PR #349, opened 2013-05-01
MSN -> Skype

matthew-kerle:
- This commit replaces MSN with Skype. All MSN usernames are converted
  to Skype format.

_Side Note: there is actually a fair amount of Skype API for call /
chat / conference buttons, might be a fun mod eventually_

Signed-Off: Matthew "Labradoodle-360" Kerle lab360@simplemachines.org


emanuele45:
Please note that this will break badly the upgrade.
You don't have to change any upgrade_{something}.sql where {something} is **not** 2_1!

What you have to do is restore the current changes you have done to upgrade_1_0 and upgrade_1_1 and add the ALTER TABLE to upgrade_2_1.sql alone (for all the three database files with the appropriate differences in the query if any).

matthew-kerle:
Yeah, thanks for catching that. Josh explained it to me, we should be all set now :)

Issue #326, opened 2013-04-02
Immediate registration


redonesmf:
By default SMF allows "immediate" registration for new members. My suggestion is that this is removed by default instead adding some new checks for newly registered members.

matthew-kerle:
Thanks for pointing this out @redonesmf, I definitely would agree that this should be incorporated. Are you interested in working on the Pull Request for this, if so, that'd be fantastic!

We should probably either set it to Email Activation by default, or put the dropdown for selection as part of the installation script. I wouldn't be opposed to either route, really.

redonesmf:
I don't mind helping, its been quite some time since I last went through any SMF code to be honest.

matthew-kerle:
Well if you'd like to that'd be awesome. If not, I'll do my best to get to it here at some point. If you have any specific questions feel free to shoot me a message somewhere.

illori:
this has been merged [and fixed] please close

matthew-kerle:
This was merged in earlier...closed!

PR #351, opened 2013-05-02
Issue: #326 - Immediate Registration

matthew-kerle:
- Registration Mode: Added the ability to select the registration mode
  upon installing SMF to provide further control from unwanted
  registrations before the setting can be changed from the admin panel or
  for people who are unaware it exists. The setting defaults to Immediate
  Registration (registration_mode = 0).
- Cleaned up the syntax of the forum_settings template, removed invalid
  HTML5.

Reference: Issue - #326 by @redonesmf

Signed-Off: Matthew "Labradoodle-360" Kerle |
lab360@simplemachines.org

### DIFF
--- a/other/install.php
+++ b/other/install.php
@@ -957,6 +957,17 @@ function ForumSettings()
 				updateSettingsFile(array('db_character_set' => 'utf8'));
 		}
 
+		//-- Set the registration mode...
+		$smcFunc['db_query']('', '
+			UPDATE {db_prefix}settings
+			SET value = {int:reg_mode}
+			WHERE variable = {string:this_variable}',
+			array(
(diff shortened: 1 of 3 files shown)


Oldiesmann:
Merged the changes in, but I still don't understand GitHub's "Use the command line" instructions. When I do it that way it fixes the changes locally but doesn't push them to the pull request :|

other/install.php
```diff
@@ -957,6 +957,17 @@ function ForumSettings()
 				updateSettingsFile(array('db_character_set' => 'utf8'));
 		}
 
+		//-- Set the registration mode...
```
Arantor, line 960: This query is made before the settings table is even set up. It is DatabasePopulation() which performs the creation of the relevant table.

Better still, why not just integrate it directly into the install file and populate it when the table is created, like $smf_version or $current_time are?

PR #353, opened 2013-05-05
Add an ignoreProtected parameter for removing and adding members to grou...

colinschoen:
...ps. Signed off by Colin Schoen colin@simplemachines.org

### DIFF
--- a/Sources/Subs-Membergroups.php
+++ b/Sources/Subs-Membergroups.php
@@ -209,9 +209,10 @@ function deleteMembergroups($groups)
  * @param array $members
  * @param array $groups = null if groups is null, the specified members are stripped from all their membergroups.
  * @param bool $permissionCheckDone = false
+ * @param bool $ignoreProtected = false
  * @return boolean
  */
-function removeMembersFromGroups($members, $groups = null, $permissionCheckDone = false)
+function removeMembersFromGroups($members, $groups = null, $permissionCheckDone = false, $ignoreProtected = false)
 {
 	global $smcFunc, $user_info, $modSettings, $sourcedir;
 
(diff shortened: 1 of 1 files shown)


Sources/Subs-Membergroups.php
```diff
@@ -312,7 +313,7 @@ function removeMembersFromGroups($members, $groups = null, $permissionCheckDone
 	$groups = array_diff($groups, $implicitGroups);
 
 	// Don't forget the protected groups.
-	if (!allowedTo('admin_forum'))
+	if (!allowedTo('admin_forum') && !ignoreProtected)
```
eurich: !ignoreProtected would validate a constant. Haven't checked the details but I'd expect a variable check like !empty($ignoreProtected). You could also set $ignoreProtected = null and do an !isset($ignoreProtected) instead.

Sources/Subs-Membergroups.php
```diff
@@ -487,7 +489,7 @@ function addMembersToGroup($members, $group, $type = 'auto', $permissionCheckDon
 	if (!allowedTo('admin_forum') && $group == 1)
 		return false;
 	// ... and assign protected groups!
-	elseif (!allowedTo('admin_forum'))
+	elseif (!allowedTo('admin_forum') && !ignoreProtected)
```
eurich: same here..

Sources/Subs-Membergroups.php
colinschoen: I could change it to a constant, good idea. I will see if anyone else has any comments otherwise I will go for it.

emanuele45: Why would you make it a constant?
I think eurich wanted to say that you forgot the $ in front of ignoreProtected, that's all. ;)

colinschoen: Woops, I didn't mean make a constant I meant change it from a boolean, but I suppose that is fine as it is in line $permissionCheckDone. Thanks, I didn't notice I forgot the $. I will update that. Cheers Ema.

colinschoen:
That should fix it.

colinschoen:
Bump

Issue #358, opened 2013-05-14
Installer doesn't work


Arantor:
Fatal error, where it's trying to insert a row to {db_prefix}settings and can't because it doesn't exist.

See https://github.com/SimpleMachines/SMF2.1/pull/351 for more information.

(I should add, this was reported to us on wedge.org in the hope that we could answer their question.)

Devilwillcry4you:
Tried installing with the fix above and still getting  the same error.
Table 'skinsn_smf.smf_settings' doesn't exist

emanuele45:
Tested with both MySQL and SQLite.
Always completed the installation process.

Are you sure you removed the relevant code from install.php?

Just to be sure, download [this package](https://github.com/emanuele45/playpen/archive/install_fix.zip).

Devilwillcry4you:
Thanks,  I got it working last night.

colinschoen:
Ema, can this be merged soon as it is inhibiting anyone from installing the 2.1 repo?

wsabransky:
Tested @emanuele45 's fix. No problems. Should be good to merge.

Arantor:
Fix is confirmed as working as expected in all respects.

illori:
@Oldiesmann can you close this it is fixed

matthew-kerle:
As per Illori's request, closed!

PR #359, opened 2013-05-14
This should fix issue #358

emanuele45:
Signed-off-by: emanuele emanuele45@gmail.com


ghost:
Getting the below error on step 3 page after implementing this fix locally.
Notice: Undefined variable: db_persist in install.php on line 364

Oldiesmann:
@ziycon I'm not sure how that error can be caused by this fix, as this doesn't change/add anything related to a $db_persist variable.

ghost:
@Oldiesmann I placed here as this was 'installer' related and I only noticed after this issue, in hindsight I should have created another issue for it, either way I've created a PR under #369 to fix the $db_persist issue.

PR #362, opened 2013-05-15
Fix MARK ALL MESSAGES AS READ seen by guest

illori:
Fix for issue 356 

this does not exist in 2.1

``` php
$settings['show_mark_read']
```

 does not exist in 2.1


illori:
``` php
$settings['show_mark_read']
```

this is another dumb per theme setting so i was looking in the wrong place for it. imo it should not be per theme and imo it should not exist at all.

PR #364, opened 2013-05-15
Theme variants broken

illori:
fix issue 363

missing closing bracket


illori:
apparently this was not incorrect to start with. can this be reverted? thanks. see beta thread in the beta board.

illori:
@Oldiesmann can you revert the change to the index.template.php file from this commit?

PR #370, opened 2013-05-26
moderation actions when not full moderator

### LINKED ISSUE #367: moderation actions when not full moderator
if you give a user group some moderation powers [for example to approve items awaiting moderation] the link to click should be under the "more" drop down and located on the right side of the post. in this case the unapprove link shows up on the right side and not under the more drop down.

if you give permission to split a topic you can then view the area where the more should be but split does not show up, just the unapprove but without the proper css applied

if you have the permission to unapprove topics and remove messages the drop down does not show either.

looks like the drop down does not show correctly unless you give permission to modify posts

image i dont think that is how it should look

XinYenFon:
I think this resolves #367 

### DIFF
--- a/Themes/default/Display.template.php
+++ b/Themes/default/Display.template.php
@@ -579,13 +579,16 @@ function template_main()
 			echo '
 							<li class="quick_edit"><img src="', $settings['images_url'], '/icons/modify_inline.png" alt="', $txt['modify_msg'], '" title="', $txt['modify_msg'], '" class="modifybutton" id="modify_button_', $message['id'], '" style="cursor: pointer; margin: 0;" onclick="oQuickModify.modifyMsg(\'', $message['id'], '\')" />', $txt['quick_edit'], '</li>';				
 		
+			echo '
+							<li class="post_options">';
+		
 		// Can the user modify the contents of this post?
 		if ($message['can_modify'])
 			echo '
-							<li class="post_options"><a href="', $scripturl, '?action=post;msg=', $message['id'], ';topic=', $context['current_topic'], '.', $context['start'], '">', $txt['post_options'], '</a>';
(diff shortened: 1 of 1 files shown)


illori:
this does not fix the issue, it moves the more and unapproved link to the left side.

Alt Text

XinYenFon:
I'm using latest GitHub revision.
As the Usergroup (with problematic permission)
cap1
As admin
asadmin

illori:
i took the whole file from your repo so it should be the same version you are using

the user group has the following moderation permissions:
approve items awaiting moderation

your screenshot is after the thread has been unapproved, mine is before it has been unapproved.

XinYenFon:
Still same for me, can you attach or pastebin your display.template? Please use pastebin.mozilla.org because of my government's stupid blocking rules I can't use pastebin.

illori:
it is the same exact file from here https://raw.github.com/Antes/playground/ee0c11cc7685b7cec094b31c6329375ed5ddd94e/Themes/default/Display.template.php which is the commit from above.

i copied that over my existing file

XinYenFon:
Both are same and I'm sorry I have no clue why its happening.

illori:
even tested in firefox and same issue i am seeing here.

illori:
i looked into the issue further, what i am seeing in part can be reproduced if you do not allow the user to reply to the topic but can approve the messages, so further fixes may be required

if the quote button is missing the "more" button has nothing to attach to or so it seems.

also i may recommend along with this fix to disable the more button to automatically act as a modify button if the user can modify the message. people will tend to click on the more button and it should not act as a clickable button.

illori:
http://pastebin.ca/2385380

``` css
.quickbuttons li.post_options {
    padding: 0 4px 0 4px;
    border-radius: 1px 4px 4px 1px;
}
```

PR #385, opened 2013-06-02
Post Moderation should obey amount of messages per page

NanoSector:
As reported here:
http://www.simplemachines.org/community/index.php?topic=504033.msg3546394#msg3546394

Hereby a fix is provided.

### DIFF
--- a/Sources/PostModeration.php
+++ b/Sources/PostModeration.php
@@ -213,8 +213,11 @@ function UnapprovedPosts()
 	);
 	list ($context['total_unapproved_topics']) = $smcFunc['db_fetch_row']($request);
 	$smcFunc['db_free_result']($request);
+	
+	// Limit to how many? (obey the user setting)
+	$limit = !empty($options['messages_per_page']) ? $options['messages_per_page'] : $modSettings['defaultMaxMessages'];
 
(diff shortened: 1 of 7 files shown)


NanoSector:
Ugh. I hate Git, it has no logic. If added can you please apply the _second commit only_ only in a new commit? Thanks.

emanuele45:
This is not enough.
Changing this you just change the paginations, but not the amount of "rows" showed.
To completely fix the issue, you have also to change the query few lines below the one you changed, in particular:

```
LIMIT ' . $context['start'] . ', 10',
```

to something like:

```
LIMIT ' . $context['start'] . ', ' . !empty($options['messages_per_page']) ? $options['messages_per_page'] : $modSettings['defaultMaxMessages'],
```

even though ti would be much nicer to fix that query to use the the array to pass values...

NanoSector:
This okay? :)

emanuele45:
Much better, yes.

NanoSector:
I'll fix it up for a merge tomorrow.

Oldiesmann:
Yoshi, to remove the unwanted commit from this, switch to the branch and run this command:

```
git revert --no-edit -s 1e1a893
```

NanoSector:
resubmitting.

PR #390, opened 2013-06-04
Added MySQLi database abstraction layer

gnif:
This commit adds a MySQLi abstraction layer wihout duplication all the code in DbExtra-mysql.php etc. This should be used as the default as the mysql_\* API is being deprecated as well as has problems with persistant connections on versions of PHP later then 4.0.

See: http://php.net/manual/en/mysqli.persistconns.php

> The problem with persistent connections is that they can be left in unpredictable states by clients. For example, a table lock might be activated before a client terminates unexpectedly. A new client process reusing this persistent connection will get the connection "as is". Any cleanup would need to be done by the new client process before it could make good use of the persistent connection, increasing the burden on the programmer.
> 
> The persistent connection of the mysqli extension however provides built-in cleanup handling code.

### DIFF
--- a/Sources/ManageMaintenance.php
+++ b/Sources/ManageMaintenance.php
@@ -131,10 +131,10 @@ function MaintainDatabase()
 	global $context, $db_type, $db_character_set, $modSettings, $smcFunc, $txt, $maintenance;
 
 	// Show some conversion options?
-	$context['convert_utf8'] = $db_type == 'mysql' && (!isset($db_character_set) || $db_character_set !== 'utf8' || empty($modSettings['global_character_set']) || $modSettings['global_character_set'] !== 'UTF-8') && version_compare('4.1.2', preg_replace('~\-.+?$~', '', $smcFunc['db_server_info']()), '<=');
-	$context['convert_entities'] = $db_type == 'mysql' && isset($db_character_set, $modSettings['global_character_set']) && $db_character_set === 'utf8' && $modSettings['global_character_set'] === 'UTF-8';
+	$context['convert_utf8'] = ($db_type == 'mysql' || $db_type == 'mysqli') && (!isset($db_character_set) || $db_character_set !== 'utf8' || empty($modSettings['global_character_set']) || $modSettings['global_character_set'] !== 'UTF-8') && version_compare('4.1.2', preg_replace('~\-.+?$~', '', $smcFunc['db_server_info']()), '<=');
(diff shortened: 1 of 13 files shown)


Oldiesmann:
This is a good start, but you need to make MySQLi versions of the other two files (DbExtra and DbPackages) as well, and we'll need an install sql file so users can actually install it with MySQLi.

Also, don't forget to sign-off on your commits. We can't accept commits that haven't been signed-off on

gnif:
@Oldiesmann: See the hack on lines 36-37, this prevents it needed these files as the SQL syntax has not changed, just the extension. There is also code in the site that checks if the db_driver is 'mysql' and has special use cases for it, so it is important that the driver still report as mysql.

As for signing off, will do.

gnif:
Hrmm, this hack is proving not worth its trouble, let me re-work this.

gnif:
Pull request updated, now supports the installer and upgrade script. I have tested the installer but I am unable to test the upgrade script as I do not have a SMF forum to upgrade.

illori:
if you wish for this to be merged in please sign off your PR and commits so that you agree to the DCO
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/DCO.txt

eurich:
the commit is signed-off, AFAIK that's fine it is.

illori:
well everyone else has been asked to sign off their PR's so i assume there is a reason for doing so.

emanuele45:
Probably who asked didn't know that sign the PR is not particularly useful, and the most important thing is sign each commit (this is the only thing "brought around" in the history, not the "PR"). :wink:

norv:
Only the commits need sign-off. As per the existing development procedure on copyright/licensing, documented in the readme in the repository.

When starting up on github, there have been only a few cases when PRs have been signed-off, because people have forgot to do it on commits, and they were non-copyrightable commits anyway. So by exception we have accepted a few sign-offs on PR. But it's not otherwise meaningful or too useful.

Attempts to make sign-off on PRs into a "rule" of sorts are meaningless and not thought-out.

norv:
Thank you for the work on this, @gnif. I think you're right, the explicit support for mysqli is a good thing.

gnif:
New lines removed, vim was trying to be smart and adding them for C/C++ development.

gnif:
Squashed the commits into a single clean commit.

gnif:
If this PR is ok, can you please merge it... it would be great to get this in for 2.1 as many hosts will need this functionality.

PR #401, opened 2013-06-05
jQuery 2.0.2

XinYenFon:


jdarwood007:
Are we sure nothing broke by this? 1.9 had BC breakage and removed support for certain browsers.  We need to make sure that things still work and that we still have the BC support we need.

emanuele45:
For certain it makes IE 8 unsupported. :angel:

matthew-kerle:
I was thinking about this when I was at work, but ema is right, it breaks some things for IE8. The jump unsupports IE6-8. If I recall correctly, our stance was to keep IE8 support which would mean we can't merge this...

Oldiesmann:
We definitely shouldn't break IE8 compatibility - IE8 is the latest version available for Windows XP, which is still (sadly) fairly well-used.

butch2k:
The jQuery 1.9.x branch is equivalent to the 2.0.x branch in
functionnality, the later dropping the IE6-8 support and being as such a
bit lighter but w/o any added value.

XinYenFon:
Ouch this is my mistake sorry, I thought IE9 is available for windows XP, or we can use jQuery migrate.

matthew-kerle:
I'm sure IE9 is in fact available for Windows XP, but what Michael is saying is that IE8 is what's installed by default. And at this point we're keeping IE8 support.

No, migrate is a really bad idea for something as large as SMF. There's no way we'd distribute that.

emanuele45:
So you are saying Microsoft is wrong?
http://windows.microsoft.com/en-us/internet-explorer/products/ie-9/system-requirements
http://answers.microsoft.com/en-us/ie/forum/ie9-windows_other/ie9-for-xp-why-doesnt-internet-explorer-9-work-on/e8113f20-b149-4763-b4d4-562d1da524b6

XinYenFon:
I didn't want to link Turkish, XP cames with IE6 and IE7/8 upgrade option.

jQuery Migrate: https://github.com/jquery/jquery-migrate/#readme

matthew-kerle:
No. I was just saying that at the very least IE8 was pre-installed on Windows XP.

And I know very well what jQuery migrate is and how it handles things, it's not going to be implemented.

butch2k:
jQuery migrate is mostly useful for the $.browser calls as in jQuery 1.9+
feature detection is the way to go. Using jQuery 1.9 and Modernizer is the
recommended way.

eurich:
For the Record: The first version of XP came with IE6 pre-installed. Dunno if they changed that over the years, maybe the latest CDs / DVDs contained IE8 "out of the box"?!?

PR #420, opened 2013-06-12
Clarify file type

colinschoen:
http://www.simplemachines.org/community/index.php?topic=505692.0

### DIFF
--- a/Themes/default/languages/Themes.english.php
+++ b/Themes/default/languages/Themes.english.php
@@ -14,7 +14,7 @@ $txt['theme_remove'] = 'remove';
 $txt['theme_remove_confirm'] = 'Are you sure you want to permanently remove this theme?';
 
 $txt['theme_install'] = 'Install a New Theme';
-$txt['theme_install_file'] = 'From a file';
+$txt['theme_install_file'] = 'From an archive (zip)';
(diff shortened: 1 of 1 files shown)


matthew-kerle:
Yeah, I read the thread. I still think it might be best to put "ie." or something similar so it doesn't appear to be the only file type.

illori:
@colinschoen please update your PR per the suggestion in the thread

matthew-kerle:
Colin, any word here? I'd like to get this in.

Arantor:
Well, there are only two formats accepted in the bowels of the package system (which this sort of reuses), which means you're either dealing with .zip or .tar.gz/.tgz (which are the same thing with a different name)

You could just list all three, or as a best compromise, .zip/.tar.gz even though pretty much every theme does say .zip on it. (And technically, it would be e.g. not i.e. if you're getting picky. e.g. is for giving an example, i.e. is for stating 'that is to say' for a definite answer rather than an example)

Arantor:
I actually took the comments above into consideration to reword it, and also in future please note that we cannot merge in commits that are not signed off. Signing off constitutes providing a DCO for the commit in question.

illori:
thanks arantor.

PR #425, opened 2013-06-14
Make the post moderation center obey max posts

NanoSector:
As reported here:
http://www.simplemachines.org/community/index.php?topic=504033.msg3546394#msg3546394

Hereby a fix is provided.

PR #427, opened 2013-06-15
Yoshi postmoderationfix

emanuele45:
Reference: #425
Signed and added the global.

Thanks @Yoshi2889 for the code. ;)


matthew-kerle:
Again, thanks to Norv for pointing out the globals. And good catch on the start, I was wondering about that myself.

PR #429, opened 2013-06-15
Fix background image on infobox... why a checkmark?

illori:


Oldiesmann:
For some reason there's a merge conflict with this, though I have no idea why.

norv:
Probably because the commit also contains a change to Help english file, which has been already pulled in IIRC.

illori:
no it is not the same fix, it is similar. it has that _blank fix.

maybe that will help fix this?

PR #442, opened 2013-06-18
Move report to moderator to post area instead of dropdown

ghost:
Signed-off-by: Bryan Deakin runic@simplemachines.org


emanuele45:
And why not into the dropdown? :confused:

PR #443, opened 2013-06-18
Here we go: all happy the db backup is gone

emanuele45:
There may be left some unused functions in some db-layer, but at least the basics are there: db backup is gone once and for all. No more complains for broken databases.


matthew-kerle:
Thanks ema! Excited about this one. Oldies, can we merge it? :P

emanuele45:
That's really, really odd you are exited about remove something (that works, except nobody has ever wasted time to test it) since remove is the easiest thing to do in the world.

I'm instead a bit sad to have wasted time fixing something in the hope it would have been useful to someone just to discover nobody cares about it....meh.
Next time I'll be more careful and will not spend my time at all.

Kindred-999:
wow...  I am upset that
a) no one (apparently) tested that emanuele fixed the problem
b) people insisted that it be removed, even after it was fixed
c) Oldies merged the pull, based on a and b

PR #456, opened 2013-06-20
Fix to check version of files correctly on upgrade.

ghost:

Change to allow file versions to be checked correctly on upgrade.

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -987,7 +987,7 @@ function WelcomeLogin()
 
 	// Do a quick version spot check.
 	$temp = substr(@implode('', @file($boarddir . '/index.php')), 0, 4096);
-	preg_match('~\*\s@version\s+(.+)[\s]{2}~i', $temp, $match);
+	preg_match('~\*\s@version\s+(.+)[\s]{2}\s~i', $temp, $match);
 	if (empty($match[1]) || $match[1] != SMF_VERSION)
 		return throw_error('The upgrader found some old or outdated files.<br /><br />Please make certain you uploaded the new versions of all the files included in the package.');
 


emanuele45:
Why do you need that? ???
The \s is already there...

ghost:
It's failing on some setups without the extra \s at the end of the pattern.

emanuele45:
Know exactly on what setups fails would probably help in define if it is an error in the code or in some specific version of php or whatever.
As far as i can see that code has been the same for a while now and I can't remember issues (not saying of course you don't have any).

Anyway, since the \s is already there, something like:
`preg_match('~\*\s@version\s+(.+)[\s]{1,2}~i', $temp, $match);`
would probably lead to the same result.
That said, I have two doubts:
1) no idea why the \s is within square brackets (I know it was there, just thinking out loud),
2) why the fix you proposed should fix the issue since it requires one more space and so it should be more restrictive... :confused: (not saying the fix is wrong, I'm just a bit confused by the fact that an apparently more strict pattern matches the result, while a more relaxed one fails)

ghost:
From further investigation I've found that as the code is:
`preg_match('~\*\s@version\s+(.+)[\s]{2}~i', $temp, $match);
if (empty($match[1]) || $match[1] != SMF_VERSION)`

it fails due to `$match[1]` returning 'SMF 2.1 Alpha 1 ' and `SMF_VERSION` returns 'SMF 2.1 Alpha 2.1', the difference being that `$match[1]` returns extra white space at the end so the if statement passes.

So the easiest fix I could think of was to replace:
`preg_match('~\*\s@version\s+(.+)[\s]{2}~i', $temp, $match);`
with:
`preg_match('~\*\s@version\s+(.+)[\s]{2}\s~i', $temp, $match);`

The reason I decided this was due to myself so far being the only person coming across this issue that I know of I'm putting it down to my setup. The only other thing I can think of that could cause this to happen could be the changes in the copyright in all the files where the new copyright added in extra white space.

emanuele45:
The point is that if it worked well so far on all the other systems and is failing only in yours, changing the code the way you did would potentially make it fail on any other system except yours... a bit unfair :P
`preg_match('~\*\s@version\s+(.+)[\s]{2,}~i', $temp, $match);`
a minimum of two spaces should work as well and avoid any possible drawback.

ghost:
I get your point, I'm going to try and get someone to test it out based on my settings. Will post back once I have some results.

matthew-kerle:
Any word on this, @ziycon?

ghost:
@Labradoodle-360, been trying to get people to replicate the issue but there are other issues occurring also on upgrade from 2.0 to 2.1 alpha so working on those before getting to this.

PR #463, opened 2013-06-24
Ban fixes

emanuele45:
Ban needs quite a bit of testing.
TBH I'm not sure I ported **all** the fixes, I may have missed something.

The PR includes also the fix for:
http://www.simplemachines.org/community/index.php?topic=506463.0

### DIFF
--- a/Sources/ManageBans.php
+++ b/Sources/ManageBans.php
@@ -365,175 +365,176 @@ function BanEdit()
 		foreach ($context['ban_errors'] as $error)
 			$context['error_messages'][$error] = $txt[$error];
 	}
-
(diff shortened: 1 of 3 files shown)


Oldiesmann:
Went ahead and merged this in so we can start testing it (Lab and I are hoping to get a beta out soon). Anything in particular we should be looking for?

emanuele45:
Good question, I suppose if the bans and the triggers are properly inserted from all the possible places (add new ban/trigger, edit trigger/ban, the profile, etc.), both in case of correct inserts and in case of errors (i.e. posting a non-valid IP or host), etc.

A bit all the combinations...

PR #472, opened 2013-06-27
fixed the "black background" problem on Firefox

PortaMx:
add a background:#ffff to jquery.sceditor.default.css

Signed-off-by: PortaMx suport@portamx.com


ghost:
kinda cant replicate this :/

Oldiesmann:
None of us have been able to replicate it. Illori even tested FF22 in XP (on a virtual machine).

matthew-kerle:
Even if this was merged, we'd probably want it to be: background: #fff; instead.

PortaMx:
Well .. after long search, I found the problem .. it was a old css class from a theme was I have long time ago installed on my FF ... (gruble) .. so forgott this.

But .. I found a other problem ..
make a code area with long lines they not break.
If you now scroll the code area, the background for the code stopped on the right edge of the editor range.
This can by simple fixed with change the .code css from display:block to display:inline-block.
Now this will show correctly ...

Oldiesmann:
I'll go ahead and close this one then. We can submit a separate PR for the code issue.

PR #486, opened 2013-06-28
Easy toggle

NanoSector:
Signed-off-by: Rick Kerkhof rick.2889@gmail.com (my commit is signed)

Prefered method of creating toggles is toggleElementEvent(), because it creates the event and assigns the correct class to the image (if any).

Syntax (toggleElement):
toggleElement(element to toggle (jQuery selector), the item that toggles it (jQuery selector), any image options (see below), the duration of the animation (string or int))

toggleElementEvent is nearly identical, with exception of the last parameter:
toggleElementEvent(element to toggle (jQuery selector), the item that toggles it (jQuery selector), any image options (see below), is collapsed? (bool))

imageoptions consists of:
visibleimage: The image that is shown when the toggled element is visible. (default: upshrink.png, URL to image path)
hideimage: The image that is shown when the toggled element is hidden. (default: upshrink2.png, URL to image path)

visibletxt: The hover text that is shown when the toggled element is visible (default: hide_txt, or "Hide", string)
hidetxt: The hover text that is shown when the toggled element is hidden (default: show_txt or "Show", string)
Both visibletxt and hidetxt are disregarded when the toggle "button" is NOT an image.

isimage: If the toggle "button" is an image. (bool)
eimage: An external image, if wanted. (jQuery selector)

This function aims to replace smc_Toggle.

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -1813,6 +1813,9 @@ function loadTheme($id_theme = 0, $initialize = true)
 		'ajax_notification_text' => JavaScriptEscape($txt['ajax_in_progress']),
 		'ajax_notification_cancel_text' => JavaScriptEscape($txt['modify_cancel']),
 		'help_popup_heading_text' => JavaScriptEscape($txt['help_popup']),
+		'header_is_collapsed' => JavaScriptEscape(!empty($options['collapse_header'])),
+		'show_txt' => JavaScriptEscape($txt['show']),
+		'hide_txt' => JavaScriptEscape($txt['hide']),
 	);
 
 	// Add the JQuery library to the list of files to load.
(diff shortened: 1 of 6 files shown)


Oldiesmann:
What's the advantage of this over the current functionality, besides using jQuery? Also, if a guest collapses/expands something, is that remembered across page loads like SMF currently does (with cookies)?

NanoSector:
"You don't need the single quotes around true and false here unless you specifically want to pass them as strings instead of boolean values."
JavaScript needs them as a string format.

"What's the advantage of this over the current functionality, besides using jQuery?"
It doesn't require 30 lines of complicated code which nobody understands.

"Also, if a guest collapses/expands something, is that remembered across page loads like SMF currently does (with cookies)?"
No, but that can be changed.

Oldiesmann:
If you can have it remember things for guests (SMF currently uses a cookie to do that), then we should be good to replace the existing code with this.

illori:
also the current code has a glitch in it, on the board index if you collapse and expand the header when you expand it over expands and has to collapse up.

NanoSector:
I did not fix that in this pr, fix it by adding width: 100% on #inner_something.

PR #496, opened 2013-07-04
admin.css remove images

illori:
Credit goes to ARG


emanuele45:
Small suggestion: if the credit goes to someone else, better specify it in the commit message rather than in the PR message. ;)
I usually use the structure:

> commit message - thanks to ARG for the suggestion/code/report/whatever

so in that case for example:

> admin.css remove images - thanks to ARG for the ...

Issue #93, opened 2012-07-26
WYSIWYG: missing features / bugs


eurich:
the wysiwyg is nice in general but two features are IMHO important have been wiped out:
1) the button-status isn't updated anymore, thus you don't knwo the current selected formattings..
2) the toggle button to switch between wysiwyg and non-wysiwyg mode

emanuele45:
1) https://github.com/samclarke/SCEditor/issues/25

2) it's there and is the same (and same place) as in SMF...at least here...did I forgot to add something?

eurich:
2) yep, found it :P

Trekkie101:
1. URLs added without http:// screw up.

ChainyEO:
Is an old version of this SCEditor currently being used in SMF2.1? Just that the Button State still appears to be missing in the editor as used in SMF.

Oldiesmann:
Looks like it. The comment at the beginning of jquery.sceditor.js says it's 1.3.7. I'll look into getting that updated soon.

@Spuds, is this something you can help us with?

emanuele45:
The one in the branch master is usually the latest: SCEditor v1.4.3 (rather big on the first page of the repo :P).

That said 85% of the editor would work out of the box just copying the files, some more exotic (like...dunno the [nobbc][ol], [quote], [code][/nobbc]) requires some rewrites of the internals in order to work _properly_.

What I did the first time was to just edito the SCEditor "core" files, though this is not very cleaver because it then requires more maintenance in order to upgrade.

Spuds added another file with Elk/SMF specific code.

The only thing the most likely cannot be added extending the core is the ability to show the buttons on two lines, that requires a change that may not be worth copy just for that, but dunno.

PR #525, opened 2013-07-23
Update sceditor to latest version

Oldiesmann:
This PR takes the changes made to SCEditor 1.3.7 (or whatever version we currently have in 2.1) and applies them to the latest version (1.4.3 as of this PR). A complete list of changes applied is available at http://smfgit.oldiesmann.us/index.php?topic=4.0. A brief test on my forum indicates that this should work fine, but I'd like others to test it as well before we pull this in, especially since three blocks of code which we modified in jquery.sceditor.bbcode.js weren't found in the new version.

I have not made any changes to SMF itself for this, so other tweaks may be required as well.

Ref: #93

### DIFF
--- a/Themes/default/css/jquery.sceditor.default.css
+++ b/Themes/default/css/jquery.sceditor.default.css
@@ -6,6 +6,7 @@ html, body, p, code:before, table {
 html>body>p, html>body>div {
 	font: 83.33%/150% "Segoe UI", "Helvetica Neue", "Liberation Sans", "Nimbus Sans L", Arial, sans-serif;
 }
+
 ul, ol {
 	margin-top: 0;
 	margin-bottom: 0;
@@ -31,7 +32,9 @@ code {
 	border: 1px solid #aaa;
 	white-space: pre;
 	padding: .25em;
+	display: block;
 }
+
 code:before, code {
 	display: block;
 	text-align: left;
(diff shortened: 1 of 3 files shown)


XinYenFon:
After the patch

patch

Oldiesmann:
I'll look into that and see why the code I added to handle rows isn't working.

I am also seeing a couple errors in the JS:

`[01:29:11.667] TypeError: $.sceditor.setCommand is not a function @ http://smfgit.oldiesmann.us/Themes/default/scripts/jquery.sceditor.js?alph21:5397`

`[01:29:12.376] TypeError: $(...).data(...).createPermanentDropDown is not a function @ http://smfgit.oldiesmann.us/index.php?action=post;topic=4.0;last_msg=13:431`

I'll have to see what's going on with those as well.

Oldiesmann:
Not sure what the issues are here...

It seems to be using its own toolbar (with our buttons) rather than ours, so no rows are being defined. It also seems to be ignoring the fact that setCommand and createPermanentDropDown are defined.

@norv, @Spuds or @emanuele45 - any help on this?

emanuele45:
mmm...assuming what you want are the two lines of icons, that's one of the few changes that must be done to the core of the editor.
At the moment I don't have the code handy, will check later.

Spuds:
TypeError: $.sceditor.setCommand (  ... this was replaced with $.sceditor.command.set(  in the 1.4 branch

TypeError: $(...).data(...).createPermanentDropDown ... not sure, if you are sure its initialized then it could simply be in the wrong namespace.

two lines of icons ... that was emanuales thing ... you could try (this is a guess) placing all of the top row ones in to a single group by only using the array(), where you want the new row, then target the css groups "sceditor-group" to flow as you want

Oldiesmann:
Closing this one in favor of #526.

PR #526, opened 2013-07-25
SCEditor updated to 1.4.3 - thanks @Spuds for the directions ;D

emanuele45:
Replaces #525

Before merge: obviously test it, but also please note that a (css) class for the active button is missing. I if remember it I'll add one tomorrow morning, otherwise open an issue because something should be added.


emanuele45:
Also: almost all the SMF-specific code is in the new file.
I did 3 (or 4 don't remember) changes to the original SCEditor files:
1) one to remember the tabindex
2) one to avoid that url and irul got mixed
3) one to support multiple lines of buttons
4) one to support multiple lines of smileys
3 and 4 are questionable and I would drop the functionalities from SMF itself (also because very few use them anyway.
2 is questionable as well tough it affects anyone (including people not using the WYSIWYG, so is not easy to ignore).

Themes/default/scripts/jquery.sceditor.smf.js
```diff
@@ -0,0 +1,693 @@
...
+		txtExec: ["[pre]", "[/pre]"],
+		exec: function () {
+			this.wysiwygEditorInsertHtml('<pre>', '</pre>');
+		}
+	}
+);
+$.sceditor.command.set(
+	'move', {
+		tooltip: 'Teletype',
+		txtExec: ["[[move]", "[/move]"],
```
XinYenFon: Not a big deal but extra [ here :)

Oldiesmann: Tooltip is wrong as well, though that's not a big deal since we're pulling in tooltips from the language files anyway.

XinYenFon:
Glow & Shadow are not working on input(text) area, but no problem on posts.
glowshadow

Themes/default/scripts/jquery.sceditor.smf.js
emanuele45: Not true, because the text there **is** the index that identify the string in the language file **and** en_US doesn't use a language file.

emanuele45:
@Antes glow and shadow are not meant to work in WYSIWYG mode... :stuck_out_tongue: 
Seriously, no, never even tried to make them work...

emanuele45:
And: pushed a couple of other commits, including update of language strings too.

Oldiesmann:
Anything else that needs to be done before we can pull this in?

emanuele45:
1st answer the questions.
2nd verify why Antes has all the icons on the same line even though there is a separator after the color picker.

XinYenFon:
@emanuele45 its not a bug or something oh sorry to say, its happening when you resize the editor and give it more width (no CSS).

Oldiesmann:
@emanuele45 Which questions?

Oldiesmann:
@emanuele45 Any updates? I'd like to get this merged in soon so we can work on getting a beta out.

emanuele45:
Well, apparently nobody cares, so the best option would be to just remove the edits I did to the original SCE code and live with the differences.
I would just keep the edit 2 (as per https://github.com/SimpleMachines/SMF2.1/pull/526#issuecomment-21592168) and remove anything else.

XinYenFon:
[color=purple][b][center]sadasdasdasd[/center][/b][/color] > Center breaking color and b code.
[center][color=purple][b]sadasdasdasd[/b][/color][/center] > this has no problem.

[b][move]asdasdasdasd[/move][/b] > Move breaking B code (probably any other bbcode)
[move][b]asdasdasdasd[/b][/move] > this has no problem

Oldiesmann:
@emanuele45: "Nobody cares" about what? I still haven't gotten an answer as to what (if anything) needs to be done before this can be merged in. That's all I want to know.

@Antes: Does it work when the actual post is made? I will admit that many people won't know which order is "proper" for the tags, but if the messed-up ordering works when the post is made, it might be something we can live with. If it's still broken when the post is made, then we'll need to look into it further.

XinYenFon:
no its not. I think same happened with bold [b]

emanuele45:
@Antes that's parse_bbc behaviour and it has always been like that since 1.0 (search the bug reports board for bbc block or similar): block-level tags (i.e. center, code, quote, br, hr, list, move, probably others as well) break non-block level.

@Oldiesmann it's not very difficult to follow a link (assuming it's difficult to read the discussion from the beginning and not stumble on the very first comment I made to this PR, on the other hand you are right, it was not formulated in the form of a question, though they were in the [blog post](http://www.simplemachines.org/community/index.php?topic=508994.0) (<== this is a link) I made about a month ago).

Just to make it easier to find again: should we keep the 4 edits to SCEditor code and make it slightly more difficult to upgrade it, or forget about them (so that upgrade means simply replace the code and update the plugins) and live with the different behaviour from previous versions of SMF?

emanuele45:
Closing this in favour of #599

PR #537, opened 2013-07-26
More PM-related bugfixes

### LINKED ISSUE #532: warning sent to user via pm has link to user with id 0
when the guest pm is sent the users name is a link to the profile, in this case to the profile with id =0 which defaults to the user that is viewing the message, if the user is a guest that sent the message the link to profile should not show on the username.

http://localhost:8080/github2.1/index.php?action=profile;u=0

also somewhat related to this when you hover over the username a partial popup shows which has no information as the user is a guest.

the link to the profile with user id 0 also occurs when a guest posts in the forum. but in that case the popup has some information filled in.

Oldiesmann:
Fixes #532

### DIFF
--- a/Sources/PersonalMessage.php
+++ b/Sources/PersonalMessage.php
@@ -809,8 +809,8 @@ function MessageFolder()
 
 			// Seperate query for these bits!
 			$subjects_request = $smcFunc['db_query']('', '
-				SELECT pm.id_pm, pm.subject, pm.id_member_from, pm.msgtime, IFNULL(mem.real_name, pm.from_name) AS from_name,
-					IFNULL(mem.id_member, 0) AS not_guest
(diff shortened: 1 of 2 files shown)


Sources/PersonalMessage.php
```diff
@@ -809,8 +809,8 @@ function MessageFolder()
 
 			// Seperate query for these bits!
 			$subjects_request = $smcFunc['db_query']('', '
-				SELECT pm.id_pm, pm.subject, pm.id_member_from, pm.msgtime, IFNULL(mem.real_name, pm.from_name) AS from_name,
-					IFNULL(mem.id_member, 0) AS not_guest
+				SELECT pm.id_pm, pm.subject, IFNULL(pm.id_member_from, 0) AS not_guest, pm.msgtime, IFNULL(mem.real_name, pm.from_name) AS from_name,
+					mem.id_member
```
emanuele45, line 813: /me feels this break https://github.com/Oldiesmann/SMF21/blob/4a0450d59ac272c8897aa8fcaf624e2a2aa23e59/Sources/PersonalMessage.php#L914 (not checked, just guess)

PR #539, opened 2013-07-26
Don't link to "issue warning" when viewing your own profile

### LINKED ISSUE #533: issue warning visible for admin in profile
Profile.php
now

``` php
'enabled' => in_array('w', $context['admin_features']) && $modSettings['warning_settings'][0] == 1 && (!$context['user']['is_owner'] || $context['user']['is_admin']),
```

fixed

``` php
'enabled' => in_array('w', $context['admin_features']) && $modSettings['warning_settings'][0] == 1,
```

if you are admin the link to issue warnings shows up then you get an error that you are not allowed to access this section.

the link does not show up for your own profile if you are not admin, so why show it to admin?

i think there was some discussion about letting users view their warnings, so maybe this was part of it?

Oldiesmann:
Fixes #533

### DIFF
--- a/Sources/Profile.php
+++ b/Sources/Profile.php
@@ -295,12 +295,12 @@ function ModifyProfile($post_errors = array())
 				),
 				'issuewarning' => array(
 					'label' => $txt['profile_issue_warning'],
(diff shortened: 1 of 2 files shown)


emanuele45:
Just a small suggestion: if you plan to fix few issue, it's not necessary to send 10 PRs for 10 bug fixes: fix all the bugs you want, and then send a single PR. Less emails for everybody. :P

Issue #40, opened 2012-05-25
SQLite compatibility


norv:
SMF 2.0 is only supports officially SQLite 2.8.
SQLite 3.x is backwards incompatible (or not guaranteed) with 2.x.
The standard PHP driver for SQLite 3.x is the PDO driver. Which SMF 2.0 database layer does not support.

So the question is: are we going to implement SQLite 3.x support for 2.1, or just fix 2.8?
My previous answer: support 3.x in _any_ next version of SMF.

Notes:
2.x is extremely outdated. (this is an understatement)
3.x is almost certainly on the hosts supporting SQLite.
Once a database is migrated to 3.x, one cannot go back.
They both can be installed in parallel and work side by side on a server. SQLite has designed its 3.x so that it has different functions to create/open/work with a database, to enable that.
SMF uses the SQLite 2 functions, i.e. install.php first lines: sqlite_open()
http://www.php.net/manual/en/function.sqlite-open.php

> > (PHP 5 < 5.4.0, PECL sqlite >= 1.0.0)
> > This function only support database of SQLite 2 or below. For SQLite 3, you must use PDO.

Ref:
http://www.php.net/manual/en/sqlite.installation.php

Spuds:
Hummm ... no comments ... I would have thought the 2 people using sqlite would have chimed in :D .... 

Yes it should be moved to 3/PDO however I don't think its worth the effort to make the change for such a seldom used feature.  Also given is scant usage, testing it will take forever ... really we are still finding postgre and sqlite errors from the early days of 2.0 which is all about test coverage and real life usage. 

IMO in 3.0 I would drop support for all but MySQL, this was a nice experiment but the gain vs the effort or development and support is just not worth it.  If thats not possible, maybe move on to a more robust abstraction layer, or maybe you have!

IchBin:
The database abstraction at the time was a nice shiny new object. Everyone thought it was the coolest things since sliced bread. After seeing how little it's been used and how much trouble it's been to fix and support, I'd agree at this point to let it go the way of the wind. Would be better if we just left it abstract enough that someone else can plug into it if they so desire. Off with it's head!

Fustrate:
Instead of dropping support for database engines, we should be supporting them through PDO.

We also need to get rid of this horrible abstraction layer and use something more standard. I don't care if it's our own or if we use a well-established code base, but the regex-based replacement stuff we have now is crazy, IMO.

IchBin:
[...]

PR #545, opened 2013-07-29
Removing sqlite support

ghost:
This is just a start to remove sqlite support from future versions.


emanuele45:
xref #40

Oldiesmann:
When did we decide we weren't going to support SQLite anymore? I don't think it's a good idea to just drop it, especially as there's no reason we can't keep supporting it. If it's because 2.8 is outdated, we should look into adding support for 3.x as well.

ghost:
I was starting off with this work, not that it has to be merged but gave it a start anyway, the next step would be to update the upgrader to take the sqlite database on upgrade to 2.1 and convert it to mysql, a bit of discussion would need to take place with emanuele around the licensing of his converter.

http://www.simplemachines.org/community/index.php?topic=507820.msg3577612#msg3577612

Arantor:
I can give you two good reasons not to continue supporting it.

Firstly, lack of use. Hardly anyone uses SQLite currently - and that's not just related to limited/no support for SQLite 3 in the system (I see code in 2.1 that mentions it). SMF 2.x has had SQLite support for years but even going back to RC days, hardly anyone used it. It seems a bit silly to pour all the time and effort into checking compatibility with every change that gets made knowing full well that almost no-one will use it.

Secondly, even taking into account the potential userbase, there is a much larger problem with SQLite: it just doesn't scale. It cannot scale up to the levels that MySQL and PostgreSQL can do. Given that, I do not ever recommend its use; I cannot think of a single application where SQLite would actually be appropriate (and before anyone suggests 'running it off a USB stick for specialised applications', you already have Apache and PHP to deal with and you can already get MySQL in a portable form that would suit)

To me, the benefit just doesn't outweigh the maintenance or development time requirements.

jdarwood007:
I am going to agree with removing Sqlite support.  Although I would love to see us pick up another database engine then.
I am not talking about Percona or MariaDB as those will be supported due to their current drop in replacement status for MySQL and will possibly someday need their own database layers, not at the moment though unless their is changes in the horizon that will break MySQL compatibility with them.

Arantor:
I would note that there is SQLite 3 support code in 2.1's current state, which muddies the problem even more.

That said, at this stage I'm tempted to say leave SQLite where it is unless there are major breaking issues with it; we don't really want to be doing huge overhauls at this stage of the day.

jdarwood007:
Sounds fine to me.  We leave it for the period of 2.1 and announce its removal during SMF 2.1 Final announcement, which should give enough time for any forums (I said any) to migrate away.

Arantor:
Works for me. @Oldiesmann what's your take on this?

Oldiesmann:
"Announce its removal during the final 2.1 announcement" - this means we'll be removing it in 2.2/3.0, right? If so I'm fine with that.

Arantor:
Yup that's what we're talking about.

Oldiesmann:
Sounds good to me.

ghost:
Based on the over all consensus do you want this PR removed?

Arantor:
I don't think it needs to be removed, though I think we can just close the issue for now? It may be useful in future for review purposes :)

XinYenFon:
No need to remove get this PR as patch and store it we/you can apply it on 2.2 and/or 3.0

https://github.com/SimpleMachines/SMF2.1/pull/545.patch

NanoSector:
If we aren't removing this, can we at least close this so it doesn't show up in the list?

Arantor:
Yeah, I think we should do that, just close it for now - we can always find it later.

PR #548, opened 2013-07-31
PHP 5.5 Compatibility

jdarwood007:
These are changes that are needed to support PHP 5.5.  So far this appears to be /e preg_replace fixes.  This has required changing over to callbacks and some curry functions.  Everything changed will need testing.  I've done testing to ensure things still appear to work, but am not sure if I missed something in testing.  I am also not sure if I missed any or any other 5.5 replacements are needed.

### DIFF
--- a/Sources/Admin.php
+++ b/Sources/Admin.php
@@ -127,7 +127,7 @@ function AdminMain()
 					'subsections' => array(
 						'general' => array($txt['mods_cat_security_general']),
 						'spam' => array($txt['antispam_title']),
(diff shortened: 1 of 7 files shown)


jdarwood007:
The only other note is that MySQL functions are now considered deprecated.  Should we offer this in upgrade.php if we detect they are using mysql to change over to mysqli?  It looks like somebody has already written a mysqli database file.  So we just need to move them over if they wish.  Future versions we may force it though.

Oldiesmann:
If the mysqli stuff works, I'm fine with offering to switch to that during the upgrade. I'm not sure how complete it is though - I thought it still used regular MySQL functions in some areas.

PR #554, opened 2013-08-02
// If they are for sure registered, let other people to know about it

MissAllSunday:
This is just to pass the newly created user ID for others to use.

 I needed a way to set up a notification when an user was registered and integrate_register for obvious reasons doesn't provide the user ID. 

There's no pass by reference since at this point there's nothing you can do, this is just for reference only.


ghost:
Great to see you MissAllSunday :)

PR #556, opened 2013-08-03
Moveable packages dir

jdarwood007:
This is a feature request basically so we can deny it if needed for 2.1.  I am proposing we add support in SMF to be able to move the packages directory.  I am accomplishing this by adding a new Settings.php variable, having upgrade.php add that and updated all revelvant parts of SMF to use that directory.

I've done some testing and haven't seen anything broken or bad side effects from it yet.

Signed off by me


Oldiesmann:
I can't really think of why/when you'd want to move the Packages directory to be honest, but if there's a good reason for doing so, I'm fine with adding this.

illori:
for security would be the only reason. why does the packages folder need to be in an area that can be accessed by the internet?

if this is pulled in repair_settings.php needs to be updated to support fixing this path.

jdarwood007:
Exactly,  For the same reasons you can/should move your Sources directory.

matthew-kerle:
This is awesome Jeremy, thanks so much for working on it! I'm kind of out of commission right now due to something that happened in my life, but I'd love to see this get pulled into SMF 2.1 before it goes beta! :smile:

jdarwood007:
@Labradoodle-360 Whenever you can you are welcome to contribute what you got.  Like all of our features added, we will need it well tested.

PR #557, opened 2013-08-03
 Added option to convert to MySQLi during upgrade

jdarwood007:
This should help us make peoples migration to MySQLi easier to do.

Signed-off-by: SleePy

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -1343,10 +1343,13 @@ function UpgradeOptions()
 	if (empty($db_type))
 		$changes['db_type'] = 'mysql';
 
+	// For now we offer a option, this may change in future versions when mysql is completely removed.
+	if (!empty($_POST['convertMysql']) && $db_type == 'mysql')
+		$changes['db_type'] = '\'mysqli\'';
+
 	// Maybe we haven't had this option yet?
 	if (empty($packagesdir))
 		$changes['packagesdir'] = '\'' . fixRelativePath($boarddir) . '/Packages\'';
-
 
 	// @todo Maybe change the cookie name if going to 1.1, too?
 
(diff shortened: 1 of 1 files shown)


Oldiesmann:
Once the link and merge conflicts are fixed, is this good to go, or do we need to test it out more?

Grudge:
Couple of concerns:
1) Shouldn't there be a check included that mysqli is actually supported on the server? Something like function_exists('mysqli_connect')? It's perfectly possible that PHP might have been compiled with MySQL and not MySQLi, breaking the install - leading onto:
2) The proposed implementation is that this is checked by default. I wonder whether this should be unchecked (at least on first release), and only made default on the second release when we can be more confident it doesn't break things?

tinoest:
Wouldn't a check to see if class_exists(mysqli) be more suitable? Although from a client point of view both mysql and mysqli perform very similar. Either using mysqlnd or libmysql and you can pretty much fall back onto mysql if mysqli fails with no discernible impact to the end user.

Arantor:
I'd argue that since the $smcFunc implementation of MySQLi uses the functions, testing the class is actually unreliable. I can actually imagine hosts that might disable the mysqli functions but not the class, for example.

jdarwood007:
You are correct Grudge,  We should check that at least the basic mysqli functionality is there.  I will update my PR here in a few days (when I have time to do so).

Although it is by default included with php 5 and somebody would have to explicitly remove that support or be using php 4 still to have the issue occur.  As well, it is a transparent change in the respect you can change between them at this point without noticing the slightest difference. 

At some point we should say mysqli is our primary supported DBA and start making use of its functions.  http://php.net/mysqli.overview  As pointed out by their official docs, its still maintained and supports more features than mysql functions do.

jdarwood007:
That should do it.  Git can be a pain sometimes when you have merge conflicts..

Grudge:
Sorry to be a pain Sleepy, but I think that the option should not be presented if the function does not exist. If mysqli isn't built onto the server for some reason, and someone checks this box, their install will stop working altogether, and it will generate support requests. I'd therefore say that if MySQLi will not work, then we should not present the option - this is the approach we take on install (etc) in regards to showing details on supported databases.

jdarwood007:
True enough, I was thinking maybe present the option there for some reason, but honestly you are right, if it doesn't exist it would break no mater what.

PR #559, opened 2013-08-06
Add SQLite3 support

Oldiesmann:
This commit adds support for installing/running SMF on an SQLite3 database. It currently does not support upgrading from 2.x to 3.x, but that functionality can be added later if needed.

Note that I haven't tested everything with this, but none of the actual SQL changed - just the PHP side of things. The basic functionality (installation, clicking around, posting, clearing the error log, etc.) works fine, so I don't expect anything else to break.

### DIFF
--- a/Sources/DbExtra-sqlite3.php
+++ b/Sources/DbExtra-sqlite3.php
@@ -1,0 +1,369 @@
+<?php
+
+/**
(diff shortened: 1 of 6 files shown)


Oldiesmann:
Since there aren't any comments on this I'm going to go ahead and pull it in.

PR #563, opened 2013-08-09
Login's fail with debug mode

jdarwood007:
This is straight forwarder and looks like it was debugging code that got left in place.


emanuele45:
/me is innocent... lol

I thought this was removed a long time ago... :blush:

jdarwood007:
Darn it, I need to redo this commit.. I have no idea how those other commits made it into there.  I branched from the master not from my working one...

jdarwood007:
I was able to correct it I think...

PR #571, opened 2013-08-13
Quote Action

### LINKED ISSUE #565: $txt['quote'] in translation
This is something that affects translations into many languages.

$txt[['quote']  - in index.english-utf8.php - This is used to place the word "Quote" both on the button in the top right hand side of a post, and also to place the word "Quote" above a quotation in a post.

This is OK in English, but in many other languages it causes bit of a problem as the verb form "to quote"  is different to the noun form "a quote". In the case of German, they have made do with the noun form in both cases (Zitat), but this looks strange next to the other buttons, which all use the verb form. In the case of the Russian translation, they have used the noun form, but that means it says "to quote" above each quotation, which is far from ideal ! I translate the Esperanto version and it has the same problem...

Would it be possible to create an extra string to help solve this? Thank you!

Here's a picture of the problem:

quote1

XinYenFon:
This will resolves #565

### DIFF
--- a/Themes/default/Display.template.php
+++ b/Themes/default/Display.template.php
@@ -586,12 +586,12 @@ function template_main()
 			// Can they reply? Have they turned on quick reply?
 			if ($context['can_quote'] && !empty($options['display_quick_reply']))
 				echo '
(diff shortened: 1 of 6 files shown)


ChainyEO:
@Antes 

I should have responded earlier, but thank you for fixing this one!

PR #577, opened 2013-08-18
integrate_register_check

MissAllSunday:
Another hook.

Make it easier to any anti-spam mods to check the user, add their own reg_error entries and/or end the execution before the user gets added to the DB.


emanuele45:
`s/$/&/` :stuck_out_tongue:

PR #578, opened 2013-08-18
Declare vars on specific places in poster list

MissAllSunday:
This is just an idea to let mod authors to add their own data to the poster list area on each message.

It is a simple declared var echoed on very specific locations inside the poster list. Made it to be fillable via integrate_prepare_display_context hook.

The custom HTML var is declared on a per-message basics.

I didn't wrapped any var on any HTML tag on purpose, the idea is to let multiple mods to add their own independent HTML code.

Things to look out for newbie mod authors:
- The data needs to be appended, not assigned. Perhaps a nice note on each comment would help.
- Each mod needs to properly check their resulting HTML before append it, there is no special checking other than a mere !empty() To avoid adding more logic to the template file and make it easier for theme authors to relocate the vars if needed. This includes adding proper classes and IDs to any HTML tag too.
- Each mod needs to check for things like $message['member']['is_guest'] and any other "visibility" check.
- Each mod should take care of proper indentation and tabs and append their code with a new line at the beginning. This isn't really mandatory but it would be nice if the resulting HTML still follows the original indentation.

Since this deals with view code, the exact place and even the names can be changed, I just chose some of the most common places mod authors use to add their data.

Suffices to say, this isn't a panacea, it won't cover every specific need out there but will surely help to reduce template edits at least in the most common places.

If going further with the idea, a couple more vars can be declared for before and after the post body itself or other places for things like a ratings mod or something like that.


emanuele45:
Why not expand the custom profile fields with new placements? :angel:

MissAllSunday:
Because not all the data that would be displayed comes from profile fields. Besides, it looks pretty hackish to do stuff like $output['member']['custom_fields']['-1'] to add a new custom field, not to mention you still need to edit Display.php to do that anyway.

Custom profile fields aren't a per-message feature, some mods would like to work with data in every message.

Some vars that are unique to a single post are the post count, the message ID, the time, etc, all of them can be used by several mods on a per-message basis.

Each custom profile field is limited to a li with a single "custom" class, limiting what a mod can do in terms of HTML handling.

Not really a fan of setting a key => values for each field; as flexible as arrays are, having each field stuffed as an array value pretty much limits what you can do with them not to mention it feels incredible square.

Having a pure string may not be as nice as having everything separated in array values but it offers far more flexibility to the mod author.

Less logic in the template, no foreach, no if/else statements, this makes it easier for themers to not mess around with vars and logic that most of the time they do not understand.

Lastly, SMF already does this kind of thing in vars like insert_after_template or html_headers and some admin vars like $context['settings_message']

Handling/displaying HTML via arrays just isn't the way to go.

PR #579, opened 2013-08-21
New package-info.xml commands for Release 2.1

Underdog-01:
This is a proposal for adding the ability to use < input > tags in the package-info.xml file.
It gives mod authors the ability to use HTML5 input and textarea for installing or uninstalling their modifications where the data can be processed in files accessed via the < code > tags.

I authorize the use of this additional code for any releases of the SMF software package.

Signed,
Underdog @ webdevelop.comli.com  
Underdog-01 @ github.com        

Please feel free to email me or leave me a message on my forum with any questions or feedback.
Thank you.

### DIFF
--- a/Sources/Packages.php
+++ b/Sources/Packages.php
@@ -1002,10 +1002,8 @@ function PackageInstall()
 				// This is just here as reference for what is available.
 				global $txt, $boarddir, $sourcedir, $modSettings, $context, $settings, $forum_version, $smcFunc;
 
-				$context['new_inputs'] = array();
-				if (!empty($_POST['new_inputs']))
-					$context['new_inputs'] = $_POST['new_inputs'];	
-
+				$context['new_inputs'] = !empty($_POST['new_inputs']) ? $_POST['new_inputs'] : array();
+				
 				// Now include the file and be done with it ;).
 				if (file_exists($packagesdir . '/temp/' . $context['base_path'] . $action['filename']))
 					require($packagesdir . '/temp/' . $context['base_path'] . $action['filename']);
--- a/Sources/Packages.php
+++ b/Sources/Packages.php
@@ -1002,6 +1002,10 @@ function PackageInstall()
 				// This is just here as reference for what is available.
 				global $txt, $boarddir, $sourcedir, $modSettings, $context, $settings, $forum_version, $smcFunc;
 
+				$context['new_inputs'] = array();
+				if (!empty($_POST['new_inputs']))
+					$context['new_inputs'] = $_POST['new_inputs'];	
+
 				// Now include the file and be done with it ;).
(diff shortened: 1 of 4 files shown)


Underdog-01:
Changes committed to Release2.1 branch... done.
I fixed the typos, changed some if/else to switch/case and changed some comparisons to typesafe.
I'll test it again later today to make sure the switch/case is working properly.
It does the same thing as if/else I suppose but Imo looks neater (aesthetically).

Underdog-01:
Tested... works fine.

Oldiesmann:
We need each individual commit to be signed off, not the pull request itself. When committing things, use the -s flag (eg `git commit -asm 'Commit message here'`)

Underdog-01:
I did it manually to all of them.

Oldiesmann:
Commenting on the commits doesn't really work as those will get lost when we merge this in. It needs to be done the proper way (in the commit message for each commit).

Grudge:
I'm not sure I like the proposed implementation of this, it doesn't align with the general packages-info.xml schema. why would we introduce a custom way of signalling data with {} (etc) rather than using the accepted means of XML, such as:
<input_field type="checkbox" default_value="checked" xxxx />

I'd also suggest that if this were supported then the mod author should not be given the opportunity to set style or classes - this should be implemented in a standard way (so the user sees any custom inputs much like they do on a settings screen).

Given I've not seen much demand for this previously, I think we're in danger of over implementing a solution. Finally, at the point the fields are shown the user has not decided to go ahead with the install. As such all data should be htmlspecialchar'd, as I believe the current implementation could enable remote code execution by a malicious mod author.

Effectively I'd propose a standardised, simpler implementation. If demand by authors requires mode flexibility then this could be looked at as a later phase.

ghost:
make sense that does :)

Arantor:
There is certainly a place for extending the package system to provide simple mod settings, and leave more complex settings (like multiple pages) to conventional hooking.

I actually did something like that in Wedge, and I found it very useful, but I'm not sure the current suggestion is really the way to do it :/

Arantor:
Having re-reviewed this, it's not what I originally thought (settings page declared from package-info, which does have proven use to simplify mods adding settings) but configuration for the installer itself... I cannot think of a time when this would actually be needed over and above the ability to flag things for deletion on uninstall and SMF can do that itself if the package is set up properly in the first place.

As it stands, I think I'm going to have to say thank you -Underdog- for your submission, but that I can't see it as something we can use at this time, even putting aside some of the technical concerns raised above.

That said, if you can provide a use-case where such settings would be useful in a mod, please do comment and we can reopen this and work together on getting it in to the core; if there is a truly viable use case for it, let's provide for it, but I'm struggling to think of one.

PR #580, opened 2013-08-22
Disregard -> Unwatch

Oldiesmann:


jdarwood007:
Shouldn't upgrade.php be updated as well to change that column if needed?

Oldiesmann:
Fixed in #581

PR #586, opened 2013-08-24
Fixes 580 pr

jdarwood007:
Submitting this as a simple fix to the pull request 580 which changed disregarded to unwatched.  SMF's upgrade script gracefully handles database errors.

I don't have the capacity to test the PostgreSQL or sqlite changes, but the code is based on other similar parts of our upgrade scripts and thus should work.  Please let me know if they need adjusted though.

### DIFF
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -203,6 +203,11 @@ VALUES
 	('enable_unwatch', 0);
 ---#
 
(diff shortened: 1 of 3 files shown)


Oldiesmann:
I thought we weren't worrying about people already using SMF development versions?

jdarwood007:
Yea, but it isn't a hard fix.  But meh, reject if needed.

PR #587, opened 2013-08-24
Fixes 582 template issue

### LINKED ISSUE #582: token verification failed ignore boards
enable ignore boards in admin panel

try to ignore a board in your profile, when you save the page it will give a token verification failed

jdarwood007:
Fixes #582

Oddly enough, this is the only occurrence of the template_profile_save.  I wonder if we shouldn't adjust all of our profile template to use that.

Signed-off-by: jdarwood007

### DIFF
--- a/Themes/default/Profile.template.php
+++ b/Themes/default/Profile.template.php
@@ -2539,7 +2539,13 @@ function template_profile_save()
 					</dl>';
 
 	echo '
-					<div class="righttext">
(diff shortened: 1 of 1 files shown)


emanuele45:
As I said elsewhere: the template was (in 1.0 and 1.1) used in more profile pages (4 to be precise), while in 2.0 has been removed and kept only in one place.

PR #590, opened 2013-08-26
Fall back to the default theme when loading jquery.sceditor.css (fixes #589)...

### LINKED ISSUE #589: jquery.sceditor.css called for each theme
currently jquery.sceditor.css is called from each theme, it should only be called from the default theme.

Oldiesmann:


emanuele45:
There was no need for that: fallback on default is the default.

PR #599, opened 2013-09-02
Sceditor 1.4.4

emanuele45:


Oldiesmann:
Continuing the discussion from #526 and the dev blog post you referenced:

Keep the functionality to remember tabindex and have multiple lines of bbcode buttons since this is what people are used to.

I don't really know about the url/iurl thing. I'm not sure how many people even know that iurl exists, let alone how many of those actually use it.

Multiple lines of smileys is something some people like, and as we've supported this since SMF 1.0, it'd be nice to keep it if we can.

I realize it will make upgrading a bit more difficult in the future, but I think it's best to keep useful functionality.

emanuele45:
> Keep the functionality to remember tabindex and have multiple lines of bbcode buttons since this is what people are used to.

The tabindex I agree.
The multiple lines it can be achieved with css alone, so I'd prefer that way because it is the single biggest change to SCE code.

> I don't really know about the url/iurl thing. I'm not sure how many people even know that iurl exists, let alone how many of those actually use it.

Since that affects me personally (yes, I'm very selfish :stuck_out_tongue: ) I kept that change.
Anyway, the problem is that all url become iurl, and most likely, based on history, this is **not** what you want...

> Multiple lines of smileys is something some people like, and as we've supported this since SMF 1.0, it'd be nice to keep it if we can.

In fact that depends entirely on my own "custom" code, so it's not a problem.

Anyway, just for the record: all the custom code is marked with an `@SMF code:` and a comment why it was added.
There is one that is tricky:
https://github.com/emanuele45/playpen/blob/8bc55372e91c046e5fc04008a679fb63f7047167/Themes/default/scripts/jquery.sceditor.bbcode.js#L1450
I feel it's necessary but  don't remember why.
I'm also too lazy to test it.
Though it may even be no more necessary due to:
https://github.com/emanuele45/playpen/blob/8bc55372e91c046e5fc04008a679fb63f7047167/Themes/default/scripts/jquery.sceditor.bbcode.js#L1436

Dunno.

Oldiesmann:
I misread your comments on iurl/url - I thought you said it makes all iurl tags become url tags, which isn't a big deal. If it's the other way around, we definitely need to keep that fix.

Not sure why the change on line 1450 would be needed. Without testing things, the code on line 1436 might be there to prevent problems if there's no defined style attribute for that element.

XinYenFon:
This seems odd, Content got from : http://support.amd.com/us/kbarticles/Pages/AMDCatalyst13-10WINBetaDriver.aspx

modify message 2013-09-15 15-25-50

emanuele45:
Now should be fixed.

XinYenFon:
I'm not sure this is related to SCEditor or this patch but
modify message 2013-09-15 16-14-16

emanuele45:
Technically is fault of the broken markup of the page:

```
<ul>
<li>AMD Catalyst 13.10 Beta includes all of the features/fixes found in AMD Catalyst 13.8 Beta2</li>
<li>PCI-E bus speed is no longer set to x1 on the secondary GPU when running in CrossFire configurations</li>
<li>Rome Total War 2:  Updated AMD CrossFire™ profile</li>
<ul>
<li>improves CrossFire scaling up to 20% (at 2560x1600 with extreme settings) </li>
...
</ul>
```

Do you see the `<ul>` after `Updated AMD CrossFire™ profile`? It should be opened from inside the `<li>` and not after the `<li>` is closed:

```
<ul>
<li>AMD Catalyst 13.10 Beta includes all of the features/fixes found in AMD Catalyst 13.8 Beta2</li>
<li>PCI-E bus speed is no longer set to x1 on the secondary GPU when running in CrossFire configurations</li>
<li>Rome Total War 2:  Updated AMD CrossFire™ profile
<ul>
<li>improves CrossFire scaling up to 20% (at 2560x1600 with extreme settings) </li>
...
</ul>
</li>
```

Also the outcome of the paste is highly browser-dependant: FF works (except for what you reported), Opera converts it to plain text, Chrome shows only the "headings" and nothing inside the `<ul>`s...

XinYenFon:
There is a problem with [img] tag
modify message 2013-09-15 18-21-52

emanuele45:
mmm...is a problem with img or with html?

XinYenFon:
[html][/html] working on post (posted) and preview.

[img] tag only only visually working when i disable source mode.

emanuele45:
Works fine here, I used this code:

```
[html]<img src="https://fbstatic-a.akamaihd.net/rsrc.php/v2/y3/r/0PnClccPFXs.png" />[/html]

troll

[img]https://fbstatic-a.akamaihd.net/rsrc.php/v2/y3/r/0PnClccPFXs.png[/img]
```

And I can see both images in the preview, and the one in the [img] tag in the WYSIWYG.

live627:
May I request a status update on this?

emanuele45:
I can't see if it merges properly or not.
The bug reported by @Antes was probably the one about preg_replace.

Oldiesmann:
@emanuele45 It still says it can be automatically merged, so no issues there.

Arantor:
I have a crazy idea. I'm going to merge this and then see what's still broken after.

PR #605, opened 2013-09-08
 & Changing "Edit message icon" string to "Message Icons" which looks be...

Grudge:
& Changing "Edit message icon" string to "Message Icons" which looks better in the current admin template.


live627:
OMG IT'S GRUDGE!!

mcblaber:
He's alive! o.o

illori:
i really dont agree with this change, now we have Message Icons and Edit Smileys which seems a confusing, we can view message icons and no place to edit them as the menu shows, yet we can add/edit smileys easily given the titles of each section of the page. we should be consistent, so Edit message icons seems to be the best option.

PR #611, opened 2013-09-14
Update Credits for 2.1

Acans:
- Adds Corporate Team Section
- Adds SMF friends and GitHub Contributors to Special Thanks
- Updates Team Members to those who are currently/were on the team after the release of 2.0.
- Changes 2.0 to 2.1 :P


XinYenFon:
I maybe missing some details but generally looks awesome :)

Acans:
I went through the old credits and check them with resignation topics as well as any new team members and I (think) I didn't miss anybody.

Anything you can think of missing Antes?

eurich:
hey, yet again you're removing old DEVs from the credits?

XinYenFon:
Current credits page;
credits

emanuele45:
@Antes what TE means is that the list includes only people that worked on 2.1, while the "line" is 2.x, so it should include people that worked on both versions of the software.
That's how it is in the current SMF credits page.

Spuds:
> TE means is that the list includes only people that worked on 2.1

Sadly it does not even do that properly ...

eurich:
What I basically mean: SMF was made by lots of people (unknown, compuart, grudge and many others..) and those people should get credits... not the team nor the NPO made this software ;-)

ChainyEO:
I'm a mere translator of this software, but I think it's a big shame that there's a constant issue about the credits. It's spoiling the otherwise very friendly and happy atmosphere here. Why not just make it simple and include ALL developers that have worked on the project and then everyone will be happy. 

The current developers are obviously extemely important. Without them, SMF would quickly die off. But the former developers must always be credited, too. The current developers are presumably using a huge amount of code created by those former developers.

I've just looked at the credits for Firefox and Inkscape - both are very long lists. Are they only the active developers? Do they delete former developers from the list?

I can imagine that the credits would be very complicated if things were divided up into who contributed precisely to which version of SMF and how much was contributed by each developer etc. Now that SMF is here on Github, it is possible to find out more precise details about the contributions of the developers.

ChainyEO:
The Credits don't have to fit neatly on a small and concise page. There's nothing wrong in putting a big long list of developers on it. There's lots of room.

XinYenFon:
I have no disagreement about putting all friends to the list.

eurich:
> The current developers are obviously extemely important. Without them, SMF would quickly die off. But the former developers must always be credited, too. The current developers are presumably using a huge amount of code created by those former developers.

Exactly this is why I am complaining.. I've spend thousands of hours for SMF (rewriting a vast amount of templates to semantic html, fixing lots of bugs, developing more than 20 converters ...)

All I want is "Credits" for my work ;) 

But yet again the SMF team (and NPO) seems to be more important than the people who made the software. 

Sad days for the contributors ..

ChainyEO:
Just out of interest, why was Thorsten Eurich (and many others) deleted from the list of developers? Why not just leave all the names there, all the developers that have at some stage contributed? Why is a big long list a problem?

XinYenFon:
and you are right @eurich , I agree we should add all former/current developers under Development team list, and all other friends under friends.

Spuds:
Just to note, don't solely go the developer badge ... thats a misnomer as to who contributed

There are current and former members that never had a developer badge but supplied large amounts of code to the core.  I've always felt that was improper in terms of acknowledgement of who the true "developers" were. 

If you want to make a distinction, then perhaps, it should be if you provided code (and sure set a reasonable threshold if desired) and were a badged member then why not acknowledge them (all of them).  Trying to separate Current/Former by time/place/badge distinctions is prone to inaccuracies and needlessly consumes time IMO

XinYenFon:
Well @Spuds if we can create that list I don't see why not we list Code Contributors (1.x/2.x not separated list).

I already asked team if we can create a list for all friends with their former (primary) position. So in credits we can do small boxes with Nick (name) - Position or as like current system

Former Developers
Former Customizers...

Spuds:
I don't have an problems with that really, I guess some names can get listed twice, you could even break out a section of former/current lead developers as a special thanks for the special hell that can be ... the data can be sliced and grouped as desired.

But in the end lets not loose perspective of whats most important here  ....  that people are not left out for what often amounts to 100's or 1000's of hours of work because they had the wrong badge or wrong time-frame or whatever.  If they contributed code (of significance) they should be acknowledged in the developer role call period.

XinYenFon:
Better or worse :cat: didn't want to list everything its just an try don't get me wrong.

credits 2013-09-14 18-40-23

eurich:
tbh it's still  wrong.. why are code contributors sorted under the section "Former team members" ? not everyone who contributed code is a former team member..

Besides this, the order of the sections is IMHO derogatory for the contributors.

The SMF team still needs to learn one important thing: The **software** (and it's contributors) should be the focus, not the team nor the NPO or whatever else..

XinYenFon:
@eurich who give the impression to you i don't know but leave that NPO prio behind. I'm trying to make things acceptable for everyone, as i want to clear something code contributors group is for people @Spuds pointed people who send their fixes/codes but not developer.

Please feel free to shoot me mail/pm or mention :) I'll try to do my best to help solving this unfortunate problem and make everyone happy because people on that list deserves that.

credits 2013-09-15 09-44-45

ChainyEO:
In this discussion about the credits I've read over and over again that some developers don't feel that they are being shown enough respect in the credits page. I'm just wondering how other projects solve this kind of problem - perhaps we should follow their example? 

I can understand that it might seem a bit demeaning to place former developers much further down the page into a section called 'Former developers'. They are not even listed on the page, but linked to, which makes it even worse. Why not just include the names on the actual page?

Regarding the order of things on the Credits page, I suppose some might think that whatever order is chosen, someone will be offended.  But much of this discussion seems to be Developers vs. SMF Team, which can only be a bad thing...

I'm not sure that I see the point in differentiating between 'developer' and 'code contributor', but perhaps the developers here could explain? Anyway, I have a small question. If we made an alphabetical list of all the developers and code contributors (including all SMF versions), just how many people would be in that list? I'm asking as I'm trying to figure out why we can't just place all the names in the Developer part of the the section "SMF Team". Perhaps "Former Developers" could appear just under "Developers", same for "Former Customizers" etc?

Oldiesmann:
One thing I wanted to clear up here... Nobody is trying to exclude anyone from the list of credits. In the past we have generally only listed current/former team members who were involved with the current version (eg if you were a developer for 1.0 but didn't contribute anything towards 1.1, you weren't listed as a developer there). The initial update of the credits committed here is simply following that same trend. We can easily change things as desired by everyone.

matthew-kerle:
How was the order of the list defined per group...? For instance, I am second to last on the Developers list, even though I was the last one to go to the friends group...

Acans:
Team lead, current than former in alphabetical order.

The new version on my local repo (waiting for my adsl to get connect to my new address) tho combines current and former in alphabetical order.

MissAllSunday:
Why do you make things so overly complicated? why is there a need to segregate people according to trivial considerations such as the timeframe when they were team members?
- Set devs, current and former at the top, this includes all devs of 2.0 since 2.1 still use large portions of it. And since you're at it, include 1.1 and 1.0, as a way to show respect. Sort them alphabetically and be done with it.
- Team members, ALL active members during 2.1 dev cycle, there shouldn't be any "former" and "current" list since former and current is relative to time not contributions. No sub-team or lead differentiation (since, again, its relative, members go from one to another sub-team and you had and will have many leads until release), just a plain list of ALL members, sort them alphabetically and be done with it. 
- Code contributors. This is the most tricky one since a code contributor can be someone who only adds a comment line or an empty line or some other trival stuff gotta be careful.
- Third party copyrights
- NPO stuff.

Acans:
1, 4 & 5. Aye

2: Aye, however if we are including developers from 1.0/1.1, shouldn't we include former team members as well as a sign of respect? Respect to the support, customizers, doc writers, many of whom shaped SMF in many ways. Seeing as there is no credits page on the Wiki for former Doc Writers or Credits page on the Cus site for Former Customizers, both who contributed their time to the project. Not to mention Support and Localizers for their time?

3: That's why I hyperlinked to https://github.com/SimpleMachines/SMF2.1/graphs/contributors/ instead of trying to define clear rules on what warrants a mention/entry for Code contributors.

Oldiesmann:
Motoko-chan has said that he doesn't want to be listed in the credits. Other than that, it looks good.

Arantor:
I have no idea why the credits actually need to be in the software at all. Just have a page on sm.org documenting who did what for SMF as a whole. Since even the people who shaped 1.0 still have a strong presence on 2.1 at the code level, just have the page as a final summation of 'who's shaped SMF to make it what it is today'. There's no need to pigeonhole people by position; by that definition I've been part of several teams in the past (despite only having two different badges), just list everyone who helped on sm.org (and the current team if you so wish) and move it out of the software entirely.

illori:
if that would include a link to a page on sm.org from the "credits" page coreisp has said that is a bad idea due to the bandwidth that would be used for the page to be viewed each time the spiders etc view the page.

also i am pretty sure spuds does not want to be listed in the credits.

Arantor:
Well, there's no need to have three links in the footer. The sm.org argument is a bit thin since two of them already point to sm.org (one to the front page, one to the licence page)

There are people who don't want to be listed in the credits as contributors, just as there are people who don't want their real name listed for whatever reason (as per the source)... you don't have to put those people there if they choose not to be, you just have a list of everyone else.

I'm still not seeing the problem with this, though.

illori:
complain to coreisp who is the one that said he did not want a link to sm.org from the credits page which could add 150 million more hits per month(?) to sm.org

Arantor:
So just have two links in the footer, not three. Still not seeing the problem with this.

Acans:
I've already spoken to Spuds via PM, as explained in the commit description.

Issue #517, opened 2013-07-18
name of default theme in 2.1


illori:
currently the default theme of 2.0 and 2.1 has the same name... curve in the admin panel

this will cause many headaches for the support people, as they will not know if someone has tried to use the old 2.0 default theme in 2.1 or if they are using the 2.1 default theme.

the default theme name needs to be changed [it appears in the drop down and elsewhere in the themes and layout area]

it requires a db update to whatever the name should be, IMO something different then curve to stop confusion.

Oldiesmann:
I thought it was called "Curve 2" at some point, but I agree that we need to change it to avoid confusion. Any suggestions on the new name?

illori:
i am not great at coming up with names... if we want to stick with something "curve" like why not something like ... "curvature"

matthew-kerle:
The theme name is "Curve II", and I personally like it. That's honestly what it is...an updated version of SMF Curve, nothing more. I think since we're about to beta, we should stick with the naming structure and not add another thing to change.

illori:
and that will be very confusing to people posting and telling what theme they are using, they will leave off the II and support staff will think they are trying to use the 2.0 default theme on 2.1.

matthew-kerle:
However, you know that if they're using the default theme on SMF 2.1, it's indeed Curve II rather than "SMF Curve" or "Curve I". And likewise, if they say they're using SMF 2.0, you know they're using SMF Curve and not SMF Curve II.

I do see the potential for confusion, but in almost every case you'll ask for version number, or the actual file, and that'll eliminate any confusion.

illori:
either way the theme has to be renamed, currently it is named curve in the database.

ghost:
im with Lab on this, I dont think name is as important as you think, no matter what name is chosen people wont like it so I think Curve2 or Curve II is best for it, as for the name change in db to change it to Curve2 or CurveII thats easy.

illori:
the name "Curve2" instead of "Curve II" would be fine with me, as people will be less likely to leave out the 2 as it is more part of the name, without the space.

ghost:
if oldies agrees we can just call it Curve2 ... getting marketing ideas n ow

Oldiesmann:
I don't care what we call it. Whatever makes it easier for you guys

illori:
if we are to stick to curve, then curve2 is fine, but i think curvature would work as well. sounds like more people here think curve2 would be best.

PR #613, opened 2013-09-17
Rename default theme and uninstall old ones when upgrading

Oldiesmann:


emanuele45:
1) Why did you change the 1_1 upgrade file?
2) Why did you change the 2_0 upgrade file?
3) `$classic_dir = $GLOBALS['boarddir'] . '/Themes/classic';` won't work if the Themes directory is not in $boarddir (and don't say people are not doing that, because [we had to fix a very similar issue last year](http://www.simplemachines.org/community/index.php?topic=489333.msg3441031#msg3441031)).
4) As far as I understand, the upgrade 1_1 and 2_0 were **installing** a theme, while now they are removing it...it's a bit different... 1_1 did install Babylon, now is removing classic, 2_0 did install Core, while now is removing Babylon.
Consequence of 1, 2 and 4: if one of these two files would end up now into the SVN repo from which we generate the install and upgrade packages people would have a broken upgrade from 1.0 to 1.1. or from 1.x to 2.0 because the themes that are supposed to be installed during these processes are not going to be installed.
5) For the nth time: would you kindly give others the time to have a look... oh darn, waste of time.

Oldiesmann:
1 and 2. Because it makes sense to keep all things that would apply when upgrading from 1.1 or 2.0 in the appropriate files.
3. Existing code in the 1.1 and 2.0 upgrade SQL files used similar methods for detecting whether a particular theme was installed, so I didn't even think about the fact that some people might not have Themes in $boarddir.
4. Why would either of those files end up in SVN?
5. I didn't realize there was a problem with not doing so. I've made dozens of commits over the past few weeks and this is the first time anyone has said anything about me not leaving them for comments/review before merging them in. #517 received all of two comments in two months, so I assumed it wasn't something many people cared about.

Oldiesmann:
Further comment on point 4. I honestly don't see how that could lead to broken upgrades anyway. My code specifically sets anyone using those themes to "Forum/Board Default".

emanuele45:
1 and 2 do change completely the behaviour of the upgrader: before your changes it was installing a new theme if it didn't exist, now it is removing one, so if you use that file to upgrade from 1.0 to 1.1 you end up without any babylon theme (that was a perfectly valid theme at the time), instead if you use it to upgrade from 1.1 to 2.0 you will not have Core installed (and Core was a perfectly valid theme for 2.0).

Also, let's number it 2a:
https://github.com/Oldiesmann/SMF21/blob/fa00c29e349721e84fcf5dff568578b19a2fb50b/other/upgrade_2-0_mysql.sql#L333
changes `ID_THEME` to `id_theme`, so:
https://github.com/SimpleMachines/SMF2.1/pull/613/files#diff-d7e190b444405cb0da14496d36d053ccL2772
is technically wrong, `id_theme` was correct (yes, I know it's still valid in MySQL).

3) That's that point: it was broken from the start.

4) Why would you not let previous versions benefit from your fixes? And an upgrade that doesn't do what is supposed to do (i.e. install new perfectly valid themes) isn't it broken?

5) As long as I remember we (Spuds, Norv and me) implicitly agreed to wait a bit for someone else review before merging our own branches unless it was an obvious and "critical" bug, if the "merge-soon" way is the current way, go ahead with it, it's just a matter of define how people work together.

Arantor:
1 & 2... this sounds like a lack of regression testing and the need to retest it on all the major permutations (1.0 to 1.1, 1.1 to 2.0, 2.0 to 2.1, 1.1 to 2.1)

5) I wondered about that, but I'm of the view of taking a bit of common sense with it; the stuff I've done thus far has been minor bug fixes and while I did start out with branching and merging, that seemed an awful waste of time for minor fixes. The new thing I'm doing - multi language, multi answer Q&A - probably should have been in a new branch too, but whatever :P I will be letting people review that before including but honestly I'm not actually sure it's particularly worth it since it will have been tested before inclusion.

Oldiesmann:
The changing of id_theme to ID_THEME was unintentional - I certainly wasn't trying to break things there, though as you pointed out it'll still work in MySQL.

I can fix the broken theme detection. That's not a difficult to do.

For the other thing, we can move that to the 2.1 upgrade file, but then we run into problems if you run the upgrade after 2.1 is installed and have themes with those names, you unintentionally uninstall themes that could have been perfectly valid. (As we all know, some users think they have to use the full upgrade file or at least run upgrade.php when upgrading between minor versions). I'm not sure what the likelihood of someone having themes with those names is, but it's still a possibility.

PR #614, opened 2013-09-17
Don't reveal users' email addresses when notifying moderators of reporte...

### LINKED ISSUE #612: user email exposed when reporting post
why when a user reports a post to moderator does it have to be sent by the users email address?

it seems that it would be easier for the users if the default forum email address was used and just had the users name in the email.

there could be times that the users do not read that their email address will be exposed and send the report anyway and it comes back to hurt them in the future [for example being able to use a whois and find out information about the user if they have their own domain].

i really dont see why the email is the users, the email could also be caught by spam filters as not being sent from the same server as the email is from.

Oldiesmann:
...d posts

Fixes #612

### DIFF
--- a/Sources/SendTopic.php
+++ b/Sources/SendTopic.php
@@ -529,11 +529,11 @@ function ReportToModerator2()
 		$smcFunc['db_insert']('',
 			'{db_prefix}log_reported_comments',
 			array(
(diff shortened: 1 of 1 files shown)


emanuele45:
Who are you answering (or asking for details) to if the person reporting the post is a guest?

PR #627, opened 2013-09-25
Fix a minor kind of SQL injection

emanuele45:
There may be others.

### DIFF
--- a/Sources/PersonalMessage.php
+++ b/Sources/PersonalMessage.php
@@ -1155,7 +1155,7 @@ function MessageSearch2()
 
 		for ($k = 0, $n = count($possible_users); $k < $n; $k++)
 		{
-			$possible_users[$k] = trim($possible_users[$k]);
+			$possible_users[$k] = addcslashes(trim($possible_users[$k]), '\\\'');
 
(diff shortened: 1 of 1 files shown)


Oldiesmann:
The problem is the user string isn't being run through the DB escape functions - SMF just escapes a few things that might cause problems in "LIKE":

`$userString = strtr($userString, array('%' => '\%', '_' => '\_', '*' => '%', '?' => '_'));`

After that, SMF removes quotes from around the names with regex, and whatever is left from that is passed to the query, unescaped.

We can't use addcslashes to fix it because PostgreSQL and SQLite don't use \ for escaping quotes. We need to pass the user list through the appropriate db escape function.

emanuele45:
Actually the issue is not the quote but the . ;)
Anyway, yes probably use quote is better, but then do it in the autosuggest too. ;)

Oldiesmann:
Sorry if my previous reply was confusing. The slash is indeed the problem, but the solution is that we need to properly escape usernames by running them through $smcFunc['db_escape_string'] before passing them off to the query - not calling addcslashes().

PR #631, opened 2013-09-27
Attachment extension

MissAllSunday:
Add a generic binary extension to files saved in attachments dir

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -3345,7 +3345,7 @@ function template_css()
  * Get an attachment's encrypted filename. If $new is true, won't check for file existence.
  * @todo this currently returns the hash if new, and the full filename otherwise.
  * Something messy like that.
- * @todo and of course everything relies on this behavior and work around it. :P.
+ * @todo and of course everything relies on this behaviour and work around it. :P.
  * Converters included.
  *
  * @param $filename
@@ -3395,7 +3395,8 @@ function getAttachmentFilename($filename, $attachment_id, $dir = null, $new = fa
 	else
 		$path = $modSettings['attachmentUploadDir'];
 
-	return $path . '/' . $attachment_id . '_' . $file_hash;
+	// Add a generic extension, kudos to FileZilla!
+	return $path . '/' . $attachment_id . '_' . $file_hash . '.iso';
 }
 
 /**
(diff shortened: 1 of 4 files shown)


Arantor:
Do we need to do anything with the upgrader if we merge this in?

Sources/Subs.php
```diff
@@ -3395,7 +3401,21 @@ function getAttachmentFilename($filename, $attachment_id, $dir = null, $new = fa
...
-	return $path . '/' . $attachment_id . '_' . $file_hash;
+	// Add a generic extension to the full path
+	$return_file = $path . '/' . $attachment_id . '_' . $file_hash . $FileZilla_ext;
+
+	// Is this an old attachment?
+	if (!file_exists($return_file))
+	{
+		// Lets check if there is one without the extension
+		if(file_exists($path . '/' . $attachment_id . '_' . $file_hash))
+		{
```
live627, line 3412: I think these two if() statements can be combined.

jdarwood007:
What is the purpose of this?  We drop extensions for a reason, so why are we adding one back, even a generic version?

Arantor:
We drop the original extension to prevent the file being handled by the system in case the attachments folder is 1) web accessible and 2) normal .htaccess isn't going to work for whatever reason (file missing, not using Apache, whatever)... this means files that are .php can't be executed as PHP which would be a bad security issue.

On the other hand, we have _extremely regular_ issues with support because of FileZilla. FileZilla, by default, treats any file without an extension as a text file. Regardless of content. That means any time a default installation of FileZilla touches the SMF attachments folder, the backup will be irretrievably damaged unless _everything_ is a text file (and invariably isn't)

The best way to solve both problems is to have a generic extension for all files, that won't be executed as PHP or similar and will at the same time prevent FileZilla doing what FileZilla does (which has been filed as a bug with FileZilla for _years_ but they assert that it is not a bug)

This is not even remotely a new problem, I draw your attention to http://www.simplemachines.org/community/index.php?topic=374178.0

NanoSector:
So what's wrong with this commit now?

MissAllSunday:
Still needs some testing, the code seems simple too, perhaps another check or another approach would fit best.

We also need to decide what to do on upgrading.

Sources/Subs.php
MissAllSunday, line 3412: Yep! :) let me see if I still have the branch on local.

See "help squad"? I didn't overreact! :P

Arantor:
The upgrader should fix all files. It should also back fill the mime type and file hash columns.

Issue #476, opened 2013-06-27
profile, show drafts page header missing


illori:
using left side menu, there is no header showing for the show drafts page.

when you use the top menu it has a "show drafts - username" in the header

Issue #381, opened 2013-05-30
integrate_clean_cache


interlab:
http://www.simplemachines.org/community/index.php?topic=454883.msg3522050#msg3522050

for example if you use library which uses own folders of a cache which you too would like to clear in case of forum cache flush

Subs.php

Find:

``` php
    // Invalidate cache, to be sure!
    // ... as long as Load.php can be modified, anyway.
    @touch($sourcedir . '/' . 'Load.php');
    clearstatcache();
```

Replace:

``` php
    // Invalidate cache, to be sure!
    // ... as long as Load.php can be modified, anyway.
    @touch($sourcedir . '/' . 'Load.php');

    call_integration_hook('integrate_clean_cache');

    clearstatcache();
```

PortaMx:
This is also interesting if you have memory cache like ACP ..

interlab:
(sorry my english)

if you use libraries which generate temporal files, by means of this lever you will be able to clear them:
Twig, Doctrine ORM, CSSMin, JSMin, JSMinPlus, ...

PR #355, opened 2013-05-07
Issue: #338 - $user_info['id'] === $int;

matthew-kerle:
- $user_info['id'] and $context['user']['id'] are now cast to integers.

Thanks to @The-Craw for pointing this out.

Signed-Off: Matthew "Labradoodle-360" Kerle |
lab360@simplemachines.org

Issue #338, opened 2013-04-23
$user_info['id'] === $int;


ghost:
I found out today after about 3 hours of debugging and randomly placing var_dump() functions that $user_info['id'] and $context['user']['id'] are both strings. This is evil, and I'm mad about it. :P 

``` PHP
var_dump(
    array($user_info['id'], $context['user']['id'])
);

/* Output:
 * array(2) {
 *     [0]=> string(1) "1"
 *     [1]=> string(1) "1"
 * }
 */
```

matthew-kerle:
Well you do know that you're always able to submit a Pull Request and fix it yourself :)

joshuaadickerson:
`$id_member = $check && ($user_settings['is_activated'] == 1 || $user_settings['is_activated'] == 11) ? $user_settings['id_member'] : 0;`

to

`$id_member = $check && ($user_settings['is_activated'] == 1 || $user_settings['is_activated'] == 11) ? (int) $user_settings['id_member'] : 0;`

Though, replacing that seems like a waste. It should be a simple if instead.

ghost:
Yeah, that got it, @joshuaadickerson. Thanks. :)

ghost:
![](http://i2.kym-cdn.com/entries/icons/original/000/005/365/awwyea.jpg)

PR #632, opened 2013-09-27
! Cast the user id to an int when we first get it set up because that's ...

Arantor:
...the right thing to do. Essentially as per #338 and #355 but without any edit conflicts.

! Add a new hook for cache cleaning as per #381.
! Brought the profile/show drafts to look the same as show posts, in terms of providing a description and the extra header for sidebar mode.


Arantor:
This also fixes #476.

PR #639, opened 2013-09-27
Attachment help strings

### LINKED ISSUE #618: Split up $helptxt['attachment_manager_settings']
In the "Attachment Settings" section of the admin center, we currently have one big helptxt popup that covers most of the settings (but not all of them), and some of the other settings have their own helptxt popups as well. I think it would be best to split this giant language string up into individual ones for each setting. That way there's less reading involved and users don't have to scroll through a bunch of text to find info on the one setting they want more info on.

MissAllSunday:
resolves #618 Split up $helptxt['attachment_manager_settings']

I created a few more help strings for settings that didn't have one. Since apparently I'm not that good with English I didn't actually write any long description, I just copy/paste what was already there.


XinYenFon:
I asked Irisado to review stuff for you since he doesn't have GH account

From the first of the two links in that git post:

+$helptxt['attachment_manager_settings'] = 'Attachments are files that members can upload, and attach to a post.<br /><br />'

Remove the comma before 'and'.

+$helptxt['attachmentNumPerPostLimit'] = ' Select the number of attachments a person can make, per post.'; 

The comma before 'per' is strange.  I would remove it, unless it is there because there are other options in another string which replace it, such as per thread or something similar.  In which case, a comma is still strange, a colon would be more usual in written English.

+$helptxt['attachmentShowImages'] = 'If the uploaded file is a picture, this will show it underneath the post.

I'm not sure about the 'it' in this sentence.  If the context is that by clicking a button, the picture will display, then 'it' is okay, but if the context is that attaching an image will result in the image being displayed, then I would use the following wording: 'If the uploaded file is a picture, it will be displayed underneath the post'.

All the other green strings look okay to me.

In the second link:

+$helptxt['attachmentDirFileLimit'] = 'Max umber of files per directory help text'

That should be 'number' ;).

+$helptxt['automanage_attachments'] = 'Decide how the attachments will be handle

Needs a 'd' added to 'handle', 'handled'

+$helptxt['max_image_height'] = 'The maximum height an attached image can have.'

Normally, we would say in English 'the maximum height an attached image can be'.

I hope that helps.

MissAllSunday:
It would be far better to actually provide more content to those help strings.

PR #643, opened 2013-09-28
Don't require a RTL version of a theme variant

live627:
Most theme variants might not supply their own RTL layouts, so let''s load it only if present.

### DIFF
--- a/Themes/default/index.template.php
+++ b/Themes/default/index.template.php
@@ -114,7 +114,7 @@ function template_html_above()
 		echo '
 		<link rel="stylesheet" type="text/css" href="', $settings['theme_url'], '/css/rtl.css?alp21" />';
 
-	if (!empty($context['theme_variant']))
+	if (!empty($context['theme_variant']) && file_exists($settings['theme_url'], '/css/rtl', $context['theme_variant'], '.css'))
 		echo '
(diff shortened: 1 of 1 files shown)


emanuele45:
/me doesn't like file_exists in a theme file

Also, considering rtl.css is copied when a new theme is created copying and existing one, the likelihood the file is shipped with a theme (even if it doesn't work) is very high.

Also: `$settings['theme_url']` maybe it's better `$settings['theme_dir']`?

And finally: why not instead use loadCSSFile?

live627:
> Also: $settings['theme_url'] maybe it's better $settings['theme_dir']?0

Yep, forgot to change it when I C&P'd it

> And finally: why not instead use loadCSSFile?

Mmm, maybe all CSS should got through that function and be within the bowels of loadTheme()?

Arantor:
The entire variants thing is kind of iffy anyway :P

Arantor:
Most themers don't even use theme variants anyway. Those that do invariably need to provide RTL variants anyway because their main CSS is so infused with the unique-to-theme bits that a simple RTL variation may not work anyway, and if DzinerStudio's anything to go by, they end up shipping RTL variants for each main variant anyway.

I think this one is ultimately not going anywhere for 2.1 - but for 3.0 we need to rethink how we do a lot of this stuff anyway.

Issue #635, opened 2013-09-27
moderation center with missing titles with sidebar menu


illori:
i believe due to https://github.com/SimpleMachines/SMF2.1/commit/651820084e4b8aaaf4064c5a6d0f7d0185b9349f when using the sidebar menu in the moderation center many areas are missing headers

Arantor:
It's passed on! This sidebar is no more! It has ceased to be! It's expired and gone to meet its maker! It's a stiff! Bereft of life, it rests in peace! 

It's kicked the bucket, it's shuffled off this mortal coil, run down the curtain and joined the bleedin' choir invisibile!! THIS IS AN EX-SIDEBAR!

(sorry, needed an easter egg somewhere)

PR #657, opened 2013-10-03
Moderation center

### LINKED ISSUE #505: moderation center latest news
why do moderators need to know the latest news from smf? i think it would be more useful to add a block for unapproved posts/attachments instead, or even membership requests for those that can approve the requests

MissAllSunday:
Fixes #505 couldn't reproduce #635
- Puts the moderators notes outside the rest of blocks, uses all width available.
- Removes moderation settings for latest news and notes.
- Adds the time stamp to each note. Dunno whats the best place for it so I just put it at the beginning.

I know its a bit hackish, however, I wanted to avoid breaking as less things as possible.

Dunno if the latest news block will be used somewhere else, I just remove it.

### DIFF
--- a/Sources/ModerationCenter.php
+++ b/Sources/ModerationCenter.php
@@ -224,11 +224,12 @@ function ModerationHome()
 	$context['page_title'] = $txt['moderation_center'];
 	$context['sub_template'] = 'moderation_center';
 
+	// Handle moderators notes.
+	ModBlockNotes();
+
 	// Load what blocks the user actually can see...
-	$valid_blocks = array(
-		'n' => 'LatestNews',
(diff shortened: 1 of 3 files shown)


MissAllSunday:
Oh, I couldn't reproduce #635 because I have to enable the sidebar. didn't knew that. Oh well, theres a proposal to remove it anyways.

live627:
But why remove the sidebar?

MissAllSunday:
It is only a proposal.

Arantor:
We did propose to remove the sidebar for a variety of reasons, not least solving some other issues (because it was already supposed to have been removed anyway by now)

Issue #426, opened 2013-06-14
Collapse/Expand categories


PortaMx:
What I don't understand ..
Why do the category collapse/expand send a request directly?
Why not collapse the category with the smc_toggle and send a ajax request to the server?
I think that is simple and faster to collase/expand ...

emanuele45:
Probably mainly because nobody wrote it.

If you intend to, remember that on expanding it implies an ajax call to retrieve the category structure because if a category is collapsed nothing from that category is loaded.

Arantor:
It would be faster for the user, sure, but it would necessitate more work from the server on board index.

Would it be nice to AJAXively get the partial board index and inject that when expanding?

beratdogan:
I'm working on it.

Arantor:
Closing this issue and referring to 663 instead.

PR #663, opened 2013-10-04
Collapse/Expand Categories with AJAX

beratdogan:
Adds AJAX support to toggle categories without any page refreshings, like wanted here #426.

I can't use slideToggle() effect because it doesn't support tables. For this reason I used fadeToggle() effect.

Please ignore the ce4d586 commit.


Arantor:
On the face of it it will collapse a category with AJAX but if you try to expand a category it doesn't appear to fetch the data to repopulate the category.

I'm also not sure what the deal is with binary files being marked as modified, you might want to check your local git configuration file to make sure that autocrlf is set to false (so it always adheres to what's in the repo)

emanuele45:
@Arantor he just loads the data of any category irrespective of the collapsed/uncollapsed status, so on "expand" everything is already there because is just `display:none`.

@grafitus then I have a couple of "suggestions":
1) if it is something like that you may just use smc_Toggle and even get rid of the collapsed_categories table
2) when you collapse/uncollapse the button doesn't change "state".
3) I wouldn't use toggle (on tables has a weird effect of delaying the collapse and it looked simply broken when I first tried it).
oh, and 4) there is no need to add another parameter (namely ajax), you can use xml that is already associated with most of the xml actions. ;) (What speaks here is my own experience of adding yet another useless param to the url :P)

Arantor:
@emanuele45 I thought the board index query didn't bother to load the data in those situations? I haven't examined the commit, though.

And yes, there is the infamous ;xml parameter...

Even better, of course, if you're sending something via AJAX properly, check for the X-Requested-With header being set to XMLHttpRequest.

emanuele45:
> @emanuele45 https://github.com/emanuele45 I thought the board index
> query didn't bother to load the data in those situations? I haven't
> examined the commit, though.
> 
> I thought the same, but apparently we were wrong... lol
> The only thing that skip is all the "preparation" starting from:

```
// Let's save some typing.  Climbing the array might be slower, anyhow.
```

and on.

Arantor:
Another thing to rewrite in the logic then.

live627:
This can be closed now.

PR #664, opened 2013-10-05
Third level menus cannot have subsections

live627:
My new theme uses arrows on .subsections, and this fix prevents them from going on the third level menus, as that is the deepest level of nesting that SMF allows.

### DIFF
--- a/Themes/default/GenericMenu.template.php
+++ b/Themes/default/GenericMenu.template.php
@@ -167,7 +167,7 @@ function template_generic_menu_dropdown_above()
 					$url = isset($sub['url']) ? $sub['url'] : (isset($area['url']) ? $area['url'] : $menu_context['base_url'] . ';area=' . $i) . ';sa=' . $sa;
 
 					echo '
-							<li ', !empty($area['subsections']) ? ' class="subsections"' : '', '>
+							<li>
 								<a ', !empty($sub['selected']) ? 'class="chosen" ' : '', ' href="', $url, $menu_context['extra_parameters'], '">', $sub['label'], '</a>
 							</li>';
 				}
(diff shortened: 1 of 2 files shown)


Arantor:
I'm curious about whether this will have any repercussions in SMF, or indeed why you need what seems to be a 4th level menu (which is usually bad for usability)? Or have I missed the point?

live627:
Eh what? 4th level?

Currently, SMF 2.1 always adds that class to 3rd level because the condition will always return true.

Arantor:
Surely a third level menu having a subsection would create a fourth level menu?

Arantor:
OK, so I had a look at this again. I'm still not seeing what benefit this provides. I don't get why you need to change core SMF code to achieve this... if the arrows are in CSS, use CSS to select the right thing, if it's JS, use selectors to select the right thing.

I guess I'm just not seeing how this benefits the core. :(

Oldiesmann:
Although SMF shouldn't be adding the subsections class there, the fact that it's causing arrows to show up in your theme but not the default theme indicates that it's a problem with your CSS that's causing the bug to appear.

deliciousbamboo:
> I guess I'm just not seeing how this benefits the core. :(

Thats because there are none, this is a fix the core because of poor theme coding, nothing else

Arantor:
That's what I thought.

live627:
> Thats because there are none, this is a fix the core because of poor theme coding, nothing else

Pretty much the gist (no pun intended) of it.

See my latest commit; my theme does not have that hack, which is why the problem shows.

live627:
Bump. Is this going into the codebase?

Arantor:
Can't be automatically merged for some reason.

live627:
#1396

PR #668, opened 2013-10-06
More PHP 5.5 fixes found while working on 2.0 PHP 5.5 Combat

jdarwood007:
I found some more PHP 5.5 changes needed while working on the 2.0 Combat fixes.

Some of these where not fun to fix, and hopefully through my limited testing no more regressions occur.  I tested all of them the best I could to ensure they function as they should.

Signed-off-by: jdarwood007

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,3 +8,4 @@
 *.tgz binary
 *.zip binary
 *.tar.gz binary
+*.ttf binary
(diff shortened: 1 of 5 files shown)


jdarwood007:
Thanks for looking those over Spuds :+1: 
I am hoping that is really the last of them.  I found those this time by doing:
grep -Rn "preg_replace" | grep "e'"

I already did searches for these previously
grep -Rn "~ie"
grep -Rn "~e"

PR #670, opened 2013-10-06
! Make expandPages() nest properly: Fixes #669

### LINKED ISSUE #669: expandPages() doesn't nest properly
Once the three dots are clicked, the new page links that are loaded get wrapped into the expand_pages span. Also, the new link to expand pages is missing the proper class.

live627:


live627:
It works in  all my tests, incl. the message index and topic display.

PR #681, opened 2013-10-07
Add some text to differentiate these 2 exact queries

MissAllSunday:
Frivolous PR, Post.php has the exact same query on 2 different places, this just adds some text to differentiate them. Quick and dirty fix...

### DIFF
--- a/Sources/Post.php
+++ b/Sources/Post.php
@@ -493,7 +493,7 @@ function Post($post_errors = array())
 		// Previewing an edit?
 		if (isset($_REQUEST['msg']) && !empty($topic))
 		{
-			// Get the existing message.
+			// Get the existing message. Previewing.
 			$request = $smcFunc['db_query']('', '
 				SELECT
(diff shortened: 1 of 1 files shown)


emanuele45:
/me wonders when SMF will decide to use functions instead of repeating the exact same queries over and over and over... (Disclaimer: just a comment floating in my mind, nothing against you or your PR, MAS... especially if the thing you are brandishing in your picture is what I think it is :P )

MissAllSunday:
I thought about converting it to a function but in all honesty, the chances of mod authors actually changing this specific part of the code are small, as far as I know, only Niko (https://github.com/nikop/Related-Topics) and I (https://github.com/MissAllSunday/saiom) have actually changed those two specific places. So this is just a selfish PR actually :P

As for an overhaul, yes it is needed but 2.1 isn't the best place for it :(

And yes, it is a machete :P

Arantor:
@emanuele45 It's amongst the many things for 3.0, however at this stage such a change is not something we can entertain for too much for 2.1.

PR #689, opened 2013-10-08
Mod center improvements

MissAllSunday:
As per Mantis 4416 bug report.

This PR adds some functionality to the mod center
- Adds a delete message.button
- Adds a ban this user button, handles guest posting too.
- Adds a confirmation message when doing an action (ignore, close, delete, open) to an report.

There are some things that do bother me, 
- Theres a nasty inline-double ternary to find out which action the user is doing.
- I added some new text strings so the usual check must be done, it feels weird typing "un-ignore" or "un-ignored".
- Dunno if a mod_log should be set when deleting a message.
- Is it worth to also apply this changes to the report details page?

### DIFF
--- a/Sources/ManageBans.php
+++ b/Sources/ManageBans.php
@@ -536,6 +536,27 @@ function BanEdit()
 					$context['ban_suggestions']['other_ips'] = banLoadAdditionalIPs($context['ban_suggestions']['member']['id']);
 				}
 			}
+
+			// We come from the mod center.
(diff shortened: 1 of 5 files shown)


Arantor:
I'm honestly not sure what is still needed here. I kind of think we need to try it out and see what's needed. Not worried about the ternary stuff at this stage.

Deleting message should push it automatically to the modlog because it uses action=deletemsg?

PR #691, opened 2013-10-08
Performance enhancement to SSI's ssi_recentTopics

live627:
Use two faster queries instead of that slow monstrosity.

### DIFF
--- a/SSI.php
+++ b/SSI.php
@@ -451,33 +451,52 @@ function ssi_recentTopics($num_recent = 8, $exclude_boards = null, $include_boar
 	// Find all the posts in distinct topics.  Newer ones will have higher IDs.
 	$request = $smcFunc['db_query']('substring', '
 		SELECT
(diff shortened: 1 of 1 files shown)


SSI.php
```diff
@@ -451,35 +451,54 @@ function ssi_recentTopics($num_recent = 8, $exclude_boards = null, $include_boar
...
 			'include_boards' => empty($include_boards) ? '' : $include_boards,
 			'exclude_boards' => empty($exclude_boards) ? '' : $exclude_boards,
 			'min_message_id' => $modSettings['maxMsgID'] - 35 * min($num_recent, 5),
 			'is_approved' => 1,
 		)
 	);
+	$topics = array();
+	while ($row = wesql::fetch_assoc($request))
+		$topics[$row['id_topic']] = $row;
+	wesql::free_result($request);
```
emanuele45, line 476: /me feels there is something wrong here. :P

Arantor:
o.O I didn't study it that closely, I just looked over the changes to the query as a whole which to me was the more important part :P

Arantor:
And fixed.

SSI.php
live627, line 476: Major oops.

PR #700, opened 2013-10-10
! Making members pending approval and moderation reports more prominent....

Arantor:
... The moderation menu, like most of the time, has it wrong anyway and a separate "accident" may happen to it in the future, but for now, let's get rid of icky (and frequently wrong, as well as some old) language strings and be a bit daring.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -2884,10 +2884,8 @@ function setupThemeContext($forceload = false)
 		$_SESSION['unread_messages'] = $user_info['unread_messages'];
 
 		if (allowedTo('moderate_forum'))
-		{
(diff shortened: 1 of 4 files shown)


illori:
Alt text

i have one moderation report, seems that a guest can view the 1 that is waiting which it should not be able to do.

i also get a few errors in index.template.php related to href active_button title being undefined

i also get an error of undefined index moderate on line 4049 of subs.php

if the user is logged in the Reported posts text along with the number of reports shows in the header as it should only if you can do something with the reports

illori:
that seems to be an issue with the cache and once it got cleared the issue went away.

PR #705, opened 2013-10-11
+ Allow users to turn off whether bans go into the error log or not. And...

Arantor:
... if they do, let's have a category just for them.


Oldiesmann:
Thank you! This is long overdue :+1:

PR #722, opened 2013-10-14
Proxy Server IP header support

jdarwood007:
This adds support for SMF for handling Proxy Servers, such as a CDN or Reverse Proxy.  We can now specify to do nothing, autodetect (SMF default) or only allow a single header to respond.  As well we setup allowed IP ranges. The main reason behind this is security and for ease of expandability to other headers that obscure proxy headers use.

I should note that most servers are supposed to handle these securely as it is, but it doesn't hurt for us to be sure as well.  This also adds a function to SMF which may be useful in the future, but for now was used for proxy servers that have you use CIDR entries.

Final note is Subs-Charset got pulled in on this.  It appears to be line ending differences that need fixed.  If needed I can attempt to remove it, otherwise I would just let us pull the changes and fix the line endings to match what we are now forcing with .gitattributes


Themes/default/Register.template.php
```diff
@@ -170,6 +170,15 @@ function verifyAgree()
 					</dl>';
 	}
 
+		echo '
```
live627: Why is this here? Not the line, but this entire code block..... a seemingly unrelated part of this commit?

Arantor: We were doing some testing based on some stuff discussed in the SMF chatroom, and this may be related but it's not part of the feature AFAIK.

jdarwood007: Ooops, I forgot to revert those changes before committing.  Just uploaded a commit to revert that.

Oldiesmann:
Looks good to me. The Subs-Charset thing could be the line-endings, but I've never had that problem on my end (using GitHub For Windows, the "posh-git" thing GH includes or SourceTree/regular git via Mingw), so I have no idea why some people are having that problem but others aren't. I guess it can't hurt to pull it in though.

Arantor:
OK, so I changed up Subs-Charset so hopefully this should not happen so much in the future.

Oldiesmann:
This is showing a merge conflict now.

Arantor:
It might be nice to also support http://dev.simplemachines.org/mantis/view.php?id=1564 as part of this, where we show both IP addresses to the user.

butchs:
Interesting I just did something similar but with an emphasis on spoof detection.  I am not sure but I think you may have missed repeated multiple message-header field per RFC2616?

Please note that Cf-Connecting-Ip should be changed to HTTP_CF_CONNECTING_IP since you are using $_SERVER to get the HTTP IP address.   Cf-Connecting-Ip was originally used to verify that CF is online.

PR #728, opened 2013-10-16
Typo fix & 'unable to verify referring URL' fix

HoverClub:
Subscription settings message doesn't make sense?

Potential fix for the 'Unable to verify...' issue.


HoverClub:
New to GitHub :$

Arantor:
Firstly, please ensure you sign off your commits before pushing them to us otherwise we can't use them.

Secondly, that second commit doesn't actually work?

emanuele45:
I've never heard of that issue.

Also using str_[i]replace what would happen if the domain is for example `www.mydomain.com` and the referrer is a domain like `www.mywww.domain.com`?
I feel the subsequent comparison would give green light when instead the two are clearly different domains opening the door for possible attacks.

Arantor:
I haven't either, thinking about it. Certainly HTTP_REFERER is implicitly untrustable because it's from the client and plenty of security software strips it for 'security'

But ultimately consider what we're talking about: the referer being provided has a domain that it shouldn't. I'm not sure this is actually strictly an SMF issue, it is at least also a configuration issue where the site is accepting www. as a subdomain where it isn't configured as such, means that maybe wwwRedirect isn't working quite as it should.

HoverClub:
Apologies for the typo - it was pasted from a working site to the repository, obviously incorrectly!

As to signing off - I can't see where I do that?

This "unable.." error happens sporadically on a lot of sites - I've seen it many times even with www ReWrite in htaccess.  Strictly it's not an SMF issue (other than using an unreliable variable) BUT if would be a lot simpler for most users if it was fixed in SMF (then they don't have to learn how to config a web server).

emanuele45:
Are you sure the issue of the referring url issue is the www?
I'm asking because there is a bug in 2.0 (fixed in 2.1) about referring URL, but the problem was related to collapsed "things" (usually the header) that was re-setting the collapsed state at every page load creating one version of the referring url problem.

HoverClub:
I've noted that bug but I've seen this error on other actions.  Isn't the error just a trigger?  Question is why does the user agent use a different http_referer value on a jsoption query?  This is what I've seen:

```
[HTTP_HOST] => somesite.com
[HTTP_REFERER] => http://www.somesite.com/forum/
[REQUEST_URL] => http://somesite.com/forum/index.php?action=jsoption;var=collapse_header;val=1;bcb6bc7fc4e=da2d970eefa68620a16e891769cdfa81;time=1381921671818
```

}

You have no control over the referer_url sent by the browser therefore it shouldn't be relied upon for security.  Removing any www. will make it more tolerant of user and server config errors but maybe it shouldn't be used at all?

emanuele45:
That's the error I mentioned about, is not due to the "www" part, but to the fact that the header is collapsed.
Try uncollapsing the header and it won't give you the error.

Arantor:
Psst, collapse_header went a bit back ;)

emanuele45:
Ops...I missed that one... :angel: 
Though the url posted by HoverClub refers to collapse_header, so it is probably from 2.0 that is still buggy in that respect.

PR #736, opened 2013-10-17
More upgrade bugs :/

Oldiesmann:


emanuele45:
Commenting here just not to open a bug report, but I'm not sure the postgre and sqlite version of the upgrade work...
AFAIK none of the two support multiple INSERTs and upgrade_query just sends the query string to $smcFunc['db_query'] that executes the query as it is.

Oldiesmann:
I've already accounted for that in the appropriate files... In the MySQL file it implodes() the array and inserts everything in one go. In PostgreSQL and SQLite, we loop through the array via foreach() and insert one row at a time.

Arantor:
And this is why we should be using $smcFunc['db_insert'] :tongue:

PR #738, opened 2013-10-18
fix missing txt string $txt['board_moderator_groups']

illori:


Oldiesmann:
Fixed via #746.

PR #777, opened 2013-10-28
Some more goodies for SMF

Arantor:


Arantor:
Bah. Yes, somehow my repo got screwed up vs the master one so that last commit is basically remerging everything that's been added lately that wasn't me.

PR #786, opened 2013-11-02
! Fix a few calendar bugs

Oldiesmann:


Arantor:
Are we not interested in fixing the upgrader too?

I seem to recall publishing updated versions of the SQL files for that...

PR #789, opened 2013-11-03
fixing missing id in input

illori:
this should be correct, i did not test all of them.

### DIFF
--- a/Sources/Profile.php
+++ b/Sources/Profile.php
@@ -739,12 +739,12 @@ function loadCustomFields($memID, $area = 'summary')
 		if ($row['field_type'] == 'check')
 		{
 			$true = (!$exists && $row['default_value']) || $value;
-			$input_html = '<input type="checkbox" name="customfield[' . $row['col_name'] . ']" ' . ($true ? 'checked="checked"' : '') . ' class="input_check" />';
+			$input_html = '<input type="checkbox" name="customfield[' . $row['col_name'] . ']" id="customfield[' . $row['col_name'] . ']"' . ($true ? 'checked="checked"' : '') . ' class="input_check" />';
 			$output_html = $true ? $txt['yes'] : $txt['no'];
 		}
 		elseif ($row['field_type'] == 'select')
 		{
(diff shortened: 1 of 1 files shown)


live627:
HTML ID cannot have brackets. Use underscore instead.

illori:
well it did validate for me with the brackets in place on 2.0 using a html5 theme.

emanuele45:
I think they are at least "annoying" to use in javascript, but I may be wrong.
Also:

```
 id="customfield[' . $row['col_name'] . ']"' . ($true ? 'checked="checked"
```

will generate invalid HTML when checked because it will result in:

```
 id="customfield[something]"checked="checked"
```

without space between the two.
Same for the type text and the textarea.

Oldiesmann:
I did find one site that said brackets weren't valid for IDs, but I can't find any official reference to that. However, because brackets have special meaning in CSS (see http://www.w3.org/TR/CSS21/selector.html#attribute-selectors), I think it would be best to use underscores instead.

live627:
w3 trollfest! hahaa :P

emanuele45:
http://www.w3.org/TR/html5/dom.html#the-id-attribute

> The value must be unique amongst all the IDs in the element's home subtree and must contain at least one character. The value must not contain any space characters.
> 
> There are no other restrictions on what form an ID can take; in particular, IDs can consist of just digits, start with a digit, start with an underscore, consist of just punctuation, etc.

Arantor:
That's HTML5. We're not using HTML5 yet, therefore the rules from XHTML 1.0 take precedence which are not the same and derive from XML

As per http://www.w3.org/TR/REC-xml/#sec-attribute-types an id must match the Name production which is defined in http://www.w3.org/TR/REC-xml/#NT-Name - and [ and ] are not on the list of NameChars or the NameStartChar.

So finding an alternative it is.

emanuele45:
I thought the doctype was changed a while ago when Lab was working on moving to html5, sorry. ;)

Arantor:
It has occurred to me to do so, however.

Arantor:
Well, now we're using an HTML5 doctype, we can fix this.

PR #817, opened 2013-11-07
Version check for themes

MissAllSunday:
As per mantis report 4754:
- Checks if a mod was installed via theme install, yes, it can happen...
- Adds an < install > tag, works the same as the packager manager one.
- Uses the version tag to check for possible updates.
- Solve a couple of todo tasks.

### DIFF
--- a/Sources/Themes.php
+++ b/Sources/Themes.php
@@ -1518,11 +1518,10 @@ function ThemeInstall()
 				'theme_templates' => 'templates',
 				'based_on' => 'based-on',
 			);
+
 			foreach ($xml_elements as $var => $name)
-			{
 				if (!empty($theme_info_xml[$name]))
 					$install_info[$var] = $theme_info_xml[$name];
-			}
 
 			if (!empty($theme_info_xml['images']))
 			{
@@ -1536,11 +1535,14 @@ function ThemeInstall()
 
 		if (isset($install_info['based_on']))
 		{
+			// No need for elaborated stuff when the theme is based on the default one.
 			if ($install_info['based_on'] == 'default')
 			{
 				$install_info['theme_url'] = $settings['default_theme_url'];
 				$install_info['images_url'] = $settings['default_images_url'];
 			}
+
+			// Custom theme based on another custom theme, lets get some info.
 			elseif ($install_info['based_on'] != '')
 			{
 				$install_info['based_on'] = preg_replace('~[^A-Za-z0-9\-_ ]~', '', $install_info['based_on']);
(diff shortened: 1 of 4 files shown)


Sources/Themes.php
```diff
@@ -1366,7 +1371,7 @@ function ThemeInstall()
 		package_flush_cache();
 
 		$theme_name = $_REQUEST['copy'];
-		$images_url = $boardurl . '/Themes/' . basename($theme_dir) . '/images';
+		$images_url = $themedir . '/' . basename($theme_dir) . '/images';
```
emanuele45, line 1374: `$themedir != $boardurl . '/Themes'`

Sources/Themes.php
```diff
@@ -1443,42 +1453,65 @@ function ThemeInstall()
...
 
-	// Something go wrong?
+	// Let us proceed with the install.
 	if ($theme_dir != '' && basename($theme_dir) != 'Themes')
 	{
 		// Defaults.
 		$install_info = array(
 			'theme_url' => $boardurl . '/Themes/' . basename($theme_dir),
-			'images_url' => isset($images_url) ? $images_url : $boardurl . '/Themes/' . basename($theme_dir) . '/images',
+			'images_url' => isset($images_url) ? $images_url : $themedir . '/' . basename($theme_dir) . '/images',
```
emanuele45, line 1479: Same as above

Sources/Themes.php
MissAllSunday, line 1374: Apparently one can't make mistakes...   eww, I would offer the same "help" back but I don't give a rat ass about other projects nor do I wait to know what other people are doing so I can "borrow" it... I mind my own business... you should do the same...

emanuele45, line 1374: Fuck.
It's the first time I reach that point with a person.

I'm pointing out errors in all the code I read, if you think I didn't like you, you were wrong, now you are right.
If you prefer to have broken code in the repo, help yourself.

MissAllSunday, line 1374: Nope, you're being snarky and you're only "helping" or at least you take special consideration whenever I submit something...

I don't prefer to have broken code, thats why betas are for...

I respond the very same way I've been treated, if you keep being snarky and sarcastic this is what you're going to get, deal with it. Like I said, just mind your own business and I'll mind mine, everyones happy.

emanuele45, line 1374: OMG.
LOL

Feel free to live happily in your dreaming world.

MissAllSunday, line 1374: Now that you will refrain from having any kind of interaction with me I'll think my dreaming world will be a much better place, thanks!

PR #824, opened 2013-11-07
Remove extra legend

live627:
Post Moderation has two identical legends; this removes the top one.


Oldiesmann:
The top legend was added a while back to make it more visible.

Arantor:
I'd probably fix the spacing on the top one and remove the bottom one, personally.

Oldiesmann:
So you'd like to make the top legend horizontal instead of vertical? I can look into that

Arantor:
No, that's not what I meant. The top one is jammed up against the main content, as opposed to the bottom one that has proper spacing between the bottom of the permissions grid and the legend. I'd rather drop the bottom one and add some space between the top one and the top of the permissions grid.

PR #837, opened 2013-11-09
! Make update notifier look better by simplifying it, using a single CSS...

live627:
... class, and by removing inline CSS

_How to test_

Modify $forum_version to 'SMF 2.0.1'

### DIFF
--- a/Themes/default/Admin.template.php
+++ b/Themes/default/Admin.template.php
@@ -116,7 +116,7 @@ function template_admin()
 					<script type="text/javascript"><!-- // --><![CDATA[
 						var oAdminIndex = new smf_AdminIndex({
 							sSelf: \'oAdminCenter\',
-			
+
 							bLoadAnnouncements: true,
 							sAnnouncementTemplate: ', JavaScriptEscape('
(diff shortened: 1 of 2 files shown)


live627:
42cb0a88ccc8ca94e0d6b96bc4f5ae67ec4e620e adds them.

My editor automatically strips them upon save.

Arantor:
Um, I don't see any blank lines being added in that commit, I see only indentation being fixed. This commit adds a bunch of blank lines throughout it that weren't there before.

live627:
Eh what? It's removing whitespace.

Oldiesmann:
@Arantor Those blank lines have apparently been there a while, as they were there when I made my changes earlier. His commit is removing extra tabs that I inadvertently added in 42cb0a88ccc8ca94e0d6b96bc4f5ae67ec4e620e. I'm not sure when the blank lines were added, but we can clear those out at some point.

Arantor:
Wait... it's now different to before because when I first looked at this commit it was just adding blank lines - not removing empty lines with just tabs...

PR #841, opened 2013-11-09
Use .infobox more

live627:
Use .infobox more

### DIFF
--- a/Themes/default/ManageMaintenance.template.php
+++ b/Themes/default/ManageMaintenance.template.php
@@ -18,7 +18,7 @@ function template_maintain_database()
 	// If maintenance has finished tell the user.
 	if (!empty($context['maintenance_finished']))
 		echo '
-			<div class="maintenance_finished">
+			<div class="infobox">
 				', sprintf($txt['maintain_done'], $context['maintenance_finished']), '
 			</div>';
 
@@ -113,7 +113,7 @@ function template_maintain_routine()
 	// If maintenance has finished tell the user.
 	if (!empty($context['maintenance_finished']))
 		echo '
-			<div class="maintenance_finished">
+			<div class="infobox">
 				', sprintf($txt['maintain_done'], $context['maintenance_finished']), '
 			</div>';
 
@@ -252,7 +252,7 @@ function template_maintain_members()
 	// If maintenance has finished tell the user.
 	if (!empty($context['maintenance_finished']))
 	echo '
-		<div class="maintenance_finished">
+		<div class="infobox">
 			', sprintf($txt['maintain_done'], $context['maintenance_finished']), '
 		</div>';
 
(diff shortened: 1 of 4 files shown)


Oldiesmann:
You might want to use .noticebox instead as that's similar to the styles that are currently used.

live627:
But that isn't green.

Oldiesmann:
I see that now. I was thinking they were closer because noticebox uses an ! icon like the box for the styles you're removing do. For some reason I was thinking they both used the same icon, but they don't.

However, after comparing .infobox with the maintenance box (what you're proposing we remove), I think it would be better to change the style for .infobox to what's used for the"task completed" thing, as it looks much nicer. That way we can still drop the CSS and use infobox while making other areas that use infobox look a bit nicer.

infobox
task_completed

Kindred-999:
perhaps define the icon to be displayed?  like (!) (?)

Oldiesmann:
Even if we add an icon to infobox, I still think the maintenance box looks much nicer.

NanoSector:
We should have consistency througout the software. Having different types of information/success boxes kills that.

Oldiesmann:
I agree that we need some consistency. The trick will be finding the best way to handle things, as there are no less than 6 different "box" types defined in various places.

Two different maintenance ones (the ones that this PR would remove) - see first image above (second one, which is used to tell you that the selected scheduled tasks were run successfully, is identical but has square corners instead of rounded ones)

descbox - a generic transparent box with rounded corners (example: Admin -> Main -> Package Manager - the "No packages installed yet" box)
infobox - 1st image above
noticebox - similar to infobox, but with orange borders, a yellowish-orange background and an orange exclamation point icon
errorbox - similar to noticebox but with red borders, a reddish-pink background a red exclamation point icon

Live examples of each:
Maintenance 1 - Admin -> Maintenance -> Forum Maintenance -> Routine. Click "Run task now" for any task but the file version check one.
Maintenance 2 - Admin -> Maintenance -> Scheduled Tasks. Run one of the scheduled tasks now.
descbox - Admin -> Main -> Package Manager. It's the "No packages installed yet" box
infobox - Go to a settings page in the admin center and save some settings. This is the box telling you they've been saved successfully
noticebox - Admin -> Main -> Package Manager -> File Permissions. It's the big box with the "Please Note" text.
warningbox - put the forum in maintenance mode or copy upgrade.php to the forum directory.

My suggestion here would be to merge the two maintenance boxes with infobox, and modify the infobox CSS to be similar to the two maintenance boxes. Then we can figure out a way to standardize things from there (a template of some sort that specifies the type of box and its contents would work maybe).

live627:
fireshot screen capture 125 - forum maintenance - localhost_smf2_1_index_php_action admin area maintain sa routine activity cleancache

How's this? It was too light before.

Arantor:
infobox was generally meant as a replacement for the old profile_success format, and everything indicating a successful action should really use that (much as errorbox is for anything that indicates a failure)

Oldiesmann:
Looks much better now :)

PR #919, opened 2013-11-20
Don't show debug info in the popups

live627:
omg those tabs....

I don't think anything's broken. Famous last words! (At least, not in my tests)

EDIT: Damn formatting


Arantor:
The XML parameter should _not_ be added. This causes other things to happen that probably shouldn't, like the fact you're causing it to issue a Content-Type: text/xml header which will definitely cause issues there in some browsers (especially IE)

live627:
Forgot about the header. Should a constant be used instead, such as POPUP?

PR #921, opened 2013-11-20
! Don't allow moving of topics to boards that the user can't move topics...

Oldiesmann:
... on...

### DIFF
--- a/Sources/MoveTopic.php
+++ b/Sources/MoveTopic.php
@@ -59,13 +59,13 @@ function MoveTopic()
 		if ($id_member_started == $user_info['id'])
 		{
 			isAllowedTo('move_own');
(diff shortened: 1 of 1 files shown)


Arantor:
The logic of this doesn't actually work. You're getting the list of boards that a user has that permission in and using that as the possible destinations... what happens if you have, say, an archive board that you want to move _to_ but otherwise lock down?

Now you have to give them that permission for those boards as well as the boards they can move _from_. Honestly it would be better to base on the boards where the user could create a new topic. And make it optional because it would interfere with sites who do actually need it to work as it currently is without rejigging all the permissions again.

emanuele45:
Hey, the code was commented out by Unknown in 2004 (21/04/2004) ... time to have it back probably. :P (we brought it back too O:-))

Anyway, agree, better the post_new.

PR #922, opened 2013-11-20
! Enforce new "move topic" restrictions for quick moderation as well

Oldiesmann:


Arantor:
And everything I said in https://github.com/SimpleMachines/SMF2.1/pull/921#issuecomment-28933100 applies here too.

PR #923, opened 2013-11-20
Theme stuff

MissAllSunday:
A bigger refactor for installing and overall handling themes.
- Separates the 3 options to install themes into 3 different functions.
- Adds a helper file with, well, helper functions.
- Adds more checks for common mistakes.
- Removes some old code when copying a file from another based on theme, I haven't see any theme based on another then other than the default one so the copy a file feature goes straight to the default theme now.
- A couple of hooks, just because.
- Saves some queries on specific cases, no more inner joins either.
- A small template change to separate the 3 install options.
- Adds a enable/disable feature, themes marked as enable will be the only ones eligibles to use.
- When deleting a theme it also gets deleted from the Themes folder.
- Sets $modSettings['enableThemes'] for a quick and easy way to get all enabled themes.
- The version tag now holds the theme version rather than the forum's one.
- Themes made for 2.0 or lower would not be installed.
- Themes now require a <install for"">  tag.
- When a theme is installed via a package, it gets dumped on a temp dir first, if the install fails it gets removed, on success it renames the temp dir to the new theme's name.


Oldiesmann:
Just wanted to let you know that I'm not ignoring this. I just haven't had a chance to look at it thoroughly yet. I'll try to go through it when I'm off on Monday.

MissAllSunday:
Sure! Thanks :D

En Sat, 23 Nov 2013 11:31:47 -0600, Michael Eshom  
notifications@github.com escribió:

> Just wanted to let you know that I'm not ignoring this. I just haven't  
> had a chance to look at it thoroughly yet. I'll try to go through it  
> when I'm >off on Monday.
> 
> —
> Reply to this email directly or view it on GitHub.

## 

Sources/Subs-Themes.php
```diff
@@ -0,0 +1,587 @@
...
+ * @return array The theme info as an array.
+ */
+function get_single_theme($id)
+{
+	global $smcFunc, $modSettings;
+
+	// No data, no fun!
+	if (empty($id))
+		return false;
+
```
Oldiesmann, line 33: It would be a good idea to make sure $id is an int before passing it to call_integration_hook and the query. Even though it's not likely that it won't be an int, you're better safe than sorry.

Sources/Subs-Themes.php
```diff
@@ -0,0 +1,587 @@
...
+			'theme_values' => $themeValues,
+			'id_theme' => $id,
+			'no_member' => 0,
+		)
+	);
+
+	while ($row = $smcFunc['db_fetch_assoc']($request))
+		$single[$row['variable']] = $row['value'];
+
+	// Fix the path and tell if its a valid one.
```
Oldiesmann: For this section, you're probably better off setting all this in the while loop - otherwise you'll get errors if for some reason the query doesn't return anything.

Sources/Subs-Themes.php
```diff
@@ -0,0 +1,587 @@
...
+		}
+
+		$context['themes'][$row['id_theme']]['known'] = in_array($row['id_theme'], $knownThemes);
+		$context['themes'][$row['id_theme']]['enable'] = in_array($row['id_theme'], $enableThemes);
+		$context['themes'][$row['id_theme']][$row['variable']] = $row['value'];
+	}
+
+	$smcFunc['db_free_result']($request);
+
+	return $context['themes'];
```
Oldiesmann: Since $context is a global variable, there's no need to return it here - you can just set it and it'll be accessible immediately after calling the function.

Sources/Subs-Themes.php
```diff
@@ -0,0 +1,587 @@
...
+
+	// OK, is this a newer version of an already installed theme?
+	if (!empty($context['to_install']['version']))
+	{
+		$to_update = array();
+		$request = $smcFunc['db_query']('', '
+			SELECT id_theme, variable, value
+			FROM {db_prefix}themes
+			WHERE id_member = {int:no_member}
+				AND id_member = {int:no_member}
```
Oldiesmann: You don't need the "AND id_member ..." line here since you have the same thing on the line above.

Sources/Subs-Themes.php
MissAllSunday: Yep, I'm aware of that :) I'll did it like that because I don't like to depend on globals that much, on some specific cases (namely instantiated methods) relying on a global var would be somehow problematic, by returning the value you can keep working without using globals. Don't think the return affects that much.

Sources/Subs-Themes.php
```diff
@@ -0,0 +1,587 @@
...
+		$to_update = $smcFunc['db_fetch_assoc']($request);
+		$smcFunc['db_free_result']($request);
+
+		// Got something, lets figure it out what to do next.
+		if (!empty($to_update) && !empty($to_update['version']))
+			switch (compareVersions($context['to_install']['version'], $to_update['version']))
+			{
+				case 0: // This is exactly the same theme.
+				case -1: // The one being installed is older than the one already installed.
+				default: // Any other possible result.
```
Oldiesmann: The "default" case should be the last one in the list. Otherwise it will hit that if the value is 1.

Sources/Themes.php
```diff
@@ -884,56 +813,54 @@ function RemoveTheme()
 	validateToken('admin-tr', 'request');
 
 	// The theme's ID must be an integer.
-	$_GET['th'] = isset($_GET['th']) ? (int) $_GET['th'] : (int) $_GET['id'];
+	$themeID = isset($_GET['th']) ? (int) $_GET['th'] : (int) $_GET['id'];
 
 	// You can't delete the default theme!
 	if ($_GET['th'] == 1)
```
Oldiesmann: If you're not going to use $_GET['th'], you need to change this line to use $themeID instead.

Sources/Themes.php
```diff
@@ -1288,394 +1217,289 @@ function PickTheme()
...
+	// Create the temp dir.
+	mkdir($dirtemp, 0777);
 
-	elseif (isset($_REQUEST['theme_dir']) && $method == 'path')
-	{
-		if (!is_dir($_REQUEST['theme_dir']) || !file_exists($_REQUEST['theme_dir'] . '/theme_info.xml'))
-			fatal_lang_error('theme_install_error', false);
+	// Hopefully the temp directory is writable, or we might have a problem.
+	if (!is_writable($dirtemp))
+		fatal_lang_error('theme_install_write_error', 'critical');
```
Oldiesmann: You might want to try chmodding it to 755 first before throwing an error. `@chmod($dirtmp, '0755');`

Sources/Subs-Themes.php
```diff
@@ -0,0 +1,596 @@
...
+ */
+function get_single_theme($id)
+{
+	global $smcFunc, $modSettings;
+
+	// No data, no fun!
+	if (empty($id))
+		return false;
+
+	// Make user its an int.
```
live627: Grammar error detected.

Sources/Subs-Themes.php
```diff
@@ -0,0 +1,596 @@
...
+		$smcFunc['db_insert']('insert',
+			'{db_prefix}themes',
+			array('id_theme' => 'int', 'variable' => 'string-255', 'value' => 'string-65534'),
+			$inserts,
+			array('id_theme', 'variable')
+		);
+
+	// Update the known and enable Theme's settings.
+	updateSettings(array('knownThemes' => strtr($modSettings['knownThemes'] . ',' . $id_theme, array(',,' => ','))));
+	updateSettings(array('enableThemes' => strtr($modSettings['enableThemes'] . ',' . $id_theme, array(',,' => ','))));
```
Oldiesmann: You only need to call updateSettings once to update both settings. To avoid having a ridiculously long line of code, just do something like this:

``` php
$known = strtr($modSettings['knownThemes'] . ',' . $id_theme, array(',,' => ','));
$enable = strtr($modSettings['enableThemes'] . ',' . $id_theme, array(',,' => ','));
updateSettings(array('knownThemes' => $known, 'enableThemes' => $enable));
```

Sources/Subs-Themes.php
```diff
@@ -0,0 +1,596 @@
...
+	$enable = array_diff($enable, array($themeID));
+
+	// Back to good old comma separated string.
+	$known = strtr(implode(',', $known), array(',,' => ','));
+	$enable = strtr(implode(',', $enable), array(',,' => ','));
+
+	// Update the enableThemes list.
+	updateSettings(array('enableThemes' => $enable));
+
+	// Fix it if the theme was the overall default theme.
```
Oldiesmann, line 516: Again no need to call updateSettings twice here - just update knownThemes at the same time as enableThemes and eliminate the "else" block.

Sources/Themes.php
```diff
@@ -205,25 +190,7 @@ function ThemeList()
...
-				'base_theme_dir' => 'base_theme_dir',
-				'base_theme_url' => 'base_theme_url',
-				'base_images_url' => 'base_images_url',
-			)
-		);
-		$themes = array();
-		while ($row = $smcFunc['db_fetch_assoc']($request))
-			$themes[$row['id_theme']][$row['variable']] = $row['value'];
-		$smcFunc['db_free_result']($request);
+		$themes = get_all_themes(false);
```
Oldiesmann: Since you're setting a context variable in get_all_themes and not returning anything now, you need to do `get_all_themes(false);` followed by `$themes = $context['themes'];`. Otherwise $themes will just get set to true (or false if the function fails).

Sources/Subs-Themes.php
```diff
@@ -0,0 +1,596 @@
...
+		$to_update = $smcFunc['db_fetch_assoc']($request);
+		$smcFunc['db_free_result']($request);
+
+		// Got something, lets figure it out what to do next.
+		if (!empty($to_update) && !empty($to_update['version']))
+			switch (compareVersions($context['to_install']['version'], $to_update['version']))
+			{
+				case 0: // This is exactly the same theme.
+				case -1: // The one being installed is older than the one already installed.
+					fatal_lang_error('package_get_error_theme_no_new_version', false, array($context['to_install']['version'], $to_update['version']));
```
Oldiesmann: Since the default case is the same as what happens for 0 and -1, you can move the "case 0:" and "case -1:" statements to right above "default:", then get rid of this line and the one below it.

Sources/Subs-Themes.php
```diff
@@ -0,0 +1,596 @@
...
+							'version' => 'version',
+							'id_theme' => $to_update['id_theme'],
+						)
+					);
+
+					// Done with the update, tell the user about it.
+					$context['to_install']['updated'] = true;
+					$context['to_install']['id'] = $to_update['id_theme'];
+
+					return $context['to_install'];
```
Oldiesmann: $context doesn't need to be returned here. Also, because you have `$context['to_install']['id'] = theme_install($context['to_install']);` elsewhere, the "return" here will cause $context['to_install']['id'] to contain a copy of the parent array in some situations. Your best bet here would be to either remove the return value below (setting $context['to_install']['id'] directly instead) or just set $context['to_install']['updated'] here and return the ID instead of setting $context['to_install']['id'] directly.

Oldiesmann:
That should do it :+1:. Now the real testing can begin :wink:

MissAllSunday:
Thanks!

Its gonna be a disaster :P :D 

ahh these github smilies are way different, gotta type tongue: tongue   so, :tongue:

PR #924, opened 2013-11-20
! More tweaks related to moving topics

Oldiesmann:
1. Don't display the "Move Topic" button if there aren't any other boards where you can move this topic
2. Don't display a "move" quick mod icon if there aren't any other boards where you can move this topic


Arantor:
And again, everything I said in https://github.com/SimpleMachines/SMF2.1/pull/921#issuecomment-28933100 applies here too.

PR #929, opened 2013-11-22
! Spellcheck enhancements

Oldiesmann:
This adds support for PHP's enchant extension by abstracting the spell check functions (spell_init, spell_check and spell_init functions) and also fixes the bug causing the Spell Check button to throw a JS error.


Oldiesmann:
Thanks for the heads-up. I'll look into that and make changes as necessary.

Oldiesmann:
According to the website for the Python interface for enchant, it works in UTF-8 only... http://pythonhosted.org/pyenchant/api/enchant.html

> Note that Enchant works in UTF-8 internally, so passing an ASCII string to a dictionary for a language requiring Unicode may result in UTF-8 strings being returned.

Should I just revert the bulk of that commit (basically everything except the button fix) or just force it to only use enchant if the charset is UTF-8? Obviously converting between character sets isn't a realistic option (we could do it, but it doesn't make any sense to do so, especially in this situation).

Arantor:
Yeah, I figured Enchant was probably UTF-8 only as there wasn't a way to pass a charset to it.

We could do the conversion between character sets, both ways in fact, if iconv is available and iconv isn't unrealistic an option at all. But it would likely be better to wait for us to go UTF-8 only first rather than trying to handle both.

pspell isn't going away just yet, it may be deprecated but that doesn't mean missing, just as ext/mysql is deprecated in PHP 5.5 but you'll still find plenty of hosts that have it around.

There is a much bigger question: do we need spell checking in the core? Is it actually worth it these days to have the support headache and whatnot for a feature that is actually - as far as I can tell - not that heavily used (especially because the browser can do it)?

Oldiesmann:
I would say at least leave it for 2.1, as there's no harm in removing it and some might complain if we do remove it (as of May, IE apparently didn't have it, as per http://www.simplemachines.org/community/index.php?topic=502943.0 though that might have changed now - I rarely use IE). We could always add it as an official mod, but the problem there is spell check currently works with every language SMF supports, and having a mod support 30+ languages could be a pain.

I'm also not sure I like the idea of converting words with iconv, even if it is reasonable, for the simple fact that it makes things even more confusing for people. If you have enchant enabled but not pspell, then you have to be UTF-8 (which I'm guessing a good portion of the SMF userbase doesn't even know about) or you need iconv (which may or may not be available - I'm not sure how common it is in terms of typical shared-hosting setups).

Arantor:
iconv is pretty common these days - and 2.1 defaults to UTF-8 on a new installation. So in those situations it's not so much of an issue, bearing in mind we would definitely need to strip the trailing .utf8 in the locale. Anyone coming from 2.0 or earlier could default to pspell if they had it enabled already.

Oldiesmann:
In that case I can modify it to use iconv if available. I actually can make it use pspell by default by switching things around slightly if you think it would be better to do that instead.

Arantor:
Well... it's not a case of 'better'. Anyone who uses spell check now will likely want it to continue working - and what we have does work if pspell etc. is properly configured. Having enchant would be nice, but it's a configuration we can't properly support because while we can tell users about pspell configuration... I can't even get enchant to work on WampServer and not for the want of trying.

iconv will handle the back and forth if available, quite literally iconv('ISO-8859-1', 'UTF-8', $text) to get it ready for enchant, iconv('UTF-8', 'ISO-8859-1//TRANSLIT', $text) going the other way. But making spell check dependent on both these two things by default is problematic.

If we keep spell check, pspell should be the default, followed by enchant if both enchant and iconv are available. There is a bigger question of whether we keep spell check or not. Honestly, I'm not that fussed about keeping it, not even particularly as a mod, though as a mod it's actually relatively easy to make happen because most of the translations we should need are already available.

Oldiesmann:
It's set up now to fall back on pspell if enchant isn't available or the appropriate dictionary can't be loaded, so we're not forcing either option at this point - it just uses enchant first if it's there, otherwise it'll use pspell.

I still think it would be best to leave spell check in for 2.1 and look at removing it / making a mod out of it for 3.0. IE didn't even add spell check until version 10, so it might still be useful to those using older versions of IE.

Arantor:
Works for me.

PR #952, opened 2013-11-30
Notify is no longer permission driven, but simply available to all. Keep it simple, folks!

Arantor:


AaronVanGeffen:
Hmm, can I ask what the rationale behind this change is? Looking at the changes, it is now effectively no longer possible to disable notifications altogether. Granted, this is a use case most forums won't face, but I can see some use for disabling notifications for certain user groups...

Arantor:
Why would you ever want to disable notifications for certain groups?

But there's a much larger overhaul of notifications coming and making each branch of notifications permission dependent is extremely problematic.

AaronVanGeffen:
If, for some reason, your forum cannot or should not send out any emails, for example. Sure, it won't happen often, but I used this once on an intranet that just didn't have access to a mail server. (Of course, users had to be added manually, too...)

Anyway, I was under the impression that 2.1 was feature frozen --which apparently it is not-- so I'll sit back and see what it is you're going to do with it.

Arantor:
Trouble is, we're still a good couple of years behind the competition - even with my breaking feature freeze to get this stuff implemented.

The thing is, I don't see how it is meaningful for us to have something that represents additional complexity when it suits only a (very small) fraction of the userbase. I'd rather provide something specific for them - and only them - if demand warrants it.

Don't forget there are other types of notification SMF already sends out e.g. new registration notifications, the notifications for pending approvals... just having that permission only deals with a user facing thing, not the other stuff.

PR #963, opened 2013-11-30
No need to center it...

XinYenFon:


NanoSector:
What's "it"?

XinYenFon:
Enable & Fill all profile fields and go to your profile you'll see website & email icons are lower than others.

PR #982, opened 2013-12-05
Update index.css

Hristo-:
Boards on Ignore Boards Options page should be in 2 columns (.floatleft and .floatright), but due to 90% that was not possible.

P. S. Registered here a few hours ago, have no idea if this is the right way to do it :/

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -3292,7 +3292,7 @@ img.centericon {
 .ignoreboards {
 	margin: 0 2%;
 	padding: 0;
-	width: 90%;
+	width: 45%;
 }
 .ignoreboards a {
 	font-weight: bold;


MissAllSunday:
Hi @Hristo- thank you for your interest and the PR :)

Yes, git and/or github can be somehow complicated to use, on the repo main page there are some instructions on how to branch and submit pull requests (PR) https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/README.md

Just be sure to sign off all your commits otherwise we won't be able to use your code.

You can delete your branch both form your local repo and your own fork repo, doing that will close this PR

Hristo-:
Thank you for the clarification. I deleted the repo a few hours ago and the PR is still open though.

I think I figured out how to proceed, but I have one question regarding the sigh-off - is it enough if I include in each PR's description the next text:

Signed-off-by: Hristo- "my email at github"

P. S. The "How to create Git aliases" link inside Readme is not working.

MissAllSunday:
Oh yeah, pretty weird it wasn't deleted when you deleted your branch.

Well, it would be preferred if you sign off every commit you made, that way we can be absolutely sure you are in agreement with our Developer's certificate of origin.

Most of the UIs for git does have a way for making the sign off pretty easy, I use the default windows UI and it has a handy button for it or if you prefer to use the command line thats where aliases comes in handy too.

Thanks for reporting the broken link, I'll update it with a link to git's wiki instead: https://git.wiki.kernel.org/index.php/Aliases

OK then I'm gonna close this PR and we're looking forward to see your contributions :smile:

PR #984, opened 2013-12-05
.ignoreboards width fix

Hristo-:
Boards on Ignore Boards Options page should be in 2 columns (.floatleft
and .floatright), but due to 90% width  that was not possible.


Arantor:
Thanks :)

Issue #944, opened 2013-11-28
minor: Password flood error generates multiple undefined/general PHP errors 


HoverClub:
If the password flood check fails (load.php, line 310) the resulting fatal_lang_error tries to load the theme which relies on the permissions being available - they haven't been loaded yet so multiple PHP errors are generated as the code loads templates, etc..  

Backtrace:
smf_main->loadUserSettings->validatePasswordFlood (fails)->fatal_lang_error->LoadTheme->isAllowed['admin_forum'] 

A possible fix might be to return false for any permission if the permissions are not available (security.php line 921 - allowedTo, etc.    if (empty($user_info) OR empty($user_info['permissions'])) return false; ?

Example of the issue:
http://www.simplemachines.org/community/index.php?topic=513216.msg3627169#msg3627169

HoverClub:
I've also just found other errors from $user_info itself not being populated - all through the failed flood check as above.  

Could  $user_info be populated with the default guest values at the start of loadUserSettings (then replaced by any "real" member data) or is that likely to cause other issues?

Oldiesmann:
I think returning false if $user_info or $user_info['permissions'] are empty should be enough. There's no point in populating an array with a bunch of dummy data just to avoid PHP errors when there are easier ways around it.

emanuele45:
In loadTheme $user_info is already populated as far as I can see (loadUserSettings is called before loadTheme).

And if $user_info contains data, but 'permissions' is not an array, it means there is something wrong somewhere else, because loadUserSettings initialize 'permissions' as an empty array.

HoverClub:
If the flood test (during loadUserSettings) fails then it bails out to fatal_lang_error BEFORE loadUserSettings has populated the $user_info array.  It then spews out around 70 PHP errors/warnings every time a flood check fails - that's a pain for admins looking at logs and trying to work out what's gone wrong.  

A lot of the system won't work properly if $user_info isn't valid so surely it should be populated as early as possible (and certainly, as in this case, before any function that can bypass the normal load sequence, is called)?

emanuele45:
Then, if $user_info is empty and allowedTo returns false.

HoverClub:
It's not enough only fixing permission checks  ..... template_header->setupThemeContext requires $user_info[groups], $user_info[is_guest], $user_info[language], etc., etc.  All these irrelevant errors are generated because a failed flood check bypasses the normal load sequence.

emanuele45:
[...]

PR #985, opened 2013-12-05
Fixes

MissAllSunday:
OK, this is yet another attempt for handling attachments extensions, it is meant to be done on upgrading and it should cover the old attachment system, meaning there shouldn't be any need for legacy code.

It also fixes some other small issues.


Arantor:
Looks good, though will need some serious testing. Before we merge it, what's the state of #944 being fixed here?

MissAllSunday:
> Looks good, though will need some serious testing. Before we merge it,  
> what's the state of #944 being fixed here?
> 
> —
> Well, the fix didn't work that well, it caused a bunch or errors, all  
> related to $user_info been treated as not empty so I just reverted it back  
> to check $user_info. We can re-open the issue if there is more info about  
> it.

PR #988, opened 2013-12-06
Token admin-tm could not be verified.

live627:
This just moves code.


Oldiesmann:
You really just need to move the `createToken('admin-tm');` statement, not that entire block of code, though keeping all createToken statements together makes the code more readable, so I won't complain if you want to move all of them.

live627:
I know.

Moving all the code makes sense to me. None of the template code is needed upon save.

PR #992, opened 2013-12-06
The upgrader has fallen behind in regards to themes

live627:
Can we delete the 1.0 file already? It was sunset... a year ago.

2 0 pgsql and sqlite files are so different I'm lost on them.

### DIFF
--- a/other/upgrade_1-1.sql
+++ b/other/upgrade_1-1.sql
@@ -368,7 +368,7 @@ if ((!isset($modSettings['smfVersion']) || $modSettings['smfVersion'] <= '1.1 RC
 ---}
 ---#
 
----# Updating the knownThemes setting.
+---# Insert the knownThemes setting.
 ---{
 $request = upgrade_query("
 	SELECT id_theme
@@ -384,6 +384,7 @@ if (!empty($inserts))
 		INSERT INTO {$db_prefix}settings
 			(variable, value)
 		VALUES
+			('knownThemes', '" . implode(',', $inserts) . "')");
 			('enableThemes', '" . implode(',', $inserts) . "')");
 ---}
 ---#
--- a/other/upgrade_1-1.sql
+++ b/other/upgrade_1-1.sql
@@ -368,6 +368,26 @@ if ((!isset($modSettings['smfVersion']) || $modSettings['smfVersion'] <= '1.1 RC
 ---}
 ---#
 
+---# Updating the knownThemes setting.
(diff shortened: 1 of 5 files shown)


Oldiesmann:
No, we can't get rid of the 1.0 file. Doing so would mean anyone still using 1.0 who wants to upgrade would have to download 1.1 or 2.0 first, upgrade to that, then download 2.1 and repeat the process. Granted, there likely aren't many people still using 1.0, but as long as we can support upgrading from it, we should. This might change for 3.0 though.

For this change though, you only need to touch the upgrade_2-1 files. There's no need to modify the others.

Oldiesmann:
Looking at this again, I'm actually not sure this is necessary. Setting SMF so that only the default theme is enabled upon upgrade really isn't such a bad idea, because 99% of the people running upgrade.php will be coming from an earlier version of SMF (not simply upgrading between minor releases, such as from 2.1 to 2.1.1), and there's really no reason why we should assume that all themes are now compatible with 2.1. It's a tough call - do we just assume users have already upgraded their themes, or do we just set it so that only the default is enabled and leave it up to the users to re-enable any others that they might want?

Either way, if we do make this change, my point above still stands - we only need to change the upgrade_2-1 SQL files.

Arantor:
Given that there are non trivial changes to the theme, I'd be inclined to declare them all incompatible regardless... in fact I'm pretty sure they will be incompatible as far as the theme code is now concerned as theme-info.xml files won't have the version in them.

MissAllSunday:
Yep, when making the upgrade your current installed theme will be used but it will have some glitches here and there, granted, not as bad as the conversion from 1.1.x to 2.0 but its still pretty much unusable.

I vote for making curve2 set as the default theme when upgrading, the issue here is that the theme they were using will still be marked as installed and the user can set it as the default theme again.

I suppose we can set those themes as disable and let them know that this particular theme won't work on 2.1.

Oldiesmann:
The current upgrade sets "enableThemes" to 1, which is part of the process. We just need to finish the rest of the necessary code changes to make it so that users can't select themes that aren't enabled. I do agree we should just force the new default upon upgrade, and let admins sort out the rest later. We might get some complaints, but it's better than people whining that their theme is broken.

Oldiesmann:
#1002 and #1003 finish up the themes stuff, so I think this can be closed now.

live627:
k then I'll close this

PR #1004, opened 2013-12-07
! PostgreSQL fixes

Oldiesmann:

### DIFF
--- a/other/upgrade_2-1_postgresql.sql
+++ b/other/upgrade_2-1_postgresql.sql
@@ -1,4 +1,4 @@
-/* ATTENTION: You don't need to run or use this file!  The upgrade.php script does everything for you! */
+/* ATTENTION: You don't need to run or use this file! The upgrade.php script does everything for you! */
 
 /******************************************************************************/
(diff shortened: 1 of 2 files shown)


other/upgrade_2-1_postgresql.sql
```diff
@@ -213,7 +213,7 @@ upgrade_query("
 	ALTER COLUMN session SET DEFAULT '';
 
 	ALTER TABLE {$db_prefix}log_errors
-	ALTER COLUMN session SET default '                                                                ';");
+	ALTER COLUMN session SET default '                                ';");
```
emanuele45, line 216: Why?
It's a char(64), why remove 32 of the 64 white spaces?

Oldiesmann, line 216: Sorry. That was a mistake on my part. I removed "IF NOT EXISTS" but my editor left a space in its place, so I was replacing double spaces with a single space. I thought I undid that when it came up saying 80-something instances were replaced, but I guess it didn't completely fix it. I'll fix it shortly.

PR #1009, opened 2013-12-08
Go Up Buttons

Hristo-:
I think they need a fix

### DIFF
--- a/Themes/default/Display.template.php
+++ b/Themes/default/Display.template.php
@@ -701,7 +701,7 @@ function template_main()
 	echo '
 			<div class="pagesection">
 				', template_button_strip($context['normal_buttons'], 'right'), '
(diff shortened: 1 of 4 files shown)


Arantor:
I agree with this change as it reverts the go up/down buttons back to how they worked in 2.0 (i.e. go up/down to the top/bottom of primary content), but each individual commit should really be signed off as opposed to the PR as a whole...

Hristo-:
I see, we have to Sign-off the changes of each file even if they are part of one PR?

Then, should I close this PR and submit a separate PR for each file, or I can use the option to comment some line and add a sign that way?

XinYenFon:
you don't have to actually open PR for each of your changes, do your changes push to your repo (don't forget to check sign-off)
- SourceTree has little sign-off checkbox
- GitGui has a button to write sign-off.

PR #1013, opened 2013-12-10
+ Add alerts/notification preferences for group requests

Oldiesmann:
Initial functionality for handling group requests with the new setup in 2.1. The actual background task to handle the notification and alert will be added once that system is functional.

### DIFF
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -1885,6 +1885,7 @@ function alert_configuration($memID)
 			'msg_report_reply' => array('alert' => 'yes', 'email' => 'yes', 'permission' => array('name' => 'moderate_board', 'is_board' => true)),
 		),
 		'members' => array(
+			'request_group' => array('alert' => 'yes', 'email' => 'yes'),
 			'register_new' => array('alert' => 'yes', 'email' => 'yes'),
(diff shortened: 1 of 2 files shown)


Arantor:
The entire point of the background tasks system has been neutralised in this. You're not supposed to do anything regarding fetching the list of possible users, or getting preferences at all in the foreground. The foreground's deal is literally to get the bare minimum possible and fire off the background task and let the background task deal with everything else.

If we're going to do it all frontside, 1) there's no point having the background tasks and 2) you end up slowing down all user-responsive tasks by doing all that stuff every time it applies - before the user gets feedback. That's why you push it to the background.

I was kind of hoping you'd actually wait for an example to build off, rather than trying to second guess how it's supposed to work, and now I'm going to end up rewriting it to make it work that way. And damn, this also causes fun and exciting edit conflicts.

Oldiesmann:
I can rewrite it to make it work the proper way as needed. I assumed from talking to you in IRC that the background task simply handled the notifications and alerts, not the whole "pull the data for who should be notified" thing as well.

Arantor:
I don't recall giving that impression :confused: The idea is to delegate, to do as little work as possible so we can get back to the user sooner, and let all the real work happen in the background.

I'm in the middle of adding one example once it's finished testing and that will hopefully give a better idea of how it's supposed to work.

PR #1024, opened 2013-12-11
Release 2.1

Xarcell:
I have added a exclusive & custom default avatar that I designed in illustrator to use with SMF2.1 & it's forks.


Hristo-:
I definitely like it and if my vote counts I would like to see it as part of 2.1
And to save some bytes here is a lossless compression of it:
default-avatar-optimized

I see you say "default". Does that mean default avatar for members who does not use/pick their own? I use AvatarSelect to force the users to choose between a male or a female silhouette avatar. But if I have to use just one avatar I would choose something inhuman (do I use it right?!), like big question mark or something similar to what skype use(ed). As I said I like this one, but IMO it's not gender/age/race-less. You may laugh, but when I saw it my first though was something like "that is a Korean boy" :)... now I do not think like that, but my first impression was in those lines.

Arantor:
I like it - but I think we do need to have a generic genderless icon as well as male/female, so those who choose not to identify themselves with either gender as well as both genders currently supported will be catered for - I can imagine that some female users would not exactly appreciate a male-looking avatar by default.

Hristo-:
When I looked for good default female and male avatars the best ones I found were these:
http://brandonmathis.com/projects/fancy-avatars/demo/

Xarcell:
I am working on other male/female icons. A whole set actually. Just trying to work out the facial features as it isn't easy to do.

Oldiesmann:
Interesting find Hristo. Those might be usable since they're under an open source license (we're free to do what we want with them as long as we give credit to the author and provide a link to the license). I'd like to see what Xarcell can come up with first before we use those though.

Xarcell:
hrmmmm... being new to Github and all, I did not mean to combine the default avatar and index.php issue.

Xarcell:
When I say working on a whole set, I mean non-silhouettes(although I'm struggling getting the facial features right). Is the request here to have a default genderless silhouette avatar, as well as a default male & female silhouette avatar?

Oldiesmann:
The request is indeed to have gender-specific default avatars in addition to the genderless one.

As far as the other commit goes - you need to use branches if you want to work on multiple unrelated things at one time. Since you committed the index.php thing to the same branch as the avatar thing, it got added to this pull request.

PR #1049, opened 2013-12-14
We don't use latest-themes any more, let's stop having it.

Arantor:


Xarcell:
I'm glad this was removed.

PR #1055, opened 2013-12-15
! Add default value for "member_group_request" alert (and the other two ...

Oldiesmann:
...as well in Postgres)


Arantor:
I knew I'd put them in for PostgreSQL... but I put them in the wrong place >_< I'll clean that up in a minute.

PR #1081, opened 2013-12-19
Board descriptions should really be shown inside boards, not just because a user chooses so.

Arantor:


Xarcell:
Some idiots apply long descriptions to their boards(like 3-4 paragraphs). This option helped reduced scrolling to get to the important content. Although, I'm not saying I'm against the changes, I'm actually for them. However, I can see why the option was added to begin with.

Arantor:
Except the board descriptions will _always_ be shown on the board index too anyway, and that might sort of suggest that they don't.

PR #1084, opened 2013-12-19
Master

brayancruces:
Copyright


Arantor:
Firstly, the master branch is not where development is happening. I'm also not entirely sure what you're trying to do - this was already previously merged into SMF and subsequently reverted because of issues caused. As you will note, the PR has a variety of conflicts that make it unable to be easily merged.

PR #1097, opened 2013-12-21
Moving the calendar CSS to its own file

live627:


Xarcell:
Why create more CSS files, you don't need to.

Oldiesmann:
It decreases the size of index.css and prevents loading unnecessary CSS on every page (why do we need to load the calendar CSS on every page if we only use it in the calendar?).

Xarcell:
Why have the extra HTTP request, and additional files that a themer has to tamper with? Splitting up php files I understand, but the CSS file I don't as cached by the browser anyway. Most mobile browsers have a bottleneck and only download 2 files at a time. The more HTTP requests you have, the higher your bounce rate will be.

Even the admin.css being shouldn't be separated, but is acceptable because it's use is very limited.

As long as the CSS file in question doesn't exceed 100k or 4096 selectors, there is no beneficial reason to start splitting it up. If done properly, the calendar CSS should be no more than 8-12 lines of code anyway.

Arantor:
Because index CSS is already close to 100KB, the calendar is not enabled by default and even if it were, the CSS really doesn't need to be loaded every page because even on sites that use the calendar, it's still a fraction of the page views in most cases of everything else.

And if you look, it's significantly more than 8-12 lines of code being moved.

Xarcell:
When I said 8-12 lines, I meant there shouldn't be, not that there is. I'll have to go back and check in my fork how many lines I was able to reduce it too, without changing the look of it.

matthew-kerle:
I definitely agree with starting to split up the CSS for actions like Calendar that won't always be present and have a lot of CSS. It makes things more modular and organized as well. There is definitely a lot of Calendar CSS.

XinYenFon:
Also this needs some love ^^ #1377 coz of :paw_prints:

PR #1109, opened 2013-12-22
Release 2.1

Xarcell:
I added a default avatar image to the theme directory. No code added or altered. I also reworked the favicon to be white fulcrum, on SMF blue background, with a 2px border radius. IMO, it looks alot smoothier than the current one and not so rough looking.


Oldiesmann:
The avatar looks nice, but I'd prefer one that actually looks like a person (genderless and gender-specific) rather than our logo.

Xarcell:
Lol, that makes me want to shoot somebody. I had a person, then it was requested that it be changed to a symbol. I made a ethic-less/gender-less person, but then style & age became an issue for some. So then it was requested that it be turned into a symbol. Which I did a few of. Then it was requested that it should be the SMF logo. So that's what I used.

Some kind of outline or agreement among the powers that be would be nice to have. This way I wouldn't have to spend time creating so many trying to make everyone happy. If the team can decide indefinitely what they want, I will be happy to make it.

Arantor:
Actually, it wasn't a 'request' but a possible suggestion, out of curiosity to see how it might look.

Xarcell:
I understand, but I think a decision among the team should be made so that the work doesn't become counter-productive.

Xarcell:
Has a decision been made on this?

PR #1135, opened 2014-01-01
Avatar stuff

MissAllSunday:
- Removes the custom_avatar_enabled setting, making it the default and only behavior.
- Handles moving old attachment avs on upgrade
- Sets some safety checks for custom_avatar and server stored dir
- Left Sources/Avatar.php for legacy/for some reason the upgrade file failed miserably.

Hope this covers everything its suppose to cover :tongue:


Arantor:
One thing before I test this... while at a glance it looks good, does it safely handle upgrading a site that already has a custom avatar folder? (And if so, does it leave it in place or does it move things?)

Also, on that note, does the installer also check that the folder is writable?

Arantor:
Also, while I think about it, what happens if when upgrading, the folder isn't writable? Does it stop and demand the user make that folder writable before continuing?

MissAllSunday:
In essence, if the custom dir is already set then the upgrade use that dir to move the avatars to it, if not, it sets the default one. The upgrade doesn't touch any avatar that was already moved, well, the upgrade assumes that all already moved avatars have 1 as the attachment type.

The installer doesn't  check if the custom_avatar dir is writable nope, only the upgrader.

I did thought about what happens if the custom dir cannot be writable, I didn't add something to tell the upgrader to stop, I probably should :tongue:

Arantor:
Cool, sounds good to me.

The installer is easy, there's already a list of folders that need to be checked for being writable, it's no huge deal to add another to that. The upgrader might check, can't remember if it does or not (should be checking at least for Settings.php to be writable)

Sources/ManageAttachments.php
```diff
@@ -330,9 +329,18 @@ function ManageAvatarSettings($return_config = false)
...
+			$_POST['custom_avatar_dir'] = $boarddir .'/custom_avatar';
+
+		if (empty($_POST['custom_avatar_url']))
+			$_POST['custom_avatar_url'] = $boardurl .'/custom_avatar';
+
+		if (empty($_POST['avatar_directory']))
+			$_POST['avatar_directory'] = $boarddir .'/avatars';
+
+		if (empty($_POST['avatar_url']))
+			$_POST['avatar_url'] = $boardurl .'/avatars';
```
jdarwood007, line 342: I don't really like modifying the POST vars, but my knowledge says that is the only way to do it with SMF currently.

Arantor, line 342: Yup, because when it gets fed into saveDBSettings it will look for $_POST values.

Arantor:
Bah I broke it :(

MissAllSunday:
It's been a while since the last time I ranted about the command line :tongue:  

I'll take care of it. :+1:

Arantor:
Thanks :)

MissAllSunday:
My goodness! after a lot of failed commits I finally was able to resolve the merge conflicts Ha! looks like I also committed some recent changes but I finally did it!

github instructions are fine but they are designed for me to hardcode the merge directly into the main release21 branch which obviously I ain't going to do so I merge it with my own repo's release2.1 and merge that repo with this one.

All of this just to remove a ternary :tongue:

Oldiesmann:
Hate to break it to you, but it looks like something broke it again. It's still showing merge conflicts for me :frowning:

MissAllSunday:
OK done, it was  a silly typo on my part as well as the removal of / in self closing rags for HTML5

PR #1140, opened 2014-01-03
Disappearing text on folder permissions fix.

ghost:
Potential fix for disappearing text on file permissions page, ref: 0004866 old bug tracker.


Oldiesmann:
This doesn't actually fix anything - it just changes the output of the code so it's on the same line as the previous line of HTML code, rather than being on its own line.

It's weird that this is only happening for the Smileys directory though - if you click on any other sub-directories that contain images, it simply prevents you from clicking on that subdirectory again.

ghost:
This fix worked locally for me, it seems to only happen on second level directories that have no sub-directories. The issue is happening in `Packages.template.php` at line 1344-1348, when it's moving the childNode before the link it will only move the image tag and not pick up the text `$name`, removing the white space before the image tag solved the issue on my local testing.

There are two functions looking after the generation of the file tree, starting on line 1300 which works fine and on line 1627 which is where the directories with the issues are being generated.

It happens on any second level directories if they have no sub-folder, if you add in a new folder to Themes and add just files into it you will get the same result as the Smiley sub-directories.

I had to clear out my browser cache for the change to be picked up.

ghost:
From further investigation it looks like all second level directories that have no sub-folders for example Smileys -> aaron are being picked up like so `[object Text][object HTMLImageElement]` with the preceding text object being either empty or just whitespace where as any other directories with no sub-folders for example Themes -> default -> css are being picked up as `[object HTMLImageElement][object Text]`.

This workaround will fix the issue grand but for a permanent fix it needs to be looked at how the `$context['file_tree']` is defined in `Packages.php`on line 1981, if you specify each sub-folder with a 'list_contents' => true, it will fix the issue. The problem is that this requires all sub-directories to be defined in the file_tree array for Smilies and Themes at minimum, two options come to mind:
1) Set a default global value for 'list_contents' set to true
2) Add a new array element to the file_tree array when a new smiley pack or theme is installed.

Open to the floor for suggestions as to the best way to tackle this issue.

Oldiesmann:
I'll go ahead and pull this in for now until we can find a more permanent/better fix. It's a pretty complicated setup, especially for something that's so rarely used.

PR #1152, opened 2014-01-07
! The beginnings of a conversion to HTML5. About time we did this, reall...

Arantor:
...y.


matthew-kerle:
Not sure why you just didn't fix the conflict issue with my Pull Request because it did all of this and WAY more. Fixed a ton of inline CSS. Added a backwards compatibility doctype and script for older browsers. And fixed hundreds of instances of invalid HTML5.

Arantor:
I have my reasons. Let's just leave it at that.

AaronVanGeffen:
@Xarcell Looking at Arantor's changes, IE8 will render these pages just fine. You'll only need the 'HTML5 shim' JavaScript bit if you're using elements that are new in HTML5 (e.g. nav, section, header), which is not (yet) the case here.

Arantor:
There are some things that need fixing for validation purposes but I wasn't planning on a complete tag conversion just yet.

PR #1160, opened 2014-01-09
Gravatar support

ghost:
Rough addition of Gravatar support for profile avatars.


Xarcell:
Is there an option to turn this off? I'm not fond of gravatar because it drastically slows page load in forums.

Oldiesmann:
Yes of course it will be optional. However, it does not slow things down nearly as much as you seem to think it does. This site uses gravatar as do several others I'm a member of and there is no noticeable reduction in the speed of things.

Arantor:
Of course this could also solve the default avatar debate _innocent whistle_

Xarcell:
@Oldiesmann I know Xenforo came out with Gravatar support, then they turned it into an option because of the slow loading. Gravatar just didn't have the power back then. Maybe this has changed and their servers can handle it.

@Arantor I reckon the default avatar issues has been brought to a point of causing too much stress I reckon. I don't want to push things any further. Gravatar does provide a default avatar, but what if Gravatar is disabled?

Arantor:
Not just XenForo - IPB has it too. In fact, in IPB it's flat out by default, to the point where it actually is broken on localhost installations (because they use the feature of Gravatar to push default avatars to Gravatar)

Do note that you're talking about a 3 year timespan... lots has changed since then.

PhuriousGeorge:
Are we talking about making use of Gravatar by default (option is on in the admin panel ofc)? Or making it select able at registration?

Arantor:
We're definitely looking at including it as an avatar choice. Choice at registration... not so much. Default avatar... it's an option but Xarcell makes a good point about not doing that. So I suspect we'll push it to being a choice (like gallery or upload is a choice) and then see how people find it from there.

Xarcell:
If your set on adding Gravatar, then would it even be necessary to even allow upload-able avatars? Seems to me that Gravatar replaces this function somewhat, but then again I dunno.

After taking a quick read at Gravatar, it looks as though you can still define your own default avatar, but you could still even use the avatar from a theme's folder or default theme.

Arantor:
Sure it would. Gravatar is one of those useful things to have - but I don't often remember to change my Gravatar even when I might well want a different one - and of course if Gravatar is disabled then you still need uploadable.

The whole 'default avatar' thing is a bit iffy when trying to push to Gravatar (so we'd have to probably keep it separate)... I haven't checked the very most recent version of IPB, but up to 3.4.5 at least, if you run it on a localhost, non-uploaded avatars are flat out broken because the request URL to Gravatar includes the URL to the default avatar, so that Gravatar can download it for you - but of course, Gravatar can't load a URL from http://localhost/ and so it fails.

PhuriousGeorge:
Personally I'd rather still have the option for uploaded avatars instead of relying on Gravatar and people to be forced to go there to gain individuality

Arantor:
There is precisely zero chance of removing uploadable avatars.

XinYenFon:
So we are not going to put a little check box on registration for Gravatar about "set my avatar as Gravatar"

Arantor:
I'm not exactly enthusiastic about that. Convince me :tongue:

Xarcell:
IMO, I see no reason to add that on registration if we are offering a default avatar. Most people go to their profile to make changes anyway for other things. I would leave only necessities for registration, because the more fields you add/require the more it will increase your bounce rate. I read an article once that once you hit about 7 fields required to fill out on registration, people start to leave before finishing the process.

Oldiesmann:
We could have an option for "automatically use Gravatar if possible" so users' avatars would automatically be set to their Gravatar if it exists, which is what many sites seem to do now (and yes, I said "option" - something that the admin would have to enable with the rest of the Gravatar features, since not everyone would want this even if Gravatar is enabled).

PhuriousGeorge:
If the current stance is to apply a default avatar to users upon registering, I think @Oldiesmann 's suggestion makes the most sense.

ghost:
So are we adding the option on registration to say 'if a gravatar exists' use it, only if the membergroup allows the gravatar option?

Oldiesmann:
If we add Gravatar support, it would be nice to have an "automatically use the member's Gravatar if it exists" option as well for new members. Permissions here would be a bit tricky though since there would be no point in enabling that option if new members can't have Gravatars to begin with. Maybe we could just ignore permissions in that one situation?

Xarcell:
I'm not sure about that. Some user's might not want to use their gravatar, and this option would force them to use it. I have registered at sites to lurk before commiting to a community, and would rather not use my gravatar unless I'm commited. So I see how this can benefit the admin, but will it benefit all user's?

XinYenFon:
may i ask what is missing or needed here? :)

MissAllSunday:
Theres a merge conflict now and I'm still unsure what are we gonna do.

ghost:
What's the merge conflict and I'll have a look at it?

XinYenFon:
conflict

Dragooon:
From what I can see it requires specifying the Gravatar's hash right? IMO we should change that to allow specifying gravatar's email and simply calculate the hash ourselves.

Arantor:
That sort of depends on whether you want to allow the user to specify a separate Gravatar address or reuse their normal account email address... I know in That Which We Do Not Speak Of, I stored it in the avatar field as gravatar:// for normal (mirroring using http:// for URL avatars) and gravatar://email@domain.com for secondary Gravatar email.

Dragooon:
Currently it asks for their gravatar URL, IMO we should change that field and simply allow filling their separate gravatar address and fill it with their default e-mail by default.

Arantor:
What happens if a user changes their email address?

Dragooon:
I guess of its same as gravatar email then change it too otherwise leave
that alone?

Arantor:
That's kind of why I thought about not storing it as part of the Gravatar so it would pull it from the user's account to avoid data duplication.

Dragooon:
But that wouldn't allow user's to specify their own e-mails instead if they use something else. Hash is less user friendly to be honest

Arantor:
Sure it would if it was done as I suggested ;) You reuse the avatar column in the database, just use the word gravatar on its own or similar to indicate 'generate hash out of user's email' or gravatar:user@domain to indicate 'generate hash out of that email'

live627:
Is this going in or will it be rejected?

Dragooon:
This still has a merge conflict and I'd like to e able to enter e-mail instead of hash

live627:
Mind if I finish this up then?

Dragooon:
Sure, would appreciate it.

live627:
Aye aye, skipper! :P

Dragooon:
Closing because of #1886

PR #1162, opened 2014-01-09
More old stuff to go buh-bye

Arantor:


Xarcell:
Glad to see this gone.

PR #1165, opened 2014-01-10
! Quick reply might be permanently on, but there's no need to show it if...

Arantor:
... the user doesn't have permission to see it.


Xarcell:
THANK YOU...

PR #1193, opened 2014-01-15
Fix to stop javascript error being thrown in the console.

ghost:
Small change to stop an error being thrown in the browser console when the default message icon is selected from a reply via an ajax call.

### DIFF
--- a/Themes/default/scripts/script.js
+++ b/Themes/default/scripts/script.js
@@ -593,7 +593,8 @@ function submitThisOnce(oControl)
 // Deprecated, as innerHTML is supported everywhere.
 function setInnerHTML(oElement, sToValue)
 {
-	oElement.innerHTML = sToValue;
+	if ('innerHTML' in oElement)
+		oElement.innerHTML = sToValue;
 }
 
 function getInnerHTML(oElement)


matthew-kerle:
My guess without looking at the code more than those couple lines is that it throws an error to the console if the element doesn't exist. You obviously can't use the innerHTML method on an element that doesn't exist.

So instead of checking to see if innerHTML is a callable method in the element, why not check to see if the element that's being inputted as an argument ACTUALLY exists. If it does, innerHTML will be there. If it doesn't, innerHTML won't be there.

When it comes down to it, your fix is doing the same thing, but checking to see if the element exists is probably a better idea, it's also on a higher level than a method of the element.

Arantor:
Or how about actually studying the code in question before delivering a sermon on how everyone is doing it wrong.

In this case neither the suggested fix nor the suggested course of action are actually correct.

matthew-kerle:
What's up with being rude to people who are trying to volunteer their time and services to help? And for the record, I never said he was doing it wrong. I said it should do the same thing.

Arantor:
I have a problem with people who come along and tell us how things 'should' work but not actually want to invest in _doing_ it. In this case, for example, checking that the object being passed in exists before checking that innerHTML is usable... the stated code actually does that already as a by-product of the test it's doing, and so no changes are actually required.

I also have issues with 'I haven't checked the code but it's probably ...' when in the time taken to reply, you could have checked the behaviour described with a modern browser which would have given you the backtrace, and told you exactly what the problem is (a div that doesn't exist in the display template in a lot of cases)

I just don't appreciate having my time wasted is all. It's not like I don't have hundreds of bugs still to investigate and fix, some of them stemming back years in SMF, like yesterday fixing the calendar that had been broken for 18 months.

emanuele45:
Considering that innerHMTL as stated in the comment above the function is supported everywhere and the error I'm getting is

```
TypeError: oElement is null
```

I feel like @Labradoodle-360 is right... without even looking at the code. :stuck_out_tongue:

Arantor:
Well... if the element doesn't exist, that's fine for the stated change because the if should return false.

Oh look, what's that? A backtrace you say? script.js, line 1318.

```
setInnerHTML(document.getElementById('modified_' + this.iCurMessageId), oMessage.getElementsByTagName('modified')[0].childNodes[0].nodeValue);
```

Yes, it's a missing element. We all figured that out. Basically Lab was showing off his knowledge of how we should do things by saying that we should test it exists before testing whether it has .innerHTML. Sort of like saying we could test isset() before testing !empty() on something specific, when the second one will cover it.

Or we could, as I suggest, fix the actual problem: not have a missing element being passed to setInnerHTML in the first place. Because fixing the root cause is so much better than fixing the surface problems.

PR #1195, opened 2014-01-15
Removed unnecessary parameter

ghost:
Removed unnecessary parameter, where default value 'post' was being passed into checksession().


Arantor:
Code cleanliness is good :)

Issue #1215, opened 2014-01-17
Void Elements - Removing ending slash


ghost:
Topic to discuss removing the unnecessary ending slash on void elements as 2.1 is going to be HTML5 compliment. There are between 2000 and 3000 of these instances, it wont have any real performance impact but it will remove between 2000 - 3000 unnecessary characters from the codebase.

Arantor:
Better than that, it's actually 4000-6000 unnecessary characters since convention is to put a space in front of them!

Though, what's to discuss? It's going to get done :tongue:

ghost:
Great, didn't know it was on the list :+1:

Arantor:
It's on my mental list but a reminder is always good. :)

Oldiesmann:
This should be good now, but I'll leave it open in case any issues with my PR or this thing in general pop up.

ghost:
Haven't come across any issues to date since this PR was merged.

Arantor:
We've seen a few but honestly at this point we should be closing the issue and resolving any as we come across them.

PR #1216, opened 2014-01-17
The ' />' method of closing so-called 'void' tags is no longer required ...

Oldiesmann:
...in HTML5, per #1215

I believe this covers everything. I basically did a find/replace on all SMF files with phpDesigner, then undid changes to a few places where we don't want the changes to be made (Wireless template and anything dealing with XML pretty much).


Arantor:
There is a very good reason we do not do find/replace automatically for this... there are now guaranteed to be fringe bugs attached to this.

The preparse step converts line ending characters to br tags. All pre-existing posts are going to have XMLish tags in them, new ones may or may not, so we can't just ignore them. And we have to be careful with nl2br() calls for the same reason.

ghost:
I started this change for sources and there are at min 10-15 instances where a find and replace will cause issues.

Arantor:
We actually discussed this in IRC, the principle issues are around posting, with preparsecode and unpreparsecode, and no issues were found because unpreparsecode is capable of handling both styles. (This change has been tested.)

PR #1220, opened 2014-01-18
Don't tether smc_PopupMenu to #top_info

live627:
If #top_info is not found within the popup's parents, it closes immediately after it opens. This makes it more dynamic.


Arantor:
Planning on using this in your own mods, eh?

live627:
On a sidenote, it is so nice to write without worrying about bytes.

PR #1221, opened 2014-01-18
Email address

### LINKED ISSUE #1191: User changing email address issues
## *\* Old Bug Report **

Still present in 2.1 alpha, tested and confirmed.

1) When `Require reactivation after email change` is enabled and a user changes their email address they are logged out, it isn't entirely clear that when a user changes their email that they will be logged out.

2) After the user changes their email and are logged out the 'Modify Profile' and 'Actions' options are still visible.

MissAllSunday:
Fixes #1191
- Adds an alert box when someone changes their email and $modSettings['send_validation_onChange'] is set.
- Moves out the validation message to its own action.

The new text string needs better wording:

$txt['email_change_logout'] = 'You decided to change your email. You will be logged out. Please reactivate your account.';

And the page title probably shouldn't be $txt['profile']

Any suggestions?


live627:
Since you decided to change your email, you will need to reactivate your account. You will now be logged out.

Sources/Profile-Modify.php
```diff
@@ -181,6 +181,12 @@ function $preload:		A function that is used to load data required for this eleme
 			'subtext' => $txt['valid_email'],
 			'log_change' => true,
 			'permission' => 'profile_password',
+			'js_submit' => !empty($modSettings['send_validation_onChange']) ? '
+				if (document.forms.creator.email_address.value != "'. $cur_profile['email_address'] .'")
+				{
+					alert("'. $txt['email_change_logout'] .'");
```
live627: You should escape the string with JSE..

matthew-kerle: Also, if we have jQuery at our disposal, why mix in raw JavaScript? I thought the goal was to move from raw JavaScript to fully jQuery. Or at least not write more raw JavaScript when we have jQuery available.

live627: Matt, no one took the incentive AFAICT.

Arantor: I find this conversation hilarious and ironic. Surely someone on the dev team during that time should have known why it wasn't used? (Answer: backwards compatibility with mods, not spending large chunks of time rewriting existing, working code for the sake of a little bit of brevity. For example, there are mods which use setInnerHTML rather than including jQuery to do it for them.) So using raw JS now is really a continuation of that.

siddhartha-gupta: Apologies, for dropping in between, but as per my knowledge pure/native JS runs faster than the jQuery.

I've ran few test cases for dom element selection, event binding and even for event bubbling, and found out native JS was a pure winner. You can feel the real difference while running the website/web app on a mobile device, as due to the hardware restriction mobile device really shows you the efficiency of the code. Few test cases can be found over here

http://jsperf.com/jquery-vs-js1
http://jsperf.com/jquery-vs-javascript-performance-comparison/14

So, it'd be better if someone feels that porting the code to jquery is not going to hamper the performance, then only the porting task should be started.

MissAllSunday: AFAIK there is no mixing, the goal is still there but I won't start doing it until there is a plan behind, I will not randomly start porting code just because. If we're going to port, we will use unobtrusive JS, meaning theres plenty more than just replacing raw JS with jQuery. You should know that already.

I agree, the page needs to use the new tools for adding JS, I'll do that.

Arantor: Um, jQuery IS native JS. It's all just native JS. The difference is whether you call a few lines, or many many more lines to do the same thing for you... So naturally non JS would be faster.

siddhartha-gupta: @Arantor ... umm yes and no about jQuery being JS.

jQuery is a library/subset of javascript. It has been built on top of javascript to simplify the code JS dev's used to write.
There is another JS lib called 'zepto' which is smaller than jQuery and has some nice features, but its way slower than jQuery. So, the question arises, if 2 libraries are based on JS how come they are different in terms of performance?

Not going into much of explanation, try to find out the difference between 'getElementById' and 'querySelectorAll'. Such small things make huge difference.

I know it won't make much difference for SMF as its more of a PHP software, but when it comes to JS development, all these small things makes a huge difference.

Arantor:
What do you mean 'yes and no'. jQuery is native JS. Examine the source yourself to see. It's not 'based on' JS, it IS JS. 2 libraries built in JS that are written differently, sure there are going to be differences in performance. getElementById (native JS method of the DOM) should be faster than querySelectorAll (native JS method of the DOM in more recent browsers) because of the difference of what they're doing... and if one uses the inefficient one in places it shouldn't, it's going to be slow. Same holds for straight JS vs jQuery - badly written straight code vs efficient use of jQuery = jQuery wins in performance. But good straight JS vs jQuery (e.g. just for things like getElementById, or in this case form targeting), straight JS will win every time because it doesn't have the weight of jQuery to contend with.

A better example would be PHP.

``` php
$x = 1;
$y = 2;

$z = $x + $y; // straight example

function add($param1, $param2)
{
  return $param1 + $param2;
}

$z = add($x, $y); // abstracted example
```

Both are PHP (just as jQuery is JavaScript), and the latter is an example of some operations in jQuery; if you're targeting a form element or you're targeting an element by id, you can use environment native methods to do it, rather than invoking the weight of something constructed on top of that environment - just as I am here to create an add function. That's all jQuery is doing - abstracting away environment native functions in a convenient and potentially easier to use wrapper.

In this particular case where this all started, you're replacing one line for one line for the sake of a few bytes of space and a lot of overhead you wouldn't need to use...

siddhartha-gupta:
I'm well aware of inside out of jQuery as its part of my job to play with JS (a lot of JS) on daily basis, and its not only about writing codes, its about optimisation and performance. I do a heck lot of benchmarking too, to measure every bit & nit of performance I can extract out of codes.

My sole point was that before porting JS of SMF to jQuery, one should consider the performance impact too, rest is up to community.

Just my thoughts.

PR #1247, opened 2014-01-20
WrapAction() won't assign a sub template if funtion is set

live627:


emanuele45:
I fail at understand why you want to use WrapAction to catch an action at all (and not the "proper" tool that is the action array), but that may be just me being dumb.
What you are writing is a sort of dispatcher for your own mod, that will fail any time another mod will try to do the same. Write the dispatcher into the code instead...

Arantor:
WrapAction is primarily for themes, not mods. But WrapAction does do one thing that previously might not have been done - setting a fallback action (but there's a hook for that too)

emanuele45:
I know what it was meant to be, and I saw how it has been changed, yet the only meaningful thing you can do with WrapAction that you cannot do with the action array is a 404 mod that acts on non-existing actions (i.e. action you anyway shouldn't know anything about, otherwise you'd use the "proper" tool that is the action array, unless you write code that invents actions and way to deal with invented actions on-the-fly).

Arantor:
Or you do what Bloc did and create themes with additional pages that add new actions without touching the action array or hooks - even back in 1.1. So, no, that's not the only meaningful thing you can do.

emanuele45:
If you are using a hook, you are not doing it from a theme (and if you are using a hook from a theme you are badly cheating the system in any possible way and manner, even for the "(kind of-)MVC-like" way SMF is coded), so no, your example doesn't sound very valid to me.

But I suppose I'm wasting my and your time, so, okay.

Arantor:
But that's the point: hooks are fine and dandy for hook cases and mods. But there's been several themes that did this without using hooks, that's the point. That's why WrapAction exists and is called from $settings - so a theme can create that functionality without having to call hooks, even back to 1.1.x. It isn't MVCish at all, but it's there and it has been used.

emanuele45:
And now you are going down the way "it has been there since forever". Yes, it has been there.
Has it been (in the past) changed giving it much more power than it should have to begin with? No, it has been changed now.
Is live627 a themer? I don't think so (at least [he didn't publish any at sm.org](http://custom.simplemachines.org/themes/index.php?action=profile;u=154736) => redirect to forum profile), I have always seen him do mods.
Should I expect live627 using this new function in a theme? Based on the above no.
Should I expect live627 using this new function in a mod? According to Bayes, yes.
Now, back on live627's actual code:

```
 if (isset($_GET['action']))
   $settings['catch_action'] = array(
     'template' => 'Custom',
     'function' => ucfirst($_GET['action']),
     'sub_template' => $_GET['action'],
   );
```

In order to check if `$_GET['action']` is set and then setting a catch_action based on that, he has to use another hook (integrate_actions? Likely), he cannot do it in [integrate_fallback_action](https://github.com/SimpleMachines/SMF2.1/blob/17f8048ab86efb78ed0aca0493fd7d266e865462/index.php#L372), because that hook is fired **after** the check that returns WrapAction, and if he is doing that in integrate_fallback_action, then he has to actually return himself WrapAction as fallback action, but then what's the reason to set $settings anyway and use a function that is in the middle of a 2k lines file? Just return the action you want.

Oh well, to me it sounds odd.

Arantor:
Yes, it's been there since forever. It has a use. There's no need to remove it or break it just because we don't necessarily do things the same way any more.

It's not our fault if live is abusing functionality that's been there for ages :tongue: For all I know he is using it in a theme which is how it is meant to be used (declare it up in template_init, job done without having to do anything else) I know he did publish a theme on his own site at one time, for all I know he's updating that theme for 2.1 and using this function.

Does it really matter? It's a feature that's been around, that has been used, whose behaviour was not entirely logical. Just because there are alternatives now does not mean that we have to remove this functionality. It doesn't hurt to leave it, and it is more logical for a themer than having a theme set hooks to achieve it when they're already going to be configuring various things in $settings anyway.

Not being funny but if you don't like what we're doing, or how we're doing it, you don't _have_ to look at it.

PR #1251, opened 2014-01-20
Info center

XinYenFon:
infocenter

### DIFF
--- a/Themes/default/BoardIndex.template.php
+++ b/Themes/default/BoardIndex.template.php
@@ -293,8 +293,8 @@ function template_ic_block_recent()
 
 	// This is the "Recent Posts" bar.
 	echo '
-			<div class="title_barIC">
-				<h4 class="titlebg">
+			<div class="cat_bar">
+				<h4 class="catbg">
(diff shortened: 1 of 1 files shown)


ghost:
I'd have to agree with Arantor on this, I'd prefer it the way it currently is.

Arantor:
No, that's not what I meant. This commit is to fix the broken styling of the headers but the way it does it is not the way we'd planned to do it.

ghost:
Regardless i think there is way to much gold/yellow in the 'info center' there, it's just too in your face, almost emphasises it over the actual board listings.

illori:
also this does not fix this issue in all the other locations were it happens

XinYenFon:
may i ask what route you want to use @Arantor ?
@illori other locations :( (stats page?)

Arantor:
The problem is that the info center - and plenty of other places - are all reusing the h3 and h4 catbg/titlebg classes and just adjusting them on demand.

What we should be doing is creating new classes and replacing those ones.

PR #1272, opened 2014-01-24
SSI code cleanup/CSS stuff

ghost:


Arantor:
Can't rely on using classes defined in the main CSS file for SSI stuff; the whole point of SSI is including into other sites - where SMF standard classes typically will not be available.

ghost:
Grand, always an exception to the rule. :)

Arantor:
Yeah, that's one of the fun and exciting problems to deal with >_<

PR #1362, opened 2014-02-05
Polished the SMF favicon. 

Xarcell:
Centered fulcrum and added a 2px border radius. Just a small tweak to make it look a tad better.

Signed-off-by: Draven Xarcell Vestatt draven.vestatt@xarpixels.com

Here is a screenshot of a new versus old comparison.
screenshot_smf-favicon-new-vs-old


mcblaber:
It does definitely look better. Although if we're updating the favicon we should probably make it so it looks good on high DPI screens too. ;)

Xarcell:
This simply replaces the old one. So that will do for now. 

Making a high DPI favicon is actually quite complicated, it's not as simple as making the icon, because every single browser fetches the sizes differently. So as far as I know there is currently no way to make all the browsers happy.

I will do some more research to see if anyone has came up with a solution.

Xarcell:
I'm not longer contributing to the SMF project.

PR #1375, opened 2014-02-08
half_content

live627:
- Add .half_content
- Have the stats page use it
- Misc. CSS cleanup


XinYenFon:
I think its best you get latest changes and look at this again :) Especially #1377

PR #1377, opened 2014-02-08
New Title_bar & Sub_Bar with cleaning

### LINKED ISSUE #1012: Link Styling Issue
Minor issue. But under 'My Community - Info Center' the title links for 'Forum Stats' & 'Users Online' are the same colour as the background so there not visible.

XinYenFon:
I'm still not sure if i missed something but i can still correct errors if i made any :)

fixes #1012

### DIFF
--- a/Themes/default/Admin.template.php
+++ b/Themes/default/Admin.template.php
@@ -216,8 +216,8 @@ function template_credits()
 
 	// The most important part - the credits :P.
 	echo '
-							<div id="credits_sections" class="cat_bar">
-								<h3 class="catbg">
+							<div id="credits_sections" class="sub_places">
+								<h3 class="subg">
 									', $txt['admin_credits'], '
 								</h3>
 							</div>
@@ -1484,7 +1484,7 @@ function template_php_info()
 				{
 					// heading row for the settings section of this categorys settings
 					echo '
-								<tr class="titlebg">
+								<tr class="catbg">
 									<td align="center" width="33%"><strong>', $txt['phpinfo_itemsettings'], '</strong></td>
 									<td align="center" width="33%"><strong>', $txt['phpinfo_localsettings'], '</strong></td>
 									<td align="center" width="33%"><strong>', $txt['phpinfo_defaultsettings'], '</strong></td>
--- a/Themes/default/Admin.template.php
+++ b/Themes/default/Admin.template.php
@@ -168,8 +168,8 @@ function template_credits()
 
 					<div id="admincenter">
 						<div id="support_credits">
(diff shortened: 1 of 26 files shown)


Xarcell:
I actually disagree with the renaming. I know Arantor said it was needed, and he's usually right, but personally I don't think it's needed.

I think it should be styled like this: "#info_center .cat_bar" instead of ".info_bar". I said that because the general styling is the same as ".cat_bar", but usually with the margin/padding adjusted. I reckon it really just boils down to design preference.

Heck if it was really up to me, I wouldn't even use unnecessary wrappers. So instead of:

```
<div class="roundframe" id="info_center">
        <div class="cat_bar">
            <h3 class="catbg">
```

it would be

`<div id="info_center" class="roundframe">
        <h3 class="catbg">`

I know we have to think about backwards compatibility here for mods, but we already keep changing it inch by inch to the point we've already broken compatibility. Pages load faster with less elements you have on the page.

Arantor:
Now do all the same for all the other places where catbg/cat_bar/titlebg/title_bar are redeclared and have a completely different visual and semantic style. Which is pretty much everywhere you see a white background with orange flanking - but using the same classes.

XinYenFon:
I think you have to be more specific here @Arantor... Unless you are not meaning you want to discard Curve2 and write a new theme for SMF 2.1 :)

Arantor:
I'm merely pointing out the problem with what @Xarcell is suggesting: you would have to re-apply that same specific change in markup everywhere that catbg and titlebg and is friends are reused inappropriately which is in many places. Along with the many interesting places that h3 and h4 are used inappropriately.

I've already said that I think the way to fix this is to ditch the theme entirely and create one where everything is actually consistent with everything else, stylistically speaking, because currently the theme isn't and I'm not sure it would be quicker to fix all the (many) bugs than to start over and build it right in the first place.

ETA: There is little to no chance of _me_ doing that with 2.1. It is merely my opinion that this is the way it should be done, and would be the way I would approach it were I to do it. However, I have no desire to do so.

XinYenFon:
heh it looks w/o quote i can't really see who you are replying :P - You are completely right about theme. But I'm still not sure I'm doing it right or wrong.

Arantor:
You're basically doing what I originally suggested, though I called it sub_bar and subbg, but the principle holds.

XinYenFon:
I'm awful at naming things if you give me list I'll change them in next commit.

Arantor:
Nah, infobar is fine, IMHO. I don't really see it as my problem, I am merely adding my opinion and letting others make of that what they wish.

Issue #1402, opened 2014-02-09
Installer does not handle DB ports properly (or at all)


Arantor:
If there is a database issue during installation for MySQL, the port will be appended to the hostname (in normal MySQL connection style)

So you end up with localhost:3306, then localhost:3306:3306 and so on, as reported in http://www.simplemachines.org/community/index.php?topic=516836.0

Also, there is no support for running PostgreSQL on a custom port, as reported in http://www.simplemachines.org/community/index.php?topic=397030.0

These can be combined into a single item: create a separate standalone item in the installer for DB port, default to whatever the platform's defaults are (hide for SQLite). If the platform's default is used, don't bother saving to $db_port in Settings.php, if not save it to that and everything else should, theoretically, work.

Oldiesmann:
The changes made by #1421 should fix this. I also added code to upgrade.php to split the host and port if you have "localhost:3306" or something similar as the db_server value. Before I pull all this in I want to do some testing to make sure it works properly with default ports (I don't have any instances of db servers running on non-standard ports) and also need to double-check as to how things are currently handled in 2.0.

Oldiesmann:
Going to close this for now as the above PR should take care of everything here. Feel free to reopen this or open a new issue if it's discovered that anything further needs to be done in this area.

Arantor:
Like the comment I left in the PR, you mean...

PR #1421, opened 2014-02-11
Improved DB port handling

Oldiesmann:
This should fix the issues brought up in #1402:
- Add a new input field for db port (hidden for SQLite of course)
- $db_port is only set if something other than default
- Adds support for specifying a non-standard port for PostgreSQL
- If port is specified, it's passed to smf_db_initiate via the db_options parameter
- Added code to upgrade.php to split port off from db_server if needed ($db_server = 'host:port'; => $db_server = 'host'; and $db_port = port;)


Xarcell:
Thanks Oldiesmann for taking the time to squash some bugs.

PR #1425, opened 2014-02-12
Update PackageGet.php

Underdog-01:
! Paths containing many files over tax the memory despite any large presets prior to the loop. Allocating it prior to the loop solves the issue. 

### DIFF
--- a/Sources/PackageGet.php
+++ b/Sources/PackageGet.php
@@ -471,6 +471,10 @@ function PackageGBrowse()
 
 			if ($package['count'] > 1)
 				$context['list_type'] = 'ol';
+				
+			// memory may be needed within the loop where many files exist within the directory
+			if (round(memory_get_peak_usage()/1048576) > ((int)ini_get('memory_limit') - 16))
+				@ini_set('memory_limit', '512M');
 		}
 
 		$packageSection++;
--- a/Sources/PackageGet.php
+++ b/Sources/PackageGet.php
@@ -247,6 +247,7 @@ function PackageGBrowse()
 
 	// Might take some time.
 	@set_time_limit(600);
+	@ini_set('memory_limit', '32M');
 
 	// Read packages.xml and parse into xmlArray. (the true tells it to trim things ;).)
 	require_once($sourcedir . '/Class-Package.php');
@@ -471,10 +472,6 @@ function PackageGBrowse()
 
 			if ($package['count'] > 1)
 				$context['list_type'] = 'ol';
-				
-			// memory may be needed within the loop where many files exist within the directory
-			if (round(memory_get_peak_usage()/1048576) > ((int)ini_get('memory_limit') - 16))
-				@ini_set('memory_limit', '512M');
 		}
 
 		$packageSection++;
--- a/Sources/PackageGet.php
+++ b/Sources/PackageGet.php
@@ -247,7 +247,6 @@ function PackageGBrowse()
 
 	// Might take some time.
 	@set_time_limit(600);
-	@ini_set('memory_limit', '32M');
 
 	// Read packages.xml and parse into xmlArray. (the true tells it to trim things ;).)
 	require_once($sourcedir . '/Class-Package.php');


emanuele45:
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs.php#L2967

Underdog-01:
Your suggestion uses 88 more bytes per occurrence & is counter-productive. 
Using all the pre-fab SMF functions is not always necessary, especially concerning one line of basic code. My recommendation stands.

Underdog-01:
Also I would like to add that in the function you linked it might be a good idea to allocate more memory slightly prior to the available memory being exhausted.

emanuele45:
1. If you feel you need that amount of memory, you are likely already screwed (unless you have a server or a VPS).
2. If you feel you need that much memory, it's useless to waste time checking inside a loop if you are using it and if not increase the amount, just do it at the beginning and hope for the best (also because increase the size may or may not work depending on server config, so it may even be something completely useless).
3. Are you arguing about 88 bytes \* 1000 loops (let's say, so 88K) over 512 MB? Seriously?
4. ini_get could return for example `1G` or other odd things, that would be translated in your code to 1
5. it may have a meaning if you are dynamically increasing the amount of memory (e.g. `ini_set('memory_limit', (ini_get('memory_limit') + 32) . 'M');`), but considering you are not able to increase the amount of memory "forever", **and** over a certain limit you'll hit some timeout issue, it is really worth consider to convert everything to some kind of dynamic memory allocation?
6. [memory_get_peak_usage](http://www.php.net/manual/en/function.memory-get-peak-usage.php) `(PHP 5 >= 5.2.0)`: last time I checked 2.1 was php 5.1+, but I didn't follow very well the discussions around, so it may be it has changed, or anyway it can be changed rather easily.

Of course these are just my own personal thoughts, and since I'm just another commenter, do not feel like my comments have anything to do with SMF. (This is just a standard disclaimer to avoid misunderstandings, some people (not you) have tendency to think everything I do is against them, so better take a moment to explain that what I write is my own opinion and should not influence any of your contributions to the project. ;))

Oldiesmann:
I'm inclined to agree with emanuele on this. I don't see why we need 512MB of memory there. Even with the huge number of packages on simplemachines.org, I've never run into memory issues on a server with a memory_limit of 32M (though 3GB allocated to the VPS). 512MB of RAM is absolutely ridiculous for a PHP script.

Also, as far as the PHP version goes, we were planning to bump it up to 5.3 at some point (in part to fix some of the issues caused by the PHP 5.5 fixes), but it's not really useful to try to get the memory usage, as we don't even know how reliable that function is or whether some hosts might decide it's not something they want their users to have access to (some hosts still insist on disabling functions that might be dangerous on an improperly-configured server). If there's a chance we will run out of memory, we should just try to increase it once, rather than setting it at a random point inside a loop.

I'm not convinced this is needed though. Simplemachines.org has tons of packages on its "package server" and I've never run into any memory issues while trying to browse it from my forum, on a server with a memory_limit of 32M.

Underdog-01:
When I was testing this on a live forum the other day, it was not working for me when I attempted to adjust the memory limit prior to the loop which is why I came up with this edit. 

  Prior to today, I did test this on 3 live forums and 1 local and the behavior was the same. Now today when I test it without any memory_limit edits it is working just fine.

  Somehow I was getting an odd result on that live forum that has a php.ini setting of 128M & atm can not figure why it was so (the file was as the default).   

  I am aware that adjusting that setting within the loop does not seem necessary as the 1 setting prior to the loop should work for as long as the script is running. However, at the time this was the behavior I was experiencing. Atm I am puzzled as to the behavior of the script I was witnessing. 

 Unless something was somehow in that folder that was causing the issue that was since removed? This is the only possibility I can think of atm!?

  Also this was reported to me in a PM for a supposed new forum and from a SMF member that was also a new user. You'll have to excuse this comment but I feel as though I have fallen for a ruse. :(  

  Apologies.. indeed very puzzling to me at this moment. I have removed this edit.

```
The next issue I spot & attempt to fix will involve a second tester/witness.
```

---

  Also as I said, Imo the setMemoryLimit function needs an adjustment. If the needed memory is exceeded that function will not be processed prior to php crashing.

Underdog-01:
emanuele45,

  In response to your last paragraph... No I do not think you have anything against me. I make mistakes sometimes & if it is pointed out then I learn + admit it. 
  I will say that there are some individuals which I can not tolerate (ie. those that talk cr*p to people all the time and use excuses as to why they should be able to.. no names.) but you nor Oldiesmann are definitely not in that list.

---

emanuele45 & Oldiesmann,

  I do not know why I was seeing those results where the memory was spiking when reading file names from that dir path. I swear something was up with that specific directory the other day because it was crashing on 4 different forums (the user that reported it, my forum, Skhilled forum and a local install). Oddly enough, putting the memory_limit within the loop worked even though I knew it should not have been applied that way. My mistake was not stopping to realize that something else must have been wrong to cause the behavior I was witnessing. 

  What I do not understand is that now it appears to be working just fine  (without any edits to the default files) but how can this be unless something that was in that directory was causing it? Also if something was done within the directory since my initial report (ie. removal of a corrupted file, chars in a file name causing the issue somehow, etc.), I was not informed,

PR #1427, opened 2014-02-13
Half content

live627:
A redo of #1375 that should be much fresher

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -4299,16 +4299,16 @@ h3.titlebg, h4.titlebg, .titlebg, h3.subbg, h4.subbg, .subbg {
 	background: none;
 	color: #555;
 	font-family: "Tahoma", sans-serif;
-	font-weight: bold;	
+	font-weight: bold;
 	overflow: hidden;
-	padding: 6px 12px 5px 12px;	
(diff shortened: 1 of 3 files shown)


XinYenFon:
my community - statistics center 2014-03-03 22-19-38

Themes/default/css/index.css
```diff
@@ -326,6 +326,12 @@ h3.largetext {
 .main_section, .lower_padding {
 	padding-bottom: 6px;
 }
+.half_content {
+	-moz-box-sizing: border-box; -webkit-box-sizing: border-box; box-sizing: border-box;
```
XinYenFon, line 330: You don't need this actually because "#statistics div" already gives that to all divs.

Themes/default/css/index.css
```diff
@@ -326,6 +326,12 @@ h3.largetext {
...
+	width: 49%;
```
XinYenFon, line 331: better you increase this to 50% and add float: left :)

XinYenFon:
and i think it would be better if you add; 

#statistics .windowbg2 {
    background: none;
}

Themes/default/css/index.css
live627, line 330: There's more areas that have two columns; the mod center is one.

Issue #1392, opened 2014-02-09
CSS/JS resources should always use proper load functions


Arantor:
Two reasons:
1) Means they get logged properly for the debugging information
2) Means that come release time, no-one has to go through and change hundreds of instances of ?alp21 to ?beta or whatever. At least one of the loaders already has facilities for including this cache buster.

Proper use will also cover off http://dev.simplemachines.org/mantis/view.php?id=3268

Arantor:
Since it was brought to my attention that I wasn't detailed enough, and even though I took the trouble to see what others had added before me, I guess I can summarise this for everyone else. Even though these functions are in use in the source already.

loadCSSFile is the first, which accepts a filename as its simplest case, but accepts a variety of options in an array as the second parameter, these are all outlined in Load.php where the function is defined. An id can also be provided for caching purposes but is not explicitly required.

loadJavascriptFile is the second. Like loadCSSFile it accepts a filename as the simplest case which will normally be relative to the theme's directory. Again there is an array of options as the second parameter and an id for third. Of the options, 'default_theme' => true is the most useful, forcing the script to be loaded from the default theme. Again, the list of options is outlined in Load.php where the function is defined.

MissAllSunday:
It was me, apparently the IRC staff only isn't that "staff-only anymore"...  it was made out of frustration, I already figure it out how it works. Not a big deal. Ask for logs to know exactly what I wrote to avoid further misunderstandings.

Arantor:
Oh, I saw the quote. Fairly hard to misinterpret it. If you're going to make snide remarks about me, do it to my face. Mind you, I thought it was fairly hard to misinterpret "I don't care how fast anyone works as long as stuff gets done", but I guess we're both wrong.

Also, when the login form is loaded AJAXively, template layers don't get run, meaning that template_javascript() never gets run...

MissAllSunday:
It was not about you, it was in general, if it was other person who wrote the report I would have said exactly the same thing. In fact, I thought exactly the same about other report I was looking at.

Once again, please ask for logs, I didn't wrote this: "I don't care how fast anyone works as long as stuff gets done"  so whoever told you or show you the quote is modifying the source, please ask for logs to avoid further misunderstandings. Thank you.

Arantor:
[...]

PR #1433, opened 2014-02-15
Css/jsstuff

MissAllSunday:
oops, forgot to add a description... this basically covers #1392 whenever possible.

There are some instances where the loaders cannot be used so I added $modSettings['browser_cache'] to them

### DIFF
--- a/Sources/Admin.php
+++ b/Sources/Admin.php
@@ -563,7 +563,7 @@ function AdminHome()
 	);
 
 	if ($context['admin_area'] == 'admin')
-		loadJavascriptFile('admin.js', array('default_theme' => true, 'defer' => false), 'smf_scripts');
+		loadJavascriptFile('admin.js', array('default_theme' => true, 'defer' => false), 'smf_admin');
 }
 
 /**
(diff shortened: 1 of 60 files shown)


Xarcell:
Good work.

Oldiesmann:
This looks good, but why do we have a function to reset $modSettings['browser_cache'] and why is it being called on every page load? We just need to set it during the install/upgrade and that's it - there should be no reason to reset it afterwards.

MissAllSunday:
Yep, thats true, the reason why theres a function (or at least what I understood from the old bugtracker talk) was to force the browser to call the files directly when a mod or something else added new css/js files, that was the original intention, the original code didn't added the current forum version, it added a random generated seed.

It all depends if we want to use this and follow the original intention or just use it to identify the files.

Oldiesmann:
If mods need to add their own files with different caching requirements, they should be using their own setting rather than trying to change the one used by SMF. It's best to just let us use our own and then mods can either use the same one or set their own if they want to do it differently (since some portal mods might want to set it based on portal version instead).

MissAllSunday:
Yep, fair enough, I'll do the change.

PR #1436, opened 2014-02-19
im fields

### LINKED ISSUE #1391: Migrate some fields to custom fields
Some of the fields we have currently as hardcoded would be better being migrated to custom fields. The upgrader could do this - it would only create (say) ICQ if any users had an ICQ field previously. It would end up decluttering the memberlist and the topic display and the members table itself, while allowing for better control from the admin in the long run.

ICQ, AIM, YIM and Skype are obvious candidates, Location is another, and potentially so is Gender (which would also deal with the issues around multiple genders as we have seen before). This would also cover off http://dev.simplemachines.org/mantis/view.php?id=1402 at the same time.

MissAllSunday:
Fixes #1391

Yes, I tried to use the old PR but the patch file had too many issues and there was no branch to cherry-pick the commits from.


live627:
I think this is more complete than my old attempt.

PR #1446, opened 2014-02-27
Update RepairBoards.php

Underdog-01:
! Some fields in the boards table do not have default values and must be defined


Underdog-01:
illori,

  My apologies. I was multitasking many things at once & likely made an error here.. 
I came across the issue while working on someones database & did of course create a fresh installation of a smf db config to double check the defaults but must have been looking at the wrong db?! 

  Again apologies.. I will remove the 2 pull requests that were done in error.

PR #1449, opened 2014-03-01
Update Xml.template.php

Underdog-01:
! missing required xml stylesheets (rendered pages were throwing errors)

### DIFF
--- a/Themes/default/Xml.template.php
+++ b/Themes/default/Xml.template.php
@@ -12,9 +12,10 @@
 
 function template_sendbody()
 {
-	global $context;
+	global $context, $settings;
 
 	echo '<', '?xml version="1.0" encoding="', $context['character_set'], '"?', '>
+<?xml-stylesheet href="' . $settings['theme_url'], '/css/index', $context['theme_variant'], '.css?fin20"?>
 <smf>
 	<message view="', $context['view'], '">', cleanXml($context['message']), '</message>
 </smf>';
@@ -22,9 +23,10 @@ function template_sendbody()
 
 function template_quotefast()
 {
-	global $context;
+	global $context, $settings;
 
 	echo '<', '?xml version="1.0" encoding="', $context['character_set'], '"?', '>
+<?xml-stylesheet href="' . $settings['theme_url'], '/css/index', $context['theme_variant'], '.css?fin20"?>
 <smf>
 	<quote>', cleanXml($context['quote']['xml']), '</quote>
 </smf>';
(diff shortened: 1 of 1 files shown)


Underdog-01:
One can add a specific xml css stylesheet within the default theme and change the file locations to it since a standard plain template can be used here for xml display. Still my point is without the stylesheet xml templates show an error.

Oldiesmann:
Please take a minute to review what things actually do before deciding they need to be fixed. The functions in that file are used for various AJAX functions throughout SMF, and the XML from those templates is never displayed to the end user. The stylesheets are already loaded by the time the information from that XML is displayed to the user via JavaScript.

Underdog-01:
The style sheet issue appeared in SMF 2.0.7 but now when I test it in SMF 2.1 you are correct as it does not display an error regarding a missing style sheet.

There is an issue with xml display though... 
The page crashes because of named entities (ie. &nbsp) which is an issue no matter what it is used for but that is another matter.  This happens in both SMF 2.1 & SMF 2.0.7 versions.

Oldiesmann:
There's no point in using `&#160;` instead of `&nbsp;` and similar, for the same reason we don't need to add XML stylesheets. The content of the XML tags is injected into HTML and displayed to users. It's valid when it's displayed even if it's not valid when viewed in XML context.

Underdog-01:
So what is the point of having the xml appear when ;xml is included in the url?
ie. http://www.simplemachines.org/community/index.php?action=help;xml
I thought it was for search engine bots/crawlers.
When that is done the page breaks & does not render properly when it hits the named entities.

If it is for rss feeds, ie. feedburner also obviously states its an invalid page due to the named entity error.

Oldiesmann:
The ";xml" is a shortcut to tell loadTheme() to load Xml.template.php, thinking that the contents of what we're displaying are fully Xml. I'm not sure why things were done this way, but that's how things work. It was never designed to do things like what you demonstrated.

Underdog-01:
I have recently witnessed a forum redirect to a crashed xml page when auto-logged out due to the 'time logged in' setting (ie. default 60 mins). This is how I discovered what I am conversing here.
Imo something not right there with the code setup.

Oldiesmann:
That really shouldn't happen unless SMF is trying to call something via AJAX at the exact moment their session expires. I've never heard of that happening before, so it must be a pretty rare occurrence. Definitely something to keep an eye on though.

Underdog-01:
Ajax shoutbox set on refresh? I put a small patch in the main index.php & testing it atm. 
The patch allows the shoutbox to work but redirects someone to the wireless template when ;xml is included in the url.. not sure if that will bugger something else though.

Oldiesmann:
Really the only way to get around this issue is to add an array of action => permission pairs, then when ";xml" is in the URL, strip out the action, check the permission and redirect to the action without ;xml if they can't do it. I'm not sure how well that will work though.

Oldiesmann:
Going to go ahead and close this one since we don't really need stylesheets for the XML templates, but the issue you mentioned is still something we need to investigate further, so I'll take a look at that when I get time.

PR #1455, opened 2014-03-01
Update install_2-1_sqlite3.sql

Underdog-01:
! missing required field 'location' (ref. install.php - currently line 1465)


Underdog-01:
With a fresh installation it throws an error without this column because it attempts a query with it in install.php yet I just found it being dropped in the upgrade script.

... not in the source files therefore I assume this should be removed from install.php

MissAllSunday:
I am aware of this and all other related issues, On a recent commit  moved the gender, location, aim, yim, skype etc to cust profile fields. No need to add them back.

I'm also aware the current upgrade process is broke due to those changes, no need to open yet more reports about it.

PR #1466, opened 2014-03-06
Fix various errors in Upgrade process (other/upgrade_2-1_mysql.sql)

nwsw:
I have been running the upgrade.php tool on a real forum dataset. This process revealed two serious issues: PM Labels cause big problems for the upgrade, and existing attachments don't work after the upgrade.

This request now includes three commits to deal with this:
1. e73bd602fb900a736eaeb39e85c9c4efea06dcbc:
   This works around the fatal errors created by commit 754c249ef32948df6fe17bae0a03c3379a7a2afe. This needed to be done before I could actually test/fix the other problems. Obviously it can be ignored if a real fix is added.
2. cd453e1f718281a96858929007c0f58389ca530a
   This fixes several problems with the handling of PM Labels. It also fixes a query error in the attachment handling.
3. 37e977b2bafc58905f7254c764ef6ba2e026a0f7
   This fixes the attachment rename processing. The biggest problem in this code was missing braces around the _Check if the av is an attachment_ hook.


nwsw:
Additional fixes posted to the mysql version of the 2.1 sql script.

Merge or don't...whatever.

nwsw:
I updated my first comment with a full description of the upgrade problems.

MissAllSunday:
Thank you for your contributions but you need to sign off your commits for use to merge them. All of them.

nwsw:
Technically, both of
- upgrade_2-1_postgresql.sql and
- upgrade_2-1_sqlite.sql 

appear to suffer from the same PM Labels and attachment upgrade problems. I do not have a setup in which to test this, so I did not touch those areas.

Update: I see now that you said the following in the merge:

> like I already said, we are AWARE of the issues and solved them already, now we will have to re-do our code) and this isn't something critical, I'm pretty sure the install/upgrade will continue to get errors as we add more stuff to it, its only natural since we aren't even on alpha stage.

Since you are apparently aware of these issues and solved them already, I will stop worrying about the other sql files. Apparently, I have been wasting my time on these problems.

Issue #858, opened 2013-11-11
Board index needs rewrite


Arantor:
The board index code needs a rewrite. Firstly the way it tries to grab board moderators is just asking for trouble in terms of multiplying out the number of rows pulled when there are moderators, and this should really be a separate query. Perhaps for each board, indicate whether there are moderators or not and query them only if there are.

More importantly, it's been an issue for some time that InnoDB got the order wrong because InnoDB stores by PK order and SMF has since forever been trying to coerce the boards table into display order. While some workarounds were documented for InnoDB, with more recent changes to MySQL (5.6+), it seems that this predication of ordering won't consistently work even in MyISAM.

Any place we query the boards table, we need to explicitly ensure that we get the ordering correct and don't rely on the old semantics to do it for us.

Dragooon:
Hmm, doing a simple query over boards with sorting by board_order and child_level still produces a file sort, although I guess it'll be much faster since it's a much smaller data set?

The idea I had was to create a new function called getTreeOrder which stores the entire category and board tree order in a long term cache and another function called sortBoards which takes an array of boards and sorts it with respect to the order defined. This way we avoid direct order by in heavy queries and still get properly sorted

Dragooon:
Also, as far as moderators go, wouldn't a separate single query to fetch the moderators and their names be much faster?

EDIT: Just realised you said the same thing

Arantor:
Considering that you're pulling from two sets of moderator tables (member as moderator, group as moderator, each of those with respective table joins), hiving that off to separate queries would be useful. This, of course, assumes you actually want to display all the moderators on the board index, if you didn't want to display such, you could cut all that stuff out.

This issue was mostly about trying to reorganise the board index to avoid all the mess of board/group moderators to make it faster; #887 covers the actual order-by issue, including discussing all the issues around filesorts and changing the TEXT columns to VARCHARs which should limit the chance of filesorts (filesorts primarily occur when the size of memory that would need to be allocated is too large, and when you deal with such, MySQL will allocate the maximum row size for every row for sorting, so you end up dealing with rows > 64KB per row and filesorts are guaranteed at that point)

Dragooon:
It'll still hit filesort if you only fetch smf_boards table (id_board and id_cat), because of the two column order which is not using any indexes. A separate query should be the faster solution here, but even then I can't seem to avoid filesort but this is a lot faster due to the much smaller data set.

Arantor:
[...]

PR #1478, opened 2014-03-21
Added ORDER BY to Subs-BoardIndex.php, hopefully fixes #887

### LINKED ISSUE #887: Boards/Categories sorted after upgrade
I upgraded my alpha build 3 times (2.0.x to 2.1 Alpha) - (2.1 Alpha to 2.1 Alpha) - (2.1 Alpha to 2.1 Alpha) and my boards/categories sorted differently from as it needs to be.

But if i edit a board (move up/down) everything fixes itself.

Apache version  2.4.7
PHP version     5.4.23
MySQL version   5.5.35-cll

margarett:
Added ORDER BY

As far as I've tested, this fixes:
- Wrong board order (different from admin)
- Wrong board order when adding local moderators
- Posts in child boards and empty parent do not show in Board Index

All tested with MySQL 5.6

Signed-off-by: margarett bruno.m.alves@gmail.com

### DIFF
--- a/Sources/Subs-BoardIndex.php
+++ b/Sources/Subs-BoardIndex.php
@@ -70,7 +70,8 @@ function getBoardIndex($boardIndexOptions)
 			LEFT JOIN {db_prefix}attachments AS a ON (a.id_member = m.id_member)' : '') . '
 		WHERE {query_see_board}' . (empty($boardIndexOptions['countChildPosts']) ? (empty($boardIndexOptions['base_level']) ? '' : '
 			AND b.child_level >= {int:child_level}') : '
-			AND b.child_level BETWEEN ' . $boardIndexOptions['base_level'] . ' AND ' . ($boardIndexOptions['base_level'] + 1)),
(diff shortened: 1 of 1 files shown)


Oldiesmann:
While it does fix the issues, it's not really the best solution. See #858 for more info.

margarett:
Looking at it, I think that we can just remove the moderators and group moderators tables, transforming those into text/varchar fields at the boards tables, ids separated by commas. 
Or is there a reason I'm missing for having dedicated tables?

This part (local moderators) seems pretty obvious to change, the rest of the huge query I'm not that sure. Everything there seems really necessary. 

I can also think about getting independent queries for the members table (in order to avoid that massive joins)

What do you think?

Oldiesmann:
I honestly can't see any advantage to doing that, or even having a flag to indicate whether or not a board has moderators. Is saving a couple of queries (which, when split out from the big one, are actually pretty simple as far as queries go - pulling two columns and using an INNER JOIN and a WHERE) really that important?

It may be best to consult with Arantor on this and see if he can shed some more light on this situation.

Dragooon:
This can be closed as it was deprecated by #1637

Oldiesmann:
Deprecated by #1637

PR #1493, opened 2014-04-01
Update Reports.template.php

gevv:
Theme Parse Error!


XinYenFon:
Good catch but you need to sign-off your commits (each of them), otherwise we cannot use your commits :(

PR #1522, opened 2014-04-16
Delete topic

XinYenFon:
This turns remove post into remove topic if the post is the first of the topic.


emanuele45:
IIRC remove topic and remove message are two different permissions.

Oldiesmann:
They are indeed two separate permissions.

Issue #648, opened 2013-09-30
Invalid profile field values never throw errors


Arantor:
When editing profile fields, invalid values are never made clear to the user. This is most notable with custom profile fields where a regex is used to validate the field, but it only checks and removes the value if it doesn't match. This probably applies to other fields too.

An error should be being thrown back to the user to tell them that something doesn't fit.

MissAllSunday:
closed via #1528

PR #1528, opened 2014-04-18
Invalid profile values

MissAllSunday:


MissAllSunday:
WTF, I didn't even had the chance to type a description...

Anyway... this fixes #648 Invalid profile field values never throw errors

at least partially, theres still the admin registration but since that stuff doesn't have any error handler I didn't touch it.

It also fixes some minor stuff here and there  and adds new places for custom profile fields

PR #1549, opened 2014-04-23
NO HTML allowed :P

### LINKED ISSUE #1548: karma labels should not accept html
it can cause issues...

XinYenFon:
fixes #1548

### DIFF
--- a/Themes/default/Display.template.php
+++ b/Themes/default/Display.template.php
@@ -471,7 +471,7 @@ function template_main()
 
 function template_single_post($message, $force_alternate = null)
 {
-	global $context, $settings, $options, $txt, $scripturl, $modSettings;
+	global $context, $settings, $options, $txt, $scripturl, $modSettings, $smcFunc;
 	static $alternate = false;
 	if ($force_alternate !== null)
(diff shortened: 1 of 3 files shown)


emanuele45:
I think you should do that before saving the value into the database (and definitely not in the template ;)).

XinYenFon:
that would be better ofc, but i didn't find any clue about it (Took a look at how its done in board names).

Oldiesmann:
Karma is probably going to be removed at some point anyway, but this needs to be done when we're saving settings, not when the setting is loaded in the templates.

matthew-kerle:
Since when is it cool to use htmlspecialchars in a template?

matthew-kerle:
You can always modify it in the source file and then the template can grab it.

$modSettings['varOne'] = $smcFunc['htmlspecialchars']($modSettings['varOne'], ENT_QUOTES);

That in the source, and then just use the var in the template and it's clean.

XinYenFon:
https://github.com/SimpleMachines/SMF2.1/pull/1556 ?

matthew-kerle:
Ah my bad. Good to see. :)

PR #1477, opened 2014-03-20
Div styling

XinYenFon:
Removed table styling from boardindex & messageindex

PR #1592, opened 2014-04-30
Use smc_Toggle to toggle categories

live627:


XinYenFon:
is it possible to add the same effect info_center has? Thanks it looks awesome!

live627:
not without using divs.

XinYenFon:
So after https://github.com/SimpleMachines/SMF2.1/pull/1477 ? :)

live627:
yeah

PR #1611, opened 2014-05-04
nobbc tags wouldn't work if they encompassed multiple lines. 10 years...

live627:


MissAllSunday:
I believe we still have php 5.1 as minimum, thus, we can't really use anonymous functions :(

PR #1619, opened 2014-05-06
Ajaxified likes

### LINKED ISSUE #801: Likes with AJAX
Putting this here so I don't forget, likes currently work just fine without JS. Needs AJAX goodness as well.

MissAllSunday:
fixes #801

Decompose each like subaction into its own method, each method can be called directly and will produce a generic response. Handles both ajax calls and normal calls. If we do move to php 5.2 the response method can be easily adapted to return more data as a json object.

Hooks now need to return an array of data.

There are a few new language strings so the usual check needs to be done.


XinYenFon:
If post was liked by other person before after i like a post and click "x other person" link part it redirects to page (beside the popup). I need to reload the page to see the popup again.

MissAllSunday:
Should be fixed now.

XinYenFon:
yea thanks :)

Edit: One thing i noticed; while I'm using my admin account like/unlike action is so fast but when changed to regular user (no extra permisssion) it takes up to 4 seconds to complete the action (like/unlike)

MissAllSunday:
it shouldn't have any difference, the class doesn't implement any permissions besides {query_see_board}

I tested it with a normal account and i couldn't notice any difference.

PR #1632, opened 2014-05-11
Search API cleanup

Dragooon:
Introduce search_api_interface and search_api in order define the full functionality and capabilities of a Search API.  Simplify the extending classes in order to improve the code reusability and readability.

Don't merge yet, I'd like to address SimpleMachines#1563 first.

### DIFF
--- a/Sources/Class-SearchAPI.php
+++ b/Sources/Class-SearchAPI.php
@@ -23,7 +23,7 @@ interface search_api_interface
 	 * @param array $query_params
 	 * @return boolean
 	 */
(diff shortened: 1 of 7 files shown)


Dragooon:
Okay, added the relevant callbacks for moving and deleting a post and topic. There's still an issue when posts removed are recycled which are done by creating a new topic and moving the post into it

Oldiesmann:
Is the remaining issue an easy one to fix at all, or something that should even be dealt with?

Dragooon:
I don't think it needs to be dealt with, the ID of the message remains the same and so does the body.

PR #1634, opened 2014-05-12
BBC & Collapse on topics default on

XinYenFon:
fixes #1633
fixes #1558 
Signed-off-by: Antes antes@simplemachines.org

PR #1638, opened 2014-05-13
! BBCode buttons should be enabled by default

Oldiesmann:


Oldiesmann:
Closing this since Antes took care of it with #1634

PR #1646, opened 2014-05-23
Should fix #1604 (if I understood correctly)

### LINKED ISSUE #1604: delete_subscription query identifier for PGSQL
The PGSQL abstraction library defines a replacement for 'delete_subscription' as for removing a LIMIT 1 in the query, except the query that is referenced does not have a LIMIT 1, so it should be removed from both the PGSQL abstraction library and ManagePaid.php.

NanoSector:


NanoSector:
Shit, wrong branch.

PR #1655, opened 2014-05-23
Get rid of pregReplaceCurry and use closures instead

Dragooon:
Perks of PHP 5.3+, we can use closures instead of currying our way to pass scope into the callbacks, saves us from a lot of create_function hell

### DIFF
--- a/Sources/Modlog.php
+++ b/Sources/Modlog.php
@@ -606,7 +606,6 @@ function list_getModLogEntries($start, $items_per_page, $sort, $query_string = '
 	}
 
 	// Do some formatting of the action string.
-	$callback = pregReplaceCurry('list_getModLogEntriesCallback', 3);
 	foreach ($entries as $k => $entry)
 	{
 		// Make any message info links so its easier to go find that message.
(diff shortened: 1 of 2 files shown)


jdarwood007:
Should note this was done this way as we supported older PHP versions.  If we have dropped everything below 5.2 now then this can be done.

jdarwood007:
Also, my personal belief is that creating the functions with create_function and assigning it to a variable.  I want to say that its more memory/resource friendly to do

Dragooon:
> Should note this was done this way as we supported older PHP versions. If we have dropped everything below 5.2 now then this can be done.

We decided to bump to 5.3 (at which point closures were introduced) so we're good.

> Also, my personal belief is that creating the functions with create_function and assigning it to a variable. I want to say that its more memory/resource friendly to do

Sorry but that's completely wrong. Functions created with create_function are stuck in memory heap until the end of execution, causing a lot of memory usage. This was the entire reason 2.0.7 became more memory heavy than it's predecessors. pregReplaceCurry gets called hundreds of times by parse_bbc/parsesmileys on a BBC heavy posts consuming lots of memory and causing white screens. Closures are a lot more memory friendly (and can be assigned to variables too) and much more readable, I fixed this in the 2.0.8 patch by using closures but since we've already bumped to 5.3 for 2.1 I'd rather like to use closures since they're cleaner and easier to follow.

Arantor:
Functions created with create_function are declared as normal functions with a funky name preventing them appearing in the result of things like get_defined_functions(). 5.3's anonymous functions are really instances of the Closure class and from a strict allocation point of view, a closure should consume slightly more memory just because it's got the class stuff hanging around too (which in theory means if you can find it you can also apply Reflection, hence the overhead)

On the other hand, using a proper closure will actually be faster than a create_function() created function since all you're actually doing with create_function is creating a variable function. The result is not a function but a string and it's called just as SMF does in plenty of other places; it's how the rest of $smcFunc works. The overhead of calling that in theory should be slightly higher than calling a closure, but I've not benchmarked any of this to verify.

Really, though, the memory use difference between a single create_function and a single closure is negligible. The key difference here is that you're not creating a bajillion of the beasts. 5.3.2 is officially - according to the installer - the minimum required for 2.1 (I have no idea why 5.3.2 vs 5.3.0 or vs 5.3.8 or any other version; at least I can think of a reason for 5.3.8)

jdarwood007:
Well thats good to know.  I always thought create_function was slightly faster to function and thus why we used it.

We do have many other instances of create_function that we should look at as well then and address if they are the best possible usage.

Arantor:
Actually, no, create_function was only used because it was more convenient than the alternatives. Some of the uses (parse_bbc comes to mind) are just about avoiding lots and lots of separate definitions, where inline but otherwise anonymous functions would be more appropriate.

$smcFunc of course is all about conditional compilation, e.g. the strtolower function being directed to mb_strtolower if available and native strtolower if not, and avoiding the hit of the if branch every function call. Similar deal with htmlspecialchars; it's cheaper to accept a tiny penalty of variable calling each time than it is to accept the hit of an if branch every time you call it (like whether you're doing UTF-8 or not in preg functions)

It's really down to whether you need true conditional compilation or whether you're just acting on local variables that need to be pulled into scope or not.

Oldiesmann:
5.3.2 was chosen because we plan to upgrade password hashing to SHA256 still, and I was told that was the minimum version for supporting SHA256. After further investigation, it looks like there was a bit of confusion here - `crypt()` does not support SHA256 before PHP 5.3.2, but `hash()` does support it as of PHP 5.1.2 (when that was added as a default extension in PHP).

In retrospect it might be good to support both methods (even though crypt is a bit different than hash), but we could also just make the installer throw an error if hash isn't available for some reason (I don't see that happening much, but some hosts do weird things with PHP, so I guess it's possible).

Issue #1538, opened 2014-04-20
0005001: Login error with non-ascii charaters


XinYenFon:
In SMF usernames are not case sensitive when login. But if you give wrong size letter in username with non-ascii characters when login, smf gives password error in first attempt to try login. Second time works.
1. Create user: Ääkkönen
2. Login with "Ääkkönen" and correct password -> works fine.
3. login with "ääkkönen" and correct password -> it gives password error. Try again with same "ääkkönen" with correct password, second time everything works fine. Happends every time.

It does that with usernames with non-ascii characters, like öä etc.
Mantis: http://dev.simplemachines.org/mantis/view.php?id=5001

PR #1674, opened 2014-05-26
Update hashing algorithm for passwords to BCrypt

Dragooon:
Previously hashes were computed as sha1(lower(username) . password), update this to add a layer of sha256 so that it becomes sha256(sha1(lower(username) . password) based on ideas by @jdarwood007.

The cookie password hash becomes sha256(password_hash . salt) instead of sha1(password_hash . salt), but the login hash remains same as before. This provides additional protection as someone with the hash from the database can't login directly simply by submitting that into the login form, they'd still need the original password to be able to login

EDIT: Updated this PR to move to entirely to bcrypt instead, and drop client side hashing. See commits for reasoning


live627:
Shouldn't `strtolower()` be using `$smcFunc['strtolower']()` in order to allow for Unicode characters?

Dragooon:
I'm not sure, it was strtolower all the places plus it's simply used to calculate the hash so it doesn't affect anything as such. I'd say leave it alone since it might break something? Or should we change it?

live627:
I'm pretty sure it'll fix #1538

Dragooon:
That is a whole rat's nest of issues, using $smcFunc doesn't really fix it, it breaks it for uppercase letters since the client side can't handle making them properly lower case (as we always hash on client side, the password is incorrect in case the user tries to enter upper case first letter or something). Using JS's toLowerCase + $smFunc['strtolower'] _might_ fix it but I'm not 100% certain, since the client side will do it correctly but on the off chance that the server doesn't have mb_strtolower, it becomes off on the server side.

Dragooon:
Okay since we're moving away from sha1 anyway, I figured go full length and implement bcrypt instead. This requires to bump up the min. version to 5.3.7 but that's fine.

PR #1676, opened 2014-05-26
[WIP] Convert create_function calls to true anonymous functions

Dragooon:
With the recent bump in minimum PHP requirement as PHP 5.3, we can use closures and true anonymous functions, which are better, more readable and less memory intensive.

Currently I believe I got almost all of them but a few are still left, and these are barely tested. So don't merge them yet and any help testing them will be appreciated. Thanks!

### DIFF
--- a/Sources/Groups.php
+++ b/Sources/Groups.php
@@ -93,34 +93,35 @@ function GroupList()
 					'value' => $txt['name'],
 				),
 				'data' => array(
-					'function' => create_function('$rowData', '
+					'function' => function ($rowData)
+					{
 						global $scripturl;
 
 						// Since the moderator group has no explicit members, no link is needed.
-						if ($rowData[\'id_group\'] == 3)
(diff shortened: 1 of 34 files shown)


Sources/Groups.php
```diff
@@ -93,34 +93,35 @@ function GroupList()
 					'value' => $txt['name'],
 				),
 				'data' => array(
-					'function' => create_function('$rowData', '
+					'function' => function ($rowData)
+					{
 						global $scripturl;
```
live627: Can't this use the 'use' keyword?

Dragooon: For $scripturl? It can, but it was this way almost all the places so it'd be required at all the places
/me is lazy

Dragooon:
The last commit will make @live627 happy :D

Dragooon:
Uhh...I think I kinda messed up this PR :P

PR #1677, opened 2014-05-27
Remove old PHP compatibility code

Dragooon:
With the recent bump to 5.3, we don't need most of the stuff. I'm still leaving sha1 functions since it's used in Login2 function for old passwords compatibility. Delete them too?


Sources/Subs-Compat.php
```diff
@@ -224,234 +127,13 @@ function sha1_rol($num, $cnt)
...
-
-/**
- * Compatibility function.
- * crc32 doesn't work as expected on 64-bit functions - make our own.
- * http://www.php.net/crc32#79567
- * @param $number
- */
-if (!function_exists('smf_crc32'))
-{
-	function smf_crc32($number)
```
live627, line 315: grep.exe -nHIirF -- smf_crc32 (in directory: C:\Users\John\Documents\repos\SMF2.1\Sources)
./Class-Graphics.php:662:       $out .= $tmp . pack('N', smf_crc32($tmp));
./Class-Graphics.php:669:           $out .= $tmp . pack('N', smf_crc32($tmp));
./Class-Graphics.php:682:           $out .= $tmp . pack('N', smf_crc32($tmp));
./Class-Graphics.php:688:       $out .= $tmp . pack('N', smf_crc32($tmp));
./Class-Graphics.php:698:if (!function_exists('smf_crc32'))
./Subs-Compat.php:313:if (!function_exists('smf_crc32'))
./Subs-Compat.php:315:  function smf_crc32($number)
./Subs-Package.php:128: // smf_crc32 and crc32 may not return the same results, so we accept either.
./Subs-Package.php:129: if ($crc['crc32'] != smf_crc32($data) && $crc['crc32'] != crc32($data))
./Subs-Package.php:3184:if (!function_exists('smf_crc32'))
./Subs-Package.php:3193:    function smf_crc32($number)

Dragooon, line 315: Good point, I've added it back in

Dragooon:
It's not for pre PHP 4.2 installations. Installations upgrading from pre PHP 4.2 systems would be using smf's SHA implementation, which can be off on Linux systems (according to the comments) so we need it in order to make sure they don't break the passwords. Although it's fairly low priority and I am this close to removing it as is.

Issue #1203, opened 2014-01-16
RTL compatibility for install & upgrade pages


ghost:
## *\* Old Bug Report **

Install and upgrade pages not RTL compatible.
http://dev.simplemachines.org/mantis/view.php?id=4749

PR #1685, opened 2014-05-27
Install and Upgrade RTL Compat

matthew-kerle:
Added right to left support on both the install and upgrade scripts. Fixed an undefined error and a stray indentation difference. Pretty short and sweet.

Signed Off: Matthew P. Kerle / Labradoodle-360 creativewebdsign@gmail.com


matthew-kerle:
This fixes: https://github.com/SimpleMachines/SMF2.1/issues/1203

PR #1689, opened 2014-05-28
Update Load.php

butchs:
Editing the way cache files determine when they are expired.  My tests showed that this is more reliable than the old method.  (I hope I can signoff this with TenForFox?)

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -2853,7 +2853,7 @@ function cache_put_data($key, $value, $ttl = 120)
 				@unlink($cachedir . '/data_' . $key . '.php');
 			else
 			{
-				$cache_data = '<' . '?' . 'php if (!defined(\'SMF\')) die; if (' . (time() + $ttl) . ' < time()) $expired = true; else{$expired = false; $value = \'' . addcslashes($value, '\\\'') . '\';}' . '?' . '>';
+				$cache_data = '<' . '?' . 'php if (!defined(\'SMF\')) die; $duration = \'' . (time() + $ttl) . '\'; $value = \'' . addcslashes($value, '\\\'') . '\';}' . '?' . '>';
 
 				// Write out the cache file, check that the cache write was successful; all the data must be written
 				// If it fails due to low diskspace, or other, remove the cache file
@@ -2930,13 +2930,19 @@ function cache_get_data($key, $ttl = 120)
 			break;
 		default:
 			// Otherwise it's SMF data!
-			if (file_exists($cachedir . '/data_' . $key . '.php') && filesize($cachedir . '/data_' . $key . '.php') > 10)
+			if (is_file($cachedir . '/data_' . $key . '.php') && is_readable($cachedir . '/data_' . $key . '.php') && filesize($cachedir . '/data_' . $key . '.php') > 10)
 			{
+				$expired = true;
 				// php will cache file_exists et all, we can't 100% depend on its results so proceed with caution
 				@include($cachedir . '/data_' . $key . '.php');
-				if (!empty($expired) && isset($value))
-				{
-					@unlink($cachedir . '/data_' . $key . '.php');
+				if ($duration === null) $duration = 0;
+					if((int)$duration > time()) {
+						$expired = false;
+					}
+				unset($duration);
+
+				if ((bool)$expired !== false && isset($value)) {
+					@unlink($cache_filename);
 					unset($value);
 				}
 			}
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -2933,15 +2933,15 @@ function cache_get_data($key, $ttl = 120)
 			if (is_file($cachedir . '/data_' . $key . '.php') && is_readable($cachedir . '/data_' . $key . '.php') && filesize($cachedir . '/data_' . $key . '.php') > 10)
 			{
 				$expired = true;
-				// php will cache file_exists et all, we can't 100% depend on its results so proceed with caution
-				@include($cachedir . '/data_' . $key . '.php');
+				// php will cache file_exists et all.
+				include($cachedir . '/data_' . $key . '.php');
 				if ($duration === null) $duration = 0;
-					if((int)$duration > time()) {
+					if((int) $duration > time()) {
 						$expired = false;
 					}
 				unset($duration);
 
-				if ((bool)$expired !== false && isset($value)) {
+				if ((bool) $expired !== false && isset($value)) {
 					@unlink($cache_filename);
 					unset($value);
 				}


live627:

butchs:
I was working on it...

Dragooon:
You have to sign-off the commit, not the PR.

NanoSector:
Also, please use proper indentation.

``` php
if((int)$duration > time()) {
```

Please use a space there as well.

butchs:
I called it "$allotted" in my mod but thought you would not like that here...

So, if you do not like what I call "$duration" then advise.

If not, I never claimed to be a classically trained programmer, nor do I claim to have any training, so please elaborate.

NanoSector:
It's okay. This happens to the best of us :)

Your $duration is fine, I meant a space between the (int) and $duration.

Arantor:
You don't need is_file and is_readable (you only really need is_readable). If you're casting $expired to a bool, you don't need to do typesafe comparison on it. Unsetting variables especially ones that are going to fall out of scope shortly is not necessary and is often not a good idea. Trying to pre-empt the GC is surprisingly prone to bad things happening.

As far as spaces go, that's all about following the established coding standards throughout SMF. Like using Allman style braces rather than K&R style, and not adding braces for a one line if-branch.

I am curious as to why the move of the if() test from the cache file to the core, which is the fundamental point of all of this. I'm really not sure what benefit it actually has in practice, unless you're doing your own caching and creating way way way too many cache items where disk space is actually potentially a concern.

butchs:
I will take the logical order with your questions and start from the bottom.  The core cache test says that is is not reliable.  Since I do not want to deal with other caches due, to reasons that I will not discuss in this thread, I have been working with disk cache.  No there were not that many files in my test.  In fact, only two or three at a time...  

As I was testing various file names with strategic debugging code I discovered that the standard SMF cache randomly failed before the cache expiration time.  To make a long story short, I first tried to keep the "if" statement but it kept failing.  The code, albeit overkill, is what worked consistently.  Since then I tested it with hundreds of files with no failures.

I never  met Allman or K&R?  They sound like nice chaps.

is_file & is_readable:  Good practice per IBM (http://www.ibm.com/developerworks/library/os-php-readfiles/index.html) recommends to use: "file_exists" and "is_readable".
I replaced "file_exists" with "is_file" because "Is_file" is faster.  But answer 15 at (http://stackoverflow.com/questions/792899/is-file-or-file-exists-in-php#792909) states otherwise.
So I ended up with "is_file", "is_readable" and (filesize > 10).  The latter was not looked at.

Unlink will completely delete the file but unset deletes the cache content.   I unset "duration" just because I wanted to make sure it was gone.  I left the "value" there because it was there before and is not unset when unlinked.  Both "duration" and "value" are part of the cache file.  Here is a nice read on the difference (http://phppot.com/php/php-unlink-vs-unset/).

Maybe you have a point with "pre-empt the GC" by using variables that fall out of scope.  I was not sure if they will fall out of scope so I did what I did because I assumed otherwise.  What I did worked so, I moved on...  Thought I would share it here...  You are welcome to try and TEST other methods?  I for one will like to see the effect on the GC...

Arantor:
Firstly, I think you misunderstand the context of 'reliable' in that context. The specific point that it refers to in terms of reliability is with regards to the file_exists part, not the file's contents. file_exists is served from the stat cache (and so is is_file and is_readable for that matter). The actual expiration test itself is not considered unreliable and as far as I know did not need replacing. As a rule: if you do not understand the context of why a piece of code is what it is, do not just change it based on an assumption. Clarify it first.

There are several circumstances I can think of where the cache will return false for data legitimately even though the TTL has not expired. This is not a fault. The entire point of a cache is that it is not supposed to be a reliable data storage replacement. What really concerns me is that code has been replaced on the assumption that previous code didn't work properly without actually finding out the cause. I suspect this is the reason you actually feel you must use the file cache because you seem to be under the impression it is a persistent cache; it is not and should be removed during the daily maintenance run in any event.

Allman or K&R are coding styles. Allman style is used throughout SMF (braces on their own lines) while K&R (braces on the line to which it relates) is not. I am a strong proponent of code quality and have been extremely critical of the overall decline in code quality in SMF over the years. Following establishing coding style and standards is a small but important facet of that, and honestly it doesn't matter whether you prefer one or the other: SMF uses Allman style, and any contributions that do not follow SMF's coding style should generally be rejected.

The thing about is_file and is_readable is that you can generally rely on it not being a directory. The names of things are not entirely predictable meaning that it would be interesting to see a genuine case where a directory appeared, meaning that it's possibly an edge case you do not need to be concerned about.

I do not understand where you brought unlink into this, presumably you misread my comments? I made no mention of it as it is not directly relevant.

I also found a bug in your code. Your code tests that $duration is null; if the file was not loaded, $duration will be undefined and will throw an error accordingly, though it will indeed match null. POC, on the assumption that nonexistantfilewithsomevariablein.php would declare $somevariable if it existed...

```
<?php

@include('nonexistantfilewithsomevariablein.php');
if ($somevariable === null)
    echo 'was null after all!';

?>
```

results in:

```
( ! ) Notice: Undefined variable: somevariable in C:\wamp\www\null.php on line 4
Call Stack
#   Time    Memory  Function    Location
1   0.0003  371944  {main}( )   ..\null.php:0
was null after all!
```

You must not rely on the file being included because there are times it will not be. It's a breeding ground for race conditions - multiple almost-concurrent requests, the first one unlinks an expired cache, but because of statcache, it might still be flagged as existing and passing both is_file and is_readable; this is why it is @ with include rather than the older require code. You must therefore assume any and all variables declared in it will actually not be declared.

I still have no idea why you are outputting it as a string in the first place and then casting it to an int. If you want to use it as an int, output it as an int in the first place. You can still happily do a type-sensitive comparison against null (or even do an isset which should return false on null values)

Lastly about the GC: the thing is, I don't think you understand what the effects of doing so actually are from a systems engineering point of view, and while I haven't benchmarked it, I don't need to drop a hammer to know that it's going to fall down.

OK, here's the deal. Inside the bowels of PHP, there is a beast called the symbol table. This is effectively (though not literally) an array containing the lists of every variable and every value of those variables. When you unset from the table, you're effectively fragmenting the table manually. There are times it makes sense to do so - typically if you have large blocks of data in there, where it's beneficial to dump it as soon as you're done with it.

The problem is, you're trying to reclaim memory but there's no guarantee you actually _will_. Ultimately it's still going to go through the GC that will rebuild the symbol table. It will fragment the table more, and encourage the GC to reclaim that memory sooner, even if the CPU would be better off not doing so at that point in time. It will, guaranteed, do so at end of script anyway when all memory is returned to the OS. In plenty of cases the GC won't even run until the end of script depending on how big you're getting and how deeply nested you're going with other things (e.g. function stack). The danger is that you end up pushing the GC to reclaim memory more often than it actually needs to, by telling it there's potentially memory it can reclaim.

It's a trade off between CPU and memory. Most of the time you're simply not using enough memory to justify trying to pre-empt it like that. As for benchmarking it, there are two problems. Firstly, you can't actually measure the things at work - you can't reliably predict if or when the GC will run, as it's something in the bowels of the Zend engine, not in the PHP upper core, and secondly you're also guaranteed to be subject to observer effect where any attempt you make at PHP level to attempt to benchmark it will ultimately affect the benchmarks themselves.

Dragooon:
To add to what Arantor said, how did you test this to find it more reliable?

butchs:
As you know I have been working on IPv6 integration with my forum firewall mod.  One of the problems was with my mostly php (Java only for support) Challenge page for IP addresses that did not pass  and suspect Countries.

After some consideration of the the naming conventions I decided to move from sessions to disk cache.  The cache needed to be secure and not slow things down.  

I tested it in my offline MAMP test server.  I do not have any fancy dancey testing software like Arantor.  I test and validate each and every variable/ function live manually just like an old assembly debugger.  I print it if I can, I store the data in a traceable method or stop the code.    Change this and that until I am happy.

Since this is not work and I am not making CSV, ISO or GMP documentation I do not always keep records.  I kept most of lone IPv6 test files but the cache was edited directly in the code, there were so many changes and in this case I did not keep the legacy. 

From memory, I added an echo statement to the point where the cache was to be checked, loaded and saved that said something like 
echo '********\*  CHECKING CACHE *******(BR)'; 
Then I added echos to key variables like $value and $expired.  Something like 
echo $value . ' <-----  Current cache value(BR)';

For Cache_put:
echo time() + $ttl . ' <----  cache Put expired time(BR);
echo time() . ' <---  cache put start time(BR)';
echo $ttl . ' <---  cache put start ttl(BR)';

For cache_get I did something like 
if ($expired !== false) echo '************\*  CACHE EXPIRED  *******(BR)';
echo $value ' <-----  Cache value(BR)';
echo $time ' <-----  Current time(BR);

I made my own a cache check function to see if the cache file is there.  If not, I did something like:
echo '********\*  CACHE FILE MISSING *******(BR); 

Basically I echoed all major functions/ variables.  If I could not get it to display I manually stopped the code at a strategic point.

Each page of the data was copied and pasted in a file.  The values were manually added, studied them and confirmed.  Then I waited a few days to clear the head and checked again to see if I missed something.  Repeated until the results were acceptable.

Good grief....  Why do you want this kind of detail?

Arantor:
I want this kind of detail because changing a core function seemingly on a whim is bad practice! If it's actually broken, the correct procedure would have been to file a bug report with appropriate details, because peer review is a good thing. Let someone else review it, see if it's reproducible, and if it is, _then_ do something about it. Establish that it actually is not working by its design - or that someone who isn't you thinks there's a design flaw - before trying to 'fix' it.

It's one of the reasons I bother reading the stuff in here because at some point, someone has to care about code quality and it seems like there's few enough people that do as it is!

As far as 'fancy tools', I'm just using XDebug, which has been the standard PHP debug extension for some years (and is bundled with WampServer amongst others, though not MAMP which is a shame because it is extremely useful). I find it extremely useful for getting call stacks for debugging purposes. You should try it rather than making a snarky comment. It's not my fault you choose not to use tools available for free that would help you.

Mind you, we're dealing with a team that routinely adds new commits without any kind of testing, so frankly this is still probably an improvement! But do not assume that I'm being a hard-ass for you specifically. I am this way with anyone who doesn't follow good practice for software development. I am even more scathing in private as some people here will know only too well.

butchs:
The comment in the php code that says "we can't 100% depend on its results so proceed with caution" kinda sorta tells me that it has been established that it actually is not working by design.  The fact that it has been this way for some time leads me to believe that some do not care.  Originally I planned not to share but I had nightmares...  And...  No I will not describe my nightmares.  ;)

You forgot dancey.

Kindred-999:
As PM, I have to agree with Arantor here...
we have plans to completely redesign for 3.0.
For 2.1 - Unless something is actually broken or there is a clear code reason - redesigning a core function that works should be considered carefully and documented closely.

Arantor:
Can't depend on its result... let's look at the FULL comment:

```
// php will cache file_exists et all, we can't 100% depend on its results so proceed with caution
```

That's the point. file_exists' value is served from the statcache. By definition it is not a guarantee of the file actually existing. Nothing more or less is meant by that comment. Nothing to do with reliability about anything else, simply that file_exists is not and should not be considered absolutely guaranteed to be accurate.

PR #1690, opened 2014-05-28
Update Security.php

VBGAMER45:
Save extra queries on page load rebuildModCache(); does not need to be called by guests.

### DIFF
--- a/Sources/Security.php
+++ b/Sources/Security.php
@@ -487,8 +487,11 @@ function banPermissions()
 		$user_info['mod_cache'] = $_SESSION['mc'];
 	else
 	{
-		require_once($sourcedir . '/Subs-Auth.php');
-		rebuildModCache();
+		if (!$user_info['is_guest'])
+		{
+			require_once($sourcedir . '/Subs-Auth.php');
+			rebuildModCache();
+		}
 	}
 
 	// Now that we have the mod cache taken care of lets setup a cache for the number of mod reports still open


live627:

jdarwood007:
While usually true live627, I think we can let ones like this in.  The amount of code added/changed is fairly small and fairly difficult to copyright a simple if statement.

VBGAMER45:
What is the process for signed off? You can use any of my code for SMF 2.1
and any future version under any license.

On Wed, May 28, 2014 at 10:40 AM, Jeremy D notifications@github.com wrote:

> While usually true live627, I think we can let ones like this in. The
> amount of code added/changed is fairly small and fairly difficult to
> copyright a simple if statement.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/SimpleMachines/SMF2.1/pull/1690#issuecomment-44423859
> .

## 

---

Free Forum Hosting for your community needs
http://www.createaforum.com
Community and Forum Solutions
http://www.ezcommunitysuite.com

Oldiesmann:
The process for signing off varies depending on what you're using to commit the code. For command line, just add the -s flag to the commit command (git commit -asm "message"). For graphical clients (TortoiseGit, SourceTree, GitHub for Windows, etc.) there's usually a box you can check to do it.

VBGAMER45:
So what's next. Should I download this repo and go for there?

I don't see a way in the web interface to signoff...

VBGAMER45:
I am going to withdraw this one...Looks like that function is required to be called/loaded by guests but will summit a different patch to reduce the querys used.

PR #1695, opened 2014-05-28
SQL Query not needed for guests since they can not be a group moderator

VBGAMER45:
SQL Query not needed for guests since they can not be a group moderator
Signed-off-by: vbgamer45 vbgamer45@gmail.com


Dragooon:
Shouldn't this also be done for board moderators?

Oldiesmann:
Probably should @Dragooon. Definitely something we should look into at least.

PR #1697, opened 2014-05-28
Likes perm

### LINKED ISSUE #1678: download packages page broken
looks like the section under download new packages is not showing

MissAllSunday:
- Adds json response support to Likes class
- Add a collapsed icon on Download new packages sectio fixes #1678
- Adds and implements likes permissions and a master setting.
- Removes karma feature.

Don't really like how the like permissions were named, specially the likes_like.
Didn't add a real description for the likes admin setting, not good at it.


XinYenFon:
We are going to offer Karma as a modification, we need to put a checkbox to confirm they want to remove karma table, otherwise that could be really wrong. All their data will be wiped.

Dragooon:
Did a quick read and most of it looks fine, but I agree with @Antes, we shouldn't remove existing Karma data in case someone wants to keep that.

Also the mentions placeholders are for me, correct?

Dragooon:
Eh...I guess my recent changes are causing merge conflicts with this PR

Sources/PackageGet.php
```diff
@@ -174,7 +174,24 @@ function PackageServers()
...
-		$(\'.new_package_content\').css(\'display\') == \'none\' ? $(\'.new_package_content\').show(\'slow\') : $(\'.new_package_content\').hide(\'slow\'); 
+		var collapseState = $(\'.new_package_content\').css(\'display\');
+		var icon = $(\'.download_new_package\').children(\'span\');
+		var collapsedDiv = $(\'.new_package_content\');
+
+		if (collapseState == \'none\')
+		{
+			collapsedDiv.show(\'slow\');
+			icon.removeClass(\'toggle_down\').addClass(\'toggle_up\');
+			icon.prop(\'title\', '. JavaScriptEscape($txt['hide']) .');
```
live627, line 185: This feels weird. Shouldn't it use attr() instead?

Dragooon, line 185: .prop() and .attr() essentially do the same thing but .prop() is generally the recommended way of doing things since it deals with Javascript properties instead of HTML attributes (checkout the jQuery docs for .prop() to know more)

MissAllSunday:
Moved the karma deleting stuff to upgrade.php, added a checkbox and fixed the merge conflict.

Dragooon:
So...this is good for merge?

MissAllSunday:
yep

MissAllSunday:
oops, just found out I made a huge boo boo 

The queries for removing the karma settings are all borked, I didn't get the actual error because for some reason php_version_check() is actually returning false on my php version: 5.3.13

I'll create another PR

PR #1698, opened 2014-05-28
Convert create_functions to true anonymous functions/closures

Dragooon:
Making another PR since i kinda messed up the last one, with the recent bump in minimum PHP requirement as PHP 5.3, we can use closures and true anonymous functions, which are better, more readable and less memory intensive.

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -133,10 +133,26 @@ function reloadSettings()
 		},
 		'strtolower' => $utf8 ? function ($string)
 		{
+			global $sourcedir;
+
+			if (!function_exists('mb_strtolower'))
+			{
+				require_once($sourcedir . '/Subs-Charset.php');
+				return utf8_strtolower($string);
+			}
+
 			return mb_strtolower($string, 'UTF-8');
 		} : 'strtolower',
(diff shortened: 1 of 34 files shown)


Sources/Load.php
```diff
@@ -109,38 +121,55 @@ function reloadSettings()
...
+			}
+		},
+		'substr' => function ($string, $start, $length = null) use ($utf8, $ent_check, $modSettings)
+		{
+			$ent_arr = preg_split('~(&#' . (empty($modSettings['disableEntityCheck']) ? '\d{1,7}' : '021') . ';|&quot;|&amp;|&lt;|&gt;|&nbsp;|.)~' . ($utf8 ? 'u' : '') . '', $ent_check($string), -1, PREG_SPLIT_DELIM_CAPTURE | PREG_SPLIT_NO_EMPTY);
+			return $length === null ? implode('', array_slice($ent_arr, $start)) : implode('', array_slice($ent_arr, $start, $length));
+		},
+		'strtolower' => $utf8 ? function ($string) use ($sourcedir)
+		{
+			if (!function_exists('mb_strtolower'))
```
Daniel15, line 133: This check will run every time `strtolower` is called, you probably want to just check it once.

PR #1705, opened 2014-05-29
Mass rename user_likes -> member_likes for harmony

live627:


Dragooon:
Oh man...:P, I'm holding these renames for the moment until we've merged rest of the likes and mentions stuff, then I'll see.

PR #1711, opened 2014-05-29
Bump required PHP version to 5.3.7

Dragooon:


Arantor:
I would not require 5.3.7. I would require 5.3.8. Specifically because I know crypt() is used internally for things and 5.3.7's implementation is broken.

See http://php.net/releases/5_3_8.php

Dragooon:
Only md5 is broken in 5.3.7, we only use blowfish.

Arantor:
There are instances of crypt() in the bowels of SMF unrelated to password hashing.

Dragooon:
Ah crap you're right

PR #1730, opened 2014-05-31
Fix board index always showing "Hide" on collapsing category

NanoSector:


NanoSector:
Antes will commit this one.

PR #1732, opened 2014-05-31
! Removed some unused image files

XinYenFon:
! Fixed the Quote button radius (effets of last-child)


Arantor:
As per comment on commit, admin/openid.png is actually in use and cannot just be removed.

XinYenFon:
seems those parts are not using CSS :P I'll revert it now.

PR #1741, opened 2014-06-03
Update Subs-Graphics.php

butchs:
I did things differently because of this error:  PHP Fatal error:  Allowed memory size of 33554432 bytes exhausted (tried to allocate 36 bytes) in .../Sources/Subs-Graphics.php on line 300 using SMF 2.0.

Logic:
1.  This functions checks if GD is installed and adds a flag for GD2.  I wanted to get rid of the array created with get_extension_funcs so I replaced it with extension_loaded.  I did not look for a functions_exists GD function for this feature to avoid a fight. Besides extension_loaded seemed to be a good fit.  :0
2.  function_exists is all you need to check for GD2 (ref https://api.drupal.org/api/drupal/modules!system!image.gd.inc/function/image_gd_check_settings/7).  If you look it up you will find that their original code was a slightly improved version of this function.

### DIFF
--- a/Sources/Subs-Graphics.php
+++ b/Sources/Subs-Graphics.php
@@ -242,16 +242,10 @@ function checkImageContents($fileName, $extensiveCheck = false)
  */
 function checkGD()
 {
-	global $gd2;
-
-	// Ddetermine if GD2 is installed and store it in a global.
-	$gd2 = function_exists('imagecreatetruecolor');
-		
-	// Check to see if GD is installed.
-	if (function_exists('extension_loaded') && extension_loaded('gd'))
-		return true;
-
-	return false;
+    global $gd2;
+    // imagecreate is in both versions of GD, but imagecreatetruecolor only got added in GD2.
+    $gd2 = function_exists('imagecreatetruecolor');
+    return $gd2 || function_exists('imagecreate');
 }
 
 /**
--- a/Sources/Subs-Graphics.php
+++ b/Sources/Subs-Graphics.php
@@ -244,15 +244,13 @@ function checkGD()
 {
 	global $gd2;
 
-	// Check to see if GD is installed and what version.
-	if (function_exists('extension_loaded') && extension_loaded('gd')) {
-
-		// Also determine if GD2 is installed and store it in a global.
-		$gd2 = function_exists('imagecreatetruecolor');
+	// Ddetermine if GD2 is installed and store it in a global.
+	$gd2 = function_exists('imagecreatetruecolor');
 		
+	// Check to see if GD is installed.
+	if (function_exists('extension_loaded') && extension_loaded('gd'))
 		return true;
-	}
-	$gd2 = false;
+
 	return false;
 }
 
--- a/Sources/Subs-Graphics.php
+++ b/Sources/Subs-Graphics.php
@@ -245,7 +245,7 @@ function checkGD()
 	global $gd2;
 
 	// Check to see if GD is installed and what version.
-	if (function_exists(extension_loaded) && extension_loaded('gd')) {
+	if (function_exists('extension_loaded') && extension_loaded('gd')) {
 
 		// Also determine if GD2 is installed and store it in a global.
 		$gd2 = function_exists('imagecreatetruecolor');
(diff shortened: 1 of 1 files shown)


Arantor:
OK, so we still have the different braces style to how SMF does it.

We also have code that will generate errors because PHP's parser will assume that extension_loaded is a constant of value extension_loaded. It'll _work_ but it'll throw an error, as it should be a quoted string. Slightly concerned this wasn't picked up in testing.

Seems to me we could simplify all of this. As of PHP 4.3 (bearing in mind we're on 5.3!), checking first for imagecreate being available would cover both GD and GD2, and then checking for imagecreatetruecolor just for the GD2 test. Keep the code as simple as possible if changes are required. And doesn't require any memory allocation to handle; no array of available functions (which potentially might be very very many in 2.0.7!) is required.

Of course we could just make GD2 a dependency in the installer and never have to test for it after that; a host that doesn't support GD2 is not a host you want to be staying on. This isn't 2002 any more.

butchs:
Yes imagecreate was for both only because I was simply duplicating the original function.  GD2 dependency is not my decision.  But if you want just GD2 then a simple function_exists('imagecreatetruecolor') should be all you need.

Arantor:
That's the thing, the routine has to separate GD and GD2 detection as it currently stands.

butchs:
ok.

Dragooon:
Why do you need to check for extension_loaded if imagecreatefromtruecolor is already defined? Shouldn't that be enough?

Arantor:
The bowels of the CAPTCHA differentiate between GD and GD2 (hence the very existence of $gd2), but if imagecreate exists and imagecreatetrucolor does not, you got GD1... I was thinking more like:

```
function checkGD()
{
    global $gd2;
    // imagecreate is in both versions of GD, but imagecreatetruecolor only got added in GD2.
    $gd2 = function_exists('imagecreatetruecolor');
    return $gd2 || function_exists('imagecreate');
}
```

Dragooon:
Can you get GD1 with PHP 5.3?

Arantor:
In theory, yes, though I've never seen it. (Edited comment above to add better version of function!)

But of course you could just streamline it all to make GD2 a dependency rather than an optional extra (which is no big deal, if hosts are running PHP 5.3 they're almost certainly running GD2) and then test it in the installer.

As a bonus you'd get to drop all the Themes/default/fonts/\* folders (like Themes/default/fonts/anon)

butchs:
I looked into this more and I discovered that GD older than version 2 had **imagecreattruecolor**.  If php was compiled with a library older than GD2 it will throw an error and state that **imagecreattruecolor** requires GD2 to work.

I am no expert on this subject but, how about "imagegd2", which will require at least GD 2.01?

> function checkGD()
> {
>     global $gd2;
>     // imagecreate is in both versions of GD, but imagecreatetruecolor only got added in GD2.
>     $gd2 = **function_exists('imagegd2')**;
>     return $gd2 || function_exists('imagecreate');
> }

Arantor:
There are only a few places where GD vs GD2 is actually checked. Virtually all of them are for using imagecreatetruecolor vs imagecreate.

If you wanted to be more thorough you could check existence of imagecreatetruecolor and imagecopyresampled which are the things SMF actually depends on existing - but over the last 10 years, if a system didn't have imagecreatetruecolor, it didn't have anything else of GD2 either, and it's not been a problem in that time from a function availability perspective.

It's only even an issue now for a totally unrelated situation - that trying to get the list of available things uses too much memory, not because any of the underlying assumptions have actually changed.

butchs:
Assuming SMF will never change what we do I am ok with Arantors proposed June 6 edit.  If all agree, I will update the commit.

Arantor:
I'd honestly prefer dropping GD1 support, going with GD2 + ttf support fully, and taking the opportunity to drop a raft of legacy stuff, but in absence of that, agreed ;)

Oldiesmann:
There should be a blank line after the "global" line and also before "return", but looks good otherwise.

Arantor:
And spaces vs tabs?

Issue #1740, opened 2014-06-02
SMF Error's during clean SQLite3 install


tinoest:
```
[Mon Jun 02 20:52:20 2014] [error] [client 127.0.0.1] PHP Fatal error:  Uncaught exception 'Exception' with message 'Unable to open database: unable to open database file' in /Users/tino/SMF2.1/Sources/Subs-Db-sqlite3.php:65\nStack trace:\n#0 /Users/tino/SMF2.1/Sources/Subs-Db-sqlite3.php(65): SQLite3->__construct('/Library/Docume...')\n#1 /Users/tino/SMF2.1/Sources/Load.php(2732): smf_db_initiate('localhost', '/Library/Docume...', '', '', 'smf_', Array)\n#2 /Users/tino/SMF2.1/index.php(63): loadDatabase()\n#3 {main}\n  thrown in /Users/tino/SMF2.1/Sources/Subs-Db-sqlite3.php on line 65, referer: http://127.0.0.1/SMF2.1/index.php
```

The above just resulted in a blank screen via the web browser, I checked the logs and got the above and deduced that it was due to permissions. If something could be returned to the screen that would be useful. 

```
[Mon Jun 02 20:57:11 2014] [error] [client 127.0.0.1] PHP Fatal error:  Allowed memory size of 268435456 bytes exhausted (tried to allocate 130968 bytes) in /Users/tino/SMF2.1/Sources/Subs-Db-sqlite3.php on line 1154, referer: http://127.0.0.1/SMF2.1/install.php?step=5
```

It was previously complaining at 128M , I don't see how it needs that much memory to do anything mind. I can't allocate any more as the VM only has 384M allocated to it.

tinoest:
```
[Mon Jun 02 21:09:20 2014] [error] [client 127.0.0.1] PHP Fatal error:  Call to a member function finalize() on a non-object in /Library/WebServer/Documents/SMF2.1/Sources/Subs-Db-sqlite3.php on line 747, referer: http://127.0.0.1/SMF2.1/install.php?step=2
```

I forgot the above first time round, not sure what's calling that at the moment.

Oldiesmann:
We can probably add code to catch common exceptions like that, but the fact that the error wasn't displayed at all is due to the PHP configuration - specifically PHP's "display_errors" setting. In most cases you don't want errors displayed to the end user like that.

[...]

PR #1756, opened 2014-06-04
SMF Error's during clean SQLite3 install (#1740)

tinoest:
Added check for if the $handle is a object

### DIFF
--- a/Sources/Subs-Db-sqlite3.php
+++ b/Sources/Subs-Db-sqlite3.php
@@ -305,6 +305,10 @@ function smf_db_quote($db_string, $db_values, $connection = null)
  */
 function smf_db_fetch_array($handle)
 {
+
+	if(!is_object($handle))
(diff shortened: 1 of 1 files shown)


Oldiesmann:
You forgot to sign off. Also, there shouldn't be a new line at the end of the file.

tinoest:
Removed that line, its automatically added when I use vim. 

By sign off I guess you mean the following? 

Signed-off-by: tinoest

tinoest:
Grrrrrr, its being added automatically every time I use the github client I have. Its not vim as  I thought it was.

Oldiesmann:
No, I mean actually signing off the commits, in the GitHub client. Generally you can check a box to do it, or you can just type "Signed-off-by: <username> email@adddress.here" in the commit message. I'll accept the commits this time but in the future things need to be signed off properly.

PR #1773, opened 2014-06-06
Basic Markdown Support for Modifications

matthew-kerle:
- Basic markdown support for modifications which allows mod authors to
  have their readme parsed using the GitHub famous markdown.

Signed-Off: Matthew P. Kerle creativewebdsign@gmail.com


live627:
This seems like a very nice thing to have.

matthew-kerle:
Yeah, I agree. I think that it's kind of annoying to have to format a readme using BBC when mod authors are starting to move more and more towards having GitHub repositories (or at least for development) and already have a nicely formatted readme there. I think they come out more modern looking anyway, and a lot easier to write on the fly.

It's a very simple thing to hook in, and I think SMF 2.1 would be a nice time to pull it in. I would definitely use it with my modifications.

The one thing that might have to be fixed here is attaching the code's license somewhere. It's supposed to be in the directory, but I'm not a fan of putting a license file in ./Sources, so I was going to see if people do want to implement this in SMF 2.1, how the licensing should be handled (can we just put a docblock on the top of the file with the license?)

Arantor:
1. Scope creep. There's already enough to do without adding great chunks of additional code for a sub-feature that few enough people use already.
2. If there's an MD parser in the core, the very next question will be 'can we use it for posts'? Then it'll be 'can we intermix MD and bbcode'? You have no idea the hell that will spawn if you try to do these things.
3. I see from the issue list that there are some HTML5 incompatibilities in Parsedown that would need to be addressed from quality control perspective, assuming quality is still a goal for this project (e.g. issue 156 talks about using style instead of align for aligning table elements; this is not valid HTML5)

Long and short: couldn't recommend for 2.1. For a mythical 3.0, by all means put it on the table.

Dragooon:
Gotta have to agree with Arantor here, MD parser is nice and all but simply for modifications is a no-go and we can't have another type of parser for posts with BBC being already there.

@Arantor: Issue 156 is a few other hooks?

Oldiesmann:
Adding this much code just to support something few mod authors would use really isn't worth the trouble, especially since many would also expect us to start supporting markdown in other places as well.

Arantor:
@Dragooon: Issue 156 of the Parsedown project - https://github.com/erusev/parsedown/issues/156 - as if I'd meant issue 156 here, I'd have linked it appropriately ;)

PR #1774, opened 2014-06-07
* Reworking the page index to be formatted by the template

live627:
- Document the function more
- Fix page indexes in RTL mode (Antes! Bad kitty! >:D
- Jaavascript won’t work right. The event doesn’t seem to update. I can expand the first 50 pages, but the next click adds the first 50 again. Help? :D

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -535,16 +535,18 @@ function updateSettings($changeArray, $update = false)
  * an example is available near the function definition.
  * $pageindex = constructPageIndex($scripturl . '?board=' . $board, $_REQUEST['start'], $num_messages, $maxindex, true);
  *
- * @param string $base_url
- * @param int $start
- * @param int $max_value
- * @param int $num_per_page
- * @param bool $flexible_start = false
- * @param bool $show_prevnext = true
+ * @param string $base_url The basic URL to be used for each link.
+ * @param int &$start The start position, by reference. If this is not a multiple of the number of items per page, it is sanitized to be so and the value will persist upon the function's return.
+ * @param int $max_value The total number of items you are paginating for.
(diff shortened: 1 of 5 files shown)


Dragooon:
Fun question, why not move the whole thing to templates itself? It's simply a page index, doesn't have any significance to source whatsoever.

Arantor:
Oh but it does, incidentally. It has a fun and exciting side property that it sanitises the values being imported. For example, if you're doing something that's 20 per page and you supply a start of 15, constructPageIndex is the one that will be sanitising the value to be at a page boundary rather than an arbitrary start point.

As for styling, moving it to templates is fine for the purposes of offering up better styling - but if sufficient markup is provided, so that most things can be done with CSS, wouldn't that be enough too?

live627:
> Fun question, why not move the whole thing to templates itself?

Wedge did that. IMO, it was a mistake...

Arantor:
I seem to recall I did it. I came to realise it wasn't ideal for this reason. But with suitable classes and whatnot, would that not be enough anyway?

live627:
Maybe it could be rewritten as a class in 3.0, if someone gets tit going.

Arantor:
Don't look at me.

live627:
Definitely not you.

Arantor:
That's a good thing, I don't want the job anyway :tongue: Yes, it probably should be in a class or something in 3.0. Or at least sanely introduced into however the input sanitisation works rather than having two separate jobs done by one thing.

live627:
I know.

Dragooon:
Fair enough I guess, merging this

live627:
Now fix the JS, m'kay?

PR #1786, opened 2014-06-09
 Broken html possible from fatal_error() (fixes #1780)

### LINKED ISSUE #1780: Broken html when not-approved user requests password-change email
image

p.S.

Also -- user actions should not be met with an error response.  
"I'm sorry, user, you can't do that" is not a software error, and shouldn't look like one.

live627:

### DIFF
--- a/Sources/Errors.php
+++ b/Sources/Errors.php
@@ -129,7 +129,7 @@ function fatal_error($error, $log = 'general')
 	if (empty($txt))
 		die($error);
 
-	setup_fatal_error_context($log ? log_error($error, $log) : $error, $error);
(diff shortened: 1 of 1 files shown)


Arantor:
Yes. In SMF terminology, a fatal error is any kind of error where all execution paths will be terminated and divert the user to an error message without any option of displaying a form again. In this case, the fact the account is unapproved will mean trying to do anything with it (other than approving) is going to be kicked back to an error message without any other choices (thus fatal error)

Arantor:
Change the second parameter to null rather than omit it. Or change the signature of setup_fatal_error_context() to have the second parameter default to null. The isset check should still return false on a value of null.

PR #1796, opened 2014-06-09
!disregard message confusing to users

AngelinaBelle:
Clarify what disregard/ignore actually means


illori:
i think this version still needs some work, it has removed the part of the message that lets the user know this applies to all moderators. users in 2.0 have had issues with not knowing this.

Oldiesmann:
I think "Moderators will not see any future reports ..." would be better than "No moderator will see any future reports".

AngelinaBelle:
Howzzat?

Oldiesmann:
Sounds good :+1:

PR #1810, opened 2014-06-12
! install.php php < 5.3 syntax errors

AngelinaBelle:
Now, install.php will notify user that php version is too low.
Closures in install.php replaced with create_function.


XinYenFon:
you should sign-off each commit :/

AngelinaBelle:
Sorry.  I will try again.

Issue #1798, opened 2014-06-10
install.php/upgrade.php should support a lower PHP version than 5.3


Dragooon:
The base install.php/upgrade.php should ideally support PHP 4 or at least PHP 5+, otherwise we won't even be able to tell the user that their PHP version isn't supported.

Arantor:
For the installer, agreed. For the upgrader, no, because by then it's already too late anyway.

Oldiesmann:
The installer already includes code to complain about the PHP version being too low:

``` php
    // Check the PHP version.
    if ((!function_exists('version_compare') || version_compare($GLOBALS['required_php_version'], PHP_VERSION, '>')))
    {
        $incontext['warning'] = $txt['error_php_too_low'];
    }
```

The only thing is it's treated as a warning and not an error, so theoretically you could get past that, but if we at least warn them that their PHP version is too low, we can't be responsible for problems caused by ignoring the warning.

Dragooon:
Installer will throw a parse error for versions below PHP 5.3, essentially
my fault since I introduced closures to fix some other bug.

AngelinaBelle:
Closure is used only twice.  Can it be replaced by create_function and global  $script_url?

Oldiesmann:
If it will still generate an error before they get to that part of the install script, then we should indeed fix it.

Dragooon:
It should just die and show the system requirements, before doing anything
else.

Arantor:
That's just it - you have to remove the use of a closure and reintroduce the old code because it will throw a syntax error before it even gets to do a version test. Already one user has tried to run the installer and it fails because he's using 5.2 or lower.

AngelinaBelle:
I had the same problem on. A free web host.
So,something like:
https://github.com/AngelinaBelle/SMF2.1/compare/patch-4
Needs to be tested, of course.

Arantor:
At a glance I see no reason why that wouldn't work, actually. create_function returns a string that names the function, and $smcFunc['strtolower'] is just fine with variable functions :)

AngelinaBelle:
#1811 addresses issues in install.php
#1812 adresses one of the issues in upgrade.php

However, Upgrade.php still loads QueryString.php before checking php version.

This results in Parse error: syntax error, unexpected T_FUNCTION in /home/a8358262/public_html/smf/Sources/QueryString.php on line 620

PR #1811, opened 2014-06-12
! [Installer] Closures caused syntax errors in install.php for php < 5.3.8

AngelinaBelle:
Use create_function instead, so installer can let user know that php version is too low.

ref: #1798 
ref: #1810 (improperly signed off)


Dragooon:
Your indention is still off :P, you need to use tabs instead of spaces. Also, there should be a space after the first argument, like create_function('$string', 'functionblah'

PR #1812, opened 2014-06-12
! [Upgrade] Closures caused syntax errors in upgrade.php for php < 5.3.8

AngelinaBelle:
ref: #1798 

However, there is still an error: QueryString.php is loaded before the php version check is done, causing syntax error for php v 5.2.

Signed-off-by: Angelina Belle angelinabelle@hotmail.com

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -792,7 +792,7 @@ function redirectLocation($location, $addForm = true)
 // Load all essential data and connect to the DB as this is pre SSI.php
 function loadEssentialData()
 {
-	global $db_server, $db_user, $db_passwd, $db_name, $db_connection, $db_prefix, $db_character_set, $db_type;
+	global $txt, $db_server, $db_user, $db_passwd, $db_name, $db_connection, $db_prefix, $db_character_set, $db_type;
 	global $modSettings, $sourcedir, $smcFunc, $upcontext;
 
 	// Do the non-SSI stuff...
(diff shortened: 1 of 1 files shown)


live627:
Sooo, status?

AngelinaBelle:
I was told this fix was not wanted.

live627:
Mind blown.

Dragooon:
No it's wanted but this fix alone won't fix for < PHP 5.3, a file or two needs to be included after ensuring the requirements are met. Although by this time they've already overwritten all the file so most of the things are messed up but we can still prevent the DB changes

AngelinaBelle:
http://www.simplemachines.org/community/index.php?topic=523980.msg3709746
Neither the PM nor the lead dev weighed in, but seems to be the general consensus among the participants.
This much is tested, and I coded the next bit, but did not test it.

The point seemed to be that, if anyone has gone so far as to upload all the files, their website is already messed up, so it hardly matters that the upgrades fails on a php syntax error rather than gracefully explaining to them that they need to upgrade php, and so it would be a waste of the team 's time to test this.

I don't agree, but it is not worth my time to fight a decision that seems already to have been made.
And it is possible that this is currently seen as a low priority that might be taken up later.  Who knows.  

The other bit of work I did is in https://github.com/AngelinaBelle/SMF2.1/tree/patch-2

AngelinaBelle:
Oops.  Was typing while you posted, dragoon. 
Ok then.  A second pull request #1910  should finish that off.  But I have not tested #1910

BitFros7y:
Why just not require 5.3.8 (or even never?) Every host has at least 5.3.29 as PHP version, just because many people use closures. And most hosts even have 5.4.31 version as its one of the latest stable releases. And current official stable release is 5.5.15 (at least atm). Many frameworks pushed their required version for php to 5.3.8 just because of the closures, so why apply these fix if we can just push php version requirements and use all new stuff in the same time? If you ask me, i would even skip 5.3 and jump right to 5.4 as when the 2.1 comes to beta, PHP 5.3 would be old and 5.4 will be supported everywhere (If it isnt already...)

Dragooon:
@qICEp: We are PHP 5.3.8+, this PR was exclusively for upgrade.php to support PHP 5.0+ so that if an user is upgrading on a server that doesn't have PHP 5.3, it can notify them instead of throwing a parse error.

Oldiesmann:
Due to the way PHP works, this is really the only way we can properly tell them their version is too old. It's ugly, but it's probably the best solution.

Oldiesmann:
Is this something we still want to fix? If so we should probably do so before we release Beta 1.

Dragooon:
Closing this, see #2535

PR #1834, opened 2014-06-17
removed hook call

Underdog-01:
This causes a php error when it turns existing string to an array.
It seems the hooks to the admin section are already added prior to this hook call.
Perhaps there is some editing in place that is half finished?

Either way.. install a mod that uses hooks and test it out for yourself.
Removing the hook resolves the issue and the mod is still added prior to it.
Even if another method is going to be implemented using this hook call, it needs to be fixed so it works.

### DIFF
--- a/Sources/Admin.php
+++ b/Sources/Admin.php
@@ -448,9 +448,6 @@ function AdminMain()
 		}
 	}
 
-	// Let them modify admin areas easily.
-	call_integration_hook('integrate_admin_areas', array(&$admin_areas));
-
 	// Make sure the administrator has a valid session...
 	validateSession();
 


MissAllSunday:
Whats the exact error you saw?

I don't really get from where exactly do you get a "string", $admin_areas has always been an array and your functions needs to treat it as such.

How exactly are you using that hook?

Underdog-01:
Prior to changing anything, add a couple of mods that use hooks and then attempt to access your admin section.

This:
    if (isset($admin_include_data['file']))
        require_once($sourcedir . '/' . $admin_include_data['file']);
... $admin_include_data['file'] becomes an array after it uses the hook call and throws an error with the require command. Also if you remove the hook, mods are still added to the admin so therefore something is awry.

.. you can change the above to a for loop which executes all the necessary requires but then you get a template error. someone changed something in this file recently to add the hook in that place but evidently did not test things because the changes cause an error. Also whatever previously added things to admin is still in place prior to the hook because if you remove the hook tthings are still added.

MissAllSunday:
What are the exact hooks you're using?  and how are you using them?

AngelinaBelle:
Looks like integrate_admin_areas
Someone else had indicated that it was called twice, but I could not find the other call, because I did not look hard enough

In createMenu
        call_integration_hook('integrate_' . $menu_context['current_action'] . '_areas', array(&$menuData));

Underdog-01:
AngelinaBelle,

  Ahh yes, I did not find it just yet so thank you for pointing it out. Since the other hook call works as expected and is already implemented then my suggestion is to remove the one noted in this pull request.

AngelinaBelle:
Quite possibly, a comment in createMenu would help -- one that included the possible hook names that will result, such as integrate_admin_areas etc. so that others may find them when they search.

Another option, before accepting a pull request to add a hook, is to require the requester to test the use of the hook to make sure hooked code is run exactly one time.

MissAllSunday:
Can't really delete the hook if we want to keep compatibility with 2.0 which is something I want to keep as much as possible.

So, a work around here would be changing the actual name of the hooks for call_integration_hook('integrate_' . $menu_context['current_action'] . '_areas', array(&$menuData));  something as simple as:

call_integration_hook('integrate_' . $menu_context['current_action'] . '_menu', array(&$menuData));

Would suffice but theres also the inconvenient that "_menu" doesn't really reflect what the actual hook is doing although I argue that creating entire sections from a function designed to create menus doesn't really have much sense to me.  Plus, the hook allows the creation of subsections and sections as well and not just areas.

AngelinaBelle:
If createMenu is being overloaded with code to create areas, then perhaps a separate function for createArea is required.  If that place in the code (createMenu, or near createMenu) is the wrong place for integrate_admin_areas, then maybe it should be moved to be closer to where it should be.

As for compatibility -- how many mods use that integration hook? I bet it is not easy to search all mods for that string to find the answer to this one.

Oldiesmann:
It's an extremely useful hook regardless of how many mods actually use it. Removing it isn't really an option.

I'm trying to figure out how $admin_include_data['file'] gets turned into an array with a simple call to a hook. I don't see anything in call_integration_hook or call_hook_helper that would do this. Also, what template error do you get if you change the code to loop through it?

Underdog-01:
That hook is being used to add custom mod links to the admin menu array and is necessary although I was reporting that for some reason the hook is being called twice because if I remove it from that specific line the data is still being added to that array.

  With SMF 2.0,  I was using array_merge_recursive to add to a specific part of that array which caused an issue with the changes made previously in this file. I just changed my routine a bit which seems to be working with this hook in place (specific new example within my hook code: $admin_areas['layout']['areas'] += $array;).. An example of the code change I had to make is in my github repo for PersonalizedBBC if you are curious as to what I was previously doing to cause the error. ref. https://github.com/Underdog-01/Personalized_BBC/commit/dd2984a5d4f145cd48656b71aa498f339bcb5654#diff-8ced2fb063aa8f1d53908c843b897e2aL97

  Although if I stumbled upon an issue that needs to be resolved because the hook is being called twice then that is good. However if the intention was to purposely use the hook a second time to use some part of the multidimensional array then perhaps ignore my report.

Regards.

Oldiesmann:
It is indeed being called twice due to the new hook in createMenu() (`call_integration_hook('integrate_' . $menu_context['current_action'] . '_areas', array(&$menuData));`), so the call in Admin.php can be removed now. I was just a bit confused as to what was going on here.

AngelinaBelle:
Please put appropriate comments in CreateMenu so this won't be a mystery the next time it comes up.

MissAllSunday:
Ha! and I just realize how pointless my last comment was Haahahah

AngelinaBelle:
I really just thought you were saying createMenu is a weird place for creating an entire area, and that the hook call should be moved someplace else more sensible.
That is not pointless at all.

The current location, and use of a string makes it actually impossible to find this hook, or the other similar ones, by searching.  So a comment at that point in the code listing which hooks will get called from there by out-of-the-box SMF would be really useful, and probably prevent some kind of double-hooking in the future.

MissAllSunday:
Nop, I meant that it is pointless to say we shouldn't remove the hook for compatibility since its really irrelevant.

PR #1838, opened 2014-06-19
! s and u tags could not be disabled in editor + construct is tidier

Underdog-01:
- changed if/else to array & foreach construct
- if/else had incorrect conditions anyhow since "i" was repeated in error
  (more to come with this file since bbc button array is not working properly atm)

### DIFF
--- a/Sources/Subs-Editor.php
+++ b/Sources/Subs-Editor.php
@@ -1650,6 +1650,18 @@ function create_control_richedit($editorOptions)
 			),
 		);
 
+		$disabled_AltTags = array(
+			'b' => 'bold',
+			'i' => 'italic',
+			'u' => 'underline',
+			's' => 'strike',
+			'img' => 'image',
+			'url' => 'link',
+			'sup' => 'superscript',
+			'sub' => 'subscript',
+			'hr' => 'horizontalrule',
+		);
+
 		// Allow mods to modify BBC buttons.
 		// Note: pass the array here is not necessary and is deprecated, but it is ketp for backward compatibility with 2.0
 		call_integration_hook('integrate_bbc_buttons', array(&$context['bbc_tags']));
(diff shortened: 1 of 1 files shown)


AngelinaBelle:
This fixes the bug of s and u tags not being dealt with properly, and looks like easier-to-maintain code.

Sources/Subs-Editor.php
```diff
@@ -1650,6 +1650,18 @@ function create_control_richedit($editorOptions)
...
+			'img' => 'image',
+			'url' => 'link',
+			'sup' => 'superscript',
+			'sub' => 'subscript',
+			'hr' => 'horizontalrule',
+		);
+
 		// Allow mods to modify BBC buttons.
 		// Note: pass the array here is not necessary and is deprecated, but it is ketp for backward compatibility with 2.0
 		call_integration_hook('integrate_bbc_buttons', array(&$context['bbc_tags']));
```
live627, line 1667: Maybe pass $disabled_AltTags into the hook? Rename it to $disabled_editor_tags?

Underdog-01:
I can change the name of the array used if you think it is more suitable.
I can also add the array to the hook call but I don;t see the point atm since the hooked BBC's show up on the BBC admin page whereas they can be unchecked. The loop that follows should disable them anyway.

live627:
his pull request contains a merge conflict now due to your other PR

Underdog-01:
I edited the code to match the other PR.
If it still gives a conflict let me know & I will close this one and start another without the last part.

live627:
Look for the green icon below this.

XinYenFon:
you didn't sign-off your last two commit :(

Underdog-01:
It won't let me edit the actual title relating to the code edit.
I will just close this one and do it again.

Underdog-01:
keep in mind that I will be posting more PR's for this file after I take the liberty to fix the WYSIWYG editor. the bbc array needs to include the html from the bbc tag (added via hook or otherwise) to have that behave properly but the current code for the WYSIWYG function needs to be reworked first. 
Atm its buggy in SMF 2.1 as is the behavior for SMF 2.0.8... I will look at it when time permits.
Any hooked bbc tags that do not include the html in the array will behave the same... mod authors will have to opt in html to the array to have it behave different for the WYSIWYG editor .. if you follow what I am getting at?

PR #1886, opened 2014-06-23
Gravatars

live627:
Closes  #1160

Can be turned off for all you purists out there.


Dragooon:
I take it this allows entering gratar e-mail instead of hash?

Issue #1881, opened 2014-06-23
upgrade.php still trying to use sqlite for mysqli


illori:
the upgrade.php script is still trying to call sqlite upgrade files for mysqli

PR #1892, opened 2014-06-23
Fix the upgrader so it actually works and doesn't crap on Settings.php

NanoSector:


NanoSector:
This should fix #1881.

PR #1894, opened 2014-06-23
Cust profile field stuff

### LINKED ISSUE #1889: gender none icon used in profile
it is confusing that the gender icon in the profile for the none option is the same as when the user is online. it should be something else so we can see it is different from others used elsewhere in the UI.

also most may not understand what the alt text "none" means as there is nothing in that area to tell that it is the persons gender.

MissAllSunday:
fixes #1889
- Changes option "None" to "Disabled"
- Doesn't print the option if the "Disabled" option is selected.

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -1243,6 +1243,10 @@ function loadMemberContext($user, $display_custom_fields = false)
 
 			$value = $profile['options'][$custom['col_name']];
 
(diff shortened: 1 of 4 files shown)


Dragooon:
Gender: Disabled.....haaaahahahahaahahah

MissAllSunday:
Better than  Sex: Disabled  :tongue:

PR #1911, opened 2014-06-24
Post Buttons reordered, as in SMF 2.0

AngelinaBelle:
Order: Post, Preview, Spell Check, Save Draft

### DIFF
--- a/Themes/default/GenericControls.template.php
+++ b/Themes/default/GenericControls.template.php
@@ -143,26 +143,39 @@ function template_control_richedit_buttons($editor_id)
 	echo '
 		<span class="smalltext">
 			', $context['shortcuts_text'], '
-		</span>
-		<input type="submit" value="', isset($editor_context['labels']['post_button']) ? $editor_context['labels']['post_button'] : $txt['post'], '" tabindex="', $context['tabindex']++, '" onclick="return submitThisOnce(this);" accesskey="s" class="button_submit">';
+		</span>';
 
-	if ($editor_context['preview_type'])
+	if (!empty($context['drafts_pm_save']))
+	{
+		// The PM draft save button
 		echo '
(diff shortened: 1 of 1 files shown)


live627:
- Quick reply also needs this fix.
- Tabindex ordering is backwards. Side effect of using the float.
- The float seems like a hack to me. Could it be replaced with a display of inline-block?
- Why are the buttons outside of the frame?

AngelinaBelle:
- quick reply SHOULD be reordered the same way.  Why isn't template_control_richedit_buttons being reused there?
- It is simple to reorder the tabindex.  See https://github.com/AngelinaBelle/SMF2.1/tree/patch-3. I can do that.
- Float was chosen because it is responsive and reorders properly when RTL.css is loaded.  inline-block was not mentioned for these buttons.  Discussion at http://www.simplemachines.org/community/index.php?topic=523420.msg3704112#msg3704112
- When you say "why are the buttons outside the frame" what do you mean? To me, they look like this:
  image

live627:
> quick reply SHOULD be reordered the same way. Why isn't template_control_richedit_buttons being reused there?

Excceept that it's its woowwn code in the DSisplay template. Probably should reuse the editor code... need more dev input.

> Float was chosen because it is responsive and reorders properly when RTL.css is loaded. inline-block was not mentioned for these buttons. Discussion at http://www.simplemachines.org/community/index.php?topic=523420.msg3704112#msg3704112

So is inline-block, AFAICT.

> When you say "why are the buttons outside the frame" what do you mean?

T'was a bug. I fixed it today.

AngelinaBelle:
I feel competent to add the code to fix the tabindex for the floated version, but not to understand why float was chosen over inline-block.

Glad to hear the "outside frame" was not my bug!

live627:
Can you send that other branch in a pull request? I want to get this done.

AngelinaBelle:
#1942

PR #1918, opened 2014-06-25
Add  proper timezone support

live627:


Oldiesmann:
How hard would it be to implement an "automatically detect DST" thing with this? I realize that DST is a tricky thing to deal with, but it might be a nice feature to add if we can since we have proper timezone support now.

live627:
Meh... this  new coded deals with it already.

PR #1920, opened 2014-06-25
Theme fixes...

XinYenFon:
- img to css (xx.png)
- removed core_post class (windowbg/2 now handle it)
- removed some inline css
- increased padding of roundframe (9px to "12px 16px")


live627:
Spodey senses say new bug is introduced.

Issue #1757, opened 2014-06-04
Weird interaction between account settings and iPad/Safari and iPad/chrome  save password feature.


AngelinaBelle:
Please use html5 form attribute autocomplete="off" to work around several buggy browsers that are over eager to jam a saved username into a form with a password field in it.  

<hr>

This is not a problem in 2.0.7.
When I go to the account settings page, my ipad automatically replaces my email address with the username I used when I logged in.

Since I wanted to fix another field in that page, I decided I needed to fix the email address at the same time.  SMF complained that I had changed my email address and would therefore be logged out.  

But it did not log me out -- presumably because I had changed it to exactly what it had been before.
![image](https://cloud.githubusercontent.com/assets/3267025/3177479/a20f38e8-ec10-11e3-9d9c-dcf4e0605f33.jpg)
![image](https://cloud.githubusercontent.com/assets/3267025/3177477/a20cdc06-ec10-11e3-967d-e29a267b5eac.jpg)
![image](https://cloud.githubusercontent.com/assets/3267025/3177478/a20d2288-ec10-11e3-8968-b3a9563fb2b0.jpg)

![image](https://cloud.githubusercontent.com/assets/3267025/3177481/abfaca98-ec10-11e3-86d3-2199c767d2a7.jpg)

illori:
sounds like this may be a dup of #1191 which should be solved

AngelinaBelle:
Not a dup. The primary issue is that SMF and Safari for iPad are not compatible here.
So SMF is not iOS friendly.  Can any MAC users report in?

Safari thinks the email input field is meant for my username, and so automatically fills in the username it just remembered for me for this site.   It also automatically fills in my password at the bottom of the page.

Can the page tell Safari not to try this trick on this page, or would the fix involve renaming the input field.

The confusing bit about the changed but unchanged email address is just baggage; the email address was changed back to its database value by the time I submitted the form.

illori:
looking at the code for that element i might be willing to say this is an issue with your browser and not smf.

```
<input id="email_address" class="input_text" type="email" value="myemailaddress@emailaddress.com" size="30" name="email_address">
```

AngelinaBelle:
Issue with Safari for iPad. IOS 7.1.1 not compatible with that page. Don't know about chrome.

illori:
[...]

PR #1954, opened 2014-06-27
! iPad autocomplete was replacing email address in account settings

AngelinaBelle:
Fake name and password fields prevent Safari and chrome from replacing email in account settings.


AngelinaBelle:
Fixes #1757

AngelinaBelle:
Please merge to work around the ios over eager autocomplete.  This PR makes SMF much friendlier to iPad users.

Oldiesmann:
Won't those fake inputs be visible to the user?

AngelinaBelle:
They are outside the visible window because of the absolute positioning.  If you use hidden, Safari is smart enough to ignore them.

PR #1962, opened 2014-06-27
allow hooked BBcodes to use WYSIWYG/Source toggle

Underdog-01:
Mod author paramters for the bbc array:
- includes backward compatibility for previous bbc array keys (code, description, before, after)
- mod authors must include code (<-name), description, before (and after if applicable) keys into bbc array
- if they want to use the WYSIWYG editor to toggle their BBCode
  they must include new keys of "html", "allowed_children" (<- defaults to null)
  and "block_lvl" (<- defaults to inline if 0 or not included in array)
- for the WYSIWYG editor to ignore toggling the BBC html,
  the "html" key should be empty or not included in the bbc array
- code works for all BBCode types
- the key of allowed_children defaults to null which allows all children unless otherwise specified within the array
  the # can be entered for this setting which will allow all text but will not execute a BBcode placed within. 
- other default BBC with the custom tag posted within will behave normally.
- setting for allowed_children examples:
  empty or null -> allow all children BBC's,
  "#" will treat other BBC's contained within as text,
  b, url, hr -> execute specific BBC HTML only (separated by commas)
- url's within custom bbc tags will not toggle to url, img, etc.
  unless those specific bbc tags are included ( :) <-- preferred behavior!!!)

---

Explanation of behavior:

I will suggest that some general information about the SCEditor is provided in the SMF wiki
  for mod authors regarding the use of the SCEditor within the SMF infrastructure.  
ie. for custom BBC mods they need to be informed that if they intend to enable WYSIWYG for their mod
  that it should be tested for conflicts with existing BBC's
else give their admin menu an option to enable/disable the feature with specific help text for that option.

This edit takes advantage of css styling elements and an existing sce css class
  to provide the desired effect within the html class function.
Meaning that the existing format class function for custom BBCodes added via the hook is ignored
  as it uses css style/class elements within the html class function for the desired effect.
  It behaves very similar to the existing behavior of the form/html class
  (as executed within ../jquery.sceditor.bbcode.js). 
The author has informed me via email that changes are pending regarding the above type of behavior
  or at least the ability to control it using the format class function (if I understood the response correctly).
This is because currently for SMF 2.1 the default behavior of the editor will change those to [img] and [url] BBC tags
  as it is hard coded into it via functions within jquery.sceditor.smf.js
  although this is not a desired effect for custom BBCodes.

---

Undesired effects of SCEditor within SMF 2.1 (this is default behavior without this edit):
- posting html within posts does not get converted to BBC renderring the usage of source/WYSIWYG toggling
  within the editor as only converting existing BBC to source and back again.
- meaning that people posting html into posts does not get converted to BBC (although smileys text still gets converted)
- disabled BBcodes are still controlled by the SCEditor which is incorrect behavior
  (needs $.sceditor.plugins.bbcode.bbcode.remove("[BBCODE]"); function added) 
- remove formatting option does not seem to function?

Notes:
$.sceditor.plugins.bbcode.bbcode.set(format: ...) (specifically the format class)
  has been removed due to incorrect IE behavior (the html class for this edit takes care of all display).
Adding before, after & html attributes to the BBC array for existing default SMF BBCodes
  will change their behavior (if desired) as they will use the conditions of the edit.

### DIFF
--- a/Sources/Subs-Editor.php
+++ b/Sources/Subs-Editor.php
@@ -1763,13 +1763,6 @@ function create_control_richedit($editorOptions)
 									return \'' . preg_replace(array('~\$1~', '~\$2~', '~\$3~'), array('\' + content + \'', '\' + attributes[0] + \'', '\' + attributes[1] + \''), '<div style="display:inline;" class="sceditor-ignore">' . $tag['html'] . '</div>') . '<span style="display: none;visibility: hidden;">' . str_replace((!empty($attrs[1]) ? $attrs[1] : ''), '\' + attributes[0] + \',\' + attributes[1] + \'', $tag['before']) . '\' + content + \'' . (isset($tag['after']) ? $tag['after'] . '</span>' : '</span>') . '\';
 								else
 									return \'' . preg_replace(array('~\$1~', '~\$2~'), array('\' + content + \'', '\' + attributes[0] + \''), '<div style="display:inline;" class="sceditor-ignore">' . $tag['html'] . '</div>') . '<span style="display: none;visibility: hidden;">' . str_replace((!empty($attrs[1]) ? $attrs[1] : ''), '\' + attributes[0] + \'', $tag['before']) . '\' + content + \'' . (isset($tag['after']) ? $tag['after'] . '</span>' : '</span>') . '\';
-						},
-						format: function($element, content) {
-							var element = $element[0];
-							if (element === "undefined")
-								return \'' . $tag['before'] . '\' + content + \'' . (isset($tag['after']) ? $tag['after'] : '') . '\';
-
-							return \'' . str_replace((!empty($attrs[1]) ? $attrs[1] : ''), '\' + elemement + \'', $tag['before']) . '\' + content + \'' . (isset($tag['after']) ? $tag['after'] : '') . '\';
 						}
 					}
 				);';
--- a/Sources/Subs-Editor.php
+++ b/Sources/Subs-Editor.php
@@ -1715,24 +1715,65 @@ function create_control_richedit($editorOptions)
 				if ((!empty($tag['code'])) && empty($context['disabled_tags'][$tag['code']]))
 				{
 					$tagsRow[] = $tag['code'];
+					$tag['allow_children'] = !empty($tag['allow_children']) && strpos($tag['allow_children'], 'null') === false ? array_map('trim', explode(',', $tag['allow_children'])) : array();
+
 					if (isset($tag['image']))
 						$bbcodes_styles .= '
 			.sceditor-button-' . $tag['code'] . ' div {
 				background: url(\'' . $settings['default_theme_url'] . '/images/bbc/' . $tag['image'] . '.png\');
 			}';
 					if (isset($tag['before']))
-					{
 						$context['bbcodes_handlers'] .= '
 				$.sceditor.command.set(
 					' . javaScriptEscape($tag['code']) . ', {
 					exec: function () {
-						this.wysiwygEditorInsertHtml(' . javaScriptEscape($tag['before']) . (isset($tag['after']) ? ', ' . javaScriptEscape($tag['after']) : '') . ');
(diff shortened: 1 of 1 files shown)


Underdog-01:
One other thing I would like to state is that $1, $2, $3 ($2 & $3 are optional) should be contained within the BBcode html. They represent {content}, {option1}, {option2} in that order.
I believe using $1, $2 and $3 is/was the default behavior for BBcodes but if I am incorrect with that assumption then please let me know.

This edit works very well although I only tested it in 3 popular browsers (IE, Mozilla & Chrome).
Testing in Safari, Opera and mobile browsers is pending.

Regards.

live627:
> remove formatting option does not seem to function?remove formatting option does not seem to function?

SCE thing; only applies to selected text.

Same effect at http://www.sceditor.com/

Underdog-01:
Ok I understand.

I adjusted the edit to not include the format class as it was affecting IE behavior.
With this edit the html class takes care of all display so it was unnecessary anyhow.

An example BBcode mod to use for SMF 2.1 is located here:
http://www.web-develop.ca/index.php?topic=92.msg834#msg834

The above linked mod already includes the necessary data within the BBcode array (except for the restricted internal tags option). If anyone is willing to test other BBC mods that use the hook please do so. 

Although aside from the backward compat behavior (whereas the WYSIWYG/Source toggle has no effect on said BBCode) you will need to adjust those mods to include their html within the BBCode array. 

Regards.

live627:
This is not really working. Code gets sporadically eaten and JS errors get spewed.

Why so complex? Why not simplify html to `html: ' . $tag['html'] .`? Also, format is missing, but should be optional. Maybe you could do the same with all properties? That way, we could point mod authors to the SCE docs for the values...

Underdog-01:
I added allowsEmpty:true, which might have been the cause of your issue if you were testing empty
bbc tags. This actually works well for me when I test it without "eating" any user input although I only tested in the 3 browsers that I stated earlier. 

When I use the js validator in Chrome it appears ok although Firebug in Firefox shows many issues on the page.  Although when I check the page without any of my edits, the Firebug errors still show up so I do not think it is this edit causing them. Did you check for js errors on a default SMF 2.1 edit post page (without this edit) ?

For some odd reason I can not get the format class to work atm... I will work on it and yes I think it is a good idea to pass other various settings via the BBCode array.

Regards.

Oldiesmann:
I'm not seeing any JS errors on my test board using Firefox's built-in dev tools (CTRL+SHIFT+I or Menu -> Developer -> Toggle Tools). I don't have Firebug so I don't know what errors you're getting.

Underdog-01:
I find the SMF [code] display easier to read so I posted the warnings/errors here:
http://www.web-develop.ca/index.php?topic=93.msg835#msg835

This is what Firebug reports while editing a post in the current SMF 2.1 without the edit shown here (default template). These are warnings and not errors although I am not sure what the dev team deems as misdemeanors. 

I am going to try the code using format directly in the js file to see if it works that way because I am puzzled as to why format will not work for me using the $context to add it.

Underdog-01:
I ran the error checks again cycling through HTML, CSS, SCRIPT, DOM, NET and COOKIES just once and then back to CONSOLE to show the issues. I get the same 0 errors and 4 warnings for both.
The script does have an issue regarding getElementById whereas something is not behaving correctly but it does not seem to be related to my edit.

  As I stated, I will work on getting this to work using the format class although Imo the issue may be related to the JS warning.

Oldiesmann:
SMF doesn't have files named "object.js" or "reps.js" so I'm not sure where those are coming from.

Underdog-01:
That's the warnings that firebug displays. 

 I don't know what to make of the behavior I see here for using the $.sceditor.plugins.bbcode.bbcode.set() function. No matter what it seems to not fire the format parameter for me and the only way I seem to be able to make it work is with the trick I did here that uses only the html parameter.   If I attempt to use it as documented, it converts the bbcode to html but not back again. 

Atm I am stumped.

Underdog-01:
Due to recent editor updates and with testing those updates, I see that BBCodes added via the hook are now able to use the WYSIWYG editor therefore I will close this PR.
Nice job as I see it now works quite well.

Issue #1380, opened 2014-02-09
Auto saving draft in QR mode without full editor causes errors


Arantor:
To reproduce, have auto saving drafts enabled, do not use the full editor in QR mode, and go to write a message.

```
Uncaught TypeError: Cannot read property 'contentWindow' of undefined drafts.js?alp21:25
smf_DraftAutoSave.init drafts.js?alp21:25
(anonymous function)
```

This refers to:

```
var oIframeWindow = oIframe.contentWindow || oIframe.contentDocument;
```

but without the full editor defined, oIframe doesn't exist. This also means drafts will not be saved automatically.

live627:
Fixed by #1964

PR #1964, opened 2014-06-28
Remove the option to use the full editor on quick reply

live627:


XinYenFon:
uh?

live627:
you saying you don't like this?

XinYenFon:
Quick reply is for quick typing (so I'm not against) but could be nice to know the reasons (or I'm probably missing the discussion :P).

live627:
I'm trying to find Arantor's report

live627:
#1380
http://www.simplemachines.org/community/index.php?topic=524417.0

illori:
there was no other way to fix the issue then to remove the full editor in quick reply?

illori:
also since i had this set before your change... the buttons are still showing as if you changed nothing.

AngelinaBelle:
This topic has a confusing title. Several people have read it to mean the full editor on quick reply has gone away.
It really means that full editor is always on quick reply.

Oldiesmann:
Isn't there a way we can fix the issues without forcing everyone to use full editor on quick reply? There are plenty of people who would prefer to keep quick reply the way it's always been without the editor.

Oldiesmann:
This has now been reverted in #2011 (though the top bar where the buttons would normally go is still showing with the editor disabled), but spell check and auto-saving of drafts still seem to work fine even with the full editor disabled.

illori:
did you test with your developer tools open and using the console to check for errors?

Oldiesmann:
Just tested it again with dev tools open this time. No errors and both spell check and auto-saving of the draft work fine still.

illori:
i am not getting any errors either. is it expected that we still see the full set of buttons as if the editor is still full?

Oldiesmann:
Yes

illori:
then i guess the issue is resolved

PR #1965, opened 2014-06-28
A few random fixes

live627:


NanoSector:
Your second commit isn't signed off. Otherwise, nice commits :)

PR #1978, opened 2014-06-28
edit fixes JS error/warning

Underdog-01:

### DIFF
--- a/Themes/default/scripts/smf_jquery_plugins.js
+++ b/Themes/default/scripts/smf_jquery_plugins.js
@@ -49,7 +49,7 @@
 			var iPosy = 0;
 			
 			if (!event)
-				var event = window.event;
+				let event = window.event;
 				
 			if (event.pageX || event.pageY)
 			{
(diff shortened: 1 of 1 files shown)


live627:
Can I ask how to reproduce the error?

Underdog-01:
I adjusted the edit.

I use Komodo editor which has a syntax checker. 
The file parameter is set to javascript and then flags all the errors within the file (red underline). 
I think it flagged the warning because event had to be defined as something previously in the onset of the function (not a new var) and the condition is checking if it is false then assigning it as another value. var is only used when assigning a new undefined var.

It also shows other warnings in the file about anonymous functions returning no value whereas a return false; at the end of them fixes it.

NanoSector:
Komodo flags them as strict warnings. They're not required to fix. http://i.imgur.com/WWF3BDD.png

Underdog-01:
I was not sure which misdemeanor warnings are ignored for JS and suppose it depends on each dev's preference. .

live627:
False flag.

PR #1991, opened 2014-06-29
Call helper

### LINKED ISSUE #1895: call static method in ManageSettings.php
SMF 2.0.8
php 5.4

error: Fatal error: Call to undefined function MyClass::mystaticmethod() in ...\Sources\ManageSettings.php on line 198

find:

``` php
    // Call the right function for this sub-acton.
    $subActions[$_REQUEST['sa']]();
```

replace:

``` php
    $func = strpos($subActions[$_REQUEST['sa']], '::') !== false ? explode('::', $subActions[$_REQUEST['sa']]) : $subActions[$_REQUEST['sa']];
    // Call the right function for this sub-acton.
    if (is_array($func))
        call_user_func($func);
    else
        $func();
```

MissAllSunday:
Another round for finding calls to sub-actions.
- My other search missed some variants like $sub_action and $subaction thus a few other places were still using the string directly fixes #1895  /me hopes...
- call_helper() will now return a boolean false/log an error if the given string cannot produce a callable var.
- Removes some is_callable() checks and instead it places an !empty()

### DIFF
--- a/Sources/Admin.php
+++ b/Sources/Admin.php
@@ -512,7 +512,7 @@ function AdminMain()
 		$call = call_helper($admin_include_data['function'], true);
 
 	// Is it valid?
-	if (is_callable($call))
+	if (!empty($call))
 		call_user_func($call);
 }
 
--- a/Sources/Admin.php
+++ b/Sources/Admin.php
@@ -959,7 +959,7 @@ function AdminLogs()
 	);
 
 	require_once($sourcedir . '/' . $log_functions[$subActions][0]);
-	$log_functions[$subActions][1]);
+	call_helper($log_functions[$subActions][1]);
 }
 
 /**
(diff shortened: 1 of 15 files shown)


Sources/Admin.php
```diff
@@ -922,7 +922,7 @@ function AdminLogs()
 
 	call_integration_hook('integrate_manage_logs', array(&$log_functions));
 
-	$sub_action = isset($_REQUEST['sa']) && isset($log_functions[$_REQUEST['sa']]) && empty($log_functions[$_REQUEST['sa']]['disabled']) ? $_REQUEST['sa'] : 'errorlog';
+	$subActions = isset($_REQUEST['sa']) && isset($log_functions[$_REQUEST['sa']]) && empty($log_functions[$_REQUEST['sa']]['disabled']) ? $_REQUEST['sa'] : 'errorlog';
```
live627: Shouldn't this be singular?

PR #2011, opened 2014-07-02
I guess nostalgia trumps logic

live627:
Happy now?

### DIFF
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -1157,6 +1157,7 @@ function Display()
 	$editorOptions = array(
 		'id' => 'message',
 		'value' => '',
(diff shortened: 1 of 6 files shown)


Oldiesmann:
I was hoping we could actually fix the issues caused by having the full editor disabled in quick reply as well...

AngelinaBelle:
Isn't it possible to just hide the fancy stuff for those who want a clean-looking quick reply?
How about offering "clean looking quick reply" as a mod instead of another option?

Oldiesmann:
Going to go ahead and merge this in so we can look at what needs to be fixed with drafts and spellcheck. I'd rather have a proper fix for the issues than forcing this on people (or at least have time to figure out if the issues can be fixed first).

PR #2026, opened 2014-07-04
When making a copy from the default theme, include index.responsive.css

MissAllSunday:
Makes things easier.

If this isn't the intended behavior for index.responsive then we need to force loading the file from the default theme,

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -1846,7 +1846,7 @@ function loadTheme($id_theme = 0, $initialize = true)
 	loadCSSFile('index.css');
 
 	// Here is my luvly Responsive CSS
-	loadCSSFile('index.responsive.css');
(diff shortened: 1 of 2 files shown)


XinYenFon:
Hmm... Actually we can;

1) Merge this
2) Force default
3) Remove file and put inside index.css

But we probably go for 3. But not sure when? Its better let Lead developer decides that :)

MissAllSunday:
It would probably be better if we create every new theme with a responsive file  then let themres decide if thye want to use the one form default or use current by simple keep or remove the responsive file, if its removed it means the themer wants to use the default one, if it was kept then use that one.

loadCSS can check if the current theme has that file and if it doesn't have it, load the one from the default theme.

XinYenFon:
Good for me :)

PR #2049, opened 2014-07-14
Clean up the CSS a bit, removing prefixed entries where not needed.

NanoSector:
Also added an IE compatibility stylesheet. A different mechanic is needed to actually load it.


XinYenFon:
No point to do this. IE8/9 doesn't support linear-gradient. Also you are just removing the -ms- prefix from one place and create new file to use it. http://caniuse.com/css-gradients

XinYenFon:
Actually it would be awesome if you add white (#FFFFFF) fallback to those gradients

Oldiesmann:
Looking at the "icanuse" site, it looks like dropping the -webkit-linear-gradient declarations would actually be a bad idea, as the following still need it:

Safari on iOS 6.1 or earlier (probably not many iOS 6.1 devices still in use, but you never know)
Default Android browser except on Android 4.4 (this covers all but a small percentage of devices)
Opera Mobile (latest version)
Default BlackBerry browser (as of BB 10 at least)

XinYenFon:
probably my changes created the conflict. :/ #2054

PR #2050, opened 2014-07-14
Fix #1569

### LINKED ISSUE #1569: Internal admin settings do not adhere to limits
Some settings that are int or float also declare min/max values. While these may be enforced through the front end, these are never enforced in the back end.

saveDBSettings needs to validate and clamp the values within ranges if appropriate.

NanoSector:

### DIFF
--- a/Sources/ManageServer.php
+++ b/Sources/ManageServer.php
@@ -1068,10 +1068,10 @@ function saveDBSettings(&$config_vars)
 		}
 		// Integers!
 		elseif ($var[0] == 'int')
-			$setArray[$var[1]] = (int) $_POST[$var[1]];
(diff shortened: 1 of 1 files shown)


Oldiesmann:
This won't work entirely because the 3rd item in the array won't always be the max value and it also doesn't check against minimum values. Since we have min and max keys for the config var array, just check for $var['min'] and $var['max']. Also, if $var['min'] isn't set, we're implying a minimum of 0.

Oldiesmann:
If min is set, $var['min'] will be set. If max is set, $var['max'] will be set. That's how I implemented it. $var[2] will always be the third item in the array, which may or may not be what you're looking for, so I don't recommend using that.

NanoSector:
Ok I thought about harmonising the panels altogether, but it seems as if there's nothing documented nor the "correct" way to do things. And I _really_ do not feel like backtracing SMF again.

PR #2055, opened 2014-07-14
Stats for likes

MissAllSunday:
Adds two more stats, top liked messages and top liked users, top liked users only takes into consideration liked messages (content_type == 'msg').

Might want to check the queries for getting the stats. Perhaps they could be improved.

I also encounter a token validation error while saving both mentions and likes settings, after some debugging it turns out the inline permissions were causing it, init_inline_permissions() needed to create the token that save_inline_permissions() validates, creating said token inside  init_inline_permissions() appers to fix the issue but more testing would be needed.

### DIFF
--- a/Sources/Stats.php
+++ b/Sources/Stats.php
@@ -149,7 +149,7 @@ function DisplayStats()
 	);
 	$context['latest_member'] = &$context['common_stats']['latest_member'];
 
-	// Male vs. female ratio - let's calculate this only every four minutes.
+	// Let's calculate gender stats only every four minutes.
 	$disabled_fields = isset($modSettings['disabled_profile_fields']) ? explode(',', $modSettings['disabled_profile_fields']) : array();
 	if (!in_array('gender', $disabled_fields))
 	{
--- a/Sources/Stats.php
+++ b/Sources/Stats.php
@@ -567,29 +567,29 @@ function DisplayStats()
 		$context['stats_blocks']['liked_users'] = array();
 		$max_liked_users = 1;
 		$liked_users = $smcFunc['db_query']('', '
-			SELECT l.id_member, l.content_id, l.content_type, COUNT(mem.id_member) AS count, mem.id_member AS liked_user, mem.real_name, mem.posts
+			SELECT m.id_member AS liked_user, COUNT(l.content_id) AS count, mem.real_name
 			FROM {db_prefix}user_likes AS l
 				INNER JOIN {db_prefix}messages AS m ON (l.content_id = m.id_msg)
 				INNER JOIN {db_prefix}members AS mem ON (m.id_member = mem.id_member)
 			WHERE content_type = {literal:msg}
-				AND mem.posts > {int:no_posts}
-				AND mem.id_member > {int:no_posts}
-			GROUP BY mem.id_member
+				AND m.id_member > {int:zero}
+			GROUP BY m.id_member
 			ORDER BY count DESC
 			LIMIT 10',
 			array(
 				'no_posts' => 0,
+				'zero' => 0,
 			)
 		);
 
 		while ($row_liked_users = $smcFunc['db_fetch_assoc']($liked_users))
 		{
 			$context['stats_blocks']['liked_users'][] = array(
-				'name' => $row_liked_users['real_name'],
 				'id' => $row_liked_users['liked_user'],
 				'num' => $row_liked_users['count'],
 				'href' => $scripturl . '?action=profile;u=' . $row_liked_users['liked_user'],
-				'link' => '<a href="' . $scripturl . '?action=profile;u=' . $row_liked_users['liked_user'] . '">' . $row_liked_users['real_name'] . '</a>'
+				'name' => $row_liked_users['real_name'],
+				'link' => '<a href="' . $scripturl . '?action=profile;u=' . $row_liked_users['liked_user'] . '">' . $row_liked_users['real_name'] . '</a>',
 			);
 
 			if ($max_liked_users < $row_liked_users['count'])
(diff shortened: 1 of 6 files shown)


Sources/Stats.php
```diff
@@ -521,6 +521,93 @@ function DisplayStats()
...
 		cache_put_data('stats_total_time_members', $temp2, 480);
 
+	// Likes.
+	if (!empty($modSettings['enable_likes']))
+	{
+		// Liked messages top 10.
+		$context['stats_blocks']['liked_messages'] = array();
+		$max_liked_message = 1;
+		$liked_messages = $smcFunc['db_query']('', '
+			SELECT m. id_msg, m.subject, m.likes, m.id_board, m.id_topic, t.approved
```
live627: Extra space in `m. id_msg`

Sources/Stats.php
```diff
@@ -521,6 +521,93 @@ function DisplayStats()
...
+			$context['stats_blocks']['liked_messages'][] = array(
+				'id' => $row_liked_message['id_topic'],
+				'subject' => $row_liked_message['subject'],
+				'num' => $row_liked_message['likes'],
+				'href' => $scripturl . '?topic=' . $row_liked_message['id_topic'] . '.msg' . $row_liked_message['id_msg'],
+				'link' => '<a href="' . $scripturl . '?topic=' . $row_liked_message['id_topic'] . '.msg'. $row_liked_message['id_msg'] .'">' . $row_liked_message['subject'] . '</a>'
+			);
+
+			if ($max_liked_message < $row_liked_message['likes'])
+				$max_liked_message= $row_liked_message['likes'];
```
live627: Need space before `=`

Themes/default/css/index.css
```diff
@@ -2243,6 +2243,9 @@ dl {
 .stats_icon.starters, .stats_icon.people {
 	background-position: -80px -48px;
 }
+.stats_icon.liked_users, .stats_icon.liked_messages{
+	background: url(../images/generic_icons.png) no-repeat -64px 0;
```
live627, line 2247: Maybe this could be `background-position:  -64px 0;`

I haven't tested this yet.

MissAllSunday, line 2247: I just copied it from the main likes code

MissAllSunday:
Theres a problem with the most liked users query, will check that further.

Themes/default/css/index.css
Oldiesmann, line 2247: The "background" property is a shortcut that allows you to specify all the individual properties on a single line. This is no different than this:

```
background-image: url(../images/generic_icons.png);
background-repeat: no-repeat;
background-position: -64px 0;
```

MissAllSunday:
OK heres another attempt to fix the liked users query.

And now that my lappy exploded, this PC has "origin" AKA, this repo as the default when pushing changes, thats why I've been creating new branches directly, it looks like I can specify the default to push repo somewhere. Gosh I hate CL...

Dragooon:
Querying members table for real_name separately might be faster rather than doing a join for every message even if it's by the same member and group by m.id_member instead (I'm not 100% sure, theoretically it might be but I haven't tested it).

MissAllSunday:
Ok, how about this one?

``` php
        // Liked users top 10.
        $context['stats_blocks']['liked_users'] = array();
        $max_liked_users = 1;
        $liked_users_ids = array();
        $liked_users = $smcFunc['db_query']('', '
            SELECT m.id_member AS liked_user, COUNT(l.content_id) AS count
            FROM {db_prefix}user_likes AS l
            INNER JOIN {db_prefix}messages AS m ON (l.content_id = m.id_msg)
            WHERE content_type = {literal:msg}
                AND m.id_member > {int:zero}
            GROUP BY m.id_member
            ORDER BY count DESC
            LIMIT 10',
            array(
                'no_posts' => 0,
                'zero' => 0,
            )
        );

        while ($row_liked_users = $smcFunc['db_fetch_assoc']($liked_users))
        {
            $liked_users_ids[] =  $row_liked_users['liked_user'];
            $context['stats_blocks']['liked_users'][$row_liked_users['liked_user']] = array(
                'id' => $row_liked_users['liked_user'],
                'num' => $row_liked_users['count'],
                'href' => $scripturl . '?action=profile;u=' . $row_liked_users['liked_user'],
            );

            if ($max_liked_users < $row_liked_users['count'])
                $max_liked_users = $row_liked_users['count'];
        }

        $smcFunc['db_free_result']($liked_users);

        $liked_users_names = $smcFunc['db_query']('', '
            SELECT real_name, id_member
            FROM {db_prefix}members
            WHERE id_member IN({array_int:members})',
            array(
                'members' => $liked_users_ids,
            )
        );

        while ($row_liked_users_name = $smcFunc['db_fetch_assoc']($liked_users_names))
                $context['stats_blocks']['liked_users'][$row_liked_users_name['id_member']] = $context['stats_blocks']['liked_users'][$row_liked_users_name['id_member']] + array(
                    'name' => $row_liked_users_name['real_name'],
                    'link' => '<a href="' . $scripturl . '?action=profile;u=' . $row_liked_users_name['id_member'] . '">' . $row_liked_users_name['real_name'] . '</a>',
                    'percent' => round(($context['stats_blocks']['liked_users'][$row_liked_users_name['id_member']]['num'] * 100) / $max_liked_users),
                );

        $smcFunc['db_free_result']($liked_users_names);
```

I needed to add the 

AND m.id_member > {int:zero}

because it is possible to like the very first message made by Simple Machines which has 0 as its id_member

Oldiesmann:
Actually, on my server, the all-in-one query is significantly faster - 0.0036 seconds vs. 0.046 seconds (and that's just for the query to get the IDs of users whose posts have been liked - not counting the one to pull the names of those members). I'd just stick with that one for now since it's fast and we don't have to run two queries to get everything.

Dragooon:
Ahh, messages doesn't have an index for id_member. Okay, I guess it should be fine.

live627:
cos I can :P

PR #2068, opened 2014-07-19
Implement simple language select

live627:


ChainyEO:
The language selector is a nice idea and I'm sure that many people will find it useful. Is there an option to hide it, though, even if you do have more than one language installed?  

As Illori mentioned, there is indeed a slight problem with the alignment of the items. The Search field and the dropdown list "Entire Forum" seem to be correctly aligned, but then the "Search" button looks a bit low, and the language selector looks too high:

lang selector zoomed

XinYenFon:
Search button is not low, but bigger. Also fixed the placement issue here https://github.com/Antes/kittenground/commit/aa32b84acc3fdb547aa04b3d60f6e7d26f5db925

PR #2084, opened 2014-07-21
some css cleaning

XinYenFon:
fixes #2049 


live627:
Does adding these globals fix something?

XinYenFon:
yea; https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/ManageBans.php#L131

live627:
ah yes I see now :P

PR #2099, opened 2014-07-23
Update editor CSS to SCE v1.4.5-dev

live627:


Oldiesmann:
My only concern with this is that using minified CSS will make it more difficult for theme authors to theme the editor.

XinYenFon:
they can get normal version from author's repo or they can use unminify things.

Oldiesmann:
Yes, but it's still hard to theme because you can't preview your changes simply by tweaking some CSS and uploading the file - you either have to hack the code to not use the minified version or you have to minify your file after every change. It's a lot more work with little additional benefit.

XinYenFon:
simply rename your unminified version to abc.min.css and you're done? :D

PR #2101, opened 2014-07-23
Scheduled tasks and external tasks

MissAllSunday:
There is a hook for adding external tasks which is great but its limited to plain functions.
Using the task name for everything works great but it would be better to have a specific  column for a callable with a larger field (varchar(60)).

This change allows adding full methods to run scheduled task, it is fully compatible with SMF standard function tasks and/or old modifications.

A simple example for adding an external scheduled tasks is as follows:

``` php
<?php

include 'SSI.php';

    add_integration_function('integrate_autotask_include', '$sourcedir/Test.php|Test::scheduledTask#');

    $smcFunc['db_insert'](
        'insert',
        '{db_prefix}scheduled_tasks',
        array(
            'id_task' => 'int',
            'next_time' => 'int',
            'time_offset' => 'int',
            'time_regularity' => 'int',
            'time_unit' => 'string',
            'disabled' => 'int',
            'task' => 'string',
            'callable' => 'string',
        ),
        array(
            0, 0, 0, 1, 'd', 1, 'test', '$sourcedir/Test.php|Test::scheduledTask#',
        ),
        array(
            'id_task',
        )
    );
```

``` php
<?php

class Test
{
    public function scheduledTask()
    {
        global $smcFunc;

        $smcFunc['db_query']('', '
            UPDATE {db_prefix}messages
            SET body = CONCAT(body, {string:notice})
            WHERE id_msg = {int:m}',
            array(
                'm' => 1,
                'notice' => '<br /> It works! it works! muahahahahah',
            )
        );

        return true;
    }
}

```


Oldiesmann:
I assume the "how many alerts to display" commit wasn't supposed to be part of this?

MissAllSunday:
Yes and I added another commit to revert the changes: https://github.com/MissAllSunday/SMF2.1/commit/d96993b137917587e3707e1bc4df5dc15e3e7ba5

Oldiesmann:
Somehow missed that :)

PR #2107, opened 2014-07-24
Rewrite a text string (thanks to Illori)

NanoSector:


ChainyEO:
> Note that logging of every spider action only occurs if tracking is set to either "high" or "very high". Detail of every spiders action is only logged if tracking is set to "very high"

Detail? Does this mean that if it's set at "high", then the logging takes place, but no details are recorded? Feels like these sentences could be made clearer. Why have we completely deleted the original sentence about server strain?

PR #2108, opened 2014-07-24
Fix #2106

### LINKED ISSUE #2106: aggressive spider log useless info...
with the spider log on aggressive it should tell you what topic the spider has viewed

instead it is showing stuff like this...

```
a:4:{s:5:"topic";i:37;s:9:"prev_next";s:4:"prev";s:5:"board";i:1;s:10:"USER_AGENT";s:72:"Mozilla/5.0 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)";}
```

NanoSector:

### DIFF
--- a/Sources/ManageSearchEngines.php
+++ b/Sources/ManageSearchEngines.php
@@ -592,7 +592,9 @@ function logSpider()
 		{
 			$url = $_GET + array('USER_AGENT' => $_SERVER['HTTP_USER_AGENT']);
 			unset($url['sesc'], $url[$context['session_var']]);
-			$url = serialize($url);
+			
+			// Encoding so we are sure the variable doesn't get messed with.
+			$url = base64_encode(serialize($url));
 		}
 		else
 			$url = '';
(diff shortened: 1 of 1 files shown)


live627:
At the risk of sounding dumb, how would I reproduce this issue?

illori:
set your spider settings to the top 2 options and get some spider data in your database and go from there. if you need some data just pm me on sm.org and i can give you my table.

NanoSector:
Is this still relevant? Some people were against this implementation.

illori:
i have since removed the entries in the spider log and had none since so i cant confirm if we still need this or not.

NanoSector:
Yeah this still applies... 

Notice: Undefined index: viewing in /mnt/data/Web/smf21/Sources/ManageSearchEngines.php on line 788

and the same garbage:
Yahoo!  Today at 05:02 AM   a:2:{s:5:"board";i:1;s:10:"USER_AGENT";s:83:"Mozilla/5.0 (compatible; Yahoo! Slurp; http://help.yahoo.com/help/us/ysearch/slurp)";}

Oldiesmann:
I'm still confused as to how base64 encoding is going to fix anything here.

NanoSector:
It seemed to fix this at the time since apparently the string got corrupted while being transferred to the database.

illori:
yep still an issue and still has the error on line 788

Oldiesmann:
The undefined index thing is because "viewing" is never set anywhere.

illori:
@Oldiesmann can we get this merged in? if we can not can you give a good reason why this can not fix a bug that exists in the software?

Oldiesmann:
I still don't understand how this fixes anything. If there are unexpected characters in the serialized string, they're still going to be there regardless of whether you encode it.

illori:
then can you try to find another way to fix this? this did fix it when i tested it. i should still have that db dump you could test with to see if you can fix it another way.

MissAllSunday:
Instead of base64, why don't simply use htmlspecialchars?

If the serialized string got corrupted because it was saved with some weird characters inside then convert anything to a  safe, storable character before doing the serialize.

The undefined index "viewing" is a totally different issue altogether, it means there wasn't any "viewing" key on the array, this can be caused by several different things and I doubt a corrupted string will fire this error.A corrupted string will simple choke when unserialize() was used.

Oldiesmann:
The "viewing" is set based on the "url" column of the database, so that should never be undefined.

I still don't understand how encoding a serialized string will fix this though, as MySQL shouldn't be doing anything to the characters in that string.

PR #2123, opened 2014-07-29
More icon doohickies 

live627:


XinYenFon:
the hot/very hot topic things removed from SMF 2.1 why you keeping them?

PR #2170, opened 2014-08-08
Update team members

colinschoen:
Signed of by Colin Schoen (colin@simplemachines.org)


Dragooon:
SA, Relyana etc don't want their full names or were they not added?

colinschoen:
That's what was in there previously. I didn't modify it if they were already added.

PR #2190, opened 2014-08-14
Unneeded code

MissAllSunday:
This was code I introduced for handling methods before call_helper(), now call_helper() can handle calls to instantiated methods so this is redundant, it wasn't used by anyone but me so its safe to remove it.


Dragooon:
Sold

PR #2206, opened 2014-08-19
! Fix syntax errors in Who.php

SSPMark:
Any page that was loading up Sources\Who.php was generating a syntax error due to missing commas.


live627:
Once you sign off your commit (you can amend it), I will merge it in.

SSPMark:
Done. Forgot about that bit, thanks.

live627:
Thanks :)

PR #2209, opened 2014-08-20
! don't show reg agreement links/page if its not enabled. fixes #2208

### LINKED ISSUE #2208: hide terms and rules if registration agreement is disabled
if the admin does not require their users to agree to the registration agreement the terms and rules link should not show up as it is not the rules for that forum.

XinYenFon:

### DIFF
--- a/Themes/default/Help.template.php
+++ b/Themes/default/Help.template.php
@@ -174,7 +174,8 @@ function template_manual()
 function template_terms()
 {
 	global $txt, $context, $modSettings;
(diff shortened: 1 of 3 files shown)


MissAllSunday:
Looks good, perhaps just set an empty line between declaring globals and the if conditional. Same for the else and the end of the first echo.

XinYenFon:
yeap, missed that :) but i never see SMF does line break before else ? that's why i didn't do that too.

live627:
looks good to me.

PR #2210, opened 2014-08-20
Fixes #2013 --> Undefined index: host

### LINKED ISSUE #2013: Undefined index: host
\Themes\Sources\Subs-Auth.php on line 134
\Themes\Sources\Session.php on line 41

Happens after installer creating admin account

margarett:
Signed-off-by: margarett bruno.m.alves@gmail.com

### DIFF
--- a/other/install.php
+++ b/other/install.php
@@ -1446,7 +1446,7 @@ function DeleteInstall()
 {
 	global $txt, $incontext;
 	global $smcFunc, $db_character_set, $context;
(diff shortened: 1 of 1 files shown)


live627:
dafuck, really that simple?

margarett:
Yup, that takes out the undefined error ;)

Unfortunately there is another issue "attached", as the auto login after install doesn't work...

Dragooon:
Merging this in for now since it fixes the undefined index. The not logging in issue is odd but I'll take a look later.

PR #2212, opened 2014-08-21
Just a bit of a visual indicator that user is logged in.

SSPMark:
Seems like the going trend these days is to put the users avatar next to their username in the top menu bar (just like here on GitHub) when the user is logged in. It gives a bit more of a visual appearance that the user is in fact logged in where they might not catch it with just text.

Just a thought.

### DIFF
--- a/Themes/default/Profile.template.php
+++ b/Themes/default/Profile.template.php
@@ -47,22 +47,9 @@ function template_profile_popup()
 	// Unlike almost every other template, this is designed to be included into the HTML directly via $().load()
 
 	echo '
-		<div class="profile_user_info">';
-
-	// Firstly, an avatar.
-	if (!empty($context['user']['avatar']))
-		echo '
-			<a href="', $scripturl, '?action=profile" class="avatar">', $context['user']['avatar']['image'], '</a>';
-
-	// Then a couple of bits and pieces that might be interesting
-	echo '
-			<ol>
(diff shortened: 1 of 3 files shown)


SSPMark:
Here's a visual reference.

top-bar-avatar

SSPMark:
Better alignment of the avatar.

top-bar-avatar-adjusted

Dragooon:
I like it

live627:
The avatar is also in the user menu. Should that stay or go?

SSPMark:
I was on the fence on that. It's redundant, but still fits as it's not visible all the time.

SSPMark:
Okay, after messing around with it a bit more, I went ahead and removed the avatar from the user menu.

top-bar-avatar-no-avatar-popup

Oldiesmann:
I personally liked it better the way it was before. Plus, we already have a visual indicator that the user is logged in - their username.

SSPMark:
I know, but switching welcome text with username text and your average user may not catch that they are, in fact, logged in or even worse, logged in as someone else on a shared computer. With the avatar, it pops out more. A lot of sites these days make your avatar ever present when you're logged in. See GitHub, Facebook, Disqus, Google, Xbox.com, Playstation website, etc. so people are likely used to it.

The ones who don't do this, such as Twitter and Dropbox for example, serve a completely different page and look to you when you're not actually logged in versus when you are so it's easier to catch.

PR #2214, opened 2014-08-21
Give "$software_year" automatic Year Update!

M50Scripts:
I updated the "$software_year" to "'.date('Y').'" which will auto update the "Year" in every new years eve :) better than doing it manually (save time, yeah!)

### DIFF
--- a/index.php
+++ b/index.php
@@ -21,7 +21,7 @@
  */
 
 $forum_version = 'SMF 2.1 Alpha 1';
-$software_year = '2014';
+$software_year = ''.date('Y').'';
 
 // Get everything started up...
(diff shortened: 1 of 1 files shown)


MissAllSunday:
Hi, thank you for your contribution.

There are a few problems with it though, as per our [Developer's Certificate of Origin](https://github.com/SimpleMachines/SMF2.1/blob/master/DCO.txt) every commit we receive has to be signed by the comitter.

The year doesn't have to change every year, it only needs to change when new code is added, on our case, this typically means when we do a new release.

Thanks for taking the time to contribute :smile:

PR #2220, opened 2014-08-25
Floating cat_bar's should get rounded corners.

SSPMark:
If a cat_bar is visually floating and not attached to any container, like the category headers on the index page, then they should have rounded corners all around. This one I'm just tossing up as an idea, if you guys don't like it, feel free to reject.

Before:
post-rounded-corners-before
After:
post-rounded-corners-after

If you guys like this one, and _before_ you merge in the pull request, let me know and I'll make the changes to all cat_bar's that aren't visually attached to a container.

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -3247,6 +3247,7 @@ div#pollmoderation {
 }
 #forumposts .cat_bar {
 	margin: 0 0 -4px 0;
+	border-radius: 6px;
 }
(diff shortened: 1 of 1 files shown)


XinYenFon:
we also have option how many people viewing this topic thingy, that'll look bit weird with that... Maybe better not to touch it for now.

SSPMark:
Okay, sounds good. I'll leave it alone for now.

PR #2221, opened 2014-08-25
Connected board description to title bar on board index.

SSPMark:
On the topic listing for a board, the description bar was floating away from the title_bar style used for the "Subject / Started by" etc bar. Here's an example of it connected. Looks nicer I think.

Before:
messageindex-spacing-before
After:
messageindex-spacing-after


XinYenFon:
This is really nice catch :) thanks

PR #2223, opened 2014-08-26
! Small fix for user_info (drop)

XinYenFon:
! Lets make it look bit better if its under QuickReply ;)
screenshot-localhost 2014-08-26 20-31-38
screenshot-localhost 2014-08-26 20-32-04

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -777,13 +777,13 @@ img.sort, .sort {
 	max-height: 50px;
 }
 #profile_menu .profile_user_info {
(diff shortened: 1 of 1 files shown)


SSPMark:
Doh! My fault, it's not an issue with QuickReply(?), it's with short names. I forgot to test that.

Good catch. :smile:

XinYenFon:
Nah I don't write proper PR messages :P, so GH getting my commit message(s)...

SSPMark:
oh haha

PR #2225, opened 2014-08-28
Just some random typos.

SSPMark:
Don't know what this file is used for, but it had some random typos in it. :wink: 

~~Also, about the large blocks of changes with the names, my Sublime Text changed the line endings of those sections to match the rest of the file.~~ It removed the whitespace at EOL.

The actual fixes are on lines 49 and 58.


SSPMark:
Guess if I'm fixing typos, may as well keep going.

Oldiesmann:
Looks good, but whatever editor you're using is sticking new lines at the end of the file when there weren't any there originally.

SSPMark:
Yeah, I have that turned on for consistency on my own projects so that all files are formatted the same (new line at EOF, Unix line endings, no trailing whitespace at EOL).

For SMF, it looks like some files have new lines at EOF and some don't so I didn't think it mattered. I can turn it off for this project only if you don't want new lines on the files that don't have them to begin with.

MissAllSunday:
It will help to have a new line at the end, unfortunately we cannot remove the ?> at the end of each file (Package manager still has the "end" instruction on it) but it cannot hurt to have a new line at the end, I was going to do that but it has to be nearly one of the last things to do before a formal release to make sure everthing its alright.

SSPMark:
Do you mean it looks like I removed the ?> from the end, or that the ?> has to be the very last thing at the end of the file, therefore the new line should be removed? If the latter, I'll got back and remove the new lines.

SSPMark:
Just in case, I removed them for now and updated my project specific settings so that moving forward, it will not add them until it's decided to do so. :smile:

MissAllSunday:
Idially, a PHP file shouldn't have a closing >? and will have a new line at the end, we cannot remove the ?> because the packager manager still use that as reference for mod authors to add code at the end of a file but we could make all source files to end with a new line. This isn't mandatory and its not really that important for our enviroment.

But yeah, at one point we should prob check for uniformity, either all files with or without a newline at the end.

SSPMark:
Sounds good, I'll keep an eye out in the future if that changes. For now I'll leave them as is.

PR #2231, opened 2014-08-31
Improve topic/board alert's preferences

### LINKED ISSUE #1320: Merge Watch this topic & Watch
Hai,

Today i saw new enchantment for the topic systematic and as far as i can see there is an upcoming notify (aka user-tagging) system coming and i think we should take GitHub model in this like they use in repos (Not Watching / Watching / ...)

Our model can be;
- Not Watching (You get notification only when you mentioned)
- Watching without mails (You get notification only via Alerts area)
- Watching with mails (You get notifications via Alerts area & mails)
- Ignoring (You'll not get any mail or alert even when you mentioned)

Dragooon:
Previously watching/unwatching and alert/notify preferences were separate, unify them and make them a better UX. Fixes #1320


Dragooon:
Currently WIP, don't merge

Sources/Display.php
```diff
@@ -1199,10 +1215,32 @@ function Display()
...
 		'mark_unread' => array('test' => 'can_mark_unread', 'text' => 'mark_unread', 'image' => 'markunread.png', 'lang' => true, 'url' => $scripturl . '?action=markasread;sa=topic;t=' . $context['mark_unread_time'] . ';topic=' . $context['current_topic'] . '.' . $context['start'] . ';' . $context['session_var'] . '=' . $context['session_id']),
-		'unwatch' => array('test' => 'can_unwatch', 'text' => ($context['topic_unwatched'] ? '' : 'un') . 'watch', 'image' => ($context['topic_unwatched'] ? 'un' : '') . 'watch.png', 'lang' => true, 'url' => $scripturl . '?action=unwatchtopic;topic=' . $context['current_topic'] . '.' . $context['start'] . ';sa=' . ($context['topic_unwatched'] ? 'off' : 'on') . ';' . $context['session_var'] . '=' . $context['session_id']),
 		'print' => array('test' => 'can_print', 'text' => 'print', 'image' => 'print.png', 'lang' => true, 'custom' => 'rel="nofollow"', 'url' => $scripturl . '?action=printpage;topic=' . $context['current_topic'] . '.0'),
+		'notify' => array(
+			'lang' => true,
+			'test' => 'can_set_notify',
+			'text' => 'notify_topic_' . $context['topic_notification_mode'],
+			'elements' => array(
```
live627: Should this be sub_buttons?

XinYenFon:
would it be better if we remove this from general buttons and put top-right side of the topic, and it needs to be bit more down and right. (1366x768)

screenshot-localhost 2014-09-01 01-43-02

Themes/default/languages/index.english.php
```diff
@@ -833,4 +833,14 @@
...
+
+// Notification post control
+$txt['notify_topic_0'] = 'Not Following';
+$txt['notify_topic_1'] = 'No Alerts or Emails';
+$txt['notify_topic_2'] = 'Receive Alerts';
+$txt['notify_topic_3'] = 'Receive Emails and Alerts';
+$txt['notify_topic_0_desc'] = 'You will not receive any emails or alerts for this post and it will also not show up in your unread replies and topics list. You will still receive @mentions for this topic.';
+$txt['notify_topic_1_desc'] = 'You will not receive any emails or alerts but only @mentions by other members.';
+$txt['notify_topic_2_desc'] = 'You will receive alerts for this topic.';
+$txt['notify_topic_3_desc'] = 'You will receive alerts as well as alerts for this topic,';
```
live627: typo

XinYenFon: I think this would be better: "You will receive both alerts and emails for this topic."

Edit: oops ^^ yeap emails...

SSPMark: Agree with Antes, that flows a little better. Though, I'd say 'emails' instead of 'mails', unless that's what you meant Antes. Plus, seeing as the heading is 'Receive Emails and Alerts' best if the text follows suit by putting the word emails first. Same goes for the 'No  Alerts or Emails' text, they should all be in the same order IMO.

Oldiesmann:
Great to see progress being made on this!

Dragooon:
Can anyone give this a quick test again?

XinYenFon:
ok :P this is good to go for now.

Dragooon:
Added support for per-board preferences

XinYenFon:
Its working 0.o

but one simple thing, if you change your setting to receive alert on new topic (on that board), you shouldn't be getting this:

"-user- posted -topic_name- in -board_name- - -time-"
it should be;
"-user- started(posted) a new topic -topic_name- in -board_name- - -time-

---

Notice: Undefined offset: 0 in C:\wamp\www\kittenground\Sources\Profile.php on line 87

PR #2240, opened 2014-09-04
Make call_helper to also support closures

MissAllSunday:
This is a tricky one, haven't realize that you could also pass a closure to call_helper but its more common that I imagined, usage of createList() heavily relies on closures.

In theory this should sufice to detect any closure and just return the object, downside of this is the case when create_function() is used but since we're now on 5.3, modders shouldn't really use that anyway.

I'm tempted to change the && for a || just in case someone passes an already instantiated method as a var:

$var = $this->foo();

call_helper($var);

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -4205,6 +4205,10 @@ function call_helper($string, $return = false)
 	if (empty($string))
 		return false;
 
+	// Is this a closure?
+	if (is_object($string) && ($string instanceof Closure))
+		return $string;
+
 	// Stay vitaminized my friends...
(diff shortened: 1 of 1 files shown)


live627:
> I'm tempted to change the && for a || just in case someone passes an already instantiated method as a var:

This might be a good idea. But then again, wouldn't is_object alone suffice for this and catch closures all the same?

Dragooon:
Actually, wouldn't just $string instanceof Closure suffice? is_object would return true for a lot of other things, and those other things won't really make any sense in our case since we wouldn't know what to do with them.

MissAllSunday:
yep, that makes sense

PR #2241, opened 2014-09-04
Changes on display

XinYenFon:


Themes/default/Display.template.php
```diff
@@ -23,9 +23,32 @@ function template_main()
 			</div>';
 	}
 
-	// Show the anchor for the top and for the first message. If the first message is new, say so.
+	// Show new topic info here?
 	echo '
+		<div id="display_head">
+			<h2 class="display_title">'. $context['subject']. '', ($context['is_locked']) ? '<span class="generic_icons lock"></span>' : '','', ($context['is_sticky']) ? '<span class="generic_icons sticky"></span>' : '','</h2>
+			<p>',$txt['started_by'],' ', $context['topic_poster_name'],', ', $context['topic_started_time'],'</p>';
```
live627, line 30: This seems a little messy to me...

I'd change it to:
`<h2 class="display_title">', $context['subject'], ($context['is_locked']) ? '<span class="generic_icons lock"></span>' : '', ($context['is_sticky']) ? '<span class="generic_icons sticky"></span>' : '', '</h2>`

Oldiesmann:
Two things...
1. You removed the code to display the poll question.
2. Why are we removing those icons? Similar icons (using generic_icons rather than topic-icon) are shown on the messageindex.

XinYenFon:
1) I'm just showing Question as title (beside static Poll)
2) yes that's why, consistency... and using less stuff.

Oldiesmann:
Ok. Didn't see either of those :tongue: 

I still think the poll question looks fine where it is though. Don't see any reason to put it up so far above the poll options.

Oldiesmann:
Looking at this again, I just realized something. Only 2 of the 8 icons from the topic icons sprite that are currently used are also in generic icons. However, with a slight tweak to the determineTopicClass function we could also use another 8 of them ("my" topic/poll vs a topic/poll someone else started), so we should probably combine those as well.

The rest are for "hot" and "very hot" topics, which is no longer a feature in SMF.

XinYenFon:
I did this because smf does not allow second questions also removed margin and some padding they look more close.

-----Original Message-----

PR #2243, opened 2014-09-04
Content Slider

XinYenFon:
This is mostly default non-minimized version of this work. Mostly default because some options (Next-Prev icons - captions for img) disabled for now.

Non-minimized version takes 25kb of space, while minimized version only takes 12kb.
Slider can be found at https://github.com/booncon/slippry


Themes/default/css/index.css
```diff
@@ -4051,3 +4032,381 @@ h3.profile_hd {
...
+#smf_slider li {
+	padding: 5px 10px;
+}
+.sy-slides-wrap {
+	padding: 0 0 15px 0;
+}
+.sy-box {
+	margin: 0 0 10px 0;
+}
+/**
```
live627: Should all the code below be in its own file?

XinYenFon:
moved the css to its own file and loaded only on boardindex :paw_prints:

PR #2246, opened 2014-09-05
Slightly larger checkboxes. WIP - Feedback requested.

SSPMark:
Okay, so, this one is another one of those "let me know what you think and I'll work on it more" things. :stuck_out_tongue:
1. The checkboxes feel really small to me. I don't know if it's the new theme overall, or just the styled text inputs that are being used. So increased the size ever so slightly.
2. Should we clear the browser default styling and customize it to look more like the text inputs? i.e. white bg and black thin border?

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -106,6 +106,11 @@ input.input_check, input.input_radio {
 	background: none;
 	vertical-align: top;
 }
+/* Slightly larger checkboxes. */
+input.input_check {
+  width: 16px;
+  height: 16px;
+}
 /* Give disabled text input elements a different background color. */
 input[disabled].input_text {
 	background: #eee;


XinYenFon:
I didn't notice any difference?

SSPMark:
Doh, that's what I get for not cross browser testing right out the gate. They stay the same in Firefox because in Firefox they're larger already. So, this really just fixes Chrome which renders them a few pixels smaller.

Before: checkboxes-before After: checkboxes-after

Interesting note, in IE, they're the same size as Firefox with no CSS, but add my CSS and they are even larger.

Will have to test more thoroughly if it's decided as a good change.

XinYenFon:
Seems nice especially for people like @Kindred-999 :D :D

colinschoen:
Agreed, this seems like a useful tweak.

SSPMark:
Okay, I'll finish it up and mark it when I'm done.

live627:
Mark, any news?

SSPMark:
I'm terribly sorry, but time just got away from me. Feel free to merge it
in and pick up where I left off. What I've got so far at least makes it so
that the size of the checkboxes in Chrome match that of Firefox and
Internet Explorer. Otherwise I can try to finish it up (i.e. full testing)
later this week, or at the lest, this weekend.

PR #2247, opened 2014-09-06
Content Slider

XinYenFon:
test this version, more work will be done later...


Dragooon:
Do we really need a 25kb JS file on every page load?

XinYenFon:
It's not loading on every page only boardindex if news fader enabled

-----Original Message-----

XinYenFon:
Also side note (I'm currently on road), this is non-minimized version, also I didn't do any cleanup, I just want to know if its worth to be in so i can do the cleaning and end it :)

PR #2286, opened 2014-09-11
Stuff

### LINKED ISSUE #2269: Generic lists call for items even when there aren't any
Applies to both 2.0 and 2.1

Some lists don't bother to specify a number of get-count function, because they don't need to paginate, so the items must always be fetched. That's fine.

But if $list_context['total_num_items'] is defined, we already know how many items there are, and if this is 0, there's no point in calling $listOptions['get_items']['function'], might as well avoid the query in that case, since we already know there won't be any rows from it.

It's not a huge deal but every little helps when it comes to performance, right?

MissAllSunday:
fixes #2269

and add the new placements for cust_fields on PersonalMessages.template  seems the commit where I added those was never actually merged, must have been pushed on one of the lose branches I have :(

It seems theres a merge conflict, will go ahead and create the PR so github can tell me where the conflict is


MissAllSunday:
...and merge conflict resolved :smile:

Themes/default/PersonalMessage.template.php
```diff
@@ -347,17 +385,29 @@ function loadLabelChoices()
...
 					foreach ($message['member']['custom_fields'] as $custom)
 						if (empty($custom['placement']) || empty($custom['value']))
 							echo '
-				<li class="custom">', $custom['title'], ': ', $custom['value'], '</li>';
-				}
+				<li class="custom ', $custom['col_name'] ,'">', $custom['title'], ': ', $custom['value'], '</li>';
 
 				// Are we showing the warning status?
 				if ($message['member']['can_see_warning'])
+
```
Oldiesmann: This will break the if statement...

MissAllSunday: oops, yeah, sorry about that, I'll add a fix soon.

PR #2288, opened 2014-09-11
Normalize files

MissAllSunday:
- Removes lose spaces and tabs.
- Adds ?> at the end of files
- Removes new line at the end of line
- Adds a new line before the ?> this is merely for those mod authors out there that still uses the "end" search param, I hope they all remember to add their own new lines after and before the code they're inserting.  A vague attempt to prevent clunky results.
- Changes the version from Alpha 1 to Beta 1


Dragooon:
We need ?> at the end of files for the package manager right? Otherwise I'd rather not have them

MissAllSunday:
Yes, the package manager still has the "end" search param, that feature uses the ?> as reference so we cannot really remove that.

I'm all for deleting that feature from the package manager but a lot of mod authors does use it, specially those who adds scheduled tasks.

Theres other minor bugs related to the "end" feature, for example, it doesn't work on css or js files since those files doesn't have a ?>

Dragooon:
Leave it then I guess, killing position=end isn't really required for 2.1

MissAllSunday:
It will "encourage" mod authors to use hooks or add other methods for inserting their code so i''l be in favor of removing that but yeah, it isn't really mandatory or needed.

Oldiesmann:
Not sure what, but one of the recent PRs I merged in caused a conflict with this...

MissAllSunday:
Solved I hope...

Oldiesmann:
Yep. That fixed it.

PR #2322, opened 2014-09-14
Interfaces

MissAllSunday:
Usage of SMF_BackgroundTask as an interface.

Since we're on 5.3 why not use a proper interface, this allows custom notification classes to implement the interface while still be able to extend a third party class or even implement another third party interface, it also gives back full control of the __constructor magic method back to each notification class.

Meh... TBH this is just a selfish thing, I'm pretty sure I'm gonna be the only one implementing while also extending a third party class. Then again, most of the stuff I've been adding will not be fully used by mod authors anyway :cry:

### DIFF
--- a/Sources/tasks/CreatePost-Notify.php
+++ b/Sources/tasks/CreatePost-Notify.php
@@ -14,6 +14,11 @@
 
 class CreatePost_Notify_Background implements SMF_BackgroundTask
 {
+	public function __construct($details)
+	{
+		$this->_details = $details;
+	}
+
 	public function execute()
 	{
 		global $smcFunc, $sourcedir, $scripturl, $language, $modSettings, $language;
--- a/Sources/tasks/CreatePost-Notify.php
+++ b/Sources/tasks/CreatePost-Notify.php
@@ -12,7 +12,7 @@
  * @version 2.1 Beta 1
  */
 
-class CreatePost_Notify_Background extends SMF_BackgroundTask
+class CreatePost_Notify_Background implements SMF_BackgroundTask
 {
 	public function execute()
 	{
--- a/Sources/tasks/CreatePost-Notify.php
+++ b/Sources/tasks/CreatePost-Notify.php
@@ -14,6 +14,8 @@
 
 class CreatePost_Notify_Background implements SMF_BackgroundTask
 {
+	protected $_details = array();
+
 	public function __construct($details)
 	{
 		$this->_details = $details;
--- a/Sources/tasks/EventNew-Notify.php
+++ b/Sources/tasks/EventNew-Notify.php
@@ -16,6 +16,11 @@
 
 class EventNew_Notify_Background implements SMF_BackgroundTask
 {
+	public function __construct($details)
+	{
+		$this->_details = $details;
+	}
+
 	public function execute()
  	{
  		global $sourcedir, $smcFunc, $user_profile;
--- a/Sources/tasks/EventNew-Notify.php
+++ b/Sources/tasks/EventNew-Notify.php
@@ -16,6 +16,8 @@
 
 class EventNew_Notify_Background implements SMF_BackgroundTask
 {
+	protected $_details = array();
+
 	public function __construct($details)
 	{
 		$this->_details = $details;
--- a/Sources/tasks/EventNew-Notify.php
+++ b/Sources/tasks/EventNew-Notify.php
@@ -14,7 +14,7 @@
  * @version SMF 2.1 Beta 1
  */
 
-class EventNew_Notify_Background extends SMF_BackgroundTask
+class EventNew_Notify_Background implements SMF_BackgroundTask
 {
 	public function execute()
  	{
--- a/Sources/tasks/GroupReq-Notify.php
+++ b/Sources/tasks/GroupReq-Notify.php
@@ -16,6 +16,8 @@
 
 class GroupReq_Notify_Background implements SMF_BackgroundTask
 {
+	protected $_details = array();
+
 	public function __construct($details)
 	{
 		$this->_details = $details;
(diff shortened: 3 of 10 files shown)


Oldiesmann:
I guess I can see the usefulness of this, but it seems like a waste if you don't think anyone's going to take advantage of it, especially since we now have to declare a constructor in each class rather than in a single place. @Arantor, what are your thoughts on this?

MissAllSunday:
Its not really a waste, its not just the implements and also extends thing, an interface is what should be used here.

You can also have the __constructor in the interface to force anyone to also set it, in fact, it would be a good thing to do it.

MissAllSunday:
We can get rid of __constructors altogether by using a setter instead:

``` php
$bgtask = new $task_details['task_class']();
$bdtask->setDetails($details);
```

The interface should have a setter defined, voila, no need for any constructor unless the mod author really needs to use one, actually, this will solve my issues too since I would have the __constructor completely free and independent from the interface.

Dragooon:
I don't remember the exact task class and I'm on phone so I might be wrong
here, but you can make an abstract class implementing the interface which
defines a constructor for all. Less code that way rather than defining same
setters for everyone.

Arantor:
The interface would define what methods must provided and what signatures they should take; it would not declare the constructor. So whatever you do, you pretty much need an abstract unless redeclaring that constructor and private class members every time sounds like fun.

The same deal applies if you have a setter: it doesn't matter much what you call it, whether it's called __construct or setDetails, it's still a generic method that applies to all background tasks, for which a base background task is an ideal construct to practice DRY.

Since that's the case, the choices were 'an abstract' or 'an abstract implementing an interface', and honestly, declaring the actual execution method as an abstract has much the same practical effect as enforcing it implementing an interface. I just saved the code and memory of actually making an interface.

Elegant? Not especially. Improper? Possibly. But the point of having cron.php in the first place was to minimise the overhead, which also includes not defining things unnecessarily. The BackgroundTask serves two functions: while not just providing common functionality and ensuring that tasks fulfil a certain method's availability (which is really all an interface would get you here too), there's a slight degree of protection.

Just as the paid subs API enforces that the relevant string is part of the file before trying to load it, this at least guarantees that it is _intentionally_ a background task, in case of someone doing something they shouldn't be somehow.

FWIW: There may be some subtlety or context I'm missing but I don't understand what benefits you will get out of this. Considering they will ALL be called from the same place with the same calling information, you don't need to do anything unique in each constructor because they're all receiving the same, but even if you did need to change it you could have left it as it was, implemented a different __construct() in the inheriting class and simply called `parent::__construct()` appropriately.

And otherwise now you have to declare everything for every class rather than having an abstract that otherwise took care of the boring stuff for you.

MissAllSunday:
NVM, this has nothing to do with performance, neither an interface or an abstract class has any significant improvements over the other.

If you wan't to enforce something, use the proper tool for it. Otherwise don't. Want to save bytes? theres assembly. DRY on SMF? if you said so...

I wanted to  setting a set of rules for everyone to follow while still maintain your design freedom. Too bad no one actually sees this. I'll find a workaround for this, not a big deal..

Arantor:
Performance: the intent for cron.php was to be minimalistic. Having both an interface and an abstract seemed more bloated to me than having just an abstract which managed the same basic task of enforcing the bg-task's side of the promise.

Was it the wrong tool for the job? I didn't think so at the time. In a pure architectural sense, you're right and for something larger and more complex than a class with two methods I'd certainly be agreeing with you. But for something with two methods, going for the 'more correct' route is something akin to overkill, in my opinion. But that's all it is: an opinion.

I think I understand what you were trying to do, I just didn't see why it necessitated having to repeat yourself in every class when you could completely have your design freedom without having to change all the classes to fit - I don't get what is so wrong with how I originally built it, since I outlined how one could maintain design freedom within ones' own bg tasks should such a thing be necessary.

DRY: just because SMF doesn't practice this very well at present doesn't mean it shouldn't be done. I will admit I was trying to encourage a DRY style approach to things simply because I think that's a sane mentality to practice and it has long irked me that SMF doesn't.

I would love to see an example of the code you have in mind for this where what you did would make it easier for you to have your design freedom and what benefits it would provide. That's probably my biggest issue with this, actually: I cannot envisage a situation where someone building a task would find it easier your way than mine, but I'd love for you to show me where I'm doing it wrong.

(You are capable of that, you know. Examples of why something would be better like this go a long way, because you know what? I'm not the quickest on the up-take of things. I can't see it right now. Doesn't mean you can't show me why it would be better your way.)

MissAllSunday:
I use dependency Injection to pass my whole environment to my class, on inner classes there isn't any issue since DI takes care of that but for classes that are instantiated externally, such as this case, I need to extend my master class to be able to use my environment:

``` php
class some_background_class extends my master_class
{
    function __construct()
    {
        parent::__construct();
    }
```

There is no multiple inheritance in PHP. Thats the whole issue.

I want to follow the rules SMF_BackgroundTask  sets, I just do not want to be limited on what I can or cannot do. I want to **Implement** SMF_BackgroundTask rules. I want full freedom on how I handle my notification classes, thats one of the very primordial reasons of Interfaces usage.

Sure, I can just do this:

``` php
class some_background_class
{
    function __construct()
    {
        $this->_app = new master_class();
    }
```

But thats not the point here. SMF_BackgroundTask doesn't need to tell me _how_ I do things, only needs to tell me _what_ I need to do.

You don't need an abstract and an Interface, just an Interface.

Don't want a setter? thats fine:

``` php
return $bgtask->execute($details);
```

Now, if you find boring or a "performance bloke" having to type on every class:

$this->_details = $details; 

then thats just wrong...

> Doesn't mean you can't show me why it would be better your way.)

Whats the point? you said it yourself, just "show you"... from that to actually accept it, theres a huge gap and we both know thats not gonna happen, you will find something, somewhere to refute what I just typed, it is futile to keep arguing about this.

Issue #2318, opened 2014-09-14
delete comment on moderation report


illori:
if you try to delete a comment on a moderation report you get an error that you are not allowed to delete the comment [as admin and the comment was made by the same user]

![Alt text](https://imagizer.imageshack.us/v2/697x130q90/537/KsKUQy.png)

also the error that you can not delete the comment shows in the error log, i dont think this should show in the error log.

the following error also shows at the top of the page
Notice: Undefined index: comment in F:\wamp\www\github2.1\Sources\ReportedContent.php on line 476

there is also one more error in the log

```
Type of error: Undefined
```

8: Undefined variable: context
/github2.1/index.php?action=moderate;area=reportedposts;sa=handlecomment;rid=8
F:/wamp/www/github2.1/Sources/Subs-ReportedContent.php
Line: 590

Oldiesmann:
The "undefined variable: context" issue has already been fixed. The other one will be fixed shortly.

illori:
not fixed. Notice: Undefined variable: user_info in F:\wamp\www\github2.1\Sources\Subs-ReportedContent.php on line 509

if the user is not the author of the comments

illori:
also with this change the user can not edit their own comment, the buttons dont show up. they only show up for the admin.

please test this with an account that is not an admin.

illori:
so we have a missing global $user_info in subs-reportedcontent.php. we also have one other issue.

```
'can_edit' => allowedTo('admin_forum') || (($user_info['id'] == $row['id_member']) && allowedTo('moderate_forum')),
```

why have a comment edit feature when only those that can moderate forum members can edit the comments they have made?

illori:
related to this, editing comments does not work either for non-admin users.

PR #2329, opened 2014-09-16
fix the rest of #2318 editing/deleting comments in reports

illori:

### DIFF
--- a/Sources/ReportedContent.php
+++ b/Sources/ReportedContent.php
@@ -476,7 +476,7 @@ function HandleComment()
 		$comment_owner = $user_info['id'] == $comment['id_member'];
 
 		// Nope! sorry.
-		if (!allowedTo('admin_forum') && !$comment_owner && !allowedTo('moderate_forum')) 
+		if (!allowedTo('admin_forum') && !$comment_owner) 
 			fatal_lang_error('report_action_message_delete_cannot');
 
 		// All good!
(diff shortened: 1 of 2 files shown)


NanoSector:
Confirmed working.

Oldiesmann:
The problem is this now also makes it such that anyone with moderate_forum can edit comments made by other users. While the moderate_forum permission isn't a light one, I'm not entirely sure we should do this. If we do want to do it that way, then we don't need to check admin_forum because anyone who has admin_forum will also have moderate_forum.

illori:
It had half of that already. I just made it work.

illori:
we already had 

```
'can_edit' => allowedTo('admin_forum') || (($user_info['id'] == $row['id_member']) && allowedTo('moderate_forum')),
```

which allowed users with moderate_forum to see the buttons but they did nothing as the rest did not allow it. do we want to allow it or not? i dont see it as a major issue either way.

Oldiesmann:
The existing code limited users to editing their own comments if they had moderate_forum. Your change makes it so that they can edit ANY user's comments with the moderate_forum permission.

NanoSector:
It doesn't show the edit button for someone else's comments.

illori:
Yes it does show for anyone with that permission. I did not pick that one
to be used if you really don't like it we can change it but I don't think
it needs to be.

PR #2358, opened 2014-09-22
Allow custom mods to add "big" admin icons

Underdog-01:
- css for "big" admin icon image files added by custom mods 

### DIFF
--- a/Sources/Subs-Menu.php
+++ b/Sources/Subs-Menu.php
@@ -135,7 +135,14 @@ function createMenu($menuData, $menuOptions = array())
 						if (isset($area['icon_class']) && empty($menu_context['sections'][$section_id]['areas'][$area_id]['icon']))
 							$menu_context['sections'][$section_id]['areas'][$area_id]['icon_class'] = $menu_context['current_action'] . '-menu-icon ' . $area['icon_class'];
 						elseif (isset($area['icon']))
+						{
+							if ((substr($area['icon'], -4) === '.png' || substr($area['icon'], -4) === '.gif') && file_exists($settings['theme_dir'] . '/images/admin/big/' . $area['icon']))
+								$menu_context['sections'][$section_id]['areas'][$area_id]['icon_file'] = $settings['theme_url'] . '/images/admin/big/' . $area['icon'];
+							elseif ((substr($area['icon'], -4) === '.png' || substr($area['icon'], -4) === '.gif') && file_exists($settings['default_theme_dir'] . '/images/admin/big/' . $area['icon']))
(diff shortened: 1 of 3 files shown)


Underdog-01:
I tested this in current versions of IE, Firefox, Chrome, Safari and Opera browsers.
No issues with display of multiple additions.

live627:
Can you get rid of the blank lines at the end of the files? Then this can go in I think.

Underdog-01:
I did not add those and have no idea why it is suddenly adding them automatically.
Is this something new that I have to change in GitHub settings somehow?
I hope it did not add them again after I attempted to remove them.

Oldiesmann:
GitHub's web editor does that automatically for some reason. The only way to remove them is to edit them on your computer and use a Git client like SourceTree or TortoiseGit to commit them. The GitHub For Windows program should work for that as well but SourceTree and Tortoise are much better.

MissAllSunday:
It is considered a good practice to leave a new line at the end of files, thats why github does it, we do not follow that implementation for several reasons, mostly handling legacy code.

I actually expect this and other things gets added eventually so another round of clean up will have to be performed before a proper release is made.

XinYenFon:
I don't understand why we need this if you add another big item just add large-admin-menu-icon your-icon class to CSS and problem solved. Either point to static file or add your icon to sprite (not recommended).

Underdog-01:
what about creating a .gitconfig file and adding a [core] directive?
like this... ( ,,,,  = tab for this example since I can not use tab here in comments)

[core]
....blank-at-eof = false

.. I think that may disable how it is auto adding the line break at the end of files
or if the config file exists you can use a shell command:..
git config core.blank-at-eof = false

If I have read the git documentation correctly that should disable the error it flags for no empty line at the end of a file and therefore it will not auto add that line.

Underdog-01:
The empty line at the end of files does not occur on any of my own repositories.
If it is still occurring here and this change is acceptable perhaps someone can quickly edit those out when/if the edits are actually committed?

Antes, 

  Shouldn't the design be to facilitate mod additions and Isn't that what hooks are all about?
This edit allows a mod author to easily add large icons via the admin_areas hook.
What you suggest involves adding css to the $context['html_headers'] and/or a css file (without direct file edits)   

Regards.

Underdog-01:
There is now some kind of merge conflicts.
I will close this edit and attempt it anew.

MissAllSunday:
You can pull the recent release-2.1 branch on to yours and whatever git client you're using should tell you about the conflict and mark the edits to merge.

Or just do it from the CM: https://help.github.com/articles/resolving-a-merge-conflict-from-the-command-line

Underdog-01:
I just closed it and started a new one whereas I also figured out how to omit the end of file line breaks.
Thanks for the info Suki !!

ref. https://github.com/SimpleMachines/SMF2.1/pull/2374

PR #2374, opened 2014-09-23
Allow mod authors to add big icons to admin menu

Underdog-01:
- created new pull request due to file conflicts


Underdog-01:
ok I figured out how to avoid those lines at the end of files now.
I never used to have that issue with any previous PR's I submitted using the Git editor and do not know why it is only occurring now.
Anyhow from now on I will simply create new patches and then edit the files locally using Notepad++ which seems to work.

Regards.

XinYenFon:
I still don't understand why you can't add image? you need to use more CSS yes but I find this additions unnecessary.

Underdog-01:
Why make it necessary for mod authors to do more than add to the menu hook?
Imo the code should be designed to add to the hook without further action. The hooks are there to perform a necessary task (in this case add to the admin menu array), why should adding function to SMF involve more steps in that regard? Why not design the hook/code to perform the full task of adding to the menu?

PR #2376, opened 2014-09-24
Revert "Allow mod authors to add big icons to admin menu"

XinYenFon:
next time push something into core trying first and make sure nothing gets broken down @live627 

This PR also reverts your changes because it breaks things @Underdog-01 

---

Further readers, this PR created by "revert" function inside GitHub so cannot be signed-off


Underdog-01:
I figured out what the issue is whereas I made the mistake of using hyphens instead of underscores.
The fix will be posted shortly..

ref. https://github.com/SimpleMachines/SMF2.1/pull/2378

PR #2377, opened 2014-09-24
Someone made a typo

Underdog-01:
- missing semicolon to end PHP line


Underdog-01:
Why is there so much white-space in this file?
I was actually going to remove most of it but it appears the variables may be grouped?
If that is the case then I grouped this edit with the $txt error string below it.

Oldiesmann:
Whoops! Don't know how I missed that.

And yes, the strings are grouped. Most of the language files are like that. In this case I probably should have put that one with $txt['mc_add_note'] at the bottom, but it really doesn't matter too much.

PR #2378, opened 2014-09-24
change hyphens to underscores

Underdog-01:
Adjust css class names from using hyphens to underscores


Underdog-01:
My apologies as I initially made the mistake of using hyphens instead of underscores for css class names. This should fix any issues.

Regards.

Issue #2382, opened 2014-09-27
issue with installing packages containing hooks


Underdog-01:
Here is a test package:
http://www.web-develop.ca/index.php?action=dlattach;topic=92.0;attach=163

You will notice one error in the log for install only. 
I believe it is due to the way in which the load_file function has been configured.
Imo this was introduced by Suki from this reference:
https://github.com/SimpleMachines/SMF2.1/commit/fbd85f8920801a715ab5e1415cdbe1e99d4d91b4

The actual error message is being logged from within the call_helper function.
This occurs when a function name is piped through load_file and that function returns the function name (failing to load a file). Since the value of the return variable is not false, it proceeds further to eventually log an error.

What would be a preferred method to fix this?
Perhaps have the load_file function return an array ie.. array('file', $string) or array('func', $string) depending on it being a file or not? and/or.. some condition prior to the error logging?

Regards.

MissAllSunday:
load_file will return false if no file was located but a file was indicated on the string.

If the error is been logged by SMf it means your $string hit one of the checks, which leaves me to believe there is some sort of quirkiness in the way load_file is been used or the content of $string.

Question is, how exactly are you adding your hooks?  your package doesn't contain any files or references.

MissAllSunday:
OK, you're using plain old 2.0 calls.

Your $string been a function name doesn't make any difference, is_callable can happily accept a string.

The issue isn't on load_file() that function won't do anything if there isn't a | symbol, none of your hooks has it so it will just return the exact string your gave it to it.  Both load_file and call_helper only do things if you tell them to do things, otherwise they just return what was passed.

The issue here is how are you calling your functions and more specifically, when, many people do not know but the place and timing while installing is important, its always advisable to load any files you're gonna use before calling any function on them.

You're using 

``` xml
<readme parsebbc="true">SMF2/readme_PersonalizedBBC.txt</readme>
```

So, the hook you just added is actually been used while you're installing the mod, try to move that part **after** you've done all you have to do.

MissAllSunday:
Oh, scratch that...

You're using 

``` php
add_integration_function('integrate_bbc_codes','PersonalizedBBC_codes', true);
```

That "true" on 2.1 means "file" rather than "permanent" https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs.php#L4082

So, add_integration_function is actually adding a pipe which result in a flawed process on call_helper.

[...]

PR #2395, opened 2014-10-01
reload settings table during mod installation

Underdog-01:
- mods that alter the settings table during installation (ie. for adding hooks) need to have their possible values executed else errors may be logged during the process

ref.
https://github.com/SimpleMachines/SMF2.1/issues/2382


Underdog-01:
botched.
redoing PR

PR #2407, opened 2014-10-04
allow mod authors to add their own menu icon

Underdog-01:
- allow mod authors to dynamically add elements to the menu
- this can include background image, DOM events, etc.


Underdog-01:
I could have wrote this similar to the way I introduced custom admin "big" icons but decided to go this route to allow possible DOM event triggers (ie. changing the image onmouseover).
An example of just using this to display a custom icon for the menu (using a hook to add to the array):
'custom' => 'style="background-image:url(\'' . $settings['images_url'] . '/board.png\'); width:16px; height:16px;"',

PR #2408, opened 2014-10-05
Token fixes

Dragooon:


Sources/ManageServer.php
```diff
@@ -558,9 +557,9 @@ function ModifyLoadBalancingSettings($return_config = false)
 	if (isset($_GET['save']))
 	{
 		// Stupidity is not allowed.
-		foreach ($_POST as $key => $value)
```
Dragooon, line 561: Damn this should've stayed $_POST

PR #2410, opened 2014-10-06
Template fixes

XinYenFon:


Oldiesmann:
There's already an anchor with an ID of "top" (the forum title/logo that links to the main page of the forum), so this will cause invalid code. I suggest leaving "main_content_section" as is and just fixing "bot".

PR #2422, opened 2014-10-10
Allow access to loaded member's data 

MissAllSunday:
Just move the hook call a few lines below to get access to the sender's data.

### DIFF
--- a/Sources/Profile-View.php
+++ b/Sources/Profile-View.php
@@ -233,9 +233,6 @@ function fetch_alerts($memID, $all = false, $counter = 0)
 	}
 	$smcFunc['db_free_result']($request);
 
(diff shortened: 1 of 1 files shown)


live627:
Should $senders be passed to the hook as well?

MissAllSunday:
Well, $memberContext is already available on global scope and as much as it kills me to see that behavior, i seems its the preferred way to go.

MissAllSunday:
Oh you mean the $senders var? I dunno, that data is already included in $alerts, seems $senders was created just to gather all the senders IDs

Oldiesmann:
SMF does that so we don't have to call loadMemberData individually for each sender. There's no need to pass it to the hook since the sender IDs will already be available in $alerts.

PR #2424, opened 2014-10-12
Quote beautifying

Hristo-:
I do not like how the default quotes look like, so I made some changes on my 2.0 forum and finally decided to make a pull request for 2.1. Here how the quote will look after the changes (tested in FF, Pale Moon, Opera new and old, Chrome):
Post quote
post_quote

PM quote
pm_quote

Post quote on narrow window
post_narrow_quote

I know many people think the default theme should look simple and something like this would be better as mod. Usually I tend to agree with this, but I think quotes really need some beautifying.

My reasoning to remove the quotation mark:
1. After the changes it's much more obvious that the block is a quote, so I do not think they are needed anymore.
2. Not all languages use these opening quotation marks. For example in German and Bulgarian the opening quotation marks are these: "„". In Spanish, Russian and French are these: "«". 

Bulgarian - http://bg.wikipedia.org/wiki/%D0%9A%D0%B0%D0%B2%D0%B8%D1%87%D0%BA%D0%B8
French - http://fr.wikipedia.org/wiki/Guillemet
German - http://de.wikipedia.org/wiki/Anf%C3%BChrungszeichen
Russian - https://ru.wikipedia.org/wiki/%D0%9A%D0%B0%D0%B2%D1%8B%D1%87%D0%BA%D0%B8
Spanish - http://es.wikipedia.org/wiki/Comillas

I guess many more languages do not use "“" as opening quotation marks.
3. As image they looked good in all browsers, now every browser renders them more or less differently.


XinYenFon:
Too bad you need to sign-off your commits :/ Other than my personal opinion no to this as well.

Hristo-:
Made the commit through the site and could not figure out how to sign-off. If there is a point to sign it I will think something out.

Oldiesmann:
If you don't see an option to sign off, just add the "Signed-off-by: ..." line at the end of the commit message (that's really all that signing off does anyway). @Antes, why do you say no to this? I think it looks nice.

XinYenFon:
Breaks the simplicity of the theme and its not looking good.

Hristo-:
Good to know that, Oldiesmann. Thank you!

I really have no desire to argue which one is more user friendly and/or looks better so it's up to you guys and gals. If you decide this one is worth it I will re-commit.

illori:
IMO i think this should be a mod and/or theme change. what we have now looks good enough to me. if the admin of the forum wants it changed they can tweak it or change themes to something they like better.

maybe our quote could use a better way to identify the links as links and not just plain text. in 2.0 it is in bold in 2.1 this has been lost.

Oldiesmann:
Going to close this since most don't seem to like the proposed changes.

PR #2425, opened 2014-10-12
HTML5 Validation Improvement(s)

XinYenFon:


Underdog-01:
I am glad to see some HTML5 css compliance being applied but please note that ie. the use of width within a tag and strong tags are also more or less deprecated. These should also be applied as css class else use direct style elements.  
Imo a lot of HTML5 css compliance is still pending.

XinYenFon:
As much as time allows me, my prio will be the removing this deprecated things and use CSS others are just extra sorry to say.

Underdog-01:
I may have the time to apply css fixes for all the templates tomorrow since I have the day off. 
Although I will only do this after/when this PR is dealt with to thwart file conflict issues.
Testing every template through a filter (HTML tidy at HTML5 setting) is fairly basic but this will take time as you say.

XinYenFon:
You can send pr to my repo as well, pretty much same thing you contribute to pr with your commits, beside opening pr. I'll work on this tomorrow more, I'm currently having sleep issues tbh it's nearly 7 am :/ and can't sleep.

-----Original Message-----

Issue #2360, opened 2014-09-22
Default avatar not showing in the topheader miniprofile


ahrasis:
Profile and display page do show default avatar automatically but topheader miniprofile doesn't show it automatically. It has to be chosen in profile page. Is this intended?

XinYenFon:
Not really sure (can't remember) but isn't better not to show default avatar on that place, that'll remind user that they don't have avatar? (or I'm thinking soooo weird which is a very normal thing)...

ahrasis:
Not weirdo or something but I just think that would defeat the purpose of default avatar. Might as well do not show it elsewhere too.

jsqx:
People bring back display avatar in mini popup. In SMF 2.0 avatar is display at top profile, here should be the same, this miniature, which is now isn't fine.

![avatar](https://cloud.githubusercontent.com/assets/8003629/4377563/106733ea-4350-11e4-9a54-f472dae63b39.png)

XinYenFon:
It was like that but changed. I don't think we revert again.

Oldiesmann:
It should be left where it is for now. If after the first beta more people prefer it where it was, we can change it back later.

jsqx:
Why you off-load something what looks good? Mini Popup is empty now without display avatar and bring it back soon.

ahrasis:
This already fixed but awaiting PR to be completed. Please close this once it is completed or I will close it, whoever the earlier.

Oldiesmann:
Fixed in #2428

PR #2428, opened 2014-10-13
Default avatar not showing in the topheader miniprofile #2360

ahrasis:
This is to fix issue #2360 where default avatar not showing in the
topheader miniprofile.

https://github.com/SimpleMachines/SMF2.1/issues/2360

Fix is not at the theme level as suggested earlier but Subs.php. IMO
this will ensure that no bug for having no avatar since default avatar
is supposed to be always in place.

Do check.

### DIFF
--- a/Sources/ManageSettings.php
+++ b/Sources/ManageSettings.php
@@ -315,7 +315,7 @@ function ModifyLayoutSettings($return_config = false)
 			array('int', 'defaultMaxMembers'),
 		'',
 			// Stuff that just is everywhere - today, search, online, etc.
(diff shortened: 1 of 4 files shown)


Oldiesmann:
You need to sign off on your commits or we can't accept them.

ahrasis:
Signed-off-by: Ahmad Rasyid Ismail ahrasis@gmail.com

Oldiesmann:
Unfortunately you've now added completely unrelated commits to this PR. Unless the commits are related, you should either work on one thing at a time or create a separate branch for each thing you want to work on, so you can create separate PRs for each thing.

ahrasis:
I think I better delete this and create branches for each one of them. I am still learning how github actually works

PR #2429, opened 2014-10-13
adjust bitmask & get_html_translation_table output

Underdog-01:
- the function of get_html_translation_table actually filters spaces to the entity of & # 39; (without the 0) so this has to be added to the replacement array
- the default document charset encoding varies dependant on PHP version and should be specified 

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -793,7 +793,7 @@ function un_htmlspecialchars($string)
 	static $translation = array();
 
 	if (empty($translation))
-		$translation = array_flip(get_html_translation_table(HTML_SPECIALCHARS, ENT_QUOTES)) + array('&#039;' => '\'', '&nbsp;' => ' ');
+		$translation = array_flip(get_html_translation_table(HTML_SPECIALCHARS, ENT_QUOTES | ENT_HTML5)) + array('&#039;' => '\'', '&#39;' => '\'', '&nbsp;' => ' ');
 
 	return strtr($string, $translation);
 }
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -790,10 +790,11 @@ function timeformat($log_time, $show_today = true, $offset_type = false)
  */
 function un_htmlspecialchars($string)
 {
+	global $context;
 	static $translation = array();
 
 	if (empty($translation))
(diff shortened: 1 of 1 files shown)


Underdog-01:
I just noticed that ENT_HTML5 was only introduced in PHP 5.4.0+
Shall I remove that bitmask parameter?
What is the minimal PHP version to use SMF 2.1 now? Is it 5.4 whereas I can leave that part as is?

MissAllSunday:
Its 5.3.8

Underdog-01:
Suki,
Thank you.

I have removed the edit for the bitmask parameter as it is only compatible with PHP 5.4.0+
If the SMF 2.1 minimal PHP version is bumped up at some point in time for whatever reason I will reintroduce that specific bitmask parameter.

Underdog-01:
perhaps more is required here.
Please note that function also performs different default encoding dependent on PHP version.
PHP 5.4.0+ encodes it as UTF-8 while anything prior to that encodes it at ISO-8859-1.
That encoding paramter can be put within the function to override the default behavior.

MissAllSunday:
SMF uses $context['charset'] to overwrite whatever default charset a function has.

Underdog-01:
Then that should also be added to this function which I will do right now..
Do note that it is $context['character_set']

Underdog-01:
If you look at the reference for that function (supported charsets):
http://php.net/manual/en/function.get-html-translation-table.php#refsect1-function.get-html-translation-table-description

.. you will see that only certain ones are accepted else it will go the the default and display an error. 

Would it be prudent to perhaps fill an array with the accepted charsets and if the current smf charset setting is not in that array then force it to default to UTF-8 ?? 

This will thwart any possible errors if other charsets are allowed/possible for SMF 2.1

MissAllSunday:
As far as I know, all the charsets SMF uses are included on that list since we haven't see any mayor issue (I remember a few, isolated ones on mantis, cannot remember them all) on previous versions and usage of  htmlspecialchars() which was used on several places pretty much rawly.

Underdog-01:
I will leave this PR as is then since I see no further necessary edits to it.
This fix is required for use of the un_htmlspecialchars function. 
Tested while installing mods that use it whereas it flagged an error which is what drew my attention to the issue. 
Regards.

PR #2433, opened 2014-10-14
Relative Time And Date For SMF 2.1

ahrasis:
I would love to see Relative Time And Date For SMF 2.1. Since I wrote
the said mod for SMF 2.0, I wish it'll become a part of SMF 2.1.

### DIFF
--- a/Sources/ManageSettings.php
+++ b/Sources/ManageSettings.php
@@ -315,7 +315,7 @@ function ModifyLayoutSettings($return_config = false)
 			array('int', 'defaultMaxMembers'),
 		'',
 			// Stuff that just is everywhere - today, search, online, etc.
-			array('select', 'todayMod', array($txt['today_disabled'], $txt['today_only'], $txt['yesterday_today'])),
+			array('select', 'todayMod', array($txt['today_disabled'], $txt['today_only'], $txt['yesterday_today'], $txt['rltv'])),
 			array('check', 'onlineEnable'),
 		'',
 			// This is like debugging sorta.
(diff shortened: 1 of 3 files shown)


live627:
mehh...

XinYenFon:
why? Its better when its a mod tbh...

ahrasis:
May be so, may be not. Anyway, it is up to you all.

XinYenFon:
There must be a reason other than “I wrote this for SMF x and I’d be very happy to see in core SMF y”. Dozens of mods (more popular than this feature) waiting outside but we are not adding them to core, because they don’t fit to core.

Don’t get me wrong, we are grateful to all your contribution(s). Thanks 😊

From: Ahmad Rasyid Ismail
Sent: ‎Wednesday‎, ‎October‎ ‎15‎, ‎2014 ‎3‎:‎36‎ ‎AM
To: SimpleMachines/SMF2.1
Cc: Mert Alınbay

May be so, may be not. Anyway, it is up to you all.

ahrasis:
Why? From what I can see this has been useful and popular to many in various other forums and platforms. Most people don't even bother about details of time and date anymore. If they need so, it is already SMF default anyway. So why not give people an extra option to display them relatively instead? Really, I don't mind at all whether this become a part of future SMF. It is also can be done in a lot of better ways by SMF team. This is just an idea. The code is nothing really.

MissAllSunday:
Perhaps turning this into a mod and see how it goes?  how popular it is?

If this is already a mod, what are its statics? how's the mod going?

There is still time to implement something like this but I would like to see if it will be relevant to a majority of users

Oldiesmann:
http://custom.simplemachines.org/mods/index.php?mod=3892 - only 298 downloads since July.

live627:
Also of note, one made by me - http://custom.simplemachines.org/mods/index.php?mod=2432

Not very popular.  Could this request be closed now?

Oldiesmann:
Closing this due to lack of evidence that this would be a popular feature among users.

PR #2434, opened 2014-10-14
QuickReply Code Moved To Function Template

ahrasis:
This is to ease modification which moves the quickreply box. With this,
the quickreply code is now moved into a function template_quickreply().

This function thus allows quickreply box to be moved anywhere without
moving the whole quick reply code. They justneed to find and move
template_quickreply(); to where they want it to be in Display page.

### DIFF
--- a/Themes/default/Display.template.php
+++ b/Themes/default/Display.template.php
@@ -213,100 +213,8 @@ function template_main()
 	echo '
 			<div id="display_jump_to">&nbsp;</div>';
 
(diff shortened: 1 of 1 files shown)


MissAllSunday:
@ahrasis Can you made the changes Oldiesmann propose so we can merge this in?

Thanks.

live627:
Is this going in?

XinYenFon:
only if OP does it right way.

PR #2436, opened 2014-10-14
Add Info Center Permission

ahrasis:
If this is useful, I wish to commit the info center permission for SMF 2.1.


XinYenFon:
I'd rather avoid adding new settings, better we look ways to remove some existing ones.

Oldiesmann:
I don't think this is necessary either. We have enough permissions as it is and I don't think many people will find this one useful.

ahrasis:
You may be right. It is just my thought that this could reduce support after seeing this mod http://custom.simplemachines.org/mods/index.php?mod=1335 was in high demand anyway.

Oldiesmann:
Is there still a high demand for it? That mod hasn't been updated in 2 years. I haven't seen any support requests for that sort of thing recently either.

Dragooon:
Yeah I'd like to avoid this as well, this is best as a mod.

PR #2440, opened 2014-10-18
Delete alerts

MissAllSunday:
Adds some new functions to showalerts page:
- Delete and mark as un/read buttons for every alert.
- Mass delete or un/read alerts
- Pagination

Also adds some generic functions on Profile-Modify.php

### DIFF
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -2099,6 +2099,44 @@ function alert_markread($memID)
 	updateMemberData($memID, array('alerts' => 0));
 }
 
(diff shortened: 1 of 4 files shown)


Dragooon:
Does this fix the issues Bigguy's been having?

smfbigguy:
I have not tried this yet but will be doing an update of my site in the next couple days. So we will see by then. I hope it does.

MissAllSunday:
Nope, not really, it just adds a way to delete alerts.

Oldiesmann:
Is this good to go now?

MissAllSunday:
yes

PR #2455, opened 2014-10-24
! Don't use the same ID for multiple elements (fixes #2450)

### LINKED ISSUE #2450: Validation errors
I'm getting some validation errors in several parts of my test board.
- Admin (main page)
  Line 599, Column 33: Element object is missing one or more of the following attributes: data, type.
  `<object id="quick_search">`
  Line 600, Column 137: Element form not allowed as child of element object in this context. (Suppressing further errors from this subtree.)
  …ion=admin;area=search" method="post" accept-charset="UTF-8" class="floatright">

Oldiesmann:
Note that this is still a work in progress. I have verified that it works for the instances mentioned in #2450, but have not verified if it's a problem anywhere else yet. I need to make sure this doesn't break anything else before it gets pulled in.

### DIFF
--- a/Themes/default/Display.template.php
+++ b/Themes/default/Display.template.php
@@ -194,7 +194,7 @@ function template_main()
 	// Show the page index... "Pages: [1]".
 	echo '
 			<div class="pagesection">
-				', template_button_strip($context['normal_buttons'], 'right'), '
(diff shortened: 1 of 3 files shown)


Oldiesmann:
Closing this as further investigation has shown that the ID isn't actually used anywhere.

PR #2467, opened 2014-10-28
Stuff

MissAllSunday:
Just some stuff I encounter when using the tools for integrating alerts and like system.
- Add a generic getter for mod authors to access like properties, either via the hook or the callback. This way the property cannot be mutated by mod authors.
- Add a custom_alert param to _validLikes so mod authors can disable the generic alert that is created on every like, this allows mod authors to create their own custom alert with their own custom arguments
- Make _validLikes to be overwritten by integrate_valid_likes hook instead of been completely replaced, this allows mod authors to change whatever default value they want to change without having to re-declare all options.

Will probably add more as I finish the notification stuff on my mod so I'll leave this open for now. I also need to add an "are you sure?" check to delete/mark read alerts too.

I'm still unsure about the custom_alert option as this could lead to some nasty results if 2 or more mods uses this option or if a mod wants the generic alert but another one does not... better go the safe route and do a

if ($this->_type == 'msg')

like the rest of the code does.

### DIFF
--- a/Sources/Likes.php
+++ b/Sources/Likes.php
@@ -21,40 +21,40 @@ class Likes
 	/**
 	 *@var boolean Know if a request comes from an ajax call or not, depends on $_GET['js'] been set.
 	 */
-	protected $_js = false;
+	public $js = false;
 
 	/**
-	 *@var string If filled, its value will contain a string matching a key on a language var $txt[$this->_error]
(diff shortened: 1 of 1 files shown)


MissAllSunday:
Oh, I still wanted to add some stuff :P 

I add them in another PR.

Issue #1768, opened 2014-06-05
Text pasted into post can create bad font BBC


AngelinaBelle:
I copied the Chrome version from another browser window into the WYSIWYG post editor.
The result was 
[font='Segoe UI', Tahoma, sans-serif]33.0.1750.146, Windows 7[/font]

Which rendered as shown above, because the font BBC cannot accept parameters in that format.
It kind of spoils the wysiwyg thing.  It always looks fine in the wysiwyg editor, but never after parse_BBC gets hold of it.

http://test2.turtleshellprod.com/index.php?topic=34.0

Oldiesmann:
That sounds like more of an issue with the bbcode parser/editor itself rather than SMF's implementation of the editor. Might want to file an issue on the SBBCodeParser project (https://github.com/samclarke/SBBCodeParser) instead.

AngelinaBelle:
Me?  I am not sure how to tell whether this is an SBBCodeParser issue or whether it has to do with how SMF has implemented it, defined the font tag, or what have you.

I can't even find add_bbcode in the repo here...

If anyplace, the html -> bbcode magic seems to happen in jquery.sceditor.bbcode.js on line 941 (?). I don't understand how the html attribute got so hashed,though.  It is as though additional string were inserted later, and the quotes were not stripped.

AngelinaBelle:
So [font='Segoe UI', Tahoma, sans-serif]Version[/font] is bad, but
[font=Segoe UI, Tahoma, sans-serif]Version[/font] is better, and results in <font style="font-family: Segoe UI, Tahoma, sans-serif;" ... >.  

Perhaps the rule for creating the [font] tag from html could be re-defined to strip all single-quotes since parse_bbc clearly does not like them.

I don't know if this is actually designed behavior in the sBBCodeParser or not, so I have started with a question at http://www.sceditor.com/documentation/plugins/bbcode/#comment-1432547366

AngelinaBelle:
This is designed behavior for sceditor.
Because some font names contain spaces, they must be enclosed in quotes.
Giving a comma-separated list of several font families is valid CSS -- http://www.w3schools.com/CSSref/pr_font_font-family.asp

Because the SMF font tag is designed for only one font family, there is an incompatibility between  SBBCodeparser and  SMF.
There are two choices to fix it:  change parse_bbc or change SBBcodeparser.

[...]

PR #2490, opened 2014-11-06
font family Signed-off-by: snrj <snrj_od_@hotmail.com>

snrj:
fix
Signed-off-by: snrj snrj_od_@hotmail.com

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -1215,7 +1215,7 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 			array(
 				'tag' => 'font',
 				'type' => 'unparsed_equals',
-				'test' => '(([A-Za-z0-9_\-]+|&#39;.*&#39;),\s*)+([A-Za-z0-9_\-]+|&#39;.*&#39;)\]',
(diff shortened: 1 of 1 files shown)


MissAllSunday:
I get a Warning: preg_match(): Compilation failed: unmatched parentheses  on PHP 5.5.12 when I apply this change

MissAllSunday:
Is there any specific reason as to why we aren't using the old regex from 2.0? Seems to still do the trick, less convulsive, easier to read too.

Been searching but cannot find anything, I suppose is because the way the new editor works or some compatibility with old BBC with simple commas?

live627:
The one in 2.0 does not support commas.

MissAllSunday:
And where does this commas come from? Why was it necessary to add support for commas? is there any bug report you can link?

Again, I'm trying to find any relevant info so I can fix this, if you have any relevant info, please provide as much details as you can.

Oldiesmann:
See #1768...

MissAllSunday:
It seems that is a different issue?  the tag font has been broken for way longer than that bug report.

If the change to the font tag was made to address that issue, I would assume whoever did that would had closed the bug report on a commit or something like that.

PR #2492, opened 2014-11-07
! Pop_content size does not need to be that big fixes #2485

### LINKED ISSUE #2485: large font size in help popup
is there any reason why we need a larger then normal font size in the help popup? i think 1em is a better size then 1.11em that it is now.

the larger font IMO makes it almost "scream" at you when you view the popup.

XinYenFon:
! Linktree & topic info needs to be close to each (highschool sweethearts?)

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -2759,9 +2759,6 @@ dl.addrules dt.floatleft {
 }
 .popup_content {
 	color: #222;
-	/* @todo Can't really see any reason for pop-up text to be smaller than standard body text.*/
-	/* Perhaps we should just remove the descritpion class from the div in the template. */
-	font-size: 1.11em;
 	line-height: 1.6em;
 	max-height: 20em;
 	overflow: auto;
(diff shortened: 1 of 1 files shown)


Oldiesmann:
Why are you killing .dropmenu? That class is used for the menus.

XinYenFon:
That classes are not used in menus they are defined by @Dragooon for Follow/Alert section. They are actually breaking some parts of the menu as well.

Oldiesmann:
.dropmenu, .dropmenu a and .dropmenu a:hover are indeed used.

Line 26 of GenericMenu.template.php

```
                    <ul class="dropmenu" id="dropdown_menu_', $context['cur_menu_id'], '">';
```

Oldiesmann:
Looking at this further at your request... I didn't know that we had already defined .dropmenu earlier in the CSS. According to Chrome, the only part of the 2nd .dropmenu that's used is the background color (but toggling that doesn't change the look at all), so we should be good merging this in.

PR #2493, opened 2014-11-10
cleaning some stuff and leftovers

XinYenFon:
Yey for one less setting? :P After @Dragooon 's changes on notification setting(s) on Board/Topic level this alone setting is just not needed...

:feet: @GitHub not cool to change :paw_prints to feet...


Sources/Notify.php
```diff
@@ -188,7 +188,7 @@ function TopicUnwatch()
 			$mode = (int) $_GET['mode'];
 			$alertPref = $mode <= 1 ? 0 : ($mode == 2 ? 1 : 3);
 
-			if (!$modSettings['enable_unwatch'] && empty($mode))
+			if (empty($mode))
```
Oldiesmann, line 191: This line and the one below it can be removed now since we always have "unwatched" topics now.

PR #2496, opened 2014-11-15
Generic fixes

### LINKED ISSUE #1545: Glow & shadow BBC
While glow actually makes a drop-shadow (1 right - 1 Down & no glow), the shadow makes a shadow (1 up - 1 left).

A shadow should go right & Down, while glow should not move.

Fix 4 glow (Subs.php line # 1208):´

```
<span style="text-shadow: $1 0px 0px 3px">
```

Fix 4 shadow:
Let the standard shadow "side" be "right", which is more normal.

Another major issue is that the current BBC checks for IE.

```
'before' => isBrowser('ie') ?
```

All that code for IE, doesn't Work, at least not in IE 10 as well as emulated IE7-9. AFAIR, IE9-11 does support normal text-shadow.
I was also under the impression, support 4 IE browsers older than IE9 would be dropped, so the Whole "isBrowser(ie)" for those BBC's could be dropped entirely.

So glow could be entirely rewritten:

```

            array(
                'tag' => 'glow',
                'type' => 'unparsed_commas',
                'test' => '[#0-9a-zA-Z\-]{3,12},([012]\d{1,2}|\d{1,2})(,[^]]+)?\]',
                'before' => '<span style="text-shadow: $1 0px 0px 3px">',
                'after' => '</span>',
            ),
```

And something similar for the shadow BBC.

### LINKED ISSUE #1684: Remove useless/unused bbc's
Old discussion URL - http://www.simplemachines.org/community/index.php?topic=516416

I think its about time we really remove the obsolute bbc's that 99.9% of SMF users don't use. Such as glow, shadow, marquee, and FTP. This will help make the slow parse_bbc a bit faster and can open up the possibility for more useful bbc's in the future. (Like a spoiler bbc for instance.)

### LINKED ISSUE #2453: smileys issues
old smileys drop down should go away.
"These smileys will appear in the popup smiley box, when posting a new forum message or Personal Message. " text in admin panel needs to be corrected.
ideally the [more] button should look like a button, same with the close link
the popup should be more centered in the window rather then left aligned.
on mobile browser the scroll bar in the popup does not show, could we add the same that shows for alerts so users will know they can scroll?

also is it possible to allow the popup to be moved and resized?

XinYenFon:
! Glow changes applied (fixes #1545)
! Shadow changes applied
! Text changes on popup/more (fixes #2453)

Since glow/shadow get some changes no need to remove them (they look fine enough) :P fixes #1684


Dragooon:
The change for schemaless URLs won't actually allow having [url=//site.com], would it (which is what @Arantor has created the issue about)? It'll simply change default schema from http:// to whatever the schema the user is accessing the site from, which can cause problems for SSL users.

Oldiesmann:
That's my understanding.

Oldiesmann:
Since SMF appears to already match URLs without no schema and no "//" on them, theoretically all we'd have to do to support this is just add `&& strpos($data[0], '//') !== 0` to those conditional statements.

Arantor:
I really don't think it's quite that simple. Take a look at fixTags.

XinYenFon:
yea that seems bit _lovely_

XinYenFon:
How about we add protocols as http:// https:// and // since 
`[html]<a href="//mysite.com">MySite</a>[/html]` has no problem... I'm not sure how this $replace thing works... I need to look for it too... (not gonna lie bit confused)...

Arantor:
Of course the html tag version works without any problems, it's completely unchecked for such things.

You will need to be more creative in fixing it than that. I will see if I fixed this at some point in the past, because I think I did.

Issue #2545, opened 2014-11-30
Profile fields (aim, icq, msn, yim, location, gender) problems


nwsw:
There are a couple of issues with the custom profile fields:
1. None of the new custom profile fields stored in `{db_prefix}themes` get migrated over during an upgrade.
2. The gender field apparently needs to be specifically converted from an integer to a string.
3. When a user changes their profile, all of the custom profile fields get added to the database, even if they are all empty.

nwsw:
The primary upgrade problem is caused by the following code in the 2.1 database upgrade scripts:

``` php
// We cannot do this twice
if (@$modSettings['smfVersion'] < '2.1')
```

The `$modSettings['smfVersion']` is actually set to **2.1 dev0**, which will not pass this test. This check appears in multiple locations in the 2.1 database upgrade script.

With regard to the custom profile fields, the following check seems to be a satisfactory replacement:

``` php
// We cannot do this twice
// First see if we still have a gender column
$results = $smcFunc['db_list_columns']('{db_prefix}members');
if (in_array('gender', $results))
```

Dragooon:
The fix for this was merged

PR #2546, opened 2014-11-30
Fix the upgrade migration of custom profile fields in #2545

nwsw:
The upgrade_2-1_postgresql.sql changes are untested, but do match the changes to upgrade_2-1_mysql.sql that are tested and confirmed to fix the upgrade issues in #2545.

Signed-off-by: Noteworthy Software, Inc online@noteworthysoftware.com

### DIFF
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -859,7 +859,7 @@ if (in_array('gender', $results))
 		FROM {db_prefix}members');
 
 	$inserts = array();
-	$genderTypes = array(1=>"Male",2=>"Female");
+	$genderTypes = array(1 => 'Male', 2 => 'Female');
 	while ($row = $smcFunc['db_fetch_assoc']($request))
 	{
 		if (!empty($row['aim']))
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -850,13 +850,16 @@ INSERT INTO `{$db_prefix}custom_fields` (`col_name`, `field_name`, `field_desc`,
 ---# Converting member values...
 ---{
 // We cannot do this twice
-if (@$modSettings['smfVersion'] < '2.1')
+// First see if we still have a gender column
+$results = $smcFunc['db_list_columns']('{db_prefix}members');
+if (in_array('gender', $results))
 {
 	$request = $smcFunc['db_query']('', '
 		SELECT id_member, aim, icq, msn, yim, location, gender
 		FROM {db_prefix}members');
 
 	$inserts = array();
+	$genderTypes = array(1=>"Male",2=>"Female");
 	while ($row = $smcFunc['db_fetch_assoc']($request))
 	{
 		if (!empty($row['aim']))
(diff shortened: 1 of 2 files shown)


other/upgrade_2-1_mysql.sql
```diff
@@ -850,13 +850,16 @@ INSERT INTO `{$db_prefix}custom_fields` (`col_name`, `field_name`, `field_desc`,
...
+// First see if we still have a gender column
+$results = $smcFunc['db_list_columns']('{db_prefix}members');
+if (in_array('gender', $results))
 {
 	$request = $smcFunc['db_query']('', '
 		SELECT id_member, aim, icq, msn, yim, location, gender
 		FROM {db_prefix}members');
 
 	$inserts = array();
+	$genderTypes = array(1=>"Male",2=>"Female");
```
Dragooon: This needs to be $genderTypes = array(1 => 'Male', 2 => 'Female');, we don't use double quotes for simple strings and have space between the assignment operator

other/upgrade_2-1_postgresql.sql
```diff
@@ -907,13 +907,16 @@ INSERT INTO {$db_prefix}custom_fields (col_name, field_name, field_desc, field_t
...
+// First see if we still have a gender column
+$results = $smcFunc['db_list_columns']('{db_prefix}members');
+if (in_array('gender', $results))
 {
 	$request = $smcFunc['db_query']('', '
 		SELECT id_member, aim, icq, msn, yim, location, gender
 		FROM {db_prefix}members');
 
 	$inserts = array();
+	$genderTypes = array(1=>"Male",2=>"Female");
```
Dragooon: Same as above

other/upgrade_2-1_mysql.sql
```diff
@@ -850,13 +850,16 @@ INSERT INTO `{$db_prefix}custom_fields` (`col_name`, `field_name`, `field_desc`,
 ---# Converting member values...
 ---{
 // We cannot do this twice
-if (@$modSettings['smfVersion'] < '2.1')
```
Oldiesmann, line 853: If this is removed, the query to pull the IM fields, location and gender will fail if those columns aren't there. That's specifically why we have that check there.

nwsw, line 853: If you don't remove it, the included code will never be executed.

The other places that use this same condition are broken.for the same reason as described in #2545. The broken places are from the following:  dc573be2bc7565e3ba5c9b985115cc2931b3a0ed, 3e0417c56016bcae7859ccdbe93590ad88e1b25d, and af915c8fee045a0454233922556c94e5d67dbd9f.

If you do not uderstand why this is an issue, then you should review the following area of the upgrade.php script:

``` php
    // All possible files.
    // Name, <version, insert_on_complete
    $files = array(
        array('upgrade_1-0.sql', '1.1', '1.1 RC0'),
        array('upgrade_1-1.sql', '2.0', '2.0 a'),
        array('upgrade_2-0_' . $type . '.sql', '2.1', '2.1 dev0'),
        array('upgrade_2-1_' . $type . '.sql', '3.0', SMF_VERSION),
    );
```

Oldiesmann, line 853: I didn't know that was done. The easiest way to get around this issue is to only execute those particular sections of code if the version is 2.1 dev0.

Dragooon, line 853: Wouldn't checking for columns like it is being done in this PR be enough?

Oldiesmann, line 853: I didn't even notice that somehow. That should work for this at least. We'll still need to adjust anything else that checks for version < 2.1 in this file.

PR #2563, opened 2014-12-05
!Bug fix default author email $thisPackage->fetch to $listing->fetch

VBGAMER45:
!Bug fix default author email $thisPackage->fetch to $listing->fetch

Signed-off-by: Jonathan Valentin vbgamer45@gmail.com

### DIFF
--- a/Sources/PackageGet.php
+++ b/Sources/PackageGet.php
@@ -307,7 +307,7 @@ function PackageGBrowse()
 	if ($listing->exists('default-author'))
 	{
 		$default_author = $smcFunc['htmlspecialchars']($listing->fetch('default-author'));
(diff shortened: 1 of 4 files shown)


live627:
Looks good to me, but please remember to keep related changes in future requests.
(Each commit here seems unrelated)

Thanks.

VBGAMER45:
Yeah I don't know how that happened. I committed each one separately. I did the pull request online at github after the first commit. Then it seems to have grouped the next two commits in the same request...

Oldiesmann:
Git operates on the assumption that anything committed to a particular branch is related, so if you commit multiple things to the same branch, they'll all get attached to the most recent open PR (regardless of when that PR is opened relative to when things are committed). In order to avoid this, create a separate branch for each set of related changes (or just focus on one thing at a time and wait for the PR to be handled before committing anything else to that branch).

VBGAMER45:
Ok thanks that helps will do that for the future.

PR #2589, opened 2014-12-08
! Slider update: 1.2.1 > 1.2.7

XinYenFon:


live627:
I'm testing this now and have found that the last two commits are not needed. Furthermore, I would use the below in the template:

```
    <script>
        jQuery("#smf_slider").slippry({
            pause: ', $settings['newsfader_time'],',
            captions: false,
            controls: false,
            useCSS: true
        });
    </script>';
```

XinYenFon:
I was actually going for a cleanup for JS file, I'm aware of those. But that also works. :P

XinYenFon:
@live627 my job is here done. Please test and let me know :)

Themes/default/BoardIndex.template.php
```diff
@@ -35,7 +35,10 @@ function template_newsfader()
 		</ul>
 		<script>
 			jQuery("#smf_slider").slippry({
-				speed: ', $settings['newsfader_time'],'
+				speed: ', $settings['newsfader_time'],',
```
live627: Speed is the duration during which it fades; pause is the interval between the transitions and seems to be the one we want to stay true to the admin setting..

Themes/default/BoardIndex.template.php
```diff
@@ -35,7 +35,10 @@ function template_newsfader()
...
+				adaptiveHeight: 0,
```
live627, line 39: Not a bool.

XinYenFon, line 39: I used original values from JS, in this version its using !0 or 0, instead true/false.

PR #2608, opened 2014-12-17
MySQL Storage Engine support

Oldiesmann:
This big collection of changes does the following:
- Use InnoDB (or, for log_floodcontrol and log_online, MEMORY) by default for new installs (if InnoDB is available - otherwise falls back to MyISAM)
- Add support for specifying a storage engine with $smcFunc['db_create_table']
- Add support for Fulltext indexes for InnoDB tables when running MySQL 5.6.4 or newer
- Add storage engine to array returned by smf_db_table_structure (currently not used in SMF, but some may find it useful)
- Minor code cleanup here and there

Note: In order to allow log_online to be MEMORY, the "url" column in this table was changed from text to varchar(1024). For consistency, I also made this change in PostgreSQL even though there is no benefit for doing so.

### DIFF
--- a/Sources/DbPackages-mysql.php
+++ b/Sources/DbPackages-mysql.php
@@ -86,7 +86,7 @@ function db_packages_init()
  * @param string $table_name The name of the table to create
  * @param array $columns An array of column info in the specified format
  * @param array $indexes An array of index info in the specified format
- * @param array $parameters Currently not used
+ * @param array $parameters Extra parameters. Currently only 'engine', the desired MySQL storage engine, is used.
  * @param string $if_exists What to do if the table exists.
  * @param string $error
  * @return boolean Whether or not the operation was successful
@@ -94,6 +94,8 @@ function db_packages_init()
 function smf_db_create_table($table_name, $columns, $indexes = array(), $parameters = array(), $if_exists = 'ignore', $error = 'fatal')
 {
 	global $reservedTables, $smcFunc, $db_package_log, $db_prefix, $db_character_set;
+
+	static $engines = array();
 
 	// Strip out the table name, we might not need it in some cases
 	$real_prefix = preg_match('~^(`?)(.+?)\\1\\.(.*?)$~', $db_prefix, $match) === 1 ? $match[3] : $db_prefix;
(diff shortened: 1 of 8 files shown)


Sources/DbPackages-mysql.php
```diff
@@ -145,7 +147,29 @@ function smf_db_create_table($table_name, $columns, $indexes = array(), $paramet
...
+		}
+
+		$smcFunc['db_free_result']($get_engines);
+	}
+
+	// If we don't have this engine, or didn't specify one, default to InnoDB or MyISAM
+	// depending on which one is available
+	if (!isset($parameters['engine']) || !in_array($parameters['engine'], $engines))
+	{
+		$parameters['engine'] = in_array('InnoDB', $engines) ? 'InnoDB' : 'MyISAM';
```
live627, line 169: So then I can only use either InnoDB or MyISAM when creating a table?

other/upgrade_2-1_mysql.sql
```diff
@@ -1710,13 +1710,17 @@ WHERE variable='enableOpenID' OR variable='dh_keys';
...
----# Changing url column size from 255 to 1024
+---# Changing url column size in log_spider_hits from 255 to 1024
 ALTER TABLE {$db_prefix}log_spider_hits
 CHANGE `url` `url` varchar(1024) NOT NULL DEFAULT '';
 ---#
 
+---# Changing url column in log_online from text to varchar(1024)
+ALTER TABLE {$db_prefix}log_online
+CHANGE `url` `url` varchar(1024) NOT NULL DEFAULT '';
+
```
live627, line 1723: Will the upgraded sessions table stay in MyISAM then?

Sources/DbPackages-mysql.php
Oldiesmann, line 169: No. That code just chooses the default storage engine if you don't specify one or the one you specified isn't supported.

other/upgrade_2-1_mysql.sql
Oldiesmann, line 1723: Yes. We're not changing the storage engine for upgrades in part because it might take quite a while to do so on larger forums.

live627, line 1723: Why not kill all sessions? It's no big deal... at best, user won't notice, and at worst, they need to log in again.

Oldiesmann, line 1723: Because there's no point in doing so for just one table when that would still leave a bunch of other tables with different storage engines compared to what you'd get with a fresh install.

PR #2617, opened 2014-12-22
Stuff

MissAllSunday:
Adds the remaining custom profile fields places to the summary page

This was somehow confusing:

if ($field['placement'] == x || empty($field['output_html']))

So I change it to

if ($field['placement'] == x && !empty($field['output_html']))

Which seems more obvious to me unless I'm missing something.


Oldiesmann:
The new code is basically doing the same as the old code, just in a slightly different way...

Original code:
continue if placement isn't 0 or output_html is empty

New code:
display field if placement is 0 and output_html isn't empty

However, there's a slight problem with your new code. Originally it would only output the `<dl>` and `</dl>` if there was actually something to show (and the field hadn't already been shown). Now it's always spitting out the `<dl>` and `</dl>` before we even check whether the field should be shown.

MissAllSunday:
I don't really follow what you're saying. Yes, the new code is doing the same as the old code, I just removed the redundancy of using continue; I did the same on Display.template.

The old code had a check on empty()  which will check for empty vars and return true, I changed that part to !empty()  to check for non empty vars which is what we really need.

The "should be shown" part is implicit in the placement field, if the field matches the check, it gets printed, thats all it does.

Oldiesmann:
The problem is we're echoing out containers even if we have nothing to show there.

The original code only output the `<ul>` and `</ul>` if we actually had at least one field to show. The new code outputs those tags automatically regardless. The code for the profile fields below the signature is similar - output a div and dl for them regardless of whether there's something to show.

MissAllSunday:
mmm, OK, I see it now, this doesn't happen on the Display template because each placement gets filled on Display.php  but the code in Profile.php its pretty different and there doesn't seems to be a good way to do on profile.php what I did on display, at least not without messing with other things (I'm not particularly fond of the way the whole profile code works) meh, will revert back to the old code.

MissAllSunday:
OK, found a way to categorize the custom profile fields by placement but it requires some major changes to a few profile files :disappointed: 

Would it be OK if we create a new array on function summary() and use that array to print them? we won't be messing with loadCustomFields() and the printing code would be much more user friendly

MissAllSunday:
OK, added a few changes

$context['print_custom_fields'] categorizes each custom fields, now we can check for each specific placement and can safely remove the continue; and $shown bits. Its still possible to shown an empty ul if all fields within a placement are empty but thats highly unlikely and if it does happen, it would mean something else is borked anyway.
- Added $context['cust_profile_fields_placement'] early in the process to have a unique var that holds are possible placements, this makes it easier for anybody who would want to add more placements, or course there are still a few edits to make but having a unique var does help a lot.
- Moved integrate_pre_load call to the bottom of reloadSettings() to give mod authors a chance to play with all the context vars that are created on that function. This also means integrate_stable_icons is redundant now since pre_load can handle stable icons too.

Oldiesmann:
Didn't see your question until now, but feel free to do that if you want. Whatever makes things simpler in the long run :)

MissAllSunday:
I already did :tongue:

Oldiesmann:
Is this ready to go now?

MissAllSunday:
yep!

PR #2622, opened 2014-12-25
Prevent username shadowing due to case-sensitive name checks in PostgreSQL

sexwarrior:
Signed-off-by: Milosz Gaczkowski milosz@omgomg.eu

Currently, PostgreSQL will allow case-insensitive logins but case-sensitive registrations. A critical consequence of that is that a newly-registered _UsErNaMe_ may easily shadow _username_, rendering one of them unable to log in. Particularly problematic if all of the forum's admins are targetted.

Overall, it's worth keeping in mind that PostgreSQL's LIKE is case-sensitive, unlike MySQL's LIKE. Since we claim to support PostgreSQL, we should be conscious of this and ensure that the case-insensitive ILIKE is used appropriately. This issue likely affects more than just registrations, but this commit at least deals with the giant security hole.


live627:
Commitf must be signed off as per README.md

sexwarrior:
Commit message and pull request amended.

PR #2624, opened 2014-12-25
Alphabetise the BBC list

live627:


Sources/Subs.php
```diff
@@ -1664,6 +1664,7 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 		{
 			if (isset($temp_bbc))
 				$bbc_codes = $temp_bbc;
```
live627, line 1666: @Oldiesmann can you do a SVN blame on this and let me know why it's there? I think I've already figured it out. Ma ybe broken, too...

Oldiesmann, line 1666: $bbc_codes is a static variable. It's done as a way to cache the bbcode tags so we don't have to define the big array each time the function is called in a particular situation. This code is there to prevent the tags from being cached if we're only parsing specific tags (as opposted to all of them).

Issue #2699, opened 2015-01-28
Sources / tasks does not show up in detailed version check


illori:
do we want these files to show up? i think it is a good idea in case they get patches in future versions.

Issue #2698, opened 2015-01-28
files missing header


illori:
Class-TOTP.php, Subs-Password.php does not have the normal header all the other files do.

at least we should have a version string preset

PR #2702, opened 2015-01-29
Stuff2

### LINKED ISSUE #2700: drafts.english.php version string
the version string in this file reads // Version: 2.1 Beta 1; Profile it is not the Profile language string page, as a result it does not show a version in the detailed version check

```
// Version: 2.1 Beta 1; Drafts
```

that seems to make it show up in the list.

MissAllSunday:
Fixes #2700,  #2699 and #2698

Creates a new var for the tasks folder.

Is it pertinent to add an index.php on the tasks folder?


Oldiesmann:
We should probably add an index.php file to that folder as well.

MissAllSunday:
OK, done.

XinYenFon:
Hold this PR till @Dragooon comments here, we have to know if he made any contribution to the;
- Class-TOTP.php
- Subs-Password.php

otherwise SM/SMF cannot claim any copyright onto those files (that won't be ethic).

MissAllSunday:
We aren't claiming anything.

I left the original copy and license header intact, I added the SMF headers, thats all I did, that doesn't mean we are claiming that code.

BTW we have done this before with Class-Graphics.php  https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Class-Graphics.php#L4  since the 2.0 days with absolutely no issues.

XinYenFon:
http://www.businessdictionary.com/definition/copyright.html

You are simply putting SMF's copyright onto that file. Which is why I asked is there any change(s) done by people worked on it.

MissAllSunday:
Please take a look at the actual headers: https://github.com/MissAllSunday/SMF2.1/commit/72d0274eecba2c71afcf72c7b23fb58e68a55f38

https://github.com/MissAllSunday/SMF2.1/blob/72d0274eecba2c71afcf72c7b23fb58e68a55f38/Sources/Class-TOTP.php

https://github.com/MissAllSunday/SMF2.1/blob/72d0274eecba2c71afcf72c7b23fb58e68a55f38/Sources/Subs-Password.php

Both classes were released with the MIT license which makes this change absolutely legal. What the MIT license require is to keep a copyright notice which is what we are doing already.

The SMF headers were put after the original ones, the original headers weren't changed and still reference the code itself.

MissAllSunday:
I want to expand this as it is important.

Every contributor that has ever submitted code to SMF still retains all their copyrights. The same happens here, the class authors still retain their full copyright. SM/SMF cannot take away your copyright on the code you wrote. That will never gonna happen.

The SMF headers makes reference to SMF the package, not the individual classes. 

On both files it is clear which copyright notice belongs to what, there is no ambiguous or otherwise shady behavior. We are in perfectly legal stand here.

XinYenFon:
true, but I actually asked for not including the following line, afaik headers needed for version checking etc... as for second line yes you are also true (actually both two posts you made are nothing but true), but as for respect and work ethics (unwritten rules), I asked you to not include the copyright line so we just don't claim any copyright on files if we didn't do any changes on them. :)

`* @copyright 2015 Simple Machines and individual contributors`

MissAllSunday:
And again... there is respect and there is ethics...

Again we aren't claiming any copyright on any file, those files are included on SMF the package, its the package as a whole the one SM is claiming copyright on, not these single isolated classes. Please do try to understand this.

I'm sorry but I won't be repeating myself over this. The original authors still retain full copyright. We haven't claiming anything, SM CAN'T claim anything.  There is no lack of ethics and/or respect.

If I were to be mean and disrespectful I would have removed all copyrights and license headers from both files but that wasn't the case... so please stop this. There is no point in keep arguing about this.

live627:
@Antes That is a non-issue

@MissAllSunday Can I merge this?

XinYenFon:
@live627 yea sure, if we decide otherwise in the future we can change it again.

Oldiesmann:
One note about the index.php added in there... `dirname(dirname(__FILE__))` will always return the path to the Sources directory (at least by default), so the file will always just exit rather than redirect to the board URL. Might be better to just redirect to `dirname(dirname(__FILE__)) . '/index.php'`, which in turn will redirect to the board url.

MissAllSunday:
OK, I copied the index.php from the default folder and added support for $tasksdir on upgrade

MissAllSunday:
One more change, added the $context['allowed_html_tags'] early on the process to keep track of the allowed HTML tags on a single place, it also helps mod authors to skip yet another file edit.

PR #2708, opened 2015-02-02
Dev

### LINKED ISSUE #2367: Add Packages via Server General gzinflate(): data error
I get general gzinflate(): data error when accessing action=admin;area=packages;get;sa=browse;server=1;relative=feature_enhancement

Details
Type of error: General
2: gzinflate(): data error
Subs-Package.php
Line: 311

Cannot reproduce for others like administrative or theme_enhancements. But the above can be reproduced.

### LINKED ISSUE #2593: Silverlight XAP files in mod packages break zip files in the package manager
I have a mod that bundles a Silverlight component as a fallback for missing browser functionality; this is a dependency in my package that I am not able or willing to change.

Now, 2.1 had a rewrite of read_zip_data. Previously it was done by stepping through the file but there were issues with this for some reason, and it was rewritten in 2.1 to explode() the data on the PK file header magic bytes.

Ordinarily this wouldn't be a huge problem, but a Silverlight XAP bundle is really a renamed zip file, so this gets caught up in the explode(), meaning that the zip files contains x files but x+2 (or more) files will be reported as there are additional file entries returned, and the comparison of $file_sections to $zip_info['files'] around line 265 of Subs-Package will then fail accordingly.

Now, I can package my bundle as a tar.gz which would work - but as far as I'm concerned this is a regression from 2.0.

Note that any mod that contains zip files in it for any reason, while unlikely, will have the same problem - this includes actual .zip files, Open Office XML files, OpenOffice/LibreOffice files and of course, Silverlight files.

live627:
Notes:
- This was taken from the Wedge SVN
- As noted in the file, @Arantor has authored it, who has given us permission to use it in 
  #2593 (not  t hat it needs granted, since it _is_ BSD, after all, but courtesy, and all of those things)
- Nao has not touched its working code that I could see 
- All I did was change the class name and used our uni form file header
- Note: Should be backwards compatible. **Needs A LOT of testing.**

Fix #2367
Fix #2593

### DIFF
--- a/Sources/Class-ZipExtract.php
+++ b/Sources/Class-ZipExtract.php
@@ -1,0 +1,463 @@
+<?php
+/**
+ * Zip extraction class.
+ *
+ * Simple Machines Forum (SMF)
+ *
+ * @package SMF
+ * @author Simple Machines http://www.simplemachines.org
+ * @copyright 2015 Simple Machines and individual contributors
+ * @license http://www.simplemachines.org/about/smf/license.php BSD
+ *
+ * @version 2.1 Beta 1
+ * @author see contributors.txt
+ */
+
+/**
+ * Note: this file was authored entirely by Peter Spicer,
+ * who chose to make it available under the 3-clause BSD license.
+ */
+
+class ZipExtract
+{
+	public $zipname = '';
+	public $zip_fd = 0; // The file descriptor for file operations on the current zip file
+	private $_content = false;
+
+	public function __construct($zipname)
+	{
+		if (!is_callable('gzopen'))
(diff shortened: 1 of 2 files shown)


Arantor:
:+1: Looks good at a glance.

live627:
Cool! Now just waiting on input from someone else.

Arantor:
And indeed testing since I didn't, I just glanced at it to make sure it was done in what looked the right way ;)

live627:
Fuck waiting any longer, I'm merging this and assume that SOMEONE other than me has tested this.

I want feedback! "I'm busy with work, will test this over the weekend"" would be better than nothing.

PR #2711, opened 2015-02-03
Release 2.1s

live627:


XinYenFon:
this fixes some problems, but for example LE not outputting td element and its content now @live627

PR #2713, opened 2015-02-04
Board ordering issue (should fix #2710)

### LINKED ISSUE #2710: Issue with child-boards display when having posts or not
As reported here
http://www.simplemachines.org/community/index.php?topic=533020.0

This can easily be reproduced:
_Parent board
*_ Child
**\* Grand-child 1
**\* Grand-child 2
(everything with no posts)
You create a topic in Grand-child 1, then go back to "Parent". Grand-child 1 disappeared as child of "Child"
You create a topic in "Child", everything is back in place :)

### LINKED ISSUE #2712: board #1 missing?
on the topics maintenance page board # 1 is missing from the specific boards expanded area for remove old posts, it is also missing from the move topics drop downs at the bottom of that page.

@Oldiesmann was looking at the top part last week.

live627:
This is a c&p from getBoardTree() https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs-Boards.php#L1318

Apparently, this also fixes #2712.

### DIFF
--- a/Sources/Subs-BoardIndex.php
+++ b/Sources/Subs-BoardIndex.php
@@ -63,7 +63,8 @@ function getBoardIndex($boardIndexOptions)
 			LEFT JOIN {db_prefix}log_boards AS lb ON (lb.id_board = b.id_board AND lb.id_member = {int:current_member})') . '
 		WHERE {query_see_board}' . (empty($boardIndexOptions['countChildPosts']) ? (empty($boardIndexOptions['base_level']) ? '' : '
 			AND b.child_level >= {int:child_level}') : '
-			AND b.child_level BETWEEN ' . $boardIndexOptions['base_level'] . ' AND ' . ($boardIndexOptions['base_level'] + 1)),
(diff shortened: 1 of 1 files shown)


margarett:
I don't think it fixes #2712 as there is an own query in function MaintainTopics()
I was pretty much asleep when I tested this yesterday but I think that board #1 is lost directly in the first while loop after the query so it didn't even reach Subs-Boards. Although I'm not totally sure anymore :tongue:

margarett:
Actually, live, did you test this?
The thing is... #2710 happens in when in a board with child boards. At that time, $boardIndexOptions['include_categories'] is false, which means that smf_categories isn't joined and, as such, your fix causes a "Unknown column 'c.cat_order' in 'order clause'"
So you should drop the c.cat_order from the ORDER BY line :wink:

Emperornero:
I was going to post this earlier, I cannot confirm that this fixes either problem that are referenced in #2712 or #2710.

live627:
ok thanks for letting me know.

Issue #2740, opened 2015-02-17
img tag to accept title as optional


illori:
currently we do not have a title as part of the img tag, why not add it? it could be useful for some. maybe more useful then the alt attribute we already support.

margarett:
I made it here and it seems to work :) I'll submit a PR for this

live627:
a22ee5b9e1c683b7cf1ad29e0321d997029cf1fc

PR #2750, opened 2015-02-19
Adds optional "title" argument for the img tag as suggested in #2740

margarett:


margarett:
These were supposed to be 3 different PRs because they are different stuff... Yay me :tongue:

Oldiesmann:
That's why branching is useful - create a separate branch for each thing you're working on :tongue:

live627:
But I'll merge this anyway.

PR #2762, opened 2015-03-05
Selected text to quote

MissAllSunday:
As per this www.simplemachines.org/community/index.php?topic=533350.0

This PR offers pretty basic support for handling selected text to quote.

It uses a separate quote button, this button will appear on mouseup event if the user selected some text on that .inner div

It was build to be as unobtrusive as possible to not mess up with the existing quote system so everything is contained in a single mouseup event.

The quickbuttons sf-js-enabled sf-arrows li does not really like long text so the text I chose for this: "Quote selected text" looks weird.

I deliberately did not include the check on XMLHttpRequest

Needs tons of testing of course.

### DIFF
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -1276,6 +1276,9 @@ function Display()
 	// topic.js
 	loadJavascriptFile('topic.js', array('default_theme' => true, 'defer' => false), 'smf_topic');
 
+	// quotedText.js
+	loadJavascriptFile('quotedText.js', array('default_theme' => true, 'defer' => true), 'smf_quotedText');
+
 	// Mentions
 	if (!empty($modSettings['enable_mentions']) && allowedTo('mention'))
 	{
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -961,7 +961,7 @@ function Display()
 	$context['topic_notification'] = !empty($user_info['id']) ? $topicinfo['notify_prefs'] : array();
 	// 0 => unwatched, 1 => normal, 2 => receive alerts, 3 => receive emails
 	$context['topic_notification_mode'] = !$user_info['is_guest'] ? ($context['topic_unwatched'] ? 0 : ($topicinfo['notify_prefs']['pref'] & 0x02 ? 3 : ($topicinfo['notify_prefs']['pref'] & 0x01 ? 2 : 1))) : 0;
-	
+
 	$attachments = array();
 
 	// If there _are_ messages here... (probably an error otherwise :!)
--- a/Sources/Post.php
+++ b/Sources/Post.php
@@ -1172,6 +1172,9 @@ function Post($post_errors = array())
 		loadJavascriptFile('mentions.js', array('default_theme' => true, 'defer' => true), 'smf_mention');
 	}
 
+	// quotedText.js
+	loadJavascriptFile('quotedText.js', array('default_theme' => true, 'defer' => true), 'smf_quotedText');
+
 	// Finally, load the template.
 	if (WIRELESS && WIRELESS_PROTOCOL != 'wap')
 		$context['sub_template'] = WIRELESS_PROTOCOL . '_post';
--- a/Themes/default/Display.template.php
+++ b/Themes/default/Display.template.php
@@ -568,7 +568,7 @@ function template_single_post($message)
 									', $txt['post_awaiting_approval'], '
 								</div>';
 	echo '
-								<div class="inner" id="msg_', $message['id'], '"', $ignoring ? ' style="display:none;"' : '', '>', $message['body'], '</div>
+								<div class="inner" data-msgid="', $post['id'], '" id="msg_', $message['id'], '"', $ignoring ? ' style="display:none;"' : '', '>', $message['body'], '</div>
 							</div>';
 
 	// Assuming there are attachments...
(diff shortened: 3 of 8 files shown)


XinYenFon:
It looks nice on basics :) What i find
- .quickbuttons li a.quote_selected_button missing from CSS (just another line for .quickbuttons li a.quote_button)
- clicking outside of "inner div" does not dismiss the selected quote option and it quotes the selected text.

MissAllSunday:
The quote_selected_button is just for reference it  doesn't add any style, the other thing is on purpose to avoid crashing the client with multiple calls to onmouseup event.

margarett:
Forgive my GH ignorance :P but to test this I can just download your "quote" branch and setup 2.1 with it?

XinYenFon:
Because its under quickbuttons, it adds icon.
screenshot-localhost 2015-03-06 21-00-51

MissAllSunday:
Since this is a small Pr you can just directly copy the added changes into your local branch, just be sure to delete them if you are planing on keep on using that branch.

antes, yes I know its there I deliberately put it there, yes thats the weird thing I was talking about I thought that the li or the ul class will be capable of handling the icon without me having to use custom css.

XinYenFon:
margarett, You can get patch (add .patch to url) and open Git Bash (alike) navigate to your repo (create different branch > checkout) then use "git apply 2762.patch"

suki, nope, not in quickbuttons :)

live627:
This is easy to abuse. Try to quote the same post twice...

MissAllSunday:
As in clicking the button twice?  I don't know what you mean, more details are needed.

XinYenFon:
@MissAllSunday when i click inside "inner div" now it dismisses the Quote Selected button but doesn't do it for outside of "inner div".

Yeap, try to click twice (or more) based on your click rate, it quotes more than twice.

MissAllSunday:
I already said thats totally intentional. The whole event is attached to the .inner class, anything that happens outside that class isn't taken into consideration.

I still don't understand what you mean by click rate, the on click event doesn't use on. it uses one() which indicates that there will only be one click. When the butotn is clieked it gets hided again, preventing it form being clicked again

Need more details, where does this happen? on which button? do you mean the normal quote button? I did not modify that button.

XinYenFon:
well, did you tried to click on quote selected button more than once?

Also, using only inner div to dismiss button is surely weird way to do it and I'm 100% sure people call it bug. But I'm not going back there.

MissAllSunday:
OK, add an onmuseup even to the whole document to see what happens... If you are sure this is a bug then propose a fix for it.

When I tried to do more than one click on the "Quote selected text" it gets hidden after the first click...  DO NOTE that I said the "Quote selected text" and NOT the normal "Quote" button...

Again.... for the Nth time.,... need MORE details... which button are you actually clicking?

Make sure that the event attached to the "Quote selected text"  has one() instead of on()

$(document).one('click', '#quoteSelected_' + oSelectedID + ' a', function(e){

XinYenFon:
Well, I'm not saying its a bug, you designed it that way but I'm sure others will agree with me, regular users see that as bug. But if you want to keep it that way that's ok as well.

Edit: yeap after your last commit that seems fixed :)

margarett:
OK, so I finally got to test this.
The thing with the "double click" seems fine as I can't reproduce it. So I select a text from the post area and the button appears. I try to click it very fast and only the first click has any effect.

The "dismiss" might actually be an issue, simply because it memorizes whatever you selected.
Reproduce: Have a reply like this:

reply 123

another line

and another

Then select just the word "reply". Click outside of the post area (so that the "reply" text is not selected anymore but the button is still shown -- the "Quote selected text" button!)
Now repeat this, randomly selecting text from the reply text and clicking OUT OF THE POST area (again, like this, the text is not selected anymore but the button doesn't disappear)
When you did this several times, finally click the "Quote selected text" button. It will introduce everything that you selected before. Eg:

[quote author=admin link=msg=3 date=1426036567]reply [/quote]

[quote author=admin link=msg=3 date=1426036567]another [/quote]

[quote author=admin link=msg=3 date=1426036567]another[/quote]

[quote author=admin link=msg=3 date=1426036567]line[/quote]

[quote author=admin link=msg=3 date=1426036567]another line[/quote]

[quote author=admin link=msg=3 date=1426036567]and [/quote]

[quote author=admin link=msg=3 date=1426036567]123[/quote]

This actually works like a nice multi-quote :tongue: although I'm not sure if this is by design or not...

BTW: can we have this in the full reply page? :smiley:

MissAllSunday:
It wasn't designed like that, at least not intentionally, its just the nature of the code the one that allows this kind of behavior.

This can be further abused by selecting something completely out of the scope of the .inner div, for example the entire main menu so I need to find a way to only select the text of the .inner div we are working on.

MissAllSunday:
OK, after checking this further, the whole approach is wrong, the on click event shouldn't be inside the onmouseup one...

Gotta re-do this all over again.

MissAllSunday:
OK, made a few improvements:
- Added it to the post page
- I forgot JS is way more picky than PHP so !var is not the same as var == false. In this case we should use == and not ===
- Moved the code to its own file.
- Added a check to make sure the selected text really belongs to the div that fired the onmouseup event
- Added the class to the css file.

The weird "multiquote" stuff still happens, I actually kinda like it but I do see some people getting all confuse about it.

The button still doesn't hide intil you click the div again, dunno how to solve this really, perhaps a setTimeout()?

XinYenFon:
Notice: Undefined variable: post in C:\wamp\www\SMF2.1\Themes\default\Display.template.php on line 571
" id="msg_14">

mikekaganski:
Triple click (to select the whole line) on the *last* line of the message disables the "Quote selected text" button.

For test: open https://forumooo.ru/index.php?topic=9244.msg63373;topicseen#msg63373 ; try to triple-click at the message's last line "В notepad++ использую тот плагин, но в Индизайне тоже проверил, с тем же результатом.".

Unlike the case when you drag mouse to select, the triple click seems to get something past the correct area, and the button gets disabled.

Triple-clicking other lines works fine.

The site uses SMF 2.1.2.

Issue #2766, opened 2015-03-08
incorrect text in edit membergroup


illori:
You can upload custom images to the default theme directory to be able to select them here.

this text is not correct, the image must be in the images/membericons folder for each theme that is being used, not just the default theme.

margarett:
I never really used this. How does this work, exactly? Can the same group have different stars in different themes?

margarett:
Strange behavior...
The only icons that show up for selection are the one in default theme's "membericons" folder (even if using other theme and other icons exist in that theme). But if said icons don't exist in other theme, it doesn't fallback to default's ones. IMO it should, don't you think?

illori:
cant hurt. 

i dont agree with your change here https://github.com/margarett/SMF2.1/commit/3b8d47028aa1a0b6d108a8f2d00dae9d8080f963#diff-7fa5ecbd8c738e177ea868d01f023d59L82

sure it can fall back to default, but we should still be prompting the user to put the images in their themes folder.

that is why the text on the left is wrong but the path was correct.

margarett:
I tested with more than one of the themes available for 2.1B1 and I always reproduced the behavior.
**No matter which images are in these theme's "membericons" folder, only the default theme's icons are available for selection.**
 If the same icon exists in the custom theme, that is used. If not, no image is loaded (the replacement "*" is shown)

So IMO, what I changed makes sense, because it is as SMF behaves. Unless I missed something, which is highly likely :tongue: 

edit: damn markup :angry:

illori:
well if that is how it works in 2.1. it is a MAJOR regression from 2.0.

if you select a badge/image and put a different one in your theme does the different one show up in the posts?

margarett:
If it has the same name, yes.
Until my PR, if it wasn't there (same name) nothing would be shown (just the "asterisk"). Now it falls back to the default theme (which was originally selected in admin panel)

illori:
that is fine, but you are telling people not to upload the image to their theme. this will make users think that they can no longer have different images for each theme.

margarett:
Can this issue be closed? The PR was merged so... ;)

illori:
i dont think so. i still dont agree with the change to default. you still can have the images in your custom themes and right now it makes it look like you can not from this point forward.

margarett:
But the fact is that you can only select icons in default theme's folder :P
Maybe it requires a help popop or something?

illori:
[...]

PR #2785, opened 2015-03-20
Fixes and improves #2766

margarett:
Fixes the text about where the membergroup icons should be placed and adds a fallback in Load.php so that, if the icon doesn't exist in the current theme's "membericons" folder, falls back to default


Themes/default/languages/ManageMembers.english.php
```diff
@@ -79,7 +79,7 @@
 $txt['membergroups_settings'] = 'Membergroup Settings';
 $txt['groups_manage_membergroups'] = 'Groups allowed to change membergroups';
 $txt['membergroups_select_permission_type'] = 'Select permission profile';
-$txt['membergroups_images_url'] = 'Themes/{theme}/images/membericons/';
+$txt['membergroups_images_url'] = 'Themes/default/images/membericons/';
```
live627, line 82: Might want to revert this change.

margarett:
@live627 check the discussion here
https://github.com/SimpleMachines/SMF2.1/issues/2766

I don't know if this was the intended behavior or not (although it makes MUCH sense to me as it is) but the suggested text is adequate to how it works ;)

PR #2790, opened 2015-03-24
Make spamProtection() more generic

MissAllSunday:
Added a second param. Sometimes you just don't want the function to horribly die with a fatal lang error

Wanted to add a hook there but it was already added :tongue:

### DIFF
--- a/Sources/Security.php
+++ b/Sources/Security.php
@@ -1145,13 +1145,15 @@ function spamProtection($error_type)
 		'reporttm' => $modSettings['spamWaitTime'] * 4,
 		'search' => !empty($modSettings['search_floodcontrol_time']) ? $modSettings['search_floodcontrol_time'] : 1,
 	);
-	call_integration_hook('integrate_spam_protection', array(&$timeOverrides));
+
 
 	// Moderators are free...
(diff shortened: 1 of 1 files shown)


Sources/Security.php
```diff
@@ -1176,7 +1179,9 @@ function spamProtection($error_type)
 	if ($smcFunc['db_affected_rows']() != 1)
 	{
 		// Spammer!  You only have to wait a *few* seconds!
-		fatal_lang_error($error_type . '_WaitTime_broken', false, array($timeLimit));
+		if (!$only_return_result)
+			fatal_lang_error($error_type . '_WaitTime_broken', false, array($timeLimit));
+
```
live627, line 1184: I think this trap should return false if $only_return_result  is true.

edit nvm I'm dumb.....

Sources/Security.php
```diff
@@ -1153,6 +1154,8 @@ function spamProtection($error_type)
 	else
 		$timeLimit = 2;
 
+	call_integration_hook('integrate_spam_protection', array(&$timeOverrides, &$timeLimit));
+
```
live627, line 1158: ???

PR #2796, opened 2015-03-26
Use PharData to handle ZIP files

### LINKED ISSUE #2730: install package from package repo
i dont have a real server to test on, just my localhost.

if you use the SM.org package server and try to install any package that has a download link i get the error message related to missing temp folder or the archive is empty. anyone else have this issue?

i tested on my 2.0 install on the same localhost install and had no issues i was offered to install the mod.

### LINKED ISSUE #2737: Package Manager --> White page
Subject says it all. Trying to open PackMan returns a white page (after some seconds "Waiting for localhost), nothing in SMF's error log, nothing in PHP's log

live627:
This library won't choke on GitHub's generated files. And it's faster than ZipExtract.

Fix #2730 
Fix #2737


margarett:
This is a great idea :)
Unfortunately there's something wrong... I uploaded again my Custom Actions zip to Packages folder and tried to install it. PackMan doesn't choke (great!) but there is a "Modification parse error" in all files.
If I put the folder (uncompressed) inside Packages, the error is gone
If I repack the MOD (unzip/zip), the error is also gone.

So there is still something weird with GH's zip files...

live627:
>  but there is a "Modification parse error" in all files

Fixed.

PR #2799, opened 2015-03-28
Use PharData to create package backups

### LINKED ISSUE #2141: Package backup is broken
The routine that creates a backup of the installation prior to installing a mod is broken. The archive files produced are damaged where 7zip reports an error due to content being after the end of file marker. Simply the file becomes too big for the original routine to handle and it seems to mash the tar headers (since the gzip part is fine, it's the inner tar that's broken)

This applies to 2.0 and 2.1 as far as I can tell.

live627:
Well, files seem complete now.

Also, log those errors!

Fix #2141

### DIFF
--- a/Sources/Packages.php
+++ b/Sources/Packages.php
@@ -2519,7 +2519,7 @@ function PackagePermissionsAction()
 		if (empty($context['total_items']))
 		{
 			/**
-			 * Counts all the directoies under a given path
+			 * Counts all the directories under a given path
 			 *
 			 * @param type $dir
 			 * @return integer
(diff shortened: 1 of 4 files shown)


Sources/Packages.php
```diff
@@ -2520,7 +2519,7 @@ function PackagePermissionsAction()
 		if (empty($context['total_items']))
 		{
 			/**
-			 * Counts all the directorys under a given path
+			 * Counts all the directoies under a given path
```
XinYenFon: directories :P

PR #2803, opened 2015-03-30
Added some help texts missing in Manage Server area

### LINKED ISSUE #2792: Two-Factor Authentication enabled?
i have not enabled this feature but on the cookies and sessions page the drop down shows "enabled". it should not show enabled if i have not selected it to be enabled.

why have a drop down when the only options are enable/disable? a checkbox could do that easily.

i did try to log in and i was not redirected to the config page for 2fa.

does the admin really need to setup their own account for 2fa before others can?

the text says "Please enable 2FA in your account in order to be able to force 2FA on other users!"

margarett:
Fixes #2792 
Maybe some texts require a developer review, though...


ChainyEO:
I think it might be good to make a couple of changes to the language:

> $helptxt['tfa_mode'] = 'You can add a second level of security to your forum by enabling <a href="http://en.wikipedia.org/wiki/Two_factor_authentication">Two Factor Authentication</a>. 2FA forces your users to ~~add a~~ enter a machine-generated code after the regular login. ~~You need to configure 2FA to yourself~~ You need to enable 2FA for your own account before ~~you are able to force~~ forcing it to other users!

live627:
> forcing it to other users

Wouldn't it read better as 

> forcing it tor other users

ChainyEO:
I think it's understandable either way, but for perfect clarity you could say: 

> ...before forcing it to be a requirement for other users.

PR #2814, opened 2015-04-05
Don't place an arbitrary limitation on recent posts in SSI based on message ID

sexwarrior:
This commit removes the min_message_id limitations from ssi_recentTopics() and ssi_recentBoards(). The current limitation has been found to break if one of these functions is used only on a small subset of boards in the forum - they don't get posted in often enough, and so the min_message_id limit becomes more restricting than num_recent, ultimately leading to a smaller array than requested being returned. A (MUCH) more lenient limit may need to be considered in the future if efficiency becomes an issue.

Signed off by: Miłosz Gaczkowski milosz@omgomg.eu

### DIFF
--- a/SSI.php
+++ b/SSI.php
@@ -277,7 +277,7 @@ function ssi_recentPosts($num_recent = 8, $exclude_boards = null, $include_board
 
 	// Let's restrict the query boys (and girls)
 	$query_where = '
-		m.id_msg >= {int:min_message_id}
+		1=1
 		' . (empty($exclude_boards) ? '' : '
 		AND b.id_board NOT IN ({array_int:exclude_boards})') . '
 		' . ($include_boards === null ? '' : '
@@ -289,7 +289,6 @@ function ssi_recentPosts($num_recent = 8, $exclude_boards = null, $include_board
 		'is_approved' => 1,
 		'include_boards' => $include_boards === null ? '' : $include_boards,
 		'exclude_boards' => empty($exclude_boards) ? '' : $exclude_boards,
-		'min_message_id' => $modSettings['maxMsgID'] - 25 * min($num_recent, 5),
 	);
 
 	// Past to this simpleton of a function...
@@ -461,7 +460,7 @@ function ssi_recentTopics($num_recent = 8, $exclude_boards = null, $include_boar
 		FROM {db_prefix}topics AS t
 			INNER JOIN {db_prefix}messages AS ml ON (ml.id_msg = t.id_last_msg)
 			LEFT JOIN {db_prefix}boards AS b ON (b.id_board = t.id_board)
-		WHERE t.id_last_msg >= {int:min_message_id}' . (empty($exclude_boards) ? '' : '
+		WHERE 1=1' . (empty($exclude_boards) ? '' : '
 			AND b.id_board NOT IN ({array_int:exclude_boards})') . '' . (empty($include_boards) ? '' : '
 			AND b.id_board IN ({array_int:include_boards})') . '
 			AND {query_wanna_see_board}' . ($modSettings['postmod_active'] ? '
@@ -472,7 +471,6 @@ function ssi_recentTopics($num_recent = 8, $exclude_boards = null, $include_boar
 		array(
 			'include_boards' => empty($include_boards) ? '' : $include_boards,
 			'exclude_boards' => empty($exclude_boards) ? '' : $exclude_boards,
-			'min_message_id' => $modSettings['maxMsgID'] - 35 * min($num_recent, 5),
 			'is_approved' => 1,
 		)
 	);


Kindred-999:
As I understand it, the so called "arbitrary" limitation is to prevent server overload and removing it could, potentially kill your server performance with that SSI request. So efficiency _IS_ already considered in the original coding. I disagree with the request to remove the limit and believe this pull should be rejected and/or refactored.

Additionally, the coding, "where 1=1"   really?

sexwarrior:
> prevent server overload and removing it could, potentially kill your server performance with that SSI request

ssi_recentTopics() and ssi_recentBoards() are lighter on the database than a lot of SMF's core functionality. In other words, if efficiency was a factor here, we'd already have a much bigger issue elsewhere.

That said, as explained in the original comment, a significantly more lenient and better-planned limit may be considered. A non-functional method is not efficient by any stretch of the imagination - who cares if it does something fast if it doesn't actually do what's advertised?

> Additionally, the coding, "where 1=1" really?

That exact same idiom is already present in SSI.php, and throughout SMF as a whole. Keeping things consistent is not a bad thing. The reason for that should be fairly clear, given the large amount of ternary operations in the queries, which is common in this project.

live627:
ssi_recentTopics() not very light at all.. II can get slow on big boards.... even after I split its  query....

sexwarrior:
I'm not saying it's very light. I'm saying there already are heavier things which aren't being restricted.

Would you more willing to accept this if disabling the feature-breaking restrictions was an optional flag rather than a hardcoded change?

Kindred-999:
not really...   we are actually REMOVING optional settings in places as we move forward... most especially options which could break things if improperly used.  Adding a new system hanging option is not on the roadmap.
PERHAPS increasing the limit a little... but even than is problematic.
This really sounds more like mod material -- especially since I think this is a fairly niche issue since I have never actually encountered the issue on 15 different sites that I have run over the years... Nor is it a frequently reported/complained about issue.

sexwarrior:
I would like to see some evidence of the "system hanging" claim. SMF frequently performs heavier operations than this, so if you were right, you'd see a hanged system every time someone performs a search. The reason that doesn't happen is because mature DBMSes are pretty good at what they do. We already have a perfectly good restriction in the SQL query - the LIMIT clause. It's the DBMS's job to optimise the query, and, to no one's surprise, they usually do the trick.

The fact that you failed to spot a broken feature is nothing extraordinary. It happens to the best of us, really. No one is able to evaluate all use cases, which is where the strengths of a collaborative system comes into play. If you find find it unlikely that someone would make 125 posts in a different board than the one you're currently trying to display, I don't know what to say to you. 1250 _might_ be a safe-ish magic number, but even then it's an arbitrary restriction with no good reason stated for it.

I particularly do not enjoy the appeal to authority you're trying to instil here. If you have a case, make it, and back it up with evidence. "I run 15 sites and never ran into the problem" is a worthless statement. Would you also dismiss RTL theme fixes because you never ran into problems on your LTR sites?

Kindred-999:
You have missed my point...   Your view of this as a "broken feature" is particular to YOU.
In running multiple sites over the last decade, I have never encountered an issue in which the use of the SSI lists caused the problem that you seem to think is a major issue. MY statement above is the shorthand of -- "I have used the SSI functions extensively, across multiple sites and multiple integrations. The limit that you call an issue has never arisen in any of the sites that I have run, despite being frequently used. The reason that YOU see it as an issue is particular to your own setup with hidden/inaccessible boards, etc."

It is not an "appeal to authority" - it is a statement from a position of knowledge and authority backed by experience. At the time, I did not feel like typing it all out for you, however... and _I_ "do not enjoy" your assumptions otherwise nor your silly comparison.

Remember, we have to code the software to run small, large and even enterprise level forums. So, while you might not see the query as "expensive" for your own uses, it does get very much so on larger systems, and the fact that "there are heavier queries" does not mitigate this factor.

We appreciate the effort at contribution. Please continue to submit contributions like this.
However, at this time, I do not see this pull request as likely to be accepted as submitted, for the reasons documented above.

While a slight increase in the limit might be accepted, the total removal is not really an option.

sexwarrior:
I honestly don't understand why you're so dismissive of my claim about there being more DB-intensive functions in place in SMF. This is the best way to know that this change wouldn't be problematic, or that it wouldn't be more problematic than the rest of SMF. If a PlusSearch2() doesn't break your forum, then neither will this. And if PlusSearch2() does break your forum, you really have bigger issues than SSI integration.

It sounds like you chose to never use ssi_recentTopics() or ssi_recentPosts() to only show a small subset of boards within your forums. Like I said, I don't fault you for that (contrary to you faulting me for using a function well within its documented purpose), but it's a completely reasonable use case. You also seem to think that choosing not to use the function in such a way on multiple occasions somehow makes you an authority. The RTL/LTR comparison is spot on - "I chose not to use x, therefore x is unimportant" is the stance you've taken in many of your comments.

This isn't a matter of me crying about my own workflow (We applied the change in our own forum, as we have done with many other things which we feel SMF doesn't quite hit the spot with for us. Most of them are genuinely a matter of preference, and I don't make pull requests for those). This is a matter of saving other users a pain of trying to debug the function only to find out that it's because of a poorly chosen magic number. Or, worse still, only to find out that they don't know how to resolve it and give up (we can't assume all forum admins to be competent coders). This isn't about me being upset about my use case. It's about me hoping to save someone else some frustration while setting SSI up.

But it's clear we won't get anywhere with this approach. We disagree strongly, and I doubt either of us is gonna convince the other. Let's try to compromise and find a limit that's acceptable from your point of view and allows the function to operate correctly, at least anecdotally.

Personally, I think multiplying the magic number by a factor of 10 is the absolute minimum to ensure correct functioning of ssi_recentTopics(). Even that is gonna be a stretch. I imagine you'll want to go quite much lower than that.

MissAllSunday:
Just add another param to the function, optional of course, for those who want to mess around with that value.   If we do not want to break compatibility with 2.0.x series then something hackish as using a global $context value will do the trick too.

Oldiesmann:
I do see where the current code could cause issues if you have a board that isn't posted in very often, but removing that altogether can cause problems on larger forums. A better solution would be to either increase that limit some (maybe 500 posts instead of 175?) or move the "35" to a parameter instead.

PR #2839, opened 2015-04-17
Attachment uploads and inline attachment display

MissAllSunday:
PR this so it can be further tested. I got my head so stuffed with this stuff that I am probably missing some serious but quite easy to find errors :laughing: :
- Adds minimal inline attachment support, can be expanded pretty easily but it all depends on what we want.
- Adds [dropzone.js](https://github.com/enyo/dropzone) to upload attachments before posting. The user can upload the attachments and then leave them as is and the post will act as 2.0.x does, however, if there are multiple attachments and you upload 1 and leave the rest, only the one uploaded previously will be taken into consideration, this is the way the script works and haven't found a way around that.
- Leaves the "regular" attachment process (almost) intact, as documented on the topic at SMF, I did have to change a few things here and there but the idea was to left the process as "untouched" as possible.
- Has UI issues, I'm not good building user interfaces, just tried to add classes to everything so it can all be handled via css, its all documented on the relevant topic at SMF.
- Needs testing! specially the delete part as it is somehow fragile and perhaps should be re-written to properly handle any errors. There is also the things I raised on SMF and how are we gonna handle it.


XinYenFon:
Pushing the limits :P
- after upload it gives link with extra "]"
- BBC tag to insert attachment in the post:: - has extra ":"
- If you upload a file with eg pdf format, it tries to embed (insert) 
- you cannot upload multiple files at the same time
- if you set limits to 0 (unlimited) system stops working and counts 0 as 0 (none)
- (Notice: Undefined index: attach_restrict_attachmentNumPerPostLimit_MB in C:\wamp\www\kittenground\Sources\Post.php on line 1152) < dunno what is this? I just set everything to 10000. (Probably related to limit > error The current php.ini setting 'post_max_size' may not support this. )
- no need to put 100% on inline just leave it be :)
- #postAttachment dd not needed let it roam wild :P

MissAllSunday:
> after upload it gives link with extra "]"

Where?  which link?

> BBC tag to insert attachment in the post:: - has extra ":"

It has one, doe you mean that one?

> If you upload a file with eg pdf format, it tries to embed (insert)

If you drag and drop a file outside the drag zone then yes, the browser will automatically try and open that file, thats a standard browser behavior not related to this script.

> you cannot upload multiple files at the same time

I already addressed that on the SMF topic.

> if you set limits to 0 (unlimited) system stops working and counts 0 as 0 (none)

What limits?

> (Notice: Undefined index: attach_restrict_attachmentNumPerPostLimit_MB in C:\wamp\www\kittenground\Sources\Post.php on line 1152) < dunno what is this? I just set everything to 10000. (Probably related to limit > error The current php.ini setting 'post_max_size' may not support this. )

This isn't related to this PR.

> no need to put 100% on inline just leave it be :)

online where? 

> #postAttachment dd not needed let it roam wild :P

Yes it is still needed.

XinYenFon:
Notice: Undefined offset: 0 in C:\wamp\www\kittenground\Sources\Subs-Attachments.php on line 952

on post page.

margarett:
Was now testing this. Looks REALLY good. But soooo good that it hurts :laughing: 
Could not catch Antes's error.

The only thing that I think is missing is a button to upload everything at once. So imagine that I point 10 files at once to be uploaded, I have to click "Upload" in each of them, which is annoying...

Final suggestion: maybe reduce the size of the boxes. They are too big, especially when there's no image preview
img1

margarett:
Oh, almost forgot: kudos @MissAllSunday for this. B U T FUL ;-)

margarett:
OK, now I caught Antes's error :tongue: 
Some stuff I noticed:
- preview (with attachments inserted) didn't work and threw some errors in the log
- after that I left the post screen (by going back to BoardIndex). Then I got back to post a new topic, but now I have 2 errors in the log just by entering the post page:
  Undefined index: id
  File: C:/xampp/htdocs/smf21/Themes/default/Post.template.php
  Lines: 416 and 417

Did the browser kept something? Let me just restart my browser ;-)
Edit: yes, restarting the browser makes that go away.

live627:
Any updates @MissAllSunday?

MissAllSunday:
Nope! :cry: the undefined error is most likely a flaw in the code's logic. I've seen it I just cannot narrow it down when/where it happens.

Whats been driving me crazy is adding a "upload/Cancel all" buttons. Adding a cancel/upload all button is easy, whats difficult is getting the single "cancel" and upload buttons for each file to work again.  The script has no way to identify a single file and when .submit() is called it will upload all files and not just the single one you want to upload...

When a single upload is been handle, the script passes the entire object data which includes all uploaded files so far, calling .submit() will automatically call it for all fiels and not just the one we want to :sob:

MissAllSunday:
Progress!!!

Turns out the demo from where I started coding this uses jquery each() and an index var as a unique identifier but turns out thats not really a good way to really identify a single file.... specially if the user cancels and upload several files.

So, I've been slowly replacing that "index" identifier with a "uniqueIID" which is basically the same as index BUT BUT BUT it doesn't  get overwrittien if a new file is added to the list, small but very important difference...

MissAllSunday:
nah.... spoke too soon... the whole freaking thing is based on each() .... if I change the keys the whole script collapses... AWESOME!

MissAllSunday:
AND 7 hours later...

Fixed the undefined issue and the cancelAll button annoying issues too... still need to fix the upload all button  and also to decide what exactly the attach BBC tag is suppose to show when the attachment is not an image... currently it shows nothing, not even the name... and thus a bunch of errors appear because the message is seem as "empty" by parseBBC() 

Also... the delete attachment option is pretty much borked, it doesn't even delete the attachment...

Oldiesmann:
Why not just show the standard link to download the attachment like what we show at the end of the post if it's not an image?

live627:
@MissAllSunday I see you added some commits to this. Is it ready?

XinYenFon:
What I find so far;
- [ ] Even if you cancel all items on the list one by one. Upload All - Cancel All buttons does not disappear.
- [ ] Upload all button, uploads first attachment, rest stuck at 100% (or 99%)
- [ ] Cancel all, visually removes uploaded attachment(s) but posting reply/topic shows the uploaded attachment(s).
- [ ] Delete this attachment is not working.
- [ ] If you attach something bigger than the allowed size, system counts the file size and does not allow you to upload the files.
- [ ] Upload All button uploads some type of files even its not allowed (eg you can upload something file.suki), not sure if this is a security problem ? Maybe they can upload and run malicious code via this (still not sure if its possible).
- [ ] If I attach non-image file and use inline-option then preview the topic/reply I'll get;
  screenshot-localhost 2015-08-16 17-08-48
- [ ] and one last error not sure how I manage to get this :( sorry
  screenshot-localhost 2015-08-16 17-09-59

MissAllSunday:
Nope, it isn't ready

XinYenFon:
- [x] If I upload a file bigger than allowed size, error message showing max-size as 1kb.
- [x] Delete this attachment has confirmation message but ignoring it (clicking cancel does not cancel the "delete" activity.
- [x] Sucessfully deleted message is going below item
  screenshot-localhost 2015-09-16 00-26-35
- [ ] Error if I preview non-image files. 8: Undefined index: is_image - ?action=post2;board=1;preview;xml - Subs.php Line: 1158

MissAllSunday:
I fixed the confirmation issue.
Yes, the script sometimes shows 1kb, thats why I didn't use the default message and instead added a harcoded line made with PHP so make sure the script is overwritting the default line with the one printed on the source code.

I can't replicate the message going below the correct item.
- Added an upload/cancel all buttons
-  Added a total progress bar.

MissAllSunday:
OK I fixed the "the message going below the correct item" issue.

XinYenFon:
- [ ] I still cannot preview the inline image on post preview. But it works as expected after post.

and the issue above still.

XinYenFon:
oh and awesome work!

MissAllSunday:
What issue above? can you please be a little more specific?

XinYenFon:
I listed ? "Error if I preview non-image files. 8: Undefined index: is_image - ?action=post2;board=1;preview;xml - Subs.php Line: 1158 "

MissAllSunday:
yeah... you made it sound like there was 2 issues... the preview one and another, thats what I interpreted when you used  "and the issue above still". It makes it sound as if you are referring to another issue "above" the preview one or if the "the message going below the correct item" is still an issue.  thats why I asked you to clarify that.

margarett:
Heya Suki, great work (again)

I found an issue with going back from the post page and back in. Let me try to explain in steps
1 - Go to the post page (I tried with both new post and reply) and write something. Attach a file and upload it.
2 - Now, instead of submitting the post, just press "Go back" on your browser
3 - Enter the post page again (eg, a new post, but it's the same with a reply)
You'll get the file you uploaded before already attached to this new post and there is an error in the log

```
Type of error: Undefined
8: Undefined index: id_thumb
http://localhost/smf21/index.php?action=post;board=1.0
File: C:/xampp/htdocs/smf21/Sources/Post.php
Line: 1200
```

This behavior only goes away after you close your browser or logout

MissAllSunday:
Thanks, yes those are the pesky "session attachments" I was talking about, should be fixed now, I forgot those session files actually has their own thumbnails so with a few changes that thubmnail id should be used.

I also added support for the WYSIWYG editor

margarett:
I tested again and I still get more or less the same behavior: when I reenter the post page I still get the previous attachments. Now I get no error and I the attachment ready to be used (green background and all :smile: )
I just don't know if this is a good behavior (but this is a design decision)

MissAllSunday:
yes it is by design, "session attachments" are considered as "uploaded attachments", they have a unique attach ID and for all purposes they are valid attachments the only difference is that those attachments aren't assigned to any message.

session attachments will indeed show just as if they were uploaded on every new message or reply page that particular user tries to make until the session is over or the attachments are assigned to a message or the attachment are deleted.

2.0.x does exactly the same, it shows "already attached" files as checkboxes and the user can "un-check" those checkboxes if they do not want to assign those attachments on the message the user is currently creating.  this is the same but instead of adding checkboxes, the attachments are fully displayed and with a link to delete each file.

This same principle can be used to store "draft attachments", if someone saves a draft and that draft has some attachments, when the user loads the draft, load the attachments associated with that draft too.

margarett:
Gotcha :smiley: 

I'll test some more :wink:

XinYenFon:
Well I didn't find any issues in my tests.

margarett:
@MissAllSunday, tested this again, this time even tried cross-board permissions.
Everything seems fine and, IMO, this should be merged towards B3 ;)

Just a small note: when an inline attachment is to be shown in a place not allowed, this is presented:
"not_allowed_to_see" (no quotes)
I would suggest a better description. Other than that, all is very very, _very_ good :open_mouth:

MissAllSunday:
Yes, the function returns a text if there was some error, it cannot horribly die while parsing and we don't want to set up too much logic into it so it returns a $txt key and lets whoever called that function to handle it.

XinYenFon:
- if you upload files it calculates the total size but when you delete one (or all) it does not recalculate the size of attached files (cancel all clears it).
- If you upload max allowed quantity of files then delete one (or more) you cannot upload new one(s).
- After deleting an attachment system only changes first one's background others stay as it was in original.

new changes looking amazing :)

margarett:
Actually @Antes the first point seems to work here, it seems to recalculate the bar nicely when removing files.

@MissAllSunday, just an additional note: if you edit a post which already has attached the max number of files, the "old" list of attachments is shown.

margarett:
Another tricky one, related to what Antes said. Here's how I reproduce it:
- max 3 attachments por post.
- create a post and attach 2 files. Post it
- edit the post and select 2 more files to attach (one at a time). The second gets red and says "Sorry, you aren't allowed to post any more attachments." Good.
- Now go and remove the third attachment (the first you selected when editing the post)
  --> the remaining attachment (which should be allowed to be attached) still keeps the error state and cannot be uploaded

XinYenFon:
I'm using Firefox 41.0.2

margarett:
Me too :)

XinYenFon:
I can't use the attachment system on MS Edge actually. @margarett are you uploading the file or just adding then removing ? I'm talking about the calculation which is not made after upload then delete.

margarett:
Yes, you are correct, sorry.
Eg: Upload 2 files, delete one of them, the bar isn't updated.

About MS Edge: yeah... Who knows what we will need to support it properly :open_mouth: 
Everything JS-related will be a royal PITA.

live627:
Last I heard, Edge doesn't support the File API--thus, half the internet is
broken when using it.

MissAllSunday:
Upload works fine on edge,  it is only the drag and drop part the one that doesn't work due to edge changing how it handles it I got a fix for it but I'm waiting for a better solution from dropzone if there isn't then I'll apply it

MissAllSunday:
OK, I believe I fixed all the issues you guys mentioned earlier.

margarett:
Cool beans :) almost everything looks perfect.
The only issue I can still identify is the one I described, regarding the maximum attachments number
https://github.com/SimpleMachines/SMF2.1/pull/2839#issuecomment-152367674

The same behavior occurs if you exceed the maximum size per post. Eg:
- Upload files which almost exceed the max size per post. Post.
- Edit the post, add another file which exceeds the max size per post. You get an error and you can't upload it. Good.
- Add another (very small) file. Again you get the error about the max size.
- Cancel the third file, leaving just the already attached files and the last (very small) file. It still keeps the "max size" error and you can't upload it.

MissAllSunday:
I believe I go that issue and a few other ones as well, including previewing a non image attachment in WYSIWYG mode which was a real pain in the posterior...  I didn't want to hit the server with an ajax request to fetch an attachment details and see if its an image so I use a couple of BBC params to get that info directly form the source itself.

margarett:
Merge it, I say!!!! :smiley: 

Thank you ;-)

wintstar:
Great work! My congratulations, Thanks :smile:

MissAllSunday:
En Thu, 12 Nov 2015 02:29:01 -0600, wintstar notifications@github.com  
escribió:

> Great work! My congratulations, Thanks
> —
> Reply to this email directly or view it on GitHub.
> Thanks!

## 

Mavn:
Some issues with attachments
1, Attachments doesn't work for basic user(admin all work fine). Add attach -> upload(fine) -> save post and post is clear without attachment.

MissAllSunday:
Thats related to the permissions system, somehow regular users end up without the "post_attachment"  permission even though it is assigned to them.  Not really something related to the inline attachment system itself.

mmm further testing this, it seems it is kinda related, doing a var dump on Post.php exposes 44 or so permissions meanwhile doing the same dump on Attachments.php only shows 23...

Mavn:
Another one:
Basic user (with captcha protection) if user enter wrong captcha his all attachments will be clear
I don't know maybe it's the same Issue with permissions, but I think it's will be need testing too...

MissAllSunday:
ugh, I forgot about captcha... thank you for the report i'll take a look at it.

MissAllSunday:
About the regular user issue, yep, that was a permission issue, the allowedTo() call didn't know which board the message was suppose to be posted on so the "post_attachment" check always returned false.

This commit should fix that: https://github.com/MissAllSunday/SMF2.1/commit/5590fec577a4b0e108bc7026d46da18c4580e81a

As for the captcha issue, it should be a matter of saving the current attachments and re-display them almost exactly as handling already added attachments when editing a message.

Mavn:
Thank you, your  fix is working :) 
I find another one(using one file for this):
Add attach -upload -> after click delete, add attach -upload  -> click delete, add attach -upload -> click delete - add attache -upload - save post ... And see all attachements with one id below post :)

MissAllSunday:
I further tested the "captcha issue" and I cannot reproduce it.

If an attachment was "fully" uploaded, that is, it has an ID, then if the user fails the captcha those files are going to still be there, they are "catched" by the mock file system.

If the user just drags and dropS a file  WITHOUT uploading it and tries to post the message there should be a warning message telling the user that there are some files there and the user needs to either upload them or cancel them.

It is possible to bypass this warning of course, maybe jquery failed to attach that event to the post button or something like that, in that case, well, there isn't much we can do, those files aren't fully uploaded yet so theres no real harm on "losing them".

badmonkey72:
Fantastic work MissAllSunday!  There is one concern I've found.  Using the stock Android browser (ok, assuming "stock" on a Cyanogenmod based ROM), selecting Add Files leads to a "No apps can perform this action" error.  However, it works fine using Firefox for Android.  Do you have any ideas for a fix?  Thanks!

MissAllSunday:
I'm afraid there isn't a solution for this or at least not an easy one, if  
a web browser doesn't support the  input type file tag then there is  
nothing we can do about it, it will be problematic to offer support for  
every custom browsers out there.

Googling that error "No apps can perform this action" it seems more  
related to your Android device, perhaps the browser you are using has some  
setting to prevent file uploads or that app doesn't know how to handle or  
how to have access to your files inside your cellphone, I dunno.

badmonkey72:
File selection works fine on mine for the old style attachments, so browser capability isn't the issue.  If this breaks things for Android this may well shut out a huge number of users.  I will try it on my wife's completely stock phone to be sure.

badmonkey72:
Yes, it's broken on a completely stock Android tablet.  Don't take what I'm saying as criticism, it isn't that at all.  I'm only trying to help.  ;)  Don't think of it as a "custom" browser.  This is a large percentage of users at this point.

MissAllSunday:
I just posted an attachment using the default browser on my phone and also using opera and in both cases I was able to upload a file

Oldiesmann:
The error is definitely a problem with your setup. On my phone (LG G3), I get a popup asking me which app I want to use to upload the file (camera, documents, etc.).

badmonkey72:
Is that using the stock browser (not Chrome, not FF)?  I have 4 Android devices, one modded, 3 stock.  This is broken on all 4 when using the stock browser.  It does work using FF and Chrome.

Mavn:
1. problem with russian in file name 
   
   files save without filename, and insert files also without 
2. Insert, upload and delete bottons jumping on the sreen with various resolution

MissAllSunday:
@Mavn Can you test this fix: https://github.com/MissAllSunday/SMF2.1/commit/fc18b37c300cfa468ee05a4fa3046c91b44a9788 ? I tested it with a few accented Spanish filenames and it all seems fine, just want to make sure it works.

Mavn:
I tested, but it doesn't work

MissAllSunday:
mmm can you paste here whats the result from the call to save an attachment?  add a console.log(response);

after this:

// Server returns an array.
        response = responseText.files[0];

in smf_fileUpload.js  and add some attachments while dev tools or firebug is open

PR #2841, opened 2015-04-22
Generic fixes

### LINKED ISSUE #2821: remove HTML from board descriptions
allowing HTML in the board descriptions is actually a potential security issue and it leads to regualr issue reports from admins who screw up their site layout (and is unnecessary for the most part)

It would be my inclination to state that we should allow just basic text... but it has been pointed out that there is some valid use for basic formatting and links. So, I recommend basic BBC (bold, italic,underline, hr and links) and ignore or remove everything else..
(allowing images leads the the highest number of reports, so I'd suggest getting rid of those - or forcing a CSS on images in that section to be minimal if it is deemed necessary to keep images)

### LINKED ISSUE #2823: remove/disallow HTML in membergroup descriptions
we should be removing/cleaning HTML form the membergroup descriptions.

### LINKED ISSUE #2840: Wrong count of board totals with unapproved topics
1 - created a board and set it to post moderation.
2 - with a test user created a topic.
3 - admin replied to said topic but not approve the topic

Now, on Board Index, that board has totals: "1 posts, 0 topics". And shows it for everyone.
Recount totals does not fix this.

Should these replies count towards totals?

MissAllSunday:
-  Allow basic BBC tags on board and cat descriptions fixes #2821
-  Allow basic BBC tags on membergroup descriptions fixes #2823
  
  Parsing bbc on every board description has consequences, however, due the lack of a way to parse static tags without parsing dynamically ones, using parse_bbc() with a list of allowed tags seems the way to go.
- If the topic is unapproved the message automatically becomes unapproved too fixes #2840
  
  Instead of messing around with already convulsed queries, let use try a different approach, each message under an unapproved topics becomes unapproved automatically, yes I realize people will ask why their messages were set as unapproved (I am THE ADMINZZZZ) and stuff but hey, better complain about that than complaining their forum is slow or acting bizarrely.


MissAllSunday:
ugh, just realize that the OP cannot see the replies... meh... it looks like some queries will have to be changed....

margarett:
Yeah, I would say that we can leave that "unnaproved" stuff as it is. 2.0 behaves exactly the same ;)

MissAllSunday:
The reply been approved needs to be addressed though as it can be viewed as a privacy issue.

I just chose the "unapprove the reply too" approach because it seems easier to do that than having to modify every every query where a message is displayed.

MissAllSunday:
OK, the OP now can see all unapproved replies

MessageIndex still shows 0 replies for topics with unapproved replies, dunno if this is something we need to fix.

Meanwhile added some more generic fixes:
- Remove the "forum_width" theme setting.
- Allow dynamically created DOM elements to be able to use the "you sure" on click event.
- Use .sourceEditorInsertText(text) to insert the partial quote exactly where the cursor is.

illori:
dont you need to remove

```
// Save some database hits, if a width for multiple wrappers is set in admin.
    if (!empty($settings['forum_width']))
        echo '
    <style type="text/css">#wrapper, .frame {width: ', $settings['forum_width'], ';}</style>';
```

 from index.template.php?

also 

```
array(
            'id' => 'forum_width',
            'label' => $txt['forum_width'],
            'description' => $txt['forum_width_desc'],
            'type' => 'text',
            'size' => 8,
        ),
```

from Settings.template.php?

XinYenFon:
also the strings from language file.

MissAllSunday:
What other language strings are there to remove?

illori:
None that I saw.

XinYenFon:
oh lol yea my bad sorry :D you already removed the required strings.

MissAllSunday:
Added a couple of things:
- Add a comma if there are custom columns added via hooks. Otherwise the implode will mess up the query
- Updated the at.js script to its version 1.2.0, added the minified version.

MissAllSunday:
Added more stuff:
- $topicInfo is now $context['topicInfo'] so that mods can reference it freely.
- $settings['avatars_on_indexes'] has been split into two, $settings['avatars_on_messageIndex'] and $settings['avatars_on_boardIndex'] to have more control over each situation.
- if either of those settings are set, the query will load the attachment info, on 2.0.x the avatar will only be displayed if it was an external url or one from the predefined galleries.
- avatar on message index now also loads the first post member avatar
- Load::set_avatar_data() is a generic function to help load member's avatars with the given info, used in both boardIndex and messageIndex
- hook tag in package-info.xml now supports an "object" action to let mod authors to determinate if their hook needs to be instantiated
- hook call didn't passed the "permanent" argument to add/remove integration_function making it fail to load the hooks file.
- credits tag in package-info now supports an "licenseurl" action to help mod authors to link to their respective license page
- $context['copyrights']['mods'] has been moved right after $context['credits_modifications'] and it displays its info in the same place $context['credits_modifications'] does.

live627:
I thought `$topicInfo` was already a global.

MissAllSunday:
It is, however, I rather prefer to keep all this type of vars as keys inside $context as it provides some some of uniformity compared to having a stand alone var been globalized.

live627:
I agree, helps curb confusion

MissAllSunday:
Damn it, I keep forgetting this hasn't been merged yet.

Oldiesmann:
I can go ahead and merge it in if everything's ready to go.

MissAllSunday:
Sure, I've been using this branch on my site for quite some time without any major issues.

MissAllSunday:
Addressing this: http://www.simplemachines.org/community/index.php?topic=537345.0

SMF does not use the refresh param in any instance, in fact all admin setting page uses header location as a way to "refresh" the current admin page when settings are been saved, this of course gets copied by every mod too.

I'm kinda curious to know which mod prompted this error it might be just one single mod and we will never hear about it ever again but if we do, its a simple fix anyway.

I didn't remove the $refresh param to keep up compatibility, thing is, the hook in 2.1 also passes the $permanent var which might confuse some mod authors... I don't know if this particular hook was used by some mod or if its safe to just replace the $refresh var with $permanent and hope whatever mod it was using this hook gets properly updated.

PR #2842, opened 2015-04-23
Allow mod authors to mess things up!

MissAllSunday:
Adds a way to modify the predefined css classes of each element in BoardIndex. MessageIndex and Display.

It also allows topic in MessageIndex to have multiple css classes:

multipleclasses


live627:
Shouldn't this also be applied to the unread pages?

PR #2844, opened 2015-04-26
Small change - CLOSED

ghost:
quick way for users to decide if they want icons in nav tabs

### DIFF
--- a/Themes/default/Settings.template.php
+++ b/Themes/default/Settings.template.php
@@ -124,6 +124,11 @@ function template_options()
 			'label'  => $txt['drafts_show_saved_enabled'],
 			'default' => true,
 		)
+		array(
+			'id' => 'button_icon',
+			'label'  => $txt['button_icon_enabled'],
+			'default' => true,
+		)
 	);
 }
 
--- a/Themes/default/index.template.php
+++ b/Themes/default/index.template.php
@@ -451,8 +451,8 @@ function template_menu()
 	{
 		echo '
 						<li id="button_', $act, '"', !empty($button['sub_buttons']) ? ' class="subsections"' :'', '>
-							<a', $button['active_button'] ? ' class="active"' : '', ' href="', $button['href'], '"', isset($button['target']) ? ' target="' . $button['target'] . '"' : '', '>
-								', $button['icon'], $button['title'], '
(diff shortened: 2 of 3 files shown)


MissAllSunday:
- May be better to change the if condition with a ternary, for simplicity.
- We cannot merge your commit as you didn't sign it.

XinYenFon:
Is this setting really needed theme authors can remove icons if they don't want it in their themes.

-----Original Message-----

MissAllSunday:
I'll be more inclined towards a theme setting, defined in index.template, that way themers doesn't have to manually remove the icons. Just define and set the var.

ghost:
modify it as you wish not a massive issue, giving users more flexibility will save them having to edit the code.  End of day if you dont include easy to make it a mod as alot of people will ask for this.

MissAllSunday:
@Pouvik  This really isn't an user UI, you are assuming all themes will have icons which will not always be the case.

This will be better if there was a var defining if the the current theme uses icons and then and only then show that setting to the users.  Your PR is fine it just lacks the necessary logic to handle all possible cases.

Your last sentence didn't made much sense.

ghost:
Then why dont you respect the fact I have closed this, and add the var yourself and submit this idea in a fix of your own, its not like its hard to do.  Now please respect the fact that this has been CLOSED!

MissAllSunday:
The PR itself is closed yes, the conversation is not, while it is your right to close this, the conversation might still going on since more people can be interested on sharing ideas and/or improvements.

You can chose not to receive notifications for this conversation, thats within your rights too.

ghost:
well I am requesting you drop it, if you like idea then work on it yourself!

MissAllSunday:
Don't see the need to drop this since we are already discussing this idea here. If we agree then a new PR will be submitted, meanwhile discussions will be held right here. Thank you for the idea.

Anyway... back on topic, the $setting var to declare icons usage should be done, dunno about the user setting though, that might be overkilled.

ghost:
discussion has not been on that until now discussion has been on you being ignorant and failing to comply with a request!

MissAllSunday:
Your whole message didn't make any sense.

PR #2847, opened 2015-04-30
Fixes001

FearThe1337:
Some minor fixes which for errors which kept appearing in my error log.


MissAllSunday:
Which errors?

FearThe1337:
Errors on $condition not being defined. Its only being defined once as '' in an [else statement](https://github.com/FearThe1337/SMF2.1/commit/de06b0c76e77228114888bfff006663c508e0867#diff-a2aa83d2eff230e8a37219a7407dbf4eL476).

In the case of duplicate permissions, it happens when creating a new group based on maintenance. It will warn about a key already existing (It tries to enter the same permission 2x),

Third one, not sure what causes it, but it happens a lot somehow.

PR #2854, opened 2015-05-08
Restoring Span For $button['title']

ahrasis:
This PR is only a suggestion for easy customization from this topic:
http://www.simplemachines.org/community/index.php?topic=536271.0. The
text menu can be easily removed by css display:none with the link and
icons left intact if the span is restored. There is no other easy way to
customize removing the text from menu so far without the span. While I
am pulling this request, I will leave this totally to the SMF Team
discretion to decide.

Signed-off-by: "ahrasis" Ahmad Rasyid Ismail ahrasis@gmail.com

### DIFF
--- a/Themes/default/css/responsive.css
+++ b/Themes/default/css/responsive.css
@@ -40,7 +40,11 @@
 		margin: 0 0 0 -85px;
 	}
 	.dropmenu li {
-		width: 50%;
+		width: 10%;
+	}
+	.dropmenu li span.textmenu
+	{
+		display:none;
 	}
 	.dropmenu li a {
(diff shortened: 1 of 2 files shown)


XinYenFon:
Making menu only icon in default and listing them side-by-side will disable its ability to use secondary/third menu on mobile device. That's why they are listed one-by-one in the list. Please revert changes in responsive.css rest is good to go :)

live627:
First commit was cherry picked https://github.com/SimpleMachines/SMF2.1/commit/5e6d8824877ac266883645bf73463f16b8031b11

PR #2859, opened 2015-05-15
Native languages support

### LINKED ISSUE #2857: "Native" language names
Should we include this?  http://www.simplemachines.org/community/index.php?topic=536359.msg3811781#msg3811781

Needs to be done differently of course, each language will have to define its own "native" name in a string inside index.language and use that string. Should only use entities or otherwise be "safe" in whatever encoding the currnet user is using.

MissAllSunday:
Fixes #2857
- Adds support for "Native" language names.
- Tries to load the target file using the current encoding, it heavily depends on the language's encoding file.

Did some test with the Spanish files but I have absolutely no idea how this will behave if someone is using Russian (windows-1251) and wants to display another equally weird encoding language like Greek.


MissAllSunday:
ugh, nope, nevermind, this is failing even with simple entities...

MissAllSunday:
OK, been toying with this using different approaches, I don't think it will be possible to actually do this as there is no sane way to get a string in x encoding to another x encoding.

The only reasonable way to implement this is to "represent" every native language name in HTML entities, so Russian for example:

$txt['native_name'] = 'русский';

becomes:

$txt['native_name'] = '& #x440;& #x443;& #x441;& #x441;& #x43A;& #x438;& #x439;';

And if thats the case, then it will be far more easy to store a txt file on SMF.org with a list of all supported languages and its "html entities" name and fetch that instead of loading the entire index.language file and messing up with file encodings.

MissAllSunday:
A different approach:
- Loads the index.language file and tries to encode it to utf8.
- If the file was successfully encoded,get the string, pass it to un_htmlspecialchars() and use it.
- Stores the langlist as a serialize array in a $modSettings entry.
- If something miserably fails, fallback to the old lang name.

This needs every native language name to be "written" in full HTML entities.

Are mb_  functions safe to use or should we add fallbacks for weird servers?

live627:
mb isn't enabled by default so I'd imagine that many hosts leave it off.

> Loads the index.language file and tries to encode it to utf8.

Will that really work? I mean, like, there are sooo many charsets that SMF
uses.I'm shooting in the dark here, Can the functions detect the charset,
or do they need manually fed?

Until we go utf8 only, this endeavor might be for naught....

MissAllSunday:
The file_get_contents_utf8() func is supposed to convert to utf8 no matter what charset the original file has  but if mb_ functions aren't that reliable and iconv isn't an option either, we can remove that part and just rely on each language using full entities.

Oldiesmann:
iconv is enabled by default in PHP, so that should be a fairly safe choice.

MissAllSunday:
The convert the whole file to utf8 wasn't that necessary, was just an extra safety layer.

I still feel awkward trying to load the entire file just to look for a var, can be messy if a site happens to have a bunch of languages installed (like certain smf support forum :smile: ).

It also feels weird to store actual entities as a serialized string as there is also the (overly paranoid?) concern about using serialize() itself on data we don't fully have control over it, perhaps it might be better to just implode/explode the language names.

live627:
What to do with this?

MissAllSunday:
Loading the entire file was unnecessary, reverted that and used a simpler fopen() call.
Move the $txt['native_name'] closer to the top to save some cycles.

XinYenFon:
I'm curious why this is not tagged for Beta3 ? :)

jdarwood007:
I would just like to note we should be using json instead of serialize functions for consistency of removal of serialize from the code base.

MissAllSunday:
Yes, I'm in the process of doing just that but in order to do that, we need to take older installations into account and perform the change from serialize to json on upgrade, unfortunately, hunting down each serialized setting its a royal pain in the posterior...

Theres also the notification system, it currently uses serialize because some issues with json and non-utf8 strings, the notification system does include saving large chunks of user supplied data which can and surely will be encoded in weird encodings making json functions to fail or otherwise return malformed data so either way we chose, serialize or json we will still face issues.

jdarwood007:
Yea, especially any user supplied should be json.  Serialize has been known for a while to be full of security issues and shouldn't be trusted with any user input.  Removing it in general just makes the code base consistent.

Oldiesmann:
Could we possibly use iconv or similar functions to work around the encoding issues with that?

MissAllSunday:
Not really, converting to and from a weird encoding will only cause issues, thats why I suggested to use full HTML entities to be sure whatever weird encoding an user is using won't interfere and the language names will always be displayed correctly.

MissAllSunday:
Since we now use utf8 encoding issues shouldn't be a problem anymore...

PR #2863, opened 2015-05-17
Replace the define SIMPLE_ACTION with an entry in $context.

FearThe1337:
Required for error handling in the print page. If the printing view is disabled on the forums, it should be displayed as non-simple action and therefor requires the simple_action to be dynamic.


MissAllSunday:
Looks good, would you mind to change both ifs to:

```
if (!empty($context['simple_action']))
```

For clarity.

FearThe1337:
Updated to use the empty function.

MissAllSunday:
Thanks!

PR #2869, opened 2015-05-22
Fix #2765 - Post merging resulting in wrong stats

### LINKED ISSUE #2765: issues with merging topics
http://www.simplemachines.org/community/index.php?topic=534371.0

seems that when you merge topics the count on the board index may be incorrect for number of topics in the board.

depending on number of topics, this may make the board show as having no topics at all.

seems that after the merge the number of topics in the board [if both topics in the same board] go down by 2.

FearThe1337:
- Fixes merging topics without a redirect post having wrong amount of replies
- Fixes wrong board totals for merges.
- Fixes a check for when dealing with only unapproved posts.

### DIFF
--- a/Sources/RemoveTopic.php
+++ b/Sources/RemoveTopic.php
@@ -240,8 +240,9 @@ function RemoveOldTopics2()
  * @param array/int $topics The topics to remove (can be an id or an array of ids).
  * @param bool $decreasePostCount if true users' post count will be reduced
  * @param bool $ignoreRecycling if true topics are not moved to the recycle board (if it exists).
+ * @param bool $updateBoardCount if true topics the board totals are adjusted.
  */
-function removeTopics($topics, $decreasePostCount = true, $ignoreRecycling = false)
+function removeTopics($topics, $decreasePostCount = true, $ignoreRecycling = false, $updateBoardCount = true)
 {
 	global $sourcedir, $modSettings, $smcFunc;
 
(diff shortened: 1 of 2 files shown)


margarett:
I have been trying this branch. I still found an error in maths (I know how messy that is...)

How to reproduce:
I had 1 topic in the "General Discussion" board, 3 replies.
*\* In BoardIndex, the board has 1 topic, 4 posts. Good :)
I split it in half, so I got 2 topics, 1 reply each.
*\* In BoardIndex, the board has 2 topics, 4 posts. Good :)
I merge the 2 topics back together, leaving the redirect topic.
*\* In BoardIndex, the board has 3 topics, 5 posts. Not good :cry: 5 posts is good, 3 topics not, it should be only 2.

Not leaving the redirection topic behind produces correct counts :stuck_out_tongue_winking_eye:

FearThe1337:
Interesting, I tried with and without redirect post and the counts seemed fine.
I'll try to check tomorrow as it's almost midnight for me at the moment :)

Edit: Just woke up and took a quick look. Weird thing is, I actually knew this at one point and forgot to fix it :confused: 
The problem was/is that I am increasing the post count of all boards with 1 to account for the redirect topic, however, there will be 1 less redirect topic, namely the merged topic. Still, I should not have missed this during testing however.

FearThe1337:
Ok, that should fix it :rabbit:

FearThe1337:
Now also updates subject. (Else check for errors would complain, and I assume it makes it easier for people to find the redirect based on the subject)

margarett:
I haven't yet tested your new fixes, although I noticed that many code changes are using spaces instead of tabs, which is something against the coding guidelines :(

FearThe1337:
Oops. I normally always change them before doing a commit, as my editor always enters spaces instead of tabs. Must of forgotten to change them in my last commit (Update search index correctly).
Submitted a new commit to fix those.

margarett:
Gave this another test. Split and merge all seem to work fine and post count is always correct.

Thank you for this :)

Issue #2868, opened 2015-05-21
Feature request: Memcached via sockets


heidi666:
Pretty self-explanatory, SMF only supports port access to memcached. I'm running socket access on my main server so having the forum being able to use memcached as well would be great.

FearThe1337:
Seems extremely easy to change. Ill make a PR for this.

Edit: [PR](https://github.com/SimpleMachines/SMF2.1/pull/2871)

PR #2871, opened 2015-05-22
Add support for unix sockets for memcached. #2868

FearThe1337:
Memcache PHP documentation states the following:

> This parameter may also specify other transports like unix:///path/to/memcached.sock to use UNIX domain sockets, in this case port must also be set to 0.

Memcached PHP documentation:

> As of version 2.0.0b1, this parameter may also specify the path of a unix socket filepath ex. /path/to/memcached.sock to use UNIX domain sockets, in this case port must also be set to 0.

Best solution seemed to be to detect if there is a / or not. Since syntax differs for both memcached implementations.


Sources/Load.php
```diff
@@ -3235,7 +3235,20 @@ function get_memcached_server($level = 3)
 	global $memcached, $db_persist, $cache_memcached;
 
 	$servers = explode(',', $cache_memcached);
-	$server = explode(':', trim($servers[array_rand($servers)]));
+	$server = trim($servers[array_rand($servers)]);
+	
+	$port = 0;
```
nanaya, line 3240: probably better inside the socket-specific block? Will also match the comment.

PR #2882, opened 2015-05-25
Various password recovery fixes. (Fixes #2848)

### LINKED ISSUE #2848: Password recovery by email broken when a user has recovery questions.
Recovery by email seems to redirect to the same page without sending an email and the selection choice text header is broken (Not sure how to call it, it used to say something like choose an recovery option).

2015-05-02 20_51_03-authentication reminder

FearThe1337:
- Allow users to recover their account using their secret question if no email is set.
- Default to secret question if the user has no email address yet has recovery questions.
- Allow users to get their password emailed if they have a secret question.

Signed-off-by: FearThe1337 git@fearthe1337.com

### DIFF
--- a/Sources/Reminder.php
+++ b/Sources/Reminder.php
@@ -85,8 +85,7 @@ function RemindPick()
 		FROM {db_prefix}members
 		WHERE ' . $where . '
 		LIMIT 1',
-		array_merge($where_params, array(
-		))
+		$where_params
 	);
 	// Maybe email?
 	if ($smcFunc['db_num_rows']($request) == 0 && empty($_REQUEST['uid']))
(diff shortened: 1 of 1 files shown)


illori:
in what cases would a user not have an email address? in SMF users are required to have an email address.

FearThe1337:
Good point, however in the case it does occur, should we not just allow them to recover using their recovery questions?
It seems weird to just deny recovery, even if there are recovery options available.

Also, I guess it might be perhaps possible from migrated forums?

illori:
IMO even a forum migrated to SMF would have email. 

if it were to fail, maybe going back to the idea of a contact form would be a better idea.

FearThe1337:
Adjusted.

illori:
if this has been resolved, the issue should be closed.

PR #2894, opened 2015-06-02
Allow 2FA authentication to be preserved for 30 days

Dragooon:
This allows an user to identify a computer and prevent having to enter the 2FA code every time they need to use it


Dragooon:
I'll merge this after we get beta 2 out

Oldiesmann:
If it's ready to go, merge it now. There's no reason this should wait until after Beta 2 since we're not going to be getting it out before next week at the earliest anyway.

PR #2910, opened 2015-06-27
Bug fix: Avatars can't load using a HTTPS url & Give default avatar to last_post in MessageIndex.php and Subs-BoardIndex.php

HorusGoul:


tinoest:
Just curious but why doesn't SMF use '//<url>' this will either be http or https depending on what your currently viewing it under. 

https://en.m.wikipedia.org/wiki/Wikipedia:Protocol-relative_URL

PR #2914, opened 2015-07-06
MySQL Installer Updates

tallona:
Updated some keywords to use standard words for easier reading, added 'idx_' prefix for all indexes, removed NOT NULL from primary key columns as it is implicitly added when a column is defined as a primary key.

Signed-off-by: Adam Tallon adamtallon@gmail.com

### DIFF
--- a/other/install_2-1_mysql.sql
+++ b/other/install_2-1_mysql.sql
@@ -1825,12 +1825,12 @@ INSERT INTO {$db_prefix}scheduled_tasks
 	(id_task, next_time, time_offset, time_regularity, time_unit, disabled, task, callable)
 VALUES
 	(1, 0, 0, 2, 'h', 0, 'approval_notification', ''),
-	(3, 0, 60, 1, 'd', 0, 'daily_maINTenance', ''),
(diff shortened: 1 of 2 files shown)


live627:
Neat.

other/install_2-1_mysql.sql
```diff
@@ -1825,12 +1825,12 @@ INSERT INTO {$db_prefix}scheduled_tasks
 	(id_task, next_time, time_offset, time_regularity, time_unit, disabled, task, callable)
 VALUES
 	(1, 0, 0, 2, 'h', 0, 'approval_notification', ''),
-	(3, 0, 60, 1, 'd', 0, 'daily_maintenance', ''),
+	(3, 0, 60, 1, 'd', 0, 'daily_maINTenance', ''),
```
live627: This will break SMF. Please revert.

tallona:
Thanks, added that change.

live627:
There are still a bajillion inserts that have been modified to use the wrong case. Can you also revert those?

tallona:
Sorry about that, find and replace can be your friend but can also be your enemy at the best of times :)

PR #2917, opened 2015-07-10
fix XHTML tags which are invalid and obsolete on HTML5

Rupurudu:
Since we are switcing to XHTML 1.0 > HTML 5 some of our html is invalid
and some are obsolete, which is this pull fixes.

These include:
Removal of CDATA tags (invalid on HTML5)
Removal of type attribute on link and script tags (purely advisory,
works without it)

Signed-off-by: Mustafa Can Elmacı mustafacan@elmaci.net


live627:
Thanks

XinYenFon:
Nice thanks :)

PR #2921, opened 2015-07-12
license

live627:


XinYenFon:
@Kindred-999 can you take a peak at this ?

Kindred-999:
I am confused....    what is the purpose of this?
we already have the line: 
- @license http://www.simplemachines.org/about/smf/license.php BSD 
  in the file headers
  If this is going to reproduce that, then it would need to use number bullets instead of just *
  additionally, the copyright year is wrong....

live627:
> I am confused.... what is the purpose of this?I am confused.... what is
> the purpose of this?

Standards

PR #2948, opened 2015-07-26
Fix missing jQuery file; add mentions for all sc-editors on enabled p…

AnthonyCalandra:
…ages. Signed-off-by: Anthony Calandra anthony@anthony-calandra.com

This commit fixes an error being thrown in the topic display where the browser can't find `jquery.atwho.js` since it doesn't exist. Also, adjusted the jQuery selectors that attach the `atwho` plugin to all sc-editors when loaded on the page; this includes adding support for quick reply and creating PMs.

### DIFF
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -1293,7 +1293,7 @@ function Display()
 	// Mentions
 	if (!empty($modSettings['enable_mentions']) && allowedTo('mention'))
 	{
-		loadJavascriptFile('jquery.atwho.js', array('default_theme' => true, 'defer' => true), 'smf_atwho');
+		loadJavascriptFile('jquery.atwho.min.js', array('default_theme' => true, 'defer' => true), 'smf_atwho');
 		loadJavascriptFile('mentions.js', array('default_theme' => true, 'defer' => true), 'smf_mention');
 	}
 }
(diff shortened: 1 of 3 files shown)


live627:
Thanks for this, although I don't think mentions should be allowed in PMs.

AnthonyCalandra:
@live627 The way I see it, allowing PMs have the mention feature makes it easier to provide links to member profiles; maybe useful for moderators or administrators. Although, iirc it doesn't create an anchor in the PM message... I can fix that if I know it will be included.

live627:
Thing is, I'm not sure if the mentions mechanism is tied into the alerts
system. Notifications of the mention shouldn't propagate if they can't see
the message.

I don't remember offhand and I don't have the code handy right now to check.

illori:
it is tied into the alerts system, so if the user cannot ever see the message they should not receive an alert. if that is not checked for this then it should not be added.

AnthonyCalandra:
@Illori I did not receive a notification when testing but I'll look into the code to verify this when I get some time.

illori:
so if you dont receive the notification why enable the feature?

AnthonyCalandra:
@illori It's an easier way to link to member's profiles imo.

illori:
and how often do you have to link to a users profile when you are pming someone? not too often. usually typing out their name is enough.

live627:
I don't reference people in PMs at all simply because there is no need.

AnthonyCalandra:
Then don't add it.

live627:
If you reverse  your changes to `Sources/PersonalMessage.php`, I can merge this..

PR #2967, opened 2015-08-09
Improve group requests

live627:
-  Add an interface to see old/closed requests
-  Add membergroup request alerts


Themes/default/Profile.template.php
```diff
@@ -2065,6 +2065,8 @@ function highlightSelected(box)
 				<input type="hidden" name="', $context['session_var'], '" value="', $context['session_id'], '">
 				<input type="hidden" name="u" value="', $context['id_member'], '">
 			</form>';
+
+	template_show_list('group_request_list');
```
XinYenFon: Creating undefined func. error

XinYenFon:
This seems to not related to this PR but... Group request creating the alert but only showing the number (or bump the number) after you open Alerts and reload the page.

Accepting the membergroup request does nothing, I can go back and request again.

live627:
The background task thingies are a royal PITA to debug  since undefined function calls and other syntax errors get eaten up by the GIF output.

Maybe I can change that since I replaced the mage request JS with `$.ajax()`.

live627:
> This seems to not related to this PR but... Group request creating the alert but only showing the number (or bump the number) after you open Alerts and reload the page.

I don't know what you mean...

> Accepting the membergroup request does nothing

I hacked in some voodoo magic to fix that.

Oldiesmann:
One extremely minor issue... This PR adds a blank line to the end of several files that previously didn't have one (Groups.php, ManageSettings.php, Profile-Modify.php, Subs-Notify.php and the upgrade SQL files).

XinYenFon:
okay seems issue from my side. But still a question in mind, with reason or not refusing not sending Alert or any notification to user ?

XinYenFon:
oh and this created in errorLog

```
Type of error: Critical
The database value you're trying to insert does not exist: id_member
Function: logActions
File: .../Sources/Logging.php
Line: 527 
http://localhost/kittenground/cron.php 
```

live627:
> But still a question in mind, with reason or not refusing not sending Alert or any notification to user ? 

Your questions are seriously malformed, thus I can't decipher your message.

User should get notifications of acceptance or rejection, with or without a reason. Make sure thee related preferences are checked, and update your branch, as my latest commits fix some errors.

XinYenFon:
Typing from phone is not a good thing. I'll check things based on your suggestions.

live627:
If no one has any further comments, I'll merge this in.

XinYenFon:
I don't think any further pre-tests needed. We can test deeply after merged into core code. Go for it.

PR #2970, opened 2015-08-12
Birthday

live627:
Convert birthday emails to the alerts system


live627:
This depends on some bug fixes in #2967

PR #2972, opened 2015-08-13
Email digest

live627:
It now uses the alert prefs.


live627:
This is more broken than anticipated.

PR #2976, opened 2015-08-15
Remove all traces of the defunct WAP code

live627:
Because progress!


XinYenFon:
Something I support but I'm not sure. Opera Mini does not support HTML5/CSS3 yet it offers compressed web-mode which saves dozens of data.

live627:
Turbo mode?

Rupurudu:
Opera mini is used on very low end dumb phones. These devices need several seconds to render pages. On big pages with heavy css and giant images, these devices generally crash.

Opera turbo per-renders pages on opera servers, speeding up the rendering on the phone.

Personally, I'm opting for removing the wap code, since these devices are very rare.

live627:
I was asking for clarification on the mention of Mini's "compressed" mode which sounds like turbo. Anyway, it does not use WAP but magic on their proxy servers.

http://www.opera.com/turbo

PR #2984, opened 2015-08-17
Phpdoc Updates

Oldiesmann:
This massive set of commits cleans up, updates and adds phpdoc comments to all the functions in the Sources and template files.


Oldiesmann:
Note that I am aware of the merge conflict and will figure that out in the coming days...

PR #2997, opened 2015-08-22
Hi

GrCOTE7:
I think using GitHub for language's files could be a good solution...

If you're agree, can you merge this or better initiate a new GitHub project, for all languages...?


XinYenFon:
Nope... We are not accepting this type of translation please apply to translation group.

edit: typo

PR #2999, opened 2015-08-22
Fix PMs permission checking

socram8888:
Signed-off-by: Marcos Vives Del Sol socram8888@gmail.com

`list()`, as the PHP manual itself says, can be only used on numeric arrays. Therefore it was failing when used on an array which has "allowed" and "denied" indices instead.


live627:
Nice catch!

PR #3013, opened 2015-08-29
Fixed mistakes in function ssi_recentTopics

OVer711:
id_first_msg, id_last_msg


Oldiesmann:
See my comments on the changes... Also, you need to sign off on your commits or we can't use them.

MissAllSunday:
Unfortunately, we cannot accept this as the commit isn't signed off by the author, aditionally, it will be better to add both last and first poster info as thats the default behaviour on SMF.

PR #3017, opened 2015-08-30
Dev

live627:


live627:
Forgot to mention:

This enables formatted posts to be selectivity quoted.

PR #3018, opened 2015-09-01
Don't ignore $_SESSION['login_url']...

Oldiesmann:


live627:
How  would I reproduce this?

Oldiesmann:
Using ssi_login() is easiest. Create a file with this code:

``` php
<?php
include_once('/path/to/SSI.php');
ssi_login('url_goes_here');
?>
```

Without this fix, it should just redirect you to the board index. With the fix it should redirect you to whatever site you specified.

PR #3036, opened 2015-09-07
Simplify loadCSSFile and loadJavascriptFile 

live627:
Simplify loadCSSFile and loadJavascriptFile so they don't seem so complicated with all those needless params.

Also respect base theme.


MissAllSunday:
This brakes the ability to use external files.

default_theme is a useful param too. I used on my mods to avoid the need to copy all my css and js files on every theme the forum has and avoid support request because some css file was missing on their recently installed theme.

socram8888:
@MissAllSunday I the fallback on missing .js and .css file is still there. It's just that it's now forcefully enabled. If it fails to find on the current theme it attempts to use default's.

MissAllSunday:
Yes but I somehow prefer the old method, there is a slight difference between  a "fallback" and explicitly require the file to be loaded from the default theme (skipping the current theme altogether).  There might be times when you want to skip the current theme, perhaps the current theme has a file the same name as your mod file, highly impossible but you never know.

Anyway, I'm planning on adding a quick way to "disable" files for themers to use to give them more flexibility on decide what they want to use and what not.

live627:
The same methodology is used in loadLanguage().

MissAllSunday:
Yeah but the external files needs to be added back, SMF uses it to load the CDN versions of jquery.

Issue #3026, opened 2015-09-05
WYSIWYG and mentions/quote selected


Oldiesmann:
Per http://www.simplemachines.org/community/index.php?topic=539497.0

Mentions: Doesn't work with WYSIWYG disabled ("Uncaught TypeError: $this.inputor.caret is not a function jquery.atwho.min.js 1:6731")

Quote selected: Doesn't work with WYSIWYG enabled ("Not well formed <unknown> 1:75" - possibly an XML error?)

Mentions work fine with WYSIWYG enabled and the "quote selected" feature works fine with WYSIWYG disabled.

Oldiesmann:
Interesting find here... If I load the jquery.caret.min.js file, mentions will work in non-WYSIWYG mode, though Firefox will still spit out a bunch of "Not well-formed" errors (the exact same one it gives for trying to use "quote selected" in WYSIWYG mode). I have no idea what it thinks isn't well-formed though because it says "<unknown> 1:75" (meaning character 75 on line 1 of "unknown"). I don't get any "not well-formed" errors in Chrome at all though.

Oldiesmann:
The mentions issue was fixed with #3044, but the quote selected issue still remains.

MissAllSunday:
I'll take a look at it.

PR #3044, opened 2015-09-09
Partial fix for #3026

Oldiesmann:
This fixes the issue with mentions not working in non-WYSIWYG mode in the quick reply box. However, something is causing Firefox to complain about "not well-formed" JS somewhere along the lines with this fix. Everything still works though.


live627:
Seems you forgot to branch.

Oldiesmann:
Yeah, I initially committed that fix to the release-2.1 branch and forgot to fix the new branch I made while fixing the other one. It's fixed now though.

PR #3046, opened 2015-09-10
[WIP] THEME CODE

### LINKED ISSUE #3033: Presentation in Sources again
```
    // Are we showing debugging for templates?  Just make sure not to do it before the doctype...
    if (allowedTo('admin_forum') && isset($_REQUEST['debug']) && !in_array($sub_template_name, array('init', 'main_below')) && ob_get_length() > 0 && !isset($_REQUEST['xml']))
    {
        echo '
<div style="font-size: 8pt; border: 1px dashed red; background: orange; text-align: center; font-weight: bold;">---- ', $sub_template_name, ' ends ----</div>';
    }
```

Ok, so why does this have to be defined by inline CSS? Why not just bung an existing class on the div and be done with it? Some admins may not want the thing having an orange background and a dashed red border. I sure wouldn't. That'd be as ugly as a sack full of arseholes.

It's only a notification of a template name.

live627:
Current work of condensing the CSS. Several areas will still be broken. Brief summary of changes:
- Add a 16 column grid that uses inline block display
- Combine progress bars and stats bars
- Popup uses centerbox (nobody cares 🍩)
- The meat and gravy: remove extra/unused styles

This will also make more things responsive. More centralized code  = easier to maintain, amirite?

Fixes  #3033


Themes/default/BoardIndex.template.php
```diff
@@ -108,7 +110,7 @@ function template_main()
 
 				echo '
 
-						<p class="board_description">', $board['description'] , '</p>';
+						</h4><p>', $board['description'] , '</p>';
```
XinYenFon: why reverting this ?

Themes/default/MessageIndex.template.php
```diff
@@ -124,7 +126,7 @@ function template_main()
 		{
 		echo '
 	<div id="description_board" class="generic_list_wrapper">
-		<h3>', $context['name'], '</h3>
+		<h3 class="floatleft">', $context['name'], '&nbsp;-&nbsp;</h3>
```
XinYenFon, line 60: or this ?

Themes/default/BoardIndex.template.php
live627: mistake, I'll lfi that soon.

Themes/default/MessageIndex.template.php
live627, line 60: also mistake, I'll lfi that soon.

XinYenFon, line 60: oh lol sorry :P

StealthWombat:
Umm, what is all this "grid on board index" stuff in aid of? The class names have no relation to what they are doing on the board index, and there are far more of them. How is this supposed to be better than what is already there? We've just re-done the index to use table-cell display anyway.

ETA: Just as one obvious point to consider: what if someone decides they don't want text-align: center; on the stats div? Sure, it's styled that way by default, but people have a habit of changing presentation in themes.

StealthWombat:
Oh crap. I missed those tiny yellow tags. They really should do something less stupid there. Ok, nvm. Prefixes r us I suppose.

Although I'm fairly sure this could be handled with CSS 2.1. Used to be, IIRC.

(IMO "requires prefix" = "partial support")

XinYenFon:
Since you started why don't we go one step forward and remove `<table>` from the SMF and use CSS model tables ?

StealthWombat:
@Antes: for data which actually is tabular, it'd only be worth changing the markup if you were also going to change the presentation on smaller screen to allow more data to be displayed. IOW, if you were planning on doing something that an actual table wouldn't allow you to do.

If you're going to stick with dropping cells to keep everything on on line on small screens, a table is just as good or even (arguably) better for the dreaded religion of semantix, and will make no difference to presentation.

So it comes down to what you want the finished product to look like on small screens.

XinYenFon:
Considering with those changes CSS going more framework alike position, I think using CSS table-display far more better and modern. Currently we are dropping alot of data in admin panel, are they really needed (maybe?) but surely we can give more data to admins from mobile view.

PR #3094, opened 2015-10-04
Update things with the 2.0.11 security patch

live627:
Ready for review.

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -434,7 +434,7 @@ function loadUserSettings()
 			{
 				if (!empty($_COOKIE[$tfacookie]))
 				{
-					list ($tfamember, $tfasecret) = @unserialize($_COOKIE[$tfacookie]);
+					list ($tfamember, $tfasecret) = safe_unserialize($_COOKIE[$tfacookie]);
 
 					if ((int) $tfamember != $id_member)
 						$tfasecret = null;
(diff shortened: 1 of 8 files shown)


live627:
Ready for review.

jdarwood007:
Use json instead.

Oldiesmann:
I agree that we should be using json instead, since it's standard as of 5.2.0 (minimum for 2.1 is 5.3.8).

Oldiesmann:
I'll look into this a bit more in the coming days and see what I can do. It's going to be a bit of a pain due to having to convert any existing data that's serialized, but it's doable, and some of the data is stuff that was only added in 2.1 to begin with.

Oldiesmann:
Just realized something that's going to make this a pain to deal with... According to the PHP manual, all string data passed to json_encode must be UTF-8 encoded. I'm not sure how MySQL will handle that when we're not using UTF-8...

jdarwood007:
I see 2 cases:
1. We ignore it and users have the possibility of having problems.  I can't think why since the page has the charset set that we are working with in the database that we would see a problem.
2. Force all users to migrate to UTF-8.  There is no reason not nowadays.

At the very least if we don't force/give an option during upgrade to migrate to UTF-8, all new installs should have set default UTF-8 and warnings should popup if they try to use a non UTF-8 version.

live627:
Maybe when the upgrade is complete, tell users to migrate to utf8 if not
already done and give a link to the proper area. SMF already has a section
in the admin panel for this.

Oldiesmann:
At this point we're planning to just force the upgrade to UTF-8. The main reason we didn't do that initially is because of the amount of time it will take to do that on larger forums, but it's a good option and will simplify other things. I'm going to look at ripping out the existing conversion code and plopping it into upgrade.php instead (with some changes as needed).

live627:
Good luck. I'd help, but I'm completely burnt out right now and need a long
break.

Oldiesmann:
Closing this since it's not necessary now that all the JSON stuff is done.

PR #3100, opened 2015-10-15
Force UTF-8 conversion on upgrade

Oldiesmann:
Here is the initial commit to force UTF-8 on upgrade. Note that this hasn't been tested yet - it's mostly just modified from the previous convertutf8() function in ManageMaintenance.php. Also, it doesn't properly handle unsupported character sets at the moment. Currently if an unsupported character set is detected, SMF shows the user a warning and defaults to converting from ISO-8859-1 (the user can choose which supported character set they want to convert from). We need to either figure out how to properly handle those or just skip trying to convert for those as well.


jdarwood007:
I would suggest we move it into the upgrade.php so we don't have to eval that block of code every time it loads.  I know its for MySQL only but its our popular database type and would keep it from being slowing things due to eval being slow.

jdarwood007:
You need to do it after the upgrades to the database.  Unless you want to make the UTF-8 conversion steps compatible pre 2.0 databases, it was designed to work with 2.0 databases.  I.E. in 1.1 we used upper case and camel case whereas in 2.0 its lowercase and underscores.  The upgrader is supposed to support upgrading even from YabbSE.  As it is now, this would fail for any database pre 2.0.

Oldiesmann:
I guess I can add the code to change serialized things to JSON to upgrade.php as well then, since we'll need to do that after we do the UTF-8 conversion.

other/upgrade.php
```diff
@@ -3590,7 +3591,466 @@ function smf_strtolower($string)
 		return mb_strtolower($string, 'UTF-8');
 	require_once($sourcedir . '/Subs-Charset.php');
 	return utf8_strtolower($string);
-};
+}
+
+/**
+ * Handles converting your database to UTF-8
+ */
+function convertUtf8()
```
live627, line 3595: This should be TitleCase.

other/upgrade.php
```diff
@@ -3590,7 +3591,466 @@ function smf_strtolower($string)
...
+{
+	global $db_character_set, $sourcedir, $smcFunc, $modSettings, $language, $db_prefix, $db_type;
+
+	// First make sure they aren't already on UTF-8 before we go anywhere...
+	if ($db_type == 'postgresql' || ($db_character_set === 'utf8' && !empty($modSettings['global_character_set']) && $modSettings['global_character_set'] === 'UTF-8'))
+	{
+		echo 'Database already UTF-8. Skipping.';
```
live627, line 3602: wut

jdarwood007, line 3602: From what @Oldiesmann explained PostgreSQL ignores charset really as its already UTF-8.

Oldiesmann, line 3602: PostgreSQL doesn't have proper charset/collation support like MySQL does. You can only set the charset and collation at the database level, and only when the database is created. It does however have good support for automatic conversion between the client and the database charset. In most cases it will be UTF-8 already anyway (that's the default setting and cPanel doesn't give you an option to choose the charset or collation when creating the db).

Has anyone tested this to see how well it works? I don't have a good board to test this with at the moment, though I could theoretically copy my existing database and run tests on that.

XinYenFon:
http://localhost/utf8/upgrade.php?step=3&substep=0&data=YTo2OntzOjc6ImN1cnN0ZXAiO2k6MztzOjQ6ImxhbmciO3M6NzoiZW5nbGlzaCI7czozOiJyaWQiO2k6NTI4O3M6NDoicGFzcyI7aTo1OTM4NztzOjU6ImRlYnVnIjtiOjE7czoyOiJqcyI7aToxO30=&filecount=1

screenshot-localhost 2015-11-05 01-12-46

Oldiesmann:
Antes has said that this does work as intended, but it doesn't really show what it's doing, and the code to exit if your forum is already UTF-8 doesn't work properly, so I guess we need to tweak that some. I'll try to look into this more in the coming days.

XinYenFon:
Can't you check settings.php and look for `$db_character_set` if its set to utf8 then skip the process ?

PR #3108, opened 2015-10-24
Fixed the footer to be the same height on all pages

Justyne:
Signed-off-by: Justyne justyne@gmail.com

### DIFF
--- a/other/install.php
+++ b/other/install.php
@@ -2112,7 +2112,7 @@ function template_install_above()
 		', $txt['lang_rtl'] == true ? '<link rel="stylesheet" href="Themes/default/css/rtl.css?alp21">' : '' , '
 		<script src="Themes/default/scripts/script.js"></script>
 	</head>
-	<body>
+	<body><div id="footerfix">
 		<div id="header">
 			<h1 class="forumtitle">', $txt['smf_installer'], '</h1>
 			<img id="smflogo" src="Themes/default/images/smflogo.png" alt="Simple Machines Forum" title="Simple Machines Forum">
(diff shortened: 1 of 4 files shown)


other/upgrade.php
```diff
@@ -3837,6 +3838,7 @@ function template_upgrade_below()
 			</div>
 			</div>
 		</div>
+		</div>
```
XinYenFon, line 3841: extra tab here.

other/install.php
```diff
@@ -2112,7 +2112,7 @@ function template_install_above()
 		', $txt['lang_rtl'] == true ? '<link rel="stylesheet" href="Themes/default/css/rtl.css?alp21">' : '' , '
 		<script src="Themes/default/scripts/script.js"></script>
 	</head>
-	<body>
+	<body><div id="footerfix">
```
XinYenFon, line 2115: pick one style :P either do body-div or body(new-line)div

XinYenFon:
You forgot to sign-off last commit. If you fix them I think we are good to go here (maybe pick new ID instead footerfix).

XinYenFon:
Also I have to admit, I'm not really fan of adding extra container to template system.

Arantor:
If it were possible to fix any other way than adding new divs, it might have been doable, but having a footer that's half the page height looks terrible.

As for the adding of tags, that was mostly because no-one really wants to go through and fix all the indentation everywhere (when it's already not consistent as it is)

XinYenFon:
seems so, I think this is good to go :)

PR #3115, opened 2015-11-02
+ Give a quick command for linux users on chmod changes

jdarwood007:
This is something small that I would like to see added.  It simply provides a linux chmod option for any files we need to change permissions on.  I really like using this on most of my systems as its far more convenient to change permissions from my shell than from FTP.


jdarwood007:
Good thing I'm a programer not a English teacher.

PR #3119, opened 2015-11-04
+ Added some additional hooks

jdarwood007:
Some hooks that would be very handy to add into SMF.

### DIFF
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -1481,7 +1481,7 @@ function Download()
 
 	// Do we have a hook wanting to use our attachment system? We use $attachRequest to prevent accidental usage of $request.
 	$attachRequest = null;
-    call_integration_hook('integrate_download_request', array(&$attachRequest));
+	call_integration_hook('integrate_download_request', array(&$attachRequest));
 	if (!is_null($attachRequest) && is_resource($attachRequest))
 		$request = $attachRequest;
 	else
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -1479,23 +1479,31 @@ function Download()
 
 	$_REQUEST['attach'] = isset($_REQUEST['attach']) ? (int) $_REQUEST['attach'] : (int) $_REQUEST['id'];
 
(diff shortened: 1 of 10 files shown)


jdarwood007:
I should note I did little testing with these.  But they seem solid.

I also realize I didn't sign off, using a new Git GUI tool.  I tried via command line to amend the message but it complains about a fast forward error.  Can we just ignore that so I don't have to try and figure out how to add that.

Sources/Display.php
```diff
@@ -1479,23 +1479,31 @@ function Download()
 
 	$_REQUEST['attach'] = isset($_REQUEST['attach']) ? (int) $_REQUEST['attach'] : (int) $_REQUEST['id'];
 
-	// This checks only the current board for $board/$topic's permissions.
-	isAllowedTo('view_attachments');
+	// Do we have a hook wanting to use our attachment system? We use $attachRequest to prevent accidental usage of $request.
+	$attachRequest = null;
+    call_integration_hook('integrate_download_request', array(&$attachRequest));
```
XinYenFon: My OCD :P

Oldiesmann:
Besides the sign off thing, the only issue I see is that the indentation is inconsistent in several areas.

I'm not sure if we can ignore the sign off requirement. I did last night, but only because I was tired and wasn't paying attention...

jdarwood007:
Fixed those up and signed off.

Sources/ManageAttachments.php
```diff
@@ -1530,6 +1537,8 @@ function RepairAttachments()
 		for (; $_GET['substep'] < $thumbnails; $_GET['substep'] += 500)
 		{
 			$to_remove = array();
+			$ignore_ids = array(0);
+			call_integration_hook('integrate_repair_attachments_nomsg', array(&$ignore_ids, $_GET['substep'], $_GET['substep'] += 500));
```
jdarwood007, line 1541: This is properly tabbed, just displays weird in GitHub.

PR #3123, opened 2015-11-13
Use JSON encoding instead of serialize

Oldiesmann:
This replaces nearly every use of serialize() and unserialize() with json_encode() and json_decode() respectively. It also adds code for converting existing serialized data to JSON-encoded data during the upgrade, plus a setting to skip that once we've converted the data.

Note that this currently does not handle cookie data, but this can be added as well (we need to determine which format the cookie data is in then convert it if necessary).

### DIFF
--- a/Sources/Admin.php
+++ b/Sources/Admin.php
@@ -38,7 +38,7 @@ function AdminMain()
 	require_once($sourcedir . '/Subs-Menu.php');
 
 	// Some preferences.
-	$context['admin_preferences'] = !empty($options['admin_preferences']) ? unserialize($options['admin_preferences']) : array();
+	$context['admin_preferences'] = !empty($options['admin_preferences']) ? json_decode($options['admin_preferences']) : array();
 
 	/** @var array $admin_areas Defines the menu structure for the admin center. See {@link Subs-Menu.php Subs-Menu.php} for details! */
 	$admin_areas = array(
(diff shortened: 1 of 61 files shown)


MissAllSunday:
All json_decode calls needs to pass a boolean true as second argument to tell the function to return an array instead of an object :+1:  :) or cast the object as an array but that seems unreliable.

Oldiesmann:
Thanks for the tip. I've updated the PR to reflect that. I also realized earlier that I forgot to add the code to actually execute the new function as well, so that's been fixed too.

XinYenFon:
...upgrade.php?step=3&substep=0&data=YTo2OntzOjc6ImN1cnN0ZXAiO2k6MztzOjQ6ImxhbmciO3M6NzoiZW5nbGlzaCI7czozOiJyaWQiO2k6NDAzMDtzOjQ6InBhc3MiO2k6MTU4NjY7czo1OiJkZWJ1ZyI7YjoxO3M6MjoianMiO2k6MTt9&filecount=1

`Database already UTF-8. Skipping.Fixing the 'extra' column in the log_actions table... done. Fixing the log_online table... done. Unknown column 'id_rules' in 'field list'`

Oldiesmann:
Ran some more tests and can't figure out why it's running into trouble with some values as everything it ran into problems with (mostly empty arrays - 'a:0:{}') was valid otherwise. It did skip the things it had problems with though, so I'll go ahead and merge this then continue to investigate.

PR #3126, opened 2015-11-14
Update Admin.template.php

Underdog-01:
Step 1 of a 3 part pull request (please see other pull requests in this timeframe)

HTML5 Compliance
- object tag requires type
- form tag can not be used within h3 tag  
- div tag can not be used within h3 tag
- search results caption was barely visible

### DIFF
--- a/Themes/default/Admin.template.php
+++ b/Themes/default/Admin.template.php
@@ -1280,16 +1280,17 @@ function template_admin_search_results()
 
 	echo '
 						<div id="section_header" class="cat_bar">
-							<h3 class="catbg">
-								<div id="quick_search">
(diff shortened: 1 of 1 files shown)


Oldiesmann:
In the future, if you're making related changes to multiple files, keep all of them in one PR rather than spreading them out across multiple PRs. It's harder to keep track of things if you're opening a separate PR for each file.

Underdog-01:
I attempted to edit each file in my fork at once prior to commiting the PR and there did not seem to be a link to send them as 1 PR. It showed me 3 links with 1 PR for each file change.
I will look for the instruction to do that properly.. My apologies.

Oldiesmann:
Your best bet is to use a client like Sourcetree or Github for Windows and edit the files directly on your computer, then push the changes to GitHub that way.

PR #3128, opened 2015-11-16
Update index.php

Underdog-01:
- deprecated function throws an E_WARNING for PHP 5.3.0 and above
- either check version or get rid of it altogether

### DIFF
--- a/index.php
+++ b/index.php
@@ -25,7 +25,7 @@ $software_year = '2015';
 
 // Get everything started up...
 define('SMF', 1);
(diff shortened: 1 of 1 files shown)


Underdog-01:
I realize the error is supressed but the function concerned is indeed deprecated for PHP 5.3.0 and will acually crash PHP for the upcoming 7.0.0 version (beyond this branch anyhow). 
The SMF 2.1 documentation states you devs decided to go with PHP 5.1.0 as a minimum so I suppose the edit I just posted applies.

Issue #2881, opened 2015-05-25
Wrong text lines in email recovery.


FearThe1337:
When filling nothing in at the enter username/email field you get the following message:

```
That username does not exist.
```

When filling in a non-existing username or email you get the following message:

```
There are no usernames associated with that email.
```

Surely, I assume that the first text should be something like: "You did not fill in any username or email, please go back and fill in either your username or email"
And the second one: "There were no accounts found for...." (Not only email, as usernames are valid too!)

tallona:
I would even suggest that one message is used for all failed long attempts regardless of if there is no user with that username/email, if the field is empty or if it's a failed login attempt as you are giving a potential hacker hints as to what is and isn't a valid user account that could be targeted.

frandominguezl:
This is solved @live627 @colinschoen

live627:
closing thus then...

PR #3141, opened 2015-11-21
Generalization of dl/dt/dd +fix.

XinYenFon:
Please read commit messages for further information.

### DIFF
--- a/Sources/LogInOut.php
+++ b/Sources/LogInOut.php
@@ -268,7 +268,7 @@ function Login2()
 	// Let them try again, it didn't match anything...
 	if ($smcFunc['db_num_rows']($request) == 0)
 	{
-		$context['login_errors'] = array($txt['username_no_exist']);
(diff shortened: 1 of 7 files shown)


Oldiesmann:
Why are we returning an error of "There are no usernames with that email" when they enter an invalid username? I'm fine with returning a generic "invalid login" message, but users are going to get confused if they see "there are no users with that email" when they didn't even enter an email address in the first place.

XinYenFon:
Actually it needs calculation at first place if input is empty it shouldn't proceed with request. But same error message makes sure we are not giving anything away.

https://github.com/SimpleMachines/SMF2.1/issues/2881#issuecomment-122863865

Oldiesmann:
I understand about having the same error for everything, but it should at least make sense ("Invalid username or password" or something like that rather than "No username found for that email address")

PR #3142, opened 2015-11-23
Update ManageMaintenance.php

Underdog-01:
- loop to optimize tables caused PHP timeout
- time delay is more aesthetic as opposed to using only set_time_limit()
- $context['optimized_tables'] still used for the template loop


Underdog-01:
Perhaps no one agrees with the way I went about this edit?
If there an issue with this edit please elaborate.

Doesn't anyone else get a PHP timeout when optimizing the database with the original code (before these edits)?
Just using - set_time_limit - works but Imo makes the admin hang without a progress meter.
This edit at least lets the admin know it is in the process of optimizing.

Oldiesmann:
I'm not sure if that code is actually used anymore to be honest with you (unless it's an option in the admin center still). I know we removed the option to auto optimize things a while back for performance reasons or something like that. I'll take another look at it later this afternoon though.

Underdog-01:
Perhaps auto optimization was omitted but manually doing so is still available.
It wasn't working properly though as it would timeout/crash using the default PHP setting.
As I said, just set_time_limit can be used but then it leaves the admin hanging while it optimizes tables.
Also it is not returning the proper amount of tables optimized after the process which also occurs prior to this edit therefore Imo something else is awry within the optimize function itself.

PR #3161, opened 2015-12-03
Fix attachment directory issues...

Oldiesmann:
This should fix most (if not all) of the issues with the $modSettings['attachmentUploadDir'] setting. It forces $modSettings['attachmentUploadDir'] to always be an encoded array regardless of how we're handling attachments.


jdarwood007:
Looks like some white space issues as well.

PR #3186, opened 2015-12-11
Upgrade issues

### LINKED ISSUE #3184: Remaining issues with upgrade
Following #3155 there are still some issues with the upgrade tool.
I've been trying to track this but it's the first time I'm looking into it really.

1 - no sub-template for "Step 6: Convert serialized strings to JSON"
Cause: typo in upgrade.php

```
    $context['sub_template'] = isset($_GET['xml']) ? 'serialize_json_xml' : 'serialize_json';
```

Should be $upcontext. Fixing this brings the template back but it still does nothing

2 - Step 6 doesn't perform any operation, just stays at:

> Completed 0 out of 13 tables.
> Current Table: "background_tasks"

"Continue" also does nothing.

margarett:
Fixes #3184

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -1633,7 +1633,7 @@ function backupTable($table)
 {
 	global $is_debug, $command_line, $db_prefix, $smcFunc;
 
-	if ($is_debug && $command_line)
+	if ($command_line)
 	{
 		echo "\n" . ' +++ Backing up \"' . str_replace($db_prefix, '', $table) . '"...';
 		flush();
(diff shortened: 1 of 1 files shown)


Oldiesmann:
So with this commit everything is working now?

margarett:
As much as I can tell, yes. I noticed that I probably removed too many $is_debug which might show debug info in CLI mode without the --debug parameter though. 
This relates to JSON, please check my comments on the issue and also the private dev board on a possible issue with the UTF8 template ;)

Oldiesmann:
I've never run it with --debug in the shell but it already seems to show everything that would be shown in debug mode when running from a browser so I'm not too concerned with that.

I'll take a look at the rest of the stuff in a bit and see what I can do.

PR #3191, opened 2015-12-13
Do not use DEFAULT on TEXT/BLOB

socram8888:
`upgrade.php` was failing to check permissions, as my version of MySQL apparently won't allow `DEFAULT`s on `TEXT`/`BLOB` colums (see https://stackoverflow.com/questions/3466872/why-cant-a-text-column-have-a-default-value-in-mysql)


Oldiesmann:
You need to sign off on the commit in order for us to use it.

socram8888:
Done

PR #3205, opened 2015-12-21
add "Drop table if exists" to smf_db_table_sql in pg

albertlast:
In the mysql version of the function smf_db_table_sql came a drop before the create statement.

I added the same logic to the pg version of this function.

### DIFF
--- a/Sources/DbExtra-postgresql.php
+++ b/Sources/DbExtra-postgresql.php
@@ -235,6 +235,9 @@ function smf_db_table_sql($tableName)
 	// This will be needed...
 	$crlf = "\r\n";
 
+	// Drop it if it exists.
+	$schema_create = 'DROP TABLE IF EXISTS ' . $tableName . ';' . $crlf . $crlf;
+	
 	// Start the create table...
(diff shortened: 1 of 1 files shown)


XinYenFon:
Please read: https://github.com/SimpleMachines/SMF2.1#how-to-contribute - we cannot accept PRs/Commits w/o sign-off.

albertlast:
Signed-off-by: albertlast

XinYenFon:
it needs to be "commit message".

albertlast:
i did a rebase to add this.

I should notice that IF EXISTS for table and index exists with the pg 8.2 and higher,
this project got current as requirement pg 8.0 and higher.

illori:
i am not sure that is a good enough sign off. the linked to page on our readme says

It is enough to include in your commit comment "Signed-off by: " followed by your name and email address (for example:  Signed-off-by: Angelina Belle angelinabelle1@hotmail.com )

so you should include your email address in the sign off. this would apply to your other PR's as well.

without your email address we really have no way to prove who has given us a sign off from what i understand.

albertlast:
I got my own point of view about this topic,
but I added a mail address to every commit.

PR #3206, opened 2015-12-21
PG 9.5 Ignore Support

albertlast:
again a smale db related pull request.

This added ignore support for the pg 9.5+,
pg 9.5 is current as rc available.
see http://www.postgresql.org/docs/9.5/static/sql-insert.html

Signed-off-by: albertlast

### DIFF
--- a/Sources/DbSearch-postgresql.php
+++ b/Sources/DbSearch-postgresql.php
@@ -30,9 +30,19 @@ function db_search_init()
 			'db_create_word_search' => 'smf_db_create_word_search',
 			'db_support_ignore' => false,
 		);
-		
-		//pg 9.5 got ignore support
-	if (version_compare($smcFunc['db_get_version'],'9.5.0','>='))
+        
+	db_extend();
+	
+	//pg 9.5 got ignore support
+	$version = $smcFunc['db_get_version']();
+	// if we got a Beta Version
+	if (stripos($version, 'beta') !== false)
+	    $version = substr($version, 0, stripos($version, 'beta')).'.0';
+	// or RC
+	if (stripos($version, 'rc') !== false)
+	    $version = substr($version, 0, stripos($version, 'rc')).'.0';
+	
+	if (version_compare($version,'9.5.0','>='))
 		$smcFunc['db_support_ignore'] = true;
 }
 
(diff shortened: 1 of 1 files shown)


albertlast:
Signed-off-by: albertlast

albertlast:
The pr got a bug,
reason db_get_version bring the string back (and had to call db_extend).
String is "9.5rc1" doesn't work with version_comare from php.
I'm attempted to change the function to integer version nummer (server_version_num).
The result would be 90500 (easy to compare),
problem is that this infos are shown in the admin panel there would be the integer than too.

Oldiesmann:
Easiest way around this is to just look for "beta" or "rc" in the string and only look at the first three characters if so (since you'll never have "9.5.1rc1" or anything like that).

Something like this:

```
$version = $smcFunc['db_get_version'];
if (stripos($version, 'beta') !== false || stripos($version, 'rc') !== false)
    $version = substr($version, 0, 3);
```

Sources/DbSearch-postgresql.php
```diff
@@ -91,11 +105,16 @@ function smf_db_search_query($identifier, $db_string, $db_values = array(), $con
 
 	if (isset($replacements[$identifier]))
 		$db_string = preg_replace(array_keys($replacements[$identifier]), array_values($replacements[$identifier]), $db_string);
-	elseif (preg_match('~^\s*INSERT\sIGNORE~i', $db_string) != 0)
+	if (preg_match('~^\s*INSERT\sIGNORE~i', $db_string) != 0)
```
albertlast, line 108: I don't understand the intention,
why here is a elseif used.

I got some error around this,
so i replace this with a if.

Oldiesmann, line 108: What error did you get? The "elseif" is used because we want to do that if $replacements[$identifier] isn't set.

albertlast, line 108: The query insert_into_log_messages_fulltext goes wrong,
it's take only the if part but it need to take the elseif part too.

live627, line 108: I don't know either, but it looks like an error, from reading the code..

Sources/DbSearch-postgresql.php
```diff
@@ -30,6 +30,20 @@ function db_search_init()
 			'db_create_word_search' => 'smf_db_create_word_search',
 			'db_support_ignore' => false,
 		);
+        
+	db_extend();
+	
+	//pg 9.5 got ignore support
+	$version = $smcFunc['db_get_version']();
+	// if we got a Beta Version
+	if (stripos($version, 'beta') !== false)
```
live627, line 39: Pleas use a tab instead of spaces for indenting here.

albertlast:
Fixed the tab and allman style
about the elseif thing i got no solution because the query insert_into_log_messages_fulltext
needs both

PR #3207, opened 2015-12-22
PG Install Fix

albertlast:
$has_innodb was not setted when we came from pg
the privilegs check works only with mysql

Signed-off-by: albertlast


other/install.php
```diff
@@ -1081,6 +1081,8 @@ function DatabasePopulation()
 			$replaces['START TRANSACTION;'] = '';
 			$replaces['COMMIT;'] = '';
 		}
+	} else {
```
live627: Use Allman style, please.

albertlast:
done

live627:
thanks

PR #3208, opened 2015-12-22
Fix Pattern in smf_db_search_query

albertlast:
Dunno why this is needed,
but I got error so long this was not right.


Oldiesmann:
How did my commits get attached to this?

albertlast:
I created a new one :)

PR #3209, opened 2015-12-22
Pg fulltext search

albertlast:
This is a try to implement the pg fulltext search.

Three kind are included in this pr.
On the one hand the big changes about fulltext,
than a fix for the calculation of the table and index size and
fix of the pattern search.

The pattern fix got an own pr.

And I do rebase to far....

One exisiting "problem" is that I dont know,
how to access the setting.php to get the forum language.
Would help to optimize the search, current is english hardcoded on the place where index is created.


albertlast:
today define the default language to simple,
reason:
I know how to access the setting parameter,
but it can happen that pg version doesn't got this language,
for this case i would need to place a backup.
In SearchAPI-Fulltext.php I would need this infos if orginal language is used or backup.

For this logic i got current no idea how to solve this.

Infos about simple:
Simple doesn't ignore stopwords and doesn't try to find the stem of the word. With simple every group of characters separated by a space is a lexeme; the simple text search config is pratical for data like a person's name for which we may not want to find the stem of the word.
src: http://rachbelaid.com/postgres-full-text-search-is-good-enough/

Sources/DbSearch-postgresql.php
```diff
@@ -40,7 +40,7 @@ function db_search_init()
  */
 function smf_db_search_support($search_type)
 {
-	$supported_types = array('custom');
+        $supported_types = array('fulltext');
```
Oldiesmann, line 43: You don't want to remove 'custom' from that array. Doing so will prevent the "custom" search index from working on new installs and will cause all kinds of problems for anyone upgrading.

Sources/SearchAPI-Fulltext.php
```diff
@@ -215,14 +218,26 @@ public function indexedWordQuery(array $words, array $search_data)
...
-			foreach ($words['indexed_words'] as $fulltextWord)
+                        
+                        if($smcFunc['db_title'] == "PostgreSQL"){
+                            $row = 0;
+                            foreach ($words['indexed_words'] as $fulltextWord) {
+                                $query_params['boolean_match'] .= ($row <> 0 ? '&' : '');
+                                $query_params['boolean_match'] .= (in_array($fulltextWord, $query_params['excluded_index_words']) ? '!' : '') . $fulltextWord . ' ';
+                                $row++;
+                            } 
+                        }else
```
Oldiesmann, line 229: The "else" here needs to be on its own line.

Oldiesmann:
A couple of things here (in addition to my comments on specific lines)...

First, how did my commits get attached to this? Those changes have already been merged in and I don't know what will happen if we try to merge them in again.

Second, there's a way we can detect which languages are available to use, right? If so, we should be able to determine whether fulltext support is available based on that and the db version.

albertlast:
Created a new one.

Issue #273, opened 2013-01-09
PostgreSQL Native Fulltext Search


Oldiesmann:
It might be a bit late in the development to implement this, but it's definitely something we should consider. As of version 9.0, PostgreSQL has native support for full text searching.

From the original entry on Mantis ( http://dev.simplemachines.org/mantis/view.php?id=4657 ) :

> On a moderately large forum, the custom search index does not work well -- it literally takes weeks
> to construct, and the resulting index is an order of magnitude larger than the original messages,
> and takes so long to search that just doing a brute search can have better performance. It would 
> greatly benefit Postgresql database users to have a native full text search index (documentation
> here http://www.postgresql.org/docs/9.0/static/textsearch.html)

emanuele45:
What would be needed?
Complete PSQL noob here...I know how to set it up and use it a bit for debugging, but not much more.

Oldiesmann:
We'd simply need to create another file containing the appropriate queries for the fulltext search (like we have one for fulltext in MySQL), then modify the SMF code (if necessary) to detect whether or not it's possible. The documentation linked to above gives some good examples about setting it up and searching, though it's still somewhat complicated.

jdarwood007:
Oldies,
Would this be able to make any beta of 2.1?  Your the native PostgreSQL user and know the most about it, so implanting it would be the easiest for you to do.

Oldiesmann:
I've looked at the manual info on fulltext searching and don't have a 
cluewhere to start. Postgres' implementation is much different than 
MySQL's. However, Tinoestsaid he's worked with it and is going to look 
into it for us.
Michael "Oldiesmann" Eshom
Christian Oldies Fan
Cincinnati, Ohio

tinoest:
I don't run a forum large enough to test the performance benefits of switching to this searching , however do we wish to create your own index's and if so would you prefer a GiST or a GIN? The following link explains it better than I could http://www.postgresql.org/docs/8.4/static/textsearch-indexes.html , the other option is just using the default full text engine and not creating any index as an addition to that.

Oldiesmann:
Based on what's outlined there, I think a GiN index would be better, particularly due to GiST producing false positives.

Arantor:
This would also mean bumping the minimum PGSQL version from 8.0 to 9.0, though I presume that's not a problem here?

I am concerned at the amount of work this would generate this late in the 2.1 day.

tinoest:
[...]

PR #3212, opened 2015-12-23
Pg fulltext search 2

albertlast:
A Clean version of https://github.com/SimpleMachines/SMF2.1/pull/3209
with the custom support and "else" in a own line.

Missing is the use of default language(default_text_search_config) of the db user.
More complicate solution would be to give the admin the choice,
based on the table pg_ts_config

related issue: https://github.com/SimpleMachines/SMF2.1/issues/273

### DIFF
--- a/Sources/ManageSearch.php
+++ b/Sources/ManageSearch.php
@@ -223,53 +223,54 @@ function EditSearchMethod()
 		checkSession('get');
 		validateToken('admin-msm', 'get');
 
-	    if ($db_type == 'postgresql') {
-		$smcFunc['db_query']('', '
-			DROP INDEX IF EXISTS {db_prefix}messages_ftx',
-			array(
-			    'db_error_skip' => true,
-			)
-		);
-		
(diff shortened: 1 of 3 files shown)


albertlast:
From technical point of view I think the topic is done.

Some helpful query,
to see the language
show default_text_search_config;

to change the language
ALTER USER <username> SET default_text_search_config = <language>;

available language
select cfgname from pg_ts_config;

Sources/ManageSearch.php
```diff
@@ -223,6 +223,37 @@ function EditSearchMethod()
...
+			SHOW default_text_search_config',
+			array(
+			    
+			)
+		);
+		
+		if ($request !== false && $smcFunc['db_num_rows']($request) == 1)
+		{
+		    $row = $smcFunc['db_fetch_assoc']($request);
+		    $language_ftx = $row['default_text_search_config'];
```
live627: Please use tabs for indenting all the code.

albertlast:
done

PR #3215, opened 2015-12-24
Index for case insensitve search on member_name and real_name

albertlast:
I noticed some query with LOWER(member_name) or LOWER(real_name),
i think this table can be get big, also created some index.


matthew-kerle:
Sweet.

Oldiesmann:
There are now merge conflicts, so you need to fix those before we can merge this in.

albertlast:
done

PR #3218, opened 2015-12-25
Unlogged table in pg

albertlast:
I noticed some table which are from type like caching,
exp. session or log_online table

These are brilliant for unlogged table type,
This kind is faster to fill,
but they lost there complete data when the pg-server get crashed.
more information: http://michael.otacoo.com/postgresql-2/unlogged-table-performance-in-postgresql-9-1/

This feature works only with pg 9.1+, but I placed a version check so this will
no harm the min pg requirement.

Give me feedback if you want go this way.
When yes than i will implement a upgrade script and
give me a hint which table could be unlogged too.

### DIFF
--- a/other/install.php
+++ b/other/install.php
@@ -1083,6 +1083,25 @@ function DatabasePopulation()
 		}
 	}
 
+	// PostgreSQL-specific stuff - unlogged table
+	if ($db_type == 'postgresql')
+	{
+		$result = $smcFunc['db_query']('', '
+			SHOW server_version_num'
+		);
+		if ($result !== false)
+		{
+			while ($row = $smcFunc['db_fetch_assoc']($result))
+				$pg_version = $row['server_version_num'];
+			$smcFunc['db_free_result']($result);
+		}
+		
+		if(isset($pg_version) && $pg_version >= 90100)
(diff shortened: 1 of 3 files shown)


margarett:
If you allow your "sessions" table to die, your users will need to login again even if they chose login "Forever".
You are only looking at log_floodcontrol and log_online (you can see which tables are MEMORY tables in MySQL's installer :wink: )

albertlast:
Like the blog post sayed, unlogged table are designe for table how does store session data.

This table only "dies" when the pg server is shutdown in unnormal way,
not like the memory table in mysql which are empty after every mysql server restart,
for this special situation it's from my point of view okay to login again.

But when wanted i can undo the session table.

margarett:
I understand that ;-)

Either by server restart or crash (highly unlikely but still...) you shouldn't kill your stored sessions, IMO. That's why I compared it with MEMORY ;-)

live627:
This is feedback.

albertlast:
Need I to do something?

live627:
I can merge this if you're done here.

albertlast:
I implement the upgrade path and
test the raw sql directly vs database.

But I see no possibilty to test my upgrade logic directly in the upgrade_2-1_postgresql.sql

PR #3220, opened 2015-12-25
Static get db version

albertlast:
I think it's not needed to call everytime the query behind the db get version.

So this is a offer to prohibit this behavior.


MissAllSunday:
Thanks!

MissAllSunday:
Ugh, I missed it but can you please put the return on its own line?  SMF uses an Allman style and will be better to preserve that as much as possible, thanks.

Issue #2968, opened 2015-08-09
constructBanQueryIp() will return wrong query


Elsensee:
I'm talking about this: https://github.com/SimpleMachines/SMF2.1/blob/939347d4bfdba0194744996d34f9d29abe6544ec/Sources/Security.php#L1273

Let's say you got the range 123.241.98.29-129.23.53.100 in the database and the IP you want to check against that is: 124.10.89.211
The comparison for block1 will be: `124 BETWEEN 123 AND 129` which is true, of course.
For block2 it will be: `10 BETWEEN 241 AND 23` which is false, although the IP is in that range.

Just stumbled across this issue when trying to test the performance of various solutions for IP banning and just wanted to see if you got the same issue. ;)

joshuaadickerson:
This issue comes up every now and then. The issue with fixing it is pretty much just resources.

Besides documentation, which I think is needed, I think the next easiest step is to create rows in the database for the range. So, if your first block is 123 and the last is 129, you would have 7 rows dedicated to this range. The first would have all of its "low"s matching the start and the high would be 255. The last would have all of its "high"s matching the end and the low would be 0. Every row between those (the 5 others) would be 0-255 for all parts. If you have something like 123.23.53.100 - 123.241.98.29 then all of the rows would have the first part's high and low equal to 123.

This is a hack but it works well. It only adds a few rows per range. I would say to make it less hacky the ban system should make it a group.

Not necessarily directed at you, @Elsensee, but if you have thousands of bans, you probably should look in to what you are doing. If you are trying to block a range, you should talk to your host and get it blocked at a lower level in your stack. If you are trying to block someone that keeps on evading your bans, what makes you think adding more IPs is going to stop them? Once again, maybe get your host involved. There are lists of public proxies out there. You could have your host ban those. There are services that do this too. 

If it is spammers I would recommend making it harder for them. Check their IP extensively in the registration process. If it is one of a known spammer from any of the SPAM lists out there, make them jump through hoops to register or just don't let them at all. If they do register, make their posts require some sort of human verification and then make them approval only. Make it cost them time which equals money. For the first 10 unapproved posts, make them take 1 minute between each. For the next 10 make it 30 seconds. That is 15 minutes that it takes for them to post. They won't want to waste that much time on your site.

Sorry for the OT discussion. I should bring that to the forum.

wintstar:
//OT// Only times for information. Elsensee is a developer in the phpBB forum system.

[...]

PR #3222, opened 2015-12-26
Change ban IP address field 

albertlast:
Man this is a mess.

i don't know if this pr is helpfull,
but it give a good overview where all problem raise up...

I implemented the pg and mysql support,
but i'm not able to test mysql.

One of the problems are, that pg take the ip address als "string" and mysql need it as integer,
reason for 'ip_address'.
Maybe would work for both string...

related to issue: https://github.com/SimpleMachines/SMF2.1/issues/2968

technical done,
~~upgrade is missing~~


Oldiesmann:
I'm not sure how this is a good solution compared to how we're currently handling things. Also, this would require that we force people to have PHP compiled with bcmath (which is optional) and we'd need to do even more work in the upgrade to convert the existing IP bans, which could take quite some time on larger forums.

albertlast:
The big pro for this kind of implementation is
"which could take quite some time on larger forums" here we can now use index.

The code is less complex.

The most code came from fallback for mysql and
the wildcard implementation, which not exist current.

There is no need to distinguish of ipv4 and 6.

Edit:
From my point of view the biggest disadvantages of the old solution ist,
that there no safe way if the ip address is are ipv4 or ipv6.
127|0|0|1|0|0|0|0 could be a ipv4 but 127:0:0:1:0:0:0:0 is a valid ipv6 too.

jdarwood007:
The bcmath thing is important since its optional.

@Oldiesmann,  Upgrading shouldn't be too difficult.  We have the upgrader add the low and high columns and just call a massive update query.
Something like such:

```
UPDATE ban_items
SET ip_low = IF(ip_low5 > 0, CONCAT(ip_low1, ':', ip_low2, ':', ip_low3, ':', ip_low4, ':', ip_low5, ':', ip_low6, ':', ip_low7, ':', ip_low8), CONCAT(ip_low1, '.', ip_low2, '.', ip_low3, '.', ip_low4))
WHERE ip_low1 > 0
```

Adding the where into there would be ok so we can still do it in 1,000 item batches.

albertlast:
well it's importent to know that the ip_low and ip_high are varbinary in mysql,
so a function update sql would be:

``` sql
UPDATE ban_items
SET ip_low = IF(ip_low5 > 0, 
                INET6_ATON(
                  CONCAT(ip_low1, ':', ip_low2, ':', ip_low3, ':', ip_low4, ':', ip_low5, ':', ip_low6, ':', ip_low7, ':', ip_low8)
                  )
                  , INET6_ATON(
                    CONCAT(ip_low1, '.', ip_low2, '.', ip_low3, '.', ip_low4)
                  )
               )
WHERE ip_low is null
```

But this only works on mysql version where the function INET6_ATON exists,
the version where you can find this is >5.6.2, mariadb >10.0.12 or in any mysql where this udf is exists:
https://bitbucket.org/skion/mysql-udf-ipv6

on every other version you need to load this data in php and use the fallback functions to convert this in right format.
Here a link to my test case for mysql: http://sqlfiddle.com/#!9/7e4cd/1/0

PG upgrade:

``` sql
UPDATE smf_ban_items
SET ip_low = (
    case when ip_low5 > 0 then CONCAT(ip_low1, ':', ip_low2, ':', ip_low3, ':', ip_low4, ':', ip_low5, ':', ip_low6, ':', ip_low7, ':', ip_low8) 
    else CONCAT(ip_low1, '.', ip_low2, '.', ip_low3, '.', ip_low4) 
    END
    )::inet
WHERE ip_low is null
```

```
```

jdarwood007:
Is the IPv4 version also using INET6_ATON?  We could make it a CASE usage in MySQL to limit our need to specifics, but I see we would have to use ::net at the end, which MySQL doesn't do.

albertlast:
This "::net" only works in postgresql.

For mysql it's essential that the INET6_ATON from mysql is used for ipv4 and ipv6 or
the fallback methode from Security.php inet_ptod.

I should mention that definition of "-" is changed in this pr.

It's look like in the current develop version is the - used to define a range of a block of the ip address.
In this pr is the - use to splitt the lower and the upper bound of the whole ip address.
current: 192.168.0-10.\* is internal 192.168.0.0 - 192.168.10.255
pr: 192.168.*-192.168.10.\* is internal 192.168.0.0 - 192.168.10.255

I already mention that i finde the current solution to complex internaly and for the user the benefit is smale or negative when he use ipv6 (with the short version of 0000:0000 -> ::)

albertlast:
For me was on to many places a if for postgresql or not.
So i decided to add a new datatype "inet" to Subs-Db-\* in which the whole magic is done.

This pr use the function to write null values to the database,
to do so is the pr needed: https://github.com/SimpleMachines/SMF2.1/pull/3241

Oldiesmann:
I still don't like that this uses the BCMath functions since that's an optional extension. Do we really need to use those instead of regular math functions?

albertlast:
For me is important that we got the same result as the nativ function of mysql,
to keep flexible for the future.

But I'm not deep enough in php to get another way as this which I copy from stackoverflow.

The intention here to use bcmath I think is that ipv6 got a 128 bit integer.

albertlast:
I don't believe it,
but it's look like the solution could be very easy for mysql.

See the last commit, would be good if some there with mysql and can test this.

Added {nullVar} as null string identifier.

Here the mysql upgrade command

``` sql
UPDATE ban_items
SET ip_low = case when ip_low8 > 0 THEN
    UNHEX(
        CONCAT(LPAD(hex(ip_low1),4,'0'),LPAD(hex(ip_low2),4,'0'),LPAD(hex(ip_low3),4,'0'),
        LPAD(hex(ip_low4),4,'0'),LPAD(hex(ip_low5),4,'0'),LPAD(hex(ip_low6),4,'0'),LPAD(hex(ip_low7),4,'0'),LPAD(hex(ip_low8),4,'0'))
    )
ELSE
    unhex(
        hex(
            INET_ATON(concat(ip_low1,'.',ip_low2,'.',ip_low3,'.',ip_low4))
        )
    )
END,
ip_high = case when ip_high8 > 0 THEN
    UNHEX(
        CONCAT(LPAD(hex(ip_high1),4,'0'),LPAD(hex(ip_high2),4,'0'),LPAD(hex(ip_high3),4,'0'),
        LPAD(hex(ip_high4),4,'0'),LPAD(hex(ip_high5),4,'0'),LPAD(hex(ip_high6),4,'0'),LPAD(hex(ip_high7),4,'0'),LPAD(hex(ip_high8),4,'0'))
    )
ELSE
    unhex(
        hex(
            INET_ATON(concat(ip_high1,'.',ip_high2,'.',ip_high3,'.',ip_high4))
        )
    )
END
```

Should again someone test with mysql.

albertlast:
- Implemented db driven upgrade path
- found a bug in existing solution that the smallint on pg is to small for FFFF::

Sources/Subs.php
```diff
@@ -3433,53 +3433,73 @@ function getAttachmentFilename($filename, $attachment_id, $dir = null, $new = fa
...
+			{
+				$ip_parts[1] .= $mode . $max;
+				$valid_high = inet_pton($ip_parts[1]);
+				$count++;
+				if ($count > 9) break;
+			}
+		}
+	    
+	    if($valid_high && $valid_low)
+	    {
```
live627: I see mixed spaces and tabs. Please use only tabs to indent the code.

other/upgrade_2-1_mysql.sql
```diff
@@ -1908,4 +1880,28 @@ DROP INDEX idx_topic on {$db_prefix}messages;
...
+--- update ban ip with ipv6 support
+/******************************************************************************/
+
+ALTER TABLE {$db_prefix}ban_items ADD COLUMN ip_low varbinary(16);
+ALTER TABLE {$db_prefix}ban_items ADD COLUMN ip_high varbinary(16);
+
+UPDATE {$db_prefix}ban_items
+SET ip_low = 
+    UNHEX(
+        hex(
```
live627, line 1895: Why hex and then unhex? Am I missing something?

albertlast, line 1895: I don't know another way to convert this in varbinary.
the hex convert the int to hex,
the unhex convert this to varbinary.

live627:
I really want to merge this in. Anything left that needs doing here?

albertlast:
From my point of view I didn't see anything to do.

live627:
I'm going to merge this in now... because I can, I suppose?

/me waits for the fallout

PR #3227, opened 2015-12-30
Fix Translation exp. for IP ban address

albertlast:
Set a point after the 19 to get a valid ipv4.
Reason for 19. for ones that lower as the 20. which already in high part and
it's less risk that he ban his self,
when he copy the exp.


Themes/default/languages/Admin.english.php
```diff
@@ -153,7 +153,7 @@
 $txt['ban_description'] = 'Here you can ban troublesome people either by IP, hostname, username, or email.';
 $txt['ban_add_new'] = 'Add new ban';
 $txt['ban_banned_entity'] = 'Banned entity';
-$txt['ban_on_ip'] = 'Ban on IP (e.g. 192.168.10-20.*)';
+$txt['ban_on_ip'] = 'Ban on IP (e.g. 19.2.168.10-20.*)';
```
XinYenFon, line 156: I think it would be better to give example for each (IPv4 : 192.168.x not like this, even i'm hardly understanding this.

albertlast, line 156: My intention is to fix to invalid ip which is shown,
when you like to change more feel free to make an own pr with this change.

tinoest, line 156: Your pull request results in a invalid IPv4 address being displayed though? You'd need to remove the .\* to make it valid.

albertlast, line 156: My interpretation of 19.2.168.10-20.\* that the result is
19.2.168.10 is the lower bound of ip address and
20.255.255.255 the upper bound of the ip address range.

Another interpretation of this could be lower bound 19.2.168.10 and upper 19.2.168.20.

When the last one is the expected behavior i would recommand to change this.

The reading of this is to far complex
worst case (in ipv6): 0011-ffff:0031-ffff:0000-0022:0000-0011:0011-*:0031-*:0011-*:0000-*
my version: 0011:0031::0011:0031:0011:0000-ffff:ffff:0022:0011:*

advantage on both side of the - we got valid ip or not --> less complex logic
I should mention that this logic is implemented in my pr of https://github.com/SimpleMachines/SMF2.1/pull/3222

tinoest:
That's not clear to me, having the following would be clearer;

19.2.168.10 - 20.\* . *. \* 

I read the original as 192.168.10.\* - 192.168.20.\* if you expand it out. 

I can understand the technical reasons behind your comment, just don't think it's clear what your trying to convey.

albertlast:
Both solutions:
19.2.168.10 - 20.*
19.2.168.10 - 20.*.*.*
works in the implementation from me.

But then I would say this pr is not needed. Because I misinterpret the exp.

jdarwood007:
The interpretation of that line is 192.168.10-20.*
Which is 192.168.10.1 to 192.168.20.255

No valid cidr for that, close would be a /21 from 192.168.8.0 to 192.168.15.255

PR #3229, opened 2015-12-31
update isValidIPv6 function

albertlast:
The preg_match which is used it's me to complex.

So I update this function with nativ php valid check.


albertlast:
yeah you are right

i did a smale performance test and the result is,
that the nativ version is 25% faster as the preg_match version (php 7.0.1).

``` php
$var = '::1';
$var2 = false;
$time_start = microtime_float();
for($i=0;$i<100000;$i++)
    $var2 = isValidIPv6($var);

$time_end = microtime_float();
```

Oldiesmann:
If there are errors with the regex, we should fix those instead of just assuming any string with a colon in it is a valid IPv6 address.

Sources/QueryString.php
```diff
@@ -330,10 +330,11 @@ function cleanRequest()
  */
 function isValidIPv6($ip)
 {
-	if (preg_match('~^((([0-9A-Fa-f]{1,4}:){7}[0-9A-Fa-f]{1,4})|(([0-9A-Fa-f]{1,4}:){6}:[0-9A-Fa-f]{1,4})|(([0-9A-Fa-f]{1,4}:){5}:([0-9A-Fa-f]{1,4}:)?[0-9A-Fa-f]{1,4})|(([0-9A-Fa-f]{1,4}:){4}:([0-9A-Fa-f]{1,4}:){0,2}[0-9A-Fa-f]{1,4})|(([0-9A-Fa-f]{1,4}:){3}:([0-9A-Fa-f]{1,4}:){0,3}[0-9A-Fa-f]{1,4})|(([0-9A-Fa-f]{1,4}:){2}:([0-9A-Fa-f]{1,4}:){0,4}[0-9A-Fa-f]{1,4})|(([0-9A-Fa-f]{1,4}:){6}((\b((25[0-5])|(1\d{2})|(2[0-4]\d)|(\d{1,2}))\b)\.){3}(\b((25[0-5])|(1\d{2})|(2[0-4]\d)|(\d{1,2}))\b))|(([0-9A-Fa-f]{1,4}:){0,5}:((\b((25[0-5])|(1\d{2})|(2[0-4]\d)|(\d{1,2}))\b)\.){3}(\b((25[0-5])|(1\d{2})|(2[0-4]\d)|(\d{1,2}))\b))|(::([0-9A-Fa-f]{1,4}:){0,5}((\b((25[0-5])|(1\d{2})|(2[0-4]\d)|(\d{1,2}))\b)\.){3}(\b((25[0-5])|(1\d{2})|(2[0-4]\d)|(\d{1,2}))\b))|([0-9A-Fa-f]{1,4}::([0-9A-Fa-f]{1,4}:){0,5}[0-9A-Fa-f]{1,4})|(::([0-9A-Fa-f]{1,4}:){0,6}[0-9A-Fa-f]{1,4})|(([0-9A-Fa-f]{1,4}:){1,7}:))$~', $ip) === 0)
-		return false;
-
-	return true;
+	//looking for :
+	if (strpos($ip , ':') === false ) return false;
```
live627: Please put the return in its own line.

albertlast:
done

PR #3230, opened 2016-01-01
PHP 7 Session Handling

albertlast:
I played today with php 7 and noticed that php 7 got a stricter version of customer session handling,
see: http://stackoverflow.com/questions/34117651/php7-symfony-2-8-failed-to-write-session-data

So I modified some session function that they return true or false.
Based on the information if the db action was successful or not.

The intial error was: http://snag.gy/hDZPk.jpg

Error-msg in log:
2: Unknown: Session callback expects true/false return value

2: Unknown: Failed to write session data (user). Please verify that the current setting of session.save_path is correct (D:\Programme\xampp\tmp)

### DIFF
--- a/Sources/Session.php
+++ b/Sources/Session.php
@@ -177,7 +177,7 @@ function sessionWrite($session_id, $data)
 			array('session_id')
 		);
 
-	return $result;
+	return ($smcFunc['db_affected_rows']() == 0 ? false : true);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


Dragooon:
I'm guessing this is backwards compatible until 5.2.7?

albertlast:
I'm dunno.
But the target of smf 2.1 is php 5.3+

PR #3237, opened 2016-01-05
Add a db engine to our support information

jdarwood007:

This is so we can see things like Percona and MariaDB in the support information.  At this point the purpose of this is only for support usage incase ever in the future we have Percona, MariaDB or MySQL based issues.


jdarwood007:
I didn't know if PostgreSQL had forks that we support.  So if we do, a separate commit should be made to properly add those in as I can't test it very well.

PR #3239, opened 2016-01-05
Logging.php FixEmptyJson

albertlast:
I notice that the url field in log_online is a json field and
a valid form for empty is {}.

But this solution is not best(but the easiest) solution from my point of view.
The best solution would be to change the database field nullable and write null into this field.

When wanted I can build the other solution too.

### DIFF
--- a/Sources/Logging.php
+++ b/Sources/Logging.php
@@ -62,7 +62,7 @@ function writeLog($force = false)
 		$serialized = json_encode($serialized);
 	}
 	else
-		$serialized = '';
+		$serialized = '{}';
 
 	// Guests use 0, members use their session ID.
 	$session_id = $user_info['is_guest'] ? 'ip' . $user_info['ip'] : session_id();


MissAllSunday:
I will leave it as it is to be honest, an empty value means we don't need to decode it only to end up with an empty array, to me its much easier to do this:

$var = !empty($var) ? json_decode($var, true) : array();

The empty string var makes it uniform with how SMF treats vars.  SMF takes advantage of PHP's dynamic typing, so every var in SMF its an empty string unless told otherwise.

The same I told you on your null PR applies here, SMF doesn't really use null values in its DB fields.

albertlast:
The bigger picture is here to use later one, the json functions from the database (see mysql 5.7 and pg 9.2),
and like in the other pr would be nullable field do the same.
But here we got the possible to choose between two possibles ways: use {} or null

MissAllSunday:
We have a minimum versions for mysql (5.0.3) and postgresql (8.0), we need to stick with what those min versions can offer, we shouldn't code for specific versions as that will be quite problematic to support.

albertlast:
This is a good point which I like to address:
- the minimum of postgresql should be 8.2 reason: https://github.com/SimpleMachines/SMF2.1/pull/3205
- When you realy plan to implement full text search in pg you need 8.3.
- All version below 9.1 are no more supported http://www.postgresql.org/support/versioning/ ,
  so I would recommend to set the minimum to 9.1.
- Your pg user base is smaler as the mysql user base to there is no need to get the time frame as support time like the mysql
- Min php version is 5.3(released 2009) so it reasonable to take the postgresql version from this time 8.4
- I got some pr which need pg 9.2(logged table) or 9.5(replace/insert ignore) there I implement a version check to not raise up the min pg version
- features like the json support of the database is a big advantage when you use it
  - Exp. the native search directly in the database without string operator

MissAllSunday:
We don't set min versions based on the vendors support for it. We set min  
versions based on mainstream usage and our own internal goals and needs.

Both the mysql and psgre versions must be similar, we can't realistically  
support cutting edge postgre if there are no feature equivalent in mysql  
and vice-versa.

Check versions aren't really a good idea, it creates a maintenance hell  
and the software already suffers on that end, we def don't want to add  
more burden to it, so any PR that relies on a check version must likely  
won't be merged in.

albertlast:
And how many user use pg 8.0? From where you get the information that the mainstream is 8.0?
How many user you got how use pg in general?

With pg 8.3 was fulltext search added and with pg 9.5 pg and mysql again get similar (replace and insert ingore was added) so based on this we would need to target pg 9.5. 
So what is "similar"?

One of your internals goal should be to be fast as board and
this you can archive by using "bleeding edge" feature.

One of the other target would be to got a clean database schema which common include nullable fields.

Oldiesmann:
At this point I think we can safely drop support for anything older than PostgreSQL 8.3, but we really can't go too much higher than that as 8.4 is the latest version available in the official repos for CentOS 6, which is still widely used. You can of course manually install newer versions, but doing so is a bit of a pain, especially if PostgreSQL is used in production environments (since that would mean anything using a PostgreSQL database wouldn't work until the upgrade and data migration were complete).

I also agree that we need to keep version checks to a minimum, as it gets too confusing otherwise.

MissAllSunday:
> One of your internals goal should be to be fast as board and
> this you can archive by using "bleeding edge" feature.

Sorry but nope!  Our internal goals are set to provide a usable (as in being able to run it under the vast majority of server configurations out there) and fast software, we need to find balance between those two main factors.

We can't happily use "cutting edge" software. Sure, newer versions usually means improved and better code and thus we should be using them  but theres a lot of factors we need to take into account when deciding our minimum versions.

albertlast:
The factors are not transparent,
this is the reason that i can't follow you definition of min version.

Like what are the majority of server configuration? I know that you alread mention in this pr, but from where and how many?

To came back to the topic, why no use of null here? would be perfect of the one side the empty got the same behavior by null value as by '' and other side the database infos are right.
To remember '' is no valid json, but null is okay.

MissAllSunday:
I honestly don't know how to explain this better to you, every major php software has its own min versions specified, we have ours too, we can't happily raise our min versions everytime we want, thats not how this works.

For server statistics/usage and related data, theres a few websites that offers that kind of info, from server OS to PHP versios as well as DB engine usage as well, you can get a clearer idea why we don't use "cutting edge software".

As for null values, I still haven't see any valid point, I'm sorry. null isn't the same as an empty string, we use an empty string plus dynamic typing.

If null values are going to be introduced then theres gotta be a very important reason to do so and I still haven't see any.

albertlast:
One of the imporant reason is, that smf ist not able to decide if the banned ip address is ipv4 or ipv6.
When you ban the ipv6 address 127:0:0:1:0:0:0:0 smf will convert this to ipv4 -> 127.0.0.1

Okay then I close this pr (it's your problem not my).

MissAllSunday:
There is no problem,  that's what I've been trying to explain to you. 

If there are issues with ips then it should be solved for that specific issue. Trying to apply a fix that can potentially broke other things it's not really a good fix. 

While I do appreciate your initiative, your solution is kinda drastic, to make a reference, you are trying to cut off an entire arm just to fix a broken finger.

If the real issue is SMF not handling IPs correctly then I'm pretty sure we can fix that without messing with the rest of the code and without modifying our min versions.

If there is absolutely no other alternative then yes, lets push our min versions but thats the very, very last option, let us try a few other alternatives first :)

Issue #3226, opened 2015-12-30
Ban IP entry


albertlast:
By the develop of the "new" version ban fields(https://github.com/SimpleMachines/SMF2.1/pull/3222) I noticed some errors which I fixed in my pr,
but the future for this pr is unknown that is the reason that I want to create this list of issues of
existing solution:
- Wildcard ban on high doesn't work
  -- exp. 19.168.10-20.*
- Short form for ipv6 doesn't work
  -- exp. ::1
- ipv6 entry will converted (unwanted) to ipv4
  -- exp. 20:20:0:1:0:0:0:0
- Wildcard on single ipv6 doesn't work (with ipv4 no problem)
  -- exp. 20:20:*

A solution for the unwanted convertig could be the fields ip_low5-8 and ip_high5-8 change to nullable fields.

PR #3241, opened 2016-01-05
Allow Null Value to write to Database

albertlast:
Null Fields are importent tool in the database,
so we should allow to use them.

### DIFF
--- a/Sources/Class-Null.php
+++ b/Sources/Class-Null.php
@@ -1,0 +1,38 @@
+<?php
+
+/**
+ * Classes used for reading gif files (in case PHP's GD doesn't provide the
+ * proper gif-functions).
+ *
+ * Gif Util copyright 2003 by Yamasoft (S/C). All rights reserved.
+ * Do not remove this portion of the header, or use these functions except
+ * from the original author. To get it, please navigate to:
+ * http://www.yamasoft.com/php-gif.zip
+ *
+ * Simple Machines Forum (SMF)
+ *
+ * @package SMF
+ * @author Simple Machines http://www.simplemachines.org
+ * @copyright 2015 Simple Machines and individual contributors
+ * @license http://www.simplemachines.org/about/smf/license.php BSD
+ *
+ * @version 2.1 Beta 2
+ */
+
+if (!defined('SMF'))
+	die('No direct access...');
+
+/**
+ * Class nullClass
+ *
+ * An implementation for nullValues to the database
+ */
+
+class nullClass{
+    
+    public function getValue()
+    {
+	return null;
+    }
+}
+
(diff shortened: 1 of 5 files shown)


MissAllSunday:
We shouldn't really comment out code, a better solution would be to create a new token and compare the match against it, something like {null} to tell SMF that value must be inserted as null.

albertlast:
And how you know that the token came from programm and not from intput of the user?
Personaly I was thinking to create a null class as token.

MissAllSunday:
Uh, nope! not  talking about user inputs, SMF uses widlcards to know which  
type the content should be treated as, for example, int, string, literal,  
etc

So if a mod uses a null wildcard {null:var} then SMF should save that as a  
null.

I honestly don't see any real benefit of adding a null type wildcard,  
historically SMF doesn't use null values and I can see some issues in the  
way SMF retrieves modSettings values, theres also the way PHP handles null  
values, many mod authors aren't really aware of the difference between a  
var been null, been empty and been set.

albertlast:
In my special case this problem came from ip address and
I implement a own type (inet) this need to be nullable otherwise exist no way realize afterwards
if the user put a ip address in or not (the database field is not able to safe 0 or '').

One of problem the missing of nullable field can you finde here:
ipv6 entry will converted (unwanted) to ipv4 -- exp. 20:20:0:1:0:0:0:0
https://github.com/SimpleMachines/SMF2.1/issues/3226
Normaly you would null all field which are not use,
so you had a easy token to see if the row is a ipv4 or ipv6 entry.
But here it's missing and you had no chance to say if the ip address is a ipv4 or ipv6 address.

MissAllSunday:
Then I suggest to look for an alternative for that specific case only or add a null wildcard but comment out code its not really a solution.

albertlast:
I added a nullClass and reactivate the checks,
for on my "ban ip address" pr this works good.

The field got the values = new nullClass(); and later i check in inet if the $replacement got this object.

MissAllSunday:
Adding a full class and file everywhere isn't necessary.

What I meant by adding a wildcard was adding a new case inside the switch ($matches[1])  in each DB file

There are cases for int, string, array_string, array_int, date, float, identifier and raw

If you really need to add a null value to the DB then add a new case on that switch to exclusively handle null cases.

Or you could use the already available {raw} wildcard, that wildcard will literally pass whatever you want (including null values) to the DB.

So, really, there is no need to add an entire new class.

albertlast:
The itention of this pr is to define a token which i can use in the other pr in matches case part and there is the handling of the special case of null already placed.

But I can't use null directly because the

```
if (!isset($values[$matches[2]]))
```

came befor the matches case part.

Oldiesmann:
You don't have to use "null" directly. You can use something like "nullval" to get around that particular issue.

albertlast:
Yeah currently I use the class for this,
but when you like a string give me one and I will implement.

The idea to use a class is that easy to distinguish if this value came from user input or program logic(when i see right is this a hot topic around smilies).
When this not require the string solution works as well.

albertlast:
So the other pr use now the string 'null' as token,
also no need more for this pr.

PR #3246, opened 2016-01-07
Fixes016

FearThe1337:
Fixes 2 issues, see commit messages.


FearThe1337:
Just adding:
Found the error message of the first commit in the error log of my forums. Which was done by a moderator trying to delete their own message, hence the second fix.

Issue #3261, opened 2016-01-21
MySQL 5.7 and updated topics


socram8888:
Due to MySQL 5.7 having deprecated `ORDER BY`s where the clause isn't in the SELECT, the following code doesn't work and fails to execute.

``` sql
SELECT DISTINCT t.id_topic
FROM smf_topics AS t
INNER JOIN smf_messages AS m ON (m.id_topic = t.id_topic AND m.id_member = 865)
LEFT JOIN smf_log_topics AS lt ON (lt.id_topic = t.id_topic AND lt.id_member = 865)
LEFT JOIN smf_log_mark_read AS lmr ON (lmr.id_board = t.id_board AND lmr.id_member = 865)
WHERE t.id_board IN (3, 4, 5, 6, 13, 16, 20)
AND t.id_last_msg >= 863236
AND (IFNULL(lt.id_msg, IFNULL(lmr.id_msg, 0))) < t.id_last_msg
AND t.approved = 1 AND lt.unwatched != 1
ORDER BY t.id_last_msg DESC
LIMIT 0, 50
```

Error is:

> # 3065 - Expression #1 of ORDER BY clause is not in SELECT list, references column 'etl_smf.t.id_last_msg' which is not in SELECT list; this is incompatible with DISTINCT

So options I see are:
- Reenable the deprecated feature
- Change the SQL code, probably using a subquery to list all the posts and then running a SELECT DISTINCT on the outer

This is the only code I've noticed so far has stopped working in MySQL 5.7 but might not be the only one.

albertlast:
For the documentation this error happen here:
https://github.com/SimpleMachines/SMF2.1/blob/4d2e3339cee682bce6bdce715d82e6ab2e4532aa/Sources/Recent.php#L1122

This error is already fix in the pg query: https://github.com/SimpleMachines/SMF2.1/blob/02d286edbb0341ca05f47b8f7f85437184a306e0/Sources/Subs-Db-postgresql.php#L318

I would suggest we take the pg solution as general solution?
Would be mean that we add the col t.id_topic to the select fields in the recent query.

live627:
We should change the code.

albertlast:
This issue should be fixed with the pr #3262 
@socram8888 can you please test the the pm system?
work order: send from one user to another a pm,
go into the inbox of receiver
and press on subject to change the order of pm are listed

give me feedback if there a error is raised up.

PR #3262, opened 2016-01-21
fix SQL Syntax in recent and remove replacments in pg subs #3261

albertlast:
Fix the error in the sql syntax
see: https://github.com/SimpleMachines/SMF2.1/issues/3261

### DIFF
--- a/Sources/Recent.php
+++ b/Sources/Recent.php
@@ -1118,8 +1118,8 @@ function UnreadTopics()
 				))
 			);
 		else
-			$request = $smcFunc['db_query']('unread_replies', '
-				SELECT DISTINCT t.id_topic
+			$request = $smcFunc['db_query']('', '
+				SELECT DISTINCT t.id_topic,'.$_REQUEST['sort'].'
 				FROM {db_prefix}topics AS t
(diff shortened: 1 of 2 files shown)


socram8888:
This AFAIK will fail, as DISTINCT works with tuples: if somebody posts two messages on the same thread, it'll be shown twice in the "Updated topics" section (vs once as it does currently).

What I had done was replacing the `DISTINCT` by a `GROUP BY t.id_topic`, and was testing it in my server before doing a PR.

albertlast:
Nope this will not double the rows,
reason both columns are from the same table.

here a smale exp: http://sqlfiddle.com/#!9/035bf/1/0

and other columsn shouldn't use because the postgresql sql was hardcoded vs the one column(t.id_last_msg),
so this problem would raise up earlier.

socram8888:
Okay, suppose in your example, "a" is thread ID, and "b" is last post ID.

In current implementation, it would return only "1". Using your modification, note it returns "1,2" and "1,3". This is, it returns thread 1 twice.

albertlast:
Yes but id_topic is primary key, 
so you will never seen 1,2 and 1,3 exists at the same time.

socram8888:
Oh derp, you're right, sorry!

PR #3263, opened 2016-01-21
correct the pg replacement for order by null

albertlast:
To delete the order by null it's more clean what
the intention of the "order by null" is.

It's a optimization to supress mysql to sort the date, https://blog.mozilla.org/it/2012/09/05/should-you-use-group-by-order-by-null-by-default/
postgresql didn't sort data without a order by.

technical it's the same but this solution looks nicer.


live627:
Could ORDER BY also be removed from the query? It doesn't GROUP BY anything. https://github.com/SimpleMachines/SMF2.1/blob/19ee85ff8761b792ea3e9ed630a947f45f93ee68/cron.php#L132-L141

albertlast:
yeah your are right,
there is no need for the order by in the first place.

albertlast:
pr got broken

PR #3264, opened 2016-01-21
improvment of exist real/membername check

albertlast:
Some improvments around realname and membername check.

This query got already a lower(col) for db who are cs,
so it's exists no reason to use ilike.

Under some special circumstance we can now use the index from: https://github.com/SimpleMachines/SMF2.1/pull/3215

And when no wildcard is used, we can use = instead of LIKE

### DIFF
--- a/Sources/Subs-Members.php
+++ b/Sources/Subs-Members.php
@@ -905,13 +905,14 @@ function isReservedName($name, $current_ID_MEMBER = 0, $is_name = true, $fatal =
 		SELECT id_member
 		FROM {db_prefix}members
 		WHERE ' . (empty($current_ID_MEMBER) ? '' : 'id_member != {int:current_member}
-			AND ') . '({raw:real_name} '.$operator.' LOWER({string:check_name}) OR {raw:member_name} '.$operator.' LOWER({string:check_name}))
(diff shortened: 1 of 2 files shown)


Sources/Subs-Members.php
```diff
@@ -896,13 +896,16 @@ function isReservedName($name, $current_ID_MEMBER = 0, $is_name = true, $fatal =
...
+	$operator = (strpos($checkName, '%') || strpos($checkName, '_') ? 'LIKE' : '=' );
 
 	// Make sure they don't want someone else's name.
-	$request = $smcFunc['db_query']('case_insensitive', '
+	$request = $smcFunc['db_query']('', '
 		SELECT id_member
 		FROM {db_prefix}members
 		WHERE ' . (empty($current_ID_MEMBER) ? '' : 'id_member != {int:current_member}
-			AND ') . '({raw:real_name} LIKE {string:check_name} OR {raw:member_name} LIKE {string:check_name})
+			AND ') . '({raw:real_name} '.$operator.' LOWER({string:check_name}) OR {raw:member_name} '.$operator.' LOWER({string:check_name}))
```
matthew-kerle: Should be: ' . $operator . ' not '.$operator.'.

albertlast: Did you mention the spaces?

jdarwood007: Actually should use {raw:operator} instead.  It looks cleaner and can be read easier.

PR #3268, opened 2016-01-31
fix PHP7 Login Logout error msg 

albertlast:
the two error msg on logout and the one on login are fixed by this pr


XinYenFon:
Tested, works awesome.

PR #3271, opened 2016-02-01
return static Var in DB get version in a own line

albertlast:
as you requested, https://github.com/SimpleMachines/SMF2.1/pull/3220
i move the return in a own line


MissAllSunday:
Thanks!

PR #3272, opened 2016-02-01
fix SQL syntax in Display

albertlast:
the use of having is here wrong,
having is in sql standard only allow on col where aggregate function like count,max.. is used

and again less specific pg code


Oldiesmann:
You need to keep the check for the "group by" fix. It's needed for MySQL 5.0.50 and 5.0.51, and we support 5.0.3 and newer in 2.1 (even though 5.0.3 is quite old by now).

albertlast:
This group by fix is not needed when having is not used.

PR #3273, opened 2016-02-01
remove not needed HAVING sql statements

albertlast:
I convert this from mysql tolerated sql statements to valid sql standard statements.


jdarwood007:
https://en.wikipedia.org/wiki/Having_(SQL)

Having is a valid SQL language syntax.  I don't see why removal is needed.

albertlast:
Having is valid sql statement right, but it's only allow on aggregate value:"aggregate values meet the specified conditions" 
aggregate value you get only by aggregate function like count, min,max etc.

this is the reason that you see on the wiki only statement like having  count(*) > 1

so the right approach is to use the where in this statements

PR #3276, opened 2016-02-06
fixed smiley search in admin panel

albertlast:
When i was in the admin panel and used the search function
by quick search as input "asd" category "task/settings"
then i got errors like:

 0e7b0e7d6d3ea23830a821df567d9a6f
Type of error: Undefined
8: Undefined offset: 1
http://localhost/SMF2.1/index.php?action=admin;area=search
G:/github/SMF2.1/Sources/ManageSmileys.php
Line: 129

After the change of the explode it's works,
but i'm not sure why it's before.


Oldiesmann:
I can't reproduce this issue. The fact that the change fixes it for you would indicate that the smiley set names aren't stored properly in the database.

PR #3277, opened 2016-02-06
convert all sql IFNULL to coalesce

albertlast:
i noticed that some queries got nested ifnull express like the display:
IFNULL(lt.id_msg, IFNULL(lmr.id_msg, -1))

I look in the documentation from mysql and see that this function only support two expression,
that is the reason that i convert everything to the sql standard COALESCE,
this function got no limit of the amount of exp.

overall three reasons:
1. this is the standard way
2. no need to create ifnull function in the install process of pg db
3. no limit of amount of expressions

### DIFF
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -156,11 +156,11 @@ function Display()
 		SELECT
 			t.num_replies, t.num_views, t.locked, ms.subject, t.is_sticky, t.id_poll,
 			t.id_member_started, t.id_first_msg, t.id_last_msg, t.approved, t.unapproved_posts, t.id_redirect_topic,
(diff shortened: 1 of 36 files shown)


Underdog-01:
requires MySQL 5.6 
ensure SMF 2.1 minimal requirements are updated to reflect these changes
ref. http://wiki.simplemachines.org/smf/SMF2.1:Features#Minimum_requirements
(prior to this commit .. MySQL 5.5 was necessary & page was also not updated)

albertlast:
coalesce exists with 5.0 (maybe earlier) http://dev.mysql.com/doc/refman/5.0/en/comparison-operators.html

Underdog-01:
I was looking at the w3resource site which seems to be incorrect.
The MySQL main website as you referenced must be correct and it appears to have been introduced in MySQL 5.0

PR #3282, opened 2016-02-13
Do Topic Starter sort in PHP

albertlast:
I think it's a better approach to do the sort in php as in rdbms
- less db specific code
- less loc
- code is easier

### DIFF
--- a/Sources/Stats.php
+++ b/Sources/Stats.php
@@ -450,7 +450,7 @@ function DisplayStats()
 	$max_num = 1;
 	while ($row_members = $smcFunc['db_fetch_assoc']($members_result))
 	{
-		$i = $members[$row_members['id_member']];
+		$i = array_search($row_members['id_member'], array_keys($members));
 		$context['stats_blocks']['starters'][$i] = array(
 			'name' => $row_members['real_name'],
 			'id' => $row_members['id_member'],
@@ -462,7 +462,7 @@ function DisplayStats()
 		if ($max_num < $members[$row_members['id_member']])
 			$max_num = $members[$row_members['id_member']];
 	}
-	krsort($context['stats_blocks']['starters']);
+	ksort($context['stats_blocks']['starters']);
 	$smcFunc['db_free_result']($members_result);
 
 	foreach ($context['stats_blocks']['starters'] as $i => $topic)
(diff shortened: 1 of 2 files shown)


Sources/Stats.php
```diff
@@ -437,22 +437,21 @@ function DisplayStats()
...
 			'member_list' => array_keys($members),
-			'top_topic_posters' => implode(',', array_keys($members)),
 		)
 	);
 	$context['stats_blocks']['starters'] = array();
 	$max_num = 1;
 	while ($row_members = $smcFunc['db_fetch_assoc']($members_result))
 	{
-		$context['stats_blocks']['starters'][] = array(
+		$i = array_search($row_members['id_member'], array_keys($members));
```
live627: `$i = $members[$row_members['id_member']]`

Sources/Stats.php
```diff
@@ -437,22 +437,21 @@ function DisplayStats()
...
+		$context['stats_blocks']['starters'][$i] = array(
 			'name' => $row_members['real_name'],
 			'id' => $row_members['id_member'],
 			'num' => $members[$row_members['id_member']],
```
live627: `'num' => $i`

Sources/Stats.php
```diff
@@ -463,6 +462,7 @@ function DisplayStats()
 		if ($max_num < $members[$row_members['id_member']])
 			$max_num = $members[$row_members['id_member']];
 	}
+	ksort($context['stats_blocks']['starters']);
```
live627: This is backwards.

http://php.net/manual/en/function.krsort.php

albertlast:
The first idea was,
to evade the whole sorting stuff and set the array on the right index,
but php didn't allow stuff like this.

So I take your advice.
I keep on line 453 with $i = ... , because otherwise the line 454 get to big.

live627:
I think  I  foresee an issue with this. Two or more members with the same stat won't show right.

live627:
Only one will show; the rest are clipped.

albertlast:
I didn't noticed this too,
move back to the first solution.

live627:
I haven't tested it, but it seems the first solution won't sort correctly either.

The solution here is to sort the array according to the order of $members.

albertlast:
In my test area this works: http://snag.gy/BPxG3.jpg

PR #3303, opened 2016-02-29
fix add spider -> doublicate key constraint

albertlast:
when the id is direct given the sequence will not updated,
get douplicate key conflict on the first insert without hard id

by adding a new spider
in managedsearchengine line 417 try to insert a new row without id -> error
reason is that the sequence stand on 1.

btw this problem exists since 2.0.x

### DIFF
--- a/other/install_2-1_mysql.sql
+++ b/other/install_2-1_mysql.sql
@@ -2061,27 +2061,27 @@ VALUES (':)', 'smiley.gif', '{$default_smiley_smiley}', 0, 0),
 #
 
 INSERT INTO {$db_prefix}spiders
-	(id_spider, spider_name, user_agent, ip_info)
-VALUES (1, 'Google', 'googlebot', ''),
-	(2, 'Yahoo!', 'slurp', ''),
-	(3, 'MSN', 'msnbot', ''),
-	(4, 'Google (Mobile)', 'Googlebot-Mobile', ''),
-	(5, 'Google (Image)', 'Googlebot-Image', ''),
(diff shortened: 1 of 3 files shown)


live627:
I thought the installer skipped SQL commands if they generated errors.

albertlast:
This problem raised up,
after the installation when we try to add a spider per admin interface.

jdarwood007:
@albertlast 

Change the install from this

```
CREATE SEQUENCE {$db_prefix}spiders_seq;
```

To this:

```
CREATE SEQUENCE {$db_prefix}spiders_seq START WITH 21;
```

I think that resolves your issue properly without changing a bunch of other things.  If we add more default spiders, we will need to change that to our max id + 1

albertlast:
This is the wrong way.

The user is able to add spiders by his own since smf 2.0,
when you add spiders later on you can define the id in install process,
but in the upgrade process the next id neet to take.

So when the user define two spider, the new one gets 23 and not 21.

Right would be to take out the id's from mysql too,
to keep even.

jdarwood007:
The installer only runs once so no worries.  Look at how we create categories and boards during the install process.

albertlast:
Yeah i see,
you do it wrong and fixed afterwards

Do it right in the first place,
is a better way to provide the samelogic between mysql an pg.

live627:
Show us, sensei.

albertlast:
i all ready done this,
next step would be to even the whole insert statements.
that would i like to do in another pr.

PR #3304, opened 2016-03-01
Adds integration hooks to many SSI functions

Sesquipedalian:
It would be nice to be able to access many of the existing SSI functions via integration hooks, so this adds a bunch. (Not every SSI function gets one, of course; just the ones where it makes sense.)

In general, call_integration_hook() has been positioned so that it passes a given SSI function's output to the hook just before returning the output, so that the hooked code can make any final changes it needs to.

For the ssi_todaysCalendar() variants (ssi_todaysBirthdays, ssi_todaysHolidays, ssi_todaysEvents, and ssi_todaysCalendar), the same hook is used in each case, and call_integration_hook() simply passes on the $eventOptions array as a parameter so that the hooked code can handle different cases as needed.

For ssi_news(), nothing is passed to the hook, because the only relevant data is already in $context.

### DIFF
--- a/SSI.php
+++ b/SSI.php
@@ -1010,12 +1010,12 @@ function ssi_queryMembers($query_where = null, $query_where_params = array(), $q
 
 	if (empty($members))
 		return array();
-
-	// Load the members.
-	loadMemberData($members);
 	
 	// If mods want to do somthing with this list of members, let them do that now.
 	call_integration_hook('integrate_ssi_queryMembers', array(&$members));
+
+	// Load the members.
+	loadMemberData($members);
 
 	// Draw the table!
 	if ($output_method == 'echo')
--- a/SSI.php
+++ b/SSI.php
@@ -439,6 +439,9 @@ function ssi_queryPosts($query_where = '', $query_where_params = array(), $query
 			);
 	}
 	$smcFunc['db_free_result']($request);
+	
+	// If mods want to do somthing with this list of posts, let them do that now.
+	call_integration_hook('integrate_ssi_queryPosts', array(&$posts));
 
 	// Just return it.
 	if ($output_method != 'echo' || empty($posts))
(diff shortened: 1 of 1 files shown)


SSI.php
```diff
@@ -998,6 +1013,9 @@ function ssi_queryMembers($query_where = null, $query_where_params = array(), $q
 
 	// Load the members.
 	loadMemberData($members);
+	
+	// If mods want to do somthing with this list of members, let them do that now.
+	call_integration_hook('integrate_ssi_queryMembers', array(&$members));
```
live627: this should be before loadMemberData

Sesquipedalian: Ah, yes. You are right.

SSI.php
```diff
@@ -1312,6 +1336,9 @@ function ssi_recentPoll($topPollInstead = false, $output_method = 'echo')
 	}
 
 	$return['allowed_warning'] = $row['max_votes'] > 1 ? sprintf($txt['poll_options6'], min(count($sOptions), $row['max_votes'])) : '';
+	
+	// If mods want to do somthing with this list of polls, let them do that now.
+	call_integration_hook('integrate_ssi_recentPoll', array(&$return, $topPollInstead));
```
live627, line 1341: this should be before $return['allowed_warning']

Derp. nvm.

live627:
Please sign off your commits.

SSI.php
Sesquipedalian, line 1341: Really? I don't really get why...

Edit: Fair enough, I'll nvm. :)

Sesquipedalian:
I don't much like the idea of publicly posting my raw email address for spam bots to pick up, but if the ... [at] ... [dot] ... convention isn't good enough to count as signing off on my code, alright then.

live627:
Signing the PR isn't good enough, sorry...

Sesquipedalian:
Fixed now.

live627:
Ah thanks.

PR #3308, opened 2016-03-01
Minify

MissAllSunday:
Implements https://github.com/matthiasmullie/minify to minify cs and js files.

Exclusive by default, only those files that explicitly set the "minimize" key in $params will be added. External files cannot be added.

This is designed to minify only files that are required everywhere and for everyone, on a vanilla installation the following files are minified:
- scripts.js
- theme.js
- smf_jquery_plugins.js
- index.css
- responsive.css

There are no "on the fly" files, all minimized files are physically created inside their respective folders in default theme, a one day cache entry was added to make sure the files aren't re-created on every requests and also to make sure possible changes on minified files are being merged.


live627:
Very nice!

PR #3311, opened 2016-03-02
Fixes invalid and suboptimal XML feed output

Sesquipedalian:
- Ensures Atom's rel="self" link always has the correct URL in the href attribute.
- Adds the atom:link element and the Atom xmlns to RSS2 feeds, as per the RSS2 spec.
- Replaces the invalid 'modified' element with 'updated' in the Atom news feed.


live627:
Thank you for your submission.

Issue #3301, opened 2016-02-29
Undefined variable: ip_array


lurkalot:
Not sure if this is a bug of not, or if it's been reported and fixed, thought I'd mention it anyway.

I went to Admin > Forum > Search Engines.  

Set Search Engine Tracking Level = Standard
Set Apply restrictive permissions from group = Disabled
Set Show spiders in the online list = Show Spider Names

Within a few minutes I got two errors in the log, (now have nineteen errors all the same type).

Guest
Today at 08:41 PM
a1deccd17653562588e401be2c4d795c
Type of error: Undefined
8: Undefined variable: ip_array 
http://cctestsite.info/testsite3/index.php?topic=2.0 
/homex/xxxxxx/xxxxxx/cctestsite.info/testsite3/Sources/Subs.php
Line: 3510

Guest
Today at 08:34 PM
1e6daff224f2312def4eeddf6378df85
Type of error: Undefined
8: Undefined variable: ip_array 
http://cctestsite.info/testsite3/index.php? 
/homex/xxxxxxx/xxxxxxx/cctestsite.info/testsite3/Sources/Subs.php
Line: 3510 

I also noticed in the users online page a IP address from the guest as 0.0.0.0
![whosol](https://cloud.githubusercontent.com/assets/6047983/13404352/235d9116-df11-11e5-9be7-0a18491ab952.jpg)

MissAllSunday:
ip2range function in Subs.php has had some changes recently, make sure you got the most recent code from this repo.

If you have an old installation then you will need to manually add the follow columns:

ALTER TABLE {$db_prefix}ban_items ADD COLUMN ip_low varbinary(16);
ALTER TABLE {$db_prefix}ban_items ADD COLUMN ip_high varbinary(16);

and delete some these:

ALTER TABLE {$db_prefix}ban_items
DROP ip_low1,
DROP ip_low2,
DROP ip_low3,
DROP ip_low4,
DROP ip_low5,
DROP ip_low6,
DROP ip_low7,
DROP ip_low8;

MissAllSunday:
Oh, nevermind, I can reproduce this on my test board too.

albertlast:
ther should be two error,
- not valid ip address
- undefined var if the ip address are not valid (this i fixed)

btw in line 3510 doesn't exist this return -> the Subs.php is not update to date?

MissAllSunday:
I'm getting the following error:

```
[...]

PR #3316, opened 2016-03-03
make IP2 ipv6 compatible

albertlast:
fix the issue https://github.com/SimpleMachines/SMF2.1/issues/3301
i'm not sure why i need to this, but with this it's works.


live627:
It's much cleaner, that's for sure.

PR #3317, opened 2016-03-03
remove constructBanQueryIP

albertlast:
I see no good reason to use this function,
instead off a clean version.

### DIFF
--- a/Sources/Security.php
+++ b/Sources/Security.php
@@ -179,8 +179,8 @@ function is_not_banned($forceCheck = false)
 		{
 			if ($ip_number == 'ip2' && $user_info['ip2'] == $user_info['ip'])
 				continue;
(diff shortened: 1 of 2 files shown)


Sources/Security.php
```diff
@@ -179,7 +179,8 @@ function is_not_banned($forceCheck = false)
 		{
 			if ($ip_number == 'ip2' && $user_info['ip2'] == $user_info['ip'])
 				continue;
-			$ban_query[] = constructBanQueryIP($user_info[$ip_number]);
+			$ban_query[] = ' {inet:ip} BETWEEN bi.ip_low and bi.ip_high';
+			$ban_query_vars['ip'] = $user_info[$ip_number];
```
live627: One param for two different ip addressees...

albertlast: yes bad mistake

PR #3331, opened 2016-03-08
Fixes broken 'Remove empty BBC' regex

Sesquipedalian:
The previous regex didn't make sense. The problematic part was:

```
(?>(?R)|.)*?
```

That made an atomic group that first tried to recursively match nothing (which of course fails), and then tried to match any character except a line break, which only fails if the BBCode actually IS completely empty. Then the group was lazily repeated. As a result, this line was actually removing all BBCodes that were NOT empty!

The new version simply looks for zero or more whitespace characters between the BBCode tags. This does what the previous version seems to have been intended to do.


Sesquipedalian:
Actually, it occurs to me that even this fixed version is likely to cause problems. I've fixed it so that it only matches empty BBCode tags, but it could still match against _fake_ BBCode tags. For example, the regex as it stands in 08ff4bf will remove not only `[b][/b]` and `[url]      [/url]`, but also `[sarcasm]   [/sarcasm]` and `[guess][/guess]`. Granted, that is likely to be a rare situation. Usually when users create fake BBCode for humorous effect, they put some content in between the fake tags. But even so, the right way to handle this would be to make sure that the regex only matches against valid BBCode tags.

The trouble is, to do so we would need to pull in the array of valid tags from parse_bbc(). But in that case, it probably makes more sense just to wait and kill off any empty BBCode during parse_bbc() rather than here in preparsecode().

Thoughts?

PR #3336, opened 2016-03-11
Makes SCEditor more responsive to on-the-fly changes in display size

Sesquipedalian:
SCEditor uses JavaScript to calculate the dimensions of the iframe and textarea on page load. But this doesn't work when the display size changes after page load, e.g. if a mobile user rotates their device. This can be fixed by setting the min-width and max-width values in a few key places. Technically, this makes SCEditor's calculations pointless, since they are overridden by the CSS anyway, but that doesn't hurt anything.

### DIFF
--- a/Themes/default/css/jquery.sceditor.css
+++ b/Themes/default/css/jquery.sceditor.css
@@ -146,6 +146,8 @@ div.sceditor-grip {
 	font-weight: 700;
 	border-radius: 4px;
 	background-clip: padding-box;
+	min-width: 100%;
+	max-width: 100%;
 }
 .sceditor-container {
 	font-weight: normal;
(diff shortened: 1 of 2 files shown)


XinYenFon:
Did you test those changes with resizing the textbox since you can do that as well.

Sesquipedalian:
Yes I did. With these changes it can still be resized vertically just like always. It can't be arbitrarily resized horizontally, but that's no loss.

XinYenFon:
One last question actually, did you resize and rotate your device ?

Sesquipedalian:
Yes. It works perfectly on an iOS device, at any rate. I don't have an Android device to test on, but I can't imagine how it could be different. The behaviour of max-width and min-width is well defined and consistent on all mobile browsers. 

The only place this could have been be an issue is with a browser (nothing later than IE6) that doesn't support the max-width and min-width CSS properties. But that's not a problem, since SCEditor's JavaScript is still operating and will act as a fallback in that case.

live627:
Earliest IE we support is 9.

Issue #3257, opened 2016-01-19
Textboxes too wide in Chrome


Oldiesmann:
In Chrome, many of the textboxes are too wide, which looks weird and causes horizontal scrollbars in some cases. Compare these two screenshots. Top one is Firefox, bottom one is Chrome.

![search_firefox](https://cloud.githubusercontent.com/assets/89023/12426549/2e814af8-bea8-11e5-96fb-9e8ccca80241.png)

![search_chrome](https://cloud.githubusercontent.com/assets/89023/12426544/26740dc8-bea8-11e5-9fb4-e379e27d0d15.png)

PR #3346, opened 2016-03-16
Another regression

live627:
There's a reason black cats r evil. heh


XinYenFon:
WutFace ? That's put in there because of overflowing issue Chrome causing cc @Oldiesmann

Sesquipedalian:
What does this overflow issue involve, @Antes?

XinYenFon:
https://github.com/SimpleMachines/SMF2.1/issues/3257

PR #3348, opened 2016-03-16
Generic Improvements on Curve2

XinYenFon:

### DIFF
--- a/Themes/default/Errors.template.php
+++ b/Themes/default/Errors.template.php
@@ -225,7 +225,7 @@ function template_show_file()
 		$is_target = $line_num == $context['file_data']['target'];
 		echo '
 			<tr>
-				<td ', $is_target ? ' class="current righttext" : "righttext">==&gt;' : '>', $line_num , ':</td>
+				<td class="righttext', $is_target ? ' current">==&gt;' : '">', $line_num , ':</td>
 				<td style="white-space: nowrap;', $is_target ? ' border: 1px solid black;border-width: 1px 1px 1px 0;':'','">', $line, '</td>
 			</tr>';
 	}
--- a/Themes/default/Errors.template.php
+++ b/Themes/default/Errors.template.php
@@ -225,7 +225,7 @@ function template_show_file()
 		$is_target = $line_num == $context['file_data']['target'];
 		echo '
 			<tr>
-				<td align="right"', $is_target ? ' class="current">==&gt;' : '>', $line_num , ':</td>
+				<td ', $is_target ? ' class="current righttext" : "righttext">==&gt;' : '>', $line_num , ':</td>
 				<td style="white-space: nowrap;', $is_target ? ' border: 1px solid black;border-width: 1px 1px 1px 0;':'','">', $line, '</td>
 			</tr>';
 	}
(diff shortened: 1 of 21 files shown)


Themes/default/Errors.template.php
```diff
@@ -225,7 +225,7 @@ function template_show_file()
 		$is_target = $line_num == $context['file_data']['target'];
 		echo '
 			<tr>
-				<td align="right"', $is_target ? ' class="current">==&gt;' : '>', $line_num , ':</td>
+				<td ', $is_target ? ' class="current righttext" : "righttext">==&gt;' : '>', $line_num , ':</td>
```
Oldiesmann: This is incorrect. This will cause it to output `<td class="current righttext" : "righttext">==&gt;"` if $is_target is true. If the order in which the classes are defined doesn't matter then you can just do this: `<td class="righttext', $is_target ? ' current">==&gt;' : '">'`. Otherwise you'll need to do it like this: `<td ', $is_target ? 'class="current righttext">==&gt;' : 'class="righttext">'`

Themes/default/Reports.template.php
```diff
@@ -111,12 +111,12 @@ function template_main()
 				// Shaded?
 				if ($column_number == 0 && !empty($table['shading']['left']))
 					echo '
-					<td align="', $table['align']['shaded'], '" class="table_caption"', $table['width']['shaded'] != 'auto' ? ' width="' . $table['width']['shaded'] . '"' : '', '>
```
Oldiesmann, line 114: Since we're setting the alignment based on a variable, we should handle that accordingly rather than assuming they'll always be the default values. Just set `class="', $table['align']['shaded'], 'text"` or something similar.

Themes/default/Errors.template.php
XinYenFon: `<td class="righttext', $is_target ? ' current">==&gt;' : '">'` is cleaner. Considering its already has align right righttext needs to be there as well.

Themes/default/css/index.css
```diff
@@ -57,6 +58,10 @@ textarea, textarea.editor {
 input, button, select {
 	font-size: 0.9em;
 }
+/* Height fixation */
+input, select {
```
live627, line 62: May I ask why?

XinYenFon, line 62: because they have different heights ? and different browsers natively handle height/width differently. We need a standart on that.

live627, line 62: Won't that break select lists? See the avatar selection...

`select:not([size])`

XinYenFon, line 62: Yea seems they are lacking class. That side of SMF also needs improvements then. I'll take care after this mid-term exam.

XinYenFon, line 62: I changed stuff but probably missing some more lists around if you find any you can use "select_list" class to fix them.

live627, line 62: May I know the reason for introduction yet another CSS class?

Why not just use :not? http://caniuse.com/#feat=css-sel3

XinYenFon, line 62: How you gonna target :not( what select ?)

live627, line 62: https://github.com/SimpleMachines/SMF2.1/pull/3348/files/2a16456c03021c3dcede045f4abc81c63d46233c#r57670510

XinYenFon, line 62: lol my bad :P missed that...

live627:
lmk  when I can merge this

XinYenFon:
anytime you want?

PR #3354, opened 2016-03-18
Implements responsive design for the calendar

Sesquipedalian:
- As the screen gets smaller, first we force the mini calendars to hide, then we switch the main calendar into a vertical list view. This involves manipulating the CSS to make all the table cells display as blocks and various other bits of CSS chicanery.
- To help avoid long words stretching table cells wider than we want, this also implements CSS hyphenation on browsers that support it, and word-breaking on browsers that don't. Note that to make hyphenation work, the lang attribute needs to be set on a parent element somewhere (see SimpleMachines#3351).
- This also does a bit of tidying in the CSS file.
- Removes some redundant classes in the template, adds some others where needed, and lightly rearranges a few elements. Also removes a couple of `<br class="clear">` elements that were no longer needed.

### DIFF
--- a/Themes/default/Calendar.template.php
+++ b/Themes/default/Calendar.template.php
@@ -51,7 +51,7 @@ function template_main()
 	}
 
 	// Close our wrapper.
-	echo '<br class="clear">
+	echo '
 	</div>';
 }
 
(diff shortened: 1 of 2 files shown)


Sesquipedalian:
Demo available at http://stovell.info/forum/index.php?action=calendar;year=2016;month=2

PR #3356, opened 2016-03-21
Does a better job declaring the lang attribute

Sesquipedalian:
- Doesn't assume that the primary language subtag is two characters long (there are many defined in BCP 47 that are three characters long). Also, keeps all subtags that are defined in the locale code, such as script and region tags. In some cases, the subtags are unnecessary but harmless (e.g. 'fr-FR') while in others they are very important (e.g. 'zh-Hans' vs. 'zh-Hant')
- TODO: As of 2016-03-21, several of SMF's language files use deprecated locale strings. These should be updated for SMF 2.1.

### DIFF
--- a/Themes/default/index.template.php
+++ b/Themes/default/index.template.php
@@ -83,9 +83,9 @@ function template_html_above()
 {
 	global $context, $settings, $scripturl, $txt, $modSettings, $mbname;
 
-	// Show right to left and the character set for ease of translating.
+	// Show right to left, the language code, and the character set for ease of translating.
 	echo '<!DOCTYPE html>
-	<html', $context['right_to_left'] ? ' dir="rtl"' : '', !empty($txt['lang_locale']) ? ' lang="' . substr($txt['lang_locale'], 0, 2) . '"' : '' , '>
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Someone ought to go through the language packs available at http://download.simplemachines.org/?smflanguages and update the locale strings to fix any that are using deprecated values. For example, `sr-YU` ought to be replaced with `sr-BA`, and `zh-CN` and `zh-TW` ought to be replaced with `zh-Hans-CN` and `zh-Hant-TW`. I would volunteer to do it, but the language packs don't seem to be available on GitHub.

XinYenFon:
@Sesquipedalian neither they are generated by hand. :) I'll pass that info to our localization team.

Sesquipedalian:
Cool. This page should be helpful for them: https://www.w3.org/International/articles/language-tags. Also, there's this handy lookup tool to find the current language tags in the IANA registry: http://r12a.github.io/apps/subtags/

Sesquipedalian:
Btw, in case it wasn't clear, this pull request isn't dependent on updating the localization tags. That was just an additional suggestion on my part.

PR #3360, opened 2016-03-24
Default to server timezone for new members

### LINKED ISSUE #3234: timezone forum default
when you register an account there is an option to use the forum default for time offset. this does not seem to work. a new user seems to get time zone UTC when using this option but local time in their profile when viewed as someone else shows the correct time. in that users profile it does show them as UTC and not the forum default timezone.

Sesquipedalian:
- Fixes #3234
- Question: Could/should we do anything smarter than this, like looking up the timezone of the requesting IP address or something?


live627:
> Question: Could/should we do anything smarter than this, like looking up the timezone of the requesting IP address or something?

Nah; that could easily be spoofed by a VPN.

asmith20002:
I think it's best to put the comma , at the end of that line. That's how SMF does it anyway. It's better for future possible modifications.

PR #3361, opened 2016-03-25
Don't require Settings_bak.php on fresh install

Sesquipedalian:
- is_writable() fails on non-existent files, but on a fresh install there is no Settings_bak.php yet.

### DIFF
--- a/other/install.php
+++ b/other/install.php
@@ -502,9 +502,11 @@ function CheckFilesWritable()
 		'Smileys',
 		'Themes',
 		'agreement.txt',
-		'Settings.php',
-		'Settings_bak.php'
+		'Settings.php'
 	);
+	if (file_exists(dirname(__FILE__) . '/Settings_bak.php'))
+		$writable_files[] = 'Settings_bak.php';
+	
 	foreach ($incontext['detected_languages'] as $lang => $temp)
 		$extra_files[] = 'Themes/default/languages/' . $lang;
 


jdarwood007:
To improve that you should have it attempt to touch/create the backup file and if it can't, fail saying the backup file is not writable.  Otherwise server permissions may not allow it to create the file and then no backup file exists.

Sesquipedalian:
I disagree. The right time to check whether Settings_bak.php can be created and written to is when we actually want to do so. But the install process never actually writes anything to that file. On a fresh install it doesn't exist yet because doesn't need to—there is nothing to back up yet. The only time Settings_bak.php will already exist is if the user is reinstalling over a previous install (maybe a botched one). So if Settings_bak.php is there, we should make sure it is writable, but if it isn't we should just move on.

jdarwood007:
That doesn't seem right.  Settings_bak.php should contain the same information as Settings.php after the installer writes it.  That way if Settings.php gets corrupted or broken somehow, its a simple fix for the admin to replace the broken Settings.php with the one from the backup.

I've had this save the day when sometimes Settings.php would get written blank (for various reasons).

Sesquipedalian:
Go ahead and take a look at install.php. It never does anything with Settings_bak.php apart from this check. You might be thinking of upgrade.php, which writes to it several times.

jdarwood007:
I thought it did write the backup as well.  But looking at 2.0 and 1.1 it didn't.

Still not sure if we should ignore if we can't write a backup file though, seems like we may give ourselves more issues for support later.

Sesquipedalian:
We don't ignore it. There are checks and warnings and such when we do try to write to it. Those just happen when the need arises, which is as it should be. 

Strictly speaking, install.php shouldn't even be doing the conditional check that I put into this PR. I only included it as a "better safe than sorry" measure in case something farther down the line somewhere assumed that the writability of the file had already been established during install. To my knowledge there are no such places, but I'm not omniscient. :)

Sesquipedalian:
So, since this PR is still sitting here, does that mean it won't be merged unless I change it so that Settings_bak.php is created during the installation process? @jdarwood007 and I have had some discussion, but unless I'm misreading the conversation, no resolution was reached. What needs to happen for this to be pulled?

Sesquipedalian:
I'll add that I think if Settings_bak.php is going to be required and/or created in the install process, it makes more sense to just have the file included in the distribution to begin with, like it is in 2.0, rather than creating it on the fly.

jdarwood007:
A developer hasn't responded or merged so it is up to them to do so.  I'm not on the dev team so I don't have rights to merge PRs from the SMF repo.

PR #3370, opened 2016-03-29
Log erros and member_ip(2) with ipv6 support

albertlast:
I want only update the log errors but i needed in the same place the member_ip to update...
So this big change was not intented.


live627:
Actually, member IP data really should be retained.

PR #3372, opened 2016-03-30
poster_ip with ipv6 support

albertlast:
The same as the others.


live627:
Actually, message IP data really should be retained.

PR #3373, opened 2016-03-30
Big ipv6

albertlast:
I want to let you decide if it better many small pr or one big
this big one got allready resolved conflicts.

I created this in first place for my self to test my pr.

messages.poster_ip, member.ip, member.ip2 need converting
Found a solution in pg where wrong ipadress will be ingored -> every field in pg will be converted
Solution for mysql is missing.
only floodcontrol table get truncated because the ip field is pkey

Status:
done:
log_actions.ip
log_banned.ip
log_errors.ip,members.member_ip, members.member_ip2
messages.poster_ip
log_floodcontrol.ip, log_online.ip
log_reported_comments.ip
member_logins.ip, member_logins.ip2

not a real one:
spiders.ip_info


live627:
One big PR is better. All these changes are interrelated in that they change IP handling. 

(side benefit for me: i have only one branch to test now!)

albertlast:
The topic is from technical pov done,
for mysql is coverting path missing,
someone with a good idea how to realize?

All field with the exception of floodcontrol are null able,
not valid ip address will be converted into null.

live627:
> for mysql is coverting path missing,
> someone with a good idea how to realize?

Just changing the column type should suffice. I know that works on MariaDB 10, I'm hoping that MySQL behaves similarity.

albertlast:
changing the col does a converting but the result is wrong:
http://sqlfiddle.com/#!9/1eb5d7/1/0
column a should be the same as the INET6_ATON(b) column

jdarwood007:
Seems like we can recast it to fix it:
http://sqlfiddle.com/#!9/1eb5d7/7

albertlast:
The solution needs to be without INET6_ATON because this function exists only with mysql 5.6+.

jdarwood007:
Well using the inet was a simple example to show the data still exists in working order, we could do the ugly pull all data after casting it and let php fix it.  Cast only works  for mysql 5.5 though.

live627:
some Googling brought this up http://stackoverflow.com/a/14539582

albertlast:
The stepp two is key Problem,
Where a solution is needed.

Von: John Rayes notifications@github.com
Gesendet: 02.04.2016 9:55 AM
An: SimpleMachines/SMF2.1
Cc: albertlast
Betreff: Re: [SimpleMachines/SMF2.1] [WIP] Big ipv6 (#3373)

some Googling brought this up http://stackoverflow.com/a/14539582

On Fri, Apr 1, 2016 at 9:15 PM, Jeremy D notifications@github.com wrote:

> Well using the inet was a simple example, we could do the ugly pull all
> data after casting it and let php fix it. Cast only works for mysql 5.5
> though.
> 
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> https://github.com/SimpleMachines/SMF2.1/pull/3373#issuecomment-204641766

---

jdarwood007:
If you go for the 2 column approach (I had the same idea while looking quickly last night), you still have to use PHP since you can't use INET.  So you would have to pull all data down that needs converted and push it up with update statements.  There is no simple massive query that will handle this since we have to combat previous MySQL versions.

We could make it if we have a high enough MySQL/Percora/etc, version that has the ability to use INET6, to upgrade those that way, otherwise fall back to a slower upgrade method of updating each row one by one.

Another solution, could be to build a ip table, have the IP in clear text, and in varbinary.  You would have to pull all distinct ips from the existing tables, convert them and then insert them into the ip table.  But from there you can issue a mass update query to convert everything.

albertlast:
I like your last solution,
so i implented this solution.

I can imagine some edges where this could broke,
but this can befixed when this problems are known.

For this reason i was realy picky and choose only this column which live want to be converted.

I forgot to mention,
someone with a mysql installation should test this(special the upgrade process) and
maybe a backup of the db :)

jdarwood007:
If its only going to be used in the upgrade, move the function to the upgrade file, also use {raw:name} etc as needed rather than uncleanly accepting the data

live627:
But the `raw` keyword takes input as is and doesn't do anything.

https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs-Db-mysqli.php#L253

live627:
IOW, I don't see the point of _using_ `raw`

albertlast:
The missing benefit and the harder to read let me no see good reason to use raw,
but when you want i can do it.

jdarwood007:
@live627 Mostly because it follows our coding standards of clean readable code.  Its easy to understand the input you are adding is being put directly into the query as is, not being sanitized.

We really should have a type to handle things like column/table names in the query to properly make sure it is safe to use and doesn't contain "bad stuff".

live627:
@SimpleMachines/smf-developers-2-1 post test results here

other/upgrade.php
```diff
@@ -5761,4 +5761,63 @@ function doTheDelete(theCheck)
...
+ * newCol needs to be a varbinary(16) null able field
+ * return true or false
+ */
+function MySQLConvertOldIp($targetTable,$oldCol,$newCol)
+{
+	//mysql default max length is 1mb http://dev.mysql.com/doc/refman/5.1/en/packet-too-large.html
+	$max = 10000;
+	$arIp = array();
+	
+	$request = $smcFunc['db_query']('', 'DROP TABLE IF EXISTS {db_prefix}ip_table');
```
live627, line 5777: ( ! ) Fatal error: Function name must be a string in D:\www\SMF2.1\upgrade.php on line 5775

other/upgrade.php
```diff
@@ -5761,4 +5761,63 @@ function doTheDelete(theCheck)
...
+	$request = $smcFunc['db_query']('', 'select distinct '.$oldCol.' from {db_prefix}'.$targetTable);
+	while($row = $smcFunc['db_fetch_assoc']($request))
+		$arIp[] = $row[$oldCol]; 
+	$smcFunc['db_free_result']($request);
+	
+	$insertStart = 'insert into {db_prefix}ip_table(oldip, newip) values';
+	$query = $insertStart;
+	$impArray = array();
+	$x = 0;
+	for($i = 0; $i < length($arIp); $i++)
```
live627: ( ! ) Fatal error: Call to undefined function length() in D:\www\SMF2.1\upgrade.php on line 5794

live627:
You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'table smf_members a join smf_ip_table b on a.member_ip_old = b.oldip set a.mem' at line 1

other/upgrade_2-1_mysql.sql
```diff
@@ -1926,3 +1926,139 @@ DROP ip_high2,
...
+
+---# create pk
+ALTER TABLE {$db_prefix}log_floodcontrol ADD PRIMARY KEY (ip,log_type)
+---#
+
+/******************************************************************************/
+--- update log_online ip with ipv6 support without converting
+/******************************************************************************/
+---# delete old column
+ALTER TABLE {$db_prefix}log_online DROP COLUMN ip;
```
live627, line 2035: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'ALTER TABLE smf_log_online DROP COLUMN ip' at line 2

albertlast, line 2035: Can't find the sql error: http://sqlfiddle.com/#!9/165b08/1/0

maybe the column didn't exists on your database anymore? (because you try sometime to run this upgrade)

okay the sql line before looks special... i added a ; at the end.

live627, line 2035: > okay the sql line before looks special... i added a ; at the end.

that fixed it

albertlast:
Need i to do something,
or are you testing?

PR #3377, opened 2016-04-02
Stuff

MissAllSunday:
- fixes an issue with minify changing vars making unable to collapse categories 
- Change code in Sources/ModerationCenter.php to be php 5.3 compatible
- renamed a few vars to avoid possible var name collision.
- Move a check for atachment previous, prevents undefined errors if no attachment was found.

### DIFF
--- a/Sources/ModerationCenter.php
+++ b/Sources/ModerationCenter.php
@@ -397,8 +397,10 @@ function ModBlockNotes()
 				)
 			);
 
-			$note_owner = $smcFunc['db_fetch_assoc']($get_owner)['id_member'];
+			list($note_owner) = $smcFunc['db_fetch_row']($get_owner);
 			$smcFunc['db_free_result']($get_owner);
+
+			$note_owner = $note_owner['id_member'];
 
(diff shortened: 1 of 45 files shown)


Sources/ModerationCenter.php
```diff
@@ -397,9 +397,11 @@ function ModBlockNotes()
 				)
 			);
 
-			$note_owner = $smcFunc['db_fetch_assoc']($get_owner)['id_member'];
+			list($note_owner) = $smcFunc['db_fetch_row']($get_owner);
 			$smcFunc['db_free_result']($get_owner);
 
+			$note_owner = $note_owner['id_member'];
```
live627: $note_owner[0]

live627: Oh wait... that entire line shouldn't be there unless you get rid of list().

MissAllSunday: ah yes

live627:
Is this okay to get merged?

MissAllSunday:
not yet sadly, still need to push some more stuff

PR #3402, opened 2016-04-22
Compatibility fix

Underdog-01:
simple casting of array


Underdog-01:
Did I miss something from the original PR?
Is there perhaps another method someone is going to implement to fix this issue?
Just curious :)

PR #3405, opened 2016-04-23
Fix IP display bug that affected ~3% of the ips + 1 minor typo.

FearThe1337:
Due to the data being binary a dot or colon can occur in the binary data. Respectively the bug triggers in an IP address containing 46 and 58.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -4847,12 +4847,14 @@ function inet_ptod($ip_address)
 }
 
 /**
- * @param binary $bin An IP address in IPv4, IPv6
+ * @param binary $bin An IP address in IPv4, IPv6 (Either string (postgresql) or binary (other databases))
  * @return string The IP address in presentation format or false on error
  */
 function inet_dtop($bin)
 {
-	if(strpos($bin,'.')!==false || strpos($bin,':')!==false)
+	global $db_type;
+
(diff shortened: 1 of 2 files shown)


Sources/Subs.php
```diff
@@ -4852,9 +4852,6 @@ function inet_ptod($ip_address)
  */
 function inet_dtop($bin)
 {
-	if(strpos($bin,'.')!==false || strpos($bin,':')!==false)
-		return $bin;
```
albertlast: to remove this is a error,
in the furture unknown didn't exists anymore see https://github.com/SimpleMachines/SMF2.1/pull/3373
and this lines are needed for pg

FearThe1337: Then pg should use a new function. This one specifically mentions binary data and it is failing to use binary data correctly due to those checks.

albertlast: please replace if(strpos($bin,'.')!==false || strpos($bin,':')!==false) 
with if(isValidIP($bin))

should be a better solution as deletion the whole logic

albertlast:
this solution is from pg pov good

MissAllSunday:
Hello @FearThe1337 it seems recent changes introduced conflicts with this PR, would you minf resolving those conflicts so this can be merged?

PR #3413, opened 2016-04-27
Prevent some SQL errors

live627:


albertlast:
would be not the right way to use {raw:cur_table} ?

Oldiesmann:
We should be using raw here in my opinion. That's exactly what that option was created for - to prevent problems that could be caused by escaping things that we don't want escaped.

jdarwood007:
If we want to feel safer, we could create 2 more identifiers calling them {column:name} and {table:name}.  Which would simply validate that its a valid name for a table or column ([A-Za-z0-9_-]+ is all we use) and not pass it into making sure its a database safe string (like how {stiing:name} works.

albertlast:
i got the same idea some days ago,
but this could lead you in new problems.

My idea was not only place this identifiers i also wanted to add the right quotes to this
for mysql would be this `(most time) and pg ".
exp. select {col:foo} from {table:bar}
result: select`col1`from`smf_table1`

but when you add quotes,
than you are need realy cautious with casesentive name of table and col name.

exp: smf_ table1 with col1
select CoL1 from `smf_table1` --> work
select `CoL1` from `smf_table1` --> doesn't work
select `col1` from `smf_table1` --> work

so this mean without quotes the name are "caseinsentive"(not directly),
with quotes the name are casesensitive.

Issue #3427, opened 2016-05-02
upgrade.php error


illori:
. +++  add the new one done.
 +++  convertError: mysqli_free_result() expects parameter 1 to be mysqli_result, boolean given File: D:\wamp\www\github2.1\upgrade.php Line: 5778Error: mysqli_free_result() expects parameter 1 to be mysqli_result, boolean given File: D:\wamp\www\github2.1\upgrade.php Line: 5783Error: The database value you're trying to insert does not exist: ip0<em>(upgrade.php-5810)</em> File: D:\wamp\www\github2.1\Sources\Subs-Db-mysqli.php Line: 852Error: Undefined index: ip0 File: D:\wamp\www\github2.1\Sources\Subs-Db-mysqli.php Line: 183Error: The database value you're trying to insert does not exist: ip0<em>(upgrade.php-5810)</em> File: D:\wamp\www\github2.1\Sources\Subs-Db-mysqli.php Line: 852Error: Undefined index: ip0 File: D:\wamp\www\github2.1\Sources\Subs-Db-mysqli.php Line: 183Error: Wrong value type sent to the database. IPv4 or IPv6 expected.(ip0)<em>(upgrade.php-5810)</em> File: D:\wamp\www\github2.1\Sources\Subs-Db-mysqli.php Line: 852Error: inet_pton(): Unrecognized address  File: D:\wamp\www\github2.1\Sources\Subs-Db-mysqli.php Line: 262Error: mysqli_free_result() expects parameter 1 to be mysqli_result, boolean given File: D:\wamp\www\github2.1\upgrade.php Line: 5811Error: mysqli_free_result() expects parameter 1 to be mysqli_result, boolean given File: D:\wamp\www\github2.1\upgrade.php Line: 5816Error: mysqli_free_result() expects parameter 1 to be mysqli_result, boolean given File: D:\wamp\www\github2.1\upgrade.php Line: 5819 done.

albertlast:
so i found three error related to this,
- one is typo mistake which is fixed with this pr: https://github.com/SimpleMachines/SMF2.1/pull/3428
- the other one is that my sql commands (ALTER from line 1956 - 1967) are missing
- it's look like i don't need to call db_free_result after a dml command 

the second error should be a task from smf team @Oldiesmann  @live627  or someone else...

the sql commands needs to bo done before my MySQLConvertOldIp function is executed.

albertlast:
problem one and three are fixed,
[...]

PR #3433, opened 2016-05-05
harded  MySQLConvertOldIp

albertlast:
now with less potential to fail and more debug infos
could fix the issue: https://github.com/SimpleMachines/SMF2.1/issues/3427


albertlast:
jdarwood aproach looks better: https://github.com/SimpleMachines/SMF2.1/pull/3436

PR #3455, opened 2016-05-21
Stuff

MissAllSunday:
Some other random fixes

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -1947,15 +1947,15 @@ function loadTheme($id_theme = 0, $initialize = true)
 	);
 
 	$simpleAreas = array(
-		'popup',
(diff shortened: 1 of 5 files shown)


Sources/Load.php
```diff
@@ -1947,15 +1947,15 @@ function loadTheme($id_theme = 0, $initialize = true)
 	);
 
 	$simpleAreas = array(
-		'popup',
-		'alerts_popup',
+		'profile' => 'popup',
+		'profile' => 'alerts_popup',
```
jdarwood007, line 1951: That won't work.  You defined the same key twice.

MissAllSunday, line 1951: ah yes.... then we will have to switch those...

PR #3464, opened 2016-05-26
Optionally pass BBC params to validate function

Sesquipedalian:
- If a BBCode's validate function is written to accept a fourth parameter, it can receive the final, calculated values of the $params array. This is especially useful if the values the user entered for the BBCode's parameters are required in the validate function.
- If a BBCode's validate function does not expect the fourth parameter, it will of course be ignored by PHP. So this is fully backwards compatible.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -2253,7 +2253,7 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 				$data = substr($data, 4);
 
 			if (isset($tag['validate']))
-				$tag['validate']($tag, $data, $disabled);
+				$tag['validate']($tag, $data, $disabled, $params);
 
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
FYI, this is in preparation for a PR to add support for image parameters (e.g. width, height) to the attach BBCode. I'm submitting these changes in two separate PRs because passing the BBCode parameters to the validate function is a significant change in its own right. It opens a lot of new possibilities for what can be done with BBCode. I expect that mod authors will find these new possibilities quite intriguing.

MissAllSunday:
Is this ready for merging @Sesquipedalian ?

Sesquipedalian:
Yes. Once it is merged, I will submit a separate pull request for the attachment params code that builds on this PR

MissAllSunday:
cool, thanks!

PR #3465, opened 2016-05-26
Avoid id conflicts between mobile and desktop versions of main menu

### LINKED ISSUE #3463: many duplicate ID's give error on validate html
Because the mobile menu is loaded allways (even it's not a mobil device) on Validation it gives many, many errors on duplicate id's.
Whey not check (with php) if a mobile device detected ?
It's very simple to do that .. I can spend the code for that ;-)

Sesquipedalian:
Fixes #3463

### DIFF
--- a/Themes/default/index.template.php
+++ b/Themes/default/index.template.php
@@ -327,7 +327,7 @@ function template_body_above()
 					<div class="popup_window description">
 						<div class="popup_heading">', $txt['mobile_user_menu'],'
 						<a href="javascript:void(0);" class="generic_icons hide_popup"></a></div>
-						', template_menu(), '
+						', template_menu(true), '
 					</div>
 				</div>';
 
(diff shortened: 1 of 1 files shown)


XinYenFon:
I think its better solution to remove IDs and use classes. Too many new phones coming out to market, I'm not really sure if mobile-detector can hold up with it. Also Curve2 is not a mobile theme, its a responsive theme so solutions should be at responsive stage.

Sesquipedalian:
Removing ids could work too, yes. Even so, this solution is a responsive one rather than a browser detection solution.

XinYenFon:
@Sesquipedalian https://github.com/Antes/kittenground/commit/d04c4b041beec6b08453662e7ca2515416550e39

Sesquipedalian:
That's fine with me. Go ahead and submit that as a PR and close this one.

PR #3467, opened 2016-05-28
Stuff

MissAllSunday:
This PR fixes a few mistakes and bugs I introduced on my previous PR.

It also introduces a couple of features:
- Implements smf_json_decode()  which is a wrapper for json_decode, has error handler and also makes sure to always return an array to avoid SMF breaking due to malformed json data.
- function smc_resize()  a small function to make images with the bbc_img class responsive, a live example can be found at my site: http://missallsunday.com/ the Cinco de Mayo topic has a large image and it gets automatically resized to fit the current screen size.  The function ca be used on any element/selector.

### DIFF
--- a/Sources/Admin.php
+++ b/Sources/Admin.php
@@ -39,7 +39,7 @@ function AdminMain()
 	require_once($sourcedir . '/Subs-Menu.php');
 
 	// Some preferences.
-	$context['admin_preferences'] = !empty($options['admin_preferences']) ? json_decode($options['admin_preferences'], true) : array();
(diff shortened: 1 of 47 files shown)


XinYenFon:
I'm curious about smc_resize, bbc_image already have max-width: 100%. How this can be bypassed?

MissAllSunday:
It won't be bypassed or take into account.

XinYenFon:
I just can't figure out where smc_resize helps us. Considering bbc_img already applied to all img tags.

PR #3468, opened 2016-05-29
Too many guesses and reminder broken in login pop-up.

the-real-blackh:
Two types of errors during login were not handled properly:
1. fatal_lang_error messages such as 'failed_login_threshold' (too many password guesses).
2. The reminder page if you guessed the wrong password too many times.
In both cases it was just jumping back to the home page, leading the user to think they had successfully logged in - but they have no privileges.
We fix this by making it handle these cases better in handling the response to the ajax request.

### DIFF
--- a/Themes/default/Login.template.php
+++ b/Themes/default/Login.template.php
@@ -76,14 +76,23 @@ function template_login()
 								method: "POST",
 								data: form.serialize(),
 								success: function(data) {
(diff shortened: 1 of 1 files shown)


Oldiesmann:
You need to sign off on your commits or we can't accept them.

the-real-blackh:
Sorry... re-submitting.

PR #3472, opened 2016-05-29
Two types of errors during login were not handled properly:

the-real-blackh:
1. fatal_lang_error messages such as 'failed_login_threshold' (too many
   password guesses).
2. The reminder page if you guessed the wrong password too many times.
   In both cases it was just jumping back to the home page, leading the user
   to think they had successfully logged in - but they have no privileges.
   We fix this by making it handle these cases better in handling the
   response to the ajax request.
   Signed-off-by: Stephen Blackheath docks.cattlemen.stephen@blacksapphire.com

### DIFF
--- a/Themes/default/Login.template.php
+++ b/Themes/default/Login.template.php
@@ -78,14 +78,23 @@ function template_login()
 								method: "POST",
 								data: form.serialize(),
 								success: function(data) {
-									if (data.indexOf("<bo" + "dy") > -1)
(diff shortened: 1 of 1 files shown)


the-real-blackh:
To reproduce 1: enter the wrong password 10 times or more so you lock the user out.
To reproduce 2: enter the wrong password 3 times (this number depends on configuration).

MissAllSunday:
Thank you for your contribution! :smile:

PR #3473, opened 2016-05-30
Minify adjustments

jdarwood007:
Using methods with the same name as their classes is needed for PHP 4 so this can be removed as it is depurated in PHP 7.
Fix FTP connection needing a login token on upgrader.
Require upgrader to make sure minified files can be wrote to.
Have SMF retry css/js processing if it can't minify, to prevent broken forums

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -3213,8 +3213,9 @@ function template_footer()
  *  - if defered is set function will output all JS (source & inline) set to load at page end
  *
  * @param bool $do_deferred If true will only output the deferred JS (the stuff that goes right before the closing body tag)
+ * @param bool $skip_minify If true will skip minify process
  */
-function template_javascript($do_deferred = false)
+function template_javascript($do_deferred = false, $skip_minify = false)
 {
 	global $context, $modSettings, $settings;
 
@@ -3256,7 +3257,7 @@ function template_javascript($do_deferred = false)
 			continue;
 
 		// By default all files don't get minimized unless the file explicitly says so!
-		if (!empty($js_file['options']['minimize']) && !empty($modSettings['minimize_files']))
+		if (empty($skip_minify) && !empty($js_file['options']['minimize']) && !empty($modSettings['minimize_files']))
 		{
 			if ($do_deferred && !empty($js_file['options']['defer']))
 				$toMinifyDefer[] = $js_file;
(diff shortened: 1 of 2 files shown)


jdarwood007:
So I upgraded another one of my many test forums to the latest 2.1 and it broke afterwards due to being unable to write minify files.  The adjustments to minify process here has it bail out of minify and load regular files if it can't write.

This isn't checking to see if it can minify for a single theme only or not.  It simply bails out.  Somebody can work that in but it requires some rewrites of the minify process.

This makes the process of minify much safer as the minify process failing causes the forum to be unusable.

live627:
Someone should really merge this.

Sources/Subs.php
```diff
@@ -3438,7 +3448,7 @@ function custMinify($data, $type, $do_deferred = false)
 		{
 			loadLanguage('Errors');
 			log_error(sprintf($txt['file_not_created'], $toCreate), 'general');
-			continue;
+			return false;
```
MissAllSunday, line 3451: How will this  behave with multiple themes?

Lets say there are 3 active themes and the second one fails, with a return the third one will never be processed.

other/upgrade.php
```diff
@@ -1042,6 +1036,9 @@ function WelcomeLogin()
 	$writable_files = array(
 		$boarddir . '/Settings.php',
 		$boarddir . '/Settings_bak.php',
+		$boarddir . '/db_last_error.php',
+		$modSettings['theme_dir'] . '/css/minified.css',
+		$modSettings['theme_dir'] . '/scripts/minified.js',
```
MissAllSunday, line 1041: The minify process also creates a  minified_defer.js  file

jdarwood007:
With multiple themes based on how I understand the minify code works, it would fail for the current theme?  I believe it rolls all css into 1 file even for multiple themes?

We can add a minified_defer.js if needed.  The fix at least makes the forum usable if minify process fails.  What defer items are broken by this?

If this PR isn't merged, we need to resolve this as it is a forum breaking issue that holds up the release as when the minify process fails the forum is not usable (no css or js).

MissAllSunday:
On Fri, 10 Jun 2016 19:27:48 -0500, Jeremy D notifications@github.com  
wrote:

> With multiple themes based on how I understand the minify code works, it  
> would fail for the current theme? I believe it rolls all css into 1 file  
> even for >multiple themes?
> No, with multiple themes every theme gets a minify file that contains all  
> files that particular theme uses and has been labeled for minification.  
> Some files are grabbed from the default theme.
> 
> We can add a minified_defer.js if needed. The fix at least makes the  
> forum usable if minify process fails. What defer items are broken by  
> this?

None, I was solely talking about adding that file to that list.

> If this PR isn't merged, we need to resolve this as it is a forum  
> breaking issue that holds up the release as when the minify process  
> fails the forum is not >usable (no css or js).

Well yes, thing is, support for multi themes is a must, that needs to be  
resolved.

jdarwood007:
I'm not sure what to do for multiple theme support.  The main part of this code ties into the already existing part of the minify process that reports an error when it can't minify and returns in a way that we know it failed and can exit safely.

MissAllSunday:
I've been thinking about moving the echo part into the function so there's no need to return any value and we can still have control over the way the minimized files are created and we still have control if the process fails and make sure there's a safe fallback

jdarwood007:
Well if you got a plan then we can hold off on this PR or merge this and then you commit changes over it.

MissAllSunday:
Will do when I get home

MissAllSunday:
So yeah that return needs to be changed back to a continue, multiple themes support is currently broken.

jdarwood007:
I was going to say that the minify process didn't really handle multiple themes properly from what I could tell.  But I see your changes and that should work.

Issue #3475, opened 2016-06-01
ssi_examples.php updates


illori:
minor issue in ssi_examples.php there is a link to the old docs site... http://docs.simplemachines.org/index.php?topic=400.0 it does redirect to the wiki, but maybe we should update the link? also http://www.simplemachines.org/community/index.php?topic=14906.0 could be removed as the links it contains are on the page on the wiki.

should link to http://wiki.simplemachines.org/smf/Category:SSI

also minor issue in ssi_examples.shtml the copyright year in the footer is incorrect and it does not tell you when the file was last modified, it has the text but no date.

illori:
also on ssi_examples.php there is a question mark after advanced functions, it does not seem to function.

[edit] it seems if you hover over it, you do get some text but it is not apparent that you need to hover as in SMF you would get a popup when clicking it.

illori:
if you view the sample 1 website the code is all showing in one line instead of a large code block, this makes it hard to read and it looses its formatting.

XinYenFon:
Fixed some stuff in SSI-examples, that popup will be too much. I also suggest removing sample1 and offer more powerful things at wiki. Its already way too simple and not useful for most.

illori:
if you have suggestions to add more to the wiki feel free to create a page or open a topic on it.

XinYenFon:
Anything left/forgot ?

illori:
the code block for sample 1 is still one line.

XinYenFon:
I think its best to remove those very old samples out and write something more modern to Wiki and close this :)

PR #3486, opened 2016-06-14
small fixes

XinYenFon:
This should improve print page little bit (also fixes quote looking). Did some changes on SSI based on #3475

cc: @illori 

and yes I forgot to open this PR :P


illori:
I was wondering where the PR had gone.

PR #3502, opened 2016-06-21
The Admin preferences in the themes table are not updated

PortaMx:
Signed-off-by: Feline support@portamx.com


MissAllSunday:
And yet, this isn't a good way to solve this....

The upgrade process ALREADY handles serialize to json process.  If the admin preferences aren't being converted during the upgrade process then the issues is on the upgrade process.


Lets try to fix the root issue rather than trying to patch its consequences.

PR #3513, opened 2016-07-04
Methods with the same name as their class will not be constructors in…

jdarwood007:
… a future version of PHP


MissAllSunday:
Shouldn't it be renamed to __construct  ?

jdarwood007:
This was a PHP4 compatibility item.  If you look, it just calls the connect method directly.  We could change it to construct if we want, or leave it to where we call the new class and then have to issue the connect method.

We dropped this from the upgrade.php before, this got missed.

MissAllSunday:
OK I see it now.

PR #3527, opened 2016-07-19
Proxy servers bug fix, step 1

### LINKED ISSUE #3449: Proxy servers question
My forum is behind a proxy server mainly for ddos protection.

The $_SERVER var contains both HTTP_X_FORWARDED_FOR and HTTP_X_REAL_IP for the user ip. In the admin security settings I've tried both of them for the select box and filled in the proxy IP, yet when I check a user profile I still get the proxy server IP for their ip address. Am I doing something wrong?

asmith20002:
Fixes #3449

Tested on a live server behind proxy. (by the looks of the bug, it seems it was never tested before)

Now SMF successfully records member's real IP into member_ip2 column in member's table. However the member_ip1 still shows the proxy server IP which should be discussed further. (hence "step 1" in the title)

Signed-off-by: Alex Smith asmith_20002@yahoo.com

### DIFF
--- a/Sources/ManageServer.php
+++ b/Sources/ManageServer.php
@@ -406,7 +406,7 @@ function ModifyGeneralSecuritySettings($return_config = false)
 		'',
 			array('select', 'frame_security', array('SAMEORIGIN' => $txt['setting_frame_security_SAMEORIGIN'], 'DENY' => $txt['setting_frame_security_DENY'], 'DISABLE' => $txt['setting_frame_security_DISABLE'])),
 		'',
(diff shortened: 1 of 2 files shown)


Rupurudu:
We should keep member_ip1 as proxy server's IP. Otherwise, people can fake their IP with using these headers.

asmith20002:
It's been a month since I committed this change. Any feedback why it's not getting merged? Seems like nobody has even reviewed. (because the bugs are so obvious)

MissAllSunday:
I did, I'm just not sure about the last part.

asmith20002:
Which last part?

MissAllSunday:
This one: https://github.com/SimpleMachines/SMF2.1/pull/3527/files#diff-94ffaea296841584d2e4a6cfe8611aa0R252

asmith20002:
The `isset()` in this line:
`if (isset($modSettings['proxy_ip_servers']))`

is replaced with `!empty()` to prevent warning.

This line:
`// Ignore if this is not set.
if (!isset($_SERVER[$proxyIPheader]))
       continue;`

is also there to prevent warning. I was testing this on a live server and selected a header which wasn't there and it started to clog my error log pretty bad.

PR #3530, opened 2016-07-31
The calendar now supports event times

Sesquipedalian:
Lots of inter-related changes in this, including:
- Database changes (adds three new columns to calendar table)
- Subs-Db-\* can now accept and handle data for SQL TIME columns
- A bunch of new and reorganized code (especially in Subs-Calendar.php)
- Rewritten UI for creating/editing events in Calendar.template.php and Post.template.php, plus corresponding CSS changes
- New JavaScripts to keep event editing UI both pleasant and sane
- Changes to calendar display templates (HTML & CSS) to incorporate event time info without looking like a mess
- iCal export can now handle both all day events and events with time and timezone data
- The time zone select menu has been improved. It is now always up-to-date and accurate, and the admin can choose to show the time zones from a given country at the top of the list if desired.

Backwards compatibility with the previous system has been maintained in two ways:
- Existing events in the database are handled as expected. No updates to existing data are required.
- The code accepts multiple input formats from $_POST, including full support for the format used in SMF 2.0 (i.e. year, month, day, and optional span parameters). So any old mods (or imperfectly updated themes) that rely on this behaviour will continue to work.

Themes and mods can now enjoy a great deal of flexibility in how they choose to supply date and time info for events.
- Year, month, day, hour, minute, and second can be given as individual parameters, as a date string and a time string, or as single datetime strings. The default theme for SMF 2.1 uses date strings and time strings.
- Input can either specify a start date (and optional time) and end date (and optional time), or a start date (and optional time) plus a 'span' parameter to indicate the duration of the event in days. If both end date/time parameters and and a span parameter are provided, the span parameter is ignored.
- If time info is provided, a timezone parameter should also be supplied. If no timezone is given, the default timezone will be used.
- The 'allday' parameter can (and should) be used to specify whether an event does/should have start and end times. If 'allday' is given a value equivalent to true, any time and timezone parameters will be discarded. If false, time info will be used as supplied (or calculated if required and possible). If 'allday' is not set, the code will attempt to infer the best course of action based on whatever parameters have been provided.
- The default implementation always uses ISO format (YYYY-MM-DD) for entering/displaying dates in the UI input fields, regardless of the user's preferred format. This is mostly for the sake of some of the JavaScript, whose date formatting abilities are much weaker than PHP's. Apart from keeping the JS happy, there's no reason the date fields couldn't use any format that PHP understands.

Speaking of JavaScript, three new ones have been included: ~~pikaday.js~~ jquery-ui.datepicker.js, jquery.timepicker.js, and datepair.js. The first two are a date picker and time picker (obviously). The third coordinates the start and end dates/times, which both serves as a first-line defense against user error (end dates before start dates, etc.) and provides some creature comforts.

Displaying time information on the calendar required some rearrangement of elements in order to make space without looking like a mess. Probably more needs to be done to make it look really good, but this is passable. If @Antes can think of ways to improve it further, that's fine by me.

For the record, dealing with calendars is a royal pain in the ass.


albertlast:
i'm wonderd that you only need one timezone for start AND end.
I would expect that both got there own.

Sources/Subs-Db-postgresql.php
```diff
@@ -202,6 +202,13 @@ function smf_db_replacement__callback($matches)
 				smf_db_error_backtrace('Wrong value type sent to the database. Date expected. (' . $matches[2] . ')', '', E_USER_ERROR, __FILE__, __LINE__);
 		break;
 
+		case 'time':
+			if (preg_match('~^([0-1]?\d|2[0-3]):([0-5]\d):([0-5]\d)$~', $replacement, $time_matches) === 1)
+				return sprintf('\'%02d:%02d:%02d\'', $time_matches[1], $time_matches[2], $time_matches[3]);
```
albertlast, line 207: I would like to see here: TO_TIMESTAMP('15:15:20', 'HH24:MI:SS')::TIME
maybe with a different pattern but something in this direction.

possible pattern are here: https://www.postgresql.org/docs/9.5/static/functions-formatting.html

Sources/Subs-Db-mysql.php
```diff
@@ -214,6 +214,13 @@ function smf_db_replacement__callback($matches)
 				smf_db_error_backtrace('Wrong value type sent to the database. Date expected. (' . $matches[2] . ')', '', E_USER_ERROR, __FILE__, __LINE__);
 		break;
 
+		case 'time':
+			if (preg_match('~^([0-1]?\d|2[0-3]):([0-5]\d):([0-5]\d)$~', $replacement, $time_matches) === 1)
+				return sprintf('\'%02d:%02d:%02d\'', $time_matches[1], $time_matches[2], $time_matches[3]);
```
albertlast, line 221: STR_TO_DATE is the similiar function like pg : https://dev.mysql.com/doc/refman/5.5/en/date-and-time-functions.html#function_str-to-date
pattern find you there: https://dev.mysql.com/doc/refman/5.5/en/date-and-time-functions.html#function_date-format

Sources/Subs-Db-mysqli.php
```diff
@@ -238,6 +238,13 @@ function smf_db_replacement__callback($matches)
 				smf_db_error_backtrace('Wrong value type sent to the database. Date expected. (' . $matches[2] . ')', '', E_USER_ERROR, __FILE__, __LINE__);
 		break;
 
+		case 'time':
+			if (preg_match('~^([0-1]?\d|2[0-3]):([0-5]\d):([0-5]\d)$~', $replacement, $time_matches) === 1)
+				return sprintf('\'%02d:%02d:%02d\'', $time_matches[1], $time_matches[2], $time_matches[3]);
```
albertlast, line 251: STR_TO_DATE is the similiar function like pg : https://dev.mysql.com/doc/refman/5.5/en/date-and-time-functions.html#function_str-to-date
pattern find you there: https://dev.mysql.com/doc/refman/5.5/en/date-and-time-functions.html#function_date-format

Sources/Subs-Db-postgresql.php
Sesquipedalian, line 207: I don't understand your suggestion. The `smf_db_replacement__callback()` function validates and formats the values using PHP before anything is passed to the database. If you want to use SQL functions for that job, that would need to happen at a different, later stage in the process.

Sesquipedalian, line 207: ... Or do you mean that instead of returning a string like `15:15:20` that it should instead return `TO_TIMESTAMP('15:15:20', 'HH24:MI:SS')::TIME`?

Sesquipedalian:
> i'm wonderd that you only need one timezone for start AND end.
> I would expect that both got there own.

I thought about that, but decided not to implement it because:
1. It would very rarely be useful (scheduling plane flights, and that's about it).
2. It would exponentially complicate the code.
3. It would clutter the interface to provide the option.

If a user ever feels the need to enter plane flight information onto their forum's calendar, they can just figure out the time zone difference themselves.

Sources/Subs-Db-postgresql.php
albertlast, line 207: Sure otherwise is it "luke" That the rdbms reconize as time and in the right format.

---

Gesendet: 31.07.2016 20:53
An: SimpleMachines/SMF2.1
Cc: albertlast; Comment
Betreff: Re: [SimpleMachines/SMF2.1] The calendar now supports event times (#3530)

> @@ -202,6 +202,13 @@ function smf_db_replacement__callback($matches)
>                               smf_db_error_backtrace('Wrong value type sent to the database. Date expected. (' . $matches[2] . ')', '', E_USER_ERROR, **FILE**, **LINE**);
>               break;
> -             case 'time':
> -                     if (preg_match('~^([0-1]?\d|2[0-3]):([0-5]\d):([0-5]\d)$~', $replacement, $time_matches) === 1)
> -                             return sprintf('\'%02d:%02d:%02d\'', $time_matches[1], $time_matches[2], $time_matches[3]);

... Or do you mean that instead of returning a string like `15:15:20` that it should instead return `TO_TIMESTAMP('15:15:20', 'HH24:MI:SS')::TIME`?

---

You are receiving this because you commented.
Reply to this email directly or view it on GitHub:
https://github.com/SimpleMachines/SMF2.1/pull/3530/files/b7128f146b6bf1a9673095524c7ff7f004466d1c#r72908810

Sesquipedalian, line 207: Well, according to [this](https://www.postgresql.org/docs/9.5/static/datatype-datetime.html#DATATYPE-DATETIME-INPUT), it is perfectly fine to give date and time literal input.

If you still think it is a problem, you will probably also want to make similar changes for all the other data types that `smf_db_replacement__callback()` handles as well. If so, it would make more sense to implement all those changes at once in a separate PR. 

But as I said, I really don't see the need. The current system already does things in a valid way.

albertlast, line 207: Sure the database can try to handle this,
but this is a sign of good programming to give the database a hint in which format your
time or data are.

you already find such solution for inet: '\'%1$s\'::inet'
for string and numbers is not needed such solution.

XinYenFon:
Well I spotted some stuff,
- SQL changes are in related files but when you run upgrade.php its not adding them.
- I can't select an end day for event?
- Time cannot be limited (well surely it needs to be between 00:00 to 23:59 (or AM/PM)).
- Mobile users can't see post event on calendar

Unfortunately I don't have much time to look further for now. But one thing last thing I want to mention, in post event page the header part has 2 style (one with dl/dd/dt which we use in core, one with divs which you introduced) we need to decide which needs to go and keep the styling in template as one, so we don't have to write more code to make them responsive as well. Code (or style) needs to have a standard (or base whichever you call it).

Sesquipedalian:
> SQL changes are in related files but when you run upgrade.php its not adding them.

Hm. Okay. I'm not sure why that would be. Who should I talk to for advice on how to fix that?

> I can't select an end day for event?

The "Allow events to span multiple days" (i.e. `$modSettings['cal_allowspan']`) setting needs to be activated in the Calendar settings. 

However, that brings up another question. The `cal_allowspan` setting is useless and redundant now. The same effect can be accomplished by setting `cal_maxspan` to a value of 1. So, perhaps `call_allowspan` should be removed during the upgrade process. Thoughts, anyone?

> Time cannot be limited (well surely it needs to be between 00:00 to 23:59 (or AM/PM)).

It's supposed to be. Sounds like something is going wrong with the timepicker JavaScript. I'll see if I can reproduce the problem

> Mobile users can't see post event on calendar

I think that was already like that, but I'll commit something to fix the issue.

> dl/dd/dt

I think that can be changed easily enough. :)

XinYenFon:
Instead using extra classes use current ones. For example if you need to float we already have floatleft / floatright. Also you can embed the new parts into "post_header" probably save you dozens of problems (styling).

XinYenFon:
```
Warning: timezone_open(): Unknown or bad timezone () in C:\wamp\www\kittenground\Sources\Subs-Calendar.php on line 148

Warning: date_create() expects parameter 2 to be DateTimeZone, boolean given in C:\wamp\www\kittenground\Sources\Subs-Calendar.php on line 148

Warning: timezone_open(): Unknown or bad timezone () in C:\wamp\www\kittenground\Sources\Subs-Calendar.php on line 149

Warning: date_create() expects parameter 2 to be DateTimeZone, boolean given in C:\wamp\www\kittenground\Sources\Subs-Calendar.php on line 149

Warning: date_format() expects parameter 1 to be DateTimeInterface, boolean given in C:\wamp\www\kittenground\Sources\Subs-Calendar.php on line 157

Warning: date_format() expects parameter 1 to be DateTimeInterface, boolean given in C:\wamp\www\kittenground\Sources\Subs-Calendar.php on line 158

Warning: date_timezone_set() expects parameter 1 to be DateTime, boolean given in C:\wamp\www\kittenground\Sources\Subs-Calendar.php on line 160

Warning: date_timezone_set() expects parameter 1 to be DateTime, boolean given in C:\wamp\www\kittenground\Sources\Subs-Calendar.php on line 161

Warning: date_time_set() expects parameter 1 to be DateTime, boolean given in C:\wamp\www\kittenground\Sources\Subs-Calendar.php on line 162

Warning: date_time_set() expects parameter 1 to be DateTime, boolean given in C:\wamp\www\kittenground\Sources\Subs-Calendar.php on line 163

Warning: date_format() expects parameter 1 to be DateTimeInterface, boolean given in C:\wamp\www\kittenground\Sources\Subs-Calendar.php on line 165

Warning: date_format() expects parameter 1 to be DateTimeInterface, boolean given in C:\wamp\www\kittenground\Sources\Subs-Calendar.php on line 166
```

Enabled SMF's debug and saw this on ?action=calendar / topic view.

Also what I have in my mind is this, literally zero extra css usage. Not saying what you did wasn't better, I like the way you did it. But we need to keep consistency and we need to release the product at some point, if we change dl usage into fieldset we need to change all, which I'm sure you understand that's something we want to avoid for this version. Do mind I'll change the New Poll page as well, because its weird as it is first input is empty is not the way it needs to be.

screenshot-kittenground 2016-08-02 14-07-49

Snip I used

```
    // Are you posting a calendar event?
    if ($context['make_event'])
    {
        echo '
                        <hr class="clear">
                        <dt>
                            <span', isset($context['post_error']['no_event']) ? ' class="error"' : '', '>', $txt['calendar_event_title'], '</span>
                        </dt>
                        <dd>
                            <input type="text" id="evtitle" name="evtitle" maxlength="255" size="55" value="', $context['event']['title'], '" tabindex="', $context['tabindex']++, '" class="input_text">
                        </dd>';
        // If this is a new event let the user specify which board they want the linked post to be put into.
        if ($context['event']['new'] && $context['is_new_post'])
        {
            echo '
                        <dt>
                            ', $txt['calendar_post_in'], '
                        </dt>
                        <dd>
                            <select name="board">';
                foreach ($context['event']['categories'] as $category)
                {
                    echo '
                                        <optgroup label="', $category['name'], '">';
                    foreach ($category['boards'] as $board)
                        echo '
                                            <option value="', $board['id'], '"', $board['selected'] ? ' selected' : '', '>', $board['child_level'] > 0 ? str_repeat('==', $board['child_level'] - 1) . '=&gt;' : '', ' ', $board['name'], '&nbsp;</option>';
                    echo '
                                        </optgroup>';
                }
                echo '
                                    </select>
                        </dd>
                        <hr class="clear">
                        <h3>', $txt['calendar_event_options'], '</h3>
                        <dt>
                            ', $txt['start'], '
                        </dt>
                        <dd>
                            <input type="text" name="start_date" id="start_date" maxlength="10" value="', $context['event']['start_date'], '" tabindex="', $context['tabindex']++, '" class="input_text date_input start" data-type="date">
                            <input type="text" name="start_time" id="start_time" maxlength="11" value="', $context['event']['start_time'], '" tabindex="', $context['tabindex']++, '" class="input_text time_input start" data-type="time"', !empty($context['event']['allday']) ? ' disabled' : '', '>
                        </dd>
                        <dt>
                            ', $txt['end'], '
                        </dt>
                        <dd>
                            <input type="text" name="end_date" id="end_date" maxlength="10" value="', $context['event']['end_date'], '" tabindex="', $context['tabindex']++, '" class="input_text date_input end" data-type="date"', (empty($modSettings['cal_allowspan']) || $modSettings['cal_maxspan'] == 1) ? ' disabled' : '', '>
                            <input type="text" name="end_time" id="end_time" maxlength="11" value="', $context['event']['end_time'], '" tabindex="', $context['tabindex']++, '" class="input_text time_input end" data-type="time"', !empty($context['event']['allday']) ? ' disabled' : '', '>
                        </dd>
                        <dt>
                            <label for="allday">', $txt['calendar_allday'], '</label>
                        </dt>
                        <dd>
                            <input type="checkbox" name="allday" id="allday"', !empty($context['event']['allday']) ? ' checked' : '', ' tabindex="', $context['tabindex']++, '">
                        </dd>
                        <dt>
                            ', $txt['calendar_timezone'], '
                        </dt>
                        <dd>
                            <select name="tz" id="tz"', !empty($context['event']['allday']) ? ' disabled' : '', '>';

            foreach ($context['all_timezones'] as $tz => $tzname)
                echo '
                                <option value="', $tz, '"', $tz == $context['event']['tz'] ? ' selected' : '', '>', $tzname, '</option>';

            echo '
                            </select>
                        </dd>
                    </dl>';
        }
    }
    else
        echo '
                    </dl>';
```

edit: typo :(

Sesquipedalian:
Okay. The reason for the fieldsets was that the existing templates already used fieldsets, divs, and uls. I was trying to stick as closely as I could to what was already there. If you wanted to rip out the fieldset/div/ul structure in favour of dl/dt/dd anyway, that makes sense me. :)

One thing, though, is that the start and end date and time fields need to be contained within a single wrapper of some sort for the JavaScript to work correctly. It might work if we tell the JS to treat the entire `#post_header` itself as that wrapper, but if not, we'll have to nest the start & end section somehow.

XinYenFon:
Ok then keep it as is :)

Sesquipedalian:
> Ok then keep it as is :)

Fair enough. 

> [debug errors]

I'm trying to reproduce that, but so far I can't. Is there some other debug function besides adding `$db_show_debug = true;` to Settings.php that I should know about? Also, you mentioned before that the SQL upgrade didn't add the columns to the calendar table in the database. Did you manually add those yourself in the meantime?

XinYenFon:
no, that's the debug func I'm using as well. yes I added tables manually (otherwise site stops working).

Sesquipedalian:
What result do you get if you replace line 148 of Subs-Calendar.php with this?

```
if (!empty($row['start_time']) && !empty($row['end_time']) && !empty($row['timezone']))
```

Sources/Subs-Db-postgresql.php
Sesquipedalian, line 207: Okay, I tried implementing this, but it produced error messages. It is possible that I implemented it incorrectly, of course. But for now I am just going to leave it as it is. If you want to try implementing this change yourself in another PR after this PR is accepted, @albertlast, that would be fine with me. :)

Sesquipedalian:
> SQL changes are in related files but when you run upgrade.php its not adding them

I can't figure out why this is happening—or rather, why it isn't happening. I guess I don't understand the upgrade process very well yet. Perhaps @MissAllSunday, @Oldiesmann, @live627, @jdarwood007, @albertlast, or someone else could educate me (or at least show me what I've missed)?

other/upgrade_2-1_mysql.sql
```diff
@@ -2441,3 +2441,10 @@ if ($doChange)
 ---# Add the new log banned ip
 ALTER TABLE {$db_prefix}log_online ADD COLUMN ip VARBINARY(16);
 ---#
+
+--# Add start_time and end_time support to the calendar
```
live627: you need three dashes, not two

https://github.com/SimpleMachines/SMF2.1/wiki/SQL-file-for-upgrade

Sesquipedalian: Ah-ha! I figured it was some silly mistake on my part. Thanks, @live627.

live627: np

Sources/Subs-Db-postgresql.php
albertlast, line 207: Is there any place in smf where the time format is defined?
Or is the format everytime HH24:MM:SS regardless what is configured?

Sesquipedalian, line 207: > Is there any place in smf where the time format is defined?

This if statement is that definition. The regular expression `^([0-1]?\d|2[0-3]):([0-5]\d):([0-5]\d)$` will only match valid times given in 24 hour format with both minutes and seconds defined. If the input value matches, the sprintf function rewrites it to ensure that it is in exactly the correct format. The return value is therefore absolutely guaranteed to be a time literal in HH24:MM:SS format.

| Input | Result |
| --- | --- |
| 19:45:32 | '19:45:32' |
| 03:30:00 | '03:30:00' |
| 3:30:00 | '03:30:00' |
| 23:15 | error |
| 23:15:00 | '23:15:00' |
| 10:30 AM | error |
| 32:30:00 | error |
| 13:68:21 | error |
| 13:22:99 | error |
| 2016-08-03 | error |
| foostring | error |

As you can see, the regular expression is willing to tolerate input where the hour leaves off the leading zero, but the sprintf function will always add the missing zero if it is not there.

albertlast, line 207: This was not my intention in my question,
the user is able to customize the time format:
timeformat

And my question how react when the format is not hh24:mm:ss ?
Or did you ignore this setting and work all the time with hh24:mm:ss?

Sesquipedalian, line 207: That setting just controls how times are shown to the user. The code internally uses HH24:MM:SS for everything. 

Here is the flow:
1. The user enters a time on the form in any format that PHP understands. For example, `1:30 PM`.
2. The `setEventStartEnd()` function parses the input and converts it to HH24:MM:SS format. For example, `13:30:00`
3. The `insertEvent()` (or `modifyEvent()`) function sends `13:30:00` to the appropriate $smc function for insertion into the database.
4. The `smf_db_replacement__callback()` function checks the time string again to make sure it is in HH24:MM:SS format, and rewrites it in that format again, just to be extra sure.
5. The SQL call inserts the value into the database.

When the user wants to view an event, the HH24:MM:SS time string is retrieved from the database and reformatted for display using the setting you mentioned.

Sesquipedalian:
> However, that brings up another question. The `cal_allowspan` setting is useless and redundant now. The same effect can be accomplished by setting `cal_maxspan` to a value of 1. So, perhaps `call_allowspan` should be removed during the upgrade process. Thoughts, anyone?

I have gone ahead and implemented this change. That setting is used nowhere outside of the code I am dealing with here, so I see no reason not to make the change. I suppose that if there are any mods that depended on this setting in some way, those mods will need to be updated, but mod authors should already expect that they will need to update their mods for 2.1 anyway, so meh.

Sources/Subs-Db-postgresql.php
albertlast, line 207: good than i can place this coding after your commit

Sesquipedalian:
> Time cannot be limited (well surely it needs to be between 00:00 to 23:59 (or AM/PM)).

Are you still seeing this, @Antes? Expected behaviour is:
1. If start date and end date inputs have the same value, the end time input will be limited to a range between whatever time is in the start time input and 23:59:59.
2. If the end date is some date after the start date, the end time will offer a full 24 hour range.
3. Trying to select an end date before the start date will cause the start date to change, too. If necessary, the start time will also be changed to make sure it is before the end time.
4. Changing the start date and/or start time will automatically change the end date and/or end time in order to preserve the current duration of the event.

albertlast:
One question,
why you use 4 fields for time (two for end/start date and two for end/start time)?
i would expected that here only two fields are needed (datetime start, datetime end).

Sesquipedalian:
If the user checks the "All day" input, we need to disable the time input. One could use JavaScript to change what a single input field will accept, but it is a lot easier to just toggle the "disabled" property on the time input field. Plus, the JavaScripts that I use to make the UI nicer all assume that date and time are separate inputs.

However, if a mod or theme wants to implement another UI that uses two datetime input fields, that will be fine. I wrote the backend code to be able to handle input in that format.

albertlast:
I would say "all day" is start date + 0 o'clock and end date + 23:59:59

here some points versus this:
- database index is easier when this only one field
- input != database side

Sesquipedalian:
Oh, I thought you were talking about the user interface, not the database. Disregard my previous answer.

There reason why I chose to add start_time and end_time as separate database columns is to maintain backwards compatibility. Doing it this way means:
1. Existing events in the database will continue to work as intended, without any changes during an upgrade
2. Existing mods that expect start_date and end_date to be dates rather than datetimes will continue to work

If I were building a completely new system, I probably would not have architected it the way it is now. But I had to work with what was already there, and this was the cleanest way to extend the existing functionality without breaking anything.

XinYenFon:
I can't set time for event. It still takes it as "all day" event. There is a problem with saving start/end time.

Sesquipedalian:
@Antes, could you please run this SQL query for me and tell me the result?

```
select *
from INFORMATION_SCHEMA.COLUMNS
where TABLE_NAME='smf_calendar'
```

XinYenFon:
```
<?xml version="1.0" encoding="utf-8"?>
<!--
- phpMyAdmin XML Dump
- version 4.5.2
- https://www.phpmyadmin.net
-
- Host: 127.0.0.1
- Generation Time: Aug 22, 2016 at 10:28 AM
- Server version: 5.7.9
- PHP Version: 7.0.7
-->

<pma_xml_export version="1.0" xmlns:pma="http://www.phpmyadmin.net/some_doc_url/">
    <!--
    - Structure schemas
    -->
    <pma:structure_schemas>
        <pma:database name="INFORMATION_SCHEMA" collation="utf8_general_ci" charset="utf8">
            <pma:table name="COLUMNS">
                CREATE TEMPORARY TABLE `COLUMNS` (
                  `TABLE_CATALOG` varchar(512) NOT NULL DEFAULT '',
                  `TABLE_SCHEMA` varchar(64) NOT NULL DEFAULT '',
                  `TABLE_NAME` varchar(64) NOT NULL DEFAULT '',
                  `COLUMN_NAME` varchar(64) NOT NULL DEFAULT '',
                  `ORDINAL_POSITION` bigint(21) unsigned NOT NULL DEFAULT '0',
                  `COLUMN_DEFAULT` longtext,
                  `IS_NULLABLE` varchar(3) NOT NULL DEFAULT '',
                  `DATA_TYPE` varchar(64) NOT NULL DEFAULT '',
                  `CHARACTER_MAXIMUM_LENGTH` bigint(21) unsigned DEFAULT NULL,
                  `CHARACTER_OCTET_LENGTH` bigint(21) unsigned DEFAULT NULL,
                  `NUMERIC_PRECISION` bigint(21) unsigned DEFAULT NULL,
                  `NUMERIC_SCALE` bigint(21) unsigned DEFAULT NULL,
                  `DATETIME_PRECISION` bigint(21) unsigned DEFAULT NULL,
                  `CHARACTER_SET_NAME` varchar(32) DEFAULT NULL,
                  `COLLATION_NAME` varchar(32) DEFAULT NULL,
                  `COLUMN_TYPE` longtext NOT NULL,
                  `COLUMN_KEY` varchar(3) NOT NULL DEFAULT '',
                  `EXTRA` varchar(30) NOT NULL DEFAULT '',
                  `PRIVILEGES` varchar(80) NOT NULL DEFAULT '',
                  `COLUMN_COMMENT` varchar(1024) NOT NULL DEFAULT '',
                  `GENERATION_EXPRESSION` longtext NOT NULL
                ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
            </pma:table>
        </pma:database>
    </pma:structure_schemas>

    <!--
    - Database: 'INFORMATION_SCHEMA'
    -->
    <database name="INFORMATION_SCHEMA">
        <!-- Table COLUMNS -->
        <table name="COLUMNS">
            <column name="TABLE_CATALOG">def</column>
            <column name="TABLE_SCHEMA">kitten</column>
            <column name="TABLE_NAME">smf_calendar</column>
            <column name="COLUMN_NAME">id_event</column>
            <column name="ORDINAL_POSITION">1</column>
            <column name="COLUMN_DEFAULT">NULL</column>
            <column name="IS_NULLABLE">NO</column>
            <column name="DATA_TYPE">smallint</column>
            <column name="CHARACTER_MAXIMUM_LENGTH">NULL</column>
            <column name="CHARACTER_OCTET_LENGTH">NULL</column>
            <column name="NUMERIC_PRECISION">5</column>
            <column name="NUMERIC_SCALE">0</column>
            <column name="DATETIME_PRECISION">NULL</column>
            <column name="CHARACTER_SET_NAME">NULL</column>
            <column name="COLLATION_NAME">NULL</column>
            <column name="COLUMN_TYPE">smallint(5) unsigned</column>
            <column name="COLUMN_KEY">PRI</column>
            <column name="EXTRA">auto_increment</column>
            <column name="PRIVILEGES">select,insert,update,references</column>
            <column name="COLUMN_COMMENT"></column>
            <column name="GENERATION_EXPRESSION"></column>
        </table>
        <table name="COLUMNS">
            <column name="TABLE_CATALOG">def</column>
            <column name="TABLE_SCHEMA">kitten</column>
            <column name="TABLE_NAME">smf_calendar</column>
            <column name="COLUMN_NAME">start_date</column>
            <column name="ORDINAL_POSITION">2</column>
            <column name="COLUMN_DEFAULT">0001-01-01</column>
            <column name="IS_NULLABLE">NO</column>
            <column name="DATA_TYPE">date</column>
            <column name="CHARACTER_MAXIMUM_LENGTH">NULL</column>
            <column name="CHARACTER_OCTET_LENGTH">NULL</column>
            <column name="NUMERIC_PRECISION">NULL</column>
            <column name="NUMERIC_SCALE">NULL</column>
            <column name="DATETIME_PRECISION">NULL</column>
            <column name="CHARACTER_SET_NAME">NULL</column>
            <column name="COLLATION_NAME">NULL</column>
            <column name="COLUMN_TYPE">date</column>
            <column name="COLUMN_KEY">MUL</column>
            <column name="EXTRA"></column>
            <column name="PRIVILEGES">select,insert,update,references</column>
            <column name="COLUMN_COMMENT"></column>
            <column name="GENERATION_EXPRESSION"></column>
        </table>
        <table name="COLUMNS">
            <column name="TABLE_CATALOG">def</column>
            <column name="TABLE_SCHEMA">kitten</column>
            <column name="TABLE_NAME">smf_calendar</column>
            <column name="COLUMN_NAME">end_date</column>
            <column name="ORDINAL_POSITION">3</column>
            <column name="COLUMN_DEFAULT">0001-01-01</column>
            <column name="IS_NULLABLE">NO</column>
            <column name="DATA_TYPE">date</column>
            <column name="CHARACTER_MAXIMUM_LENGTH">NULL</column>
            <column name="CHARACTER_OCTET_LENGTH">NULL</column>
            <column name="NUMERIC_PRECISION">NULL</column>
            <column name="NUMERIC_SCALE">NULL</column>
            <column name="DATETIME_PRECISION">NULL</column>
            <column name="CHARACTER_SET_NAME">NULL</column>
            <column name="COLLATION_NAME">NULL</column>
            <column name="COLUMN_TYPE">date</column>
            <column name="COLUMN_KEY">MUL</column>
            <column name="EXTRA"></column>
            <column name="PRIVILEGES">select,insert,update,references</column>
            <column name="COLUMN_COMMENT"></column>
            <column name="GENERATION_EXPRESSION"></column>
        </table>
        <table name="COLUMNS">
            <column name="TABLE_CATALOG">def</column>
            <column name="TABLE_SCHEMA">kitten</column>
            <column name="TABLE_NAME">smf_calendar</column>
            <column name="COLUMN_NAME">id_board</column>
            <column name="ORDINAL_POSITION">4</column>
            <column name="COLUMN_DEFAULT">0</column>
            <column name="IS_NULLABLE">NO</column>
            <column name="DATA_TYPE">smallint</column>
            <column name="CHARACTER_MAXIMUM_LENGTH">NULL</column>
            <column name="CHARACTER_OCTET_LENGTH">NULL</column>
            <column name="NUMERIC_PRECISION">5</column>
            <column name="NUMERIC_SCALE">0</column>
            <column name="DATETIME_PRECISION">NULL</column>
            <column name="CHARACTER_SET_NAME">NULL</column>
            <column name="COLLATION_NAME">NULL</column>
            <column name="COLUMN_TYPE">smallint(5) unsigned</column>
            <column name="COLUMN_KEY"></column>
            <column name="EXTRA"></column>
            <column name="PRIVILEGES">select,insert,update,references</column>
            <column name="COLUMN_COMMENT"></column>
            <column name="GENERATION_EXPRESSION"></column>
        </table>
        <table name="COLUMNS">
            <column name="TABLE_CATALOG">def</column>
            <column name="TABLE_SCHEMA">kitten</column>
            <column name="TABLE_NAME">smf_calendar</column>
            <column name="COLUMN_NAME">id_topic</column>
            <column name="ORDINAL_POSITION">5</column>
            <column name="COLUMN_DEFAULT">0</column>
            <column name="IS_NULLABLE">NO</column>
            <column name="DATA_TYPE">mediumint</column>
            <column name="CHARACTER_MAXIMUM_LENGTH">NULL</column>
            <column name="CHARACTER_OCTET_LENGTH">NULL</column>
            <column name="NUMERIC_PRECISION">7</column>
            <column name="NUMERIC_SCALE">0</column>
            <column name="DATETIME_PRECISION">NULL</column>
            <column name="CHARACTER_SET_NAME">NULL</column>
            <column name="COLLATION_NAME">NULL</column>
            <column name="COLUMN_TYPE">mediumint(8) unsigned</column>
            <column name="COLUMN_KEY">MUL</column>
            <column name="EXTRA"></column>
            <column name="PRIVILEGES">select,insert,update,references</column>
            <column name="COLUMN_COMMENT"></column>
            <column name="GENERATION_EXPRESSION"></column>
        </table>
        <table name="COLUMNS">
            <column name="TABLE_CATALOG">def</column>
            <column name="TABLE_SCHEMA">kitten</column>
            <column name="TABLE_NAME">smf_calendar</column>
            <column name="COLUMN_NAME">title</column>
            <column name="ORDINAL_POSITION">6</column>
            <column name="COLUMN_DEFAULT"></column>
            <column name="IS_NULLABLE">NO</column>
            <column name="DATA_TYPE">varchar</column>
            <column name="CHARACTER_MAXIMUM_LENGTH">255</column>
            <column name="CHARACTER_OCTET_LENGTH">765</column>
            <column name="NUMERIC_PRECISION">NULL</column>
            <column name="NUMERIC_SCALE">NULL</column>
            <column name="DATETIME_PRECISION">NULL</column>
            <column name="CHARACTER_SET_NAME">utf8</column>
            <column name="COLLATION_NAME">utf8_general_ci</column>
            <column name="COLUMN_TYPE">varchar(255)</column>
            <column name="COLUMN_KEY"></column>
            <column name="EXTRA"></column>
            <column name="PRIVILEGES">select,insert,update,references</column>
            <column name="COLUMN_COMMENT"></column>
            <column name="GENERATION_EXPRESSION"></column>
        </table>
        <table name="COLUMNS">
            <column name="TABLE_CATALOG">def</column>
            <column name="TABLE_SCHEMA">kitten</column>
            <column name="TABLE_NAME">smf_calendar</column>
            <column name="COLUMN_NAME">id_member</column>
            <column name="ORDINAL_POSITION">7</column>
            <column name="COLUMN_DEFAULT">0</column>
            <column name="IS_NULLABLE">NO</column>
            <column name="DATA_TYPE">mediumint</column>
            <column name="CHARACTER_MAXIMUM_LENGTH">NULL</column>
            <column name="CHARACTER_OCTET_LENGTH">NULL</column>
            <column name="NUMERIC_PRECISION">7</column>
            <column name="NUMERIC_SCALE">0</column>
            <column name="DATETIME_PRECISION">NULL</column>
            <column name="CHARACTER_SET_NAME">NULL</column>
            <column name="COLLATION_NAME">NULL</column>
            <column name="COLUMN_TYPE">mediumint(8) unsigned</column>
            <column name="COLUMN_KEY"></column>
            <column name="EXTRA"></column>
            <column name="PRIVILEGES">select,insert,update,references</column>
            <column name="COLUMN_COMMENT"></column>
            <column name="GENERATION_EXPRESSION"></column>
        </table>
        <table name="COLUMNS">
            <column name="TABLE_CATALOG">def</column>
            <column name="TABLE_SCHEMA">kitten</column>
            <column name="TABLE_NAME">smf_calendar</column>
            <column name="COLUMN_NAME">start_time</column>
            <column name="ORDINAL_POSITION">8</column>
            <column name="COLUMN_DEFAULT">NULL</column>
            <column name="IS_NULLABLE">YES</column>
            <column name="DATA_TYPE">time</column>
            <column name="CHARACTER_MAXIMUM_LENGTH">NULL</column>
            <column name="CHARACTER_OCTET_LENGTH">NULL</column>
            <column name="NUMERIC_PRECISION">NULL</column>
            <column name="NUMERIC_SCALE">NULL</column>
            <column name="DATETIME_PRECISION">0</column>
            <column name="CHARACTER_SET_NAME">NULL</column>
            <column name="COLLATION_NAME">NULL</column>
            <column name="COLUMN_TYPE">time</column>
            <column name="COLUMN_KEY"></column>
            <column name="EXTRA"></column>
            <column name="PRIVILEGES">select,insert,update,references</column>
            <column name="COLUMN_COMMENT"></column>
            <column name="GENERATION_EXPRESSION"></column>
        </table>
        <table name="COLUMNS">
            <column name="TABLE_CATALOG">def</column>
            <column name="TABLE_SCHEMA">kitten</column>
            <column name="TABLE_NAME">smf_calendar</column>
            <column name="COLUMN_NAME">end_time</column>
            <column name="ORDINAL_POSITION">9</column>
            <column name="COLUMN_DEFAULT">NULL</column>
            <column name="IS_NULLABLE">YES</column>
            <column name="DATA_TYPE">time</column>
            <column name="CHARACTER_MAXIMUM_LENGTH">NULL</column>
            <column name="CHARACTER_OCTET_LENGTH">NULL</column>
            <column name="NUMERIC_PRECISION">NULL</column>
            <column name="NUMERIC_SCALE">NULL</column>
            <column name="DATETIME_PRECISION">0</column>
            <column name="CHARACTER_SET_NAME">NULL</column>
            <column name="COLLATION_NAME">NULL</column>
            <column name="COLUMN_TYPE">time</column>
            <column name="COLUMN_KEY"></column>
            <column name="EXTRA"></column>
            <column name="PRIVILEGES">select,insert,update,references</column>
            <column name="COLUMN_COMMENT"></column>
            <column name="GENERATION_EXPRESSION"></column>
        </table>
        <table name="COLUMNS">
            <column name="TABLE_CATALOG">def</column>
            <column name="TABLE_SCHEMA">kitten</column>
            <column name="TABLE_NAME">smf_calendar</column>
            <column name="COLUMN_NAME">timezone</column>
            <column name="ORDINAL_POSITION">10</column>
            <column name="COLUMN_DEFAULT">NULL</column>
            <column name="IS_NULLABLE">YES</column>
            <column name="DATA_TYPE">varchar</column>
            <column name="CHARACTER_MAXIMUM_LENGTH">80</column>
            <column name="CHARACTER_OCTET_LENGTH">240</column>
            <column name="NUMERIC_PRECISION">NULL</column>
            <column name="NUMERIC_SCALE">NULL</column>
            <column name="DATETIME_PRECISION">NULL</column>
            <column name="CHARACTER_SET_NAME">utf8</column>
            <column name="COLLATION_NAME">utf8_general_ci</column>
            <column name="COLUMN_TYPE">varchar(80)</column>
            <column name="COLUMN_KEY"></column>
            <column name="EXTRA"></column>
            <column name="PRIVILEGES">select,insert,update,references</column>
            <column name="COLUMN_COMMENT"></column>
            <column name="GENERATION_EXPRESSION"></column>
        </table>
    </database>
</pma_xml_export>
```

Sesquipedalian:
Hm. That all looks to be in order. 

To clarify, do you mean that you cannot enter times in the interface, or that you can enter them, but they are not being recorded in the database?

XinYenFon:
yes.

Sesquipedalian:
Yes to which one?

XinYenFon:
I can enter time, but its not recorded :laughing:

Sesquipedalian:
Hm. Anything in your error logs? I just tested again on both my local and remote servers, and creating events with times worked correctly in both cases.

XinYenFon:
I started to suspect its related to timezone setting in my local build. While my local has `date.timezone = "Europe/London"` I'm still having `Warning: timezone_open(): Unknown or bad timezone ()` error.

Sesquipedalian:
At what line numbers does that error happen?

Sesquipedalian:
... and in which files, of course. :)

XinYenFon:
screenshot-kittenground 2016-08-22 21-27-18

Sesquipedalian:
Thanks. That section is actually trying to process information retrieved from the database, rather than trying to save to the database. What happens if you go to line 146 in Subs-Calendar.php and replace this:

```
if (isset($row['start_time']) && isset($row['end_time']) && isset($row['timezone']))
```

with this:

```
if (!empty($row['start_time']) && !empty($row['end_time']) && !empty($row['timezone']))
```

XinYenFon:
yeap, no more issues. but I still can't set specific time for events.

Also I noticed something weird. lets say I set all day event 15th - 16th. When i go back to edit, it takes the event one day back (14th - 15th). Save go back one more back and so on.

Sesquipedalian:
Okay. I think that somehow the timezone on your event was entered into the database as an empty string, and that has been causing the problem.

Sesquipedalian:
> Also I noticed something weird. lets say I set all day event 15th - 16th. When i go back to edit, it takes the event one day back (14th - 15th). Save go back one more back and so on.

That's a function of datepair.js. Calendar applications like Google Calendar, Apple's iCal, etc., typically do the same thing. My understanding is that they do this to make it easier for users to change the time when an event starts without messing with the event duration. The scenario goes like this:
1. "Let's set up a meeting on Tuesday from 2:00 PM to 3:00 PM"
2. "Oh wait, we need to reschedule that to 1:00PM instead."

In a situation like this, users typically intend for the meeting to still be one hour long. They are just moving it to a different time in the day. The person who wrote datepair.js noticed that the professional calendar apps did this, and so he recreated it in his script.

XinYenFon:
I don't think its a good feature. User can go back to edit description only and overlook the date (or assume it stay same).

Sesquipedalian:
Wait, you mean it is changing it on page load? Oh. I thought you meant that it was adjusting the end date/time when the user interacted with the start date/time. 

Yeah, that would be another effect of having the broken time zone info in the database entry for the event.

I've just pushed some commits to ensure that the timezone info entered into the database will always be valid. Please delete all your existing events and try creating new ones using the updated code.

XinYenFon:
Unfortunately nothing changed for me I tested it on 1ac1e1b commit. Same 2 behavior still exist.

Sesquipedalian:
Ugh. Can you give me an SQL dump of the contents of your smf_calendar table?

XinYenFon:
[kitten.sql.zip](https://github.com/SimpleMachines/SMF2.1/files/431101/kitten.sql.zip)

Sesquipedalian:
How about now, @Antes?

XinYenFon:
Everything same. Putting one day back and the issue I can't set start/end time still exist.

Sesquipedalian:
What about with a fresh install and fresh database? I have never yet been able to reproduce the issue you are seeing, so I am starting to suspect that something has gone wrong with your specific setup.

XinYenFon:
I took your branch and `Fatal error: Maximum function nesting level of '256' reached, aborting!`. I'll try with fresh copy from here and your patch.

XinYenFon:
I did a fresh install, then checkout to this PR, run upgrade.php and same output

`Fatal error: Maximum function nesting level of '256' reached, aborting! in .\Sources\Load.php on line 135`

Sesquipedalian:
> `Fatal error: Maximum function nesting level of '256' reached, aborting! in .\Sources\Load.php on line 135`

What the...? That's a runaway recursion in the htmlspecialchars function. None of my code touches or even calls that function.

XinYenFon:
To keep sanity, I remove the image. Issue solved with ec15640. But unfortunately main issues still same.

Sesquipedalian:
Ah. Apparently something went wrong when calling log_error. I've removed that now. 

Still, this is revealing. It looks like that root of the problem is that you have an invalid time zone selected in `$modSettings['default_timezone']`. What do you have selected there?

XinYenFon:
Europe/London.

Sesquipedalian:
PHP seems to think otherwise. If you run the following script, you'll see that the output is `true`.

```
<?php
$modSettings['default_timezone'] = 'Europe/London';
if (isset($modSettings['default_timezone']) && in_array($modSettings['default_timezone'], timezone_identifiers_list()))
    echo 'true', PHP_EOL;
else
    echo 'false', PHP_EOL;
?>
```

That if statement is the same as on line 220 of Load.php (in this PR). That means that whatever is in your real $modSettings['default_timezone'] is not in PHP's list of time zone identifiers.

My guess is that you actually still have the value that SMF first put there during install. Prior to    885853b, SMF would give $modSettings['default_timezone'] a value like `Etc/GMT+7` or`Etc/GMT-4` or whatever. PHP is willing to use these as values for `date_default_timezone_set()`, but it seems that they aren't fully supported in more advanced DateTime functions, so using one of them for $modSettings['default_timezone'] would break my code.

So in 885853b I've changed the install logic so that it sets a proper time zone identifier (e.g. `Europe/London` rather than `Etc/GMT+0`) as the initial value for $modSettings['default_timezone']. And in 512f86f I've pasted the same code into Load.php so that if an invalid tzid is found in $modSettings['default_timezone'], the code will set it aside and find a proper one to use instead.

Sesquipedalian:
> Also I noticed something weird. lets say I set all day event 15th - 16th. When i go back to edit, it takes the event one day back (14th - 15th). Save go back one more back and so on.

I was finally able to reproduce this. Turns out that it was a bug the date picker script I was using. I've now replaced it with another (JQuery UI's date picker) and that seems to have solved it nicely.

So, @Antes, between this new JavaScript and the changes I made relating to time zones, have those two bugs been solved in your tests?

XinYenFon:
Date no longer going back. But still I can't save the start/end time.

Sesquipedalian:
Well, at least one is solved...

Sesquipedalian:
I'm still working on figuring out why event time zones are not being set for you, @Antes. I need some more diagnostic information from your install in order to pin down where the problem is and how to reproduce it. So, I'd like you to do the following, if you don't mind:
1. Checkout the latest version of this branch
2. Manually replace the contents of Sources/Subs-Calendar.php with the contents of this text file: [Subs-Calendar.txt](https://github.com/SimpleMachines/SMF2.1/files/455683/Subs-Calendar.txt)
3. Create a new event.
4. This modified version of Subs-Calendar.php will generate one (or more) messages in the Error Log. Please copy and paste the content of the error messages here.

matthew-kerle:
Can you provide a screenshot of the calendar actions that you changed? I'm curious to see what they look like now. I put in a lot of work on SMF 2.1's calendar a while back.

Sources/Calendar.php
```diff
@@ -219,18 +219,31 @@ function CalendarMain()
...
+	// We need these for all kinds of useful functions.
 	require_once($sourcedir . '/Subs-Calendar.php');
+	require_once($sourcedir . '/Subs.php');
 
 	// Cast this for safety...
 	if (isset($_REQUEST['eventid']))
 		$_REQUEST['eventid'] = (int) $_REQUEST['eventid'];
 
+	// We want a fairly compact version of the time, but as close as possible to the user's settings.
+	if (preg_match('~%[HkIlMpPrRSTX](?:[^%]*%[HkIlMpPrRSTX])*~', $user_info['time_format'], $matches) == 0 || empty($matches[0]))
```
matthew-kerle, line 237: I believe that SMF always wants to wrap if / else with brackets, no? Been a while

Sources/Calendar.php
```diff
@@ -365,6 +423,55 @@ function CalendarPost()
 	$context['linktree'][] = array(
 		'name' => $context['page_title'],
 	);
+
+	loadCSSFile('jquery-ui.datepicker.css', array('defer' => false), 'smf_datepicker');
+	loadCSSFile('jquery.timepicker.css', array('defer' => false), 'smf_timepicker');
+	loadJavascriptFile('jquery-ui.datepicker.min.js', array('defer' => true), 'smf_datepicker');
```
matthew-kerle: Can `loadCSSFile` and `loadJavascriptFile` not take arrays? Really surprising if not...also, it should be `loadJavaScript`.

Sources/Calendar.php
```diff
@@ -365,6 +423,55 @@ function CalendarPost()
...
+	loadJavascriptFile('jquery.timepicker.min.js', array('defer' => true), 'smf_timepicker');
+	loadJavascriptFile('datepair.min.js', array('defer' => true), 'smf_datepair');
+	addInlineJavascript('
+	$("#allday").click(function(){   
+		$("#start_time").attr("disabled", this.checked);
+		$("#end_time").attr("disabled", this.checked);
+		$("#tz").attr("disabled", this.checked);
+	});
+	$( "#event_time_input .date_input" ).datepicker({
```
matthew-kerle: Extra spaces here around the parens.

Sources/Calendar.php
```diff
@@ -365,6 +423,55 @@ function CalendarPost()
...
+		dateFormat: "yy-mm-dd",
```
matthew-kerle, line 434: `yy-mm-dd` is that default date format for SMF now...? I would think it should be default of `mm-dd-yyyy`.

Sources/Calendar.php
Sesquipedalian: Not always, no. Or if it is SMF's convention, it seems to get broken very often.

Sources/Calendar.php
Sesquipedalian: No, and no. :)

Sources/Calendar.php
Sesquipedalian: For whatever reason, the date picker script uses `yy` to indicate a four digit year number.

As for the decision to use ISO format for dates, it was in part driven by JavaScript's relatively weak date parsing capabilities compared to PHP's. If it were possible to use whatever the user had chosen as their preferred date format, I would have just told the date picker script to expect input in that format. But there are plenty of possible combinations that the JavaScript can't handle, so I had to choose a single, reliable standard for that. ISO format is compact, unambiguous, and international, whereas `mm-dd-yyyy` format is idiosyncratically American and ambiguous when dealing with real values. An American would read `10-02-2016` as the second day of October in the year 2016, while a Brit would read it as the tenth day of February.

Sources/Calendar.php
matthew-kerle: I think that it might actually be for single-line if / else no brackets in SMF.

Sources/Calendar.php
matthew-kerle: That would be a very nice addition...I might actually throw in a PR for this functionality. Definitely unappealing to have to make multiple calls to a ~~method~~ _function_ instead of passing in an array.

And actually, it is Java<strong>S</strong>cript: https://en.wikipedia.org/wiki/JavaScript

Sources/Calendar.php
matthew-kerle: Can we not take in the user preferred datetime format? Also have a look at Moment.js, might be useful here.

Sesquipedalian:
Screen shots for @Labradoodle-360 :)

Creating/editing an event, with date picker shown:
screen shot 2016-09-07 at 1 15 11 pm
Time picker:
screen shot 2016-09-07 at 1 16 35 pm
With "All day" set:
screen shot 2016-09-07 at 1 17 06 pm
Event displayed on the calendar in month view:
screen shot 2016-09-07 at 1 18 32 pm
Event displayed on the calendar in week view:
screen shot 2016-09-07 at 1 19 28 pm

matthew-kerle:
Thanks for the screenshots, it looks awesome. Super stoked to have this in.

Sources/Calendar.php
Sesquipedalian: That might be a very useful PR. 

And yes, I am aware that the proper spelling is JavaScript, but the function is named `loadJavascriptFile`, so that's what I need to use here.

matthew-kerle: Oh, I know it's `loadJavascriptFile` by its definition, my argument was maybe it should be changed to reflect the proper spelling. You definitely can't use it here if that's now how its defined.

Sesquipedalian: In theory, changing the function's name to `loadJavaScriptFile` would be nice. But I'm sure there are many mods currently written to call `loadJavascriptFile` that would break if we changed it now.

Sources/Calendar.php
Sesquipedalian: As I noted in the PR's opening comments:

> Apart from keeping the JS happy, there's no reason the date fields couldn't use any format that PHP understands.

So, if you or someone else were to create a subsequent PR to make the JavaScript date parsing and formatting more robust, that would be lovely. So long the JS outputs a string in the input field in a format that PHP can parse, it can be whatever you like.

XinYenFon:
screenshot-localhost 2016-09-09 02-37-54

Sesquipedalian:
Thanks, @Antes. What do you get if you try to create an event that isn't set to "All day"?

XinYenFon:
Updated the image above to be exact. @Sesquipedalian 

I set UTC timezone and random start/end time.

Sesquipedalian:
Thanks again! Just to clarify, do you mean you set your server time zone to UTC in the admin centre, or that you set this event's time zone to UTC?

XinYenFon:
Server (Admin Panel) timezone set to Europe/London.

Sesquipedalian:
Okay, this is helping me zero in on the issue. Can you run it one more time, but using this version of Subs-Calendar.php instead?

XinYenFon:
Still empty unfortunately.

Sesquipedalian:
But what error message did it produce? The code in the latest attached version of Subs-Calendar.txt produces even more content in the error message to help me hone in on exactly where the the issue lies. I _think_ what is happening is that date_parse() is producing some warnings on your install that it isn't on mine, so I need to see what those are.

Sesquipedalian:
Actually, here, use this:

XinYenFon:
screenshot-localhost 2016-09-09 03-28-47

Sesquipedalian:
And one more time, even more verbosely than before:
[Sub-Calendar.txt](https://github.com/SimpleMachines/SMF2.1/files/462971/Sub-Calendar.txt)

Sorry to hit you with so many, by the way. :)

XinYenFon:
I can't see the start/end time in calendar/topic but db has those values now.

Sesquipedalian:
Well, that sounds like improvement. What error did the latest version of Subs-Calendar.txt generate?

XinYenFon:
report_1

Sesquipedalian:
Yes, it looks like everything went into the database correctly that time. But you still don't see the times displayed on the calendar and in the topic?

XinYenFon:
yeap, when i go back to edit mode, it shows those empty.

Sesquipedalian:
Hm. Okay then. Here is yet another version that will spit out a similar message to the error log when you enter the event edit screen. Please try it out and then post the content of the error message it generates.
[Sub-Calendar.txt](https://github.com/SimpleMachines/SMF2.1/files/463005/Sub-Calendar.txt)

Sesquipedalian:
Whenever you get a chance to run a test with the version of Subs-Calendar that I attached in my last post, it will be greatly appreciated, @Antes. :)

XinYenFon:
This one seems worsen things :P now my error log doesn't show start/end time in debug info.

Sesquipedalian:
Can you post another shot of the error log messages?

> This one seems worsen things :P now my error log doesn't show start/end time in debug info.

That's really weird. I didn't change anything at all in that section. In fact, I was a bit surprised when you said that that part suddenly did start working for you, since all I am doing is gathering lots of diagnostic info. I haven't done anything yet to change the way the code actually works. 

Was there anything at all that you did differently the time when it did work? Maybe it was the difference between editing an existing event vs. creating a new event, for example.

Again, I appreciate your help running all these tests for me, @Antes. Since I cannot reproduce the bug on my own system, it is hard to figure out what the issue is except by doing this.

Sesquipedalian:
By the way, what version of PHP are you using? Also which database, and what version of that?

XinYenFon:
php 5.6.19
mysql 5.7.11

screenshot-localhost 2016-09-12 22-02-33

XinYenFon:
I wiped my local install and even my WAMP then reinstalled everything, result is same. Start/end time saving on first "post" but when edit (return) it removes start/end time. Return of start_time is false even the db entry is not empty, and its returning boolean, I guess it needs to return string like start_date.

```
...\default\Post.template.php:19:boolean false
...\default\Post.template.php:20:string '2016-09-15' (length=10)
```

Sesquipedalian:
Thanks, @Antes. That is helpful information. Question: when you enter the edit screen to edit an event that previously had start and end times, does the "All day" checkbox also become checked when the page loads?

XinYenFon:
No. It doesn't.

Sesquipedalian:
That's helpful to know, thanks. It means something is going wrong in preparing the start time and end time strings for presentation after they have been retrieved from the database. Do the changes in 51ddd8e help?

XinYenFon:
Unfortunately.

Sesquipedalian:
Darn. Okay, let's try the following:
1. Checkout the latest version of this PR
2. Create a new event with start and end times. (Please make sure to link it to a topic when you create it.)
3. After the event is created, please look at it in the topic display and make sure all of the relevant info is shown. It should look similar to this:
   screen shot 2016-09-20 at 7 52 08 pm
4. Replace the contents of Sources/Post.php with the contents of this file:
   [Post.txt](https://github.com/SimpleMachines/SMF2.1/files/483957/Post.txt)
5. Click the edit icon next to the event info in the topic display.
6. When the edit page loads, don't save any changes. Instead, go to the error log.
7. In the error log, you should find two new error messages. One will be from line 281 of Post.php, and the other from line 330. Please capture a screenshot of both error messages and paste it here for me.

Sources/Subs-Calendar.php
```diff
@@ -1025,6 +1143,302 @@ function getEventProperties($event_id)
...
+	$end_day = isset($_REQUEST['end_day']) ? $_REQUEST['end_day'] : $today['mday'];
+	$end_hour = isset($_REQUEST['end_hour']) ? $_REQUEST['end_hour'] : ($today['hours'] < 23 ? $today['hours'] + 1 : $today['hours']);
+	$end_minute = isset($_REQUEST['end_minute']) ? $_REQUEST['end_minute'] : ($today['hours'] < 23 ? $today['minutes'] : 59);
+	$end_second = isset($_REQUEST['end_second']) ? $_REQUEST['end_second'] : 0;
+	$start_date = isset($_REQUEST['start_date']) ? $_REQUEST['start_date'] : sprintf('%04d-%02d-%02d', $start_year, $start_month, $start_day);
+	$start_time = isset($_REQUEST['start_time']) ? $_REQUEST['start_time'] : sprintf('%02d:%02d:%02d', $start_hour, $start_minute, $start_second);
+	$end_date = isset($_REQUEST['end_date']) ? $_REQUEST['end_date'] : sprintf('%04d-%02d-%02d', $end_year, $end_month, $end_day);
+	$end_time = isset($_REQUEST['end_time']) ? $_REQUEST['end_time'] : sprintf('%02d:%02d:%02d', $end_hour, $end_minute, $end_second);
+	$start_datetime = isset($_REQUEST['start_datetime']) ? $_REQUEST['start_datetime'] : $start_date . ' ' . $start_time;
+	$end_datetime = isset($_REQUEST['end_datetime']) ? $_REQUEST['end_datetime'] : $end_date . ' ' . $end_time;
```
jdarwood007, line 1191: Can you clean these better?  Lots of those look like they should be ints or in formats that can be validated.

Sources/Subs-Calendar.php
```diff
@@ -1025,6 +1143,302 @@ function getEventProperties($event_id)
...
+	$start_minute = isset($eventOptions['minute']) ? $eventOptions['minute'] : (isset($_POST['minute']) ? $_POST['minute'] : null);
+	$start_second = isset($eventOptions['second']) ? $eventOptions['second'] : (isset($_POST['second']) ? $_POST['second'] : null);
+	$end_year = isset($eventOptions['end_year']) ? $eventOptions['end_year'] : (isset($_POST['end_year']) ? $_POST['end_year'] : null);
+	$end_month = isset($eventOptions['end_month']) ? $eventOptions['end_month'] : (isset($_POST['end_month']) ? $_POST['end_month'] : null);
+	$end_day = isset($eventOptions['end_day']) ? $eventOptions['end_day'] : (isset($_POST['end_day']) ? $_POST['end_day'] : null);
+	$end_hour = isset($eventOptions['end_hour']) ? $eventOptions['end_hour'] : (isset($_POST['end_hour']) ? $_POST['end_hour'] : null);
+	$end_minute = isset($eventOptions['end_minute']) ? $eventOptions['end_minute'] : (isset($_POST['end_minute']) ? $_POST['end_minute'] : null);
+	$end_second = isset($eventOptions['end_second']) ? $eventOptions['end_second'] : (isset($_POST['end_second']) ? $_POST['end_second'] : null);
```
jdarwood007, line 1309: Same here, no validation that data is as we expect it to be and just accepting raw user input.

Sesquipedalian: The validation section starts on line 1317.

Sources/Subs-Calendar.php
Sesquipedalian: Good catch. Looks like I forgot to finish implementing the validation in this function. I've now added that in dfc0e8e.

XinYenFon:
Latest commits seems to fix everything for me. But I noticed something, you can set minutes unlike how i set them. I can set time like; 1 PM, 1:30 PM etc... you can set 1:32 PM / 3:35 AM etc...

Sesquipedalian:
> Latest commits seems to fix everything for me.

Excellent!

> But I noticed something, you can set minutes unlike how i set them. I can set time like; 1 PM, 1:30 PM etc... you can set 1:32 PM / 3:35 AM etc...

You can if you type directly into the input field. The timepicker dropdown offers values at 30 minute intervals, but that's just for convenience.

XinYenFon:
Oh... that's wonderful and its good to know probably most of the users will skip or won't notice that.

Sources/Subs-Calendar.php
jdarwood007: Yea, I was thinking more along the lines of assuring they are ints in most causes (as day, hour, minute and second should always be this) and other simple validation.  I'm not sure what dirty input into strtotime would cause, but it seems we should only make sure sensible data should even make it there.  Since you already setting up the vars, this isn't too much other than (int) on most everything.

Sesquipedalian: We certainly want to validate the input, yes. However, casting everything to `int` is not necessary or desirable in this situation. It is better to rely on PHP's date parser, which can handle anything from "2016-09-23 10:00:00" to "fourth friday of sept 2016 10:00 am". If the parser receives input too malformed for it to make sense of, it'll return false.

Even so, I still do more checks after the date parser has had a turn with the input:

The year, month, and day inputs are validated with `checkdate()`, which is more thorough and more relevant than just testing for integer values.

The hour, minute, and second inputs are validated by first running them through `sprintf()` (only integer values will get through the format I have there) and then `date_parse()`, which (re)converts all the input to integers (if it is valid) and reports errors and/or warnings if anything was malformed in the input.

By the time the input goes through all that, it is not only guaranteed to be clean, but moreover guaranteed to comprise a valid date.

Oldiesmann:
Not sure if this is ready to go yet but there's a merge conflict with install_2-1_mysql.sql now.

Sesquipedalian:
> Not sure if this is ready to go yet but there's a merge conflict with install_2-1_mysql.sql now.

The conflict has been resolved now. At the moment I am doing some final reviews to make sure it is ready to go. I'll post a comment here as soon as I am ready to give it the final go-ahead. :)

Sesquipedalian:
In the midst of all this work with calendars and time zones, I noticed that the return value from `smf_list_timezones()` was both incomplete and inaccurate. I've now rewritten with the following improvements:
- Generates the list of available time zones dynamically based on the most recent time zone info PHP has available. Therefore, always up to date and accurate.
- Returns exactly one representative time zone identifier for each and every unique set of time zone rules. No duplicates, and nothing missing.
- The admin can now choose to prioritize time zones from certain countries. If most of your members are from Canada, for example, you can enter "CA" in the admin settings to show Canadian time zones at the top of the select list.

Sesquipedalian:
And with that, I think this PR is ready to merge, @Oldiesmann.

Sesquipedalian:
Figures. There always has to be one last bug. _Now_ it's ready. :)

Sources/Subs-Calendar.php
```diff
@@ -980,6 +1073,10 @@ function getEventProperties($event_id)
 			c.id_event, c.id_board, c.id_topic, MONTH(c.start_date) AS month,
 			DAYOFMONTH(c.start_date) AS day, YEAR(c.start_date) AS year,
 			(TO_DAYS(c.end_date) - TO_DAYS(c.start_date)) AS span, c.id_member, c.title,
+			MONTH(c.end_date) AS end_month, DAYOFMONTH(c.end_date) AS end_day, YEAR(c.end_date) AS end_year,
+			HOUR(start_time) AS hour, MINUTE(start_time) AS minute, SECOND(start_time) AS second,
```
albertlast: two things:
1. you should use the table alias (c) in this and next line  start_time -> c.start_time ...
1. in my eyes this table call is not good use, because the heavy use of time function month, dayofmonth, year, hour, minute and second().
   Better would be to keep the line 1079 and remove 1077, 1078,
   the remaining information generate in php.
   Reasons:
2. the timefunction are different in mysql and pg 
3. sql statement get cleaner

Sesquipedalian: Yes, that would be better. I'll implement those changes soon.

Sesquipedalian: 3bfcb48 + ce807d4

equinoxmatt:
Any plans to get this merged in?  Does it need more testing?

frandominguezl:
@equinoxmatt According to Antes on IRC, Beta 4 development hasn't started yet. We should wait for it then.

colinschoen:
As soon as we get Beta 3 out, this will be the priority to get merged in first thing for Beta 4. Thanks for the work on this.

Sesquipedalian:
That said, if you want to check out the branch and test it now, that'd be great, @equinoxmatt.

equinoxmatt:
@Sesquipedalian - I will look at checking out the branch in the next few days and giving it a look.

live627:
@Sesquipedalian - what's the status on this? Gotta move forward.

Also: conflicts on ./other/upgrade_2-1_mysql.sql

Sesquipedalian:
> what's the status on this? Gotta move forward.

It's ready to merge, so far as I am concerned. 

The recent commits just:

1. added location support
2. corrected a few minor issues

The location support is just a simple text field, so the code for that was easy and seems to work very well. As for correcting minor issues, we'll probably find some more after the merge, but that's what beta testing is for. :)

> Also: conflicts on ./other/upgrade_2-1_mysql.sql

Resolved now.

Sesquipedalian:
FYI, I'm also working on developing a list view for the calendar, in addition to the month and week views. That, however, will be submitted as a separate PR since it is a different feature.

live627:
Alright cool... will now merge this in so that the real testing begins. Seems to me that hardly anyone tests PRs.

PR #3532, opened 2016-08-05
Dynamic Load Average Caching

sicommnend:
This little bit changes the cache levels depending on the server load.

I am not going to force this to be accepted in the core, either accept or reject.


XinYenFon:
You have to sign-off your commits.

sicommnend:
Will do.

PR #3533, opened 2016-08-05
Abstracts adding <meta> tags to the output

Sesquipedalian:
~~The property and content values for the Open Graph <meta> tags are stored as key/value pairs in $context['open_graph']. Mods can add/overwrite OG tags either by calling `addOpenGraph()` directly, or by calling `call_integration_hook('integrate_theme_context')` and then editing the content of $context['open_graph'] themselves.~~

Moves all content related `<meta>` tag info into `$context['meta_tags']`. This will allow mods to add/edit/remove arbitrary meta info (e.g. Open Graph tags) simply by accessing `$context['meta_tags']`—ideally in a function that hooks into the `integrate_theme_context` hook, but theoretically anywhere.

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -2529,23 +2529,6 @@ function addInlineJavascript($javascript, $defer = false)
 }
 
 /**
- * Add info for an Open Graph <meta> tag to include in the HTML head
- *
- * @param string $property The value for the meta tag's property attribute
- * @param string $content The value for the meta tag's content attribute
- * @return void|bool Adds the passed values to the $context['open_graph'] array or returns if either value is empty
- */
-function addOpenGraph($property, $content)
-{
-	global $context;
-
-	if (empty($property) || empty($content))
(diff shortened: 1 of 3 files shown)


jdarwood007:
With how basic the function is, maybe we could just rename it addMetaTag or similar and cleanup all meta data in the theme.  Because as it is, anything can call the function to add meta data.

Sesquipedalian:
Not a bad idea. I'll need to adjust the input parameters a bit in order to allow setting arbitrary properties, but that should be trivial.

Sesquipedalian:
On further consideration, the function is unnecessary. We can just add meta info to $context directly.

XinYenFon:
I think its good to go.

Sesquipedalian:
> Can you define $context['meta_tags'] as an empty array first?  Ideally and preferable very early in SMF's process

Done as requested, @MissAllSunday.

PR #3534, opened 2016-08-11
Normalize files

MissAllSunday:
- remove tabs and spaces
- typos
- remove new line after closing tag
- add closing tag where needed

Signed-off-by: Jessica González suki@missallsunday.com


jdarwood007:
Could we just get rid of the closing ?>

There isn't a benefit to having it there.  Removing it prevents somebody or a mod from accidentally adding code to the end of the file improperly.

MissAllSunday:
It was my intention to do so as it will also solve the awkward css and js thing when adding code to the end of those files 

But since the package manager still has the "end" parameter it was decided to leave the closing tag for compatibility purposes

PR #3539, opened 2016-08-19
Sync Db Schema from MySQL and Pg

albertlast:
mysql was the leading platform for the pg defintion
but i changed also the mysql more cleaner way

What I did:
- Remove useless information in mysql code like
  mediumint(8) --> mediumint -- mysql ingnore the information of (8) when the datatype is mediumint, smallint or tinyint
- set in pg code the same default value as there are in mysql
- try to set in pg a datatype which match the mysql datatype int(10) unsinged -> bigint
- sync not null

found in pg in table pm_recipients a col which didn't exists in mysql (labels) and
reorder sometable so that they in the same order like mysql file are.

### DIFF
--- a/other/install_2-1_mysql.sql
+++ b/other/install_2-1_mysql.sql
@@ -6,7 +6,7 @@
 #
 
 CREATE TABLE {$db_prefix}admin_info_files (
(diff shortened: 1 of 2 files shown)


MissAllSunday:
Awesome! thanks!

PR #3543, opened 2016-08-20
Adds more required entries to the $xmlActions array

Sesquipedalian:


Sources/Load.php
```diff
@@ -1982,7 +1982,6 @@ function loadTheme($id_theme = 0, $initialize = true)
 		'stats',
 		'notifytopic',
 		'notifyboard',
-		'signup',
```
Sesquipedalian: Removed because @MissAllSunday already chose to deal with `signup` another way in 796368c

MissAllSunday:
mmm, I thought I already merged this one?

Sesquipedalian:
> mmm, I thought I already merged this one?

[Second verse, same as the first!](https://github.com/SimpleMachines/SMF2.1/pull/3537)

PR #3546, opened 2016-08-22
Update install_2-1_postgresql.sql

Underdog-01:
- ip column in log_online table should be varbinary(16) for ipv4 & ipv6 support


albertlast:
Nope,
inet supportet ipv4/ipv6 and works better as the varbinary hack from mysql.

MissAllSunday:
Yep, it will be better to leave this to inet_ function

Underdog-01:
Yes I was looking at a db reference created more than a month ago in error.
These changes have already been implemented (as an inet reference) to create the varbinary(16) table.
My apologies.

Regards.

PR #3550, opened 2016-08-24
[wip] Partial upgrade

albertlast:
here my idea from the board,
step 4 looks good
the step after 4 not so good


live627:
I assume this should save the current step in the database. IIRC the upgrader already does so in Settings.php.

albertlast:
The main  idea is,
to be able test new upgrade routine and allow user upgrade from beta 2 to beta 3, beta3 to beta4 etc.

PR #3553, opened 2016-08-26
Added wilcard to db_list_tables call - fixes UTF-8 conversion issue

equinoxmatt:
This fixes the issue with UTF-8 conversion.  Added the wildcard for the prefix back in (e.g smf_%) which was removed in a previous commit.

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -3922,7 +3922,7 @@ function convertUtf8()
 
 		// Get a list of table names ahead of time... This makes it easier to set our substep and such
 		db_extend();
-		$queryTables = $smcFunc['db_list_tables'](false, $db_prefix);
+		$queryTables = $smcFunc['db_list_tables'](false, $db_prefix . '%');
 
 		$upcontext['table_count'] = count($queryTables);
 		$file_steps = $upcontext['table_count'];


XinYenFon:
heh #3554

equinoxmatt:
Closing this until I checked out some more of the issues mention in #3554

XinYenFon:
No need to close this, you can add more commits unless this PR breaks more :laughing:

jdarwood007:
Not to mention this does look like the issue.  But I'm trying to figure out where that happened.  Oldiesmann's original commit where it was in the upgrade .sql file uses the wildcard, but we had the issue then.  https://github.com/SimpleMachines/SMF2.1/commit/69e37fe8fc2f10fa3a4b7ac148c3d24ee4257430

Still trying to track down where it got into upgrade.php without the wildcard and using that function.

jdarwood007:
Found it in https://github.com/SimpleMachines/SMF2.1/commit/dfd0a19614d6d9ac6cb0306f1dc6034c4e8be475#diff-795f3202b17cb6bc3d4b771d8c6c9eaf

Seems to have happened when we where fixing it to handle timeouts properly.

XinYenFon:
btw you forgot to sign-off your commit :laughing: :feet:

PR #3556, opened 2016-08-26
Edited some CSS aspects

frandominguezl:
- I added some margin-bottom for on and off images to be aligned with the username (in the poster info) and with the user status (in the user profile).
  I reduced the margin-left of the user status text from 20px to 5px.
  I added "cursor: pointer" to the "More..." button of the quick buttons in topics.
  I changed the color of the button "button button_strip_notify" (next to print button) from Black to 346, as same as the other buttons.
  I deleted the padding property of the class "icons" to align the icons with the avatar and the rest of the items of poster info.
- It was a Firefox issue. I've added another commit to add the solution for Firefox and deleted the margin-bottom properties from on and off classes.

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -946,6 +946,7 @@ img.sort, .sort {
 /* Keep a consistent size when wrapped in pagesection. */
 .pagesection .button {
 	font-size: 0.889em;
+	color: #346;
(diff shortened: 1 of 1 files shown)


XinYenFon:
Looks good. Can be merged anytime.

PR #3559, opened 2016-08-27
Prettier mod button selection indicators

Sesquipedalian:
Instead of square brackets, use spans with `class="amt"` to indicate the number of selected posts on the relevant moderation buttons.


XinYenFon:
Nice :)

PR #3562, opened 2016-08-30
Refactor upgrade/install scripts to use FTP Class in Class-Packages.php

equinoxmatt:
The FTP class used in upgrade.php does not accept constructor values.  Unlike the one found in core.

connect() method needs to be called to make connection.  Also, error handling checks for use of last_message which was not set when connection failed.


equinoxmatt:
@MissAllSunday @Oldiesmann

Don't merge this for now - is there a reason that the FTP class in /Sources/Class-Package.php is not used?

equinoxmatt:
Modified commit - now uses /Sources/Class-Package.php FTP class with some core class modifications.

MissAllSunday:
There isn't a special reason as far as I know.

PR #3565, opened 2016-09-03
Allows URLs to contain [ ] and makes parse_bbc() generally smarter

### LINKED ISSUE #1187: BBCode parsing has trouble with urls containing braces
The following url, when pasted into a SMF forum post, will hopelessly confuse the bbcode parser (this url was taken from the address bar while browsing around in Firefox, which has the unusual behavior of decoding urls):

http://www.example.com/products?q[filters][toyline]=3533

I don't understand why this should be an issue, as it does not have any valid bbcode markup in it (no closing tags). This is using SMF 2.0.2.

### LINKED ISSUE #3366: URL with array parameter in parse BBC bug
Parse bbc breaks when there are array in URL parameters:

example.com/index.php?var[]=value

Sesquipedalian:
- The [url] and [iurl] BBCodes optionally allow the URL to be quoted (e.g. `[url="http://www.example.com"]`). This required changes in both Subs.php's `parse_bbc()` function and Subs-Post.php's `fixTag()` function.
- The [url] and [iurl] BBCodes have smarter validation functions in parse_bbc(). Previously, `http://` was blindly prepended to any string that didn't begin with `http://` or `https://`. Now, we check if a valid URI scheme is present (even if it isn't an http one), and only prepend `http://` if there really and truly is no valid scheme present.
- Autolinking uses a smarter (and much more legible) regex to detect plain text URLs.
- Autolinking now uses the form `[url="http://www.example.com"]http://www.example.com[/url]` to wrap detected URLs.
- parse_bbc() is now smarter about how it detects BBCodes as it moves through a message. Previously, the parser broke up the message text at evey `[` character, which meant that random square brackets could interfere with the parser in some cases (e.g. when trying to detect and autolink URLs that include square brackets in the URL's query parameters).  Now we use a regex to detect where the next actually valid BBCode begins, and jump straight to that position. This means the parser doesn't need to concern itself with `[` and `]` characters that are not part of a valid BBCode.
- jquery.sceditor.smf.js has been updated to handle creating url BBCodes with a quoted parameter. However, jquery.sceditor.bbcode.min.js does not understand how to translate a BBCode with the form `[url="http://example.com/index.php?var[]=value"]foo[/url]` into an HTML element with the form `<a href="http://www.example.com/index.php?var[]=value">foo</a>`, so toggling the editor between source mode and WYSIWYG mode will cause such [url] BBCodes not to render as beautifully in the editor's WYSIWYG mode. Fortunately, this is a merely aesthetic issue, and does not cause any problems with the message text itself or with how it will be parsed and displayed by SMF afterwards. It would be nice if jquery.sceditor.bbcode.min.js were updated to fix this, but that would need to be done by the fine folks at SCEditor, rather than by us.

Fixes #3366
Fixes #1187

At some point, it would be a good idea to write some code to periodically grab the latest list of valid TLDs from http://data.iana.org/TLD/tlds-alpha-by-domain.txt to use at the relevant place in parse_bbc(). That's a job for another PR, though.

### DIFF
--- a/Sources/Subs-Post.php
+++ b/Sources/Subs-Post.php
@@ -412,7 +412,12 @@ function fixTag(&$message, $myTag, $protocols, $embeddedUrl = false, $hasEqualSi
 
 	$replaces = array();
 
-	if ($hasEqualSign)
+	if ($hasEqualSign && $embeddedUrl)
+	{
+		$quoted = preg_match('~\[(' . $myTag . ')=&quot;~', $message);
+		preg_match_all('~\[(' . $myTag . ')=' . ($quoted ? '&quot;(.*?)&quot;' : '([^\]]*?)') . '\](?:(.+?)\[/(' . $myTag . ')\])?~is', $message, $matches);
+	}
+	elseif ($hasEqualSign)
 		preg_match_all('~\[(' . $myTag . ')=([^\]]*?)\](?:(.+?)\[/(' . $myTag . ')\])?~is', $message, $matches);
 	else
 		preg_match_all('~\[(' . $myTag . ($hasExtra ? '(?:[^\]]*?)' : '') . ')\](.+?)\[/(' . $myTag . ')\]~is', $message, $matches);
@@ -453,7 +458,7 @@ function fixTag(&$message, $myTag, $protocols, $embeddedUrl = false, $hasEqualSi
 			$replace = $protocols[0] . '://' . $replace;
 
 		if ($hasEqualSign && $embeddedUrl)
-			$replaces[$matches[0][$k]] = '[' . $this_tag . '=' . $replace . ']' . (empty($matches[4][$k]) ? '' : $matches[3][$k] . '[/' . $this_close . ']');
+			$replaces[$matches[0][$k]] = '[' . $this_tag . '=&quot;' . $replace . '&quot;]' . (empty($matches[4][$k]) ? '' : $matches[3][$k] . '[/' . $this_close . ']');
 		elseif ($hasEqualSign)
 			$replaces['[' . $matches[1][$k] . '=' . $matches[2][$k] . ']'] = '[' . $this_tag . '=' . $replace . ']';
 		elseif ($embeddedUrl)
--- a/Sources/Subs-Post.php
+++ b/Sources/Subs-Post.php
@@ -270,7 +270,7 @@ function fixTags(&$message)
 		array(
 			'tag' => 'url',
 			'protocols' => array('http', 'https'),
-			'embeddedUrl' => true,
+			'embeddedUrl' => false,
 			'hasEqualSign' => false,
 		),
 		// [url=http://...]name[/url]
@@ -284,7 +284,7 @@ function fixTags(&$message)
 		array(
 			'tag' => 'iurl',
 			'protocols' => array('http', 'https'),
-			'embeddedUrl' => true,
+			'embeddedUrl' => false,
 			'hasEqualSign' => false,
 		),
 		// [iurl=http://...]name[/iurl]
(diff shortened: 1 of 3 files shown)


Sesquipedalian:
Examples in use [here](http://stovell.noip.me/~jon/dev/index.php?topic=3.msg3#msg3)

Message text used:

```
This is an unquoted url bbcode with square brackets in the URL: [url=http://www.example.com/index.php?var[foo]=value]evil link[/url]. As you can see, it isn't parsed correctly.  :'( 

This is an quoted url bbcode with square brackets in the URL: [url="http://www.example.com/index.php?var[foo]=value"]not so evil link[/url]. It is parsed correctly.  :) 

This is a plain text url for autolinking: http://www.example.com/index.php?var[foo]=value Also parsed correctly. :D 

And now let's try to confuse the auto linker:

For an example of a valuably variable foo, go to http://www.example.com/index.php?var[foo]=value! Have you been to http://www.example.com/index.php?var[foo]=value? In parentheses (http://www.example.com/index.php?var[foo]=value), it is still parsed correctly. Even in brackets [http://www.example.com/index.php?var[foo]=value], this URL is no longer evil. 8) 
```

Resulting HTML:

```
<div class="inner" data-msgid="3" id="msg_3">This is an unquoted url bbcode with square brackets in the URL: <a href="http://www.example.com/index.php?var[foo" class="bbc_link" target="_blank">=value]evil link</a>. As you can see, it isn't parsed correctly. &nbsp;<img src="http://stovell.noip.me/~jon/dev/Smileys/default/cry.gif" alt=":'(" title="Cry" class="smiley"> <br><br>This is an quoted url bbcode with square brackets in the URL: <a href="http://www.example.com/index.php?var[foo]=value" class="bbc_link" target="_blank">not so evil link</a>. It is parsed correctly. &nbsp;<img src="http://stovell.noip.me/~jon/dev/Smileys/default/smiley.gif" alt=":)" title="Smiley" class="smiley"> <br><br>This is a plain text url for autolinking: <a href="http://www.example.com/index.php?var[foo]=value" class="bbc_link" target="_blank">http://www.example.com/index.php?var[foo]=value</a> Also parsed correctly. <img src="http://stovell.noip.me/~jon/dev/Smileys/default/cheesy.gif" alt=":D" title="Cheesy" class="smiley"> <br><br>And now let's try to confuse the auto linker:<br><br>For an example of a valuably variable foo, go to <a href="http://www.example.com/index.php?var[foo]=value" class="bbc_link" target="_blank">http://www.example.com/index.php?var[foo]=value</a>! Have you been to <a href="http://www.example.com/index.php?var[foo]=value" class="bbc_link" target="_blank">http://www.example.com/index.php?var[foo]=value</a>? In parentheses (<a href="http://www.example.com/index.php?var[foo]=value" class="bbc_link" target="_blank">http://www.example.com/index.php?var[foo]=value</a>), it is still parsed correctly. Even in brackets [<a href="http://www.example.com/index.php?var[foo]=value" class="bbc_link" target="_blank">http://www.example.com/index.php?var[foo]=value</a>], this URL is no longer evil. <img src="http://stovell.noip.me/~jon/dev/Smileys/default/cool.gif" alt="8)" title="Cool" class="smiley"> </div>
```

screen shot 2016-09-02 at 11 08 41 pm

Sesquipedalian:
~~Don't merge this yet. I just found one more way that handling an url with square brackets in it can fail. Will work on a fix for that soon.~~
Fixed as of 05f1fb9

Sesquipedalian:
I now consider this branch to be ready for merging.

matthew-kerle:
I'm excited to see this one go in, very nice work.

Sesquipedalian:
Thank you. :)

MissAllSunday:
Indeed, thank you!

Issue #3504, opened 2016-06-29
The function inet_dtop($bin) don't work allways correct..


PortaMx:
I found a lot of ipv6 adresses the have a  . (dot) or : (doublepoint) as binary.
So the test of '.' or ':' failed.

I change that to:
`function inet_dtop($bin)
{
    if(empty($bin))
        return '';

```
$ip_address = @inet_ntop($bin);
if($ip_address === false)
    return '';

return $ip_address;
```

}`

Now all works correct ..

albertlast:
see pr https://github.com/SimpleMachines/SMF2.1/pull/3405

PortaMx:
Yeah .. I see .. but THAT gives a runtime error on **inet_ntop** if the ip not a correct ip4 or ip6 address.
inet_ntop returns false on bad ip's and also a runtime error .. (tested)

Refer: http://php.net/manual/en/function.inet-ntop.php

albertlast:
I can't follow you.
Can you please post the code from inet_dtop which u use for this test?

PortaMx:
See my first Post .. the code is inside this 😉

albertlast:
I mean after you applied the change from pr #3405

PortaMx:
No .. my code works perfect, so I see no necessary to change a running code ...

tinoest:
But you're code isn't good coding practice, you're silencing a error with the @ sign. It's better to not have the error in the first place. 

You also state there is a error when using the code from pr #3405, is this the case or not, as your last comment contradicts your second comment.

PortaMx:
Without the @ it will gives a runtime error, that fills the error log ,,,
You can simple test that if you have a good debugger (like phped or similar) .. set a breakpoint before, manipulate the binary ip, then step into .. bäng .. runtime error ;-)
The problem is here, that you can't test the binary if is have a correct ip ... and php itself have no function for this .. in the net you find many function for this, but at this point that is overdressed.

albertlast:
Your solution doesn't work with postgresql,
take the solution from pr #3405

PortaMx:
Yeah .. Yesterday I have installed a Postgres DB, so I can also test all with postgres,
I added the postgres check and it works fine ..

albertlast:
You only need to place the pg check,
more change is not needed.

the intention of . and : check was only for pg.

MissAllSunday:
[...]

PR #3570, opened 2016-09-08
Handle IP address strings in inet_dtop()

### LINKED ISSUE #3548: inet_ntop(): Invalid in_addr value
www/beta/Sources/Subs.php
Line: 4985

This only happens when testing on localhost, the admin IP will be (string) 127.0.0  which that function cannot handle properly.

It can also happen with a few mods, I know I have one that explicitly uses 127.0.0.

Sesquipedalian:
Fixes #3548

### DIFF
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -1407,6 +1407,13 @@ function prepareDisplayContext($reset = false)
 	censorText($message['body']);
 	censorText($message['subject']);
 
+	$start = microtime(true);
+	for($i = 0; $i < 1000; $i++){
+		$foo = parse_bbc($message['body'], $message['smileys_enabled'], $message['id_msg']);
+	}
+	$time_elapsed_secs = microtime(true) - $start;
+	log_error('1000 iterations of parse_bbc took ' . $time_elapsed_secs, 'general', __FILE__, __LINE__);
+
 	// Run BBC interpreter on the message.
 	$message['body'] = parse_bbc($message['body'], $message['smileys_enabled'], $message['id_msg']);
 
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -1407,13 +1407,6 @@ function prepareDisplayContext($reset = false)
 	censorText($message['body']);
 	censorText($message['subject']);
 
-	$start = microtime(true);
-	for($i = 0; $i < 1000; $i++){
-		$foo = parse_bbc($message['body'], $message['smileys_enabled'], $message['id_msg']);
-	}
-	$time_elapsed_secs = microtime(true) - $start;
-	log_error('1000 iterations of parse_bbc took ' . $time_elapsed_secs, 'general', __FILE__, __LINE__);
-
 	// Run BBC interpreter on the message.
 	$message['body'] = parse_bbc($message['body'], $message['smileys_enabled'], $message['id_msg']);
 
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -5031,15 +5031,14 @@ function inet_ptod($ip_address)
  */
 function inet_dtop($bin)
 {
-	if(empty($bin))
-		return '';
-
 	global $db_type;
 
-	if ($db_type == 'postgresql')
-		return $bin;
-
-	$ip_address = inet_ntop($bin);
+	if(empty($bin))
+		$ip_address =  '';
+	elseif (!filter_var($bin, FILTER_VALIDATE_IP) === false || $db_type == 'postgresql')
+		$ip_address = $bin;
+	else
+		$ip_address = inet_ntop($bin);
 
 	return $ip_address;
 }
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -5033,22 +5033,22 @@ function inet_dtop($bin)
 {
 	global $db_type;
 
-	// An unencoded IPv4 address? Just return it.
+	// An unpacked IPv4 address? Just return it.
 	if (filter_var($bin, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4) !== false)
 		$ip_address = $bin;
 
-	// A binary encoded address? Decode it.
-	// Need to check this first because a binary encoded IP can also be a valid IPv6.
+	// A packed IP address? Unpack it.
+	// Need to check this first because a packed IP might take a form that validates as an IPv6.
 	elseif (filter_var(@inet_ntop($bin), FILTER_VALIDATE_IP) !== false)
 		$ip_address = inet_ntop($bin);
 
-	// An unencoded IPv6 address? Just return it.
+	// An unpacked IPv6 address? Just return it.
 	elseif (filter_var($bin, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6) !== false)
 		$ip_address = $bin;
 
 	// No valid IP address? Return an empty string.
 	else
-		$ip_address = '';
+		$ip_address = false;
 
 	return $ip_address;
 }
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -5034,26 +5034,24 @@ function inet_dtop($bin)
 	global $db_type;
 
 	if (empty($bin))
-		$ip_address = '';
+		return '';
 
 	// An unpacked IPv4 address? Just return it.
-	elseif (filter_var($bin, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4) !== false)
-		$ip_address = $bin;
+	if (filter_var($bin, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4) !== false)
+		return $bin;
 
 	// A packed IP address? Unpack it.
 	// Need to check this first because a packed IP might take a form that validates as an IPv6.
-	elseif (filter_var(@inet_ntop($bin), FILTER_VALIDATE_IP) !== false)
-		$ip_address = inet_ntop($bin);
+	$unpacked = @inet_ntop($bin);
+	if ($unpacked !== false)
+		return $unpacked;
 
 	// An unpacked IPv6 address? Just return it.
-	elseif (filter_var($bin, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6) !== false)
-		$ip_address = $bin;
+	if (filter_var($bin, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6) !== false)
+		return $bin;
 
 	// No valid IP address? Return an empty string.
-	else
-		$ip_address = '';
-
-	return $ip_address;
+	return '';
 }
 
 /**
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -5033,15 +5033,12 @@ function inet_dtop($bin)
 {
 	global $db_type;
 
-	// No IP address given?
-	if(empty($bin))
-		$ip_address =  '';
-
 	// An unencoded IPv4 address? Just return it.
-	elseif (filter_var($bin, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4) !== false)
+	if (filter_var($bin, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4) !== false)
 		$ip_address = $bin;
 
-	// A binary encoded IPv6 address? Decode it. Need to check this first because a binary encoded IPv6 can also be a valid IPv6.
+	// A binary encoded address? Decode it.
+	// Need to check this first because a binary encoded IP can also be a valid IPv6.
 	elseif (filter_var(@inet_ntop($bin), FILTER_VALIDATE_IP) !== false)
 		$ip_address = inet_ntop($bin);
 
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -5033,8 +5033,11 @@ function inet_dtop($bin)
 {
 	global $db_type;
 
+	if (empty($bin))
+		$ip_address = '';
+
 	// An unpacked IPv4 address? Just return it.
-	if (filter_var($bin, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4) !== false)
+	elseif (filter_var($bin, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4) !== false)
 		$ip_address = $bin;
 
 	// A packed IP address? Unpack it.
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -5027,7 +5027,7 @@ function inet_ptod($ip_address)
 
 /**
  * @param binary $bin An IP address in IPv4, IPv6 (Either string (postgresql) or binary (other databases))
- * @return string The IP address in presentation format or false on error
+ * @return string The IP address in presentation format or an empty string on error
  */
 function inet_dtop($bin)
 {
@@ -5046,9 +5046,9 @@ function inet_dtop($bin)
 	elseif (filter_var($bin, FILTER_VALIDATE_IP, FILTER_FLAG_IPV6) !== false)
 		$ip_address = $bin;
 
-	// No valid IP address? Return false.
+	// No valid IP address? Return an empty string.
 	else
-		$ip_address = false;
+		$ip_address = '';
 
 	return $ip_address;
 }
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -5042,7 +5042,7 @@ function inet_dtop($bin)
 		$ip_address = $bin;
 
 	// A binary encoded IPv6 address? Decode it. Need to check this first because a binary encoded IPv6 can also be a valid IPv6.
-	elseif (filter_var(@inet_ntop($bin), FILTER_VALIDATE_IP, FILTER_FLAG_IPV6) !== false)
+	elseif (filter_var(@inet_ntop($bin), FILTER_VALIDATE_IP) !== false)
 		$ip_address = inet_ntop($bin);
 
 	// An unencoded IPv6 address? Just return it.
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -5036,15 +5036,22 @@ function inet_dtop($bin)
 	// No IP address given?
 	if(empty($bin))
 		$ip_address =  '';
-	// A IPv4 address? Just return it.
+
+	// An unencoded IPv4 address? Just return it.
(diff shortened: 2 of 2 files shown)


albertlast:
This will bring back the problem of https://github.com/SimpleMachines/SMF2.1/issues/3504
so my recommandation do not merge this.
The problem of #3548 is that ip adress is hardcoded and not this function.

MissAllSunday:
Can a token can be used instead of hardcoding the IP?  something like {ip} or similar that can be dynamically changed to a suitable var?

albertlast:
Yes this idea i already explain in the issue https://github.com/SimpleMachines/SMF2.1/issues/3548#issuecomment-243973860
this would be the right way.

MissAllSunday:
What about a wildcard like {localip}  or {localhost}  and let the appropriate DB file to fill it accordingly?   much like other wildcards such as {literal} or {empty}

The IP for the initial message on install.php can be harcoded as thats the only place it is used.  We can instruct mod authors to use the wildcard if they are planning on do some kind of fake posting.

albertlast:
yeah something like that.

Sesquipedalian:
That's all well and good for solving the problem of manually entering `127.0.0.1` as the IP for the "Welcome to SMF" message. But there are mods that create table entries that record IP addresses, and mods will call this function and pass it unpacked string values. If, as @albertlast says, a packed IPv6 address can also be a valid IPv6 address, then this PR as it stands now will not be good enough for filtering IPv6 addresses given as raw strings. But we need something to bulletproof this function so that it can handle unpacked IP addresses.

Sesquipedalian:
I've just committed a change so that filter_var is only used to detect IPv4 addresses. For now I've left the logic for dealing with IPv6 addresses as simply "If this is a PostgreSQL database, assume $bin is already an unencoded IPv6 address. Otherwise, decode $bin using inet_ntop()."

Sesquipedalian:
For the record, I still think we need a different way to check whether $bin is an encoded or decoded IPv6. Merely asking what database we are using is not reliable at all.

Sesquipedalian:
Hm. Looking at #3504, I think PortaMX was onto something. What she did wasn't good enough by itself, but if used in concert with filter_var(), we might get something pretty good.

Sesquipedalian:
With 89731b5, the only way to get a false positive is if $bin is a valid, unpacked IPv6 address that just so happens to produce another valid IPv6 address when run through inet_ntop(). That is a vanishingly rare occurrence, and the worst that can happen is that we'll return an incorrect (but still valid) IPv6 address. And since even then the incorrect value will be consistent, any code that relies on the inet_dtop() function (e.g. to enforce bans, etc.) will still work.

Of course, I also agree that using wildcards like {localip} or {localhost} would be a good idea. The changes in this PR are instead meant to defend against mods and whatnot that might not pack IP address values to binary as expected.

albertlast:
I don't like the solution:
first
as mod dev you know if the ip adress is a string or binary ( we do in smf in the same way)
if your ip adress is a string everything is fine -> use inet as query datatype
if you ip adress is binary use this function

secondly
unoptimize runtime for pg

thirt
to many different output behavior of this function

four
you got then different datatype in the db some varbinary with binary ipv4 and ipv6
and some ipv4 with string
then you got many problem to work with this --> you need to convert this data once

From my side the "issue" doesn't exists which you try to solve.
When you like to solve the last issue "existing mod data converting" than feel free,
to place a converting function for the mod dev.
So that them got a standard way to solve this issue.

Another solution would be to create a second function,
maybe with the name inet_ntop_safe() (or different...)
and place there your logic.

Sesquipedalian:
I don't understand the reasoning for your objection. :)

> first
> as mod dev you know if the ip adress is a string or binary ( we do in smf in the same way)
> if your ip adress is a string everything is fine -> use inet as query datatype
> if you ip adress is binary use this function

My changes amount to basic defensive coding practices. We simply cannot be sure that the incoming data will always be in the expected format. If we could, there would be no need for this function to begin with. Since the incoming data string might or might not be packed into a binary format, we need to test whether it is or not. Simply checking what sort of database we are using as a proxy for checking the actual data is not reliable. _We need to test the **actual data**_.

> secondly
> unoptimize runtime for pg

I don't understand what you mean. Could you elaborate on that a bit for me, please?

> third
> to many different output behavior of this function

The output is always either a valid IP address, or `false` if the incoming data is not a valid IP address. It is very consistent.

> four
> you got then different datatype in the db some varbinary with binary ipv4 and ipv6
> and some ipv4 with string
> then you got many problem to work with this --> you need to convert this data once

By all means, convert the data to the preferred type during upgrade. That's a very good idea. But even if we do, we cannot assume that the data passed to this function will always be in the proper format.

> From my side the "issue" doesn't exists which you try to solve.
> When you like to solve the last issue "existing mod data converting" than feel free,
> to place a converting function for the mod dev.
> So that them got a standard way to solve this issue.

But doing the check here causes no harm whatsoever. I honestly cannot understand why you object. If the incoming data is in the correct format (a packed string for MySQL, or an unpacked string for PostgreSQL), this code will return the desired output. My changes do nothing whatsoever to interfere with your desired output. All my changes do is ensure that if someone else does something wrong, we can recover gracefully.

> Another solution would be to create a second function,
> maybe with the name inet_ntop_safe() (or different...)
> and place there your logic.

Why do that when we can put the test here in this function?

albertlast:
To make it short,
You make the function safe not smf,
because you try to hide real issue.
The real issue there are strings in a varbinary field.
The behaviour of both is highly different when you try to select them by sql.
String doesn't support between operation.

This is the reason I don't support in this pr.

---

Gesendet: 09.09.2016 06:31
An: SimpleMachines/SMF2.1
Cc: albertlast; Mention
Betreff: Re: [SimpleMachines/SMF2.1] Handle IP address strings in inet_dtop() (#3570)

I don't understand the reasoning for your objection. :)

first
as mod dev you know if the ip adress is a string or binary ( we do in smf in the same way)
if your ip adress is a string everything is fine -> use inet as query datatype
if you ip adress is binary use this function

My changes amount to basic defensive coding practices. We simply cannot be sure that the incoming data will always be in the expected format. If we could, there would be no need for this function to begin with. Since the incoming data string might or might not be packed into a binary format, we need to test whether it is or not. Simply checking what sort of database we are using as a proxy for checking the actual data is not reliable. We need to test the actual data.

secondly
unoptimize runtime for pg

I don't understand what you mean. Could you elaborate on that a bit for me, please?

third
to many different output behavior of this function

The output is always either a valid IP address, or false if the incoming data is not a valid IP address. It is very consistent.

four
you got then different datatype in the db some varbinary with binary ipv4 and ipv6
and some ipv4 with string
then you got many problem to work with this --> you need to convert this data once

By all means, convert the data to the preferred type during upgrade. That's a very good idea. But even if we do, we cannot assume that the data passed to this function will always be in the proper format.

From my side the "issue" doesn't exists which you try to solve.
When you like to solve the last issue "existing mod data converting" than feel free,
to place a converting function for the mod dev.
So that them got a standard way to solve this issue.

But doing the check here causes no harm whatsoever. I honestly cannot understand why you object. If the incoming data is in the correct format (a packed string for MySQL, or an unpacked string for PostgreSQL), this code will return the desired output. My changes do nothing whatsoever to interfere with your desired output. All my changes do is ensure that if someone else does something wrong, we can recover gracefully.

Another solution would be to create a second function,
maybe with the name inet_ntop_safe() (or different...)
and place there your logic.

Why do that when we can put the test here in this function?

##

Sesquipedalian:
Yes, it is bad if there are strings in a varbinary field. I am not arguing with that.

But the solution to that is to create a filter on data going **into** the database to make sure the data is saved in the correct format.

In this function we are dealing with data coming **out** of the database. The format in which the data was stored in the database is no longer relevant here. What is relevant here is whether PHP can use the data. In this function, it doesn't matter if the data was stored correctly or not. We are no longer dealing with the storage. We are no longer dealing with the database at all. We are only dealing with the data itself.

In short, if you are worried about data integrity and making sure that we store data in the correct format (and it is good to be worried about that), the right place to do that is in the `inet_ptod()` function, not the `inet_dtop()` function.

Sesquipedalian:
Think of it like this: the database is like the fuel storage tank at a gas station, and PHP is like a car, and the data is like the fuel itself.

We want the data to be clean at all times, just like we want the gasoline to be clean at all times. We want the fuel to be clean inside the storage tank, and we want the fuel going into the car to be clean. So, we create two filters:
1. A filter that ensures that fuel going into the storage tank is clean. **This is what you feel most concerned with, @albertlast.**
2. A filter in the car's fuel intake that ensures the fuel the car takes in is clean. **This is what the changes in this PR are concerned with.**

Because the driver of the car can never be sure that the fuel in the storage tank really is clean, it is necessary to have a filter on the car. When you insist that we should not merge this PR, it is like a gas station owner insisting that the car should not have a fuel filter because the fuel in the storage tank is supposed to be clean. That does not make sense. What does make sense is to have **both** filters in place and active.

albertlast:
your are right about output filter,
Because your are right you solution is wrong.
In smf exists no db output validation,
So when you like feel free to implement this.

But this function is the not responsible for this job.

So you can waste your time on this topic or fix the real issue.
Real issue are:
The installer do the wrong thing
There is no validation for db output

but here i see no solution.

---

Gesendet: 09.09.2016 07:32
An: SimpleMachines/SMF2.1
Cc: albertlast; Mention
Betreff: Re: [SimpleMachines/SMF2.1] Handle IP address strings in inet_dtop() (#3570)

Think of it like this: the database is like the fuel storage tank at a gas station, and PHP is like a car, and the data is like the fuel itself.

We want the data to be clean at all times, just like we want the gasoline to be clean at all times. We want the fuel to be clean inside the storage tank, and we want the fuel going into the car to be clean. So, we create two filters:
1.  A filter that ensures that fuel into the storage tank is clean. This is what you feel most concerned with, @albertlasthttps://github.com/albertlast.
2.  A filter in the car's fuel intake that ensures the fuel the car takes in is clean. This is what the changes in this PR are concerned with.

Because the driver of the car can never be sure that the fuel in the storage tank really is clean, it is necessary to have a filter on the car. When you insist that we should not merge this PR, it is like a gas station owner insisting that the car should not have a fuel filter because the fuel in the storage tank is supposed to be clean. That does not make sense. What does make sense is to have both filters in place and active.

##

Sesquipedalian:
Yes, the installer does the wrong thing, and that needs to be changed. But that is one specific instance of a more general problem. There are other ways that we might end up with improperly formatted IP addresses coming from the database. Since this is a general problem, we need a general solution. That is what this PR provides.

You see "no solution" because you think that this PR is meant to solve a different problem than it is actually meant to solve. This PR is **not** solving the problem of incorrectly stored data. That is a different problem that needs a different solution. This PR **is** solving the problem of the `inet_dtop()` function not handling the variably formatted data it claims to be able to handle. The `inet_dtop()` function claims to be able to handle IP addresses in either packed (i.e. binary) or unpacked strings. Without this PR, the function does not do that reliably. With the code in this PR, it does do that reliably. The problems happening with storage in the database are a separate issue. Regardless whether we eventually solve all the storage problems, this PR should be included anyway, so that the function will do what it says it will do.

albertlast:
When the database output would be there this change could be good,
But this is a wanted behaviour to fund the issue where program does insert data without convert this in the right format.

Without the first thing is this pr dangerous because is hide the real issues.

---

Gesendet: 09.09.2016 08:20
An: SimpleMachines/SMF2.1
Cc: albertlast; Mention
Betreff: Re: [SimpleMachines/SMF2.1] Handle IP address strings in inet_dtop() (#3570)

Yes, the installer does the wrong thing, and that needs to be changed. But that is one specific instance of a more general problem. There are other ways that we might end up with improperly formatted IP addresses coming from the database. Since this is a general problem, we need a general solution. That is what this PR provides.

You see "no solution" because you think that this PR is meant to solve a different problem than it is actually meant to solve. This PR is not solving the problem of incorrectly stored data. That is a different problem that needs a different solution. This PR is solving the problem of the inet_dtop() function not handling the variably formatted data it claims to be able to handle. The inet_dtop() function claims to be able to handle IP addresses in either packed (i.e. binary) or unpacked strings. Without this PR, the function does not do that reliably. With the code in this PR, it does do that reliably. The problems happening with storage in the database are a separate issue. Regardless whether we eventually solve all the storage problems, this PR should be included anyway, so that the function will do what it says it will do.

##

Sesquipedalian:
That's like saying that it is better not to have a fuel filter in my car so that you can use my car exploding as an indication to you that you should install a better filter on your storage tank.

This PR is not hiding "the real issue." There are **two** "real issues" here, and this is dealing with the other one. Specifically, the `inet_dtop()` function is broken. It doesn't do what it says it will do. This PR makes it do what it says it will do. If the symptoms of the storage issue are covered up by fixing this function, that doesn't mean we shouldn't fix the function. That means we should not forget to also fix the storage issue. These are two separate problems that **both** need to be fixed. Not fixing one problem because it makes it easier to remember to fix the other is silly. If you need something to remind you to fix the storage problem, create a new issue report for it. That's what the issue tracker is for.

Sesquipedalian:
To be clear, the issue in #3548 is **triggered** by the storage problem, but the **cause** of #3548 is that the `inet_dtop()` function is broken.

albertlast:
With your solution the problem of the setup would be raised up later/to late -> this pr hide other issue.

This keep this pr dangerous.

---

Gesendet: 09.09.2016 08:37
An: SimpleMachines/SMF2.1
Cc: albertlast; Mention
Betreff: Re: [SimpleMachines/SMF2.1] Handle IP address strings in inet_dtop() (#3570)

That's like saying that it is better not to have a fuel filter in my car so that you can use my car exploding as an indication to you that you should install a better filter on your storage tank.

This PR is not hiding "the real issue." There are two "real issues" here, and this is dealing with the other one. Specifically, the inet_dtop() function is broken. It doesn't do what it says it will do. This PR makes it do what it says it will do. If the symptoms of the storage issue are covered up by fixing this function, that doesn't mean we shouldn't fix the function. That means we should not forget to also fix the storage issue. These are two separate problems that both need to be fixed. Not fixing one problem because it makes it easier to remember to fix the other is silly. If you need something to remind you to fix the storage problem, create a new issue report for it. That's what the issue tracker is for.

##

Sesquipedalian:
That's simply false. The storage issue has already been raised. You have stated it quite clearly already. You and @MissAllSunday have already discussed ways to fix the storage issue by fixing the installer. Just don't forget to do what you have already decided to do, and you will fix it.

But even if it somehow were true that merging this PR would somehow make everyone forget that the storage issue existed, that isn't a reason not to merge this PR. This PR is not dangerous. It fixes a problem that needs to be fixed regardless of what happens with the database storage issue. Conversely, even once you do fix the installer and the storage issue, this PR will still be needed, because without it the `inet_dtop()` function will still be broken.

It is true that forgetting to fix the storage issue would be dangerous, but again, this PR will not make you or anyone else forget the storage issue. The only thing that will make you forget the storage issue is if you have a bad memory or if you fail to make a report for it in the issue tracker.

albertlast:
the storage problem wouldn't raised up when this patch would be already in place.
the older version of this function already got somthing simliar and this is the reason that it take so long,
until this bug raised up.

Here the benchmark:

```
mysql bin ipv4            2.173054933548
mysql bin ipv4 safe       4.7309100627899
postgresql bin ipv4       1.148197889328
postgresql bin ipv4 safe  1.9698660373688
mysql bin ipv6            2.0489330291748
mysql bin ipv6 safe       4.5167911052704
postgresql bin ipv6       1.1420891284943
postgresql bin ipv6 safe  6.8907370567322
```

http://pastebin.com/i9qYvaS3

in short your function is 2x - 6x slower as the existing version.
And this only for one special case that the wrong data is inserted by someone.

Sesquipedalian:
Maybe it is slower. It is also **_not broken_**. Your objections to fixing the function make no sense.

albertlast:
Btw please fix the null/empty handling.
The database field can be null,
but smf need an empty string.

Sesquipedalian:
No problem. Since the function's documentation wrongly states that the function should return false on error, I'll fix that as well.

albertlast:
Like I saied this no error when the binary is empty.

Sesquipedalian:
Of course. I am simply using it in the sense that you yourself meant it when you originally wrote 'or false on error' in the documentation. :)

Sesquipedalian:
I've done some optimizations for performance on this code. Unfortunately, handling an unpacked IPv6 address still takes a relatively long time compared to other scenarios, but I don't see any way around that. Even so, I don't think speed of execution is a relevant concern here. A _single_ execution of `parse_bbc()` takes several orders of magnitude more time to complete than this function. If we are looking for speed optimizations, this is not a very important function to look at.

albertlast:
To make a recape,
you want a function which bevahior is different and for different use.
Now you try to change an existing function to fit this needs.
And this bad, i mean really bad.

This function is for convert ip field (which could be varbinary or inet) db output into string and
for this porpose special optimize.
When need something different create an own function how fit this need,
don't use this function for your needs.

There is now function for you need and this function is the wrong place.

Sesquipedalian:
No, I'm not trying to make this function do anything different than what it already is supposed to do. You aren't listening to what I have been saying, and therefore you misunderstand what I am doing.

I am making this function actually check **the data itself** to see what it really is. Your version makes assumptions about the data based on nothing more than the value of `$db_type`. Making assumptions like that is bad programming. When the function receives the data, it needs to check the data. Failure to do so is just begging for errors to arise. If you refuse to fix this now, eventually someone will have to fix it anyway.

The right thing to do is to test the value of `$bin` itself. Testing the value of `$db_type` and then making guesses about the value of `$bin` is the wrong thing to do. We need to replace `($db_type == 'postgresql')` with something that tests the value of `$bin`. Until that is changed, this function remains bad code.

albertlast:
Sure your do different,
you want this function work not only for db converting and
your input can be binary or string for mysql and pg.
But binary is for pg wrong and string is wrong for mysql.

So you need to write an own function.

There is no need to check the data when the data is correct and
your are not able to insert wrong data in pg -> no need to check.

Sesquipedalian:
> your are not able to insert wrong data in pg

I did not know about this. That changes the situation, and I no longer see a reason to keep this PR open. What you are doing in #3574 is better, @albertlast. :)

PR #3571, opened 2016-09-08
Fixes #3453

### LINKED ISSUE #3453: template_button_strip inconsistencies [minor]
Files (where issue was spotted):
- BoardIndex.php
- Compat.template.php

The button array generated circa line 130 in BoardIndex.php uses a "lang" (bool) and a "text" (string) key - presumably with the intention that the lang key being false would have the "text" key directly displayed, whereas an unset or true value would have $txt[text key] displayed instead.

Compat.template.php defines the template_button_strip function that is used to display the button. However, this uses the  $txt array and pays no attention to the "lang" key whatsoever.

If enforcing the best practice of language strings then the lang key could be removed in BoardIndex.php (and any other locations where the function is called - Recent, Memberlist, Reports, MessageIndex, Display, Xml, Calendar, index). Usually the button array is defined in the source file, not template. Conversely, if allowing direct text strings then it'd be simpler to change the definition in Compat.template.php

**Also,** the test key is used in Calendar.php (202) referring to "can_post" as a string - the template_button_strip function only checks the issetness and emptynes of the test key - it doesn't pass it through allowedTo - template_button_strip should probably do the allowedTo test if the value is a string (or array).

No SMF mods should be using the lang key as false (as per mod guidelines) so it _shouldn't_ break any mods.

Sesquipedalian:
- Compat.template.php serves no purpose any more, since its only function was to provide a version of the `template_button_strip()` function for really old themes. Support for those really old themes is dropped in SMF 2.1, so we no longer need this.
- The 'test' checks for all standard button strips have been turned into contextual checks at the time the button strip array is created, as in #3557.
- The 'lang' key for buttons appears to be a vestigial element from old versions of SMF that no longer serves any purpose. It has therefore been removed.


MissAllSunday:
Awesome, thanks!

PR #3573, opened 2016-09-09
More urlbbc

### LINKED ISSUE #1389: Auto linking URLs doesn't handle non ASCII characters
Complex issue.
http://dev.simplemachines.org/mantis/view.php?id=360
http://dev.simplemachines.org/mantis/view.php?id=2449
http://dev.simplemachines.org/mantis/view.php?id=4359
http://www.simplemachines.org/community/index.php?topic=387983

Using the url bbcode is fine, it's merely the auto linking regex.

### LINKED ISSUE #1565: Schemaless links and images
With the growing trend for schemaless links (beginning solely with // rather than http:// or https:// to avoid security issues), it would be very handy if it were actually possible to use these in the img or i/url bbcodes as currently it is not.

### LINKED ISSUE #3575: subs.php error
i have been informed that php version 5.4.16 has an issue.

Fatal error: Can't use function return value in write context in C:\wamp\www\github2.1\Sources\Subs.php on line 1278

Call Stack:
 0.0270 871424 1. {main}() C:\wamp\www\github2.1\upgrade.php:0
 0.1510 885224 2. loadEssentialData() C:\wamp\www\github2.1\upgrade.php:132

Sesquipedalian:
1. Bug fixes
   - Fixes an unreported bug when autolinking `mailto:` URLs
   - Fixes #3575 
2. Supports schemeless URLs (Fixes #1565)
   - Raw domain names (e.g. `www.example.com`) and schemeless URLs (e.g. `//www.example.com`) are linked without specifying the `http:` scheme. This applies both in BBCode (e.g. `[url]` and `[iurl]`) and in the autolinker. 
3. Adds support for IRIs in the autolinker (Fixes #1389)
   - Changes the regex to permit Unicode characters in relevant parts of a detected URL. HTML5 supports IRIs, so it is perfectly valid to put UTF8 characters (without percent encoding them) into a web address in an HTML5 document. Whenever an IRI link is encountered, the browser will handle everything that needs to be handled. See http://www.w3.org/International/articles/idn-and-iri/ for more info.
   - Removes some calls to filter_var() because FILTER_SANITIZE_URL and FILTER_VALIDATE_URL only support URIs, not IRIs. Since we no longer have the filters in place as a secondary check for invalid URLs, we'll have to rely on the regex alone. But that should be fine; it's no different than the situation prior to commit 8db70d2.

### DIFF
--- a/Sources/Subs-Post.php
+++ b/Sources/Subs-Post.php
@@ -439,7 +439,7 @@ function fixTag(&$message, $myTag, $protocols, $embeddedUrl = false, $hasEqualSi
 
 		if (!$found && $protocols[0] == 'http')
 		{
-			if (substr($replace, 0, 1) == '/')
+			if (substr($replace, 0, 1) == '/' && substr($replace, 0, 2) != '//')
 				$replace = $domain_url . $replace;
 			elseif (substr($replace, 0, 1) == '?')
 				$replace = $scripturl . $replace;
--- a/Sources/Subs-Post.php
+++ b/Sources/Subs-Post.php
@@ -449,7 +449,7 @@ function fixTag(&$message, $myTag, $protocols, $embeddedUrl = false, $hasEqualSi
 				$this_tag = 'iurl';
 				$this_close = 'iurl';
 			}
-			else
+			elseif (substr($replace, 0, 2) != '//')
 				$replace = $protocols[0] . '://' . $replace;
 		}
 		elseif (!$found && $protocols[0] == 'ftp')
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -1275,7 +1275,8 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 				{
 					if (isset($disabled['url']))
 						$tag['content'] = '$1';
-					if (empty(parse_url($data[0], PHP_URL_SCHEME)))
+					$scheme = parse_url($data[0], PHP_URL_SCHEME);
+					if (empty($scheme))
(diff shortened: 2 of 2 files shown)


MissAllSunday (commented):
Lets try this new review thing :P 

The only issue I see is parse_url() not been able to handle schemaless urls on php lower than 5.4.7.

I tested it on PHP 5.3.29 and it logs an error:  Undefined variable: fullUrl

on File: /public_html/Sources/Subs.php
Line: 1928

This is when trying to autolink a schemaless url:

http://missallsunday.com/secret/
www.missallsunday.com
//missallsunday.com/

Sesquipedalian:
I forgot to add an `else` statement in da643bc. That error probably would have popped up on any PHP version.

Antes:
Weird. I can't add screenshot to review section FeelsBadMan.

capture

XinYenFon (approved):
Didn't find any problem in PR. Tested with different link schemes. Following links used in the test;

```
http://www.simplemachines.org
https://www.simplemachines.org
//simplemachines.org
//www.simplemachines.org
www.simplemachines.org
```

Sesquipedalian:
> The only issue I see is parse_url() not been able to handle schemaless urls on php lower than 5.4.7.

I've checked everything over, and I don't think that PHP bug should affect anything in this PR. [According to the manual](http://php.net/manual/en/function.parse-url.php#example-4673), the bug affected what ended up in the `path` value when parsing a schemeless URL. The `scheme` value will still be empty, and that's all I deal with in this code.

PR #3574, opened 2016-09-10
better error handling for inet_dtop

albertlast:
It's supress the warning and safe it in the smf log.
With less performance impact as #3570.

```
mysql bin ipv4           2.2889740467072
mysql bin ipv4 safe      4.1099700927734
postgresql bin ipv4      1.2278430461884
postgresql bin ipv4 safe 1.2439939975739
mysql bin ipv6           2.1695320606232
mysql bin ipv6 safe      4.0392029285431
postgresql bin ipv6      1.2550032138824
postgresql bin ipv6 safe 1.2721769809723
```

http://pastebin.com/ZraQt3eu

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -4982,9 +4982,31 @@ function inet_dtop($bin)
 	if ($db_type == 'postgresql')
 		return $bin;
 
-	$ip_address = inet_ntop($bin);
+	//http://stackoverflow.com/questions/1241728/can-i-try-catch-a-warning
+	set_error_handler(function($errno, $errstr, $errfile, $errline, array $errcontext)
+		{
+			// error was suppressed with the @-operator
+			if (0 === error_reporting())
+			{
+				return false;
+			}
 
-	return $ip_address;
+			throw new ErrorException($errstr, 0, $errno, $errfile, $errline);
+		}
+	);
+
+	try
+	{
+	 $var = inet_ntop($bin);
+	 restore_error_handler();
+	 return $var;
+	}
+	catch(ErrorException $e)
+	{
+	    log_error($e->getMessage(),'general', $e->getFile(), $e->getLine());
+		restore_error_handler();
+		return false;
+	}
 }
 
 /**
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -4969,8 +4969,9 @@ function inet_ptod($ip_address)
 }
 
 /**
+ * Convert the ip field of a database output in a string
  * @param binary $bin An IP address in IPv4, IPv6 (Either string (postgresql) or binary (other databases))
- * @return string The IP address in presentation format or false on error
+ * @return string The IP address in presentation format, null will be an emtpy string or false on error
  */
 function inet_dtop($bin)
 {
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -5004,7 +5004,7 @@ function inet_dtop($bin)
 	}
 	catch(ErrorException $e)
 	{
-	    log_error($e->getMessage(),'general', $e->getFile(), $e->getLine());
+	    log_error($e->getMessage().'<br/>'.$e->getTraceAsString(),'general', $e->getFile(), $e->getLine());
 		restore_error_handler();
 		return false;
 	}


Sesquipedalian:
But @albertlast, the problem with the function is not that we don't want to see an error message. The problem is that the function does not reliably do its job. Where an error message gets logged is beside the point.

albertlast:
It's do perfect his job,
it's convert the output from db into string.
When the function get a string instead of an binary with mysql is this an error.
So this implentation do the right thing.

Sesquipedalian:
But the job of this function is not "to convert the output from the db into string." It's job is to convert `$bin` into a presentable IP address. The data in `$bin` could come from anywhere. Any other function might call `inet_dtop()` and feed it some data. In plain SMF 2.1, it is only called from code that you have written. But once mods get involved, there is no telling where the mod will have gotten the data that they want to pass to `inet_dtop()`. So, `inet_dtop()` needs to be able to handle all the use cases it claims to be able to handle.

albertlast:
And here you see the different understanding of this function.
this function is for converting db ip field to string fast as possible.
When you need something similar create an own function and don't use this function.

---

Gesendet: 10.09.2016 08:22
An: SimpleMachines/SMF2.1
Cc: albertlast; Mention
Betreff: Re: [SimpleMachines/SMF2.1] better error handling for inet_dtop (#3574)

But the job of this function is not 'to convert the output from the db into string'. It's job is to convert $bin into a presentable IP address. The data in $bin could come from anywhere. Any other function might call inet_dtop() and feed it some data. In plain SMF 2.1, it is only called from code that you have written. But once mods get involved, there is no telling where the mod will have gotten the data that they want to pass to inet_dtop(). So, inet_dtop() needs to be able to handle all the use cases it claims to be able to handle.

##

Sesquipedalian:
1. Do not confuse (a) what you as a human being intended this function to be used for, with (b) what the function instructs the computer to do. You wrote the function so that you could convert IP addresses retrieved from the database to a presentation format. Fine. But the computer doesn't know anything about what you intended. All it knows is what the instructions actually say. And this function doesn't say anything about "retrieved from the database." This function only says, "Turn this chunk of data into a presentable IP address." It doesn't matter what you intended the function for. The instructions you actually wrote are all that matter, and right now, those instructions are broken. If you want to make sure that this function will only be used for that one specific task, you need to write more and better instructions in the function.
2. I don't want or need another function. I want to make sure that this function validates the data in `$bin` before feeding it to `inet_ntop()`. That's it. That's all. In my PR, I also added some code to try to recover gracefully if `$bin` contains bad data, but that was just extra pleasantries. What fundamentally needs to happen is that `if ($db_type == 'postgresql')` needs to be replaced with a test that will reliably validate the real, actual data before acting on it. If you can come up with a solution that works better and faster than what I came up with, by all means do so. But unless and until that inadequate test is replaced with a proper one, this function will remain broken and poorly coded.

albertlast:
Update the documentation to make cleare the usecase of this function.

Sesquipedalian:
This change in documentation makes your intended use case clearer. That's good.

Unfortunately, this version of the function still does not test the value of `$bin` before acting on it. Until you replace `($db_type == 'postgresql')` with something that checks the actual data in `$bin`, this function will remain poorly coded.

Sources/Subs.php
```diff
@@ -4982,9 +4983,31 @@ function inet_dtop($bin)
 	if ($db_type == 'postgresql')
```
Sesquipedalian, line 4983: Testing the value of `$db_type` is not a reliable way to know the value of `$bin`. You need to replace this test with something that tests the value of `$bin`.

albertlast: There is no need to test, when the output came from pg
because your are not able to save wrong data in pg-> no need for checks.

Sesquipedalian: For the sake of argument, let's pretend for a moment that it is somehow impossible to write data in the wrong type to a column in PostgreSQL. If we pretend that, then here is the logic of the function as it is now:

Are we using a PostgreSQL database?
1. Yes.
   - Then assume $bin contains a plain IP address in a string, and return it. 
2. No.
   - Then assume $bin contains an IP address that has been packed into binary format, and blindly feed it to inet_ntop().

The problem here is that you are **_assuming_** that using MySQL somehow guarantees that $bin is an IP address in binary format. Using MySQL does **not** guarantee that at all. As we've seen in the installer already, it is easy for data in the wrong type to be written to this column, and therefore there is no guarantee that $bin really will be what it should be. You must do something to check if $bin really is what it should be.

Of course, if we don't pretend that it is somehow impossible for PostgreSQL to contain bad data, then we need to check $bin in that case, too.

In short, just because the data _should be_ in a certain format does not mean that it _is_ in that format. You must check if it is or not.

Sesquipedalian: If we pretend that using PostgreSQL somehow guaranteed that `$bin` could never carry incorrectly formatted data, your code would at least need to look like this

```
if ($db_type == 'postgresql')
    return $bin;
else
{
    // code to test whether $bin is a normal string or has been packed into binary format
}
```

Sesquipedalian:
> your are not able to save wrong data in pg

I did not know about this feature of PostgreSQL. That makes a significant difference. Since that is the case, I recognize that what you are doing in this PR is better than what I was doing in #3570.

Also, as I look over my previous comments here, I realize that I wasn't always as courteous in my tone as I should have been, so I'd like to apologize to you for that, @albertlast. Regardless what the better way to code this was, I should not have let my frustration lead me to being rude. Sorry about that.

MissAllSunday:
Isn't there another way to handle this? it seems hackish to try and catch a warning, isn't there a way to prevent the warning?

albertlast:
sure i could prevent the warning with @ but i want the information if something goes wrong.
And than more information for logging where the shit goes wrong.
from my point of view i didn't understand that smf got already an own error handler

MissAllSunday:
By prevent I meant finding a way for the error to not happen not to  
suppress it.  Why does the error happen?

## 

albertlast:
The error happen when someone like the installer write ip string directly to the mysql database,
to finde this issue I want to log this information.

MissAllSunday:
Then wouldn't it be easier to prevent that from happen?

We can instruct mod authors to stop hardcoding the IP  or set a check  
before anything gets saved into the DB, check if theres a hardcoded string  
and convert it to a suitable value according to the DB engine being used.

## 

albertlast:
the checks are there,
but when you write directysl to the database than the checks are ingored and
this function didn't got enought information to fix this after wards.

MissAllSunday:
Then something is wrong with those checks,  all our db insert operations  
are done via $smcFunc['db_insert']  and $smcFunc['db_query']  for  
updating.  We can add/reinforce any checks there  or use a wildcard like I  

## suggested.

albertlast:
nope there is no wrong,
when you look at the install script you see there are not done by db_insert

MissAllSunday:
OK, if the issue is the hardcoded IP on install, we can either remove that  
or replace it with a suitable value for each db engine, or just leave it  
empty,  how does the ip_ functions will react if the IP is empty?

## 

albertlast:
empty is an allowed value

MissAllSunday:
OK then, can we try leaving the IP field empty when inserting the first  
message to the DB?

## 

albertlast:
empty = null yes no problem,
like i said for the install thing would by valid place holder a better idea.

Sesquipedalian:
> Isn't there another way to handle this? it seems hackish to try and catch a warning, isn't there a way to prevent the warning?

Unfortunately, there doesn't seem to be any good way to test `$bin` except by running it through `inet_ntop()` and seeing what happens.

> sure i could prevent the warning with @ but i want the information if something goes wrong.
> And than more information for logging where the shit goes wrong.
> from my point of view i didn't understand that smf got already an own error handler

What about something like this?

```
$unpacked = @inet_ntop($bin);
if ($unpacked !== false)
    return $unpacked;
else
{
    log_error('IP address "' . $bin . '" is stored with an incorrect datatype in the database.');
    return '';
}
```

Then the error message would point to the real issue (an incorrectly cast value in the database) rather than looking like a problem in the code. Plus, you could avoid having to use the try/catch workaround.

Sesquipedalian:
Or even this?

```
$unpacked = @inet_ntop($bin);
if ($unpacked !== false)
    return $unpacked;
else
{
    log_error('IP address "' . $bin . '" is stored with an incorrect datatype in the database.');

    if (filter_var($bin, FILTER_VALIDATE_IP) !== false)
        return $bin;
    else
        return '';
}
```

 This way, you'll still get the warning about the database problem, but code that relies on getting a valid IP address will still work. It is important for, e.g., enforcing bans that we do our best to return a valid IP address if we possibly can.

albertlast:
no we don't return ip's when they not valid.
because the data is garbage, when the data is garbage the where operation are not working and
everthing which based on get fucked up.

Sesquipedalian:
Well, `(filter_var($bin, FILTER_VALIDATE_IP) !== false)` will only evaluate to true in the case where someone wrote a valid IP address as a raw string to the database. If the data really is garbage, it won't validate. But maybe I'm missing something (that's already happened once in this discussion).

But even if the extended version doesn't work for reasons I don't understand at this point, what about the shorter version that simply returns an empty string?

albertlast:
Missing information,
where the problem is happen.
End of the day you try to replicate exception function -> than use exception.

Let's come back to the point where @MissAllSunday say is hackish way.
Which good way exists in php tio handling warning?
You can keep them untouched than they display on the display of the user without logging in smf or
supress them but than you don't know where the problem is happen.

With the error handler from stackoverflow,
the dev can >decide< if he want supress the warning(with @),
when not the warning would be collected be the error handler.

Based on this:
I see no problem to raise warning to exception,
even more would i like to see this as standard bevhavior for smf.
Because it lat us implement an better error handling.

MissAllSunday:
By hackish I mean we're treating the symptoms rather then the root of the issue. 

If we can prevent the error from happening then let's do that. 

Sesquipedalian:
> If we can prevent the error from happening then let's do that.

We can fix the problem in the installer easily enough using the method you and @albertlast have already discussed (and we should), and I believe @albertlast has already done the necessary work to make sure that the $smc database functions will only write IP addresses in the correct type to the database. But there is nothing we can do to prevent code outside of our control from writing data in the wrong type. The fundamental problem is that MySQL will accept raw string input for a field that is supposed to be varbinary. The only things we can do about the situation are either (a) to do something to test the data when we retrieve it and try to recover if it is bad, or (b) to say "Don't do that" to mod writers and consider it to be "not our problem" if someone does write raw strings to our varbinary field despite what we told them.

So even once we have done everything we can to make sure that our code never writes bad data to the database, we still need something to deal with the situation if someone else does.

MissAllSunday:
Yes I agree and understand your point, however, theres only too much we can do to prevent third party code from mess things up.

To put an example, theres mods for 2.0 that doesn't use our  $smcFunc DB layer, they're directly using plain mysql functions, meaning they don't work on postgre and theres little to no concern about sanitizing values before being inserted to the DB. Yes, despise all of our efforts to tell them otherwise. 

So, I wouldn't worry too much about that.

Sesquipedalian:
Exactly. And that means that, apart from fixing the installer (which should been done no matter what), the question is, what else we should do?

A. Be friendly to forum admins by trying our best to recover gracefully if some bad code writes the data in the wrong type.
B. Focus on developer concerns by logging an error about the bad data and returning an empty value.
C. Log the error _and_ do our best to recover gracefully.

Choosing solution (A) will mean fewer bug reports once we release SMF 2.1. Users will be generally happier—except, of course, if the presence of incorrectly cast data in their database causes problems some day. But there will be less incentive to mod authors to update their mods to behave properly.

Solution (B) will generate more bug reports once we release. The folks who hang out on the support boards at simplemachines.org will need to get used to saying "You must be using a mod that doesn't follow proper practices." This will give mod authors more incentive to fix their code. But we'll also get complaints about how this causes various features and functions of the forum to break—including rather important ones like bans and anti-spam.

Solution (C) would give us the best of both the other options, in my opinion. We'd still get the error messages that would prompt mod authors to fix their code, but SMF would still function correctly in most cases in the meantime.

Sesquipedalian:
A further thought: perhaps we should include a function that the admin can run from the Maintenance area to find and repair broken IP addresses in the database. Then if a mod or something ever does make a mess this way, the admin can fix it without having to do surgery directly on the database. I believe @albertlast already wrote some code to do this in the upgrader, so maybe that code could be pasted into such a function easily.

albertlast:
a and c are not able to realize because you don't know from where the data came,
is this a database? which table? which row?

An repair function is not so easy i think,
mostly run time,
because you need to select the full table with ip fields and check them in php and than update again.
Could easy run ~10 min.

MissAllSunday:
As much as I like to go the extra mile, offer support and overall be friendly to forum admins, theres only so  much we can do about it. Thing is, if we do take that route then we will also  have to start supporting external modifications, which will add a significant amount of work load to ourselves.

We are also currently focusing on the wrong side of things, we should be encouraging mod authors to write better code rather than trying to fix their code for them or try to hide their mistakes.  Again, try to prevent this to happen rather than patching it, hiding it or otherwise cover it.

If we take the route of trying to cover other people's bad code then those people will continue to write bad code and we will have to add more patches and cover more places, etc, etc. t becomes a vicious circle.

I rather prefer bad code to be exposed and correctly fixed, that way we are attacking the real issue, which is preventing bad code from being written in the first place.

Sesquipedalian:
Okay then. In that case, it looks like the existing code in `inet_dtop()` already does what we want it to do even without the changes in this PR. In fact, it is probably best to stick with SMF's standard error reporting system so that we can keep those error messages prominent.

MissAllSunday:
I'm planning on writing some documentation to try and cover any major change between 2.0 and 2.1, including this one.  Also we will make sure any future mod submitted to our modsite to follow proper and good coding standards

albertlast:
Than how do you like the error handling?
The old version or the new version @MissAllSunday 
When you want top keep the old one,
than i would like to close this pr.

Sesquipedalian:
> An repair function is not so easy i think,
> mostly run time,
> because you need to select the full table with ip fields and check them in php and than update again.
> Could easy run ~10 min.

Maybe, but SMF already includes similar functions for doing things like converting HTML entities to UTF8 characters, which would in theory take even more time to execute. SMF handles that by breaking the database entries up into several batches, and then operating on one batch at a time. I think that using the same approach for fixing malformed IP addresses would work.

MissAllSunday:
I would agree on a repair function, yes, SMF handles the runtime issues by splitting the load and using a simple refresher.

As for the error, I prefer if like it currently is (the old one), to expose the code.

Issue #3383, opened 2016-04-08
smtp mail bug


abolabo:
Hello guys!
Please replace $modSettings['smtp_host'] with getenv('SERVER_NAME') in line
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs-Post.php#L1319
Your code not working with gmail smtp correctly.

live627:
Understand that we cannot just blindly change things around. That causes
bugs.   We don't want that.

On Fri, Apr 8, 2016 at 12:25 PM, Dmitry Suprunenko <notifications@github.com

abolabo:
Yes, i understand it. 
Please read about EHLO and blocking http://faculty.cs.niu.edu/~rickert/cf/bad-ehlo.html
Your code gives smtp-server's URI instead sender host name. it's a bug 100%.

Best wishes from AbanteCart Developers Team. :-)
https://github.com/abantecart

jdarwood007:
The problem is we connect to the specified smtp_host and then issue a ehlo with that stmp_host name.  So gmail appears to be blocking this.  We should be attempting it seems to get the FQDN (which SERVER_NAME should specify) and using that in our EHLO.

Of course incorrectly configured servers that do not setup the SERVER_NAME with their FQDN attribute may either cause this to say localhost or be blank.  Both of which can cause a bad EHLO as well.

live627:
>  Your code gives smtp-server's URI instead sender host name. it's a bug 100%.

It is user-configurable from the admin panel, no?

https://github.com/SimpleMachines/SMF2.1/blob/4a67433f28be2c0c1b63c87fd41f209116046a4f/Sources/ManageMail.php#L313

asmith20002:
@abolabo 
You gotta check SMF error log to see what gmail responded.

I use gmail smtp with SMF 2.1 beta 2 with absolutely no problems. But for the first time, gmail rejects it and you gotta configure some security settings in gmail to allow your website access.

In my case even after setting everything right, gmail was still rejecting me because of my password. Apparently when you get rejected by gmail, it also saves your password on those failed attempts. So even when I fixed all the security checks, it was seeing my password as suspicious. So after changing my password, it began to work.

abolabo:
[...]

PR #3576, opened 2016-09-11
Fixed issue with EHLO/HELO command sent to Gmail 

tinoest:
Fixes the issue raised in #3383

### DIFF
--- a/Sources/Subs-Post.php
+++ b/Sources/Subs-Post.php
@@ -270,7 +270,7 @@ function fixTags(&$message)
 		array(
 			'tag' => 'url',
 			'protocols' => array('http', 'https'),
-			'embeddedUrl' => true,
+			'embeddedUrl' => false,
 			'hasEqualSign' => false,
 		),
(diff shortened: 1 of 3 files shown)


Sources/Subs-Post.php
```diff
@@ -1316,12 +1316,22 @@ function smtp_mail($mail_to_array, $subject, $message, $headers)
 	if (!server_parse(null, $socket, '220'))
 		return false;
 
+	// Try and determine the servers name, fall back to the mail servers if not found
+	$helo		= FALSE;
```
MissAllSunday: Just a couple of cosmetic things, SMF doesn't use uppercase for boolean values and theres 2 tabs before the = sign, we usually use a single space.

tinoest: Changed them over, reverted to coding standards at work where its capitals for FALSE and TRUE and lined up = if possible.

MissAllSunday:
Thanks!

Unfortunately, it seems that not all of your commits are signed off.

PR #3586, opened 2016-09-18
Checking in changes to support STARTTLS

tinoest:
Added Support for STARTTLS command if using port 587 and the server supports it. 
Updated server_response params to reflect change. 

Signed-off-by: tinoest@gmail.com

### DIFF
--- a/Sources/ManageMail.php
+++ b/Sources/ManageMail.php
@@ -309,7 +309,7 @@ function ModifyMailSettings($return_config = false)
 			array('int', 'mail_quantity'),
 		'',
 			// SMTP stuff.
-			array('select', 'mail_type', array($txt['mail_type_default'], 'SMTP')),
+			array('select', 'mail_type', array($txt['mail_type_default'], 'SMTP', 'SMTP - STARTTLS')),
 			array('text', 'smtp_host'),
 			array('text', 'smtp_port'),
 			array('text', 'smtp_username'),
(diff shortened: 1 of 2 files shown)


jdarwood007:
Instead of auto assuming port 587 will use TLS.  This should go in as a modSetting requiring the admin to check use STARTTLS

tinoest:
STARTTLS is supported on port 587 (RFC 3207) and possibly port 25 (mentioned in RFC 2476)

I can add the option of it being a config but in its current form it'll only work if the server responds and says it can deal with it anyway.

jdarwood007:
Well you also get the problems where people are running on non-standard ports.  So assuming that 587 will always be used for TLS would cause problems there.  Standards are not always followed, some people even run them on 465 as well.

tinoest:
465 is an encrypted port already when you connect. So its not valid for STARTTLS, the only two I know off are port 25 and port 587, where you could use STARTTLS. 

I'll ammend as you suggest, although I don't think its required personally. At least if its not returned then it will get ignored in its current implementation even if the user enables it.

jdarwood007:
Well we try to be compliant with RFCs, but not everyone follows RFCs and some setups do use different ports for various reasons. I've seen STARTTLS used on port 465 and other random ports. So changing this prevents it from becoming an issue for those who want to use non standard ports with STARTTLS.

ColdrickSotK:
I came to check if anyone had created an issue or PR for this with a view to writing one myself if not. This one works nicely!

MissAllSunday:
Are the changes good @jdarwood007 ?  I don't know much about ports and supported mail standards so I leave it up to yo guys to decide.

jdarwood007:
Yes.   The changes do allow for non standard ports to be used.  So I see no issue here.

MissAllSunday:
cool, thanks!

PR #3587, opened 2016-09-19
Credits

MissAllSunday:
A few issues with the install and remove the hardcoded ip ion install

### DIFF
--- a/other/install_2-1_mysql.sql
+++ b/other/install_2-1_mysql.sql
@@ -1717,8 +1717,8 @@ VALUES ('xx', 'Standard', '0'),
 #
 
 INSERT INTO {$db_prefix}messages
-	(id_msg, id_msg_modified, id_topic, id_board, poster_time, subject, poster_name, poster_email, poster_ip, modified_name, body, icon)
-VALUES (1, 1, 1, 1, UNIX_TIMESTAMP(), '{$default_topic_subject}', 'Simple Machines', 'info@simplemachines.org', '', '', '{$default_topic_message}', 'xx');
+	(id_msg, id_msg_modified, id_topic, id_board, poster_time, subject, poster_name, poster_email, modified_name, body, icon)
+VALUES (1, 1, 1, 1, UNIX_TIMESTAMP(), '{$default_topic_subject}', 'Simple Machines', 'info@simplemachines.org', '', '{$default_topic_message}', 'xx');
 # --------------------------------------------------------
 
 #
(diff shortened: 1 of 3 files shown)


other/install_2-1_mysql.sql
```diff
@@ -1718,7 +1718,7 @@ VALUES ('xx', 'Standard', '0'),
 
 INSERT INTO {$db_prefix}messages
 	(id_msg, id_msg_modified, id_topic, id_board, poster_time, subject, poster_name, poster_email, poster_ip, modified_name, body, icon)
-VALUES (1, 1, 1, 1, UNIX_TIMESTAMP(), '{$default_topic_subject}', 'Simple Machines', 'info@simplemachines.org', '127.0.0.1', '', '{$default_topic_message}', 'xx');
+VALUES (1, 1, 1, 1, UNIX_TIMESTAMP(), '{$default_topic_subject}', 'Simple Machines', 'info@simplemachines.org', '', '', '{$default_topic_message}', 'xx');
```
albertlast: an emtpy string is no valid ip address,
please fill null in this field,
for this you got two ways:
first:
take the poster_ip field out from the insert statement plus the ''
or:
replace '' with null

recommended way is the first one because is use the default value of the field in the database.

other/install_2-1_postgresql.sql
```diff
@@ -2267,7 +2267,7 @@ VALUES ('xx', 'Standard', '0'),
 
 INSERT INTO {$db_prefix}messages
 	(id_msg, id_msg_modified, id_topic, id_board, poster_time, subject, poster_name, poster_email, poster_ip, modified_name, body, icon)
-VALUES (1, 1, 1, 1, {$current_time}, '{$default_topic_subject}', 'Simple Machines', 'info@simplemachines.org', '127.0.0.1', '', '{$default_topic_message}', 'xx');
+VALUES (1, 1, 1, 1, {$current_time}, '{$default_topic_subject}', 'Simple Machines', 'info@simplemachines.org', '', '', '{$default_topic_message}', 'xx');
```
albertlast: an emtpy string is no valid ip address,
please fill null in this field,
for this you got two ways:
first:
take the poster_ip field out from the insert statement plus the ''
or:
replace '' with null

recommended way is the first one because is use the default value of the field in the database.

jdarwood007:
Wait, blank ip isn't valid for poster ip?  Won't that cause issues elsewhere?  I know SMF OOB doesn't do anything that would cause a blank ip or should.  But there is always those 3rd party integrations and API usage that may have left it blank.

MissAllSunday:
It didn't cause any issues leaving that field empty when I was testing that change

albertlast:
To handle this easier,
i changed the behavior of the inet type #3588 
so when you do an normal insert you can use ''
this will be converted internaly to null and inet_dtop https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs.php#L5038
convert null value back to ''.

On the end of the day only two kind of value are allowed on database, 
a ipaddress(v4 or v6) or null.
Empty string is non of this.

PR #3601, opened 2016-10-03
Better logic for [img] BBC re: schemeless URLs and image proxy

Sesquipedalian:
In b9a5018, I missed the second form of [img] entirely, and the code I used in the first form would never use a schemeless link even if we actually did want one (i.e. if the image proxy was not enabled). Now [img] will behave consistently in both forms, and will use schemeless URLs when appropriate.


MissAllSunday:
Thanks!

PR #3609, opened 2016-10-06
Prevent autocomplete in password fields

sprak3000:
Best security practice is to have password fields set to not autocomplete. This PR updates all found password fields to have the autocomplete="off" attribute/value.

Signed-off-by: Luis Cruz <sprak3000+github@gmail.com>

### DIFF
--- a/SSI.php
+++ b/SSI.php
@@ -1198,7 +1198,7 @@ function ssi_login($redirect_to = '', $output_method = 'echo')
 					<td><input type="text" id="user" name="user" size="9" value="', $user_info['username'], '" class="input_text"></td>
 				</tr><tr>
 					<td style="text-align: right; border-spacing: 1"><label for="passwrd">', $txt['password'], ':</label>&nbsp;</td>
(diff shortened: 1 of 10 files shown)


Sesquipedalian:
Actually, allowing password managers to do their job is better for security. Password managers encourage users to use really good passwords, and the security benefit of that far, far outweighs any potential risks. That is why modern browsers now disregard the autocomplete attribute. Adding this would have no benefit. Mostly it would just annoy users if they used a browser that, for some reason, didn't just ignore the autocomplete attribute anyway.

sprak3000:
You are correct; this would have been better stated as a privacy best practice rather than a security practice. I still prefer to see password fields have this on regardless of browser defaults. There are still use cases for it, but I can also understand the desire to not have attribute bloat even on a small scale. YMMV

matthew-kerle:
Having autocomplete off does not prevent password managers to do their jobs...they target inputs that have type="password"...so having autocomplete off has no affect on those programs.

Sesquipedalian:
Indeed, which is part of why this change would have no benefit. It wouldn't work anyway, and even if it did, it isn't something that should work.

PR #3610, opened 2016-10-06
add inet support for table manage functions

albertlast:
Allow to create ip field by using the smf api and
reverse function specially for mysql 
varbinary + length 16 = inet
inet = varbinary and length 16


Sesquipedalian:
Making inet fields part of SMF's database abstraction layer is a great idea. :)

PR #3618, opened 2016-10-12
Fixes #3616

### LINKED ISSUE #3616: Web Links
Now some words are parsed as links

Sesquipedalian:
Once support for schemeless URLs was added, the removed bit of code was redundant and could produce false positives with certain strings.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -1883,12 +1883,7 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 		|										#  or
 		(?<=^|\W)(?=//)							# A boundary followed by two slashes (for schemeless URLs like "//example.com")
 	)						
-	(?:
-		/{1,3}									# 1-3 slashes
-		|										#	or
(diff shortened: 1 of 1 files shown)


Yariksat:
Thank you, there are still such a false alarm

Sesquipedalian:
In the case of `M.TV`, it will be linked by the autolinker because `M.TV` is a valid domain name. There is simply no way for a computer to know whether a human thinks that a string is supposed to be a domain name or not. All the computer knows is that the string does in fact match the format for a domain name.

So basically, we can either turn both `example.com` and `m.tv` into links, or neither of them. If your users start to complain about false positives on your forum, you can always disable the autolinker in your admin settings.

Yariksat:
Understood thanks .
Tell or say аnd where is this option  is disabled in the admin panel ?

Sesquipedalian:
It is in Features and Settings >> Bulletin Board Code.

Yariksat:
Exactly thank you.
Now everything is is just fine

PR #3636, opened 2016-10-17
PG added basic datatypes

albertlast:

### DIFF
--- a/Sources/DbPackages-postgresql.php
+++ b/Sources/DbPackages-postgresql.php
@@ -661,6 +661,9 @@ function smf_db_calculate_type($type_name, $type_size = null, $reverse = false)
 			'mediumtext' => 'text',
 			'largetext' => 'text',
 			'inet' => 'inet',
+			'time' => 'time without time zone',
+			'datetime' => 'timestamp without time zone',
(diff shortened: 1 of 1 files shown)


Sources/DbPackages-postgresql.php
```diff
@@ -661,6 +661,9 @@ function smf_db_calculate_type($type_name, $type_size = null, $reverse = false)
 			'mediumtext' => 'text',
 			'largetext' => 'text',
 			'inet' => 'inet',
+			'time' => 'time without time zone',
+			'datetime' => 'timestamp without time zone',
+			'timestamp' => 'timestamp without time zone',
```
jdarwood007, line 666: Is it a mistake that both timestamp and datetime have the same value?

albertlast: Sure in mysql are they nearly the same too : http://dev.mysql.com/doc/refman/5.7/en/datetime.html
only the min max values is different and pg only know one datatype: https://www.postgresql.org/docs/9.6/static/datatype-datetime.html

jdarwood007: Ok, just making sure.

colinschoen:
LGTM

PR #3638, opened 2016-10-18
"loadJavascriptFile" -> "loadJavaScriptFile"

matthew-kerle:
This PR changes the function name from "loadJavascriptFile" to "loadJavaScriptFile".

Signed-Off: Matthew P. Kerle matthew.kerle@gmail.com

### DIFF
--- a/Sources/Admin.php
+++ b/Sources/Admin.php
@@ -30,7 +30,7 @@ function AdminMain()
 	// Load the language and templates....
 	loadLanguage('Admin');
 	loadTemplate('Admin');
-	loadJavascriptFile('admin.js', array(), 'smf_admin');
+	loadJavaScriptFile('admin.js', array(), 'smf_admin');
 	loadCSSFile('admin.css', array(), 'smf_admin');
 
 	// No indexing evil stuff.
@@ -572,7 +572,7 @@ function AdminHome()
 	);
 
 	if ($context['admin_area'] == 'admin')
-		loadJavascriptFile('admin.js', array('defer' => false), 'smf_admin');
+		loadJavaScriptFile('admin.js', array('defer' => false), 'smf_admin');
 }
 
 /**
--- a/Sources/BoardIndex.php
+++ b/Sources/BoardIndex.php
@@ -138,7 +138,7 @@ function BoardIndex()
 
 	if (!empty($settings['show_newsfader']))
 	{
-		loadJavascriptFile('slippry.min.js', array(), 'smf_jquery_slippry');
+		loadJavaScriptFile('slippry.min.js', array(), 'smf_jquery_slippry');
 		loadCSSFile('slider.min.css', array(), 'smf_jquery_slider');
 	}
 }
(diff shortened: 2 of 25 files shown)


Sesquipedalian:
I think this would break a lot of mods for very little gain.

matthew-kerle:
If by very little gain you mean spelling something correctly but no functional gain, you're correct. But it'd be good to have this spelled right, and now is the time to do it. If we wait even until a release candidate it'll never happen because then there would be too many mods. SMF 2.1 is in beta right now, meaning things could change, and that's a risk any mod developers take on -- most mod authors this change won't be a big deal at all.

I'm a developer for SimpleDesk, so I can change all of our usages with that modification very easily. Beyond that, any other modifications that this would break where the developer doesn't want to update for, I can provide the package to them with the updated version.

Sesquipedalian:
> Beyond that, any other modifications that this would break where the developer doesn't want to update for, I can provide the package to them with the updated version.

So, you are volunteering to go through the entire SMF mods database and make sure all of them are updated to offer both an SMF 2.0 version and a SMF 2.1 version? That sounds like a lot of work...

I just don't think this change is worthwhile, for the same reason that no one has ever corrected "REFERER" to "REFERRER" in the HTTP header spec. This is a change that would break a lot of software that is not in our control and precipitate a huge amount of work to fix, for what is really just a cosmetic difference.

matthew-kerle:
-snip- Cursing, being extremely rude and shouting is not tolerated here.

XinYenFon:
wow... that was not cool... and expect something soon.

matthew-kerle:
-edit- snip

Sesquipedalian:
> -snip-

What the...? I think you read a different tone into my words than I meant, @Labradoodle-360. I'm just interacting with you in a bit of friendly debate, not trying to be rude or denigrate you.

Also, I'm not on the SMF team. Like you, I'm just a contributor, so don't blame them if you have an issue with me. I'm just some random guy on the internet.

Anyway, if `loadJavascriptFile()` was introduced in SMF 2.1, then that changes the situation completely. I was under the (apparently mistaken) impression that it was in SMF 2.0. If it was introduced in 2.1, then yes, by all means let's fix the spelling issue.

Sesquipedalian:
Hm. I see you closed this PR while I was typing my last message. Perhaps you would consider reopening it? If the function was introduced in 2.1, then I am in agreement with your code changes.

Sesquipedalian:
In fact, if you resubmit, @Labradoodle-360, perhaps you could also take care of `addJavascriptVar()` and `addInlineJavascript()`?

emanuele45:
It's sad to see insults are the only defensive argument, when it was enough to link the PHP documentation http://php.net/manual/en/functions.user-defined.php

Sesquipedalian:
> Note: Function names are case-insensitive, though it is usually good form to call functions as they appear in their declaration.
> -- [PHP Manual](http://php.net/manual/en/functions.user-defined.php)

Oh, hey. I didn't realize that. Thanks, @emanuele45! Learn something new every day. :)

PR #3639, opened 2016-10-18
Fix Inline Edit Styling

matthew-kerle:
Fixes the inline edit styling for messages.

Signed-Off: Matthew P. Kerle matthew.kerle@gmail.com


XinYenFon:
You forgot to sign-off commit(s).

PR #3641, opened 2016-10-20
Converts sort_bbc_tags callback for usort to an anonymous function

Sesquipedalian:


albertlast:
Why should be this better as this befor?
and how it works with php 7.1?

Sesquipedalian:
Functionally, it isn't really any different at all. But there has been a push in SMF 2.1 to convert callback functions that are only used once (like this one) into closures (i.e. inline anonymous functions). It just makes for neater, more self-contained code, that's all. :)

This change should work just fine in PHP 7. PHP 7 added some new abilities regarding closures, but it doesn't break anything.

colinschoen:
LGTM

PR #3655, opened 2016-10-25
CSS fixes

frandominguezl:
Adjusting here some CSS aspects;
- The overflow property set to .roundframe caused to display the login box like this: http://bit.ly/2dFRNvT
- Adjusted top margin of .overall_progress on install to centralize it just a bit


XinYenFon:
Those changes are not necessary.

frandominguezl:
Sorry, didn't notice this before

PR #3662, opened 2016-11-04
Add if exists upgrade for pg upgrade

albertlast:
since we now use 9.1 as min version we can no use if not exists syntax
in our create table statements -> less posible to get problems

this branche based on qanda fix pr

### DIFF
--- a/other/upgrade_2-1_postgresql.sql
+++ b/other/upgrade_2-1_postgresql.sql
@@ -9,7 +9,7 @@ CREATE SEQUENCE {$db_prefix}member_logins_seq;
 ---#
 
 ---# Creating login history table.
-CREATE TABLE {$db_prefix}member_logins (
+CREATE TABLE IF NOT EXISTS {$db_prefix}member_logins (
 	id_login int NOT NULL default nextval('{$db_prefix}member_logins_seq'),
 	id_member int NOT NULL,
 	time int NOT NULL,
@@ -548,7 +548,7 @@ CREATE SEQUENCE {$db_prefix}background_tasks_seq;
 ---#
 
 ---# Adding the table
-CREATE TABLE {$db_prefix}background_tasks (
+CREATE TABLE IF NOT EXISTS {$db_prefix}background_tasks (
 	id_task int default nextval('{$db_prefix}background_tasks_seq'),
 	task_file varchar(255) NOT NULL default '',
 	task_class varchar(255) NOT NULL default '',
@@ -660,7 +660,7 @@ ADD COLUMN alerts int NOT NULL default '0';
 ---# Adding the new table for alerts.
 CREATE SEQUENCE {$db_prefix}user_alerts_seq;
 
-CREATE TABLE {$db_prefix}user_alerts (
+CREATE TABLE IF NOT EXISTS {$db_prefix}user_alerts (
 	id_alert int default nextval('{$db_prefix}user_alerts_seq'),
 	alert_time int NOT NULL default '0',
 	id_member int NOT NULL default '0',
@@ -679,7 +679,7 @@ CREATE INDEX {$db_prefix}user_alerts_alert_time ON {$db_prefix}user_alerts (aler
 ---#
 
 ---# Adding alert preferences.
-CREATE TABLE {$db_prefix}user_alerts_prefs (
+CREATE TABLE IF NOT EXISTS {$db_prefix}user_alerts_prefs (
 	id_member int NOT NULL default '0',
 	alert_pref varchar(32) NOT NULL default '',
 	alert_value smallint NOT NULL default '0',
@@ -1089,7 +1089,7 @@ ALTER TABLE {$db_prefix}members
 ---# Creating drafts table.
 CREATE SEQUENCE {$db_prefix}user_drafts_seq;
 
-CREATE TABLE {$db_prefix}user_drafts (
+CREATE TABLE IF NOT EXISTS {$db_prefix}user_drafts (
 	id_draft int NOT NULL default nextval('{$db_prefix}user_drafts_seq'),
 	id_topic int NOT NULL default '0',
 	id_board smallint NOT NULL default '0',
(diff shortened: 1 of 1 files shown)


other/upgrade_2-1_postgresql.sql
```diff
@@ -2055,15 +2055,19 @@ CREATE INDEX {$db_prefix}ban_items_id_ban_ip ON {$db_prefix}ban_items (ip_low,ip
...
+	AS
+	$$
+	BEGIN
+	   RETURN (trim(val))::inet;
+	EXCEPTION
+	   WHEN OTHERS THEN RETURN NULL;
+	END;
+	$$ LANGUAGE plpgsql;"
+);
+---}
```
rturja, line 2070: still getting:

```
Error!
ERROR: unterminated dollar-quoted string at or near "$$ LANGUAGE plpgsql" LINE 1: $$ LANGUAGE plpgsql ^
```

and

```
Nov  4 21:20:47 cerebro postgres[79657]: [3-1] ERROR:  unterminated dollar-quoted string at or near "$$
Nov  4 21:20:47 cerebro postgres[79657]: [3-2]  BEGIN
Nov  4 21:20:47 cerebro postgres[79657]: [3-3]     RETURN (trim(val))::inet" at character 76
Nov  4 21:20:47 cerebro postgres[79657]: [3-4] STATEMENT:  CREATE OR REPLACE FUNCTION migrate_inet(val IN anyelement) RETURNS inet
Nov  4 21:20:47 cerebro postgres[79657]: [3-5]  AS
Nov  4 21:20:47 cerebro postgres[79657]: [3-6]  $$
Nov  4 21:20:47 cerebro postgres[79657]: [3-7]  BEGIN
Nov  4 21:20:47 cerebro postgres[79657]: [3-8]     RETURN (trim(val))::inet
Nov  4 21:20:47 cerebro postgres[79657]: [4-1] ERROR:  syntax error at or near "EXCEPTION" at character 1
Nov  4 21:20:47 cerebro postgres[79657]: [4-2] STATEMENT:  EXCEPTION
Nov  4 21:20:47 cerebro postgres[79657]: [4-3]     WHEN OTHERS THEN RETURN NULL
Nov  4 21:20:47 cerebro postgres[79657]: [5-1] ERROR:  unterminated dollar-quoted string at or near "$$ LANGUAGE plpgsql" at character 1
Nov  4 21:20:47 cerebro postgres[79657]: [5-2] STATEMENT:  $$ LANGUAGE plpgsql
```

Interestingly enough if I copy and paste the function creation directly to pgadmin, it works without fault

albertlast: Sure it work in pgadmin directly ther i created this function and hop this run in smf too :(

albertlast: please try something for me got this line: https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs-Db-postgresql.php#L403
if (empty($modSettings['disableQueryCheck']))
-->
if (empty($modSettings['disableQueryCheck']) && false)

rturja: Albert you're a hero! :)

Got my board up and running. The problem with the plpgsql function actually was that I pasted the new one on top of the old and then forget to remove the old... So pilot error. Board is up and running, update all done.

I'll do a fresh update from existing data and then let our forum users loose on the beta install :)

Thanks a lot! -Reko

albertlast: so is this line changes needed or works this with the pr changes?

albertlast: So i checked to  day the code and i can't understand fully why this happen.
Because the query check should be already disable by the upgrade_query function and
when this plpgsql function already exists should this replaced by this function (CREATE OR REPLACE FUNCTION migrate_inet) because of the replace in the statement.

rturja: This worked with the change to Subs-DB-postgresql.php

I got the upgrade working from the scratch- almost. Below are some notes:

Installation got stuck at:

```
Nov  5 13:31:55 cerebro postgres[12375]: [3-1] ERROR:  index "verstas_log_group_requests_id_member" does not exist
Nov  5 13:31:55 cerebro postgres[12375]: [3-2] STATEMENT:  DROP INDEX verstas_log_group_requests_id_member
Nov  5 13:33:42 cerebro postgres[12648]: [3-1] ERROR:  column "disregarded" does not exist
Nov  5 13:33:42 cerebro postgres[12648]: [3-2] STATEMENT:  
Nov  5 13:33:42 cerebro postgres[12648]: [3-3]          ALTER TABLE verstas_log_topics
Nov  5 13:33:42 cerebro postgres[12648]: [3-4]          RENAME disregarded TO unwatched

```

And continued only after page reload

After the database changes and hitting continue blank screen with:

`ERROR: relation "verstas_background_tasks" does not exist LINE 3: FROM verstas_background_tasks ^`

Is there a way to do the JSON step without running the whole upgrade script btw?

Then there's a slight typo in Sources/ManageSearch.php line 247
it is `CREATE INDEX smf_messages_ftx ON smf_messages`
when it should be                                  `CREATE INDEX smf_messages_ftx ON {db_prefix}messages` 

-Reko

albertlast: Need to check why this change in subs-db-pg is needed...

The first issue with the index is fixed

background_tasks i hope is fixed by his self...

json duno

last is fixed by a new pr https://github.com/SimpleMachines/SMF2.1/pull/3663

albertlast:
So checked the error about the missing ; ( the mention change in Subs-DB-postgresql.php)
but this issue is not there in my env --> so i can not fix this when this issue i keep on your side.

other/upgrade_2-1_postgresql.sql
rturja: With the latest changes update went through without a hitch, thanks a million again for your work, Now onwards to configure the test board and let the hordes loose on it...

MissAllSunday:
Hi @albertlast Are all these recent PRs compatible with Mysql?

Oldiesmann:
@MissAllSunday This one won't affect MySQL at all as it only changes the upgrade_2-1_postgresql.sql file. I'm not sure about the others.

albertlast:
Sure they are compatible,
some fixes could effect mysql too.
But than in a good manner like this one: https://github.com/SimpleMachines/SMF2.1/pull/3678/files

MissAllSunday:
OK, let me know when this one and https://github.com/SimpleMachines/SMF2.1/pull/3679 are ready to be merged.

albertlast:
This here is done the other need some love

Issue #3671, opened 2016-11-07
Missing Charset / Collate information for pg install / upgrade


albertlast:
Since we use pg 9.1+, now.
We can define for everytable/col an own collate.
So we need to add to the create table statement in install/upgrade process the collate information.
Otherwise the user need to choose the right one by creation of the database.

live627:
#3679

PR #3679, opened 2016-11-09
Fix missing global char set fix #3669

### LINKED ISSUE #3669: Unicode strings are not cut cleanly in main view "Last post" section
See the picture; o with umlauts "ö" gives unexpected result. I've fixed this in 2.0.x, but can't remember where I did fix it...

unicode

albertlast:
In the pg upgrade process this information was missing.
I think it's doesn't hurt to set this value everytime.

related issue #3669

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -3299,6 +3299,13 @@ function convertUtf8()
 	// First make sure they aren't already on UTF-8 before we go anywhere...
 	if ($db_type == 'postgresql' || ($db_character_set === 'utf8' && !empty($modSettings['global_character_set']) && $modSettings['global_character_set'] === 'UTF-8'))
 	{
+		$smcFunc['db_insert']('replace',
+			'{db_prefix}settings',
+			array('variable' => 'string', 'value' => 'string'),
+			array(array('global_character_set', 'UTF-8')),
+			array('variable')
+		);
+		
 		return true;
 	}
 	else


rturja:
I think better - and safer approach - is checking the contents of column character_set_name from information_schema.character_sets first. If that is set to UTF8 (note the missing dash!) then it's safe to set the value. If it's something else, then it's probably better to leave gloabal_character_set as it is.

Changing the encoding in postgres - if there's any special characters isn't just setting a value, but involves dump & restore, so I think it is better to take safe approach here and check the encoding before setting the value.

Having this backported into 2.0.x installer and postgres updater would be lovely too though.

albertlast:
why without - ?
The mysql installation path take UTF-8...

rturja:
Sorry for not being clearer. Postgres stores the utf character set as UTF8, not UTF-8 in information_schema.character_sets.character_set_name

rturja:
So of course the SMF character set variable should be set as UTF-8 as it is done now, but the check for postgres database charset should test for UTF8.

albertlast:
okay than i to do more things and befor i had to fix this point: https://github.com/SimpleMachines/SMF2.1/issues/3671

albertlast:
@rturja can your please look at db_character_set in your Settings.php and
say what value is setting there?

rturja:
finnish-utf8 (in both 2.0.12 and the 2.1)

albertlast:
well this make less sense...
this value from the setting will be parse to the database in this form 
SET NAMES 'finnisch-utf8';
When you run this command directly in your database you should get an error message.
A correct value would be utf8 so i would like to try something.
Please correct this setting value and look if something change and
than delete the global_character_set in your database and look if something change again.

rturja:
No wonder that made little sense. In hurry I checked $language. $db_character_set is utf8 in both Settings.php files (2.0.x and 2.1).

If I remove global_character_set from database, the problem with handling unicode returns. Also the problem is there if value for global_character_set is empty.

albertlast:
did you try after remove of the global_character_set,
to create a new poste and look if he get the same problem?

rturja:
If I remove global_character_set the problem with cutting UTF strings returns with both old and new posts.

albertlast:
And what is the behavior of smf when you put UTF-8 as global_character_set?

albertlast:
@Oldiesmann in my eyes this pr is ready
because the value is need to be in the database
and actual this value is missing in pg installation

PR #3680, opened 2016-11-11
Keeps the autolinker up-to-date with the full list of top level domains

Sesquipedalian:
The `set_tld_regex()` function creates an optimized regex to match all known top level domains and stores it in $modSettings['tld_regex']. The regex is updated once per week (or possibly more often if necessary for some reason) using a cron job. If the regex is discovered to be missing or invalid during day-to-day use, it will be regenerated from a hard-coded list of basic TLDs for the short term, and a background update via cron will be scheduled to fix it properly in the near future.

This PR also tweaks `build_regex()` for cleaner handling of multibyte character encodings like UTF-8.

Finally, this PR also improves the regexes used for autolinking URLs and email address.

### DIFF
--- a/Sources/ScheduledTasks.php
+++ b/Sources/ScheduledTasks.php
@@ -1297,7 +1297,7 @@ function scheduled_birthdayemails()
  */
 function scheduled_weekly_maintenance()
 {
-	global $modSettings, $smcFunc;
+	global $modSettings, $smcFunc, $sourcedir;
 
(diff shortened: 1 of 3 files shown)


Sesquipedalian:
Unless someone can find some bugs that I can't, this is ready to go.

Oldiesmann:
Is this ready to go now?

Sesquipedalian:
Yes.

Issue #3693, opened 2016-11-28
memcached procedural functions


MissAllSunday:
My server doesn't recognize memcached procedural functions, it returns false for var_dump(function_exists('memcached_get'));  

But it does recognize Memcached class:

```php
$s = new Memcached;

$s->addServer('server', port);
$s->add ('key', 'value', 3600);
var_dump($s->getAllKeys ());
```

prints out:

array(2) { [0]=> string(3) "key" [1]=> string(3) "value" }

Memcached values:

Version	3.0.0b1
libmemcached version	1.0.18

PHP Version	7.0.13

By reading https://github.com/php-memcached-dev/php-memcached/issues/225  it seems theres a few issues when using php7 and memcached 3.0.0 is still a beta so this might not be a huge issue.

But it would be nice to add support for Memcached class if it exists and the check for procedural fails in case the procedural functions get deprecated/removed later on.

jdarwood007:
Should just move it all to the class method instead of supporting both.

MissAllSunday:
With our current min versions, is it safe to do so? I don't know much about memcached and its development.

I would also like to have a real check to see if theres any real memcached server running, the current code only checks for the php side of it, some hosts, specially shared host might have the extension enabled but theres no server or access to that memcached server.

Oldiesmann:
I've submitted a PR to change all the functions to OOP instead of procedural. No verification of the servers is needed as SMF will try to connect before trying to do anything and bails if the $memcached variable is false (see the get_memcached_server() function in Sources/Load.php).

jdarwood007:
Cool, I will test it when I can.

PR #3697, opened 2016-12-01
Switch to OOP for memcache

Oldiesmann:
This commit changes all the procedural functions for memcache(d) to the OOP equivalents, per #3693. Note that I haven't tested any of this yet as I don't have memcache on my server.

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -3238,6 +3238,7 @@ function cache_quick_get($key, $file, $function, $params, $level = 1)
  * - It supports:
  *	 Xcache: http://xcache.lighttpd.net/wiki/XcacheApi
  *	 memcache: http://www.php.net/memcache
+ *   memcached: http://www.php.net/memcached
  *	 APC: http://www.php.net/apc
  *   APCu: http://www.php.net/book.apcu
  *	 Zend: http://files.zend.com/help/Zend-Platform/output_cache_functions.htm
@@ -3270,7 +3271,7 @@ function cache_put_data($key, $value, $ttl = 120)
 	{
 		case 'memcached':
 			// The simple yet efficient memcached.
-			if ((function_exists('memcached_set') || function_exists('memcache_set')) && isset($cache_memcached) && trim($cache_memcached) != '')
+			if ((class_exists('memcached') || class_exists('memcache')) && isset($cache_memcached) && trim($cache_memcached) != '')
 			{
 				// Not connected yet?
 				if (empty($memcached))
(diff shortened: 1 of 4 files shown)


albertlast:
From my pov you should wrap the two classes into one class

MissAllSunday:
When testing this I got an error:

Uncaught Error: Call to undefined method Memcached::connect()  the server where I'm testing this only has memcached support so it passes the firsts tests and creates a new instance of memcached but it fails because the code thinks the instance is a memcache one and not memcached.

jdarwood007:
That is because memcacheD does not have a connect method.  It appears for it, you create the object and then run addServers.  According to some notes, when using FastCGI that doesn't shutdown this can be unsafe as it doesn't check for duplicate serveres
https://secure.php.net/manual/en/memcached.addserver.php

I was thinking if we don't mind, doing a quick rewrite here for the caching.  I would like to rewrite it up similar to how our SearchAPI works, but a CacheAPI instead.

So we would have the following files
- Class-CacheAPI.php
- CacheAPI-memcache.php
- CacheAPI-memcached.php
- CacheAPI-apc.php
- CacheAPI-apcu.php
- CacheAPI-zend.php
- CacheAPI-xcache.php
- CacheAPI-SMF.php

The class would define a few methods
- isSupported (Checks if we can use this cache method)
- connect (Sets up anything we need to before hand, this would be handled in reloadSettings before we do anything really, would also do a key setup).
- setKey (Allows for changing the key)
- getKey (Gets the key, useful if we have different keys we are running).
- getData (Saves data to cache)
- putData (Gets data from cache)
- settings (This would allow us when choosing the cache method, to define additional modSettings if needed)
- quit (Shuts downs the cache, we could possibly also use the __destruct to point to this if the object is destroyed).

We would add to Load.php a loadCache($overrideCache = null) which would check our setup and figure out what caching method we need to use and get it running, saving the object instance of the class in a global such as $cacheAPI or similar.  It would also setup the key.  It takes 1 parameter as a override to override the $cache_method we defined in Settings.php.  loadCache would return the $cacheAPI instance, if the override is set, we wouldn't change the $cacheAPI global.
The cache_(get|put)_data would be changed to basically get the object again from $cacheAPI and running the related get/pull.  These exist to keep compatibility.

This then makes it easier for mod authors to add additional caching methods that are retained during upgrades or replacing the file completely.  It also allows people using SSI.php to call the CacheAPI to do their own/separate caching method.  They would have to use the $cacheAPI object returned by loadCache rather than the cache_(get|put)_data functions in SMF.

MissAllSunday:
> That is because memcacheD does not have a connect method.

Yes, is a logic problem, either treat memcache and memcached as completely separate or built a wrapper like you suggest.

Oldiesmann:
Closing this in favor of #3703

PR #3699, opened 2016-12-03
Fixes #3624

### LINKED ISSUE #3624: Inserting smileys in a message
When you insert attached image to the body of the message is not displayed after the smilies

Sesquipedalian:
Problem was the newlines inserted by the attach BBC validation function. Since parse_bbc() uses newlines as delimiters to decide where to parse smilies, these extra newlines were causing confusion.


MissAllSunday:
awesome, thanks!

Issue #3644, opened 2016-10-22
 Xcache


Yariksat:
The forum included  Xcache
When adding a new section in the admin panel
PHP Fatal error:  xcache_clear_cache(): xcache.admin.user and/or xcache.admin.pass settings is not configured. Make sure you've modified the correct php ini file for your php used in webserver. in /home/p2777/public_html/Sources/Subs.php on line 4203
        case 'xcache':
            if (function_exists('xcache_clear_cache'))
            {
                //
                if ($type === '')
                {
                    xcache_clear_cache(XC_TYPE_VAR, 0);
                    xcache_clear_cache(XC_TYPE_PHP, 0);
                }
                if ($type === 'user')
                    xcache_clear_cache(XC_TYPE_VAR, 0);
                if ($type === 'data')
    ==>             xcache_clear_cache(XC_TYPE_PHP, 0);
            }
            break;

PR #3703, opened 2016-12-05
Move the caching support over to an API

### LINKED ISSUE #3613: Replace apc* cache with acpu*
Just my note or anyone is able to do this.  APC was removed in PHP 5.6 and is not maintained or supported anymore.  However APCu is maintained and supported even in PHP 7.1.  APCu is APC without the opcache.  All functions I can find seem to be directly the same.  It should be safe to do a PR and replace all APC cache functions with APCu and update text strings as needed.

Should also migrate the modSetting from apc to apcu if its set.

### LINKED ISSUE #3693: memcached procedural functions

jdarwood007:
This implants a Class-CacheAPI that defines our supported setup.  Our supported cache methods have been converted over.

This conflicts with #3697
This resolves #3693 
This resolves #3613

Please test this well.  I've tested with SMF, APCu and Memcache.


MissAllSunday:
My server isn't using the memcached class, it correctly detects memcached but it continues to use smf file based cache

I believe its because of this:

```php
	$tryAccelerator = !empty($overrideCache) ? $overrideCache : !empty($cache_accelerator) ? $cache_accelerator : 'smf';
```


!empty($cache_accelerator) will always return false since that var ($cache_accelerator) isn't defined prior to that check so $tryAccelerator will always be assigned to "smf";

MissAllSunday:
$cache_accelerator wasn't globalized.

After added to the global scope I now get this error:

PHP Fatal error:  Uncaught Error: Call to a member function get() on null in /Sources/CacheAPI-memcached.php:103

jdarwood007:
Oh my, your right.  My testing yesterday was falling back to SMF file caching.  Try the most recent changes, I forgot to call the connect.

albertlast:
The whole class thing i personaly only know from java topics and
one thing which looks to me "wrong" that you got set and get function for like defaultttls (this is good) and
on the other side public vars ttl.

Is that a good idea to keep them public instead of make them private or protected? (maybe that doesn't work in php i dont know...)

jdarwood007:
Well if we didn't want them public, we would have to make them protected, since the base class gets extended and contains those members.

Does it offer us any advantage here?  There isn't any sensitive stuff being passed around really.  The only stuff I could think off is the memcache instances and its private from the class it extends.

MissAllSunday:
Got a new error:

Fatal error:  Uncaught Error: Call to a member function getServerList() on null in /Sources/CacheAPI-memcached.php:60

It seems $this->memcached is never created as an instance of Memcached class

jdarwood007:
Are you sure its memcached and not memcache?

https://secure.php.net/manual/en/memcached.getserverlist.php

That should exist

MissAllSunday:
Thats not the issue, the issue is that  $this->memcached is always null,  it doesn't hold an instance of Memcached.  There should be a new instance somewhere in the memcached class:

$this->memcached = new Memcached;

jdarwood007:
I used $this->memcache for memcached as well. I can change that, its just a naming convention at that point

jdarwood007:
Oh wait, I see now.  Getting confused with memcache and memcached

MissAllSunday:
nope....

public function connect()  uses the property $this->memcached  which is defined as null.  It never holds an instance of Memcached  so it remains as null.

When connect() gets called and $currentServers is defined, $this->memcached is still null.

Before $this->memcached is used, it must contain an instance of Memcached:

$this->memcached = new Memcached;
 $currentServers = $this->memcached->getServerList();

jdarwood007:
Yep.  Fixed it hopefully?

MissAllSunday:
yep but now I got a new one:

Memcached::set() expects at most 3 parameters, 4 given	/Sources/CacheAPI-memcached.php
Line: 125

the set() method in memcached is slightly different than that one memcache uses.

jdarwood007:
Oh so it is, it doesn't offer compression methods in the set.

MissAllSunday:
All working now, will leave it for a couple of days to see if theres any more issues

MissAllSunday:
oops, nvm, a nre error:

Json decode error: Syntax error, malformed JSON	h
/Sources/Load.php
Line: 3379

return empty($value) ? null : smf_json_decode($value, true);

The json data is the one coming from memcached, will take a look at it.

jdarwood007:
Is $value returning as something else?

I detect if $value was exactly false (=== test) and return null, otherwise return the value.  This is supposed to catch anything that may be considered empty such as 0 but still a stored value.

Do you happen to know what is being cached?  You can dump the $key to find out.

MissAllSunday:
yeah thats what I've been trying to do, it seems the error is generated every time this url gets called:

https://missallsunday.com/index.php?favicon=ico

I don't know if that param (favicon) is included on a vanilla SMF installation or its a part of some of the mods I have installed. It fails and gets redirected back to my index.

 Can't find any code that handles that particular url so I can't detect what exactly its been retrieved from the cache.

jdarwood007:
That looks like a mod.  I don't recall any such thing, nor does it work on my test site. https://dev.sleepycode.com/smf21git/index.php?favicon=ico

MissAllSunday:
I think my browser its generating that url and simpleSEF is redirecting it. Thats the only mod I have that could be causing it.

MissAllSunday:
Just disabled simpleSEF and the error is gone so yeah, its related to it, will have a look and fix it within the mod: https://github.com/MissAllSunday/SimpleSEF/issues/6

albertlast:
> Does it offer us any advantage here? There isn't any sensitive stuff being passed around really. The only stuff I could think off is the memcache instances and its private from the class it extends.

sure is there advantage, when you got only set and get function public, your are able to place check logic if the value is correct for this var for exp $keyprefix var is a Ä or ü or ` valid chars?
other advantage is your can place update logic "when the ttl is changed than i had to update the whole cache" with public vars you got no chance todo so.

one additional note i would like to see that prefix function and the prefix key got the same name
$keyPrefix vs setPrefix

jdarwood007:
Well I suspect that UTF-8 would be supported in keys for all the caching APIs, since they support storing UTF-8 data.

I fixed up those changes.  I prefer set/get/put/etc at the start of most methods because it defines a clear accepted action (updating/changing data or retrieving)

Sources/CacheAPI-apc.php
```diff
@@ -0,0 +1,98 @@
...
+		return parent::isSupported() && $supported;
+	}
+
+	/**
+	 * Connects to the cache method. If this fails, we return false, otherwise we return true.
+	 *
+	 * @access public
+	 * @return boolean Whether or not the cache method was connected to.
+	 */
+	public function connect()
```
live627: these functions that don't do anything should be moved to the base class, less duplication.

Sources/CacheAPI-apc.php
```diff
@@ -0,0 +1,98 @@
...
+			return apc_store($key . 'smf', $value, $ttl);
+	}
+
+	/**
+	 * Closes connections to the cache method.
+	 *
+	 * @access public
+	 * @return bool Whether or not we could close connections.
+	 */
+	public function quit()
```
live627: these functions that don't do anything should be moved to the base class, less duplication.

Sources/CacheAPI-apcu.php
```diff
@@ -0,0 +1,98 @@
...
+		return parent::isSupported() && $supported;
+	}
+
+	/**
+	 * Connects to the cache method. If this fails, we return false, otherwise we return true.
+	 *
+	 * @access public
+	 * @return boolean Whether or not the cache method was connected to.
+	 */
+	public function connect()
```
live627: these functions that don't do anything should be moved to the base class, less duplication.

Sources/CacheAPI-apcu.php
```diff
@@ -0,0 +1,98 @@
...
+			return apcu_store($key . 'smf', $value, $ttl);
+	}
+
+	/**
+	 * Closes connections to the cache method.
+	 *
+	 * @access public
+	 * @return bool Whether or not we could close connections.
+	 */
+	public function quit()
```
live627: these functions that don't do anything should be moved to the base class, less duplication.

Sources/CacheAPI-smf.php
```diff
@@ -0,0 +1,169 @@
...
+		return parent::isSupported() && $supported;
+	}
+
+	/**
+	 * Connects to the cache method. If this fails, we return false, otherwise we return true.
+	 *
+	 * @access public
+	 * @return boolean Whether or not the cache method was connected to.
+	 */
+	public function connect()
```
live627: these functions that don't do anything should be moved to the base class, less duplication

Sources/CacheAPI-smf.php
```diff
@@ -0,0 +1,169 @@
...
+	{
+//		$this->setPrefix('');
```
live627: Debugging leftover...

Sources/CacheAPI-smf.php
```diff
@@ -0,0 +1,169 @@
...
+		}
+	}
+
+	/**
+	 * Closes connections to the cache method.
+	 *
+	 * @access public
+	 * @return bool Whether or not we could close connections.
+	 */
+	public function quit()
```
live627: these functions that don't do anything should be moved to the base class, less duplication.

Sources/CacheAPI-xcache.php
```diff
@@ -0,0 +1,97 @@
...
+		return parent::isSupported() && $supported;
+	}
+
+	/**
+	 * Connects to the cache method. If this fails, we return false, otherwise we return true.
+	 *
+	 * @access public
+	 * @return boolean Whether or not the cache method was connected to.
+	 */
+	public function connect()
```
live627: Needs authentication here to be able to clear all, related to #3644

Sources/CacheAPI-xcache.php
```diff
@@ -0,0 +1,97 @@
...
+			return xcache_set($key, $value, $ttl);
+	}
+
+	/**
+	 * Closes connections to the cache method.
+	 *
+	 * @access public
+	 * @return bool Whether or not we could close connections.
+	 */
+	public function quit()
```
live627: these functions that don't do anything should be moved to the base class, less duplication.

Sources/CacheAPI-zend.php
```diff
@@ -0,0 +1,101 @@
...
+		return parent::isSupported() && $supported;
+	}
+
+	/**
+	 * Connects to the cache method. If this fails, we return false, otherwise we return true.
+	 *
+	 * @access public
+	 * @return boolean Whether or not the cache method was connected to.
+	 */
+	public function connect()
```
live627: these functions that don't do anything should be moved to the base class, less duplication.

Sources/CacheAPI-zend.php
```diff
@@ -0,0 +1,101 @@
...
+			return output_cache_put($key, $value);
+	}
+
+	/**
+	 * Closes connections to the cache method.
+	 *
+	 * @access public
+	 * @return bool Whether or not we could close connections.
+	 */
+	public function quit()
```
live627: these functions that don't do anything should be moved to the base class, less duplication.

Sources/Class-CacheAPI.php
```diff
@@ -0,0 +1,271 @@
...
+	 */
+	public function getDefaultTTL()
+	{
+		return $this->ttl;
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public function getData($key, $ttl = null)
```
live627, line 246: These empty stubs should be removed since they all should already have been implemented in the extending objects.

Sources/CacheAPI-smf.php
```diff
@@ -0,0 +1,169 @@
...
+			// Work around Zend's opcode caching (PHP 5.5+), they would cache older files for a couple of seconds
+			// causing newer files to take effect a while later.
+			if (function_exists('opcache_invalidate'))
+				opcache_invalidate($cachedir . '/data_' . $key . '.php', true);
+
+			if (function_exists('apc_delete_file'))
+				@apc_delete_file($cachedir . '/data_' . $key . '.php');
+
+			// php will cache file_exists et all, we can't 100% depend on its results so proceed with caution
+			@include($cachedir . '/data_' . $key . '.php');
```
live627, line 71: Could we move to JSON here? IMO it would be better than abusing PHP files for this...

live627 (changes requested):
- the @inheritdoc blocks can and should be on all of the extending classes
- you forgot to include a clear() met5hod... https://github.com/SimpleMachines/SMF2.1/blob/aafc00003fae1129ed2e0567e3e5402b8e484787/Sources/Subs.php#L3950
- i hate jqwarry

Sources/Class-CacheAPI.php
jdarwood007: Fixing this causing abstract errors on the individual APIs themselves.  Marking the individual APIs abstract causes more errors.  Not sure if we want to do it this way.

Sources/CacheAPI-xcache.php
jdarwood007: PHP 7 doesn't appear to have xCache so I can't test this, but implanting the code that may possibly work.

jdarwood007:
I left the JSON issue out of this for now.  We should discuss if we want to move the SMF file caching over to JSON separately.  File Based caching is fairly stable compared to when it was first released.  I don't know if we want to cross that hurdle again by moving to JSON right now.

jdarwood007:
As well, I moved the clear function to Load.php with the other cache stuff.  Makes more sense I think to have all the caching code together in Load.php rather than in Subs.php as well.  We can revert this change if needed.

Themes/default/languages/ManageSettings.english.php
```diff
@@ -131,6 +131,9 @@
 $txt['memcached_cache'] = 'Memcached';
 $txt['zend_cache'] = 'Zend Platform/Performance Suite';
 $txt['xcache_cache'] = 'XCache';
+$txt['cache_xcache'] = 'XCache settings';
```
albertlast: cache_xcache vs xcache_cache it's from my pov to near,
i would like to see somthing like that cache_xcache -> xcache_setting
to keep inline with the other fields and let to be clear.

jdarwood007:
Hows that.  I think that makes it more clear.

PR #3706, opened 2016-12-05
Fixes #3705

### LINKED ISSUE #3705: Session ID string too long for database field
Lately I have been getting the following error pretty frequently when logging in or logging out. These are database errors and therefore block the loading of the page with a big "Database Error" warning message as well as making entries in the error log. Here are sample entry from the error log.

```
f534a2c08a4a60ac536d4319d0fde30e
Type of error: Database
Database Error: Data too long for column 'session' at row 1
http://stovell.noip.me/~jon/dev/index.php?action=login2;sa=check;member=1
File: /Users/jon/Sites/dev/Sources/Logging.php
Line: 129
```

Here are the relevant entries from the list of queries in the debug output generated when the error occurred:

```
UPDATE dev_log_online
SET log_time = 1480969094, ip = unhex('c630a1e3'), url = '{\"action\":\"logout\",\"USER_AGENT\":\"Mozilla\\/5.0 (Macintosh; Intel Mac OS X 10_12_1) AppleWebKit\\/602.2.14 (KHTML, like Gecko) Version\\/10.0.1 Safari\\/602.2.14\"}'
WHERE session = 'ebjtrtug8lbudjb4ffev76lrg881b1eqbapqcsre8hadm782t4juvuunbhaotddnc6ehmvu4g8in92r4dtpfu1mdfnco47to80os692' 
in .../Sources/Logging.php line 102, which took 0.0007751 seconds at 0.04076791 into request.

DELETE FROM dev_log_online
WHERE id_member = 1 
in .../Sources/Logging.php line 122, which took 0.00082707 seconds at 0.04165888 into request.
REPLACE INTO dev_log_online(`session`, `id_member`, `id_spider`, `log_time`, `ip`, `url`)
VALUES
   ('ebjtrtug8lbudjb4ffev76lrg881b1eqbapqcsre8hadm782t4juvuunbhaotddnc6ehmvu4g8in92r4dtpfu1mdfnco47to80os692', 1, 0, 1480969094, unhex('c630a1e3'), '{\"action\":\"logout\",\"USER_AGENT\":\"Mozilla\\/5.0 (Macintosh; Intel Mac OS X 10_12_1) AppleWebKit\\/602.2.14 (KHTML, like Gecko) Version\\/10.0.1 Safari\\/602.2.14\"}') 
in .../Sources/Logging.php line 129, 
Notice: Undefined index: t in /Users/jon/Sites/dev/Sources/ViewQuery.php on line 133
which took 0 seconds.

INSERT INTO dev_log_errors(`id_member`, `log_time`, `ip`, `url`, `message`, `session`, `error_type`, `file`, `line`)
VALUES
   (1, 1480969094, unhex('c630a1e3'), SUBSTRING('?action=logout;ea2037243fee=f534a2c08a4a60ac536d4319d0fde30e', 1, 65534), SUBSTRING('Database Error: Data too long for column \'session\' at row 1', 1, 65534), 'f534a2c08a4a60ac536d4319d0fde30e', 'database', SUBSTRING('/Users/jon/Sites/dev/Sources/Logging.php', 1, 255), 129) 
in .../Sources/Errors.php line 112, which took 0.00079012 seconds at 0.04379797 into request.
```

According to one of the comments in [the PHP manual page for `session_id()`](http://php.net/manual/en/function.session-id.php), a session id string may be up to 128 characters long. In the sample above, we can see that the session id string is 103 characters long. 

The problem is, SMF configures the `session` field in the `{db_prefix}log_online` table to expect no more than 64 characters. Other tables are similarly configured. My guess is that the best thing would be to change these fields to expect up to 128 characters.

Sesquipedalian:


Sesquipedalian:
Does this look good to you, @albertlast?

albertlast:
Sure it's fine.
When i look in the issue the possible length of session could be anything between 64 - 128 characters.
In mysql would that mean we could generate many used space without content (default behavior of char fields).

So maybe we should change the datatype in mysql like in pg to varchar.
http://dev.mysql.com/doc/refman/5.7/en/char.html

Sesquipedalian:
Done.

PR #3714, opened 2016-12-08
Update SCEditor

### LINKED ISSUE #3711: scEditor Update
Leaving this here so we don't forget about it. Version 1.5.0 of scEditor was released recently (now at 1.5.2 after a couple more releases to fix bugs), fixing various bugs and adding support for jQuery 3 (though we don't officially support jQuery 3 and I'm not sure what if anything we need to do to support that - something else we can look into). We're still using 1.4.7 (the last version before 1.5.0 was released), so we should look into upgrading the editor in 2.1 to take advantage of these bug fixes.

colinschoen:
This fixes #3711


XinYenFon:
Lovely, so far what i found;

- The origin issue (https://github.com/samclarke/SCEditor/issues/492) reported fixed but on visual mode in editor its still looking broken, end result is also fixed.
source
visual

- (Minor) blockquote css definition inside the SCEditor itself is not matching with our style so we need to use our own changes (which we already have in current version). Please revert the changes on blockquote & blockquote cite:before (jquery.sceditor.default.css)

- (Minor) For styling again, can you please remove the background / border-bottom on div.sceditor-group (jquery.sceditor.css)

Sesquipedalian:
We probably don't want to overwrite the changes that were made in #3336. Without them, SCEditor doesn't do a very good job responding to on-the-fly changes in the size of the browser viewport.

colinschoen:
Thanks, missed that. Will fix.

colinschoen:
Those changes have been added back.

live627:
Is this 1.5.2?

live627:
@colinschoen Is this 1.5.2?

colinschoen:
Sorry missed that. Yes, this is `1.5.2`.

live627:
I added some padding to the smiley row and reapplied my previous fix to the smiley popup.

XinYenFon (approved):
Looks good.

PR #3718, opened 2016-12-10
Fix #3647

### LINKED ISSUE #3647: JSON conversion for themes fails
The upgrade fails to complete the conversion to json on the themes table.

The output shows the second to last request with this response data:

```
<?xml version="1.0" encoding="UTF-8"?>
    <smf>
        <get key="_">1477184841414</get>
    <table num="12">themes</table>
        </smf>
```

And the last response I get from the server with this data:

```
 done.<?xml version="1.0" encoding="UTF-8"?>
    <smf>
        <get key="_">1477184841415</get>
    <table num="13">themes</table>
        </smf>
```

I'm thinking the $debug needs to go away here at about Line 3933

```
                            if ($is_debug || $command_line)
                                echo ' done.';
```

live627:

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -1301,7 +1301,7 @@ function BackupDatabase()
 				return upgradeExit();
 		}
 
-		if ($is_debug && $command_line)
+		if ($command_line)
 		{
 			echo "\n" . ' Successful.\'' . "\n";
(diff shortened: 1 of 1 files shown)


colinschoen (approved):
LGTM, approval

PR #3719, opened 2016-12-11
Lint all the things

live627:
Someone should enable builds on Travis CI

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -5806,7 +5806,10 @@ function build_regex($strings, $delim = null)
 				$sub_regex = $index_to_regex($value, $delim);
 
 				if (count(array_keys($value)) == 1)
-					$new_key .= explode('(?'.'>', $sub_regex)[0];
+				{
+					$new_key_array = explode('(?'.'>', $sub_regex);
+					$new_key .= $new_key_array[0];
+				}
 				else
 					$sub_regex = '(?'.'>' . $sub_regex . ')';
 			}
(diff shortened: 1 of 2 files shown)


Sources/Subs.php
```diff
@@ -5806,7 +5806,10 @@ function build_regex($strings, $delim = null)
 				$sub_regex = $index_to_regex($value, $delim);
 
 				if (count(array_keys($value)) == 1)
-					$new_key .= explode('(?'.'>', $sub_regex)[0];
```
colinschoen, line 5809: What is this change for?

colinschoen (changes requested):
Not sure what Subs is being changed for.

Sources/Subs.php
live627: PHP 5.3 compatibility.

colinschoen:
@CoreISP or @Oldiesmann (or whoever has owners access) can we enable the travis CI extension.

colinschoen (approved):

colinschoen:
Bump

jdarwood007:
@colinschoen,

I am asking @Oldiesmann to setup travis.  Although at this point I'm not sure what use it would be without some proper tools to test with.  Do you have stuff in mind?  We do have a unreleased PHP unit that was for testing BBC and a few other things sitting around.  I think if you are willing to take on that PHP unit, we could get it updated, released and pushed for Travis CI to use.

colinschoen:
Thanks, yes willing to take that on.

jdarwood007:
https://travis-ci.org/SimpleMachines/SMF2.1

@Oldiesmann has set it up btw.

PR #3727, opened 2016-12-14
remove the pm_email_notify fix #3724

### LINKED ISSUE #3724: This table pm_email_notify need ??
TABLE {$db_prefix}members DROP INDEX 'pm_email_notify'   upgrade 2.1 b*
delete Sources/Subs-Members.php
`672  'new_pm', 'pm_prefs', 'show_online', 'pm_email_notify',`

albertlast:
Like i mention in #3724 i'm not 100% sure if this the right way.

The problem is mysql:
Because of the table lock,
the sql statement run on error if the col is already deleted.

The pg way is fast and run again a members table without the col will create no error message.

### DIFF
--- a/Sources/Subs-Members.php
+++ b/Sources/Subs-Members.php
@@ -669,7 +669,7 @@ function registerMember(&$regOptions, $return_errors = false)
 	// Right, now let's prepare for insertion.
 	$knownInts = array(
 		'date_registered', 'posts', 'id_group', 'last_login', 'instant_messages', 'unread_messages',
-		'new_pm', 'pm_prefs', 'show_online', 'pm_email_notify',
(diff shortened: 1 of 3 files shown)


jdarwood007:
The upgrader is supposed to silently handle errors during alter operations to handle being ran multiple times.  If its not doing this, we need to look at why.

albertlast:
i mention only the behavior of the sql statement not the smf part.
I don't know how smf handle the situation when the sql runs on error because the col is already gone.

jdarwood007:
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/other/upgrade.php#L2142

albertlast:
Than should be everything fine.

PR #3731, opened 2016-12-17
Fixes #3367

### LINKED ISSUE #3367: you must enter a value between 1 and 525600
enter a valid username in the log in popup, enter an incorrect password and check the box to always stay logged in. try to log in. log in will fail. try to log in with the correct password and the always stay logged in box checked. get error message about invalid value.

frandominguezl:
Fixes #3367 
Signed-off by: Frankie "d3vcho" <d3vcho@gmail.com>


colinschoen (approved):
LGTM, approval

PR #3733, opened 2016-12-19
Removed db_insert_sql

albertlast:
I didn't understand the idea behind the function AND
can't finde any use in the code--> remove


jdarwood007:
Its actually only used in the tool to convert from sqlite to mysql far as I can tell.  But in that case the function is overridden for a localized version.

Oldiesmann:
It's left over from when you could backup your database from the admin center (in SMF 1.x and early versions of 2.0 - it was ultimately removed because it didn't work very well and there are better options for backing up your database). I guess the code for that was never removed.

equinoxmatt:
If it has no use, I think it is worth removing it.

colinschoen (approved):
LGTM, approval

PR #3736, opened 2016-12-22
added return value to db_insert

albertlast:
I saw many places where after db_insert the db_insert_id is calling
and i realize that the db_insert function is able to insert multipe values at once but then the db_insert_id get then broken.

So with this pr I solve this issues your are able to get directly the id back and
when your insert multiple values your get multiple id's.

To handle this different needs I place different returnmods = 
0 do nothing (default)
1 return the id of the last row
2 return an array of all id of every row

The challange is here that mysql is less potential as pg.
in pg last insert id works on all insert mode, on all type of fields and unlimited number of fields in the table
in mysql last insert works only on normal insert (no replace or ignore), only on ai fields and only one field

The first limitation of mysql I implemented in pg that the return only works on method == ''
for the second I don't see any checks
the last one when multiple keys are provided I take the first one -> the first one needs to be the ai field.

When this pr is merged,
the next step would be to remove the places where db_insert_id is used.

### DIFF
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -701,8 +701,10 @@ function smf_db_error($db_string, $connection = null)
  * @param array $keys The keys for the table
  * @param bool $disable_trans Whether to disable transactions
  * @param resource $connection The connection to use (if null, $db_connection is used)
+ * @param returnmode 0 = nothing(default), 1 = last row id, 2 = all rows id as array; every mode runs only with method = ''
+ * @return value of the first key, behavior based on returnmode
  */
-function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $disable_trans = false, $connection = null)
+function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $disable_trans = false, $connection = null, $returnmode = 0)
 {
 	global $smcFunc, $db_connection, $db_prefix;
 
@@ -754,6 +756,21 @@ function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $dis
 		),
 		$connection
 	);
+	
+	if(!empty($keys) && (count($keys) > 0) && $method == '' && $returnmode > 0)
+	{
+		if ($returnmode == 1)
+			$return_var = smf_db_insert_id($table, $keys[0]) + count($insertRows) - 1;
+		else if ($returnmode == 2)
+		{
+			$return_var = array();
+			$count = count($insertRows);
+			$start = smf_db_insert_id($table, $keys[0]);
+			for ($i = 0; $i < $count; $i++ )
+				$return_var[] = $start + $i;
+		}
+		return $return_var;
+	}
 }
 
 /**
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -701,7 +701,7 @@ function smf_db_error($db_string, $connection = null)
  * @param array $keys The keys for the table
  * @param bool $disable_trans Whether to disable transactions
  * @param resource $connection The connection to use (if null, $db_connection is used)
- * @param returnmode 0 = nothing(default), 1 = last row id, 2 = all rows id as array; every mode runs only with method = ''
+ * @param int returnmode 0 = nothing(default), 1 = last row id, 2 = all rows id as array; every mode runs only with method = ''
  * @return value of the first key, behavior based on returnmode
  */
 function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $disable_trans = false, $connection = null, $returnmode = 0)
--- a/Sources/Subs-Db-mysqli.php
+++ b/Sources/Subs-Db-mysqli.php
@@ -759,8 +759,10 @@ function smf_db_error($db_string, $connection = null)
  * @param array $keys The keys for the table
  * @param bool $disable_trans Whether to disable transactions
  * @param object $connection The connection to use (if null, $db_connection is used)
+ * @param returnmode 0 = nothing(default), 1 = last row id, 2 = all rows id as array; every mode runs only with method = ''
+ * @return value of the first key, behavior based on returnmode
  */
-function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $disable_trans = false, $connection = null)
+function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $disable_trans = false, $connection = null, $returnmode = 0)
 {
 	global $smcFunc, $db_connection, $db_prefix;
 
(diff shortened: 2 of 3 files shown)


Sources/Subs-Db-postgresql.php
```diff
@@ -776,22 +788,40 @@ function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $dis
...
+					', $insertRows).$replace.$returning,
 				array(
 					'security_override' => true,
 					'db_error_skip' => $method == 'ignore' || $table === $db_prefix . 'log_errors',
 				),
 				$connection
 			);
+			
+			//get from the last insert the return value
+			if($with_returning && $request !== false && $row = $smcFunc['db_fetch_assoc']($request, count($insertRows) - 1))
```
colinschoen: `if (` instead of `if(`

Sources/Subs-Db-mysqli.php
```diff
@@ -807,6 +809,21 @@ function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $dis
 		),
 		$connection
 	);
+	
+	if(!empty($keys) && (count($keys) > 0) && $method == '' && $returnmode > 0)
+	{
+		if ($returnmode == 1)
```
colinschoen: Switch here too

Sources/Subs-Db-mysql.php
```diff
@@ -749,6 +751,21 @@ function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $dis
 		),
 		$connection
 	);
+	
+	if(!empty($keys) && (count($keys) > 0) && $method == '' && $returnmode > 0)
+	{
+		if ($returnmode == 1)
```
colinschoen, line 762: Can we use a switch here.

Sources/Subs-Db-postgresql.php
```diff
@@ -776,22 +788,40 @@ function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $dis
...
+			{
+				if ($returnmode == 1)
```
colinschoen: Switch

Sources/Subs-Db-postgresql.php
```diff
@@ -776,22 +788,40 @@ function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $dis
...
+						$return_var[] = $row[$keys[0]];
+				}
+			}
+			
 	}
 
 	if ($priv_trans)
 		$smcFunc['db_transaction']('commit', $connection);
+	
+	if($with_returning && !empty($return_var))
```
colinschoen: `if (` instead of `if(`

colinschoen (changes requested):
Nits

Sources/Subs-Db-mysql.php
albertlast: With switch your mean case?
Why should be a case here better as if?
use more space/loc and we got only two possibilities

live627: why use a switch if only two permutations?

Sources/Subs-Db-mysql.php
```diff
@@ -696,8 +696,10 @@ function smf_db_error($db_string, $connection = null)
  * @param array $keys The keys for the table
  * @param bool $disable_trans Whether to disable transactions
  * @param resource $connection The connection to use (if null, $db_connection is used)
+ * @param returnmode 0 = nothing(default), 1 = last row id, 2 = all rows id as array; every mode runs only with method = ''
```
live627: invalid docblock, should be `@param int $returnmode`

live627 (changes requested):

Sources/Subs-Db-mysql.php
colinschoen: Agreed, not needed. Apologies, disregard this comment.

Sources/Subs-Db-mysqli.php
colinschoen: Disregard

Sources/Subs-Db-postgresql.php
colinschoen: Disregard

Sources/Subs-Db-mysql.php
albertlast: Int would be not fully true,
from technical pov is int[] when returnmode is 2 and
when you use pg string and string[] is possible, too.

That was I mean with:

> only on ai fields -- for the second I don't see any checks

i mean we could hide this behavior if we that don't mention ...

Whoops.. wrong docblock i was looking...

Sources/Subs-Db-postgresql.php
```diff
@@ -788,39 +788,46 @@ function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $dis
...
+			if ($returnmode === 2)
+				$return_var = array();
+
+			while($row = $smcFunc['db_fetch_assoc']($request) && $with_returning)
 			{
-				if ($returnmode == 1) 
-					while($row = $smcFunc['db_fetch_assoc']($request))
-						$return_var = $row[$keys[0]];
-				elseif ($returnmode == 2)
+				if (is_int($row[$keys[0]])) // try to emulate mysql limitation
```
albertlast: I'm not 100% sure if I should do this,
but this in my eyes the only way to check if the fields is a ai field(int).

The main fokus is for person how develop something on smf with pg and
hop the stuff is running on mysql too.
But on the other side maybe is this behavior for this mod/addon (not only we develop for smf)
not important...

Sources/Subs-Db-postgresql.php
```diff
@@ -776,22 +788,47 @@ function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $dis
...
+		);
+
+		if ($with_returning && $request !== false)
+		{
+			if ($returnmode === 2)
+				$return_var = array();
+
+			while($row = $smcFunc['db_fetch_assoc']($request) && $with_returning)
+			{
+				if (is_int($row[$keys[0]])) // try to emulate mysql limitation
```
albertlast, line 815: I'm not 100% sure if I should do this,
but this in my eyes the only way to check if the fields is a ai field(int).

The main fokus is for person how develop something on smf with pg and
hop the stuff is running on mysql too.
But on the other side maybe is this behavior for this mod/addon (not only we develop for smf)
not important...

albertlast:
So I clean up and reduce the dupplicate of logic.
And place a check if the return value is a integer.

But like I already mansion,
i'm not sure how fare i should emulate the limitation of mysql in pg.

implementet behavior
- only mode ''
- only the first key
- only int

albertlast:
@live627 oldiesmann mention that we doesn't want api change in rc1.
so u are sure about for targeting this for rc1?

live627:
changed it back

PR #3744, opened 2016-12-27
Add reCAPTCHA

live627:


Themes/default/GenericControls.template.php
```diff
@@ -266,27 +266,37 @@ function template_control_verification($verify_id, $display_type = 'all', $reset
...
-		// Do the actual stuff - image first?
-		if ($i == 0 && $verify_context['show_visual'])
+		// Do the actual stuff
+		if ($i == 0)
 		{
-			if ($context['use_graphic_library'])
+			if ($verify_context['can_recaptcha'])
+			{
 				echo '
+				<div class="g-recaptcha centertext" data-sitekey="' . $verify_context['recaptcha_public_key'] . '" data-theme="' . $verify_context['recaptcha_theme'] . '"></div><br>
```
XinYenFon: I guess you forgot to change this 😆 

```
<div class="g-recaptcha centertext" data-sitekey="' . $modSettings['recaptcha_site_key'] . '" data-theme="' . $modSettings['recaptcha_secret_key'] . '"></div><br>
```

colinschoen (commented):
Excellent, will review soon.

Sources/Subs-Editor.php
```diff
@@ -2134,6 +2145,23 @@ function create_control_verification(&$verificationOptions, $do_test = false)
...
+		{
+			require_once($sourcedir . '/ReCaptcha/autoload.php');
+
+			$reCaptcha = new \ReCaptcha\ReCaptcha($modSettings['recaptcha_secret_key']);
+
+			// Was there a reCAPTCHA response?
+			if (isset($_POST['g-recaptcha-response']))
+			{
+				$resp = $reCaptcha->verify($_POST['g-recaptcha-response'], $_SERVER['REMOTE_ADDR']);
+
```
jdarwood007, line 2158: Should use $user_info['ip'] instead?

jdarwood007:
Would it be possible to make this a API instead?  So we can swap out/add additional anti-spam verification methods at ease?  Would suggest this be a multiple allowed methods and convert questions to a method as well as the image verification.   Would be a big change in the code, but overall allows new methods to be added with ease.

live627:
You mean convert this to separate objects? Interesting idea.

Themes/default/GenericControls.template.php
live627: indeed.

live627:
This would delay the beta even more. Can the rewrite wait?

jdarwood007:
Beta 4 milestone?  It could wait.

Overall just a suggestion as anti-spam methods come and go fairly quickly, although recaptcha has stuck around and done a nice job.  It would just mean if a new method comes to be, a simple drop in of the files would enable its use, rather than changing more code.

live627:
Cool. I want to get this in B4.

jdarwood007:
I don't think we can do a API change like that in RC. So if we don't make it an API in B3/B4 then it wont make 2.1.

Oldiesmann (commented):
I agree with SleePy about moving this to an API. Other than that I'm fine with adding support for this.

Sesquipedalian:
I'll be glad to see this go in. 👍

live627:
I don't have the energy, motivation to do a full API change like people want.

PR #3747, opened 2016-12-28
Update headers

live627:


Oldiesmann (changes requested):
At this point we should hold off on changing the version until after we release Beta 3. I'm fine with going ahead and changing the year since we're obviously not going to get a Beta 3 out before 2017.

live627:
ok, only the year is updated now.

colinschoen (approved):
LGTM, approval

PR #3760, opened 2016-12-29
Remove unused code in ManageLanguages.php

### LINKED ISSUE #3730: 2 errors in ManageLanguages.php after installing language pack
Received this two errors after installing Spanish_es-utf8 (Beta 2, from Admin Panel) in the latest GitHub build:

Undefined index: images
Invalid argument supplied for foreach()

Both in line 379 of ManageLanguages.php

colinschoen:
Fixes #3730.

`$context['theme_names']` is unused.
`$theme_indexes` is unused. 

I see no need for this logic to be here.

### DIFF
--- a/Sources/ManageLanguages.php
+++ b/Sources/ManageLanguages.php
@@ -358,9 +358,6 @@ function DownloadLanguage()
 			// Add the context data to the main set.
 			$context['files']['lang'][] = $context_data;
 		}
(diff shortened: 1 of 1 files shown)


Oldiesmann (approved):
Looks good to me. I think this might be leftover code from when we had some language-specific theme images (such as the one that showed "new" next to a new post). As far as I can see it's not actually used anywhere anymore.

colinschoen:
Thanks for reviewing.

Issue #3583, opened 2016-09-13
Not Support In MariaDB ?


creativefull:
```
Critical Error!
The installer was unable to detect any database support in PHP. Please ask your host to ensure that PHP was compiled with the desired database, or that the proper extension is being loaded.
```

why ? i'm using archlinux
php -> 7.0.9
MariaDB -> 10.1.16

so i need mysql to run this application ?

tinoest:
Can you provide the output of your phpinfo please? 

MariaDB is a drop in replacement for MySQL. 

So SMF should work. 

Just tested on a local VM works ok with;

mysql --version
mysql  Ver 15.1 Distrib 10.0.26-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2

albertlast:
with 10.1.9-MariaDB no problem too.

creativefull:
mysql  Ver 15.1 Distrib 10.1.16-MariaDB, for Linux (x86_64) using readline 5.1

albertlast:
i don't understand what you want tell us with that,
but your php need mysqli support,
because smf 2.1 support mysql and mysqli.
And php 7.x didn't support mysql_\* anymore.

creativefull:
i don't know ,.. when i create a project with mysqli it's alright,

maybe i'm missing a configuration in smf ?

tinoest:
No your not, MySQLi is the latest version of the MySQL libs. PHP7 doesn't support the old MySQL libs so it wouldn't work with them. 

If you want to use the older deprecated libraries then downgrade you php version.

albertlast:
Maybe we should place a auto switch, so that smf 2.1 + php 7 + mysql > mysqli

---

Von: Martyn Brown notifications@github.com
Gesendet: 14.09.2016 7:55 AM
An: SimpleMachines/SMF2.1
Cc: albertlast; Comment
Betreff: Re: [SimpleMachines/SMF2.1] Not Support In MariaDB ? (#3583)

No your not, MySQLi is the latest version of the MySQL libs. PHP7 doesn't support the old MySQL libs so it wouldn't work with them.

If you want to use the older deprecated libraries then downgrade you php version.

##

tinoest:
It should be removed from the install options. 

Although I'd hope people could read the docs on PHP's website.

albertlast:
But this info is in configuration and below 7 is okay to use mysql

---

Von: Martyn Brown notifications@github.com
[...]

Issue #3554, opened 2016-08-26
ConvertUTF8 issue on upgrade


XinYenFon:
I just wanted to open this issue as track and discuss. Well actually @equinoxmatt send me a fix and its working. I noticed some small issues, upgrade.php now seems to trying (?) to do utf8 & json convert at same time and its causing upgrade.php stuck (after completing the process). I rerun the upgrade.php and its completed w/o problem. Function converted all tables but {prefix}_user_likes.

I tested some basics of forum after upgrade everything seems working fine (do mind I used empty DB, we're still not sure about the function working perfectly).

![screenshot-localhost 2016-08-26 03-29-58](https://cloud.githubusercontent.com/assets/1669995/17990385/f128e06a-6b3d-11e6-87fb-bb9a219b0681.png)

equinoxmatt:
Hmm, I didn't get that when I tried it,  I will take another look tomorrow.  I noticed some really odd stuff with the MySQLConvertOldIp() function.  Might be worth another issue,  but some of the subStep code seems to be not behaving correctly. 

Not sure if it is all somehow related?  I will do some more digging over the next few days.

XinYenFon:
I did another test run today, and this is still happening. Way to reproduce; "Install non-utf8 SMF 2.0 then upgrade to 2.1 (latest)".

Do mind upgrade process actually ends but user_likes table stays as is.

sbulen:
I don't think user_likes is in vanilla 2.0.

Was this from a mod, in 2.0?   Or is this a 2.0 table being setup incorrectly?

Oldiesmann:
user_likes is a new table in 2.1.

sbulen:
Is the mentions table OK?   How about the moderator_groups table?  
I'm trying to review this code, they all appear to be created similarly in the upgrade script.

equinoxmatt:
Looking a bit more into this today.  The reason we get the odd progress bar update is the following line:

`$upcontext['table_count'] = 13;`
	`$upcontext['cur_table_num'] = $_GET['substep'];`
`	$upcontext['cur_table_name'] = isset($keys[$_GET['substep']]) ? $keys[$_GET['substep']] : $keys[0];`
`	$upcontext['step_progress'] = (int) (($upcontext['cur_table_num'] / $upcontext['table_count']) * 100);
`

table_count is set to 13,  with 13 harcoded table names.  So in the convertToUTF8,  you had 75 tables.  The calculation is this:

[...]

PR #3761, opened 2016-12-29
Clean up upgrade.php

NanoSector:
When running upgrade.php I noticed a lot of issues pop up. I'm going to clean it up as much as I can to hopefully make maintenance easier.


NanoSector:
Do we really need CleanupMods? It seems to be touching a whole lot of stuff the upgrader shouldn't really be doing (it parses a modification's XML file to see if it's still valid, something the package manager in the admin panel does just fine). If you've just upgraded you really shouldn't be touching old mods.

albertlast:
When your already touching the upgrade.php,
an "step" who is missing the config rewrite from mysql to mysqli see the part of autoswitch: https://github.com/SimpleMachines/SMF2.1/issues/3583

jdarwood007:
CleanupMods was added in 2.0 beta phase as a way to attempt to automatically reinstall mods.  It caused a lot of problems, so it was just disabled until we could get around to resolving those issues.  It never made it by the time 2.0 was gold.

NanoSector:
We probably should look into recreating it in a sane way, then. A lot has changed and we should be able to use the Subs files' functions now instead of manually checking packages.

sbulen:
Thanks Yoshi...   While in Upgrade.php, you might want to take a peek at issue #3554

NanoSector:
I'll take a look, thanks.

NanoSector:
Why do we populate the entire user array during upgrade? Shouldn't it be enough to just check if the user has admin access?

live627:
The upgrader is still a complete mystery to me. Many thanks for fixing it.

NanoSector:
I kind of want to avoid a last-minute rewrite to further stretch the next release but I really feel that a new upgrade/installation framework should be put in place. It is a big mess indeed, lots of old code.

The thing that worries me most about the current upgrader is that there is a lot of code copied from the Sources files. IMO, this is a direct consequence of having huge functions in many of those files. For instance, checkLogin copies huge chunks of password checking etc. from LogInOut.php, but it cannot use the functions inside LogInOut.php because they were written for a specific use case (and can't be used for any other use case). This in the end means that if we ever decide to change the behavior of those functions (or even worse, fix a security issue), Upgrade.php will not reflect these changes, possibly leaving vulnerabilities open. Fixing this then gets messy because some chunks are copied directly from the Sources files, some are altered a bit, so we're essentially forced to fix the same thing twice.

Fixing this mess is possible by splitting code like LogInOut.php into smaller, specialized functions; Upgrade.php can then simply directly include LogInOut.php and utilize these functions. This avoids code duplication as the same code can be reused internally too and allows easier maintenance of code (and as a bonus gets us a much higher code rating in Scrutinizer).

I'm not sure if this should be done in 2.1 though as it pretty much means rewriting half the codebase, but it is something I would love to see in 3.0.

Oldiesmann:
At this point I think it's best to just clean up upgrade.php as much as we can and look at rewriting everything for the next version of SMF. It would be a lot of work to completely rewrite all of that at this point, especially given how long 2.1 has been in development.

NanoSector:
Yeah. I'll probably get some more cleaning up done today.

NanoSector:
When digging through the code I noticed we only recommend to backup your tables when upgrading from YaBB SE (and I changed that to always recommend it). That got me thinking though, do we really want to have the upgrade script backup tables? I mean, it's nice to have it, but shouldn't we encourage the user to make a full backup of their database first, since that's better than just copying tables?

other/upgrade.php
```diff
@@ -28,15 +35,6 @@
...
-		'version' => '5.0.3',
-		'version_check' => 'return min(mysql_get_server_info(), mysql_get_client_info());',
-		'utf8_support' => true,
-		'utf8_version' => '5.0.3',
-		'utf8_version_check' => 'return mysql_get_server_info();',
-		'alter_support' => true,
-	),
 	'postgresql' => array(
 		'name' => 'PostgreSQL',
 		'version' => '8.0',
```
tinoest: From this comment here; http://www.simplemachines.org/community/index.php?topic=547286.msg3890058#msg3890058

PostgreSQL is meant to be version 9.1 minimum now.

jdarwood007:
Yabb != Yabb SE.
We still support upgrading from Yabb SE (see the upgrade_1.0.sql).  Yabb needs a converter.


Have we tested a SMF 1.1 or 2.0 install upgrading to 2.1 with mysqli?  Does the upgrader handler that smoothly?

NanoSector:
But we have converters for both of them:
http://www.simplemachines.org/community/index.php?topic=140776.0

I would test but I only have php7 on my systems, that doesn't work with
smf2.0 or 1.1.

Op wo 4 jan. 2017 01:45 schreef Jeremy D <notifications@github.com>:

NanoSector:
Sorry, didn't notice your comment; I'll fix that tomorrow.

Op di 3 jan. 2017 21:28 schreef Martyn Brown <notifications@github.com>:

live627:
@Yoshi2889 do you wish to add more changes to this? Can I merge this?

sbulen:
I'm looking forward to this set of changes!

NanoSector:
I don't think I'd want to add more changes, thanks for fixing that postgre
issue. I cannot test the whole thing since I only have access to PHP 7 now
so if someone can test if it still upgrades fine from both 1.1 and 2.0 I
think it's good to go.

Op zo 8 jan. 2017 06:11 schreef sbulen <notifications@github.com>:

other/install.php
```diff
@@ -523,9 +523,8 @@ function CheckFilesWritable()
 		'Themes',
 		'agreement.txt',
 		'Settings.php'
```
Oldiesmann: You're missing a comma at the end of line 525

NanoSector:
Didnt we change the mysqli think back to mysql? I think Upgrade.php has not been changed accordingly, then..

live627:
I already changed it.

NanoSector:
Ah, cool.

PR #3775, opened 2017-01-02
Rewrite db_insert_id call

albertlast:
rearrange the paramets of the function
because the other are never called.


jdarwood007:
The old way works just fine though correct?

albertlast:
Sure I didn't touch the old way.

albertlast:
Drop this pr in favor #3777 to got only on pr with all changes.

PR #3776, opened 2017-01-02
added mini files to git ignore

albertlast:
I think it's a good idea to add this two files
to the ingore list from git.


colinschoen (approved):
LGTM, approval

PR #3777, opened 2017-01-02
Align transaction handling

albertlast:
Drop the transaction handling in pg because mysql doesn't support this.
And i'm lazy -> i build this pr on the rewrite db_insert_id pr
to see only the real changes: https://github.com/albertlast/SMF2.1/commit/df30aa0f3a5118f0f256569e03e2e1697837be65

### DIFF
--- a/Sources/RepairBoards.php
+++ b/Sources/RepairBoards.php
@@ -264,7 +264,7 @@ function loadForumTests()
 				$memberStartedID = (int) getMsgMemberID($row['myid_first_msg']);
 				$memberUpdatedID = (int) getMsgMemberID($row['myid_last_msg']);
 
-				$smcFunc['db_insert']('',
+				$newTopicID = $smcFunc['db_insert']('',
 					'{db_prefix}topics',
 					array(
 						'id_board' => 'int',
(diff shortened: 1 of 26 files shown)


jdarwood007:
MySQL supports transactions http://dev.mysql.com/doc/refman/5.7/en/commit.html

I'm having trouble finding when it was added, but far as I can tell it has existed since 3.23.28 and is only supported on InnoDB tables (no MyISAM).

albertlast:
i know that it's support transaction,
but i don't like that db_insert works differently between mysql and pg.
the other way would be to add this to mysql too.

jdarwood007:
Ok, I see your commit now.  Removing it from the insert function is fine, if we need to call the transaction we can do it ourselves rather than in the insert function.

Sources/RepairBoards.php
```diff
@@ -402,12 +401,11 @@ function fix_full_processing:
 							'xx',
 							1,
 						),
-						array('id_topic')
+						array('id_topic'),
```
live627: This should be `id_msg`

Sources/RepairBoards.php
```diff
@@ -538,12 +535,11 @@ function fix_full_processing:
 						'xx',
 						1,
 					),
-					array('id_topic')
+					array('id_topic'),
```
live627: Same as above.

live627 (changes requested):

PR #3778, opened 2017-01-02
Added SQLite Database Caching

tinoest:
Added SQLite Database Based Cache, in some instances its faster than filebased. Simple Class currently, if its wanted then I'll add additional features/checking to ensure its more stable.


Sources/CacheAPI-sqlite.php
```diff
@@ -0,0 +1,145 @@
...
+
+		parent::__construct();
+
+		// Set our default cachedir.
+		$this->setCachedir();
+
+		$database = $this->cachedir.'/'.'SQLite3Cache.db3';
+
+		$this->cacheDB = new SQLite3($database);
+		$this->cacheDB->busyTimeout(1000);
```
jdarwood007, line 48: Instead of connecting here in the __construct(), I suggest using the connect() method instead.  __construct() is called every time the API is called, while connect() is only called when we actually intend to use it.  __construct would also get called for when visiting the cache settings page as well.

Sources/CacheAPI-sqlite.php
```diff
@@ -0,0 +1,145 @@
...
+
+		return parent::isSupported() && $supported;
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public function getData($key, $ttl = null)
+	{
+		$ttl		= time();
```
jdarwood007: Was the TTL here to force the cache to always pull valid entries?  How do we handle old entries (automatic cleanup/garbage collection)?

Sources/CacheAPI-sqlite.php
```diff
@@ -0,0 +1,145 @@
...
+
+    if (!isset($context['settings_post_javascript']))
+      $context['settings_post_javascript'] = '';
+
+    $context['settings_post_javascript'] .= '
+      $("#cache_accelerator").change(function (e) {
+        var cache_type = e.currentTarget.value;
+        $("#cachedir_sqlite").prop("disabled", cache_type != "sqlite");
+      });';
+  }
```
jdarwood007: Minor note on this is the tabbing is off.

Themes/default/languages/Login.english.php
```diff
@@ -16,6 +16,7 @@
 $txt['need_username'] = 'You need to fill in a username.';
 $txt['no_password'] = 'You didn\'t enter your password.';
 $txt['incorrect_password'] = 'Password incorrect';
+$txt['integrate_validate_login_failure'] = 'Integrate Validate Login Failure';
```
jdarwood007: Doesn't seem related?

jdarwood007 (changes requested):
Thanks for your PR.

Sources/CacheAPI-sqlite.php
tinoest: Will do.

Sources/CacheAPI-sqlite.php
tinoest: I copied and pasted, its set to ts=2 what does SMF use?

Themes/default/languages/Login.english.php
tinoest: It's not I'll remove it from this PR was from something else I was working on.

Sources/CacheAPI-sqlite.php
tinoest: I haven't added that yet, will do shortly. 
I couldn't work out what the passed in ttl was meant to be used for in this inherited class as it didn't make sense to me, when you put data it makes sense but getting I'm not so sure.

jdarwood007: We issue the ttl in the get as well as put as some cache methods need it.  But it can also be used to check to make sure that the cache is still valid.

Sources/CacheAPI-sqlite.php
jdarwood007: We use tabs, not spaces for indentation of code.

tinoest: Ok, well on my screen it's indented correctly. So can only be my setting of ts in vim

live627: I can push a commit to fix that...

tinoest: Please do, I can then pull your version into my source, unless you want to do it when it's complete and ready for merging

Sources/CacheAPI-sqlite.php
tinoest: So anything less than now - ttl can be cleared? As it doesn't seem to contain a timestamp so I assume is relative to now. Also the replace into in the put will update any that are expired it's just those we've not accessed often which will get left. I don't see that as a huge issue looking at the other cache methods they don't seem to do garbage collection here

jdarwood007: $ttl still contains the local valid value (such as 90 seconds).

For the get in this case, I would recommend making sure the ttl stored in the cache is above time() - $ttl.

For garbage collection I would recommend deleting anything order than time(), because you store the $ttl (in the put) as time() + $ttl.

SMF 2.1 includes in its $db_show_debug cache hits and misses.  So using misses will help you adjust and see what cached items are not valid or are being rejected and thus we are not using the cache.

Sources/CacheAPI-sqlite.php
live627: I pushed it to this PR.

Sources/CacheAPI-sqlite.php
tinoest: I don't think this should be done in the get request. It's meant to speed things up and it's just adding extra code which if it's a key which already exists will be cleared in the next put request anyway.

I'll add a cron to be called to clear out any data which isn't changed much.

tinoest: Actually this now no different to how the other cache methods work, so either they all should change behaviour or I'll leave it as is.

Sources/CacheAPI-sqlite.php
tinoest: Thanks, pulled it into mine and made one change.

Issue #3498, opened 2016-06-19
Upgrade problem (2.0.11 to 2.1 b2)


bobrenja:
download http://download.simplemachines.org

upgrade error 

![2016-06-19 1 04 06](https://cloud.githubusercontent.com/assets/20020367/16174556/5958ca0a-35cf-11e6-825f-a8bea22da728.png)

XinYenFon:
You can upload images here and also can you write your server-details (php-mysql versions) for better debug.

bobrenja:
MySQL 5.0.67
Php 5.4.45

albertlast:
Make a new copy of you 2.0.11 in mysql,
use the github version https://github.com/SimpleMachines/SMF2.1/archive/release-2.1.zip
move the upgrage\* files from others to the root  directory and
try again to upgrade.

bobrenja:
Where to take russian for release 2.1?

albertlast:
I don't know where the language file are,
but in the worst case (when you realy need them) take the beta2 language packages.

bobrenja:
Settings.php move the  files from others&

No login input

![2016-06-19 9 18 45](https://cloud.githubusercontent.com/assets/20020367/16175674/e90ec562-35fe-11e6-92eb-787da6c54048.png)

albertlast:
Than try without,
maybe after the upgrade process.

bobrenja:
Settings.php move the  from others?

albertlast:
Nope

bobrenja:
Use the github version https://github.com/SimpleMachines/SMF2.1/archive/release-2.1.zip
Write "Thank you for choosing to upgrade to SMF 2.1 Beta 3...."

**So it should be?**

![2016-06-19 11 21 42](https://cloud.githubusercontent.com/assets/20020367/16176194/603bc34a-3610-11e6-9369-e5cf59ab4090.png)

no files

Themes/default/css/**minified.css**
Themes/default/scripts/**minified.js**
Themes/default/scripts/**minified_deferred.js**

iRickyJ:
You can create them manually and set them to chmod 644

PS: I also had to create them , why not to provide these files by default with package ?

albertlast:
I think this is a real issue @MissAllSunday  that the file handling is wrong

bobrenja:
So, these files should be in the package, or they need to be downloaded separately ??
They need to work offline or not?
If needed, where to download?

[...]

PR #3787, opened 2017-01-03
Stop Tracking Minified Files in Git

colinschoen:


XinYenFon:
? https://github.com/SimpleMachines/SMF2.1/pull/3776 ?

colinschoen:
Adding a file to `.gitignore` does no good if it is already being tracked by the repository from an earlier commit. This removes the files from the repository. Essentially undoing 4d5138c. Not sure why they were added in the first place.

XinYenFon:
hmm... I think you should reset your local repo to SMF2.1/release-2.1.

git reset --hard should help you on that.

colinschoen:
Why would I `git reset` What am I resetting to? `HEAD`? The minified files currently exist in the `HEAD` of `release-2.1`. This commit removes the blank versions of the minified files which shouldn't be necessary since they are created automatically.

XinYenFon:
If I recall correctly there are certain cases SMF cannot create those files, that's reason those files created in the first place.

Ref: https://github.com/SimpleMachines/SMF2.1/issues/3498

colinschoen:
Hmm annoying. Then I suppose there will always will be the issue of modified tracked files that we don't care about, unless there is some other git trickery that I don't know of.

colinschoen:
This might work: 

https://git-scm.com/docs/git-update-index

`git update-index --skip-worktree Themes/default/scripts/minified.js`
`git update-index --skip-worktree Themes/default/css/minified.css`

XinYenFon:
There won't be if you disable minimizing on localhost or just reset to SMF2.1/release-2.1 (not HEAD), btw HEAD is not this repo's HEAD its your origin repo's HEAD. That's why your resets won't fix it.

If you add this in your .git/config

```
[remote "upstream"]
   url = git://github.com/SimpleMachines/SMF2.1.git
   fetch = +refs/heads/*:refs/remotes/upstream/*
   fetch = +refs/pull/*/head:refs/remotes/upstream/pr/*
```

This will enable this repo as your upstream (less trouble), then you can use following code to reset your local repo to latest build in this repo;

```
git reset --HARD upstream/release-2.1
git push -f origin release-2.1
```

colinschoen:
I was talking about `HEAD` prior to committing. I don't want to have to reset every time I load the forum. The solution I posted above works. Closing this pull request.

XinYenFon:
You should not be resetting your repo unless something bad happens, you just pull & push without committing to your origin/release-2.1, for PRs you use branching. Just typing it here so if someone reads in future to solve similar problem can use either of the methods.

colinschoen:
No, I'm not ever running `git reset`, but I would have to use that or `git stash` without using my other solution in order to stop have git showing the minified files were changed after loading any forum page. Changing the upstream won't change which files are tracked by git so I don't see how that would fix the issue.

PR #3790, opened 2017-01-03
Fixes layout in dl.settings lists (e.g. member search form)

Sesquipedalian:


Sesquipedalian:
Thoughts, @Antes?

Sesquipedalian:
By the way, despite what GitHub says [should be the case](https://help.github.com/articles/requesting-a-pull-request-review/), I am not being shown the option to request reviews on the pull requests I submit. Perhaps the repo is configured to deny "Contributors" the ability to do that? If so, could this be changed?

XinYenFon:
You can't request from me, I don't have any push access to this repo. Try requesting from @colinschoen @live627 @Oldiesmann

Sesquipedalian:
No, I mean I can't request reviews from anyone. The option isn't there.

Sesquipedalian:
screen shot 2017-01-03 at 3 35 38 pm
No gear icon available, so no review requests for me. :(

Themes/default/css/index.css
```diff
@@ -610,11 +610,17 @@ dl.settings dt.windowbg {
 	clear: both;
 }
 dl.settings dd {
-	width: 42%;
+	width: 50%;
 	float: right;
 	overflow: auto;
 	margin: 0 0 3px 0;
 }
+dl.settings dd + dd {
```
live627: Does this syntax work in IE9?

colinschoen:
Hmm if possible we should try to fix that so you can request reviews. @CoreISP @Oldiesmann. Are you able to change that?

Themes/default/css/index.css
Sesquipedalian: Yes. It's a CSS2 selector, supported in IE7+.

Sesquipedalian: https://msdn.microsoft.com/en-us/library/hh781508(v=vs.85).aspx#combinators

live627: neat, I didn't know that.

Themes/default/css/index.css
```diff
@@ -597,9 +597,9 @@ dl.settings {
 	padding: 5px;
 }
 dl.settings dt {
-	width: 56%;
+	width: 45%;
```
XinYenFon: Changing this will change alot in SMF, are you sure it won't break anything else ?

XinYenFon:
I think we can do this without touching CSS.

screenshot-localhost 2017-01-05 01-23-23
screenshot-localhost 2017-01-05 01-23-38

Themes/default/css/index.css
Sesquipedalian: From what I can tell, it is currently broken everywhere else, too. But I can change the percentages back if you prefer. Those were mostly just cosmetic.

Sesquipedalian:
Okay, @Antes. That looks nice, too. What is the HTML and CSS for that approach?

XinYenFon:
Zero CSS changes. Very little HTML changes (tho I cleaned up HTML bit so it might not look that very little lol).

https://github.com/SimpleMachines/SMF2.1/compare/release-2.1...Antes:search_ui?expand=1 not sure if you can see this diff but (too lazy to create a patch for you).

Sesquipedalian:
That's tidier HTML, yes. The use of the empty `<dt>` is kind of funky, though. Here's an idea: what if you moved `<label><strong>', $txt['mlist_search_filter'], ':</strong></label>` into the `<dt>` and then nested the list of options inside the `<dd>`? I think that would be more semantic and more visually straightforward.

Sesquipedalian:
Here's what my revised code looks like. As you can see when you look at the diff, @Antes, I incorporated elements of your changes, too.

screen shot 2017-01-04 at 4 46 25 pm

screen shot 2017-01-04 at 4 45 58 pm

XinYenFon (approved):
Even better big thanks :)

(dayum autocorrect to Turkish :P)

colinschoen (approved):
LGTM, approval

PR #3795, opened 2017-01-04
Uses CSS3 syntax for pseudo-element selectors

### LINKED ISSUE #3791: Should use CSS3 pseudo-element syntax
Since IE9 is the lowest IE that SMF 2.1 will support, we can upgrade our CSS to use the CSS3 double-colon (::) syntax for pseudo-elements. Microsoft themselves [recommend this](https://msdn.microsoft.com/en-us/library/hh781508(v=vs.85).aspx#pseudo-elements) for IE9+.

Sesquipedalian:
Fixes #3791

### DIFF
--- a/Themes/default/css/calendar.css
+++ b/Themes/default/css/calendar.css
@@ -297,7 +297,7 @@ span.hidelink {
 	#main_grid .calendar_table .days:not(.disabled) {
 		min-height: 45px;
 	}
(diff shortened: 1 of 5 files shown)


XinYenFon (approved):

colinschoen (approved):
LGTM, approval

live627 (approved):

PR #3796, opened 2017-01-04
Html bbc fix

### LINKED ISSUE #3794: HTML BBC fails
Type of error: General
2: strtr() expects parameter 1 to be string, array given
https://test2.turtleshellprod.com/index.php?action=post2;start=0;board=4
/.../Sources/Subs.php
Line: 823

tried the following post
testing
[html]HTML STUFF[/html]
and the entire content is removed...
this happens in both preview and post

Sesquipedalian:
Fixes #3794
Also, wraps the content of the [html] BBC in a `<div>`, since, after all, that BBC is supposed to be a block level one.


colinschoen (approved):
LGTM, approval

PR #3798, opened 2017-01-05
Image proxy should serve images even if they aren't cached

live627:


colinschoen (approved):
LGTM, approval. 

Side note: This would be a good class to use the new autoloader functionality to load  require_once($sourcedir . '/Class-CurlFetchWeb.php');

jdarwood007:
If you don't issue an exit after the header then it will continue to process the script.  Also should use redirectexit.

colinschoen:
Sorry missed that. Or perhaps just an `else` surrounding the rest of the remaining logic.

PR #3802, opened 2017-01-05
Adds a list view to the calendar

Sesquipedalian:
Think of this PR as an addendum to #3530. It adds a list view to the calendar, so that the user can view the calendar in list, month, or week formats. [Live demo](http://stovell.noip.me/~jon/dev/index.php?action=calendar;viewlist;year=2017;month=1;day=5)

The list view is somewhat similar to the compact "Upcoming Calendar" block in the info centre on the board index, but it gives more fulsome info and lets the user specify date ranges that they want to view.

The main changes are:

- A couple new functions to gather and assemble calendar info for display on the list view
- A new UI screen to show the list view
- The date selection dropdown and post event button have been moved from the bottom to the top (just below the title bar), and buttons to switch between list, month, and week views have been added
- To allow the user to switch between list, month, and week views, the calendar action now handles `viewlist`, `viewmonth`, and `viewweek` URL parameters. (Previously, it only handled the presence or absence of `viewweek`.) If none of these are specified, the default is list view (since, IMHO, the list view is more helpful for the sort of things a forum calendar is used for).

### DIFF
--- a/Sources/Calendar.php
+++ b/Sources/Calendar.php
@@ -112,7 +112,7 @@ function CalendarMain()
 		$context['calendar_view'] = $modSettings['calendar_default_view'];
 
 	// Don't let search engines index the non-default calendar pages
-	if ($context$context['calendar_view'] !== $modSettings['calendar_default_view'])
+	if ($context['calendar_view'] !== $modSettings['calendar_default_view'])
 		$context['robot_no_index'] = true;
 
 	// Get the current day of month...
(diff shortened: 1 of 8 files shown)


Sesquipedalian:
List view
screen shot 2017-01-05 at 1 22 01 pm

Month view
screen shot 2017-01-05 at 1 22 59 pm

Week view
screen shot 2017-01-05 at 1 31 40 pm

Admin option
screen shot 2017-01-05 at 2 21 46 pm

live627:
Automatic PHP linting is the best thing I added to this repo.

PR #3803, opened 2017-01-05
Database error "not exist: real_name_high"

tdausner:
Error message on Member list, default sort, last page:

http://my-forum.de/index.php?action=mlist;sort=real_name;start=13980
The database value you're trying to insert does not exist: real_name_high
Function: MLAll File: [smfRootFolder]/Sources/Memberlist.php
Line: 382

Run on smf V2.0.12

### DIFF
--- a/Sources/Memberlist.php
+++ b/Sources/Memberlist.php
@@ -305,7 +305,7 @@ function MLAll()
 	if ($use_cache && $_REQUEST['sort'] === 'real_name' && !isset($_REQUEST['desc']))
 	{
 		$first_offset = $_REQUEST['start'] - ($_REQUEST['start'] % $cache_step_size);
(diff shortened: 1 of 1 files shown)


live627 (changes requested):
Please try again using the latest files

https://github.com/SimpleMachines/SMF2.1/archive/release-2.1.zip

Oldiesmann:
A couple of things...

First off we can't accept this commit unless you sign off on it (simply adding "Signed-off-by: you <you@email.tld>" to the description is enough)
Also the code you're patching exists in two different places - it's also on line 326
Lastly, keep in mind that this site is for 2.1 not 2.0.x (although the issue you're reporting appears to be a problem in 2.1 as well).

albertlast:
in my eyes could this possible be a wrong solution,
when do a floor the last members could be missed.
Safer way would be to change the where statement instead of use between 
would be a >= <= and
drop the <= when you on the end of the list.
another question is,
if the upper ever is needed or only a mem.real_name >= {string:real_name_low} is enough.

Issue #3737, opened 2016-12-23
Inconsistency in folder navigation protections


sbulen:
The subfolders under /smf utilize different strategies to disallow direct navigation.  Some folders utilize .htaccess (which will work only on apache), and some utilize an index.php that will redirect the user to the home page (or just up one level).  Some folders use both.  

I.e., you use your browser & navigate directly to smf/attachments, you will get a 403 error.  
If you navigate to smf/themes you get redirected to the home page.
If you navigate to folder with neither an .htaccess nor an index.php, you start navigating files & folders with your browser.

If you read the actual .htaccess files, there are a lot of inconsistencies across them.  Some appear to be more deep than others & some are more minimal.   
If you read the index.php files, there are inconsistencies there too, but I'm not sure how meaningful these differences are.   

I suggest use of one consistent method for *ALL* folders and *ALL* subfolders.  

This is likely a very low priority, but I'm documenting it here for awareness/discussion.

sbulen:
In 2.0, I've noticed that many folders placed out there by mods do not have an .htaccess or an index.php.  You can navigate directly there and view and execute the source code files.   (e.g., Sources/LikePosts/DB)

Should similar folder treatment be required by mods?   

I wonder if we need a "safety net" utility to propagate the .htaccess/index.php files.  Something on the maintenance menu?

sbulen:
In a 2.1 vanilla install, in the Sources/minify folders, there are a few that don't have an index.php or .htaccess.

sbulen:
Closed

PR #3804, opened 2017-01-06
Folder protections consistent

sbulen:
Pretty sure this is all that's needed to resolve Inconsistency in folder navigation protections #3737.
Tests OK.


sbulen:
Sorry, still struggling a bit with git...

live627:
git is such a git

Issue #3813, opened 2017-01-07
Upgrader Step 2 - Cannot find database source script! Please check file Subs-Db-mysqli.php is within your forum source directory.


frandominguezl:
http://www.simplemachines.org/community/index.php?topic=551149

Something related to #3801 ?

tinoest:
Yes, seems that install.php and upgrade.php haven't been fully cleaned up / switched 

#3761 should either contain it or it should be done after that is merged to save people duplicating work.

illori:
running via CLI  i get this
Warning: mysqli_get_server_info() expects parameter 1 to be mysqli, null given in D:\wamp\www\github2.1\upgrade.php(2079) : eval()'d code on line 1

Call Stack:
    0.0180    1403544   1. {main}() D:\wamp\www\github2.1\upgrade.php:0
    0.1800    1882616   2. cmdStep0() D:\wamp\www\github2.1\upgrade.php:303
    0.1810    1866560   3. db_version_check() D:\wamp\www\github2.1\upgrade.php:2958
    0.1810    1868672   4. eval('global $db_connection; return min(mysqli_get_server_info($db_connection), mysqli_get_client_info());') D:\wamp\www\github2.1\upgrade.php:2079
    0.1810    1868872   5. mysqli_get_server_info() D:\wamp\www\github2.1\upgrade.php(2079) : eval()'d code:1

running in browser i get 
Fatal error: Call to undefined function safe_serialize() in D:\wamp\www\github2.1\upgrade.php on line 399
Call Stack
#	Time	Memory	Function	Location
1	0.0190	1420888	{main}( )	..\upgrade.php:0
2	0.1070	1958920	upgradeExit( )	..\upgrade.php:336

live627:
@illori #3761 should fix that

illori:
now a new error 
Fatal error: Call to undefined function throw_error() in D:\wamp\www\github2.1\upgrade.php on line 481

Call Stack:
    0.0120    1162920   1. {main}() D:\wamp\www\github2.1\upgrade.php:0
    0.0360    1182528   2. loadEssentialData() D:\wamp\www\github2.1\upgrade.php:154

live627:
@illori can you apply https://github.com/live627/SMF2.1/commit/fbd4e0a034f8343f772f8d230f44df87898f4cac?

illori:
that gives me via cli 
```
No direct access...
```

nothing else but the above.

illori:
[...]

PR #3801, opened 2017-01-05
replace mysql with mysqli and delete mysqli file fix #3774

albertlast:
This pr based on the issue #3774 

@Yoshi2889 well you should than revoke you changes from https://github.com/SimpleMachines/SMF2.1/pull/3761/commits/49d6fe4729f7f38863b2c5d3f485c64547cc8991 sry

PR #3817, opened 2017-01-07
Removing traces off MySQLi from #3801, related to #3813

tinoest:
Signed-Off-By: tinoest tinoest@gmail.com

### DIFF
--- a/other/install.php
+++ b/other/install.php
@@ -27,7 +27,7 @@ require_once('Sources/Class-Package.php');
 
 // Database info.
 $databases = array(
-	'mysqli' => array(
+	'mysql' => array(
 		'name' => 'MySQLi',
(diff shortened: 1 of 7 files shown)


other/install.php
```diff
@@ -27,7 +27,7 @@
 
 // Database info.
 $databases = array(
-	'mysqli' => array(
+	'mysql' => array(
 		'name' => 'MySQLi',
```
live627: Should the database name also be changed?

tinoest: It should, I need to spin up a vm with MySQL on and test this. I'll get it done some time today and let you know when it's ready to commit to the upstream.

tinoest:
Tested this and it works ok on PHP 5.6.29 and MariaDB 10.0.28

PR #3821, opened 2017-01-09
Change og_image to URL

jsDotx3:
Because og_image is a URL.
Signed-off-by: José <jsdotx3@gmail.com>


live627:
Note to self: this depends on  #3822

Issue #3772, opened 2017-01-02
Attachment size issues - minor but annoying


sbulen:
Two issues, minor but annoying.  Trying to attach photos to a message, I see:

(1)  What is displayed as a maximum is incorrect.  If an image is rejected, the message displayed by SMF says I have surpassed a 2K limit, not a 2M limit.   My setting in Attachment Settings for max attachment is 2000 KB.   

(2) Attachment size is calculated differently vs displayed.  This is just the normal 1000 vs 1024 issue.  I have a 1.9M file that is rejected on a 2M limit.  The file size is actually 1.91 MB (2,009,563 bytes).  SMF says it is a 1.9MB file...

So... SMF says I have a 2K limit when I have a 2M limit, and shows me it is rejecting a 1.9MB file when it tells me it is rejecting this limit.  It should be displaying a 2M limit, and that it is rejecting a 2M+ file...   

_I'd also look at avatars, I'm assuming they behave similarly._

frandominguezl:
1) Nice catch!

2) Yeah, seems like we're working with MiB (1024) instead of MB (1000), but we're confusing them both... Possible solutions are to change all limits from 1024 to 1000, or to change all places were MB is used to MiB...

sbulen:
Screenshots attached...  Forgot to attach them earlier....
![attach_rejected_size](https://cloud.githubusercontent.com/assets/23568484/21594553/bde356e4-d0d8-11e6-94ab-6d12f249c1db.JPG)
![attach_size](https://cloud.githubusercontent.com/assets/23568484/21594552/bde31cce-d0d8-11e6-8728-26a0aec62f04.JPG)

XinYenFon:
Try to increase max attachment size per post as well, I think that's a bug related to "0" value.

sbulen:
D3vcho's fix addressed the first issue above, where it stated something flat out inaccurate.  Thanks!   

The 2nd part is still outstanding.  It now says "max attachment size is 2000 KB", and still displays 1.9MB as the file size as pictured above.  

Max attachment size doesn't change this behavior.  

It would be more friendly & understandable if it displayed the attachment size in K also.   

FYI, it's not using true K (1024), it's using 1000's.  This file's size is 2,009,563 bytes, or 1.91 MB, or 1962 K.  

To be most clear it should display 2009 K as this pic's size, which clearly exceeds the 2000 K limit set by the admin.

sbulen:
[...]

PR #3824, opened 2017-01-10
Attachment size issues - minor but annoying #3772

sbulen:
Signed by Shawn Bulen, bulens@pacbell.net


sbulen:
Maybe I'm overdoing it?

I'm signing my signed signatures with a signature...

PR #3828, opened 2017-01-12
Fixed option "show shildren"

jsDotx3:
Option "show shildren", It does not work because operation logic OR (||), and I change OR (||) to AND (&&).

Topic: http://www.simplemachines.org/community/index.php?topic=551289.0

Signed-off-by: José jsdotx3@gmail.com

### DIFF
--- a/Themes/default/MessageIndex.template.php
+++ b/Themes/default/MessageIndex.template.php
@@ -26,7 +26,7 @@ function template_main()
 			</div>';
 	}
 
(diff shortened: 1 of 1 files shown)


Oldiesmann:
This isn't a bug. The option allows you to show sub-boards on every page of topics within the parent board. Your change causes SMF to **only** show sub-boards on the first page of topics within a board, and then only when that box is checked (with your change, if that box isn't checked, sub-boards aren't shown within the board at all).

jsDotx3:
Yes, but the option not is visible when change. Review the topic, have image and explication.

PR #3830, opened 2017-01-13
Resp menu

### LINKED ISSUE #3765: Mobile menu scrolling issue
found this while bug hunting

http://www.simplemachines.org/community/index.php?topic=549457.msg3896243#msg3896243

I have no way to test

From @sicommnend 
> You have to set the overflow to hidden for the HTML document when you set the overlay. This prevents the document underneath from scrolling while the lights are dim.

### LINKED ISSUE #3786: main menu oddity
Shrink the windows down to trigger the mobile view...

the pop-up menu icon appears to the left **AND** the main menu remains...

Honestly, I think that the main menu should convert to icons only at tablet size... and then to pop-up menu only at phone size... either way, however, the popup and regular menu should not be on the screen at the same time

menu

XinYenFon:
Test this PR with the following issues: fixes #3765 & fixes #3786 (cc: @Kindred-999)

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -2641,8 +2641,7 @@ dl.addrules dt.floatleft {
 	box-shadow: 0 3px 6px rgba(0,0,0,0.5);
 	border: 1px solid #777;
 	border-radius: 7px 7px 3px 3px;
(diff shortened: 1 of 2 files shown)


frandominguezl:
It fixes #3786 but not #3765

PR #3835, opened 2017-01-13
Added db_last_error.php to install check, corrected auto load of jQuery

### LINKED ISSUE #3114: Install database error for missing db_last_error.php
Warning: file_put_contents(/smf/db_last_error.php): failed to open stream: Permission denied in /smf/Settings.php on line 195
1. We should include this file (empty) by default in our installer.
2. We should make sure on install/upgrade that this file exists and is writable.

tinoest:
Fixes #3114 

### DIFF
--- a/Sources/ManageMaintenance.php
+++ b/Sources/ManageMaintenance.php
@@ -1725,7 +1725,8 @@ function MaintainRecountPosts()
 	// Lets get a group of members and determine their post count (from the boards that have post count enabled of course).
 	$request = $smcFunc['db_query']('', '
 		SELECT /*!40001 SQL_NO_CACHE */ m.id_member, COUNT(m.id_member) AS posts
-		FROM {db_prefix}messages AS m, {db_prefix}boards AS b
+		FROM {db_prefix}messages AS m 
+		INNER JOIN {db_prefix}boards AS b ON m.id_board = b.id_board
 		WHERE m.id_member != {int:zero}
 			AND b.count_posts = {int:zero}
 			AND m.id_board = b.id_board ' . (!empty($modSettings['recycle_enable']) ? '
@@ -1779,7 +1780,8 @@ function MaintainRecountPosts()
 			PRIMARY KEY (id_member)
 		)
 		SELECT m.id_member
-		FROM {db_prefix}messages AS m,{db_prefix}boards AS b
+		FROM {db_prefix}messages AS m
+		INNER JOIN {db_prefix}boards AS b ON m.id_board = b.id_board
 		WHERE m.id_member != {int:zero}
 			AND b.count_posts = {int:zero}
 			AND m.id_board = b.id_board ' . (!empty($modSettings['recycle_enable']) ? '
--- a/Sources/ManageMaintenance.php
+++ b/Sources/ManageMaintenance.php
@@ -1725,7 +1725,7 @@ function MaintainRecountPosts()
 	// Lets get a group of members and determine their post count (from the boards that have post count enabled of course).
 	$request = $smcFunc['db_query']('', '
 		SELECT /*!40001 SQL_NO_CACHE */ m.id_member, COUNT(m.id_member) AS posts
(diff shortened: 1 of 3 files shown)


Oldiesmann:
Don't we need to make other changes to support jQuery 3.1.1?

tinoest:
Not that I can see, the rest were done which I could find in my previous pull request #3807 which has already been merged.

Sources/ManageMaintenance.php
```diff
@@ -1725,7 +1725,7 @@ function MaintainRecountPosts()
 	// Lets get a group of members and determine their post count (from the boards that have post count enabled of course).
 	$request = $smcFunc['db_query']('', '
 		SELECT /*!40001 SQL_NO_CACHE */ m.id_member, COUNT(m.id_member) AS posts
-		FROM ({db_prefix}messages AS m, {db_prefix}boards AS b)
+		FROM {db_prefix}messages AS m, {db_prefix}boards AS b
```
jdarwood007: Wait when was this done?

We don't use multiple from statements.  We use joins to bring in additional tables to make it clear what we are doing.  http://stackoverflow.com/questions/894490/sql-left-join-vs-multiple-tables-on-from-line

tinoest:
No idea, I just corrected it so it ran on PG aswell as MySQL.

jdarwood007:
https://github.com/SimpleMachines/SMF2.1/commit/f02d980afcf5a2526a28bb3a9838c029ceb1b4fd#diff-ca7ff2a34b39d18aa0a1659d91368b9bR1827

Looks like Spuds.  Can we get this fixed up the proper way and use a join instead?

Oldiesmann:
While we're on the subject of cleaning up queries, we should also clean up the conditional comments, since the only ones we use are for MySQL 4.x, which we no longer support (going by what we strip out of queries for PostgreSQL).

tinoest:
@Oldiesmann did you want me to do that in this Pull Request also? As it's turned into a more a of a general fixes one.

Sources/ManageMaintenance.php
```diff
@@ -1725,7 +1725,8 @@ function MaintainRecountPosts()
 	// Lets get a group of members and determine their post count (from the boards that have post count enabled of course).
 	$request = $smcFunc['db_query']('', '
 		SELECT /*!40001 SQL_NO_CACHE */ m.id_member, COUNT(m.id_member) AS posts
-		FROM ({db_prefix}messages AS m, {db_prefix}boards AS b)
+		FROM {db_prefix}messages AS m 
+		INNER JOIN {db_prefix}boards AS b ON m.id_board = b.id_board
 		WHERE m.id_member != {int:zero}
 			AND b.count_posts = {int:zero}
 			AND m.id_board = b.id_board ' . (!empty($modSettings['recycle_enable']) ? '
```
albertlast: please remove here and in the other query the join statement  "AND m.id_board = b.id_board"
your already define the join statement in the ON part.

tinoest: I left it in due to the !empty part, its already joined so I didn't see any performance impact.

Sources/ManageMaintenance.php
```diff
@@ -1725,7 +1725,8 @@ function MaintainRecountPosts()
 	// Lets get a group of members and determine their post count (from the boards that have post count enabled of course).
 	$request = $smcFunc['db_query']('', '
 		SELECT /*!40001 SQL_NO_CACHE */ m.id_member, COUNT(m.id_member) AS posts
-		FROM ({db_prefix}messages AS m, {db_prefix}boards AS b)
+		FROM {db_prefix}messages AS m 
+		INNER JOIN {db_prefix}boards AS b ON m.id_board = b.id_board
```
Oldiesmann: The INNER JOIN clause should be indented here

Sources/ManageMaintenance.php
Oldiesmann: I agree with @albertlast  here. No reason to leave this part in if we're doing an INNER JOIN.

PR #3837, opened 2017-01-13
optimize DB initial check fixes #3834

### LINKED ISSUE #3834: "postg_fetch_assoc"
Subs-Db-postgrespl.php, line 37:

```
if (!isset($smcFunc['db_fetch_assoc']) || $smcFunc['db_fetch_assoc'] != 'postg_fetch_assoc')
```

Is "postg_fetch_assoc" really what should be in there? I would think "smf_db_fetch_assoc" would be right, based on what I see just a few lines below that.

albertlast:
I think an easier check is a better solution for #3834

### DIFF
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -33,7 +33,7 @@ function smf_db_initiate($db_server, $db_name, $db_user, $db_passwd, $db_prefix,
 	global $smcFunc, $mysql_set_mode;
 
 	// Map some database specific functions, only do this once.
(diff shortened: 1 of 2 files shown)


live627 (changes requested):
`smf_db_initiate()` can be called more than once in the install and upgrade scripts.

albertlast:
Sure that is the reason why i keep the first express.

In the existing solution are two express combined with or:
The first part which check if the db is already initalize: !isset($smcFunc['db_fetch_assoc'])
and the second part which check if this the right one: $smcFunc['db_fetch_assoc'] != 'mysqli_fetch_assoc'

in my eyes the second check is unneeded.

live627 (approved):

PR #3839, opened 2017-01-14
Better Coding Style for curl class

albertlast:
Try to remove this findings: https://scrutinizer-ci.com/g/SimpleMachines/SMF2.1/issues/release-2.1/files/Sources/Class-CurlFetchWeb.php?orderField=path&order=asc&fileId=Sources%2FClass-CurlFetchWeb.php&honorSelectedPaths=0

in my eyes the curl class is broken,
three types of options
default_options, user_options and options
but to fix this is not target of this pr.

### DIFF
--- a/Sources/Class-CurlFetchWeb.php
+++ b/Sources/Class-CurlFetchWeb.php
@@ -57,6 +57,41 @@ class curl_fetch_web_data
 		CURLOPT_SSL_VERIFYHOST	=> 0, // stop cURL from verifying the peer's host
 		CURLOPT_POST			=> 0, // no post data unless its passed
 	);
+	
+	/**
+	 * @var int Maximum number of redirects
+	 */
+	public $max_redirect;
+	
+	/**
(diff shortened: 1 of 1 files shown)


Oldiesmann:
You can ignore the messages about "Property `property` does not exist". For some reason Scrutinizer thinks we're trying to access the values of those properties rather than setting them. Defining properties inside methods (especially in the constructor) is standard practice in many programming languages when working with OOP. Declaring all the variables before hand won't do anything because we're not actualy doing anything with them.

albertlast:
Nope Scrutinizer is here right.
Decleration of a attribut had to be done in the class head.
The initialize of this attribut can be done somewhere else or
when the scrope of this is only to be exists in methode than is this fine too.
But this var are used across the class so they had to be define correctly.

And this you can't do for exp in java because that is stupid.

live627 (approved):
@Oldiesmann this is good coding practice, and ends up in less dev confusion.

Issue #3833, opened 2017-01-13
Upgrade Issue - Can't get past Step 4


sbulen:
I can't get past Step 4.  It keeps telling me there is no response for 30 seconds, but my system seems to be OK otherwise.  Logs are fairly clean.  

This is my first test run on a complete copy of my prod system - with all the prod data - moving from 2.0 to 2.1.  Running PHP 5.6, & MySql 5.7.  

I suspect it is confused what step it is on.  Note in the URL it says it's on **_step 2_**, but it was actually on **_step 4_**.  

Maybe that's the real issue, it's lost.  See screenshot.  

![cantgetpaststep4](https://cloud.githubusercontent.com/assets/23568484/21918320/fd265a98-d904-11e6-86f3-3ff777e838be.JPG)

sbulen:
Just installed a brand new 2.0 instance & upgraded to 2.1 and this did NOT occur.  It executed properly to completion.

sbulen:
Just repeated the upgrade process for the entire site (yes, in test) and the same thing happened.  I can't get past step 4.   Running the very latest off of github.  

EXACT same location - after saying "Add an order value to each existing cust profile field... done", it clearly pauses for a while & issues the error message above.

sbulen:
However, this time I see this entry in my MySQL log:
2017-01-14T17:33:58.960641Z 52 [Note] Aborted connection 52 to db: 'unconnected' user: <my user> host: <myhost> (Got an error reading communication packets)

Reading up on this...  Lots of sources of this error, among them network hardware errors (not likely, it's localhost...), lack of commits, etc., settings too low in MySql config, or PHP unexpected termination.

sbulen:
I suspect an issue in the Messenger Fields migration.   

If I delete the "Add an order value..." and "Converting member values" sections from upgrade_2-1_mysql.sql (rows 957-1062) it executes successfully to completion.

jdarwood007:
Try in the "Converting member values" to change the $limit to something lower like 200.

sbulen:
Hmmm...   Didn't work.  I have it all the way down to 200.  

I've got ~18000 members.  It just seems to run away with that volume on that command that tries to take the old columnar data & convert it.

sbulen:
Looks to me like its job is to take the old messenger contents from the members table and convert them to references in the themes table.  

It didn't add a single row into the themes table after all that effort.

sbulen:
There are two issues here...
[...]

PR #3842, opened 2017-01-15
Fix upgrade migration of fields from members to themes

sbulen:
To address: Upgrade Issue - Can't get past Step 4 #3833

### DIFF
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -1022,22 +1022,22 @@ if (!empty($select_columns))
 		while ($row = $smcFunc['db_fetch_assoc']($request))
 		{
 			if (!empty($row['aim']))
-				$inserts[] = array($row['id_member'], -1, 'cust_aolins', $row['aim']);
+				$inserts[] = array($row['id_member'], 0, 'cust_aolins', $row['aim']);
 
 			if (!empty($row['icq']))
-				$inserts[] = array($row['id_member'], -1, 'cust_icq', $row['icq']);
+				$inserts[] = array($row['id_member'], 0, 'cust_icq', $row['icq']);
 
 			if (!empty($row['msn']))
-				$inserts[] = array($row['id_member'], -1, 'cust_skyp', $row['msn']);
+				$inserts[] = array($row['id_member'], 0, 'cust_skyp', $row['msn']);
 
 			if (!empty($row['yim']))
-				$inserts[] = array($row['id_member'], -1, 'cust_yim', $row['yim']);
+				$inserts[] = array($row['id_member'], 0, 'cust_yim', $row['yim']);
 
 			if (!empty($row['location']))
-				$inserts[] = array($row['id_member'], -1, 'cust_loca', $row['location']);
+				$inserts[] = array($row['id_member'], 0, 'cust_loca', $row['location']);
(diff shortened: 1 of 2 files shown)


other/upgrade_2-1_mysql.sql
```diff
@@ -1046,7 +1046,7 @@ if (!empty($select_columns))
 				'{db_prefix}themes',
 				array('id_member' => 'int', 'id_theme' => 'int', 'variable' => 'string', 'value' => 'string'),
 				$inserts,
-				array('id_theme', 'id_member', 'variable')
+				array('id_member', 'id_theme', 'variable', 'value')
```
albertlast: value field can't be a pk field.
and the pk is defined as 'id_theme', 'id_member', 'variable' -> i would keep the old arrengment

other/upgrade_2-1_mysql.sql
```diff
@@ -1022,22 +1022,22 @@ if (!empty($select_columns))
 		while ($row = $smcFunc['db_fetch_assoc']($request))
 		{
 			if (!empty($row['aim']))
-				$inserts[] = array($row['id_member'], -1, 'cust_aolins', $row['aim']);
+				$inserts[] = array($row['id_member'], 0, 'cust_aolins', $row['aim']);
```
albertlast: -1 looks as the right default value (the datebase can handle -1)
the same for the other changes.

albertlast (changes requested):

albertlast:
Please consider that all mysql changes needs to be sync with pg upgrade when ther are right,
what in my eyes here not the true is.

other/upgrade_2-1_mysql.sql
albertlast: i looked on the wrong col,
the -1 is in theme_id col...
here are -1 value not allowed -> 0 or even 1 could be right

other/upgrade_2-1_mysql.sql
sbulen: OK.  Will revise & test.

sbulen:
Added the pg updates.  I don't have pg, so I can't test them, but they're simple enough. 

I left the pk line alone.  That makes sense.  Thanks for the insight.  

FYI - I have tested with a theme_id of 0 as well as a theme_id of 1 here.  The theme_id of 0 I believe corresponds to the default, and is a little more consistent with what I read elsewhere in this code.  

When you use 0, the profiles of forum members are visible with this info throughout the forum, which I think is desired.

albertlast:
Someone else had to say if theme_id 0 is a valid value for this context,
i only mention the database side.

other/upgrade_2-1_mysql.sql
live627: https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Load.php#L1753

Probably should be `1`

sbulen: I think you're right...   Let me fix this...
I just added a bunch of these attributes manually online, & they all get added with a theme_id of 1.
I think the data migration should mimic the manual data entry, so I'll set it to 1.

sbulen:
Changed to theme_id of 1, per discussion above.  Thanks for the input!

Issue #3841, opened 2017-01-15
Upgrader - : unserialize(): Error at offset 0 of 15 bytes in ... upgrade.php on line 125


sbulen:
I've started seeing this at the top of the Upgrader recently:

Notice
: unserialize(): Error at offset 0 of 15 bytes in (path) \upgrade.php on line 125

albertlast:
In general would be better I think,
if we safe this information as json and not as serialze.

sbulen:
When re-testing the upgrade, I restored the DB but kept the settings.php file around.  (I thought this was safe, it had my environment info...)

The 2.1 upgrade adds a bunch of entries to the settings.php file.   These settings are causing the error.   If I restore the original settings.php file when I restore the DB, I do not get this error.  

These are the lines in Settings.php causing the issue:
> $upgradeData = #remove#;
> $image_proxy_secret = '47baf8c8b7f0fbc7fce8';

So...   Due to the comment #remove# (which also negates the semicolon), this ends up executing as:
> $upgradeData = $image_proxy_secret = '47baf8c8b7f0fbc7fce8';

If you attempt to re-use this Settings.php file, $upgradeData gets this weird value, and causes this error message when it attempts to deserialize this value.  

I was going to close this issue out, since there is no issue when I start with the 2.0 settings.php.  

But then I wondered -  Do we really want to leave these lines in Settings.php?   Including:
> $upgradeData = #remove#;

Or are we supposed to remove it as the comment suggests?

sbulen:
OK, to be safest, I just submitted a PR to set $upgradeData to null upon successful completion of an upgrade.  Tested & it works fine.  (If nothing else, we're setup for 3.0!)

PR #3849, opened 2017-01-17
Set $upgradeData to null after successful upgrade

sbulen:
To fix this issue:   Upgrader - : unserialize(): Error at offset 0 of 15 bytes in ... upgrade.php on line 125 #3841

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -1318,7 +1318,7 @@ function DeleteUpgrade()
 	$changes = array(
 		'language' => '\'' . (substr($language, -4) == '.lng' ? substr($language, 0, -4) : $language) . '\'',
 		'db_error_send' => '1',
-		'upgradeData' => '#remove#',
+		'upgradeData' => '\'\'',
 	);
 
 	// Are we in maintenance mode?


jdarwood007:
Its actually supposed to be removed (the entire line).  I wonder if something else broke elsewhere.

jdarwood007:
https://github.com/SimpleMachines/SMF2.1/commit/41ec30d7a54edf399c67ab21a1cf76ef51bef6b6#diff-cccb32785017365b9e98cf35be9ed3dfL1627

That commit there removed the code that handled this.

@live627

sbulen:
I could not find current code to remove the line, or I would have gone that route.

Setting it to null works, and gets rid of the error ("unserialize(): Error at offset 0 of 15 bytes in upgrade.php on line 125").

sbulen:
FYI, I'm pretty sure this is the only instance of #remove# in SMF.   If this were used elsewhere, we should definitely add that old code back.  But if it's not needed elsewhere, & null works, I'd say let's keep it simple.  

Make sense?

live627:
> 41ec30d#diff-cccb32785017365b9e98cf35be9ed3dfL1627
>
> That commit there removed the code that handled this.

That was when I wanted to recycle that function. Didn't notice it removed stuff.

sbulen:
FWIW, I really think we should not have put the #remove# in there, even temporarily, due to this syntax error.  

It causes different issues for different folks, depending on the contents of their settings.php file.

PR #3850, opened 2017-01-17
Improves handling of attachment MIME types

Sesquipedalian:
Since SMF 2.1 requires PHP 5.3 as a minimum, we can use mime_content_type() to better identify an attachment's MIME type.

Also, a MIME type string can be up to 128 characters long, so this updates the length of the mime_types column in the attachments table to accommodate that.

### DIFF
--- a/Sources/Attachments.php
+++ b/Sources/Attachments.php
@@ -254,6 +254,11 @@ class Attachments
 			// No errors, YAY!
 			if (empty($errors))
 			{
+				// The reported MIME type of the attachment might not be reliable.
+				// Fortunately, PHP 5.3+ lets us easily verify the real MIME type.
+				if (function_exists('mime_content_type'))
+					$_FILES['attachment']['type'][$n] = mime_content_type($_FILES['attachment']['tmp_name'][$n]);
+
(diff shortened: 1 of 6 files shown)


live627:
Dumb question: Does `mime_content_type()` read the file like `exif_imagetype ()` does?

Also, according to the manual, it's in PHP 4 http://php.net/manual/en/function.mime-content-type.php

Sesquipedalian:
> Dumb question: Does `mime_content_type()` read the file like `exif_imagetype ()` does?

I think it uses a variety of tricks depending on the file being examined. The manual refers interested parties to the `magic.mime` file.

> Also, according to the manual, it's in PHP 4

Yes, but its parent extension only became enabled by default in 5.3. http://php.net/manual/en/fileinfo.installation.php

Sesquipedalian:
It's probably worth noting that all of our other checks to determine the correct MIME type for image files, etc., are still in place. This is just adding another layer so that (1) we don't need to trust the user's browser to give us a correct initial MIME type, and (2) we won't end up with a blank MIME type even if the browser doesn't tell us one.

PR #3852, opened 2017-01-18
RSS improvements

Sesquipedalian:
This PR does three main things:

1. Makes feeds much more mod friendly. 
    - Abstracts the process for adding attributes to XML elements in feeds.
    - Allows mods to extend a feed's XML by adding extra elements
    - Allows mods to force CDATA wrapping or namespace prefixes for certain elements
2. Allows attachments to be included as enclosures in feeds
    - Can be enabled/disabled by the admin
    - Respects `view_attachments` permission
    - Makes mods for podcasting, broadcatching, etc., possible
3. Fixes some unreported bugs with Atom feeds (we still weren't generating valid ones in some cases).


Sesquipedalian:
Okay, this is done. I've tested every combination of sub-action and type, and the expected data appears in the expected format in every case. All standard feed types were tested with the W3C's [feed validator](https://validator.w3.org/feed/), and SMF's proprietary feeds were checked using `xmllint` as well as visually inspected and compared with the output of prior versions of SMF.

Sesquipedalian:
Bleh. That's a lot of commits. I'm going to squash them and resubmit this PR.

Issue #3825, opened 2017-01-11
Attachments - Max attachments not honored


sbulen:
Set it to 4, it'll let you upload 100...  

Setting is "Max number of attachments per post"

Kindred-999:
Confirmed with both admin and normal user...

PR #3855, opened 2017-01-18
Fixes maxfiles check in dropzone

sbulen:
Attachments - Max attachments not honored #3825

Default dz formatting...   Ugly but functional.

### DIFF
--- a/Themes/default/scripts/smf_fileUpload.js
+++ b/Themes/default/scripts/smf_fileUpload.js
@@ -1,8 +1,12 @@
 function smf_fileUpload(oOptions)
 {
-	// Before anything else, overwrite some stuff...
+	// Check if the file should be accepted or not...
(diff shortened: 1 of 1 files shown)


sbulen:
In the original dropzone.js script, the prototype.accept function does all the error checking, including maxfiles.   

In smf_fileUpload.js, the prototype.accept function had been overwritten to simply return OK.   The idea being that there were new versions of all of the error logic...       

EXCEPT maxfiles...  Now we had no error checking for maxfiles.  

The solution was to put the maxfiles check back in, using the original dropzone.js as an example.  

The done function returns the error message.  The .emit function broadcasts the error to the rest of dropzone to act upon.  These two actions mimic dropzone's default response to exceeding maxfiles.  

Any questions, let me know,

PR #3856, opened 2017-01-18
RSS Improvements

Sesquipedalian:
1. Makes the internal logic better and smoother
	- Restructures data arrays to remove any need for kludges
	- Abstracts the process for adding attributes to XML elements in feeds
	- Cleans up, generalizes, and simplifies sundry parts of the code
2. Makes feeds much more mod friendly
	- Allows mods to extend a feed's XML by adding extra elements to both a feed and its entries
	- Allows mods to add/modify namespace declarations
	- Allows mods to force CDATA wrapping or namespace prefixes for certain elements
3. Allows attachments to be included as enclosures in feeds
	- Can be enabled/disabled by the admin
	- Respects `view_attachments` permission
	- Makes mods for podcasting, broadcatching, etc., possible
4. Fixes some unreported bugs with Atom feeds (we still weren't generating valid ones in some cases).


Sesquipedalian:
I believe this is done now. I've tested every combination of sub-action and type, and the expected data appears in the expected format in every case. All standard feed types were tested with the W3C's [feed validator](https://validator.w3.org/feed/), and SMF's proprietary feeds were checked using `xmllint` as well as visually inspected and compared with the output of prior versions of SMF. No errors appear in the error log.

live627:
so much for squashing commits (I didn't think that was necessary though)

Sesquipedalian:
> so much for squashing commits

Oh well. It seemed like a good idea at the time. 😊

PR #3857, opened 2017-01-19
More RSS improvements

Sesquipedalian:
I found a few final improvements to make. Mostly just some bulletproofing.

- Fixes two small bugs in fix_possible_url()
	- Now fixes URLs anywhere in the string, not just at the start
	- Passes the correct delimiter character to preg_quote()
- Enforces consistent sanitization for all element attributes and values
- Ensures required feed elements are not empty
- Adds default support for icons and copyright notices to RSS and Atom feeds, because why not?
- Consolidates a few stray bits of redundant code


live627:
`Parse error: syntax error, unexpected ',' in ./Sources/News.php on line 363`

PR #3861, opened 2017-01-22
Error when modifying a post

ThomasFok:
When modifying a post, I got this error.

Only variables should be passed by reference
Sources/Drafts.php 
Line:468

The argument of censorText() is a reference. It causes error when the return value of a function is passed to it directly.


XinYenFon:
You have to sign-off each commit, otherwise we can't accept it.

ThomasFok:
Signed-off-by: Thomas Fok <thomas@melody-soft.com>

Oldiesmann:
You have to sign off on the commits themselves, not the PR

live627:
Sign added to squashed commit.

Issue #3827, opened 2017-01-11
Attachments - orphaned file issues


sbulen:
Several file management issues that I believe are related:

(1) Add a few attachments, & press the "Upload" button.   Navigate away without pressing "Post".  Start a 2nd post.  Start to add an attachment to your new post, & **one of** the prior attachments is already in your list...  I'm not sure what you want it to do here, but likely not that...    (Either all or none & clean up after itself...)

(2)  Set your max attachment directory space to something low like 4000K.   Add a bunch of attachments & exceed this.   Upload what you can, & press Post.   
    2A -  It will not log a critical error or notify the administrators that the directory is full.  
    2B -  It will leave post_tmp* files in your file system.  The only way to remove these files is at the file system level.  They are not visible when using "Browse Files" and they are not cleaned up when you run the Attachment Integrity Check.
    2C -  It displays the wrong errors to the user - it will use the 'internal' errors, not the user friendly translated errors.    E.g., the error will state "ran_out_of_space".   

(3)  After doing several tests like the above, check the # of files in your attachment folder.   There are too many [hash].dat files and post_tmp* files.   They don't show up when you "browse" files under Admin | Attachments | Browse.  And they don't go away after running file maintenance.   To verify, browse your attachments and delete ALL of them.   Run the Attachment Integrity Check.  You should have 0 files in the attachment folder now, but you have more...  

The ONLY way to delete these orphaned files is directly, at the file system level (file manager).  Go ahead & do so...   Interestingly, running another Attachment Integrity Check immediately after this will squawk about files it didn't know about a minute earlier...   

Either "Browse Files" is missing files or the Attachment Integrity Check is missing files...   

It appears the dropzone attachment UI leaves orphans around - it doesn't always clean up after itself.

sbulen:
My theory now on (3) above is that all of the [hash].dat files are valid, but browse & Attach Integrity Check are acting on old info, specifically, your previous visit to Admin | Attachments.

sbulen:
(3) Above is repeatable with the following steps:   Add a bunch of files (I always use images) to a post.   Upload.   Navigate away without pressing the 'Post' button.    If the upload was successful, these files exist on disk in [hash].dat format, but are not visible when you use Browse Files (Admin, Attachments & Avatars).  They do not go away when you do an attachment check.  They are, however, included in the stats under File Maintenance.  

So:
 - Files that error out leave behind the post_tmp* orphans.
[...]

PR #3862, opened 2017-01-22
Fix issues with attachment critical error processing

sbulen:
Addresses some of the issues noted here: Attachments - orphaned file issues #3827

Specifically, these issues are addressed:
(2) Set your max attachment directory space to something low like 4000K. Add a bunch of attachments & exceed this. Upload what you can, & press Post.
2A - It will not log a critical error or notify the administrators that the directory is full.
2B - It will leave post_tmp* files in your file system. The only way to remove these files is at the file system level. They are not visible when using "Browse Files" and they are not cleaned up when you run the Attachment Integrity Check.
2C - It displays the wrong errors to the user - it will use the 'internal' errors, not the user friendly translated errors. E.g., the error will state "ran_out_of_space".

### DIFF
--- a/Sources/Attachments.php
+++ b/Sources/Attachments.php
@@ -331,10 +331,11 @@ class Attachments
 				'mime_type' => isset($attachment['type']) ? $attachment['type'] : '',
 				'id_folder' => isset($attachment['id_folder']) ? $attachment['id_folder'] : $modSettings['currentAttachmentUploadDir'],
 				'approved' => !$modSettings['postmod_active'] || allowedTo('post_attachment'),
-				'errors' => $attachment['errors'],
+				'errors' => array(),
 			);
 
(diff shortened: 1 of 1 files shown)


sbulen:
The core issue here is the mismatched elseif...   The indentation is misleading.  The elseif is matched to an indented if.   So what happens is that the main "else" exception logic - the processing of errors - never executes.  

And...  Once the mismatched else was fixed...

The attachment errors structure has the "internal" errors in it, e.g., "ran_out_of_space", that should not be passed back to the users.   The exception logic steps thru the internal errors & produces user-friendly errors & admin notifications .  Thus, the output errors should start off as empty, or it will have 2 versions of the errors - internal & external.  This is what I saw in testing & fixed.  

I hope this makes sense.  Any questions, let me know.

PR #3864, opened 2017-01-24
Fix #3843

### LINKED ISSUE #3843: SCEditor incorrectly formats and interprets [url] BBCode
It is easier to give instructions on how to reproduce this one than it is to describe it. Do the following, and you will see the issue:

1. In WYSIWYG mode, insert a link to some URL. Initially, everything appears to the eye to be okay (but it isn't—see below on previewing, etc.).
screen shot 2017-01-15 at 4 33 17 am
2. Toggle from WYSIWYG mode to source mode. Instead of converting the `<a>` element to a `[url]` BBCode, it is converted to a `[member]` BBCode with an undefined parameter.
screen shot 2017-01-15 at 4 35 57 am
3. Toggle back to WYSIWYG. SCEditor tries to render this as a mention.
screen shot 2017-01-15 at 4 37 42 am
4. Toggle to source mode again. Now the `<a>` get converted to an `[attach]` BBCode!
screen shot 2017-01-15 at 4 39 09 am

This issue does not only affect toggling between editor modes. It also affects the actual data passed to SMF's preparser (and therefore also the data that will be saved in the database). To see this, just get a post preview at any point in the process outlined above. Try to get a preview after completing step 1 above, and you will see this:
screen shot 2017-01-15 at 4 47 23 am
Try after completing step 3, and you will see this:
screen shot 2017-01-15 at 4 49 08 am

ThomasFok:
Missing typeof operator seems to be the cause.

### DIFF
--- a/Themes/default/scripts/jquery.sceditor.smf.js
+++ b/Themes/default/scripts/jquery.sceditor.smf.js
@@ -437,17 +437,17 @@ $.sceditor.plugins.bbcode.bbcode.set(
 			if (url.substr(0, 7) === 'mailto:')
 				return '[email=' + url.substr(7) + ']' + content + '[/email]';
 
-			if (element.attr('target') !== undefined)
+			if (typeof element.attr('target') !== "undefined")
 				return '[url=\"' + decodeURI(url) + '\"]' + content + '[/url]';
 
 			// A mention?
-			else if (element.attr('data-mention') !== "undefined")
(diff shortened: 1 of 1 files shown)


jdarwood007:
Can you sign off your commit please?

live627:
uhh ,sleepy, is it not there??

jdarwood007:
Oh it is upon further review, GitHub didn't make it easy to see at first.  Just making sure as all commits need signed off.

PR #3865, opened 2017-01-24
Remove php close tag

albertlast:
Scrutinizer says:
> It is not recommended to use PHP's closing tag ?> in files other than templates.
>Using a closing tag in PHP files that only contain PHP code is not recommended as you might accidentally add whitespace after the closing tag which would then be output by PHP. This can cause severe problems, for example headers cannot be sent anymore.
>A simple precaution is to leave off the closing tag as it is not required, and it also has no negative effects whatsoever.

From the reason it's sound right to do so.

260 issues are fixed

### DIFF
--- a/Packages/backups/index.php
+++ b/Packages/backups/index.php
@@ -5,5 +5,3 @@ if (file_exists(dirname(dirname(__FILE__)) . '/index.php'))
 	include (dirname(dirname(__FILE__)) . '/index.php');
 else
 	exit;
-
-?>
--- a/Packages/index.php
+++ b/Packages/index.php
@@ -14,5 +14,3 @@ if (file_exists(dirname(dirname(__FILE__)) . '/Settings.php'))
 // Can't find it... just forget it.
 else
 	exit;
-
-?>
--- a/SSI.php
+++ b/SSI.php
@@ -2354,5 +2354,3 @@ function ssi_recentAttachments($num_attachments = 10, $attachment_ext = array(),
 	echo '
 		</table>';
 }
-
-?>
--- a/Smileys/aaron/index.php
+++ b/Smileys/aaron/index.php
@@ -5,5 +5,3 @@ if (file_exists(dirname(dirname(__FILE__)) . '/index.php'))
 	include (dirname(dirname(__FILE__)) . '/index.php');
 else
 	exit;
-
-?>
--- a/Smileys/akyhne/index.php
+++ b/Smileys/akyhne/index.php
@@ -5,5 +5,3 @@ if (file_exists(dirname(dirname(__FILE__)) . '/index.php'))
 	include (dirname(dirname(__FILE__)) . '/index.php');
 else
 	exit;
-
-?>
--- a/Smileys/default/index.php
+++ b/Smileys/default/index.php
@@ -5,5 +5,3 @@ if (file_exists(dirname(dirname(__FILE__)) . '/index.php'))
 	include (dirname(dirname(__FILE__)) . '/index.php');
 else
 	exit;
-
-?>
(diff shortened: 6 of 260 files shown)


jdarwood007:
Unfortunately for now we need to leave the closing ?>.  SMF's package manager has problems when doing the End of File replacement operations as it depends on finding ?> to undo the operation during a uninstall.

Right now there isn't a good solution other than adding something stupid other than /* EOL */ comment to the end.

live627:
> Right now there isn't a good solution other than adding something stupid other than /* EOL */ comment to the end.

Some mods specifically look for `?>`

jdarwood007:
Yes and those mods are doing it wrong.  We have a operation to handle the end of the file.  Breaking them with removing ?> is not in my scope of worries, it is the other stuff that breaks due to it.

albertlast:
Should I replace "?>" with "/* EOL */" ?
When this solution is working for the packege manager than i'm fine with that.

jdarwood007:
Package manager would have to be updated to handle looking for /* EOL */ instead of ?> as well.

It should be discussed and honestly not sure if we want to tackle this with 2.1 looking to get another beta release soon and then go RC.

Sesquipedalian:
In the long term, I agree with @albertlast, but I can see the point about not making this change yet. 

Maybe for now @jdarwood007 could just add a Travis check to make sure there is no whitespace trailing after `?>` in any of our PHP files.

jdarwood007:
Shouldn't be too difficult to write a script to do this.  Off the top of my head, I would need to find the last 100 bytes of the file and then look for with regex ~\?>([\s]+)$~i and check the result is not empty.

Sesquipedalian:
Here's a function I've used elsewhere that might be helpful with that:

```
function tailCustom($filepath, $lines = 1, $adaptive = true) {
	// Open file
	$f = @fopen($filepath, "rb");
	if ($f === false) return false;

	// Sets buffer size
	if (!$adaptive) $buffer = 4096;
	else $buffer = ($lines < 2 ? 64 : ($lines < 10 ? 512 : 4096));

	// Jump to last character
	fseek($f, -1, SEEK_END);

	// Read it and adjust line number if necessary
	// (Otherwise the result would be wrong if file doesn't end with a blank line)
	if (fread($f, 1) != PHP_EOL) $lines -= 1;

	// Start reading
	$output = '';
	$chunk = '';

	// While we would like more
	while (ftell($f) > 0 && $lines >= 0) {

		// Figure out how far back we should jump
		$seek = min(ftell($f), $buffer);

		// Do the jump (backwards, relative to where we are)
		fseek($f, -$seek, SEEK_CUR);

		// Read a chunk and prepend it to our output
		$output = ($chunk = fread($f, $seek)) . $output;

		// Jump back to where we started reading
		fseek($f, -mb_strlen($chunk, '8bit'), SEEK_CUR);

		// Decrease our line counter
		$lines -= substr_count($chunk, PHP_EOL);

	}

	// While we have too many lines
	// (Because of buffer size we might have read too many)
	while ($lines++ < 0) {

		// Find first newline and remove all text before that
		$output = substr($output, strpos($output, PHP_EOL) + 1);

	}

	// Close file and return
	fclose($f);
	return trim($output);
}
```

albertlast:
for end of line why we don't use http://php.net/manual/de/function.file-get-contents.php ?
With php 7.1.0 we could use the shortcut to jump directly to the end.

jdarwood007:
file_get_contents works, but when you think about it, you are loading a lot of data into memory and dumping it.  Was trying to be more efficient, but it can be done.

@Sesquipedalian Your welcome to add the commit for that.
My development process for the others ones was running the find:
find . -name "*.php" -exec php other/buildtools/check-smf-license.php {} 2>&1 \;

Then when I found a file it was picky about, investigate and run it directly via:
cd /root/of/smf/repo
php other/buildtools/check-smf-license.php ./Local/Path/To/File.php
I.e: php other/buildtools/check-smf-license.php ./Sources/Subs.php

After I made it happy with that file I moved on running the find again, till it came out clean (or fixed any issues it really found).

albertlast:
The files are not big enough to be a problem,
but i think it's not the right time for changing this part.

PR #3869, opened 2017-01-25
Ignores ./other/db_last_error.php in license check

Sesquipedalian:


jdarwood007:
My bad.  I totally forgot about that.
Should add as well:
````
	'\./cache/data_[A-Za-z0-9-_]\.php',
````
To ignore any cache files.

PR #3876, opened 2017-01-26
Clean up after old UTF-8 languages

Oldiesmann:
This commit focuses on cleaning up language things for forums that were converted to UTF-8 prior to upgrading to 2.1. Specifically, it does three things:
1. Update the default forum language (dropping "-utf8" from the name)
2. Update the list of installed languages (dropping "-utf8" from the internal name and " UTF-8" from the external name, while also accounting for the possibility that the "regular" version of at least one language is already installed)
3. Update the language for each user (dropping "-utf8" from the language name)

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -908,7 +908,7 @@ function checkLogin()
 // Step 1: Do the maintenance and backup.
 function UpgradeOptions()
 {
-	global $db_prefix, $command_line, $modSettings, $is_debug, $smcFunc, $packagesdir, $tasksdir;
+	global $db_prefix, $command_line, $modSettings, $is_debug, $smcFunc, $packagesdir, $tasksdir, $language;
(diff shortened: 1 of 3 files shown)


jdarwood007:
You will need to test to make sure that the langauge list isn't serialized still here.  Since that in a normal upgrade occurs after the database changes.

ThomasFok:
I got this error when I execute upgrade.php

image

I don't know much about SQL. But I guess line 2514 of upgrade_2-1_mysql.sql should be `--{`. There is same typo in upgrade_2-1_postgresql.sql

ThomasFok:
One more error.  

image

I guess the correct sql is 

```sql
UPDATE {$db_prefix}members
SET lngfile = REPLACE(lngfile, '-utf8', '');
```

Note: There are two underscore in `$db_prefix` and spelling of `members` is wrong.

Issue #3877, opened 2017-01-27
Upgrader:  Freezes, loses graphics, loses buttons, errors


sbulen:
Upgrader is behaving erratically today.  Brand new download from github.  Issues:
- it loses the 'continue' button
- it loses graphics
- it freezes
- error message about not finding english
- if I can get it to finish (sometimes enter works despite no continue button) I have received this message:
       * Notice: Undefined index: upgrade_upgrade_utility in (path)\upgrade.php on line 3334

![upgrader-lost-graphics-button](https://cloud.githubusercontent.com/assets/23568484/22359249/64e6bcb8-e3fa-11e6-93c7-c6d5f66edbfb.JPG)
![upgrader-lost-graphics-freezes](https://cloud.githubusercontent.com/assets/23568484/22359250/64eaf242-e3fa-11e6-8449-2ab7091d2cbd.JPG)
![upgrader-errors](https://cloud.githubusercontent.com/assets/23568484/22359248/64e648aa-e3fa-11e6-9fc2-419fb52672ef.JPG)

Oldiesmann:
That undefined constant thing is my fault. Will fix that .

Oldiesmann:
#3880 fixes the "undefined constant" error. That might fix some of the other issues as well, so try again and let me know what else is still broken.

sbulen:
Still getting all these new errors.  I can't make it to the end to confirm the undefined constant error is OK.

sbulen:
Looks like we got the Continue button back.  

The '30 seconds' message usually comes up when the DB fails.  I think we have a syntax error early in the script now.   
![image](https://cloud.githubusercontent.com/assets/23568484/22392032/2590c09e-e4aa-11e6-99e6-0fae410bdd76.png)

sbulen:
Waitaminute...   I needed to restore my old Settings.php - didn't realize it got putzed...  Let me try again...

sbulen:
Getting this error now:
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '} } updateSettings(array('langList' => json_encode($newlangs)))' at line 1

![image](https://cloud.githubusercontent.com/assets/23568484/22392774/e6389072-e4b1-11e6-9d3f-34da43fca8a0.png)

sbulen:
It's the apostrophe on line 2521

sbulen:
It's the brace to indicate it's PHP not SQL...   Testing...

[...]

PR #3883, opened 2017-01-28
Fix some typos in upgrader

sbulen:
Addresses some of the problems listed in this issue:
Upgrader: Freezes, loses graphics, loses buttons, errors #3877


sbulen:
The upgrader doesn't complete properly at the moment...  This at least gets past the DB issues.

PR #3886, opened 2017-01-28
Fixes #3723

### LINKED ISSUE #3723: Quick edit post
**message**

Quick edit -> save

frandominguezl:
Thanks to albertlast for the fix

### DIFF
--- a/Themes/default/scripts/topic.js
+++ b/Themes/default/scripts/topic.js
@@ -436,11 +436,11 @@ QuickModify.prototype.modifySave = function (sSessionId, sSessionVar)
 	var i, x = new Array(),
 		oCaller = this,
 		formData = {
-			subject : document.forms.quickModForm['subject'].value.replace(/&#/g, "&#38;#").php_to8bit(),
-			message : document.forms.quickModForm['message'].value.replace(/&#/g, "&#38;#").php_to8bit(),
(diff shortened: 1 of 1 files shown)


ThomasFok:
Should all `php_to8bit()` function call be removed?

albertlast:
The problem with php_to8bit() is that we don't know,
what there task was.
So maybe be some alternative code is needed to do this work what the php_to8bit() do before.

Oldiesmann:
The function is supposed to "convert a string to an 8 bit representation (like in PHP)" based on the comment right before it's defined in script.js. I have no idea what that means specifically or why that's even necessary.

albertlast:
d3vcho try this with chines symbole and keep working

frandominguezl:
@Oldiesmann This is ready to merge. Travis fails because instead of "Signed-off-by" I wrote "Signed-off by". Should I push a fix to that?

albertlast:
when i see correctly you do the end of line thing again

frandominguezl:
@albertlast that's not the error Travis is throwing

jdarwood007:
FYI, You don't need to specify manually the signed off by in your message.  Most UI clients have a checkbox or option called "Sign off" and from the command line you add -s or --signoff

PR #3887, opened 2017-01-28
Fix attachment filename problem if the filename starts with non-ASCII character.

ThomasFok:
If the filename of attachment starts with a non-ASCII character (like Chinese), the filename is missing after posting.

For example, the filename of the png file is `你好.png`. But after posting, it shows:

image


Oldiesmann:
Does `$_FILES['attachment']['name'][$n]` always contain just the filename? The `basename()` function is used to get just the last part of the path (so in this case just the filename) rather than the entire path. According to the PHP manual, `basename()` will work with multibyte characters only if `setlocale()` is used properly and set to the appropriate locale, so if we have entire paths in that array we may need to find a workaround (since we can't prevent people from uploading files with names containing characters that don't correspond with their locale).

ThomasFok:
I am not sure. But if `$_FILES['attachment']['name']` contains the full path of the file being uploaded, the browser is disclosing the directory structure of the client machine to the server. And it is a security hole.

sbulen:
I just spent a ton of time looking at attachments, & I am pretty sure there is no directory information in the $_File reference.   (I'm still learning this, so "pretty sure" is about as good as I can do...)

At this point in the process, it has already been uploaded - so it would not point to a directory on the user's computer.  

The logic in question is doing further checks before putting in into post_tmp* format in the SMF attachment directory.   (Where it will do even further checks before renaming it to [hash].dat format & storing an attachment record...  SMF'ers do lots of checks...)

I believe the file name includes ONLY the file name (which does come from the client).   But it is placed in the server's temporary directory until we do the move_uploaded_file() to put it in the SMF directory.  I don't think the server's temp directory is included in the string - I never saw directory information while in the step debugger looking at this stuff.  The temp directory must be assumed at this point.  

More here:
http://php.net/manual/en/features.file-upload.post-method.php

And here:
http://php.net/manual/en/function.move-uploaded-file.php

ThomasFok:
This comment from `function.move-uploaded-file.php` is worrying. Maybe we can replace `basename()` with our own implementation.

```
// basename() may prevent filesystem traversal attacks;
```

ThomasFok:
Replaced `basename` with a function dedicated to sanitize filename.

sbulen:
Interesting point...    We don't know anything about the name itself.    It comes straight from the client.

sbulen:
This explains the need:
http://www.acunetix.com/websitesecurity/php-security-3/

sbulen:
I'd consider going back to basename () and using setlocale ().   I have found a few comments out there like this:
_This is still a bug in 2016, but setlocale(LC_ALL,'en_US.UTF-8'); works._

One example:
http://stackoverflow.com/questions/4451664/make-php-pathinfo-return-the-correct-filename-if-the-filename-is-utf-8

ThomasFok:
`setlocale` affects lots of functions. It may break existing code.

sbulen:
Very good point.  

We do call setlocale in a couple other places, eg., in subs.php.  I suspect that call is intended to be global.   Maybe that call needs a tweak to make basename work?

sbulen:
This is an excellent bug, enabling true UTF8 support...  

Just did some poking around...   This is what I've found:
  -  The setlocale call in Subs is very specifically targeted to representing time, nothing else; it is the only setlocale call I can find in SMF
  -  basename is used ALL OVER SMF...  Packages, Languages, Themes, Hooks, Smileys & Attachments
  -  In Attachment processing, it's used in ManageAttachments.php & Attachments.php

I think if we want "full" utf8 support for all _SMF internals_ (e.g., in package names), we probably need a global setlocale approach to address Packages, Languages, Themes, & Hooks.   I think that's an SMF-org level decision & activity beyond the scope of this bug.  

My recommendation is to proceed as follows...   For attachments, I think replacing basename is safe it you put the file-traversal prevention edits in your function.   There was a good example in the stack overload article above that stripped unsafe characters.  I'd consider putting your new version in Subs-Attachments.php or possibly even Subs.php, and using it for all 3 uses of basename in ManageAttachments.php & Attachments.php.   (If you leave out the ManageAttachments call, the file may break on subsequent file maintenance activity.)   

Smileys may need a further look; user-uploaded Smileys may need similar treatment.  If so, Sub.php may be a more appropriate place for your function.  

Does this make sense?

ThomasFok:
Why does `processAttachments()` exist in both `Subs-Attachments.php` and `Attachments.php`? What is the story of these two files?

sbulen:
I know one is used in the user-facing upload process.  

I suspect the other is used in internal file maintenance, but I never confirmed that.  IIRC, they're pretty different.

sbulen:
@ThomasFok -  I wonder if adding setlocale with the following characteristics will work:
      setlocale(LC_CTYPE,'en_US.UTF-8');

Somewhere up there, like in index.php.   The LC_CTYPE should restrict its scope to character processing & leave things like currency & time alone.  

That way we can rely on basename() and pathinfo() safely throughout.

ThomasFok:
@sbulen Adding `setlocale` to index.php, of course, can fix this particular bug. It is up to the project management team to decide whether it is appropriate to make such change at this stage of development.

sbulen:
@ThomasFok   Yep.  I believe the proper way to pose the question is thru a PR!  (After testing it, of course...)

reloadsettings() in load.php may be more appropriate than index.php, as that's where most of the settings work is done, and it's one of the first things called by index.php.  

After some thought, I now think it's important to do, given that basename() is used throughout SMF.  Otherwise the behavior of basename() is left somewhat up to chance, as you found during testing.

sbulen:
@ThomasFok I wonder if something else is going on here.   I tried to reproduce the problem, but can't.  See the picture below.  Could you post a string here for me to try?   I used IE, Chrome, & a variety of strings.

I tried a few strings in Chinese, Lao & Khmer...   All worked fine...   Did I choose the wrong strings?

utf8-tests-1

ThomasFok:
@sbulen I use the filename `你好.png`. But I think any Chinese string should be the same. Maybe the default locale of your webserver is already UTF-8.

If I execute the following php file, I get `.png` as output.

```php
<?php
$filename = "你好.png";
echo basename ( $filename );

sbulen:
What do you see when you do the following? 
echo setlocale(LC_ALL, "0");

This is supposed to display your current settings.

My system comes back with one simple "C", explained here...
http://unix.stackexchange.com/questions/87745/what-does-lc-all-c-do

I'd also be interested in seeing the default charset in php:
echo ini_get("default_charset");

The setlocale functions appear to be very OS/installation dependent.  If telling php to anticipate utf8 (using an ini_set) solves the issue, that may be a better solution.

ThomasFok:
I get a letter "C" as well.  But `echo ini_get("default_charset");` returns UTF-8.

sbulen:
I believe the "C" means use your OS defaults for locale-dependent functions, and the php ini setting means use UTF8.  These match my settings.   (I'm testing primarily on a local Windows server.)  

My conclusion is that something at the OS level isn't playing nice, & needs to be overridden.  I bet others will have the same experience.  

Have you tested the proposed setlocale(LC_CTYPE,'en_US.UTF-8') fix?  Did it work?  Note that it fails in my environment - Windows uses a different set of parameters.  But that's OK, just allow it to fail if you go that route.  setLocale comprehends it may be used in different environments & allows you to pass it several different strings to try to get a hit...

I would support either solution: the setlocale change (if it tests out OK) or completing the edits you started above.  I lean slightly towards setlocale, as I think it is a bit more standards based & allows us to use functions like basename.  And it will make the current usage of basename safer throughout SMF.  

I'd be interested in knowing what the other SMF folks think...    @illori , @Oldiesmann , @colinschoen  do you have a recommendation on this question?

ThomasFok:
` setlocale(LC_CTYPE,'en_US.UTF-8')` can fix the problem.

Oldiesmann:
@ThomasFok Will that work with every language, and regardless of what characters are in the filename?

ThomasFok:
UTF-8 is backward compatible with ASCII.  My code should work with any languages provided that the string is in UTF-8 encoding.

XinYenFon:
I tested this PR with Turkish chars (Ğ ğ Ü ü Ş ş İ ı Ö ö Ç ç) result was good.
screenshot-localhost 2017-02-15 17-22-37

XinYenFon (approved):

sbulen:
@Antes, the pr, in its current state, does not provide the security checks desired to prevent files stem traversal attacks.    It does make utf8 work, though.

An alternative approach is proposed in the above discussion.

sbulen:
@ThomasFok  - 
The PR:  Issue setlocale to enable utf8 processing in certain php functions #3948
was merged.  Let us know if this resolved your issues!

ThomasFok:
@sbulen Yes, it did.

PR #3889, opened 2017-01-28
Fix issue freezing upgrader

sbulen:
Fixes one of the issues in:
Upgrader: Freezes, loses graphics, loses buttons, errors #3877

This is causing the freeze.  

I'm not sure this logic is necessary?  I don't think there is a langFile setting in smf_settings.  If it was in 2.0 would it have been JSON encoded?   @Oldiesmann

### DIFF
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -2510,25 +2510,6 @@ ADD COLUMN location VARCHAR(255) NOT NULL DEFAULT '';
 /******************************************************************************/
 --- Cleaning up after old UTF-8 languages
 /******************************************************************************/
----# Update the forum language list
----{
-	$langlist = json_decode($modSettings['langList'], true);
-	$newlangs = array();
-
-	// Strip "-utf8" from any languages and change them to regular if necessary
-	foreach($langlist as $internal => $external)
-	{
-		// Make sure we wont end up with a duplicate here...
-		if (!array_key_exists(str_ireplace('-utf8', '', $internal), $newlangs))
-		{
-			$newlangs[str_ireplace('-utf8', '', $internal)] = str_ireplace(' UTF-8', '', $external);
-		}
-	}
-
-	updateSettings(array('langList' => json_encode($newlangs)));
----}
----#
-
 ---# Update the members' languages
 UPDATE {$db_prefix}members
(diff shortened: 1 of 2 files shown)


live627:
@Oldiesmann no need to duplicate part of getLanguages()

https://github.com/SimpleMachines/SMF2.1/blob/9d4ead31bab83c5ab42878c40748d10d4a029fdc/Sources/Load.php#L2796

sbulen:
@live627 - Given the 'real-time' mid-upgrade need, I don't think the 'getLanguages' call helps here.    

The upgrader faces an interesting challenge...   Historical language settings in settings.php & the user profiles may have -utf8 in them if the site had been converted to utf8.  Going forward, we are not using language files with that suffix, utf8 is assumed.  So we need to do some string tweaks mid-upgrade or things break, because there isn't a -utf8 language file.  

Most of this was done perfectly.  I don't understand why this piece is needed, and it is causing issues.  It may yet need to be tweaked & re-added.

Oldiesmann:
@live627 What part of getLanguages am I duplicating? The only thing that calling getLanguages would do is ensure that the langList is up to date based on what language files are available (or were available whenever the cache was last updated, depending on the situation). It doesn't clean up the old "-utf8" languages. We can't assume the user has deleted all the old language files or uploaded the new ones yet.

@sbulen What part of that code is causing the freeze? By the time we get to the end of the upgrade SQL file, everything should have been converted to JSON already.

sbulen:
I agree it doesn't duplicate getLanguages.  

@Oldiesmann : The convert to JSON step comes much later, it's step 6.  The DB changes are step 4.  But that's not the issue.  

Even without this code, langList is populated.  Since it's new in 2.1, it's empty, so Load.php initializes it with 'english' in JSON format.  Since langList is new in 2.1, I don't think it will ever have a value of -utf8 in there.  

So I'm pretty sure this step is not needed.  langList is properly populated already.

As to where it is failing, I don't know exactly (it's eval'd, so the errors are somewhat opaque).  I was hoping you could test it out & determine the source of that, if it is determined that this step is needed.  

_But without this PR, the upgrader does not run to completion._  Which makes it difficult for folks to test & contribute.

Oldiesmann:
I didn't realize langList was a new setting in 2.1. In that case the only thing we need is the part to strip "-utf8" from the "lngfile" values in the members table.

PR #3893, opened 2017-01-29
Drop default on Text and Blobs fixing #3385

### LINKED ISSUE #3385: BLOB/TEXT column 'txt' can't have a default value
On windows mysql 5.7.9 php 5.6.16  while trying to perform an upgrade.

Setting mysql mode to MYSQL40 doens't seem to work.  Seems to be caused by this bit:

``` php
foreach ($columns as $column)
                        {
                            $updates_blob .= '
                                CHANGE COLUMN `' . $column['Field'] . '` `' . $column['Field'] . '` ' . strtr($column['Type'], array('text' => 'blob', 'char' => 'binary')) . ($column['Null'] === 'YES' ? ' NULL' : ' NOT NULL') . (strpos($column['Type'], 'char') === false ? '' : ' default \'' . $column['Default'] . '\'') . ',';
                            $updates_text .= '
                                CHANGE COLUMN `' . $column['Field'] . '` `' . $column['Field'] . '` ' . $column['Type'] . ' CHARACTER SET ' . $charsets[$upcontext['charset_detected']] . ($column['Null'] === 'YES' ? '' : ' NOT NULL') . (strpos($column['Type'], 'char') === false ? '' : ' default \'' . $column['Default'] . '\'') . ',';
                        }
```

albertlast:
Fixed #3385

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -2795,9 +2795,9 @@ function ConvertUtf8()
 						foreach ($columns as $column)
 						{
 							$updates_blob .= '
-								CHANGE COLUMN `' . $column['Field'] . '` `' . $column['Field'] . '` ' . strtr($column['Type'], array('text' => 'blob', 'char' => 'binary')) . ($column['Null'] === 'YES' ? ' NULL' : ' NOT NULL') . (strpos($column['Type'], 'char') === false ? '' : ' default \'' . $column['Default'] . '\'') . ',';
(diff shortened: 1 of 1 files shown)


Oldiesmann:
Since the code just looks to see if type name contains "char", wouldn't this also prevent setting default values for varchar columns?

albertlast:
Oops than mark the issue as solved

Issue #3901, opened 2017-01-30
Upgrader:  Various issues impacting stability & performance


sbulen:
I'm still looking at the upgrader.   I was wondering why sometimes you get the "30 seconds" exception & sometimes you don't.  I looked more closely at the issues, & under the covers, there are a number of things big & small that should be fixed.   The upgrader somehow forces its way through it...  I'm a little surprised it completes successfully.  

Over the next few days I'll be testing & submitting fixes to these issues.

Some examples that I found upon reading the code near 30-sec dropouts:
(1) ~~Syntax error creating the mentions table.~~  Edit - mentions table is OK 
(2) Mentions table is built differently in the upgrader than in the installer.
(3) log_topics.  First the unwatched col is created.  No problem.  But then there is a verb that tries to rename a non-existent column (disregarded) to unwatched.  This will always fail, first, because disregarded doesn't exist, and second, because unwatched does.   
(4) Drops columns before the index it's exclusively used in (one example: member_ip_old).  Since the column is gone, the index is gone & dropping the index is a syntax error.   Reversing the order will be more efficient & will not have syntax issues.  
(5) Speaking of which, it drops the index on the messages table, where it created it on the members table...   

I think the root cause here is that these statements are run indirectly by the upgrader, and the errors are very well hidden & easily missed.  If I can figure out how to make these more visible & debuggable, I'll do so...

illori:
disregard did exist in prior versions of SMF 2.1 so it does not hurt to have that part still there for those that did use that version.

sbulen:
I figured it was an old column.   But this will never remove the disregarded column, even if it still exists, due to the syntax.   It basically says:
  -  create unwatched
  -  rename disregarded to unwatched      <= this will always fail, since unwatched will exist already

We could either try the rename first or explicitly delete it.  But the existing code will not accomplish what is desired.

sbulen:
The upgrader definitely allows for reruns.  upgrade_query(), line 1863+, is written to ignore errors like dropping tables that don't exist, and adding column names that already exist.  

But at some point, it appears the upgrader stopped reporting ALL errors.  And some of these are pretty bad, like the syntax errors.  

It'd be nice if it ignored what we wanted it to ignore, and reported all other errors.  

Some are causing the upgrader to hang a bit - hence the 30 sec timeout warnings.  (A side note: I only get the warnings on my InnoDB install.  Apparently it is more strict than MyISAM.)

sbulen:
[...]

PR #3905, opened 2017-02-01
Some installer & upgrader tweaks

sbulen:
This is to address some of the issues reported here:
Upgrader: Various issues impacting stability & performance #3901

I don't have an instance of pg, so I'd appreciate it if a pg user could look closely at this.

### DIFF
--- a/other/install_2-1_mysql.sql
+++ b/other/install_2-1_mysql.sql
@@ -1179,9 +1179,9 @@ CREATE TABLE {$db_prefix}user_likes (
 CREATE TABLE {$db_prefix}mentions (
   content_id INT DEFAULT '0',
   content_type VARCHAR(10) DEFAULT '',
-  id_mentioned INT DEFAULT 0,
-  id_member INT(10) UNSIGNED NOT NULL DEFAULT 0,
-  `time` INT NOT NULL DEFAULT 0,
+  id_mentioned INT DEFAULT '0',
+  id_member INT(10) UNSIGNED NOT NULL DEFAULT '0',
+  `time` INT NOT NULL DEFAULT '0',
   PRIMARY KEY (content_id, content_type, id_mentioned),
   INDEX content (content_id, content_type),
   INDEX mentionee (id_member)
--- a/other/install_2-1_mysql.sql
+++ b/other/install_2-1_mysql.sql
@@ -1179,9 +1179,9 @@ CREATE TABLE {$db_prefix}user_likes (
 CREATE TABLE {$db_prefix}mentions (
   content_id INT DEFAULT '0',
   content_type VARCHAR(10) DEFAULT '',
-  id_mentioned INT DEFAULT '0',
-  id_member INT(10) UNSIGNED NOT NULL DEFAULT '0',
-  `time` INT NOT NULL DEFAULT '0',
+  id_mentioned INT DEFAULT 0,
+  id_member INT(10) UNSIGNED NOT NULL DEFAULT 0,
+  `time` INT NOT NULL DEFAULT 0,
   PRIMARY KEY (content_id, content_type, id_mentioned),
   INDEX content (content_id, content_type),
   INDEX mentionee (id_member)
--- a/other/install_2-1_postgresql.sql
+++ b/other/install_2-1_postgresql.sql
@@ -1717,9 +1717,9 @@ CREATE INDEX {$db_prefix}user_likes_liker ON {$db_prefix}user_likes (id_member);
 CREATE TABLE {$db_prefix}mentions (
   content_id int default '0',
   content_type varchar(10) default '',
-  id_mentioned int default 0,
-  id_member int NOT NULL default 0,
-  time int NOT NULL default 0,
+  id_mentioned int default '0',
+  id_member int NOT NULL default '0',
+  time int NOT NULL default '0',
   PRIMARY KEY (content_id, content_type, id_mentioned)
 );
 
--- a/other/install_2-1_postgresql.sql
+++ b/other/install_2-1_postgresql.sql
@@ -1717,9 +1717,9 @@ CREATE INDEX {$db_prefix}user_likes_liker ON {$db_prefix}user_likes (id_member);
 CREATE TABLE {$db_prefix}mentions (
   content_id int default '0',
   content_type varchar(10) default '',
-  id_mentioned int default '0',
-  id_member int NOT NULL default '0',
-  time int NOT NULL default '0',
+  id_mentioned int default 0,
+  id_member int NOT NULL default 0,
+  time int NOT NULL default 0,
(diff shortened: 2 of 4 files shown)


other/install_2-1_mysql.sql
```diff
@@ -1179,9 +1179,9 @@ CREATE TABLE {$db_prefix}user_likes (
 CREATE TABLE {$db_prefix}mentions (
   content_id INT DEFAULT '0',
   content_type VARCHAR(10) DEFAULT '',
-  id_mentioned INT DEFAULT 0,
-  id_member INT(10) UNSIGNED NOT NULL DEFAULT 0,
-  `time` INT NOT NULL DEFAULT 0,
+  id_mentioned INT DEFAULT '0',
```
albertlast: Any special reason for using single quotes?
Usually your use for integer value no quotes because they are no strings.

sbulen: Mainly to be consistent.  Single quotes are used throughout all of these .sql files.  For installer & upgrader.  

Even the pg ones.  (I wish I could test pg, unfortunately I can't...)

The only exceptions are these 3 lines, where none were used, and which generate syntax errors.

sbulen: I have tested all of this code, for the installer & upgrader, but for MYSQL only.

albertlast: it's sql standard not to use single quote für integer,
but the conversion from string to int is not so hard,
that is reason why this is working...

but this keep wrong from standard pov.

so in my eyes keep the right thing right and wrong think wrong,
or fix the wrong ones, but don't make the right one to wrong one only because it's looks better.

Oldiesmann: I'm not sure why we would ever use single quotes around integer values like that. It's a good idea to always pass them as INTs rather than strings. For now you can just fix it for this PR and we can do a separate one to clean up the rest.

sbulen: MySQL does not accept default 0.  It requires default '0'.   

MySQL does a lot of things that are not ansi-standard.  That's one of the reasons folks like pg.  

I think pg just tolerates it...

sbulen:
EDIT:  This note is inaccurate, see further discussion below.....

> MySQL does not accept default 0.   It requires default '0'.
> To verify, paste the syntax into the pmpMyAdmin window. 
> This change is to address the upgrader freezing up when it hits this verb, while mysql tries to sort things out.

albertlast:
why do you believe that? 
see http://sqlfiddle.com/#!9/bccac0/1/0

sbulen:
I don't trust the online syntax checkers, to be honest, they've failed me before.   Most searches I've done on this question don't help, as there are some actually quite lengthy debates out there on this topic.  About 60% agree with you - no quotes.  The MySQL manual provides inconsistent examples - where default 0 is used as well as default '0' is used.

This is what I know:   
(1) It fails syntax check when using phpMyAdmin to test.  
(2) The upgrader freezes temporarily; it normally does this when handed a syntax error.  
(3) Literally every other line in the install & upgrade .sql scripts use the quotes.  I suspect a php setting or edit in the upgrade code.  

Over the next few days, I was going to debug the various freezes more closely & see exactly what was returned.  I don't think the upgrader should just ignore errors like that.   It's opaque to us, even with debug on.  We can't see what is really happening.

sbulen:
How's that?

Do the other changes make sense?

sbulen:
I'm game to remove all the single quotes.   

Give me a few days to test it a bunch of times...   I REALLY want the Upgrader to stay stable.

I'll do the Installer and Upgrader, pg and mysql.   

I can't test pg, though.

Oldiesmann:
What version of MySQL are you running where you're getting syntax errors with `DEFAULT 0` instead of `DEFAULT '0'`? It works fine on my server (MariaDB 10.1).

screenshot_20170202_111624

sbulen:
Attached are the screenshots of the error I get in phpMyAdmin.   This is a php 5.6.19/mysql 5.7.11 environment.  I can get the table created with or without the quotes, despite the errors.  

I do not get these errors in any other environment.  I'm willing to chalk it up as either a bug in this particular configuration or a fluke in the environment, which is why I reverted the change.   

What is a bit more concerning is that the '0' format is used perfectly consistently throughout all 8 .sql files in the /other directory - for SMF 1x and 2x, for mysql & pg, and for the installer & the upgrader.  I believe these 3 lines are the only exceptions.  (I'm a little confused why, despite all that history, these 3 lines drew such attention!)

To be honest, with regards to removing all the quotes across all 8 files, I'm concerned with making such a big change, affecting hundreds of lines of code across all engines/versions.  It feels a bit reckless.  Given that we don't actually just RUN the sql directly, we run it thru our code that pre-processes it (depending on how written, either parse_sql() and/or an smcfunc and/or upgrade_query() and/or an eval()), I'm not 100% convinced yet we can rely on pure SQL as a guideline.  

I made the - possibly invalid - assumption that it was done this way for a purpose.  (No snark intended.)

I admit I'm still learning why that code does what it does. 

syntax-error-1
syntax-error-2

albertlast:
There are no mysql errors,
only some phpmyadmin errors are shown.

sbulen:
Thanks Albert.   I've gone full Elsa.  I've let it go.  As explained above, I removed the quotes from this PR.  

I am not signing up to change all of the other values to be consistent with int default 0, though, for reasons explained above.  I highly suggest a quick scan of all 8 .sql files in the /other directory.  It's that way in 2.0 as well.

It is actually an interesting discussion, & there are some pretty good threads out there:
http://stackoverflow.com/questions/6781976/mysql-quote-numbers-or-not

The argument from the security-minded folks who argue for the quotes, though, seems to only apply for queries.  Not table definitions.  The solution to add quotes applies to the php processing between the form & mysql.  So I see no harm either way, in table definition, to use default 0 or default '0'.

sbulen:
An explanation of the remaining changes on this PR:
 - removed the (3) on log_topics to make it consistent with the installer script
 - made id_member unsigned to make it consistent with the installer script
 - separated adding the column to log_topics from initializing it, to allow the upgrader to work in smaller more granular steps (and make it more consistent with the rest of our .sql files)
 - removed old 'disregarded' column using a drop instead of a rename; the rename won't work anymore since adding the new column was added explicitly prior
 - addressed the typo on removing the indexes - the table is members, not messages
 - reversed the order - removed the index before dropping the column; this is more efficient, since dropping the column would also remove the index, this allows the upgrader to do the work in two more granular steps

Any questions let me know

sbulen:
Is anything else needed on this PR?   Please advise.  Thanks,

albertlast (approved):

colinschoen:
Will review soon.

colinschoen (approved):
LGTM, approval

PR #3912, opened 2017-02-03
Generates the user's theme options list from Settings.template.php

Sesquipedalian:
- Previously, we were using hard-coded HTML in Profile.template.php, which defeated most of the purpose of `theme_options()` in Settings.template.php (namely, to allow themes to define their own sets of options). This PR changes that so that the user is shown a list of options according to what is defined in Settings.template.php. The new method is very similar to what is used to show the default options to the admin.
- This PR also rearranges the options in Settings.template.php into a more logical order.
- This PR also adds support for separators and titled sections in the theme settings/options.

### DIFF
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -1794,6 +1794,9 @@ function theme($memID)
 {
 	global $txt, $context;
 
+	loadTemplate('Settings');
+	loadSubTemplate('options');
+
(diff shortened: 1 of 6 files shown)


Sesquipedalian:
This is ready now. Here's a screenshot.

screen shot 2017-02-03 at 5 33 11 pm

PR #3918, opened 2017-02-05
maybe a better way to find the filename

albertlast:
to be ownest i don't know how to test this,
but i think is a bette way to finde the filename.
http://php.net/manual/en/function.basename.php


sbulen:
The problem with basename() is that it is not UTF8 compatible, unless you issue a setlocale() properly somewhere.   
I'm not quite sure where we should do the setlocale(), & the proper values need some thought, too.  
More discussion over here: Fix attachment filename problem if the filename starts with non-ASCII character. #3887

sbulen:
basename() is used throughout SMF, but will have the UTF8 issues that @ThomasFok discovered.  

I think it would be appropriate to have a setlocale() way up front, in index.php.   That would need some testing though.

Issue #3909, opened 2017-02-01
Skype icon not shown on poster information


frandominguezl:
After entering an username in the skype field (Profile > Forum Profile), the icon is not shown on the poster information

frandominguezl:
Alright, so, dumb of me I just changed the placement in "Bottom poster info". Now, the curious thing is that after saving that, on bottom poster info there's also displayed Gender even though is not set to be placed there...

![](https://i.gyazo.com/2c0d5382acc88df9d1f6a7c661d820fe.png)


![](https://i.gyazo.com/a695f489ffe9909922c5112e4621637a.png)

Oldiesmann:
Gender is shown "with icons" as it should be. "Bottom Poster Info" shows things below all of the other poster info (icons, location, IP, etc.).

frandominguezl:
Before changing skype to "Bottom Poster Info", Gender was not displayed there. After changing it, it's now displayed.

Oldiesmann:
It's possible that the display_fields setting in the database got messed up during an upgrade due to some of the issues we've had in the past with conversion from serialize to JSON. In the screenshot you posted though, both the gender and Skype icons are in the proper place based on your settings.

frandominguezl:
Then I don't understand the behaviour of this, since both Skype & Gender were set as "with icons" and none of them appeared before. That's why I first opened this issue.

Oldiesmann:
Definitely sounds like the "displayFields" setting got corrupted at some point. That setting contains a JSON-encoded array of information about all of the custom profile fields. In earlier versions of SMF 2.1, we used serialize() instead of JSON for storing arrays in the database, and there have been some issues with that conversion process during upgrades in the past. When you changed the Skype field, that setting got updated, so everything is displaying properly now.

sbulen:
The JSON for displayfields appears to be built properly upon a fresh upgrade.

Like D3vcho, it feels like I chased it around a bit.  Almost like displaying the icons "kicked in" after I started putzing with it.  I don't suspect the JSON though.

Oldiesmann:
It's related to #3654 I believe

frandominguezl:
I updated the files of my hosted test forum (not localhost) and seems the error is not happening. Can anybody test this, please?

sbulen:
This is really a good catch...   

I suspect a disconnect between the migrated field names and what the app is using...
The app is looking for cust_skype, but the upgrader is using cust_skyp.   
[...]

PR #3925, opened 2017-02-08
Align custom fields with app

sbulen:
Some profile fields names are not aligned with app expectations.  Check out:
Skype icon not shown on poster information #3909

Also confirmed installer files were aligned - all good.

Tests out OK.


frandominguezl:
Just not sure why did you change yim to yahoo; I thought yim standed for "Yahoo Instant Messenger" and that's why yim was used instead of yahoo. Am I wrong?

sbulen:
Internally the app is calling that field cust_Yahoo, not cust_yim, so it was not seeing the migrated data.   Look at the screenshots I provided in the issue.

frandominguezl:
Ew yes, my bad!

PR #3931, opened 2017-02-12
Fix #3930

### LINKED ISSUE #3930: typo
themes\default\languages\Admin.english.php (1 hit)
	Line 430: $txt['attachment_transfer_desc'] = 'Transfer files betwen directories.';

BETWEEN

frandominguezl:


XinYenFon:
You have problem with file endings fix that issue.

frandominguezl:
Travis build failed again. It fails with and without touching the end of file. There should be something wrong here. Can you check @jdarwood007 ?

jdarwood007:
You removed extra spacing at the end of the file.

screen shot 2017-02-13 at 2 38 19 pm

frandominguezl:
Yes, this is the last commit I did just to test. But the previous one in this PR and in another one, shows the same error even without touching it.

Oldiesmann:
The problem is that whatever editor you're using is adding an extra blank line after the closing PHP tag

PR #3937, opened 2017-02-15
Improvements on mobile

### LINKED ISSUE #3903: Scroll don't work on small screens
```
@media (max-width: 480px) {
	/* Stay where you are in mobile popup enabled */
	body {
		overflow: hidden;
	}
```
Why scroll disabled here? Forum unusable on small screens.

XinYenFon:
fixes #3903 
closes #3923

### DIFF
--- a/Themes/default/Admin.template.php
+++ b/Themes/default/Admin.template.php
@@ -1605,7 +1605,7 @@ function template_admin_quick_search()
 	global $context, $txt, $scripturl;
 	if ($context['user']['is_admin'])
 		echo '
-								<span class="floatright">
+								<span class="floatright admin_search">
 									<span class="generic_icons filter centericon"></span>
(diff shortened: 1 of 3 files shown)


XinYenFon:
@jdarwood007 seems Travis fails on GPG Sign-off's as well 😆

jdarwood007:
I haven't tested GPG signed stuff, But I was looking into using GPG myself.
Seems my debugger code is faulty and didn't capture the secondary message so I don't know why its still failing.

Oldiesmann:
Looks like it's failing because there's no "signed-off-by" line in the commit when you use GPG signing. There is a `verify-commit` flag for git, but I don't know if we can have Travis run `git verify-commit` for us.

jdarwood007:
Travis runs a git clone command, so it should be easily done to run a git verify-commit.  It needs the actual commit though which is what is currently failing here I think.  The commit has parents (from the merges) which contain the sign off message.  So I think I need to rebuild the code to find all parents and verify if any of them have signed off.

PR #3943, opened 2017-02-17
Moving build tools to a separate repo

jdarwood007:
This should setup Travis to work properly using a separate repo to handle all the build tools.  This will prevent clutter here from these build tools from making too much noise on the main branch and allow me to debug them easier.


jdarwood007:
Ok, this should be ready to go now.  Seems to work as expected.

PR #3947, opened 2017-02-20
div table

snrj:


snrj:
index.css or ......
`.ssitable{
	display: table;
	width: 100%;
}
.ssitable_row {
	display: table-row;
}

.ssitable_cell, .ssitable_head {
	border: 1px solid #eee;
	display: table-cell;
	padding: 3px 10px;
}
.ssitable_heading {
	background-color: #EEE;
	display: table-row;
	font-weight: bold;
}
.ssitable_footer {
	background-color: #EEE;
	display: table-row;
	font-weight: bold;
}
.ssitable_body {
	display: table-row-group;
}`

XinYenFon:
Well, this is surely nice to have which I also planned for a long time (from html table to css table), unfortunately due to time limitations we decided to skip this conversion for this version.

I'm not going to approve those changes but still last word comes from the development team. Sorry once more and big thanks for taking your time.

albertlast:
What is the advantage of this solution?
And Signed-off-by: ... is missing in the commit log.

Oldiesmann:
@Antes do you have any further info on the limitations? I'm not entirely sure what this would do to begin with.

XinYenFon:
@Oldiesmann i mean once we start converting html table 's into CSS Tables we should be doing it for entire software not just for SSI, which is the time consuming part.

snrj:
remove

PR #3948, opened 2017-02-23
Issue setlocale to enable utf8 processing in certain php functions

sbulen:
php functions such as basename() and pathinfo() require setlocale to be set in order to properly operate with the utf8 character set.  

basename() and pathinfo() are safe & preferred methods operate on uploaded files, and provide protection from file traversal attacks.  Both are used throughout SMF today.  This PR makes them utf8-friendly.

This PR negates the need to write our own version of pathinfo or basename, & replaces PR #3887 

Feedback welcome,

### DIFF
--- a/Themes/default/languages/index.english.php
+++ b/Themes/default/languages/index.english.php
@@ -16,7 +16,7 @@ $txt['lang_dictionary'] = 'en';
 $txt['lang_spelling'] = 'american';
 
 // Ensure you remember to use uppercase for character set strings.
-$txt['lang_character_set'] = 'ISO-8859-1';
+$txt['lang_character_set'] = 'UTF-8';
 // Character set and right to left?
 $txt['lang_rtl'] = false;
 // Number format.
--- a/Themes/default/languages/index.english.php
+++ b/Themes/default/languages/index.english.php
@@ -16,7 +16,8 @@ $txt['lang_dictionary'] = 'en';
 $txt['lang_spelling'] = 'american';
 
 // Ensure you remember to use uppercase for character set strings.
-$txt['lang_character_set'] = 'UTF-8';
+// Leave this entry at ISO-8859-1 for the language editor
+$txt['lang_character_set'] = 'ISO-8859-1';
 // Character set and right to left?
 $txt['lang_rtl'] = false;
 // Number format.
--- a/Themes/default/languages/index.english.php
+++ b/Themes/default/languages/index.english.php
@@ -16,7 +16,6 @@ $txt['lang_dictionary'] = 'en';
 $txt['lang_spelling'] = 'american';
 
 // Ensure you remember to use uppercase for character set strings.
-// Leave this entry at ISO-8859-1 for the language editor
 $txt['lang_character_set'] = 'ISO-8859-1';
 // Character set and right to left?
 $txt['lang_rtl'] = false;
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -2651,7 +2651,7 @@ function loadLanguage($template_name, $lang = '', $fatal = true, $force_reload =
 				$found = true;
 
 				// setlocale is required for basename() & pathinfo() to work properly on the selected language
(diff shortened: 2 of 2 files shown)


albertlast:
I think it's the right way,
only i ask my self if maybe important to change language.
In your pr is this hardcoded "en_US" i mean this is better as nothing,
but maybe we should be able to change this value by settings/vars?

sbulen:
@albertlast - Great question.  Note that we focused on LC_CTYPE - which focuses on character processing.  

_If we needed to specify another locale category, e.g., LC_MONETARY or LC_TIME, we would absolutely have had to offer an SMF setting to control it._  

I did a number of searches on this issue, and they all used that exact string to resolve it.  Given that it supports all UTF8 characters, I'm not sure another language setting would make a difference?  

Note that these locale strings are OS dependent as well.  setlocale() will try anything you feed it and ignore errors - it understands you may be feeding it a few things to try to provide multi-OS support.  The errors we observed were Linux specific, and this is the Linux string to remedy the problem.

XinYenFon:
@sbulen why are you setting this as en_US.utf8 ? You should call $txt['lang_locale'] which returns locale in the language active without .utf8 part.

sbulen:
I wasn't aware of $txt['lang_locale']...  Hadn't run across it yet!  

How is it updated/maintained?  Is it safe to use in setLocale()?   If so, I'll make that change.   It'll take me a few days, I'm now travelling again & won't have access to my dev environment until Sunday/Monday.  

I see this entry in index.english.php as well:
    $txt['lang_character_set'] = 'ISO-8859-1';

Is this correct, or should it reflect UTF-8?   If so, I'll do that as well.  

Interesting $txt['lang_character_set'] is used throughout the app, SMF will utilize it over UTF-8.  This likely also needs to be addressed & tested.  We have a lot of string functions that are being forced to use 8859 charsets during processing (e.g., upper & lower case processing, string comparisons).  This sure looks like it needs to be fixed as well.  If we wish to say we're UTF-8, I bet we should actually go UTF-8!

sbulen:
Updated per request

Themes/default/languages/index.english.php
```diff
@@ -16,7 +16,7 @@
 $txt['lang_spelling'] = 'american';
 
 // Ensure you remember to use uppercase for character set strings.
-$txt['lang_character_set'] = 'ISO-8859-1';
+$txt['lang_character_set'] = 'UTF-8';
```
XinYenFon: Please keep this as is, our language editor already taking care of this part based on the output :)

XinYenFon (changes requested):

sbulen:
@Antes - OK...  But just to be sure...  

I think this setting is overwriting UTF-8 processing in multiple places throughout the app.  For example, in Load:
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Load.php#L75
Also:
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Load.php#L2077

A grep reveals there are numerous other places this is used, and potentially negating UTF-8 processing.   

If $txt['lang_character_set'] is intended to be used by the language editor, then maybe these other instances should be changed to use $modSettings['global_character_set']?   

I guess my question is:  When should the one be used over the other?

XinYenFon:
That probably requires code cleanup because SMF 2.1 didn't started with utf8 only.

sbulen:
How's this?

XinYenFon (approved):

XinYenFon:
whelp my stupidity forgot to say, can you please revert (remove) comment line in language file. If you drop by IRC some time, i can explain how our LE works and how your comment changes description of string :/

sbulen:
Done.   

I thought the comment would be helpful - folks such as myself would think it was missed.

sbulen:
@Antes - I'll take you up on that offer!   I haven't used IRC in about 20 years, though...   Let me know what server/room you're on & I'll find a way there.

XinYenFon:
@sbulen http://www.simplemachines.org/community/index.php?topic=183880.0

PR #3949, opened 2017-02-25
Italicize hidden users

### LINKED ISSUE #3933: who's online page hidden user
on the who's online page, it does not clearly tell you if someone is hidden or visible as it does on the list in the info center. if you inspect the source on a user that should be hidden you see there is a <span class="member hidden"> but no css that goes along with it. i think we should add that it is italics to match the info center format.

colinschoen:
Fixes #3933 

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -3669,6 +3669,9 @@ p.information img {
 .action_who select {
 	margin-top: -1px !important;
 }
+span.member.hidden {
+	font-style: italic;
+}
 
 /* Icons (converted from img to css) */
 /* Stay here till I find a better place for you guys */


XinYenFon:
@illori can you confirm this is better to separate the hidden & non-hidden users on who's online page?

illori:
i have not applied the fix myself but it looks like it should do the trick yes.

colinschoen:
@illori Please let me know if you give approval on this. I'd like a second set of eyes prior to merging even if it is a small change.

illori:
applied the fix, it had no effect. the class has a space in the name, so in this case it still has no effect. i changed
`<span class="member', $member['is_hidden'] ? '_hidden' : '', '">` in who.template.php and changed the . to _ and then it worked without issue.... also required to disable Minimize css and JavaScript files as the change was not being seen as the minimized file was not being updated fast enough [when does it get updated?]

XinYenFon:
You should disable minimization on dev-branch its not really effective for development (or testing PRs).

illori:
I ended up doing that for testing which was helpful. The issue here needs a
little bit more work.

XinYenFon:
It is working as expected in my local environment. Try hard refresh on the page or validate your index.css via visiting inside your browser and check the added class in it.

illori:
I got it to work with the changes I posted. I don't have time to retest
right now. If you say it works that is good enough for me.

XinYenFon:
@colinschoen is there anything specific we are awaiting for to merge this ?

PR #3952, opened 2017-02-28
Fixes #3913 and CSS typo

### LINKED ISSUE #3913: Use this function to inform the moderators and administrators of an abusive or wrongly posted message.
Use this function to inform the moderators and administrators of an abusive or wrongly posted message.
is wrongly really a valid word? seems it is a valid word but i dont think it is commonly used.
https://en.wiktionary.org/wiki/wrongly

maybe change the wording to something like this...

"Use this function to inform the moderators and administrators of an abusive or message posted in error."

frandominguezl:


Themes/default/languages/index.english.php
```diff
@@ -480,7 +480,7 @@
 
 $txt['page_created_full'] = 'Page created in %1$.3f seconds with %2$d queries.';
 
-$txt['report_to_mod_func'] = 'Use this function to inform the moderators and administrators of an abusive or wrongly posted message.';
+$txt['report_to_mod_func'] = 'Use this function to inform the moderators and administrators of an abusive or message posted in error.';
```
Sesquipedalian: The old version was awkward, but this is grammatically broken. How about "an abusive message or one posted in error," or more simply "a problematic message"?

Sesquipedalian (changes requested):

Themes/default/languages/index.english.php
colinschoen: +1 for the latter.

Sesquipedalian: Alternatively, if we still want to highlight the idea of reporting abusive messages, we could try "Use this function to inform the moderators of an abusive or otherwise problematic message."

colinschoen (approved):
LGTM, approval

Sesquipedalian (approved):

PR #3953, opened 2017-02-28
Sorts topics by message time rather than message id

Sesquipedalian:
Under normal circumstances, sorting by id works well enough. But if messages and topics were imported from another source, such as another forum database or via an RSS posting mod or something, it is possible that an older message will have a higher id number than a newer message. Since the user normally expects sorting by the first or last post to return a list sorted chronologically rather than by SMF's internal id numbers, this change simply makes sure that we deliver as expected in all cases.

I don't *think* there could be any negative consequences of this change to the message index sorting, but if anyone can think of any issues that it might raise, please point them out.


Sesquipedalian:
Hm. I'm not sure what Travis is complaining about now. The log for the failed job is completely empty.

XinYenFon:
@Sesquipedalian you forgot to sign-off your commit ?

Sesquipedalian:
D'oh!

XinYenFon:
Its failing because of outage in AWS stuff (https://www.traviscistatus.com/).

jdarwood007:
Kicked off a rebuild for you.

However we leave the default sort by ID as it is more effiicent in a database than other sorting methods.  When you issue a ORDER BY, there is some work to be done and this on large forums this can really slow databases because of the manual sorting done in filesorting/temp tables.
https://dev.mysql.com/doc/refman/5.7/en/order-by-optimization.html

albertlast:
yeah it's a better idea to keep the sort by id.
your should rewrite the converter to fit this,
as to break smf.

Sesquipedalian:
What if we added two more indexed columns to the topics table called `first_msg_time` and `last_msg_time`? Then we could issue an ORDER BY against those columns in the same way that we already can against `id_first_msg` and `id_last_msg`, right? We'd need to add some code to update these values in the topics table when new posts are created, but that won't be hard.

albertlast:
Why you not update the id_topic so that it in the right order,
or why you not insert the data in the right order?

Why we try to fix something what is outside?

Sesquipedalian:
Topic ids should never, ever change once they are created. For example, if someone shares a link to http://www.example.com/forum/index.php?topic=78.0, but then we renumber the topic ids, that link will point to the wrong topic. Also, many other parts of the code refer to topics by id and store topic ids in other tables. Renumbering topic ids is not a good idea.

As for inserting data in the right order, I'm talking about when data is imported from other sources by various means. When we add a new topic from another source into our database, we need to give it an incremented id number. But the content of that topic may be minutes, hours, or years old. It does not make sense to present a three year old topic as though it were "newer" than a three day old topic, even if the three year old topic was imported into our database after the three day old topic.

albertlast:
The importing of data from "outside" is no function which smf support,
so is this use case not supported by smf -> smf is wrong side to fix this issue.

Sesquipedalian:
There are mods that can bring this situation about, as well as various converting, importing, and merging scripts. But how exactly one could end up with topics whose id numbers and first/last post times do not coincide is beside the point. The point is this: if the user says she wants the topics to be sorted chronologically, we should sort them chronologically. 

So I return to my question: if we added two more indexed columns to the topics table called `first_msg_time` and `last_msg_time`, and then sort based on those values rather than on `id_topic` and `id_last_msg`, does anyone see any reason why that would not work?

jdarwood007:
id_first_msg and id_last_msg are primarily used for being able to retrieve those messages quickly for things such as the MessageIndex where we want to display the first and last message details.  They are not used in sorting.  Using MIN/MAX here is slower.  We only use MIN/MAX during a rebuild/recount on forum maintenance.

If this is happening because of a conversion, the converter should do the ORDER BY itself.  Its better to let it do it and eat the extra temp tables (/filesorts) during the conversion than all the time.

Sesquipedalian:
> id_first_msg and id_last_msg are primarily used for being able to retrieve those messages quickly for things such as the MessageIndex where we want to display the first and last message details. They are not used in sorting. Using MIN/MAX here is slower. We only use MIN/MAX during a rebuild/recount on forum maintenance.

No, `id_last_msg` is the default sort key used when displaying the message index. [See here.](https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/MessageIndex.php#L246) We ORDER BY `is_sticky` and then whatever is set in `$_REQUEST['sort']`, which is usually `id_last_msg`. [See here.](https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/MessageIndex.php#L296)

> If this is happening because of a conversion, the converter should do the ORDER BY itself. Its better to let it do it and eat the extra temp tables (/filesorts) during the conversion than all the time.

There are a number of ways that this could happen. I stumbled across the issue when testing out an RSS posting bot mod that preserved the original publication date of the posts it created. The posts would correctly be dated to, say, 2012 or whenever, but in the message index they would be sorted into positions that made them appear to be newer than posts from last week. This situation had nothing to do with the RSS posting bot process itself, and everything to do with the way the MessageIndex.php orders the topics by id number rather than by date. Any other process that adjusts the dates on posts could produce a similar situation. So long as MessageIndex.php sorts topics by id when asked to sort them by date, the potential for this discrepancy exists.

The point is that regardless of *how*, or even *if*, a backdated topic was added to the database, we want the message index to sort the topics chronologically when the user requests that they be sorted chronologically. There is no good way to guarantee that will happen if we are actually sorting by id numbers.

jdarwood007:
Well I guess I was wrong on how MessageIndex handles it.  However id_first_msg and id_last_msg also contain a index, which helps prevent the file sort.

Sesquipedalian:
Yup. Which is why I am thinking that I will revert the current changes in this PR and instead add `first_msg_time` and `last_msg_time` columns and index them.

Sesquipedalian:
I'm interested to @albertlast's opinion on whether this is the best way to index the `first_msg_time` and `last_msg_time` columns. Both columns are written to at the same time, but only one or the other is read in any given query.

albertlast:
Well the first issue that your update syntax is wrong for pg:
http://stackoverflow.com/questions/7869592/how-to-do-an-update-join-in-postgresql
https://www.postgresql.org/docs/9.1/static/sql-update.html

Exp:

UPDATE smf_topics AS t
SET t.first_msg_time = mf.poster_time, t.last_msg_time = ml.poster_time
			from smf_messages AS mf, smf_messages AS ml
		WHERE 
t.id_first_msg = mf.id_msg 
and t.id_last_msg = ml.id_msg 
and (t.first_msg_time = 0 OR t.last_msg_time = 0) 

Personaly i like more the "with" syntax which is shown in the stackoverflow

The is in my eyes no other possibiltiy at the index topic.

Sources/MoveTopic.php
```diff
@@ -608,9 +608,12 @@ function moveTopics($topics, $toBoard)
 			// If not, update.
 			if ($row['first_msg'] != $topicMaxMin[$row['id_topic']]['min'] || $row['last_msg'] != $topicMaxMin[$row['id_topic']]['max'])
 				$smcFunc['db_query']('', '
-					UPDATE {db_prefix}topics
-					SET id_first_msg = {int:first_msg}, id_last_msg = {int:last_msg}
-					WHERE id_topic = {int:selected_topic}',
+					UPDATE {db_prefix}topics AS t
+						INNER JOIN {db_prefix}messages AS mf ON (t.id_first_msg = mf.id_msg)
+						INNER JOIN {db_prefix}messages AS ml ON (t.id_last_msg = ml.id_msg)
```
albertlast: Doesn't work for pg.

Sources/RemoveTopic.php
```diff
@@ -904,6 +904,18 @@ function removeMessage($message, $decreasePostCount = true)
 					)
 				);
 
+			// Ensure the recycle topic has the correct first/last message times
+			$smcFunc['db_query']('', '
+				UPDATE {db_prefix}topics AS t
+					INNER JOIN {db_prefix}messages AS mf ON (t.id_first_msg = mf.id_msg)
+					INNER JOIN {db_prefix}messages AS ml ON (t.id_last_msg = ml.id_msg)
```
albertlast: pg

Sources/RemoveTopic.php
```diff
@@ -988,6 +1000,18 @@ function removeMessage($message, $decreasePostCount = true)
 		call_integration_hook('integrate_remove_message', array($message));
 	}
 
+	// Update the first/last message times in the original topic
+	$smcFunc['db_query']('', '
+		UPDATE {db_prefix}topics AS t
+			INNER JOIN {db_prefix}messages AS mf ON (t.id_first_msg = mf.id_msg)
+			INNER JOIN {db_prefix}messages AS ml ON (t.id_last_msg = ml.id_msg)
```
albertlast: pg

Sources/RemoveTopic.php
```diff
@@ -1409,13 +1433,17 @@ function mergePosts($msgs, $from_topic, $target_topic)
 
 		// Update the topic details for the source topic.
 		$smcFunc['db_query']('', '
-			UPDATE {db_prefix}topics
+			UPDATE {db_prefix}topics AS t
+				INNER JOIN {db_prefix}messages AS mf ON (t.id_first_msg = mf.id_msg)
+				INNER JOIN {db_prefix}messages AS ml ON (t.id_last_msg = ml.id_msg)
```
albertlast: pg

Sources/RemoveTopic.php
```diff
@@ -1442,13 +1470,17 @@ function mergePosts($msgs, $from_topic, $target_topic)
 
 	// Finally get around to updating the destination topic, now all indexes etc on the source are fixed.
 	$smcFunc['db_query']('', '
-		UPDATE {db_prefix}topics
+		UPDATE {db_prefix}topics AS t
+			INNER JOIN {db_prefix}messages AS mf ON (t.id_first_msg = mf.id_msg)
+			INNER JOIN {db_prefix}messages AS ml ON (t.id_last_msg = ml.id_msg)
```
albertlast: pg

Sources/RepairBoards.php
```diff
@@ -296,6 +296,18 @@ function fix_full_processing:
 						'newTopicID' => $newTopicID,
 					)
 				);
+
+				// Ensure the salvage topic has the correct first/last message times
+				$smcFunc['db_query']('', '
+					UPDATE {db_prefix}topics AS t
+						INNER JOIN {db_prefix}messages AS mf ON (t.id_first_msg = mf.id_msg)
+						INNER JOIN {db_prefix}messages AS ml ON (t.id_last_msg = ml.id_msg)
```
albertlast: pg

Sources/RepairBoards.php
```diff
@@ -429,6 +441,18 @@ function fix_full_processing:
 						1
 					);
 
+					// Ensure this topic has the correct first/last message times
+					$smcFunc['db_query']('', '
+						UPDATE {db_prefix}topics AS t
+							INNER JOIN {db_prefix}messages AS mf ON (t.id_first_msg = mf.id_msg)
+							INNER JOIN {db_prefix}messages AS ml ON (t.id_last_msg = ml.id_msg)
```
albertlast: pg

Sources/RepairBoards.php
```diff
@@ -574,9 +598,19 @@ function fix_full_processing:
 					)
 				);
 
-				updateStats('subject', $newTopicID, $txt['salvaged_poll_topic_name']);
-
+				// Ensure the poll's topic has the correct first/last message times
+				$smcFunc['db_query']('', '
+					UPDATE {db_prefix}topics AS t
+						INNER JOIN {db_prefix}messages AS mf ON (t.id_first_msg = mf.id_msg)
+						INNER JOIN {db_prefix}messages AS ml ON (t.id_last_msg = ml.id_msg)
```
albertlast: pg

Sources/RepairBoards.php
```diff
@@ -619,11 +653,14 @@ function fix_full_processing:
...
 
 				$smcFunc['db_query']('', '
-					UPDATE {db_prefix}topics
-					SET id_first_msg = {int:myid_first_msg},
-						id_member_started = {int:memberStartedID}, id_last_msg = {int:myid_last_msg},
-						id_member_updated = {int:memberUpdatedID}, approved = {int:firstmsg_approved}
-					WHERE id_topic = {int:topic_id}',
+					UPDATE {db_prefix}topics AS t
+						INNER JOIN {db_prefix}messages AS mf ON (t.id_first_msg = mf.id_msg)
+						INNER JOIN {db_prefix}messages AS ml ON (t.id_last_msg = ml.id_msg)
```
albertlast: pg

Sources/ScheduledTasks.php
```diff
@@ -471,6 +471,16 @@ function scheduled_daily_maintenance()
 			'oldLogins' => time() - (!empty($modSettings['loginHistoryDays']) ? 60 * 60 * 24 * $modSettings['loginHistoryDays'] : 2592000),
 	));
 
+	// Ensure all topics have valid first/last message times, in case a mod or something didn't do it properly
+	$smcFunc['db_query']('', '
+		UPDATE {db_prefix}topics AS t
+			INNER JOIN {db_prefix}messages AS mf ON (t.id_first_msg = mf.id_msg)
+			INNER JOIN {db_prefix}messages AS ml ON (t.id_last_msg = ml.id_msg)
```
albertlast: pg

Sources/SplitTopics.php
```diff
@@ -721,6 +724,19 @@ function splitTopic($split1_ID_TOPIC, $splitMessages, $new_subject)
 		)
 	);
 
+	// Ensure both topics have the correct first/last message times
+	$smcFunc['db_query']('', '
+		UPDATE {db_prefix}topics AS t
+			INNER JOIN {db_prefix}messages AS mf ON (t.id_first_msg = mf.id_msg)
+			INNER JOIN {db_prefix}messages AS ml ON (t.id_last_msg = ml.id_msg)
```
albertlast: pg

Sources/SplitTopics.php
```diff
@@ -1637,23 +1653,26 @@ function MergeExecute($topics = array())
...
-					id_member_updated = {int:current_user},
-					id_poll = 0,
-					approved = 1,
-					num_replies = 0,
-					unapproved_posts = 0,
-					id_redirect_topic = {int:redirect_topic},
-					redirect_expires = {int:redirect_expires}
-				WHERE id_topic = {int:old_topic}',
+				UPDATE {db_prefix}topics AS t
+					INNER JOIN {db_prefix}messages AS m ON (t.id_first_msg = m.id_msg)
```
albertlast: pg

Sources/SplitTopics.php
```diff
@@ -1715,6 +1734,18 @@ function MergeExecute($topics = array())
 		)
 	);
 
+	// Ensure the merged topic has the correct first/last message times
+	$smcFunc['db_query']('', '
+		UPDATE {db_prefix}topics AS t
+			INNER JOIN {db_prefix}messages AS mf ON (t.id_first_msg = mf.id_msg)
+			INNER JOIN {db_prefix}messages AS ml ON (t.id_last_msg = ml.id_msg)
```
albertlast: pg

Sources/Subs-Post.php
```diff
@@ -2411,10 +2417,13 @@ function approvePosts($msgs, $approve = true, $notify = true)
 	// ... next the topics...
 	foreach ($topic_changes as $id => $changes)
 		$smcFunc['db_query']('', '
-			UPDATE {db_prefix}topics
-			SET approved = {int:approved}, unapproved_posts = unapproved_posts + {int:unapproved_posts},
-				num_replies = num_replies + {int:num_replies}, id_last_msg = {int:id_last_msg}
-			WHERE id_topic = {int:id_topic}',
+			UPDATE {db_prefix}topics AS t
+				INNER JOIN {db_prefix}messages AS mf ON (t.id_first_msg = mf.id_msg)
+				INNER JOIN {db_prefix}messages AS ml ON (t.id_last_msg = ml.id_msg)
```
albertlast: pg

albertlast (changes requested):
- SQL Syntax for pg doesn't work
- recommandation use for mysql bigint without unsigned like for pg (than they got the same min/max range)

Sesquipedalian:
Thanks, @albertlast. 

1) Will the syntax you provided also work for MySQL? We need something that will work for both, of course.
2) ~We might need to find and fix some other queries as well. I modelled this query on another I found already in the code.~ Never mind. My memory failed me. :)

Sesquipedalian:
[Here](https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/MessageIndex.php#L290) is the query that I based mine on. Does it need to be changed for PG compatibility as well?

Sesquipedalian:
Instead of trying to find a syntax that would work for both MySQL and PostgreSQL, I've just changed my code so that it builds the query differently for each case.

Sesquipedalian:
> recommandation use for mysql bigint without unsigned like for pg (than they got the same min/max range)

I have no reason to disagree with you that this would be better. I chose to use unsigned INT for MySQL only because I wanted these columns to exactly match `poster_time` in the messages table. 

If you think a signed BIGINT column would be better for storing Unix timestamps, probably the best idea would be for you to submit a PR to change that in all the columns that store that kind of value.

albertlast:
> Thanks, @albertlast.
>
>    Will the syntax you provided also work for MySQL? We need something that will work for both, of course.
>    We might need to find and fix some other queries as well. I modelled this query on another I found already in the code. Never mind. My memory failed me. :)

there is no commone sql string which is supported by both
The only clean solution would be to create a update function db layer which do the magic.
But this would many effort i think,
so your solution could be better.

The bigint is the think,
i find this to many work to changes this in effery field (would by huge impact on updater scriper),
so only for new stuff we could look that we got the same technical limites on both side.

sbulen:
I admit I haven't been following this thread closely, but at face value, I'd say embedding the different syntaxes in the code strays from SMF coding standards.   

The purposes of the smcFuncs are to do exactly the magic described above - an abstracted DB function layer - you write the updates once with a call to smcFunc, & the differences are dealt with the different MySql & PG implementations (in Subs-Db-\*.php, DbPackages-\*.php & DbExtra-\*.php).  

I would look very closely at these existing functions first.  If a specific function is missing, then add it...  But I would not recommend embedding db-specific logic in there.   

My 2 cents...   It's possible I'm off base, I admit I'm still learning, but this sounded unusual given SMF's excellent library of functions for this very purpose...

Sesquipedalian:
I looked through the existing $smcFunc db-* functions, and I didn't find anything to do this job. One could in theory try to abstract it all into a db-update function, but as @albertlast rightly said, that would take a lot of effort to do correctly. Such a function would either need to be restricted to handling only simple updates (in which case it is hardly worth doing) or else be able to handle complex updates (and my knowledge of SQL is not nearly strong enough to create that).

sbulen:
Which part of the syntax isn't supported by both pg & mysql? I don't think I see it... Inner joins are supported by both...   That's the most obscure part in there that I see...

pg is a lot more strict, & tends to only support ANSI standards.   MySQL usually operates ANSI, but adds a lot of goofy stuff.

Sesquipedalian:
JOIN is allowed for SELECT in Postgres, but not for UPDATE. Instead, Postgres wants the additional tables listed after the FROM.

sbulen:
I see...

I think for the most part, updates are close enough & work with a 'db_query'.  But that's not what we're doing here.  Their 'update from' syntaxes are different.  I think that focusing on the 'update from', instead of trying to do the much broader 'update' will allow you to write a new smcFunc that will work.  

A very crude outline would look something like:

MySQL smf_db_update_from:
	UPDATE target_tbl
	INNER JOIN src_tbl ON join_expr
	SET set_expr

Postgresql smf_db_update_from:
	UPDATE target_tbl
	SET set_expr
	FROM src_tbl
	WHERE join_expr

DB-inspecific call from code:
	smcfunc['db_update_from'], (src_tbl, target_tbl, join_expr, set_expr)
		src_tbl = 'source'
		target_tbl = 'target'
		join_expr = 'target_tbl.join_col = src_tbl.join_col'
		set_expr = 'target_field = src_tbl.src_field'

(Sorry for the crude pseudo-code, I'm not at my home computer & I'm trying to do this mostly from memory.)

Don't do the broader 'update', which I agree over-complicates things.   Just a thought...

Sesquipedalian:
That might be feasible.

jdarwood007:
Is there a reason we have to call a join or from in a update?  Personally I have disliked this functionality.
For the upgrade query this would be fine.

For the SMF side, in createPost we currently hard code time().  We should change that into $msgOptions['poster_time'] and default that to time() if not specified.  Then the section that handles creating a topic or updating it can do so correctly.

sbulen:
I think the 'update from' can be far more efficient, especially where large tables involved.  E.g., if the source is 2,000,000 rows and the target is 20...   Or vice versa...   In that example, you should only need to retrieve 40 rows, 20 from each, not 2M+.  

In theory, you could do the same thing with a query => an array in php, followed by an update.  But would need to create an array with 2,000,000 rows in memory in the example above.  Or, another approach: strings of queries could be used to make it efficient (e.g., a query to produce the array of the 20, to drive a subsequent set of queries to get the other 20).   

The 'update from' is more straightforward.

sbulen:
Nicely done!

Sesquipedalian:
Thanks. :)

Sesquipedalian:
> For the SMF side, in createPost we currently hard code time(). We should change that into $msgOptions['poster_time'] and default that to time() if not specified. Then the section that handles creating a topic or updating it can do so correctly.

That was a good idea, @jdarwood007. Done. :)

Sources/Subs-Db-postgresql.php
```diff
@@ -922,4 +923,50 @@ function smf_db_escape_wildcard_string($string, $translate_human_wildcards = fal
...
+		return;
+
+	$from = array();
+	foreach ($joined as $join)
+	{
+		if (empty($join['name']) || empty($join['alias']) || empty($join['condition']))
+			continue;
+
+		$from[] = $join['name'] . ' AS ' . $join['alias'];
+		$where = (!empty($where) ? $where .' AND ' : '') . $join['condition'];
```
albertlast: would use brackets to differ between where and join condition
exp.
$where = "asd = 1 or bsf = 2";
$join['condition'] = "b.id = a.id"

result of the existing solution:
where asd = 1 or bsf = 2 and b.id = a.id -> could lead in a wrong result
with brackets:
where (asd = 1 or bsf = 2) and (b.id = a.id) -> safe

Sources/Subs-Db-mysql.php
```diff
@@ -920,4 +921,48 @@ function smf_is_resource($result)
...
+	if (empty($table['name']) || empty($table['alias']) || empty($set))
+		return;
+
+	$joins = array();
+	foreach ($joined as $join)
+	{
+		if (empty($join['name']) || empty($join['alias']) || empty($join['condition']))
+			continue;
+
+		$joins[] = 'INNER JOIN ' . $join['name'] . ' AS ' . $join['alias'] . ' ON (' . $join['condition'] . ')';
```
albertlast: please let the "inner" away this works fine without this word.
the inner thing is similiar to "the dead person is dead" (we know that he is dead we don't need call it twice.)
and the same with inner, we know that a inner because there is no left or right.

albertlast:
yeah looks like good approache.

Sources/Subs-Db-postgresql.php
Sesquipedalian: Good catch.

albertlast:
> I have no reason to disagree with you that this would be better. I chose to use unsigned INT for MySQL only because I wanted these columns to exactly match poster_time in the messages table.
>If you think a signed BIGINT column would be better for storing Unix timestamps, probably the best idea would be for you to submit a PR to change that in all the columns that store that kind of value.

Maybe i should this quesion before,
but why we use unixtimestemp instead of normal timestamp?
normale timestamp = date + time -> 09.03.2017 5:29

Sesquipedalian:
> Maybe i should this quesion before,
but why we use unixtimestemp instead of normal timestamp?
normale timestamp = date + time -> 09.03.2017 5:29

I don't know, but changing it now would probably be a large and difficult task.

Sources/Subs-Db-postgresql.php
```diff
@@ -922,4 +923,53 @@ function smf_db_escape_wildcard_string($string, $translate_human_wildcards = fal
...
+			continue;
+
+		$from[] = $join['name'] . ' AS ' . $join['alias'];
+		$where = (!empty($where) ? $where .' AND ' : '') . '(' . $join['condition'] . ')';
+	}
+	if (empty($from))
+		return;
+
+	// Postgres doesn't like prefixes on the columns to be set
+	$set = preg_replace('~\b' . $table['alias'] . '\.\b~', '', $set);
```
albertlast, line 1015: From which kind of prefix we talk here?

Sesquipedalian: Table name or alias prefixes. As it says in [the Postgres documentation](https://www.postgresql.org/docs/9.1/static/sql-update.html) regarding the `column` definition for `UPDATE` syntax:

> Do not include the table's name in the specification of a target column — for example, UPDATE tab SET tab.col = 1 is invalid.

albertlast:
okay i see that datetime is hard to use,
because design error of poster_time...
and many other places where unixtimestamp is used instead of datetime.

I only want to mention how to convert a unixtimestamp to datetime
mysql FROM_UNIXTIME(1447430881)
pg to_timestamp(1447430881)

BloberSh:
I may come late but what about this PR?

It could be good to sort topic by message time. 

Also are you talking about sorting it by last topic activity or just last created topic

I see it "Merged" but there is some commits after this. So what's the state of this?

Sesquipedalian:
> Also are you talking about sorting it by last topic activity or just last created topic?

Both.

joshuaadickerson:
The RSS posting bot should have gone to the end of the feed and then worked its way forward. As everyone else has said, the converter should reorder the posts and you should figure out how to forward broken links.

I can see a reason to backdate posts though. 

If you're going to do that, you should change all the sorting to use that. Not sure if I saw it, but then your indexes should be {date, id}. You sort by date and then use the id as a tie-breaker.

I don't think this big of a change is worth it for your tiny use-case.

Sesquipedalian:
> The RSS posting bot should have gone to the end of the feed and then worked its way forward. As everyone else has said, the converter should reorder the posts and you should figure out how to forward broken links.

That wouldn't make any difference. And the functioning of RSS posting bot is beside the point here, anyway. I only mentioned it to explain how I stumbled on this issue.

The issue is that when we say we are sorting chronologically, we should actually sort chronologically.

> I can see a reason to backdate posts though.
If you're going to do that, you should change all the sorting to use that.

That's what this PR does.

illori:
are we going to do something with this PR?

Sesquipedalian:
I hope so. But it isn't the sort of thing that I would want to merge unilaterally. I'm still waiting for some feedback from other official devs.

albertlast:
The key issue on this pr is,
that it not fix a smf issue.

But create workaround for parts which are not connected with smf,
also the reason that topic_id can not be changed because urls didn't work any more,
is not so good.

Because with use of smf the other parts of the url change any way, too -->  so old url never work after this migration.

Would be much more reason able, to add a col(by your "mod") with the old_topic_id so when you call
the old format url version that the mapping can be realized.

Sesquipedalian:
I still disagree with your interpretation of the situation, @albertlast. The issue is not caused by the mod in question. The mod is just how I discovered the issue. The issue instead is that when we claim to be sorting topics chronologically, we are actually sorting them by id.

Sources/RepairBoards.php
```diff
@@ -640,11 +595,13 @@ function fix_full_processing:
 						MIN(ma.id_msg) END ELSE
 					MIN(mu.id_msg) END AS myid_first_msg,
 					CASE WHEN MAX(ma.id_msg) > 0 THEN MAX(ma.id_msg) ELSE MIN(mu.id_msg) END AS myid_last_msg,
+					t.first_msg_time, t.last_msg_time, mf.poster_time AS first_poster_time, ml.poster_time AS last_poster_time,
```
albertlast, line 598: This col are missing in the group by part

albertlast (changes requested):

albertlast:
Please had in mind that we support mysql below 5.7 so you need maintain the complet list of columns in group by and
not only the unfunction dep.

Sesquipedalian:
I'm closing this PR now. I would have liked to see this change make it in, but if we are trying to get a release candidate out, it is too late. Maybe in SMF 3.0

sbulen:
Sorry about that.   A lot of work...    

Probably a good call for the reason you stated.

Sesquipedalian:
Yeah, I actually did a bunch more work that I never pushed up to GitHub, in order to handle splitting and merging, tracking unread messages, etc., properly using timestamps rather than ID numbers. It's nearly done, as a matter of fact, but as I looked at it I realized that it had turned into a significant overhaul—too much for this stage in the game.

I may turn it into a mod for the 2.1 branch. It won't be very interesting for end users by itself, but it will make things like importing, scheduled posts, backdating, etc., possible for other mods to implement.

jdarwood007:
Thanks.  I agree it would be a big overhaul to do this.  Mostly because everything in SMF is designed around the id rather than the timestamp.

As well this could cause unforeseen problems as timestamps are known for being wrong. Case in point, the server has the wrong time, the admin fixes it up (and or gets ntp working) and now the time is drifting to correct itself and messages are now sorted wrong.  At least with id based sorting, even if the time is wrong, the messages are in the correct order.

PR #3954, opened 2017-03-03
Set displayFields upon install

sbulen:
displayFields wasn't initialized, which meant custom_fields were not displayed upon install.

To address the issue: Skype icon not shown on poster information #3909

Tested out OK on mysql.  I don't have a pg database to test with, but updates were straightforward.


frandominguezl:
That's it, it's working. Thanks :)

colinschoen (approved):
LGTM, approval

Issue #3919, opened 2017-02-05
Normal member can see hidden ones in the online list


realdigger:
![hidden](https://cloud.githubusercontent.com/assets/1187218/22630164/f02671c2-ec0d-11e6-8bbe-25b89beb9ddb.png)

Oldiesmann:
Were you logged in as a regular member at that time? Only admins and users with certain admin permissions should be able to see hidden members in that list

realdigger:
We have some reports about this issue from regular members. I test this from regular member account and guest. Both of them can see my hidden admin account online.
I checked some forum settings and found out that this only happens when using Memcached as caching accelerator.

realdigger:
Same issue with file based caching.

illori:
do they have moderator forum members permission? as this grants the access to view users that hidden.

i tested this on my 2.1 install and cant duplicate the issue if the user viewing the online list does NOT have moderate forum members permission.

frandominguezl:
Can't duplicate the issue

sbulen:
Couple of questions - 
(1) Do all hidden IDs ALWAYS appear when cache is enabled?  
(2) Are you changing settings (e.g., permissions) and retesting?   
(3) Is this a migrated forum, where you just completed a 2.0 => 2.1 upgrade?  

Knowing a little more about the exact steps may help.  This could be that some cache needs invalidation upon permission changes or cache changes, etc.   OR..  It could also be that we shouldn't place hidden members' info in cache.

sbulen:
Confirmed.  Yes, hidden users display when caching is enabled.  I'll take a peek at this...

sbulen:
You know...  I don't think this should be cached at all.  Cache doesn't appear to be unique per user...   So...  Once cache is built, it's used by whoever logs on next...    At first I thought that hashed key on the cache file name might include a user ID, but it doesn't appear to (and its set by the cache api).   

The solution here appears to be to simply not use cache for members online.  

If I'm off base, someone let me know...

colinschoen:
Fixed with #3956.

PR #3956, opened 2017-03-05
Do not cache members online data, it's unique per user

sbulen:
Fixes Normal member can see hidden ones in the online list #3919


sbulen:
Problem here is that if cache is built by an admin, who can see hidden users, then that is placed into cache.  The next users logging on see this info, including info for hidden users.  
Thus, this shouldn't be cached.

sbulen:
Also note that buddy info is embedded in there as well.

colinschoen (approved):
LGTM, approval

Issue #3957, opened 2017-03-07
Attachments - discarded upon preview


sbulen:
Attachments discarded upon preview.  To reproduce:
1. Add a new message
2. Type in something
3. Attach some photos
4. Upload attachments
5. Preview 
6. Post

If you click on the 'insert' button after uploading, you will get the message: "No message is associated with this attachment" in the message body after pressing Post.

![image](https://cloud.githubusercontent.com/assets/23568484/23640333/c118d53c-02a1-11e7-9d61-12e7bd797e17.png)

sbulen:
The cause of this was my prior PR:  Clear out attachment info for new posts #3879

I need to be more selective about clearing out the $_SESSION attachment info.  

Will need to look a little deeper...

sbulen:
The intent of the original logic was to restore previously uploaded attachments to $context.   There were issues with this, mainly that it only properly restored one of the attachments and discarded the others (leaving orphans in the file system).   

So...    Resume attachment processing or start afresh whenever the user presses Post & attachments are found partially processed in  $_session?   

I'm now leaning towards fixing the old attachment restore logic.

Input welcome...

sbulen:
To recap, since this info is spread across a couple of issues...

Imagine you author a post, and provide - and upload - multiple attachments.  Now imagine you navigate away before pressing Post (possibly on purpose, possibly on accident, the app doesn't know your motives...).  

What should happen the next time you press Post (or Reply)?   Bear in mind, you *might* be trying to recreate the post you just lost.  OR, you *might* be on a different board/topic entirely, starting a new message.  Again, the app doesn't know your intent.   

At this point in time, your previous message text is lost.  HOWEVER...  Oddly, uploaded attachments are tracked.  

The original logic here tried & failed to restore the attachments to $context, so they would be attached to your current post.  It failed, and only attached the first attachment to your post, and left the rest as orphans in your DB & filesystem.  

I thought it was odd to have one of your old attachments auto-attached to a new, blank, message.  It felt random.  My previous PR cleared those out.  It feels much cleaner to start afresh.  

(It turns out that PR inadvertently affected previews, & needs fixing.  Hence this issue.)

My only reservation with clearing everything out is that it is quite possible the attachments were large, and compiling them and attaching and uploading them could have been a chunk of work.  Re-presenting them would give the author the opportunity to delete them purposefully or re-use them.   

So...

Start afresh?   Or re-present the previously uploaded attachments on a new post?

[...]

PR #3960, opened 2017-03-11
Clean up prior attachment info when starting new message

sbulen:
Fixes: Attachments - discarded upon preview #3957

Needed to be a bit more selective when clearing out old $_SESSION attachment info.


colinschoen (approved):
LGTM, approval

Issue #3928, opened 2017-02-10
Security edits on .jpgs


sbulen:
In 2.1, I have received invalid security warnings uploading .jpgs.  Someone also reported this on the support site for 2.0 the other day:
http://www.simplemachines.org/community/index.php?topic=552034.0

I traced my 2.1 issue to the checkImageContents call in subs-graphics.    It's finding a hit on the preg_match, even the lighter one *without* extensive security checks selected.    The 2.0 issue reported that option was de-selected as well.

My jpgs were taken on a nikon, and edited in the latest version of lightroom.

sbulen:
I've been doing some reading on this, & I think that checkImageContents may need a new set of edits.   In my case, the preg_matches are returning false positives, and there are probably better ways to do this today.  

A helpful read - especially the response that says all uploaded images are evil:
http://stackoverflow.com/questions/6484307/how-to-check-if-an-uploaded-file-is-an-image-without-mime-type

My first thought is to change our minimal validation to use getImageSize(), and the extensive validation to use imageCreateFromString().   

This one may be important, as it involves security and it appears to be impacting 2.0.  

I'd appreciate feedback before diving in too deep here.

tinoest:
I've tended to use the following, if its any use at all. mime_content_type is just to catch when getimagesize parses an file as a image when it shouldn't do so. You could ensure that the first imageType matches the second from getimagesize if you wanted to be really cautious. 



       function check_image($filename) {

        $imageType      = mime_content_type($filename);
        if(!in_array($imageType, array('image/gif', 'image/jpeg', 'image/tiff', 'image/bmp', 'image/png'))) {
                return false;
        }               

        $imageData      = getimagesize($filename);
        if(!is_array($imageData)) {
                return false;
        }

        if( $imageData[0] == 0 ) {
                return false;
        }

        if( $imageData[1] == 0 ) {
                return false;
        }

        $imageType      = $imageData[2];     
        if(!in_array($imageType , array(IMAGETYPE_GIF , IMAGETYPE_JPEG ,IMAGETYPE_PNG , IMAGETYPE_BMP))) {
                return false;
        }

        return true;
       }

sbulen:
Thanks, that's very helpful.  And consistent with the other things I've read.

sbulen:
There is another current thread out there in 2.0 support with the same issue:
http://www.simplemachines.org/community/index.php?topic=544243.0

sbulen:
More reading on sanitizing images:
[...]

PR #3961, opened 2017-03-12
Clean up old security checks

sbulen:
Our string search mechanism to find suspicious code in attachments is highly problematic, and rejects an extremely high percentage of valid photos.  

The existing security checks fall into 3 categories:
  -  <% and <? are the short form of inline php & js; these reflect older attack vectors that should be locked down by most hosts today
  -  [CFZ]WS is part of the pattern for inline flash
  -  flash, script, html, eval, etc - are paranoid checks looking for suspicious terms.  Some are tied to old forms of IE attacks, some I couldn't find any documentation for at all as valid, current vulnerabilities

In this PR, for the paranoid check, I left in all of the old checks.  I would suggest using this security option ONLY if you have extreme mistrust in your ISP's security config.

For the non-paranoid check, I was torn between removing it entirely, or leaving in minimal php & flash checks.  I eventually opted for keeping the php & flash checks, but making them more robust to significantly reduce false positives.  

SMF photo sharing sites have significant issues due to these edits in 2.0 and will continue to have issues in 2.1 without this change.  If approved, I suggest pushing these changes back into 2.0 as well.   

Discussion threads, for background:
  -  https://github.com/SimpleMachines/SMF2.1/issues/3928
  -  https://github.com/elkarte/Elkarte/issues/2874

This is a short-to-mid-term solution.  Long term, we should find a better approach, since ANY string search technique will return false positives sooner or later.  The goal should be to only filter out truly infected photos, not random photos...

### DIFF
--- a/Sources/Subs-Graphics.php
+++ b/Sources/Subs-Graphics.php
@@ -211,8 +211,9 @@ function checkImageContents($fileName, $extensiveCheck = false)
 		// Though not exhaustive lists, better safe than sorry.
 		if (!empty($extensiveCheck))
 		{
-			// Paranoid check. Some like it that way.
-			if (preg_match('~(iframe|\\<\\?|\\<%|html|eval|body|script\W|[CF]WS[\x01-\x0C])~i', $prev_chunk . $cur_chunk) === 1)
+			// Paranoid check.  Use this if you have reason to distrust your host's security config.
+			// Will result in MANY false positives, and is not suitable for photography sites.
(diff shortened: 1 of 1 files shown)


sbulen:
Some example threads of users with issues in 2.0:
http://www.simplemachines.org/community/index.php?topic=552034.0
http://www.simplemachines.org/community/index.php?topic=544243.0

I believe it is related to the photo upload "white page" issue...  Some of my pix that fail erroneously in 2.1 cause white page failures in 2.0 for that thread.  I don't think the post completes successfully:
http://www.simplemachines.org/community/index.php?topic=547585.msg3884517#msg3884517

Issue #3962, opened 2017-03-13
Upgrader - remnants of old themes around, prevents package install


sbulen:
Just tried testing an install of a package, and received the following error:  
![image](https://cloud.githubusercontent.com/assets/23568484/23839990/00fa3d14-075f-11e7-8a47-e7aed5d554ec.png)

It appears there are references to old themes in the themes table.  The package cannot be installed because it cannot make a backup first.  It can't make the backup, because the core theme directory doesn't exist in 2.1.

I've confirmed the old entries are in the themes table...  At this point, I am assuming that's where the reference to core comes from in the above message.

sbulen:
Confirmed, backups first try to backup your themes...  I believe the upgrader should clear out all the old theme data.  This is the applicable code in subs-package.php:
![image](https://cloud.githubusercontent.com/assets/23568484/23873724/3a2a1f0a-07f0-11e7-9aa4-f822cd6fe5c2.png)

sbulen:
Note that the Upgrader already attempts to clean these up, but I see issues with its current approach: 
 - It only cleans up core
 - It only cleans up core if it still exists on the file system
 - It assumes all other theme data should be left alone in the themes table

I believe we are now working under the premise that the themes are going to be wiped out as part of the upgrade.  2.0 themes are not guaranteed to work in 2.1.  We will ask folks to get rid of their old Themes folder as part of the upgrade process.  Thus, the old themes, including core, should _not_ be found in the file system.  

So...  

I think that the upgrader logic should be changed to wipe out all details for _all_ themes except the new default theme.  This cleanup should not be dependent on what it finds in the file system.  

Feedback welcome.

illori:
2.0 themes will never work on SMF 2.1. but i have yet to see it advised to user to delete their old themes at the time of the upgrade. also some may not delete core as it is a theme they did not install in the first place if they are told to delete themes.

sbulen:
@illori :  We need a common vision for 2.0 => 2.1 migration, in order to know what to build to.  

The only option that I see that makes sense is the one I have been told from the beginning, when I asked how to help out on the beta:  Delete all folders except attachments, avatars & settings.php.  Copy new files over & run the upgrader.

[...]

PR #3964, opened 2017-03-14
Clean up old theme data

sbulen:
Fixes: Upgrader - remnants of old themes around, prevents package install #3962

Users must start afresh with themes upon upgrade due to the significant differences between 2.0 & 2.1.  This fixes the issue where old theme settings were left around in the themes table, and were reflected in the knownThemes settings in the settings table.  

The extra old theme settings were, for example, causing backups to fail, since references to old theme folders were found in the themes table, & the backup process iterates those when making backups.  

Test out OK, and fixes all issues reported in #3962 

I don't have pg, I simply copied the MySQL stuff over, I believe it's simple enough it should be supported.

### DIFF
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -924,13 +924,15 @@ foreach ($toMove as $move)
 /******************************************************************************/
 --- Cleaning up after old themes...
 /******************************************************************************/
----# Delete all theme data except for the current default
+---# Disabling old themes
 ---{
 $smcFunc['db_query']('', '
 	DELETE FROM {db_prefix}themes
-	WHERE id_theme != {int:default_theme}',
+	WHERE id_theme != {int:default_theme}
+	  AND variable IN ({array_string:remove_settings});',
 	array(
 		'default_theme' => 1,
+		'remove_settings' => array('theme_dir','theme_url'),
 	)
 );
 ---}
@@ -941,7 +943,7 @@ $smcFunc['db_query']('', '
 $smcFunc['db_query']('', '
 	UPDATE {db_prefix}settings
 	SET value = {string:default_theme}
-	WHERE variable = {string:known_themes}',
+	WHERE variable = {string:known_themes};',
 	array(
 		'default_theme' => '1',
 		'known_themes' => 'knownThemes',
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -924,29 +924,45 @@ foreach ($toMove as $move)
 /******************************************************************************/
 --- Cleaning up after old themes...
 /******************************************************************************/
----# Disabling old themes
+---# Clean up settings for unused themes
 ---{
-$smcFunc['db_query']('', '
-	DELETE FROM {db_prefix}themes
-	WHERE id_theme != {int:default_theme}
-	  AND variable IN ({array_string:remove_settings});',
+// Fetch list of theme directories
+$request = $smcFunc['db_query']('', '
+	SELECT id_theme, variable, value
+	  FROM {db_prefix}themes
+	WHERE variable = {string:theme_dir}
+	  AND id_theme != {int:default_theme};',
 	array(
 		'default_theme' => 1,
-		'remove_settings' => array('theme_dir','theme_url'),
+		'theme_dir' => 'theme_dir',
 	)
 );
----}
----#
-
----# Setting knownThemes to current default
----{
+// Check which themes exist in the filesystem & save off their IDs 
+// Dont delete default theme(start with 1 in the array), & make sure to delete old core theme
+$known_themes = array('1');
+$core_dir = $GLOBALS['boarddir'] . '/Themes/core';
+while ($row = $smcFunc['db_fetch_assoc']($request))	{
+	if ($row['value'] != $core_dir && is_dir($row['value'])) {
+		$known_themes[] = $row['id_theme'];
+	}
+}
+// Cleanup unused theme settings
+$smcFunc['db_query']('', '
+	DELETE FROM {db_prefix}themes
+	WHERE id_theme NOT IN ({array_int:known_themes});',
+	array(
+		'known_themes' => $known_themes,
+	)
+);
+// Set knownThemes
+$known_themes = implode(',', $known_themes);
 $smcFunc['db_query']('', '
 	UPDATE {db_prefix}settings
-	SET value = {string:default_theme}
-	WHERE variable = {string:known_themes};',
+	SET value = {string:known_themes}
+	WHERE variable = {string:known_theme_str};',
 	array(
-		'default_theme' => '1',
-		'known_themes' => 'knownThemes',
+		'known_theme_str' => 'knownThemes',
+		'known_themes' => $known_themes,
 	)
 );
 ---}
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -924,45 +924,29 @@ foreach ($toMove as $move)
 /******************************************************************************/
 --- Cleaning up after old themes...
 /******************************************************************************/
----# Checking for "core" and removing it if necessary...
+---# Delete all theme data except for the current default
 ---{
-// Do they have "core" installed?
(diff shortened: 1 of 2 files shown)


jdarwood007:
This affects inter-upgrades between 2.1 Beta 2 and 2.1 Beta 3 as well though.  Which if the theme supports the new beta, would uninstall it.

This also has a side effect of removing a themes settings.  So if all is needed is more current files, then the admin is required to setup any changes again.

sbulen:
Understood.  

But the _primary_ purpose of the upgrader is to help move folks from 2.0 to 2.1.   If we leave all the old 2.0 settings out there, problems happen.  (In this case, old 2.0 theme data prevented package installation.)  We must clear out the old 2.0 settings.  

And the upgrader cannot tell the difference between a 2.0 theme setting & a 2.1beta theme setting. 

Yes, I think this means that a beta => beta upgrade will require re-application of themes.   I think they need to re-install mods also.   

We cannot make decisions based on supporting beta-to-beta migrations _at the expense of supporting 2.0 to 2.1 migrations_.  Therein lies madness - in the form of a neverending beta!

jdarwood007:
However, with us using more hooks, technically we can support not stripping all mods and may not need to strip them if they are compatible.

Maybe this.  Move this to a setting during upgrade setup and call it something like "Cleanup themes".  Would be a option based setup with 3 options.

- Remove unsupported themes and settings (Default)
- Remove all non-default themes and settings
- Do Nothing.

We could then add to the theme_info.xml the ability to say a SMF version range (like we do for customizations) and the upgrader can then identify if the theme supports the version we are upgrading to or not to leave those themes alone.  If the folder doesn't exist, the theme_info.xml doesn't exist or doesn't have the version detection, we treat it as unsupported.

We would want to make sure during CLI we can specify this option.  The default option would be the safest method of trying to detect if we can support that theme or not.

albertlast:
In my eyes is this requirment wrong,
i provide a partial update logic this was refused because beta to beta upgrade is not supported.

And now every upgrade step had to solve this issue by his own.
When you(as smf) like to got the possibility to upgrade beta to beta,
than provide centrtal function to controll the flow.

At the moment i didn't see anything --> the requirement didn't exists.

sbulen:
I'm with Albert on this.  

4 years ago, I think that would have been an excellent idea.  This late in the beta process, I think we should be very selective when introducing new approaches & significant enhancements.  It's time to stop moving the goalposts.  

I'm not sure there is much of a problem anyway, since I don't think there are any 2.1 themes out there.

jdarwood007:
Well there isn't any public 2.1 themes out there.  Antes and I are working on the new theme for the website and there is lots of theme setting data in it to make it work properly in both production and development environments (due to url and path differences).  I upgrade my test smf instance now and then along the development process and would be very annoying if I have to keep reinstalling the theme and adjusting those theme settings again.

If this is an issue for #3962, then we could fix the code that handles the backups to ignore themes under certain conditions (such as missing) and not throw errors.

sbulen:
It is far more preferable to build towards a solution that completely removes the 2.0 themes & mods - that we already know won't work - as part of the upgrade process.  

Leaving all those old settings & code around is dangerous.  I believe you will be dealing with prod issues for years due to dirty old code & settings laying around like little time bombs.  

_We already know 2.0 themes & mods won't work...  Why on earth would we want to keep them around, only to have to code around them?_

jdarwood007:
Because a theme or mod developer can upgrade their customizations and they will work again.  Just as I have done for all the mods and site theme we use for SimpleMachines.  No settings are lost so nothing needs reset, but it works as soon as they are installed.  With hooks, I may not even need to worry about reinstalling them if the files become fresh again.

sbulen:
Yes, they "can"...   But will they?   I think half of my 2.0 mods are abandoned by their authors at this point.  How many mod authors are going to rewrite for 2.1?  I'd be shocked if more than 25% do.   If they do, they will likely leverage new techniques and not write the same mod the same way.   Meanwhile, we're going to have 75% of odd code laying about...  

I do think it's a good bet mod authors will re-use settings & DB tables wherever possible.  

But I worry about incompatibility on almost every other front.  

Like...   Language tables...   We're going to get new language tables.   How will that work?   Delete the old ones & overwrite?  Mods will definitely have to be reinstalled then.  

Packages...   Do we keep the old package info around?   Will it still work, despite all the differences in 2.0 to 2.1?   What about errors (e.g., with languages) in deinstall?   Do we have to start telling people to ignore errors in the package manager?   

I just sense there are enough differences that we should make a clean break.  It works today, and will not require onerous, complex new design & development (that is not spec'd out yet & that I bet nobody will agree on...)

The mods and themes you want to keep will almost certainly need new installs to fully work properly anyway.

sbulen:
How did 1.x to 2.x work?   Any lessons to be learned from that?

jdarwood007:
All previous upgrades (and not just 1.x to 2.x, yabbse is/was supported), mods would be removed by the core nature of having to replace files.  The settings/columns they left didn't get removed.  Reinstalling them would have the mods do whatever they did, which in most cases they would start working again if they used the same settings.

Themes for 2.0 it looks like we just ignored them.  I do remember some support cases where we had to tell people to either upload a new theme or add ?theme=1 to the url to get to the default to change theme options.  We did inject the "core" theme (1.1 default theme) and transfer all themes that had settings telling us they where based on core to the core theme to fix them up.  Themes based on core could potentially have worked again.

We did have at one point code to try and convert themes over to be supported.  I assume this was mostly fixing up function names and any variables that got renamed.  Doesn't look like it made it into the upgrader.

illori:
we already do a version check on installing themes in SMF 2.1, so that part is already covered. this in part was added to prevent installing old themes that will not work with SMF 2.1.

sbulen:
OK...  I'll rework & test this PR to remove the part where I delete all the settings.   

To fix the backup, two approaches come to mind:
  -  jdarwood's excellent suggestion above to confirm the directory exists before backing it up
  -  Narrowing the settings deletion to remove only the theme's 'path' settings

I'm leaning towards removing the path settings, just in case there are other functions that use similar logic, for example:
  -  The normal backup function
  -  Repair settings
  -  Applying theme mods during theme installation... It may think the old items are still around...

It just feels a little safer to remove the paths.  Feedback welcome.

sbulen:
Let me know if you think this approach will work.

sbulen:
Not directly related, but there is a lot of interesting information in Arjen's 3/23/17 post in this thread:
Theme Error: Database error, given array of string values is empty. (query_where) #3982

That is a real-world example of an admin approaching the upgrade.   There is actually a lot to unpack there:
  -  Like me, he cleared out folders in advance
  -  Like me, he was surprised to find theme references in there
  -  He hit the Delete button in his test environment, and it _deleted the theme on his main site!!!!_   This is because the old theme_dir setting is there upon upgrade, still pointing to where the DB was restored from...  This is a potential hazard for anybody who places their test environment alongside prod.  

I don't think this helps with the ultimate resolution, just more background to chew over.   If anything, this made me want to go back to my original push here, which was to delete all theme info other than the default.  I tend to lean towards making things as bulletproof as possible.  I know there are other perspectives, though.

jdarwood007:
Incorrect urls/paths is outside of the upgrader.  That is what we have repair_settings.php for.  In a normal environment, this wouldn't be a problem.  A true separate testing server would also prevent this.

sbulen:
I think I have a solution that will make us all happy:  Delete theme settings for themes that don't exist in the file system.   If the theme_dir exists, don't delete the theme settings.  This handles all the scenarios: folks cleaning up the folder before upgrade AND folks leaving themes around.  

Let me test this idea & submit a new PR if it works out OK.  

Sound like a good compromise?

jdarwood007:
I think that resolves it.

sbulen:
I think it's ready.  Backups work.  No stray theme settings.  Allows folks to clear out Themes before upgrade & start afresh.  Allows beta-to-beta updates.  

I believe this even resolves:
Theme Error: Database error, given array of string values is empty. (query_where) #3982

sbulen:
One caveat:  I don't have a pg environment to test...

albertlast:
I see no bad stuff for pg in your changes.

colinschoen (approved):
LGTM, approval

Issue #3963, opened 2017-03-13
package_path lingers


sbulen:
I believe $modSettings['package_path'] was an old smf_settings entry that held the packages path at one point.  I have an old $modSettings['package_path'] entry in my 2.0 forum.  

As near as I can tell, package_path was replaced by $packagesdir (in settings.php) & should no longer be used.  However, it is still referenced in code in Subs-Package.php.   

So...  If my understanding is correct, we should:
  -  Clean up that old logic from subs-package.php
  -  Add an entry in the upgrader to delete the old package_path setting

One example, it tries to set the ftp path using it here:
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs-Package.php#L873

sbulen:
This looks like a very easy change, there is only one question in my mind...

For the most part, this is an easy substitution in the code from $modSettings['package_path']) (which was loaded from the settings table) to $packagesdir (now loaded from settings.php). 

However...  In Subs-Package.php, when a valid FTP path is determined, it then issues an updatesettings to update 'package_path'.  I think this is/was questionable...  Updating the settings might make sense if there were a separate setting for the ftp path somewhere; it just doesn't feel right setting the packages path to it, as this feels like it might be disruptive to packages processing.  

I also simply have issues updating critical settings in the background with the user unaware...

So, my recommendation is to use $packagesdir in its place, however, NOT to issue an updateSettings/updateSettingsFile to overwrite the configured packages directory with the FTP directory as the old logic suggests.  

Feedback welcome...

jdarwood007:
I introduced this setting.  It's purpose was to allow the packages directory to be moved, thus allowing more secure configurations by moving this outside of the web accessible location.  A common host setup is to need 755 or 777 on the packages directory to upload customizations and handle the temp directory during their install/uninstall.

It sounds like the modSetting was a older usage for detecting the correct path to be working in for a package.  I would have to dive deeper into its meaning, but I think removing it wouldn't break anything.

sbulen:
Thanks.

I ran across this while looking into a package install issue.  It was a holdover from an earlier version, and had a strange partial path in it.  Reading the code, it looks like it would cause havoc for ftp users under certain circumstances.

jdarwood007:
[...]

PR #3966, opened 2017-03-15
Remove old unused package_path setting

sbulen:
package_path is an legacy setting, but it still appears in subs-packages, and is populated with (old) path info in my 2.0 forum.   Replaced instances of package_path with its current version, $packagesdir, and removed the setting in the upgrader.  

Addresses: package_path lingers #3963

Testing, I was able to confirm the old setting is gone, & that package installation & de-installation still work fine.  

This may have addressed problems with FTP paths - it would be helpful if an FTP user could test that out.


sbulen:
I'm thinking I should leave it be & withdraw the PR. Package_path does seem to have a purpose, and you have explained the funky path - it's not the dangerous broken stray setting I thought it was.

PR #3967, opened 2017-03-15
[Fixes #3965] Stops sending notifications regarding ignored users

### LINKED ISSUE #3965: Members recieve notification alerts about ignored users.
I recieve notification alert when ignored user quote my post. I think this is wrong behavior.

colinschoen:
Adds check to verify we aren't sending a notification if the user who updated a topic is on the ignore list. 

Partially-Fixes #3965 


sbulen:
I think the user noted that there were others as well, e.g., Likes.   I'm certain Likes-Notify needs similar treatment.  

I don't know how many other scriptlets in sources\tasks need similar treatment, Birthdays?  Events?

colinschoen:
Ah ok, I thought this was exclusive. I'll do a deeper dig and push another PR for those.

colinschoen:
Can I get a LGTM or review for just this one case from anyone here prior to merging, please.

colinschoen:
Self-approving...

PR #3968, opened 2017-03-17
Normalize query in sub boardindex

albertlast:
Define the postgresql solution as default one,
mysql can handle this kind of query too

### DIFF
--- a/Sources/Subs-BoardIndex.php
+++ b/Sources/Subs-BoardIndex.php
@@ -44,7 +44,7 @@ function getBoardIndex($boardIndexOptions)
 		);
 
 	// Find all boards and categories, as well as related information.  This will be sorted by the natural order of boards and categories, which we control.
-	$result_boards = $smcFunc['db_query']('boardindex_fetch_boards', '
+	$result_boards = $smcFunc['db_query']('', '
 		SELECT' . ($boardIndexOptions['include_categories'] ? '
 			c.id_cat, c.name AS cat_name, c.description AS cat_desc,' : '') . '
 			b.id_board, b.name AS board_name, b.description,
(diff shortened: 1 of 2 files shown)


jdarwood007:
Do we know the performance impact of changing a query like this?  I worry about changing queries like this mostly for big board users.

albertlast:
They should run the same,
but you got smf.org  feel free to test vs your data and give me feedback.

jdarwood007:
Yes we could do that and see what would happen.

What is interesting is all your changing is wrapping a CASE around the COALESCE.  If that is all and MySQL supports it just fine, then why didn't we do that natively in 2.0 rather than do that workaround.

The workaround leads me to believe either a performance issue or version compatibility reasons.  Weirdly enough, trying to search sm.org for boardindex_fetch_boards seems to give me lots of false positives.  Seems like the search engine isn't doing something right.

albertlast:
the old and new version are nativ,
the new version is working on mysql and pg.

the old version is only shorter from writer style,
but internaly the database need to the same.
-> this is no workaround

main motivation is to get the same result under mysql as under pg

jdarwood007:
Yes I agree.

I went and looked into the SVN code and it doesn't tell us why other than this was first added when we initially added support for PostgreSQL across the codebase.  The developer comment was adding initial support for postgresql and changed basically the entire codebase to do that.

albertlast:
When you compare the beta 2 version with the git version: https://github.com/SimpleMachines/SMF2.1/compare/v2.1-beta.2...release-2.1#diff-122ecf81f7a41384b7e471f58fe5a3a0L275
than you see that i did already many times,
define the postgresql query as default one for both.

That is the reason that for me normal to do so.

jdarwood007:
I'm not questioning your work.  I'm just questioning why it was done that way originally if there was support to do it your way originally.  I can't find any information onto as why.  It was part of the original code translation to support PostgreSQL, so leaving it at that.

colinschoen:
Agreed, that is an odd. This is better, but I'm a bit suspicious this may have been deliberate too. I'll do some digging.

albertlast:
@colinschoen any reason to hold this back and not merge?

colinschoen:
I haven't been able to find anything so at this point no. I'll do one final dig and then merge tomorrow assuming no issues.

PR #3970, opened 2017-03-17
Calendar birthday query with index (pg only)

albertlast:
With my test board 2k user
I reduce the run time from 0,004 sec to 0,0006 sec


albertlast:
From my side this topic is done,
feel free to comment.

PR #3971, opened 2017-03-17
fatal_lang_error $log can be string or false

albertlast:
updated the doc
in Scrutinizer 289 issues are fixed


albertlast:
@colinschoen i think is uncritical which should easy to merge

colinschoen (approved):
Thank you

LGTM, approval

PR #3974, opened 2017-03-18
Fixed datatype check in ManageNews

albertlast:
Main idea is to protect here for possible security issues.


colinschoen (approved):
LGTM, approval

PR #3976, opened 2017-03-19
Do the mb4 fix only when needed

albertlast:
Postgresql support from default the complet utf8,
in mysql is this related to the used collation when you use the newer one (utf8mb4_general_ci).

utf8mb4 collation exists with mysql 5.5+ and is recommand to use instead of uf8
but we support 5.1 so the default value is false for mb4 support.

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -84,7 +84,7 @@ function reloadSettings()
 		{
 			return $string;
 		};
-	$fix_utf8mb4 = function($string) use ($utf8,$smcFunc)
+	$fix_utf8mb4 = function($string) use ($utf8, $smcFunc)
 	{
 		if (!$utf8 || $smcFunc['db_mb4'])
 			return $string;
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -84,9 +84,9 @@ function reloadSettings()
 		{
 			return $string;
 		};
-	$fix_utf8mb4 = function($string) use ($utf8)
+	$fix_utf8mb4 = function($string) use ($utf8,$smcFunc)
 	{
-		if (!$utf8)
+		if (!$utf8 || $smcFunc['db_mb4'])
 			return $string;
 
 		$i = 0;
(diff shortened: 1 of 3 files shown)


jdarwood007:
Is there any conversion/upgrade needed to worry about by using this collation?  Or is this a compatible collation?

albertlast:
utf8 is a subset of mb4,
so i would expect that are no issues.
see this link for more information: http://stackoverflow.com/questions/766809/whats-the-difference-between-utf8-general-ci-and-utf8-unicode-ci

sbulen:
MySQL's "UTF8" is, in fact, a misnomer.  For some reason, they only supported up to 3 bytes, not 4, of multibyte characters as true UTF8 should.  I.e., MySQL's UTF8 is not UTF8...   MySQL later started supporting UTF8MB4, which actually provides true UTF8 support.   

So everything is named incorrectly in the MySQL world.  There are some languages that are not fully supported in "UTF8".  MB4 fixes that.  

What I don't know is, whether all the places we have the string 'utf8' hard-coded throughout SMF will seamlessly support MB4.  I suspect this will need testing.  

MB4 may need more character space due to having another byte.  More described here:
https://dev.mysql.com/doc/refman/5.5/en/charset-unicode-conversion.html

albertlast:
when you new version (which is in beta stadium) even the mb4 didn't support "all" character,
that is the reason that mysql 8 cames up with utf8mb4_900 
http://www.dataarchitect.cloud/sushi-beer-an-introduction-of-utf8-support-in-mysql-8-0/
cache version:
http://webcache.googleusercontent.com/search?q=cache:pu0Z3T6LykAJ:www.dataarchitect.cloud/sushi-beer-an-introduction-of-utf8-support-in-mysql-8-0/+&cd=4&hl=de&ct=clnk&gl=de&client=firefox-b

albertlast:
@colinschoen is here any issue with this pr?

Sources/Load.php
```diff
@@ -84,9 +84,9 @@ function reloadSettings()
 		{
 			return $string;
 		};
-	$fix_utf8mb4 = function($string) use ($utf8)
+	$fix_utf8mb4 = function($string) use ($utf8,$smcFunc)
```
colinschoen: Super nit: 

($utf8, $smcFunc)

Otherwise LGTM

colinschoen (approved):
LGTM, approval

asmith20002:
Has this been solved? I'm using MySQL 5.7.18 with Php 7.1 and SMF 2.1beta 3.

Sometimes my I get this in error log: (the message is NOT inserted into db)
Incorrect string value: '\xF0\x9F\x92\x99' for column 'body' at row 1
Subs-Posts.php
line 1773


I changed the body column from utf8_general_ci into utf8mb4_general_ci. Does that fix the problem?

PR #3977, opened 2017-03-19
PostgreSQL Check for utf8 database charcode

albertlast:
In postgresql world is that default database (template1) utf8 codec,
but maybe someone somewhere doesn't use the default stuff
this code check this.

need to add a text message for error message


Oldiesmann:
Is this really necessary? We're already using SET NAMES to specify that anything we're sending is UTF-8, and PostgreSQL has good support for charset conversion.

albertlast:
utf8 got so many character which are normale charset doesn't got,
for example the smylies.
so when you set the client charset to utf8 and the database is latin,
the database loos because the can not transform the smylie into a latin symbole,

Like i saied the default template for every database is template1 and this
is utf8, but maybe someone did something wrong than he got problems.

PR #3978, opened 2017-03-19
Fix Security, Check the attachment id

albertlast:
we pipe on many places $_REQUEST['attach'] directly into getAttachmentFilename as attachment_id without testing if there only integer value.
so instead of change on everycall i place a check in the function


albertlast:
@colinschoen this too

colinschoen (approved):
LGTM, approval

PR #3979, opened 2017-03-19
Validation Class for smf

albertlast:
was looking for some good validation stuff in smf for exp email validation and
didn't found something usefull.
So i jump to elkarte adapt there solution for us. 

i'm not 100% if is all working,
but is this better as the existing one.

### DIFF
--- a/Sources/Class-Data_Validator.php
+++ b/Sources/Class-Data_Validator.php
@@ -1,0 +1,1318 @@
+<?php
+/**
+ * Simple Machines Forum (SMF)
+ *
+ * @package SMF
+ * @author Simple Machines http://www.simplemachines.org
+ * @copyright 2017 Simple Machines and individual contributors
+ * @license http://www.simplemachines.org/about/smf/license.php BSD
+ *
+ * @version 2.1 Beta 3
+ */
+if (!defined('SMF'))
+	die('No direct access...');
+/**
+ * Class used to validate and transform data
+ * based on https://github.com/elkarte/Elkarte/blob/development/sources/subs/DataValidator.class.php
+ *
+ * Initiate
+ *    $validation = new Data_Validator();
+ *
+ * Set validation rules
+ *    $validation->validation_rules(array(
+ *      'username' => 'required|alpha_numeric|max_length[10]|min_length[6]',
+ *      'email'    => 'required|valid_email'
+ *    ));
+ *
+ * Set optional sanitation rules
+ *    $validation->sanitation_rules(array(
+ *      'username' => 'trim|strtoupper',
+ *      'email'    => 'trim|gmail_normalize'
+ *    ));
+ *
+ * Set optional variable name substitutions
+ *    $validation->text_replacements(array(
+ *      'username' => $txt['someThing'],
+ *      'email'    => $txt['someEmail']
+ *    ));
+ *
+ * Set optional special processing tags
+ *    $validation->input_processing(array(
+ *      'somefield'    => 'csv',
+ *      'anotherfield' => 'array'
+ *    ));
+ *
+ * Run the validation
+ *    $validation->validate($data);
+ * $data must be an array with keys matching the validation rule e.g. $data['username'], $data['email']
+ *
+ * Get the results
+ *    $validation->validation_errors(optional array of fields to return errors on)
+ *    $validation->validation_data()
+ *    $validation->username
+ *
+ * Use it inline with the static method
+ * $_POST['username'] = ' username '
+ * if (Data_Validator::is_valid($_POST, array('username' => 'required|alpha_numeric'), array('username' => 'trim|strtoupper')))
+ *    $username = $_POST['username'] // now = 'USERNAME'
+ *
+ * Current validation can be one or a combination of:
+ *    max_length[x], min_length[x], length[x],
+ *    alpha, alpha_numeric, alpha_dash
+ *    numeric, integer, boolean, float, notequal[x,y,z], isarray, limits[min,max]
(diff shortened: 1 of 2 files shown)


emanuele45:
It's not my intention to be a pain, neither a stalker, but BSD has 3 requirements, and the first is:

> 1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.

Here the (c) in the original file has been lost, I guess a typo. :smiley: 

And, to be frank, I feel the @ author reporting SMF is *slightly* misleading.

albertlast:
To be ownest i don't understand the copyright stuff,
i see in elkarte many files which are look simliar to smf stuff,
but there are elkarte copyright without hint of smf... for exp. Errors.template.php

Well travis didn't like the old and new version copyright stuff...
need some advice...

emanuele45:
Everything in ElkArte started with the SMF headers along with the ElkArte ones. Then, for the code wrote entirely by ElkArte Contributors (e.g. Errors.class.php (the template has the SMF note) or the DataValidator class), of course the (C) was only of those that wrote the code from scratch.
Then there is code that still retains the SMF (C) notice, but is somehow difficult to match with anything still in SMF because it changed so much that the only SMF code left are probably few comparisons.

albertlast:
@jdarwood007 how should be licence file look like?

XinYenFon:
It should be reverse in this file.

@emanuele45 nah its cool, its your project's right to protect your copyrighted material.

jdarwood007:
If the intention is to get a single source for all validations, then we should make all code use this.  But I feel that may be a big step for 2.1 at this stage to want to tackle that.

The licensing check script isn't built to handle 3rd party/additional lines that don't conform to our normal licensing syntax.  For other 3rd party code, the licensing check script actually excludes those directories/files.  Our headers are not present anywhere in those sources.

albertlast:
The idea is to centralize the valid rule for everything,
here did elkarte a good job to create a good amount of rules.

but it's not target to convert now in this version all inputs to this kind of validatin.

Oldiesmann:
Need to look into this and find out why it's failing Travis checks, but Travis is down at the moment:

```
[michael@antergos ~]$ ping -c4 travis-ci.org
PING travis-ci.org (75.101.145.225) 56(84) bytes of data.

--- travis-ci.org ping statistics ---
4 packets transmitted, 0 received, 100% packet loss, time 3028ms
```

Sources/Subs.php
```diff
@@ -16,6 +16,8 @@
 if (!defined('SMF'))
 	die('No direct access...');
 
+require_once($sourcedir . '/Class-Data_Validator.php');
+
```
jdarwood007, line 20: We don't call stuff like this here.

If its needed at startup, should go in the index/ssi.php with the other startup require_once calls.  If another function will always call it to start it up, that function can do the require

Sources/Class-Data_Validator.php
```diff
@@ -0,0 +1,1326 @@
...
+	 * @param string $field
+	 * @param mixed[] $input
+	 * @param mixed[]|null $validation_parameters array or null
+	 */
+	protected function _validate_valid_url($field, $input, $validation_parameters = null)
+	{
+		if (!isset($input[$field]))
+			return;
+
+		if (!preg_match('`^(https{0,1}?:(//([a-z0-9\-._~%]+)(:[0-9]+)?(/[a-z0-9\-._~%!$&\'()*+,;=:@]+)*/?))(\?[a-z0-9\-._~%!$&\'()*+,;=:@/?]*)?(\#[a-z0-9\-._~%!$&\'()*+,;=:@/?]*)?$`', $input[$field], $matches))
```
Sesquipedalian, line 1102: Why is this using a regex rather than `filter_var($input[$field], FILTER_VALIDATE_URL)`?

Also, it might be valuable to add a `_validate_valid_iri()` function to handle user entered URLs that include non-ASCII characters. A simplified version of [this regex](https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs.php#L1886) should be helpful for that.

Sources/Class-Data_Validator.php
```diff
@@ -0,0 +1,1326 @@
...
+	 * @param string $field
+	 * @param mixed[] $input
+	 * @param mixed[]|null $validation_parameters array or null
+	 */
+	protected function _validate_valid_color($field, $input, $validation_parameters = null)
+	{
+		if (!isset($input[$field]))
+			return;
+
+		// A color can be a name: there are 140 valid, but a similar list is too long, so let's just use the basic 17
```
Sesquipedalian, line 1243: Why not use all of them? The full list is available at https://www.w3schools.com/cssref/css_colors.asp. It should just be a matter of copying and pasting them in.

Sesquipedalian (changes requested):

Sesquipedalian:
Just to be clear, the fact that I requested changes doesn't necessarily mean that I have an opinion about whether we should include this file in 2.1 or not. In theory, I like the idea of having a single class to handle all of our validation, but @jdarwood007 is right that switching over to using this for all our data validation would be a major change. We should probably have a discussion of it on the [Development Contributors](https://www.simplemachines.org/community/index.php?board=257.0) board before making any decisions.

albertlast:
I see not high chance for this topic to get merge.

jdarwood007:
While I don't see this making the 2.1 release.  Is there anything external to the project we can do in the mean time?
If we pushed some unit testing, would this help us work towards this?

FYI, for anything beyond 2.1, this is great.

PR #3981, opened 2017-03-21
Optimize the amount of queries for calendar

albertlast:
It could be happen that the same member got many event,
the query how get the timezone from the member is called once per event.

this code "cache" the result for every member.

### DIFF
--- a/Sources/Subs-Calendar.php
+++ b/Sources/Subs-Calendar.php
@@ -1639,9 +1639,16 @@ function buildEventDatetimes($row)
 function getUserTimezone($id_member = null)
 {
 	global $smcFunc, $context, $sourcedir, $user_info, $modSettings;
+	static $member_cache = array();
 
 	if (is_null($id_member) && $user_info['is_guest'] == false)
 		$id_member = $context['user']['id'];
+	
+	//check if the cache got the data
+	if(isset($id_member) && isset($member_cache[$id_member]))
+	{
+		return $member_cache[$id_member];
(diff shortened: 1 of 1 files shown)


Sesquipedalian (approved):

Sources/Subs-Calendar.php
```diff
@@ -1639,9 +1639,16 @@ function buildEventDatetimes($row)
 function getUserTimezone($id_member = null)
 {
 	global $smcFunc, $context, $sourcedir, $user_info, $modSettings;
+	static $member_cache = array();
 
 	if (is_null($id_member) && $user_info['is_guest'] == false)
 		$id_member = $context['user']['id'];
+	
+	//check if the cache got the data
+	if(isset($id_member) && isset($member_cache[$id_member]))
```
colinschoen: `if (`

colinschoen (changes requested):
Nit

colinschoen:
Any reason we aren't using `cache_put_data` and `cache_get_data`?

albertlast:
Sure,
wouldn't solve the issue instead of ~2k time we query the databse,
we then 2k times access the cache -> the runtime keep the same.

With the static call we redurce the amounts of calls.

Sesquipedalian:
Yeah, a static variable is the better solution here for sure.

colinschoen (approved):
Makes sense. This looks like a good improvement. LGTM, approval

Issue #3972, opened 2017-03-18
Problem reading arabic text after upgrading.


kipaslampu:
Im succesfull upgrade my smf forum to this smf 2.1
My problem is, its cannot read the old forum post and topic in arabic after upgrading.

The font show like  SaÃ¢â‚¬â„¢id
Thanks.

sbulen:
A few questions - 
Was your forum UTF8 before the upgrade?
Which do you use - mysql or postgresql?
What version of SMF were you running before the upgrade?

sbulen:
Possibly related...
On the 2.0 support forum, a similar issue was reported:
http://www.simplemachines.org/community/index.php?topic=552591.msg3918546#msg3918546

kipaslampu:
> A few questions -
Was your forum UTF8 before the upgrade?
Which do you use - mysql or postgresql?
What version of SMF were you running before the upgrade?

My forum is not utf8 before upgrade. After this problem im change it manually through my webmin to utf8 but the problem is still running.
Im using mysql
Before upgrade im using smf 2.0.7

sbulen:
Were you latin1-swedish-ci before the upgrade?

sbulen:
Or Cp1256?

kipaslampu:
> Were you latin1-swedish-ci before the upgrade?

Exactly. Before upgrade the langguage is using latin swedish.

sbulen:
Let me confirm the steps that took place:

You completed the 2.1 upgrade, but it did not properly convert your forum to utf8.

You noticed this, and your Web admin manually converted you after the 2.1 process.

My questions:

This was all in your TEST environment, correct?

We're there any errors during the 2.1 upgrade?

Did the utf8 step appear to execute?    Did it take a lot of time?

Can you explain the process used to convert to utf8 after the 2.1 upgrade?    Did your web admin simply change the tables in phpmyadmin?

sbulen:
kipaslampu - 

Do you have a search index built in 2.0?  If so, is it the 'full text' or the 'custom' index?

I'm still looking into this....

kipaslampu:
> This was all in your TEST environment, correct?
YES
We're there any errors during the 2.1 upgrade?
No errors
Did the utf8 step appear to execute? Did it take a lot of time?
UTF8 not show the progress bar... like bypass
Can you explain the process used to convert to utf8 after the 2.1 upgrade? Did your web admin simply change the tables in phpmyadmin?
[...]

Issue #3980, opened 2017-03-19
UTF8 forum in 2.0 gets 'mangled' when upgrading to 2.1


Bopske:
I have tested the upgrade to 2.1 once again. 

My 2.0.13 test forum was running full UTF8 without any issues. I have downloaded the 2.1 package from Github yesterday and performed an upgrade today. 

The result is not good. Somehow the UTF8 conversion seems to be started, even though the database is already fully UTF8. The result is that special characters are mangled bij the conversion. Not just on screen, but in the database too.

Before upgrade:
![before1](https://cloud.githubusercontent.com/assets/24902644/24082907/9d1a1afc-0ccd-11e7-8a07-51213dc81315.jpg)
![before2](https://cloud.githubusercontent.com/assets/24902644/24082908/9d1ba764-0ccd-11e7-8256-5d178a6e2a97.jpg)

After upgrade:
![after1](https://cloud.githubusercontent.com/assets/24902644/24082913/adc6bd10-0ccd-11e7-8c33-4df8746aeb04.jpg)
![after2](https://cloud.githubusercontent.com/assets/24902644/24082914/adc8621e-0ccd-11e7-9b18-813a956b5686.jpg)

Then when I check the language in the upgraded forum, it shows as ISO-8859-1, is that not supposed to be  UTF-8

![taal](https://cloud.githubusercontent.com/assets/24902644/24082903/787e9fec-0ccd-11e7-8932-c50734b28d69.jpg)

sbulen:
A few questions:
 - Was this a fresh 2.0 to 2.1 upgrade?   (Some people rerun the upgrader)
 - In 2.0.13, in your Settings.php file, what is the value for $language?
 - In 2.0.13, in your Settings.php file, what is the value for $db_character_set?
 - Were you able to run the 2.1 upgrade successfully at an earlier date?

sbulen:
I don't think the character set displayed is a real problem.  I'm pretty sure that is lifted from the $txt['lang_character_set'] setting in the language file.  

It is my understanding that $txt['lang_character_set'] setting was to be left at iso-8859-1 due to the behavior of the SMF language editor.  

That said, it is very confusing to have that displayed here, though, as it will display ISO-8859-1 even if the charset is UTF8.  Thoughts?   @Antes ?

sbulen:
[...]

PR #3984, opened 2017-03-23
Do not re-encode data already utf8

sbulen:
This change ensures that individual columns are encoded to UTF8 ONLY if they aren't already UTF8...

Re-encoding UTF8 content that is already UTF8 causes it to be garbled.   We see this whenever:
  -  A site admin manually converts a site to UTF8, without using the SMF conversion function
  -  A prior UTF8 attempt conversion failed, leaving mixed charsets

These two scenarios are problematic because the system may not detect that the site has been converted to UTF8 (e.g., global_character_set may not be set).   So, it does the UTF8 conversion again...

Fixes: Problem reading arabic text after upgrading. #3972

_* Consider this change in 2.0 in ManageMaintenance.php as well... *_

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -2818,7 +2818,7 @@ function ConvertUtf8()
 								$table_charsets[$charset] = array();
 
 							$table_charsets[$charset][] = $column_info;
-							}
+						}
 					}
 				}
 			}
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -2812,13 +2812,13 @@ function ConvertUtf8()
 					{
 						list($charset) = explode('_', $collation);
 
-					// Build structure of columns to operate on organized by charset; only operate on columns not yet utf8
-					if ($charset != 'utf8') {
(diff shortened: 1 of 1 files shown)


sbulen:
As background info, we have seen three recent examples of garbled text:
  - Problem reading arabic text after upgrading. #3972
  - UTF8 forum in 2.0 gets 'mangled' when upgrading to 2.1 #3980
  - http://www.simplemachines.org/community/index.php?topic=552591.msg3918546#msg3918546

Note that the 3rd item in the list above is specific to the 2.0 UTF8 conversion logic, which is identical to the upgrader UTF8 conversion logic.

sbulen:
This PR also addresses this issue: Upgrader - "Full" fixed-length fields can abort UTF8 conversion #4019

sbulen:
Let me know if there are any questions here.  

I really think this PR is a good safety measure that will help avoid data corruption for some users down the road.  I think this is a pretty safe change as well: there is no need to convert columns already converted.

PR #3989, opened 2017-03-25
Supplies some time zone abbreviations PHP doesn't know

Sesquipedalian:


Sesquipedalian:
Derp. Made a couple dumb mistakes. Just going to resubmit.

Issue #3992, opened 2017-03-28
Upgrader - collation set inconsistently, db may not be fully UTF8


sbulen:
In some instances, the 2.1 upgrader will build new tables in latin1_swedish_ci, not UTF8 as expected:
![new-tables-not-utf8](https://cloud.githubusercontent.com/assets/23568484/24388741/9aae62fe-1331-11e7-884e-8a1cd50fbbe9.JPG)

The problem is that the upgrader is inconsistent in how it sets charset & collation.  It uses two techniques.  The first method is substituting 'ENGINE=MYISAM;' with a string containing the collation:
![image](https://cloud.githubusercontent.com/assets/23568484/24388807/021b952e-1332-11e7-8193-eec17a5bbe22.png)

The second method is doing a substitution within parse_sql based on $db_collation:
![image](https://cloud.githubusercontent.com/assets/23568484/24388854/3dd6221e-1332-11e7-97c8-eb311b7cc464.png)

Some tables are setup to use method 1:
![image](https://cloud.githubusercontent.com/assets/23568484/24388937/c4dd4bc0-1332-11e7-838a-7899758f94a0.png)

And some are setup to use method 2:
![image](https://cloud.githubusercontent.com/assets/23568484/24388894/7c716754-1332-11e7-9f3c-b874546b0437.png)

The issue is with method 2...   It appears to me that _the upgrader actually never sets $db_collation_.  It appears blank at execution time.  But this is not normally visible, since the UTF8 conversion step in the 2.1 upgrader steps through the tables and fixes any tables not yet utf8_general_ci.   It appears the UTF8 conversion step covers up this issue...  

HOWEVER...  What happens when the DB has already been converted to UTF8, for example, using the 2.0 maintenance tool?

It appears the 2.1 upgrader skips the UTF8 conversion process, leaving the tables exactly as they were created.  They are created using the DB-level default charset & collation, normally latin1_swedish_ci.  

The simplest fix is to set $db_collation in the upgrader.   (A functioning micro-kludge.)  Or, cut over all new tables to using Method 1 by deleting the {$db_collation]..  

I believe the steps to verify are simple:  Convert your DB to UTF8 using the 2.0 maint menu.  Confirm your DB-level default collation is still latin1-swedish-ci (the MySQL default).  Run the 2.1 upgrader.  Look at the collation for the 5 tables in the first image above after migration.

Input welcome...  Am I missing something????   I'm surprised I never noticed this before...

[...]

PR #3993, opened 2017-03-28
Set collation type consistently

sbulen:
This addresses the issue where some tables were not built using the UTF8 charset.  

This sometimes happened when the DB had already been converted to UTF8 in 2.0.  The UTF8 conversion step gets bypassed, and new tables are built using the MySQL defaults, which may not be UTF8.

Fixes: Upgrader - collation set inconsistently, db may not be fully UTF8 #3992

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -2914,7 +2914,7 @@ function ConvertUtf8()
 		}
 	}
 	$_GET['substep'] = 0;
-	return false;
+	return true;
 }
 
 function serialize_to_json()
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -1329,7 +1329,7 @@ function DatabaseChanges()
 
 		return true;
 	}
-	return false;
+	return true;
 }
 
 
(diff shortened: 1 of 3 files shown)


sbulen:
Please hold up on approval...  

Based on discussion over on #3992 , let me tidy up some old code that isn't applicable anymore, and is potentially dangerous to leave around.  I'll need another day or two to change & retest.

sbulen:
OK, this has been tested & works.

Both methods - $db_collation & the MyISAM string substitution now work.

Removed the logic that set the target charset/collation based on the members table.  It needs to move to UTF8, no matter what is in the members table.  

The $db_collation method was specified consistently in the upgrade_2-0_mysql.php.
Now that it works, no changes were needed to upgrade_2-0_mysql.php to set the collation & charset to utf8 correctly (specifically, for tables that were dropped & re-added).

The MyISAM method is now used consistently throughout upgrade_2-1_mysql.php.

Tested with scenarios where UTF8 conversion was needed & where the conversion was not needed.

This fixes #3992 .

Issue #2182, opened 2014-08-13
Create an entire backup (excluding smileys, avatars and attachments) of the SMF install.


illori:
this feature is listed on the options page under package manager. is this a way to disable the auto backup that is created when you install a mod? if so the string of it is misleading. i would think that when i save this page it will create a backup, but it does not. also if this is the package manager backup it does not include every file, for example the Settings.php file is not backed up when you install a mod.

Oldiesmann:
Yes, that's a way to disable the backup. We can tweak the wording slightly to include "when a package is installed, uninstalled or upgraded" or something similar.

It only backs up files that a mod would touch - template/language/js/css files, the Sources directory, index.php, agreement.txt, the SSI files (SSI.php, ssi_examples.php and ssi_examples.shtml) and subscriptions.php (though we should add cron.php to that list for 2.1). This creates a smaller package size and uses less memory. There's really no need to back up Settings.php as no mod should ever be editing that file, and you've already got Settings_bak.php anyway.

illori:
you might not want a mod to edit Settings.php but we have seen a few [i dont know if they actually got approved], it cant hurt to backup a very important file like Settings.php.

yes changing the wording like that would work.

margarett:
Here's one :tongue: 
http://custom.simplemachines.org/mods/index.php?mod=3919

I don't recall any guidelines about not using Settings.php so I'd say we either explicitly deny it in the guidelines or this auto-backup should probably include Settings.php.

Oldiesmann:
That's a special case, but really nothing should be touching Settings.php directly. If new settings are needed, they should be added via saveSettings, which will automatically back up Settings.php before saving changes.

margarett:
So that really has to go into the guidelines... ;)

Can saveSettings be used in a package?

Arantor:
In theory, sure, but there are surprisingly few good reasons to edit Settings.php. Without seeing the package I couldn't tell you if this was done for a good reason or not. Usually not necessary though.

sbulen:
I submitted a PR to change the wording of the prompt per suggestions above.

I couldn't tell if y'all ended up aligned on backing up settings.php or not.   I did not do that.  I can add it to the backup if desired.   Let me know.  

I have a (slight) learning towards NOT backing it up, because that's just another place to worry about security-wise, given that it has critical logon information embedded within.

sbulen:
@illori - let me know if this is sufficient to close the issue.

[...]

PR #3995, opened 2017-03-30
Correct package backup prompt

sbulen:
Fixes: Create an entire backup (excluding smileys, avatars and attachments) of the SMF install. #2182

Changed the setting label to be more accurate.


colinschoen (approved):
LGTM, approval

PR #3998, opened 2017-04-04
Package install backup fixes

sbulen:
Addresses: Create an entire backup (excluding smileys, avatars and attachments) of the SMF install. #2182

3 changes:
  -  Made the backup prompt txt more accurate
  -  Added settings.php to backup
  -  Fixed glitch in backup filename dupe detection logic, in case package gets installed multiple times

### DIFF
--- a/Sources/Subs-Package.php
+++ b/Sources/Subs-Package.php
@@ -2980,7 +2980,7 @@ function package_create_backup($id = 'backup')
 
 	$files = array();
 
-	$base_files = array('index.php', 'SSI.php', 'agreement.txt', 'cron.php', 'ssi_examples.php', 'ssi_examples.shtml', 'subscriptions.php', 'settings.php');
+	$base_files = array('index.php', 'SSI.php', 'agreement.txt', 'cron.php', 'ssi_examples.php', 'ssi_examples.shtml', 'subscriptions.php');
 	foreach ($base_files as $file)
 	{
 		if (file_exists($boarddir . '/' . $file))
--- a/Sources/Subs-Package.php
+++ b/Sources/Subs-Package.php
@@ -2980,7 +2980,7 @@ function package_create_backup($id = 'backup')
 
 	$files = array();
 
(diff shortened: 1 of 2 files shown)


jdarwood007:
I believe Settings.php is purposefully left out because we have Settings_bak.php and you wouldn't want somebody to guess the backup filename (highly unlikely) and be able to get ahold of the database credentials.  The file can be recreated with some minimal work.

sbulen:
@jdarwood007 - this was explicitly requested over in the request: Create an entire backup (excluding smileys, avatars and attachments) of the SMF install. #2182 .    

@illori - thoughts?  Differing opinions here.  

I am OK with it either way.

illori:
we have had a number of people in the past that have had issues with their Settings.php file getting corrupted and their Settings_bak.php file was not usable as a replacement. otherwise there was no backup of the file for the user to use. that bug should be fixed but it does not hurt to have a backup of such a vital file.

emanuele45:
Jeremy D wrote on 05/04/17 02:35:
>
> I believe Settings.php is purposefully left out because we have Settings_bak.php and you wouldn't want somebody to guess the backup filename (highly unlikely) 

Not really **that** unlikely: provided most of the install have or will a patch installed at some point, guess the name is just a matter of guessing the patch installation date.
Of course there is the caveat that the webserver shall not honour .htaccess, but... I have the feeling (maybe wrong) this is more likely (IIS?) than have troubles with a backup of Settings.php.

But of course it's better to open the entire SMF install-base to a potential security hole just to avoid... how many? One request per month related to a broken Settings.php?... oh it's not 1 per month it's 5?... oh, it's not 5 per month, it's "a number" in the whole history of SMF? That's a huge amount indeed. :stuck_out_tongue:

sbulen:
Good points, @emanuele45 .  Thanks for the input.  It feels like a battle of edge conditions.  

I'll defer to @illori and team.

Oldiesmann:
We have Settings_bak.php if for some reason Settings.php gets broken. Including it in the package backups is a huge security risk and one I'd rather not take, though I don't know how many people are even aware of that feature.

sbulen:
Done.

PR #4003, opened 2017-04-06
Don't send like notifications for ignored users

colinschoen:
Again fixing #3965 

### DIFF
--- a/Sources/tasks/Likes-Notify.php
+++ b/Sources/tasks/Likes-Notify.php
@@ -31,7 +31,8 @@ class Likes_Notify_Background extends SMF_BackgroundTask
 		if ($this->_details['content_type'] == 'msg')
 		{
 			$request = $smcFunc['db_query']('', '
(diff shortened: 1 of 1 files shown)


Sources/tasks/Likes-Notify.php
```diff
@@ -74,6 +76,11 @@ public function execute()
 		if ($author == $this->_details['sender_id'])
 			return true;
 
+		// If the person who sent the notification is on this person's ignore list, do nothing.
+        if (!empty($ignored_members) && in_array($this->_details['sender_id'], $ignored_members)) {
+            return true;
+        }
+
```
jdarwood007: Minor but tabbing needs corrected

colinschoen: Thank you, been having a hell of a time with my IDE on this. I think it is the VIM plugin for it.

colinschoen:
Should be fixed now.

PR #4007, opened 2017-04-09
FixDoubleQuotes in PG fix #4005

### LINKED ISSUE #4005: Postgresql search issue with "
When you search for "in berlin" than the double quotes create a sql syntay error

`WHERE to_tsvector('pg_catalog.german',body) @@ to_tsquery('pg_catalog.german','"in berlin"')`

the solution could be to replace " with two single quotes ''

albertlast:
Fix #4005

### DIFF
--- a/Sources/DbSearch-postgresql.php
+++ b/Sources/DbSearch-postgresql.php
@@ -118,6 +118,9 @@ function smf_db_search_query($identifier, $db_string, $db_values = array(), $con
 			$db_values['db_error_skip'] = true;
 		}
 	}
+	
+	//fix double quotes
+	$db_values = str_replace('"',"'",$db_values);
 
 	$return = $smcFunc['db_query']('', $db_string,
 		$db_values, $connection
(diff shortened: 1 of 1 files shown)


Sources/DbSearch-postgresql.php
```diff
@@ -118,6 +118,9 @@ function smf_db_search_query($identifier, $db_string, $db_values = array(), $con
 			$db_values['db_error_skip'] = true;
 		}
 	}
+	
+	//fix double quotes
+	$db_values = str_replace('"',"'",$db_values);
```
jdarwood007: db_values is an array?

albertlast: sure but str_replace can handle array and return array.
And is recommanded to use instead of preg_replace() when search string is easy:
http://php.net/manual/en/function.str-replace.php

> If you don't need fancy replacing rules (like regular expressions), you should always use this function instead of preg_replace().

jdarwood007: Ahh, so it does.  Never noticed that.  Carry on.

colinschoen: Super nit: 

`$db_values = str_replace('"', "'", $db_values);`

colinschoen (changes requested):
nits

colinschoen (approved):
LGTM, approval

PR #4012, opened 2017-04-17
centralize the language setting for fulltext search

albertlast:
give the user/admin the option to change the fulltext search language by
changing the the search_language in the settings table

### DIFF
--- a/Sources/DbSearch-postgresql.php
+++ b/Sources/DbSearch-postgresql.php
@@ -29,6 +29,7 @@ function db_search_init()
 			'db_search_support' => 'smf_db_search_support',
 			'db_create_word_search' => 'smf_db_create_word_search',
 			'db_support_ignore' => false,
+			'db_search_language' => 'smf_db_search_language',
 		);
 
 	db_extend();
@@ -152,6 +153,41 @@ function smf_db_create_word_search($size)
 			'string_zero' => '0',
 		)
 	);
+}
+
+/**
+* Return the language for the textsearch index
+*/
+function smf_db_search_language()
+{
+	global $smcFunc, $modSettings;
+	
+	$language_ftx = 'english';
+	
+	if (!empty($modSettings['search_language']))
+		$language_ftx = $modSettings['search_language'];
+	else
+	{
+		$request = $smcFunc['db_query']('','
+			SHOW default_text_search_config',
+			array()
+		);
+
+		if ($request !== false && $smcFunc['db_num_rows']($request) == 1)
+		{
+			$row = $smcFunc['db_fetch_assoc']($request);
+			$language_ftx = $row['default_text_search_config'];
+
+			$smcFunc['db_insert']('replace',
+				'{db_prefix}settings',
+				array('variable' => 'string', 'value' => 'string'),
+				array('search_language', $language_ftx),
+				array('variable')
+			);
+		}
+	}
+
+	return $language_ftx;	
 }
 
 ?>
--- a/Sources/ManageSearch.php
+++ b/Sources/ManageSearch.php
@@ -231,19 +231,7 @@ function EditSearchMethod()
 				)
 			);
 
-			$request = $smcFunc['db_query']('', '
-				SHOW default_text_search_config',
-				array()
-			);
-
-			$language_ftx = 'simple';
-
-			if ($request !== false && $smcFunc['db_num_rows']($request) == 1)
-			{
-				$row = $smcFunc['db_fetch_assoc']($request);
-				$language_ftx = $row['default_text_search_config'];
-			}
-
+			$language_ftx = $smcFunc['db_search_language']();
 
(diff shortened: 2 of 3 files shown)


Sesquipedalian:
1. How does the admin set $modSettings['search_language']? I don't see any references to that setting anywhere else in SMF.
2. How do you plan to implement $modSettings['search_language'] for MySQL databases?

albertlast:
To the first point,
without this change the admin had to change this value directly in the postgres.conf see https://www.postgresql.org/docs/9.6/static/textsearch-configuration.html
with this pr the admin is able to change this value directly in the settings table with a db admin tool.
The pr is build that it create automaticaly $modSettings['search_language']

2. mysql fulltext search doesn't support different language
for exp. you had to maintaine the stop word list by your self: https://dev.mysql.com/doc/refman/5.7/en/fulltext-stopwords.html

Sesquipedalian:
I think you will need to build something into the UI to allow the admin to set this value. The admin should never need to resort to editing the database directly in order to configure a setting like this. Since this is only applicable to PostgreSQL, you will probably want to use a conditional check against `$db_type` to decide whether or not to show the setting in the admin UI.

albertlast:
because this is only setting for pg and this is a real specif one,
i see no reason to do so.
and any way is this unrelated to this pr.

Sesquipedalian:
Even if it is specific to Postgres, we don't expect admins to adjust SMF settings by directly editing the database. If you want admins to be able to control this setting, you need to build a UI for it.

albertlast:
Than he is not able to change this.
the main focus of this pr is the centralize of the needed code and
this do the pr very good.

Sesquipedalian:
Then you should remove the if/else in DbSearch-postgresql.php that refers to $modSettings['search_language']. You can add that stuff back in later in another pull request when you build the UI for it.

albertlast:
i do this for performance reason.

Sesquipedalian:
But really, why not just include the UI to control this setting? All you'd need to do is add a couple of lines in the EditSearchSettings() function in Sources/ManageSearch.php.

albertlast:
because my pr's got a single focus and
the idea where i place special pg setting bring ups many new question.
Because there exists many special pg stuff which i could place some where.

Sesquipedalian:
It doesn't bring up any questions in this case. This setting should be controlled in the search settings admin panel. Just add $db_type to the global declaration at the start of the function, and then add something like this just after $config_vars is first defined:

```
	if ($db_type == 'postgresql')
		$config_vars = array_merge($config_vars, array('', array('text', 'search_language')));
```

That's all there is to it. :)

Sesquipedalian:
Oh, and adding something like this to the Search.english.php language file:

```
$txt['search_language'] = 'whatever description of this setting you want to give';
```

Sesquipedalian:
Another option might be to replace `$modSettings['search_language']` in your existing PR with `$language`. Then you would always be using the default forum language as the Postgres search language. I'm not sure if that is what you want or not, @albertlast, but if it is, that might be the easiest way to do so.

albertlast:
you see the frontend stuff goes fast in many different direction what is not connected to the intention of this pr,
language and search_language are different.
search language needs to be a language which postgresql support and
in a form which pg understand: http://stackoverflow.com/questions/39751892/get-full-list-of-full-text-search-configuration-languages

Sesquipedalian:
Well then, just go with my first suggestion. Because if you want to add an a setting that the admin is supposed to configure, you need to include a way for the admin to configure it. The feature isn't complete until both the front and back ends are included.

albertlast:
than feel free to build a frontend for this,
when so easie.
the intention of this pr is to centralize the code and this do it fine.

Sesquipedalian:
But you didn't just centralize the code. You also added a setting that the admin apparently needs to configure in order to make your code work as expected. If your code works without that setting being configured, then why did you include it at all? If your code does not work unless that setting is configured, you need to include a way for the admin to configure it.

I understand that you want to keep your PR focused on one thing. But to do that, you either need to take out the part that depends on a $modSettings value, or you need to fully build the things you need to build in order to use a $modSettings value. I don't care which you do, but you need to do one or the other.

albertlast:
No the code works without action of the admin,
the value will be set by the code automatically.

Sesquipedalian:
So the admin is never supposed to configure it?

albertlast:
Sure otherwise i had to place something to do so.

Sesquipedalian:
In that case, it looks liit's the value of `$language_ftx` will only ever be the PG default, so why not just hard code the value into your code?

albertlast:
the "default" came from the pg conf,
that can be different on every pg setup.

Sesquipedalian:
Ah, now I see. Okay, that makes sense then. :)

Sesquipedalian (approved):

PR #4016, opened 2017-04-22
HTTPs fetching Simple Machines files

jdarwood007:
SM.org is now https, lets make use of it.  For fetching files we currently allow fallback to http.  In the future we should remove this.


Sources/ScheduledTasks.php
```diff
@@ -1246,12 +1246,16 @@ function scheduled_fetchSMfiles()
...
-		$server = empty($file['path']) || (substr($file['path'], 0, 7) != 'http://' && substr($file['path'], 0, 8) != 'https://') ? 'http://www.simplemachines.org' : '';
+		$server = empty($file['path']) || (substr($file['path'], 0, 7) != 'http://' && substr($file['path'], 0, 8) != 'https://') ? 'https://www.simplemachines.org' : '';
 		$url = $server . (!empty($file['path']) ? $file['path'] : $file['path']) . $file['filename'] . (!empty($file['parameters']) ? '?' . $file['parameters'] : '');
 
 		// Get the file
 		$file_data = fetch_web_data($url);
 
+		// If we got an error - maybe https isn't working, try http?
+		if ($file_data === false && substr($url, 0, 30) == 'https://www.simplemachines.org')
+			$file_data = fetch_web_data(str_replace('https://www.simplemachines.org', 'http://www.simplemachines.org', $url));
```
albertlast: In my eyes is the entire function a big backdoor.
missing validation of the connection (checks for ssl certs etc.)
missing validation of the files

albertlast: And the http fallback make the https stuff sensless.

jdarwood007: There isn't any built in method currently for our system to check the ssl certs.  The fallback is to currently allow the system to fall back incase port 443 or something else (like a weird php configuration I don't know about) prevents being able to fetch data over 443.  I thought about taking out http completely but wasn't sure.

Thinking about it if they don't have direct access to the machine, this would only occur with dns poisoning.  If they had direct access to the machine, they could change the DNS, but then also inject a new trusted root ca and self sign a certificate that would be trusted as well. There isn't stuff yet in the code to check for HSTS or similar either.

albertlast: and what is about the validation of the file which your try to download?
like i said the entire program part need to be rewriten.
the solution of download a html page and try you best to validate is fare away to be safe.

jdarwood007: If you got an idea of how to better validate the file, we can make things change for 2.1.  Previous versions of SMF still check against http.

I'm wanting to convert the latest-*.js files to be json based rather than js based.  But they currently do a little more than what you can do with json based results including handling some unique javascript coding to make things pop better in the admin panel for update warnings.

albertlast: Create a json based on specific schema, one field for the version number, another about the news, another field for notes with specif level (there is no need in my eyes for js)

jdarwood007: Not to argue the point, but that doesn't solve much here other than being able to inject javascript.  There isn't any validation of the received data still.

albertlast: in my eyes is the easy solution to disable html/js stuff which are coming by the json file:
http://stackoverflow.com/questions/20855482/preventing-html-and-script-injections-in-javascript

other/install.php
```diff
@@ -309,7 +297,7 @@ function load_lang_file()
 		<p>In some cases, FTP clients do not properly upload files with this many folders.  Please double check to make sure you <span style="font-weight: 600;">have uploaded all the files in the distribution</span>.</p>
 		<p>If that doesn\'t help, please make sure this install.php file is in the same place as the Themes folder.</p>
 
-		<p>If you continue to get this error message, feel free to <a href="http://support.simplemachines.org/">look to us for support</a>.</p>
+		<p>If you continue to get this error message, feel free to <a href="https://support.simplemachines.org/">look to us for support</a>.</p>
```
Sesquipedalian, line 300: Using `https` to fetch the files makes sense, but why change this link? We use `http` in similar links everywhere else in the codebase.

other/install.php
```diff
@@ -1906,7 +1894,7 @@ function template_install_below()
 		</div></div>
 		<div id="footer">
 			<ul>
-				<li class="copyright"><a href="http://www.simplemachines.org/" title="Simple Machines Forum" target="_blank" class="new_win">SMF &copy; 2017, Simple Machines</a></li>
+				<li class="copyright"><a href="https://www.simplemachines.org/" title="Simple Machines Forum" target="_blank" class="new_win">SMF &copy; 2017, Simple Machines</a></li>
```
Sesquipedalian, line 1897: This should probably remain as `http`

other/upgrade.php
```diff
@@ -723,7 +707,7 @@ function WelcomeLogin()
 				<li>Source Directory: ' . $boarddir . '</li>
 				<li>Cache Directory: ' . $cachedir_temp . '</li>
 			</ul>
-			If these seem incorrect please open Settings.php in a text editor before proceeding with this upgrade. If they are incorrect due to you moving your forum to a new location please download and execute the <a href="http://download.simplemachines.org/?tools">Repair Settings</a> tool from the Simple Machines website before continuing.';
+			If these seem incorrect please open Settings.php in a text editor before proceeding with this upgrade. If they are incorrect due to you moving your forum to a new location please download and execute the <a href="https://download.simplemachines.org/?tools">Repair Settings</a> tool from the Simple Machines website before continuing.';
```
Sesquipedalian, line 710: This should probably remain as `http`

other/upgrade.php
```diff
@@ -3451,7 +3435,7 @@ function template_upgrade_below()
 		</div>
 		<div id="footer">
 			<ul>
-				<li class="copyright"><a href="http://www.simplemachines.org/" title="Simple Machines Forum" target="_blank" class="new_win">SMF &copy; 2017, Simple Machines</a></li>
+				<li class="copyright"><a href="https://www.simplemachines.org/" title="Simple Machines Forum" target="_blank" class="new_win">SMF &copy; 2017, Simple Machines</a></li>
```
Sesquipedalian, line 3438: This should probably remain as `http`

other/upgrade.php
```diff
@@ -3772,7 +3756,7 @@ function template_upgrade_options()
 						<td width="100%">
 							<label for="stat">
 								Allow Simple Machines to Collect Basic Stats Monthly.<br>
-								<span class="smalltext">If enabled, this will allow Simple Machines to visit your site once a month to collect basic statistics. This will help us make decisions as to which configurations to optimise the software for. For more information please visit our <a href="http://www.simplemachines.org/about/stats.php" target="_blank">info page</a>.</span>
+								<span class="smalltext">If enabled, this will allow Simple Machines to visit your site once a month to collect basic statistics. This will help us make decisions as to which configurations to optimise the software for. For more information please visit our <a href="https://www.simplemachines.org/about/stats.php" target="_blank">info page</a>.</span>
```
Sesquipedalian, line 3759: This should probably remain as `http`

other/upgrade.php
```diff
@@ -4504,7 +4488,7 @@ function doTheDelete(theCheck)
 		echo '<br> Upgrade completed in ', $totalTime, '<br><br>';
 
 	echo '<br>
-			If you had any problems with this upgrade, or have any problems using SMF, please don\'t hesitate to <a href="http://www.simplemachines.org/community/index.php">look to us for assistance</a>.<br>
+			If you had any problems with this upgrade, or have any problems using SMF, please don\'t hesitate to <a href="https://www.simplemachines.org/community/index.php">look to us for assistance</a>.<br>
```
Sesquipedalian, line 4491: This should probably remain as `http`

Sesquipedalian (changes requested):
I think we should keep `http` in the clickable HTML links for consistency with the rest of the codebase.

other/install.php
XinYenFon: Why ?

Sesquipedalian: For consistency with the rest of the codebase. Alternatively, @jdarwood007 could go through everything and change all the links to use https. But we should use one or the other consistently.

jdarwood007:
Where else am I missing http? usage on our site stuff?  All hyperlinks should be fixed.

I didn't change the license header as I wasn't sure if I could.

Sesquipedalian:
Using grep to find all occurrences of `http://[^\.]+.simplemachines.org` (excluding those in the license blocks) produced the following list. I haven't checked every one exhaustively, so exactly how you will want to handle each case is up to you. :)

./LICENSE
./README.md
./Sources/Admin.php
./Sources/Class-CurlFetchWeb.php
./Sources/Help.php
./Sources/ManageLanguages.php
./Sources/ManageSmileys.php
./Sources/News.php
./Sources/ScheduledTasks.php
./Sources/Stats.php
./Sources/Themes.php
./Themes/default/css/index.css
./Themes/default/index.template.php
./Themes/default/languages/Admin.english.php
./Themes/default/languages/EmailTemplates.english.php
./Themes/default/languages/Help.english.php
./Themes/default/languages/Install.english.php
./Themes/default/languages/ManageBoards.english.php
./Themes/default/languages/Packages.english.php
./Themes/default/languages/Themes.english.php
./Themes/default/languages/index.english.php
./other/install.php
./other/install_2-1_mysql.sql
./other/install_2-1_postgresql.sql
./other/readme.html
./other/upgrade.php
./other/upgrade_1-1.sql
./other/upgrade_2-0_mysql.sql
./other/upgrade_2-0_postgresql.sql
./ssi_examples.php
./ssi_examples.shtml

jdarwood007:
Searched out and found some.  Leaving out the license and copyright changes for now.  Until I get word if changing those is appropriate.

My grep search: http://([^.]+)\.simplemachines\.org

Sesquipedalian:
For interest's sake and so you can hunt them all down yourself, the command I used for that was
`grep -ER 'http://[^\.]+.simplemachines.org' . | grep -Ev '@(author|license)' | grep -Eo '^[^:]+' | sort -u`

jdarwood007:
Ahh, I found I had a filter in place on my editor tool when I was doing a massive file search, which caused it to exclude a bunch of files.

Updated some more files.  I left out some of them on purpose or the 3rd party sites didn't use https yet so I ignored them.  This updates lots of language files as well.

Sesquipedalian:
Latest commit changes a lot of URLs to https that should not have been. You probably want to revert these changes and try again using a search and replace pattern like this:

search: `http://([^\.]+.simplemachines.org)`
replace: `https://$1`

jdarwood007:
I updated the 3rd party urls as well that supported https.  Things like php.net, apache.org, jquery.org, etc.  I followed some and updated them with their updated urls if they had a redirect.

Sesquipedalian:
https://www.php.net/manual/en/index.php, etc., returns a 404. Also in spots like licenses, URLs for namespace declarations (e.g. in News.php), etc., the http needs to remain. In the case of the DTDs they *must* remain http.

Sources/News.php
```diff
@@ -225,11 +225,11 @@ function ShowXmlFeed()
 	// Allow mods to add extra namespaces and tags to the feed/channel
 	$namespaces = array(
 		'rss' => array(),
-		'rss2' => array('atom' => 'http://www.w3.org/2005/Atom'),
-		'atom' => array('' => 'http://www.w3.org/2005/Atom'),
+		'rss2' => array('atom' => 'https://www.w3.org/2005/Atom'),
+		'atom' => array('' => 'https://www.w3.org/2005/Atom'),
```
Sesquipedalian: Namespace declaration URLs should not be changed.

Sources/ReCaptcha/RequestMethod/Curl.php
```diff
@@ -33,7 +33,7 @@ class Curl
 {
 
     /**
-     * @see http://php.net/curl_init
+     * @see https://www.php.net/curl_init
```
Sesquipedalian: https://www.php.net/curl_init returns a 404. Same problem elsewhere.
Perhaps try normalizing everything to https://secure.php.net instead of https://www.php.net

Sources/News.php
jdarwood007: Do you have any reference for that? I never knew that.  These namespace urls do support https.

Sesquipedalian: The feeds won't validate with anything but the canonical http URL. Try one in the [W3 feed validator](https://validator.w3.org/feed/). Using https in the namespace declaration URL results in the following error message:

```
Sorry

This feed does not validate.
line 2, column 0: feed is in an invalid namespace: https://www.w3.org/2005/Atom [help]
<feed xmlns="https://www.w3.org/2005/Atom">

In addition, interoperability with the widest range of feed readers could be improved by implementing the following recommendation.
line 2, column 0: Use of unknown namespace: https://www.w3.org/2005/Atom [help]
<feed xmlns="https://www.w3.org/2005/Atom">
```

jdarwood007: Good enough for me.  I already reversed it.  Just didn't know if any official documentation existed supporting it either way.

Sesquipedalian:
https://www.opensource.org times out, but https://opensource.org works.

other/upgrade.php
albertlast: https shoulde be the commone connection -> https is fine here in my eyes

other/upgrade.php
albertlast: https shoulde be the commone connection -> https is fine here in my eyes

other/upgrade.php
albertlast: https shoulde be the commone connection -> https is fine here in my eyes

other/upgrade.php
albertlast: https shoulde be the commone connection -> https is fine here in my eyes

other/install.php
albertlast: https shoulde be the commone connection -> https is fine here in my eyes

other/install.php
albertlast: https shoulde be the commone connection -> https is fine here in my eyes

jdarwood007:
@Sesquipedalian Thanks, I must have updated those bad.  Yea they are one of those weird sites where the subdomains is failing that really shouldn't (you should sign both example.com and www.example.com)

Saying example.com reminds me, I did update some of the domains I saw we had that used examples to use example.com exclusively, as its purpose is documented per RFC 2606 and RFC 6761.

Sources/News.php
Sesquipedalian: The official documentation is at https://www.w3.org/TR/xml-names/, but basically, if one changes the URI in the namespace declaration in any way, one is in fact declaring an entirely different namespace. For XML documents that require certain namespaces to be defined (both Atom and RSS2 do), changing the URI means that you are not defining the expected namespace and instead have defined some other, unknown namespace.

Also relevant from that documentation is that the URI does not need to be a working URL. It just needs to be a unique and permanent string that matches the criteria for a URI. What the URI is and whether it would point to any actual document if treated as a URL is beside the point.

With uncommon namespaces (e.g. our own `xmlns:smf="http://www.simplemachines.org"`), it is probable that most XML processors won't already know what it is and will therefore just accept whatever is defined in the XML document. But with common namespaces (e.g. Atom's `xmlns:atom="http://www.w3.org/2005/Atom"`)

Sources/News.php
```diff
@@ -233,8 +233,8 @@ function ShowXmlFeed()
 			'dc' => 'http://purl.org/dc/elements/1.1/',
 		),
 		'smf' => array(
-			'' => 'http://www.simplemachines.org/xml/' . $_GET['sa'],
-			'smf' => 'http://www.simplemachines.org/',
+			'' => 'https://www.simplemachines.org/xml/' . $_GET['sa'],
+			'smf' => 'https://www.simplemachines.org/',
```
Sesquipedalian: Since they are namespace declarations rather than linkable URLs, these should stay http. In reality, most XML processors aren't going to recognize SMF's proprietary namespace, so in real world applications this change wouldn't make much difference. But to be standards compliant, we should keep this namespace exactly as it has always been defined.

Sources/Themes.php
```diff
@@ -1452,7 +1452,7 @@ function InstallCopy()
 
 	// Lets add a theme_info.xml to this theme.
 	$xml_info = '<' . '?xml version="1.0"?' . '>
-<theme-info xmlns="http://www.simplemachines.org/xml/theme-info" xmlns:smf="http://www.simplemachines.org/">
+<theme-info xmlns="https://www.simplemachines.org/xml/theme-info" xmlns:smf="https://www.simplemachines.org/">
```
Sesquipedalian: Namespaces again. These should stay http.

Sources/Who.php
```diff
@@ -772,10 +772,10 @@ function Credits($in_admin = false)
...
 		),
 		'software' => array(
-			'<a href="http://jquery.org/">JQuery</a> | &copy; John Resig | Licensed under <a href="http://github.com/jquery/jquery/blob/master/MIT-LICENSE.txt">The MIT License (MIT)</a>',
-			'<a href="http://cherne.net/brian/resources/jquery.hoverIntent.html">hoverIntent</a> | &copy; Brian Cherne | Licensed under <a href="http://en.wikipedia.org/wiki/MIT_License">The MIT License (MIT)</a>',
-			'<a href="http://www.sceditor.com/">SCEditor</a> | &copy; Sam Clarke | Licensed under <a href="http://en.wikipedia.org/wiki/MIT_License">The MIT License (MIT)</a>',
-			'<a href="http://wayfarerweb.com/jquery/plugins/animadrag/">animaDrag</a> | &copy; Abel Mohler | Licensed under <a href="http://en.wikipedia.org/wiki/MIT_License">The MIT License (MIT)</a>',
+			'<a href="https://jquery.org/">JQuery</a> | &copy; John Resig | Licensed under <a href="https://github.com/jquery/jquery/blob/master/MIT-LICENSE.txt">The MIT License (MIT)</a>',
+			'<a href="https://github.com/briancherne/jquery-hoverIntent">hoverIntent</a> | &copy; Brian Cherne | Licensed under <a href="https://en.wikipedia.org/wiki/MIT_License">The MIT License (MIT)</a>',
```
Sesquipedalian: Looks like the official URL for this one is https://briancherne.github.io/jquery-hoverIntent/

Sources/Who.php
```diff
@@ -785,9 +785,9 @@ function Credits($in_admin = false)
 			'<a href="https://github.com/matthiasmullie/minify">Minify</a> | &copy; Matthias Mullie | Licensed under <a href="http://en.wikipedia.org/wiki/MIT_License">The MIT License (MIT)</a>',
 		),
 		'fonts' => array(
-			'<a href="http://openfontlibrary.org/en/font/anonymous-pro"> Anonymous Pro</a> | &copy; 2009 | This font is licensed under the SIL Open Font License, Version 1.1',
-			'<a href="http://openfontlibrary.org/en/font/consolamono"> ConsolaMono</a> | &copy; 2012 | This font is licensed under the SIL Open Font License, Version 1.1',
-			'<a href="http://openfontlibrary.org/en/font/phennig"> Phennig</a> | &copy; 2009-2012 | This font is licensed under the SIL Open Font License, Version 1.1',
+			'<a href="https://fontlibrary.org/en/font/anonymous-pro"> Anonymous Pro</a> | &copy; 2009 | This font is licensed under the SIL Open Font License, Version 1.1',
+			'<a href="https://fontlibrary.org/en/font/consolamono"> ConsolaMono</a> | &copy; 2012 | This font is licensed under the SIL Open Font License, Version 1.1',
+			'<a href="https://openfontlibrary.org/en/font/phennig"> Phennig</a> | &copy; 2009-2012 | This font is licensed under the SIL Open Font License, Version 1.1',
```
Sesquipedalian: Missed one. :)
Should be https://fontlibrary.org/en/font/phennig

Sources/minify/src/JS.php
```diff
@@ -15,7 +15,7 @@
 class JS extends Minify
 {
     /**
-     * Var-matching regex based on http://stackoverflow.com/a/9337047/802993.
+     * Var-matching regex based on https://stackoverflow.com/a/9337047/802993.
```
Sesquipedalian, line 18: This redirects back to the http version

jdarwood007: It does, but only through the shorturl redirect that stackoverflow does. I suspect that is something on their end that will change/be fixed someday.  It still produces a valid cert, works and gets you to the same content.

jdarwood007: https://meta.stackexchange.com/questions/223728/permalink-short-urls-lose-https
There is a bug report on it.

Sesquipedalian: Fair enough

Sesquipedalian (approved):
LGTM

Sources/minify/src/JS.php
jdarwood007: Just to add some meta to this since it came across my feeds.  https://nickcraver.com/blog/2017/05/22/https-on-stack-overflow/
They have enabled it completely.

Issue #3996, opened 2017-04-01
UTF8 conversion - template not displayed


sbulen:
The UTF8 conversion always executes 'in the background', meaning, the template does not kick off & tell you what it's doing.   You want the template displayed when the UTF8 update executes.  

This is EASILY fixed by making the UTF8 conversion end-of-step return 'false' like all of the other multi-substep steps.  E.g., compare the utf8 conversion end of step:
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/other/upgrade.php#L2915

With the backup end of step:
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/other/upgrade.php#L1210

Or the JSON end of step:
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/other/upgrade.php#L3190

Or the DB-changes end of step:
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/other/upgrade.php#L1320

Returning false properly displays the template...  

_**HOWEVER...**_   

Then the continue button doesn't work and you can't move past the UTF8 step.  I have not found a way to enable the template - that doesn't leave you stuck in the UTF8 step.  

There are a few little step & substep management code snippets used in the JSON and backup steps that are not used in the UTF8 step.  Even adding these, I still haven't been able to get it work properly.  Specifically, the continue button remains 'disabled' at the end of the UTF8 step.  It's driving me nuts...   I'm probably missing something simple somewhere...

sbulen:
THIS one...  Is a PITA...  Every time I poke it with a stick, it pokes back...

sbulen:
For some reason the UTF8 template is only being displayed once, at the very end, where other templates are displayed only once, at the very start.  

The little inline javascript in the template is supposed to "watch" and only enable the button when it sees it work on the last item.  But the javascript never "sees" the last table complete for the utf8 step, because the template is painted too late in the process at "step completed"...   

My theory is that the root cause is a problem with step & substep tracking.  The upgrader actually tracks steps & substeps in multiple places.  (That's why the URL is wrong as the upgrader executes...)   

For the UTF8 conversion step, for some reason the step & substep aren't reset.  I believe this is why the template doesn't kick off as expected.  

Fixing the step/substep issue may make the 'restart' links work as well.  They don't work at the moment, because the step & substep used by the restart (in the red restart dialog box) are the wrong ones; they take you back much earlier in the process.

sbulen:
[...]

PR #4020, opened 2017-04-24
Fix utf8 template

sbulen:
Several changes to get the template to display during the UTF8 conversion step.  While in there, I brought  some consistency across the templates, so the steps behave similarly (scrollbars, buttons, display etc).   

Addresses: UTF8 conversion - template not displayed #3996

While I was in there I also:
  -  Confirmed the subtemplate exists - part of the issue is that it had been calling a non-existent subtemplate...
  -  The connect was generating internal "out of sequence" warnings on subsequent calls; used a ping instead, as that will attempt a reconnect; I think this will be helpful during long-running upgrades
  -  SourceURL directives allow debugging of inline .js using your browser's debugger

Tested on large & small forums, with & without backups, with & without debug info displayed, and with & without a UTF8 conversion needed.  Testing combos was needed, because sometimes steps affected subsequent steps' behavior (another part of the reason the template didn't display)...  This PR resolves this. The only quirk outstanding is that the very first tablename is displayed (not executed, displayed) 2x if debug is enabled during UTF8 conversion...  I gave up on that one after spending WAY too much time on it...

### DIFF
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -58,6 +58,7 @@ function smf_db_initiate($db_server, $db_name, $db_user, $db_passwd, $db_prefix,
 			'db_escape_wildcard_string' => 'smf_db_escape_wildcard_string',
 			'db_is_resource'            => 'smf_is_resource',
 			'db_mb4'                    => false,
+			'db_ping'                   => 'mysqli_ping',
 		);
 
 	if (!empty($db_options['persist']))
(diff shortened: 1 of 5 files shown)


sbulen:
I did not test pg (don't have access in my env), & I did not test using the cmdline (mainly because I don't know how...).

sbulen:
Fixed the quirk identified above (first table displayed 2x - it was driving me nuts...).  New commit forthcoming after a couple more tests.  While I was at it displayed the table names consistently with other steps as well (stripped out the $db_prefix).

PR #4021, opened 2017-04-24
Uses $context['posting_fields'] for all fields in post form header

Sesquipedalian:
Benefits:

1. The logic for setting up these fields is handled in source rather than in the theme
2. It is easier to ensure uniform formatting for all fields
3. Mods can add custom fields above, below, or between standard fields using nothing but the 'integrate_post_end' hook
4. Every dt/dd pair in the list is automatically assigned a unique `pf_*` CSS class

### DIFF
--- a/Sources/Post.php
+++ b/Sources/Post.php
@@ -1326,20 +1326,19 @@ function Post($post_errors = array())
 	);
 
 	// Icons are fun.
-	$message_icon_select = '<select name="icon" id="icon" onchange="showimage()">';
+	$context['posting_fields']['icon'] = array(
+		'dt' => $txt['message_icon'],
+		'dd' => '<select name="icon" id="icon" onchange="showimage()">',
+	);
 	foreach ($context['icons'] as $icon)
 	{
-		$message_icon_select .= '
+		$context['posting_fields']['icon']['dd'] .= '
 							<option value="' . $icon['value'] . '"' . ($icon['value'] == $context['icon'] ? ' selected' : '') . '>' . $icon['name'] . '</option>';
 	}
-	$message_icon_select .= '
+	$context['posting_fields']['icon']['dd'] .= '
 						</select>
 						<img id="icons" src="' . $context['icon_url'] . '">';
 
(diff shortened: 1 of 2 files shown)


Sesquipedalian:
How does this look to you, @Antes?

XinYenFon:
Well, putting those DL/DD/DT in source not a huge deal but putting img/icons to source will limit the theme authors especially the people who is using font icons in their themes.

Sesquipedalian:
> but putting img/icons to source will limit the theme authors especially the people who is using font icons in their themes.

How so? They can edit the icons using the `integrate_load_message_icons` hook, just as before. Or if that isn't enough, they can completely rebuild `$context['posting_fields']['icon']` to be whatever they want using the `integrate_post_end` hook.

XinYenFon:
Do they require extra modification to integrate the hooks? If your answer is yes then my answer is no.

Sesquipedalian:
No, no extra mods besides the theme itself would be necessary. For example, if a theme author didn't want to use the standard message icon select menu contained in `$context['posting_fields']['icons']['dd']`, she could either (a) include a hooked function in the theme's files to replace it with whatever she wanted, or (b) just process the content of `$context['posting_fields']['icons']` however she liked directly in her Post.template.php. Everything is just in $context, after all. Since themes can define and use hooks just like any other mod, all the necessary code can be included in the theme package regardless of which of these methods a theme author might prefer.

Sesquipedalian:
Clearer explanation:

If a theme author wants to do things the old way, they are completely able to do that. This PR doesn't take anything away. It just adds more info to $context. If a theme wants to ignore it, change it, or use it, is entirely up to the theme author. If the theme author wants to change it, she can either include a hooked function in her theme's files to do so, or she can do it directly in her theme's custom version of Post.template.php. In short, this PR is just giving more options and flexibility; it doesn't take anything away.

XinYenFon:
Fine enough for me, but surely requires some documentation.

Sesquipedalian:
Documentation added.

PR #4022, opened 2017-04-24
Removes « » around "last edit" string after using quick edit

Sesquipedalian:


XinYenFon:
LGTM.

illori:
but why do we need to do this? i think the symbols helps to identify that this is the edit by text.

XinYenFon:
Those type of things needs to be done by CSS and I think SMF already does that (if not still needs to be done by CSS).

Sesquipedalian:
> but why do we need to do this? i think the symbols helps to identify that this is the edit by text.

This change makes the HTML inserted by the JavaScript consistent with what the theme template generates on page load. See https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Themes/default/Display.template.php#L677

Sesquipedalian:
Here's how it looks in the default theme when an edited post is first loaded.
screen shot 2017-04-27 at 10 47 53 am


If someone likes the « » characters enough to put them back into the theme (I don't), that should be done, as @Antes says, by editing the CSS rather than by hard coding those characters into the HTML.

XinYenFon:
That header part of the post requires some work in general.

Sesquipedalian:
> That header part of the post requires some work in general.

Fair enough, but what you said about using CSS rather than HTML to style the "Last Edit" message is still true.

illori:
also looks like if you have an edit reason in place the double arrows dont show up at all with the edit text.

illori:
also the italics dont show up with the edit text, well maybe it does? might be related to #4028 it seems to come and go depending on the refresh.

Sesquipedalian:
That all sounds like the expected behaviour, @illori. At least as it exists right now, Curve for SMF 2.1 doesn't wrap the "Last Edit" message in « » characters or render it in italics. The only exception has been the bit of JavaScript that this PR fixes in order to bring it into line with the rest.

illori:
well it does it, when the subject does not shows the text does have the double arrows and it is in italics as it is in SMF 2.0.


 then i refreshed the page and it changes



something is not consistent here.

XinYenFon:
Its not a huge deal, we can handle the inconsistency with enabling CSS.

Sesquipedalian:
That inconsistency is what this PR corrects, @illori

illori:
It was not clear to me. Then I guess this should get committed

colinschoen (approved):
LGTM, approval

PR #4024, opened 2017-04-25
Fixes a bug where incorrect times were loaded when editing an event

Sesquipedalian:


Sesquipedalian:
On second thought, I'll just roll this into #3990.

PR #4026, opened 2017-04-27
Allows mods to easily add custom settings to template_modify_board & template_modify_category

Sesquipedalian:


Sesquipedalian:
Thoughts, @Antes?

XinYenFon:
I have nothing to say. LGTM.

Sesquipedalian:
Cool. I just like to get your input about theme template stuff.

colinschoen (approved):
LGTM, approval

DiegoAndresCortes:
@colinschoen Would you mind explaining how $context['custom_board_settings'] is being used?
Do you have to define this array on the mod function for this hook?
What happens if multiple mods want to add board settings? The array will be overwritten, or am I missing something?

colinschoen:
Yes, they will create or mutate the `$context['custom_board_settings']` array depending on if it has been previously defined. Mods that want to add `custom_board_settings` will check if the array has been defined and preserve the original key-value pairs and add their own if it is already defined.

Sesquipedalian:
@DiegoCortes, you just need to append some data to `$context['custom_board_settings']`. For example, to add a simple checkbox, you'd do something like this in your hooked function:
```
function mod_name_edit_board()
{
	global $context, $txt;

	loadLanguage('mod_name');

	if (isset($context['board']['is_new']) && $context['board']['is_new'] === true)
		$context['board']['mod_name_enable'] = 0;

	$context['custom_board_settings']['mod_name_enable'] = array(
		'dt' => '<strong>' . $txt['mod_name_enable'] . ':</strong>',
		'dd' => '<input type="checkbox" name="mod_name_enable"' . (!empty($context['board']['mod_name_enable']) ? ' checked="checked"' : '') . '>',
	);
}
```

Note that if you hook into `integrate_edit_board`, you will probably also want to hook into some or all of the following as well:

- `integrate_pre_modify_board`
- `integrate_modify_board`
- `integrate_pre_boardtree`
- `integrate_boardtree_board`

DiegoAndresCortes:
Yeah I figured it out after 30 minutes, it was the only thingy missing as it is barely new.
Thanks @sesquipedalian

PR #4030, opened 2017-04-30
More rss improvements

Sesquipedalian:
- Adds language declaration to RSS and Atom feeds
- Uses value of $settings['og_image'] for feed icon image, if available
- Appends board parameter to generated RSS links when appropriate
- Sets default feed format to RSS2 instead of SMF's proprietary format
- Tidies some code and corrects some documentation
- Use the [tag URI scheme](https://en.wikipedia.org/wiki/Tag_URI_scheme), which is designed and recommended for exactly this purpose, to generate GUIDs for feed items

### DIFF
--- a/Sources/News.php
+++ b/Sources/News.php
@@ -651,6 +651,9 @@ function getXmlMembers($xml_format)
 	$data = array();
 	while ($row = $smcFunc['db_fetch_assoc']($request))
 	{
+		// Create a GUID for each member using the tag URI scheme
+		$guid = 'tag:' . parse_url($scripturl, PHP_URL_HOST) . ',' . gmdate('Y-m-d', $row['poster_time']) . ':member=' . $row['id_member'];
+
 		// Make the data look rss-ish.
 		if ($xml_format == 'rss' || $xml_format == 'rss2')
 			$data[] = array(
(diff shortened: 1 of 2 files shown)


colinschoen (changes requested):
nit: The GUID creation is repeated several times. It might be nice to encapsulate this in a function. 

All else LGTM

Sesquipedalian:
I'm not sure what you mean, @colinschoen. Each of the four `getXml*` functions has its own version because each one has different needs. Three of those functions use while loops to go through the rows returned from the DB query, and each row needs its own GUID. But all of that is as it should be. Am I missing something?

Sesquipedalian:
Looking again at this in the light of day and after a good cup of coffee, I just don't think a separate function is worthwhile for creating the GUIDs, @colinschoen. The only part that is actually the same in every case is just the `'tag:' . parse_url($scripturl, PHP_URL_HOST) . ','` bit. Everything after that is unique to each case. A function would need three parameters and a global variable declaration, but all it would perform would be a one line string concatenation.

colinschoen:
Sounds good. LGTM, approval

colinschoen (approved):

PR #4033, opened 2017-05-01
some improvments arround getAttachmentFilename

albertlast:
Mostly clean up the code.

Fix one possble logic error in subs line 3707
we check there if "Are we using multiple directories?" but don't if the dir is set.
I'm not 100% sure if a null array element can exists,
so i place a fallback when null is provided.

motivation to look into was this topic here: https://www.simplemachines.org/community/index.php?topic=553601.0

### DIFF
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -3161,7 +3161,7 @@ function profileSaveAvatarData(&$value)
 		$url = parse_url($_POST['userpicpersonal']);
 		$contents = fetch_web_data($url['scheme'] . '://' . $url['host'] . (empty($url['port']) ? '' : ':' . $url['port']) . str_replace(' ', '%20', trim($url['path'])));
 
-		$new_filename = $uploadDir . '/' . getAttachmentFilename('avatar_tmp_' . $memID, false, null, true);
+		$new_filename = $uploadDir . '/' . getAttachmentFilename('avatar_tmp_' . $memID, null, false, true);
 		if ($contents != false && $tmpAvatar = fopen($new_filename, 'wb'))
 		{
 			fwrite($tmpAvatar, $contents);
@@ -3265,7 +3265,7 @@ function profileSaveAvatarData(&$value)
 				if (!is_writable($uploadDir))
 					fatal_lang_error('attachments_no_write', 'critical');
 
-				$new_filename = $uploadDir . '/' . getAttachmentFilename('avatar_tmp_' . $memID, false, null, true);
+				$new_filename = $uploadDir . '/' . getAttachmentFilename('avatar_tmp_' . $memID, null, false, true);
 				if (!move_uploaded_file($_FILES['attachment']['tmp_name'], $new_filename))
 					fatal_lang_error('attach_timeout', 'critical');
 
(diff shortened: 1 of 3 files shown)


colinschoen:
Can you explain these changes?

albertlast:
the most changes are false null change,
this is because false is only allow on the 4th position and null only allow on thirth.

and the is_array check do two things in one,
first check if we run on multiple directories and
second if the expected behavior is there (we exptect in the next code line that is an array).

Sources/Subs-Attachments.php
```diff
@@ -1195,7 +1195,7 @@ function loadAttachmentContext($id_msg, $attachments)
 							$thumb_mime = 'image/' . $thumb_ext;
 
 						$thumb_filename = $attachment['filename'] . '_thumb';
-						$thumb_hash = getAttachmentFilename($thumb_filename, false, null, true);
+						$thumb_hash = getAttachmentFilename($thumb_filename, 0, null, true);
```
Sesquipedalian: This change doesn't make any difference, since the `false` will be cast to an integer anyway.

Sources/Profile-Modify.php
```diff
@@ -3161,7 +3161,7 @@ function profileSaveAvatarData(&$value)
 		$url = parse_url($_POST['userpicpersonal']);
 		$contents = fetch_web_data($url['scheme'] . '://' . $url['host'] . (empty($url['port']) ? '' : ':' . $url['port']) . str_replace(' ', '%20', trim($url['path'])));
 
-		$new_filename = $uploadDir . '/' . getAttachmentFilename('avatar_tmp_' . $memID, false, null, true);
+		$new_filename = $uploadDir . '/' . getAttachmentFilename('avatar_tmp_' . $memID, null, false, true);
```
Sesquipedalian: With this change you would be setting the `$attachment_id` parameter to `null` and the `$dir` parameter to `false`. Changing `$attachment_id` doesn't make any difference, since it will just be cast to an integer. But the changing `$dir` from `null` to `false` will cause problems. `$dir` is used as a key to the `$modSettings['attachmentUploadDir']` array. A `null` key will return nothing, but a `false` key will be treated like a `0` key, i.e. the first key of the array. That's not what we want in `getAttachmentFilename()`.

The same applies to the other similar changes in this PR.

Sources/Subs.php
```diff
@@ -3704,7 +3704,7 @@ function getAttachmentFilename($filename, $attachment_id, $dir = null, $new = fa
 		$file_hash = sha1(md5($filename . time()) . mt_rand());
 
 	// Are we using multiple directories?
-	if (!empty($modSettings['currentAttachmentUploadDir']))
+	if (is_array($modSettings['attachmentUploadDir']))
```
Sesquipedalian, line 3707: In theory, `!empty($modSettings['currentAttachmentUploadDir'])` will always be the same as `is_array($modSettings['attachmentUploadDir'])`. But I suppose a mod or something could mess with the value of `$modSettings['currentAttachmentUploadDir']`, so this change makes sense.

Sesquipedalian (changes requested):
Changing `!empty($modSettings['currentAttachmentUploadDir'])` to `is_array($modSettings['attachmentUploadDir'])` makes sense, but the other changes would either do nothing or create bugs.

colinschoen (approved):
Ok this looks fine, now. I'm not sure what the now reverted changes were for.

LGTM, approval

Sesquipedalian:
LGTM

PR #4036, opened 2017-05-03
Adds support for partial & resumable downloads to showAttachments()

Sesquipedalian:
This is pretty straightforward. If the client used the HTTP Range header to request a partial download of an attachment, this PR serves up the requested part of the file. Based closely on some tried and true code found in multiple places on the 'Net.


colinschoen (approved):
Very nice. LGTM, approval

PR #4038, opened 2017-05-07
Prevent direct calls to SSI.php functions,

### LINKED ISSUE #4034: Preventing "?ssi_function=`any ssi function name`" from overriding my page
Lets assume I have this very simple `derp.php` page where I want to use SMF's user system:

    <?php
    require('./SSI.php');

    if($user_info['is_guest'])
        echo 'you are a guest, get out!';
    ?>

Accessing it as `derp.php?ssi_function=welcome` will prevent normal control flow in `derp.php`.
How can I prevent/disable this feature?

PS: This is my current workaround, but it prevents me from using GET on the page:

    <?php
    if(count($_GET))
    	return header('Location: .');
    
    require('./SSI.php');

jdarwood007:
unless we have opted not to for some reason.  Fixes #4034

### DIFF
--- a/SSI.php
+++ b/SSI.php
@@ -177,6 +177,9 @@ if (!isset($_SESSION['USER_AGENT']) && (!isset($_GET['ssi_function']) || $_GET['
 // Have the ability to easily add functions to SSI.
 call_integration_hook('integrate_SSI');
 
+// Ignore a call to ssi_* functions if we are not using SSI.php
+if (empty($modSettings['allow_ssi_functions_anywhere']) && isset($_GET['ssi_function']) && basename($_SERVER['PHP_SELF']) !== 'SSI.php')
+	unset($_GET['ssi_function']);
 // Call a function passed by GET.
 if (isset($_GET['ssi_function']) && function_exists('ssi_' . $_GET['ssi_function']) && (!empty($modSettings['allow_guestAccess']) || !$user_info['is_guest']))
 {


jdarwood007:
The travis job is failing for hhvm.  But I believe its related to this: https://github.com/travis-ci/travis-ci/issues/7712

Sesquipedalian:
I am not sure that this PR as it stands now will solve the issue adequately. If the allow_ssi_functions_anywhere setting is enabled, the problematic behaviour will still occur. We need to find a way for that setting to be enabled while still preventing $_GET parameters from allowing unwanted code execution.

jdarwood007:
You can still use SSI.php?ssi_function=welcome and such.  You just can't use it on any other named scripts.  I added the hidden mod setting incase there is a not foreseeable reason that you need to have the ability to use ssi_function outside of SSI.php.

Sesquipedalian:
Weird. I could have sworn `$modSettings['allow_ssi_functions_anywhere']` already existed. I must have been having déjà vu. Never mind. :)

Sesquipedalian (approved):
LGTM

colinschoen:
@jdarwood007 

> Yes adding dist: trusty to the config got HHVM working

colinschoen (approved):
LGTM, approval

jdarwood007:
Yes, but my testing of specifying trusty also breaks PHP 5.3 testing.  Since 2.1 still supports 5.3, I don't know if I should remove the call or not.  On a technical aspect, I don't think we would come across anything that was removed in 5.3 at this stage in 2.1's development cycle.

bit2shift:
An improved workaround from #4034 for currently deployed forums where patching `SSI.php` isn't possible:

     <?php
    +if(isset($_GET['ssi_function']))
    +{
    +	unset($_GET['ssi_function']);
    +	return header('Location: ?' . http_build_query($_GET));
    +}

jdarwood007:
The patch for SSI.php is fairly straight forward and safe to make even on 2.0.  It is handled directly in SSI.php so no modifications are needed in any other scripts.

bit2shift:
I know, but I honestly disagree with having a `$modSettings['allow_ssi_functions_anywhere']` which, as far as I know, doesn't exist outside of this pull request.
The SSI functions [SHOULD NOT](https://www.ietf.org/rfc/rfc2119.txt) be callable when accessing scripts that require SSI.php.

jdarwood007:
You can leave it out.

If the modSetting is not in the settings table, it won't take effect for this and thus disable using SSI.php functions outside of SSI.php.  However as 2.1 is supposed to be getting stable, I didn't want to submit a PR that could change cause a backwards functionality breakage.  I would think in future versions we could remove that setting, giving time for anyone using this (likely noone) time to fix it.

PR #4039, opened 2017-05-09
Fixes #4028

### LINKED ISSUE #4028: edit bug
enable in topic settings show subjects in topics
then make a new post
wait long enough to edit the post that the edit text should show up. 
before you edit the post you see the subject of the post above the date
after you use quick edit to edit the post you see the date and to the right the of the date without the subject showing
refresh the page and the subject now returns to its proper place above the date.

i tried to add a reason for the edit and i dont see that showing up anywhere.

if you edit on the full edit screen and get returned to the topic the subject shows up without issue.

Sesquipedalian:
Instead of mucking about with JavaScript setting or unsetting the subject text, this just uses CSS to hide or show it.

### DIFF
--- a/Themes/default/Display.template.php
+++ b/Themes/default/Display.template.php
@@ -655,7 +655,7 @@ function template_single_post($message)
 
 	//Some people don't want subject ... The div is still required or quick edit breaks...
 	echo '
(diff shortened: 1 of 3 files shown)


colinschoen (approved):
LGTM, approval

PR #4042, opened 2017-05-10
Better control flow regarding $_GET['ssi_function']

### LINKED ISSUE #4034: Preventing "?ssi_function=`any ssi function name`" from overriding my page

bit2shift:
This is an ~improvement~ **alternative** to #4038 (fixes #4034).
According to SMF's wiki page "[Add a Function to your Website](https://wiki.simplemachines.org/smf/Add_a_Function_to_your_Website)", there are two distinct ways of calling the SSI functions.
This implies that you can use either A or B, but not both (`GET /mypage.php?ssi_function=welcome`).

Signed-off-by: Danilo Margarido <bit2shift@gmail.com>


Sesquipedalian (changes requested):
I agree with @bit2shift that it would be better not to use an empty setting to create an override. I don't think we need to worry about breaking old mods that relied on erroneous behaviour. If any exist that do that (doubtful), they should not have been doing so in the first place and it will be easy for them to fix whatever they were doing. Since mod authors know that there are lots of changes in 2.1, they already expect that they will need to update their mods to make them compatible, and so maintaining backwards compatibility for this sort of thing is not necessary. I mean, maintaining backwards compatibility if possible to support mods that were functioning according to best practices is one thing; maintaining backwards compatibility for a previously unknown bug is another.

However, I think the best solution is simply to remove the `empty($modSettings['allow_ssi_functions_anywhere']) && ` bit from @jdarwood007's code. Your alternative works, @bit2shift, but simply unsetting the unwanted $_GET key at the point #4038 does is simpler and cleaner.

So, @bit2shift, could you revert these changes and instead just take out the unnecessary bit from the previous code?

bit2shift:
If you/ really+/ want to invoke `isset($_GET['ssi_function'])` 3 times in a row instead of 1, alright!
And no, the existing code after @jdarwood007's patch is everything but simpler and cleaner.
Unsetting `ssi_function` at line 193 is necessary for `ssi_shutdown()` to display the footer.

bit2shift:
From [SMF's function DB on `ssi_shutdown()`](https://support.simplemachines.org/function_db/index.php?action=view_function;id=510):
image

Sesquipedalian:
> If you/ really+/ want to invoke `isset($_GET['ssi_function'])` 3 times in a row instead of 1, alright!

You make a good point. On further consideration, LGTM.

Sesquipedalian (approved):

Sesquipedalian (changes requested):
Only thing to do is remove the trailing whitespace at the end of the file.

bit2shift:
I'll have to do that from the desktop. The GitHub editor forces a trailing newline upon committing.

Sesquipedalian:
I did not know that. That's rather annoying.

bit2shift:
That should do it.

Sesquipedalian (approved):
LGTM

bit2shift:
Ubuntu 12.04 doesn't have boost 1.49, which is required for HHVM.
The PPA needs to be added manually with [`sudo add-apt-repository ppa:mapnik/boost`](https://github.com/facebook/hhvm/wiki/Building-and-installing-HHVM-on-Ubuntu-12.04/537e5a8840f8377505dc4080e5b63467b88aea6f#installing-boost-149).

jdarwood007:
Is that what it needs?  Thanks, I haven't had time to look into fixing HHVM.
I was going to move us to the trusty builds for travis, but they don't have PHP 5.3, and I haven't had time to figure that out.

Your welcome to update our travis.xml file with that added ppa with this PR as well.

bit2shift:
image

bit2shift:
On a further note, HHVM builds started failing on the 4th of May.
A closer inspection reveals that Travis switched from HHVM 3.6.6 to HHVM 3.18.2, hence the reason for the failures.

Sesquipedalian:
Nicely done

PR #4043, opened 2017-05-10
Preserves image attribs on attachments when toggling editor view

Sesquipedalian:
Previously, if the text of a post contained something like `[attach width=600 name=outtosea.png type=image/png]4[/attach]`, toggling the editor between WYSIWYG and source views would remove the `width=600` parameter (same for height, title, and alt parameters). This fixes that.


colinschoen (approved):
Nice fix. LGTM, approval

PR #4044, opened 2017-05-11
Lets user specify width and height when inserting image attachments

Sesquipedalian:
This PR does two things:

1. Lets user specify width and height when inserting image attachments. Two extra input fields are shown for image attachments so the user can type in an optional value for each. If either field contains a numerical value when the "Insert" button is clicked, the entered value will be used as a width/height parameter on the inserted BBCode.
2. Rearranges the layout of the attachments list, which was kinda messy before.


Sesquipedalian:
Here's a screenshot. If the user were to click the insert button right now, `[attach width=480 name=abusimbel.jpg type=image/jpeg]8[/attach]` would be inserted into the post's body.

screen shot 2017-05-11 at 4 31 26 pm

XinYenFon:
IMO, separation of insert & delete buttons are kinda against SMF's usage. We always put related buttons next-to each. While I don't have official voice anymore, I suggest this keeps that way (at least for this version) to keep consistency.

Also I can't see the image-size inputs, I'm on 25831bae... which seems the latest commit.

Sesquipedalian:
> IMO, separation of insert & delete buttons are kinda against SMF's usage. We always put related buttons next-to each. While I don't have official voice anymore, I suggest this keeps that way (at least for this version) to keep consistency.

Fair enough. I'll change that. :)

> Also I can't see the image-size inputs, I'm on 25831ba... which seems the latest commit.

It only shows up for attachments that have a MIME type that starts with `image`. Can you send the file that is failing so I can test with it, @Antes? The email address I use to sign my commits will be fine.

XinYenFon:
chrome_2017-04-16_16-58-51

Sesquipedalian:
Thanks, @Antes. Fixed now.

PR #4046, opened 2017-05-12
Fix MS Smart Quotes Converter Bug

### LINKED ISSUE #4045: Smart quotes converter bug
In the bowels of Subs.php is a function called sanitizeMSCutPaste, which contains the following definition:

```
	$findchars_utf8 = array(
		"\xe2\80\x9a",	// single low-9 quotation mark
		"\xe2\80\x9e",	// double low-9 quotation mark
		"\xe2\80\xa6",
```
These lines should contain `\x80` rather than `\80` for them to actually detect the bytes correctly.

colinschoen:
Fixes #4045 


bit2shift:
Wouldn't it be better if `check-eof.php` was fixed to allow *an optional* trailing newline?
\*cough* `'~\?>\s{2,}$~'` \*cough*

And with that, the next `preg_match` would have to be `'~\?>~'`.
Or better yet, merge both into `'~\?>\n$~'`, since a trailing newline is a *de facto* standard.

Sesquipedalian (approved):
LGTM

Sesquipedalian:
@bit2shift, you will probably find the discussion in #3865 helpful.

bit2shift:
Yet @albertlast didn't correctly put a trailing newline on the files he removed the closing tag from, like he should have.
SMF's package manager should be relying on `\n$` to undo stuff near EOF.

Sesquipedalian:
Perhaps so. It would likely be better to open a new issue for this, @bit2shift, rather than hijacking this discussion. :)

PR #4048, opened 2017-05-14
Adds a float bbc

Sesquipedalian:
This BBC creates a div that floats to the left or right. It takes the following forms:
```
[float=left][/float]
```
```
[float=left max=45%][/float]
```

- The value of the main parameter can be either `left` or `right`.
- The optional `max` parameter sets the max-width of the floated div. If not set, the default value of `45%` is used.
- The units for the max parameter can be any type of CSS unit: `%`, `px`, `em`, `vw`, etc. If a number is given without units, the unit defaults to `px`.
- On narrow screens, the max-width of the div will be overridden to a value of `100%` in order to prevent goofy looking layouts.
- Editor toolbar buttons are included.
- In the editor's WYSIWYG mode, the floated div has a dashed grey border to give the user some visual feedback.


Sesquipedalian:
Post display in desktop sized browser window:
screen shot 2017-05-14 at 1 40 48 am

WYSIWYG editor in desktop sized browser window:
screen shot 2017-05-14 at 1 44 06 am

Plain text editor:
screen shot 2017-05-14 at 1 59 07 am


Post display in narrow browser window (less than 480px wide):
screen shot 2017-05-14 at 1 41 44 am

WYSIWYG editor in narrow browser window (less than 480px wide):
screen shot 2017-05-14 at 1 45 08 am

XinYenFon:
Seems the BBC ignoring the line break for first character (kinda confused here actually 😆) ?

Content;
```
[float=left]Some Words[/float]
[float=right]Another[/float]



Some line
And a another one.
And a another one.
```

Output;
screenshot-localhost-2017-05-16-01-49-35

Sesquipedalian:
That's odd. By copying and pasting your input, I got this output:
screen shot 2017-05-15 at 5 00 25 pm

It looks like your test copy of index.css is somehow missing this bit:
```
.bbc_float {
	clear: both;
	margin-bottom: 1em;
	/* default max-width to use if none was specified in the BBC */
	max-width: 45%;
}
.bbc_float.floatleft {
	margin-right: 1em;
}
.bbc_float.floatright {
	margin-left: 1em;
}
```

XinYenFon:
But why you are clearing both side? It'll just convert the float values into block, still the line breaks should be adding "br" as expected.

Sesquipedalian:
I don't really understand your question yet. `<br>` is treated as an inline element. Clearing the floated `<div>` elements doesn't affect the `<br>` elements at all.

Sesquipedalian:
The reason I set the floated `<div>`s to `clear: both` is that, in the context of writing a post, using a float like this is probably going to be used for things like asides and "pull quotes" in longer, blog-style posts. For that kind of use, the post author probably doesn't want two floated elements to stack horizontally next to each other, but to make a pleasant reading layout.

But I suppose I could change it so that the floatleft elements only clear to the left, and the float right elements only clear to the right.

Sesquipedalian:
Hmm. Are you talking about how the extra line breaks are disappearing after the floats in your example, @Antes? If so, I think that is actually a general bug with how `parse_bbc()` deals with line breaks after block level BBCodes. 

For example:
```
[quote]Some Words[/quote]




Some line
And a another one.
And a another one.
```
Result:
screen shot 2017-05-15 at 5 55 10 pm

XinYenFon:
Well that's another topic i guess ^^.

Sesquipedalian:
> Well that's another topic i guess ^^.

Agreed. :)

So, leaving aside that separate bug, how does this PR look to you now?

XinYenFon:
Here let me give you my mark of approval fine sir: 🐾

Sesquipedalian:
lol

PR #4051, opened 2017-05-16
Clears the floats only on the side they are floated to

Sesquipedalian:
This was supposed to be included in cc45cd7, but I forgot to add the changes in this file


XinYenFon:
🐾

PR #4055, opened 2017-05-16
Fixes #4052

### LINKED ISSUE #4052: Deleting too much white space after BBCs with trim=outside
As seen [here](https://github.com/SimpleMachines/SMF2.1/blob/9fcacbe8a7e7d6fca1c71110ebbbfe6fd05750d0/Sources/Subs.php#L2133), `parse_bbc()` deletes `<br>` elements and white space after BBCs with `trim` set to `outside` (or `both`). What we want is to delete only one line of trailing white space, but in fact we are deleting *all* the `<br>` elements and white space until we run into something else. This means that it is impossible to add extra line breaks after, e.g., a `[quote]` BBC. For comparison, SMF 2.0.x behaves as intended.

The weird thing is, the only change in these lines of code between 2.0.x and 2.1 is `<br />` → `<br>`. In fact, the current behaviour in 2.1 appears to be exactly what these lines of code have always said to do. As far as I can tell, it is actually **2.0.x** that is behaving in a buggy fashion! My only guess is that there is actually some other bug in 2.0 that is cancelling out this one.

At any rate, I think we need to change these lines so that we delete only one line's worth of white space after a BBC that trims outside whitespace.

Sesquipedalian:
After block level BBCs that trim outside whitespace but do not have required parents, such as [quote], [float], etc., we only want to trim one line of whitespace. But after ones that do have required parents, such as [li], [td], etc., we want to trim out all trailing whitespace and line breaks until we reach the next element. For reasons I haven't bothered to fully figure out, something about the way parse_bbc() worked in SMF 2.0.x made that work even though the code said to trim all the trailing whitespace regardless. At any rate, in 2.1 we need to state explicitly what we actually want to happen, so this PR does that.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -2131,16 +2131,15 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 				$pos2 = $pos - 1;
 
 				// See the comment at the end of the big loop - just eating whitespace ;).
-				if (!empty($tag['block_level']) && substr($message, $pos, 4) == '<br>')
-					$message = substr($message, 0, $pos) . substr($message, $pos + 4);
+				$whitespace_regex = '';
+				if (!empty($tag['block_level']))
+					$whitespace_regex .= '(&nbsp;|\s)*<br>';
+				// Trim one line of whitespace after unnested tags, but all of it after nested ones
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Found a better and more efficient way to do it.

Sesquipedalian:
Found a better and more efficient way to do it. In my tests, it works perfectly now.

PR #4059, opened 2017-05-17
Fix Preview Symlies fixing #4058

### LINKED ISSUE #4058: Emoji characters in message break the post preview function
Original report [here](https://www.simplemachines.org/community/index.php?topic=553707.0)

If the message body contains an emoji character (not to be confused with SMF's smileys), trying to get a message preview results in this error message:

> The message has the following error or errors that must be corrected before continuing:
> The message body was left empty.

Once posted, the message will show up perfectly well in the topic display. It seems that only the preview is affected.

albertlast:
Without this stuff i was able to place start(🌟) in the preview.
Tested on pg and mysql.
fix #4058

### DIFF
--- a/Themes/default/Post.template.php
+++ b/Themes/default/Post.template.php
@@ -630,9 +630,9 @@ function template_main()
 						{
 							// Handle the WYSIWYG editor.
 							if (textFields[i] == ', JavaScriptEscape($context['post_box_name']), ' && $("#', $context['post_box_name'], '").data("sceditor") != undefined)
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Nice, @albertlast. :)

What happens if you only get rid of the `.php_to8bit()` part and keep the `.php_urlencode()`? I suspect the conversion of higher level UTF-8 characters to 8 bit is the issue. I think it would be best to keep the url encode part if we can.

Sesquipedalian:
>What happens if you only get rid of the `.php_to8bit()` part and keep the `.php_urlencode()`? 

Never mind, I just tested it myself, and it didn't work.

Sesquipedalian (approved):
LGTM

PR #4062, opened 2017-05-20
Fixes #4060 

### LINKED ISSUE #4060: Upgrader - fix needed
While reviewing my recently merged upgrader PRs, I noticed that the net code merged was slightly different than code submitted.   (Github bug?)

I believe the issue can cause the upgrader to freeze under some circumstances &/or cause one of the templates to be displayed improperly.

I will test a fix with the intended code in the next day or two.

(My PRs were merged in a different order than submitted, and this resulted in slightly different net code....    Very interesting...)

sbulen:
A cleanup item needed after recent PRs.  

While in there, reduced the size of the debug section further, to prevent pushing the copyright info beyond the end of the screen (on a 1920x1080 using Chrome).  

Tested a few variants to make sure things work as expected - tested on a DB that needed a UTF8 conversion & on one that didn't.  Tested with & without debug info.  All seems OK.

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -2898,7 +2898,7 @@ function ConvertUtf8()
 		}
 	}
 	$_GET['substep'] = 0;
-	return true;
+	return false;
 }
 
 function serialize_to_json()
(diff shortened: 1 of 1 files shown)


sbulen:
Thanks, @Oldiesmann !

PR #4065, opened 2017-05-23
Update Subs-Db-mysql.php

Underdog-01:
- any modification or future SMF update that includes additions/changes for where clauses contained within the $user_info array (using the query_ prefix) will be included using this method.
- Imo this should also be applied to the SMF 2.0.15 update

Author:    Underdog <github.underdog@gmail.com>
Date:      Mon May 22 23:07:25 2017 -0400

On branch patch-22

Changes to be committed:
modified:   Sources/Subs-Db-mysql.php


albertlast:
Any reason why not the Subs-Db-postgresql is touched?
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs-Db-postgresql.php#L131

Sources/Subs-Db-mysql.php
```diff
@@ -172,11 +172,9 @@ function smf_db_replacement__callback($matches)
 	if ($matches[1] === 'db_prefix')
 		return $db_prefix;
 
-	if ($matches[1] === 'query_see_board')
-		return $user_info['query_see_board'];
-
-	if ($matches[1] === 'query_wanna_see_board')
-		return $user_info['query_wanna_see_board'];
+	$user_queries = array_values(preg_grep('/^query_/', array_keys($user_info)));
```
Sesquipedalian: No need to use a regex here. It'd be much more efficient to use strpos() in this case.

Sesquipedalian (commented):
I'm not sure that this is a very helpful change. At the very least, strpos() should be used rather than preg_grep(). But moreover, since we already know the only two permitted values, the current method of doing strict equality tests and returning immediately on success is the most efficient approach possible.

Sesquipedalian:
Can you give an example, @Underdog-01, of how a mod would want to use this change? What sort of "query_*" WHERE clause placeholder do you imagine a mod would want to add?

jdarwood007:
@Sesquipedalian,

I can give an example from SimpleDesk, where we do use a query_see_ticket.  To accomplish our needs we use a wrapper around queries that make use of it.  Although I myself am not sure of this implantation being the correct method of doing so.

```
/**
 *	Query wrapper around $smcFunc['db_query'] for helpdesk specific operations.
 *
 *	This function provides a basic wrapper around SMF's internal $smcFunc['db_query'] function, adding the parameter {query_see_ticket}
 *	to it, specifically so that ticket visibility can be enforced in a query without being aware of the specific rules for the user.
 *
 *	If not previously loaded, user permissions are loaded with {@link shd_load_user_perms()}.
 *
 *	@param string $identifier SMF-style query identifier for database backend-specific replacements
 *	@param string $db_string Standard SMF 2.0 style database query, including {query_see_ticket} if appropriate
 *	@param array $db_values Standard SMF 2.0 style hash map of parameters to inject into the query
 *	@param resource $connection A database connection variable, if supplied, to override the one used by SMF by default
 *	@return resource Standard database query resource, suitable for processing with other $smcFunc['db_*'] functions
 *
 *	@see shd_load_user_perms()
 *	@since 1.0
*/
function shd_db_query($identifier, $db_string, $db_values = array(), $connection = null)
{
	global $user_info, $smcFunc;

	$replacements = array(
		'{query_see_ticket}' => $user_info['query_see_ticket'],
	);

	$db_string = str_replace(array_keys($replacements), array_values($replacements), $db_string);
	$db_values['user_info_id'] = $user_info['id'];

	return $smcFunc['db_query']($identifier, $db_string, $db_values, $connection);
}

```

Underdog-01:
I have had to manually edit popular mods due to the change.
SMF Arcade & SMF Project tools also use it. SP might use it as well but can't remember atm if I had to make an edit for it. 
This likely goes for other mods those authors have developed and also those I have not mentioned.
ie. Niko, Sinan, Arantor, etc.

SMF itself uses the $user_info array to pass info for db queries in this fashion for quite some time now so it makes sense to me to allow dynamic usage for mod authors.
I will change it to use strpos but I do so like one liners. 
Everyone so concerned about those milliseconds adding up.

albertlast:
I keep missing an awnser of my question,
where keep the support of this for postgres?

Underdog-01:
@albertlast: done.

Sesquipedalian:
Okay, that explanation makes sense. Just trim off the trailing whitespace after the closing `?>`, and this should be good to go, in my opinion.

Underdog-01:
done.
My editor also removed some other rogue white-spaces

Sesquipedalian (approved):
LGTM

colinschoen (approved):
LGTM approval

albertlast (approved):

PR #4068, opened 2017-05-25
Set Default SQL_MODE

albertlast:
set the sql_mode at the default value of the newest version from mysql:
https://dev.mysql.com/doc/refman/5.7/en/faqs-sql-modes.html
three motivation:
- the different mysql got different default set of sql_mode to garentee the same bevahior on
the different mysql version we had to set something
- take the newest set of sql_mode because ther will be the comon version of the life time of smf2.1
- higher chance that queries which run on mysql run on pg too

because i did already many thing for pg where i clean the queries,
the impact on smf2.1 is not high,
but many addon could run in issues with this change.

btw set sql_mode and set SESSION sql_mode are the same,
in my eyes it's only cleaner that we do only session stuff.

### DIFF
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -30,7 +30,7 @@ if (!defined('SMF'))
  */
 function smf_db_initiate($db_server, $db_name, $db_user, $db_passwd, $db_prefix, $db_options = array())
 {
-	global $smcFunc, $mysql_set_mode;
+	global $smcFunc;
 
 	// Map some database specific functions, only do this once.
 	if (!isset($smcFunc['db_fetch_assoc']))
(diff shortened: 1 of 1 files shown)


jdarwood007:
I like the idea, but what sort of breakage could we possibly see from this?  Beta 3 is tagged and intentions would be to go RC after that I think.  I would worry about this possibly causing issues.  Although it may solve some if MariaDB and PecronaDB support them as well and it adjusts/fixes issues with them.

A good question for this is if we are using persistent connections.  Do we need to do this every time we startup SMF, or only if we are using a new connection?

albertlast:
well on smf 2.1 we shouldn't see any breaks,
like i said the queries i fixed already because pg stuff.

On addon/portal side queries which a writing wrong will now stop working,
but this problem could even raised up without this changes (see point one in my list).

the common issue in smf 2.1 was the missing col in group by.
https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_only_full_group_by

Normaly you need only this once per connection,
but in my eyes is the handling of "is this connection new or already existing" more complex
instead of setting this value on every run again.

Sesquipedalian:
> well on smf 2.1 we shouldn't see any breaks,
like i said the queries i fixed already because pg stuff.
On addon/portal side queries which a writing wrong will now stop working,
but this problem could even raised up without this changes (see point one in my list).

That's good enough for me.

PR #4069, opened 2017-05-26
Fixed a few SM Stat collection bugs

jdarwood007:
This resolves the upgrader breaking the stat collection.  As well fixes a HTTP 1.1 RFC violation of issuing a close http header after sending data.


colinschoen (changes requested):
LGTM approval

colinschoen (approved):
Mistaken request changes.

jdarwood007:
All good.  Want to leave it open until we decide if we want to bring the code and UI for this into the ACP so admins can enable/disable at will.

jdarwood007:
This should resolve all bugs and add the ability to turn it on/off in the ACP.

other/upgrade_2-1_mysql.sql
```diff
@@ -1351,6 +1351,36 @@ DELETE FROM {$db_prefix}themes
...
+	// First get the original value
+	$request = $smcFunc['db_query']('', '
+		SELECT value
+		FROM {db_prefix}settings
+		WHERE variable = {literal:allow_sm_stats}');
+	if ($smcFunc['db_num_rows']($request) > 0 && $row = $smcFunc['db_fetch_assoc']($request))
+	{
+		if (!empty($row['value']))
+		{
+			$smcFunc['db_insert']('',
```
albertlast: Any reason why not use replace methode?

other/upgrade_2-1_postgresql.sql
```diff
@@ -1362,6 +1362,36 @@ DELETE FROM {$db_prefix}themes
...
+	// First get the original value
+	$request = $smcFunc['db_query']('', '
+		SELECT value
+		FROM {db_prefix}settings
+		WHERE variable = {literal:allow_sm_stats}');
+	if ($smcFunc['db_num_rows']($request) > 0 && $row = $smcFunc['db_fetch_assoc']($request))
+	{
+		if (!empty($row['value']))
+		{
+			$smcFunc['db_insert']('',
```
albertlast: Any reason why not use replace methode?

jdarwood007: The values should not exist prior to the conversion and the original value doesn't exist after the conversion or if you choose to enable stat collection prior to starting the db upgrade.

albertlast: in my eyes you get no benefit,
because your beliefe "ther should be no value".
but why risking that instead of use replace and "you don't care what ever ther is".

Sources/Stats.php
```diff
@@ -779,7 +779,7 @@ function SMStats()
 		'topics' => $modSettings['totalTopics'],
 		'boards' => 0,
 		'php_version' => $serverVersions['php']['version'],
-		'database_type' => strtolower($serverVersions['db_server']['title']),
+		'database_type' => strtolower($serverVersions['db_engine']['version']),
```
albertlast, line 782: With this you get the engine but not any more the database version.
In my eyes is version more importent as the engine.
Engine with version would be fine too.
But maybe you should splitt this in two information,
database_type (with the version nummer for the sake of backward compatibility)
database_vendor to know which product.

jdarwood007: The ['db_server']['title'] will only return things like 'mysqli' or 'postgresql'.  Not really useful for it to say 'mysqli' when we need to know 'mysql', 'percona', or 'mariadb'.  ['db_engine']['version'] sends that of those as it is what the engine is reported as from the database itself.  The title is derived from $smcFunc['db_title'] directly and it is something we control in our Subs-Db-*.php directly.  The db_engine comes from $smcFunc['db_get_engine'].  This data is pulled from Subs-Admin.php in the getServerVersions funciton.

In Example on my test server, it would send this data:
  'database_type' => string 'mysqli version' (length=14)
  'database_version' => string '5.6.36-82.0-log' (length=15)

After the change it sends:
  'database_type' => string 'percona' (length=7)
  'database_version' => string '5.6.36-82.0-log' (length=15)

To me that feels like the correct information.

albertlast: I miss somehow the database_version...

albertlast (approved):

Sesquipedalian:
Is this good to go now, @jdarwood007?

jdarwood007:
Yes, the 2.1 version is ready.  I've got to fix up the 2.0 stuff still, but that will come in the patch later on.

jdarwood007:
I guess the only other thing I can think of, is if we should make this default on.

I really think we should as stat collection of things like php versions can help us make better decisions regarding what we should be supporting.  If we see enough low usage of a specific version of a software, ending its support would be better than continuing to attempt support for it.

Sesquipedalian:
> I guess the only other thing I can think of, is if we should make this default on.

I think so.

PR #4070, opened 2017-05-27
Truncate excessively long URLs

### LINKED ISSUE #4049: Errors when trying to log online activity with really long URLs
Today a would-be hacker attempted to get into my forum by appending a bunch of SQL to the URL parameters. It didn't work at all, but it did make an unnecessary mess in my error log.

Specifically, because the appended SQL made the URL really long, it was too big to fit into the url field in the ${db_prefix}_log_online table. Because the hacker made a series of automated attempts, my error log was flooded with complaints about the URL being too long.

One way to avoid the unnecessary error messages would be to truncate the URL string to a reasonable length before trying to insert it into the log_online table. The potential downside of this approach would be that in cases where the URL being accessed exceeded the maximum length, the log wouldn't report the entire URL. I'm not sure if this would be a problem, though.

Another option would be to raise the limit on the field. Right now it is a VARCHAR field with a length of 1024. But no matter how high we raised it, there would always be the possibility that an even longer URL would cause this problem.

Right now I am leaning towards the truncation option, but I am wondering if anyone else has some thoughts on the matter.

colinschoen:
Truncates array that is then serialized and stored in the `log_info` table's `url` column. 

@Sesquipedalian (tagged since I can't request a review from you)

Fixes #4049


albertlast:
i got a theoritcal issue in my mind where i dont know how to test this...
since url use utf-8 (or somthing similiar) ther could be char which are in the mb4 charset.

when i got url with mb4 char than the size of the string get bigger when i save them in mysql (no pg issue here).
exp: 
native
😐 = size one
html style
& #128528; (added space after the &) = size 10

colinschoen:
Hmm strlen(😐) is `4`. 

https://gist.github.com/anonymous/ebcc0514431da2f5e181332d7d468492

Load with ?test=😐

albertlast:
well when i run this in pg: select length('😐')
i got size one but 4 > 1 so should be no issue.

but for myql:
And how long is this when you let them run throw:
https://github.com/SimpleMachines/SMF2.1/blob/8a60e5ee9a8a04adae9150115537a56f4b039ed7/Sources/Load.php#L87
?

maybe should i mention that a smcFunc('db_mb4') exists that give your the hint if the database is supporting mb4 or not.
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs-Db-mysql.php#L60

Sesquipedalian:
Actually, URLs have to be in the ASCII character set. It is permissible in an HTML 5 document to write a URL involving non-ASCII characters using the UTF-8 character set (that is, to write it as though it were an IRI instead of a URI), but the browser will re-encode it to ASCII. Any domain name involving non-ASCII characters is converted to an ASCII representation using the Punycode algorithm, while the rest is percent-encoded.

The upshot of that for our purposes here is that a non-ASCII character can never appear in a valid URL. Only its percent-encoded representation can. So we don't need to worry about higher level UTF-8 characters messing anything up.

colinschoen:
Good to know. Let me know if you think any changes are needed.

sbulen:
Thanks for the helpful explanation!

albertlast (approved):

albertlast:
@Sesquipedalian but how this is with get paremeter,
when i place there ?asdf=😐
then it's look like the information is utf8 and not asci

albertlast (changes requested):
revoke my approvement

colinschoen:
Bump @Sesquipedalian when you get a chance

Sesquipedalian:
> @Sesquipedalian but how this is with get paremeter,
> when i place there ?asdf=😐
> then it's look like the information is utf8 and not asci

I just went to `http://stovell.noip.me/~jon/dev/index.php?asdf=😐`, and the `log_online` table shows this in the `url` field:
```
{"asdf":"&#128528;","USER_AGENT":"Mozilla\/5.0 (Macintosh; Intel Mac OS X 10_12_4) AppleWebKit\/603.1.30 (KHTML, like Gecko) Version\/10.1 Safari\/603.1.30"}
```

As you can see, the emoji character was changed to an entity on the fly. :)

Sesquipedalian:
> Bump @Sesquipedalian when you get a chance

I just got back from my trip this weekend. Still catching up on things.

One thing I noticed with this code is that if a parameter with an excessively long string is nestled in between shorter parameters, everything after the long param is truncated. For example, when I go to `http://stovell.noip.me/~jon/dev/index.php?board=1.0;asdf=Lorem_ipsum...a_augue;sort=replies` (with a bunch more characters in the `asdf` param), the log only shows `{"board":1,"USER_AGENT":"Mozilla\/5.0 (Macintosh; Intel Mac OS X 10_12_4) AppleWebKit\/603.1.30 (KHTML, like Gecko) Version\/10.1 Safari\/603.1.30"}`. The sort parameter isn't logged at all.

Instead of doing an all-or-nothing approach to adding elements to the array returned from `truncateArray()`, would it be possible to truncate each element to a reasonable length and then return that? Then we could still list all the parameters, even if some have truncated values.

... But on the other hand, I'm not sure if it is worthwhile to do that much work for the `log_online` table. The contents of that table are pretty ephemeral, after all.

Sesquipedalian:
Actually, it turned out not to be so much work after all. This seems to work:
```
function truncateArray($arr, $max_length=2048)
{
	$curr_length = array_sum(array_map("strlen", $arr));
	if ($curr_length <= $max_length)
		return $arr;
	else
	{
		// Truncate each element's value to a reasonable length
		$param_max = floor($max_length/count($arr));

		foreach ($arr as $key => &$value)
			$value = substr($value, 0, $param_max - strlen($key) - 5);

		return $arr;
	}
}
```

Sesquipedalian (changes requested):
It'd be better to truncate long values to a reasonable size rather than dropping the overly long value and all subsequent ones. See comment in main discussion for a suggestion on how we might do this.

colinschoen:
But do we really want to change values of parameters. Keep in mind the only time a user should be having something truncated is if it is excessively long, likely an intrusion attempt or a bot.

Sources/Logging.php
```diff
@@ -17,6 +17,29 @@
 	die('No direct access...');
 
 /**
+ * Truncate the GET array to a specified length
+ * @param array $arr The array to truncate
+ * @param max_length $max_length The upperbound on the length
+ *
+ * @return array The truncated array
+ */
+function truncateArray($arr, $max_length=2048) {
```
Sesquipedalian: If the maximum length for the database field is 2048 characters, the value of `$max_length` here should be less so that there will always be enough room for the user agent string, etc.

Sesquipedalian:
> But do we really want to change values of parameters. Keep in mind the only time a user should be having something truncated is if it is excessively long, likely an intrusion attempt or a bot.

The truncation will still only be triggered if the overall length is too much. In a situation where this happens, I'd think that knowing what all the URL parameters were, even if the full parameter value is truncated, would be more valuable information.

colinschoen:
Ok. I agree. Updated. Changed `max_length` to 1900 for some wiggle room.

Sesquipedalian (approved):

colinschoen:
Thanks for reviewing.

PR #4073, opened 2017-05-30
Fix image proxy from throwing 404 on initial image load.

colinschoen:
Credit: https://www.simplemachines.org/community/index.php?topic=553914.msg3926081#msg3926081


tinoest:
I'm not sure this is the best solution reading the whole file, there are lots of places it will return false and those are all checked by is_int, as per the comment in the forum thread this won't check correctly. 

I would change it so that a true or false is returned and to check against that. 
`
		if ($responseCode != 200) {
			return false;
		}
`

Should allow you to do that along with reverting the last change. 

Then a 
`
if (!$response) {
		// Throw a 404
		header('HTTP/1.0 404 Not Found');
		exit;
}
`

As you don't seem to care about what the response code was from the curl request.

colinschoen:
Agreed, thanks. This should be fixed now.

PR #4076, opened 2017-05-31
Message Index sorting by invalid columns

### LINKED ISSUE #4075: Non-default sorting on middle pages
Sometimes when sorting the message index by specific columns and you have several pages + aren't on the first or last page, it'll try to `order by` invalid column names.

> http://example.com/index.php?board=14.440;sort=last_post
> Unknown column 'tid_last_msg' in 'order clause'
> File: Sources/MessageIndex.php Line: 305

It applies the `identifier` sanitization method to functions as well:
> http://example.com/index.php?board=14.440;sort=starter
> Unknown column 'COALESCE(memfreal_name, mfposter_name)' in 'order clause'
> File: Sources/MessageIndex.php Line: 305

DuncanFairley:
Fixes #4075
The values in `$_REQUEST['sort']` are already pulled from `$sort_methods` so they're already safe.

### DIFF
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -272,7 +272,7 @@ function smf_db_replacement__callback($matches)
 
 		case 'identifier':
 			// Backticks inside identifiers are supported as of MySQL 4.1. We don't need them for SMF.
-			return '`' . strtr($replacement, array('`' => '', '.' => '')) . '`';
+			return '`' . strtr($replacement, array('`' => '', '.' => '`.`')) . '`';
 		break;
 
 		case 'raw':
--- a/Sources/MessageIndex.php
+++ b/Sources/MessageIndex.php
@@ -293,7 +293,7 @@ function MessageIndex()
 				LEFT JOIN {db_prefix}members AS meml ON (meml.id_member = ml.id_member)' : '') . '
 			WHERE t.id_board = {int:current_board}' . (!$modSettings['postmod_active'] || $context['can_approve_posts'] ? '' : '
 				AND (t.approved = {int:is_approved}' . ($user_info['is_guest'] ? '' : ' OR t.id_member_started = {int:current_member}') . ')') . '
-			ORDER BY is_sticky' . ($fake_ascending ? '' : ' DESC') . ', {identifier:sort}' . ($ascending ? '' : ' DESC') . '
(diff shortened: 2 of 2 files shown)


Sources/Subs-Db-mysql.php
```diff
@@ -272,7 +272,7 @@ function smf_db_replacement__callback($matches)
 
 		case 'identifier':
 			// Backticks inside identifiers are supported as of MySQL 4.1. We don't need them for SMF.
-			return '`' . strtr($replacement, array('`' => '', '.' => '')) . '`';
```
colinschoen, line 275: Why is this change needed now if you are using `raw`?

colinschoen (changes requested):

Sources/Subs-Db-mysql.php
DuncanFairley: I feel that the `itentifier` filter should either accept table.field syntax, or return an error - instead of just mangling the input. A couple sanitization libraries I've used seem to accept table.field.

colinschoen: Agreed.

colinschoen (approved):
LGTM approval

albertlast:
@colinschoen this pr is in my eyes wrong in many direction
it's do only mysql stuff (pg is effected to),
with the replace of the point with the point the sql issue keep there.

colinschoen:
I'll take a closer look tomorrow this does need to be fixed for PG but the solution presented works for MySQL.

albertlast:
#4077 fix this issue correctly for both databases

DuncanFairley:
In #4077 you drop the aliased table name which would break other sort columns from `$sort_methods` like `mf.subject`, as there's multiple `messages` tables joint you'd get an ambiguous error.

albertlast:
first important to know,
your solution works never because you had the table alias and the col nam in the same quote.
see http://sqlfiddle.com/#!9/582dd8/2

the order by work that the col name of the select statement part is taken
and there can only be exists one messages all other needs to be a different name.

DuncanFairley:
But my change doesn't only put quotes around the whole thing... That was the point of my change. 
http://sandbox.onlinephpfunctions.com/code/be39d01ea1ee2214075bf7d7bbabf9a06e03f0d0
> Outputs \`table\`.\`column\`

albertlast:
And why did you change the type of the sql statement from identifier to raw?

albertlast:
okay gotcha because of the coalesec

PR #4078, opened 2017-05-31
Fix the point handling in identifier for pg

albertlast:
Copy the changes from mysql

### DIFF
--- a/Sources/Subs-Db-postgresql.php
+++ b/Sources/Subs-Db-postgresql.php
@@ -227,7 +227,7 @@ function smf_db_replacement__callback($matches)
 		break;
 
 		case 'identifier':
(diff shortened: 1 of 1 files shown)


Sources/Subs-Db-postgresql.php
```diff
@@ -227,7 +227,7 @@ function smf_db_replacement__callback($matches)
 		break;
 
 		case 'identifier':
-			return '"' . strtr($replacement, array('`' => '', '.' => '')) . '"';
+			return '"' . strtr($replacement, array('`' => '', '"."' => '')) . '"';
```
colinschoen: Are you sure this is what you want? Don't you want to change the value, not the key in the replace.

colinschoen (changes requested):

colinschoen (approved):
LGTM approval

PR #4082, opened 2017-06-01
Added gui for pg fulltext index

albertlast:
It's frontend extension for the pr: https://github.com/SimpleMachines/SMF2.1/pull/4012

first i placed this setting under database stuff,
than i got the idea "would be cool in search setting"
moved and realize that there not select list exists,
that is the reason why this setting now again unter database settings.

plaese check the text stuff.
@Sesquipedalian maybe for your intresting

grafik

### DIFF
--- a/Themes/default/languages/Help.english.php
+++ b/Themes/default/languages/Help.english.php
@@ -1,7 +1,7 @@
 <?php
 // Version: 2.1 Beta 3; Help
 
-global $helptxt;
+global $helptxt, $scripturl;
 
 $txt['close_window'] = 'Close window';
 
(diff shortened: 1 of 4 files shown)


jdarwood007:
So if I understand this properly, PG has a language that the search index is in?

How do multi-lingual forums handle this with PG?  Does one language take precedence over others and the rest suffer?  I would think it wouldn't care and would return results being fully UTF-8 compatible.

albertlast:
Every database (also mysql) need a language for the index creating/using.
In mysql exists "one fit all" and in you had the possible to choose.
For this exists in pg the lanuage "simple" which do the same job,
so when you got a multi language board simple could be the right choice.

The language setting to normalize the text col,
make things from pural to singular or drop words how a common like "a".

All this stuff is not related to utf-8,
so when you got a englisch index and search for chines word you will find this.

Themes/default/languages/Help.english.php
```diff
@@ -403,6 +403,10 @@
 	</ul>';
 
 $helptxt['autoFixDatabase'] = 'This will automatically fix broken tables and resume as if nothing happened. This can be useful, because the only way to fix it is to REPAIR the table, and this way your forum won\'t be down until you notice. It does email you when this happens.';
+$helptxt['pgFulltextSearch'] = 'This setting allow your to define the language wich is used for the fulltext search of postgresql, the language settings are independent from your smf settings and not related to the installed language for smf. <br/>
+When you change the language you need to rebuild the fulltext search index to find under Search -> Search Method. (Fulltext Index)<br/>
+The listed language based on the postgresql installation,
+when you miss a language ask your database admin.';
```
Sesquipedalian: How about the following?
```
$helptxt['pgFulltextSearch'] = 'This setting defines the language to use for PostgreSQL\'s fulltext search. Choose the language that most closely matches the language your forum actually uses. If your forum\'s language is not listed, or if your forum is multi-lingual, choose the "simple" option. <br/>
This setting is independent of your main SMF language settings and not related to the installed language for SMF. <br/>
When you change this setting you need to [url=' . $scripturl . '?action=admin;area=managesearch;sa=method]rebuild the fulltext search index[/url].<br/>
If a language you need is not listed, ask your database admin to install PostgreSQL language support for that language.';
```

You'll need to add $scripturl to the globals at the start of this file, of course.

Sesquipedalian:
The code looks good, but the help text could be improved. See my review comment for a suggestion of how this could go.

Sesquipedalian (approved):

PR #4087, opened 2017-06-03
added insert as method for return

albertlast:
for me was methode  '' = insert
forgot that methode 'insert' exists too.

### DIFF
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -768,7 +768,7 @@ function smf_db_error($db_string, $connection = null)
  * @param array $columns An array of the columns we're inserting the data into. Should contain 'column' => 'datatype' pairs
  * @param array $data The data to insert
  * @param array $keys The keys for the table
- * @param int returnmode 0 = nothing(default), 1 = last row id, 2 = all rows id as array; every mode runs only with method = ''
+ * @param int returnmode 0 = nothing(default), 1 = last row id, 2 = all rows id as array; every mode runs only with method = '' or 'insert'
  * @param object $connection The connection to use (if null, $db_connection is used)
  * @return value of the first key, behavior based on returnmode
  */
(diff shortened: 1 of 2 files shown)


Sources/Subs-Db-mysql.php
```diff
@@ -825,7 +825,7 @@ function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $ret
 		$connection
 	);
 	
-	if(!empty($keys) && (count($keys) > 0) && $method == '' && $returnmode > 0)
+	if(!empty($keys) && (count($keys) > 0) && ($method === '' || $method === 'insert') && $returnmode > 0)
```
Sesquipedalian: In this case, it would be more efficient to test if `$queryTitle === 'INSERT'` rather than if `($method === '' || $method === 'insert')`

Sesquipedalian (changes requested):

albertlast:
Well i like to be that the check is the same on mysql and pg side.
pg doesn't hav this line of code.

Sesquipedalian:
I see your point, yes. My suggestion would only make the difference of a few microseconds, after all.

Sesquipedalian (approved):

Sesquipedalian:
There's a conflict that needs to be resolved before merging.

Sesquipedalian:
Hey @colinschoen or @live627, do you know why Travis is complaining about an unsigned commit in bdbd6c3? It's just a merge commit, so it shouldn't need to be signed.

Oldiesmann:
@jdarwood007 Any ideas on why travis is failing on the merge commit?

colinschoen:
Does travis even differentiate between merge commits and other commits? They look the same.

jdarwood007:
Because Travis is looking at the pr merge and the merge itself for comments mentioning the sign off by.  It can fail as it only looks so far back in the history.  Its more difficult for me to program the automatic detection of the sign off than it is for us to check.

I'm thinking we could tie in https://www.clahub.com/ as a way to ensure they have signed the DCO, which I think reduces as much worry about signing off commits.  The CLA would cover the DCO and our generic CLA (if different).

PR #4088, opened 2017-06-03
Benchmark for smf

albertlast:
Something I like to wish since start of my dev,
a way the benchmark smf.

for different motivation:
- changes on database stuff (dev stuff)
- changes on hardware (change webhoster)
- changes on software releases (change of php or database version)

the second and last point i believe it's a common issue and not special like the first one.

Two modes exists:
User creation
Post creation

Both takes a minute and try to create as mutch as possible.


jdarwood007:
Should note that registerMember is a really really slow function.  I believe during my testing it came down to the password creation as hash_password was designed to be slow.

albertlast:
Well this in my eyes no problem.

I did some test in my env.

mysql
-user 81
-topic 72

postgres
-user 112
-topic 7789

Dunno why the numbers with pg so mutch higher than mysql (i believe my mysql is not so good).
But it's shows that the user creation is mutch more cpu/php heavy task is and
the topic run more database releated is.

And this in my eyes a good mix.

albertlast:
Added a third methode,
this try to access a topic x times in one minutes.

This benchmark shows realy good how many topic a user can access,
because this stuff depend on many things like size of the board and activate features.
It's show realy good the status of the board,
but this make it's hard for comparing accross with different board.

Because it's show how a mod, feature or size effect the board,
make this test veary valuable for board admins.

Also I try to add in the description what the test focus on.

Sources/ManageMaintenance.php
```diff
@@ -2285,4 +2295,257 @@ function get_hook_info_from_raw($rawData)
 	return $hookData;
 }
 
+/**
+ * Benchmakr for User creation tries to create as many as possible in 1 minute
```
colinschoen: `Benchmakr` -> `benchmark`

Sources/ManageMaintenance.php
```diff
@@ -2285,4 +2295,257 @@ function get_hook_info_from_raw($rawData)
...
+	removeTopics($postid);
+	
+	$context['benchmark_result']['amount'] = $count;
+	$context['benchmark_result']['test_name'] = $txt['benchmark_post'];
+	
+	deleteMembers($userID);
+}
+
+/**
+ * Benchmakr for Post reads tries to access as many as possible in 1 minute
```
colinschoen: `Benchmakr` -> `Benchmark`

Themes/default/languages/ManageMaintenance.english.php
```diff
@@ -230,4 +230,13 @@
 $txt['enable_maintenance1'] = 'Due to the size of your forum, it is recommended to place your forum in "maintenance mode" before you start the backup.';
 $txt['enable_maintenance2'] = 'To proceed, due to the size of your forum, please place your forum in "maintenance mode".';
 
+$txt['benchmark_usercreate'] = 'Creating User';
+$txt['benchmark_post'] = 'Creating Posts';
+$txt['benchmark_usercreate_info'] = 'Create as many users as possible in one minute <br/> This test is cpu heavy';
+$txt['benchmark_post_info'] = 'Create as many posts as possible in one minute <br/> This test is db insert imperformance heavy';
```
colinschoen: `imperformance` -> `performance`

Themes/default/languages/ManageMaintenance.english.php
```diff
@@ -230,4 +230,13 @@
 $txt['enable_maintenance1'] = 'Due to the size of your forum, it is recommended to place your forum in "maintenance mode" before you start the backup.';
 $txt['enable_maintenance2'] = 'To proceed, due to the size of your forum, please place your forum in "maintenance mode".';
 
+$txt['benchmark_usercreate'] = 'Creating User';
+$txt['benchmark_post'] = 'Creating Posts';
+$txt['benchmark_usercreate_info'] = 'Create as many users as possible in one minute <br/> This test is cpu heavy';
+$txt['benchmark_post_info'] = 'Create as many posts as possible in one minute <br/> This test is db insert imperformance heavy'; 
+$txt['benchmark_topic'] = 'Benchmark methodes';
```
colinschoen: `Benchmark methodes` -> `Benchmark methods`

Themes/default/languages/ManageMaintenance.english.php
```diff
@@ -230,4 +230,13 @@
...
+$txt['benchmark_postread'] = 'Reading Posts';
+$txt['benchmark_postread_info'] = 'Access a topic how often it can in one minute <br/> This test dipence heavy on the size of your board and active features';
```
colinschoen: `Repeatedly access a topic for 1 minute. <br /> This test depends heavily on the size of your board and the active features.`

colinschoen (changes requested):
This benchmarking has some side effects of affecting statistics and increasing unique IDs substantially if run often.

albertlast:
The statistics are not touched as fare i see in my test env.
I see no problem with raising the unique id's in db tables,
max int is 2147483647 in my test was max 8k for topic creating stuff...

jdarwood007:
Would this only have use for development/testing?

I'm not seeing this as something that is meant for a production site.  Perhaps it would be best split off into a separate utility that we could expand and maintain as part of benchmark and/or unit testing

albertlast:
i already sayed my point of in the first commit,
in short i see need on admin side too.

XinYenFon:
I disagree this is a bloated addition to core.

Sesquipedalian:
This kind of benchmarking is definitely useful, @albertlast. But I agree with the others that it would be better to develop it as a separate mod or utility. That would give you more flexibility to adjust the various types of benchmarks and add new ones over time. By the look of things, you should be able to do everything via integration hooks.

albertlast:
I take from the last comments basicly: "The dev do the thing" 
Can someone show me the person
how test smf with all variation of php version are supported,
all databases types and version,
different sizes of boards,
different smf configuration (cache setting/mode etc.),
with the different combination of mods and
hardware/webspace configuration which is smf running?

When you can not mention this guy than is my eye not provided by the dev "guy" and
the smf admin had to do this stuff.

Sesquipedalian:
What we are saying is that if you make it a mod for this, then anyone who feels the need to do some benchmarking can just install the mod and do as much as they like. It should be pretty easy to make a mod out of this, and in fact you should be able to do it entirely using integration hooks.

albertlast:
But i will create no mod,
this type of information are essential,
when this type is not essential we should drop function like calendar too.

XinYenFon:
What? How you correlate calendar with benchmark ?

colinschoen:
The issue is that 99% of forum owners won't have too much use for the utility. It is however extremely useful for the 1%, but that means it is likely better as a mod to install for those that will use it.

XinYenFon:
On the other hand, you can tweak https://github.com/SimpleMachines/tools/blob/master/Populate.php to do benchmark as well as the populate or create whole new file like populate (benchmark.php) to make it happen.

albertlast:
Populate is the wrong tool to benchmark your smf,
please don't use it for this case...

Sesquipedalian:
Using the `integrate_admin_areas` and `integrate_manage_maintenance` hooks, you should be able to make a mod for this that does everything you want, @albertlast. The consensus seems to be that this would be the best approach to follow.

albertlast:
When you mean this benchmark would be good for dev,
why you not keeping this and activate only when $db_show_debug=true; is on.

From my side will be no mod.

colinschoen:
The issue isn't that it is hidden or visible, but that it inflates the code base which we have to maintain for a feature that will likely be used by ~1% of the users. What is your hesitate to convert this into a mod?

albertlast:
Why you not create a mod the code is there?
Easy anwser the issue the support.

colinschoen:
^ Sorry, I don't understand. Can you rephrase.

albertlast:
provide a mod create many support task and i'm not willig to do so,
when you mean that is important than feel free to take the code an create mod.

colinschoen:
Sure thing I will make the mod.

jdarwood007:
We can spin up a new repo to contain the mod call it smf benchmark.  Contain it in there as a mod.  As a plus, allow us to easily call it from a benchmark.php as cron/cli.  So then us devs could pull down the mod and latest 2.1 dev, then have the cron script go through, running the benchmark and sending the results out (either in post or email).  As a plus, that benchmark.php could then be a git hook as well :)

albertlast:
@colinschoen what is the status of the mod?

colinschoen:
Low on the priority list. Planned but not started.

albertlast:
I was interesting how mod creation works in smf,
so i take this pr and created a mod: https://custom.simplemachines.org/mods/index.php?mod=4155
It's nice on the one side and on the other side it's pain.

PR #4090, opened 2017-06-04
Generic fixes

NanoSector:
This includes a lot of cleanups and various small fixes.

### DIFF
--- a/SSI.php
+++ b/SSI.php
@@ -25,9 +25,6 @@ global $db_connection, $modSettings, $context, $sc, $user_info, $topic, $board, 
 global $smcFunc, $ssi_db_user, $scripturl, $ssi_db_passwd, $db_passwd, $cachedir;
 
 // Remember the current configuration so it can be set back.
-$ssi_magic_quotes_runtime = function_exists('get_magic_quotes_gpc') && get_magic_quotes_runtime();
-if (function_exists('set_magic_quotes_runtime'))
-	@set_magic_quotes_runtime(0);
 $time_start = microtime();
 
 // Just being safe...
@@ -193,8 +190,6 @@ if (basename($_SERVER['PHP_SELF']) == 'SSI.php')
 unset($_GET['ssi_function']);
 
 error_reporting($ssi_error_reporting);
-if (function_exists('set_magic_quotes_runtime'))
-	@set_magic_quotes_runtime($ssi_magic_quotes_runtime);
 
 return true;
 
--- a/Sources/CacheAPI-sqlite.php
+++ b/Sources/CacheAPI-sqlite.php
@@ -24,8 +24,16 @@ class sqlite_cache extends cache_api
 	 * @var string The path to the current $cachedir directory.
 	 */
 	private $cachedir = null;
+
+    /**
+     * @var SQLite3
+     */
 	private $cacheDB = null;
-	private $cacheTime = null;
+
+    /**
+     * @var int
+     */
+	private $cacheTime = 0;
 
 	public function __construct()
 	{
--- a/Sources/CacheAPI-sqlite.php
+++ b/Sources/CacheAPI-sqlite.php
@@ -25,14 +25,14 @@ class sqlite_cache extends cache_api
 	 */
 	private $cachedir = null;
 
-    /**
-     * @var SQLite3
-     */
+	/**
+	 * @var SQLite3
+	 */
 	private $cacheDB = null;
 
-    /**
-     * @var int
-     */
+	/**
+	 * @var int
+	 */
 	private $cacheTime = 0;
 
 	public function __construct()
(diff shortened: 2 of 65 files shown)


Sources/Subs-Db-postgresql.php
```diff
@@ -661,9 +661,9 @@ function smf_db_unescape_string($string)
  * @param array $columns An array of the columns we're inserting the data into. Should contain 'column' => 'datatype' pairs
  * @param array $data The data to insert
  * @param array $keys The keys for the table
- * @param int returnmode 0 = nothing(default), 1 = last row id, 2 = all rows id as array; every mode runs only with method = ''
+ * @param int $returnmode 0 = nothing(default), 1 = last row id, 2 = all rows id as array; every mode runs only with method = ''
  * @param resource $connection The connection to use (if null, $db_connection is used)
- * @return value of the first key, behavior based on returnmode
+ * @return mixed value of the first key, behavior based on returnmode. null on no data
```
albertlast: Text is different in compare of the mysql description. if vs on

Sources/CacheAPI-sqlite.php
```diff
@@ -24,8 +24,16 @@ class sqlite_cache extends cache_api
 	 * @var string The path to the current $cachedir directory.
 	 */
 	private $cachedir = null;
+
+    /**
+     * @var SQLite3
+     */
```
jdarwood007: This isn't the only place I see, but lots of spaces rather than tabs are being used.

NanoSector: Blame phpStorm, sorry. I'll see if I can get this fixed.

NanoSector: Should be fixed now I think?

jdarwood007: Looks like it

SSI.php
```diff
@@ -25,9 +25,6 @@
 global $smcFunc, $ssi_db_user, $scripturl, $ssi_db_passwd, $db_passwd, $cachedir;
 
 // Remember the current configuration so it can be set back.
-$ssi_magic_quotes_runtime = function_exists('get_magic_quotes_gpc') && get_magic_quotes_runtime();
-if (function_exists('set_magic_quotes_runtime'))
-	@set_magic_quotes_runtime(0);
```
Sesquipedalian, line 30: According to install.php and upgrade.php, our minimum is still PHP 5.3.8. If there has been a recent change that is compatible only with 5.4+, we need either to revert that change or to deliberate about whether to raise our minimum PHP level. In the meantime, we need to leave the magic quotes stuff in place.

Sesquipedalian (changes requested):

Sources/Profile.php
```diff
@@ -26,8 +26,8 @@
 function ModifyProfile($post_errors = array())
 {
 	global $txt, $scripturl, $user_info, $context, $sourcedir, $user_profile, $cur_profile;
-	global $modSettings, $memberContext, $profile_vars, $post_errors, $user_settings;
-	global $db_show_debug, $smcFunc;
+	global $modSettings, $memberContext, $profile_vars, $post_errors;
+	global $smcFunc;
```
Sesquipedalian: Why not put $smcFunc on the same line as the rest of the global variables?

Sesquipedalian (changes requested):

SSI.php
NanoSector: If 2.0.14 raised the minimum to 5.4 I don't think SMF 2.1 works with anything lower than that anyway. Especially since some code was ported from 2.1 to 2.0.14 (including the DB class causing the issue).

Sources/Profile.php
NanoSector: Dunno, will fix.

SSI.php
Sesquipedalian: I believe 2.0.14 requires a minimum of 5.3.0.

Sesquipedalian:
I'm not sure we should have merged 	b405e08, @Oldiesmann. Unless we raise the minimum PHP version to 5.4, we still need to deal with the magic quotes stuff.

XinYenFon:
@Sesquipedalian I think its best to start SMF 2.1 from php5.5 and recommend php7.

Sesquipedalian:
We should probably have a discussion about this. https://www.simplemachines.org/community/index.php?topic=554431.0

XinYenFon:
Unfortunately, I don't have access to such boards anymore.

Sesquipedalian:
Just moved it to the Development Contributors board.

XinYenFon:
I don't have access to Development boards.

Sesquipedalian:
I would think you should at least have access to Development Contributors. @colinschoen, can we fix that?

colinschoen:
Looking into this.

Sesquipedalian:
#4101

PR #4093, opened 2017-06-04
Fixed display of cached hit ratio

albertlast:
The array didn't exists any more for the next echo whre the implode use this var,
so i define this on the start of the if.

### DIFF
--- a/Sources/Logging.php
+++ b/Sources/Logging.php
@@ -260,6 +260,7 @@ function displayDebug()
 
 	if (!empty($modSettings['cache_enable']) && !empty($cache_hits))
 	{
+		$missed_entires = array();
 		$entries = array();
 		$total_t = 0;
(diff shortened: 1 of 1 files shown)


Sources/Logging.php
```diff
@@ -260,6 +260,7 @@ function displayDebug()
 
 	if (!empty($modSettings['cache_enable']) && !empty($cache_hits))
 	{
+		$missed_entires = array();
```
NanoSector: typo: missed_entries

NanoSector: it's actually broken in the original code too. might be best to just fix it to avoid confusion.

albertlast: already done

colinschoen (approved):
LGTM approval

PR #4094, opened 2017-06-04
Pg cache

albertlast:
I notice in my test env is no cache avaible by side of file cache and
this one is slow.

So I added this cache for postgresql and
again i notice that the smf db_* magic is to mutch overhead for a cache.

so i used native pg_* with prepared statment stuff to be safer as the other cache solution but be faster as
file cache solution.

### DIFF
--- a/Sources/CacheAPI-pg.php
+++ b/Sources/CacheAPI-pg.php
@@ -32,25 +32,17 @@ class pg_cache extends cache_api
 	 */
 	public function connect()
 	{
-		global $db_prefix, $smcFunc;
-
-		$request = $smcFunc['db_query']('','
-			SELECT 1 
+		global $db_prefix, $smcFunc, $db_connection;
+		
+		pg_prepare($db_connection, '', 'SELECT 1 
 			FROM   pg_tables
-			WHERE  schemaname = {string:schema}
-			AND    tablename = {string:cache}
-			',
-			array(
-			'schema' => 'public',
-			'cache' => $db_prefix . 'cache',
-			)
-		);
+			WHERE  schemaname = $1
+			AND    tablename = $2');
+			
+		$result = pg_execute($db_connection, '', array('public', $db_prefix . 'cache'));
 
-		if($smcFunc['db_num_rows']($request) === 0)		
-			$smcFunc['db_query']('',
-				'CREATE TABLE {db_prefix}cache (key text, value text, ttl bigint, PRIMARY KEY (key))',
-				array()
-			);
+		if(pg_affected_rows($result) === 0)
+			pg_query($db_connection, 'CREATE TABLE {db_prefix}cache (key text, value text, ttl bigint, PRIMARY KEY (key))');			
 	}
 
 	/**
@@ -71,21 +63,17 @@ class pg_cache extends cache_api
 	 */
 	public function getData($key, $ttl = null)
 	{
-		global $smcFunc;
+		global $db_prefix, $smcFunc, $db_connection;
 		
 		$ttl = time() - $ttl;
-		$request = $smcFunc['db_query']('','
-			SELECT value FROM {db_prefix}cache WHERE key = {string:key} AND ttl >= {int:ttl} LIMIT 1
-			',
-			array(
-				'key' => $key,
-				'ttl' => $ttl,
-			)
-		);
-		if($smcFunc['db_num_rows']($request) === 0)
+		pg_prepare($db_connection, '', 'SELECT value FROM ' . $db_prefix . 'cache WHERE key = $1 AND ttl >= $2 LIMIT 1');
+			
+		$result = pg_execute($db_connection, '', array($key, $ttl));
+		
+		if(pg_affected_rows($result) === 0)
 			return null;
 		
-		$res = $smcFunc['db_fetch_assoc']($request);
+		$res = pg_fetch_assoc($result);
 		
 		return $res['value'];
 	}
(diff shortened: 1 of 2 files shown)


Sources/CacheAPI-pg.php
```diff
@@ -0,0 +1,134 @@
...
+	public function connect()
+	{
+		global $db_prefix, $smcFunc;
+
+		$request = $smcFunc['db_query']('','
+			SELECT 1 
+			FROM   pg_tables
+			WHERE  schemaname = {string:schema}
+			AND    tablename = {string:cache}
+			',
```
NanoSector: Should be put on the previous line

NanoSector: Multiple occurrences

NanoSector:
LGTM.

Sources/CacheAPI-postgres.php
```diff
@@ -0,0 +1,129 @@
...
+			return false;
+
+		$result = pg_query($db_connection, 'SHOW server_version_num');
+		$res = pg_fetch_assoc($result);
+		
+		if($res['server_version_num'] < 90500)
+			return false;
+		
+		return true;
+	}
```
jdarwood007: Don't forget "$test = true" is what tests if we can use the API, and without it we try to actually use it.  It should fail after this if the parent doesn't show support.  The parent fails isSupported if caching is disabled.  For your code, I think the easiest is to just replace the return true with return parent::isSupported(). At least if testing so maybe need to evaluate $test first

albertlast: to be ownest i don't understand...
i take the sqlite as template and this ingored the test flag too.

jdarwood007: I think this would work:
```
return $test ? true : parent::isSupported();
```

isSupported handles 2 tests.

First where $test = true, it checks to see if this is a supported API.  This is mainly used for the ACP where we present options for ability to use the API.  We are only seeing if its possible this API is usable, such things as if the function exists.

The second is there $test = false, it does those tests and the actually makes sure its usable.  Such as if cache is enabled, or in the case of memcache it makes sure the settings exist and are valid for it to work.

Sources/CacheAPI-postgres.php
```diff
@@ -0,0 +1,129 @@
...
+		global $db_prefix, $db_connection;
+
+		pg_prepare($db_connection, '', 'SELECT 1 
+			FROM   pg_tables
+			WHERE  schemaname = $1
+			AND    tablename = $2');
+
+		$result = pg_execute($db_connection, '', array('public', $db_prefix . 'cache'));
+
+		if(pg_affected_rows($result) === 0)
```
colinschoen: `if (`

Sources/CacheAPI-postgres.php
```diff
@@ -0,0 +1,129 @@
...
+		global $smcFunc, $db_connection;
+		
+
+		if ($smcFunc['db_title'] !== 'PostgreSQL')
+			return false;
+
+		$result = pg_query($db_connection, 'SHOW server_version_num');
+		$res = pg_fetch_assoc($result);
+		
+		if($res['server_version_num'] < 90500)
```
colinschoen: `if (`

Sources/CacheAPI-postgres.php
```diff
@@ -0,0 +1,129 @@
...
+	public function getData($key, $ttl = null)
+	{
+		global $db_prefix, $db_connection;
+
+		$ttl = time() - $ttl;
+		pg_prepare($db_connection, '', 'SELECT value FROM ' . $db_prefix . 'cache WHERE key = $1 AND ttl >= $2 LIMIT 1');
+			
+		$result = pg_execute($db_connection, '', array($key, $ttl));
+		
+		if(pg_affected_rows($result) === 0)
```
colinschoen: `if (`

Sources/CacheAPI-postgres.php
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public function putData($key, $value, $ttl = null)
+	{
+		global  $db_prefix, $db_connection;
+
+		if(!isset($value))
```
colinschoen: `if (`

colinschoen (changes requested):
nits

NanoSector:
This can be autofixed by adding the spacing patches Scrutinizer suggests:
https://scrutinizer-ci.com/g/SimpleMachines/SMF2.1/inspections/0e5d1992-ce14-4490-9a2f-744abb6d3597/patches?status=new

Tried doing it myself but it freaks out.

jdarwood007:
@Yoshi2889 ,
Its why we don't use their auto fixer as it seems to cause more problems than normal.  Adjusting the settings may make it get more accurate to our coding style though.

albertlast:
@colinschoen any issue are here or is this pr fine?

Sesquipedalian:
Is this satisfactory to you now, @colinschoen?

Sources/CacheAPI-postgres.php
```diff
@@ -0,0 +1,129 @@
...
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public function getData($key, $ttl = null)
+	{
+		global $db_prefix, $db_connection;
+
+		$ttl = time() - $ttl;
+		pg_prepare($db_connection, '', 'SELECT value FROM ' . $db_prefix . 'cache WHERE key = $1 AND ttl >= $2 LIMIT 1');
```
albertlast: I would like to test if can do this prepare stuff only once and
check if this make any different runtime/performance side
--> don't merge this pr

Sesquipedalian: Okay

albertlast:
It work,
i call the prepare statment only once ~ 40% less runtime.

Now it can be merged.

albertlast:
I forgot that i'm working in a class...

PR #4086, opened 2017-06-03
Set $packagesdir & $tasksdir

sbulen:
Set $packagesdir & $tasksdir

Fixes: Installer - Paths & setting not set #3999

PR #4095, opened 2017-06-04
Define more directories to put in Settings.php

### LINKED ISSUE #3999: Installer - Paths & setting not set
Brand new downloads from git.  Just installed.  Copied the Install*.* files and settings.php to the main directory prior to install.  

Issues (/questions) noted:

(1) Several paths do not appear to be set properly in settings.php, they still say (__FILE__).  Most are set properly and specify the full path.  
(2) Repair_settings does not populate/show paths for these, it appears to be unexpected.  Repair_settings does, in fact, repair the paths if you click on the recommended values & press save.  
(3) Repair_settings complains about a setting not set - queryless URLs - should that be set by the installer?

Minor things, but I don't think you should have to go to repair_settings.php to complete a plain vanilla install.  

install-paths-1
install-paths-2
install-paths-3

NanoSector:
Should fix #3999

### DIFF
--- a/other/install.php
+++ b/other/install.php
@@ -936,6 +936,7 @@ function ForumSettings()
 			'boardurl' => $_POST['boardurl'],
 			'boarddir' => addslashes(dirname(__FILE__)),
 			'sourcedir' => addslashes(dirname(__FILE__)) . '/Sources',
+			'tasksdir' => addslashes(dirname(__FILE__)) . '/Sources/tasks',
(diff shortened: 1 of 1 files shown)


sbulen:
Is this a dupe of Set $packagesdir & $tasksdir #4086?

NanoSector:
Oh oops didn't see you already had one up :) yes, I'll close this one.

Op zo 4 jun. 2017 19:44 schreef sbulen <notifications@github.com>:

Issue #4083, opened 2017-06-02
Upgrader - fulltext search doesn't work after upgrade


sbulen:
If your forum uses a fulltext search, and a UTF8 upgrade is needed, the index is dropped by the upgrader as part of the ConvertUTF8 step.

There is logic to issue a warning message, to tell the user to re-create the index, but this message doesn't display properly.  I believe the proper fix is to get this message to display.  

My initial thought is that this is a low priority, mainly because I suspect few folks are using the fulltext search method.

sbulen:
The logic in question is specific to Mysql. 

I do not know if there is an equivalent issue with pg fulltext search???

albertlast:
that are the fact that i know so fare:
fulltext search with index for pg exists in smf since smf2.1 beta 3,
mysql fulltext search index on innodb i believe is not supported.

sbulen:
@albertlast and others....   The current upgrader logic assumes that if pg is used, the charset is already UTF8.  Is this a valid assumption?

albertlast:
Well the default database in pg use utf8 and
smf had not the possiblity to change the charset
sooo it's the best assumption.
Any convert stuff doesn't work here.

at the installer we place a check for the utf:
```
 function() {
			$request = pg_query('SHOW SERVER_ENCODING');
			
			list ($charcode) = pg_fetch_row($request);
			
			if ($charcode == 'UTF8')			
				return true;
			else
				return false;

```

sbulen:
Good, I hadn't seen that.  Thanks.

So the only issue here is to get the warning message to display properly for MySQL fulltext users upon UTF8 conversion.

PR #4097, opened 2017-06-04
Upgrader - Display warning when fulltext index dropped

sbulen:
Fixes: Upgrader - fulltext search doesn't work after upgrade #4083

The warning message is now properly displayed when the fulltext index is dropped.

fulltextmsg


Sesquipedalian (changes requested):
For the sake of clarity, it would be good to append "in the admin area after upgrading is complete" to that warning message.

sbulen:
Done.
fulltext2

Sesquipedalian (approved):

PR #4100, opened 2017-06-06
Fixes #4099

### LINKED ISSUE #4099: Installer & Upgrader - array_keys() expects parameter 1 to be array error message
Whenever there's a bunch of changes, I attempt a fresh install & upgrade.   

Just tried a fresh install, all brand new code from git, and got the following message.  Will dig a little deeper, but just getting this out there for folks to be aware of...

This floods the screen during the database population step as well as the admin account creation step:

Warning
: array_keys() expects parameter 1 to be array, null given in
C:\Program Files (x86)\EasyPHP-Devserver-16.1\eds-www\VAN21\Sources\Subs-Db-mysql.php
on line
175

They also appear in the error log as you navigate around the application.

Sesquipedalian:


sbulen (approved):
This all looks good & tests out OK

PR #4105, opened 2017-06-07
Passes $boardOptions in integrate_modify_board hook

Sesquipedalian:
... Because that info is very likely to be needed.


Sesquipedalian:
FYI, I'm passing `$boardOptions` unreferenced because we only need read-access to it in this hook. `integrate_pre_modify_boards` lets a mod edit the content of `$boardOptions`.

colinschoen (approved):
LGTM

PR #4109, opened 2017-06-07
Memberlist Fix PG Customfield sort and  improve MySQL

albertlast:
coalesece was used wrong (used double quotes) (pg fix) and
let the custom fields only join when we need them and than only the who is sorted
not all (mysql improvment).

### DIFF
--- a/Sources/Memberlist.php
+++ b/Sources/Memberlist.php
@@ -327,13 +327,17 @@ function MLAll()
 		$limit = $second_offset - ($memberlist_cache['num_members'] - $_REQUEST['start']) - ($second_offset > $memberlist_cache['num_members'] ? $cache_step_size - ($memberlist_cache['num_members'] % $cache_step_size) : 0);
 	}
 
+	$custom_fields_qry = '';
+	if (!empty($context['custom_profile_fields']['join'][$col]))
+		$custom_fields_qry = $context['custom_profile_fields']['join'][$col];
+
 	// Select the members from the database.
 	$request = $smcFunc['db_query']('', '
 		SELECT mem.id_member
 		FROM {db_prefix}members AS mem' . ($_REQUEST['sort'] === 'is_online' ? '
 			LEFT JOIN {db_prefix}log_online AS lo ON (lo.id_member = mem.id_member)' : '') . ($_REQUEST['sort'] === 'id_group' ? '
 			LEFT JOIN {db_prefix}membergroups AS mg ON (mg.id_group = CASE WHEN mem.id_group = {int:regular_id_group} THEN mem.id_post_group ELSE mem.id_group END)' : '') . '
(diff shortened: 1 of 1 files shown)


Sources/Memberlist.php
```diff
@@ -327,13 +327,17 @@ function MLAll()
...
+		$custom_fields_qry = $context['custom_profile_fields']['join'][$col];
+
 	// Select the members from the database.
 	$request = $smcFunc['db_query']('', '
 		SELECT mem.id_member
 		FROM {db_prefix}members AS mem' . ($_REQUEST['sort'] === 'is_online' ? '
 			LEFT JOIN {db_prefix}log_online AS lo ON (lo.id_member = mem.id_member)' : '') . ($_REQUEST['sort'] === 'id_group' ? '
 			LEFT JOIN {db_prefix}membergroups AS mg ON (mg.id_group = CASE WHEN mem.id_group = {int:regular_id_group} THEN mem.id_post_group ELSE mem.id_group END)' : '') . '
-			' . (!empty($context['custom_profile_fields']['join']) ? implode(' ', $context['custom_profile_fields']['join']) : '') . '
+			' . $custom_fields_qry . '
```
Sesquipedalian, line 340: Does this not still need to be imploded?

albertlast: What need to be imploded?
you only need to join the custom fields when you sort them.

You could keep the imploded when you decide only to support postgres,
because pg ignored the join part when not needed,
but mysql is stupid,
so it do the left join even it's not needed.

Sources/Memberlist.php
```diff
@@ -327,13 +327,17 @@ function MLAll()
 		$limit = $second_offset - ($memberlist_cache['num_members'] - $_REQUEST['start']) - ($second_offset > $memberlist_cache['num_members'] ? $cache_step_size - ($memberlist_cache['num_members'] % $cache_step_size) : 0);
 	}
 
+	$custom_fields_qry = '';
+	if (!empty($context['custom_profile_fields']['join'][$col]))
+		$custom_fields_qry = $context['custom_profile_fields']['join'][$col];
```
Sesquipedalian: I think you might want to use `$_REQUEST['sort']` instead of `$col` here. `$col` will be set to the key of whatever the last element of `$context['columns']` happens to be, and I don't think that is what we want.

Sources/Memberlist.php
Sesquipedalian: Okay, now I understand what you are doing here. Yes, this part is fine, but I think you need to make a change higher up.

Sesquipedalian (changes requested):
I think you meant `$_REQUEST['sort']`, not `$col`

albertlast:
somehow the change of $_REQUEST['sort'] got lost,
any way added this again.

Sesquipedalian (approved):
LGTM

PR #4112, opened 2017-06-08
Fixes #4085

### LINKED ISSUE #4085: Installer - Undefined index & func name must be a string
I can't get the installer to complete at the moment.  White screen, & the message below:
```
Notice: Undefined index: htmlspecialchars in yadayada\VAN21\Sources\Errors.php on line 69

Fatal error: Function name must be a string in yadayada\VAN21\Sources\Errors.php on line 69

```
Brand new code sync'd from Github.  Vanilla install.

sbulen:
This addresses: Installer - Undefined index & func name must be a string #4085

(1) Support for https streams is checked, and if not found, a warning is displayed.
(2) The error functions are properly loaded - for the last 2 steps anyway.  So, AT THE VERY LEAST, any error will appear in the error log and these screwy 'Notice: Undefined index: htmlspecialchars' messages should not be seen after database population.  

See screenshots below.
no_https
https_error

### DIFF
--- a/Themes/default/languages/Install.english.php
+++ b/Themes/default/languages/Install.english.php
@@ -26,6 +26,7 @@ $txt['good_luck'] = 'Good luck!<br>Simple Machines';
 
 $txt['install_welcome'] = 'Welcome';
 $txt['install_welcome_desc'] = 'Welcome to SMF. This script will guide you through the process for installing %1$s. We\'ll gather a few details about your forum over the next few steps, and after a couple of minutes your forum will be ready for use.';
+$txt['install_no_https'] = 'Your environment does not support https streams.  Certain functions, e.g., receiving updates from SimpleMachines.org, will not work.';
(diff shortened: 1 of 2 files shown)


sbulen:
The only thing I'm unsure of here is how much of the old bootstrap code (if any) is redundant with reloadSettings () and should now be removed in the last steps.  

Eg., that custom load_database() unique to the installer?

These changes test clean, I just suspect some cleanup may be appropriate.    

Feedback/direction welcome on this point...

Sesquipedalian:
> The only thing I'm unsure of here is how much of the old bootstrap code (if any) is redundant with reloadSettings () and should now be removed in the last steps.

If the code in this PR is working as expected in its current state, then I'd say we should merge it now and then do any cleanup to remove redundant code in a separate PR.

sbulen:
OK.  All my tests were good.

Sesquipedalian:
In that case, LGTM.

PR #4114, opened 2017-06-09
Upgrader - Confirm https support

sbulen:
Confirm https support in the upgrader.

Similar to PR #4112 , which added the check to the installer.  

upgr_no_https


colinschoen (approved):
LGTM approval

PR #4115, opened 2017-06-10
More method for db_insert returning id

albertlast:
Today i notice that i got a wrong understanding from replace method of mysql,
so i was able to add replace to supported method for returning the ai.
So only insert ingore keep unsupported,
i would try later this day to add also for ingore.

### DIFF
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -767,7 +767,7 @@ function smf_db_error($db_string, $connection = null)
  * @param array $columns An array of the columns we're inserting the data into. Should contain 'column' => 'datatype' pairs
  * @param array $data The data to insert
  * @param array $keys The keys for the table
- * @param int returnmode 0 = nothing(default), 1 = last row id, 2 = all rows id as array; every mode runs only with method = '' or 'insert'
+ * @param int returnmode 0 = nothing(default), 1 = last row id, 2 = all rows id as array; every mode runs only with method != 'ignore'
  * @param object $connection The connection to use (if null, $db_connection is used)
  * @return mixed value of the first key, behavior based on returnmode. null if no data.
  */
@@ -824,7 +824,7 @@ function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $ret
 		$connection
 	);
 
-	if(!empty($keys) && (count($keys) > 0) && ($method === '' || $method === 'insert') && $returnmode > 0)
+	if(!empty($keys) && (count($keys) > 0) && $method !== 'ignore' && $returnmode > 0)
 	{
 		if ($returnmode == 1)
 			$return_var = smf_db_insert_id($table, $keys[0]) + count($insertRows) - 1;
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -767,7 +767,7 @@ function smf_db_error($db_string, $connection = null)
  * @param array $columns An array of the columns we're inserting the data into. Should contain 'column' => 'datatype' pairs
  * @param array $data The data to insert
  * @param array $keys The keys for the table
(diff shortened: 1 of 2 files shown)


albertlast:
So this pr is done,
would be good if someone is able to test specialy the ingore part,
is very complex the solution...

Sources/Subs-Db-mysql.php
```diff
@@ -776,13 +776,24 @@ function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $ret
 	global $smcFunc, $db_connection, $db_prefix;
 
 	$connection = $connection === null ? $db_connection : $connection;
+	
+	$return_var;
```
jdarwood007: Typo?

Sources/Subs-Db-mysql.php
```diff
@@ -811,24 +822,82 @@ function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $ret
...
+					$where_string += key($indexed_columns[$x]) . ' = '. $insertRows[$i][$x];
+					if (($x + 1) < $count2)
+						$where_string += ' AND ';
+				}
+
+				$request = $smcFunc['db_query']('','
+					SELECT `'. $keys[0] . '` FROM ' . $table .'
+					WHERE ' . $where_string . ' LIMIT 1',
+					array()
+				);
```
jdarwood007, line 879: Only issue I have with this, is we are assuming that $keys[0] contains the PK.  I should be true for 99% of cases, but I wonder if that is still true.

I believe the entire need for us specifying the PK in the builder was SQLite had issues with inserts and had to rebuild indexes or something of the sort.  I don't think its even relevant for just MySQL and PostgreSQL.

Sources/Subs-Db-mysql.php
albertlast: i need the var across the methode,
to get the right scrope i defined this var on start.

Sources/Subs-Db-mysql.php
albertlast: The pk field is essential for pg and
mysql only allow that one field in a table is a ai field,
so i had to assume that the first pk field is the ai field.

similiar logic you find on pg side: https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs-Db-postgresql.php#L769
like i said this is for the limitation on mysql side.

Sources/Subs-Db-mysql.php
jdarwood007: Shouldn't you specify a = null then?  Just seems weird to see it like this.

Sources/Subs-Db-mysql.php
jdarwood007: Well developers should be following the proper logic and making the AI field the first in the list for the keys.  Doesn't mean it will happen 100%, but it could.  I guess we call those bugs in that code.

Sources/Subs-Db-mysql.php
albertlast: i hope empty() keeps triggered

jdarwood007: https://secure.php.net/empty

Yes, null is considered empty.  But you can also test for null with is_null or === null.  Which would make it different than 0 or false.

Sources/Subs-Db-mysql.php
albertlast: so it's fine how it is?

Sesquipedalian: Approval, @jdarwood007?

jdarwood007: I've voiced my concerns, if no others agree, then merge it.  Worse case is it becomes a coding standard for the keys to contain the pki as the first item.  We could already use the passed pki info to figure it out ourselves.

Sesquipedalian: I don't have an opinion either way. SQL is not my strong suit, so I'm just waiting to see if you two can reach a conclusion or if another dev decides to weigh in.

PR #4117, opened 2017-06-11
added random_bytes

albertlast:
to protect that we got the depracted message of this function: http://php.net/manual/de/function.mcrypt-create-iv.php
i added the new one random_bytes() function on the the place before we call mcrypt-create-iv().

### DIFF
--- a/Sources/Subs-Password.php
+++ b/Sources/Subs-Password.php
@@ -103,7 +103,13 @@ namespace {
 			} else {
 				$buffer = '';
 				$buffer_valid = false;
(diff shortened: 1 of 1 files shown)


Sesquipedalian (approved):
LGTM

jdarwood007:
Would changing this require any needed password upgrades?

Sesquipedalian:
No, it's just a method for generating the password salt.

jdarwood007:
Thanks, LGTM.

PR #4120, opened 2017-06-12
Newsletter fixes

### LINKED ISSUE #4116: Newsteller variable issue
Heyo, 

when sending an email with the newsteller, the variable {$member.link} outputs {$member.name} instead of the link. Just sent a bunch of emberrasing emails haha

Cheers, kat

Oldiesmann:
This addresses a couple of issues:
- Always include a link for {$member.link} even when not sending via HTML (fixes #4116)
- Flawed logic in the code caused the newsletter to not get sent in most cases

### DIFF
--- a/Sources/ManageNews.php
+++ b/Sources/ManageNews.php
@@ -1008,7 +1008,7 @@ function SendMailing($clean_only = false)
 			$message = str_replace($from_member,
 				array(
 					$row['email_address'],
-					!empty($_POST['send_html']) ? '<a href="' . $scripturl . '?action=profile;u=' . $row['id_member'] . '">' . $cleanMemberName . '</a>' : ($context['send_pm'] ? '[url=' . $scripturl . '?action=profile;u=' . $row['id_member'] . ']' . $cleanMemberName . '[/url]' : $cleanMemberName),
+					!empty($_POST['send_html']) ? '<a href="' . $scripturl . '?action=profile;u=' . $row['id_member'] . '">' . $cleanMemberName . '</a>' : ($context['send_pm'] ? '[url=' . $scripturl . '?action=profile;u=' . $row['id_member'] . ']' . $cleanMemberName . '[/url]' : $scripturl . '?action=profile;u=' . $row['id_member']),
 					$row['id_member'],
 					$cleanMemberName,
 				), $_POST['message']);
--- a/Sources/ManageNews.php
+++ b/Sources/ManageNews.php
@@ -985,7 +985,7 @@ function SendMailing($clean_only = false)
 		foreach ($rows as $row)
 		{
 			// Force them to have it?
-			if (empty($context['email_force']) || empty($prefs[$row['id_member']]['announcements']))
+			if (empty($context['email_force']) && empty($prefs[$row['id_member']]['announcements']))
 				continue;
 
 			// What groups are we looking at here?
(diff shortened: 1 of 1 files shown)


Sources/ManageNews.php
```diff
@@ -1008,7 +1008,7 @@ function SendMailing($clean_only = false)
 			$message = str_replace($from_member,
 				array(
 					$row['email_address'],
-					!empty($_POST['send_html']) ? '<a href="' . $scripturl . '?action=profile;u=' . $row['id_member'] . '">' . $cleanMemberName . '</a>' : ($context['send_pm'] ? '[url=' . $scripturl . '?action=profile;u=' . $row['id_member'] . ']' . $cleanMemberName . '[/url]' : $cleanMemberName),
+					!empty($_POST['send_html']) ? '<a href="' . $scripturl . '?action=profile;u=' . $row['id_member'] . '">' . $cleanMemberName . '</a>' : ($context['send_pm'] ? '[url=' . $scripturl . '?action=profile;u=' . $row['id_member'] . ']' . $cleanMemberName . '[/url]' : $scripturl . '?action=profile;u=' . $row['id_member']),
```
Sesquipedalian, line 1012: How about `$cleanMemberName . ' (' . $scripturl . '?action=profile;u=' . $row['id_member'] . ')'`?

Oldiesmann: We already have {$member.name} for the member name though, so I'm not sure why we need to include it with the member link when we're not sending HTML emails

Sesquipedalian: As far as I can tell, the idea of `{$member.link}` seems to be to give a clickable version of the member's name that links to the member's profile. I figured that writing the name followed by the URL to the profile was the best way to approximate that in a plain text email.

Maybe we should have both a `{$member.link}` variable and a `{$member.url}` variable? Then `{$member.link}` would give the clickable name in HTML emails and PMs or the nearest approximation we can provide in plain text emails. Meanwhile, `{$member.url}` would provide just the raw `$scripturl . '?action=profile;u=' . $row['id_member']` string so that admins could explicitly include that even in HTML emails whenever they chose.

Sesquipedalian:
I suppose we should probably also give the same treatment to `$cleanLatestMember` a bit [earlier](https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/ManageNews.php#L864) in the function.

Sources/ManageNews.php
Oldiesmann: The helptext for `{$member.link}` simply says "The current member's link". Most people would expect that to be a link to the current member's profile, so I'm not sure adding a `{$member.url}` would be very useful to people. On a related note we're doing the same thing with `{$latest_member.link}`.

Oldiesmann:
Found the issue with newsletters not being sent for me (at least one of them - my non-admin test account is still not getting them) so I changed the name and comments for this PR...

Oldiesmann:
Still looking into the other newsletter issue here... If I change my regular test account to a global mod account it gets found by the query to get members to send emails to. I'll have to play with it more tomorrow and see what I can come up with.

Sources/ManageNews.php
Sesquipedalian: Fair enough. :)

Issue #4121, opened 2017-06-12
Polls with multiple choices


kat-kta:
Heyo

Just wanna point out that polls where you can choose more than one answer don't fill up the statsbar accordingly. If there are 13 possible choices and 13 people in total voted, 10 out of them for choice # 1 (but also for any ammount of other choices cause it's multiple choice), its stats bar only gets a width of 23.3% (even though a majority - 10/13 people - voted for that choice).

I guess that's because the system takes the total no. of **votes** to divide into percantage and not the no. of **people who voted**. 

Cheers, kat

PR #4122, opened 2017-06-13
Fix multi choice fixing #4121

albertlast:
Placed a switch for the case we got a multi choice vote
#4121


Sesquipedalian (approved):
Nicely done

PR #4128, opened 2017-06-14
Updates version to Beta 4

Sesquipedalian:
This may in fact become RC1, but we won't label it as such until it is ready to release.

### DIFF
--- a/SSI.php
+++ b/SSI.php
@@ -8,7 +8,7 @@
  * @copyright 2017 Simple Machines and individual contributors
  * @license http://www.simplemachines.org/about/smf/license.php BSD
  *
- * @version 2.1 Beta 3
+ * @version 2.1 Beta 4
  */
 
 // Don't do anything if SMF is already loaded.
--- a/Sources/Admin.php
+++ b/Sources/Admin.php
@@ -10,7 +10,7 @@
  * @copyright 2017 Simple Machines and individual contributors
  * @license http://www.simplemachines.org/about/smf/license.php BSD
  *
- * @version 2.1 Beta 3
+ * @version 2.1 Beta 4
  */
 
 if (!defined('SMF'))
--- a/Sources/Attachments.php
+++ b/Sources/Attachments.php
@@ -10,7 +10,7 @@
  * @copyright 2017 Simple Machines and individual contributors
  * @license http://www.simplemachines.org/about/smf/license.php BSD
  *
- * @version 2.1 Beta 3
+ * @version 2.1 Beta 4
  */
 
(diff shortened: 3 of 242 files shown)


Sesquipedalian:
Nothing really to see, @colinschoen. Just want to make sure you sign off before it is merged.

XinYenFon:
IMO, you guys should not be labeling versions which never gonna see a daylight. This probably create confusion in community and lead questions related "where is beta4", "what happened to beta4" etc...

Sesquipedalian:
It hasn't been decided yet whether it will be a release candidate, but we need to update it to something higher than Beta 3 to avoid confusion in the meantime.

XinYenFon:
I think its best to keep it as beta3 or Nightly, in an ideal development process version in the headers should be either Nightly-<latest_commit_number> or just Nightly. Versioning should be changed upon the release candidate (to avoid confusion I'm not meaning RC of final product) commit of that version.

IMO, its best time to bring the script written by @jdarwood007 into the action so nightly releases may be distinguished and distributed far more easily. With that possible change not only users can find nightly builds easier they also have static data related their nightly build so developers and support members can identify the problems (or solutions).

Sesquipedalian:
If there is a better solution, I am open to it. We probably need to hear more from @jdarwood007 and @colinschoen at this point, since a bump to Beta 4 is what was decided in discussion on the forum.

XinYenFon:
oh I see 😆 sorry some old habits still getting used to not seeing the team boards.

Sesquipedalian:
Can you still not see the Development Contributors board? That's where we talked about this one.

XinYenFon:
Nope.

colinschoen:
Working on fixing that still.

albertlast:
ther is now new feature planed, i see no reason to call this a beta.

Sesquipedalian:
Saying that there are no new features *planned* is not the same as saying that we are declaring that no new features will be *accepted*, so I am not certain that we are officially in RC stage at this point.

Sesquipedalian:
Also, I would suggest that even when we declare that we are officially in the RC stage, we allow minor additions like new integration hooks to be added right up until we officially release RC 1.

albertlast:
Sure that are not features in my eyes so ther is no limitation on extension like this.

Sesquipedalian:
BTW @jdarwood007 and/or @live627: It looks like Travis is broken again. :(

colinschoen (approved):
LGTM

colinschoen:
@Antes You should have access to that board now.

PR #4133, opened 2017-06-16
Unbind the click event each time the function is called

### LINKED ISSUE #4084: Issue with "Quote selected text"
I open this issue because i didn't found any on this.
Php version 7.0
SMF beta 3

When you go to a topic, you select a text on a post, you decide that this text is not what you want, you select an other part of the text on the same post. This time it's the right selection. Then you click on the "Quote selected text" and the problem come up. It will insert into the quick reply the first and the second selected texts and not only the last.

Hope it's understandable

BloberSh:
Fix #4084

It also remove the ability to select nonadjacent parts of text on firefox.
Chrome, IE and Edge don't support this so no problems.

Signed-off-by: Blober <contact@blober.net>

### DIFF
--- a/Themes/default/scripts/quotedText.js
+++ b/Themes/default/scripts/quotedText.js
@@ -118,6 +118,9 @@ $(function() {
 			msgID : $(this).data('msgid'),
 		};
 
+		// Revome any 'click' event to the button.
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Works nicely. Thanks for your submission!

PR #4134, opened 2017-06-16
implement seek-method for display of topic

albertlast:
Well it database world you try to evade paging
so somthing like that: 
select * from messages LIMIT 15, 3000
is curse.

one of the common case where can evade this is,
when you jump from page to page,
here you can use the seek method which garantee the same run time on every page no different if you call the first page or the 300th.

For this stuff i had to two things:
build a base (subs.php and querystring.php) and
then implement this for board display topics.

example when you normale call a topic thr url look like that:
index.php?topic=123364.30
ther is no change but when you now try to access the next page:
index.php?topic=123364.45.M993457
M means next page/upper limit/max
993457 the id of the message

similiar it's look when you try the page before:
http://localhost/SMF2.1/index.php?topic=123364.15.L993457
L means before page/lower limit/min
993457 the id of the message

The effect old:
         FROM smf_messages
         WHERE id_topic = 123364
         ORDER BY id_msg DESC
      LIMIT 15 OFFSET 2905
   in ...\Sources\Display.php line 866, which took 0.00132585 

         FROM smf_messages
         WHERE id_topic = 123364
         AND id_msg < 996414
         ORDER BY id_msg DESC
         LIMIT 15
   in ...\Sources\Display.php line 899, which took 0.0003171

### DIFF
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -441,8 +441,8 @@ function Display()
 	elseif (isset($_REQUEST['all']))
 		$_REQUEST['start'] = -1;
 
-	// Construct the page index, allowing for the .START method...
-	$context['page_index'] = constructPageIndex($scripturl . '?topic=' . $topic . '.%1$d', $_REQUEST['start'], $context['total_visible_posts'], $context['messages_per_page'], true);
+	// Construct allowing for the .START method...
+	//$context['page_index'] = constructPageIndex($scripturl . '?topic=' . $topic . '.%1$d', $_REQUEST['start'], $context['total_visible_posts'], $context['messages_per_page'], true);
 	$context['start'] = $_REQUEST['start'];
 
 	// This is information about which page is current, and which page we're on - in case you don't like the constructed page index. (again, wireles..)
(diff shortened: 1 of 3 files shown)


Sources/Subs.php
```diff
@@ -543,16 +543,22 @@ function updateSettings($changeArray, $update = false)
...
+function constructPageIndex($base_url, &$start, $max_value, $num_per_page, $flexible_start = false, $show_prevnext = true, $options = null)
 {
 	global $modSettings, $context, $smcFunc, $settings, $txt;
 
 	// Save whether $start was less than 0 or not.
 	$start = (int) $start;
 	$start_invalid = $start < 0;
+	
+	if (!empty($options['low_id']))
+		$low_id = $options['low_id'];
```
NanoSector, line 559: might want to typeforce this with (int). $max_id too

Sources/Display.php
```diff
@@ -841,37 +823,84 @@ function Display()
...
+				'start' => $start,
+				'max' => $limit,
+			)
+		);
+	}
+	else //next or before page
+	{
+		$firstIndex = 0;
+		
+		if ($start_char === 'M')
```
NanoSector, line 874: can be simplified;
```php
$ascending = $start_char === 'M' ? true : false;
$page_operator = $start_char === 'M' ? '>' : '<';
```
slightly cleaner. also not sure why you're using === here vs ==

NanoSector: in fact $ascending can be simplified as 
```php
$ascending = $start_char === 'M';
```
no need for the explicit true and false there.

Sources/Subs.php
albertlast: Reason why i said to me that i don't need this,
because i say in query side that is had to be a int: https://github.com/SimpleMachines/SMF2.1/pull/4134/files/e2b10fd24660a764ae3f5566800eb29b49348231#diff-eeb7dcd590ea5ad96bd502e7c294017bR889
but maybe not and bad idea to force here too

PR #4137, opened 2017-06-16
Trim

Sesquipedalian:
1. Fixes a bug with trimming breaks after block_level BBC.
Specifically, if a block level BBC was nested inside another block level, and the inner block level was NOT followed by a `<br>`, we might still trim off four characters after the inner BBC's closing tag.

2. ~Adds a 'trim_breaks' BBC param to help with layout for the float BBC
When 'trim' is set to 'outside' or 'both', 'trim_breaks' causes the whitespace trimming to always remove as many `<br>` elements as it can. This is the same thing that we already do for nested BBC like [li] and [tr] inside lists and tables, but now there is a way to specify it for other BBC that need it, such as the [float] BBC.~ Fixes a discrepancy between the editor's CSS and the standard display CSS.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -1040,13 +1040,9 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 			  not block level, and will not be implicitly closed as easily.
 			  One break following a block level tag may also be removed.
 
-			trim: if set to 'inside', whitespace after the begin tag will be
+			trim: if set, and 'inside' whitespace after the begin tag will be
 			  removed.  If set to 'outside', whitespace after the end tag will
-			  meet the same fate. If set to 'both', both will be done.
-
-			trim_breaks: if set to true, whitespace trimming after the end tag
-			  will trim as many <br> elements as it can. This is mostly useful
-			  for block_level tags.
+			  meet the same fate.
 
 			validate: except when type is missing or 'closed', a callback to
 			  validate the data as $data.  Depending on the tag's type, $data
@@ -1293,7 +1289,6 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 					$data = $class . $css;
 				},
 				'trim' => 'outside',
-				'trim_breaks' => true,
 				'block_level' => true,
 			),
 			array(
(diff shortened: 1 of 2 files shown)


colinschoen (approved):
LGTM

jdarwood007:
Do you have a before and after example of this?

Sesquipedalian:
Of the bug, or of how 'trim_breaks' works?

Sesquipedalian:
Using this as the post text in all cases:
```
[float=right max=33%][quote]Lorem ipsum dolor sit amet, consectetuer adipiscing elit.[/quote][/float]

Nullam sagittis nunc quis massa. Curabitur a pede vel risus mollis dignissim. Aenean nec massa. Vivamus at tellus. Sed ac nunc.
```

1. Result without this PR:
screen shot 2017-06-16 at 4 38 54 pm
Notice how the first four characters of the closing `[/float]` were cut away. That's because there was no line break after the closing `[/quote]`

2. Result with 005fc92:
screen shot 2017-06-16 at 4 42 26 pm
Bug is fixed, but we still have leftover line breaks after the float that make the layout look odd. One could argue that the user should just remove all the line breaks in their entered text, but one could also argue the same about how we handle trailing line breaks inside lists.

3. Result with 90f337e:
screen shot 2017-06-16 at 4 52 37 pm
Layout fixed.

Sesquipedalian:
... Hm. On second thought, maybe 90f337e isn't such a good idea after all. If the user really wants to add a bunch of line breaks after a float, we should probably let them.

jdarwood007:
Yea the bug as well. Just trying to understand it from a logic of the output.

Sesquipedalian:
So, does that mean it looks good to you at this point, @jdarwood007?

jdarwood007:
Yea, so with that last commit reverting one of your examples, it shows up like the second one?

Sesquipedalian:
> Yea, so with that last commit reverting one of your examples, it shows up like the second one?

Yes. But simply changing the raw text to the following gets rid of the extra line break before the "Nullam..." paragraph.
```
[float=right max=33%][quote]Lorem ipsum dolor sit amet, consectetuer adipiscing elit.[/quote][/float]
Nullam sagittis nunc quis massa. Curabitur a pede vel risus mollis dignissim. Aenean nec massa. Vivamus at tellus. Sed ac nunc.
```

Issue #4136, opened 2017-06-16
2.1 upgrade issue Incorrect date value: '0000-00-00' for column 'birthdate'


VBGAMER45:
Occurs on MySQL 5.7
 it seems like MySQL 5.7 doesn't like 0000-00-00 as a date. You'd have to use either an arbitrary date like 1970-01-01 or null.

Report link
https://www.simplemachines.org/community/index.php?topic=554635.0

albertlast:
This error should be not mysql 5.7 related because we force on every mysql,
that NO_ZERO_DATE is enabled https://dev.mysql.com/doc/refman/5.5/en/sql-mode.html#sqlmode_no_zero_date
https://github.com/SimpleMachines/SMF2.1/pull/4068
merged after beta 3 release.

Birthdate is a not null field (stupid smf logic) but the default value 0001-01-01 is a valid value.

FDinolfo:
I can only go by what I'm seeing here. Let me know of any other additional info or log entries you want and I'll bring them over here.

albertlast:
Oldiesmann will look into (he will create the pr),
i already did some changes
maybe he find more stuff and do finish the upgrade part: https://github.com/albertlast/SMF2.1/commit/dc2ce578de957da45d3a06b29bc8a96f56b2a735

PR #4139, opened 2017-06-17
Date fixes

Oldiesmann:
Per the MySQL manual, MySQL 5.7 only "supports" date values as low as 1001-01-01, though other values may work as well. This PR updates SMF's usage of 0001-01-01 to compensate for this.

### DIFF
--- a/other/upgrade_2-1_postgresql.sql
+++ b/other/upgrade_2-1_postgresql.sql
@@ -3,7 +3,7 @@
 /******************************************************************************/
 --- Fixing dates...
 /******************************************************************************/
---# Updating old values
+---# Updating old values
 UPDATE {$db_prefix}calendar
 SET start_date = '1001-01-01'
 WHERE start_date < '1001-01-01';
@@ -23,13 +23,15 @@ WHERE stat_date < '1001-01-01';
 UPDATE {$db_prefix}members
 SET birthdate = '1001-01-01'
 WHERE birthdate < '1001-01-01';
+---#
 
---# Changing default values
-ALTER TABLE {$db_prefix}calendar CHANGE start_date start_date date NOT NULL DEFAULT '1001-01-01';
-ALTER TABLE {$db_prefix}calendar CHANGE end_date end_date date NOT NULL DEFAULT '1001-01-01';
-ALTER TABLE {$db_prefix}calendar_holidays CHANGE event_date event_date date NOT NULL DEFAULT '1001-01-01';
-ALTER TABLE {$db_prefix}log_spider_stats CHANGE stat_date stat_date date NOT NULL DEFAULT '1001-01-01';
-ALTER TABLE {$db_prefix}members CHANGE birthdate birthdate date NOT NULL DEFAULT '1001-01-01';
+---# Changing default values
+ALTER TABLE {$db_prefix}calendar ALTER COLUMN start_date SET DEFAULT '1001-01-01'::date;
+ALTER TABLE {$db_prefix}calendar ALTER COLUMN end_date SET DEFAULT '1001-01-01'::date;
+ALTER TABLE {$db_prefix}calendar_holidays ALTER COLUMN event_date SET DEFAULT '1001-01-01'::date;
+ALTER TABLE {$db_prefix}log_spider_stats ALTER COLUMN state_date SET DEFAULT '1001-01-01'::date;
+ALTER TABLE {$db_prefix}members ALTER COLUMN birthdate SET DEFAULT '1001-01-01'::date;
+---#
 
 /******************************************************************************/
 --- Adding new settings...
--- a/other/upgrade_2-1_postgresql.sql
+++ b/other/upgrade_2-1_postgresql.sql
@@ -5,24 +5,24 @@
 /******************************************************************************/
 ---# Updating old values
 UPDATE {$db_prefix}calendar
-SET start_date = '1001-01-01'
-WHERE start_date < '1001-01-01';
+SET start_date = concat_ws('-', CASE WHEN EXTRACT(YEAR FROM start_date) < 1001 THEN 1001 END, EXTRACT(MONTH FROM start_date), EXTRACT(DAY FROM start_date))::date
+WHERE EXTRACT(YEAR FROM start_date) < 1001;
 
 UPDATE {$db_prefix}calendar
-SET end_date = '1001-01-01'
-WHERE end_date < '1001-01-01';
+SET end_date = concat_ws('-', CASE WHEN EXTRACT(YEAR FROM end_date) < 1001 THEN 1001 END, EXTRACT(MONTH FROM end_date), EXTRACT(DAY FROM end_date))::date
+WHERE EXTRACT(YEAR FROM end_date) < 1001;
 
 UPDATE {$db_prefix}calendar_holidays
-SET event_date = '1001-01-01'
-WHERE event_date < '1001-01-01';
+SET event_date = concat_ws('-', CASE WHEN EXTRACT(YEAR FROM event_date) < 1001 THEN 1001 END, EXTRACT(MONTH FROM event_date), EXTRACT(DAY FROM event_date))::date
+WHERE EXTRACT(YEAR FROM event_date) < 1001;
 
 UPDATE {$db_prefix}log_spider_stats
-SET stat_date = '1001-01-01'
-WHERE stat_date < '1001-01-01';
+SET stat_date = concat_ws('-', CASE WHEN EXTRACT(YEAR FROM stat_date) < 1001 THEN 1001 END, EXTRACT(MONTH FROM stat_date), EXTRACT(DAY FROM stat_date))::date
+WHERE EXTRACT(YEAR FROM stat_date) < 1001;
 
 UPDATE {$db_prefix}members
-SET birthdate = '1001-01-01'
-WHERE birthdate < '1001-01-01';
+SET birthdate = concat_ws('-', CASE WHEN EXTRACT(YEAR FROM start_date) < 1001 THEN 1001 END, CASE WHEN EXTRACT(MONTH FROM birthdate) < 1 THEN 1 ELSE EXTRACT(MONTH FROM birthdate) END, CASE WHEN EXTRACT(DAY FROM birthdate) < 1 THEN 1 ELSE EXTRACT(DAY FROM birthdate) END)::date
+WHERE birthdate EXTRACT(YEAR FROM birthdate) < 1001;
 ---#
 
 ---# Changing default values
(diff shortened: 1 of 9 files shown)


other/upgrade_2-1_postgresql.sql
```diff
@@ -1,6 +1,37 @@
...
+UPDATE {$db_prefix}log_spider_stats
+SET stat_date = '1001-01-01'
+WHERE stat_date < '1001-01-01';
+
+UPDATE {$db_prefix}members
+SET birthdate = '1001-01-01'
+WHERE birthdate < '1001-01-01';
+
+--# Changing default values
+ALTER TABLE {$db_prefix}calendar CHANGE start_date start_date date NOT NULL DEFAULT '1001-01-01';
```
albertlast: wrong syntax for pg:
ALTER TABLE {$db_prefix}calendar ALTER COLUMN start_date SET DEFAULT '100-01-01'::date;

albertlast (approved):

Oldiesmann:
Still looking into this further. I just realized that we can't just set the date to 1001-01-01 where the date is earlier than that because it will screw up a bunch of stuff.

albertlast:
Because the Birthday issue,
I was believing that Need to be fix on early Stage.


Von: Michael Eshom <notifications@github.com>
Gesendet: 19.06.2017 6:55 vorm.
An: SimpleMachines/SMF2.1
Cc: albertlast; Comment
Betreff: Re: [SimpleMachines/SMF2.1] Date fixes (#4139)


Still looking into this further. I just realized that we can't just set the date to 1001-01-01 where the date is earlier than that because it will screw up a bunch of stuff.

Oldiesmann:
Think I got everything this time. Note that I haven't tested the PostgreSQL part of the upgrade, but I believe that will work, unless there's a simpler way of doing it.

Sesquipedalian:
Two questions:

1. Why `1001-01-01` rather than `1000-01-01`? According to [the manual](https://dev.mysql.com/doc/refman/5.7/en/datetime.html), that is the bottom end of the range of supported dates.
2. What is the reason for not simply using `0000-00-00`? It falls outside the supported range, true, but the manual states that this is the "zero" value used whenever an invalid date is given. Since the code affected by this PR deals with the situation where the user has not entered a valid birth date, I don't understand at this point why we aren't just going with MySQL's built-in way of dealing with that rather than creating our own workaround.

Oldiesmann:
1. I must have misread the manual. I thought it said 1001-01-01 was the bottom end.
2. Didn't see that part either.

I'll look into addressing those situations in the coming days since it's 3:12am here

Oldiesmann:
@Sesquipedalian Now that I'm not half asleep I can explain the 0000-00-00 thing. This whole PR started as a result of #4136. When MySQL is in strict mode, 0000-00-00 isn't allowed (see https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sql-mode-strict and https://stackoverflow.com/questions/36374335/error-in-mysql-when-setting-default-value-for-date-or-datetime#36374690) unless you use INSERT IGNORE or UPDATE IGNORE. Since most users have no control over this, it's best to just skip that value altogether.

Sesquipedalian:
Thanks for the explanation, @Oldiesmann. :)

other/upgrade_2-1_postgresql.sql
```diff
@@ -1,6 +1,39 @@
 /* ATTENTION: You don't need to run or use this file! The upgrade.php script does everything for you! */
 
 /******************************************************************************/
+--- Fixing dates...
```
Sesquipedalian, line 4: Looks like upgrade_2-1_postgresql.sql still needs the `1001` → `1000` change.

Sesquipedalian (changes requested):

albertlast:
Why we take the day and month from dates wher they wrong?
For exp. SET start_date = DATE(CONCAT(1000, '-', MONTH(start_date), '-', DAY(start_date)))

Oldiesmann:
@albertlast Currently for holidays that repeat on the same date every year (Christmas, etc.), we use 0004 for the year. In those cases we just want to change the year to 1000 without changing the day or month. In the past we've also used 0004 at times for birthdays when the user wants to enter their birthday without having the forum display their age. It should automatically change the day and month to 1 when those are incorrect as well unless I missed something.

other/upgrade_2-1_postgresql.sql
```diff
@@ -5,32 +5,32 @@
...
 UPDATE {$db_prefix}log_spider_stats
-SET stat_date = concat_ws('-', CASE WHEN EXTRACT(YEAR FROM stat_date) < 1001 THEN 1001 END, EXTRACT(MONTH FROM stat_date), EXTRACT(DAY FROM stat_date))::date
-WHERE EXTRACT(YEAR FROM stat_date) < 1001;
+SET stat_date = concat_ws('-', CASE WHEN EXTRACT(YEAR FROM stat_date) < 1000 THEN 1000 END, EXTRACT(MONTH FROM stat_date), EXTRACT(DAY FROM stat_date))::date
+WHERE EXTRACT(YEAR FROM stat_date) < 1000;
 
 UPDATE {$db_prefix}members
-SET birthdate = concat_ws('-', CASE WHEN EXTRACT(YEAR FROM start_date) < 1001 THEN 1001 END, CASE WHEN EXTRACT(MONTH FROM birthdate) < 1 THEN 1 ELSE EXTRACT(MONTH FROM birthdate) END, CASE WHEN EXTRACT(DAY FROM birthdate) < 1 THEN 1 ELSE EXTRACT(DAY FROM birthdate) END)::date
-WHERE birthdate EXTRACT(YEAR FROM birthdate) < 1001;
+SET birthdate = concat_ws('-', CASE WHEN EXTRACT(YEAR FROM start_date) < 1000 THEN 1000 END, CASE WHEN EXTRACT(MONTH FROM birthdate) < 1 THEN 1 ELSE EXTRACT(MONTH FROM birthdate) END, CASE WHEN EXTRACT(DAY FROM birthdate) < 1 THEN 1 ELSE EXTRACT(DAY FROM birthdate) END)::date
```
albertlast: start_date = birthdate

other/upgrade_2-1_postgresql.sql
```diff
@@ -5,32 +5,32 @@
...
+WHERE birthdate EXTRACT(YEAR FROM birthdate) < 1000;
```
albertlast: invalid sql syntax -> remove birthdate (the left one)

albertlast (changes requested):

albertlast:
Well then would exists a possible better solution:
instead of 
`SET birthdate = DATE(CONCAT(IF(YEAR(birthdate) < 1000, 1000, YEAR(birthdate)), '-', IF(MONTH(birthdate) < 1, 1, MONTH(birthdate)), '-', IF(DAY(birthdate) < 1, 1, DAY(birthdate)))) `
replace with
`SET birthdate = birthdate + interval '1000' year`

motivation is that you evade issues with date like 29.02.2017
exp:
29.02.2016 + interval '1' year  -> 28.02.2017 (good)
'29.02.2017'::date -> error date out of range (this can happen by your string operation)

Sesquipedalian (approved):

Sesquipedalian:
I've approved for my requested changes. I'll leave it to you two to reach a solution regarding @albertlast's requested changes. :)

Oldiesmann:
@albertlast I think you're misunderstanding the code, as it can't end up with an invalid date

We're never adding 1 to anything. We change the year to 1000 if it's less than 1000, change the month to 1 if it's 0 and change the day to 1 if it's 0. Otherwise we use the existing values (if year is at least 1000, we leave it alone; if the month isn't 0, we leave it alone; if the day isn't 0, we leave it alone). In addition, 1000 was a leap year so 1000-02-29 is still a valid date.

In addition, adding 1000 will lead to unexpected results if the year is anything other than 0. In some cases we've used 0004 as the year for birthdays when the user doesn't want their age revealed.

IF is just a shorthand version of the standard CASE WHEN:
`IF(YEAR(birthdate) < 1000, 1000)` is the same as `CASE WHEN YEAR(birthdate) < 1000 THEN 1000 ELSE YEAR(birthdate) END`

albertlast:
Why you believe is 1000 a leap year?
is divided by 100 -> no leap year
when i try to insert in mysql '1000-02-29' i got an error
same for pg

sbulen:
It's a little more convoluted than that...  From Wikipedia: "There is a leap year every year whose number is perfectly divisible by four - except for years which are both divisible by 100 and not divisible by 400. The second part of the rule effects century years. For example; the century years 1600 and 2000 are leap years, but the century years 1700, 1800, and 1900 are not."   

How perfect do we need to be?  Are we scheduling birthdays for 2100?    ;^)

Oldiesmann:
Apparently 1000 actually was a leap year according to the Julian calendar which was in use at the time. The simplest solution here would be to use 1004 as the minimum year instead of 1000 (and for consistency use it across SMF).

Sesquipedalian:
> Apparently 1000 actually was a leap year according to the Julian calendar which was in use at the time. The simplest solution here would be to use 1004 as the minimum year instead of 1000 (and for consistency use it across SMF).

What problem is that a solution for? A.D. 1000 was a leap year according to both the Julian and Gregorian calendars, so why not use it?

albertlast:
Well the issue is that is no leap year like i saied before,
to prove this point i saied the rule (divided by 100) and show two program(mysql and pg)
wehre the result show that 29.2.1000 didn't exists.

Sesquipedalian:
Derp. Never mind. I made a calculation error in my head. :)

other/upgrade_2-1_postgresql.sql
```diff
@@ -1,6 +1,39 @@
 /* ATTENTION: You don't need to run or use this file! The upgrade.php script does everything for you! */
 
 /******************************************************************************/
+--- Fixing dates...
+/******************************************************************************/
+---# Updating old values
+UPDATE {$db_prefix}calendar
+SET start_date = concat_ws('-', CASE WHEN EXTRACT(YEAR FROM start_date) < 1000 THEN 1000 END, EXTRACT(MONTH FROM start_date), EXTRACT(DAY FROM start_date))::date
+WHERE EXTRACT(YEAR FROM start_date) < 1000;
+
```
albertlast: 1000 -> 1004
next queries too

jdarwood007:
While the divide by 100 rule is correct for the Gregorian calendar, the Gregorian calendar wasn't introduced until 1582.  So technically the year 1000 is only on the Julian calendar.  So by all accounts, 1000 should be  a leap year, with 1700 being the first year to apply the divide by 100, as 1600 passed the divide by 400 rule.

I suspect most systems go the easy route and just do the math and don't take into account anything prior to 1600 using the old system.

PR #4143, opened 2017-06-20
better join for display query

albertlast:
Take the id_member_started from topic instead of the id_messages -> easier join plan.
less time for queryplaner to find a good solution.
result:
in ...\Sources\Display.php line 171, which took 0.00295496 seconds at 0.04309702 into request.
in ...\Sources\Display.php line 171, which took 0.00209904 seconds at 0.04312682 into request.


colinschoen (approved):
LGTM approval

PR #4144, opened 2017-06-21
Fixes #1896

### LINKED ISSUE #1896: number of reported members does not show in moderate menu
if there is more then 1 reported members there should be a 1 in the moderate menu, currently there is nothing.

sbulen:
Fixed reported member count in mod menu.  

Main problem was that it needed to be more specific when checking the session variables, i.e., checking $_SESSION['rc']['member_reports'] instead of just $_SESSION['rc'].  Prior logic always ensured ['rc'] was set, so it was never empty...  Since it never thought it was empty, it was never recalculated via a call to recountOpenReports()...  

While in there, noted that the context variables weren't loaded when recountOpenReports() was called.  Thus it took multiple executions of this logic to get the menu entries right.  Now it can get them right the first time thru.  

Everything tests OK.

All this mod menu logic is buried in banPermissions(), which doesn't feel like a proper location for it...  I'll move if to a new location if desired.   Just tell me where...

### DIFF
--- a/Sources/Security.php
+++ b/Sources/Security.php
@@ -465,26 +465,22 @@ function banPermissions()
 	}
 
 	// Now that we have the mod cache taken care of lets setup a cache for the number of mod reports still open
-	if (!empty($_SESSION['rc']['reports']) && $_SESSION['rc']['time'] > $modSettings['last_mod_report_action'] && $_SESSION['rc']['id'] == $user_info['id'])
+	if (!empty($_SESSION['rc']['reports']) && !empty($_SESSION['rc']['member_reports']) && $_SESSION['rc']['time'] > $modSettings['last_mod_report_action'] && $_SESSION['rc']['id'] == $user_info['id'])
+	{
 		$context['open_mod_reports'] = $_SESSION['rc']['reports'];
+		$context['open_member_reports'] = $_SESSION['rc']['member_reports'];
+	}
 	elseif ($_SESSION['mc']['bq'] != '0=1')
 	{
 		require_once($sourcedir . '/Subs-ReportedContent.php');
 		$context['open_mod_reports'] = recountOpenReports('posts');
-	}
-	else
-		$context['open_mod_reports'] = 0;
-
-	// Figure out how many reported members need attention
-	if (!empty($_SESSION['rc']['member_reports']) && $_SESSION['rc']['time'] > $modSettings['last_mod_report_action'] && $_SESSION['rc']['id'] == $user_info['id'])
-		$context['open_member_reports'] = $_SESSION['rc']['member_reports'];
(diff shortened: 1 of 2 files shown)


Sesquipedalian:
> All this mod menu logic is buried in banPermissions(), which doesn't feel like a proper location for it... I'll move if to a new location if desired. Just tell me where...

That does sound like an odd place for it, yes. You're more familiar with this part of the code than I am at this point, so I'm going to flip the question back to you. Where do you think would be a better place to put this logic?

If you decide to move it, someone can review the changes once you commit them. If you decide not to, then this look good as it stands.

sbulen:
Fixed some timing issues caused by the fact that both stats share the same timestamp...  This would make the stats sometimes not update properly in the mod menu.

sbulen:
I will hunt around for a new home for this code tomorrow.  First thought is Subs-ReportedContent.php.
In the meanwhile, it does test OK.  
Maybe we can wrap this issue up before its 3rd birthday!

Sources/Security.php
```diff
@@ -465,25 +465,22 @@ function banPermissions()
 	}
 
 	// Now that we have the mod cache taken care of lets setup a cache for the number of mod reports still open
-	if (!empty($_SESSION['rc']) && $_SESSION['rc']['time'] > $modSettings['last_mod_report_action'] && $_SESSION['rc']['id'] == $user_info['id'])
+	if (!empty($_SESSION['rc']['reports']) && !empty($_SESSION['rc']['member_reports']) && $_SESSION['rc']['time'] > $modSettings['last_mod_report_action'] && $_SESSION['rc']['id'] == $user_info['id'])
+	{
```
jdarwood007: Correct me if I'm wrong, but this means the cache is only valid if ['rc']['reports'] and ['rc']['member_reports'] are both not 0, instead of ['rc'] just existing.  So if you have no member reports open, it is going to ignore the cache completely?

sbulen: That is a good point - underutilization of the cache.  isset() would be better than !empty()...   

I was avoiding a check on just ['rc'] to avoid issues like the initial bug, where ['rc'] was set (due to 'reports' being in there) so it then assumed 'member_reports' was in there when it wasn't.   Since the function is written to do one or the other, not both, we shouldn't assume to always find both.  Hope this makes sense...  

I'll test isset()...

sbulen:
Changes made per @jdarwood007 's suggestion.  Thanks for the input.  Tests OK.   

I think I'm done with this.  I had tried moving all the moderator-related logic out of banPermissions() to a few other places that I thought made better sense, but none worked.  So I left that logic where it is.

Sesquipedalian (approved):

PR #4151, opened 2017-06-22
Topic without list fixing #4149

### LINKED ISSUE #4149: Undefined offset: 2 in QueryString.php line 194
Fresh load from github.  I get this navigating around a newly upgraded site.  

Notice
: Undefined offset: 2 in
\Sources\QueryString.php
on line
194

Oddly, I don't get a similar message on a newly INSTALLED site.

albertlast:
i doesn't understand list() fully,
but this function behavior is different between 5.x and 7.x http://php.net/manual/en/function.list.php
maybe a dangerous function for us.

But this should fix #4149

### DIFF
--- a/Sources/QueryString.php
+++ b/Sources/QueryString.php
@@ -192,10 +192,7 @@ function cleanRequest()
 		// Dots are useful and fun ;).  This is ?topic=1.15.
 		elseif (strpos($_REQUEST['topic'], '.') !== false)
 		{
-			$parts = explode('.', $_REQUEST['topic']); 
-			list ($_REQUEST['topic'], $_REQUEST['start']) = $parts;
-			if (count($parts) > 2)
-				$_REQUEST['page_id'] = $parts[2];
+			list ($_REQUEST['topic'], $_REQUEST['start'], $_REQUEST['page_id']) = explode('.', $_REQUEST['topic'] . '.');
 		}
 
 		$topic = (int) $_REQUEST['topic'];
--- a/Sources/QueryString.php
+++ b/Sources/QueryString.php
@@ -192,13 +192,9 @@ function cleanRequest()
 		// Dots are useful and fun ;).  This is ?topic=1.15.
 		elseif (strpos($_REQUEST['topic'], '.') !== false)
 		{
-			$parts = array();
-			$parts = explode('.', $_REQUEST['topic']);
-			$size = count($parts);
-			$_REQUEST['topic'] = $parts[0];
-			if ($size > 1)
-				$_REQUEST['start'] = $parts[1];
-			if ($size > 2)
+			$parts = explode('.', $_REQUEST['topic']); 
+			list ($_REQUEST['topic'], $_REQUEST['start']) = $parts;
+			if (count($parts) > 2)
 				$_REQUEST['page_id'] = $parts[2];
 		}
 
--- a/Sources/QueryString.php
+++ b/Sources/QueryString.php
@@ -191,7 +191,16 @@ function cleanRequest()
 			list ($_REQUEST['topic'], $_REQUEST['start']) = explode('/', $_REQUEST['topic']);
 		// Dots are useful and fun ;).  This is ?topic=1.15.
 		elseif (strpos($_REQUEST['topic'], '.') !== false)
-			list ($_REQUEST['topic'], $_REQUEST['start'], $_REQUEST['page_id']) = explode('.', $_REQUEST['topic']);
+		{
+			$parts = array();
+			$parts = explode('.', $_REQUEST['topic']);
+			$size = count($parts);
+			$_REQUEST['topic'] = $parts[0];
+			if ($size > 1)
+				$_REQUEST['start'] = $parts[1];
+			if ($size > 2)
+				$_REQUEST['page_id'] = $parts[2];
+		}
 
 		$topic = (int) $_REQUEST['topic'];
 


frandominguezl:
Why would this be dangerous?

albertlast:
because php 7 fill the var in a different order as the php 5 does.

frandominguezl:
Oh, yes... I just see it ^^

sbulen:
I don't think the php7 index reverse order is impacting this, as it refers to using arrays with [] as the list() parameters - that is when it will determine the index numerical value.  (See:  http://php.net/manual/en/migration70.incompatible.php)   

I think the issue I found here is much more basic, we were trying to find the 3rd value where there wasn't a 3rd value.  Bear in mind my failure was in php 5.6.  

I admit I'm confused, because it did not fail on another (almost empty, freshly installed) instance, with a similar x.y URL (2 values, one period).  Why did one x.y URL fail & another x.y URL not fail?  That makes me think this particular logic is not always invoked, i.e., your intended benefit might not always occur.

albertlast:
The last one is optional,
but it look like that i got my env not setted on notice level that is the reason that i got no notice..

any way this pr fix this,
feel free to give me an advice if i had to write this i another way.

jdarwood007:
There is no difference programing wise how list is handled.  It is how it assigns the vars in list is changed, but doesn't change how they are used.  It does appear to throw errors differently.  Although why we don't see the "PHP Notice" in either version and only the "Notice" is a good question.

```
PHP 5.4.16 (cli) (built: Nov  6 2016 00:29:02)
Copyright (c) 1997-2013 The PHP Group
Zend Engine v2.4.0, Copyright (c) 1998-2013 Zend Technologies
[jeremy@localhost ~]$ php -r 'list($a, $b, $c) = array(1,2,3); var_dump($a,$b,$c);'
int(1)
int(2)
int(3)
[jeremy@localhost ~]$ php -r 'list($a, $b, $c) = array(1,2); var_dump($a,$b,$c);'
PHP Notice:  Undefined offset: 2 in Command line code on line 1
int(1)
int(2)
NULL
```

```
Zeus:~ jeremy$ php -v
PHP 7.1.1 (cli) (built: Jan 21 2017 13:36:56) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2017 Zend Technologies
Zeus:~ jeremy$ php -r 'list($a, $b, $c) = array(1,2,3); var_dump($a,$b,$c);'
int(1)
int(2)
int(3)
Zeus:~ jeremy$ php -r 'list($a, $b, $c) = array(1,2); var_dump($a,$b,$c);'
PHP Notice:  Undefined offset: 2 in Command line code on line 1

Notice: Undefined offset: 2 in Command line code on line 1
int(1)
int(2)
NULL
```

sbulen:
If this tests out ok, I'd say go for it.  I'm concerned about impact to beta testers at the moment.   

I wasn't able to look/test deeper last night, earliest I will be able to is tomorrow.    

The nagging concern in the back of my mind is that it works sometimes...  It would be helpful to understand what is happening/not happening at those times.

albertlast:
this code is changed after beta 3 release so the beta tester are not effected.
that it work sometimes i believe is related that you to some time "next page" or "prev page" in this cases all three var are filled in other case where you jump directly to one page only the frist two are filled.

sbulen:
Any fresh downloads from github are affected.  

The installation that works is a fresh 2.1 install.  Maybe there aren't enough posts?  No pagination at all.  I didn't see a check for size or # of entries in that logic, though.  

The installation that failed is a fresh 2.1 upgrade of a vanilla 2.0 install.  That one has about 15 posts (different languages - I use it to test UTF8 upgrades).  No pagination at all there either.   

I don't understand why one worked & the other didn't...  It feels like we're missing something...

albertlast:
For me makes no sens what du talking,
Tze Topic issue is fixed


Gesendet: 23.06.2017 5:09 nachm.
An: SimpleMachines/SMF2.1
Cc: albertlast; Author
Betreff: Re: [SimpleMachines/SMF2.1] Topic without list fixing #4149 (#4151)


Any fresh downloads from github are affected.

The installation that works is a fresh 2.1 install. Maybe there aren't enough posts? No pagination at all. I didn't see a check for size or # of entries in that logic, though.

The installation that failed is a fresh 2.1 upgrade of a vanilla 2.0 install. That one has about 15 posts (different languages - I use it to test UTF8 upgrades). No pagination at all there either.

I don't understand why one worked & the other didn't... It feels like we're missing something...

sbulen:
I have tested this change & it works.  

Further, I have confirmed that the exact same logic is executed in both of my environments.  In fact, using the old logic, the exact same results are encountered - topic & start are set, & pageid is null.  

The remaining issue is the fact that the error is treated differently between a fresh install (no error) and an upgrade (error printed at the top of the page).  

I do not understand why a fresh install behaves differently than an upgrade here.  I will log a separate ticket for this.

albertlast:
Maybe different error levels,
see that error is only a "notice" so it's the lowest version of an "error" which you can get.

sbulen:
Let's continue the discussion for this new error over in #4152

Sesquipedalian (changes requested):
An easier approach might be the following:
```
list ($topic, $start, $page_id) = explode('.', $_REQUEST['topic']);
foreach (array('topic', 'start', 'page_id') as $key)
	$_REQUEST[$key] = $$key;
```

albertlast:
Looks very evil,
$$ is a bad thing.

```
$parts = explode('.', $_REQUEST['topic']);
list($_REQUEST['topic'], $_REQUEST['start']) = $parts;
if($parts == 3 )
 $_REQUEST['page_id'] = $parts[2];
```

would be better

Sesquipedalian:
> $$ is a bad thing.

How so? PHP's variable variables are one of its great features.

albertlast:
this is kind of "magic" and this evil stuff like eval.

Sesquipedalian:
It's not magic. It's a clearly documented and predictable part of PHP. http://php.net/manual/en/language.variables.variable.php

albertlast:
You find eval too in the documentation: http://php.net/manual/de/function.eval.php
so what is the point?
you loop to the same thing like my version only make the code harder to read.
And i'm not 100% sure if this solve the notice issue...

Sesquipedalian:
> you loop to the same thing like my version only make the code harder to read.

I think my version is easier to read, and it is more efficient.

>And i'm not 100% sure if this solve the notice issue...

At the very least, it will take care of any issues involving how list handles assigning values to array elements. The notice issue is another matter.

albertlast:
the notice issue is the reason for this pr,
so the other points i don't care feel free to change afterwards.

And it wouldn't solve the possible issue with the assigning stuff..

sbulen:
The original issue was due to the fact that a particularly ugly warning was issued in one of my test environments when attempting to assign a value to the non-existent 3rd node in the topic within the list() command.  

I suspect the same thing will happen here:
```
list ($topic, $start, $page_id) = explode('.', $_REQUEST['topic']);
foreach (array('topic', 'start', 'page_id') as $key)
	$_REQUEST[$key] = $$key;
```
...because it is still referencing a non-existent object in list() - it's the exact same line that failed.  

I like albertlast's last example, as it only does the assignment after confirming the 3rd node exists.   I think he wants count($parts) though.  

(I have no issue with the variable variable...  I do need to squint at 'em to interpret them properly though...)

Sesquipedalian:
> ...because it is still referencing a non-existent object in list() - it's the exact same line that failed.

Derp. It turns out that I was testing in an environment that was configured not to give the notice. 

@albertlast's latest version works well. 👍

Sesquipedalian (approved):

Sesquipedalian:
... Actually, here's an even simpler solution:

```
list ($_REQUEST['topic'], $_REQUEST['start'], $_REQUEST['page_id']) = explode('.', $_REQUEST['topic'] . '.');
```

By appending that trailing `.`, we can guarantee that we have at least three elements in the exploded array. `$_REQUEST['page_id']` will either be a value or an empty string.

Sesquipedalian (approved):

PR #4156, opened 2017-06-25
Update Load.php

Underdog-01:
SQL column does not appear to have a default value of '' and/or NOT NULL resulting in a possible undefined log error.
I came across this issue on a test forum while accessing part of the admin section.


Sources/Load.php
```diff
@@ -1280,10 +1280,10 @@ function loadMemberData($users, $is_name = false, $set = 'normal')
 		while ($row = $smcFunc['db_fetch_assoc']($request))
 		{
 			// If the image proxy is enabled, we still want the original URL when they're editing the profile...
-			$row['avatar_original'] = $row['avatar'];
+			$row['avatar_original'] = !empty($row['avatar']) ? $row['avatar'] : '';
 
 			// Take care of proxying avatar if required, do this here for maximum reach
-			if ($image_proxy_enabled && !empty($row['avatar']) && stripos($row['avatar'], 'http://') !== false)
+			if ((!empty($row['avatar'])) && $image_proxy_enabled && stripos($row['avatar'], 'http://') !== false)
```
Sesquipedalian: What's the point of this change?

Sesquipedalian (changes requested):

Sesquipedalian:
If the avatar column can be NULL, then we probably want to change the installer and upgrader files to make it a NOT NULL.

Sources/Load.php
Underdog-01: To ensure the column has a value prior to using stripos .. I did not see an error from that line but put the edit in to cover the possibility as I have seen undefined errors with my own projects on similar logic. The brackets could go around the first 2 variables or your other suggestion I also thought of prior can be implemented which will cover any null possibility.

Underdog-01:
I can implement:  ~ not null default '' ~ but prior to doing that I need to check/test that there is not an attempt to insert a null value anywhere in the current SMF 2.1 coding. I will test things out this weekend & implement the edits unless someone else reads this & beats me to it.

jdarwood007:
There isn't a case where that can happen though as you just flipped the first 2 checks around and the strpos is the 3rd check.

Sources/Load.php
Sesquipedalian: As @jdarwood007 said below,

>There isn't a case where that can happen though as you just flipped the first 2 checks around and the strpos is the 3rd check.

This change doesn't actually do anything.

XinYenFon:
extra txt.

Sesquipedalian:
This branch is kind of messed up now, @Underdog-01. If you don't mind, please make a fresh branch, reimplement your finalized changes in that, and then submit it as a new PR.

Underdog-01:
Agreed.. but it will have to be Friday evening.

Sesquipedalian:
No problem. Thanks.

PR #4157, opened 2017-06-26
Small improve in ScheduledTasks

albertlast:
get sql align (removed inner)
replace delete from with truncate
removed old db checks

### DIFF
--- a/Sources/ScheduledTasks.php
+++ b/Sources/ScheduledTasks.php
@@ -170,9 +170,9 @@ function scheduled_approval_notification()
 		SELECT aq.id_msg, aq.id_attach, aq.id_event, m.id_topic, m.id_board, m.subject, t.id_first_msg,
 			b.id_profile
 		FROM {db_prefix}approval_queue AS aq
-			INNER JOIN {db_prefix}messages AS m ON (m.id_msg = aq.id_msg)
-			INNER JOIN {db_prefix}topics AS t ON (t.id_topic = m.id_topic)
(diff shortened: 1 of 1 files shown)


jdarwood007:
What would the difference between JOIN and INNER JOIN be here?

albertlast:
it's the same -> no different
one of the reason why i said this are small changes.

jdarwood007:
Well then to keep consistent I think we should either keep everything with INNER JOIN or remove INNER from all queries everywhere.

I would think that specifying INNER makes it clearer what type of join is being performed, even if we know the default is INNER.

albertlast:
i hat this discussion...
"inner join" is redundancy,
like "dead corpse", it's clear that the corpse is dead we don't need add "dead" in front of this.

And when i take you logic for different topic like "if" which you can write in "if(){} else {}" or "( ? : )",
would you now start to write every short version to the long version (= if(){} )?

jdarwood007:
I was more saying we should do it consistently rather than just to ScheduledTasks.php.  We should correct all queries to follow this.

Sesquipedalian:
LGTM. Do you want me to merge this now, @albertlast, or are you planning to replace `INNER JOIN` with just `JOIN` everywhere?

albertlast:
the inner join stuff i only replace when i touched the file,
when you like to keep every where the same i could try to make a own pr with the target to change this in whole project.

Sesquipedalian:
Okay. In that case, I'll go ahead and merge this now.

PR #4159, opened 2017-06-27
Improve Calendar runtime

albertlast:
Notice two points:
one unnessary query for the case of the user
timeformat get's slow when callet high amount (13k in my case)

Last one I look into, tomorrow.


colinschoen (approved):
LGTM, approval

PR #4160, opened 2017-06-28
Much faster timeformat

albertlast:
Notice on calendar call that this function takes ~78% runtime,
issues was that the setlocal get called 13k times.

This changes reduce the call at one when process_safe is off (default) in
theory a different process can change setlocal and then our runtime is affected too.
But this is in my eyes less potential that this happen -> so i take the fast way.

left side old, right side new
grafik


albertlast:
I realize that timeformat is used across smf,
so the complet board get faster

Sesquipedalian:
Very nice!

PR #4161, opened 2017-06-28
Removed loop in db replace

albertlast:
i notice in the display benchmark,
that 30 % of the runtime getlost in the replace stuff.

the user_info stuff was essential the issue,
without understand what it do i try to do i revamp the idea in a more faster way

left old right new
grafik

maybe @Sesquipedalian can comment if the code keep doing the same.

### DIFF
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -170,11 +170,8 @@ function smf_db_replacement__callback($matches)
 	if ($matches[1] === 'db_prefix')
 		return $db_prefix;
 
-	if (!empty($user_info) && strpos($matches[1],'query_') !== false)
-	{
-		if (isset($user_info[$matches[1]]))
-			return $user_info[$matches[1]];
-	}
+	if (isset($user_info[$matches[1]]) && strpos($matches[1], 'query_') !== false)
+		return $user_info[$matches[1]];
 
 	if ($matches[1] === 'empty')
 		return '\'\'';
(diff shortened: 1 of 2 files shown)


Sources/Subs-Db-mysql.php
```diff
@@ -170,11 +170,10 @@ function smf_db_replacement__callback($matches)
 	if ($matches[1] === 'db_prefix')
 		return $db_prefix;
 
-	if (!empty($user_info))
+	if (!empty($user_info) && strpos($matches[1],'query_') !== false)
 	{
-		foreach (array_keys($user_info) as $key)
-			if (strpos($key, 'query_') !== false && $key === $matches[1])
-				return $user_info[$matches[1]];
+		if (isset($user_info[$matches[1]]))
```
jdarwood007:
```
if (!empty($user_info) && strpos($matches[1],'query_') !== false && isset($user_info[$matches[1]]))
```

I would wonder if doing a query identifier would be faster than doing a strpos match.  If we really want to save even more time.

Sesquipedalian: If we are checking `isset($user_info[$matches[1]])` we don't even need to check `!empty($user_info)`, so we could reduce this even further to just:
```
if (isset($user_info[$matches[1]]) && strpos($matches[1],'query_') !== false)
```

Sesquipedalian: To be more clear, I mean that the best option to replace these loops is probably this:
```
if (isset($user_info[$matches[1]]) && strpos($matches[1], 'query_') !== false)
	return $user_info[$matches[1]];
```

It's basically the same idea that you already implemented, @albertlast, but with the unnecessary bits trimmed out.

Sesquipedalian (approved):

PR #4165, opened 2017-06-30
Try to get the time from Database

albertlast:
I notice even after my optimization i got a high time consume by the subs-calendar,
highest issue are the time is used to build the different time information for the events.

 so i came up with this proof of concept where i do the most stuff on database side instead of php.
this poc lowered the start time of the index page from ~1,1 to 0,7 sec -> 300-400 ms safed time

so i would like to hear for exp from @Sesquipedalian if this would be a way or it's possible to improve
the getevent function that this not needed or maybe a hybrid (where the database deliever more but not all data)

i should mention that this query takes 2 ms instead of 0,5 ms.
on the risk side i'm not 100% sure if i can do all this things in mysql.

### DIFF
--- a/Sources/Subs-Calendar.php
+++ b/Sources/Subs-Calendar.php
@@ -135,10 +135,39 @@ function getEventRange($low_date, $high_date, $use_permissions = true)
 	$low_object = date_create($low_date);
 	$high_object = date_create($high_date);
 
+	$modSettings['disableQueryCheck'] = true;
 	// Find all the calendar info...
 	$result = $smcFunc['db_query']('calendar_get_events', '
 		SELECT
-			cal.id_event, cal.title, cal.id_member, cal.id_topic, cal.id_board,
+			to_char(cal.end_date, \'Mon DD, YYYY\') "end_date_local"
+			,to_char(cal.end_date, \'Mon DD, YYYY\') "end_date_orig"
+			,cal.end_time "end_time_local"
+			,cal.end_time "end_time_orig"
+			,cal.end_date::timestamp "end_datetime"
+			,to_char(cal.end_date, \'DD\') "end_day"
+			,to_char(cal.end_time, \'HH\') "end_hour"
+			,to_char(cal.end_time, \'MI\') "end_minute"
+			,to_char(cal.end_time, \'SS\') "end_second"
+			,to_char(cal.end_date, \'MM\') "end_month"
+			,to_char(cal.end_date, \'YYYY\') "end_year"
+			,to_char(cal.start_date, \'Mon DD, YYYY\') "start_date_local"
+			,to_char(cal.start_date, \'Mon DD, YYYY\') "start_date_orig"
+			,cal.start_time "start_time_local"
+			,cal.start_time "start_time_orig"
+			,cal.start_date::timestamp "start_datetime"
+			,to_char(cal.start_date, \'DD\') "start_day"
+			,to_char(cal.start_time, \'HH\') "start_hour"
+			,to_char(cal.start_time, \'MI\') "start_minute"
+			,to_char(cal.start_time, \'SS\') "start_second"
+			,to_char(cal.start_date, \'MM\') "start_month"
+			,to_char(cal.start_date, \'YYYY\') "start_year"
+			,cal.end_date - cal.start_date "diff_days" 
+			,(cal.end_date + coalesce( cal.end_time , \'00:00:00\'::time)) "test_time"
+			,to_char(cal.end_date + coalesce( cal.end_time , \'00:00:00\'::time) at time zone \'UTC\', \'YYYY-MM-DD"T"HH24:MI:SSOF\') "end_iso_gmdate"
+			,to_char(cal.start_date + coalesce( cal.start_time , \'00:00:00\'::time) at time zone \'UTC\', \'YYYY-MM-DD"T"HH24:MI:SSOF\') "start_iso_gmdate"
+			,extract(epoch from cal.end_date + coalesce( cal.end_time , \'00:00:00\'::time)) "end_timestamp"
+			,extract(epoch from cal.start_date + coalesce( cal.end_time , \'00:00:00\'::time)) "start_timestamp"
+			,cal.id_event, cal.title, cal.id_member, cal.id_topic, cal.id_board,
 			cal.start_date, cal.end_date, cal.start_time, cal.end_time, cal.timezone, cal.location,
 			b.member_groups, t.id_first_msg, t.approved, b.id_board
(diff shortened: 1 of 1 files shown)


jdarwood007:
I will be honest, that is a ugly looking query.
Instead of passing strings directly, should use the {string:handle} to pass things like DD, MM, etc even if its static text.  Just looks cleaner.

albertlast:
that is a >poc< that mean,
that i only do the min of needed stuff to be able to test this.

Sesquipedalian:
Hm. The potential speed boost is enticing, but I have a couple of concerns.

1. **Code fragmentation**

	`buildEventDateTimes()` consolidates into one function all the code needed to build a complete list of date and time info for an event. Before I created it, the calendar related code was a mess in this regard: there were multiple versions of code to do essentially the same job, and none of them were consistent with each other. It was awful. I'm sure I don't need to explain to anyone here why it is better to have a single version of this code contained in a function that other functions can call whenever necessary. :)

	So, my first concern with having the database do the job instead of PHP is that we would go back to having multiple, redundant versions of the same code scattered all over the place. At the very least, that would be repetative. Worse, that would make it very easy for inconsistencies between the different versions to arise in the future.

	If you can figure out a way to consolidate the db_query code for the different calendar functions into a single helper function, maybe we could make that work.

2. **Localization**

	`buildEventDateTimes()` calls `timeformat()` in order to localize a number of the date and time strings according to the user's preferred format. If you want to do all the formatting in the query, you will need to figure out a way to incorporate the user's preferred format into it. I'm not sure if MySQL and PostgreSQL even have equivalents for all of PHP's `strftime()` formatting options. If they don't, it will be impossible to make this work. If they do, you will need to figure out a way to translate the user's preferred date and time formats from `strftime()` syntax to SQL syntax and then incorporate those translated format strings into your query.


It might be possible to address both of these concerns, and if so, I would be very interested to see what you come up with. But if not, we might just need to accept the slower performance we get using PHP instead of SQL to do this job.

albertlast:
This why i ask if we able to speed up more the php version,
like the buildEventDateTimes calls many times the timeformat where maybe any array with all date could be done faster on similiar stuff,
also i notice multiple calls of date_create on the same data:
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs-Calendar.php#L174
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs-Calendar.php#L1565

some simliar calls exists.

strftime is possible in to_char part,
https://stackoverflow.com/questions/33009076/change-postgresql-date-language-from-request
your limitation are this locals which is database support (which can be different to the php)

like i mention in the start post,
maybe a hybrid way could be the right way,
wher we try get more data directly from the database and the rest(like strtime) we do in php

Sesquipedalian:
Maybe. It might be worth investigating further, at least. :)

In the meantime, I just noticed that `buildEventDatetimes()` was re-creating the values for `$date_format` and `$time_format` every time it ran. I've just submitted #4169, which should improve performance in `buildEventDatetimes()` a bit.

Sesquipedalian:
> like the buildEventDateTimes calls many times the timeformat where maybe any array with all date could be done faster on similiar stuff,

I think you might be right to suspect that timeformat() is a cause of a lot of the slowdown. PHP's modern date objects are pretty quick to work with, so the several calls to timeformat() are likely suspects here.

Unfortunately, timeformat() doesn't accept arrays. But maybe we could write a new function to replace timeformat() that would be more efficient. There are several places in buildEventDateTimes() where I have to do silly things like `timeformat(strtotime(date_format($start_object, 'Y-m-d H:i:s')), $date_format, 'none')` in order to get a value that timeformat() will handle properly. If we could eliminate the need for all that, it might help a lot.

albertlast:
we could also extend timeformat that it accept array and does then also return array.

Sesquipedalian:
As I think about it some more, I suspect that the best way forward might be to change the syntax we use in `$user_info['time_format']`. Right now we use the old-fashioned [strftime syntax](http://php.net/manual/en/function.strftime.php) instead of PHP's native [date syntax](http://php.net/manual/en/function.date.php) that works most directly with PHP's modern date objects (which are both fast and convenient to work with). Because of that, `buildEventDateTimes()` has to do a lot of work translating PHP's date objects into strings and Unix timestamps for `timeformat()` and back again. If we changed from using strftime syntax to native date syntax, a whole bunch of the work that `buildEventDateTimes()` has to do could be eliminated. For that matter, so would a lot of the work that `timeformat()` itself has to do. If we can do everything with date objects and native date syntax, all of this code will become both simpler and faster.

Oldiesmann:
The only issue is it would be a slight pain to update the preferred formats if we switch to using date() instead of strftime(), but we could probably handle that by simply mapping the strftime() flag with the equivalent date() flag in an array then doing an str_replace. While not every strftime flag has an equivalent for date, the ones that don't have an equivalent for date are things we're not likely to run into anyway (more advanced stuff like date and/or timestamp formatted based on locale, week number according to ISO-8601:1988 standards, etc.)

albertlast:
Did we know that we got better performance (as the existing solution) when we go this way?

Sesquipedalian:
We'd need to develop and test a solution to know for sure, but I expect so. If we can use date objects consistently and avoid translating into and out of string formats, then all the processing will be done at the C level. If we do it properly, that should be significantly faster.

albertlast:
Look like that one of the main issue is that date and dateTime class are not able to provide local namens for weekdays etc.
https://stackoverflow.com/questions/5941121/php-strftime-date-or-datetime-which-is-better https://php.earth/doc/general/date-time
so it's doesn't look that this solve any issue.

i would like to mention that the runtime of strftime is not so high 
grafik
you see on the picture that timeformat takes 27 of 30% for his self and only 1.78% is used by strftime.

By disable part of the codes,
i came up with the result that the manuel local stuff: https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs.php#L819
takes ~17% of the 27% runtime --> more than 50% is spend here.

Oldiesmann:
If we switched we'd be using the [date](https://www.php.net/date) function, not the DateTime class.

As far as that code that takes a long time - I'll have to look into that further when I have time and figure out what exactly we're doing there. It seems like a ridiculous amount of work for something pretty trivial.

albertlast:
Like i said in comment before in my eyes is date or strftime not the issue,
the issue is the amount of function calls here.

So i added a pr which provide a fetch all function,
with this we could call the buildEventDatatimes once (after this function is modified by someone).

The next step would be that the timeformat is called once.

albertlast:
created #4202  with a php version

PR #4169, opened 2017-07-01
Makes $date_format & $time_format static in buildEventDatetimes()

Sesquipedalian:
This should improve performance a bit in buildEventDatetimes().


albertlast:
in my test case it's look like that this doesn't speed up the runtime (get a litte bit slower).

left old version right your changes:
grafik

Sesquipedalian:
Does your test case use a single calendar event, or multiple ones? Declaring a static variable will increase performance on the second and subsequent executions of buildEventDateTimes().

albertlast:
you see can see on the pic that buildeventdatetimes was called 268 times and
this is the amount of calls when i call the index page of my board (without going into calendar area).

Issue #4173, opened 2017-07-03
json_(de|en)code in $smcFunc


jdarwood007:
Was there a reason we didn't do this?

Since the json functions are heavily used in our code, it would make sense we move them into $smcFunc so if in the future, like the serialized issue, we have to fix something we can just override it in $smcFunc and be done?

I forsee basically adding 2 items to $smcFunc
$smcFunc['json_encode'] = 'json_encode';
$smcFunc['json_decode'] = 'smf_json_decode';

The code would look a bit cleaner and again, if we need to clean something up later it would be easier.

We put mostly db stuff and utf-8 handling in $smcFunc right now, but that isn't to say we can't do additional stuff in there like this heavily used functions.

Sesquipedalian:
That suggestion makes sense to me.

sbulen:
FYI - there's a bunch of such functions out there in Load.php.  smcFunc isn't just DB stuff...  

I'm not sure everyone knows they're there.   Even string functions:
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Load.php#L130

Things get added to smcFunc in a few different places...   Hard to tell what the whole set is.

jdarwood007:
Most of stuff is added in Load.php and then the database abstraction stuff adds as needed for its different pieces.

I will work on getting things converted over.  Lots of places json is used and I need to make sure that $smcFunc is global before using it in a function.

PR #4177, opened 2017-07-05
Update all usages of json

jdarwood007:
Changed all json_(smf_de|en)code\(  to $smcFunc['json_\2code'](
Hopefully ensured $smcFunc is global in said functions

I have not tested this yet.  Getting it here for review and testing.  It should all work just fine though if I didn't miss a $smcFunc global somewhere.

I purposely did not change install or upgrade files because they only appear to load up a basic $smcFunc when they initialize the database.  Sometimes it loads up htmlspecialchars on its own.

Proxy.php can't be updated as it doesn't load SMF, so $smcFunc isn't around.

There is one call in Load.php that I didn't update, I added a comment for that which is because it happens prior to $smcFunc being populated with non database stuff.

### DIFF
--- a/Sources/Admin.php
+++ b/Sources/Admin.php
@@ -25,7 +25,7 @@ if (!defined('SMF'))
 function AdminMain()
 {
 	global $txt, $context, $scripturl, $modSettings, $settings;
-	global $sourcedir, $options, $boarddir;
+	global $smcFunc, $sourcedir, $options, $boarddir;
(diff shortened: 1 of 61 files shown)


jdarwood007:
I've tested as much as I can.  I haven't seen anything broken.

I feel this is ready to go.

jdarwood007:
This fixes #4173

PR #4178, opened 2017-07-05
added Fetch All to DB API

albertlast:
added $smcFunc['db_fetch_all'] to mysql and pg
could be use full where you try to proceed many row sets at once like the time topic: https://github.com/SimpleMachines/SMF2.1/pull/4165

### DIFF
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -936,7 +936,7 @@ function smf_is_resource($result)
 function smf_db_fetch_all($request)
 {
 	// Return the right row.
-	return @mysqli_fetch_all($request);
+	return mysqli_fetch_all($request);
 }
 
 ?>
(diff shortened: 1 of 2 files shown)


Sources/Subs-Db-mysql.php
```diff
@@ -59,6 +59,7 @@ function smf_db_initiate($db_server, $db_name, $db_user, $db_passwd, $db_prefix,
 			'db_is_resource'            => 'smf_is_resource',
 			'db_mb4'                    => false,
 			'db_ping'                   => 'mysqli_ping',
+			'db_fetch_all'				=> 'smf_db_fetch_all',
```
colinschoen: The tabbing looks off.

colinschoen (changes requested):

sbulen:
Do we want the @s in the return statements?

I don't think we want errors ignored.  Also, none of the other simply called functions use it in that fashion, e.g., mysqli_get_server_info().  We should be consistent.

albertlast:
About the @ i see no consistency,
In Subs-Db-postgresql.php every fetch methode use @ 
and we don't ignore the error because of https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Load.php#L3047
when you like i can do in pg with @ and in mysql without,
but i don't see any reason for this...

Oldiesmann:
I'm not sure why we're suppressing errors for PostgreSQL, but we aren't for MySQL at all, so at the very least be consistent that way. We should look into why we are suppressing errors in PostgreSQL for the fetch functions though.

PR #4184, opened 2017-07-08
Fixing some commas

### LINKED ISSUE #4183: comma setting by ::after looks wrong
First I notice that this comma looks wrong:
grafik
I would expect that the comma is after new,
when a new element is there

Than i looked in my bigger test area: 
grafik
More or less the same behavior .

Then when you mark a board as read is look strange:
grafik

All this is done by the
`.children a::after {
    content: ", ";
}
`
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Themes/default/css/index.css#L2939

Is this behavior realy wanted?

XinYenFon:
With this PR I added "strong" as class as well, this is useful for many people who doesn't want to add extra element to their markup also SMF's core can benefit from it to eliminate strong element.

fixes #4183

### DIFF
--- a/Themes/default/BoardIndex.template.php
+++ b/Themes/default/BoardIndex.template.php
@@ -142,7 +142,7 @@ function template_main()
 					foreach ($board['children'] as $child)
 					{
 						if (!$child['is_redirect'])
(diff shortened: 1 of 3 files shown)


XinYenFon:
@jdarwood007 seems after updates Travis failing on GPG keys ?

albertlast (approved):

XinYenFon:
I added opacity/transition trick (very light) to new button so SMF can actually look little bit modern 😆 and some updates you can read on commit messages.

sbulen:
You don't mention *whose* soul you've devoured...

PR #4185, opened 2017-07-08
Stop Edit Board when no board was found

albertlast:
I can't see any way how someone was able to do so,
but this case exists: https://www.simplemachines.org/community/index.php?topic=555050.0;topicseen
any way this should be allow us to handle this case nicely.

### DIFF
--- a/Sources/ManageBoards.php
+++ b/Sources/ManageBoards.php
@@ -840,12 +840,13 @@ function EditBoardSettings($return_config = false)
 	$smcFunc['db_free_result']($request);
 	
 	if (empty($recycle_boards))
-		return;
-
-	require_once($sourcedir . '/Subs-Boards.php');
-	sortBoards($recycle_boards);
-
-        $recycle_boards = array('') + $recycle_boards;
+	{
+		require_once($sourcedir . '/Subs-Boards.php');
+		sortBoards($recycle_boards);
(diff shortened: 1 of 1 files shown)


NanoSector:
This should be reversed. Instead of returning when no recycleable boards were found, you should wrap https://github.com/albertlast/SMF2.1/blob/548351610305f7bb001ce228286b78018059a45d/Sources/ManageBoards.php#L845-L846 in `if (!empty($recycle_boards))`

Sources/ManageBoards.php
```diff
@@ -838,11 +838,15 @@ function EditBoardSettings($return_config = false)
 	while ($row = $smcFunc['db_fetch_assoc']($request))
 		$recycle_boards[$row['id_board']] = $row['cat_name'] . ' - ' . $row['board_name'];
 	$smcFunc['db_free_result']($request);
-
-	require_once($sourcedir . '/Subs-Boards.php');
-	sortBoards($recycle_boards);
-
-        $recycle_boards = array('') + $recycle_boards;
+	
+	if (empty($recycle_boards))
```
Oldiesmann: Shouldn't that be `!empty`?

albertlast: yes is better

PR #4192, opened 2017-07-10
Fixes #4190 Where SMF isn’t yet loaded again properly but we are usin…

### LINKED ISSUE #4190: Upgrader - Undefined index: json_decode
Ran into this running the upgrader just now

image

jdarwood007:
…g SMF functions.


sbulen:
Sorry...  That would have been me with Fixes #4085 PR #4112 ...   Addressing the fact that the upgrader couldn't properly report errors.  

So I brought up enough of SMF for error reporting, once there was a DB to report the error into... 

Problem with the upgrader is it has got to bootstrap itself, which makes it a little more challenging.

PR #4197, opened 2017-07-14
Revert seek method

### LINKED ISSUE #4196: Topic display no longer pages correctly (EDIT: has security implications, too)
It looks like #4134 introduced a bug with proper paging in topics. There are multiple ways the bug shows up, but it is really all the same bug. #4151 tried to fix part of the situation, but more problems remain.

Assume in the following examples that topic 53 contains 35 replies, and the messages per page option is set to 15.

## Old, expected behaviour
1. Go to http://www.example.com/forum/index.php?topic=53.10
2. Be served a page containing the original post plus replies 1-14 (total 15). The page index indicates that this is page 1 of 3.
3. Go to http://www.example.com/forum/index.php?topic=53.32
4. Be served a page containing replies 30-35 (total 6). The page index indicates that this is page 3 of 3.
5. Create a new post, which becomes reply 36.
6. After posting, be served a page containing replies 30-36 (total 7). The page index indicates that this is page 3 of 3.
7. Go to http://www.example.com/forum/index.php?topic=53.37
8. Be served a page containing replies 30-36 (total 7). The page index indicates that this is page 3 of 3.
9. Go to http://www.example.com/forum/index.php?topic=53.38
10. Be served a page containing replies 30-36 (total 7). The page index indicates that this is page 3 of 3.

## New, buggy behaviour
1. Go to http://www.example.com/forum/index.php?topic=53.10
2. Be served a page containing replies 10-24 (total 15). The desired post (reply 10) is first. The page index indicates that this is page 1 of 3.
3. Go to http://www.example.com/forum/index.php?topic=53.32
4. Be served a page containing replies 32-35 (total 4).  The page index indicates that this is page 3 of 3.
5. Create a new post, which becomes reply 36.
6. After posting, be served a page containing just  reply 36 (total 1).  The page index indicates that this is page 3 of 3.
7. Go to http://www.example.com/forum/index.php?topic=53.37
8. Be served a page containing no replies. The page index indicates that this is page 3 of 3. SMF's error log contains the error message "Undefined offset: 0" referring to line 920 of Display.php.
9. Go to http://www.example.com/forum/index.php?topic=53.38
10. Be served a page reporting the following error:
```
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '-1' at line 5
File: /Users/jon/Sites/dev/Sources/Display.php
Line: 868


FROM dev_messages
WHERE id_topic = 53
ORDER BY id_msg DESC
LIMIT 0, -1
```
SMF's error log indicates that this happened on line 868 of Display.php

## Cause
This bug appears to occur because #4134 moved the call to `constructPageIndex()` lower down in the `Display()` function. Among other things, `constructPageIndex()` normalizes the value of `$_REQUEST['start']` to an even multiple of `$context['messages_per_page']`. Doing this avoids all the buggy behaviour described above.

## Security related implications
Using the new code, when would-be hackers try to access, e.g., `http://www.example.com/forum/index.php?topic=53.0&quot;%20and%20&quot;x&quot;%3D&quot;x`, the following error message is logged, referring to line 868 of Display.php:
```
Wrong value type sent to the database. Integer expected. (start)
Function: Display
```

That means that `$_REQUEST['start']` is not being sanitized the way it should be before being passed to `$smcFunc['db_query']()`. It is good that we have multiple layers of security, which is my forum hasn't been hacked yet, but the fact that this error was generated at all means that at least one of our layers of security has been compromised by this bug. The sanitization provided by `constructPageIndex()` silently gets rid of all that sort of SQL-hacker cruft before it has a chance to get anywhere.

## Recommendation
@albertlast, I think you will need to rethink how you have structured the code in Display.php to support your new seek method.

In the meantime, I think we need to revert the following commits (i.e. #4134 and #4151):
9c19ce5fe7d62270949f5ea102da42cd7459d935
21f395b40ac4b7a8f406e1fdf094411d0eede748
1631908b7c441a86e98f69c70a5491d06849dd7c
e2b10fd24660a764ae3f5566800eb29b49348231
a36277b84303eca1fbed27cd61130e49d402fe75

Once @albertlast comes up with a new version of his seek method, we can reimplement it. :)

Sesquipedalian:
Fixes #4196 

@albertlast, I hope you can develop and submit a new implementation of your seek method.

### DIFF
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -441,8 +441,8 @@ function Display()
 	elseif (isset($_REQUEST['all']))
 		$_REQUEST['start'] = -1;
 
-	// Construct allowing for the .START method...
-	//$context['page_index'] = constructPageIndex($scripturl . '?topic=' . $topic . '.%1$d', $_REQUEST['start'], $context['total_visible_posts'], $context['messages_per_page'], true);
+	// Construct the page index, allowing for the .START method...
+	$context['page_index'] = constructPageIndex($scripturl . '?topic=' . $topic . '.%1$d', $_REQUEST['start'], $context['total_visible_posts'], $context['messages_per_page'], true);
 	$context['start'] = $_REQUEST['start'];
 
 	// This is information about which page is current, and which page we're on - in case you don't like the constructed page index. (again, wireles..)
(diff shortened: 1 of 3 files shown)


albertlast:
why we need to revert this?
we need to fix stuff...

Sesquipedalian:
So that SMF continues to work in the meantime. Once you come up with a new version, we can add it back in. :)

albertlast:
smf work this is my eye a special case and this is beta so it can be so long i fix this...

Sesquipedalian:
How long do you think it will take you to fix it? If you think it will be a quick fix, then just go ahead and make a PR to do so, and we can close this one. If you think it will take some time, then it makes more sense to revert for now and wait for you to submit a new version in a new PR.

albertlast:
This "critical" bug take you a month to find, than you had some days to wait.

Sesquipedalian:
I first noticed the problems it causes several weeks ago, but I've been busy with other things, and so it has taken me this long to track down the cause.

However, the reason I am suggesting that we revert for now is not that I think the bug is critical (although throwing database errors is a big one), but because it looked to me like fixing it would involve a significant rewrite, and therefore it might take you a while. I'm not sure how much time you have available, so I figured that if we did it this way, it would keep pressure off of you. This way you can take your time. :)

But like I said, if you think this will be a quick fix, then I will gladly close this PR. :)

albertlast:
this is realy shit,
you notice long time ago and after long time you need to rush now.

Sesquipedalian:
I'm not trying to pressure you, @albertlast. The whole idea of reverting for now is so that you *don't* need to rush out a fix.

However, I have just discovered some security-related implications of #4196, too. Right now,  `$_REQUEST['start']` is not being sanitized properly before being passed to `$smcFunc['db_query']()`. Fortunately, other layers of our security have made sure that nothing bad has actually happened yet, but even so, this bug does weaken our security. I've updated my report in #4196 to describe this in more detail and labelled it as a serious bug. Because the bug is now known to have security implications, I want to get it closed as soon as possible.

Sesquipedalian:
EDIT: Oops, I posted this reply on the wrong issue. The following is actually in reply to https://github.com/SimpleMachines/SMF2.1/pull/4198#issuecomment-315554838

No, @albertlast. 

1. These so-called special cases, which will happen quite commonly once 2.1 is released, must be handled correctly. Hackers will intentionally try to exploit this weakness. Regular users screw stuff up all the time. It is not acceptable to fill up the admin's error log with "critical error" messages every time an incorrect URL parameter is given, and even less acceptable to show unexpected content or database error messages to end users when they type something incorrectly or simply click on an outdated bookmark link. This must be fixed.

2. This statement is incorrect:
  > When a post is deletet the old version didn't help you too,
  > you could try to access page two in a topic wher no page exists anymore.
  
In reality, if a user tried to go to page 4 of a topic with only 3 pages, the old version would simply serve up page 3. The user’s error was handled silently, gracefully, and reliably.

albertlast:
1. where is the weakness? you mean that we look if the first char is a L or M or did you mention the int check? everything checked  --> show me the hacking case

2. but my post was on page 4 is now delete ther is doesn't help me that i got forwarded to page 3.
and for stuff like this you can implement a fallback, the normale post information are ther.

Sesquipedalian:
> show me the hacking case

I haven't found a completely successful one yet. But because setting `$_REQUEST['page_id']` lets an end user not just to request a certain message or page, but rather to give the database instructions on *how to find* data, it opens a strong (if at this point still theoretical) possibility that a hacker will find a way to use it to retrieve entirely unexpected information.

> but my post was on page 4 is now delete ther is doesn't help me that i got forwarded to page 3.

If you try to go to page 4 when there is no page 4, the old method shows you the last page that does exist. The new method shows you this:
screen shot 2017-07-15 at 1 38 06 pm
It doesn't matter why page 4 doesn't exist, or whether it used to exist and now doesn't. Nor does it matter why page 4 no longer exists if it used to. All that matters in this discussion is how we handle the situation if page 4 is requested when page 4 doesn't exist. The new method does not handle it well.

> and for stuff like this you can implement a fallback, the normale post information are ther.

The problem is that there currently is no such fallback. Once you create such fallbacks and checks, then we will be able to include your new, improved version of your seek method. Until such fallbacks and checks actually exist and have be tested and proven, this code is not ready to be included in SMF.

albertlast:
Yeah you didn't find any working hacking because ther no one,
the checks are ther when you don't like them than it's nice like you already done extend them.

smf is team project,
your are part of the team so you can provide the fallback too.

Sesquipedalian:
You are right that it is a team effort and that we need to work together to come up with solutions to problems. I have made the contributions that I can think of to help deal with some of the bugs in the seek method. But the remaining bugs that I described in https://github.com/SimpleMachines/SMF2.1/pull/4198#issuecomment-315521247 are serious and need to be dealt with before the new method can safely be included in SMF. Unfortunately, the only ideas I have been able to come up with to solve the remaining bugs would all perform worse than the old method, and might cause other problems anyway.

As I said at the beginning in #4196, I think it would be great to implement a version of your seek method for retrieving posts, because it is much faster. But unless and until the problems with the current version are solved, we need to pull it out of the release-2.1 branch and put it back into a feature development branch so that it can be worked on further.

I recommend that you create a new branch based on the current version of release-2.1  in your alberlast/SMF2.1 fork here on GitHub, and use that branch to continue developing your new seek method. Meanwhile, I will merge #4197 so that we don't have broken code in the release-2.1 while we wait. Once all the bugs have been worked out in your development branch, we can merge the new, improved version back in.

jdarwood007:
I would have to see a POC to understand how it could be a security issue.

But since manipulating user input (especially GET) is easily generating database errors, we need to correct that.

PR #4198, opened 2017-07-14
Fix Seeking Fixing #4196

### LINKED ISSUE #4196: Topic display no longer pages correctly (EDIT: has security implications, too)

albertlast:
Fix #4196

needs to be clean up tomorrow.

### DIFF
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -441,6 +441,19 @@ function Display()
 	elseif (isset($_REQUEST['all']))
 		$_REQUEST['start'] = -1;
 
+	$_REQUEST['start'] = (int) $_REQUEST['start'];
+	$start_invalid = $_REQUEST['start'] < 0;
+	
+	// Make sure $start is a proper variable - not less than 0.
+	if ($start_invalid)
+		$_REQUEST['start']= 0;
+	// Not greater than the upper bound.
+	elseif ($_REQUEST['start'] >= $context['total_visible_posts'])
+		$_REQUEST['start'] = max(0, (int) $context['total_visible_posts']- (((int) $context['total_visible_posts'] % (int) $context['messages_per_page']) == 0 ? $context['messages_per_page'] : ((int) $context['total_visible_posts'] % (int) $context['messages_per_page'])));
+	// And it has to be a multiple of $num_per_page!
+	else
+		$_REQUEST['start'] = max(0, (int) $_REQUEST['start'] - ((int) $_REQUEST['start'] % (int) $context['messages_per_page']));
+	
 	// Construct allowing for the .START method...
 	$context['start'] = $_REQUEST['start'];
 
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -842,10 +842,50 @@ function Display()
 	else
 		$start_char = null;
 
-	if ($start_char === 'M' || $start_char === 'L')
+	if (isset($start_char) && ($start_char === 'M' || $start_char === 'L'))
 		$page_id = substr($_REQUEST['page_id'], 1);
+	else
+		$start_char = null;
 
 	$limit = $context['messages_per_page'];
+	
+	if (isset($start_char))
+	{
+		$firstIndex = 0;
+		
+		if ($start_char === 'M')
+		{
+			$ascending = true;
+			$page_operator = '>';
+		}
+		else
+		{
+			$ascending = false;
+			$page_operator = '<';
+		}
+		
+		$request = $smcFunc['db_query']('', '
+			SELECT id_msg, id_member, approved
+			FROM {db_prefix}messages
+			WHERE id_topic = {int:current_topic} 
+			AND id_msg '. $page_operator . ' {int:page_id}'. (!$modSettings['postmod_active'] || $approve_posts ? '' : '
+			AND (approved = {int:is_approved}' . ($user_info['is_guest'] ? '' : ' OR id_member = {int:current_member}') . ')') . '
+			ORDER BY id_msg ' . ($ascending ? '' : 'DESC') . ($context['messages_per_page'] == -1 ? '' : '
+			LIMIT {int:limit}'),
+			array(
+				'current_member' => $user_info['id'],
+				'current_topic' => $topic,
+				'is_approved' => 1,
+				'blank_id_member' => 0,
+				'limit' => $limit,
+				'page_id' => $page_id,
+			)
+		);
+		
+		// Fallback
+		if ($smcFunc['db_num_rows']($request) < 1)
+			unset($start_char);
+	}
  
 	// Jump to page
 	if (empty($start_char))
@@ -877,39 +917,6 @@ function Display()
 				'blank_id_member' => 0,
 				'start' => $start,
 				'max' => $limit,
-			)
-		);
-	}
-	else //next or before page
-	{
-		$firstIndex = 0;
-		
-		if ($start_char === 'M')
-		{
-			$ascending = true;
-			$page_operator = '>';
-		}
-		else
-		{
-			$ascending = false;
-			$page_operator = '<';
-		}
-		
-		$request = $smcFunc['db_query']('', '
-			SELECT id_msg, id_member, approved
-			FROM {db_prefix}messages
-			WHERE id_topic = {int:current_topic} 
-			AND id_msg '. $page_operator . ' {int:page_id}'. (!$modSettings['postmod_active'] || $approve_posts ? '' : '
-			AND (approved = {int:is_approved}' . ($user_info['is_guest'] ? '' : ' OR id_member = {int:current_member}') . ')') . '
-			ORDER BY id_msg ' . ($ascending ? '' : 'DESC') . ($context['messages_per_page'] == -1 ? '' : '
-			LIMIT {int:limit}'),
-			array(
-				'current_member' => $user_info['id'],
-				'current_topic' => $topic,
-				'is_approved' => 1,
-				'blank_id_member' => 0,
-				'limit' => $limit,
-				'page_id' => $page_id,
 			)
 		);
 	}
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -849,6 +849,9 @@ function Display()
 
 	$limit = $context['messages_per_page'];
 	
+	$messages = array();
+	$all_posters = array();
+	
 	if (isset($start_char))
 	{
 		$firstIndex = 0;
@@ -856,12 +859,12 @@ function Display()
 		if ($start_char === 'M')
 		{
 			$ascending = true;
-			$page_operator = '>';
+			$page_operator = '>=';
 		}
 		else
 		{
 			$ascending = false;
-			$page_operator = '<';
+			$page_operator = '<=';
 		}
 		
 		$request = $smcFunc['db_query']('', '
@@ -877,16 +880,35 @@ function Display()
 				'current_topic' => $topic,
 				'is_approved' => 1,
 				'blank_id_member' => 0,
-				'limit' => $limit,
+				'limit' => $limit + 1,
 				'page_id' => $page_id,
 			)
 		);
 		
+		$found_msg = false;
+		
 		// Fallback
 		if ($smcFunc['db_num_rows']($request) < 1)
 			unset($start_char);
+		else
+		{
+			while ($row = $smcFunc['db_fetch_assoc']($request))
+			{
+				if ($row['id_msg'] != $page_id)
+				{
+					if (!empty($row['id_member']))
+						$all_posters[$row['id_msg']] = $row['id_member'];
+					$messages[] = $row['id_msg'];
+				}
+				else
+					$found_msg = true;
+			}
+			//page_id not found? -> fallback
+			if (!$found_msg)
+				unset($start_char);
+		}
 	}
- 
+
 	// Jump to page
 	if (empty($start_char))
 	{
@@ -919,22 +941,19 @@ function Display()
 				'max' => $limit,
 			)
 		);
-	}
-	
 
-	$messages = array();
-	$all_posters = array();
-	while ($row = $smcFunc['db_fetch_assoc']($request))
-	{
-		if (!empty($row['id_member']))
-			$all_posters[$row['id_msg']] = $row['id_member'];
-		$messages[] = $row['id_msg'];
+		while ($row = $smcFunc['db_fetch_assoc']($request))
+		{
+			if (!empty($row['id_member']))
+				$all_posters[$row['id_msg']] = $row['id_member'];
+			$messages[] = $row['id_msg'];
+		}
 	}
-	
+
 	// Before Page bring in the right order
 	if (!empty($start_char) && $start_char === 'L')
 		krsort($messages);
-	
+
 	// Construct the page index, allowing for the .START method...
 	$page_options = array(
 		'low_id' => $messages[0],
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -905,7 +905,11 @@ function Display()
 			}
 			//page_id not found? -> fallback
 			if (!$found_msg)
+			{
+				$messages = array();
+				$all_posters = array();
 				unset($start_char);
+			}
 		}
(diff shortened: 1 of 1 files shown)


albertlast:
the clean up version could go in two direction,
extend the constructPageIndex about a test mode (would i realize by extending the options array) or
create a new function (checkStart) in sub with the test stuff lett it call from constructPageIndex
and display.
@Sesquipedalian any comment or different idea?

Sesquipedalian:
This fixes the problems with `$_REQUEST['start']` having the wrong value. :thumbsup: I'd be happy to make some suggestions for how it could be tidied up. Let me play with it a bit and see what I come up with. :)

In the meantime, I'm afraid the seek method itself is still causing problems.

1. The minor bug: If I go directly to the last page (page 3 using my example in #4196), the 'L' seek value is incorrectly set to the id of the last message displayed, rather than the first.

2. The major bug: If the numerical part of the `$_REQUEST['page_id']` doesn't match correctly with the id of the message that is *supposed* to be the first or last message on the page, the content of other messages will be displayed instead of the content that should have been. This is probably easier to explain by examples.

If you go right now to my test site at http://stovell.noip.me/~jon/dev/index.php?topic=53.15, you will see a series of posts whose content matches their reply number on the thread. For example, the content of reply 15 is '15'. The message id of that message is 69. 

  1. If you go to http://stovell.noip.me/~jon/dev/index.php?topic=53.15.M68, everything works fine. Message id 69 shows up as reply 15, and message id 83 shows up as reply 29.

  2. If you go to http://stovell.noip.me/~jon/dev/index.php?topic=53.15.M67, wrong messages appear. Message id 68 shows up as reply 15, and message id 82 shows up as reply 29.

  3. If you go to http://stovell.noip.me/~jon/dev/index.php?topic=53.15.M69, wrong messages appear. Message id 70 shows up as reply 15, and message id 84 shows up as reply 29.

  4. If you go to a L or M value that exceeds the min or max id of the messages in the topic, such as http://stovell.noip.me/~jon/dev/index.php?topic=53.0.L10 or http://stovell.noip.me/~jon/dev/index.php?topic=53.0.M1000, you will get an "Undefined index: 0" error and no messages will be displayed.

albertlast:
I'm not sure if you understand the "concept" behind this correctly,

ther are three parts of information,
the topic, the page and msg_id
53.15.M68

page and msg_id are the same only different the information is displayed.
When i came back to you example
53.15.M68 = good because 15 and msg_id 68 are the same
53.15.M67
53.15.M69 are wrong because 15 means msg_id 68 but you send different msg_id.

To make it short M69 and M67 are not valid parameter.

about 4. we talk about message_id (from the message table) so the value can be
between 1 - max int of pk field

Sesquipedalian:
I understand that. My point is that this method includes nothing to detect and fix incorrect message id values, and it needs to. It is not enough to say, "That's a wrong parameter, don't do that." For the software to work reliably, incorrect parameters like these need to be detected and corrected. Otherwise errors are inevitable and will happen frequently.

albertlast:
Why should be 10 or 100000 incorrect?
L or M say only the message id needs to high oder lower not spezial one.

Sesquipedalian:
> Why should be 10 or 100000 incorrect?

Because values outside the range of the actual message ids in the topic generate an error and display no messages.

> L or M say only the message id needs to high oder lower not spezial one.

And therein lies the problem. Because there are no checks to see if the value is is a special one, it is trivial to pass an incorrect value and end up retrieving the wrong range of messages or even generating an error.

Put another way, the problem with this method is that it only works by trusting the URL parameter to be correct. But we can *never* trust user input to be correct. As I have demonstrated in my examples above, it is trivial to produce wrong behaviour and generate errors simply by editing the URL in the browser's address bar. If this method is going to be robust and reliable, it cannot simply trust the incoming URL parameters. Instead it must somehow protect itself against incorrect values. 

The goal here needs to be that if I am on page 2 of a topic, I will always be shown the messages that actually belong on page 2 of that topic. If it is possible by altering the URL parameters to be on page 2 and yet see messages that belong on page 1 or page 3, then the system is broken.

Sesquipedalian:
Besides editing URLs in the browser's address bar, I just thought of another way this problem could arise. If a user saves a link to, e.g., http://stovell.noip.me/~jon/dev/index.php?topic=53.15.M68, and then the number of visible messages in the topic changes for some reason (e.g messages become approved/unapproved or deleted, a moderator splits the topic, etc., etc.), the formerly working link will now produce wrong behaviour.

albertlast:
You try to create issue wher no is,
when the message id is out of range the user see nothing,
nothing is okay,
we can extend here the error handling that no error log is written but ther is no check for message_id needed,
only a check on the query how return the value to do the right thing when empty.

we don't trust the user input that is the reason that i already place some basic checks and
you try to extend them on some right scenario and some useless secenario like this here.

ther are common cases like the user jump from page to page and uncommon like you mention.
The common had the higher prio and not special of the special case.

When a post is deletet the old version didn't help you too,
you could try to access page two in a topic wher no page exists anymore.

And i don't understand why you not awnser my question...

> the clean up version could go in two direction,
extend the constructPageIndex about a test mode (would i realize by extending the options array) or
create a new function (checkStart) in sub with the test stuff lett it call from constructPageIndex
and display.

Sesquipedalian:
> If you go to a L or M value that exceeds the min or max id of the messages in the topic, such as http://stovell.noip.me/~jon/dev/index.php?topic=53.0.L10 or http://stovell.noip.me/~jon/dev/index.php?topic=53.0.M1000, you will get an "Undefined index: 0" error and no messages will be displayed.

Latest commit solves this problem. 👍 

Other problems in https://github.com/SimpleMachines/SMF2.1/pull/4198#issuecomment-315521247 remain.

albertlast:
Ther are no bug left in https://github.com/SimpleMachines/SMF2.1/pull/4198#issuecomment-315521247
this only when someone does wrong (manually),
than wrong things are happen.

And this wrong happen doesn't hurt smf and the security is not in denger.

Sesquipedalian:
> this only when someone does wrong (manually),
than wrong things are happen.

No. As I explained earlier, if a user saved a bookmark to a URL that worked correctly at the time he saved it, but then a moderator deleted some of the messages, or merged another topic into the bookmarked topic, or anything like that, the bookmarked URL would now produce unexpected results. The user did nothing wrong. The moderator did nothing wrong. The blame for the incorrect results lies entirely with the code.

albertlast:
Yeah when you bookmark a board and
the domain changed than you lost too.

So it's part of the internet that things get lost.

Sesquipedalian:
This is not the same type of problem. You are describing 404 errors when things cannot be found. The problem here is that the correct information is available to be found, but the server retrieves and displays different information instead because the URL query is not being handled well.

albertlast:
the provided information are wrong,
so they get a wrong result...
The last pr fix the issue when a msg got deleted.

Sesquipedalian:
> the provided information are wrong,
so they get a wrong result...

Again, it is not enough to say, "That's a wrong parameter, don't do that." By allowing the URL to specify both a 'start' parameter and a 'page_id' parameter, this code has created a situation where we can end up presenting any arbitrary post in the thread as if it were any other arbitrary post in the thread. Going to http://stovell.noip.me/%7Ejon/dev/index.php?topic=53.30.M53, for example, will present the original post in the topic as if it were reply 30. That's not acceptable, nor is it acceptable to blame the user for supplying conflicting URL parameters. The code needs to detect and fix conflicting URL parameters like this, or else the code is not doing its job properly.

When you first submitted the code for this new seek method, you described it as a way to improve performance while maintaining backward compatibility with the existing method. But at this point it has become clear that it does not maintain backward compatibility. Your response regarding this core problem is to insist that users everywhere must actively monitor and manage the obscure URL parameters of every page they visit if they want to avoid getting misleading results. Requiring that would extremely hostile to end users, to the point of making SMF virtually unusable for the average person. The support staff would be inundated with complaints and bug reports, and rightly so, because SMF would be behaving unpredictably for reasons that average users could not understand. In short, you are now asking for reliable functionality to be sacrificed for the sake of performance.

So the way I see it, there are two options for moving forward:

1. You make changes to how the seek method works so that if the user goes to page 2 of a topic, she will always be shown the messages that belong on page 2 of that topic, *even if $_REQUEST['page_id'] has an incorrect value*.

2. We remove the seek method from SMF.

If you want to pursue option 1, I just need to know how much time you think it will take to do so. If you think you can do it quickly, fine. If you think it will take you a while, then I will temporarily remove the existing version of the seek method code, and await your new, improved version. 

If you don't want to pursue option 1, then I'll implement option 2.

albertlast:
Again a missunderstanding from you side.

Why you believe that the user want to page 2?
He want ther where the message_id 30 is,
fix your paging that it know page 1 is.

Any way i got a new idea, how to implement this,
but i don't understand how the approvel stuff is working.

What happen when you got 10 messages in a topic how needs approval?
Ther shown so long they are not approved,
what happen when the 5. is approved and the other not.
what happen when a 11. get approved and the 10 messages before not.
how see what?

Sesquipedalian:
> Why you believe that the user want to page 2?
He want ther where the message_id 30 is,
fix your paging that it know page 1 is.

`?topic=53.30` means "Show me whatever page contains the 30th message in topic 53" (not "Show me the message whose message_id is 30 in topic 53"). That's what the `$_REQUEST['start']` parameter is used for in the `Display()` function, under both the original method and your seek method. The page index, reply numbering, etc., are all set up based on that value.

> Any way i got a new idea, how to implement this

Okay, great. Then I will merge #4197 for now while you work on your new approach. :)

> but i don't understand how the approvel stuff is working.
What happen when you got 10 messages in a topic how needs approval?
Ther shown so long they are not approved,
what happen when the 5. is approved and the other not.
what happen when a 11. get approved and the 10 messages before not.
how see what?

That gets figured out a bit higher up in `Display()`, in lines 212–238. The code there figures out how many messages there really are, how many are approved, how many are unapproved, and how many the user can actually see. In general, the last of these is the approved posts + their own unapproved posts (if any) + any posts they are allowed to approve (if any).

albertlast:
Idea is that i safe for commone pages the first id_msg in topic table in a new field(page_index) where the information are safed in json.
So when a user came want to page 2 which is 30 (when we default 15 page limit),
than we can look in the topic table of the field and get the message id of page 2.

Everytime the post is deletet or added,
we can update this information in the topic page_index.

So when a user access index.php?topic=53.30 we can use seek and
when he came with index.php?topic=53.32 we know that we don't had to seek.

In my eyes this only work if the topic is free of approval stuff,
which should be easy to detect based on line 217
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Display.php#L217

@Sesquipedalian any ideas/complains?

Sesquipedalian:
Well, if you were going to do that, you might as well just store an JSON array in the topics table that tracked a complete list of the message ids and then just use that info to skip searching for rows in the message table entirely. (But I'm not recommending that, just musing about it.)

Rather than storing paging info in the database or in URL parameters, you might want to consider storing it in `$_SESSION`. Since your original goal here was to make it faster to load the next or previous page of posts when the user is viewing a topic, the relevant information for this purpose is really a matter of what the user is currently looking at, rather than anything intrinsic to the topic itself. That makes `$_SESSION` the best place to track such data.

Sesquipedalian:
Maybe something like this could work:

1. If the user is viewing a topic, you keep track in `$_SESSION` of the following paging info about the current page: (1) the topic id (e.g. 53), (2) the reply numbers of the first and last messages shown on the current page (e.g. 15 and 29), and (3) the message_id numbers of the first and last messages shown on the current page (e.g. 69 and 83). 
2. If the next page the user loads is also viewing a topic, you first check to see if it is the same topic id as was stored in `$_SESSION` from last time.
   - If it is the same topic, you then 
        1. check to see if the current value of `$_REQUEST['start']` is not more than `$context['messages_per_page']` less than the stored reply number of the first message from the last page or not more than `$context['messages_per_page']` more than the stored reply number of the last message from the last page. If so, then you know that the user is now on the previous or next page of the topic.
        2. Because you know that, you can then use the seek method to get the batch of messages to display.
   - If any of the tests above turn out to be false, then you know that the user did not simply move to the previous or next page of the topic, and so you fall back to the normal method for finding the messages to display.

PR #4199, opened 2017-07-15
Sanitizes $_REQUEST['start'] for topic display in QueryString.php

Sesquipedalian:
Plugs a potential security hole by making sure that `$_REQUEST['start']` is always a clean and expected value before Display.php ever even sees it. This ensures that any random cruft (or evil SQL appended by would-be hackers) is dropped before before the start value is passed to the database abstraction layer.

@albertlast will probably want to add something here to sanitize `$_REQUEST['page_id']` according to his expected values for that URL parameter.

### DIFF
--- a/Sources/QueryString.php
+++ b/Sources/QueryString.php
@@ -195,10 +195,38 @@ function cleanRequest()
 			list ($_REQUEST['topic'], $_REQUEST['start'], $_REQUEST['page_id']) = explode('.', $_REQUEST['topic'] . '.');
 		}
 
-		$topic = (int) $_REQUEST['topic'];
+		// Topic should always be an integer
+		$topic = $_GET['topic'] = $_REQUEST['topic'] = (int) $_REQUEST['topic'];
 
-		// Now make sure the online log gets the right number.
-		$_GET['topic'] = $topic;
+		// Start could be a lot of things...
+		// ... a simple number ...
+		if (is_numeric($_REQUEST['start']))
+		{
+			$_REQUEST['start'] = (int) $_REQUEST['start'];
+		}
+		// ... or a specific message ...
+		elseif (strpos($_REQUEST['start'], 'msg') === 0)
+		{
+			$virtual_msg = (int) substr($_REQUEST['start'], 3);
+			$_REQUEST['start'] = $virtual_msg === 0 ? 0 : 'msg' . $virtual_msg;
+		}
+		// ... or whatever is new ...
+		elseif (strpos($_REQUEST['start'], 'new') === 0)
+		{
+			$_REQUEST['start'] = 'new';
+		}
+		// ... or since a certain time ...
+		elseif (strpos($_REQUEST['start'], 'from') === 0)
+		{
+			$timestamp = (int) substr($_REQUEST['start'], 4);
+			$_REQUEST['start'] = $timestamp === 0 ? 0 : 'from' . $timestamp;
+		}
+		// ... or something invalid, in which case we reset it to 0.
+		else
+		{
+			$_REQUEST['start'] = 0;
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Would this work to sanitize `$_REQUEST['page_id']` according to your specifications, @albertlast?
```
if (isset($_REQUEST['page_id']))
{
	$page_id_firstchar = substr($_REQUEST['page_id'], 0, 1);
	if (in_array($page_id_firstchar, array('L', 'M')))
		$_REQUEST['page_id'] = $page_id_firstchar . (int) substr($_REQUEST['page_id'], 1);
	else
		unset($_REQUEST['page_id']);
}
```

If so, I could add another commit to this PR to add that.

jdarwood007:
Please remember that is_numeric is not just whole numbers and could contain things like exponents for example.  Those would fail in queries as well.  It also matches decimals.

albertlast:
The page_id is my eye already checked in here: https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Display.php#L827
maybe a int check is missing (like i say in my eyes db layer had to his job here) but this would be the right place to extend the checks.

Sesquipedalian:
> Please remember that is_numeric is not just whole numbers and could contain things like exponents for example. Those would fail in queries as well. It also matches decimals.

That's why I cast it as an integer in that case. If someone tries `?topic=12.+0123.45e6`, the value of `$_REQUEST['start']` will end up as `123`.  If someone tries `?topic=12.0x539`, the value of `$_REQUEST['start']` will end up as `0`.  If someone tries `?topic=12.1337e0`, the value of `$_REQUEST['start']` will end up as `1337`.

Sesquipedalian:
> The page_id is my eye already checked in here: https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Display.php#L827
maybe a int check is missing (like i say in my eyes db layer had to his job here) but this would be the right place to extend the checks.

There you are *evaluating* the page_id and making decisions based on it. What I am talking about is simply *sanitizing* it so that when you evaluate it you can trust that it doesn't contain any unpleasant surprises.

albertlast:
Can you please look in to the code?
I check if the first char is m or l,
when not the whole logic is skipped.
And i check on db layer if $page_id is a int.

This the same logic as you use and had the same idea to protect from malformed data.

edit: okay i missing setting start_char at null when no m or l is provided,
but this nothing which can not be added.

Sesquipedalian:
The whole point of QueryString.php's `cleanRequest()` function is to clean the `$_REQUEST` variables. This is the correct place for that to happen. So, `$_REQUEST['page_id']` should be sanitized here, and then evaluated in Display.php. That's the way SMF is designed.

Sesquipedalian:
Since our current debate is a philosophical one about where the initial data sanitization should happen, rather than about whether my proposed code to sanitize `$_REQUEST['page_id']` will work, I have gone ahead and added it to this PR.

albertlast:
Would be than not better to throw the stuff in $context,
after the checks.

Sesquipedalian:
> Would be than not better to throw the stuff in $context,
after the checks.

Perhaps, yes. But that would be a big change to how SMF works in general. It's probably something to consider for SMF 3.0 rather than 2.1.

jdarwood007:
The stuff cleaned in QueryString.php is the initial load up stuff that can cross multiple ares of the code base (as $board, $topic, etc are).  In the future, it should contain a proper handler for all input to be cleaned.

If its input being cleaned in a function used across multiple sections, that function should clean it.  If its only used in a specific area, we should clean it before using it in that area.

colinschoen (approved):
LGTM

PR #4201, opened 2017-07-16
Raise min PG Version to 9.2

albertlast:
I mean we got the discussion last year,
let's bring up the important points:
- the rule for me for smf is the oldest supported version - 1 -> 9.2 get eol sep 2017
- PG get ~once a year a new version when we release smf 2.1 we need to support the existing version (5 versions) + the first eol = 6 additionaly smf 2.1 will stay after release many year (once per year a new version) which get us for examp 5 year runtime 5 additional version of which we should support 
6 + 5 =11 to keep the support "easy" we should not go lower as one below supported
- i got the problem that me as mostly the main "maintainer" for pg support is not able to download/run eol version (i need to look in the documentation if something is supported)

here you find the supported/eol information: https://www.postgresql.org/support/versioning/

Good things:
- range datatypes (the start and the end are one field) -> posible improvment in calendar area https://wiki.postgresql.org/wiki/What%27s_new_in_PostgreSQL_9.2#Range_Types
- first tier of json support https://wiki.postgresql.org/wiki/What%27s_new_in_PostgreSQL_9.2#JSON_datatype

### DIFF
--- a/other/install.php
+++ b/other/install.php
@@ -51,7 +51,7 @@ $databases = array(
 	),
 	'postgresql' => array(
 		'name' => 'PostgreSQL',
-		'version' => '9.1',
+		'version' => '9.2',
 		'function_check' => 'pg_connect',
 		'version_check' => '$request = pg_query(\'SELECT version()\'); list ($version) = pg_fetch_row($request); list($pgl, $version) = explode(" ", $version); return $version;',
 		'supported' => function_exists('pg_connect'),
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -37,7 +37,7 @@ $databases = array(
 	),
 	'postgresql' => array(
 		'name' => 'PostgreSQL',
-		'version' => '9.1',
+		'version' => '9.2',
 		'version_check' => '$version = pg_version(); return $version[\'client\'];',
 		'always_has_db' => true,
 	),


tinoest:
I know I argued the other way before, but I'd raise it to 9.4 now as most distributions have that as their minimal version. Or 9.1, for some reason 9.2/9.3 isn't as common.

albertlast:
yeah i would like to raise it to 9.4

colinschoen:
Seems reasonable to me. Any concerns from others?

Sesquipedalian:
> Any concerns from others?

I have no opinion on this matter.

albertlast:
Could we go directly to 9.4?

jdarwood007:
What it is the general version being supported by hosts?  As long as we target that fairly well, we are good to raise the version to what the users can be expected to have.

On the other hand, if there is no reason to upgrade (no real features improvements we make use of yet), then there isn't a reason to push.

albertlast:
Like i said in start comment,
the reason like that "we no need new features so can keep at the existing version" wrong in my eyes
because we had to support every new version and so the amount of supported version get higher every year.

Oldiesmann:
The latest available Postgres version on CentOS 7.3 is 9.2.7, so raising it to 9.2 should be OK.

albertlast:
okay then merge this pr.

PR #4202, opened 2017-07-16
More speed calendar

albertlast:
For doesn't look that some one try to get the stuff better,
i try again.
The result is this pr,
which got the same/better runtime as the database solution.
~800 ms -> 400ms

One thing i don't understand from where and why we use %l instead of %I 
this small L didn't exists in strftime doc: http://php.net/manual/de/function.strftime.php

so i place a replace...

maybe @Sesquipedalian can explain this.

left old behavior right new bevhaior
grafik

### DIFF
--- a/Sources/Subs-Calendar.php
+++ b/Sources/Subs-Calendar.php
@@ -1591,14 +1591,12 @@ function buildEventDatetimes($row)
 	$end['iso_gmdate'] = gmdate('c', $end['timestamp']);
 
 	// Strings showing the datetimes in the user's preferred format, relative to the user's time zone
-	$test = timeformat($start['timestamp'], $date_format . ' § ' . $time_format);
-	$test_array = explode(' § ', timeformat($start['timestamp'], $date_format . ' § ' . $time_format));
-	list($start['date_local'], $start['time_local']) = explode('§', timeformat($start['timestamp'], $date_format . '§' . $time_format));
-	list($end['date_local'], $end['time_local']) = explode('||',timeformat($end['timestamp'], $date_format . '||' . $time_format));
+	list($start['date_local'], $start['time_local']) = explode(' § ', timeformat($start['timestamp'], $date_format . ' § ' . $time_format));
+	list($end['date_local'], $end['time_local']) = explode(' § ',timeformat($end['timestamp'], $date_format . ' § ' . $time_format));
 
 	// Strings showing the datetimes in the user's preferred format, relative to the event's time zone
-	list($start['date_orig'], $start['time_orig']) = explode('||',timeformat($start['timestamp'], $date_format . '||' . $time_format, 'none'));
-	list($end['date_orig'], $end['time_orig']) = explode('||',timeformat($end['timestamp'], $date_format . '' . $time_format, 'none'));
+	list($start['date_orig'], $start['time_orig']) = explode(' § ',timeformat($start['timestamp'], $date_format . ' § ' . $time_format, 'none'));
+	list($end['date_orig'], $end['time_orig']) = explode(' § ',timeformat($end['timestamp'], $date_format . ' § ' . $time_format, 'none'));
 
 	// The time zone identifier (e.g. 'Europe/London') and abbreviation (e.g. 'GMT')
 	$tz = date_format($start_object, 'e');
--- a/Sources/Subs-Calendar.php
+++ b/Sources/Subs-Calendar.php
@@ -130,7 +130,11 @@ function getBirthdayRange($low_date, $high_date)
 function getEventRange($low_date, $high_date, $use_permissions = true)
 {
 	global $scripturl, $modSettings, $user_info, $smcFunc, $context, $sourcedir;
+	static $timezone_array = array();
 	require_once($sourcedir . '/Subs.php');
+	
+	if (empty($timezone_array['default']))
+		$timezone_array['default'] = timezone_open(date_default_timezone_get());
 
 	$low_object = date_create($low_date);
 	$high_object = date_create($high_date);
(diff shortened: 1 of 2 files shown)


Sources/Subs-Calendar.php
```diff
@@ -1578,16 +1589,12 @@ function buildEventDatetimes($row)
 	$end['iso_gmdate'] = gmdate('c', $end['timestamp']);
 
 	// Strings showing the datetimes in the user's preferred format, relative to the user's time zone
-	$start['date_local'] = timeformat($start['timestamp'], $date_format);
-	$start['time_local'] = timeformat($start['timestamp'], $time_format);
-	$end['date_local'] = timeformat($end['timestamp'], $date_format);
-	$end['time_local'] = timeformat($end['timestamp'], $time_format);
+	list($start['date_local'], $start['time_local']) = explode(' § ', timeformat($start['timestamp'], $date_format . ' § ' . $time_format));
```
Sesquipedalian, line 1592: Clever idea! 👍

Sources/Subs-Calendar.php
```diff
@@ -1578,16 +1589,12 @@ function buildEventDatetimes($row)
...
+	list($end['date_local'], $end['time_local']) = explode(' § ',timeformat($end['timestamp'], $date_format . ' § ' . $time_format));
```
Sesquipedalian: Nit: space missing before `timeformat` here and elsewhere.

Sources/Subs-Calendar.php
```diff
@@ -130,7 +130,11 @@ function getBirthdayRange($low_date, $high_date)
 function getEventRange($low_date, $high_date, $use_permissions = true)
 {
 	global $scripturl, $modSettings, $user_info, $smcFunc, $context, $sourcedir;
+	static $timezone_array = array();
```
Sesquipedalian, line 133: This is a good approach. 👍

Sources/Subs.php
```diff
@@ -829,6 +829,9 @@ function timeformat($log_time, $show_today = true, $offset_type = false, $proces
 	if ($context['server']['is_windows'] && strpos($str, '%e') !== false)
 		$str = str_replace('%e', ltrim(strftime('%d', $time), '0'), $str);
 
+	// fix unkown %l with %I
```
Sesquipedalian: `%l` is part of the strftrime format specifications. The German version of the strftrime documentation is incomplete. See the English version for a description of `%l` in comparison to `%I`: http://php.net/manual/en/function.strftime.php

Sesquipedalian (commented):
The changes in Subs-Calendar.php are great. 👍 

Regarding `%l`, the problem is just that the German version of PHP's documentation is missing some info. 

Please revert the changes in Subs.php and add the missing spaces. Then this PR can be merged. :)

Sesquipedalian:
> left old behavior right new bevhaior

I am curious about why `buildEventDateTimes()` is being called 715 times in your tests. Are there 715 events and birthdays shown on the page you are testing? Because it should only be called once per event.

colinschoen (approved):
LGTM

albertlast:
My test board try to emulate a mid - big tier smf board,
so i got 2k user with birthday +
i generated "some" events a cross the year.

in this case i got ~ 500 normal events,
~200 birthday across three months.

Sources/Subs.php
albertlast: well i forgot to mention my motivation for this "fix",
when i keep the %L(small) strftime return false instead of a string mit date and time:

```
<?php
echo 'test';
$test = strftime('%l:%M am');
var_dump($test);
```
->
```
test

D:\Programme\xampp2\htdocs\test.php:4:boolean false
```

when i use %i(high) 

```
<?php
echo 'test';
$test = strftime('%I:%M am');
var_dump($test);
```

```
test

D:\Programme\xampp2\htdocs\test.php:4:string '05:42 am' (length=8)
```


So on my side strftime doesn't work with %L ...

php 7.1.6

Sesquipedalian: I wonder if that is due to a difference in locale. Here is my text code and results:
```
<?php
date_default_timezone_set('America/Edmonton');

$currentLocal = setlocale(LC_ALL, 0); // Get name of current locale

echo $currentLocal, PHP_EOL;
$test = strftime('%l:%M am');
var_dump($test);
```
Result:
```
C
string(8) "12:50 am"
```
What locale does your machine report?

Sesquipedalian:
> My test board try to emulate a mid - big tier smf board,
> so i got 2k user with birthday +
> i generated "some" events a cross the year.
>
> in this case i got ~ 500 normal events,
> ~200 birthday across three months.

Okay, thanks for clarifying for me. I was worried for a bit there that the function was somehow being called more times than in should have been.

Sources/Subs.php
albertlast:
```
<?php
$currentLocal = setlocale(LC_ALL, 0); // Get name of current locale

echo $currentLocal, PHP_EOL;
$test = strftime('%l:%M am');
var_dump($test);

date_default_timezone_set('America/Edmonton');

$currentLocal = setlocale(LC_ALL, 0); // Get name of current locale

echo $currentLocal, PHP_EOL;
$test = strftime('%l:%M am');
var_dump($test);
```

Result:
```
C

D:\Programme\xampp2\htdocs\test.php:6:boolean false

C

D:\Programme\xampp2\htdocs\test.php:14:boolean false
```

Sesquipedalian: Hm. According to PHP's strftime documentation, supported formats vary from system to system. It all depends on which ones were included in your system's C library:

> Not all conversion specifiers may be supported by your C library, in which case they will not be supported by PHP's strftime().

Additionally, different operating systems deal with unsupported formats differently:

> As the output is dependent upon the underlying C library, some conversion specifiers are not supported. On Windows, supplying unknown conversion specifiers will result in 5 E_WARNING messages and return FALSE. On other operating systems you may not get any E_WARNING messages and the output may contain the conversion specifiers unconverted.

In light of this, we need to come up with something more robust to deal with unsupported formats, because it can vary quite a lot from system to system and we can't assume anything. I've started working on something to do that now, and I will submit it in a PR soon. In the meantime, if you could take this bit out of this PR before we merge it, that'd make it easier for me to avoid merge conflicts when putting together my PR.

Sesquipedalian (approved):

PR #4206, opened 2017-07-17
Handles unsupported strftime formats more robustly

Sesquipedalian:
This seems to work well on my *nix systems. I don't have Windows, though, so if someone (maybe @albertlast?) could test it there, that'd be good.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -792,7 +792,7 @@ function timeformat($log_time, $show_today = true, $offset_type = false, $proces
 			// Time and Date Stamps
 			'c' => '%F %T', 'D' => '%m/%d/%y', 'F' => '%Y-%m-%d', 's' => '&#37;s', 'x' => '%F',
 			// Miscellaneous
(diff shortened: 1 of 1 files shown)


jdarwood007:
Since this doesn't even need to be done on every page load.  Could this be cached up for a day?

Sesquipedalian:
> Since this doesn't even need to be done on every page load. Could this be cached up for a day?

Good idea.

albertlast:
here some time about Page created:
with cache 1sec 
without cache 1.4 sec

left cache right without
grafik


The existing version (with replace otherwise doesn't work for me): 0,37 sec
grafik

in my eyes is the using of caching here realy wrong.

Sesquipedalian:
What do your tests show now, @albertlast?

albertlast:
Looks perfect from runtime:
left your version right my old
grafik
And i can't see any bad behavior of smf.

Sesquipedalian:
Excellent.

PR #4210, opened 2017-07-19
Avoids repeating attempts to download TLD list on isolated servers

Sesquipedalian:
Previously, if the server couldn't retrieve https://data.iana.org/TLD/tlds-alpha-by-domain.txt, another attempt would be scheduled in the background tasks queue. As a result, isolated servers or ones that couldn't reach iana.org for whatever reason could end up in a loop of trying and failing to update the value of `$modSettings['tld_regex']`. With this change, we give up after a failed update attempt and don't try again until the next regularly scheduled weekly update.


colinschoen (approved):
LGTM (fun)

PR #4211, opened 2017-07-19
Language fixes

Gwenwyfar:
Just fixing a few strings. The `<br>` tag was removed as it has no place in the middle of that phrase. In search, you can select multiple boards or all, not just one, so I changed it to plural.


colinschoen (approved):
LGTM

sbulen:
THANK YOU....

PR #4213, opened 2017-07-22
More phpDoc fixes

NanoSector:
Signed-off-by: Yoshi2889 <rick.2889@gmail.com>

### DIFF
--- a/Sources/Subs-Db-postgresql.php
+++ b/Sources/Subs-Db-postgresql.php
@@ -499,7 +499,12 @@ function smf_db_query($identifier, $db_string, $db_values = array(), $connection
 
 /**
  * affected_rows
- * @param resource $connection
+ *
+ * !!! Not sure what this is supposed to do.
+ * @param mixed $result
+ *
+ * @return int
+ *
  */
 function smf_db_affected_rows($result = null)
 {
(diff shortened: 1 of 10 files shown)


Sources/Subs-Db-postgresql.php
```diff
@@ -499,7 +499,12 @@ function smf_db_query($identifier, $db_string, $db_values = array(), $connection
 
 /**
  * affected_rows
- * @param resource $connection
+ *
+ * !!! Not sure what this is supposed to do.
```
albertlast: Return the affected rows of a query.
see http://php.net/manual/en/function.pg-affected-rows.php

Sources/Subs-Db-postgresql.php
```diff
@@ -518,7 +523,7 @@ function smf_db_affected_rows($result = null)
  *
  * @param string $table The table (only used for Postgres)
  * @param string $field = null The specific field (not used here)
- * @param resource $connection = null The connection (if null, $db_connection is used)
+ * @param resource $connection = null The connection (if null, $db_connection is used) (not used here?)
```
albertlast: The mysql version can use it:
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs-Db-mysql.php#L541

albertlast (changes requested):

colinschoen (approved):
Pending albertlast's requested changes. LGTM

PR #4219, opened 2017-07-30
Fixes #4218

### LINKED ISSUE #4218: SSL Assumptions
I believe I have found two coding issues/assumptions both possibly impacting SSL.  

(1) The assumption that if force_ssl is set, that SSL is enabled/active.  This is not true, as force_ssl set to '0' means it is disabled.  Checking is_set(force_ssl) is insufficient.  Since force_ssl is new to 2.1, this only impacts 2.1 code.

(2) $_SERVER['HTTPS'] is used in several places, and what I've read indicates it is used slightly differently between IIS and *nix.  In *nix, it is either 'on' or it's not set (never 'off').  Some IIS servers have it 'on' or 'off'.  Similar thing, we cannot assume it is 'off' when it is not set; that will work for *nix but not all IIS users.  

The $_SERVER['HTTPS'] assumption possibly impacts 2.0 IIS users.  

I'm going to audit usages of both & provide some examples.

sbulen:
Remove unnecessary/confusing check.

### DIFF
--- a/Sources/ManageServer.php
+++ b/Sources/ManageServer.php
@@ -301,7 +301,7 @@ function ModifyCookieSettings($return_config = false)
 		array('localCookies', $txt['localCookies'], 'db', 'check', false, 'localCookies'),
 		array('globalCookies', $txt['globalCookies'], 'db', 'check', false, 'globalCookies'),
 		array('globalCookiesDomain', $txt['globalCookiesDomain'], 'db', 'text', false, 'globalCookiesDomain'),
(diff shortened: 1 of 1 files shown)


jdarwood007:
While it seems unnecessary, but there could be server configurations sending 1 instead of on.

sbulen:
None of the info I've read includes a check for 1/0.  They all account for Unix being on or unset, and (one flavor of) IIS being on or off.   

There are at least a dozen similar checks in SMF, this is the outlier.   (If a check for 1/0 were important, we'd have to replicate it everywhere else...)

Good news is I don't think this hurts anything...   Pretty sure it will always return false, so the OR makes the logic dependent on the other variable.

Oldiesmann:
If we're removing that check then we can just change the check to `strtolower($_SERVER['HTTPS']) != 'on'` instead of the more confusing `!(strtolower($_SERVER['HTTPS']) == 'on')`

sbulen:
Done.

PR #4223, opened 2017-08-01
Update some leftover tags

Gwenwyfar:
This updates some `<br />`, `<hr />` and `<b>` tags still lying around.
Someone may want to check for the usage of these in some functions, this only updates the templates/other output HTML and language files.


albertlast:
It's not needed to mention in the head or somewhere else,
that you side now html5?
By side what is with the mods how use not html5?

Gwenwyfar:
I'm not sure I understand what you're saying, @albertlast, but I hope you're happy now with the new title? Honestly I don't see what difference it makes, I'm just being clear on what was modified here.

albertlast:
My question got awnser in the irc.

PR #4225, opened 2017-08-03
Fixes #4011

### LINKED ISSUE #4011: Set HTTPS Board URL on Install
When installing SMF http:// is set as the protocol even if force SSL is enabled in the installer. We should set the board url to be https:// if that is where the installer is loaded from or force SSL is enabled.

sbulen:
If user selects to use SSL, ensures all URLs are https://.   It does this by ensuring $boardurl is aligned with the SSL setting before setting any others, e.g., themes - everything else keys off of $boardurl.

Also ensures you cannot select SSL if you have no cert installed.   

Further, ensures you MUST use SSL if you have an http:// => https:// redirect active.

If folks like this approach, I'll do similar checks for the upgrader next.


sbulen:
Hold on on this for a bit, found an issue...  Will sub a PR in a day or two...

sbulen:
Fixed.  Ended up doing it & retesting this morning.

sbulen:
The upgrader's needs were much simpler.  So...  The two routines added to sub.php may in fact only be needed by the installer.  It may make sense to move them there... 

Unless...    

Folks think it makes sense to add similar edits to the force_ssl setting in the admin control panel.

 I'm open to input.

PR #4226, opened 2017-08-05
Seek for messages in topics

albertlast:
This is the next try to implement seek for pages in topics.
Handle four cases:
next page
before page
current page (when someone push refresh)
start page (it's mostly the same as current page)

@Sesquipedalian can you please look for issue of this version?

### DIFF
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -876,15 +876,15 @@ function Display()
 		$start_char = null;
 
 	$limit = $context['messages_per_page'];
-	
+
 	$messages = array();
 	$all_posters = array();
-	
+
 	if (isset($start_char))
 	{
 		$firstIndex = 0;
 
-		if ($start_char === 'M' or $start_char === 'S' or $start_char === 'C')
+		if ($start_char === 'M' or $start_char === 'C')
 		{
 			$ascending = true;
 			$page_operator = '>=';
@@ -927,18 +927,18 @@ function Display()
 		{
 			while ($row = $smcFunc['db_fetch_assoc']($request))
 			{
-				if ($row['id_msg'] != $page_id || $start_char === 'C')
+				// Check if the start msg is in our result
+				if ($row['id_msg'] == $page_id)
+					$found_msg = true;
+
+				// Skip the the start msg if we not in mode C
+				if ($start_char === 'C' || $row['id_msg'] != $page_id)
 				{
 					if (!empty($row['id_member']))
 						$all_posters[$row['id_msg']] = $row['id_member'];
 
 					$messages[] = $row['id_msg'];
-
-					if ($start_char === 'C' && $row['id_msg'] == $page_id) // find the start message of current page
-						$found_msg = true;
 				}
-				else
-					$found_msg = true; // find the start message of next or before page
 			}
 
 			// page_id not found? -> fallback
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -931,6 +931,7 @@ function Display()
 				{
 					if (!empty($row['id_member']))
 						$all_posters[$row['id_msg']] = $row['id_member'];
+
 					$messages[] = $row['id_msg'];
 
 					if ($start_char === 'C' && $row['id_msg'] == $page_id) // find the start message of current page
@@ -993,7 +994,7 @@ function Display()
 			$messages[] = $row['id_msg'];
 		}
 
-		// get the message in this order how they shown
+		// Sort the messages into the correct display order
 		if (!$ascending)
 			sort($messages);
 	}
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -875,7 +875,7 @@ function Display()
 	{
 		$firstIndex = 0;
 		
-		if ($start_char === 'M' or $start_char === 'S')
+		if ($start_char === 'M' or $start_char === 'S' or $start_char === 'C')
 		{
 			$ascending = true;
 			$page_operator = '>=';
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -842,45 +842,153 @@ function Display()
 		call_integration_hook('integrate_poll_buttons');
 	}
 
-	// Calculate the fastest way to get the messages!
-	$ascending = empty($options['view_newest_first']);
-	$start = $_REQUEST['start'];
-	$limit = $context['messages_per_page'];
-	$firstIndex = 0;
-	if ($start >= $context['total_visible_posts'] / 2 && $context['messages_per_page'] != -1)
+	// Check if we can use seek
+	if (isset($_SESSION['page_topic']) && isset($_SESSION['page_next_start']) && $_SESSION['page_topic'] == $topic && $_SESSION['page_next_start'] == $_REQUEST['start'])
 	{
-		$ascending = !$ascending;
-		$limit = $context['total_visible_posts'] <= $start + $limit ? $context['total_visible_posts'] - $start : $limit;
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Sorry, I got swamped with real life stuff to take care of. I will take a look at this as soon as I can. :)

albertlast:
When it take to long we could merge this pr and
let everyone test this.
in the worst case do a revert again.

Sources/Display.php
```diff
@@ -842,45 +842,163 @@ function Display()
...
+		$start_char = 'C';
+		$page_id = $_SESSION['page_first_id'];
+	}
+	elseif ($start == 0) //special case start page
+	{	// special case start page
+		$start_char = 'C';
+		$page_id = $context['topicinfo']['id_first_msg'];
+	}
+	else
+		$start_char = null;
```
Sesquipedalian: This would avoid checking the value of $_SESSION['page_topic'] more than once:
```
	// Check if we can use the seek method to speed things up
	if (isset($_SESSION['page_topic']) && $_SESSION['page_topic'] == $topic)
	{
		// User moved to the next page
		if (isset($_SESSION['page_next_start']) && $_SESSION['page_next_start'] == $start)
		{	
			$start_char = 'M'; 
			$page_id = $_SESSION['page_last_id'];
		}
		// User moved to the previous page
		elseif (isset($_SESSION['page_before_start']) && $_SESSION['page_before_start'] == $start)
		{	
			$start_char = 'L';
			$page_id = $_SESSION['page_first_id'];
		}
		// User refreshed the current page
		elseif (isset($_SESSION['page_current_start']) && $_SESSION['page_current_start'] == $start)
		{
			$start_char = 'C';
			$page_id = $_SESSION['page_first_id'];
		}
	}
	// Special case start page
	elseif ($start == 0)
	{
		$start_char = 'C';
		$page_id = $context['topicinfo']['id_first_msg'];
	}
	else
		$start_char = null;
```
Also, this would make the comments a bit clearer in English and would format them according to SMF's standard coding style. :)

Sources/Display.php
```diff
@@ -842,45 +842,163 @@ function Display()
...
+		if ($smcFunc['db_num_rows']($request) < 1)
+			unset($start_char);
+		else
+		{
+			while ($row = $smcFunc['db_fetch_assoc']($request))
+			{
+				if ($row['id_msg'] != $page_id || $start_char === 'C')
+				{
+					if (!empty($row['id_member']))
+						$all_posters[$row['id_msg']] = $row['id_member'];
```
Sesquipedalian: Nit: please add a line break after this to improve legibility.

Sources/Display.php
```diff
@@ -842,45 +842,163 @@ function Display()
...
+					$messages[] = $row['id_msg'];
+
+					if ($start_char === 'C' && $row['id_msg'] == $page_id) // find the start message of current page
+						$found_msg = true;
+				}
+				else
+					$found_msg = true; // find the start message of next or before page
```
Sesquipedalian: This else statement is probably a bug. It will set $found_msg to true in a number of cases where I doubt that is what you really intended. For example, if the first retrieved row from the database matched the previous if statement, then the rest of the retrieved rows probably won't match. The result is that $found_msg is probably *always* going to be true by the time you finish looping through all the retrieved rows.

Sources/Display.php
```diff
@@ -842,45 +842,163 @@ function Display()
...
+		);
+
+		while ($row = $smcFunc['db_fetch_assoc']($request))
+		{
+			if (!empty($row['id_member']))
+				$all_posters[$row['id_msg']] = $row['id_member'];
+			$messages[] = $row['id_msg'];
+		}
+
+		// get the message in this order how they shown
```
Sesquipedalian: Suggested: `// Sort the messages into the correct display order`

Sources/Display.php
```diff
@@ -842,45 +842,163 @@ function Display()
...
+		if (!$ascending)
+			sort($messages);
 	}
+
+	// Save the next start of the next page and the end of the page before
```
Sesquipedalian: Suggested: `// Remember the paging data for next time`

Sesquipedalian (changes requested):
Looks pretty great. I've made a few suggestions, and I found one spot where the logic to deal with unexpected results appears to be a bit buggy. I could not find a way to turn that into a problematic result, but you should still probably take a look at it.

Sources/Display.php
albertlast: The else cover this case:
you go the next or previous page and
the page_id was deleted/get lost some how.

Testcase: go to a thread with 3 pages
go to page 3, delete somehow the first msg on page three and call then page 2 than this
fallback get active.

To make it clear i describe now the same thing from coding side:
Common is that we need the amount of msg_id which is needed to show ->
when you want to show 15 msg = limit 15.
To cover this case described above we set limit + 1,
so the sql return 15 + (start page_id) we need this for check but we don't want
this deliver to the next logic part.
In short we don't want the start page_id in $messages.

albertlast:
Some how i didn't see all you comment at once,
so i created multiple commits for you comments.

I hope i cover all the things.

Sources/Display.php
Sesquipedalian: Okay, I think I see what you mean. In that case, you can probably simplify your code a bit by replacing this:
```
if ($row['id_msg'] != $page_id || $start_char === 'C')
{
	if (!empty($row['id_member']))
		$all_posters[$row['id_msg']] = $row['id_member'];

	$messages[] = $row['id_msg'];

	if ($start_char === 'C' && $row['id_msg'] == $page_id) // find the start message of current page
		$found_msg = true;
}
else
	$found_msg = true; // find the start message of next or before page
```
with this:
```
if ($row['id_msg'] == $page_id)
	$found_msg = true;

if ($start_char === 'C' || !$found_msg)
{
	if (!empty($row['id_member']))
		$all_posters[$row['id_msg']] = $row['id_member'];

	$messages[] = $row['id_msg'];
}
```

They are logically equivalent, but the second version is easier to understand.

Sources/Display.php
```diff
@@ -842,45 +842,171 @@ function Display()
...
+	
 	$messages = array();
 	$all_posters = array();
-	while ($row = $smcFunc['db_fetch_assoc']($request))
+	
+	if (isset($start_char))
+	{
+		$firstIndex = 0;
+
+		if ($start_char === 'M' or $start_char === 'S' or $start_char === 'C')
```
Sesquipedalian: Should `$start_char === 'S'` be `$start_char === 'L'` here?

Sources/Display.php
albertlast: When I read correctly (for me is this version harder to  read)
```
if ($row['id_msg'] == $page_id)
	$found_msg = true;

if ($start_char === 'C' || !$found_msg)
{
	if (!empty($row['id_member']))
		$all_posters[$row['id_msg']] = $row['id_member'];

	$messages[] = $row['id_msg'];
}
```
Doesn't skip the 16th dataset for the L and M case.
```
if ($row['id_msg'] == $page_id)
	$found_msg = true;

if ($start_char === 'C' || $row['id_msg'] != $page_id)
{
	if (!empty($row['id_member']))
		$all_posters[$row['id_msg']] = $row['id_member'];

	$messages[] = $row['id_msg'];
}
```
This would be logically equivalent in my eyes.

Sesquipedalian: Fair enough. :)

albertlast:
So can we merge this?

Sesquipedalian (approved):

Sesquipedalian:
Nice to see this go in. :)

PR #4227, opened 2017-08-06
Avoid redundant requires when integrate_autoload hook is used

MissAllSunday:
If a mod is using integrate_autoload allow it to set the first param in each $actionArray entry as false to avoid redundant calls to require_once(). Same behavior as Profile and Admin files.


colinschoen (approved):
LGTM

Issue #4224, opened 2017-08-02
"Enable basic HTML" tags not working


Gwenwyfar:
Some tags for the "enable basic HTML" feature are not working properly.

Here's a screenshot with the output for the following posts:
```
<b>Bold</b>
<em>Em</em>
<i>i</i>
<u>U</u>
<s>S</s>
<ins>ins</ins>
<del>del</del>
<a href="somelink">Somelink</a>
<img src="/upgrade-ver/Themes/default/images/smflogo.png" alt="Someimage" />
<pre>Pre</pre>
<blockquote>blockquote</blockquote>
<br />
<hr />
```

Second post (`strong` is not on the list, but it should be since `em` is). Both `<br/>` and `<br>` should work. both hrs are working fine.
```
<strong>Strong</strong>
<br>
<hr>
```
![tags](https://user-images.githubusercontent.com/28991853/28870661-83454abc-7757-11e7-991b-167e52a43885.png)

Edit: Adding http:// or https:// to the anchor or image tags works, but shouldn't it still be parsed as HTML on both cases?

illori:
the html has to be valid html 5 or we need to put in more effort to convert it to be valid html 5.

Gwenwyfar:
The html for the image tag is completely valid, and the link could also be valid without http://. Both of them accept relative paths.

Gwenwyfar:
Did a few more tests, and `mailto:` works, but `tel:` doesn't. `#id` also doesn't (but this latter is probably not going to get much use).

sbulen:
I'll start chasing a few of these down.

sbulen:
PR #4230 submitted, fixes hr, br & strong.

I've started looking into relative paths.  Current a & img code works fine on fully qualified URLs at the moment.  

I think tel: would be very valuable in a mobile context.  I'll look into that, too.  

Not sure #id is worth it, though...

Gwenwyfar:
I think for #id as long as it works at the end of a full URL it should be good enough (gave it a try with a full URL with a message ID but it returned another broken tag :P)

sbulen:
In plain link in SMF it work, right?   E.g., the following works in a post:
http://www.vguitarforums.com/smf/index.php?topic=20867.msg154930#msg154930

And in a post in SMF, this works:
`<a href="http://www.vguitarforums.com/smf/index.php?topic=20867.msg154930#msg154930">Link test 2</a>`

Gwenwyfar:
[...]

PR #4230, opened 2017-08-08
Fix some html tags - Issue #4224

sbulen:
This addresses br, hr & strong issues.


sbulen:
3 final fixes here:
(1) Fixed preg_match so you won't get lopsided [url]s.
(2) Fixed html mailto: - there was a conflict between 'auto url' logic and the html logic - now prevents 'auto url' logic from stepping on the html logic.
(3) Added tel: html element.  Works great on mobile.  Points to skype on desktops, if skype is there...

html anchor variants now supported are:
```
<a href="http(s)://somelink">Sometext</a>
<a href="ftp(s)://somelink">Sometext</a>
<a href="mailto:somelink">Sometext</a>
<a href="tel:somelink">Sometext</a>

```

PR #4236, opened 2017-08-12
Icons & images

Gwenwyfar:
Just polishing up the icons. The size of some was doubled so it doesn't look stretched on some screens with higher pixel density.


XinYenFon (approved):

Themes/default/css/index.css
```diff
@@ -446,19 +446,26 @@ a img {
...
 	overflow: hidden;
 	content: '';
 	vertical-align: middle;
 	margin: -5px 5px 0 5px;
-	opacity: 0.8;
+	border: 1px solid #c5c5c5;
+	border-radius: 3px;
+	box-shadow: 0 1px 2px rgba(0,0,0,.2),
+		0 1px 1px #fff inset,
+		0 -5px 4px rgba(0,0,0,.1) inset;
```
frandominguezl, line 458: Shouldn't this be 5px instead of -5px?

frandominguezl: Um, nevermind, it's okay, I forgot to add the final inset ^^

frandominguezl (approved):

PR #4237, opened 2017-08-14
Css

Gwenwyfar:
Just some more details fixed/improved. A lot of edits, but these are mostly minor. Just getting these in before we start having conflicts.


Themes/default/css/index.css
```diff
@@ -7,14 +7,12 @@ body {
...
-	margin: 0;
-	padding: 0;
 	display: flex;
 	flex-direction: column;
 	min-height: 100vh;
 }
-/* <p> is a possible exception for margins. Must test that one a bit. */
-p, ul, ol, li, dl, dd ,dt, fieldset, form {
+/* General reset */
+* {
```
XinYenFon, line 15: Well this is an important step, not that I'm against but requires some careful testing to be safe.

Themes/default/css/index.css
```diff
@@ -392,12 +385,12 @@ blockquote cite::before {
...
 }
 /* Styling for BBC tags */
-.bbc_link:link, .bbc_link:visited {
+.bbc_link {
 	border-bottom: 1px solid #A8B6CF;
 }
 .bbc_link:hover {
 	text-decoration: none;
-	border-bottom: 1px solid #346;
+	border-bottom-color: #346;
```
XinYenFon: changes like this requires consistency, either go for all in CSS or just revert it.

XinYenFon (changes requested):
Rest looks fine.

Themes/default/css/index.css
Gwenwyfar: Margin/padding should be fine, browser default ones are too inconsistent to be relied onto. Border-box could cause some harm, but at most it'll break somewhere that widths are used that conflict with padding, better to find those early and they're very easy to fix. I looked at a wide range of pages and didn't find any problems so far, other than the two already fixed.

Themes/default/css/index.css
Gwenwyfar: Let's leave it like that for now then.

XinYenFon (approved):

Themes/default/css/index.css
XinYenFon: Those type of padding issues can be fixed in RC period.

XinYenFon (approved):

frandominguezl (approved):

PR #4238, opened 2017-08-15
Fixing and optimize search

albertlast:
Create Valid Temp table, 
"drop table if exits",
removed db_error_skip, 
optimiz order for pg, 
force using fulltext index


albertlast:
this pr is done please merge.

PR #4243, opened 2017-08-16
Remove input_* classes

Gwenwyfar:
Should help make the the code cleaner, these classes have no real use on js/css (there is the `input[type="*"]` selector for them), it's just bloat. Only css uses them so this shouldn't affect anything else.

upgrade.php and install.php were not included.


XinYenFon (changes requested):
Upgrade.php & Install.php missing.

Gwenwyfar:
I've included upgrade.php and install.php. The classes there don't seem to be used at all (no css seems to be linked to them either), but I'm not 100% sure, so someone may want to confirm that.

PR #4245, opened 2017-08-18
Fixes #4011

### LINKED ISSUE #4011: Set HTTPS Board URL on Install

sbulen:
If user selects to use SSL, ensures all URLs are https://. It does this by ensuring $boardurl is aligned with the SSL setting before setting any others, e.g., themes - everything else keys off of $boardurl.

Also ensures you cannot select SSL if you have no cert installed.

Further, ensures you MUST use SSL if you have an http:// => https:// redirect active.

If folks think it makes sense to add similar edits to the force_ssl setting in the admin control panel, I'll do so.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -5989,12 +5989,13 @@ function build_regex($strings, $delim = null, $returnArray = false)
 	return $regex;
 }
 
-/*
-//*** Check if current domain has a cert
-// Adapted from: https://stackoverflow.com/questions/27689147/how-to-check-if-domain-has-ssl-certificate-or-not
-// * @param string $url to check, in $boardurl format (no trailing slash).
-*/ 
-function ssl_cert_found($url) {
+/**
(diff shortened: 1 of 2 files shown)


sbulen:
I had to revise & retest due to newly developed merge conflicts.  

FYI - I have tested:
  -  Windows & linux, both apache
  -  With certs & without
  -  With redirects & without 
  -  With forums in subdirectories & not

That said, I can't test everything...  E.g., I haven't tested with subdomains, or IIS.  

Any questions, let me know,

colinschoen (changes requested):
Can you fix the doc block formatting.

sbulen:
Done.  Yep, they were ugly.

colinschoen (approved):
LGTM

PR #4246, opened 2017-08-18
[PG] Drop Seq more nicely

albertlast:
Added if exists and another place a drop,
so that possibility something goes wrong is less.


Sources/DbPackages-postgresql.php
```diff
@@ -498,6 +498,13 @@ function smf_db_change_column($table_name, $old_column, $column_info)
 		else
 		{
 			$smcFunc['db_query']('', '
+				DROP SEQUENCE IF EXISTS ' . $table_name . '_seq',
```
colinschoen: Why is this necessary?

colinschoen (approved):
LGTM approval

colinschoen:
Nice find, thank you.

PR #4247, opened 2017-08-19
Reverts to serialize in this case; adds brief explanation of logic

Sesquipedalian:
There's no need to break out to $smcFunc's JSON functions in this case. Since the use is entirely internal to this function, it is faster, simpler, and cleaner to use a built in function.

Also, @jdarwood007 wanted to know why we go through the effort to do this at all. Hopefully my comment makes it clearer. :)

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -5075,8 +5075,6 @@ function smf_list_timezones($when = 'now')
 		// First, get the set of transition rules for this tzid
 		$tzinfo = timezone_transitions_get($tz, $when, $later);
 
-		$tzinfo[0]['abbr'] = fix_tz_abbrev($tzid, $tzinfo[0]['abbr']);
-
 		// Use the entire set of transition rules as the array *key* so we can avoid duplicates
 		$tzkey = serialize($tzinfo);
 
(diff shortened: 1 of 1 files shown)


jdarwood007:
I assume you want to know why I put it into $smcFunc?

I put into $smcFunc as we can easily update the code base later should another issue arise, similar to the serialized security issues.  We already have a smc_json_decode to handle some situations during decoding.  Its possible we may run into this with encoding later.   I would almost advocate we do this for any function that handles raw user data at times.

From my testing, the difference between the straight calls and $smcFunc calls was very minor.

Sesquipedalian:
> I assume you want to know why I put it into $smcFunc?

No, the evolution was clear enough. I initially wrote the code to use serialize, which was subsequently changed in a general sweep to replace all uses of serialize with json_encode, and which was in turn changed in a general sweep to use an abstracted $smcFunc call.

I only tagged you in the PR because you had put the following comment into the code: `// !!! TODO: Why encode this and then decode it here?` Taking that out and adding my comment to make the reason clearer was dealing with that. :)

> I put into $smcFunc as we can easily update the code base later should another issue arise, similar to the serialized security issues. We already have a smc_json_decode to handle some situations during decoding. Its possible we may run into this with encoding later. I would almost advocate we do this for any function that handles raw user data at times.

I agree with doing that for any function that handles raw user data. It is just that this one never does. I'm working purely with internal PHP data here, so a good ol' serialize is the best choice in this case.

jdarwood007:
Ok that was because we serialized the data and then a little bit later unserialize it.  There was no caching or other stuff being done.  We went from one foreach to another in the same function.  It didn't make sense.  I assume there is a faster way than serializing the data to check for duplicates.  Even if we used a custom function to do so.

Sesquipedalian:
I started writing a long explanation, and then I realized that I don't really need to bother with unserialize if I just append one bit of data to the array I build the first time. :)

PR #4249, opened 2017-08-20
Better gender stat querry

albertlast:
Added the default theme id,
to evade possible issues.


Sesquipedalian:
Why should the gender stats only include people who are using the default theme?

sbulen:
I think he was trying to avoid double-counting if someone had multiple themes installed.  Trying to get one count per user.

albertlast:
We talk in the irc about this topic and
we came to the result that the profile data is safed everytime in theme id 1.
So it's don't care which theme you use the gender information asre safed in theme id 1.
End of the day is protect us from double counting like sbulen mention.

Sesquipedalian:
Okay, that makes sense. Thanks for the explanation.

PR #4250, opened 2017-08-20
Removes obsolete fix_serialized_columns() function

### LINKED ISSUE #4127: Should fix_serialized_columns() be removed?
The `fix_serialized_columns()` function in Subs-Charset.php "Fixes corrupted serialized strings after a character set conversion" in the `id_action` and `extra` fields of the the log_actions table. But the strings in those fields are JSON encoded now, not serialized. So as far as I can see, the function does nothing now and can be removed.

Sesquipedalian:
Closes #4127


XinYenFon (changes requested):
Is there any specific reason you left this 😆 

https://github.com/SimpleMachines/SMF2.1/blob/a878de567d832dc284d62757e2b4a759be0a39bf/Sources/ManageMaintenance.php#L679-L681

Sesquipedalian:
D'oh!

PR #4256, opened 2017-08-22
Ensure 'alltags' matching is case insensitive.

gizmo71:
Otherwise, old posts in the form [URL=x]foo[/url] aren't parsed correctly.
This brings both uses of $alltags_regex into line.

Signed-off-by: Dave 'Gizmo' Gymer <github@davegymer.org>


Sesquipedalian (approved):
LGTM. Good catch.

PR #4257, opened 2017-08-22
Convert the logo to svg

Gwenwyfar:
Just switching the logo for a better quality version. Colors were changed to the same ones currently used in simplemachines.org.


Sesquipedalian:
SMF 2.1 currently supports IE10+. IE10 and 11 support SVG, but they are buggy about it. Worse, 11 fixed some of 10's bugs, but introduced others, IIRC.

So, have you tested this in both IE10 and IE11?

Gwenwyfar:
We should have a full "versions supported" list publicly available somewhere (or if we do, where is it?)

In any case, IE supports SVG since IE9, and its bugs in plain SVG images are mainly related to scaling. This SVG has width and viewbox defined, so it should not be an issue.

Here are some screenshots for edge, IE 11 and 10 if you want to be absolutely sure...
edge
ie10
ie11

XinYenFon:
@Sesquipedalian not our problem. If we are going to hold anything for browser related bugs we shouldn't be doing anything.

frandominguezl:
We should also have in mind that not too much people use IE9-10 (not even IE11) nowadays

jdarwood007:
IE11 is still in good use even nowadays.  Edge doesn't support all of the legacy features that IE still supports that are used in corporate environments.  Its becoming better but still in use.  Microsoft SharePoint is a fine example of something you need IE to get the full feature set out of.

Gwenwyfar:
@Antes: Agreed. In particular for anything that is purely cosmetic like this. We'd spend 90% of our time chasing after bugs instead of actually doing anything if we were to make it all perfect in every browser. Legacy support should be limited to making sure all basic functions work. In fact we'd have to scrap many current features that don't all work 100% in every browser...

For SVG, it already has widespread use for years as well.

Sesquipedalian:
> Here are some screenshots for edge, IE 11 and 10 if you want to be absolutely sure...

Excellent.

Sesquipedalian:
In general I agree with the sentiments here about support for old IE and its bugs. However, in this particular case there is no pressing reason why we *need* to use an SVG, so I wanted to make sure that we didn't break backwards compatibility needlessly for this very prominent logo. Since it doesn't, I'm more than happy to see this merged. :)

Gwenwyfar:
Great, thanks :)

@Sesquipedalian: It may have been just a poor choice of words then.

Also, for future reference, editors like inkscape can save SVGs in formats that are compatible to supporting browsers.

PR #4261, opened 2017-08-23
Fix nextSubStep to be nextSubstep.

gizmo71:
A single instance had the wrong case.


Sesquipedalian:
Strictly speaking, this doesn't matter because PHP is case-insensitive. But I like consistency, too.

PR #4262, opened 2017-08-23
Calls to mysqli_insert_id require the connection, not the result set.

gizmo71:


Sesquipedalian:
@albertlast

albertlast (approved):

jdarwood007:
The $rs variable is named badly but it should contain the resource identifier.  If none is specified it falls back to the db_connection.

albertlast:
When you look,
the method is used only in upgrade script 1.1 and 2.0.
And on both places without var,
sooo i think we only clean up.

jdarwood007:
Then it would be ok I guess,  Going off the php docs for mysql_insert_id is what tells me what should be there.

PR #4264, opened 2017-08-25
Implements boardurl token substitution

### LINKED ISSUE #3726: $settings['theme_url']
Each $settings['theme_url'] and $settings['default_theme_url'] doesn't respect SSL. It still uses http after change to SSL. This causes most  ajax request to be blocked by the browser. loadJavaScriptFile() also uses $settings['default_theme_url'] which prevents scripts form being loaded.

When doing the switch to SSL a small function needs to be called to change all DB stored urls to the new schema  or store all local urls as as a token {localURL} and do a replacement when retrieved from the DB.

The urls can be changed from within the admin panel but its not very obvious, might be better to just do it automatically for the user when doing the change to SSL or as a routine maintenance task.

Sesquipedalian:
As discussed in #3726, this commit implements seamless substitution of the value of $boardurl with a token when reading and writing to the database.

This approach solves a few problems:

* Activating or deactivating SSL will not break existing URLs in settings
* Changing the domain or path of the forum will not break existing URLs in settings
* Quotes and links in posts that refer to other posts will always use the proper value of $boardurl

It works by (1) replacing all occurrences of the value of $boardurl with the literal string `{$boardurl}` just before writing in smf_db_query(), and then (2) performing the opposite operation immediately after reading using smf_db_fetch_assoc() or smf_db_fetch_row().

Note:
1. During a fresh install, we no longer replace `{$boardurl}` in the install SQL files with the value of $boardurl, because we want to use that token in the database now.
2. During an upgrade to 2.1, we replace all literal references to the value of $boardurl in posts and personal messages with the token.

Fixes #3726

### DIFF
--- a/Sources/Subs-Db-postgresql.php
+++ b/Sources/Subs-Db-postgresql.php
@@ -389,12 +389,12 @@ function smf_db_query($identifier, $db_string, $db_values = array(), $connection
 		// Inject the values passed to this function.
 		$db_string = preg_replace_callback('~{([a-z_]+)(?::([a-zA-Z0-9_-]+))?}~', 'smf_db_replacement__callback', $db_string);
 
-		// Replace $boardurl with a token (we restore it when we retrieve the value)
-		$db_string = str_replace($boardurl, '{$boardurl}', $db_string);
-
 		// This shouldn't be residing in global space any longer.
 		$db_callback = array();
 	}
+
+	// Replace $boardurl with a token (we restore it when we retrieve the value)
+	$db_string = str_replace($boardurl, '{$boardurl}', $db_string);
 
 	// Debugging.
 	if (isset($db_show_debug) && $db_show_debug === true)
--- a/Sources/Subs-Db-postgresql.php
+++ b/Sources/Subs-Db-postgresql.php
@@ -995,8 +995,17 @@ function smf_db_escape_wildcard_string($string, $translate_human_wildcards = fal
  */
 function smf_db_fetch_all($request)
 {
+	global $boardurl;
+
 	// Return the right row.
-	return @pg_fetch_all($request);
+	$rows = @pg_fetch_all($request);
+
+	if (is_array($rows))
+		array_walk_recursive($rows, function(&$column) use ($boardurl) {
+			$column = str_replace('{$boardurl}', $boardurl, $column);
+		});
+
+	return $rows
 }
 
 ?>
--- a/Sources/Subs-Db-postgresql.php
+++ b/Sources/Subs-Db-postgresql.php
@@ -641,9 +641,12 @@ function smf_db_fetch_row($request, $counter = false)
 		$row = @pg_fetch_row($request, $db_row_count[(int) $request]++);
 	}
 
-	array_walk($row, function(&$column) use ($boardurl) {
-		$column = str_replace('{$boardurl}', $boardurl, $column);
-	});
+	if (is_array($row))
+	{
+		array_walk($row, function(&$column) use ($boardurl) {
+			$column = str_replace('{$boardurl}', $boardurl, $column);
+		});
+	}
 
 	return $row;
 }
@@ -672,9 +675,10 @@ function smf_db_fetch_assoc($request, $counter = false)
 		$row = @pg_fetch_assoc($request, $db_row_count[(int) $request]++);
 	}
 
-	array_walk($row, function(&$column) use ($boardurl) {
-		$column = str_replace('{$boardurl}', $boardurl, $column);
-	});
+	if (is_array($row))
+		array_walk($row, function(&$column) use ($boardurl) {
+			$column = str_replace('{$boardurl}', $boardurl, $column);
+		});
 
 	return $row;
 }
--- a/Sources/Subs-Db-postgresql.php
+++ b/Sources/Subs-Db-postgresql.php
@@ -626,7 +626,7 @@ function smf_db_error($db_string, $connection = null)
  */
 function smf_db_fetch_row($request, $counter = false)
 {
-	global $db_row_count;
+	global $db_row_count, $boardurl;
 
 	if ($counter !== false)
 		$row = pg_fetch_row($request, $counter);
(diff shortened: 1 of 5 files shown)


Sources/Subs-Db-postgresql.php
```diff
@@ -626,14 +629,26 @@ function smf_db_fetch_row($request, $counter = false)
 	global $db_row_count;
```
albertlast: $boardurl is missing

Sources/Subs-Db-postgresql.php
```diff
@@ -648,14 +663,24 @@ function smf_db_fetch_assoc($request, $counter = false)
...
```
albertlast: $boardurl is missing

albertlast (changes requested):

albertlast:
In my eyes is your implementation to mutch "magic".
The problem you see here:
grafik
The stuff get interpreted and even in code blocks.

Sources/Subs-Db-postgresql.php
Sesquipedalian: Oops. :)

Sesquipedalian:
That was just because I forgot to add `$boardurl` to the global variable declaration in the PostgreSQL versions of `smf_db_fetch_row()` and `smf_db_fetch_assoc()`. Should be fixed now.

other/upgrade_2-1_mysql.sql
```diff
@@ -2066,6 +2066,14 @@ UPDATE {$db_prefix}personal_messages SET body = REPLACE(REPLACE(body, '[blue]',
 ---#
 
 /******************************************************************************/
+--- Tokenizing references to forum URL in posts and personal messages
+/******************************************************************************/
+---# Replacing $boardurl with token
+UPDATE {$db_prefix}messages SET body = REPLACE(body, '{$boardurl}', CONCAT('{$', 'boardurl}')) WHERE body LIKE '%{$boardurl}%';
+UPDATE {$db_prefix}personal_messages SET body = REPLACE(body, '{$boardurl}', CONCAT('{$', 'boardurl}')) WHERE body LIKE '%{$boardurl}%';
+---#
```
jdarwood007: Hope you realize what that sort of update this does on a multi-million post forum.  This will be a killer query.

Sesquipedalian: Not much different that all the BBC replacement queries immediately above it.

jdarwood007: wow those will also be killers.  Well I guess accept this PR as it is.  We should address this separately.  I couldn't upgrade sm.org with that.  A LIKE wildcard (%) on the messages table would kill the database to try and figure that out.  If it could use the sphinx search index, then it wouldn't be that bad.

Sesquipedalian: If you can come u p with a better method, I'd be 100% in favour of that.

jdarwood007: Well again, this would be outside of this PR.

But I think we could create a scheduled task, call it something like "Background Upgrade Tasks".  This would have the code to process this a set of messages at a time (such as lets say 100 of them).  We would track this upgrade process in a separate table from the settings table (to avoid causing a recreate of the settings cache for every update).  This would also be able to upgrade personal messages.

If we do it this way, the users would experience for a short while the possibility of the old bbc codes not working and the boardurl replacement not functioning on older posts.  Once the background upgrader went through, it would resolve it.

The best part is, this allows us to easily keep cleaning up parts of our database that are heavy hitters for large forums, while keeping the main upgrader faster.

This of course only works with things like messages and pms where it would be acceptable for old BBC to not work.  If we did it on other things, we may have to have combat code in our system to handle this gracefully.

sbulen:
IMO - I believe this  is too broad & too risky a change to be made at this late stage.  I bet we'll be finding new issues for months.  

And...  Why?  We already provide the tools to fix this problem in #3726  - repair_settings.php & the admin console.  These issues are easily diagnosed & solved.  This is literally a 5 minute fix for an admin.  This is not a necessary change at this point.  

I really like the token idea, & even better, I really like the schemeless URL idea.  but I think these are 3.0 discussions.  

If we *MUST* do this now, the mass {boardurl} substitutions should be done via the upgrader.  Not in the app.  We should rethink the approach to do the heavy lifting during upgrade, and then focus on updates only when URLs are stored - preferably ONLY in admin functions.  No massive updates should be needed in the normal course of events.

I also think repair_settings.php needs some changes to support this properly.

Sorry, I guess I'm the contrarian today...

albertlast:
The fetch_all stuff you missing on both database files.

I would like to mention that i'm agree with sbulen that the implentation way got to high impact.
The use of a token is possible fine,
but to manipulate the output of sql did we not before and this with good reason.
We don't know on db side which kind of response is wanted.
Fopr exp. when i write the token in a code block,
than i exptected that this keep the token and is not transformet in output the url of the board.
Also from performance side I don't like it,
because you go over all col and all row set to find this token.

Sesquipedalian:
Tokenizing posts during the upgrade process isn't really necessary anyway. Since @jdarwood007 is right that it would be a very heavy task on large boards, I've taken that out now.

Sesquipedalian:
> The fetch_all stuff you missing on both database files.

Sesquipedalian:
> The fetch_all stuff you missing on both database files.

Added now.

albertlast:
i notice but the key issue keep,
that this pr do ouput modifcation.

Sesquipedalian:
Do you mean that you want to be able to write the literal string `{$boardurl}` in a post without it being changed into the actual value of $boardurl?

albertlast:
When i do this in a [code][/code] yes.
But ther exists many cases where this behavior is unwanted or does hurt.
To give you "taste" where the stuff get unfunny.
I could create a user with nick {$boardurl}lol and the url is than displayed,
which is show in which complexity you try to get into.

recommendation, do this in this way we do for ip stuff ->
the programm part hat to handle this case not the db layer.

But like already mention this solution would to many impact in my eyes in the state of smf 2.1 now.

Sesquipedalian:
> I could create a user with nick {$boardurl}lol and the url is than displayed

That is a very good point.

Issue #4096, opened 2017-06-04
Deprecating compatibility functions


NanoSector:
Hopping through the codebase, I found the following things...

- [ ] un_htmlspecialchars - PHP 5.1 and above includes htmlspecialchars_decode. I'm not sure how this function differs from it (it says it removes things but it just replaces things from reading the code).
- [ ] html_to_bbc, bbc_to_html, legalise_bbc (and sort_array_length) - I know the notes say that they shouldn't be removed due to mod compatibility, however mods have to be updated for 2.1 anyway so authors could move to a different implementation.
- [ ] theme_postbox - Used in SMF 1.1. Need I say more :x 
- [ ] smf_setcookie has a note that it can be removed since SMF requiers PHP >= 5.3.8 now.

live627:
Huh, I always  thought  un_htmlspecialchars() and htmlspecialchars_decode() did the same thing.

Sesquipedalian:
> un_htmlspecialchars - PHP 5.1 and above includes htmlspecialchars_decode. I'm not sure how this function differs from it (it says it removes things but it just replaces things from reading the code).

Yeah, they look the same to me. We should probably keep it around for old mods, but we could replace the current content of the function with a simple `return htmlspecialchars_decode($string)`.

> html_to_bbc, bbc_to_html, legalise_bbc (and sort_array_length) - I know the notes say that they shouldn't be removed due to mod compatibility, however mods have to be updated for 2.1 anyway so authors could move to a different implementation.

In theory the mods should be updated, sure, but I doubt they all will. It doesn't hurt us to keep these functions around for now.

> theme_postbox - Used in SMF 1.1. Need I say more :x

No, you needn't. Death to it!

> smf_setcookie has a note that it can be removed since SMF requiers PHP >= 5.3.8 now.

Ditto for this one.

XinYenFon:
@Sesquipedalian backward compat broken for a long time so anything unnecessary should better be dead in sense of consistency.

But the thing is, anything removed from SMF (Source) should be noted in wiki so mod authors can update their mods far more easily.

NanoSector:
We need a wiki page with all the developer changes anyway, like hooks added/changed and stuff.

colinschoen:
i'll make a post in the doc team boards to see if they can get a page template sorted for us and we can start updating it.

frandominguezl:
[...]

PR #4266, opened 2017-08-25
Fix #4098

### LINKED ISSUE #4098: critical error
why is this error a critical error in the error log? i dont think this requires the admin attention to do something. it does not even tell us what account was being attempted to log into.
Sorry, but you've reached your login attempts threshold. Please wait 30 seconds and try again later.

NanoSector:
Introduce 'login' error category and add two types to this category. Also change up some strings a bit.

### DIFF
--- a/Sources/Errors.php
+++ b/Sources/Errors.php
@@ -85,6 +85,7 @@ function log_error($error_message, $error_type = 'general', $file = null, $line 
 		'cron',
 		'paidsubs',
 		'backup',
+        'login',
 	);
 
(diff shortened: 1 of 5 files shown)


NanoSector:
Woops, commits reference #4096 but needs to be #4098 of course.

Sesquipedalian (changes requested):
Nit: weird indenting. Please convert to tabs.

NanoSector:
Should be fixed

Sesquipedalian (approved):

PR #4275, opened 2017-08-30
Fixes #4273

### LINKED ISSUE #4273: Install - Cookie names must not be empty
Latest commit atm. ( 6ad29c7a6bf0569b7833f6debe96c45454b035f0 )
https://smf.local/install.php?step=5
```
Warning
C:\xampp\VirtualHosts\SMF2.1\Sources\Subs-Auth.php
on line
837
```
After install finishes i used login link and everything is fine so its just minor installer problem.

sbulen:
Installer - An innocent victim of some recent cleanup...

### DIFF
--- a/other/install.php
+++ b/other/install.php
@@ -1564,7 +1564,7 @@ function AdminAccount()
 function DeleteInstall()
 {
 	global $smcFunc, $db_character_set, $context, $txt, $incontext;
-	global $current_smf_version, $databases, $sourcedir, $forum_version, $modSettings, $user_info, $db_type, $boardurl, $cachedir;
+	global $current_smf_version, $databases, $sourcedir, $forum_version, $modSettings, $user_info, $db_type, $boardurl, $cachedir, $cookiename;
 
 	$incontext['page_title'] = $txt['congratulations'];
 	$incontext['sub_template'] = 'delete_install';


NanoSector:
I tracked down the behavior:

```
[20:57:29] <+NanoSector> $cookiename is defined in Settings.php
[20:58:21] <albertlast> in the install phase could be this empty?
[20:58:42] <+NanoSector> oh, i get it
[20:58:49] <+NanoSector> it's defined in Settings.php but that doesn't make it a global
[20:59:17] <+NanoSector> so it's required to define it as a global in DeleteInstall() *before* Settings.php is loaded
[20:59:24] <+NanoSector> https://github.com/SimpleMachines/SMF2.1/blob/7e16ce2edf88a1e4eb521a05c0bbd9ac99a3de59/other/install.php#L1564-L1573
[20:59:27] <Botergos> [NanoSector] SMF2.1/install.php at 7e16ce2edf88a1e4eb521a05c0bbd9ac99a3de59 · SimpleMachines/SMF2.1 · GitHub
[20:59:39] <+NanoSector> since you're saying "oh, if you encounter $cookiename, make it a global"
```

essentially, any places we load Settings.php we should global these variables. Or, you know, do it in Settings.php.

Sesquipedalian:
> it's defined in Settings.php but that doesn't make it a global

This is incorrect. All the variables in Settings.php are already global variables, because PHP is in the global scope at the time when those variables are defined.

It would have been correct to say, "It's defined in Settings.php, so it is a global, but that doesn't make it automatically available in a local context."

NanoSector:
That is not true. Settings.php is included inside a function (in the scope of the function), so these variables only exist inside that function unless told otherwise. Including a file doesn't simply 'break out' of the current scope.

This is further acknowledged in this manual entry: https://secure.php.net/manual/en/language.variables.scope.php
```Using global keyword outside a function is not an error. It can be used if the file is included from inside a function.```

Sesquipedalian:
I'm afraid you are mistaken in this case, my friend. :) Please see http://php.net/manual/en/language.variables.scope.php and http://php.net/manual/en/function.include.php. Especially pertinent is this statement:

> When a file is included, the code it contains inherits the variable scope of the line on which the include occurs. Any variables available at that line in the calling file will be available within the called file, from that point forward. However, all functions and classes defined in the included file have the global scope.

NanoSector:
That especially outlines what I just said. “the code it contains inherits the variable scope of the line on which the include occurs.” E.g. if it is included inside a function, the scope stays inside that function. Functions and classes in that file will be in global, variables will not be.

Van: Jon Stovell
Verzonden: maandag 4 september 2017 01:20
Aan: SimpleMachines/SMF2.1
CC: NanoSector; Comment
Onderwerp: Re: [SimpleMachines/SMF2.1] Fixes #4273 (#4275)

I'm afraid you are mistaken in this case, my friend. :) Please see http://php.net/manual/en/language.variables.scope.php and http://php.net/manual/en/function.include.php. Especially pertinent is this statement:
When a file is included, the code it contains inherits the variable scope of the line on which the include occurs. Any variables available at that line in the calling file will be available within the called file, from that point forward. However, all functions and classes defined in the included file have the global scope.

Sesquipedalian:
But index.php isn't inside a function when it includes Settings.php. It is in the global scope. There's nothing else we need to do.

NanoSector:
Index.php is not, no. OTHER occurrences might not be the same and we should investigate them.

Van: Jon Stovell
Verzonden: maandag 4 september 2017 01:48
Aan: SimpleMachines/SMF2.1
CC: NanoSector; Comment
Onderwerp: Re: [SimpleMachines/SMF2.1] Fixes #4273 (#4275)

But index.php isn't inside a function when it includes Settings.php. It is in the global scope. There's nothing else we need to do.

PR #4277, opened 2017-08-30
Remembers moderator's real group in loadMemberData and loadMemberContext

Sesquipedalian:
Because $user_profile[$id]['member_group'] gets overwritten for moderators in loadMemberData(), there has been no easy way, short of a new query, to get the name of a moderator's primary membergroup. (The id number of the moderator's true primary group is retained, but not the group's name.)

With this change, $user_profile[$id]['primary_group'] and $memberContext[$user]['primary_group'] keep track of the name of member's true primary group.


Sesquipedalian:
By way of explanation, while a default install of SMF 2.1 doesn't do anything with this info, it is entirely plausible that other themes and/or mods will use it. (Indeed, I can fully guarantee that some will. :wink:)

PR #4282, opened 2017-09-04
Fixes #3726

### LINKED ISSUE #3726: $settings['theme_url']

sbulen:
This PR syncs all URLs to the force_ssl setting.  If force_ssl = 2, then all key URLs are set to use the https:// scheme.  A force_ssl value of 1, 0, or unset results in all URLs getting set to the http:// scheme.  

Updates are not made to URLs that have been moved out from under $boardurl.  The assumption is that the admin moved them for a reason, so leave the scheme alone.  

If no cert is found & force_ssl is set to 0 (disabled), the user is not allowed to change it to something disastrous.

The user must press "Save" for the logic to kick in.  Note that it doesn't check for a change in force_ssl - so, simply bringing up the screen and pressing "Save" may actually fix errors...  This may seem a little odd, but the URL values being updated aren't displayed on the screen.  And if they are out of sync with force_ssl, pressing "Save" really should fix them...  

The help text was updated to warn that URLs may get changed.

Tested in Windows & Linux, with certs & without, with redirects & without, and with some of the affected folders moved outside of $boardurl to confirm they'd be left alone.


Sources/ManageServer.php
```diff
@@ -213,6 +227,141 @@ function ModifyGeneralSettings($return_config = false)
...
+	if ($new_force_ssl == 2)
+		$newval = strtr($boardurl, array('http://' => 'https://'));
+	else
+		$newval = strtr($boardurl, array('https://' => 'http://'));
+	updateSettingsFile(array('boardurl' => '\'' . addslashes($newval) . '\''));
+
+	$new_settings = array();
+
+	// Check $smileys_url, but only if it points to a subfolder of $boardurl
+	if (BoardurlMatch($modSettings['smileys_url'])) {
```
Sesquipedalian: Nit: SMF's coding style requires a line break before the `{`

... I honestly don't know *why* we do it that way, but thus did the Ancient Powers declare that it should be done.

Themes/default/languages/Help.english.php
```diff
@@ -616,7 +616,7 @@
 $helptxt['alert_pm_new'] = 'Notifications of new personal messages do not appear in the Alerts pane, but appear in the "My Messages" list instead.';
 $helptxt['alert_event_new'] = 'This will send out an alert or email as requested if there is a new calendar event added. However, if that event is posted and a topic is added, you will not get an alert for the event if you\'re already following that board - the alert from following the board would cover this.';
 
-$helptxt['force_ssl'] = '<strong>Test SSL and HTTPS on your server properly before enabling this, it may cause your forum to become inaccessible.</strong> Enable maintenance mode if you are unable to access the forum after enabling this';
+$helptxt['force_ssl'] = '<strong>Test SSL and HTTPS on your server properly before enabling this, it may cause your forum to become inaccessible.</strong> Enable maintenance mode if you are unable to access the forum after enabling this.<br><br><strong>Changing this setting will update all forum URLs to match your selection.</strong>';
```
Sesquipedalian: The phrase "all forum URLs" might confuse people. Naive admins might think this will update URLs in the text of forum posts, while knowledgable admins will be worried that customized URLs in their settings will be overwritten. I suggest something like this:

`Changing this setting will update "' . $boardurl . '" to "' . (str_replace('http:', 'https:', $boardurl) . '" in several key settings for your forum, including the URLs for your theme files and images, smileys, avatars, etc. Customized URLs in these settings will not be affected.`

Sesquipedalian (changes requested):

Sources/ManageServer.php
sbulen: Done.

Themes/default/languages/Help.english.php
sbulen: Done.  Let me know if you like the new wording.  I was hesitant to use $boardurl in the message, because there are times when it doesn't match Settings.php.   (SMF changes it in memory sometimes...)  It was confusing to me to see a value displayed that I knew was wrong.

Sesquipedalian (approved):
LGTM

Issue #2872, opened 2015-05-23
Weird character handling in title of topics.


FearThe1337:
When creating a topic with the following subject: 

```
&#32;
```

The editor automatically converts the subject to 

```
&#38;#32;
```

Not sure if related to #1066. (Was trying to find a way to get that bug, and tried random stuff in the title)

Also noticed that for example:

```
&#160

changes into

&#38;#160
```

Oldiesmann:
Something appears to be getting double-encoded there - 38 is the HTML entity for the & character (though "&amp" is usually used instead)

Sesquipedalian:
Fixed in #4283

PR #4283, opened 2017-09-04
Literal entities

### LINKED ISSUE #4203: bbc code bug
a post with this message
` [code]'&#39;', '&#96;', '&#123;', '&#125;'[/code] `
get as result `'&#39;', '&#96;', '{', '}'`

In the db look like that
`"[code]&#39;&amp;#39;&#39;, &#39;&amp;#96;&#39;, &#39;&#123;&#39;, &#39;&#125;&#39;[/code]"`

i would guess ther are missing amp;

This bug i copy from elkart.
https://github.com/elkarte/Elkarte/issues/2961

Sesquipedalian:
This fixes #4203, which was caused by several not entirely consistent ways of trying to deal with encoded entities in user supplied text. Because SMF 2.1 uses only UTF-8 character encoding, these mostly-but-not-completely functional workarounds are no longer necessary.


Sesquipedalian:
I strongly encourage testers to try to find ways to break this code. I didn't find any ways to produce unexpected behaviour, but I may not have thought of everything.

Sesquipedalian:
@albertlast

albertlast:
looks better,
when you safe your post as draft and post then you got the wrong output:
grafik

Sesquipedalian:
Thanks, @albertlast. I will take a look at the drafts scripts soon.

albertlast:
Reporting to moderator does thing wrong when you click on preview:
grafik
(the preview is right) but the "enter comment" changed ther to a different behavior,
dunno what is right...
i would expect that enter comment and preview are got the same output...

Sesquipedalian:
I think this solves problems when saving drafts and reporting to mods. Care to give it a test, @albertlast?

albertlast:
well when i create a subject with `&#32;` then this get `&#38;#32;` (it's no new bug)

the other cases looks good.

Sesquipedalian:
> well when i create a subject with &#32; then this get &#38;#32; (it's no new bug)

Do you mean when creating a new post using `?action=post`?

albertlast:
Well yes i mean post in the first way,
but when you create a replay this issue happen to:
grafik

On database side you see that amp stuff is working right:
grafik


Btw the issue came from here: https://github.com/SimpleMachines/SMF2.1/issues/2872

albertlast:
Found the src of issue: https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Themes/default/scripts/script.js#L1446
changed this line to:
`document.forms[sFormName][aElementNames[i]].value = document.forms[sFormName][aElementNames[i]].value;`

And it's works fine.

Themes/default/scripts/script.js
```diff
@@ -1443,7 +1443,7 @@ function smc_saveEntities(sFormName, aElementNames, sMask)
 	for (var i = 0, n = aElementNames.length; i < n; i++)
 	{
 		if (aElementNames[i] in document.forms[sFormName])
-			document.forms[sFormName][aElementNames[i]].value = document.forms[sFormName][aElementNames[i]].value.replace(/&#/g, '&#38;#');
+			document.forms[sFormName][aElementNames[i]].value = document.forms[sFormName][aElementNames[i]].value;
```
albertlast: I know i recommend this "solution",
when i look deeper in this code than this code doesn't do any thing more? --> can we drop this?

Sesquipedalian: Yes. That's what 4c50dbe does.

albertlast (approved):

PR #4285, opened 2017-09-04
Everybody loves DLs

MissAllSunday:
Some dirty old dev removed our precious definition lists, some even older dev got really upset for some reason... 

Suffice to say, this is top priority!11!eleventeen!1!!

### DIFF
--- a/Themes/default/Who.template.php
+++ b/Themes/default/Who.template.php
@@ -223,19 +223,21 @@ function template_credits()
 		<div class="cat_bar">
 			<h3 class="catbg">', $txt['credits_modifications'], '</h3>
 		</div>
-		<div class="windowbg noup">';
+		<div class="windowbg noup">
+			<ul>';
 
 		// Display the credits.
 		if (!empty($context['credits_modifications']))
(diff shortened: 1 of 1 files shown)


XinYenFon:
Hi,

Just confused why you are using double dd under dl tag, I thought it requires dt/dd to be valid?

thanks for the fix.

MissAllSunday:
Modification credits do not have a term/name like the other credits does so using a definition list for them doesn't make much sense. I would use a ul/li combo here but the old dev was too concern about his precious dls so I put them back. 😏 

edit, typos

XinYenFon:
Fine by me 😆

MissAllSunday:
I can change the PR as needed, just need to know what should be done.

- keep on using dls and insert a blank dt to validate againts html 5
- use ul/li
- use br

Sesquipedalian:
A ul list makes sense to me.

XinYenFon:
+1 on ul/li

Oldiesmann:
I agree with using ul/li as well

Issue #4287, opened 2017-09-06
Mention didn't check if user is able to access this topic


albertlast:
When you mention a person X in a topic where he had no access,
he keep getting the notification that he was mention:

d3vcho was able to reproduce this error:
https://i.gyazo.com/b72d180e8bd8b1922b559cdcf9f2fd82.png

 the post is censored.

Possible root of this evil would be here:
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/tasks/CreatePost-Notify.php#L51

Missing check of the array $msgOptions['mentioned_members'] if the mention user is able to access.
other placce could be https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs-Post.php
that we didn't add any person here how had no access.

frandominguezl:
To reproduce this, you just need to create (or edit) a board's permissions to make it invisible to a group of users and then just mention an user who is not able to see that board.

albertlast:
when i look on elkarte side they fixed on notify side:
https://github.com/elkarte/Elkarte/blob/development/sources/subs/MentionType/MentionBoardAccessAbstract.php#L60
(i believe that is the code how do the check)

PortaMx:
😀 this we fixed it weeks ago .. 
and I can say, it's very simple 😋 .. for mentions and quotes ..

XinYenFon:
@PortaMx unless you are planning to contribute that fix towards this software please do not flood.

PortaMx:
I have posted many bug and also fixes  .. but nothing of them was accepted by the people here.
So I stop reporting of bug and solved code and make my own work .. you understand ? yes ? no ? shit happens 😋
But .. I can you say that the fix is very simple .. a little bit code and moving exist code at a other place ..that's all

MissAllSunday:
I had a similar issue with Breeze custom notifications:  https://github.com/MissAllSunday/Breeze/blob/develop/Sources/Breeze/BreezeLog.php#L85

It isn't as nice as I would like it to be but it works, the extra query cannot be avoided unless you want to add some kind of dynamic cache or had previously loaded the needed data. I mitigated it by grouping as much messages as I was able to group together, thus avoiding doing multiple queries.

@PortaMx Your code wasn't accepted because it didn't meet the minimal criteria for SMF standards, it was almost incomplete or just plain wrong

PortaMx:
[...]

PR #4288, opened 2017-09-07
hide alerts fixing #4287

albertlast:
My try to solve the issue #4287
possible issue is here when you got a limit of 10 and
the last 10 are hide then you didn't see the 11th.

it's not the clean version which i like to do,
but i didn't got a clear vision how to create better version.

### DIFF
--- a/Sources/Profile-View.php
+++ b/Sources/Profile-View.php
@@ -325,17 +325,26 @@ function fetch_alerts($memID, $all = false, $counter = 0, $pagination = array())
 			continue;
 		if (isset($alert['extra']['board']))
 			if ($boards[$alert['extra']['board']] == $txt['board_na'])
+			{
 				unset($alerts[$id_alert]);
+				continue;
+			}
 			else
 				$alerts[$id_alert]['extra']['board_msg'] = $boards[$alert['extra']['board']];
 		if (isset($alert['extra']['topic']))
 			if ($alert['extra']['topic'] == $txt['topic_na'])
+			{
 				unset($alerts[$id_alert]);
+				continue;
+			}
 			else
 				$alerts[$id_alert]['extra']['topic_msg'] = $topics[$alert['extra']['topic']];
 		if ($alert['content_type'] == 'msg')
 			if ($msgs[$alert['content_id']] == $txt['topic_na'])
+			{
 				unset($alerts[$id_alert]);
+				continue;
+			}				
 			else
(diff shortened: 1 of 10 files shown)


Sources/Profile-View.php
```diff
@@ -324,11 +324,20 @@ function fetch_alerts($memID, $all = false, $counter = 0, $pagination = array())
 		if (!empty($alert['text']))
 			continue;
 		if (isset($alert['extra']['board']))
-			$alerts[$id_alert]['extra']['board_msg'] = $boards[$alert['extra']['board']];
+			if ($boards[$alert['extra']['board']] == $txt['board_na'])
+				unset($alerts[$id_alert]);
+			else
+				$alerts[$id_alert]['extra']['board_msg'] = $boards[$alert['extra']['board']];
```
jdarwood007: I'm only looking at the code here and not the context, but why are we checking this against $txt?

albertlast: all alerts where intialize with the txt https://github.com/albertlast/SMF2.1/blob/82522334aa8ee193ba505ed2a8c354be1ce8259e/Sources/Profile-View.php#L263
and i placed the check here if the intial text is in here or changed.
When the array got the intial value, the user doesn't hav the access right -> remove this from the list.

frandominguezl:
This is the result after applying the fix;

Sources/Profile-View.php
jdarwood007: Ok, seems weird we are doing the N/A, but ok.

albertlast:
well the last pr fix the issue that the numbers of new alerts showing is wrong.

what it do,
everytime a new alert come in,
the alerts field it's empty (null).
When then the first read action happen on alerts i look if the field is empty,
when not all is fine,
when empty(null) than i calc the amount.

but to be honest it's not feeling right.

alternativ:
I could place the count logic into updateMemberData but then i need to know the {query_see_board} of the receiving user, this could evade us from changing the table designe because keep not null able...

albertlast:
I'm very happy to bring a new solution of counting alerts.

first i revert the last two changes and added a new way,
this way do three things.

First it add a new function build_query_board($userid) to get the {query_see_board} from not current users.
Secondly it's fix fetch_alert where you can provide with $memID but this is calc with rights of the current user -> now it's calc with the provided memID.
Third updateMemberData calc alerts now on the query_see_board of every different user.

Idanban12 (approved):

PR #4289, opened 2017-09-08
Fixes #3393

### LINKED ISSUE #3393: View vs issue likes
If a user has permission to like something, but not to view, liking things will fail with the generic $txt['likes_error'] message, and it's not obvious that both permissions have to be given.

Also, can guests like things now? If not, the group should be removed from the Configuration > Features > Likes page... (Think this is the right area, am at work)

sbulen:
All users even guests can see likes.

Note (1): I could not test the SSI function - would appreciate another set of eyes & /or testing by someone.  
Note (2): I kept the 'cannot_view_likes' error message, as it is used as a generic trap message in Likes logic.

### DIFF
--- a/Themes/default/Display.template.php
+++ b/Themes/default/Display.template.php
@@ -794,7 +794,7 @@ function template_single_post($message)
 	}
 
 	// And stuff below the attachments.
-	if ($context['can_report_moderator'] || !empty($context['can_like']) || $message['can_approve'] || $message['can_unapprove'] || $context['can_reply'] || $message['can_modify'] || $message['can_remove'] || $context['can_split'] || $context['can_restore_msg'] || $context['can_quote'])
+	if ($context['can_report_moderator'] || !empty($modSettings['enable_likes']) || $message['can_approve'] || $message['can_unapprove'] || $context['can_reply'] || $message['can_modify'] || $message['can_remove'] || $context['can_split'] || $context['can_restore_msg'] || $context['can_quote'])
 	echo '
 							<div class="under_message">';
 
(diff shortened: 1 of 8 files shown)


frandominguezl (approved):

frandominguezl:
There are 2 text strings left from view likes; L240 and L241 from ManagePermissions.english.php

sbulen:
Yep, missed 'em...  Will fix asap, but may not be able to get to them until tomorrow.

sbulen:
Done.

Themes/default/Display.template.php
```diff
@@ -794,7 +794,7 @@ function template_single_post($message)
 	}
 
 	// And stuff below the attachments.
-	if ($context['can_report_moderator'] || !empty($context['can_see_likes']) || !empty($context['can_like']) || $message['can_approve'] || $message['can_unapprove'] || $context['can_reply'] || $message['can_modify'] || $message['can_remove'] || $context['can_split'] || $context['can_restore_msg'] || $context['can_quote'])
+	if ($context['can_report_moderator'] || !empty($context['can_like']) || $message['can_approve'] || $message['can_unapprove'] || $context['can_reply'] || $message['can_modify'] || $message['can_remove'] || $context['can_split'] || $context['can_restore_msg'] || $context['can_quote'])
```
Sesquipedalian: Here and in the similar statement below to close the div, you should probably replace `!empty($context['can_see_likes'])` with `!empty($modSettings['enable_likes'])`, rather than just deleting `!empty($context['can_see_likes'])`.

Sesquipedalian (changes requested):

Themes/default/Display.template.php
sbulen: Not everybody can like though...  What about folks in groups who don't have those permissions?

sbulen: Oh, I see it...  Will change & test, give me a few...

sbulen: Good catch!

sbulen:
Done.

Sesquipedalian (approved):

PR #4291, opened 2017-09-11
Better error handling check var in error

albertlast:
Check the var if isset before we use them.

### DIFF
--- a/Sources/Errors.php
+++ b/Sources/Errors.php
@@ -490,7 +490,7 @@ function log_error_online($error, $sprintf = array())
 	if (SMF == 'SSI' || SMF == 'BACKGROUND')
 		return;
 
-	$session_id = $user_info['is_guest'] ? 'ip' . $user_info['ip'] : session_id();
+	$session_id = isset($user_info['is_guest']) && $user_info['is_guest'] ? 'ip' . $user_info['ip'] : session_id();
 
 	// First, we have to get the online log, because we need to break apart the serialized string.
 	$request = $smcFunc['db_query']('', '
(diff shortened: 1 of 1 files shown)


Sesquipedalian (changes requested):
`!empty($user_info['is_guest'])` would be more concise.

Sources/Errors.php
```diff
@@ -490,7 +490,7 @@ function log_error_online($error, $sprintf = array())
 	if (SMF == 'SSI' || SMF == 'BACKGROUND')
 		return;
 
-	$session_id = $user_info['is_guest'] ? 'ip' . $user_info['ip'] : session_id();
+	$session_id = !empty($user_info['is_guest']) && $user_info['is_guest'] ? 'ip' . $user_info['ip'] : session_id();
```
Oldiesmann: The `&& $user_info['is_guest']` part is redundant here. If $user_info['is_guest'] is true, !empty will also be true.

albertlast: this only a catch when $user_info doesn't exists that we don't try to access  $user_info['is_guest'].
I could create error wher this happen that $user_info was not there.

Oldiesmann: See https://www.php.net/empty.

The `!empty()` call will return true if `$user_info['is_guest']` is true (or set to some other "non empty" value). The call will return false if `$user_info['is_guest']` is false (or some other "empty" value) or doesn't exist in the first place (including if `$user_info` isn't set). There's no need to check whether `$user_info` is true again after the `!empty()` call.

Sesquipedalian (approved):

PR #4296, opened 2017-09-14
Fixes #3247

### LINKED ISSUE #3247: Too many login attempts gives A LOT of errors due menu and other template related code.
I have been seeing errors in my error log about non existing indexes and such.

```
2: array_intersect(): Argument #2 is not an array
8: Undefined index: permissions
```

 I obtained the following stacktrace:

```
/forum/Sources/Security.php:918    array_intersect
/forum/Sources/Subs.php:3861    allowedTo
/forum/Sources/Subs.php:2978    setupMenuContext
/forum/Sources/Subs.php:3101    setupThemeContext
/forum/Sources/Subs.php:2757    template_header
/forum/Sources/Errors.php:399    obExit
/forum/Sources/Errors.php:199    setup_fatal_error_context
/forum/Sources/LogInOut.php:856    fatal_lang_error
/forum/Sources/Load.php:446    validatePasswordFlood
/forum/index.php:144    loadUserSettings
/forum/index.php:118    smf_main
```

It seems that the user's permissions are not loaded (Since they do not get logged in) but somehow it is expected to exist later on.
In addition no ip adress information is logged for these errors.

Edit, using a MySQL query, based on the fact that the ip logging fails I obtained several more related errors and messages.

``` SQL
select count(*),url,message,stack from smf_log_errors smp WHERE smp.ip = '' GROUP BY message,stack;
```

It seems there are a total of 66 different error messages associated with this.

sbulen:
loadTheme() can, in certain circumstances, get called very early in the logon process, e.g., via fatal_lang_error() via validatePasswordFlood() via loadUserSettings().  In these instances, $user_info is not set, and this triggers 80+ error messages, e.g., undefined index, etc.  ($user_info is set later in loadUserSettings(), after the call to validatePasswordFlood()...)

loadTheme() already attempts to deal with missing $user_info, but it does so by filling in blanks in $context['user'].  This PR loads 'guest' values in $user_info as well.  

Feedback welcome.  I considered a rewrite of loadUserSettings() to call validatePasswordFlood() later in the process, but I wasn't sure this is the ONLY scenario when loadTheme() or fatal_lang_error() would get called so early.

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -1944,20 +1944,22 @@ function loadTheme($id_theme = 0, $initialize = true)
 			'ignoreusers' => array(),
 		);
 		// Note we should stuff $user_info with some guest values also...
-		$user_info['id'] = 0;
-		$user_info['is_guest'] = true;
-		$user_info['is_admin'] = false;
-		$user_info['is_mod'] = false;
-		$user_info['username'] = $txt['guest_title'];
-		$user_info['language'] = $language;
-		$user_info['email'] = '';
-		$user_info['smiley_set'] = '';
-		$user_info['permissions'] = array();
-		$user_info['groups'] = array();
-		$user_info['ignoreusers'] = array();
-		$user_info['possibly_robot'] = true;
-		$user_info['time_offset'] = 0;
-		$user_info['time_format'] = $modSettings['time_format'];
+		$user_info = array(
+			'id' => 0,
+			'is_guest' => true,
+			'is_admin' => false,
+			'is_mod' => false,
+			'username' => $txt['guest_title'],
+			'language' => $language,
+			'email' => '',
+			'smiley_set' => '',
+			'permissions' => array(),
+			'groups' => array(),
+			'ignoreusers' => array(),
+			'possibly_robot' => true,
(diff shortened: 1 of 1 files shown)


Sources/Load.php
```diff
@@ -1942,6 +1943,21 @@ function loadTheme($id_theme = 0, $initialize = true)
...
+		$user_info['username'] = $txt['guest_title'];
+		$user_info['language'] = $language;
+		$user_info['email'] = '';
+		$user_info['smiley_set'] = '';
+		$user_info['permissions'] = array();
+		$user_info['groups'] = array();
+		$user_info['ignoreusers'] = array();
+		$user_info['possibly_robot'] = true;
+		$user_info['time_offset'] = 0;
+		$user_info['time_format'] = $modSettings['time_format'];
```
Oldiesmann: Can't we just define all of these in a single array() statement like we do with `$context['user']` above?

sbulen: Yes, I'll need a day or two to change & retest....

sbulen:
Requested changes made.

frandominguezl (approved):

PR #4305, opened 2017-09-16
Use cache folder for db_last_error

jdarwood007:
Also, Make sure we only verify minified files are writable if we have it enabled or have not set it yet.

Not all forums run with having world writable files in public folders.  This changes this to use the cache folder.  Which should be writable and per a admins configuration can be outside of the any public folders.

The only thing this doesn't cover it cleaning up Settings.php.  But so far none of the other changes done also cleanup Settings.php or prep it for a improved upgraded version.

### DIFF
--- a/Sources/Logging.php
+++ b/Sources/Logging.php
@@ -183,29 +183,29 @@ function writeLog($force = false)
  */
 function logLastDatabaseError()
 {
-	global $boarddir;
+	global $boarddir, $cachedir;
 
 	// Make a note of the last modified time in case someone does this before us
-	$last_db_error_change = @filemtime($boarddir . '/db_last_error.php');
+	$last_db_error_change = @filemtime($cachedir . '/db_last_error.php');
 
 	// save the old file before we do anything
-	$file = $boarddir . '/db_last_error.php';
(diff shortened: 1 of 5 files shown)


Oldiesmann:
From a support standpoint it's best to change Settings.php on upgrade as well. Otherwise it gets too confusing and people who upgrade from previous versions of SMF won't be able to take advantage of features such as this.

Sesquipedalian:
> The only thing this doesn't cover it cleaning up Settings.php. But so far none of the other changes done also cleanup Settings.php or prep it for a improved upgraded version.

Would you be willing to write some code to do that in the upgrader? Perhaps as part of another PR so that we can merge this one in the meantime.

jdarwood007:
I could look into that.  Would it be too late in this stage to do that?

The only way I see this being possible is having the Settings.php as a template in the upgrader, injecting the updated variables into said template, then testing to make sure the variables match between the current and new one, then finally writing the file.

Oldiesmann:
Can't we just use the existing functionality for updating Settings.php?

jdarwood007:
There isn't any template, the existing updater just updates vars and/or adds them if missing to the end.

Sesquipedalian:
> The only thing this doesn't cover it cleaning up Settings.php. 

I've put together some code to do this now that seems to work nicely. So I'm going to merge this now and then submit a PR with those further changes.

PR #4306, opened 2017-09-16
Add getVersion to Cache API

jdarwood007:
I haven't been able to properly test all of these, so please help with this.  Most of them are borrowed from what Subs-Admin.php uses.

### DIFF
--- a/Sources/CacheAPI-apc.php
+++ b/Sources/CacheAPI-apc.php
@@ -74,6 +74,14 @@ class apc_cache extends cache_api
 		$this->invalidateCache();
 		return true;
 	}
+
+	/**
+	 * {@inheritDoc}
+	 */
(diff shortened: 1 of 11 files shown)


albertlast:
The pg cache keep working,
did i see somewhere on the admin panel the version info (didn't find them)?

For pg/sql/file cache i miss a function/class method to implement house keeping,
idea would be once in a month or so to delete the old entries.

jdarwood007:
Wouldn't be a bad idea, can tie it into a scheduled task to do so.

Its in a few places in the ACP but they have to use the direct loading at this point unless I want to call the loadCacheAccelerator multiple times to get all of that.  This fixes a call I missed in the admin panel and sets it up so we can get the version info easily.  Mostly would be for support needs.

PR #4310, opened 2017-09-17
Change BoardIndex Templates to be more flexible

jdarwood007:
This technically adds a feature, but this feature prevents modifications from needing to modify the template if they need to inject a "fake" board.  This also allows us to more easily handle redirects differently than boards.

@Antes You may want to review this.
@Gwenwyfar Antes thinks you may be interested.

### DIFF
--- a/Themes/default/BoardIndex.template.php
+++ b/Themes/default/BoardIndex.template.php
@@ -132,7 +132,7 @@ function template_main()
 }
 
 /**
- * Outs the board icon for a standard board.
+ * Outputs the board icon for a standard board.
  *
  * @param array $board Current board information.
  */
@@ -145,7 +145,7 @@ function template_bi_board_icon($board)
 }
 
 /**
- * Outs the board icon for a redirect.
+ * Outputs the board icon for a redirect.
  *
  * @param array $board Current board information.
  */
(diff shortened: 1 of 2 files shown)


XinYenFon (approved):

Themes/default/BoardIndex.template.php
```diff
@@ -178,6 +132,140 @@ function template_main()
 }
 
 /**
+ * Outs the board icon for a standard board.
```
Sesquipedalian: Nit: "Outs the..." is weird English. This should probably be "Outputs the..." or (better) "Prints the..."

Themes/default/BoardIndex.template.php
```diff
@@ -178,6 +132,140 @@ function template_main()
...
+	global $txt;
+
+	echo '
+						<p>', comma_format($board['posts']), ' ', $txt['redirects'], '
+						</p>';
+}
+
+/**
+ * Outs the board lastposts for a standard board or a redirect.
+ * This gets hidden on mobile if $board['last_post']['id'] is empty by the main template.
```
Sesquipedalian: Nit: this comment is unclear.

Sesquipedalian (changes requested):
I like the approach. Just requesting a couple adjustments to the comments.

jdarwood007:
Thanks, I like it as well.  SimpleDesk and other mods will be able to easily add "fake" boards now with custom content for the all the columns.  The code is easier to read as well and should be easier on theme developers to make changes to it.

Updated the comments.

Sesquipedalian (approved):

Issue #1571, opened 2014-04-27
Fulltext indexes are supported on mediumtext columns


Arantor:
There is code internally to juggle the body column around based on post size but it refuses to upgrade the column size to mediumtext if there's a fulltext index on it.

Both MyISAM and InnoDB at least in MySQL 5.5 support fulltext indexes on mediumtext columns, very likely that extends to at least MySQL 5.1 and probably even 5.0 (the documentation is somewhat unclear on this point; testing would be required)

Switching it to mediumtext permanently would solve at least one outstanding bug around posts potentially being truncated inappropriately.

Oldiesmann:
The only thing I can really find regarding fulltext indexes on mediumtext columns is this bug report from 2007: http://bugs.mysql.com/bug.php?id=26677. That seems to prove that MySQL 5 does indeed support fulltext indexes on MEDIUMTEXT columns.

However, fulltext indexes are not supported for InnoDB tables prior to 5.6, per the MySQL manual (http://dev.mysql.com/doc/refman/5.5/en/fulltext-search.html ).

Arantor:
Could have sworn it worked on 5.5/InnoDB rather than 5.6... in any case, your installer still brute forces everything to MyISAM anyway currently so what's the problem?

Oldiesmann:
Just making sure we cover everything I guess. Plus, I could have sworn we went through and changed the installer so that you can specify a preferred engine for a table and have it fall back to MyISAM if that engine wasn't available.

Arantor:
Nope, #772 lists that as a thing to be done.

albertlast:
Can we close this issue?
because this min mysql version is to low to look at this topic.

Arantor:
But it isn't. If the installer still forces things to MyISAM, the minimum of 5.0.3 would allow this to simplify all the code away.

And if it goes InnoDB, the existing behaviour is still wrong anyway.

albertlast:
The Installer Force innodb.


________________________________
Von: Arantor <notifications@github.com>
Gesendet: 22.08.2017 7:39 nachm.
An: SimpleMachines/SMF2.1
Cc: albertlast; Comment
Betreff: Re: [SimpleMachines/SMF2.1] Fulltext indexes are supported on mediumtext columns (#1571)


But it isn't. If the installer still forces things to MyISAM, the minimum of 5.0.3 would allow this to simplify all the code away.

And if it goes InnoDB, the existing behaviour is still wrong anyway.

[...]

PR #4311, opened 2017-09-19
Fixes #4000

### LINKED ISSUE #4000: Installer vs Upgrader - DB differences
I have noticed a few of these by chance, so I performed a more rigorous comparison.  

Install a vanilla 2.0, then upgrade it to 2.1.  Install a vanilla 2.1.  Export the data structures & compare.  

How different should the DBs be?  

Differences I am seeing are:
  -  Different defaults - only a few of these
  -  Indexes missing - only a few of these
  -  Differences in signed vs unsigned - only a few of these
  -  Differences in text types - e.g., text vs mediumtext - several of these
  -  Differences in int types - e.g., int vs mediumint - several of these
  -  Differences in int display lengths - e.g., int(3) vs int(4) - **_LOTS_** of these
  -  Differences in index names - **_LOTS_** of these

Suggested actions:
  -  Different defaults **_- need alignment, case-by-case_**
  -  Indexes missing **_- Add the MISSING one..._**
  -  Differences in signed vs unsigned **_- Change to Unsigned..._**
  -  Differences in text types - e.g., text vs mediumtext **_- Unsure, input welcome_**
  -  Differences in int types - e.g., int vs mediumint **_- Unsure, input welcome_**
  -  Differences in int display lengths - e.g., int(3) vs int(4)  **_-  IGNORE_**
  -  Differences in index names  **_-  IGNORE_**

Most of the default differences are NULL vs ''.  Input welcome.

This is all MySQL - I don't have a pg to do an equivalent comparison.  

Thoughts please.  I think MOST of these can be safely ignored, but not all.  

differences-1
differences-2
differences-3
differences-4
differences-5
differences-6

(*** Note: I converted everything to InnoDB so these would not show up as differences...***)

sbulen:
Align column and index definitions across installs & upgrades.

This PR addresses a number of inconsistencies between the databases of an installed 2.1 forum and a forum that was upgraded to 2.1.  I worked with 3 databases:
 - a vanilla 2.1 install
 - a vanilla 2.0.14 install upgraded to 2.1
 - a mature 2.0.14 forum (patched since smf1.1) upgraded to 2.1

Global Changes:
 - Made id_member consistent throughout
 - Made the ipv6 varbinaries consistent throughout
 - Made all indexes align across all 3 DBs listed above.  In general, added indexes that were in the install but missing from the upgrade _& vice-versa_.  In addition, my mature 2.0 forum had several duplicate indexes with slightly different spellings.  I am assuming I'm not alone here - so I added a bunch of index drops to clean those funky alternate spellings up.  
 - Same with defaults - added defaults that were in the install but missing from the upgrade & vice-versa.
 - Aligned data types across install & upgrade.  
 - Wherever there were differences, I used the mature 2.0 forum as a tiebreaker...  

What I did *NOT* do:
 - I did not remove all of the 'unsigned' attributes (sorry, @albertlast !), rather, I made them consistent across installs & upgrades - too many & it felt risky, I felt this should wait for 3.0
 - I did not rename all indexes to make them consistent...  Too many, I felt this should wait for 3.0.  I looked to ensure the index definitions matched, if not the names.  We support old versions of MySQL that did not allow index renames, thus, the fix would have required rebuilding MOST indexes...
 - I did not modify display lengths (e.g., int(10) vs int(11)) - too many, I felt this should wait for 3.0
 - I did not attempt pg at this time, mainly due to familiarity (again, sorry @albertlast ).  Also, we need to agree on this set of changes first...

My two biggest questions:
 - mediumtext vs text...  My old 2.0 forum had MANY fields defined as mediumtext that are text in 2.1.  This PR as submitted issues commands to change those back to TEXT.  I realize there is some history there.  Should I leave these TEXT conversions in there or remove them?  (This would eliminate most of the changes in the Legacy Column section.)
 - My old 2.0 forum had a default value of 1 for id_profile in smf_board_permissions.  2.1 has a value of 0.  Which is correct, 0 or 1?  It is left as 0 at the moment.


other/install_2-1_mysql.sql
```diff
@@ -90,8 +90,8 @@ CREATE TABLE {$db_prefix}ban_groups (
 CREATE TABLE {$db_prefix}ban_items (
   id_ban MEDIUMINT UNSIGNED AUTO_INCREMENT,
   id_ban_group SMALLINT UNSIGNED NOT NULL DEFAULT '0',
-  ip_low VARBINARY(16),
-  ip_high VARBINARY(16),
+  ip_low VARBINARY(16) DEFAULT NULL,
+  ip_high VARBINARY(16) DEFAULT NULL,
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/install_2-1_mysql.sql
```diff
@@ -240,7 +240,7 @@ CREATE TABLE {$db_prefix}log_actions (
   id_log TINYINT UNSIGNED NOT NULL DEFAULT '1',
   log_time INT(10) UNSIGNED NOT NULL DEFAULT '0',
   id_member MEDIUMINT UNSIGNED NOT NULL DEFAULT '0',
-  ip VARBINARY(16),
+  ip VARBINARY(16) DEFAULT NULL,
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/install_2-1_mysql.sql
```diff
@@ -276,7 +277,7 @@ CREATE TABLE {$db_prefix}log_activity (
 CREATE TABLE {$db_prefix}log_banned (
   id_ban_log MEDIUMINT UNSIGNED AUTO_INCREMENT,
   id_member MEDIUMINT UNSIGNED NOT NULL DEFAULT '0',
-  ip VARBINARY(16),
+  ip VARBINARY(16) DEFAULT NULL,
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/install_2-1_mysql.sql
```diff
@@ -335,7 +336,7 @@ CREATE TABLE {$db_prefix}log_errors (
   id_error MEDIUMINT UNSIGNED AUTO_INCREMENT,
   log_time INT(10) UNSIGNED NOT NULL DEFAULT '0',
   id_member MEDIUMINT UNSIGNED NOT NULL DEFAULT '0',
-  ip VARBINARY(16),
+  ip VARBINARY(16) DEFAULT NULL,
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/install_2-1_mysql.sql
```diff
@@ -353,7 +354,7 @@ CREATE TABLE {$db_prefix}log_errors (
 #
 
 CREATE TABLE {$db_prefix}log_floodcontrol (
-  ip VARBINARY(16),
+  ip VARBINARY(16) NOT NULL,
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/install_2-1_mysql.sql
```diff
@@ -422,8 +423,8 @@ CREATE TABLE {$db_prefix}log_online (
   log_time INT(10) NOT NULL DEFAULT '0',
   id_member MEDIUMINT UNSIGNED NOT NULL DEFAULT '0',
   id_spider SMALLINT UNSIGNED NOT NULL DEFAULT '0',
-  ip VARBINARY(16),
-  url VARCHAR(2048) NOT NULL,
+  ip VARBINARY(16) DEFAULT NULL,
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/install_2-1_mysql.sql
```diff
@@ -500,7 +501,7 @@ CREATE TABLE {$db_prefix}log_reported_comments (
   id_report MEDIUMINT NOT NULL DEFAULT '0',
   id_member MEDIUMINT NOT NULL,
   membername VARCHAR(255) NOT NULL DEFAULT '',
-  member_ip VARBINARY(16),
+  member_ip VARBINARY(16) DEFAULT NULL,
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/install_2-1_mysql.sql
```diff
@@ -703,8 +704,8 @@ CREATE TABLE {$db_prefix}members (
   time_offset float NOT NULL DEFAULT '0',
   avatar VARCHAR(255) NOT NULL DEFAULT '',
   usertitle VARCHAR(255) NOT NULL DEFAULT '',
-  member_ip VARBINARY(16),
-  member_ip2 VARBINARY(16),
+  member_ip VARBINARY(16) DEFAULT NULL,
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/install_2-1_mysql.sql
```diff
@@ -703,8 +704,8 @@ CREATE TABLE {$db_prefix}members (
...
+  member_ip2 VARBINARY(16) DEFAULT NULL,
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/install_2-1_mysql.sql
```diff
@@ -747,8 +748,8 @@ CREATE TABLE {$db_prefix}member_logins (
   id_login INT(10) AUTO_INCREMENT,
   id_member MEDIUMINT NOT NULL DEFAULT '0',
   time INT(10) NOT NULL DEFAULT '0',
-  ip VARBINARY(16),
-  ip2 VARBINARY(16),
+  ip VARBINARY(16) DEFAULT NULL,
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/install_2-1_mysql.sql
```diff
@@ -747,8 +748,8 @@ CREATE TABLE {$db_prefix}member_logins (
...
+  ip2 VARBINARY(16) DEFAULT NULL,
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/install_2-1_mysql.sql
```diff
@@ -782,7 +783,7 @@ CREATE TABLE {$db_prefix}messages (
   subject VARCHAR(255) NOT NULL DEFAULT '',
   poster_name VARCHAR(255) NOT NULL DEFAULT '',
   poster_email VARCHAR(255) NOT NULL DEFAULT '',
-  poster_ip VARBINARY(16),
+  poster_ip VARBINARY(16) DEFAULT NULL,
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/upgrade_2-1_mysql.sql
```diff
@@ -41,10 +41,10 @@ ALTER TABLE {$db_prefix}members CHANGE birthdate birthdate date NOT NULL DEFAULT
 ---# Adding login history...
 CREATE TABLE IF NOT EXISTS {$db_prefix}member_logins (
 	id_login INT(10) AUTO_INCREMENT,
-	id_member MEDIUMINT(8) NOT NULL,
-	time INT(10) NOT NULL,
-	ip VARBINARY(16) NOT NULL DEFAULT '',
-	ip2 VARBINARY(16) NOT NULL DEFAULT '',
+	id_member MEDIUMINT(8) NOT NULL DEFAULT '0',
+	time INT(10) NOT NULL DEFAULT '0',
+	ip VARBINARY(16) DEFAULT NULL,
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/upgrade_2-1_mysql.sql
```diff
@@ -41,10 +41,10 @@ ALTER TABLE {$db_prefix}members CHANGE birthdate birthdate date NOT NULL DEFAULT
...
+	ip2 VARBINARY(16) DEFAULT NULL,
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/upgrade_2-1_mysql.sql
```diff
@@ -2082,8 +2083,8 @@ DROP INDEX idx_id_board on {$db_prefix}topics;
 /******************************************************************************/
 ---# Add columns to ban_items
 ALTER TABLE {$db_prefix}ban_items
-ADD COLUMN ip_low varbinary(16),
-ADD COLUMN ip_high varbinary(16);
+ADD COLUMN ip_low varbinary(16) DEFAULT NULL,
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/upgrade_2-1_mysql.sql
```diff
@@ -2082,8 +2083,8 @@ DROP INDEX idx_id_board on {$db_prefix}topics;
...
+ADD COLUMN ip_high varbinary(16) DEFAULT NULL;
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/upgrade_2-1_mysql.sql
```diff
@@ -2135,7 +2136,7 @@ if ($doChange)
 ---#
 
 ---# Add the new one
-ALTER TABLE {$db_prefix}log_actions ADD COLUMN ip VARBINARY(16);
+ALTER TABLE {$db_prefix}log_actions ADD COLUMN ip VARBINARY(16) DEFAULT NULL;
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/upgrade_2-1_mysql.sql
```diff
@@ -2154,7 +2155,7 @@ if ($doChange)
 ---#
 
 ---# Add the new log banned ip
-ALTER TABLE {$db_prefix}log_banned ADD COLUMN ip VARBINARY(16);
+ALTER TABLE {$db_prefix}log_banned ADD COLUMN ip VARBINARY(16) DEFAULT NULL;
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/upgrade_2-1_mysql.sql
```diff
@@ -2173,7 +2174,7 @@ if ($doChange)
 ---#
 
 ---# Add the new ip columns to log errors
-ALTER TABLE {$db_prefix}log_errors ADD COLUMN ip VARBINARY(16);
+ALTER TABLE {$db_prefix}log_errors ADD COLUMN ip VARBINARY(16) DEFAULT NULL;
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/upgrade_2-1_mysql.sql
```diff
@@ -2200,8 +2201,8 @@ if ($doChange)
 
 ---# Add the new ip columns to members
 ALTER TABLE {$db_prefix}members
-ADD COLUMN member_ip VARBINARY(16),
-ADD COLUMN member_ip2 VARBINARY(16);
+ADD COLUMN member_ip VARBINARY(16) DEFAULT NULL,
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/upgrade_2-1_mysql.sql
```diff
@@ -2200,8 +2201,8 @@ if ($doChange)
...
+ADD COLUMN member_ip2 VARBINARY(16) DEFAULT NULL;
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/upgrade_2-1_mysql.sql
```diff
@@ -2253,7 +2254,7 @@ if ($doChange)
 ---#
 
 ---# Add the new ip column to messages
-ALTER TABLE {$db_prefix}messages ADD COLUMN poster_ip VARBINARY(16);
+ALTER TABLE {$db_prefix}messages ADD COLUMN poster_ip VARBINARY(16) DEFAULT NULL;
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/upgrade_2-1_mysql.sql
```diff
@@ -2333,7 +2334,7 @@ if ($doChange)
 ---#
 
 ---# Add the new ip column for log online
-ALTER TABLE {$db_prefix}log_online ADD COLUMN ip VARBINARY(16);
+ALTER TABLE {$db_prefix}log_online ADD COLUMN ip VARBINARY(16) DEFAULT NULL;
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/upgrade_2-1_mysql.sql
```diff
@@ -2352,7 +2353,7 @@ if ($doChange)
 ---#
 
 ---# Add the new ip column for reported comments
-ALTER TABLE {$db_prefix}log_reported_comments ADD COLUMN member_ip VARBINARY(16);
+ALTER TABLE {$db_prefix}log_reported_comments ADD COLUMN member_ip VARBINARY(16) DEFAULT NULL;
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/upgrade_2-1_mysql.sql
```diff
@@ -2374,8 +2375,8 @@ if ($doChange)
 ---#
 
 ---# Add the new ip columns for member logins
-ALTER TABLE {$db_prefix}member_logins ADD COLUMN ip VARBINARY(16);
-ALTER TABLE {$db_prefix}member_logins ADD COLUMN ip2 VARBINARY(16);
+ALTER TABLE {$db_prefix}member_logins ADD COLUMN ip VARBINARY(16) DEFAULT NULL;
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/upgrade_2-1_mysql.sql
```diff
@@ -2374,8 +2375,8 @@ if ($doChange)
...
+ALTER TABLE {$db_prefix}member_logins ADD COLUMN ip2 VARBINARY(16) DEFAULT NULL;
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/upgrade_2-1_mysql.sql
```diff
@@ -2394,7 +2395,7 @@ if ($doChange)
 ---#
 
 ---# Add the new log banned ip
-ALTER TABLE {$db_prefix}log_online ADD COLUMN ip VARBINARY(16);
+ALTER TABLE {$db_prefix}log_online ADD COLUMN ip VARBINARY(16) DEFAULT NULL;
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/install_2-1_mysql.sql
```diff
@@ -90,8 +90,8 @@ CREATE TABLE {$db_prefix}ban_groups (
 CREATE TABLE {$db_prefix}ban_items (
   id_ban MEDIUMINT UNSIGNED AUTO_INCREMENT,
   id_ban_group SMALLINT UNSIGNED NOT NULL DEFAULT '0',
-  ip_low VARBINARY(16),
-  ip_high VARBINARY(16),
+  ip_low VARBINARY(16) DEFAULT NULL,
```
albertlast: a null able field is by sql def default null -> so no need to write this down.

other/install_2-1_mysql.sql
```diff
@@ -266,7 +266,8 @@ CREATE TABLE {$db_prefix}log_activity (
   posts SMALLINT UNSIGNED NOT NULL DEFAULT '0',
   registers SMALLINT UNSIGNED NOT NULL DEFAULT '0',
   most_on SMALLINT UNSIGNED NOT NULL DEFAULT '0',
-  PRIMARY KEY (date)
+  PRIMARY KEY (date),
+  INDEX idx_most_on (most_on)
```
albertlast: i find no case in code where searching for most_on,
so useless index?

other/install_2-1_mysql.sql
```diff
@@ -320,8 +321,8 @@ CREATE TABLE {$db_prefix}log_comments (
 #
 
 CREATE TABLE {$db_prefix}log_digest (
-  id_topic MEDIUMINT UNSIGNED NOT NULL,
-  id_msg INT(10) UNSIGNED NOT NULL,
+  id_topic MEDIUMINT UNSIGNED NOT NULL DEFAULT '0',
+  id_msg INT(10) UNSIGNED NOT NULL DEFAULT '0',
```
albertlast, line 324: default value should be empty,
other wise you can insert data which is not valid.

other/install_2-1_mysql.sql
```diff
@@ -320,8 +321,8 @@ CREATE TABLE {$db_prefix}log_comments (
 #
 
 CREATE TABLE {$db_prefix}log_digest (
-  id_topic MEDIUMINT UNSIGNED NOT NULL,
-  id_msg INT(10) UNSIGNED NOT NULL,
+  id_topic MEDIUMINT UNSIGNED NOT NULL DEFAULT '0',
```
albertlast, line 323: default value should be empty,
other wise you can insert data which is not valid.

other/install_2-1_mysql.sql
```diff
@@ -794,6 +795,7 @@ CREATE TABLE {$db_prefix}messages (
   PRIMARY KEY (id_msg),
   UNIQUE idx_id_board (id_board, id_msg),
   UNIQUE idx_id_member (id_member, id_msg),
+  UNIQUE idx_id_topic (id_topic, id_msg),
```
albertlast: the index INDEX idx_current_topic (id_topic, id_msg, id_member, approved),
already exists so this index is a duplicate (same field in the same order).

other/install_2-1_mysql.sql
```diff
@@ -1181,7 +1183,7 @@ CREATE TABLE {$db_prefix}mentions (
   content_id INT DEFAULT '0',
   content_type VARCHAR(10) DEFAULT '',
   id_mentioned INT DEFAULT 0,
-  id_member INT(10) UNSIGNED NOT NULL DEFAULT 0,
+  id_member MEDIUMINT(10) UNSIGNED NOT NULL DEFAULT 0,
```
albertlast: the size (10) is here wrong when you use mediumint.

other/upgrade_2-1_mysql.sql
```diff
@@ -41,10 +41,10 @@ ALTER TABLE {$db_prefix}members CHANGE birthdate birthdate date NOT NULL DEFAULT
 ---# Adding login history...
 CREATE TABLE IF NOT EXISTS {$db_prefix}member_logins (
 	id_login INT(10) AUTO_INCREMENT,
-	id_member MEDIUMINT(8) NOT NULL,
-	time INT(10) NOT NULL,
-	ip VARBINARY(16) NOT NULL DEFAULT '',
-	ip2 VARBINARY(16) NOT NULL DEFAULT '',
+	id_member MEDIUMINT(8) NOT NULL DEFAULT '0',
```
albertlast: size (8) is here wrong -> remove.

other/upgrade_2-1_mysql.sql
```diff
@@ -1289,7 +1289,7 @@ CREATE TABLE IF NOT EXISTS {$db_prefix}mentions (
 	content_id INT DEFAULT '0',
 	content_type VARCHAR(10) DEFAULT '',
 	id_mentioned INT DEFAULT 0,
-	id_member INT(10) UNSIGNED NOT NULL DEFAULT 0,
+	id_member MEDIUMINT(10) UNSIGNED NOT NULL DEFAULT 0,
```
albertlast: size (10) is here wrong -> remove.

albertlast (changes requested):

other/install_2-1_mysql.sql
Oldiesmann: @albertlast It's used for statistical stuff. Sources/Stats.php lines 92, 188, 612 and 721

other/install_2-1_mysql.sql
sbulen: Is there no need to enforce the UNIQUE constraint?   

This exists in 2.0, so it exists in upgraded forums, which is why I added it back.

other/install_2-1_mysql.sql
sbulen: It's also used in Subs_MembersOnline.php.

other/upgrade_2-1_mysql.sql
sbulen: MEDIUMINT does in fact support display width, see:
https://dev.mysql.com/doc/refman/5.7/en/numeric-type-overview.html

other/install_2-1_mysql.sql
sbulen: OK I'll change all of these.   Some had it, some didn't, all I did was make it consistent one way or the other.  I'll remove.  

It was a 50-50 bet...

albertlast:
When Id_msg is pk field = unique,
Then I can't imagen how pk + any column should not be unique.


Gesendet: 19.09.2017 07:06
An: SimpleMachines/SMF2.1
Cc: albertlast; Mention
Betreff: Re: [SimpleMachines/SMF2.1] Fixes #4000 (#4311)


@sbulen commented on this pull request.

In other/install_2-1_mysql.sql<https://github.com/SimpleMachines/SMF2.1/pull/4311#discussion_r139598124>:

> @@ -794,6 +795,7 @@ CREATE TABLE {$db_prefix}messages (
   PRIMARY KEY (id_msg),
   UNIQUE idx_id_board (id_board, id_msg),
   UNIQUE idx_id_member (id_member, id_msg),
+  UNIQUE idx_id_topic (id_topic, id_msg),


Is there no need to enforce the UNIQUE constraint?

This exists in 2.0, so it exists in upgraded forums, which is why I added it back.

other/install_2-1_mysql.sql
sbulen: Are you sure you want these removed?   They were inserted to bring it consistent with similar columns elsewhere thoughout the DB...   And 2.0, & upgraded DBs...

other/install_2-1_mysql.sql
sbulen: Are you sure you want these removed?   They were inserted to bring it consistent with similar columns elsewhere thoughout the DB...

other/install_2-1_mysql.sql
sbulen: See above.   MEDIUMINT does honor display width.

other/upgrade_2-1_mysql.sql
sbulen: See above.

other/install_2-1_mysql.sql
sbulen: PK + any field = **not necessarily** unique.  Having one field a pk doesn't make any combo unique.  I don't think I follow your logic there.  

2.0 has these 3 UNIQUE constraints.  

I'm not sure I see a reason to remove any of them.  If we DO want to remove them, we must do it across installs & upgrades, not just one or the other.

other/upgrade_2-1_mysql.sql
albertlast: nope doesn't supprt,
it's only ingored so that some person can do mistake without harming ther self:
http://sqlfiddle.com/#!9/51eed/1/0

sbulen:
@albertlast , @Oldiesmann - any input on my 2 biggest questions above?

I will be out of town for a few days & will make the suggested changes & retest when I return.

other/install_2-1_mysql.sql
albertlast: Show me a case where pk field is unique and pk field + any col is not unique.

other/install_2-1_mysql.sql
albertlast: but never in the where statement and never as group by where a index support could be happen -> useless index.

albertlast:
I see no question marke,
what are you two big questions?

Okay now i see them.
Mediumtext stuff i remember this issues: #1571 
but when i look into this topic this issue is more a problem when you use mysql below 5.6...
ther for smf can not use fulltext search with innodb (so only new installation are effected).

im no friend of this default value, when you ask me the right value is in the most case null,
but this is not smf style, in this context i see 0 as commone default value.

other/install_2-1_mysql.sql
sbulen: I see your point...

Should none of the 3 indexes be unique?

The unique was removed from the install, but we would need to take further steps to rebuild it to make upgraded DBs align.

albertlast: I'm realy sure that i'm the person how removed the unique index from installer.
I "see" no issue that the installer is different as the upgrade version,
on the other side some of the stuff could create issues on upgrade progress side (the droping take no time but creation of one can take long time...)

so the change of unique to a normal index can bring up many issue with no benefit.

other/upgrade_2-1_mysql.sql
sbulen: albertlast - 
Yes it is supported.  The documentation is clear about support, and it is also clear it does not truncate.  If you go back to your sqlfiddle example, and change it to int(1), you will get the exact same result.  
I am aiming for consistency here.  I'd hate to remove these display widths only to find some code somewhere that now is off by a few bytes...  
Make sense?  Buddy, pal, friend???

other/install_2-1_mysql.sql
sbulen: @albertlast - wouldn't an index be needed for the MAX(most_on) clauses?  Otherwise full table scans would be needed to find the MAX value, right?

If so, I think it only comes down to how big log_activity gets.  My forum has ~3500 rows, and is ~115K in size.  I believe this is only displayed on the board index.  

Thoughts?   I would lean towards leaving it in given the MAX clause...

other/install_2-1_mysql.sql
sbulen: OK, so I'll leave it as-is, where both sets are UNIQUE and consistent.

sbulen (commented):
Regarding my 2 questions: 

Q1 - Mediumtext - 
I think my biggest concern here is that my 2.0 forum has **24 columns** as MEDIUMTEXT where 2.1 has them as TEXT.  Important columns - like 'body' and 'value' and 'description' and 'message' and 'reason'...  I do not know how these were changed.  A mod?  Was it that way in 1.1?  If this were done for/by some users deliberately, to avoid message truncation, then "fixing" them back to TEXT (as this PR does now...) may truncate data.  

Since this may truncate real data (and it might be unique to me?) I am going to get rid of these TEXT conversions for these 24 columns.  This feels safest.  Feedback definitely welcome.  

Q2 - default of 0 for id_profile - 
I will leave the PR as is.  This is consistent with the 2.1 changes.  For now I will assume it fixed some old 2.0 quirk.

other/upgrade_2-1_mysql.sql
albertlast: when you look across the installer file,
you find no place where mediumint got a size value.
So you leave the existing consistency.

Please keep in mind that the structure of the installer is leading and
upgrader script needs to follow.

other/install_2-1_mysql.sql
albertlast: yes and no
you can use index on max() = yes
but the qry with group by and the other columns should prohibt the index using.

End of the day you can test this by creating the index in you env and use "explain" to look into if the index is used than you know if this realy or not.

sbulen: Removing most_on based on @albertlast 's suggestion.
Looking at the explain plan for the complex queries in question, the index is not used.

other/upgrade_2-1_mysql.sql
sbulen: Yes.... Clearly, SOMEONE went thru and deleted them all...

There are a bunch across the upgrader.  And I can't find any in my old 2.0 environment.  

For consistency, I'll remove all display widths on all mediumints.

other/install_2-1_mysql.sql
sbulen: We touched base & agreed we'll leave these as-is.

other/install_2-1_mysql.sql
albertlast: My understanding was that you revert the change of "UNIQUE idx_id_topic (id_topic, id_msg),"
and keep only the both other unique index.

other/install_2-1_mysql.sql
sbulen: Note on floodcontrol, this was made NOT NULL to be consistent with the logic that rebuilds it in the upgrader.

albertlast: varbinary(16) = ip_field >needs to be null able<, not null is not allowed in any ip field.

other/install_2-1_mysql.sql
sbulen: Can't do that to the upgrader, as you pointed out above.

Two options....     3 unique indexes (as above), or, live with the inconsistency (which is sloppy).  

Which would you prefer?

other/install_2-1_mysql.sql
sbulen: To get the PR thru, I'll make this change....

other/install_2-1_mysql.sql
albertlast: well i describe in the upper comment that is a bad idea to build up a index in the upgrade process,
but to drop doesn't take much time.
So i would take option three: drop the index in the upgrade process, too.

sbulen: Drop it from both the installer and upgrader?

Is the index not needed?

albertlast: The index(UNIQUE idx_id_topic (id_topic, id_msg)) hat possible two function:
- Check if the combination of cols is unique, but we came already to the result that the pk field by his own is already unique -> for this case the index is useless
- Make the access two the col id_topic, id_msg faster, but we got already the index; INDEX idx_current_topic (id_topic, id_msg, id_member, approved) -> for this case already a better index exists

we came to the result both cases where the index could be needed already catch up -> no need for this index.
Yes drop it from installer and upgrader.

sbulen: Ok.  Will do.

other/install_2-1_mysql.sql
sbulen: For the record, since ip is in the primary key, NOT NULL is implicit - it will be enforced by the db engine.  
What this means is that removing NOT NULL will result in no change.

albertlast (approved):

other/upgrade_2-1_mysql.sql
```diff
@@ -2547,11 +2547,16 @@ ALTER TABLE {$db_prefix}log_actions
 ADD INDEX id_topic_id_log (id_topic, id_log);
 ---#
 
----# Updating log_activity
+---# Updating log_activity mostOn
```
XinYenFon: typo :/

sbulen: What is the typo?   

Oddly, over the years, there have been multiple spellings of this index.   Hopefully this cleans up all of them.

sbulen: @Antes ?   What is the typo - please be specific?

XinYenFon: damn my bad actually I read O (capital o) as 0...

sbulen: Thx,

PR #4318, opened 2017-09-22
Added help text for custom require field fix #4315

### LINKED ISSUE #4315: custom profile field, show on registration and require entry
if you require a field to be completed on registration it also seems to require it to be filled out on the profile view. i dont think it should as there is no flag that states it has to be filled out in the profile.

albertlast:
Should fix the issue #4315


Themes/default/languages/Help.english.php
```diff
@@ -623,4 +623,6 @@
 
 $helptxt['enable_sm_stats'] = 'If enabled, this will allow Simple Machines to visit your site once a month to collect basic statistics. This will help us make decisions as to which configurations to optimize the software for. For more information please visit our <a href="https://www.simplemachines.org/about/stats.php" target="_blank">info page</a>.';
 
+$helptxt['field_reg_require'] = 'When the field get require, than is this field also mandatory on profile changes.';
```
frandominguezl: I believe it's not properly worded; should be something like: "If the field is required on registration, it will also be required on profile changes."

frandominguezl (approved):

PR #4324, opened 2017-09-23
Align total progress with updates needed

sbulen:
Upgrader...  If you have had a lot of attachments deleted, or if you have a lot of avatars, max(id_attach) can overstate the number of attachments by as much as 20-30%.  This change makes the progress bar more accurate for this very long-running step, especially for forums with tens of thousands of attachments.


albertlast:
And what is with PostgreSQL? 
Same line of code exists there too

sbulen:
Done.

albertlast (approved):

PR #4327, opened 2017-09-23
Positioning improvements on top_menu and mobile menu changes.

XinYenFon:


Gwenwyfar (commented):
Should move #alerts_menu to max-width 550px, it breaks between 480 and 540.

PR #4328, opened 2017-09-23
Add width limit to the theme

Gwenwyfar:
Plus some small fixes. This prevents the page from becoming hard to read on very large monitors. The limit value is all in one place so it's easy to edit if anyone wishes to.

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -982,13 +982,18 @@ img.sort, .sort {
 
 /* Box-shadow only on this one. */
 #wrapper {
-	margin: 0 auto;
 	width: 90%;
 	clear: both;
 	background: #fff;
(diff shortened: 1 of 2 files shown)


Sesquipedalian:
In 2.0, the admin could set the width to a preferred value. Rather than hard-coding 1200px, perhaps we should bring back that system and use it to set the max-width?

Gwenwyfar:
If you want to, go for it, but I don't think it would make much difference. Also, note that the old selection was for `width`, which is now fixed at a percentage/used through the theme in more than one place. This change only adds a `max-width`. If you bring the option back, it would have to be a definition of value in pixels only, as percentage would be redundant.

Sesquipedalian:
Fair enough. :)

PR #4335, opened 2017-09-25
Allow create table to update table structure

albertlast:
I find the idea on elkart: https://github.com/elkarte/Elkarte/issues/2118

What it do:
When we create table which already exists,
- it's rename the old one (delete the posible existing one)
- create the new table
- copy all data from the old table to the new table where column name are the same

Some issue exists,
~~i got no idea how to test this and~~
the filling from the old table to the new one could take time...

### DIFF
--- a/Sources/DbPackages-mysql.php
+++ b/Sources/DbPackages-mysql.php
@@ -83,6 +83,7 @@ function db_packages_init()
  *  	- 'ignore' will do nothing if the table exists. (And will return true)
  *  	- 'overwrite' will drop any existing table of the same name.
  *  	- 'error' will return false if the table already exists.
+ *		- 'update' will update the table if the table already exists (no change of ai field and only colums with the same name keep the data)
  *
  * @param string $table_name The name of the table to create
  * @param array $columns An array of column info in the specified format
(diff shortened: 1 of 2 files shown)


sbulen:
This will dramatically increase duration of 2.1 upgrades.  And I'm concerned about keeping mod data around...

This is pretty risky this late in the cycle.  Significant testing is needed.

I agree that a more systematic method for applying changes during upgrades is needed.  I suggest exploring options in 3.0.

albertlast:
And why you believe that this function is used for the upgrade process?

This api extension is optional and only desined for mods.

sbulen:
Sorry, missed that...   Guess I have upgrades on my mind...

My primary concern remains testing.

albertlast:
So i tested this pr on my pg and mysql env and it's work.
```
		$fields = array();
		$fields[] = array('name' => 'id_test', 'type' => 'int', 'size' => 10, 'unsigned' => true, 'auto' => true);
		$fields[] = array('name' => 'charfield', 'type' => 'varchar', 'size' => 10);
		$fields[] = array('name' => 'intfield3', 'type' => 'int', 'size' => 10, 'null' => true );
		$fields[] = array('name' => 'intfield5', 'type' => 'int', 'size' => 10, 'null' => true );
		$smcFunc['db_create_table'](
			'{db_prefix}create_table_check', 
			$fields,
			array(array('columns' => array('id_test'), 'type' => 'primary')),
			array(),
			'update'
		);
```
I was able to remove some fields and add new one and the "old columns" keep ther data.
The speed of this progres looks good, too.

albertlast:
I should mention when you add a colum,
the new column need a default value or need to be null able.

PR #4341, opened 2017-09-30
Misc template and css fixes

Gwenwyfar:
See commit descriptions for details of each.


Gwenwyfar:
The formatting commit is the first step of a revision in all of the templates. There is still a lot to be done. I'll also do a second smaller run on formatting. This will likely be the largest commit of the revision.

Issue #4349, opened 2017-10-03
Endless error loop


albertlast:
More or less a reminder for me,
in some cases it's happen that somewhere an error is raised up,
than the error handling try to save this,
but create an own error and then the endless loop runs.

Got some idea how to detect/stop this endless loop,
but i need to create first this case somehow and then look what work better.

PR #4352, opened 2017-10-04
Use perpared query for error db logging

albertlast:
One of my issue in pg world where i could get an error is because the hacking attemp
can create again a hacking attemp reason for #4349

To get a safe way to save the data i invented this new db api which can handle the information in a more nicer way.


Sources/Subs-Db-mysql.php
```diff
@@ -60,6 +60,7 @@ function smf_db_initiate($db_server, $db_name, $db_user, $db_passwd, $db_prefix,
 			'db_mb4'                    => false,
 			'db_ping'                   => 'mysqli_ping',
 			'db_fetch_all'              => 'smf_db_fetch_all',
+			'db_error_insert' => 'smf_db_error_insert',
```
frandominguezl: Can you keep the indentation?

albertlast: when you look across the different db file than you notice that the laste line is the correct one.
But i align the subs db one now.

PR #4353, opened 2017-10-08
Better Session Cookie handling fix #4297

### LINKED ISSUE #4297: 3 Login Errors
When there is a cookie/session issue, trying to login gives inappropriate error messages.  

To reproduce:
(1) Under Admin | Maintenance | Server Settings | Cookies & Sessions, change the value for "Enable Local Storage of Cookies".   (Changing this value properly affects _future_ cookie behavior, but screws up _active_ sessions...)
(2) Log out.
(3) Try to logon 3 times in rapid succession.  Quick is important.  

By the way, you are now LOCKED OUT...  You must delete your SMF cookie to log back in.  

The errors:  
(1)  When logging on, you get "Your session timed out when posting.  Please go back and try again."  Which makes no sense...  
login-1

(2)  On your 3rd attempt, it does not properly display the error message.  When trying to report "Sorry, but you've reached your login attempts threshold for account xxxxxx. Please wait 30 seconds and try again later.", instead, all you get is:
login-2

(3) The code checks for 5 login attempts.  Why did it fail after 3?  The problem is that validatePasswordFlood() is called from multiple places during login, doubling up on the count of attempts.   Your 3rd attempt internally is 5/6 due to the double-counting.   Not a huge issue, but an indication that some streamlining is in order.  

I considered logging these as 3 separate issues.  If you want me to do so I'll split them up.  

I do think this is kind of important, as it appears to me that there are certain admin tasks & settings updates that will basically lock folks out...  The only way to recover is to delete the cookie.

albertlast:
Since i saw for #4297 no pr and was able to produce the error number one,
i created this pr.

about case one,
i was only to create this case when i jump from my mysql env to pg and the other way around.

Main issue here was that validatePasswordFlood create a new session_id which get us to the point where
login session from the form and the actual running one are different.

after fixing this,
the next problem was that the load.php got the old cookie data and
recognized that member_id  + password doesn't match,
so it setted the $id_member at 0 without empty the usersetting and empty the cookie (so that we don't in the issue on the next call)
Load 472 - 476 handle now this case

Such for "backup" i also place another check in 562 of load that we can be sure that $user_setting is empty.

@sbulen do check your case

Also this pr fix the other two cases in the issue (they based on the first issue)

Side note, when you got multiple env like me,
it's important the cookie name are different OR the cookie setting are the same.

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -469,12 +469,17 @@ function loadUserSettings()
 
 			// Wrong password or not activated - either way, you're going nowhere.
 			$id_member = $check && ($user_settings['is_activated'] == 1 || $user_settings['is_activated'] == 11) ? (int) $user_settings['id_member'] : 0;
+			if ($id_member === 0 && isset($_COOKIE[$cookiename]))
+			{
+				setLoginCookie(-3600, 0);
+				$user_settings = array();
+			}
 		}
 		else
 			$id_member = 0;
 
 		// If we no longer have the member maybe they're being all hackey, stop brute force!
-		if (!$id_member)
+		if (!empty($id_member))
 		{
 			require_once($sourcedir . '/LogInOut.php');
 			validatePasswordFlood(
@@ -553,6 +558,10 @@ function loadUserSettings()
 
 			if ($row['total'] > 0 && !in_array($action, array('profile', 'logout')) || ($action == 'profile' && $area != 'tfasetup'))
 				redirectexit('action=profile;area=tfasetup;forced');
+		}
+		else if (!empty($user_settings) && empty($id_member))
+		{
+			$user_settings = array();
 		}
 	}
 
--- a/Sources/LogInOut.php
+++ b/Sources/LogInOut.php
@@ -809,9 +809,6 @@ function validatePasswordFlood($id_member, $member_name, $password_flood_value =
 	// Only if they're not validating for 2FA
 	if (!$tfa)
 	{
-		require_once($sourcedir . '/Subs-Auth.php');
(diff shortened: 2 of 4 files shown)


sbulen:
Yes, I have been looking into this.  I think your solutions will work.  I will test.  

I have been trying to get to the root cause.  I believe the source of the issue is with the logout logic leaving the cookie in a weird state.   Your changes here catch that weird state & address it.  But I have been trying to figure out exactly how it got so weird...

sbulen:
Albert - 
These fixes work for the admin, but for some reason do not protect end users.  I'm worried more about the broader impact on end users after admin updates.  

To reproduce:  
Create a user, george
Logon as george in one browser (Firefox)
Logon as your admin user in a completely separate browser (chrome)
Toggle the localCookie setting as admin, & SAVE the update (chrome)
Logout george (Firefox)
Try to logon as george (Firefox)

As before, george is locked out until he deletes his SMF cookie.  The message is better, though:

image

I still think we need to find out exactly why logging out screws up the cookie.

albertlast:
I'm very confident that i already describe the issue: https://github.com/SimpleMachines/SMF2.1/issues/4297#issuecomment-335015411

albertlast:
@sbulen try this
but you need first the new version of the cookie -> delete the login cookie of george in your browser 
and login new.

sbulen:
This version tests OK for both the end user (george) & the admin.  
This PR appears to properly handle broken cookies/sessions.

I am still curious as to how it was broken - e.g., how did the password go bad?  I have a theory, which I'll document over in Issue #4297 .

PR #4356, opened 2017-10-14
Cookie preg_match should check for json, not serialized strings

sbulen:
In the cookie/session handling logic, cookie values were being checked vs serialized strings; they need to be checked for json encoded strings.   Found & addressed 2 instances of preg_matches that needed to be updated

Two other changes:
 - No need to de-serialize the cookie string.  It can't get that far...
 - During login, session checks should not default to normal post/get session checks.  You're not posting a message, you're trying to login & create a new session.  

This is the last PR to address #4297 (login failures after cookie setting changes).  There has been existing logic in setLoginCookie to detect cookie state changes, but it was not being invoked due to the preg_match fail.  This is why the cookies/sessions were not re-established properly after cookie settings were changed by the admin.

### DIFF
--- a/Sources/LogInOut.php
+++ b/Sources/LogInOut.php
@@ -173,7 +173,7 @@ function Login2()
 		redirectexit();
 
 	// Are you guessing with a script?
-	checkSession('login');
+	checkSession();
 	validateToken('login');
 	spamProtection('login');
 
--- a/Sources/LogInOut.php
+++ b/Sources/LogInOut.php
@@ -169,7 +169,7 @@ function Login2()
 		redirectexit();
 
 	// Are you guessing with a script?
-	checkSession();
+	checkSession('login');
 	validateToken('login');
 	spamProtection('login');
 
--- a/Sources/LogInOut.php
+++ b/Sources/LogInOut.php
@@ -97,17 +97,25 @@ function Login2()
 
 	if (isset($_GET['sa']) && $_GET['sa'] == 'salt' && !$user_info['is_guest'])
 	{
+		// First check for 2.1 json-format cookie in $_COOKIE
 		if (isset($_COOKIE[$cookiename]) && preg_match('~^\[\d{1,7},"([a-fA-F0-9]{128})?",\d{1,14}(,[0-3])?\]$~', $_COOKIE[$cookiename]) === 1)
 		{
 			list (,, $timeout) = $smcFunc['json_decode']($_COOKIE[$cookiename], true);
 		}
-		elseif (isset($_SESSION['login_' . $cookiename]))
+		// Try checking for 2.1 json-format cookie in $_SESSION
+		elseif (isset($_SESSION['login_' . $cookiename]) && preg_match('~^\[\d{1,7},"([a-fA-F0-9]{128})?",\d{1,14}(,[0-3])?\]$~', $_SESSION['login_' . $cookiename]) === 1)
 		{
-			list (,, $timeout) = $smcFunc['json_decode']($_SESSION['login_' . $cookiename]);
-
-			// Try for old format
-			if (is_null($timeout))
-				list (,, $timeout) = safe_unserialize($_SESSION['login_' . $cookiename]);
+			list (,, $timeout) = $smcFunc['json_decode']($_SESSION['login_' . $cookiename], true);
+		}
+		// Next, try checking for 2.0 serialized string cookie in $_COOKIE
+		elseif (isset($_COOKIE[$cookiename]) && preg_match('~^a:[34]:\{i:0;i:\d{1,7};i:1;s:(0|128):"([a-fA-F0-9]{128})?";i:2;[id]:\d{1,14};(i:3;i:\d;)?\}$~', $_COOKIE[$cookiename]) === 1)
+		{
+			list (,, $timeout) = safe_unserialize($_COOKIE[$cookiename]);				
+		}
+		// Last, see if you need to fall back on checking for 2.0 serialized string cookie in $_SESSION
+		elseif (isset($_SESSION['login_' . $cookiename]) && preg_match('~^a:[34]:\{i:0;i:\d{1,7};i:1;s:(0|128):"([a-fA-F0-9]{128})?";i:2;[id]:\d{1,14};(i:3;i:\d;)?\}$~', $_SESSION['login_' . $cookiename]) === 1)
+		{
+			list (,, $timeout) = safe_unserialize($_SESSION['login_' . $cookiename]);				
 		}
 		else
 			trigger_error('Login2(): Cannot be logged in without a session or cookie', E_USER_ERROR);
--- a/Sources/LogInOut.php
+++ b/Sources/LogInOut.php
@@ -97,13 +97,9 @@ function Login2()
 
 	if (isset($_GET['sa']) && $_GET['sa'] == 'salt' && !$user_info['is_guest'])
 	{
-		if (isset($_COOKIE[$cookiename]) && preg_match('~^a:[34]:\{i:0;i:\d{1,7};i:1;s:(0|128):"([a-fA-F0-9]{128})?";i:2;[id]:\d{1,14};(i:3;i:\d;)?\}$~', $_COOKIE[$cookiename]) === 1)
+		if (isset($_COOKIE[$cookiename]) && preg_match('~^\[\d{1,7},"([a-fA-F0-9]{128})?",\d{1,14}(,[0-3])?\]$~', $_COOKIE[$cookiename]) === 1)
 		{
 			list (,, $timeout) = $smcFunc['json_decode']($_COOKIE[$cookiename], true);
-
-			// That didn't work... Maybe it's using serialize?
-			if (is_null($timeout))
-				list (,, $timeout) = safe_unserialize($_COOKIE[$cookiename]);
 		}
 		elseif (isset($_SESSION['login_' . $cookiename]))
 		{
--- a/Sources/Security.php
+++ b/Sources/Security.php
@@ -593,11 +593,7 @@ function isBannedEmail($email, $restriction, $error)
  * Will check GET, POST, or REQUEST depending on the passed type.
  * Also optionally checks the referring action if passed. (note that the referring action must be by GET.)
  *
- * @param string $type The type of check (post, get, request, or login), where
- *      post - checks for a session timeout
- *      get - performs a strict session check
- *      request - performs a strict session check
- *      login - does none of the above, as they are not appropriate for logging in
+ * @param string $type The type of check (post, get, request)
  * @param string $from_action The action this is coming from
  * @param bool $is_fatal Whether to die with a fatal error if the check fails
  * @return string The error message if is_fatal is false.
--- a/Sources/Security.php
+++ b/Sources/Security.php
@@ -593,7 +593,11 @@ function isBannedEmail($email, $restriction, $error)
  * Will check GET, POST, or REQUEST depending on the passed type.
  * Also optionally checks the referring action if passed. (note that the referring action must be by GET.)
  *
- * @param string $type The type of check (post, get, request)
+ * @param string $type The type of check (post, get, request, or login), where
+ *      post - checks for a session timeout
+ *      get - performs a strict session check
+ *      request - performs a strict session check
+ *      login - does none of the above, as they are not appropriate for logging in
  * @param string $from_action The action this is coming from
  * @param bool $is_fatal Whether to die with a fatal error if the check fails
  * @return string The error message if is_fatal is false.
--- a/Sources/Subs-Auth.php
+++ b/Sources/Subs-Auth.php
@@ -38,9 +38,18 @@ function setLoginCookie($cookie_length, $id, $password = '')
 
 	// The cookie may already exist, and have been set with different options.
 	$cookie_state = (empty($modSettings['localCookies']) ? 0 : 1) | (empty($modSettings['globalCookies']) ? 0 : 2);
-	if (isset($_COOKIE[$cookiename]) && preg_match('~^\[\d{1,7},"([a-fA-F0-9]{128})?",\d{1,14}(,[0-3])?\]$~', $_COOKIE[$cookiename]) === 1)
+	if (isset($_COOKIE[$cookiename]))
 	{
-		$array = $smcFunc['json_decode']($_COOKIE[$cookiename], true);
+		// First check for 2.1 json-format cookie
+		if (preg_match('~^\[\d{1,7},"([a-fA-F0-9]{128})?",\d{1,14}(,[0-3])?\]$~', $_COOKIE[$cookiename]) === 1)
+		{
+			$array = $smcFunc['json_decode']($_COOKIE[$cookiename], true);		
+		}
+		// If necessary, fall back on checking for 2.0 serialized string cookie
+		elseif (preg_match('~^a:[34]:\{i:0;i:\d{1,7};i:1;s:(0|128):"([a-fA-F0-9]{128})?";i:2;[id]:\d{1,14};(i:3;i:\d;)?\}$~', $_COOKIE[$cookiename]) === 1)
+		{
+			$array = safe_unserialize($_COOKIE[$cookiename]);
+		}
 
 		// Out with the old, in with the new!
 		if (isset($array[3]) && $array[3] != $cookie_state)
--- a/Sources/Subs-Auth.php
+++ b/Sources/Subs-Auth.php
@@ -38,13 +38,9 @@ function setLoginCookie($cookie_length, $id, $password = '')
 
 	// The cookie may already exist, and have been set with different options.
 	$cookie_state = (empty($modSettings['localCookies']) ? 0 : 1) | (empty($modSettings['globalCookies']) ? 0 : 2);
-	if (isset($_COOKIE[$cookiename]) && preg_match('~^a:[34]:\{i:0;i:\d{1,7};i:1;s:(0|128):"([a-fA-F0-9]{128})?";i:2;[id]:\d{1,14};(i:3;i:\d;)?\}$~', $_COOKIE[$cookiename]) === 1)
+	if (isset($_COOKIE[$cookiename]) && preg_match('~^\[\d{1,7},"([a-fA-F0-9]{128})?",\d{1,14}(,[0-3])?\]$~', $_COOKIE[$cookiename]) === 1)
 	{
 		$array = $smcFunc['json_decode']($_COOKIE[$cookiename], true);
-
-		// Legacy format
-		if (is_null($array))
-			$array = safe_unserialize($_COOKIE[$cookiename]);
 
 		// Out with the old, in with the new!
 		if (isset($array[3]) && $array[3] != $cookie_state)


jdarwood007:
Upgrades from 2.0 to 2.1 will cause all users to log out with this change as the stored data will be different.  I think it would be better if we tried the json first and then fall back to trying to get it via serialized data.  Sometime in the future we can completely remove it.

sbulen:
The preg_match had been written to look at the serialized string...   I just rewrote it to look at the json string...  The problem is that the preg_match is for one or the other.  

I.e., it will never get to the serialized string check if the preg_match filter just before it checks for json.  

Or, we could get rid of the preg_match altogether?   It's meant to ensure the cookie looks OK.

Sources/LogInOut.php
```diff
@@ -169,7 +165,7 @@ function Login2()
 		redirectexit();
 
 	// Are you guessing with a script?
-	checkSession();
+	checkSession('login');
```
albertlast: type login didn't exist: https://github.com/SimpleMachines/SMF2.1/blob/e750119e8d604456faa760c7a6f5fdc7050a60f4/Sources/Security.php#L596

albertlast (changes requested):

Sources/LogInOut.php
sbulen: Correct.   Any value other than post or get.  Simply bypassing those post/get edits that are not appropriate for logging in.

albertlast: I see no need for this,
otherwise you need to extend the checksession function,
because you use a type which didn't exists -> wrong.

sbulen: Maybe just add a comment at the beginning of checksession to that effect?

The problem is that with no parameter, it assumes you are making a post to the forum.

Sources/LogInOut.php
```diff
@@ -97,13 +97,9 @@ function Login2()
 
 	if (isset($_GET['sa']) && $_GET['sa'] == 'salt' && !$user_info['is_guest'])
 	{
-		if (isset($_COOKIE[$cookiename]) && preg_match('~^a:[34]:\{i:0;i:\d{1,7};i:1;s:(0|128):"([a-fA-F0-9]{128})?";i:2;[id]:\d{1,14};(i:3;i:\d;)?\}$~', $_COOKIE[$cookiename]) === 1)
+		if (isset($_COOKIE[$cookiename]) && preg_match('~^\[\d{1,7},"([a-fA-F0-9]{128})?",\d{1,14}(,[0-3])?\]$~', $_COOKIE[$cookiename]) === 1)
```
albertlast: both pattern check, the old and new one, are wrong.
Since my update the cookie look like so:
`{"0":1,"1":"7a30e3ff27d981d93f57fb0aaa2065b96c3144535fbbeebb3aa1e9fa7e68d771e127f8283fb66ca828e8cdf543ff9b14eccd5a5b731c8c885f683a266ce4e217","2":1697263514,"3":3,"path":"\/SMF2.1\/"}`

Sources/LogInOut.php
albertlast: because you do a post to the forum:
grafik

jdarwood007:
@sbulen ,

Since in SMF 2.0 it is serialized, upon upgrade to 2.0, all users will be in effect logged out due to the cookie changing formats.  I was suggesting we support the old method as a fallback for now.  The old method doesn't need to support TFA, just cookies for login session.  Cookies will get updated over time to use the json format.

sbulen:
@jdarwood007 - Understood.  I was jut pointing out that due to the preg_match, the "fallback" would never be reached.  We would have to restructure the logic a little bit, maybe even remove the preg_match altogether.  

I was asking if you thought it was OK to remove the preg_match.  

Note that  @albertlast has removed the preg_match in his PR #4358, which is intended to address issues from PR #4353.  My issue with albertlast's approach is exactly what you point out - he has redefined the data portion of the SMF login cookie. 

With the data portion of the cookie redefined, users cannot logon in 2.1 post migration without deleting their cookies.  

The root cause of the original problem ( #4297 ) is in fact this issue with the preg_match.   This logic worked great in 2.0, and the only thing that broke it in 2.1 was the fact that the preg_match was still looking at the data portion of the cookie as if it were a serialized string, when it is now json.  Changing or removing the preg_match is all that is needed to fix #4297 .  

This is a challenge with Github development...  Two people can work on different incompatible fixes to the same problem at the same time.  

I'll leave it to the dev team to determine how to proceed.  I'll withdraw this PR if requested.

Sources/LogInOut.php
sbulen: albertlast - I didn't pick up on the fact that you redefined the data portion of the cookie.  I think you should revisit that decision.

Sources/LogInOut.php
sbulen: The post is of the logon form, not a post of a new message.  Edits like checking for session timeout are intended for posting new messages, and do not make sense during login.

jdarwood007:
Honestly, I would leave the preg_match as its a good sanity check prior to attempting to load the cookie.  It should properly read the cookie for all cases, which can be difficult but this makes it safer since it is handling user input

sbulen:
Thanks @jdarwood007 .   If the dev team wants to keep the old cookie format, I'll mod this PR.  

If the dev team wants to rewrite the cookie format, all such changes should be in PR #4353 & PR #4358 , and this PR should be closed.

sbulen:
I followed the suggestions here & made the changes.   All tests well, but on the old cookie format.

albertlast:
> With the data portion of the cookie redefined, users cannot logon in 2.1 post migration without deleting their cookies.

Sure they can, it will be migrated like the old version

> Honestly, I would leave the preg_match as its a good sanity check prior to attempting to load the cookie. It should properly read the cookie for all cases, which can be difficult but this makes it safer since it is handling user input

This make no sense: When the user got a valid cookie we check,
when not valid cookie we do nothing???

Sources/LogInOut.php
albertlast: We are here in login2, this function is only called at login and this is a post (see my screenshot)

jdarwood007:
@albertlast 
If their cookie doesn't match what we know to be a valid json string for a login session and doesn't match a valid serialized string for a login session, we can assume that the cookie is dirty and contains input we don't want to try to parse.  We know exactly what that cookie should contain and thats why I think its safe and easy here to do the check.

albertlast:
The cookie is at the point already parsed and
ther is no check "when pattern is wrong fix the cookie".
This is here the right place to do any check and fix it when possible.

Sources/LogInOut.php
```diff
@@ -97,21 +97,25 @@ function Login2()
 
 	if (isset($_GET['sa']) && $_GET['sa'] == 'salt' && !$user_info['is_guest'])
 	{
-		if (isset($_COOKIE[$cookiename]) && preg_match('~^a:[34]:\{i:0;i:\d{1,7};i:1;s:(0|128):"([a-fA-F0-9]{128})?";i:2;[id]:\d{1,14};(i:3;i:\d;)?\}$~', $_COOKIE[$cookiename]) === 1)
+		// First check for 2.1 json-format cookie in $_COOKIE
+		if (isset($_COOKIE[$cookiename]) && preg_match('~^\[\d{1,7},"([a-fA-F0-9]{128})?",\d{1,14}(,[0-3])?\]$~', $_COOKIE[$cookiename]) === 1)
```
albertlast, line 101: As fare my test goes the regex pattern didn't work for json...

sbulen: I've tested it repeatedly.  Works fine.    The difference of is probably that you're using your redefined cookies.

albertlast: http://www.phpliveregex.com/p/lEX here my test case

sbulen: Here is a screenshot of the point in question.  Note the format of the data portion of the cookie.  Note also that, in the debugger, that line of code would not have been executed without a successful preg_match.

Note also that this is the logic for detecting cookie state changes - the heart of the original issue. 
This logic worked fine in 2.0.  And it works fine in 2.1 after updating the preg_match.   No need to redefine the cookie.

image

albertlast:
This is no json, please use a json when you try
To define a pattern for this.


Gesendet: 16.10.2017 6:39 vorm.
An: SimpleMachines/SMF2.1
Cc: albertlast; Mention
Betreff: Re: [SimpleMachines/SMF2.1] Cookie preg_match should check for json, not serialized strings (#4356)


@sbulen commented on this pull request.

In Sources/LogInOut.php<https://github.com/SimpleMachines/SMF2.1/pull/4356#discussion_r144753763>:

> @@ -97,21 +97,25 @@ function Login2()

        if (isset($_GET['sa']) && $_GET['sa'] == 'salt' && !$user_info['is_guest'])
        {
-               if (isset($_COOKIE[$cookiename]) && preg_match('~^a:[34]:\{i:0;i:\d{1,7};i:1;s:(0|128):"([a-fA-F0-9]{128})?";i:2;[id]:\d{1,14};(i:3;i:\d;)?\}$~', $_COOKIE[$cookiename]) === 1)
+               // First check for 2.1 json-format cookie in $_COOKIE
+               if (isset($_COOKIE[$cookiename]) && preg_match('~^\[\d{1,7},"([a-fA-F0-9]{128})?",\d{1,14}(,[0-3])?\]$~', $_COOKIE[$cookiename]) === 1)


Here is a screenshot of the point in question. Note the format of the cookie. Note also that, in the debugger, that line of code would not have been executed without a successful preg_match.

Note also that this is the logic for detecting cookie state changes - the heart of the original issue.

[image]<https://user-images.githubusercontent.com/23568484/31595528-4564755e-b1f1-11e7-9801-c05099f31f00.png>

sbulen:
https://en.functions-online.com/json_encode.html?command={"value":"array(7,'abcd0123',44444,2)"}

https://en.functions-online.com/json_decode.html?command={"json":"[7,\"abcd0123\",44444,2]"}

albertlast:
This is a json with a key where you got an array,
Which is a valid json,
In your screenshot you doesn't got a json with a key.


Gesendet: 16.10.2017 12:05 nachm.
An: SimpleMachines/SMF2.1
Cc: albertlast; Mention
Betreff: Re: [SimpleMachines/SMF2.1] Cookie preg_match should check for json, not serialized strings (#4356)


https://en.functions-online.com/json_encode.html?command={"value":"array(7,'abcd0123',44444,2)"}<https://en.functions-online.com/json_encode.html?command=%7B%22value%22:%22array(7,'abcd0123',44444,2)%22%7D>

https://en.functions-online.com/json_decode.html?command={"json":"[7,\"abcd0123\",44444,2]"}<https://en.functions-online.com/json_decode.html?command=%7B%22json%22:%22%5B7,%5C%22abcd0123%5C%22,44444,2%5D%22%7D>

sbulen:
https://en.functions-online.com/json_encode.html?command={"value":"array(7,'abcd0123',44444,2)"}

https://en.functions-online.com/json_decode.html?command={"json":"[7,\"abcd0123\",44444,2]"}

sbulen:
This emulates the code exactly, as is demonstrated in my screenshot.

albertlast:
And why your are sure that this format is used and not this:
`{"0":7,"1":"abcd0123","2":44444,"3":2}`
Which create the same output in your test case,
what only show that it wrong from the start to make pattern check.

sbulen:
(1) The debugger screenshot above shows what is actually being returned.  
(2) The utility which I provided the link to is in alignment with the screenshot.  
(3) Testing reveals the proper 2.0 behavior has been restored.  The change made the issues go away.

albertlast:
and the question keeps up, how you safe this json array is used instead of a json object.

sbulen:
I did not alter the cookie in any way.  What is passed to the cookie is a simple array, as it has always been, in 2.0 & 2.1.  As noted before, I do not think it was necessary or desirable to redefine the cookie contents.

sbulen:
Sorry the links above are wonky.  I guess the site is having ssl issues.  Screenshots here are in alignment with SMF behavior:
image
image

albertlast:
The links work fine when you replace en with www, so i was able to test this.
And based on this i came up with the question how you safe that the json keeps an array json and
get not a object json.

sbulen:
By reading the code, and seeing it is being passed a simple array, not an object.  

Further, by looking at it in the debugger and observing the behavior, confirming the above.   

Finally, by testing, and confirming that the logic for cookie state changes is once again operational.

albertlast:
this is not a good check, when a mod write a key as a string like me with "path" than the thing get a object,
so this is very danger zone where you move.

jdarwood007:
When in doubt, check the RFC: https://tools.ietf.org/html/rfc7159
Check Page 11 for examples.  Both are valid, but we shouldn't be using a object here, should be an array

albertlast:
Can you please cover your "should" with reason?
On php side it's not important which kind of json this is,
because we get this anyway as array.
which get me back to point this pattern check is nitpicking without sense.

Sesquipedalian:
It seems like the changes to the cookie format in #4353 are creating more problems than they solve. After some discussion by the dev team over on simplemachines.org, it seems like the best course is to revert #4353 and start over at solving #4297.

sbulen:
I believe this one should have been kept, unless there are other concerns.  It addresses issue #4297.

The commit that adds 'login' to the checksession call fixes the erroneous messages.

The fixes to the preg_match solve the errors related to sessions & cookies.

Sesquipedalian:
These changes are independent from those in #4353?

sbulen:
Yes.  Unfortunately we both worked on solutions to #4297 independently.

Sesquipedalian:
Just read over the comments again and I now see that, yes, this is independent. Reopening.

sbulen:
OK, as promised I did another round of testing.  Results were good.  Specifically:
  -  All symptoms of #4297 have been addressed
  -  The cookie is now properly cleared upon logout when cookie state changes (localCookie or globalCookie setting changed)

This round of testing was on a UNIX host, all prior rounds of tests were on Windows.  

Any questions, let me know.

albertlast:
As i saying in one of my issues,
you don't check the session_var and sc value any more which is provided by the login form,
you should fix this.

sbulen:
I understand you don't agree with the approach of passing a 'login' value to checkSession.  I will defer to the dev team's input on that.  

FYI - checkSession is called ~200x throughout SMF.  It does MANY checks on the session, including timeout, etc.   The timeout check & strict comparison make no logical sense in the middle of a login, hence the completely non-sensical messages reported in #4297 .  Passing 'login' bypasses these checks _alone_, and allows checkSession to continue with all other session checks.   

Again, I will defer to the dev team's input.

albertlast:
This stuff can't be checked 200x because it's only be part of the login form.

Since you mention that the issue https://github.com/SimpleMachines/SMF2.1/issues/4366 is solved by this pr and
this is not fixed -> you testes are incomplete.

sbulen:
As noted above, I confirmed the cookie is now properly being destroyed after cookie state changes.  That issue is confirmed to be adfressed.

albertlast:
#4366 is not talking about this,
the case is here that a board is moved from /smf/ to /smf/sub,
and this case is not coverd by your pr (i already tested this with your pr):
grafik

sbulen:
Ok.  No, this PR does not cover actual board relocations.

albertlast:
I know,
but than please edit you comment in the issue or
cover relocation with this pr.

albertlast:
Good than only the missing check of session_var and sc is in my eyes.
The test case here for is,
open the login form, edit in browser ide (f12) the numbers in the hidden input and
notice that the login is successfull -> no check are done.
unchaged:
grafik

changed:
grafik
--> successfull login -> checks are broken (this behavior is not wanted)

in the "live dev" version you got an error (this is right behavior)

emanuele45:
sbulen wrote on 24/10/17 07:05:
>
> I understand you don't agree with the approach of passing a 'login'
> value to checkSession. I will defer to the dev team's input on that.
>
> FYI - checkSession is called ~200x throughout SMF. It does MANY checks
> on the session, including timeout, etc. The timeout check & strict
> comparison make no logical sense in the middle of a login, hence the
> completely non-sensical messages reported in #4297
> <https://github.com/SimpleMachines/SMF2.1/issues/4297> . Passing
> 'logon' bypasses these checks alone, which makes sense to me because
> they are illogical during login. Note that checkSession continues with
> all other session checks.
>

Out of curiosity: why are these checks illogical to you during the login?

albertlast:
Since you change only cookie stuff and doesn't mention that you want to change the login process,
i expected that everything which is working before keeps working after this pr.

And why i raised attention about this topic,
because the stuff of checking this var doesn't work anymore.

sbulen:
Look at the original issue, #4297.  You would get locked in an unending 'session timed out' error loop while trying to login.   That is what was illogical - you didn't have a session to timeout.

After changing to 'login', you would at least get a more appropriate message informing you your cookie had issues.   

(And converting the preg_match to json fixed the actual cookie issue.)

sbulen:
I had been tempted to use 'get' instead of 'login'.  One could argue that existing edit made sense.  Again, I'm deferring to dev input.

Sources/LogInOut.php
Sesquipedalian: What happens if you leave this as `checkSession();`?

sbulen: Assume for a second that we split this PR into two PRs, one that addressed the preg_match issue & this change.  I will first answer as if this change were made in isolation:  WITHOUT this change, the issues over in #4297 appear - you get the 'session timed out' error.  WITH this change, you get an appropriate 'there is a problem with your cookie' message.  

My reasoning for this change: the first edit in checkSession() compares the session var to the prior value - which will fail, because you were logged off, i.e., you don't have a prior session.  That's why I think those first two checks in checkSession() don't make sense during logon.  

Since fixing the pregmatch pattern solves the problem (the other change in this PR), leaving this at checkSession() actually makes no difference.  The problem is fixed.  But I made this change because, as indicated above, I believe that those first edits at the top of checkSession() do not apply for logins and may bite us under other circumstances down the road.   

Note that checkSession() is invoked ~200 times throughout the application, under all sets of circumstances (message posts, admin tasks, etc...).   I think the initial edits at the top of checkSession for stale sessions make more sense for message posts & setting updates, but not for logging in.

Sesquipedalian: Thanks for the explanation, @sbulen. I agree with @albertlast's reading of the situation regarding `checkSession()`. If this PR works without changing `checkSession()` to `checkSession('login')`, I would much prefer to leave it as it was.

sbulen: Done.

Sesquipedalian: Thanks

Sesquipedalian:
With `checkSession()` reverted back to the original, I like the way the logic is laid out in this solution to #4297. But before I merge it, I also want to read @albertlast's answers to my questions on #4373. I also like some of the things he is doing there to address the other issues he has identified. I may ask you both to make a few more tweaks to your PRs so that we can merge in the best from both of them.

albertlast:
With the reverting of checkSession() you got the problem,
that the message error session timeout... apears again.

Sesquipedalian:
> With the reverting of checkSession() you got the problem,
that the message error session timeout... apears again.

Indeed it does. Perhaps I wasn't clear, but that's actually what I was asking about, @sbulen: Would these incorrect error messages still show up if we don't passing a bogus parameter to `checkSession()`?

Since it does, we need to go back to the drawing board. Bypassing the security checks here using `checkSession('login')` isn't a good option, but neither is removing `setLoginCookie(-3600, 0)` from `validatePasswordFlood()`, as in #4373.

The root issue must be something deeper. In 2.0.x, `checkSession()` is called at this point in `Login2()` and `setLoginCookie(-3600, 0)` is called near the beginning of `validatePasswordFlood()`, and no problems arise. Something else in the logic flow must have changed in order to cause this mess.

sbulen:
I'll take a peek over the next few days - very busy at work.   My initial thought is that other changes in the interim have changed behavior.

sbulen:
I have not performed a merge on this branch recently, which is why it still works for me.  I.e., this branch works:
https://github.com/sbulen/SMF2.1/tree/login_tweaks

So...  There have been changes in this area over the last few months that make it so this PR won't work now.  I'll keep digging when I can.

sbulen:
If I take the very latest from Github, and merge in the login_tweaks branch, everything works.  I cannot reproduce the problem reported in #4297 ...   (Specifically, change local or global cookie settings, log out & try to log back on...)   

This PR tests OK.   What steps were you following to make it fail?  
  
So now I've tested this PR both ways & it works both ways - with the code as it was in October & the code as it is today.

Sesquipedalian:
> This PR tests OK. What steps were you following to make it fail?

1. Log in as admin user.
2. Go to admin area, toggle "Enable local storage of cookies" setting, and save.
3. Log out.
4. Log in as admin user.

Sesquipedalian:
Before you do any more work on this, @sbulen, give me a chance to test some things.

sbulen:
Those have been my exact test steps throughout.   Make sure you're testing with the unmodified cookie format.

Sesquipedalian:
I am. ;)

sbulen:
The only thing this PR does at the moment is update the preg_match to json.  If we're keeping the preg_match at all, we need this change, since the cookie is now encoded in json.  The only wrinkle is trying to keep support for 2.0/serialized in addition to 2.1/json; it's slightly complicated because of that.  

The root cause of the problem was that due to the serial/json mismatch, the cookie & session weren't destroyed upon logout.  But the member's hash salt value gets updated upon logout.  So, upon logon, the system gets confused as it finds a cookie that contains a hashed password, that is out of date - as it is based on the old salt.  If it is allowed to clean up after itself properly on logout, the issues go away.  

The specific logic that it can't get to is the "out with the old, in with the new" logic in setLoginCookie, in subs-auth around line 50.  This is where the cookie should be destroyed and rebuilt.  (setLoginCookie is also called during logout.)  This logic exists specifically for "cookie_state" changes, and by "cookie_state", it means a change in the local or global cookie settings, i.e., the circumstances that trigger this bug.

albertlast:
A weaknes of this pr is,
that you don't define which type of json your want.
But you define with only one as right with your preg match.

sbulen:
It supports the json pattern for the current cookie definition.   You are correct in that if we redefine the cookie somehow (e.g., as an object) then this would need to change.

albertlast:
and where you define that the json is saved as json array?

sbulen:
The primary location is here, in setLoginCookie, around line 58:
`$data = $smcFunc['json_encode'](empty($id) ? array(0, '', 0) : array($id, $password, time() + $cookie_length, $cookie_state));`
No other parameters are passed to json_encode, so the default logic is used (objects are not forced).   Note the data is an array as well.

sbulen:
This has been superseded by PR #4481, so I'll close.

PR #4358, opened 2017-10-15
added new json pattern and right json_decode call fix #4357

### LINKED ISSUE #4357: New login issues
I am having issues with logon/logoff functions on the current github version.  

I get the "You were unable to login.  Please check your cookie settings."  

Sometimes I get:
Fatal error: Uncaught Error: Cannot use object of type stdClass as array in \Sources\Load.php:428 Stack trace: #0 index.php(176): loadUserSettings() #1 index.php(150): smf_main() #2 {main} thrown in \Sources\Load.php on line 428

I was re-testing to make sure the fixes for the prior login issues PR worked ( #4353 ).  To reproduce, login as admin, change cookie settings (either local or global setting), log off & try to log back on.

albertlast:
In the last pr i changed the cookie format and i was unkown that we check the pattern of the cookie,
so this fix this issue and
since the last changes it can be happen that the user run into https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Load.php#L423
which never work because of the missing/wrong attribute for the function call.
@jdarwood007 or was this with intention?

Both issues are defined here: #4357

The pr: #4356 try to fix the first issue too (don't merge both)

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -420,7 +420,7 @@ function loadUserSettings()
 	elseif (empty($id_member) && isset($_SESSION['login_' . $cookiename]) && ($_SESSION['USER_AGENT'] == $_SERVER['HTTP_USER_AGENT'] || !empty($modSettings['disableCheckUA'])))
 	{
 		// @todo Perhaps we can do some more checking on this, such as on the first octet of the IP?
-		$cookie_data = $smcFunc['json_decode']($_SESSION['login_' . $cookiename]);
+		$cookie_data = $smcFunc['json_decode']($_SESSION['login_' . $cookiename], true, false);
 
 		if (empty($cookie_data))
 			$cookie_data = safe_unserialize($_SESSION['login_' . $cookiename]);
--- a/Sources/LogInOut.php
+++ b/Sources/LogInOut.php
@@ -97,7 +97,11 @@ function Login2()
 
 	if (isset($_GET['sa']) && $_GET['sa'] == 'salt' && !$user_info['is_guest'])
 	{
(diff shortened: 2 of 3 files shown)


albertlast:
In my last pr i deleted the pattern check,
i see no need for this -> for performance is no check better.

sbulen:
@albertlast @jdarwood007 - These two last PRs from albertlast have redefined the data portion of the cookie.  Any hopes of supporting cookies across the 2.0 to 2.1 conversion are gone at this point.  Forum users must clear their cookies to logon post conversion.  

I should point out that redefining the data portion of the cookie was not necessary; paths are supported in cookies without placing them within the data portion of the cookie.  This was working fine in 2.0, and also in 2.1, and did not need to be redefined.

albertlast:
And how do you know what the path of the cookie is?

sbulen:
If you really need to know the path of the cookie, all you need to do is use one of the php calls to get it, such as this:
http://php.net/manual/en/function.session-get-cookie-params.php

We really did not need to rewrite the cookie.  That is unnecessary.

albertlast:
you should read the doc: 
> It should be noted that this gets the session cookie ini file parameters, not the parameters from the cookie itself.

So this doesn't work
Please read the documentation of this: https://stackoverflow.com/questions/2802760/is-it-possible-to-get-a-stored-cookies-path
in short only data is send to the webserver

Sesquipedalian:
It seems like the changes to the cookie format in #4353 are creating more problems than they solve. After some discussion by the dev team over on simplemachines.org, it seems like the best course is to revert #4353 and start over at solving #4297.

PR #4359, opened 2017-10-15
Templates revision

Gwenwyfar:
Round 1, start!

Only small changes for the most part. Round 2 will be removing inline styles and other html cleanup.


Sesquipedalian:
There are some conflicts to resolve before this can be merged.

XinYenFon (approved):

PR #4370, opened 2017-10-21
Removes obsolete Download() code from Display.php

Sesquipedalian:
I kept the function name and turned it into an alias of showAttachment(), just in case there are some old mods out there somewhere that call this function directly.


colinschoen (approved):
LGTM approval

Issue #4368, opened 2017-10-20
Unsafe cookie deleting


albertlast:
Since reverting pr https://github.com/SimpleMachines/SMF2.1/pull/4363
the cookie get deleting by setting time back to now() - 1 hour,
which can lead in issue because of time zone stuff.

Issue #4367, opened 2017-10-20
Json object from cookie date


albertlast:
the code line https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Load.php#L423
could create json object instead of an array,
since decline pr which it fixed you keeping this.

Issue #4366, opened 2017-10-20
Cookie will be not destroy when the board move to subfolder


albertlast:
When you got local cookie and change the board from exp. /smf to /smf/test
the cookie get lost smf is not able to delete/destroy the cookie
This issue is reintroduce by  https://github.com/SimpleMachines/SMF2.1/pull/4363

sbulen:
This is a dupe of #4297

albertlast:
Nope this issue the not mention,
The base of this issue is the missing path info in data.


________________________________
Von: Jon Stovell <notifications@github.com>
Gesendet: 21.10.2017 12:04 vorm.
An: SimpleMachines/SMF2.1
Cc: albertlast; Author
Betreff: Re: [SimpleMachines/SMF2.1] Cookie will be not destroy when the board move to subfolder (#4366)


Closed #4366<https://github.com/SimpleMachines/SMF2.1/issues/4366>.

albertlast:
i checked the pr from sbulen and he didn't fix the issue.

sbulen:
Note a correction: this issue is not a dupe of #4297.

jdarwood007:
By board you mean forum I assume?

albertlast:
Sure,
Like i already mention,
the issue for this is the missing information of the path from the cookie (which was already fixed in my pr)

sbulen:
My problem with this issue is that I think it ends up defeating the purpose of the localCookie setting.

Local cookies are intended to allow support for different forums in different subfolders.  I.e., each forum is in a unique subfolder and there is a corresponding cookie.  If you make cookie behavior independent of the subfolder, you have essentially disabled localCookies, i.e., they are now global.  

The online help indicates that keeping the subfolders distinct is important for some ISPs that are organized in that way.  You do not want the forum software or cookie operating outside of its scope.  

If you *DO* want it to use the cookie for the whole structure, independent of the subfolders, then you should de-select Local Cookies.

albertlast:
The behavior of the localcookies didn't change,
so your comment make no sense.

Issue #4365, opened 2017-10-20
While login a new session id is generated


albertlast:
Since revert pr https://github.com/SimpleMachines/SMF2.1/pull/4363
you get a new session id while login process which generate the error
that the session runs out.

sbulen:
This will be addressed by #4356

albertlast:
And where there?

sbulen:
The extra session was created because the cookie state change was not detected. Fixing the preg_match restores the cookie & session cleanup logic that was working in 2.0.

albertlast:
Nope the session keeps changing.
So instead of fixing this,
you didn't check the session id any more which is my eyes a new security issue.

sbulen:
New session IDs are generated upon login on purpose.  This is a security measure:
https://stackoverflow.com/questions/22965067/when-and-why-i-should-use-session-regenerate-id

The code to do so is in setLoginCookie.

The "extra session" bug had to do with logging OUT.   Cleaning up the cookie and session takes place on logout, not login.   

This wasn't working upon cookie state changes.  Properly detecting the cookie state change was required for this to work.  Otherwise, it tried to delete the wrong session ID, and left an extra session record out there.

albertlast:
When you look into the code you solve this issue by not checking the session_var and the sc field from the login form,
which is the security issue.
The reason why you not checking this,
that the new session_id is generated before login process is done.

PR #4373, opened 2017-10-26
Cookie session3

albertlast:
The third version of my cookie pr,
it do many things:
* Change the json format from array to object in the data field of the cookie
* It enforce this behavior
* Add the path key
* Check every smf login cookie (if valid or not)
  * Check if the smf login got the new valid format
  * Check if the path is right
  * Check if key 3 is right
  * If something is wrong it  create a new logon cookie
* When the cookie should be deleted it set the valid time at 1 instead of time()-3600
* Solve the issue of renegrated multiple php session generation in the login process (we generate once after login) https://github.com/SimpleMachines/SMF2.1/issues/4365
* Fix a issue in load.php that he json from the cookie was not converted to an php array (no open issue for this exists)
* Enforce the new json object schema on Login2 which is only called when you try to login --> should no one harm?

Since it check the path in full detail issue https://github.com/SimpleMachines/SMF2.1/issues/4366 and https://github.com/SimpleMachines/SMF2.1/issues/4364 is gone
https://github.com/SimpleMachines/SMF2.1/issues/4368 is resolved by setting cookie expire time at 1
since the pr enforce the json object we are safe the the json is not sometime a array or object solve https://github.com/SimpleMachines/SMF2.1/issues/4367
The long list of checks independent if the cookie got a valid format or not solve the issue https://github.com/SimpleMachines/SMF2.1/issues/4362

Since the issue https://github.com/SimpleMachines/SMF2.1/issues/4297 based on some of this issues,
the issue is fixed, too.

In different to other pr,
this pr keep the security check on sc and session_var untouched -> it keep working.

Possible issue could be exists in the upgrade process,
since smf 2.0 doesn't work realy with pg i can't test the upgrade process with this changes.
Since the update work "silently",
which means you keep loggedin with the old format cookie even the pr is active,
i see no high chance that the upgrad process got issue with this.

This pr is in conflict with https://github.com/SimpleMachines/SMF2.1/pull/4356

### DIFF
--- a/Sources/LogInOut.php
+++ b/Sources/LogInOut.php
@@ -809,9 +809,6 @@ function validatePasswordFlood($id_member, $member_name, $password_flood_value =
 	// Only if they're not validating for 2FA
 	if (!$tfa)
 	{
-		require_once($sourcedir . '/Subs-Auth.php');
-		setLoginCookie(-3600, 0);
-
 		if (isset($_SESSION['login_' . $cookiename]))
 			unset($_SESSION['login_' . $cookiename]);
 	}
--- a/Sources/LogInOut.php
+++ b/Sources/LogInOut.php
@@ -97,13 +97,9 @@ function Login2()
 
 	if (isset($_GET['sa']) && $_GET['sa'] == 'salt' && !$user_info['is_guest'])
 	{
-		if (isset($_COOKIE[$cookiename]) && preg_match('~^a:[34]:\{i:0;i:\d{1,7};i:1;s:(0|128):"([a-fA-F0-9]{128})?";i:2;[id]:\d{1,14};(i:3;i:\d;)?\}$~', $_COOKIE[$cookiename]) === 1)
+		if (isset($_COOKIE[$cookiename]) && preg_match('~^{"0":\d{1,7},"1":"[0-9a-f]{0,128}","2":\d{1,14}(,"3":\d{1}(,"path":"\\\\/.*")?)?}$~' , $_COOKIE[$cookiename]) === 1)
 		{
 			list (,, $timeout) = $smcFunc['json_decode']($_COOKIE[$cookiename], true);
-
-			// That didn't work... Maybe it's using serialize?
-			if (is_null($timeout))
-				list (,, $timeout) = safe_unserialize($_COOKIE[$cookiename]);
 		}
 		elseif (isset($_SESSION['login_' . $cookiename]))
 		{
--- a/Sources/Subs-Auth.php
+++ b/Sources/Subs-Auth.php
@@ -33,7 +33,7 @@ function setLoginCookie($cookie_length, $id, $password = '')
 
 	$id = (int) $id;
 
-	$time = ($cookie_length > 0 ? time() + $cookie_length : 1);
+	$time = ($cookie_length >= 0 ? time() + $cookie_length : 1);
 
 	// If changing state force them to re-address some permission caching.
 	$_SESSION['mc']['time'] = 0;
--- a/Sources/Subs-Auth.php
+++ b/Sources/Subs-Auth.php
@@ -38,8 +38,15 @@ function setLoginCookie($cookie_length, $id, $password = '')
 
 	// The cookie may already exist, and have been set with different options.
 	$cookie_state = (empty($modSettings['localCookies']) ? 0 : 1) | (empty($modSettings['globalCookies']) ? 0 : 2);
-	if (isset($_COOKIE[$cookiename]) && preg_match('~^a:[34]:\{i:0;i:\d{1,7};i:1;s:(0|128):"([a-fA-F0-9]{128})?";i:2;[id]:\d{1,14};(i:3;i:\d;)?\}$~', $_COOKIE[$cookiename]) === 1)
+	if (isset($_COOKIE[$cookiename]))
 	{
+		// Valid Format
+		$validFormat = preg_match('~^{"0":\d{1,7},"1":"[0-9a-f]{0,128}","2":\d{1,14}(,"3":\d{1}(,"path":"\\\\/.*")?)?}$~' , $_COOKIE[$cookiename]) === 1;
+
+		// Get the path and check it
+		$cookie_url = url_parts(!empty($modSettings['localCookies']), !empty($modSettings['globalCookies']));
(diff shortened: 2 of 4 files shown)


colinschoen:
@sbulen Do you have feedback on this?

sbulen:
albertlast & I respectfully disagree on this.   The initial issue was #4297 .  While in there, albertlast saw a number of things he didn't like (#4362 #4364 #4365 #4366 #4367 #4368).  This PR attempts to address all of these concerns.

The problem in my mind is that to address these, this PR redefines the SMF login cookie.  I do not think we need or want to redefine the cookie at this stage.  Redefining the cookie is not needed to address the initial problem, #4297.   

The next question is about the other 6 open issues.  The other 6 issues referenced above, in my mind, are not significant enough to warrant a redefinition of the cookie.  As an example, attempting to preserve the cookie/session across forum moves to a subfolder.  That is a rare case, a true edge condition.  I think it is quite alright to have folks log back in again after forum relocation to a subfolder, & allow normal browser & smf garbage collection deal with any stray cookies & sessions.  I do not think it is worth the risk to the broader SMF codebase of redefining the cookie.  I am concerned we will be fighting additional cookie/session issues for a while.  (I may be overly conservative regarding such changes.)

Unfortunately, albertlast & I worked on #4297 in parallel.  I submitted PR #4356 , which conflicts with this PR.  albertlast of course prefers the approach in this PR, because it covers some of his other concerns.

albertlast:
The key different between my solution and sbulen is,
that i don't disable the security checks from the login form.

colinschoen:
All right. I haven't been keeping up on any of the cookie PRs. I'll do some reading and take a look at all of these. Thanks for the work on this issue, both @albertlast and @sbulen

Sources/LogInOut.php
```diff
@@ -97,13 +97,9 @@ function Login2()
 
 	if (isset($_GET['sa']) && $_GET['sa'] == 'salt' && !$user_info['is_guest'])
 	{
-		if (isset($_COOKIE[$cookiename]) && preg_match('~^a:[34]:\{i:0;i:\d{1,7};i:1;s:(0|128):"([a-fA-F0-9]{128})?";i:2;[id]:\d{1,14};(i:3;i:\d;)?\}$~', $_COOKIE[$cookiename]) === 1)
+		if (isset($_COOKIE[$cookiename]) && preg_match('~^{"0":\d{1,7},"1":"[0-9a-f]{0,128}","2":\d{1,14}(,"3":\d{1}(,"path":"\\\\/.*")?)?}$~' , $_COOKIE[$cookiename]) === 1)
```
Sesquipedalian: The inclusion of a "path" key-value pair is to address #4366, correct? Is it also part of solving #4297, and if so, how?

Sources/LogInOut.php
```diff
@@ -809,9 +805,6 @@ function validatePasswordFlood($id_member, $member_name, $password_flood_value =
 	// Only if they're not validating for 2FA
 	if (!$tfa)
 	{
-		require_once($sourcedir . '/Subs-Auth.php');
-		setLoginCookie(-3600, 0);
```
Sesquipedalian, line 813: Why has this been removed?

Sources/LogInOut.php
Sesquipedalian, line 100: The inclusion of a "path" key-value pair is to address #4366, correct? Is it also part of solving #4297, and if so, how?

Sources/Subs-Auth.php
```diff
@@ -33,37 +33,49 @@ function setLoginCookie($cookie_length, $id, $password = '')
 
 	$id = (int) $id;
 
+	$time = ($cookie_length >= 0 ? time() + $cookie_length : 1);
```
Sesquipedalian, line 36: I like this approach.

Sources/LogInOut.php
albertlast: Solves only #4366 by knowing the full path of the cookie.

Sesquipedalian: Thanks

Sources/LogInOut.php
albertlast: It's the key issue from #4297.
Because of this line of code we generate a new php session while we didn't call the checksession.
Small explain:
Login form with form_session id 123 -> send -> validatePasswordFlood -> change php_session id to 567 -> checksession call -> check form_session id(123) and php_session id(567) -> error "sesion time out.." 
so this issue: https://github.com/SimpleMachines/SMF2.1/issues/4365
get fixed by this and is reason why we get session timeout.
 
Bad:
session exists -> start login -> new session -> validate -> success -> new session 
Right:
session exists -> start logn -> validate -> success -> new session

Sesquipedalian: I'm not sure that it is the key issue. Here's the code from the 2.0.x branch:

```
function validatePasswordFlood($id_member, $password_flood_value = false, $was_correct = false)
{
	global $smcFunc, $cookiename, $sourcedir;

	// As this is only brute protection, we allow 5 attempts every 10 seconds.

	// Destroy any session or cookie data about this member, as they validated wrong.
	setLoginCookie(-3600, 0);

	if (isset($_SESSION['login_' . $cookiename]))
		unset($_SESSION['login_' . $cookiename]);

```

This code doesn't lead to such problems in 2.0.x, and the relevant logic for how we might reach this point is unchanged in 2.1. 

Removing the code to reset the cookie from `validatePasswordFlood()` is a problem. If the user supplied the wrong login information, the cookie needs to expire right away.

albertlast: validatePasswordFlood() is called before checksession(),
so you had two possibility:
- call it later
- don't change the session

it choose last one

The issue in my test case only exists when you got a smfcookie which is not valid.
what i do to generate the issue:
Login in my pg env, 
change the setting to my mysql env,
try to login into my mysql env.
important: got the same cookie name

And sure exists the issue in 2.0.x also.

Sesquipedalian: I just tried it on 2.0.x, and everything behaved as it should.

albertlast: Okay maybe different  (I don't care 2.0.x).
For smf 2.1 it's work this way.

albertlast: So i checked smf 2.0.x this doesn't call validatePasswordFlood() in the login process.

albertlast: What is the idea by removing here the LoginCookie?

PR #4374, opened 2017-10-26
Don't forget custom_avatar_url

sbulen:
In loadTheme(), there is some logic that deals with alternate URLs.  These can include anything from detected https/http differences, to using an IP address to load the board, to using alias URLs.  

When a valid 'alternate' is detected, it updates all of the board URLs in $modSettings in memory prior to loading the theme.   It does not update custom_avatar_url.  

I believe that there will be instances, for example, when somebody invokes a page via http on an https forum or vice-versa, where this may cause avatar issues.


colinschoen (approved):
LGTM approval

Nice catch

Issue #4343, opened 2017-09-30
Missing avatars breaking layout


jdarwood007:
![image](https://user-images.githubusercontent.com/732116/31048028-faedf870-a5c9-11e7-8ce6-09698312d32a.png)

Can we make that look better when the avatar is missing?

Gwenwyfar:
I'll look into it (unless @Antes wants to do this one)

XinYenFon:
Doesn`t it require to show Default Avatar if user missing avatar ?

jdarwood007:
The files may have been missing or the url bad.  I didn't see why it wasn't loading the avatar, just noticed what the layout looked like when it was.
Its hard to click the link to the PM when its like this as well.

sbulen:
When doing an upgrade, you MUST combine/overwrite the old folders with the new ones, to get the proper default icons.  They are different than in 2.0.

illori:
combine/overwrite which old folders with what new ones?

XinYenFon:
Default avatar introduced in SMF 2.1, so no, overwrite does not required. Its a code related issue. While unread_alerts fetches the default avatar properly (?), unread_messages does not.

sbulen:
Maybe i could have worded that better.  Since the defaults images are new, they must find their way into the existing avatar & custom avatar folders during an upgrade.  I.e., by copying the new folders that have the default images on top of the existing folders (merging the contents).

In addition, the custom avatar folder did not have a fixed/predetermined name in 2.0.  Many (most? all?) users will have custom avatar folders named something other than custom_avatar.  

I found it easiest to rename my "avs" folder to custom_avatar in 2.0, prior to the upgrade.

Upgrade instructions will need to address these needs or we will have lots of "missing" avatars.  This issue would not have happened had a proper upgrade occured.

jdarwood007:
While that fixes it for included avatars, external avatars that are missing could also cause this template issue.

sbulen:
Correct.  The code should be fixed.

Just pointing out that the upgrade process & instructions are an important factor for rollout, or we will see a lot of missing avatars.

Gwenwyfar:
I have submitted a fix, but we may need to expand it to some other places where images/avatars are used that may have the same problem.

Gwenwyfar:
So we have the same alt text displaying on topics, profile and user menu, although the text doesn't break anything on those.

Gwenwyfar:
[...]

PR #4376, opened 2017-10-28
Fixes for top_info avatars

Gwenwyfar:
Fixes image scaling and #4343.

This is how it will display on images that fail to load:
broken-image

On some non-supporting browsers, nothing will show (the alt text will also remain hidden, to not break the layout).

Image used is my own.


frandominguezl (approved):

XinYenFon:
I don't think adding yet another default (fallback) avatar is the solution, core system should be loading a default avatar when one does not exist.

Gwenwyfar:
If the theme images folder is broken then the entire theme will be, not just avatars as is the case here. We could use the same image as avatarless members, but that would become hard to differentiate from broken avatars.

XinYenFon:
If the image folder is broken, user needs to fix bigger problem. Such additions only required if there is no fallback to current solution.

Gwenwyfar:
If this can be fixed another way then whoever can do that feel free to do so.

jdarwood007:
Problem would be css does not know the avatars location, since that url is separate from the themes url.

XinYenFon:
@Gwenwyfar there shouldn't be any need of any fix, the broken (non-existent etc...) avatars should fallback to default avatar, if default avatar does not exist in the situation that means either code lacks the fallback to default avatar or setup has issues related to locating the default avatar. What you are offering is actually a fallback for default avatar.

Gwenwyfar:
Exactly. If the problem can be fixed with the default avatar fallback, then we don't need this fix.

I made this assuming the paths to avatars don't work (for whatever reason) and that this may be a common situation, but if the default avatar is already serving that purpose there is no need for this, if it can be made to work correctly in all cases (although keeping the transparent alt doesn't hurt in any case)

colinschoen:
I agree with @Antes. If we are displaying this second-fallback image something is seriously wrong and it may create more confusion as to why or where this is actually loading from if all the paths are wrong.

colinschoen:
Closing per rationale above.

Gwenwyfar:
Ok. I'll submit the other fixes included in this in another PR.

PR #4377, opened 2017-10-28
Allow self-signed certs

sbulen:
Realized while testing that the ssl check does not honor self-signed certs.  It's possible some sites want to use self-signed.  Also, not honoring self-signed will hinder folks' ability to test.  

Tested in windows & linux, with cert & without.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -6016,18 +6016,11 @@ function build_regex($strings, $delim = null, $returnArray = false)
  */
  function ssl_cert_found($url) {
 
-	global $db_show_debug;
- 
 	// Ask for the headers for the passed url, but via https...
 	$url = str_ireplace('http://', 'https://', $url) . '/';
 
 	$result = false;
-	$params = array('ssl' => array('capture_peer_cert' => true));
-
-	// Allow support for self-signed certs, but only if $db_show_debug is true
-	if (!empty($db_show_debug))
-		$params['ssl'] += array('verify_peer' => true, 'allow_self_signed' => true);
-
+	$params = array('ssl' => array('capture_peer_cert' => true, 'verify_peer' => true, 'allow_self_signed' => true));
 	$stream = stream_context_create ($params);
 
 	$read = @fopen($url, 'rb', false, $stream);
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -6020,7 +6020,7 @@ function build_regex($strings, $delim = null, $returnArray = false)
 	$url = str_ireplace('http://', 'https://', $url) . '/';
 
 	$result = false;
-	$stream = stream_context_create (array("ssl" => array("capture_peer_cert" => true)));
+	$stream = stream_context_create (array("ssl" => array("capture_peer_cert" => true, "verify_peer" => true, "allow_self_signed" => true)));
 	$read = @fopen($url, "rb", false, $stream);
 	if ($read !== false) {
 		$cont = stream_context_get_params($read);
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -6016,15 +6016,24 @@ function build_regex($strings, $delim = null, $returnArray = false)
  */
  function ssl_cert_found($url) {
 
+	global $db_show_debug;
+ 
 	// Ask for the headers for the passed url, but via https...
 	$url = str_ireplace('http://', 'https://', $url) . '/';
 
 	$result = false;
-	$stream = stream_context_create (array("ssl" => array("capture_peer_cert" => true, "verify_peer" => true, "allow_self_signed" => true)));
-	$read = @fopen($url, "rb", false, $stream);
+	$params = array('ssl' => array('capture_peer_cert' => true));
+
+	// Allow support for self-signed certs, but only if $db_show_debug is true
+	if (!empty($db_show_debug))
+		$params['ssl'] += array('verify_peer' => true, 'allow_self_signed' => true);
+
+	$stream = stream_context_create ($params);
+
+	$read = @fopen($url, 'rb', false, $stream);
 	if ($read !== false) {
 		$cont = stream_context_get_params($read);
-		$result = isset($cont["options"]["ssl"]["peer_certificate"]) ? true : false;
+		$result = isset($cont['options']['ssl']['peer_certificate']) ? true : false;
 	}
     return $result;
 }


albertlast:
I would not do that hardcoded because this are security issues and
since "Let's Encrypt" it's for the most cases (beside test cases) easy to get real certs.

sbulen:
To be clear, the choice whether to enable https is under admin control and is not hard-coded.

This is a helper function, used by the installer and the admin control panel,  that determines whether the https options are disabled or not.   This change enables an admin to use a self signed cert.

albertlast:
But the admin had not the controll if self signed is allowed or not.

sbulen:
Corect.   And now s/he has that choice.

albertlast:
Wher he can choose?
In my eyes you hard coded that self signed are allowed(he had not the choice if he want it or not).

sbulen:
The admin can choose whether to use https on the installer and in the acp.

The admin will know what type of cert s/he is using.

Adding settings for settings for settings is overkill and confusing.  This change leaves the admin with greater flexibility without unnecessary complexity.

tinoest:
I don’t think this is a good change, self signed certs still throw the error when you visit the site. You can get dev certs from letsencrypt ( or other certificate authorities ) if you need them. 

I can see this causing problems for those that don’t understand the difference between them.

albertlast:
In my eyes this "feature" needs to be able to disable (and this should be the default setting of this).

sbulen:
I really think it should be up to the admin, and that we should not prohibit the admin from doing so.   

In practice, I think the biggest need is for test environments on local LAMP & WAMP servers.  Let's Encrypt doesn't do this, actually:
https://community.letsencrypt.org/t/can-i-use-letsencrypt-in-localhost/21741

There are pretty complex workarounds, here is one:
https://gist.github.com/andrewn/b30375074398ab94361dc607ac7c0e12

But the bottom line is that for testing, I think that most folks will use self-signed & ignore the warning.  Do a quick search and this is still the predominant method recommended out there for testing locally.

tinoest:
Then as @albertlast said it should be a check box to enable this functionality. Not the decision being taken away from them.

sbulen:
The admin already has options to enable https - it is already under admin control.

albertlast:
we don't talk about >https< we talk about self signed certs.

sbulen:
I don't believe adding new options adds any additional protection.  It just makes the UI more complex.  

I favor simplicity & choice over complexity & constraints.   It's a different design approach.  Legitimately differing opinions.

albertlast:
than keep the feature out,
because you and we already mention this is only intresting for dev and
the most user/admin smf are not dev.

tinoest:
But you’re making it more complex as you’ll get questions on I have a cert but I’m still getting a error in x browser. 

They should have a option to enable that functionality for dev or testing purposes.

sbulen:
But that will leave an unnecessary constraint.  Again, I believe in giving the admins the freedom to choose these things.  We really shouldn't make it more complex than necessary.

sbulen:
What might be a good longer-term thing, maybe a 3.0 topic, may be a 'test' vs 'prod' setting or settings.  But SMF makes no such distinction today in the 2.0 branch that I know of.

tinoest:
I don’t see how that isn’t giving them the freedom. If anything it is making them more aware of their choice. 

I’m just wary of it causing support questions which could be avoided.

sbulen:
Understood, & I agree it's a valid point.  

I'm more concerned, however, on support calls from folks not understanding the options, selecting them unwisely, not aware of the cross dependencies, or not being able to test.   

When there are differing opinions, I defer to the dev team as the final arbiters of consistency & usability; that's their role.

jdarwood007:
For now, I would suggest tying this into !empty($db_show_debug).  If that is set, we are most likely in a dev environment as it is and a self signed would be fine.  This doesn't make it a feature, but makes it function in a developers local environment for testing purposes.  In future versions we can take on a better solution.

sbulen:
That makes sense.

sbulen:
Change made.  It's a little awkward, but it does work.   
I'm about to update the PR description with some usage notes.

albertlast (approved):

colinschoen:
Grr I really don't like polluting $db_show_debug as this has nothing to do with what that setting's name implies. Can we think of a better setting to use to enable this? This seems like a nontrivial side effect.

sbulen:
I am not aware of any other test/prod setting.

As noted above, i dont think we should restrict usage of self-signed certs.  If an admin wants to uae one, s/he should be allowed to do so.  I'd rather leave the PR as initially submitted.

When I initially added this cert check, I was solving the problem of sites basically disappearing when set to https without a cert.  I can imagine plenty of prod uses for self-signed: small private clubs, teams, faculties, etc, where self-signed meets their needs.

colinschoen:
I’m leaning towards allowing self-signed in any context.

tinoest:
I can’t think of any reason to use a self signed cert when it is so easy to generate a valid one. If you’re not comfortable with letsencrypt and the command line then use;

https://www.sslforfree.com/

It’s not good practice to use self signed certificates anymore, they are susceptible to man in the middle attacks as anyone can issue a similar cert and google will not increase your page rankings either for having a https site. 

I get it on testing ( kinda ) but really don’t think it’s a good idea.

colinschoen:
The thing is that anyone who knows how to use a self-signed certificate probably knows about the pros and cons of using one. I don't think we should restrict forum administrators from setting up a self-signed cert should they have a specific case where it is useful for them.

colinschoen:
@sbulen Can we go back to having this enabled by default and not hiding it behind `$db_show_debug`

albertlast:
@colinschoen the basic of https is not only to encrypt the connection,
is also important to know that you connected to the right "server".
when you allow self signed cert than not only a admin can use a server with self signed,
also "bad person" with a man in the middle attack with a self signed cert.

https://security.stackexchange.com/questions/8110/what-are-the-risks-of-self-signing-a-certificate-for-ssl

Ther is no good reason to allow them in general,
when you don't want to check for dev,
than keep the feature out but not lower the default setting of smf.

sbulen:
Updated, per request.

colinschoen (approved):
LGTM approval

After discussion and consideration it is decided to allow self-signed certificates based on the fact that browsers actively warn users if a certificate is not signed by a trusted certificate authority and the subset of forum owners who know how to set up a self signed certificate also are familiar with the risks associated with it. Furthermore, many other similar projects permit self-signed certificates to be used with their software (PHPBB, Owncloud, Wordpress...).  

Thanks @sbulen for this PR.

PR #4378, opened 2017-10-31
Custom profile fields

MissAllSunday:
Adds two new places for profile fields, before and after the username. This is literally before and after the username, inside the h4 tag. Good for crown mods or other type of "custom usernames" modifications.

It also removes the hardcoded switch I forgot to remove.

image


Sources/PersonalMessage.php
```diff
@@ -1083,29 +1083,7 @@ function prepareMessageContext($type = 'subject', $reset = false)
...
-				case 5:
-					$output['custom_fields']['above_member'][] = $custom;
-					break;
-				case 6:
-					$output['custom_fields']['bottom_poster'][] = $custom;
-					break;
-				default:
-					$output['custom_fields']['standard'][] = $custom;
-			}
+			$output['custom_fields'][$context['cust_profile_fields_placement'][$custom['placement']]][] = $custom;
```
albertlast, line 1086: Would be nice to place a comment here with the position how exists otherwise this place is such generic,
that this information is hard to see.

MissAllSunday: Thats kinda the point, to be generic. I did it like this to have a single point where custom fields are defined, makes it easier for anyone to add more.

albertlast: Sure and it's fine but the existing point are not visible anymore or
this somewhere already define?

MissAllSunday: On Sources/Load.php https://github.com/SimpleMachines/SMF2.1/blob/c939cab292f53d2de9058fde91c8f5493692aa5f/Sources/Load.php#L324

albertlast (approved):

frandominguezl:
Looks good 👍

colinschoen (approved):
LGTM

While testing this I noticed another issue where we aren't properly sanitizing HTML for custom profile fields even when NO HTML is selected as the validation. Tracking at #4379

PR #4380, opened 2017-11-03
Add a "nohtml" mask validation

### LINKED ISSUE #4379: XSS Profile Fields
HTML tags are not being sanitized from profile field values creating a stored cross site scripting vulnerability.

screen shot 2017-11-02 at 00 28 08
screen shot 2017-11-02 at 00 27 56
screen shot 2017-11-02 at 00 27 40

MissAllSunday:
fixes #4379

### DIFF
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -1150,8 +1150,8 @@ function makeCustomFieldChanges($memID, $area, $sanitize = true, $returnErrors =
 
 	$errors = array();
 
-	// if ($sanitize && isset($_POST['customfield']))
-	// 	$_POST['customfield'] = htmlspecialchars__recursive($_POST['customfield']);
+	if ($sanitize && isset($_POST['customfield']))
+		$_POST['customfield'] = htmlspecialchars__recursive($_POST['customfield']);
 
 	$where = $area == 'register' ? 'show_reg != 0' : 'show_profile = {string:area}';
 
@@ -1200,8 +1200,7 @@ function makeCustomFieldChanges($memID, $area, $sanitize = true, $returnErrors =
 			// Any masks?
 			if ($row['field_type'] == 'text' && !empty($row['mask']) && $row['mask'] != 'none')
 			{
-				// $value = $smcFunc['htmltrim']($value);
-
+				$value = $smcFunc['htmltrim']($value);
 
 				// Try and avoid some checks. '0' could be a valid non-empty value.
 				if (empty($value) && !is_numeric($value))
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -1150,8 +1150,8 @@ function makeCustomFieldChanges($memID, $area, $sanitize = true, $returnErrors =
 
 	$errors = array();
 
-	if ($sanitize && isset($_POST['customfield']))
-		$_POST['customfield'] = htmlspecialchars__recursive($_POST['customfield']);
+	// if ($sanitize && isset($_POST['customfield']))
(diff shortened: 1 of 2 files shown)


Sources/Profile-Modify.php
```diff
@@ -1199,7 +1199,15 @@ function makeCustomFieldChanges($memID, $area, $sanitize = true, $returnErrors =
 			// Any masks?
 			if ($row['field_type'] == 'text' && !empty($row['mask']) && $row['mask'] != 'none')
 			{
-				if ($row['mask'] == 'email' && (!filter_var($value, FILTER_VALIDATE_EMAIL) || strlen($value) > 255))
+				if($row['mask'] == 'nohtml' && ($value != strip_tags($value) || (!filter_var($value, FILTER_SANITIZE_STRING))))
```
colinschoen: nit: `if(` -> `if (`

colinschoen (changes requested):
Nit

colinschoen:
This is also failing for empty fields.

Sources/Profile-Modify.php
colinschoen: Should be comparing return value of filter_var to $value as the return value of filter_var is the sanitized string.

MissAllSunday:
An improved version it is still pretty weak againts xss attacks, un_htmlspecialchars is used when displaying the fields to offer some flexibility so htmlspecialchars() is out.  I still suggest to make "nohtml" mask as a default if none is selected and build a real function capable of at least protect against the most [common cases ](www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet)

Either that or completely disallow HTML on custom fields. It already supports BBC (although I couldn't find the option on the admin panel). 

Since I'm comparing the raw value with its un_htmlspecialchars() version, html entities will trigger the check.

Sources/Profile-Modify.php
```diff
@@ -1193,13 +1193,29 @@ function makeCustomFieldChanges($memID, $area, $sanitize = true, $returnErrors =
...
+
+				if ($row['mask'] == 'nohtml' && ($valueReference != strip_tags($valueReference) || $value != filter_var($value, FILTER_SANITIZE_STRING) || preg_match('/<(.+?)[\s]*\/?[\s]*>/si', $valueReference)))
+				{
+					if ($returnErrors)
+						$errors[] = 'custom_field_nohtml_fail';
+
+					else
+						$value = '';
+				}
+				elseif ($row['mask'] == 'email' && (!filter_var($value, FILTER_VALIDATE_EMAIL) || strlen($value) > 255))
```
colinschoen, line 1218: `!filter_var($value, FILTER_VALIDATE_EMAIL)` -> `$value != filter_var($value, FILTER_VALIDATE_EMAIL)`

colinschoen (changes requested):

Sources/Profile-Modify.php
MissAllSunday: Well, I didn't modified that part but its actually correct.  Theres a difference between VALIDATE and SANITIZE options.
For the nohtml mask there is no VALIDATE_HTML_SOMETHING filter which is why we indeed have to use a comparison check !=

For the mail mask however, FILTER_VALIDATE_EMAIL will return false if the filter fails which is what that piece of code is checking against. 

So, doing !filter_var($value, FILTER_VALIDATE_EMAIL) is the same as $value == filter_var($value, FILTER_VALIDATE_EMAIL)

colinschoen: Thank you... missed this.

colinschoen (approved):
LGTM

Thanks!

PR #4383, opened 2017-11-05
Use numeric keys for multiple options fields

### LINKED ISSUE #4381: Gender Icon profile fields Multilanguage Bug
The Gender icon  Status ist not display in other language. As an example in German.

The bug is in INPUT. 
`generic_icons gender_{INPUT}`
Input is the replacement for  None, Male or Female. But when it is translated in the admin center => Custom profile fields,  it results to  keine Angabe, Männlich oder Weiblich. 

gender_german-1

gender_german-2

gender_german-3
gender_german-4

gender_german-5

Possible solution would be to change the value in index. css. But I find it to be a less than optimal solution.

=> http://www.stephan-frank.de/board/index.php/topic,12.msg24.html#msg24

MissAllSunday:
In order to properly support translations for multiple option fields, a key and a value needs to be introduced. The value can be translated into whatever you want and the key will be used to reference said value.

Gender fields are now referenced by value:

0 => None
1 => Male
2 => Female

By taking advantage of how arrays are constructed, there is no need to store the key somewhere else or having it stored along with the value.

This means any new option added to the css will have to be added by its number:

.my_custom_gender_3
.other_custom_gender_4

Buts its a minor issue compared to adding support for translatable fields. Also added a {KEY} token to make it easier for the user to add the key to their enclosing options.

Changed the word "disabled" for "none". The install sql fields where using None instead of Disabled, I have no idea why I chose "Disabled" but "None" looks nicer. 

fixes #4381


albertlast:
When i read to code correctly than is this not a generic extension to customfields that they are multilanguage able right?
So only gender is with this solution able to be existing with different language?

MissAllSunday:
All fields with multiple options will now be referenced by value instead of input. Genders is the only default field that has multiple options thats why we can apply the needed changes.

Other multiple fields will also have to use the {KEY} option if they want to reference the field somewhere else like an image or something like that, if the field isn't referenced then they can still use input normally.

albertlast:
Okay.
I'm suprised that mostly no changes in the upgrade script are.
Did we already save the key of the multi field in the database?

MissAllSunday:
The options are stored as a comma separated string. Using explode on it will automatically give us the keys since PHP will treat it as a numeric array

wintstar:
For gender None 
replace 
index.css
```
.generic_icons.gender_None {
	background: none;
}
```
to
```
.generic_icons.gender_0 {
	background: none;
}
```

Sources/Load.php
```diff
@@ -1540,12 +1540,24 @@ function loadMemberContext($user, $display_custom_fields = false)
 			$value = $profile['options'][$custom['col_name']];
 
 			// Don't show the "disabled" option for the "gender" field.
-			if ($custom['col_name'] == 'cust_gender' && $value == 'Disabled')
+			if ($custom['col_name'] == 'cust_gender' && $value == 'None')
```
colinschoen: This has the side effect of assigning Male the key `0` and Female the key `1`.

Themes/default/css/index.css
```diff
@@ -1701,7 +1701,7 @@ tr.windowbg td, tr.windowbg2 td, .table_grid tr td {
 .generic_icons.general, .generic_icons.boards, .generic_icons.topics_views {
 	background-position: -31px -31px;
 }
-.generic_icons.gender_Male {
+.generic_icons.gender_1 {
```
colinschoen, line 1721: If no changes are made to the logic where None is skipped because it is a special value then this should be changed from `.generic_icons.gender_1` to `.generic_icons.gender_0`.

Themes/default/css/index.css
```diff
@@ -1688,7 +1688,7 @@ tr.windowbg td, tr.windowbg2 td, .table_grid tr td {
 .generic_icons.invalid {
 	background-position: -109px -5px;
 }
-.generic_icons.gender_Female {
+.generic_icons.gender_2 {
```
colinschoen, line 1708: If no changes are made to the logic where None is skipped because it is a special value then this should be changed from `.generic_icons.gender_2` to `.generic_icons.gender_1`.

colinschoen (changes requested):
Apologies for delay in reviewing. 

Nice work. Small changes...

colinschoen (approved):
LGTM approval

colinschoen:
Thanks.

PR #4384, opened 2017-11-05
Wrong var used in Sources/Subs-BoardIndex.php

### LINKED ISSUE #4382: Avatars on board index hidden setting doesn't work
Putting aside the fact that the hidden setting doesn't do anything visible when enabled (which is by design rather than an oversight), the logic doesn't work on the board index correctly. In Subs-BoardIndex.php, it this:

```
		if (!empty($settings['avatars_on_boardIndex']))
			$this_last_post['member']['avatar'] = set_avatar_data(array(
				'avatar' => $row_board['avatar'],
				'email' => $row_board['email_address'],
				'filename' => !empty($row['member_filename']) ? $row_board['member_filename'] : '',
			));
```

The last element of that array should use $row_board not $row.

MissAllSunday:
fixes #4382


colinschoen (approved):
LGTM

Thank you

PR #4388, opened 2017-11-18
Fulltext search edit - fixes #4387

### LINKED ISSUE #4387: Search - fulltext not allowed if db not myisam
On a newly installed 2.1 forum, pulled up the Search Method form, and received the following prompt:
image

Since it was a new install, the tables were innodb, and this failed the old 2.0 edit that required myisam.  

I do think InnoDB supports fulltext these days, mysql 5.6+.  I do not know if it is syntactically identical with myisam.  

So...  I think we need some experimentation to determine whether to remove the myisam restriction, or modify it somehow, to make it smarter based on the version of mysql.

sbulen:
There was a problem with the version check that prevented this from working properly.  

Fixes #4387 

Tested by creating fulltext indexes and performing searches on MyISAM and InnoDB.  Also confirmed edit works properly by preventing creation of a fulltext index on MySQL 5.5 for InnoDB.


albertlast:
Since many distro provide mariadb instead of mysql,
you should maybe extend the check also for mariadb (i know that smf didn't support mariadb),
would safe many support tickets...

sbulen:
Since MariaDB supports the same engines (InnoDB and MyISAM), I don't think anything needs to change here to provide MariaDB support.

In theory, it is "drop-in" compatible with MySQL.  Like Percona.

MariaDB offers several additional engines (Aria), but a lot more work would need to be done to support those.  I don't think we want to go there.

albertlast:
The possible issue is that db_get_version returns 10.1.21-MariaDB instead of 5.7.0,
but your are lucky because mariadb 10 ~ mysql 5.6.
When the feature exists since 5.7 then you had to beware of which db you running,
because of mysql it would be 5.7 and in maria it would be 10.1 (10 wouldn't support this).

sbulen:
Yikes...   Their numbering strayed.  

So much for "drop in" compatibility.  That is now out the door.

Looks like innodb supports fulltext from 10.0.5 on:
https://mariadb.com/kb/en/library/full-text-index-overview/

I think the MariaDB folks shot themselves in the foot.  

I'm not so sure we can maintain a mapping of versions - it would have to be feature by feature, since they have abandoned MySQL version numbering.  That is not a reasonable expectation.  Over time, our code would have to be peppered with individual checks.  

I will leave this as-is.

colinschoen (approved):
LGTM approval

PR #4391, opened 2017-11-25
Handle version_comment better

albertlast:
Should evade the bug from https://www.simplemachines.org/community/index.php?topic=557286.0

### DIFF
--- a/Sources/DbExtra-mysql.php
+++ b/Sources/DbExtra-mysql.php
@@ -394,7 +394,14 @@ function smf_db_get_engine()
 	if (!empty($db_type))
 		return $db_type;
 
-	$request = $smcFunc['db_query']('', 'SELECT @@version_comment');
+	$request = $smcFunc['db_query']('', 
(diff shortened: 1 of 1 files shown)


sbulen (changes requested):
Unfortunately, the MySQL folks are rethinking how this works...   This will be deprecated in 5.7.6, in favor of the performance schema:
https://dev.mysql.com/doc/refman/5.7/en/performance-schema-variable-table-migration.html

We should think about some fallback logic.

albertlast:
Or we raise the min version of mysql to 5.0.22 / 5.1.10,
than is the bug fixed and we don't need change anything.

btw 5.0.22 was released in the year 2006...

sbulen:
I believe this change already went in with PR #4392.   This is a dupe.

albertlast:
Yes and no.
Yes the change already is inlcuded,
but since no one is awnser >my< question about the future of mysql version in the dev board,
i can't finish this pr.

PR #4392, opened 2017-11-25
Engine to vendor

albertlast:
This PR is a spin off pr: #4391
Since i don't know for sure if this change get merge.

The word "engine" is in mysql context clear defined as storage engine and storage engine are
myisam, innodb, xtradb, blackhole etc.
but this function didn't return the engines it return the vendor of the db -> right name would be vendor.

### DIFF
--- a/Sources/DbExtra-mysql.php
+++ b/Sources/DbExtra-mysql.php
@@ -394,7 +394,14 @@ function smf_db_get_engine()
 	if (!empty($db_type))
 		return $db_type;
 
-	$request = $smcFunc['db_query']('', 'SELECT @@version_comment');
+	$request = $smcFunc['db_query']('', 
+				'SELECT VARIABLE_VALUE
+				 FROM INFORMATION_SCHEMA.GLOBAL_VARIABLES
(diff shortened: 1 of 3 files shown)


jdarwood007:
I'm fine with it as long as we don't modify what gets sent up to the SM Stat collection without consulting myself.  As I would be the one to fix it on the server side to handle that change.

albertlast:
@jdarwood007 i never look in the code but when i look in this line https://github.com/albertlast/SMF2.1/blob/654db6417a9b0b61addf029f90bfa05e7867501e/Sources/DbExtra-mysql.php#L417
then the function return fail when the db is a mysql mysql is this wanted?

jdarwood007:
No that works.  If the comment is not empty, it attempts to match.  If no match happens the else doesn't happen and it returns with a mysql.  If the comment is empty, we return fail.

This could lead to some other fork possibly getting detected as mysql though.  I'm not aware of any other major forks at this time though.

colinschoen (approved):
LGTM approval

PR #4393, opened 2017-11-29
Removed Session id Check

albertlast:
In my eyes is this useless test and
could maybe root of the issue for https://www.simplemachines.org/community/index.php?topic=554292.0
be.

### DIFF
--- a/Sources/Session.php
+++ b/Sources/Session.php
@@ -190,9 +190,6 @@ function sessionDestroy($session_id)
 {
 	global $smcFunc;
 
-	if (preg_match('~^[A-Za-z0-9,-]{16,64}$~', $session_id) == 0)
-		return false;
-
 	// Just delete the row...
 	$smcFunc['db_query']('', '
 		DELETE FROM {db_prefix}sessions


colinschoen:
Can you advise why this is a useless check? Why would a session ID ever be created that doesn’t fall into these constraints.

albertlast:
Two reason:
- when the check failed than we return false,
 which we don't catch somewhere so we get in a state where this create errors
- it doesn't solve any issues

alternativ solution would be to return true if the check fails and don't do the sql stuff,
but i didn't see the benefit of doing so.

Arantor:
Why wouldn't you remove it from all the session functions?

The reason it's there is because the session ID is user supplied, and PHP for a long time didn't sanitise correctly either. I'm not totally sure when this was fixed in PHP itself, but it was certainly a bug in 4.3 and possibly 4.4. Old bug, presumed fixed, but best to double check.

albertlast:
In some othercases it surve maybe an puporse,
but here it create an error (because it return false) and
evade a database delete action (where we don't care if it success or not).

sbulen:
To me it appears that albertlast's concern is valid; I don't think that check is adding value, it just increases risk of leaving stray sessions around.  

sessionDestroy is a callback, called from PHP, not directly from SMF code.  The session ID is supplied by PHP.  As far as I can see, it is not user supplied, or even SMF supplied.  (Take a peek at session_set_save_handler() in Sessions.php, & also look for all instances of sessionDestroy.) 

On the other hand, if this change is deemed risky, the worst case scenario of NOT implementing this change is that the stray session is there until garbage collection.

Arantor:
You can substitute the function name of sessionDesroy for literally all of the other session functions. They're all callbacks, they all have the same check.

In all cases, even the dodgy ones, it's still formulated out of data provided by the user.

sbulen:
@Arantor - Is your recommendation to keep the session check?  

Since the sessionID is generated by PHP, I'm confused by your note that it is "still formulated out of data provided by the user".   I'm trying to understand the risk is here, & if it is current.  Could you explain the risk?  

I do agree that if we want to get rid of it, we should get rid of the similar checks in all SMF session callbacks.

Arantor:
PHP gets the data from user cookies and is trusting that the data is in the right format. It provably wasn't sanitised correctly in earlier versions of PHP.

sbulen:
We're not talking about the contents here, just the session ID.

Arantor:
Yes, and that still comes out of the cookie. How else could it possibly get to PHP?

sbulen:
@Arantor ...  True.  If I understand it correctly, PHP is getting the sessionID it from the PHPSESSIONID cookie (unless it was renamed...).

So....     I believe you are suggesting that this check should be kept.  The reason is that the sessionID that PHP is passing here does in fact come from a cookie.  As such, the preg_match is a minimal check to ensure that outlandish values are not passed.  It's far from perfect, but it's better than nothing.

Arantor:
Not necessarily from the PHPSESSID cookie, but potentially the SMF cookie, yup. It's only a test to ensure the cookie is possibly valid and it should only be a problem if it doesn't cover the possible range of values. If the regex isn't broad enough, then that needs to be fixed.

albertlast:
Then we should return true in the case when the check failed?

Arantor:
How about doing some debugging instead of just blowing it away?

As in: when this happens, log an error for the time being, return false as is done today and then look at what the error log contains. It's possible the regex needs updating. It's possible there's malicious behaviour. Without actually examining what is going on, it's impossible to know what the correct behaviour really is.

PR #4394, opened 2017-12-04
Update the favicon

Gwenwyfar:
It's using the same proportions as the new svg logo, and updated to contain all .ico favicon sizes. Total size could've been smaller but this image in particular kept resulting in a lot of aliasing in other converters.

Since it uses transparency it will work better for places outside the browser tab, and in different backgrounds. Color and style were a mix of old favicon vs new curve theme.

Samples:
favicon-sample
dark


frandominguezl (approved):
Looking good

illori:
i would post this on the team boards to get further input, i am not sure if this violates our trademarks by doing this change to the image. as far as i know it has always been black inside the fulcrum and not the other way around, so this is a bit of a major change.

XinYenFon:
A very fine upgrade

colinschoen (approved):
Very nice work. 

LGTM approval (pending no legal or trademark issues with this change. I’ll leave that discussion to the team boards per Illori’s suggestion)

sbulen:
To me, that looks like the logo on github for SMF.  

Check here:
https://github.com/SimpleMachines

Gwenwyfar:
@sbulen it is, that one was updated a few months back, and the previous (almost identical) one was one of the main bases for it. So the current Github logo, the new 2.1 header logo and this favicon are identical. Hoping we can update the others soon as well, once trademark issues are sorted :)

colinschoen:
Per discussion in team boards this seems fine. Merging.

PR #4399, opened 2017-12-07
PharData sometimes wants extensions; Fixes #4395

### LINKED ISSUE #4395: Installing theme from package doesn't work
When installing a new theme through the package, you are given an error: "_An error occurred while installing the theme_". No errors in the error log; multiple PHP versions were tested (5.4, 5.5, 5.6, 7.0 and 7.1).

Installing from a folder works fine. There are a couple packages in the theme site that can be tested, all of them seem to give this error.

Tests done on a live server, not localhost.

Edit: Could be related to #2995

sbulen:
This fixes the theme upload issue reported in #4395 .
Tested for both wamp & hosted 'nix.  

While in there, fixed error reporting for this function.

### DIFF
--- a/Sources/Subs-Package.php
+++ b/Sources/Subs-Package.php
@@ -246,11 +246,10 @@ function read_zip_file($file, $destination, $single_file = false, $overwrite = f
 		// This may not always be defined...
 		$return = array();
 
-		// Note that hosted unix platforms sometimes need an extension; win usually has .tmp & thats ok
-		if (stristr($file, '.zip') === false && stristr($file, '.tmp') === false) {
-			@rename($file, $file . '.zip');
-			$file = $file . '.zip';
-		}
+		// Some hosted unix platforms require an extension; win may have .tmp & that works ok
+		if (strtolower(substr($file, -4)) != '.zip' && strtolower(substr($file, -4)) != '.tmp')
+			if (@rename($file, $file . '.zip'))
+				$file = $file . '.zip';
 
 		$archive = new PharData($file, RecursiveIteratorIterator::SELF_FIRST, null, Phar::ZIP);
 		$iterator = new RecursiveIteratorIterator($archive, RecursiveIteratorIterator::SELF_FIRST);
--- a/Sources/Subs-Package.php
+++ b/Sources/Subs-Package.php
@@ -246,6 +246,12 @@ function read_zip_file($file, $destination, $single_file = false, $overwrite = f
 		// This may not always be defined...
 		$return = array();
 
+		// Note that hosted unix platforms sometimes need an extension; win usually has .tmp & thats ok
+		if (stristr($file, '.zip') === false && stristr($file, '.tmp') === false) {
+			@rename($file, $file . '.zip');
+			$file = $file . '.zip';
+		}
+
 		$archive = new PharData($file, RecursiveIteratorIterator::SELF_FIRST, null, Phar::ZIP);
 		$iterator = new RecursiveIteratorIterator($archive, RecursiveIteratorIterator::SELF_FIRST);
 
@@ -314,6 +320,7 @@ function read_zip_file($file, $destination, $single_file = false, $overwrite = f
 	}
 	catch (Exception $e)
 	{
+		log_error($e->getMessage(), 'general', $e->getFile(), $e->getLine());
 		return false;
 	}
 }
--- a/Sources/Subs-Package.php
+++ b/Sources/Subs-Package.php
@@ -247,7 +247,7 @@ function read_zip_file($file, $destination, $single_file = false, $overwrite = f
 		$return = array();
 
 		// Some hosted unix platforms require an extension; win may have .tmp & that works ok
-		if (strtolower(substr($file, -4)) != '.zip' && strtolower(substr($file, -4)) != '.tmp')
+		if (!in_array(strtolower(pathinfo($file, PATHINFO_EXTENSION)), array('zip', 'tmp')))
 			if (@rename($file, $file . '.zip'))
 				$file = $file . '.zip';
 


jdarwood007:
Would a file like this break this?: test.zip.tmp.tar.gz

sbulen:
@jdarwood007 - That would be treated as a tar.gz and would not be processed by this .zip logic.

To the broader question, I believe layered extensions fail anyway with the same UnexpectedValueException.  Specifically, in my testing, the xxxx.tmp.zip pattern failed on the win platform, which is why I don't rename it if it is already .tmp.  

If you think it is safer, I could restructure this to ensure the extensions are at the end.

sbulen:
I went ahead & only checked the last extension.  It does seem a little safer.  Tests OK on both platforms.

jdarwood007:
Ok, I was just checking as extension verification is a tricky thing and not safe to blindly trust.  Mime reading would be read better, but slower.

sbulen:
@jdarwood007 - This is almost backwards...   Phar wants an extension on unix or it blows up with this message:  
`UnexpectedValueException: Cannot create phar '/tmp/phpX3O29S', file extension (or combination) not recognised or the directory does not exist in /blah/blah/public_html/VAN21/Sources/Subs-Package.php:249`

So we are ADDING the .zip extension to a file that didn't have an extension in order to make Phar happy.  

BTW the way SMF gets to this zip logic is via the file signature 'magic number'.

Sources/Subs-Package.php
```diff
@@ -246,6 +246,11 @@ function read_zip_file($file, $destination, $single_file = false, $overwrite = f
 		// This may not always be defined...
 		$return = array();
 
+		// Some hosted unix platforms require an extension; win may have .tmp & that works ok
+		if (strtolower(substr($file, -4)) != '.zip' && strtolower(substr($file, -4)) != '.tmp')
+			if (@rename($file, $file . '.zip'))
+				$file = $file . '.zip';
+
```
Sesquipedalian: This would be more reliable:
```
if (!in_array(strtolower(pathinfo($file, PATHINFO_EXTENSION)), array('zip', 'tmp')) && @rename($file, $file . '.zip'))
	$file = $file . '.zip';
```

Sesquipedalian (changes requested):

Sources/Subs-Package.php
Sesquipedalian: Actually, I would think that it's probably better to skip the .tmp option entirely and just force a rename to a zip extension in all cases.
```
if (strtolower(pathinfo($file, PATHINFO_EXTENSION) !== 'zip' && @rename($file, $file . '.zip'))
	$file = $file . '.zip';
```

sbulen:
As noted above, .tmp.zip failed on Windows.  That's why I did it that way.

Some fixes work on Win, some work on Unix.  I'm testing both, so we don't loop around & do this again again...

I'll look into the other simplification & retest.

Sesquipedalian:
Does `file.tmp` fail on *nix?

Sesquipedalian:
Also, when you say it failed on Windows, do you mean it failed in IIS on Windows or in Apache on Windows?

sbulen:
It failed on apache on windows.   I don't recall testing .tmp on unix - I probably did but that was months ago...   

The reason I tend to 'layer' ifs is because PHP says you shouldn't make assumptions about the order of operations.  I.e., it may test the rename first - but you don't want it to do that first thing, the rename must be dependent on the other conditions.  (Force of habit...  In most other languages I've worked on over the years that would be a no-no due to making assumptions about the order of ops...)

It kind of drives me nuts that we assume the order of operations so much.  But it does appear to work.  

(E.g., every time we do an  if(!empty($x) && $x > 5) I'm shocked we don't have variable not set issues because it may in fact evaluate the $x > 5 before confirming $x is set...  It really does seem to go left to right.)

sbulen:
It'll probably take me an hour or two to test thoroughly.  So before I make changes, I'd like to confirm...   

The only change being requested is to scooch the rename up into the prior if, right?   Removing an if?

Sesquipedalian:
> The only change being requested is to scooch the rename up into the prior if, right? Removing an if?

No, the main change being requested is that you use `pathinfo($file, PATHINFO_EXTENSION)` rather than `substr($file, -4)` to extract the file extension.

> PHP says you shouldn't make assumptions about the order of operations

Er, where does it say that? I'm pretty sure that inside `if` conditions, the order is always left to right. In functions that _do_ something (e.g. `list()`) it might happen in either direction, but that's a different type of situation. In _control_ structures like `if`, not having a predictable order of operations to control things with would render the language completely unusable.

sbulen:
Will do.  I'll get to this tonight.  

Regarding order of operations, the part that stuck out in my mind was:
> Operator precedence and associativity only determine how expressions are grouped, they do not specify an order of evaluation. PHP does not (in the general case) specify in which order an expression is evaluated and code that assumes a specific order of evaluation should be avoided, because the behavior can change between versions of PHP or depending on the surrounding code.

From:
http://php.net/manual/en/language.operators.precedence.php

OTOH, the section on that page regarding 'associativity' does state that && is 'left' associated, e.g., A && B && C is evaluated as (A && B) && C.   But the quote above is pretty clear that grouping doesn't imply order of evaluation either, so even the left associativity shouldn't make a difference, i.e., C can still be evaluated first.  

I am an old-timer, and assuming evaluation left-to-right was always a rookie mistake that usually led to problems.  

Old habits...

(And, for the record, having the true logical order & dependency expressed in the if is far more readable in my mind!  ;^)  )

Sesquipedalian:
Well, there you go. Alright then, based on that, sticking with the three line structure makes sense.

sbulen:
Done.  Tested on win & unix.

For the record, .tmp did work on unix.

Sesquipedalian (approved):

PR #4403, opened 2017-12-10
Remove defined E_STRICT check

albertlast:
because since php 5.4 is E_STRICT in E_ALL included
http://php.net/manual/de/function.error-reporting.php

Would be also a good idea to not set the error_reporting level when not wanted (read the setting file),
but this not the scope of this pr.


colinschoen (approved):
LGTM approval

Issue #4402, opened 2017-12-10
Found a each (deprecated on php 7.2)


PortaMx:
list ($k, $v) = each($_GET); on line 18?? in Load.php ....

colinschoen:
Closing...

PR #4404, opened 2017-12-10
removed each() fixing #4402

albertlast:
since we call each once should be this call the same
related issue #4402

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -1851,7 +1851,8 @@ function loadTheme($id_theme = 0, $initialize = true)
 				redirectexit('wwwRedirect');
 			else
 			{
-				list ($k, $v) = each($_GET);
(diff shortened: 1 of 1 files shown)


colinschoen (approved):
LGTM approval

jdarwood007:
Am I missing something.  I don't understand how this code is supposed to work?  Not what @albertlast  has fixed, but in general.

I think its trying to pull al the $_GET data back into a string to pass back to the redirectexit.  But if I understand that correctly, its just pulling the first key and value.  Which means multiple parameters in GET are not being returned in this case.

albertlast:
You understand the code is correct,
it's take only the first pair of get parameter (because each is called once).

For me and this pr is only the focus/scope to remove the each,
when you want to change the logic feel free to create a own pr.

PR #4405, opened 2017-12-10
Unset Internal Attachment Details

MissAllSunday:
Removes stuff that shouldn't be passed to the client.


colinschoen (approved):
LGTM approval

colinschoen:
Nice find.

PR #4412, opened 2017-12-18
Less queries

albertlast:
I take some time to find some unnessary queries,
found only two queries (it's hard to search for them).

MessageIndex i was able to rewrite the query to the existing on by using of correlated subquery

ProfileView evade a useless roundtrip when both trips do the same

### DIFF
--- a/Sources/MessageIndex.php
+++ b/Sources/MessageIndex.php
@@ -339,11 +339,18 @@ function MessageIndex()
 		$message_index_tables = array();
 		$message_index_wheres = array();
 		call_integration_hook('integrate_message_index', array(&$message_index_selects, &$message_index_tables, &$message_index_parameters, &$message_index_wheres, &$topic_ids));
+		
+		if (!empty($modSettings['enableParticipation']) && !$user_info['is_guest'])
+			$enableParticipation = true;
+		else
+			$enableParticipation = false;
 
 		$result = $smcFunc['db_query']('substring', '
 			SELECT
 				t.id_topic, t.num_replies, t.locked, t.num_views, t.is_sticky, t.id_poll, t.id_previous_board,
 				' . ($user_info['is_guest'] ? '0' : 'COALESCE(lt.id_msg, COALESCE(lmr.id_msg, -1)) + 1') . ' AS new_from,
+				' . ( $enableParticipation ? ' COALESCE(( SELECT 1 FROM {db_prefix}messages parti WHERE t.id_topic = parti.id_topic and parti.id_member = {int:current_member} LIMIT 1) , 0) as is_posted_in,
+				'	: '') . '
 				t.id_last_msg, t.approved, t.unapproved_posts, ml.poster_time AS last_poster_time, t.id_redirect_topic,
 				ml.id_msg_modified, ml.subject AS last_subject, ml.icon AS last_icon,
 				ml.poster_name AS last_member_name, ml.id_member AS last_id_member,' . (!empty($settings['avatars_on_indexes']) ? ' meml.avatar, meml.email_address, memf.avatar AS first_member_avatar, memf.email_address AS first_member_mail, COALESCE(af.id_attach, 0) AS first_member_id_attach, af.filename AS first_member_filename, af.attachment_type AS first_member_attach_type, COALESCE(al.id_attach, 0) AS last_member_id_attach, al.filename AS last_member_filename, al.attachment_type AS last_member_attach_type,' : '') . '
@@ -512,7 +519,7 @@ function MessageIndex()
 				'is_locked' => !empty($row['locked']),
 				'is_redirect' => !empty($row['id_redirect_topic']),
 				'is_poll' => $modSettings['pollMode'] == '1' && $row['id_poll'] > 0,
-				'is_posted_in' => false,
+				'is_posted_in' => ($enableParticipation ? $row['is_posted_in'] : false),
 				'is_watched' => false,
 				'icon' => $row['first_icon'],
 				'icon_url' => $settings[$context['icon_sources'][$row['first_icon']]] . '/post/' . $row['first_icon'] . '.png',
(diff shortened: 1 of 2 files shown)


jdarwood007:
How would this affect queries for big forums?

albertlast:
I hope you mean the messageindex change:

You could awnser the question by your self.
The old query use a group by + limit 20
 which result in that you select between 0 and unlimited(user can be write more than once in a topic) rowset on disk side.
The subquery runs 20 times (when the limit is 20) and select between 0 and 20 rowsets(because of the limit 1 without group by).

And also the overhead(build qry, run qry, evaluate) is gone in this solution. 

in short:
This in improvment is specialy for big boards.

In my run test of the "big" qry is take the same time as before.
and i got a mid size test board ~500k messages.

jdarwood007:
Yes I was referring to the messageindex change.  That is a heavy query and a frequent one ran.  So on Big Boards they will notice if that query becomes unoptimized.
Generally because of how heavy things like this and whos viewing queries can take on a database, most big boards don't use them.

Sounds good if we are improving it then and big boards should see an improvement.

colinschoen:
I'll review this soon.

Sources/MessageIndex.php
```diff
@@ -339,11 +339,18 @@ function MessageIndex()
...
+		if (!empty($modSettings['enableParticipation']) && !$user_info['is_guest'])
+			$enableParticipation = true;
+		else
+			$enableParticipation = false;
 
 		$result = $smcFunc['db_query']('substring', '
 			SELECT
 				t.id_topic, t.num_replies, t.locked, t.num_views, t.is_sticky, t.id_poll, t.id_previous_board,
 				' . ($user_info['is_guest'] ? '0' : 'COALESCE(lt.id_msg, COALESCE(lmr.id_msg, -1)) + 1') . ' AS new_from,
+				' . ( $enableParticipation ? ' COALESCE(( SELECT 1 FROM {db_prefix}messages parti WHERE t.id_topic = parti.id_topic and parti.id_member = {int:current_member} LIMIT 1) , 0) as is_posted_in,
```
colinschoen: Please include the `AS` for consistency. 

`SELECT 1 FROM {db_prefix}messages parti` -> `SELECT 1 FROM {db_prefix}messages AS parti`

Sources/Profile-View.php
```diff
@@ -699,7 +699,7 @@ function showPosts($memID)
 		}
 
 		// Make sure we quit this loop.
-		if ($smcFunc['db_num_rows']($request) === $maxIndex || $looped)
+		if ($smcFunc['db_num_rows']($request) === $maxIndex || $looped || $range_limit == '')
```
colinschoen, line 702: This is more of a question with the existing code than your change:

Am I missing something? The body of this loop seems to ever run the same query 1-2 times. Why is this `while` loop in place?

colinschoen (changes requested):

Sources/Profile-View.php
albertlast: The first run try something,
when it not successed it try to do the same again but with range_limit  = ''.
Dunno how i get to the state that range_limit is in the first run not empty.

Sources/MessageIndex.php
albertlast: the reason consistency is the only valid reason to do so.
because it not needed technicaly.

colinschoen (approved):
LGTM approval

I can take a closer look at the existing code later. Your changes look fine.

Arantor:
Does this actually make it faster? I can imagine on Postgres that it would, but my initial testing says it makes it slower on MySQL, sometimes more than 3 times slower (with SQL_NO_CACHE to avoid the cache affecting things). Not to mention that a user can configure how many rows per page on the message index meaning that it's possible the subquery can be more like 50 times rather than 20.

In addition, the single query also should be able to be satisfied on MySQL solely from the index, which it now won't be. I'd be interested in getting real world numbers from MySQL seeing how it makes up 99.99% of the install base...

albertlast:
1. When you do database benchmark you always expected that the data is warm,
so the use of SQL_NO_CACHE is a mistake from the start.

2. Since the old solution use a group by the amount of selected rowset >= shown rowset
to get a better fealing:
the second query in the old solution select rowset between 0 - max_amount_row_of_messages
new version the subquery select works without group by selected rowset = shown rowset
-> the amount of selected data is between 0 - limit(20 or 50)

3. A internal single database select is faster than a php single select

4. when you calc the improvment than you had to sum all times of the queries and php
old: runtime qry 1 + php overhead of qry1  + runtime qry 2 + php overhead of qry 2 = run_time_old
new: runtime qry 1 + php overhead of qry1 = run_time_new

In short i guess in small setup the solution could run maybe slower,
but on medium/larg setup it's faster.
When I do improvment than I got medium/larg setup in mind and not small.

Arantor:
The problem with benchmarking on the assumption that the data is warm is that in MySQL's cache you need to make sure the cache isn't hiding the real performance of the query. The same query run 50 times in a row will run slowly once and considerably faster every subsequent time. Need to eliminate that because you can't rely on the query cache.

I'd still want to see numbers showing that on a MySQL setup, this will actually be faster, and ideally numbers from a real forum, rather than a constructed benchmark because the reality is that a constructed benchmark forum has odd distributions of data that don't match a real forum (virtually every benchmark forum DB I've ever see has a far more equally distributed user:post ratio than ever exists on a real forum, for example)

What happens on a multi million post forum? Does it still scale on that, on MySQL, in the expected way? Sadly I can't get hold of the 80-odd million post database any more, but an anonymised dump from sm.org would be an interesting dataset in terms of proving it does or doesn't scale better. I share SleePy's concerns about it.

I've asked a friend to give me a copy of their 111k post database to see how that kind of thing scales, will see how that goes.

Arantor:
OK, so I've been playing with that database (it's 111k posts from the board index but turns out there's 226k posts in total) a bit to try to get a feel for improvement. Interestingly with a larger database on MySQL, the difference isn't so huge between the two - but it seems to perform about the same, give or take a 0.01 second.

Would still love to see some benchmarking on real MySQL installs but that's the biggest I can get my hands on right now.

PR #4426, opened 2017-12-24
Disable autocomplete on text validation field

albertlast:
The input field where you had to enter the code from the image (registration),
got no autocomplete off,
and in my eyes it's make no sense.


colinschoen (approved):
LGTM approval

PR #4429, opened 2017-12-24
Avoid using fopen when checking cert - sometimes disabled - Fixes #4407

### LINKED ISSUE #4407: SSL certificate detection doesn't work on many shared hosts
Similar to #4406, ssl_cert_found relies on being able to open a stream wrapper (only this time, with a straight stream context creation rather than relying on file_get_contents)

This issues two warnings:

2: fopen(): https:// wrapper is disabled in the server configuration by allow_url_fopen=0

2: fopen(https://example.com/): failed to open stream: no suitable wrapper could be found

What makes this doubly funny is that this is a URL that's already set to https:// in the first place...

sbulen:
Modified the ssl_cert_found routine to avoid using fopen.
Some hosts may have allow_url_fopen disabled in php.ini.

Feedback welcome.  

Tested with allow_url_fopen enabled & disabled.
Tested on wamp & linux.
Tested with & without certs.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -6016,17 +6016,18 @@ function build_regex($strings, $delim = null, $returnArray = false)
  */
  function ssl_cert_found($url) {
 
-	// Ask for the headers for the passed url, but via https...
-	$url = str_ireplace('http://', 'https://', $url) . '/';
-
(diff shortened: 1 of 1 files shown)


colinschoen (approved):
LGTM approval

PR #4430, opened 2017-12-25
Removed Default on Credits field Fixes #4428

### LINKED ISSUE #4428: Install issue:  BLOB, TEXT, GEOMETRY or JSON column 'credits' can't have a default value
I'm attempting a vanilla 2.1 install, latest github version.  

wamp server, MySQL 5.7, php 5.6.

Error message is:  Line #455: BLOB, TEXT, GEOMETRY or JSON column 'credits' can't have a default value

image

albertlast:
Since the field should be never empty : https://github.com/SimpleMachines/SMF2.1/blob/832c57538c8a2ff60e56aa0d1b00c0639cf8dc70/Sources/Packages.php#L1178
Should be not an issue when the field got no default value and is not null able.

Issue #4428

Because this is an mysql special issue (no sql standard) i forgot this limitation all the time.


colinschoen (approved):
LGTM approval

Issue #4425, opened 2017-12-23
Website on registration doesn't work (missing field)


Arantor:
Assuming we fix #4424, the field doesn't work anyway because the website URL part of the field is defined as a 'url' type field and doesn't get output anyway.

![image](https://user-images.githubusercontent.com/1156911/34323292-c845823c-e839-11e7-8fff-a1174cf99377.png)

frandominguezl:
Working on a fix now for #4424 and #4425

frandominguezl:
Alright, #4424 was an easy fix, but trying to fix this one ends up in weird stuff. If you set the field type to 'text', the field is displayed but it won't keep the website data after you register. So, something is wrong with 'url' type.

Plus, if you fail to solve the captcha, when the screen reloads, the website info is lost (while your username and email are still there).

frandominguezl:
This can be closed @colinschoen

PR #4432, opened 2017-12-25
Doing website fields in register more consistent

### LINKED ISSUE #4424: Website on registration
Not sure why you would do this, but if you selected for a user to add website on registration from the profile fields area, it won't work - because the user doesn't get given the extra permission for profile_website_own inside Register.php to futz with the check made in the profile code.

frandominguezl:
This Pull Request fixes #4424 and #4425 

Thanks to @albertlast for helping me with the fixes for #4425


frandominguezl:
I used Atom to submit this Pull Request and made it via GitHub Desktop and I'm getting again the same error with Travis... "End of file contains extra spaces" while everything is good...

Issue #4444, opened 2017-12-29
Session 'failed to decode' messages


sbulen:
Not all fields in the session are properly encoded.  This results in occasional 'failed to decode' messages.  

I believe the bug is this:  the $_SESSION['last_ss'] is populated with the contents of the last search.  When UTF8 characters are searched for, they are not properly encoded here.   The session encoding & decoding has a problem with this.  

To reproduce:  Search for any special UTF8 characters...   Try this: 'Река неслася; бедный чёлн'

This occurs in 2.1 & in 2.0 at the moment.  This even occurs if your DB is properly setup as UTF8.  

Note that everything works perfectly.  Your search works!  But this message is logged in the background & your session is destroyed & rebuilt.

colinschoen:
Good find.

albertlast:
I found by this description an issue in the postgresql search stuff,
but i'm not able to create any session issue by using this information.

I tested this in my mysql and pg env.
My testet scenario:
Fulltext on and off
Quick Search box and normal search window
Mysql and PostgreSQL

sbulen:
Note from @Arantor over on SMF:

> Try changing the @ini_set('session.serialize_handler', 'php'); to @ini_set('session.serialize_handler', 'php_serialize');
> 
> Requires PHP 5.5.4

albertlast:
Did it solve this?

sbulen:
@albertlast - Hey, I just woke up & got my coffee!   I haven't tried it yet...   

Also, I've been doing a lot of reading first...    I suspect this will work, as most threads that discuss this setting state it will address "known problems" with session_encode with "special characters".   (Some threads suggest hijacking the session encoding algorithm altogether, as they don't think this is sufficient.  Lots of threads out there with alternative session encoding approaches...)

In hindsight, I believe this is the very reason SMF rarely stuffs content into $_SESSION at all.  So I suspect there may be a better way, one that doesn't require bumping the minimum php version.   

Still thinking...

sbulen:
@albertlast - To reproduce, simply search for a string that has multibyte characters:
![utf8-search-issue-1](https://user-images.githubusercontent.com/23568484/34463787-068678f2-ee1c-11e7-9bde-0ac4ad383eaa.JPG)
You will then see this error in the error log:
![utf8-search-issue-2](https://user-images.githubusercontent.com/23568484/34463788-09f91364-ee1c-11e7-80e1-02b923cc9dcf.JPG)
[...]

PR #4445, opened 2017-12-30
Safer way to create a tsquery

albertlast:
By using the test string from issue #4444 
i found an issue on pg fulltextsearch side,
this pr fix this.
But it didn't fix the issue from #4444.


colinschoen (approved):
LGTM approval

PR #4446, opened 2017-12-30
Implement housekeeping interface for Caching fixes #4438

### LINKED ISSUE #4438: Housekeeping interface for Caching
Like I mention here: https://github.com/SimpleMachines/SMF2.1/pull/4306
an "interface" where i could add some housekeeping would be nice.

For pg and sqlite i would like do a monthly/weekly(maybe the admin can choose) run where i truncate complete table.

Maybe @jdarwood007 could you add something like that?

albertlast:
and implement this for pg and also fix a bug in postgres cache table creation

@tinoest since you provide the sqlite cache maybe you could help me or add a own pr to add sqlite logic

Issue #4438


tinoest:
I can later next week. Shouldn’t be to difficult. Why do you move to a new table? 

Why not just delete old entries?

albertlast:
I assume that the amount of data which are olded is "high".

Delete of data is slower and when you delete table of table,
you got data which is marked as deleted on the hdd without be "fully deleted".

See also: https://www.tutorialspoint.com/sqlite/sqlite_truncate_table.htm

jdarwood007:
Wouldn't it be better here to use a transaction for this?  Since it could take a while to create the temp, empty the old table and recreate the new one?  During that time any parallel calls to the cache table would cause some database errors possibly.

albertlast:
Let's go to the possible reason for transaction:
Is the data important? no (we can create the data when ever we like to do so)
What happen when someone select the data in the moment we truncate the table? Nothing he get the data and we truncate this table in the same moment.
Would it possible to do this in a transaction? In common no in the special case of postgresql yes.

The last point is very important for your to know.
SQL define that a transaction is only valid so long you do dml operation(delete, inset, update),
when you do ddl operation (create, alter, truncate...) the transaction get be commited.
But pg is here special and allow ddl inside a transaction.

So in theory could one case happen:
We truncate the table,
after this a user create a new cache entry and
then we insert the data from temp table and in the temp exists the samedata set(same key maybe different value).
In my eyes a very small chance that this happen,
but it's not hard catch this issue.
So i extend the pr about this.

jdarwood007:
Fair enough!  Was just thinking about a race condition that could occur on very busy websites.  Although I doubt they would be using pg caching.  Most likely would be using another caching method then.

albertlast:
@colinschoen @Sesquipedalian can we merge this?

Sesquipedalian:
@colinschoen assigned himself to review this, so I've been leaving it to him. :)

Sesquipedalian (approved):

Issue #4451, opened 2017-12-31
Gender Field


wintstar:
Last Pull request: 
https://github.com/SimpleMachines/SMF2.1/pull/4445

Forum version: SMF 2.1 Beta 4 (more detailed)
Current SMF version: SMF 2.1 Beta 3
GD version: bundled (2.1.0 compatible) 
MySQLi engine: MariaDB 
MySQLi version: 10.1.25-MariaDB 
Alternative PHP Cache: 5.1.2 
PHP: 7.1.11 (more detailed) 
Server version: Apache/2.4.29 (Win32) OpenSSL/1.0.2l PHP/7.1.11 
Fix: 
https://github.com/SimpleMachines/SMF2.1/pull/4383

Fix not work with the new pull request today. Not by new install or update. Key is not read from the database. This field is not displayed in the post view.

![gender_2](https://user-images.githubusercontent.com/2753863/34460868-009a76f0-ee1c-11e7-89d2-ba3bfc6a0eb2.png) 

![gender_1](https://user-images.githubusercontent.com/2753863/34460869-02b49254-ee1c-11e7-879d-ecae16c1e48d.png) 

![gender_3](https://user-images.githubusercontent.com/2753863/34460870-04545914-ee1c-11e7-8743-799c0ddd503b.png) 

![gender_4](https://user-images.githubusercontent.com/2753863/34461077-cab536f0-ee21-11e7-8da2-9cea87813d63.png)

wintstar:
It would be much easier to set this as the default field. Why the change from 2.0. *? 
It worked in all languages. 

> Never change a runing system

albertlast:
Soi the pr got merge,
please test this again with the changes(notice that also database changes are in ther)

wintstar:
Database was updated with upgrader. It works with the post view, not work in memberlist and not in profile view. With the default language English set in the profile, it does not work there either. There is always the problem with "KEY".

![gender_7](https://user-images.githubusercontent.com/2753863/34638816-0436458e-f2d4-11e7-8f7f-aa7e89d4d830.png) 

![gender_6](https://user-images.githubusercontent.com/2753863/34638819-0d13aa34-f2d4-11e7-8a0c-663424a17c16.png)

![gender_5](https://user-images.githubusercontent.com/2753863/34638822-14f5352e-f2d4-11e7-8e7b-a44a0eacf320.png)

albertlast:
it doesn't solve the issue,
[...]

PR #4452, opened 2017-12-31
Fix Broken Logic for Loading Profile Fields

colinschoen:
I noticed these issues when investigating issue #4451. This doesn't fix that issue, but addresses other issues in PR #4383. 

Specifically this adds the `"options":["None","Male","Female"]` key to the serialized JSON for the `displayFields` setting. 

It also fixes some logic where the `currentKey` variable was being improperly set for custom fields with multiple options.

@MissAllSunday I echo the concerns of @albertlast after looking into this more... are you sure that we don't need any changes in the upgrade script? If you can please take a look to review this and have a look to confirm we don't need changes into the upgrade script that would be helpful :). 


colinschoen:
@albertlast If you have time please do review as well. I'd like another set of eyes before merging.

albertlast (approved):

colinschoen:
Thanks for reviewing.

MissAllSunday:
yep, totally forgot about displayfields. That should be the only thing to consider when upgrading

PR #4457, opened 2018-01-01
Set user_info language while install process fixes #4453

### LINKED ISSUE #4453: Installer step 5 umlauts bug
Last Pull request:
#4445

Forum version: SMF 2.1 Beta 4 (more detailed)
Current SMF version: SMF 2.1 Beta 3
GD version: bundled (2.1.0 compatible)
MySQLi engine: MariaDB
MySQLi version: 10.1.25-MariaDB
Alternative PHP Cache: 5.1.2
PHP: 7.1.11 (more detailed)
Server version: Apache/2.4.29 (Win32) OpenSSL/1.0.2l PHP/7.1.11
Browser: Chrome Version 63.0.3239.108 (64-Bit)

Step 1 to 4 have the charset utf-8. But step 5 have charset ISO-8859-1.  That makes the bug with the German umlauts.  [The German language files correspond to the file encoding utf-8](http://www.stephan-frank.de/board/index.php/topic,11.0.html).

umlauts_bug_installer

albertlast:
In the last step we do scheduled_fetchSMfiles
while the process the loadLanguage() is called,
which didn't find any "loaded" language and try to guess which is loaded/or could be.
Which got than the issue we get a diffrent language is loaded as we use in the setup process.

So i extend the load_lang_file about the information which language we load (was suprised that is no otherwhere already known) and
set this language as $user_info['language'] so that the loadLanguage() know which language we like.

issue: #4453

### DIFF
--- a/other/install.php
+++ b/other/install.php
@@ -325,7 +325,7 @@ function load_lang_file()
 	// And now include the actual language file itself.
 	require_once(dirname(__FILE__) . '/Themes/default/languages/' . $_SESSION['installer_temp_lang']);
 	
-	// Which language we load? And also assume that he like his language.
+	// Which language did we load? Assume that he likes his language.
 	preg_match('~^Install\.(.+[^-utf8])\.php$~', $_SESSION['installer_temp_lang'], $matches);
 	$user_info['language'] = $matches[1];
 }
--- a/other/install.php
+++ b/other/install.php
@@ -259,7 +259,7 @@ function initialize_inputs()
 // Load the list of language files, and the current language file.
 function load_lang_file()
 {
-	global $txt, $incontext;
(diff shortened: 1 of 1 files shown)


other/install.php
```diff
@@ -324,6 +324,10 @@ function load_lang_file()
 
 	// And now include the actual language file itself.
 	require_once(dirname(__FILE__) . '/Themes/default/languages/' . $_SESSION['installer_temp_lang']);
+	
+	// Which language we load? And also assume that he like his language.
```
colinschoen: nit: Which language did we load? Assume that he likes his language.

colinschoen (changes requested):
Nit

colinschoen (approved):
LGTM approval

PR #4461, opened 2018-01-01
Extend target="_blank" about rel="noopener" Fixes #4460

### LINKED ISSUE #4460: Secure external links
Wouldn't it be better to make external links (e. g. with BBCode) make safer with rel=noopener?

Look here https://mathiasbynens.github.io/rel-noopener ?

albertlast:
If you like the idea from #4460 than this would be the pr for this.
This change effect also existingslinks


albertlast:
Since i changed also the copyright text, travis failed.

colinschoen:
This shouldn't be necessary for internal links (non-user generated links point to SMF pages) right?

albertlast:
Sure they are needed,
where you know that sm.org links to sm.org?
specialy when you don't use https.

colinschoen:
We know that `$scripturl` links to sm.org. E.g. This isn't some user generated URL so why is this useful to add the `rel=noopener` attribute [here](https://github.com/albertlast/SMF2.1/blob/cbf849b37fccdc450456ddd4bd8520b6d39b2293/Sources/Logging.php#L306)?

albertlast:
Your don't know where the dns shows.

colinschoen:
I suppose that is true (but would be a pretty involved attack requiring control of the DNS resolver the victim is using) however this link is not using `target="_blank"` so it isn't opening in a new page so why is having control of `window.opener` even useful at this point? I guess technically the user could use the back button and maybe `window.opener` attacks could still be successful? I'm not even sure if that would work.

albertlast:
And what is the issue to use even for sm.org side this technical?
And sure use the link target="_blank" .

colinschoen:
Not sure how I missed that `target="_blank"` is indeed there for all of the links you changed :). This LGTM I don't see how it could hurt.

colinschoen (approved):
LGTM approval

I will handle updating the travis build script.

colinschoen:
Will be merging once https://github.com/SimpleMachines/BuildTools/pull/3 is approved.

PR #4463, opened 2018-01-01
Use UTF-8 in setup and index Fixes #4454

### LINKED ISSUE #4454: ISO-8895-1 in install and index.english.php
i had to ask if ther any reason why we got in
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Themes/default/languages/index.english.php#L19
and
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Themes/default/languages/Install.english.php#L5
ISO-8895-1 instead of UTF-8?

albertlast:
With this change your able to use utf chars in the middle of install process,now.
Which i think is good specialy for the case when no lang pack exists for your language.

issue: #4454

### DIFF
--- a/Themes/default/languages/Install.english.php
+++ b/Themes/default/languages/Install.english.php
@@ -2,7 +2,7 @@
 // Version: 2.1 Beta 4; Install
 
 // These should be the same as those in index.language.php.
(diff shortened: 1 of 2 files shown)


colinschoen (approved):
LGTM approval

PR #4468, opened 2018-01-03
Fixes #4462

### LINKED ISSUE #4462: Weird displaying in Register
After enabling Timezone field to show in register, this is how it displays:

Gwenwyfar:
plus issues with other inputs


colinschoen (approved):
LGTM approval

frandominguezl:
The fix works wood, but wouldn't it be better to set it to 50% instead of 100%?
It looks kinda weird for me that some "boxes" are from size X, and then that one has a different size.

Gwenwyfar:
This is only a general fix for the entire forum so nothing "breaks loose" in any situation. Defining individual sizes are more on a case-by-case basis. I'll look into that for the upcoming revisions :)

PR #4471, opened 2018-01-05
SMF needs more socks to keep its feet warm.

Archer70:
Click any part of the body 25 times in less than 15 seconds. Enjoy. 😁


wintstar:
Sorry to ask But what does this have to do with SMF?

decoded

Spam?

XinYenFon:
Hmm it looks like a nice easter egg ^^

@wintstar, @Archer70 is a trusted friend and developer.

illori:
i really dont see the need to add more easter eggs to the code. at one point it was suggested to remove almost all of them as they are not found and not necessary. we are trying to move toward a stable release we dont need to be adding unnecessary things at this point.

frandominguezl (approved):

colinschoen:
Let's put this on hold until after RC1. If we don't have any issues with that we can look into adding some fun things. Perhaps the community can have some input on this. For now I will close, but happy to consider at a later date.

frandominguezl:
@Sesquipedalian #BringTheSocksBack

PR #4472, opened 2018-01-05
Database Debug time moved, added php timer and get the code from viewquery inline with logging

albertlast:
Something i don't like was that the "start" time of the db runtime was middle of road,
so php code came afterwards and some where in the end of road is than the query,

So i moved the "start timer" for the query directly in front of query fire statement.
Than i remember in the history we got sometimes,
very slow code php/replacement code.
To know this would be nice.

So i set another start point of start the function to know how mutch time it take on the php side.
Reason why i had to extend the language file, logging and viewquery,
to show this new information.

the result look so:
grafik
i'm not 100% sure if the php stuff is needed,
but to place the start timer for db on a reasonable places is in my eye very important.


Themes/default/languages/index.english.php
```diff
@@ -847,7 +847,7 @@
 $txt['debug_queries_used_and_warnings'] = 'Queries used: %1$d, %2$d warnings.';
 $txt['debug_query_in_line'] = 'in <em>%1$s</em> line <em>%2$s</em>, ';
 $txt['debug_query_which_took'] = 'which took %1$s seconds.';
-$txt['debug_query_which_took_at'] = 'which took %1$s seconds at %2$s into request.';
+$txt['debug_query_which_took_at'] = 'which took %1$s seconds in php %2$s in db at %3$s into request.';
```
Sesquipedalian: 'which took %1$s seconds to prepare in PHP and %2$s to execute in the database at %3$s into request.'

Sources/ViewQuery.php
```diff
@@ -74,11 +74,11 @@ function ViewQuery()
 	<body id="help_popup">
 		<div class="tborder windowbg description">';
 
-	foreach ($_SESSION['debug'] as $q => $query_data)
+	foreach ($_SESSION['debug'] as $q => $qq)
```
Sesquipedalian: Nit: `$query_data` is a more descriptive name for a variable than `$qq`. Please change it back so that future developers will have an easier time reading your code. :)

Sesquipedalian (changes requested):

Sources/ViewQuery.php
albertlast: it's not my code,
this is the var name of logging.php

Sesquipedalian (approved):

PR #4474, opened 2018-01-06
Added httpsOn function how is checking if the connection fixes #4473

### LINKED ISSUE #4473: Error with Forcing SSL
After I installed SMF (using version 3b2b583), I tried to visit the index page but failed.
image
I enabled the HTTPS service but it still said "You can only login via HTTPS".
It's really confusing.
BTW the environment I'm using is as follows.
Virtual Host(Linux), Nginx, MySQL 5.5.35, PHP 7.0
Any things I can help?

albertlast:
to the webserver or load balancer is done by https.

Since the check got more meat to check and it used on many placed i decide to create a function how got
the complet logic.
I'm not fully able to test this change because i got no ssl env.

Could also fix the issue #4473

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -6181,12 +6181,12 @@ function build_query_board($userid)
 function httpsOn()
 {
 	$secure = false;
-	
+
 	if (isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] == 'on') 
 		$secure = true;
 	elseif (!empty($_SERVER['HTTP_X_FORWARDED_PROTO']) && $_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https' || !empty($_SERVER['HTTP_X_FORWARDED_SSL']) && $_SERVER['HTTP_X_FORWARDED_SSL'] == 'on') 
 		$secure = true;
-	
+
 	return $secure;
 }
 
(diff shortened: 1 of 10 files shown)


Sefank (approved):
I've tested and the bug was fixed.

jdarwood007:
Great catch on that.  Load balancers would have indeed failed without that forward check.

LGTM

Sources/Subs.php
```diff
@@ -6173,4 +6173,21 @@ function build_query_board($userid)
...
+ * 
+ * @return boolean true if connection used https
+ */
+function httpsOn()
+{
+	$secure = false;
+
+	if (isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] == 'on') 
+		$secure = true;
+	elseif (!empty($_SERVER['HTTP_X_FORWARDED_PROTO']) && $_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https' || !empty($_SERVER['HTTP_X_FORWARDED_SSL']) && $_SERVER['HTTP_X_FORWARDED_SSL'] == 'on')
```
Sesquipedalian, line 6187: Rather than `elseif`, why not just use `||` in the `if`'s condition?

albertlast: first because it's suck to read,
secondly forward stuff is because of load balancer.(to make clear where the https is)
third performance wise shouldn't be a point -> read able > performance

Sesquipedalian: Fair enough.

PR #4478, opened 2018-01-07
Error count only when admin fixes #4477

### LINKED ISSUE #4477: Error count query runs way too often
The query that fetches the error count for the admin menu runs in reloadSettings. This is way too early to be running it, because now it means that regardless of what happens on that page, you've run that query.

It only needs to run if the user is an admin (since they won't be able to get to the error log otherwise anyway), and it doesn't actually need to run any earlier than setupMenuContext as that's the only place it's really needed. (Though it might be prudent to rename it from $context['num_errors'] if it's moved, to avoid messing with the error log page itself that also uses the same item name)

albertlast:
This is pr to the issue: #4477
@Arantor Since you request the changes, please test this pr.


Arantor:
I didn't 'request the change', I saw what looked to me like a problem. The reality is that I won't be running 2.1 in production, but until I finish the refactoring I'm doing, I might as well point out the issues that I find along the way.
 
That said, running the query in the log_error function is only of value if the error happens before the menu is built; if it happens after that it won't be incremented anyway.

Sesquipedalian:
> That said, running the query in the log_error function is only of value if the error happens before the menu is built; if it happens after that it won't be incremented anyway.

True, but the way @albertlast has this code, the query won't be run in `log_error()` once `$context['num_errors']` has been defined in `setupMenuContext()`. The subsequent incrementation is still useless, but there won't be a useless query.

If we really want to make sure that the error count indicator reflects all the errors generated along the way, we probably need to append a placeholder during `setupMenuContext()` and then insert the final value of `$context['num_errors']` into it during `obExit()`.

Arantor:
True, if it occurs at that point, the query is not run. But if it occurs before setupMenuContext, it will be off by one since the insert happens, then the fetch count then an increment...

Sources/Errors.php
```diff
@@ -123,6 +123,19 @@ function log_error($error_message, $error_type = 'general', $file = null, $line
...
+			$query = $smcFunc['db_query']('', '
+				SELECT COUNT(id_error)
+				FROM {db_prefix}log_errors',
+				array()
+			);
+
+			list($context['num_errors']) = $smcFunc['db_fetch_row']($query);
+			$smcFunc['db_free_result']($query);
+		}
+
```
Sesquipedalian: Should have an `else` here.

Sesquipedalian (changes requested):

Sources/Errors.php
albertlast: Nope it get last number from db,
but this part is called when a new error come in.
--> get the db number and add one to get the new count.

Sesquipedalian: The latest error was already inserted into the database on line 123 above, so the query retrieves the correct error number at that moment. Incrementing the variable after the query sets it to one more than the number of errors.

For example, suppose that the error log already has 3 errors recorded in it when the script starts. Then a new error occurs sometime before we reach `setupMenuContext()`. Here's what will happen:

1. `$smcFunc['db_error_insert']($error_info);` inserts the new error into the database. There are now 4 errors recorded in the database.
2. `list($context['num_errors']) = $smcFunc['db_fetch_row']($query);` sets `$context['num_errors']` to 4.
3. `$context['num_errors']++;` sets `$context['num_errors']` to 5.

albertlast:
Stupid github editor added something wrong at the eof -.-
need to edited again...

Sesquipedalian (approved):

PR #4481, opened 2018-01-10
Cookie

### LINKED ISSUE #4297: 3 Login Errors

### LINKED ISSUE #4362: corrupt cookies doesn't get fixed
When a mod change the cookie, an old version get the smf cookie in a different format or a bad guy,
then the cookie doesn't get fixed,
since we fix only cookies when the are not corrupted: https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs-Auth.php#L41

Possible fix is provided in pr https://github.com/SimpleMachines/SMF2.1/pull/4358

### LINKED ISSUE #4364: Cookie will be not destroyed when you change from global to local
Since https://github.com/SimpleMachines/SMF2.1/pull/4363 you introduce
that the cookie get not deleted when you change from global to local cookie

### LINKED ISSUE #4365: While login a new session id is generated

### LINKED ISSUE #4366: Cookie will be not destroy when the board move to subfolder

### LINKED ISSUE #4367: Json object from cookie date

### LINKED ISSUE #4368: Unsafe cookie deleting

Sesquipedalian:
Fixes #4297 
Fixes #4362 
Fixes #4364 
Fixes #4365
Fixes #4366 
Fixes #4367 
Fixes #4368

### DIFF
--- a/Sources/Subs-Auth.php
+++ b/Sources/Subs-Auth.php
@@ -60,8 +60,14 @@ function setLoginCookie($cookie_length, $id, $password = '')
 	}
 
 	// Get the data and path to set it on.
-	$data = $smcFunc['json_encode'](empty($id) ? array(0, '', 0, $cookie_state) : array($id, $password, time() + $cookie_length, $cookie_state));
+	$data = empty($id) ? array(0, '', 0, $cookie_state) : array($id, $password, time() + $cookie_length, $cookie_state);
 	$cookie_url = url_parts(!empty($modSettings['localCookies']), !empty($modSettings['globalCookies']));
+
+	// Allow mods to add custom info to the cookie
+	$custom_data = array();
+	call_integration_hook('integrate_cookie_data', array($data, &$custom_data));
+
+	$data = $smcFunc['json_encode'](array_merge($data, $custom_data));
 
 	// Set the cookie, $_COOKIE, and session variable.
 	smf_setcookie($cookiename, $data, time() + $cookie_length, $cookie_url[1], $cookie_url[0]);
--- a/Sources/Subs-Auth.php
+++ b/Sources/Subs-Auth.php
@@ -36,34 +36,41 @@ function setLoginCookie($cookie_length, $id, $password = '')
 	// If changing state force them to re-address some permission caching.
 	$_SESSION['mc']['time'] = 0;
 
+	// Extract our cookie domain and path from $boardurl
+	$cookie_url = url_parts(!empty($modSettings['localCookies']), !empty($modSettings['globalCookies']));
+
 	// The cookie may already exist, and have been set with different options.
 	$cookie_state = (empty($modSettings['localCookies']) ? 0 : 1) | (empty($modSettings['globalCookies']) ? 0 : 2);
 	if (isset($_COOKIE[$cookiename]))
 	{
 		// First check for 2.1 json-format cookie
-		if (preg_match('~^{"0":\d+,"1":"[0-9a-f]*","2":\d+(,"3":\d\b)?~', $_COOKIE[$cookiename]) === 1)
-			$array = $smcFunc['json_decode']($_COOKIE[$cookiename], true);
+		if (preg_match('~^{"0":\d+,"1":"[0-9a-f]*","2":\d+,"3":\d\b,"4":"[^"]+"~', $_COOKIE[$cookiename]) === 1)
+			$array = @$smcFunc['json_decode']($_COOKIE[$cookiename], true);
 
 		// Legacy format (for recent 2.0 --> 2.1 upgrades)
 		elseif (preg_match('~^a:[34]:\{i:0;i:\d+;i:1;s:(0|128):"([a-fA-F0-9]{128})?";i:2;[id]:\d+;(i:3;i:\d;)?~', $_COOKIE[$cookiename]) === 1)
-			$array = safe_unserialize($_COOKIE[$cookiename]);
+			$array = @safe_unserialize($_COOKIE[$cookiename]);
 
 		// Out with the old, in with the new!
-		if (isset($array[3]) && $array[3] != $cookie_state)
+		if (isset($array[3]) && $array[3] != $cookie_state || isset($array[4]) && $array[4] != $cookie_url[1])
 		{
 			// Maybe we need to temporarily pretend to be using local cookies
 			if ($cookie_state == 0 && $array[3] == 1)
-				$cookie_url = url_parts(true, false);
+				$old_cookie_url = url_parts(true, false);
 			else
-				$cookie_url = url_parts($array[3] & 1 > 0, $array[3] & 2 > 0);
+				$old_cookie_url = url_parts($array[3] & 1 > 0, $array[3] & 2 > 0);
 
-			smf_setcookie($cookiename, $smcFunc['json_encode'](array(0, '', 0, $cookie_state), JSON_FORCE_OBJECT), 1, $cookie_url[1], $cookie_url[0]);
+			// If the paths don't match, feed ALL the cookies to the Cookie Monster!
+			$paths = rsort(array_unique(array($cookie_url[1], $old_cookie_url[1]), $array[4]));
+
+			// OM NOM NOM
+			foreach ($paths as $path)
+				smf_setcookie($cookiename, $smcFunc['json_encode'](array(0, '', 0, $cookie_state, $path), JSON_FORCE_OBJECT), 1, $path, $cookie_url[0]);
 		}
 	}
 
 	// Get the data and path to set it on.
-	$data = empty($id) ? array(0, '', 0, $cookie_state) : array($id, $password, time() + $cookie_length, $cookie_state);
-	$cookie_url = url_parts(!empty($modSettings['localCookies']), !empty($modSettings['globalCookies']));
+	$data = empty($id) ? array(0, '', 0, $cookie_state, $cookie_url[1]) : array($id, $password, time() + $cookie_length, $cookie_state, $cookie_url[1]);
 
 	// Allow mods to add custom info to the cookie
 	$custom_data = array();
--- a/Sources/Subs-Auth.php
+++ b/Sources/Subs-Auth.php
@@ -38,12 +38,14 @@ function setLoginCookie($cookie_length, $id, $password = '')
 
 	// The cookie may already exist, and have been set with different options.
 	$cookie_state = (empty($modSettings['localCookies']) ? 0 : 1) | (empty($modSettings['globalCookies']) ? 0 : 2);
-	if (isset($_COOKIE[$cookiename]) && preg_match('~^{"0":\d+,"1":"[0-9a-f]*","2":\d+(,"3":\d\b)?~', $_COOKIE[$cookiename]) === 1)
+	if (isset($_COOKIE[$cookiename]))
 	{
-		$array = $smcFunc['json_decode']($_COOKIE[$cookiename], true);
+		// First check for 2.1 json-format cookie
+		if (preg_match('~^{"0":\d+,"1":"[0-9a-f]*","2":\d+(,"3":\d\b)?~', $_COOKIE[$cookiename]) === 1)
+			$array = $smcFunc['json_decode']($_COOKIE[$cookiename], true);
 
-		// Legacy format
-		if (is_null($array))
+		// Legacy format (for recent 2.0 --> 2.1 upgrades)
+		elseif (preg_match('~^a:[34]:\{i:0;i:\d+;i:1;s:(0|128):"([a-fA-F0-9]{128})?";i:2;[id]:\d+;(i:3;i:\d;)?~', $_COOKIE[$cookiename]) === 1)
 			$array = safe_unserialize($_COOKIE[$cookiename]);
 
 		// Out with the old, in with the new!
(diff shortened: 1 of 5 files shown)


Sesquipedalian:
Regarding the ability for mods to add custom data to the cookie, it is important to note that the regexes in this PR don't look for `\]$` or `}$`. This change means that if mods append more data to the end of the `$data` array, the regex will not fail.

Sources/Subs-Auth.php
```diff
@@ -49,15 +49,26 @@ function setLoginCookie($cookie_length, $id, $password = '')
...
 	}
 
 	// Get the data and path to set it on.
-	$data = $smcFunc['json_encode'](empty($id) ? array(0, '', 0) : array($id, $password, time() + $cookie_length, $cookie_state));
+	$data = empty($id) ? array(0, '', 0, $cookie_state) : array($id, $password, time() + $cookie_length, $cookie_state);
 	$cookie_url = url_parts(!empty($modSettings['localCookies']), !empty($modSettings['globalCookies']));
 
+	// Allow mods to add custom info to the cookie
+	$custom_data = array();
+	call_integration_hook('integrate_cookie_data', array($data, &$custom_data));
```
albertlast: Integration hook doesn't work,
when you got the regexp on the otherside which not allow addional stuff.

albertlast:
and why not place a `.*}$` ?

Sesquipedalian:
Because there is no need.

Sesquipedalian:
@sbulen and @albertlast, if you could each test this in order to make sure it works on your configurations, that'd be helpful.

sbulen:
I will definitely help out when I can - at the moment I think I can spend real time on it Saturday.  (Work has mushroomed; and family matters keep pulling me out of town at random times... Sorry about that...)

Sesquipedalian:
There's nothing to apologize for.

albertlast:
I ask my self why you not fixing the issue https://github.com/SimpleMachines/SMF2.1/issues/4368
by using this loc:
https://github.com/albertlast/SMF2.1/blob/937c6d27960e4aa0e09e9dce388927a9f16ae5e7/Sources/Subs-Auth.php#L36

the only one different than is the path logic,
which is a maybe a complex topic but needed to be address in my eyes.
Because to change/delete a cookie you need to know the path,
smf do it by predicting this by using the cookie_state.
Which do the similiar job (it's partly the path information) but it lost the detail information.
This leads into the posibility that you lost the controll about this cookie: https://github.com/SimpleMachines/SMF2.1/issues/4366

So when you make a clean approache you would delete the cookie_state and add path,
maybe for lazy reason you could like me keep the cookie_state and add path,
but cookie_state get useless when you got path.

sbulen:
The path always comes from $boardurl.  Its not a mystery.  The global & local settings just tell you whether to honor its subfolders & subdomains or not.

albertlast:
Like I mention a complex topic,
It's not so easy as sbulen think.
Imagen following process:
You got a forum under localhost/board
And move it to localhost/board/old
Next someone login to your board under the old board in cookie path is now /board
Than came the move the existing cookie keeps their, 
Now the user try to logout smf send to the browser to delete the cookie with path /board/old 
The result is that the existing cookie will be not deleted because there is no cookie with this path.

Sesquipedalian:
> I ask my self why you not fixing the issue #4368

I was waiting to hear from you two whether your tests show that this PR fixes the issues it addresses, and then I was going to ask you to put the changes you made to address #4368 into their own PR based on this one. But I suppose I could just implement them myself in this PR. Either way is fine with me.

> the only one different than is the path logic,
> which is a maybe a complex topic but needed to be address in my eyes.

I agree. The scenario you describe could cause problems. It would only be a problem for users who had logged in forever when the board was moved this way, but even so, that is a problem that we should solve.

> So when you make a clean approache you would delete the cookie_state and add path,
> maybe for lazy reason you could like me keep the cookie_state and add path,
> but cookie_state get useless when you got path.

No, we still need the $cookie_state information even if we include the path information. $cookie_state also tells us what to put in the cookie's domain.

albertlast:
>     So when you make a clean approache you would delete the cookie_state and add path,
>     maybe for lazy reason you could like me keep the cookie_state and add path,
>     but cookie_state get useless when you got path.
> 
> No, we still need the $cookie_state information even if we include the path information. $cookie_state also tells us what to put in the cookie's domain.

This is one of the important shift/changes,
the current cookie_state we get from smf/database based on this we generat the current path and then we check this with the path we got from cookie,
if different -> change the cookie information.
Because of this logic the information in the cookie about the cookie state is not needed any more.

Sesquipedalian:
No, `url_parts()` can return different values for `$cookie_url[0]` depending on the values of its parameters.

albertlast:
jeah and this information came from $modSettings['localCookies'] and $modSettings['globalCookies']
not from the cookie data.

Sesquipedalian:
The logic [here](https://github.com/SimpleMachines/SMF2.1/pull/4481/files#diff-fcc054f557dd135fb1af3324ba04d286R58) uses the value stored in the cookie to determine which parameters to give to `url_parts()`

albertlast:
Yes so long you don't know the path you had to look into the cookie_state which state it got and
based on this you build the path of the cookie.
But when the path is in the cookie_data than the solution is that: https://github.com/albertlast/SMF2.1/blob/937c6d27960e4aa0e09e9dce388927a9f16ae5e7/Sources/Subs-Auth.php#L63
Maybe i should mention in you question about path that i don't only use the path for the issue stuff...,
i use this for any easier logic to address the old and new one cookie also.

Sesquipedalian:
Yes, I see what you mean. But I'm using `$array[3]` on [this line](https://github.com/SimpleMachines/SMF2.1/pull/4481/files#diff-fcc054f557dd135fb1af3324ba04d286R58) to make sure we get the correct _domain_ for the cookie (i.e. `subdomain.example.com` or `.example.com`) from `url_parts()`, not the correct _path_. If I don't store the value of `$cookie_state`, I'll need to store the value of `$cookie_url[0]`. That wouldn't really be any simpler, and it would needlessly break backwards compatibility with the 2.0 style cookies.

Sesquipedalian:
In case it wasn't clear, the reason we need to make sure that we get the correct domain at that part of the code is to address the bug that I described in [this comment](https://github.com/SimpleMachines/SMF2.1/issues/4297#issuecomment-356573593) regarding forums hosted on subdomains.

albertlast:
The cookie_state is as fare i know zero or one,
global or local how this affect this subdomain stuff?
exp: my.board.com/smf
for my understanding was global affected only the path -> / (unrelated where the board ist)
Local : path -> /smf

Sesquipedalian:
Assume the forum is hosted at http://subdomain.example.com/path/to/forum/ in the following sequence of events:

1. The admin enables local storage for the cookie. 
2. The user logs in. SMF sends a cookie to the user's browser with the domain `subdomain.example.com` and the path `/path/to/forum/`.
3. The admin now disables local storage for the cookie.
4. The user returns to the site, and SMF detects that the user's cookie is no longer correct.

The question is, what happens next? 

Right now, this is what will happen next:

5. SMF sends an immediately expiring cookie to the user's browser with an empty domain and path `/`
6. The browser rejects this new cookie, since it doesn't match the existing cookie. The existing cookie is retained.
7. SMF sends another new cookie using the correct login information with an empty domain and path `/`. The browser doesn't accept this either, and continues using the old cookie.
8. The user receives login errors.

If we use [this code](https://github.com/albertlast/SMF2.1/blob/937c6d27960e4aa0e09e9dce388927a9f16ae5e7/Sources/Subs-Auth.php#L63): this is what will happen next:

5. SMF sends an immediately expiring cookie to the user's browser with an empty domain and path `/path/to/forum/`
6. The browser rejects this new cookie, since it doesn't match the existing cookie. The existing cookie is retained.
7. SMF sends another new cookie using the correct login information with an empty domain and path `/`. The browser doesn't accept this either, and continues using the old cookie.
8. The user receives login errors.

If we use [this code](https://github.com/SimpleMachines/SMF2.1/pull/4481/files#diff-fcc054f557dd135fb1af3324ba04d286R58), this is what will happen next:

5. SMF sends an immediately expiring cookie to the user's browser with the domain `subdomain.example.com` and path `/path/to/forum/`.
6. The browser accepts this new cookie. The existing cookie is expired immediately.
7. SMF sends another new cookie using the correct login information with an empty domain and path `/`. The browser accepts this new cookie.
8. The user logs in successfully.

albertlast:
Yeah but you make the same mistake than twice.
- You try to guess that the path could be 
- You try to guess the subdomain

I mean the result is better as the existing code.

But the idea from me is, instead of play the guessing game,
safe the needed information directly in the cookie -> when the path and domain is needed than same both.
i mean you "compress" both information in 0 or 1 sure is ther are lost about information.

Sesquipedalian:
Actually, it is a value of 0, 1, 2, or 3.

sbulen:
My understanding is as follows...   This might be stating the obvious, but it may help...

It's a bit-wise value.  First bit is global & 2nd bit is local.  In binary:
00 = 0 = global off, local off
01 = 1 = global off, local on
10 = 2 = global on, local off
11 = 3 = global on, local on

Global & local are not mutually exclusive, this is just weirdness in terms.   Global is whether the scope is for the entire TLD or whether the cookie is specific to a subdomain.   Local is whether the scope is for a specific subfolder or all subfolders.  

So I don't understand why the UI makes them mutually exclusive, I don't think they are... 

albertlast is right in that it is just a yes/no, and does not specify subdomains or subfolders.  So, actually, the term "compressed" is accurate.   And it is a loss**ful** compression...   You lose information.  

If you really want to redo this, you should just keep it VERY simple and have ONE setting: the scope of the cookie, expressed in a URL format...  If you want all subdomains, then don't specify a subdomain.  If you want all subfolders, then don't specify a subfolder...

sbulen:
I had been hoping to test today, but couldn't get a clean install due to issue #4483 .   I always like working from a clean, current git install as a proper foundation for a test, but I wasn't able to get there.  

I realize I am a lone voice in the wilderness on this topic, but I do think this PR has gotten very complex, and, in my mind, the code has gotten much more difficult to read.  

And most of this complexity has to do with trying to handle forum moves across subdomains/TLD and/or across subfolders vs forum root.   I still maintain this is not a problem.   

I think that, if you move a forum, you may get a stray cookie, and you may have to log in again, and this is _expected behavior_.  And we can & should rely on the SMF garbage cleanup to clear stray sessions & the browser garbage cleanup to clean up stray cookies.  

I actually don't think we have ever had a bug logged by a user upset because s/he had to logon again after moving a forum.  I think they expect that may happen.  So we are adding unnecessary complexity to address unreported bugs.  At a time when we are trying to get a stable codebase for RC - the worst possible time to do this.

I strongly suggest simplifying this, and leaving most of this code as it was.

Sesquipedalian:
> I think that, if you move a forum, you may get a stray cookie, and you may have to log in again, and this is expected behavior. And we can & should rely on the SMF garbage cleanup to clear stray sessions & the browser garbage cleanup to clean up stray cookies.

The trouble is, if the user set their cookie never to expire (i.e. they chose to stay logged in forever), the browser won't delete the now-broken cookie.

The code here is complex because adding support for global cookies is more complex than it seems, and the initial code to do it didn't get everything right.

sbulen:
And what is the problem with that stray cookie?   Worst case scenario is that it stays until someone clears their cache.  Not a huge problem.   

I think the complexity, readability and risk far outweigh that rare stray cookie that will only happen that once in a very long while when you move a forum.

Sesquipedalian:
You have convinced me, gentlemen, to ditch the cookie state code and instead to store both the domain string and the path string. 

We still need to recognize and process the old cookie state values for the 2.0 format cookies, so I have moved the cookie state logic into the self-contained `if` block so that we only use it when we find one of those.

albertlast:
the only "posible" improvment about expire i could see,
that you update the setLoginCookie doc behavior.

PR #4489, opened 2018-01-14
Added language support for recaptcha fixes #4331

### LINKED ISSUE #4331: The google recapcha is not shown in the users language ..
as the title say ...
the problem: the recapcha api is load in the template .. not on the head.
but that is a MUST .. then the api?hl=langcode will work correctly.
So you have to check in the Load.php (where all js is loaded), if the api need for the current request and if so, then load here.
If you do that, then the ?hl=lang works correctly !!
And .. the api langcode, you can put in the index.language.php ..

Also this need a little bit more code, so if this works on ALL situations correct.
As exampe .. 
if a guest do a search like "test", make the recapcha auth and show the first page.
then he make a new search with the same value ("test") no recapcha is shown !!
So bad guest can stress the system if he do this many times !!! (not good)

Also on Preview a Post / PM it's a good option to get the user a "HINT" that he have to complete the reCapcha before he **SEND** the message. At this stage it is no ERROR if the capcha not completed !!

Thanks for reading this ;-)

albertlast:
Since the language is commonly the dictory lang i place a fallback when recaptcha_lang didn't exists:
https://developers.google.com/recaptcha/docs/language
issue: #4331

### DIFF
--- a/Themes/default/GenericControls.template.php
+++ b/Themes/default/GenericControls.template.php
@@ -295,11 +295,13 @@ function template_control_verification($verify_id, $display_type = 'all', $reset
 			}
 
 			if ($verify_context['can_recaptcha'])
+			{
 				$lang = (isset($txt['lang_recaptcha']) ? $txt['lang_recaptcha'] : $txt['lang_dictionary']); 
 				echo '
 				<div class="g-recaptcha centertext" data-sitekey="' . $verify_context['recaptcha_site_key'] . '" data-theme="' . $verify_context['recaptcha_theme'] . '"></div>
 				<br>
 				<script type="text/javascript" src="https://www.google.com/recaptcha/api.js?hl='.$lang.'"></script>';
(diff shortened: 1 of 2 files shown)


XinYenFon (approved):

frandominguezl (approved):

Themes/default/GenericControls.template.php
```diff
@@ -295,10 +295,11 @@ function template_control_verification($verify_id, $display_type = 'all', $reset
 			}
 
 			if ($verify_context['can_recaptcha'])
+				$lang = (isset($txt['lang_recaptcha']) ? $txt['lang_recaptcha'] : $txt['lang_dictionary']);
```
jdarwood007: Got to wrap {} around this if you are going to define another variable.

Sesquipedalian (approved):

PR #4490, opened 2018-01-14
Update Subs-Post.php

Underdog-01:
- create empty arrays prior to attempting to use them (using a one liner)
- parse_bbc(false) always returns an empty string therefore do not attempt to use it as an array in a for loop

If this (the 2nd edit) is supposed to be an attempt to filter $code['tag'] through the parse_bbc function then it should be done within the loop itself. If this is the intention then either change the code to do that or put a notation in this PR and I will make it as such.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,13 +1,5 @@
-# Autodetect text files
-* text=auto
-
-# ...Unless the name matches the following
-# overriding patterns
-
-# Definitively text files 
-*.txt text
-*.c text
-*.h text
+# Force our line endings to be LF, even for Windows
+* text
 
 # Set certain files to be binary
 *.png binary
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,13 @@
-# Force our line endings to be LF, even for Windows
-* text
+# Autodetect text files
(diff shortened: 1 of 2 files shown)


Sesquipedalian:
Actually, `parse_bbc(false)` returns the complete array of BBCode data (see [here](https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs.php#L1761)). With this proposed change, `$codes` would be empty.

Underdog-01:
Yes you're right as I misread the code.. parse_bbc(false) returns the array of bbcodes.
However I am experiencing errors when someone other than an admin sends a PM.
This is using a PHP 7.2 environment.

Perhaps it is an ambiguity problem. Hopefully I'm now reading it correctly and am going to test the code shown below to see if it resolves the errors.
`$codes = parse_bbc(false);

foreach ($codes as $code)`

Sesquipedalian:
If it works for an admin but not for other users, changing this won't make any difference.

Underdog-01:
Yes.. I traced the issue to my own mod.
I apologize for the bother.

PR #4491, opened 2018-01-15
Removed Force SSL Auth and Reg fixes #2939

### LINKED ISSUE #2939: SSL setting needs work - SSL for Auth doesn't return to http
On my 2.1 Beta2 install, I have set "Enable SSL for Authentication (Login and Register)".  

After login, the forums stay in SSL mode, it doesn't return back to plain http.

The forums work appropriately the same otherwise in either http or https.

albertlast:
Removed the force_ssl_auth,
force_ssl_complet got the number 1 now.
To be lazy i changed all "if" to greater zero checks.

Would be nice if someone is able to test this pr,
i got no ssl env.

Issue #2939

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -1754,7 +1754,7 @@ function loadTheme($id_theme = 0, $initialize = true)
 		$member = empty($user_info['id']) ? -1 : $user_info['id'];
 
 		// Disable image proxy if we don't have SSL enabled
-		if (empty($modSettings['force_ssl']) || $modSettings['force_ssl'] < 1)
+		if (empty($modSettings['force_ssl']))
 			$image_proxy_enabled = false;
 
 		if (!empty($modSettings['cache_enable']) && $modSettings['cache_enable'] >= 2 && ($temp = cache_get_data('theme_settings-' . $id_theme . ':' . $member, 60)) != null && time() - 60 > $modSettings['settings_updated'])
@@ -1838,7 +1838,7 @@ function loadTheme($id_theme = 0, $initialize = true)
 		return;
 
 	// Check to see if we're forcing SSL
-	if (!empty($modSettings['force_ssl']) && $modSettings['force_ssl'] > 0 && empty($maintenance) &&
+	if (!empty($modSettings['force_ssl']) && empty($maintenance) &&
 		!httpsOn() && SMF != 'SSI')
 	{
 		if (isset($_GET['sslRedirect']))
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -2003,7 +2003,7 @@ function loadTheme($id_theme = 0, $initialize = true)
 	if (!isset($context['javascript_vars']))
 		$context['javascript_vars'] = array();
 
-	$context['login_url'] = (!empty($modSettings['force_ssl']) && $modSettings['force_ssl'] < 1 ? strtr($scripturl, array('http://' => 'https://')) : $scripturl) . '?action=login2';
+	$context['login_url'] =  $scripturl . '?action=login2';
 	$context['menu_separator'] = !empty($settings['use_image_buttons']) ? ' ' : ' | ';
 	$context['session_var'] = $_SESSION['session_var'];
 	$context['session_id'] = $_SESSION['session_value'];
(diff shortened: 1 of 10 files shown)


Sources/Load.php
```diff
@@ -1754,7 +1754,7 @@ function loadTheme($id_theme = 0, $initialize = true)
 		$member = empty($user_info['id']) ? -1 : $user_info['id'];
 
 		// Disable image proxy if we don't have SSL enabled
-		if (empty($modSettings['force_ssl']) || $modSettings['force_ssl'] < 2)
+		if (empty($modSettings['force_ssl']) || $modSettings['force_ssl'] < 1)
```
Sesquipedalian: This could just be `if (empty($modSettings['force_ssl']))`

Sources/Load.php
```diff
@@ -1838,7 +1838,7 @@ function loadTheme($id_theme = 0, $initialize = true)
 		return;
 
 	// Check to see if we're forcing SSL
-	if (!empty($modSettings['force_ssl']) && $modSettings['force_ssl'] == 2 && empty($maintenance) &&
+	if (!empty($modSettings['force_ssl']) && $modSettings['force_ssl'] > 0 && empty($maintenance) &&
```
Sesquipedalian: This could just be `if (!empty($modSettings['force_ssl']) && empty($maintenance) &&`

Sources/ManageServer.php
```diff
@@ -255,7 +255,7 @@ function AlignURLsWithSSLSetting($new_force_ssl = 0)
 	require_once($sourcedir . '/Subs-Admin.php');
 
 	// Check $boardurl
-	if ($new_force_ssl == 2)
+	if ($new_force_ssl > 0)
```
Sesquipedalian: This could just be `if (!empty($new_force_ssl))`

Sesquipedalian (changes requested):
Checking for `empty()` or `!empty()` is all you need to do if `$modSettings['force_ssl']` is now a binary value. Please change that., then I'll test functionality.

Sources/ManageServer.php
albertlast: Is empty in any kind better as this solution?

albertlast: Run small test ~ 1mio calls the different is small,
when you do 10 mio calls "> 0" is faster then empty().
empty: 3.9425780773163
"> 0": 3.2624399662018

~ 20% faster (but empty() and > 0 are fast already)

Sesquipedalian: We need to use `empty()` because the variable might not be set at all. Once we've done `empty()`, there is no reason to also test whether it is 0 or greater than 0.

albertlast: You're sure that can happen? https://github.com/albertlast/SMF2.1/blob/a9bd146a16bcdcd866500aa82f6905148df843ef/Sources/ManageServer.php#L252
But like i said the diff between the solution is margin so i don't care which way is choosen.

Sesquipedalian:
Sorry if I wasn't clear, but the three spots I commented on were just examples. Please make similar changes in the rest of the PR as well.

Sesquipedalian (approved):
Everything seems to work nicely.

Sesquipedalian:
Before merging, I would like to know @colinschoen's thoughts on whether this is a change we want to make. 

Pro: getting rid of the "Enable SSL for Authentication" mode would solve some outstanding bugs and make things simpler.

Con: "Enable SSL for Authentication" is a useful feature that many users would appreciate. It gives the benefits of SSL when needed while avoiding the headaches and requirement for the image proxy, etc.—or at least it would, if it worked as intended.

For my own part, I'm inclined to drop it. I wish there were time to fix it properly, but I want to get RC1 out the door at long last. That also seems to be the general consensus in the discussion in #2939, but I don't see any comments from you there, Colin, and I want to be sure that we are all on the same page before going ahead.

colinschoen (approved):
Thanks Jon. It is a useful feature and something we should still keep tracking, but I agree it shouldn't block the RC. Let's keep the issue open, but merge this for now. Thanks for your work on this @albertlast

Issue #1178, opened 2014-01-12
Calendar Events Missing Additonal Options


ghost:
## **Old Bug Report**

Description below taken from original bug report.

This bug applys to non-admin users with "additional options" permssions such as announce topics or upload attachments. I am running 1.1.10 and there are several boards configured with different permssions.

Creating a event/topic by following the link from the calendar page results in the user only having the additional options "return to this topic" and "Don't use smileys".

What I think should happen is whatever permissions the user has on the board given in the option "Default board to post events in: " should be applied.

This is the link from the calendar with missing permissions:

```
index.php?action=calendar;sa=post;month=12;year=2009
```

If I change the link manually it has correct permissions:

```
index.php?action=calendar;sa=post;month=12;year=2009;board=7
```

This is the link from "link to calendar" button with correct permissions:

```
index.php?action=post;calendar;msg=87;topic=57.0
```

Option discussed to resolve was, to use a Javascript to determine the selected board and show/hide the attachments part depending if the user is allowed to post attachments in the selected board and ignore the attachments if the user doesn't have the permission to post attachments in the selected board.

Sesquipedalian:
> Option discussed to resolve was, to use a Javascript to determine the selected board and show/hide the attachments part depending if the user is allowed to post attachments in the selected board and ignore the attachments if the user doesn't have the permission to post attachments in the selected board.

The basic idea is good, but it might be trickier to implement than expected. Turning off and hiding checkboxes is easy enough, so handling the basic post options is not a problem. But dealing with the attachments will be much more complicated.

To illustrate, consider this scenario:

1. The user has permission to create events and add attachments in board 1
2. The user has permission to create events, but not to add attachments, in board 2
3. The user selects board 1 as the destination board in which to post her event. Since she can add attachments in board 1, SMF allows her to add some attachments to the event's post.
4. The user decides to change the destination board for the event to board 2...

Now the question is, what should we do with the attachments the user already uploaded? Should we delete her attachments immediately? Should we hide the attachments interface from her but keep the attachments themselves in case she changes her mind? If we do the latter, how and when do we make sure that her attachments do not get posted on board 2?

Since the attachments system is pretty complex, I'd like to hear what @MissAllSunday thinks about this.

Sesquipedalian:
At this point, here's the situation with this:

Suppose that Joe is a non-admin user who has permission to create events and who has permission to post attachments in Board 1 but not in Board 2.

[...]

PR #4495, opened 2018-01-17
Fixes #4493

### LINKED ISSUE #4493: Incorrect posting options when posting via ?action=calendar;sa=post
For a description, see the "If the admin option to allow events that are not attached to posts is disabled" section of [this comment](https://github.com/SimpleMachines/SMF2.1/issues/1178#issuecomment-307890381) as well as the relevant parts of the original bug description in #1178.

Sesquipedalian:
This fixes #4493 (and thus the main issue in #1178) as well as taking care of some to-do items in `Post()`.

The overall issue is that the logic in `Post()` was built on the assumption that `$board` would be set, and then makes exceptions to that assumption when creating an event. Those exceptions had a "bolted on" feel to them, inasmuch as they worked well enough to allow the core task of creating a post to complete successfully, but they didn't cover every aspect. To fix this properly has required (1) adjusting the logic in `Post()` so that it expects from the outset that `$board` might be empty and, as a prerequisite, (2) adding any/all logic to `allowedTo()` and `isAllowedTo()`.

The latter is a very neat and clean change. `allowedTo()` and `isAllowedTo()` have been given an optional third parameter, `$any`, with a default `false` value. Normally, `allowedTo('post_new', $boards)` will return true if the current user can make new posts in **all** of the boards listed in `$boards`. If the `$any` parameter is set to true, as in `allowedTo('post_new', $boards, true)`, the function will return true if the current user can make new posts in **at least one** of the boards listed in `$boards`.

The former is a little more involved:
1. Some code that was already being used in the special `if ($context['make_event'])` block up has been moved near the top of the function. As a result, if `$board` is empty, we build a list of boards the user could post in. 
2. Calls to `allowedTo()` and `isAllowedTo()` in several relevant places now take advantage of the change to those function described above in order to check for permissions on any of those boards when deciding what options to show the user.
3. If necessary, we now add a select menu to the form forcing the user to select a board to post in. This ensures that `$board` will be set for `Post2()`.
4. Some variable assignment statements have been organized to make them neater.
5. We do a final `allowedTo('announce_topic')` check before sending an announcement. (That really ought to have been there already.)
6. One or two minor fixes, tidying, etc.

It is important to note that in order to fix #4493, we have to show all the options to the user that they have on *any* board they might ultimately submit the post to. There's no way to know in advance what board they are going to post in if they are coming from `?action=calendar;sa=post`, so we can only give them the options and then send them back to the posting screen if they choose an option that is not permitted to them on the board they select. Fortunately, `Post2()` does check all these things (well, it did check all but `announce_topic`, and now it checks that, too) before acting, so if the user selected an option on the form that they didn't have permission for, `Post2()` will kick them back to the form with error messages. Notably, at this point `$board` will be set and so the user will not have the opportunity to select a different board.

### DIFF
--- a/Sources/Post.php
+++ b/Sources/Post.php
@@ -1514,7 +1514,7 @@ function Post2()
 		if (isset($_POST['lock']))
 		{
 			// Nothing is changed to the lock.
-			if ((empty($topic_info['locked']) && empty($_POST['lock'])) || (!empty($_POST['lock']) && !empty($topic_info['locked'])))
+			if (empty($topic_info['locked']) == empty($_POST['lock']))
 				unset($_POST['lock']);
 
 			// You're have no permission to lock this topic.
--- a/Sources/Post.php
+++ b/Sources/Post.php
@@ -53,9 +53,31 @@ function Post($post_errors = array())
 	$context['notify_prefs'] = (array) array_pop($temp);
 	$context['auto_notify'] = !empty($context['notify_prefs']['msg_auto_notify']);
 
-	// You must be posting to *some* board.
-	if (empty($board) && !$context['make_event'])
-		fatal_lang_error('no_board', false);
+	// Not in a board? Fine, but we'll make them pick one eventually.
+	if (empty($board) || $context['make_event'])
+	{
+		// Get ids of all the boards they can post in.
+		$post_permissions = array('post_new');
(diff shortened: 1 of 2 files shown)


Sesquipedalian:
Apparently Travis was having some kind of an outage when I submitted this.

PR #4502, opened 2018-01-21
Fixes #1066. Makes $smcFunc['htmltrim'] faster and more effective.

### LINKED ISSUE #1066: blank message name?
http://www.simplemachines.org/community/index.php?action=profile;area=showposts;u=83023

this user has somehow created 2 messages without text in the subject line. this means that the subject link does not show up in the message. should we not allow blank text like this in the subject?

the source shows

`<a href="http://www.simplemachines.org/community/index.php?topic=201161.msg3413450#msg3413450"> </a>
<a href="http://www.simplemachines.org/community/index.php?topic=455983.msg3648216#msg3648216"> </a>`

but when viewing the post you cant see the link at all

Sesquipedalian:
For those wondering how much more effective this is, here's the result of benchmarking using a string of 50 different Unicode whitespace and control characters:

```
Length of initial string: 50

Entities for initial string: &#0;&#9;&#10;&#11;&#13;&#160;&#173;&#8192;&#8193;&#8194;&#8195;&#8196;&#8197;&#8198;&#8199;&#8200;&#8201;&#8202;&#8203;&#8204;&#8205;&#8206;&#8207;&#8232;&#8233;&#8234;&#8235;&#8236;&#8237;&#8238;&#8239;&#8287;&#8288;&#8289;&#8290;&#8291;&#8292;&#8293;&#8294;&#8295;&#8296;&#8297;&#8298;&#8299;&#8300;&#8301;&#8302;&#8303;&#12288;&#65279;

Time for 1000000 iterations of old preg_replace: 7.2233781814575
Final length (should be 0): 25

Time for 1000000 iterations of new preg_replace: 4.361584186554
Final length (should be 0): 0
```

As shown, the new method got them all, and did it much faster.

Fixes #1066

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -76,7 +76,7 @@ function reloadSettings()
 	$utf8 = (empty($modSettings['global_character_set']) ? $txt['lang_character_set'] : $modSettings['global_character_set']) === 'UTF-8';
 
 	// Set a list of common functions.
-	$ent_list = '&(#' . (empty($modSettings['disableEntityCheck']) ? '\d{1,7}' : '021') . '|quot|amp|lt|gt|nbsp);';
+	$ent_list = '&(?:#' . (empty($modSettings['disableEntityCheck']) ? '\d{1,7}' : '021') . '|quot|amp|lt|gt|nbsp);';
 	$ent_check = empty($modSettings['disableEntityCheck']) ? function($string)
 		{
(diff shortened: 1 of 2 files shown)


Sesquipedalian:
I noticed some nearby spots where we were needlessly reduplicating the logic that creates `$ent_list`, so I fixed those, too.

PR #4503, opened 2018-01-22
Dont proxy images for bots, fixes #4448

### LINKED ISSUE #4448: Proxy Enhancement:  Don't serve up bots
After enabling the proxy on my 2.0 site, I noticed CPU up 2x & bandwidth up 3x.  I believe the crux of the problem is serving up images for search bots.  

I believe we should NOT serve up images thru the proxy for suspected bots.  

We probably shouldn't be serving up these images (originating from other sites) for search engines anyway.

This impacts 2.0 & 2.1.

sbulen:
Don't proxy images for bots.  

This fixes issues where site recrawls, even by good bots like Google, can cause your cache to grow by many GB of data & get completely randomized - useless as a cache.   

Also, bots probably should "see" the image's proper link anyway.  Give the image's original source page proper credit.

At this point in time, Google does not penalize sites for mixed-content pages (http & https), so this will not hurt Google's ranking of your forum's content.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -1415,13 +1415,13 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 				{
 					global $image_proxy_enabled, $image_proxy_secret, $boardurl, $user_info;
 
-					if (!empty($user_info['possibly_robot']))
-						return;
-
 					$data = strtr($data, array('<br>' => ''));
 					$scheme = parse_url($data, PHP_URL_SCHEME);
 					if ($image_proxy_enabled)
 					{
+						if (!empty($user_info['possibly_robot']))
+							return;
+
 						if (empty($scheme))
 							$data = 'http://' . ltrim($data, ':/');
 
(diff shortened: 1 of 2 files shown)


Sources/Subs.php
```diff
@@ -1413,7 +1413,10 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 				'content' => '<img src="$1" alt="{alt}" title="{title}"{width}{height} class="bbc_img resized">',
 				'validate' => function (&$tag, &$data, $disabled)
 				{
-					global $image_proxy_enabled, $image_proxy_secret, $boardurl;
+					global $image_proxy_enabled, $image_proxy_secret, $boardurl, $user_info;
+
+					if (!empty($user_info['possibly_robot']))
+						return;
```
Sesquipedalian: This should go inside the `if ($image_proxy_enabled)` block.

Sources/Subs.php
```diff
@@ -1436,7 +1439,10 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 				'content' => '<img src="$1" alt="" class="bbc_img">',
 				'validate' => function (&$tag, &$data, $disabled)
 				{
-					global $image_proxy_enabled, $image_proxy_secret, $boardurl;
+					global $image_proxy_enabled, $image_proxy_secret, $boardurl, $user_info;
+
+					if (!empty($user_info['possibly_robot']))
+						return;
```
Sesquipedalian: This should go inside the `if ($image_proxy_enabled)` block.

Sesquipedalian (changes requested):

sbulen:
I'll make the change & retest.

sbulen:
Done.  Tests OK.
Tested on a wamp server, php5.6, MySQL 5.7.

Sesquipedalian (approved):

PR #4505, opened 2018-01-23
Fixes #1189

### LINKED ISSUE #1189: Word Censor Should Disable Option to be Limited to Whole Words in UTF-8
## *\* Old Bug Report **

Still present in 2.1 alpha.

Original report:
The word censor provides the option for restricting censor matches to whole words, or simply match in place.

When matching whole words, this relies on using the \w marker (or is it \W, I forget, either way it's a PCRE control character), which is fine -- until you're in UTF-8 mode.

As covered by http://www.simplemachines.org/community/index.php?topic=363219.msg2612779#msg2612779 and specifically I'll requote the paragraph from the PHP manual on the subject:
Quote

`Matching characters by Unicode property is not fast, because PCRE has to search a structure that contains data for over fifteen thousand characters. That is why the traditional escape sequences such as \d and \w do not use Unicode properties in PCRE.`

Since PCRE does not support \w in UTF-8 mode, the option is actually pointless (since it doesn't work) so the option should be removed when in UTF-8 mode.

Sesquipedalian:
The bug fix consists in using `[\p{L}\p{M}\p{N}_]` instead of `\w` when using UTF-8. 

Although the PHP manual warns that "Matching characters by Unicode property is not fast," testing shows that using `[\p{L}\p{M}\p{N}_]` in the word censoring regex takes only 9% longer `\w` in PHP 5.6, and only 7% longer in PHP 7.1. Maybe in 2010, when the original bug report was filed, the performance difference was large enough to warrant the suggestion in #1189 to drop the whole word only option for Unicode, but in 2017 there's no reason to do so.

Besides fixing the bug, I've also optimized the regexes further in a couple of ways:
1. When creating the final regex for each censored term, I use the faster `\b` word boundary metacharacter whenever possible.
2. Instead of `(?<=^|\W)` and `(?=$|\W)` (or `(?<=^|[^\p{L}\p{M}\p{N}_])` and `(?=$|[^\p{L}\p{M}\p{N}_])` for Unicode), I'm using `(?<!\w)` and `(?!\w)` (or `(?<![\p{L}\p{M}\p{N}_])` and `(?![\p{L}\p{M}\p{N}_])`). This is functionally equivalent, but it is more concise and it is slightly faster since it doesn't include any alternation.

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -2984,14 +2984,17 @@ function censorText(&$text, $force = false)
 		// Quote them for use in regular expressions.
 		if (!empty($modSettings['censorWholeWord']))
 		{
+			$charset = empty($modSettings['global_character_set']) ? $txt['lang_character_set'] : $modSettings['global_character_set'];
+
 			for ($i = 0, $n = count($censor_vulgar); $i < $n; $i++)
 			{
 				$censor_vulgar[$i] = str_replace(array('\\\\\\*', '\\*', '&', '\''), array('[*]', '[^\s]*?', '&amp;', '&#039;'), preg_quote($censor_vulgar[$i], '/'));
-				$censor_vulgar[$i] = '/(?<=^|\W)' . $censor_vulgar[$i] . '(?=$|\W)/' . (empty($modSettings['censorIgnoreCase']) ? '' : 'i') . ((empty($modSettings['global_character_set']) ? $txt['lang_character_set'] : $modSettings['global_character_set']) === 'UTF-8' ? 'u' : '');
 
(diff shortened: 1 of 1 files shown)


colinschoen:
@albertlast This seems like your forte. Want to have a look?

albertlast:
regex is a curse,
this is all what i know about this.

colinschoen:
Haha fair enough :)

Sesquipedalian:
How about I just say that I'm very, very sure that this is good? ;)

colinschoen (approved):
LGTM i'll take your word on it :)

PR #4509, opened 2018-01-23
Adds autoloader to SSI.php. Fixes #3959

### LINKED ISSUE #3959: Guests recieve error on pages using SSI when guest access is disabled
I have a website using SSI, /forum/SSI.php is included in my /index.php with `$ssi_guest_access = false` (default). If someone visits /index.php as a guest they receive the following error:

`
Fatal error: Uncaught Error: Class 'MatthiasMullie\Minify\CSS' not found in /var/www/site/public/forum/Sources/Subs.php:3617 Stack trace: #0 /var/www/site/public/forum/Sources/Subs.php(3529): custMinify(Array, 'css') #1 /var/www/site/public/forum/Themes/default/index.template.php(99): template_css() #2 /var/www/site/public/forum/Sources/Load.php(2355): template_html_above() #3 /var/www/site/public/forum/Sources/Subs.php(3263): loadSubTemplate('html_above', true) #4 /var/www/site/public/forum/Sources/Subs.php(2858): template_header() #5 /var/www/site/public/forum/SSI.php(157): obExit(NULL, true) #6 /var/www/site/public/index.php(14): require_once('/var/www/site/p...') #7 {main} thrown in /var/www/site/public/forum/Sources/Subs.php on line 3617`

However, visitors that are logged into the forum do not receive the error and SSI works as intended. 

If I disable Minify CSS/JS in the forum settings then the error doesn't appear for guests, and they are redirected to the forum login as intended.

If `$ssi_guest_access = true` then the error doesn't appear.

Sesquipedalian:


Sesquipedalian:
Better idea implemented in c299c0e: move the `spl_autoload_register()` call from index.php into Load.php. Then it is available to both index.php and SSI.php from one place.

jdarwood007:
Source/template files should not do any execution by themselves.  I.E, you shouldn't be able to call Load.php and have it do anything (forgetting for now the kill line at the top).

Sesquipedalian:
I would argue that this should be an exception, but fine.

jdarwood007:
Well that would be up to @colinschoen if we want to allow this.  This is just what I know from what past developers would say.  I'm a old dev, I know the old rules, they may not be relevant anymore.

Sesquipedalian:
Meh. I don't care enough to argue. :)

PR #4510, opened 2018-01-23
Query debug

### LINKED ISSUE #4508: Undefined index: s on line 511 of Subs-Db-mysql.php
I've started getting this error sometimes since merging #4472.

The cause of the issue is that when `!empty($_SESSION['debug_redirect'])` is true, `$db_count` is incremented twice, once [here](https://github.com/SimpleMachines/SMF2.1/pull/4472/commits/eef44bda06f04efae95036d8892435141d63cdf8#diff-40a16e287462d60f366698be5e160838R358) and again [here](https://github.com/SimpleMachines/SMF2.1/pull/4472/commits/eef44bda06f04efae95036d8892435141d63cdf8#diff-40a16e287462d60f366698be5e160838R449).

Looking at #4472 again, I think we can accomplish @albertlast's goal in that PR more easily by simply moving the old debugging block below the block where we clean the query string.

Sesquipedalian:
Fixes #4508


Sources/Logging.php
```diff
@@ -249,10 +249,10 @@ function displayDebug()
 	$warnings = 0;
 	if (!empty($db_cache))
 	{
-		foreach ($db_cache as $q => $query_data)
+		foreach ($db_cache as $q => $qq)
```
albertlast: On my pr you mention that qq is bad and
now it's good?
https://github.com/SimpleMachines/SMF2.1/pull/4472#discussion_r162849616

Sources/Subs-Db-postgresql.php
```diff
@@ -470,6 +441,29 @@ function smf_db_query($identifier, $db_string, $db_values = array(), $connection
 			smf_db_error_backtrace('Hacking attempt...', 'Hacking attempt...' . "\n" . $db_string, E_USER_ERROR, __FILE__, __LINE__);
 	}
 
+	// Debugging.
```
albertlast: I would like to see this code below "// Set optimize stuff" block,
idea would be that the code should be than the same as mysql part.

Sesquipedalian: The optimization block expects `$db_cache[$db_count]['q']` to already have a value.

Sesquipedalian: ... But I suppose there is no strong reason why it would need to be left that way.

Sources/Logging.php
Sesquipedalian: No, I just missed this when reimplementing those changes. :)

Sources/Subs-Db-postgresql.php
```diff
@@ -491,10 +485,7 @@ function smf_db_query($identifier, $db_string, $db_values = array(), $connection
 	}
 
 	if (isset($db_show_debug) && $db_show_debug === true)
```
albertlast: I'm not sure if i clear,
so i would expect that the code is placed here

albertlast (approved):

PR #4511, opened 2018-01-24
Fixes bugs with two factor authentication

### LINKED ISSUE #3592: tfa bypass
enable tfa on an account. open a new browser [not the one you used to setup tfa] and log in with tfa with checking the box to stay logged in on the tfa page. logout, right after you logout click the login link, fill in your username and password and log in, you get an error message to wait 2 minutes to log in. click the back button... it takes you back to the forum index and you are logged in. IMO if you have not entered your tfa it should not log you or it should give you a different error message.

if you try another browser and attempt to log in, you should get prompted to enter your tfa but you dont. you get to the error page telling you to wait 2 minutes. when you try to click the back button you dont get redirected to the index page when you click back, you just go in a loop on the error page that you need to wait 2 minutes. even clicking on the home button ends me back on the same error page, so i cant even browse the forum as a guest. the URL gets changed to http://localhost:8080/github2.1/index.php?action=logintfa and even manually removing the action does not get me back to the normal board index.

tried the last part again, removed the action and i got prompted to enter the tfa... maybe this needs to be looked into further.

Sesquipedalian:
Fixes #3592
- Moving the `redirectexit('action=logintfa');` out of `loadUserSettings()` into `Login2()` solved the endless loop problem.
- Getting rid of the `$preserve` stuff (which wasn't accomplishing anything useful anyway) helped to solve the problem where clicking the back button after receiving the "You must wait two minutes" warning would log the user in.

Also cleans up the TFA login UI a bit.


Sesquipedalian:
Give this a try, @illori.

illori:
if you are in a rush, you can ask others. i might not have time for a good few days.

Sesquipedalian:
No worries.

Sources/Load.php
```diff
@@ -656,9 +658,9 @@ function loadUserSettings()
 		{
 			$tfa_data = $smcFunc['json_decode']($_COOKIE[$cookiename . '_tfa'], true);
 
-			list ($id, $user, $exp, $domain, $path, $preserve) = $tfa_data;
+			list ($id, $user, $exp, $domain, $path) = array_pad((array) $tfa_data, 5, '');
```
equinoxmatt: Since the only variable that is actually used is $exp, can we not refactor this?  Clean up various parts of the codebase are touched.

Sesquipedalian: Yes, I suppose we can. Good catch, @equinoxmatt.

Sesquipedalian:
I've tested this out a fair bit myself, and it all seems to work. Lacking any testing or reviews from others, I'm going to go ahead and merge it.

PR #4512, opened 2018-01-26
Implement 2.0.15 User Search Fixes

### LINKED ISSUE #4500: Need to implement "Some User search Bugs" fix from 2.0.15 patch
It looks like some of the code from this bug fix has been implemented in 2.1, but not all of it.

@colinschoen, since (IIRC) you came up with that fix in the 2.0.15 patch, it might be best if you take point on this.

colinschoen:
Pull in fixes from 2.0.15.

Fixes #4500

### DIFF
--- a/Sources/Subs-Auth.php
+++ b/Sources/Subs-Auth.php
@@ -394,6 +394,7 @@ function findMembers($names, $use_wildcards = false, $buddies_only = false, $max
 		$names = explode(',', $names);
 
 	$maybe_email = false;
+	$names_list = array();
 	foreach ($names as $i => $name)
 	{
 		// Trim, and fix wildcards for each name.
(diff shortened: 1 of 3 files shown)


jdarwood007:
considering this is a direct copy of the 2.0.15 fixes, I think its safe to merge.
LGTM

Sesquipedalian (approved):

colinschoen:
One small change, but mostly the same. I tested it too. Thanks for reviewing.

Sources/Subs-Auth.php
```diff
@@ -425,22 +429,23 @@ function findMembers($names, $use_wildcards = false, $buddies_only = false, $max
...
+	$real_name_search = $real_name . ' ' . $comparison . ' ' . implode( ' OR ' . $real_name . ' ' . $comparison . ' ', $names_list);
+
 	// Search by username, display name, and email address.
 	$request = $smcFunc['db_query']('', '
 		SELECT id_member, member_name, real_name, email_address
 		FROM {db_prefix}members
-		WHERE ({raw:member_name_search}
-			OR {raw:real_name_search} {raw:email_condition})
+		WHERE (' . $member_name_search . '
+			OR ' . $real_name_search . ' ' . $email_condition . ')
```
albertlast, line 441: This doesn't look like a improvement, 
Why you removed the raw?

PR #4513, opened 2018-01-26
Adds a function to the upgrader to tell Settings.php to store db_last_error.php in the cache

Sesquipedalian:
Complements #4305


jdarwood007:
LGTM

Although seeing that old fix for blanking the Settings file, I wonder if its still relevant.  I don't even know when that showed up in SMF, but its been around a very long time.

Sesquipedalian:
Yeah, it seems questionable to me, too. But it doesn't hurt anything either.

PR #4514, opened 2018-01-27
Reg: Save website_* when error and drop captcha value fixes #4456

### LINKED ISSUE #4456: Profile Fields Values not Persisted on Registration w/ Errors
When profile fields that `$_POST` to different name/s than the `registration_fields` key their value is not persisted through registration form submissions with errors. 

An example is the `Website` profile field which when enabled to `Show on Registration` is actually two form inputs: `Website title`, `Website URL`. These inputs $_POST to `website_title` and `website_url` respectively which don't match the key in `registration_fields`. Perhaps we need to store the names of all inputs for each `Profile Field` and check for those [here](https://github.com/SimpleMachines/SMF2.1/blob/ae07c3e0b8b0285414af2364330e6a2253af9b4a/Sources/Register.php#L170)

albertlast:
I'm not sure if i insert the logic in the right way "but it works" and
also i find use less to save the captcha code from the input.

issue: #4456

### DIFF
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -574,7 +574,6 @@ function loadProfileFields($force_reload = false)
 			'size' => 50,
 			'permission' => 'profile_website',
 			'link_with' => 'website',
-			'value' => (!empty($_POST['website_title'])? $smcFunc['htmlspecialchars'] ($_POST['website_title'], ENT_QUOTES) : ''),
 		),
 		'website_url' => array(
 			'type' => 'url',
@@ -592,7 +591,6 @@ function loadProfileFields($force_reload = false)
 				return true;
 			},
 			'link_with' => 'website',
-			'value' => (!empty($_POST['website_url'])? $smcFunc['htmlspecialchars'] ($_POST['website_url'], ENT_QUOTES) : ''),
 		),
 	);
 
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -574,6 +574,7 @@ function loadProfileFields($force_reload = false)
 			'size' => 50,
 			'permission' => 'profile_website',
 			'link_with' => 'website',
+			'value' => (!empty($_POST['website_title'])? $smcFunc['htmlspecialchars'] ($_POST['website_title'], ENT_QUOTES) : ''),
 		),
 		'website_url' => array(
 			'type' => 'url',
@@ -591,6 +592,7 @@ function loadProfileFields($force_reload = false)
 				return true;
 			},
 			'link_with' => 'website',
+			'value' => (!empty($_POST['website_url'])? $smcFunc['htmlspecialchars'] ($_POST['website_url'], ENT_QUOTES) : ''),
 		),
 	);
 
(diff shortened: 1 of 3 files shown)


albertlast:
@Sesquipedalian @colinschoen can we merge this?

Sources/Profile-Modify.php
```diff
@@ -574,6 +574,7 @@ function $preload:		A function that is used to load data required for this eleme
 			'size' => 50,
 			'permission' => 'profile_website',
 			'link_with' => 'website',
+			'value' => (!empty($_POST['website_title'])? $smcFunc['htmlspecialchars'] ($_POST['website_title'], ENT_QUOTES) : ''),
```
Sesquipedalian: Doing this here would set `website_title` to an empty string when viewing the user's profile page, wouldn't it?

Sources/Profile-Modify.php
```diff
@@ -591,6 +592,7 @@ function $preload:		A function that is used to load data required for this eleme
 				return true;
 			},
 			'link_with' => 'website',
+			'value' => (!empty($_POST['website_url'])? $smcFunc['htmlspecialchars'] ($_POST['website_url'], ENT_QUOTES) : ''),
```
Sesquipedalian: Doing this here would set `website_url` to an empty string when viewing the user's profile page, wouldn't it?

Sesquipedalian (changes requested):
I don't think this is the right place to make changes in order to fix #4456.

Sources/Profile-Modify.php
albertlast: I checked and it's realy happen...
Where would be the right place to fix this?

Sesquipedalian: Probably in the main `Register()` function in Register.php, somewhere inside [this if block](https://github.com/SimpleMachines/SMF2.1/blob/ae07c3e0b8b0285414af2364330e6a2253af9b4a/Sources/Register.php#L155-L176).

Sources/Register.php
```diff
@@ -166,6 +166,16 @@ function Register($reg_errors = array())
 		$user_info['permissions'] = array_merge($user_info['permissions'], array('profile_account_own', 'profile_extra_own', 'profile_other_own', 'profile_password_own', 'profile_website_own', 'profile_blurb'));
 		$reg_fields = explode(',', $modSettings['registration_fields']);
 
+		// Website is a little different
+		if (in_array('website', $reg_fields))
+		{
+			unset($reg_fields['website']);
+			if (isset($_POST['website_title']))
+				$cur_profile['website_title'] = $smcFunc['htmlspecialchars']($_POST['website_title']);
+			if (isset($_POST['website_url']))
```
Sesquipedalian, line 175: This looks pretty good. However, we need the user to supply both the title and the url of their website, so it would be better to do a single check for both, like this:

```
if (!empty($_POST['website_title']) && !empty($_POST['website_url'])
{
	$cur_profile['website_title'] = $smcFunc['htmlspecialchars']($_POST['website_title']);
	$cur_profile['website_url'] = $smcFunc['htmlspecialchars']($_POST['website_url']);
}
```

Sesquipedalian (changes requested):

Sources/Register.php
albertlast: Please had in mind what the scope of this pr is.
The scope is that when you reg doing something wrong exp. you write the wrong auth code,
than you return to the reg form and this pr fix the issue that in this case the title and url are empty -> we want also partly save.

Sesquipedalian: Yes, you are right. That makes sense then.

Sesquipedalian (approved):

PR #4515, opened 2018-01-27
Check if $next_c is not empty

albertlast:
could be happen when the tag is the last char
Should fix this: https://www.simplemachines.org/community/index.php?topic=558590.0

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -2259,7 +2259,7 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 			if (strtolower(substr($message, $pos + 1, $pt_strlen)) != $possible['tag'])
 				continue;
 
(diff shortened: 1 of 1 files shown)


Sources/Subs.php
```diff
@@ -2259,7 +2259,11 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 			if (strtolower(substr($message, $pos + 1, $pt_strlen)) != $possible['tag'])
 				continue;
 
-			$next_c = $message[$pos + 1 + $pt_strlen];
+			$next_c = isset($message[$pos + 1 + $pt_strlen])? $message[$pos + 1 + $pt_strlen] : '';
```
colinschoen: nit: Space after `)` and before `?`.

colinschoen (changes requested):

colinschoen (approved):
LGTM approval

PR #4517, opened 2018-01-28
Fix missing $db_persist global, fixes #4516

### LINKED ISSUE #4516: Persistent connection issue (2.0 & 2.1)

```
		if (!$db_connection)
		{
			if (empty($db_persist))
			else
		}
```

sbulen:
Without this change, the 'persistent connection' setting doesn't really work; it just creates a new connection every time.  

With this change, you can see that subsequent connects use next to 0 resources.  Without this change, sessionWrite() always needs a new db connection, increasing overhead.   

**_This change is also needed in 2.0_**
smf21_persist_compare

### DIFF
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -374,7 +374,7 @@ function smf_db_query($identifier, $db_string, $db_values = array(), $connection
 	// Get a connection if we are shutting down, sometimes the link is closed before sessions are written
 	if (!is_object($connection))
 	{
-		global $db_server, $db_user, $db_passwd, $db_name, $db_show_debug, $ssi_db_user, $ssi_db_passwd;
+		global $db_server, $db_user, $db_passwd, $db_name, $db_show_debug, $ssi_db_user, $ssi_db_passwd, $db_persist;
 
 		// Are we in SSI mode?  If so try that username and password first
 		if (SMF == 'SSI' && !empty($ssi_db_user) && !empty($ssi_db_passwd))


colinschoen (approved):
LGTM approval

jdarwood007:
BTW: This affects 2.0 as well.

jdarwood007:
Also this logic is flawed, as its depending on $db_presist from Settings.php.  If a custom database handler is called and this condition occurs, it will use the $db_presist from Settings.php rather than the $db_options.  In fact if a error occurs at all using a custom connection string, it will fail to work on reconnect because all the database information is used from Settings.php.

Seems like we need to make all the database connection info private local var when calling initialize, then reuse it as needed.

colinschoen:
Makes sense. 

Reverted per comments above.

sbulen:
Should I do the $db_options, as @albertlast suggested?   

I'm lost as to why that is safer, even reading jdarwood note above.

This version did solve the performance issue I was setting out to fix.

jdarwood007:
For a default install of SMF with no changes this fix will resolve that problem.

The problem I see results in when you start using SMF APIs/calls outside of SMF.  Such as for example if you use SSI to get SMF going, then use the database setup to initialize a new database connection, to another database.  That would mean a different user, password, database and possibly db_persist among other db_options.  During a event that would trigger this issue in the first place, the process to reinitialize the connection would fail to reconnect you to the proper database since it would try to use the information from the Settings.php, rather than what it was initially supplied with when we called the process to initialize the connection.

jdarwood007:
Also any usage of $ssi_db_(user|passwd) would also have this same resulting issue.

PR #4518, opened 2018-01-28
Revert "Fix missing $db_persist global, fixes #4516"

### LINKED ISSUE #4516: Persistent connection issue (2.0 & 2.1)

colinschoen:
Reverts SimpleMachines/SMF2.1#4517

### DIFF
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -374,7 +374,7 @@ function smf_db_query($identifier, $db_string, $db_values = array(), $connection
 	// Get a connection if we are shutting down, sometimes the link is closed before sessions are written
 	if (!is_object($connection))
 	{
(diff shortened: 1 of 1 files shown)


sbulen:
Something about needing to work for DBs not described in settings.php.    Also not liking the reconnect logic.

Apparently "working" is insufficient these days.

colinschoen:
@Sesquipedalian Discussion here: #4517

Sesquipedalian:
I meant, why was 43ad686 (#4515) reverted? It doesn't have anything to do with the database connection issue.

colinschoen:
Weird, I don't know why that was added to this history, but if you actually look at the files changed it was not reverted. Here is the relevant file and line at `release-2.1` HEAD: 

https://github.com/SimpleMachines/SMF2.1/blob/ac85d8c0fb1b070aaabfe622cf02839682c7eafb/Sources/Subs.php#L2264

Sesquipedalian:
That is odd. But at any rate, that answers my question.

PR #4519, opened 2018-01-28
Use the database directly on connection and use mysqli_query fixes #4516 partly

### LINKED ISSUE #4516: Persistent connection issue (2.0 & 2.1)

albertlast:
instead of smcfunc query
Thanks to @sbulen i notice that mysql logic got a reconnect logic.
I used this new knowledg to check in which case a common smf 2.1 env call this.

Two reason are founded:
- mysql_real_connect needs a schema/database or will go fail
- in the smf_db_initiate is the call $smcFunc['db_query'] stupid because the $connection is not known at this point -> needs to be a mysqli_query

Whith this changes the issue #4516 get partly fixed -> no double connection call any more.

### DIFF
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -73,11 +73,15 @@ function smf_db_initiate($db_server, $db_name, $db_user, $db_passwd, $db_prefix,
 
 	$success = false;
 
+	// Windows doesn't support socket
+	if (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN' && $db_server == 'localhost')
+			$db_server = '127.0.0.1';
+
 	if ($connection) {
 		if (!empty($db_options['port']))
-			$success = mysqli_real_connect($connection, $db_server, $db_user, $db_passwd, $db_name, $db_options['port'], null, $flags);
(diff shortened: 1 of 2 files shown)


sbulen:
I have tested this & confirmed it fixes the double connect problem in 2.1.

albertlast:
@colinschoen @Sesquipedalian than please merge

sbulen:
My only minor concern with this is that the prior logic connected without the db_name, but followed up with a db_name selection ~line 93.   I don't know why it did it in two steps like that, but it did.  

So, if we are going to connect directly to the db_name, then the select isn't needed ~line 93.  

If we are going to continue to honor the 'dont_select_db' setting, then we should leave it in 2 steps and I suspect we should not specify dbname at connect time.

albertlast:
Waiting for conformation that this code is running in unix/linux env: https://pastebin.com/5WVizEvz
But i already changed the code to support socket connection.

To use socket in mysql env, hostname needs to be localhost
in pg env, hostname needs to be empty.

Also the last commit solved the concern from @sbulen

XinYenFon:
@Arantor , just curious what you think about this.

Arantor:
Please don't drag me into this.

PR #4527, opened 2018-02-04
Fix own Theme Select/change fixes #4525

### LINKED ISSUE #4525: Board themes not working properly
Per report here: https://www.simplemachines.org/community/index.php?topic=558121.0

> SMF offers the option to set a theme by board, and not force it on anyone who has a specific theme selected.
For this to work, the user needs to have the "Forum or Board Default" theme option selected in their profile.
For an admin user, this does not work, and so an admin is unable to see the board specific themes without forcing them for everyone.

albertlast:
Issue #4525

### DIFF
--- a/Sources/Themes.php
+++ b/Sources/Themes.php
@@ -925,7 +925,7 @@ function PickTheme()
 		$_GET['th'] = (int) $_GET['th'];
 
 		// Save for this user.
-		if (!isset($_REQUEST['u']) || !allowedTo('admin_forum'))
+		if (!isset($_REQUEST['u']) || !allowedTo('admin_forum') || (!empty($_REQUEST['u']) && $_REQUEST['u'] == $user_info['id']))
 		{
 			updateMemberData($user_info['id'], array('id_theme' => (int) $_GET['th']));
 
--- a/Sources/Themes.php
+++ b/Sources/Themes.php
@@ -993,7 +993,7 @@ function PickTheme()
 		{
 			updateMemberData((int) $_REQUEST['u'], array('id_theme' => (int) $_GET['th']));
 
-			if (!empty($_GET['vrt']))
+			if (!empty($_GET['vrt']) && $_GET['th'] != 0)
 			{
 				$smcFunc['db_insert']('replace',
 					'{db_prefix}themes',
@@ -1005,6 +1005,20 @@ function PickTheme()
 
 				if ($user_info['id'] == $_REQUEST['u'])
 					$_SESSION['id_variant'] = 0;
+			}
+			elseif ($_GET['th'] == 0)
+			{
+				// Remove any custom variants.
+				$smcFunc['db_query']('', '
+					DELETE FROM {db_prefix}themes
+					WHERE 
+						variable = {string:theme_variant}
+						AND id_member = {int:id_member}',
+					array(
+						'theme_variant' => 'theme_variant',
+						'id_member' => (int) $_REQUEST['u'],
+					)
+				);
 			}
 
 			redirectexit('action=profile;u=' . (int) $_REQUEST['u'] . ';area=theme');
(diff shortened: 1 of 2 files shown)


Gwenwyfar (approved):

Sources/Themes.php
```diff
@@ -991,10 +991,6 @@ function PickTheme()
 		// Change a specific member's theme.
 		else
 		{
-			// The forum's default theme is always 0 and we
-			if (isset($_GET['th']) && $_GET['th'] == 0)
-					$_GET['th'] = $modSettings['theme_guests'];
-
```
live627, line 997: this breaks theme variants

albertlast: Can you describe in more detail, 
What are variant and how it break.

Sesquipedalian: Theme variants allow a single theme to have multiple colour schemes. See [here](https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Themes/default/index.template.php#L95) in index.template.php, [here](https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Load.php#L2150) in Load.php, and [here](https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Themes.php#L686), [here](https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Themes.php#L783), and [here](https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Themes.php#L1145) in Themes.php

albertlast: When i set a theme as default which got a variant than the stuff get broken without this logic?

albertlast: I created some variants for the curve2 template and didn't see any issue,
by side that i'm not able to define a variant as default.
So from my side is ther no issue.

albertlast:
Can we merge this pr?

Sources/Themes.php
illori: how are you creating the variant? if you are creating a new theme then you are doing it incorrectly.
for SMF 2.0 this theme shows how it can be done https://custom.simplemachines.org/themes/index.php?lemma=2308

albertlast: i know,
but i'm not able to create this bug.
But this "workaround" create a bug which i can see.

-> So when no one is able to create the bug for this workaround,
than i see no reason to keep this.

live627:
The bug can be fixed by moving this below updateMemberData

Sources/Themes.php
live627: The bug can be fixed by moving this below updateMemberData

albertlast: Lets play your idea in our mind:
We got two styles: 
Curve2 with id 1
Lunar with id 2

The default(with id 0) shows at Curve2.
Let's assum curve2 got 3 variants and lunar 2.

Now we choose default (id 0) variant 3 (id 2),
the memberdata got as theme id 0 (this is good because we get now that what we selected)
Then the theme table get update ther we insert theme id 1 with variant id 2.

When now the admin change the default theme from curve2 to lunar,
the logic get broken.

Based on this is my guess:
You can't choose a variant on the default themen(because it can be changed every minute by the admin),
when someone choose the default the entry in the theme table needs to be deletet.

albertlast:
@colinschoen @Sesquipedalian what is with this pr?

Sesquipedalian:
My available time is limited, and this one takes a bit more work than usual to test properly. :)

albertlast:
When you don't hav the time to test this.
Than merge this pr and wait for the feedback of the tester,
when they find issue we could revert this pr or do improve this one,
since ther are no database changes is ther no permant "damage" possible.

Sesquipedalian:
Okay, I finally found some time to set up a theme with variants and test this. It seems to fix the bug and I don't see any problems.

Sesquipedalian (approved):

PR #4528, opened 2018-02-05
Better looking error log

albertlast:
I saw this version from @live627 at elkarte: https://www.elkarte.net/community/index.php?topic=4910.0
and i liked so i created a smf version.

Old:
grafik
New:
grafik

I removed the magnifier from the file and place the filter function on the word "File"

Maybe @Antes or @Gwenwyfar could say about this something.

### DIFF
--- a/Themes/default/Errors.template.php
+++ b/Themes/default/Errors.template.php
@@ -154,9 +154,9 @@ function template_error_log()
 
 		if (!empty($error['file']))
 			echo '
-							<a href="', $scripturl, '?action=admin;area=logs;sa=errorlog', $context['sort_direction'] == 'down' ? ';desc' : '', ';filter=file;value=', $error['file']['search'], '" title="', $txt['apply_filter'], ': ', $txt['filter_only_file'], '"><span class="generic_icons filter"></span></a>
 							<div>
-								', $txt['file'], ': ', $error['file']['link'], '<br>
+								<a href="', $scripturl, '?action=admin;area=logs;sa=errorlog', $context['sort_direction'] == 'down' ? ';desc' : '', ';filter=file;value=', $error['file']['search'], '" title="', $txt['apply_filter'], ': ', $txt['filter_only_file'], '">'
+				. '					', $txt['file'], '</a>: ', $error['file']['link'], '<br>
 								', $txt['line'], ': ', $error['file']['line'], '
 							</div>';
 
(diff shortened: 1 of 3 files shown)


XinYenFon:
with all due respect but new one looks more complicated.

albertlast:
Where is in your eyes the complexity?
I mean the time formation are on the right,
the content on the left.
Less space used in the height.

Gwenwyfar:
Looks better for me. What exactly is "::1" supposed to be, though?

albertlast:
::1 is the ip address and ich1 is my nickname in my test bord.

Gwenwyfar (approved):

frandominguezl:
Can't test this right now, but just to check, does it keeps responsive the design, right?

Gwenwyfar:
Yeah, pretty much everything in Curve2 is in percentages, so it scales to a fair amount. Any smaller details can be sorted out later.

albertlast:
Old:
grafik
New:
grafik

But i got not so mutch knowledg about design stuff,
so maybe some improvment we could also here

frandominguezl:
Fine by me.

live627:
the icons should be unselectable

albertlast:
the icons(magnifier?) are needed to filter one session/user/etc...
i see no alternative way to implement this.

albertlast:
Possible alternative solution would be to replace the "File:" with the icon.
grafik
The user experience could be here more consistent.

albertlast:
The last commit move the line number into the file:
grafik

illori:
i assume that is the session id on the right side? why do we need that logged in the first place? IMO to most admin that is not necessary to figure out the errors cause.

even to a point the IP address is not that much required.

Sesquipedalian:
I would keep the IP address. It can be relevant for certain types of errors. But the session id is useless. It would be much more useful to include the member id of the member who triggered the error.

Sesquipedalian:
> It would be much more useful to include the member id of the member who triggered the error.

Derp. We already record that in the log. Never mind. 😅

albertlast:
Since you can get multiple sesson from on ip i see no advantage to removed this.
Also the session id is already on last position and
with the free space we win nothing,
BUT we could work like git and only show the last/first x chars.

Old:
grafik
New:
grafik

XinYenFon:
For better responsiveness, add the following after (responsive.css);
```
	#admincenter input[type="text"], #quick_search input[type="search"],
	#admincenter select, #quick_search select {
		width: 90%;
	}
```
```
	.error_info {
		width: 100% !important;
		padding: 0 !important;
		float: none;
	}
```

colinschoen (approved):
LGTM approval

PR #4529, opened 2018-02-09
Droped reconnect logic in mysql fixes #4516

### LINKED ISSUE #4516: Persistent connection issue (2.0 & 2.1)

albertlast:
Since the reconnect didn't exists in mysql_* world and smf 2.0.x works fine with it and
i see more issues then solutions about this logic.
So i came to the idea to drop the logic.
Issue: #4516

### DIFF
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -368,34 +368,6 @@ function smf_db_query($identifier, $db_string, $db_values = array(), $connection
 	// Decide which connection to use.
 	$connection = $connection === null ? $db_connection : $connection;
 
-	// Get a connection if we are shutting down, sometimes the link is closed before sessions are written
-	if (!is_object($connection))
-	{
-		global $db_server, $db_user, $db_passwd, $db_name, $db_show_debug, $ssi_db_user, $ssi_db_passwd;
-
-		// Are we in SSI mode?  If so try that username and password first
-		if (SMF == 'SSI' && !empty($ssi_db_user) && !empty($ssi_db_passwd))
-		{
-			if (empty($db_persist))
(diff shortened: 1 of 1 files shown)


sbulen:
I gave this one a shot but I'm now getting lots of errors, just bringing up posts:
```
Warning
: mysqli_prepare() expects parameter 1 to be mysqli, null given in
C:\Program Files (x86)\EasyPHP-Devserver-16.1\eds-www\VAN21\Sources\Subs-Db-mysql.php
on line
973


Warning
: mysqli_stmt_bind_param() expects parameter 1 to be mysqli_stmt, null given in
C:\Program Files (x86)\EasyPHP-Devserver-16.1\eds-www\VAN21\Sources\Subs-Db-mysql.php
on line
981


Warning
: mysqli_stmt_execute() expects parameter 1 to be mysqli_stmt, null given in
C:\Program Files (x86)\EasyPHP-Devserver-16.1\eds-www\VAN21\Sources\Subs-Db-mysql.php
on line
982
```

albertlast:
Oops didn't notice that this is the open pr...
But like i ask in the issue:

@sbulen are ther more warning from other files/line of code?
Or ways to reproduce this.

sbulen:
I just did a vanilla 2.1 install, & navigated to the Welcome message.

albertlast:
Moved the install file from others,
install smf4 fresh
jump to welcome post
nothing happen

only the firstindex error appears,
because this got fixed in a different pr.

albertlast:
@Sesquipedalian @colinschoen can we merge this?

sbulen:
I have retested, & this looks OK now.

PR #4534, opened 2018-02-09
Changed CLI table from tinytext to varchar fixes #4530

### LINKED ISSUE #4530: BLOB/TEXT column 'txt' can't have a default value
upgrade.php is giving this output and nothing else. it worked fine yesterday but not today.

BLOB/TEXT column 'txt' can't have a default value

albertlast:
@illori please test this.
Issue: #4530

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -2368,7 +2368,7 @@ Usage: /path/to/php -f ' . basename(__FILE__) . ' -- [OPTION]...
 	$create = $smcFunc['db_create_table']('{db_prefix}priv_check', array(array('name' => 'id_test', 'type' => 'int', 'size' => 10, 'unsigned' => true, 'auto' => true)), array(array('columns' => array('id_test'), 'primary' => true)), array(), 'overwrite');
 
 	// ALTER
-	$alter = $smcFunc['db_add_column']('{db_prefix}priv_check', array('name' => 'txt', 'type' => 'varchar', 'size' = 255, 'null' => false, 'default' => ''));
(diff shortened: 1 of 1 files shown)


other/upgrade.php
```diff
@@ -2368,7 +2368,7 @@ function cmdStep0()
 	$create = $smcFunc['db_create_table']('{db_prefix}priv_check', array(array('name' => 'id_test', 'type' => 'int', 'size' => 10, 'unsigned' => true, 'auto' => true)), array(array('columns' => array('id_test'), 'primary' => true)), array(), 'overwrite');
 
 	// ALTER
-	$alter = $smcFunc['db_add_column']('{db_prefix}priv_check', array('name' => 'txt', 'type' => 'tinytext', 'null' => false, 'default' => ''));
+	$alter = $smcFunc['db_add_column']('{db_prefix}priv_check', array('name' => 'txt', 'type' => 'varchar', 'size' => 255, 'null' => false, 'default' => ''));
```
Sesquipedalian: The equivalent on line 645 uses `'size' => 4`. It doesn't really make a difference, since we drop the table in the very next command, but for the sake of consistency, we may as well use the same size in both cases. So, could you please change this from `'size' => 255` to `'size' => 4`?

Sesquipedalian (changes requested):

Sesquipedalian (approved):

Sesquipedalian:
Does this fix the issue for you, @illori?

illori:
yes it does.

PR #4535, opened 2018-02-09
Added housekeeping for imageproxy fixes #4449

### LINKED ISSUE #4449: Proxy Enhancement:  Cache size setting
After watching my (2.0 proxy) cache grow over 600MB of files a day, I realized we probably need some cache management...

I propose a user-configurable cache size setting, that is enforced via a Daily maintenance task.   (It grows too quickly for Weekly maintenance, on my site anyway...)

This applies to 2.0 & 2.1.

albertlast:
This is a houskeeping implemtation,
which try to delete all files which get older than 5 days.
In my eyes is this solution very fast (since we don't need to read the file) and
allow so to run accross many files.

@sbulen please test the code (i got not the env for this)

issue: #4449


Sesquipedalian:
This works as a basic housekeeping method, but the benefit of @sbulen's [proxy-maint-cron.php](https://github.com/sbulen/sjrbTools/blob/master/proxy-maint-cron.php) is that it reduces the image cache down to a certain size and prioritizes the files based on the date they were lasted accessed instead of just deleting anything that was last modified more than a certain number of days ago. I think it would be better to use an approach like that.

jdarwood007:
I like the fact you kept the name the same as the cacheAPI one.  If we start getting more of these, it will be easy to create a loop and run these all quickly and easily.

sbulen:
In my experience the proxy cache is highly volatile - I believe this should be daily, not weekly.  Or, outside of the standard daily/weekly in its own script, so admins can set the frequency.

albertlast:
in my eyes the imageproxy is a cache and
when you look in to our cache logic you see,
every entry got a ttl and the same way use this pr.
The proxy-maint-cron create to mutch i/o in my eyes and
it's posible that images keep ther for ever -> this no cache than any more, the images are saved.
When you save image you need the copyright to do so,
when you can be 100% sure that you had them,
than feel free to do so.

albertlast:
Small update,
in the pr https://github.com/SimpleMachines/SMF2.1/pull/4550 i added the cache information into the http header -> the browser will use his own cache
So from proxy side is the usage not visable any more.

Also when one of the both pr get merge the other need to needs resolve conflict.

proxy.php
```diff
@@ -195,9 +201,38 @@ protected function cacheImage($request)
...
+			}
+
+			closedir($handle); 
+		}
+	}
 }
 
-$proxy = new ProxyServer();
-$proxy->serve();
+if(empty($proxyhousekeeping))
```
colinschoen: `if(` -> `if (`

Sources/ScheduledTasks.php
```diff
@@ -458,6 +458,17 @@ function scheduled_daily_maintenance()
 			'oldLogins' => time() - (!empty($modSettings['loginHistoryDays']) ? 60 * 60 * 24 * $modSettings['loginHistoryDays'] : 2592000),
 	));
 
+	// Run Cache housekeeping
+	if (!empty($image_proxy_enabled))
+	{
+		global $proxyhousekeeping;
```
colinschoen: ^ Can we add this as a constructor argument to ProxyServer which stores the value as a property.

colinschoen (changes requested):

proxy.php
```diff
@@ -195,9 +201,38 @@ protected function cacheImage($request)
...
+	public function housekeeping()
+	{
+		$path = $this->cache . '/';
+		if ($handle = opendir($path)) {
+
+			while (false !== ($file = readdir($handle)))
+			{ 
+				$filelastmodified = filemtime($path . $file);
+
+				if((time() - $filelastmodified) > ($this->maxDays * 86400))
```
colinschoen: `if(` -> `if (`

Sources/ScheduledTasks.php
albertlast: The issue is here that this logic:
https://github.com/albertlast/SMF2.1/blob/5340d887d2a01da781ebb86da83a1359f0975476/proxy.php#L232-L236
run in this moment when you load the file with require_once,
wihtout loading the file i can't set the constructor.

Sesquipedalian:
@colinschoen: All your requested changes have been implemented. Can we merge this?

colinschoen (approved):
LGTM

PR #4548, opened 2018-02-15
Define load order of css fixes #4544 

### LINKED ISSUE #4544: Responsive.css should load last
To avoid over-usages of !important and more importantly avoid having issues related custom integrations etc... this file should be loaded last so its content will become the most important without usage of !important.

albertlast:
Extend the params about order_pos,
index is in antes eyes the start = 1 and responsive the last = 9000
The implementation allow also that two css want the same position,
than the first get the orginal position and the second get the nearst free position.

Issue: #4544

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -2145,7 +2145,7 @@ function loadTheme($id_theme = 0, $initialize = true)
 	loadCSSFile('responsive.css', array('force_current' => false, 'validate' => true, 'minimize' => true, 'order_pos' => 9000), 'smf_responsive');
 
 	if ($context['right_to_left'])
-		loadCSSFile('rtl.css', array('order_pos' => 20), 'smf_rtl');
+		loadCSSFile('rtl.css', array('order_pos' => 200), 'smf_rtl');
 
 	// We allow theme variants, because we're cool.
 	$context['theme_variant'] = '';
@@ -2168,9 +2168,9 @@ function loadTheme($id_theme = 0, $initialize = true)
 
 		if (!empty($context['theme_variant']))
 		{
-			loadCSSFile('index' . $context['theme_variant'] . '.css', array('order_pos' => 30), 'smf_index' . $context['theme_variant']);
+			loadCSSFile('index' . $context['theme_variant'] . '.css', array('order_pos' => 300), 'smf_index' . $context['theme_variant']);
 			if ($context['right_to_left'])
-				loadCSSFile('rtl' . $context['theme_variant'] . '.css', array('order_pos' => 40), 'smf_rtl' . $context['theme_variant']);
+				loadCSSFile('rtl' . $context['theme_variant'] . '.css', array('order_pos' => 400), 'smf_rtl' . $context['theme_variant']);
 		}
 	}
 
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -2490,7 +2490,7 @@ function loadCSSFile($fileName, $params = array(), $id = '')
 	{
 		// find a free number/position
 		while (isset($context['css_files_order'][$params['order_pos']]))
-				$params['order_pos']++;
+			$params['order_pos']++;
 		$context['css_files_order'][$params['order_pos']] = $id;
 		
 		$context['css_files'][$id] = array('fileUrl' => $fileUrl, 'filePath' => $filePath, 'fileName' => $fileName, 'options' => $params);
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -2139,13 +2139,13 @@ function loadTheme($id_theme = 0, $initialize = true)
 	$settings['lang_images_url'] = $settings['images_url'] . '/' . (!empty($txt['image_lang']) ? $txt['image_lang'] : $user_info['language']);
 
 	// And of course, let's load the default CSS file.
-	loadCSSFile('index.css', array('minimize' => true), 'smf_index');
+	loadCSSFile('index.css', array('minimize' => true, 'order_pos' => 1), 'smf_index');
 
 	// Here is my luvly Responsive CSS
-	loadCSSFile('responsive.css', array('force_current' => false, 'validate' => true, 'minimize' => true), 'smf_responsive');
+	loadCSSFile('responsive.css', array('force_current' => false, 'validate' => true, 'minimize' => true, 'order_pos' => 9000), 'smf_responsive');
 
 	if ($context['right_to_left'])
-		loadCSSFile('rtl.css', array(), 'smf_rtl');
+		loadCSSFile('rtl.css', array('order_pos' => 20), 'smf_rtl');
 
 	// We allow theme variants, because we're cool.
 	$context['theme_variant'] = '';
@@ -2168,9 +2168,9 @@ function loadTheme($id_theme = 0, $initialize = true)
 
 		if (!empty($context['theme_variant']))
 		{
-			loadCSSFile('index' . $context['theme_variant'] . '.css', array(), 'smf_index' . $context['theme_variant']);
+			loadCSSFile('index' . $context['theme_variant'] . '.css', array('order_pos' => 30), 'smf_index' . $context['theme_variant']);
 			if ($context['right_to_left'])
-				loadCSSFile('rtl' . $context['theme_variant'] . '.css', array(), 'smf_rtl' . $context['theme_variant']);
+				loadCSSFile('rtl' . $context['theme_variant'] . '.css', array('order_pos' => 40), 'smf_rtl' . $context['theme_variant']);
 		}
 	}
 
@@ -2428,11 +2428,15 @@ function loadSubTemplate($sub_template_name, $fatal = false)
  *  - ['rtl'] (string): additional file to load in RTL mode
  *  - ['seed'] (true/false/string): if true or null, use cache stale, false do not, or used a supplied string
  *  - ['minimize'] boolean to add your file to the main minimized file. Useful when you have a file thats loaded everywhere and for everyone.
+ *  - ['order_pos'] int define the load order, when not define it's loaded in the middle, before index = - 500, after index = 500, middle = 3000 end(after responsive) = 10000
  * @param string $id An ID to stick on the end of the filename for caching purposes
  */
 function loadCSSFile($fileName, $params = array(), $id = '')
 {
 	global $settings, $context, $modSettings;
+	
+	if (empty($context['css_files_order'])) 
+			$context['css_files_order'] = array();
 
 	$params['seed'] = (!array_key_exists('seed', $params) || (array_key_exists('seed', $params) && $params['seed'] === true)) ? (array_key_exists('browser_cache', $modSettings) ? $modSettings['browser_cache'] : '') : (is_string($params['seed']) ? ($params['seed'] = $params['seed'][0] === '?' ? $params['seed'] : '?' . $params['seed']) : '');
 	$params['force_current'] = isset($params['force_current']) ? $params['force_current'] : false;
@@ -2440,6 +2444,7 @@ function loadCSSFile($fileName, $params = array(), $id = '')
 	$params['minimize'] = isset($params['minimize']) ? $params['minimize'] : false;
 	$params['external'] = isset($params['external']) ? $params['external'] : false;
 	$params['validate'] = isset($params['validate']) ? $params['validate'] : true;
+	$params['order_pos'] = isset($params['order_pos']) ? (int) $params['order_pos'] : 3000;
 
 	// If this is an external file, automatically set this to false.
 	if (!empty($params['external']))
@@ -2482,7 +2487,14 @@ function loadCSSFile($fileName, $params = array(), $id = '')
 
 	// Add it to the array for use in the template
 	if (!empty($fileName))
+	{
+		// find a free number/position
+		while (isset($context['css_files_order'][$params['order_pos']]))
+				$params['order_pos']++;
+		$context['css_files_order'][$params['order_pos']] = $id;
+		
 		$context['css_files'][$id] = array('fileUrl' => $fileUrl, 'filePath' => $filePath, 'fileName' => $fileName, 'options' => $params);
+	}
 
 	if (!empty($context['right_to_left']) && !empty($params['rtl']))
 		loadCSSFile($params['rtl'], array_diff_key($params, array('rtl' => 0)));
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -2436,7 +2436,7 @@ function loadCSSFile($fileName, $params = array(), $id = '')
 	global $settings, $context, $modSettings;
 	
 	if (empty($context['css_files_order'])) 
-			$context['css_files_order'] = array();
+		$context['css_files_order'] = array();
 
 	$params['seed'] = (!array_key_exists('seed', $params) || (array_key_exists('seed', $params) && $params['seed'] === true)) ? (array_key_exists('browser_cache', $modSettings) ? $modSettings['browser_cache'] : '') : (is_string($params['seed']) ? ($params['seed'] = $params['seed'][0] === '?' ? $params['seed'] : '?' . $params['seed']) : '');
 	$params['force_current'] = isset($params['force_current']) ? $params['force_current'] : false;
@@ -2492,7 +2492,7 @@ function loadCSSFile($fileName, $params = array(), $id = '')
 		while (isset($context['css_files_order'][$params['order_pos']]))
 			$params['order_pos']++;
 		$context['css_files_order'][$params['order_pos']] = $id;
-		
+
 		$context['css_files'][$id] = array('fileUrl' => $fileUrl, 'filePath' => $filePath, 'fileName' => $fileName, 'options' => $params);
 	}
 
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -3633,6 +3633,9 @@ function template_css()
 	$toMinify = array();
 	$normal = array();
 
+	ksort($context['css_files_order']);
+	$context['css_files'] = array_merge(array_flip($context['css_files_order']), $context['css_files']);
+
 	foreach ($context['css_files'] as $id => $file)
 	{
 		// Last minute call! allow theme authors to disable single files.


Sesquipedalian (dismissed):
This looks good, except that responsive.css should not override CSS files added by mods. All the standard CSS files should have order_pos values between 0 and 2999, and then the mods from 3000 and up.

XinYenFon:
I disagree responsive.css should override everything, this will encourage mod authors to use default classes more and properly. Mods shouldn't be editing the core/default classes they should add their own if required.

Sesquipedalian:
> I disagree responsive.css should override everything, this will encourage mod authors to use default classes more and properly. Mods shouldn't be editing the core/default classes they should add their own if required.

That would depend on what the mod was doing. Depending on what the mod does, overriding the default classes might be appropriate. Just as with anything else, it is up to the mod authors to make sure their mods don't break anything.

Besides, there is nothing we can do to stop mod authors from overriding responsive.css if they want to. They would merely need to set an `order_pos` of 9001 on their own CSS file, and they will have done it.

albertlast:
Yeah this behavior is wanted,
they need to be active force ther css to be loaded later.
When the do nothing,
The css get loaded in the middle.

Sesquipedalian:
I suppose this is a disagreement over design philosophy. My understanding is that SMF is intended to be as mod-friendly as possible. Part of that means allowing mods to override aspects of our default CSS, including our responsive CSS, if they choose to. It also means that if the mod breaks the CSS with the changes, it is the responsibility of the modder to fix their mod so it doesn't. 

Conversely, forcing the standard CSS to override the CSS added by the mod is tantamount to trying to protect the end user from the modder. That's not mod-friendly. Indeed, it will probably make it very frustrating for the modder. She will be very confused why the CSS changes she needed to implement in her mod seem to work on desktops, but then on mobile screens they are ignored. In such a case, she will inevitably try to find some way to force her changes to override the standard CSS. The best case scenario is that she discovers the `order_pos` parameter and assigns her custom CSS file an order_pos of 9001. More likely is that she will simply shove her CSS into the HTML head, or (worst of all) try to perform a file edit directly on responsive.css itself. None of these are ideal.

In the end, I simply do not believe that trying to "protect" the end user from the modder is a good approach. It is our job to provide a good base product. It is the modder's job to write mods that work without breaking anything. It is the end user's job to check whether the mods they install break anything. If the end user installs a mod and then discovers that it breaks the CSS on small screens, they go and bother the mod author to fix it.

Besides which, trying to "protect" the end user from the modder is not something we can do anyway. There is no point in putting minor speed-bumps like this in the way of the modders, because they will just bypass or overwrite them so that they can make their mod work the way they want it to.

If either of you have a counter-argument to make, I am more than happy to give you the opportunity to convince me. (Seriously, I am. 😊) But unless and until that happens, I have no plan to merge this until `9000` is changed to `2999`.

albertlast:
This is no protection here,
this order exists for smf by self.
Because we got we know some files which needed load on start in spec order,
than some css where we don't care (reason why the default behavior is to add css in the middle).
And some files where should be load at the "end".
When you want to add css file you need to define where you want to add this css file,
on start = you need a number below 1, in the middle = you need to do nothing or the end = over 9000

This behavior is for css person easy to understand because they had to work with css load order and z-index.
And this solution is right the same.

Btw.
In my eyes is antes and gwenwyfar for gui responsible when they say it's a good solution,
than the solution is good.

XinYenFon:
@Sesquipedalian I see your point, you clearly saw mine :) Simply I'm ok with both just my way is to protect end user and enforce standards on mod authors so things work better for everything. If a mod author embraces default classes (without editing them) and introducing extra classes when needed (which is what we call proper coding) then mod author does not have to work more on supporting different themes which lowers the support time and gives mod author more time to improve his/her modification. Its not that we are only protecting end-user with enforcing default classes on mod authors we are increasing total quality in code, or shall say at least it is what i hope to see :)

wintstar:
This one's already discovered => [loadCSS](https://github.com/filamentgroup/loadCSS)

Gwenwyfar:
Responsive css should always load last as default. The only case scenarios where mod authors may want to specify custom css loading later is if they need to modify something in the responsive code itself, or something related to it. For most other standard custom css uses, it is not necessary nor wanted.

You're right in saying custom code should come after every other default css, but it doesn't work the same way with the responsive code. Having it come before their custom code will actually be the thing that may break it. If they modify a default value that comes from other files for example, that responsive also adjusts later, the whole chain in responsive.css will be broken, but it will still look fine in desktop.

If they have the option to simply change the index to load order if necessary, I see no problem with this. Actually, this is the best way to do it. Otherwise we risk the chance of having broken responsive code anytime authors add anything, which they may not even notice.

I'd say we just should make sure there is proper documentation on usage of the index, if they need to load it later, they will know how, and we leave a good standard for most use cases.

Sesquipedalian:
@Antes:
> If a mod author embraces default classes (without editing them) and introducing extra classes when needed (which is what we call proper coding) [...]

We are agreed that this is the ideal approach... except, of course, if the _purpose_ of the mod is to change something in the CSS. For example, [this mod](https://custom.simplemachines.org/mods/index.php?mod=4162)'s core functionality revolves around messing with the CSS. In this particular case, the mod works by injecting some custom CSS directly into the HTML head (and would therefore override responsive.css regardless of the order in which we loaded external CSS files), but the point remains that here we see a clear example of a situation where it would be wrong for responsive.css to override the mod's CSS.

> then mod author does not have to work more on supporting different themes which lowers the support time and gives mod author more time to improve his/her modification.

I don't see how that follows. Ensuring compatibility with Curve2's responsive.css doesn't do anything to ensure compatibility with other themes. Other themes will, by definition, have different content in their index.css and responsive.css files, so I don't see how ensuring a certain load order would do much to make supporting other themes any easier.

> Its not that we are only protecting end-user with enforcing default classes on mod authors we are increasing total quality in code, or shall say at least it is what i hope to see :)

To my mind, this goal is more likely to be achieved if the mod CSS is loaded after responsive.css, not before. If a mod's changes to an existing class are loaded **after** the standard responsive CSS, then it is clear and straightforward that the mod is _always_ responsible to include both (1) a basic version of their changes and (2) any responsive changes that may be required. In contrast, if we load the mod's CSS **before** responsive.css, it becomes rather difficult to say whether the mod should be responsible for including some responsive changes or not. The correct answer will vary from case to case, which means it will be less predictable, which means we'll see many more examples of inexperienced mod authors making the wrong choice.

@Gwenwyfar:
> You're right in saying custom code should come after every other default css, but it doesn't work the same way with the responsive code. Having it come before their custom code will actually be the thing that may break it. If they modify a default value that comes from other files for example, that responsive also adjusts later, the whole chain in responsive.css will be broken, but it will still look fine in desktop.

I like how you are framing this in terms of making sure that the chain of CSS priorities doesn't become broken, but I am not convinced by the content of the argument itself. My reason is this:

As everyone in this conversation already knows, each successive CSS file that we load changes the state of the overall CSS ruleset that the browser will apply. Indeed, the whole reason that we want to control the load order is so that we can be certain of exactly what the state of the ruleset is at every step in the process. So far, so good. But the thing is, responsive.css has been written assuming the known state that arises from loading the the rest of standard Curve2 CSS files. So long as we load responsive.css immediately after the rest of the standard Curve2 CSS files, everything works nicely: the rules in responsive.css are applied to a ruleset that is in a known state, and therefore we get nice, predictable results. But if we insert a mod's CSS into the load sequence before responsive.css, then when responsive.css is loaded, its changes will be applied blindly to a ruleset that is in an _unknown_ state. When that happens, we can no longer predict what sort of results we might get, which is the recipe for problems.

So, if we keep the mod's CSS out of the load sequence until we are done loading _all_ the standard CSS, including responsive.css, then we can maintain a consistent known state up until we have done everything we want the standard CSS to do. Once that is done, we can then turn over the ruleset in a known state to the mod so that it can do with the ruleset whatever it wants to do. If the mod proceeds to make a mess out of what we turned over to it, that's a separate matter and not our fault or our responsibility. But if we design our software so that  _by default_ it shoves unknown CSS files into the load sequence before loading responsive.css, then it is _our_ fault when responsive.css is applied to a ruleset that is in an unknown state.

Gwenwyfar:
Both ways will have problems with one mod or another. But leaving responsive in the middle not only has a much higher chance of giving problems, but many problems that will arise may not even be seen by the author. There's a big difference between it and the other files. Something as simple as changing the width of an element may break the responsive code if it doesn't come last, and the author will have much more trouble doing it all over again, or finding out how to even fix it (if he even sees the issue), and it may be something he wouldn't even need to touch otherwise. Between that and a lesser chance of problems that will most likely be noticed right away, the latter is preferable. 

There's no way around having no problems with mods here, it's the choice of less harm done. Most mods do small changes to the css, and these are exactly the ones that may break without being noticed. And even for complex modifications, authors may still prefer to leave their code in the middle and add their own responsive code last, because that is easier to work with than having to rewrite responsive code of your own on top of an existing one.

Sesquipedalian:
> leaving responsive in the middle not only has a much higher chance of giving problems,

<sup>*[citation needed]*</sup>

> Something as simple as changing the width of an element may break the responsive code if it doesn't come last,

It is equally the case that something as simple as changing the width of an element may break the responsive code if it **does** come last.

> the author will have much more trouble doing it all over again, or finding out how to even fix it

On the contrary, I believe it will make debugging much easier for the modder. If the mod is only loaded after all the standard CSS, then the modder can know that the issue is not due to something in the standard CSS overriding his custom CSS. The source of the problem can only possibly be in his mod's files, so he can confine his debugging investigation to them.

> (if he even sees the issue)

If the modder fails to test his mod correctly, the problem lies entirely with him and his bad practices. I do not believe that the best approach is to encourage him to continue in his bad practices.

(As an aside, I find it interesting that your argument here seems to be that we should load custom CSS before responsive.css in order to help badly written mods to work, whereas @Antes' argument was that he hoped doing so would push modders to improve the quality of their code.)

> And even for complex modifications, authors may still prefer to leave their code in the middle and add their own responsive code last, because that is easier to work with than having to rewrite responsive code of your own on top of an existing one.

It might or might not be easier in the sense of enabling lazy habits and poor quality coding to (sometimes) work. But it does not make things easier for people who want to code their modifications well. 

If a modder decides to change a preexisting CSS class, it is incumbent on the modder to ensure that his code works at all screen resolutions. If the mod's CSS is loaded after responsive.css, then the modder can put all his CSS into a single, self-contained file that will behave entirely predictably. That single file can contain both the general redefinition as well as any media queries. Not only will the changes be held all in one place, but they will all be loaded together at once and at the same base priority, which will effectively make the change atomic. That situation is simple and easy for the modder to work with: one file that just loads and does its thing, and if the changes don’t work right away on narrow screens, he can add a media query in the very same file to fix it, and then he’s done. In contrast, if the mod’s CSS is loaded before responsive.css and turns out not to play nicely with the latter, the modder first needs to analyze how responsive.css is interacting with his CSS, figure out what exactly is causing the break and why, then write *another* CSS file to override responsive.css, then learn how to get his second file to load after responsive.css so that he can override it, … and then finally ask himself why he doesn’t just load all his custom CSS after responsive.css and be done with it.

> There's no way around having no problems with mods here, it's the choice of less harm done.

Agreed. But overall, less harm will be done when (a) responsive.css is loaded into a ruleset in a known state rather than one in an unknown state, and (b) mods are able to start from a known state and finish in a known state without the expectation that they are going to get clobbered by another CSS file added later in the load sequence.

albertlast:
The documentation is here: https://github.com/SimpleMachines/SMF2.1/pull/4548/files#diff-e44bf29380cd171e044f5bf3ac967c8dR2431
It's define recommanded ranges where you want to put into your css.

a) When you want to load the css on a state where responsive.css is loaded than you go in 10k range.
b) you are talk about modS but you don't think ins modS, specialy when you got multiply modS than you had to manage position of you css files -> so you order of css can be get different when other mods exists in your setup
But the b) exists in smf 2.0.x and your solution keep this up in a more daramatic way as here implemented because everthing is at the end in your solution instead of only how are wanted.

Also when you don't like it,
you can place a mod for this to change this behavior.

Sesquipedalian:
Yes, I know, @albertlast. You documented the behaviour nicely and clearly, and the code is very straightforward. It's a good solution. :)

My concern is only that on [this line](https://github.com/SimpleMachines/SMF2.1/pull/4548/files#diff-e44bf29380cd171e044f5bf3ac967c8dR2447) you have defined the default `order_pos` value as 3000 (i.e. the "middle" position). I believe the default position should instead be set to the "end" position after responsive.css. I don't mind if you do it by lowering the `order_pos` for responsive.css from 9000 to 2999 or by raising the default from 3000 to 10000. Either way will be fine.

Gwenwyfar:
You seem to be basing yourself on the assumption responsive.css is similar to other files, and if it were, you'd be completely right. But there are some big differences:

First is that responsive doesn't focus on styling anything, that is all done on all the other files. Basic styling changes will go mostly unaffected. Responsive css is normally kept exclusive to width, display, size, font size definitions, etc. Anything else is avoided as that is what the main files are for.

Second is that unlike other files, responsive.css has an entire chain of css one after the other for the same elements. This means that if you replace one piece of it, you may well be replacing not one, but 5, each inside a different media query. So instead of having to fix one single bad replacement, you have to hunt down the entire chain of it (or, as is standard, you simply don't break it to begin with).

And lastly, this is not something for beginners, and this is not something we should expect most modders or most simple changes to deal with. We will have many mods breaking it, and even if they test properly, it is still going to be our fault for forcing them to fix it. Now, if something breaks because of the opposite, that is much easier to fix, and modders that are making large modifications are usually more experienced.

`
It might or might not be easier in the sense of enabling lazy habits and poor quality coding to (sometimes) work. But it does not make things easier for people who want to code their modifications well.`
That isn't "lazy habits", that's a standard and avoids needless replacements and messy !importants everywhere. This is how you do things properly. If I was writing a mod that used css I'd expect to not have to fix the basics.

`
(As an aside, I find it interesting that your argument here seems to be that we should load custom CSS before responsive.css in order to help badly written mods to work, whereas @Antes' argument was that he hoped doing so would push modders to improve the quality of their code.)`
The assumption is that people who are making small modifications should not have their code broken, and that is how even experienced people should be doing it. It's rather silly to force a standard that people will have to work their way around to fix, when for most cases they don't need to. Antes argument is still valid but I don't see it as a major factor. The thing with small mods/beginners is simply that they would be affected the most with this change.

And quite to the contrary of what you think about responsive.css would happen: Any mod that modified the base code and loads later than the responsive.css may break an entire chain of css. If it were left in the middle, the modder would not even have to do anything, but since we load his custom code last, he now has to deal with a whole chain of broken css, or go and change the load order so that it works properly. This is what we want to avoid.

The best use for modders is to have the standard mod code in the middle, and then his custom responsive css (if he has any) loaded after the main responsive.css, that is how the priorities go when you are working with that kind of file. In some cases it's better to not even add a second responsive file, but edit the existing one (although this can be problematic for mods to do). In others, there may not even be the need to move custom responsive.css up, such as an extra isolated element or page somewhere. Add to this that: how many modders do you think will bother with responsive code, when even many of the "responsive" themes in the theme site are barely responsive at all? That is a much more infrequent occurrence than the usual basic css changes, which most will be doing, and which should be loaded before any responsive code.

We're not tying modders hands and forbidding them from adding anything of their own last. We are setting a good default that will work better for most cases and leaving the choice open for those creating responsive css, or those that may need it for other reasons.

Sesquipedalian:
I understand what you are saying, @Gwenwyfar. Your concern is that you want a modder to be able to put this into her mod.css file:
```
#searchform .roundframe {
	padding: 50px;
}
```
... without also needing to include this in order to prevent the responsive design from breaking:
```
@media (max-width: 480px) {
	#searchform .roundframe {
		padding: 5px;
	}
}
```

You are right that this would be easier for the modder... or rather, it would be if it were not for the fact that we also have built-in CSS minification, and mod CSS files are not included in the minification by default, whereas responsive.css is. 

Because minification compresses all the minified files into one, anything that is not included in the minified version must either be loaded before it or after it. That means that manipulating the load sequence in order to allow the modder to avoid needing to add a media query is simply not going to work. If minification is turned off, the modder could get away with adding only the first bit of code above to her mod.css file. But if minification is turned on and her file is not included in it, she is going to need to add the second bit as well. So there is really no escaping from the reality that if a modder decides to alter a pre-existing CSS rule, she is required to take on responsibility for both the general rule as well as any responsive rules.

That being the case, it is clearer, simpler, and therefore ultimately easier for the modder if we keep the requirements and expectations consistent regardless of whether minification is turned on or off.

Sesquipedalian:
Well, I suppose we *could* create a situation that would allow her to escape from having to include the media query in mod.css, but only by doing one of the following:

1. Leave responsive.css out of minified.css, so that it can continue to be loaded after everything else.
2. Force (or at least default) everything with an `order_pos` less than that of responsive.css to be included in minified.css.

Both of those are undesirable options. 

So to my mind, it remains the case that, if a modder decides to alter a pre-existing CSS rule rather than creating a new one, she should be responsible for handling any responsive design changes that her decision requires. If she can't handle doing that, it is her job to learn how.

Sesquipedalian:
Just in case it wasn't clear, you have in fact convinced me that your way is right, @Gwenwyfar, aside from the part where we need to figure out what to do about minification.

albertlast:
Since the browser take the css from the cache in the common case, 
I don't understand what the the benefit of minified css is?

Sesquipedalian:
It reduces bandwidth consumption and the number of HTTP requests.

albertlast:
Well when we only would like to reduce the bandwidth (which is in my eyes near at zero because of cache),
we could minified all css by ther own and don't compress them to one file.

Sesquipedalian:
> (which is in my eyes near at zero because of cache)

Whether this is true or not will vary from server to server. Because we are merely inserting links to static files in the HTML head, we actually have no control over any of the HTTP headers that will be sent for those files. The browser will only cache the files if the underlying server software provides the instructions to do so, and we can't influence that.

albertlast:
Well the webserver had more task than running our php stuff,
when you server doesn't react as wanted then you should improve you webserver...

This is the behavior which i expected:
grafik
You see the css is taken from the cache because the eTag/lastChanged doesn't changed,
to handle this information is task of the webserver.

Sesquipedalian:
> when you server doesn't react as wanted then you should improve you webserver...

Yes, one should do that. But unfortunately we can't rely on that happening.

albertlast:
well like i said i understand the need to minified the css,
specialy for the first call since you open the browser the posibility is "high" that the information is not
cached by you browser.
Could we improve by minified all files by them self.

But to managed a big minified css,
which got posibility to got more css as needed -> size is bigger as the single css
i see no advantaged beside we had to manged the possible complex logic which could lead in more issue and more runtime as deliver all minified css version.

Sesquipedalian:
As I've been thinking more about this, I've realized that even if we set aside the question of what to do with CSS files added by mods and only consider the standard CSS files, we still need to figure out what to do about the way that minifying some CSS files but not others messes with the load order, and therefore the precedence, of different rule sets. 

I think I might have an idea for how to solve that, but I will need a little time to work on it. If my idea works, I will merge this PR and then submit another PR that builds on it. In the meantime, I'm just going to leave this PR open.

Sesquipedalian:
I tried out my idea to deal with the minification issue, and it worked nicely. So I am going to merge this now and then submit a PR with my further changes.

PR #4550, opened 2018-02-16
overhaul of proxy.php

albertlast:
Based on the idea/code from https://www.simplemachines.org/community/index.php?topic=558395.0
i decide to look more deeply in the code.
And did major changes:
- Removed subs.php -> better performance
- added request check of http/https (dunno why this make it but it doesn't hurt)
- stream line the return value of cacheImage to int
- define 3 return values -1 bad, 0 to big, 1 good
- based on this we can do different things
- define utc as default timezone for caching stuff no issue with daylight saving time
- added the header data that this is a cache and the browser can cache the data by his self for the same ttl like we got -> the proxy cache is less called

Since the proxy file works also with smf 2.0.x i guess @sbulen could maybe test this version and
comment on issue around this.

### DIFF
--- a/proxy.php
+++ b/proxy.php
@@ -31,6 +31,9 @@ class ProxyServer
 
 	/** @var string The cache directory */
 	protected $cache;
+	
+	/** @var int time() value */
+	protected $time;
 
 	/**
 	 * Constructor, loads up the Settings for the proxy
@@ -109,10 +112,7 @@ class ProxyServer
 			$this::redirectexit($request);
 		}
 
-		$old_timezone = date_default_timezone_get();
-		date_default_timezone_set('UTC');
-		$time = time();
-		date_default_timezone_set($old_timezone);
+		$time = $this->getTime();
 
 		// Is the cache expired?
 		if (!$cached || $time - $cached['time'] > (5 * 86400))
@@ -193,12 +193,9 @@ class ProxyServer
 		// Validate the filesize
 		if ($response['size'] > ($this->maxSize * 1024))
 			return 0;
-		
-		$old_timezone = date_default_timezone_get();
-		date_default_timezone_set('UTC');
-		$time = time();
-		date_default_timezone_set($old_timezone);
-		
+
+		$time = $this->getTime();
+
 		return file_put_contents($dest, json_encode(array(
 			'content_type' => $headers['content-type'],
 			'size' => $response['size'],
@@ -218,6 +215,24 @@ class ProxyServer
 	{
 		header('Location: ' . $request, false, 301);
 		exit;
+	}
+	
+	/**
+	 * Helper function to call time() once with the right logic
+	 * 
+	 * @return int
+	 */
+	protected function getTime()
+	{
+		if (empty($this->time))
+		{
+			$old_timezone = date_default_timezone_get();
+			date_default_timezone_set('UTC');
+			$this->time = time();
+			date_default_timezone_set($old_timezone);
+		}
+
+		return $this->time;
 	}
 }
 
--- a/proxy.php
+++ b/proxy.php
@@ -136,11 +136,11 @@ class ProxyServer
 			exit;
 
 		$max_age = $time - $cached['time'] + (5 * 86400);
-		header('Content-type: ' . $cached['content_type']);
-		header('Content-length: ' . $cached['size']);
-		header('Cache-Control: public, max-age=' . $max_age );
-		header('Last-Modified: ' . gmdate('D, d M Y H:i:s', $cached['time']) . ' UTC');
-		header('ETag: ' . $eTag);
+		header('content-type: ' . $cached['content_type']);
+		header('content-length: ' . $cached['size']);
+		header('cache-control: public, max-age=' . $max_age );
+		header('last-modified: ' . gmdate('D, d M Y H:i:s', $cached['time']) . ' UTC');
+		header('etag: ' . $eTag);
 		echo base64_decode($cached['body']);
 	}
 
--- a/proxy.php
+++ b/proxy.php
@@ -102,13 +102,15 @@ class ProxyServer
 
 		// Did we get an error when trying to fetch the image
 		$response = $this->checkRequest();
-		if ($response === -1) {
+		if ($response === -1)
+		{
 			// Throw a 404
 			header('HTTP/1.0 404 Not Found');
 			exit;
 		}
 		// Right, image not cached? Simply redirect, then.
-		if ($response === 0) {
+		if ($response === 0)
+		{
 			$this::redirectexit($request);
 		}
 
@@ -183,7 +185,8 @@ class ProxyServer
 		$responseCode = $curl_request->result('code');
 		$response = $curl_request->result();
 
-		if (empty($response) || $responseCode != 200) {
+		if (empty($response) || $responseCode != 200)
+		{
 			return -1;
 		}
 
--- a/proxy.php
+++ b/proxy.php
@@ -122,6 +122,13 @@ class ProxyServer
 				$this->serve();
 			$this::redirectexit($request);
 		}
+		
+		$eTag = '"' . substr(sha1($request) . $cached['time'], 0, 64) . '"';
+		if (!empty($_SERVER['HTTP_IF_NONE_MATCH']) && strpos($_SERVER['HTTP_IF_NONE_MATCH'], $eTag) !== false)
+		{
+			header('HTTP/1.1 304 Not Modified');
+			exit;
+		}
 
 		// Make sure we're serving an image
 		$contentParts = explode('/', !empty($cached['content_type']) ? $cached['content_type'] : '');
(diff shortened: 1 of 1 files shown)


proxy.php
```diff
@@ -159,41 +165,74 @@ protected function isCached($request)
 	 *
 	 * @access protected
 	 * @param string $request The image to cache/validate
-	 * @return bool|int Whether the specified image was cached or error code when accessing
+	 * @return int -1 error, 0 to big, 1 is good
```
colinschoen: *to* -> *too*

proxy.php
```diff
@@ -159,41 +165,74 @@ protected function isCached($request)
...
-		$request = $curl->get_url_data($request);
-		$responseCode = $request->result('code');
-		$response = $request->result();
+		$curl_request = $curl->get_url_data($request);
+		$responseCode = $curl_request->result('code');
+		$response = $curl_request->result();
 
-		if (empty($response))
-			return false;
+		if (empty($response)) {
```
colinschoen: `|| $responseCode != 200`

proxy.php
colinschoen: 1 is good -> 1 valid image

proxy.php
```diff
@@ -71,7 +72,7 @@ public function checkRequest()
 			if (!mkdir($this->cache) || !copy(dirname($this->cache) . '/index.php', $this->cache . '/index.php'))
 				return false;
 
-		if (empty($_GET['hash']) || empty($_GET['request']))
+		if (empty($_GET['hash']) || empty($_GET['request']) || ($_GET['request'] === "http:") || ($_GET['request'] === "https:"))
```
colinschoen, line 75: ^ What are these cases for?

colinschoen (dismissed):

proxy.php
albertlast: see https://www.simplemachines.org/community/index.php?topic=558395.msg3962226#msg3962226

albertlast:
By looking at some google pages,
i notice that the header stuff is always low.
So i changed the proxy stuff in the ways that this match.

tinoest:
That’s what the rfc states is the correct way to do it.

tinoest:
https://tools.ietf.org/html/rfc7540#section-8.1.2

However IE11 likes the first letter capitalised. So Content-Type: 

https://tools.ietf.org/html/rfc7230#section-3.2 States that it should be case insensitive for the receiver.

equinoxmatt:
Just a quick question regarding if statements.  Can we try and make it consistent?

Some have the opening bracket on the same line, others have them on the below line and some have none at all.

Is there a style guide? If there is, can we stick to it?  Better yet, start moving to PSR-2?

albertlast:
The normal style guide for smf is that bracket get in the new line,
but one line code doesn't need brackets.
Since the style stuff would hide my changes i'm not intrested in this pr to change the style to the smf standard.
Feel free to do this afterwards.

Sesquipedalian:
> Since the style stuff would hide my changes i'm not intrested in this pr to change the style to the smf standard.

On the lines where you are making changes anyway, it wouldn't hurt anything to also correct the style. For example, [here](https://github.com/SimpleMachines/SMF2.1/pull/4550/files#diff-95dde0693614aded154dc51012aba3b8R111) you may as well just move the bracket down to the next line. There are a few other places in the PR that you could do the same.

albertlast:
@Sesquipedalian i changed this.

Sesquipedalian:
@albertlast: Thanks. 

@colinschoen: How does this PR look to you now?

albertlast:
change the internal timezone to gmt

Sesquipedalian:
Why did you change from UTC to GMT? UTC never varies, but GMT switches between daylight saving time and standard time.

albertlast:
First gmt didn't got dst and more important the standard say you had to use gmt: https://stackoverflow.com/questions/21120882/the-date-time-format-used-in-http-headers

Sesquipedalian:
I stand corrected. :)

proxy.php
```diff
@@ -101,32 +102,47 @@ public function serve()
...
 			header('HTTP/1.0 404 Not Found');
 			exit;
 		}
+		// Right, image not cached? Simply redirect, then.
+		if ($response === 0)
+		{
+			$this::redirectexit($request);
+		}
+
+		$time = $this->getTime();
```
Sesquipedalian, line 117: I don't think you need the `getTime()` function. Changing the value of the default time zone doesn't affect the result of `time()`.
```
date_default_timezone_set('America/Edmonton');
$t1 = time();
date_default_timezone_set('GMT');
$t2 = time();
print_r(array($t1, $t2));
/* Result: 
Array
(
    [0] => 1520449992
    [1] => 1520449992
)
*/
```

proxy.php
Sesquipedalian: Rather than testing those specific cases, it would be better to test if `$_GET['request']` is a valid URL. This should do it:
```
$_GET['request'] = filter_var($_GET['request'], FILTER_VALIDATE_URL);
if (empty($_GET['hash'] || empty($_GET['request']))
	return false;
```

Note: Unfortunately, `filter_var($url, FILTER_VALIDATE_URL)` doesn't work for URLs that contain non-ASCII characters, so this will fail for an image hosted at, e.g., http://www.exämple.com/image.jpg. But I am just finishing putting together a PR to fix that. So for now go ahead and use `$_GET['request'] = filter_var($_GET['request'], FILTER_VALIDATE_URL);`, and then I will replace that line once the other PR is merged.

proxy.php
```diff
@@ -159,41 +175,71 @@ protected function isCached($request)
...
 		$curl = new curl_fetch_web_data(array(CURLOPT_BINARYTRANSFER => 1));
-		$request = $curl->get_url_data($request);
-		$responseCode = $request->result('code');
-		$response = $request->result();
-
-		if (empty($response))
-			return false;
+		$curl_request = $curl->get_url_data($request);
+		$responseCode = $curl_request->result('code');
+		$response = $curl_request->result();
```
Sesquipedalian, line 186: Right now, the proxy only works if the curl library was included when PHP was compiled (it usually is, of course, but not always). It would be nice to include here [some of the fallback logic](https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs-Package.php#L3141-L3253) from the `fetch_web_data()` function in Subs-Package.php.

Sesquipedalian (changes requested):

proxy.php
albertlast: Well the whole subs* stuff is my eyes is to mutch overhead for this that we don't needs this in the common case,
on the second hand we provide with the external proxy stuff a second solutions.
When your enverment doesn't met this requirement than it's "bad luck".

proxy.php
albertlast: It's not the intention here to test the url,
only to catch some cases where the stuff is empty,
when the url is "http:" or "https:" means that it got nothing.

proxy.php
Sesquipedalian: Sorry if I wasn't clear. I don't mean that we should include the entire Subs-Package file. I mean that we should copy and paste (with necessary changes) the relevant code from that function into this one.

proxy.php
Sesquipedalian: > when the url is "http:" or "https:" means that it got nothing.

That's a basic form of URL validation. If you are going to check that here, you may as well validate it entirely. And really, it is a good idea to validate the URL at this point. If the URL isn't even valid, then we know the proxy will fail to retrieve anything from it, so we should abort. I think you were on the right track to validate here. :)

proxy.php
Sesquipedalian: But if you would prefer not to do this, that's fine. I can do it myself afterwards. :)

proxy.php
albertlast: What happen when if it failed?
Brutforce is not the issue in my eyes because of the hash check,
so url came from smf and i would suggest that smf already checked the url -> no need to do this here.
And the existing solution didn't do it, too.

proxy.php
albertlast: Feel free to add this after this pr.

Sesquipedalian: Okay, that's fine.

proxy.php
Sesquipedalian: `filter_var($_GET['request'], FILTER_VALIDATE_URL)` will return `false` if `$_GET['request']` is not a valid URL. Then the `empty($_GET['request'])` will be triggered in the if statement.

Sesquipedalian: You are correct that the existing code didn't check the URL properly, but it should have. The fact it wasn't being checked properly is the reason why you had to add ` || ($_GET['request'] === "http:") || ($_GET['request'] === "https:")` to the if statement. If the URL was being properly checked, adding those cases would not have been necessary to begin with.

colinschoen (approved):
LGTM

Apologies for delay

albertlast:
Well i was planing to remove the timezone logic
two ways:
@Sesquipedalian provide this change on his pr
or i will create a small pr today with this changes 
i see no need for a revert of this pr.

PR #4551, opened 2018-02-18
[js] added avatar fallback when not able to load fixes #4343

### LINKED ISSUE #4343: Missing avatars breaking layout

albertlast:
Use bubbling method -> no problem when img avatar get added at runtime and no issue when other event are added
this will only affected img html tags how got the avatar class
and when the src != avatar default is (protect us from endless loop).

issue: #4343

### DIFF
--- a/Themes/default/scripts/script.js
+++ b/Themes/default/scripts/script.js
@@ -1665,3 +1665,20 @@ $(function()
 		return typeof actOnElement !== "undefined" ? smfSelectText(actOnElement, true) : smfSelectText(this);
 	});
 });
+
+function avatar_fallback(e) {
+    var e = window.e || e;
+	var default_avatar = '/avatars/default.png';
+	var default_url = document.URL.substr(0,document.URL.lastIndexOf('/')) + default_avatar;
+	
+    if (e.target.tagName !== 'IMG' || !e.target.classList.contains('avatar') || e.target.src === default_url )
+        return;
+
+	e.target.src = default_url;
+	return true;
+}
+
+if (document.addEventListener)
+    document.addEventListener("error", avatar_fallback, true);
+else
+    document.attachEvent("error", avatar_fallback);


albertlast:
@jdarwood007 can you please test this pr?

Themes/default/scripts/script.js
```diff
@@ -1665,3 +1665,20 @@ $(function()
 		return typeof actOnElement !== "undefined" ? smfSelectText(actOnElement, true) : smfSelectText(this);
 	});
 });
+
+function avatar_fallback(e) {
+    var e = window.e || e;
+	var default_avatar = '/avatars/default.png';
+	var default_url = document.URL.substr(0,document.URL.lastIndexOf('/')) + default_avatar;
```
Sesquipedalian, line 1672: This would still only work if the default avatar image is available at $boarddir/avatars/default.png, which means that if the default avatar image isn't in the expected location, this often won't be able do much to help.

But if you change this
```
	var default_avatar = '/avatars/default.png';
	var default_url = document.URL.substr(0,document.URL.lastIndexOf('/')) + default_avatar;
```

to this
```
	var default_url = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAGQAAABkCAMAAABHPGVmAAABlVBMVEUAAABVfqBVfqBVfqBVfqBVfqBVfqBVfqBVfqBVfqBqjKiSpraqtr+6wMXEx8nKy8tVfqBVfqB5lq2lsr7AxMjMzMxVfqBghaSdrbvAxMhVfqCwusKDnLG4v8WCnLG7wcZ3lKy3v8RVfqBcgqKuuMFVfqCcrbpVfqB8l65VfqCBm7DFyMpVfqCotL9VfqBvj6nBxchVfqCXqbhVfqBVfqBliKbDxslVfqCGnrJVfqCst8BVfqBVfqBVfqDGyMrKystVfqBVfqC8wca+w8eNo7VehKNVfqB7l66Qpba+w8fJysvBxchVfqBVfqBih6WLobSksr28wsbLy8xVfqBVfqBhhqSptb/MzMxVfqBykauZqrnIycpVfqB5lq2isLzExslVfqBVfqCotL/Cxchqi6ibrLqJoLPLy8uls76PpLbCxshbgqLHycpVfqCjsb1agaKhsLxVfqCIoLNykatcg6OvucGUqLdrjKjFx8misb1kiKaRpbZVfqBhhqTEx8mFnbKerruvucFVfqC6wMVVfqBVfqDKystLizbtAAAAh3RSTlMACBorOENKAA4nVGd3hZCXBCJbc4uZEFBujBd8X4NfhlmDAk96IW0PXBFekil2ClaNHWowBVKOFGEseTc/RpKXTBmHiWVQJltmiJWMEypRY3OImAsfUXeYJVhrlAlacY8HIHWNVGxil3Rljk+SLXNOcShiV098aFWQcVJnL1CPYG57OYZCSZb7UfAQAAAC6klEQVR42u3X91MaQRTAcQEliBAP5aGcYDm7Yj+xg4YIxi5YKIpYsXfsveTvzsUklojHwe6b0Rm+P9/MZ27udvdtSrJkyT5fMrkiNU2pTEtVyGVf4ilF8pOqdHWGRvs1M/OrVpOhTlchIIwuK1sPT+mzs3QMbcSQk2uEVxlzcwx0ETYrD96Ul8XSROQmM0TJbJLTQ9j8AohaQT5LC1EVFsE7FRWq6CBccQm8W0kxRwUpLQORykppIOUVRjHEWFFOAVFoQDSNghyprKoWR6qrKokRSw3EqMZCitTW1cdC6utqCZGGRohZYwMh0pQXG8lrIkP4ZmtsxNrMEyEtrSCh1hYi5Hm1i696IqStXQrS3kaEdJilIOYOIqRTLwXRdxIhXTYpiK2LCFHapSB2JRHSDZLq/nBvgv9N8P8u/HWCv+Lx9y78XRj/PME/GfHPePxpBX/uwp8g8Wdh/Kke/36Cf9P6PHdGRtWT/y264fjeq+hRMaQIzzpzXH2aH/1Rkf4Bs6bPleNk+cQRZnBoeGR0zOYGkdy2sdGR4aFBJiHEMz5hmnRMgYSmHJOmiXFP3IjK6/NrAyC5gNbv86riQqYtM8FZN8SVezY4Y5mWjITm1PMLkEAL8+q5kDSEXVzSQ4LplxZZCYgsfbkoDAkXLlpOl8VC2OEVGxBlWxlmxRGdywzEmV06EYRzrq4BhdZWndx7CL9etgFU2ihb56Mjod5gACgVCPaGoiH85pYdqGXf2uTfItz2DlBtZ5t7g+z6A3SRgH/3f6RhTzAoK3sNrxHDvgOo59g3vES4gwggFDngXiDjh2EMJHw4/ozIjvSAkv5I9oQ4RwGpUec/xOM7xkKOfZ6/iPcE0Drx/kH40wE8ZOCUf0TkZ4DYmfwROY9gIpHz3whzcYmJXF4wAjJ4BahdDQqI7hoXudYJyPkNLnJzLiC3d7jI3a2A3FtxEeu9gDz04yL9D4/IT9QeEcuDErUHi4Dgl0SSSBL5KMgvTLdZPzZWxnQAAAAASUVORK5CYII=';
```

then it will work no matter what happens.

Sesquipedalian (changes requested):

Themes/default/scripts/script.js
albertlast: i got so many question about your comment...
Why shouldn't be the default not ther?
When this is could happen, why we not check this on the upgrade process, if the directory and file are there?
Why you don't want to show the admin that his smf setup is wrong?
When the default image get change how you remeber to change this js file?

Sesquipedalian: One way that the image load error could occur would be if the user was supposed to use the default avatar, but `/avatars/default.png` didn't exist for some reason. In that case, this fallback would create the same error again.

But you are right that we should use `default.png` if it is available. So in this function I recommend using jQuery's `$.get()` to check if `default.png` exists. If it does, great. If not, then change the value of `default_url` to the data URI that I suggested in my previous comment.

albertlast: When the default.png didn't exists this existing check detect this(https://github.com/SimpleMachines/SMF2.1/pull/4551/files#diff-1395d4e6d032272170e3f17499906d1bR1674):
```
e.target.src === default_url 
```
So we could place here fallback logic.

But i don't like it to place binary in the js file,
which is 99,999% of the time useless -> garbage.
In my eyes the admin should know that his smf setup is not valid -> so that he can fix it.
Based on this i would place a check in upgrade process.

Sesquipedalian: > i don't like it to place binary in the js file

I don't particularly like it either, but if the goal is to stop a missing avatar from messing up the layout, then we need to make sure some data is provided.

> In my eyes the admin should know that his smf setup is not valid -> so that he can fix it.

That would be better accomplished by generating errors in the error log.

> i would place a check in upgrade process.

That is a good idea, yes, but it would not help if the file is deleted sometime after the upgrade.

A better option might be to add some code in the PHP that will (re-)create default.png and blank.png in the `$modSettings['avatar_dir']` directory if they are missing. We could set up a function in Subs.php that will use `file_put_contents()` to write the data to disk if the file is missing, and then return the URL to the file, and then make an AJAX call to that function from inside your JavaScript error function. Your error function will receive the correct URL to use for default.png in response to the AJAX call, and then you can update the image src attribute with that response.

albertlast: >That is a good idea, yes, but it would not help if the file is deleted sometime after the upgrade.

How often happen to you that you delete the file?
Why should someone delete this?

Since ther is no fallback in existing pr,
the admin and everyone else see that that this file is missing and
know by looking into the browser that this file is missing.
We could do also place a "error message" in a div box somewhere in the html code on client side,
when the file is not found.

Gwenwyfar: I wouldn't add data URI into js either. We can prevent the missing avatars from ruining the layout by simply hiding the alt text, so it is fine if the admin moves the image or it is broken somehow. Who knows, maybe they removed it even because they want nothing to show. 

There's other ways the avatar could be broken (including users putting a bad link), seems to me like these cases are more important than the admin breaking things/not installing properly. We really don't want to force people to go around replacing js code in data URI to change the default avatar...

A possible alternative would be to add a class instead, and change the content in the css.

Sesquipedalian: Alright, you've convinced me. I still think this would be a good idea...

> A better option might be to add some code in the PHP that will (re-)create default.png and blank.png in the $modSettings['avatar_dir'] directory if they are missing. We could set up a function in Subs.php that will use file_put_contents() to write the data to disk if the file is missing, and then return the URL to the file, and then make an AJAX call to that function from inside your JavaScript error function. Your error function will receive the correct URL to use for default.png in response to the AJAX call, and then you can update the image src attribute with that response.

... so that we can restore accidentally deleted default avatar files if necessary. But something like that would lie outside the scope of this PR.

Sesquipedalian (approved):

Sesquipedalian:
Note: my approval just means that I have withdrawn my previous request for changes. I'd still like to hear from @jdarwood007 and/or @Gwenwyfar whether this PR solves the problem in their tests before merging it.

jdarwood007:
I've been busy with RL stuff so haven't really had time to look into it

- Kinda annoying that we have to resort to a javascript based solution here to fix a theme issue.  Not sure why we can't present a default avatar when loading up the user settings for the current user.
- I swear SMF had its own hook for handling on load that prevented conflicts with other on load handlers.  Been a while since I looked into that code
- I see both sides about how the image is loaded.  Either way is fine really, but we should remember that SMF has support for using images and scripts in different urls, so this would only for for OOB setups that haven't changed around things.

albertlast:
From my point of view is that not a theme issue,
the problem is that user could use avatar image which could get be broken.
This issue happen on client side and to handle issue on client side is js perfect.
What you mean with "on load" is see no use of such the thing in my pr.
what is "oob"?
The url how is used for get the default avatar is the root of the smf not the of images/scriptes.

Gwenwyfar:
A css fallback is possible, but it is not compatible with Safari, so it would display no avatar in it.

We had discussed this on IRC and one other thing that could be done is remove the alt text, so that browsers handle it in their default mode (no alt to deal with). Upon looking further we should probably do this in any case:

>  When a screen reader comes across null alt text, it will completely skip over the image, not even announcing its presence. Someone who is fully sighted would easily skip right over a decorative image, null alt text allows someone using a screen reader to do the same thing.

I don't think it's very relevant to have an alt text for blind people there, if it displays even when no avatar is present.

As for the code itself: I'll do some testing later, but perhaps a simpler way to do it that may fix the URL issue is make the JS check only for the error, and add a class when one happens. Then we deal with it in the css.

jdarwood007:
@albertlast Not theme url, but avatar and custom avatar url.

Sesquipedalian:
> As for the code itself: I'll do some testing later, but perhaps a simpler way to do it that may fix the URL issue is make the JS check only for the error, and add a class when one happens. Then we deal with it in the css.

That sounds like an excellent idea, @Gwenwyfar.

albertlast:
From my point of view,
it doesn't matter if a css class is added or url ist changed.
But since i'm not the person how will provide this css stuff,
i would like to see to get merge this pr.
How ever like can change the logic afterwards of this.

Sesquipedalian:
Fair enough, @albertlast. :)

@Gwenwyfar, after I merge this, if you would like to take a stab at changing the content of albertlast's `avatar_fallback()` function so that it sets a CSS class, that'd be great. That approach would be very clean and much easier for modders and themers to work with, so I'd like to see it happen if you can make it work.

PR #4555, opened 2018-02-21
Links directly to messages in permalinks on topic display

Sesquipedalian:
In 54683d86e, @Oldiesmann rightly changed the URL in `$message['link']` to `$scripturl . '?msg=' . $message['id_msg']` in order to "prevent things from breaking if a post is moved, etc." This implements the same change for `$message['href']`.


colinschoen (approved):
LGTM

PR #4561, opened 2018-02-21
Allow array in $_REQUEST fixes #4558

### LINKED ISSUE #4558: quick moderation icons/lock
i changed to quick moderation icons, then i used the lock option. error log reports the following
	Type of error: General
2: strlen() expects parameter 1 to be string, array given http://localhost:8080/github2.1/index.php?action=quickmod;board=1.0;actions%5B111%5D=lock;b393f9d=d524a94797fb7ccadc396da9e276d3ea
D:/wamp/www/github2.1/Sources/Logging.php (Line 28)

albertlast:
Quick fix use $_get in form of &action[1234]='sticky' -> this is an array on php side,
the logging doesn't supported this.

issue: #4558

### DIFF
--- a/Sources/Logging.php
+++ b/Sources/Logging.php
@@ -25,7 +25,13 @@ if (!defined('SMF'))
  */
 function truncateArray($arr, $max_length=1900)
 {
-	$curr_length = array_sum(array_map("strlen", $arr));
+	$curr_length = 0;
+	foreach ($arr as $key => $value)
+		if (is_array($value))
+			foreach ($value as $key2 => $value2)
+				$curr_length += strlen ($value2);
+		else
+			$curr_length += strlen ($value);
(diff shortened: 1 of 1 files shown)


Sesquipedalian (approved):

Sources/Logging.php
```diff
@@ -25,15 +25,25 @@
  */
 function truncateArray($arr, $max_length=1900)
 {
-	$curr_length = array_sum(array_map("strlen", $arr));
+	$curr_length = 0;
+	foreach ($arr as $key => $value)
```
equinoxmatt, line 29: Just a question RE syntax.... why the lack of { }? Is that the SMF coding style?

Sesquipedalian: For one line things, sure.

PR #4562, opened 2018-02-22
Support to search for ipaddress

albertlast:
the search doesn't work without this.
Support the same syntax like ban mask:
- 127.0.0.1
- 127.*
- 127.0.0.1-127.0.0.100

only wildcard at the end supported

Thanks to the person how is to busy to write down this issue in github,
but got the time to write down in a random board.

### DIFF
--- a/Sources/ManageMembers.php
+++ b/Sources/ManageMembers.php
@@ -255,6 +255,14 @@ function ViewMemberlist()
 			'ip' => array(
 				'db_fields' => array('member_ip'),
 				'type' => 'inet'
+			),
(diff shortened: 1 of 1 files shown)


albertlast:
Since this error got claimed by the same nice guy and
the issue is the same file and i'm lazy.

I fixed in this pr also the member and post groups search,
btw i found it confusing that we got memberGROUPS and on the other hand post GROUPS.
So one with space and one without...
grafik

Sesquipedalian (approved):
Nicely done.

PR #4563, opened 2018-02-23
[WIP] Updates credits

Sesquipedalian:
If anyone sees anything amiss in this update, please let me know.

Also, the info in the marketing section is known to be incorrect, but I do not know what it should be. If anyone can fill me in on which names should be in that section, and in which order, that would be very helpful.


colinschoen (approved):
LGTM

illori:
i have proposed some changes on the forum related to this. please do not complete the merge until changes are made.

PR #4564, opened 2018-02-24
Lower all header

albertlast:
only proxy.php i evade because ther is a pr how fix this

Since we got man different style and the rfc say it's should be low.
I changed all header which i found to lower case.

related discussion: https://github.com/SimpleMachines/SMF2.1/pull/4550#issuecomment-368209482


tinoest:
I think due to IE11 it’s best to leave it as Content-Type, ETag etc this could break to much otherwise.

albertlast:
Even google response to ie11 in low:
grafik
so see no issue here

I also notice that for exp. my apache convert this information to Upper case all words:
content-type -> Content-Type
which is fine i guess,
but we should provide the information in the standard definition way,
when the webserver decide to do it different than he can.

Sesquipedalian:
@tinoest, do you have a link explaining your concern regarding IE11?

tinoest:
@Sesquipedalian just recall when IE11 first came out that it liked Content-Type rather than content-type. This was a while ago mind so they might of fixed that issue. http/2 rfc states it must be lowercase. So switch to all lowercase. 

https://tools.ietf.org/html/draft-ietf-httpbis-http2-17#section-8.1.2

PR #4566, opened 2018-02-25
Minify improvements

### LINKED ISSUE #4544: Responsive.css should load last

Sesquipedalian:
- Minimizes ~CSS files loaded before responsive.css~ all CSS files by default
- Improves custMinify in several ways:
  - Now generates distinct minified files for different sets of input files. This allows us to serve up minified files that are tailored for different pages, which means we can freely include whatever files are necessary for a given page, without worrying that we are going to serve unnecessary bloat on other pages as a result.
  - Changes the return value so that it is always an array of relevant file info.
  - When minimizing JavaScript files, tries to preserve the async attribute if possible.

This is a follow-up to #4548 that helps to further fix #4544.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,9 +13,9 @@ Packages/*.zip
 Packages/*/
 attachments/
 /upgrade.php
-Themes/default/css/minified.css
-Themes/default/scripts/minified.js
-Themes/default/scripts/minified_deferred.js
+Themes/default/css/minified*.css
+Themes/default/scripts/minified*.js
+Themes/default/scripts/minified_deferred*.js
 
 # Compiled source #
 ###################
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -2434,14 +2434,13 @@ function loadSubTemplate($sub_template_name, $fatal = false)
 function loadCSSFile($fileName, $params = array(), $id = '')
 {
 	global $settings, $context, $modSettings;
-	
-	if (empty($context['css_files_order'])) 
+
+	if (empty($context['css_files_order']))
 		$context['css_files_order'] = array();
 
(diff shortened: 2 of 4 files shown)


Sesquipedalian:
On further consideration, I've realized that because this is now happening:

> Now generates distinct minified files for different sets of input files. This allows us to serve up minified files that are tailored for different pages, which means we can freely include whatever files are necessary for a given page, without worrying that we are going to serve unnecessary bloat on other pages as a result.

... that there is actually no reason why we shouldn't include all CSS files in the minimized file by default. Since we now create a different minified_[hash].css file for each combination of input files, and minimizing doesn't alter the relevant content of the input files in any way, they can and should all go in.

Importantly, this change means that we no longer need a file named responsive.css to exist in the theme's CSS in order to make the logic work. That's good, because we don't want to force the theme designer to follow any particular naming conventions just to make our software work.

Sesquipedalian:
Would you mind taking a moment to test this, @Gwenwyfar? It seems to work flawlessly on my machines, but I'd like your input, too.

Gwenwyfar:
Seems to be working fine :) I'm not sure I completely understand what you mean with this though:

> this change means that we no longer need a file named responsive.css to exist in the theme's CSS in order to make the logic work. That's good, because we don't want to force the theme designer to follow any particular naming conventions just to make our software work.

Doesn't this mean we can remove the documentation saying the responsive file should be named responsive.css?

Sesquipedalian:
> I'm not sure I completely understand what you mean with this though

The logic for choosing which CSS files to minimize no longer requires a file named responsive.css. The initial version of this PR used the position of responsive.css in the load order as part of its decision-making process. Now it doesn't. Instead it just minimizes everything (except files that explicitly opt out) in order.

> Doesn't this mean we can remove the documentation saying the responsive file should be named responsive.css?

I already [did](https://github.com/SimpleMachines/SMF2.1/commit/ecd980b6883145af4fbdc046bf081a7bd3fe2aff#diff-9e8c5b8ee93aa642bd9f1ff71d59cc31L107). :)

Gwenwyfar:
I see. Looks great then :)

Ops, that's what I get for seeing the last few commits backwards, thought the "responsive.css" documentation had just been moved about :x

Gwenwyfar (approved):

PR #4568, opened 2018-02-26
Catch Warning from mail()

albertlast:
By testing another error,
my test env remember me that i got no mail server setting up:
grafik
In my eyes to mutch information to a $random guy,
so i suppress the error message on user interface side but log this in the smf log.

### DIFF
--- a/Sources/Subs-Post.php
+++ b/Sources/Subs-Post.php
@@ -686,11 +686,29 @@ function sendmail($to, $subject, $message, $from = null, $message_id = null, $se
 
 		foreach ($to_array as $to)
 		{
-			if (!mail(strtr($to, array("\r" => '', "\n" => '')), $subject, $message, $headers))
+			set_error_handler(function($errno, $errstr, $errfile, $errline)
+				{
+					// error was suppressed with the @-operator
+					if (0 === error_reporting()) {
+						return false;
+					}
+
+					throw new ErrorException($errstr, 0, $errno, $errfile, $errline);
+				}
+			);
+			try
 			{
-				log_error(sprintf($txt['mail_send_unable'], $to));
-				$mail_result = false;
+				if (!mail(strtr($to, array("\r" => '', "\n" => '')), $subject, $message, $headers))
+				{
+					log_error(sprintf($txt['mail_send_unable'], $to));
+					$mail_result = false;
+				}
 			}
+			catch(ErrorException $e)
+			{
+				log_error($e->getMessage(), 'general', $e->getFile(), $e->getLine());
+			}
+			restore_error_handler();
 
 			// Wait, wait, I'm still sending here!
 			@set_time_limit(300);
--- a/Sources/Subs-Post.php
+++ b/Sources/Subs-Post.php
@@ -707,6 +707,8 @@ function sendmail($to, $subject, $message, $from = null, $message_id = null, $se
 			catch(ErrorException $e)
 			{
 				log_error($e->getMessage(), 'general', $e->getFile(), $e->getLine());
+				log_error(sprintf($txt['mail_send_unable'], $to));
+				$mail_result = false;
 			}
 			restore_error_handler();
 


equinoxmatt:
I really like the idea and I agree it shouldn't be blasted out for all to see.

However, setting an error handler just for once function seems overkill.  Can we either look at setting the error handler globally or doing this another way?

Not familiar with how errors are generally handled across the codebase, however this concept could be used elsewhere.

albertlast:
smf use a error handler how did his job good,
but in this case is that not error, it's warning which get printed in the commone cases.
But in this case i would like that this not happen,
so i setup a special error handler for this spot.

equinoxmatt:
Yep, I see what you were doing and I agree. I just wonder if this can be made reusable if it is needed it elsewhere

Sesquipedalian:
> I just wonder if this can be made reusable if it is needed it elsewhere

SMF rarely ever uses exceptions, so there isn't much opportunity to reuse something like this. 

EDIT: If you are curious, @equinoxmatt, you can check out SMF's custom error handler in the `smf_error_handler()` function in Errors.php.

equinoxmatt:
Fair enough :)

Will take a look 👍

albertlast:
This code do something differently,
It change warning to exception.
So that we can suppress this.

equinoxmatt:
Could we suppress the warning with an @ operator check the result and then log if false using error_get_last()?

I don't know, I like the idea, it's just setting/unsetting the error handler for one function seems too much.

Ignore me if I am being to awkward, get it merged in :)

albertlast:
When you do @ in front of,
than only one error get logged compare yellow vs not yellow area:
grafik

Sesquipedalian:
I definitely prefer giving the admin some information on why the email could not be sent.

jdarwood007:
Whats the performance impact of setting and restoring a anonymous error handler a lot of times?  Thinking more here along the lines of newsletters.

Sesquipedalian:
That's a good point, @jdarwood007. We should move the `set_error_handler()` and `restore_error_handler()` outside the `foreach()` loop. Would you mind tossing together a quick PR to do that, @albertlast?

albertlast:
Because this question i build this benchmark:
```
<?php
$db_string = 1;
$time_old = 0;
$time_new = 0;
$temp = false;
$max = 1000;
$sleep = 25000; // 1k = 1ms

set_time_limit(600);

$start = microtime(true);
set_error_handler(function($errno, $errstr, $errfile, $errline)
{
	// error was suppressed with the @-operator
	if (0 === error_reporting()) {
		return false;
	}

	throw new ErrorException($errstr, 0, $errno, $errfile, $errline);
}
);
for($i = 0; $i < $max; $i++)
{
	try
	{
		usleep($sleep);
	}
	catch(ErrorException $e)
	{
		log_error($e->getMessage(), 'general', $e->getFile(), $e->getLine());
		log_error(sprintf($txt['mail_send_unable'], $to));
		$mail_result = false;
	}
}
restore_error_handler();			

$start = microtime(true);
for($i = 0; $i < $max; $i++)
{
			set_error_handler(function($errno, $errstr, $errfile, $errline)
				{
					// error was suppressed with the @-operator
					if (0 === error_reporting()) {
						return false;
					}

					throw new ErrorException($errstr, 0, $errno, $errfile, $errline);
				}
			);
			try
			{
				usleep($sleep);
			}
			catch(ErrorException $e)
 			{
				log_error($e->getMessage(), 'general', $e->getFile(), $e->getLine());
 				log_error(sprintf($txt['mail_send_unable'], $to));
 				$mail_result = false;
 			}
			restore_error_handler();
}
	
var_dump($time_old);
var_dump($time_new);
```
The sleep is the mail() call.
Result:
test.php:66:float 31.262226104736 (this is the once logic)
test.php:67:float 31.256489038467 (this is do it in every loop)

Because i see no differenc here and the possibly that also other code react than differently,
which maybe is not wanted, i see no need for this change.

Sesquipedalian:
Thanks, @albertlast. There's nothing like some good numbers to settle a question. Based on those results, I agree that there's no need for a change after all.

jdarwood007:
Yup fine with that.  Was just ensuring we where not going to do something performance killing especially on newsletter sending

PR #4569, opened 2018-02-26
Check $cur_profile['email_address'] before used

albertlast:
Well to create this issue you had to activate "send_validation_onChange"
but even than i was not fully able to reproduce this error but this should solve this any way.

related issue: https://www.simplemachines.org/community/index.php?topic=559067.0

### DIFF
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -167,7 +167,7 @@ function loadProfileFields($force_reload = false)
 			'js_submit' => !empty($modSettings['send_validation_onChange']) ? '
 	form_handle.addEventListener(\'submit\', function(event)
 	{
-		if (this.email_address.value != "'. $cur_profile['email_address'] . '")
+		if (this.email_address.value != "'. (!empty($cur_profile['email_address']) ? !empty($cur_profile['email_address']) : '') . '")
 		{
 			alert('. JavaScriptEscape($txt['email_change_logout']) . ');
 			return true;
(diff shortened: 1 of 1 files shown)


Sources/Profile-Modify.php
```diff
@@ -167,7 +167,7 @@ function $preload:		A function that is used to load data required for this eleme
 			'js_submit' => !empty($modSettings['send_validation_onChange']) ? '
 	form_handle.addEventListener(\'submit\', function(event)
 	{
-		if (this.email_address.value != "'. $cur_profile['email_address'] . '")
+		if (this.email_address.value != "'. (!empty($cur_profile['email_address']) ? !empty($cur_profile['email_address']) : '') . '")
```
Sesquipedalian: I think you probably meant to write this:
```
(!empty($cur_profile['email_address']) ? $cur_profile['email_address'] : '')
```

Sesquipedalian (changes requested):
Copy/paste error?

Sesquipedalian (approved):

jdarwood007:
Whats the error here?  email is a requirement of all profiles, so not sure how you are getting a blank email to trigger this.

smfbigguy:
People were trying to register on my site. I could see them. When they did I would get the error I posted on SMF. After I applied the fix I was able to register perfectly. I have seen this happening on my site for about 4 or 5 days. Numerous people generated that error. That is all I can tell you really.

albertlast:
In this moment when the user accept the licence(in the regestry process) the profile stuff is builded up,
but at this moment the user didn't provide any mail information.

PR #4570, opened 2018-02-26
Docker setup

equinoxmatt:
Apache, PHP 5.4  (with xdebug), MySql

I choose PHP 5.4 as this is the minimum version for SMF, so we should be testing everything on PHP 5.4.

Does someone want to give it a go?

Instructions:

docker-compose up -d
docker-compose logs -f (if you want to see the build logs)

Copy files from other folder (we could do this as part of the setup).
http://localhost:8080/install.php
In the DB settings, you can use the database container name (smf21_database_1).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM php:5.4-apache
+RUN apt-get update
+RUN apt-get install -y zip libxml2-dev libpng-dev
+RUN docker-php-ext-install mysql simplexml mbstring gd mysqli
+RUN pecl install xdebug-2.4.0 && docker-php-ext-enable xdebug
+RUN echo "xdebug.remote_enable=on" >> /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini
+RUN echo "xdebug.remote_connect_back=on" >> /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini
+# nasty hack to solve perms issues
+RUN chmod 777 /var/www/html
+
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,15 @@
+version: '2'
+services:
+  smf:
+    build: .
+    ports:
+      - "8080:80"
+    volumes:
+      - ./:/var/www/html
+  database:
+    image: mysql
+    restart: always
+    ports:
+      - "3306:3306"
+    environment:
+      MYSQL_ROOT_PASSWORD: password


Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM php:5.4-apache
+RUN apt-get update
+RUN apt-get install -y zip libxml2-dev libpng-dev
+RUN docker-php-ext-install mysql simplexml mbstring gd mysqli
+RUN pecl install xdebug-2.4.0 && docker-php-ext-enable xdebug
+RUN echo "xdebug.remote_enable=on" >> /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini
+RUN echo "xdebug.remote_connect_back=on" >> /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini
+# nasty hack to solve perms issues
+RUN chmod 777 /var/www/html
```
colinschoen, line 9: This should be more specific to only directories that need to be writable (e.g. attachments)

colinschoen (changes requested):

Dockerfile
equinoxmatt: Absolutely, I agree.  Docker didn't like it when I tried to do it that way.  Before this gets merged (if accepted), I will fix this. I just need to find the solution.

My first priority is to get someone else to test this.

I work on Linux & mac. This was tested on Linux (ubuntu), going to try on my macbook tomorrow.

docker-compose.yml
```diff
@@ -0,0 +1,15 @@
...
+      - "8080:80"
+    volumes:
+      - ./:/var/www/html
+  database:
+    image: mysql
+    restart: always
+    ports:
+      - "3306:3306"
+    environment:
+      MYSQL_ROOT_PASSWORD: password
```
jdarwood007, line 15: That's secure looking!

Dockerfile
Sesquipedalian: Also, `chmod 775` should be sufficient.

docker-compose.yml
equinoxmatt: This is meant for development, nothing else.

If we want to make this ready for production, someone else with more docker experience can take this over.

equinoxmatt:
Just tried this on the macbook, worked fine.  Full environment up and running within 5 minutes and one command 👍

Sesquipedalian:
I'm open to discuss it, but as I think about it, I'm not sure that including this in the 2.1 repo is warranted. Docker seems pretty useful, but it isn't part of our shared development toolset. 

Of course, if you want to use Docker to deploy your test installs, @equinoxmatt, by all means do so. :) But rather than committing these files to the repo, it might be better to add their names to .gitignore or even your local .git/info/exclude.

equinoxmatt:
No problem, it was more to help other devs than anything else.  It is really useful for those that don't want to have multiple PHP's versions running natively. e.g I run php5./php7 for most projects, being able to run SMF separate from everything else is very useful.  Feel free to share the files on your internal developer forum. 👍

Sesquipedalian:
Nice. I now plan to look into using Docker myself. Thanks for putting me onto it, @equinoxmatt.

jdarwood007:
Could include it in the buildTools repo for people who would fine it useful.  Could have the script do a git clone of the current revision from git and then do the rest.

Sesquipedalian:
That sounds like a good idea.

PR #4578, opened 2018-03-03
Check for not empty profile field fixes #4523

### LINKED ISSUE #4523: Profile summary Custom fields template creates empty gabs/elements
example:
https://github.com/SimpleMachines/SMF2.1/blob/ac85d8c0fb1b070aaabfe622cf02839682c7eafb/Themes/default/Profile.template.php#L157

then this line checks if there are any, (there are always ones if you make one, but when the output
`if (!empty($field['output_html']))`
 is not entered/filled in by the profile holder then this check still makes the div?)
`if (!empty($context['print_custom_fields']['above_member']))`
This is on all costum fields within that summary template.

there needs to be more checks otherwise the profile summary template screwes up with empty gabs/elements.

possible fix:
```
// Are there any custom profile fields for above the name?
	if (!empty($context['print_custom_fields']['above_member']))
	{
		$fields = '';
		foreach ($context['print_custom_fields']['above_member'] as $field)
		{
			if (!empty($field['output_html']))
				$fields = $fields.'<li>'. $field['output_html']. '</li>';
		}
		if ($fields)
			echo'<div class="custom_fields_above_name"><ul>', $fields, '</ul></div><br>';
	}
```

albertlast:
My target of this pr was to keep the reading for person easy how do gui stuff,
reason why i splitt the vars to keep the indentation easy.
issue: #4523

### DIFF
--- a/Themes/default/Profile.template.php
+++ b/Themes/default/Profile.template.php
@@ -157,19 +157,26 @@ function template_summary()
 	// Are there any custom profile fields for above the name?
 	if (!empty($context['print_custom_fields']['above_member']))
 	{
-		echo '
+		$above = '
 			<div class="custom_fields_above_name">
 				<ul>';
 
+		$fields = '';
 		foreach ($context['print_custom_fields']['above_member'] as $field)
 			if (!empty($field['output_html']))
-				echo '
+				$fields .= '
 					<li>', $field['output_html'], '</li>';
 
-		echo '
+		if (!empty($fields))
+		{
+			echo $above;
+			echo $fields;
+			echo '
 				</ul>
 			</div>
 			<br>';
+		}
+
 	}
 
 	echo '
--- a/Themes/default/Profile.template.php
+++ b/Themes/default/Profile.template.php
@@ -157,10 +157,6 @@ function template_summary()
 	// Are there any custom profile fields for above the name?
 	if (!empty($context['print_custom_fields']['above_member']))
 	{
-		$above = '
-			<div class="custom_fields_above_name">
-				<ul>';
-
 		$fields = '';
 		foreach ($context['print_custom_fields']['above_member'] as $field)
 			if (!empty($field['output_html']))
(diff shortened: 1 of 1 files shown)


Themes/default/Profile.template.php
```diff
@@ -157,19 +157,26 @@ function template_summary()
 	// Are there any custom profile fields for above the name?
 	if (!empty($context['print_custom_fields']['above_member']))
 	{
-		echo '
+		$above = '
```
Sesquipedalian: Instead of defining $above here and then echoing it later, it'd be cleaner to move the original echo statement down to where you've put `echo $above;`

Sesquipedalian (changes requested):
One minor change requested. Other than that, LGTM.

albertlast:
I know but the reading for someone who is not a dev is so I guess better,
Because the order and the spaceing are easier to read.


Gesendet: Montag, 5. März 2018 04:50
An: SimpleMachines/SMF2.1
Cc: albertlast; Author
Betreff: Re: [SimpleMachines/SMF2.1] Check for not empty profile field fixes #4523 (#4578)


@Sesquipedalian requested changes on this pull request.

One minor change requested. Other than that, LGTM.

In Themes/default/Profile.template.php<https://github.com/SimpleMachines/SMF2.1/pull/4578#discussion_r172085898>:

> @@ -157,19 +157,26 @@ function template_summary()
        // Are there any custom profile fields for above the name?
        if (!empty($context['print_custom_fields']['above_member']))
        {
-               echo '
+               $above = '


Instead of defining $above here and then echoing it later, it'd be cleaner to move the original echo statement down to where you've put echo $above;

Sesquipedalian:
Maybe. Personally, I find it easier to think about it the other way: one part doing the setup, and one part doing the echoing.

albertlast:
I mean the existing code define the html code in this order in which is displayed,
In your mention way you define first the middle party and then the first and then the end part. 

But to be clear I don't care, 
I only want clear up the intention of the existing code. 
So when you like I can change this.

Sesquipedalian:
Yes, please do. :)

albertlast:
done

Sesquipedalian (approved):

PR #4580, opened 2018-03-03
Delete custom fields entries when empty fixes #4579

### LINKED ISSUE #4579: Empty custom fields display on memberlist
From https://www.simplemachines.org/community/index.php?topic=554449.0:

> If a member has a custom field with icons filled out and they appear on the memberlist, the icon is there as it should be, however, if the custom field is left empty, the memberlist still shows everything in the "Enclosed Within Text" section of said custom field but the {INPUT} part.

albertlast:
When you empty a custom field the theme entry should be deleted.

issue: #4579

### DIFF
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -1167,6 +1167,7 @@ function makeCustomFieldChanges($memID, $area, $sanitize = true, $returnErrors =
 		)
 	);
 	$changes = array();
+	$deletes = array();
 	$log_changes = array();
 	while ($row = $smcFunc['db_fetch_assoc']($request))
 	{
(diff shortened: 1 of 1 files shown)


Gwenwyfar:
Tested and fixes the issue upon editing and for any new fields created. But existing fields continue displaying when empty unless you edit them twice, should anything be done about that?

albertlast:
Maybe the upgrader when the issue also exists in smf2.0.x,
on a clean 2.1 setup shouldn't happen,

Sources/Profile-Modify.php
```diff
@@ -1253,26 +1254,43 @@ function makeCustomFieldChanges($memID, $area, $sanitize = true, $returnErrors =
 					'member_affected' => $memID,
 				),
 			);
-			$changes[] = array(1, $row['col_name'], $value, $memID);
-			$user_profile[$memID]['options'][$row['col_name']] = $value;
+			if (empty($value))
+			{
+				$deletes = array('id_theme' => 1 , 'variable' => $row['col_name'], 'id_member' => $memID);
```
Sesquipedalian, line 1259: I think this will only ever delete one row per save, won't it? This should probably use `$deletes[]` here, and then use a more advanced query below to delete all variables.

Sesquipedalian (changes requested):

Sources/Profile-Modify.php
albertlast: Nope makeCustomFieldChanges() is called once for every field.

Sesquipedalian: Are you sure about that? The [query](https://github.com/SimpleMachines/SMF2.1/pull/4580/files#diff-73d40d702b0964f8eba5d0e9d7ea31b1R1159) retrieves all the fields in the profile area and then the [while](https://github.com/SimpleMachines/SMF2.1/pull/4580/files#diff-73d40d702b0964f8eba5d0e9d7ea31b1R1172) loops through them. Thats why `$changes[]` is used [here](https://github.com/SimpleMachines/SMF2.1/pull/4580/files#diff-73d40d702b0964f8eba5d0e9d7ea31b1R1264) rather than just `$changes`.

albertlast: Yes this query call only the >settings< for every fields.

Maybe because a field could got multipley values.
But this code delete the complet entry when value is empty.

Sesquipedalian (approved):

PR #4593, opened 2018-03-08
More proxy.php changes

Sesquipedalian:
Follows up on #4550.


Sesquipedalian:
Moving fetch_web_data() into Subs.php makes this PR look much bigger than it really is. All the actually significant changes are just in proxy.php.

Sesquipedalian:
@sbulen, would you mind testing this version of proxy.php to see if it works for you?

proxy.php
```diff
@@ -50,7 +47,7 @@ public function __construct()
 		global $image_proxy_enabled, $image_proxy_maxsize, $image_proxy_secret, $cachedir, $sourcedir;
 
 		require_once(dirname(__FILE__) . '/Settings.php');
-		require_once($sourcedir . '/Class-CurlFetchWeb.php');
+		require_once($sourcedir . '/Subs.php');
```
albertlast, line 50: I like to point out,
that the Subs,php is garbage for the default behavior,
only in the special case when curl is not avaible (which is rar) the subs are needed,
for this load on every call this garbage (~5k lines) is useless.

proxy.php
```diff
@@ -78,7 +75,11 @@ public function checkRequest()
 			if (!mkdir($this->cache) || !copy(dirname($this->cache) . '/index.php', $this->cache . '/index.php'))
 				return false;
 
-		if (empty($_GET['hash']) || empty($_GET['request']) || ($_GET['request'] === "http:") || ($_GET['request'] === "https:"))
+		// Basic sanity check
+		$_GET['request'] = filter_var($_GET['request'], FILTER_VALIDATE_URL);
```
albertlast, line 79: When the url(the url was checked by smf) and hash match each other,
i see no need to validate again the url.

albertlast:
I would like to see two kinds of benchmark,
- runtime of cached content (compare existing solutions and your solutions)
- runtime of caching new content (again existing vs your pr)

sbulen:
@sesquipedalian - will do.  Tied up until Sunday, but will get to it then.   Have other old defects to retest also.

albertlast:
So i did some bechmark,
placed a mod_rewrite to img2/* test.jpg
created 1k url with hashes
called jmeter 1k single thread every image once

sproxy = sesqu implementation
oproxy = orginal
alt = without curl

grafik
numbers in ms

from what i see is the non curl version from this pr the fastes solution what we got.

reading of caching values are no diff.

proxy.php
Sesquipedalian: If `$_GET['request']` can be just `http:`, then clearly the URL is not guaranteed to be valid.

Sesquipedalian:
That is surprising and fascinating. Thanks for that info, @albertlast. This makes me wonder if we should get rid of the curl method entirely. We'd need to do more testing before making any such decision, of course, but it is worth considering.

tinoest:
@sesquipedalian some hosts disable fsockopen, which I think the alternative is using having a quick look at the code. So curl should be there still.

Sesquipedalian:
Hm, I hadn't thought of that. Still, perhaps we should change the order in which we try things.

sbulen:
Changing the order is an excellent idea.   

Excellent analysis, @albertlast .

As tinoest noted, we will find every possible combination of things that are disabled (or misconfigured) out there.  This is why SMF code is peppered with backups - fetch_web_data() is a classic example of that.  It evolved based on issues encountered - in a good way.  Which relates to my notes above...  We need to be a defensive as remotely possible (and try to paint the pic no matter what), since our code will encounter every situation possible out there...

PR #4596, opened 2018-03-08
Move post images size limit to css

### LINKED ISSUE #1401: Max image size during posting
There are several issues with this.

Firstly: overly large images may never get processed to have their size read.
http://www.simplemachines.org/community/index.php?topic=505111.0
http://www.simplemachines.org/community/index.php?topic=505050.msg3555178#msg3555178

Secondly: much as it's still not _quite_ a bug that it's not behaving 'as expected' (it IS behaving as designed, but the design is from some years before CSS had things like max-width)
http://www.simplemachines.org/community/index.php?topic=518083.0

Basically, the use of size should probably be generally dropped from SMF's bbc. There's the legacy thing to consider, but forcing the size in the code just isn't going to work out so well.

Now, the resized image details (as to whether the full vs resized toggle is necessary) can be determined via the naturalWidth and naturalHeight properties of said image, which are accessible in all good HTML5 browsers. IE8 of course needs help, http://www.jacklmoore.com/notes/naturalwidth-and-naturalheight-in-ie/

Thirdly, there's no way to prevent over-sized images, e.g. for mobile users. There may not be a good way to solve that.

Gwenwyfar:
Fixes the host of issues with the old resize function. This does create one small conflict with the existing code: images that get resized through css will no longer have the "resized" class, so the "click to expand" function only works with manually resized images instead. I will make a separate PR to add a popup to expand images inside posts and perhaps for attachments as well.

The css removed fixes one additional issue with any manually defined height being ignored.

Fixes #1401

### DIFF
--- a/Sources/Subs-Post.php
+++ b/Sources/Subs-Post.php
@@ -362,66 +362,6 @@ function fixTags(&$message)
 		return "$m[1]" . preg_replace("~action(=|%3d)(?!dlattach)~i", "action-", "$m[2]") . "[/img]";
 	}, $message);
 
-	// Limit the size of images posted?
-	if (!empty($modSettings['max_image_width']) || !empty($modSettings['max_image_height']))
-	{
-		// Find all the img tags - with or without width and height.
-		preg_match_all('~\[img(\s+width=\d+)?(\s+height=\d+)?(\s+width=\d+)?\](.+?)\[/img\]~is', $message, $matches, PREG_PATTERN_ORDER);
-
-		$replaces = array();
-		foreach ($matches[0] as $match => $dummy)
-		{
-			// If the width was after the height, handle it.
-			$matches[1][$match] = !empty($matches[3][$match]) ? $matches[3][$match] : $matches[1][$match];
-
-			// Now figure out if they had a desired height or width...
-			$desired_width = !empty($matches[1][$match]) ? (int) substr(trim($matches[1][$match]), 6) : 0;
-			$desired_height = !empty($matches[2][$match]) ? (int) substr(trim($matches[2][$match]), 7) : 0;
-
-			// One was omitted, or both.  We'll have to find its real size...
-			if (empty($desired_width) || empty($desired_height))
-			{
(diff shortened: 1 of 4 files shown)


Sesquipedalian:
I haven't tested yet, but it looks good at first glance.

> I will make a separate PR to add a popup to expand images inside posts and perhaps for attachments as well.

I wouldn't make it a popup. Just a simple `onclick` on the image to toggle the `resized` class on or off should do the job.

Sesquipedalian:
Something like `$(this).toggleClass('resized');` should do it.

Gwenwyfar:
I think a popup will make the resize more useful than it is now, and we can include the source link in it. For the attachments it would replace the new window it opens now. A resize is constrained to the post area, in a popup we can use the entire screen for the larger images. Would actually be even better if we can cycle through images in it, but I'm leaving this as a nice extra if time allows, so I'll just go with your idea until then ;)

Sesquipedalian:
Now I understand what you mean. That would be nice, yes. But it also sounds an awful lot like _adding a new feature_ when we are trying to get RC1 out the door. 😉

Sesquipedalian:
Anyway, it occurs to me that instead of adding an `onclick` to every image, it'd be better to just add something like this to script.js:
```
$('.postarea .bbc_img').each(function(index, item)
{
	$(item).click(function(e)
	{
		$(item).toggleClass('resized');
	});
});
```
Note: untested.

Gwenwyfar:
We already have a popups function that we use elsewhere, so it would just be a matter of replacing the resize/new window with them, but yes, it's a bit of work to figure how they work and actually get it done :P

Well, here's the updated resizing. I'll need to think over if we need to remove the existing resized class or if it has any use for the css. I'll remove it on my next revision PR if not.

XinYenFon:
You can take a peek at how we serve the mobile menu, you can do very much the same with special class which allows you to customize the pop section in the end you don't require backgrounds etc... while serving full screen images.

But I think the tricky (or at least for me) part is to make sure you don't load every image in pop section as well, I think with proper JS you can take the image and show it in the pop screen (I kinda confused myself in here but I hope you get the idea).

PR #4599, opened 2018-03-09
External proxy doesn't like urlencode

albertlast:
When you look at the changes from https://www.simplemachines.org/community/index.php?topic=558946.msg3962443#msg3962443
and play with the example from https://github.com/willnorris/imageproxy
you notice that the proxy doesn't like urlencode

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -447,7 +447,7 @@ function loadUserSettings()
 
 			if (!empty($modSettings['force_ssl']) && $image_proxy_enabled && stripos($user_settings['avatar'], 'http://') !== false && empty($user_info['possibly_robot']))
 				if ($image_proxy_enabled === 2 && !empty($image_proxy_url))
-					$user_settings['avatar'] = $image_proxy_url . urlencode($user_settings['avatar']);
+					$user_settings['avatar'] = $image_proxy_url . $user_settings['avatar'];
 				else
 					$user_settings['avatar'] = strtr($boardurl, array('http://' => 'https://')) . '/proxy.php?request=' . urlencode($user_settings['avatar']) . '&hash=' . md5($user_settings['avatar'] . $image_proxy_secret);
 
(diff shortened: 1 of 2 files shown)


Sesquipedalian:
The behaviour of that particular proxy script is unusual. The standard behaviour for such scripts is to take the URL as a query parameter.

We probably need to make our code for the external proxy field smarter. Besides whether to urlencode or not, we also need a way to allow arbitrary proxy URL structures. For example, some scripts might require stuff to be appended after the requested URL.

As well, we ought to add some hooks to make it very easy for mods to alter the external proxy behaviour.

I think I have an idea for how to accomplish both of these goals.

albertlast:
From where you take the knowlegd that is common to use urlencode for image proxy?
can you show exp.(smf code doesn't count) ?

Sesquipedalian:
> From where you take the knowlegd that is common to use urlencode for image proxy?

From the fact that `http://localhost/200/https://willnorris.com/logo.jpg` isn't a syntactically valid URL, which is why the software you linked to needs to perform non-standard parsing of the path in order to extract the requested URL from it. Native methods for doing that don't exist, because that way of passing the requested URL is just wrong.

> can you show exp.(smf code doesn't count) ?

There are [many](https://github.com/search?utf8=✓&q=image+proxy+language%3APHP&type=Repositories&ref=advsearch&l=PHP&l=).

albertlast:
All this projects are mutch smaller than this solution which i linked and
the nginx works in the same way like the shown project.
So from what i see,
is that the standard to use the normal url and not the urlencoded version.

Sesquipedalian:
> All this projects are mutch smaller than this solution which i linked

Even if that were true, it wouldn't change anything. https://yourlogicalfallacyis.com/special-pleading

But moreover, as I look at the variety of ways that different proxy scripts work, I have realized that there is no way for us to build any sort of generic support for them, and it is a mistake to try to do so. Providing support for other proxies should be handled via mods.

albertlast:
I know that everyone do it differently,
but in my eyes this are two big parts of the market which we can serve by this easy solution,
(this was idea from the start less amount of work for max outcome).

We could centralize the code where mods could hook up ther logic.
Exp. this code:
```
$user_settings['avatar'] = $image_proxy_url . $user_settings['avatar'];
```
get to
```
$user_settings['avatar'] = $image_proxy_url . external_proxy($user_settings['avatar']);
```

The external_proxy function is default implement with empty logic to support this big kinds of image proxy,
but we could place a hook where mods could add the different other logic.

```
function external_proxy($url)
{
hook(&$url);
return $url;
}
```

with this approache is see everything option posible,
could implement in this way for exp. the internal proxy
the hook logic needs to in this way than:
```
hook($url)
{
$url = '?request=' . urlencode($url) . '&hash=' . md5($url.$salt);
}
```

albertlast:
@tinoest i notice you got a mod for external proxy,
would be this implementation help you?

tinoest:
I did it a slightly different way, it was primarily only to test if the issues were resolved for that person in the bug report. 

Having a hook means I don’t have to parse the entire page. So it’s less resource intensive but it wouldn’t solve the reason I went down the route I did.

Extending the proxy is one way and I think it should have a hook to do so, but it’s not the only way to achieve the same results, parsing the page as I did in the quick and simple modification also works.

Sesquipedalian:
I like your idea here, @albertlast. I'm incorporating it into a PR that will, I hope, provide an easier and more robust way to handle all the image proxy code, including both the internal proxy and any external ones.

PR #4608, opened 2018-03-10
Consolidates & simplifies process of getting proxy urls; allows mods to add alternative proxies

Sesquipedalian:
The `get_proxied_url()` function will return the correct URL to use for URLs that may or may not require a proxy. If the proxy is disabled, we are not using SSL, or the requested URL is already an HTTPS url, then the original URL is returned. Otherwise, the proxied URL is returned.

Mods can easily implement any sort of alternative proxy they want via the 'integrate_proxy' hook. That hooks simply passes the `$url` and `$proxied_url` variables so that the mod can overwrite `$proxied_url` with whatever it wants based on `$url`.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -2875,7 +2875,7 @@ function highlight_php_code($code)
  */
 function get_proxied_url($url)
 {
-	global $boardurl, $modSettings, $image_proxy_enabled, $image_proxy_secret;
+	global $boardurl, $image_proxy_enabled, $image_proxy_secret;
 
 	// Only use the proxy if enabled and necessary
 	if (empty($image_proxy_enabled) || empty($modSettings['force_ssl']) || parse_url($url, PHP_URL_SCHEME) === 'https')
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -1413,7 +1413,7 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 				'content' => '<img src="$1" alt="{alt}" title="{title}"{width}{height} class="bbc_img resized">',
 				'validate' => function (&$tag, &$data, $disabled)
 				{
-					global $image_proxy_enabled, $image_proxy_secret, $boardurl, $user_info;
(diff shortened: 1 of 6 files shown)


Sesquipedalian:
Okay, I think this is done. 
@sbulen, @albertlast, @tinoest: Any feedback on this approach?

albertlast:
Well you added the javascript bug again in the proxy selection...
was placed here: https://github.com/SimpleMachines/SMF2.1/pull/4608/commits/43a38abd68533b9c508fcdec4f540f93652d692a

Sesquipedalian:
00fab2f re-implemented your fix for that. :)

Sources/Subs.php
```diff
@@ -2870,6 +2864,37 @@ function highlight_php_code($code)
...
+ *
+ * @param string $url The original URL of the requested resource
+ * @return string The URL to use
+ */
+function get_proxied_url($url)
+{
+	global $boardurl, $modSettings, $image_proxy_enabled, $image_proxy_secret;
+
+	// Only use the proxy if enabled and necessary
+	if (empty($image_proxy_enabled) || empty($modSettings['force_ssl']) || parse_url($url, PHP_URL_SCHEME) === 'https')
```
albertlast: In the load php is already a check for force_ssl,
when disable it set $image_proxy_enabled at false:
https://github.com/SimpleMachines/SMF2.1/blob/1bbcf2bf43d82a61cb1bb94533e8b15a0a5247f4/Sources/Load.php#L1759-L1761
no need for double check empty($modSettings['force_ssl'])

albertlast:
well the js got two bugs,
one you found,
the second is the missing true.

albertlast:
Well i got mixed feelings about to get back to two modes (true or false) for image_proxy,
the "existing" solution got three modes (off/internal/external).
But beside of the posibility to skip the own code i couldn't describe this in words...

maybe my issue that we provide no external proxy support by default,
even the easiest implementation is now gone.

Sources/Subs.php
Sesquipedalian: Good catch.

PR #4612, opened 2018-03-11
Adds better support for IRIs (international URLs)

Sesquipedalian:
1. Adds a Punycode library and several functions to make using it easy.
2. Uses those functions to add IRI-compatible URL validation at several places throughout the codebase, including parse_bbc()'s autolinker, the image proxy, and the registration form.


equinoxmatt (approved):
LGTM :+1:

PR #4614, opened 2018-03-11
Allow ip range in ip tracking

albertlast:
Allow to search for ip range and
drope the range check for whois provider (get to complex with ipv6...)

### DIFF
--- a/Sources/Profile-View.php
+++ b/Sources/Profile-View.php
@@ -1893,7 +1893,7 @@ function TrackIP($memID = 0)
 	if (isset($_REQUEST['searchip']))
 		$context['ip'] = ip2range(trim($_REQUEST['searchip']));
 
-	if (count($context['ip']) != 2)
+	if (count($context['ip']) !== 2)
 		fatal_lang_error('invalid_tracking_ip', false);
 
 	$ip_string = array('{inet:ip_address_low}','{inet:ip_address_high}');
@@ -1904,7 +1904,7 @@ function TrackIP($memID = 0)
 
 	$ip_var = $context['ip'];
 
-	if ($context['ip']['low'] !=  $context['ip']['high'])
+	if ($context['ip']['low'] !==  $context['ip']['high'])
 		$context['ip'] = $context['ip']['low'] . ' - ' . $context['ip']['high'];
 	else
 		$context['ip'] = $context['ip']['low'];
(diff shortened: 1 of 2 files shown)


Sources/Profile-View.php
```diff
@@ -1885,33 +1885,38 @@ function TrackIP($memID = 0)
...
 		$context['base_url'] = $scripturl . '?action=profile;area=tracking;sa=ip;u=' . $memID;
 	}
 
 	// Searching?
 	if (isset($_REQUEST['searchip']))
-		$context['ip'] = trim($_REQUEST['searchip']);
+		$context['ip'] = ip2range(trim($_REQUEST['searchip']));
 
-	if (isValidIP($context['ip']) === false)
+	if (count($context['ip']) != 2)
```
equinoxmatt: !==

Sources/Profile-View.php
```diff
@@ -1885,33 +1885,38 @@ function TrackIP($memID = 0)
...
+	$ip_string = array('{inet:ip_address_low}','{inet:ip_address_high}');
+	$fields = array(
+			'ip_address_low' => $context['ip']['low'],
+			'ip_address_high' => $context['ip']['high'],
+		);
+
 	$ip_var = $context['ip'];
-	$ip_string = '= {inet:ip_address}';
+
+	if ($context['ip']['low'] !=  $context['ip']['high'])
```
equinoxmatt: !==

Sources/Profile-View.php
albertlast: Can you explain?

Sources/Profile-View.php
albertlast: Can you explain?

equinoxmatt: Sorry, in this context, both of your checks can be done using strict type comparision. Good practice.

Sources/Profile-View.php
equinoxmatt: See below, but basically you are comparing an integer to an integer, therefore you can use strict type comparision (!==/===) (good practice).

albertlast: But which issue solve this?

equinoxmatt: It doesn't solve any issue as such. It is good practice to use strict type comparision, unless there is a specific reason you cannot do so.

equinoxmatt (approved):

colinschoen (approved):
LGTM

PR #4615, opened 2018-03-11
let's call the function of db_allow_persistent

albertlast:
stupid php can return mostly everything here...
> Note: When querying boolean values
> 
> A boolean ini value of off will be returned as an empty string or "0" while a boolean ini value of on will be returned as "1". The function can also return the literal string of INI value.


sbulen:
I just tested this & it works now.   Thanks, al.

Sesquipedalian (approved):

PR #4629, opened 2018-03-15
Templates revision - part 3

Gwenwyfar:
Let's get this out before it grows into a little monster. There aren't many things left that will involve editing the templates from here onwards.

### DIFF
--- a/Themes/default/css/calendar.css
+++ b/Themes/default/css/calendar.css
@@ -265,10 +265,6 @@ span.hidelink {
 	-ms-hyphens: auto;
 	hyphens: auto;
 	overflow-wrap: break-word;
-	word-wrap: break-word;
-}
-#chrome #main_grid td.days {
-	word-break: break-word;
 }
 
 /* At less than 1024px wide, #main_grid needs all the space it can get */
--- a/Themes/default/css/calendar.css
+++ b/Themes/default/css/calendar.css
@@ -264,6 +264,7 @@ span.hidelink {
 	-webkit-hyphens: auto;
 	-ms-hyphens: auto;
 	hyphens: auto;
+	word-wrap: break-word; /* IE fallback */
 	overflow-wrap: break-word;
 }
 
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -2407,6 +2407,14 @@ dl {
 	width: 100%;
 	float: left;
 }
+
+/* Signature preview */
+
+#preview_signature, #preview_signature_display {
+	width: 100%;
+	overflow: hidden;
+}
+
 /* Issue a warning */
 #warn_body {
 	width: 100%;
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -591,10 +591,6 @@ dl.settings img {
 	margin: 0 10px 0 0;
 	vertical-align: middle;
 }
-/* help icons */
-dl.settings dt a span {
-	vertical-align: top;
-}
 
 /* a general table class */
 table.table_grid {
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -205,7 +205,7 @@ body#help_popup {
 	margin: 0 2px;
 	padding: 2px 4px;
 	font-size: 9px;
-	font-family: verdana, monospace, sans-serif;
+	font-family: verdana, sans-serif;
 	box-shadow: 1px 2px 2px rgba(0,0,0,0.1), 0 -2px 4px rgba(0,0,0,0.1) inset;
 	opacity: 0.7;
 	transition: 0.3s all;
@@ -2186,6 +2186,11 @@ dl {
 #basicinfo h4 {
 	font-size: 1.4em;
 	font-weight: 100;
+	-webkit-hyphens: auto;
+	-ms-hyphens: auto;
+	hyphens: auto;
+	word-wrap: break-word; /* IE fallback */
+	overflow-wrap: break-word;
 }
 #basicinfo h4 span.position {
 	font-size: 0.8em;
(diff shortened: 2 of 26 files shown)


equinoxmatt:
Been looking at this for a short while tonight. I am not the best person to comment, but I took a look at your commits individually and checked your branch out. It seems fine, I can't see any issues.

I won't 'approve' it, because this isn't my area of expertise, but LTGM :+1:

wintstar:
@Gwenwyfar, the text in the reports is somewhat small.  Is there a way to make this a little bigger?
`index.php?action=admin;area=reports`

admin.css line 31
```
.table_caption, tr.table_caption td {
    color: #000;
    font-size: 10px;
    font-weight: bold;
}
```

Gwenwyfar:
This PR is already long enough, but I'll add font size revision to one of my next ones. There are some oddly sized fonts elsewhere as well.

Themes/default/css/calendar.css
```diff
@@ -265,10 +265,6 @@ span.hidelink {
 	-ms-hyphens: auto;
 	hyphens: auto;
 	overflow-wrap: break-word;
-	word-wrap: break-word;
```
Sesquipedalian: Apparently, IE and Edge < 18 don't support the standard `overflow-wrap`, so we need to keep `word-wrap: break-word;` for them. 
https://caniuse.com/#search=overflow-wrap

Themes/default/css/index.css
```diff
@@ -205,13 +205,14 @@ body#help_popup {
 	margin: 0 2px;
 	padding: 2px 4px;
 	font-size: 9px;
-	font-family: verdana, sans-serif; /* @todo Might add extra fallbacks. Possibly monospace. */
+	font-family: verdana, monospace, sans-serif;
```
Sesquipedalian: This change doesn't make a lot of sense to me. If we define a generic `monospace` before the generic `sans-serif`, then the latter will never be used. So I think we only need one, right? Personally, I'd stick with `sans-serif` rather than `monospace` for this.

Themes/default/css/index.css
```diff
@@ -2234,8 +2314,9 @@ dl {
 	box-shadow: 0 1px 0 #fff inset;
 	padding-top: 12px;
 	clear: both;
-	word-break: hyphenate;
-	word-wrap: break-word;
```
Sesquipedalian, line 2238: Because support for `hyphens` is dependent not only on browser, but also [on language](https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens#Browser_compatibility), we should keep `word-wrap: break-word; overflow-wrap: break-word;` as a fallback.

Themes/default/css/index.css
```diff
@@ -3121,10 +3169,10 @@ div#pollmoderation {
 .poster {
 	float: left;
 	/* Don't set this in em.It will eat too much space if people need to set large text sizes. */
-	/* Better to just break the few words here, and leave more space for actual topic content. */
 	width: 160px;
-	word-break: hyphenate;
-	word-wrap: break-word;
```
Sesquipedalian, line 3127: As noted above, because support for `hyphens` is dependent not only on browser, but also [on language](https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens#Browser_compatibility), we should keep `word-wrap: break-word; overflow-wrap: break-word;` as a fallback.

Themes/default/css/index.css
```diff
@@ -3227,7 +3275,7 @@ div#pollmoderation {
 	border-top: 1px solid #bfbfbf;
 	box-shadow: 0 1px 0 #fff inset;
 	min-height: 85px;
-	word-break: hyphenate;
+	word-break: break-all;
```
Sesquipedalian: `word-break: break-all;` is not very nice to readers. It'd be better to use `hyphens: auto;` (with the vendor prefixes, of course).

Sesquipedalian (changes requested):
Mostly just some changes regarding hyphenation and word wrapping.

Sesquipedalian:
Apparently GitHub doesn't allow comments on just any random line, so I couldn't add one for `#basicinfo h4` on line 2186 of index.css. That selector needs to have the following added to its ruleset in order to fix #4719:
```
word-wrap: break-word;
overflow-wrap: break-word;
-webkit-hyphens: auto;
-ms-hyphens: auto;
hyphens: auto;
```

Sesquipedalian:
I just noticed some other spots where `overflow-wrap` is being used without `word-wrap` being set as as compatibility fallback for IE. That should be fixed everywhere.

Gwenwyfar:
Just to update on this, I'm hoping to make the fixes soon, as I'm finally getting some free time back :)

XinYenFon:
I think at this stage its better to just remove the conflict and merge this big one in and do rest of the fixes later.

Themes/default/css/index.css
Gwenwyfar: Hyphens will not break some links and other overly long words. And unfortunately text-ellipsis needs white-space: nowrap which we don't want here either. And outside those cases there is enough space in this area that hyphens is not necessary.

Gwenwyfar:
Nevermind, got the quirks to replicate once more. We may need to watch out those places to see if they aren't breaking in strange ways, still. Those rules are still unreliable.

Themes/default/css/index.css
Sesquipedalian: If I can't convince you to use `hyphens: auto`, please at the very least use `word-wrap: break-word` instead of `word-break: break-all`. Using `word-wrap: break-word` will only break the word if absolutely necessary, whereas `word-break: break-all` will break any and all words whenever they reach the edge of the container.

Gwenwyfar: You're right, I missed this.

Sesquipedalian (approved):

Gwenwyfar:
Can you handle the conflicts, Sesq? I could do it but it's going to also merge the PR (if there's another way through GitKraken I haven't it figured out yet)

PR #4630, opened 2018-03-16
Minify fix

Sesquipedalian:
1. Fixes bugs retrieving and returning cached minified file info
2. Fixes a very minor bug in cache_get_data()
3. Minimizes most JavaScript files, because there is zero reason not to.


albertlast:
Possible reason to not minfied not now,
could be that the development get more complex,
but i do no frontend stuff ... i don't care.

Sesquipedalian:
> Possible reason to not minfied not now,
could be that the development get more complex,

Not an issue. When I'm working on JavaScript or CSS, I just uncheck the "Minimize CSS and JavaScript files" option in the admin panel. :)

Sesquipedalian:
Actually, I think it would be better to break this up into separate PRs.

PR #4632, opened 2018-03-16
Fixes bugs retrieving and returning cached minified file info

Sesquipedalian:
1. Fixes bugs retrieving and returning cached minified file info.
2. Fixes very minor bug in cache_get_data()


equinoxmatt (approved):
LGTM :+1:

PR #4633, opened 2018-03-16
Adds housekeeping for minimized CSS and JS files

Sesquipedalian:
Prevents old minimized files from accumulating in the theme directories.

The minimized files are removed in the following situations:

1. When the admin turns off minimization.
2. Whenever a mod is installed or uninstalled.
3. During the weekly maintenance.

### DIFF
--- a/Sources/ManageSettings.php
+++ b/Sources/ManageSettings.php
@@ -231,6 +231,10 @@ function ModifyBasicSettings($return_config = false)
 		saveDBSettings($config_vars);
 		$_SESSION['adm-save'] = true;
 
+		// Do a bit of housekeeping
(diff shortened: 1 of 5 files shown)


equinoxmatt (approved):
LGTM :+1: 

The only thing I would say, is there a way we can log any errors when removing files, rather than silently failing?

albertlast:
Maybe a bit picky,
but the function name "unlinkMinified" doesn't sound right to me.
By the name i would guess that i could give him a array with a list of minified files how get deleted than,
but this is not the case here.
Since you call it housekeeping i would call the function in the similiar way,
maybe "housekeepingMinified"

PR #4634, opened 2018-03-16
Minimizes most JavaScript

Sesquipedalian:
Minimizes most JavaScript files, because there is zero reason not to.

### DIFF
--- a/Sources/Admin.php
+++ b/Sources/Admin.php
@@ -30,7 +30,7 @@ function AdminMain()
 	// Load the language and templates....
 	loadLanguage('Admin');
 	loadTemplate('Admin');
-	loadJavaScriptFile('admin.js', array(), 'smf_admin');
+	loadJavaScriptFile('admin.js', array('minimize' => true), 'smf_admin');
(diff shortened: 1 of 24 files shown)


equinoxmatt:
Why not turn minimize on by default and if a particular call wants to not minimize (for some reason), it can be overridden?

Seems better than changing all these function calls.

I appreciate you may know something I don't, but I thought I would just put it out there.

Sesquipedalian:
We can minimize CSS by default because loading CSS is pretty straightforward and predictable. Once @albertlast built us a nice system to control load order, that was all the control we needed.

In contrast, JavaScript can have quite complex dependency structures. We can know in advance what our own scripts depend on, but there's no way to predict in advance when and how scripts added by mods might want to be loaded. So we need to leave it up to the modder to decide what to do.

Sesquipedalian:
By the way, in case it wasn't clear, the minimization process concatenates all the minimized data into a single file. If it kept each file separate, then there would be no reason not to do them all. But concatenation can potentially affect load order and dependency for JS.

PR #4638, opened 2018-03-18
handle empty message_labels array

albertlast:
Catch ``count(): Parameter must be an array or an object that implements Countable``
on line 1000 and 1072


albertlast:
Issue: https://www.simplemachines.org/community/index.php?topic=559398.0

colinschoen (approved):
LGTM

Let's verify this fixes the issue for the user.

albertlast:
Since the issue exists also in pg (was mitigated by the db layer),
i was able to reproduce this error on pg and notice after changing the sql issue was gone (no workaround needed).

albertlast:
@colinschoen pr got verify at sm.org

PR #4640, opened 2018-03-19
Improves loading of JavaScript regarding async and defer

Sesquipedalian:
All our supported browsers fully support the async and defer attributes for loading JavaScript files, and the best performance comes from loading all the scripts in the head and using those attributes. Why? This way the browser can download the JS files while simultaneously parsing the HTML into the DOM. In contrast, putting the script files at the end of the document means the browser only starts to download the scripts after it is done parsing the HTML.

(Well, the last sentence above may not be 100% true for all modern browsers. Apparently some scan the HTML looking for JS files to pre-fetch. But that is just an attempt to mimick the best practices that this change implements properly.)

For a handy explanation of the reasons for this, see https://bitsofco.de/async-vs-defer. For a comparative demo, see http://contentloaded.com/async-defer

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -3539,8 +3539,8 @@ function template_footer()
 
 /**
  * Output the Javascript files
- * 	- tabbing in this function is to make the HTML source look good proper
- *  - if defered is set function will output all JS (source & inline) set to load at page end
+ * 	- tabbing in this function is to make the HTML source look good and proper
+ *  - if defered is set function will output all JS set to load at page end
  *
  * @param bool $do_deferred If true will only output the deferred JS (the stuff that goes right before the closing body tag)
  */
@@ -3551,8 +3551,11 @@ function template_javascript($do_deferred = false)
 	// Use this hook to minify/optimize Javascript files and vars
 	call_integration_hook('integrate_pre_javascript_output', array(&$do_deferred));
 
-	$toMinify = array();
-	$toMinifyDefer = array();
+	$toMinify = array(
+		'standard' => array(),
+		'defer' => array(),
+		'async' => array(),
+	);
 
 	// Ouput the declared Javascript variables.
 	if (!empty($context['javascript_vars']) && !$do_deferred)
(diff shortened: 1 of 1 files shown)


emanuele45:
Personal curiosity: how does the async loading cope with the inline javascript all over the places in the template?

Sesquipedalian:
> Personal curiosity: how does the async loading cope with the inline javascript all over the places in the template?

The async attribute should be used on self-contained code that doesn't depend on anything else. If your external script will depend on any inline scripts, you should use defer on the external script instead of async.

albertlast:
By reading the first acrticel,
came i up with the question of inline js: https://github.com/Sesquipedalian/SMF2.1/blob/6824b9a983187adc6a9ef4f1521376937a1147eb/Sources/Subs.php#L3632
when you add a inline script you can define this as defer.
In the background smf2.1 do put the js in a script tag on start or when defer on the end(without defer).

But based on the arctiel, would be it not smart to load both type script at the start/sameplace and only define one script tag as defer?

Sesquipedalian:
> when you add a inline script you can define this as defer.

The defer and async attributes are ignored on inline scripts. (More information [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script).) Instead, inline scripts are always loaded and executed as soon as the HTML parser reaches them. So the only way to "defer" an inline script is to load it at the end of the document.

albertlast:
And to add the defer script to the ``<head>`` is not a good idea i guess?

Sesquipedalian:
> And to add the defer script to the `<head>` is not a good idea i guess?

Adding an external script to the head with the `defer` attribute is a good idea. But adding an inline script to the head is a bad idea if you want that script to be deferred.

emanuele45:
Jon Stovell wrote on 20/03/18 02:14:
>
>     Personal curiosity: how does the async loading cope with the
>     inline javascript all over the places in the template?
>
> The async attribute should be used on self-contained code that doesn't
> depend on anything else. If your external script will depend on any
> inline scripts, you should use defer on the external script instead of
> async.
>

My question was a "light hint" to the fact SMF has several inline
scripts that rely on the main script (either script.js, topic.js,
admin.js, etc) being loaded (for example any time an "object" is
instantiated like smc_Toggle as random example, but there are more, or a
function from a file is called inline during the loading of the page).

So my real question was: what will happen to those inline scripts when
the file is loaded asynchronously?
My suspect is that they will choke because the file is not yet loaded
and parsed and as such the function/class is not yet available, and your
answers seem to confirm my suspect, unless I misunderstood them.

But of course this is an edge case that will likely happen "only" the
first time a person opens the page of the forum (or at least any time he
will open a page that requires to load a new file).

Sesquipedalian:
Thanks for clarifying. The answer is simple: none of those are being loaded asynchronously. In fact, I don't believe any of SMF's standard scripts are loaded asynchronously (I'm on my phone at the moment, so I can't double check), but only either loaded in the head with normal loading or else deferred.

PR #4646, opened 2018-03-21
Source view was being loaded when WYSIWYG was requested.

NicolasBonet:


Gwenwyfar:
Don't forget you need to sign off your commits before they can be merged ;)

PR #4652, opened 2018-03-21
Don't let .tar at the end of the name for compressed theme installation.

NicolasBonet:
Signed-off-by: Nicolas Bonet <nibogo2@gmail.com>

### DIFF
--- a/Sources/Themes.php
+++ b/Sources/Themes.php
@@ -1331,7 +1331,7 @@ function InstallFile()
 
 	// Get the theme's name.
 	$name = pathinfo($_FILES['theme_gz']['name'], PATHINFO_FILENAME);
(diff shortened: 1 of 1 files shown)


Sesquipedalian (approved):

XinYenFon:
@Sesquipedalian - commit did not signed-off by sender.

Sesquipedalian:
@Antes: Good catch. I should have double checked. It's odd that the Travis script didn't pick that up.

@NicolasBonet: Did you amend the commit message and then force push it? If so, please don't do that in the future. It can really mess things up. At any rate, I will have to revert this PR for now. Please resubmit it with a properly signed commit (use `git commit -s` instead of just `git commit`).

PR #4656, opened 2018-03-23
Ensures "deferred" inline scripts run after deferred external scripts

Sesquipedalian:
Prior to #4640, when we deferred scripts merely by inserting them into the document right before the `</body>` tag, we always loaded the external scripts first and then the inline script after that. In #4640 we started loading the external scripts using the defer attribute instead. But the defer attribute causes a script not to be executed until after the DOMContentLoaded event. This is functionally equivalent to inserting the script after the closing `</html>`. But our deferred inline script was still being loaded and executed before the DOMContentLoaded event. That could cause problems if the inline script depended on the external script.

This PR solves the problem by simply ensuring the inline script isn't executed until after the DOMContentLoaded event. Because the external scripts with the defer attribute appeared earlier in the HTML, they will be executed before the content of the inline script. Thus the correct load order is maintained.

### DIFF
--- a/Sources/Calendar.php
+++ b/Sources/Calendar.php
@@ -476,8 +476,8 @@ function CalendarPost()
 		'name' => $context['page_title'],
 	);
 
-	loadCSSFile('jquery-ui.datepicker.css', array('defer' => false), 'smf_datepicker');
-	loadCSSFile('jquery.timepicker.css', array('defer' => false), 'smf_timepicker');
+	loadCSSFile('jquery-ui.datepicker.css', array(), 'smf_datepicker');
+	loadCSSFile('jquery.timepicker.css', array(), 'smf_timepicker');
 	loadJavaScriptFile('jquery-ui.datepicker.min.js', array('defer' => true), 'smf_datepicker');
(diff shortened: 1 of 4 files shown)


albertlast:
Since i oberserver this topic from some distance i had to ask a litte bit from "big picture".
When i count correctly we got 5 or 6 different kind of js.
- Head without any attribut
- Head with async
- Head with deferred
- Body inline start
- Body inline deferred
- "random" script tag from theme

What is load order of this?
- Head w/o, Head async, Body inline start, random script, Head deferred, inline deferred
- Head w/o, Body inline start, Head async, random script, Head deferred, inline deferred
- ??

Sesquipedalian:
> * Head w/o, Head async, Body inline start, random script, Head deferred, inline deferred
> * Head w/o, Body inline start, Head async, random script, Head deferred, inline deferred

Either of these could be correct at any given time. It is not possible to know exactly when "Head async" will load. Async scripts are loaded as soon as the browser has finished fetching them. If "Head async" is already in the browser's cache, it will be loaded before "Body inline start." Otherwise, it will be loaded whenever it becomes available. Perhaps that will happen after "random script" but before "Head deferred." Perhaps it will not be until after "inline deferred." It all depends on how long the script file takes to download from the server.

PR #4666, opened 2018-03-25
Re-allowing HTML like tags and preventing XSS

NicolasBonet:
Signed-off-by: Nicolas Bonet <nicolas.bonet@globant.com>

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -1558,7 +1558,7 @@ function loadMemberContext($user, $display_custom_fields = false)
 
 			// Enclosing the user input within some other text?
 			if (!empty($custom['enclose']))
-				$value = strtr($custom['enclose'], array(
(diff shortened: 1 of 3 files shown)


Sesquipedalian:
I've had a chance to test this now, @NicolasBonet. Unfortunately, I found a bug. 

If I paste `<a href="http://www.example.com">example</a>` into the Location field, then the input field will break on the next time the profile page loads:
screen shot 2018-04-03 at 11 58 24 pm

Sesquipedalian (changes requested):
More robust handling of input is required.

Sesquipedalian:
Since @NicolasBonet hasn't replied, I'm going to close this for now. If and when he gets a chance to push more commits to address the bug I found, we can look at it again.

PR #4670, opened 2018-03-26
Fixes undefined errors on profile

### LINKED ISSUE #4669: Undefined errors each time you access to an user's profile
You get various undefined errors each time you access to any user's profile. Here's an example;

frandominguezl:
Fixes #4669


Sesquipedalian:
Thanks for this, @d3vcho. I believe #4672 will be a better fix, but I'll leave this open for now.

Issue #4677, opened 2018-03-30
Can't open the admin panel


Sefank:
When open the admin panel, it comes that
> Parse error: in /home/bae/app/Sources/Load.php(3055) : eval()'d code on line 51

frandominguezl:
Are you using the latest GitHub available version?

Sefank:
Yes, using commit ver b017107.

wintstar:
Which language package do you use?

I have with this commit no problems.

XinYenFon:
can't reproduce this issue.

albertlast:
Test please the pr #4680 maybe the issue get fixed or the error message get more clear.

Sefank:
Yes, it's a language package error. Thanks for the pr #4680 to find where it's wrong.

albertlast:
And the language pack is not the offical from smf2.1 beta4?

wintstar:
@albertlast, on simpelmachines.org either a note should be added to the downloads that the 2.1 language packs are not up to date or the language packs should be updated. Some of the error messages here, could thus be avoided.

Sefank:
@albertlast The offical one from smf2.1 beta 4? Should it be got from this repo? The language editor shows that the English version was only updated to beta 3.

albertlast:
The beta 4 language file got some new text array entries so the beta 3 doesn't work in beta 4

Sefank:
OK I will check them. Anyway thanks~

wintstar:
> The beta 4 language file got some new text array entries so the beta 3 doesn't work in beta 4

@albertlast, then this should be published on simpelmachines.org.

albertlast:
@wintstar since it the second time in a short time that you tell me what sm team had to do...
please look at my status "Collaborator" that mean i provide some amount of code but i'm not part of the sm.org team.

wintstar:
Sorry, @albertlast. After your great work here, I assumed you are part of the team.

> On simpelmachines.org either a note should be added to the downloads that the 2.1 language packs are not up to date or the language packs should be updated. Some of the error messages here, could thus be avoided.
>>The beta 4 language file got some new text array entries so the beta 3 doesn't work in beta 4

@Sesquipedalian, you could take it on here?

albertlast:
This got in my eyes fixed,
since we got here two problems
eval got issue of missing language stuff -> got fixed by pr
language file from beta 3 was used -> don't do it

albertlast:
[...]

PR #4680, opened 2018-03-30
Removed Template Eval

albertlast:
Could fix issue: #4677 
but eaven it not fix this.
The whole eval stuff is slow, error mesage are garbage and debugging sucks -> see no reason to keep this.

btw. The helptext is wrong (direct opposite is the right case).

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -3023,7 +3023,7 @@ function censorText(&$text, $force = false)
 }
 
 /**
- * Load the template/language file using eval or require? (with eval we can show an error message!)
+ * Load the template/language file using require
  * 	- loads the template or language file specified by filename.
  * 	- uses eval unless disableTemplateEval is enabled.
  * 	- outputs a parse error if the file did not exist or contained errors.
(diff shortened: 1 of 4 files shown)


albertlast:
@Arantor antes mention that you got reason why eval is ther.

albertlast:
The response from the issue #4677 shows that eval create a fog around the real issue.
When the eval keep in, the support team of sm.org get a hard time when some one got a issue,
where eval is included.

Arantor:
The reason for template eval is that if a template has a fatal error at the parse stage, it can be caught and a usable error can be shown to users.

I know that people have trouble with template eval, though honestly for someone competent, it's at worst a few minutes extra detective work. The line number is guaranteed correct by the Zend engine, but the file will be wrong... though it's hardly a difficult process to identify which file it will be in almost every case.

albertlast:
For caught error we don't need eval,
we already did this on other places where such behavior is needed,
that we change temp the error handler of php/smf in a more strict one.
exp. https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs-Post.php#L628

> Parse error: in /home/bae/app/Sources/Load.php(3055) : eval()'d code on line 51

doesn't help any one.

Arantor:
The idea is to cope if index.template.php has a syntax error. Back in 2003 it seemed like a good idea. But it's your call if you want to remove it, don't drag me into it.

Sesquipedalian:
Given what the PHP manual page for eval() says, I'm inclined to agree with @albertlast on this. In 2003 it probably was a good idea. I'm not sure that it still is helpful in 2018.

Sesquipedalian (approved):

Sesquipedalian:
Any thoughts to share, @colinschoen? If not, I'm inclined to merge this.

Sesquipedalian:
Comments from @MissAllSunday would also be welcome. :)

colinschoen (approved):
LGTM

Times have changed.

PR #4692, opened 2018-04-04
Exposes the 'cron_is_real_cron' setting to the admin

### LINKED ISSUE #3916: Bad SEO, bots doing scheduled task.
I seen this in the 2.0 branch and same applies to 2.1.

In Load.php
`		if (isBrowser('possibly_robot'))
		{
			// @todo Maybe move this somewhere better?!
			require_once($sourcedir . '/ScheduledTasks.php');
			// What to do, what to do?!
			if (empty($modSettings['next_task_time']) || $modSettings['next_task_time'] < time())
				AutoTask();
			else
				ReduceMailQueue();
		}`

If the user is a bot then the scheduled task is being called if there is one set. Problem I see with this is that some robots, such as GoogleBot see page speed as a indicator of site rankings. Under the right circumstances this offloading to the bots might hurt a forums web presence.

I suggest in this instance we don't treat the bots any differently.

Sesquipedalian:
Fixes #3916 by giving the admin a way to disable the "web cron" JavaScript.

Of course, this means an admin could screw things up by disabling the "web cron" method without actually configuring a real cron task to run cron.php. But I've included lots of warnings and helpful explanations, ~~so if the admin is an idiot, (1) it's his own fault, and (2) its trivial for him to figure out how to fix his mistake~~ and I've also included a little function that checks once every 24 hours to make sure that scheduled tasks have been running as expected and automatically reverts to the standard, JavaScript-based method if they have not.


albertlast:
I like it that you able to exclude the run from the client side.
But the "revert" stuff i got two question:
Why you don't create an error (which get logged)?
Why 24h? ( i mean a server can be down for 24h because of crash/upgrade/backup),

Maybe you should create an error when the cron is called by not js,
while js crone(the standard) is active so that the admin see that you revert his changes.

And maybe would be nice to mention in the help or so,
what the url to cron.php is.

Sesquipedalian:
> Why you don't create an error (which get logged)?

[I do](https://github.com/SimpleMachines/SMF2.1/pull/4692/files#diff-6b95ae55062797979a2f26e651403101R6573). 🙂 The purpose of that error message is to inform the admin if the web cron method was turned back on, and why. Then the admin will know to take appropriate action.

> Why 24h? ( i mean a server can be down for 24h because of crash/upgrade/backup)

I chose 24 hours because the daily maintenance script is supposed to be run every 24 hours. If that's not happening, there will be problems. 

If the server went down for more than 24 hours, the admin will already be aware of why the web cron method was reactivated. The admin can turn it back off at any time.

> Maybe you should create an error when the cron is called by not js,
while js crone(the standard) is active so that the admin see that you revert his changes.

I think it makes more sense to give the error if cron is supposed to be working, but hasn't been. After all, nothing bad will happen if the admin uses both a real cron job and the web cron method at the same time, so logging an error when both are being used doesn't make much sense. It only makes sense to log an error if neither are being used.

> And maybe would be nice to mention in the help or so,
what the url to cron.php is.

Good idea. 👍

albertlast:
When you apply a backup your 24 hours logic get fired also,
Reason why I got concern about this magic.

Sesquipedalian:
Yes, but like I said, it doesn't hurt anything if 'cron_is_real_cron' is turned off even though a real cron job exists. All that happens in that case is that both methods will be used to run cron.php. So yes, it would be turned off unnecessarily in that scenario, but it also wouldn't do any harm.

PR #4693, opened 2018-04-05
Improves sanitization during registration

Sesquipedalian:
1. This makes it harder for people to spoof usernames by inserting unusual whitespace characters, because now we eliminate all of those.
2. Strips out all control characters from user input, not just line endings.

### DIFF
--- a/Sources/Register.php
+++ b/Sources/Register.php
@@ -284,7 +284,16 @@ function Register2()
 	foreach ($_POST as $key => $value)
 	{
 		if (!is_array($_POST[$key]))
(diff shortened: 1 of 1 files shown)


jdarwood007:
Since 2.1 is UTF8 only, we don't really need to pull the UTF8 check and just force it.

Sesquipedalian:
> Since 2.1 is UTF8 only, we don't really need to pull the UTF8 check and just force it.

One of these days soon, we should go through all the code to make sure that there are no loopholes that could allow other character encodings to be used and then rip out all these tests. In the meantime, though, I'd rather code defensively. :)

PR #4704, opened 2018-04-06
Fix Upgrade from 2.0.x pg to 2.1 pg

albertlast:
Found on the way a special bug which you get,
when you call the upgrade.php without being logged in.


Sources/Subs-Db-postgresql.php
```diff
@@ -1017,4 +1002,33 @@ function smf_db_custom_order($field, $array_values, $desc = false)
 	return $return;
 }
 
+/**
+ * Function which return the information if the database support native replace inserts
```
colinschoen: nit: support -> supports

colinschoen (changes requested):
LGTM (with nit)

colinschoen (approved):
LGTM

PR #4705, opened 2018-04-06
Change content -> posts.

### LINKED ISSUE #4696: like any content
in permissions we have "Can like any content", currently we only have likes in posts, i think this should be changed to "Can like any posts" since this better reflects how this feature works. we would need to update the help text as well if this is changed.

colinschoen:
Fixes #4696


Sesquipedalian:
Failing the EOF line ending check.

colinschoen:
Thanks

MissAllSunday:
Still think the change isn't needed, the likes is pretty much designed to work with anything that has a unique id, not just posts.

colinschoen:
Yep, I created this to have the discussion here. I think it is reasonable that mod authors will extend the functionality too, but let's give Illori a chance to respond.

tinoest:
I don’t see what’s up with ‘content’ myself. Surely a forum is made up of ‘content’ be that a thread or a post or users who contribute to the site. If you like a post you might be liking the user who created the ‘content’ rather than the post itself. Or even the ‘content’ of the post.

illori:
i think since by default SMF only allows liking on posts, this fits. if we were to add liking to profiles for example then this would be fine. it is just defining for the admin what they are allowing users to like. we dont even know that mods will extend this or how they will do so.

XinYenFon:
if they extend the permission usage which I think its possible then the wording requires another change. You don't have to be this "core" on certain things. This PR is not required and should not be merged into the core-code.

Sesquipedalian:
I'm in favour of leaving it as "content" rather than "posts."

colinschoen:
Agreed, closing.

PR #4706, opened 2018-04-07
Show better text when spell check is not avaible

albertlast:
The core issue is that the exiting text is only shown,
when the function are not ther,
but the text tell a different story.


albertlast:
Okay looked wrong in the code,
the text is shown everytime.
But it easy for the user to know that red is death and not red is good.

Themes/default/languages/Admin.english.php
```diff
@@ -551,6 +551,7 @@
 $txt['enableEmbeddedFlash_warning'] = 'may be a security risk!';
 $txt['enableSpellChecking'] = 'Enable spell checking';
 $txt['enableSpellChecking_warning'] = 'this does not work on all servers.';
+$txt['enableSpellChecking_notavailable'] = 'no pspell or enchant found on this server.';
```
Sesquipedalian: "This is not available on your server."

Themes/default/languages/Admin.english.php
```diff
@@ -551,6 +551,7 @@
 $txt['enableEmbeddedFlash_warning'] = 'may be a security risk!';
 $txt['enableSpellChecking'] = 'Enable spell checking';
 $txt['enableSpellChecking_warning'] = 'this does not work on all servers.';
```
Sesquipedalian: $txt['enableSpellChecking_available'] = "This is available on your server.";

Sources/ManagePosts.php
```diff
@@ -198,7 +198,7 @@ function ModifyPostSettings($return_config = false)
 			array('check', 'removeNestedQuotes'),
 			array('check', 'enableEmbeddedFlash', 'subtext' => $txt['enableEmbeddedFlash_warning']),
 			// Note show the warning as red if: pspell not installed and (enchant not installed or not using UTF-8 and iconv not installed)
-			array('check', 'enableSpellChecking', 'subtext' => ($can_spell_check ? $txt['enableSpellChecking_warning'] : ('<span class="alert">' . $txt['enableSpellChecking_warning'] . '</span>'))),
+			array('check', 'enableSpellChecking', 'subtext' => ($can_spell_check ? $txt['enableSpellChecking_warning'] : ('<span class="alert">' . $txt['enableSpellChecking_notavailable'] . '</span>'))),
```
Sesquipedalian: array('check', 'enableSpellChecking', 'subtext' => ($can_spell_check ? $txt['enableSpellChecking_available'] : ('<span class="alert">' . $txt['enableSpellChecking_notavailable'] . '</span>')), 'disabled' => !$can_spell_check),

Sesquipedalian (changes requested):
If the functions are not available, we should disable the setting. Also, we might as well give the user a more specific subtext message in both cases.

Themes/default/languages/Admin.english.php
albertlast: Need we any message when it's available?

Sesquipedalian (approved):

PR #4710, opened 2018-04-10
Restores "Mark Read" button on message index

### LINKED ISSUE #4707: Missing "Mark Read" option in Message Index
In SMF 2.0.x we had the "Mark Read" option on the top of the message index. This option seems to have disappeared in SMF 2.1.x, and we think it should be nice to have it back since it's easier to mark topics as read using that option rather than using checkboxes and the dropmenu.

Sesquipedalian:
Fixes #4707


colinschoen (approved):
LGTM

PR #4715, opened 2018-04-15
Fix background jobs for Notify, AddMembergroup and handling ip null values

albertlast:
Notice that some task of my background table doesn't got deleted and
this is the result of this hunt.

- Notify methode didn't return true
- member_affected look like the right one to get some information how got new rights
- since ip value are null able and we decide not to allow null on Subs-Db* layer we need to convert the null to null string


colinschoen (approved):
LGTM

PR #4721, opened 2018-04-22
Fix timezone list format use of 'empty'

gizmo71:
Don't use `empty` on an expression, just negate it as a boolean.

Otherwise, PHP will report "Can't use function return value in write context".

On my PHP installation, this prevented the upgrader script from even starting.

Signed-off-by: Dave 'Gizmo' Gymer <github@davegymer.org>


Sesquipedalian (approved):

Sesquipedalian:
Oops. I forgot to test for PHP 5.4 compatibility. Thanks for the PR, @gizmo71. 🙂

PR #4725, opened 2018-04-24
Uses proper CSS spec for word-wrap and hyphens

### LINKED ISSUE #4719: Basic info username with lots of characters and no spaces
https://github.com/SimpleMachines/SMF2.1/blob/dd179b23c765f51a490c9464e4db0cd06aa86137/Themes/default/css/index.css#L2100

this id css need extra rules as if displayname at url ?action=profile contains lots of characters with no spaces (+20) the name flows outside element. 

This will fix it for sure in IE. & Chrome, not tested other browsers.
```
word-break: break-all;
-ms-word-break: break-all;
```

Sesquipedalian:
Fixes #4719

### DIFF
--- a/Themes/default/css/calendar.css
+++ b/Themes/default/css/calendar.css
@@ -261,11 +261,12 @@ span.hidelink {
 
 /* Break long words in calendar table cells to avoid layout problems */
 #main_grid td.days {
-	-webkit-hyphens: auto;
-	-ms-hyphens: auto;
-	hyphens: auto;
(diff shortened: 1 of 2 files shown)


Gwenwyfar:
Most of these have already been fixed in #4629. And we don't need the -moz- rules.

Sesquipedalian:
Excellent. If there is anything in here that isn't already in #4629, please go ahead and add it.

PR #4726, opened 2018-04-24
Fix Child boards get marked new where their is a new post in a grand … fixes #1734

### LINKED ISSUE #1734: 0000908: Child boards get marked new where their is a new post in a grand child that isn't visible.
For example the dev board here gets marked new when posts are made in child boards of it that we can't see.

Ref: http://dev.simplemachines.org/mantis/view.php?id=908

albertlast:
…child that isn't visible

This pr do two things:
- don't update the the parent board with the new message id
- run from the bottom to the top to search for is_read false and mark all level above in the same way

@Sesquipedalian is that a good way?

issue #1734

### DIFF
--- a/Sources/Subs-BoardIndex.php
+++ b/Sources/Subs-BoardIndex.php
@@ -47,7 +47,7 @@ function getBoardIndex($boardIndexOptions)
 	$result_boards = $smcFunc['db_query']('', '
 		SELECT' . ($boardIndexOptions['include_categories'] ? '
 			c.id_cat, c.name AS cat_name, c.description AS cat_desc,' : '') . '
-			b.id_board, b.name AS board_name, b.description,
+			b.child_level, b.id_board, b.name AS board_name, b.description,
 			CASE WHEN b.redirect != {string:blank_string} THEN 1 ELSE 0 END AS is_redirect,
 			b.num_posts, b.num_topics, b.unapproved_posts, b.unapproved_topics, b.id_parent,
 			COALESCE(m.poster_time, 0) AS poster_time, COALESCE(mem.member_name, m.poster_name) AS poster_name,
@@ -64,7 +64,7 @@ function getBoardIndex($boardIndexOptions)
 			LEFT JOIN {db_prefix}log_boards AS lb ON (lb.id_board = b.id_board AND lb.id_member = {int:current_member})') . '
 		WHERE {query_see_board}' . (empty($boardIndexOptions['countChildPosts']) ? (empty($boardIndexOptions['base_level']) ? '' : '
 			AND b.child_level >= {int:child_level}') : '
-			AND b.child_level BETWEEN ' . $boardIndexOptions['base_level'] . ' AND ' . ($boardIndexOptions['base_level'] + 1)) . '
+			AND b.child_level >= ' . $boardIndexOptions['base_level']) . '
 			ORDER BY ' . (!empty($boardIndexOptions['include_categories']) ? 'c.cat_order, ' : '') . 'b.child_level, b.board_order',
 		array(
 			'current_member' => $user_info['id'],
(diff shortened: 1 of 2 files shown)


Sesquipedalian (commented):
The basic idea here is good. In order to solve the issue, we need to do the job in `getBoardIndex()` rather than in `updateLastMessages()`. However, the logic you are using here could be made much more efficient by using a recursive function and `array_map()`, or something along those lines.

Sesquipedalian:
In order to make a good recommendation for how to improve the efficiency of this code, I would need to develop my own code. But if I do that, I might as well just submit it as a new pull request. So, @albertlast, would you prefer to take another shot at it yourself based on the suggestion in my review, or would you prefer if I put together a new PR? I'm fine with either.

albertlast:
Make please a own pr,
So that I can see where you try to go,
When your solution is better than we can take your pr.

albertlast:
Since i now back on pc,
i can look deeper what you mean and
I don't understand you.

Because the main task is done by getBoardIndex in updateLastMessages I delete only some code,
which is wrong with the new logic.

And i see no performance issue here,
the logic should work fine by over 1k boards.

Sesquipedalian:
When I wrote this:

>  The basic idea here is good. In order to solve the issue, we need to do the job in `getBoardIndex()` rather than in `updateLastMessages()`.

I was saying that you had the right idea when you decided to take some logic out of `updateLastMessages()` and add some to `getBoardIndex()`.

When I wrote this:

> However, the logic you are using here could be made much more efficient by using a recursive function and array_map(), or something along those lines.

I was saying that the code you added to `getBoardIndex()` is somewhat convoluted and could made simpler and more efficient.

Is that clearer now? :)

albertlast:
okay,
i looked in the php doc for array_map since ther is no chance to define the order in which the array is
worked through,
i see no possibility to implement the logic into array_map.

For the existing logic it's essential that you start the work from the bottom throw the top (and skip the "top top" board lvl 0)

And since we speaking here possible performance improvment (the logic is fine),
i see the problem not to merge this pr and comeback with a "better" solution.

emanuele45:
Indeed the logic is fine, what matters here for performances (and actually also the logic, is when you have:
```
- root 
    |-b1
        |-b11-1
            |-b111-1
            |-b111-2
            |- ...
            |- b111-1000
    |-b2
        |-b21-1
            |-b211-1
            |-b211-2
            |- ...
            |- b211-1000
    |-b3
        |-b31-1
            |-b311-1
            |-b311-2
            |- ...
            |- b311-1000
```
In the "current" code, if you are visiting b1, you loop over 6 rows (b1, b2, b3, b11-1, b21-1, b31-1), with the new one you loop over 3006 (b1, b2, b3, b11-1, b21-1, b31-1, b111-1...b1111-1000, b211-1...b2111-1000, b311-1...b3111-1000).
Of course, visiting B11-1 or any other higher level there would be no difference, but considering the root (aka board index) is usually the most hit, with the new code it would mean query and loop over all boards at each and every page load.
Indeed an edge case anyway, so maybe not even worth mentioning.

Of course, unless I misread the code.

albertlast:
This is not a issue of the logic,
It's a mysql issue.
For pg/mysql8.0/Mariadb 10.2 I could place a better query, which run only throw the selected branches.

Sesquipedalian:
I've had a chance to upload a PR tackling this issue now, @albertlast. I didn't end up using `array_map()` myself either, but I did find a way to avoid looping over the array twice. The key change compared to your code was using `$row_boards[$row_board['id_board']] = $row_board;` instead of just `$row_boards[] = $row_board;` when creating the array. By giving known keys to each element in `$row_boards`, I could manipulate the values of parent boards directly without needing to loop multiple times.

PR #4727, opened 2018-04-25
Added Join Permission System for query_see_board_join

albertlast:
Motivation: 
Scaling on mide size boards

I notice that my midsize board with 100 boards takes many time for the board index select,
bottleneck was the find_in_set stuff which is used for permission check in {query_see_board}.

So i added and normalize the permissions data by adding another table which can be used in a join statement,
to make the convertion easy i create another query "var" query_see_board_join,
which work in the most case like a drop in replace 

Execution time of board index query (for none admin user)
old: "Execution time: 1.663 ms"
new: "Execution time: 0.249 ms"

old:
https://github.com/SimpleMachines/SMF2.1/blob/a1f315def7a730949906e952af4429a662619696/Sources/Subs-BoardIndex.php#L47-L74
new:
https://github.com/albertlast/SMF2.1/blob/5dd5c95a0be9bbba6d62ce1b8d0d8212ccdc40f2/Sources/Subs-BoardIndex.php#L47-L75


Sesquipedalian:
So, the basic idea here is to use the board_permissions_view table to cache the information about which boards a member group has access to, so that we can quickly query the board_permissions_view table instead of performing the complex FIND_IN_SET query to get that data.

I do like the idea of remembering that data rather than performing the complex query every time, but I wonder if it would be better to simply cache that data rather that creating an entire new table, query system, and code structure.

If you were to cache the list of id_board values for each group and then retrieve them from the cache at [this point](https://github.com/SimpleMachines/SMF2.1/pull/4727/files#diff-6b95ae55062797979a2f26e651403101L6301) in the process, you could set the value of `$query_part['query_see_board']` to a simple `(b.id_board IN (<list of integers>)`. If the cached value didn't exist or was expired, you'd need to do a query to create it, of course, so that would be slower the first time, but after that it should be faster.

My reason for suggesting this alternative approach is that, because we really are trying to get RC1 out soon, we need to avoid making major structural changes at this point. Major changes like this will be appropriate for SMF 3.0, though, so don't delete this code entirely. Just save it for later. :)

Sesquipedalian (commented):
So, the basic idea here is to use the board_permissions_view table to cache the information about which boards a member group has access to, so that we can quickly query the board_permissions_view table instead of performing the complex FIND_IN_SET query to get that data.

I do like the idea of remembering that data rather than performing the complex query every time, but I wonder if it would be better to simply cache that data rather that creating an entire new table, query system, and code structure.

If you were to cache the list of id_board values for each group and then retrieve them from the cache at [this point](https://github.com/SimpleMachines/SMF2.1/pull/4727/files#diff-6b95ae55062797979a2f26e651403101L6301) in the process, you could set the value of `$query_part['query_see_board']` to a simple `(b.id_board IN (<list of integers>)`. If the cached value didn't exist or was expired, you'd need to do a query to create it, of course, so that would be slower the first time, but after that it should be faster.

My reason for suggesting this alternative approach is that, because we really are trying to get RC1 out soon, we need to avoid making major structural changes at this point. Major changes like this will be appropriate for SMF 3.0, though, so don't delete this code entirely. Just save it for later. :)

albertlast:
First thing,
we got a n:m relationship -> which mean we need a second table to solve the relationship.
This table is until now missing which i fixed in this pr, so this is not a caching table.

Like i said it's  not a caching.

You could implement this when you planing to introduce a caching -> which i didn't done by this pr.

I don't trust the story of rc1 get released,
so i keep doing stuff when something is wrong like here.

albertlast:
I keep the old/wrong one only because  of backward compatibility.

albertlast:
And @Sesquipedalian what is the status of this pr?

jdarwood007:
Looking at this, its a major change.  I would think that @colinschoen would need to give a approval as the lead dev if we want to push this PR.

Also to nitpick, I dislike using "grp" for the group id.  It should follow the format id_group to remain consistent across the code and database structure.

Also any change like this shouldn't change how the existing query_see_board works.  Sites like SM.org do use this outside of SMF.  Some mods even use it such as portals.  Will make upgrading a lot more work for them.  It should be kept around for legacy, but recommend action is for developers to switch to it.

albertlast:
Sure it's a major change.

I renamed and sync the datatype def.

And I know that the query_see_board should not be toucht -> reason why i not touch this and keep both exists.

Sesquipedalian:
My stance on this PR remains unchanged. The proposed board_permissions_view table is designed simply to track which boards a given membergroup has access to. This could be accomplished with much less interference in the existing code by caching a list of accessible board ids for each membergroup. 

For example, if membergroup 5 can access boards 1, 2, 8, and 12, we would cache that list of board ids using a label like `board_permissions_view_group5` or whatever. Then when building `query_see_board`, we would simply retrieve the cached value and use `b.id_board IN (1,2,8,12)` as the value of `$query_part['query_see_board']`. No muss, no fuss.

albertlast:
I hope you notice that you see the issue,
but instead of fixing this you try to build a workaround.

And this show me that the existing is wrong,
like i said before the data is not normalize and this get us to the issue we got now.

You need a table how resolve the n:m relationship between membergroups and boards and
the wrong old solution keeps in place.

Sesquipedalian:
@albertlast, would it be possible to make this work with the `(SELECT DISTINCT id_board FROM ... )` subquery in the WHERE instead of a JOIN? Perhaps if we used something like `b.id_board IN (SELECT DISTINCT id_board FROM ... )` as the value of `$query_part['query_see_board']`, we could get the benefits of using the table without having to change the rest of the code structure.

albertlast:
Well...
All pg version and only newer mysql would be rewrite a:
``select * from smf_boards a where a.id_board in (select id_board from smf_board_permissions_view where id_group = 9)``
to a join:
```
Hash Semi Join  (cost=4.76..113.54 rows=101 width=182) (actual time=0.048..0.143 rows=102 loops=1)"
  Hash Cond: (a.id_board = smf_board_permissions_view.id_board)"
  ->  Seq Scan on smf_boards a  (cost=0.00..107.01 rows=101 width=182) (actual time=0.007..0.071 rows=102 loops=1)"
  ->  Hash  (cost=3.49..3.49 rows=102 width=4) (actual time=0.031..0.031 rows=102 loops=1)"
        Buckets: 1024  Batches: 1  Memory Usage: 12kB"
        ->  Seq Scan on smf_board_permissions_view  (cost=0.00..3.49 rows=102 width=4) (actual time=0.005..0.018 rows=102 loops=1)"
              Filter: (id_group = 9)"
              Rows Removed by Filter: 17"
Planning time: 0.203 ms"
Execution time: 0.170 ms"
```
https://www.percona.com/blog/2010/10/25/mysql-limitations-part-3-subqueries/

So when your idea is to sneaky backstabe many mysql setup be my guest,
but for me is this move/idea realy evil.
Gesendet: Mittwoch, 4. Juli 2018 23:15
An: SimpleMachines/SMF2.1
Cc: albertlast; Mention
Betreff: Re: [SimpleMachines/SMF2.1] Added Join Permission System for query_see_board_join (#4727)


@albertlast<https://github.com/albertlast>, would it be possible to make this work with the (SELECT DISTINCT id_board FROM ... ) subquery in the WHERE instead of a JOIN? Perhaps if we used something like b.id_board IN (SELECT DISTINCT id_board FROM ... ) as the value of $query_part['query_see_board'], we could get the benefits of using the table without having to change the rest of the code structure.

albertlast:
Well...
All pg version and only newer mysql would be rewrite a:
``select * from smf_boards a where a.id_board in (select id_board from smf_board_permissions_view where id_group = 9)``
to a join:
```
Hash Semi Join  (cost=4.76..113.54 rows=101 width=182) (actual time=0.048..0.143 rows=102 loops=1)"
  Hash Cond: (a.id_board = smf_board_permissions_view.id_board)"
  ->  Seq Scan on smf_boards a  (cost=0.00..107.01 rows=101 width=182) (actual time=0.007..0.071 rows=102 loops=1)"
  ->  Hash  (cost=3.49..3.49 rows=102 width=4) (actual time=0.031..0.031 rows=102 loops=1)"
        Buckets: 1024  Batches: 1  Memory Usage: 12kB"
        ->  Seq Scan on smf_board_permissions_view  (cost=0.00..3.49 rows=102 width=4) (actual time=0.005..0.018 rows=102 loops=1)"
              Filter: (id_group = 9)"
              Rows Removed by Filter: 17"
Planning time: 0.203 ms"
Execution time: 0.170 ms"
```
https://www.percona.com/blog/2010/10/25/mysql-limitations-part-3-subqueries/

So when your idea is to sneaky backstabe many mysql setup be my guest,
but for me is this move/idea realy evil.

albertlast:
what would be if we provide both solution @Sesquipedalian ?
the "new" version from me already use a different query var,
your solution use the existing query var,
so both solution could exists by side.

Sources/Subs.php
```diff
@@ -6295,11 +6295,19 @@ function build_query_board($userid)
...
+	{
 		$query_part['query_see_board'] = '1=1';
+		$query_part['query_see_board_join'] = '';
+	}
 	// Otherwise just the groups in $user_info['groups'].
 	else
+	{
 		$query_part['query_see_board'] = '((FIND_IN_SET(' . implode(', b.member_groups) != 0 OR FIND_IN_SET(', $groups) . ', b.member_groups) != 0)' . (!empty($deny_boards_access) ? ' AND (FIND_IN_SET(' . implode(', b.deny_member_groups) = 0 AND FIND_IN_SET(', $groups) . ', b.deny_member_groups) = 0)' : '') . (isset($mod_cache) ? ' OR ' . $mod_cache['mq'] : '') . ')';
-
+		$query_part['query_see_board_join'] = 'JOIN (SELECT DISTINCT id_board FROM ' . $db_prefix . 'board_permissions_view WHERE (id_group IN ( '. implode(',', $groups) .') and deny = 0) '
```
jdarwood007: Can we use {db_prefix} here or is there a reason for this?

jdarwood007: I was reminded that this is called from {query_see_board_join}, so it can't be done here unless we looped through it twice or with a db_quote.  Not worth fixing for that.

albertlast:
@jdarwood007 @Sesquipedalian rewrited the pr,
in the mention fashion from sesqui.
could we merge this now?

Sesquipedalian:
Okay @albertlast, I just have a few questions at this point:

1. Earlier you said that using a SELECT subquery in the WHERE would be backstabbing the database engine and really evil. Do the changes you made in 74ca217 mean that you've changed your opinion on that?

2. As of 451722c, query_see_board_join isn't being used anywhere. So, is there a reason to build it in the `build_query_board()` function anymore?

Sources/Subs-Boards.php
```diff
@@ -655,6 +655,45 @@ function modifyBoard($board_id, &$boardOptions)
 				'selected_board' => $board_id,
 			))
 		);
+	
+	// Do permission sync
+	if (!empty($boardUpdateParameters['deny_groups']))
+		$insert = array();
```
Sesquipedalian: There should probably be braces around the statements after this if.

albertlast:
About 
1. I read about this topic and notice that in my test env every database do rewrite internaly to somekind of join when exists is used which is the wanted behavior, on the otherside when the unwanted behavior is went (many sub query calls) the solution would be same fast as the existing solution -> which is okay
and in the end of the day this version is easier intigrated into smf

2. well posible for the freedome to use a different way to do the permission check, but from techincal pov exists no reason

Sesquipedalian (approved):

PR #4730, opened 2018-04-26
Improves the language file editor

### LINKED ISSUE #1699: Language editor can fail due to PHP's max_input_vars setting
The language editor in the admin center can fail at times due to PHP's "max_input_vars" setting. The reason for this is that for each language string in the file, SMF sends two values to the server: one containing the string's original value and one containing the value sent from the form (edited or not). Then, if the value sent from the form is the same as the original value, we don't change that string. This limits the number of calls to `cleanLangString()` and also means we don't have to save anything if nothing was actually changed.

The problem is, this causes the number of $_POST items to grow to a ridiculous number for larger language files, and, in some cases, will prevent you from being able to edit certain language files (I ran into this issue with the Admin language file, which has well over 500 language strings in it).

There are about 3 ways we can handle this:
1. Add .htaccess and .user.ini files to the directories as needed to increase input_max_vars.
2. Split up some entries in the Admin language file (and others as needed) to prevent them from having so many language strings
3. Stop the comparison and just assume that everything has changed, even if it hasn't.

I would assume things were done this way for a reason (besides "don't save changes to the file if nothing actually changed"), so I wanted further input on this situation before we decide how to fix it (or even if it should be fixed).

Sesquipedalian:
* Cleans up the UI
* Fixes #1699
* Groups text strings by type & gives users some help for editing them
* Adds a system to allow the admin to add and/or remove strings from language files. To avoid potential accidents screwing up someone's forum, only certain safe files allow adding and removing.
* When editing or removing a string, comments out the old version rather than actually deleting it. This should make it easier for an admin to restore default values with a text editor if a mess is made.
* Improves the regexes involved in this operation


Sesquipedalian:
Having tested this for a while myself, I am confident in it.

PR #4737, opened 2018-04-28
New hook (integrate_movetopic2_end)

dragomano:
Adds integration hook for additional operations in MoveTopic2 function.


illori:
we cannot accept this PR without you signing off the commit.

PR #4738, opened 2018-04-28
New hook (integrate_movetopic2_end)

dragomano:
Adds integration hook for additional operations in MoveTopic2 function.

### DIFF
--- a/Sources/MoveTopic.php
+++ b/Sources/MoveTopic.php
@@ -355,6 +355,8 @@ function MoveTopic2()
 	// Notify people that this topic has been moved?
 	sendNotifications($topic, 'move');
 
+	call_integration_hook('integrate_movetopic2_end');
+
 	// Why not go back to the original board in case they want to keep moving?
 	if (!isset($_REQUEST['goback']))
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Thanks for this pull request. The code looks good, and we'd like to include it. The only problem is that, for legal reasons regarding SMF's licensing, we require every merged commit to include a "Signed-off-by" line in its commit message. As it says in SMF'S readme,

> sign-off your commits, to acknowledge your submission under the license of the project.
It is enough to include in your commit comment "Signed-off by: " followed by your name and email address (for example: `Signed-off-by: Angelina Belle <angelinabelle1@hotmail.com>`)

If you are using the command line to make your Git commits, the easiest way to do that is to use the `-s` switch when you make your commit, like this:
```
git commit -s
```

Alternatively, you can just manually add the necessary text to your commit messages, but that gets tedious really quickly if you have to do it more than once.

Anyway, if I can get you to close this PR and resubmit it one more time with that Signed-off-by line in the commit message, then we will be glad to merge it.

Sesquipedalian (approved):

PR #4739, opened 2018-04-29
Show banned text on Login popup fixes #4603

### LINKED ISSUE #4603: banned user trying to login
if you use the login popup while you are banned, it just shows "error" and nothing else where the rest of the page shows your banned message. i think the popup should show the same error if you are not allowed to log in.

albertlast:
Was unkown to me,
that we can catch status code.
grafik

The login form is not part of the xml response -> no chance to display this.

issue: #4603

### DIFF
--- a/Themes/default/scripts/script.js
+++ b/Themes/default/scripts/script.js
@@ -329,7 +329,10 @@ function reqOverlayDiv(desktopURL, sHeader, sIcon)
 		},
 		statusCode: {
 			500: function() {
-				oPopup_body.html(banned_text);
+				if (sHeader == 'Login')
+					oPopup_body.html(banned_text);
+				else
+					oPopup_body.html('500 Internal Server Error');
 			}
 		}
 	});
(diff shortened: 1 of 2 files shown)


Themes/default/scripts/script.js
```diff
@@ -326,6 +326,11 @@ function reqOverlayDiv(desktopURL, sHeader, sIcon)
 		},
 		error: function (xhr, textStatus, errorThrown) {
 			oPopup_body.html(textStatus);
+		},
+		statusCode: {
+			500: function() {
+				oPopup_body.html(banned_text);
```
Sesquipedalian: It would be good to do this: 
```
if (sHeader == 'Login')
	oPopup_body.html(banned_text);
else
	oPopup_body.html('500 Internal Server Error');
```

Then if the Ajax call ever receives a 500 status code response for something besides the login form, it won't display the banned message.

Sesquipedalian (changes requested):

Sesquipedalian (approved):

PR #4740, opened 2018-04-29
Log backtrace on error

albertlast:
Issue like that: https://www.simplemachines.org/community/index.php?topic=559941.0
show me that we need more information from where the issue came,
to get the information i added backtrace to error log:

added a new icon to the error log
grafik
and this is the popup
grafik

### DIFF
--- a/Sources/Errors.php
+++ b/Sources/Errors.php
@@ -115,8 +115,11 @@ function log_error($error_message, $error_type = 'general', $file = null, $line 
 	// Make sure the category that was specified is a valid one
 	$error_type = in_array($error_type, $known_error_types) && $error_type !== true ? $error_type : 'general';
 
+	// Collect backtrace
+	$backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 10);
+
 	// Don't log the same error countless times, as we can get in a cycle of depression...
-	$error_info = array($user_info['id'], time(), $user_info['ip'], $query_string, $error_message, (string) $sc, $error_type, $file, $line);
+	$error_info = array($user_info['id'], time(), $user_info['ip'], $query_string, $error_message, (string) $sc, $error_type, $file, $line, $backtrace);
 	if (empty($last_error) || $last_error != $error_info)
 	{
 		// Insert the error into the database.
--- a/Sources/Errors.php
+++ b/Sources/Errors.php
@@ -115,8 +115,10 @@ function log_error($error_message, $error_type = 'general', $file = null, $line 
 	// Make sure the category that was specified is a valid one
 	$error_type = in_array($error_type, $known_error_types) && $error_type !== true ? $error_type : 'general';
 
-	// Collect backtrace
+	// Collect backtrace and remove himself
 	$backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 10);
+	array_shift($backtrace);
+	$backtrace = json_encode($backtrace);
 
 	// Don't log the same error countless times, as we can get in a cycle of depression...
 	$error_info = array($user_info['id'], time(), $user_info['ip'], $query_string, $error_message, (string) $sc, $error_type, $file, $line, $backtrace);
--- a/Sources/Errors.php
+++ b/Sources/Errors.php
@@ -118,7 +118,7 @@ function log_error($error_message, $error_type = 'general', $file = null, $line 
 	// Collect a backtrace (but leave out the call to log_error)
 	$backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 20);
 	array_shift($backtrace);
-	$backtrace = json_encode($backtrace);
+	$backtrace = !empty($smcFunc['json_encode']) ? $smcFunc['json_encode']($backtrace) : json_encode($backtrace);
 
 	// Don't log the same error countless times, as we can get in a cycle of depression...
 	$error_info = array($user_info['id'], time(), $user_info['ip'], $query_string, $error_message, (string) $sc, $error_type, $file, $line, $backtrace);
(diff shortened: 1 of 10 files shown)


Sources/Errors.php
```diff
@@ -115,8 +115,13 @@ function log_error($error_message, $error_type = 'general', $file = null, $line
 	// Make sure the category that was specified is a valid one
 	$error_type = in_array($error_type, $known_error_types) && $error_type !== true ? $error_type : 'general';
 
+	// Collect backtrace and remove himself
+	$backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 20);
+	array_shift($backtrace);
+	$backtrace = json_encode($backtrace);
```
Sesquipedalian: Should use `$smcFunc['json_encode']()` here. Right now it is just an alias for the standard json_encode function, but that could change in a future version.

other/upgrade_2-1_mysql.sql
```diff
@@ -2638,3 +2638,11 @@ UPDATE {$db_prefix}settings
 SET value = replace (value, '\n{$default_fugue_smileyset_name', '')
 WHERE variable = 'smiley_sets_names';
 ---#
+
+/******************************************************************************/
+--- Change log_erros add backtrace
```
Sesquipedalian: "Add backtrace to log_error"

Sources/Errors.php
```diff
@@ -115,8 +115,13 @@ function log_error($error_message, $error_type = 'general', $file = null, $line
 	// Make sure the category that was specified is a valid one
 	$error_type = in_array($error_type, $known_error_types) && $error_type !== true ? $error_type : 'general';
 
+	// Collect backtrace and remove himself
```
Sesquipedalian: "Collect a backtrace (but leave out the call to log_error)"

other/upgrade_2-1_postgresql.sql
```diff
@@ -2463,3 +2463,11 @@ UPDATE {$db_prefix}settings
 SET value = replace (value, '\n{$default_fugue_smileyset_name', '')
 WHERE variable = 'smiley_sets_names';
 ---#
+
+/******************************************************************************/
+--- Change log_erros add backtrace
```
Sesquipedalian: "Add backtrace to log_error"

Themes/default/Errors.template.php
```diff
@@ -256,4 +256,24 @@ function template_attachment_errors()
...
+
+	echo '<!DOCTYPE html>
+<html', $context['right_to_left'] ? ' dir="rtl"' : '', '>
+	<head>
+		<meta charset="', $context['character_set'], '">
+		<title>Backtrace</title>
+		<link rel="stylesheet" href="', $settings['theme_url'], '/css/index', $context['theme_variant'], '.css', $modSettings['browser_cache'], '">
+	</head>
+	<body>
+	', var_dump(!empty($context['error_backtrace'])? $context['error_backtrace'] : ''),'
```
Sesquipedalian: `print_r()` might be more appropriate than `var_dump()` here.

Sesquipedalian (changes requested):
This is a great idea! Just a few minor changes needed.

Themes/default/Errors.template.php
albertlast: the display of print_r is not usefull,
reason why i keep stick on var_dump

Sources/Errors.php
albertlast: I evade with intention the use of json_encode since we are here in the error handling area,
so i can't besure if we already cross the point where this function exists.

jdarwood007: I would suggest using then function_exists($smcFunc['json_encode']).  The code is there to ensure if something like serialized ever pops up (not that I can see with what it supports), we have a easy way to mass fix the API without changing tons of lines of code.  json_ functions are used to handle user input and thus why we don't need to do it for all functions.

Themes/default/Errors.template.php
Sesquipedalian: Yes it is, and it is easier to read.

Here is the output using `<pre>', var_dump(!empty($context['error_backtrace'])? $context['error_backtrace'] : ''), '</pre>`
screen shot 2018-04-29 at 5 57 37 pm

Here is the output using `<pre>', print_r(!empty($context['error_backtrace'])? $context['error_backtrace'] : ''), '</pre>`
screen shot 2018-04-29 at 5 58 13 pm

Also, please wrap the output in `<pre></pre>` so that whitespace is preserved properly. Without them, the output looks like this in my tests:
screen shot 2018-04-29 at 6 04 32 pm

Sources/Errors.php
albertlast: I don't like it,
to got two function how do the same,
but i added this to last pr.

Themes/default/Errors.template.php
albertlast: Done also i added a css class,
so when someone like to change the style he is free todo so.

Sesquipedalian:
Looking good so far. I have one more idea for improvement. It would be good to move [these lines of code](https://github.com/SimpleMachines/SMF2.1/pull/4740/files#diff-c3661331ad8815ccef396da83b6e02f5R118) up to [this location](https://github.com/SimpleMachines/SMF2.1/pull/4740/files#diff-c3661331ad8815ccef396da83b6e02f5L40). Then we could delete some redundant lines inside [this if statement](https://github.com/SimpleMachines/SMF2.1/pull/4740/files#diff-c3661331ad8815ccef396da83b6e02f5R42).

Sesquipedalian:
... Actually, the array_shift and json encoding would need to go after that if statement, I suppose. But I'm sure you get the idea. 🙂

albertlast:
Want we in general when debug is enable that we got the complet (with methode parameters) backtrace? 
https://github.com/SimpleMachines/SMF2.1/pull/4740/files#diff-c3661331ad8815ccef396da83b6e02f5L47

albertlast:
I took a little different route about remove duplicate code:
the array_shift and json encode keep in the lower,
my thoughts are in a error loop the error methode could be a reason for the error -> i want the complete trace
in a "normale" error the error part is not part of the issue.
on the other hand i like the idea when enable debug to get are more complex backtrace instead of the version without args.

Sesquipedalian (approved):

Sesquipedalian:
That makes sense to me.

PR #4742, opened 2018-04-30
Backtrace UI and quick bug fix

Sesquipedalian:
* Quick fix for a bug I missed in my testing of #4740
* Gives a prettier UI for backtrace info.


Sesquipedalian:
screen shot 2018-04-30 at 4 17 41 pm

Sesquipedalian:
Merging ASAP to avoid broken error logging on MySQL.

DiegoAndresCortes:
I was wondering if this error is related
```

Warning
: mysqli_stmt_bind_param() expects parameter 1 to be mysqli_stmt, boolean given in
C:\xampp\htdocs\SMF2.1\Sources\Subs-Db-mysql.php
on line
994

Warning
: mysqli_stmt_execute() expects parameter 1 to be mysqli_stmt, boolean given in
C:\xampp\htdocs\SMF2.1\Sources\Subs-Db-mysql.php
on line
995

Warning
: mysqli_stmt_bind_param() expects parameter 1 to be mysqli_stmt, boolean given in
C:\xampp\htdocs\SMF2.1\Sources\Subs-Db-mysql.php
on line
994

Warning
: mysqli_stmt_execute() expects parameter 1 to be mysqli_stmt, boolean given in
C:\xampp\htdocs\SMF2.1\Sources\Subs-Db-mysql.php
on line
995
```

I was updating a mod to SMF 2.1 and ran into this after upgrading my local site, it wasn't happening before.
I'm not sure what is causing it, but it always appears after submitting a form and some other random situations. Anything I should change that I'm missing?

DiegoAndresCortes:
So after coming back to this I realized that it only occurs when the error log is triggered. So if I get a fatal_error alert, these warnings pop up.
Still don't know why but I'll try to keep looking into it

DiegoAndresCortes:
So apparently I'm supposed to add the second parameter in the function stating if it should be logged or indicating what type of error is.
I would have figured this out pretty fast but I was on vacations so..

Anyways it'd have been nice if someone had told me this was necessary, since it didn't seem to be before.

PR #4752, opened 2018-05-06
php7+ crash on windows 1803 strftime format %Z

albertlast:
I notice on my env that apache crash since windows 10 1803

Test code for this issue:
```
<?php
$value = strftime('%Z');
echo $value.'%Z';
echo '<br>';

setlocale(LC_TIME,'en_US.UTF-8');
$value = strftime('%Z');
echo $value.'%Z';
echo '<br>';

setlocale(LC_CTYPE,'en_US.UTF-8');
$value = strftime('%Z');
echo $value.'%Z';
echo '<br>';
```


jdarwood007:
Did you file a bug report with PHP as well?  Seems relevant to them to patch this as well.

albertlast:
nope

wintstar:
I had the same problem. Your fix works for me.

albertlast:
Since i got the confirmation from wintstar i open a bug report on php

jdarwood007:
Awesome.  If we can get confirmed affected versions and if they are mainstream in distros, I can see fixing this.  If its not a release making distros, then it may not be worth it.
I.e, if you downloaded something like wamp or iis config tool.

albertlast:
What is not worth?

MissAllSunday:
Fix the issue on SMF side. He means to wait until we know which versions are affected by this issue and patch it up if popular distros ship those versions.

Although I could argue (without having exact figures of course) that a good chunk of mod authors do use tools like xampp or wamp in windows to develop and that a good amount of those users are also using win10 so it might be worth to fix even if this doesn't affect any other major distributions.

albertlast:
When you read the documention on php side,
you notice that %z and %Z are not supported anyway on windows.
So it make less sense to test this.

wintstar:
You might also consider that this bug could affect hosters with Windows Server. Or not?

albertlast:
@Sesquipedalian @colinschoen would be nice if we get the fix asap,
the development without this fix is very hard...

albertlast:
Add more generic way and set z to the list of unsupportedFormats

jdarwood007:
If Z isn't supported on windows, I'm fine with removing it if the server is windows based.  Although I don't see that mentioned in the php.net/date at all.

albertlast:
You had to look here: http://php.net/manual/de/function.strftime.php

Sesquipedalian:
I think you could accomplish the same goal more simply by adding ` || ($context['server']['is_windows'] && in_array($format, array('z', 'Z')))` to the end of [this if statement](https://github.com/SimpleMachines/SMF2.1/pull/4752/files#diff-6b95ae55062797979a2f26e651403101R839).

albertlast:
Nope this point is to late.
The logic needs to be in front of: `` $value = @strftime('%' . $format); ``

Sesquipedalian:
In that case, I would insert the following bit of code into [here](https://github.com/SimpleMachines/SMF2.1/pull/4752/files#diff-6b95ae55062797979a2f26e651403101L821) immediately before the `$value = @strftime('%' . $format);` line.

```
// Avoid a crashing bug with PHP 7 on certain versions of Windows
if ($context['server']['is_windows'] && in_array($format, array('z', 'Z')))
{
	$unsupportedFormats[] = $format;
	continue;
}
```

albertlast:
Done

Sources/Subs.php
```diff
@@ -824,6 +826,13 @@ function timeformat($log_time, $show_today = true, $offset_type = false, $proces
 		{
 			foreach($strftimeFormatSubstitutions as $format => $substitution)
 			{
+				// Avoid a crashing bug with PHP 7 on certain versions of Windows
+				if ($context['server']['is_windows'] && in_array($format, $unsupportedFormatsWindows))
```
Sesquipedalian, line 830: Should be `in_array($format, array('z', 'Z'))`

Sesquipedalian (changes requested):

Sources/Subs.php
albertlast: In my eyes is worst to read and hard to maintaine,
reasone why i keep to define the array in front of the function.

Sesquipedalian: Further evidence that I should go to bed now.

Sesquipedalian (approved):

PR #4754, opened 2018-05-08
Mock as someone else when we do background task fixes #4753

### LINKED ISSUE #4753: fetch_alterts doesn't load user data
Based on the backtrace from https://www.simplemachines.org/community/index.php?topic=559941.msg3971592#msg3971592
i think it's important that fetch_alterts load the user data from $memID,
because the follow logic based around the idea that $user_info is setted.

albertlast:
the underlaying api expect that the curent user = the user is who get the stuff
The orginal issue: https://www.simplemachines.org/community/index.php?topic=559941.0
came from CreatePost_Notify_Background which is created when you mention someone for exp.

I'm not sure if the logic perfect...
but what it do -> check if the $user_info is the same how is targeting by this process,
when not we save the old one, get the new one process throw the process and resetet on the end to the old one.

Issue #4753

### DIFF
--- a/Sources/Profile-View.php
+++ b/Sources/Profile-View.php
@@ -222,7 +222,7 @@ function fetch_alerts($memID, $all = false, $counter = 0, $pagination = array(),
 	if (empty($user_info) || $user_info['id'] !== $memID)
 	{
 		$user_old = !empty($user_info) ? $user_info : null;
-		loadMemberData($memID, false, 'minimal', true);
+		loadMemberData($memID, false, 'minimal');
 		$user_info = $user_profile[$memID];
 	}
 
--- a/Sources/Profile-View.php
+++ b/Sources/Profile-View.php
@@ -213,10 +213,18 @@ function summary($memID)
  */
 function fetch_alerts($memID, $all = false, $counter = 0, $pagination = array(), $withSender = true)
 {
-	global $smcFunc, $txt, $scripturl, $memberContext;
+	global $smcFunc, $txt, $scripturl, $memberContext, $user_info, $user_profile;
 
 	$query_see_board = build_query_board($memID);
 	$query_see_board = $query_see_board['query_see_board'];
+
+	// are we someone else?
+	if (empty($user_info) || $user_info['id'] !== $memID)
+	{
+		$user_old = !empty($user_info) ? $user_info : null;
+		loadMemberData($memID, false, 'minimal');
+		$user_info = $user_profile[$memID];
+	}
 
 	$alerts = array();
 	$request = $smcFunc['db_query']('', '
@@ -374,6 +382,8 @@ function fetch_alerts($memID, $all = false, $counter = 0, $pagination = array(),
 			$alerts[$id_alert]['text'] = str_replace($search, $repl, $txt[$string]);
 		}
 	}
+
+	$user_info = !empty($user_old) ? $user_old : null;
 
 	return $alerts;
 }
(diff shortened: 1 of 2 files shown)


Sesquipedalian (approved):

Sesquipedalian:
Hm. According to the reporting user, these changes didn't fix the problem. I am a bit surprised by that.
https://www.simplemachines.org/community/index.php?topic=559941.msg3971785#msg3971785

Do you mind following up with him, @albertlast?

albertlast:
When you look at the line of code where the issue happen,
you notice that the issue is now a different one.

The new one get from register-notify

Without looking deeper the issue could be,
that the new user didn't got at this time no member entry,
so in this case the question what is the "default" user when the user didn't exists?

albertlast:
@Sesquipedalian maybe i was not fully clear, could you tell me what kind of user default value we got for non existing user?

Sesquipedalian:
> When you look at the line of code where the issue happen,
>you notice that the issue is now a different one.
>
>The new one get from register-notify

Yes, you are right. But that means the problem is unlikely to be missing info for guest users, since guests can't be sent notifications. It looks like there the time_format value is missing for an admin or a global moderator on the forum of the person who reported the bug. I'm not sure how that happened (I suspect that the person was mucking around in the database directly) but the best solution is probably just to make sure that all the required values are defined inside `timeformat()` itself.

albertlast:
timeformat is user dependent and
this why we're here with this pr.

Sesquipedalian:
> timeformat is user dependent and
> this why we're here with this pr.

Yes, but Register-Notify.php doesn't send anything to guests. It only sends things to members with the 		moderate_forum permission (see [here](https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/tasks/Register-Notify.php#L31)). So the error must be happening for a member that has that permission but does not have a value set for `$user_info['time_format']` for some reason.

The code I've just submitted in #4759 should solve the immediate problem. However, I still also want to merge your code in 976885e. You are entirely right that we need to make sure we are using the correct set of `$user_info` data in `fetch_alerts()` so that we send the alerts with the correct time format. I don't think we need the code in d7a0c78, though, so if you could please revert that commit, then I will merge this PR and #4759.

albertlast:
Well i forgot the signed-off for the revert,
is this a issue?

Sesquipedalian:
> Well i forgot the signed-off for the revert,
is this a issue?

Amend the commit and force push it, and that should be fine.

albertlast:
rebase fix the issue.

Sesquipedalian:
Great.

Sources/Profile-View.php
```diff
@@ -213,11 +213,19 @@ function summary($memID)
...
+	global $smcFunc, $txt, $scripturl, $memberContext, $user_info, $user_profile;
 
 	$query_see_board = build_query_board($memID);
 	$query_see_board = $query_see_board['query_see_board'];
 
+	// are we someone else?
+	if (empty($user_info) || $user_info['id'] !== $memID)
+	{
+		$user_old = !empty($user_info) ? $user_info : null;
+		loadMemberData($memID, false, 'minimal');
```
Sesquipedalian: Oh, I just noticed that you are calling `loadMemberData($memID, false, 'minimal')` here. It needs to be `loadMemberData($memID, false, 'profile')` in order to get the member's `time_format` value. That's probably why the errors were still happening.

Sesquipedalian (changes requested):
Change 'minimal' to 'profile'

albertlast:
got fixed

Sesquipedalian (approved):

PR #4757, opened 2018-05-10
In membersAllowedTo, only checks mod groups for the board in question

Sesquipedalian:
This check in `membersAllowedTo()` definitely should be there.


colinschoen (approved):
LGTM

PR #4758, opened 2018-05-13
Update txt sentence

MissAllSunday:


Themes/default/languages/ManagePermissions.english.php
```diff
@@ -206,7 +206,7 @@
 $txt['permissionname_move_own'] = 'Own topic';
 $txt['permissionname_move_any'] = 'Any topic';
 $txt['permissionname_lock'] = 'Lock topics';
-$txt['permissionhelp_lock'] = 'This permission allows a user to lock a topic. This can be done in order to make sure no one can reply to a topic. Only uses with a \'Moderate board\' permission can still post in locked topics.';
+$txt['permissionhelp_lock'] = 'This permission allows an user to lock a topic. This can be done in order to make sure no one can reply to a topic. Only users with a \'Moderate board\' permission can still post in locked topics.';
```
Sesquipedalian: Actually, "a user" is correct. Because "user" is pronounced "you-ser," there is no need for the "n." Just another example of how English is weird.

Sesquipedalian (changes requested):

smfbigguy:
The first part of the line you changed reads better. It should be changed to:

This permission allows a  user to lock a topic. This can be done in order to make sure no one can reply to a topic. Only users with a \'Moderate board\' permission can still post in locked topics.';

Sorry don't mean to be critical or anything it just reads better like this.

Themes/default/languages/ManagePermissions.english.php
smfbigguy: Sorry I didn't see you posted below the code there.

colinschoen (approved):
LGTM

PR #4759, opened 2018-05-13
Ensures required values are set in timeformat()

Sesquipedalian:


albertlast:
the user_info is wrong user when timeformat is called by background task and
$modSettings['time_offset'] got not the value from the user.

albertlast (changes requested):
Ensure that you the setting from the user for them the background task is running,
otherwise you send data in wrong format etc. out.

Sesquipedalian:
Yes. There are two separate problems.

1. We need to ensure we are using the correct set of `$user_info`. Your code in #4754 addresses that.

2. We need to ensure that all required values in `timeformat()` are defined, just in case any are missing from whichever set of `$user_info` data we are using. This PR addresses that.

albertlast:
When the time_format is not setted when i do loadMemberData,
than loadMemberData got an issue and not the timeformat function in my eyes...

albertlast:
now we didn't know if the fix works...

PR #4768, opened 2018-05-20
Update Dropzone to 5.4.0

### LINKED ISSUE #4766: Large attachment uploads do not complete
When uploading a (very) large attachment, the Dropzone script seems to stop sending data after a certain amount has been uploaded. The upload simply stops and no further progress is made. In my tests, this happened at approximately 18.9 MB.

When I reverted fcca6bf (a.k.a. #4436), the issue disappeared.

I suspect that the cause may have something to do with the upgrade to Dropzone 5.2. That version introduced [chunked uploads](https://gitlab.com/meno/dropzone/wikis/faq#chunked-uploads).

MissAllSunday:
It seems upgrading to 5.4.0 fixes this, at least it does on my test env.

fixes #4766 


Sesquipedalian:
Unfortunately, the error still persists in my tests. It's odd, though. I've found that the error only occurs in some environments, and not in others. But in the ones where it does occur, it is consistently solved by reverting fcca6bf, so I don't think it is (just) a server issue. I haven't yet figured out what the difference might be, but I am looking into it.

In the meantime, upgrading DropZone to 5.4.0 is still a good idea, so I'm going to merge this but leave #4766 open.

PR #4770, opened 2018-05-21
Adds `timeout: null` to default Dropzone options

### LINKED ISSUE #4766: Large attachment uploads do not complete

Sesquipedalian:
Fixes #4766

### DIFF
--- a/Themes/default/scripts/smf_fileUpload.js
+++ b/Themes/default/scripts/smf_fileUpload.js
@@ -20,6 +20,7 @@ function smf_fileUpload(oOptions) {
 		autoQueue: false,
 		clickable: '.fileinput-button',
 		currentUsedSize: 0,
+		timeout: null,
 		smf_insertBBC: function (file, w, h) {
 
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Do you see any potential issues with this @MissAllSunday? As explained [here](https://github.com/SimpleMachines/SMF2.1/issues/4766#issuecomment-390770378), I think all this does is restore the behaviour that was the default in the earlier versions of dropzone.min.js that we were using before.

MissAllSunday:
Nope, shouldn't be an issue 👍

MissAllSunday (approved):

PR #4771, opened 2018-05-22
Sceditor update

### LINKED ISSUE #4763: Code tags
I hate to do this. It should be a simple fix though.....I think. Code tags do not copy and paste right. This could be a wysiwyg editor problem really. If copy and pasting code into the editor and then using code tags, it treats it like a paragragh not a code block. No formating at all.

Sesquipedalian:
- ~Fixes #4763~
- Upgrades SCEditor to v2.1.3
- Makes logic for setting up the editor a bit more legible
- Fixes some oddities with the AutoYouTube plugin


Sesquipedalian:
Of #4763.

Sesquipedalian:
Since this PR doesn't actually fix #4763, and I've determined that the issue is with SCEditor's AutoYouTube plugin, I'm simply going to disable and remove the AutoYouTube plugin, file a bug report over on the SCEditor repo, and call it a day. If and when Sam Clarke rewrites the AutoYouTube plugin, someone can release a mod to add the AutoYouTube plugin.

jdarwood007:
Tested this PR.  Don't see any issues with the Sceditor.  Youtube bbc works fine still.  Auto youtube is taken out as expected.

PR #4772, opened 2018-05-22
Looks for fallback images if the expected smiley image is not found

### LINKED ISSUE #4767: Text smileys broken on upgrade
In my 2.0 forum, we use the 'classic' smiley set.  After upgrade, we are still on the 'classic' set - which is .gif based.  

Using a text smiley - e.g., "8)" for "cool" - is broken after upgrade, because SMF 2.1 assumes your default smileys are always .pngs.   It can't find "cool.png" in the classic folder, and displays a broken link.  

In short, all text-based smileys are broken after the upgrade if your forum default points to a .gif based set.

Sesquipedalian:
Fixes #4767

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -2775,6 +2775,9 @@ function parsesmileys(&$message)
 	if ($user_info['smiley_set'] == 'none' || trim($message) == '')
 		return;
 
+	// Maybe a mod wants to implement an alternative method (e.g. emojis instead of images)
+	call_integration_hook('integrate_smileys', array(&$smileyPregSearch, &$smileyPregReplacements));
+
 	// If smileyPregSearch hasn't been set, do it now.
 	if (empty($smileyPregSearch))
 	{
@@ -2858,7 +2861,7 @@ function parsesmileys(&$message)
 			}
 		}
 
-		$smileyPregSearch = '~(?<=[>:\?\.\s' . $non_breaking_space . '[\]()*\\\;]|(?<![a-zA-Z0-9])\(|^)(' . implode('|', $searchParts) . ')(?=[^[:alpha:]0-9]|$)~' . ($context['utf8'] ? 'u' : '');
+		$smileyPregSearch = '~(?<=[>:\?\.\s' . $non_breaking_space . '[\]()*\\\;]|(?<![a-zA-Z0-9])\(|^)(' . build_regex($searchParts) . ')(?=[^[:alpha:]0-9]|$)~' . ($context['utf8'] ? 'u' : '');
 	}
 
 	// Replace away!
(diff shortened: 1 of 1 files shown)


frandominguezl (approved):

sbulen:
I am getting undefined index errors with this PR:
```
Type of error: Undefined 
8: Undefined index:
 https://shawnbulen.com/VGF21/index.php?topic=20418.0
 /yada/yada/public_html/VGF21/Sources/Subs.php (Line 2867)

```
And the smileys aren't always showing:
image

Sesquipedalian:
Empty your cache in the Maintenance admin panel, then try again.

sbulen:
No improvement...  

Note I replicated the same symptoms on a pure vanilla 2.1 environment (no upgrade run).

sbulen:
These errors are being generated even when there are no smilies...

Sesquipedalian:
Derp. Just a dumb error on my part. Please try again now.

Sesquipedalian:
Based on [further discussion](https://github.com/SimpleMachines/SMF2.1/issues/4773#issuecomment-391483067), I now propose this PR as a temporary fix until I can implement a better, permanent solution.

sbulen:
.gifs still don't work.  They are no longer broken links - but they don't work.  Both as forum defaults, and when personal sets are chosen.  

If you like, I can provide admin credentials to my test site online so you can test for yourself.  

Link to the post:
https://shawnbulen.com/VGF21/index.php?msg=151471

sbulen:
Note that personal smiley selections for .gif based sets still show up as broken links:
image

sbulen:
The problem with this PR is the change from implode to build_regex.  If you restore that line back to using implode, parseSmilies works for all extensions.  

Also, a recommendation - if you can set the extension outside of the loop thru the smilies, then you can get away with one filesystem hit instead of 20+.  E.g., you really don't have to check all 20+ smiley extensions individually.  If you find the first one is a .gif, I suggest making the assumption that all 20+ in that directory are .gifs.  There shouldn't be mixed extensions within the same smiley folder.  

===================

Note this only solves part of the problem - displaying smilies in messages.   

There are at least two other places where .gif smilies are still broken:  The smiley icons in the editor when posting a message and the smiley icons displayed when changing your personal smiley set under your profile, Look and Layout.

Sesquipedalian:
This should fix the problem with using build_regex. I'm not going to do much else with this PR, though, since it is really just a temporary, band-aid solution until I can create a proper fix.

sbulen:
This works for posts!

The limitations above still hold - profile Look & Layout & the acp Smiley Settings page still have issues with .gifs.

sbulen:
A thought...

Why should we specify the extension?  It's already in the file system...  It really should be sufficient to only specify the smiley folder.

If you follow the suggestion above, we only need one single file-exists check out of this logic.  (As opposed to the 20+ it does today.)

In other words, with a tweak, this may be good as a long term solution.  

Other benefits: No other upgrader, installer, ACP or profile changes needed.

Even the single file check may be avoided if you cache the end result...  Rethink the caching and we may be closer than we think...   Done, even...

Just thinking aloud...

PR #4775, opened 2018-05-26
Better logic to determine is_read status; makes countChildBoards more efficient.

### LINKED ISSUE #1734: 0000908: Child boards get marked new where their is a new post in a grand child that isn't visible.

Sesquipedalian:
1. Fixes #1734
2. Improves performance when `countChildPosts ` is enabled

This code is inspired by @albertlast's ideas in #4726 to put all the query results into an array and then process that array and to loop through them in reverse order so that relevant values in the descendent boards can affect the ancestor boards.

The key change compared to albertlast's code is using `$row_boards[$row_board['id_board']] = $row_board;` instead of just `$row_boards[] = $row_board;` when creating the array. By giving known keys to each element in `$row_boards`, we can manipulate the values of parent boards without needing to loop over the array multiple times.

Not only does this allow us to fix the original bug, but it also allowed us to greatly improve performance when the `countChildPosts ` setting is enabled. Previously, we had to do a bunch of inefficient work to construct the `$parent_map` tree in order to figure out where and how to propagate post and topic counts from children to parents. This is no longer necessary, so propagating counts from children to parents is now quick and painless. Moreover, because constructing `$parent_map` could become _very_ slow if there were multiple levels of children involved, the database query previously limited the results to only the top level and immediate child boards when the `countChildPosts ` was enabled; now that artificial limitation is unnecessary and has been removed. So now, when that setting is enabled, the post and topic totals for a board will include not only the totals for the board itself and its immediate children, but for the board itself and all of its descendants. This obviously fits much better with the intended purpose of the setting, and now we can do it without a performance penalty.

A final point to note: The query now uses `(CASE WHEN COALESCE(lb.id_msg, 0) >= b.id_last_msg THEN 1 ELSE 0 END) AS is_read` instead of `(CASE WHEN COALESCE(lb.id_msg, 0) >= b.id_msg_updated THEN 1 ELSE 0 END) AS is_read`. This means the `id_msg_updated` column is no longer used when creating the board index. It might be worth investigating whether that column is even necessary any more. If not, another PR to remove it might be warranted.


albertlast:
well from my test is see no performance diff (forum with 100 boards),
i think you php code is faster but you query is slower as my cte query(which select only 1 row instead of 50 in my case) so they get the same speed on end of the day.

When this pr get merge i would add a pr with the cte version of the board query,
so that modern rdbms can run faster on subboard view.

Sesquipedalian:
> When this pr get merge i would add a pr with the cte version of the board query,
so that modern rdbms can run faster on subboard view.

That sounds like a good plan to me.

Sesquipedalian:
Regarding the setting I've added in ef7456b:

The upgrader will set the `boardindex_max_depth` setting to an initial value of 1 in order to preserve behaviour from SMF 2.0. However, the installer will set it to an initial value of 5. This seemed to me like a reasonable default value to use on a brand new forum. If @jdarwood007 or anyone else has strong feelings about that decision, I'm open to changing it.

PR #4776, opened 2018-05-27
Upgrader Smiley Issues - Fixes #4773

### LINKED ISSUE #4773: Smiley settings problems on upgrade
The logic for adding the 2.1 Alienine & Fugue smiley sets to smf_settings is failing. I believe Fugue is failing because it's the default, and there is already a default. Alienine is failing because of syntax. Note the directories are all there in the filesystem, but the names are confused because the smf_settings values are confused:
image

The logic for updating these values in the upgrader needs to be fixed.

Another problem:  In 2.0, the default was alienine, so in the /default folder are alienine .gifs.   In 2.1, the default is fugue, so in the /default folder are fugue .pngs.  As a result, when installing the 2.1 files on top of the 2.0 files, you now get a mix of alienine & fugue in the /default folder.  I suggest leaving the default at alienine, so the /default folder would include alienine .gifs & .pngs post installation.  

Another consideration...  Some folks may want to just delete the Smileys folder prior to upgrading.  If this were to happen, the old 2.0 values in the settings table would point to missing folders & smileys.  It might be helpful to confirm they exist.  (We did this with themes, over here:  https://github.com/SimpleMachines/SMF2.1/pull/3964/files)

This issue was split from #4767 .

sbulen:
This addresses a # of issues reported over in #4773 : 
- 2.0 smiley settings were doubled up during the 2.1 upgrade
- The /default folder had two completely different smiley sets in it after upgrade
- The smiley_sets_names had strange values like {$default_alienine_smileyset_name} 

Also took steps to ensure alignment between the various smiley settings (e.g., no set name without a folder) and the file system (don't migrate old Smileys no longer in the file system).  

Also got rid of the /default folder.  It was just an alias for the alienine set - use its real name going forward.

Tested on wamp & on unix, both installs & upgrades, with custom smiley sets & without, etc., etc...


sbulen:
@Sesquipedalian - This can easily be enhanced to store extensions.  

I don't know if you want another setting to include extensions (e.g., smiley_sets_types = '.gif, .gif, .png')?   

Or if you'd like a json-encoded set of values in smiley_sets_known (e.g., smiley_sets_known = {"alienine":".png","fugue":".png","aaron":".gif","classic":".gif"}).

Let me know your preference, or if you have another approach in mind (DB?).  I can do the installer & upgrader changes if needed.

other/upgrade_2-1_mysql.sql
```diff
@@ -2609,34 +2609,52 @@ WHERE
...
+// Confirm they exist in the filesystem; bypass default which is getting removed...
+foreach ($dirs AS $ix => $dir)
+	if (is_dir($modSettings['smileys_dir'] . '/' . $dir . '/') && !empty($setnames[$ix]) && $dir != 'default')
+		$combined[$dir] = $setnames[$ix];
+
+// Add our lovely new smiley sets if not already there...
+if (!array_key_exists('fugue', $combined))
+	$combined['fugue'] = $txt['default_fugue_smileyset_name'];
+if (!array_key_exists('alienine', $combined))
+	$combined['alienine'] = $txt['default_alienine_smileyset_name'];
```
jdarwood007, line 2628: Won't this break if $txt is in a different language and these are translated?

sbulen: No, I don't think so.  That is just the user friendly name for the dropdown menu.  Tests fine.  (It's not the key, it's the value...)

What *did* concern me about using $txt for the names had more to do with extensibility.  Specifically, does this mean that every new custom smiley set needs to add a $txt value for its name? Where? Or is this specific to the SMF default smileysets?  How would a smileyset creator enable language support?   (Or do I worry too much about such things?)

$txt usage predated this PR & I tried to work with the existing paradigm.  I do think it will work if we deem that $txt smileyset names are for SMF sets only.  (They are *only* used during install/upgrade.)

The only other question I had was about the deletion of the old /default folder.   I really like cleaning up after things...   I know SMF folks don't necessarily like to do so - it's safer to leave it laying around.  So...  I guess my question is whether the deletion step can stay (my preference) or if it should go (more SMF-ish...).  Hope this makes sense...

jdarwood007: Ok thats fine then.
I would say it should only delete the default files.  Leaving any customized ones alone.

sbulen: OK - then I'll leave it as-is.

jdarwood007:
@sbulen How does your code here work for a fresh install of 2.1 that we then run the upgrader on?

sbulen:
I tested installs & upgrades, wamp & lamp.  Keeping old smilies & deleting them.  About 8 combos.  

All tested ok.

sbulen:
I just retested the beta-to-beta scenario, given there have been a lot of changes.  Tested OK.  Properly cutover to the new directory structure.   But you MUST run the upgrader for the new logic to bring the smiley settings in sync with the file system.

PR #4777, opened 2018-05-30
Added mb4 changeable by Settings.php

albertlast:
Based on this thread: https://www.simplemachines.org/community/index.php?topic=560579.0
i came with the idea up,
what should we do when someone change to mb4.
My guesss is that to change the default behavior by settings.php would be the easiest approache.


frandominguezl (approved):

colinschoen (approved):
LGTM

PR #4781, opened 2018-06-03
More space in top_menu

dragomano:
It is needed for Russian language, as example.
normal_view

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -846,7 +846,7 @@ img.sort, .sort {
 }
 /* The profile/pm menus are declared off .dropmenu li ul for consistency but have other characteristics. */
 .top_menu {
-	width: 25em;
+	width: 30em;
 }
 .top_menu .login {
 	width: 100%;


wintstar:
For german to small

profil-menue

albertlast:
In german case i would say you could use a different word like "Nachrichten" or "Meldungen" or "Mitteilungen" etc.

Sesquipedalian:
We should really just use a single column of variable width rather than two columns of fixed width. Then the differing widths of the menu items in different localizations will never cause problems.

Sesquipedalian:
@Gwenwyfar

Gwenwyfar:
I'm aware of the issue. The popup itself has to be a fixed size because it is absolute positioned, which makes things a little more complicated because making two columns is also "fixed" at 50% (else it looks messy). Might be able to throw some flexboxes into it to make it dynamic. Increasing the fixed size doesn't solve much.

Sesquipedalian:
I was procrastinating for a minute or ten, and tried out an idea. It seems to have worked. 🙂 #4782

Sesquipedalian:
Thanks for submitting this, @dragomano. Although we decided to go with the the changes in #4782 instead, your pull request was instrumental in bringing the problem to our attention. 🙂

PR #4782, opened 2018-06-04
Improves layout of profile menu when it contains long words

Sesquipedalian:


Gwenwyfar:
I'll have to say I don't think that's a very aesthetically-pleasing solution, but it does work fine to prevent broken language content.

Also: If content goes over the min-width, the icon is still split from the word in two lines, but maybe there aren't any words that long.

Sesquipedalian:
I had a better idea. This version uses CSS columns and hyphens so that we can continue to have two columns but not have to worry about long words breaking things.

Here's what it looks like with a long word in the profile links when hyphenation is supported:
screen shot 2018-06-03 at 10 02 16 pm

To simulate what will happen when hyphenation is not supported for a given language, here's a version with hyphenation disabled:
screen shot 2018-06-03 at 10 02 51 pm

In short, this approach ensures that the profile menu expands to the necessary width. Do you like this version better, @Gwenwyfar?

Gwenwyfar:
Looks good to me :)

I'm only trying to figure out now why the columns are misaligned, even when everything uses one line. Seems to be because of the top-margin, using only bottom it aligns correctly.

And if you'll allow me to be nitpicky, the width of the icons or the margin could change and you now have a new place tied up to that (not so good for theming). But I think it's good enough as it is, we could do a fix after the other pending PR, there may be other places that will need the same fix and we can get them all together, and removing a fixed size will be easier with some changes there.

Sesquipedalian:
> I'm only trying to figure out now why the columns are misaligned, even when everything uses one line. Seems to be because of the top-margin, using only bottom it aligns correctly.

Good catch. It was indeed due to the margins on the list items. But even using just a bottom margin I could still end up with misaligned list items in some cases. Apparently this is because margins can be broken across columns, so that, e.g., the bottom margin of an item right at the very bottom of a column might spill over to the top of the next column. I think that's counterintuitive behaviour, but apparently the people in charge of these things decided that this is how it should work, so... 😒

Anyway, I found that if I use line-spacing instead of margins, everything stays lined up nicely even when some list items contains more than one line of text. Which makes sense, if you think about it.

screen shot 2018-06-04 at 3 28 25 pm

Sesquipedalian:
> And if you'll allow me to be nitpicky, the width of the icons or the margin could change and you now have a new place tied up to that (not so good for theming).

Are you referring to the text-indent?

Gwenwyfar:
Oh, silly but fully intentional behaviors abound, if you look closely :unamused:

> Are you referring to the text-indent?

Yep.

Gwenwyfar (approved):
Looks good to merge.

PR #4785, opened 2018-06-13
Send out an approval confirmation

jazzl0ver:
Please, merge my PR


frandominguezl:
Commits need to be signed before merging, as follows:

> It is enough to include in your commit comment "Signed-off by: " followed by your name and email address (for example: Signed-off-by: Angelina Belle <angelinabelle1@hotmail.com>)

jazzl0ver:
Just signed it. Please, check if it's ok

frandominguezl:
The commit now looks good, but the feature is hardcoded and I'm not sure if the developers want it that way.

jazzl0ver:
Well.. In my opinion, it's quite logical. When a forum requires a registration approval, a user receives a letter with the following text (_register_pending_ template):
```
...
Before you can login and start using the forum, your request will be reviewed and approved.  When this happens, you will receive another email from this address.
```
So, the user is awaiting for the approval confirmation, which is not going to be sent by default. I believe this is wrong.

frandominguezl (approved):

Sources/Profile-Actions.php
```diff
@@ -53,6 +53,20 @@ function activateAccount($memID)
...
+			'USERNAME' => $user_profile[$memID]['member_name'],
+			'PROFILELINK' => $scripturl . '?action=profile;u=' . $memID,
+			'FORGOTPASSWORDLINK' => $scripturl . '?action=reminder',
+		);
+
+		require_once($sourcedir . '/Subs-Post.php');
+		loadLanguage('Login');
+
+		$emaildata = loadEmailTemplate('admin_approve_accept', $replacements, $language);
+		sendmail($user_profile[$memID]['email_address'], $emaildata['subject'], $emaildata['body'], null, null, false, 0);
```
colinschoen: `sendmail($user_profile[$memID]['email_address'], $emaildata['subject'], $emaildata['body'], null, null, false, 0);`

Should be

`sendmail($user_profile[$memID]['email_address'], $emaildata['subject'], $emaildata['body'], null, 'accapp' . $memID, $emaildata['is_html'], 0);`

colinschoen (changes requested):

Sources/Profile-Actions.php
jazzl0ver: changed

frandominguezl:
Looks like you forgot to sign the commit.

jazzl0ver:
Yeah, sorry about that. Is it better now?

frandominguezl:
Travis CI will likely complain because your file edits contain extra spaces at the end:

screenshot_1

edit: And yes, it complained

jazzl0ver:
fixed. however, this is kinda ridiculous to complain for the final CR

Sources/Profile-Actions.php
```diff
@@ -969,4 +984,4 @@ function subscriptions($memID)
 		$context['sub_template'] = 'user_subscription';
 }
 
-?>
```
frandominguezl: Nope, seems like it wasn't fixed. Travis will complain again...

frandominguezl:
That one last commit will make it 👍

jazzl0ver:
sorry for the mess. first tries were to remove the CR via web - no success

frandominguezl:
Don't worry, that has happened to me a few times ^^

jdarwood007:
Normally this isn't needed because when the admin uses the normal function from the admin center, it asks if you want to email or not. See ManageMembers.php Line 1156

This only gets called if you click the link from the users profile. What should be done is a secondary link that says approve and send email.  That way the original method is still retained.

jazzl0ver:
@jdarwood007 I don't agree with you. Users are really getting confused b/c they don't get a confirmation  they were told about in the initial letter.

jazzl0ver:
You know, this is kinda strange. I've fixed an important issue which users are facing every day. This fix does not break anything - it just implements the expected behavior. Nobody will complain if users start receiving confirmation letters they were told about. Why don't you guys want to merge the fix?

jdarwood007:
It fixes what sometimes is expected.  Sometimes the admin may want to approve the profile and not send a email.  Sometimes they may want to approve and send the email.  Both could be expected.

@Sesquipedalian, @MissAllSunday and @colinschoen may be able to weigh in on if this should merge or not.

jazzl0ver:
> Sometimes the admin may want to approve the profile and not send a email
@jdarwood007 can you pls show at least a single reason why admin might want not to send the approval to a user after approving his/her account (and knowing that user is expecting the approval)? this sounds ridiculous, sorry.

MissAllSunday:
Indeed I can see instances where an admin would not want to send an email.  We already got both options, if you want to send an email, use the admin function, if not, use the link in the profile page.

I also could argue that the admin function already has both options too.  Perhaps the link in the profile page should re-direct to the admin page instead of duplicating functionality.

jazzl0ver:
@MissAllSunday I know you don't have to but can you explain me what might be a reason for an admin not to send an approval email after a user got a letter, telling him that he **will get** the approval? Why an admin might want to lie to his/her users?

MissAllSunday:
You aren't lying to anyone. It might be something as simple as you telling the user via something else, whatsapp, fb messenger, dunno that their account has been approved. An account might be a dummy/clone/subaccount that does not require sending an email. Your forum can be hosted on a crappy shared hosting plan with no access (or very limited) to sending mails, etc.

**But thats not the point**. I'm not against telling the user about their approval status... I'm saying theres already a functionality for that, if you want to tell the user then use that functionality.

I think it is redundant to have two places where an user gets approved, the admin one seems fine to me and already offers the option to send out an email if the admin chooses to do so.

jazzl0ver:
> I'm saying theres already a functionality for that, if you want to tell the user then use that functionality.

When you open a user account that's awaiting for an approval, you have a single link "approve" which does not send out email:
image
Do you think the better way is to add one more link which will send out an approval?

MissAllSunday:
I think a better way is to use that link to redirect the user to the admin setting for approve new registrations:

image

No need to duplicate functionality.

jdarwood007:
Sounds great to me.  We can do some slight adjustments to the search member function so the link can pass the member id directly and present a single search page with that member.  The search member functionality is there, just needs tweaked to handle this.

Then all the related code could be dropped from the profile section, letting the admin section handle it.

jazzl0ver:
While the redirection resolves the issue (and I'm glad it's gonna be resolved somehow!), I don't think it's a good way to resolve, since it will require extra clicks (= extra time) for an admin to work on the user. Why not just copy that pop-down menu from the Manage Members page to the Summary page? Thus we save time for an admin and keep all possible options.

MissAllSunday:
Because if we do that then we will have to do the same thing for any other user-related admin setting. Besides, there would be no extra clicks involved, once the admin reaches the approval section s/he can approve as many users as they want from there. It is actually a lot more time consuming to go to each users profile page and click the link.

Approving an user is an admin setting and it belongs in the admin section.

jazzl0ver:
> It is actually a lot more time consuming to go to each users profile page and click the link.

@MissAllSunday I agree, in general. But this particular situation we faced is that we get to that Summary page right from the letter which admin receives when a new user is registered.

MissAllSunday:
Then add a link to the approval section in the mail along side keeping the existing one.

Sesquipedalian:
> ... When this happens, you will receive another email from this address.

It seems to me that this sentence should be removed from the email template text. Unless we remove the admin's option to approve without sending an email, it will always be possible that this statement will prove false.

Past that, it seems to me that @jazzl0ver and @MissAllSunday are both right. The default behaviour should be to send the approval notification unless the admin explicitly chooses not to. But it also makes the most sense for the link on the profile page to redirect the admin to the ACP. The latter would make the former moot.

So here's what I'd like to see in this PR instead:

1. That sentence gets removed.
2. The link on the profile page is changed into a link to `?action=admin;area=viewmembers;sa=browse;type=approve;`
3. The order of items in the admin's select menu on that page is changed to "Approve and send email," "Approve (no email)," ...

jazzl0ver:
> The default behaviour should be to send the approval notification unless the admin explicitly chooses not to. 

Absolutely! Thanks God - somebody shares my point of view! :)

Sesquipedalian (changes requested):
Here's what I'd like to see in this PR instead:

1. Remove "When this happens, you will receive another email from this address" from the email template.
2. Change the link on the profile page into a link to `?action=admin;area=viewmembers;sa=browse;type=approve;`
3. Change the order of items in the admin's select menu on that page to "Approve and send email," "Approve (no email)," ...

jazzl0ver:
@Sesquipedalian any chances you could implement that functionality by yourself? Version 2.0.15 (I'm currently running) has that kind of things in different files rather the one on github and I don't really have time to install the github's version for testing and fixing. The fix you suggest is completely different from the one I supplied.
I can create a new issue if it's required.

jdarwood007:
2.0 wont receive requests such as this.  2.0 is feature locked and security and major bug fixes only.

MissAllSunday:
I'll take a look at it, will push to this same branch so keep it open please.

PR #4786, opened 2018-06-14
Small fix

dragomano:
Height fix for input type="file" elements
file_input_auto


frandominguezl:
Commits need to be signed before merging, as follows:

> It is enough to include in your commit comment "Signed-off by: " followed by your name and email address (for example: Signed-off-by: Angelina Belle <angelinabelle1@hotmail.com>)

PR #4787, opened 2018-06-15
Height fix for input type="file" elements

dragomano:


colinschoen:
Can you describe the visual change this makes.

dragomano:
The input element does not fit completely due to the hard-coded height. I see this problem only on Firefox.
current_view_of_input_button
normal_view

colinschoen (approved):
LGTM

PR #4788, opened 2018-06-15
Raise pg min version to 9.4

albertlast:
Postgresql 11 is now in the beta phase,
so it get release in the next months (before smf 2.1 get released).

Time for our yearly raise of supported pg version like the last one https://github.com/SimpleMachines/SMF2.1/pull/4201

since pg 9.3 is not well spreaded we can jump directly from 9.2 to 9.4,
i had to check whats is new stuff of 9.3 and 9.4 if ther features are ther which we could use.

This would be a different pr than.

### DIFF
--- a/other/install.php
+++ b/other/install.php
@@ -51,7 +51,7 @@ $databases = array(
 	),
 	'postgresql' => array(
 		'name' => 'PostgreSQL',
-		'version' => '9.2',
+		'version' => '9.4',
 		'function_check' => 'pg_connect',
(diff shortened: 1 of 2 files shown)


jdarwood007:
@colinschoen @Sesquipedalian @Oldiesmann ,

Any thoughts here?

colinschoen:
JSONB looks interesting ([ref](https://wiki.postgresql.org/wiki/What%27s_new_in_PostgreSQL_9.4#JSONB_Binary_JSON_storage))

Raising to min version to 9.4 sounds good to me.

PR #4791, opened 2018-06-23
Installer/upgrader templates revision - part 1

Gwenwyfar:
Getting started on these two, the templates here will need more work than the other ones on the html level. This is just mild formatting, next up will be removing the inline styles/table layout. This revision will also make both adaptive.

### DIFF
--- a/other/install.php
+++ b/other/install.php
@@ -1897,74 +1897,75 @@ function template_install_above()
 
 	echo '<!DOCTYPE html>
 <html', $txt['lang_rtl'] == true ? ' dir="rtl"' : '', '>
-	<head>
-		<meta charset="', isset($txt['lang_character_set']) ? $txt['lang_character_set'] : 'UTF-8', '">
-		<meta name="robots" content="noindex">
-		<title>', $txt['smf_installer'], '</title>
-		<link rel="stylesheet" href="Themes/default/css/index.css?alp21">
-		<link rel="stylesheet" href="Themes/default/css/install.css?alp21">
-		', $txt['lang_rtl'] == true ? '<link rel="stylesheet" href="Themes/default/css/rtl.css?alp21">' : '', '
+<head>
+	<meta charset="', isset($txt['lang_character_set']) ? $txt['lang_character_set'] : 'UTF-8', '">
+	<meta name="robots" content="noindex">
+	<title>', $txt['smf_installer'], '</title>
+	<link rel="stylesheet" href="Themes/default/css/index.css?alp21">
+	<link rel="stylesheet" href="Themes/default/css/install.css?alp21">
+	', $txt['lang_rtl'] == true ? '<link rel="stylesheet" href="Themes/default/css/rtl.css?alp21">' : '', '
 
-		<script src="Themes/default/scripts/jquery-3.2.1.min.js"></script>
-		<script src="Themes/default/scripts/script.js"></script>
-	</head>
-	<body><div id="footerfix">
-		<div id="header">
-			<h1 class="forumtitle">', $txt['smf_installer'], '</h1>
-			<img id="smflogo" src="Themes/default/images/smflogo.svg" alt="Simple Machines Forum" title="Simple Machines Forum">
-		</div>
-		<div id="wrapper">
-			<div id="upper_section">
-				<div id="inner_section">
-					<div id="inner_wrap">';
+	<script src="Themes/default/scripts/jquery-3.2.1.min.js"></script>
+	<script src="Themes/default/scripts/script.js"></script>
+</head>
+<body>
+	<div id="footerfix">
+	<div id="header">
+		<h1 class="forumtitle">', $txt['smf_installer'], '</h1>
+		<img id="smflogo" src="Themes/default/images/smflogo.svg" alt="Simple Machines Forum" title="Simple Machines Forum">
+	</div>
+	<div id="wrapper">
+		<div id="upper_section">
+			<div id="inner_section">
+				<div id="inner_wrap">';
 
 	// Have we got a language drop down - if so do it on the first step only.
 	if (!empty($incontext['detected_languages']) && count($incontext['detected_languages']) > 1 && $incontext['current_step'] == 0)
 	{
 		echo '
-						<div class="news">
-							<form action="', $installurl, '" method="get">
-								<label for="installer_language">', $txt['installer_language'], ':</label>
-								<select id="installer_language" name="lang_file" onchange="location.href = \'', $installurl, '?lang_file=\' + this.options[this.selectedIndex].value;">';
+					<div class="news">
+						<form action="', $installurl, '" method="get">
+							<label for="installer_language">', $txt['installer_language'], ':</label>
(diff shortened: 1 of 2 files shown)


Gwenwyfar:
And that's most of all formatting for upgrader done for. The remaining for installer will come later.

other/install.php
```diff
@@ -1897,74 +1897,75 @@ function template_install_above()
...
-		</div>
-		<div id="wrapper">
-			<div id="upper_section">
-				<div id="inner_section">
-					<div id="inner_wrap">';
+<head>
+	<meta charset="', isset($txt['lang_character_set']) ? $txt['lang_character_set'] : 'UTF-8', '">
+	<meta name="robots" content="noindex">
+	<title>', $txt['smf_installer'], '</title>
+	<link rel="stylesheet" href="Themes/default/css/index.css?alp21">
```
XinYenFon, line 1904: ?alp21 needs to be dynamic check loadCSSFile function (same applies the next line as well).

other/install.php
```diff
@@ -2127,30 +2129,32 @@ function template_chmod_files()
 					<td width="26%" valign="top" class="textbox"><label for="ftp_server">', $txt['ftp_server'], ':</label></td>
 					<td>
 						<div style="float: ', $txt['lang_rtl'] == false ? 'right' : 'left', '; margin-', $txt['lang_rtl'] == false ? 'right' : 'left', ': 1px;"><label for="ftp_port" class="textbox"><strong>', $txt['ftp_port'], ':&nbsp;</strong></label> <input type="text" size="3" name="ftp_port" id="ftp_port" value="', $incontext['ftp']['port'], '" /></div>
-						<input type="text" size="30" name="ftp_server" id="ftp_server" value="', $incontext['ftp']['server'], '" style="width: 70%;" />
+						<input type="text" size="30" name="ftp_server" id="ftp_server" value="', $incontext['ftp']['server'], '" style="width: 70%;">
```
XinYenFon, line 2132: Inline width should be moved to install.css (ignoring the following input's inline style).

other/install.php
```diff
@@ -2202,7 +2206,7 @@ function template_database_settings()
 			<tr id="db_server_contain">
 				<td width="20%" valign="top" class="textbox"><label for="db_server_input">', $txt['db_settings_server'], ':</label></td>
 				<td>
-					<input type="text" name="db_server" id="db_server_input" value="', $incontext['db']['server'], '" size="30" /><br>
+					<input type="text" name="db_server" id="db_server_input" value="', $incontext['db']['server'], '" size="30"><br>
```
XinYenFon, line 2209: Input sizes should be consistent and predefined for all (ignoring rest size issues).

other/install.php
```diff
@@ -2378,15 +2382,15 @@ function template_populate_database()
 	if (!empty($incontext['failures']))
 	{
 		echo '
-				<div style="color: red;">', $txt['error_db_queries'], '</div>
-				<ul>';
+		<div style="color: red;">', $txt['error_db_queries'], '</div>
```
XinYenFon, line 2385: All main colors should be predefined in index.css, if not should be added (ignoring rest inline-coloring).

other/upgrade.php
```diff
@@ -76,7 +76,7 @@
 
 // The helper is crucial. Include it first thing.
 if (!file_exists($upgrade_path . '/upgrade-helper.php'))
-    die('upgrade-helper.php not found where it was expected: ' . $upgrade_path . '/upgrade-helper.php! Make sure you have uploaded ALL files from the upgrade package. The upgrader cannot continue.');
+	die('upgrade-helper.php not found where it was expected: ' . $upgrade_path . '/upgrade-helper.php! Make sure you have uploaded ALL files from the upgrade package. The upgrader cannot continue.');
```
XinYenFon, line 79: Why did you added extra tab here? I think its unnecessary

other/upgrade.php
```diff
@@ -3374,9 +3378,9 @@ function move_db_last_error_to_cachedir()
 	}
 }
 
-/******************************************************************************
-******************* Templates are below this point ****************************
-******************************************************************************/
+/* ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+                        Templates are below this point
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ */
```
XinYenFon, line 3383: You should follow SMF's comment standard in here. While I understand the reasoning consistency should be fixed.

other/upgrade.php
```diff
@@ -3581,17 +3587,17 @@ function updateStepProgress(current, max, overall_weight)
 	$mins = (int) ($elapsed / 60);
 	$seconds = $elapsed - $mins * 60;
 	echo '
-								<br> <br> <br> <br> <br>
-								<div class="smalltext" style="padding: 5px; text-align: center;"><br>', $txt['upgrade_time_elapsed'], ':
-									<span id="mins_elapsed">', $mins, '</span> ', $txt['upgrade_time_mins'], ', <span id="secs_elapsed">', $seconds, '</span> ', $txt['upgrade_time_secs'], '.
-								</div>';
+				<br> <br> <br> <br> <br>
```
XinYenFon, line 3590: I think this should be converted to margin instead of many following br's.

other/upgrade.php
```diff
@@ -3600,70 +3606,71 @@ function template_upgrade_below()
...
+						</div>';
 
 	echo '
-								<div class="righttext" style="margin: 1ex;">';
+						<div class="righttext" style="margin: 1ex;">';
 
 	if (!empty($upcontext['continue']))
 		echo '
-									<input type="submit" id="contbutt" name="contbutt" value="', $txt['upgrade_continue'], '"', $upcontext['continue'] == 2 ? ' disabled' : '', ' class="button">';
+						<input type="submit" id="contbutt" name="contbutt" value="', $txt['upgrade_continue'], '"', $upcontext['continue'] == 2 ? ' disabled' : '', ' class="button">';
```
XinYenFon, line 3629: 'name' values should be removed as they are obsolete.

XinYenFon (changes requested):

Gwenwyfar:
I'm not adding any new changes to this PR. It is what it is: mostly formatting. I'll have some of those suggestions in mind for the next one(s) :)

other/upgrade.php
Gwenwyfar: Look more closely ;)

other/upgrade.php
Gwenwyfar: If we are going by overall consistency the whole section is not consistent, and it is also non-existent in installer. Maybe removing it is a better choice, or leave an inline comment on both files.

jdarwood007:
PR was tested and I don't see anything broken and the PR is ok.

@Gwenwyfar Will you be able to fix up the remaining issues Antes mentioned in a future PR?

Gwenwyfar:
Yes, I already have some changes in local for part 2. Just splitting them so we don't have a giant PR to review.

PR #4794, opened 2018-06-24
 cte support added for board index 

albertlast:
It's a combination of @Sesquipedalian pr https://github.com/SimpleMachines/SMF2.1/pull/4775 and
my pr https://github.com/SimpleMachines/SMF2.1/pull/4774

I was waiting that the pr from seq get merge but it take to mutch time,
so i clone his pr and added my cte stuff around this.

It do the same as before:
When we are not on the index page and we got a database how support cte -> run the cte query version.


Sources/Subs-Db-mysql.php
```diff
@@ -1024,4 +1025,29 @@ function smf_db_native_replace()
...
+ * @return boolean true or false
+ */
+function smf_db_cte_support()
+{
+	global $smcFunc;
+	static $return;
+
+	if (!empty($return))
+		return $return;
+
```
jdarwood007: Nit picking this, but if $return is false because support doesn't exist, then this check is always ignored and the efficiency is lost.

Sources/Subs-BoardIndex.php
```diff
@@ -43,35 +43,81 @@ function getBoardIndex($boardIndexOptions)
...
-			LEFT JOIN {db_prefix}log_boards AS lb ON (lb.id_board = b.id_board AND lb.id_member = {int:current_member})') . '
-		WHERE {query_see_board}' . (empty($boardIndexOptions['countChildPosts']) ? (empty($boardIndexOptions['base_level']) ? '' : '
-			AND b.child_level >= {int:child_level}') : '
-			AND b.child_level BETWEEN ' . $boardIndexOptions['base_level'] . ' AND ' . ($boardIndexOptions['base_level'] + 1)) . '
-			ORDER BY ' . (!empty($boardIndexOptions['include_categories']) ? 'c.cat_order, ' : '') . 'b.child_level, b.board_order',
-		array(
-			'current_member' => $user_info['id'],
-			'child_level' => $boardIndexOptions['base_level'],
-			'blank_string' => '',
-		)
```
jdarwood007: You call {int:id_parent} but don't pass it in here.

albertlast:
Thanks for spotting this issues,
got fixed now.

jdarwood007:
Seems to work ok for me.  @Sesquipedalian will need to approve a change like this.

albertlast:
Keep waiting for you approvment @Sesquipedalian 
like the other pr from me

frandominguezl:
He just started reviewing pull requests. Just be patient.

Sesquipedalian (approved):
Nice stuff.

PR #4795, opened 2018-06-25
Follows the user's email notification settings correctly

### LINKED ISSUE #4783: Sending email notifications is unreliable
For example, if a user is set up to receive email notifications immediately (but only for the first unread message), they may or may not actually be sent an immediate notification. Moreover, they may be sent a daily or weekly digest email even though that is not the setting they have selected.

I haven't yet been able to discover a pattern for how and why this happens, but it happens frequently.

This should be fixed before RC1.

Sesquipedalian:
Fixes #4783


colinschoen (approved):
LGTM

Nice fix!

Sources/ScheduledTasks.php
```diff
@@ -644,13 +644,14 @@ function scheduled_daily_digest()
...
+	$members_sent = array();
 	foreach ($members as $mid => $member)
 	{
-		$frequency = !empty($prefs[$mid]['msg_notify_pref']) ? $prefs[$mid]['msg_notify_pref'] : 1;
+		$frequency = isset($prefs[$mid]['msg_notify_pref']) ? $prefs[$mid]['msg_notify_pref'] : 0;
 		$notify_types = !empty($prefs[$mid]['msg_notify_type']) ? $prefs[$mid]['msg_notify_type'] : 1;
 
 		// Did they not elect to choose this?
-		if ($frequency == 4 && !$is_weekly || $frequency == 3 && $is_weekly || $notify_types == 4)
+		if ($frequency < 3 || $frequency == 4 && !$is_weekly || $frequency == 3 && $is_weekly || $notify_types == 4)
```
albertlast, line 654: Could you add some () to make it easier to read?

PR #4800, opened 2018-06-28
Binary clock icons are reversed

Archer70:
One would reasonably assume that blue is "on" and white is "off," but that's not how it is.

imgur

In the above image, reading the blue icons, the time would be 30:49:46 or 210:49:46, depending on how you want to add the second column. Reading the white icons yields a much more reasonable time of 15:36:39.

This affects both 2.0 and 2.1. The screenshot is from https://www.simplemachines.org/community/index.php?action=clock

### DIFF
--- a/Sources/Calendar.php
+++ b/Sources/Calendar.php
@@ -649,8 +649,8 @@ function clock()
 {
 	global $smcFunc, $settings, $context, $scripturl;
 
-	$context['onimg'] = $settings['images_url'] . '/bbc/bbc_bg.png';
-	$context['offimg'] = $settings['images_url'] . '/bbc/bbc_hoverbg.png';
+	$context['onimg'] = $settings['images_url'] . '/bbc/bbc_hoverbg.png';
+	$context['offimg'] = $settings['images_url'] . '/bbc/bbc_bg.png';
 
 	$context['page_title'] = 'Anyone know what time it is?';
 	$context['linktree'][] = array(


wintstar:
Text is hardcoded, can not translated.

https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Calendar.php#L655

https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Calendar.php#L658

https://github.com/SimpleMachines/SMF2.1/blob/9259a92335c8f2920ace6b738111d56bb550536a/Themes/default/Calendar.template.php#L950

https://github.com/SimpleMachines/SMF2.1/blob/9259a92335c8f2920ace6b738111d56bb550536a/Themes/default/Calendar.template.php#L1050

https://github.com/SimpleMachines/SMF2.1/blob/9259a92335c8f2920ace6b738111d56bb550536a/Themes/default/Calendar.template.php#L1028

https://github.com/SimpleMachines/SMF2.1/blob/9259a92335c8f2920ace6b738111d56bb550536a/Themes/default/Calendar.template.php#L1121

Archer70:
> Text is hardcoded, can not translated.

Wasn't that done on purpose because it's an Easter egg?

wintstar:
I think "Easter eggs" should also be "enjoyed" in own language. Otherwise it doesn't make sense.

Archer70:
If I can get a confirmation from a dev that they want Easter egg strings translated, I don't have a problem making those changes pertaining to the clock. Until then, I going to assume it was done this way on purpose, in the same way that "The Book of Unknown" isn't internationalized. :)

colinschoen:
Easter eggs can be internationalized, that is fine. This isn't something we need to go back and change existing ones, however.

Sesquipedalian:
I would not consider this a priority.

NanoSector:
Can we at least keep this issue open even if it is low priority? This should be a quick fix anyway.

Gwenwyfar:
I agree with Nano, it's such a minor change.

Sesquipedalian:
If someone wants to submit a PR to fix it, by all means do so. :) I just don't see a need to clutter up our issues list any more than necessary.

Archer70:
> If someone wants to submit a PR to fix it, by all means do so. :)

You may notice a little "Merge" button below. :)

Sesquipedalian:
Derp. That's what I get for reading this on my phone. I honestly thought I was reading a bug report, not a pull request. 😅

Archer70:
No problem! Glad it was just a misunderstanding. :smile:

PR #4805, opened 2018-07-02
Small fix for hook (integrate_forum_stats)

dragomano:
integrate_forum_stats hook doesn't loaded if $_SESSION['expanded_stats'] is empty.


colinschoen (approved):
LGTM

PR #4807, opened 2018-07-03
Address problem with upgrading http forum via https - fixes #4790

### LINKED ISSUE #4790: Upgrader generating http urls while in https mode
Some content is being blocked by the browser in the upgrader because of http urls. Seems to be only the css and js urls.

sbulen:
Bring default_theme_url in line with how the upgrader was invoked.
Prevents .css & .js from disappearing during your upgrade, causing formatting havoc.
Note fetchSMfiles stepped on it, which is why it needed to be set in 2 places.
Tested in wamp & linux.


jdarwood007:
👍

PR #4809, opened 2018-07-03
Update help text for custom profile fields

### LINKED ISSUE #4784: Regex masks in custom fields aren't delimited
When registering or editing a regex-masked custom field, the following error is logged:
`preg_match(): No ending delimiter '^' found`

It originates from [Register.php](https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Register.php#L470) and [Profile-Modify.php](https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Profile-Modify.php#L1231)

Technically, I could simply surround the regex with `~`, but it'd be better if it was done in code.
```
preg_match('~' . substr($row['mask'], 5) . '~', $value)
```

Archer70:
Fixes #4784 
> Fix the bloody help section by making the delimiter requirement absurdly verbose


colinschoen (approved):
LGTM

PR #4814, opened 2018-07-04
Fixes #4624

### LINKED ISSUE #4624: messages menu item
the messages menu item when within the inbox takes you to writing a new PM, i think it should take you back to the inbox instead. more then likely you want to view your existing PM's rather then trying to write a new PM each time.

frandominguezl:


jdarwood007:
Isn't this already in the code?
https://github.com/SimpleMachines/SMF2.1/blame/release-2.1/Sources/PersonalMessage.php#L238

frandominguezl:
What I'm doing is to remove the sa=send part, which addresses the linked issue.

PR #4821, opened 2018-07-07
Cut error url after 500 chars

albertlast:
While doing "stuff" i created a error which resulted in a endless loop,
the issue of the endless loop was that the $error length was bigger than the url was long,
which created the next error, which created the next error again...

To cut throw this i limited the error length at 500 chars,
because the field is 1024 chars long.

Alternative solution for pg would be to convert this varchar field to jsonb.


albertlast:
Maybe the error message shouldn't attached to the url json
since we got already a error log which handle the stuff already...

PR #4819, opened 2018-07-06
Less logic better result for alert

albertlast:
fixes : https://github.com/SimpleMachines/SMF2.1/issues/4801
please test this.

PR #4834, opened 2018-07-13
Fix undefined indexes from user_info overrides.

jdarwood007:
This reverts #4754
Fixes #4288 and #4287
Related to #4819 
I wasn't able to reproduce the original issue (https://www.simplemachines.org/community/index.php?topic=559941.0) even after just reverting #4754.  I believe #4759 fixed the original issue.

If anyone can give me the full steps to reproduce the errors from these PR/issues after this PR has been applied, let me know.


albertlast (changes requested):
This pr fix one issue and open the other again,
so i'm not a friend of this.

jdarwood007:
@albertlast has suggested that the alerts should track id_board as its own column.  Which would make more sense and simplify this process.
I think that is a great idea and we can either target that here, or target that in its own PR.

betonkamil:
close if it solves the problem

Sesquipedalian (dismissed):
With this PR, I found that alerts were not being created at all for mentions. I don't know why.

Sesquipedalian:
Okay, the more I try to untangle exactly what bug or bugs this PR is meant to fix, the most twisted up it seems to get. @jdarwood007, can you tell me exactly what the bug is that this is meant to solve? That should help me know what I am trying to test in order to review this properly.

jdarwood007:
Recent commits have removed the undefined indexes.

However I still have concerns over overloading $user_info for handling fetching alerts for another user.  Seems the logic is flawed and broken if we have to overload $user_info to just handle fetching alerts.

I don't care whether this gets merged or not.  I've stated my concerns, but if we don't want to fix that concern then we can move on.

Sesquipedalian:
Okay, so at this point this PR is best understood as giving an alternative to overloading $user_info, because that is potentially dangerous, correct? If so, then I agree that it is better not to overload $user_info for the reasons you've explained in the past, and I will test this PR based on that understanding.

Sources/Subs.php
```diff
@@ -386,13 +386,13 @@ function updateMemberData($members, $data)
...
-					$val .= 'WHEN id_member = ' . $v . ' THEN '. count(fetch_alerts($v, false, 0, array(), false)) . ' ';
+					$val .= 'WHEN id_member = ' . $v . ' THEN '. alert_count($v, false) . ' ';
 				$val = $val . ' END';
 				$type = 'raw';
 			}
 			else
 			{
-				$blub = fetch_alerts($members, false, 0, array(), false);
+				$val = alert_count($members, false);
 				$val = count($blub);
```
albertlast: Should be not this line deleted?

Sources/Profile-Modify.php
```diff
@@ -2253,20 +2253,45 @@ function alert_delete($toDelete, $memID = false)
...
-	global $smcFunc;
+	global $smcFunc, $user_info;
 
 	if (empty($memID))
 		return false;
 
+	// Leave a place holder here for now.
+	// MySQL uses JSON_CONTAINS() to find data, Postgresql seems to be able to natively use it with JSONB column type.
+	// Maybe a $smcFunc['db_json_supported']?
+	if (false)
```
albertlast: How would look the json approached?

Sources/Subs.php
jdarwood007: Yes it should.

Sources/Profile-Modify.php
jdarwood007: It has been so long, I honestly am not 100% sure, but the notes say that postgres and mysql handled json differently.

I don't think its a case for something in 2.1 at this point.  Maybe a future version we can identify a way to natively handle json string searches.

Sesquipedalian (approved):

albertlast:
since some issue appear,
revert this pr and fix this while rc2 phase?
or is this essential for rc1?

PR #4840, opened 2018-07-15
Fixes HTML and BBC in emails sent in plain text

### LINKED ISSUE #4826: Full topic notifications contain BBC, HTML code
Hi all,

I'm not sure if this is related to v2.1 (I'm using Beta 4 RC1 branch install), but when user chooses to receive full message body of the message in notification, if text was formatted all that code is included in a message which doesn't look good at may confuse users and perhaps trigger spam filters. Example:

https://pastebin.com/E0j9NqP0

Is it possible that code could be removed from the notification message?

Thank you!

Reference: https://www.simplemachines.org/community/index.php?P=32d1cc244e18fe39b4a82e122fd7329e;topic=561075.msg3979087#msg3979087

jdarwood007:
Fixes #4826


colinschoen (approved):
LGTM

PR #4841, opened 2018-07-15
Fix a undefined indexes when changing file permissions

jdarwood007:
Reproduce
1. ACP > Packages > File Permissions
2. Choose to use a predefined permission profile (any is fine)
3. Enter in any FTP details, doesn't need to be valid
4. Click Make Changes
5. When the timeout page opens, the undefined indexes are loaded.

As well I noticed the undefined indexes here are not logged.  Not sure why.


colinschoen (approved):
LGTM

PR #4843, opened 2018-07-15
Allow maxchars of 84 when replying to a topic

### LINKED ISSUE #1574: Search result with to many options
If you make a search result and chooses to have the result shown as messages, you have 3 options below each result - "Reply", "Quote" & "Notify".

I really don't see the point of being able to reply directly on a search result. For that matter, the post could be on page 2 of 100 in a topic.
_If_ You choose to click the "Reply" or the "Quote" button, You are redirected to the editor in that topic. So You can't really see what other people replied, hence it's an odd situation to be in.
I know You can see the resent few posts in that topic, but there's no guaranty You can actually get a view about the Whole conversation in the topic.

### LINKED ISSUE #1759: Reply to message exceeds subject maximum length
This issue was already present in versions 2.0.x but I never got around to posting it earlier. 

If a message has a subject that is between 78 and 80 characters long and you reply to it then RE: is added to the subject to make the total length of the subject longer than 80 characters. In version 2.0.x users received an error about exceeding the maximum allowed length and had to manually change it. In my tests with 2.1 I see the error in a flash but it goes away and the message is saved with the longer subject. 

A solution I think would be to trim the subject when RE: is added.

jdarwood007:
Fixes #1759

### DIFF
--- a/Sources/Post.php
+++ b/Sources/Post.php
@@ -1364,7 +1364,7 @@ function Post($post_errors = array())
 	// Gotta have a subject.
 	$context['posting_fields']['subject'] = array(
 		'dt' => '<span id="caption_subject"' . (isset($context['post_error']['no_subject']) ? ' class="error"' : '') . '>' . $txt['subject'] . '</span>',
-		'dd' => '<input type="text" name="subject" value="' . $context['subject'] . '" size="80" maxlength="80" required>',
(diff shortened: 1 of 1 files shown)


jdarwood007:
My  bad, meant #1759

frandominguezl (approved):

colinschoen (approved):
LGTM

PR #4844, opened 2018-07-16
Upgrade can migrate our Settings file to a new version

jdarwood007:
The upgrader allows the migration our Settings.php file to a new version, if determined by a preset conditions that it is allowed to.  These conditions looks for initially integrations and advanced configuration setups and if found we do not allow the migration.  It then determines if variables are missing in the Settings.php that are added in new versions and if so suggests a migration.

The option is configurable during upgrade and can be overridden by the admin.  A copy of the Settings is made called Settings_org.php prior to the migration.  However is upgrade.php is ran multiple times and the option checked multiple times, it is possible to loose the original Settings prior to any migration.

This may require additional information and Settings.php example data to reliably determine if a migration is safe.


wintstar:
Some have also added the debug mode to Settings.php:
`$db_show_debug = true;`

Would it make sense to add this to the function migrateSettingsFile($changes)?

jdarwood007:
If that is detected, we don't enable a migration of the settings file.  I did that as I assumed this is most likely a developer setup and to avoid converting the file.

We could migrate it.  Just is a option that isn't present in the default settings file.

PR #4846, opened 2018-07-16
Alert Moderator if the sticky/lock actions have been taken already

### LINKED ISSUE #2173: conflicting moderation actions
is there a way SMF can warn a moderator when a moderation action was already completed on a thread while you are on the thread?

for example if you are on thread A and moderator is on the same thread and locks the thread before you do, there is nothing to tell you that you have unlocked the thread instead of locking it. same applies to moving a thread.

jdarwood007:
Fixes #2173

### DIFF
--- a/Sources/Topic.php
+++ b/Sources/Topic.php
@@ -60,6 +60,12 @@ function LockTopic()
 	else
 		isAllowedTo('lock_any');
 
+	// Another moderator got the job done first?
+	if (isset($_GET['sa']) && $_GET['sa'] == 'unlock' && $locked == '0')
+		fatal_lang_error('error_topic_locked_already', false);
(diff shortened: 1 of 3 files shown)


Sources/Topic.php
```diff
@@ -60,6 +60,12 @@ function LockTopic()
 	else
 		isAllowedTo('lock_any');
 
+	// Another moderator got the job done first?
+	if (isset($_GET['sa']) && $_GET['sa'] == 'unlock' && $locked == '0')
+		fatal_lang_error('error_topic_locked_already', false);
+	elseif (isset($_GET['sa']) && $_GET['sa'] == 'lock' && ($locked == '1' || $locked == '2'))
+		fatal_lang_error('error_topic_locked_already', false);
```
illori: you are using the same language string here for both, do you mean to unlocked string for one of these?

PR #4848, opened 2018-07-16
Ensure signature image limit don't break the forum

### LINKED ISSUE #2205: (+2.0) Workaround for signature images bigger than allowed size
Steps to reproduce:
(needs a REGULAR user account. Admins seem to always allow images regardless of size)
1. Put in an image of lesser or equal than allowed size. The forum will not apply any resizing (the image passed the size validation).
2. Save the signature.
3. Replace the image with an image of bigger size than allowed size. Do not modify signature. Because the forum previously allowed the image, it will not re-check it and show the bigger image anyway.

Examples:
https://nanosector.ddns.net/smf20/index.php?action=profile;area=summary;u=5
https://nanosector.ddns.net/smf21/index.php?action=profile;area=summary;u=2
Both forums have the max size set at 50x50. And that's clearly not 50x50 :P
(as you can see, it also applies in 2.0)

jdarwood007:
Also, signature settings are stupidly setup here.  Not sure what storing all the settings in a string benefits, just requires more code to figure out settings.

Fixes #2205

### DIFF
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -359,6 +359,17 @@ function Display()
 	$context['signature_enabled'] = substr($modSettings['signature_settings'], 0, 1) == 1;
 	$context['disabled_fields'] = isset($modSettings['disabled_profile_fields']) ? array_flip(explode(',', $modSettings['disabled_profile_fields'])) : array();
 
+	// Prevent signature images from going outside the box.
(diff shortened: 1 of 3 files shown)


colinschoen (approved):
LGTM

PR #4849, opened 2018-07-16
Fix Quick Moderation display issues

jdarwood007:
@Antes and @Gwenwyfar Any issues with this breaking mobile or something else I'm not seeing?


Gwenwyfar:
Looks good to me. The only thing "breaking" is a problem that was already there (when buttons use two lines, the margin is not correct).

PR #4851, opened 2018-07-17
Cleanup

XinYenFon:
Read commit note(s) for info.

### DIFF
--- a/Themes/default/ManageMembers.template.php
+++ b/Themes/default/ManageMembers.template.php
@@ -20,13 +20,13 @@ function template_search_members()
 	echo '
 	<div id="admincenter">
 		<form action="', $scripturl, '?action=admin;area=viewmembers" method="post" accept-charset="', $context['character_set'], '" id="admin_form_wrapper">
+			<input type="hidden" name="sa" value="query">
 			<div class="cat_bar">
 				<h3 class="catbg">
 					<span class="floatleft">', $txt['search_for'], '</span>
 					<span class="smalltext floatright">', $txt['wild_cards_allowed'], '</span>
 				</h3>
 			</div>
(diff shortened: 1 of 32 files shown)


Themes/default/ManageMembers.template.php
```diff
@@ -27,7 +27,7 @@ function template_search_members()
 				</h3>
 			</div>
 			<input type="hidden" name="sa" value="query">
-			<div class="windowbg noup">
+			<div class="windowbg">
```
Gwenwyfar: This input here will break the + sequence.

Themes/default/ManagePaid.template.php
```diff
@@ -26,7 +26,7 @@ function template_modify_subscription()
 
 	if (!empty($context['disable_groups']))
 		echo '
-			<div class="information noup">
+			<div class="information">
```
Gwenwyfar: title + .information + .windowbg will leave .information with bottom rounded borders. Unless we can find a solution for that, it needs to remain there.

Themes/default/PersonalMessage.template.php
```diff
@@ -1046,7 +1046,7 @@ function template_send()
 
 	echo '
 		<form action="', $scripturl, '?action=pm;sa=send2" method="post" accept-charset="', $context['character_set'], '" name="postmodify" id="postmodify" class="flow_hidden" onsubmit="submitonce(this);">
-			<div class="roundframe noup">';
+			<div class="roundframe">';
```
Gwenwyfar: Need to move the title here. Better semantics as a bonus.

Gwenwyfar (changes requested):

Gwenwyfar (approved):

PR #4854, opened 2018-07-18
Inner Join breaks hooks using the attachment system.

jdarwood007:
If a hook is using the attachment system and does not reference or has a id_msg, this will result in no rows being returned.  Two solutions are to use a left join or ignore this when $attachRequest from is a resource.

### DIFF
--- a/Sources/ShowAttachments.php
+++ b/Sources/ShowAttachments.php
@@ -120,7 +120,7 @@ function showAttachment()
 		}
 
 		// Previews doesn't have this info.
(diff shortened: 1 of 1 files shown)


albertlast:
Which select should use a jeft join?

jdarwood007:
The one that is ran after this.  With what i'm doing we dont need to make it a left join, we simply ignore the block of code that is validating if it has a msg id attached to it.  Which integrations may not have

PR #4873, opened 2018-07-26
Fixes hardcoded text strings #4858

### LINKED ISSUE #4858: Hardcoded text strings in upgrader
Self-explanatory. There are still many hardcoded strings left in upgrader.

frandominguezl:
This fixes #4858 

1.  What to do with text that will only appear if **$command_line** is true? For example, [this](https://github.com/SimpleMachines/SMF2.1/blob/6543da8353b96816adfc31cb12db8e7723fab104/other/upgrade.php#L3225) or [this](https://github.com/SimpleMachines/SMF2.1/blob/6543da8353b96816adfc31cb12db8e7723fab104/other/upgrade.php#L3284) _(and a few others)._

Did I miss anyone? :)

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -4413,7 +4413,7 @@ function template_database_changes()
 							if (!attemptAgain)
 							{
 								document.getElementById("error_block").style.display = "";
-								setInnerHTML(document.getElementById("error_message"), "', sprintf($txt['upgrade_repondtime'], ($timeLimitThreshold * 10)), '"<a href="#" onclick="retTimeout(true); return false;">"', $txt['upgrade_respondtime_clickhere'], '"</a>';');
+								setInnerHTML(document.getElementById("error_message"), "', sprintf($txt['upgrade_repondtime'], ($timeLimitThreshold * 10)), '" + "<a href=\"#\" onclick=\"retTimeout(true); return false;\">', $txt['upgrade_respondtime_clickhere'], '</a>");
 							}
 							else
 							{
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -4218,7 +4218,7 @@ function template_database_changes()
 								if (retryCount > 10)
 								{
 									document.getElementById("error_block").style.display = "";
-									setInnerHTML(document.getElementById("error_message"), ', $txt['upgrade_loop'], ' + sDebugName);';
+									setInnerHTML(document.getElementById("error_message"), "', $txt['upgrade_loop'], '" + sDebugName);';
 
 	if ($is_debug)
 		echo '
@@ -4400,7 +4400,7 @@ function template_database_changes()
 							if (!attemptAgain)
 							{
 								document.getElementById("error_block").style.display = "";
-								setInnerHTML(document.getElementById("error_message"), ', sprintf($txt['upgrade_repondtime'], ($timeLimitThreshold * 10)), '<a href="#" onclick="retTimeout(true); return false;">', $txt['upgrade_respondtime_clickhere'], '</a>';');
+								setInnerHTML(document.getElementById("error_message"), "', sprintf($txt['upgrade_repondtime'], ($timeLimitThreshold * 10)), '"<a href="#" onclick="retTimeout(true); return false;">"', $txt['upgrade_respondtime_clickhere'], '"</a>';');
 							}
 							else
 							{
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -447,19 +447,32 @@ function load_lang_file()
 		header('Cache-Control: no-cache');
 
 		echo '<!DOCTYPE html>
-		<html>
-			<head>
-				<title>SMF Upgrader: Error!</title>
-			</head>
-			<body style="font-family: sans-serif;"><div style="width: 600px;">
-				<h1 style="font-size: 14pt;">A critical error has occurred.</h1>
-				<p>This upgrader was unable to find the upgrader\'s language file or files.  They should be found under:</p>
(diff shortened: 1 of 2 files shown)


Gwenwyfar:
There's also lines 586, 788, 1207, 1208, 2104 (whole var), 2247, 3902, 3966, 4091, 4093, 4134, 4234 and 4416. Besides the ones with $command_line you already mentioned.

jdarwood007:
Is this ready yet or more commits to come?

frandominguezl:
I'd like to know if the text that appears when $command_line is true needs to be localized as well. If it isn't, it's ready to go. If it is, I'd need one more commit maybe.

jdarwood007:
It could be, but nobody has complained about it since it was introduced in 2.0 upgrader.  Not many admins use the CLI upgrade, generally only those with large boards that don't want to deal with browser timeouts during the upgrade process.

So fix it if you would like.

frandominguezl:
Thanks for clarifying. There's no need to localize them, at least for now.

You can merge this :)

other/upgrade.php
```diff
@@ -4235,7 +4218,7 @@ function onItemUpdate(oXMLDoc)
 								if (retryCount > 10)
 								{
 									document.getElementById("error_block").style.display = "";
-									setInnerHTML(document.getElementById("error_message"), "Upgrade script appears to be going into a loop - step: " + sDebugName);';
+									setInnerHTML(document.getElementById("error_message"), ', $txt['upgrade_loop'], ' + sDebugName);';
```
jdarwood007: Need Double Quotes here, breaks javascript

other/upgrade.php
```diff
@@ -4417,7 +4400,7 @@ function retTimeout(attemptAgain)
 							if (!attemptAgain)
 							{
 								document.getElementById("error_block").style.display = "";
-								setInnerHTML(document.getElementById("error_message"), "Server has not responded for ', ($timeLimitThreshold * 10), ' seconds. It may be worth waiting a little longer or otherwise please click <a href=\"#\" onclick=\"retTimeout(true); return false;\">here<" + "/a> to try this step again");
+								setInnerHTML(document.getElementById("error_message"), ', sprintf($txt['upgrade_repondtime'], ($timeLimitThreshold * 10)), '<a href="#" onclick="retTimeout(true); return false;">', $txt['upgrade_respondtime_clickhere'], '</a>';');
```
jdarwood007: Need Double Quotes here, breaks javascript

jdarwood007 (changes requested):

other/upgrade.php
```diff
@@ -447,35 +447,19 @@ function load_lang_file()
...
-		<p>This upgrader was unable to find the upgrader\'s language file or files.  They should be found under:</p>
-
-		<div class="directory">', dirname($_SERVER['PHP_SELF']) != '/' ? dirname($_SERVER['PHP_SELF']) : '', '/Themes/default/languages</div>
-
-		<p>In some cases, FTP clients do not properly upload files with this many folders. Please double check to make sure you <strong>have uploaded all the files in the distribution</strong>.</p>
-		<p>If that doesn\'t help, please make sure this install.php file is in the same place as the Themes folder.</p>
-		<p>If you continue to get this error message, feel free to <a href="https://support.simplemachines.org/">look to us for support</a>.</p>
-	</body>
-</html>';
+		<html>
```
Gwenwyfar: This reverts the changes in the recent upgrader PR.

Gwenwyfar (changes requested):

other/upgrade.php
frandominguezl: It doesn't makes sense to call some text strings if we couldn't find any language files.

illori:
are we not translating the change actually made to the database? for example 

> Transform default from gif to png done.

it seems that this will always be in english correct?

frandominguezl:
Where is it?

illori:
In the sql file.

frandominguezl:
No there's no need to translate that.

other/upgrade.php
Gwenwyfar: Yes, but you removed the css changes when you reverted the strings :P

Gwenwyfar (approved):

other/upgrade.php
```diff
@@ -4417,7 +4413,7 @@ function retTimeout(attemptAgain)
 							if (!attemptAgain)
 							{
 								document.getElementById("error_block").style.display = "";
-								setInnerHTML(document.getElementById("error_message"), "Server has not responded for ', ($timeLimitThreshold * 10), ' seconds. It may be worth waiting a little longer or otherwise please click <a href=\"#\" onclick=\"retTimeout(true); return false;\">here<" + "/a> to try this step again");
+								setInnerHTML(document.getElementById("error_message"), "', sprintf($txt['upgrade_repondtime'], ($timeLimitThreshold * 10)), '"<a href="#" onclick="retTimeout(true); return false;">"', $txt['upgrade_respondtime_clickhere'], '"</a>';');
```
jdarwood007:
```
setInnerHTML(document.getElementById("error_message"), "', sprintf($txt['upgrade_repondtime'], ($timeLimitThreshold * 10)), '" + "<a href=\"#\" onclick=\"retTimeout(true); return false;\">', $txt['upgrade_respondtime_clickhere'], '</a>");
```
That should fix up this line.

frandominguezl: Is it needed to scape the double quotes?

jdarwood007: It corrects it so it works correctly, otherwise you get a javascript errors.

frandominguezl: Should be done now

jdarwood007 (approved):

PR #4874, opened 2018-07-26
example how to handle bb code issue

albertlast:
It's an >example<
It cycle like the other upgrade script throw the database,
when the time get short the page recalled.

The important is that we use the index/primarykey to jump from one batch to the next,
to evade paging issues.

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -3134,7 +3134,7 @@ function ConvertBBCode()
 		$res = pg_fetch_assoc($result);
 		$db_return = pg_fetch_all($result);
 		$substep = count($db_return);
-		$substep = $db_return[$substep]['id'];
+		$substep = $db_return[$substep]['id_msg'];
 	}
 	
 
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -3085,6 +3085,79 @@ function ConvertUtf8()
 	return false;
 }
 
+function ConvertBBCode()
+{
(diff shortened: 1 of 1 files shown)


LexArma:
Why do you feel we need to edit the message bodies? It is a heavy operation however we do it, alters data that does not belong to us, and it is not quaranteed to be error free: the db makes no difference between [code]code[/code] and [code]examples of bbc use such as [green] color [/green] [/code]. It is better we do not force updates to existing messages. That should be left to the admin of the forum to decide what to do.

albertlast:
You don't provide the admin any tools for this,
so he will do nothing,

LexArma:
We do provide tools. We hide the nonfunctioning bbc, add hooks to help work them, and we can assist people on the forums if they want do the replace  you are suggesting - but the decision should be theirs, not ours.

albertlast:
Like i mention in the issue this code could be also use in a cronjob,
which convert silently over time the code,
so i see no issue with base idea,
where logic is placed is my eyes not important.

wintstar:
You can insert albertlast idea into the update script as a selection option. There the Admin can set a check mark if he wants that. Or display a message after the update that you can do this as a cron job in the admin center.

jdarwood007:
@albertlast 
I mentioned the silent background upgrader could also be done as another option.  The first option of just doing nothing with the bbc and letting it so mod authors could restore the functionality was better liked.
A silent background upgrader can be useful for more complex and intensive queries/processes, but involves us supporting it in the long term and ensuring the code continues to support it enough to not break during the silent upgrade process.

PR #4875, opened 2018-07-27
Assign the array to saveTriggers function fixes #4867

### LINKED ISSUE #4867: Cannot properly edit ban trigger
Applies on 22bf54f, probably earlier too.

1. Create a ban with a trigger (doesn't matter what)
2. From the ban list, go to modify the ban you just made
3. Select modify on any of the existing ban triggers
4. Edit the value
5. Save
6. Note that the value has not updated

Type of error: General 
2: Invalid argument supplied for foreach()
 http://localhost/SMF2.1/index.php?action=admin;area=ban;sa=edittrigger
 D:/xampp_php7/htdocs/SMF2.1/Sources/ManageBans.php (Line 930)

Backtrace:
#0: smf_error_handler()
#1: saveTriggers()
#2: BanEditTrigger()
#3: call_helper()
#4: Ban()
#5: AdminMain()

albertlast:
In my test with this changes,
the logic run fine.

@Arantor please test this pr

Issue #4867

### DIFF
--- a/Sources/ManageBans.php
+++ b/Sources/ManageBans.php
@@ -1616,9 +1616,7 @@ function BanEditTrigger()
 	elseif (isset($_POST['edit_trigger']) && !empty($_POST['ban_suggestions']))
 	{
 		// The first replaces the old one, the others are added new (simplification, otherwise it would require another query and some work...)
-		saveTriggers(array_shift($_POST['ban_suggestions']), $ban_group, 0, $ban_id);
-		if (!empty($_POST['ban_suggestions']))
-			saveTriggers($_POST['ban_suggestions'], $ban_group);
+		saveTriggers($_POST['ban_suggestions'], $ban_group, 0, $ban_id);
 
 		redirectexit('action=admin;area=ban;sa=edit' . (!empty($ban_group) ? ';bg=' . $ban_group : ''));
 	}


jdarwood007:
I'm not sure if the suggested fix is correct.  This is related to #4877 as well now.

I'm in the opinion that we either need to fix the interface as mentioned in #4877 or if we leave it, it should accept any additional triggers selected as well as updating the existing one.  @Sesquipedalian, @colinschoen or @MissAllSunday have an opinion on this?

Sesquipedalian:
My honest opinion is that the current way of editing ban triggers is needlessly overcomplicated. The "Add new ban" and "Edit ban" forms should be identical except for the title text and the fact that the values of "Edit ban" start with the current values instead of being empty. Then when the admin edits the ban as a whole, she could edit the ban triggers right there on the same form instead of being forced to go to these unnecessary extra forms.

However, the faster and easier fix would be to simplify the logic in the problematic part of `BanEditTrigger()` so that if `$_REQUEST['bi']` is not empty, we only modify the specified ban trigger and don't try to do anything clever to create new ones. Then the "Edit ban trigger" template should be changed so that it only shows a field for the existing trigger type of the trigger being edited. So for example, if the admin goes to `
?action=admin;area=ban;sa=edittrigger;bg=1;bi=2` and ban item 2 is a hostname ban, then the form should only show an input field for a hostname ban.

Sesquipedalian:
The changes in `BanEditTrigger()` would be to replace this:

```
		// The first replaces the old one, the others are added new (simplification, otherwise it would require another query and some work...)
		saveTriggers(array_shift($_POST['ban_suggestions']), $ban_group, 0, $ban_id);
		if (!empty($_POST['ban_suggestions']))
			saveTriggers($_POST['ban_suggestions'], $ban_group);

```

with this:

```
		saveTriggers(array_slice($_POST['ban_suggestions'], 0, 1), $ban_group, 0, $ban_id);
```

Then in `template_ban_edit_trigger()` we'd just need to add some if ... elseif logic to decide which input field to show.

albertlast:
When you combine this pr with https://github.com/SimpleMachines/SMF2.1/pull/4877 than you got in my eyes this behavior which you mention.
So i see no need to change here any logic.

Sesquipedalian:
No, this PR's changes to `BanEditTrigger()` would still allow new ban triggers to be created via `?action=admin;area=ban;sa=edittrigger;bg=1;bi=2`. The changes in #4877 don't really do anything to prevent multiple trigger criteria from being accepted and processed. Fiddling with the form's JavaScript only makes the user interface easier to work with; it doesn't change the possibilities for what can and cannot be done with the form's data.

albertlast:
That you didn't create new one do this pr,
the js pr add the gui changes to this.

PR #4877, opened 2018-07-28
Better Ban Edit flow

albertlast:
While debugging the Edit Ban issue,
i was thinik i could add multiple rules to one ban entity (stupid me).

After fixing the bug I think the checkboxes here: 
grafik
Are technical radio buttons,
but i don't want to introduce here radio button,
sooo i build js stuff to emulate radio buttons(when you marke one the others get unchecked).

And i changed the behavior of the Triggers checkbox,
in edit mode it's only a reset button,
because to check them all make no sense here.

The behavior in add mode is/should be untouched.

Since i used here jquery i convert the other stuff also in jquery,
to keep looking all the same.


illori:
there are cases where you do want to have more then 1 checked, so this does not work. i dont see a need to have this merged, the current function is not broken.

albertlast:
Like i mention in the intro of this pr,
you can't marke more than one (the code php only read the first one),
only in the case of "add" you can marke more than one.

And in this case you add multiple ban entities.

jdarwood007:
What @albertlast is saying here is that when editing an existing trigger, it doesn't make sense to be able to add additional triggers.

However it does work when editing a trigger to be able to check additional items and they will get added.

If we do change this to how @albertlast is suggesting, I would suggest hiding other options and removing the checkboxes completely then as it will end up causing confusion from how the new ban trigger form works.

albertlast:
With changes of today the bavior did change,
in the base you can't change the type any more,
this happen by not allow the change the check box:
grafik

@jdarwood007  @illori  is that what you like to see?

sbulen:
I agree with @illori 's note above.  The existing function works fine.  This should likely be closed.

PR #4880, opened 2018-07-28
Bbc removal

### LINKED ISSUE #4872: Old BBC Removal
It was brought to light that the upgrader process for old BBC removal is a killer during the upgrade process.  A quick team discussion and a suggestion has arose that seems to be a good course of action.

The problem comes down to using a LIKE against a text field in a table containing lots of data.  In all databases, this is a slow process.  Even with fulltext indexes this can still be slow.  Not all forums run full text.  In addition some upgrade scenarios can cause the fulltext to be removed to facilitate other upgrade needs.  Thus it is unsafe to trust a fulltext index to save the day.

The suggestion is as follows:
* Revert change to run the upgrader changes to remove the "old" bbc
* Add a section in the bbc parser for old bbc codes handling
* Add a hook for old bbc codes, allow accessing the entire list of "old" bbc
* Any bbc in the old bbc list will basically parse out as straight text.

The idea here is to silently just parse out as text any bbc in the old list.  This prevents the need for the upgrader to remove them.

Mod authors will need to use 2 hooks if they wish to revive a removed bbc.  One hook to add it to the bbc list, the second hook to modify the "old" bbc list.  Mod authors would read the array of "old" bbc and remove the one they want to support.

If anyone wants to work on this, they are welcome to at any time.  I'm marking it as  RC1 issue as this is a show stopper and will kill large forums during the upgrade process.

MissAllSunday:
fixes #4872

I added the list of legacy BBC tags on Load.php to allow mods/third party code to have access to the list on other places they might need it.

No new hooks are required, integrate_pre_load is inmediatly called after the list definition. Also removed the youtube bbc tag since the editor no longer supports it.

Still don't know how to actually enfore the list of legacy BBC tags, might as well just let mod authors to add the tags they want to re-add.


Gwenwyfar:
Isn't this missing shadow, move and glow?

jdarwood007:
Is it still displaying the legacy unhandled bbc or hiding them?

MissAllSunday:
I took the list from the upgrade process, I don't know which tags were removed or if theres a list somehwere.

Currently it doesn't do anything with them, they are displayed as plain regular text.

MissAllSunday:
Added the rest.

I still don't really know how to use that list, compare each BBC tag againts that list each time code is parsed seems like a waste of cycles TBH.

There was a huge fuzz about keeping old and outdated BBC tags intact but IMO we shouldn't encourage their usage, instead we should encourage either their removal or their upgrade to something modern, hence no list and provide an official mod/script to either remove them or upgrade them.

Sources/Load.php
```diff
@@ -369,6 +369,12 @@ function reloadSettings()
 		'red',
 		'green',
 		'blue',
+		'acronym',
+		'font',
```
Gwenwyfar: Font tag still exists.

jdarwood007:
I wouldn't do any comparing.  I would say after the $codes array is defined we simply do:
```
		// Handle legacy bbc codes.
		foreach ($context['legacy_bbc'] as $bbc)
			$codes[] = array(
				'tag' => $bbc,
				'before' => '',
				'after' => '',
			);
```

Simple code that is easily expanded and managed.  So if a tag is in legacy bbc, we just output text, if a mod author implants support, they simply remove from the legacy.  For example:

```
unset($context['legacy_bbc'][array_search('ftp', $context['legacy_bbc'])]);
```

MissAllSunday:
That would work yeah.

jdarwood007 (approved):

PR #4884, opened 2018-08-01
Error in RepairBoards.php

DiegoAndresCortes:
So apparently the description for the category is not being added and it causes the task to crash before it can fix and repair the problems caused by missing subjects, messages, topics, boards, etc.

If there's a better fix I don't mind this being dismissed.

Regards

### DIFF
--- a/Sources/RepairBoards.php
+++ b/Sources/RepairBoards.php
@@ -1768,8 +1768,8 @@ function createSalvageArea()
 	{
 		$salvageCatID = $smcFunc['db_insert']('',
 			'{db_prefix}categories',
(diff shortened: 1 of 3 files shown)


DiegoAndresCortes:
I also added a small fix for the GenericList.template.php

frandominguezl:
Tests throw an error because of the end of file extra spaces. Can you remove them?

DiegoAndresCortes:
So I accidentally removed the fork, do I send a new pr?

colinschoen:
Yeah we will close this PR once you submit the new one.

Issue #4882, opened 2018-07-30
Duplicate menus


Gwenwyfar:
We're duplicating the menus html (one for desktop and one for mobile), there should only be one, used for both.

DiegoAndresCortes:
I just did it, where/how can I post it for y'all to review it?

Regards

PR #4885, opened 2018-08-01
Possible fix for the double menu

DiegoAndresCortes:
I came up with this idea to try to solve #4882 

Let me know what you think


albertlast:
From technical side you got here the same issue like the other pr,
the end of line got changed.

Themes/default/index.template.php
```diff
@@ -339,23 +339,21 @@ function template_body_above()
...
+	// Show the menu here, according to the menu sub template, followed by the navigation tree.
 	// Load mobile menu here
 	echo '
 				<a class="menu_icon mobile_user_menu"></a>
-				<div id="mobile_user_menu" class="popup_container">
-					<div class="popup_window description">
-						<div class="popup_heading">', $txt['mobile_user_menu'], '
-						<a href="javascript:void(0);" class="generic_icons hide_popup"></a></div>
-						', template_menu(), '
+				<div id="mobile_user_menu" class="popup_container menu_popup_container">
```
Gwenwyfar: Can you use only one main class in the container here? Cleaner html this way (without the menu_popup_heading, etc)

Gwenwyfar (changes requested):
There's also a leftover padding from .popup_window in desktop view, needs to be removed.

Gwenwyfar:
Thanks for doing this. Only fixes the main menu though, could you do the submenus also?

DiegoAndresCortes:
I fixed it I think it's fine now.
I didn't know the genericmenu uses the double menu as well, I'll have a look at it later today

jdarwood007:
Can you fix the line endings please?

DiegoAndresCortes:
Sure, once I finish the genericmenu.

DiegoAndresCortes:
Fixed and updated for genericmenu

DiegoAndresCortes:
By the way, I forgot to mention that I decided to put the html code that wraps the function outside just in case a themer wants to easily replace/remove it, which is very likely.

Gwenwyfar (approved):

PR #4892, opened 2018-08-04
Add Privacy System to smf

albertlast:
Since it look like that some basic function should be included: https://www.simplemachines.org/community/index.php?topic=560206.0
I start with the topic and look how far i come.

- [x] Admin UI-Setting
- [x] User Download Section
- [x] Admin Download Secttion
- [x] Policy Management:
- [x] Create/Update
- [x] Delete
- [x] Evaluate which user is aggreed to which version
- [x] Enforce old aggrement to update to new one
- [x] Display the approved and new policy to the user
- [x] Integration into registration

Defined scope of this pr to deliever the core function of the core function,
not in scope is to provide finall looking or texting version.
For this is the topic to big in my eyes to do this all by my self.

Needs the fetch_all changes from pr: https://github.com/SimpleMachines/SMF2.1/pull/5048

### DIFF
--- a/Sources/ManageSettings.php
+++ b/Sources/ManageSettings.php
@@ -65,6 +65,7 @@ function ModifyFeatureSettings()
 		'likes' => 'ModifyLikesSettings',
 		'mentions' => 'ModifyMentionsSettings',
 		'alerts' => 'ModifyAlertsSettings',
+		'privacy' => 'ModifyPrivacySettings',
 	);
 
 	loadGeneralSettingParameters($subActions, 'basic');
@@ -94,6 +95,9 @@ function ModifyFeatureSettings()
 			),
 			'alerts' => array(
 				'description' => $txt['notifications_desc'],
+			),
+			'privacy' => array(
+				'label' => $txt['privacy'],
 			),
 		),
 	);
@@ -2273,6 +2277,46 @@ function ModifyAlertsSettings()
 	// Override the description
 	$context['description'] = $txt['notifications_desc'];
 	$context['sub_template'] = 'alert_configuration';
+}
+
+/**
+ * Config array for changing privacy settings
+ * Accessed  from ?action=admin;area=featuresettings;sa=privacy;
+ *
+ * @param bool $return_config Whether or not to return the config_vars array
+ * @return void|array Returns nothing or returns the $config_vars array if $return_config is true
+ */
+function ModifyPrivacySettings($return_config = false)
+{
+	global $txt, $scripturl, $context;
+
+	$config_vars = array(
+		array('check', 'enable_privacy_userexport'),
+		array('permissions', 'privacy_userexport_own'),
+		array('permissions', 'privacy_userexport_others'),
+	);
+
+	call_integration_hook('integrate_privacy_settings', array(&$config_vars));
+
+	if ($return_config)
+		return $config_vars;
+
+	// Saving?
+	if (isset($_GET['save']))
+	{
+		checkSession();
+
+		call_integration_hook('integrate_save_privacy_settings');
+
+		saveDBSettings($config_vars);
+		$_SESSION['adm-save'] = true;
+		redirectexit('action=admin;area=featuresettings;sa=privacy');
+	}
+
+	$context['post_url'] = $scripturl . '?action=admin;area=featuresettings;save;sa=privacy';
+	$context['settings_title'] = $txt['privacy'];
+
+	prepareDBSettingContext($config_vars);
 }
 
 ?>
--- a/Sources/ManageSettings.php
+++ b/Sources/ManageSettings.php
@@ -2330,7 +2330,7 @@ function ModifyPrivacySettings($return_config = false)
  */
 function ModifyPolicySettings($return_config = false)
 {
-	global $txt, $scripturl, $context, $sourcedir, $modSettings;
+	global $txt, $scripturl, $context, $sourcedir, $modSettings, $smcFunc;
 	
 	// Needed for the WYSIWYG editor.
 	require_once($sourcedir . '/Subs-Editor.php');
@@ -2347,7 +2347,7 @@ function ModifyPolicySettings($return_config = false)
 	// Now create the editor.
 	$editorOptions = array(
 		'id' => 'policy_text',
-		'value' => !empty($modSettings['policy_text'.$currentVersion]) ? $modSettings['policy_text'.$currentVersion] : '',
+		'value' => !empty($modSettings['policy_text' . $currentVersion]) ? $modSettings['policy_text' . $currentVersion] : '',
 		'height' => '250px',
 		'width' => '100%',
 		'labels' => array(
@@ -2362,9 +2362,20 @@ function ModifyPolicySettings($return_config = false)
 	
 	call_integration_hook('integrate_policy_settings', array(&$config_vars));
 	
-	$context['policy']['old'] = 100;
-	$context['policy']['not'] = 50;
-	$context['policy']['fresh'] = 10000;
+	$request = $smcFunc['db_query']('', '
+		SELECT count( case when th.value is null then 1 end) "not",
+			count( case when th.value is not null and th.value != {string:policy_version} then 1 end) "old",
+			count( case when th.value = {string:policy_version} then 1 end) "fresh"
+		FROM smf_members mem
+		LEFT JOIN smf_themes th ON (mem.id_member = th.id_member AND th.id_theme = 1 AND th.variable = {string:policy_approved})',
+		array(
+			'policy_version' => 'policy_text' . $currentVersion,
+			'policy_approved' => 'policy_approved',
+		)
+	);
+	
+	list ($context['policy']['not'], $context['policy']['old'], $context['policy']['fresh']) = $smcFunc['db_fetch_row']($request);
+	$smcFunc['db_free_result']($request);
 
 	if ($return_config)
 		return $config_vars;
--- a/Sources/ManageSettings.php
+++ b/Sources/ManageSettings.php
@@ -2387,24 +2387,42 @@ function ModifyPolicySettings($return_config = false)
 
 		call_integration_hook('integrate_save_policy_settings');
 
-		if (!empty($_REQUEST['save_new_policy']))
+		if (!empty($_REQUEST['enforce_new']))
 		{
-			$currentVersion++;
+			$smcFunc['db_query']('', '
+				UPDATE {db_prefix}themes c
+				SET "value" = {string:value}
+				WHERE EXISTS (
+					SELECT a.id_member
+					FROM {db_prefix}themes a
+					LEFT JOIN {db_prefix}themes b ON (a.id_member = b.id_member and b.variable = {string:bvar})
+					WHERE a.variable = {string:avar} and a."value" = {string:aval} 
+						AND ( b.value != {string:bval}  or b.value is null)
+						AND a.id_member = c.id_member
+				)
+				AND c."variable" = {string:avar}',
+				array(
+					'value' => '0',
+					'bvar' => 'policy_approved',
+					'avar' => 'policy_isvalid',
+					'aval' => '1',
+					'bval' => 'policy_text' . $currentVersion,
+				)
+			);
+		}
+		elseif (!empty($_REQUEST['save_new_policy']) || !empty($_REQUEST['update_policy']))
+		{
+			if (!empty($_REQUEST['save_new_policy']))
+				$currentVersion++;
+			
 			$config_vars[] = array('text', 'policy_text'.$currentVersion);
 			$_POST['policy_text'.$currentVersion] = $_REQUEST['policy_text'];
 			$config_vars[] = array('text', 'policy_version');
 			$_POST['policy_version'] = 'policy_text'.$currentVersion;
 			unset($config_vars[1]);
+			saveDBSettings($config_vars);
 		}
-		elseif (!empty($_REQUEST['update_policy']))
-		{
-			$config_vars[] = array('text', 'policy_text'.$currentVersion);
-			$_POST['policy_text'.$currentVersion] = $_REQUEST['policy_text'];
-			$config_vars[] = array('text', 'policy_version');
-			$_POST['policy_version'] = 'policy_text'.$currentVersion;
-			unset($config_vars[1]);
-		}
-		saveDBSettings($config_vars);
+
 		$_SESSION['adm-save'] = true;
 		redirectexit('action=admin;area=featuresettings;sa=policy');
 	}
--- a/Sources/ManageSettings.php
+++ b/Sources/ManageSettings.php
@@ -2425,7 +2425,7 @@ function ModifyPolicySettings($return_config = false)
 			)
 			AND variable = {string:avar}';
 			
-			$smcFunc['db_query']('managePrivacy',
+			$smcFunc['db_query']('managePrivacyNew',
 				$select,
 				array(
 					'value' => '0',
@@ -2464,33 +2464,20 @@ function ModifyPolicySettings($return_config = false)
 		call_integration_hook('integrate_manage_policy_settings');
 		
 		// set user with the policy invalid
-			if ($smcFunc['db_title'] == 'MySQLi')
-				$select = '
-					UPDATE {db_prefix}themes c
-					JOIN (
-						SELECT a.id_member
-						FROM {db_prefix}themes a
-						LEFT JOIN {db_prefix}themes b ON (a.id_member = b.id_member and b.variable = {string:bvar})
-						WHERE a.variable = {string:avar} and a.value = {string:aval} 
-							AND b.value = {string:bval}
-					) d ON (d.id_member = c.id_member)
-					SET value = {string:value}
-					WHERE c.variable = {string:avar}';
-			else
-				$select = '
-					UPDATE {db_prefix}themes
-					SET value = {string:value}
-					WHERE EXISTS (
-						SELECT a.id_member
-						FROM {db_prefix}themes a
-						LEFT JOIN {db_prefix}themes b ON (a.id_member = b.id_member and b.variable = {string:bvar})
-						WHERE a.variable = {string:avar} and a.value = {string:aval} 
-							AND b.value = {string:bval}
-							AND a.id_member = {db_prefix}themes.id_member
-					)
-					AND variable = {string:avar}';
+		$select = '
+			UPDATE {db_prefix}themes
+			SET value = {string:value}
+			WHERE EXISTS (
+				SELECT a.id_member
+				FROM {db_prefix}themes a
+				LEFT JOIN {db_prefix}themes b ON (a.id_member = b.id_member and b.variable = {string:bvar})
+				WHERE a.variable = {string:avar} and a.value = {string:aval} 
+					AND b.value = {string:bval}
+					AND a.id_member = {db_prefix}themes.id_member
+			)
+			AND variable = {string:avar}';
 
-		$smcFunc['db_query']('',
+		$smcFunc['db_query']('managePrivacyInvalid',
 				$select,
 			array(
 				'value' => '0',
@@ -2502,32 +2489,19 @@ function ModifyPolicySettings($return_config = false)
 		);
 		
 		// empty users
-			if ($smcFunc['db_title'] == 'MySQLi')
-				$select = '
-					UPDATE {db_prefix}themes c
-					JOIN (
-						SELECT a.id_member
-						FROM {db_prefix}themes a
-						LEFT JOIN {db_prefix}themes b ON (a.id_member = b.id_member and b.variable = {string:bvar})
-						WHERE a.variable = {string:avar} and a.value = {int:aval} 
-							AND b.value = {string:bval}
-					) d ON (d.id_member = c.id_member)
-					SET c.value = {string:value}
-					WHERE c.variable = {string:avar}';
-			else
-				$select = '
-					UPDATE {db_prefix}themes
-					SET value = {string:value}
-					WHERE EXISTS (
-						SELECT a.id_member
-						FROM {db_prefix}themes a
-						LEFT JOIN {db_prefix}themes b ON (a.id_member = b.id_member and b.variable = {string:bvar})
-						WHERE a.variable = {string:avar} and a.value = {string:aval} 
-							AND b.value = {string:bval}
-							AND a.id_member = {db_prefix}themes.id_member
-					)
-					and variable = {string:bvar}';
-		$smcFunc['db_query']('',
+		$select = '
+			UPDATE {db_prefix}themes
+			SET value = {string:value}
+			WHERE EXISTS (
+				SELECT a.id_member
+				FROM {db_prefix}themes a
+				LEFT JOIN {db_prefix}themes b ON (a.id_member = b.id_member and b.variable = {string:bvar})
+				WHERE a.variable = {string:avar} and a.value = {string:aval} 
+					AND b.value = {string:bval}
+					AND a.id_member = {db_prefix}themes.id_member
+			)
+			and variable = {string:bvar}';
+		$smcFunc['db_query']('managePrivacyEmptyusers',
 			$select,
 			array(
 				'value' => '',
--- a/Sources/ManageSettings.php
+++ b/Sources/ManageSettings.php
@@ -2419,7 +2419,7 @@ function ModifyPolicySettings($return_config = false)
 					SELECT a.id_member
 					FROM {db_prefix}themes a
 					LEFT JOIN {db_prefix}themes b ON (a.id_member = b.id_member and b.variable = {string:bvar})
-					WHERE a.variable = {string:avar} and a."value" = {string:aval} 
+					WHERE a.variable = {string:avar} and a.value = {string:aval} 
 						AND ( b.value != {string:bval}  or b.value is null)
 						AND a.id_member = c.id_member
 				)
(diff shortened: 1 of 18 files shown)


albertlast:
While working on this,
came the question up,
which field should be included for profile data,
messages and personal messages.

To give you an idea of the status,
personal data:
[privacySMFprofile-9.zip](https://github.com/SimpleMachines/SMF2.1/files/2260364/privacySMFprofile-9.zip)
messages:
[privacySMFmessages-3.zip](https://github.com/SimpleMachines/SMF2.1/files/2260365/privacySMFmessages-3.zip)
personal messages:
[privacySMFpmessages-1.zip](https://github.com/SimpleMachines/SMF2.1/files/2260367/privacySMFpmessages-1.zip)

frandominguezl:
https://www.simplemachines.org/community/index.php?topic=560645.0

> - Data export, so users may (if you allow them to) export their profile data. (Profile including IP address(-history), posts (optional), personal messages (optional)). In the future, SMF might get an option to restore the basic profile of another site.

albertlast:
The url is not visable by contributer.

wintstar:
> The url is not visable by contributer. https://www.simplemachines.org/community/index.php?topic=560206.0

Yes, i can not read this topic when i logged in.

> The topic or board you are looking for appears to be either missing or off limits to you.

illori:
that is a topic in a semi private topic @frandominguez03 please only link to public topics on the forum.

frandominguezl:
@illori Isn't the topic on SMF Team Blog board?

illori:
no this one is https://www.simplemachines.org/community/index.php?topic=560645.0

frandominguezl:
That's the one I linked in my message.

illori:
oh sorry it was @albertlast that linked to the private topic.

frandominguezl:
Anyway he's saying that he doesn't have access to the topic I linked, which means he doesn't have access to that board. Can you have a look at it when you have a spot?

illori:
should be fixed now.

wintstar:
> should be fixed now.

No, i can not read topic=**560206.0**, linked from alberlast.

frandominguezl:
No, you're not supposed to read that. There was an error in which the one I linked couldn't be viewed by albertlast because he belongs to a special group.

illori:
that is in a private board, it should not have been linked here in the first place.

wintstar:
> that is in a private board, it should not have been linked here in the first place.

Then a participation makes no sense.

Themes/default/languages/ManageSettings.english.php
```diff
@@ -101,6 +101,22 @@
 // Like settings.
 $txt['enable_likes'] = 'Enable likes';
 
+// Privacy settings.
+$txt['enable_privacy_userexport'] = 'Enable Userdata Export by the User';
+$txt['privacy_userexport_own'] = 'Allow to Download ther own Data';
```
frandominguezl: ther -> their

albertlast: Your comment remind me,
to define the scope of this pr.

I will change this one thing,
but please look over the complet text after the pr is merge,
so you can check if the text make sense in the context where it's used.

frandominguezl: Sure, don't worry. I'll take a look at all the text strings once you're finished.

albertlast:
Since ther are no database changes included should be easy for everyone to test this pr,
so please test this.

frandominguezl:
I get the following error when accessing the "Policy" tab under Configuration > Features and Options

screenshot_1

Backtrace:

screenshot_2

EDIT1: Got these 2 new errors after; Profile > Actions > Download Profile > Download the posts of your account:

screenshot_3

Backtraces:
1st one
screenshot_4

2nd one
screenshot_5

Same errors appeared after clicking "Download the personal messages data of your account" on the same route, with different lines:
screenshot_6

albertlast:
Did some base testing with my mysql env,
should be working better now.

wintstar:
Great job @alberlast   :+1:

Missing language String for "current"
Admin.template.php line 1717
```
		echo '
									<dt>', ($row['new'] ? 'current' : '<input type="submit" name="delete_policy" value="' . $row['name'] . '" class="button">') , '</dt><dd>', $row['amount'], '</dd>';
```
privacy-string-current

albertlast:
Added current txt language,
reordner mysql and pg query
minor adjustments of the "pg" sql in a standard way which run on pg,sqlite,oracle and ms sql

wintstar:
@albertlast could you please take a look at this=>   
https://www.simplemachines.org/community/index.php?topic=560645.msg3984212#msg3984212 ?  
Thank you :smiley:

albertlast:
I don't understand why you mention issue there and not here

wintstar:
I didn't want to have a discussion here. Assume that others see it differently and then here is a bigger discussion.

albertlast:
Your mention issue should be fixed, now.

wintstar:
You're great.  👍 

Language strings are missing when creating a report for group permissions. Only a missing one is shown in the screenshot, but I think it is more.

privacy-group-permission

wintstar:
Fix works. I have not found any more bugs.

frandominguezl (approved):

wintstar:
Now that the policy exists, is still agreement.txt is needed?

NanoSector:
One thing that bothers me about this PR is that a lot of things have a capital first letter where they do not need them. It's minor, though.

image

What I'd suggest;

* Enable exporting of user data
* Allow users to download their own data
* Allow users to download other persons data

Also, in the Policy settings, the Policy overview isn't exactly clear at first glance to me since the boxes seem editable but are not. The Policy Management header below that seems...broken?
image

Checking "Enable policy" and clicking save doesn't work.

Perhaps it would be nice to see if we can ship a default policy that covers the most common boards and SMF's base features. That way we can have this enabled by default.

@wintstar I don't think it's a good idea to scrap agreement.txt (but why is it a text file...), since agreement.txt contains behavioral policies versus data policies.

frandominguezl:
We'll handle the GUI and language improvements since the PR is merged. Albertlast stated that he'd be only commiting the core idea of this.

NanoSector:
That still leaves the matter of the non-functional enable policy checkbox.

albertlast:
Whick save button you use?

NanoSector:
The one next to the enable policy checkbox, above the box where you input the policy.

albertlast:
in the moment is when you use the button for update/save policy you update the status of the policy modus too.
In this way it's work already before the last commit,
it's oaky?

NanoSector:
I also tried that but that didn't work either, I'll re-test later.

wintstar:
Now privacy is activated or deactivated in the profile. Its works. 

Wouldn't it be better to check if there are posts and private messages to display the download buttons or not? If there are none and you click on the buttons nothing happens.

albertlast:
When you click and nothing is ther,
you got a infobox that no data was found.

wintstar:
> When you click and nothing is ther,
> you got a infobox that no data was found.

This does not work in my test board. No info box is displayed

albertlast:
Did you use maybe a custom theme?
grafik

wintstar:
No, Curve2. I wil check the language string. No diplay in english and german.

wintstar:
Yes, it works now. Browser cache was the cause.

albertlast:
@Sesquipedalian any status about this here?

Sesquipedalian:
There's just a lot to test in this PR and my free time this autumn has been more limited than I had expected, that's all. I've added this to the RC2 milestone so that we can be sure to include it in the final release without having it hold up the release of RC1 in the meantime.

Sesquipedalian (changes requested):
I haven't yet been able to fully review this yet, so I may need to request more changes later, but here is one request:

SMF already has a system for exporting user profile information that we can and should use. Just go to `index.php?action=.xml;sa=profile;u=20;type=smf` and it will output the profile info for user 20 (if you have permission to view it, of course). So instead of these new functions exporting user profile data, all we need is to add a link in the profile template pointing to that URL. Since this method already exists and since XML data is easier to manipulate than CSV data if necessary later, it'd be better to use this approach.

So, please change this PR to direct the user to that address in your `template_getProfileData()` function, and then delete the functions to generate the CSV file.

Sesquipedalian:
As a followup to my review comment above, I see that your `getProfileData()` function can export profile info, posts, and private messages. The existing XML export method doesn't currently have a way to export posts or private messages, but I can create one easily enough. In fact, I am already going to be creating that soon for SMF 2.0, so it will be easy enough to port that over to 2.1.

wintstar:
I see only one problem. It is not multilingual.

profile-xml

albertlast:
Well i looked short in the function (News.php ShowXMLFeed) and it takes my pov the wrong route,
first issue that it use xml, xml is a dead format.

It's hard to read when you output many data,
it's got a huge overheader,
the importing of xml into excel(similiar tools) or in a database is very hard.
This format doesn't fit in here.

Csv is the right way of doing.

The second issue is the authority check is wrong for this use case (i need to be able to define if they allowed to download ther own data and or of some others).

it use the {query_see_board} auth check which is wrong to do this here.

The code look like run very slow not good to export hugh amount data,
easy to ddos this function.

So in my recommandtion drop the old stuff or don't use this for this use case here.

Sesquipedalian:
XML is not at all a dead format. It may have a higher overhead, but it is also better at communicating information reliably between systems. CSV is a fragile format where meaning is dependent on the order of the columns. If the output from one forum includes columns that another forum doesn't expect, or has in a different order, or anything like that, using CSV as an import/export format is going to be messy and possibly break things.

Still, I'd say that there are uses for both formats, and we can and should offer both export formats. Then the user can choose the format that best meets her needs.

Sesquipedalian:
Hm. I thought I posted the following earlier, but I don't see it here now.

I think it is fair to say that there are valid use cases for both XML and CSV. But don't worry about the XML export. I will put that part together to add after this gets merged.

wintstar:
One could also simply create a page where the user can print his data. Printing not only in paper form but also as PDF is possible with most operating systems.

Sources/Profile-Actions.php
```diff
@@ -969,4 +969,196 @@ function subscriptions($memID)
...
+		foreach($profile as $key => &$value)
+		{
+			if (is_array($value))
+				$value = $smcFunc['json_encode']($value);
+		}
+		$profileData[0] = array_keys($profile);
+		$profileData[1] = $profile;
+		call_integration_hook('integrate_getProfile_profile', array(&$profileData));
+	}
+	elseif ($mode == 'messages') // messages
```
Sesquipedalian, line 1013: A user could have thousands of posts, and if he does, processing all of them at once like this will overwhelm the server—either it will run out of memory, or it will run out of time, or both. So, this needs some sort of method to work through them in manageable batches. Writing to a temporary file and automatically reloading the page every few seconds, like I did with the XML version, would probably work well here.

Sources/Profile-Actions.php
```diff
@@ -969,4 +969,196 @@ function subscriptions($memID)
...
+		);
+		$profileData = $smcFunc['db_fetch_all']($request);
+		if (!is_array($profileData) || empty($profileData))
+			redirectexit('action=profile;area=getprofiledata;u=' . $memID . ';nofound=1');
+		array_unshift($profileData, array_keys($profileData[0]));
+		$smcFunc['db_free_result']($request);
+
+		call_integration_hook('integrate_getProfile_messages', array(&$profileData));
+	}
+	elseif ($mode == 'pmessages')
```
Sesquipedalian, line 1037: Exporting personal messages could also overwhelm the server, so this will need a method to process them in manageable batches, too.

Sources/Profile-Actions.php
```diff
@@ -969,4 +969,196 @@ function subscriptions($memID)
...
+		}
+		$profileData[0] = array_keys($profile);
+		$profileData[1] = $profile;
+		call_integration_hook('integrate_getProfile_profile', array(&$profileData));
+	}
+	elseif ($mode == 'messages') // messages
+	{
+		
+		$request = $smcFunc['db_query']('','
+			SELECT id_msg, id_topic, poster_time, subject, modified_time, modified_name, modified_reason, body, likes, poster_ip
```
Sesquipedalian, line 1017: Don't forget to use your nice little `int_dtop()` function to decode the IP address (here and elsewhere).

Sources/Profile-Actions.php
```diff
@@ -969,4 +969,196 @@ function subscriptions($memID)
...
+			redirectexit('action=profile;area=getprofiledata;u=' . $memID . ';nofound=1');
+		array_unshift($profileData, array_keys($profileData[0]));
+		$smcFunc['db_free_result']($request);
+
+		call_integration_hook('integrate_getProfile_messages', array(&$profileData));
+	}
+	elseif ($mode == 'pmessages')
+	{
+		$request = $smcFunc['db_query']('','
+			SELECT pm.msgtime, pm.subject, pm.body
```
Sesquipedalian, line 1040: This query grabs PMs from the user and PMs to the user, but the exported data doesn't indicate who was the author and who were the recipients. It would be more helpful if each row identified both.

Sources/Profile-Actions.php
```diff
@@ -969,4 +969,196 @@ function subscriptions($memID)
...
+		foreach ($removeFields as $value)
+		{
+			unset($profile[$value]);
+		}
+		foreach($profile as $key => &$value)
+		{
+			if (is_array($value))
+				$value = $smcFunc['json_encode']($value);
+		}
+		$profileData[0] = array_keys($profile);
```
Sesquipedalian, line 1009: I like that you are providing a header row here. Please add one to the post and personal messages CSVs as well.

Sources/Profile-Actions.php
```diff
@@ -969,4 +969,196 @@ function subscriptions($memID)
...
+	
+	if (empty($_REQUEST['activity']))
+		return;
+	
+	$mode = $_REQUEST['activity'];
+
+	$profileData = array();
+	if ($mode == 'profile') // profile
+	{
+		loadMemberData($memID, false, 'profile');
```
Sesquipedalian, line 997: It would be easier and better to use `loadMemberContext()` here. It formats the member data into a nicer form for reading, you wouldn't need to unset security-related fields, and the custom profile fields could easily be presented as distinct columns instead of a blob of JSON data.

Sources/ManageSettings.php
```diff
@@ -2275,4 +2281,293 @@ function ModifyAlertsSettings()
...
+	// Store the ID for old compatibility.
+	$context['post_box_name'] = $editorOptions['id'];
+	
+	call_integration_hook('integrate_policy_settings', array(&$config_vars));
+	
+	$request = $smcFunc['db_query']('', '
+		SELECT count( case when th.value is null then 1 end) novalid,
+			count( case when th.value is not null and th.value != {string:policy_version} then 1 end) outdated,
+			count( case when th.value = {string:policy_version} then 1 end) fresh
+		FROM smf_members mem
```
Sesquipedalian: `smf_` → `{db_prefix}`

Sources/ManageSettings.php
```diff
@@ -2275,4 +2281,293 @@ function ModifyAlertsSettings()
...
+		LEFT JOIN smf_themes th ON (mem.id_member = th.id_member AND th.id_theme = 1 AND th.variable = {string:policy_approved})',
```
Sesquipedalian: `smf_` → `{db_prefix}`

Sources/ManageSettings.php
```diff
@@ -2275,4 +2281,293 @@ function ModifyAlertsSettings()
...
+		array(
+			'policy_version' => 'policy_text' . $currentVersion,
+			'policy_approved' => 'policy_approved',
+		)
+	);
+	
+	list ($context['policy']['novalid'], $context['policy']['outdated'], $context['policy']['fresh']) = $smcFunc['db_fetch_row']($request);
```
Sesquipedalian: Minor language change: "invalid" would be better that "novalid"

Sources/ManageSettings.php
```diff
@@ -2275,4 +2281,293 @@ function ModifyAlertsSettings()
...
+}
+
+/**
+ * Config array for changing policy settings
+ * Accessed  from ?action=admin;area=featuresettings;sa=policy;
+ *
+ * @param bool $return_config Whether or not to return the config_vars array
+ * @return void|array Returns nothing or returns the $config_vars array if $return_config is true
+ */
+function ModifyPolicySettings($return_config = false)
```
Sesquipedalian, line 2331: This function needs some logic to ensure that `enable_policy_function` cannot be set to true unless `policy_version` and any other necessary values are defined. 

Right now, if the admin makes the mistake of enabling the policy function without setting up everything else first, the admin will be redirected on every page load to the page where he is prompted to accept the policy. But the admin will be unable to do so because there is no policy to accept, and so he will be stuck in an endless loop and will be unable to fix his forum except by editing his database directly.

Sources/ManageSettings.php
```diff
@@ -2275,4 +2281,293 @@ function ModifyAlertsSettings()
...
+		redirectexit('action=admin;area=featuresettings;sa=policy');
+	}
+	elseif (isset($_GET['manage']))
+	{
+		checkSession();
+		
+		call_integration_hook('integrate_manage_policy_settings');
+		
+		// set user with the policy unvalid
+			if ($smcFunc['db_title'] == 'MySQLi')
```
Sesquipedalian: I'd rather avoid having two versions of the queries in this section. If necessary, feel free to create another `$smcFunc['db_***']` function to handle this cleanly.

Sources/Register.php
```diff
@@ -50,6 +50,10 @@ function Register($reg_errors = array())
 	$context['require_agreement'] = !empty($modSettings['requireAgreement']);
 	$context['registration_passed_agreement'] = !empty($_SESSION['registration_agreed']);
 	$context['show_coppa'] = !empty($modSettings['coppaAge']);
+	
+	// Do we nee them to agree to the policy agreement, secondly?
+	$context['require_policy'] = !empty($modSettings['enable_policy_function']);
```
Sesquipedalian, line 55: Here would be another good place to add more logic to make sure that all the necessary variables are defined before requiring the policy.

Themes/default/Admin.template.php
```diff
@@ -1635,4 +1635,94 @@ function template_admin_quick_search()
 								</span>';
 }
 
+/**
+ * This shows the policy settings
+ */
+function template_edit_policy()
```
Sesquipedalian, line 1641: The layout of this template is a bit messy. But we can fix that later. :)

index.php
```diff
@@ -212,6 +212,15 @@ function smf_main()
 	// Make sure that our scheduled tasks have been running as intended
 	check_cron();
 
+	// Privacy logic only when enabled, got no valid version, try not to call the policy data
+	if (!empty($modSettings['enable_policy_function']) && !$user_info['is_guest'])
```
Sesquipedalian: It might be a good idea to add ` && !$user_info['is_admin']` to this. Then if anything goes wrong with the privacy policy, the admin can still get to the admin centre to fix it.

Sesquipedalian (changes requested):

Sources/ManageSettings.php
albertlast: Make less sense to place sutch special query in a generic env,
this query is only here needed.

Sources/Profile-Actions.php
albertlast: Like you see here: https://github.com/SimpleMachines/SMF2.1/pull/4892#issuecomment-410521610
they get already added.

Sources/Profile-Actions.php
albertlast: This is indended,
because otherwise you export userdata from a different user in your data and
never allowed you to download his data.

Sources/ManageSettings.php
albertlast: Since i check everywhere if enabled_policy_function is true:
https://github.com/SimpleMachines/SMF2.1/blob/66f502c31e0e630d42c3b3a60a56b9345736743b/index.php#L216
I see no need here todo any checks in the settings.

Sources/Profile-Actions.php
albertlast: 100k messages takes 500 ms -> looks good to me
the file stuff got many weakness 
it created by his self a new gdpr topic
it could be infected by evil software
it take more time to proceed -> the cpu load is overall higher

Sources/Profile-Actions.php
albertlast: duplicate to message,
we only need to talk at one place.

Themes/default/Admin.template.php
albertlast: Like i said,
i don't want to invest time in stuff in topic where i realy bad.

Sources/Profile-Actions.php
Sesquipedalian: The displayed names of other users don't count as private data, so it is perfectly fine to include them in the downloaded information.

Sesquipedalian: Just to clarify, I'm asking for `pm.from_name` and `GROUP_CONCAT(COALESCE(mem.real_name, mem.member_name)) AS to_names` to be added to the SELECT in this query (which will require joining the members table, of course), and for those values to be included in the output columns.

Themes/default/Admin.template.php
Sesquipedalian: No problem. When I said, "But we can fix that later," I meant that we can deal with it in another PR later sometime. 🙂

Sources/Profile-Actions.php
albertlast: In my eyes you don't need this,
would be enough to mark if you send or receive the message and maybe the technical user_id from the other.

Sources/Profile-Actions.php
Sesquipedalian: How much memory was used to process those 100,000 messages all at once?

Sources/Profile-Actions.php
Sesquipedalian: As a rule, human beings find it easier to understand who a message was sent from or sent to if the name of the person is given than if an ID number is given. Please use the names.

Sources/Profile-Actions.php
Sesquipedalian: Yes, you are right that there is a header row. Unfortunately, it is filled with numeric keys instead of column names, so it isn't useful at all. Please change that.

Sources/ManageSettings.php
Sesquipedalian: We do our best to keep all alternate SQL queries in the Subs-Db-* files at all times. That ensures that if changes need to be made in the future, we know exactly where to look to find all the code that needs to be changed. If MySQL and PostgreSQL can't run the same query to do the same task in this case, then an abstraction function is required in order to comply with SMF's coding practices. I won't be able to merge this until the change is made.

Sources/Profile-Actions.php
albertlast: Nope please read the start post,
you need to apply the pr #5048
otherwise the fetch_all is different across pg and mysql.

Sources/ManageSettings.php
albertlast: got fixed.

Sources/Profile-Actions.php
Sesquipedalian: Fair enough. 🙂

Sources/Profile-Actions.php
LiroyvH: @Sesquipedalian That's very important indeed, especially problematic on more aggressively resource-limited shared hosting. Plus hosting many boards on a server executing that kind of queries can become a nuisance. Always assume low resources is probably a good idea.

albertlast: It needs to be realistic,
100k message  is more than sm.org got
And you will not host setup like this on freehost

Sesquipedalian: That's not an answer to the question, @albertlast. 🤨

There are very good reasons why SMF uses paging to perform any operation involving large numbers of table rows. Those reasons apply here just as much as they do anywhere else.

albertlast: you didn't awnser the question for the xml solution.

I test some solution,
handle the size better.

Sesquipedalian:
Hey @albertlast. As you know, we are also working on adding code to SMF 2.0,16 to make it compliant with the GDPR. Some of the code in this PR can be used in 2.0.16, but other parts require a different approach in order to work within the confines of 2.0. So in order to maintain as much compatibility as possible between 2.0 and 2.1 and thereby make upgrading from 2.0 to 2.1 easier, my plan is to make my own local copy of this PR, integrate the changes we're making in 2.0 into it, and then push it all up as a new PR for 2.1. Once that is ready, I'll close this PR (and #5081, since it will also become obsolete) so that we can focus on testing the final version and get this task done. In the meantime, I am going to leave this PR open.

albertlast:
For better testing reason i would like to merge a newer version of smf into this pr,
but i'm not sure if the backwards transfer is than more complex.

Yes feel free to adapt this pr.

I did some poc with possibilty to generate the csv directly from the database side,
but the stuff feels not right and only works with postgres...

albertlast:
@Sesquipedalian for 2.1 we could also try to use generators: http://php.net/manual/en/language.generators.overview.php
this was added by php 5.5,
but i see no big problem to provide this function only with php 5.5+

PR #4902, opened 2018-08-09
Added db_port and db_mb4 supprt to the upgrade fixes #4901

### LINKED ISSUE #4901: upgrade.php ignoring $db_port
i upgraded my wamp install, i now use MariaDB that is running on port 3307. it works fine for the forum but when running upgrade.php it errors out with "Warning: mysqli_real_connect(): (HY000/2002): No connection could be made because the target machine actively refused it."

if i modify the $db_server to `$db_server = 'localhost:3307';` upgrade.php will run but the port is removed and upgrade.php will fail the next run as the port has been removed.

albertlast:
issue: #4901

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -578,7 +578,7 @@ function loadEssentialData()
 		if (empty($db_connection))
 		{
 			$options = array('non_fatal' => true);
-				// Add in the port if needed
+			// Add in the port if needed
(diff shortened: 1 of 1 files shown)


illori:
confirmed this does resolve the issue.

jdarwood007 (approved):

PR #4907, opened 2018-08-13
Fix visual brokenness of the Ignore Boards profile area

NanoSector:

Before:
image

After:
image

### DIFF
--- a/Themes/default/Profile.template.php
+++ b/Themes/default/Profile.template.php
@@ -2192,10 +2192,10 @@ function template_groupMembership()
  */
 function template_ignoreboards()
 {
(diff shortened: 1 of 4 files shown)


NanoSector:
Adding WIP tag because there are more places where this is used.

NanoSector:
image

This should do... However it does not work well when on mobile/small resolutions.

jdarwood007:
screen shot 2018-08-13 at 6 08 12 pm

NanoSector:
I've changed it to a grid, but I'm not sure if I'm happy with this...
image

jdarwood007 (approved):

PR #4908, opened 2018-08-14
Remove the Flash BBC tag and corresponding admin panel setting

NanoSector:


colinschoen (approved):
LGTM

PR #4911, opened 2018-08-14
Fix #4897: Implement jQuery smiley selector

### LINKED ISSUE #4897: Smileys followup
This is a bit of a placeholder...   

After recent smiley set issues (#4773 & #4767 ) it was decided to rewrite the smiley set code to make it more efficient.  There is a temporary solution out there (#4772 ) that does not address all presentations of smileys, e.g., user selection still doesn't properly display non-png smileys:
image

NanoSector:
This selector iterates over the most common file extensions to find one that's available. Worked in my testing with Firefox.

### DIFF
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -461,7 +461,7 @@ function loadProfileFields($force_reload = false)
 			'preload' => function() use ($modSettings, &$context, $txt, $cur_profile, $smcFunc)
 			{
 				$context['member']['smiley_set']['id'] = empty($cur_profile['smiley_set']) ? '' : $cur_profile['smiley_set'];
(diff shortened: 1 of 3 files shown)


NanoSector:
Did I miss any common extensions? BMP maybe?

jdarwood007 (approved):

jdarwood007:
I'm not going to worry about bmp.  I don't think they should be a supported extension.

sbulen:
I think there was at least one more location...    The editor?

NanoSector:
I forgot SVG in the extensions list.

@sbulen: I think that's a whole different problem and has to be dealt with in another way. Perhaps we can reuse some of this code.

sbulen:
Ok.  I'll log the editor issue.

PR #4914, opened 2018-08-14
Fix #4913

### LINKED ISSUE #4913: Undefined index: calendar_start_day
image

This goes away after setting a start day in the Look & Layout options.

NanoSector:
Cleans up the profile template settings page, and also sets a default value for calendar_start_day on said page. I'm not happy with having those keys settings hardcoded; might look at a better way.


jdarwood007 (approved):
Looks good, need CI job restarted and if its good this can merge

PR #4928, opened 2018-08-19
[WIP] Added classes for hidden and inline_block

XinYenFon:
Lets say this is a WIP for now.


Gwenwyfar:
Is that not tied to js?

SychO9:
the hidden class is used to hide elements when the screen is under 720px adding the class to index.css will mess a lot of things, hidden in this case should be renamed to something else that hides elements no matter the view

XinYenFon:
I don't have time to work on this sorry.

PR #4948, opened 2018-08-19
Unwatch bug

SychO9:
This test breaks the option to Unwatch, there is already a similar test in line 132.

If you go to a topic, choose not following, you refresh, it wouldn't be "Not Following" and won't be added in your profile's Unwatched topic list


SychO9:
I'm sorry there are so many commits, kind of new to Github & Git
Gonna close this and post again in a more proper way

PR #4953, opened 2018-08-22
PMs, Alerts & few GenericLists Responsiveness adjustments

SychO9:
Mostly Tables

### DIFF
--- a/Themes/default/GenericList.template.php
+++ b/Themes/default/GenericList.template.php
@@ -72,7 +72,7 @@ function template_show_list($list_id = null)
 	}
 
 	echo '
-		<table class="table_grid" ', !empty($cur_list['width']) ? ' style="width:' . $cur_list['width'] . '"' : '', '>';
+		<table class="table_grid" id="', $list_id, '" ', !empty($cur_list['width']) ? ' style="width:' . $cur_list['width'] . '"' : '', '>';
 
 	// Show the column headers.
 	$header_count = count($cur_list['headers']);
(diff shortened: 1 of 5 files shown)


SychO9:
total_num

Is this intended or should I add a fix here ?

Gwenwyfar:
I'm not sure, what is the second number for?

XinYenFon:
first one is for error log second is users needs approval

SychO9:
The first is for error log, the second is for member approval

Gwenwyfar (commented):
I don't think we should be using "mobile_hidden" classes, but instead giving a name to each item if they don't have a identifying class/id yet. Not only because this adds a useful class for both desktop and mobile, but it is more semantic and it makes it clear what you are hiding (if you choose to), and easier to modify at will or hide content in different size sets.

SychO9:
I assumed it'd be fine since there's the "hidden" class which hides some elements under 720px screen view, but I'll edit all of this and use specific classes instead, should I remove remove instances where I used the hidden class as well ?

Gwenwyfar:
The responsive file hasn't been worked much in a while, it still needs a major revision (work has been focused in the other files). If there's other hidden classes doing the same they shouldn't be there either.

I'll have to check to confirm, but I'd think so. We may need to go over the ones already being used in another PR as well if that's the case.

jdarwood007:
The "hidden" class is to hide it always.  Its a shortcut to writing style="display: none;" in our code and then having javascript toggle it as needed.  This is generally for elements that again we are using javascript to toggle them on or off.

SychO9:
Should be good now

Gwenwyfar (approved):

PR #4958, opened 2018-08-25
Fix Quoting with SCeditor

### LINKED ISSUE #2845: Inconsistent behaviour between quick edit and normal edit.
Noticed the following 2 issues in the current git version.

Creating a post with a malformed(?) quote tag:

```
[quote=29-04-2015]
- Fixed frontpage link going to wrong location
[/quote]
```

The BB Code displays fine, however if you edit the post, it gets rewritten to

```
[quote]
- Fixed frontpage link going to wrong location
[/quote]
```

(Removes the =TEXT part)
example_screenshot

Second issue

When doing a quick edit in a topic with no replies for a while no warnings appear, while a warning for no recent post does appear for editing the non-quick way.
(Also, It mentions reply while editing? seems a different thing)

edit_warning

jdarwood007:
This fixes #2845

As SMF allows "[quote=text]message[/quote]" it is lost during sceditor when it converts bbc to html and then html back to bbc code.  The simplest method is to tell sceditor that this is a "author", which is how the bbc parser treats it in SMF.  This will cause all bbc to be updated to "[quote author=text]message[/quote]".


jdarwood007:
Also it seems PersonalMessage.php line endings got converted to CRLF somehow.  So correcting that.

PR #4968, opened 2018-08-28
Restores YouTube BBC to parse_bbc()

### LINKED ISSUE #4969: YouTube BBC should not have been removed
#4880 incorrectly removed support for the [youtube] BBC from `parse_bbc()`, due to the mistaken belief that the editor didn't support it. In fact, the editor does support it. (What we don't support is the AutoYouTube plugin for SCEditor, because that causes weird line break issues with the [code] BBC.)

Sesquipedalian:
Fixes #4969

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -1887,6 +1887,13 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 				'before' => '<span style="color: white;" class="bbc_color">',
 				'after' => '</span>',
 			),
+			array(
+				'tag' => 'youtube',
+				'type' => 'unparsed_content',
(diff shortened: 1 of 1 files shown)


jdarwood007 (approved):

MissAllSunday:
If the youtube bbc is indeed restored, I wouldn't use an iframe directly as it could cause security issues. What I would do is instead use a div with a data-attribute for the yt ID and use javascript to either dynamically create an iframe on the DOM  or directly use embed or object HTML5 tags to render the video.

Sesquipedalian:
Thanks for mentioning the possible security issues, @MissAllSunday. I just looked it up, and apparently YouTube now [recommends](https://developers.google.com/youtube/player_parameters#Manual_IFrame_Embeds) adding an `origin` parameter to the src URL of the iframe in order to prevent malicious JavaScript from being able to hijack the YouTube player. I'll open a new issue and submit a quick PR to address that.

PR #4975, opened 2018-08-29
Preventing other profiles from crashing

DiegoAndresCortes:
This will prevent that the user gets an unexpected error (unable to load template) when viewing a different profile.
It only happens if the admin decided to disable permissions to view other profiles to a group and left the send pm options enabled.

Since it'd be the only button active the file would try to load a function to show something right before telling him he's not allowed to view other profiles.

Perhaps there's a better way to resolve this, but this one is the easiest for sure.

### DIFF
--- a/Sources/Profile.php
+++ b/Sources/Profile.php
@@ -357,6 +357,7 @@ function ModifyProfile($post_errors = array())
 					'label' => $txt['report_profile'],
 					'custom_url' => $scripturl . '?action=reporttm;' . $context['session_var'] . '=' . $context['session_id'],
 					'icon' => 'warning',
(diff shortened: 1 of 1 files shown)


jdarwood007:
After testing this I still receive the unable to load template error

I created a new group with unique permissions.  Removed the ability to view membergroups.  Applied to a test user.  Login as test user.  Viewed another profile and get error.

DiegoAndresCortes:
I just tested it back on a clean install and it seems to work fine for me.
Could also check for the existence of a function if preferred, but requires to check it on 2 lines.

SychO9:
I tested this, it did fix the problem

jdarwood007 (approved):

Issue #4972, opened 2018-08-28
Possible bug with permissions in Who.php


DiegoAndresCortes:
So I noticed this yesterday and decided to install a fresh one in a different folder to test it and result is the same.
1- I registered another account to be a regular user and went into action=who
2- With admin account accessed the admin center and went to Permissions page

![indice](https://user-images.githubusercontent.com/3318127/44730498-b901dc00-aaa6-11e8-8c89-0a39ebdc91cd.png)

So the regular user can see what the admin account is doing. It's not showing the actual text of the action, but just because it's supposed to show $txt['who_hidden']
It happens in multiple areas but I'm not sure I've visited all of them.

DiegoAndresCortes:
BTW the title is referring to the permissions to see these actions, not to the actual permissions area, that's just one area I tested to get the picture.

albertlast:
In my eyes it is more a posible privacy issue,
instead of a permission issue.

DiegoAndresCortes:
Of course, by permissions I just meant the lack of permissions from the regular user to see these actions.

Sorck:
IIRC SMF 2.X says something like "Nothing, or nothing you can see" if the user does not have permission to view a specific action. Direct the admin to ?action=admin;area=manageboards and you can see the case I have described even in SMF 2.1.

In Who.php we have ```determineActions()``` that works out what to show. Admin pages use different logic to everything else - the 'area' is treated as the action for the purposes of calculating if you can see it. Basically we just need to add permissions for every admin 'area' in the ```$allowedActions``` array.

However, it's possible the sub-actions aren't being properly permission checked. That's something to look into further.

PR #4976, opened 2018-08-29
Permissions for Who's Online admin/mod centre

Sorck:
Do not reveal admin and mod centre usage unless allowed to see it as per issue SimpleMachines/SMF2.1#4972

* Adds in all admin centre areas (except index) properly to the permissions checking array in Who.php
* Removes fallbacks that say something along the lines of 'using the admin panel' - only those with access to the admin panel should be able to see this.
* Does *not* add in the mod centre areas as they are harder to calculate permission for and have names that are likely to conflict with other actions.

Signed-off-by: James Robson <git@sorck.net>

### DIFF
--- a/Sources/Who.php
+++ b/Sources/Who.php
@@ -273,20 +273,42 @@ function determineActions($urls, $preferred_prefix = false)
 		'ban' => array('manage_bans'),
 		'boardrecount' => array('admin_forum'),
 		'calendar' => array('calendar_view'),
+		'corefeatures' => array('admin_forum'),
 		'editnews' => array('edit_news'),
+		'featuresettings' => array('admin_forum'),
+		'languages' => array('admin_forum'),
+		'logs' => array('admin_forum'),
 		'mailing' => array('send_mail'),
+		'mailqueue' => array('admin_forum'),
 		'maintain' => array('admin_forum'),
 		'manageattachments' => array('manage_attachments'),
 		'manageboards' => array('manage_boards'),
+		'managecalendar' => array('admin_forum'),
+		'managesearch' => array('admin_forum'),
+		'managesmileys' => array('manage_smileys'),
+		'membergroups' => array('manage_membergroups'),
 		'mlist' => array('view_mlist'),
 		'moderate' => array('access_mod_center', 'moderate_forum', 'manage_membergroups'),
+		'modsettings' => array('admin_forum'),
+		'news' => array('edit_news', 'send_mail', 'admin_forum'),
 		'optimizetables' => array('admin_forum'),
+		'packages' => array('admin_forum'),
+		'paidsubscribe' => array('admin_forum'),
+		'permissions' => array('manage_permissions'),
+		'postsettings' => array('admin_forum'),
+		'regcenter' => array('admin_forum', 'moderate_forum'),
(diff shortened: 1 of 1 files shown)


Sources/Who.php
```diff
@@ -405,10 +427,6 @@ function determineActions($urls, $preferred_prefix = false)
 			{
 				if (allowedTo($allowedActions[$actions['action']]))
 					$data[$k] = $txt['whoallow_' . $actions['action']];
-				elseif (in_array('moderate_forum', $allowedActions[$actions['action']]))
-					$data[$k] = $txt['who_moderate'];
-				elseif (in_array('admin_forum', $allowedActions[$actions['action']]))
-					$data[$k] = $txt['who_admin'];
 				else
```
jdarwood007, line 430: We should leave this here, so it still shows the generic admin/mod messages for those without access.

Sorck: Do we consider it a privacy/security issue to show an admin working in the admin centre if that user has no way to view the admin centre?

Assuming that doesn't matter, they could be expanded quite a bit for completeness as admin_forum is not a permission needed for many admin centre areas. We may also need another integration hook to cover any other actions that may want to declare themselves as a mod centre/admin panel page.

We could then either do it as an array with a list of permissions that defines the admin panel, or a list of areas that are within the admin panel.
```php
// If a page has one of these permissions then it is in the approapriate major action, like the admin or moderation centre.
// the admin and moderate items can initially be populated from the 'admin' and 'moderate' keys in $allowedActions
$majorActionPermissions = array(
    'admin' => $allowedActions['admin'],
    'moderate' => $allowedActions['moderate'],
);
// Allow mods to add to the major action areas
call_integration_hook('integrate_who_major_actions', &$majorActionPermissions);

// Check if we should show a major action message
foreach ($majorActionPermissions as $action => $permissions)
    if (!empty(array_dif($permissions, $allowedActions[$actions['action']]))
        $data[$k] = $txt['who_' . $action];
```

Or if it was area based.
```php
// Used to find out if the current area is part of the mod centre/admin centre
$majorActionAreas = array(
    'admin' => array(
        'admin',
        'packages',
        'etc',
    ),
    'moderate' => array(
        'moderate',
    ),
);
// Allow mods to add to the major actions
call_integration_hook('integrate_who_major_actions', &$majorActionAreas);

// Only check against major action messages if we are in a large area
if (isset($actions['area']))
    foreach ($majorActionAreas as $action => $areas)
        if (in_array($action, $areas))
            $data[$k] = $txt['who_' . $action];
```

Benefits:
* mods can add themselves as major actions with sub areas
* mods can declare their areas as being in the admin centre or mod centre

jdarwood007: We have had the setup that way for a long time, even possibly in YabbSE.  So I wouldn't consider it a risk.  Not much can be identified by what they are doing.  Just that they are in a mod or admin area.

The hook sounds great, but need others to provide input on how specific we get.  I would go with the first and let the mod authors handle that.

DiegoAndresCortes: So $allowedActions would be for normal non-admin/mod actions only?
If it's like that it sounds great, I wouldn't mind either of the suggestions both are nice.

PR #4985, opened 2018-09-01
Adds PHP version to issue template

frandominguezl:


fredbuluku:
> implement php in the simple machine to unlock the system.

PR #4987, opened 2018-09-02
Added Port and MB4 Support to install fixes #4929

### LINKED ISSUE #4929: install with db non-standard port
similar fix to https://github.com/SimpleMachines/SMF2.1/commit/4abb70448660a498ff5b666f04c31f2e318a5efe needs to be added so the installer will connect to a non-default port. currently it returns error it cannot connect.

if i add the port after localhost it works fine.

albertlast:
Issue #4929

### DIFF
--- a/other/install.php
+++ b/other/install.php
@@ -721,7 +721,7 @@ function CheckFilesWritable()
 function DatabaseSettings()
 {
 	global $txt, $databases, $incontext, $smcFunc, $sourcedir;
-	global $db_server, $db_name, $db_user, $db_passwd;
+	global $db_server, $db_name, $db_user, $db_passwd, $db_port, $db_mb4;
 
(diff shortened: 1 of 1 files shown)


other/install.php
```diff
@@ -853,14 +853,23 @@ function DatabaseSettings()
 
 		// Attempt a connection.
 		$needsDB = !empty($databases[$db_type]['always_has_db']);
-		$db_connection = smf_db_initiate($db_server, $db_name, $db_user, $db_passwd, $db_prefix, array('non_fatal' => true, 'dont_select_db' => !$needsDB));
+
+		$options = array('non_fatal' => true, 'dont_select_db' => !$needsDB);
```
jdarwood007, line 857: Never noticed the double negative here on $needsDB before.

jdarwood007 (approved):

PR #4993, opened 2018-09-02
Improve secret q/a by using a more modern hashing method.

### LINKED ISSUE #3117: Improve Secret answer hashes in profiles
This is our current handling of Secret answers

```
        'secret_answer' => array(
            'type' => 'text',
            'label' => $txt['secret_answer'],
            'subtext' => $txt['secret_desc2'],
            'size' => 20,
            'postinput' => '<span class="smalltext" style="margin-left: 4ex;">[<a href="' . $scripturl . '?action=helpadmin;help=secret_why_blank" onclick="return reqOverlayDiv(this.href);">' . $txt['secret_why_blank'] . '</a>]</span>',
            'value' => '',
            'permission' => 'profile_password',
            'input_validate' => function (&$value)
            {
                $value = $value != '' ? md5($value) : '';
                return true;
            },
        ),
```

We should improve this, its better than plaintext, but not by much at this point in today's rainbow tables and cpu/gpu speeds.
1. Should we improve this to handle multiple questions/answers?
2. What should we change the method to?  I'm fine with at least going to a salted sha1, but we could also use the password_hash method.

jdarwood007:
Fixes #3117 by using hash_password.  We do not convert old answers as we have no way to verify them.  We accept both md5 and hash_password_verify.  Due to the varying lengths of the two methods, we don't need to try and figure out the length first.  This doesn't attempt to convert the secret question upon successful answers.

### DIFF
--- a/Sources/Reminder.php
+++ b/Sources/Reminder.php
@@ -357,7 +357,7 @@ function SecretAnswer2()
 	$smcFunc['db_free_result']($request);
 
 	// Check if the secret answer is correct.
-	if ($row['secret_question'] == '' || $row['secret_answer'] == '' || md5($_POST['secret_answer']) != $row['secret_answer'])
+	if ($row['secret_question'] == '' || $row['secret_answer'] == '' || (!hash_verify_password($row['member_name'], $_POST['secret_answer'], $row['secret_answer']) && md5($_POST['secret_answer']) != $row['secret_answer']))
 	{
 		log_error(sprintf($txt['reminder_error'], $row['member_name']), 'user');
 		fatal_lang_error('incorrect_answer', false);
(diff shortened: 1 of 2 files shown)


Sources/Reminder.php
```diff
@@ -357,7 +357,7 @@ function SecretAnswer2()
 	$smcFunc['db_free_result']($request);
 
 	// Check if the secret answer is correct.
-	if ($row['secret_question'] == '' || $row['secret_answer'] == '' || md5($_POST['secret_answer']) != $row['secret_answer'])
+	if ($row['secret_question'] == '' || $row['secret_answer'] == '' || (!hash_verify_password($row['member_name'], $_POST['secret_answer'], $row['secret_answer']) && md5($_POST['secret_answer']) != $row['secret_answer']))
```
Sesquipedalian: Please add a comment to explain to future developers that the md5 check is there for the sake of backwards compatibility with old stored values.

Sesquipedalian (changes requested):

jdarwood007:
Added a nice big comment explaining it all.

Sesquipedalian (approved):

PR #4995, opened 2018-09-03
Cleanup ViewMembers to improve mod handling

### LINKED ISSUE #3350: Admin manage members hook placement
Hello,

Wanted to do it myself but thought to ask some opinions first. In ManageMembers.php, we have this:

```
$subActions = array(
    'all' => array('ViewMemberlist', 'moderate_forum'),
    'approve' => array('AdminApprove', 'moderate_forum'),
    'browse' => array('MembersAwaitingActivation', 'moderate_forum'),
    'search' => array('SearchMembers', 'moderate_forum'),
    'query' => array('ViewMemberlist', 'moderate_forum'),
);

call_integration_hook('integrate_manage_members', array(&$subActions));
```

And then later we have the $context['tabs'] getting defined. So if I add something via hook to $subActions, I can't catch $context['tabs'] to add it there as well. (of course if I want it to be a tab)

Shouldn't the $context['tabs'] go to top, between the $subActions and the hook? (or move the hook down below tabs?)

jdarwood007:
Fixes #3350 by moving code around to allow the hook to be called at a better time.

Also improves the code readability by improved tab creations.

isAllowedTo is now called later in the process to allow this, but should not create any notifiable difference.

### DIFF
--- a/Sources/ManageMembers.php
+++ b/Sources/ManageMembers.php
@@ -38,11 +38,14 @@ function ViewMembers()
 	);
 
 	// Default to sub action 'index' or 'settings' depending on permissions.
-	$_REQUEST['sa'] = isset($_REQUEST['sa']) && isset($subActions[$_REQUEST['sa']]) ? $_REQUEST['sa'] : 'all';
+	$context['current_subaction'] = isset($_REQUEST['sa']) && isset($subActions[$_REQUEST['sa']]) ? $_REQUEST['sa'] : 'all';
 
 	// Load the essentials.
 	loadLanguage('ManageMembers');
 	loadTemplate('ManageMembers');
+
+	// Fetch our activation counts.
+	GetMemberActivationCounts();
 
 	// For the page header... do we show activation?
 	$context['show_activate'] = (!empty($modSettings['registration_method']) && $modSettings['registration_method'] == 1) || !empty($context['awaiting_activation']);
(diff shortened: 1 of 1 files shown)


Sorck:
This creates a breaking change - you will find any tabs added by mods can no longer be accessed as $subActions array is used right after the old call_integration_hook().

A similar issue cropped up a week or two back #4971

jdarwood007:
How so?  We call the hook allowing changes to $subAction, then start using it.  Your hook should have done all the work at this point to set itself up in $subAction and any context changes it needs.  After this the sub actions are handled.

I'm just not seeing the logic in what would be broken with this change.

Sorck:
So we first create the ```$subActions``` array https://github.com/SimpleMachines/SMF2.1/blob/2c08e6f8354938bf11bb7d9e0f8a95dc61c8c565/Sources/ManageMembers.php#L32-L38

[old hook call was here]

We then use the contents of array to set ```$_REQUEST['sa']``` to a value that exists in ```$subActions```.
https://github.com/SimpleMachines/SMF2.1/blob/2c08e6f8354938bf11bb7d9e0f8a95dc61c8c565/Sources/ManageMembers.php#L40-L41

With the hook called later on (after ```$context['tabs']``` is setup) you won't be able to add anything useful to $subActions as the logic for selecting which action to show (based on ```$_REQUEST['sa']```) will only ever be called with the initial values.

Also, the way 'is_selected' is determined in ```$context['tabs']``` relies on ```$_REQUEST['sa']``` being set as in the second step.


The suggestion from #3350 of another hook would solve it - keep the current one named as is for backwards compatability and add either a "_post" or maybe "_context" hook?

jdarwood007:
Seems this could be solved by using a $context['current_subaction'] (Think it kinda exists already) and defining that.  Let the hook handle if it changes.  Then make sure all is_selected is handled better.

Sesquipedalian (approved):

PR #5005, opened 2018-09-09
Allow mod authors the option to override the id theme without changing the url

tinoest:
This is to remove the need to edit the source file of Load.php like TinyPortal currently does when it has a custom theme set on a article, code modification is here; https://github.com/Tinyportal/TinyPortal/blob/master/install.xml#L64

I couldn't seem to get load_theme hook to work correctly as it sets other sections before anything else gets a chance to.


jdarwood007:
Shouldn't this be a bit more down as otherwise it could get overridden again by a board theme?

tinoest:
It could be yes, I just went for the same place TP already modifies. 

I’ll move it down later today.

Sesquipedalian (approved):

PR #5023, opened 2018-09-25
 Always uses a fully qualified domain name in EHLO/HELO to SMTP host

### LINKED ISSUE #3997: Using smtp_host as EHLO argument is not a good idea.
Subs-Post.php have this code

```
	if (empty($helo))
		$helo = $modSettings['smtp_host'];

```
But $modSettings['smtp_host'] can be set to something like ssl://smtp.gmail.com which is not hosname or IP as EHLO command should send. I think better use $helo = 'localhost'; here.

Sesquipedalian:
Fixes #3997 by bringing our behaviour into compliance with [RFC 5321](https://tools.ietf.org/html/rfc5321).

### DIFF
--- a/Sources/Subs-Post.php
+++ b/Sources/Subs-Post.php
@@ -1322,22 +1322,9 @@ function smtp_mail($mail_to_array, $subject, $message, $headers)
 		elseif (function_exists('php_uname'))
 			$helo = php_uname('n');
 
-		// Local domains won't cut it, so look up the canonical one for $boardurl in DNS
+		// If the hostname isn't a fully qualified domain name, we can use the host name from $boardurl instead
 		if (empty($helo) || strpos($helo, '.') === false || substr_compare($helo, '.local', -6) === 0 || (!empty($modSettings['tld_regex']) && !preg_match('/\.' . $modSettings['tld_regex'] . '$/', $helo)))
-		{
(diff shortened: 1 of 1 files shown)


Sources/Subs-Post.php
```diff
@@ -1310,17 +1312,37 @@ function smtp_mail($mail_to_array, $subject, $message, $headers)
...
+		elseif (function_exists('php_uname'))
+			$helo = php_uname('n');
+
+		// Local domains won't cut it, so look up the canonical one for $boardurl in DNS
+		if (empty($helo) || strpos($helo, '.') === false || substr_compare($helo, '.local', -6) === 0 || (!empty($modSettings['tld_regex']) && !preg_match('/\.' . $modSettings['tld_regex'] . '$/', $helo)))
+		{
+			$helo = parse_url($boardurl, PHP_URL_HOST);
+
+			if (function_exists('gethostbyaddr') && function_exists('dns_get_record'))
+			{
```
jdarwood007: Shouldn't we test $helo here before going to dns?

Sesquipedalian: Meh. I realized that going to DNS was overkill anyway. If the hostname in $boardurl could get a valid result from a DNS lookup, then that hostname was already perfectly good for these purposes. And if it couldn't, it's not like we have any viable options left anyway.

PR #5024, opened 2018-09-26
Gets rid of some obsolete functions, updates documentation on others

### LINKED ISSUE #4096: Deprecating compatibility functions

Sesquipedalian:
Fixes #4096


Sources/Subs-Editor.php
```diff
@@ -87,13 +86,9 @@ function($m) use (&$i)
...
 /**
- * !!!Compatibility!!!
- * This is no more needed, but to avoid break mods let's keep it
- * Run it it shouldn't even hurt either, so let's not bother remove it
+ * Converts HTML to BBC
+ * As of SMF 2.1, only used by ManageBoards.php (and possibly mods)
  *
- * The harder one - wysiwyg to BBC!
- *
- * @deprecated since version 2.1
```
albertlast, line 96: Is this function and the others really not depracted?

Sesquipedalian: `html_to_bbc()` is still called in a few places in ManageBoards.php, so it isn't actually deprecated. And both `feetchTagAttributes()` and `legalise_bbc()` are called by `html_to_bbc()`, so they aren't either. Only `bbc_to_html()` is truly deprecated.

emanuele45: https://en.wikipedia.org/wiki/Deprecation
> The feature is considered extraneous, and will be removed in the future in order to simplify the system as a whole.

Commenting just because I was the responsible for putting these notes and never actually formalized the meaning with which I put them, that was as a warning for future generation of developers and various coders not to use them, or use them knowing that they will be removed at some point (likely in 3.0, but since 3.0 will allow time travels (Or was it 4? I forgot.) it may be unnecessary).

Sesquipedalian: It seems to me that as long as there is a use case in the base software for converting HTML to BBC, a function like html_to_bbc() isn't extraneous or projected to be removed in the future.

jdarwood007 (approved):
@Sesquipedalian, Can you fix up the merge conflict and this is ready to go.

Issue #4749, opened 2018-05-05
Last post link Browser position problem


gevv:
Hi,

Many just me forum users reported the same problem

Different browsers different versions have the same problem

Click last post link  the browser position does not go to the last message section

Does go to the quick replay text writing area

I have the same problem I tested on different smf 2.1 sites

[video.zip](https://github.com/SimpleMachines/SMF2.1/files/1976429/video.zip)

![untitled 3](https://user-images.githubusercontent.com/5096676/39660645-5697e168-504c-11e8-8953-54318fb6867c.png)

albertlast:
In your video,
the link show direct to the first unread message i guess (since the video is bad on information i had to guess this)

gevv:
animated gif

![ezgif-1-c033a2d182](https://user-images.githubusercontent.com/5096676/39714685-89f2bf22-5233-11e8-9bf1-93d80a8e036a.gif)

albertlast:
like i said before,
the url information are not shown,
without this we cant follow you.

wintstar:
Same thing I already posted here https://github.com/SimpleMachines/SMF2.1/issues/4657. In Google Chrome Browser the bug still exists

albertlast:
like i mention in your cmment i'm not able to reproduce this with chrome

wintstar:
I can only detect it in Chrome. I reinstalled the browser, addons disabled. The bug didn't go away. Maybe it has something to do with the PC settings, processor or something else. I'm not the only one who has the bug.
But it's weird that it's only in chrome for me. That's why I switched to Firefox. What I noticed with Chrome was very negative, that although the browser cache was emptied several times, changes are not visible. Until much later.  It's very strange about this bug.

smfbigguy:
The url is my site. This is a bug. If you click any link that is supposed to get you to the last post in a thread it does not. It sends you to the quick reply box instead of the last post in that thread. I use FF and it happens with that browser.

albertlast:
What i asking for was the options which are sendet,
the domaine i don't care.

wintstar:
@albertlast, my testboard.

/index.php/topic,4.msg103/topicseen.html#msg103
/index.php/topic,4.msg104.html#new

It seems to be browser independent. I have problems with Chrome and everything is fine with Firefox.  smfbigguy writes Firefox causes problems. I think Javascript is the cause of this problem on some PC / server configurations.

albertlast:
[...]

PR #5039, opened 2018-10-10
Added sceditor undo js fixes #5035

### LINKED ISSUE #5035: sceditor.undo.js
took a download today of recent file getting error message when visiting posts, The file "sceditor.undo.js" could not be located on the current or the default theme and therefore it wasn't included in the minified file.

No changes made to coding and its using most up todate version took it about 15 mins ago

albertlast:
On the issue #5035 i see two ways to fix this:
remove the line or add the file again.

This pr take the last approache.
I guess the disabling of this plugin is somekind of temp solution and files removing and adding
could hurt in the upgrade process...
keep the file and update this when undo is fixed is a good way i guess.


Sesquipedalian:
No, adding the plugin back in at this point will only bring back the bug in #4749. We can add it once a fixed version is available from SCEditor, but not until then.

albertlast:
When I read the config right,
Then the plugin keeps disabled,
Even this files get loaded.

Sesquipedalian:
Why would we want to load an unused, malfunctioning file?

jdarwood007:
I would not mind seeing it readded once https://github.com/samclarke/SCEditor/issues/711 has been fixed.  This is the underlaying bug that caused its removal.

The bug seems related to the SCEditor itself having had adaption changes and the plugin not being updated to handle those changes.  Basically sounds like the plugin is technically not compatible with SCEditor in its current version.

Sesquipedalian:
Thanks @jdarwood007. Once the plugin has been fixed and released, I'll be glad to add it back in again.

PR #5047, opened 2018-10-12
Run cache check only when enabled

albertlast:
I try to do some performance optimization,
this is one of the results.

This pr is done.

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -3316,7 +3316,11 @@ function loadDatabase()
 */
 function loadCacheAccelerator($overrideCache = null, $fallbackSMF = true)
 {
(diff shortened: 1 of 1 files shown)


jdarwood007:
Can it check to make sure override is false still?  I added that as a future thing for customizations or complicated workarounds.

albertlast:
You mean?
``if (empty($cache_enable))`` --> ``if (empty($cache_enable) && empty($overrideCache))``

jdarwood007:
Yes, or I guess is_null since that is what is the default value.

Just ensuring if we are attempting to override the normal, that we can bypass the check.

albertlast:
When  null and false are the same,
than it's smf coding standard to use empty.

jdarwood007 (approved):

PR #5048, opened 2018-10-12
Reduce the call of functions

albertlast:
Well the idea less amount of context switch = faster,
but to be ownest my php env is so "fast" that i didn't measure any improvments.

What i notice in my profile data,
the smf_db_fetch_assoc  and smf_db_fetch_row take "big" amount of time in compare to the native part which is called internaly:
grafik

Since i believe the script runs long enough that a day switch is between,
I define $now as static -> again less amount of calls.

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -49,7 +49,8 @@ function reloadSettings()
 		if (!$request)
 			display_db_error();
 		$rows = $smcFunc['db_fetch_all']($request);
-		foreach ($rows as $row) {
+		foreach ($rows as $row)
+		{
 			$modSettings[$row['variable']] = $row['value'];
 		}
 		$smcFunc['db_free_result']($request);
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -1809,8 +1809,7 @@ function loadTheme($id_theme = 0, $initialize = true)
 				)
 			);
 			// Pick between $settings and $options depending on whose data it is.
-			$rows = $smcFunc['db_fetch_all']($result);
-			foreach ($rows as $row)
+			foreach ($smcFunc['db_fetch_all']($result) as $row)
 			{
 				// There are just things we shouldn't be able to change as members.
 				if ($row['id_member'] != 0 && in_array($row['variable'], array('actual_theme_url', 'actual_images_url', 'base_theme_dir', 'base_theme_url', 'default_images_url', 'default_theme_dir', 'default_theme_url', 'default_template', 'images_url', 'number_recent_posts', 'smiley_sets_default', 'theme_dir', 'theme_id', 'theme_layers', 'theme_templates', 'theme_url')))
@@ -1825,7 +1824,6 @@ function loadTheme($id_theme = 0, $initialize = true)
 					$themeData[$row['id_member']][$row['variable']] = substr($row['variable'], 0, 5) == 'show_' ? $row['value'] == '1' : $row['value'];
 			}
 			$smcFunc['db_free_result']($result);
-			unset($rows);
 
 			if (!empty($themeData[-1]))
 				foreach ($themeData[-1] as $k => $v)
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -48,13 +48,11 @@ function reloadSettings()
 		$modSettings = array();
 		if (!$request)
 			display_db_error();
-		$rows = $smcFunc['db_fetch_all']($request);
-		foreach ($rows as $row)
+		foreach ($smcFunc['db_fetch_all']($request) as $row)
 		{
 			$modSettings[$row['variable']] = $row['value'];
 		}
 		$smcFunc['db_free_result']($request);
-		unset($rows);
 
 		// Do a few things to protect against missing settings or settings with invalid values...
 		if (empty($modSettings['defaultMaxTopics']) || $modSettings['defaultMaxTopics'] <= 0 || $modSettings['defaultMaxTopics'] > 999)
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -48,9 +48,14 @@ function reloadSettings()
 		$modSettings = array();
 		if (!$request)
 			display_db_error();
-		while ($row = $smcFunc['db_fetch_row']($request))
-			$modSettings[$row[0]] = $row[1];
+		$rows = $smcFunc['db_fetch_all']($request);
+		foreach ($rows as $row) {
+			$modSettings[$row['variable']] = $row['value'];
+		}
+		//while ($row = $smcFunc['db_fetch_row']($request))
+		//	$modSettings[$row[0]] = $row[1];
 		$smcFunc['db_free_result']($request);
+		unset($rows);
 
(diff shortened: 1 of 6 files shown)


Sources/Load.php
```diff
@@ -48,9 +48,13 @@ function reloadSettings()
...
 			display_db_error();
-		while ($row = $smcFunc['db_fetch_row']($request))
-			$modSettings[$row[0]] = $row[1];
+		$rows = $smcFunc['db_fetch_all']($request);
+		foreach ($rows as $row)
+		{
+			$modSettings[$row['variable']] = $row['value'];
+		}
 		$smcFunc['db_free_result']($request);
+		unset($rows);
```
jdarwood007, line 54: Do we need to assign to the temp $rows variable or can we just call a straight through?

albertlast: I don't understand your question.
What you mean with call straight through ( I understand the words but not what kind of coding style you want)

albertlast:
To make my point clear,
i build a test script
```
<?php
$db_server = 'localhost';
$db_name = 'smf';
$db_user = 'postgres';
$db_passwd = '****';
$querylimit  = 100;

$sql = 'SELECT id_msg, id_topic FROM smf_messages limit ' . $querylimit;

// smf row api

$start = microtime(true);

while ($row = smf_db_fetch_row($request))
    $trash = $row;

$time_smfrow = microtime(true) - $start;
unset($request);

// php row api

$start = microtime(true);

while ($row = pg_fetch_row($request))
    $trash = $row;

$time_phprow = microtime(true) - $start;
unset($request);

// smf assoc api

$start = microtime(true);

while ($row = smf_db_fetch_assoc($request))
    $trash = $row;

$time_smfassoc= microtime(true) - $start;
unset($request);

// php assoc api

$start = microtime(true);

while ($row = pg_fetch_assoc($request))
    $trash = $row;

$time_phpassoc= microtime(true) - $start;
unset($request);

// smf all api

$start = microtime(true);

if (!empty($rows))
    foreach ($rows as $row)
        $trash = $row;

$time_smfall= microtime(true) - $start;
unset($request);
unset($rows);


var_dump($time_smfrow);
var_dump($time_phprow);
var_dump($time_smfassoc);
var_dump($time_phpassoc);
var_dump($time_smfall);

/**
 *
 */
{
	// Return the right row.
}

/**
 * A PostgreSQL specific function for tracking the current row...
 *
 * @param bool|int $counter The row number in the result to fetch (false to fetch the next one)
 * @return array The contents of the row that was fetched
 */
function smf_db_fetch_row($request, $counter = false)
{
	global $db_row_count;

	if ($counter !== false)
		return pg_fetch_row($request, $counter);

	// Reset the row counter...
	if (!isset($db_row_count[(int) $request]))
		$db_row_count[(int) $request] = 0;

	// Return the right row.
	return @pg_fetch_row($request, $db_row_count[(int) $request]++);
}

/**
 * Get an associative array
 *
 * @param int|bool $counter The row to get. If false, returns the next row.
 * @return array An associative array of row contents
 */
function smf_db_fetch_assoc($request, $counter = false)
{
	global $db_row_count;

	if ($counter !== false)
		return pg_fetch_assoc($request, $counter);

	// Reset the row counter...
	if (!isset($db_row_count[(int) $request]))
		$db_row_count[(int) $request] = 0;

	// Return the right row.
	return @pg_fetch_assoc($request, $db_row_count[(int) $request]++);
}
````

Output:
D:\Programme\xampp2\htdocs\test7.php:70:float 0.00078988075256348

D:\Programme\xampp2\htdocs\test7.php:71:float 0.00030303001403809

D:\Programme\xampp2\htdocs\test7.php:72:float 0.00083398818969727

D:\Programme\xampp2\htdocs\test7.php:73:float 0.00033998489379883

D:\Programme\xampp2\htdocs\test7.php:74:float 0.00011301040649414


We see fetch_all is the best and
the smf api add a hugh amount of "overhead" to the native/php call.

File: Sources/Load.php (hunk 1 above)
Comment (jdarwood007) [reply Sources/Load.php:54]:
```
		foreach ($smcFunc['db_fetch_all']($request) as $row)
```

Comment (albertlast) [reply Sources/Load.php:54]: Look like it works,
needed to change the api of db_fetch_all a little bit (it's return now always an array).

**Comment by jdarwood007** [comment]
I will need to do some digging but see this comment in the php documentation:

> Available only with mysqlnd.

> As mysqli_fetch_all() returns all the rows as an array in a single step, it may consume more memory than some similar functions such as mysqli_fetch_array(), which only returns one row at a time from the result set. Further, if you need to iterate over the result set, you will need a looping construct that will further impact performance. For these reasons mysqli_fetch_all() should only be used in those situations where the fetched result set will be sent to another layer for processing.

We need to check to ensure that mysqlnd only wont be a problem.  The memory usage is a note, but shouldn't cause that much extra memory I hope.

**Comment by albertlast** [comment]
Okay i was thinking (stupid me) that the fetch_assoc api on smf side is the same across mysql and pg,
but in reality the mysql got only request, while pg got the function to set the offset.

So i decide to drop the pg feature and use directly the pg_* functions,
this are only 3x slower as the fetch_all stuff.

For mysqli fetch_all I added a check and a fallback logic and
define that fetch_all runs in assoc mode.

**Comment by live627** [comment]
@jdarwood007 would using fetch_all be any different memory-wise than loading each row into a array? I'm just trying to ensure that we can proceed despite the warning.

**Comment by jdarwood007** [comment]
From a logical standpoint yes, because we overwrite $row and thus keep using the same data in memory/freeing up old one automatically.  This will need to dump it all into a single array for usage to parse it out.

**Comment by albertlast** [comment]
Since we talk in the peak about ~100 rows,
the memory used is not relevant.

**Comment by albertlast** [comment]
By extending my script:
```
// get memory

$querylimit  = 10;
$sql = 'SELECT id_msg, id_topic FROM smf_messages limit ' . $querylimit;

$start = memory_get_usage();
var_dump(memory_get_usage() - $start);
unset($start);
unset($rows);
unset($request);

$querylimit  = 100;
$sql = 'SELECT id_msg, id_topic FROM smf_messages limit ' . $querylimit;

$start = memory_get_usage();
var_dump(memory_get_usage() - $start);
unset($start);
unset($rows);
unset($request);

$querylimit  = 1000;
$sql = 'SELECT id_msg, id_topic FROM smf_messages limit ' . $querylimit;

$start = memory_get_usage();
var_dump(memory_get_usage() - $start);
unset($start);
unset($rows);
unset($request);
```
i get this results (in bytes):
D:\Programme\xampp2\htdocs\test7.php:84:int 4256

D:\Programme\xampp2\htdocs\test7.php:95:int 41744

D:\Programme\xampp2\htdocs\test7.php:106:int 404720

So we walkt in worst case about ~41kb in point of time (after the loop the array is gone and used ram get free)

With select * on messages table we get:
D:\Programme\xampp2\htdocs\test7.php:84:int 22304

D:\Programme\xampp2\htdocs\test7.php:95:int 225680

D:\Programme\xampp2\htdocs\test7.php:106:int 2250440

Issue #5009, opened 2018-09-10
[minor] Displaying: Unapproved account deletions


SychO9:
#### Description
so You have to approve both registrations and account deletions, when you go to Manage Members > Awaiting approval and switch between the two filters, the current filter is always set to
**Displaying:** Unapproved account deletions
it doesn't change

### Steps to reproduce
1. Create two account, make it so that the first one requires approval, and turn on Account Deletion approval
2. Once you get both notifications, go to Awaiting approval table and switch between the filters

### Environment (complete as necessary)
- Version/Git revision:  (77419e0)[https://github.com/SimpleMachines/SMF2.1/commit/77419e05f0ea79ffbef095ec70b49351ecf2416a]
- Database Type: MySQL
- Database Version: 5.6.41-84.1
- PHP Version: 7.0.19

Issue #5008, opened 2018-09-10
Members waiting for approval


DiegoAndresCortes:
#### Description
I was wondering why the admin is not notified about members waiting for approval whenever the option is changed in the registration settings.
If I switch from _Admin approval_ to _Immediate registration_ we not longer know if there are still users with their account waiting to be activated.
It could happen that a forum has multiple administrators and someone changed this setting without notice for the others and they didn't approve all of the members.
Has this been discussed before?

### Steps to reproduce
1. Enable _Admin Approval_ in registration settings
2. Register an account, you will get an information message saying an Admin needs to approve your account.
3. Admin switches to _Immediate Registration_. He's no longer aware of users waiting for approval until he visits the members section in the admin center.

Issue #4998, opened 2018-09-07
Add/Remove buddy profile link


SychO9:
#### Description
The link to add or remove someone as a buddy shows if $context['can_have_buddy'] is true
but when someone isn't able to edit their own account settings, it doesn't show up Even though you can still add/remove anyone as a buddy from your profile buddy list
noticed this in 2.0 and it's the same in 2.1

### Steps to reproduce
1. Remove the permission to edit own account settings from a certain group
2. Add/Remove buddy link doesn't show up but you can still add/remove from your profile

### Environment (complete as necessary)
- Version/Git revision: 74e6905
- Database Type: MySQL
- Database Version: 5.6.41-84.1
- PHP Version: 7.0.19

### Additional information/references

PR #5049, opened 2018-10-12
Displaying filter, Add/Remove buddy, unapproved Members amount

SychO9:
1. Fixes Displaying X filter #5009 
For example If you change the type of approval from new registrations to account deletions or underaged approvals the Displaying x filter message doesn't switch to the current filter
This pr also makes the display of current filter better.

2. Fixes Add/Remove Buddy link permission in profile #4998 
The buddy/ignore list section of the profile uses profile_extra_own perm, however the Buddy link in another person 's profile uses profile_identity_own perm, so they have to both use the same perm otherwise it doesn't make sense

3. Fix unapproved members logic #5008 
The logic used checks if the registration method is approval or profile deletions require approval in order to show the amount of unapproved member registrations/deletions which is wrong. If you have X unapproved members but then you change the registration method to anything other than approval AND you turn off profile deletion approval. You no longer get the amount of unapproved members.

It should just check if there's anything to approve at all and show how many, no needless checks

### DIFF
--- a/Sources/ManageMembers.php
+++ b/Sources/ManageMembers.php
@@ -1012,8 +1012,8 @@ function MembersAwaitingActivation()
 	if (!empty($context['show_filter']) && !empty($context['available_filters']))
 		$listOptions['additional_rows'][] = array(
 			'position' => 'above_column_headers',
-			'value' => '<strong>' . $txt['admin_browse_filter_show'] . ':</strong> ' . $context['available_filters'][0]['desc'],
-			'class' => 'smalltext floatright',
+			'value' => '<strong>' . $txt['admin_browse_filter_show'] . ':</strong> ' . ((isset($context['current_filter']) && isset($txt['admin_browse_filter_type_'.$context['current_filter']])) ? $txt['admin_browse_filter_type_'.$context['current_filter']] : $context['available_filters'][0]['desc']),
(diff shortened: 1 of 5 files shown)


Sesquipedalian:
Thanks for this, @SychO9. Could you please edit the initial comment to add a description of what the pull request does?

Also, please make sure the description includes the phrase "Fixes #______" (where the blank is an issue number) for any and all issues that this solves. Not only does that make it easier for us to know exactly what issues your PR addresses when we review it, but GitHub has a nice feature where it will automatically close the issues that you mention using that phrase when the PR is merged. That makes it much easier for us to manage the list of issues.

SychO9:
Yea should've done that my bad, done.
Although the commits messages do reference the issues so it'll close them if merged

SychO9:
Forgot to mention, I changed coppaType value from 1 to 0 because 1 refers to "Approve underaged accounts" and 0 refers to "Reject underaged accounts" so instead of NOT incrementing unapprovedMembers when coppaType != 1 it should not increment when != 0

You can test it by turning on coppa approval  and then registering with underaged account, the amount of unapproved members will Not increment

jdarwood007 (approved):

jdarwood007:
FYI, the Fixes #### should be in the commit message.  GH will auto close any issues with Fixes catch.

PR #5050, opened 2018-10-12
Remove the mobile hidden class

SychO9:
Signed-off-by: SychO <sychocouldy@gmail.com>

### DIFF
--- a/Sources/ManageMembers.php
+++ b/Sources/ManageMembers.php
@@ -470,7 +470,6 @@ function ViewMemberlist()
 							'member_name' => false,
 						),
 					),
-					'class' => 'hidden',
 				),
 				'sort' => array(
 					'default' => 'member_name',
@@ -523,7 +522,6 @@ function ViewMemberlist()
 							'member_ip' => false,
 						),
 					),
-					'class' => 'hidden',
 				),
 				'sort' => array(
 					'default' => 'member_ip',
@@ -563,7 +561,6 @@ function ViewMemberlist()
 
 						return $difference;
 					},
-					'class' => 'hidden',
 				),
 				'sort' => array(
 					'default' => 'last_login DESC',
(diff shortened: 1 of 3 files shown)


Sesquipedalian:
Is there a particular issue that this is meant to address?

SychO9:
Basically the hidden class was introduced to replace "style='display:none;'" especially for js #5036 
I found out that the hidden class was also used to hide elements on mobile view, which is something we don't want to have. so this pr removes that specific usage and hides the elements in a better way.

Incase you meant if there were existing issues addressing this, there aren't.

jdarwood007:
@Gwenwyfar Thoughts?

Sounds like a fix then.  We should have checked that, but hidden was too generic of a class for targeting just mobile usage.  Especially considering its global usage across other projects.

SychO9:
Indeed, good thing it was only used in these couple of files, makes it easier to get rid of.

XinYenFon:
or you could have renamed hidden to mobile_hidden :) using less generic classes in core template is not really a good way to do things.

SychO9:
True, but considering the CSS currently hides elements using unique class names and not generic ones, it'd be somewhat messy to add a generic class and use it in only a couple of files.

However it'd be very practical for mods to have a few classes that hide elements in different screen sizes.

XinYenFon:
For the record, first draft of responsive(css) was very hardcoded, end shape was designed to use more generic classes, but since I had less time and motivation it stayed as is. There should be mobile_480 / mobile_720 / mobile_hidden to hide things not `#ID > Class`

SychO9:
I also use generic classes as such for my themes.
Whether we should use generic classes or not is something that needs discussion, I'm only following how the team does things
https://github.com/SimpleMachines/SMF2.1/pull/4953#pullrequestreview-149360334

And if it was decided that generic classes Should be used. I'd make the changes.
There are advantages in both using them and not in my opinion so I don't mind either ways.

Gwenwyfar (approved):

Gwenwyfar:
Using mobile_hidden classes isn't very semantic nor very practical/versatile, so this looks good to me. We could certainly add them in the css for use by mods and custom themes but I don't think we should use them ourselves.

PR #5052, opened 2018-10-13
Call less timeformat in boardindex

albertlast:
Another pr from the category less call more speed,
this pr got as focus to reduce the calls of timeformat,
which is one of the pricy functions we got.

When we got board with many sub (not sub sub) boards,
than we call for everyone the timeformat function,
because it could be this one with the lasted post.

End of the day only one sub board (with his sub sub ... boards) can win -> we show only one,
this code move the logic for calling the timeformat at the end where we know how is the winner.

This pr is completed.


jdarwood007:
@albertlast 
Is there any reason we couldn't fix up the moderators foreach loop as well to prevent doing 2 foreach loops?  Your code is a bit cleaner since your using & calls, so it could be done for the moderators as well.

albertlast:
I don't understand where the connection to this pr is.
But could you link this loc which you got in your mind.

albertlast:
Combined both logic in once.

jdarwood007 (approved):

PR #5054, opened 2018-10-14
installer mysql remove mysql_errno call fixes  #5055

### LINKED ISSUE #5055: Installer call mysql_errno when table already exists
#### Description
The installer called the mysql_errno,
when you install on a mysql env smf where already some tables exists.

https://github.com/SimpleMachines/SMF2.1/blob/96a42cadbb181ad961dc873ca8e835d7068946f6/other/install.php#L1203-L1208

### Steps to reproduce
1. 
2. 

- Version/Git revision: newest
- Database Type:
- Database Version:
- PHP Version:

albertlast:
This place called the mysql_errno,
in my php 7.2 env is that deadly.

fixed by removing the check, call always mysqli_errno.

issue:  #5055

### DIFF
--- a/other/install.php
+++ b/other/install.php
@@ -1199,13 +1199,9 @@ function DatabasePopulation()
 
 		if ($smcFunc['db_query']('', $current_statement, array('security_override' => true, 'db_error_skip' => true), $db_connection) === false)
 		{
-			// Use the appropriate function based on the DB type
-			if ($db_type == 'mysql' || $db_type == 'mysqli')
-				$db_errorno = $db_type . '_errno';
-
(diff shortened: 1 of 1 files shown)


live627:
Could that perhaps use $smcFunc instead?

albertlast:
dunno,
atm it's call mysql_errno or mysqli_errno,
in the time wher it called i don't know which function are initialized.
and i don't want spend more time on this.

So feel free create a pr which take another route.

live627:
...aaaand it's gone 😛 

What I suggest doesn't exist.

jdarwood007 (approved):

PR #5057, opened 2018-10-14
Add Legacy Settings to Settings file and template

### LINKED ISSUE #5027: Upgrade to 2.1 - Unnecessary UTF8 step
#### Description
I just upgraded a test forum to 2.1.  It was already utf8, but the utf8 migration steps were repeated.  

I think when you choose the option to move to a new settings file, it may miss some of the old settings?

### Steps to reproduce
1. Have a 2.0 forum that's aready utf8
2. Upgrade it to 2.1; choose the 'migrate to new settings' option

- Version/Git revision: Latest as of today
- Database Type: mysql
- Database Version: 5.7
- PHP Version:5.6

jdarwood007:
Fixes #5027


sbulen (commented):
Wont this bypass the utf8 conversion step for folks who aren't on utf8?

In 2.0, this is set for those who have run the conversion in the admin panel, but not set for those who haven't.  We need to account for both possibilities.

jdarwood007:
It shouldn't because the modSettings also needs to be set to UTF-8.  So this would fix Settings.php, then after it runs once, the Settings table has this info and the conversion should not occur again.

I also fixed it to bring over the setting, which should get updated with the rest of the conversion process.

sbulen:
Just tested & confirmed the utf8 step was bypassed appropriately for a DB already UTF8.

PR #5058, opened 2018-10-14
change to select profile field broke backward compatibility

Underdog-01:
Recent changes to the profile fields template broke some backward compatibility and restricted select entries to non-numeric.

Imo non-numeric only for select entries was not necessary.
I suggest having devs use a disabled array to add/restrict disabled options.
Also having the name blank is not proper HTML.. it is supposed to be disabled with a proper name entry.


sbulen:
The change was part of a timezone fix, & it affected four templates, calendar, post, profile & register.  (#4708 )

Is there a reason you focused only on the profile template?

Underdog-01:
I didn't look up other select drop-boxes as of yet but came across the profile issue with the Arcade mod that I support.
Any mod using strictly numerical key values for the options will have them all disabled. 
Unfortunately with the current setup (assuming this may be uniform for other select drop-downs) it will require mod devs to update their mods to have the numerical values changed.
Imo using numerical values for select options may be quite common therefore I would push for a different solution to what you described.

Underdog-01:
This edit will likely be altered.
The specifics can be read here: 
https://github.com/SimpleMachines/SMF2.1/issues/5060

Themes/default/Profile.template.php
```diff
@@ -1560,7 +1560,7 @@ function template_edit_options()
 					if (is_array($field['options']))
 						foreach ($field['options'] as $value => $name)
 							echo '
-							<option', is_numeric($value) ? ' value="" disabled' : ' value="' . $value . '"', $value === $field['value'] ? ' selected' : '', '>', $name, '</option>';
+							<option', (!empty($field['disabled']) && is_array($field['disabled']) && in_array($value, $field['disabled']) ? ' disabled' : ''), ' value="' . $value . '"', $value === $field['value'] ? ' selected' : '', '>', $name, '</option>';
```
Sesquipedalian: The basic idea here is good, but more work needs to be done before this could be merged without breaking existing functionality.

1. Please use `$field['disabled_options']` rather than `$field['disabled']`.
2. Add a line after [this one](https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Profile-Modify.php#L542) that sets 'disabled_options' to an array populated with any and all numeric keys returned from `smf_list_timezones()`.

Sesquipedalian (changes requested):

Themes/default/Profile.template.php
Underdog-01: I can do those edits but want to ask a question prior to pushing the changed:
When viewing the page source of a disabled tag it just shows value ... it does not show value=""
Would you like that same behavior or not to bother omitting the value as I suggested?
Keep in mind that disabled elements are not submitted with the form (they are ignored).

Underdog-01:
Requested changes have been applied.

Although in this case range(0, (count(smf_list_timezones()) -1)) will yield the same result.
The actual keys of that array are not numeric by default.

Am I misunderstanding the possible content of that array?
Are you attempting to thwart mods, edits and/or user inputs that may add numeric keys to that array but by default the keys are not numeric?
If this is the case then my edit to the profile source file needs to have one of the array_keys() function removed.

Underdog-01:
Your first sentence in this comment answered my question:
https://github.com/SimpleMachines/SMF2.1/issues/5060#issuecomment-430080990

It has been changed to what I described above.

Sesquipedalian (approved):

Sesquipedalian:
Thanks for your contribution. 🙂

PR #5062, opened 2018-10-16
$cache_enable should be defined as a global var in SSI

### LINKED ISSUE #5061: Should the cache work under SSI?
#### Description
Because right now it doesn't work.

### Steps to reproduce
1. Enable file cache
2. cache get/put do nothing

- Version/Git revision:last Thursday
- Database Type:blah
- Database Version:blah
- PHP Version:7.2

Missing global declaration

This implies that I have a fix and will send it in upon confirmation.

live627:
Fixes #5061


jdarwood007 (approved):

jdarwood007:
Don't think I need to test this one, since it looks like a simple global adding.

PR #5072, opened 2018-10-20
Add a new page for users to disable tfa

### LINKED ISSUE #4941: disable tfa
seems that when you want to disable tfa, it just does it when you click on the link. i think it should require you to enter your password at least or something so you dont disable it by mistake.

live627:
Fixes #4941

The tfa code generally is sloppy and feels rushed.


live627:
Oh wow. Seems that I can't rely on Profile.php to do its thing because of its specific code order. IOW, this here pull request simply does not work. I cannot use the built-in password check and check for the save request.

Thus, I can either
- use the integration hook, that requires editing Profile.php
- Reinvent the wheel as is done in `tfasetup`. This seems to be the lest insane option.

live627:
I'm seriously overthinking this.

jdarwood007:
Is this ready or still a WIP?

live627:
is done

Sesquipedalian (approved):
Works nicely

PR #5076, opened 2018-10-21
Replace php_os with directory_separator to detect windows

albertlast:
php_os is the os where the php version was build, could lead in the wrong direction and it's slow
10k loop:
php_os 0.022707939147949
directory_separator: 0.0039470195770264


jdarwood007:
Well its a constant rather than a function, so would expect it to be much faster.

Can the other calls just make use of the $context['server']['is_windows'] check or should they keep doing the constant check?

albertlast:
Since i don't want introduce new dependency across the code and
the check is really less time consumption.
I see no need todo so.

jdarwood007 (approved):

PR #5078, opened 2018-10-21
use js to generate qr fixes #5000

### LINKED ISSUE #5000: 2FA secret leaked to third party and possibly stored in browsing history
#### Description
When the QR code with the authenticator setup secret is generated, it is a link to a third party (Google) QR code generating service. This means the user secret is being leaked during 2FA setup via browsing history and sent to a third party. 

https://github.com/SimpleMachines/SMF2.1/blob/74e690529fe0e9ed4f7ba9c077ce10be2677b7d8/Sources/Class-TOTP.php#L348-L352

Partial improvement could be gained from a self hosted QR code generation script with a link, even better would be a base 64 encoded image included in the page source. If the server doesn't have GD then we can just let the user manually type in their authenticator secret into their apps.

albertlast:
used the fork of qrcode.js https://github.com/KeeeX/qrcodejs

change the url removed the google api call and removed the urlencode (dunno if we maybe need this)

someone need to test this.

issue #5000

### DIFF
--- a/Themes/default/Profile.template.php
+++ b/Themes/default/Profile.template.php
@@ -3155,6 +3155,7 @@ function template_tfasetup()
 									', $txt['tfa_code_invalid'], '
 								</div>';
 
+	loadJavaScriptFile('qrcode.js', array('force_current' => false, 'validate' => true));
 	echo '
 								<input type="text" name="tfa_code" size="25"', !empty($context['tfa_error']) ? ' class="error"' : '', !empty($context['tfa_value']) ? ' value="' . $context['tfa_value'] . '"' : '', '>
 								<input type="submit" name="save" value="', $txt['tfa_enable'], '" class="button">
@@ -3164,7 +3165,10 @@ function template_tfasetup()
 						</form>
 					</div>
 					<div class="floatright tfa_qrcode">
-						<img src="', $context['tfa_qr_url'], '" alt="">
+						<div id="qrcode"></div>
+						<script type="text/javascript">
+							new QRCode(document.getElementById("qrcode"), "' , $context['tfa_qr_url'],'");
+						</script>
 					</div>';
 
 	if (!empty($context['from_ajax']))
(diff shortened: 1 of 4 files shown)


albertlast:
@jdarwood007 the related issue is rc1 related...
so i'm not sure if than a good plane to set this pr as rc2

jdarwood007:
Do you think we can get enough testing to ensure nothing breaks for RC1?

albertlast:
how ever decided to make the issue rc1,
was thinking that i guess.

but i got no test env for this here,
so i can't test this pr.

I only want to see that rc1 get released so i created some pr for the open rc1 issues.

jdarwood007:
I will move it over to RC1, @Sesquipedalian can move it to RC2 if needed.

Sesquipedalian:
When I go to `?action=profile;area=tfasetup`, no QR code is shown. The JavaScript error log reports the following:
```
[Error] ReferenceError: Can't find variable: QRCode
	Global Code (index.php:391)
```
 It appears that the JavaScript isn't being loaded correctly.

Sesquipedalian (changes requested):
The JavaScript does not appear to be loading correctly.

Themes/default/Profile.template.php
```diff
@@ -3155,6 +3155,7 @@ function template_tfasetup()
 									', $txt['tfa_code_invalid'], '
 								</div>';
 
+	loadJavaScriptFile('qrcode.js', array('force_current' => false, 'validate' => true));
```
Sesquipedalian: This is probably the issue. As a rule, `loadJavaScriptFile()` needs to be called in the source files, not the template files.

albertlast: Should i call this in a different way or
move the call to the source file?

Sesquipedalian: Just move that call to the source file. 🙂

albertlast:
So i move the logic,
please test again.

albertlast:
Sorry i didn't notice that sesqui provide the action url,
by using index.php?action=profile;area=tfasetup
i get now a qr code shown which my phone reconized as 2fa code -> the 2fa app open.

Sesquipedalian (approved):

Sesquipedalian:
This works very nicely. 👍

PR #5075, opened 2018-10-20
Update index.css

live627:
Fixes #5064

Fix suggested by @Antes

Issue #5064, opened 2018-10-16
Printing randomness makes the page go berserk


live627:
![screenshot_2018-10-16 login](https://user-images.githubusercontent.com/431799/47052632-06710000-d15e-11e8-9a35-dcb07ed1d688.png)

### Steps to reproduce
1. print something in a source file
2. view in firefox
3. this will come up when someone gets an error in debug mode since they get spit out at the top

jdarwood007:
Is this something we can logically fix?

@Gwenwyfar or @Antes have feedback on that?  Assuming this is printing text before the < HTML> declaration.

XinYenFon:
Try changing `.popup_container` height to max-height

PR #5085, opened 2018-10-24
Revert "Update index.css"

### LINKED ISSUE #5084: Popups not displaying properly
#### Description
Our popup overlays are now displaying only at the top of the window rather than overlaying the whole window.  For example:
screen shot 2018-10-23 at 3 06 10 pm

### Steps to reproduce
1. Log in
2. 

- Version/Git revision: 207d2dd
- Database Type:
- Database Version:
- PHP Version:

Introduced in #5075
@live627

live627:
Reverts SimpleMachines/SMF2.1#5075

Fixes #5084 
Reopens #5064

Quick fix until we can figure out something better.

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -2630,7 +2630,7 @@ dl.addrules dt.floatleft {
 	top: 0;
 	left: 0;
 	width: 100%;
-	max-height: 100%;
+	height: 100%;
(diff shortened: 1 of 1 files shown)


jdarwood007:
@Gwenwyfar Is this the right thing to do until more time can be involved in figuring out the right solution?

PR #5087, opened 2018-10-25
Escapes null character when saving data to the cache

### LINKED ISSUE #5083: Possible issue in file cache backend
#### Description
There's always been some concern about the reliability of the file cache backend. I think we have a case where it actively has been shown to break.

If the cached content contains a null byte, the addcslashes call to write the data will truncate the string.

### Steps to reproduce
1. Have content that contains a null byte
2. Push it to the cache with cache_put_data
3. Observe result file cache is malformed

- Version/Git revision: master
- Database Type: n/a
- Database Version: n/a
- PHP Version: n/a

See https://www.simplemachines.org/community/index.php?topic=560471.0 for more including a dump of a modSettings cache entry at 20489 bytes, so not a file write boundary issue (like we saw back in the day). I don't know for sure if this is the problem but it's a reasonable bet right now.

Sesquipedalian:
Fixes #5083

### DIFF
--- a/Sources/CacheAPI-smf.php
+++ b/Sources/CacheAPI-smf.php
@@ -100,7 +100,7 @@ class smf_cache extends cache_api
 			@unlink($cachedir . '/data_' . $key . '.php');
 		else
 		{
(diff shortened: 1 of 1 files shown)


live627:
Should pull requests be tagged for a specific release? Helps set expectations and keeps development in check.

PR #5091, opened 2018-10-26
Stats Get the right top topic starter

albertlast:
In 2.1 i remove the order from 441 query for performance reason,
but forgot to adjust the limit.

This should now help to get the right top 10 users.

@live627 maybe you got another approache.

### DIFF
--- a/Sources/Stats.php
+++ b/Sources/Stats.php
@@ -441,7 +441,7 @@ function DisplayStats()
 		SELECT id_member, real_name
 		FROM {db_prefix}members
 		WHERE id_member IN ({array_int:member_list})
-		LIMIT 10',
+		LIMIT 20',
 		array(
 			'member_list' => array_keys($members),
 		)
@@ -464,6 +464,10 @@ function DisplayStats()
 	}
 	ksort($context['stats_blocks']['starters']);
 	$smcFunc['db_free_result']($members_result);
+
+	// remove alle not top 10
+	for($i = 10; $i < 20; $i++)
+		unset($context['stats_blocks']['starters'][$i]);
 
 	foreach ($context['stats_blocks']['starters'] as $i => $topic)
 	{
--- a/Sources/Stats.php
+++ b/Sources/Stats.php
@@ -450,6 +450,10 @@ function DisplayStats()
 	while ($row_members = $smcFunc['db_fetch_assoc']($members_result))
 	{
 		$i = array_search($row_members['id_member'], array_keys($members));
+		// skip all not top 10
+		if ($i >= 10)
+			continue;
+
 		$context['stats_blocks']['starters'][$i] = array(
 			'name' => $row_members['real_name'],
 			'id' => $row_members['id_member'],
@@ -463,10 +467,6 @@ function DisplayStats()
 	}
 	ksort($context['stats_blocks']['starters']);
 	$smcFunc['db_free_result']($members_result);
-
-	// remove alle not top 10
-	for($i = 10; $i < count($members); $i++)
-		unset($context['stats_blocks']['starters'][$i]);
 
 	foreach ($context['stats_blocks']['starters'] as $i => $topic)
 	{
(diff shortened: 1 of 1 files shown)


Sources/Stats.php
```diff
@@ -451,6 +450,10 @@ function DisplayStats()
 	while ($row_members = $smcFunc['db_fetch_assoc']($members_result))
 	{
 		$i = array_search($row_members['id_member'], array_keys($members));
+		// skip all not top 10
+		if ($i >= 10)
+			continue;
```
Sesquipedalian, line 455: You might as well use `break` instead of `continue` here.

albertlast: Since the data is not sorted,
Reason why ksort is called later on,
Is continue the right call.

Sesquipedalian:
(This comment isn't about the changes in this PR itself, but it is about related code.)

@albertlast, I noticed that the query right before the one you are working on here is noted in the comments for being slow, because it needs to do lots of GROUP BY and ORDER BY sorting. That got me wondering if replacing that chunk of code with the following might be faster.
```
	if (($members = cache_get_data('stats_top_starters', 360)) == null)
	{
			SELECT id_member_started
			FROM {db_prefix}topics' . (!empty($modSettings['recycle_enable']) && $modSettings['recycle_board'] > 0 ? '
			WHERE id_board != {int:recycle_board}' : ''),
			array(
				'recycle_board' => $modSettings['recycle_board'],
			)
		);
		$rows = $smcFunc['db_fetch_all']($request))
		$members = array_count_values($rows);
		arsort($members);
		$members = array_slice($members, 0, 10, true);

		cache_put_data('stats_top_starters', $members, 360);
	}
```
As you can see, it uses that nice `$smcFunc['db_fetch_all']` function you so kindly added for us recently, and I suspect that using `array_count_values()`, `arsort()`, and `array_slice()` to get the ids of the top 10 members might be faster than asking the database to use GROUP BY, ORDER BY, and COUNT. If I'm right, this might also make your life a little easier in the code you are working on here, because `$members` will already have been sorted and cut down to ten values.

If you get a moment, would you mind testing this out? If it doesn't help performance much, don't worry about it, but if it does feel free to add that change to this PR.

EDIT: added `true` parameter to the end of `array_slice()`.

live627:
@Sesquipedalian has been looking at my code arsenal 😛

albertlast:
well i don't know if this is a joke or you mean this honest...
first the code is not working,
then i try to simulate the "idea"

```
<?php
$db_server = 'localhost';
$db_name = 'smf';
$db_user = 'postgres';
$db_passwd = '****';
$loop  = 100;


$start = microtime(true);
for($i = 0; $i < $loop; $i++)
{
    // old
    $sql = '
    SELECT id_member_started, COUNT(*) AS hits
    FROM smf_topics
    GROUP BY id_member_started
    ORDER BY hits DESC
    LIMIT 20';
    $members = array();
    pg_free_result($request);
    ksort($rows);
}
	
$start = microtime(true);

for($i = 0; $i < $loop; $i++)
    // new
    {
        $sql =  '
            SELECT id_member_started
            FROM smf_topics';
        //print_r($rows);
        pg_free_result($request);
        $members2 = count($rows);
        arsort($rows);
        $members = array_slice($rows, 0, 10, true);
    }

var_dump($time_old);
var_dump($time_new);

/**
 *
 */
{
	// Return the right row.
}
```
D:\Programme\xampp2\htdocs\test12.php:51:float 1.4883460998535

D:\Programme\xampp2\htdocs\test12.php:52:float 26.885273933411

I hope we all agree that 1.4 is lower as 26.8

Maybe you ask yourself how big smf_topics is,
in my test env i try to simulate a mid size forum -> 77234 rows.

Sesquipedalian:
Thanks for testing that for me, @albertlast. The reason the code didn't work in your test is that in your test you used `count()` where you were supposed to use `array_count_values()`. But that doesn't matter, since the results are clearly much slower than the existing method once there are lots of topics. Feel free to ignore the suggestion. :)

albertlast:
and i used count because array_count_values didn't work.

please had in mind that fetch_all shouldn't be use so long this pr https://github.com/SimpleMachines/SMF2.1/pull/5048 is not merge.
since the output format between mysql and pg is different and got fixed in this pr.

Sesquipedalian (approved):

PR #5093, opened 2018-10-26
4-byte Unicode characters for user names

### LINKED ISSUE #5088: 4-byte Unicode characters for user names
#### Description
Entering 4-byte Unicode characters for a user name during registration or manually adding a user causes a database error 

### Steps to reproduce
1. Navigate to the register page as a new user
2. Enter a user name that contains 4-byte characters from ie. HTML decimal entities
3. example: 🅻🅴🅼🅾🅽
4. db error: 
Incorrect string value: '\xF0\x9F\x85\xBB\xF0\x9F...' for column 'member_name' at row 1
C:\wamp64\www\smf21site4\Sources\Subs-Members.php
Line: 709

- Database Type: MySQL / MariaDB
- Database Version: MySQL 8.0.12 / MariaDB 10.3.10
- PHP Versions: PHP 5.6 - PHP 7.2.11

I have a fix for this which restricts the name entry to 3-bytes that I can enter as a PR but I would like to know the intended behavior prior to doing that.
If 4-byte Unicode characters are intended then this will require a utf8mb4 charset for MySQL or MariaDB.

Underdog-01:
- if $smcFunc['db_mb4'] is false then database appears to be 3-byte therefore mb4 user name entered
will reflect a warning message & not allow the name
- if utf character set is not being used site wide ($context['character_set']) then input is restricted to 3-byte and same warning message displayed
- if warning message is put in $reg_errors[] then it may be logged so user input should be filtered to change illegal chars to "????"

EDIT by live627: Fixes #5088

### DIFF
--- a/Sources/Subs-Members.php
+++ b/Sources/Subs-Members.php
@@ -464,10 +464,7 @@ function registerMember(&$regOptions, $return_errors = false)
 	}
 
 	// Spaces and other odd characters are evil...
-	$regOptions['username'] = trim(preg_replace('~[\t\n\r \x0B\0' . ($context['utf8'] ? '\x{A0}\x{AD}\x{2000}-\x{200F}\x{201F}\x{202F}\x{3000}\x{FEFF}' : '\x00-\x08\x0B\x0C\x0E-\x19\xA0') . ']+~' . ($context['utf8'] ? 'u' : ''), ' ', $regOptions['username']));
-
-	// 4-byte Unicode characters are not permitted within user names where db_mb4 is false
-	$regOptions['username'] = !$smcFunc['db_mb4'] && max(array_map('ord', str_split($regOptions['username']))) >= 240 ? preg_replace_callback('/[\x{10000}-\x{10FFFF}]/u', function($name) use($smcFunc){return $smcFunc['htmlspecialchars']($name[0]);}, $regOptions['username']) : $regOptions['username'];
+	$regOptions['username'] = $smcFunc['htmlspecialchars'](trim(preg_replace('~[\t\n\r \x0B\0' . ($context['utf8'] ? '\x{A0}\x{AD}\x{2000}-\x{200F}\x{201F}\x{202F}\x{3000}\x{FEFF}' : '\x00-\x08\x0B\x0C\x0E-\x19\xA0') . ']+~' . ($context['utf8'] ? 'u' : ''), ' ', $regOptions['username'])));
 
 	// @todo Separate the sprintf?
 	if (empty($regOptions['email']) || !filter_var($regOptions['email'], FILTER_VALIDATE_EMAIL) || strlen($regOptions['email']) > 255)
--- a/Sources/Subs-Members.php
+++ b/Sources/Subs-Members.php
@@ -464,7 +464,10 @@ function registerMember(&$regOptions, $return_errors = false)
 	}
 
 	// Spaces and other odd characters are evil...
-	$regOptions['username'] = $smcFunc['htmlspecialchars'](trim(preg_replace('~[\t\n\r \x0B\0' . ($context['utf8'] ? '\x{A0}\x{AD}\x{2000}-\x{200F}\x{201F}\x{202F}\x{3000}\x{FEFF}' : '\x00-\x08\x0B\x0C\x0E-\x19\xA0') . ']+~' . ($context['utf8'] ? 'u' : ''), ' ', $regOptions['username'])));
+	$regOptions['username'] = trim(preg_replace('~[\t\n\r \x0B\0' . ($context['utf8'] ? '\x{A0}\x{AD}\x{2000}-\x{200F}\x{201F}\x{202F}\x{3000}\x{FEFF}' : '\x00-\x08\x0B\x0C\x0E-\x19\xA0') . ']+~' . ($context['utf8'] ? 'u' : ''), ' ', $regOptions['username']));
+
+	// Convert character encoding for non-utf8mb4 database
+	$regOptions['username'] = $smcFunc['htmlspecialchars']($regOptions['username']);
 
 	// @todo Separate the sprintf?
 	if (empty($regOptions['email']) || !filter_var($regOptions['email'], FILTER_VALIDATE_EMAIL) || strlen($regOptions['email']) > 255)
--- a/Sources/Subs-Members.php
+++ b/Sources/Subs-Members.php
@@ -473,7 +473,7 @@ function registerMember(&$regOptions, $return_errors = false)
 	elseif (max(array_map('ord', str_split($regOptions['username']))) >= 240 && stripos($context['character_set'], 'utf') === false)
 	{
 		$reg_errors[] = array('lang', 'name_invalid_character');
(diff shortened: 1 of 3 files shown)


albertlast:
well ther is already a function how do the stuff and more user friendly: $smcFunc['htmlspecialchars']
https://github.com/SimpleMachines/SMF2.1/blob/310ab996658cc3a13c2c055f2601de4d731f59cc/Sources/Load.php#L135-L138

Like i mention in the issue, we already got a function how do the stuff,
this is this function.

Underdog-01:
htmlspecialchars used in that function you referenced only encodes strings/chars to named entity equivalents.
The Unicode characters I gave in my example (from issues) do not have named entity equivalents and only exist as decimal entities.

albertlast:
did you try it?

Underdog-01:
Without looking I assume $fix_utf8mb4() takes care of the encoding.
This will now trip 2 SMF custom functions that use php standard functions to do what I did in one php function.  If you see my point.

albertlast:
you solution keeps to complex...

```
$regOptions['username'] = $smcFunc['htmlspecialchars']($regOptions['username']);
```
more should be not needed,
all checks and stuff around are not needed to solve this issue.

Underdog-01:
No, doing it that way will convert every character to an entity even when it is not necessary and make the user name column a complete mess of entities. I tested this and what you propose does just that.

Even my original concept will change every character in a string that may contain some characters that do not need conversion.

Since names are limited to a few characters in length (display-wise), I believe it is prudent to only change chars that need changing.

Plus chars should only be converted when it is a utf character set otherwise it will change it to ??? which is not what the user enters as a name therefore must be checked. 
This includes utf-16, utf-32, etc. & not just utf-8 which is why I think it is necessary to pass the character_set parameter to the smcFunc entity function when a utf char set is being used.

I post this using the smcFunc function although Imo mb_convert_encoding is simpler by not jumping (function) hoops to perform the necessary conversion.

albertlast:
first things first = smf 2.1 is utf only
no iso no utf-16 or 32.
The function do the right thing at the right time,
look where it all it used.

Underdog-01:
Always UTF-8 ?

Why do I see ISO-8859-1 as the default charset from Load.php in that quoted code up there?

albertlast:
Because you didn't read the code...
ther stand $utf8.

Underdog-01:
Well, it's a one liner now.

albertlast:
i keep asking my self,
if the call of the function wouldn't be enough...
the function switch ther behavior by therself when mb4 ist enabled/supported.

Underdog-01:
Yes but if mb4 is disabled then your suggestion converts all characters.
All characters may not need to be converted.
Any 3-byte characters can remain as is without being converted.

albertlast:
Since i don't trust you,
i build a test case:
```
<?php
$text = 'i❤me😎🅻🅴🅼🅾🅽';
$utf8 = true;
$smcFunc['db_mb4'] = false;
$modSettings['disableEntityCheck'] = false;

function entity_fix__callback($matches)
{
	if (!isset($matches[2]))
		return '';

	$num = $matches[2][0] === 'x' ? hexdec(substr($matches[2], 1)) : (int) $matches[2];

	// we don't allow control characters, characters out of range, byte markers, etc
	if ($num < 0x20 || $num > 0x10FFFF || ($num >= 0xD800 && $num <= 0xDFFF) || $num == 0x202D || $num == 0x202E)
		return '';
	else
		return '&#' . $num . ';';
}

$ent_check = empty($modSettings['disableEntityCheck']) ? function($string)
{
    $string = preg_replace_callback('~(&#(\d{1,7}|x[0-9a-fA-F]{1,6});)~', 'entity_fix__callback', $string);
    return $string;
} : function($string)
{
    return $string;
};
$fix_utf8mb4 = function($string) use ($utf8, $smcFunc)
	{
		if (!$utf8 || $smcFunc['db_mb4'])
			return $string;

		$i = 0;
		$len = strlen($string);
		$new_string = '';
		while ($i < $len)
		{
			$ord = ord($string[$i]);
			if ($ord < 128)
			{
				$new_string .= $string[$i];
				$i++;
			}
			elseif ($ord < 224)
			{
				$new_string .= $string[$i] . $string[$i + 1];
				$i += 2;
			}
			elseif ($ord < 240)
			{
				$new_string .= $string[$i] . $string[$i + 1] . $string[$i + 2];
				$i += 3;
			}
			elseif ($ord < 248)
			{
				// Magic happens.
				$val = (ord($string[$i]) & 0x07) << 18;
				$val += (ord($string[$i + 1]) & 0x3F) << 12;
				$val += (ord($string[$i + 2]) & 0x3F) << 6;
				$val += (ord($string[$i + 3]) & 0x3F);
				$new_string .= '&#' . $val . ';';
				$i += 4;
			}
		}
		return $new_string;
    };
    
$smcFunc['htmlspecialchars'] = function($string, $quote_style = ENT_COMPAT, $charset = 'ISO-8859-1') use ($ent_check, $utf8, $fix_utf8mb4) 
        { 
            return $fix_utf8mb4($ent_check(htmlspecialchars($string, $quote_style, $utf8 ? 'UTF-8' : $charset))); 
       };

print_r($text);
echo '<br>';

print_r($smcFunc['htmlspecialchars']($text));
```
with $smcFunc['db_mb4'] = false;
``
i❤me😎🅻🅴🅼🅾🅽<br>i❤me&#128526;&#127355;&#127348;&#127356;&#127358;&#127357;
``
with $smcFunc['db_mb4'] = true;
``
i❤me😎🅻🅴🅼🅾🅽<br>i❤me😎🅻🅴🅼🅾🅽
``

Result -> you only need to call the function and this do the rest.

Underdog-01:
@albertlast: 
I have to apologize as I was in error.

Honestly, I was testing what you initially informed me about that function's behavior prior to responding here & was seeing different results (all special chars). This is why I was pushing for doing the variable filtering. 

However, today I started a test anew with a ut8_general_ci db and used strictly the smcFunc whereas everything worked as it should. 

I changed it by just adding it to the existing line of code as I figure that to be more efficient.

Underdog-01:
I did notice something else regarding installing SMF with a utf8-mb4 db though.
Perhaps I should file another report.

albertlast:
In my eyes you do to many things in this one line,
the code hart to read --> should be two lines.


Since it look like that you primary issue get fixed by the line of change,
it would you harm when you real try to extend the scope of this pr with last commit.

timeformat is one of priciest function which exists in smf,
so first we need to see where problem how the change effect this (two screen shot would be nice)
and than we to look if other solutions exists.

I hope you realize while reading that make a own pr worth.

Underdog-01:
Changed as requested.

I didn't mean to merge the two PR's... I use desktop Git & it did so automatically because this PR is still open.

albertlast (approved):

Underdog-01:
Are we also applying this to the current build?

live627:
Sorry, what do you mean? You asking which release this goes into?

Underdog-01:
This PR is still open yet it seems to have been applied to rc2.
How about closing it and applying it to the current beta build?

Sesquipedalian:
It has been marked for evaluation and (assuming it passes evaluation) inclusion in RC2. We haven't yet released RC1, so this will wait in the queue until after that.

Underdog-01:
O.K. thanks for the explanation.

PR #5094, opened 2018-10-26
The top topic starters list didn't show everyone correctly

live627:
- Make the code just a little bit cleaner. This codebase could always use some cleanup.
- Optimize a query to remove the filesort and to not use a temporary table. Makes little difference on a SSD, but there you have it.

Fixes issue discussed in https://www.simplemachines.org/community/index.php?topic=562849.msg3993413#msg3993413

### DIFF
--- a/Sources/Stats.php
+++ b/Sources/Stats.php
@@ -415,20 +415,20 @@ function DisplayStats()
 	if (($members = cache_get_data('stats_top_starters', 360)) == null)
 	{
 		$request = $smcFunc['db_query']('', '
-			SELECT id_member_started, COUNT(*) AS hits
+			SELECT id_member_started, COUNT(*)
 			FROM {db_prefix}topics' . (!empty($modSettings['recycle_enable']) && $modSettings['recycle_board'] > 0 ? '
 			WHERE id_board != {int:recycle_board}' : '') . '
-			GROUP BY id_member_started
-			ORDER BY hits DESC
-			LIMIT 20',
+			GROUP BY id_member_started',
 			array(
 				'recycle_board' => $modSettings['recycle_board'],
 			)
 		);
 		$members = array();
-		while ($row = $smcFunc['db_fetch_assoc']($request))
-			$members[$row['id_member_started']] = $row['hits'];
+		while ($row = $smcFunc['db_fetch_row']($request))
+			$members[$row[0]] = $row[1];
 		$smcFunc['db_free_result']($request);
+		arsort($members);
+		$members = array_slice($members, 0, 10, true);
 
 		cache_put_data('stats_top_starters', $members, 360);
 	}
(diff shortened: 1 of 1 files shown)


albertlast:
Did you really test you solution?
This solution is slower,
because php needs to do more stuff.

```
<?php
$db_server = 'localhost';
$db_name = 'smf';
$db_user = 'postgres';
$db_passwd = 'asdf';
$loop  = 100;


$start = microtime(true);
for($i = 0; $i < $loop; $i++)
{
    // old
    $sql = '
    SELECT id_member_started, COUNT(*) AS hits
    FROM smf_topics
    GROUP BY id_member_started
    ORDER BY hits DESC
    LIMIT 20';
    $members = array();
    pg_free_result($request);
    ksort($rows);
}
	
$start = microtime(true);

for($i = 0; $i < $loop; $i++)
    // new
    {
        $sql =  '
        SELECT id_member_started, COUNT(*)
        FROM smf_topics
        GROUP BY id_member_started';
        //print_r($rows);
        $members = array();
		while ($row = @pg_fetch_row($request))
			$members[$row[0]] = $row[1];
		arsort($members);
		$members = array_slice($members, 0, 10, true);
        pg_free_result($request);;
    }

var_dump($time_old);
var_dump($time_new);

/**
 *
 */
{
	// Return the right row.
}
```

D:\Programme\xampp2\htdocs\test13.php:53:float 1.4927799701691

D:\Programme\xampp2\htdocs\test13.php:54:float 1.8526408672333

--> you loose
stop trying to make stuff slower please...

Why you both think php can be faster than a database?

Maybe some addiotion infos for this query the explain look like so:
"Limit  (cost=2198.66..2198.71 rows=20 width=12) (actual time=17.334..17.336 rows=20 loops=1)"
"  ->  Sort  (cost=2198.66..2203.66 rows=2001 width=12) (actual time=17.334..17.335 rows=20 loops=1)"
"        Sort Key: (count(*)) DESC"
"        Sort Method: top-N heapsort  Memory: 26kB"
"        ->  HashAggregate  (cost=2125.41..2145.42 rows=2001 width=12) (actual time=16.938..17.162 rows=2001 loops=1)"
"              Group Key: id_member_started"
"              ->  Seq Scan on smf_topics  (cost=0.00..1739.27 rows=77227 width=4) (actual time=0.011..3.799 rows=77234 loops=1)"
"Planning time: 0.131 ms"
"Execution time: 17.498 ms"

Important for you is that the "temp table" is not written on the hdd,
it's written to the memory like you did in php,
but since your solution transfer more data AND is not optimize for doing this stuff,
your are not able to be fast with this kind of solution.

albertlast:
This pr and pr: https://github.com/SimpleMachines/SMF2.1/pull/5091
fix both the same issue.

albertlast:
I asked my self why you reopen this,
Like I shown, your logic is wrong from the start.

live627:
I accidentally deleted the branch.

albertlast:
I already show you that your pr doesn't tweak anything(because the problem doesn't exists).
So to close this was the right way.

live627:
Oink! I'm a pig!

https://www.simplemachines.org/community/index.php?topic=562849.msg3993583#msg3993583

This is the change there. Note that people have said that this solves their issue.

I'm done arguing with you over this one.

albertlast (changes requested):
This pr is not a improvment.

PR #5095, opened 2018-10-27
Disable Minimize of file as default fixes #5067

### LINKED ISSUE #5067: "Minimize CSS and JavaScript files" add very high page creation time and loading time
#### Description
When the option is enabled page creation get doubled or more.

Question should be this realy be default enabled?

Should be the help text warn about the issue of higher page creation time?
### Steps to reproduce
1. Admin
2. Gereral setting

- Version/Git revision:
- Database Type:
- Database Version:
- PHP Version:

For $reason i got actualy a unstable/small internet connection
with minimized the page loading (measured by ff) 3.84 sec
without minimized page loading takes 1.92 sec

albertlast:
For me it's very important,
when a user change to smf2.1 or compare his 2.1 setup with his 2.0 setup,
that the forum feels fast.

And in the actual state of development this feature do more the oposite,
on my fast dev i notice some more time and on freehost setup the calling time (network view of browser dev view) adds mutch more time when enabled.

So to prrovide the admin/user the best result at this time,
i woud like to disable this feature by default and this do this pr.

issue: https://github.com/SimpleMachines/SMF2.1/issues/5067


Underdog-01:
I agree with it being disabled by default.

When it has to initially compile it's certainly slower.
For backward compatibility it should be disabled by default.
Mods have to add to the minimize hooks, do they not?

MissAllSunday:
Is up to the mod authors to add their code to the minimize pool, most of the time they don't add that much style code and if they do, it will be more convenient for them to ship their code separately, specially if they overwrite default styles.

As for the process, I still cannot duplicate the issue, if the minimize files are being generate on each request then we need to know why it happens and fix it up rather than disabling it and kick aside the real issue.

albertlast:
When you enable this function on all setup where i look,
notice that the calling time of the browser get up,
even when i refresh the page multiple times.

This function must be on a simple refresh provide the same or faster call time,
measure in browser dev view.

MissAllSunday:
Yeah but is it the result of the request itself? or is the result of loading the actual minimized file?

albertlast:
result of the actual minimized file implemention.
reason why i would like to disable it by default for the time being.

MissAllSunday:
> result of the actual minimized file implemention.


That doesn't really help, sorry. I need to know if the actual file being fetched by the browser is the one taking to long to load, post a screenshot of your dev tools showing all the files in a request and their loading times

something like this:

image


I really need more details about this,. Which browser did you use? do you have browser cache disabled? does it happens with other browsers too? Does your minimized file gets created correctly?  which files causes the lag? the css ones or the js ones?

Disabling it won't fix the issue.

albertlast:
Go to your test env -> enabled the minimized do some refresh and look at the browser tools
than disable it and look again you notice the call of the page get faster with disabled.

MissAllSunday:
> Go to your test env -> enabled the minimized do some refresh and look at the browser tools
than disable it and look again you notice the call of the page get faster with disabled.

No.

I need you to provide evidence of the issue you created #5067  if no evidence can be provided then theres nothing to fix.

albertlast:
i added this information in the issue.

tinoest:
I have it slow on the first load, all subsequent loads are fast as the files have been minimised. Is this not the case with you @albertlast?

albertlast:
fast is only when it faster than non minimized version.

tinoest:
Mine is faster than non minimised after the first load.

albertlast:
Can you please provide pic like i did?
And this information would be important in the issue not in this pr.

The measurement i recommand to disable cron function of the time where you do this.

Issue #5096, opened 2018-10-28
Alter table generates wrong SQL for type boolean


tinoest:
#### Description
See steps to reproduce..

### Steps to reproduce
1.  Create a database table with int as default with db_create_table 
2. Then try to change it to type boolean using db_change_column. It will add the size attribute incorrectly 

### Environment (complete as necessary)
- Version/Git revision: Any
- Database Type: MariaDB
- Database Version: 10
- PHP Version: 7.0

### Additional information/references

Created SQL is 

ALTER TABLE mytable CHANGE COLUMN `column_int` `column_int` boolean(1) NOT NULL default '1' 

Should be 

ALTER TABLE mytable CHANGE COLUMN `column_int` `column_int` boolean NOT NULL default '1'

albertlast:
mysql didn't got boolean right?
Right would be than:
ALTER TABLE mytable CHANGE COLUMN ``column_int`` ``column_int`` tinyint NOT NULL default '1'

tinoest:
> mysql didn't got boolean right?
> Right would be than:
> ALTER TABLE mytable CHANGE COLUMN `column_int` `column_int` tinyint NOT NULL default '1'

Which will mean it doesn't work correctly in pgsql. type boolean is correct for both.

albertlast:
Test please the pr

albertlast:
Please had in mind,
because of the lazy implementation of #5916 mod author etc. can't check if a table got already a boolean field or not.
So please don't forget this.

PR #5097, opened 2018-10-28
Add boolean as datatype for tables #5096

albertlast:
i woud like to use boolean on pg side,
but it would create many confusion so i defined the mysql way for pg as default....

issue: #5096

### DIFF
--- a/Sources/DbPackages-mysql.php
+++ b/Sources/DbPackages-mysql.php
@@ -561,6 +561,7 @@ function smf_db_calculate_type($type_name, $type_size = null, $reverse = false)
 	{
 		$types = array(
 			'inet' => 'varbinary',
+			'boolean' => 'tinyint',
 		);
 	}
 	else
--- a/Sources/DbPackages-postgresql.php
+++ b/Sources/DbPackages-postgresql.php
@@ -735,7 +735,7 @@ function smf_db_calculate_type($type_name, $type_size = null, $reverse = false)
 			'time' => 'time without time zone',
 			'datetime' => 'timestamp without time zone',
 			'timestamp' => 'timestamp without time zone',
-			'boolean' => 'smallint',
+			'boolean' => 'boolean',
 		);
 	}
 	else
--- a/Sources/DbPackages-postgresql.php
+++ b/Sources/DbPackages-postgresql.php
@@ -735,6 +735,7 @@ function smf_db_calculate_type($type_name, $type_size = null, $reverse = false)
 			'time' => 'time without time zone',
 			'datetime' => 'timestamp without time zone',
 			'timestamp' => 'timestamp without time zone',
+			'boolean' => 'smallint',
 		);
 	}
 	else


tinoest (commented):
This doesn’t work with UPDATE SET column = NOT column in PGSQL says type must be Boolean

albertlast:
you cant use not.
in pg you would write "update foobar set col = false"
but you can't doit because mysql wouldn't understand this.
so you had to use 0.

tinoest:
When you want to toggle a value you can use NOT in both MySQL and PGSQL, this is accepted for both. Please leave the type as boolean for pgsql to allow this.

albertlast:
Not is not allowed: https://www.postgresql.org/docs/11/static/datatype-boolean.html

tinoest:
> Not is not allowed: https://www.postgresql.org/docs/11/static/datatype-boolean.html

It's not a datatype, its a logical operator.

https://www.postgresql.org/docs/9.6/static/functions-logical.html

As I said UPDATE tableName SET columnName = NOT columnName WHERE id = 1 

Will toggle the value in columName as shown in the table in that link. 


a | NOT a
TRUE | FALSE
FALSE | TRUE
NULL | NULL

albertlast:
i'm sure that you will got no fun out of this,
because the return type are different between mysql and pg which will lead your in issue on php side.

generic style woulde be to use case.

tinoest:
> i'm sure that you will got no fun out of this,
> because the return type are different between mysql and pg which will lead your in issue on php side.
> 
> generic style woulde be to use case.

https://dev.mysql.com/doc/refman/8.0/en/logical-operators.html#operator_not

They are the same, other than the handling of values which are not 0 or 1. Although boolean should limit it to null, 0 or 1 anyway in both database engines.

albertlast:
they are not the same,
when you insert/update a field you had to use '1' as string for both systems and
the output of mysql is 1 and on pg side it's t.

tinoest:
PHP treats true and 1 the same aslong as you don't use the === operator. 

This all works other than smf_db_replacement__callback also doesn't support boolean / bool. So you have to use raw instead.

albertlast:
you should use string and as value 0 or 1.

tinoest:
That doesn’t work with pgsql ( or I dont think it does )

albertlast:
http://sqlfiddle.com/#!9/3fe4c8/1/0

works with both rdbms

tinoest:
That doesnt have it in a where clause. It also crashes for me when I try that.

albertlast:
https://www.db-fiddle.com/f/rtPa3h5odAv3bj14YghioM/0
here alternative,
end of the day it work.

albertlast:
@Sesquipedalian the related issue is rc2 please fix it

Sesquipedalian:
According to [this](http://www.mysqltutorial.org/mysql-boolean/), MySQL treats BOOLEAN as a synonym for TINYINT(1), so if we tell it to change a column's type to BOOLEAN, it will transparently set the column to TINYINT(1) without any extra work on our part.

I think that means that all we need to do to solve #5096 is to add a line to smf_db_calculate_type() in DbPackages-mysql.php to set $type_size to null if $type_name is boolean. If so, then something like the following is all we should need, right?

```
function smf_db_calculate_type($type_name, $type_size = null, $reverse = false)
{
	// MySQL is actually the generic baseline.

	$type_name = strtolower($type_name);
	// Generic => Specific.
	if (!$reverse)
	{
		$types = array(
			'inet' => 'varbinary',
		);
	}
	else
	{
		$types = array(
			'varbinary' => 'inet',
		);
	}

	// Got it? Change it!
	if (isset($types[$type_name]))
	{
		if ($type_name == 'inet' && !$reverse)
		{
			$type_size = 16;
			$type_name = 'varbinary';
		}
		elseif ($type_name == 'varbinary' && $reverse && $type_size == 16)
		{
			$type_name = 'inet';
			$type_size = null;
		}
		elseif ($type_name == 'varbinary')
			$type_name = 'varbinary';
		else
			$type_name = $types[$type_name];
	}
	elseif ($type_name == 'boolean')
		$type_size = null;

	return array($type_name, $type_size);
}
```

albertlast:
Would be nice if you think on postgres too.
The changes are the minimum for both db.

Side note tinyint didn't got any size value.

albertlast:
Maybe i got you wrong...
but the code work already in the way like you mention.

i define here how to translate the boolean type:
https://github.com/SimpleMachines/SMF2.1/blob/00a1d623a7e2c86c4f9d9b4dfdf96d967618cddc/Sources/DbPackages-mysql.php#L564

here is than true:
https://github.com/SimpleMachines/SMF2.1/blob/00a1d623a7e2c86c4f9d9b4dfdf96d967618cddc/Sources/DbPackages-mysql.php#L575

and then the boolean get translated to tinyint:
https://github.com/SimpleMachines/SMF2.1/blob/00a1d623a7e2c86c4f9d9b4dfdf96d967618cddc/Sources/DbPackages-mysql.php#L590

Sesquipedalian:
> Maybe i got you wrong...
> but the code work already in the way like you mention.
> 
> i define here how to translate the boolean type:
> 
> [SMF2.1/Sources/DbPackages-mysql.php](https://github.com/SimpleMachines/SMF2.1/blob/00a1d623a7e2c86c4f9d9b4dfdf96d967618cddc/Sources/DbPackages-mysql.php#L564)
> Line 564 in [00a1d62](/SimpleMachines/SMF2.1/commit/00a1d623a7e2c86c4f9d9b4dfdf96d967618cddc)
>  			'boolean' => 'tinyint', 
> here is than true:
> 
> [SMF2.1/Sources/DbPackages-mysql.php](https://github.com/SimpleMachines/SMF2.1/blob/00a1d623a7e2c86c4f9d9b4dfdf96d967618cddc/Sources/DbPackages-mysql.php#L575)
> Line 575 in [00a1d62](/SimpleMachines/SMF2.1/commit/00a1d623a7e2c86c4f9d9b4dfdf96d967618cddc)
>  	if (isset($types[$type_name])) 
> and then the boolean get translated to tinyint:
> 
> [SMF2.1/Sources/DbPackages-mysql.php](https://github.com/SimpleMachines/SMF2.1/blob/00a1d623a7e2c86c4f9d9b4dfdf96d967618cddc/Sources/DbPackages-mysql.php#L590)
> Line 590 in [00a1d62](/SimpleMachines/SMF2.1/commit/00a1d623a7e2c86c4f9d9b4dfdf96d967618cddc)
>  			$type_name = $types[$type_name];

What I am saying is that I don't think we need to do any of that for MySQL. According to the link I gave earlier, MySQL is fine with being told to set a columns type to BOOLEAN, and will happily set and retrieve true and false values from that column. *Internally*, MySQL will convert BOOLEAN to TINYINT(1) when saving data, and covert the 0 or 1 back to a boolean value when retrieving the data. 

So, we don't need to do anything in PHP to convert between BOOLEAN and TINYINT(1). MySQL will take care of all that automatically for us. The only thing we need to do is to make sure that in this function, if the column type is boolean, then `$type_size` is set to null so that we don't end up trying to change the automatic TINYINT(1) into TINYINT(16) or something.

albertlast:
you see the behavior here: https://www.db-fiddle.com/f/ocnZpTzsQmxEadG4Mv9vas/1
yes boolean get converted to tinyint (mysql ingore size value on int datatypes).

But the output get complex (mysql return the values sometimes in 0 and 1, sometimes true and false and pg return this in true and false or t and f,
so the coder need to know what he is doing.

albertlast:
What do we want to do with this pr? @Sesquipedalian

Sesquipedalian:
I'm not sure yet. I'll take a fresh look at it soon and see what I think.

albertlast:
@Sesquipedalian how the fresh look looks like?

Sesquipedalian:
Having looked again, I still think that what I said in https://github.com/SimpleMachines/SMF2.1/pull/5097#issuecomment-468130527 is correct. The best way to solve #5096 is to set $type_size to null when $type_name is boolean. This only applies to the MySQL version.

MissAllSunday:
So, what's the status on this? are we going with setting $type_size to null for mysql when $type_name is boolean?  If so, @albertlast can you do the change?

albertlast:
i can do this,
but we all accept the fact that we got no chance on mysql to reverse the datatype from tinyint to boolean, right?

MissAllSunday:
I'm afraid I'm not following you, are you saying there's some kind of backwards compatibility issue?

albertlast:
when you use this function here: https://github.com/SimpleMachines/SMF2.1/blob/00a1d623a7e2c86c4f9d9b4dfdf96d967618cddc/Sources/DbPackages-mysql.php#L567-L572
you can't detect a boolean field.

MissAllSunday:
Do we need to?  is there a specific case where this can happen?

albertlast:
When the mod try to check if the column is boolean and
use for this the api,
than you would need this function.

MissAllSunday:
I'm afraid I still don't know why we cannot apply the change that @Sesquipedalian suggested.

albertlast:
Since his solution is more complex than my solution (got one line of code) i see no need to change it.

MissAllSunday:
Well, I'm taking about what should be the correct way to handle this issue, if the solution turns out to be more than one line of code or too complex we still need to  do it.

albertlast:
From my POV I would implanted this version and
Wait what happen,
Since it's very special use case, 
I see no issue for the core.

MissAllSunday:
What will happen with the case @tinoest mentioned?

PR #5100, opened 2018-10-30
Fix minimized caching

albertlast:
Look like the minimized never worked correctly...
two issue i found while looking at this.
First the path to the minimized files was never push down to the cache --> was always null:
grafik
Secondly the case was never checked when we get the file from the cache if the file is existing in the fs.

maybe related to: #5067

In general i didn't see the issue fixed by this,
because the dependency to the cache is not clear and without cache the minimized stuff make no sense.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -3997,7 +3997,7 @@ function custMinify($data, $type)
 	}
 
 	// And create a one day cache entry.
-	cache_put_data('minimized_' . $settings['theme_id'] . '_' . $type . '_' . $hash, array($toCache, $async, $defer), 86400);
+	cache_put_data('minimized_' . $settings['theme_id'] . '_' . $type . '_' . $hash, array($toCreate, $async, $defer), 86400);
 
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
> Secondly the case was never checked when we get the file from the cache if the file is existing in the fs.
>
>In general i didn't see the issue fixed by this,
>because the dependency to the cache is not clear and without cache the minimized stuff make no sense.

You are right. Using the cache to record whether we've created a minified file creates a dependency, and we can't assume that the file exists now just because there is a record in the cache that we created it. But as you say, the changes is this PR aren't really enough to solve the issue. You are correct that we need to use a `file_exists()` check [here](https://github.com/SimpleMachines/SMF2.1/pull/5100/files#diff-6b95ae55062797979a2f26e651403101R3918), but once we're doing that, it is pointless to use the cache at all.

Don't worry about changing this one further though, @albertlast. Before I checked in here this morning, I wrote an alternative (#5101) that should solve the issue entirely.

PR #5101, opened 2018-10-30
Doesn't try to use cache to track minified files

### LINKED ISSUE #5004: Comment needed in minified files about not editing them
I'm noticing we might begin to have a lot of problems with support with people editing the minified files (in particular the css files). Any thoughts on what we could do to reduce the confusion? I'm thinking if we move them to another folder it may discourage people from editing it, and having only the real files in view is a bonus that is useful to everyone.

Could also make the minified script add a comment in the beginning about this.

### LINKED ISSUE #5067: "Minimize CSS and JavaScript files" add very high page creation time and loading time

Sesquipedalian:
Fixes #5067

Using the cache wasn't working because:

1. It only told us if we'd **created** the minified file before, not whether it **exists** now.
2. It only worked if the cache was enabled, and would unnecessarily re-create the minified file every time if the cache was not enabled.

The only good solution to these problems is to check whether the file actually exists or not. But once we are doing that, there is no reason to use the cache at all. So this gets rid of all the cache-related logic in `custMinify()` (and `deleteAllMinified()`) and simply checks for the actual minified file. Some other lines of code have also been moved around as side effects of this change, but the underlying logic is the same.

Also fixes #5004

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -4027,16 +4027,13 @@ function deleteAllMinified()
 		{
 			foreach (glob(rtrim($theme['dir'], '/') . '/' . ($type == 'css' ? 'css' : 'scripts') . '/minified*.' . $type) as $filename)
 			{
-				// Remove the cache entry
-				if (preg_match('~([a-zA-Z0-9]+)\.' . $type . '$~', $filename, $matches))
-					cache_put_data('minimized_' . $theme['id'] . '_' . $type . '_' . $matches[1], null);
-
 				// Try to delete the file. Add it to our error list if it fails.
 				if (!@unlink($filename))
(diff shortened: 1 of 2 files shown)


albertlast (approved):
Works now
Work without cache
File_exists checks works for css and js

PR #5102, opened 2018-10-30
Some scrutinizer fixes

albertlast:
Well walk throw the list of stuff: https://scrutinizer-ci.com/g/SimpleMachines/SMF2.1/issues/release-2.1
And choose some to fix.


albertlast:
i would like to see that this kind of pr get merge in small pieces,
everytime again (not for rc1)
So wish would be to be target for rc2

jdarwood007 (approved):
Most of these are just documentation, but some are ensuring the strict type checking, which should in theory cause no issues.  But incase we assumed 0 == false which is true and changing to 0 === false would be false, could cause some checks to fail that otherwise worked.  If we don't merge this for RC4, we won't be merging it for final because of any unforeseen issues with these could make the gold release.

The logic seems sounds and I suggest we merge and if any issues are discovered, we can fix those before the final.

PR #5104, opened 2018-10-31
Assume a denial if permissions have not been loaded yet

### LINKED ISSUE #4716: Error Sources/Security.php
Error : array_intersect(): Argument 2 is not an array

Sources/Security.php: 918
if (count(array_intersect($permission, $user_info['permissions'])) != 0)

Solution:

Sources/Security.php: 905
if (empty($user_info) || !isset($user_info['permissions']))

live627:
Fix #4716

### DIFF
--- a/Sources/Security.php
+++ b/Sources/Security.php
@@ -903,7 +903,7 @@ function allowedTo($permission, $boards = null, $any = false)
 		return true;
 
 	// You're never allowed to do something if your data hasn't been loaded yet!
-	if (empty($user_info))
+	if (empty($user_info) || !isset($user_info['permissions']))
 		return false;
 
 	// Administrators are supermen :P.


Sources/Security.php
```diff
@@ -906,6 +906,11 @@ function allowedTo($permission, $boards = null, $any = false)
 	if (empty($user_info))
```
Sesquipedalian: Why not just add ` || !isset($user_info['permissions'])` to this line instead?

Sesquipedalian (changes requested):

Sources/Security.php
live627: Then I'd lose the comment.

Sesquipedalian: 😆 Well, I suppose you could move the comment up under the existing one.

live627:
will do it tomorrow

Sesquipedalian (approved):

PR #5113, opened 2018-11-03
Notfiy

live627:
- [x] Remove scheduled task for sending approval notifications
- [x] Remove old setting to send approval emails from the old system
- [x] Create task for unapproved attachments
- [ ] ~~Create task for unapproved events~~

- The approvals emails seem to go out every two hours since the task was scheduled for that. I haven't verified this because I have never enabled that option thingummy. (Also didn't know of it, but shhh. My sekrit. No tell.)
- The settings area becomes empty. 2.0 had a toggle for each panel. Why is it gone?


live627:
I actually may not need to create a task for event creation because that is handled by the ApprovePost task. The old system didn't really differentiate between events and regular topics anyway.

Also, that task could be renamed. Any suggestions? All I came up with is UnapprovedPost.

Sesquipedalian:
> The settings area becomes empty. 2.0 had a toggle for each panel. Why is it gone?

I suppose because 2.1 moved all the user's notification settings into 
`?action=profile;area=notification`. We could include the relevant settings in the Moderation centre's setting panel as aliases to the same settings. But I'm not sure if that'd make it easier for users to discover the relevant settings or if it would just make things more confusing because there were two places to set the same thing.

> I actually may not need to create a task for event creation because that is handled by the ApprovePost task. The old system didn't really differentiate between events and regular topics anyway.

Is that true even when an event is created without being linked to a post?

> Also, that task could be renamed. Any suggestions? All I came up with is UnapprovedPost.

Maybe "ApprovalNeededPost"? It awkward, but it isn't ambiguous.

live627:
> Is that true even when an event is created without being linked to a post?

I didn't know that was possible.

Sesquipedalian:
It is indeed. Go to `?action=admin;area=managecalendar;sa=settings` and activate the "Allow events not linked to posts" setting.

live627:
> Also, that task could be renamed.

MessageApproval

Sesquipedalian:
> MessageApproval

That would still leave it unclear whether this sends a notification that a message has been approved or that a message needs to be approved.

live627:
Can unlinked events be unapproved? I can't find the option. Might I not be looking hard enough?

Sesquipedalian:
No, it doesn't appear that unlinked events can be unapproved. That probably shouldn't be the case, but dealing with that would fall outside the parameters of this PR. I suppose for now we'll just not worry about that part. If later we add some code to allow unlinked events to be unapproved, we can create a background task to notify moderators about unapproved events then.

Sesquipedalian:
Is this ready for testing now?

live627:
Yes, it is.

Sesquipedalian:
Great. I'll test this as soon as I can.

Sources/tasks/CreateAttachment-Notify.php
```diff
@@ -0,0 +1,160 @@
...
+	 */
+	public function execute()
+	{
+		global $smcFunc, $sourcedir, $scripturl, $modSettings, $language;
+
+		// Validate the attachment does exist and is the right approval state.
+		$request = $smcFunc['db_query']('', '
+			SELECT a.id_attach, m.id_board, m.id_msg, m.id_topic, m.id_member, m.subject
+			FROM {db_prefix}attachments AS a
+				INNER JOIN {db_prefix}messages AS m ON (m.id_msg = a.id_msg)
```
albertlast, line 33: Like always and like the old query,
i keep recommand to drop the word inner here.

live627: Why? Can you explain, please?

albertlast: * First it's optional since 1992
* Secondly you don't use the opposite "outer" when you use left oder right join -> inconsistency
* Third the second point show that left or right give you already the knowledg if this an outer or inner join

live627: This is how queries are done everywhere in SMF.

Perhaps this is something to look into once the next major version of SMF starts being developed.

Sesquipedalian (approved):

Sesquipedalian:
This all seems to work nicely. And with that done, I now need to talk with @jdarwood007 about getting the ball rolling to release SMF 2.1 RC1. 😄

PR #5114, opened 2018-11-04
Put back missing ORDER BY so that ssi_recentTopics is ordered as per SMF2.0.

tinoest:


albertlast:
In my eyes not needed,
Do the sort Manuel after this by using php.

tinoest:
This breaks compatibility with the old way of doing things. It’s also more efficient to do it in the SQL than php. Not ignoring the fact you don’t have the key to sort on anyway.

albertlast:
Sql didn't allow to sort col,
Which you didn't have in select.
So this query doesn't meet the requirement.

tinoest:
> Sql didn't allow to sort col,
> Which you didn't have in select.
> So this query doesn't meet the requirement.

Added it to the SELECT so that it works as expected.

albertlast:
Sry when I confused you,
You're free to add the sort,
When you add this col to the select,
Like now,
You could add sort.
Without sort it's "random" if the value sorted or not.

live627:
Simple test script

```
<?php

require_once('./SSI.php');

ssi_recentTopics();
displayDebug();
```

This fixes the ordering so that latest posts are always first. I forgot to do that way back when I split the queries (do a git blame). Now a filesort is performed.

albertlast (approved):

PR #5117, opened 2018-11-06
Check if $extra is an array

albertlast:
Should fix the issue: https://www.simplemachines.org/community/index.php?topic=563040.0
Even when this not solve the issue,
the check should be ther.


albertlast:
the pr solve the issue.

PR #5125, opened 2018-11-08
Indentation cleanup, general formatting and variable renaming

Gwenwyfar:
Just whitespace changes. Cheers for mass auto-converting.


Gwenwyfar:
Some more general formatting while we're at it.

I believe it doesn't make any difference, but just to note, there were two instances of "FALSE" and "NULL" that got changed to lowercase. Other than that, this is the same as above, only spaces and whitespace changes.

live627:
You   are correct in that they don't make a difference - PHP doesn't care for case in  those keywords.

Sources/ManagePermissions.php
```diff
@@ -163,7 +163,7 @@ function PermissionIndex()
 	$query = $smcFunc['db_query']('', '
 		SELECT id_group, id_parent, group_name, min_posts, online_color, icons
 		FROM {db_prefix}membergroups' . (empty($modSettings['permission_enable_postgroups']) ? '
-		WHERE min_posts = {int:min_posts}' : '') . '
+			WHERE min_posts = {int:min_posts}' : '') . '
 		ORDER BY id_parent = {int:not_inherited} DESC, min_posts, CASE WHEN id_group < {int:newbie_group} THEN id_group ELSE 4 END, group_name',
```
Sesquipedalian: I don't think we normally indent this. We should keep it consistent with the rest.

Sources/ManageServer.php
```diff
@@ -302,7 +302,7 @@ function AlignURLsWithSSLSetting($new_force_ssl = 0)
 	$request = $smcFunc['db_query']('', '
 		SELECT id_theme, variable, value
 		  FROM {db_prefix}themes
-		 WHERE variable in ({string:themeurl}, {string:imagesurl})
+		WHERE variable in ({string:themeurl}, {string:imagesurl})
```
Sesquipedalian, line 305: There's some weird indentation happening in this whole query. It should probably all be fixed, not just this one line.

Sources/ManageServer.php
```diff
@@ -323,7 +323,7 @@ function AlignURLsWithSSLSetting($new_force_ssl = 0)
 			$smcFunc['db_query']('', '
 				UPDATE {db_prefix}themes
 				   SET value = {string:theme_val}
-				 WHERE variable = {string:theme_var}
+				WHERE variable = {string:theme_var}
```
Sesquipedalian, line 327: There's some weird indentation happening in this whole query, too. It should probably all be fixed, not just this one line.

Sources/Subs-Db-mysql.php
```diff
@@ -978,18 +978,18 @@ function smf_db_fetch_all($request)
...
 		return;
 
 	if (empty($mysql_error_data_prep))
-			$mysql_error_data_prep = mysqli_prepare($db_connection,
-				'INSERT INTO ' . $db_prefix . 'log_errors(id_member, log_time, ip, url, message, session, error_type, file, line, backtrace)
-													VALUES(		?,		?,		unhex(?), ?, 		?,		?,			?,		?,	?, ?)'
-			);
+		$mysql_error_data_prep = mysqli_prepare($db_connection,
+			'INSERT INTO ' . $db_prefix . 'log_errors(id_member, log_time, ip, url, message, session, error_type, file, line, backtrace)
+			VALUES(		?,		?,		unhex(?), ?, 		?,		?,			?,		?,	?, ?)'
```
Sesquipedalian: Those long runs of whitespace between values probably aren't doing anything useful.

Sources/Subs-Db-mysql.php
```diff
@@ -978,18 +978,18 @@ function smf_db_fetch_all($request)
...
 	if (empty($db_connection))
 		return;
 
 	if (empty($mysql_error_data_prep))
-			$mysql_error_data_prep = mysqli_prepare($db_connection,
-				'INSERT INTO ' . $db_prefix . 'log_errors(id_member, log_time, ip, url, message, session, error_type, file, line, backtrace)
-													VALUES(		?,		?,		unhex(?), ?, 		?,		?,			?,		?,	?, ?)'
-			);
+		$mysql_error_data_prep = mysqli_prepare($db_connection,
+			'INSERT INTO ' . $db_prefix . 'log_errors(id_member, log_time, ip, url, message, session, error_type, file, line, backtrace)
```
Sesquipedalian: A line break and an extra tab between `log_errors` and the parenthesis would make this much more legible.

Sources/Subs-Db-postgresql.php
```diff
@@ -966,18 +968,18 @@ function smf_db_fetch_all($request)
...
 	if (empty($db_connection))
 		return;
 
 	if (empty($pg_error_data_prep))
-			$pg_error_data_prep = pg_prepare($db_connection, 'smf_log_errors',
-				'INSERT INTO ' . $db_prefix . 'log_errors(id_member, log_time, ip, url, message, session, error_type, file, line, backtrace)
-													VALUES(		$1,		$2,		$3, $4, 	$5,		$6,			$7,		$8,	$9, $10)'
-			);
+		$pg_error_data_prep = pg_prepare($db_connection, 'smf_log_errors',
+			'INSERT INTO ' . $db_prefix . 'log_errors(id_member, log_time, ip, url, message, session, error_type, file, line, backtrace)
```
Sesquipedalian: Again, a line break and an extra tab between `log_errors` and the parenthesis would make this much more legible.

Sources/Subs-Db-postgresql.php
```diff
@@ -966,18 +968,18 @@ function smf_db_fetch_all($request)
...
+			VALUES(		$1,		$2,		$3, $4, 	$5,		$6,			$7,		$8,	$9, $10)'
```
Sesquipedalian: Again, the whitespace between values can be removed.

Sources/Subs.php
```diff
@@ -199,7 +199,7 @@ function updateStats($type, $parameter1 = null, $parameter2 = null)
 				$result = $smcFunc['db_query']('', '
 				SELECT SUM(num_topics + unapproved_topics) AS total_topics
 				FROM {db_prefix}boards' . (!empty($modSettings['recycle_enable']) && $modSettings['recycle_board'] > 0 ? '
-				WHERE id_board != {int:recycle_board}' : ''),
+					WHERE id_board != {int:recycle_board}' : ''),
```
Sesquipedalian, line 202: We don't normally indent WHERE clauses differently than FROM clauses.

Sources/Subs.php
```diff
@@ -259,9 +259,9 @@ function updateStats($type, $parameter1 = null, $parameter2 = null)
 			$smcFunc['db_query']('', '
 			UPDATE {db_prefix}members
 			SET id_post_group = CASE ' . $conditions . '
-					ELSE 0
-				END' . ($parameter1 != null ? '
-			WHERE ' . (is_array($parameter1) ? 'id_member IN ({array_int:members})' : 'id_member = {int:members}') : ''),
+			ELSE 0
+			END' . ($parameter1 != null ? '
+				WHERE ' . (is_array($parameter1) ? 'id_member IN ({array_int:members})' : 'id_member = {int:members}') : ''),
 				array(
```
Sesquipedalian, line 265: We don't normally indent WHERE clauses differently than FROM clauses.

Sources/Subs.php
```diff
@@ -2222,7 +2223,7 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 					{
 						$email_regex = '
 						# Preceded by a non-domain character or start of line
-						(?<=^|[^\p{L}\p{M}\p{N}\-\.])
+							(?<=^|[^\p{L}\p{M}\p{N}\-\.])
```
Sesquipedalian: Please change this one back.

Sesquipedalian (changes requested):

Sesquipedalian:
This PR gives me the warm fuzzies.

Gwenwyfar:
> This PR gives me the warm fuzzies.

Is that good or bad? xD

Also, in the terms of getting everything consistent, I felt like half the time we indent break;, and the other half we don't (same level as case), so I left most alone after a while...

Sources/ManageServer.php
Gwenwyfar: What would you say we should use as default indentation for queries? I don't normally deal with them and I can't quite get something in my head between the inconsistencies and the different format some editors suggest me. Though I see we always indent AND.

frandominguezl:
Apparently, good! https://www.urbandictionary.com/define.php?term=warm%20fuzzies

Sources/ManageServer.php
Sesquipedalian: As a general rule, we align the SQL keywords that start a main logical clause of a query (e.g. SELECT, FROM, WHERE, UPDATE, ORDER BY, etc.) each on their own line and using the same indentation. 

The main exception is the JOIN keywords (JOIN, INNER JOIN, LEFT JOIN, etc.) which are always indented one more tab stop. This is because JOIN ultimately functions to expand the number of tables used in the FROM clause, and so we indent it in a way that shows it as subordinate to FROM.

Keywords that function more like inline logic within a clause (e.g. AND, OR, IN, AS, ASC, DESC, etc.) can either be left inline or moved to a new line that is indented another tab stop in order to show that it is subordinate to the clause it is part of.

---

The reason that this particular case is weird is simply that someone decided to right-align the major keywords (SELECT, FROM, WHERE) instead of left-aligning them. Just take out the extra spaces in front of those words, and this one should be good.

Sesquipedalian:
> Also, in the terms of getting everything consistent, I felt like half the time we indent break;, and the other half we don't (same level as case), so I left most alone after a while...

It's no different than anything else in PHP. If it's inside a function or governed by an `if` or `case` statement or whatever, then indent it another tab stop the same way that you would anything else.

Sources/ManageServer.php
Gwenwyfar: Alright. I'll do a second revision for those and the breaks then.

Gwenwyfar:
Hahaha. Well, then I'll indulge a little more. Last time I did one of those was with Oldiesmann in the template files, so I thought I might drive you a little crazy with it.

What about those? https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/PersonalMessage.php#L2669
There's few, and I'd guess there should be no `;`, but I'm not 100% sure.

Sesquipedalian:
You are correct. The semicolon should not be there.

live627:
Conflicting files
- Sources/Subs.php
- other/upgrade.php

Gwenwyfar:
> Conflicting files
> 
>     * Sources/Subs.php
> 
>     * other/upgrade.php

Conflicts make me a sad panda. I still don't know how to fix them nicely in a PR (and github's interface is not nicely). I'll leave it to you.

other/upgrade_1-1.sql
```diff
@@ -1009,8 +1009,8 @@ if (!$has_attachmentType_column)
 	$ID_MSG = array();
 	while ($row = smf_mysql_fetch_assoc($request))
 	{
-		$clean_name = strtr($row['filename'], '������������������������������������������������������������', 'SZszYAAAAAACEEEEIIIINOOOOOOUUUUYaaaaaaceeeeiiiinoooooouuuuyy');
-		$clean_name = strtr($clean_name, array('�' => 'TH', '�' => 'th', '�' => 'DH', '�' => 'dh', '�' => 'ss', '�' => 'OE', '�' => 'oe', '�' => 'AE', '�' => 'ae', '�' => 'u'));
+		$clean_name = strtr($row['filename'], '������������������������������������������������������������', 'SZszYAAAAAACEEEEIIIINOOOOOOUUUUYaaaaaaceeeeiiiinoooooouuuuyy');
```
Gwenwyfar: Ugh... Phpstorm, what have you done. Was it the file encoding? :/

Edit: Ok, so it got changed to UTF-8 from Windows format. Didn't even know we had anything around that wasn't UTF-8.

Gwenwyfar: Apparently the conversion itself got borked, rather than the encoding itself, so I left it as UTF-8 and just fixed the characters. If we really want it on windows encoding for some reason I can change it back.

Gwenwyfar:
I was going to make one commit for each variable type, but there were many more language ones than I expected, so I'm splitting it in further commits. This is about half the language files done.

$.sceditor.locale may need to have the copy/paste strings updated if they are being used as var names, could someone check this? Though it seems we don't use them, but mods might.

Sesquipedalian (approved):

jdarwood007:
@frandominguez03 
Closed issue, but this PR contained $txt changes.  Any issues with getting those to LE?

frandominguezl:
We should release RC1 language packs once we officially release 2.1 RC1

Gwenwyfar:
@Sesquipedalian did you check with the editor locale strings inside the js file? Or is it all good?

Sesquipedalian:
Do you mean the changes you made to the $editortxt variables? If so, yes, and it all works nicely.

Gwenwyfar:
Alright then :)

Wasn't sure whether some in the js files were the variable name or duplicate strings, the way they got named, heh.

PR #5127, opened 2018-11-09
Upgrade PG refresh sequence values

albertlast:
When a new scheduled_tasks got added by the 2.1 upgrade process,
a duplicate id error was raised up by postgres,
reason is that the 2.0 setup doesn't do the right thing to update the sequence.

To fix this issue overall,
I added a "refresh" block over all sequence which get created in 2.0.


Sesquipedalian:
One question: If we include a fix for the original problem in 2.0.16, such that the values will already be correct for someone upgrading from 2.0.16 to 2.1, will this code still work?

albertlast:
sure the code only set "all" sequence to the max value of each table

Sesquipedalian (approved):

PR #5130, opened 2018-11-09
Language fixes

Gwenwyfar:


Themes/default/languages/EmailTemplates.english.php
```diff
@@ -319,8 +319,7 @@
 $txt['alert_unapproved_post_subject'] = 'New Unapproved Post: {SUBJECT}';
 $txt['alert_unapproved_post_body'] = 'A new post has been made which needs to be approved: \'{SUBJECT}\'
 
-You can approve or reject this post by clicking the link below:
-
+You can approve or reject this post on the link below:
```
illori, line 322: i dont agree with this change. you need the part about clicking on it to make sense

Themes/default/languages/EmailTemplates.english.php
```diff
@@ -334,8 +333,7 @@
 $txt['alert_unapproved_topic_subject'] = 'New Unapproved Topic: {SUBJECT}';
 $txt['alert_unapproved_topic_body'] = 'A new topic has been made which needs to be approved: \'{SUBJECT}\'
 
-You can approve or reject this topic by clicking the link below:
-
+You can approve or reject this topic on the link below:
```
illori, line 336: same here.

Themes/default/languages/EmailTemplates.english.php
```diff
@@ -356,7 +354,7 @@
 
 {REASON}
 
-You can approve or reject this application by clicking the link below:
+You can approve or reject this application on the link below:
```
illori, line 357: and here

Themes/default/languages/EmailTemplates.english.php
Gwenwyfar: The point here is to change the "by" because it could be interpreted as "you will do an action by clicking the link", when the purpose is to say "you can do said action after following the link". Any other suggestions to have that effect?

Or does it add two links? Then it would be correct as is, but it says "link", not "links".

Themes/default/languages/EmailTemplates.english.php
smfbigguy: on should be with

live627: `from`?

illori: "by using the link below", technically you dont click on a phone or tablet etc.

smfbigguy: You can approve or reject this application on the link below:

Should be

You can approve or reject this application with the link below:

Gwenwyfar: Or "at" perhaps?

illori: i think we are making it more complicated trying to change it. i think we should just leave it as is.

Gwenwyfar: I don't see how any of the options are complicated, but neither do I want to spend time discussing about such a small thing. So if you could just make a decision we'll roll with it.

Sesquipedalian: @illori's "using" is the best suggestion.

Sesquipedalian:
I added @illori's "by using" suggestion to this PR and then merged in manually.

PR #5138, opened 2018-11-11
Random compat 2.0.17

MissAllSunday:
Moves random_compat() files to their own folder on Sources/


albertlast:
Maybe you should read the related https://github.com/SimpleMachines/SMF2.1/pull/5077,
before wasting your time on idea which got already decline by the smf dev team.

Sesquipedalian:
> Maybe you should read the related #5077,
before wasting your time on idea which got already decline by the smf dev team.

😕 When was it rejected? I thought you just put it there because that's what you decided to do.

EDIT: Oh, I see. Originally you had the library file separate, and then @jdarwood007 asked if it could be moved into Subs-Compat.php. That was probably the wrong idea on his part. Having it separate like your original idea and like @MissAllSunday is doing in this PR is better for long term maintenance. If I'd thought more about it at the time, I probably would have suggested that too, rather than cramming the whole library into one file like I suggested before.

Sesquipedalian:
@MissAllSunday, to fix the Travis build failures, you'll need to add some appropriate regexes [here](https://github.com/SimpleMachines/BuildTools/blob/master/check-smf-license.php#L48). Then we should be able to merge this.

albertlast:
> 
> 
> > Maybe you should read the related #5077,
> > before wasting your time on idea which got already decline by the smf dev team.
> 
> 😕 When was it rejected? I thought you just put it there because that's what you decided to do.
> 
> EDIT: Oh, I see. Originally you had the library file separate, and then @jdarwood007 asked if it could be moved into Subs-Compat.php. That was probably the wrong idea on his part. Having it separate like your original idea and like @MissAllSunday is doing in this PR is better for long term maintenance. If I'd thought more about it at the time, I probably would have suggested that too, rather than cramming the whole library into one file like I suggested before.

You forced me to do two stupid things,
one throw everything one file and
secondly call the reloadsetting in context of the upgrade process -.-

Sesquipedalian:
Well, what I said was that we needed to include all the library functions, not just one that depends on others that weren't included. And then I was the one who compiled the library functions into a single block of code and gave it to you so that you could paste it into the file. So it's not like you did much work there, and thus you don't have much to complain about if we change our minds after further consideration.

Besides which, I'm not sure why you are acting so grumpy and rude about this anyway. Everyone's code gets changed sooner or later. That's the way this works. If you find yourself getting upset by that, perhaps there are some anger issues that you need to work on?

albertlast:
This had nothing to do with anger,
i mention four times,
that it's a bad idea to place the reloadsetting into the upgrade(specialy with the background we want to release rc1).

Sesquipedalian:
Indeed you did. And then #5124 was merged. So what exactly are you complaining about here?

Sesquipedalian:
@MissAllSunday, please also include the necessary changes in install.php and upgrade.php.

Sesquipedalian (changes requested):
Please also include the necessary changes in install.php and upgrade.php.

other/upgrade.php
```diff
@@ -1010,7 +1010,11 @@ function checkLogin()
 				$upcontext['user']['id'] = 1;
 				$upcontext['user']['name'] = 'Administrator';
 			}
-			$upcontext['user']['pass'] = $smcFunc['random_int'](0, 60000);
+
+            if (!is_callable('random_int'))
```
albertlast, line 1014: The code is not well aligned -> because you use space and the code is using tab here.

other/install.php
```diff
@@ -1553,7 +1553,10 @@ function AdminAccount()
 		}
 		elseif ($_POST['username'] != '')
 		{
-			$incontext['member_salt'] = substr(md5($smcFunc['random_int']()), 0, 4);
+            if (!is_callable('random_int'))
```
albertlast, line 1556: The code is not well aligned -> because you use space and the code is using tab here.

PR #5147, opened 2018-11-17
On DB Replace call check if $keys not empty

albertlast:
while looking over the code of ehportal from @Underdog-01 
i notice call like this:
```
$smcFunc['db_insert']('replace',
	'{db_prefix}sp_parameters',
	array(
		'id_block' => 'int',
		'variable' => 'text',
		'value' => 'text',
	),
	$old_parameters,
	array()
);
```

Is working,
what is not wanted because the array() is empty and we are in replace mode,
but we need to know the keys to be sure that replace is working.

Extend the check and add the special behavior on replace mode.

So two rules exist on replace mode:
* keys needs to be mention
* keys needs to be part of column array (this rule is already ther)

Btw. checked all 82 calls of smf2.1 if they hurt the "new" rule but nothing needs to change.


live627:
I'm really not sure about this as it could upset existing code. Many queries in SMF as well as many mods leave`$keys` empty.  Most queries also leave `$method` blank, so the default of `replace` is used.

I'm going to wait for @Sesquipedalian to weigh in on this. (This might get rejected)

albertlast:
Since this is already essential before,
but was not checked by the db api,
smf is not effected by his own.

live627:
I did find a few insertion queries with empty $key, in some background tasks, Subs-Attachments, and PersonalMessage. Definitely not "many" as I previously thought.

The major concern here is still third party mods.

albertlast:
When you leave $methode blank,
than your are not in replace mode.

reason why call like this:
https://github.com/SimpleMachines/SMF2.1/blob/54e2dbcce94b0832b9cae7a9aa1494b19bea9279/Sources/Subs-Attachments.php#L739-L748
keep working.

albertlast:
@Sesquipedalian is this something for rc2?

Sources/Subs-Db-postgresql.php
```diff
@@ -707,7 +707,7 @@ function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $ret
 	$table = str_replace('{db_prefix}', $db_prefix, $table);
 
 	// Sanity check for replace is key part of the columns array
-	if ($method == 'replace' && count(array_intersect_key($columns, array_flip($keys))) !== count($keys))
+	if ($method == 'replace' && (empty($keys) || count(array_intersect_key($columns, array_flip($keys))) !== count($keys)))
```
MissAllSunday: Looks good but can you separate the empty($keys) into its own check?  just so we can provide a more detailed error message and help whoever uses the replace method to debug their code.

MissAllSunday (changes requested):

Sources/Subs-Db-postgresql.php
albertlast: done

MissAllSunday: cool, thanks!

MissAllSunday (approved):

Sources/Subs-Db-postgresql.php
```diff
@@ -648,9 +648,15 @@ function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $ret
...
 
 	// Sanity check for replace is key part of the columns array
-	if ($method == 'replace' && count(array_intersect_key($columns, array_flip($keys))) !== count($keys))
-		smf_db_error_backtrace('Primary Key field missing in insert call',
-			'Change the method of db insert to insert or add the pk field to the columns array', E_USER_ERROR, __FILE__, __LINE__);
+	if ($method == 'replace')
+	{
+		if (empty($keys))
+			smf_db_error_backtrace('When using the replace mode, the key column is a required entry.',
+				'Change the method of db insert to insert or add the pk field to the key array', E_USER_ERROR, __FILE__, __LINE__);
```
wintstar, line 655: Why are texts of error messages not to be translated?

albertlast: So the same reason like they are in other places in db layer not able to be translated.
- DB layer is called very very early (no information about the user exists)
- To post not english version of this error in the support board doesn't help
- People how are not able to read the english text are not able to solve this kind of issue

PR #5157, opened 2018-11-21
integrate_spam_protection was called a little late

live627:
-  `integrate_spam_protection` probably should be renamed to `integrate_floodcontrol so it matches everything else in the function
- Seems useless to pass `$timeLimit` through to integration


MissAllSunday:
I can't remember if this hook existed in 2.0 probably no but might as well be sure, if not, can you do the change and solve the conflicts to merge this.

PR #5167, opened 2018-11-23
Fix $last_id in mysql fixes #5164

### LINKED ISSUE #5164: The variable $last_id seems to be never defined.
Found by Scrutinizer

https://github.com/SimpleMachines/SMF2.1/blob/b641970a1945f084792ebd9d43c4403a3b12b241/Sources/Subs-Db-mysql.php#L819

https://scrutinizer-ci.com/g/SimpleMachines/SMF2.1/issues/release-2.1/files/Sources/Subs-Db-mysql.php?selectedLabels%5B0%5D=6&orderField=path&order=asc&honorSelectedPaths=0

@albertlast

albertlast:
Issue: #5164

### DIFF
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -816,7 +816,7 @@ function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $ret
 			);
 			$new_id = $smcFunc['db_insert_id']();
 
-			if ($last_id != $new_id) //the inserted value was new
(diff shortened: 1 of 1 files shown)


albertlast:
Housekeeping would be wrong here,
mysql or database is right here since it's a real bug.
Lucky for us is that this function is new and no one is using this.

live627:
sorry, my bad.

/me was high on rusty turkey 😛

Issue #5171, opened 2018-11-25
Image proxy - drops image, reason unknown


sbulen:
#### Description
I just picked 21 random http: images from my forum. With the proxy off, all 21 show. In 2.0, with the proxy ON, 19 show & 2 are dropped. But in 2.1, with the proxy ON, 14 show & 7 are dropped... I see at least 4 different symptoms on the 7 items dropped. ("dropped" means not displayed at all.) I've confirmed this on WAMP & on Linux.

This one is odd, I have no idea why it is not working...

This image is properly cached & displayed in 2.0.15, but not in 2.1.

### Steps to reproduce
Test link:
http://www.ovationgallery.com/Gallery%20Storm/OG-storm-to-nutv-01.jpg

### Environment (complete as necessary)
- Version/Git revision: today's
- Database Type: mysql
- Database Version: 5.7
- PHP Version: 5.6

### Additional information/references

sbulen:
I believe this is also caused by the % being translated to %25 by a call to rawurlencode().   

See #5170 .   

Root cause is a dupe, so I'll close this one.

PR #5174, opened 2018-11-26
Handle spaces that are already encoded

### LINKED ISSUE #5170: Image proxy - drops image due to rawurlencode
#### Description
I just picked 21 random http: images from my forum.  With the proxy off, all 21 show.  In 2.0, with the proxy ON, 19 show & 2 are dropped.  But in 2.1, with the proxy ON, 14 show & 7 are dropped...  I see at least 4 different symptoms on the 7 items dropped.  ("dropped" means not displayed at all.)  I've confirmed this on WAMP & on Linux.  

This one is odd, sockets works & curl works but the image isn't being cached or displayed.  The host-level header is returning a 403, but the image header & curl both work OK.  

This image is properly cached & displayed in 2.0.15, but not in 2.1.  

### Steps to reproduce
Test link:
http://i145.photobucket.com/albums/r227/Netskenmac/Roland%20Mobile%20Cube%20Amp/AmpMeasuringTape2.jpg

- Version/Git revision:  today's
- Database Type: mysql
- Database Version: 5.7
- PHP Version: 5.6

sbulen:
Fixes #5170 .   The image proxy was having issues with URLs that contained %20.  

I considered two solutions, the first was to simply remove the call to iri_to_url() from proxy.php.  That felt safest, and produced the same results during image proxy tests.  

But this solution is more in line with the intent of using iri_to_url() - to cleanse & standardize url usage.

Tested on both WAMP & Linux.  Solves a # of dropped image issues with the image proxy.

### DIFF
--- a/proxy.php
+++ b/proxy.php
@@ -89,9 +89,6 @@ class ProxyServer
 		if (md5($request . $this->secret) != $hash)
 			return false;
 
-		// Ensure any non-ASCII characters in the URL are encoded correctly
-		$request = iri_to_url($request);
-
 		// Attempt to cache the request if it doesn't exist
 		if (!$this->isCached($request))
 			return $this->cacheImage($request);
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -6533,6 +6533,11 @@ function iri_to_url($iri)
 {
 	global $sourcedir;
 
+	// First do a quick check to ensure there are no encoded values already
+	// These get mangled by rawurlencode()...
+	if (preg_match('~%\d{2}~', $iri) === 1)
+		return($iri);
+
 	$host = parse_url((strpos($iri, '://') === false ? 'http://' : '') . ltrim($iri, ':/'), PHP_URL_HOST);
 
 	if (empty($host))
@@ -6546,8 +6551,6 @@ function iri_to_url($iri)
 	$iri = substr_replace($iri, $encoded_host, $pos, strlen($host));
 
 	// Encode any disallowed characters in the rest of the URL
-	// First deal with spaces that may already be encoded...
-	$iri = strtr($iri, array('%20'=>' '));
 	$unescaped = array(
 		'%21'=>'!', '%23'=>'#', '%24'=>'$', '%26'=>'&',
 		'%27'=>"'", '%28'=>'(', '%29'=>')', '%2A'=>'*',
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -6533,11 +6533,6 @@ function iri_to_url($iri)
 {
 	global $sourcedir;
 
-	// First do a quick check to ensure there are no encoded values already
-	// These get mangled by rawurlencode()...
-	if (preg_match('~%\d{2}~', $iri) === 1)
-		return($iri);
-
 	$host = parse_url((strpos($iri, '://') === false ? 'http://' : '') . ltrim($iri, ':/'), PHP_URL_HOST);
 
 	if (empty($host))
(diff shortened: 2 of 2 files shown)


albertlast:
Would it be not better to decode the url in general?

sbulen:
I agree this feels incomplete...  But putting a decode immediately before the encode (the next line of code) isn't a good approach either.  You might as well drop both.  

The root cause of the issue here is that this function assumes the content is not encoded.  But it frequently is encoded.  The vast majority of the time, folks are copying the string from a url or a 'share link' or a 'copy image address' from their browser.  

I'll try a different approach...

albertlast:
Looks for me not right,
you only look if possible one escape encoded char exists.
But it make not clear if the complete url ist encoded.

What would be hapen when you decode a non encoded url?

sbulen:
The problem here is that encoding *any* already-encoded character mangles that character.  Eg, %20 is becoming %2520, and %25 would become %2525, etc.  

There is no definitive way to determine if the url has already been encoded or not.  

This latest approach leaves the string alone if it finds something that looks like %99.  A very conservative approach.

sbulen:
How's this? 

At the same time, it feels a little more purist though a little more dangerous...  It feels like excessive processing of these URLs, which have been user-specified & have probably been copied from somewhere.  If we see more dropped pix caused by this processing, I'd suggest removing the call to iri_to_url() from proxy.php altogether.  

But this does work.  I have confirmed it fixes both #5170 & #5171 on both WAMP & Linux.

sbulen:
The above didn't work on all impacted URLs.  I really don't think we should do the call to iri_to_url().   

More test data:

Photo: 6
http://www.guitarmachine.com/Home%20page%20white%20photo.jpg
[img]http://www.guitarmachine.com/Home%20page%20white%20photo.jpg[/img]
==================================================================================================
Photo: 40
http://www.ovationgallery.com/Gallery%20Storm/OG-storm-to-nutv-01.jpg
[img]http://www.ovationgallery.com/Gallery%20Storm/OG-storm-to-nutv-01.jpg[/img]
==================================================================================================
Photo: 41
http://www.itcelectronics.com/images/products/PHL%2030-10002.jpg
[img]http://www.itcelectronics.com/images/products/PHL%2030-10002.jpg[/img]
==================================================================================================
Photo: 50
http://i145.photobucket.com/albums/r227/Netskenmac/Roland%20Mobile%20Cube%20Amp/AmpMeasuringTape2.jpg
[img]http://i145.photobucket.com/albums/r227/Netskenmac/Roland%20Mobile%20Cube%20Amp/AmpMeasuringTape2.jpg[/img]
==================================================================================================
Photo: 179
http://www.midiguitar.net/group/EV-5%20Info/ev-5-schematics.png
[img]http://www.midiguitar.net/group/EV-5%20Info/ev-5-schematics.png[/img]
==================================================================================================
Photo: 181
http://3.bp.blogspot.com/--4xvZoFVOPM/TYGaGTaOkAI/AAAAAAAAADU/7EV8tjJy6WE/s400/DSCN1435%2B%25282%2529.JPG
[img]http://3.bp.blogspot.com/--4xvZoFVOPM/TYGaGTaOkAI/AAAAAAAAADU/7EV8tjJy6WE/s400/DSCN1435%2B%25282%2529.JPG[/img]
==================================================================================================

proxy.php
```diff
@@ -89,9 +89,6 @@ public function checkRequest()
 		if (md5($request . $this->secret) != $hash)
 			return false;
 
-		// Ensure any non-ASCII characters in the URL are encoded correctly
-		$request = iri_to_url($request);
```
Sesquipedalian, line 93: Don't remove this line. Instead, all you should need to do is change it to:
```
$request = iri_to_url(rawurldecode($request));
```

Sesquipedalian (changes requested):

proxy.php
Sesquipedalian: Disregard my previous comment. The issue was just that iri_to_url() wasn't correctly handling double-encoded `%` characters. #5194 fixes it properly.

PR #5180, opened 2018-11-29
revamp database escaping fixes #5178

### LINKED ISSUE #5178: Malformed JSON for $modSettings['displayFields'] when using Postgres cache
#### Description
Set up a local SMF install with PostgreSQL as the database. I set the cache option to use PostgreSQL as well since that's available. When viewing the topic I posted, I get the following error:
```
Json decode error: Syntax error, malformed JSON
http://localhost/smf/index.php?topic=2.msg3
/srv/http/smf/Sources/Load.php (Line 1566)
```

### Steps to reproduce
1. Set up SMF with PostgreSQL cache
2. View a post made by a registered user

- Database Type: PostgreSQL
- Database Version: 11.1
- PHP Version: 7.2.12

Backtrace:
```
#0: smf_json_decode()
Called from /srv/http/smf/Sources/Load.php on line 1568
#1: loadMemberContext()
Called from /srv/http/smf/Sources/Display.php on line 1564
#2: prepareDisplayContext()
Called from /srv/http/smf/Themes/default/Display.template.php on line 254
#3: template_main()
Called from /srv/http/smf/Sources/Load.php on line 2460
#4: loadSubTemplate()
Called from /srv/http/smf/Sources/Subs.php on line 3164
#5: obExit()
Called from /srv/http/smf/index.php on line 153
```

This is the actual JSON for $modSettings['displayFields']:
```
[{"col_name":"cust_icq","title":"ICQ","type":"text","order":"1","bbc":"0","placement":"1","enclose":"<a class=\\"icq\\" href=\\"\\/\\/www.icq.com\\/people\\/{INPUT}\\" target=\\"_blank\\" title=\\"ICQ - {INPUT}\\"><img src=\\"{DEFAULT_IMAGES_URL}\\/icq.png\\" alt=\\"ICQ - {INPUT}\\"><\\/a>","mlist":"0"},{"col_name":"cust_skype","title":"Skype","type":"text","order":"2","bbc":"0","placement":"1","enclose":"<a href=\\"skype:{INPUT}?call\\"><img src=\\"{DEFAULT_IMAGES_URL}\\/skype.png\\" alt=\\"{INPUT}\\" title=\\"{INPUT}\\" \\/><\\/a> ","mlist":"0"},{"col_name":"cust_yahoo","title":"Yahoo! Messenger","type":"text","order":"3","bbc":"0","placement":"1","enclose":"<a class=\\"yim\\" href=\\"\\/\\/edit.yahoo.com\\/config\\/send_webmesg?.target={INPUT}\\" target=\\"_blank\\" title=\\"Yahoo! Messenger - {INPUT}\\"><img src=\\"{IMAGES_URL}\\/yahoo.png\\" alt=\\"Yahoo! Messenger - {INPUT}\\"><\\/a>","mlist":"0"},{"col_name":"cust_loca","title":"Location","type":"text","order":"4","bbc":"0","placement":"0","enclose":"","mlist":"0"},{"col_name":"cust_gender","title":"Gender","type":"radio","order":"5","bbc":"0","placement":"1","enclose":"<span class=\\" generic_icons gender_{KEY}\\" title=\\"{INPUT}\\"><\\/span>","mlist":"0",,"options":["None","Male","Female"]}]
```

albertlast:
Well the pr got more complex/intresting as expected...

The root cause is that mysql know two escaping symbole,
on the one side it know the sql default one ``''`` (double single quote)
and on the other side it know ``\`` blackslash.

This pr disable the last methode -> only one way exists now

Benefits:
the mysql and pg setup could be the same on displayfields (fix the issue)
the amount of ``\`` you insert is the amount which you get back
the undecided smylie work with ``:-\`` instead of two ``\``
in mysql db_query the "// Remove the string escape for better runtime" needs to do this only for one way (otherwise we would had to add both ways) -> less complexity

issue: #5178


Sesquipedalian (dismissed):

sbulen:
My only concern with this (assuming it's been thoroughly tested) is disallowing backslash escapes in the connect string.  

That is basically enforcing a new standard, & I'm not sure what might break as a result.  

I am fairly certain some working mods will break.

albertlast:
This change didn't affected cases like the connect string,
Because the backslash is used for php and not mysql.
This is another reason why this change is important to provide a better distinction between the cases.

Since the changes are easy to adept,
I see the mod topic no problem.

albertlast:
@Sesquipedalian the related issue is marked as rc2,
but htis pr doesn't got marked.
what is here the plan?

Sources/Subs-Db-mysql.php
```diff
@@ -100,7 +100,7 @@ function smf_db_initiate($db_server, $db_name, $db_user, $db_passwd, $db_prefix,
 	if (empty($db_options['dont_select_db']) && !@mysqli_select_db($connection, $db_name) && empty($db_options['non_fatal']))
 		display_db_error();
 
-	mysqli_query($connection, 'SET SESSION sql_mode = \'ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION\'');
+	mysqli_query($connection, 'SET SESSION sql_mode = \'ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION,NO_BACKSLASH_ESCAPES\'');
```
Sesquipedalian: For the sake of easier reading, could you change this to use double quotes around  the overall string? Then it could look like this:
`mysqli_query($connection, "SET SESSION sql_mode = 'ONLY_FULL_GROUP_BY,STRICT...`

Sources/Subs-Db-mysql.php
```diff
@@ -414,7 +414,7 @@ function smf_db_query($identifier, $db_string, $db_values = array(), $connection
 		$old_pos = 0;
 		$pos = -1;
 		// Remove the string escape for better runtime
-		$db_string_1 = str_replace('\\\'', '', $db_string);
+		$db_string_1 = str_replace('\'\'','',$db_string);
```
Sesquipedalian: For the sake of easier reading, could you change this one to use double quotes around  the overall string, too?

Sesquipedalian (changes requested):

Sources/Subs-Db-mysql.php
albertlast: Both are good idea,
will change it later on.

albertlast: To get a better understanding i split the modes

live627: So why aren't we using `$smcFunc` for this?

albertlast: First, Why we should?
Secondly, it doesn't work.

live627: - it handles escaping things for us
-    no? It accepts a connection object as the third param

albertlast: ther is no escaping, you would use the raw method --> provide nothing
feel free to do a pr with this changes.

The result would be, that the mysql connection get unstabler and slower --> got no problem with this.

Sesquipedalian (approved):

Sesquipedalian:
Hm. I just noticed a few places in some of the older upgrade files where `addslashes()` appears to be used to escape SQL strings. Even without the changes in this PR, that's a [bad idea](http://php.net/manual/en/function.addslashes.php) and should be fixed. With the changes in this PR, those _need_ to be fixed before this can be merged. Can you check these and fix any that need to be changed, @albertlast?

The files that need to be checked for possible problems:
upgrade_1-0.sql (lines 827 and 911)
upgrade_2-0_mysql.sql (lines 463–476, 1711–1721, and 2033–2034)
upgrade_2-0_postgresql.sql (lines 1081 and 1082)

albertlast:
Postgres did already working fine,
no check is needed.
On the other files i will look into.

albertlast:
Should be fixed now.

jdarwood007:
If we are changing the older upgrade files, we need to test upgrading from them.

albertlast:
from my pov the upgrade files for < 2.0 should be deleted here,
because no one is able to test this.

other/upgrade_2-0_mysql.sql
```diff
@@ -1460,20 +1460,27 @@ while ($_GET['m'] < $totalActions)
...
+			SET file = SUBSTRING({string:file}, 1, 255),
+				line = {int:line},
+				message = SUBSTRING({string:message}, 1, 65535)
+			WHERE id_error = {int:id_error}
+			LIMIT 1',
+			array(
+				'file' => $row[file],
+				'line' => $row[line],
+				'message' => $row[message],
+				'id_error' => $row[id_error],
```
Sesquipedalian: These all need single quotes around the array key (i.e. `$row['file']`, etc.).

other/upgrade_2-0_mysql.sql
```diff
@@ -2030,8 +2037,22 @@ while ($_GET['m'] < $totalActions)
...
+				id_topic = {int:topic_id}, 
+				id_msg = {int:msg_id}, 
+				extra =  {string:extra}
+			WHERE id_action = {int:id_action}',
+			array(
+				'board_id' => $board_id,
+				'topic_id' => $topic_id,
+				'msg_id'   => $msg_id,
+				'extra'    => $row[extra],
+				'id_action' => $row[id_action],
```
Sesquipedalian: These need single quotes around the array key (i.e. `$row['extra']`, etc.).

Sesquipedalian (changes requested):
Missing some single quotes.

albertlast:
got fixed

Sources/Subs-Db-mysql.php
```diff
@@ -100,7 +100,13 @@ function smf_db_initiate($db_server, $db_name, $db_user, $db_passwd, $db_prefix,
...
 		display_db_error();
 
-	mysqli_query($connection, 'SET SESSION sql_mode = \'ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION\'');
+	$sql_mode = '';
+	// Set the mysql default mode
+	$sql_mode .= (empyt($sql_mode) ? '' : ',') . 'ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION';
+	// Disable backlash escape, only '' is than a valid one
+	$sql_mode .= (empyt($sql_mode) ? '' : ',') . 'NO_BACKSLASH_ESCAPES';
+
+	mysqli_query($connection, "SET SESSION sql_mode = '" . $sql_mode . "'");
```
Sesquipedalian: There are typos in the lines above. Since those need to be fixed anyway, why not just put everything back on a single line?
```
mysqli_query($connection, "SET SESSION sql_mode = 'ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION,NO_BACKSLASH_ESCAPES'");
```

Sesquipedalian (changes requested):
Another minor change needed

Sources/Subs-Db-mysql.php
albertlast: i was guessing,
that i would look more cleaner,
when the different set of mods are grouped and mention in the comment what they do and
when you want to disable a specif behavior that you only comment out the right line.

Sesquipedalian: Well, if you want to do that, the cleanest way would be to use an array and then implode it to make the final string. Something like this, perhaps:
```
$sql_mode = array(
	'ONLY_FULL_GROUP_BY',
	'STRICT_TRANS_TABLES',
	'NO_ZERO_IN_DATE',
	'NO_ZERO_DATE',
	'ERROR_FOR_DIVISION_BY_ZERO',
	'NO_AUTO_CREATE_USER',
	'NO_ENGINE_SUBSTITUTION',
	'NO_BACKSLASH_ESCAPES',
);
mysqli_query($connection, "SET SESSION sql_mode = '" . implode(',', $sql_mode) . "'");
```

If you want to make it really friendly for mods, you could insert an integration hook between the array and the `mysqli_query()` call in order to give the mods edit-free access to the `$sql_mode` array.

albertlast: Mods we can't support here,
Since the connection to the db is not up

albertlast:
the array idea i like.

jdarwood007:
While we can't test the 1.0 or 1.1 very well, we still officially support upgrading from yabsee 1.5 all way to the latest version.  I do think for 3.0 we should drop this and make it a conversion for yabbse 1.5, SMF 1.0 and 1.1. Only support upgrading from 2.0 or 2.1.  Just saying that by changing this, its possible we may introduce a bug with upgrading on those versions.  You don't need to run the version with old PHP, just need the database imported into a database server to do the upgrade.

As a very small note, the original devs used single quotes over double quotes because of the slight speed improvements because of the variable handling that double quotes allows is slightly slower.  I don't have information for modern PHP or know of a way to do a test on that though.

live627:
To fix the relevant issue, ba60e1448f708b59cdbf36b55713656c9f9826e5 and ca3c9dc40a5b551ab5de0740880b70700842daa5 need cherrypicked . IMO,t the rest off this PR should wait for the next version.

albertlast:
so is this pr target of rc4 @live627 ?

jdarwood007:
If this doesn't merge for RC4, I don't see it making it to final.  The type of change here is not something we can introduce just before final.  Therefor we should determine if this wants to be made into 2.1 at all.  If we do, it needs merged now and can't be merged after RC4.

I can setup a test VM with really old PHP versions to get SMF 1.0 working, get some test data (mostly testing backslash \) escape issues.

Sesquipedalian:

PR #5186, opened 2018-12-01
Internal admin search should use call_user_func()

live627:
Allows for class methods


MissAllSunday:
Forgot to mention, perhaps using call_helper() would help out with keeping things tidy up.

live627:
you can. I find it hella confusing.

PR #5188, opened 2018-12-03
Retrieve cache info after caching, not before...

### LINKED ISSUE #5185: Image proxy - efficiency - always writes to cache twice
#### Description
The image proxy always writes newly cached files to the cache two times, causing unnecessary I/O.  This is because the checkRequest function writes to the cache, and it is called two times during an initial serve():
https://github.com/SimpleMachines/SMF2.1/blob/0931566f119074833310105db08cfaa58a14b844/proxy.php#L115
https://github.com/SimpleMachines/SMF2.1/blob/0931566f119074833310105db08cfaa58a14b844/proxy.php#L134

The second call is intended to refresh the cache for aged items, but the `!$cached || ` logic will guarantee it is invoked for previously uncached images.  

### Steps to reproduce
1. Either in a trace or via the debugger, observe how often this line is called during an initial cache:
https://github.com/SimpleMachines/SMF2.1/blob/0931566f119074833310105db08cfaa58a14b844/proxy.php#L220

- Version/Git revision:  Today's
- Database Type:  MySQL
- Database Version: 5.7
- PHP Version: 5.6

sbulen:
Fixes #5185 

I wasn't sure how far to go with this.  It is still needs to do a file_get_contents() immediately after caching a file for serve() to work.  This is odd, because it just wrote the file, & should know what it just did...  This could be addressed by having cacheImage() update object properties for time, content_type & size.  Then serve() would have access to those properties without a physical IO.   If requested, I'll make these changes as well.

### DIFF
--- a/proxy.php
+++ b/proxy.php
@@ -117,9 +117,6 @@ class ProxyServer
 	public function serve()
 	{
 		$request = $_GET['request'];
-		$cached_file = $this->getCachedPath($request);
-		$cached = json_decode(file_get_contents($cached_file), true);
-
 		// Did we get an error when trying to fetch the image
 		$response = $this->checkRequest();
 		if ($response === -1)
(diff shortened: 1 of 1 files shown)


albertlast:
Why you not define a class property and
assign to this some kind of value in the cacheimage part and
check in the serve part if the property is empty or not.

live627:
>  This could be addressed by having cacheImage() update object properties for time, content_type & size. Then serve() would have access to those properties without a physical IO. If requested, I'll make these changes as well

Could you please make those changes?

sbulen:
Tests well.  Writes only when it needs to & reads only when it needs to...

live627 (approved):

PR #5189, opened 2018-12-03
Packman browse fixes

live627:
- Download some random mods
- Download at least four avatar packs, install any two
- Sort by installed time, nothing happens
- Apply commits one and two. Sorting avatars works, except that some vanish.
- Apply the fifth commit. Sort avatars again. All four avatars now show.
- Download even more mods. Change their type to randomness. You can leave then unzipped.
- Observe the unknown packages all are enumerated as 1. The third commit fixes this.

### DIFF
--- a/Sources/Packages.php
+++ b/Sources/Packages.php
@@ -1359,6 +1359,10 @@ function PackageBrowse()
 	$context['page_title'] .= ' - ' . $txt['browse_packages'];
 
 	$context['forum_version'] = SMF_FULL_VERSION;
+	$context['available_modification'] = array();
+	$context['available_avatar'] = array();
+	$context['available_language'] = array();
+	$context['available_unknown'] = array();
 	$context['modification_types'] = array('modification', 'avatar', 'language', 'unknown');
 
 	call_integration_hook('integrate_modification_types');
@@ -1491,13 +1495,6 @@ function PackageBrowse()
 
 	$context['sub_template'] = 'browse';
 	$context['default_list'] = 'packages_lists';
-
-	// Empty lists for now.
-	$context['available_mods'] = array();
-	$context['available_avatars'] = array();
-	$context['available_languages'] = array();
-	$context['available_other'] = array();
-	$context['available_all'] = array();
 
 	$get_versions = $smcFunc['db_query']('', '
 		SELECT data FROM {db_prefix}admin_info_files WHERE filename={string:versionsfile} AND path={string:smf}',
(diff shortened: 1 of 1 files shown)


Sources/Packages.php
```diff
@@ -1356,6 +1356,10 @@ function PackageBrowse()
 	$context['page_title'] .= ' - ' . $txt['browse_packages'];
 
 	$context['forum_version'] = $forum_version;
+	$context['available_modification'] = array();
+	$context['available_avatar'] = array();
+	$context['available_language'] = array();
+	$context['available_unknown'] = array();
```
Sesquipedalian, line 1365: After resolving conflicts, I suggest removing these four lines and instead using a foreach to loop through `$context['modification_types']` to set these variables. If you put that loop after the hook, then we'll be able to handle any types added by the hook, too.

live627: Eh,  like this?

```	call_integration_hook('integrate_modification_types');

	require_once($sourcedir . '/Subs-List.php');
	foreach ($context['modification_types'] as $type)
	{
		$context['available_' . $type] = array();

		// Use the standard templates for showing this.
		$listOptions = array(
```

Sesquipedalian: Yes.

live627:
Ugh..

Sesquipedalian:
> Sesquipedalian requested this change. It makes no sense to me.

Er, that might be because the content of 569c515 is not the change I requested. I was asking for this:
```
function PackageBrowse()
{
	global $txt, $scripturl, $context, $sourcedir, $smcFunc;

	$context['page_title'] .= ' - ' . $txt['browse_packages'];
	$context['forum_version'] = SMF_FULL_VERSION;
	$context['modification_types'] = array('modification', 'avatar', 'language', 'unknown');

	call_integration_hook('integrate_modification_types');

	require_once($sourcedir . '/Subs-List.php');

	foreach ($context['modification_types'] as $type)
	{
		$context['available_' . $type] = array();

		// Use the standard templates for showing this.
		...
```

live627:
Ahem.

Your requested change could possibly introduce an edge case issue. Seems that your goal is for hooks to not have to declare `available_trash`. So I sorted that  in the best possible way I knew how . Declare the variable in the same function that writes to it.

My original solution simply declared the four empty context arrays before the hook is called. Which was mostly moving them from after the lists were created, where they were just dead weight.

The hooked functions already have to declare `$global $context` so that `$context['modification_types']`. Was your intention to not have to bring `$context` into scope? That'll end well...

I do not want to revisit this issue in the future. This aims to fix bugs, not create new ones. Ahem.

I'm out. Maybe that wasn't enough time for me to simmer down.

Sesquipedalian:
You make a good point. Please feel free to revert back to your original solution.

live627:
Commit undone.

PR #5193, opened 2018-12-10
In updateMemberData(), includes correct file for alert_count()

### LINKED ISSUE #5176: Alerts are broke
I think this says it all really:

https://www.simplemachines.org/community/index.php?topic=563540.0

1. 
2. 

- Version/Git revision:
- Database Type: mariadb
- Database Version: 
- PHP Version: 7.1 (7.0 at the time of upgrade)

Sesquipedalian:
This should fix #5176

### DIFF
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -2302,6 +2302,8 @@ function alert_count($memID, $unread = false)
 		$query_see_board = build_query_board($memID);
 		$query_see_board = $query_see_board['query_see_board'];
 	}
+	else
+		$query_see_board = '{query_see_board}';
 
 	// Find only the boards they can see.
 	if (!empty($possible_boards))
@@ -2309,11 +2311,10 @@ function alert_count($memID, $unread = false)
 		$request = $smcFunc['db_query']('', '
 			SELECT id_board
 			FROM {db_prefix}boards AS b
-			WHERE ' . (!empty($query_see_board) ? '{raw:query_see_board}' : '{query_see_board}') . '
+			WHERE ' . $query_see_board . '
 				AND id_board IN ({array_int:boards})',
 			array(
 				'boards' => $possible_boards,
-				'query_see_board' => !empty($query_see_board) ? $query_see_board : '',
 			)
 		);
 		while ($row = $smcFunc['db_fetch_assoc']($request))
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -2312,7 +2312,7 @@ function alert_count($memID, $unread = false)
 				AND id_board IN ({array_int:boards})',
 			array(
 				'boards' => array_keys($possible_boards),
-				'query_see_board' => $query_see_board
+				'query_see_board' => !empty($query_see_board) ? $query_see_board : '',
 			)
 		);
 		while ($row = $smcFunc['db_fetch_assoc']($request))
@@ -2323,7 +2323,7 @@ function alert_count($memID, $unread = false)
 	// Now check alerts again and remove any they can't see.
 	foreach ($alerts as $id_alert => $extra)
 		if (!isset($boards[$extra['board']]))
-			unset($alerts[$id_alert]);		
+			unset($alerts[$id_alert]);
 
 	return count($alerts);
 }
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -2322,7 +2322,7 @@ function alert_count($memID, $unread = false)
 
 	// Now check alerts again and remove any they can't see.
 	foreach ($alerts as $id_alert => $extra)
-		if (!isset($boards[$extra['board']]))
+		if (isset($extra['board']) && !isset($boards[$extra['board']]))
 			unset($alerts[$id_alert]);
 
 	return count($alerts);
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -2289,11 +2289,12 @@ function alert_count($memID, $unread = false)
 	{
 		$alerts[$row['id_alert']] = !empty($row['extra']) ? $smcFunc['json_decode']($row['extra'], true) : array();
 
-		// Only add to possible boards ones that are not empty and that we haven't set before.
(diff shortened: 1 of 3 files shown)


Sesquipedalian:
Please test this to see if it resolves the issue, @smfbigguy.

smfbigguy:
Not that I see.. It creates an error in Profile-modify.php:



    Type of error: Critical
    Error message: Invalid value inserted or no type specified.
    Function: alert_count
    File: /home/****/public_html/Sources/Profile-Modify.php
    Line: 2315
    URL of page causing the error: https://www.whatsurbeef.net/index.phphttps://www.whatsurbeef.net/cron.php

smfbigguy:
OK, I am going to try to explain this. You gave me an edit to put into Subs.php, which I did.  It references Profile-Modify.php This gives errors and alerts do not work. I edited that file awhile ago. When I edit a file I rename the original and leave it on the server. I had another Profile-Modify file named Profile-Modify1.php I used that file and the alerts are working now, sort of. They are working better than before anyway. 
[Profile-Modify1.zip](https://github.com/SimpleMachines/SMF2.1/files/2663499/Profile-Modify1.zip)

smfbigguy:
So I put your edit in the Suns.php but mine looks like this:

include_once($sourcedir . '/Profile-Modify1.php');

smfbigguy:
query_see_board keeps giving errors on line 2315 in Profile-Modify.php

smfbigguy:
Backtrace screenshot

screenshot_2018-12-10 backtrace

smfbigguy:
The screenshot you see is from when I changed things back to the original Profile-Modify.php So it is still erroring.

smfbigguy:
screenshot_2018-12-10 backtrace
Sorry to keep posting. I am getting alerts but not properly. It will show 3 pages of alerts but there are no alerts. When you switch pages it tells you on load of the new page you have no alerts. When an alert comes in you would not know it if you did not check alerts. This concludes my posting for now, lol.

Sesquipedalian:
Thanks, @smfbigguy. Please try again with this latest commit. It should resolve that error message from line 2315.

smfbigguy:
Still gettin errors from that line. 

Invalid value inserted or no type specified.
Function: alert_count

Sesquipedalian:
> Invalid value inserted or no type specified.

I found a bit more broken logic that I think was causing that. Care to try again with the latest commit?

smfbigguy:
This:
https://github.com/SimpleMachines/SMF2.1/pull/5193/files/72b62a376610d707f3802be30d1f9cd86facd966..667891019a02acd7bdc2b5a6dda9610f02cfbc0c

pretty much stopped the errors in the log and now I do have alerts back.

smfbigguy:
Just waitin on some members to show up now...........lmao

smfbigguy:
One problem I see. How do I get rid of alerts that I was getting when this was broken. I keep seeing I have 22 alerts but there are none. Other than this I think it all works fine so far.

smfbigguy:
Just thought I would toss up some screenshots. In this first one it says I have 22 alerts.
alerts

But I don't. I have two.

alerts1

If we go into the alerts it shows 3 pages. But there is not 3 pages. There is only two alerts.

nopages

If we delete the two alerts I do have. all the pages are gone and I have no alerts.

noalerts

Next time I get an alert, it shows I have 22 again.

jdarwood007:
Might need to add a recount function for alerts to the maintenance page.  Kinda outside of the scope of this issue though.

smfbigguy:
Would I be correct in saying that the only way to get the count right would be to go into the database and do some stuff there. ??

jdarwood007:
Receiving a new alert should cause it to recount correctly as well for that member.

smfbigguy:
Well, things are not re-counting correctly as it stands. I guess clearing the table in the database or something might make it work proper. ?? As of right now alerts work fine as far as I am concerned, it's just the count is off.

live627:
Hey,  don't clear the members table! I still want to at your site.

smfbigguy:
Lol, I meant the alerts table or notifications table or whatever it might be called. There has to be away to do it in phpmyadmin but I have not played around in there for awhile.

EDIT: Not doin it tonight either, to late.

Sources/Profile-Modify.php
```diff
@@ -2296,7 +2296,7 @@ function alert_count($memID, $unread = false)
 	$smcFunc['db_free_result']($request);
 
 	// If this isn't the current user, get their boards.
-	if (isset($user_info) && $user_info['id'] != $memID)
+	if (!isset($user_info) || $user_info['id'] != $memID)
```
albertlast, line 2300: you access $user_info without knowing if it exists,
this case could happen when this is called by the cronjob i think.

Sesquipedalian: Actually, that's the opposite of what this does. ;)

albertlast: hmm because of the "or" php stop the evaluation of the right side because the left side is already true...

Sesquipedalian: Exactly. :)

Sesquipedalian:
> One problem I see. How do I get rid of alerts that I was getting when this was broken. I keep seeing I have 22 alerts but there are none. Other than this I think it all works fine so far.

I suspect that this is due to a discrepancy between `fetch_alerts()` and `count_alerts()`. Both start out by retrieving all a user's alerts from the database, but then `fetch_alerts()` weeds out any that refer to boards that the user can't see, topics that don't exist, and/or messages that don't exist, while `count_alerts()` only weeds out ones that refer to boards that the user can't see. So if you have alerts in your database that refer to non-existent topics or messages, `count_alerts()` will include them while `fetch_alerts()` won't.

I'll put together another PR to deal with that once this one is merged.

smfbigguy:
Sounds good.

smfbigguy:
I update the files on my site. I think alerts are working now perfectly. My alert count went down.....to 1. I really hated to post this. Please don't open this again if you don't have to. All I am sayin is everything seems to be working. I just have one alert I can not find anywhere.

PR #5197, opened 2018-12-12
Improve the Query "Posting activity by time"

albertlast:
The problem mention here: https://www.simplemachines.org/community/index.php?topic=563957.0
is right in my eyes,
the mention solution is wrong.

To limit the readed amount of data i used a subquery,
important step is to use the id_msg for sorting(not poster_time),
which allowed newer mysql version (8.0+) and postgres to do a backwards index scan. 

I'm not sure if 10k messages maybe not to high,
from a statistic pov 1001 message should be enough.

### DIFF
--- a/Sources/Profile-View.php
+++ b/Sources/Profile-View.php
@@ -1401,12 +1401,13 @@ function statPanel($memID)
 			SELECT poster_time, id_msg 
 			FROM {db_prefix}messages WHERE id_member = {int:current_member}
 			ORDER BY id_msg DESC
-			LIMIT 10000
+			LIMIT {int:max_messages}
 		) a
 		GROUP BY hour',
 		array(
 			'current_member' => $memID,
 			'time_offset' => (($user_info['time_offset'] + $modSettings['time_offset']) * 3600),
+			'max_messages' => 1001,
 		)
 	);
 	$maxPosts = $realPosts = 0;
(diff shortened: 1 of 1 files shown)


realdigger:
I tested query from this commit on smf 2.0.15 database and MariaDB 10.1.37

For member with 50000 messages it takes 30.286 s
For member with 100000 messages it takes 52.040 s

Forum have 7562652 messages and 59243 topics.

albertlast:
Which index you got and which engine you use?

realdigger:
MyISAM

PRIMARY	id_msg
UNIQUE	id_topic, id_msg
UNIQUE	id_board, id_msg
UNIQUE	id_member, id_msg
INDEX	id_member, id_topic
INDEX	id_member, id_board
INDEX	id_board
INDEX	id_topic
INDEX	poster_ip(15), id_topic
INDEX	approved
INDEX	id_member, approved, id_msg
INDEX	id_topic, id_msg, id_member, approved
INDEX	id_member, poster_ip, id_msg
INDEX	subject

albertlast:
When i read the manuell right,
you need innodb.
Which is the default engine in smf 2.1 setup.
The important index is this here ``UNIQUE id_member, id_msg``

When i see right,
mariadb doesn't support backward index scan atm.

For big setups and your could be a big setup,
i (i'm not a smf team member) recommand to use postgres for smf 2.1.
Based on this knowledg here the next worst would be mysql and the most worst db would be mariadb.

Sesquipedalian:
@realdigger, how does that compare with the performance you get without this PR?

albertlast:
With the limit of 1001 messages i'm below 1ms in pg and mariadb 10.1 with innodb.

live627:
Just as a side note this looks like it will fix the missing data bug as seen in https://www.simplemachines.org/community/index.php?action=profile;area=statistics;u=2

PR #5203, opened 2018-12-17
Changes format of attach BBC to something more robust

### LINKED ISSUE #5187: If attach name has square brackets, can't insert the attach to message body properly
#### Description
If filename you want to attach contains square brackets, and you insert it to the message body, it will give "No message is associated with this attachment." error and the attachment will be linked in the end of the message.

### Steps to reproduce
1. create a photo with square brackets in the name, like "my_awesome_picture[not].jpg"
2. attach it to a message and insert the link to this attachment into the message body
3. post it

- Version/Git revision:  SMF 2.1 Beta 4 (nightly)
- Database Type: mariadb
- Database Version: 10.1.31
- PHP Version: 7.2

Sesquipedalian:
Basically, this changes `[attach name=foo.jpg type=image/jpeg]31[/attach]` to `[attach id=31]foo.jpg[/attach]`. 

Why? Three reasons:

1. Better consistency with the [url] and [img] BBCodes.
2. Improved functionality. In particular, when embedding a non-image attachment, entering 
```Please read [attach id=32]this document[/attach]```
will now result in
```Please read <a href="http://www.example.com/index.php?action=dlattach;topic=120.0;attach=32">this document</a>```
instead of 
```Please read <a href="http://www.example.com/index.php?action=dlattach;topic=120.0;attach=32">readme.txt</a>```
3. Fixes #5187
4. Eliminates the type parameter. It was only there to help SCEditor's WYSIWYG mode, and was otherwise useless, so I implemented some JavaScript to handle it transparently.
5. Cleans up some messiness in the attachment UI.

~I've maintained backwards compatibility with the old format so that [attach] BBCodes created before this gets merged will still work afterwards. If this were pegged for RC 1, I wouldn't even bother to do that, but since it is pegged for RC 2, we probably ought to.~ I've decided that this should go into RC1 (and I'm merging it shortly) because it would be silly and a long term pain in the butt if we were to release an already deprecated form of this BBC.


Darknico:
Hi! After this edit not work correctly.
If inline attach not view in bottom is disable, post [attach name=foo.jpg type=image/jpeg]31[/attach].
after enable, not change this and not work correctly.
If edit post, automatic change in [attach id=31]foo.jpg[/attach]

sorry my bad english

---- EDIT

re-test and no problem... mmmhhh.. I don't know....I think have a old build before (not remember..)
It is ok now

PR #5204, opened 2018-12-17
Doesn't try to autolink URLs inside [html] BBC or HTML img elements 

### LINKED ISSUE #5192: Enableling "Automatically link posted URLs" causes malformed img tags in posts
#### Description
--Description Here--
If Automatically link posted URLs is enabled, posts with html tags in the body are malformed when output. <a href= is inserted into the src= attribute.

The problem is in the parse_bbc function somewhere. If the call to that function in Display.php is commented out, the links are correct. Disabling Automatically link posted URLs also eliminated the problem.  
### Steps to reproduce
1. enable Automatically link posted URLs
2. View a message with embedded html 

- Version/Git revision: 2.1 Beta 4
- Database Type: MySQL
- Database Version: 5.6
- PHP Version: 5.6

Sesquipedalian:
Fixes #5192


Sesquipedalian:
Tested and working nicely.

Issue #5214, opened 2018-12-19
PG bigint vs int upgrade and install


albertlast:
When I compare pg schema upgrade with install version,
Than I notice many fields which are int in install version and bigint in upgrade version,
How do we want handle this?
For me is bigint the better way.

Sesquipedalian:
Which fields are affected by the int vs. bigint difference?

Sesquipedalian:
Actually, never mind. I should just check it myself instead of making you tell me.

Better questions: what makes bigint better in these cases, and should we use bigint for the same columns in MySQL in order to keep everything consistent?

albertlast:
bigint is used for two reason
- to get similiar number range like mysql int UNSIGNED
- lazyness, the amount of space is doubled in compare from int(4bytes) to bigint(8bytes),
since space is not an issue in our day it's small optimization without any benefit

the problem on pg side (and maybe on mysql side too) when you change the datatype the complet table get rewriten,
so to change an existing table could take very long (reason why i didn't transfer the change to pg upgrade in the first place)

I think in further version (2.1 not any more),
we should drop unsinged stuff and replace it with bigint.
To align all stuff.

Sesquipedalian:
Thanks for the explanation. I am going to leave this to your judgement. If you think it would be better for upgrade_2-1_postgresql.sql to update those columns to bigint, do it. If not, don't.

PR #5215, opened 2018-12-19
Fix all schema Diff between pg upgrade and pg install

albertlast:
Thanks to https://www.postgrescompare.com
the comparsion was easy.

The only diff which exists afters this is the diff of bigint in the install version and int field in the upgrade version.
Notice here: https://github.com/SimpleMachines/SMF2.1/issues/5214

the pr is done.

travis error get fixed by the pr #5218


albertlast:
@Sesquipedalian fixed the merge conflict

PR #5221, opened 2018-12-20
Renaming

### LINKED ISSUE #5122: Naming consistency
We have multiple names (variables, css classes and so on) that have sneaked in as camelcase or other format. These should be changed to match everything else.

Note: This doesn't apply to some widely used ones by mod/theme authors that were brought over from 2.0, like $modSettings. It would probably cause too much confusion to rename those now.

Gwenwyfar:
Didn't get to check much on sources nor the templates in full, but should be a tad tidier at least.

I actually still have another set of changes on local, for Profile.english.php, but it's such a large portion of that file that I'm not sure if we want to go forward with it. I can include it if so.

Fixes #5122


Themes/default/GenericControls.template.php
```diff
@@ -236,7 +236,7 @@ function template_control_verification($verify_id, $display_type = 'all', $reset
 			{
 				$lang = (isset($txt['lang_recaptcha']) ? $txt['lang_recaptcha'] : $txt['lang_dictionary']);
 				echo '
-				<div class="g-recaptcha centertext" data-sitekey="' . $verify_context['recaptcha_site_key'] . '" data-theme="' . $verify_context['recaptcha_theme'] . '"></div>
+				<div class="google_recaptcha centertext" data-sitekey="' . $verify_context['recaptcha_site_key'] . '" data-theme="' . $verify_context['recaptcha_theme'] . '"></div>
```
Sesquipedalian: Please revert this one. "g-recaptcha" comes from Google's own scripts. The script probably doesn't care about the class name of this div, but I'd rather err to the side of safety.

Themes/default/languages/Post.english.php
```diff
@@ -64,7 +64,7 @@
 // Use numeric entities in the below two strings.
 $txt['reported_post'] = 'Reported post';
 $txt['reported_to_mod_by'] = 'by';
-$txt['rtm10'] = 'Submit';
+$txt['submit_report'] = 'Submit';
```
Sesquipedalian: It would be more consistent with our naming conventions if this were `$txt['report_submit']` instead of `$txt['submit_report']`

Sesquipedalian (changes requested):
Just a couple changes needed. Also, I'm unsure why you felt it was necessary to change `generic_icons` to `main_icons`...

Sesquipedalian:
FYI, Travis is failing because of a problem on their end, not because of problems in this PR.

Gwenwyfar:
I won't lie that it's a bit of a pet peeve... I don't like names that don't make sense, and this doesn't in many ways. Imagine if we started adding "generic" to all of our classes just because they are the default. And the icons are not generic, just the main ones. I'm not completely sure but I think at some point it was named just "icons" (or at least there were leftovers of an attempt of this in the code), which makes much more sense, but since we already have two sets of them (admin icons), "main icons" matches that. Also you can then add whatever_icons later and it leaves the definition of which ones are what clearer.

Sesquipedalian:
Fair enough.

Sesquipedalian (approved):

Sesquipedalian:
Regarding the Profile language strings, I would like to hear from @frandominguez03 before renaming a huge number of $txt variables. If it makes no difference to the translators, then it would probably be fine. But if it would create a bunch of work for them, it's not worth it.

So let's say this: I will go ahead and merge this PR now, and you can put your proposed changes to the Profile $txt variables into a separate PR. When you submit that PR, tag @frandominguez03 in the PR description so that he can give us a translator's perspective on the matter. If he thinks it is fine and won't cause delays for them, great, but otherwise we'll have to skip it.

Sesquipedalian:
Actually, I've changed my mind. I want to get a test build of RC1 into the hands of our official beta testers team by Friday evening so that we can have any hope of getting some of them to test it over the holidays. We're out of time if we are going to meet that goal.

frandominguezl:
While I agree with the changes made, I'm not sure it's worth putting them on stage at this moment. We're still pending the Crowdin thing, and I don't really know how Language Editor handles this (maybe, if we update the var names, it will show up as new string to be translated).
I'd say to wait before going ahead with the language changes.

Sesquipedalian:
Thanks, @frandominguez03. I agree. We won't bother with renaming any other $txt variables at this point.

PR #5226, opened 2018-12-22
Fixes awkward handling of legacy BBC

### LINKED ISSUE #5283: Flash is disabled on the toolbar with no way to fix
#### Description
The WYSIWYG toolbar has a Flash button on it on a fresh install - but it is greyed out, even though the Flash bbcode is enabled.

image

The original 'enable Flash' option appears to have been removed though the Flash bbcode is still present.

image

This feels unfinished - if supporting backwards compatibility is a thing, leave the bbcode in but disable it in the editor so it still works if editing legacy posts, while hiding it in the toolbar for new users.

Flash support is going to disappear in browsers from next summer onwards (even if official EOL is 2020), might as well properly disable it in new installs, and leave just enough in upgrades to work if not nicely.

### LINKED ISSUE #5284: Legacy bbcodes - inconsistent handling
#### Description
I get that notionally, old bbcodes should be hidden, e.g. glow, shadow. And the rendered version, fine, it doesn't break, it just doesn't do anything.

However, they're still listed in the admin panel in the bbcode section, which looks like they should be supported. If they're not meant to be really supported, would it be possible to remove them from the admin panel?

There's a similar situation with the color bbcodes - black, blue, green, red, white - these aren't real bbcode in the conventional sense because they're all converted during preparse to be color bbcodes, so it seems like these should be removed and let the preparser (plus any during-upgrade conversion) deal with it.

Sesquipedalian:
With these changes, legacy BBC can never be used in new posts, but they will continue to work if they are already present in pre-existing posts. This is because preparsecode() will strip them out (even if they are enabled in the admin settings), whereas parse_bbc() will treat them normally.

To make the difference between normal BBC and legacy BBC clearer to the admin, the two groups have been visually separated on the admin's BBC settings page and some help text has been added. This is purely a UI matter. There are no changes to how information about disabled BBC is stored in the database.

Also, the legacy BBC are now disabled by default for new installs, but are left alone during upgrades.

Edit: 
Fixes #5283
Closes #5284


Sesquipedalian:
I going to go ahead and merge this now, in part because all testing has shown that it works well and in part to make things simpler for the Documentation team (so that they don't need to write one version of BBCode documentation for RC1 and a different version for RC2 and beyond).

PR #5227, opened 2018-12-22
RC1 last minute stuff

Sesquipedalian:


albertlast:
Did you not miss a space between the C and 1?

Never mind, i looked in the archieve,
smf called the beta version alway with a space betwen the char and the number and rc always without a space.

Sesquipedalian:
> Never mind, i looked in the archieve,
> smf called the beta version alway with a space betwen the char and the number and rc always without a space.

Yeah, I was surprised by that too. 🤷‍♂️

albertlast:
Just for note,
i see the need,
that after a short periode of time (one or two weeks) the version had to renamed again
into something like,
RC2 or github or nightly,
only for the reason that we can distinguish download version vs github version.

Sesquipedalian:
We might, yes. But if I succeed with my plan to switch us over to a "release early, release often" model after RC1 is released, the time between releases henceforth should be short enough that it won't matter as much any more.

PR #5229, opened 2018-12-22
Fix mobile menu height

SychO9:
The important rule is not necessary here, 
This change does not re-open the issues fixed here #5086

mobile menu fix

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -2677,7 +2677,7 @@ dl.addrules dt.floatleft {
 	position: relative;
 	background: none;
 	margin: 0;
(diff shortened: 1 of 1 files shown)


Gwenwyfar (approved):

XinYenFon:
The only reason I added important to that because to prevent the future issues if something somehow overrides the rule... more like idiot-proofing. 😆 

Tho adding important to 480p (#main_menu .popup_container, #genericmenu .popup_container) seems to fix the height background issue...

SychO9:
I think it's best not to add any **!important** rules unless we really have to

PR #5231, opened 2018-12-22
Fixing Website Sort and Group sort for pg

albertlast:
Error msg: ERROR: COALESCE types character varying and boolean cannot be matched
``LINE 6: ....website_url) > 0 DESC, COALESCE(mem.website_url, 1=1) ASC, ..``

You get this when smf run with pg and you go to memberlister and order the list by website.


Sesquipedalian:
Have you tested to make sure that this does not change behaviour with MySQL?

albertlast:
sure did make no changes to mysql

PR #5240, opened 2018-12-23
Fix $last_id in mysql fixes #5234

### LINKED ISSUE #5234: sources/Subs-Db-mysq.php $last_id
Looks like last_id should be $old_id in the code below since $last_id is not defined any where

`			$old_id = $smcFunc['db_insert_id']();

			$smcFunc['db_query']('', '
				' . $queryTitle . ' INTO ' . $table . '(`' . implode('`, `', $indexed_columns) . '`)
				VALUES
					' . $insertRows[$i],
				array(
					'security_override' => true,
					'db_error_skip' => $table === $db_prefix . 'log_errors',
				),
				$connection
			);
			$new_id = $smcFunc['db_insert_id']();

			// the inserted value was new
			if ($last_id != $new_id)
			{
				$ai = $new_id;
			}
			// the inserted value already exists we need to find the pk
			else
			{`

albertlast:
Issue #5234

### DIFF
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -819,7 +819,7 @@ function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $ret
 			$new_id = $smcFunc['db_insert_id']();
 
 			// the inserted value was new
-			if ($last_id != $new_id)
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Thanks

PR #5242, opened 2018-12-23
Update Readme with new technical information

### LINKED ISSUE #5237: other/readme.html bump up php version review database versions and paths to sytle and css is wrong
other/readme.html bump up php version review database versions

Seems to say php 5.3.8 is supported along with mysql 5.0

And the path to logo and style css is wrong since the readme.html on the build is the root folder instead of the other folder.

albertlast:
Partly fix #5237

I only fixed the technical information.

### DIFF
--- a/other/readme.html
+++ b/other/readme.html
@@ -122,8 +122,8 @@
 					</li>
 					<li>Any of the following database systems
 						<ul>
-							<li><a href="https://www.mysql.com/" target="_blank" rel="noopener">MySQL</a> 5.0.22 or higher.</li>
-							<li><a href="https://mariadb.com/" target="_blank" rel="noopener">MariaDB</a> 5.1.00 or higher.</li>
+							<li><a href="https://www.mysql.com/" target="_blank" rel="noopener">MySQL</a> 5.0.22/5.1.10 or higher.</li>
+							<li><a href="https://mariadb.com/" target="_blank" rel="noopener">MariaDB</a> 5.1.10 or higher.</li>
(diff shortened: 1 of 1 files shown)


other/readme.html
```diff
@@ -136,7 +137,7 @@ <h2 id="recommendations">Recommendations for best performance</h2>
 					<li>the <a href="http://aspell.net/" target="_blank" rel="noopener">GNU Aspell</a> 0.50 or higher and <a href="https://ftp.gnu.org/gnu/aspell/dict/" target="_blank" rel="noopener">its dictionaries</a> for spell checking functionality.</li>
```
live627, line 137: Do we even support aspell? I don't remember.

albertlast: I see php code how call pspell, which require aspell -> yes?

SychO9:
https://github.com/SimpleMachines/SMF2.1/blob/12fa4d4f2379c6af7333a6f20604af90cf37d073/other/readme.html#L286-L292

Shouldn't php be 7.0.0 for upgrade recommandations here as well ?

albertlast:
Well yes this part should be updated too
From technical pov is the newest version = best.
This is valid for php, mysql, maria and pg.

other/readme.html
smfbigguy: You need both aspell and pspell for spellcheck to work as far as I know.

Oldiesmann: In 2.1 we also support [enchant](https://secure.php.net/enchant) for spell check.

PR #5264, opened 2018-12-24
Undo renaming of "smfAnnouncements" div to avoid breaking the news

Oldiesmann:

### DIFF
--- a/Themes/default/Admin.template.php
+++ b/Themes/default/Admin.template.php
@@ -33,7 +33,7 @@ function template_admin()
 									</h3>
 								</div>
 								<div class="windowbg nopadding">
-									<div id="smf_news">', $txt['smf_news_cant_connect'], '</div>
+									<div id="smfAnnouncements">', $txt['smf_news_cant_connect'], '</div>
 								</div>
 							</div>';
 
--- a/Themes/default/Admin.template.php
+++ b/Themes/default/Admin.template.php
@@ -121,7 +121,7 @@ function template_admin()
 									%message%
 								</dd>
 							'), ',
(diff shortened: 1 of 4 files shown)


albertlast:
i "forced" oldy to give a proper describtion,
here it is:

The javascript to display the SMF news in the admin center expects the div to have an ID of "smfAnnouncements". By default that div displays a text string saying "You are unable to connect to simplemachines.org's latest news file.". The JavaScript replaces that text with the appropriate news that it gets from simplemachines.org
So if the div is named "smf_news" instead of "smfAnnouncements", the JavaScript won't work

Gwenwyfar:
And this is why I hate doing changes without testing everything first... but oh well. 

The javascript was also renamed, are you sure this fixes it?

https://github.com/SimpleMachines/SMF2.1/pull/5264/files#diff-b67fb0c40e61a708da98389c4692c021L124

Edit: Just tested it and it doesn't. Must be something else in the javascript.

Oldiesmann:
I'm not sure what that change does exactly, but the JavaScript to actually display the news items is pulled directly from simplemachines.org, so we'd have to change it on that end for this renaming to work. That JS expects it to be smfAnnouncements.

Oldiesmann:
Didn't see the comment about the JS being renamed until @SychO9 pointed out the changes... Latest commit should take care of those as well

jdarwood007:
Note, its not that we can't change it, but this change has to go through approval from the lead developer and Site team (Me).  My handling will be needing to update the javascript code to both support all editions properly.  Typically this is something we should do early on in the development process.

PR #5266, opened 2018-12-24
Change the error message to a monospace type

albertlast:
Postgres try always give a hint where the error,
more precicly than mysql do:
grafik
Look at the arrow

To that work,
the text needs to be in a monospace and space needs to be shown.
old:
grafik

new:
grafik

@Gwenwyfar what think you?


Gwenwyfar:
Looks fine to me. But what about other messages that display there?

albertlast:
Well all kind of error message would be displayed in monospace.
The other fields i would keep them like them are.

Gwenwyfar:
Well, I suppose it's ok. But I'd use something more like a generic monospace or similar, it matches better the default font (less of a style clash between the two).

albertlast:
Feel free to give me a list,
i only choose some random monospaces.

Themes/default/css/admin.css
```diff
@@ -624,6 +624,8 @@ h3.grid_header {
 #error_log span.error_message {
 	overflow: hidden; /* This changes how the text wraps around the float */
 	display: block;
+	white-space: pre-wrap;
+	font: 86%/150% "Courier", "Inconsolata", "Apercu Mono", "Nitti", "Monaco";
```
Sesquipedalian: Instead of manually defining the font here, I suggest adding `span.error_message` to this list:
https://github.com/SimpleMachines/SMF2.1/blob/888b2677c8bf640776f56d446101c42a4ca4492b/Themes/default/css/index.css#L78

Note: you will first need to merge recent changes to release-2.1 into this PR in order to do that.

Sesquipedalian (dismissed):

Themes/default/css/admin.css
albertlast: the white-space wraper needs to stay ther?

Gwenwyfar (approved):

PR #5286, opened 2018-12-26
Implement new logic for sorting fixes #5278

### LINKED ISSUE #5278: Initial values for custom profile fields ordering
#### Description
When first presented with the custom profile fields on a fresh install, the ordering looks like this:

image

Out of the box, you can switch location and gender around without any problems, but in order to move location before Skype (assuming you ever cared to), you need to hit up twice (once to move it to 3 rather than 4, and then once again to move it to 2)

Doing so also causes there to be a gap before Gender, when the order numbers will be 1, 2, 3, 5.

albertlast:
Since the pr is final,
some motivation the existing solution got two issues:

First it handle gabs between field_order very bad -> when between two fields a gab of 10 int is,
than you need to push 10 time the field up to be in front of
Secondly when two fields got the same field_order things get horrible wrong:
grafik
In this case your are not able to get out of this corrupt state.

First issue get fixed because we generated for all field everytime the field_order,
so the admin/user had to click once and the field move in wanted direction.
Second issue get fixed by the same logic,
since we build the order everytime new,
the field_order get corrected after the update of one field.

To forbid(by using unique key) the case of doublicate field_order is my eyes not an option.

issue #5278

Since the numbers of field_order got less important work this pr very well with the pr  #5289

### DIFF
--- a/Sources/ManageSettings.php
+++ b/Sources/ManageSettings.php
@@ -1712,7 +1712,7 @@ function EditCustomProfiles()
 	// We need this for both moving and saving so put it right here.
 	$order_count = custFieldsMaxOrder();
 
-	if ($context['fid'])
+	if ($context['fid'] && !isset($_GET['move']))
 	{
 		$request = $smcFunc['db_query']('', '
 			SELECT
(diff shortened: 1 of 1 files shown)


albertlast:
@Arantor would be nice if you test this pr.

sbulen:
I tested with an install & an upgrade with lots of custom fields.  All on WAMP.  Looked good.

MissAllSunday (approved):

PR #5289, opened 2018-12-26
Remove the number of Field Order in the display

albertlast:
grafik
It's look cleaner and give the user less chance to get confused.

For compare the "old" one.
grafik


DiegoAndresCortes:
@Sesquipedalian Can we update the milestone of some of these 'Theme' pr's, most have _Final_ with small changes

frandominguezl:
I agree with Diego. I believe if we're going to push in Theme changes, we would need to test them all before we go final. I think RC3 is a good opportunity to do so.

Sesquipedalian (approved):

PR #5294, opened 2018-12-28
Fixes problems with load_lang_file() in upgrader; also makes most strings in upgrader localizable

### LINKED ISSUE #5273: upgrade.php hardcoded error text
#### Description
Error text message are hardcoded and cannot be translated

Example
https://github.com/SimpleMachines/SMF2.1/blob/cc6b057dda6ec6cdadfb610ef385113c6e7e82ba/other/upgrade.php#L636 

- Version/Git revision: Pull Request 5620
- Database Type: MySQL
- Database Version: 5.6.42-log 
- PHP Version: 7.1.22

It would be advantageous if the texts could be translated.

Sesquipedalian:
Fixes #5273

I ended up revising the upgrader's version of `load_lang_file()` a good deal in the process of creating this PR, because it was only sort of working. A lot of the code appeared to have been copied directly from the similar function in install.php, but a lot of assumptions that are true in the installer are not in the upgrader.

Anyway, this revised version of `load_lang_file()` is called once relatively early in the process before any database calls are made, and then again after the theme_dir path has been retrieved from the database.

During the first call, the function checks whether the languages files exist at the standard location (i.e. `./Themes/default/languages`) and loads them if it can. If not, then it falls back to defining a few necessary $txt variables to get us through until we can try again with the correct path information. (The first part of this is the same method that `load_lang_file()` used before this PR.)

During the second call, the function first checks whether the theme_dir path is actually any different than the default path it already tried. If it is the same, then it returns without bothering to repeat any unnecessary work. Otherwise, it uses the theme_dir path to find and load the install language file for the appropriate language.

The upshot of all this is that if the language files are in the standard location, they can be localized from (almost) the very beginning. If they are not in the standard location, the upgrader will use English for the first few potential error messages until it can localize them.


other/upgrade.php
```diff
@@ -540,8 +540,9 @@ function redirectLocation($location, $addForm = true)
 // Load all essential data and connect to the DB as this is pre SSI.php
 function loadEssentialData()
 {
-	global $db_server, $db_user, $db_passwd, $db_name, $db_connection, $db_prefix, $db_character_set, $db_type, $db_port;
-	global $db_mb4, $modSettings, $sourcedir, $smcFunc;
+	global $db_server, $db_user, $db_passwd, $db_name, $db_connection,
```
frandominguezl: Missing semicolon

frandominguezl:
I'm quite sure there are a lot of places around install.php and upgrade.php in where die() is used with hardcoded text. Not sure if it's worth localizing all those errors.

wintstar:
> I'm quite sure there are a lot of places around install.php and upgrade.php in where die() is used with hardcoded text. Not sure if it's worth localizing all those errors.

It is always worthwhile to put users to peace. There are many users of SMF who do not speak English. And I think SMF should be user friendly.

albertlast:
Well this is not fully right,
for exp. when the user get a german error message.

How should he solve this,
when the orginal error message is in english where the solution already known?

That this a real issue,
show in 2.0.X the login issue with old thema + new security patches.

Where the english error message got well known and
the user know how to to solve this,
but the german version he get lost.

albertlast:
A nice angel,
add some other information to this topic
``
if you don't have db connection, you cannot load $settings, so you don't have the path to the default theme and without it you don't have the path to the language files, so it's not possible to load language files and that's the reason why this message was not internationalized to begin with.
``

frandominguezl:
While I agree with @wintstar as a non-English native speaker, I do think that this kind of messages with sensitive information should remain untranslated for mere support issues.

Do have in mind that most of support given in Simple Machines forum is in English. If users find an error in their language and ask for support in the proper board in simplemachines.org, they probably will get an answer much later than if they do in English.

wintstar:
If it is in German for example, you can usually help yourself. Then you do not need a request to SMF.

albertlast:
This is less the case,
only the error message is in german,
doesn't mean you know what you had to do.

frandominguezl:
> If it is in German for example, you can usually help yourself. Then you do not need a request to SMF.

Sorry I don't agree with you. Not everyone has the proper knowledge to know how to solve some issues.

Sesquipedalian:
> A nice angel,
> add some other information to this topic
> `if you don't have db connection, you cannot load $settings, so you don't have the path to the default theme and without it you don't have the path to the language files, so it's not possible to load language files and that's the reason why this message was not internationalized to begin with.`

Actually, `load_lang_file()` uses hard coded paths to find the language files, so that's not true.

Sesquipedalian:
@frandominguez03 and @wintstar, can you both please test this? I was planning to leave the original problem of localizing upgrader strings until RC2, but I discovered some problems with `load_lang_file()` that really ought to be fixed before we release RC1.

wintstar:
I'm already translating the new strings. Reporting me back.

wintstar:
I would like to make a suggestion for this line of text:
`$txt['error_unexpected_template_call'] = "\n" . 'Error: Unexpected call to use the %1$s template. Please copy and paste all the text above and visit the SMF support forum to tell the Developers that they\'ve made a mistake; they\'ll get you up and running again.';
`
> Please copy and paste all the text above and visit the SMF support forum to tell the Developers that they\'ve made a mistake; they\'ll get you up and running again.
>
 I think the statement "the Developers that they\'ve made a mistake" too hard, I think that would be better:

> Please copy and paste all of the above text and visit the SMF Support Forum to let the developers know that there is a bug; so they can fix it.

wintstar:
My tests with still pull request 5620 and the fix here were successful.

Could this warning message also be made translatable?
```
Warning: require_once(C:\xampp\htdocs\smf21/Settings.php): failed to open stream: No such file or directory in C:\xampp\htdocs\smf21\index.php on line 40

Fatal error: require_once(): Failed opening required 'C:\xampp\htdocs\smf21/Settings.php' (include_path='\xampp\php\PEAR') in C:\xampp\htdocs\smf21\index.php on line 40
```

The installation and used with the language files from SMF 2.0.* with the prefix *-utf8 should not be more possible.

install_with_utf-8-files

install_with_utf-8-files-1

Sesquipedalian:
That's a raw PHP error message, not something we control. I suppose I could build in a test to see if Settings.php exists and die with our own error message if it doesn't, but I don't think that would accomplish much. If we can't find Settings.php in the expected location, we probably couldn't find the language files in their expected location either, so whatever error message we emit probably won't be localized anyway.

wintstar:
Ok. I assume that there will be a documentation for 2.1, how to install and update. But what about the old "-utf8".php language files. Can this be prevented, that they can be used?

Sesquipedalian:
> But what about the old "-utf8".php language files. Can this be prevented, that they can be used?

That probably shouldn't be happening. I'll need to figure out why those files are not being filtered out, and fix it.

Sesquipedalian:
Okay, I think I've now managed to fix all the problems related to language support in the upgrader. It seems to be quite solid at this point—or at least, I can't find any more ways to break it.

PR #5297, opened 2018-12-28
Reply to Warning PM handling Issue #5267

### LINKED ISSUE #5267: Private Message sent via Warning system has no recipient when replying
#### Description
A user that receives a warning can reply to the warning Private Message.
However, no recipient is auto-populated.

If a Warning PM purposely has no recipient, it shouldn't be made possible to reply. Otherwise, SMF should populate an admin/moderator user or group (of choice?) as recipient.

### Steps to reproduce
1. Create warning for user
2. Log in as user and view PM warning
3. "Reply with quote"
4. Type reply and send -> Results in "no recipient" error

- Version/Git revision: 2.1 beta 3 - git clone from 21/12/2018
- Database Type: MySQL
- Database Version: 5.7.14
- PHP Version: 7.0.10

shnazzle:
prevent showing "reply with quote" button when opening a warning PM notification and adjust notification template for English to inform users to contact admin

fix #5267


live627:
Rebase completed successfully.

PR #5299, opened 2018-12-29
Define ubuntu version for Travis

albertlast:
Let's see if the test are running well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: xenial
+dist: trusty
 sudo: required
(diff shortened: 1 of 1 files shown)


albertlast:
Look like xenia only support php 5.6+
but since we drop the support for php 5.3 we could used now trusty.

And trusty was already used by the build system,
but maybe is better to mention which ubuntu version we want,
instead of hope that the default is the right one.

Sesquipedalian (approved):

PR #5302, opened 2018-12-30
Substitute for array_column() for php 5.4

### LINKED ISSUE #5301: Upgrader - problem with php 5.4 - usage of array_column()
#### Description
Upgrader - problem with php 5.4 - usage of array_column() here:
https://github.com/SimpleMachines/SMF2.1/blob/3f92b91bfec21917d27afa787cf3a7c90cf2949e/other/upgrade_2-1_mysql.sql#L2610

That's my code...  I wasn't aware array_column() was only supported in php 5.5+...  

### Steps to reproduce
1. Attempt a 2.1 upgrade using php 5.4.  It will hang on the new smiley logic.

- Version/Git revision: Today's
- Database Type: Mysql
- Database Version: 5.7
- PHP Version: 5.4

sbulen:
Fixes #5301 

Testing thus far looks good.  Doing further tests.  Getting this out there for visibility/review.

### DIFF
--- a/other/upgrade-helper.php
+++ b/other/upgrade-helper.php
@@ -419,3 +419,24 @@ function smf_mysql_real_escape_string($string)
 	global $db_connection;
 	mysqli_real_escape_string($db_connection, $string);
 }
+
+/**
+ * Substitute for array_column() for use in php 5.4
+ *
+ * @param $array to search
+ * @param $col to select
+ * @param $index to use as index if specified
+ * @return array of values of specified $col from $array
+ */
+function array_column_upgr($array, $col, $index = NULL)
+{
+	$output = array();
+	foreach($array AS $row)
+	{
+		if ($index === NULL)
+			$output[] = $row[$col];
+		else
+			$output[$row[$index]] = $row[$col];
+	}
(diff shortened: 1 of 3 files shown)


Sesquipedalian:
Nice, but it doesn't include any error checking. I recommend pasting [this bit of code](http://php.net/manual/en/function.array-column.php#118831) into upgrade-helper.php instead of that function. Then you won't even need to remove array_column() from the SQL files.

Sesquipedalian:
For the sake of time, I went ahead and implemented that change myself, @sbulen, so don't worry about it.

sbulen:
ACK!  I've been testing a new version all morning...

Sesquipedalian:
Oh. Sorry. 🙁

Sesquipedalian:
For what it's worth, here is what was ultimately merged in: 98d5cfd. I took this PR and then made another commit (0880e56) on top of it.

sbulen:
NP.

I'll do another round of testing with all the new logic to be safe.

Sesquipedalian:
Okay, great. Thanks!

PR #5305, opened 2018-12-31
ssl_cert_found workaround for php 5.4 and 5.5 fixes #5304

### LINKED ISSUE #5304: Php 5.4 - Issue with cert detection?
#### Description
I am having a problem with cert detection in php 5.4.   I can't tell if this is an environment config issue or a real issue with the code.  This can cause issues with attempting to enable SSL.  

### Steps to reproduce
1.  In a PHP 5.4 environment that has a valid cert, do a fresh install & try to enable SSL
OR...
2. In a PHP 5.4 environment that has a valid cert, in the ACP, Disable SSL then attempt to re-enable it

- Version/Git revision:  Todays
- Database Type: MySQL
- Database Version:  5.6
- PHP Version:  5.4

This always returns false for me in 5.4:
https://github.com/SimpleMachines/SMF2.1/blob/98d5cfd420ee2db29b1fd793751cc06e5ba4eb05/Sources/Subs.php#L6327

albertlast:
For the complete story look at https://github.com/SimpleMachines/SMF2.1/issues/5304#issuecomment-450644639
issue #5304

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -6326,19 +6326,21 @@ function build_regex($strings, $delim = null, $returnArray = false)
  */
 function ssl_cert_found($url)
 {
-	// This Check works only with PHP 5.6+
-	if (version_compare(PHP_VERSION, '5.6.0', '<'))
-	{
-		return true;
-	}
-
 	// First, strip the subfolder from the passed url, if any
 	$parsedurl = parse_url($url);
 	$url = 'ssl://' . $parsedurl['host'] . ':443';
 
 	// Next, check the ssl stream context for certificate info
+	if (version_compare(PHP_VERSION, '5.6.0', '>='))
+		$ssloptions = array("capture_peer_cert" => true, "verify_peer" => true, "allow_self_signed" => true);
+		
+	else if (extension_loaded('openssl')) // php below 5.6 needs openssl
+		$ssloptions = array("capture_peer_cert" => true);
+	else
+		return true; 
+
 	$result = false;
-	$context = stream_context_create(array("ssl" => array("capture_peer_cert" => true, "verify_peer" => true, "allow_self_signed" => true)));
+	$context = stream_context_create(array("ssl" => $ssloptions));
 	$stream = @stream_socket_client($url, $errno, $errstr, 30, STREAM_CLIENT_CONNECT, $context);
 	if ($stream !== false)
 	{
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
I recommend this:
```
function ssl_cert_found($url)
{
	// This check won't work without OpenSSL
	if (!extension_loaded('openssl'))
		return true;

	// First, strip the subfolder from the passed url, if any
	$parsedurl = parse_url($url);
	$url = 'ssl://' . $parsedurl['host'] . ':443';

	// Next, check the ssl stream context for certificate info
	if (version_compare(PHP_VERSION, '5.6.0', '<'))
		$ssloptions = array("capture_peer_cert" => true);
	else
		$ssloptions = array("capture_peer_cert" => true, "verify_peer" => true, "allow_self_signed" => true);

	$result = false;
	$context = stream_context_create(array("ssl" => $ssloptions));
	$stream = @stream_socket_client($url, $errno, $errstr, 30, STREAM_CLIENT_CONNECT, $context);
	if ($stream !== false)
	{
		$params = stream_context_get_params($stream);
		$result = isset($params["options"]["ssl"]["peer_certificate"]) ? true : false;
	}
	return $result;
}
```

Sesquipedalian (changes requested):
See discussion comment for recommended changes.

albertlast:
Since php 5.6+ doesn't need open ssl i changed your code a little bit.

sbulen:
Do we have to get rid of support for self-signed certs?  Not a smart move, makes it impossible to test https locally...

albertlast:
Well you got two option,
use modern php or disable openssl (which is by default disabled).

Sesquipedalian:
> Since php 5.6+ doesn't need open ssl

Are you sure about that? It looks to me like PHP 5.6 is still using OpenSSL for this job.

Sesquipedalian:
> Do we have to get rid of support for self-signed certs? Not a smart move, makes it impossible to test https locally...

If I understand [this](http://php.net/manual/en/migration56.incompatible.php#migration56.incompatible.peer-verification) correctly, SSL certificates were not verified by default prior to PHP 5.6.  I believe that means self-signed certificates should pass by default on PHP 5.4 and 5.5. (I could be wrong, though.)

sbulen:
I just tested this PR in its current form on my two environments:
 - WAMP, php 5.6, self-signed - works OK
 - linux hosted, php5.4, real cert - works OK

In both instances I am not prevented from enabling SSL.

albertlast:
Sure like i said when openssl is not enable the test is always true,
because ther is no test:
https://github.com/SimpleMachines/SMF2.1/blob/20eed227d665c2aa0e2d093044525da293f868ab/Sources/Subs.php#L6340

PR #5307, opened 2019-01-01
Add id_thumb index to Attachments

albertlast:
In some statement is id_thumb part of where or join,
on bigger setup this type of query get very very slow: https://www.simplemachines.org/community/index.php?topic=564301.0

This pr fix this issue for 2.1

Since my target is that smf 2.1 is well designed for bigger setup,
are pr like this (small changes but big help for this kinds of setup) very important.

### DIFF
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -2919,6 +2919,6 @@ VALUES ('Independence Day', '1004-07-04'),
 --- Add Attachments index
 /******************************************************************************/
 ---# Create new index on Attachments
-CREATE INDEX idx_id_thumb ON {$db_prefix}attachments(id_thumb);
+CREATE INDEX idx_id_thumb ON {$db_prefix}attachments (id_thumb);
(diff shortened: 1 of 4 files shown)


Sesquipedalian (approved):

albertlast:
sry forgot to add idx_ to the mysql setup file.

other/upgrade_2-1_mysql.sql
```diff
@@ -2914,3 +2914,11 @@ VALUES ('Independence Day', '1004-07-04'),
 	('Labor Day', '2030-09-02'),
 	('D-Day', '1004-06-06');
 ---#
+
+/******************************************************************************/
+--- Add Attachments index
+/******************************************************************************/
+---# Create new index on Attachments
+CREATE INDEX idx_id_thumb ON {$db_prefix}attachments(id_thumb);
```
Sesquipedalian: Nit: Should probably have a space between `{$db_prefix}attachments` and `(id_thumb)`

PR #5311, opened 2019-01-02
Correctly processes BBCodes with all optional parameters

Sesquipedalian:
Previously, if a BBCode's parameters were all optional, parse_bbc() would nevertheless refuse to process it unless at least one parameter was specified in the post text. With this change, parse_bbc() will accept a BBCode with no parameters specified if all of its parameters are optional.


live627:
Going where no dev has gone before, brave soul.

Sesquipedalian:
Next up...

albertlast:
You notice the issue https://www.simplemachines.org/community/index.php?topic=564275.0 ?

Sesquipedalian:
That wasn't supposed to be merged yet... Oh well. No harm done.

PR #5312, opened 2019-01-02
Changes year to 2019, cleans up some whitespace, formatting, etc.

Sesquipedalian:


other/upgrade.php
```diff
@@ -3755,7 +3756,7 @@ function template_upgrade_below()
 	</div><!-- #footerfix -->
 	<div id="footer">
 		<ul>
-			<li class="copyright"><a href="https://www.simplemachines.org/" title="Simple Machines Forum" target="_blank" rel="noopener">SMF &copy; 2018, Simple Machines</a></li>
+			<li class="copyright"><a href="https://www.simplemachines.org/" title="Simple Machines Forum" target="_blank" rel="noopener">SMF &copy; ' . $GLOBALS['software_year'] . ', Simple Machines</a></li>
```
albertlast: Without testing i would guess,
that $GLOBALS is not defined here or?

Sesquipedalian: $GLOBALS is a [superglobal](http://php.net/manual/en/language.variables.superglobals.php), so it is always already defined and available everywhere.

However, I'm about to push another commit that switches to using a constant for this, because this value is never supposed to be changed during the execution of the script.

Sesquipedalian:
By the way, @albertlast, does 6721296 look good to you? Aside from the whitespace changes, I also took out a few parentheses that didn't appear to be doing anything. I don't see any difference in behaviour, but your review would be appreciated.

PR #5332, opened 2019-01-07
Add smiley fixes

### LINKED ISSUE #5328: Upload new smiley does not work
#### Description
Adding a new smiley still uses the default of "use existing file" instead of the files you selected to upload.

### Steps to reproduce
1. Add a new smiley through upload
2. The new smiley will be one of the selected existing files.

- PHP Version: 5.6

### LINKED ISSUE #5331: Cannot upload smileys with different extensions
Even if you have sets with different extensions, you cannot upload files matching that extension.

Error message:
"All smileys must have the same filename."

"Filename" must be a comparison of the full filename. I tried gifs and pngs with the same name excluding the extension.

In addition to this, maybe this should not even be an error at all, because I don't see why we want to block whatever extension you want to use on any set.

sbulen:
Fixes #5331 - multi-extension add erroneous message
Fixes #5328 - upload new smiley issue

Also tightened edits to prevent uploading the wrong file/extension into a set's folder.

I have one question regarding the solution to a 'special characters' issue I discovered here, in this case, the space in "Alienine's Set" & "Fugue's Set".  Internally, this name is used within a variable name.  PHP doesn't let you use spaces or periods in variable names, & swaps them out with underscores unannounced.  Thus, some of the internal edits & processing were getting bypassed because it didn't get a hit on the set name.  I just swapped out dots & spaces with underscores on the _other_ field before the comparison to address.  

My question is - is there a php function to do this?  I couldn't find one.

It would be cleaner in the long term to use a supported function for the variable name conversion that is going on.  Good description of this conversion is in the answer here:
https://stackoverflow.com/questions/17092398/post-spaces-converted-in-underscores

### DIFF
--- a/Sources/ManageSmileys.php
+++ b/Sources/ManageSmileys.php
@@ -778,7 +778,7 @@ function AddSmiley()
 				'code' => 'string-30', 'filename' => 'string-48', 'description' => 'string-80', 'hidden' => 'int', 'smiley_order' => 'int',
 			),
 			array(
-				$_POST['smiley_code'], $_POST['smiley_filename'], $_POST['smiley_description'], $_POST['smiley_location'], $smiley_order,
+				$_POST['smiley_code'], pathinfo($_POST['smiley_filename'], PATHINFO_FILENAME), $_POST['smiley_description'], $_POST['smiley_location'], $smiley_order,
 			),
 			array('id_smiley')
 		);
--- a/Sources/ManageSmileys.php
+++ b/Sources/ManageSmileys.php
@@ -702,8 +702,8 @@ function AddSmiley()
 					fatal_lang_error('smileys_upload_error_blank');
 
 				if (empty($newName))
-					$newName = basename($_FILES[$name]['name']);
-				elseif (basename($_FILES[$name]['name']) != $newName)
+					$newName = pathinfo($_FILES[$name]['name'], PATHINFO_FILENAME);
+				elseif (pathinfo($_FILES[$name]['name'], PATHINFO_FILENAME) != $newName)
 					fatal_lang_error('smileys_upload_error_name');
 			}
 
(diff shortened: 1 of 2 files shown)


Sources/ManageSmileys.php
```diff
@@ -702,14 +702,15 @@ function AddSmiley()
...
+					$newName = pathinfo($_FILES[$name]['name'], PATHINFO_FILENAME);
+				elseif (pathinfo($_FILES[$name]['name'], PATHINFO_FILENAME) != $newName)
 					fatal_lang_error('smileys_upload_error_name');
 			}
 
 			foreach ($context['smiley_sets'] as $i => $set)
 			{
-				$set['name'] = un_htmlspecialchars($set['name']);
+				// Name must be treated differently; it's used within a variable name & php converted spaces & periods to underscores
+				$set['name'] = strtr(un_htmlspecialchars($set['name']), ' .', '__');
```
Sesquipedalian: I'm not sure why this change is necessary. A variable name (e.g. `$foo`) can't contain spaces or dots, true, but the index keys of an array certainly can. For example, this is perfectly valid: `$foo['I am the very model of a modern Major-General'] = "I've information vegetable, animal, and mineral";`. Since the names of the individual smiley files are not written into our code as variables, I don't think you need to do this.

Also, if you do this, how will you handle converting these back to the actual names when required?

Sesquipedalian:
> My question is - is there a php function to do this? I couldn't find one.

I don't know of one. But this would get them all: 
```
$set['name'] = preg_replace('~[\x20\x2e\x5b\x80-\x9f]~', '_', un_htmlspecialchars($set['name']));
```

sbulen:
Tested & put out the more robust filter.

sbulen:
Closing this as it is superseded by #5342

PR #5342, opened 2019-01-13
Implements an improved system to handle smileys with varying filenames & extensions

### LINKED ISSUE #5328: Upload new smiley does not work

### LINKED ISSUE #5330: Smiley list doesn't use correct sorting
The first sort type says "Default", but all it does is sort by the IDs. Clicking it does nothing.

### LINKED ISSUE #5331: Cannot upload smileys with different extensions

### LINKED ISSUE #5336: Improve custom smiley handling & upgrading
#### Description
This discussion is spawned from #5331 , where @Gwenwyfar pointed out that multi-extension support exists in 2.0 - but for custom smileys only.  Only the default smileys are stuck on .gif in 2.0.  

And...  Given the current 2.1 design assumes one extension per set, this can pose a problem upon 2.1 upgrade:
image

In effect, 2.0 operates in a "hybrid" mode, where the default smileys are stuck at .gifs across all sets, but custom smileys can be in other formats, jpg, png, gif, etc.  

### Proposal
Use a "hybrid" approach in 2.1, treating custom & default smileys differently...   Do this by:
 - Stripping extensions for _default smileys only_ throughout the app, including the upgrader
 - Dynamically applying extensions to the _default smileys only_
 - Storing full filenames, _extensions included,_ for custom smileys; the same name & extension must be used across sets

This will be complex, but doable.  

Benefits:
 - Fixes the upgrade bug noted above
 - We can restore the quirky but functional 2.0 "copy this file across sets" function
 - No DB changes needed

### Steps to reproduce the bug above
1. Use a custom smiley with a non-.gif image in a 2.0 environment
2. Upgrade to 2.1

_If we agree on this "hybrid" approach, we likely need to revert #5324 , or early RC adopters will lose the extensions on their custom smileys._  

Thoughts, @Sesquipedalian ?

Sesquipedalian:
Tracks smiley filename info in a new smiley_files table. In day-to-day functionality (i.e. the editor and smiley parser) this required little more than changing the queries. The installer and upgrader were a little more involved, but not too big a deal. ManageSmileys has had a more significant number of changes, not only to make it work with the new smileys filename system, but also to fix a number of reported and unreported bugs I encountered in the process.

Fixes #5328
Fixes #5330 
Fixes #5331
Fixes #5336
Closes #5332

For those who are curious, I initially tried out the JSON based idea that I discussed in #5336, and it worked fine. But by the time I was done updating ManageSmileys, I figured that if we were going to do everything that was necessary to fix this set of problems, we might as well do it right the first time. So I went with the database table solution after all.


albertlast:
Did you know why the filename length 48 is?
The number sound a little bit random.

other/upgrade_2-1_postgresql.sql
```diff
@@ -2591,46 +2591,30 @@ CREATE INDEX {$db_prefix}messages_likes ON {$db_prefix}messages (likes DESC);
...
-			('smiley_sets_exts', '')");
----}
+---# Adding the new `smiley_files` table
+CREATE TABLE IF NOT EXISTS {$db_prefix}smiley_files
+(
+	id_smiley smallint NOT NULL DEFAULT '0',
+	smiley_set varchar(48) NOT NULL DEFAULT '',
+	filename varchar(48) NOT NULL DEFAULT '',
+	PRIMARY KEY (id_smiley, smiley_set, filename)
+) ENGINE=MyISAM;
```
albertlast: Postgres doesn't got engine

Sesquipedalian:
I have no idea.

other/install_2-1_postgresql.sql
```diff
@@ -1516,14 +1516,25 @@ CREATE SEQUENCE {$db_prefix}smileys_seq;
...
+#
+# Table structure for table `smiley_files`
+#
+
+CREATE TABLE {$db_prefix}smiley_files
+(
+		id_smiley SMALLINT NOT NULL DEFAULT '0',
+		smiley_set VARCHAR(48) NOT NULL DEFAULT '',
+		filename VARCHAR(48) NOT NULL DEFAULT '',
+		PRIMARY KEY (id_smiley, smiley_set, filename)
```
albertlast: Based on the structur id_smiley and smiley_set should be pk,
you idea could it be to use a index only scan?

Sources/Subs-Editor.php
```diff
@@ -1852,110 +1830,29 @@ function create_control_richedit($editorOptions)
...
 				$request = $smcFunc['db_query']('', '
-					SELECT code, filename, description, smiley_row, hidden
-					FROM {db_prefix}smileys
-					WHERE hidden IN (0, 2)
-					ORDER BY smiley_row, smiley_order',
+					SELECT s.code, f.filename, s.description, s.smiley_row, s.hidden
+					FROM {db_prefix}smileys AS s
+						JOIN {db_prefix}smiley_files AS f ON (s.id_smiley = f.id_smiley)
+					WHERE s.hidden IN (0, 2)
+						AND f.smiley_set = {string:smiley_set}' . (empty($modSettings['smiley_enable']) ? '
```
albertlast, line 1845: Personaly (ther exists not wrong or right) would move ``f.smiley_set = {string:smiley_set}`` to the  on ( ) statement part,
since the outcome is the same ... is that ony a personal thing.

Sources/Subs.php
```diff
@@ -2951,49 +2951,39 @@ function parsesmileys(&$message)
...
+		// Load the smileys in reverse order by length so they don't get parsed incorrectly.
+		if (($temp = cache_get_data('parsing_smileys_' . $user_info['smiley_set'], $cache_time)) == null)
 		{
-			// Load the smileys in reverse order by length so they don't get parsed wrong.
-			if (($temp = cache_get_data('parsing_smileys', 480)) == null)
+			$result = $smcFunc['db_query']('', '
+				SELECT s.code, f.filename, s.description
+				FROM {db_prefix}smileys AS s
+					JOIN {db_prefix}smiley_files AS f ON (s.id_smiley = f.id_smiley)
+				WHERE f.smiley_set = {string:smiley_set}' . (empty($modSettings['smiley_enable']) ? '
```
albertlast, line 2964: Same here `` f.smiley_set = {string:smiley_set}`` to the ON ( ) part.

Sources/Subs.php
```diff
@@ -2951,49 +2951,39 @@ function parsesmileys(&$message)
...
+					AND s.code IN {array_string:default_codes}' : '') . '
+				ORDER BY LENGTH(s.code) DESC',
```
albertlast, line 2966: When you got "many" smileys the opartion length() is very pricy,
personal i see no gain to use it here,
recommandation is to do only a ``order by s.code``

other/upgrade_2-1_postgresql.sql
```diff
@@ -2681,6 +2694,12 @@ if (!array_key_exists($modSettings['smiley_sets_default'], $filtered))
 ---}
 ---#
 
+---# Drop obsolete filename column from smileys table
+ALTER TABLE {$db_prefix}smileys
+DROP COLUMN filename;
```
albertlast: i would like to see here ``DROP COLUMN IF EXISTS filename;``

Sources/Subs-Editor.php
```diff
@@ -134,52 +134,30 @@ function html_to_bbc($text)
...
-		if (empty($modSettings['smiley_enable']))
-		{
-			$smileysfrom = array('>:D', ':D', '::)', '>:(', ':)', ';)', ';D', ':(', ':o', '8)', ':P', '???', ':-[', ':-X', ':-*', ':\'(', ':-\\', '^-^', 'O0', 'C:-)', '0:)');
-			$smileysto = array('evil.png', 'cheesy.png', 'rolleyes.png', 'angry.png', 'smiley.png', 'wink.png', 'grin.png', 'sad.png', 'shocked.png', 'cool.png', 'tongue.png', 'huh.png', 'embarrassed.png', 'lipsrsealed.png', 'kiss.png', 'cry.png', 'undecided.png', 'azn.png', 'afro.png', 'police.png', 'angel.png');
+		// Get all our actual smiley codes
+		$request = $smcFunc['db_query']('', '
+			SELECT code
+			FROM {db_prefix}smileys
+			WHERE code IN ({array_string:smiley_codes})
+			ORDER BY LENGTH(code) DESC',
```
albertlast, line 145: Like i said on different place,
i see not the need for using length here.

Sources/Profile-Modify.php
```diff
@@ -458,11 +458,26 @@ function $preload:		A function that is used to load data required for this eleme
...
+					FROM {db_prefix}smiley_files AS f
+						JOIN {db_prefix}smileys AS s ON (s.id_smiley = f.id_smiley)
+					WHERE s.code = {string:smiley}',
+					array(
+						'smiley' => ':)',
+					)
+				);
+				if ($smcFunc['db_num_rows']($result) !== 0)
+					list ($filenames) = $smcFunc['db_fetch_row']($result);
+				$smcFunc['db_free_result']($result);
```
albertlast, line 479: I don't get fully the idea here,
because you "search" for ``:)`` and return the filename --> okay
but the different smiley_set could got different filename for this.
And you choose "randomly" the first and drop all other filenames.

albertlast (changes requested):
Well not all are "request for changes"

Sources/Profile-Modify.php
albertlast: When this is technical correct,
than a limit 1 would be nice.

Sesquipedalian: That was just a plain old mistake on my part. Good catch.

other/upgrade_2-1_postgresql.sql
Sesquipedalian: Good idea.

Sources/Subs.php
Sesquipedalian: It doesn't matter for the SQL itself, but in the following code it is very important that we work through the codes from longest to shortest. So if we didn't sort them by length in the query, we'd have to sort them by length in PHP anyway.

other/install_2-1_postgresql.sql
Sesquipedalian: If you think that only id_smiley and smiley_set need to be primary keys, I believe you.

other/upgrade_2-1_postgresql.sql
Sesquipedalian: Oops. Good catch.

other/install_2-1_postgresql.sql
albertlast: Well yes only this two are pk.

Sometimes in the rdbms world you try to enable index only scan: https://stackoverflow.com/questions/36533644/analog-of-index-only-scans-from-postgresql-in-mysql
for this you add first the "important" column and after this you add optional column to enable the rdbms to use index only scan.

other/upgrade_2-1_mysql.sql
```diff
@@ -2647,7 +2647,7 @@ if (!array_key_exists($modSettings['smiley_sets_default'], $filtered))
 
 ---# Drop obsolete filename column from smileys table
 ALTER TABLE {$db_prefix}smileys
-DROP COLUMN filename;
+DROP COLUMN IF EXISTS filename;
```
albertlast: Mariadb support this with 10.0 https://stackoverflow.com/questions/173814/using-alter-to-drop-a-column-if-it-exists-in-mysql
and mysql 8.0 not even now -> don't do it in mysql upgrade

other/install_2-1_postgresql.sql
Sesquipedalian: Thanks for that link. Always good to learn something new. 🙂

Sources/Subs-Editor.php
Sesquipedalian: I wondered a little about that when I was writing it. I chose to do it in the WHERE simply because that is what I see most commonly in existing SMF code, but my first instinct was to do it your way. Now I'll know for the future that there's no reason not to do that. 🙂

other/upgrade_2-1_mysql.sql
Sesquipedalian: Thanks.

Sources/Subs-Editor.php
albertlast: Well it's depent,
my awnser would be complete wrong if you use a left join,
than WHERE and ON are completly different.

Only when you do a normal join than this comment is right from me.

albertlast:
Just for my intrest how do the code handle case when a smiley is used which doesn't exists in the current smiley set?
Exp. smiley ``:)`` didn't exists in my own smiley set 'asdf',
what happen then?

Sources/ManageSmileys.php
```diff
@@ -346,15 +344,20 @@ function EditSmileySets()
 				// Exclude the smileys that are already in the database.
 				$request = $smcFunc['db_query']('', '
 					SELECT filename
-					FROM {db_prefix}smileys
-					WHERE filename IN ({array_string:smiley_list})',
+					FROM {db_prefix}smiley_files
+					WHERE filename IN ({array_string:smiley_list})
```
albertlast, line 363: This query doesn't take the pk index,
when you want to change this,
you could add `` AND id_smiley <> ''`` this should enable mysql and pg to use the pk index.

Sources/ManageSmileys.php
```diff
@@ -1696,28 +1838,34 @@ function ImportSmileys($smileyPath)
...
 	}
 	$dir->close();
 
 	// Exclude the smileys that are already in the database.
 	$request = $smcFunc['db_query']('', '
 		SELECT filename
-		FROM {db_prefix}smileys
-		WHERE filename IN ({array_string:smiley_list})',
+		FROM {db_prefix}smiley_files
+		WHERE filename IN ({array_string:smiley_list})
```
albertlast: here again when you want to use the pk index,
than you need to add id_smiley to the where statement part.

Sources/ManageSmileys.php
albertlast: okay i'm wrong even with the fake where statement the index can't be used

Sources/ManageSmileys.php
albertlast: okay i'm wrong even with the fake where statement the index can't be used

Sources/ManageSmileys.php
Sesquipedalian: So, do you think I should make filename a primary key again after all?

Sesquipedalian:
> Just for my intrest how do the code handle case when a smiley is used which doesn't exists in the current smiley set?
> Exp. smiley `:)` didn't exists in my own smiley set 'asdf',
> what happen then?

Before 049d599, it didn't handle it well at all.

Sesquipedalian:
Hey @sbulen, would you mind testing this PR with your awesome suite of test environments? If it works for you as well as it has for me, I'll be very happy to merge this thing and kill the last known bug holding up RC1.

EDIT: To be clear, I'm particularly hoping that you can conduct those thorough tests of yours on the install and upgrade processes in this PR. As long as the database changes are carried out flawlessly, that'll be good enough for RC1; we can fix any remaining functionality bugs in the PHP code later.

Sources/ManageSmileys.php
albertlast: This wouldn't change the result.

The problem is that the id column is not included in where part -> pk index can't be used.
But I don't see any workaround to solve this in that case here.

But maybe it's not a problem to got here no index support.

Sesquipedalian: > But maybe it's not a problem to got here no index support.

This function is rarely used, so it is probably not a problem worth worrying about, at any rate.

albertlast:
Since pg 11 you can add optional column to an index,
when i find time, i maybe create a pr with add some check in upgrade and install code if pg >= 11 and when yes,
i would change the pk index in the format that filename is added as optional col.

see https://www.postgresql.org/docs/11/sql-createindex.html "include"

in my eyes not important thing, i only see here the first the chance to use it :)

other/upgrade_2-1_postgresql.sql
```diff
@@ -2654,22 +2641,64 @@ upgrade_query("
...
-
-upgrade_query("
-	UPDATE {$db_prefix}settings
-	SET value = '" . $smcFunc['db_escape_string'](implode("\n", array_column($filtered, 0))) . "'
+	SET value = '" . $smcFunc['db_escape_string'](implode("\n", $filtered)) . "'
 	WHERE variable = 'smiley_sets_names'");
 
+// Does the filename column exist? (Need to check in case the upgrader has been run before)
+$request = $smcFunc['db_query']('', '
+	SHOW COLUMNS
```
albertlast: Without testing i'm very sure that this command didn't work in pg,
but why you don't use db_list_columns https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/DbPackages-postgresql.php#L36 ?

Sesquipedalian: > but why you don't use db_list_columns

Because at 1:00 AM my intellect was still relatively sharp, but my memory was clearly not. 😆

Thanks for the much better suggestion.

Sesquipedalian: By the way, I'm really glad to have you around, @albertlast. As you've probably noticed, my knowledge of SQL is basic, so your expertise is much appreciated.

sbulen:
First test was a new install, and I'm getting "Sorry, Guest, you are banned from using this forum"

I took current GH & applied 5342... This is what I get when I try to login my newly created forum. 

image

sbulen:
~~I am assuming the above is not related to this PR, so logged issue #5347 .~~

EDIT: I now think the above issue is due to this PR.

sbulen:
More luck with the upgrader, but I'm getting a lot of these errors, just navigating around the forum:
image

That line of code is in parsesmileys().

sbulen:
On an upgraded DB, it's not recognizing any smiley sets on the Edit Smileys form or the Set Smiley Order form:
image
image

The new table is empty:
image

And there is no filename in the smileys table anymore, so there are no filenames anywhere???:
image

sbulen:
My first suspicion is that the upgrader doesn't have access to db_list_columns....

sbulen:
I'm pretty sure db_list_columns is a problem.   You'll need to require DbPackages-xxxx.php in the upgrader.

Next problem - if you rerun the upgrader, you kill your settings.  Since the logic builds up from the filename column, then deletes the filename column, you cannot rerun it...   There needs to be a check in the upgrader to avoid rerunning the logic.

sbulen:
Typo: Filenames instead of filename when populating the smileys table...  (the in_array check)
Took a while to find that one...

sbulen:
Once you're past the above two issues (db_list_columns & the typo) the upgrader issues listed above go away & you can test it for real, it looks **_EXCELLENT_**....   You fixed those 2.0 add smiley bugs...  This is much better than what is out there today.  

I've only found 2 issues:
1)  Adding an existing smiley doesn't seem to work - it does nothing.  For my test, I chose one of the blanks, blank.gif under one of the 2.0 sets.   (in 2.0, it would copy what it could without overwriting anything, & then you could then use the blank smiley)
2)  I'm getting a lot of array to string conversion errors:
image

sbulen:
Some final thoughts for today:
 - I think the install issue above is due to this PR somehow...  I don't get it on a plain GH download.  I only get it after applying this PR.  
 - I have only tested on WAMP today.  When we get past some of the above issues I'll retest on WAMP & on linux.

sbulen:
I have confirmed that my linux environment has the same installer issue as my WAMP environment that even as Guest I am banned & cannot logon.  

I am fairly certain that the error message is coming from here:
https://github.com/SimpleMachines/SMF2.1/blob/bfe39eb474316c14dc923b1ee7cd66d57de084fa/Themes/default/scripts/script.js#L338

The banned message is really erroneous...  It's reporting an http 500 while trying to execute the ajax call in reqOverlayDiv when painting the login form.  

I believe the source of the http 500 is an improperly loaded template.  This error message should be improved - it's got nothing to do with a ban.  

To get to true root cause, you gotta figure out why the template isn't loading...

sbulen:
On a newly installed forum, smiley_enable is not set.  When it is not set, the sql in parsesmileys doesn't work - it's not escaped properly.  This is the code ~ line 2957 in subs.php, the query under "// Load the smileys in reverse order".   Even on an upgraded forum, if you set smiley_enable to 0 or delete it, you will get this error.  It causes the template header load to fail.  

This shows the stack of the failure & the code in question:
image

sbulen:
The question in my mind about the above is WHY it's not escaped properly.  I thought db_query did that for you...   I think the proper fix would be to db_query...

albertlast:
What is the error message of the query?

Sesquipedalian:
> My first suspicion is that the upgrader doesn't have access to db_list_columns....

> Typo: Filenames instead of filename when populating the smileys table... (the in_array check)
> Took a while to find that one...

Well, db_list_columns is used a number of times in the upgrader files, but that typo seems to explain the issue quite well. Should be fixed in ba5c9bf.

> On a newly installed forum, smiley_enable is not set. When it is not set, the sql in parsesmileys doesn't work - it's not escaped properly. This is the code ~ line 2957 in subs.php, the query under "// Load the smileys in reverse order". Even on an upgraded forum, if you set smiley_enable to 0 or delete it, you will get this error. It causes the template header load to fail.
> 
> This shows the stack of the failure & the code in question:
> image

I forgot to put parentheses around the array of strings in the query, causing a syntax error. Should be fixed in 1ace7c8.

sbulen:
Excellent!   Good to see you back!

Got much further this time around.  

Went to post...  Hadn't enabled anything yet...   Just got this, whole new build using the lastest & greatest:
image

Same thing, missing parens...

sbulen:
Trying to add a .jpg smiley, got this:
image

The add smiley issue using existing file issue listed above still exists.  Nothing happens.

sbulen:
Still seeing these, possibly related to the issues noted above, but maybe not:
image

sbulen:
Ah....   It's the other way around.  The above messages (array to string conversion) appear when you _successfully_ add a smiley.  They don't seem to cause any issues, the add appears successful.

Sesquipedalian:
> 1. Adding an existing smiley doesn't seem to work - it does nothing.  For my test, I chose one of the blanks, blank.gif under one of the 2.0 sets.   (in 2.0, it would copy what it could without overwriting anything, & then you could then use the blank smiley)

That'll be because the code explicitly ignores blank.* files. Try with anything else and it works.

> 2. I'm getting a lot of array to string conversion errors:

Fixed in bc0005d

sbulen:
On an upgrade, my 2.0 .jpg custom icon wasn't honored for some reason:
image

sbulen:
Oh, why'd you exclude blank?   I like that dude.  I feel like that sometimes...

sbulen:
Nevermind on the "missing" .jpg.   That was because the default in 2.0 was 'default', so it became fugue, and fugue won't have the custom smileys in it upon upgrade.  

I guess folks who have a lot of custom smileys will need to copy the custom smileys to the new sets' folders.

sbulen:
OK, on both WAMP & linux, I have done installs & upgrades, and the only issues found were all noted above.  I see more fixes came in during my testing - those were not included.  

I performed two variants of beta-to-beta upgrades, a pre-5342 to post-5342 and a post-5342 to a post-5342.  Post-to-post was fine.   

Pre-to-post had issues - it converted everything to .png, including all my old .gif smileys???  This may be disruptive, as I suspect it will be very common.   I tested this because I know a lot of the team does this.  

In short, I think everything looks great other than beta-to-betas, which will be a problem for some.  

I will repeat the tests again Saturday _(not tomorrow - can't miss the Friday night jam!)_, to include all of tonight's patches, too.   

The only tests I couldn't get to are based on my production 2.0 => 2.1 upgrades.  Those take hours, I'll do them over the weekend.  Since my forum has been patched since 1.1, I sometimes find issues there that a "younger" SMF forum won't experience.

sbulen:
Some examples of the beta-to-beta issue described above.   I find the results very confusing...
image

This is what it looks like when you modify one of the smileys:
image

Sesquipedalian:
> Some examples of the beta-to-beta issue described above. I find the results very confusing...

That's a fine example of why upgrading from beta to beta is unsupported. But while looking again at that bit of code I saw a more robust way to handle the supported scenario, and in the process happened to solve the problem for people who upgrade from beta to beta anyway.

albertlast:
With the commit e22f79f on a fresh install i get this in my pg env:
grafik

grafik

albertlast:
Okay for $reason the smiley_files is empty after install.

other/install.php
```diff
@@ -1372,6 +1372,54 @@ function DatabasePopulation()
...
+		foreach ($smiley_set_extensions as $set => $ext)
+			$smiley_inserts[] = array($row['id_smiley'], $set, $smiley_filenames[$row['code']] . $ext);
+	}
+	$smcFunc['db_free_result']($request);
+
+	$smcFunc['db_insert']('ignore',
+		'{db_prefix}smiley_files',
+		array('id_smiley' => 'int', 'smiley_set' => 'string-48', 'filename' => 'string-48'),
+		$smiley_inserts,
+		array('id_smiley', 'smiley_set', 'filename')
```
albertlast: Throw 'filename' away (since it's not pk anymore) and
the pg install works fine.

Sesquipedalian: Thanks. f29aa8b should fix this one.

Sources/ManageSmileys.php
```diff
@@ -771,19 +819,39 @@ function AddSmiley()
...
+
+		$smcFunc['db_insert']('ignore',
+			'{db_prefix}smiley_files',
 			array(
-				$_POST['smiley_code'], $_POST['smiley_filename'], $_POST['smiley_description'], $_POST['smiley_location'], $smiley_order,
+				'id_smiley' => 'int', 'smiley_set' => 'string-48', 'filename' => 'string-48',
 			),
-			array('id_smiley')
+			$inserts,
+			array('id_smiley', 'smiley_set', 'filename')
```
albertlast: Wrong pk filename

Sources/ManageSmileys.php
```diff
@@ -929,35 +1043,49 @@ function EditSmileys()
...
+				foreach ($filenames as $set => $filename)
+					$inserts[] = array($_POST['smiley'], $set, $filename);
+
+				$smcFunc['db_insert']('replace',
+					'{db_prefix}smiley_files',
+					array(
+						'id_smiley' => 'int', 'smiley_set' => 'string-48', 'filename' => 'string-48',
+					),
+					$inserts,
+					array('id_smiley', 'smiley_set', 'filename')
```
albertlast: Wrong pk

Sources/ManageSmileys.php
```diff
@@ -1736,21 +1891,45 @@ function ImportSmileys($smileyPath)
...
+		foreach ($filename_array as $set => $filename)
+			$inserts[] = array($new_id_smiley, $set, $filename);
+
+		$smcFunc['db_insert']('replace',
+			'{db_prefix}smiley_files',
+			array(
+				'id_smiley' => 'int', 'smiley_set' => 'string-48', 'filename' => 'string-48',
+			),
+			$inserts,
+			array('id_smiley', 'smiley_set', 'filename')
```
albertlast: wrong pk

other/upgrade_2-1_mysql.sql
```diff
@@ -2605,22 +2592,57 @@ upgrade_query("
...
+	}
+	$smcFunc['db_free_result']($request);
+
+	if (!empty($inserts))
+	{
+		$smcFunc['db_insert']('ignore',
+			'{db_prefix}smiley_files',
+			array('id_smiley' => 'int', 'smiley_set' => 'string-48', 'filename' => 'string-48'),
+			$inserts,
+			array('id_smiley', 'smiley_set', 'filename')
```
albertlast: wrong pk

other/upgrade_2-1_postgresql.sql
```diff
@@ -2654,22 +2641,57 @@ upgrade_query("
...
+	}
+	$smcFunc['db_free_result']($request);
+
+	if (!empty($inserts))
+	{
+		$smcFunc['db_insert']('ignore',
+			'{db_prefix}smiley_files',
+			array('id_smiley' => 'int', 'smiley_set' => 'string-48', 'filename' => 'string-48'),
+			$inserts,
+			array('id_smiley', 'smiley_set', 'filename')
```
albertlast: wrong pk

albertlast (changes requested):
I hope i found all places

other/install.php
albertlast: yes the install works now

albertlast:
tested pg upgrade from 2.0.15 to 2.1, works fine.

albertlast (approved):

sbulen:
Tested installs on wamp & linux.

Only issue found is that the Add Smiley using an existing file still doesn't work - no matter what file is used.   No response.  

Everything else looks clean, _all_ other fixes appear to be working.  No error messages in the log.  

Looking at upgrades next...

sbulen:
I am having issues with custom smileys after an upgrade.   The Edit Smileys screen shows that files don't exist when they do...  

The .gif ones still work.  Looks like the extensions all got changed to .gif...

I think in 2.0, you can safely assume default smileys are all .gifs, but not custom smileys.  

image

image

sbulen:
I tried to modify one to fix it, and the filenames for that smiley all got changed to erroneous values - these should have been goofysmiley.jpg:
image

Sesquipedalian:
Is that a beta to beta upgrade where you are seeing the incorrect file extensions, @sbulen?

sbulen:
No - a 2.0 with custom smileys, one .gif, one .jpg, & one .png.

sbulen:
The good news is they are all correctable in the UI:
image
image

Sesquipedalian:
Hm.  Can you show me what the table data was for those smileys in the 2.0 table?

sbulen:
Thankfully, yes, saved off:
image

Sesquipedalian:
Thanks.

sbulen:
I'm going to nuke it & start over to double-check everything...

sbulen:
VERY strange...  I just repeated the upgrade & they all got changed to .png this time around, not .gif...   
image

No idea why I got a slightly different result...

sbulen:
Adding a new smiley SET that exists in the filesystem isn't working:
image

image

And you cannot fix it by attempting to modify the smileys:
image

The entries aren't in the smiley_files table:
image

sbulen:
If it's any help, here are the log entries made during the above:
image

sbulen:
My availability is spotty the next few hours - a few errands & a function.  I free up again later tonight (pacific time).  
My office is closed Monday, so I have plenty of availability to retest Sun & Mon.

Sesquipedalian:
I think 7003997 should fix the issue with setting extensions on custom smileys during upgrade. I'll try to deal with the rest tomorrow.

Sesquipedalian:
> VERY strange... I just repeated the upgrade & they all got changed to .png this time around, not .gif...
> image
> 
> No idea why I got a slightly different result...

Regarding this bit, when you select a different set in the dropdown, the filename changes to show the one used for that particular set, just like the preview image itself does. In this screenshot you're viewing Fugue's set (PNG files), whereas in the previous one you were viewing Akyhene's set (GIF files). That's all.

other/upgrade_2-1_postgresql.sql
```diff
@@ -2681,6 +2708,12 @@ if (!array_key_exists($modSettings['smiley_sets_default'], $filtered))
 ---}
 ---#
 
+---# Drop obsolete filename column from smileys table
+ALTER TABLE {$db_prefix}smileys
+DROP COLUMN IF EXISTS filename;
```
sbulen: I'm nervous about dropping the filename column unconditionally here.  Any hiccup above, and this is executed, then the filenames are lost.  I would suggest doing this via db_query within the code block above.

Sesquipedalian:
If this isn't ready, it's darn close. Give it another run, @sbulen.

Sources/ManageSmileys.php
```diff
@@ -1846,35 +1846,130 @@ function ImportSmileys($smileyPath)
...
 
-	// Exclude the smileys that are already in the database.
+	// Get the smileys that are already in the database.
+	$existing_smileys = array();
 	$request = $smcFunc['db_query']('', '
-		SELECT filename
-		FROM {db_prefix}smiley_files
-		WHERE filename IN ({array_string:smiley_list})
-			AND smiley_set = {string:smiley_set}',
+		SELECT *
```
albertlast: Why a star?
One of the db basic stuff,
don't use a star...
https://stackoverflow.com/a/3640051

albertlast (changes requested):

sbulen:
Vanilla 2.1 tests.   Adding a set has issues.   It looks good here:
image

But smiley sets looks like this:
image

Modifying an individual smiley:
image

Smiley order:
image

A sampler of the errors in case it helps:
image

sbulen:
Note that all variants of "Add Smileys" now work (use existing, use one for all sets, use a unique one for each set).

Sesquipedalian:
Ah. That'll be because you didn't import the smiley files when you created the new set. I'll add some checks to the add and edit functions to handle that situation.

In the meantime, can I get you to go to the Smiley Sets page, click the link to modify the Classic set, and then click the link to import the detected files? I couldn't find any more ways to break it on my machines, but maybe you can. 😉

sbulen:
Hmmm....   I always left the import feature alone, because (in prior incarnations) it would literally add smiley codes with strange values...  And I didn't want new made-up codes, I only wanted the set, with the existing codes.   It used to work fine just adding a new set.  

I'll try the import function.

sbulen:
Upgrade isn't working - probably a syntax error in there:
image
****

sbulen:
This is what the import function did, including the strange smiley codes I mentioned earlier.  I think they come from the filenames:
image

sbulen:
Bear in mind the import function won't know what codes you intend for each file...   There is no source for the "code" in the filename or in the define set screen.  That's why it creates new ones...  IIRC, this is 2.0 behavior as well.  

If you want to try to sync it up on import, I think you'll have to do it by filename (sans extension).  But that thinking does not hold anymore, now that we allow different filenames per smiley...  

To be honest, I think we need to get rid of the import function now that we allow different filenames per smiley.   I cannot think of a way to uniquely ID a smiley row to ~~update...~~ connect the filename to...

sbulen:
I was trying to clean this up by deleting the set & starting over, but deleting the set doesn't remove the smileys from the smiley_files table:

"classic" is no longer here:
image

But it's here:
image

Sesquipedalian:
> Bear in mind the import function won't know what codes you intend for each file... There is no source for that info in the filename or in the define set screen. That's why it creates new ones... IIRC, this is 2.0 behavior as well.
> 
> If you want to try to sync it up on import, I think you'll have to do it by filename (sans extension). But that thinking does not hold anymore, now that we allow different filenames per smiley...
> 
> To be honest, I think we need to get rid of the import function now that we allow different filenames per smiley. I cannot think of a way to uniquely ID a smiley row to update...

Actually, if you take a look at the code in ImportSmileys(), you'll see that it now detects likely matches for existing smileys among the imported files by comparing the imported file's name to the names of the files used by other sets. 
https://github.com/SimpleMachines/SMF2.1/blob/94a9f571a9b0a366a93658b7db17e78d7370cebd/Sources/ManageSmileys.php#L1885-L1918

If a match is found, it assigns the file to the existing smiley rather than creating a new smiley. 
https://github.com/SimpleMachines/SMF2.1/blob/94a9f571a9b0a366a93658b7db17e78d7370cebd/Sources/ManageSmileys.php#L1921-L1970

Not only that, but running the import process for one smiley set will also intelligently fix the matching smileys in any *other* smiley sets that are missing files for them.

It's really rather clever, if I do say so myself. :wink: Give it a try.

sbulen:
The upgrader error seems to be due to the insertion of the values.  

My first take is that the "special way for ignore method with returning", ~line 800 in subs-db-mysql, is building a query that looks pretty awkward.  

I can't tell if we're calling it incorrectly, or if this is little-used logic that is wonky...

This is where it's built:
image

So what ends up being passed to db_query is:
image

Sesquipedalian:
> This is what the import function did, including the strange values I mentioned earlier:
> image

Oh, I missed this comment. I can't tell from your screenshot which set that is. Is is the same one you just ran the import function for, or a different one? When I tested just now, I did not see those results. For example, when I duplicate ./Smileys/aaron to ./Smileys/aaron_copy in the filesystem and then create a set for it like this:
screen shot 2019-01-21 at 2 07 04 pm
... everything imports and matches up nicely:
screen shot 2019-01-21 at 2 18 10 pm


Is there anything unusual about the set of files that is giving you trouble? How can I reproduce?

sbulen:
I was getting similar results on all sets in my WAMP environment.   About to retest that function with the new code.

Sesquipedalian:
> The upgrader error seems to be due to the insertion of the values.
> 
> My first take is that the "special way for ignore method with returning", ~line 800 in subs-db-mysql, is building a query that looks pretty awkward.
> 
> I can't tell if we're calling it incorrectly, or if this is little-used logic that is wonky...

@albertlast would be the best one to tell us that. At any rate, since the only problem with the upgrader now is this very thing that is supposed to put up a safeguard if there are problems, I think I should just revert 946491f.

Sesquipedalian:
> I was getting similar results on all sets in my WAMP environment. About to retest that function with the new code.

Yes, please do.

sbulen:
I'm still getting the same thing.  

These are the steps I followed:
 - Copy a set from another forum  (classic, which for some crazy reason, my forum guys like...) 
 - Select "New Smiley Set" from the smiley sets form
 - I selected the import smileys already in this directory checkbox 

First random guess is it's not getting a hit because classic uses .gifs and the other sets are .pngs.  

image

sbulen:
My preference would be to drop the import function, & have the add smiley set function do the adds.  I think educated guesswork here would be OK - try to find one good default in the filesystem per smiley code.  

The import function basically spams your smileys table with a bunch of goofy entries...

Sesquipedalian:
> First random guess is it's not getting a hit because classic uses .gifs and the other sets are .pngs.

Ah. Yeah, that would do it. I've been testing with the Aaron and Akyhne sets installed, so I have GIF files already in my table. Shouldn't take too long to fix.

> The import function basically spams your smileys table with a bunch of goofy entries...

What if it just skipped files for which it didn't find matching smileys, instead of creating new codes for them? That would avoid the problem while still allowing a batch import. Otherwise, manually adding a file for each smiley will become quite a tedious task for the admin.

sbulen:
> What if it just skipped files for which it didn't find matching smileys, instead of creating new codes for them? That would avoid the problem while still allowing a batch import. Otherwise, manually adding a file for each smiley will become quite a tedious task for the admin.

I like the idea of never adding smiley rows on import - so we never see the :xxx: codes.  

But...  At that point, why bother having add set & import two separate functions?

Sesquipedalian:
> But... At that point, why bother having add set & import two separate functions?

A good question...

other/upgrade_2-1_mysql.sql
```diff
@@ -2631,16 +2631,15 @@ if (in_array('filename', $smileys_columns))
 
 	if (!empty($inserts))
 	{
-		$results = $smcFunc['db_insert']('ignore',
+		$smcFunc['db_insert']('ignore',
 			'{db_prefix}smiley_files',
 			array('id_smiley' => 'int', 'smiley_set' => 'string-48', 'filename' => 'string-48'),
 			$inserts,
 			array('id_smiley', 'smiley_set'),
```
albertlast: You miss the comma in both files.

Sesquipedalian: That's fine. PHP is different from SQL in this regard.

sbulen: Which comma?  I'm not seeing it...

sbulen:
I'm still getting this on an upgrade:
image

other/upgrade_2-1_mysql.sql
sbulen: Oh, the EXTRA comma...   Got it...  Yes, that's a problem & causes the issue below.

sbulen:
Note: Logged #5354 so we can come back to that db_insert issue at another time.

other/upgrade_2-1_mysql.sql
sbulen: @Sesquipedalian - This isn't an array, this is a series of parameters in a function call...   If you add that comma, the following ')' is now "unexpected".  I've tested & removing the comma allows the upgrader to proceed.  @albertlast is on target there.

Looks like it will be legal in PHP 7.3...
https://laravel-news.com/php-trailing-commas-functions

Sesquipedalian:
> I'm still getting this on an upgrade:

Bleh. I've changed it back way to the method that is known to work properly.

sbulen:
FYI - it really was just that comma...   albertlast was correct.

sbulen:
I saw the new import options, and chose Import, but I'm still seeing the following.

This was a new vanilla 2.1 install, & I tried to add my classic set:
image

Sesquipedalian:
> I saw the new import options, and chose Import, but I'm still seeing the following.

Oh, that's because af839d6 addresses that problem. 292596e was about creating new sets and  avoiding :xxx: codes.

other/upgrade_2-1_mysql.sql
Sesquipedalian: Fair enough. I'm often wrong, after all. 🙂

Sesquipedalian:
I think importing your Classic set will work now, @sbulen. It worked when I tested under the same conditions here.

sbulen:
OK - the two big issues for the day have been resolved.  I am able to complete an upgrade, and I am able to import a smiley set when no previous set with that extension existed (without spamming the smileys table...).  I have also been able to use a couple variants of Add Smiley.  I've been able to do this on both a vanilla 2.1 install & an upgraded 2.0 => 2.1 forum.   Editor looks good & profile updates look good.  No errors logged whatsoever.   All these tests were on my WAMP server.  

I have not completed the full suite of tests (all the same tests, but with different settings, e.g., custom on & selection off, or, custom off & selection on, etc.).  I haven't tried a beta-to-beta, & I haven't tried anything on linux yet...  I think all that will need to wait until tomorrow night.  I'm tapping out for now...   

Excellent work, Squid!  I admire your stamina, taking it from albertlast & myself for the last week or so.  (We've got a bit of a yin & yang thing going on...)

image

sbulen:
Oh no...  I see more changes came in during the last round of testing...   ACK....   

Again, I'll retest everything tomorrow...  But we're very, very close...

Sesquipedalian:
Excellent. All I need to know now is whether installing and upgrading works in all environments. Any further functionality issues can wait until after RC1.

sbulen:
OK this is where we are at.  I've done installs & upgrades on both WAMP & linux and they all ran well.  Editor & profile changes are perfect throughout.  Add Smileys work, all flavors.  Setting the order works. Pretty much all the screens work.  

The one exception I have is when adding a set.  After doing so, I start getting errors just navigating around the smiley admin screens, in particular, whenever I click to go to Add Smiley.  

These are the errors:
image

The errors don't seem to hurt anything...  

However, the added set, if it didn't have, for example, all the default smileys or all the custom smileys, looks, as expected, like this:
image
(cute, aren't they?)

The only real bug I've had, that's a limitation, is that I cannot fix the 'missing' ones.   I.e., if I click on the Modify link for one of the little guys missing a filename to try to correct it, I get this on both WAMP and Linux:
image

This is only a problem after adding a set, and only when trying to fill in the blanks, if some of the smileys are missing, e.g., if you had added custom smileys.  If the "set" added is "complete", no errors.  

(And yes, I had copied the missing smiley files to the smiley folder via the file system...  I was using the UI to select them to make the set complete....)

sbulen:
(cough, cough...  mulligan...  cough, cough...)

sbulen:
I think I know where the "extension" errors came from - there were stray files in the smiley set folder that I copied over that didn't have an extension...   Once deleted, I stopped getting the errors.

Sesquipedalian:
> OK this is where we are at. I've done installs & upgrades on both WAMP & linux and they all ran well. Editor & profile changes are perfect throughout. Add Smileys work, all flavors. Setting the order works. Pretty much all the screens work.

Hooray!

------------------------------

> The one exception I have is when adding a set. After doing so, I start getting errors just navigating around the smiley admin screens, in particular, whenever I click to go to Add Smiley.
> 
> These are the errors:
> image
> 
> The errors don't seem to hurt anything...

Oh, that's just because I forgot to do an `isset()` check to account for anything in the directory without an extension (or only an extension, like `.htaccess`). That'll be trivial to fix.

---------------------------------

> However, the added set, if it didn't have, for example, all the default smileys or all the custom smileys, looks, as expected, like this:
> image
> (cute, aren't they?)
> 
> The only real bug I've had, that's a limitation, is that I cannot fix the 'missing' ones. I.e., if I click on the Modify link for one of the little guys missing a filename to try to correct it, I get this on both WAMP and Linux:
> image
> 
> This is only a problem after adding a set, and only when trying to fill in the blanks, if some of the smileys are missing, e.g., if you had added custom smileys. If the "set" added is "complete", no errors.
> 
> (And yes, I had copied the missing smiley files to the smiley folder via the file system... I was using the UI to select them to make the set complete....)

Hm. I can't reproduce that. This is what I did when I tried:

1. Duplicated the default version of `./Smileys/aaron` to `./Smileys/aaron_copy`
2. Added a new smiley by uploading a file. (Just to be safe, I made sure it was one that I had never, ever, ever created before for any of these tests.)
3. Created a new smiley set for aaron_copy.
4. Immediately went to the Edit Smileys page and (as expected) saw that the custom smiley was missing from the newly created set:
screen shot 2019-01-22 at 10 25 17 pm
5. Clicked the Modify link for the custom smiley, and saw this:
screen shot 2019-01-22 at 10 25 44 pm

sbulen:
I've never seen a missing file highlighted like that...  What I have seen is that when there is a missing file, it defaults to the first file (afro).   

Tonight, it wasn't even doing that...

I hate to ask, but are you sure that commit is up there?

Sesquipedalian:
Yup. a6c3c10d38

sbulen:
Random thought...   One thing I've done a little different than what you described is that when adding sets I immediately made them the new default...   By checking the little box while adding them...

Sesquipedalian:
> Random thought... One thing I've done a little different that what you described is that when adding sets is I immediately made them the new default...

That was the missing piece. I reproduced it now. Should be a quick fix. Give me a couple minutes.

Sesquipedalian:
Done.

sbulen:
That worked!

sbulen:
Well done, Squid!   

Now I'm off to sleep...

Sesquipedalian:
Very well. Thanks for all your testing and input, @sbulen. Once Travis is done, I'll merge this, and then good things can begin to happen. 😁

PR #5349, opened 2019-01-17
Fix Double space by moving &nbsp; fixes  #5348

### LINKED ISSUE #5348: Adding more than one space adds semicolon ; to end of link.
Adding more than one space adds semicolon to end of link, and breaks the link.  

Example,
Lets add a link, https://cctestsite.info/testsite3/ a space and some more text. The link works fine.

Lets add a link, https://cctestsite.info/testsite3/   a double space and some more text. Now you can see it's added ; to the end of the url, and broken the link.

RC1 a few days old, so this might have been fixed. https://github.com/SimpleMachines/SMF2.1/compare/98d5cfd...release-2.1

albertlast:
Issue  #5348

### DIFF
--- a/Sources/Subs-Post.php
+++ b/Sources/Subs-Post.php
@@ -225,9 +225,9 @@ function preparsecode(&$message, $previewing = false)
 
 	// Restore white space entities
 	if (!$previewing)
-		$message = strtr($message, array('  ' => '&nbsp; ', "\n" => '<br>', $context['utf8'] ? "\xC2\xA0" : "\xA0" => '&nbsp;'));
+		$message = strtr($message, array('  ' => ' &nbsp;', "\n" => '<br>', $context['utf8'] ? "\xC2\xA0" : "\xA0" => '&nbsp;'));
 	else
-		$message = strtr($message, array('  ' => '&nbsp; ', $context['utf8'] ? "\xC2\xA0" : "\xA0" => '&nbsp;'));
+		$message = strtr($message, array('  ' => ' &nbsp;', $context['utf8'] ? "\xC2\xA0" : "\xA0" => '&nbsp;'));
 
 	// Now let's quickly clean up things that will slow our parser (which are common in posted code.)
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
In my tests, only the first of these two commits is necessary. Also, the second commit changes the logic. So unless I am missing something here, the second commit should be reverted.

albertlast:
The second fix the " ;" beahvior when you use the edit (not quick edit)
You need to edit more than once, i guess.

Sesquipedalian:
Now I see, yes.

The trouble is, I'm nervous about changing the behaviour of preparsecode() to address this. Since this only happens when there are multiple spaces after a plain URL, I suspect that the issue is actually a problem with the autolinker regex in parse_bbc().

albertlast:
Feel free to create a own pr,
I don't need to fix this by myself.


Gesendet: Freitag, 18. Januar 2019 08:25
An: SimpleMachines/SMF2.1
Cc: albertlast; Author
Betreff: Re: [SimpleMachines/SMF2.1] Fix Double space by moving &nbsp; fixes #5348 (#5349)


Now I see, yes.

The trouble is, I'm nervous about changing the behaviour of preparsecode() to address this. Since this only happens when there are multiple spaces after a plain URL, I suspect that the issue is actually a problem with the autolinker regex in parse_bbc().

Sesquipedalian:
Yeah, I think that's the real source of the problem. This raw message text:
```
Lets add a link, https://cctestsite.info/testsite3/&nbsp; a double space and some more text
```
is being parsed into 
```
Lets add a link, <a href="https://cctestsite.info/testsite3/&nbsp" class="bbc_link">; a double space and some more text
```
instead of
```
Lets add a link, <a href="https://cctestsite.info/testsite3/" class="bbc_link">&nbsp; a double space and some more text
```

Apparently I didn't think of this possibility when I created the new autolinker regex. Give me a bit, and I think I can fix it.

PR #5351, opened 2019-01-19
Implements permissions for BBCodes

Sesquipedalian:
- To impose permissions restrictions for a given BBC, the BBC tag simply needs to be appended to the $context['restricted_bbc'] array via the integrate_pre_load hook.
- The admin panel UI will show options to grant permission to use the restricted BBC on the main permissions page for each membergroup, as well as showing an inline permissions form for each restricted BBC at the bottom of the BBC settings page.
- The permission to use a restricted BBC is saved in the permissions table as `bbc_*`, where `*` is the tag for the BBC. For example, permission to use SMF's standard [html] BBC is stored as `bbc_html`.
- Mods adding their own restricted BBC do not need to add language strings for use on the permissions forms. Basic strings for that task are taken care of automatically. However, if a mod wants to add a $helptxt string for an individual BBC, it is free to do so.

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -377,7 +377,11 @@ function reloadSettings()
 	$context['legacy_bbc'] = array(
 		'acronym', 'bdo', 'black', 'blue', 'flash', 'ftp', 'glow',
 		'green', 'move', 'red', 'shadow', 'tt', 'white',
+	);
 
+	// Define a list of BBC tags that require permissions to use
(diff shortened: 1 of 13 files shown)


Sesquipedalian:
Also, this fixes some wonky logic in the handling of `$excluded_groups` in `init_inline_permissions()`.

wintstar:
Please do not forget the language strings for the permissions in Reports.english.php.

PR #5355, opened 2019-01-22
Fix the select of Mysql Insert Returning fixes #5354

### LINKED ISSUE #5354: db_insert - "special way for ignore method with returning" issue
#### Description
I think that "special way for ignore method with returning", ~line 800 in subs-db-mysql, is building a query that looks pretty awkward.  Initial suspicion is that this is not used very often.

The WHERE clause is built very strangely:
image

image

Found it while looking into #5342

albertlast:
Now it's work,
please had in mind because mysql is mysql...
this is not realy a fast way.

A normal insert(with ignore) is one query with many values.
In the case of returning(doesn't matter if mode1 or 2) you could got in the worst case
per row two query.

Issue: #5354

maybe the two query things is not needed..
to use returning with ignore or replace means in the common sense that one or more columns you insert,
are pk fields... so i could print them directly out without checking the db response.

Possible more compley would it be when you got on tables unique index,
when you look at smf 2.1 setup file ther are ~11 unqiue,
but the most are miss used only 5 are correctly and this around 4 tables --> not a common use case.

In short,
when you do a insert how is not using replace/ignore we need to look at the ai field on returning,
when replace or ignore is used we can return directly the index value from the $data array.

### DIFF
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -804,7 +804,7 @@ function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $ret
 		$ai = 0;
 		for ($i = 0; $i < $count; $i++)
 		{
-			$old_id = $smcFunc['db_insert_id']();
+			$old_id = $smcFunc['db_insert_id']($table);
 
 			$smcFunc['db_query']('', '
 				' . $queryTitle . ' INTO ' . $table . '(`' . implode('`, `', $indexed_columns) . '`)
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Where do we actually use the return feature, anyway? I assume that you built this so that you could use for something, but I have no idea what that something was.

albertlast:
It's a replacement of 
https://github.com/SimpleMachines/SMF2.1/blob/7bd43ba47756955e6bd7d6767d345c4168501dff/Sources/Subs-Db-postgresql.php#L532

Sesquipedalian:
Included in #5384

albertlast:
Well it's not included

albertlast:
But why close this pr?

Sesquipedalian:
Because I clicked on the wrong link in the PR list and thought I was closing #5384

Sesquipedalian (approved):

PR #5357, opened 2019-01-22
Prevents AJAX calls to fetch alerts from being logged in log_online

### LINKED ISSUE #5356: Who's Online almost always reports the member activity as "Viewing their own profile"
#### Description
See https://www.simplemachines.org/community/index.php?topic=564454.0.

- Version/Git revision: 31ffeb3

Looking at the log_online table, the problem appears to be that AJAX calls to fetch alerts are being logged. This action should be excluded from logging.

Sesquipedalian:
Fixes #5356

### DIFF
--- a/index.php
+++ b/index.php
@@ -196,19 +196,48 @@ function smf_main()
 	if (!empty($topic) && empty($board_info['cur_topic_approved']) && !allowedTo('approve_posts') && ($user_info['id'] != $board_info['cur_topic_starter'] || $user_info['is_guest']))
 		fatal_lang_error('not_a_topic', false);
 
-	// Do some logging, unless this is an attachment, avatar, toggle of editor buttons, theme option, XML feed etc.
+	// Do some logging, unless this is an attachment, avatar, toggle of editor buttons, theme option, XML feed, popup, etc.
 	$no_stat_actions = array(
-		'clock', 'dlattach', 'findmember', 'jsoption', 'likes', 'loadeditorlocale',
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Try now, @smfbigguy.

index.php
```diff
@@ -196,10 +196,19 @@ function smf_main()
...
 	// Do some logging, unless this is an attachment, avatar, toggle of editor buttons, theme option, XML feed etc.
-	if (empty($_REQUEST['action']) || !in_array($_REQUEST['action'], $no_stat_actions))
+	$no_stat_actions = array(
+		'clock', 'dlattach', 'findmember', 'jsoption', 'likes', 'loadeditorlocale',
+		'modifycat', 'requestmembers', 'smstats', 'suggest', 'about:unknown',
+		'.xml', 'xmlhttp', 'verificationcode', 'viewquery', 'viewsmfile',
+	);
+	$no_stat_areas = array(
+		'profile' => 'alerts_popup',
+	);
```
jdarwood007: Might be better as a multi-dimensional array here.

Sesquipedalian: An excellent idea.

smfbigguy:
Well the link you posted lands me here, lol. Yes GH is very weird trying to get around it sometimes, lol. I have applied all these fixes and will have to wait til someone comes online. I hope I got it all right, it's 4:30 in the morning here.

smfbigguy:
I have had two members online so far and no one is viewing their profile. I know it's just two members so I will wait a bit longer but I think this is fixed. :)

PR #5358, opened 2019-01-22
Fixes some logic for deciding what question to print in verification

Sesquipedalian:
Before this, if reCAPTCHA was enabled, visual verification was disabled, and verification questions were required, it was possible for undefined index errors to be generated.


Sesquipedalian:
Since this (1) is small, (2) is simple and obvious, (3) has been tested under all conditions that could affect it, and (4) fixes a (minor) security-related problem, I'm going to go ahead and merge it for RC1 now.

PR #5359, opened 2019-01-24
Remove unneeded col alter of ip in log_online

albertlast:
To alter the column make no sense,
because later on we alter it again to inet and
this alter create a issue when you try to upgrade the upgrade.

In this case pg try to convert inet datatype of ip to bigint which is not possible.

### DIFF
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -2316,25 +2316,6 @@ ALTER TABLE {$db_prefix}member_logins ADD COLUMN ip2 VARBINARY(16);
 ---#
 
 /******************************************************************************/
---- Update log_online ip with ipv6 support without converting
-/******************************************************************************/
----# Delete old column log banned ip
----{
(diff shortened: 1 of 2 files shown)


Sesquipedalian:
Does a similar change need to be made in the MySQL file?

albertlast:
I didn't found this,
Your could look too
If you find the line in mysql

Sesquipedalian:
Looks like it, yes.

First time:
https://github.com/SimpleMachines/SMF2.1/blob/04cffc29c20e5cdde059434cbf2bdeeb9a3ab928/other/upgrade_2-1_mysql.sql#L2257-L2274

Second time:
https://github.com/SimpleMachines/SMF2.1/blob/04cffc29c20e5cdde059434cbf2bdeeb9a3ab928/other/upgrade_2-1_mysql.sql#L2318-L2335

The only difference is the comments.

albertlast:
Well this is something different,
but yes we could remove this duplicate, too.

PR #5364, opened 2019-01-26
Removes grid CSS for dl.settings from install.css

### LINKED ISSUE #5353: Can't upgrade or install using IE11
When trying to install or upgrade 2.1 using IE11, the username and password boxes are merged together in such a way you can't add the details to them.  This only seems to happen in IE11

Two screenshots below, performing an upgrade, one in IE11, the other in Edge which is for comparison only, it's fine in Edge, Firefox, and Chrome.

2 1upgradeie11

2 1upgradeother

Sesquipedalian:
Fixes #5353

### DIFF
--- a/Themes/default/css/install.css
+++ b/Themes/default/css/install.css
@@ -119,25 +119,11 @@ h1.forumtitle {
 	max-height: 8.4em; /* 6 lines of text */
 	line-height: 1.4em;
 }
-dl.settings {
-	display: -ms-grid;
-	display: grid;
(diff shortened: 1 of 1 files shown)


Gwenwyfar:
Looks like although IE  "supports" grid, it puts everything in the first row, and you need to change it manually... So I think this should do, this wasn't much of a special use case either way.

Gwenwyfar (approved):
Looks good enough. We can tweak it later if necessary.

PR #5374, opened 2019-01-28
Avoids breaking stuff when the files and DB have mismatched versions

### LINKED ISSUE #5123: Avatars stored in attachments directory not moved during upgrade
#### Description
In 2.0 and earlier, the default option for avatars is to store them in the attachments directory. This option is no longer available in 2.1, but the avatars are not moved over to the custom_avatars directory during the upgrade, resulting in the avatar not showing for anyone who uploaded one prior to the upgrade.

### Steps to reproduce
1. Upload an avatar in 2.0 with the default settings (store in attachments directory)
2. Upgrade to 2.1

- Database Type: MaraiDB
- Database Version: 10.0.36
- PHP Version: 7.1.23

There is a hidden option in 2.0 (and possibly in 1.1 as well) to move avatars from the attachments directory to a custom avatars directory, so it shouldn't be too hard to resolve this - just port the code over from that.

Sesquipedalian:
Fixes #5123

Previously, if the admin uploaded an upgrade package, but then someone navigated to index.php before the admin finished the upgrade process, various errors and messes would be made. Among other things, `$modSettings['attachmentUploadDir']` could be corrupted, preventing custom avatars from being migrated correctly and generally causing havoc with attachments.

This commit does two things to fix the problem:

1. Adds some sanity checking when trying to json_decode the value of `$modSettings['attachmentUploadDir']` in `reloadSettings()`.
2. Detects when the SMF version defined in the PHP files doesn't match with the value of `$modSettings['smfVersion']` and, if so, redirects to the upgrader (unless the upgrader itself called `reloadSettings()`, in which case no intervention occurs). If upgrade.php doesn't exist in the expected location, dies with a message telling the admin to go get the upgrader package.

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -22,7 +22,7 @@ if (!defined('SMF'))
 function reloadSettings()
 {
 	global $modSettings, $boarddir, $smcFunc, $txt, $db_character_set;
-	global $cache_enable, $sourcedir, $context, $forum_version;
+	global $cache_enable, $sourcedir, $context, $forum_version, $boardurl;
 
 	// Most database systems have not set UTF-8 as their default input charset.
 	if (!empty($db_character_set))
@@ -64,10 +64,26 @@ function reloadSettings()
 
 		// We explicitly do not use $smcFunc['json_decode'] here yet, as $smcFunc is not fully loaded.
 		if (!is_array($modSettings['attachmentUploadDir']))
-			$modSettings['attachmentUploadDir'] = smf_json_decode($modSettings['attachmentUploadDir'], true);
+		{
+			$attachmentUploadDir = smf_json_decode($modSettings['attachmentUploadDir'], true);
+			$modSettings['attachmentUploadDir'] = !empty($attachmentUploadDir) ? $attachmentUploadDir : $modSettings['attachmentUploadDir'];
+		}
 
 		if (!empty($cache_enable))
 			cache_put_data('modSettings', $modSettings, 90);
+	}
+
+	// Going anything further when the files don't match the database can make nasty messes (unless we're actively installing or upgrading)
+	if (!defined('SMF_INSTALLING') && !empty($modSettings['smfVersion']) && version_compare(strtolower(strtr($modSettings['smfVersion'], array('SMF ' => '', ' ' => '.'))), strtolower(strtr($forum_version, array('SMF ' => '', ' ' => '.'))), '!='))
+	{
+		// Wipe the cached $modSettings values so they don't interfere with anything later
+		cache_put_data('modSettings', null);
+
+		// Redirect to the upgrader if we can
+		if (file_exists($boarddir . '/upgrade.php'))
+			header('location: ' . $boardurl . '/upgrade.php');
+
+		die('SMF file version (' . strtr($forum_version, array('SMF ' => '')) . ') does not match SMF database version (' . strtr($modSettings['smfVersion'], array('SMF ' => '')) . ').<br>Run the SMF upgrader to fix this.<br><a href="https://wiki.simplemachines.org/smf/Upgrading">More information</a>.');
 	}
 
 	$modSettings['cache_enable'] = $cache_enable;
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -74,7 +74,7 @@ function reloadSettings()
 	}
 
 	// Going anything further when the files don't match the database can make nasty messes (unless we're actively installing or upgrading)
-	if (!defined('SMF_INSTALLING') && !empty($modSettings['smfVersion']) && version_compare(strtolower(strtr($modSettings['smfVersion'], array('SMF ' => '', ' ' => '.'))), strtolower(strtr($forum_version, array('SMF ' => '', ' ' => '.'))), '!='))
+	if (!defined('SMF_INSTALLING') && (!isset($_REQUEST['action']) || $_REQUEST['action'] !== 'admin' || !isset($_REQUEST['area']) || $_REQUEST['area'] !== 'packages') && !empty($modSettings['smfVersion']) && version_compare(strtolower(strtr($modSettings['smfVersion'], array('SMF ' => '', ' ' => '.'))), strtolower(strtr($forum_version, array('SMF ' => '', ' ' => '.'))), '!='))
 	{
 		// Wipe the cached $modSettings values so they don't interfere with anything later
 		cache_put_data('modSettings', null);
--- a/other/install.php
+++ b/other/install.php
@@ -14,6 +14,7 @@
 define('SMF_VERSION', '2.1 RC1');
 define('DB_SCRIPT_VERSION', '2-1');
 define('SMF_SOFTWARE_YEAR', '2019');
+define('SMF_INSTALLING', 1);
 
 $GLOBALS['required_php_version'] = '5.4.0';
 
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -15,6 +15,7 @@
 define('SMF_VERSION', '2.1 RC1');
 define('SMF_LANG_VERSION', '2.1 RC1');
 define('SMF_SOFTWARE_YEAR', '2019');
+define('SMF_INSTALLING', 1);
 
 /**
  * The minimum required PHP version.


Sesquipedalian:
This appears to work in all my tests, but it'd be nice if @sbulen could run it through his gauntlet.

Sources/Load.php
```diff
@@ -64,12 +64,28 @@ function reloadSettings()
...
+			$modSettings['attachmentUploadDir'] = !empty($attachmentUploadDir) ? $attachmentUploadDir : $modSettings['attachmentUploadDir'];
+		}
 
 		if (!empty($cache_enable))
 			cache_put_data('modSettings', $modSettings, 90);
 	}
 
+	// Going anything further when the files don't match the database can make nasty messes (unless we're actively installing or upgrading)
+	if (!defined('SMF_INSTALLING') && !empty($modSettings['smfVersion']) && version_compare(strtolower(strtr($modSettings['smfVersion'], array('SMF ' => '', ' ' => '.'))), strtolower(strtr($forum_version, array('SMF ' => '', ' ' => '.'))), '!='))
+	{
```
jdarwood007, line 78: How would this work with the ```$forum_verison = '2.1.1';``` and ```$modSettings['smfVersion'] = '2.1.0';```.  Such in the case of upgrading to a minor patch version?

sbulen:
This does not fix #5123...

I'll see if I can duplicate the corruption issue you describe...

Sources/Load.php
Sesquipedalian:
```
<?php

$forum_version = 'SMF 2.1.1';
$modSettings['smfVersion'] = '2.1.0';

if (!empty($modSettings['smfVersion']) && version_compare(strtolower(strtr($modSettings['smfVersion'], array('SMF ' => '', ' ' => '.'))), strtolower(strtr($forum_version, array('SMF ' => '', ' ' => '.'))), '!='))
	echo 'different';
else
	echo 'same';
```

Result: 'different'

Sesquipedalian: So, patches will always need to update both the DB value and the string in index.php if they update either. But that's just good practice anyway.

Sesquipedalian:
> This does not fix #5123...

No? It does for me. 

First, here's how I am able to reproduce the issue consistently. I start off with the simplest configuration for attachment directories in 2.0, with `$modSettings['attachmentUploadDir']` containing a simple path string for a single directory. Then I unpack the upgrade package into the forum's root folder, overwriting the old 2.0 files with the 2.1 files. At this point I check the settings table in the database again just to make sure that `attachmentUploadDir` still contains the same value (a simple path string). Then I go to the forum's index URL in my browser, which spews a bunch of error messages (or rather, it did prior to this PR). Then I check `attachmentUploadDir` in the settings table again. The value has now been changed to `{"1":[]}`. After that, I run the upgrader, and get the results described in #5123.

I've found that if I manually change `{"1":[]}` back to the original path string before running the upgrader, it will handle the avatars and attachments successfully. I've also found that if I reset the database and all files back to their original state and then repeat the complete process above except skipping the step where I went to index.php before going to upgrade.php, then the upgrader will again handle the avatars and attachments successfully.

It is important to reset the settings table to a known good state before each test. (I do it by dropping the whole database and restoring from a backup.) If the setting starts out already corrupted to `{"1":[]}`, there's no way to fix it and the changes in this PR will not prevent the behaviour described in #5123.

To be clear, this PR doesn't fix that corrupted value. It is meant to prevent the corruption from happening.

Sesquipedalian:
The reason the corruption causes the missing avatars is on these lines:
https://github.com/SimpleMachines/SMF2.1/blob/4fa7e97e66008a73a81dff5b5c238510f9b241b6/other/upgrade_2-1_mysql.sql#L342-L356

Because `$currentFolder` will be empty when `$modSettings['attachmentUploadDir']` is set to `{"1":[]}`, the result of `file_exists($oldFile)` will always be false, and so the file is skipped instead of moved.

Sesquipedalian:
I just tested again, this time starting with multiple attachment directories configured in the 2.0 forum prior to unpacking the upgrade package. That worked too with this PR. All avatars were migrated to the custom_avatar folder correctly and appear as expected in 2.1.

sbulen:
I have, in fact, seen the {"1":[]} issue before & was never able to reproduce it on demand.  I'm trying to duplicate it how you described reported above.  

My point is that it is not the same issue reported in #5123 .   It's a different issue.  (And in terms of rollout, I am much more concerned with #5123.)

Testing...

jdarwood007:
> The reason the corruption causes the missing avatars is on these lines:
> [SMF2.1/other/upgrade_2-1_mysql.sql](https://github.com/SimpleMachines/SMF2.1/blob/4fa7e97e66008a73a81dff5b5c238510f9b241b6/other/upgrade_2-1_mysql.sql#L342-L356)
> 
> Lines 342 to 356 in [4fa7e97](/SimpleMachines/SMF2.1/commit/4fa7e97e66008a73a81dff5b5c238510f9b241b6)
> 
>  else 
>  { 
>  	$oldFile = $currentFolder . '/' . $row['id_attach'] . '_' . $row['file_hash']; 
>  	$newFile = $currentFolder . '/' . $row['id_attach'] . '_' . $row['file_hash'] .'.dat'; 
>   
>  	// Make sure it exists... 
>  	if (!file_exists($oldFile)) 
>  		$oldFile = false; 
>  } 
>   
>  if (!$oldFile) 
>  { 
>  	// Existing attachment could not be found. Just skip it... 
>  	continue; 
>  } 
> Because `$currentFolder` will be empty when `$modSettings['attachmentUploadDir']` is set to `{"1":[]}`, the result of `file_exists($oldFile)` will always be false, and so the file is skipped instead of moved.

This shouldn't trigger because https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/other/upgrade_2-1_mysql.sql#L279 says to only do it if json isn't done yet.

Also hate the ugly code that is attempting to cleanup the custom avatars folder.  Not sure why this had to change.

sbulen:
This is what I have confirmed:

- I can reproduce  the {"1":[]} issue following the steps Sesquipedelian outlined.  
- The fixes in #5374 fix it, mainly by preventing the corruption to begin with.  The corruption isn't caused by upgrade.php, it's caused by index.php... 
- I have confirmed installs & upgrades, in WAMP & in linux, work great & I can no longer reproduce the {"1":[]} issue in either environment.  

I think this PR can move forward.  What I will do next is attempt to document the original intent of #5123 better...

Sesquipedalian:
If there is another way to trigger #5123, please do let me know. The corruption of `$modSettings['attachmentUploadDir']` to `{"1":[]}` is the only way I have discovered.

Sesquipedalian:
> This shouldn't trigger because https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/other/upgrade_2-1_mysql.sql#L279 says to only do it if json isn't done yet.

As @sbulen rightly noted, the corruption is actually caused by index.php rather than upgrade.php. That's because `reloadSettings()` would blindly run `smf_json__decode()` on `$modSettings['attachmentUploadDir']` if it wasn't an array, assuming that if it could only either be an array or a JSON string. In a situation where the PHP files have been upgraded to 2.1, but the database hasn't been upgraded yet, that assumption is false.

> Also hate the ugly code that is attempting to cleanup the custom avatars folder. Not sure why this had to change.

Do you mean [this](https://github.com/SimpleMachines/SMF2.1/pull/5374/files#diff-e44bf29380cd171e044f5bf3ac967c8dR69)? If so, that's explained in the paragraph above. If you actually mean the chunk starting [here](https://github.com/SimpleMachines/SMF2.1/pull/5374/files#diff-e44bf29380cd171e044f5bf3ac967c8dR76), which is uglier than the other by far, that doesn't do anything to clean up the avatars or attachments. It just aborts execution if a certain No Good, Very Bad mess is detected so that we don't do any damage.

EDIT: Or are you referring to this, which attempts to fix `$modSettings['attachmentUploadDir']` on the fly?
https://github.com/SimpleMachines/SMF2.1/blob/4fa7e97e66008a73a81dff5b5c238510f9b241b6/Sources/Load.php#L289-L298
If so, I didn't change that, although I actually think it shouldn't be done here at all. As the comment says, this should be done in the upgrader, not on the fly. If the consensus is that I should remove it from here and put it into the upgrader where it belongs, I'll be more than happy to do that.

Sesquipedalian:
Note to self: In SMF 3.0, this whole kludge of a system should be torn down and replaced with something to handle attachment directories more elegantly and efficiently.

sbulen:
OK...   Short version - I have no issues anymore.  

The initial intent of #5123 was focused on problems with avatars not displaying properly because they were still in the attachments directory or a custom avatar directory not named '/custom_avatar'. 

The upgrader has logic to deal with that, and in my testing tonight, this logic worked perfectly in both linux & wamp.  It even properly deals with alternate namings of the custom avatar folder.  I even tested a mix, with some custom avatars in the attachments folder & some in the "avs" folder, and it properly dealt with that as well.  

(When I first attempted 2.1 upgrades in 2017, it was a completely different matter, custom avatars were a mess for me, and that's what was in the back of my mind.  But I have spent a lot of time fixing issues with _my forum_, including the oddity that someone had configured avatars & custom avatars to use the same folder years back (which I had to painfully correct by hand)...  In hindsight, I now suspect that was the root of my old issues...  I'm happy to report it all works now...)

jdarwood007:
> Do you mean [this](https://github.com/SimpleMachines/SMF2.1/pull/5374/files#diff-e44bf29380cd171e044f5bf3ac967c8dR69)? If so, that's explained in the paragraph above. If you actually mean the chunk starting [here](https://github.com/SimpleMachines/SMF2.1/pull/5374/files#diff-e44bf29380cd171e044f5bf3ac967c8dR76), which is uglier than the other by far, that doesn't do anything to clean up the avatars or attachments. It just aborts execution if a certain No Good, Very Bad mess is detected so that we don't do any damage.
> 
> EDIT: Or are you referring to this, which attempts to fix `$modSettings['attachmentUploadDir']` on the fly?

Looking more at the code in the upgrader which is moving custom avatars directory around and trying to cleanup the directory.  Just seems like things could go wrong or expose files if its not truely empty.

Sesquipedalian:
Okay, now I understand. We should take a look at that later.

PR #5376, opened 2019-01-29
Uses constants instead of variables for forum version and year

Sesquipedalian:
For the sake of improving developer sanity and making some things a bit simpler, this replaces variables like `$forum_version` and `$software_year` with some nice constants (specifically, `SMF_VERSION`, `SMF_FULL_VERSION`, and `SMF_SOFTWARE_YEAR`).

Given that these values are supposed to be constant, it makes sense to define them as constants. Not only does this let us eliminate silliness like `strtr($forum_version, array('SMF ' => ''))` from numerous places, but it also lets us get rid of a few silly workarounds that had to be put in place just in case a mod had clobbered the variables.

These changes do require an update to some of our buildTools scripts, so this PR will fail Travis and Scrutinizer for now. I've already written the changes for buildTools, but I won't push them to the buildTools repo until we decide to merge this.

### DIFF
--- a/SSI.php
+++ b/SSI.php
@@ -251,6 +251,52 @@ function ssi_shutdown()
 }
 
 /**
+ * Show the SMF version.
+ */
+function ssi_version($output_method = 'echo')
+{
+	if ($output_method == 'echo')
+		echo SMF_VERSION;
+	else
+		return SMF_VERSION;
+}
+
+/**
+ * Show the full SMF version string.
+ */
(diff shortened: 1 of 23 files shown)


Sesquipedalian:
@jdarwood007, it occurs to me that this change might require some tinkering with the scripts that will build patch releases for 2.1.x in the future. Will that be a problem, do you think?

jdarwood007:
We could work around any issues as needed.

PR #5380, opened 2019-01-29
Tests before acting in postgres 'update ban ip with ipv6 support' step

### LINKED ISSUE #5377: Errors while upgrading PostgreSQL
#### Description
A couple of minor issues occur when trying to upgrade on PostgreSQL from a recent Git build to the most recent one.

First, you get an error when attempting to upgrade the bans table to support IPv6 because there is no longer an "ip_low1" column. The only easy way to get around this is to either ignore errors from this or check whether an upgrade is needed.

Second, if the index already exists, you also get an error.

### Steps to reproduce
1. Attempt to upgrade a forum running 2.1 on PostgreSQL.

- Database Type: Postgres
- Database Version: 10.6 (server) / 11.1 (local test site)
- PHP Version: 7.3

The first one should be relatively simple to fix but I haven't had time to look into it yet. Basically we just need to see if there's an "ip_low1" column in the bans table before we run the query to update that table.

The second issue can be fixed by adding ` IF NOT EXISTS` to the `CREATE INDEX` statement.

Sesquipedalian:
Fixes #5377 

Give this a try and let me know if it works for you, @Oldiesmann.

### DIFF
--- a/other/upgrade_2-1_postgresql.sql
+++ b/other/upgrade_2-1_postgresql.sql
@@ -2308,18 +2308,42 @@ DROP INDEX IF EXISTS {$db_prefix}topics_id_board;
 --- update ban ip with ipv6 support
 /******************************************************************************/
 ---# add columns
-ALTER TABLE {$db_prefix}ban_items ADD COLUMN ip_low inet;
-ALTER TABLE {$db_prefix}ban_items ADD COLUMN ip_high inet;
+---{
+$request = upgrade_query("
+	SELECT column_name
+	FROM INFORMATION_SCHEMA.COLUMNS WHERE table_name = '{$db_prefix}ban_items';
+");
+$upcontext['ban_items_columns'] = $smcFunc['db_fetch_all']($request);
+$smcFunc['db_free_result']($request);
+
+if (!in_array('ip_low', $upcontext['ban_items_columns']))
+	upgrade_query("
+		ALTER TABLE {$db_prefix}ban_items ADD COLUMN ip_low inet;
+	");
+
+if (!in_array('ip_high', $upcontext['ban_items_columns']))
+	upgrade_query("
+		ALTER TABLE {$db_prefix}ban_items ADD COLUMN ip_high inet;
+	");
+---}
 ---#
 
 ---# convert data
-UPDATE {$db_prefix}ban_items
-SET ip_low = (ip_low1||'.'||ip_low2||'.'||ip_low3||'.'||ip_low4)::inet,
-	ip_high = (ip_high1||'.'||ip_high2||'.'||ip_high3||'.'||ip_high4)::inet
-WHERE ip_low1 > 0;
+---{
+if (in_array('ip_low1', $upcontext['ban_items_columns']))
+	upgrade_query("
+		UPDATE {$db_prefix}ban_items
+		SET ip_low = (ip_low1||'.'||ip_low2||'.'||ip_low3||'.'||ip_low4)::inet,
+			ip_high = (ip_high1||'.'||ip_high2||'.'||ip_high3||'.'||ip_high4)::inet
+		WHERE ip_low1 > 0;
+	");
+
+unset($upcontext['ban_items_columns']);
+---}
 ---#
 
 ---# index
+DROP INDEX IF EXISTS {$db_prefix}ban_items_id_ban_ip;
 CREATE INDEX {$db_prefix}ban_items_id_ban_ip ON {$db_prefix}ban_items (ip_low,ip_high);
 ---#
 


albertlast:
You forgot the other hundred places,
Where the check is needed.

other/upgrade_2-1_postgresql.sql
```diff
@@ -2308,18 +2308,42 @@ DROP INDEX IF EXISTS {$db_prefix}topics_id_board;
...
-ALTER TABLE {$db_prefix}ban_items ADD COLUMN ip_high inet;
+---{
+$request = upgrade_query("
+	SELECT column_name
+	FROM INFORMATION_SCHEMA.COLUMNS WHERE table_name = '{$db_prefix}ban_items';
+");
+$upcontext['ban_items_columns'] = $smcFunc['db_fetch_all']($request);
+$smcFunc['db_free_result']($request);
+
+if (!in_array('ip_low', $upcontext['ban_items_columns']))
```
albertlast, line 2319: the better code would be
```
$smileys_columns = $smcFunc['db_list_columns']('{$db_prefix}ban_items');
if (in_array('ip_low', $smileys_columns))
{
```

albertlast:
You miss checks here:
https://github.com/SimpleMachines/SMF2.1/blob/8eeff0432c727a35a21491266d6ca37531042be8/other/upgrade_2-1_postgresql.sql#L2345-L2444
and here:
https://github.com/SimpleMachines/SMF2.1/blob/8eeff0432c727a35a21491266d6ca37531042be8/other/upgrade_2-1_postgresql.sql#L2454-L2492
and here:
https://github.com/SimpleMachines/SMF2.1/blob/8eeff0432c727a35a21491266d6ca37531042be8/other/upgrade_2-1_postgresql.sql#L2699-L2706

Do you no think that i ask for general solution because the issue are not only ~30 line codes???

albertlast (commented):
The php code in sql file make it hard to read,
add the essential feature which i asked you and stuff could be solved in a nice way.

other/upgrade_2-1_postgresql.sql
Oldiesmann: I know it's less code to do it that way, but it seems a bit silly to go through the DB abstraction layer when we already know what DB system we're using.

albertlast: It's less code and the smf default way to retrieve the information,
the sql command which sequi use is very abnomarl way in pg world to retrieve the information.

so the used code is wrong in many ways.

Sesquipedalian:
To be honest, @albertlast, your explanation of your proposed general solution was not very clear to me. From what I can tell, the idea is basically to keep a permanent log of the upgrader steps that have been completed so that they can be skipped if the upgrader is run again. If my understanding is correct, I don't think it is a very good solution. Any time the code in the upgrader changes, that log will become invalid. Moreover, merely checking a log of steps that were completed in the past does not tell us anything about the integrity of the data now. It is better and safer to check the existing data at runtime.

albertlast:
I know,
reason why i ask for a differently solution, https://github.com/SimpleMachines/SMF2.1/issues/5377#issuecomment-458681586
where we could place a check in front of step and when true,
that we skip the step.

Sesquipedalian:
So, is the idea that if `$upcontext['skipStep'] === true`, then we don't execute any more of the substeps within the current major step?

albertlast:
Yes,
This would clean up upgrade file and
Allow to place the needed logic.


Gesendet: Mittwoch, 30. Januar 2019 08:33
An: SimpleMachines/SMF2.1
Cc: albertlast; Mention
Betreff: Re: [SimpleMachines/SMF2.1] Tests before acting in postgres 'update ban ip with ipv6 support' step (#5380)


So, is the idea that if $upcontext['skipStep'] === true, then we don't execute any more of the substeps within the current major step?

Sesquipedalian:
Hm. Yes, that does make a lot more sense than what I thought you were saying before.

Making this idea work probably just needs some additional conditions to test on this line:
https://github.com/SimpleMachines/SMF2.1/blob/8eeff0432c727a35a21491266d6ca37531042be8/other/upgrade.php#L1909

... and some code added inside this `if` block to reset the `$upcontext['skipStep']` to false:
https://github.com/SimpleMachines/SMF2.1/blob/8eeff0432c727a35a21491266d6ca37531042be8/other/upgrade.php#L1934

I'll need to check whether this change would work when JavaScript is disabled and when running from the command line, but my guess is that it should be possible.

If I start on a PR to implement this idea, can you start going through the upgrade_2-1_*.sql files to add the conditional checks to the steps that need them?

albertlast:
Sure,
would be `` $upcontext['skipStep'] = true`` than offical solution or
did we use some different kind?

Sesquipedalian:
Yes, I'm using `$upcontext['skipStep']`.

Sesquipedalian:
Please see #5382, @albertlast.

PR #5382, opened 2019-01-31
Adds ability to skip DB change substeps in the upgrader

Sesquipedalian:
If a PHP code block within one of the upgrade*.sql files sets $upcontext['skipStep'] to true, then the rest of the substeps within the current main step of the database changes will be skipped. Once the upgrader reaches the next main step (meaning, a line formatted like `--- Adding new settings...`), $upcontext['skipStep'] will be set back to false, and the upgrader will continue normally. This is especially useful for conducting sanity checks before applying changes.

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -155,6 +155,8 @@ if (isset($upgradeData))
 	$upcontext['updated'] = $upcontext['user']['updated'];
 
 	$is_debug = !empty($upcontext['user']['debug']) ? true : false;
+
+	$upcontext['skipStep'] = !empty($upcontext['user']['skipStep']);
 }
 
 // Nothing sensible?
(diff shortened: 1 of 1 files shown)


albertlast:
well it look like some bugs exists,
on the one hand i think you need define the $upcontext in eval call to be global too,
like $smcfunc etc.
and for some reason the $upcontext['skipStep'] value get lost between the substeps.

Sesquipedalian:
Thanks. I'll try to take a look tonight. In the meantime, can I get you to add $upcontext to the eval() in your local test environment and then test with the "Output extra debugging information" setting both on and off?

albertlast:
I added already upcontext to the globals of eval,
but like i said the information get lost somewhere than after this.

Sesquipedalian:
Try now, @albertlast.

PR #5383, opened 2019-01-31
Run upgrade twice fixes #5377

### LINKED ISSUE #5377: Errors while upgrading PostgreSQL

albertlast:
Thanks to @Sesquipedalian we could now create upgrade file which do a upgrade check in a easy fashion

One additional thing,
i removed the ' from "// PHP won't suppress" with itention,
because it's create a syntax highilight issue --> syntax highlight > right writing

atm is the code not tested

issue #5377

### DIFF
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -30,13 +30,29 @@ WHERE YEAR(birthdate) < 1004;
 ALTER TABLE {$db_prefix}calendar CHANGE start_date start_date date NOT NULL DEFAULT '1004-01-01';
 ALTER TABLE {$db_prefix}calendar CHANGE end_date end_date date NOT NULL DEFAULT '1004-01-01';
 ALTER TABLE {$db_prefix}calendar_holidays CHANGE event_date event_date date NOT NULL DEFAULT '1004-01-01';
-ALTER TABLE {$db_prefix}log_spider_stats CHANGE stat_date stat_date date NOT NULL DEFAULT '1004-01-01';
 ALTER TABLE {$db_prefix}members CHANGE birthdate birthdate date NOT NULL DEFAULT '1004-01-01';
+---#
+
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}log_spider_stats');
+$upcontext['skipStep'] = !in_array('state_date', $table_columns);
+---}
+---#
+
+---# Changing default values
+ALTER TABLE {$db_prefix}log_spider_stats CHANGE stat_date stat_date date NOT NULL DEFAULT '1004-01-01';
 ---#
 
 /******************************************************************************/
 --- Adding new settings...
 /******************************************************************************/
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}member_logins');
+$upcontext['skipStep'] = count($table_columns) > 0 ? true : false;
+---}
+---#
 
 ---# Adding login history...
 CREATE TABLE IF NOT EXISTS {$db_prefix}member_logins (
@@ -480,6 +496,12 @@ elseif (empty($modSettings['json_done']))
 /******************************************************************************/
 --- Adding support for logging who fulfils a group request.
 /******************************************************************************/
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}log_group_requests');
+$upcontext['skipStep'] = in_array('act_reason', $table_columns);
+---}
+---#
 
 ---# Adding new columns to log_group_requests
 ALTER TABLE {$db_prefix}log_group_requests
@@ -499,6 +521,13 @@ ADD INDEX `idx_id_member` (`id_member`, `id_group`);
 /******************************************************************************/
 --- Adding support for <credits> tag in package manager
 /******************************************************************************/
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}log_packages');
+$upcontext['skipStep'] = in_array('credits', $table_columns);
+---}
+---#
+
 ---# Adding new columns to log_packages ..
 ALTER TABLE {$db_prefix}log_packages
 ADD COLUMN credits TEXT NOT NULL;
@@ -507,6 +536,7 @@ ADD COLUMN credits TEXT NOT NULL;
 /******************************************************************************/
 --- Adding more space for session ids
 /******************************************************************************/
+
 ---# Altering the session_id columns...
 ALTER TABLE {$db_prefix}log_online
 CHANGE `session` `session` VARCHAR(128) NOT NULL DEFAULT '';
@@ -521,6 +551,13 @@ CHANGE `session_id` `session_id` VARCHAR(128) NOT NULL DEFAULT '';
 /******************************************************************************/
 --- Adding support for MOVED topics enhancements
 /******************************************************************************/
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}topics');
+$upcontext['skipStep'] = in_array('redirect_expires', $table_columns);
+---}
+---#
+
 ---# Adding new columns to topics ..
 ALTER TABLE {$db_prefix}topics
 ADD COLUMN redirect_expires INT(10) UNSIGNED NOT NULL DEFAULT '0',
@@ -530,6 +567,13 @@ ADD COLUMN id_redirect_topic MEDIUMINT UNSIGNED NOT NULL DEFAULT '0';
 /******************************************************************************/
 --- Adding new scheduled tasks
 /******************************************************************************/
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}scheduled_tasks');
+$upcontext['skipStep'] = in_array('callable', $table_columns);
+---}
+---#
+
 ---# Adding a new column "callable" to scheduled_tasks table
 ALTER TABLE {$db_prefix}scheduled_tasks
 ADD COLUMN callable VARCHAR(60) NOT NULL DEFAULT '';
@@ -625,6 +669,12 @@ ADD COLUMN deny_member_groups VARCHAR(255) NOT NULL DEFAULT '';
 /******************************************************************************/
 --- Adding setting for max depth of sub-boards to check for new posts, etc.
 /******************************************************************************/
+---# upgrade check
+---{
+$upcontext['skipStep'] = isset($modSettings['boardindex_max_depth']);
+---}
+---#
+
 ---# Adding the boardindex_max_depth setting.
 INSERT INTO {$db_prefix}settings
 	(variable, value)
@@ -696,6 +746,13 @@ if (!empty($member_groups))
 /******************************************************************************/
 --- Adding support for category descriptions
 /******************************************************************************/
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}categories');
+$upcontext['skipStep'] = in_array('description', $table_columns);
+---}
+---#
+
 ---# Adding new columns to categories...
 ALTER TABLE {$db_prefix}categories
 ADD COLUMN description TEXT NOT NULL;
@@ -704,6 +761,13 @@ ADD COLUMN description TEXT NOT NULL;
 /******************************************************************************/
 --- Adding support for alerts
 /******************************************************************************/
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}members');
+$upcontext['skipStep'] = in_array('alerts', $table_columns);
+---}
+---#
+
 ---# Adding the count to the members table...
 ALTER TABLE {$db_prefix}members
 ADD COLUMN alerts INT(10) UNSIGNED NOT NULL DEFAULT '0';
@@ -852,6 +916,13 @@ DROP COLUMN disregarded;
 /******************************************************************************/
 --- Fixing mail queue for long messages
 /******************************************************************************/
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}membergroups');
+$upcontext['skipStep'] = in_array('icons', $table_columns);
+---}
+---#
+
 ---# Altering mil_queue table...
 ALTER TABLE {$db_prefix}mail_queue
 CHANGE body body mediumtext NOT NULL;
@@ -966,7 +1037,7 @@ foreach ($toMove as $move)
 	$image = explode('#', $move);
 	$image = $image[1];
 
-	// PHP won't suppress errors when running things from shell, so make sure it exists first...
+	// PHP wont suppress errors when running things from shell, so make sure it exists first...
 	if (file_exists($modSettings['theme_dir'] . '/images/' . $image))
 		@rename($modSettings['theme_dir'] . '/images/' . $image, $modSettings['theme_dir'] . '/images/membericons/'. $image);
 }
@@ -1022,6 +1093,13 @@ $smcFunc['db_query']('', '
 /******************************************************************************/
 --- Messenger fields
 /******************************************************************************/
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}custom_fields');
+$upcontext['skipStep'] = in_array('field_order', $table_columns);
+---}
+---#
+
 ---# Adding new field_order column...
 ALTER TABLE {$db_prefix}custom_fields
 ADD COLUMN field_order SMALLINT NOT NULL DEFAULT '0';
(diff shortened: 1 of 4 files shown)


albertlast:
well the second commit from my side did some stuff where i'm not sure which route i want to go...
when we insert values and ther already values with this name.
Want we delete them or keep them untouched?
atm i delete/override them.

Oldiesmann:
We need to leave them alone. We especially don't want to change things in the settings table unless absolutely necessary as that will only annoy users.

albertlast:
thanks @Sesquipedalian i was able to change the upgrader in a way where no sql error appears in the pg log while running upgrader multiple times.

i used three major approach to checks
for column exists
```
---# upgrade check
---{
$table_columns = $smcFunc['db_list_columns']('{db_prefix}log_group_requests');
$upcontext['skipStep'] = in_array('act_reason', $table_columns);
---}
---#
```

for table exists
```
---# upgrade check
---{
$table_columns = $smcFunc['db_list_columns']('{db_prefix}user_drafts');
$upcontext['skipStep'] = count($table_columns) > 0 ? true : false;
---}
---#
```

for column typ check
```
---# upgrade check
---{
$column_info = upgradeGetColumnInfo('{db_prefix}member_logins', 'ip');
if (stripos($column_info['type'], 'inet') !== false)
	$upcontext['skipStep'] = true;
---}
---#
```

Sesquipedalian:
Excellent. Would you mind looking over the MySQL file as well?

albertlast:
added mysql too
mysql got no error log of sql statements so i only transfer the changes from pg to mysql

on pg side i need a test if the upgrade work from 2.0.x to 2.1.
the upgrade to upgrade is already tested

on mysql side both tests are open atm.

Sesquipedalian:
Thanks, @albertlast. Just to clarify, what does "both tests are open" mean?

albertlast:
mysql 2.0 to 2.1
mysql 2.1 to 2.1

Sesquipedalian:
Okay. But does saying that those tests are "open" mean that they have not been tested yet?

albertlast:
we need someone how test this,
yes.

Sesquipedalian:
When upgrading MySQL, using `$smcFunc['db_list_columns']()` to test whether a table exists causes the upgrade to halt. For example, the following code halts the process during a 2.0 → 2.1 upgrade because the member_logins table does not exist when it is run:
```
---# upgrade check
---{
$table_columns = $smcFunc['db_list_columns']('{db_prefix}member_logins');
$upcontext['skipStep'] = count($table_columns) > 0 ? true : false;
---}
---#
```

Sesquipedalian:
Since we can use `CREATE TABLE IF NOT EXISTS` in both MySQL and PostgreSQL, maybe we don't need to bother with a test before doing that. Alternatively, `$smcFunc['db_list_tables']()` from DbExtra-*.php might be appropriate here.

Sesquipedalian:
By the way, it is okay to do a check in the middle of a step. For example, in the following, the first and second sub-steps would be performed, but the third and fourth sub-steps would not:
```
--- Foo...
---# first sub-step...
< SQL code goes here >
---#

---# second sub-step...
---{
< php code goes here >
---}
---#

---# upgrade check
---{
$bar = false;
$upcontext['skipStep'] = $bar;
---}
---#

---# third sub-step...
< SQL code goes here >
---#

---# fourth sub-step...
---{
< php code goes here >
---}
---#

```

albertlast:
I use the table check as sign/ indicator if the step needs to be run or not.


Gesendet: Samstag, 2. Februar 2019 01:12
An: SimpleMachines/SMF2.1
Cc: albertlast; Mention
Betreff: Re: [SimpleMachines/SMF2.1] Run upgrade twice fixes #5377 (#5383)


Since we can use CREATE TABLE IF NOT EXISTS in both MySQL and PostgreSQL, maybe we don't need to bother with a test before doing that. Alternatively, $smcFunc['db_list_tables']() from DbExtra-*.php might be appropriate here.

albertlast:
I use this already on one place,
But I don't like it.
Because from maintenance pov it's harder to find/see.


Gesendet: Samstag, 2. Februar 2019 01:21
An: SimpleMachines/SMF2.1
Cc: albertlast; Mention
Betreff: Re: [SimpleMachines/SMF2.1] Run upgrade twice fixes #5377 (#5383)


By the way, it is okay to do a check in the middle of a step. For example, in the following, the first and second sub-steps would be performed, but the third and fourth sub-steps would not:

--- Foo...
---# first sub-step...
< SQL code goes here >
---#

---# second sub-step...
---{
< php code goes here >
---}
---#

---# upgrade check
---{
$bar = false;
$upcontext['skipStep'] = $bar;
---}
---#

---# third sub-step...
< SQL code goes here >
---#

---# fourth sub-step...
---{
< php code goes here >
---}
---#

Sesquipedalian:
> I use the table check as sign/ indicator if the step needs to be run or not.

Well, we either need to use a different function to do the test when the table may not exist yet, or else we need to `CREATE TABLE IF EXISTS` before using this function.

Sesquipedalian:
I think it would be best to use an `$upcontext['skipStep'] = true;` test only in places where it is absolutely necessary in order to avoid errors. In places where the existing logic can already handle the different possibilities, such as `CREATE TABLE IF EXISTS` statements, we should just leave it alone.

other/upgrade_2-1_mysql.sql
```diff
@@ -30,13 +30,29 @@ WHERE YEAR(birthdate) < 1004;
...
 ALTER TABLE {$db_prefix}calendar CHANGE end_date end_date date NOT NULL DEFAULT '1004-01-01';
 ALTER TABLE {$db_prefix}calendar_holidays CHANGE event_date event_date date NOT NULL DEFAULT '1004-01-01';
-ALTER TABLE {$db_prefix}log_spider_stats CHANGE stat_date stat_date date NOT NULL DEFAULT '1004-01-01';
 ALTER TABLE {$db_prefix}members CHANGE birthdate birthdate date NOT NULL DEFAULT '1004-01-01';
 ---#
 
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}log_spider_stats');
+$upcontext['skipStep'] = !in_array('state_date', $table_columns);
```
Sesquipedalian: 1. Why test this one but none of the others?
2. Do we need to bother testing this anyway? I don't think altering the default value of a column to the same default it already has will cause any errors, so we can probably just leave it alone.

other/upgrade_2-1_mysql.sql
```diff
@@ -1190,6 +1268,13 @@ ALTER TABLE `{$db_prefix}members`
 /******************************************************************************/
 --- Adding support for drafts
 /******************************************************************************/
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}user_drafts');
```
Sesquipedalian: This will fail and cause the upgrader to halt if the table doesn't exist. Also, I don't think we need to test this anyway.

other/upgrade_2-1_mysql.sql
```diff
@@ -1291,6 +1383,13 @@ ADD COLUMN likes SMALLINT(5) UNSIGNED NOT NULL DEFAULT '0';
 /******************************************************************************/
 --- Adding support for mentions
 /******************************************************************************/
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}mentions');
```
Sesquipedalian: This will fail and cause the upgrader to halt if the table doesn't exist. I don't think we need to test this.

other/upgrade_2-1_mysql.sql
```diff
@@ -1462,6 +1561,13 @@ $smcFunc['db_free_result']($file_check);
 /******************************************************************************/
 --- Upgrading "verification questions" feature
 /******************************************************************************/
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}qanda');
```
Sesquipedalian: This will fail and cause the upgrader to halt if the table doesn't exist. I don't think we need to test this.

other/upgrade_2-1_mysql.sql
```diff
@@ -1607,6 +1713,13 @@ $request = upgrade_query("
 /******************************************************************************/
 --- Upgrading PM labels...
 /******************************************************************************/
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}pm_recipients');
```
Sesquipedalian: I don't think we should do this test. The "Moving label info to new tables and updating rules" sub-step contains lots of different things that it needs to check and manipulate—much more than this simple test can account for.

other/upgrade_2-1_postgresql.sql
```diff
@@ -212,13 +212,29 @@ WHERE EXTRACT(YEAR FROM birthdate) < 1004;
...
 ALTER TABLE {$db_prefix}calendar ALTER COLUMN end_date SET DEFAULT '1004-01-01'::date;
 ALTER TABLE {$db_prefix}calendar_holidays ALTER COLUMN event_date SET DEFAULT '1004-01-01'::date;
-ALTER TABLE {$db_prefix}log_spider_stats ALTER COLUMN state_date SET DEFAULT '1004-01-01'::date;
 ALTER TABLE {$db_prefix}members ALTER COLUMN birthdate SET DEFAULT '1004-01-01'::date;
 ---#
 
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}log_spider_stats');
+$upcontext['skipStep'] = !in_array('state_date', $table_columns);
```
Sesquipedalian: 1. Why test this one but none of the others?
2. Do we need to bother testing this anyway? I don't think altering the default value of a column to the same default it already has will cause any errors, so we can probably just leave it alone.

other/upgrade_2-1_postgresql.sql
```diff
@@ -212,13 +212,29 @@ WHERE EXTRACT(YEAR FROM birthdate) < 1004;
...
+---# Changing default values
+ALTER TABLE {$db_prefix}log_spider_stats ALTER COLUMN state_date SET DEFAULT '1004-01-01'::date;
+---#
+
 /******************************************************************************/
 --- Adding new settings...
 /******************************************************************************/
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}member_logins');
```
Sesquipedalian: This will fail and cause the upgrader to halt if the table doesn't exist. I don't think we need to test this anyway.

other/upgrade_2-1_postgresql.sql
```diff
@@ -1383,6 +1454,13 @@ ALTER TABLE {$db_prefix}members
 /******************************************************************************/
 --- Adding support for drafts
 /******************************************************************************/
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}user_drafts');
```
Sesquipedalian: This will fail and cause the upgrader to halt if the table doesn't exist. I don't think we need to test this.

other/upgrade_2-1_postgresql.sql
```diff
@@ -1419,20 +1497,18 @@ if (@$modSettings['smfVersion'] < '2.1')
...
 	{
-		foreach ($inserts AS $insert)
-		{
-			upgrade_query("
-				INSERT INTO {$db_prefix}board_permissions
-					(id_group, id_board, permission, add_deny)
-				VALUES
-					" . $insert);
-		}
+		$smcFunc['db_insert']('replace',
```
Sesquipedalian, line 1414: Nice.

other/upgrade_2-1_postgresql.sql
```diff
@@ -1443,32 +1519,40 @@ if (@$modSettings['smfVersion'] < '2.1')
...
 INSERT INTO {$db_prefix}themes (id_theme, variable, value) VALUES ('1', 'drafts_show_saved_enabled', '1');
 ---#
 
 /******************************************************************************/
 --- Adding support for likes
 /******************************************************************************/
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}messages');
+$upcontext['skipStep'] = in_array('likes', $table_columns);
```
Sesquipedalian: What if the likes column exists in the message table, but the user_likes table doesn't? This test should be moved to a position immediately before the "Adding count to the messages table" sub-step.

other/upgrade_2-1_postgresql.sql
```diff
@@ -1490,6 +1574,13 @@ ADD COLUMN likes smallint NOT NULL default '0';
 /******************************************************************************/
 --- Adding support for mentions
 /******************************************************************************/
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}mentions');
```
Sesquipedalian: This will fail and cause the upgrader to halt if the table doesn't exist. Instead, after the `CREATE TABLE`, maybe we should test whether the indexes already exist or not.

other/upgrade_2-1_postgresql.sql
```diff
@@ -1685,6 +1776,13 @@ $smcFunc['db_free_result']($file_check);
 /******************************************************************************/
 --- Upgrading "verification questions" feature
 /******************************************************************************/
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}qanda');
```
Sesquipedalian: This will fail and cause the upgrader to halt if the table doesn't exist.

other/upgrade_2-1_postgresql.sql
```diff
@@ -1814,33 +1910,37 @@ $request = upgrade_query("
...
 	}
 ---}
 ---#
 
 /******************************************************************************/
 --- Upgrading PM labels...
 /******************************************************************************/
+---# upgrade check
+---{
+$table_columns = $smcFunc['db_list_columns']('{db_prefix}pm_recipients');
```
Sesquipedalian: I don't think we should do this test. The "Moving label info to new tables and updating rules" sub-step contains lots of different things that it needs to check and manipulate—much more than this simple test can account for.

Sesquipedalian (changes requested):

other/upgrade_2-1_postgresql.sql
albertlast: Like i said,
the upgrader is tested on pg.
so db_list_columns works nice here to test the existing of table

albertlast:
> 
> 
> I think it would be best to use an `$upcontext['skipStep'] = true;` test only in places where it is absolutely necessary in order to avoid errors. In places where the existing logic can already handle the different possibilities, such as `CREATE TABLE IF EXISTS` statements, we should just leave it alone.

yeah i see the point,
so what you define as necessary?
At the point I do all the places where sql error occured,
maybe switch to places only where upgrader stop/break?

another point would instead of checking "something" from the step,
i could check if the last changed is applied.

Sesquipedalian:
I'd define necessary as meeting at least one of these two conditions:

1. If an error while running the step causes the upgrade to halt or display an error message to the user.
2. Running the step as second time messes up the data and/or the database structure.

So, for SQL errors that we can recover from gracefully, there is no need for a check.

Sesquipedalian:
Checking the last sub-step is probably a good default, but the best choice will need to be decided on a case-by-case basis.

albertlast:
@Sesquipedalian changes are done

Sesquipedalian:
I'm not done testing yet, but I noticed that smf_db_list_columns() needs to include `table_schema` in the WHERE clause, just in case the same column name appears in more than one MySQL database.
```
function smf_db_list_columns($table_name, $detail = false, $parameters = array())
{
	global $smcFunc, $db_prefix, $db_name;

	$table_name = str_replace('{db_prefix}', $db_prefix, $table_name);

		SELECT column_name "Field", COLUMN_TYPE "Type", is_nullable "Null", COLUMN_KEY "Key" , column_default "Default", extra "Extra"
		FROM information_schema.columns
		WHERE table_name = {string:table_name}
			AND table_schema = {string:db_name}
		ORDER BY ordinal_position',
		array(
			'table_name' => $table_name,
			'db_name' => $db_name,
		)
	);
...
```

Sesquipedalian:
I can confirm that once the change described in my 
[previous commnent](https://github.com/SimpleMachines/SMF2.1/pull/5383#issuecomment-460097072) is made, running and re-running the upgrader works nicely for MySQL.

albertlast:
Will you create a pr with the schema name fix?
Because I got surprised that you merge this without adding this.

Sesquipedalian:
I did add it. I was just about to post a comment here letting you know that, but you beat me to it. 🙂

Sesquipedalian:
https://github.com/SimpleMachines/SMF2.1/commit/5b563e1e136a85a3b613cdb36bcf2143aa0e6e7a

Sesquipedalian:
I also decided to rename `$upcontext['skipStep']` to `$upcontext['skip_db_substeps']` in the merged version, just to help prevent confusion for developers reading the code in the future.

PR #5386, opened 2019-02-05
Fix sceEditor li

albertlast:
Dunno what the error message is,
but the issue looks real.

Src: https://www.simplemachines.org/community/index.php?topic=564893.0


frandominguezl:
OP says it's a syntax error

Sesquipedalian (approved):

PR #5389, opened 2019-02-07
fix split topic broken fixes #2643

### LINKED ISSUE #2643: split topic broken?
try to split a topic, go to a topic with many replies and click split in the more drop down from one of the replies. you are brought to a page to select the replies to split. click the arrow to move the replies and nothing happens.

if i refresh this page the replies get moved to the right side but they do not until i refresh the page. in 2.0 the replies would move to the right without a page refresh required.

a few related errors

 8: Undefined index: page_index
/github2.1/index.php?action=splittopics;sa=selectTopics;subname=test;topic=6.0;start2=0;move=down;msg=29;xml
F:/wamp/www/github2.1/Sources/Subs.php
Line: 572

this error shows on several different line numbers, 572, 571,654,624 all in Subs.php

albertlast:
show Error message is missing,
because i want to know if the way correct.

Issue: #2643

### DIFF
--- a/Sources/SplitTopics.php
+++ b/Sources/SplitTopics.php
@@ -624,8 +624,7 @@ function splitTopic($split1_ID_TOPIC, $splitMessages, $new_subject)
 
 	// You cannot split off the first message of a topic.
 	if ($split1_first_msg > $split2_first_msg)
-		redirectexit ('action=splittopics;sa=selectTopics;subname=' . $new_subject . ';topic=' . $split1_ID_TOPIC . '.0;start2=0');
-		//fatal_lang_error('split_first_post', false);
+		fatal_lang_error('split_first_post', false);
 
 	// We're off to insert the new topic!  Use 0 for now to avoid UNIQUE errors.
 	$split2_ID_TOPIC = $smcFunc['db_insert']('',
--- a/Sources/SplitTopics.php
+++ b/Sources/SplitTopics.php
@@ -627,7 +627,8 @@ function splitTopic($split1_ID_TOPIC, $splitMessages, $new_subject)
 
 	// You cannot split off the first message of a topic.
 	if ($split1_first_msg > $split2_first_msg)
-		fatal_lang_error('split_first_post', false);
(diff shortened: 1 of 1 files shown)


albertlast:
@Sesquipedalian did you tell me wiith the assign to rc2 that way is right?

Sesquipedalian:
Using `redirectExit()` is the right idea, but it would be better to do it near the start of the process in `SplitExecute()` rather than at the last moment in `splitTopic()`.

Just replace this:
```
	// Redirect to the selector if they chose selective.
	if ($_POST['step2'] == 'selective')
	{
		$_REQUEST['subname'] = $_POST['subname'];
		return SplitSelectTopics();
	}
```
... with this:
```
	// Redirect to the selector if they chose selective.
	if ($_POST['step2'] == 'selective')
		redirectexit ('action=splittopics;sa=selectTopics;subname=' . $_POST['subname'] . ';topic=' . $topic . '.0;start2=0');
```

Sesquipedalian:
Oh, and just in case it wasn't clear, you can also restore the `fatal_lang_error('split_first_post', false);` line back to what it was.

Sesquipedalian:
> @Sesquipedalian did you tell me wiith the assign to rc2 that way is right?

Sorry, I didn't see this reply right away. The answer to this question is no. Marking this for RC2 just means that I think this PR needs to be dealt with before RC2 is finished. When I think the code in the PR is good, I will give it an "Approved" review (or just merge without a review it if I am feeling lazy 😉).

albertlast:
I keep looking at you recommandation and
i didn't understand them.

The check is placed on the position where the issue get detected(in my version).
you code disable the function of spliting posting,
because you always redirect the user...

Sesquipedalian:
> The check is placed on the position where the issue get detected(in my version).
> you code disable the function of spliting posting,
> because you always redirect the user...

No, my suggestion will only redirect the user if she is coming from the page that asks her which method she wants to use to split the topic. Once she has been redirected to the page that handles selective splitting, everything will proceed as normal.

albertlast:
Okay i didn't understand fully why,
but it work.

But ther some more issue in split operation view...

Sesquipedalian (approved):

PR #5393, opened 2019-02-08
Fixes serialized data corrupted by conversion to UTF-8 during upgrade

### LINKED ISSUE #5388: Possible SMF 2.1 upgrade json issue with displayFields
Link https://www.simplemachines.org/community/index.php?topic=564909.msg4005682#msg4005682

I did notice some html in json maybe that is corrupteing it
<a href="{INPUT}"></a>

Sesquipedalian:
Fixes #5388


albertlast:
For the display fields I would like to see that we generate the complete value new.
You don't like the idea?

Sesquipedalian:
It's a good idea for that particular case, but this problem could affect any string of serialized data containing non-ASCII  characters, so a general solution is needed.

PR #5401, opened 2019-02-08
Change mods column in members to nullable

albertlast:
search all columns in members which got no default value and are not nullable and are not smf default columns and
changed them to be nullable.

to catch issues like this https://www.simplemachines.org/community/index.php?topic=564951.0

atm is the pr not tested

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -2017,7 +2017,7 @@ function parse_sql($filename)
 				}
 
 				// @todo Update this to a try/catch for PHP 7+, because eval() now throws an exception for parse errors instead of returning false
-				if (eval('global $db_prefix, $modSettings, $smcFunc, $txt, $upcontext; ' . $current_data) === false)
+				if (eval('global $db_prefix, $modSettings, $smcFunc, $txt, $upcontext, $db_name; ' . $current_data) === false)
 				{
 					$upcontext['error_message'] = 'Error in upgrade script ' . basename($filename) . ' on line ' . $line_number . '!' . $endl;
 					if ($command_line)
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -2931,3 +2931,26 @@ VALUES ('Independence Day', '1004-07-04'),
 CREATE INDEX idx_id_thumb ON {$db_prefix}attachments (id_thumb);
 ---#
 
+/******************************************************************************/
+--- Fix members mods columns
+/******************************************************************************/
+---# make them nullable
+---{
+$request = upgrade_query("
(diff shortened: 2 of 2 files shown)


VBGAMER45:
Maybe also do the same for other tables i saw this one in messages table
https://www.simplemachines.org/community/index.php?topic=564909.msg4006012#msg4006012

MissAllSunday:
Since we do not know what tables can be modified by a mod, perhaps it will be good to do a mapping for all of them and apply the change to each.

albertlast:
Would be nice if someone can test this first

live627:
If anyone decides to merge this in and things randomly break, I will not shy away from general nastiness (aka "I told you so")

albertlast:
Since the fields are already nullable,
only no one did tell this mysql (the old mysql version ignore this information),
did i see no problems.

On the other hand i see no other solution.

sbulen:
Pretty much any table could be modded...  

We either need to go all-in or all-out...   Either this expands to do ALL tables, or, we get rid of STRICT.

albertlast:
for me it's soudn random to go the "all way" and this is not strict mode,
this is the default mode.

sbulen:
I tested this & it hangs on my WAMP forum during the install on this step:
image

albertlast:
Maybe the sql error was the problem.

sbulen:
For the record, I took a vanilla 2.0 forum, installed the adkportal, & looked at the columns that meet your criteria (default NULL, not nullable, not PRI) and there are 41 columns that meet the criteria.   So, once your logic executes, it will be altering other existing columns in members, boards & topics:
image

albertlast:
when you look at the code,
i always blacklist some columns: 
https://github.com/SimpleMachines/SMF2.1/blob/73c5ae85c1183d8f7077c31661476c35f6a00d52/other/upgrade_2-1_mysql.sql#L2944

sbulen:
Just pointing out that it's not all of them...

EDIT:  Is there a reason you didn't blacklist message_labels, openid_uri?

albertlast:
When you look at you screenshot,
in members table only three columns(massage_labels,openid_uri and adk_...) get fixed,
which looks right to me.

sbulen:
I don't understand why you would blacklist some vanilla SMF columns but not others?

albertlast:
because the vanilla are right,
only the mod column are wrong.

sbulen:
message_labels, openid_url are vanilla columns...

albertlast:
they are not part of the 2.1 setup any more

sbulen:
Got it.  In fact, they are removed during the upgrade.  Good.

Issue #5200, opened 2018-12-15
JavaScript error: new_replies is not defined


Oldiesmann:
### Steps to reproduce
1. Preview a post (new topic or reply to existing one)
2. Check the console in your browser's development tools (CTRL+SHIFT+I on Chrome)

```
Uncaught ReferenceError: new_replies is not defined
    at smc_preview_post.onDocSent (script.js?beta21:1835)
    at Object.success (script.js?beta21:90)
    at i (jquery.min.js:2)
    at Object.fireWith [as resolveWith] (jquery.min.js:2)
    at A (jquery.min.js:4)
    at XMLHttpRequest.<anonymous> (jquery.min.js:4)
```
Relevant code:
```js
	// Remove the new image from old-new replies!
	for (i = 0; i < new_replies.length; i++)
		document.getElementById(this.opts.sNewImageContainerID.replace('%ID%', new_replies[i])).style.display = 'none';
```

### Additional information/references
This code is meant to remove the "new" icon from any new replies made while you're posting, but new_replies is never defined anywhere.

frandominguezl:
I'm not much into JavaScript. It outputs an error because new_replies is defined as a new Array right below the for loop, which doesn't makes too much sense. A solution would be to declare the array before the loop, but then it will iterate over an empty array.

If we have a look at:
https://github.com/SimpleMachines/SMF2.1/blob/29559c8787971aeff5779da79ce2efedde004a6b/Themes/default/scripts/script.js#L1848

Makes more sense to define the array inside that if condition, change that line to:
`new_replies[i] = newPosts[i].getAttribute("id");`

And then call the for loop right after that one. Something like this:

```
if (numNewPosts != 0)
	{
		var newPostsHTML = '<span id="new_replies"><' + '/span>';
		var tempHTML;
		var new_replies = new Array();
		for (var i = 0; i < numNewPosts; i++)
		{
			new_replies[i] = newPosts[i].getAttribute("id");

			ignoring = false;
			if (newPosts[i].getElementsByTagName("is_ignored")[0].firstChild.nodeValue != 0)
				ignored_replies[ignored_replies.length] = ignoring = newPosts[i].getAttribute("id");

[...]

PR #5402, opened 2019-02-09
Misc minor fixes

Sesquipedalian:


Themes/default/scripts/script.js
```diff
@@ -1860,7 +1855,8 @@ smc_preview_post.prototype.onDocSent = function (XMLDoc)
 		var tempHTML;
 		for (var i = 0; i < numNewPosts; i++)
 		{
-			new_replies[new_replies.length] = newPosts[i].getAttribute("id");
+			// new_replies[new_replies.length] = newPosts[i].getAttribute("id");
```
frandominguezl, line 1858: What's with this line?

Sesquipedalian: Apparently I forgot to go back and delete it after I was done testing.

frandominguezl: Out of curiosity, how did you test it? I just came with that solution because it seemed legit, but I didn't know exactly how to do it.

Also, I can't seem to find intialization of new_replies. Is it needed in JS?

Sesquipedalian: > Out of curiosity, how did you test it?

In one browser window, I logged in as testuser1 and started a reply. In a different, private browser window, I logged in as testuser2 and posted a reply to the same topic. Then I went back to the first browser window and tried to post the reply. When I was sent back to the post authoring page with the usual warning message, I checked the JS console to see if the error reported in #5200 had occurred. It had not.

> I can't seem to find intialization of new_replies. Is it needed in JS?

Odd. I have a clear memory of adding `var new_replies = new Array();` to the script, but now I don't see it.

Anyway, variable declaration is a bit weird in JavaScript. If a variable is declared using the `var` keyword, it will be local to that function only. If it is declared without the `var` keyword, it will be a global variable. Meanwhile, a line like `new_replies[i] = newPosts[i].getAttribute("id");` will create the `new_replies` array implicitly if it does not already exist. So the code will work without the `var new_replies = new Array();` line, but it makes the `new_replies` variable global instead of local. It would be good form to fix that by adding the declaration.

Sesquipedalian: `var new_replies = new Array();` added in 2e48b89

PR #5406, opened 2019-02-10
Fix avatar not shown when editing it

### LINKED ISSUE #5119: Uploaded/URL avatars not shown when editing profile
#### Description
When you go to edit your forum profile, if you have chosen an avatar type other than gravatar, your avatar isn't shown. This is because the `` to show the avatars is inside the "avatar_gravatar" div, which is hidden when you haven't selected Gravatar.

### Steps to reproduce
1. Go to your profile and either upload an avatar via URL or upload one.
2. After you save the profile, the avatar you chose isn't displayed except in the user menu at the top left.

- Database Type: MariaDB
- Database Version: 10.0.36
- PHP Version: 7.1.23

The reason for this is that the `` to display your avatar is inside the "avatar_gravatar" div, which is hidden when you have chosen an option other than Gravatar. If you select the "Choose an avatar from the gallery" option after uploading an avatar or specifying one via URL, it will also show up to the right of the select box.

Relevant code from Profile.template.php:
```php
	// if the user is able to use Gravatar avatars show then the image preview
	if (!empty($context['member']['avatar']['allow_gravatar']))
	{
		echo '
								<div id="avatar_gravatar">
									<img src="' . $context['member']['avatar']['href'] . '" alt="">';```

### PR DESCRIPTION
**Pull request by SychO9**
Fixes #5119 
Makes it so that the avatar is shown on the chosen method's section

### DIFF
--- a/Themes/default/Profile.template.php
+++ b/Themes/default/Profile.template.php
@@ -2892,8 +2892,8 @@ function template_profile_avatar_select()
 									<div>
 										<select name="file" id="file" size="10" style="display: none;" onchange="showAvatar()" onfocus="selectRadioByName(document.forms.creator.avatar_choice, \'server_stored\');" disabled><option></option></select>
 									</div>
-									<div>
-										<img id="avatar" src="', !empty($context['member']['avatar']['allow_external']) && $context['member']['avatar']['choice'] == 'external' ? $context['member']['avatar']['external'] : $modSettings['avatar_url'] . '/blank.png', '" alt="Do Nothing">
+									<div class="edit_avatar_img">
+										<img id="avatar" src="' . $context['member']['avatar']['href'] . '" alt="">
 									</div>
 									<script>
 										var files = ["' . implode('", "', $context['avatar_list']) . '"];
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: Themes/default/Profile.template.php [hunk 1]
```diff
@@ -2892,8 +2892,8 @@ function template_profile_avatar_select()
 									<div>
 										<select name="file" id="file" size="10" style="display: none;" onchange="showAvatar()" onfocus="selectRadioByName(document.forms.creator.avatar_choice, \'server_stored\');" disabled><option></option></select>
 									</div>
-									<div>
-										<img id="avatar" src="', !empty($context['member']['avatar']['allow_external']) && $context['member']['avatar']['choice'] == 'external' ? $context['member']['avatar']['external'] : $modSettings['avatar_url'] . '/blank.png', '" alt="Do Nothing">
+									<div class="edit_avatar_img">
+										<img id="avatar" src="' . $context['member']['avatar']['href'] . '" alt="">
```
Sesquipedalian: This would be better:
```
<img id="avatar" src="', $context['member']['avatar']['choice'] == 'server_stored' ? $context['member']['avatar']['href'] : $modSettings['avatar_url'] . '/blank.png', '" alt="">
```

Sesquipedalian (changes requested):

Themes/default/Profile.template.php
SychO9: yea, my bad

Sesquipedalian (approved):

PR #5409, opened 2019-02-10
Check for Phar

albertlast:
To catch this case a litte bit better: https://www.simplemachines.org/community/index.php?topic=564987.0

### DIFF
--- a/Themes/default/languages/Packages.english.php
+++ b/Themes/default/languages/Packages.english.php
@@ -171,6 +171,7 @@ $txt['package_uploaded_successfully'] = 'The package has been uploaded successfu
 $txt['package_modification_malformed'] = 'Malformed or invalid modification file.';
 $txt['package_modification_missing'] = 'The file could not be found.';
 $txt['package_no_zlib'] = 'Sorry, your PHP configuration doesn\'t have support for <strong>zlib</strong>. Without this, the package manager cannot function. Please contact your host about this for more information.';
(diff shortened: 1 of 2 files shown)


Themes/default/languages/Packages.english.php
```diff
@@ -171,6 +171,7 @@
 $txt['package_modification_malformed'] = 'Malformed or invalid modification file.';
 $txt['package_modification_missing'] = 'The file could not be found.';
 $txt['package_no_zlib'] = 'Sorry, your PHP configuration doesn\'t have support for <strong>zlib</strong>. Without this, the package manager cannot function. Please contact your host about this for more information.';
+$txt['package_no_phar'] = 'Sorry, your PHP configuration doesn\'t have support for <strong>phar</strong>. Without this, the package manager cannot function. Please contact your host about this for more information.';
```
MissAllSunday, line 174: Perhaps now that there is another "no support for x" string might be a good idea to make it generic, IE: 

"Sorry, your PHP configuration doesn\'t have support for <strong>{library}</strong>. Without this, {feature} cannot function. Please contact your host about this for more information."

This will also help mod authors too.

albertlast: well i see this not as task of this pr,
feeld free to do a own pr or to duplicate this pr (than i can close this),
where you add this.

Sesquipedalian: @MissAllSunday's idea is a good one, but if @albertlast would rather have that left to another PR, that's fine.

MissAllSunday: Sure I'll add a generic string

live627 (approved):

PR #5410, opened 2019-02-10
Allow hooks to silently fail.

### LINKED ISSUE #4926: Package Manager Hook install can cause error log entries
When installing a mod using <hooks> you may receive multiple error log entries.

```
Hook call: function "shd_parse_wikilinks" in file /sdsmf21/Sources/sd_source/Subs-SimpleDesk.php could not be called.
The callable shd_parse_wikilinks could not be called.
Hook call: function "shd_bbc_codes" in file /sdsmf21/Sources/sd_source/Subs-SimpleDesk.php could not be called.
The callable shd_bbc_codes could not be called.
```

In Packages.php we have the code that handles hooks:
```
			elseif ($action['type'] == 'hook' && isset($action['hook'], $action['function']))
			{
				if ($action['reverse'])
					remove_integration_function($action['hook'], $action['function'], true, $action['include_file'], $action['object']);
				else
					add_integration_function($action['hook'], $action['function'], true, $action['include_file'], $action['object'], true);
			}
```

The add_integration_function updates the $modSettings[$hook] with the new information.  Which can cause SMF later on to use this thinking the hook is loaded up.

Seems like it should be simple to add another parameter and have the function ignore updating the $modSettings when coming from a package manager operation.

Actually no, that won't fix it up as then later on we have this:
```
		// Assuming we're not uninstalling, add the entry.
		if (!$context['uninstalling'])
		{
			// Reload the settings table for mods that have altered them upon installation
			reloadSettings();
```

The reloadSettings() call causes $modSettings to be repopulated as well.  But wait there is more!

add_integration_function has another surprise for us.

```
		updateSettings(array($hook => implode(',', $permanent_functions)));
```

UpdateSettings does this:
```
		$modSettings[$variable] = $value;
```

Seems no mater what there isn't an easy way to escape this.

Thoughts?

This isn't a 2.1 show stopper, but a real annoyance as more people use <hooks>

jdarwood007:
This fixes #4926 by adding a context check to allow Package Manager to specify that we are silently ignoring this.  This is the simplest of fixes without complex changes to the code to do the correct fixes in preventing reloading settings from being called.

### DIFF
--- a/Sources/Packages.php
+++ b/Sources/Packages.php
@@ -1026,6 +1026,10 @@ function PackageInstall()
 			}
 			elseif ($action['type'] == 'hook' && isset($action['hook'], $action['function']))
 			{
+				// Set the system to ignore hooks, but only if it wasn't changed before.
+				if (!isset($context['ignore_hook_errors']))
+					$context['ignore_hook_errors'] = true;
+
 				if ($action['reverse'])
 					remove_integration_function($action['hook'], $action['function'], true, $action['include_file'], $action['object']);
 				else
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -2072,7 +2072,7 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 			foreach ($section as $code)
 				$alltags[] = $code['tag'];
 		}
-		$alltags_regex = '(?>\b' . build_regex(array_unique($alltags)) . '\b|' . build_regex(array_keys($itemcodes)) . ')';
+		$alltags_regex = '(?' . '>\b' . build_regex(array_unique($alltags)) . '\b|' . build_regex(array_keys($itemcodes)) . ')';
 	}
 
 	$pos = -1;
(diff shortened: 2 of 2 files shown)


Sources/Packages.php
```diff
@@ -1026,6 +1026,10 @@ function PackageInstall()
 			}
 			elseif ($action['type'] == 'hook' && isset($action['hook'], $action['function']))
 			{
+				// Set the system to ignore hooks, but only if it wasn't changed before.
+				if (!isset($context['ignore_hook_errors']))
```
MissAllSunday, line 1030: Looks good, just curios about this, does this mean that if a mod doesn't want to ignore errors the mod author needs to explicitely set it to false?  if so, wouldn't it be better to define and set to true this key before any hook is run to make it the default behaviour?

jdarwood007: Well that would be possible.  But this is only being set during package install if a hook is used.
Technically it means another part of SMF could trigger this as well or another mod.

This is purposely handling just the hook handling where as mentioned the reloadSettings is called during the package install and then due to how other parts of SMF are executed afterwords, hooks may get called too early on, or without other needed hooks.

MissAllSunday: Ok cool!

PR #5411, opened 2019-02-10
PharData fixes for create backups

jdarwood007:
When you use open_basedir restrictions, PharData does not handle open_basedir restrictions.  The only bypass is to ignore the errors.  We selectively ignore only those errors to ensure we capture all other errors properly.


live627:
You should `return true` to signal PHP that you have handled the error.

live627 (approved):

PR #5413, opened 2019-02-10
Old icon cleanup

XinYenFon:
I left last icon part for compat but not sure if I should because anyone using old PNG request should update to class...

### DIFF
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -1388,7 +1388,7 @@ function editBuddyIgnoreLists($memID)
 	$context[$context['profile_menu_name']]['tab_data'] = array(
 		'title' => $txt['editBuddyIgnoreLists'],
 		'description' => $txt['buddy_ignore_desc'],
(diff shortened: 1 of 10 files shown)


live627:
Not so sure this is RC material...

Should this be saved for a future version (like 2.2 idk)?

XinYenFon:
I think this can slip into RC2, but its your call. You can use it whenever you want with or without credit. I'm only interested in helping and building :)

Gwenwyfar:
I don't see any problem with it, though it would need further adjustments.

Sesquipedalian:
I like this, since it allows themes more control over their choice of icons. But I'll leave it up to @Gwenwyfar to decide when/if to approve it. I've tentatively added the PR to the Final milestone for now, pending her decision.

XinYenFon:
oh btw I made a mistake and pushed a fix here... I should be doing it in another PR :| sorry.

XinYenFon:
this is no longer valid PR, some parts already merged and some are unnecessary.

PR #5415, opened 2019-02-11
Fix site slogan alignment & size

### LINKED ISSUE #5133: Slogan text size and alignment.
Dunno if this is intended behaviour, but I noticed when you add your own text for a slogan it seems to be the same size as the default logo text (forum name).  The slogan also sits higher in the container than the logo (forum name).  I'm guessing this will also increase the height of the container unnecessarily.

When you remove the slogan text and let it display the default SMF image it appears to sit at the same height as the forum name. 

21slogan1
21slogan2

SychO9:
Fixes #5133 
1.4em should be better than 1.8em for the size

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -1071,7 +1071,7 @@ h1.forumtitle a {
 #siteslogan, img#smflogo {
 	padding-right: 2px;
 	line-height: 50px;
-	font-size: 1.8em;
+	font-size: 1.4em;
 }
 /* Tweak the SMF logo */
 img#smflogo {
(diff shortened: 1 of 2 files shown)


Gwenwyfar:
Thanks :)

Gwenwyfar (approved):
This looks better with images too, though it might require to change to two lines for mobile.

SychO9:
> This looks better with images too, though it might require to change to two lines for mobile.

well under 720px the siteslogan is hidden

Gwenwyfar:
Ah, you're right. I forgot how many things we're just hiding in mobile...

PR #5420, opened 2019-02-12
Misc small css fixes

Gwenwyfar:
... and a language one that sneaked in :P

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -3449,7 +3449,7 @@ ul.post_options li {
 }
 #moderationbuttons_strip {
 	float: left;
-	margin: 4px 0 4px -4px;
+	margin: 4px 0;
 }
 #moderationbuttons ul li {
 	padding: 0 6px 4px 0;
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -3394,7 +3394,7 @@ ul.post_options li {
 #post_draft_options .settings strong {
 	color: #555;
 }
-#post_confirm_strip /* < That may be deprecated. */, #shortcuts {
+#post_confirm_buttons, #shortcuts {
 	padding: 12px 0;
 }
 #post_confirm_buttons .smalltext {
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -677,7 +677,6 @@ img.sort, .sort {
 	display: inline-block;
 	margin: 0 2px 0 1px;
 	font-size: 0.9em;
-	line-height: 1.9em;
 }
 /* Disable default focus outlines */
 .dropmenu a {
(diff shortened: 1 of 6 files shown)


SychO9:
could you also add a fix for this ?
there's an issue with the z-index set on **.popup_window**, it's value is **99** in *index.css*, but it's meant for mobile, that should be moved to *responsive.css* as it's creating issues like this

zindex

Gwenwyfar:
Will need to make a small change for mobile for the boardindex fix, but I'll wait until the responsive file revision is merged. 

And the menu fix is more of a placeholder, there's more to fix in the menu. Better do it all in one go.

Gwenwyfar:
The main idea was to remove the text shadow (it looks very bad inside selected text), but it looks like removing text shadow  also removes color for some weird reason, so I took the chance to add custom colors as well. Should keep it consistent and nicely contrasted with the theme. We can tweak the colors if needed.

Themes/default/css/index.css
```diff
@@ -735,6 +740,9 @@ img.sort, .sort {
 #main_menu {
 	margin: 0 0 4px 0;
 }
+#main_menu .popup_window, #main_menu .popup_container {
+	z-index: unset;
```
SychO9, line 743: This doesn't quiet fix it though :p, the point of having the z-index on popup container is because on mobile it's needed, just not needed on desktop, so just removing it from line and adding it to responsive.css would fix it

But I just remembered you probably don't want to edit responsive.css right now :/ so ..

**Edit**: nevermind I wasn't properly testing, just gotta remove z-index: 5; from 2659, since it's already set on responsive.css

Gwenwyfar: Yeah, was avoiding editing responsive.css.

Hmm, it looked ok on my testing. I'll give it a better look. And since Sesq is merging the responsive PR soon so we can get back to meddling with it :)

Gwenwyfar: Checked again and still looks ok. You use chrome, right? It normally has a different idea of how to work with z-index.

smfbigguy: This is how it looks on my site.
screenshot_2019-02-28 start new topic

Gwenwyfar: What browser/version is that? I just tested on chrome and it also looks ok. But I see you have not added the latest changes of this PR, the popup should be small as well.

SychO9: Well it looks okay on desktop but not on mobile, here's what I mean
onmob
Best way of fixing this is to remove **z-index: 99;** from **.pupup_window** and **_z-index: 5;_** from **_.popup_container_** in **index.css**, removing those doesn't affect the mobile view since those values are already set in **responsive.css**

and of course reverting the css change above

smfbigguy: Firefox65 64 bit

Gwenwyfar:
After the longer time away from SMF and with the time I've had to think things over, it's time I moved on.

Some things have changed for the better since I left, but there's such a huge gap to go that I don't want any further involvement. Good luck for everyone. Maybe some day if things change enough, I might consider returning; but right now, that day is only a distant possibility.

I'm going to close my open PRs because I don't want them in my list anymore, but anyone is free to copy them and submit as their own. Sarabada.

PR #5423, opened 2019-02-13
bbc_html permission fixes

### LINKED ISSUE #5417: Permissions error, html bbc
#### Description
From #5351 
```
Subquery returns more than 1 row
File: C:\Apache24\htdocs\Upgrader\Sources\ManagePermissions.php
Line: 2369
```

### Steps to reproduce
1. Create two membergroups with independant permissions
2. Give each group permission to admin forum **or** mange groups **or** manage perms
3. Go to manage permissions page

- Version/Git revision: 2e48b89

Sesquipedalian:
- Fixes #5417 
- Closes some loopholes that could have allowed membergroups to keep the bbc_html permission after the prerequisite permissions were taken away.

### DIFF
--- a/Sources/ManagePermissions.php
+++ b/Sources/ManagePermissions.php
@@ -2387,7 +2387,7 @@ function loadIllegalBBCHtmlGroups()
  */
 function removeIllegalBBCHtmlPermission($reload = false)
 {
-	global $context;
+	global $context, $smcFunc;
 
 	if (empty($context['permissions_excluded']['bbc_html']) || $reload)
 		loadIllegalBBCHtmlGroups();
--- a/Sources/ManagePermissions.php
+++ b/Sources/ManagePermissions.php
@@ -614,6 +614,7 @@ function SetQuickGroups()
 		if ($_POST['add_remove'] == 'clear')
 		{
 			if ($permissionType == 'membergroup')
+			{
 				$smcFunc['db_query']('', '
 					DELETE FROM {db_prefix}permissions
 					WHERE id_group IN ({array_int:current_group_list})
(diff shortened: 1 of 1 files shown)


Sources/ManagePermissions.php
```diff
@@ -2340,30 +2355,19 @@ function loadIllegalBBCHtmlGroups()
...
-	}
+	$context['permissions_excluded']['bbc_html'] = array(-1, 0);
 
 	$minimum_permissions = array('admin_forum', 'manage_membergroups', 'manage_permissions');
 
 	$request = $smcFunc['db_query']('', '
 		SELECT id_group
 		FROM {db_prefix}membergroups
-		WHERE id_group NOT IN (1, COALESCE((
+		WHERE id_group != 1 AND id_group NOT IN (
```
live627, line 2365: Did `id_group` `1` in the `NOT IN` not work?

albertlast: Nope that doesn't work,
you can't combine a single value and a subquery (which return more than one value) in on ``IN``.

Sesquipedalian: The way I had it before was creating an array with a integer and then a sub-array. In PHP the equivalent would have been `array(1, array(2,4,5,6))` when what we wanted was `array(1,2,4,5,6)`. Maybe I could have mucked about some more with `COALESCE` and the sub-query to get the same effect in the end, but this works cleanly.

albertlast: Alternative solution would be:
```
SELECT id_group
FROM {db_prefix}membergroups
WHERE id_group NOT IN (1, COALESCE((
WHERE id_group NOT IN (
	SELECT DISTINCT id_group
	FROM {db_prefix}permissions
	WHERE permission IN ({array_string:permissions})
		AND add_deny = {int:add}
	UNION
	SELECT 1
)
```
but i never see this style of sql command in smf

PR #5426, opened 2019-02-14
Resize the textareas instead of sceditor-container

### LINKED ISSUE #5370: Large smileys list overflows the entire textarea
If your smileys list is too big, it will overflow and cover the entire textarea when making a new post.

live627:
Fixes #5370

### DIFF
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -1371,7 +1371,7 @@ function Display()
 			'post_button' => $txt['post'],
 		),
 		// add height and width for the editor
-		'height' => '250px',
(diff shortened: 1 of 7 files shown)


DiegoAndresCortes:
this produces a strange behavior in the editor and limits the max-height

@live627 @Sesquipedalian can we test this one again

_removing **position: relative** will cause this among other things_
Screenshot_67

_the js added causes this_
Screenshot_69


Edit: the max-height issue I had was after I touched the files this is not a concern, only the other stuff is

PR #5431, opened 2019-02-14
Removed Topic starter to see non approved messages fixes #5159

### LINKED ISSUE #5159: normal member can see an unapproved message
#### Description
Hi,

member (not moderator, not permission)
can see an unapproved message
only in his own topic

### Steps to reproduce
1. 
2. 

- Version/Git revision:
- Database Type:
- Database Version:
- PHP Version:

albertlast:
since @SychO9  did the most work

issue: #5159

### DIFF
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -200,7 +200,7 @@ function Display()
 	}
 
 	// Short-cut to know if this user can see unapproved messages.
-	$approve_posts = (allowedTo('approve_posts') || $context['topicinfo']['id_member_started'] == $user_info['id']);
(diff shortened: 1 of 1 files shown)


live627:
uhh... 

I'm unsure of this change. I'll explain after testing this.

SychO9:
yea there might be more to it, I never said it was the only change needed :3

Sesquipedalian:
This part does need to be changed, yes, but if we only change this then other old bugs will come back. I'm going to review all the logic regarding unapproved posts and topics so that we can fix all of this properly.

PR #5432, opened 2019-02-14
Added permission_view sync logic to ManageMembergroups fixes #5428

### LINKED ISSUE #5428: Board Permissions from membergroup edit doesn't work
#### Description
https://www.simplemachines.org/community/index.php?topic=565051
When deciding which boards a membergroup can see from the membergroup's edit page, it doesn't work

Git Revision: 42da6b7

albertlast:
I'm not fully sure how to test this,
@SychO9 could u test this?
Issue: #5428

### DIFF
--- a/Sources/ManageMembergroups.php
+++ b/Sources/ManageMembergroups.php
@@ -474,6 +474,7 @@ function AddMembergroup()
 		{
 			// Only do this if they have special access requirements.
 			if (!empty($changed_boards[$board_action]))
+			{
 				$smcFunc['db_query']('', '
 					UPDATE {db_prefix}boards
 					SET {raw:column} = CASE WHEN {raw:column} = {string:blank_string} THEN {string:group_id_string} ELSE CONCAT({raw:column}, {string:comma_group}) END
@@ -486,6 +487,30 @@ function AddMembergroup()
 						'column' => $board_action == 'allow' ? 'member_groups' : 'deny_member_groups',
 					)
 				);
+
+				$smcFunc['db_query']('', '
+					DELETE FROM {db_prefix}board_permissions_view
+					WHERE id_board in({array_int:board_list}) 
+						AND id_group = {int:group_id} 
+						AND deny = {int:deny}',
+					array(
+						'board_list' => $changed_boards[$board_action],
+						'group_id' => $id_group,
+						'deny' => $board_action == 'allow' ? 0 : 1,
+					)
+				);
+
+				foreach ( $changed_boards[$board_action] as $board_id)
+					$insert[] = array($id_group, $board_id, $board_action == 'allow' ? 0 : 1);
+
+				$smcFunc['db_insert']('insert',
+					'{db_prefix}board_permissions_view',
+					array('id_group' => 'int', 'id_board' => 'int', 'deny' => 'int'),
+					$insert,
+					array('id_group', 'id_board', 'deny')
+				);
+			}
+
 		}
 
 		// If this is joinable then set it to show group membership in people's profiles.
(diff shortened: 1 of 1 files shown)


SychO9:
This is good for adding new groups, But it doesn't handle editing groups,
you need to do the same in the EditMembergroup() function but slightly different

instead of this query 
```
				$smcFunc['db_query']('', '
					DELETE FROM {db_prefix}board_permissions_view
					WHERE id_board in({array_int:board_list}) 
						AND id_group = {int:group_id} 
						AND deny = {int:deny}',
					array(
						'board_list' => $changed_boards[$board_action],
						'group_id' => $id_group,
						'deny' => $board_action == 'allow' ? 0 : 1,
					)
				);
```
use this right before looping through **array('allow', 'deny')**
```
			$smcFunc['db_query']('', '
				DELETE FROM {db_prefix}board_permissions_view
				WHERE id_group = {int:group_id}',
				array(
					'group_id' => (int) $_REQUEST['group'],
				)
			);
```

it's because we want to remove all the old deny values before inserting new ones

albertlast:
Okay this should handle the case now too.

Sources/ManageMembergroups.php
```diff
@@ -771,6 +797,14 @@ function EditMembergroup()
 			foreach ($accesses as $group_id => $action)
 				$changed_boards[$action][] = (int) $group_id;
 
+			$smcFunc['db_query']('', '
+				DELETE FROM {db_prefix}board_permissions_view
+				WHERE id_group = {int:group_id}',
+				array(
+					'group_id' => (int) $_REQUEST['group'],
+				)
+			);
```
Sesquipedalian, line 806: For the sake of security, it would be better if the DELETE query had a condition to make sure it only acted on boards that we know for certain are going to have new values inserted for them.

Sources/ManageMembergroups.php
```diff
@@ -814,6 +849,18 @@ function EditMembergroup()
 							'column' => $board_action == 'allow' ? 'member_groups' : 'deny_member_groups',
 						)
 					);
+
+					$insert = array();
+					foreach ( $changed_boards[$board_action] as $board_id)
+						$insert[] = array((int) $_REQUEST['group'], $board_id, $board_action == 'allow' ? 0 : 1);
+
```
Sesquipedalian, line 856: Please do the DELETE query here, with a `WHERE id_group = {int:group_id} AND id_board IN ({array_int:boards})` condition.

Sesquipedalian (changes requested):

Sources/ManageMembergroups.php
albertlast: This forearch loop run only for existing boards (where you got an allow or deny),
not for all boards,
exp. when a group lost his allow for board 2, the loop with var  $changed_boards[$board_action] will not included board id 2 since it got deleted

-> i can't place here the delete logic, because the part is only called when you add permission.

@SychO9 could explain it better.

Sources/ManageMembergroups.php
albertlast: see the msg below,
we don't know which boards got deleted,
we only know which got added.
And for the "added", we don't know if they are changed or not.

For make sure that the data 100% sync with the old format data,
is the way right here.

In the worst case had a group no access to boards,
but keep able to access admin side to fix things --> no special critcal checks needed.

Sesquipedalian: I suppose you are right.

My concern was that if we deleted all entries for a group and then the submitted data only set permissions for some boards for that group, it would be possible to clear the deny permissions so that a group that was formerly blocked from a certain board could potentially gain access to it again.

But since we can rely on the security checks to prevent random hackers from making these changes, I suppose I was just being paranoid. So never mind. 🙂

Sesquipedalian (approved):

Issue #5433, opened 2019-02-14
Issues with conversion to JSON


sbulen:
#### Description
We have already had reports of two users whose upgrades have frozen during the Convert to JSON step.  

Not sure what's gong on here yet, but seems to be a pattern.

### Steps to reproduce
1. Can't yet...   It's out there in the wild...   

### Additional information/references
Thread 1:
https://www.simplemachines.org/community/index.php?topic=565072.0

Thread 2:
https://www.simplemachines.org/community/index.php?topic=564909.msg4005686#msg4005686

sbulen:
I don't have the spider tracking on in my site, so to attempt to reproduce, I used some of the data provided in the forum from above:
![image](https://user-images.githubusercontent.com/23568484/52822958-20691380-3068-11e9-9c91-7b719e9482fd.png)

~~During the 2.0.15 => 2.1 upgrade, 2 of 3 rows were not converted to JSON.  My initial suspicion is that there is an issue with our unserialize logic, because online testers properly detect serialized content here:~~


~~The rows not converted cause malformed JSON errors in the app:~~


EDIT:  Red herring....  All rows above convert fine...

jdarwood007:
Are you able to get the data shown there so we can analyze it?

sbulen:
It's all from the thread shown above - here's a link to the specific post:
https://www.simplemachines.org/community/index.php?topic=565072.msg4007008#msg4007008

sbulen:
I have asked that user via PM to see if he can provide the full file somehow...

sbulen:
The user sent me his spider_hits file, 455K entries, and...

It runs just fine through the upgrader for me (WAMP environment).   It executes to completion in only a minute or two.  

Observation 1:  There are, in fact, a great # of url entries in the spider_hits table that do not get translated to JSON.  These all appear to be because the USER_AGENT field is very long, and gets truncated, as this was restricted to varchar(255) bytes in 2.0.    ~85K out of the 455K spider_hits entries are broken.  

These entries are corrupt & irrecoverable, in 2.0 & 2.1.  Viewing them in the UI generates malformed JSON errors, but otherwise do not seem to hurt anything.  

We may need to think of something better to do with these?  Warn users to prune the spider hits log?  Drop them in the upgrader?  Ignore them?  They are currently left in the serialized format in the spider_hits table.

Examples:
```
[...]

PR #5439, opened 2019-02-16
Write empty array for invalid serialized string

live627:
Tentative fix for #5433

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -3435,19 +3435,22 @@ function serialize_to_json()
 								if ($temp === false)
 									$temp = @safe_unserialize(fix_serialized_data($row[$col]));
 
-								if ($temp === false && $command_line)
+								if ($temp === false)
 								{
-									echo "\nFailed to unserialize " . $row[$col] . "... Skipping\n";
-								}
-								// If unserialize failed, it's better to leave the data alone than to overwrite it with an empty JSON value
-								elseif ($temp !== false)
-								{
-									$row[$col] = json_encode($temp);
+									if ($command_line)
+										echo "\nFailed to unserialize " . $row[$col] . "... Skipping\n";
 
-									// Build our SET string and variables array
-									$update .= (empty($update) ? '' : ', ') . $col . ' = {string:' . $col . '}';
-									$vars[$col] = $row[$col];
+									$temp2 = @json_decode($row[$col], true);
+									if (!empty($temp2))
+										$temp = array();
+									elseif ($command_line)
+										echo "... Skipping";
 								}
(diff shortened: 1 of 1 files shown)


other/upgrade.php
```diff
@@ -3435,19 +3435,22 @@ function serialize_to_json()
...
 									$temp = @safe_unserialize(fix_serialized_data($row[$col]));
 
-								if ($temp === false && $command_line)
+								if ($temp === false)
 								{
-									echo "\nFailed to unserialize " . $row[$col] . "... Skipping\n";
+									if ($command_line)
+										echo "\nFailed to unserialize " . $row[$col] . "... Skipping\n";
+
+									$temp2 = @json_decode($row[$col], true));
```
albertlast: Travis found here a ``)`` to mutch.

live627:
@sbulen may I have your opinion on this?

sbulen:
LGTM.   As near as I can tell, the data is shot & irrecoverable anyway...

other/upgrade.php
```diff
@@ -3435,19 +3435,22 @@ function serialize_to_json()
...
 								{
-									echo "\nFailed to unserialize " . $row[$col] . "... Skipping\n";
+									if ($command_line)
+										echo "\nFailed to unserialize " . $row[$col] . "... Skipping\n";
+
+									$temp2 = @json_decode($row[$col], true);
+									if (!empty($temp2))
+										$temp = array();
+									elseif ($command_line)
+										echo "... Skipping";
```
Sesquipedalian, line 3447: Checking if the value is already JSON is a good idea. But I think this would be better for this block:
```
$temp = @safe_unserialize($row[$col]);

if ($temp === false)
	$temp = @safe_unserialize(fix_serialized_data($row[$col]));

if ($temp === false)
	$temp = smf_json_decode($row[$col], true, false);

if ($temp === null)
{
	$temp = array();

	if ($command_line)
		echo "\nFailed to unserialize " . $row[$col] . ". Setting to empty value.\n";
}
...
```

Sesquipedalian (changes requested):

other/upgrade.php
live627: `fix_serialized_data` never returns `false`

Sesquipedalian: No, but `safe_unserialize()` does.

live627:
I see that this is now merged in. This probably still has the bug where invalid/truncated serialized strings end up as `false` instead of an empty array.

Sesquipedalian:
I added another commit to take care of that.

PR #5440, opened 2019-02-16
Prevent ad block plus from hiding the like button

live627:


SychO9:
I don't think removing the class is a good idea, it can be really useful for customizations
Also you probably have to edit the one in SSI too

Sesquipedalian:
Using a different class name would be better than using none at all. Maybe "smflikebutton" would get past AdBlock?

Sesquipedalian (changes requested):
Using a different class name would be better than using none at all. Maybe "smflikebutton" would get past AdBlock?

Sesquipedalian (approved):

PR #5443, opened 2019-02-17
Don't use the formated percent value for calculation

albertlast:
$activity['percent'] = formated output = bad for calculation stuff
$activity['posts_percent'] = float value = good for calculation stuff

issue: https://www.simplemachines.org/community/index.php?topic=565120.0


albertlast:
got tested: https://www.simplemachines.org/community/index.php?topic=565120.msg4007426#msg4007426

Sesquipedalian (approved):

PR #5453, opened 2019-02-19
change pm_ignore_list to text fixes #5446

### LINKED ISSUE #5446: pm_ignore_list new data type is too small
#### Description
From: https://www.simplemachines.org/community/index.php?topic=565144.0

> ```
> ---# Updating members pm_ignore_list
> ALTER TABLE {$db_prefix}members
> MODIFY COLUMN pm_ignore_list VARCHAR(255) NOT NULL DEFAULT '';
> ```
> I have a lot of members where they have an ignore list that contains sigificantly more that 255 chars,

albertlast:
since buddy_list is text
i see to high chance this could break on more places...

please had in mind that text datatype in mysql didn't support default values.

issue: #5446

### DIFF
--- a/other/install_2-1_mysql.sql
+++ b/other/install_2-1_mysql.sql
@@ -702,7 +702,7 @@ CREATE TABLE {$db_prefix}members (
 	new_pm TINYINT UNSIGNED NOT NULL DEFAULT '0',
 	alerts INT(10) UNSIGNED NOT NULL DEFAULT '0',
 	buddy_list TEXT NOT NULL,
-	pm_ignore_list VARCHAR(255) NOT NULL DEFAULT '',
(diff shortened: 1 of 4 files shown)


Sesquipedalian:
Is there a reason why we we couldn’t just change it from VARCHAR(255) to, say, VARCHAR(1023) or VARCHAR(4095) or something like that?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

albertlast:
like i said the counter field of pm_ignore_list is buddy_list and this is a text datatype too.
over 1k+ could lead us in the 64k issue of mysql
1k could be in cases to low

text fix all this points

Sesquipedalian:
Fair enough.

Sesquipedalian (approved):

Issue #4871, opened 2018-07-25
Use full editor toolbar in Quick Reply setting problem


gevv:
Hi, 

"Use full editor toolbar in Quick Reply" setting Only the smile sets are invisible 

![notm2](https://user-images.githubusercontent.com/5096676/43202981-a496f87c-9025-11e8-8d7f-66630a891e4d.png)

![notm](https://user-images.githubusercontent.com/5096676/43203196-3607ef50-9026-11e8-8f80-a0d54ad04225.png)

--------------------------------------------------------------
![marked](https://user-images.githubusercontent.com/5096676/43202988-a8455234-9025-11e8-800d-709f002a061e.png)

![marked2](https://user-images.githubusercontent.com/5096676/43202999-ad1a1ec0-9025-11e8-9757-4c4502352bee.png)

smfbigguy:
From what I have been told this is default for SMF 2.1

Gwenwyfar:
Could just rewrite the language string to be more clear.

smfbigguy:
Use smiley toolbar in quick reply  ??

illori:
actually if that is the only change, i would be tempted to just remove the option completely. but changing the text would be easier.

XinYenFon:
I think removing this extra not so needed setting would be best.

Gwenwyfar:
I only wonder how does this affect the server in the number of requests, when you have a large amount of smileys in a larger forum. Other than that I don't see any use for the setting either.

live627:
I removed the setting awhile back but put it back after someone complained. Maybe I broke it in my haste to restore it.

Gwenwyfar:
I'm reopening this as there's still the language issue.

woodholly:
So smiles disabled by default ?
Wasted 2 hours to find this issue and 3 hours to find where this idiotic "Use full editor toolbar in Quick Reply" located.
PS: /index.php?action=admin;area=theme;sa=reset; - down there.

PR #5456, opened 2019-02-20
Remove use_editor_quick_reply

live627:
A possible solution for #4871

The option currently only disables the smilies list. I decide to remove it, seems quite pointless.


Sesquipedalian:
Agreed.

Sesquipedalian (approved):

PR #5458, opened 2019-02-20
Insert quotes into the editor sanely

### LINKED ISSUE #5306: Quoting a post adds it on end of message
Quoted posts insert in the wrong place.

Click reply. Type your message,  Now assuming you want to add a quoted post above your text.
Nudge the text down a bit.  and then click above it where you want the quoted post to appear.

Scroll the page down to the post you want to quote, and click quote.  The quoted post seems to get tacked onto the end of your message instead of above it where you clicked.  Pretty sure it shouldn't work like that.

2.1 RC1
https://github.com/SimpleMachines/SMF2.1/compare/98d5cfd...release-2.1

live627:
Fixes #5306

### DIFF
--- a/Sources/Post.php
+++ b/Sources/Post.php
@@ -2711,9 +2711,6 @@ function QuoteFast()
 
 	$moderate_boards = boardsAllowedTo('moderate_board');
 
-	// Where we going if we need to?
-	$context['post_box_name'] = isset($_GET['pb']) ? $_GET['pb'] : '';
-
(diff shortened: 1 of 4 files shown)


Sesquipedalian:
Seeing more JavaScript moved out of the templates into the *.js files gives me the warm fuzzies. 😄

live627:
I would have done more, but then it would not be RC material.

XinYenFon:
Minimal JS usage in template is just making me soooo happy...

PR #5459, opened 2019-02-21
Shows empty categories if necessary on manage boards page

### LINKED ISSUE #5002: Global Moderator Right to create Category Issue

Sesquipedalian:
Fixes #5002 by showing all categories on the manage boards page, even if there are no boards visible to the user in that category.
Note that the manage_boards permission no longer implies access to all boards.

Also enables the integrate_pre_boardtree hook to do its job properly.


Arantor:
Strongly disagree with this approach as per comments on other issue.

PR #5461, opened 2019-02-22
Some Posts & Theme pages UI Adjustments

SychO9:


Gwenwyfar:
Looks good, as always :)

I wonder if it wouldn't be better to put "report to admin" under a "more" tab like in posts though. It takes quite a bit of space and attention.

Gwenwyfar (approved):

SychO9:
yea I think you're right, it does take space :/
I'll move it there

SychO9:
I wonder why no one has made a generic **quickbuttons** template with a hook to it, the layout is used in many places and all of them require manual edit to add button to :/

Gwenwyfar:
Yeah, there's too many things used in the theme that we could have a generic template for :/

There haven't been a lot of front end devs working on 2.1 for much of the development though, so that's one reason.

Sesquipedalian:
> I wonder why no one has made a generic **quickbuttons** template with a hook to it, the layout is used in many places and all of them require manual edit to add button to :/

Make it so

SychO9:
See #5918

PR #5472, opened 2019-02-28
 Simplifies logic for manage_boards permission giving access to all boards

### LINKED ISSUE #5002: Global Moderator Right to create Category Issue

Sesquipedalian:
Fixes #5002 
Closes #5459

This is an alternative to #5459. This PR continues with the change introduced in 2.1 that meant anyone with the manage_boards permission could see all boards, but also simplifies some logic regarding that behaviour. In particular, these lines in `build_query_board()` now grant anyone who can manage boards access to all boards:
https://github.com/SimpleMachines/SMF2.1/blob/1c530f0b0badea95818990ec2665d2c4bdb3ba4a/Sources/Subs.php#L6581-L6582

This approach has two advantages over the previous method:

1. It's simpler. Instead of needing to use workarounds in the `EditBoard2()` and `EditMembergroup()` functions in order to grant access to all boards to membergroups with the permission, this just cuts to the chase by checking for the permission in `build_query_board()`.
2. If the permission is later revoked from a membergroup, the group automatically reverts back to whatever board access it had before the permission was given. Under the previous method, if the permission was revoked after the workaround to explicitly grant access to all boards had succeeded, the membergroup would continue to have access to all boards until the admin edited its access.

In order to address the issue pointed out by @jdarwood007 in https://github.com/SimpleMachines/SMF2.1/pull/5396#issuecomment-465805040, this PR also adds some logic to the upgrader to remove the manage_boards permission from anyone who doesn't already have access to all boards. So, if the Global Moderator group has the permission on a 2.0.x forum, but does not have access to all boards in the forum, then the permission will be removed during the upgrade to 2.1. However, if the Global Moderator group does already have access to all boards, then it can retain the permission.

### DIFF
--- a/Sources/ManageBoards.php
+++ b/Sources/ManageBoards.php
@@ -663,15 +663,6 @@ function EditBoard2()
 					$boardOptions['deny_groups'][] = (int) $group;
 			}
 
-		// People with manage-boards are special.
-		require_once($sourcedir . '/Subs-Members.php');
-		$board_managers = groupsAllowedTo('manage_boards', null);
-		$board_managers = array_diff($board_managers['allowed'], array(1)); // We don't need to list admins anywhere.
-		// Firstly, we can't ever deny them.
-		$boardOptions['deny_groups'] = array_diff($boardOptions['deny_groups'], $board_managers);
-		// Secondly, make sure those with super cow powers (like apt-get, or in this case manage boards) are upgraded.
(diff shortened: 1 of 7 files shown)


Sesquipedalian:
@jdarwood007, @Arantor, thoughts?

Arantor:
I like it 👍

PR #5473, opened 2019-02-28
Fix split topics - Select posts to split UI

### LINKED ISSUE #5470: Split Topic - Select posts to split. | show post in the wrong order
#### Description
I got a test topic which got multi pages and
it start with:
grafik

When i go into selective split i start with this(the view is fine):
grafik

When i move now the the post with 8 (the second post):
grafik
you notice that the message with 1q move one down and "noch mal" get on the first place,
this "noch mal" is the first post on the next page of the topic.

And so it keep doing,
when i move the next post 9 to the right,
the next post get on the top:
grafik

The problem is not that it is shown,
the order is in my eyes wrong.

### Steps to reproduce
1.  selected post split
2.  move the second post (you could the first one too, but this case is not supported)

SychO9:
Fixes #5470

@albertlast can you confirm this fixes the issue ?


albertlast (approved):
fix the issue

PR #5474, opened 2019-02-28
Some changes to get normal work of integrate_pre_messageindex hook

dragomano:

### DIFF
--- a/Sources/MessageIndex.php
+++ b/Sources/MessageIndex.php
@@ -322,12 +322,14 @@ function MessageIndex()
 		$enableParticipation = false;
 
 	$sort_table = '
-		SELECT t.id_topic, t.id_first_msg, t.id_last_msg
+		SELECT t.id_topic, t.id_first_msg, t.id_last_msg' . (!empty($message_index_selects) ? (', ' . implode(', ', $message_index_selects)) : '') . '
 		FROM {db_prefix}topics t
 		' . (empty($sort_methods_table[$context['sort_by']]) ? '' : $sort_methods_table[$context['sort_by']]) . '
+		' . (!empty($message_index_tables) ? implode("\n\t\t\t\t", $message_index_tables) : '') . '
 		WHERE t.id_board = {int:current_board} '
 			. (!$modSettings['postmod_active'] || $context['can_approve_posts'] ? '' : '
 			AND (t.approved = {int:is_approved}' . ($user_info['is_guest'] ? '' : ' OR t.id_member_started = {int:current_member}') . ')') . '
+		ORDER BY is_sticky' . ($fake_ascending ? '' : ' DESC') . ', ' . $_REQUEST['sort'] . ($ascending ? '' : ' DESC') . '
 		LIMIT {int:maxindex}
 			OFFSET {int:start} ';
 
--- a/Sources/MessageIndex.php
+++ b/Sources/MessageIndex.php
@@ -88,8 +88,35 @@ function MessageIndex()
 		$context['unapproved_posts_message'] = sprintf($txt['there_are_unapproved_topics'], $untopics, $unposts, $scripturl . '?action=moderate;area=postmod;sa=' . ($board_info['unapproved_topics'] ? 'topics' : 'posts') . ';brd=' . $board);
 	}
 
+	// Default sort methods.
+	$sort_methods = array(
+		'subject' => 'mf.subject',
+		'starter' => 'COALESCE(memf.real_name, mf.poster_name)',
+		'last_poster' => 'COALESCE(meml.real_name, ml.poster_name)',
+		'replies' => 't.num_replies',
+		'views' => 't.num_views',
+		'first_post' => 't.id_topic',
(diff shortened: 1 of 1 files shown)


Sources/MessageIndex.php
```diff
@@ -326,7 +328,6 @@ function MessageIndex()
 		WHERE t.id_board = {int:current_board} '
 			. (!$modSettings['postmod_active'] || $context['can_approve_posts'] ? '' : '
 			AND (t.approved = {int:is_approved}' . ($user_info['is_guest'] ? '' : ' OR t.id_member_started = {int:current_member}') . ')') . '
-		ORDER BY is_sticky' . ($fake_ascending ? '' : ' DESC') . ', ' . $_REQUEST['sort'] . ($ascending ? '' : ' DESC') . '
```
albertlast: why you throw this out?

dragomano: This line does not affect the script but causes an error when using a hook.
Just test it.

albertlast: what kind of error?

albertlast: since ther is a limit with offset,
the result get changed by this removing.

Sesquipedalian: I'm curious about how this line caused an error while using a hook. But at any rate, it is useless since the very same line appears at the end of the main query below—which is where it belongs.

albertlast: like i said it's not the same,
so need more information what the issue is here.
the var name already mention,
the table should be sorted.

Sesquipedalian: Oh, yes, you are right, @albertlast. As you said, since there's a LIMIT and OFFSET in this subquery, an ORDER BY needs to be here. I'm not sure how I missed that.

Sesquipedalian: Maybe we can/should remove this one instead?
https://github.com/SimpleMachines/SMF2.1/blob/d25c92908fbbdeb9330f7bc35e92caec58616e11/Sources/MessageIndex.php#L362

albertlast: since we don't know what we try to fix,
it make no sense to guess.

Sesquipedalian: No, I don't mean as a fix for whatever error @dragomano has been getting. I mean that it doesn't seem like the second ORDER BY clause is necessary at all, since the subquery has already ordered the results.

albertlast: nope could be changed in the middle of the road.

dragomano: Ok, I have made other changes.

dragomano: > what kind of error?

> Unknown column 'rating' in 'order clause'

I just tried to add another kind of sorting using hooks, using a third-party mod's table and had got errors. The suggested changes eliminate these errors and at the same time do not change the logic of the engine ([test forum with these changes](https://dragomano.ru/smf21/index.php?board=1.0;language=english), pay attention on Rating column that work with any sorting and added via hook).

albertlast: Why not introduce two new var for the sort part?
I mean in normal case you don't want to add in the sort table column,
because it's affect the performance of the query.

Sources/MessageIndex.php
```diff
@@ -88,8 +88,35 @@ function MessageIndex()
...
+		'first_post' => '',
+		'last_post' => ''
+	);
+
+	// Bring in any changes we want to make before the query.
+	call_integration_hook('integrate_pre_messageindex', array(&$sort_methods, &$sort_methods_table));
+
 	// We only know these.
-	if (isset($_REQUEST['sort']) && !in_array($_REQUEST['sort'], array('subject', 'starter', 'last_poster', 'replies', 'views', 'first_post', 'last_post')))
+	if (isset($_REQUEST['sort']) && !in_array($_REQUEST['sort'], array_keys($sort_methods)))
```
live627, line 119: `if (!isset($_REQUEST['sort']) || !isset($sort_methods[$_REQUEST['sort']]))`

Sources/MessageIndex.php
dragomano: Example?

Sources/MessageIndex.php
Sesquipedalian: Actually, the subsequent code in this function does not necessarily want `$_REQUEST['sort']` to be set at this point. It is important that if it is set at all, it is set to a valid value, but not being set is also valid.

Sources/MessageIndex.php
albertlast: well an own set for the sort table query part

live627:
@Sesquipedalian what's the plan for this PR?

Sesquipedalian:
> @Sesquipedalian what's the plan for this PR?

For one of us to review it again and then merge it for RC3 if it is good. :)

Sesquipedalian (approved):

Sesquipedalian:
...And that review didn't take long to complete.

PR #5478, opened 2019-03-02
Fixes #5477

### LINKED ISSUE #5477: New message anchor doesn't show sometimes
From https://www.simplemachines.org/community/index.php?topic=565241.0

This has been aggravating to try and track down. And I'm not so sure that I came up with the right solution.

I've figured out how to reliably reproduce this. A great boon to this is to disable/comment out the query that writes to log_topic in Display.php.

-  User A posts a reply to a topic.
- User B then views that topic . The new anchor is not found (verify in dev console). Refresh the page to see it appear.
Things seem to be cached in the session, to clear that, go view another topic. Populate.php is very handy.

My solution is at https://www.simplemachines.org/community/index.php?topic=565241.msg4009132#msg4009132

Affected code: "seek method" whatever the hell. https://github.com/SimpleMachines/SMF2.1/commit/1178ca3f01b8bd95a3fde25181650a995c59b16e from #4226

live627:


Sesquipedalian:
Hey @live627, want to add a fix for #5548 to this PR?

live627:
No. That seems to be a different issue. I want one issue per PR.

Sesquipedalian:
So it is. I misread it entirely.

PR #5487, opened 2019-03-07
Theme

live627:


SychO9 (commented):
Okay so this does : ✔ Stop users from changing to other themes when they are not allowed to.

However if you switch **[Allow members to select their own themes.]** On but uncheck a theme in **[Themes that the user is able to select:]**
users are still able to switch to that theme using the url I mentioned on the related issue, I think the array of available themes for users is **$modSettings['knownThemes']**

Also if **[Allow members to select their own themes.]** is off, users can access the theme picking page, they can't change their theme ofc, but they can still see a list of all themes, ~~do we want that ?~~

Oh and there seems to be a Js issue on the picking page, when you select a variant from the default theme, the change made here #699 caused the issue.

live627:
> However if you switch [Allow members to select their own themes.] On but uncheck a theme in [Themes that the user is able to select:]
users are still able to switch to that theme using the url I mentioned on the related issue, I think the array of available themes for users is $modSettings['knownThemes']

I pushed fixes for those. I'll work on the rest tomorrow.

 I must also put some of this code into 2.0, it's a mess waiting to happen.

SychO9 (commented):
hmmm, with the additional changes a normal user can't select the forum's default theme, and can still select a theme which is unchecked in the admin panel, there's enableThemes that contains an array of enabled themes, and there's knownThemes an array of themes which users can switch to.

Also I just realized, if we stop users from accessing the pick theme page when they're not allowed to, doesn't that mean they can't change variant either ? unless there's an alternative way of changing variants, or unless they're not supposed to be able to change variants if they can't change themes.

live627:
> Also I just realized, if we stop users from accessing the pick theme page when they're not allowed to, doesn't that mean they can't change variant either ? unless there's an alternative way of changing variants, or unless they're not supposed to be able to change variants if they can't change themes.

I'm guessing that variants shouldn't be changed if theme switching is off because the link in the profile goes away.

Sources/Themes.php
```diff
@@ -920,62 +920,49 @@ function PickTheme()
...
+	// Have we made a decision, or are we just browsing?
 	if (isset($_POST['save']))
 	{
-		// Which theme?
-		foreach ($_POST['save'] as $k => $v)
-			$_GET['th'] = (int) $k;
+		checkSession();
 
+		$id_theme = (int) key($_POST['save']);
 		if (isset($_POST['vrt'][$k]))
```
SychO9: $k is undefined here so can't change variants, everything else is working as it should now
just gotta make sure that whatever changes made here don't recreate the issue #4527 is meant to fix

Sources/Themes.php
```diff
@@ -879,6 +879,24 @@ function EnableTheme()
...
+ * @return bool
+ */
+function canPickTheme($id_member, $id_theme)
+{
+	global $modSettings, $user_info;
+
+	return
+		allowedTo($user_info['id'] == $id_member ? 'profile_extra_own' : 'profile_extra_any')
+		&& (allowedTo('admin_forum') || $id_theme == 0 || in_array($id_theme, explode(',', $modSettings['knownThemes'])))
+		&& (!empty($modSettings['theme_allow']) || allowedTo('admin_forum'));
```
SychO9, line 897: I think it'd be good to also check that the theme isn't disabled, I was able to change to a theme that was disabled by the admin, although the layout remains as the default, as soon as the theme is enabled my layout changed to it.

SychO9 (commented):
Also maybe this is a little unrealated to the PR itself but, when a theme is unchecked from the **Themes that the user is able to select:** list, it gets unlisted in the **Theme Settings** area,

Themes.php lines **239** & **201**, I think it should be true

Sources/Themes.php
live627: thhat's what the in_array on line 896 should do. I just tried it and it works. Am I missing something?

live627: I have a feeling that you, as admin, are changing the theme of another user.

SychO9: Well here's the thing, **$modSettings['knownThemes']** contains the themes which users are allowed to switch to, but **$modSettings['enableThemes']** contains the enabled themes, in 2.1 the ability to completely disable a theme was introduced, you can do that from the Theme Settings admin area, so right now **canPickTheme()** returns true with a normal user id & a disabled theme for parameters
enableThemes

live627:
oh  get it now... a disabled theme can still be in knownThemes

live627:
So... once a theme is disabled, only manually editing the database can bring it back. This should be fixed IMO.

SychO9:
oh damn, that wasn't the case until I told you about the get_all_themes parameter, basically that function either shows themes which belong to **knownThemes** OR those which belong to **enableThemes**, and the admin area should show all of them regardless :/

live627:
ed0b61ff68f0f666f8ab622c13f73fa46203e038 fixes that, so all themes are listed in Theme Settings. Disabled themes are not shown in Manage and Install as always.

SychO9 (commented):
Great !

* Now should the board themes selection box show enabled themes ? I think not
https://github.com/SimpleMachines/SMF2.1/blob/59d83924079374b849b0e4a196fd6515385c5c17/Sources/ManageBoards.php#L579-L591
* Should an admin be able to switch to a disabled theme ?

\* live627 goes crazy..

live627:
1. Yes it should show enabled themes because a board theme may not be user selectable. Do you mean it should skip disabled themes? Yah, that it should.
2. No.

SychO9:
I actually meant it shouldn't show disabled themes but wrote enabled themes instead lol, anyway it's showing disabled themes currently, and the admin is able to switch to disabled themes as well

live627:
> * live627 goes crazy..

Effacing 18 clowns is even crazier.

SychO9 (approved):
Alright all looks good :thumbsup:

live627:
I think I found one last bug: Users can specify a theme in the URL that can't be user selected

SychO9:
yup I confirm

Gwenwyfar:
Do we really want to disallow admins from selecting disabled themes? By default they are allowed to do pretty much anything, and they might want to test it without anyone else being able to use it.

SychO9:
if they don't want anyone else to be able to use a theme then they should use the **Themes that the user is able to select** Option, otherwise what would be the difference between completely disabling a theme and hiding it from normal members.

Gwenwyfar:
Ah, okay. I must have missed that. 2.0 only had a single option for everyone that I can remember (Or I missed it there too).

PR #5488, opened 2019-03-07
Check if the ascending is changed between calls fixes #5486

### LINKED ISSUE #5486: Breaking topics with posts order option
#### Description
When you have a topic with for example 3 pages, leave the page open and edit
**Show most recent posts at the top in topic view** from **Look & Layout**, go to the topic(which you left open) click on a page item
```
Type of error: Critical 
Function: Display
 http://localhost/SMF21/index.php?topic=1.0
 C:/Apache24/htdocs/SMF21/Sources/Display.php (Line 935)
```

### Steps to reproduce
1. Open a topic with at least 3 pages
2. edit **Show most recent posts at the top in topic view** in a new tab
3. go back to the topic tab and change the page
Might wanna try a few times to reproduce this

albertlast:
I'm not 100% sure if a isset check should be placed too,
normaly the var should be setted in the same moment when $_SESSION['page_topic'] is setted.
So when $_SESSION['page_topic'] is not setted the $_SESSION['page_ascending'] var should be also not setted and
vice versa.

issue: #5486

### DIFF
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -864,7 +864,7 @@ function Display()
 	$ascending = empty($options['view_newest_first']);
 
 	// Check if we can use the seek method to speed things up
-	if (isset($_SESSION['page_topic']) && $_SESSION['page_topic'] == $topic)
+	if (isset($_SESSION['page_topic']) && $_SESSION['page_topic'] == $topic && $_SESSION['page_ascending'] == $ascending)
 	{
 		// User moved to the next page
 		if (isset($_SESSION['page_next_start']) && $_SESSION['page_next_start'] == $start)
@@ -1022,6 +1022,7 @@ function Display()
 	$_SESSION['page_next_start'] = $_REQUEST['start'] + $limit;
 	$_SESSION['page_current_start'] = $_REQUEST['start'];
 	$_SESSION['page_topic'] = $topic;
+	$_SESSION['page_ascending'] = $ascending;
 
 	$smcFunc['db_free_result']($request);
 	$posters = array_unique($all_posters);


Sources/Display.php
```diff
@@ -864,7 +864,7 @@ function Display()
 	$ascending = empty($options['view_newest_first']);
 
 	// Check if we can use the seek method to speed things up
-	if (isset($_SESSION['page_topic']) && $_SESSION['page_topic'] == $topic)
+	if (isset($_SESSION['page_topic']) && $_SESSION['page_topic'] == $topic && $_SESSION['page_ascending'] == $ascending)
```
Sesquipedalian, line 867: > I'm not 100% sure if a isset check should be placed too,
> normaly the var should be setted in the same moment when $_SESSION['page_topic'] is setted.
> So when $_SESSION['page_topic'] is not setted the $_SESSION['page_ascending'] var should be also not setted and
> vice versa.

It would be best to include `isset($_SESSION['page_ascending'])`, just as a precaution. Without it, bored troublemakers could flood a forum's error log by manipulating their browser's stored session variables.

Sesquipedalian (changes requested):

Sources/Display.php
albertlast: When you like I can do it,
But your mention way doesn't work. 
The session is saved on server side, 
When your are evil you switch your session id but then you got complete different package of seib vars.

Sesquipedalian: 🤦‍♂️ I'm not sure what I was thinking there. Of course that's correct. Never mind what I said before. 🙂

Sesquipedalian (approved):

PR #5489, opened 2019-03-08
optimize/speed up some queries in ip tracking

live627:
ref https://www.simplemachines.org/community/index.php?topic=565147.0


jdarwood007:
Could we avoid using raw format here?

live627:
no, why?

Sources/Profile-View.php
```diff
@@ -1886,14 +1886,15 @@ function list_getUserErrors($start, $items_per_page, $sort, $where, $where_vars
  */
 function list_getIPMessageCount($where, $where_vars = array())
 {
-	global $smcFunc;
+	global $smcFunc, $user_info;
 
 	$request = $smcFunc['db_query']('', '
-		SELECT COUNT(*) AS message_count
+		SELECT COUNT(id_msg)
 		FROM {db_prefix}messages AS m
```
albertlast, line 1890: Change the alias here from m to b would be better,
as you raw version.

Sources/Profile-View.php
```diff
@@ -1886,14 +1886,15 @@ function list_getUserErrors($start, $items_per_page, $sort, $where, $where_vars
  */
 function list_getIPMessageCount($where, $where_vars = array())
 {
-	global $smcFunc;
+	global $smcFunc, $user_info;
 
 	$request = $smcFunc['db_query']('', '
-		SELECT COUNT(*) AS message_count
+		SELECT COUNT(id_msg)
```
albertlast, line 1889: The most benefit you got by changing the join,
how mutch is the time dif here when you change from id_msg to * ?

Sources/Profile-View.php
```diff
@@ -1913,24 +1914,23 @@ function list_getIPMessageCount($where, $where_vars = array())
...
-	global $smcFunc, $scripturl;
+	global $smcFunc, $scripturl, $user_info;
 
 	// Get all the messages fitting this where clause.
-	// @todo SLOW This query is using a filesort.
 	$request = $smcFunc['db_query']('', '
 		SELECT
 			m.id_msg, m.poster_ip, COALESCE(mem.real_name, m.poster_name) AS display_name, mem.id_member,
 			m.subject, m.poster_time, m.id_topic, m.id_board
 		FROM {db_prefix}messages AS m
```
albertlast: Again change this to b would be better as the raw way.

albertlast (changes requested):
Change the alias to b or create a new query_see_message would be mutch more cleaner way in my eyes-
And i like to see what the diff is between  * and id_msg

albertlast:
I like the base idea to throw out boards join where is not needed,
but the way i don't like.

jdarwood007:
@live627 ,

Raw should be avoided if at all possible.  Its bypassing all the security checks that we do in queries when processing the string.  Not that the string you are using shouldn't already be clean, just safer to avoid usage of this to ensure we don't open up any security holes by assuming the data is cleaned already.  The PR by @albertlast seems like it will do this for us.

I think if this gets accepted, the changes by #5491 should include changing this and merging it as well.

live627:
This now depends on #5491. Once that gets merged, this can go in.

albertlast:
I waiting a test from you id_msg vs * benchmark. 
Because the SQL code get awkward,
It's should provide any benefit

live627:
you already saw it at https://www.simplemachines.org/community/index.php?msg=4009918

albertlast:
Based on the pr here,
i guess you not only change the column name you also changed the join stuff.

live627:
the join was still there in that test

removing the join brought the query down to ~40ms

albertlast:
which mysql version you got?
And can you please post explain of this two queries?

Issue #5492, opened 2019-03-08
Recent topic + attachment to many queries


albertlast:
#### Description
When you enable recent topic and got a normal(limited) user,
than every attachment in a message create 4 additional queries

### Steps to reproduce
1. enable recent messages
2. look how many queries are used on index page (refresh so long until you reached the lowest number)
3. add a post with attachment
4. go to the index again (like step 2) and refresh so long you get the lowest number of quries

The delta should be 4 queries and this number get higher per post with attachment

### Environment (complete as necessary)
- Version/Git revision:
- Database Type:
- Database Version:
- PHP Version:

### Additional information/references
```
      SELECT a.id_msg AS msg, m.id_topic AS topic, m.id_board AS board
      FROM smf_attachments AS a
         LEFT JOIN smf_messages AS m ON (m.id_msg = a.id_msg)
      WHERE id_attach = 34
      LIMIT 1
   in ...\Sources\Subs-Attachments.php line 1071, which took 0.00052714 seconds at 0.0657959 into request.

      SELECT MIN(bp.add_deny) AS add_deny
      FROM smf_boards AS b
         INNER JOIN smf_board_permissions AS bp ON (bp.id_profile = b.id_profile)
         LEFT JOIN smf_moderators AS mods ON (mods.id_board = b.id_board AND mods.id_member = 2)
         LEFT JOIN smf_moderator_groups AS modgs ON (modgs.id_board = b.id_board AND modgs.id_group IN (2, 7))
      WHERE b.id_board IN (92)
         AND bp.id_group IN (2, 7, 3)
         AND bp.permission IN ('view_attachments')
         AND (mods.id_member IS NOT NULL OR modgs.id_group IS NOT NULL OR bp.id_group != 3)
      GROUP BY b.id_board
   in ...\Sources\Security.php line 944, which took 0.00092292 seconds at 0.06647396 into request.

         SELECT
            a.id_attach, a.id_folder, a.id_msg, a.filename, a.file_hash, COALESCE(a.size, 0) AS filesize, a.downloads, a.approved, m.id_topic AS topic, m.id_board AS board,
            a.width, a.height,
            COALESCE(thumb.id_attach, 0) AS id_thumb, thumb.width AS thumb_width, thumb.height AS thumb_height
         FROM smf_attachments AS a
            LEFT JOIN smf_attachments AS thumb ON (thumb.id_attach = a.id_thumb)
[...]

PR #5493, opened 2019-03-08
Cache allowedTo value

albertlast:
Well this is an attempt to redurce the amount of queries produce by the allowedTo operation,
which can get high like in the issue #5492

Create a key(hash) based on the userid+input and safe the return value if the is a more complex check.

Fix partly #5492 instead of 4 quries per attachment we are now on 2.

### DIFF
--- a/Sources/Security.php
+++ b/Sources/Security.php
@@ -897,6 +897,7 @@ function checkSubmitOnce($action, $is_fatal = true)
 function allowedTo($permission, $boards = null, $any = false)
 {
 	global $user_info, $smcFunc;
+	static $prem_cache = array();
 
 	// You're always allowed to do nothing. (unless you're a working man, MR. LAZY :P!)
 	if (empty($permission))
@@ -924,6 +925,9 @@ function allowedTo($permission, $boards = null, $any = false)
 	}
 	elseif (!is_array($boards))
 		$boards = array($boards);
+
+	if (!empty($prem_cache[hash('md5', $user_info['id'].implode($permission).implode($boards).$any)]))
+			return $prem_cache[hash('md5', $user_info['id'].implode($permission).implode($boards).$any)];
 
 	$request = $smcFunc['db_query']('', '
 		SELECT MIN(bp.add_deny) AS add_deny
@@ -955,18 +959,24 @@ function allowedTo($permission, $boards = null, $any = false)
 				break;
 		}
 		$smcFunc['db_free_result']($request);
+		$return = $result;
 		return $result;
 	}
 
 	// Make sure they can do it on all of the boards.
-	if ($smcFunc['db_num_rows']($request) != count($boards))
-		return false;
+	if ($smcFunc['db_num_rows']($request) != count($boards) && empty($return))
+		$return = false;
 
-	$result = true;
-	while ($row = $smcFunc['db_fetch_assoc']($request))
-		$result &= !empty($row['add_deny']);
-	$smcFunc['db_free_result']($request);
+	if (empty($return))
+	{
+		$result = true;
+		while ($row = $smcFunc['db_fetch_assoc']($request))
+			$result &= !empty($row['add_deny']);
+		$smcFunc['db_free_result']($request);
+		$return = $result;
+	}
 
+	$prem_cache[hash('md5', $user_info['id'].implode($permission).implode($boards).$any)] = $result;
 	// If the query returned 1, they can do it... otherwise, they can't.
 	return $result;
 }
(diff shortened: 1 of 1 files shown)


Sources/Security.php
```diff
@@ -955,18 +959,24 @@ function allowedTo($permission, $boards = null, $any = false)
...
-		return false;
+	if ($smcFunc['db_num_rows']($request) != count($boards) && empty($return))
+		$return = false;
 
-	$result = true;
-	while ($row = $smcFunc['db_fetch_assoc']($request))
-		$result &= !empty($row['add_deny']);
-	$smcFunc['db_free_result']($request);
+	if (empty($return))
+	{
```
jdarwood007, line 972: Unless I read this wrong, the previous line may set this as false when it used to return false, but then we start processing this with $result = true and returning $result.  Potentially it could mean we get a false positive here.

Might be better to set $return to null, then check !is_null or $return !== null

albertlast: Yes empty is here wrong, 
I need to use here isset.

Sources/Security.php
```diff
@@ -897,6 +897,7 @@ function checkSubmitOnce($action, $is_fatal = true)
 function allowedTo($permission, $boards = null, $any = false)
 {
 	global $user_info, $smcFunc;
+	static $prem_cache = array();
```
Sesquipedalian: Minor issue: please rename `$prem_cache` to `$perm_cache` here and throughout.

Sources/Security.php
```diff
@@ -925,6 +926,9 @@ function allowedTo($permission, $boards = null, $any = false)
 	elseif (!is_array($boards))
 		$boards = array($boards);
 
+	if (isset($prem_cache[hash('md5', $user_info['id'].implode($permission).implode($boards).$any)]))
```
Sesquipedalian: Here and throughout, please use delimiters between each component of the hashed strings, including between the imploded array elements, like this:
```
return $perm_cache[hash('md5', $user_info['id'] . '-' . implode(',', $permission) . '-' . implode(',', $boards) . '-' . $any)];
```
This will prevent possible collisions if, say, $boards contains `12, 32` in one case and `1, 2, 32` in another case.

Sources/Security.php
```diff
@@ -925,6 +926,9 @@ function allowedTo($permission, $boards = null, $any = false)
...
+			return $prem_cache[hash('md5', $user_info['id'].implode($permission).implode($boards).$any)];
```
Sesquipedalian: Minor issue: This line is indented too much.

Sources/Security.php
```diff
@@ -955,18 +959,23 @@ function allowedTo($permission, $boards = null, $any = false)
...
 		}
 		$smcFunc['db_free_result']($request);
-		return $result;
+		$return = $result;
 	}
 
 	// Make sure they can do it on all of the boards.
-	if ($smcFunc['db_num_rows']($request) != count($boards))
-		return false;
+	if ($smcFunc['db_num_rows']($request) != count($boards) && !isset($return))
```
Sesquipedalian: Rather than ` && !isset($return)`, you can just use an `elseif` here.

Sources/Security.php
```diff
@@ -955,18 +959,23 @@ function allowedTo($permission, $boards = null, $any = false)
...
+	if (!isset($return))
+	{
+		$result = true;
+		while ($row = $smcFunc['db_fetch_assoc']($request))
+			$result &= !empty($row['add_deny']);
+		$smcFunc['db_free_result']($request);
+		$return = $result;
+	}
 
+	$prem_cache[hash('md5', $user_info['id'].implode($permission).implode($boards).$any)] = $result;
```
Sesquipedalian: Just for the sake of easier reading, please add another line of whitespace between this line and the following comment.

Sesquipedalian (changes requested):

Sources/Security.php
```diff
@@ -925,6 +926,11 @@ function allowedTo($permission, $boards = null, $any = false)
 	elseif (!is_array($boards))
 		$boards = array($boards);
 
+	$cache_key = hash('md5', $user_info['id'] . '-' . implode(',', $permission) . '-' . implode(',', $boards) . '-' . $any);
+
+	if (isset($perm_cache[$cache_key]))
+			return $perm_cache[$cache_key];
```
Sesquipedalian: Indented too much.

Sesquipedalian (changes requested):
Looks good. Just one last bit of whitespace to correct.

Sources/Security.php
albertlast: I don't understand what you mean.

Sesquipedalian: There are 3 tabs before this line when there should only be 2.

Sesquipedalian (approved):

live627 (approved):
Looking at this again, I see that my initial assessment was incorrect.

PR #5500, opened 2019-03-11
Fix search highlighting to not mangle/expose some HTML [topic 550840

live627:


live627:
Before
Untitled (Recovered)

After
Untitled

albertlast:
what is the [topic xxxxx] thing?

PR #5502, opened 2019-03-11
Doc fixes

Oldiesmann:
Various fixes to the docblock comments to make [apiGen](https://github.com/ApiGen/ApiGen) work properly. See http://www.oldiesmann.us/smfdocs/ for an example of the output and individual commits for more info.

This is still a WIP and more commits might be coming as I have more time to play with ApiGen


jdarwood007:
@uses is supposed to be valid: https://docs.phpdoc.org/references/phpdoc/tags/uses.html

Oldiesmann:
@jdarwood007 It is valid - just not the way we're using it.

jdarwood007:
Can we fix it properly by referencing things the right way then?

Oldiesmann:
> Can we fix it properly by referencing things the right way then?

Completely forgot about this until now. I'll play around with it again soon when I have time and see what I can do.

Oldiesmann:
Closing this one out as it's outdated. A lot has changed in the past year. Will resubmit a newer one when I have time.

live627:
Well, most of your changes were still valid.

PR #5515, opened 2019-03-13
Fix regular members board permission

### LINKED ISSUE #5514: board permissions, regular members
If you create a new board and you only give regular members access, they still won't see the board

Deny boards option is off
All groups are deselected but the regular members

SychO9:
Fixes #5514

regular members group id = 0
$boardUpdateParameters['member_groups'] is a string containing group ids seperated by commas, so if we only check regular members we get 0 which doesn't pass the test for empty()

### DIFF
--- a/Sources/Subs-Boards.php
+++ b/Sources/Subs-Boards.php
@@ -668,7 +668,7 @@ function modifyBoard($board_id, &$boardOptions)
 	);
 
 	// Do permission sync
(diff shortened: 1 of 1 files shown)


live627 (approved):

Sesquipedalian:
Since this involves access permissions, I'm going to go ahead and put in into RC2.

live627:
ok, wassn't sure  if this would mess with the release cycle

PR #5525, opened 2019-03-18
Tidy the UI structure's css code

SychO9:
@Gwenwyfar check it out when you got time


XinYenFon (approved):

Gwenwyfar (approved):
For whatever this is worth... Looks fine from a glance at the code. No longer have SMF to test on though.

pouvik:
This works fine however for keeping code clean  would it not be better for the existing #footer in responsive.css to become

```
#footer {
		padding: 10px;
		text-align: center;
	}
```

saves space and will do the same thing also stop duplication

SychO9:
no because the #footer you are referring to is not in the same media query block, so not the same scope.

pouvik:
was just a suggestion man thats all :)

PR #5543, opened 2019-03-26
Generic quickbuttons template

SychO9:
I'm not sure this is a good change at this stage, but that's for you devs to decide.

Here's what this PR does
* Adds a generic template for the quickbuttons layout and adds a hook for each set of quickbuttons array to enable manipulation without modifications.
* Removes the **create_button** function, it is only used in a few spots in the moderation section to remove icons using the **use_buttons** template setting.
* Removes the **use_button** template setting, there is no use in a setting that only makes a difference in a couple sections of the forum.

### DIFF
--- a/Sources/Recent.php
+++ b/Sources/Recent.php
@@ -454,9 +454,6 @@ function RecentPosts()
 		$context['posts'][$counter]['can_quote'] = $context['posts'][$counter]['can_reply'] && $quote_enabled;
 	}
 
-	// Allow last minute changes.
-	call_integration_hook('integrate_recent_RecentPosts');
-
 	// Last but not least, the quickbuttons
 	foreach ($context['posts'] as $key => $post)
 	{
@@ -482,6 +479,9 @@ function RecentPosts()
 			),
 		);
 	}
+
+	// Allow last minute changes.
+	call_integration_hook('integrate_recent_RecentPosts');
 }
 
 /**
--- a/Sources/Recent.php
+++ b/Sources/Recent.php
@@ -456,6 +456,32 @@ function RecentPosts()
 
 	// Allow last minute changes.
 	call_integration_hook('integrate_recent_RecentPosts');
+
+	// Last but not least, the quickbuttons
+	foreach ($context['posts'] as $key => $post)
+	{
+		$context['posts'][$key]['quickbuttons'] = array(
+			'reply' => array(
+				'label' => $txt['reply'],
+				'href' => $scripturl.'?action=post;topic='.$post['topic'].'.'.$post['start'],
+				'icon' => 'reply_button',
+				'show' => $post['can_reply']
+			),
+			'quote' => array(
+				'label' => $txt['quote_action'],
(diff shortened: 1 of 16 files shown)


Sources/Recent.php
```diff
@@ -456,6 +456,32 @@ function RecentPosts()
 
 	// Allow last minute changes.
 	call_integration_hook('integrate_recent_RecentPosts');
+
+	// Last but not least, the quickbuttons
+	foreach ($context['posts'] as $key => $post)
```
live627, line 458: This should be  before the hook.

Themes/default/ReportedContent.template.php
```diff
@@ -206,18 +180,27 @@ function template_viewmodreport()
...
+		),
+		'close' => array(
+			'text' => $context['report']['closed'] ? 'mc_reportedp_open' : 'mc_reportedp_close',
+			'url' => $scripturl.'?action=moderate;area=reportedposts;sa=handle;closed='.(int) !$context['report']['closed'].';rid='.$context['report']['id'].';'.$context['session_var'].'='.$context['session_id'].';'.$context['mod-report-closed_token_var'].'='.$context['mod-report-closed_token'],
+			'icon' => 'close'
+		)
+	);
+
+					// Report buttons 
+					template_button_strip($report_buttons, 'right');
```
Sesquipedalian: Nit: Please fix the indentation of these lines.

Themes/default/ReportedContent.template.php
```diff
@@ -499,18 +460,26 @@ function template_viewmemberreport()
...
+		),
+		'close' => array(
+			'text' => $context['report']['closed'] ? 'mc_reportedp_open' : 'mc_reportedp_close',
+			'url' => $scripturl.'?action=moderate;area=reportedmembers;sa=handle;closed='.(int)!$context['report']['closed'].';rid='.$context['report']['id'].';'.$context['session_var'].'='.$context['session_id'].';'.$context['mod-report-closed_token_var'].'='.$context['mod-report-closed_token'],
+			'icon' => 'close'
+		)
+	);
+
+					// Report buttons
+					template_button_strip($report_buttons, 'right');
```
Sesquipedalian: Nit: Please fix the indentation of these lines.

Sources/Subs.php
```diff
@@ -4407,32 +4407,6 @@ function text2words($text, $max_chars = 20, $encrypt = false)
...
- * Creates an image/text button
- *
- * @param string $name The name of the button (should be a main_icons class or the name of an image)
- * @param string $alt The alt text
- * @param string $label The $txt string to use as the label
- * @param string $custom Custom text/html to add to the img tag (only when using an actual image)
- * @param boolean $force_use Whether to force use of this when template_create_button is available
- * @return string The HTML to display the button
- */
-function create_button($name, $alt, $label = '', $custom = '', $force_use = false)
```
Sesquipedalian: Rather than deleting this function, it would be better to simply mark it as deprecated in its documentation block. That will better maintain backwards compatibility for mods and themes from 2.0.

Sesquipedalian (changes requested):

Sesquipedalian (approved):

PR #48, opened 2012-06-23
To save a few queries or whales?

live627:
! Save a couple of queries in the search (11) and quick mod11) by
combining permission  checks
- Added support for seperating board results by permissions

PR #5544, opened 2019-03-27
Fetch all four board permissions at once in Recent.php

live627:


live627:
This new syntax was introduced in #48

Sources/Recent.php
```diff
@@ -417,13 +417,16 @@ function RecentPosts()
 		)
 	);
 
+	// Create an array for the permissions.
+	$boards_can = boardsAllowedTo(array('post_reply_own', 'post_reply_any', 'delete_own', 'delete_any'), true, false);
```
albertlast: Can we not reuse the $permissions array?

live627: a good idea. Also will help with mod compatibility.

PR #5545, opened 2019-03-27
Fetch all board permissions once for the attach BBC

live627:
This will be my attempt at fixing #5492 

I'm not sure that computing a hash for every distinct permission check is a good idea, even if it's only md5. ref #5493. That's probably a bit much just to save a query or two.

### DIFF
--- a/Sources/Subs-Attachments.php
+++ b/Sources/Subs-Attachments.php
@@ -905,6 +905,10 @@ function assignAttachments($attachIDs = array(), $msgID = 0)
 function parseAttachBBC($attachID = 0)
 {
 	global $board, $modSettings, $context, $scripturl, $smcFunc;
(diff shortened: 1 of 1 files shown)


albertlast:
the #5493 solution depend of the amount of attachments,
in my case it's redurce the amount of quries by 17 not one or two.

live627:
And in my case it was 6: two attach codes in three posts each.

But yea, that was two queries per code saved. This also saves queries
across boards, meaning that only one permission query is called even if the
attach bbcodes are in different boards.

Sesquipedalian:
The general solution in #5493 and this more specific solution in #5545 both seem like good ideas to me. Since they appear to be fully compatible with each other, I think we'll merge both.

PR #5559, opened 2019-03-30
One last version update

### LINKED ISSUE #5558: Searching in SMF Admin gives error in log
RC2 

Searching in SMF Admin gives me a error in log, but the search still works.

Type of error: Critical 							
The selected search API the forum is using is out of date - falling back to standard search. Please check file SearchAPI-Standard.php.							 http://localhost/testrc2/index.php?action=admin;area=search
--

live627:
Fix #5558 
@Sesquipedalian may add commits to this branch as needed.


Sesquipedalian (commented):
This would fix the problem for now, but since the code involved is nearly useless to begin with, I'd rather solve the issue permanently as in #5563.

PR #5563, opened 2019-03-31
 Simplifies min/max version range for search and cache APIs

### LINKED ISSUE #5558: Searching in SMF Admin gives error in log

Sesquipedalian:
The previous way required us to update the `$version_compatible` strings in Class-SearchAPI.php and Class-CacheAPI.php with every release, in a misguided and ineffective attempt to future-proof against API changes.

Fixes #5558

### DIFF
--- a/Sources/Class-CacheAPI.php
+++ b/Sources/Class-CacheAPI.php
@@ -137,7 +137,7 @@ interface cache_api_interface
 	 * @access public
 	 * @return string the value of $key.
 	 */
-	public function getMiniumnVersion();
+	public function getMinimumVersion();
 
 	/**
 	 * Gets the Version of the Caching API.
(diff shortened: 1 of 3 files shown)


live627:
Should the version strings be constants?

Sesquipedalian:
> Should the version strings be constants?

To clarify, do you mean changing the `$version_compatible` and `$min_smf_version` variables into constants?

live627:
yes

Sesquipedalian:
Hm... probably not, no. It is possible that mods might need to change those values when extending the class. 

For example, maybe a change in SMF 2.1.7 will break the Sphinx search mod in a way that requires two versions of the Sphinx mod to be released: Sphinx_A that is compatible with 2.1 RC1 through 2.1.6, and Sphinx_B that is compatible with 2.1.7 through 2.1.999. That's not a very likely scenario, but not an impossible one.

MissAllSunday:
But they can redefine the constant if they extend the class and we use "static" when referencing those constants.

Now, I can argue than those vars are suppose to be changed, in that case a constant isn't the best case for that, better put some getters and setters for those vars to allow  configuration by whoever needs it.

live627 (approved):

PR #5566, opened 2019-04-01
Alerts system

### LINKED ISSUE #4950: [suggestion] Minor Alerts & Messages Suggestion
While browsing I noticed that when I receive a notification, I find myself clicking on the member's link instead of the reply to the topic/post itself (or other kinds on alerts such as reports) 

I'm thinking either having the first link be the important one, or have the entire box of each item take to the subject notified about, I do believe small details like this matter and make a difference.

### LINKED ISSUE #5408: Clicking Alert takes you to wrong place
If someone quotes one of my posts I get an alert.  Clicking that alert takes me to that exact post regardless of whether there is post before or after it.  Great, that's what I'd expect.

If someone posts to the topic, I get an alert telling me X posted a reply to that topic.. Clicking that alert takes me to the first post in that topic which means I have to go wade through all the posts to find it. It should take you to the post it's alerting you to, shouldn't it?

### LINKED ISSUE #5412: Alerts should show as read when read.
For example you have 42 alerts showing at top of screen. Surely clicking on that alert and viewing whatever it alerted you to, it should mark itself as read, instead of still saying you have 42 alerts it should say you have 41 alerts.  

You have to go to all alerts and mark them read manually.  This isn't right is it?

SychO9:
This is my proposed PR that closes #4950 and closes #5408 and closes #5412


Sources/Profile-View.php
```diff
@@ -446,13 +449,62 @@ function showAlerts($memID)
 
 	require_once($sourcedir . '/Profile-Modify.php');
 
+	// Are we opening a specific alert ?
+	if (!empty($_REQUEST['alert']))
```
albertlast, line 453: Do we need some kind of authority check here,
if he is allowed to open this?

SychO9: oh yes you're probably right, should check if the alert belongs to the member

Sources/Profile-View.php
```diff
@@ -446,13 +449,66 @@ function showAlerts($memID)
...
+			WHERE id_alert = {int:alert}',
+			array(
+				'alert' => $alert_id
+			)
+		);
+		$alert = $smcFunc['db_fetch_assoc']($request);
+		$smcFunc['db_free_result']($request);
+
+		// Can the user see this alert ?
+		if($context['user']['id'] != $alert['id_member'])
```
albertlast: I'm not 100% sure if this var and array is everytime setted and
on the other hand i was expecting that you use $memID.

SychO9: Huh, wonder what I was thinking there

SychO9:
I wonder if it'd be a good idea to use the og_image for an alert that doesn't have a sender avatar
ogloke

live627:
This didn't really fix #5408 since that link was not modified.

Problem is that the relevant alerts don't use the message id. Perhaps that could be fixed by using new: `?topic=6666.new#new`

SychO9:
really ? did you try with the PR's changes and then recreate a "X replied to Y " alert and see ? cause I'm pretty positive it fixes it

from what I've noticed, that type of alert has 2 links, one is the poster's, the other is the topic's,
however the alert has a **content_link** value in its extra db json array, the value of content_link is a correct link containing `.new#new` at the end, and since this PR prioritizes the value of content_link if it exists, that means that alert takes to the newest post

live627:
the popup is fine; the issue is in the "All alerts" page.

SychO9:
Oh this commit should fix that, it basically adds an icon with a link to the alert
https://github.com/SimpleMachines/SMF2.1/pull/5566/commits/6a25c5a987eebe2b2bee2ddde6a4259738127990

live627:
that link is generated in fetch_alerts()

https://github.com/SimpleMachines/SMF2.1/pull/5566/commits/3b172fbea4948c48072f3516de683df7c89d47de#diff-01cc3a672ae60fbabf662b0bb5727534R341

live627 (approved):
Excellent! I'm ready to merge this if you're finished with.

live627:
>I wonder if it'd be a good idea to use the og_image for an alert that doesn't have a sender avatar

I have no opinion on this. Could @Gwenwyfar weigh in on this?

SychO9:
One last thing, I'm seeing a 'icon' for alerts in the template file but nothing that actually creates that item on the source file, unless I'm missing something
https://github.com/SimpleMachines/SMF2.1/blob/e9b4fe9cd17257dc5ec43051eea5d6cb8da83833/Themes/default/Profile.template.php#L112

I wonder if we could ask @Arantor if he remembers anything :p

Arantor:
I remember thinking I wanted to support icons but in practice avatars were better so this could really go.

SychO9:
> I remember thinking I wanted to support icons but in practice avatars were better so this could really go.

Oh I see, it can go then thank you

PR #5571, opened 2019-04-04
Message icon checks enable option

dragomano:


MissAllSunday:
Just a thought, wouldn't it be better to have this setting as a $theme setting? this way themers can define if they want to use custom icons.  Unless there's already a setting for it, I haven't checked.

Or perhaps make it so if there's an icon on the current theme's folder, use it, otherwise use the default one.

PR #5575, opened 2019-04-05
Always parses BBC using the recipent's language in email notifications

### LINKED ISSUE #1407: Email notifications involving parsed BBC can be in the wrong language
There are several instances where BBC gets parsed before sending out notifications/announcements/whatever. It is always done so that parsing occurs before language files get reloaded (and in most cases, index.lang.php never gets reloaded anyway)

While minor, it does mean that if the sender and recipient have different languages, the sender's language is used, not the recipient's. E.g. the quote bbc uses 'Quote from: <name>' - but if the sender is English and the recipient is, say, German, the English text gets used.

Solution: when switching language, force reload index+Modifications for the new language and reset back to whatever language it was on beforehand.

Originally reported in 2006 (and still valid) in http://dev.simplemachines.org/mantis/view.php?id=52

Sesquipedalian:
Doing this requires using `parse_bbc()` in the langauge of the recipient of the email notification rather than in the language of the current user.

Main changes:
- `parse_bbc()` can now remember versions of the $bbc_codes array for multiple languages.
- For new post notifications, the body text of the message is now censored and parsed in CreatePost-Notify.php rather than in Subs-Post.php's `sendNotifications()`. This allows us to tailor the output of `parse_bbc()` depending on the recipient's language.
- Similarly, some code has been moved around and modified in the `sendpm()` function of Subs-Post.php.

~TODO: This code temporarily resets `$txt` to a blank slate, populates it with the contents of the index language file for the recipient's language, and then restores `$txt` to its original state when we are done. This is fine for parsing standard BBC, but it will know nothing about `$txt` strings stored in other files, e.g. in those added by mods. Should probably do something about that...~ ← Done.

Fixes #1407 (which dates all the way back to 2006: http://dev.simplemachines.org/mantis/view.php?id=52)

### DIFF
--- a/Sources/tasks/CreatePost-Notify.php
+++ b/Sources/tasks/CreatePost-Notify.php
@@ -211,8 +211,9 @@ class CreatePost_Notify_Background extends SMF_BackgroundTask
 
 				censorText($parsed_message[$receiver_lang]['subject']);
 				censorText($parsed_message[$receiver_lang]['body']);
+
 				$parsed_message[$receiver_lang]['subject'] = un_htmlspecialchars($parsed_message[$receiver_lang]['subject']);
-				$parsed_message[$receiver_lang]['body'] = trim(un_htmlspecialchars(strip_tags(strtr(parse_bbc($parsed_message[$receiver_lang]['body'], false), array('<br>' => "\n", '</div>' => "\n", '</li>' => "\n", '&#91;' => '[', '&#93;' => ']')))));
+				$parsed_message[$receiver_lang]['body'] = trim(un_htmlspecialchars(strip_tags(strtr(parse_bbc($parsed_message[$receiver_lang]['body'], false), array('<br>' => "\n", '</div>' => "\n", '</li>' => "\n", '&#91;' => '[', '&#93;' => ']', '&#39;' => '\'')))));
 			}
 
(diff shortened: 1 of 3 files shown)


live627:
Unrelated: I don't have access to Mantis. I thought it was closed.

Sesquipedalian:
Mantis is closed now, yes. That commit message was originally authored... well, a while ago.

Sources/tasks/CreatePost-Notify.php
```diff
@@ -194,18 +197,38 @@ public function execute()
 			else
 				continue;
 
+			$receiver_lang = empty($data['lngfile']) || empty($modSettings['userLanguage']) ? $language : $data['lngfile'];
+
+			// Censor and parse BBC in the receiver's language. Only do each language once.
+			if (empty($parsed_message[$receiver_lang]))
+			{
+				if ($receiver_lang != $user_info['language'])
```
live627: Err, I don't think that `$user_info` is available in cron.

Sesquipedalian: Derp. Although `$user_info` itself is, `$user_info['language']` isn't set.

Sesquipedalian: Fixed in 3d64b0e

PR #5578, opened 2019-04-05
Minor edits

SychO9:


Sources/Profile-View.php
```diff
@@ -480,13 +481,13 @@ function showAlerts($memID)
...
-		elseif (isset($alert['content_action'], $alert['id_member_started']) && $alert['content_action'] === 'buddy_request')
-			$link = $scripturl . '?action=profile;u=' . $alert['id_member_started'];
 		elseif (isset($alert['content_action']) && $alert['content_action'] === 'group_request')
 			$link = $scripturl . '?action=moderate;area=groups;sa=requests';
+		elseif (isset($alert['content_action'], $alert['id_member_started']) && ($alert['content_type'] === 'member' || $alert['content_action'] === 'buddy_request'))
+			$link = $scripturl . '?action=profile;u=' . $alert['id_member_started'];
 		elseif (isset($alert['content_type'], $alert['extra']['event_id']) && $alert['content_type'] === 'event')
 			$link = $scripturl . '?action=calendar;event=' . $alert['extra']['event_id'];
-		elseif (isset($alert['content_action'], $alert['content_id']) && $alert['content_action'] === 'like')
+		elseif (isset($alert['content_action'], $alert['content_id']) && $alert['content_type'] === 'msg')
```
Sesquipedalian: If you want to check that `$alert['content_type'] === 'msg'` here, it would be better to use this:
```
elseif (isset($alert['content_action'], $alert['content_id'], $alert['content_type']) && $alert['content_action'] === 'like' && $alert['content_type'] === 'msg')
```
```

Sesquipedalian (changes requested):

Sources/Profile-View.php
SychO9: Actually I should just change the content_action in the isset to content_type aswell, I don't want to check for 'like' content action because all msg type alerts take to the exact message and not just the 'like', that's why I changed it

live627 (approved):

PR #5583, opened 2019-04-06
Some fixes to announcement options

live627:
-  hide the checkbox in the registration form if users cannot disable announcements
- only save the pref if said modsetting is on
- populate that checkbox with the admin pref
- always show that pref in the admin area

### DIFF
--- a/Sources/Post.php
+++ b/Sources/Post.php
@@ -2704,6 +2704,7 @@ function AnnouncementSend()
 
 	foreach ($rows as $row)
 	{
+		$context['start'] = $row['id_member'];
 		// Force them to have it?
 		if (empty($prefs[$row['id_member']]['announcements']))
 			continue;
@@ -2730,7 +2731,6 @@ function AnnouncementSend()
 		}
 
 		$announcements[$cur_language]['recipients'][$row['id_member']] = $row['email_address'];
-		$context['start'] = $row['id_member'];
 	}
 
 	// For each language send a different mail - low priority...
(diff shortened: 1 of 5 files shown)


vort7u:
Thanks for these improvements, after the changes I have these errors when I check alerts
Screenshot_2019-04-07 Error Log

SychO9:
did you make the changes manually one by one or did you copy paste the files changed from live's `notify_announcements` branch ?

vort7u:
I copy and paste files changed from live's notify_announcements branch, errors appears only when a new user registration is notified.

SychO9:
I see, which means the error you're receiving isn't caused by this PR

vort7u:
Oops ... sorry :flushed:

SychO9:
oh no don't worry about it, just gotta make sure

vort7u:
I modified the files manually and everything works correctly without errors.
Thanks

Sesquipedalian:
Hm. This looks like a case where it would have been good to wait for a review from another dev before merging. 

Unfortunately, the approach taken here runs against the GDPR, which forbids sites from forcing users to receive announcements. The GDPR-compliant solution would be to correct any inconsistencies in the direction of saying that users can _always_ disable announcements—which was the reasoning behind prior coding decisions in 2.1, which have been aimed towards having GDPR compliance built in.

Of course, I am also aware that not all forum admins consider themselves to be subject to the GDPR and do want the ability to force users to receive announcements, so I am not necessarily convinced that entirely removing that ability would be the right move either. But if we want it to be easy for SMF 2.1 to comply with the GDPR while still visibly offering options to admins that would violate the GDPR, then we're going to need to build a "force GDPR mode" toggle, similar to the one we've put into 2.0.16, that will override any settings that would otherwise violate the GDPR. That'll take a bunch more work to implement than it would to simply remove non-compliant options like this.

PR #5589, opened 2019-04-10
Improves functionality of integrate_createAttachment hook

Sesquipedalian:
Previously, the integrate_createAttachment hook could be used to change existing values in $attachmentOptions, but it couldn't be used to handle custom values and columns. As a result, mods would be forced to make file changes if they wanted to populate a custom column they'd added to the attachments table. Now they can do that via the hook.

For safety's sake, these changes also add measures to prevent mods from breaking the standard behaviour if they screw up.

### DIFF
--- a/Sources/Subs-Attachments.php
+++ b/Sources/Subs-Attachments.php
@@ -696,39 +696,25 @@ function createAttachment(&$attachmentOptions)
 			$attachmentOptions['fileext'] = '';
 	}
 
-	// This defines the columns for the insert query.
+	// This defines which options to use for which columns in the insert query.
 	// Mods using the hook can add columns and even change the properties of existing columns,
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
I just thought of a more concise and elegant way to do this. Please, nobody get trigger-happy on merging this until I have time to implement it.

MissAllSunday:
Might as well mark this PR as "In progress" or "to do" can't remember the name github uses but that thing was actually created for cases like this one. 

Also the PR is flagged for final so I doubt anyone will merge it right now.

Sesquipedalian:
My understanding is that one can only mark a PR as a draft when it is first created. If there's a way to put a normal PR into draft status after it is created, I don't know what it is.

PR #5605, opened 2019-04-17
Alert glitches fixes

SychO9:
Each commit has an explanation of the fix.

### DIFF
--- a/Themes/default/Profile.template.php
+++ b/Themes/default/Profile.template.php
@@ -125,6 +125,8 @@ function template_alerts_popup()
 						ajax_indicator(false);
 						$("#alerts_menu_top span.amt").remove();
 						$("#alerts_menu div.alerts_unread").html(data);
+						if (typeof localStorage != "undefined")
+							localStorage.setItem("alertsCounter", 0);
 					}
 				);
 				return false;
--- a/Themes/default/scripts/alerts.js
+++ b/Themes/default/scripts/alerts.js
@@ -1,4 +1,15 @@
 var pingTime = 10000;
+var notificationIcon = '/favicon.ico';
+
+$(function ()
+{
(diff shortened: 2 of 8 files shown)


Themes/default/scripts/alerts.js
```diff
@@ -51,7 +51,7 @@ var updateAlerts = function ()
 		$.each(alerts, function(index, item)
 		{
 			var notification = notify.createNotification(new_alert_title, {
-				body: $(item).find('div.details:first > span').text(),
+				body: $(item).find('div.details:first span.alert_text').text(),
 				icon: smf_images_url + '/blank.png'
```
SychO9: Also, shouldn't we allow admin to choose an icon for the desktop notifications instead of always a blank image ?

Sesquipedalian: I'd say just use the OpenGraph image, falling back to the favicon.

Themes/default/Profile.template.php
```diff
@@ -125,6 +125,8 @@ function(data) {
 						ajax_indicator(false);
 						$("#alerts_menu_top span.amt").remove();
 						$("#alerts_menu div.alerts_unread").html(data);
+						if (typeof localStorage != "undefined")
+							localStorage.setItem("alertsCounter", 0);
```
MissAllSunday, line 129: What will happen if the browser doesn't support localStorage?

SychO9: If the browser doesn't support localStorage then there'll be no need to update the localStorage's alertsCounter value in the first place, because there'd be none

https://github.com/SychO9/SMF2.1/blob/ebbdad9f3806b1dc86fd912f9006dc0d5ee26478/Themes/default/scripts/alerts.js#L54-L55

Sesquipedalian (approved):

PR #5611, opened 2019-04-23
Calendar fixes

### LINKED ISSUE #4930: default view of calendar
current default is list view, i think this should be changed to month view. right now you go to the calendar after enabling it and think it may be broken due to showing nothing at all. the admin may not notice that list is highlighted and changing to month may make stuff show up. [this is given that there are no holidays in the given month like August]

### LINKED ISSUE #4931: change month view in calendar
if you leave the default as list and switch to month view then change the month drop down to the next month and click on view, it goes back to the list view to show you, it should stay in the current view.

### LINKED ISSUE #4932: per user change default of calendar view
i think it might not hurt to let the user select their own default for the calendar view, currently it looks like only the admin can set the default and have it set for everyone. some may like a week or list view instead of monthly and they can only switch for the current view of the calendar, when they move back to another page and back to the calendar the view does not stay.

### LINKED ISSUE #4933: date for calendar in list view does not follow user selected date format
i went in my profile and changed the date format, so the year would be last, this is reflected when showing the current time at the top of the forum. when in list view in the calendar it shows 2018-8-19 where it should show August, 19, 2018 or 8-19-2018. 

when you switch to the month view it has a drop down to select month and year, maybe worth using the same in the list view?

### LINKED ISSUE #4934: calendar settings in admin panel has no token
i thought all pages in the admin panel should have tokens so you cant go back to them and edit and save again. seems this may be missing for the calendar settings page.

### LINKED ISSUE #4936: Show Previous / Next Month Links hides weekly navigation in week view
if you disable the previous/new month links, the week view then looses its ability to move to another week, it will only show the current week with no navigation to another week.

### LINKED ISSUE #4937: weekly calendar view drop downs for month/year
this view should have a drop down to change the week that is viewed not just month/year. you can change to the next month but not the next week using the drop down. the drop down that we have in the list view currently should be present here so we can easily change which week is in view.

when we change the month currently it only shows the first week of that month which may include the last week of the prior month not the complete first week of the given month.

Sesquipedalian:
Fixes #4930 
Fixes #4931 
Fixes #4932 
Fixes #4933 
Fixes #4934 
Fixes #4936 
Fixes #4937

Also consolidates some redundant code (64df5cd and 2d75867), modernizes some code (d0a88cc), removes some unnecessary settings (618f323), and fixes an unreported bug in the upgrader (a898557).


Themes/default/languages/Themes.english.php
```diff
@@ -152,6 +152,6 @@
 
 // Open Graph
 $txt['og_image'] = 'Open Graph image';
-$txt['og_image_desc'] = 'Link to your Open Graph optimized image, suggested size 175x175px<br><span class="smalltext">You can read more about here <a href="http://ogp.me/">Open Graph</a></span>';
+$txt['og_image_desc'] = 'Link to your Open Graph optimized image, suggested size 175x175px<br><span class="smalltext">You can read more about Open Graph <a href="http://ogp.me/">here</a></span>';
```
MissAllSunday: Can we link the word "Open Graph" instead of the "here"?

Sesquipedalian: Sure.

MissAllSunday: I suggest something like:

read more about <a href="http://ogp.me/">Open Graph</a>

MissAllSunday (dismissed):

PR #5613, opened 2019-04-23
Retires old entries in log_packages when a package is upgraded.

### LINKED ISSUE #5612: Upgrading a mod does not mark previous version as uninstalled in log_packages table
#### Description
When the package manager executes an upgrade to a mod that already has an older version installed, it does not mark the old version as uninstalled in the log_packages table. This creates issues if the admin later uninstalls the new version of the mod.

### Steps to reproduce
1. Install an older version of a mod (e.g. [this one](https://github.com/SimpleMachines/SMF2.1/files/3109927/Paragrapher.tar.gz).)
2. Upload a new version of the mod package that includes update instructions in its package_info.xml file (e.g. [this one](https://github.com/SimpleMachines/SMF2.1/files/3109727/Paragrapher.tar.gz).)
3. Click the Upgrade button in the package manager and install the upgrade.
4. Look in the `log_packages` table in the database and see that both the old version and the new version of the mod are listed as currently installed. (Specifically, the `time_removed` column has a value of 0 and the `install_state` column doesn't.)
5. Go back to the package manager and uninstall the mod.
6. After uninstalling, observe how the package manager claims that the old version of mod is still installed, even though no version is actually installed.
7. Delete the mod in the package manager.
8. Check the `log_packages` table again, and see that the old version is still listed as installed.
9. Upload the new version of the mod package again.
10. Attempt to install the mod. Observe how the package manager wants to execute the upgrade instructions rather than the install instructions, despite the fact that the mod is not actually installed at all.

- Version/Git revision: e4f44b9
- Database Type:
- Database Version:
- PHP Version:

The package manager should probably mark any old versions of the mod as uninstalled when it successfully completes an upgrade to a new version of the mod.

Sesquipedalian:
Fixes #5612. Turned out to be a pretty straightforward fix.

### DIFF
--- a/Sources/Packages.php
+++ b/Sources/Packages.php
@@ -1100,8 +1100,8 @@ function PackageInstall()
 			{
 				$smcFunc['db_query']('', '
 					UPDATE {db_prefix}log_packages
-					SET install_state = {int:not_installed}, member_removed = {string:member_name}, id_member_removed = {int:current_member},
-						time_removed = {int:current_time}
+					SET install_state = {int:not_installed}, member_removed = {string:member_name},
+						id_member_removed = {int:current_member}, time_removed = {int:current_time}
 					WHERE package_id = {string:package_id}
 						AND id_install = {int:install_id}',
 					array(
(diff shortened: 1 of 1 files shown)


Sources/Packages.php
```diff
@@ -1119,6 +1119,23 @@ function PackageInstall()
 			{
 				$is_upgrade = true;
 				$old_db_changes = empty($row['db_changes']) ? array() : $smcFunc['json_decode']($row['db_changes'], true);
+
+				// Mark the old version as uninstalled
+				$smcFunc['db_query']('', '
+					UPDATE {db_prefix}log_packages
+					SET install_state = {int:not_installed}, member_removed = {string:member_name}, id_member_removed = {int:current_member},
```
albertlast: Some new lines to improve the read would be nice.

PR #5614, opened 2019-04-24
Remove port from default settings.php fixes #5608

### LINKED ISSUE #5608: Can't install from current HEAD using postgres
#### Description
This is probably a followup to #5584. Using the current git checkout, I can't connect to my (Postgres) DB at all during install. (Haven't tested MySQL because I don't have it installed.)

### Steps to reproduce
1.  Try to install using postgresql: port 5432, valid credentials with a trivial password like
```
$db_passwd = 'VejmZ8zs';
```
2. Installer fails with
```
Critical Error!
Cannot connect to the database server with the supplied data.

If you are not sure about what to type in, please contact your host.
```

The DB is running (other services are using it), so I don't think there's an issue there. I've tried several passwords, with and without special characters, including those that caused the problems in #5584.

I don't understand enough PHP to debug this further.

- Version/Git revision: c84bf54d547e3e2bc008bacac00048c8370725a5
- Database Type: Postgres
- Database Version: 9.4.21
- PHP Version: 5.6.40

albertlast:
The installer is not written in the way that the port stuff is handle well and
i got no reaction from @live627 

issue: #5608

### DIFF
--- a/other/Settings.php
+++ b/other/Settings.php
@@ -79,13 +79,6 @@ $cookiename = 'SMFCookie11';
  */
 $db_type = 'mysql';
 /**
- * The database port
- * Default options: 3306 for mysql, 5432 for postgresql
- *
- * @var int
- */
-$db_port = 3306;
-/**
  * The server to connect to (or a Unix socket)
  *
  * @var string
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -1321,17 +1321,6 @@ function UpgradeOptions()
 		if ($db_port != ini_get('mysqli.default_port'))
 			$changes['db_port'] = (int) $db_port;
 	}
-	elseif (!empty($db_port))
-	{
-		// If db_port is set and is the same as the default, set it to ''
-		if ($db_type == 'mysql')
-		{
-			if ($db_port == ini_get('mysqli.default_port'))
-				$changes['db_port'] = '\'\'';
-			elseif ($db_type == 'postgresql' && $db_port == 5432)
-				$changes['db_port'] = '\'\'';
-		}
-	}
 
 	// Maybe we haven't had this option yet?
 	if (empty($packagesdir))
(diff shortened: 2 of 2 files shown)


Sesquipedalian:
I don't think this is a good solution. The right solution would set the value of `$db_port` correctly when the user selects PostgreSQL in the installer.

This is the part of the code that I think needs to be adjusted:
https://github.com/SimpleMachines/SMF2.1/blob/f6eeb773ae5c05b0e21d0c00fae38d4e6d0f096f/other/install.php#L797-L805

I haven't tested this, but I think adding the following immediately after that if statement should solve the issue:
```
elseif ($db_type == 'postgresql')
	$vars['db_port'] = 5432;
```

albertlast:
First i asked me which issue got fixed by this change?
Secondly keep unix socket connection working?

Why we spend time on something which was working fine?

Sesquipedalian:
Oh, I see what you mean now. In that case, you should also remove the db_port changes made to upgrade.php in be26a66.

I also just noticed this bit of logic that doesn't make sense. We might as well fix that in this PR, too. It just needs to be changed to use the same structure as the code I quoted from install.php in my previous comment.
https://github.com/SimpleMachines/SMF2.1/blob/f6eeb773ae5c05b0e21d0c00fae38d4e6d0f096f/other/upgrade.php#L1324-L1334

albertlast:
I throw the complet logic out of the window,
when something is on default port,
we don't need to do anything in my pov.

live627:
My availability is very limited right now.

>  you should also remove the db_port changes made to upgrade.php in be26a66.

That will break the upgrader's ability to retain the port number when reformatting the settings file. (Is it "migrate"? I don't remember...)

albertlast:
I was guess that this came from:
https://github.com/SimpleMachines/SMF2.1/blob/be26a66cd5509dbe45cc6b525907cc43f71f7c74/other/upgrade.php#L5045-L5047

PR #5615, opened 2019-04-24
Reduce the amount of attachment queries fixes #5492

### LINKED ISSUE #5492: Recent topic + attachment to many queries

albertlast:
It's redurce the amount of to one for all attachment.

Idea is to "prepare" the attachment api with the needed data.

@Sesquipedalian i guess you should look into this.

issue: #5492

in my test env the amount of quries get down from 28 to 15.

### DIFF
--- a/Sources/Subs-Attachments.php
+++ b/Sources/Subs-Attachments.php
@@ -1101,16 +1101,15 @@ function getAttachMsgInfo($attachID)
 	{
 		if (empty($msgRows[$attachID]))
 			continue;
-		
+
 		$row = array(
 			'msg' => $msgRows[$attachID]['id_msg'],
 			'topic' => $msgRows[$attachID]['topic'],
 			'board' => $msgRows[$attachID]['board'],
 			);
-		
+
 		return $row;
 	}
-	
 
 	$request = $smcFunc['db_query']('', '
 		SELECT a.id_msg AS msg, m.id_topic AS topic, m.id_board AS board
@@ -1366,18 +1365,18 @@ function loadAttachmentContext($id_msg, $attachments)
 }
 
 /**
- * Gets attachment info associated with a message ID
+ * prepare the Attachment api for all messages
  *
- * @param int $msgID the message ID to load info from.
+ * @param int array $msgIDs the message ID to load info from.
  *
- * @return array.
+ * @return void.
  */
 function prepareAttachsByMsg($msgIDs)
 {
 	global $context, $modSettings, $smcFunc, $user_info, $cacheMsgID;
 	if(!isset($cacheMsgID))
 		$cacheMsgID = [];
-	
+
 	// remove all $msgIDs which we already cached
 	$msgIDs = array_flip(array_diff_key(array_flip($msgIDs), $cacheMsgID)) ;
 
@@ -1401,7 +1400,7 @@ function prepareAttachsByMsg($msgIDs)
 		$temp = array();
 		$rows = $smcFunc['db_fetch_all']($request);
 		$smcFunc['db_free_result']($request);
-		
+
 		foreach ($rows as $value)
 		{
 			if(empty($cacheMsgID[$value['id_msg']]))
--- a/Sources/Subs-Attachments.php
+++ b/Sources/Subs-Attachments.php
@@ -1106,7 +1106,7 @@ function getAttachMsgInfo($attachID)
 			'msg' => $msgRows[$attachID]['id_msg'],
 			'topic' => $msgRows[$attachID]['topic'],
 			'board' => $msgRows[$attachID]['board'],
-			);
+		);
 
 		return $row;
 	}
@@ -1146,7 +1146,7 @@ function getAttachsByMsg($msgID)
 	if (!isset($attached[$msgID]))
 	{
 		if (empty($cacheMsgID[$msgID]))
-			prepareAttachsByMsg (array($msgID));
+			prepareAttachsByMsg(array($msgID));
 
 		$temp = array();
 		$rows = $cacheMsgID[$msgID];
@@ -1358,6 +1358,7 @@ function loadAttachmentContext($id_msg, $attachments)
 function prepareAttachsByMsg($msgIDs)
 {
 	global $context, $modSettings, $smcFunc, $user_info, $cacheMsgID;
+
 	if(!isset($cacheMsgID))
 		$cacheMsgID = array();
 
@@ -1388,7 +1389,7 @@ function prepareAttachsByMsg($msgIDs)
 		foreach ($rows as $value)
 		{
 			if(empty($cacheMsgID[$value['id_msg']]))
-				$cacheMsgID[$value['id_msg']] = [];
+				$cacheMsgID[$value['id_msg']] = array();
 			$cacheMsgID[$value['id_msg']][$value['id_attach']] = $value;
 		}
 	}
--- a/Sources/Subs-Attachments.php
+++ b/Sources/Subs-Attachments.php
@@ -1146,24 +1146,8 @@ function getAttachsByMsg($msgID)
 	if (!isset($attached[$msgID]))
 	{
 		if (empty($cacheMsgID[$msgID]))
-			prepareAttachsByMsg ([$msgID]);
-/*
-		$request = $smcFunc['db_query']('', '
-			SELECT
-				a.id_attach, a.id_folder, a.id_msg, a.filename, a.file_hash, COALESCE(a.size, 0) AS filesize, a.downloads, a.approved, m.id_topic AS topic, m.id_board AS board, m.id_member,
-				a.width, a.height' . (empty($modSettings['attachmentShowImages']) || empty($modSettings['attachmentThumbnails']) ? '' : ',
-				COALESCE(thumb.id_attach, 0) AS id_thumb, thumb.width AS thumb_width, thumb.height AS thumb_height') . '
-			FROM {db_prefix}attachments AS a' . (empty($modSettings['attachmentShowImages']) || empty($modSettings['attachmentThumbnails']) ? '' : '
-				LEFT JOIN {db_prefix}attachments AS thumb ON (thumb.id_attach = a.id_thumb)') . '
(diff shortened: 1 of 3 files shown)


Sources/Subs-Attachments.php
```diff
@@ -1126,11 +1140,14 @@ function getAttachMsgInfo($attachID)
...
 {
-	global $context, $modSettings, $smcFunc, $user_info;
+	global $context, $modSettings, $smcFunc, $user_info, $cacheMsgID;
 	static $attached = array();
 
 	if (!isset($attached[$msgID]))
 	{
+		if (empty($cacheMsgID[$msgID]))
+			prepareAttachsByMsg ([$msgID]);
+/*
```
MissAllSunday: I know our minimum version supports short syntax for arrays [] but to keep some normalization and consistency with the rest of the code please use the old array format

Sources/Subs-Attachments.php
```diff
@@ -1092,11 +1092,25 @@ function getRawAttachInfo($attachIDs)
...
+	foreach ($cacheMsgID as $msgRows)
+	{
+		if (empty($msgRows[$attachID]))
+			continue;
+
+		$row = array(
+			'msg' => $msgRows[$attachID]['id_msg'],
+			'topic' => $msgRows[$attachID]['topic'],
+			'board' => $msgRows[$attachID]['board'],
+			);
```
Sesquipedalian: Please reduce indentation by one tab.

Sources/Subs-Attachments.php
```diff
@@ -1126,35 +1140,23 @@ function getAttachMsgInfo($attachID)
...
-				LEFT JOIN {db_prefix}messages AS m ON (m.id_msg = a.id_msg)
-			WHERE a.attachment_type = {int:attachment_type}
-				AND a.id_msg ' . (!empty($context['preview_message']) ? 'IN (0, {int:message_id})' : '= {int:message_id}'),
-			array(
-				'message_id' => $msgID,
-				'attachment_type' => 0,
-			)
-		);
+		if (empty($cacheMsgID[$msgID]))
+			prepareAttachsByMsg (array($msgID));
```
Sesquipedalian: Please remove unnecessary whitespace:
```
prepareAttachsByMsg(array($msgID));
```

Sources/Subs-Attachments.php
```diff
@@ -1346,4 +1348,50 @@ function loadAttachmentContext($id_msg, $attachments)
...
+			)
+		);
+		$temp = array();
+		$rows = $smcFunc['db_fetch_all']($request);
+		$smcFunc['db_free_result']($request);
+
+		foreach ($rows as $value)
+		{
+			if(empty($cacheMsgID[$value['id_msg']]))
+				$cacheMsgID[$value['id_msg']] = [];
```
Sesquipedalian:
```
$cacheMsgID[$value['id_msg']] = array();
```

Sources/Subs-Attachments.php
```diff
@@ -1346,4 +1348,50 @@ function loadAttachmentContext($id_msg, $attachments)
...
+ * @return void.
+ */
+function prepareAttachsByMsg($msgIDs)
+{
+	global $context, $modSettings, $smcFunc, $user_info, $cacheMsgID;
+	if(!isset($cacheMsgID))
+		$cacheMsgID = array();
+
+	// remove all $msgIDs which we already cached
+	$msgIDs = array_flip(array_diff_key(array_flip($msgIDs), $cacheMsgID)) ;
```
Sesquipedalian, line 1366: This works, but it would be more efficient to use the following.
```
$msgIDs = array_diff($msgIDs, array_keys($cacheMsgID));
```

Also, you could put an `else` before this line, because there is no need to do it at all if `$cacheMsgID` is empty.

Sources/Subs-Attachments.php
```diff
@@ -1346,4 +1348,50 @@ function loadAttachmentContext($id_msg, $attachments)
...
+ * prepare the Attachment api for all messages
+ *
+ * @param int array $msgIDs the message ID to load info from.
+ *
+ * @return void.
+ */
+function prepareAttachsByMsg($msgIDs)
+{
+	global $context, $modSettings, $smcFunc, $user_info, $cacheMsgID;
+	if(!isset($cacheMsgID))
```
Sesquipedalian, line 1362: Please add a line break above this line.

Sources/Subs-Recent.php
```diff
@@ -51,8 +51,29 @@ function getLastPosts($latestPostOptions)
 			'is_approved' => 1,
 		)
 	);
+
+	$rows = $smcFunc['db_fetch_all']($request);
+
+	// BBC and the entire attachments feature is enabled
+	$disabled = array();
+
+	$temp = !empty($modSettings['disabledBBC']) ? explode(',', strtolower($modSettings['disabledBBC'])) : [];
```
Sesquipedalian:
```
$temp = !empty($modSettings['disabledBBC']) ? explode(',', strtolower($modSettings['disabledBBC'])) : array();
```

Sesquipedalian (changes requested):
Two main changes requested:

1. When I ran a test of this code, I noticed that two very similar queries were run when displaying a topic with attachments: the new one in `prepareAttachsByMsg()` and another in Display.php here:
https://github.com/SimpleMachines/SMF2.1/blob/967873f3130809fda9dffe19055d01fd348aa317/Sources/Display.php#L1159-L1197
I think you can remove the code you've added to Display.php and instead just replace the query that already exists in Display.php with a call to `prepareAttachsByMsg()`.

2. The structure you use in your new `$cacheMsgID` variable is very similar to the one already being used for `$context['loaded_attachments']`. So, if you use `$context['loaded_attachments']` for this task, that would be more efficient on memory usage by avoiding duplicate data. It will also make it easier to implement my first requested change.

The rest of my change requests here are for minor fixes to formatting and style.

albertlast:
1. this is no new issue, this is already exists.
the code is written to be used in any places like in the recent thing
so i see no chance to change it,
like always feel free to change it after ward.

2. the big different is that here is not authority checks are done or
any other higher logic, it's hold only the data for the case that it's needed.
So i could move the data to context but not into the key.

Sources/Subs-Attachments.php
albertlast: ``$msgIDs = array_diff($msgIDs, array_keys($cacheMsgID));``
since the $cacheMsgID is possibe bigger of memory footprint and got a more complex structur,
i choose the other way.

Sesquipedalian:
> so i see no chance to change it,
> like always feel free to change it after ward.

I did, and then I found more efficiencies and some code that we could remove, so I did that too in a3460e3 before merging.

albertlast:
Yes you changes make sense,
beside of the sort thing,
but for this you need a "bigger" knowledg about attachment stuff which i doesn't had and
the scope of the pr over extend.

PR #5627, opened 2019-04-26
More hooks

Sesquipedalian:
Because @Gwenwyfar, @SychO9, and @kindred-999 are just crazy for them. 😉


Sesquipedalian:
They're just hooks. Imma merge 'em now.

Issue #5584, opened 2019-04-07
Installer should escape passwords


m4z:
#### Description
When running the installer, entering a database password containing `'` breaks the install process with various syntax errors, depending on the actual password. For example, using a 64-character KeePassXC-generated password starting with `$t&jz_vkCS$H*(32YsN~2D4'>JZur}$uAD$UjhX` resulted in:
```
==> /var/log/lighttpd/error.log <==
2019-04-07 19:33:43: (mod_fastcgi.c.2702) FastCGI-stderr: PHP Parse error:  syntax error, unexpected '}' in /home/some-username/public_html/Settings.php on line 104
```

Choosing the 1-character password `'` results in:
```
==> /var/log/lighttpd/error.log <==
2019-04-07 19:56:07: (mod_fastcgi.c.2702) FastCGI-stderr: PHP Parse error:  syntax error, unexpected '';' (T_CONSTANT_ENCAPSED_STRING) in /home/some-username/public_html/Settings.php on line 110
```

### Steps to reproduce
1. During installation, enter any database passwort containing `'`
2. Enjoy a beautiful "HTTP ERROR 500"

### Environment (complete as necessary)
- Version/Git revision: 2.1rc2 (on Debian 8.11 with lighttpd-1.4.35)
- Database Type: postgresql-9.4, but probably affects all
- Database Version: 9.4, but probably affects all
- PHP Version: 5.4, but probably affects all

### Additional information/references
Probably related to #2896.

Thanks for this otherwise (probably) great software! (Haven't yet managed to finish my first install. :wink: )

m4z:
This seems to be more complex: Even if avoiding `'`, other special characters don't seem to be properly escaped, because I could not successfully connect with various passwords containing special characters.
I have chosen a password without special characters for now (which worked fine), but will post more details in a few days.

live627:
I added a commit https://github.com/live627/SMF2.1/commit/1c85fd040d5873bdd5204eec31b403debff7517d that should addressees this. Can you say which special characters don't work?

m4z:
After reading the docs on PHP single quoted strings and a bit of testing, I think the culprit breaking the database login was `\`.

To expand on my second problem: If I use a password containing a backslash, the password is entered into `Settings.php` correctly:
```
[...]

PR #5628, opened 2019-04-26
Allow \ and ' be part of db pw for pg

albertlast:
i was not able to get a test case running for mysql,
so i need to believe the @live627 fix works for mysql.

Maybe to do this every time is not best way,
so would be better to place the convertion once in install process,
so that we got the right format directly in the settings.php

### DIFF
--- a/Sources/Subs-Db-postgresql.php
+++ b/Sources/Subs-Db-postgresql.php
@@ -72,6 +72,9 @@ function smf_db_initiate($db_server, $db_name, $db_user, $db_passwd, &$db_prefix
 	if (!function_exists('pg_pconnect'))
 		display_db_error();
 
+	// We need to escape ' and \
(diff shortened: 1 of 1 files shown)


Sesquipedalian (approved):

Sesquipedalian:
> Maybe to do this every time is not best way,
> so would be better to place the convertion once in install process,
> so that we got the right format directly in the settings.php

I wouldn't worry about it. `str_replace()` is not an expensive function, so it's better in this case to make things easier for the admin by keeping string in Settings.php as human-friendly as possible.

Sesquipedalian:
Am I correct in thinking that this PR should be combined with #5614 as part of fixing #5608?

albertlast:
For me is the issue: #5608 only port stuff,
where i had to update the upgrade.php into #5614 

This pr would fix: #5584 but this issue got already closed.

PR #5635, opened 2019-04-28
Update Help.english.php

m4z:
PostgreSQL also has FTI.
I don't know if the technical details all apply, though.


m4z:
Apparently the GH WebUI inserts a trailing newline.

SychO9:
Yea you have to use a different editor, also you should probably include all edits in one single PR

m4z:
Ok, will do.

m4z:

PR #5637, opened 2019-04-28
Update Help.english.php

m4z:
Add missing breaks for uniformity.


Gwenwyfar:
Thanks for taking the time to fix these :)

Same as mentioned in the other issue though, you can make a single PR with all language fixes, if possible.

m4z:

PR #5641, opened 2019-04-29
Completely hide "Users Online" section from Info Center when show_who…

### LINKED ISSUE #5632: Misnomer $modSettings['who_enabled'] and incorrect $helptxt['who_enabled']
This seems to be incorrect. The option only controls the latter option.

If it is disabled, the Info Center on the front page will still show "Users Online"
* how many guests and members are online, and
* the names of online registered users.

However, the former isn't a link anymore, so you can't find out what individual users are doing.

I proprose to either
* also hide the info center section "Users Online" if this setting is disabled, or
* find a more proper name and description for it

### Steps to reproduce
1. Read Help.*.php

- Version/Git revision: HEAD

m4z:
… is disabled

Fixes #5632 (does what the help text seems to imply)


SychO9:
that setting is only for the online list, if we go down the road of hiding online status then we'd also have to apply this everywhere including the online status on topics and profiles,

So I think the help text needs to be edited to reflect what the setting actually does, not the other way around

m4z:
Individual users can choose to turn off their Online status (if the admin allows it), or are you talking about more?
I'm a big fan of disabling such... let's call it "statistics" instead of "user tracking features". 😉 Especially for closed forums, I don't want to leak such details to "externals".

(But if this isn't in line with the vision of SMF, I'll try putting this into a mod instead and will include a more concise text for this issue.)

m4z:
PS: It would have been nice if you (or anybody) would have said which option they prefer in the issue before I spent all evening with PHP. 😉 The text option would've been "a few hours easier".

SychO9:
Well there are 101 issues you know, we can't tackle them all at once, we try to figure them out by order of their priority (milestones) and it isn't easy when the people contributing don't have much free time.

in any case, what I said is only an opinion, please don't take it as a "this is what has to be done" 😅 

> PS: It would have been nice if you (or anybody) would have said which option they prefer in the issue before I spent all evening with PHP. 😉 The text option would've been "a few hours easier".

We really do appreciate the help, the more people the better, but please don't be rude about it, I'm sorry you spent a lot of your time on it, but that's what we all do here as well.

m4z:
I understand that and didn't mean to be rude at all (it was a learning/refresher experience anyway and *I* chose to do it, so no harm done); I just mean any dev could have voiced an opinion (earlier) in the issue (before I began investing effort in one of the solutions). Trying to help, but when there's no comment at all, I'm just flying blind and hitting more things than necessary. 🚀

m4z:
This is superseded by #5852, but either I don't know the GitHub magic word for that or such a feature doesn't exist. :wink: 

I'd like to keep this around though, and work it into a full-fledged "Disable the Info Center" feature.

PR #5642, opened 2019-04-30
Fixing seeking by enabled descanding mode fixes #5561

### LINKED ISSUE #5561: Seek doesn't work by enabled descanding
Like mention in the pr: https://github.com/SimpleMachines/SMF2.1/pull/5488

Ther is actually another issue,
when ascending is enabled,
the seek doesn't work.
Since we got checks implementated the user didn't notice the issue.

I don't want to mix up this different issue in one pr and on the other hand,
i don't want to forget this issue.

albertlast:
I try to make to code a little bit easier,
in the way that $_SESSION['page_last_id'] got always the bottom message and
$_SESSION['page_first_id'] got always the top message id.

No strugle here anymore,
which one i had to take.

Second problem was in non seeking code the ascending got inverted 979,
leads me to the needs of $DBascending to get logic again easier i renamed $ascending_seeking to the same name
since it got the same meaning.

Thirt is that the $DBascending depends on the $ascending and $start_char,
in old code it's only depends on $start_char.

Issue: #5561

### DIFF
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -869,19 +869,19 @@ function Display()
 		if (isset($_SESSION['page_next_start']) && $_SESSION['page_next_start'] == $start)
 		{
 			$start_char = 'M';
-			$page_id = $ascending ? $_SESSION['page_last_id'] : $_SESSION['page_first_id'];
+			$page_id = $_SESSION['page_last_id'];
 		}
 		// User moved to the previous page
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Looks good, except for one thing. Instead of `array_values($messages)[0]`, please use `reset($messages)`. This is both more efficient and better for compatibility with PHP 5.x.

albertlast:
Well this was not fully in scope of this pr,
but i added this change to the code.

Sesquipedalian (approved):

PR #5643, opened 2019-05-01
Cache settings upon install & upgrade

### LINKED ISSUE #5520: Default cache settings upon install
They could be disabled for new installs. It is very confusing to have your fresh installed forum and the settings take long to be saved. And it won't be a needed option for most users until their forums are big enough to need it.

sbulen:
Fixes #5520 

Changes:
1. For installs, sets cache_enable in settings table to 0 (no cache), in sync with the parallel setting in Settings.php.  
2. For upgrades, migrates cache settings from the settings table to Settings.php where they are now.  Previously, the cache settings were lost upon upgrade.  
3. For upgrades, deletes unused table entry for cache_memcached.  

Notes:
- Yes, cache_enable is in two places, in Settings.php & in the settings table.  ACP keeps them in sync going forward.  I am assuming this was intentional...
- There is no real accelerator setting in 2.0 to migrate.  The only one we know for sure exists is smf file-based caching.  So if cache_enable is set, that is used.  
- What this means is users who have more advanced configs in 2.0 must revisit their cache settings after upgrade.


sbulen:
Related issue:
https://www.simplemachines.org/community/index.php?topic=567762.0

Sesquipedalian:
> * Yes, cache_enable is in two places, in Settings.php & in the settings table.  ACP keeps them in sync going forward.  I am assuming this was intentional...

Honestly, whether intentional or not, that doesn't make much sense to me. To have two synced settings is to have one setting too many. 

Moreover, this line in `reloadSettings()` means the value stored in the database is ignored anyway:
https://github.com/SimpleMachines/SMF2.1/blob/8900b7e84e5f4b1bb9b79125c1dd4686461ddbfc/Sources/Load.php#L89

Can I get you to go through the codebase and replace all occurrences of `$modSettings['cache_enable']` with `$cache_enable`, @sbulen? For the sake of backwards compatibility with mods we should probably keep that one line I linked to above, but otherwise please be ruthless. 😉 

Leaving that aside, the actual content of this PR so far looks great. 🙂

sbulen:
Will do.

Sesquipedalian:
Thanks.

sbulen:
Done.  The only line that had me scratching my head was the @ in subs_db_mysql...  Pretty sure it's fine - works today - just... looks different...

Sesquipedalian (approved):

Sesquipedalian:
> Done. The only line that had me scratching my head was the @ in subs_db_mysql... Pretty sure it's fine - works today - just... looks different...

You mean the way [this line](https://github.com/SimpleMachines/SMF2.1/pull/5643/commits/e017cb1a9a2e4f5058f67b940c64b33cbcfb040c#diff-40a16e287462d60f366698be5e160838R592) uses a string instead of an integer? Yeah, that does look odd. It won't hurt anything since PHP handles the type conversion automatically, of course, but it should properly be an integer. If you feel like fixing it, @sbulen, feel free to do so.

sbulen:
No I meant the @ in the simple assignments:
`$old_cache = @$cache_enable;`
I don't think I've seen that used elsewhere.  Since @ suppresses errors, I was wondering if this is an alternative way of checking if it exists or not; where normally we use isset() or empty()?  
I wasn't sure it was worthwhile to start pulling on that thread...

sbulen:
To be clear - I'm leaving this as-is!

Sesquipedalian:
Oh, right. I misread your comment. Yeah, that's not a convention I've seen elsewhere in our code. But it does seem to work as a terse alternative to `$old_cache = isset($cache_enable) ? $cache_enable : null;`. It has never occurred to me before to use `@` in a simple assignment this way. It's kind of a nice little trick, actually...

Sesquipedalian:
Anyway, merging now. Thanks for this one, @sbulen.

PR #5646, opened 2019-05-03
Member search time warp

### LINKED ISSUE #5274: Member search does not work properly - date mismatch?
#### Description
When using admin search, searching by registration date may not work correctly. Last active also appears to have the same problem - they don't work for me.

### Steps to reproduce
1. Install RC1 from beta board.
2. Register admin account during registration - the date registered as far as the platform is reporting is today, 25th December.
3. Go to admin > members > search for members
4. Try to search for any members registered on that date, 25/12/2018

image

5. No results are found - presumably because it's trying to match on yyyy-mm-dd format but the HTML5 datepicker doesn't automatically submit it in that format...

- Version/Git revision: RC1 build
- Database Type: MySQL
- Database Version: 10.1.22-MariaDB
- PHP Version: 7.0.18

Interestingly, if I edit the date on the user, moving it forward to 26/12/2018 and then back to 25/12/2018, search subsequently works correctly. Something about the initial creation may not work properly - if I either register or create a member from the admin panel today, this doesn't find them.

sbulen:
Fixes #5274 

The problem is that the comparisons were all against fields in "db" time (GMT), but a user entering the field is thinking in terms of _their_ time zone...  We have functions that convert from "db" time to "user" time, but I couldn't find one to convert the other way around, so needed to add one.  

Also changed the comparison from > to >= as would be expected on the "date =" search; otherwise we may miss records that have had times truncated.  

Also got rid of the inaccurate prompt.

### DIFF
--- a/Sources/ManageMembers.php
+++ b/Sources/ManageMembers.php
@@ -340,7 +340,7 @@ function ViewMemberlist()
 					}
 					elseif ($search_params['types'][$param_name] == '-')
 					{
-						$query_parts[] = $param_info['db_fields'][0] . ' < ' . un_forum_time(true, $search_params[$param_name]) + 86400;
+						$query_parts[] = $param_info['db_fields'][0] . ' < ' . (un_forum_time(true, $search_params[$param_name]) + 86400);
 					}
 					else
 						$query_parts[] = $param_info['db_fields'][0] . ' ' . $range_trans[$search_params['types'][$param_name]] . ' ' . un_forum_time(true, $search_params[$param_name]);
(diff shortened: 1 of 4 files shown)


sbulen:
Testing revealed an issue - user offset is always for today.  It may not be the right offset for a historical date, due to dst.

Heading out of town for a few days.  Will address upon my return.

sbulen:
Issue mentioned above has been addressed.

sbulen:
Speaking of pulling on sweater threads...   

DST wasn't handled properly nearly anywhere, mainly because the user offset logic was duplicated (warts & all) in timeformat().

sbulen:
I sprouted a conflict that needed to be dealt with...

sbulen:
Conflict resolved.

MissAllSunday:
@sbulen Can you resolve the conflicts to merge this in, thanks.

sbulen:
Done.

PR #5647, opened 2019-05-03
[WIP] Unsubscribe mail

albertlast:
Userstory:
Every mail from smf to an user should got a unsubscribe url included,
on the first time an unsubscribe token will be generated.

possible url schema: https://www.foobar.com/index.php?action=unsubscribe;token=123456
Maybe we could add an user id.

This url leads to an place where can unsubscribe from all (no partial unsubscribe is planed in this way),
the law say that he needs to be able to do so without login (reason why we need the token).

Since the login based only on this token,
we want to treat this user most likly as a guest.

This code is in a very early stage,
reason why i already keep up with this to see if someone else is further and
otherwise to get some advice for some issues:

Known issue:
The code is placed atm in profile env,
this env based highly on the need to know the user,
which can't be delievered by the way how it works.
To make a hole in thie area to make it possible doesn't look right to me,
so the question where to place the code?

To unsubscribe query is very straight forward,
it search for the user where the 2 bit is set and removed this,
so when a user ask for a alert + mail he fallback to alert only and
when he only asked for mail he goes to nothing.

@Sesquipedalian


Sesquipedalian:
Thanks for this, @albertlast, but there's no need to work on it any further. I've already written fully working code for this in the 2.0.x branch that I will port over to 2.1 along with the rest of the GDPR-related changes.

albertlast:
Sounds not good to throw all in one pr, 
Also you didn't mention how the system behave so many question already on this small thing..

PR #5648, opened 2019-05-04
New hook - integrate_mod_richedit_buttons

dragomano:


Themes/default/GenericControls.template.php
```diff
@@ -100,23 +100,47 @@ function template_control_richedit_buttons($editor_id)
...
-		<input type="hidden" id="id_draft" name="id_draft" value="', empty($context['id_draft']) ? 0 : $context['id_draft'], '">';
-
-	if ($context['show_spellchecking'])
-		echo '
-		<input type="button" value="', $txt['spell_check'], '" tabindex="', --$tempTab, '" onclick="oEditorHandle_', $editor_id, '.spellCheckStart();" class="button">';
-
-	if ($editor_context['preview_type'])
-		echo '
-		<input type="submit" name="preview" value="', isset($editor_context['labels']['preview_button']) ? $editor_context['labels']['preview_button'] : $txt['preview'], '" tabindex="', --$tempTab, '" onclick="', $editor_context['preview_type'] == 2 ? '' : 'return submitThisOnce(this);', '" accesskey="p" class="button">';
+	$richedit_buttons = array(
```
MissAllSunday: I like the idea of grouping the form elements into an array, however, I'm not really sure about placing a hook on a template file.

What I would like to see here is to create a new $context array with the same code you used in $richedit_buttons, put that context var in a source var and then call a new hook for it or just place it on a source file that already has a hook.

Then on the template you will just print that $context var just as you already do after your hook call.

Maybe even using $settings instead of $context to allow both mod and theme authors to define their own buttons for the editor.

MissAllSunday:
Couldn't find any issues with the jquery update but if they are, might as well merge this sooner so they can be spotted.

PR #5649, opened 2019-05-04
Fixes #5540

### LINKED ISSUE #5540: alerts errors
#### Description

sometimes error appears when all alerts are deleted

IMG-20190326-WA0000

Type of error: Undefined
8: Undefined offset: 191 /index.php?action=profile;area=alerts_popup;u=15946
/home/....../public_html/Sources/Profile-View.php (Line 392) 

### Steps to reproduce
1. 
2. 

- Version/Git revision:
- Database Type:
- Database Version:
- PHP Version: 5.6.32 

SychO9:
Fixes #5540 
Though I can't reproduce the dumped data, so this only fixes the error

### DIFF
--- a/Sources/Errors.php
+++ b/Sources/Errors.php
@@ -46,10 +46,7 @@ function log_error($error_message, $error_type = 'general', $file = null, $line 
 
 	// are we in a loop?
 	if ($error_call > 2)
-	{
-		var_dump($backtrace);
 		die('Error loop.');
-	}
 
 	// Check if error logging is actually on.
 	if (empty($modSettings['enableErrorLogging']))
--- a/Sources/Errors.php
+++ b/Sources/Errors.php
@@ -46,7 +46,10 @@ function log_error($error_message, $error_type = 'general', $file = null, $line 
 
 	// are we in a loop?
 	if ($error_call > 2)
+	{
+		var_dump($backtrace);
 		die('Error loop.');
+	}
 
 	// Check if error logging is actually on.
 	if (empty($modSettings['enableErrorLogging']))
--- a/Sources/Profile-View.php
+++ b/Sources/Profile-View.php
@@ -293,9 +293,11 @@ function fetch_alerts($memID, $all = false, $counter = 0, $pagination = array(),
 	{
 		if ($alert['content_type'] == 'msg')
 			$possible_msgs[] = $alert['content_id'];
-		elseif (isset($alert['extra']['topic']))
+
+		if (isset($alert['extra']['topic']))
 			$possible_topics[] = $alert['extra']['topic'];
-		elseif (isset($alert['extra']['board']))
+
+		if (isset($alert['extra']['board']))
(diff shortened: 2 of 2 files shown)


Sources/Profile-View.php
```diff
@@ -393,7 +393,7 @@ function fetch_alerts($memID, $all = false, $counter = 0, $pagination = array(),
 			{
 				$alerts[$id_alert]['extra']['msg_msg'] = $txt['topic_na'];
 				$alerts[$id_alert]['extra']['topic_msg'] = $topic_msgs[$alert['extra']['topic']][$alert['show_links'] ? 'link' : 'text'];
-				$alerts[$id_alert]['extra']['board_msg'] = isset($alert['extra']['board']) ? $board_msgs[$alert['extra']['board']][$alert['show_links'] ? 'link' : 'text'] : $txt['board_na'];
+				$alerts[$id_alert]['extra']['board_msg'] = isset($alert['extra']['board']) && isset($board_msgs[$alert['extra']['board']]) ? $board_msgs[$alert['extra']['board']][$alert['show_links'] ? 'link' : 'text'] : $txt['board_na'];
```
MissAllSunday: This will cause undefined errors if for some reason $board_msgs[$alert['extra']['board']] ends up being an empty array, it will pass the isset check but will show an undefined error because it will not have any key on it.

Ternaries are useful but it can quickly become a mess if more checks are added, therefor it might be better to declare a var and perform all the needed checks on it, once we've done that just pass the var to $alerts[$id_alert]['extra']['board_msg']

live627:
rte error dump is from https://github.com/SimpleMachines/SMF2.1/blob/2eafeb3b490bc3bc134ac1d7538b27d943655ff3/Sources/Errors.php#L50

and nobody wants to remove it. Pisses me off!

Sources/Errors.php
```diff
@@ -46,10 +46,7 @@ function log_error($error_message, $error_type = 'general', $file = null, $line
 
 	// are we in a loop?
 	if ($error_call > 2)
-	{
-		var_dump($backtrace);
```
albertlast: Without replacement(replace it with nothing is no replacement),
you can't remove this line of code.

albertlast (changes requested):
Make it better or keep it be.

Sources/Errors.php
SychO9: I thought this whole block was for debugging, if not I'll just revert the commit and you can deal with it then.
```
	// are we in a loop?
	if ($error_call > 2)
	{
		var_dump($backtrace);
		die('Error loop.');
	}
```

albertlast: Like i mention many many many times,
it's last >hope< when smf is in an error loop,
to get to an end AND to get the least amount of information.
In this case we can't call any smf function or do any db call or do any filesystem call.

albertlast (approved):

PR #5654, opened 2019-05-05
Fixes a Access-Control-Allow-Origin error with sub domains

jdarwood007:
This is a minor issue, but while testing SMF via SSI.php with $ssi_themes across subdomains (i.e. www.site.tld/forums is SMF and support.site.tld/ is the current page), the browser will send a error

> Access to XMLHttpRequest at 'https://www.site.tld/forums/cron.php?ts=1557074880' from origin 'https://support.site.tld' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.

To work around this, we are sending a header now in Cron.php to allow any.  This is allowed as long as cron.php does not specify credentials: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin

More information here: https://www.moesif.com/blog/technical/cors/Authoritative-Guide-to-CORS-Cross-Origin-Resource-Sharing-for-REST-APIs/
Which suggests that even *.site.tld is not allowed.  Meaning there is not easy way for us to fix this unless we check the referral and ensure it is in our domain.  Not something that can easily be done in cron.php without adding lots of code and assumptions due to lack of entire SMF loading.

### DIFF
--- a/cron.php
+++ b/cron.php
@@ -70,6 +70,10 @@ if (file_exists($cachedir . '/cron.lock'))
 // Before we go any further, if this is not a CLI request, we need to do some checking.
 if (!FROM_CLI)
 {
+	// When using sub-domains with SSI and ssi_themes set, browsers will receive a "Access-Control-Allow-Origin" error.
+	// * is not ideal but the best method to preventing this from occurring.
+	header('Access-Control-Allow-Origin: *');
+
 	// We will clean up $_GET shortly. But we want to this ASAP.
 	$ts = isset($_GET['ts']) ? (int) $_GET['ts'] : 0;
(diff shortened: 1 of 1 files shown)


albertlast:
in your case i would start the con by the crontab of the os when possible.

jdarwood007:
In most preferred cases yes it should be.  I don't see any security violations by allowing this since no data is returned.  This just allows firing off of processing for cleanups and other tasks.  Although not to be confused with scheduled tasks since they do essentially the same thing.

Sesquipedalian (approved):

PR #5667, opened 2019-05-11
Prune PM issues

### LINKED ISSUE #5640: Session verification fails when deleting all pms
Pms -> Action -> Prune Messages -> Delete all

the session verification always fails.

sbulen:
Fixes #5640 

Once past the session error, there was some logic there that caused an undefined index issue on 'f'.  Looks like this used to delete PMs for specific folders.  But the user prompts now say it's gonna delete everything, so... it's gonna delete everything...

### DIFF
--- a/Sources/PersonalMessage.php
+++ b/Sources/PersonalMessage.php
@@ -2825,7 +2825,7 @@ function MessageKillAll()
 {
 	global $context;
 
-	checkSession('get');
+	checkSession();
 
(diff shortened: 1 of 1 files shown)


Sesquipedalian (approved):

Sesquipedalian:
Interestingly enough, it looks like this logic has been around since early in the pre-release versions of 2.0, yet was never put into use.

PR #5680, opened 2019-05-25
Add support for multidimensional arrays

### LINKED ISSUE #5653: strlen() expects parameter 1 to be string, array given
Logging::truncateArray()  should be able to handle multidimensional arrays or skip them gracefully

`Type of error: General
Error message: 2: strlen() expects parameter 1 to be string, array given
/var/www/html/Sources/Logging.php
Line: 33`

MissAllSunday:
and move truncate_array to Sources/Subs.php

fixes #5653

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -7052,8 +7052,7 @@ function truncate_array($array, $max_length = 1900, $deep = 3)
 {
     $array = (array) $array;
 
-    $curr_length = 0;
-    $curr_length = array_length($array);
+    $curr_length = array_length($array, $deep);
 
     if ($curr_length <= $max_length)
         return $array;
@@ -7063,15 +7062,13 @@ function truncate_array($array, $max_length = 1900, $deep = 3)
         // Truncate each element's value to a reasonable length
         $param_max = floor($max_length / count($array));
 
-        $currentDeep = $deep - 1;
+        $current_deep = $deep - 1;
 
         foreach ($array as $key => &$value)
         {
             if (is_array($value))
-            {
-                if ($currentDeep > 0)
-                    $value = truncate_array($value, $currentDeep);
-            }
+                if ($current_deep > 0)
+                    $value = truncate_array($value, $current_deep);
 
             else
                 $value = substr($value, 0, $param_max - strlen($key) - 5);
@@ -7093,7 +7090,7 @@ function array_length($array, $deep = 3)
     $array = (array) $array;
     $length = 0;
 
-    $deepCount = $deep - 1;
+    $deep_count = $deep - 1;
 
     foreach ($array as $value)
     {
@@ -7101,10 +7098,10 @@ function array_length($array, $deep = 3)
         if (is_array($value))
         {
             // No can't do
-            if ($deepCount <= 0)
+            if ($deep_count <= 0)
                 continue;
 
-            $length += array_length($value, $deepCount);
+            $length += array_length($value, $deep_count);
         }
 
         else
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -7040,4 +7040,78 @@ function sentence_list($list)
 	return strtr($format, $replacements);
 }
 
+/**
+ * Truncate an array to a specified length
+ *
+ * @param array $array The array to truncate
+ * @param int $max_length The upperbound on the length
+ * @param int $deep How levels in an multidimensional array should the function take into account.
+ * @return array The truncated array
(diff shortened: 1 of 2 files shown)


Sources/Subs.php
```diff
@@ -7040,4 +7040,78 @@ function sentence_list($list)
...
 
+/**
+ * Truncate an array to a specified length
+ *
+ * @param array $array The array to truncate
+ * @param int $max_length The upperbound on the length
+ * @param int $deep How levels in an multidimensional array should the function take into account.
+ * @return array The truncated array
+ */
+function truncate_array($array, $max_length = 1900, $deep = 3)
```
albertlast, line 7051: why not camel case?

Sources/Subs.php
```diff
@@ -7040,4 +7040,78 @@ function sentence_list($list)
...
+{
+    $array = (array) $array;
+
+    $curr_length = 0;
+    $curr_length = array_length($array);
```
Sesquipedalian: This should probably be `array_length($array, $deep)` in order to ensure that both functions are working to the same depth.

Sources/Subs.php
```diff
@@ -7040,4 +7040,78 @@ function sentence_list($list)
...
+
+    if ($curr_length <= $max_length)
+        return $array;
+
+    else
+    {
+        // Truncate each element's value to a reasonable length
+        $param_max = floor($max_length / count($array));
+
+        $currentDeep = $deep - 1;
```
Sesquipedalian: It would be cleaner and simpler to just use `$deep--` here and replace all following uses of `$currentDeep` with just `$deep`.

Sources/Subs.php
```diff
@@ -7040,4 +7040,78 @@ function sentence_list($list)
...
+ * @param int $deep How many levels should the function
+ * @return int
+ */
+function array_length($array, $deep = 3)
+{
+    // Work with arrays
+    $array = (array) $array;
+    $length = 0;
+
+    $deepCount = $deep - 1;
```
Sesquipedalian: Previous comment about `$deep--` applies in this function, too.

Sources/Subs.php
```diff
@@ -7040,4 +7040,78 @@ function sentence_list($list)
...
+ * @param array $array The array to truncate
+ * @param int $max_length The upperbound on the length
+ * @param int $deep How levels in an multidimensional array should the function take into account.
+ * @return array The truncated array
+ */
+function truncate_array($array, $max_length = 1900, $deep = 3)
+{
+    $array = (array) $array;
+
+    $curr_length = 0;
```
Sesquipedalian: This line no longer seems necessary.

Sesquipedalian (changes requested):

Sources/Subs.php
MissAllSunday: We don't have a proper set of rules/guides for this kind of stuff but we should and for now it will be better to follow what we already have.

Sources/Subs.php
MissAllSunday: agree

Sources/Subs.php
MissAllSunday: Cleaner perhaps but I would like to keep it as two separate values for readability. That and also to prevent unwanted side effects in case there's some weird behaviors on using the same var while in recursive. One should always expect the whole "by-value function cannot modify the value of the variable that was passed to it" thing to hold true but you never know.

Sources/Subs.php
MissAllSunday: yep

Sources/Subs.php
Sesquipedalian: Yeah, officially we accept both camel case and underscore separation, but the unofficial preference has been for underscore separation for a while now. Perhaps we should update the official policy to make that preference clear.

Sources/Subs.php
Sesquipedalian: We can agree to disagree on that, then. 🙂 It doesn't make any substantial difference in this case anyway.

albertlast: $var-- is better to read than $var = $var - 1

MissAllSunday: This isn't a case of $var = $var - 1

Sesquipedalian:
Aside: For some silly reason the folks at Travis CI don't think that their mobile UI should allow me to manually trigger a rebuild from my phone. 🤨

Sesquipedalian (approved):

PR #5681, opened 2019-05-25
Menu Subsections fix

### LINKED ISSUE #5517: Profile Area Track User Subitem
#### Description
getting tons of 
```
Type of error: Undefined 
8: Undefined index: label
 http://localhost/Upgrader/index.php?action=profile;u=1
/Themes/default/GenericMenu.template.php (Line 112)
```

### Steps to reproduce
1. Create a membergroup with permission to **Manage and assign membergroups**
2. Must not have permission to **Moderate forum members**
3. Visit a user's profile
trackuser

Latest Git Revision

SychO9:
Fixes #5517

### DIFF
--- a/Sources/ManageServer.php
+++ b/Sources/ManageServer.php
@@ -112,9 +112,9 @@ function ModifySettings()
 	$settings_backup_fail = !@is_writable($boarddir . '/Settings_bak.php') || !@copy($boarddir . '/Settings.php', $boarddir . '/Settings_bak.php');
 
 	if ($settings_not_writable)
-		$context['settings_message'] = '<div class="centertext"><strong>' . $txt['settings_not_writable'] . '</strong></div><br>';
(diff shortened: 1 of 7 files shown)


Sesquipedalian (commented):
In the code added to Subs-Menu.php, please add braces to wrap each layer of subordinate code (i.e. a `foreach` under an `if` under another `foreach `, etc.).

albertlast:
To add braces here looks wrong to me here.
Since the common coding style in smf is to use braces only in cases where php it's needed and
in cases of one command this is not needed.

This changes here are the last case.

Sesquipedalian:
> To add braces here looks wrong to me here.
> Since the common coding style in smf is to use braces only in cases where php it's needed and
> in cases of one command this is not needed.
> 
> This changes here are the last case.

That is usually only true when there is only one level of subordinate code. When there are multiple levels of subordinate code, we usually use braces. (The existing code isn't completely consistent on this, of course, but it's the most common and preferred way.)

Sesquipedalian (approved):

PR #5687, opened 2019-05-28
More alert fixes

### LINKED ISSUE #5682: Cron error (again)
### Steps to reproduce
1. Latest SMF2.1
2. 

- Version/Git revision:
- Database Type:
- Database Version:
- PHP Version: 7.3

Database error, given array of integer values is empty. (members)
Function: updateMemberData

/Sources/Subs.php (Line 422)

Sesquipedalian:
- Fixes #5682
- Overhauls `alert_count()` & `fetch_alerts()` to be saner & more reliable. Among other things, this overhaul :
    - Fixes an unreported bug where `alert_count()` was not counting certain alerts.
    - Gives `fetch_alerts()` the ability to fetch by ID number. The `$all` parameter has been renamed to `$to_fetch` to reflect this change.
    - Replaces the redundant `$counter` and `$pagination` parameters of `fetch_alerts()` with `$limit` and `$offset`. The main difference is that `$pagination` took an array specifying the offset and limit to use, whereas `$offset` takes a simple integer since `$limit` (formerly known as `$counter`) already specifies the limit.
    - Uses set_avatar_data() to get alert senders' avatars rather than loading all of their $memberContext info just for that one job.
    - Standardizes some haphazard alert content_type & content_action values.
- Adds icons to show the alert type in the Alerts popup and on the Show Alerts profile page. (This required some theme changes.)
- On the Show Alerts profile page, redirects to the correct page after taking actions (mark read/unread, delete), and correctly paginates  even when given abnormal "start" values.


Sources/Profile-View.php
```diff
@@ -267,70 +271,128 @@ function fetch_alerts($memID, $all = false, $counter = 0, $pagination = array(),
 
 		if (!empty($row['sender_id']))
 			$senders[] = $row['sender_id'];
+
+		if ($row['content_type'] == 'profile')
+		{
+			$members[] = $row['content_id'];
```
SychO9: might want to initialize **$members** as an array to avoid any cases where it wouldn't reach this point

Sesquipedalian: Good idea.

Sources/Profile-View.php
```diff
@@ -217,221 +217,353 @@ function summary($memID)
...
+			elseif (in_array($msg_type, array('msg_msg', 'topic_msg', 'board_msg')))
+				$alert['extra'][$msg_type] = $txt[$msg_type == 'board_msg' ? 'board_na' : 'topic_na'];
+			else
+				$alert['extra'][$msg_type] = '(' . $txt['not_applicable'] . ')';
+		}
 
+		// If we loaded the sender's profile, we may as well use it.
+		$sender_id = !empty($alert['sender_id']) ? $alert['sender_id'] : 0;
+		if (isset($user_profile[$sender_id]))
+			$alert['sender_name'] = $user_profile[$sender_id]['real_name'];
```
SychO9: don't forget we need the sender's avatar though, I think we can get that after loading the member's context

Sesquipedalian: Bleh. You're right. I guess I'll put it back, then.

Sesquipedalian:
Until now we've been displaying avatars next to alerts in the tiny, space constrained popup list, even though this requires a bunch of time-consuming extra queries in a function that is called fairly frequently and which really needs load as quickly as possible. Conversely, we have *not* been loading and displaying avatars on the full sized alerts list page, where neither speed nor space are an issue.

Am I the only one who thinks this is completely backwards?

For my part, I would greatly prefer for the popup to display either (a) just the alert text, or else (b) an icon indicating the alert type rather than the avatar of the alert sender. Not only would such an icon be computationally inexpensive, but it would also be more useful information for the reader than an avatar would be.

Sources/Profile-View.php
Sesquipedalian: I've restored the code to load the `$memberContext` info (including the avatar) for the alert senders in my local repository, but before I commit it I'd appreciate interaction on my question below.

SychO9:
What if we were to not load the member context to get the avatar, and instead directly get the avatar only, using the set_avatar_data() function which would require selecting the avatar + email columns, and the filename from the attachments table, it wouldn't be that expensive then.

I don't hate the icons idea either to be honest, I gave it a try locally and it visually gives more information on alerts
icons

we could even go with both avatars and icons if what I said above is correct and we can get avatars in a very less expensive manner.

Sources/Profile-Modify.php
```diff
@@ -2342,110 +2347,130 @@ function alert_count($memID, $unread = false)
...
 		$request = $smcFunc['db_query']('', '
-			SELECT t.id_topic, b.id_board
+			SELECT t.id_topic
 			FROM {db_prefix}topics AS t
-				INNER JOIN {db_prefix}boards AS b ON (t.id_board = b.id_board)
-			WHERE ' . $query_see_board . '
+			WHERE ' . $qb['query_see_topic_board'] . '
 				AND t.id_topic IN ({array_int:topics})',
 			array(
 				'topics' => $possible_topics,
```
SychO9, line 2426: I am getting `Invalid value inserted or no type specified.` on this line when locking/stickying a topic

Sesquipedalian: Can you give me the backtrace info for that error and a bit more info on how to reproduce it? Was that using quick moderation or the buttons in the topic display? What were the permissions of the user that triggered the error?

SychO9:
```
#0: smf_db_error_backtrace()
Called from ...\Sources\Subs-Db-mysql.php on line 190
#2: preg_replace_callback()
#3: smf_db_query()
Called from ...\Sources\Profile-Modify.php on line 2431
#4: alert_count()
Called from ...\Sources\Subs.php on line 398
#5: updateMemberData()
Called from ...\Sources\tasks\CreatePost-Notify.php on line 358
#6: execute()
Called from ...\cron.php on line 216
#7: perform_task()
Called from ...\cron.php on line 110
```

user is an admin, the action was taken from the topic page

Sesquipedalian: Derp. I was using `{query_see_message_board` and` {query_see_topic_board}`, but those were added in a different branch. 82d8de5 should fix it.

SychO9:
well for some reason the last commit doesn't fix the error

Sesquipedalian:
> well for some reason the last commit doesn't fix the error

Double derp. 07f3a8e should *actually* fix it.

Sesquipedalian:
The alert icons are centred if there is no avatar to show, and appear in the upper right corner if there is an avatar.

Screen Shot 2019-06-02 at 6 10 48 PM

Sources/Profile-View.php
```diff
@@ -451,24 +592,12 @@ function showAlerts($memID)
...
-			)
-		);
-		$alert = $smcFunc['db_fetch_assoc']($request);
-		$smcFunc['db_free_result']($request);
+		$alert_id = (int) $_REQUEST['alert'];
 
-		// Can the user see this alert ?
-		if($memID != $alert['id_member'])
-			redirectexit();
+		$alert = array_pop(fetch_alerts($memID, $alert_id));
```
SychO9: You should assign the result to a variable first and then use it in array_pop, as I am getting 
"**Only variables should be passed by reference**" in the errolog

Sources/Profile-View.php
```diff
@@ -217,221 +217,362 @@ function summary($memID)
...
-			$alerts[$id_alert]['text'] = str_replace($search, $repl, $txt[$string]);
+
+			$alert['text'] = strtr($txt[$string], $substitutions);
 		}
+
+		// Unset the reference variable to avoid any surprises in subsequent loops.
+		unset($alert);
 	}
 
 	return $alerts;
```
SychO9, line 585: I'm thinking it would be nice to either move the hook before the return or add a new one that would maybe be called **integrate_fetch_alerts_after** there are some data that cannot be edited from the current hook like the icon

Sources/Profile-Modify.php
```diff
@@ -2342,110 +2347,97 @@ function alert_count($memID, $unread = false)
...
+			$alerts[$row['id_alert']]['visible'] = true;
 	}
 	$smcFunc['db_free_result']($request);
 
-	$possible_boards = array_unique($possible_boards);
-
-	// If this isn't the current user, get their boards.
-	if (!isset($user_info['id']) || $user_info['id'] != $memID)
+	// If we need to check board access, use the correct board access filter for the member in question.
+	if ((!isset($user_info['id']) || $user_info['id'] != $memID) && (!empty($possible_msgs) || !empty($possible_topics)))
```
SychO9, line 2373: Seems like this still needs more tweaking as I am still getting the errors
"**Invalid value inserted or no type specified.(Line 2426)**"

SychO9 (commented):
That is a lot of work you've put into it, the icons look really nice too

Sources/Profile-Modify.php
SychO9: hmm, I noticed this also happens in the main branch, so your changes did not introduce it

Sesquipedalian: That's good to know. I'll stop banging my head against this wall and look elsewhere for the cause. Can I get you to file a new bug report about it?

SychO9:
One more thing: it seems like there are a few cases where an alert would show (private topic) when it shouldn't, like the **like** alert: "**user liked your post (private topic)**"

Sources/Profile-View.php
Sesquipedalian: I just moved the call to set_alert_icon() to a point before the call to the hook. I don't see anything else that a hook would need access to.

SychO9: yes I think that is enough indeed

Sesquipedalian:
Er, I don't believe so. Alerts about likes on posts the user can't see will be filtered out and not shown at all. Alerts about likes on posts the user can see should have all the info they need to build real messages. The fallback to use`$txt['board_na']` and `$txt['topic_na']` is mostly just there in case a mod screws up or something is wrong in the database.

SychO9:
except it's showing (private topic) when I am able to see the topic :/, try to produce a like alert, and you'll see

Sesquipedalian:
> except it's showing (private topic) when I am able to see the topic :/, try to produce a like alert, and you'll see

Found it: 4918925. 🙂

PR #5688, opened 2019-05-28
give php-cs-fixer a try

MissAllSunday:
Ran php-cs-fixer for some of the most common things, heres the ruleset I used:

https://gist.github.com/MissAllSunday/b171ca6af5f2c3e1a7488318585aec15

Fixes some inconsistencies such as:

function () 
using short array syntax
spaces on blank lines
braces on new class declarations 
proper space after =

Might be a good idea to add php-cs-fixer to our travis stuff so every commit gets scanned against the same ruleset


jdarwood007:
The usage of list was consistent though it seems in our code base, even if php-cs-fixer wants to make it different.  Not sure if we should fix it or leave it.

It also seems to want to change list(,, $var) to list (, , $var), which is not what we want either.

Some of the areas had lots of tabs after => to indent these to be consistent and easier to read.

MissAllSunday:
The list () to list() is just for consistency with other functions and/or constructors.

I reverted back the list(,, $var) although that rule was set for function arguments, I don't know if theres a way to add list() to some kind of whitelist.

live627:
I always looked at `list ()` as a language construct, so the space could easily differentiate that from normal functions.

PR #5698, opened 2019-06-01
Supply missing title

### LINKED ISSUE #5277: Odd empty section header
#### Description
Visiting Admin > Maintenance > Logs > Log Settings produces three sections of settings, the middle one of which has a catbar but no text in it.

image

sbulen:
Fixes #5277


SychO9:
oh I fixed that [here](https://github.com/SimpleMachines/SMF2.1/pull/5681) and didn't notice there was an issue about it

sbulen:
It's still an issue as of today's GH.   I don't think there is a value for 'errorlog', which is why I created 'errorlog_title'.

SychO9:
no there is a value for errorlog, $txt['errorlog'] = 'Error Log';
your installation must not be updated is all

sbulen:
You are correct, sir!   Thanks - not sure how I missed that.

PR #5700, opened 2019-06-03
Fixes bugs detecting language names

### LINKED ISSUE #5699: Language dropdown should never says "English" for non-English languages
If a language pack does not yet have a proper translation in the `$txt['native_name']` variable, the language selection menu will say "English" for that language instead of whatever it should say.

The long term fix is, of course, for all the language packs to be updated properly. But in the real world, we need to add a hack to fall back to the language file name if the value of `$txt['native_name']` is `English` when the language is not actually English.

### LINKED ISSUE #5705: smf langlist setting not cleared when removing language files
#### Description
On an upgraded forum version from 2.015 to 2.1 RC2 uninstalling language did not clear smf_settings variable "langList"
Result is that the language selector is showing the value "Dutch" instead of the textstring $txt['native_name'] 

### Steps to reproduce
1. 
2. 

- Version/Git revision: 2.1 RC2, last_commit_3428c48
- Database Type:
- Database Version:
- PHP Version: 7.1

See https://www.simplemachines.org/community/index.php?topic=567178.msg4021137#msg4021137

Sesquipedalian:
- Fixes #5699 
- Fixes an unreported bug filtering out `-utf8` files.
- Fixes #5705

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -3010,7 +3010,7 @@ function getLanguages($use_cache = true)
 			while ($entry = $dir->read())
 			{
 				// Look for the index language file... For good measure skip any "index.language-utf8.php" files
-				if (!preg_match('~^index\.(.+[^-utf8])\.php$~', $entry, $matches))
+				if (!preg_match('~^index\.((?:.(?!-utf8))+)\.php$~', $entry, $matches))
 					continue;
 
 				if (!empty($langList) && !empty($langList[$matches[1]]))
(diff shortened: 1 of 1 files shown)


albertlast:
How the system behave with this change,
When English is added?
Would it be not smart to check if a language is listed twice?

sbulen:
@Sesquipedalian - Note that #5705 has an RC3 tag, so this should probably be considered an RC3 PR.

PR #5704, opened 2019-06-04
Board description

### LINKED ISSUE #5372: When editing a board, all html tags are stripped
Unless this is not what I think it is, it is not being respected, similar to the float issue:

https://github.com/SimpleMachines/SMF2.1/blob/4fa7e97e66008a73a81dff5b5c238510f9b241b6/Sources/Load.php#L353-L357

### LINKED ISSUE #5404: Float bbc gets removed from board description
Since #5271 ended being two issues, I'm opening a new one.

> 3 - Float bbc gets edited out of the description

That is still unfixed. Adding it to the allowed list will let it be parsed once you save the description, but once you go to edit the board again, it will be gone.

MissAllSunday:
fixes #5372
fixes #5404

Parsing is now done on display. On saving and on upgrading HTML gets parsed as BBC first then the ones we couldn't convert are stripped by strip_tags(), lastly, any remains gets  $smcFunc['htmlspecialchars']()


Sources/ManageBoards.php
```diff
@@ -394,13 +395,14 @@ function EditBoard()
...
 	$_REQUEST['boardid'] = isset($_REQUEST['boardid']) ? (int) $_REQUEST['boardid'] : 0;
+
 	if (!isset($boards[$_REQUEST['boardid']]))
 	{
 		$_REQUEST['boardid'] = 0;
 		$_REQUEST['sa'] = 'newboard';
 	}
 
-	if ($_REQUEST['sa'] == 'newboard')
+	if ('newboard' === $_REQUEST['sa'])
```
albertlast, line 405: why you add this "new" style every where?

other/upgrade_2-1_mysql.sql
```diff
@@ -115,6 +115,43 @@ if (@$modSettings['smfVersion'] < '2.1')
 ---}
 ---#
 
+---# Parsing board descriptions and names
+---{
+if (@$modSettings['smfVersion'] < '2.1')
```
albertlast: Please had in mind that we got rc1 to rcX support, this would break here

other/upgrade_2-1_postgresql.sql
```diff
@@ -306,6 +306,43 @@ if (@$modSettings['smfVersion'] < '2.1')
 ---}
 ---#
 
+---# Parsing board descriptions and names
+---{
+if (@$modSettings['smfVersion'] < '2.1')
```
albertlast: Same like mysql, not safe for 2.1 to 2.1 upgrade

other/upgrade_2-1_mysql.sql
```diff
@@ -115,6 +115,43 @@ if (@$modSettings['smfVersion'] < '2.1')
...
+    while ($row = $smcFunc['db_fetch_assoc']($request))
+    {
+        $inserts[] = array(
+            'name' => $smcFunc['htmlspecialchars'](strip_tags(html_to_bbc($row['name']))),
+            'description' => $smcFunc['htmlspecialchars'](strip_tags(html_to_bbc($row['description']))),
+            'id' => $row['id'],
+        );
+    }
+
+    if (!empty($inserts))
```
albertlast, line 139: The check should be unneed because the array when empty should run zero times in the foreach?

Sources/ManageBoards.php
MissAllSunday: They are Yoda conditions: https://en.wikipedia.org/wiki/Yoda_conditions

They help to prevent accidental variable assigning. We should also start to use real comparisons (value and type) whenever possible

other/upgrade_2-1_mysql.sql
MissAllSunday: The parsing should be done just once. On that same file the same version compare [was used](https://github.com/SimpleMachines/SMF2.1/blob/2a6d524410429ae4cf11fb8f1bf7b65d67f100e2/other/upgrade_2-1_mysql.sql#L96) and merged into release-2.1 branch, hence why I use it.

Sources/ManageBoards.php
albertlast: Yeah i mean the yoda conditions,
from my pov this base issue of "=" didn't exists in php and
the worst reading is a real thing.

other/upgrade_2-1_mysql.sql
albertlast: Yes,
this exists before of realsing rc1,
you change came after this thow it's to late to use this kind of checks,
because already working 2.1 setup will not run this in upgrade path which they need.

other/upgrade_2-1_mysql.sql
albertlast: Wrong guess,
you could use ``forearch((array)  $inserts as $insert)``

Sources/ManageBoards.php
MissAllSunday: PHP has always been a loosely typed language, don't really get what you mean by "didn't exists in php". If you mean SMF, yes, we have always used "loose" instead of "strict", that doesn't mean we should continue to use it on things we should be using strict compares.

As for yoda conditions, you are just changing the order, putting the constant first makes you realize it is comparing and not assigning. Helps you from stop taking things for granted too.

albertlast: I don't understand why you mix upo different thing,
the type checks is fine, nothing wrong here.

Only the yoda style is unneed because singel = had only the task to assign value not to check them.

MissAllSunday: Putting the constant first helps you from accidentally assigning a variable, take this example:

$var == 'constant';

Imaging you accidentally remove a = 

$var = 'constant';

Now $var means 'constant'; and it will always return true.  This happens way more frequent that you think, specially on large codebases touched by several people.

albertlast: but we talk about = vs === and here remove accidentally a = change not the equality check.

MissAllSunday: Yes.... thats exactly what this change does, besides doing the yoda conditions, using strict comparisons helps out too.

In resume, yoda comparisons prevents accidental assigning and helps out figure it out we are doing a compare and not an assign.

We should use strict comparisons whenever they are needed. Such as this case.

Those two things (yoda and strict) are separated entities and will not always be used together like on this example. There might be cases where only one will be applied.

And I believe I already spent a lot more comments than needed explaining this change, if there are further questions these links will help out:

https://en.wikipedia.org/wiki/Yoda_conditions
https://www.php.net/manual/en/types.comparisons.php

other/upgrade_2-1_mysql.sql
MissAllSunday: What do you recommend then?

other/upgrade_2-1_postgresql.sql
MissAllSunday: Keep in mind that we do not want to do this change during a 2.1 to 2.1 upgrade, this change should only be done from 2.0 and only once.

other/upgrade_2-1_mysql.sql
albertlast: What would be happen when you run the code twice?
Which logic get here the problem?

MissAllSunday: It should run only once, when a user goes from 2.0 to 2.1, if the same user for whatever reason does another upgrade from 2.1 to 2.1 this particular step should not be executed.

albertlast: I ask you why?
Which logic get broken here?

MissAllSunday: I don't really understand what you mean, if you ask why this has to be done once, here's why:

2.0 supports adding raw HTML to board descriptions and I think categories too, can't remember.

2.1 doesn't allow this anymore, in fact, it doesn't allow to add BBC either expect by the ones specified in $context['description_allowed_tags'].

In order to provide a better upgrade experience for those users who had HTML on their description boards, an attempt to parse their HTML into BBC is done. Any description is now saved in the DB as BBC and that BBC gets parsed on display only, this is why this step has to be done just once.

albertlast: So it's create no issue when it run twice?

MissAllSunday: It shouldn't but in practice things can and surely will get broken since already htmlspecialchars HTML will get another round of htmlspecialchars, making it look quite broken for the user.

But more importantly, why run things twice where there isn't any need for? the upgrade isn't service friendly so if we can avoid doing unneeded steps then lets do it.

albertlast: Don't let's do thing just we can,
you need to concider that you had to choose between run once and never or once and multiple time.

So instead of hurting yourself,
you could consider if it possible to write the code in a way how is able to run multiple times without hurting.

When it's very important to you to run this realy once,
than write in the setting or theme table a row that this task is done and
check on a rerun if the task is marked as done.

MissAllSunday: ???

I did consider the options and the best one is to do this step only when they upgrade from 2.0 to 2.1.

The same DB will only do this upgrade once (2.0 to 2.1), they can of course do another upgrade, in that case it will be a 2.1 to 21. upgrade on which case this step is useless and we shouldn't waste CPU cicles on it.

I really don't think a setting or a theme setting is required here, $modSettings['smfVersion'] < '2.1' fulfills our needs quite well, we only need to run this when a user goes from 2.0 to 2.1, that's it.

From what I can see, the $modSettings['smfVersion'] < '2.1' check won't pass for any RC or beta version and that's OK, in fact it is exactly what we want, we only want this piece of code to be executed when there's a 2.0 to 2.1 upgrade, all other scenarios don't matter to us.

albertlast: I find this interesting that more important for you,
that the upgrade process runs fast,
instead of reliable.

solution already exists how could fix it: https://www.simplemachines.org/community/index.php?topic=548174.0
but was decided by persons not to take this way.

To get to the point,
it can happen in the upgrade process that error apears which can let you upgrade process skip somewhere in the middle but the mod version get raised up,
when the admin start again the upgrade you your partial or never hittet upgrade skript part will be never vistid again --> weak reliability

MissAllSunday: You do realize that this issue has nothing to do with partial upgrades?

If you want to yet again give it a shot and discuss partial upgrades then open a separate issue or bump that SMF topic.

Partial upgrades from beta or RC for this particular case? not important at all, we only care about 2.0 to 2.1.

If the upgrade process fails then all the user has to do is go to admin/boards/edit board and save the page, that's it. Saving the edits will trigger the exact same piece of code the upgrade failed to do.

This isn't a "must have" or "must execute" thing, its merely a way to offer a better upgrade process, it isn't a required step and shouldn't be any issues if the upgrade fails to execute this step as its quite simple to manually make the change if needed.

Sesquipedalian: Doing a simple version check is fine in this case. However, it would be more reliable to use 
```
if (version_compare(trim(strtolower(@$modSettings['smfVersion'])), '2.1.foo', '<'))
```
rather than 
```
if (@$modSettings['smfVersion'] < '2.1')
```

Also, @MissAllSunday, I noticed that there are two other spots in the upgrade_2-1_*.sql files that use 
```
if (@$modSettings['smfVersion'] < '2.1')
```
and one that uses just
```
if (version_compare(@$modSettings['smfVersion'], '2.1', '<'))
``` 
It'd be lovely if you could also change those to use 
```
if (version_compare(trim(strtolower(@$modSettings['smfVersion'])), '2.1.foo', '<'))
```

Sesquipedalian (changes requested):
Please change all version checks in the upgrade_2-1_*.sql files to
```
if (version_compare(trim(strtolower(@$modSettings['smfVersion'])), '2.1.foo', '<'))
```

Sesquipedalian (approved):

Issue #5670, opened 2019-05-14
Stats should use https


jdarwood007:
#### Description
Stats code uses port 80:
```
		$fp = @fsockopen('www.simplemachines.org', 80, $errno, $errstr);
```

We should change this to support both port 80 and 443, defaulting to 443.  We should also have this start sending a user agent (PHP/SMF)

### Additional information/references
Mostly adding this as a note to myself to look into fixing this.  This should be fixed before final but is not any RC deal breaker.

frandominguezl:
@MissAllSunday If #5706 fixes this please add a "Has Fix" label to this issue.

MissAllSunday:
No it doesn't

PR #5706, opened 2019-06-05
Create user-agent constant SMF_USER_AGENT

MissAllSunday:
This creates a new constant to define our own user agent

#5670

### DIFF
--- a/Sources/Class-CurlFetchWeb.php
+++ b/Sources/Class-CurlFetchWeb.php
@@ -48,7 +48,7 @@ class curl_fetch_web_data
 		CURLOPT_RETURNTRANSFER	=> 1, // Get returned value as a string (don't output it)
 		CURLOPT_HEADER			=> 1, // We need the headers to do our own redirect
 		CURLOPT_FOLLOWLOCATION	=> 0, // Don't follow, we will do it ourselves so safe mode and open_basedir will dig it
-		CURLOPT_USERAGENT		=> 'Mozilla/5.0 (Windows NT 6.1; WOW64; rv:11.0) Gecko Firefox/11.0', // set a normal looking useragent
+		CURLOPT_USERAGENT		=> SMF_USER_AGENT, // set a normal looking useragent
 		CURLOPT_CONNECTTIMEOUT	=> 15, // Don't wait forever on a connection
 		CURLOPT_TIMEOUT			=> 90, // A page should load in this amount of time
 		CURLOPT_MAXREDIRS		=> 5, // stop after this many redirects
--- a/index.php
+++ b/index.php
@@ -26,8 +26,7 @@ define('SMF_VERSION', '2.1 RC2');
 define('SMF_FULL_VERSION', 'SMF ' . SMF_VERSION);
 define('SMF_SOFTWARE_YEAR', '2019');
 define('JQUERY_VERSION', '3.4.1');
-define('SMF_USER_AGENT', 'user-agent: PHP/SMF');
-
+define('SMF_USER_AGENT', 'Mozilla/5.0 (' . php_uname('s') . ' ' . php_uname('m') . ') AppleWebKit/605.1.15 (KHTML, like Gecko)  SMF/' . strtr(SMF_VERSION, ' ', '.'));
 error_reporting(E_ALL);
 $time_start = microtime(true);
 
--- a/index.php
+++ b/index.php
@@ -26,6 +26,7 @@ define('SMF_VERSION', '2.1 RC2');
 define('SMF_FULL_VERSION', 'SMF ' . SMF_VERSION);
 define('SMF_SOFTWARE_YEAR', '2019');
 define('JQUERY_VERSION', '3.4.1');
+define('SMF_USER_AGENT', 'user-agent: PHP/SMF');
 
 error_reporting(E_ALL);
 $time_start = microtime(true);
@@ -57,7 +58,7 @@ require_once($sourcedir . '/Errors.php');
 require_once($sourcedir . '/Load.php');
 
 // If $maintenance is set specifically to 2, then we're upgrading or something.
-if (!empty($maintenance) && $maintenance == 2)
+if (!empty($maintenance) &&  2 === $maintenance)
 	display_maintenance_message();
 
 // Create a variable to store some SMF specific functions in.
--- a/Sources/Stats.php
+++ b/Sources/Stats.php
@@ -816,7 +816,7 @@ function SMStats()
 
 			$out = 'POST /smf/stats/collect_stats.php HTTP/1.1' . "\r\n";
 			$out .= 'Host: www.simplemachines.org' . "\r\n";
-			$out .= SMF_USER_AGENT . "\r\n";
+			$out .= 'user-agent: '. SMF_USER_AGENT . "\r\n";
 			$out .= 'content-type: application/x-www-form-urlencoded' . "\r\n";
 			$out .= 'connection: Close' . "\r\n";
 			$out .= 'content-length: ' . $length . "\r\n\r\n";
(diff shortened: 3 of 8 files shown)


Sources/Class-CurlFetchWeb.php
```diff
@@ -48,7 +48,7 @@ class curl_fetch_web_data
 		CURLOPT_RETURNTRANSFER	=> 1, // Get returned value as a string (don't output it)
 		CURLOPT_HEADER			=> 1, // We need the headers to do our own redirect
 		CURLOPT_FOLLOWLOCATION	=> 0, // Don't follow, we will do it ourselves so safe mode and open_basedir will dig it
-		CURLOPT_USERAGENT		=> 'Mozilla/5.0 (Windows NT 6.1; WOW64; rv:11.0) Gecko Firefox/11.0', // set a normal looking useragent
+		CURLOPT_USERAGENT		=> SMF_USER_AGENT, // set a normal looking useragent
```
Sesquipedalian, line 51: At first glance, I'm uneasy about changing this line. I suspect that there may have been a good reason to set a user agent that looks like a normal browser in this case.

In theory, the user agent shouldn't cause any difference in how the remote server responds, but in practice it is entirely possible that a badly written UA filter on the far end would cause the server to respond differently to a UA that didn't look like a normal browser.

Further investigation or discussion could convince me to lay aside my concerns, of course. 🙂

MissAllSunday: These days the UA isn't really used to filter out bad guys anymore, the most common thing for filtering out bad bots and stuff like that is by specifically checking for data centers like aws or azure, thats where the bad boys are coming from. I know people who ban entire IP ranges of Amazon Web Services

albertlast:
Why hard code this ua than?
Would it be not better so save this in the settings?

To use ua for the stats collection issue,
i don't like.

Sources/Class-CurlFetchWeb.php
sbulen: I would be worried almost the other way around - not filtering requests, but treatment of responses.  This is used in our fetch_web_data routine, which is used in a few places, e.g., the image proxy.  I've seen lots of funky responses - and lack of response - from other websites while I was testing the proxy.   

We may need 'mozilla' and 'gecko', etc., for compatibility.  For interacting with old poorly-configured web servers.  _The same reasons your browser stuffs 'mozilla' and 'gecko' and 'safari' in your browser's user-agent..._  

(Yes, I worry a lot...)

MissAllSunday:
The user agent isn't something we want to change, a setting is mainly used for things that we want to change at some point in time, since we don't want to change the UA and a PHP constant by definition prevents changes, a constant seems a better choice.

Sources/Class-CurlFetchWeb.php
Sesquipedalian: We're thinking similarly, @sbulen. It is silly in theory to have to stuff the UA string with all that cruft, but silly stuff happens all the time on the Internet. 😑  There's a reason why nearly all spiders and proxies out there do it.

That said, I do like what @MissAllSunday is proposing in this PR. Maybe if we just change the value of the constant to something that looks more like a typical UA but still has the unique bits necessary to identify SMF as SMF, that would give us the best balance. We'd want to check with @jdarwood007 to make sure the scripts on simplemachines.org were ready to accept that UA before merging it in, though.

index.php
```diff
@@ -25,6 +25,7 @@
 define('SMF_VERSION', '2.1 RC2');
 define('SMF_FULL_VERSION', 'SMF ' . SMF_VERSION);
 define('SMF_SOFTWARE_YEAR', '2019');
+define('SMF_USER_AGENT', 'user-agent: PHP/SMF');
```
Sesquipedalian: Don't forget cron.php, install.php, upgrade.php, and proxy.php.

Sources/Class-CurlFetchWeb.php
jdarwood007: The stats collection scripts do not validate a user agent.

MissAllSunday: What user agent should we change the constant to?

Sesquipedalian: This would probably work for all our needs:
```
define('SMF_USER_AGENT', 'Mozilla/5.0 (' . php_uname('s') . ' ' . php_uname('m') . ') AppleWebKit/605.1.15 (KHTML, like Gecko) PHP/' . PHP_VERSION . ' SMF/' . strtr(SMF_VERSION, ' ', '.'));
```

Sesquipedalian: Oh, one more thing. Here in Class-CurlFetchWeb.php we don't want the constant to include the `user-agent: ` bit at the start, so we should define the constant as just `Mozilla/5.0 (...` and then prepend the `user-agent: ` bit as necessary elsewhere.

albertlast: Did you guys had in mind that this information is send to every page which is called from smf?

Many of this information shouldn't make public (you don't know if the connection is encrypted) and the receiver trust worthy.

Secondly you should test this agent vs many side to get a better "feeling" if this custom ua realy create no new issues.

Sesquipedalian: I don't see how this basic info could create a security issue. There are, of course, plenty of people out there with far more knowledge than me on this matter, so I could be wrong. But since none of those people working at Mozilla, Apple, Google, or Microsoft seem to think there is any risk from putting such information in their UA strings, I think we're safe.

Sesquipedalian: I suppose we can leave out the PHP version part if that makes everyone more comfortable. It's not like that bit really matters, after all.

MissAllSunday: Yeah these days sending the server name and the PHP version doesn't do much, considering it is SMF the one that is actually sending that info while trying to communicate with others rather than trying to fetch said info from them.

If you worry about someone intercepting the communication and getting the info, they just don't bother doing that, its way more easy and profitable to just smash everyone with everything and wait until something out of it works.

albertlast: That had nothing related with interception,
you changed the code of the proxy too,
so you send this information all over the place,
when ssl image caching is enabled.

Sesquipedalian (approved):

albertlast (changes requested):
The request create new issues,
like sending server information across to different sites and
also smf version with no needs.

Also the user agent could be blocked on some pages,
which would break the proxy logic,
test would be needed but no one is able to provide.

Sesquipedalian:
On further consideration, I suppose we should drop the PHP version info out of it. It isn't relevant information for a UA, and it's the one part that could conceivably have any sort of security implications. Concerns about the rest of the info are unfounded.

If you are concerned that servers won't like this UA, @albertlast, you can easily test it for yourself against any servers you like. Just use your browser's custom user agent feature or curl's `--user-agent` option on the command line to test as many URLs as you want. If you find any sites that object to `Mozilla/5.0 (Linux x86_64) AppleWebKit/605.1.15 (KHTML, like Gecko) SMF/2.1.RC2`, by all means feel free to come back and tell me how wrong I have been. But in my tests I haven't found a single one that behaved any differently with that UA than they did if I connected using a standard Chrome or Safari UA. 

FYI, the general pattern for WebKit-based browsers is `Mozilla/5.0 (<system and browser information>) AppleWebKit/<version> (KHTML, like Gecko) <extended info>`. Any server that sees a UA matching this pattern is going to accept that they are dealing with some sort of WebKit-based browser.

Sesquipedalian:
@MissAllSunday, let's go ahead and remove the PHP version info part after all. Since our own stats collection doesn't care about the UA, including it gains us nothing and we may as well not bother. Let's keep the SMF version part, though, since that could conceivably be useful info for various reasons.

albertlast (approved):

PR #5711, opened 2019-06-11
Fix reference error

SychO9:
was still getting "**Only variables should be passed by reference**" every time I open an alert (php7.3)

### DIFF
--- a/Sources/Profile-View.php
+++ b/Sources/Profile-View.php
@@ -600,7 +600,7 @@ function showAlerts($memID)
 	if (!empty($_REQUEST['alert']))
 	{
 		$alert_id = (int) $_REQUEST['alert'];
-		$alerts = $alerts = fetch_alerts($memID, $alert_id);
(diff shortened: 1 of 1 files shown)


Sources/Profile-View.php
```diff
@@ -600,8 +600,8 @@ function showAlerts($memID)
 	if (!empty($_REQUEST['alert']))
 	{
 		$alert_id = (int) $_REQUEST['alert'];
-
-		$alert = array_pop(($alerts = fetch_alerts($memID, $alert_id)));
+		$alerts = $alerts = fetch_alerts($memID, $alert_id);
```
Sesquipedalian: I think we only need one `$alerts = ` on this line. 🙂

Sesquipedalian (changes requested):

Sources/Profile-View.php
SychO9: weird how I didn't notice that..

Sesquipedalian (approved):

PR #5710, opened 2019-06-11
Check for notification opt-out setting change

sbulen:
Fixes #5607

PR #5714, opened 2019-06-12
Honor disable announcement setting

### LINKED ISSUE #5011: Allow users to disable announcements: OFF, notification still contains info how to disable it
Hi all!

I noticed that even when the setting:

Admin > Configuration > General > Allow users to disable announcements

is unchecked (set to OFF), at the bottom of notification email users still see this message:

> To unsubscribe from these announcements, login to the forum and uncheck "Receive forum announcements and important notifications by email." in your profile.

Because users cannot change this setting (as it was disabled via Admin) this part of notification should be removed.

sbulen:
Fixes #5011 

Made display of the unsub message conditional.  

Brought behavior, the prompts & the help text all in sync:  
 - If user disabling of announcements is allowed, the unsub message is in the newsletter, and the user "receive newsletter & announcement" selection is honored for both new topic announcements & newsletter mailings.  
 - If user disabling of announcements is NOT allowed, the unsub message is NOT in the newsletter, and the user "receive newsletter & announcement" selection is ignored for new topic announcements & newsletter mailings.


sbulen:
Needs #5710 to test properly...

Sesquipedalian:
FYI, this impinges directly on matters involving GDPR compliance, and I am currently doing my best to make sure that there is as much compatibility as possible between 2.0.16's GDPR compliance systems and 2.1's. So it is quite possible that this will ultimately require significant changes or may be rejected in favour of an alternative that more closely mirrors the 2.0.16 version. Just so you aren't surprised later. 🙂

sbulen:
I was wondering if the option should be removed altogether.

Note:  post thoughts like this on topics *before* folks contribute - or self-assign. So we dont waste our time.  

Trying to help with items flagged as rc3.  Think of the rc3 flag as "please help here".

Sesquipedalian:
Fair point. I dropped the ball on that one. 🙂 Also, to be clear, I'm not saying it will be rejected. I haven't even had a chance to review it yet. I just suspect that it may need some adjustments.

sbulen:
NP.  Just a reminder.

One thing that has been bugging me about this is the knowledge that some implementations are internal, e.g., for clubs, schools, employees, staff, etc.  Not public facing, some certainly even behind a firewall.  I'm wondering if all the same restrictions apply universally.  I.e., maybe there are valid uses for not allowing users to opt out.  Just thinking aloud...

Sesquipedalian (approved):

PR #5722, opened 2019-06-22
Change colum alter api of nullable

albertlast:
By looking at this code: 
```
$column = array(
	'name' => 'join_reason',
	'type' => 'tinytext',
);

$smcFunc['db_add_column']('{db_prefix}members', $column);
```
https://www.simplemachines.org/community/index.php?topic=568346.msg4022474#msg4022474

i came to the result that the mod author did everything right,
since in mysql and pg is the default behavior that  columns are nullable by default: https://www.db-fiddle.com/f/hATtEpH78Qw3eWsA3DR3JT/0

So when the author did nothing wrong,
than the db api of smf do it wrong.

The behavior is that every column is by creation not nullable AND
got no default value.
On the other hand in postgres is the default behavior that every column is nullable,
because of a coding error.

Based on this i changed the api,
first i changed 'null' to 'not_null' because this is the behavior of sql,
you don't say when you want null you say when you want not_null.

Secondly i changed the pg code in a away that it's behave now in the same way like mysql.

By afraid that the upgrade.php and maybe some mods really use 'null' i added the information back,
but without mention this in the function api.

Would be nice if this change get also into 2.0.x

### DIFF
--- a/Sources/DbPackages-mysql.php
+++ b/Sources/DbPackages-mysql.php
@@ -81,7 +81,7 @@ function db_packages_init()
  *  	'size' => Size of column (If applicable) - for example 255 for a large varchar, 10 for an int etc.
  *  		If not set SMF will pick a size.
  *  	- 'default' = Default value - do not set if no default required.
- *  	- 'null' => Can it be null (true or false) - if not set default will be false.
+ *  	- 'not_null' => Can it be null (true or false) - if not set default will be false.
  *  	- 'auto' => Set to true to make it an auto incrementing column. Set to a numerical value to set from what
  *  		 it should begin counting.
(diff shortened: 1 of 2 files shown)


MissAllSunday:
I'm kinda worry about this change, specially for backward compatibility.

albertlast:
Can you be more precisly?

MissAllSunday:
Thats the thing, I can't. It's just a gut feeling about this change being too drastic.

albertlast:
Because the mod author never want a not null without default value,
i see not the problem.

MissAllSunday:
Let's tag it for final and see its behavior.

MissAllSunday (approved):

PR #5729, opened 2019-07-07
Prune log topics

### LINKED ISSUE #5657: Should have a scheduled task for pruning log_topics
#### Description
Question: Should we have a scheduled task for pruning log_topics?

I believe 2.1 has the same behavior as 2.0, in that, log_topics can grow out of hand.  In 2.0, we have a script that is shared (informally, in a post) to clean that up.  

https://www.simplemachines.org/community/index.php?topic=212330.msg1667071#msg1667071

sbulen:
Fixes #5657


Sources/ScheduledTasks.php
```diff
@@ -1542,4 +1542,196 @@ function scheduled_remove_old_drafts()
...
+
+	// Find the oldest $maxMembers in log_boards
+	$sql = "SELECT lb.id_member, m.last_login
+		FROM {db_prefix}members m
+		INNER JOIN
+		(
+			SELECT DISTINCT id_member
+			FROM {db_prefix}log_boards
+		) lb
+		ON lb.id_member = m.id_member
```
albertlast: I would move the "on" part one tab to the right,
since it's part of the join statement or move a line up with the alias.

Sources/ScheduledTasks.php
```diff
@@ -1542,4 +1542,196 @@ function scheduled_remove_old_drafts()
...
+
+	// Find the oldest $maxMembers in log_mark_read
+	$sql = "SELECT lmr.id_member, m.last_login
+		FROM {db_prefix}members m
+		INNER JOIN
+		(
+			SELECT DISTINCT id_member
+			FROM {db_prefix}log_mark_read
+		) lmr
+		ON lmr.id_member = m.id_member
```
albertlast: I would move the on part on tab to the right,
since it's part of the join statement or move a line up with the alias.

Sources/ScheduledTasks.php
```diff
@@ -1542,4 +1542,196 @@ function scheduled_remove_old_drafts()
...
+	// Note the different cutoff, as we may need to mark these as read
+	$sql = "SELECT lt.id_member, m.last_login
+		FROM {db_prefix}members m
+		INNER JOIN
+		(
+			SELECT DISTINCT id_member
+			FROM {db_prefix}log_topics
+			WHERE unwatched = {int:unwatched}
+		) lt
+		ON lt.id_member = m.id_member
```
albertlast: I would move the on part on tab to the right,
since it's part of the join statement or move a line up with the alias.

Sources/ScheduledTasks.php
```diff
@@ -1542,4 +1542,196 @@ function scheduled_remove_old_drafts()
...
+				WHERE id_member = {int:member}";
+			$smcFunc['db_query']('', $sql,
+				array(
+					'member' => $id,
+				)
+			);
+			// Delete rows from log_topics
+			$sql = "DELETE FROM {db_prefix}log_topics
+				WHERE id_member = {int:member}
+				AND unwatched = {int:unwatched}";
```
albertlast: move the line one tabe to right,
since "and" is part of the where

Sources/ScheduledTasks.php
```diff
@@ -1542,4 +1542,196 @@ function scheduled_remove_old_drafts()
...
+			// but we just want one row each, & the ID of the last message read in each board.
+			$sql = "SELECT lt.id_member, t.id_board, MAX(lt.id_msg) AS id_last_message
+				FROM {db_prefix}topics t
+				INNER JOIN
+				(
+					SELECT id_member, id_topic, id_msg
+					FROM {db_prefix}log_topics
+					WHERE id_member =  {int:member}
+				) lt
+				ON lt.id_topic = t.id_topic
```
albertlast: I would move the on part on tab to the right,
since it's part of the join statement or move a line up with the alias.

Sources/ScheduledTasks.php
```diff
@@ -1542,4 +1542,196 @@ function scheduled_remove_old_drafts()
...
+				array('id_member' => 'int', 'id_board' => 'int', 'id_msg' => 'int'),
+				$boards,
+				array('id_member', 'id_board')
+			);
+
+			// Finally, delete this user's rows from log_topics
+			// Keep 'unwatched' settings, for when they come back
+			$sql = "DELETE FROM {db_prefix}log_topics
+				WHERE id_member = {int:member}
+				AND unwatched = {int:unwatched}";
```
albertlast: move the "and" one tab to the right

Sources/ScheduledTasks.php
```diff
@@ -1542,4 +1542,196 @@ function scheduled_remove_old_drafts()
...
+		WHERE m.last_login <= {int:cutoff}
+		ORDER BY m.last_login
+		LIMIT {int:limit}";
+	$result = $smcFunc['db_query']('', $sql,
+		array(
+			'limit' => $maxMembers,
+			'cutoff' => $cleanupBeyond,
+		)
+	);
+	// Using id_member as key basically eats dupes...
```
albertlast: Would combine all three selects to one by using union,
```
select ... from asdf
union
select ... from asdf2
```
the union would also eliminate all dupes.

Sources/ScheduledTasks.php
```diff
@@ -1542,4 +1542,196 @@ function scheduled_remove_old_drafts()
...
+	while($row = $smcFunc['db_fetch_assoc']($result))
+		$members[$row['id_member']] = $row['last_login'];
+	$smcFunc['db_free_result']($result);
+
+	// Sort by last_login & limit to $maxMembers entries
+	asort($members);
+	$members = array_slice($members, 0, $maxMembers, TRUE);
+
+	// Either mark boards as read or purge entries, based on last_login
+	foreach($members as $id => $last_login)
```
albertlast: To run for every member throw the delete 3 delete commands is not effience,
you should rather collect all member and use the `` IN ({array_int:members)`` style
for exp.
```
			// Delete rows from log_boards
			$sql = 'DELETE FROM {db_prefix}log_boards
				WHERE id_member in ( {array_int:members} )';
			$smcFunc['db_query']('', $sql,
				array(
					'members' => $ids,
				)
			);
```

albertlast (changes requested):
Would like to see some more tabs,
also a more effenice delete loop and
most important please use single quotes for sql strings and not double quotes.

Sources/ScheduledTasks.php
sbulen: Will do...

Sources/ScheduledTasks.php
sbulen: Will do...

Sources/ScheduledTasks.php
sbulen: Will do...

Sources/ScheduledTasks.php
sbulen: Will do...

Sources/ScheduledTasks.php
sbulen: Will do...

Sources/ScheduledTasks.php
sbulen: Will do...

Sources/ScheduledTasks.php
sbulen: I will test this to see if it works out to be any more efficient.  If so, will do.   (I thought smcfunc didn't like UNIONs?)

Sources/ScheduledTasks.php
sbulen: Will test, & if faster, will do.  

This could all be collapsed into just 4 queries per execution - 3 LARGE deletes & 1 LARGE insert...

sbulen:
Thanks for the input.  Will likely take a few days to test properly.

Sources/ScheduledTasks.php
albertlast: dunno how many entries this delete array can have,
alternative you could do some kind of batch size,
for exp. fire the the delete statements when the array size is 500 or the last run,
instead of to collect them all.

Sources/ScheduledTasks.php
sbulen: The UNION doesn't make anything more efficient, but it is more readable, so I am implementing that change.   

Regarding my other note:  2.0 tries (and fails) to prevent UNIONs, but 2.1 does not.

Sources/ScheduledTasks.php
sbulen: I am testing restructures of this logic, and performance improvements vary - I've been seeing _4x - 10x_ improvements in speed.  It varies based on environment (WAMP vs linux) and # of updates, but a minimum of a 4x improvement.  

Good input.

albertlast:
don't forget to replace the double quotes with single quotes

Sources/ScheduledTasks.php
sbulen: Done

sbulen:
All suggested changes made.  

Tested on WAMP w/ php7.1 & on hosted shared linux w/ php5.4.

It's pretty good at deleting millions of records pretty quickly...

Sources/ScheduledTasks.php
```diff
@@ -1542,4 +1542,178 @@ function scheduled_remove_old_drafts()
...
+		);
+	}
+
+	// Nothing left to do?
+	if (empty($markReadMembers))
+		return true;
+
+	// Find board inserts to perform...
+	$boards = array();
+	foreach($markReadMembers as $id)
```
albertlast: Is ther a reason why you run this select for every id_member single instead of run this once for all?

Sources/ScheduledTasks.php
```diff
@@ -1542,4 +1542,178 @@ function scheduled_remove_old_drafts()
...
+				FROM {db_prefix}log_topics
+				WHERE id_member = {int:member}
+			) lt ON lt.id_topic = t.id_topic
+			GROUP BY lt.id_member, t.id_board';
+		$result = $smcFunc['db_query']('', $sql,
+			array(
+				'member' => $id,
+			)
+		);
+		while($row = $smcFunc['db_fetch_assoc']($result))
```
albertlast: maybe you want to use `` $smcFunc['db_fetch_all']($result) `` instead of this while loop.

albertlast (commented):
Sry that i didn't saw this two points on my first run throw.

Sources/ScheduledTasks.php
sbulen: I tried both suggestions.  Slightly more readable & slight performance improvements.

albertlast (approved):

sbulen:
@MissAllSunday - let me know if we were still going to consider this.

One nagging thought I had about this is that '0 days' means '0 days' here.  We often use 0 to mean 'diable'.  That's not the case here.  I.e., if someone set all the parameters to 0, thinking they were disabling the feature, all 'board read' info would be purged entirely instead...  

If we're going to proceed, I'll address the merge conflicts & make 0 mean 'disabled'...

MissAllSunday:
Yes, for the sake of keep things as normalized as possible we should go for the 0 means disable as the rest of the settings behave even if this means it won't be possible to prune logs for 0 days.

sbulen:
OK.  It's been so long the diffs are unwieldy, so I am going to rebase this.  Then, tweak it a bit, & retest.

albertlast:
i would recommand to follow the sql join standard,
this mean the base table should be in the on statement part on the left side,
see pr #6202
other stuff looks good

sbulen:
Changes & retests completed.  Feedback welcome.

live627:
> i would recommand to follow the sql join standard,
> this mean the base table should be in the on statement part on the left side,

Proof?

albertlast:
Proof is delivered in the mention pr from me

MissAllSunday (approved):

PR #5730, opened 2019-07-07
Adds $smcFunc['random_bytes']() function

Sesquipedalian:
Needed for various soon-to-be submitted changes ported from 2.0.16

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -251,6 +251,18 @@ function reloadSettings()
 
 			return random_int($min, $max);
 		},
+		'random_bytes' => function($length = 8)
+		{
+			global $sourcedir;
+
+			if (!is_callable('random_bytes'))
+				require_once($sourcedir . '/random_compat/random.php');
+
+			// Make sure length is valid
+			$length = max(1, (int) $length);
+
+			return random_bytes($length);
+		},
 	);
 
 	// Setting the timezone is a requirement for some functions.


Sources/Load.php
```diff
@@ -251,6 +251,18 @@ function reloadSettings()
 
 			return random_int($min, $max);
 		},
+		'random_bytes' => function($length = 8)
```
albertlast, line 254: Any reason for 8?
sounds very short to me.

Sesquipedalian: In this case, the default is not meant to suggest a typical value. This one exists merely to prevent `random_bytes()` from generating a fatal error if someone forgets to pass a length parameter to `$smcFunc['random_bytes']()`.

albertlast: for this you add the max() check below,
with this you prevent this happen,
the default value here is based on this a suggestion.

Sesquipedalian: `max(1, (int) $length)` will not work unless `$length` is set to some non-null value first. I could have left `$length` without a default, of course, but I figured that if the point of creating a wrapper function  like this is to make things smoother and easier for developers, then defining a sane default in order to avoid errors just made sense. 

The default of 8 was just an arbitrary decision. I wanted something small but not too small. That's all.

I'm not sure why this matters to you. If you have a reason to change it, please explain.

albertlast: Well for me is random_bytes a security function in the first place and
a byte length shorter than any recommand password length sound wrong to me.

8 char length password need some hours to get cracked: https://www.betterbuys.com/estimating-password-cracking-times/
and anything based on this got variation of 8 byte intial vector...

I would set as default 64 and
when someone need less he can change it to 8.
Performance wise should be this only hurt people which use php below 7.0,
on the time where smf 2.1 get released is this a very minor issue in my eyes.

albertlast: Small test script i builded:
```
<?php
$loops = [1000,10000,100000];
$mods = [8,16,64];

foreach($loops as $loop) {
    foreach($mods as $mod) {  
        $start = microtime(true);
        for($i = 0; $i< $loop; $i++){
            $temp = random_bytes($mod);
        }
        echo ' modus ' . $mod . ' loop ' . $loop . ' runtime '. (microtime(true) - $start) . '<br>' . "\n"; 
    }
}
```

```
 modus 8 loop 1000 runtime 0.001431941986084<br>
 modus 16 loop 1000 runtime 0.0013267993927002<br>
 modus 64 loop 1000 runtime 0.0017960071563721<br>
 modus 8 loop 10000 runtime 0.013478994369507<br>
 modus 16 loop 10000 runtime 0.013724088668823<br>
 modus 64 loop 10000 runtime 0.017735004425049<br>
 modus 8 loop 100000 runtime 0.111741065979<br>
 modus 16 loop 100000 runtime 0.093127012252808<br>
 modus 64 loop 100000 runtime 0.12503004074097<br>
```

From point of performance the diff is very small.

Sesquipedalian: It has now been updated from 8 to 64.

PR #5732, opened 2019-07-11
Add Cloudflare IP Support to the auto detection

albertlast:
I like the idea from https://www.simplemachines.org/community/index.php?topic=532453.0
feel free to decide if this is right or not.

### DIFF
--- a/Sources/QueryString.php
+++ b/Sources/QueryString.php
@@ -248,11 +248,6 @@ function cleanRequest()
 		exit;
 	}
 
-	// Check for Cloudflare
-	if (isset($_SERVER['HTTP_CF_CONNECTING_IP'])) {
-		$_SERVER['REMOTE_ADDR'] = $_SERVER['HTTP_CF_CONNECTING_IP'];
-	}
-
 	// Make sure we have a valid REMOTE_ADDR.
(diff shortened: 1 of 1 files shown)


tinoest:
Would it not make more sense to have a integration hook for any potential future change then a mod for the custom implementations others use?

albertlast:
query string is called very early,
to add here hooks is dangerous.

And cloudflare by his own is very big,
so it's the most common one here.

jdarwood007:
Check a bit further down in the code, we have support for proxy headers there.

VBGAMER45:
I think this is a great idea. People will use that service often. And it is great to have support out of the box.

Kindred-999:
According to Arantor, 2.1 already HAS this capability with a switch in the admin/security section

albertlast:
but it miss on the auto detection.

vort7u:
Has anyone tested this commit?
With SMF RC4 all messages have the ip address of Cloudflare.

Sesquipedalian:
> With SMF RC4 all messages have the ip address of Cloudflare.

Please open a new issue for this. Comments on closed issues and PRs are easily overlooked.

Issue #5731, opened 2019-07-08
Link to delete Package on uninstall package


VBGAMER45:
As a mod developer one thing that annoys me is when I uninstall a package. Normally I want to delete the package. This usually involves going back to package manager file listing and finding the mod to delete.  It would be cool to add a link to delete the package on the same page that says the package has been uninstalled.

albertlast:
Could you add a image how you think this should look like?

VBGAMER45:
Sure something really basic like this screenshot


![deletepackage](https://user-images.githubusercontent.com/978002/60909644-62c40480-a244-11e9-9382-76492ad08d97.png)

MissAllSunday:
Closed via #5733

PR #5733, opened 2019-07-11
Add remove link to the finished uninstall window

albertlast:
I hope it fit somehow the base idea:
grafik

issue: #5731


LexArma:
Looks good, but I think from a usability point of view, it might be good if the text was altered to explain the button. For example "The package has been successfully uninstalled, you may now delete the package".

Sesquipedalian:
Just to prevent possible confusion, let’s have a “Delete Package” button and a “Keep Package” button. The second would really just be a link to the Browse Packages page, but it would help admins to know that they don’t have to delete the package. If there is only one button, some admins might think deleting the package is mandatory rather than optional.

MissAllSunday:
@albertlast Can you address the point Sesquipedalian make so this can be merged.

albertlast:
done
grafik

MissAllSunday (approved):

Issue #5734, opened 2019-07-12
getBoardindex do call parse_bbc which takes 94% of the run time


albertlast:
#### Description
notice in my test board,
that the generation goes up from ~ 0.060 ms to ~0.340 ms.
By cecking the trace i notice that the parse_bbc and calls below create 94% of the call time:

![grafik](https://user-images.githubusercontent.com/1782906/61103166-38a35a00-a471-11e9-94eb-a1a65c0c3512.png)

trace file: https://mega.nz/#!W8piDSKa!6qkqASjhWj80unkS8uAY_XvFW5iwHYRhJZtdjjScyik

issue is introduce by pr: #5704 from @MissAllSunday 
because of this line: 
https://github.com/SimpleMachines/SMF2.1/blob/e941ff2b0b31fe44a8aeb9ca435ff6231d5eaa26/Sources/Subs-BoardIndex.php#L263
### Steps to reproduce
1.  time notice on the start page

### Environment (complete as necessary)
- Version/Git revision: e941ff2b0
- Database Type:
- Database Version:
- PHP Version: 7.2

### Additional information/references

MissAllSunday:
Yeah, this is a side effect of moving the parsing to display.

Since the board and category descriptions won'y be updated that often, we can save the parsed description on cache

albertlast:
Well from performance point of view,
the feature double(with improved build_regex) the time of the smf start page,
so it should double the user exprience otherwise is questionable to take so mutch time for not worth function.

A middle way in my eyes could be the right way:
to safe the information twice in the database,
in one column the orginal version(this is would be the string which we use now) and display/optimize column where the parsed version is filled in.

So when you editing/changing you work with the original column and
when you display this information you use the display column

MissAllSunday:
Yeah, that's pretty much what the cache does already only without the burden on adding a new column on the db.

This really belongs to a cache entry, specially since using a proper cache system will be far better than a DB entry.

If you are worried about users not using any cache, then there isn't much we can do, perhaps instructing users to use it whenever possible.

albertlast:
from performance view, is the cache slower than database in this case.
because you need to run a select anyway, so a column more or less doesn't hurt,
but to do 193 cache calls add to the time.

MissAllSunday:
Thing is, if we see everything from a performance view then we will never finish doing performance optimizations.

We have a cache feature, this case can benefit from using said feature, lets use it. Even if there are better ways to do it.

[...]

PR #5333, opened 2019-01-07
parsebbc() fixes

Sesquipedalian:
- Fixes #5065 
- By popular demand, brings [br] out of retirement
- Some efficiency improvements in code I added earlier

PR #5735, opened 2019-07-12
Cache the value from build_regex

albertlast:
Well it mitigate the issue #5734
but doesn't fixed it.
parse_bbc is still using 61% off the runtime
but in time it goes down from 300 ms to 100ms.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -6520,16 +6520,16 @@ function set_tld_regex($update = false)
 function build_regex($strings, $delim = null, $returnArray = false)
 {
 	global $smcFunc;
-	static $cache_buids = array();
+	static $regexes = array();
 
 	// If it's not an array, there's not much to do. ;)
 	if (!is_array($strings))
 		return preg_quote(@strval($strings), $delim);
 
-	$cache_key = md5(json_encode($strings)) .'d=' . $delim . 'r=' . $returnArray;
+	$regex_key = md5(json_encode(array($strings, $delim, $returnArray)));
 	
-	if (isset($cache_buids[$cache_key]))
-		return $cache_buids[$cache_key];
+	if (isset($regexes[$regex_key]))
+		return $regexes[$regex_key];
 
 	// The mb_* functions are faster than the $smcFunc ones, but may not be available
 	if (function_exists('mb_internal_encoding') && function_exists('mb_detect_encoding') && function_exists('mb_strlen') && function_exists('mb_substr'))
@@ -6672,7 +6672,7 @@ function build_regex($strings, $delim = null, $returnArray = false)
 	if (!empty($current_encoding))
 		mb_internal_encoding($current_encoding);
 
-	$cache_buids[$cache_key] = $regex;
+	$regexes[$regex_key] = $regex;
 	return $regex;
 }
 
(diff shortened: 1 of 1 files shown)


MissAllSunday:
Just curious, why adding cache to build_regex?  by adding cache I actually meant saving the board_description var into a cache entry with the board id in the entry name to get the correct entry when displaying the boards.

albertlast:
I didn't write the code based on your comment,
i notice that build_regex is complex and get called many times with the same input,
to cache this input, output combination is good for general perfomance of smf.

Also this code improve setups where no cache is enabled.

Sesquipedalian:
It's better for the calling function to cache the result of build_regex(). If parse_bbc() repeatedly calls build_regex() with the same input, parse_bbc() is doing something wrong.

albertlast:
Yes the parse bbc should do something,
but it could happen that different function like parse bbc,preparsecode, parsesmileys, set_tld_regex or a mod
do build the same build_regex like the other function,
So to cache this result i see no pain.

To first point it look like that you try to fix parse_bbc,
but you wrote this in a way that this doesn't work:
https://github.com/SimpleMachines/SMF2.1/blob/e941ff2b0b31fe44a8aeb9ca435ff6231d5eaa26/Sources/Subs.php#L2252-L2266
Pr: #5333

Sesquipedalian:
Yes, that was a mistake on my part. I should have used an array to remember all variants of `$alltags_regex` rather than just saving and restoring the default version every time.

Sesquipedalian:
As for caching inside build_regex, the problem with that is that doing so would bloat the cache with unnecessary, redundant data. The calling function knows best how it will use the results from build_regex, so it should be left to the calling function to decide how and where to save and retrieve the results.

albertlast:
The cache is only a static var so there is no bloating  of the cache.
How should  a calling function know if a different function already ask for the same build regex?

Sources/Subs.php
```diff
@@ -6520,11 +6520,17 @@ function set_tld_regex($update = false)
 function build_regex($strings, $delim = null, $returnArray = false)
 {
 	global $smcFunc;
+	static $cache_buids = array();
```
Sesquipedalian: Please change `$cache_buids` to `$regexes` here and in the subsequent code.

Sources/Subs.php
```diff
@@ -6520,11 +6520,17 @@ function set_tld_regex($update = false)
...
 
 	// If it's not an array, there's not much to do. ;)
 	if (!is_array($strings))
 		return preg_quote(@strval($strings), $delim);
 
+	$cache_key = md5(json_encode($strings)) .'d=' . $delim . 'r=' . $returnArray;
```
Sesquipedalian: Please change `$cache_key` to `$regex_key` here and in the subsequent code.

Sesquipedalian: Please change 
```
md5(json_encode($strings)) .'d=' . $delim . 'r=' . $returnArray; 
```
to 
```
md5(json_encode(array($strings, $delim, $returnArray))); 
```
This will be more reliable.

Sesquipedalian (changes requested):
> How should  a calling function know if a different function already ask for the same build regex?

That is a very good point. Okay, you have convinced me. Please make the requested changes, and then I will merge this PR.

albertlast:
Changes are done.

Sesquipedalian (approved):

PR #5736, opened 2019-07-12
improve boards category parse_bbc fixes #5734

### LINKED ISSUE #5734: getBoardindex do call parse_bbc which takes 94% of the run time

albertlast:
the base idea is to build the best of two worlds,
on the one hand being able to modify the category name/description and board name/dscription with bbc,
on the other hand being fast to allow big board run in the same speed like small one.

since we do db changes in the process,
we can do a mutch better work on upgrade process check if the changes needs to run or not.

missing:
upgrade process is not final
edit logic is missing

issue: #5734

### DIFF
--- a/Sources/Subs-BoardIndex.php
+++ b/Sources/Subs-BoardIndex.php
@@ -95,8 +95,8 @@ function getBoardIndex($boardIndexOptions)
 	else
 		$result_boards = $smcFunc['db_query']('', '
 			SELECT' . ($boardIndexOptions['include_categories'] ? '
-				c.id_cat, c.name AS cat_name, c.description AS cat_desc,' : '') . '
-				b.id_board, b.name AS board_name, b.description,
+				c.id_cat, c.name AS cat_name, c.name_disp AS cat_name_disp,c.description AS cat_desc, c.description_disp AS cat_desc_disp,' : '') . '
+				b.id_board, b.name AS board_name,b.name_disp AS board_name_disp, b.description, b.description_disp,
 				CASE WHEN b.redirect != {string:blank_string} THEN 1 ELSE 0 END AS is_redirect,
 				b.num_posts, b.num_topics, b.unapproved_posts, b.unapproved_topics, b.id_parent,
 				COALESCE(m.poster_time, 0) AS poster_time, COALESCE(mem.member_name, m.poster_name) AS poster_name,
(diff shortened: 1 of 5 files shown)


jdarwood007:
This won't work reliably because parse_bbc can have language strings.  We already cache the board index under certain conditions which is what is needed here.

albertlast:
Than we should allow the admin to disable parse_bbc for board and catagory name/description?
In my eyes is the feature not worth that it raise the runtime by 5x (with optimization its double the time).

albertlast:
Or would be not better to ban this tages for description and name,
so that the display version is language undepend?

live627:
Use the cache sub system for this.

albertlast:
Show me that is faster or go away with this solution.

Sesquipedalian:
I'm not going to merge anything that changes the structure of the boards table at this stage.

However, @albertlast is right that it is silly to cache and retrieve each board name and description separately. We should be able to improve performance if we cache the parsed board names and descriptions all at once in a single array.

albertlast:
Since the user based on the right see different part of the forum I'm not sure how well this works. 
But we could save the original and display version in same columns by using json encoding, 
Would imply that we would extend the size of names columns,
Since the description already are text would be here no change needed.

Would this a solution too?

albertlast:
@Sesquipedalian i wait on a comment from you.

Sesquipedalian:
My idea will work if we follow this order:

1. Retrieve the array of parsed board names and descriptions from the cache.
2. Loop through the user's board tree and either:
  a. Get and use the existing parsed value from the cached array if it exists, or
  b. Run parse_bbc to generate the parsed value, use it for display, and append it to the big array of parsed values.
3. When done looping through the board tree, if we appended anything to the array of parsed values, save the new version of the array to the cache.

This method ensures that we only need to parse each board's name and description once, and it also spreads out the workload among multiple users, since the initial parsing for each board only happens when needed.

I think this will be a better approach, especially in the RC stage when we want to avoid database alterations except when absolutely necessary.

albertlast:
I see some problem because the description is a text fields,
which mean it got unlimted size,
so every description can be very big when you combind them into one array you big array get realy big...

But when i take your idea i could imagine a "middle" way.
We could save the name in "big array" and the descirption since they are text column,
we could save ther the parse value directly by side in the same column of the orginal value by using json encode or something similiar.

live627:
See for maximum numbers:
http://dev.mysql.com/doc/refman/5.0/en/storage-requirements.html

    TINYBLOB, TINYTEXT       L + 1 bytes, where L < 2^8    (255 Bytes)
    BLOB, TEXT 	             L + 2 bytes, where L < 2^16   (64 Kilobytes)
    MEDIUMBLOB, MEDIUMTEXT   L + 3 bytes, where L < 2^24   (16 Megabytes)
    LONGBLOB, LONGTEXT 	     L + 4 bytes, where L < 2^32   (4 Gigabytes)

L is the number of bytes in your text field. So the maximum number of chars for text is 2<sup>16</sup>-1 (using single-byte characters). Means 65 535 chars(using single-byte characters).

**UTF-8/MultiByte encoding**: using MultiByte encoding each character might consume more than 1 byte of space. For UTF-8 space consumption is between 1 to 4 bytes per char.

albertlast:
Nothing new to me

live627:
> I see some problem because the description is a text fields,
which mean it got unlimted size,

I just wanted to set the record straight here is all; have a good new year.

PR #5740, opened 2019-07-19
Remove not more supported mysql hints

albertlast:
With mysql 5.7 sql_no_cache got deptraced and with 8.0 got it removed.
src: https://dev.mysql.com/doc/refman/5.7/en/query-cache-in-select.html

I didn't change the hint handling in the db layer,
because maybe some mods use it.

https://github.com/SimpleMachines/SMF2.1/blob/e0848c82fbbc912b51789a2fefded6b6830f086f/Sources/Subs-Db-mysql.php#L367

https://github.com/SimpleMachines/SMF2.1/blob/e0848c82fbbc912b51789a2fefded6b6830f086f/Sources/Subs-Db-postgresql.php#L375

End of the day only the hint: "40000 USE INDEX" is used once:
https://github.com/SimpleMachines/SMF2.1/blob/e0848c82fbbc912b51789a2fefded6b6830f086f/Sources/Subs-Boards.php#L109
otherwise the queries are free from hints.

### DIFF
--- a/Sources/ManageMaintenance.php
+++ b/Sources/ManageMaintenance.php
@@ -855,13 +855,13 @@ function AdminBoardRecount()
 			// Recount approved messages
 			$request = $smcFunc['db_query']('', '
 				SELECT t.id_topic, MAX(t.num_replies) AS num_replies,
-					CASE WHEN COUNT(ma.id_msg) >= 1 THEN COUNT(ma.id_msg) - 1 ELSE 0 END AS real_num_replies
+					GREATEST(COUNT(ma.id_msg) - 1, 0) AS real_num_replies
 				FROM {db_prefix}topics AS t
 					LEFT JOIN {db_prefix}messages AS ma ON (ma.id_topic = t.id_topic AND ma.approved = {int:is_approved})
 				WHERE t.id_topic > {int:start}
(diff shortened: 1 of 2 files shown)


Sources/ManageMaintenance.php
```diff
@@ -854,14 +854,14 @@ function AdminBoardRecount()
 		{
 			// Recount approved messages
 			$request = $smcFunc['db_query']('', '
-				SELECT /*!40001 SQL_NO_CACHE */ t.id_topic, MAX(t.num_replies) AS num_replies,
-					CASE WHEN COUNT(ma.id_msg) >= 1 THEN COUNT(ma.id_msg) - 1 ELSE 0 END AS real_num_replies
+				SELECT t.id_topic, MAX(t.num_replies) AS num_replies,
+					GREATEST(COUNT(ma.id_msg) - 1, 0) AS real_num_replies
```
MissAllSunday, line 858: Just curios if its safe to use GREATEST() with our current min mysql version.  It seems its behavior changed a bit: https://bugs.mysql.com/bug.php?id=28258

albertlast: Since aggration function can't return null and 0 can't be null,
so we don't need to care about this topic.

PR #5742, opened 2019-07-21
Fix missing error messages reducing mail queue

### LINKED ISSUE #5683: Autotask undefined txt strings
#### Description
There is a case where log_error() is executed but the text strings are not defined at that point so the error in the log doesn't say anything.
Read here
https://www.simplemachines.org/community/index.php?topic=567954

sbulen:
Fixes #5683 

Since this is also called early in index.php, we should at least make sure errors are loaded.  Note that some smtp errors are in the index language file.

### DIFF
--- a/Sources/ScheduledTasks.php
+++ b/Sources/ScheduledTasks.php
@@ -607,11 +607,19 @@ function scheduled_weekly_digest()
  */
 function ReduceMailQueue($number = false, $override_limit = false, $force_send = false)
 {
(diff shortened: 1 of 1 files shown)


albertlast:
Is ``$language`` always setted at this point of time?

sbulen:
No, but the loadLanguage() routine handles that possibility.

sbulen:
Thanks, Suki!

PR #5745, opened 2019-07-21
fix function registerMember (Subs-Members.php)

javinotfound:
registerMember doesn't check if username (member_name) exists before allow to register it.

### DIFF
--- a/Sources/Subs-Members.php
+++ b/Sources/Subs-Members.php
@@ -527,7 +527,8 @@ function registerMember(&$regOptions, $return_errors = false)
 	);
 	
 	// @todo Separate the sprintf?
-	if ($smcFunc['db_num_rows']($request) != 0) {
+	if ($smcFunc['db_num_rows']($request) != 0)
+	{
 		list ($membername_db) = $smcFunc['db_fetch_row']($request);
 		if ($membername_db === $regOptions['username']) {
 			$reg_errors[] = array('lang', 'already_a_user', false);
--- a/Sources/Subs-Members.php
+++ b/Sources/Subs-Members.php
@@ -517,7 +517,7 @@ function registerMember(&$regOptions, $return_errors = false)
 	$request = $smcFunc['db_query']('', '
 		SELECT member_name
 		FROM {db_prefix}members
-		WHERE email_address = {string:email_address} 
+		WHERE email_address = {string:email_address}
 				OR member_name = {string:username}
 		LIMIT 1',
 		array(
--- a/Sources/Subs-Members.php
+++ b/Sources/Subs-Members.php
@@ -518,16 +518,28 @@ function registerMember(&$regOptions, $return_errors = false)
 		SELECT id_member
 		FROM {db_prefix}members
 		WHERE email_address = {string:email_address}
-			OR email_address = {string:username}
 		LIMIT 1',
-		array(
-			'email_address' => $regOptions['email'],
-			'username' => $regOptions['username'],
-		)
+		array('email_address' => $regOptions['email'])
 	);
+	
 	// @todo Separate the sprintf?
 	if ($smcFunc['db_num_rows']($request) != 0)
 		$reg_errors[] = array('lang', 'email_in_use', false, array($smcFunc['htmlspecialchars']($regOptions['email'])));
+
+	$smcFunc['db_free_result']($request);
+	
+	// Check if the username is in use.
+	$request = $smcFunc['db_query']('', '
+		SELECT id_member
+		FROM {db_prefix}members
+		WHERE member_name = {string:username}
+		LIMIT 1',
+		array('username' => $regOptions['username'])
+	);
+	
+	// @todo Separate the sprintf?
+	if ($smcFunc['db_num_rows']($request) != 0)
+		$reg_errors[] = array('lang', 'already_a_user', false);
 
 	$smcFunc['db_free_result']($request);
 
--- a/Sources/Subs-Members.php
+++ b/Sources/Subs-Members.php
@@ -517,25 +517,21 @@ function registerMember(&$regOptions, $return_errors = false)
 	$request = $smcFunc['db_query']('', '
 		SELECT member_name
 		FROM {db_prefix}members
-		WHERE email_address = {string:email_address}
-				OR member_name = {string:username}
+		WHERE email_address = {string:email_address} 
+			OR member_name = {string:username}
 		LIMIT 1',
 		array(
-			'email_address' => $regOptions['email'],
+			'email_address' => $regOptions['email'], 
 			'username' => $regOptions['username'],
 		)
 	);
 	
 	// @todo Separate the sprintf?
-	if ($smcFunc['db_num_rows']($request) != 0)
-	{
+	if ($smcFunc['db_num_rows']($request) != 0) {
 		list ($membername_db) = $smcFunc['db_fetch_row']($request);
-		if ($membername_db === $regOptions['username'])
-		{
+		if ($membername_db === $regOptions['username']) {
 			$reg_errors[] = array('lang', 'already_a_user', false);
-		}
-		else
-		{
+		} else {
 			$reg_errors[] = array('lang', 'email_in_use', false, array($smcFunc['htmlspecialchars']($regOptions['email'])));
 		}
 	}
--- a/Sources/Subs-Members.php
+++ b/Sources/Subs-Members.php
@@ -517,15 +517,19 @@ function registerMember(&$regOptions, $return_errors = false)
 	$request = $smcFunc['db_query']('', '
 		SELECT member_name
 		FROM {db_prefix}members
-		WHERE email_address = {string:email_address} OR member_name = {string:username}
+		WHERE email_address = {string:email_address} 
+				OR member_name = {string:username}
 		LIMIT 1',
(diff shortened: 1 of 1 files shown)


albertlast:
i don't understand why you take the mail account check out,
why you create a second query.
could be done by one query by extend the where or use union.

javinotfound:
The problem is that as it was the registerMember function a query is made to see if the email is already being used by another member, but it is not checked at any time if the username is already being used by another user.

Although I know that it could be done through a single query, when reporting which is the exact error (if it is the username that already exists, or on the contrary is the email address) I chose to make 2 queries.

In the current code it seems that the query checks that the user name entered is unique, but this does not happen because in any case 'OR email_address = {string:username}' should be 'OR member_name= {string:username}'.

albertlast:
The problem is clear,
but the solution is unclear,
i would keep the mail check and i would keep this at one query not two.

javinotfound:
The mail check is present in my fix (line 520, email is checked)

I deleted line 521 because it was a mistake (it should be 'OR member_name = {string:username}' instead of 'OR email_address = {string:username}'

I decided make two query instead of one because if you keep that in only one query, when a row is found, it could be username or email. You don't know if the problem is within the email or with the username. You will know that some of them is being used by another member, but no if it is the username or the email.

albertlast:
sure you can know this,
by using case in the colum or use union or check this in php.

javinotfound:
I edited the fix to execute only one query. Now it's ok?

Sources/Subs-Members.php
```diff
@@ -513,22 +513,24 @@ function registerMember(&$regOptions, $return_errors = false)
...
 
-	// Check if the email address is in use.
+	// Check if the email address or the username is in use.
 	$request = $smcFunc['db_query']('', '
-		SELECT id_member
+		SELECT member_name
 		FROM {db_prefix}members
-		WHERE email_address = {string:email_address}
-			OR email_address = {string:username}
+		WHERE email_address = {string:email_address} OR member_name = {string:username}
```
albertlast: get this back the old stylew and change only the member_name column

Sources/Subs-Members.php
```diff
@@ -513,22 +513,24 @@ function registerMember(&$regOptions, $return_errors = false)
...
 		LIMIT 1',
-		array(
-			'email_address' => $regOptions['email'],
-			'username' => $regOptions['username'],
-		)
+		array('email_address' => $regOptions['email'], 'username' => $regOptions['username'])
```
albertlast: get this back to the old style

Sources/Subs-Members.php
```diff
@@ -513,22 +513,24 @@ function registerMember(&$regOptions, $return_errors = false)
...
 	);
+	
 	// @todo Separate the sprintf?
-	if ($smcFunc['db_num_rows']($request) != 0)
-		$reg_errors[] = array('lang', 'email_in_use', false, array($smcFunc['htmlspecialchars']($regOptions['email'])));
-
+	if ($smcFunc['db_num_rows']($request) != 0) {
+		list ($membername_db) = $smcFunc['db_fetch_row']($request);
+		if ($membername_db===$regOptions['username']) {
```
albertlast: add a space on both side of ``===``

albertlast (changes requested):

Sources/Subs-Members.php
javinotfound: Done

Sources/Subs-Members.php
javinotfound: Done

Sources/Subs-Members.php
javinotfound: Done

javinotfound:
All changes suggested done.

Sources/Subs-Members.php
```diff
@@ -513,22 +513,28 @@ function registerMember(&$regOptions, $return_errors = false)
...
 
-	// Check if the email address is in use.
+	// Check if the email address or the username is in use.
 	$request = $smcFunc['db_query']('', '
-		SELECT id_member
+		SELECT member_name
 		FROM {db_prefix}members
-		WHERE email_address = {string:email_address}
-			OR email_address = {string:username}
+		WHERE email_address = {string:email_address}
```
SychO9:
```suggestion
		WHERE email_address = {string:email_address}
```

Sources/Subs-Members.php
```diff
@@ -513,22 +513,28 @@ function registerMember(&$regOptions, $return_errors = false)
...
+		SELECT member_name
 		FROM {db_prefix}members
-		WHERE email_address = {string:email_address}
-			OR email_address = {string:username}
+		WHERE email_address = {string:email_address} 
+				OR member_name = {string:username}
 		LIMIT 1',
 		array(
-			'email_address' => $regOptions['email'],
+			'email_address' => $regOptions['email'],
```
SychO9:
```suggestion
			'email_address' => $regOptions['email'],
```

Sources/Subs-Members.php
```diff
@@ -513,22 +513,28 @@ function registerMember(&$regOptions, $return_errors = false)
...
+			'email_address' => $regOptions['email'], 
 			'username' => $regOptions['username'],
 		)
 	);
+	
 	// @todo Separate the sprintf?
-	if ($smcFunc['db_num_rows']($request) != 0)
-		$reg_errors[] = array('lang', 'email_in_use', false, array($smcFunc['htmlspecialchars']($regOptions['email'])));
-
+	if ($smcFunc['db_num_rows']($request) != 0) {
```
SychO9:
```suggestion
	if ($smcFunc['db_num_rows']($request) != 0)
	{
```

Sources/Subs-Members.php
```diff
@@ -513,22 +513,28 @@ function registerMember(&$regOptions, $return_errors = false)
...
+		list ($membername_db) = $smcFunc['db_fetch_row']($request);
+		if ($membername_db === $regOptions['username']) {
```
SychO9:
```suggestion
		if ($membername_db === $regOptions['username'])
		{
```

Sources/Subs-Members.php
```diff
@@ -513,22 +513,28 @@ function registerMember(&$regOptions, $return_errors = false)
...
+			$reg_errors[] = array('lang', 'already_a_user', false);
+		} else {
```
SychO9:
```suggestion
		}
		else
		{
```

SychO9 (commented):
Excuse the nitpicking, but just to keep the coding style the same as SMF's

Thank you for the fix, no one had noticed this before 😄

Sources/Subs-Members.php
```diff
@@ -513,22 +513,32 @@ function registerMember(&$regOptions, $return_errors = false)
...
 
-	// Check if the email address is in use.
+	// Check if the email address or the username is in use.
 	$request = $smcFunc['db_query']('', '
-		SELECT id_member
+		SELECT member_name
 		FROM {db_prefix}members
 		WHERE email_address = {string:email_address}
-			OR email_address = {string:username}
+				OR member_name = {string:username}
```
albertlast: The tab should be not here.

javinotfound:
Sorry for not respecting SMF coding style. It's my first pull request on open source code.

All changes applied, waiting to travis.

Sources/Subs-Members.php
```diff
@@ -1530,4 +1540,4 @@ function generateValidationCode()
 	return substr(preg_replace('/\W/', '', sha1(microtime() . $smcFunc['random_int']() . $dbRand . $modSettings['rand_seed'])), 0, 10);
 }
 
-?>
\ No newline at end of file
+?>
```
albertlast: you use a different eol,
change it back to the old one.

albertlast (changes requested):
to meet the code style stuff is one the first time always a strugle.

SychO9:
It's allright, I remember my first pull request, it was a mess xD

javinotfound:
I think that all works fine, respecting SMF coding style and tabulations.

albertlast (approved):

albertlast:
I got notice that registerMember call validateUsername
https://github.com/SimpleMachines/SMF2.1/blob/e0848c82fbbc912b51789a2fefded6b6830f086f/Sources/Subs-Members.php#L476

 and in this the function isReservedName is called:
https://github.com/SimpleMachines/SMF2.1/blob/e0848c82fbbc912b51789a2fefded6b6830f086f/Sources/Subs-Auth.php#L681

so the username check should already work.
Without this pr.

javinotfound:
I thought same as you, but for any reason I think that this function (isReservedName) only checks if username is reserved (for example if contains 'admin', 'root'... because this words are reserved by default), but not if username is unique and exists. You can take 2.1 RC2 release and with SSI.php, try to register two users with same username and different email, and you will see that registrationMember will work, creating two users with same username in the forum. Perhaphs we need to change error message  un isReservedName.

albertlast:
For me the line look similiar what you do:
https://github.com/SimpleMachines/SMF2.1/blob/e0848c82fbbc912b51789a2fefded6b6830f086f/Sources/Subs-Members.php#L911-L930

So maybe this function got an issue which should get be fixed?

MissAllSunday:
This seems very similar to what SMF already does, will close for now but please feel free to create another issue with an example of how to reproduce the behavior this PR solves.

Issue #5741, opened 2019-07-20
Upgrader - challenges with 1.1 to 2.1 upgrade


sbulen:
#### Description
Having problems with dates when upgrading from 1.1 to 2.1:  
![image](https://user-images.githubusercontent.com/23568484/61583093-1d2af580-aae8-11e9-80a3-286b311f9d72.png)

### Steps to reproduce
1. Install a 100% vanilla 1.1.21
2. Upgrade to 2.1

### Environment (complete as necessary)
- Version/Git revision:  GH as of today
- Database Type:  MySQL
- Database Version:  5.7
- PHP Version: 7.1

sbulen:
I believe this is due to applying strict mode to this old 1.x syntax:
```
UPDATE {$db_prefix}calendar
SET endDate = startDate
WHERE endDate = '0001-01-01'
	OR endDate = '0000-00-00';

UPDATE {$db_prefix}calendar_holidays
SET eventDate = '0001-01-01'
WHERE eventDate = '0000-00-00';
```
In these cases, it's not trying to STORE the invalid values, it's attempting to REPAIR the invalid values.

albertlast:
The fix would be to place simliar logic from 2.1 here:
https://github.com/SimpleMachines/SMF2.1/blob/e0848c82fbbc912b51789a2fefded6b6830f086f/other/upgrade_2-1_mysql.sql#L3-L27

But the result would be,
that 1.1 upgrade create a database version which is not any more 1.1...
Maybe to remove this logic from 1.1 upgrade would be the most valid way to fix it.

sbulen:
I think we need to... Those old upgrader scripts need to be able to execute under more recent versions of mysql.

An alternative may be to associate each step with it's own sql_mode.

albertlast:
This change will be changed in a later time of the upgrade again,
so it make no sense to keep this.

sbulen:
For the record, 1.1 has a few issues, but 1.0 has MANY issues...  

It should be able to run thru...

sbulen:
The dates aren't the only issue with 1.1 - the 'previous fix detected' kind of logic has at least one issue as well:
![1-0-upgrade-9](https://user-images.githubusercontent.com/23568484/61921788-29adb480-af13-11e9-8cf5-d5ce5c3bc7ba.JPG)

sbulen:
Found another issue if you leave the "classic" theme laying around:
[...]

PR #5746, opened 2019-07-21
Better 1.1 upgrade logic

albertlast:
dunno if i find more stuff,
for the first part i looked into the calendar, calendar_holidays and member table.

All this three table get touched again in 2.1 upgrade script:
https://github.com/SimpleMachines/SMF2.1/blob/e0848c82fbbc912b51789a2fefded6b6830f086f/other/upgrade_2-1_mysql.sql#L3-L36

So it make less sense todo this in 1.1 part before.
Secondly was here mayor mistake done,
the index eventData got dropped (this is for this case here very stupid).

issue: https://github.com/SimpleMachines/SMF2.1/issues/5741

### DIFF
--- a/other/upgrade_1-1.sql
+++ b/other/upgrade_1-1.sql
@@ -622,17 +622,6 @@ CHANGE COLUMN date date date NOT NULL default '0001-01-01';
 UPDATE {$db_prefix}log_activity
 SET date = '0001-01-01'
 WHERE date = '0000-00-00';
-
-ALTER TABLE {$db_prefix}members
-CHANGE COLUMN birthdate birthdate date NOT NULL default '0001-01-01';
-
-UPDATE {$db_prefix}members
-SET birthdate = '0001-01-01'
-WHERE birthdate = '0000-00-00';
-
-UPDATE {$db_prefix}members
-SET birthdate = CONCAT('0004-', MONTH(birthdate), '-', DAYOFMONTH(birthdate))
-WHERE YEAR(birthdate) = 0;
 ---#
 
 /******************************************************************************/
--- a/other/upgrade_1-1.sql
+++ b/other/upgrade_1-1.sql
@@ -609,49 +609,7 @@ if ($do_it)
 
 ---# Updating event start and end dates...
 ALTER TABLE {$db_prefix}calendar
-DROP INDEX eventDate;
-
-ALTER TABLE {$db_prefix}calendar
-CHANGE COLUMN eventDate startDate date NOT NULL default '0001-01-01';
-
-ALTER TABLE {$db_prefix}calendar
-CHANGE COLUMN startDate startDate date NOT NULL default '0001-01-01';
-
-UPDATE {$db_prefix}calendar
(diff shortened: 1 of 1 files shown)


sbulen:
I don't think this will work - the 1.1 upgrade actually renamed some 1.0 columns and added others.  I think those parts need to be kept.  The 2.1 logic does not do the renames & new columns.  (It goes from 'eventDate' to 'startDate' and 'endDate'.)

I think the only parts that need to be removed are the references to 0000-00-00.  

Question:  If there were invalid dates, I believe those would have been caught during the mysql upgrade itself, correct?   I.e., it wouldn't support allowing those values in the columns at all?  I.e., I think they would have been required to be corrected before even running this upgrader...  If that is true, I believe it is safe to remove those checks for 0000-00-00.

albertlast:
About the renaming i had to think about...

Invalid value get notice by mysql when you do a dml command.
So you can select a invalid date but you can't insert a invalid or update a field to unvalid one.
In this case here #5741 the issue is that the column get renamed by this invalid dates get "inserted".

So the order is wrong in the upgrade script -> first change the values and then rename the col.

albertlast:
@sbulen is that so better?

sbulen:
No, same error.  I believe you really need to get rid of all references to 0000-00-00...  

(I'm running MySQL 5.7.17)

image

jdarwood007:
The upgrade scripts should not be touched unless they are currently broken to where they don't operate.  The intention of those scripts is to upgrade you to that SMF versions compliant database.  Upgrade 1-1 is meant to get you to a SMF 1.1 compliant database.

albertlast:
nothing new to it,
since the world is changing the old env not exists any more,
so you had to upgrade the of the 1-1 upgrade to be able to run in our time.

sbulen:
Of course, you're both right.  The question is what to do about it.  

I'm running MySQL 5.7, and I cannot upgrade a vanilla 1.1.21 DB due to these syntax issues.  It really seems to have a hard time with the 0000-00-00 dates.  I believe others trying to upgrade from 1.1 may experience the same, depending on their version of MySQL.  

@albertlast - I think we should leave the file as untouched as possible.   I have confirmed that removing the 4 commands that reference 0000-00-00, and modifying the one that uses it in an AND, addresses this issue.  I can run a full upgrade from 1.1 to 2.1 with that change.  

There may be other approaches - e.g., by experimenting with sql_mode to introduce a more lax mode when processing the older files.   I haven't had luck here so far, though.   

But just removing the references to 0000-00-00 works.

sbulen:
...and upon further reflection, just removing the references to 0000-00-00 just won't work.   Specifically, the various data fixes dont get applied.   

Note there are some other issues with this file - I updated the issue accordingly.  

@albertlast - I hope you don't mind I'm working on a PR to address both the 1.1 issues I've found thus far...  I have some thoughts that might address Sleepy's concerns as well.

albertlast:
Since 1.x didn't support postgres,
im not full care about.

PR #5755, opened 2019-07-27
Upgrader - 1.x issues - alt approach

### LINKED ISSUE #5741: Upgrader - challenges with 1.1 to 2.1 upgrade

### LINKED ISSUE #5750: Upgrader - challenges with 1.0 to 2.1 upgrade

### LINKED ISSUE #5751: Upgrader - challenges with 1.0 to 2.1 security
#### Description
If you upgrade from 1.0 to 2.1, your password is somehow mucked up...  You're locked out.  I've tried this a couple times, it always happens

### Steps to reproduce
1. Install 1.0.23, get it up & running fine
2. Upgrade to 2.1

- Database Type: MySQL
- Database Version: 5.7
- PHP Version: 5.6

image

### LINKED ISSUE #5797: 1.1 upgrade issue
#### Description
We have seen multiple of these today, so it's not unique to one individual.  

Let's watch for now & see if my proposed solution works:
https://www.simplemachines.org/community/index.php?topic=569571.0
https://www.simplemachines.org/community/index.php?topic=569567.0

Good news is that folks are upgrading from 1.1!!!

- Version/Git revision:  (n/a - same logic is used in 2.0 & 2.1 - fix would need to be applied to both) 
- Database Type:  mysql
- Database Version: 5.6

image

sbulen:
Fixes #5741 
Fixes #5750 
Fixes #5751 
Fixes #5797 

This supersedes #5752 .  

I've tested the following variations of upgrades:
- yabbse 1.5.5 => 2.1
- 1.0.23 (vanilla) => 2.1
- 1.1.21 (vanilla) => 2.1
- 2.0.15 (vanilla) => 2.1
- 2.0.15 (test copy of my forum) => 2.1

In each of the following environments, including all updates thru 9/9/19:
- WAMP, PHP 7.2, MySQL 5.7
- Linux, PHP 5.4, MySQL 5.7

If others could test, that'd be welcome! Especially if you have a 1.0 or 1.1 environment with real data in it...

EDIT:  Notes on yabbse upgrades:
 - Tested with 1.5.5, I know 1.5.0 has  a completely different db structure & won't work

### DIFF
--- a/other/upgrade_2-0_mysql.sql
+++ b/other/upgrade_2-0_mysql.sql
@@ -2763,7 +2763,7 @@ if (!isset($modSettings['attachment_thumb_png']))
 // Do they have "babylon" installed?
 if (file_exists($GLOBALS['boarddir'] . '/Themes/babylon'))
 {
-	$babylon_dir = $GLOBALS['boarddir'] . '/Themes/babylon';
+	$babylon_dir = smf_mysql_real_escape_string($GLOBALS['boarddir'] . '/Themes/babylon');
 	$theme_request = upgrade_query("
 		SELECT ID_THEME
 		FROM {$db_prefix}themes
@@ -2778,13 +2778,13 @@ if (file_exists($GLOBALS['boarddir'] . '/Themes/babylon'))
 		smf_mysql_free_result($theme_request);
 		unset($row);
 
-		$known_themes = explode(', ', $modSettings['knownThemes']);
+		$known_themes = explode(',', $modSettings['knownThemes']);
 
 		// Remove this value...
 		$known_themes = array_diff($known_themes, array($id_theme));
 
 		// Change back to a string...
-		$known_themes = implode(', ', $known_themes);
+		$known_themes = implode(',', $known_themes);
 
 		// Update the database
 		upgrade_query("
--- a/other/upgrade_2-0_mysql.sql
+++ b/other/upgrade_2-0_mysql.sql
@@ -2763,19 +2763,24 @@ if (!isset($modSettings['attachment_thumb_png']))
 // Do they have "babylon" installed?
 if (file_exists($GLOBALS['boarddir'] . '/Themes/babylon'))
 {
-	$babylon_dir = smf_mysql_real_escape_string($GLOBALS['boarddir'] . '/Themes/babylon');
-	$theme_request = upgrade_query("
-		SELECT ID_THEME
-		FROM {$db_prefix}themes
-		WHERE variable = 'theme_dir'
-			AND value ='$babylon_dir'");
+	$babylon_dir = $GLOBALS['boarddir'] . '/Themes/babylon';
+	$theme_request = $smcFunc['db_query']('', '
+		SELECT id_theme
+		FROM {db_prefix}themes
+		WHERE variable = {string:themedir}
+			AND value = {string:babylondir}',
+		array(
+			'themedir' => 'theme_dir',
+			'babylondir' => $babylon_dir,
+		)
+	);
 
 	// Don't do anything if this theme is already uninstalled
-	if (smf_mysql_num_rows($theme_request) == 1)
+	if ($smcFunc['db_num_rows']($theme_request) == 1)
 	{
-		$row = smf_mysql_fetch_row($theme_request);
+		$row = $smcFunc['db_fetch_row']($theme_request);
 		$id_theme = $row[0];
-		smf_mysql_free_result($theme_request);
+		$smcFunc['db_free_result']($theme_request);
 		unset($row);
 
 		$known_themes = explode(',', $modSettings['knownThemes']);
@@ -2787,9 +2792,12 @@ if (file_exists($GLOBALS['boarddir'] . '/Themes/babylon'))
 		$known_themes = implode(',', $known_themes);
 
 		// Update the database
-		upgrade_query("
-			REPLACE INTO {$db_prefix}settings (variable, value)
-			VALUES ('knownThemes', '$known_themes')");
+		$smcFunc['db_insert']('replace',
+			'{db_prefix}settings',
+			array('variable' => 'string', 'value' => 'string'),
+			array('knownThemes', $known_themes),
+			array('variable')
+		);
 
 		// Delete any info about this theme
 		upgrade_query("
@@ -2809,10 +2817,12 @@ if (file_exists($GLOBALS['boarddir'] . '/Themes/babylon'))
 
 		if ($modSettings['theme_guests'] == $id_theme)
 		{
-			upgrade_query("
-				REPLACE INTO {$db_prefix}settings
-				(variable, value)
-				VALUES('theme_guests', 0)");
+			$smcFunc['db_insert']('replace',
+				'{db_prefix}settings',
+				array('variable' => 'string', 'value' => 'string'),
+				array('theme_guests', 0),
+				array('variable')
+			);
 		}
 	}
 }
--- a/other/upgrade_2-0_mysql.sql
+++ b/other/upgrade_2-0_mysql.sql
@@ -1735,7 +1735,7 @@ if ($profileCount == 0)
 	// Update the board tables.
 	foreach ($board_updates as $profile => $boards)
 	{
-		if (empty($boards))
+		if (empty($boards) || empty($profile))
 			continue;
 
 		$boards = implode(',', $boards);
--- a/other/upgrade_2-0_postgresql.sql
+++ b/other/upgrade_2-0_postgresql.sql
@@ -926,19 +926,24 @@ if (!isset($modSettings['attachment_thumb_png']))
 // Do they have "babylon" installed?
 if (file_exists($GLOBALS['boarddir'] . '/Themes/babylon'))
 {
-	$babylon_dir = smf_mysql_real_escape_string($GLOBALS['boarddir'] . '/Themes/babylon');
-	$theme_request = upgrade_query("
-		SELECT ID_THEME
-		FROM {$db_prefix}themes
-		WHERE variable = 'theme_dir'
-			AND value ='$babylon_dir'");
+	$babylon_dir = $GLOBALS['boarddir'] . '/Themes/babylon';
+	$theme_request = $smcFunc['db_query']('', '
+		SELECT id_theme
+		FROM {db_prefix}themes
+		WHERE variable = {string:themedir}
+			AND value = {string:babylondir}',
+		array(
+			'themedir' => 'theme_dir',
+			'babylondir' => $babylon_dir,
+		)
+	);
 
 	// Don't do anything if this theme is already uninstalled
-	if (smf_mysql_num_rows($theme_request) == 1)
+	if ($smcFunc['db_num_rows']($theme_request) == 1)
 	{
-		$row = smf_mysql_fetch_row($theme_request);
+		$row = $smcFunc['db_fetch_row']($theme_request);
 		$id_theme = $row[0];
-		smf_mysql_free_result($theme_request);
+		$smcFunc['db_free_result']($theme_request);
 		unset($row);
 
 		$known_themes = explode(',', $modSettings['knownThemes']);
@@ -950,9 +955,12 @@ if (file_exists($GLOBALS['boarddir'] . '/Themes/babylon'))
 		$known_themes = implode(',', $known_themes);
 
 		// Update the database
-		upgrade_query("
-			REPLACE INTO {$db_prefix}settings (variable, value)
-			VALUES ('knownThemes', '$known_themes')");
+		$smcFunc['db_insert']('replace',
+			'{db_prefix}settings',
+			array('variable' => 'string', 'value' => 'string'),
+			array('knownThemes', $known_themes),
+			array('variable')
+		);
 
 		// Delete any info about this theme
 		upgrade_query("
@@ -972,10 +980,12 @@ if (file_exists($GLOBALS['boarddir'] . '/Themes/babylon'))
 
 		if ($modSettings['theme_guests'] == $id_theme)
 		{
-			upgrade_query("
-				REPLACE INTO {$db_prefix}settings
-				(variable, value)
-				VALUES('theme_guests', 0)");
+			$smcFunc['db_insert']('replace',
+				'{db_prefix}settings',
+				array('variable' => 'string', 'value' => 'string'),
+				array('theme_guests', 0),
+				array('variable')
+			);
 		}
 	}
 }
--- a/other/upgrade_2-0_postgresql.sql
+++ b/other/upgrade_2-0_postgresql.sql
@@ -918,192 +918,67 @@ if (!isset($modSettings['attachment_thumb_png']))
 ---#
 
 /******************************************************************************/
---- Installing new default theme...
+--- Cleaning up after old themes...
 /******************************************************************************/
 
----# Installing theme settings...
+---# Checking for "babylon" and removing it if necessary...
 ---{
-// This is Grudge's secret "I'm not a developer" theme install code - keep this quiet ;)
-
-// Firstly, I'm going out of my way to not do this twice!
-if ((!isset($modSettings['smfVersion']) || $modSettings['smfVersion'] <= '2.0 RC2') && empty($modSettings['dont_repeat_theme_core']))
+// Do they have "babylon" installed?
+if (file_exists($GLOBALS['boarddir'] . '/Themes/babylon'))
 {
-	// Check it's not already here, just in case.
+	$babylon_dir = smf_mysql_real_escape_string($GLOBALS['boarddir'] . '/Themes/babylon');
 	$theme_request = upgrade_query("
-		SELECT id_theme
+		SELECT ID_THEME
 		FROM {$db_prefix}themes
 		WHERE variable = 'theme_dir'
-			AND value LIKE '%core'");
-	// Only do the upgrade if it doesn't find the theme already.
-	if ($smcFunc['db_num_rows']($theme_request) == 0)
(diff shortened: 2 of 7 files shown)


other/upgrade_2-0_mysql.sql
```diff
@@ -2778,13 +2778,13 @@ if (file_exists($GLOBALS['boarddir'] . '/Themes/babylon'))
...
 
-		$known_themes = explode(', ', $modSettings['knownThemes']);
+		$known_themes = explode(',', $modSettings['knownThemes']);
 
 		// Remove this value...
 		$known_themes = array_diff($known_themes, array($id_theme));
 
 		// Change back to a string...
-		$known_themes = implode(', ', $known_themes);
+		$known_themes = implode(',', $known_themes);
```
albertlast, line 2792: I hope you don't forget postgres,
when you work on files which are 2.x .sql

sbulen: The pg .sql files dont have any of this, for example, deletion of old 1.x themes when checking 2.0.  

I considered trying to build them from scratch, but frankly, I don't know how.  So I can't...

The good news is that the only impacted installs are 1.x pg installs - which I doubt there are any at all.

albertlast: And what is with this here?
https://github.com/SimpleMachines/SMF2.1/blob/e0848c82fbbc912b51789a2fefded6b6830f086f/other/upgrade_2-0_postgresql.sql#L1026-L1028

sbulen: Compare upgrade_2-0_postgresql.sql to upgrade_2-0_mysql.sql....

It's ~1300 lines vs ~3000.   The pg one is woefully incomplete & completely not acceptable for migration from 1.x.  It simply isn't designed to do so today.  As an example, look at the very first set of modifications covered in the mysql version - all the column renames, etc.  NONE of that is covered in the pg version, meaning, _a pg 1.x will not be converted properly to 2.x_.

It appears to me that the pg version - in broad strokes - omits the true "1.x conversion" items but keeps the "2.x validation" items found in the mysql version.  E.g., it will add a table that doesn't exist.  

Before diving in wholesale into pg 1.x to 2.x migration changes, I think we would need to determine 2 things:
 - Are we really trying to support conversion of a pg 1.x forum to a pg 2.x forum?  Does one even exist?
 - What types of changes should be included in this script?

tinoest:
To my knowledge SMF 1.1 never supported PGSQL or at least it didn’t when I looked to switch a forum in 2009 to SMF. I don’t think an upgrader from 1.x to 2.1 is required for SMF 2.1

sbulen:
The upgrade scripts are also used in the support forum to attempt to correct DB issues with forums.  

This is why, when running a 2.1 upgrade, it runs the 2.0 upgrade step as well as the 2.1 upgrade step.  

We may have pg 2.0 forums out there, and these scripts will be run against them.  

So...  To use a specific example...  albertlast's suggestion above makes sense, in that, we have removed from the mysql 2.0 file the "let's add this old theme" and in fact replaced it with "let's remove an older now-unsupported theme" logic.  

I agree with albertlast's specific example above.  I.e., an older 2.0 pg forum may still have those funky old theme settings in them, if it had once been 1.x in the past.  

I am just concerned with how far to take it...  Especially since I don't have a pg forum to test on & I am very uncomfortable making changes that I cannot test.

sbulen:
@albertlast - done.
If you could test a 2.0 to 2.1 pg upgrade, that would be appreciated.

other/upgrade_2-0_postgresql.sql
```diff
@@ -918,192 +918,67 @@ if (!isset($modSettings['attachment_thumb_png']))
...
-// Firstly, I'm going out of my way to not do this twice!
-if ((!isset($modSettings['smfVersion']) || $modSettings['smfVersion'] <= '2.0 RC2') && empty($modSettings['dont_repeat_theme_core']))
+// Do they have "babylon" installed?
+if (file_exists($GLOBALS['boarddir'] . '/Themes/babylon'))
 {
-	// Check it's not already here, just in case.
+	$babylon_dir = smf_mysql_real_escape_string($GLOBALS['boarddir'] . '/Themes/babylon');
 	$theme_request = upgrade_query("
-		SELECT id_theme
+		SELECT ID_THEME
```
albertlast: The sql keywords are commonly in uppercase and
the other stuff in lower case -> would keep the lowercase version

other/upgrade_2-0_postgresql.sql
```diff
@@ -918,192 +918,67 @@ if (!isset($modSettings['attachment_thumb_png']))
...
 		FROM {$db_prefix}themes
 		WHERE variable = 'theme_dir'
-			AND value LIKE '%core'");
-	// Only do the upgrade if it doesn't find the theme already.
-	if ($smcFunc['db_num_rows']($theme_request) == 0)
+			AND value ='$babylon_dir'");
+
+	// Don't do anything if this theme is already uninstalled
+	if (smf_mysql_num_rows($theme_request) == 1)
```
albertlast: you are here in the postgre file ``smf_mysql_num_rows`` didn't work here.
You would make your life easier when you use the $smcFunc['db_*'] since they exists in both worlds.

other/upgrade_2-0_postgresql.sql
```diff
@@ -918,192 +918,67 @@ if (!isset($modSettings['attachment_thumb_png']))
...
-					array(0, $id_core_theme, 'name', 'Core Theme'),
-					array(0, $id_core_theme, 'theme_url', $core['theme_url']),
-					array(0, $id_core_theme, 'images_url', $core['images_url']),
-					array(0, $id_core_theme, 'theme_dir', $core['theme_dir'])
-				),
-				array()
-		);
+		$row = smf_mysql_fetch_row($theme_request);
+		$id_theme = $row[0];
+		smf_mysql_free_result($theme_request);
```
albertlast: you are here in the postgre file smf_mysql_num_rows didn't work here.
You would make your life easier when you use the $smcFunc['db_*'] since they exists in both worlds.

other/upgrade_2-0_postgresql.sql
```diff
@@ -918,192 +918,67 @@ if (!isset($modSettings['attachment_thumb_png']))
...
-				array('id_member' => 'int', 'id_theme' => 'int', 'variable' => 'string-255', 'value' => 'string-255'),
-				array(
-					array(0, $id_core_theme, 'name', 'Core Theme'),
-					array(0, $id_core_theme, 'theme_url', $core['theme_url']),
-					array(0, $id_core_theme, 'images_url', $core['images_url']),
-					array(0, $id_core_theme, 'theme_dir', $core['theme_dir'])
-				),
-				array()
-		);
+		$row = smf_mysql_fetch_row($theme_request);
```
albertlast: you are here in the postgre file smf_mysql_num_rows didn't work here.
You would make your life easier when you use the $smcFunc['db_*'] since they exists in both worlds.

other/upgrade_2-0_postgresql.sql
```diff
@@ -918,192 +918,67 @@ if (!isset($modSettings['attachment_thumb_png']))
...
 
-		// Since we want to do a replace, just delete the old settings and re-insert them
+		// Change back to a string...
+		$known_themes = implode(',', $known_themes);
+
+		// Update the database
 		upgrade_query("
-			DELETE FROM {$db_prefix}settings
-			WHERE variable IN ('theme_default', 'theme_guests', 'knownThemes')");
+			REPLACE INTO {$db_prefix}settings (variable, value)
```
albertlast: Use the replace methode of $smcfunc['db_insert'],
replace got a different syntax in pg.

other/upgrade_2-0_postgresql.sql
```diff
@@ -918,192 +918,67 @@ if (!isset($modSettings['attachment_thumb_png']))
...
+			WHERE id_theme = $id_theme");
 
-		// What about members?
+		// Set any members or boards using this theme to the default
 		upgrade_query("
 			UPDATE {$db_prefix}members
-			SET id_theme = $id_core_theme
-			WHERE id_theme = 1");
+			SET id_theme = 0
+			WHERE id_theme = $id_theme");
```
albertlast, line 974: Don't use double quotes for sql queries.

other/upgrade_2-0_postgresql.sql
```diff
@@ -918,192 +918,67 @@ if (!isset($modSettings['attachment_thumb_png']))
...
 
-		// Boards?
 		upgrade_query("
 			UPDATE {$db_prefix}boards
-			SET id_theme = $id_core_theme
-			WHERE id_theme = 1");
+			SET id_theme = 0
+			WHERE id_theme = $id_theme");
```
albertlast, line 979: Don't use double quotes for sql queries.

other/upgrade_2-0_postgresql.sql
```diff
@@ -918,192 +918,67 @@ if (!isset($modSettings['attachment_thumb_png']))
...
-
-			if (!empty($insertRows))
-				upgrade_query("
-					INSERT INTO {$db_prefix}themes
-						(id_member, id_theme, variable, value)
-					VALUES
-						" . implode(',
-						', $insertRows). ' ON CONFLICT DO NOTHING');
+			upgrade_query("
+				REPLACE INTO {$db_prefix}settings
```
albertlast: Use the replace methode of $smcfunc['db_insert'],
replace got a different syntax in pg.

albertlast (changes requested):
Based of the amount of errors in the pg file i don't need to try to upgrade from 2.0 to 2.1

other/upgrade_2-0_postgresql.sql
sbulen: albertlast - I am keeping things consistent with these other .sql files, and have copied this over from what has been in production use for quite some time.  I don't want to start making cosmetic changes at this point.

other/upgrade_2-0_postgresql.sql
sbulen: albertlast - I am keeping things consistent with these other .sql files, and have copied this over from what has been in production use for quite some time.  I don't want to start making cosmetic changes at this point.

other/upgrade_2-0_postgresql.sql
sbulen: albertlast - I am keeping things consistent with these other .sql files, and have copied this over from what has been in production use for quite some time.  I don't want to start making cosmetic changes at this point.

other/upgrade_2-0_postgresql.sql
sbulen: Hmmm...  All of the upgrade-helper.php "smf_" functions are mysql specific...  Let me think about this...

other/upgrade_2-0_postgresql.sql
sbulen: Hmmm...  All of the upgrade-helper.php "smf_" functions are mysql specific...  Let me think about this...

other/upgrade_2-0_postgresql.sql
sbulen: Hmmm...  All of the upgrade-helper.php "smf_" functions are mysql specific...  Let me think about this...

other/upgrade_2-0_postgresql.sql
sbulen: Hmmm...  All of the upgrade-helper.php "smf_" functions are mysql specific...  Let me think about this...

other/upgrade_2-0_postgresql.sql
sbulen: @albertlast - I must admit I'm struggling with how far to take this.  

In the past, the steps were rewritten for pg vs mysql - hence there are even different files for upgrade_2-0_postgresql.sql vs upgrade_2-0_mysql.sql.   And they are very different...  

Question:  If we use smcFuncs throughout... can we get to the point where we don't even need different files, e.g., upgrade_2-0_postgresql.sql vs upgrade_2-0_mysql.sql?  Or, are there functions that cannot be abstracted using smcFunc needed in the upgrade?  

Do we "smcFunc" EVERYTHING?   Do we "smcFunc" only a little?  Or do we continue to rewrite the steps for pg?

albertlast: In general i don't understand why you add smf*mysql* function.

No the upgrade process is complex enough.

I would only not add any smf*mysql* function

sbulen: Those functions are heavily used throughout the 2.0 & 2.1 upgrade scripts already...  

Looking at the original 1.x scripts, they were mostly straight-up sql & very simple php.  A custom upgrade_query() that allowed for special treatment of certain sql errors was sufficient.  The 2.x scripts, however, require much more logic to migrate data & implement data fixes.  So there is a need to use functions beyond upgrade_query().  

While doing so, it appears the intent was to _avoid_ smcFuncs.  I believe the reason for this was initial concern was that the upgrader is 'pulling itself up by its own bootstraps', meaning, that SMF may not be fully configured to run properly under the target version yet.  So the workaround was to either use direct mysqli calls (in 2.0.x) or to use the smf_ calls in the upgrade_helper (in 2.1) - which, in turn, used direct mysqli calls...  

However we got here, the problem is that we cannot re-use logic across pg & mysql in the upgrader today.  

I'll experiment a bit to see how big a deal it is to use smcFuncs.

live627: upgrade_query() already uses $smcFunc.

https://github.com/SimpleMachines/SMF2.1/blob/4b6c2b145369a171055e3d2d4f90408e445a0f41/other/upgrade.php#L2120

sbulen: Yep, for db_query.  

The question is whether we should modify all of the upgrade*.sql files to using it...  Consistently...

live627: eh...

It woulds need much testing, which translates to more delay/dev time. Maybe a bigger hammer should be put away.

albertlast: the problem is not the existing mysql function,
the problem is that you add some,
instead of using the newer api.

sbulen: @albertlast - What added function are you talking about?

frandominguezl: He means that you should use $smcFunc['db_num_rows'] or upgrade_query() instead of smf_mysql_num_rows.

sbulen: @live627 - My broader concern is that the 2.0 .sql files (e.g., in the 2.0.15 upgrade)  and the 2.1 .sql files (visible here on github) have drifted considerably.  This is not good, and I fear that forums may have different fixes applied (_or worse, not applied..._).  I believe they really should match - preferably byte-for-byte.  One of _many_ differences is that right now, 2.0 uses direct mysqli calls (with a couple smcFuncs), where 2.1 mainly uses these smf_ calls from upgrade_helper.php (which makes mysqli calls...).  

At some point - sooner rather than later - they should be brought back in sync in some agreed-upon fashion.  When doing so, I agree with albertlast that we should use the existing smcFuncs in both branches, and get rid of direct mysqli calls.

sbulen: @frandominguez03 - that's not new by any stretch.  They've been used throughout the 2.1 .sql scripts for years...

sbulen: All - What I'm going to do here is do targeted fixes for albertlast's comments.

We will look at the proper sync'ing of the files as a separate issue/PR.  It's important to get 1.x working again.

sbulen: Converted to smcFunc

other/upgrade_2-0_postgresql.sql
sbulen: Replace smcFunc used

other/upgrade_2-0_postgresql.sql
sbulen: smcFunc replace used

other/upgrade_2-0_postgresql.sql
sbulen: smcFunc used

other/upgrade_2-0_postgresql.sql
sbulen: smcFunc used

sbulen:
This is another one I'd target for RC3.  The sooner folks can upgrade cleanly, the better.

PR #5765, opened 2019-08-16
Fix some error around seek fixes #5764

### LINKED ISSUE #5764: Topic listing issues
#### Description
```
Error
Type of error: Critical
Error message:
Function: Display
File: .../Sources/Display.php
Line: 934
URL of page causing the error: http://.../index.php?topic=187.0
Backtrace information
#0: smf_db_error_backtrace()
Called from ...\Sources\Subs-Db-mysql.php on line 204
#2: preg_replace_callback()
#3: smf_db_query()
Called from ...\Sources\Display.php on line 934
#4: Display()
Called from ...\index.php on line 151
```

### Steps to reproduce
0. Turn [**Show most recent posts at the top in topic view**] ON
1. Make a topic with 2 pages
2. Mark the topic unread
3. View the 1st page
4. Go back to the board
5. Click the "New" button
6. Now click on a different page
7. Click on a different page again

```
```

- Version/Git revision: https://github.com/SimpleMachines/SMF2.1/commit/22482b33dd5704417a989e41311c397d45ee8936
- Database Type: MySQL
- Database Version: 5.7
- PHP Version: 7.3

Similar to this issue: #5486
Related PR: #5488

albertlast:
Atm i'm not fully done with testing.

Issue: #5764

### DIFF
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -906,11 +906,11 @@ function Display()
 			$DBascending = $ascending;
 			$page_operator = $ascending ? '>=' : '<=';
 		}
-        elseif ($start_char === 'L')
-        {
-            $DBascending = !$ascending;
-            $page_operator = $ascending ? '>=' : '<=';
-        }
+		elseif ($start_char === 'L')
+		{
+			$DBascending = !$ascending;
+			$page_operator = $ascending ? '>=' : '<=';
+		}
 		else
 		{
 			$DBascending = !$ascending;
@@ -1020,7 +1020,7 @@ function Display()
 		if (!$DBascending)
 			sort($messages, SORT_NUMERIC);
 		else
-            rsort($messages, SORT_NUMERIC);
+			rsort($messages, SORT_NUMERIC);
 	}
 
 	// Remember the paging data for next time
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -906,6 +906,11 @@ function Display()
 			$DBascending = $ascending;
 			$page_operator = $ascending ? '>=' : '<=';
 		}
+        elseif ($start_char === 'L')
+        {
+            $DBascending = !$ascending;
+            $page_operator = $ascending ? '>=' : '<=';
+        }
 		else
 		{
 			$DBascending = !$ascending;
@@ -1013,13 +1018,15 @@ function Display()
 
 		// Sort the messages into the correct display order
 		if (!$DBascending)
-			sort($messages);
+			sort($messages, SORT_NUMERIC);
+		else
+            rsort($messages, SORT_NUMERIC);
 	}
 
 	// Remember the paging data for next time
-	$_SESSION['page_first_id'] = $ascending ? reset($messages) : end($messages);
+	$_SESSION['page_first_id'] = $messages[0];
 	$_SESSION['page_before_start'] = $_REQUEST['start'] - $limit;
-	$_SESSION['page_last_id'] = $ascending ? end($messages) : reset($messages);
+	$_SESSION['page_last_id'] = $messages[count($messages) - 1];
 	$_SESSION['page_next_start'] = $_REQUEST['start'] + $limit;
 	$_SESSION['page_current_start'] = $_REQUEST['start'];
 	$_SESSION['page_topic'] = $topic;


SychO9:
Although this fixes the error mentioned in the issue, you can still see how messed up the order of the posts is, and I get these

```
Type of error: Undefined 
Undefined offset: -1
http://.../index.php?topic=187.6
.../Sources/Display.php (Line 1029)
```

```
Type of error: Undefined 
Undefined offset: 0
http://.../index.php?topic=187.6
.../Sources/Display.php (Line 1027)
```

frandominguezl:
I get the same, and in my case the suggested changes do not fix the error using one single page. If the option to show most recent posts at the top is enables and you reply the topic, the topic will remain bugged.

Sources/Display.php
```diff
@@ -1013,13 +1018,15 @@ function Display()
...
 		if (!$DBascending)
-			sort($messages);
+			sort($messages, SORT_NUMERIC);
+		else
+			rsort($messages, SORT_NUMERIC);
 	}
 
 	// Remember the paging data for next time
-	$_SESSION['page_first_id'] = $ascending ? reset($messages) : end($messages);
+	$_SESSION['page_first_id'] = $messages[0];
```
Sesquipedalian, line 1027: Instead of `$messages[0]`, you should use `reset($messages)` here. If `$messages` is an empty array, `$messages[0]` will produce an undefined index error, whereas `reset($messages)` will simply return `false`.

Sources/Display.php
```diff
@@ -1013,13 +1018,15 @@ function Display()
...
 	$_SESSION['page_before_start'] = $_REQUEST['start'] - $limit;
-	$_SESSION['page_last_id'] = $ascending ? end($messages) : reset($messages);
+	$_SESSION['page_last_id'] = $messages[count($messages) - 1];
```
Sesquipedalian, line 1029: Instead of `$messages[count($messages) - 1]`, you should use `end($messages)` here. If `$messages` is an empty array, `$messages[count($messages) - 1]` will evaluate to `$messages[-1]` and produce an undefined index error, whereas `end($messages)` will simply return `false`.

Sources/Display.php
```diff
@@ -906,6 +906,11 @@ function Display()
 			$DBascending = $ascending;
 			$page_operator = $ascending ? '>=' : '<=';
 		}
+		elseif ($start_char === 'L')
+		{
+			$DBascending = !$ascending;
+			$page_operator = $ascending ? '>=' : '<=';
+		}
```
Sesquipedalian, line 913: Since the only possible values for `$start_char` are `M`, `C`, and `L`, the `else` statement after this line will never be reached. This means that you can replace this entire `if ... elseif ... else` block with two lines:
```
$DBascending = ($start_char === 'M' || $start_char === 'C') ? $ascending : !$ascending;
$page_operator = $ascending ? '>=' : '<=';
```

Sesquipedalian (changes requested):
Note: I haven't tested this PR yet. These comments only deal with problems I saw when reading the code.

albertlast:
From my pov exists two ways,
support seeking only for ascending view,
good reason would be because the most board use this direction or
drop complety the seeking logic.

Sesquipedalian:
> From my pov exists two ways,
> support seeking only for ascending view,
> good reason would be because the most board use this direction or
> drop complety the seeking logic.

No, we just need to be a bit smarter about it.

The main thing about the seek method is that it only works correctly if the number of visible posts in the topic hasn't changed since the last page load. If that number has changed, then seeking will inevitably make mistakes. This is most obvious when the posts are sorted in descending order, but it is actually true regardless of the sort order. There are merely fewer situations where a human might notice the mistakes when sorting in ascending order, that's all.

Fortunately, `$context['total_visible_posts']` exists. All we need to do is remember that value in `$_SESSION['page_total_visible_posts']` and then check on each page load whether `$context['total_visible_posts']` is equal to `$_SESSION['page_total_visible_posts']`. If they are equal, then we can safely use the seek method. Otherwise, we need to fall back to the traditional method.

I think that adding this test would allow us to avoid all the weird issues we've been seeing with the seek method and also to simplify some of the code used inside the seek method block. Because the seek method code would already know that the total number of visible posts was unchanged, the various workarounds that it currently uses to try to recover from unexpected results would no longer be necessary.

albertlast:
Someting like lastchanged could be not better instead of post count right?

PR #5767, opened 2019-08-17
Fix registration no access error

### LINKED ISSUE #5582: Can't register a new account if integrated captcha is active
#### Description
If the default integrated (standard, not ReCAPTCHA) is active and you try to register a new account (a new user), it throws a "You are not allowed to access this section" error. Disabling the integrated captcha solves the problem.

https://www.simplemachines.org/community/index.php?topic=547871.0

### Steps to reproduce
1. Enable integrated captcha on registration.
2. Try to register a new account.

- Version/Git revision: RC2
- Database Type: MySQL (MyISAM)
- Database Version: 5.6.43
- PHP Version: 7.2.16

SychO9:
Fixes #5582

So the issue was that, when a 2.0 installation has Anti-Spam questions and a "*Number of verification questions user must answer*" during the upgrade when those questions get moved to the **qanda** table the lngfile column ends up empty for all questions because **$language** is undefined at that point (not added to globals)

Which means that these questions don't show up on the registration page or in the admin panel page, so when the user registers without being able to answer, error no_access happens.

### DIFF
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -1537,7 +1537,7 @@ $get_questions = upgrade_query("
 	WHERE comment_type = 'ver_test'");
 
 	while ($row = $smcFunc['db_fetch_assoc']($get_questions))
-		$questions[] = array($language, $row['question'], serialize(array($row['answer'])));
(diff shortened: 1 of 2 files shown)


sbulen:
Good sleuthing!

jdarwood007 (approved):

PR #5772, opened 2019-08-20
Updates jQuery in install.php

### LINKED ISSUE #5760: jQuery is outdated again
#### Description
jQuery has made several updates since the last public release candidate of SMF 2.1. Google Chrome is also reporting a potential vulnerability in the current version used by SMF 2.1 RC2 during page audits.

### Steps to reproduce
1. Not applicable.

- Version/Git revision: SMF 2.1 RC2
- Database Type: MySQL
- Database Version: 8.0.12 - MySQL Community Server - GPL
- PHP Version: 7.2.8

SMF 2.1 RC2 appears to be using jQuery 3.2.1, however jQuery 3.4.1 is the latest as of this bug report posting. (Bug report numbers 3712 and 4371 also reference previous updates done to the jQuery system as part of SMF 2.1's development).

frandominguezl:
Closes #5760

### DIFF
--- a/SSI.php
+++ b/SSI.php
@@ -19,6 +19,7 @@ define('SMF', 'SSI');
 define('SMF_VERSION', '2.1 RC2');
 define('SMF_FULL_VERSION', 'SMF ' . SMF_VERSION);
 define('SMF_SOFTWARE_YEAR', '2019');
+define('JQUERY_VERSION', '3.4.1');
 
 // We're going to want a few globals... these are all set later.
 global $time_start, $maintenance, $msubject, $mmessage, $mbname, $language;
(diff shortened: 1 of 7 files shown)


MissAllSunday:
To avoid missing out places to update, can you add either a constant or a setting for the jQuery version? that way it will be easier to modify and also easier to update when releasing updating patches.

frandominguezl:
I'm kinda new when coming to define constants. Any suggested places to do so?

frandominguezl:
As suggested by SychO, I've declared it in Load.php, Install.php and SSI.php. Does that look good, Suki?

MissAllSunday:
Yep :) but there's still a few missed places where constants are declared:

- cron.php 
- other/upgrade.php 
- proxy.php

We might not need the jQuery version on all those files, however, we can add it to keep things uniform.

Issue #5757, opened 2019-08-11
Uncaught TypeError: Cannot read property 'mime_type' of undefined


kungfooman:
#### Description

![image](https://user-images.githubusercontent.com/5236548/62834537-56b9d100-bc4e-11e9-9d8d-39cca7d27149.png)

### Steps to reproduce
1. Change settings to allow mp4 attachments
2. Upload some mp4 file

frandominguezl:
Where are you exactly getting this error? I tried to replicate it but I couldn't.

Can you provide more details? Files size, for example...

smfbigguy:
I have added mp4's to my install and it works fine with no errors. of course my install might be a month old or so, so something could have happened between then and now.

frandominguezl:
Mine is not that old. @kungfooman You're using latest build from GitHub or RC2?

sbulen:
I would also attempt mp4s from other sources.  

Mp4s are just containers, holding many types of contents.   IIRC, no "magic #",  so we're at the mercy of whoever encoded it to put the mime type in there.

sbulen:
I cannot duplicate this.  Had a different issue, though, which I will log separately.

kungfooman:
I tested HEAD (just cloned) and RC2, both fail with same error (beautified):

![image](https://user-images.githubusercontent.com/5236548/62849380-bad8a580-bcdf-11e9-8622-9f82941b5ad3.png)

The uploaded file size is 46MB, smaller files work! So this is probably a file size issue. But I don't know why, in my `php.ini` I allow `upload_max_filesize = 100M` and in SMF settings I allow unlimited file size (=0).

@frandominguez03 I get the error after dragging a .mp4 file on the "attachment drop zone", it looks like this after the error: 

![image](https://user-images.githubusercontent.com/5236548/62849724-05a6ed00-bce1-11e9-99c2-16d574f84643.png)

frandominguezl:
Thanks for the details. I could kind of reproduce it with a 30MB file:
![image](https://user-images.githubusercontent.com/10365222/62878673-d49eda80-bd29-11e9-95d1-15f544381ede.png)

sbulen:
Could you confirm what your post_max_size is in your php.ini file?  The default for those values, IIRC, is 30MB, and that's ~what we're seeing.  

I can reproduce this if I drop my post_max_size down to ~30MB.  

So...  I wonder if we should prevent folks from setting attachment settings larger than php.ini allows?

[...]

PR #5782, opened 2019-08-26
Embedded attachments

Sesquipedalian:
- Improves embedding of attachments via the attach BBC
    - Allows video and audio files to be embedded
    - Adds a display parameter to the attach BBC to allow the user to control whether to embed an attached file or just display a link.
    - Includes mime type info when loading an attachment's context data
- Adds `$context['browser']['is_mobile']` so that `isBrowser('mobile')` will work as an easy way to check if the client is a mobile device.
- Uses `content-disposition: inline` HTTP header when server audio & video attachments to mobile devices. This prompts mobile browsers to play the media file instead of trying to save it somewhere, which is usually the desired behaviour on mobile devices.


MissAllSunday:
@Sesquipedalian does this fixes or has something to do with #5757 ?

Sesquipedalian:
Not at the moment, but I suppose I could toss in a fix for that. [This mod](https://custom.simplemachines.org/mods/index.php?mod=4215) already contains most of the logic I'll need for that anyway.

Issue #5780, opened 2019-08-23
Tweaking some hooks?


DiegoAndresCortes:
#### Description

1.  integrate_post2_start (Post.php)
     We should include _&$post_errors_ as we don't have any other hook where we could append information to $post_errors **before** these errors are read.

2.  integrate_message_index (MessageIndex.php)
    We don't have a way to add more _where_ statements to $sort_table so if we include new data, pagination won't be loading the correct information and can't behave as expected compared with the old query of 2.0.x

3. loadBoard (Load.php)
   Seems like it's missing a hook? However, with a simple query you can define new $board_info elements depending on where you need this information so I'm not sure about this one.

albertlast:
Two the second point it's okay that is not changeable by hook.
When you change here,
You code must be able to run on big size forum,
And need to support mysql 5.0 till $today and postgres.
When you skilled enough use code replace.

DiegoAndresCortes:
4. integrate_remove_message (RemoveTopic.php)
   This hook is only invoked when the recycle board is disabled, is this correct? If a post goes to recycle bin then there's no hook to handle information

MissAllSunday:
I agree with the change on the second point. Adding the ability to modify the where statement doesn't mean SMF will have problems running on large boards. The ability is there for mod authors to use, we can't guarantee how that option will be used.

DiegoAndresCortes:
Just for keeping track #5870

PR #5784, opened 2019-08-27
[WIP] Addressing hooks information

DiegoAndresCortes:
Fixes some issues for #5780 

But needs more discussing still maybe.
And I have no clue how to approach loadBoard but doesn't seem like a big issue imo because usually $board_info is already available so we can get additional board info from different hooks.

For _integrate_remove_message_ I moved the hook outside of if statement because otherwise it's only invoked when there's no recycle bin. Also added _$row_ so we can get more info about this message in case it's gone forever. _$recycle_ suggested by Sycho as well.

For _integrate_remove_topics_ I would suggest adding a hook _before_ topics are deleted?

### DIFF
--- a/Sources/Likes.php
+++ b/Sources/Likes.php
@@ -413,7 +413,7 @@ class Likes
 		}
 
 		// Sometimes there might be other things that need updating after we do this like.
-		call_integration_hook('integrate_issue_like', array($this));
+		call_integration_hook('integrate_issue_like', array($this->_type, $this->_content, $this->_user['id'], $this->_alreadyLiked, $this->_validLikes));
 
 		// Now some clean up. This is provided here for any like handlers that want to do any cache flushing.
 		// This way a like handler doesn't need to explicitly declare anything in integrate_issue_like, but do so
--- a/Sources/Likes.php
+++ b/Sources/Likes.php
@@ -413,7 +413,7 @@ class Likes
 		}
 
 		// Sometimes there might be other things that need updating after we do this like.
-		call_integration_hook('integrate_issue_like', array($this->_type, $this->_content, $this->_user['id'], $this->_alreadyLiked, $this->_validLikes));
+		call_integration_hook('integrate_issue_like', array($this));
 
 		// Now some clean up. This is provided here for any like handlers that want to do any cache flushing.
 		// This way a like handler doesn't need to explicitly declare anything in integrate_issue_like, but do so
--- a/Sources/MessageIndex.php
+++ b/Sources/MessageIndex.php
@@ -313,8 +313,9 @@ function MessageIndex()
 	$message_index_selects = array();
 	$message_index_tables = array();
 	$message_index_wheres = array();
+	$message_index_topic_wheres = array();
 
-	call_integration_hook('integrate_message_index', array(&$message_index_selects, &$message_index_tables, &$message_index_parameters, &$message_index_wheres, &$topic_ids));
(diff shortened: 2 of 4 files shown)


Sources/MessageIndex.php
```diff
@@ -328,7 +329,8 @@ function MessageIndex()
 		' . (!empty($message_index_tables) ? implode("\n\t\t\t\t", $message_index_tables) : '') . '
 		WHERE t.id_board = {int:current_board} '
 			. (!$modSettings['postmod_active'] || $context['can_approve_posts'] ? '' : '
-			AND (t.approved = {int:is_approved}' . ($user_info['is_guest'] ? '' : ' OR t.id_member_started = {int:current_member}') . ')') . '
+			AND (t.approved = {int:is_approved}' . ($user_info['is_guest'] ? '' : ' OR t.id_member_started = {int:current_member}') . ')') . (!empty($message_index_topic_wheres) ? ' 
+			AND ' . implode("\n\t\t\t\tAND ", $message_index_topic_wheres) : ''). '
```
albertlast, line 333: Like I mentioned before I decline here any changes from author,
Since the part is fragile and important.

albertlast (changes requested):

Sources/MessageIndex.php
MissAllSunday: Yeah, it is fragile, however, we shouldn't limited what mod authors can do based on that logic.

We aim to offer freedom and options, we can't and shouldn't decide how those option will be used.

Sources/Likes.php
```diff
@@ -413,7 +413,7 @@ protected function like()
 		}
 
 		// Sometimes there might be other things that need updating after we do this like.
-		call_integration_hook('integrate_issue_like', array($this));
+		call_integration_hook('integrate_issue_like', array($this->_type, $this->_content, $this->_user['id'], $this->_alreadyLiked, $this->_validLikes));
```
MissAllSunday: Its better to pass the entire object and use setter and getters to get access to all properties.

Sources/MessageIndex.php
MissAllSunday, line 333: Obly apply the where statement when $message_index_topic_wheres is not empty, otherwise this will fail on an install with no functions for this hook.

MissAllSunday (changes requested):

Sources/MessageIndex.php
albertlast: Will @DiegoAndresCortes fix all performance and error issue which get raised up in the support board of sm.org?

MissAllSunday: The only issue that directly impacts a vanilla installation is directly applying the where var without making sure it has some value first. I already asked for that change.

It is not our duty to provide support for poorly performant mods. It is not our duty to limit what mods can do. Our duty is to allow mod authors the freedom to modify SMF.

DiegoAndresCortes: I thought it was already doing that?

MissAllSunday: You're right, didn't saw that.

MissAllSunday:
@DiegoAndresCortes are you planing on adding more changes or can we go ahead with this as it is?

MissAllSunday (approved):

DiegoAndresCortes:
Well I was interested in adding some for the mass actions, like deleting old topics, but not sure if it's actually useful atm so I'd leave it as it is right now.

MissAllSunday:
Let me know when we can merge this one then.

DiegoAndresCortes:
You can go ahead and merge I won't be adding those. If anyone wants to address anything else before Final release they can discuss it on #5780

PR #5796, opened 2019-09-09
Board category parse desc

### LINKED ISSUE #5734: getBoardindex do call parse_bbc which takes 94% of the run time

MissAllSunday:
Fixes #5734

Made a refactor on how to parse and save the parsed content on the cache.

Its a somehow atomic approach, the get and set operations are done on a per category and per board basis, this means an user who can only see a subset of the total boards won't get the load of parsing all of them.

Tested with filebased and SQLite3

### DIFF
--- a/Sources/Subs-BoardIndex.php
+++ b/Sources/Subs-BoardIndex.php
@@ -570,6 +570,9 @@ function getBoardIndex($board_index_options)
 		call_integration_hook('integrate_getboardtree', array($board_index_options, &$this_category));
 	}
 
+	// I took my time, I hurried up, the choice was mine I didn't think enough
+	setparsedDescriptions($to_parse);
+
 	return $board_index_options['include_categories'] ? $categories : $this_category;
 }
 
--- a/Sources/Subs-BoardIndex.php
+++ b/Sources/Subs-BoardIndex.php
@@ -640,11 +640,12 @@ function setParsedDescriptions($dataToParse = array())
 					'boards' => array(),
 				);
 
-				foreach ($category['boards'] as $board)
-					$already_parsed_data[$cat_id]['boards'] = array(
-						'name' => $board['name'],
-						'description' => $board['description'],
-					);
+				if (!empty($category['boards']))
+					foreach ($category['boards'] as $board)
+						$already_parsed_data[$cat_id]['boards'] = array(
+							'name' => $board['name'],
+							'description' => $board['description'],
+						);
 
 				cache_put_data('parsed_cat_description_'. $cat_id, $already_parsed_data[$cat_id], 864000);
 			}
--- a/Sources/Subs-BoardIndex.php
+++ b/Sources/Subs-BoardIndex.php
@@ -133,6 +133,7 @@ function getBoardIndex($board_index_options)
 	// Start with an empty array.
 	if ($board_index_options['include_categories'])
 		$categories = array();
+
 	else
 		$this_category = array();
 
--- a/Sources/Subs-BoardIndex.php
+++ b/Sources/Subs-BoardIndex.php
@@ -110,7 +110,7 @@ function getBoardIndex($board_index_options)
 	else
 		$result_boards = $smcFunc['db_query']('', '
 			SELECT' . ($board_index_options['include_categories'] ? '
-				b.id_cat, c.name AS cat_name, c.description AS cat_desc,' : 'b.id_cat,') . '
+				c.id_cat, c.name AS cat_name, c.description AS cat_desc,' : 'b.id_cat,') . '
 				' . (!empty($board_index_selects) ? implode(', ', $board_index_selects) : '') . ',
 				COALESCE(m.poster_time, 0) AS poster_time, COALESCE(mem.member_name, m.poster_name) AS poster_name,
 				m.subject, m.id_topic, COALESCE(mem.real_name, m.poster_name) AS real_name,
--- a/Sources/Subs-BoardIndex.php
+++ b/Sources/Subs-BoardIndex.php
@@ -250,6 +250,15 @@ function getBoardIndex($board_index_options)
 				{
 					$board_name = $row_board['board_name'];
 					$board_description = $row_board['description'];
+
+					// Always buckle up your kids!
+					if (!isset($to_parse[$row_board['id_cat']]['boards']))
+						$to_parse[$row_board['id_cat']]['boards'] = array();
+
+					$to_parse[$row_board['id_cat']]['boards'][$row_board['id_board']] = array(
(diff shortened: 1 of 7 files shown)


MissAllSunday:
Ugh, I forgot to replace the data with its parsed counterpart. Its late so I'll do it later.

albertlast:
> Some poor soul will get the hit of parsing the entire categories and boards names and descriptions but there isn't much we can do about it.

How would be to allow to disable this feature like "recent post function"?

And how long run you change on ~100 board forum in compare to the existing one?

MissAllSunday:
I don't understand what you mean by disable this feature. No it cannot be disable unless you don't use cache at all.

You won't be able to measure this against other alternatives. This approach adds to the cache as needed.  The query uses {query_see_board} which means the load of parsing the names and description will be shared among several users since all of them will have access to different boards.

Every time there's something new to parse, the cache gets refreshed and the new data gets appended to what we already have. The cache also has a long time duration since boards and categories aren't updated that frequently.

Lastly, having the cache by category means we will end up with fewer cache entries of a moderate size.

I also need to re-create the cache whenever a board is created/updated. Will add that too.

albertlast:
I don't see the bcode parsing feature in general as essential,
so i would purpose to allow to disable the parsing function to safe runtime/cpu time.

I do all the time measurements, first alternative is the version before you add you first pr,
the secodn one is the merge version right now.

for both cases: with cache and none cache.

MissAllSunday:
> I do all the time measurements, first alternative is the version before you add you first pr,
the secodn one is the merge version right now.

Cool. What I'm saying is that if you try to measure this using an admin account who can see 100 boards then you won't be able to see much improvements.

albertlast:
sm.org would already reach the point and
i'm waiting for measurments...

MissAllSunday:
> sm.org would already reach the point and

???

Sources/Subs-BoardIndex.php
```diff
@@ -95,7 +95,7 @@ function getBoardIndex($boardIndexOptions)
 	else
 		$result_boards = $smcFunc['db_query']('', '
 			SELECT' . ($boardIndexOptions['include_categories'] ? '
-				c.id_cat, c.name AS cat_name, c.description AS cat_desc,' : '') . '
+				c.id_cat, c.name AS cat_name, c.description AS cat_desc,' : 'b.id_cat,') . '
```
albertlast: i would also then change the c.id_cat to b.id_cat in the first position

albertlast (commented):
only look throw the code

albertlast:
While testing i get two kind of errors:
```
Error

    Type of error: Undefined_vars
    Error message: 8: Undefined index: description
    File: F:/github/SMF2.1/Sources/Subs-BoardIndex.php
    Line: 621
    URL of page causing the error: http://localhost/SMF2.1/index.php?

Backtrace information

    #0: smf_error_handler()
    Called from F:\github\SMF2.1\Sources\Subs-BoardIndex.php on line 621
    #1: setParsedDescriptions()
    Called from F:\github\SMF2.1\Sources\Subs-BoardIndex.php on line 548
    #2: getBoardIndex()
    Called from F:\github\SMF2.1\Sources\BoardIndex.php on line 49
    #3: BoardIndex()
    Called from F:\github\SMF2.1\index.php on line 155

```

and
```
Error

    Type of error: Undefined_vars
    Error message: 8: Undefined index: name
    File: F:/github/SMF2.1/Sources/Subs-BoardIndex.php
    Line: 620
    URL of page causing the error: http://localhost/SMF2.1/index.php?

Backtrace information

    #0: smf_error_handler()
    Called from F:\github\SMF2.1\Sources\Subs-BoardIndex.php on line 620
    #1: setParsedDescriptions()
    Called from F:\github\SMF2.1\Sources\Subs-BoardIndex.php on line 548
    #2: getBoardIndex()
    Called from F:\github\SMF2.1\Sources\BoardIndex.php on line 49
    #3: BoardIndex()
    Called from F:\github\SMF2.1\index.php on line 155
```

with this i can't test you changes,
only when i disable cache(used APCu) i can run your code fine,
ther i notice that your code run slower as the existing solution.

Sources/Subs-BoardIndex.php
```diff
@@ -470,13 +544,141 @@ function getBoardIndex($boardIndexOptions)
 		$context['latest_post'] = $latest_post['ref'];
 	}
 
+	// I took my time, I hurried up, the choice was mine I didn't think enough
+	$parsed_descriptions = setparsedDescriptions($to_parse);
```
live627: Just a tiny nitpick here... sorry... case doesn't match

MissAllSunday: case?

albertlast: i guess it should be setParsedDescriptions

MissAllSunday:
Made a refactor on how to parse and save the parsed content on the cache.

Its a somehow atomic approach, the get and set operations are done on a per category and per board basis, this means an user who can only see a subset of the total boards won't get the load of parsing all of them.

Tested with filebased and SQLite3.

Please give it a test.

albertlast:
When you switch from caching back to non caching,
stuff get lost:
grafik

Since this bug i can't fully do a performance test,
but partial:
You new version is 10ms faster as the existing code in a cached env.
Non cached env is missing of this bug mention of the start

MissAllSunday:
Added yet another refactor, grouping both categories and board caches, now there's a single cache entry for all categories and board cache is grouped by category, reducing the amount of hits to get the cached data.

MissAllSunday:
I removed support for parsing the board/cat name its just too much hassle to make sure the parsed result won't mess up all the different places where the board name is displayed.

Also due to the call to clean_cache when a board is been created/modified all cache data will be lost, I added support for keeping the cache data for all the other boards on the same category but the rest will get cleaned.

PR #5802, opened 2019-09-10
Uses "Forever" as the default time option when logging in

Sesquipedalian:
Fixes a minor annoyance that end users have repeatedly complained about over the years.


Sesquipedalian:
Honestly, a simple "Remember me on this device" checkbox is all we need, but this PR at least improves the situation.

albertlast (approved):

live627 (approved):

jdarwood007 (approved):

Issue #5806, opened 2019-09-16
Notification settings changes not saved - pg only


sbulen:
#### Description
Notification changes not getting saved.  For some users, we are seeing error messages, for some users, no error messages.

### Steps to reproduce
1. Change notification settings for a thread. 
2. Refresh page - you will see that the change was not saved.

### Environment (complete as necessary)
- Version/Git revision: Current as of today
- Database Type: pg
- Database Version: 10
- PHP Version: 5.6, 7.3

### Additional information/references
https://www.simplemachines.org/community/index.php?topic=569463.0

sbulen:
I think that part of the problem is that the code is out of sync with the DB structure, e.g., the pk is member, topic, board, but the full pk isn't referenced in some of the smcFunc insert calls:
https://github.com/SimpleMachines/SMF2.1/blob/dff71408bc055e9ad8eafc33e35b4618c800b096/Sources/Notify.php#L57
https://github.com/SimpleMachines/SMF2.1/blob/dff71408bc055e9ad8eafc33e35b4618c800b096/Sources/Notify.php#L157

MySQL must be more forgiving.

The next question is why I didn't get any errors logged...

albertlast:
If you want to stop this kind of leazy programming,
than you need to remove the default values of pk fields,
to get a fealing of the amount of work see here: https://github.com/albertlast/SMF2.1/commit/1e67630b08fe4ab450b023716e222ae00c570e97
I don't think it's a good idea to change the database schema in this way.

But yes would be intresting why no error message is catched in smf.

MissAllSunday:
Closed by #5807

PR #5807, opened 2019-09-17
[pg] Fix pk in topic/board notify

albertlast:
issue: #5806


sbulen:
Tests OK - notifications are now saved properly on pg!

PR #5816, opened 2019-10-02
Update description themers/modders?

DiegoAndresCortes:
It was a bit confusing for theme authors imo as I discussed it with @SychO9 
This should be more self explanatory (aside from reading the docs).

Any suggestions, or is this good enough

### DIFF
--- a/Themes/default/index.template.php
+++ b/Themes/default/index.template.php
@@ -89,17 +89,22 @@ function template_html_above()
 	/*
 		You don't need to manually load index.css, this will be set up for you.
 		Note that RTL will also be loaded for you.
-
-		The most efficient way of writing multi themes is to use a master
-		index.css plus variant.css files. If you've set them up properly
-		(through $settings['theme_variants']), the variant files will be loaded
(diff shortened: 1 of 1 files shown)


SychO9:
This is much better Diego !

I think it'd be really nice to also add an example since themers wont know how to exactly use the functions, something like this perhaps:
```php
loadCSSFile('filename.css', array('minimize' => true), 'smf_filename');
```

MissAllSunday (approved):

DiegoAndresCortes:
> 
> 
> This is much better Diego !
> 
> I think it'd be really nice to also add an example since themers wont know how to exactly use the functions, something like this perhaps:
> 
> ```
> loadCSSFile('filename.css', array('minimize' => true), 'smf_filename');
> ```

By default the functions use the filename for the ID so I won't add it. They can read more in the wiki once it's there maybe?

SychO9 (approved):

SychO9:
Yea, we need to get around writing the documentation for 2.1

PR #5829, opened 2019-10-14
Improves sanitization in ModifySubscription()

Sesquipedalian:


Sesquipedalian:
Hm. The check keeps showing as incomplete here, but following the details link shows that it passed. I'd restart the Travis build in order to force an update on the status here, but for some reason Travis is not giving me that option anymore. @CoreISP, has someone recently changed my GitHub or Travis access rights in a way that would prevent me from restarting Travis builds?

smfbigguy:
When I follow the details link it says this:

We couldn't find the repository
SimpleMachines/SMF2.1

PR #5840, opened 2019-10-23
Assigns quickbutton divs a class instead of an id (prevents conflicts)

Sesquipedalian:
Previously, this was creating HTML id conflicts when the template was called more than once for the same type of quickbuttons, such as on a post or personal message display page.

### DIFF
--- a/Themes/default/index.template.php
+++ b/Themes/default/index.template.php
@@ -651,7 +651,7 @@ function template_quickbuttons($list_items, $list_class = null, $output_method =
 
 	// Print the quickbuttons
 	$output = '
-		<ul class="quickbuttons' . (!empty($list_class) ? ' ' . $list_class : '') . '">';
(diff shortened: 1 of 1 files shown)


Themes/default/index.template.php
```diff
@@ -651,20 +651,20 @@ function template_quickbuttons($list_items, $list_id = null, $output_method = 'e
 
 	// Print the quickbuttons
 	$output = '
-		<ul' . (!empty($list_id) ? ' id="quickbuttons_'.$list_id.'"' : '') . ' class="quickbuttons">';
+		<ul class="quickbuttons' . (!empty($list_class) ? ' ' . $list_class : '') . '">';
```
SychO9: the custom class should probably be in the format of `quickbuttons_{class}` otherwise there is a chance there will be conflicts.

Sesquipedalian: Yeah, okay.

jdarwood007 (approved):

PR #5845, opened 2019-10-25
Marks alerts about a topic as read when the topic is viewed

Sesquipedalian:
Previously, we only marked the alert as read if you visited the topic via the alert. If you nevigated there via other means, the alert would still show as unread. This required the user to manually mark the alert as read after reading the topic itself, which quickly becomes unmanageable on an active forum.

### DIFF
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -1139,7 +1139,7 @@ function Display()
 		$smcFunc['db_query']('', '
 			UPDATE {db_prefix}user_alerts
 			SET is_read = {int:now}
-			WHERE id_member = {int:current_member}
+			WHERE is_read = 0 AND id_member = {int:current_member}
 				AND
 				(
 					(content_id IN ({array_int:messages}) AND content_type = {string:msg})
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -1136,28 +1136,31 @@ function Display()
 		}
 
 		// Mark any alerts about this topic or the posts on this page as read.
-		$smcFunc['db_query']('', '
-			UPDATE {db_prefix}user_alerts
-			SET is_read = {int:now}
-			WHERE is_read = 0 AND id_member = {int:current_member}
-				AND
-				(
-					(content_id IN ({array_int:messages}) AND content_type = {string:msg})
-					OR
-					(content_id = {int:current_topic} AND (content_type = {string:topic} OR (content_type = {string:board} AND content_action = {string:topic})))
-				)',
-			array(
-				'topic' => 'topic',
-				'board' => 'board',
-				'msg' => 'msg',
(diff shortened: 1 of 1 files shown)


Sources/Display.php
```diff
@@ -1134,6 +1134,30 @@ function Display()
 				array('id_member', 'id_board')
 			);
 		}
+
+		// Mark any alerts about this topic or the posts on this page as read.
+		$smcFunc['db_query']('', '
```
albertlast: got we a chance to know if this update is needed or not,
without doing a select?

Sesquipedalian: There's no need to do a SELECT. If the user is reading a topic, then the alerts about that topic should be marked as read.

The WHERE clause will mark alerts as read that are (1) about the topic as a whole (e.g. locked, stickied, etc.) and (2) about messages on the specific page that the user is viewing (e.g. likes, mentions, etc.). In both cases, if the page is being shown to the user, then by definition the content of those alerts has been read and the alerts should be marked as read.

This is the same logic used by social media sites, GitHub, etc., who mark alerts about content on the site as read when the user views the content, even if the user didn't look at the alert page itself to get to the content.

Make sense?

albertlast: github and co. used here an async query,
since we doesn't got something like that in php,
i asked if we know if the update query is needed to run.

Sesquipedalian: Oh, I see what you mean. 

No, I just run the query directly during page load because it is more efficient to use the information we already have than it would be to use a separate AJAX request. 

Using AJAX for this task would require the server to redo all the work to get this information again, and then we'd still need to run a very similar query to mark the alerts as read anyway. 

By running this query now, we skip all that extra work and make the most efficient use of the work we have already done.

Sesquipedalian: I suppose we could add an `if (!empty($user_info['alerts']))` before running the query, since there's no point in running it if the user doesn't have any unread alerts.

Sources/Display.php
```diff
@@ -1134,6 +1134,33 @@ function Display()
...
+					'topic' => 'topic',
+					'board' => 'board',
+					'msg' => 'msg',
+					'current_member' => $user_info['id'],
+					'current_topic' => $topic,
+					'messages' => $messages,
+					'now' => time(),
+				)
+			);
+			$user_info['alerts'] = $user_info['alerts'] - $smcFunc['db_affected_rows']();
```
jdarwood007: mysqli_affected_rows indicates it can return -1 in the case of an error.  Is this a concern?  It would essentially add 1 instead of subtracting one.

Sesquipedalian: I'll wrap it in a `max()`

PR #5851, opened 2019-10-31
Keep order of poll choices consistent

m4z:
This patch sorts poll choices the way they were input by the poll creator.

Before this, the poll choice order was inconsistent over time, see link below (in Postgres, the last-voted-on choices were shown last; I don't know or care about the details in MySQL, the patch is trivial enough that I don't expect issues on MySQL though).

Thanks to [Arantor, who did the searching that would have been the biggest part of my work for this patch](https://www.simplemachines.org/community/index.php?topic=569566.msg4036106#msg4036106).

### DIFF
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -676,7 +676,8 @@ function Display()
 			SELECT pc.id_choice, pc.label, pc.votes, COALESCE(lp.id_choice, -1) AS voted_this
 			FROM {db_prefix}poll_choices AS pc
 				LEFT JOIN {db_prefix}log_polls AS lp ON (lp.id_choice = pc.id_choice AND lp.id_poll = {int:id_poll} AND lp.id_member = {int:current_member} AND lp.id_member != {int:not_guest})
-			WHERE pc.id_poll = {int:id_poll} ORDER by pc.id_choice',
+			WHERE pc.id_poll = {int:id_poll}
+			ORDER BY pc.id_choice',
 			array(
 				'current_member' => $user_info['id'],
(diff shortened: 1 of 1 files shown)


MissAllSunday (approved):

Sources/Display.php
```diff
@@ -676,7 +676,7 @@ function Display()
 			SELECT pc.id_choice, pc.label, pc.votes, COALESCE(lp.id_choice, -1) AS voted_this
 			FROM {db_prefix}poll_choices AS pc
 				LEFT JOIN {db_prefix}log_polls AS lp ON (lp.id_choice = pc.id_choice AND lp.id_poll = {int:id_poll} AND lp.id_member = {int:current_member} AND lp.id_member != {int:not_guest})
-			WHERE pc.id_poll = {int:id_poll}',
+			WHERE pc.id_poll = {int:id_poll} ORDER by pc.id_choice',
```
albertlast: BY should be in capitalcase,
the ORDER BY should be in a new line

I'm not 100% sure if this sort shouldn't be done in php instead of db side.

m4z: > BY should be in capitalcase,
> the ORDER BY should be in a new line

I'll fix that (sometime later this week).


> I'm not 100% sure if this sort shouldn't be done in php instead of db side.

I wouldn't know how to do that.

frandominguezl: Why would we do this sort in PHP when we can do it in SQL?

MissAllSunday: In this case there isn't any advantage of using PHP versus doing it on the DB.

Sesquipedalian (approved):

PR #5855, opened 2019-11-04
Modify language vars to reflect custom avatar directory change

### LINKED ISSUE #5853: Strange error mesage when uploading avatar
This is a bit of a strange one.  I went to upload a avatar on my local test install in XAMPP, and was presented with a error on page telling me my attachments folder is not writable.

So checked everything, all folder permission were correct, and the attachment directory is writable.

Now since discovered that the issue was because my custom_avatar folder was missing, now this might be down to me, though I don't know how.  But what I'm more concerned about is the error I got on page and in the log for having the custom_avatar folder missing.

The backtrace info,

Error 
Type of error: Critical
Error message: The attachments directory is not writable
URL of page causing the error: http://localhost/tp2smf2.1rc3/index.php?action=profile;area=forumprofile;u=1

Backtrace information 

#0: fatal_lang_error()
Called from C:\xampp\htdocs\tp2smf2.1rc3\Sources\Profile-Modify.php on line 3467
#1: profileSaveAvatarData()
Called from C:\xampp\htdocs\tp2smf2.1rc3\Sources\Profile-Modify.php on line 821
#2: saveProfileFields()
Called from C:\xampp\htdocs\tp2smf2.1rc3\Sources\Profile.php on line 677
#3: ModifyProfile()
Called from C:\xampp\htdocs\tp2smf2.1rc3\index.php on line 155

I've just done a fresh install from the nightly builds, and can recreate the error very easily.  Just  rename the custom_avatar directory, add a tilde to the end or something.  Then try and upload an avatar. https://github.com/SimpleMachines/SMF2.1/compare/8d19779...release-2.1

You'll then see that exact same error in the log complete with backtrace info.

Also, while folder is renamed, if you visit, Admin > Forum > Attachments and Avatars > Avatar settings.  You'll see the following text highlighted in red.

Upload directory
 This should be a valid and writable directory, different from the server-stored directory.  This error message is correct. 

and 

The Attachments directory is not valid. This will prevent attachments from working properly.   This error message is referring to the wrong directory.

MissAllSunday:
Fixes #5853


Themes/default/languages/Admin.english.php
```diff
@@ -457,7 +457,7 @@
 $txt['avatar_select_permission'] = 'Select permissions for each group';
 $txt['avatar_download_external'] = 'Download avatar at given URL';
 $txt['option_specified_dir'] = 'Specific directory...';
-$txt['custom_avatar_dir_wrong'] = 'The Attachments directory is not valid. This will prevent attachments from working properly.';
+$txt['custom_avatar_dir_wrong'] = 'The Avatars directory is not valid. This will prevent attachments from working properly.';
```
Sesquipedalian: Should probably change “attachments” to “avatars” in the second sentence, too.

Sesquipedalian (changes requested):

Sesquipedalian (approved):

PR #5860, opened 2019-11-07
Data too long check

### LINKED ISSUE #5838: Database Error: Data too long for column 'avatar'
#### Description
errors from log:
Error
Type of error: Database
Error message:
Database Error: Data too long for column 'avatar' at row 1
E:/XAMPP/htdocs/rc2/Sources/Subs.php
Line: 426
URL of page causing the error: http://localhost/rc2/index.php?action=profile;area=forumprofile;u=1
Backtrace information
#0: smf_db_error()
Called from E:\XAMPP\htdocs\rc2\Sources\Subs-Db-mysql.php on line 489
#1: smf_db_query()
Called from E:\XAMPP\htdocs\rc2\Sources\Subs.php on line 426
#2: updateMemberData()
Called from E:\XAMPP\htdocs\rc2\Sources\Profile.php on line 701
#3: ModifyProfile()
Called from E:\XAMPP\htdocs\rc2\index.php on line 155

### Steps to reproduce
1. Forum Profile
2. Personalized picture / Specify avatar by URL
3. Google image, right click image get url ie: https://www.google.com/imgres?imgurl=http%3A%2F%2Fwww.google.com%2Flogos%2Fdoodles%2F2015%2Fgoogles-new-logo-5078286822539264.3-hp2x.gif&imgrefurl=http%3A%2F%2Fwww.google.com%2Fdoodles%2Fgoogles-new-logo&docid=hWfwYn6E9UL_wM&tbnid=wNsZSiIPru5UcM%3A&vet=10ahUKEwj62uXBvbDlAhWQfZoKHcR7DLcQMwhRKAEwAQ..i&w=881&h=400&bih=767&biw=1600&q=google%20image&ved=0ahUKEwj62uXBvbDlAhWQfZoKHcR7DLcQMwhRKAEwAQ&iact=mrc&uact=8
4. Paste the url in input field, hit enter.

- Version/Git revision: 2.1 RC2
- Database Type: MariaDB
- Database Version: 10.1.38-MariaDB - mariadb.org binary distribution
- PHP Version: Apache/2.4.38 (Win32) OpenSSL/1.1.1a PHP/7.3.2

user must get a nice error: `fatal_lang_error('your image url is way to long, please try another one or similar text', false);` message instead of this

googled the error, it is has reported earlier for smf 2.0 here: https://www.simplemachines.org/community/index.php?topic=568670.0 seems to be unnoticed

MissAllSunday:
Fixes #5838

### DIFF
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -18,8 +18,6 @@
 if (!defined('SMF'))
 	die('No direct access...');
 
-const MAX_EXTERNAL_URL_SIZE = 255;
-
 /**
  * This defines every profile field known to man.
  *
@@ -3346,13 +3344,17 @@ function profileSaveAvatarData(&$value)
 	global $modSettings, $sourcedir, $smcFunc, $profile_vars, $cur_profile, $context;
 
 	$memID = $context['id_member'];
+	$context['max_external_size_url'] = 255;
+
 	if (empty($memID) && !empty($context['password_auth_failed']))
 		return false;
 
 	require_once($sourcedir . '/ManageAttachments.php');
 
+	call_integration_hook('before_profile_save_avatar', array(&$value));
+
 	// External url too large
-	if ($value == 'external' && allowedTo('profile_remote_avatar') && strlen($_POST['userpicpersonal']) > MAX_EXTERNAL_URL_SIZE)
+	if ($value == 'external' && allowedTo('profile_remote_avatar') && strlen($_POST['userpicpersonal']) > $context['max_external_size_url'])
 		return 'bad_avatar_url_too_long';
 
 	// We're going to put this on a nice custom dir.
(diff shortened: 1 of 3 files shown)


Sources/Profile-Modify.php
```diff
@@ -18,6 +18,8 @@
 if (!defined('SMF'))
 	die('No direct access...');
 
+const MAX_EXTERNAL_URL_SIZE = 255;
```
Sesquipedalian: I wonder whether it would be better to retrieve the maximum length of the field from the database rather than defining a constant. That way, if the length of the field ever changes, the function will still use the correct value.

Sources/Subs.php
```diff
@@ -375,17 +375,24 @@ function updateMemberData($members, $data)
 	$setString = '';
 	foreach ($data as $var => $val)
 	{
-		$type = 'string';
+		switch ($var){
```
Sesquipedalian: Minor:  bracket should be on next line for SMF formatting conventions.

Sources/Profile-Modify.php
MissAllSunday: I thought about it but the amount of times this length will be changed is quite low and thus it doesn't justify to do a query to the dababase :/

We can always increase the field length on update to a more reasonable value

Sesquipedalian: Fair point. Still, instead of a constant, maybe we should make this a variable and declare it somewhere that will allow mods to change it via hooks instead of file edits.

MissAllSunday: yep, that can be done yes.

Sesquipedalian (approved):

albertlast (approved):
Personally I don't  like that you combine the fix with coding style changes in subs.
Since you mention only the fix and not that you change also some code style in subs. 

But the result look fine to me,
I only checked subs  changes.

PR #5866, opened 2019-11-08
Use constants for DB titles

### LINKED ISSUE #5817: Incorrect DB comparison / value of $smcFunc['db_title']
#### Description
A couple of places in the smileys code check against $smcFunc['db_title'] to see if it contains 'MySQL' so they can use a more specialised comparison.

Unfortunately, Subs-Db-Mysql.php lists itself as MySQLi so that comparison never plays out.

See https://github.com/SimpleMachines/SMF2.1/blob/527a3ea50166b39deb9e03303fead37cd3c9135f/Sources/ManageSmileys.php#L629 for example.

Either the entire edge case should be removed (since it currently does nothing special) or $smcFunc['db_title'] should probably return 'MySQL' instead.

MissAllSunday:
Fixes #5817


albertlast (changes requested):
Well the core problem is still there and you solution make it harder to fix. 
Core problem:
On mysql side you mention the driver and
On postgresql side the database.

So you mix different information in the same field. 

The problem get real when you look in the pdo direction.

Sesquipedalian (approved):

MissAllSunday:
This is more like an identifier rather than  the driver or the db name, thats why we don't need any other identifier for mariaDB for example.

The DB title is solely used for checking if its Postgre or MySQL/MariaDB, thats all. 2.1 won't be using PDO anytime soon.

albertlast:
Than use MySQL instead of MySQLi

MissAllSunday:
Fair enough.

albertlast (approved):

PR #5872, opened 2019-11-19
Attribute "id" is missing…

dragomano:
Attribute "id" is required, because "label for=" works only with HTML elements that have ids.


Sesquipedalian (approved):

Sesquipedalian:
Nice catch.

PR #5884, opened 2019-12-07
New hook - integrate_ini_settings (Session.php)

dragomano:

### DIFF
--- a/Sources/Session.php
+++ b/Sources/Session.php
@@ -34,6 +34,9 @@ function loadSession()
 	@ini_set('session.use_trans_sid', false);
 	@ini_set('arg_separator.output', '&amp;');
 
+	// Allows mods to change/add PHP settings
+	call_integration_hook('integrate_ini_settings');
+
 	if (!empty($modSettings['globalCookies']))
 	{
 		$parsed_url = parse_url($boardurl);
--- a/Sources/Session.php
+++ b/Sources/Session.php
@@ -35,7 +35,7 @@ function loadSession()
 	@ini_set('arg_separator.output', '&amp;');
 
 	// Allows mods to change/add PHP settings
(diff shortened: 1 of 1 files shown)


albertlast:
first question should be a mod able to change the php settings?
second is the database already up at this point?
How are programm error catched by mods at this point of time?

dragomano:
The database is not required there, $modSettings is enough :)
For example, `@ini_set('session.use_only_cookies', !empty($modSettings['some_option']));`

albertlast:
did you check how call_integration_hook is working?

dragomano:
Of course.

albertlast:
modSettings gefilled by the database... 
so you need databse.

Sources/Session.php
```diff
@@ -34,6 +34,9 @@ function loadSession()
 	@ini_set('session.use_trans_sid', false);
 	@ini_set('arg_separator.output', '&amp;');
 
+	// Allows mods to change/add PHP settings
+	call_integration_hook('integrate_ini_settings');
```
Sesquipedalian: Having a hook at this point in `loadSession()` makes sense, if the hooked code relates to loading sessions. 

This is not, however, a good place for a generic hook to change arbitrary ini settings, so the name of this hook should not suggest that to mod developers. Please change it to `integrate_load_session`. I know it is only cosmetic, but we want to communicate the right idea with the name.

Sesquipedalian (changes requested):

Sources/Session.php
dragomano: Ok, done. Currently, it is the only way to change such variables (before starting a session). If you try to change them later, there will be a warning:
`: ini_set(): A session is active. You cannot change the session module's ini settings at this time in...`

Sesquipedalian (approved):

PR #5887, opened 2019-12-11
Smallfixes

### LINKED ISSUE #5867: Text Overflow in error log
#### Description
Some fields in the error log overflow & overwrite adjacent labels/text.  

### Steps to reproduce
1. Note in the below, the 2nd item properly wraps, but the 3rd one does not:
image

- Database Type: MySQL
- Database Version: 5.7
- PHP Version: 7.1

Originally reported here:
https://www.simplemachines.org/community/index.php?topic=568063.0

XinYenFon:
fixes #5867

There are some rare situations where ID in the link gets bigger than the expected so it overlaps with the other area, breaking those IDs does no harm.


SychO9:
don't wanna be annoying but `break-all` does look better in this case xD

XinYenFon:
weird, I do recall that I used -all... This is why you dont do changes at 5 am in the morning...

SychO9 (approved):
Looks good

jdarwood007 (approved):

PR #5892, opened 2019-12-14
Error log generates a error

jdarwood007:
If the filter isn't set, we can cause a error to occur, ensure that filter is set before we compare it.


MissAllSunday (approved):

Oldiesmann:
Looks like I need to update my local test copy then. Thanks!

PR #5893, opened 2019-12-14
Alerts Configuration generates errors

jdarwood007:
When loading the Alerts configuration in ACP errors can be generated because we attempt to use $cur_profile, which from previous loadThemeOptions had specified 0 for the $memID, which results in a invalid $cur_profile.  To work around this we simply are adding a way to bypass the usage of $cur_profile and load Theme options from id_theme = 0.

This may need more testing than I gave it.


Sources/Profile-Modify.php
```diff
@@ -2897,7 +2899,7 @@ function loadThemeOptions($memID)
 			WHERE id_theme IN (1, {int:member_theme})
 				AND id_member IN (-1, {int:selected_member})',
 			array(
-				'member_theme' => (int) $cur_profile['id_theme'],
+				'member_theme' => !isset($cur_profile['id_theme']) && $defaultSettings ? 0 : (int) $cur_profile['id_theme'],
```
Sesquipedalian: Please change this to `!isset($cur_profile['id_theme']) && !empty($defaultSettings)`, just in case someone passes a goofy value to the `$defaultSettings` parameter.

Sesquipedalian (changes requested):

Sesquipedalian (approved):

PR #5895, opened 2019-12-15
Add some more hooks

jdarwood007:
Some additional hooks and changes to hooks to make more use of the hooks system.

### DIFF
--- a/Sources/ManageBans.php
+++ b/Sources/ManageBans.php
@@ -514,6 +514,9 @@ function BanEdit()
 			return confirm("' . $txt['ban_remove_selected_confirm'] . '");
 		});',
 			);
+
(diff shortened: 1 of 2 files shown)


MissAllSunday (approved):

SychO9:
I was gonna say about those hooks after looking into the AntiSpamLinks mod, we might need more from what I've seen

jdarwood007:
@SychO9 Yea these are all from converting the Anti-Spam Links mod.

This part of SMF code is not very friendly to work with.  But I've come up with something that seems to work.

SychO9:
nice, let me know how it goes and if you need help with it

jdarwood007:
I've sent you info with the link to the code.  Try it out.
I may add more to this for some other mods, but it can be pulled at any time.  Will do more PRs if needed.

PR #5896, opened 2019-12-15
Dont exceed range

### LINKED ISSUE #5890: Out of range value for column page_hits
#### Description
Some hyper-aggressive recrawls exceed the max page_hits in the log_spider_stats table.  It's an unsigned smallint, max 65K.  

Although a proper fix would be a schema change to use an int, my initial thought is to just trap for >65535 and not increment.  I'm not sure there is a ton of value in having an exact figure at that point.

image

image

sbulen:
Hyper aggressive crawlers like MSN can flood the stats records.

Fixes #5890 

I am assuming we're not going to change the schema at this point.  

Note this does not work for pg, mainly because the pg smallint is signed & the max value is 32767, not 65535.  Open to input.

### DIFF
--- a/Sources/ManageSearchEngines.php
+++ b/Sources/ManageSearchEngines.php
@@ -558,7 +558,8 @@ function logSpider()
 		$date = strftime('%Y-%m-%d', forum_time(false));
 		$smcFunc['db_query']('', '
 			UPDATE {db_prefix}log_spider_stats
-			SET last_seen = {int:current_time}, page_hits = page_hits + 1
+			SET last_seen = {int:current_time},
+				page_hits = CASE WHEN page_hits + 1 > 65535 THEN 65535 ELSE page_hits + 1 END
 			WHERE id_spider = {int:current_spider}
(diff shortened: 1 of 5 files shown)


jdarwood007:
I would bump the size of the field.  No code changes and not taking much more storage.

sbulen:
...and it would fix pg, too.

albertlast:
the idea is smart,
but like mention pg got the half amount as limit and
the size diferenz between 2 byte int and 4 byte int is not so important im complet smf database structure.
Would also change the datetype of the field.

sbulen:
Went with an INT, for conformity across pg & mysql...  

Tested installs & upgrades, both mysql & pg.  Everything was good, but, found different pg-related issues, logged separately.

albertlast (approved):

albertlast:
I like that you also removed the unsigned from the mysql of this field,
this make the min/max values between mysql and pg setup even.

MissAllSunday (approved):

PR #5897, opened 2019-12-15
Fixes errors with bad IPs

jdarwood007:
If bad data exists in the database for a binary IP, SMF will generate a error:
Warning: inet_ntop(): Invalid in_addr value in /Sources/Subs.php on line 6001

This simply tests if its a valid IP already and sends it back on through.

This doesn't fix the issue (as it doesn't know the source) and won't prevent areas that may be broken from compare operations.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -5990,17 +5990,16 @@ function inet_ptod($ip_address)
  */
 function inet_dtop($bin)
 {
-	if (empty($bin))
-		return '';
-
 	global $db_type;
 
-	if ($db_type == 'postgresql')
+	if (empty($bin))
+		return '';
+	elseif ($db_type == 'postgresql')
 		return $bin;
-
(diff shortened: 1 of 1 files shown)


albertlast:
Like mention in the irc,
i'm not sure if this a good way to hide the issue,
this is the way how the pr working.

We could place somekind of repair/check stuff,
but this could be a different pr....

Would be nice to know how the data got inserted wrong.

albertlast (approved):
in general the aproache is valid.

jdarwood007:
I will work something up to repair them.  I believe this needs a two fold to resolve.
We could log errors when we trigger this to indicate to the admin to run the repair.

MissAllSunday:
Yeah it will be cool to try and repair corrupted data, we just need to be careful not to oversee this and never implement it.

MissAllSunday (approved):

Issue #5836, opened 2019-10-21
User selected timezone not displayed properly in profile


sbulen:
#### Description
User time zone not displayed properly in profile.  

### Steps to reproduce
1. Select a new timezone.
2. Navigate away & return - it is blank.

Note it is properly set in the DB, it's just not displayed correctly.

### Environment (complete as necessary)
- Version/Git revision:  Current
- Database Type:  pg
- Database Version:  10.10
- PHP Version:  7.1

### Additional information/references
Set it:
![image](https://user-images.githubusercontent.com/23568484/67169197-d787fe00-f35e-11e9-93af-ac155709878b.png)

But it's gone when you come back:
![image](https://user-images.githubusercontent.com/23568484/67169214-ee2e5500-f35e-11e9-975a-226081d6e04e.png)

But it's saved properly in the DB:
![image](https://user-images.githubusercontent.com/23568484/67169247-1453f500-f35f-11e9-920d-05e77ae38b4e.png)

sbulen:
I am closing this.  

Refreshed my environments & not seeing it anymore.

PR #5900, opened 2019-12-17
Lock on writing cache

### LINKED ISSUE #5826: Infrequent caching error for modSettings
#### Description
Error message:  
`

### Steps to reproduce

- Database Type: various
- Database Version: various
- PHP Version: various

Sample threads:

MissAllSunday:
This is my approach to fixes #5826

It uses  rather simple method to lock a file while its been created, taking advantage of directories been able to be created and deleted but not overwritten we can bypass the race condition created when two or more process attempt to modify the same cache entry.

The function as it is return false when theres a lock on the file and thus the cache entry won't be created, it can be changed to do a re-attempt either via recursiveness or other mechanisms.

### DIFF
--- a/Sources/CacheAPI-smf.php
+++ b/Sources/CacheAPI-smf.php
@@ -111,7 +111,7 @@ class smf_cache extends cache_api
 			// Write out the cache file, check that the cache write was successful; all the data must be written
 			$fileSize = $this->filePutWithLock($file_name, $cache_data);
 
-			// Check the file againts what its suppose to have in it.
+			// Check the file against what its suppose to have in it.
 			if ($fileSize === strlen($cache_data))
 				$file_created = true;
 
--- a/Sources/CacheAPI-smf.php
+++ b/Sources/CacheAPI-smf.php
@@ -215,6 +215,26 @@ class smf_cache extends cache_api
 	{
 		return SMF_VERSION;
 	}
+
+	/**
+	 * @param string $file_name
+	 * @param string $cache_data
+	 * @return int
+	 */
+	protected function filePutWithLock($file_name, $cache_data)
+	{
+		$saved_file = 0;
+		$tmp_lock_file = $this->cachedir . '/tmp/'. str_replace('.php', '.lock', $file_name);
+
+		if (mkdir($tmp_lock_file, 0700)) {
+
+			$saved_file = file_put_contents( $file_name, $cache_data, LOCK_EX);
+
+			rmdir($tmp_lock_file);
+		}
+
+		return $saved_file;
+	}
 }
 
 ?>
--- a/Sources/CacheAPI-smf.php
+++ b/Sources/CacheAPI-smf.php
@@ -86,19 +86,20 @@ class smf_cache extends cache_api
 	public function putData($key, $value, $ttl = null)
 	{
 		$key = $this->prefix . strtr($key, ':/', '-_');
-		$file_name = $this->cachedir . '/data_' . $key . '.php';
+		$file_name = '/data_' . $key . '.php';
+		$file_path = $this->cachedir . $file_name;
 
 		// Work around Zend's opcode caching (PHP 5.5+), they would cache older files for a couple of seconds
 		// causing newer files to take effect a while later.
 		if (function_exists('opcache_invalidate'))
-			opcache_invalidate($file_name, true);
+			opcache_invalidate($file_path, true);
 
 		if (function_exists('apc_delete_file'))
-			@apc_delete_file($file_name);
+			@apc_delete_file($file_path);
 
 		// Otherwise custom cache?
 		if ($value === null)
-			@unlink($file_name);
+			@unlink($file_path);
 
 		else
 		{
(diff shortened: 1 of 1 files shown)


jdarwood007:
The rmfile outside of the protected function could still cause a race condition here.  As the protected function would return 0 in one thread while the other writes the file and then outside it would trigger a deletion on the first thread.  Is that a concern?

MissAllSunday:
mmm not sure if I follow correctly,which outside rmfile()?

Sources/CacheAPI-smf.php
```diff
@@ -86,34 +86,38 @@ public function getData($key, $ttl = null)
...
 		else
 		{
 			$cache_data = '<' . '?' . 'php if (!defined(\'SMF\')) die; if (' . (time() + $ttl) . ' < time()) $expired = true; else{$expired = false; $value = \'' . addcslashes($value, "\0" . '\\\'') . '\';}' . '?' . '>';
 
 			// Write out the cache file, check that the cache write was successful; all the data must be written
 			// If it fails due to low diskspace, or other, remove the cache file
-			$fileSize = file_put_contents($cachedir . '/data_' . $key . '.php', $cache_data, LOCK_EX);
+			$fileSize = $this->filePutWithLock($file_name, $cache_data);
+
 			if ($fileSize !== strlen($cache_data))
```
jdarwood007: Line 112 checks the file size and if it doesn't match issues a unlink. filePutWithLock will return a 0 if it can't reach a lock.

albertlast:
Is this pr affect the performance of file cache?

Sources/CacheAPI-smf.php
MissAllSunday: Changed the logic to not delete the file if the method returned false.  If returned false it means the file wasn't created on this attempt.

MissAllSunday:
> Is this pr affect the performance of file cache?

The file  cache system is already slow due to the filesystem access itself.

This is indeed a curious case. We are trying to fix a symptom rather than a cause. Whoever hits this particular bug should start looking at some more robust cache functionalities (memcached, redis).

I'm more interested on preventing corrupted/malformed files rather than trying to optimize a filesystem operation. For the SMF average user (shared hosting), the call to mkdir() won't represent a real impact on their execution times.

albertlast:
than implement a mysql version, like the pg version already exists, and drop the file based cache.

MissAllSunday:
Yeah but how many servers have that configuration enable? Adding it as an option sounds good but I wouldn't replace the file based one with it.

albertlast:
In  my eyes your are not able to fix the issue in a way that work stable  and is still faster as none cached setup.
The last point is the reason why this cache exists, 
So when this not archived the solution is useless.

MissAllSunday:
And what exactly is "the issue" to you?

For me it is to avoid having corrupted cache files due to race conditions which is what #5826 is about.

albertlast:
Yes and follow up issue is that file cache  get slower.

MissAllSunday:
Where on #5836 does it says file cache gets slower?

albertlast:
This pr do this.

MissAllSunday:
:sweat_smile: 

We really, really need to stop worrying about micro-optimizations.

albertlast:
Like the last time where you extend the runtime of smf by two times?
When you do changes where  runtime is all, 
You need to provide a benchmark how  mutch you change effects the runtime.

When you mean it's not worth than drop three support of file cache

MissAllSunday:
I'm not worried about performance, at all. Like I said, file based is already slow and no, there isn't a single thing we can do to make it faster.

I really don't see how this can double the execution time as you claim but if you do, please attach your benchmarks.

But I repeat, we really shouldn't be worrying about micro-optimizations, specially in a product like SMF where we can't control where its been deployed.

albertlast:
Performance is key,
Everything which harm this is useless.

MissAllSunday:
Better start re-writting SMF in assembly then :smile:

MissAllSunday:
> The rmfile outside of the protected function could still cause a race condition here. As the protected function would return 0 in one thread while the other writes the file and then outside it would trigger a deletion on the first thread. Is that a concern?

I couldn't find any other call to rmfile outside the API implementation.

jdarwood007:
I can't find the code i commented on originally, so it must be patched now

sbulen:
Note - this PR has spawned some merge conflicts...

sbulen:
I've been testing this.   To reproduce the error & test a fix, I wrote two bonehead scripts - one that writes cache 1000x and one that reads cache 1000x.  Running both does not in itself reproduce the issue, however, running TWO of the PUT scripts causes all kinds of issues with the GET scripts (confirming Suki's idea that it's related to conflicting writes).

Test results:
 - The T_CONSTANT_ENCAPSED_STRING remain on linux
 - This introduced a new issue with the mkdir; basically, the write conflicts have been xferred to mkdir conflicts
 - This dramatically improved successful reads on WIN

We should see 1000 reads, but it never gets there due to the T_CONSTANT_ENCAPSED_STRING errors...
smf21-PR5900-linux-fails
The errors from the apache log:
smf21-current-code-linux-err

The mkdir errors.  
smf21-PR5900-linux-error

Where this PR resulted in a dramatic improvement was on WIN...  Before this PR, the GET returned NULL almost 50% of the time:
smf21-current-code-WIN

But those are rare with this PR:
smf21-PR5900-WIN

PR #5915, opened 2020-01-04
Few more Hooks

jdarwood007:


albertlast:
Would be nice to maintain somekind of "list" when we add new hooks...

jdarwood007:
We can easily grep for all hooks in SMF.

I'm converting some mods and finding places I need hooks. So getting these PRs going to get all of them using hooks.

albertlast:
Some hooks are hard to find by a grep and
you still missing information around this when you only mention them.

PR #5924, opened 2020-01-07
Fixes #5914

### LINKED ISSUE #5914: Proxy cache cleanup
#### Description
The 2.0 proxy cache cleanup has a couple issues.  The 2.1 version is functionally similar.   

It needs to:
 - skip directories
 - skip files like index.php

See 2.0 issue here: https://www.simplemachines.org/community/index.php?topic=571162.0

Sesquipedalian:


jdarwood007:
readdir also returns '.' and '..'.  Should also ignore these.

Sesquipedalian:
> readdir also returns '.' and '..'. Should also ignore these.

No need. `is_file($path . $file)` already covers that.

PR #5925, opened 2020-01-07
Does a better job of ensuring $cachedir is defined and writable

Sesquipedalian:
Also improves path checking in Settings.php.


sbulen:
Any time you change Settings.php these days, you must also change the $settingsArray in upgrade.php.

sbulen:
I don't think we wanted the !is_dir checks on $sourcedir, etc...   I think you meant is_file...

Sesquipedalian:
> I don't think we wanted the !is_dir checks on $sourcedir, etc... I think you meant is_file...

Oh yeah, forgot about symlinks.

Sesquipedalian:
#5948

PR #5926, opened 2020-01-07
Converts global $time_start to constant TIME_START

Sesquipedalian:
Previously we were inconsistently using a constant in cron.php and a global variable elsewhere. Using a constant makes the most sense for this value.


jdarwood007:
We use should !defined incase a integration with another software uses the same define.  I don't know of any, but worry about using common names in defines causing this.  Worse case with a common name of $time_start is that we get wrong times on page generation stats.  A define will produce a fatal error if it is defined again.

Sesquipedalian:
Makes sense

PR #5930, opened 2020-01-09
Removes unnecessary escaping of some Settings.php values in installer

Sesquipedalian:
I could have sworn I included this in #5815, but apparently not. (Probably got stashed but not restored at some point.)

### DIFF
--- a/other/install.php
+++ b/other/install.php
@@ -789,12 +789,12 @@ function DatabaseSettings()
 
 		// Take care of these variables...
 		$vars = array(
-			'db_type' => addcslashes($db_type, '\'\\'),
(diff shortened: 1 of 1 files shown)


albertlast:
I'm not sure if this a good idea,
this lines are not realy old...
https://github.com/SimpleMachines/SMF2.1/commit/1c85fd040d5873bdd5204eec31b403debff7517d
was part of issue which got fixed by this.

Sesquipedalian:
> I'm not sure if this a good idea,

It is now that #5815 has been merged. We no longer need to manually escape any values passed to updateSettingsFile().

PR #5931, opened 2020-01-09
Fix undefined variable: approve_query

### LINKED ISSUE #5877: Search.php: 8: Undefined variable: approve_query
#### Description
Originally reported [here](https://www.simplemachines.org/community/index.php?topic=570651.0): Since ~2019-10-10, the GH version produces the following errors when using the search function:

```
Code: [Select]
Type of error: Undefined
8: Undefined variable: approve_query
 http://localhost:8080/smf/index.php?action=search2
 C:/wamp64/www/smf/Sources/Search.php (Line 1715)

Type of error: Undefined
8: Undefined variable: approve_query
 http://localhost:8080/smf/index.php?action=search2
 C:/wamp64/www/smf/Sources/Search.php (Line 1828)
```
These lines were last changes 6 months ago ([L1715](https://github.com/SimpleMachines/SMF2.1/blame/0009c2d072ef77f8f8f869def5986d7a9449e2b1/Sources/Search.php#L1715), [L1828](https://github.com/SimpleMachines/SMF2.1/blame/0009c2d072ef77f8f8f869def5986d7a9449e2b1/Sources/Search.php#L1828)), but I didn't see this before 2019-10-10 in my installs (but I didn't explicitly test the search function and my users don't seem to use it much). Also just confirmed this in a freshly updated nightly build from 15 minutes ago.

### Steps to reproduce
1. Install from/update to a recent version 
2. Search something (posting from entire forum or current board, or a member) in the top-right search bar
3. if the search has no result, only the first error occurs; if it has one, both appear

- Version/Git revision: HEAD
- Database Type: any? (confirmed on psql, assume sm.org reporter uses mysql)
- Database Version: any?
- PHP Version: any?

live627:
- use the mod cache instead of a whole new query
- Redoing the comments because how does English work?

Fixes #5877

### DIFF
--- a/Sources/Search.php
+++ b/Sources/Search.php
@@ -1687,22 +1687,17 @@ function PlushSearch2()
 			}
 		}
 
+		$approve_query = '';
 		if (!empty($modSettings['postmod_active']))
 		{
-			$approve_boards = boardsAllowedTo('approve_posts');
-
-			// Can approve everywhere, so search all topics.
-			if ($approve_boards === array(0))
-		 		$approve_query = '';
-
-		 	// Can't approve anywhere, so search ony their own topics and approved topics.
-		 	elseif (empty($approve_boards))
-		 		$approve_query = '
+			// Exclude unapproved topics, but show ones they started.
+			if (empty($user_info['mod_cache']['ap']))
+				$approve_query = '
 				AND (t.approved = {int:is_approved} OR t.id_member_started = {int:current_member})';
 
-		 	// Can approve in some boards, so search own, approved, and approvable topics.
-		 	else
-		 		$approve_query = '
+			// Show unapproved topics in boards they have access to.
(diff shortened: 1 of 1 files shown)


sbulen:
live - Is this for #5877?   
If so, could you put that in your initial description?  That way the issue & pr are linked.

live627:
they are linked via the commit  which is enough for github

sbulen:
Yeah, but not for us suckers trying to manually maintain the "has fix" tags.  

Can you tag them?

Or at least make them easier to find?

Sesquipedalian (approved):

XinYenFon:
may i ask why this marked for final? Its a common error users can face, better suited for upcoming RC.

sbulen:
The cat in a tie is correct.
I was able to reproduce the issue & confirm this addressed it.

m4z:
I concur, no more errors. ❤️

Issue #5858, opened 2019-11-04
When I myself move or sticky a topic, I get notified (alert + mail)


m4z:
#### Description
For details on my notification settings, please see the [original report here](https://www.simplemachines.org/community/index.php?topic=569575.0).

### Steps to reproduce
1. Move or sticky a topic
2. Receive alert and notification email

### Environment (complete as necessary)
- Version/Git revision: GH version from ~2019-10-10
- Database Type: Postgres but probably any
- Database Version: 9.6ish
- PHP Version: 5.6ish

m4z:
Also, when I @mention myself, I'm notified too (maybe only alerted; just did a quick test related to https://www.simplemachines.org/community/index.php?topic=570456.0 on one of my installs that might have incomplete mail setup, i.o.w. I didn't receive a mail notification and won't spend time on the Why right now).

PR #5932, opened 2020-01-09
exclude the current user from $to_names

live627:


albertlast (commented):
You are  sure from I'd is always maintained?

live627:
Positive. It is used a few lines down


https://github.com/SimpleMachines/SMF2.1/blob/f610b7927976c6c764bc348bce09a85e57f48810/Sources/Subs-Post.php#L1151

And declared at

https://github.com/SimpleMachines/SMF2.1/blob/f610b7927976c6c764bc348bce09a85e57f48810/Sources/Subs-Post.php#L837-L842

sbulen:
live - is this for #5858 ?

Sesquipedalian (approved):

PR #5941, opened 2020-01-10
Update Birthday-Notify.php

### LINKED ISSUE #5911: Birthdate for those with no birthdates set
### Description
The default **birthdate** column value in the smf_members table is **1004-01-01**, that means that the software by default determines that all who have not set their birth dates, their birth dates are the 1st of January

### Source code in question
I believe these are the two spots of the source code that we should tweak (add conditions that the year value must be upper than 1004 perhaps?), unless we want to edit the column's default value.

Line 81 of **Profile-View.php**
https://github.com/SimpleMachines/SMF2.1/blob/6007069567bd78d1a72dd56dc61590a074258be4/Sources/Profile-View.php#L79-L82

And the query in the birthday scheduled task
https://github.com/SimpleMachines/SMF2.1/blob/6007069567bd78d1a72dd56dc61590a074258be4/Sources/tasks/Birthday-Notify.php#L31-L48

https://www.simplemachines.org/community/index.php?topic=571106.0
https://www.simplemachines.org/community/index.php?topic=571108.0

live627:
Fix #5911

### DIFF
--- a/Sources/Profile-View.php
+++ b/Sources/Profile-View.php
@@ -78,7 +78,7 @@ function summary($memID)
 		$datearray = getdate(forum_time());
 		$context['member'] += array(
 			'age' => $birth_year <= 1004 ? $txt['not_applicable'] : $datearray['year'] - $birth_year - (($datearray['mon'] > $birth_month || ($datearray['mon'] == $birth_month && $datearray['mday'] >= $birth_day)) ? 0 : 1),
-			'today_is_birthday' => $datearray['mon'] == $birth_month && $datearray['mday'] == $birth_day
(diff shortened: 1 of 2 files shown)


SychO9:
You also need to update Line 81 of **Profile-View.php** to test that the year is valid (`$birthyear > 1004` or smth) to remove the cake icon from those kind of profiles.
https://github.com/SimpleMachines/SMF2.1/blob/6007069567bd78d1a72dd56dc61590a074258be4/Sources/Profile-View.php#L79-L82

SychO9 (approved):

Sesquipedalian (approved):

PR #5944, opened 2020-01-10
Check if ip address is empty

albertlast:
When cron job create an error the ip field can be empty(not null) and
empty inet is not allowed.
This catch the case.
affect only pg


albertlast:
notice that i fioxed already the issue in this pr: https://github.com/SimpleMachines/SMF2.1/pull/5222
for random reason is this pr defined for final...

PR #5946, opened 2020-01-10
Add index Support for Birthday task pg only

albertlast:
Since i saw the query in the pr #5945 
i notice that this query didn't use the birthday index for pg,
so this fix it.

Query execution time goes down from 0.900 ms to 0.050 ms at 2014 entries.

Don't get confused by the year,
it's literaly random.

### DIFF
--- a/Sources/tasks/Birthday-Notify.php
+++ b/Sources/tasks/Birthday-Notify.php
@@ -40,7 +40,7 @@ class Birthday_Notify_Background extends SMF_BackgroundTask
 				AND MONTH(birthdate) = {int:month}
 				AND DAYOFMONTH(birthdate) = {int:day}
 				AND YEAR(birthdate) > {int:year}
-				' . ($smcFunc['db_title'] === POSTGRE_TITLE ? 'AND indexable_month_day(birthdate) = {date:bdate}' : ''),
+				' . ($smcFunc['db_title'] === POSTGRE_TITLE ? 'AND indexable_month_day(birthdate) = indexable_month_day({date:bdate})' : ''),
 			array(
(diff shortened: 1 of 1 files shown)


jdarwood007:
Why is 2020 hardcoded?

albertlast:
Because a date need a year,
But the year is unimportant
For the result.

In 2020 we got also the 29. Feburary.

Sources/tasks/Birthday-Notify.php
```diff
@@ -39,11 +39,13 @@ public function execute()
...
 				AND MONTH(birthdate) = {int:month}
 				AND DAYOFMONTH(birthdate) = {int:day}
-				AND YEAR(birthdate) > {int:year}',
+				AND YEAR(birthdate) > {int:year}
+				' . ($smcFunc['db_title'] === POSTGRE_TITLE ? 'AND indexable_month_day(birthdate) = indexable_month_day({date:bdate})' : ''),
 			array(
 				'year' => 1,
 				'month' => $month,
 				'day' => $day,
+				'bdate' => '2020-' . $month . '-' . $day,
```
Sesquipedalian: For the sake of future developers reading this, please add a comment like "// This uses 2020 merely because it is a leap year." Otherwise, this looks good.

Sesquipedalian (changes requested):

Sources/tasks/Birthday-Notify.php
albertlast: I will change this to 1004 like every else where we need a place holder year.

MissAllSunday (approved):

PR #5957, opened 2020-01-16
Fileinfo ext

### LINKED ISSUE #5913: Installer/upgrader should check for fileinfo
#### Description
From: https://www.simplemachines.org/community/index.php?topic=571155.0

> This "undefined function dl()" error has cropped up previously in the past couple of days.
> 
> The dl() function seems to be called by PHP when a required extension has not been nominated in the php.ini file to handle an operation (eg, a file interrogation operation). Some code, somewhere, is trying to circumvent that missing object by doing a dynamic load of the required extension. The dl() function has been deprecated for quite some time as extensions in the php.ini file are the way forward.
> 
> The right answer here is to ensure the minimum extension set has been entered in the php.ini file, including (but not limited to)
> 
> [code=Windows NT]extension=php_fileinfo.dll[/code]
> 
> - or (I think) -
> 
> [code=Linux]extension=php72-fileinfo[/code]
> 

I think the extension is just fileinfo, but the notion is correct.

MissAllSunday:
fixes #5913

I didn't test this because this is a pretty niche problem (and I don't have any windows machine), you need to be using either 

- PHP 5.3
- Windows configuration (where the ext is disabled by default for some reason)
- Explicitly disabling the ext

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -848,6 +848,11 @@ function loadUserSettings()
 	call_integration_hook('integrate_user_info');
 }
 
+function loadMinUserSettings($user_id = 0)
+{
+
+}
+
 /**
  * Check for moderators and see if they have access to the board.
  * What it does:
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -848,11 +848,6 @@ function loadUserSettings()
 	call_integration_hook('integrate_user_info');
 }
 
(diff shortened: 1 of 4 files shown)


live627:
Also needs my ax!

... also whitespace...

MissAllSunday:
superseded by #6011

PR #5958, opened 2020-01-18
Scrutinizer Auto-Fixes

live627:
This commit consists of patches automatically generated for this project on https://scrutinizer-ci.com


MissAllSunday:
I much, much prefer if there was a line between if/else statements. 

```php
            if ($value === $foo) 
            {
                continue;
            }
            else 
	    {
                $value = $foo;
	    }
```

NO

```php
            if ($value === $foo) {
                continue;
            } else {
				$value = $foo;
            }
```

HELL NO!

Sources/Subs.php
```diff
@@ -1041,11 +1038,11 @@ function get_date_or_time_format($type = '', $format = '')
 			// Anything that isn't a specification, punctuation mark, or whitespace.
 			'~(?<!%)\p{L}|[^\p{L}\p{P}\s]~u',
 			// A series of punctuation marks (except %), possibly separated by whitespace.
-			'~([^%\P{P}])(\s*)(?'.'>(\1|[^%\P{Po}])\s*(?!$))*~u',
+			'~([^%\P{P}])(\s*)(?' . '>(\1|[^%\P{Po}])\s*(?!$))*~u',
```
sbulen, line 1041: Are we sure we want these changes?  Seems to change the nature of the regex...

sbulen (commented):
Automated changes always worry me...  I have fixed a LOT of bugs caused by automated changes.  Especially the removal of "unnecessary" globals...

At the very least, double-check the regexes.  I think they got broken.

Sources/Subs.php
sbulen: Ah...  Maybe OK?  Looks like an 'atomic group'.  Hadn't seen one of those before...

Sesquipedalian: This is fine. It's just adding some whitespace between concatenated parts of the regex. We concatenate the parts to make sure that `?>` in the regex will never be interpreted by PHP as a closing tag. (Normally it won't be because it is in a string, but if the line containing it is ever commented out, PHP suddenly dies a screaming death.)

Sources/Attachments.php
```diff
@@ -199,7 +199,6 @@ protected function processAttachments()
 			list ($context['attachments']['quantity'], $context['attachments']['total_size']) = $smcFunc['db_fetch_row']($request);
 			$smcFunc['db_free_result']($request);
 		}
-
```
Sesquipedalian, line 202: Like @MissAllSunday, I'm not a fan of removing all the vertical whitespace between if/elseif/else blocks. Technically it is unnecessary, sure, but sometimes it really helps legibility.

Sources/Profile-Modify.php
```diff
@@ -500,7 +500,7 @@ function $preload:		A function that is used to load data required for this eleme
 						log_error(sprintf($txt['smiley_set_dir_not_found'], $set_names[array_search($set, $context['smiley_sets'])]));
 
 						$context['smiley_sets'] = array_filter($context['smiley_sets'], function($v) use ($set)
-							{
+						{
```
Sesquipedalian: If the indentation is reduced on this line, it should be reduced on the following lines as well.

Sources/Subs-BoardIndex.php
```diff
@@ -434,7 +434,7 @@ function getBoardIndex($boardIndexOptions)
 					$board['last_post']['last_post_message'] = sprintf($txt['last_post_message'], $board['last_post']['member']['link'], $board['last_post']['link'], $board['last_post']['time'] > 0 ? timeformat($board['last_post']['time']) : $txt['not_applicable']);
 			}
 		}
-	else
+		else
```
Sesquipedalian: This one is incorrect.

Sesquipedalian (changes requested):

PR #5960, opened 2020-01-18
Add support to validate packages against validation servers

jdarwood007:
This is a feature that I am looking to add into SMF.  This was discussed briefly amongst the developers.  There may be more additions to add to this as well.  I'm tagging RC3 as we want to test this out to ensure packages are still working as expected.

The goal of this addition is to allow packages to be validated against trusted sources, to help inform admins if a source has this package (package exists).  It also serves to allow to blacklist packages based on file name, id and hash.  By doing so it should be possible for a source to mark a malicious or unsafe customization and allow the admin to make an informed decision as to install or not.

Below is a few screenshots of what happens during installs
Screen Shot 2020-01-18 at 1 17 31 PM
Screen Shot 2020-01-18 at 1 17 49 PM
Screen Shot 2020-01-18 at 1 18 18 PM

This also adds Simple Machines Download site as a package server.  The purpose is to allow patches to be directly served from our downloads site.  This can be tested by emulating 2.0.16.

As well this allows both package servers and validation servers to receive the real/emulated version.  The purpose of this is to allow package servers/validation servers to also limit/filter results based on the sent SMF Version.  We allow emulated versioning here for this purpose.

While privacy concerns may come from this, I felt adding an option to disable this could result in admins not being aware of a malicious package installed.  If this is is a concern, we can add the option if we find a way to ensure the admin is well aware of no checks being performed.

This adds a 'extra' column to package_servers table that is currently not used.  Its purpose is to be a json array of data for extending/supporting additional needs in the future should we have them.  At this time I can't think of anything to use this for, but it may become useful.

As well there is support for sending multiple packages at once, however we currently don't make use of this function.  Validation servers are expected to accept multiple validations at once and return them as expected.

Data is built into an array with the following data
- sha256_hash: A sha256 hash of the package
- file_name: The file name of the package
- custom_id: The theme/mod id of the package
- custom_type: What type of package this is.  Modification, theme, smiley, etc.

This array data should be wrapped in another array and then encoded in JSON before being sent.

Validation servers will unpack the json and loop through all arrays, performing validation tests.  Validation servers will build the data in the following array format
sha256_hash as the key: result (unknown, clean, blacklist)

This data is json encoded and returned.  Any other responses other than valid json should be ignored.
The endpoints may change still, especially for the customize site as we haven't built an endpoint for it.


sbulen:
Can we test this at this time?
 - Against 2.1 installs?
 - 2.1 Upgrades?
 - Mods?
 - Mod sites?

jdarwood007:
You can upload the 2.0.17 patch file and it will validate against the downloads server.  It will fail to install, but at least shows you it will function when 2.0 becomes final and starts getting patches.

I have a file I used for the Evil server to randomly generate a result just for testing purposes only.  I could upload that so you could setup a package server response from it.

Mods/Themes will require @SychO9 here as we need to build the endpoint into the customize site.

other/upgrade_2-1_mysql.sql
```diff
@@ -537,11 +537,55 @@ ADD INDEX `idx_id_member` (`id_member`, `id_group`);
...
+
+---# Ensure The Simple Machines Customize Site is https
+UPDATE {$db_prefix}package_servers
+SET url = 'https://custom.simplemachines.org/packages/mods'
+WHERE url = 'http://custom.simplemachines.org/packages/mods';
+---#
+
+---# Add validation to Simple Machines Customize Site
+UPDATE {$db_prefix}package_servers
+SET validation_url = 'https://custom.simplemachines.org/validate.php?api=v1;smf_version={SMF_VERSION}'
```
SychO9: can we use `https://custom.simplemachines.org/api.php?action=validate;version=v1;smf_version={SMF_VERSION}` instead please ?

SychO9 (commented):
Are theme installations not handled ?

jdarwood007:
I intended to support themes, smileys, etc.
Themes use their own installer though that doesn't use a preinstall check page (unless I'm remembering wrong).
What would be the best area to tap this into then?

Side Opinion here, themes should be treated like mods during installs...  Makes upgrades a bit difficult though.

SychO9:
oh that's right, we'd have to create a preinstall page for themes and I'm not sure how reasonable that is at this stage.

jdarwood007:
Really it would be a good idea, but yea requires a lot of changes to introduce that.  This change here is minimal to reduce any new bugs introduced by this feature.

Themes/default/Packages.template.php
```diff
@@ -50,6 +55,31 @@ function template_view_package()
 			!empty($context['failure_details']) ? '<br><br><strong>' . $context['failure_details'] . '</strong>' : '', '
 		</div>';
 
+	// Validation info?
+	if (!empty($context['validation_tests']))
+	{
+		echo '
+		<div class="cat_bar">
+			<h3 class="catbg">Validation Results</h3>
```
SychO9: hard-coded string

Sources/Subs-Package.php
```diff
@@ -3141,4 +3141,120 @@ function smf_crc32($number)
...
+
+		// Sub out any variables we support in the validation url.
+		$validate_url = strtr($server['validation_url'], array(
+			'{SMF_VERSION}' => urlencode($the_version)
+		));
+
+		$results = fetch_web_data($validate_url, 'data=' . json_encode($sendData));
+
+		$parsed_data = $smcFunc['json_decode']($results, true);
+		if (is_array($parsed_data) && isset($parsed_data['data']) && is_array($parsed_data['data']))
```
SychO9, line 3247: to follow the same code style, I think you should add curly braces to the if block

Sources/Packages.php
```diff
@@ -201,6 +201,13 @@ function PackageInstallTest()
 	// Set the type of extraction...
 	$context['extract_type'] = isset($packageInfo['type']) ? $packageInfo['type'] : 'modification';
 
+	// Get our validation info.
+	$context['validation_tests'] = package_validate_installtest(array(
+		'file_name' => $packagesdir . '/' . $context['filename'],
+		'custom_id' => !empty($packageInfo['id']) ? $packageInfo['id'] : '',
+		'custom_type' => $context['extract_type']
+	), true);
```
SychO9: I think you forgot the `true` since the function only takes a single parameter.

SychO9 (commented):
I think it would look a little better visually if we changed the cat_bar to title_bar and added icons for each state, like this

package_hash_checking

```php
	$icons = array('package_validation_status_clean' => 'approve', 'package_validation_status_unknown' => 'check', 'package_validation_status_blacklist' => 'error');

	// Validation info?
	if (!empty($context['validation_tests']))
	{
		echo '
		<div class="title_bar">
			<h3 class="titlebg">Validation Results</h3>
		</div>
		<div id="package_validation">
			<table class="table_grid">';

		foreach ($context['validation_tests'] as $id_server => $result)
		{
			echo '
			<tr class="windowbg">
				<td>', $context['package_servers'][$id_server]['name'], '</td>
				<td><span class="main_icons ', $icons[$result[$context['package_sha256_hash']]], '"></span> ', $txt[isset($result[$context['package_sha256_hash']]) ? $result[$context['package_sha256_hash']] : 'package_validation_status_unknown'], '</td>
			</tr>';
		}

		echo '
			</table>
		</div>
		<br>';
	}
```

Sources/Subs-Package.php
```diff
@@ -3141,4 +3141,122 @@ function smf_crc32($number)
...
+		);
+	}
+
+	return package_validate_send($sendData);
+}
+
+/**
+ * Sending data off to validate packages.
+ *
+ * @param string $sendData Json encoded data to be sent to the validation servers.
```
live627: type mismatch; you pass an array to this param

SychO9 (approved):

MissAllSunday (approved):

PR #5963, opened 2020-01-19
EOL to LF, even on Windows

sbulen:
After some research & experimentation, this is what keeps EOL at LF, in git, on WAMP systems.  Upon commit, CRLFs, if found, are converted to LF.  On checkout, LFs are preserved.  

Without this change, all files are converted to CRLF upon checkout in Windows.  This can impact testing & package builds - wherever code is expecting an LF but finds a CRLF instead.  

This does not take effect immediately.  Git only applies the EOL processing on checkout/commit.  Also, if no changes are detected on a file, it leaves the file alone.  To force everything in your working directory to be converted to LF after this change, there are some very complicated procedures out there...  I took a much simpler, brute force approach: I deleted all files & folders (except the hidden .git folder, where the repo is...), then issued a:
`git checkout release-2.1 --force`
Boom.  Git was forced to rebuild all the files in the working directory from the repo, honoring the new setting, and all EOLs became LFs, everywhere.  

Input welcome.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,5 @@
 # Force our line endings to be LF, even for Windows
-* text
+* text eol=lf
 
 # Set certain files to be binary
 *.png binary


Sesquipedalian:
I do all my work on *nix machines, so I'll leave it to others to test this. 🙂

jdarwood007:
I use my mac for personal stuff, so I can't test this, but its sound from the documentation.

albertlast:
In my eyes this doesn't hurt,
so no reason to do any big test here...

Sesquipedalian:
I’ll take that as an approval.

PR #5964, opened 2020-01-20
Patch 4

Underdog-01:
The hook needs to be called directly after the array is populated.
This is because the default logic that follows, changes the subaction to "browse" if it is not in the array.


Underdog-01:
BTW this PR came to my attention because I just updated one of my mods to use that hook.
It did not execute its functions until I made this edit.
It highlighted my subaction but displayed the mail  ("browse") queue template.

Sesquipedalian (approved):

PR #5969, opened 2020-01-23
Cache

live627:
This unifies much of the behavior of the various cache methods. Also fixes version detection for memcached from the admin panel (credits section) because the object wasn't initialized and needs at least one server added. 

[`$memcached->getVersion()`](https://www.php.net/manual/en/memcached.getversion.php) returns an array of servers, the keys being the servers and the values are versions.  The code expects a string, so running the array through `current()` made it hunky dory again.

The previsions code would also only return a version string for memcached if it was selected as the current running cache.

### DIFF
--- a/Sources/CacheAPI-postgres.php
+++ b/Sources/CacheAPI-postgres.php
@@ -21,18 +21,19 @@ if (!defined('SMF'))
  */
 class postgres_cache extends cache_api
 {
-	/**
-	 * @var false|resource of the pg_prepare from get_data.
-	 */
-	private $pg_get_data_prep;
+	/** @var string */
+	private $db_prefix;
 
-	/**
-	 * @var false|resource of the pg_prepare from put_data.
-	 */
-	private $pg_put_data_prep;
+	/** @var resource result of pg_connect. */
+	private $db_connection;
 
 	public function __construct()
 	{
+		global $db_prefix, $db_connection;
+
+		$this->db_prefix = $db_prefix;
+		$this->db_connection = $db_connection;
+
 		parent::__construct();
 	}
 
@@ -59,12 +60,12 @@ class postgres_cache extends cache_api
 	 */
 	public function isSupported($test = false)
 	{
-		global $smcFunc, $db_connection;
+		global $smcFunc;
 
 		if ($smcFunc['db_title'] !== POSTGRE_TITLE)
 			return false;
 
-		$result = pg_query($db_connection, 'SHOW server_version_num');
+		$result = pg_query($this->db_connection, 'SHOW server_version_num');
 		$res = pg_fetch_assoc($result);
 
 		if ($res['server_version_num'] < 90500)
@@ -182,9 +183,7 @@ class postgres_cache extends cache_api
 	 */
 	private function createTempTable()
 	{
-		global $db_connection, $db_prefix;
-
-		pg_query($db_connection, 'CREATE LOCAL TEMP TABLE IF NOT EXISTS ' . $db_prefix . 'cache_tmp AS SELECT * FROM ' . $db_prefix . 'cache WHERE ttl >= ' . time());
+		pg_query($this->db_connection, 'CREATE LOCAL TEMP TABLE IF NOT EXISTS ' . $this->db_prefix . 'cache_tmp AS SELECT * FROM ' . $this->db_prefix . 'cache WHERE ttl >= ' . time());
 	}
 
 	/**
@@ -194,9 +193,7 @@ class postgres_cache extends cache_api
 	 */
 	private function deleteTempTable()
 	{
-		global $db_connection, $db_prefix;
-
-		pg_query($db_connection, 'DROP TABLE IF EXISTS ' . $db_prefix . 'cache_tmp');
+		pg_query($this->db_connection, 'DROP TABLE IF EXISTS ' . $this->db_prefix . 'cache_tmp');
 	}
 
 	/**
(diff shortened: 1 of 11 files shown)


Sources/CacheAPI-postgres.php
```diff
@@ -41,30 +42,67 @@ public function __construct()
...
+			);
+			$this->prepareQuery(
+				'smf_cache_put_data',
+				'INSERT INTO ' . $this->db_prefix . 'cache(key,value,ttl) VALUES($1,$2,$3)
+				ON CONFLICT(key) DO UPDATE SET value = $2, ttl = $3'
+			);
+			$this->prepareQuery(
+				'smf_cache_delete_data',
+				'DELETE FROM ' . $this->db_prefix . 'cache WHERE key = $1'
+			);
```
albertlast: You only prepare the queries when the cache table didn't exists --> doesn't work,
you need to create everytime when they didn't exists.

Sources/CacheAPI-postgres.php
```diff
@@ -108,49 +131,22 @@ public function getData($key, $ttl = null)
...
-			}
-		}
+		$ttl = time() + (int) ($ttl !== null ? $ttl : $this->ttl);
 
-		$result = pg_execute($db_connection, 'smf_cache_put_data', array($key, $value, $ttl));
-
-		if (pg_affected_rows($result) > 0)
-			return true;
+		if ($value === null)
+			$result = pg_execute($this->db_connection, 'smf_cache_delete_data', array($key));
```
albertlast, line 143: Deleting of single rows cost to mutch time and doesn't provide any benefits,
deleting should keep a task of housekeeping.

Sources/CacheAPI-postgres.php
```diff
@@ -160,9 +156,7 @@ public function cleanCache($type = '')
 	 */
 	public function getVersion()
 	{
-		global $smcFunc;
-
-		return $smcFunc['db_server_info']();
+		return pg_version()['server'];
```
albertlast, line 165: Why the use of an own way to retrive the Version?

Sources/CacheAPI-postgres.php
```diff
@@ -41,30 +42,67 @@ public function __construct()
...
+	 * @param string $stmtname
+	 * @param string $query
+	 */
+	private function prepareQuery($stmtname, $query)
+	{
+		$result = pg_query_params(
+			$this->db_connection,
+			'SELECT name FROM pg_prepared_statements WHERE name = $1',
+			array($stmtname)
+		);
```
albertlast: Time waste in my eyes,
running unneeded select of instead of try to pre when it failed it's fine when not it's fine too.

albertlast (changes requested):
To many changes in my es in postgres cache with many possiblites to break stuff and get slower,
only for the idea to look "nice".

I'm not sure if the postgres cache is running well with persistent connection.

Sources/CacheAPI-postgres.php
live627: Show me a better way to check if the resource is allocated.

Sources/CacheAPI-postgres.php
live627: The SMF cache system is special in that it expects a null value to trigger a delete.

Sources/CacheAPI-postgres.php
live627: Consistency. Less faffing with globals.

Sources/CacheAPI-postgres.php
albertlast: Where is that used?
And why is not a negative ttl value used?

Sources/CacheAPI-postgres.php
albertlast: There is no check needed, 
When the prepare run well,
Than came up with no error, 
When already exists he can up with an error which needs to be ignored.

albertlast:
I'm still waiting for answer of my question and why so many selects are needed

live627:
so that we get no errors when prepared statements are allocated

albertlast:
This can only happen when you first use postgres secondly when you use persists connection.
Very rare case of an rare case and then it's enough to check the first query.

Since you didn't handle the other cases where prepare statements can create error is this construct only slow but not good.

Sources/CacheAPI-postgres.php
Sesquipedalian: If I understand you correctly, @albertlast, you mean that instead of running this:
```
$this->prepareQuery(
 	'smf_cache_delete_data',
 	'DELETE FROM ' . $this->db_prefix . 'cache WHERE key = $1'
 );
```
... we should instead just run this:
```
@pg_prepare(
 	$this->db_connection,
 	'smf_cache_delete_data',
 	'DELETE FROM ' . $this->db_prefix . 'cache WHERE key = $1'
 );
```

Do I understand you correctly?

albertlast: Not realy,
to centralize the prepare logic is fine,
but the check in form of a select is waste of time.

Sources/CacheAPI-postgres.php
albertlast: To make my point clear here,
the delete looks fine,
but here we got two issue in my eyes:
This behavior is not descripte here https://github.com/SimpleMachines/SMF2.1/blob/73cee9a121eb1047518f4fa119588d893623dc11/Sources/Class-CacheAPI.php#L83-L92
Secondly i would expect that this behavior realized by a negativ ttl instead of a null value.

about null values vs  negative ttl ->
null value we use in smf for example in ip address logic so maybe i want to save this in a cache i can't when the null value always get deleted.
negative ttl is never a value from "common" smf it's special command from cache logic.

Sources/CacheAPI-postgres.php
albertlast: The version of https://github.com/SimpleMachines/SMF2.1/pull/5969/commits/19c2abb7675ef60c322546f06ab27f7c0df53672
to combine all selects in one and
prepare only this how are not exists is a ok solution,
but the check is still usless so long you don't had a persitent connection only in this case a prepare query can exists or did i miss something?

live627: So the prepared statements only last as long as the connection? Would it be safe to forego the query for regular non-persistent connections?

albertlast: I short yes,
https://www.postgresql.org/docs/current/sql-prepare.html :
```
Prepared statements only last for the duration of the current database session.
When the session ends, the prepared statement is forgotten,
so it must be recreated before being used again.
```
reason why in the existing code a persistent check/special logic exists.

jdarwood007 (approved):

PR #5971, opened 2020-01-25
Select box sizes should not be an array

live627:
This error manifests itself in the page for the cache settings


Sesquipedalian:
It looks like this has the positions of `$config_var[4]` and `0` in the ternary backwards.

live627:
In ManageServer, find

		array('cache_accelerator', $txt['cache_accelerator'], 'file', 'select', $detected),

and replace with

		array('cache_accelerator', $txt['cache_accelerator'], 'file', 'select', 44,

Sesquipedalian:
Never mind. I missed the `!` when I was looking at this before. :)

live627:
cheers

PR #5972, opened 2020-01-25
Deny SMF.org scraping for guests or non-admins

Underdog-01:
Gathering info from SMF.org for the admin section should be limited to admin access since this is only for admins to view
I had errors in my log showing guests accessing this info and when tested I was able to view this piped data as a guest.

### DIFF
--- a/Sources/Admin.php
+++ b/Sources/Admin.php
@@ -582,11 +582,11 @@ function AdminHome()
  */
 function DisplayAdminFile()
 {
-	global $context, $modSettings, $smcFunc;
+	global $context, $modSettings, $smcFunc, $user_info;
 
 	setMemoryLimit('32M');
 
-	if (empty($_REQUEST['filename']) || !is_string($_REQUEST['filename']))
+	if (empty($_REQUEST['filename']) || !is_string($_REQUEST['filename']) || empty($user_info['is_admin']))
 		fatal_lang_error('no_access', false);
 
 	// Strip off the forum cache part or we won't find it...
--- a/Sources/Admin.php
+++ b/Sources/Admin.php
@@ -962,4 +962,5 @@ function AdminEndSession()
 
 	redirectexit();
 }
+
 ?>
--- a/Sources/Admin.php
+++ b/Sources/Admin.php
@@ -962,5 +962,4 @@ function AdminEndSession()
 
 	redirectexit();
 }
-
 ?>


jdarwood007:
Your editor added a line to the end of the file.  That needs fixed before the PR is mergeable.

What errors did you see? Regardless of admin access or not, I can't see why any errors would be thrown.

Underdog-01:
> Could not retrieve the file https://www.simplemachines.org/smf/current-version.js?version=SMF+2.1+RC2.
https://web-develop.ca/index.php?topic=434.msg2525

That is one of the errors. I get several of them daily although from different URL queries. 
The only instance of retrieving those admin files is via that function.

Also only from guests... I can not duplicate the error when I enter the same URL.
Although I have been able to view the file by: 

> https://web-develop.ca/index.php?action=viewsmfile;filename=latest-news.js

... which Imo perhaps should not be allowed for guests or non-admins.

Underdog-01:
Btw - You'll have to test on your own site or test site.
I already added the changes to my site so the above no longer applies as it shows a fatal lang error for guests as it should.

jdarwood007:
That error simply means that your server couldn't reach our servers.  Check your servers DNS and firewall to ensure it can make a https connection to our servers.

Underdog-01:
It makes the connection just fine in the admin and when I use the URL prior to the changes I made.
However the errors still appear in the log on a daily basis.

The data is for admins only.
Even if not able to replicate the logged errors, is it not prudent to restrict the scraping for admins only since the data is only to be viewed in the admin section? 
Why should a guest be able to scrape SMF.org data via my website without purpose?

Oldiesmann:
That error is likely generated as the result of an automated task. In 2.1, the files aren't accessed directly every time you visit the admin center. Instead, they're fetched every so often (depending on how often the "Fetch Simple Machines Files" task runs) and stored in the database.

sbulen:
Correct, I think there are two different things going on here.  One is the error message, the other is the ability to kinda hijack your server to have it make sm.org do tasks.  

**1...**  Regarding the "could not retrieve the file" error message, I don't see that.  I just checked my 2.1 test environments & they always run cleanly fetching the files.  No errors.  My two guesses are (1) maybe it's been fixed already, or (2) some kind of environment/network issue in that test environment.  

**2...**  Regarding the "scraping" issue, I do think that's worth discussion.  Knowing the scheduled tasks, you can get *ANY* SMF forum to serve you up some sm.org info.  Examples:
https://www.whatsurbeef.net/index.php?action=viewsmfile;filename=latest-news.js
https://www.whatsurbeef.net/index.php?action=viewsmfile;filename=current-version.js
https://www.whatsurbeef.net/index.php?action=viewsmfile;filename=detailed-version.js
https://www.whatsurbeef.net/index.php?action=viewsmfile;filename=latest-versions.txt

On the one hand, these are all valid tasks/requests, and they are designed to work this way.  On the other hand, do we really want to be able to allow _anybody_ to trigger communications from _any_ SMF site to sm.org from _anywhere_ at will?

If we don't think there is a vulnerability here, no action is needed.  The question is - is there a real vulnerability here?

Part of me is concerned that this could be used in a bot attack against sm.org.  @jdarwood007 ???

Part of me is tempted to say, "Hey, we already have a REST API!!!" and call it a day...

Sesquipedalian:
Well, they're public files, and the ability to read them gives a guest no access to anything else, so it doesn't appear to a security risk for individual forums. 

However, it's potential as a DDOS vector against simplemachines.org sounds like a reason to change it.

Underdog-01:
I figured out the issue so I will close this PR.
When I updated PHP and enabled PHP-FPM my php.ini file location changed which I wasn't aware of at the time.
The config file I was adjusting from my control panel was the old one not being used.
What happened was, my ssl ca cert file settings for ssl & curl were not set to the proper path.
After adjusting 2 php variables in the ini file to the proper path, all is well now.
They were causing functions such as file_get_contents & curl to not work using domain names although ip's were working?! Whatever... problem is fixed now. 

Although perhaps something can be done in the future regarding non-members piping SMF content via any forum in circumstances where the data being sought should be a registered users or admin usage only. Just saying...    

Thanks guys & gals.

PR #5974, opened 2020-01-27
use the correct type in the gif to png converter

live627:


jdarwood007:
If we use the type hinting like this, we should ensure all calls to Subs-Graphics.php also call Class-Graphics.php first, otherwise I suspect we will have fatal errors.

live627:
Good point. I'll use `is_a()` instead.

This function is unused, btw.

Sesquipedalian (approved):

PR #5961, opened 2020-01-18
Improves updateSettingsFile() and adds some helper functions

Sesquipedalian:
This took longer to finish than I hoped, but it should be pretty much bulletproof now.

Fixes #5934
Fixes #5938
Fixes #5939
Fixes #5943

PR #5986, opened 2020-02-02
Partially fixes #5981

### LINKED ISSUE #5981: Upgrader - CLI errors
#### Description
There are several errors reported during an upgrade via CLI that are hidden when upgrading via the browser (e.g, eval'd code).  These errors does not happen on a 2.1=>2.1 upgrade, you must be upgrading from 1.x or 2.0.  

image

These reported errors do not seem to cause failures, but should be cleaned up nonetheless.

### Steps to reproduce
1. Upgrade, via CLI, from a valid 2.0 DB

- Database Type: mysql
- Database Version: 5.7
- PHP Version: 7.3

jdarwood007:
Does not yet fix the convert_utf8 template issue

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -4946,7 +4946,7 @@ function MySQLConvertOldIp($targetTable, $oldCol, $newCol, $limit = 50000, $setS
 		$smcFunc['db_free_result']($request);
 
 		// Special case, null ip could keep us in a loop.
-		if (is_null($arIp[0]))
+		if (!isset($arIp[0]) || is_null($arIp[0]) || $arIp[0] == '')
 			unset($arIp[0]);
 
 		if (empty($arIp))
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -4891,7 +4891,7 @@ function MySQLConvertOldIp($targetTable, $oldCol, $newCol, $limit = 50000, $setS
 {
 	global $smcFunc, $step_progress;
 
-	$current_substep = $_GET['substep'];
+	$current_substep = !isset($_GET['substep']) ? 0 : (int) $_GET['substep'];
 
 	if (empty($_GET['a']))
 		$_GET['a'] = 0;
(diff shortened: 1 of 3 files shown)


sbulen:
I'll do some tests of this once #5961 has been merged...

sbulen:
This is what I am seeing as of now:
image

Sesquipedalian:
The errors on line 4923 could be solved simply by changing `if (is_null($arIp[0]))` to `if (!isset($arIp[0]))`.

other/upgrade.php
```diff
@@ -4946,7 +4946,7 @@ function MySQLConvertOldIp($targetTable, $oldCol, $newCol, $limit = 50000, $setS
 		$smcFunc['db_free_result']($request);
 
 		// Special case, null ip could keep us in a loop.
-		if (is_null($arIp[0]))
+		if (!isset($arIp[0]) || is_null($arIp[0]) || $arIp[0] == '')
```
Sesquipedalian: `!isset($arIp[0]) || is_null($arIp[0])` is redundant.

Sesquipedalian (changes requested):

other/upgrade.php
jdarwood007: So was the other check.

sbulen:
If you merge this (& keep #5981 open) I'll look into the other issues.

other/upgrade.php
sbulen: This works as-is.   empty() oughtta work here, too, covers all the bases, I think...

jdarwood007:
I'm not sure why, but I'm thinking empty would create a issue.  My thoughts are thinking it has something to do with 0s but I can't recall exactly.

sbulen:
I've been testing this extensively along with all the other upgrader changes & it works just fine, I suggest merging it.

sbulen:
Since some of these issues impact current upgrades, I think we should merge this one asap.

MissAllSunday (approved):

PR #6004, opened 2020-02-06
Tries using $smcFunc['random_bytes'] in Subs-Password.php

Sesquipedalian:
This tries loading the random_compat library's `random_bytes()` function to generate a random salt in Subs-Password.php before falling back to other methods. It first tries using `$smcFunc['random_bytes']` to do the job. If that is unavailable, it attempts to load the library file directly. If both of those fail, it moves on to other options.

Also, in the last ditch attempt to generate a random salt, this no longer tries to call `$smcFunc['random_int']`. If that were going to work, then `$smcFunc['random_bytes']` already would have. Now it checks if PHP's native `random_int()` is available (in case it is but the native `random_bytes()` isn't for some strange reason), and then falls back to `mt_rand()` just like it did before we added the random_compat library. (I'd honestly prefer to trigger an error and return null at that point, but this seems more in line with the intended design of Subs-Password.php—and since we do include the random_compat library, it's a hypothetical question anyway.)

### DIFF
--- a/Sources/Subs-Password.php
+++ b/Sources/Subs-Password.php
@@ -38,7 +38,7 @@ namespace
 		 */
 		function password_hash($password, $algo, array $options = array())
 		{
-			global $smcFunc;
+			global $smcFunc, $sourcedir;
 
 			if (!function_exists('crypt'))
 			{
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
FYI, this is just a quick side issue that came up while I've been working on the PR to migrate all the 2.0.16/GDPR changes to 2.1.

PR #6005, opened 2020-02-07
Fix relabeling a pm with the same label in pgsql

live627:
apparently  pgsql needs to know the keys to successfully ignore duplicate inserts


live627:
Poor Travis... too many redundant builds.

jdarwood007:
Gave it a kick

PR #6009, opened 2020-02-10
Ports some error handling improvements from 2.0.16

Sesquipedalian:
This is part of porting over some recent changes in 2.0.x to 2.1.

### DIFF
--- a/SSI.php
+++ b/SSI.php
@@ -54,7 +54,7 @@ if (empty($cachedir) || !is_dir($cachedir) || !is_writable($cachedir))
 	}
 }
 
-$ssi_error_reporting = error_reporting(E_ALL);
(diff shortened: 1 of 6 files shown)


live627:
You sure hiding deprecation warnings  is a good idea?

Sesquipedalian:
In production, yeah.

Sesquipedalian:
I updated the logic in order to more intelligently choose whether to report deprecation notices. In SSI.php and index.php, deprecation notices are reported if $db_show_debug is true. In install.php, they are shown for pre-release versions of SMF but not release versions. In upgrade.php, they are shown if $db_show_debug is true or if SMF is a pre-release version. Does that look good to you, @live627?

live627:
Yes. That's quite neat, actually. Andi I now know of another PHP function.

PR #6010, opened 2020-02-10
Sets limits on PM rules and rates

Sesquipedalian:
This is part of porting over some recent changes in 2.0.x to 2.1.

### DIFF
--- a/Sources/PersonalMessage.php
+++ b/Sources/PersonalMessage.php
@@ -3733,6 +3733,12 @@ function ManageRules()
 {
 	global $txt, $context, $user_info, $scripturl, $smcFunc;
 
+	// Limit the Criteria and Actions to this.
+	$context['rule_limiters'] = array(
+		'criteria' => 10,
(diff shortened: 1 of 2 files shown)


MissAllSunday (approved):

Sources/PersonalMessage.php
```diff
@@ -3776,6 +3782,7 @@ function ManageRules()
 	if (isset($_GET['apply']))
 	{
 		checkSession('get');
+		spamProtection('pm');
```
jdarwood007, line 3785: Only note here is for 2.0 we did this because it wasn't feasible to apply a better fix.  This will trigger a warning if they send a PM then go apply rules right after.
Do we want to use a more clearly defined message/trigger here?

Sesquipedalian: > Do we want to use a more clearly defined message/trigger here?

Sounds lovely. How about we merge this, and then in a subsequent PR you can replace this with something better?

PR #6011, opened 2020-02-10
Checks MIME type of user-supplied avatar images

Sesquipedalian:
This implements a more robust version of check_mime_type() than in 2.0.16, and puts it in Subs.php instead of Subs-Package.php to accompany fetch_web_data().

This is part of porting over some recent changes in 2.0.x to 2.1. It is an alternative to #5957.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -5549,14 +5549,13 @@ function fetch_web_data($url, $post_data = '', $keep_alive = false, $redirection
 }
 
 /**
- * Checks whether a file or data has the expected MIME type.
+ * Attempts to determine the MIME type of some data or a file.
  *
  * @param string $data The data to check, or the path or URL of a file to check.
- * @param string $type_pattern A regex pattern to match the acceptable MIME types.
  * @param string $is_path If true, $data is a path or URL to a file.
- * @return int 1 if the detected MIME type matches the pattern, 0 if it doesn't, or 2 if we can't check.
+ * @return string|bool A MIME type, or false if we cannot determine it.
  */
-function check_mime_type($data, $type_pattern, $is_path = false)
+function get_mime_type($data, $is_path = false)
 {
 	global $cachedir;
 
@@ -5565,13 +5564,16 @@ function check_mime_type($data, $type_pattern, $is_path = false)
 
 	// Oh well. We tried.
 	if (!$finfo_loaded && !$exif_loaded)
-		return 2;
+		return false;
+
+	// Start with the 'empty' MIME type.
+	$mime_type = 'application/x-empty';
 
 	if ($finfo_loaded)
 	{
 		// Just some nice, simple data to analyze.
 		if (empty($is_path))
-			$mime_type = finfo_buffer(finfo_open(FILEINFO_MIME), $data);
+			$mime_type = finfo_buffer(finfo_open(FILEINFO_MIME_TYPE), $data);
 
 		// A file, or maybe a URL?
 		else
@@ -5582,11 +5584,7 @@ function check_mime_type($data, $type_pattern, $is_path = false)
 
 			// URL.
 			elseif ($data = fetch_web_data($data))
-				$mime_type = finfo_buffer(finfo_open(FILEINFO_MIME), $data);
-
-			// Non-existent files obviously don't have the right MIME type.
-			else
-				return 0;
+				$mime_type = finfo_buffer(finfo_open(FILEINFO_MIME_TYPE), $data);
 		}
 	}
 	// Workaround using Exif requires a local file.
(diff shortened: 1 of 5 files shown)


Sources/Subs.php
```diff
@@ -5548,6 +5548,89 @@ function fetch_web_data($url, $post_data = '', $keep_alive = false, $redirection
...
+	$finfo_loaded = extension_loaded('fileinfo');
+	$exif_loaded = extension_loaded('exif') && function_exists('image_type_to_mime_type');
+
+	// On Windows, the Fileinfo extension may not be enabled by default.
+	if (!$finfo_loaded)
+	{
+		// Maybe we can load it dynamically? (Probably not, but give it a shot.)
+		$safe = ini_get('safe_mode');
+		$dl_ok = ini_get('enable_dl');
+		if (empty($safe) && !empty($dl_ok) && function_exists('dl'))
```
live627: holup, dl() must never be used.

Sesquipedalian: If php.ini enabled it, why not try?

live627: > This function was removed from most SAPIs in PHP 5.3.0, and was removed from PHP-FPM in PHP 7.0.0. 

https://www.php.net/manual/en/function.dl.php

Best to avoid it.

Sesquipedalian: It’s best to avoid *relying* on dl(). This isn’t relying on it. This is merely giving it a shot before giving up and moving on.

MissAllSunday: I agree is best not to relying on it or giving it a shot, its old stuff and it will always be best if we don't have code for old stuff, even if its just to give it a shot.

Sesquipedalian: > it will always be best if we don't have code for old stuff, even if its just to give it a shot.

But what's the harm in keeping the option available? Even if it is an old function that usually isn't available on modern systems, it does still work on those older systems where it is available. Axioms like "it will always be best if..." are only true when they are actually true. In this case, it doesn't apply.

live627: Check the support forums. If what you say is true, then why do people complain that they can't upload avatars because of this?

Sesquipedalian: Because 2.0.17 didn't include https://github.com/SimpleMachines/SMF2.0/pull/129

MissAllSunday: The PHP group decided to deprecate and eventually remove that function, that alone should be reason enough to avoid using it.

Its behavior is OS dependent, meaning someone using MAC won't be able to make use of it even if it can be used.  To avoid that you need to introduce boilerplate code to check for the OS.

Sesquipedalian: Well, I work on a Mac, so I'm pretty confident saying that it works the same way on Macs as on any other *nix OS. 😉 The boilerplate code to handle Windows vs. *nix is already included on the next line after these comments.

But hey, if you folks are all still convinced that it is somehow a bad idea to use a tool that would actually work just fine in those situations where it is enabled, then fine. I remain unconvinced by the objections I've heard, but I don't always have to get my way. 🙂

MissAllSunday: By this logic we should also add support for any other features the PHP group has removed or deprecated.

Sesquipedalian: No, that's not what this logic implies. That's an all-or-nothing conclusion. The logic here entails making decisions based on the situational context. As such, this logic rejects the premise that these decisions should made using _a priori_, all-or-nothing rules. To conclude that this logic thereby implies an  _a priori_, all-or-nothing rule to include support for deprecated functions is a _non sequitur_.

MissAllSunday: Nope!  the logic is this bit:

> use a tool that would actually work just fine in those situations where it is enabled

In that context, we should add support for other deprecated features since, in fact, those features will still work on situations were those are enabled.  

Heres a situational context too:  the use of create_function(). create_function() (still valid in PHP 5.4) will still work just fine in those situations where it is enabled, since there is no harm in keeping it, no harm in adding boilerplate code for supporting it and we turn a blind eye on PHP deprecations, lets added it back to parse_bbc() :)

My point is, the mere concept of it been deprecated should be reason enough to not use it.

Ultimately, it's your call.

Sesquipedalian: 1. If there were a situation where create_function() solved a problem that a closure couldn't, then your example would be analogous. But there isn't, so no, that's not the same situation.
2. You persist in your all-or-nothing reasoning. Suppose Alice says, "Red cars should never be driven, because X, Y, and Z," and Bob replies, "I agree red cars should not be driven in general, but there can exceptions under certain conditions." If Alice then replies, "If you allow exceptions under certain conditions, Bob, then you are saying all red cars should be driven at all times," then Alice is committing the [composition/division logical fallacy](https://yourlogicalfallacyis.com/composition-division), as well as the [strawman fallacy](https://yourlogicalfallacyis.com/strawman) since she is misrepresenting Bob's actual argument.

Thus, where you say, "the mere concept of it been deprecated should be reason enough to not use it", I say, "In general that is true, but there can be exceptions under specific conditions." Those specific conditions are when all of the following are true:

1. We are not relying on the deprecated function, but merely keeping the option open as a possible fallback.
2. No better alternative exists that can do the job of the deprecated function.
3. No harm will come from attempting to use the deprecated function.

Regardless, this conversation has become pointless: c7c8461.

PR #6012, opened 2020-02-10
Implements authentication secret; improves cookie hashing; uses unsubscribe tokens in email notifications

### LINKED ISSUE #5106: Need to provide unsubscribe links for announcement emails
To comply with the GDPR, we'll need to make it possible for people to unsubscribe from announcement emails without logging in.

I've already written the code to do this in SMF 2.0, so this issue is just here to remind me to port it over to 2.1.

### LINKED ISSUE #5480: Incorrect unsubscribe link for topic notifications
#### Description
See https://www.simplemachines.org/community/index.php?topic=565371.0

I'm not sure if we can provide a valid link, because the "notifytopic" action requires a valid session ID.

### LINKED ISSUE #5923: 2.0.16 Changes Needs Merged
Just a note, prior to RC3, we need to merge the changes from 2.0.16 into here.

I was going to start working on it, but if anyone has already started, please speak up.

Sesquipedalian:
This implements the authentication secret infrastructure introduced in 2.0.16 and uses it (1) for cookie hashing and (2) for generating unsubscribe tokens in email notifications in order to allow unsubscribing without logging in, which is required for GDPR compliance.

In connection with the second use, this also removes the `allow_disableAnnounce` setting and always gives the user the option to disable announcements. In 2.0.16 we added conditional checks in a bunch of places in order to override the `allow_disableAnnounce` setting when GDPR compliance mode was enabled, but that was messy and complicated. Simply removing the setting is much cleaner and saner.

Fixes #5106
Fixes #5480
Fixes #5923 (although technically it will take number of separate PRs to do that, but this one does more than others.)


Sources/Subs-Auth.php
```diff
@@ -884,15 +884,15 @@ function hash_password($username, $password, $cost = null)
...
+ * Hashes password with salt and authentication secret. This is solely used for cookies.
  *
  * @param string $password The password
  * @param string $salt The salt
  * @return string The hashed password
  */
 function hash_salt($password, $salt)
 {
-	return hash('sha512', $password . $salt);
+	return hash_hmac('sha1', sha1($password . $salt), get_auth_secret());
```
albertlast: Why we switch from sha2 to sha1, how got already first issues:
```
Additionally, SHA1 has also been deemed quite vulnerable to collision attacks
which is why all browsers will be removing support for certificates signed with SHA1 by January 2017.
SHA256 however, is currently much more resistant to collision attacks
as it is able to generate a longer hash which is harder to break.
```
https://www.keycdn.com/support/sha1-vs-sha256

jdarwood007:
Needs to go back to sha2.  The hmac should also use sha2.

Sources/Subs-Auth.php
Sesquipedalian: For simple hashes, yes, SHA1 has been broken. But this is using `hash_hmac()`, not `hash()`. [HMAC](https://en.wikipedia.org/wiki/HMAC) is a fundamentally different kind of algorithm than a simple hash, and it has not been broken. It is still 100% safe to use even MD5 in an HMAC.

Unfortunately, we cannot simply replace `hash()` with `hash_hmac()` everywhere in our code, because an HMAC is only suitable for certain situations—specifically, message authentication. But authenticating cookies is one of those situations, so switching from `hash()` to `hash_hmac()` is a good idea no matter what. 

In SMF 2.0.16 and higher, we are using an SHA1 HMAC for this task. So by using the same here, we will allow cookies set by 2.0.16+ to be accepted by 2.1 just as they are. That means that upgrading from 2.0.x to 2.1 will not require all users of a forum to log in again, which is better from a user perspective.

Sesquipedalian:
> Needs to go back to sha2. The hmac should also use sha2.

Why? Even even the authors of the "SHA-1 is a Shambles" document acknowledge that there is no threat against HMACs using SHA1. If that were not true, then the changes we made in 2.0.16 would not have solved the security issue that 2.0.16 did in fact solve.

jdarwood007:
Its still moving the main hashing back to sha1, which is what we need to be avoiding.

If we up the min PHP version 5.5, we could even move to password_hash using bcrypt.  Then use the hmac for the cookie.

Using a sha2 with a sha2 hmac requires less changes here.

Sesquipedalian:
No, the `sha1($password . $salt)` in this code isn't the main hashing. It is is merely a (pointless) extra step that lets us maintain backwards compatibility with SMF 2.0.16+. The main hashing is the `hash_hmac()` function. That's what actually does the work, and it is unaffected by that pointless initial hashing.

Of course, if we decide that there's no value in maintaining backward compatibility with the cookies from SMF 2.0.x, then sure, we can use a different algorithm for the HMAC.  At that point, though, we might as well just throw away the entire `hash_salt()` function and simply use `hash_hmac()` directly. That will make our code simpler overall, but in terms security it won't gain us anything in exchange for giving up that backwards compatibility.

(Edited after I re-read the original version and realized it sounded rather snarky, which was not intended.)

jdarwood007:
Well my opinion would be that breaking cookies with a 2.1 upgrade would be acceptable.  But at this point I've stated my opinions.  Others should chime in.  We could let a upgrade path enable old cookies.  The admin could decide to turn it off.

albertlast:
i would stick to this point: https://crypto.stackexchange.com/a/27530

> I would use HMAC-SHA256.
> 
> While poncho's answer that both are secure is reasonable, there are several reasons I would prefer to use SHA-256 as the hash:
>   - Attacks only get better. SHA-1 collision resistance is already broken, so it's not impossible that other attacks will also be possible in the future.
>    -  It allows you to depend on just one hash function, which you can also use in signature algorithms etc. where collision resistance is required.
>    -  You don't have to justify using a broken primitive. ("I heard SHA-1 is broken?")
> 
> The only potential downside is performance, but it's probably not significant in most HMAC use cases, since where performance is important you can find even faster MACs to use instead. So I would consider the above reasons sufficient in most cases to justify using SHA-256 as the HMAC hash.

Oldiesmann:
I see no problems with breaking cookies for 2.1 in the name of security. We've forced people to log in again other times for security purposes as well (when we've changed password hashing methods at least).

Sesquipedalian:
Very well then.

Sesquipedalian:
If we're going to do it, we may as well go all in. So:
1. We now use sha3-512 on systems that support it, falling back to sha512.
2. We now use `hash_pbkdf2()` to derive a unique secret key for each user based on the master authentication secret and the user's unique salt.
3. We then use that derived key to generate the HMAC of the password to include in the cookie.
4. We now use `hash_equals()` to validate the hash supplied in the user's cookie, instead of merely checking whether the strings are equal using `==`.

Sesquipedalian:
I've also bumped the length of the user's password_salt up from 16 bits (represented as 4 hexadecimal characters) to 128 bits (represented as 32 hexadecimal characters). This is to match NIST's recommended salt length for PBKDF2 key derivation.

Sources/Load.php
```diff
@@ -528,9 +529,9 @@ function loadUserSettings()
 			// As much as the password should be right, we can assume the integration set things up.
 			if (!empty($already_verified) && $already_verified === true)
 				$check = true;
-			// SHA-512 hash should be 128 characters long.
-			elseif (strlen($password) == 128)
-				$check = hash_salt($user_settings['passwd'], $user_settings['password_salt']) == $password;
+			// Password hash should be 40 characters long.
```
jdarwood007: With Sha3-512 and sha512, are those not 128 lengths?  Or am I missing something else that trims them back to 40.. Does the hmac do that?

Sources/Subs-Admin.php
```diff
@@ -529,6 +529,18 @@ function updateSettingsFile($config_vars, $keep_quotes = null, $rebuild = false)
 			'default' => 'SMFCookie11',
 			'type' => 'string',
 		),
+		'auth_secret' => array(
+			'text' => implode("\n", array(
+				'/**',
+				' * Secret key used to create and verify cookies, tokens, etc.',
+				' * Do not change this, and NEVER share it.',
+				' *',
```
jdarwood007, line 541: Wait, shouldn't this say something along the lines of changing this will instantly log out all users?  It should be changeable in the event the forum was breached to invalidate all cookies.

Sources/Subs-Auth.php
```diff
@@ -943,4 +950,60 @@ function hash_benchmark($hashTime = 0.2)
...
+
+		if (!$raw_output)
+			$output = bin2hex($output);
+		return $length ? substr($output, 0, $length) : $output;
+	}
+}
+
+if (!function_exists('hash_equals'))
+{
+	function hash_equals($known_string, $user_string)
```
jdarwood007, line 954: I see this came from the php.net/hash_equals user comments.  Do we need to credit that or is that free use?

jdarwood007 (commented):
Thanks for the changes.  I understand sha1 under hmac still may be secure, but they are showing signs of weakness in other forms (non hmac) that just says for the life of this version of the software, we should ensure any weak hashing is eliminated.  The CPU cost nowadays is fairly minimal.

Sources/Load.php
Sesquipedalian: Good catch. Will fix shortly.

Sources/Subs-Admin.php
Sesquipedalian: Yeah, that'd be better. I'll change it.

Sources/Subs-Auth.php
Sesquipedalian: Good question. I'll check.

Sesquipedalian: According to the comment submission page, "any note submitted here becomes the property of the PHP Documentation Group and will be available under the same license as the documentation." The PHP Documentation is licensed using the Creative Commons Attribution 3.0 License. So I guess the answer is yes, we do need to. I'll add a comment doing so.

Sesquipedalian:
Note: Travis build is currently failing due to a problem with Travis itself.

PR #6013, opened 2020-02-11
Refactor Cache API

### LINKED ISSUE #6110: Memcache errors
#### Description
#5969 I didn't notice that it introduced an error.
We should not call ->connect() during the loadCacheAPIs() call.  We should only call the isSupported(true) as we are checking that its supported and we are in test mode.  Test mode indicates we are not going to make any connections.  Outside of test mode we are attempting to use the caching engine and then may need to do more sanity checks.  On connect we finally attempt to connect to the caching engine.

@live627 Your thoughts here?

2: Memcache::connect(): Can't connect to :11211, php_network_getaddresses: getaddrinfo failed: Name or service not known (0)
8: Memcache::connect(): Server  (tcp 11211, udp 0) failed with: php_network_getaddresses: getaddrinfo failed: Name or service not known (0)

/Sources/CacheAPI-memcache.php (Line 76)

MissAllSunday:
This PR refactors the current cache API

- separates the abstract and the interface into their own classes
- namespace usage
- each implementation now has extend the basic abstract class and implement the cache interface otherwise SMF won't see it as a valid option.
- cleaner loadCacheAPIs() function, it does not depends on regex to get the correct class 
- cleaner structure, each API implementation goes into the Sources/cache/APIs folder
- cleans up Sources folder a bit.
- support for upgrades, it attempts to know the current cache option been used and change $cache_accelerator accordingly.
- $cache_accelerator now holds the cache class name.
Tested on Mac Mojave and Ubuntu 18.04.4 using Sqlite, File based and Memcached options.

Fixes #6110


Sources/ManageServer.php
```diff
@@ -1544,28 +1544,28 @@ function ShowPHPinfoSettings()
...
+	require_once($cacheAPIdir . '/CacheAPI.php');
 
 	$apis = array();
-	if ($dh = opendir($sourcedir))
+	if ($dh = opendir($cacheAPIImplementationdir))
 	{
 		while (($file = readdir($dh)) !== false)
 		{
-			if (is_file($sourcedir . '/' . $file) && preg_match('~^CacheAPI-([A-Za-z\d_]+)\.php$~', $file, $matches))
+			if (is_file($sourcedir . '/' . $file))
```
jdarwood007: Wrong variable on is_file here.

Sources/ManageServer.php
```diff
@@ -1544,28 +1544,28 @@ function ShowPHPinfoSettings()
...
 			{
-				$tryCache = strtolower($matches[1]);
-
 				require_once($sourcedir . '/' . $file);
-				$cache_class_name = $tryCache . '_cache';
-				$testAPI = new $cache_class_name();
+
+				$testAPI = new $file();
```
jdarwood007: How is this to work if $file is a file name and your new is trying to initialize an object?

Sources/cache/CacheApiInterface.php
```diff
@@ -0,0 +1,67 @@
...
+interface CacheApiInterface
+{
+	/**
+	 * Checks whether we can use the cache method performed by this API.
+	 *
+	 * @access public
+	 * @param boolean $test Test if this is supported or enabled.
+	 * @return boolean Whether or not the cache is supported
+	 */
+	public function isSupported($test = false);
```
jdarwood007: More of a thought here, but we do type casting in other places to be compliant with other parts of SMF.  We should make the Cache API use type casting to the full extent.

other/Settings.php
```diff
@@ -169,6 +169,12 @@
  * @var array
  */
 $cache_memcached = '';
+/**
+ * This is only used for redis. Should be a string of 'server:port'
+ *
+ * @var string
+ */
+$cache_redis = '';
```
jdarwood007: Just a thought, would it be better to use a generic variable here for all additional caching needs or leave it like this with additional settings.php variables?
The only downside to doing that is it prevents the possibility of using multiple different caching mechanisms at the same time.

other/Settings.php
```diff
@@ -225,6 +231,20 @@
...
+ */
+$cacheAPIdir = $sourcedir . '/cache';
+
+/**
+ * Path to the cache implementations.
+ *
+ * @var string
+ */
+$cacheAPIImplementationdir = $cacheAPIdir . '/implementations';
+
```
jdarwood007: Overall, do we need to build a $cacheAPIdir and $cacheAPIImplementationdir  for any reason?  Why not just build off $sourcedir?  I personally do not believe we need to do that, same with $tasksdir but I wasn't around for that implantation.  Just adds complexity to the code.

jdarwood007 (changes requested):

other/Settings.php
MissAllSunday: I'm not a fan of using $sourcedir .'/harcoded_dir/hardcoded_dir';

It's easier to reference an already build global variable than building your own. Mod authors will end up building their own anyway, might as well provide a uniform and consistent across modifications variable ;)

other/Settings.php
MissAllSunday: I thought about that as well, even moving the Redis specific parameters to a regular $modSettings key.

Will do the Redis implementation on another PR to keep things simpler but will most likely just use a regular $modSettings var instead.

Sources/cache/CacheApiInterface.php
MissAllSunday: mmm you mean type hinting?  or what do you mean by type casting?

jdarwood007: Yea I meant type hints.  Nothing more than a comment we should clean it up to go full type hinting if we are refactoring it to ensure proper compatibility with the API and no assumptions about how a method works or takes input.

I believe our min is still in PHP 5.x, which doesn't support return type declaration.  So we can't implant those unless we are moving the min to PHP 7.

MissAllSunday: yep unfortunately we can't use them :( this is as far as we can enforce the interface.

other/Settings.php
MissAllSunday: Went with a half way approach, kept cacheAPIdir, deleted the other one.

other/Settings.php
MissAllSunday: Will use a normal $modSettings var for this and will implement it on another PR.

Sesquipedalian:
Note: Travis build is currently failing due to a problem with Travis itself.

Sesquipedalian:
To fix Travis builds, please merge release-2.1 into this branch.

MissAllSunday:
This is now considered done. Please test it specially with the other cache options.

Themes/default/languages/ManageSettings.english.php
```diff
@@ -127,21 +127,30 @@
...
 $txt['cache_sqlite_settings'] = 'SQLite3 database caching settings';
-$txt['cache_memcache_settings'] = 'Memcache(d) settings';
-$txt['cache_memcache_servers'] = 'Memcache(d) servers';
-$txt['cache_memcache_servers_subtext'] = 'Example: 127.0.0.1:11211,127.0.0.2';
+$txt['cache_memcacheimplementation_settings'] = 'Memcached settings';
+$txt['cache_memcacheimplementation_servers'] = 'Memcached servers';
+$txt['cache_memcacheimplementation_servers_subtext'] = '<br>
+<div class="smalltext">
+Example: 127.0.0.1:11211,127.0.0.2
+</div>';
```
jdarwood007: We should avoid HTML as much as possible.  Especially this as its layout and not hyperlinks/formatting.

Sources/Load.php
```diff
@@ -3410,60 +3413,84 @@ function loadDatabase()
...
+	$cache_file_info = !empty($overrideCache) ? $overrideCache : (!empty($cache_accelerator) ?
+		smf_json_decode($cache_accelerator, true) : array(
+			'class_name' => CacheApi::APIS_DEFAULT,
+			'file_info' => array(
+				'basename' => CacheApi::APIS_DEFAULT .'.php',
+			),
+			'txt_key' => strtolower(CacheApi::APIS_DEFAULT),
+		));
+
+	$file_to_load = $apis_dir . '/' . $cache_file_info['file_info']['basename'];
```
live627: Not a fan of storing arrays in Settings.php. IMO short strings work the best  as  they are quick and easy for intermediate/advanced forum owner to change. (Also devs)

Themes/default/languages/ManageSettings.english.php
```diff
@@ -127,21 +127,30 @@
...
+$txt['filebased_cache'] = 'SMF file based caching';
 $txt['sqlite_cache'] = 'SQLite3 database based caching';
 $txt['postgres_cache'] = 'PostgreSQL caching';
 $txt['cachedir_sqlite'] = 'SQLite3 database cache directory';
 $txt['apc_cache'] = 'APC';
 $txt['apcu_cache'] = 'APCu';
-$txt['memcache_cache'] = 'Memcache';
-$txt['memcached_cache'] = 'Memcached';
+$txt['memcacheimplementation_cache'] = 'Memcache';
+$txt['memcachedimplementation_cache'] = 'Memcached';
```
live627, line 135: is this an error?

MissAllSunday: Nope, memcache and memcached are now separated, thus they have their own txt strings even if the only thing that changes is adding a "d".

The "Implementation" part is just being safe as to not be confused with the real Memcache and Memcached classes since we can't really fully use namespaces.

Sources/Load.php
MissAllSunday: But that's one of the key features of json strings, they are somehow easier to read than a serialized one.

Themes/default/languages/ManageSettings.english.php
MissAllSunday: To be able to do that we will have to re-do the cache admin page, just a few lines above this one there's a nice (and outdated) ul list for cache systems:

```php
$txt['caching_information'] = 'SMF supports caching through the use of accelerators. The currently supported accelerators include:
<ul class="normallist">
	<li>APC</li>
	<li>APCu</li>
	<li>Memcached</li>
	<li>SQLite3</li>
	<li>PostgreSQL</li>
	<li>Zend Platform/Performance Suite (Not Zend Optimizer)</li>
	<li>XCache</li>
</ul>
Caching will work best if you have PHP compiled with one of the above optimizers, or have memcached available. If you do not have any optimizer installed SMF will do file based caching.';
```

Sources/Load.php
live627: I see you quickly put it  back so I'm not going to continue this argument. 😃

MissAllSunday: mmm not sure what you mean, I didn't modify this part of the code on recent commits. It still stores a json encoded array.

live627: oh... when you pushed d80dacbcb1569fea41cea1ae03aa6d336dd8491e, several commits got lost.

MissAllSunday: Yeah I use rebase instead of merging, thats why old commits appears to be new ones, in a sense they are new (to the release-2.1 branch) but originally they are old.

MissAllSunday: Oh my!  you are right a lot of commits got lost 😱 will check it out why.

MissAllSunday:
It seems GitHub ate a lot of commits, will investigate why or try to bring them back.

MissAllSunday:
Found the issue, GitHub for some reason took the same branch as a different one, merged both.

live627:
IMO any major refactoring should be postponed for the next major version.

MissAllSunday:
I agree. However, this isn't a major refactor. The abstract class and interface already exists, there is no change in how the cache apis work and are implemented (besides how the current cache info is stored). 

This changes makes sure anyone wanting to add another cache engine should follow our structure and makes it easier to add more engines.

live627:
uhh. okay then...

Sources/Load.php
```diff
@@ -3507,60 +3510,78 @@ function loadDatabase()
 /**
  * Try to load up a supported caching method. This is saved in $cacheAPI if we are not overriding it.
  *
- * @param string $overrideCache Try to use a different cache method other than that defined in $cache_accelerator.
+ * @param array $overrideCache Try to use a different cache method other than that defined in $cache_accelerator.
  * @param bool $fallbackSMF Use the default SMF method if the accelerator fails.
  * @return object|false A object of $cacheAPI, or False on failure.
  */
-function loadCacheAccelerator($overrideCache = null, $fallbackSMF = true)
+function loadCacheAccelerator($overrideCache = [], $fallbackSMF = true)
```
DiegoAndresCortes: Bracket on parameter is ok?

Sources/Cache/CacheApi.php
```diff
@@ -193,7 +50,7 @@ abstract class cache_api implements cache_api_interface
 	 */
 	public function __construct()
 	{
-		$this->setPrefix();
+		$this->setPrefix('');
```
DiegoAndresCortes: Is it necessary to send a parameter?

Sources/Load.php
MissAllSunday: nope, will change it.

Sources/Cache/CacheApi.php
MissAllSunday: nope, was added by the linting script, will remove.

MissAllSunday:
Bump, we still need a proper API implementation.  $cache_accelerator still holds a plain string and a few fixes to the current cache implementation are also included.

live627:
> $cache_accelerator still holds a plain string

Can we keep that a string for 2.1 and revisit in 3.0?

MissAllSunday:
It no longer uses a json string. $cache_accelerator is no longer affected.

other/upgrade-helper.php
```diff
@@ -457,3 +458,27 @@ function array_column($input, $column_key, $index_key = null)
 		return $arr;
 	}
 }
+
+/**
+ * Creates the json_encoded array for the current cache option.
+ *
+ * @return string a json_encoded array with the selected API options
```
jdarwood007, line 465: Where is the JSON?

MissAllSunday: There is not json entry anymore, it still is a plain string I just forgot to update the dock block

PR #6024, opened 2020-02-22
Notify

### LINKED ISSUE #6022: Installer warning - Cookie names must not be empty
#### Description
Warning
D:\EasyPHP\van21\Sources\Subs-Auth.php
on line
862

image

~~The DB files are not created or populated.  You cannot load the forum, you get:~~
(EDIT: the files are in fact created...)

### Steps to reproduce
1. Install vanilla 2.1 

- Version/Git revision:  Now
- Database Type:  Mysql
- Database Version:  5.7
- PHP Version:  7.3

live627:
Breaks out the user info loading routine so that it may be useful to the background task. Fixes #6022 
Also standardize the content links for the alerts.

(Most of these changes are whitespace so I can better see what I'm doing)

ref https://www.simplemachines.org/community/index.php?topic=571923.0
ref https://www.simplemachines.org/community/index.php?topic=571858.0


Sources/Profile-View.php
```diff
@@ -363,23 +367,15 @@ function fetch_alerts($memID, $to_fetch = false, $limit = 0, $offset = 0, $with_
 		$format_type = str_replace('{scripturl}', $scripturl, $format_type);
 
 	// If we need to check board access, use the correct board access filter for the member in question.
+	$qb = array(
```
albertlast, line 370: Why note any more in else part and
why the array() syntax instead of $qb[]?

Sources/tasks/CreatePost-Notify.php
```diff
@@ -76,10 +76,7 @@ public function execute()
 		// Find the people interested in receiving notifications for this topic
 		$request = $smcFunc['db_query']('', '
 			SELECT
-				ln.id_member, ln.id_board, ln.id_topic, ln.sent,
-				mem.email_address, mem.lngfile, mem.pm_ignore_list,
-				mem.id_group, mem.id_post_group, mem.additional_groups,
-				mem.time_format, mem.time_offset, mem.timezone,
+				mem.*, ln.id_board, ln.id_topic, ln.sent,
```
albertlast, line 79: to use * is always bad: https://stackoverflow.com/questions/3639861/why-is-select-considered-harmful

Sources/Load.php
```diff
@@ -682,120 +794,73 @@ function loadUserSettings()
...
+	return $possibly_robot;
+}
+
+/**
+ * Returns an array suitable for populating `$user_info` with.
+ *
+ * @param int $id_member
+ * @param array $user_settings
+ *
+ * @return array
```
albertlast, line 839: funny note: we always define an array as return,
but you don't know what format of this array is XD

albertlast (changes requested):
select coment is curcial, the others are only notes.

live627:
As much as I really want to unify the alerts, I'm going to hold off on this and try it again when we start on the new version.

PR #6025, opened 2020-02-22
Cron user info

### LINKED ISSUE #5868: cron.php doesn't initialize smileys and attempts to use SERVER_PROTOCOL
#### Description
So I'm trying to use my crontab to run cron.php and ran into some issues:
```
Type of error: Cron
8: Undefined index: smiley_set
http://(redacted).de/cron.php
(redacted)/Sources/Subs.php (Line 3011( & 3024 & 3035 & 3037))
```
```
The database value you're trying to insert does not exist: smiley_set<newline>Function: parsesmileys
(redacted)/Sources/Subs.php (line 3263)
```
```
8: Undefined index: SERVER_PROTOCOL
(redacted)/Sources/Subs.php (line 7076)
```

Original threads (linking to Arantor's summaries near the ends of both threads):
* https://www.simplemachines.org/community/index.php?topic=569827.msg4036609#msg4036609 (problems caused by the normal javascript-based scheduled tasks)
* https://www.simplemachines.org/community/index.php?topic=570396.msg4036612#msg4036612 (crontab problems)

### Steps to reproduce
1. run 2.1 between rc2 and the GH version of 2019-10-10
2. patch [this](https://github.com/SimpleMachines/SMF2.1/pull/5222/commits/6e288d5f78e954aacbceb629ce3346d8a1b26cd9) postgres-related bug
3. enjoy the spam 😉 

- Version/Git revision: GH version of 2019-10-10
- Database Type: Postgres
- Database Version: 9.6ish
- PHP Version: 5.6ish

MissAllSunday:
Fixes #5868

Provides a new method for tasks to get info from users and use $user_info safely within context, as extra safe measure a copy of $user_info is saved on instance and revert it back when the class no longer is used.

Also sets a dummy default value for SERVER_PROTOCOL when using cron.php on CLI


Sources/Load.php
```diff
@@ -849,6 +850,71 @@ function loadUserSettings()
...
+	$user_settings_min = array();
+
+	if (empty($userId) || !is_int($userId))
+		return [];
+
+	if (empty($cache_enable) || ($user_settings_min = cache_get_data('user_settings_min-' . $user_id, 120)) == null)
+	{
+		$request = $smcFunc['db_query']('', '
+				SELECT time_offset, additional_groups, id_group, id_post_group, lngfile, smiley_set, time_offset
+				FROM {db_prefix}members
```
albertlast: Would like to see that we use here the mem allias

Sources/Load.php
```diff
@@ -849,6 +850,71 @@ function loadUserSettings()
...
+function loadMinUserSettings($user_id = 0)
+{
+	global $smcFunc, $cache_enable, $modSettings, $language;
+
+	$user_settings_min = array();
+
+	if (empty($userId) || !is_int($userId))
+		return [];
+
+	if (empty($cache_enable) || ($user_settings_min = cache_get_data('user_settings_min-' . $user_id, 120)) == null)
```
albertlast: Since the logic in this part is not complex,
is the cache checking logic more time consuming as without and
it greate a layer of complexity without need.

Sources/Load.php
```diff
@@ -849,6 +850,71 @@ function loadUserSettings()
...
+				SELECT time_offset, additional_groups, id_group, id_post_group, lngfile, smiley_set, time_offset
+				FROM {db_prefix}members
+				WHERE id_member = {int:id_member}
+				LIMIT 1',
+			array(
+				'id_member' => $user_id,
+			)
+		);
+
+		$user_settings_min = $smcFunc['db_fetch_assoc']($request);
```
albertlast: I'm not sure if you don't need some kind of check if the result amount is one.
For exp. line 883 could fail i guess.

Sources/Load.php
```diff
@@ -849,6 +850,71 @@ function loadUserSettings()
...
+				explode(',', $user_settings_min['additional_groups'])
+			);
+
+		if (!empty($user_settings_min['timezone']))
+		{
+			$tz_system = new DateTimeZone(@date_default_timezone_get());
+			$tz_user = new DateTimeZone($data['timezone']);
+			$time_system = new DateTime('now', $tz_system);
+			$time_user = new DateTime('now', $tz_user);
+			$user_settings_min['time_offset'] = ($tz_user->getOffset($time_user) - $tz_system->getOffset($time_system)) / 3600;
```
albertlast: vars get not set,
when data came from the cache (this kind of complexity i mean before)

Sources/Load.php
```diff
@@ -849,6 +850,71 @@ function loadUserSettings()
 	call_integration_hook('integrate_user_info');
 }
 
+function loadMinUserSettings($user_id = 0)
```
albertlast: Doc is missing

cron.php
```diff
@@ -332,6 +344,27 @@ public function __construct($details)
...
+	/**
+	 * Loads minimal info for the current incontext user
+	 *
+	 * @param int $user_id
+	 * @return array
+	 */
+	public function getIncontextUserInfo($user_id = 0)
+	{
+		if (empty($user_id))
+			return array();
```
albertlast: Make this check sense?

albertlast (changes requested):
Miss use of cache,
the needed complexity of the code part is not archived.

Logic is not optimize when the cron runs for more than one user,
should be designed in a away that it run once for all and not for everyone once.

Sources/Load.php
MissAllSunday: Alias are used to distinct between two or more table columns, this query uses one.

Sources/Load.php
MissAllSunday: SMF can be installed in a myriad of server configurations, your unique server env is not a realistic way to determinate if a cache checking is needed or not.

MissAllSunday:
> Logic is not optimize when the cron runs for more than one user,
> should be designed in a away that it run once for all and not for everyone once.


Can't do a one for all approach since we do not know which users are gonna be loaded.

Sources/Load.php
albertlast: would things easier to find and
it doesn't hurt you.

albertlast:
Well i checked not all,
but here https://github.com/SimpleMachines/SMF2.1/blob/59e158110f6a608e8069655db65f5dfa41fe1fbd/Sources/tasks/CreatePost-Notify.php#L136
we know all users.

Sources/Load.php
albertlast: This here had nothing do to with my env.
The thing is how many external calls you got and how high the chance is to get a cache hit.

Since every one get one notify is the chance very low that you got a cache hit.
When no cache hit is ther you do a second external call to the database.

So from what i see is that in 90% cases that you slow smf by trying to get data which is not existing,
which result that cached system is slower as a non cached system.

MissAllSunday: Not going to discuss this further unless theres solid data indicating calling cache_get_data() significantly slows down SMF.

albertlast: so you call the cache just for the lulz?

jdarwood007:
In my opinion, $user_info should not be populated unless its the current logged in user.  In a cron context, this is a guest user.  If we need to do something as the user, we should operate under the $user_profile (as Profile.php does, which also sets $cur_profile) or $memberContext (as Display.php) does.

In a consistency sake, its better if we used $cur_profile as Profile.php does then.

MissAllSunday:
Yeah, the thing is, if cron for some reason uses any SMF function which uses $user_info, the whole thing will fail since all SMF functions work around the fact that $user_info is going to be the actual current user.

In other words, SMF functions aren't pure enough to be able to "not care" which user information are they handling.

To be able to use $cur_profile we will need to modify any function that uses $user_info, something way out of scope for the current development milestones.

Sources/Load.php
```diff
@@ -849,6 +850,78 @@ function loadUserSettings()
...
+	{
+		$tz_system = new \DateTimeZone(@date_default_timezone_get());
+		$tz_user = new \DateTimeZone($user_settings_min['timezone']);
+		$time_system = new \DateTime('now', $tz_system);
+		$time_user = new \DateTime('now', $tz_user);
+		$user_settings_min['time_offset'] = ($tz_user->getOffset($time_user) - $tz_system->getOffset($time_system)) / 3600;
+	}
+
+	else
+		$user_info['time_offset'] = empty($data['time_offset']) ? 0 : $data['time_offset'];
```
jdarwood007: Wrong variable here.

Sources/Load.php
```diff
@@ -851,70 +851,79 @@ function loadUserSettings()
...
+		'smiley_set',
+		'time_offset'
+	);
 
-		$user_settings_min = $smcFunc['db_fetch_assoc']($request);
+	call_integration_hook('integrate_load_min_user_settings_columns', array(&$columns_to_load));
 
-		$smcFunc['db_free_result']($request);
+	$request = $smcFunc['db_query']('', '
+		SELECT {raw:columns}
```
jdarwood007, line 885: I'm personally not a fan of {raw: usage.  Too many chances to abuse or bypass our security checks.

Sources/Load.php
```diff
@@ -851,70 +851,79 @@ function loadUserSettings()
...
+			'columns' => implode(',', $columns_to_load)
+		)
+	);
 
-		if (!empty($cache_enable))
-			cache_put_data('user_settings_min-' . $user_id, $user_settings_min, 1800);
-	}
+	while ($row = $smcFunc['db_fetch_assoc']($request))
+	{
+		$user_settings_min[$row['id_member']] = array(
```
jdarwood007, line 896: If we are defining $user_settings_min like this, why not make it static and save on loading here?

Sources/Load.php
MissAllSunday: Abuse in what sense? if we want some flexibility via hooks then is this, a comma separated string or directly inject the implode in the query string. I prefer this method since its cleaner and passing an array offers more flexibility to mod authors..

Indeed whoever uses that hook can add anything but since the query is limited to the members table (theres no option to add other tables for example) the most they can do is break the query.

Another option is to add another type {columns} which will make some validations but theres really not much we can do to validate other than perhaps casting to string and remove empty keys.

albertlast: well... 
first you had differenty syntax across mysql and pg
secondly you could add subquery here
third build a union all select

so you can add so many tables, rows and column as you like

MissAllSunday: Which different syntax?

Yes, thats all part of what the hook provides, thats what a hook is for to let mod authors do whatever they want.Once more, is not our duty to watch mod authors and moderate what they can't or can't do.

Sesquipedalian: The question of whether we should use `{raw:columns}` or not is a red herring. It’s no more dangerous than dynamically writing the string of SQL code directly in the source code (e.g. https://github.com/SimpleMachines/SMF2.1/blob/ec38cf4b03fc1c1e9c814437dfdbca12334e6033/Sources/MessageIndex.php#L353). 

The only way to change the content of `$columns_to_load` here is via a hooked mod, which means there’s no security issue here. An attacker would first need to gain admin privileges before he could insert malicious code into this query. But if the attacker already has admin privileges, he can already do whatever he wants anyway.

Now, it is true that a badly written mod could introduce a security hole that would allow an attacker to inject malicious code via this hook, but that’s true of every mod already and doesn’t amount to an objection to @MissAllSunday’s code here.

Sources/Load.php
```diff
@@ -849,6 +850,94 @@ function loadUserSettings()
...
+ * @return array
+ * @throws Exception
+ */
+function loadMinUserSettings($users_ids = [])
+{
+	global $smcFunc, $modSettings, $language;
+	static $user_settings_min;
+
+	if (empty($user_settings_min))
+		$user_settings_min = array();
```
Sesquipedalian:
```suggestion
	static $user_settings_min = array();
```

Sources/Load.php
```diff
@@ -849,6 +850,94 @@ function loadUserSettings()
...
+
+	if (empty($users_ids))
+		return array();
+
+	// Already loaded?
+	foreach ($users_ids as $key => $user_id)
+		if (!empty($user_settings_min[$user_id]))
+			unset($users_ids[$key]);
```
Sesquipedalian:
```suggestion
	$users_ids = array_diff($users_ids, array_keys($user_settings_min));
```

Sesquipedalian (commented):
I like this. I made a couple of suggestions to make the code slightly more concise, but this looks good.

Sources/Load.php
albertlast: different syntax: mysql guys like to use ` for columns name, alias,

Sources/Load.php
albertlast: i feel overwhelmed by this code of line.

Sources/Load.php
MissAllSunday: mysql works fine with or without `

albertlast: i talk about mod authors,
they don't know what they do when they use this kind.

MissAllSunday: Please let me know when your PR for handling mysql's completely optional ` usage is done and merged to apply the changes here :)

albertlast: it's already ther: https://github.com/SimpleMachines/SMF2.1/blob/63fc00a5e9c8505bbe6a5d975251233b948f9dd0/Sources/Subs-Db-mysql.php#L278-L280

MissAllSunday: ...and all it does is stripping them back to plain strings because SMF does not use/need them.

I mean your PR changing all instances of raw and hardcoded implodes in queries to use identifier. Theres currently one usage only.

albertlast: It allow the mod author not know the details,
on pg side it's strip them again and place " on this places:
https://github.com/SimpleMachines/SMF2.1/blob/63fc00a5e9c8505bbe6a5d975251233b948f9dd0/Sources/Subs-Db-postgresql.php#L242-L244
they are designed for this case here.

MissAllSunday: Made a small test, did the following changes:

```php
	$columns_to_load = array(
		'id_member',
		'time_offset',
		'additional_groups',
		'`id_group`',
		'id_post_group',
		'lngfile',
		'smiley_set',
		'time_offset'
	);

	call_integration_hook('integrate_load_min_user_settings_columns', array(&$columns_to_load));

		SELECT {identifier:columns}
		FROM {db_prefix}members
		WHERE id_member IN ({array_int:users_ids})',
		array(
			'users_ids' => array_map('intval', array_unique($users_ids)),
			'columns' => implode(', ', $columns_to_load)
		)
	);
```

And got a nice error:

> Database Error: Unknown column 'id_member, time_offset, additional_groups, id_group, id_post_group, lngfile, smiley_set, time_offset' in 'field list'

Your proposed change won't work when using arrays, we will have to use strings which makes the change pointless anyway because when a mod author uses the hook and sees a string with no Backticks s/he will see SMF does not use `.

albertlast: array_identifier didn't exists.
identifier does support only one column,
like all others no array_*

MissAllSunday: Then how do you want to use identifier on this very own query?

albertlast: generate the identifier of the amount of columns,
or add array_identifier

MissAllSunday: first one is inefficient and unnecessary the other one is out of the scope for this PR.

albertlast: When it you to mutch work,
than take the hook out.
no one is forcing you to add this,
but when you add this it's need to be robust as possible.

MissAllSunday: Robust for who? show me exactly where and how many mod authors uses Backticks and I'll do the change.

Fact is, SMF doesn't use them, mod authors doesn't use them. This is perfectly usable as it is and its compatible with both pg and mysql. So, unless you can prove that theres a considerable amount of people who use Backticks and absolutely can't use the SMF way (and demonstrate why thy can't drop the Backticks themselves) then I'll do the change.

albertlast: tinyportal https://custom.simplemachines.org/mods/index.php?mod=97
Pretty URLs https://custom.simplemachines.org/mods/index.php?mod=636
Anti Spam https://custom.simplemachines.org/mods/index.php?mod=3851

3 of top 5 addons use backticks,
so you see the mod author build ther query based on phpmyadmin,
ther is the default to use backticks so they learn, "backticks are good".

Sesquipedalian: Problems in poorly written mods need to be fixed in the mods, not in this PR. 

Alternatively, @albertlast, if you want SMF to automatically remove those backticks from SQL statements, please feel free to add something to Subs-Db-Postgresql.php to do so.

albertlast: @Sesquipedalian the identifier do this already,
you only had to use them.

MissAllSunday: Did a quick search for backticks on TinyPortal 1.6.5. The only place where they use backticks is on their install.php file.

What I want you to understand is this, the hook priimary usage its for adding column names, column names can be easily and uniformly represented by simple string keys, this is why this:

```
	$columns_to_load = array(
		'id_member',
		'time_offset',
		'additional_groups',
		'id_group',
		'id_post_group',
		'lngfile',
		'smiley_set',
		'time_offset'
	);

```
Is a perfectly valid way to add column names. Remember, the primary usage for this hook is to add column names. Now, if a mod author wants to include a subquery or other SQL syntax  then they will have to write their query **accordingly** to how SMF builds queries. This makes sense because whenever you want to integrate with something you adapt to their way, **no the other way around**

It is far more easy for mod authors to adapt to a single unified way for writing queries than add support for every sql engine we support.

And I already spent way to much time on this, not gonna comment any further.

Sesquipedalian (approved):

PR #6030, opened 2020-02-27
fix float left/right, bug issue 5494

### LINKED ISSUE #5494: Strange behavior of img/float tag parameters in post modify function
https://www.simplemachines.org/community/index.php?topic=565710.msg4010294#msg4010294

Seems like any custom bbc parameters are lost during modification, and the float **left** parameter is always changed to **right**

tested on Firefox & Chrome

QuakeMan:
fix float left/right bug

Fixes #5494


live627 (approved):
works  as intended

Sesquipedalian (approved):

PR #6032, opened 2020-02-28
Remove uppercasing of words

BrickOzp:
Since not all languages uses uppercasing of words it can't be done
in the code. If a langauge should have uppercasing it should be done
in the language string.

### DIFF
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -304,7 +304,7 @@ function loadProfileFields($force_reload = false)
 		),
 		'passwrd1' => array(
 			'type' => 'password',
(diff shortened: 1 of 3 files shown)


jdarwood007:
Would you  mind changing the corresponding english entries as well?

BrickOzp:
English strings updated as well

chopinesque:
This is a common localization problem arising from a globalization-unaware and anglocentric coding practice. In Greek, automatic capitalization via css is OK, provided accents are dropped. But this is not done automatically and can only be resolved by adapting the body tag:

https://www.translatum.gr/forum/index.php?topic=358982.0

Sesquipedalian (approved):

PR #6033, opened 2020-03-01
fix: allow $listOptions['get_count'] to get an int 

MissAllSunday:
Right now to pass the total number of items you can pass a callable and the creatList option will call it.

But sometimes you already have that info in which case the only way to pass it is to use an anonymous function:

```php
'get_count' => [
		        'function' => function () use ($maxIndex)
		        {
		        	return $maxIndex;
		        },
		    ],
```

Which seems unnecessary and overkill.

This PR modifies createList() to allow it to directly accept an int and thus avoiding the need to execute the callable.

Its done in the "function" key rather than a new key to keep compatibility with 2.0.

### DIFF
--- a/Sources/Subs-List.php
+++ b/Sources/Subs-List.php
@@ -73,11 +73,18 @@ function createList($listOptions)
 	else
 	{
 		// First get an impression of how many items to expect.
-		if (isset($listOptions['get_count']['file']))
-			require_once($listOptions['get_count']['file']);
+		if (ctype_digit($listOptions['get_count']['function']))
+			$list_context['total_num_items'] = $listOptions['get_count']['function'];
 
-		$call = call_helper($listOptions['get_count']['function'], true);
-		$list_context['total_num_items'] = call_user_func_array($call, empty($listOptions['get_count']['params']) ? array() : $listOptions['get_count']['params']);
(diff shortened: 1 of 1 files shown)


jdarwood007:
Instead of hacking the function parameter, why not create a static value?

```
'get_count' => [
		        'value' => $maxIndex,
		    ],
```

Then the code can just use the defined value rather than a function call.

live627:
Indeed. Don't need to worry about compatibility if adding a new feature.

Sesquipedalian (approved):

Sources/Subs-List.php
```diff
@@ -73,11 +79,18 @@ function createList($listOptions)
 	else
 	{
 		// First get an impression of how many items to expect.
-		if (isset($listOptions['get_count']['file']))
-			require_once($listOptions['get_count']['file']);
+		if (!empty($listOptions['get_count']['value']) && ctype_digit($listOptions['get_count']['value']))
```
live627, line 82: I smell a bug. This won't work with integers that I know of.

MissAllSunday: Which integers do you know that won't get caught by ctype_digit?

PR #6034, opened 2020-03-01
fix: allow $listOptions['get_items'] to accept an array

MissAllSunday:
Same as #6033  but for get_items key


jdarwood007:
Similar note from your other pr.  Can we do a values instead?

Sesquipedalian (approved):

PR #6036, opened 2020-03-05
Upgrader - various issues learned via CLI

### LINKED ISSUE #5981: Upgrader - CLI errors

### LINKED ISSUE #6001: Upgrader - CLI - custom_avatar & $upgradeData
#### Description
Two issues _specific to upgrading via the CLI_:
 - If the prior version didn't have a custom_avatar directory, it is somehow not created upon upgrade
 - $upgradeData is left in Settings.php when done

### Steps to reproduce
1. Do an upgrade via CLI on a pure vanilla 2.0 or 1.x implementation with no custom avs directory setup

- Version/Git revision: Today
- Database Type: mysql
- Database Version: 5.7
- PHP Version: 7.3

Errors from trying to upload an avatar:
image

The problem is the folder & url weren't setup by the upgrader:
image

$upgradeData is still in Settings.php:
image

sbulen:
Fixes #5981 (in conjunction with #5986 )
Fixes #6001 

CLI, especially with 1.x, was generating scores of errors.  

Changes:
- The basic problem was that the '@' operator still invokes the error handler, and they get logged, even though they are OK.  Not a big deal, other than this was masking some REAL issues.  Note that although the @ operator still calls the error routine, it temporarily sets the error level to 0, so it is easy to test in the error handler.  So, the _real_ errors were...
 - Some true undefined errors were buried in there - and not visible via browser upgrades.
 - Fixed the utf8 template issue.
 - cmdstep0 is the cli version of WelcomeLogin (without a template) & needed to deal with custom_avatar

A subset of these changes are needed in 2.0 as well.

**_Testing completed..._**

### DIFF
--- a/other/upgrade-helper.php
+++ b/other/upgrade-helper.php
@@ -49,7 +49,7 @@ function getMemberGroups()
 		return $member_groups;
 
 	$request = $smcFunc['db_query']('', '
-		SELECT group_name, id_group
+		SELECT groupName, id_group
 		FROM {db_prefix}membergroups
 		WHERE id_group = {int:admin_group} OR id_group > {int:old_group}',
 		array(
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -2600,7 +2600,7 @@ function nextSubstep($substep)
 function cmdStep0()
 {
 	global $boarddir, $sourcedir, $modSettings, $start_time, $cachedir, $databases, $db_type, $smcFunc, $upcontext;
-	global $is_debug, $boardurl;
+	global $is_debug, $boardurl, $txt;
 	$start_time = time();
 
 	ob_end_clean();
@@ -2732,7 +2732,7 @@ Usage: /path/to/php -f ' . basename(__FILE__) . ' -- [OPTION]...
 	// Do we need to add this setting?
 	$need_settings_update = empty($modSettings['custom_avatar_dir']);
 
-	$custom_av_dir = !empty($modSettings['custom_avatar_dir']) ? $modSettings['custom_avatar_dir'] : $GLOBALS['boarddir'] . '/custom_avatar';
+	$custom_av_dir = !empty($modSettings['custom_avatar_dir']) ? $modSettings['custom_avatar_dir'] : $boarddir . '/custom_avatar';
 	$custom_av_url = !empty($modSettings['custom_avatar_url']) ? $modSettings['custom_avatar_url'] : $boardurl . '/custom_avatar';
 
 	// This little fellow has to cooperate...
@@ -2740,7 +2740,7 @@ Usage: /path/to/php -f ' . basename(__FILE__) . ' -- [OPTION]...
 
 	// Are we good now?
 	if (!is_writable($custom_av_dir))
-		return throw_error(sprintf($txt['error_dir_not_writable'], $custom_av_dir));
+		print_error(sprintf($txt['error_dir_not_writable'], $custom_av_dir));
 	elseif ($need_settings_update)
 	{
 		if (!function_exists('cache_put_data'))
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -2744,8 +2744,12 @@ function ConvertUtf8()
 
 	// Done it already?
 	if (!empty($_POST['utf8_done']))
-		return true;
-
+	{
+		if ($command_line)
+			return DeleteUpgrade();
+		else
+			return true;
+	}
 	// First make sure they aren't already on UTF-8 before we go anywhere...
 	if ($db_type == 'postgresql' || ($db_character_set === 'utf8' && !empty($modSettings['global_character_set']) && $modSettings['global_character_set'] === 'UTF-8'))
 	{
@@ -2756,7 +2760,10 @@ function ConvertUtf8()
 			array('variable')
 		);
 
-		return true;
+		if ($command_line)
+			return DeleteUpgrade();
+		else
+			return true;
 	}
 	else
 	{
@@ -3246,6 +3253,11 @@ function ConvertUtf8()
 			flush();
 		}
 	}
+
+	// Make sure we move on!
+	if ($command_line)
+		return DeleteUpgrade();
+
 	$_GET['substep'] = 0;
 	return false;
 }
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -1515,7 +1515,7 @@ function backupTable($table)
 function DatabaseChanges()
 {
 	global $db_prefix, $modSettings, $smcFunc, $txt;
-	global $upcontext, $support_js, $db_type;
+	global $upcontext, $support_js, $db_type, $boarddir;
 
 	// Have we just completed this?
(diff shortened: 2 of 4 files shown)


other/upgrade.php
```diff
@@ -2725,6 +2729,27 @@ function cmdStep0()
...
+
+	$custom_av_dir = !empty($modSettings['custom_avatar_dir']) ? $modSettings['custom_avatar_dir'] : $GLOBALS['boarddir'] . '/custom_avatar';
+	$custom_av_url = !empty($modSettings['custom_avatar_url']) ? $modSettings['custom_avatar_url'] : $boardurl . '/custom_avatar';
+
+	// This little fellow has to cooperate...
+	quickFileWritable($custom_av_dir);
+
+	// Are we good now?
+	if (!is_writable($custom_av_dir))
+		return throw_error(sprintf($txt['error_dir_not_writable'], $custom_av_dir));
```
live627: Missing global

sbulen: Which global?   

FYI - this corrected the issue in actual testing.

other/upgrade-helper.php
```diff
@@ -49,7 +49,7 @@ function getMemberGroups()
 		return $member_groups;
 
 	$request = $smcFunc['db_query']('', '
-		SELECT group_name, id_group
+		SELECT groupName, id_group
```
jdarwood007, line 52: Is this code intended to run prior to 2.0?  groupName will fail for any 2.0 or higher upgrades.

Maybe detect what version we are upgrading from? Or check if the column exists?

other/upgrade.php
jdarwood007: Looks like $txt

other/upgrade.php
```diff
@@ -2725,6 +2729,27 @@ function cmdStep0()
 		require_once($modSettings['theme_dir'] . '/languages/Install.' . $upcontext['language'] . '.php');
 	}
 
+	// Do we need to add this setting?
+	$need_settings_update = empty($modSettings['custom_avatar_dir']);
+
+	$custom_av_dir = !empty($modSettings['custom_avatar_dir']) ? $modSettings['custom_avatar_dir'] : $GLOBALS['boarddir'] . '/custom_avatar';
+	$custom_av_url = !empty($modSettings['custom_avatar_url']) ? $modSettings['custom_avatar_url'] : $boardurl . '/custom_avatar';
```
jdarwood007, line 2796: $boarddir is global here, don't use $GLOBALS

other/upgrade.php
sbulen: Yep, will fix.

other/upgrade-helper.php
sbulen: Correct, this routine is for 1.x only.  That is the problem being fixed here.  The getMemberGroups() function is only called from the upgrade_1-0.sql file.

other/upgrade.php
sbulen: I can make that change, but just so you're aware: $GLOBALS is used throughout.  I wasn't seeking to do a rewrite, just address the bug, in a fashion consistent with the existing code.

As noted in the comments, cmdstep0 is the CLI equivalent of WelcomeLogin(), and this code was cloned from there.

other/upgrade.php
sbulen: In hindsight, this should be made more consistent with the error processing in that function, which is CLI specific.

other/upgrade.php
jdarwood007: In that case leave it.  Should address in the future.

sbulen: I had to bring the error rpt in line with the cli errors anyway, so a couple minor changes.  I think this is better.

sbulen:
That last commit addressed undefined offset issues experienced with most upgrades:
yabb-1

sbulen:
The last issue I have found during testing is...  Interesting...

Via CLI, I am seeing a class of undefined errors that are pretty big & cause functionality gaps - but there is no corresponding gap when the upgrade is run via the browser.  

The difference is that the browser "comes up for air" at various points & restarts.  _And reloads $modSettings..._  CLI never reloads $modSettings, so the only values it knows are the ones that were there _at the start of the upgrade_.  The loss of functionality is thus worse the further back you go.

I have some thoughts to address.  If easily fixed, we're good.  If not, I think guidance needs to be that CLI isn't really appropriate for meaningful upgrades across versions.  

Some examples:
Old-upgr-1
Old-upgr-2

sbulen:
Still testing this.  Lots of permutations...

Some variations get this at the very end - looking into it:
yabb

~~I suspect this was brought about by the auth_secret changes.~~

jdarwood007:
This is why for beyond 2.1 I'm recommending we drop support for upgrading anything other than the previous major and build converters for those.  Its going to become more troublesome in the future to maintain being able to upgrade from YabbSE all the way to current.

sbulen:
Yep...

sbulen:
> This is why for beyond 2.1 I'm recommending we drop support for upgrading anything other than the previous major and build converters for those. Its going to become more troublesome in the future to maintain being able to upgrade from YabbSE all the way to current.

**_ALSO VERY IMPORTANT..._**  We need to be careful **_what gets pushed down to 2.0..._**  If we push the same changes down to 2.0.x, we will break those, too.  Some may be unavoidable (security-related).  But we need to be mindful of the impacts.

sbulen:
Testing completed.  

One minor issue outstanding that will be logged separately.

sbulen:
Since some of these issues impact current installs, I think we should merge this one asap.

other/upgrade.php
```diff
@@ -1562,7 +1573,52 @@ function DatabaseChanges()
...
+					SELECT variable, value
+					FROM {db_prefix}settings',
+					array(
+						'db_error_skip' => true,
+					)
+				);
+				$modSettings = array();
+				while ($row = $smcFunc['db_fetch_assoc']($request))
+					$modSettings[$row['variable']] = $row['value'];
+				$smcFunc['db_free_result']($request);
```
MissAllSunday, line 1592: To keep things more readable please use a space after a condition/loop is applied:

```php
	$modSettings[$row['variable']] = $row['value'];

```

Theres a few whiles and if/elses that can be separated.

MissAllSunday (changes requested):

other/upgrade.php
sbulen: Think I got 'em.  Any more, let me know.

MissAllSunday (approved):

MissAllSunday (approved):

PR #6042, opened 2020-03-08
Fix inappropriate file names

dragomano:
Currently, when we are working with hooks on Integration Hooks area, we have the problem with inappropriate file names:
Problem

After the fix, we have the proper state:
Applied_fix


MissAllSunday (approved):

MissAllSunday:
Looks good Bugo, thanks! will gonna mark it for final for now.

PR #6043, opened 2020-03-15
Better detect a working temp directory

jdarwood007:
Systems using open_basedir restrictions may receive errors with sys_get_temp_dir as it may not be properly configured to use a safe workable directory.  As well its possible that the temp directory may be readonly.  We also check if the directory is writable.

As a fall back we resort to sys_get_temp_dir again for lack of a 100% safe directory.

### DIFF
--- a/Sources/Subs-Admin.php
+++ b/Sources/Subs-Admin.php
@@ -2181,9 +2181,25 @@ function emailAdmins($template, $replacements = array(), $additional_recipients 
 		}
 }
 
-/**
+/*
  * Locates the most appropriate temp directory.
- */
+ *
+ * Systems using `open_basedir` restrictions may receive errors with
+ * `sys_get_temp_dir()` due to misconfigurations on servers. Other
+ * cases sys_temp_dir may not be set to a safe value. Additionaly
+ * `sys_get_temp_dir` may use a readonly directory. This attempts to
+ * find a working temp directory that is accessible under the
+ * restrictions and is writable to the web service account.
+ *
+ * Directories cheked against `open_basedir`:
+ *
+ * - `sys_get_temp_dir()`
+ * - `upload_tmp_dir`
+ * - `session.save_path`
+ * - `$cachedir`
+ *
+ * @return string
+*/
 function sm_temp_dir()
 {
 	global $cachedir;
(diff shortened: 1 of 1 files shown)


Sources/Subs-Admin.php
```diff
@@ -2181,4 +2181,88 @@ function emailAdmins($template, $replacements = array(), $additional_recipients
...
+
+	// Already did this.
+	if (!empty($temp_dir))
+		return $temp_dir;
+
+	// Determine if we should detect a restriction and what restrictions that may be.
+	$restriction = !empty(ini_get('open_basedir')) ? explode(':', ini_get('open_basedir')) : false;
+
+	// Prevent any errors as we search.
+	$old_error_reportig = 0;
```
live627: typo

live627:
that's a lot of duplicated code

jdarwood007:
Yea, I couldn't think of a better way to do that without slowing things down.  Thought of throwing it in a foreach loop.  But I would have to declare and store all of the variables even if I don't use them later.

jdarwood007:
Thought of something, adds another function to make it happen, but it solves it.

Sources/Subs-Admin.php
```diff
@@ -2181,4 +2181,88 @@ function emailAdmins($template, $replacements = array(), $additional_recipients
 		}
 }
 
+/**
+ * Locates the most appropriate temp directory.
+ */
```
live627: I posit  a new docblock

	/**
	 * Locates the most appropriate temp directory.
	 *
	 * Systems using `open_basedir` restrictions may receive errors with
	 * `sys_get_temp_dir()` as it may not be properly configured to use a
	 * safe workable directory. Since the temp directory may be readonly,
	 * check that it is writable.
	 *
	 * Directories cheked against `open_basedir`:
	 *
	 * - `sys_get_temp_dir()`
	 * - `upload_tmp_dir`
	 * - `session.save_path`
	 * - `$cachedir`
	 *
	 * @return string
	 */

live627:
`$ if find . -name "*.php" -exec php other/buildTools/check-eof.php {} 2>&1 \; | grep "Error:"; then exit 1; fi`

Error: Incorrect return(s) after last function/class but before EOF in ./Sources/Subs-Admin.php

MissAllSunday (approved):

PR #6048, opened 2020-03-25
Match dropzone name

### LINKED ISSUE #5989: dropzone.min.js ignores attachment settings (hardcoded 256MB limit)
I fixed all forum attachment and php.ini settings to allow files up to 2GB, but the forum still told me that my 0.7GB file was too large.

I traced it all down to: Themes/default/scripts/dropzone.min.js

I beautified the code, it contains this crucial line, which hardcodes the limit to 256MB:

```JS
maxFilesize: 256,
```

I fixed it manually to `2000` (2gb) and now large files finally work

sbulen:
Fixes #5989


sbulen:
Suki - I'm pretty sure this will do it, it tests out OK, but bouncing it off of you in case I missed something...

MissAllSunday:
yep, it seems this will do it, thanks!  still have to check that chunk stuff on dropzone and lock it on a usable version but that can be done later.

MissAllSunday (approved):

PR #6049, opened 2020-03-28
fix: return $memberContext on Load::loadMemberContext()

MissAllSunday:
To try and minimize global usage, make Load::loadMemberContext() to return $memberContext

Core code will still rely on globalizing $memberContext  but mods and external scripts can freely use the returned array as they see fit.


jdarwood007:
Don't forget the doc header.

PR #6052, opened 2020-03-29
Expose cookie

### LINKED ISSUE #6022: Installer warning - Cookie names must not be empty

sbulen:
Fixes #6022

### DIFF
--- a/other/install.php
+++ b/other/install.php
@@ -1729,7 +1729,7 @@ function DeleteInstall()
 {
 	global $smcFunc, $db_character_set, $context, $txt, $incontext;
 	global $databases, $sourcedir, $modSettings, $user_info, $db_type, $boardurl;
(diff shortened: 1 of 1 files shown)


live627 (approved):
Caused by 8b10e9e0fb8f8d3a73d19bbaa4ffaa2bc62caf61 (git blame).  and now the login cookie can be set so we don't have to login to new installs.

PR #6053, opened 2020-03-30
Add sha256_hash column to log_packages to install.sql and upgrade.sql

VBGAMER45:
Add sha256_hash column to log_packages to install.sql and upgrade.sql


mwalimu-prog:
Wow!  Wonderful @VBGAMER45

PR #6064, opened 2020-04-09
Avoids generating errors if searching for users but user list is empty

Sesquipedalian:
Saw this bug in the wild recently. PlushSearch2() already has some code to filter out empty strings in the `userspec` search parameter, but after it is done filtering, it doesn't check if there are any strings left. If there aren't, then the `$realNameMatches` array will end up being empty as well, which will in turn produce an SQL error in the "Retrieve a list of possible members" query.

This PR fixes that.

### DIFF
--- a/Sources/Search.php
+++ b/Sources/Search.php
@@ -413,8 +413,7 @@ function PlushSearch2()
 		$userQuery = '';
 	else
 	{
-		$userString = strtr($smcFunc['htmlspecialchars']($search_params['userspec'], ENT_QUOTES), array('&quot;' => '"'));
-		$userString = strtr($userString, array('%' => '\%', '_' => '\_', '*' => '%', '?' => '_'));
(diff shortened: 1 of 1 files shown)


Sources/Search.php
```diff
@@ -427,53 +427,58 @@ function PlushSearch2()
...
+						'match_possible_guest_names' => 'm.poster_name LIKE ' . implode(' OR m.poster_name LIKE ', $realNameMatches),
+					)
+				);
+			}
+			else
+			{
+				$memberlist = array();
+				while ($row = $smcFunc['db_fetch_assoc']($request))
+					$memberlist[] = $row['id_member'];
+				$userQuery = $smcFunc['db_quote'](
```
MissAllSunday, line 478: To help with readability can you add a blank space on whiles and if/else blocks?

Sesquipedalian: Sure.

PR #6065, opened 2020-04-09
fix: log_type VARCHAR(30)

### LINKED ISSUE #6031: Database Error: Data too long for column 'log_type'
#### Description
If language strings are created for WaitTime_broken for modifications, the length for column log_type in table *_log_floodcontrol is too small. This should be increased from 8 to 16.
### Steps to reproduce
1.
look here https://www.simplemachines.org/community/index.php?topic=203471.msg4047900#msg4047900

2. 
```
 Type of error: Database
Database Error: Data too long for column 'log_type' at row 1


			REPLACE INTO smf_log_floodcontrol(`ip`, `log_time`, `log_type`)
			VALUES
				(unhex('00000000000000000000000000000001'), 1582827357, 'downloads')
 http://localhost/xxx/xxx/index.php?action=downloads;sa=comment&amp;id=1
 /var/www/html/xxx/xxx/Sources/Security.php (Line 1203)
```

- Version/Git revision: SMF 2.1 RC2
- Database Type:  MYSQL
- Database Version: 5.7.28-log
- PHP Version: 7.3.14

MissAllSunday:
Fixes #6031

### DIFF
--- a/other/install_2-1_mysql.sql
+++ b/other/install_2-1_mysql.sql
@@ -370,7 +370,7 @@ CREATE TABLE {$db_prefix}log_floodcontrol (
 	ip VARBINARY(16),
 	log_time INT(10) UNSIGNED NOT NULL DEFAULT '0',
 	log_type VARCHAR(30) DEFAULT 'post',
-	PRIMARY KEY (ip, log_type(30))
+	PRIMARY KEY (ip, log_type)
 ) ENGINE={$memory};
 
 #
--- a/other/install_2-1_mysql.sql
+++ b/other/install_2-1_mysql.sql
@@ -369,8 +369,8 @@ CREATE TABLE {$db_prefix}log_errors (
 CREATE TABLE {$db_prefix}log_floodcontrol (
 	ip VARBINARY(16),
 	log_time INT(10) UNSIGNED NOT NULL DEFAULT '0',
-	log_type VARCHAR(8) DEFAULT 'post',
-	PRIMARY KEY (ip, log_type(8))
+	log_type VARCHAR(30) DEFAULT 'post',
+	PRIMARY KEY (ip, log_type(30))
 ) ENGINE={$memory};
 
 #
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -2352,7 +2352,7 @@ ALTER TABLE {$db_prefix}log_floodcontrol MODIFY log_type VARCHAR(30) NOT NULL DE
 ---#
 
 ---# Create primary key for floodcontrol
-ALTER TABLE {$db_prefix}log_floodcontrol ADD PRIMARY KEY (ip(16),log_type(30));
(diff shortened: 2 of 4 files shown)


other/install_2-1_mysql.sql
```diff
@@ -369,8 +369,8 @@ CREATE TABLE {$db_prefix}log_errors (
 CREATE TABLE {$db_prefix}log_floodcontrol (
 	ip VARBINARY(16),
 	log_time INT(10) UNSIGNED NOT NULL DEFAULT '0',
-	log_type VARCHAR(8) DEFAULT 'post',
-	PRIMARY KEY (ip, log_type(8))
+	log_type VARCHAR(30) DEFAULT 'post',
+	PRIMARY KEY (ip, log_type(30))
```
albertlast: To keep the (30) here i guess use less,
Can be deleted.

other/upgrade_2-1_mysql.sql
```diff
@@ -2347,6 +2347,10 @@ if ($doChange)
 ALTER TABLE {$db_prefix}log_floodcontrol ADD COLUMN ip VARBINARY(16);
 ---#
 
+---# Modify log_type size
+ALTER TABLE {$db_prefix}log_floodcontrol MODIFY log_type VARCHAR(30) NOT NULL DEFAULT 'post';
+---#
```
albertlast, line 2352: Need to update the pk constraint.

albertlast (changes requested):

other/upgrade_2-1_mysql.sql
MissAllSunday: What do you mean?

albertlast: Something like that
```
PRIMARY KEY (ip, log_type)
```
you need to add in the upgrade part of mysql.

MissAllSunday: But its [already there](https://github.com/SimpleMachines/SMF2.1/blob/5131f225f88dacd6bb009e6ea9b022793a028245/other/upgrade_2-1_mysql.sql#L2355)

albertlast: Okay but this differently to the install pk where (30) is defined in pk

other/upgrade_2-1_mysql.sql
```diff
@@ -2347,8 +2347,12 @@ if ($doChange)
...
+
 ---# Create primary key for floodcontrol
-ALTER TABLE {$db_prefix}log_floodcontrol ADD PRIMARY KEY (ip,log_type);
+ALTER TABLE {$db_prefix}log_floodcontrol ADD PRIMARY KEY (ip(16),log_type(30));
```
albertlast: nooooooo,
first ip is no 16 long
secondly remove the (30) from the pk definition and do the same on install part.

albertlast:
perfect

sbulen:
I tested mysql & pg, installs and updates, one each.
Everything was perfect.  Columns created/updated properly.

PR #6069, opened 2020-04-10
Remove DESC from likes index

### LINKED ISSUE #6066: MySQL 8.0 The storage engine for the table doesn't support descending indexes
#### Description
Saw this on SMF site this user ran into an issue upgrading their forum
https://www.simplemachines.org/community/index.php?topic=572637.0

Seems like it is an issue with MySQL 8.0.  table needs to be Innodb
https://www.percona.com/blog/2016/10/20/mysql-8-0-descending-indexes-can-speedup-your-queries/

- Database Type: MySQL
- Database Version: 8.0.18
- PHP Version: 7.3

sbulen:
Fixes #6066 

This brings the definition of the likes index on the messages table consistent across installs & upgrades, mysql & pg.  Turns out that the 2.1 mysql installer never created the index DESC to begin with, only the upgrader, so it has been inconsistent.

I could not find any other DESC indexes.  

I don't have a MySQL 8.0 environment to test in, hopefully this is the only issue.

Tested pg install, & various permutations of upgrades for pg & mysql.

### DIFF
--- a/other/install_2-1_postgresql.sql
+++ b/other/install_2-1_postgresql.sql
@@ -1227,7 +1227,7 @@ CREATE INDEX {$db_prefix}messages_show_posts ON {$db_prefix}messages (id_member,
 CREATE INDEX {$db_prefix}messages_id_member_msg ON {$db_prefix}messages (id_member, approved, id_msg);
 CREATE INDEX {$db_prefix}messages_current_topic ON {$db_prefix}messages (id_topic, id_msg, id_member, approved);
 CREATE INDEX {$db_prefix}messages_related_ip ON {$db_prefix}messages (id_member, poster_ip, id_msg);
-CREATE INDEX {$db_prefix}messages_likes ON {$db_prefix}messages (likes DESC);
+CREATE INDEX {$db_prefix}messages_likes ON {$db_prefix}messages (likes);
 #
 # Table structure for table `moderators`
 #
--- a/other/upgrade_2-1_postgresql.sql
+++ b/other/upgrade_2-1_postgresql.sql
@@ -2785,7 +2785,7 @@ CREATE INDEX {$db_prefix}members_birthdate2 ON {$db_prefix}members (indexable_mo
 /******************************************************************************/
 ---# Add Index for messages likes
 DROP INDEX IF EXISTS {$db_prefix}messages_likes;
-CREATE INDEX {$db_prefix}messages_likes ON {$db_prefix}messages (likes DESC);
+CREATE INDEX {$db_prefix}messages_likes ON {$db_prefix}messages (likes);
 ---#
 
 /******************************************************************************/
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -2499,7 +2499,8 @@ SET lngfile = REPLACE(lngfile, '-utf8', '');
 --- Create index for messages likes
 /******************************************************************************/
 ---# Add Index for messages likes
-CREATE INDEX idx_likes ON {$db_prefix}messages (likes DESC);
+DROP INDEX idx_likes ON {$db_prefix}messages;
+CREATE INDEX idx_likes ON {$db_prefix}messages (likes);
 ---#
 
 /******************************************************************************/


other/install_2-1_postgresql.sql
```diff
@@ -1227,7 +1227,7 @@ CREATE INDEX {$db_prefix}messages_show_posts ON {$db_prefix}messages (id_member,
 CREATE INDEX {$db_prefix}messages_id_member_msg ON {$db_prefix}messages (id_member, approved, id_msg);
 CREATE INDEX {$db_prefix}messages_current_topic ON {$db_prefix}messages (id_topic, id_msg, id_member, approved);
 CREATE INDEX {$db_prefix}messages_related_ip ON {$db_prefix}messages (id_member, poster_ip, id_msg);
-CREATE INDEX {$db_prefix}messages_likes ON {$db_prefix}messages (likes DESC);
```
albertlast, line 1230: i don't like that you want to harm postgres.
it's common that index across the different version are different,
since the database and features are differently.

other/upgrade_2-1_postgresql.sql
```diff
@@ -2785,7 +2785,7 @@ CREATE INDEX {$db_prefix}members_birthdate2 ON {$db_prefix}members (indexable_mo
 /******************************************************************************/
 ---# Add Index for messages likes
 DROP INDEX IF EXISTS {$db_prefix}messages_likes;
-CREATE INDEX {$db_prefix}messages_likes ON {$db_prefix}messages (likes DESC);
+CREATE INDEX {$db_prefix}messages_likes ON {$db_prefix}messages (likes);
```
albertlast, line 2788: same like above, don't do it.

other/install_2-1_postgresql.sql
live627: "harm" wut

sbulen: albertlast is correct that in some circumstances a DESC index can be a big improvement performance-wise:
https://www.percona.com/blog/2016/10/20/mysql-8-0-descending-indexes-can-speedup-your-queries/

My understanding is that the big benefit is if you have mixed terms in your ORDER BY clause, e.g., "ORDER BY likes DESC, name".  Which looks like what I see here in 2.1:
image

Let me think about this further...

sbulen: Actually, that was members!   This one is messages:
image

sbulen: An ideal solution might be to leave the DESC in there, and add that to the list of errors that the upgrader ignores.  

I wouldn't feel comfortable doing that without being able to test it.  I don't have access to MySQL8 anywyhere at the moment, not on my hosted server nor in my WAMP.

albertlast: you can download mysql 8.0 for wamp here: http://wampserver.aviatechno.net/

sbulen: I use EasyPHP.

sbulen:
The only place I can find this index used is in stats.php in these two places:
https://github.com/SimpleMachines/SMF2.1/blob/ec38cf4b03fc1c1e9c814437dfdbca12334e6033/Sources/Stats.php#L543
https://github.com/SimpleMachines/SMF2.1/blob/ec38cf4b03fc1c1e9c814437dfdbca12334e6033/Sources/Stats.php#L551

Given it's not used in any mixed index, I don't think there is any material performance benefit.

albertlast:
when you look at the last pr https://github.com/SimpleMachines/SMF2.1/pull/5424 how introduce line 543,
you notice that you play with fire here,
because you need a big setup to test this behavior.

End of the day i would only change mysql and
keep pg ther where it is.

sbulen:
The index you refer to has only a single item, likes (and was included in my note above).  Performance on traversing an index forwards or backwards is negligible.  Whether the index itself is DESC or ASC makes no material difference in that example.

Where a DESC index helps performance is when you have mixed ASC & DESC in your clause.  E.g., ORDER BY likes DESC, name ASC.  That's not what we have here.  

There is no case for material performance degradation by this PR.

OTOH, varying the DB definition itself by platform is a slippery slope & a practice to be avoided whenever possible.  Like here.

albertlast (approved):

gabbieian38 (approved):

PR #6070, opened 2020-04-10
 Adds an interface for disableRegisterCheck

VBGAMER45:


Sources/ManageRegistration.php
```diff
@@ -299,6 +299,7 @@ function ModifyRegistrationSettings($return_config = false)
 	$config_vars = array(
 		array('select', 'registration_method', array($txt['setting_registration_standard'], $txt['setting_registration_activate'], $txt['setting_registration_approval'], $txt['setting_registration_disabled'])),
 		array('check', 'send_welcomeEmail'),
+        array('check', 'disableRegisterCheck'),
```
live627: whitespace

VBGAMER45: Yeah....Tried to fix... You can edit if this change will be merged.

albertlast:
Is a change of the upgrade files for pg and mysql not missing?

VBGAMER45:
Added it. I think it added to this pull request. Still a mystery to me how that works...

SychO9:
> Added it. I think it added to this pull request. Still a mystery to me how that works...

It's not that hard, if you ever want to me to show you the ropes, you can always hit me up on discord.

VBGAMER45:
> 
> 
> > Added it. I think it added to this pull request. Still a mystery to me how that works...
> 
> It's not that hard, if you ever want to me to show you the ropes, you can always hit me up on discord.

I will probably will just been very busy. What was interesting about this was concerned how the commit that I pushed was going to get linked to the this pull request but it looks like github did that automatically.

jdarwood007:
If the code is already doing a empty check, then we don't need to handle this in upgrade or installs.  Just to simplify things.

PR #6073, opened 2020-04-17
Adds support for more password hashing schemes for converters

VBGAMER45:
Adds support for more password hashing schemes for converters requiring less file edits that need to be made.

Signed-off-by: Jonathan <vbgamer45@gmail.com>

### DIFF
--- a/Sources/LogInOut.php
+++ b/Sources/LogInOut.php
@@ -23,7 +23,7 @@ if (!defined('SMF'))
  *  It caches the referring URL in $_SESSION['login_url'].
  *  It is accessed from ?action=login.
  *
- * Uses Login template and language file with the login sub-template.
+ * @uses Login template and language file with the login sub-template.
  */
 function Login()
 {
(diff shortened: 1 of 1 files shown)


Sources/LogInOut.php
```diff
@@ -23,7 +23,7 @@
  *  It caches the referring URL in $_SESSION['login_url'].
  *  It is accessed from ?action=login.
  *
- * Uses Login template and language file with the login sub-template.
+ * @uses Login template and language file with the login sub-template.
```
live627: Incorrect use of tag; please revert

VBGAMER45: Damn how did that get there.

VBGAMER45: Reverted should be better now.

live627: I  guess you were working on an old file and staged too many changes at once.

VBGAMER45: Yeah more than likely I thought just pulled it down but could have been comparing from an older github release with the files on my computer.

PR #6074, opened 2020-04-17
Adds an interface for disableRegisterCheck

VBGAMER45:
Adds an interface for disableRegisterCheck 
Signed-off-by: Jonathan <vbgamer45@gmail.com>


Sesquipedalian:
Honestly, I don't think we should have a UI for this setting. There are virtually no circumstances in which it is a good idea to disable this check. Turning it off amounts to rolling out the red carpet to spammers to bulk register a ton of fake accounts in quick succession. I'd be more inclined to delete the one occurrence of this setting in Subs-Members.php than to add a UI for it.

But if you have a counterargument or a clear example of how this setting can be useful, I'll be glad to let you try to convince me. 🙂

MissAllSunday:
I'm more inclined to leave it as it is and let power users to use however they want. Their might be valid reasons (valid for them at least) But I agree not to make it easier to change it.

MissAllSunday:
We will leave it as it is in case someone really wants to disable the registration check, however lets not make it easier to do the change to prevent unwanted behavior due to admins experimenting with their admin panel.

PR #6080, opened 2020-04-20
Fixes the fix for the sceditor xd

DiegoAndresCortes:
Fixes issues caused by #5426 
and extra fixes for #5370 ??

😋

### DIFF
--- a/Themes/default/css/jquery.sceditor.css
+++ b/Themes/default/css/jquery.sceditor.css
@@ -180,7 +180,7 @@ div.sceditor-grip {
 	background-clip: padding-box;
 	min-width: 100%;
 	max-width: 100%;
-	min-height: 307px;
+	min-height: 332px;
 }
 .sceditor-container *,
 .sceditor-container *::before,
@@ -210,10 +210,10 @@ div.sceditor-dropdown div {
 	resize: none;
 	background: #fff;
 	display: flex;
-	flex-basis: 150px;
+	flex-basis: auto;
 	min-width: 100%;
 	max-width: 100%;
-	height: auto !important;
+	height: 100%;
 	min-height: 1px;
 }
 .sceditor-container iframe {
--- a/Themes/default/css/jquery.sceditor.css
+++ b/Themes/default/css/jquery.sceditor.css
@@ -8,162 +8,162 @@
  * SCEditor is licensed under the MIT license:
  *	https://opensource.org/licenses/mit-license.php
  */
- div.sceditor-grip,
- .sceditor-button div {
+div.sceditor-grip,
+.sceditor-buttondiv {
 	background-image: url("../images/icons/editor_sprite.png");
 	background-repeat: no-repeat;
 	width: 16px;
 	height: 16px;
 }
- .sceditor-button-flash div {
+.sceditor-button-flash div {
 	background-position: 0px -712px;
 }
- .sceditor-button-youtube div {
+.sceditor-button-youtube div {
 	background-position: 0px 0px;
 }
- .sceditor-button-link div {
+.sceditor-button-link div {
 	background-position: 0px -16px;
 }
- .sceditor-button-unlink div {
+.sceditor-button-unlink div {
 	background-position: 0px -32px;
 }
- .sceditor-button-underline div {
+.sceditor-button-underline div {
 	background-position: 0px -48px;
 }
- .sceditor-button-time div {
+.sceditor-button-time div {
 	background-position: 0px -64px;
 }
- .sceditor-button-table div {
+.sceditor-button-table div {
 	background-position: 0px -80px;
 }
- .sceditor-button-superscript div {
+.sceditor-button-superscript div {
 	background-position: 0px -96px;
 }
- .sceditor-button-subscript div {
+.sceditor-button-subscript div {
 	background-position: 0px -112px;
 }
- .sceditor-button-strike div {
+.sceditor-button-strike div {
 	background-position: 0px -128px;
 }
- .sceditor-button-source div {
+.sceditor-button-source div {
 	background-position: 0px -144px;
 }
- .sceditor-button-size div {
+.sceditor-button-size div {
 	background-position: 0px -160px;
 }
- .sceditor-button-rtl div {
+.sceditor-button-rtl div {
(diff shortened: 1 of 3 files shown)


DiegoAndresCortes:
I'm still testing this btw but forgot to mark as draft

DiegoAndresCortes:
I think it's ready if you want to test it

DiegoAndresCortes:
So supposedly it shouldn't shrink, but allowing the editor to shrink below the set height is the ideal solution here.

Obviously the main issue #5370 it's solved, however when resizing the box it will set this "custom" height to the editor which causes this again. **BUT** the main issue is not dealing with this when the user is not playing around with the size of the editor (the default size), which should now be fixed without causing any _other_ problems.

Well, also forgot to mention that _it would only invoke_ the issue when the user **resizes the browser window** because obviously that's the only way to increase the height of the bbc/emoticons in real time.

So naturally, this is fixed imo.

Thanks for reading then, this was indeed a tricky one.

Themes/default/css/jquery.sceditor.css
```diff
@@ -197,24 +198,24 @@ div.sceditor-dropdown div {
...
 	resize: none;
 	background: #fff;
-	display: block;
+	display: flex;
+	flex-basis: 175px;
+	flex-grow: 1;
 	min-width: 100%;
 	max-width: 100%;
-	height: auto !important;
+	height: 175px;
```
live627, line 216: Now what would happen if a textedit control/mod specifies a longer height? Should this then be  a custom data attribute?

DiegoAndresCortes: Can your JavaScript code target this using height instead of minHeight?

DiegoAndresCortes: Is there an example of this behavior also?? To look into it

DiegoAndresCortes:
I believe this one fixes the issue @live627

Themes/default/scripts/jquery.sceditor.smf.js
```diff
@@ -932,4 +932,4 @@ sceditor.formats.bbcode.set(
 		},
 		html: '<div class="videocontainer"><div><iframe frameborder="0" src="https://www.youtube.com/embed/{0}?wmode=opaque" data-youtube-id="{0}" allowfullscreen></iframe></div></div>'
 	}
-);
+);
```
XinYenFon, line 935: my sweet innocent file endings :|

sbulen:
I definitely like this behavior better, it feels more solid.  Three odd things I've seen...  

(1) You can size the window _smaller_ than available space, & create some funky behavior.  

E.g., I can shrink from here:
image

To here.  Note the text is completely gone, & the 'attachments' line can be scrunched/hidden as well:
image

(2) The other thing is that the MORE window is massive and obscures everything (100% browser zoom):
image

(3) Resize the browser window itself (to 150%) & the MORE window appears to be transparent:
image


Are these fixable?

DiegoAndresCortes:
1 is tricky I commented on this behavior above, it seems to be the best solution but I’ll play around with it a bit more and see what can be improved.

2n3 I didn’t test those, I’ll make sure to have a look.

Also I don’t have internet atm so a few days at least.

SychO9:
@sbulen why is your smiley box like that ? it should be one line,

the best way to fix that issue is for use to force the smiley box to a fixed height and avoid more lines of smileys, that's why we have the "More" button in the first place, because we only show a few of them inline and everything in the modal box.

XinYenFon:
1) Resizing issue seems from our side... 
2) That's default behavior from SCEditor side, I strongly suggest it to keep it as is...
3) `.sceditor-container, .sceditor-container div, div.sceditor-dropdown, div.sceditor-dropdown div` has low z-index... increasing it to 5 (or removing it) solves the issue (for now)***

*** The default behavior for "More" is just to extend the list unlike our approach where you create extra-pop... 

@SychO9 You can enable custom smiley ordering and create massive pain to developers...

SychO9:
> @SychO9 You can enable custom smiley ordering and create massive pain to developers...

Oh boi...

sbulen:
I've seen users with dozens of them...

XinYenFon:
Well, since (by design) SCEditor has extend (while SMF has pop), the inline elements needs to become block elements this will solve some of the issues...

For next version we can discuss how to change this feature (if required or not).

live627:
> That's default behavior from SCEditor side, I strongly suggest it to keep it as is...

No, it's custom code that I helped write.

XinYenFon:
> > That's default behavior from SCEditor side, I strongly suggest it to keep it as is...
> 
> No, it's custom code that I helped write.

hmm... why may i ask ? Default behavior seems good enough. I don't recall the reasoning behind moving the "more" part into a popup container.

DiegoAndresCortes:
So the z-index fix is the only thing we want to do for now?

XinYenFon:
yes (?) unless you guys think its good idea to revert "More" func. back into the original one

sbulen:
Is there nothing that can be done about the size of the more box?

XinYenFon:
Its size already defined inside CSS, we can simply resize it and be done with it

SychO9:
@DiegoAndresCortes can you please push a fix for the height of the "more" smileybox ? I think you just have to remove the `height: 50%` from `div.sceditor-smileyPopup` and add `max-height: 50%` & `overflow-y: auto` to `div.sceditor-smileyPopup, div.sceditor-dropdown`

I think this should go into RC3 because of how obvious the bug is :/
repost

DiegoAndresCortes:
Why is this box "fixed" instead of "absolute". Is that the expected behavior we want?

SychO9:
> Why is this box "fixed" instead of "absolute". Is that the expected behavior we want?

yea, I think if it was `absolute` it would just be floating at the top of the page, so when you click more you actually wouldn't see it until you scroll all the way up. that's why `fixed` is more appropriate.

MissAllSunday:
Hi @DiegoAndresCortes whats the status on this one? does testing is still required or do you feel its ready to go?

DiegoAndresCortes:
Should be ready

MissAllSunday (approved):

PR #6084, opened 2020-04-25
Used Count(*) where possible

albertlast:
When a count(*) is used the db had to check if the row is visible
When a count(col) is used the db had to check if the row is visible and is not null

Less checks = faster

Changed only count where no join or normal join is used.

On a plain 11 mio rows table was the different 600 ms vs 400 ms

### DIFF
--- a/Sources/Drafts.php
+++ b/Sources/Drafts.php
@@ -549,7 +549,7 @@ function showProfileDrafts($memID, $draft_type = 0)
 	// Get the count of applicable drafts on the boards they can (still) see ...
 	// @todo .. should we just let them see their drafts even if they have lost board access ?
 	$request = $smcFunc['db_query']('', '
-		SELECT COUNT(id_draft)
+		SELECT COUNT(*)
 		FROM {db_prefix}user_drafts AS ud
 			INNER JOIN {db_prefix}boards AS b ON (b.id_board = ud.id_board AND {query_see_board})
 		WHERE id_member = {int:id_member}
@@ -723,7 +723,7 @@ function showPMDrafts($memID = -1)
 
 	// Get the count of applicable drafts
 	$request = $smcFunc['db_query']('', '
-		SELECT COUNT(id_draft)
+		SELECT COUNT(*)
 		FROM {db_prefix}user_drafts
 		WHERE id_member = {int:id_member}
 			AND type={int:draft_type}' . (!empty($modSettings['drafts_keep_days']) ? '
--- a/Sources/Errors.php
+++ b/Sources/Errors.php
@@ -133,7 +133,7 @@ function log_error($error_message, $error_type = 'general', $file = null, $line 
 		if (!isset($context['num_errors']))
 		{
 			$query = $smcFunc['db_query']('', '
-				SELECT COUNT(id_error)
+				SELECT COUNT(*)
 				FROM {db_prefix}log_errors',
 				array()
 			);
--- a/Sources/Likes.php
+++ b/Sources/Likes.php
@@ -344,7 +344,7 @@ class Likes
 		global $smcFunc;
 
 		$request = $smcFunc['db_query']('', '
-			SELECT COUNT(id_member)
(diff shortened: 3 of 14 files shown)


gabbieian38 (approved):

gabbieian38 (approved):

gabbieian38 (approved):

jdarwood007:
Any information to back this up?  I remember back in the day (may be many years ago), that it was said COUNT(col) was faster than COUNT(*)
Is this different between postgresql and mysql?

MissAllSunday:
It relies on the DB engine to figure it out when using * to only check for the existence of a table, meaning old engines will still try to fetch all tables making it slower.

albertlast:
count(*) is a special word,
that had nothing do with select *.
So the db will not try to fetch all tables.

albertlast:
since wrong information get again spreaded by person how don't know...
try this:
```
CREATE TABLE IF NOT EXISTS asdf (
  id int(11) NOT NULL AUTO_INCREMENT,
  wert varchar(60) DEFAULT NULL,
  PRIMARY KEY (id)
) ENGINE=MyISAM;
CREATE TABLE IF NOT EXISTS asdf2 (
  id int(11) NOT NULL AUTO_INCREMENT,
  wert varchar(60)  DEFAULT NULL,
  PRIMARY KEY (id)
) ENGINE=INNODB;
```
fill them with ~ 10mio rows
do a
```
select count(*) from asdf2;
```
```
select count(wert) from asdf2;
```
i got time different in mysql 8.0 between 10x(innodb) till 100x(myisam) between count(*) and count(wert)
Myisam get more hurted by this syntax and
smf team "recommand" to keep on myisam for old=big setups :)

jdarwood007:
https://www.percona.com/blog/2007/04/10/count-vs-countcol/

This blog seems to indicate the favor of using count(*) is hitting indexes where count(id) is not.  At least for MyISAM.

However: https://www.percona.com/blog/2006/12/01/count-for-innodb-tables/
There is a note that InnoDB is slower at using count(*)

So not really good evidence.

MissAllSunday:
Indeed, doing a COUNT(*) in MyISAM is fast because MyISAM keeps a row count in memory, something InnoDB cannot do

albertlast:
the url are from 2006 and 2007.
 it make no sense that you talk about stuff without testing this.

count(*) was in the test mention above from me im >worst< case the same speed like count(wert) but in common it was 10x or 100x faster...

jdarwood007:
@albertlast Any references that are more recent?

I'm not against or for this change in any way right now.  I want to know the evidence behind this suspected change to support that it would be a better global change.

The only indicator I see is in the ones I linked that MyISAM will be faster than InnoDB when doing COUNT(*) without any WHERE as it can use the cached information about the table size.  All other queries are basically saying that InnoDB and MyISAM will produce similar results because both have to use indexes to find the results.  With id columns being faster if used on the indexed primary key or indexed columns that don't allow nulls.

albertlast:
got no reference i test directly by my self,
in the moment when you add a where part to the statement innodb get faster than myisam
(0,38 ms vs 2,7 seconds)
in myisam the behavior changed dynamical based on if the column is null able or not,
when is null able it goes slower, like you already mention.
The problem of the syntax of count(col) get changed in this moment too,
so its dangerours to had count(col) because the behavior depend on the definiton of the column,
where count(*) always work in the same way undepend of how the column is defined.

jdarwood007:
So only on columns with nulls.  Only 2 queries in what you changed I would have to look at.  themes and the moderators tables.  All others are hitting the primary key that is auto incrementing  and unique.  So a null can't exist on those.

albertlast:
yea and
now ask yourself what is idea behind the query,
is it
a) count the values of pk fields -> count(pk)
of
b) get the amound of rows -> count(*)

This is the information we send to the execution planer of the database,
maybe he come up with the same solution for both question, but sometimes not.

live627 (approved):

PR #6089, opened 2020-04-27
Lastpost (MessageIndex) alignment issue / Breaking things

### LINKED ISSUE #6044: Minor alignment issue - should get date & time on one line
#### Description
image

Also feels like replies/views got nudged over, too.

### LINKED ISSUE #6071: Error log display
#### Description
Something happened to the display of the error log, I hadn't seen this before.  The individual entries exceed the boundaries of the page.

### Steps to reproduce
1. Look at errors...

- Version/Git revision:  Today
- Database Type:  MySQL
- Database Version:  5.7
- PHP Version: 7.3

image

XinYenFon:
fixes #6044
fixes #6071

### DIFF
--- a/Themes/default/Errors.template.php
+++ b/Themes/default/Errors.template.php
@@ -253,7 +253,7 @@ function template_attachment_errors()
 		<div class="windowbg">
 			<div class="padding">
 				<div class="noticebox">',
-	$context['error_message'], '
+					$context['error_message'], '
 				</div>';
 
 	if (!empty($context['back_link']))
--- a/Themes/default/Errors.template.php
+++ b/Themes/default/Errors.template.php
@@ -253,7 +253,7 @@ function template_attachment_errors()
 		<div class="windowbg">
 			<div class="padding">
 				<div class="noticebox">',
-					$context['error_message'], '
+	$context['error_message'], '
 				</div>';
 
 	if (!empty($context['back_link']))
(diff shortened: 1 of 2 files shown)


SychO9 (approved):

Sesquipedalian (dismissed):
Let's also change the CSS for the word_break class from `word-break: break-all` to `word-wrap: break-word`. It achieves the same goal, but it's less disruptive in the process.

XinYenFon:
not sure about that... afaik there are some cases break-word fails due to some weird combination of `\/` that's why we agreed on breaking all... (also why break partially when you can break whole glass in one push...)

Sesquipedalian:
> not sure about that... afaik there are some cases break-word fails due to some weird combination of `\/` that's why we agreed on breaking all...

Fair enough, but then I don't think it is a good idea to apply `word-break: break-all` to the entire error_info div. We only want that applied to specific strings within the div—just the URLs, really. The error message would be better handled by `word-wrap: break-word` plus `overflow:auto` to handle any rare case where a browser bug messes up the word breaking for whatever reason.

XinYenFon:
> > not sure about that... afaik there are some cases break-word fails due to some weird combination of `\/` that's why we agreed on breaking all...
> 
> Fair enough, but then I don't think it is a good idea to apply `word-break: break-all` to the entire error_info div. We only want that applied to specific strings within the div—just the URLs, really. The error message would be better handled by `word-wrap: break-word` plus `overflow:auto` to handle any rare case where a browser bug messes up the word breaking for whatever reason.

we can handle the situation far better in next version, rework the area to show better info with better output (maybe even with -copy- button?) but for now its -imo- best to merge this as is.

MissAllSunday (approved):

PR #5081, opened 2018-10-22
Allows full XML-based export of profile, posts, and personal messages

Sesquipedalian:
1. Fixes a few bugs in XML feed generation and does a few things more intelligently.
2. Adjusts content of profile export to make it even more suitable for GDPR purposes.
3. Adds a sub-action to export a single user's posts (with relevant permissions checks)
4. Adds a sub-action to export a single user's personal messages (with relevant permissions checks)
5. ~Adds a "batch mode" method for compiling large XML datasets without overwhelming the server (this is only available for use when export ing a user's posts or personal messages).~ Adds a background task and some dedicated functions in Profile-Actions.php to handle compiling large XML datasets without overwhelming the server (this is only available for use when exporting a user's posts or personal messages for download).

PR #6096, opened 2020-04-28
Address MariaDB limitation in strict mode

### LINKED ISSUE #6094: Database error during export of profile data
#### Description
Database Error: 'smf_home.pm.msgtime' isn't in GROUP BY

### Steps to reproduce
1.  Go to Profile - Download profile data - Check all checkboxes - Run export
2.  After some time we have an error in the Error log:

- Version/Git revision:  6cda6d8e1c3b65ce80ab75de519e327b283196b2
- Database Type: MariaDB
- Database Version: 10.4.12
- PHP Version: 7.2.29

sbulen:
Fixes #6094 

@dragomano - please test!

@albertlast - input welcome.  I felt the more narrow query in the subquery made more sense.  I did A/B tests with my biggest users & no noticeable difference.

### DIFF
--- a/Sources/News.php
+++ b/Sources/News.php
@@ -64,7 +64,6 @@ function ShowXmlFeed()
 
 	// Default to latest 5.  No more than 255, please.
 	$context['xmlnews_limit'] = empty($_GET['limit']) || (int) $_GET['limit'] < 1 ? 5 : min((int) $_GET['limit'], 255);
-	$context['xmlnews_offset'] = empty($_GET['offset']) || (int) $_GET['offset'] < 1 ? 0 : (int) $_GET['offset'];
 
 	// Users can always export their own profile data
 	if (in_array($subaction, array('profile', 'posts', 'personal_messages')) && !$user_info['is_guest'] && $context['xmlnews_uid'] == $user_info['id'])
@@ -223,7 +222,7 @@ function ShowXmlFeed()
 		$feed_meta[$mkey] = strip_tags($mvalue);
 
 	// We only want some information, not all of it.
-	$cachekey = array($xml_format, $_GET['action'], $context['xmlnews_limit'], $subaction, $context['xmlnews_offset']);
+	$cachekey = array($xml_format, $_GET['action'], $context['xmlnews_limit'], $subaction);
 	foreach (array('board', 'boards', 'c') as $var)
 		if (isset($_GET[$var]))
 			$cachekey[] = $var . '=' . implode(',', $_GET[$var]);
@@ -706,10 +705,9 @@ function getXmlMembers($xml_format, $ascending = false)
 		SELECT id_member, member_name, real_name, date_registered, last_login
 		FROM {db_prefix}members
 		ORDER BY id_member {raw:ascdesc}
-		LIMIT {int:limit} OFFSET {int:offset}',
+		LIMIT {int:limit}',
 		array(
 			'limit' => $context['xmlnews_limit'],
-			'offset' => $context['xmlnews_offset'],
 			'ascdesc' => !empty($ascending) ? 'ASC' : 'DESC',
 		)
 	);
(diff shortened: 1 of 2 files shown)


Sources/News.php
```diff
@@ -2546,18 +2546,23 @@ function getXmlPMs($xml_format, $ascending = false)
...
+				INNER JOIN {db_prefix}pm_recipients AS pmr ON (pm2.id_pm = pmr.id_pm)
+				INNER JOIN {db_prefix}members AS mem ON (mem.id_member = pmr.id_member)
+			WHERE pm2.id_pm > {int:start_after}
+				AND (
+					(pm2.id_member_from = {int:uid} AND pm2.deleted_by_sender = {int:not_deleted})
+					OR (pmr.id_member = {int:uid} AND pmr.deleted = {int:not_deleted})
+				)
+			GROUP BY pm2.id_pm
+			LIMIT {int:limit} OFFSET {int:offset}
+		) nis ON nis.id_pm = pm.id_pm
```
jdarwood007: nitpick here but we have always used "AS" even though its not required, as we are verbose.  Should do that even for a sub select.

dragomano:
It works fine, thanks :)

albertlast:
my test env didn't got a big pm setup.. so i cant this from a performance pov.
Personaly i had tried before to solve this by using distinct
```
		SELECT DISTINCT pm.id_pm, pm.msgtime, pm.subject, pm.body, pm.id_member_from, pm.from_name, ' . $select_id_members_to . ' AS id_members_to, ' . $select_to_names . ' AS to_names
		FROM {db_prefix}personal_messages AS pm
			INNER JOIN {db_prefix}pm_recipients AS pmr ON (pm.id_pm = pmr.id_pm)
			INNER JOIN {db_prefix}members AS mem ON (mem.id_member = pmr.id_member)
		WHERE pm.id_pm > {int:start_after}
			AND (
				(pm.id_member_from = {int:uid} AND pm.deleted_by_sender = {int:not_deleted})
				OR (pmr.id_member = {int:uid} AND pmr.deleted = {int:not_deleted})
			)
		ORDER BY pm.id_pm {raw:ascdesc}
		LIMIT {int:limit} OFFSET {int:offset}',

sbulen:
It performs well on my test copy of my prod forum.   

I think it's either COUNT DISTINCT *or* GROUP_CONCAT...  but we need the GROUP_CONCAT function, so distinct doesn't work:
https://dev.mysql.com/doc/refman/8.0/en/group-by-functions.html

It works in my (small) pg environment.

albertlast:
Okay didn't notice the group_concat...
from a technical pov is the subqery from you differently to orginal because of position of the order by.

Limit is affected by the order by,
so to spread them into different parts of the query could lead in a different result.

sbulen:
Correct.  GROUP BY doesnt necessarily sort.  I'll tweak it a bit tomorrow...

sbulen:
*** Note I also removed the OFFSET for posts & PMs...  ***

The offset should be a #, a count of msgs, _usually a multiple of your LIMIT_.  Note that the logic saved off the last msg or pm retrieved, and subsequent calls included that in the WHERE clause.  So LIMIT by itself is appropriate here.  

Input welcome.

Sesquipedalian (changes requested):
I'm fine with removing the use of `OFFSET` from these queries, sure. I added that back in my first version of the export feature, but then I switched over to using `WHERE pm.id_pm > {int:start_after}` for greater efficiency, which means we no longer using the offset value at all.

However, if you remove `OFFSET` from these queries, you also need to remove all occurrences of `$context['xmlnews_offset']` in all files.

sbulen:
Done, & tests OK.  

I thought that was used by the xml feeds prior to #5081 ...  Interesting, it was defined, just never used.

sbulen:
@Sesquipedalian - if this looks good let me know, we can merge

Sesquipedalian (approved):

PR #6097, opened 2020-04-29
Php Unite for install on postgres

albertlast:
I got a dream,
to get an ci for install for pg powered by travis...

The result is that after some time i got a running phpunit script and
after many hours and 48 commits later i lost the fight versus travis,
see https://travis-ci.org/github/albertlast/SMF2.1/builds/680275492 .

The choosen structure is based on best practise of phpunit and composer.

But maybe the first part is a good start point for further steps.

Something what you need to prepare to run this is:
move/copy all install files from others

```
 mv ./other/install.php .
 mv ./other/install_2-1_mysql.sql .
 mv ./other/install_2-1_postgresql.sql .
 mv ./other/Settings.php .
 cp ./Settings.php ./Settings_bak.php
 mv ./other/readme.html .
 mv ./other/db_last_error.php .
```
and got a empty postgres db and change the db access data in the installTest.php

The installTest run throw the major function with the result that you got after this a full working smf,
check the result of the install steps -> true = good.


Tests/installTest.php
```diff
@@ -0,0 +1,75 @@
+<?php
+
+/**
+ * The main purpose of this file is to show a list of all errors that were
```
frandominguezl: Forgot to change file description ;)

other/install.php
```diff
@@ -432,7 +432,7 @@ function installExit($fallThrough = false)
 	}
 
 	// Bang - gone!
-	die();
+	if (!defined('SMFPHPUNIT')) die();
```
MissAllSunday, line 435: Any CI and/or Unit testing tools shouldn't interfere with the script normal flow. There shouldn't be any conditionals while unit testing otherwise it defeats the purpose of the tests as a whole.

MissAllSunday (changes requested):

other/install.php
albertlast: Than advice a better solution for this case here.

MissAllSunday: While performing unit tests you usually bootstrap the minimal script flow in order to test what you want to test.

You can have a shell script executed by travis to setup all you need before performing your tests. Including setting up a dummy DB with credentials.

Might be better to move your declared $_POST vars to .env vars to further automatize the process.

albertlast: this doesn't solve the problem here,
please provide a solution how the die part is skipped.

Like i mention the mayor idea is here to test all function related to the install process,
not the ui stuff.
Also i already change globales in a way that the install process work,
but for this didn't exists any global.

MissAllSunday: The solution is not to change the program normal flow  and instead build a proper bootstrapped setup.

In this case the proper way to handle it and to test it would be to move all the die() bits into a separate function, then that function can be mocked by PHPUNIT.

albertlast: and how get the function the knowledg?
By a global or a constant?

MissAllSunday: You don't. Thats the whole point.  SMF shouldn't have to have any testing code or conditionals.

Instead you will need to either use some tool to be able to override functions https://www.php.net/manual/es/book.runkit.php

or use bootstrap your entire test suite and use a register_shutdown_function()  to convert any exit/die calls into proper testable exceptions.  But even this will have its own set of issues.


Overall, I wouldn't recommend trying to unitest the installation, instead I would go for integration tests or behavior/functional testing

albertlast: Feel free to contribute a unite test.

since your mention solution based on changing smf code at runtime,
i see no diference to my solution.

MissAllSunday: Doing it a runtime means theres no code change inside SMF.

There should be no changes inside the codebase to add conditionals for running tests and the changes needed to be able to run unit tests are far beyond the 2.1 scope.

So no, overall theres no way to include a test suite at this point.

albertlast: since the unite test work fine on local,
means that it exists a way to add test suite atm.

MissAllSunday: If you can find a way to add tests without changing the current codebase, go ahead :)

There shouldn't be any  changes to the existing codebase to add conditionals for unit testing this is an axiom and it won't change.  Changes to help unit testing? sure, changes to condition a flow specific for unit test? nope.

I provided possible solutions, its up to you to accept them or not but as it is right now this PR cannot be merged.

albertlast: since you asked me to move the die to an own function,
mean your solution i had to change to code only for united test.

So your own solution didn't fit your requirement.

MissAllSunday: Thats correct, it didn't fit the requirements as it takes more than moving the die to its own function to fit the requirements.

As I have been told repeatedly, a proper bootstrap solution has to be implemented, alone side with a way to intercept those pesky exit/die calls and create a script to automatize the whole process.

You would also need to properly move your _POST vars to env vars and provide a bootstrap file for your suite case. Chance are you will need a script to build your DB and connect to it if you really want to test behavior and not just unit testing.

Not to mention under no circumstances test code should be shipped in the final product, thats just a big no and its reason enough to dismiss this attempt as it can be considered a safety issue.

So, once again, as it is right now, this cannot be merged and yes my suggestions are just that, suggestions to move to the right direction and aren't considered the final solution to all issues.

albertlast: To use bootstrap make no sense from the start,
since we call directly the function and skip the ui part?

Since you got only a vision which is not working and
this pr is working and real i would suggest to take it and improved it.

MissAllSunday: By bootstrap I mean phpunit bootstrapping:  https://en.wikipedia.org/wiki/Bootstrapping  since we are talking about phpunit I thought it will be obvious that I was referring to this and not the UI framework.

And no, just because it works doesn't mean it should be included. It works out of a conditional, you are basically telling SMF to behave differently if unit tests are run. That alone breaks the whole testing point and makes your tests useless since you aren't testing the real behavior of an application. You are accommodating the application to pass your test.

I once again will repeat it and will be the last time I say it:  There shouldn't be any conditionals   that changes the behavior of a script in order to satisfy testing.

albertlast: Than make a pr where you add the mention stuff and
than we could update the unite test logic.
Until this is this the best solution which we got,
since we got nothing here.

MissAllSunday:
Not gonna follow this route for unit testing

PR #6100, opened 2020-05-03
Migrate Privacy Policy system from SMF 2.0.16

### LINKED ISSUE #6075: Privacy policy support
We need to port the privacy policy system forward from 2.0.16+, including the system that records the user's acceptance of the registration agreement and privacy policy and prompts for re-acceptance when they change. The code from 2.0.x will need some changes in order to work in 2.1.

MisterVector:
This pull request migrates the privacy policy system added to SMF 2.0.16.

This closes #6075


MisterVector:
I may have missed some things as I'm not very familiar with the codebase. Just point me in the right direction!

Themes/default/Register.template.php
```diff
@@ -701,4 +727,73 @@ function template_edit_reserved_words()
...
+						<textarea cols="70" rows="20" name="policy" id="agreement">', $context['policy'], '</textarea>
+					</p>
+					<div class="information">
+						<span>', $context['policy_info'], '</span>
+					</div>
+					<div class="righttext">
+						<input type="submit" value="', $txt['save'], '" tabindex="', $context['tabindex']++, '" class="button" onclick="return resetPolicyConfirm()" />
+						<input type="hidden" name="policy_lang" value="', $context['current_policy_lang'], '" />
+						<input type="hidden" name="sa" value="policy" />
+						<input type="hidden" name="', $context['session_var'], '" value="', $context['session_id'], '" />
```
SychO9: Need to also add the token you created in the source files, `admin-regp`

a hidden input containing the token details
```php
<input type="hidden" name="', $context['admin-regp_token_var'], '" value="', $context['admin-regp_token'], '" />
```

sbulen:
Getting a couple undefineds - I think this was from editing the policies:
image

Themes/default/Agreement.template.php
```diff
@@ -0,0 +1,91 @@
...
+	{
+		echo '
+		<div class="cat_bar">
+			<h3 class="catbg">', $txt['agreement' . ($context['can_accept_agreement'] ? '_updated' : '')], '</h3>
+		</div>';
+
+		if ($context['can_accept_agreement'])
+		{
+			echo '
+		<p class="description">
```
XinYenFon: we are no longer using p element as description, please use div instead.

Themes/default/Agreement.template.php
```diff
@@ -0,0 +1,91 @@
...
+			', $txt['agreement_updated_desc'], '
+		</p>';
+		}
+		elseif (!empty($context['agreement_accepted_date']))
+		{
+			echo '
+		<p class="description">
```
XinYenFon: same here p > div

Themes/default/Agreement.template.php
```diff
@@ -0,0 +1,91 @@
...
+		echo '
+		<div class="roundframe">
+			<div>', $context['agreement'], '</div>
+		</div>';
+	}
+
+	if (!empty($context['policy']))
+	{
+		if (!empty($context['agreement']))
+			echo '<br />';
```
XinYenFon: Invalid HTML5 :) you can leave this br tag out completely because block elements and proper padding/margin should be enough to create space.

Themes/default/Agreement.template.php
```diff
@@ -0,0 +1,91 @@
...
+
+		echo '
+		<div class="cat_bar">
+			<h3 class="catbg">', $txt['privacy_policy' . ($context['can_accept_privacy_policy'] ? '_updated' : '')], '</h3>
+		</div>';
+
+		if ($context['can_accept_privacy_policy'])
+		{
+			echo '
+		<p class="description">
```
XinYenFon: p > div

Themes/default/Agreement.template.php
```diff
@@ -0,0 +1,91 @@
...
+			', $txt['privacy_policy_updated_desc'], '
+		</p>';
+		}
+		elseif (!empty($context['privacy_policy_accepted_date']))
+		{
+			echo '
+		<p class="description">
```
XinYenFon: p > div

Themes/default/Register.template.php
```diff
@@ -701,4 +727,74 @@ function template_edit_reserved_words()
...
+						<script>
+							function resetPolicyConfirm()
+							{
+								return true;
+							}
+						</script>
+					</div>
+				</form>
+			</div>
+			<span class="botslice"><span></span></span>
```
XinYenFon: slices are no longer used in SMF 2.1

Themes/default/Register.template.php
```diff
@@ -701,4 +727,74 @@ function template_edit_reserved_words()
...
+		</div>
+		<br class="clear" />';
```
XinYenFon: this should be unnecessary as well at this stage

XinYenFon (changes requested):

SychO9 (commented):
I also believe the `requirePolicyAgreement` needs to be set to 1 by default on installation, i.e make sure the value is inserted to the settings table in the sql files `install_2-1_mysql` and `install_2-1_postgresql` just like `requireAgreement`

XinYenFon:
You forgot the closing tags ,🙂

Themes/default/Register.template.php
MisterVector: Actually, this appears in a lot of other places in the SMF 2.1 files (without the trailing slash). Not sure I should touch those as they don't fall within the scope of this PR.

Sources/ManageRegistration.php
```diff
@@ -292,20 +322,33 @@ function SetReserved()
...
-		'',
-
-		array('int', 'coppaAge', 'subtext' => $txt['zero_to_disable'], 'onchange' => 'checkCoppa();', 'onkeyup' => 'checkCoppa();'),
+	'',
+		array('check', 'requireAgreement', 'text_label' => $txt['admin_agreement'], 'value' => 1, 'disabled' => 1),
+		array('warning', empty($agreement) ? 'error_no_agreement' : ''),
+		array('check', 'requirePolicyAgreement', 'text_label' => $txt['admin_privacy_policy'], 'value' => 1, 'disabled' => 1),
+		array('warning', empty($policy) ? 'error_no_privacy_policy' : ''),
+	'',
+		array('int', 'coppaAge', 'subtext' => $txt['setting_coppaAge_desc'], 'onchange' => 'checkCoppa();'),
```
SychO9: Instead of creating a new text string, please preserve the previous coppaAge setting definition line that uses `$txt['zero_to_disable']`

Sources/ManageRegistration.php
```diff
@@ -292,20 +322,33 @@ function SetReserved()
...
 	require_once($sourcedir . '/ManageServer.php');
 
+	// Do we have at least default versions of the agreement and privacy policy?
+	$agreement = file_exists($boarddir . '/agreement.' . $language . '.txt') || file_exists($boarddir . '/agreement.txt');
+	$policy = !empty($modSettings['policy_' . $language]);
+
 	$config_vars = array(
 		array('select', 'registration_method', array($txt['setting_registration_standard'], $txt['setting_registration_activate'], $txt['setting_registration_approval'], $txt['setting_registration_disabled'])),
+		array('check', 'enableOpenID'),
+		array('check', 'notify_new_registration'),
```
Sesquipedalian: Please remove these two lines. Those settings no longer exist in SMF 2.1.

Sources/ManageRegistration.php
Sesquipedalian: Yes, please restore the previous version of this line:
```
array('int', 'coppaAge', 'subtext' => $txt['zero_to_disable'], 'onchange' => 'checkCoppa();', 'onkeyup' => 'checkCoppa();'),
```

Sources/ManageRegistration.php
```diff
@@ -292,20 +322,33 @@ function SetReserved()
...
+		array('check', 'requirePolicyAgreement', 'text_label' => $txt['admin_privacy_policy'], 'value' => 1, 'disabled' => 1),
+		array('warning', empty($policy) ? 'error_no_privacy_policy' : ''),
+	'',
+		array('int', 'coppaAge', 'subtext' => $txt['setting_coppaAge_desc'], 'onchange' => 'checkCoppa();'),
 		array('select', 'coppaType', array($txt['setting_coppaType_reject'], $txt['setting_coppaType_approval']), 'onchange' => 'checkCoppa();'),
 		array('large_text', 'coppaPost', 'subtext' => $txt['setting_coppaPost_desc']),
 		array('text', 'coppaFax'),
 		array('text', 'coppaPhone'),
+	'',
+		array('check', 'announcements_default', 'disabled' => 1, 'value' => 0),
```
Sesquipedalian: This setting no longer exists, because SMF 2.1 always gives the user the ability to opt out of receiving announcement emails, so please remove these two lines.

Sources/ManageRegistration.php
```diff
@@ -342,7 +342,7 @@ function ModifyRegistrationSettings($return_config = false)
 		array('check', 'requirePolicyAgreement', 'text_label' => $txt['admin_privacy_policy'], 'value' => 1, 'disabled' => 1),
 		array('warning', empty($policy) ? 'error_no_privacy_policy' : ''),
 	'',
-		array('int', 'coppaAge', 'subtext' => $txt['setting_coppaAge_desc'], 'onchange' => 'checkCoppa();'),
+		array('int', 'coppaAge', 'subtext' => $txt['zero_to_disable'], 'onchange' => 'checkCoppa();'),
```
SychO9, line 337: also delete the new string you added since it's no longer needed 👍

Sources/ManageRegistration.php
```diff
@@ -328,6 +371,11 @@ function ModifyRegistrationSettings($return_config = false)
 		// Post needs to take into account line breaks.
 		$_POST['coppaPost'] = str_replace("\n", '<br>', empty($_POST['coppaPost']) ? '' : $_POST['coppaPost']);
 
+		// GDPR requires these settings to have certain values
+		$_POST['requireAgreement'] = 1;
+		$_POST['requirePolicyAgreement'] = 1;
+		$_POST['announcements_default'] = 0;
```
Sesquipedalian: Hm. Simply forcing these settings to always have a certain value doesn't make a lot of sense overall. 

On the one hand, if we do decide to always force these settings to have certain values, we would do better to remove the settings entirely and simply hardcode the behaviour that they control. On the other hand, I'm not sure that removing these settings entirely is actually what we want to do, either. Some of the behavioural differences between 2.0.x and 2.1 mean that we might still need these settings even in a "GDPR compliant by default" design.

So, at least for the time being, these lines should be removed.

Sources/Register.php
```diff
@@ -209,6 +225,9 @@ function Register($reg_errors = array())
 		'notify_announcements' => !empty($_POST['notify_announcements']) ? 1 : 0,
 	);
 
+	$context['announcements_ask'] = true;
+	$context['notify_announcements'] = isset($_POST['notify_announcements']) ? (bool) $_POST['notify_announcements'] : !empty($modSettings['announcements_default']);
+
```
Sesquipedalian: These lines are unneeded. `$context['announcements_ask']` doesn't exist in 2.1, and `$context['notify_announcements']` is handled elsewhere in Register.php.

Sources/Register.php
```diff
@@ -429,6 +448,9 @@ function Register2()
 		$_POST['options'] = isset($_POST['options']) ? $_POST['options'] + $_POST['default_options'] : $_POST['default_options'];
 	$regOptions['theme_vars'] = isset($_POST['options']) && is_array($_POST['options']) ? $_POST['options'] : array();
 
+	// Note when they accepted the agreement and privacy policy
+	$regOptions['theme_vars']['agreement_accepted'] = $regOptions['theme_vars']['policy_accepted'] = time();
```
Sesquipedalian: This should be changed to:
```suggestion
	if (!empty($modSettings['requireAgreement']))
		$regOptions['theme_vars']['agreement_accepted'] = time();
	if (!empty($modSettings['requirePolicyAgreement']))
		$regOptions['theme_vars']['policy_accepted'] = time();
```
(Note to self and other dev team members: We should probably change that in 2.0.x, too.)

Themes/default/Register.template.php
```diff
@@ -166,6 +165,15 @@ function verifyAgree()
...
 
+	if (!empty($context['announcements_ask']))
+		echo '
+					<dl class="register_form" id="notify_announcements_group">
+						<dt><strong><label for="notify_announcements">', $txt['notify_announcements'], ':</label></strong></dt>
+						<dd>
+							<input type="checkbox" name="notify_announcements" id="notify_announcements" tabindex="', $context['tabindex']++, '" class="input_check"', (!empty($context['notify_announcements']) ? ' checked="checked"' : ''), ' />
+						</dd>
+					</dl>';
+
```
Sesquipedalian: This can be removed. A checkbox for this is already included here: https://github.com/SimpleMachines/SMF2.1/blob/c514281ad04c910a9afd31057dea7054e05be92d/Themes/default/Register.template.php#L137-L144

Themes/default/Register.template.php
```diff
@@ -560,6 +568,18 @@ function template_admin_register()
...
+						</dt>
+						<dd>
+							<input type="checkbox" name="requireAgreement" id="requireAgreement_check" tabindex="', $context['tabindex']++, '" checked="checked" disabled="disabled" class="input_check" /><input type="hidden" name="requireAgreement" value=1>
+						</dd>
+						<dt>
+							<strong><label for="requirePolicyAgreement">', $txt['admin_register_require_policy'], ':</label></strong>
+						</dt>
+						<dd>
+							<input type="checkbox" name="requirePolicyAgreement" id="requirePolicyAgreement_check" tabindex="', $context['tabindex']++, '" checked="checked" disabled="disabled" class="input_check" /><input type="hidden" name="requirePolicyAgreement" value=1>
+						</dd>
```
Sesquipedalian: I think we can leave out these checkboxes in 2.1, and instead just hard code the behaviour in Register.php so that newly created users will always be presented with the agreement acceptance form when they first log in.

Themes/default/languages/Errors.english.php
```diff
@@ -426,6 +426,13 @@
 $txt['search_api_missing'] = 'The search API could not be found. Please contact the admin to check they have uploaded the correct files.';
 $txt['search_api_not_compatible'] = 'The selected search API the forum is using is out of date - falling back to standard search. Please check file %1$s.';
 
+// Registration Agreement
+$txt['error_no_agreement'] = 'There is no registration agreement to display!';
+$txt['error_no_privacy_policy'] = 'A privacy policy has not been created for this forum.';
+
+// Unsubscribe
+$txt['unsubscribe_invalid'] = 'The unsubscribe link that brought you here does not appear to be valid.';
+
```
Sesquipedalian, line 436: This can be removed. It already exists elsewhere in the file.

Themes/default/languages/Help.english.php
```diff
@@ -634,5 +634,9 @@
...
 $helptxt['languages_recaptcha'] = 'The language code to use for the reCAPTCHA verification widget.<br><br>See Google\'s reCAPTCHA documentation for its <a href="https://developers.google.com/recaptcha/docs/language" target="_blank" class="bbc_link">list of supported languages</a>.';
+$helptxt['allow_disableAnnounce'] = 'This will allow users to opt out of notification of topics you announce by checking the &quot;announce topic&quot; checkbox when posting.<br /><br />This setting must be enabled in order to comply with the rules of the <a href="https://ec.europa.eu/commission/priorities/justice-and-fundamental-rights/data-protection/2018-reform-eu-data-protection-rules_en" class="bbc_link">GDPR</a>.';
+$helptxt['announcements_default'] = 'This simply controls whether the the checkbox on the registration form starts as checked or unchecked.<br /><br />This setting must be disabled in order to comply with the rules of the <a href="https://ec.europa.eu/commission/priorities/justice-and-fundamental-rights/data-protection/2018-reform-eu-data-protection-rules_en" class="bbc_link">GDPR</a>.';
```
Sesquipedalian: Please remove these two lines. The settings they refer to do not exist in 2.1.

Themes/default/languages/Login.english.php
```diff
@@ -91,6 +95,7 @@
 $txt['setting_coppaPost_desc'] = 'Only applies if age restriction is in place';
 $txt['setting_coppaFax'] = 'Fax number to which approval forms should be faxed';
 $txt['setting_coppaPhone'] = 'Contact number for parents to contact with age restriction queries';
+$txt['setting_announcements_default'] = 'Enable "' . $txt['notify_announcements'] . '" by default.';
```
Sesquipedalian: We don't need this line in 2.1. The relevant setting already has a UI elsewhere.

Themes/default/languages/Login.english.php
```diff
@@ -107,6 +112,8 @@
 $txt['admin_register_group_desc'] = 'Primary membergroup new member will belong to';
 $txt['admin_register_group_none'] = '(no primary membergroup)';
 $txt['admin_register_done'] = 'Member %1$s has been registered successfully!';
+$txt['admin_register_require_agreement'] = 'Require user to accept the registration agreement';
+$txt['admin_register_require_policy'] = 'Require user to accept the privacy policy';
```
Sesquipedalian: Removing the checkboxes in `template_admin_register()` (see above) will also mean that these lines can be removed.

Themes/default/languages/Profile.english.php
```diff
@@ -533,6 +533,12 @@
 $txt['trackEdit_action_id_group'] = 'Primary Membergroup';
 $txt['trackEdit_action_additional_groups'] = 'Additional Membergroups';
 
+// Registration Agreement
+$txt['trackEdit_action_agreement_accepted'] = 'Accepted the registration agreement';
+$txt['trackEdit_action_policy_accepted'] = 'Accepted the privacy policy';
+
+$txt['export_profile_data'] = 'Download profile data';
```
Sesquipedalian: This can be removed. UI strings for the data export function are covered elsewhere.

Sesquipedalian (changes requested):
Thanks for your work on this, @MisterVector. Much appreciated.

This is an initial review, based only on reading through the code. Once these requested changes have been implemented, I'll run tests and then give a second review.

Sources/ManageRegistration.php
MisterVector: Which string specifically?

MisterVector: @SychO9 if you're referring to the `setting_coppaAge_desc` text string, that has already been removed in a different commit.

MisterVector:
@Sesquipedalian you're good to go for a second review!

MisterVector:
@sbulen An earlier commit has addressed the `setting_coppaAge_desc` error you receive. Now as for the `log_type` issue, I'm not quite sure how to address that.

As for me, I'm getting this error:

image

I'm not sure how to properly create and check the "admin-rega" token in ManageRegistration.php::EditAgreement(). I tried placing that code in respect to how the "admin-regp" token is created and checked in the EditPrivacyPolicy() function in the same file.

live627:
This is what I get

( ! ) Parse error: syntax error, unexpected ''" />' (T_ENCAPSED_AND_WHITESPACE) in C:\wamp64\www\SMF2.1\Themes\default\Register.template.php on line 764

Themes/default/Register.template.php
```diff
@@ -701,4 +706,72 @@ function template_edit_reserved_words()
...
+						<script>
+							function resetPolicyConfirm()
+							{
+								return true;
+							}
+						</script>
+					</div>
+				</form>
+			</div>
+		</div>
```
live627: yooo

Sources/ManageRegistration.php
```diff
@@ -292,16 +322,25 @@ function SetReserved()
...
 		array('select', 'registration_method', array($txt['setting_registration_standard'], $txt['setting_registration_activate'], $txt['setting_registration_approval'], $txt['setting_registration_disabled'])),
 		array('check', 'send_welcomeEmail'),
-		'',
-
-		array('int', 'coppaAge', 'subtext' => $txt['zero_to_disable'], 'onchange' => 'checkCoppa();', 'onkeyup' => 'checkCoppa();'),
+	'',
+		array('check', 'requireAgreement', 'text_label' => $txt['admin_agreement'], 'value' => 1, 'disabled' => 1),
+		array('warning', empty($agreement) ? 'error_no_agreement' : ''),
+		array('check', 'requirePolicyAgreement', 'text_label' => $txt['admin_privacy_policy'], 'value' => 1, 'disabled' => 1),
+		array('warning', empty($policy) ? 'error_no_privacy_policy' : ''),
```
live627: why are these always disabled?

Sources/Logging.php
```diff
@@ -501,6 +501,10 @@ function logActions($logs)
 			$smcFunc['db_free_result']($request);
 		}
 
+		$always_log = array('agreement_accepted', 'policy_accepted', 'agreement_updated', 'policy_updated');
+		if ((empty($modSettings['modlog_enabled']) && !in_array($action, $always_log)) || !isset($log_types[$log_type]))
```
live627, line 505: Notice: Undefined variable: log_type in C:\wamp64\www\SMF2.1\Sources\Logging.php on line 505

Sources/ManageRegistration.php
MisterVector: Those values in SMF 2.0.16 were based on `!empty($modSettings['force_gdpr'])` but since that setting doesn't exist anymore, these values are being treated as if `force_gdpr` was always true

Themes/default/Register.template.php
MisterVector: Not sure how that happened!

Sources/ManageRegistration.php
live627: Then you have it backwards. You can remove the disabled part I think.

Sources/Modlog.php
```diff
@@ -58,9 +60,11 @@ function ViewModlog()
 
 		$smcFunc['db_query']('', '
 			DELETE FROM {db_prefix}log_actions
-			WHERE id_log = {int:moderate_log}',
+			WHERE id_log = {int:moderate_log}
+			AND action NOT IN ({array_string:uneditable})'),
```
live627: A parse error occurred: Syntax error, unexpected ',' on line 64 at column 50

Oldiesmann: You've got an extra ) there before the comma

Sources/ManageSettings.php
```diff
@@ -54,6 +54,7 @@ function ModifyFeatureSettings()
 	loadLanguage('ManageSettings');
 
 	$context['page_title'] = $txt['modSettings_title'];
+	$context['show_privacy_policy_warning'] = empty($modSettings['policy_' . $language]);
```
live627, line 57: Missing globals: `$modSettings` and `$language`

Sources/ManageRegistration.php
MisterVector: I think they wanted the agreement always required, but I might be wrong.

MisterVector: @Sesquipedalian any input here?

Sources/Modlog.php
```diff
@@ -368,6 +375,9 @@ function list_getModLogEntries($start, $items_per_page, $sort, $query_string = '
 
 	$modlog_query = allowedTo('admin_forum') || $user_info['mod_cache']['bq'] == '1=1' ? '1=1' : (($user_info['mod_cache']['bq'] == '0=1' || $ignore_boards) ? 'lm.id_board = 0 AND lm.id_topic = 0' : (strtr($user_info['mod_cache']['bq'], array('id_board' => 'b.id_board')) . ' AND ' . strtr($user_info['mod_cache']['bq'], array('id_board' => 't.id_board'))));
 
+	if (!isset($context['uneditable_actions']))
+		$context['uneditable_actions'] = array();
```
live627, line 379: Missing global: $context

Themes/default/Agreement.template.php
```diff
@@ -0,0 +1,88 @@
...
+		elseif (!empty($context['agreement_accepted_date']))
+		{
+			echo '
+		<div class="description">
+			', sprintf($txt['agreement_accepted'], timeformat($context['agreement_accepted_date'], false)), '
+		</div>';
+		}
+
+		echo '
+		<div class="roundframe">
```
live627: I suggest both description  boxes to have the class `information noup` ans this main one `windowbg noup`

Screenshot_2020-05-05 Privacy Policy

XinYenFon:
I sent a small PR (https://github.com/MisterVector/SMF2.1/pull/1) to resolve logging related issue, updated the template and added agreement.txt to gitignore. Will be listed here after @MisterVector merges it.

XinYenFon:
Also forgot to mention but, upgrader needs to check if user has `requirePolicyAgreement ` set & 1, user can be coming from old version of SMF where requirePolicyAgreement does not set or not even exist which creates issues (like skipping the agreement part entirely).

Sources/ManageRegistration.php
XinYenFon: Both Agreements are forced in 2.x

XinYenFon (approved):

Sources/ManageRegistration.php
MisterVector: Therefore I think this conversation can be resolved then.

MisterVector:
Working on the upgrade scripts to set `requireAgreement` and `requirePolicyAgreement` to 1. Looking at PostgreSQL for how to replace a value, I find that `In PostgreSQL 9.5, the ON CONFLICT clause was added to INSERT`. I noticed that SMF 2.1 requires PostgreSQL 9.4, but with the addition of this statement, the version would be bumped to 9.5. Is that fine?

SychO9:
@MisterVector I believe you can just do this and not worry about those details
```php
$smcFunc['db_insert']('replace',
	'{db_prefix}settings',
	array('variable' => 'string', 'value' => 'string'),
	array('requirePrivacyPolicy', '1'),
	array('variable')
);
```

However I would wait for one of the developers to confirm this is correct as I haven't dealt with the upgrader before.

MisterVector:
I have added what I think would work for the MySQL upgrade scripts. I haven't included the PostgreSQL scripts yet. I'll wait for word on that.

MisterVector:
@XinYenFon is there a reason you added agreement.txt to .gitignore when you PR'd to my pull request branch? Was that an accident?

XinYenFon:
it is required otherwise if someone edits their agreement it will show as new file or changed file. Ping @Sesquipedalian to fix build tools...

MisterVector:
@XinYenFon That's true actually, we wouldn't want people committing their license files on accident...

@Sesquipedalian probably can skip it for now, but have you thought about inserting the agreement text into the database?

MisterVector:
CI actually hasn't built since the first commit to this PR. Once the syntax issues were taken care of, the remaining issue has been with build tools not finding agreement.txt.

live627:
- `requireAgreement` is present since v1.1
- `requirePolicyAgreement` probably should not be set since SMF does not ship with a privacy policy
- The two new checkboxes added in the admin shouldn't be permanently disabled because that defeats the purpose of them
- Along the same vein, those two checkboxes always are checked, which again, defeats the purpose.

Sesquipedalian:
The build tools are fine. 🙂 The Travis CI check is failing because the license blocks in ./Sources/Agreement.php and ./Themes/default/Agreement.template.php are missing some bits. The correct format is this:
```
/**
 * <<description of file's purpose>>
 *
 * Simple Machines Forum (SMF)
 *
 * @package SMF
 * @author Simple Machines https://www.simplemachines.org
 * @copyright 2020 Simple Machines and individual contributors
 * @license https://www.simplemachines.org/about/smf/license.php BSD
 *
 * @version 2.1 RC2
 */
```
Currently, the two files are missing the URL in the author line and `and individual contributors` in the copyright line.

XinYenFon:
@live627 true but writing a Privacy Policy does not trigger those database values to 1 in some cases. I agree some things are logically wrong and I do think those settings should not be forced on users.

MisterVector:
Wondering what is left to do. I don't yet have the queries set for the PostgreSQL upgrade scripts. @sbulen I have hopefully addressed your errors (and with the PR @XinYenFon sent, the `log_type` issue). I haven't yet figured out the error I get with `admin-rega_token` though.

Sesquipedalian:
I'm working on testing this and putting together some revisions that I will send to you, @MisterVector. As you may have noticed, my time has been constrained lately, so it might still be a few days yet before I can finish that. Your work is very much appreciated. :)

MisterVector:
Yeah I noticed you've been focusing on another important PR. Take your time man and eventually we'll get it sorted out!

MissAllSunday:
Hi @MisterVector The changes looks good codewise. Thanks!  the only minor change would be to change the direct checks against vars for empty() checks. This is mainly due to SMF historically prefers empty() checks over direct on var ones.

MissAllSunday:
Been testing this one, so far looks good, thanks!

Been wondering if you could add an info notice similar to the one an admin gets when s/he edits the Registration Agreement, this will help in getting an indication if their changes were indeed saved.

MisterVector:
image

@MissAllSunday please advise as this hasn't yet been added to the postgresql upgrade scripts. The install scripts have already been covered.

other/upgrade_2-1_mysql.sql
```diff
@@ -51,6 +51,13 @@ CREATE TABLE IF NOT EXISTS {$db_prefix}member_logins (
 ) ENGINE=MyISAM;
 ---#
 
+---# SMF 2.1 requires registration and policy agreements to be enabled so they must be set to 1.
+REPLACE INTO {$db_prefix}settings
+	(variable, value)
+VALUES ('requireAgreement', '1'),
+	   ('requirePolicyAgreement', '1');
+---#
+
```
jdarwood007: If you do this, they run upgrade script once, it adds it.  If they change the setting then run upgrade again, this will revert it.  We should only insert it if it doesn't exist.

If these are required and shouldn't be disabled, then it should be removed as a setting.

other/upgrade_1-0.sql
```diff
@@ -1953,6 +1952,11 @@ VALUES
 DELETE FROM {$db_prefix}settings
 WHERE variable = 'agreement'
 LIMIT 1;
+
+REPLACE INTO {$db_prefix}settings
+	(variable, value)
+VALUES ('requireAgreement', '1'),
+	   ('requirePolicyAgreement', '1');
```
jdarwood007: The first edit to upgrade is fine since the option is being removed, but we shouldn't otherwise touch the upgrade.

MissAllSunday:
> image
> 
> @MissAllSunday please advise as this hasn't yet been added to the postgresql upgrade scripts. The install scripts have already been covered.

As this is something that directly impacts the behavior of the forum as a whole, I had raised the concern to the rest of the team, once we made an agreement about it we can proceed further.

MisterVector:
I've had a lot of trouble installing SMF 2.1... couldn't do it on the last version of xampp that still used MySQL, and couldn't do it on the latest. Perhaps that's another thing to test for others to see if it's actually an issue for them. Perhaps I wouldn't have had an issue if I installed MySQL and PHP separately, but I don't know.

My path has typically involved installing SMF 2.0 and then later upgrading to SMF 2.1 in my PR branch.

live627:
SMF works  with mariadb (fork and drop in replacement of mysql)

MissAllSunday:
Hi @MisterVector The team has come to an agreement about leaving the requirePolicyAgreement setting off by default.

Sources/ManageRegistration.php
```diff
@@ -292,16 +322,25 @@ function SetReserved()
...
 	$config_vars = array(
 		array('select', 'registration_method', array($txt['setting_registration_standard'], $txt['setting_registration_activate'], $txt['setting_registration_approval'], $txt['setting_registration_disabled'])),
 		array('check', 'send_welcomeEmail'),
-		'',
-
-		array('int', 'coppaAge', 'subtext' => $txt['zero_to_disable'], 'onchange' => 'checkCoppa();', 'onkeyup' => 'checkCoppa();'),
+	'',
+		array('check', 'requireAgreement', 'text_label' => $txt['admin_agreement'], 'value' => 1, 'disabled' => 1),
+		array('warning', empty($agreement) ? 'error_no_agreement' : ''),
+		array('check', 'requirePolicyAgreement', 'text_label' => $txt['admin_privacy_policy'], 'value' => 1, 'disabled' => 1),
```
SychO9: > leaving the requirePolicyAgreement setting off by default.

In which case, I believe this needs to no longer be disabled, to allow users enabling it.

MisterVector:
@SychO9 I'll go back over all the changes in this PR and see what needs to be modified.

frandominguezl:
I'm currently testing this. When you click on "Sign Up", this is what you see:

image

After you click on "I ACCEPT THE TERMS OF THE AGREEMENT.", this is what you see:

image

Maybe this latter button should be placed in the first one, and change this latter to "REGISTER".

Also, if you do the change, please remove the final "." of each sentence for the buttons.

Thanks!

frandominguezl:
Another suggestion here. Maybe we could add some spacing between the button of "I AGREE" after changing the current privacy policy to a new one and logging into an account that needs to accept it to proceed navigation.

image

Thanks!

sbulen:
I noted two issues/questions during my testing:
 - I am not seeing user acceptance of the policies logged, maybe I'm looking in the wrong place? (in 2.0 they're in log_actions)
 - When an admin registers a user, that user is not required to accept the policies (in 2.0, the acceptance is still tracked & they will still have to agree before getting access to the forum, even if added by an admin)

Thank you for doing this!

MisterVector:
> 
> SMF works with mariadb (fork and drop in replacement of mysql)
> […](#)
> On Thu, Aug 6, 2020 at 1:46 PM Joel Paulien ***@***.***> wrote: I've had a lot of trouble installing SMF 2.1... couldn't do it on the last version of xampp that still used MySQL, and couldn't do it on the latest. Perhaps that's another thing to test for others to see if it's actually an issue for them. Perhaps I wouldn't have had an issue if I installed MySQL and PHP separately, but I don't know.

Previously I took the files off of master, copied them to a separate folder somewhere else, and installed SMF that way. Either I'd get an error with MariaDB when installing, or the installer would time out (odd issue).

I've just run through the same steps but instead installing off the fork itself, and everything turned out fine. So weird though!

sbulen:
It's a tangent here, but...  SMF does not formally support MariaDB.  There are discussions on issues over in the support forum on this.  Yes, installs sometimes have issues, due to conflicts between SMF db column names & MariaDB reserved words.  Also, their version #s have strayed, which can lead to other issues.  

You can usually work thru the issues, but it takes work...

Its a drag & drop replacement... usually...

MisterVector:
Hunting down another issue I found: if only the privacy policy is enabled, and a user signs up, they are not presented with the privacy policy.

sbulen:
I have confirmed that the "privacy policy only" issue was fixed.

The new user via admin account can still logon without accepting the policy, though.  There is a brief flicker on the screen like it tried to do something.

linux, php7.3

Sources/Logging.php
```diff
@@ -501,6 +501,10 @@ function logActions($logs)
 			$smcFunc['db_free_result']($request);
 		}
 
+		$always_log = array('agreement_accepted', 'policy_accepted', 'agreement_updated', 'policy_updated');
+		if ((empty($modSettings['modlog_enabled']) && !in_array($action, $always_log)) || !isset($log['log_type']))
```
live627: `$action` doesn't exist. Should be `$log['action']`

live627: Actually... incoming PR

live627: https://github.com/MisterVector/SMF2.1/pull/3

MisterVector: Does this fix the issue @sbulen reported about the agreement and policy acceptance not being logged?

sbulen:
OK, the very last thing I see here has to do with the upgrader: 

Short version:  The changes to the 1-0, 1-1 & 2-0 upgrade sql files are not needed and should be removed.  The 2-1 upgrade file only needs to add requirePolicyAgreement, as that is the only new setting being added in 2.1 (requireAgreement has been around since 1.0).  

Longer version:  Each upgrader step takes you from the PRIOR version to THAT version.  In theory, they should never be changed on subsequent versions (rarely - only when really old bugs are found...).  If the setting didn't exist back then, we don't need to add it retroactively.  

The upgrader will literally run every upgrader step from your "FROM" version to your "TO" version (yes, including the initial "FROM" version, to ensure DB completeness).  E.g., if you are going from 1.0 to 2.1, it will execute 1.0, 1.1, 2.0 & 2.1.  So we don't add new settings in all of the scripts.  Only the script for the version where the setting was added.  

The odd hiccup here is that although the new setting was added in 2.0, it was added by the GDPR switch in the app, not in the install or upgrade scripts...  In the 2.0 code, there are no changes in 1.0, 1.1 & 2.0 upgrade sql files.  The same should be true here.  We do need to ensure that single new setting (requirePolicyAgreement) is covered in the 2.1 installs & 2.1 upgrade.

sbulen:
> Working on the upgrade scripts to set `requireAgreement` and `requirePolicyAgreement` to 1. Looking at PostgreSQL for how to replace a value, I find that `In PostgreSQL 9.5, the ON CONFLICT clause was added to INSERT`. I noticed that SMF 2.1 requires PostgreSQL 9.4, but with the addition of this statement, the version would be bumped to 9.5. Is that fine?

MisterVector - I just took a peek and the ON CONFLICT DO NOTHING is used throughout the 2-1 postgresql upgrade script.  So much for 9.4 compatibility...  We'll have to update our documentation.  

You should emulate one of the existing lines & you'll be fine.   

As noted in my earlier comment, you only need to worry about the new item, requirePolicyAgreement.

sbulen:
Upgrader changes look good - thanks!

@live627 pointed out that the ON CONFLICT DO NOTHING actually receives special attention in the upgrader for pg.  So, the 9.4 minimum version still holds.  If interested, take a peek here:
https://github.com/SimpleMachines/SMF2.1/blob/6a55d9242aecd5f42abba349b488eaada71796e4/other/upgrade.php#L2222

I'll do further testing.

Some other background you may find helpful:  The upgrader has also been used over the years to help _repair_ damaged DBs.  You should be able to re-run the upgrader on a forum without breaking it or losing settings.  

You will see in upgrade_query() that duplicate table errors, duplicate column errors, duplicate row errors, deleting a column that doesn't exist, etc., are all ignored.  HOWEVER...  Also note that, in the event of critical settings/columns missing, the upgrader can be used to fix that.  It will fill in the gaps & get the DB to where it needs to be.

sbulen:
@live627 I'm getting this today, hadn't seen it before.  Looks like you removed the definition & added a check for it?

image

sbulen:
Migrating from 2.0 to 2.1, the non-English language privacy policies, when -utf8, did not make it over.  

The problem is that the languages in 2.0 may or may not have had a -utf8 suffix.  In 2.1, they are all implicitly -utf8 already...  

Post conversion here - spanish isn't found, because it doesn't have -utf8 at the end in 2.1:
image

There are 2.1 upgrader steps to change languages to strip the -utf8 from the language strings.  I think we need one here, too.  A very safe, non-destructive one... E.g., if setting "policy_\*-utf8" exists, and "policy_\*" does NOT exist, rename, otherwise do nothing.  Remember we should be able to safely rerun it.

This would get the 'updated_' history strings as well.

Might be interesting, I don't think I've seen a setting key modified before?  A delete/add may be needed?  If you need help, let me know, I'll take a stab at it.

sbulen:
For that matter, the registration agreement (the *.txt file) has the same issue...  This isn't tied to this PR, so I will write that up separately.

MisterVector:
I won't be here for the next week, so perhaps you can take a stab at the privacy policy stuff @sbulen if you want.

sbulen:
Will do.

sbulen:
Minor question/issue: I am pretty sure the notify_tokens setting is not used in 2.1.  It is in 2.0 only.  Token based unsubscription is default behavior in 2.1, so no setting is needed.  

I saw a stray reference in there to notify_tokens that I am pretty sure must be removed???

sbulen:
> For that matter, the registration agreement (the *.txt file) has the same issue... This isn't tied to this PR, so I will write that up separately.

Followup note...  After further testing, I decided against writing up an issue for this.  Unlike the privacy policies, the flat file agreements are in the language packs, which need to be reinstalled after the 2.0=>2.1 upgrade.  Doing so overwrites your existing files, _including the agreement..._  Leaving them alone (with the -utf8 suffix) actually makes it easier to recover it if needed...  Renaming them increases the likelihood they're overwritten...

I think I'll leave this as-is.  If you have a custom utf8 reg agreement, you will need to rename it manually after the upgrade.  If you have a custom non-utf8 agreement, you better not overwrite it when you install the language pack...  All in all, it seems safer as-is.  

Open to input...

sbulen:
@live627 - have you had a chance to look at the $always_log issue above?

sbulen:
Thanks @live627 !

All issues I've seen above have been addressed.  The only oddity is that policy acceptance logged in 2.0 isn't visible, but that appears to be a 2.0 issue...  Policy acceptance in 2.1 is getting logged properly & is visible under the profile edit log - but only AFTER profile edit logging is enabled.

live627:
oh, it should be logged regardlesss

live627:
I have only the agreement for my user  to accept and if I use the linktree to go to the `agreement`  action, it then shows both documents.

I need to fix this  oddity...

live627:
Seems to me that the subaction juggling can be re moved. Unless the intent was to allow the user to view either one at will. But there's no easily accessible links for that.

sbulen:
I think the only need there is to do one, the other, or both.  Whatever has been updated.

live627:
Yesh, manual URL overrides shouldn't be possible.

sbulen:
> oh, it should be logged regardlesss

Yes, I think you're right.

sbulen:
@live627 - could you take a look at these?

2 issues:
 - New user registrations are missing id_member in log_actions.  It was bugging me that I sometimes saw it blank & sometimes not.  The problem is NEW users (this is an issue in 2.0 also).  
 - This is new as of today, attempting a new user registration:

image

sbulen:
More info:
image

live627:
... this was with the profile edits log disabled, right?

Would make this much ampler if the logs could never be disabled, but hey  complexity is our old friend and people want options, right?  RIGHT? So I've been told. Developers don't know much about the human psyche. (No grudges here! Ugh...))

sbulen:
@live627, I think all the issues noted above have been dealt with:
 - No more id_log issues
 - Policy acceptance is logged whether or not the profile edit log is enabled
 - id_member is now populated upon new user registration, so the initial policy acceptance is in the log

The only oddity I am seeing now is that certain activities are flagged as having been performed by a "deleted member".  This is on all of the new user registrations:
image

sbulen:
@live627 - Good news! The policy is fixed.  

But I don't know how this didn't appear before, but now I'm getting the same thing for "gender"...  It's not even on my earlier screenshot above...(???):
image

sbulen:
Note the companion PR, #6231 , now has 2 upgrader fixes related to the privacy policy:
 - strips "-utf8" from the policy setting variable names to avoid losing your policies upon migration from 2.0 =>2,1
 - fixes gaps in id_member on log_actions, correcting data caused by the 2.0 reg bug, so the initial reg & policy acceptances are visible in the logs

MisterVector:
You guys made a lot of progress since I went on my trip!  I’m still on vacation but figured I’d check in.  From your perspective sbulen what all is left?

live627:
That error with the gender field looks like an oddity with custom   field
processing. Similar code paaths; that may be why the symptom  is similar.

sbulen:
I have just confirmed that the 'deleted member' issue on the Gender record existed prior to this PR.  I will write that up separately.

So...  **_I do think all issues have been dealt with._**  I suggest one last round of testing by everyone to confirm recent changes haven't introduced something.  

Almost there!

Sources/ManageSettings.php
```diff
@@ -227,6 +228,7 @@ function ModifyBasicSettings($return_config = false)
 		'',
 
 		// Option-ish things... miscellaneous sorta.
+		array('check', 'notify_tokens', 'disabled' => 1),
```
SychO9: what's this setting for ? it doesn't have a label in the `Features and Options` admin page, but the disabled checkbox can be seen.

other/upgrade_1-0.sql
```diff
@@ -1920,8 +1920,8 @@ upgrade_query("
 		('defaultMaxTopics', '" . (empty($GLOBALS['maxdisplay']) ? 20 : $GLOBALS['maxdisplay']) . "'),
 		('defaultMaxMembers', '" . (empty($GLOBALS['MembersPerPage']) ? 20 : $GLOBALS['MembersPerPage']) . "'),
 		('time_offset', '" . (empty($GLOBALS['timeoffset']) ? 0 : $GLOBALS['timeoffset']) . "'),
-		('cookieTime', '" . (empty($GLOBALS['Cookie_Length']) ? 60 : $GLOBALS['Cookie_Length']) . "'),
 		('requireAgreement', '" . @$GLOBALS['RegAgree'] . "')");
+		('cookieTime', '" . (empty($GLOBALS['Cookie_Length']) ? 60 : $GLOBALS['Cookie_Length']) . "')");
```
SychO9: why move this ? I think the line 1923 would need the `;` to be changed to `,`

Sources/ManageSettings.php
sbulen: Correct...  This needs to go.

It's a setting in 2.0 that doesn't apply to 2.1.  (Whether to use no-logon-necessary unsub tokens on emails...)

other/upgrade_1-0.sql
sbulen: I think you are correct here, too...

Themes/default/Register.template.php
```diff
@@ -300,7 +313,7 @@ function verifyAgree()
 				<input type="submit" name="accept_agreement_coppa" value="', $context['coppa_agree_below'], '" class="button">';
 	else
 		echo '
-				<input type="submit" name="regSubmit" value="', $txt['register'], '" tabindex="', $context['tabindex']++, '" class="button">';
+				<input type="submit" name="regSubmit" value="', $txt['agreement' . (!empty($context['require_policy_agreement']) ? '_policy' : '') . '_agree'], '" class="button" />';
```
SychO9: I don't think we want this button editted, this is the `Signup` submit button, instead we want to edit the buttons with the name `accept_agreement` in this same template

live627:
Ahh the things I do to drop eight commits...

 Totally worth it for  a cleaner branch history.

Themes/default/css/index.css
```diff
@@ -3630,6 +3630,9 @@ h3.catbg .main_icons::before, h3.catbg .icon {
 .cat_bar + .title_bar {
 	margin-top: 0;
 }
+.cat_bar:not(:first-of-type) {
+	margin: 10px 0 0 0;
+}
```
SychO9: this doesn't seem to have any effect on the the agreements page, instead it just adds extra space everywhere else, I think it should be reverted

SychO9 (commented):
One last thing and this should be good to be merged (the setting I mentioned previously also has still to be fixed)

MisterVector:
I'm back! I'll see what I can do with the remaining issues here. Thanks for helping out @live627 !

Themes/default/css/index.css
live627: oh but it does
Untitled

live627:
@MisterVector np🙂

Themes/default/css/index.css
SychO9: no it doesn't, because the element above it has bottom margin, just try and remove these rules and you'll see

live627: huh, so it dues. SMF's highly complicated CSS is very confusing to me. At least in 2.0's CSS file I could find my way around!

MisterVector:
The button text for just the privacy policy enabled needs to be changed:

image

I'm not sure how to code the solution in the best manner possible so someone else want to take a stab at it?

Also I think I'm satisfied with testing.

live627:
Will do.

Thanks for the work you put in so far!

MisterVector:
This was kind of fun! Sooooooo glad it's almost finished!

live627:
me too, then we can move on  with preparing the release!

SychO9 (approved):
I believe this is finally good to go.

Thank you @MisterVector !

sbulen:
I retested everything & it looks good!

MissAllSunday (approved):

MissAllSunday:
Thank you everyone for your efforts!

PR #6101, opened 2020-05-04
Adds support for HTML and "Styled XML" as profile data export formats

### LINKED ISSUE #6102: CLI cron.php: Call to undefined function JavaScriptEscape() in Subs.php:2184
#### Description
I'm recently updated my forum to the GH version from this weekend, and my `cron`-triggered `cron.php` fails since:
```
PHP Fatal error:  Call to undefined function JavaScriptEscape() in /home/[…]/public_html/Sources/Subs.php on line 2184
```

### Steps to reproduce
1. Run `php […]/cron.php`
2. Notice you don't receivce forum notifications anymore, and check the error mails of the account running the forum ;)

- Version/Git revision: GH version from 2020-05-02
- Database Type: postgres but probably any
- Database Version: 9.4-ish
- PHP Version: 5.6-ish

Sesquipedalian:
The heart of this PR is the use of an XSLT stylesheet to transform an XML-based export file into HTML:
- On all servers, we can offer "Styled XML" format, which is an XML-based export file with an XSLT stylesheet embedded directly into it. Virtually all web browsers have built in XSLT support, so when the user opens a "Styled XML" file in a web browser, it will render just like an HTML file.
- On a server with a typical PHP configuration, the XML → HTML transformation can be performed on the server, so that we can offer the end user a nice, simple HTML file. If PHP was compiled without XSL support, the HTML export format option will be unavailable, but "Styled XML" will still work.

Since export files are supposed to be standalone files, this code also embeds some essential CSS and JavaScript into the HTML file (or "Styled XML" file, as the case may be). Getting this to work in the most optimal way was a bit tricky at first, but I think it is working nicely now.

Finally, this PR also tries to make it easy for the end user to acquire copies of all their attachments, while also avoiding any unnecessary strain on server resources. Basically, when the user opens the HTML/"Styled XML" export file, the browser will first look for attachments in the export file's local directory. If the attachment file exists in the local directory, great. Otherwise, we show a download link that retrieves the attachment file from the server and saves it on the user's machine.

It is worth noting that these download links point to `action=profile;area=dlattach;...` rather than to `action=dlattach;...`. These special download links call the new `export_attachment()` function in Profile-Actions.php. This function is essentially just a wrapper for the normal `showAttachment()` function, except that it makes a few tweaks to better support the export use case.

Also fixes #6102.


SychO9:
Just a minor note,

perhaps the default export format should be set to **XML** instead of **HTML** ?

Sesquipedalian:
> Just a minor note,
> 
> perhaps the default export format should be set to **XML** instead of **HTML** ?

I went back and forth on that myself, and I don’t have a strong feeling either way. On the one hand, the GDPR specifically mentions “machine readable” formats in its data portability requirements, which XML is better for than HTML. On the other hand, HTML is better for human readability, which is also part of the GDPR’s requirements. 

I’m open to arguments either way.

Sesquipedalian:
> > Just a minor note,
> > perhaps the default export format should be set to **XML** instead of **HTML** ?
> 
> I went back and forth on that myself, and I don’t have a strong feeling either way. On the one hand, the GDPR specifically mentions “machine readable” formats in its data portability requirements, which XML is better for than HTML. On the other hand, HTML is better for human readability, which is also part of the GDPR’s requirements.
> 
> I’m open to arguments either way.

Hm... The thought occurs to me that it might be a good idea to make XML-with-embedded-XSLT an option that the user can request explicitly, and perhaps even to make that the default option. If we did that, then our default would support both robust machine reading and easy human reading in the same file. We'd still provide the option to download only XML or only HTML (assuming the server has the XSLTProcessor class enabled, of course) in case the user needs only one of those.

SychO9:
> XML-with-embedded-XSLT

I'm afraid I wouldn't know, I lack knowledge in that department.

---

Upon testing, HTML exports seem to never work, this error might be related, keep in mind for some reason, many cron job errors don't make it to the smf error log, and you have to look at the server error log to see if all is well.

```log
[Thu May 07 10:58:18.296507 2020] [php7:notice] [pid 161864:tid 1212] [client 127.0.0.1:52847] PHP Fatal error:  Uncaught Error: Using $this when not in object context in ...\\Sources\\tasks\\ExportProfileData.php:387
Stack trace:
#0 ...\\Sources\\Subs.php(5067): ExportProfileData_Background::add_dtd()
#1 ...\\Sources\\News.php(344): call_integration_hook()
#2 ...\\tasks\\ExportProfileData.php(159): buildXmlFeed()
#3 ...\\Sources\\tasks\\ExportProfileData.php(315): ExportProfileData_Background->exportXml()
#4 ...\\Sources\\tasks\\ExportProfileData.php(59): ExportProfileData_Background->exportHtml()
#5 ...\\cron.php(233): ExportProfileData_Background->execute()
#6 ...\\cron.php(127): perform_task()
#7 {main}

thrown in ...\\Sources\\tasks\\ExportProfileData.php on line 387, referer: http://.../index.php?action=profile;area=getprofiledata;u=1
```

**PHP**: 7.4

Sesquipedalian:
> ```
> #0 ...\\Sources\\Subs.php(5067): ExportProfileData_Background::add_dtd()
> ```

Thanks. Should be fixed in ead22e6 and beyond.

SychO9:
Now I no longer get the HTML export option, my setup has **DOMDocument** available but not **XSLTProcessor** class, is that expected behavior ?

Also can we validate `$_POST['format']` by checking that it belongs to one of the available formats ? when I tried sending a random string to the receiving route, the error log went crazy.

SychO9:
Also getting this error when including all three types of data ans selecting styled xml format

```
Error
Type of error: Cron
Error message:
8: Undefined index: id_thumb
File: .../Sources/Subs-Attachments.php
Line: 987
URL of page causing the error: http://...index.phphttp://simplemachines.smf/cron.php

Backtrace information
#0: smf_error_handler_cron()
Called from ...\Sources\Subs-Attachments.php on line 987
#1: parseAttachBBC()
Called from ...\Sources\Subs.php on line 1412
#2: {closure}()
Called from ...\Sources\Subs.php on line 2991
#3: parse_bbc()
Called from ...\Sources\News.php on line 2119
#4: getXmlPosts()
Called from ...\Sources\tasks\ExportProfileData.php on line 175
#5: exportXml()
Called from ...\Sources\tasks\ExportProfileData.php on line 320
#6: exportHtml()
Called from ...\Sources\tasks\ExportProfileData.php on line 77
#7: execute()
Called from ...\cron.php on line 233
#8: perform_task()
Called from ...\cron.php on line 127
```

Sesquipedalian:
> Now I no longer get the HTML export option, my setup has **DOMDocument** available but not **XSLTProcessor** class, is that expected behavior ?

Yes, that's expected. Without XSLTProcessor, PHP can't transform the XML into HTML.

Prior to 8b3d650, `exportHtml()` would silently fall back to "Styled XML" format (i.e. XML with an embedded XSLT stylesheet) in that situation. You may not have noticed because the results were still served with an .html file extension even though the file content was XML rather than HTML. I initially figured that this approach would reduce user confusion, since users who want something they can read don’t care about the internal data structure. However, I've since discovered that at least some browsers get a bit confused when opening a local file with an .html extension that actually contains XML data, so that didn't work as well as I'd hoped.

Anyway, as of 8b3d650:
1. HTML and "Styled XML" are fully distinct format options, both to the human user and to machines.
2. Instead of offering HTML as an option and then silently (and somewhat misleadingly) falling back to "Styled XML" when HTML cannot not be provided due to a lack of XSLT support, the HTML option is now simply disabled in that situation.

> Also can we validate `$_POST['format']` by checking that it belongs to one of the available formats ? when I tried sending a random string to the receiving route, the error log went crazy.

Oops. I thought I was doing that. I must have missed a spot. Did that happen when you were initiating an export, or when you tried to download an export file?

> ```
> 8: Undefined index: id_thumb
> ```

That appears be an independent bug that this PR happens to have exposed. The fix should be easy, though, so I might just toss it into this PR.

SychO9:
Ah it all makes sense, that is a much better way of handling things indeed.

> Oops. I thought I was doing that. I must have missed a spot. Did that happen when you were initiating an export, or when you tried to download an export file?

When initiating an export, try sending a random string as a format (by editing one of the selectbox options value) you'll notice tens of errors whenever you refresh the page after that. (the only way to stop those errors is to manually delete the export files from the exports folder, deleting from the interface doesn't work in those circumstances)

SychO9:
That fixes the tens of errors, only leaves one single error

```
Type of error: Undefined
8: Undefined index: format
 http://.../index.php?action=profile;area=getprofiledata;u=1
.../Sources/Profile-Export.php (Line 398)
```

SychO9:
hmm, another problem is when you don't send a format value to the route, error log gets filled with tens of errors.

perhaps a
```php
if (!isset($_POST['format']))
    redirectexit(...);
```
would be better to avoid any similar issues ?

sbulen:
Is this ready to test?  

I get this upon export:
image

I had asked for everything, html format.

sbulen:
For one very basic user, only one post,  I asked for all 3 types of export & found this in the log:
image

Sesquipedalian:
30420b1 should solve the "Undefined variable: file" errors, @sbulen.

That "array_walk_recursive()" error has me perplexed. It is difficult to imagine how the multiple tests on lines 1772 through 1780 could all have passed and yet `$profile` could end up null on line 1782.

sbulen:
Have you requested all 3 in a row, same minimal user, all content, as described?  Pretty easy to reproduce on my end.

(My exact user is foreign language, one post, no PMs.  I was setting up to test #6107 .)

Interesting, the two backtraces are different, so there are apparently at least 2 paths:
image
image

Sesquipedalian:
> #6089 also aims to address this.

Fair enough. a7ade21 was added merely because I'd accidentally committed a change from `overflow:hidden` to `overflow:scroll` on the error_message span in #5081, so I needed to clean up my mess. I'm happy to let you deal with it in #6089 instead (but please see my comments there).

Sesquipedalian:
> Have you requested all 3 in a row, same minimal user, all content, as described? Pretty easy to reproduce on my end.

Sorry if I was unclear. It's not that I couldn't figure out how to reproduce it. It's that I couldn't figure out why it would happen. However, I now suspect this is another manifestation of #6111 that @live627 reported yesterday. I don't have time today to test a fix for it, but that's by far the most likely culprit.

live627:
the array walk error is what inspired #6111. It even broke my unit test!

sbulen:
The array walk error does not happen if I wait for one to complete before starting the next one.

It only happens when there are multiple requests active.

sbulen:
This issue still happens even after #6112 was merged, so it wasn't that...  

But I do think it is an issue with loadMemberContext() - or at least the way it is being used.  I believe the problem is that $dataLoaded is static, and it's supposed to tell you whether the user has been loaded already.  Problem is that I think $memberContext gets initialized somewhere along the line...

If you run them one at a time, no problem.  However, if you have a few queued up, first time thru everything works OK.  Second time thru, $dataLoaded is set, so it doesn't reload $memberContext for your requested user.  What you get is an array with a key of the user, but everything else is empty.  

$memberContext[$context['xmlnews_uid']] exists, but points to null...  This is what it looks like if you print_r it (user id is 2):
```
Array
(
    [2] => 
)

```
This passes all the tests at the start of getXmlProfile, so $profile is null.

sbulen:
One other minor issue - I am seeing a .tmp.bak left around in the file system sometimes after a successful request:
image

Sesquipedalian:
Regarding loadMemberContext(), my first thought is that even after #6112, the $dataLoaded array is actually useless. The logic as it currently stands would do better to simply check whether $memberContext[$user] is set, rather than futzing around with $dataLoaded[$user] at all. 

My second thought, however, is that the function does need a way to remember whether the custom profile fields were loaded for a given user or not, which it currently does not track even though it should. We could repurpose this useless static variable to track that info for us, thereby solving two problems at once.

Sesquipedalian:
> One other minor issue - I am seeing a .tmp.bak left around in the file system sometimes after a successful request:
> image

Oh, I probably just missed a bit somewhere when I ripped out the buggy code that tried to update old exports on the fly. It’ll probably be a two second fix once I get a chance to do it. 🙂

Sesquipedalian:
If you merge the changes from #6116 into this PR on your local machine, @sbulen, does it fix the issue?

sbulen:
Good news:  No 'stray' .bak files left around & no array walk errors.  Both of those seem to have been addressed.

After download, I cannot open the 'styled' xml in XML notepad, I get this error:
image

sbulen:
I am getting many hundreds of the following errors.  Certain users only:
image

I am also seeing hundreds of these:
image

sbulen:
Similar:
image

sbulen:
I am pretty sure the above are all from the html export, and only for certain users.  Might be volume-related - the user causing 90% of the errors logged is the problem one we had issues with before.

Sesquipedalian:
> Good news: No 'stray' .bak files left around & no array walk errors. Both of those seem to have been addressed.

Excellent.

> After download, I cannot open the 'styled' xml in XML notepad, I get this error:
> image

That appears to be a bug in your XML editor. Specifically, it looks like it is ignoring or mishandling the DOCTYPE Definition near the top of the document, and is therefore failing to load the embedded XSLT stylesheet.

In the most common scenario, an XSLT stylesheet is provided in a separate file rather than embedded within the XML data file that it applies to. But the DTD in these files defines an "id" attribute for XSLT stylesheets, which allows the href attribute in the `<?xml-stylesheet ... />` processing instruction to point to the embedded XSLT rather than to an external file. Unfortunately, your XML editor is still assuming `#stylesheet` must be an external file, which means it is either ignoring the DTD or else it doesn't simply know how to handle the idea of an href attribute that points to an internal element's id.

You _might_ be able to work around this by opening the *.styled.xml file in a plain text editor and moving the `<?xml-stylesheet ... />` below the DTD statement. In theory that shouldn't make a difference, but it possibly might, depending on how your XML editor implements its internal logic.

Sesquipedalian:
The `DOMDocument::load()` errors are probably all due to an improperly escaped character breaking the entire XML file. It looks like you have one of those improperly escaped characters on line 9077 of the 24_5e7e... file, and one on line 12551 of the 3_5e7e... file.

Could you provide me with copies of those two files so that I can inspect them, @sbulen? If I can see exactly what this invalid XML is, I should be able to figure out why it is being generated incorrectly and then fix it.

sbulen:
> The `DOMDocument::load()` errors are probably all due to an improperly escaped character breaking the entire XML file. It looks like you have one of those improperly escaped characters on line 9077 of the 24_5e7e... file, and one on line 12551 of the 3_5e7e... file.
> 
> Could you provide me with copies of those two files so that I can inspect them, @sbulen? If I can see exactly what this invalid XML is, I should be able to figure out why it is being generated incorrectly and then fix it.

If you PM me an email address over on smorg & I'll email them to you.

The posts with issues are public.  I'm pretty sure it's these two:
https://www.vguitarforums.com/smf/index.php?topic=3915.msg85024#msg85024
https://www.vguitarforums.com/smf/index.php?topic=6469.msg44977#msg44977

sbulen:
> That appears to be a bug in your XML editor. Specifically, it looks like it is ignoring or mishandling the DOCTYPE Definition near the top of the document, and is therefore failing to load the embedded XSLT stylesheet.

I can open the files in browsers, etc., and even in Excel, so I think you're correct...

Sesquipedalian:
> If you PM me an email address over on smorg & I'll email them to you.

Thanks again for those, @sbulen. Yeah, each of those posts contained an illegal control character that wasn't being stripped out due to a silly mistake on my part. I was using `array_walk` (or `array_walk_recursive`) to apply `cleanXml` to the various data arrays in News.php. But `array_walk($array, $callback)` only changes the values in `$array` if the callback function takes parameters by reference, and `cleanXml` does not. The method I'm using in fb96607 fixes the problem in my tests. Please give it a try on your machines to make sure it works for you, too.

Sesquipedalian:
> I can open the files in browsers, etc., and even in Excel, so I think you're correct...

I don't know if you plan to file a bug report with them or not, but if you do you can point them to the XSLT 1.0 specification's section on [embedding stylesheets](https://www.w3.org/TR/xslt-10/#section-Embedding-Stylesheets) to describe the expected behaviour.

sbulen:
Good news:  Down from 831 errors to 6.  The remaining one seems to be special character-related also...  
Down_to_6

sbulen:
On my 'large user' test (2GB of posts), I am getting this disk is full error.  BUT...  The disk is not at all full.  
Maybe the name is mangled somehow?  Circled in yellow looks suspicious...
table_full_err

sbulen:
Note for my 2GB user - back when testing #6101 I had to drop the export_rate down to 10 for this user's export to complete.  (1000 posts, 2MB each post)  The "disk full" failure happened when I requested all 3 exports for the same guy at the same time.  I am now trying them one-at-a-time, and so far, no error.  

So, the above happened with multiple massive exports going on simultaneously.  Specifically, with massive messages.

Sesquipedalian:
> Good news: Down from 831 errors to 6. The remaining one seems to be special character-related also...

Yeah, looks like it. My guess is that it is something that the `cleanXml` function should be catching but isn't. Can you email me a copy of 5_95bb8ee... file? I'll need both the HTML and styled XML formats again, please.

Sesquipedalian:
> On my 'large user' test (2GB of posts), I am getting this disk is full error. BUT... The disk is not at all full.
> Maybe the name is mangled somehow? Circled in yellow looks suspicious...

It looks like the database is trying to create a temporary table while processing the query, and the quantity of data is too much. Hmmm.....

The URL duplication is just a minor bug in the error log display that no one has ever bothered to track down and fix.

> Note for my 2GB user - back when testing #6101 I had to drop the export_rate down to 10 for this user's export to complete. (1000 posts, 2MB each post) The "disk full" failure happened when I requested all 3 exports for the same guy at the same time. I am now trying them one-at-a-time, and so far, no error.
> 
> So, the above happened with multiple massive exports going on simultaneously. Specifically, with massive messages.

Well, I suppose I could change the system to only allow one active export at a time, but that might be more of a band-aid than a real solution to the underlying problem.

... Honestly, I think I'd rather try to tackle this issue in a separate PR. Resource management is a pretty different problem than expanding the range of supported export file formats.

sbulen:
> Can you email me a copy of 5_95bb8ee... file? I'll need both the HTML and styled XML formats again, please.

Email sent with that info.

sbulen:
> The URL duplication is just a minor bug in the error log display that no one has ever bothered to track down and fix.

OK. 

> Well, I suppose I could change the system to only allow one active export at a time, but that might be more of a band-aid than a real solution to the underlying problem.

Agreed.  We don't want to single-thread those.  In fact, you STOPPED single-threading in the middle of #6101 when we realized that an error backed the whole system up...

> ... Honestly, I think I'd rather try to tackle this issue in a separate PR. Resource management is a pretty different problem than expanding the range of supported export file formats.

If we could, I'd suggest even simpler...  If it were at all possible to trap the error & provide a more friendly message to the user/admin that they may need to revisit the settings &/or submit large requests one-at-a-time.

I do think we have forums that share lots of articles - very large posts.  This particular example is going to be infrequent, but will definitely happen.   That kind of forum, e.g., article-oriented, may need to have a much smaller export_rate value.  

The support folks may need to know that this error message is a symptom of that.

Sesquipedalian:
> > Can you email me a copy of 5_95bb8ee... file? I'll need both the HTML and styled XML formats again, please.
> 
> Email sent with that info.

Thanks again. a782bec should have that fixed. As suspected, `cleanXml` was not filtering out all the characters it was supposed to.

sbulen:
Literally hundreds of thousands of errors & still going.
image

sbulen:
Literally hundreds of thousands of errors...
image

Sesquipedalian:
What the...?

Sesquipedalian:
Found the problem. Just trying to decide on the best fix. Should be available soon.

Sources/Load.php
```diff
@@ -1800,6 +1802,8 @@ function loadMemberContext($user, $display_custom_fields = false)
 
 	call_integration_hook('integrate_member_context', array(&$memberContext[$user], $user, $display_custom_fields));
 
+	$dcf[$user] = !empty($dcf[$user]) | $display_custom_fields;
```
MissAllSunday: Use a descriptive name for variables.

Sources/News.php
```diff
@@ -333,12 +338,15 @@ function buildXmlFeed($xml_format, $xml_data, $feed_meta, $subaction)
 	$forceCdataKeys = array();
 	$nsKeys = array();
 
+	// Maybe someone needs to insert a DOCTYPE declaration?
+	$dtd = '';
```
MissAllSunday: Same here, better use a descriptive name.

Sources/News.php
```diff
@@ -715,7 +722,7 @@ function getXmlMembers($xml_format, $ascending = false)
 	while ($row = $smcFunc['db_fetch_assoc']($request))
 	{
 		// If any control characters slipped in somehow, kill the evil things
-		array_walk($row, 'cleanXml');
+		$row = filter_var($row, FILTER_CALLBACK, ['options' => 'cleanXml']);
```
MissAllSunday: Avoid using short array syntax  on several filter_var() calls

Sources/News.php
```diff
@@ -2290,7 +2298,7 @@ function getXmlPosts($xml_format, $ascending = false)
 						'content' => array(
 							array(
 								'tag' => 'name',
-								'content' => $row['poster_name'],
+								'content' => !empty($user_info['name']) ? $user_info['name'] : $user_info['username'],
```
MissAllSunday: This will use the current user data, is this intended?  I'm thinking an scenario where an admin or any other user with the relevant permission will export someone elses data.

Themes/default/scripts/profile.js
```diff
@@ -293,4 +293,32 @@ function changeVariant(iThemeId, el)
 	document.getElementById('theme_thumb_' + iThemeId).src = oThemeVariants[iThemeId][el.value]['thumbnail'];
 	document.getElementById('theme_thumb_preview_' + iThemeId).href = el.form.action + ';theme=' + iThemeId + ';variant=' + el.value;
 	document.getElementById('theme_preview_' + iThemeId).href = el.form.action + ';theme=' + iThemeId + ';variant=' + el.value;
-}
\ No newline at end of file
+}
+
+$(document).on('change', '#export_format_select', function() {
+	var selected_format = $('#export_format_select').val();
+	if (completed_formats.includes(selected_format))
```
MissAllSunday: Array.prototype.includes  is ES6 syntax, we might have problems with old browsers compatibility.

MissAllSunday (dismissed):

Sources/Load.php
Sesquipedalian: Fair enough.

Sources/News.php
Sesquipedalian: This is descriptive. 🙂  DTD is the common abbreviation for Document Type Definition. It's even [used in the XML spec itself](https://www.w3.org/TR/xml/#sec-prolog-dtd).

Sources/News.php
Sesquipedalian: Good catch. I'll add some logic to make this decision more intelligently.

Sources/News.php
MissAllSunday: Yes, inside their context its descriptive, inside SMF's context it is not. Chances are someone looking at the code will not have knowledge of the XML spec to the point of easily identify its meaning.

Themes/default/scripts/profile.js
Sesquipedalian: Do you happen to have a quick suggestion for a backwards compatible alternative? 

(Don't spend a bunch of time answering this; I can do my own homework if I need to. But if you can toss off an answer in your sleep, I won't mind saving some time.)

MissAllSunday: Theres jquery's $.inArray  or vanilla indexOf

Sources/News.php
Sesquipedalian: Fair enough.

Themes/default/scripts/profile.js
Sesquipedalian: Thanks

Sesquipedalian:
>> Literally hundreds of thousands of errors...

> Found the problem. Just trying to decide on the best fix. Should be available soon.

Fixed in 7ba1303, @sbulen. Plus, the performance of cleanXml() has been improved. 🙂 Does it look good in your tests, too?

@MissAllSunday: Thanks for that. Look good to you now?

sbulen:
Good news: All prior errors reported are gone.

Issue 1: I cannot download the results when there are multiple files involved.  I get "network error" messages when I hit the "Download All" button.  The first file always works, but subsequent files do not.  Note the files are all listed in the Downloads browser tab, and may be resumed successfully; the browser resume function works.
image

Issue 2: Large export test (2GB of posts).  Strange "Resource temporarily unavailable" errors occur during large exports.  Several of these get posted in clumps.  The exports appear to continue properly - execution resumes; total volume seems right.
image

Old issue 3: Multiple large export test (2GB of posts).  I am still getting the erroneous disk full messages when I submit more than one large export at a time.  (This disk has ~300GB free...)  Things operate well when run one-at-a-time (other than issue 2 noted above).  Execution does appear to resume.
image

sbulen:
Issue 4:  One more...  I literally just submitted this & waited for the output.  It is telling me it is out of date, due to posts.  But there have been no new posts...
image

EDIT:  Actually, I think this one is still running...???  There is still a tmp file, and a background task record.
EDIT2:  It is in fact still running.  I believe the problem happens when the screen refreshes & there is no .tmp file out there.  You get the above message.  If you refresh the screen a few seconds later, it goes back to saying the export is in progress.  I just sat around refreshing the screen a bunch of times, & this comes & goes.

sbulen:
Issue 5:  Tried to open an export in Firefox & got this:
image

sbulen:
Reading up on each...
(1) above - Chrome downloads - may be local to my PC & specific to Chrome.  Apparently Chrome doesn't like multiple big downloads; this is an often-reported problem.  If I use another browser - OR - tell Chrome to prompt for each download, everything works.
(3) above - Disk full - as discussed before.  Very large average post sizes (articles) may need special care.   
(2), (4) & (5) above - seem to be real bugs.

Oldiesmann:
> Issue 5: Tried to open an export in Firefox & got this:
> image

This is because of the XML declaration being on the same line as the closing XSL stylesheet. It needs to be on its own line, with no whitespace or anything else before it.

sbulen:
A clarifying note on the disk full: It is not referring to running out of disk space.  It's trying to create a temp table for the result of a DB query.  This is an internal DB error saying there wasn't enough temp space to deal with the query.  If there is any way to scale the query back, e.g., by removing unnecessary columns or joins, that could help.

Sesquipedalian:
### Regarding issue (1)
I wonder if adding a short delay between the automated click events in the "Download all" JavaScript could allow us to work around the Chrome bug. 

### Regarding issues (2) and (3)
I think (2) and (3) are symptoms of PHP and MySQL respectively finding their available resources overwhelmed by the quantity of data being processed at once. From what I can discover, "failed to open stream: resource temporarily unavailable" can occur when PHP simply doesn't have the available capacity at the moment to open any more files. Meanwhile, the disk full messages are being generated by the database when it tries to create a temporary table and runs out of room for some reason. I don't know why that would be when you have 300GB of free disk space, but apparently MySQL thinks that's what's happening.

I honestly don't see any solution except to throttle down the rate at which you're exporting the data. Perhaps we could try to figure out some sort of adaptive throttle that would fetch fewer rows per query if it looked like server resources were getting strained.

What value have you set for “Rate at which to process posts & personal messages for export”, @sbulen?

### Regarding issues (4) ad (5)
Both of those sound like symptoms of something going wrong while renaming files and appending data. What happens if you comment out this block of code and run the test again?
```
		if (file_exists($tempfile) && filesize($tempfile) >= 1024 * 1024 * 250)
		{
			rename($tempfile, $realfile);
			$realfile = $export_dir_slash . ++$filenum . '_' . $idhash_ext;
		}
```

sbulen:
(1) Might be worth a try.  It really sounds like this is a common Chrome issue though.  

(2) Possibly also related to 4/5.  Most of what I read out there noted that this usually means php tries opening a file in the wrong read/write mode.  E.g., trying to append to a non-existing file.  

(3) DEFINITELY a Mysql issue with multiple parallel massive queries...  I have read a few articles on this & tried all of their suggestions, including changing temp table processing to MyISAM (it defaults to InnoDB, which is resource intensive) and jacking up temp table size values (which default to 16MB).  Nothing worked.
Best article I found was this one:
https://www.percona.com/blog/2019/07/17/mysql-disk-space-exhaustion-for-implicit-temporary-tables/#:~:text=The%20maximum%20size%20for%20in,you%20can%20increase%20the%20variables.

> What value have you set for “Rate at which to process posts & personal messages for export”, @sbulen?

I have tried values as low as 2...   I am shocked 2 didn't work...  I even put a trace in there to make sure it was really using the 2...

If you recall this particular user has 1000 posts, each 2M of text.  (This user was dummied up to test large output, a 2GB user...)  This error only pops up when I run 2 or 3 exports for this user simultaneously, e.g., by submitting html & xml at the same time.  Everything eventually completes.  

(4)(5) I will try that this evening.

Sesquipedalian:
Regarding issue (3), try applying this patch. I don't have much hope that these changes will accomplish anything, but there's no harm in testing them.

[temp_table_full_test.diff.txt](https://github.com/SimpleMachines/SMF2.1/files/4732232/temp_table_full_test.diff.txt)

Sesquipedalian:
I just tried "Download All" using Chrome 83.0.4103.61, and it downloaded all the files correctly.

sbulen:
> ### Regarding issues (4) ad (5)
> Both of those sound like symptoms of something going wrong while renaming files and appending data. What happens if you comment out this block of code and run the test again?
> 
> ```
> 		if (file_exists($tempfile) && filesize($tempfile) >= 1024 * 1024 * 250)
> 		{
> 			rename($tempfile, $realfile);
> 			$realfile = $export_dir_slash . ++$filenum . '_' . $idhash_ext;
> 		}
> ```

Well, this was completely unexpected...  Frankly, I was expecting (2), (4), and (5) to go away.  And that I would have one single VERY LARGE export file.   

It started off looking like that, but as the file approached 2GB, I got (2) and (4) at the same time.  And...  I had multiple 2GB export files in there, e.g., a 1_xxxx and a 2_xxxx, like it was still splitting it up (???)...   

And then...  My 2GB+ file(s) disappeared and the whole thing started all over again - with empty files...  

I can't tell if (5) was addressed, because the files are gone.

Initial guess - there are multiple renames in there for different purposes & they're stepping on each other.

I'll try again in a bit, to see if the same thing happens again.

sbulen:
For the above tests, my export_rate was at 50, which is too high for this test.  Plus, I was seeing memory exhausted errors in my apache log.  So...

I restarted the test (with the rename commented out) with an export_rate of 5.  It has been running for 6 hours.  It is definitely in a loop - it just keeps adding more files, with the same messages.

I am seeing errors (2) and (3) in the error log.  I am not seeing (4) or (5).  No errors in the apache log.

I will try the new query tomorrow.

Sesquipedalian:
Thanks, @sbulen. That’s really helpful info. I look forward to seeing your results from the altered query. 

Meanwhile, I’m testing here with a similar nightmare dataset (a user with 1000 posts of approximately 2 MB each). The symptoms differ a bit on my machine from what you’ve reported, but the root issue that processing too much data at once is overwhelming the available system resources is the same. Like you, I’ve found that setting the value of $modSettings['export_rate'] to something low (11, in my case) allows the export to proceed, albeit at a snail’s pace. I haven’t seen repetitions in the exported data like you have, but that difference is probably just a matter of which system resources are being overwhelmed first on the different machines.

I am working on some code for more dynamic resource management in these sorts of situations.

sbulen:
> Like you, I’ve found that setting the value of $modSettings['export_rate'] to something low (11, in my case) allows the export to proceed, albeit at a snail’s pace. I haven’t seen repetitions in the exported data like you have, but that difference is probably just a matter of which system resources are being overwhelmed first on the different machines.

Mine can get to 2GB reasonably quickly - mainly because I have a Selenium IDE script pinging the home page every few seconds.  It needs activity...  

Note that this 2GB user was working - and quite stable - earlier.  See screenshot at the end of #5081 .  So I don't think it's a resource thing, I think it's a bug.  Resource issues show up in the apache logs.  

https://github.com/SimpleMachines/SMF2.1/pull/5081#issuecomment-619195556

sbulen:
> Regarding issue (3), try applying this patch. I don't have much hope that these changes will accomplish anything, but there's no harm in testing them.
> 
> [temp_table_full_test.diff.txt](https://github.com/SimpleMachines/SMF2.1/files/4732232/temp_table_full_test.diff.txt)

I completed this test & it was an overwhelming success...  

I submitted all 3 exports on my 2GB user simultaneously, and all 3 completed successfully.   Errors (2), (3), & (5) did not appear.  No apache errors & no SMF errors.  export_rate set to 5.  Since I refreshed the code, the 'rename' logic is still in place; it is no longer commented out.

Error (4) appeared briefly once; disappeared upon refresh.

_EDIT:  Execution time went from 30 minutes to do 1 to 10 minutes to do all 3...._

Sesquipedalian:
Excellent! I guess it was just the joins after all. ef9dc05 implements a permanent version of what was in that patch, and 
34665ff should minimize the opportunities for (4) to occur.

sbulen:
Getting hundreds of these errors:
image

sbulen:
Despite the above, it did complete all 3 exports in ~10 minutes, and the size is right.  
No error (4) observed.
I did not see error (5) opening the output either, however, it looked a little strange/different, with a naked DOCTYPE tag at the top.  I don't think I've seen that before:
image

Possibly related to the other error above?  Which might have to do with some of the txt cleanup that's been going on?

Sesquipedalian:
> ... Which might have to do with some of the txt cleanup that's been going on?

So it would appear, yes.

Sources/Profile-Export.php
```diff
@@ -0,0 +1,1956 @@
...
+			$stylesheet['variables'] .= "\n\t\t" . '<xsl:' . $element . ' name="' . $name . '"';
+
+			if (isset($var['xpath']))
+				$stylesheet['variables'] .= ' select="' . $var['value'] . '"/>';
+			else
+				$stylesheet['variables'] .= '>' . (!empty($var['no_cdata_parse']) ? $var['value'] : cdata_parse($var['value'])) . '</xsl:' . $element . '>';
+		}
+
+		// The top-level template. Creates the shell of the HTML document.
+		$stylesheet['html'] = '
```
MissAllSunday, line 938: Can these be moved to a template file, perhaps one function for each and returning the string rather than echoing it.

Sesquipedalian: That's a good idea.

sbulen: Sounds like changes are coming...  I'll wait until after the new commits to re-test.  Just let me know when they are complete.

Sesquipedalian: It's done now, @sbulen: 830c32b

It doesn't change much, really. It mostly just shuffles some existing code around between different files. I also took the opportunity to make this system easily extensible in case we find ourselves needing other XSLT stylesheets in the future.

sbulen:
Getting some new errors.  A small #, 12 total, running the 2GB user test:
image

Sesquipedalian:
Looks like a write error, or possibly some malformed output. What is on line 87256 of the file it is complaining about, @sbulen?

sbulen:
> Looks like a write error, or possibly some malformed output. What is on line 87256 of the file it is complaining about, @sbulen?

ACK!!! I just lost it.  

I repeated the test, and got the exact same set of errors, just in different places...  There is nothing wrong with the lines in question.  This is one of the new errors:
image

The start of that line is uneventful:
image


The end of that line is also uneventful:
image

The problem is that I just can't get this to complete with my 2GB user right now.  (Submit all 3 at once.)  

**_Note this was working great a week ago, with the exception of the sprintf errors..._**

Sesquipedalian:
Hm. That's really odd. The errors all clearly indicate that it is running into trouble while trying to read the XML, but the fact that the problem happens at different, apparently random lines of the XML file suggest that the issue is neither the XML data nor the XSLT. The only reasonable explanation I can think of is a race condition where data is being read at the same time that it is being written, but I don't see how the changes in 4413197 and 830c32b could cause that. Still, I'm not omniscient, so I could be wrong.

Can you tell me what happens when you check out each of the following commits and run the test with that version, @sbulen?
1. d32b464
2. 4413197

sbulen:
So this is what I tried:

revert 4413197, & d32b464 as requested.  
result:  Wouldn't work, due to a merge conflict.  830c32b has to be reverted as well.

revert 830c32b, 4413197, & d32b464
result:  Completed properly - but the sprintf errors returned.

revert 830c32b
result: The 'premature end of data' errors were present.  

revert 830c32b, 4413197
result: Completed perfectly.  No errors.  

The problem is with 4413197.  (Curious - why is there another new version of jquery in there?)

SychO9:
> (Curious - why is there another new version of jQuery in there?)

probably a slight mistake when merging the main branch to this one, jQuery seems to have been updated to 3.5.1 recently. So the 3.4.1 file in this branch can just be removed If I'm not mistaken.

Sesquipedalian:
That jQuery change was definitely not intended. I can only guess that I accidentally added the old version after merging upstream changes from release-2.1. 

Anyway, I’ll revert those two commits. It seems very strange to me that 4413197 could cause that misbehaviour, though.

sbulen:
> Anyway, I’ll revert those two commits. It seems very strange to me that [4413197](https://github.com/SimpleMachines/SMF2.1/commit/441319771fa9058ebd36fef3c204eef0ab8dfb71) could cause that misbehaviour, though.

My initial thought is more that it had to do with jquery.  Maybe a bug in that version for handling large quantities of data, or, bad defaults for buffer sizes, etc.  

New versions need a lot of scrutiny.  

The rest of that commit looked pretty harmless.

Sesquipedalian:
Well, jQuery isn't involved at all in server-side processing of XML by PHP, so it is difficult to see how that would affect it. But adding a simple "ID: #" to the output also shouldn't cause the behaviour you saw. The error messages very clearly seem to say that the XML data was incomplete at read time, which only makes sense if reading and writing were happening simultaneously. Hopefully the reversion in 1a35992 and the file locks I added in 574b246 & fa1dd18 will fix it, at any rate.

sbulen:
I am getting lots of these again:
image

sbulen:
I do agree that you are likely operating on the same files with multiple operations somehow.

Sesquipedalian:
Please apply this patch on top of fa1dd18 and test again, @sbulen. It should make the export process more resilient against temporarily unavailable files.
[patch.diff.txt](https://github.com/SimpleMachines/SMF2.1/files/4802878/patch.diff.txt)

sbulen:
With that diff, we're back to these:
image

My server isn't showing any signs of stress, it's just humming along.  No apache errors.

sbulen:
Seeing these after applying the patch:
image

I'll try a couple other things here...

sbulen:
If you write to a simple log, once at the beginning of execute() and once at the end, logging whether it's the start or the end of execute(), as well as $this->_details['format_settings']['extension'], you will see that the different executions may occasionally overlap.  

The errors get logged when this happens.  

My initial thought is that cron's 'claim threshold' logic is too tight.  I think it is interpreting a request that is still executing as a 'failed/abandoned' execution, and submits it again.

I can get these errors just running a single .html request on its own - for my massive 2GB user.

```
June 20, 2020, 08:42:07 PM source: ExportProfileData.php line: 40 func: execute disp: execute() start...html
June 20, 2020, 08:42:10 PM source: ExportProfileData.php line: 40 func: execute disp: execute() start...html
June 20, 2020, 08:42:11 PM TRACE:
	source: cron.php line: 278 called: sjrbTrace
	source:  line:  called: smf_error_handler_cron
	source: ExportProfileData.php line: 456 called: load
	source: ExportProfileData.php line: 81 called: exportHtml
	source: cron.php line: 233 called: execute
	source: cron.php line: 127 called: perform_task
June 20, 2020, 08:42:11 PM source: cron.php line: 278 func: smf_error_handler_cron disp: In Error Handler...  Error: DOMDocument::load(): CData section not finished
200k of text 200k of text 200k of text 200k of tex in file:/D:/EasyPHP/van21/exports/4_f4f353053d4fce4e86ca009323b1600893477b81.html, line: 87236
June 20, 2020, 08:42:11 PM TRACE:
	source: cron.php line: 278 called: sjrbTrace
	source:  line:  called: smf_error_handler_cron
	source: ExportProfileData.php line: 456 called: load
	source: ExportProfileData.php line: 81 called: exportHtml
	source: cron.php line: 233 called: execute
	source: cron.php line: 127 called: perform_task
June 20, 2020, 08:42:11 PM source: cron.php line: 278 func: smf_error_handler_cron disp: In Error Handler...  Error: DOMDocument::load(): Premature end of data in tag body_html line 87236 in file:/D:/EasyPHP/van21/exports/4_f4f353053d4fce4e86ca009323b1600893477b81.html, line: 87236
June 20, 2020, 08:42:11 PM TRACE:
	source: cron.php line: 278 called: sjrbTrace
	source:  line:  called: smf_error_handler_cron
	source: ExportProfileData.php line: 456 called: load
	source: ExportProfileData.php line: 81 called: exportHtml
	source: cron.php line: 233 called: execute
	source: cron.php line: 127 called: perform_task
June 20, 2020, 08:42:11 PM source: cron.php line: 278 func: smf_error_handler_cron disp: In Error Handler...  Error: DOMDocument::load(): Premature end of data in tag member_post line 65449 in file:/D:/EasyPHP/van21/exports/4_f4f353053d4fce4e86ca009323b1600893477b81.html, line: 87236
June 20, 2020, 08:42:11 PM TRACE:
	source: cron.php line: 278 called: sjrbTrace
	source:  line:  called: smf_error_handler_cron
	source: ExportProfileData.php line: 456 called: load
	source: ExportProfileData.php line: 81 called: exportHtml
	source: cron.php line: 233 called: execute
	source: cron.php line: 127 called: perform_task
June 20, 2020, 08:42:11 PM source: cron.php line: 278 func: smf_error_handler_cron disp: In Error Handler...  Error: DOMDocument::load(): Premature end of data in tag xml-feed line 2 in file:/D:/EasyPHP/van21/exports/4_f4f353053d4fce4e86ca009323b1600893477b81.html, line: 87236
June 20, 2020, 08:42:12 PM source: ExportProfileData.php line: 86 func: execute disp: execute() end...html
June 20, 2020, 08:42:12 PM source: ExportProfileData.php line: 40 func: execute disp: execute() start...html
June 20, 2020, 08:42:12 PM source: ExportProfileData.php line: 86 func: execute disp: execute() end...html
June 20, 2020, 08:42:13 PM source: ExportProfileData.php line: 86 func: execute disp: execute() end...html
```

SychO9:
@sbulen can you check your `background_tasks` table and make sure you haven't somehow ended up with more than one of the **same** task. Or if you cannot tell, just prune the table and re-try the export.

The reason I ask is because I can't reproduce the issues you're having no matter how many things/times I try (not the same amount of data but 8000 posts of 1MB each I would think should be enough). 

The only way I can produce similar issues is if I copy the same task in the database. (Having many of the same task happens when testing and the tasks runs into fatal errors, it does not happen nder normal circumstances)

sbulen:
I definitely see more than one when they overlap (briefly - see image below), with different task IDs.

The issue is definitely the threshold logic.  A second task shouldn't be starting at all.  

Note I have a SIDE script simulating user activity (just loading the board index), so crons are getting checked frequently.  This is more like normal forum behavior, though.  If I set 'user activity' for 5 seconds apart, I may not see issues, or only a couple.  If I set them for 1/2 sec apart (more like reality) I see more.  

I think you need a large user & simulated load to reproduce.

image

sbulen:
Possibly part of the problem?  Not sure we should be checking the time when updating a background task:
https://github.com/SimpleMachines/SMF2.1/blob/b3268cc2b2f5972134cd8d65e5131e2a64ee0df3/cron.php#L175

The task_id is unique.  At best, checking the time is unnecessary.  At worst, this may be contributing to overlapping tasks.

SychO9:
Well, at least we know that the issue is the task being duplicated (maybe not duplucated, might have to do with the task creating the next task..), I don't think you would receive all those errors if it weren't. Now how and why is it getting duplicated.

> Possibly part of the problem? Not sure we should be checking the time when updating a background task:

Maybe, but I don't see how that can lead to the duplication, need to do some more digging.

sbulen:
I *think* I know what is going on.  Not sure what the solution is yet.

First off, the cron always deletes its background task record on completion.  So, when the export functions know there is more work to do, a new background task record is added.  I.e., you are going to have a 2nd bt record, that is normal.  But it should be the very last thing to do, only once it's OK for the next chunk to proceed.  Overlap should be brief & inconsequential.

Once the background task is complete, control is returned to cron, which deletes the prior bt record.

The html export first builds an xml export (a big chunk of work; it re-uses the xml export logic), then applies the xslt (another big chunk of work).  BUT...  The xml export may have put out a 'do the next chunk' bt record in the middle of the html processing.  

If cron wakes up while the html export is doing the xslt logic, it may find a bt record put there by the xml export & start working on it...  But the prior task isn't complete yet.  We need to find a way to ensure that putting the bt for the next chunk of work out there is the very last thing that is done, even for html exports.

sbulen:
My suggestion is to move inserting the new background task record to the end of execute().  That way *all* work is done, whether for xml, html or styled.xml, before putting the bt up for grabs by cron.

FWIW, a total kludge - that works - is to set the default value for $delay to 2 seconds in the export xml logic.  That appears to be more than enough elbow room for any amount of xslt processing that may follow insertion of the bt record.  (1 might even work...)

Sesquipedalian:
That's some nice detective work, @sbulen , thanks!

SychO9:
with the last commit the tasks take more time, it's seems like they fail more often:

```
[Mon Jun 22 16:42:44.492852 2020] [php7:notice] [pid 82436:tid 1216] [client 127.0.0.1:58697] PHP Fatal error:  Maximum execution time of 10 seconds exceeded in ...\\Sources\\tasks\\ExportProfileData.php on line 437, referer: http://.../index.php?action=profile;area=getprofiledata;u=1
[Mon Jun 22 16:42:44.492852 2020] [php7:notice] [pid 82436:tid 1216] [client 127.0.0.1:58697] PHP Stack trace:, referer: http://.../index.php?action=profile;area=getprofiledata;u=1
[Mon Jun 22 16:42:44.492852 2020] [php7:notice] [pid 82436:tid 1216] [client 127.0.0.1:58697] PHP   1. {main}() ...\\cron.php:0, referer: http://.../index.php?action=profile;area=getprofiledata;u=1
[Mon Jun 22 16:42:44.493849 2020] [php7:notice] [pid 82436:tid 1216] [client 127.0.0.1:58697] PHP   2. perform_task() ...\\cron.php:127, referer: http://.../index.php?action=profile;area=getprofiledata;u=1
[Mon Jun 22 16:42:44.493849 2020] [php7:notice] [pid 82436:tid 1216] [client 127.0.0.1:58697] PHP   3. ExportProfileData_Background->execute() ...\\cron.php:233, referer: http://.../index.php?action=profile;area=getprofiledata;u=1
[Mon Jun 22 16:42:44.493849 2020] [php7:notice] [pid 82436:tid 1216] [client 127.0.0.1:58697] PHP   4. ExportProfileData_Background->exportHtml() ...\\Sources\\tasks\\ExportProfileData.php:82, referer: http://.../index.php?action=profile;area=getprofiledata;u=1
```

The block of code added inside the loop in the last commit is the cause, should be better outside the loop.

Sesquipedalian:
> with the last commit the tasks take more time, it's seems like they fail more often:
> 
> ```
> [Mon Jun 22 16:42:44.492852 2020] [php7:notice] [pid 82436:tid 1216] [client 127.0.0.1:58697] PHP Fatal error:  Maximum execution time of 10 seconds exceeded in ...\\Sources\\tasks\\ExportProfileData.php on line 437, referer: http://.../index.php?action=profile;area=getprofiledata;u=1
> [Mon Jun 22 16:42:44.492852 2020] [php7:notice] [pid 82436:tid 1216] [client 127.0.0.1:58697] PHP Stack trace:, referer: http://.../index.php?action=profile;area=getprofiledata;u=1
> [Mon Jun 22 16:42:44.492852 2020] [php7:notice] [pid 82436:tid 1216] [client 127.0.0.1:58697] PHP   1. {main}() ...\\cron.php:0, referer: http://.../index.php?action=profile;area=getprofiledata;u=1
> [Mon Jun 22 16:42:44.493849 2020] [php7:notice] [pid 82436:tid 1216] [client 127.0.0.1:58697] PHP   2. perform_task() ...\\cron.php:127, referer: http://.../index.php?action=profile;area=getprofiledata;u=1
> [Mon Jun 22 16:42:44.493849 2020] [php7:notice] [pid 82436:tid 1216] [client 127.0.0.1:58697] PHP   3. ExportProfileData_Background->execute() ...\\cron.php:233, referer: http://.../index.php?action=profile;area=getprofiledata;u=1
> [Mon Jun 22 16:42:44.493849 2020] [php7:notice] [pid 82436:tid 1216] [client 127.0.0.1:58697] PHP   4. ExportProfileData_Background->exportHtml() ...\\Sources\\tasks\\ExportProfileData.php:82, referer: http://.../index.php?action=profile;area=getprofiledata;u=1
> ```
> 
> The block of code added inside the loop in the last commit is the cause, should be better outside the loop.

Ah, yes. When I moved the code inside the loop, I seem to have inadvertently changed `@set_time_limit(MAX_CLAIM_THRESHOLD * 2);` to `@set_time_limit(MAX_CRON_TIME);` rather than `@set_time_limit(MAX_CLAIM_THRESHOLD);`. I've fixed that in e1b859d. Just for good measure, I've also added `@set_time_limit(MAX_CLAIM_THRESHOLD);` at the start of `execute()`.

Sesquipedalian:
By the way, @sbulen, thanks for all your faithful and rigorous cross-platform testing! That torture test you set up may have made me want to pull my hair out a few times, but it's made this code way better than it would have been. Much appreciated!

SychO9:
Yea, better now.

I did surprisingly get this one error, very hard to reproduce though.

```
Type of error: General
file_get_contents(): read of 8192 bytes failed with errno=13 Permission denied
http://.../index.php?action=profile;area=getprofiledata;u=1
.../Sources/Profile-Export.php (Line 235)
```

Sesquipedalian:
Oh, that was probably trying to read the progress.json file just at the moment when it was being updated. I should probably add a little loop or something to check for that possibility.

SychO9:
Here's another minor and last issue from my part:

if a post has for example
```
</div></div>
```

The styled XML's structure will be broken (when you open it in the browser), it seems to actually read that as XML code and not raw text. It's especially a problem since you can have posts with code blocks containing html or xml.

```xml
<member_post label="Post">
	<id>509</id>
	<subject label="Subject">sdfsdf</subject>
	<body label="Message">&lt;<![CDATA[/div]]>&gt;&lt;<![CDATA[/div]]>&gt;</body>
	<body_html label="HTML">&lt;<![CDATA[/div]]>&gt;&lt;<![CDATA[/div]]>&gt;</body_html>
```

sbulen:
> By the way, @sbulen, thanks for all your faithful and rigorous cross-platform testing! That torture test you set up may have made me want to pull my hair out a few times, but it's made this code way better than it would have been. Much appreciated!

Sorry about the hair!

sbulen:
> Oh, that was probably trying to read the progress.json file just at the moment when it was being updated. I should probably add a little loop or something to check for that possibility.

I am getting a lot of these when I submit multiple requests in parallel, & I have a different theory...

Why is $next_task static?  That should be instance-specific, correct?  For that matter, why are any of the properties in ExportProfileData static (with the possible exception of real_modsettings)?  

I suspect they are operating on each other's data.  My understanding of static properties is that they are like constants - used across all instances of a class (which is very different from static variables within a function).

Sesquipedalian:
> Here's another minor and last issue from my part:
> 
> if a post has for example
> 
> ```
> </div></div>
> ```
> 
> The styled XML's structure will be broken (when you open it in the browser), it seems to actually read that as XML code and not raw text. It's especially a problem since you can have posts with code blocks containing html or xml.
> 
> ```
> <member_post label="Post">
> 	<id>509</id>
> 	<subject label="Subject">sdfsdf</subject>
> 	<body label="Message">&lt;<![CDATA[/div]]>&gt;&lt;<![CDATA[/div]]>&gt;</body>
> 	<body_html label="HTML">&lt;<![CDATA[/div]]>&gt;&lt;<![CDATA[/div]]>&gt;</body_html>
> ```

Fixed in 0da5e57. 🙂

Sesquipedalian:
> > Oh, that was probably trying to read the progress.json file just at the moment when it was being updated. I should probably add a little loop or something to check for that possibility.
> 
> I am getting a lot of these when I submit multiple requests in parallel, & I have a different theory...
> 
> Why is $next_task static? That should be instance-specific, correct? For that matter, why are any of the properties in ExportProfileData static (with the possible exception of real_modsettings)?
> 
> I suspect they are operating on each other's data. My understanding of static properties is that they are like constants - used across all instances of a class (which is very different from static variables within a function).

Oh, sorry. I missed this comment.

I could always be wrong, but I'm pretty sure it was the file lock that generated the permission denied error. You are quite right that `$next_task` should not have been static, as I realized while working on efbeb97 this morning, but I don't think it would have caused that particular error. 🙂

As for why `$export_details`, `$real_modSettings`, and `$xslt_info` are all static variables, that's because they are each used in the various static functions near the end of ExportProfileData.php, and static functions need static variables.

sbulen:
Making $next_task non-static resolved the issues in my testing earlier.

Let me know when ready to retest.

Sesquipedalian:
> Making $next_task non-static resolved the issues in my testing earlier.
> 
> Let me know when ready to retest.

Fair enough. I am often wrong. 🙂

Ready to test now.

sbulen:
> > Making $next_task non-static resolved the issues in my testing earlier.
> > Let me know when ready to retest.
> 
> Fair enough. I am often wrong. 🙂
> 
> Ready to test now.

I am still curious why the other properties don't cause issues, but they do not in my testing.  They are request-specific, e.g., they are based on uid & format.  Something that should NOT be shared across instances.  But they are declared static.  E.g., xslt_info...

I don't understand why that works...  I'm missing something.  I would have thought any property instance-specific should NOT be static or the results would be indeterminate, i.e., maybe using what was declared elsewhere amongst multiple parallel requests.  

OTOH, the other items were static _for the specific purpose of using integration hooks_.  But those integration hooks are specifically for work to do mid-task _(i.e., the settings table is never really touched)_.  Since php single-threads, maybe that works  - $modsettings storage of the hook info is tied only to the current thread...

I admit I'm confused why parts are working here...

sbulen:
Only one issue I can't get past at the moment - 'huge input lookup'.  Reading around, apparently there is an 'XML_PARSE_HUGE' flag we may need to set somewhere?

I have not seen this issue in prior testing...

image

My big 2GB user, with multiple parallel requests, has _**passed**_.  

Executing against a # of my real forum users is not working at the moment due to this 'huge input lookup' issue.  Odd, because my 2GB user is the one with huge posts.

_Note:  Various issues have been encountered, e.g., the 'resource temporarily unavailable', but have all been resolved by lowering the export_rate.  We are very, very sensitive to the export_rate..._

sbulen:
I am still seeing the 'huge input lookup' error.  Note this is happening to the user I have with the massive (5MB) post that has html in it.  This was working fine until recently.
image
This error is fatal; that export request does not proceed any further. It is displayed as 'completed' in the app, but it is not really complete.

I have been running a variety of tests all day today.  All other tests I am running are looking good at the moment.  I did get one 'resource temporarily unavailable' on my 2GB user when experimenting with different settings.  Those are not fatal; the task appears to just end but the process resumes properly on the next go around.

Sesquipedalian:
1. What libxml version does phpinfo report on the affected server?
2. Does the export complete successfully for the 5 MB post after you run `git revert 0da5e57` on your local copy?

sbulen:
> 1. What libxml version does phpinfo report on the affected server?

libxml Version | 2.9.9
I've been running all these tests on a php 7.3 environment

> 2. Does the export complete successfully for the 5 MB post after you run `git revert 0da5e57` on your local copy?

Yep!

Sesquipedalian:
I made a silly mistake in 8d17721. Please try again with 8a4cafe, @sbulen. The `LIBXML_PARSEHUGE` flag should solve those error messages for you.

I think the reason you didn't see those errors prior to 0da5e57 was that `cdata_parse()` was doing things incorrectly, but in a way that would *appear* correct if the result just so happened to be well-formed XML. Specifically, if the string returned from `cdata_parse()` parsed as a valid nested XML structure (as would have been the case with the contents of the `[html]` BBCode in that enormous post of yours) then it would have been handled by the parser as a tree of smaller XML elements rather than one enormous string. Thus you were avoiding the "huge input" error because the content of the `<body_html>` element was being parsed as many smaller elements which all apparently stayed within libxml's internal limit. But once I fixed `cdata_parse()` so that it returned only strings, that enormous string exceeded the limit and produced the error. At any rate, the `LIBXML_PARSEHUGE` should resolve it.

sbulen:
Hmmm...  Haven't seen this one before:
image

Sesquipedalian:
🙄 at myself.

sbulen:
FYI, changing 'TIME STARTED' to TIME_START' of course, fixed that issue.

After making that change, everything ran to completion.  The tests I've been running are:
 - Test acct, with 2GB of posts, running styled.xml, .xml, & .html simultaneously
 - Copy of my prod forum - running my 10 biggest users & some random ones, all 3 types (styled, xml & html) simultaneously

For my extreme test case, the 2GB user, who averages 2MB per post, I have had to lower the export_rate down to 5.  It may make sense to allow very low values as low as 5.

sbulen:
Cross post.  I had made that change locally & had already rerun my tests a couple times.

Sesquipedalian:
> For my extreme test case, the 2GB user, who averages 2MB per post, I have had to lower the export_rate down to 5. It may make sense to allow very low values as low as 5.

Agreed. Out of curiosity, what happens at this point when you leave it too high? What does the failure look like?

sbulen:
> > For my extreme test case, the 2GB user, who averages 2MB per post, I have had to lower the export_rate down to 5. It may make sense to allow very low values as low as 5.
> 
> Agreed. Out of curiosity, what happens at this point when you leave it too high? What does the failure look like?

One of two failure modes:
 - 'Resource temporarily unavailable' - these seem to properly recover themselves.  Everything resumes properly the next time around.  I have not seen this one today, but I have seen it quite a lot over the last week or so.
 - Request frozen.  This one is odd.  The background task is there, and the .tmp file, etc., are all there.  No errors logged in apache or smf.  Any other requests proceed OK.  It just doesn't go anywhere further.  No timestamps are updated anywhere.  It's like it's being ignored.  When this happens, it is always the styled.xml type.  (I have not double-checked the claimed time, maybe it's possible the 'no disk space' logic kicked in & it will be idle for a day???)

sbulen:
> * Request frozen.  This one is odd.  The background task is there, and the .tmp file, etc., are all there.  No errors logged in apache or smf.  Any other requests proceed OK.  It just doesn't go anywhere further.  No timestamps are updated anywhere.  It's like it's being ignored.  When this happens, it is always the styled.xml type.  (I have not double-checked the claimed time, maybe it's possible the 'no disk space' logic kicked in & it will be idle for a day???)

I just reproduced one of those...  The timestamp is in the past, so there goes my earlier theory.  

I've been watching that dude, and it resumed after several minutes (?!?!?!)....   I now wonder if the 'abandoned' timeframe in cron.php kicked in and it resumed...  

I'll do a few more tomorrow...  Not tonight.  TIme for a beer.

sbulen:
OK, so I've run a full set of tests over the last 3 hours, after the last set of commits.

Everything ran perfectly - both my 2GB test user & a set of all my largest users running all request types in parallel.  

Some of the performance & cleanup tweaks of the last couple weeks have definitely paid off.  I have even been able to increase my export_rate across the board, and still see no errors.

When I go out of my way to deliberately break it, I see 2 modes, depending on how hard I push it:
 - Stalls...  Normally, they all operate in parallel, taking turns.  But when stressed, some requests may "stall" for a period of time.  This happens if I have many overly-large requests in parallel, and the export-rate is borderline too big.  When things quiet down, after some have completed, the stalled items come back to life & continue successfully to completion.  
 - Stops...  If I provide a totally ludicrous value for export_rate for that environment, things just don't proceed at all...  The system just kinda acts like, "You're kidding, right?"...  And does nothing.   _EDIT:  Given time, I will ultimately see memory exhausted errors in the apache log..._  

I think this is ready.

Sesquipedalian:
I’m glad to call this done. At some point we might want to try to find a way to throttle the export rate dynamically, but that can be left for another day.

PR #6107, opened 2020-05-07
trying to be clever about bbc parsing introduced a side effect

### LINKED ISSUE #6028: Mail - salutation sometimes in wrong language
#### Description
From this reported issue:
https://www.simplemachines.org/community/index.php?topic=571923.0

### Steps to reproduce
1. Set up 3 different users, in different languages
2. All 3 should be setup to receive mail notifications (straight away); have them all follow the same topic
3. Take turns posting posting from the different languages
4. This is an intermittent problem - it works fine, most of the time, but sometimes the salutation is in the wrong language

- Version/Git revision:  Today
- Database Type:  Mysql
- Database Version:  5.7
- PHP Version:  7.3

Notice the "regards" instead of "saludos":
image

live627:
ref https://www.simplemachines.org/community/index.php?topic=571923.0

fixes #6028

### DIFF
--- a/Sources/tasks/CreatePost-Notify.php
+++ b/Sources/tasks/CreatePost-Notify.php
@@ -246,20 +246,32 @@ class CreatePost_Notify_Background extends SMF_BackgroundTask
 			$user_info = $members_info[$member_id];
 
 			// Censor and parse BBC in the receiver's localization. Don't repeat unnecessarily.
-			$localization = implode('|', array($member_data['lngfile'], $user_info['time_offset'], $user_info['time_format']));
-			if (empty($parsed_message[$localization]))
-			{
-				loadLanguage('index+Modifications', $member_data['lngfile'], false);
+			loadLanguage('index+Modifications', $receiver_lang, false);
 
-				$parsed_message[$localization]['subject'] = $msgOptions['subject'];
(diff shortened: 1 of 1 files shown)


sbulen (changes requested):
Were you able to test this?  This doesn't pass parse:
_[Tue May 12 20:30:28.008992 2020] [php7:crit] [pid 3500:tid 1384] [client 127.0.0.1:53048] PHP Parse error:  syntax error, unexpected ';', expecting ')' in D:\\blah\\blah\\Sources\\tasks\\CreatePost-Notify.php on line 334, referer: https://localhost/van21/index.php?action=admin;area=scheduledtasks;done_

Please test!  Thanks,

Sesquipedalian (changes requested):
I don't think most of these changes are necessary to fix the issue. I believe all that needs to be done is to move `loadLanguage('index+Modifications', $member_data['lngfile'], false);` out of the `if` block, like this:
```
loadLanguage('index+Modifications', $member_data['lngfile'], false);

// Censor and parse BBC in the receiver's localization. Don't repeat unnecessarily.
$localization = implode('|', array($member_data['lngfile'], $user_info['time_offset'], $user_info['time_format']));
if (empty($parsed_message[$localization]))
{
	$parsed_message[$localization]['subject'] = $msgOptions['subject'];
```

Issue #6111, opened 2020-05-13
`loadMemberContext` no longer returns a bool  while various codes expect it to


live627:
See theses search results https://github.com/SimpleMachines/SMF2.1/search?q=loadMemberContext&unscoped_q=loadMemberContext

Can we agree to stop changing the API?

ref #6049 
ref #6079

I recommend a revert.

Sesquipedalian:
Agreed. The justification for the change appears to have been to "[minimize global usage](https://github.com/SimpleMachines/SMF2.1/pull/6049)." That's a nice idea in theory, but realistically it's not going to happen in 2.1, so it's not worthwhile to break a bunch of existing code merely in order to make a fairly minor gesture in that direction.

Care to put together a PR to revert the change, @live627?

sexwarrior:
Given that this fix has been [immediately undone with no documentation](https://github.com/SimpleMachines/SMF2.1/commit/95388c999ab6e3c85070987960c925c27e112b8e), does this issue need re-opening?

PR #6116, opened 2020-05-16
Fixes some faulty logic in loadMemberContext()

Sesquipedalian:
Previously we used a static variable named `$dataLoaded` to remember which members we had already loaded data for. This was both useless and buggy in multiple ways. First, it was useless because we can just check `$memberContext[$user]` directly and skip the middleman. Second, the value recorded in `$dataLoaded[$user]` could fall out of date if `$memberContext[$user]` were unset elsewhere, such as in News.php's `getXmlProfile()` function. Third, this static variable did nothing to remember whether the `$display_custom_fields` had been passed a true or false.

This PR fixes all of that.

See #6111, #6112, #6113, and https://github.com/SimpleMachines/SMF2.1/pull/6101#issuecomment-628375442 for more info.

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -1583,12 +1583,12 @@ function loadMemberContext($user, $display_custom_fields = false)
 {
 	global $memberContext, $user_profile, $txt, $scripturl, $user_info;
 	global $context, $modSettings, $settings, $smcFunc;
-	static $dataLoaded = array();
+	static $dcf = array();
 	static $loadedLanguages = array();
 
 	// If this person's data is already loaded, skip it.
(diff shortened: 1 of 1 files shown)


albertlast:
And what is dcf?

Sesquipedalian:
An abbreviation from "display custom fields."

Sesquipedalian:
Hm. Even though this change is, strictly speaking, outside the scope of #6101, the reality is that #6101 is the only known context in which we can test it, because #6101 is what exposed the bug that this is meant to fix. So, I'm going to close this PR and cherry-pick the commit into #6101 directly.

Issue #6120, opened 2020-05-16
sprintf instead of string concatenation


m4z:
#### Description
To migrate to the new translation tool, we need to convert from "dot"-string-concatenating the language strings to `sprintf` (see internal discussion for more details).

This should probably happen ASAP. I'll prepare an initial PR later today.

jdarwood007:
This is a release blocker as it prevents moving forward with the new tool.

PR #6121, opened 2020-05-16
sprintf strings

m4z:
This PHP noob is attempting to convert concat strings to sprintf, probably making it worse before it gets better…

Feel free to nudge (or 180) me into the right direction, or write superseding PRs. :wink:


m4z:
Addresses #6120.

Themes/default/languages/index.english.php
```diff
@@ -150,7 +150,7 @@
 $txt['notify'] = 'Notify';
 $txt['unnotify'] = 'Unnotify';
 // Use numeric entities in the below string.
-$txt['regards_team'] = "Regards,\nThe " . $context['forum_name'] . ' Team.';
+$txt['regards_team'] = 'Regards,' . "\n" . 'The %1$s Team.';
```
m4z: I don't understand enough about PHP (or the SMF requirements) to address the newline.

Oldiesmann: Do the double quotes mess up the `%1$s` part? If not, `$txt['regards_team'] = "Regards,\nThe %1$s Team.";` should work.

m4z: I thought they mess it up, but apparently [it should work like you suggest](https://www.php.net/manual/en/language.types.string.php#language.types.string.parsing).

m4z:
Ok, I have no clue how to do `$txt['movetopic_default']`; apparently the string is only loaded by `loadLanguage()` and then _maybe_ somehow used via `$_POST['reason']`… (not sure about that all all, though)

jdarwood007:
@m4z 
Its defined in this way
```
Zeus:Downloads jeremy$ grep -Rn "movetopic_default" ~/repos/sm/SMF2.1
~/repos/sm/SMF2.1/Sources/MoveTopic.php:111:		$temp = $txt['movetopic_default'];
~/repos/sm/SMF2.1/Sources/MoveTopic.php:114:		$txt['movetopic_default'] = $temp;
~/repos/sm/SMF2.1/Themes/default/languages/index.english.php:539:$txt['movetopic_default'] = 'This topic has been moved to ' . $txt['movetopic_auto_board'] . ".\n\n" . $txt['movetopic_auto_topic'];
```

And used in this way

```
							<dd>
								<textarea name="reason">', $txt[$type . 'topic_default'], '</textarea>
							</dd>
```

Which makes it easier to understand:
```
Zeus:Downloads jeremy$ grep -Rn "topic_default" ~/repos/sm/SMF2.1
/Users/jeremy/repos/sm/SMF2.1/Sources/MoveTopic.php:111:		$temp = $txt['movetopic_default'];
/Users/jeremy/repos/sm/SMF2.1/Sources/MoveTopic.php:114:		$txt['movetopic_default'] = $temp;
/Users/jeremy/repos/sm/SMF2.1/Themes/default/MoveTopic.template.php:103:								<textarea name="reason">', $txt[$type . 'topic_default'], '</textarea>
/Users/jeremy/repos/sm/SMF2.1/Themes/default/languages/index.english.php:539:$txt['movetopic_default'] = 'This topic has been moved to ' . $txt['movetopic_auto_board'] . ".\n\n" . $txt['movetopic_auto_topic'];
/Users/jeremy/repos/sm/SMF2.1/Themes/default/languages/index.english.php:695:$txt['mergetopic_default'] = 'This topic has been merged into ' . $txt['movetopic_auto_topic'] . '.';
```

So just convert both strings to be sprintf safe and this works.

m4z:
I can't find `$txt['attach_dir_files_missing']` being used anywhere.

Themes/default/MoveTopic.template.php
```diff
@@ -99,8 +99,18 @@ function template_redirect_options($type)
...
+	{
+		echo '
+								<textarea name="reason">', sprintf($txt[$type . 'topic_default'], $txt['movetopic_auto_board'], $txt['movetopic_auto_topic']), '</textarea>';
+	}
+	elseif ($type == 'merge')
+	{
+		echo '
+								<textarea name="reason">', sprintf($txt[$type . 'topic_default'], $txt['movetopic_auto_topic']), '</textarea>';
+	}
+	echo '
```
jdarwood007: Since you specified in the language strings the numerical indexes (%1, %2), you can just use one call here.  The other replacement will be ignored.

Themes/default/Help.template.php
```diff
@@ -29,7 +29,16 @@ function template_popup()
 	</head>
 	<body id="help_popup">
 		<div class="windowbg description">
-			', $context['help_text'], '<br>
+			';
+
+	if ($_GET['help'] == 'cal_short_months')
```
m4z: It might be more appropriate to do this at the bottom of `Sources/Help.php`, but I wasn't sure which of the three cases I needed and I feared I'd have to copy the `if` into each case. If I should move it there, tell me.

Themes/default/Help.template.php
```diff
@@ -35,6 +35,8 @@ function template_popup()
 		echo sprintf($context['help_text'], $txt['months_short'][1], $txt['months_titles'][1]);
 	elseif ($_GET['help'] == 'cal_short_days')
 		echo sprintf($context['help_text'], $txt['days_short'][1], $txt['days'][1]);
+	elseif ($_GET['help'] == 'enableSpellChecking')
+		echo sprintf($context['help_text'], ((function_exists('pspell_new') || function_exists('enchant_broker_init')) ? $helptxt['enableSpellCheckingSupported'] : $helptxt['enableSpellCheckingUnsupported']))
```
m4z: Is `$helptxt` even available in this context?

Themes/default/languages/Help.english.php
```diff
@@ -195,8 +195,8 @@
...
 $helptxt['cal_week_links'] = 'If this setting is checked, links will be added alongside each week in the calendar.';
 $helptxt['cal_prev_next_links'] = 'If this setting is checked, previous month and next month links will be added to the top of each month for easy navigation.';
-$helptxt['cal_short_months'] = 'If this setting is checked, month names within the calendar will be shortened.<br><br><strong>Enabled:</strong> ' . $txt['months_short'][1] . ' 1<br><strong>Disabled:</strong> ' . $txt['months_titles'][1] . ' 1';
-$helptxt['cal_short_days'] = 'If this setting is checked, day names within the calendar will be shortened.<br><br><strong>Enabled:</strong> ' . $txt['days_short'][1] . '<br><strong>Disabled:</strong> ' . $txt['days'][1];
+$helptxt['cal_short_months'] = 'If this setting is checked, month names within the calendar will be shortened.<br><br><strong>Enabled:</strong> %1$s 1<br><strong>Disabled:</strong> %2$s 1';
+$helptxt['cal_short_days'] = 'If this setting is checked, day names within the calendar will be shortened.<br><br><strong>Enabled:</strong> %1$s<br><strong>Disabled:</strong> %1$s';
```
m4z: Hello, Mr. Smartypants! "1" twice?

m4z: Fixed by 005b66649

m4z:
This should be testable now; I have created a test install and will begin testing tomorrow.

Themes/default/languages/index.english.php
m4z: Hmm, but then again, this creates loads of `8: Undefined variable: s` errors in my test install… :cry:

m4z: Just committed a potential fix (6fdc40e8f), which I'll test tomorrow…

Themes/default/languages/Help.english.php
```diff
@@ -381,7 +383,12 @@
...
-$helptxt['enableSpellChecking'] = 'Enable spell checking. You MUST have the pspell or enchant library installed on your server and your PHP configuration set up to use the installed library. Your server ' . ((function_exists('pspell_new') || function_exists('enchant_broker_init')) ? 'DOES' : 'DOES NOT') . ' appear to have this set up.';
+// argument(s): either $helptxt['enableSpellCheckingSupported'] or $helptxt['enableSpellCheckingUnsupported'], depending on whether the server supports one of the libs
+$helptxt['enableSpellChecking'] = 'Enable spell checking. You MUST have the pspell or enchant library installed on your server and your PHP configuration set up to use the installed library. Your server %1$s appear to have this set up.';
+// is conditionally used as part of $helptxt['enableSpellChecking']
+$helptxt['enableSpellCheckingSupported'] = 'DOES';
+// is conditionally used as part of $helptxt['enableSpellChecking']
+$helptxt['enableSpellCheckingUnsupported'] = 'DOES NOT';
```
SychO9: This bothers me a little xD

We really shouldn't be breaking phrases up like that, they won't be able to be properly translated this way.

even the way it was before the edit was wrong, what we should have is two text strings

One for when it is supported
```php
$helptxt['enableSpellCheckingSupported'] = 'Enable spell checking. You MUST have the pspell or enchant library installed on your server and your PHP configuration set up to use the installed library. Your server DOES appear to have this set up.';
```

The other for when it is not supported
```
$helptxt['enableSpellCheckingNotSupported'] = 'Enable spell checking. You MUST have the pspell or enchant library installed on your server and your PHP configuration set up to use the installed library. Your server DOES NOT appear to have this set up.';
```

It should not matter whether a big part of the string is practically duplicated, this is a better approach for the sake of sane translation, other languages translate very differently and thus imho we should not control the string with words this way.

SychO9: Or even better we should just separate the `Your server DOES/NOT appear to have this set up.` into two strings so the first bit won't be duplicated this way. and then we simply concatenate them in the source file.

m4z: I attempted to move in that direction in f0afe70ad7cad3b7ce6638715f92f4aeaa9d4868, but I'm unsure how to change the [code](https://github.com/SimpleMachines/SMF2.1/pull/6121/commits/b75bac4b4c87bb30c2c2c5c1e1973e5281263bc9), since a single `$helptxt` variable has to be passed. Do you want it to be changed, so that instead of `enableSpellChecking`, `enableSpellChecking{Supported,Unsupported}` is being passed?

SychO9: The code actually works just fine with the change, `sprintf()` either injects `$helptxt['enableSpellCheckingSupported']` or `$helptxt['enableSpellCheckingUnsupported']` at the end of the text string which is exactly what we want.

unless I didn't understand what you meant ?

Themes/default/Help.template.php
SychO9: yea It would definitely be much much better if you were to moves these conditional statements to the Source file.

The function you are looking for is `ShowAdminHelp()` located in `Help.php`. I haven't tested but I believe you want your code to be added before the last if statement in that function.

Instead of echoing you will just be assigning the result to `$context['help']`.

Also you might want to play with a `switch` block ?

MissAllSunday: Indeed, a switch block would be preferable here.

Themes/default/Profile.template.php
```diff
@@ -2698,8 +2698,16 @@ function template_error_message()
 
 		// Cycle through each error and display an error message.
 		foreach ($context['post_errors'] as $error)
-			echo '
+			if ($error == 'password_short')
```
MissAllSunday: You can move the check for "password_short" to a ternary to avoid the if/else condition:

```php
		foreach ($context['post_errors'] as $error)
		{
			$text_key_error = $error == 'password_short' ?
				printf($txt['profile_error_' . $error], (empty($modSettings['password_strength']) ? 4 : 8)) :
				$txt['profile_error_' . $error];

				echo '
				<li>', isset($txt['profile_error_' . $error]) ? $text_key_error : $error, '</li>';
		}
```

Themes/default/languages/Help.english.php
m4z: What I meant was that `Sources/Help.php` is currently passing  "`enableSpellChecking`" to `Themes/default/Help.template.php`, and because I'm not 100% sure how to change that in `Help.php`, I still left it in the template until now. Will try to fix that today.

Themes/default/Profile.template.php
m4z: I find that hard to read, but you're the boss. (I think your `echo` was indented too far, changed that. (658f963))

m4z:
I'm incapable to test `$txt['template_parse_error_details']`, I just can't manage to trigger it.

m4z:
I also don't know to create the conditions needed for `$txt['basedirectory_for_attachments_warning']` to show up; I attempted to empty `smf_settings:attachmentUploadDir`, but that just caused general errors.

m4z:
I also have no clue how to test `$txt['backtrace_info']`.

m4z:
* I have not yet tested each `who*` individually
* I'll have to do a tedious re-install to test the `switch`

Other than that (and my 3 previous comments), I have tested everything to the best of my abilities.

Sources/Help.php
```diff
@@ -163,6 +163,24 @@ function ShowAdminHelp()
...
+		case 'cron_is_real_cron':
+			$context['help_text'] = printf($context['help_text'], $boarddir, $boardurl);
+			break;
+		case 'enableSpellChecking':
+			$context['help_text'] = printf($context['help_text'], ((function_exists('pspell_new') || function_exists('enchant_broker_init')) ? $helptxt['enableSpellCheckingSupported'] : $helptxt['enableSpellCheckingUnsupported']));
+			break;
+		case 'queryless_urls':
+			$context['help_text'] = printf($context['help_text'], (isset($_SERVER['SERVER_SOFTWARE']) && (strpos($_SERVER['SERVER_SOFTWARE'], 'Apache') !== false || strpos($_SERVER['SERVER_SOFTWARE'], 'lighttpd') !== false) ? $helptxt['queryless_urls_supported'] : $helptxt['queryless_urls_unsupported']));
+			break;
+	}
```
SychO9: please change `printf` to `sprintf` you don't want to print them right there and then, you just want to replace the proper bits of the strings

Themes/default/languages/Who.english.php
```diff
@@ -117,24 +141,36 @@
...
+$txt['whoallow_mlist'] = 'Viewing the <a href="%1$s?action=mlist">memberlist</a>.';
 $txt['whoallow_optimizetables'] = 'Optimizing the database tables.';
 $txt['whoallow_repairboards'] = 'Repairing the database tables.';
-$txt['whoallow_search'] = '<a href="' . $scripturl . '?action=search">Searching</a> the forum.';
+// argument(s): $scripturl
+$txt['whoallow_search'] = '<a href="%1$s?action=search">Searching</a> the forum.';
 $txt['whoallow_search2'] = 'Viewing the results of a search.';
-$txt['whoallow_stats'] = 'Viewing the <a href="' . $scripturl . '?action=stats">forum stats</a>.';
+// argument(s): $scripturl
+$txt['whoallow_stats'] = 'Viewing the <a href="%1$s?action=stats">forum stats</a>.';
```
SychO9, line 154: none of the `whoallow_` ones are being replaced, I believe you need to do so here

https://github.com/SimpleMachines/SMF2.1/blob/017e10b91144d1d68754947e6364944c9902b348/Sources/Who.php#L428

SychO9 (changes requested):

Sources/Help.php
m4z: I think I did that because of 658f963ba11b0935557764e08befe2fc0e84a369, should I change it there as well?

m4z: Fixed in 7b61f1f

SychO9: Thanks, and no in the template files it's fine since printing there is intended, but we (normally) don't print anything from the source files. 👍

SychO9 (approved):
Apart from the two things I noticed in my previous review, everything else looks good to me, I did my best and tested every single edited text string.

Nicely done @m4z

MissAllSunday (approved):

MissAllSunday:
Is this considered done?

m4z:
> Apart from the two things I noticed in my previous review, everything else looks good to me, I did my best and tested every single edited text string.

Awesome, thanks!


> Is this considered done?

I think so. I didn't get to testing the rest of the changes yet, but IIUC, @SychO9 did it for me.

halasla (commented):
--SPAM, But Github doesn't let you delete spam.

m4z:
Have you reported the account?

PR #6133, opened 2020-05-21
small tidy up

pouvik:
While this is not massive it looks better and is correct etiquette to have the correct code.  This also creates a unified approach to br's

Signed-off-by: Deaks <bryandeakin30@gmail.com.com>


Oldiesmann:
We don't use the ` />` with img or hr so why with br? The only place I'm finding `<br />` where it's unnecessary is a single line in Sources/ManageErrors.php.

SSI.php
```diff
@@ -523,7 +523,7 @@ function ssi_queryPosts($query_where = '', $query_where_params = array(), $query
 		censorText($row['subject']);
 		censorText($row['body']);
 
-		$preview = strip_tags(strtr($row['body'], array('<br>' => '&#10;')));
+		$preview = strip_tags(strtr($row['body'], array('<br />' => '&#10;')));
```
albertlast, line 526: html5 define ``<br>`` as valid
and ``<br/>`` is wrong.

albertlast (changes requested):

pouvik:
to be precise and I can give multiple examples of this `<br />` is as accepted as much as `<br>` in html 5

The idea behind the PR is to have a uniformed approach so all br tags are the same, as it is mixed up quite abit depending where you look.  I am more than happy to change  existing br / to b and check for the other tags to make the consistent. However having both br / and br is not good practice.

albertlast:
why then not the other way around?
and change all ``<br />`` to ``<br>``

pouvik:
i am doing it now also going to checdk for others while im at it :)

PR #6134, opened 2020-05-22
<hr> and <br>

pouvik:
updating all br / and hr / tags to br and hr to bring some uniformity.

signed off by me


pouvik:
this is from convo from #6133 6133

Sources/Subs.php
```diff
@@ -3291,7 +3291,7 @@ function parsesmileys(&$message)
 function highlight_php_code($code)
 {
 	// Remove special characters.
-	$code = un_htmlspecialchars(strtr($code, array('<br />' => "\n", '<br>' => "\n", "\t" => 'SMF_TAB();', '&#91;' => '[')));
+	$code = un_htmlspecialchars(strtr($code, array('<br>' => "\n", '<br>' => "\n", "\t" => 'SMF_TAB();', '&#91;' => '[')));
```
Gwenwyfar: Is this one right? Seems like it was correct as it was both `<br / >` and `<br>`

Themes/default/Xml.template.php
```diff
@@ -107,7 +107,7 @@ function template_modifytopicdone()
 	}
 	else
 		echo '
-		<error in_subject="', $context['message']['error_in_subject'] ? '1' : '0', '"><![CDATA[', cleanXml(implode('<br />', $context['message']['errors'])), ']]></error>';
+		<error in_subject="', $context['message']['error_in_subject'] ? '1' : '0', '"><![CDATA[', cleanXml(implode('<br>', $context['message']['errors'])), ']]></error>';
```
Gwenwyfar, line 110: This is an implode, so maybe it was also correct?

Sources/Subs.php
pouvik: possibly is, I just done a search :) ill fix now

Themes/default/Xml.template.php
pouvik: this one I would think is fine as it outputing a break before error list (could be wrong just how I read the code)

Sesquipedalian (changes requested):
Sorry, but most of these are mistaken changes that would cause bugs. The only correct ones are:

- the change to the img tag on line 1455 of Subs.php
- the changes to the 'cust_skype' field in the installer and upgrader files
- the changes to readme.html

Please revert all other changes.

Sesquipedalian:
Unfortunately, your most recent commit to this pull request deleted a file that should not have been deleted. Because things have gotten a bit messy here, I am going to close this pull request and ask you to submit a new one that implements only the changes that were approved in this one.

PR #6135, opened 2020-05-23
Move -you_sure- from javascript to class

XinYenFon:


SychO9:
Perhaps we should also use `JavaScriptEscape()` on the other text strings used withing the `data-confirm` attribute while we're at it

XinYenFon:
I'd much rather prefer to handle that issue in different PR by different person. The only reason I'm doing this so javascript ($li['javascript'] won't overwrite my custom class for quickbuttons).

SychO9:
Fair enough :)

Sesquipedalian (approved):
The question is, why wasn't it already like this?

PR #6140, opened 2020-05-24
Enable file_hash for people requires to do thing with attachment

XinYenFon:
Dunno if this posses any security risk directly but, this will enable people to do some modules (like MarkDown parsing via Attachments etc...)

### DIFF
--- a/Sources/Subs-Attachments.php
+++ b/Sources/Subs-Attachments.php
@@ -1159,6 +1159,7 @@ function loadAttachmentContext($id_msg, $attachments)
 			$attachmentData[$i] = array(
 				'id' => $attachment['id_attach'],
 				'name' => preg_replace('~&amp;#(\\d{1,7}|x[0-9a-fA-F]{1,6});~', '&#\\1;', $smcFunc['htmlspecialchars']($attachment['filename'])),
+				'file_hash' => $attachment['file_hash'],
 				'downloads' => $attachment['downloads'],
 				'size' => ($attachment['filesize'] < 1024000) ? round($attachment['filesize'] / 1024, 2) . ' ' . $txt['kilobyte'] : round($attachment['filesize'] / 1024 / 1024, 2) . ' ' . $txt['megabyte'],
 				'byte_size' => $attachment['filesize'],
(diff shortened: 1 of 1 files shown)


DiegoAndresCortes:
Maybe add **fileext** too?

XinYenFon:
afaik SMF holds all attachments in .dat format, not in original format.

DiegoAndresCortes:
Well not sure really, but it's still a value in the database for when you download the attachment I think

SychO9:
> afaik SMF holds all attachments in .dat format, not in original format.

that's not what he meant, smf saves the extension of the attachment in the `fileext` column

Sesquipedalian:
Can you give a bit more explanation of why you'd need the file hash, @XinYenFon? That's really only meant for internal processing, and I'm having trouble imagining a situation in which something like a Markdown parsing mod would need the hash rather than just the URL. I'm not saying no, but I'll need some more info before I can decide whether to say yes.
> Maybe add **fileext** too?

That sounds like a reasonable idea, sure.

XinYenFon:
SMF pushed items to download not to view, you can't use PDF viewer via attachments for a reason, and parser can't parse the content since it forces to download, that's why you need to go root folders and access directly.

SychO9:
So it's so that you can directly access the file from the browser using the file hash ??

XinYenFon:
> So it's so that you can directly access the file from the browser using the file hash ??

No, I use path to access the files, -just double checked- SMF already disallows you to do direct access to attachment files even if you know the full path
> eg: ../attachments/1_8239ba7c49e755371598e8290f884a0804ba5483.dat

This should give better idea, how I do it -warning eye bleeding moment-

```
	// I have good & bad ideas, this one seems orrible!
	if (!empty($message['attachment']))
	{
		global $boarddir;

		foreach ($message['attachment'] as $attachment)
		{
			// I have good ideas and bad ideas, this feel like horrible but oh well...
			if ($attachment['file_ext'] == "md" || $attachment['file_ext'] == "mdx") {
				require_once("./parsedown.php");
				$guide_raw_file = $boarddir. '/attachments/' . $attachment['id'] . '_' . $attachment['file_hash'] . '.dat';
				$guide = file_get_contents($guide_raw_file);
				$guide_output[] = Parsedown::instance()->text($guide);
			}
		}
	}
```

SychO9:
Ah, in that case we do not need to load `file_hash` for that, just use the `getAttachmentFilename()` function, use it like so:

```php
$filepath = getAttachmentFilename($attachment['filename'], $attachment['id_attach'], $attachment['id_folder']);
```

it will give you the full path to the attachment file.

XinYenFon:
well, single_post does not load those options but, nvm I'll handle this via custom query.

PR #6150, opened 2020-06-03
Remove some deprecated code from createList

dragomano:
I found a code that is currently not used anywhere. The styles used (.mirrortab_back, .maintab_back, etc.) are missing not only in SMF 2.1 but in 2.0 too (I only found them in the CSS file SMF 1.1.x).

### DIFF
--- a/Sources/Subs-List.php
+++ b/Sources/Subs-List.php
@@ -266,10 +266,6 @@ function createList($listOptions)
 	if (isset($listOptions['javascript']))
 		$list_context['javascript'] = $listOptions['javascript'];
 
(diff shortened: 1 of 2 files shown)


Oldiesmann:
Those should still be used in the "Core" theme in 2.0 (since "Core" was the default theme in 1.1)

dragomano:
> Those should still be used in the "Core" theme in 2.0 (since "Core" was the default theme in 1.1)

Well, it means that the Core theme does not have required classes (.mirrortab_back, .maintab_back, etc.) in css files

PR #6164, opened 2020-06-25
make lngfile shorter so as not to exceed max index bytesize limit

### LINKED ISSUE #6163: quanda table lngfile key too long
#### Description
### Steps to reproduce
1. use a MySQL utf8mb4 db
2. use the quanda feature

- Version/Git revision:
- Database Type: MySQL
- Database Version: > 5.7 
- PHP Version:

live627:
fixes #6163

### DIFF
--- a/.github/config.yml
+++ b/.github/config.yml
@@ -1,0 +1,7 @@
+contact_links:
+  - name: SMF Community Support
+    url: https://www.simplemachines.org/community/index.php#category_3
+    about: Please ask and answer questions here.
+  - name: SMF Security Issue
+    url: https://www.simplemachines.org/about/smf/security.php
+    about: Please report security vulnerabilities here.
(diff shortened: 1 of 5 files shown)


albertlast:
why 30?

MissAllSunday:
I would leave it a 249, it doesn't hurt the index and we never know if sometime in the future we need to actually use a ridiculous long filename

Sesquipedalian (changes requested):
> I would leave it a 249, it doesn't hurt the index and we never know if sometime in the future we need to actually use a ridiculous long filename

live627:
Now it matches the members table.

MissAllSunday:
If we are going to pair both quanda and members table lngfile column I would set it up higher than 30, the index on both tables isn't involved in any complex queries and it will still fix the main issues reported.

Issue #6072, opened 2020-04-17
The database value you're trying to insert does not exist: real_name_high


sbulen:
#### Description
Error message: 
The database value you're trying to insert does not exist: real_name_high

### Steps to reproduce
1. Go to Members | View All Members
2. Click on the last page in a long list - specifically the one after the ellipsis:
![image](https://user-images.githubusercontent.com/23568484/79527078-932e9f80-801b-11ea-81ee-c41bd15a62a9.png)

3. You get the error:
![image](https://user-images.githubusercontent.com/23568484/79527113-a6416f80-801b-11ea-95dd-10f6f05d263f.png)

### Environment (complete as necessary)
- Version/Git revision: Current
- Database Type: MySQL
- Database Version: 5.7
- PHP Version: 7.3

### Additional information/references
I believe this is an old 2.0 issue as well.  This is the first time I can consistently reproduce it.

live627:
Can you show the query?

live627:
Whats are the contents of `$memberlist_cache`, `$first_offset`, and `$second_offset`? I'm guessing that an offset wasn't calculated correctly.

sbulen:
Here 'tiz:
![image](https://user-images.githubusercontent.com/23568484/85767824-42978a80-b6cd-11ea-8650-1eec9891261d.png)

Note I sometimes see a related notice in the log:
![image](https://user-images.githubusercontent.com/23568484/85767666-15e37300-b6cd-11ea-92f1-db7899dd5db4.png)

sbulen:
And here's the actual original error from the log:
![image](https://user-images.githubusercontent.com/23568484/85768499-f3058e80-b6cd-11ea-969a-9f1e619ca51a.png)

sbulen:
Does a second offset even make sense on the last page?

PR #6166, opened 2020-06-25
Pass $start to the memberlist columns

live627:
Issue found while looking into #6072

### DIFF
--- a/Sources/Memberlist.php
+++ b/Sources/Memberlist.php
@@ -264,7 +264,7 @@ function MLAll()
 	// Sort out the column information.
 	foreach ($context['columns'] as $col => $column_details)
 	{
-		$context['columns'][$col]['href'] = $scripturl . '?action=mlist;sort=' . $col . ';start=0';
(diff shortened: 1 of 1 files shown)


jdarwood007:
Is $_REQUEST['start'] sanitized?

Sources/Memberlist.php
```diff
@@ -426,7 +426,7 @@ function MLSearch()
 		// Build the column link / sort information.
 		foreach ($context['columns'] as $col => $column_details)
 		{
-			$context['columns'][$col]['href'] = $scripturl . '?action=mlist;sa=search;start=0;sort=' . $col;
+			$context['columns'][$col]['href'] = $scripturl . '?action=mlist;sa=search;start=' . $_REQUEST['start'] . ';sort=' . $col;
```
Sesquipedalian: It looks like `$_REQUEST['start']` has not yet been sanitized at this point in `MLSearch()`. It should probably be cast to an integer before being used here.

Sesquipedalian (changes requested):

Sesquipedalian (approved):

PR #6169, opened 2020-06-27
Fix undefined index at last page of memberlist

### LINKED ISSUE #6072: The database value you're trying to insert does not exist: real_name_high

sbulen:
Fixes #6072

### DIFF
--- a/Sources/Memberlist.php
+++ b/Sources/Memberlist.php
@@ -199,6 +199,7 @@ function MLAll()
 				'last_update' => time(),
 				'num_members' => $smcFunc['db_num_rows']($request),
 				'index' => array(),
+				'max_offset' => 0,
 			);
 
 			for ($i = 0, $n = $smcFunc['db_num_rows']($request); $i < $n; $i += $cache_step_size)
@@ -209,6 +210,10 @@ function MLAll()
 			$smcFunc['db_data_seek']($request, $memberlist_cache['num_members'] - 1);
 			list ($memberlist_cache['index'][$i]) = $smcFunc['db_fetch_row']($request);
 			$smcFunc['db_free_result']($request);
+
+			// Need to know the highest index value so we don't navigate past it...
+			if (!empty($memberlist_cache['index']))
+				$memberlist_cache['max_offset'] = max(array_keys($memberlist_cache['index']));
 
 			// Now we've got the cache...store it.
 			updateSettings(array('memberlist_cache' => $smcFunc['json_encode']($memberlist_cache)));
@@ -304,7 +309,11 @@ function MLAll()
 	if ($use_cache && $_REQUEST['sort'] === 'real_name' && !isset($_REQUEST['desc']))
 	{
 		$first_offset = $_REQUEST['start'] - ($_REQUEST['start'] % $cache_step_size);
+		if ($first_offset < 0)
+			$first_offset = 0;
 		$second_offset = ceil(($_REQUEST['start'] + $modSettings['defaultMaxMembers']) / $cache_step_size) * $cache_step_size;
+		if ($second_offset > $memberlist_cache['max_offset'])
+			$second_offset = $memberlist_cache['max_offset'];
 
 		$where = 'mem.real_name BETWEEN {string:real_name_low} AND {string:real_name_high}';
 		$query_parameters['real_name_low'] = $memberlist_cache['index'][$first_offset];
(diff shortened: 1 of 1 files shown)


sbulen:
Just an FYI - this same code & issue exists in 2.0 (it's reported regularly in my forum's logs...).  I don't think it's worth fixing in 2.0, though.

sbulen:
This bug has been oddly interesting & tricky to reproduce.  It's a function of the # of users you have and your defaultMaxMembers (members to display per page).  That second_offset calculation sometimes goes one cachesize (hardcoded to 500) too high, due to the ceiling function in the calculation.  This is easy to reproduce if you have very high defaultMaxMembers, say 200 or 300.  But, depending on the # of active users total, it can happen even at much lower values of defaultMaxMembers, _if you end up with only ONE member on the very last page..._

Bottom line, the solution is to set the max when the cache is built & make sure you never exceed it.

Sources/Memberlist.php
```diff
@@ -304,7 +304,11 @@ function MLAll()
 	if ($use_cache && $_REQUEST['sort'] === 'real_name' && !isset($_REQUEST['desc']))
 	{
 		$first_offset = $_REQUEST['start'] - ($_REQUEST['start'] % $cache_step_size);
+		if ($first_offset < 0)
+			$first_offset = 0;
```
Sesquipedalian, line 308:
```suggestion
		$first_offset = max(0, $_REQUEST['start'] - ($_REQUEST['start'] % $cache_step_size));
```

Sources/Memberlist.php
```diff
@@ -304,7 +304,11 @@ function MLAll()
...
 		$second_offset = ceil(($_REQUEST['start'] + $modSettings['defaultMaxMembers']) / $cache_step_size) * $cache_step_size;
+		if ($second_offset >= $memberlist_cache['num_members'])
+			$second_offset = $memberlist_cache['num_members'] - 1;
```
Sesquipedalian, line 311:
```suggestion
		$second_offset = min($memberlist_cache['num_members'] - 1, ceil(($_REQUEST['start'] + $modSettings['defaultMaxMembers']) / $cache_step_size) * $cache_step_size);
```

Sources/Memberlist.php
```diff
@@ -319,6 +323,8 @@ function MLAll()
 		if ($first_offset < 0)
 			$first_offset = 0;
 		$second_offset = ceil(($memberlist_cache['num_members'] - $_REQUEST['start']) / $cache_step_size) * $cache_step_size;
+		if ($second_offset >= $memberlist_cache['num_members'])
+			$second_offset = $memberlist_cache['num_members'] - 1;
```
Sesquipedalian, line 327:
```suggestion
		$second_offset = min($memberlist_cache['num_members'] - 1, ceil(($memberlist_cache['num_members'] - $_REQUEST['start']) / $cache_step_size) * $cache_step_size);
```

Sesquipedalian (changes requested):
Using min() and max() will make this cleaner. 🙂

Sources/Memberlist.php
sbulen: Cleaner?  I disagree.  It is much easier to read what it's doing with the comparison to 0...

Sources/Memberlist.php
sbulen: Again, I disagree...  It is not at all clear to see what you're aiming at here...  The suggested is very simple...   

Syntax-wise, 6 of one, half dozen of the other.  Readability-wise, I think we should not make the change.

Sources/Memberlist.php
sbulen: Same as above.

Sources/Memberlist.php
Sesquipedalian: _De gustibis non est disputandum_, I suppose. I personally find min() and max() to be clear and concise, and I prefer to avoid unnecessary `if` statements when I can. But if you have strong feelings about this, I won't dig my heels in. :)

sbulen: I think it's "gustibus"?

Not strong feelings, just don't like unnecessary changes.  Especially the requisite retests.  (And I honestly think it's less readable...  😉)

I admit my OCD side wants to redo it...  Although these range checks work, in a way, they just mask the real issues with the calc.  I think there are multiple problems: I suspect an off-by-one issue which is why the ceiling sometimes goes too high, DESC sorts have a different count of names on the last (now first...) page, & it fails pretty bad with large page sizes.   

But...  It works.   And the new range checks prevent it from erroring out no matter what.  And there are very few reports of these errors on the support forum anyway.  

I.e., it's a nit...

PR #6170, opened 2020-06-28
Update pg upgrade 2.1 file with pg 9.6 syntax

albertlast:
Since of the drop pg support of 9.4 and 9.5 we can use "new" stuff and
this is one of them.
When the columns already exists,
we got no error and the db will do other add column beside the one which alrady exists.

this feature exists since 9.6.

### DIFF
--- a/Sources/DbSearch-postgresql.php
+++ b/Sources/DbSearch-postgresql.php
@@ -34,17 +34,7 @@ function db_search_init()
 
 	db_extend();
 
-	//pg 9.5 got ignore support
-	$version = $smcFunc['db_get_version']();
-	// if we got a Beta Version
-	if (stripos($version, 'beta') !== false)
-		$version = substr($version, 0, stripos($version, 'beta')) . '.0';
-	// or RC
-	if (stripos($version, 'rc') !== false)
-		$version = substr($version, 0, stripos($version, 'rc')) . '.0';
-
-	if (version_compare($version, '9.5.0', '>='))
-		$smcFunc['db_support_ignore'] = true;
+	$smcFunc['db_support_ignore'] = true;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


albertlast:
@MissAllSunday @Sesquipedalian   this pr is essential for rc3.

MissAllSunday:
All upgrade related issues has been defered to RC4 to help us test them better before Final.

albertlast:
well than is the upgrade to rc3 unsave,
see #6255

MissAllSunday:
Yes, as with any RC product, SMF RC3 will not be bug free.

tinoest:
Whilst I don’t disagree with updating, what version of pgsql does SMF 2.0 support as the highest. That needs to be the minimum version supported by SMF 2.1.

albertlast:
i guess smf 2.0 support pg 13, so we should raise to pg 13?
Feels for me a little unreasonable but i could change it.

tinoest:
That wasn’t quite what I meant, if SMF 2.0 only supports to 9.4. SMF 2.1 has to support this, if not you can’t easily upgrade.

There needs to be an overlap.

sbulen:
I've run many pg 2.0 => 2.1 upgrades with the existing code - these aren't bugs so much as rewrites. 

If a bug, how to reproduce?  What's the error message?

albertlast:
Upgrade is designed multiple times,
when you try this in actual version you will stuck on step 4.

And a long list of errors listed down after this when you check your pg log's.
So it's hard to know if everything get covered when the error log get full.

sbulen:
I think this should have been tested & gone in with the other 9.6 changes...

sbulen (approved):
LGTM.  Tested in pg & mysql.

PR #6171, opened 2020-06-29
Remove db_native_replace logic and code pg 9.6

albertlast:
Remove the fallback code of "db_native_replace",
since it's always ther.


sbulen (approved):
LGTM.  Tested upgrade & install & all works well.

PR #6173, opened 2020-06-30
Your DBMS is not a good random number generator

sexwarrior:
This commit refactors `generateValidationCode()` to rely solely on PHP's preferred source of entropy.

The current approach is not only inefficient, but comes with virtually no guarantee of producing an unpredictable code. Database management systems are good at, well, managing databases. They're not really intended as part of your pseudo-random number generator.

Signed-off by: Miłosz Gaczkowski <milosz@omgomg.eu>

### DIFF
--- a/Sources/Subs-Db-postgresql.php
+++ b/Sources/Subs-Db-postgresql.php
@@ -330,9 +330,6 @@ function smf_db_query($identifier, $db_string, $db_values = array(), $connection
 
 	// Special queries that need processing.
 	$replacements = array(
-		'get_random_number' => array(
-			'~RAND~' => 'RANDOM',
-		),
 		'insert_log_search_topics' => array(
 			'~NOT RLIKE~' => '!~',
 		),
--- a/Sources/Subs-Members.php
+++ b/Sources/Subs-Members.php
@@ -1518,24 +1518,11 @@ function populateDuplicateMembers(&$members)
 /**
  * Generate a random validation code.
  *
- * @todo Err. Whatcha doin' here.
- *
  * @return string A random validation code
  */
 function generateValidationCode()
 {
-	global $smcFunc, $modSettings;
-
-	$request = $smcFunc['db_query']('get_random_number', '
-		SELECT RAND()',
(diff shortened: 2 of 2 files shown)


albertlast:
You could than also remove this lines:
https://github.com/SimpleMachines/SMF2.1/blob/2139a62abcac785300fbbcd1768a64fd413f347b/Sources/Subs-Db-postgresql.php#L333-L335

sexwarrior:
> You could than also remove this lines

Agreed.

Oldiesmann:
random_bytes is only available in PHP 7 or higher, while SMF currently still supports PHP 5.x.

sexwarrior:
> 
> 
> random_bytes is only available in PHP 7 or higher, while SMF currently still supports PHP 5.x.

Fair enough, I thought that was just 2.0. My bad.

It looks like SMF 2.1 is already falling back on a [userland implementation](https://github.com/paragonie/random_compat) of `random_bytes()` for PHP5. A copy is [already present in Sources](https://github.com/SimpleMachines/SMF2.1/tree/ed54e36517065b22151783a6b246347efa14e286/Sources/random_compat) and [a helper function exists to handle the fallback](https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Load.php#L260-L271).

It seems like the correct solution would therefore be to just use `$smcFunc['random_bytes']()`, analogous to things like [our current password salt generation](https://github.com/SimpleMachines/SMF2.1/blob/dcbbf523fa90d0b90f3ddd5a2e464f23f3111e4e/Sources/LogInOut.php#L120).

Does that sound about right?

Sesquipedalian (approved):
> It seems like the correct solution would therefore be to just use `$smcFunc['random_bytes']()`, analogous to things like [our current password salt generation](https://github.com/SimpleMachines/SMF2.1/blob/dcbbf523fa90d0b90f3ddd5a2e464f23f3111e4e/Sources/LogInOut.php#L120).
> 
> Does that sound about right?

Yes, it does. 👍

PR #6175, opened 2020-07-11
Undesirable results when a file doesn't exist when loading css or js.

jdarwood007:
If you call loadCSS('foo.bar'), the HTML output will be:
`	<link rel="stylesheet" href="?smf21rc2_#TIMESTAMP#">`

This is wrong.  This is produced because we only check that $fileName is not empty, not $fileUrl.  I'm not 100% sure if the intention to be able to call these without a valid file is intended, such as in the case of wanting to trigger the updateSettings call.  So instead of return, I just modified the check.

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -2387,7 +2387,7 @@ function loadTheme($id_theme = 0, $initialize = true)
 			$_SESSION['id_variant'] = $_REQUEST['variant'];
 		// User selection?
 		if (empty($settings['disable_user_variant']) || allowedTo('admin_forum'))
-			$context['theme_variant'] = !empty($_SESSION['id_variant']) ? $_SESSION['id_variant'] : (!empty($options['theme_variant']) ? $options['theme_variant'] : '');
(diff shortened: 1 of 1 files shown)


Sources/Load.php
```diff
@@ -2703,22 +2703,19 @@ function loadCSSFile($fileName, $params = array(), $id = '')
 				$fileUrl = $settings['default_theme_url'] . '/css/' . $fileName;
 				$filePath = $settings['default_theme_dir'] . '/css/' . $fileName;
 			}
-
```
MissAllSunday, line 2706: Might be good to not remove the empty lines to help with human readability.

Sources/Load.php
```diff
@@ -2387,7 +2387,7 @@ function loadTheme($id_theme = 0, $initialize = true)
 			$_SESSION['id_variant'] = $_REQUEST['variant'];
 		// User selection?
 		if (empty($settings['disable_user_variant']) || allowedTo('admin_forum'))
-			$context['theme_variant'] = !empty($_SESSION['id_variant']) ? $_SESSION['id_variant'] : (!empty($options['theme_variant']) ? $options['theme_variant'] : '');
+			$context['theme_variant'] = !empty($_SESSION['id_variant']) && in_array($_SESSION['id_variant'], $settings['theme_variants']) ? $_SESSION['id_variant'] : (!empty($options['theme_variant']) && in_array($options['theme_variant'], $settings['theme_variants']) ? $options['theme_variant'] : '');
```
MissAllSunday, line 2390: Perhaps splitting the multiple ternaries into separate vars to help with readability.

PR #6177, opened 2020-07-17
This fixes #6176 and #6180 - Major Upgrade Bug Fixes

### LINKED ISSUE #6176: Upgrade Script Error
Error
#### Description
I have a very large smf 1.1.21 forum of over 1.5 million users.  I know I should've upgraded awhile ago due to security risks etc.  However, I have been very lazy.  I kinda got some energy lately to redo my old theme into the new SMF 2.1 latest (github) version and finally push the newest SMF running on the newest PHP 7.2.

### Steps to reproduce
1. Have a large forum > 1.5 million users
2. Upgrade from 1.1.21 to 2.1 (github version)

- Version/Git revision: Latest SMF on github
- Database Type: MySQL
- Database Version: mysql  Ver 15.1 Distrib 10.2.19-MariaDB,
- PHP Version: 7.2

2020/07/17 02:06:50 [error] 9726#9726: *1574950 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 16777224 bytes) in /var/www/**REMOVED**/**REMOVED**/public_html/upgrade.php(2134) : eval()'d code on line 33" while reading response header from upstream, client: **REMOVED**, server: **REMOVED**, request: "GET /newtest998/upgrade.php?step=3&amp;substep=0&amp;data=eyJjdXJzdGVwIjozLCJsYW5nIjoiZW5nbGlzaCIsInJpZCI6MTEyNiwicGFzcyI6Njg5MCwiZGVidWciOnRydWUsImpzIjoxfQ==&xml&filecount=3&substep=91&_=1594963621919&strict_step=51 HTTP/1.1", upstream: "fastcgi://unix:/var/run/php-fpm/php-fpm.sock:", host: "**REMOVED**", referrer: "https://**REMOVED**/newtest998/upgrade.php?step=1&substep=0&data=eyJjdXJzdGVwIjoxLCJsYW5nIjoiZW5nbGlzaCIsInJpZCI6MTEyNiwicGFzcyI6Njg5MCwiZGVidWciOjAsImpzIjoxfQ=="

P.S. Sorry if this is a "noob" bug report.  I am new to github.

Fix #6177

### LINKED ISSUE #6180: Avatar/Attachment Upgrade failure - only some convert correctly
#### Description
While upgrading from 1.1.21 the users avatars were only partially updated from attachments to the new custom_avatar folder.  No errors or warnings were produced via php.

My board has around 50,000 avatars, around 30890 avatars moved to the custom_avatar folder.  The rest stayed in the attachments folder and did not set attachment_type = 1 to the other 20,000+.  Hence now 20,000+ users that had avatars no longer do.

### Steps to reproduce
1. Have large board with lots of users and attachments/avatars.
2. Upgrade from 1.1.21 to 2.1 (github version)

Version/Git revision: Latest SMF on github
Database Type: MySQL
Database Version: mysql Ver 15.1 Distrib 10.2.19-MariaDB,
PHP Version: 7.2

I will update this bug report soon when I figure out what's causing it.

Butterflysaway:
Fixes issue I published.

fixes #6176 
fixes #6180

### DIFF
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -866,7 +866,7 @@ if (in_array('notify_regularity', $results))
 	$step_progress['name'] = 'Upgrading post notification settings';
 	$step_progress['current'] = $_GET['a'];
 
-	$limit = 100000;
+	$limit = 50000;
 	$is_done = false;
 
 	$request = $smcFunc['db_query']('', 'SELECT COUNT(*) FROM {db_prefix}members');
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -314,8 +314,7 @@ if ($custom_av_dir != $GLOBALS['boarddir'] .'/custom_avatar')
 
 $request = upgrade_query("
 	SELECT COUNT(*)
-	FROM {$db_prefix}attachments
-	WHERE attachment_type != 1");
+	FROM {$db_prefix}attachments");
 list ($step_progress['total']) = $smcFunc['db_fetch_row']($request);
 $smcFunc['db_free_result']($request);
 
@@ -338,9 +337,8 @@ while (!$is_done)
 	nextSubStep($substep);
 
 	$request = upgrade_query("
-		SELECT id_attach, id_member, id_folder, filename, file_hash, mime_type
+		SELECT attachment_type, id_attach, id_member, id_folder, filename, file_hash, mime_type
 		FROM {$db_prefix}attachments
-		WHERE attachment_type != 1
 		LIMIT $_GET[a], 100");
 
 	// Finished?
@@ -349,6 +347,11 @@ while (!$is_done)
 
 	while ($row = $smcFunc['db_fetch_assoc']($request))
 	{
+		//This attachment has already been converted to the new style.
+		if($row['attachment_type'] == 1)
+		{
+			continue;
+		}
 		// The current folder.
 		$currentFolder = !empty($modSettings['currentAttachmentUploadDir']) ? $modSettings['attachmentUploadDir'][$row['id_folder']] : $modSettings['attachmentUploadDir'];
 
--- a/other/upgrade_2-1_postgresql.sql
+++ b/other/upgrade_2-1_postgresql.sql
@@ -484,8 +484,7 @@ if ($custom_av_dir != $GLOBALS['boarddir'] .'/custom_avatar')
 
 $request = upgrade_query("
 	SELECT COUNT(*)
-	FROM {$db_prefix}attachments
-	WHERE attachment_type != 1");
+	FROM {$db_prefix}attachments");
 list ($step_progress['total']) = $smcFunc['db_fetch_row']($request);
 $smcFunc['db_free_result']($request);
 
@@ -508,9 +507,8 @@ while (!$is_done)
 	nextSubStep($substep);
 
 	$request = upgrade_query("
-		SELECT id_attach, id_member, id_folder, filename, file_hash, mime_type
+		SELECT attachment_type, id_attach, id_member, id_folder, filename, file_hash, mime_type
 		FROM {$db_prefix}attachments
-		WHERE attachment_type != 1
 		LIMIT $_GET[a], 100");
 
 	// Finished?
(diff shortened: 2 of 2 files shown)


Butterflysaway:
Hello @live627 please make sure this gets included in the next SMF 2.1 if one ever gets released.  I think the developers need to add another upgrade which fixes all past upgrades to 2.1 for attachments.  This bug looks like it's been there since the beginning of 2.1.  I think you could easily add a function to recheck for attachments/avatars which failed to update.

albertlast:
First it's not smart two combine to fixes in one pr
secondly why you extend the query about attachment_type and later on you skip attachment_type 1 ?

Butterflysaway:
> First it's not smart two combine to fixes in one pr
> secondly why you extend the query about attachment_type and later on you skip attachment_type 1 ?

1. It wouldn't have merged both pull requests if the first one was merged into smf 3 days ago when I posted the bug and fix that I personally encountered. Plus I'm new to github, and I don't fully understand/care enough to figure out how to separate pull requests from one file on the same fork.

2. If you can't see why this is a problem then that's why this bug has been plaguing smf 2.1 since the beginning of 2.1.
Before this change my attachment folder of 80k attachments/avatars only converted around 1/2 properly.  After this change all of my attachments/avatars are converted properly.

Arantor (changes requested):
Instead of doing this, you should be able to fix the problem by way of instead forcing attachments to be handled in a consistent order (i.e. applying a ORDER BY to the processing query that steps through in groups). I don't believe this fixes whatever problem you actually have.

albertlast:
1. you will fine pr from me how more than one year old...

2. you second fix is none fix, i guess you run this so many time that it at some point got done or you database is too small for the amount of data,
would be also a good explaination why you got #6176 since your setup didn't match the size of your forum.

Butterflysaway:
> 1. you will fine pr from me how more than one year old...
> 2. you second fix is none fix, i guess you run this so many time that it at some point got done or you database is too small for the amount of data,
>    would be also a good explaination why you got #6176 since your setup didn't match the size of your forum.

I don't want to fight with you.  I honestly don't care if neither of these fixes get pushed or not.  It's not my job to fix smf.  I can tell though by the amount of bugs and lack of project leadership that this will likely never get finished and it would be in the best interest of my users to just use another forum platform.  However, I'm 100% confident these fixes solved my issue as I re-ran the updater on a clean copy of my 1.1.21 forum files/database several times to check.  I might just be better off forking 1.1.21 and pushing fixes to make it compatible with the latest version of php if that's how you treat people fixing your own problems here.

albertlast:
i'm not part of the smf team, should be easy to since the collaborator text beside my name.

The order by hint from arantor looks more in right direction to possible fix the issue,
also you forget postgres upgrade here to fix too.

Butterflysaway:
> Instead of doing this, you should be able to fix the problem by way of instead forcing attachments to be handled in a consistent order (i.e. applying a ORDER BY to the processing query that steps through in groups). I don't believe this fixes whatever problem you actually have.

I guess if you would like to develop a different fix that works the same way then that's fine.  I have confirmed for myself at least that this fixes the problem.  This bug affects every prior version of smf when upgrading attachments/avatars to 2.1.  You can test this if you have a 1.1.21 or 2.0.17 forum with a few hundred attachments/avatars.

This bug does not happen when you do not have any avatars and just a bunch of attachments.  It has to be a mix of attachments/avatars in the attachment folder/database.

Arantor:
In other words it works by accident for you because you have MyISAM tables and lowering the block size is why it works. You could, of course, just change the query to include ORDER BY id_attach and get all of the fix to work without *any* other changes.

Butterflysaway:
> In other words it works by accident for you because you have MyISAM tables and lowering the block size is why it works. You could, of course, just change the query to include ORDER BY id_attach and get all of the fix to work without _any_ other changes.

Pretty sure that will not work as you are later setting attachment_type = 1 which changes the total size of the rows being selected each time the updated avatar is changed to attachment_type = 1.  Either way I don't want to spend more time retesting something that works just fine in this pr.

fyi my tables are InnoDB

jdarwood007:
So the upgrade logic here is to rename all non avatars (avatars are attachment_type = 1)

First the attachment_type logic seems to be
0 == attachment
1 == avatar
3 == thumbnail

I can't find a usage of 2.

A way to fix this would be to the search where file name NOT LIKE '%.dat' as well, and then walk over that until we get no results.  Only issue is with a large forum there would be no index on the file name and this could be slow.

sbulen:
> ... you are later setting attachment_type = 1 which changes the total size of the rows being selected each time the updated avatar is changed to attachment_type = 1.

Butterflysaway is correct here. This explains why the check for type makes more sense within the processing loop than in the query.  

But others are also correct here, in that using a LIMIT without an ORDER BY may yield questionable navigation thru the set as well.  Lots of articles out there on that (& in fact I think that has bitten us before).  

So...  I think we need a companion PR to this that adds an ORDER BY here, and we should test the two PRs together.  I'll label this as final & work on doing so.

PR #6184, opened 2020-07-22
Add order by at paging process in the upgrade

albertlast:
Since we do on some places paging so it would be better,
also to add an order by statement to this --> used always the pk field.

i expect a decrease(get worst) of the performance,
since the order by force db to do stuff which they didn't do before.


Sesquipedalian:
Hey @sbulen, when you get a chance, could you review this?

sbulen (approved):
LGTM.  Tested in pg & mysql.

PR #6185, opened 2020-07-22
Sync PG upgrade logic member values with mysql

albertlast:
While adding the order by pr,
i notice the member value logic is different between pg and mysql upgrade,
since the code block got no mysql special i guess it safe to do.

what me wonders,
why was yim als possible columns in pg old upgrade in mysql not?


live627:
>  why was yim als possible columns in pg old upgrade in mysql not?

Looks like it was forgotten in https://github.com/SimpleMachines/SMF2.1/commit/447138ee98ad36a2719e71b8ea9b22ec3ce061b4

live627 (approved):

PR #6190, opened 2020-07-22
Speed up order upgrade

albertlast:
This would be an advanced version of #6184 ,
with the result that the upgrade process get faster instead of slower actual state.

This pr is mostly a show case...

170ba15 include the changes which we could do on all places where the upgrade process do paging.
In general is the idea to make advantage about the existing of the pk index to jump faster to next point,
also since it should safe it only use the methode so long no page refresh is happening,
when a page refresh happen than the next loop is done in "normal" mode with the result that the next loops again go fast.

The choosen example i had to admit make less sense since it got a high limit amount..
so maybe places like 
https://github.com/SimpleMachines/SMF2.1/blob/8104e4328223106289fcb7719f85cb5ce88b1b1a/other/upgrade_2-1_mysql.sql#L340-L344
where the limit is 100 and the possible rows can got up to 80k entries than the optimization make more sense.

So the question is,
Want we this?
Do we want this every where?

### DIFF
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -341,6 +341,7 @@ while (!$is_done)
 		SELECT id_attach, id_member, id_folder, filename, file_hash, mime_type
 		FROM {$db_prefix}attachments
 		WHERE attachment_type != 1
+		ORDER BY id_attach
 		LIMIT $_GET[a], 100");
 
 	// Finished?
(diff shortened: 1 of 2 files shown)


Sesquipedalian:
@sbulen, could you review this one, too?

@albertlast, it sounds like this PR and #6184 are alternatives. Which one are you actually suggesting we should merge?

albertlast:
well i would take #6184 since it do basic stuff for both database,
this here added a advanced  version for pg -> no effect on mysql

sbulen:
We should go with #6184 .  This is the right idea - but we should target known problem areas performance-wise, and at the moment, we're not seen any reported.

PR #6193, opened 2020-07-25
Update the credits

Arantor:


Sources/Who.php
```diff
@@ -899,4 +898,4 @@ function Credits($in_admin = false)
 	}
 }
 
-?>
\ No newline at end of file
+?>
```
MissAllSunday, line 901: As per: https://github.com/SimpleMachines/BuildTools/blob/3aff818c7ab802cb256aa540d2c15a28579f3e16/check-eof.php#L73  please remove the extra line at the end of the file to avoid a broken build.

MissAllSunday (changes requested):

Sources/Who.php
Arantor: Unfortunately that's not an option with the built-in editor in GitHub which automatically adds a line to the end of the file.

I guess we should probably close this PR because this can't be fixed in GitHub and I honestly can't be bothered to clone the repo to work around GitHub adding it automatically (there's no option to remove it)

PR #6195, opened 2020-07-25
Update the Credits

XinYenFon:


Sources/Who.php
```diff
@@ -899,4 +898,4 @@ function Credits($in_admin = false)
 	}
 }
 
-?>
\ No newline at end of file
+?>
```
MissAllSunday: As per: https://github.com/SimpleMachines/BuildTools/blob/3aff818c7ab802cb256aa540d2c15a28579f3e16/check-eof.php#L73  please remove the extra line at the end of the file to avoid a broken build.

MissAllSunday (changes requested):

MissAllSunday (approved):

PR #6199, opened 2020-07-26
Fix news queries fixing #6198

albertlast:
Do two things,
first fixes #6198
secondly re order the on statement to this schema  (base_table.col_name = join_table.col_name)
this standard in the sql world...

PR #6202, opened 2020-07-30
Join On base table on left side in news

albertlast:
move the base table to the left side of the on statement like mention here: #6199


jdarwood007:
Just to news queries here or are we talking going though all sources globally to cleanup all of them?

albertlast:
i would like to stick the scope of this pr to news.

Sesquipedalian (approved):

PR #6206, opened 2020-07-31
Modsettings cache fix

### LINKED ISSUE #5826: Infrequent caching error for modSettings

sbulen:
Fixes #5826 

This addresses the T_CONSTANT_ENCAPSED_STRING problem we see in 2.1 as well as 2.0 production.  For some forums, all users are locked out until someone clears the cache...

To reproduce the error & test a fix, I wrote two bonehead scripts - one that writes cache 1000x and one that reads cache 1000x.  Running both does not in itself reproduce the issue, however, running TWO of the PUT scripts causes all kinds of issues with the GET scripts (confirming Suki's idea that it's related to conflicting writes).

_What we want to see in these tests is 1000 successful reads, with few, if any, NULLs returned..._  

WAMP & Linux fail differently here.

(This is an alternate approach to #5900 )

**_BEFORE THIS PR - WIN:_**
Although this always successfully completes on WIN, there are a LOT of NULLS returned...  LOTS...  
smf21-current-code-WIN

**_BEFORE THIS PR - linux:_**
This NEVER returns the 1000 reads - ever - on linux.  These failures kill the script, as it is a PARSE error.  There are usually (not always) accompanied with the T_CONSTANT_ENCAPSED_STRING errors in the apache log.
smf21-current-code-linux
image

**_TESTING THIS PR - WIN:_**
Now it completes all 1000 reads successfully, with rare NULLs returned (~1 per 1000).  
smf21-sbfix-WIN
smf21-sbfix-WIN-err

**_TESTING THIS PR - linux:_**
Now it completes all 1000 reads successfully, with rare NULLs returned (~1 per 1000).  
smf21-sbfix-linux

### DIFF
--- a/Sources/CacheAPI-smf.php
+++ b/Sources/CacheAPI-smf.php
@@ -69,7 +69,13 @@ class smf_cache extends cache_api
 				@apc_delete_file($cachedir . '/data_' . $key . '.php');
 
 			// php will cache file_exists et all, we can't 100% depend on its results so proceed with caution
-			@include($cachedir . '/data_' . $key . '.php');
+			$cached = file_get_contents($cachedir . '/data_' . $key . '.php');
+			if (!empty($cached) && (substr($cached, 0, 6) === '<' . '?' . 'php ') && (substr($cached, -2, 2) === '?' . '>'))
+			{
+				$cached = substr($cached, 6, strlen($cached) - 8);
+				eval($cached);
+			}
+
 			if (!empty($expired) && isset($value))
 			{
 				@unlink($cachedir . '/data_' . $key . '.php');
--- a/Sources/CacheAPI-smf.php
+++ b/Sources/CacheAPI-smf.php
@@ -69,17 +69,18 @@ class smf_cache extends cache_api
 				@apc_delete_file($cachedir . '/data_' . $key . '.php');
 
 			// php will cache file_exists et all, we can't 100% depend on its results so proceed with caution
-			$cached = file_get_contents($cachedir . '/data_' . $key . '.php');
-			if (!empty($cached) && (substr($cached, 0, 6) === '<' . '?' . 'php ') && (substr($cached, -2, 2) === '?' . '>'))
+			$cached = json_decode(file_get_contents($cachedir . '/data_' . $key . '.php'), TRUE);
+			if (!empty($cached['expired']) && isset($cached['value']))
 			{
-				$cached = substr($cached, 6, strlen($cached) - 8);
-				eval($cached);
-			}
-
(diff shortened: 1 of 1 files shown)


jdarwood007:
eval is sometimes disabled, is there another choice?

sbulen:
eval()s are used throughout SMF, in 2.0 & 2.1...  They're used in load.php, subs.php, subs-list.php, themes, installer, & upgrader and more...

The only alternative I could think of would be to just parse it all out.  That wouldn't be too difficult.  Just wondering if it's really necessary.

jdarwood007:
Well I remember the Theme related one is able to be disabled and a performance reason for doing that.  Lists I don't believe are natively used in our base code.
If nobody has concerns I would merge it.  Just to me it sounds like it could cause other issues including performance related ones.

sbulen:
I am testing a new version that does away with putting logic in the file altogether...  Why do we do an include/eval at all?

Testing thus far shows a pretty dramatic improvement!  Instead of "very few" NULL reads, I now get zero...  In fact, I can dramatically increase the speed at which I can throw cache PUTs & GETs at SMF and still get zero NULL reads on linux.  (My wamp gets more NULLs as I speed it up, but it's still a small fraction of the existing logic, which as noted above, was almost 50%...)  

I am not so sure eval had any worse performance than include - they both have to invoke the compiler.  

The biggest problem with eval (& include...) is security.   Changing the approach tightens that up, too.  

A new commit forthcoming...  Still testing...

jdarwood007:
I like it.  Let json decoder handle validating it.  If it fails, it returns NULL anyways and the cache falls through as it should.  No PHP errors.

live627:
> I am not so sure eval had any worse performance than include - they both have to invoke the compiler.

Because we invalidate the file in the opcache... identical performance I think.

> My wamp gets more NULLs as I speed it up

Reading incomplete data. Check the filesize (add code in getData() to confirm)

sbulen:
Removed unnecessary old code per discussion over in #6211 

Turns out that file_read_contents does no lock checking whatsoever, and is thus vulnerable to returning NULL erroneously.  There are numerous articles about this, here is an example:
https://stackoverflow.com/questions/49262971/does-phps-file-get-contents-ignore-file-locking

The technique shared in this article was applied here, basically, just put a wrapper with locks around the call to file_get_contents.  

_Now, all puts & gets return 100% solid reads & writes, throughout extensive testing, across wamp & linux._

Issue #6207, opened 2020-07-31
Registration issue


Butterflysaway:
#### Description
While testing the entire forum for bugs before I push to my live site I noticed when I registered on the forum it registered 2 new members with the same name, password, different verification codes, password salts and id's, also sent out 2 emails to the same email address for both accounts.

![register_bug](https://user-images.githubusercontent.com/68415596/89088673-a3b71b80-d367-11ea-880d-46c89d72150f.png)

### Steps to reproduce
I'm still not sure how to reproduce this bug as I tried again and it only registered once, but I know for sure this has to be a bug somewhere for some reason.

### Environment (complete as necessary)
- Version/Git revision:2.1 Latest github version
- Database Type: mysql
- Database Version: mysql  Ver 15.1 Distrib 10.2.19-MariaDB, for Linux (x86_64) using readline 5.1
- PHP Version: php 7.4

### Additional information/references
No error messages or warnings were given.

Butterflysaway:
Can someone tell me if this has happened to them before?

Butterflysaway:
Turns out I must have accidentally pressed register 2 times.  I am able to reproduce it only if I press sign up 2x very fast on chrome.  I think if I enable reCaptcha that should fix the problem since each submit token has to be verified by google.

albertlast:
on db side this issue could be catched by creating an unique index over the member_name or email_address.

Butterflysaway:
Yeah I can also confirm this is a bug spanning all the way back to 1.1.21 as it happens on that version as well.

Turns out my database had around 700 duplicate registrations from 1.1.21

albertlast:
What is for you a duplicate?
the same name? the same mail? or something else?

Butterflysaway:
same name and email.  smf by default does not allow you to register another account with the same name nor email address, but this bug allows it to happen.

Butterflysaway:
@albertlast could this be fixed by using **replace** instead of **insert** for the user registration query?

albertlast:
Yes and no,
yes it would solve this issue,
no it would allow to overtake an existing account.

so in general NO would be right answer.

What is in your point of view the issue with unique index here?

Butterflysaway:
I have no issue with making them unique indexes, but before that could even be released for everybody you'd have to remove already duplicated entries of email addresses and usernames or MySQL would throw an error while trying to alter the table right?

albertlast:
[...]

PR #6208, opened 2020-08-01
Check in the welcome screen if duplicate members exists

albertlast:
Since we believe that the member_name and email_address is unique we should check this.
got the idea by reading the issue: https://github.com/SimpleMachines/SMF2.1/issues/6207

The result is shown in the welcome screen:
grafik

### DIFF
--- a/Themes/default/languages/Install.english.php
+++ b/Themes/default/languages/Install.english.php
@@ -371,6 +371,14 @@ $txt['error_dir_not_writable'] = 'The directory: %1$s has to be writable to cont
 $txt['error_cache_not_found'] = 'The cache directory could not be found.<br><br>Please make sure you have a directory called &quot;cache&quot; in your forum directory before continuing.';
 $txt['error_agreement_not_writable'] = 'The upgrader was unable to obtain write access to agreement.txt.<br><br>If you are using a linux or unix based server, please ensure that the file is chmod\'d to 777, or if it does not exist that the directory this upgrader is in is 777.<br>If your server is running Windows, please ensure that the internet guest account has the proper permissions on it or its folder.';
 $txt['error_not_admin'] = 'You need to be an admin to perform an upgrade!';
+$txt['error_duplicate_members'] = 'Found multiple username or mail address in members table, please use this database queries to find and solve them: <br> select * from %1$smembers as a
+<br> join (select member_name, count(*) from %1$smembers group by member_name having count(*) > 1) as b on (a.member_name = b.member_name)
+<br> order by a.member_name
+<br> <br> or for mail address:<br>
+select * from %1$smembers as a
+<br> join (select email_address, count(*) from %1$smembers group by email_address having count(*) > 1) as b on (a.email_address = b.email_address)
+<br> order by a.email_address
+';
 
 $txt['warning_lang_old'] = 'The language files for your selected language, %1$s, have not been updated to the latest version. Upgrade will continue with the forum default, %2$s.';
 $txt['warning_lang_missing'] = 'The upgrader could not find the &quot;Install&quot; language file for your selected language, %1$s. Upgrade will continue with the forum default, %2$s.';
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -878,6 +878,26 @@ function WelcomeLogin()
 	if (!$create || !$alter || !$drop)
 		return throw_error(sprintf($txt['error_db_privileges'], $databases[$db_type]['name']));
 
+	// Check for duplicate userdata in members table
+	$request = $smcFunc['db_query']('',
+		'select member_name, count(*) from {db_prefix}members group by member_name having count(*) > 1');
+	
+	if ($smcFunc['db_num_rows']($request) != 0)
+		$duplicateName = true;
+		
+	$smcFunc['db_free_result']($request);
+	
+	$request = $smcFunc['db_query']('',
+		'select email_address, count(*) from {db_prefix}members group by email_address having count(*) > 1');
+	
+	if ($smcFunc['db_num_rows']($request) != 0)
+		$duplicateMail = true;
+	
+	if (isset($duplicateName) || isset($duplicateMail)){
+		global $db_prefix;
+		return throw_error(sprintf($txt['error_duplicate_members'], $db_prefix));
+	}
+
 	// Do a quick version spot check.
 	$temp = substr(@implode('', @file($boarddir . '/index.php')), 0, 4096);
 	preg_match('~\*\s@version\s+(.+)[\s]{2}~i', $temp, $match);


frandominguezl:
I don't think we want to tell admins that may or may not have experience running database queries to run certain queries on their system. There should be another, more appropiate solution to this problem.

albertlast:
Than provide a better solution.

frandominguezl:
I analyzed the behaviour of the issue.

When you click register two times really fast, you complete the registration proccess and end up logged in with the account of highest ID. Let's say, if both accounts created have IDs 5 and 6, right after the registration you will be logged in with ID 6.

If you logout and then login with the account's credentials, you will be logged in with ID 5.

We can expect to be the account with higher activity the one with the lowest ID, so we, by default should tell the user what's the problem and how we meant to solve it (by deleting the _last_ account of each duplicate). If the user agrees, he can proceed with the proccess. If not, we can suggest as well to ask for support.

The main option shouldn't be telling the user to fix the problem by themselves.

albertlast:
So we should change the error text and remove the sql statement part?

frandominguezl:
Yes. We should inform the user what we're going to do, and then he'll be able to decide if he wants to continue with the upgrade process or not.

MissAllSunday:
Along side a backend fix I would also like to add a frontend fix. Perhaps a checkSubmitOnce() call and disable the button via JS on first click.

Butterflysaway:
@MissAllSunday this sounds like a great first check tbh.  That would fix 99.9% of all future duplicate registrations.  Only bots or someone intentionally trying to do that would succeed after this change.

Butterflysaway:
I will make a script to automatically remove duplicates while keeping their active version of their account if they have one.  If they have not activated either and never logged in either it shouldn't matter which account you keep of theirs.  Not sure what the best option would be if they've logged into both accounts and posted before, perhaps check the latest logged in one and keep that and remove the others.

Butterflysaway:
@frandominguez03 
Just to add - this bug can be affected by x amount of duplicate entries.  I found up to 9x duplicates in my database. with a total of 744 members containing duplicates of a total of 1618 entries.

Maybe something like this:

This simply removes any entries where id_member > first id_member and member_name is identical.

There may be times when this does not catch all duplicate email_addresses if someone changed their member_name after signing up with the duplicate accounts if you allow them in your board to do so.

You guys might want to implement a better way to delete the accounts so you can remove any posts, personal messages, alerts, etc. 

I ran this on my board to clear up the duplicates.  You can use this and make a pull request if you want or modify it in anyway.  Just a suggestion on what i personally did.

	$firstEntry = array();
	$request = $smcFunc['db_query']('', "
		SELECT duplicates.id_member, duplicates.member_name
		FROM {db_prefix}members AS duplicates
		WHERE duplicates.member_name IN (
			SELECT member_name
			FROM {db_prefix}members
			GROUP BY member_name
			HAVING COUNT(member_name) > 1
		) ORDER BY member_name ASC, id_member ASC;
	");
	{
		if(empty($firstEntry[$row['member_name']]))
		{
			$firstEntry[$row['member_name']] = $row['id_member'];
		}
	}
	foreach ($firstEntry as $k => $v)
	{
		$smcFunc['db_query']('', "
			DELETE FROM {db_prefix}members
			WHERE member_name = {string:member_name} AND id_member > {int:id_member}",
			array(
				'member_name' => $k,
				'id_member' => $v,
			)
		);
	}

albertlast:
Since you got a modern mariadb version you can us this query to delete:
```
delete from smf_member where id_member in (
	select id_member from (
		select member_name, 
		id_member, 
		row_number() over(partition by member_name order by id_member) num 
		from smf_members
	) a
	where  a.num > 1
)
```
replace member_name with email_address and done.

since we support rdbms from the stoneage a smf solution would be this:
```
delete from smf_member where id_member in (
	select id_member from (
		select a.id_member 
		from smf_members as a
		join (
			select member_name, 
			min(id_member) id_member, 
			count(*)
			from smf_members
			group by member_name
			having count(*) > 1
		) as b on (a.member_name = b.member_name and a.id_member > b.id_member)
	)
)
```

albertlast:
When should the cleaning in the upgrade process happen,
should be this happen than autmaticaly or should some king of button for cleaning process?

Butterflysaway:
@albertlast I believe it should all take place in RC3.  This way even the people who use current github repo will be able to upgrade using the official smf upgrade files once that version is pushed.  This isn't a huge problem, just an annoying one.  I would personally take a much more substantial approach to filtering out which accounts to remove which includes removing posts, personal messages, alerts, etc.  I don't plan on removing the duplicated users on my live version anytime soon at least until this is officially added to smf.  As the bug spans back to the creation of SMF (1.0), then virtually every board that runs smf has this problem and it needs to be addressed properly.  The easiest way I've found to stop this from happening again is enabling recaptcha v2 on the forum as it prevents the same form from being submitted 2 times which should be sufficient for now until the permanent fix is done.

albertlast:
This was not the quest,
i asked for which step in the upgrade process should this happen and how.

MissAllSunday:
I would prefer to add a scheduled task to handle this scenarios. 

I personally wouldn't run any query directly on the upgrade process. Check and detect, yes  and If something is detected, inform the user about it and continue with the process.

Once the process is complete, a reminder on the admin panel to run that scheduled task to take care of those duplicated entries.

albertlast:
so this wouldn't fix the issue,
would be only a workaround.

MissAllSunday:
The check is already there, so are the txt strings that can be modified to inform the user. The only thing left is building the scheduled task.

jdarwood007:
@MissAllSunday  If your going to put the unique index on there, you can't have a scheduled task fix this.  This will cause the operation to fail during the upgrade.

MissAllSunday:
Does the upgrade process gets interrupted if duplicates are found?  I still don't think the admin should be doing this during the upgrade, it is already convulsive enough to add yet another layer of complexity on top of it.

albertlast:
Than came up with a better idea and
not decline the first one so long.

in short the upgrade process get not interpruted, since the check is placed in the welcome step.

MissAllSunday:
The idea is fine. I want to know if duplicate ids are a problem doing the upgrade or if they prevent it, if not, this process can be moved to an admin section where the admin can decide when to do this rather than doing it on the upgrade.

albertlast:
This pr could changed the way to delay, but the related #6209 expect this happen before the upgrade process is started.

MissAllSunday:
Cool, this is what I wanted to know. Now, how about do the check, if there are duplicates store the result somewhere in a modSettings value then add an admin interface to let the admin do the process whenever its more convenient to them.

We already have some admin tools for recounting stuff, convert to utf8, prune logs etc etc. WE can add an interface for admins to add the unique index and also offer some more alternatives on how they want to manage duplicates, something like:


We found 3 duplicates which one you want to keep or merge them into a single account.

I think we can offer more tools and options if we move this to an admin page rather than doing it on the upgrade process.

albertlast:
Than please provide a admin page where i could lay down this feature.
Or tell me which of the existing one should be used.

MissAllSunday:
There is none. A new interface will have to be added, if you could add it great if not we can do it but it will have low priority.

albertlast:
Since it's your idea,
i let you doing the thing,
please highlight me back when your interface is done.

Sesquipedalian:
This pull request targets release-2.1, so it doesn't belong in the 3.0 milestone, and we aren't going to make this change now in the 2.1 upgrader.

That said, if a new PR implementing these changes and the ones in #6209 is submitted for release-3.0, that would be worthwhile.

PR #6211, opened 2020-08-02
This is my implementation of a JSON cache

### LINKED ISSUE #5826: Infrequent caching error for modSettings

live627:
Fixes #5826

- encodes the data with the JSON lib 
- removes the risky code evaluation that sometimes threw fatal errors
- use a reader lock to avoid reading exclusively locked files
- don't block the script if any lock could not be obtained

[A user comment on php.net about how file locks work](https://www.php.net/manual/en/function.flock.php#78318)

### DIFF
--- a/Sources/Cache/APIs/FileBased.php
+++ b/Sources/Cache/APIs/FileBased.php
@@ -55,6 +55,57 @@ class FileBased extends CacheApi implements CacheApiInterface
 		return parent::isSupported() && $supported;
 	}
 
+	private function readFile($file)
+	{
+		if (($fp = fopen($file, 'rb')) !== false)
+		{
+			if (!flock($fp, LOCK_SH | LOCK_NB))
+			{
+				fclose($fp);
+				return false;
+			}
+			$string = '';
+			while (!feof($fp))
+				$string .= fread($fp, 8192);
+
+			flock($fp, LOCK_UN);
+			fclose($fp);
+
+			return $string;
+		}
+
+		return false;
+	}
+
+	private function writeFile($file, $string)
+	{
+		if (($fp = fopen($file, 'cb')) !== false)
+		{
+			if (!flock($fp, LOCK_EX | LOCK_NB))
+			{
+				fclose($fp);
+				return false;
+			}
+			ftruncate($fp, 0);
+			$bytes = 0;
+			$pieces = str_split($string, 8192);
+			foreach ($pieces as $piece)
+			{
+				if (($val = fwrite($fp, $piece, 8192)) !== false)
+					$bytes += $val;
+				else
+					return false;
+			}
+			fflush($fp);
+			flock($fp, LOCK_UN);
+			fclose($fp);
+
+			return $bytes;
+		}
+
+		return false;
+	}
+
 	/**
 	 * {@inheritDoc}
 	 */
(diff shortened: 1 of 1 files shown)


Sources/CacheAPI-smf.php
```diff
@@ -49,67 +49,107 @@ public function isSupported($test = false)
...
+		if (file_exists($file) && ($raw = $this->readFile($file)) !== false)
 		{
-			// Work around Zend's opcode caching (PHP 5.5+), they would cache older files for a couple of seconds
-			// causing newer files to take effect a while later.
-			if (function_exists('opcache_invalidate'))
-				opcache_invalidate($cachedir . '/data_' . $key . '.php', true);
-
-			if (function_exists('apc_delete_file'))
-				@apc_delete_file($cachedir . '/data_' . $key . '.php');
-
```
jdarwood007: I think we still need these.  On systems where this would take effect, you have made it so it would keep rebuilding the cache file until it catches.  So in effect it could run the code it should be skipping dozens of times before the opcache/apc clears out.  To help simplify this, we could drop it into a sub method.

I understand we don't do it when clearing the entire cache folder.  Something should be done for that as well.

sbulen:
This tests OK in my tests - it definitely solves the original problem.  Two notes:
 - The reads return NULL sometimes, for both wamp & linux (due to...)
 - I don't think you want the LOCK_NBs

LOCK_NB causes execution to continue even if the file is unavailable.  For reads, since the return value is checked, what ends up happening is it is treated as a failed read.  This is why there are occasional NULLs returned.  Note the calling function will detect the lack of a cache & rebuild the cache unnecessarily.  I think the wait is preferred, i.e., no LOCK_NB, so the read is solid & the cache is used.

On the writes, with the LOCK_NB, you are going to return false & attempt to delete whatever was out there.  It will require a subsequent call to update the cache.  I think the wait is preferred, i.e., no LOCK_NB, because it will only take one update.

If you remove the LOCK_NBs, it will wait as normal & complete the read or write.  I have confirmed you get no NULLs if you remove them.  At that point, however, you have basically rewritten file_put_contents & file_get_contents...  

I think it would be better to remove the LOCK_NBs ~~& just use file_put_contents & file_get_contents~~ (edit: see below).  It's a bit more solid & simple.

sbulen:
I've been doing more testing...

I had only tested the _modified_ PR on linux earlier...  I just tested it on wamp now and it is perfect on wamp as well as linux - zero NULLs.  _But ONLY if you get rid of the LOCK_NBs..._  So, if modified, it may in fact have an advantage over file_put_contents & file_get_contents (which are perfect on linux but occasionally return NULLs on wamp).  I don't know why, but it does.

I'm still testing, but so far, it looks good!

Sources/CacheAPI-smf.php
live627: These calls are no longer needed because these files are no longer compiled and cached as bytecode

live627:
I think I may have another approach which is free of locks.

albertlast:
Until php 8.0 json function can be disabled right?

live627:
SMF 2.1 does JSON everywhere

albertlast:
i forgot, right

Sources/CacheAPI-smf.php
jdarwood007: Your right.  I was thinking it still gets cached, but the only one would be the file system caching it.  We still may want to call clearstatcache() then in order to inform php to no longer cache it on that side.

sbulen:
I tested the latest, and it has the same issues as Suki's version:  The cache PUT collisions take place on the rename:
image

This happens on lamp, linux seems to behave better.

As noted earlier, what worked PERFECT was the prior version of this PR, but without the LOCK_NBs (but keeping the other locks).

sbulen:
Just retested, using the latest & greatest, & this PR now has 100% success on PUTs & GETs, across wamp & linux as well!  

This has actually worked out well, learning what worked & what didn't across multiple PRs, trying different approaches to the same problem.

live627:
ok then I'll now leave it as is since it works.

Just some documentation left....

MissAllSunday (approved):

live627:
The rebase was worth it!

PR #6222, opened 2020-08-09
DeleteDrafts() only accepts one at a time

live627:


Butterflysaway:
This isn't a problem.  The DeleteDraft function allows an array of int's.

/**
 * Deletes one or many drafts from the DB
 * Validates the drafts are from the user
 * is supplied an array of drafts will attempt to remove all of them
 *
 * @param int $id_draft The ID of the draft to delete
 * @param boolean $check Whether or not to check that the draft belongs to the current user
 * @return boolean False if it couldn't be deleted (doesn't return anything otherwise)
 */

WHERE id_draft IN ({array_int:id_draft})'

live627:
You're right...  thanks for checking.

PR #6231, opened 2020-08-19
Strip -utf8 string from privacy policy settings

### LINKED ISSUE #6238: Registration - should we rename 2.0 agreements that have -utf8 in the name?
#### Description
If your 2.0 site is multi-language, and has -utf8 languages installed, it will have several .txt files that have -utf8 in the filenames.  In 2.1, everything is -utf8, so we no longer include -utf8 in the language names.  The Upgrader has gone out of its way to find all settings, etc., and rename them, to remove the -utf8.  

Should we rename these .txt files as well?  Or should we rely on the admins just recognizing this & renaming the files themselves?  

I wasn't quite sure if this was overkill or not, hence I'm asking the question...  If you weren't aware of this, it will look like you LOST all your old agreements on the 2.1 install, because they won't be found...

These 2.0 agreements won't be found in 2.1 after the upgrade unless renamed:
image

sbulen:
This is a companion PR to #6100 

Fixes #6238 

The privacy policies are language-specific, and include the languages in the setting names in the settings table.  If the language was utf8, it had a -utf8 suffix on the language name in 2.0.  Since -utf8 is stripped from language names & settings during the upgrade from 2.0 to 2.1, it now has to be stripped from the privacy settings as well, otherwise there are disconnects between the settings & the app for privacy policies entered in 2.0, & you lose privacy policies entered in 2.0.  

This issue was discovered during testing of #6100.  

Feedback welcome.  

Tested in mysql 5.7 & pg 10.

### DIFF
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -3151,11 +3151,12 @@ ALTER TABLE {$db_prefix}log_spider_stats CHANGE page_hits page_hits INT NOT NULL
 /******************************************************************************/
 ---# Strip -utf8 from policy settings
 ---{
-$utf8_policy_settings = array_filter($modSettings, function($v, $k)
-		{
-			return (substr($k, 0, 7) === 'policy_') && (substr($k, -5) === '-utf8');
-		}, ARRAY_FILTER_USE_BOTH
-	);
+$utf8_policy_settings = array();
+foreach($modSettings AS $k => $v)
+{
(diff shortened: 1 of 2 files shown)


sbulen:
Added data fix for missing id_member fields.

sbulen:
Added step to rename agreements

sbulen:
Curse you, 5.4...

MissAllSunday (approved):

PR #6244, opened 2020-09-01
RC3

MissAllSunday:
Updates the forum version to RC3

### DIFF
--- a/Themes/default/languages/Agreement.english.php
+++ b/Themes/default/languages/Agreement.english.php
@@ -1,5 +1,5 @@
 <?php
-// Version: SMF 2.1 RC2; Agreement
+// Version: 2.1 RC3; Agreement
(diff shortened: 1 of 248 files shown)


Themes/default/languages/Agreement.english.php
```diff
@@ -1,5 +1,5 @@
 <?php
-// Version: SMF 2.1 RC2; Agreement
+// Version: 2.1 RC3; Agreement
```
live627, line 2: Nice!

live627 (approved):
(and yes, I scrolled through the whole thing)

Themes/default/languages/Agreement.english.php
MisterVector: +1

PR #6247, opened 2020-09-03
Better connect errors

### LINKED ISSUE #6242: Mysql 8.0: connect issue - "authentication method unknown to the client [caching_sha2_password]"
#### Description
From this thread:
https://www.simplemachines.org/community/index.php?topic=574573.msg4066640#msg4066640

Two issues:
 - The error shown to the user provides no helpful info.  Had to insert a "die(mysqli_error($connection));" in there to get a meaningful error message to work with.  
 - The hidden error is:

> mysqli_real_connect(): The server requested authentication method unknown to the client [caching_sha2_password]

The fix was to alter the user:
mysql> ALTER USER root IDENTIFIED WITH mysql_native_password BY 'PASSWORD';

I believe this is the 2nd time we've seen this on the support forum - can't find the earlier thread...

- Database Type: MySQL
- Database Version: 8.0
- PHP Version: 7.3

Just looking at this the first time, but it looks like some mysql 8.0 instances need to be told to use the pre-8.0 "native" style of authentication:
https://dev.mysql.com/doc/refman/8.0/en/create-user.html
https://dev.mysql.com/doc/refman/8.0/en/authentication-plugins.html
https://dev.mysql.com/doc/refman/8.0/en/native-pluggable-authentication.html

sbulen:
Fixes #6242 

This PR:
 - Gets & displays error #s and error message texts in the event of a db connection issue
 - Displays #s and errors in the installer & upgrader - admin functions only
 - Removes one of the 'second guessing' steps in the installer, where it would try variants to login; the problem was the error messages would reflect the behind-the-scenes guesswork, not what the user entered...

There actually was an attempt to report connection errors in there, but it was returning empty strings...

PG was difficult, since connection error info is not available via calls such as pg_last_error.  Had to insert a try/catch around the pg connect, & save off the error details.

Feedback welcome.

**_EDIT: Based on input below, left public-facing application error messaging alone._**

### **_Connection error messages, BEFORE..._**
**_Installer:_**
21-before-installer

**_Upgrader:_**
21-before-upgrader

### **_Connection error messages, AFTER, MySql..._**
**_Installer:_**
21-after-installer

**_Upgrader:_**
21-after-upgrader

### _**Connection error messages, AFTER, PG...**_
**_Installer:_**
21-pg-after-installer

**_Upgrader:_**
21-pg-after-upgrader

### DIFF
--- a/Sources/Errors.php
+++ b/Sources/Errors.php
@@ -425,11 +425,19 @@ function display_maintenance_message()
  */
 function display_db_error()
 {
-	global $mbname, $modSettings, $maintenance;
+	global $mbname, $modSettings, $maintenance, $smcFunc;
 	global $db_connection, $webmaster_email, $db_last_error, $db_error_send, $smcFunc, $sourcedir, $cache_enable;
 
 	require_once($sourcedir . '/Logging.php');
 	set_fatal_error_headers();
+
+	// Get error info...  Recast just in case we get false or 0...
+	$error_message = $smcFunc['db_connect_error']();
+	if (empty($error_message))
+		$error_message = '';
+	$error_number = $smcFunc['db_connect_errno']();
+	if (empty($error_number))
+		$error_number = '';
 
 	// For our purposes, we're gonna want this on if at all possible.
 	$cache_enable = '1';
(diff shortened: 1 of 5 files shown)


Oldiesmann:
I believe the whole reason for the generic "Connection problems" screen was to avoid showing sensitive database info to non-admins (for security reasons). It's the same reason we don't show database error messages to non-admins when we can connect to the database.

sbulen:
Yep.  Good point.

sbulen:
@Oldiesmann - So remove the connection error details from the app one, but leave it on the installer & upgrader (admin functions)?

Oldiesmann:
That's what I would recommend

PR #6255, opened 2020-09-09
Fixing upgrade

albertlast:
- Missing IF NOT EXISTS for create sequence (work with pg 9.6+)
- added missing pk fields in insert
- convert insert to insert with ignore

For a full upgrade test, the pr #6170 is needed.


sbulen (approved):
LGTM.  Tested upgrade & install & all works well.

PR #6260, opened 2020-09-11
Add title and buttons to admin and mod log

albertlast:
Base is for me the error log:
grafik

Atm the mod log look like this:
grafik
we notice missing title and missing button on right upper side
with the pr it's look like this:
grafik

### DIFF
--- a/Sources/Modlog.php
+++ b/Sources/Modlog.php
@@ -43,7 +43,7 @@ function ViewModlog()
 
 	$context['can_delete'] = allowedTo('admin_forum');
 
-	loadLanguage('Modlog');
+	loadLanguage('Admin+Modlog');
 
 	$context['page_title'] = $context['log_type'] == 3 ? $txt['modlog_admin_log'] : $txt['modlog_view'];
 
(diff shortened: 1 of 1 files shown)


StealthWombat:
Non-critical, but a good idea.

dragomano:
It should to add `loadLanguage('Admin+Modlog');` into Line 46 to prevent possible errors.

albertlast:
under which circumstance the line is needed?

dragomano:
1. Go to Moderate -> Moderation log
2. Then we see a new undefined error in the Error log:
`Undefined index: moderation_log`
3. Replace `loadLanguage('Modlog');` with `loadLanguage('Admin+Modlog');` and we do not see this error anymore.

MissAllSunday:
@albertlast Can you add the change @dragomano suggested?

Sources/Modlog.php
```diff
@@ -163,6 +163,7 @@ function ViewModlog()
 	// This is all the information required for a watched user listing.
 	$listOptions = array(
 		'id' => 'moderation_log_list',
+		'title' => $context['log_type'] === 3 ? $txt['admin_log'] : $txt['moderation_log'],
```
MissAllSunday: Since we historically haven't perform strict checks and due to the volatile nature of global vars, please use == insead of ===

MissAllSunday (changes requested):

MissAllSunday (approved):

MissAllSunday:
Thanks!

PR #6262, opened 2020-09-12
Fixes issues with auth_secret and upgrader

### LINKED ISSUE #6261: Missing $auth_secret prevents logins
#### Description
As expected if $auth_secret is missing, we attempt to populate Settings.php with it.  But if we can't write this Settings.php we silently fail.  This will cause unexpected silent login errors since $auth_secret keeps changing, the cookie will never be valid.

We should either:
1. When $db_show_debug = $true;, show a warning if we are missing $auth_secret and can't write to Settings.php
2. Show a warning on login if $auth_secret is missing

### Steps to reproduce
1. remove $auth_secret from Settings.php
2. Set Settings.php to read only
3. Attempt to login

- Database Type: mysql
- Database Version:
- PHP Version: 7.4

jdarwood007:
This PR got a little bigger than I expected.  I've linked it as a RC4 PR for the sake of releasing RC3.

The initial part of the PR fixes #6261 
First we attempt to fix this during upgrades, but also allow the admin to randomize the auth_secret during upgrade.

Second attempt is during normal usage. We log messages indicating it is missing but only once every 15 minutes. For admins we issue a security warning as well.

As a fallback, we set secret_auth to a static value. This weakens security, but at least allows logins to occur normally.

This also fixes some editor handling issues in Subs-Admin.php with regards to ?>
As well we have updated updateDbLastError to return a bool on success/failure and optionally updates the $db_last_error into the current context, allowing if needed for other parts of the system or 3rd party APIs to recognize we have an error that was written in the current context.

The second part of this PR is fixing multiple upgrader issues I found while making the first fix.
First:
PHP Notice: Undefined index: lang_character_set in /upgrade.php on line 687
PHP Stack trace:
PHP 1. {main}() /upgrade.php:0
PHP 2. loadEssentialData() /upgrade.php:188

We resolve this by loading the index language as well. We also give a CLI friendly error message if we can't load a language.

Second:
If we specify --path=, this causes some failures in upgrade.php. I've corrected them
1. $upgrade_path is defined in the global context, so we bring it into functions where it is needed and replace usages of dirname(__FILE__) otherwise. $upgrade_path defaults to this but CLI allows us to override it.
2. Also fixes it for language path error message.

Third:
If Settings.php or Settings_bak.php can't be written, we fail. This should be writable, especially for CLI case as it should be a shell account with write access to these files, even if the web user does not have write access.

Finally:
Updated the CLI options to include rebuilding settings, resetting auth and allowing stat collection. Updated the help topic as well.


Sources/Load.php
```diff
@@ -3846,17 +3846,32 @@ function set_avatar_data($data = array())
  */
 function get_auth_secret()
 {
-	global $auth_secret, $sourcedir, $smcFunc;
+	global $context, $auth_secret, $sourcedir, $smcFunc, $db_last_error, $txt;
```
Sesquipedalian:
```suggestion
	global $context, $auth_secret, $sourcedir, $boarddir, $smcFunc, $db_last_error, $txt;
```

See below for explanation.

Sources/Load.php
```diff
@@ -3846,17 +3846,32 @@ function set_avatar_data($data = array())
...
 
 		// It is important to store this in Settings.php, not the database.
 		require_once($sourcedir . '/Subs-Admin.php');
-		updateSettingsFile(array('auth_secret' => $auth_secret));
+
+		// Did this fail?  If so, we should alert, log and set a static value.
+		if (!updateSettingsFile(array('auth_secret' => $auth_secret)))
+		{
+			$context['auth_secret_missing'] = true;
+			$auth_secret = 'none';
```
Sesquipedalian:
```suggestion
			$auth_secret = hash_file('sha256', $boarddir . '/Settings.php');
```

This will keep the $auth_secret value unique to each forum, even if not random. Since the issue is that Settings.php is not writable, this approach should keep the value stable at least until the admin fixes the permissions problem.

Themes/default/languages/Install.english.php
```diff
@@ -276,6 +276,7 @@
 $txt['upgrade_stats_collection'] = 'Allow Simple Machines to collect basic stats monthly.';
 $txt['upgrade_stats_info'] = 'If enabled, this will allow Simple Machines to visit your site once a month to collect basic statistics. This will help us make decisions as to which configurations to optimise the software for. For more information please visit our <a href="%1$s" target="_blank" rel="noopener">info page</a>.';
 $txt['upgrade_migrate_settings_file'] = 'Migrate to a new Settings file.';
+$txt['upgrade_reset_auth_secret'] = 'Reset auth_secreet to a new value.';
```
Sesquipedalian:
```suggestion
$txt['upgrade_reset_auth_secret'] = 'Reset authentication secret to a new value.';
```

Themes/default/languages/index.english.php
```diff
@@ -474,6 +474,7 @@
 $txt['not_removed_extra'] = '%1$s is a backup of %2$s that was not generated by SMF. It can be accessed directly and used to gain unauthorized access to your forum. You should delete it immediately.';
 $txt['generic_warning'] = 'Warning';
 $txt['agreement_missing'] = 'You are requiring new users to accept a registration agreement, however, the file (agreement.txt) does not exist.';
+$txt['secret_auth_missing'] = 'Unable to set secret_auth in Settings.php, this weakens your security and puts your system at risk for attacks.';
```
Sesquipedalian:
```suggestion
$txt['secret_auth_missing'] = 'Unable to set authentication secret in Settings.php. This weakens your security and puts your system at risk for attacks. Check the file permissions on Settings.php to make sure SMF can write to the file.';
```

other/upgrade.php
```diff
@@ -1401,13 +1427,22 @@ function UpgradeOptions()
 	// Ensure this doesn't get lost in translation.
 	$changes['upgradeData'] = base64_encode(json_encode($upcontext['user']));
 
+	// Update $auth_secret?
+	if (!empty($_POST['resetAuthSecret']))
+		$changes['auth_secret'] = bin2hex($smcFunc['random_bytes'](32));
+
```
Sesquipedalian: Why would we offer this? As a rule, once the auth_secret has been set, it shouldn't be changed, even during an upgrade. The only exception to the rule would be if there were a security breach that gave the attacker read-access to Settings.php itself.

Sesquipedalian (changes requested):
I'd like to see the fallback improved as suggested below. I'm also not convinced that offering to reset the auth_secret during an upgrade is a good idea.

Themes/default/languages/index.english.php
Sesquipedalian: Also, it'd be nice to change `$txt['secret_auth_missing']` to `$txt['auth_secret_missing']` here and elsewhere, just for the sake of maintaining consistency in our variable names.

Sources/Load.php
jdarwood007: Why do I need $boarddir here?  Its not being used here.

other/upgrade.php
jdarwood007: I thought it would be good option as upgrades between major versions don't happen often, and rotating out secrets like this can be beneficial to security.

Sesquipedalian: The problem is, changing the $auth_secret value doesn't just invalidate existing cookies; it also invalidates all of the unsubscribe links we've included our email notifications ever since 2.0.16 was released and since #6012 was merged for 2.1. Those new unsubscribe links are part of GDPR compliance, since the GDPR requires us to allow someone to unsubscribe their email address from a given type of notification *without needing to log into an account.* (Perhaps that's a silly requirement, but it is what it is.) But the only way to meet that requirement even halfway securely is to include a unique token in the URL parameters, and the $auth_secret value is integral to generating and verifying those tokens. So if $auth_secret changes, the unsubscribe links break, and thus the forum runs afoul of the GDPR.

The need to keep $auth_secret invariant once created is why there's no UI to change it in the Admin Centre and why the comments in Settings.php itself strongly warn the admin against changing it manually. For the same reason, we shouldn't offer to change it during upgrade.

jdarwood007: Should we do the job right then in 2.1 and split this into a secondary secret used for just notification links?
We could add a note that this will break logins and invalidate all notification unsubscribe links which may be in violation of a GDPR or similar requirements.

If we do it for notifications, we could store it in the database.  So we could store a notifications_secret and a notifications_secret_old, allowing the old secret to work for a period of time.

Preventing rotating of secrets is a bad thing.  While it allows it to be unique among all forums, it prevents security related actions of rotating those secrets on intervals, which may be another checkbox on some security compliance form somewhere.

Sesquipedalian: Splitting it into two secrets wouldn't get us any further. It would merely leave us with two secrets that would both still be subject to the same requirements that one secret would. They'd be used in almost identical ways and stored in identical ways (neither one could be stored in the database, or else a database leak would give the attacker all the info they need to forge the login authentication, so they'd have to be next to each other in Settings.php), so any policy requiring one of them to be periodically changed would equally require the other to be periodically changed.

Storing old versions of the secret and checking against is possible in theory, I suppose, but it wouldn't accomplish much. If old keys still exist and can be used to validate incoming links or cookies, then they're still active and haven't actually been rotated out.

But regardless of the above, I don't believe there's any need to rotate this particular key anyway. It never, ever leaves the server in any form. It can't be extracted from any of the tokens or cookie values we generate and send out. The only way for an attacker to get ahold of it is to gain access to the server file system itself so they can read Settings.php, at which point it's already far too late.

jdarwood007:
Any other developers agree with not adding this to the upgrader?  Just looking to confirm if this isn't a good idea.

Sesquipedalian:
> Any other developers agree with not adding this to the upgrader? Just looking to confirm if this isn't a good idea.

Obviously, I've already argued that $auth_secret should never be changed once set. But even if my arguments about that don't carry the day, I don't think a checkbox in the upgrader makes much sense as way to do that. If it were the case that $auth_secret should be changed periodically, then an optional, one-off reset wouldn't meet that need. Instead, if periodic resets were required, then it would be better to do that automatically in the background as a part of SMF's normal operation. There still wouldn't be any need for a UI, but if one really wanted to offer a UI for it, then the right place for that would be in SMF's Security settings panel, not the upgrader.

sbulen:
The part of this PR I like is handling the language issue better.  It is definitely easy to get stuck when the forum default is not english & is not loaded, and it can get confusing to try to get past that.  

I am very leery of rewriting stuff that has been working as we inch closer to a release.  We need to be increasingly strict about what changes we introduce and why.  I am very concerned about destablizing code at this late stage.  E.g., an error every 15 minutes seems plenty???

jdarwood007:
@sbulen I think the concern is more over should we allow it to be reset on the upgrader.  This otherwise is going to solve a issue that may cause support issues with users being unable to log into the forum should $auth_secret be missing in their Settings.php and we are unable to write a new $auth_secret to Settings.php
I slated this for RC4 because while its a problem, the fix should be easy enough to test and put into RC4 with minimal concerns for breakage.  You can easily test this without patching by removing $auth_secret from your Settings.php and making it read only.  You will be unable to log into the forum.

@Sesquipedalian I'm just looking for feedback to see if others agree with your points.  Since we are on 2 different sides of this opinion, another developer input will break this stalemate and give us direction to go.  Automating it sounds like something we could look to achieve in future versions of SMF.  I think secrets like this are meant to be rotated.  Like your SSL private keys.  They should change on a periodic basis to ensure that if the secret is compromised without your knowledge, that they limit the window the secret was exposed for.
I'm fine with pulling the code from the upgrader for now and revisiting this in future versions to where we have the ability to properly code in a fully automated system to keeping secrets rotated based on things like requirements for how long unsubscribe links are to be valid and a way to resend the unsubscribe link should the link become invalid (such as some webpage with a button to resend a new unsubscribe).

sbulen:
I have not seen this problem documented anywhere.  Have seen seen it reported by users on the forum?  Did you run into it during test?  How & why did it occur?

If you can't write updates to Settings.php, that's a completely different issue, and the upgrader will fail on multiple fronts.  Settings.php gets updated multiple times throughout the upgrade process.  

This feels speculative to me.  And I really want to urge everyone towards 2.1 code stability.  We need to turn the corner to making changes that really need to be made.

jdarwood007:
Occurred on my own forum because I ran upgrade via php in CLI.  For some reason it never got its $secret_auth.  Spent some time trying to figure that out.
A edge case here would also be if they recreate their Settings.php from lets say repair_settings.php, it won't fix this.  But Settings.php may be writable in that case.

sbulen:
Just confirming, you had a 2.0 environment without an $auth_secret and the CLI failed during upgrade?  Did the CLI crash out & stop or just report the error?  

I thought the recent massive rewrite of updateSettingsFile() already provided an $auth_secret if it were missing...  E.g., I have run many CLI updates from YaBB, 1.0, 1.1 & 2.0 that don't have $auth_secret & it has always fixed it...  (Then again we see over & over again that there's a lot of variability between environments...)  

I am more curious to find out why updateSettingsFile() didn't fix it than providing a second set of logic to fix it...

jdarwood007:
I'm not entirely sure how my forum was missing the secret.  It was customized and running some includes and other calls.  Maybe that prevented it from being upgraded during the upgrade.  I can't provide you why anymore, I reset Settings to factory and injected all my changes again.

I will state, without at least the minimum of this patch doing the fallback/logging, a missing auth_secret will result in a forum in which you can not log into and the admin will not receive any information or reasoning why.  Thats the problem here.  A login loop failure occurs because we attempt to generate a new auth_secret on every page load causing the cookie to become invalid on every page load.
With this fallback, we at least fall into something we can somewhat rely on (Settings.php hash) and the admin receives error log information indicating the secret is missing, which is something they can action on, we can document on and does not prevent operation of the forum.

sbulen:
Can you reproduce it?

Sesquipedalian:
> Can you reproduce it?

> You can easily test this without patching by removing $auth_secret from your Settings.php and making it read only.  You will be unable to log into the forum.

While it is true that Settings.php needs to be writable during install and upgrade, and therefore `$auth_secret` can normally be expected to have been set successfully, it is always possible for it to be changed or removed from Settings.php afterwards by any process that can affect the file. So asking for a way to reproduce the issue via normal operation doesn't make much sense in this case. @jdarwood007's changes to `get_auth_secret()` are meant to enhance our safeguards against possibilities outside our control, not problems within our control.

> @Sesquipedalian I'm just looking for feedback to see if others agree with your points.  

Of course. 🙂 

> I'm fine with pulling the code from the upgrader for now and revisiting this in future versions to where we have the ability to properly code in a fully automated system to keeping secrets rotated based on things like requirements for how long unsubscribe links are to be valid and a way to resend the unsubscribe link should the link become invalid (such as some webpage with a button to resend a new unsubscribe).

That all sounds good to me. If we build a way to reissue unsubscribe links in a future version, that'll alleviate the need to keep `$auth_secret` invariant, opening the way for periodic automatic updates. In the meantime, if you're willing to remove the reset from the upgrader, that'll address my concerns in the present.

sbulen:
> So asking for a way to reproduce the issue via normal operation doesn't make much sense in this case. @jdarwood007's changes to `get_auth_secret()` are meant to enhance our safeguards against possibilities outside our control, not problems within our control.

You can break all kinds of things by manual edits to Settings.php...  That doesn't mean we should keep making changes to the upgrader...   If the only way to reproduce a problem is to manually break Settings.php, that is not a valid test.  It should likely be fixed in that case by repair_settings.php, or another manual edit. 

We should be restricting PRs to _TRULY NECESSARY CHANGES_.  We are trying to bring this to closure.  We will not get there if we keep thinking of good ideas and rewriting stuff.  That inherently destabilizes code, and will generate a perpetual beta.  We need to snap out of a perpetual beta mindset.  PLEASE stop & ask yourselves what is truly necessary.

Also, at this time, please don't merge anything without pinging Suki.   You need to check in on our internal threads regularly.

jdarwood007:
@sbulen I won't be merging it myself.  I will let somebody else merge it.

I can back off more of the upgrader changes if needed.

Sesquipedalian (approved):

sbulen:
Ahem...  
https://www.simplemachines.org/community/index.php?topic=575241.msg4071153#msg4071153

I suspect the issue is RC => RC upgrades?

jdarwood007:
While they are missing $auth_secret, those problems look almost like some files didn't fully update everywhere.  $smcFunc is missing htmlspecialchars from his error.  Which means either Load.php isn't updated or something is unloading it.

sbulen:
Yeah...  Something is very broken there...

Themes/default/languages/Install.english.php
```diff
@@ -276,6 +276,7 @@
 $txt['upgrade_stats_collection'] = 'Allow Simple Machines to collect basic stats monthly.';
 $txt['upgrade_stats_info'] = 'If enabled, this will allow Simple Machines to visit your site once a month to collect basic statistics. This will help us make decisions as to which configurations to optimise the software for. For more information please visit our <a href="%1$s" target="_blank" rel="noopener">info page</a>.';
 $txt['upgrade_migrate_settings_file'] = 'Migrate to a new Settings file.';
+$txt['upgrade_reset_auth_secret'] = 'Reset authentication secret to a new value.';
```
Sesquipedalian: We don't need this line now, right?

Sesquipedalian (changes requested):

Sesquipedalian (approved):

MissAllSunday:
Hi @jdarwood007 can you rebase againts release2.1 to merge this in?

MissAllSunday:
There's seems to be an issue with the merge as a lot of changes done on other commits and added to master are been included into this PR, if you could revert the merge commit and do a rebase instead I think that would solve the issue.

jdarwood007:
This was a rebase...  Maybe I did that wrong.  Don't usually rebase too often.

jdarwood007:
I totally botched something in the rebase.  Good thing I pulled a diff before I attempted the rebase.  I just made a new PR against the updated code.  Not sure why the PR was broke.  The diff applied cleanly against the latest release-2.1

MissAllSunday:
Cool thanks!

PR #6264, opened 2020-09-13
Check if port is available for Memcache connection

Underdog-01:
I use Apcu for the cache & the connection error still pops up in the SMF error log for Memcache from this file/line in RC3.
This checks if the port is available prior to attempting a connection using Memcache. 
The previous PR also worked just fine in my opinion but it was forgotten or undecided for too long.


jdarwood007:
What does your cache settings look like in the Settings.php file?  It shouldn't be attempting to connect memcache.  The isSupported method sets up that you do have support for memcache and have values in the memcache setup for your Settings.php.

Underdog-01:
Please see the image:


cache_settings

jdarwood007:
Your Settings.php contains multiple settings for caching.  Its holding some memcache settings.

```
########## Cache Info ##########
/**
 * Select a cache system. You want to leave this up to the cache area of the admin panel for
 * proper detection of apc, memcached, output_cache, smf, or xcache
 * (you can add more with a mod).
 *
 * @var string
 */
$cache_accelerator = '';
/**
 * The level at which you would like to cache. Between 0 (off) through 3 (cache a lot).
 *
 * @var int
 */
$cache_enable = 0;
/**
 * This is only used for memcache / memcached. Should be a string of 'server:port,server:port'
 *
 * @var array
 */
$cache_memcached = '';
/**
 * This is only for the 'smf' file cache system. It is the path to the cache directory.
 * It is also recommended that you place this in /tmp/ if you are going to use this.
 *
 * @var string
 */
$cachedir = dirname(__FILE__) . '/cache';
```

Underdog-01:
It looks like this:

```
########## Cache Info ##########
/**
 * Select a cache system. You want to leave this up to the cache area of the admin panel for
 * proper detection of apc, memcached, output_cache, smf, or xcache
 * (you can add more with a mod).
 * @var string
 */
$cache_accelerator = 'apcu';
/**
 * The level at which you would like to cache. Between 0 (off) through 3 (cache a lot).
 * @var int
 */
$cache_enable = 1;
/**
 * This is only used for memcache / memcached. Should be a string of 'server:port,server:port'
 * @var array
 */
$cache_memcached = '';
/**
 * This is only for the 'smf' file cache system. It is the path to the cache directory.
 * It is also recommended that you place this in /tmp/ if you are going to use this.
 * @var string
 */
$cachedir = dirname(__FILE__) . '/cache';
```

Underdog-01:
.. maybe I should use quote because the dirname underscores did not show there...
There are double underscores on either side like your code example.

.. never mind... 3 graves instead of one.

Underdog-01:
> ########## Cache Info ##########
/**
 * Select a cache system. You want to leave this up to the cache area of the admin panel for
 * proper detection of apc, memcached, output_cache, smf, or xcache
 * (you can add more with a mod).
 * @var string
 */
$cache_accelerator = 'apcu';
/**
 * The level at which you would like to cache. Between 0 (off) through 3 (cache a lot).
 * @var int
 */
$cache_enable = 1;
/**
 * This is only used for memcache / memcached. Should be a string of 'server:port,server:port'
 * @var array
 */
$cache_memcached = '';
/**
 * This is only for the 'smf' file cache system. It is the path to the cache directory.
 * It is also recommended that you place this in /tmp/ if you are going to use this.
 * @var string
 */
$cachedir = dirname(__FILE__) . '/cache';

Underdog-01:
So if the $cache_enable should be set to 0 if not using memcache or memcached... shouldn't it not state that on the settings page in the SMF Admin ?

I figured it our by looking at the Settings file remarks but in the admin that is not clear.

Just an idea for that menu:

If memcache or memcached is not available and/or is not opted by the admin then do not show the Caching Level input or disable it. If it is enabled then have it show under the cache type setting and word it as Memcache/Memcached Level.
Also perhaps it should set that to 0 if neither memcache or memcached is not enabled/available.

The way it is currently done it looks like I am disabling the cache altogether if I opt "No caching" for that setting.
.. just my opinion but perhaps you or another dev sees my perspective.

Butterflysaway:
I just added a bunch of return false on the memcache functions.  Until somebody finally fixes it that's the easiest solution if you don't plan to use memcached.

jdarwood007:
So $cache_memcached is an empty string?

MissAllSunday:
This has been addressed by #6277

Thank you for your contribution.

PR #6271, opened 2020-09-26
Add webp to images support

albertlast:
Add support for the webp image format,
since imagetype stuff on php side is used,
the needed php version is 5.6.26 or 7.0.10 (we are on 5.4.0).
https://www.php.net/manual/en/image.constants.php

Could handle this in some ways:
raise the min version (would be the right way).

Do not add this or
do some version checks.

Since both last idea are stupid i guess we will take one of this?


Oldiesmann:
Considering PHP 5.4 has been EOL for 5 years now I don't see why we can't bump the min version to 5.6 at this point. Even WordPress' stats show very little usage for PHP 5.4 and 5.5 (just 4% of those reporting info to them, compared to 12.8% for 5.6).

MissAllSunday:
Is there any particular reason to add this image format? is there a request on simplemachines.org for it? perhaps some benefits?

albertlast:
get the new image standard for web stuff.
https://torquemag.io/2020/11/webp-wordpress/
https://developers.google.com/speed/webp/docs/webp_study

MissAllSunday:
None of those links declares webp as the new image standard and this requires us to either bump the min PHP version (something that is not planned at the moment) or add extra checks which adds non needed complexity.

albertlast:
So what is in your eyes the image standard?
When you ask google they say webp.
And yes rise php version would be the best solution.

Since this is a hard reason.
Von: Michel Mendiola <notifications@github.com>
Gesendet: Mittwoch, 16. Dezember 2020 00:00
An: SimpleMachines/SMF2.1
Cc: albertlast; Author
Betreff: Re: [SimpleMachines/SMF2.1] Add webp to images support (#6271)


None of those links declares webp as the new image standard and this requires us to either bump the min PHP version (something that is not planned at the moment) or add extra checks which adds non needed complexity.

MissAllSunday:
I don't think there is an image standard, webp is just the format google is pushing.

Nevertheless, the required changes overshadows the benefits. Instead I would suggest to change this PR to unify all the $allowedTypes arrays into a single $context key so mod authors can add any format they want.

albertlast:
the benefits are that we supporting what the most broads needs and
we redurce the complexity by removing old stuff.

The last php raise was in 2017 where 5.3 got dropped,
which is 3 years after eol: https://www.php.net/eol.php

based on this, we had to raise php version to 5.6,
when you personal like to had the old version than is your personal problem but not the problem of the project.

MissAllSunday:
Its not my personal choice, never has been.

albertlast:
Since smf 2.0.17 support php 7.2 and with this a migration path exists between 3 php version when i take php 7.0 as baseline,
2.0.18 could possible support php 7.4 which get us to 5 php version overlapping with 2.1
get me to the point that i see no technical reason here.

On the other hand we waist out time with version checks and workaround to support old stuff.

MissAllSunday:
Then help me out with a serious proposal, expose usage data for PHP versions, specially among shared hostings. The more data we have the more we can push towards changing the min PHP version.

And by serious I really mean a serious proposal and not just posting a couple of links and rant about how its our fault. I alone can't make the decision.

albertlast:
no i make not your job.
you hurt yourself by supporting this amount of php version and
ask me to give your reason to stop with this.

The shared hoster are not the problem of this project.

the admin had the option to change the hoster,
to stay with 2.0.x or to get in touch with hoster.

The supported php version is hugh enough even with 7.0,
since the next version of php is comming 8.0 and many will follow in the live time of 2.1.x

m4z:
https://www.php.net/usage.php (apparently from 2013) leads to [netcraft (2013)](https://news.netcraft.com/archives/2013/01/31/php-just-grows-grows.html) and [w3techs (current)](https://w3techs.com/technologies/details/pl-php), and the latter numbers aren't very beautiful: 40% (or 32%, according to the following link :confused: ) of PHP sites still use 5.x!

But: [of the sites using 5.x](https://w3techs.com/technologies/details/pl-php/5),
* 50% use 5.6 (so 20% or 16% of all PHP sites),
* 9% use 5.5 (4% or 3% of all),
* 16% use 5.4 (6% or 5% of all).

[Wordpress shows slightly different numbers](https://wordpress.org/about/stats/):
* 12.1% total for 5.6,
* 1.3% total for 5.5,
* 2.5% total for 5.4.

AFAIK, the SMF detailed stats are broken, so we can't know *our* real numbers.

The detailed market reports of various sites aren't available for free.

albertlast:
using doesn't mean they had no others to choose.

and still the question if smf had the dev power to support sutch amount of php versions.

albertlast:
@MissAllSunday requesting reopening of this pr since https://github.com/SimpleMachines/SMF2.1/pull/6501 got merged

MissAllSunday:
> Instead I would suggest to change this PR to unify all the $allowedTypes arrays into a single $context key so mod authors can add any format they want.

And the min was raised to 5.6.0

https://xkcd.com/927/

albertlast:
5.6.40 is the last version and 5.6.26 would be a good meet in the middle.

sbulen:
Note that IMAGETYPE_WEBP is only defined in php 7.1+.

albertlast:
@sbulen  since thed idea to implement webp again, this pr exists closed for this topic

jdarwood007:
@albertlast Since @sbulen has committed to supporting it.  Would you want to get this up to sync with current code and a viable PR ready?

albertlast:
@jdarwood007 pr is updated now, feel free to test the code

DiegoAndresCortes:
What about certain file inputs like this one:
https://github.com/SimpleMachines/SMF/blob/5d05c650443c3bcd208ecd9c8629f29bdbe78517/Themes/default/Profile.template.php#L2992-L2996

Should they be updated to **accept="image/*"** ?

Actually it seems to be the only one.

PR #6272, opened 2020-09-26
Drop Seeking Methode for display fixing #5764

### LINKED ISSUE #5764: Topic listing issues

albertlast:
Well the idea was nice,
but somestuff doesn't work out well.
First a topic can get changed while "looking" at this,
to make forcast, what would be the start of the next page or otherwise (depen in which direction you go) and
the problem that we support desc view where all the stuff is changing in the moment someone add a post.
Create many situation where this not working fine.

So when you got a asc view with no changing post (no deletion) stuff could work fine,
because of the otherhand the seek works very well on performance side -> constant call time undependet how many pages the thread/topic got.

fixes: #5764

### DIFF
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -860,171 +860,54 @@ function Display()
 		call_integration_hook('integrate_poll_buttons');
 	}
 
+	$limit = $context['messages_per_page'];
 	$start = $_REQUEST['start'];
 	$ascending = empty($options['view_newest_first']);
+	$firstIndex = 0;
 
-	// Check if we can use the seek method to speed things up
-	if (isset($_SESSION['page_topic']) && $_SESSION['page_topic'] == $topic && $_SESSION['page_ascending'] == $ascending)
+	// Jump to page
+	// Calculate the fastest way to get the messages!
+	if ($start >= $context['total_visible_posts'] / 2 && $context['messages_per_page'] != -1)
 	{
-		// User moved to the next page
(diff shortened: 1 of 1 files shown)


BrickOzp (approved):
Tested and it solves the problems with reverse order in topics

Sesquipedalian:
I'm sad to see this go, but I don't see any better way forward either.

live627:
I don't think this restores all the queries lost, though...

PR #6274, opened 2020-09-26
Img loading lazy

albertlast:
based on the yt https://www.youtube.com/watch?v=AActXSWxsRo
i came up with the idea that would be nice.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#Browser_compatibility

But to protect us from scrollbar issue,
loading lazy is only an option where the width and height is known by smf.

Would like to extend this to other type like thumps and other attachment which are existing in attachments table,
but i don't know how atm.

So only avatar are loading lazy.

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -1741,7 +1741,7 @@ function loadMemberContext($user, $display_custom_fields = false)
 		if (!empty($image))
 			$memberContext[$user]['avatar'] = array(
 				'name' => $profile['avatar'],
-				'image' => '<img class="avatar" src="' . $image . '" alt="" loading="lazy" width="' . $profile['attachment_width'] . 'px" height = "'. $profile['attachment_height'] . 'px">',
+				'image' => '<img class="avatar" src="' . $image . '" alt="" loading="lazy" width="' . $profile['attachment_width'] . '" height = "'. $profile['attachment_height'] . '">',
 				'href' => $image,
 				'url' => $image,
 			);
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -504,7 +504,7 @@ function loadUserSettings()
 		if (empty($cache_enable) || $cache_enable < 2 || ($user_settings = cache_get_data('user_settings-' . $id_member, 60)) == null)
 		{
 			$request = $smcFunc['db_query']('', '
-				SELECT mem.*, COALESCE(a.id_attach, 0) AS id_attach, a.filename, a.attachment_type
+				SELECT mem.*, COALESCE(a.id_attach, 0) AS id_attach, a.filename, a.attachment_type, a.width "attachment_width", a.height "attachment_height" 
 				FROM {db_prefix}members AS mem
 					LEFT JOIN {db_prefix}attachments AS a ON (a.id_member = {int:id_member})
 				WHERE mem.id_member = {int:id_member}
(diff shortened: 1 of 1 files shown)


jdarwood007:
Like the idea to use it on avatars, but it seems it would benefit more from external avatars rather than uploaded avatars.
Also a nitpick, don't really see the use to alias column names.

albertlast:
the alias is choosen because otherwise the name width and height would be very generic,
but i had no issue to take it back.

Like i said before i would like to add on many places but this need to know the dimension of the image and
for external i didn't know from where to take it.

dragomano:
It should work with [iframe](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#lazy-loading-attributes) too.

albertlast:
But we don't use them?

dragomano:
But what about the youtube tag?

Subs.php:
`array(
	'tag' => 'youtube',
	'type' => 'unparsed_content',
	'content' => '<div class="videocontainer"><div><iframe frameborder="0" src="https://www.youtube.com/embed/$1?origin=' . $hosturl . '&wmode=opaque" data-youtube-id="$1" allowfullscreen></iframe></div></div>',
	'disabled_content' => '<a href="https://www.youtube.com/watch?v=$1" target="_blank" rel="noopener">https://www.youtube.com/watch?v=$1</a>',
	'block_level' => true,
),`

dragomano:
One needs to remove spaces from this line: `height = "'. $profile['attachment_height'] . 'px"`.
 In addition, $profile['attachment_width'] and $profile['attachment_height'] might be empty, and then html will be look at `width="px" height="px"`. So, I suggest this code:
```
if (!empty($image))
	$memberContext[$user]['avatar'] = array(
		'name' => $profile['avatar'],
		'image' => '<img class="avatar" src="' . $image . '" alt="" loading="lazy"' . (!empty($profile['attachment_width']) ? ' width="' . $profile['attachment_width'] . 'px"' : '') . (!empty($profile['attachment_height']) ? ' height="'. $profile['attachment_height'] . 'px"' : '') . '>',
		'href' => $image,
		'url' => $image,
	);
```

albertlast:
line 803 and 804 should backup the case of empty,
do this not work?
and i see ono space wich needs to be removed, more liky to add one in px.

Sources/Load.php
```diff
@@ -504,7 +504,7 @@ function loadUserSettings()
 		if (empty($cache_enable) || $cache_enable < 2 || ($user_settings = cache_get_data('user_settings-' . $id_member, 60)) == null)
 		{
 			$request = $smcFunc['db_query']('', '
-				SELECT mem.*, COALESCE(a.id_attach, 0) AS id_attach, a.filename, a.attachment_type
+				SELECT mem.*, COALESCE(a.id_attach, 0) AS id_attach, a.filename, a.attachment_type, a.width "attachment_width", a.height "attachment_height"
```
MissAllSunday: Please us "AS" to follow the SQL structure.

MissAllSunday (changes requested):

Sources/Load.php
albertlast: as shoud be like outer only used when needed and like outer it never needed.
but to make you happy i added this.

MissAllSunday: By using "AS" I mean following the same SQL structure the query is already using:

`
, a.width AS attachment_width, a.height  AS attachment_height
`

Sources/Load.php
```diff
@@ -1739,7 +1741,7 @@ function loadMemberContext($user, $display_custom_fields = false)
 		if (!empty($image))
 			$memberContext[$user]['avatar'] = array(
 				'name' => $profile['avatar'],
-				'image' => '<img class="avatar" src="' . $image . '" alt="">',
+				'image' => '<img class="avatar" src="' . $image . '" alt="" loading="lazy" width="' . $profile['attachment_width'] . 'px" height = "'. $profile['attachment_height'] . 'px">',
```
dragomano: > The width and height attributes on img, iframe, embed, object, video, and, when their type attribute is in the Image Button state, input elements may be specified to give the dimensions of the visual content of the element (the width and height respectively, relative to the nominal direction of the output medium), in CSS pixels. The attributes, if specified, must have values that are valid **non-negative integers**.
https://html.spec.whatwg.org/multipage/embedded-content-other.html#attr-dim-width

So, you should use `width="' . $profile['attachment_width'] . '" height = "'. $profile['attachment_height'] . '"`, only numbers, without `px`.

MissAllSunday (approved):

PR #6277, opened 2020-09-26
Prevent a error when loading memcache information in Support info

jdarwood007:
This may be the fix for #6264

### DIFF
--- a/Sources/CacheAPI-memcache.php
+++ b/Sources/CacheAPI-memcache.php
@@ -148,7 +148,7 @@ class memcache_cache extends cache_api
 	 */
 	public function getVersion()
 	{
-		if (is_resource($this->memcache))
+		if (is_object($this->memcache))
 			return false;
 
 		// This gets called in Subs-Admin getServerVersions when loading up support information.  If we can't get a connection, return nothing.
--- a/Sources/CacheAPI-memcache.php
+++ b/Sources/CacheAPI-memcache.php
@@ -148,7 +148,14 @@ class memcache_cache extends cache_api
 	 */
 	public function getVersion()
 	{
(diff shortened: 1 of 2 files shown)


Sources/CacheAPI-memcache.php
```diff
@@ -148,7 +148,14 @@ public function cacheSettings(array &$config_vars)
...
 	public function getVersion()
 	{
-		return $this->memcache->getVersion();
+		if (is_object($this->memcache))
+			return false;
+
+		// This gets called in Subs-Admin getServerVersions when loading up support information.  If we can't get a connection, return nothing.
+		$result = $this->memcache->getVersion();
+		if (!empty($result))
+			return $result;
```
MissAllSunday, line 158: nitpick: Can you add a space between return statements?

Sources/CacheAPI-memcached.php
```diff
@@ -175,7 +175,14 @@ public function cacheSettings(array &$config_vars)
...
 	public function getVersion()
 	{
-		return current($this->memcached->getVersion());
+		if (is_object($this->memcached))
+			return false;
+
+		// This gets called in Subs-Admin getServerVersions when loading up support information.  If we can't get a connection, return nothing.
+		$result = $this->memcached->getVersion();
+		if (!empty($result))
+			return current($result);
```
MissAllSunday, line 185: same

Sources/CacheAPI-memcache.php
MissAllSunday: It looks like it was added between $result and the if statement 😅

MissAllSunday (approved):

PR #6284, opened 2020-10-02
Fixes pagination issue in exports; removes redundant tempfile creation

### LINKED ISSUE #6281: Pagination on profile exports
#### Description
Something is wrong with the pagination on the profile exports.  Like the page is off by one...
When viewing page 9, the link highlights page 8, etc...  Page 2 & 1 both highlight link 1.
In this example, I am on the last page, page 9, but 8 is highlighted. 
I cannot seem to navigate thru all 9 pages.

21rc3-export-pagination

### Steps to reproduce
1. Pick a user with a lot of posts &/or PMs.
2. Export the profile - "styled".  Download.  Browse the downloaded files.

- Database Type:  mysql
- Database Version: 5.7
- PHP Version: 7.4

Sesquipedalian:
This file_put_contents() call was redundant and caused page numbers for all page numbers after the first to be off by one.
Fixes #6281

Also implements a more reliable method to download all the export files when the user clicks the "Download All" button. The old way used the jQuery `.click()` method to simulate user interaction, which is always an inherently fragile approach and much subject to browser restrictions and quirks.

### DIFF
--- a/Sources/tasks/ExportProfileData.php
+++ b/Sources/tasks/ExportProfileData.php
@@ -373,9 +373,6 @@ class ExportProfileData_Background extends SMF_BackgroundTask
 						{
 							rename($tempfile, $realfile);
 							$realfile = $export_dir_slash . ++$filenum . '_' . $idhash_ext;
-
-							file_put_contents($tempfile, implode('', array($context['feed']['header'], $profile_basic_items, $context['feed']['footer'])), LOCK_EX);
-
 							$prev_item_count = $new_item_count = 0;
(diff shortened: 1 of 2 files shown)


sbulen (approved):
Tested & looks good!

sbulen:
Thanks, @Sesquipedalian !!!

sbulen:
And.. yes...  If you recall, testing this, I had issues, mainly in Chrome, downloading multi-page exports.  This solved that old mystery as well.

PR #6285, opened 2020-10-02
Greatly improves performance of $smcFunc['strpos']()

Sesquipedalian:
I thought I'd submitted this a while ago, but apparently not...

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -182,6 +182,8 @@ function reloadSettings()
 
 			$neg_offset = $offset < 0;
 
+			$haystack = $ent_check($haystack);
+
 			$needle_pos = false;
 			while (strlen($haystack) > 0)
 			{
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -175,28 +175,37 @@ function reloadSettings()
 		},
 		'strpos' => function($haystack, $needle, $offset = 0) use ($utf8, $ent_check, $ent_list, $modSettings)
 		{
-			$haystack_arr = preg_split('~(' . $ent_list . '|.)~' . ($utf8 ? 'u' : ''), $ent_check($haystack), -1, PREG_SPLIT_DELIM_CAPTURE | PREG_SPLIT_NO_EMPTY);
-
-			if (strlen($needle) === 1)
-			{
-				$result = array_search($needle, array_slice($haystack_arr, $offset));
-				return is_int($result) ? $result + $offset : false;
-			}
-			else
-			{
-				$needle_arr = preg_split('~(' . $ent_list . '|.)~' . ($utf8 ? 'u' : '') . '', $ent_check($needle), -1, PREG_SPLIT_DELIM_CAPTURE | PREG_SPLIT_NO_EMPTY);
-				$needle_size = count($needle_arr);
-
-				$result = array_search($needle_arr[0], array_slice($haystack_arr, $offset));
-				while ((int) $result === $result)
-				{
-					$offset += $result;
-					if (array_slice($haystack_arr, $offset, $needle_size) === $needle_arr)
-						return $offset;
-					$result = array_search($needle_arr[0], array_slice($haystack_arr, ++$offset));
-				}
+			if (strlen($haystack) < abs($offset))
 				return false;
+
+			$neg_offset = $offset < 0;
+
+			$needle_pos = false;
+			while (strlen($haystack) > 0)
+			{
+				$real_pos = $neg_offset ? strrpos($haystack, $needle) : strpos($haystack, $needle, $offset);
+
+				if ($real_pos === false)
+					return false;
+
+				$spn = substr($haystack, 0, $real_pos);
+				$spn = preg_replace('~' . $ent_list . '~' . ($utf8 ? 'u' : ''), ' ', $spn);
+
+				if ($neg_offset && strlen($haystack) - $real_pos < abs($offset))
+				{
+					$needle_pos = strlen($spn);
+					break;
+				}
+				elseif (!$neg_offset && $real_pos >= $offset)
+				{
+					$needle_pos = strlen($spn);
+					break;
+				}
+				else
+					$haystack = $neg_offset ? substr($haystack, $real_pos + strlen($needle)) : substr($haystack, 0, $real_pos + strlen($needle));
 			}
+
+			return $needle_pos;
 		},
 		'substr' => function($string, $start, $length = null) use ($utf8, $ent_check, $ent_list, $modSettings)
 		{
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -169,10 +169,12 @@ function reloadSettings()
 
 			return preg_replace('~^(?:[' . $space_chars . ']|&nbsp;)+|(?:[' . $space_chars . ']|&nbsp;)+$~' . ($utf8 ? 'u' : ''), '', $ent_check($string));
 		},
+		// Like normal strlen, except that it counts entities as single characters.
 		'strlen' => function($string) use ($ent_list, $utf8, $ent_check)
 		{
 			return strlen(preg_replace('~' . $ent_list . ($utf8 ? '|.~u' : '~'), '_', $ent_check($string)));
 		},
+		// Like normal strpos, except that it counts entities as single characters.
 		'strpos' => function($haystack, $needle, $offset = 0) use ($utf8, $ent_check, $ent_list, $modSettings)
 		{
 			if (strlen($haystack) < abs($offset))
@@ -207,6 +209,7 @@ function reloadSettings()
 
 			return $needle_pos;
 		},
+		// Like normal substr, except that it handles entities as single characters.
 		'substr' => function($string, $start, $length = null) use ($utf8, $ent_check, $ent_list, $modSettings)
 		{
 			$ent_arr = preg_split('~(' . $ent_list . '|.)~' . ($utf8 ? 'u' : '') . '', $ent_check($string), -1, PREG_SPLIT_DELIM_CAPTURE | PREG_SPLIT_NO_EMPTY);


sbulen:
IIRC, you reverted this as it caused problems with bbc in the exports?

albertlast:
What is the difference to php strpos?
When should it be used?

Sesquipedalian:
> IIRC, you reverted this as it caused problems with bbc in the exports?  

No, that's not quite right. There was a problem with calling this function during exports (or more precisely, during the creation of any XML via News.php), but the problem turned out to be simply that calling `$smcFunc['strpos']()` *at all* in that particular situation was the wrong thing to do (the situation actually called for PHP's standard `strpos()`). But in the process of figuring out what the problem really was, I did spend a while trying to improve `$smcFunc['strpos']()` in the hope that it would solve the bug we were trying to deal with at the time, and I did get you to test the improvements, which is probably why this looks familiar to you. 🙂 But once I realized the true cause of the bug, and therefore that improving improving the performance of `$smcFunc['strpos']()` was irrelevant to profile exports, I split these changes off into their own branch with the intention of submitting them as a separate pull request. And then I forgot about them. 😆

> What is the difference to php strpos?
> 
> When should it be used?

Basically, `$smcFunc['strpos']()` is similar to normal `strpos()` except that it counts HTML entities as single characters. For example, `$smcFunc['strpos']('A&amp;B', 'B')` will return 2 rather than 6.

It's useful if you want to count characters as they will appear visually to a human reader, rather than counting characters as they are stored and interpreted by a machine.

live627:
Thanks for that explanation. The wiki has nothing. Literally a blank page.

Sesquipedalian:
> Thanks for that explanation. The wiki has nothing. Literally a blank page.

I've just added some explanatory comments in Load.php itself for `$smcFunc['strlen']`, `$smcFunc['strpos']`, and `$smcFunc['substr']`. Someone else will need to take care of the wiki, though. 🙂

sbulen:
Do you have any documentation on what the problem being solved here is?  Any before/after benchmarks?  Anything to show that we haven't introduced bugs - that functionality has been matched?  

No regressions?  How exactly did you test this?  It bothers me that nothing is documented.

As we are inching closer to release, I strongly wish to avoid unnecessary changes.

Sesquipedalian:
I did quite a bit of testing and benchmarking at the time, and I'm quite confident in this code, but I suppose I could reconstruct those tests again.

sbulen:
I'm more concerned about any adverse impact to functionality...  Have you tested all places where it is used to confirm they still work properly?  From a user perspective, _actually via using the app?_  

Is the risk worth it?  How much of an improvement is there?

Please share documentation of your tests.

Before merging anything, please check in with Suki.

Sesquipedalian:
I've tested all the places `$smcFunc['strpos']()` is used, and yes, this works perfectly.

albertlast:
and how fast are you?

Sesquipedalian:
> and how fast are you?

Test code:
```
<?php

$utf8 = true;
$modSettings['disableEntityCheck'] = false;

$ent_list = '&(?:#' . (empty($modSettings['disableEntityCheck']) ? '\d{1,7}' : '021') . '|quot|amp|lt|gt|nbsp);';
$ent_check = empty($modSettings['disableEntityCheck']) ? function($string)
{
	$string = preg_replace_callback('~(&#(\d{1,7}|x[0-9a-fA-F]{1,6});)~', 'entity_fix__callback', $string);
	return $string;
} : function($string)
{
	return $string;
};

$smcFunc_strpos_old = function($haystack, $needle, $offset = 0) use ($utf8, $ent_check, $ent_list, $modSettings)
{
	$haystack_arr = preg_split('~(' . $ent_list . '|.)~' . ($utf8 ? 'u' : ''), $ent_check($haystack), -1, PREG_SPLIT_DELIM_CAPTURE | PREG_SPLIT_NO_EMPTY);

	if (strlen($needle) === 1)
	{
		$result = array_search($needle, array_slice($haystack_arr, $offset));
		return is_int($result) ? $result + $offset : false;
	}
	else
	{
		$needle_arr = preg_split('~(' . $ent_list . '|.)~' . ($utf8 ? 'u' : '') . '', $ent_check($needle), -1, PREG_SPLIT_DELIM_CAPTURE | PREG_SPLIT_NO_EMPTY);
		$needle_size = count($needle_arr);

		$result = array_search($needle_arr[0], array_slice($haystack_arr, $offset));
		while ((int) $result === $result)
		{
			$offset += $result;
			if (array_slice($haystack_arr, $offset, $needle_size) === $needle_arr)
				return $offset;
			$result = array_search($needle_arr[0], array_slice($haystack_arr, ++$offset));
		}
		return false;
	}
};

$smcFunc_strpos_new = function($haystack, $needle, $offset = 0) use ($utf8, $ent_check, $ent_list, $modSettings)
{
	if (strlen($haystack) < abs($offset))
		return false;

	$neg_offset = $offset < 0;

	if (strpos($haystack, '&#x') !== false)
		$haystack = $ent_check($haystack);

	$needle_pos = false;
	while (strlen($haystack) > 0)
	{
		$real_pos = $neg_offset ? strrpos($haystack, $needle) : strpos($haystack, $needle, $offset);

		if ($real_pos === false)
			return false;

		$spn = substr($haystack, 0, $real_pos);
		$spn = preg_replace('~' . $ent_list . '~' . ($utf8 ? 'u' : ''), ' ', $spn);

		if ($neg_offset && strlen($haystack) - $real_pos < abs($offset))
		{
			$needle_pos = strlen($spn);
			break;
		}
		elseif (!$neg_offset && $real_pos >= $offset)
		{
			$needle_pos = strlen($spn);
			break;
		}
		else
			$haystack = $neg_offset ? substr($haystack, $real_pos + strlen($needle)) : substr($haystack, 0, $real_pos + strlen($needle));
	}

	return $needle_pos;
};

function entity_fix__callback($matches)
{
	if (!isset($matches[2]))
		return '';

	$num = $matches[2][0] === 'x' ? hexdec(substr($matches[2], 1)) : (int) $matches[2];

	// we don't allow control characters, characters out of range, byte markers, etc
	if ($num < 0x20 || $num > 0x10FFFF || ($num >= 0xD800 && $num <= 0xDFFF) || $num == 0x202D || $num == 0x202E)
		return '';
	else
		return '&#' . $num . ';';
}


// --------------

// Generate a string of random entities, with needle at the very end for testing purposes.
$needle = '&amp;';
$haystack = '';
for ($i=0; $i < 255; $i++) {
	$haystack .= '&#' . str_pad(mt_rand(0, intval(hexdec('D7FF'))), 4, '0', STR_PAD_LEFT) . ';';
}
$haystack .= $needle;

$iterations = 10000;

// Run the tests
$start = microtime(true);
for ($i=0; $i < $iterations; $i++) {
	$smcFunc_strpos_old($haystack, $needle, 0);
}
echo 'Old: ' . (microtime(true) - $start) . "\n" ;

$start = microtime(true);
for ($i=0; $i < $iterations; $i++) {
	$smcFunc_strpos_new($haystack, $needle, 0);
}
echo 'New: ' . (microtime(true) - $start) . "\n" ;
```

Result:
```
Old: 2.1900990009308
New: 0.56483888626099
```

Sesquipedalian:
You know what? Never mind. This function is used so rarely that it's probably not worthwhile.

live627:
I come with some rather unexpected results. I just wrote a test that where the old implementation fails butt this one passes.
```php
$needle = '&amp;';
$haystack = '';
for ($i=0; $i < 255; $i++) {
	$haystack .= '&#' . str_pad(mt_rand(0, intval(hexdec('D7FF'))), 4, '0', STR_PAD_LEFT) . ';';
}
$haystack .= $needle . "\n\t\t\t€😄A𐍈¢РУССКИЕ";

// 255
echo $smcFunc_strpos_new($haystack, $needle); // PASS
echo "\n";
echo $smcFunc_strpos_old($haystack, $needle); // PASS
echo "\n";

// Expect 264
echo $smcFunc_strpos_new($haystack, '¢Р'); // FAIL 272
echo "\n";
echo $smcFunc_strpos_old($haystack, '¢Р'); // PASS
echo "\n";
```
About performance: I don;'t get such good numbers
```php
	$repetitions = 0;
	$start = microtime(true);
	do
	{
		$smcFunc_strpos_old($haystack, $needle);
		$repetitions += 1;
	}
	while (microtime(true) - $start < 5);

	$rate = round($repetitions / (microtime(true) - $start));
	$rate = str_pad($rate, 8, ' ', STR_PAD_LEFT);

	echo number_format($rate) . " / second\n";

	$repetitions = 0;
	$start = microtime(true);
	do
	{
		$smcFunc_strpos_new($haystack, $needle);
		$repetitions += 1;
	}
	while (microtime(true) - $start < 5);

	$rate = round($repetitions / (microtime(true) - $start));
	$rate = str_pad($rate, 8, ' ', STR_PAD_LEFT);

	echo number_format($rate) . " / second\n";
```
entity check enabled
- old 11,776 / second
- new 12,715 / second

entity check disabled
- old 73,716 / second
- new 142,840 / second

This new implantation uses much less memory whereas the old one would run out on my first large test below. (did not measure it, only time)
```php
// Worst case
test(str_repeat('A', 1e+7) . 'BB', 'BB');
test(str_repeat('aA', 1e+6) . 'aB', 'aB');
test(str_repeat('A', 1e+1) . 'BB', 'BB');
test(str_repeat('aA', 1e+1) . 'aB', 'aB');
// Best case
test(str_repeat('a', 1e+7), 'aa');
test(str_repeat('aa', 1e+6), 'aa');
test(str_repeat('a', 1e+1), 'aa');
test(str_repeat('aA', 1e+1), 'aa');
```
- 433 / second
- 286 / second
- 2,820,073 / second
- 2,669,937 / second
- 3,793 / second
- 19,256 / second
- 3,001,035 / second
- 4,814,683 / second

Is still faster, though.
```php 
// Worst case
//~ test(str_repeat('A', 1e+7) . 'BB', 'BB');
//~ test(str_repeat('aA', 1e+6) . 'aB', 'aB');
test(str_repeat('A', 1e+1) . 'BB', 'BB');
test(str_repeat('aA', 1e+1) . 'aB', 'aB');
// Best case
//~ test(str_repeat('a', 1e+7), 'aa');
//~ test(str_repeat('aa', 1e+6), 'aa');
test(str_repeat('a', 1e+1), 'aa');
test(str_repeat('aA', 1e+1), 'aa');
```
- old 31,416 / second
- new 2,827,557 / second
- old 323,575 / second
- new 2,675,053 / second
- old 886,891 / second
- new 2,997,136 / second
- old 337,376 / second
- new 4,855,230 / second

My test file [strpos.php.gz](https://github.com/SimpleMachines/SMF/files/12012907/strpos.php.gz)
```
php -v
PHP 8.1.1 (cli) (built: Dec 15 2021 10:31:43) (ZTS Visual C++ 2019 x64)
````
11th Gen Intel(R) Core(TM) i7-11700F @ 2.50GHz

**Anyway, can this be reopened for consideration?**

Sesquipedalian:
It might be a couple of days before I can do so, but sure.

PR #6286, opened 2020-10-02
Neutralizes evil control characters in preparsecode()

Sesquipedalian:
Another one that I thought I'd submitted and apparently never did...

This converts control characters (except \t, \r, and \n) to harmless Unicode symbols in preparsecode().

### DIFF
--- a/Sources/Subs-Post.php
+++ b/Sources/Subs-Post.php
@@ -32,6 +32,19 @@ function preparsecode(&$message, $previewing = false)
 
 	static $tags_regex, $disallowed_tags_regex;
 
+	// Convert control characters (except \t, \r, and \n) to harmless Unicode symbols
+	$control_replacements = array(
+		"\x00" => '&#x2400;', "\x01" => '&#x2401;', "\x02" => '&#x2402;', "\x03" => '&#x2403;',
+		"\x04" => '&#x2404;', "\x05" => '&#x2405;', "\x06" => '&#x2406;', "\x07" => '&#x2407;',
+		"\x08" => '&#x2408;', "\x0b" => '&#x240b;', "\x0c" => '&#x240c;', "\x0e" => '&#x240e;',
+		"\x0f" => '&#x240f;', "\x10" => '&#x2410;', "\x11" => '&#x2411;', "\x12" => '&#x2412;',
+		"\x13" => '&#x2413;', "\x14" => '&#x2414;', "\x15" => '&#x2415;', "\x16" => '&#x2416;',
+		"\x17" => '&#x2417;', "\x18" => '&#x2418;', "\x19" => '&#x2419;', "\x1a" => '&#x241a;',
+		"\x1b" => '&#x241b;', "\x1c" => '&#x241c;', "\x1d" => '&#x241d;', "\x1e" => '&#x241e;',
+		"\x1f" => '&#x241f;',
+	);
+	$message = strtr($message, $control_replacements);
+
 	// This line makes all languages *theoretically* work even with the wrong charset ;).
 	if (empty($context['utf8']))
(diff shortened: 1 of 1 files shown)


sbulen:
What problem is this solving?  Has there been a bug reported?

What testing has been performed?  Can you provide any documentation of the before/after?

I am very skeptical about introducing any changes that may destabilize the code.  I think we need much better documentation about the whats & whys & proof this has no adverse impacts.

Sesquipedalian:
The control characters this replaces should never ever be allowed in posts, because they can break things further down the line. But surprisingly, `preparsecode()` doesn't filter them out. Instead, we have thus far relied only on filters in various locations to remove them from output (e.g. in `cleanXml()`), and it is difficult to guarantee that we are applying such filters everywhere we should. It is best practice to remove them at the input stage, so that's what this PR does.

I chose to replace the illegal control characters with these symbols (which exist in Unicode for the express purpose of representing them) rather than simply removing them because I think it is better to be obvious about changes to user input. I could be persuaded to simply remove them if others think that would be better, though.

Sesquipedalian:
As for testing, I have tested and this introduces no appreciable performance hit, nor does it introduce any issues that I can find at all. It's a simple string substitution that changes the incoming string to make it exactly as would be if the user had typed these symbols in the first place. From there, the rest of `preparsecode()` handles the string in exactly the same way that it handles every other string.

sbulen:
Good idea, I understand what you're trying to do.  I doubt this even registers as a performance hit, that is not a particular concern.  I'm very concerned with that the user will see.  If folks are copying/pasting from whatever sources, and today see nothing, will they see these new representations in the future?  These particular substitutions may make it harder for folks to publish articles.  That's the type of thing I'm worried about.  

Please share the results of tests.  Evidence is important for documentation so we know what & why.  What will the user see that is different.

Sesquipedalian:
Note: In the example text below, GitHub wisely replaces the control character with a substitute because they know that it is an instrument of evil. On SMF, we current just accept such control characters. The original character is invisible, and so it doesn't appear visibly to the human reader on SMF.

Raw input.
```
There is a control character (U+0011) between "It" and "s" in the following line:
Its time to go home.
```

Output without this PR. Note how the control character is retained.
```
There is a control character (U+0011) between "It" and "s" in the following line:
Its time to go home.
```

Output with this PR. Note how it has been replaced with its corresponding Unicode symbol character.
```
There is a control character (U+0011) between "It" and "s" in the following line:
It␑s time to go home.
```

By showing the substitute character, we simultaneously avoid accepting the problematic characters in user input, while also making it clear to the user where the problem is. The only halfway acceptable alternative would be to silently strip out control characters and replace them with nothing. The latter approach might *seem* more user friendly at first glance, but it runs the risk of messing with user input in ways the user didn't expect, while giving no feedback about the changes. It's actually much better to show substitute characters (as we can see GitHub doing here right now, for example).

Sesquipedalian:
I suppose I could use the generic Unicode replacement character "�" (U+FFFD) rather than the specific replacement symbols, if only because the specific ones can be hard to read in some fonts. Thoughts, @sbulen?

sbulen:
I think the currently proposed characters are preferable to the dreaded black diamond...

I'm trying to place myself in the shoes of a user doing the copy/paste (or whatever) activity.   Would I rather see the chars, or nothing?   

Some use tools to try to strip them out when they copy content over.  I learned this in a couple recent support threads.   These folks might WANT to know exactly what was posted, see the alt-chars, and correct their workflow & tool settings.  Some (more, I suspect) might just want a clean post, with the chars either stripped or replaced with blanks.

I think either approach will work...   We'll just have to wait & see if someone complains...

PR #6289, opened 2020-10-04
Fix session write fail issue

### LINKED ISSUE #6239: Session bug
#### Description
Error
Type of error: General
Error message:
2: Unknown: Failed to write session data using user defined save handler. (session.save_path: /var/lib/php/session)
Unknown
URL of page causing the error: https://website.com/index.php?action=dlattach;attach=18981;type=avatar
Backtrace information
#0: smf_error_handler()

### Steps to reproduce
Don't know how to reproduce.

- Version/Git revision: 2.1 Latest
- Database Type: Mariadb
- Database Version: Mariadb latest.
- PHP Version: 7.4

If anyone knows what caused this error to occur lemme know.  It has only happened one time in 5 days on my production website.

sbulen:
Fixes #6239

### DIFF
--- a/Sources/Session.php
+++ b/Sources/Session.php
@@ -179,26 +179,13 @@ function sessionWrite($session_id, $data)
 		$db_connection = smf_db_initiate($db_server, $db_name, $db_user, $db_passwd, $db_prefix, $options);
 	}
 
-	// First try to update an existing row...
-	$smcFunc['db_query']('', '
-		UPDATE {db_prefix}sessions
-		SET data = {string:data}, last_update = {int:last_update}
-		WHERE session_id = {string:session_id}',
-		array(
-			'last_update' => time(),
-			'data' => $data,
-			'session_id' => $session_id,
-		)
+	// If an insert fails due to a dupe, replace the existing session...
+	$smcFunc['db_insert']('replace',
+		'{db_prefix}sessions',
+		array('session_id' => 'string', 'data' => 'string', 'last_update' => 'int'),
+		array($session_id, $data, time()),
+		array('session_id')
 	);
-
-	// If that didn't work, try inserting a new one.
-	if ($smcFunc['db_affected_rows']() == 0)
-		$smcFunc['db_insert']('ignore',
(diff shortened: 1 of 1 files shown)


Sesquipedalian (approved):

jdarwood007:
I'm trying to find out if there is a performance difference between replace versus update.  The only information I can find is replace will issue a delete first if a matching row on the primary key is found, so it is issuing two statements.  This would be my only concern here.  I was mostly thinking for bigger forums where a delete then insert is a table locking event, while update should be a row locking event (for InnoDB, I don't know if Memory/heap tables have the same issue).

btw, If we had pushed the min mysql version up to 5.6, we could have supported the insert on duplicate update function.  Far as I can tell, it was added in 5.6.  I can't actually find that, but thats as far back as mysql developer documentation goes.

I would otherwise say this should be approved.

jdarwood007:
Found the memory/heap reference: https://dev.mysql.com/doc/refman/5.7/en/internal-locking.html
Its a table locking.  So InnoDB would be the only one benefiting here from an update.  Unless the footnote about table locks on specific updates applies here.  Which I would only see possible figuring out with some logging.

sbulen:
No, this is not a lock issue, or that error would have been reported.  I have confirmed no problems with error logging here (that was my first thought).  

Note this PR actually reduces the # of calls to smcFunc...  The problem occurs when it is doing the second smcFunc calls.  What smcFunc will do for mysql with 'replace' is issue a single REPLACE.
https://github.com/SimpleMachines/SMF2.1/blob/523819a678ba5bcb99c81cd85cc290c6cd6aa0df/Sources/Subs-Db-mysql.php#L778

Pretty sure this can be reproduced in MyISAM or InnoDB.  The cause is explained in the Issue - two session writes too close together for the same user/session.  This can happen a LOT in 2.1, since EVERYTHING in 2.1 causes a new session, e.g., even the frequent pings to get updates to notifications, etc.

Note the fix has been confirmed to eliminate the issue in a busy production environment.

PR #6291, opened 2020-10-04
Need to keep default; it may have custom smileys in it

### LINKED ISSUE #6290: Upgrade defaults to wrong Smiley set
After upgrading from 2.0.17 to 2.1RC3 I noticed I had image placeholders in the editor where my custom smileys were supposed to be.   Turns out 2.1 doesn't carry over the default smiley set you're using, and selects its own default set. 

Before upgrade: Original 2.0.17 Smiley default set = *Alienine's Set

After Upgrade: Default after upgrade = Fugue's Set

Luckily my original set was still there so switched back as default and all is well, but it did confuse me somewhat to start with, and I would imagine it will confuse others, resulting in a few support topics.

sbulen:
Fixes #6290 

Previously, we had stopped using the 'default' smiley set folder, because it was too confusing & had been stepped on in previous versions of 2.1.  But as reported, some folks have used it & enhanced it with custom smileys.  So...  We need to keep it.

I did not keep the name 'Alienine', as that would be confusing, as there is a separate set called Alienine's Set...  So, when keeping usage of the /default smiley set folder around, went with a "2.0 Default" label.

Hope this makes sense.  Feedback welcome.

### DIFF
--- a/Themes/default/languages/Install.english.php
+++ b/Themes/default/languages/Install.english.php
@@ -45,6 +45,7 @@ $txt['default_fugue_smileyset_name'] = 'Fugue\'s Set';
 $txt['default_alienine_smileyset_name'] = 'Alienine\'s Set';
 $txt['default_aaron_smileyset_name'] = 'Aaron\'s Set';
 $txt['default_akyhne_smileyset_name'] = 'Akyhne\'s Set';
+$txt['default_legacy_smileyset_name'] = '2.0 Default';
 $txt['default_theme_name'] = 'SMF Default Theme - Curve2';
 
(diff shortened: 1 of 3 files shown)


Sesquipedalian:
It'd be nice if we could check whether the default set contains any custom smilies before choosing whether to keep it or not. This might be as simple as using `glob()` to list the contents of both the `default` and `alienine` folders and then comparing the results to see if `default` contains anything that `alienine` does not. If so, then we keep `default`. If not, we can drop it.

sbulen:
I'd prefer to keep it as-is.  Nice and simple.  No need to make it any more complicated.

live627 (approved):

PR #6294, opened 2020-10-06
Implements a better compatibilty function for hash_equals

Sesquipedalian:
The previous version used an if statement to take conditional actions, which necessarily changes the processing time, thereby defeating the purpose of the hash_equals function. This version avoids that problem. See https://www.php.net/manual/en/function.hash-equals.php#125034 for further explanation.

### DIFF
--- a/Sources/Subs-Auth.php
+++ b/Sources/Subs-Auth.php
@@ -947,8 +947,8 @@ function hash_benchmark($hashTime = 0.2)
 	return $cost;
 }
 
-// Based on code by "s rotondo90 at gmail com".
-// https://www.php.net/manual/en/function.hash-equals.php#119576
+// Based on code by "examplehash at user dot com".
+// https://www.php.net/manual/en/function.hash-equals.php#125034
 if (!function_exists('hash_equals'))
 {
 	/**
@@ -959,22 +959,20 @@ if (!function_exists('hash_equals'))
 	 */
 	function hash_equals($known_string, $user_string)
 	{
-		$ret = 0;
+		$known_string = (string) $known_string;
+		$user_string = (string) $user_string;
 
-		if (strlen($known_string) !== strlen($user_string))
+		$sx = 0;
+		$sy = strlen($known_string);
+		$uy = strlen($user_string);
+		$result = $sy - $uy;
+		for ($ux = 0; $ux < $uy; $ux++)
 		{
-			$user_string = $known_string;
-			$ret = 1;
+			$result |= ord($user_string[$ux]) ^ ord($known_string[$sx]);
+			$sx = ($sx + 1) % $sy;
 		}
 
-		$res = $known_string ^ $user_string;
-
-		for ($i = strlen($res) - 1; $i >= 0; --$i)
-		{
-			$ret |= ord($res[$i]);
-		}
-
-		return !$ret;
+		return !$result;
 	}
 }
 


sbulen:
As with others, what testing has been performed?  I'd like to see some before & afters, with some assurance there is no impact to existing functionality.

We need to get more & more strict about changes we introduce.

jdarwood007:
@sbulen Actually the comment in updated notes states the problem.  We can not do strlen comparison as otherwise it allows a timing comparison attack to occur.  While this such an attack is difficult, it has been proven to be possible.  Timing attacks are more well known nowadays because of the meltdown and spectre attacks.  For security related crypto comparisons, you can not allow any sort of optimization handling without risking a timing attack exposing information.

So this change should merge up for the sake of making a timing attack harder.  It doesn't look like it effectively changes how the function otherwise will compute the result.

sbulen:
Please read ny note above.  I had asked about testing.  I had read the note & understood the change.

jdarwood007:
Aye, wanted you to be clear this is actually more of a security related fix rather than a performance/bug fix/optimization.  So it should fall into different categories for approval because its security related.

I haven't performed any testing yet but:
- You will see this perform slower due to the nature of the fix.
- Security related change here with a very low chance of a successful attack and a very high degree of skill required exploit.
- The code is otherwise equivalent in its handling of the backwards compatibility function.

sbulen:
So what testing do you propose?  And who is going to actually do some real testing?

jdarwood007:
Are you suggesting we generate a bunch of hashes and throw them at these functions to ensure generates the same answer?  If you are searching for an edge case to show the function change produce different results, I don't know if you could find one with any naturally occurring hash. Even a dirty hash I would say finding one of those would be very difficult.

sbulen:
Something.  But not nothing.

Sesquipedalian:
This implements both the old and the new versions of our `hash_equals()` compatibility function as the own named functions. You can run it as many times as you like, @sbulen, and the output will always be "Results were the same."

```
<?php

$different = false;
for ($i=0; $i < 1000; $i++) {
	$string = bin2hex(random_bytes(mt_rand(8,32)));
	$h1 = hash_equals($string, $string);
	$h2 = smf_hash_equals_old($string, $string);
	$h3 = smf_hash_equals_new($string, $string);

	if ($h1 !== $h2 || $h1 !== $h3)
		$different = true;
}

if ($different)
	echo "Results were different!\n";
else
	echo "Results were the same.\n";


function smf_hash_equals_old($known_string, $user_string)
{
	$ret = 0;

	if (strlen($known_string) !== strlen($user_string))
	{
		$user_string = $known_string;
		$ret = 1;
	}

	$res = $known_string ^ $user_string;

	for ($i = strlen($res) - 1; $i >= 0; --$i)
	{
		$ret |= ord($res[$i]);
	}

	return !$ret;
}

function smf_hash_equals_new($known_string, $user_string)
{
	$known_string = (string) $known_string;
	$user_string = (string) $user_string;

	$sx = 0;
	$sy = strlen($known_string);
	$uy = strlen($user_string);
	$result = $sy - $uy;
	for ($ux = 0; $ux < $uy; $ux++)
	{
		$result |= ord($user_string[$ux]) ^ ord($known_string[$sx]);
		$sx = ($sx + 1) % $sy;
	}

	return !$result;
}
```

Sesquipedalian:
Does anyone object to merging this now, @sbulen or @jdarwood007?

sbulen:
That was a good test, thanks.

jdarwood007 (approved):

PR #6300, opened 2020-10-17
Fix issues with running upgrade.php from CLI

jdarwood007:
I corrected these issues so its possible to run upgrade.php via CLI without issues or needing to copy files around.

We can run the upgrade in the following ways:
smf_root# php other/upgrade.php --path=./
smf_root/other# php upgrade.php --path=../
\# php upgrade.php --path=smf_root/


MissAllSunday:
Can you rebase against release2.1 so this can be merged?

MissAllSunday (approved):

PR #6306, opened 2020-10-18
Ensure password_salt is the proper length

sbulen:
Some converted forums may have the wrong length of password_salt.

Sample issue in forum:
https://www.simplemachines.org/community/index.php?topic=575241.msg4071364#msg4071364
Converters have been fixed:
https://github.com/SimpleMachines/converters/pull/5

### DIFF
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -2552,6 +2552,11 @@ ALTER TABLE {$db_prefix}members
 MODIFY COLUMN pm_ignore_list TEXT NULL;
 ---#
 
+---# Updating password_salt
+ALTER TABLE {$db_prefix}members
+MODIFY COLUMN password_salt VARCHAR(255) NOT NULL DEFAULT '';
+---#
+
 ---# Updating member_logins id_member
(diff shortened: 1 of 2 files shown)


albertlast:
well it's a complex situation...
because this error is not happen by smf directly,
from a smf pov this error happen by a "addon" so in general i would recomand to add this in repair php thing.

since it would be smart to had in repair in general smf schema check to find this kind of issue.

jdarwood007:
The installer has it set to a varchar(255).  So the upgrader should set it to that as well.

sbulen:
I would like to confirm that syntax in pg is not destructive.

albertlast:
> 
> 
> The installer has it set to a varchar(255). So the upgrader should set it to that as well.

and what is with the others ~100 columns?

sbulen:
Recent 2.1 changes are causing issues; we now have much longer values for password_salt.  It wasn't a problem before.  And we likely have a lot of users who will be affected - converted users from a variety of sources.  We saw this within hours of putting RC3 out there...

I tested this on pg and mysql and both work fine.  We're good.

Issue #6305, opened 2020-10-18
PHP 8.0 - Status 500 when viewing package list 


tinoest:
#### Description

Status 500 returned when viewing package list in php 8.0-rc1

### Steps to reproduce
1. Install php8.0-rc1
2. go to index.php?action=admin;area=packages
 

### Environment (complete as necessary)
- Version/Git revision: be238c1c45022e7a0e5977904f3caf6caba9d388
- Database Type: pgsql
- Database Version: 11.9
- PHP Version: 8.0-rc1

### Additional information/references

PHP Fatal error:  Uncaught Error: Unknown named parameter $type in Sources/Subs-List.php:132\nStack trace:\n#0 Sources/Packages.php(1499): createList()\n#1 Sources/Subs.php(5342): PackageBrowse()\n#2 Sources/Packages.php(91): call_helper()\n#3 Sources/Admin.php(499): Packages()\n#4 index.php(173): AdminMain()\n#5 {main}\n  thrown in Sources/Subs-List.php on line 132, referer: index.php

jdarwood007:
Packages.php

```
				'params' => array('type' => $type),
```

```
				'params' => array($type),
```

albertlast:
What is plan of supporting php?
Try to suppprt more and more do we cut on some places?

jdarwood007:
No reason to cut any support at this point.

albertlast:
@jdarwood007 should be this issue not closed by the pr #6311?

jdarwood007:
Yes, I snuck the fix in for this in that PR.

PR #6311, opened 2020-10-19
Can't hash a directory

jdarwood007:


albertlast:
You delete the delete url and keep url,.
Without adding something new?

jdarwood007:
Oops, didn't mean to do that.  But I will leave the other change as it will fix up #6305 
There was one other PHP 8 issue I saw that needs fixed so far.

PR #6312, opened 2020-10-21
Installing SMF in php 8.0 failed

jdarwood007:
PHP Deprecated:  uksort(): Returning bool from comparison function is deprecated, return an integer less than, equal to, or greater than zero in /smf/Sources/Subs-Admin.php on line 1231

Hopefully I didn't foobar the logic.  It still looks right and doesn't appear to have broken anything.

### DIFF
--- a/Sources/Subs-Admin.php
+++ b/Sources/Subs-Admin.php
@@ -1221,16 +1221,13 @@ function updateSettingsFile($config_vars, $keep_quotes = null, $rebuild = false)
 	// It's important to do the numbered ones before the named ones, or messes happen.
 	uksort($substitutions, function($a, $b) {
 		if (is_int($a) && is_int($b))
-			return $a > $b ? 1 : 0;
+			return $a > $b ? 1 : ($a < $b ? -1 : 0);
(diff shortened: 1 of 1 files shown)


Sources/Subs-Admin.php
```diff
@@ -1221,13 +1221,16 @@ function updateSettingsFile($config_vars, $keep_quotes = null, $rebuild = false)
...
 		elseif (is_int($a))
 			return -1;
 		elseif (is_int($b))
 			return 1;
 		else
-			return strcasecmp($b, $a);
+		{
+			$res = strcasecmp($b, $a);
+			return $res > 0 ? 1 : ($res < 0 ? -1 : 0);
+		}
```
live627: strcasecmp() returns an int compatible with uksort()... can you reverse this change?

jdarwood007: I was thinking that uksort needed -1, 0 or 1 for return, but the docs says it can be any int.

I changed the line above because $a < $b should programmatically be -1 not 0.  0 should be for when they are equal

PR #6321, opened 2020-10-28
Uses correct time zone when suggesting default times for a new event

### LINKED ISSUE #5842: Default time for calendar events
#### Description
The default time for calendar events is in the server time, not the user time.  This is confusing, especially because it displays the user timezone right next to it...

### Steps to reproduce
1. Set a user timezone
2. Add a calendar event

- Database Type: mysql & pg
- Database Version:
- PHP Version: 7.1

In this example, server is set to utc, & user is set to PDT.  It is 7:32 pdt when this is added.  The time reflected in the create event time is "now", but in utc.  
image

Sesquipedalian:
Fixes #5842

### DIFF
--- a/Sources/Subs-Calendar.php
+++ b/Sources/Subs-Calendar.php
@@ -1441,6 +1441,9 @@ function getNewEventDatetimes()
 {
 	// Ensure setEventStartEnd() has something to work with
 	$now = date_create();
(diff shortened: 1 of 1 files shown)


sbulen (approved):
Looks good & tests out perfect.

PR #6319, opened 2020-10-26
MariaDB group by quirk

sbulen:
Fixes #6309 

Anybody have MariaDB who can test this?

PR #6323, opened 2020-10-31
Define pg sequence owned by

albertlast:
two possible issues get fixed,
first that default script can't find the sequence found by @sbulen see https://github.com/SimpleMachines/SMF2.1/pull/6319#issuecomment-719897874
second issue is, that a sequence survive a table drop.
When now a table get dropped the sequence get deleted too.

in short sequence beahvior more like a ai field -> good for smf.

### DIFF
--- a/Sources/DbPackages-postgresql.php
+++ b/Sources/DbPackages-postgresql.php
@@ -151,12 +151,15 @@ function smf_db_create_table($table_name, $columns, $indexes = array(), $paramet
 	if (!$db_trans)
 		$smcFunc['db_transaction']('begin');
 	$table_query = 'CREATE TABLE ' . $table_name . "\n" . '(';
+	$alter_sequence = '';
 	foreach ($columns as $column)
 	{
 		// If we have an auto increment do it!
 		if (!empty($column['auto']))
 		{
 			if (!$old_table_exists)
+			{
+			
 				$smcFunc['db_query']('', '
 					DROP SEQUENCE IF EXISTS ' . $table_name . '_seq',
 					array(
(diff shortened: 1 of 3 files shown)


jdarwood007:
Don't know much about postgres as I should, but since your querying global tables and modifying stuff, is there a chance this could fail in typical standard shared hosting conditions where an account may have limited access to certain data?

albertlast:
Couldn't image since are basic tabell otherwise the client is not able to work,
also i only select tables which start with the smf prefix.

had in mind, that smf 2.1 is not enabled for running in other schemas than public -> ther can not be other smf setups in the same db.

albertlast:
i limited to the public schema is this in your favor?

sbulen:
I really don't think this should be done by navigating the catalog...

Why didnt you just add ALTER SEQUENCE OWNED BY ... statements?

jdarwood007:
@albertlast I'm more concerned how this would function in shared hosting where information like this may be limited or restricted.  I don't have a postgres instance in a shared hosting space to even know.  If your saying this will always produce results even in shared hosting, then thats awesome.

albertlast:
> 
> 
> I really don't think this should be done by navigating the catalog...
> 
> Why didnt you just add ALTER SEQUENCE OWNED BY ... statements?

a source of error,
for get it on one of the 42 sequence and you will never notice to the point where it to late.
better way would be to use serial and use catalog to find the sequence when we need to access them directly.

sbulen:
Yes - The SMF database has 42 sequences.  This is known & finite.  Each is tied to a table & column.  The names have not changed over time. 

The install_2-1_postgresql & upgrade_2-1_postgresql files each simply need those 42 DDL clauses added.  

If we had to support a dynamic DB structure for some reason, it would make sense to query the catalog.  But we don't...

This is a 'keep it simple' moment.  Adding the clauses is sufficient.

albertlast:
nope 146 lines are needed for the 42 seq to be the same,
in your pourpose.
it's high source of issues.

sbulen:
And the database runs just fine without them...  No reason to do this dynamicaly mid-operation.  

The ONLY impacts are when you drop tables.  And even then...  Worst-case scenario...  You can delete the sequences if needed.  It's actually OK to leave them as-is.

sbulen:
What issues?  There is NO bug reported here...

albertlast:
i mention both above,
should i open this?

Sesquipedalian:

Sesquipedalian:
I am closing this PR since this won't be going into SMF 2.1.

@albertlast, if you still think this would be a worthwhile change, please submit a new version for SMF 3.0.

PR #6329, opened 2020-11-02
Set perms before deletions

### LINKED ISSUE #6320: Error deleting upgrade
#### Description

This has been mentioned on the board several times, e.g.:
https://www.simplemachines.org/community/index.php?topic=575295.0
https://www.simplemachines.org/community/index.php?topic=575322.msg4071730#msg4071730
https://www.simplemachines.org/community/index.php?topic=575379.msg4072110#msg4072110

I cannot reproduce it, and I can't find anything in that code that has changed...  Very odd that it keeps showing up.

Logging it here, because it seems to be real...  If anybody has seen this, please share...

sbulen:
Fixes #6320

Setting permissions prior to deletions, in the hopes this will get rid of the issues some folks have when upgrading.

### DIFF
--- a/other/upgrade-helper.php
+++ b/other/upgrade-helper.php
@@ -319,6 +319,23 @@ function quickFileWritable($file)
 }
 
 /**
+ * Delete a file.  Check permissions first, just in case.
+ *
+ * @param string $file
+ */
+function deleteFile($file)
+{
+	if (!file_exists($file))
+		return;
+
+	quickFileWritable($file);
+
+	@unlink($file);
+
+	return;
+}
+
+/**
  * UTF-8 aware strtolower function.
  *
  * @param $string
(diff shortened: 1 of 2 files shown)


other/upgrade-helper.php
```diff
@@ -318,6 +318,23 @@ function quickFileWritable($file)
...
+ * @param string $file
+ */
+function deleteFile($file)
+{
+	if (!file_exists($file))
+		return;
+
+	quickFileWritable($file);
+
+	@unlink($file);
```
live627, line 333: Should the error    suppression be removed here?

Sesquipedalian: Probably not.

Sesquipedalian (approved):

PR #6332, opened 2020-11-04
Fix issue preventing uploads by non-admins

### LINKED ISSUE #6328: Dropzone displaying error page html, not error
#### Description
I think Dropzone may have an issue when a server-side error is encountered.  Instead of displaying the error, it displays the html of the error page served up by the server...

Note, I cannot reproduce this.  Sharing in case it looks familiar to someone?

Multiple threads from the support board:
https://www.simplemachines.org/community/index.php?topic=575402.0
https://www.simplemachines.org/community/index.php?topic=575434.0

This exact dropzone issue has been reported before - apparently if there is a server-side error reported this can happen.  This thread shows how to trap the error instead of serving up the html:
https://stackoverflow.com/questions/39761736/dropzone-image-upload-error-display-how-to-remove-error

**_What you see instead (with console window open):_**
smfhelper-issue2

**_This is the error in the console:_**
smfhelper-issue3

**_If you click on the link in the console error, this is what you see:_**
smfhelper-issue

sbulen:
Fixes #6328 

Before:
attach_bug_before

After:
attach_bug_after

### DIFF
--- a/Themes/default/scripts/smf_fileUpload.js
+++ b/Themes/default/scripts/smf_fileUpload.js
@@ -7,9 +7,6 @@ function smf_fileUpload(oOptions) {
 		previewTemplate = tmp.innerHTML;
 		previewNode.parentNode.removeChild(previewNode);
 
-	if (typeof current_board == 'undefined')
-		var current_board = false;
-
 	// Default values in case oOptions isn't defined.
 	var dOptions = {
 		url: smf_prepareScriptUrl(smf_scripturl) + 'action=uploadAttach;sa=add;' + smf_session_var + '=' + smf_session_id + (current_board ? ';board=' + current_board : ''),
--- a/Themes/default/scripts/smf_fileUpload.js
+++ b/Themes/default/scripts/smf_fileUpload.js
@@ -7,6 +7,9 @@ function smf_fileUpload(oOptions) {
 		previewTemplate = tmp.innerHTML;
 		previewNode.parentNode.removeChild(previewNode);
 
+	if (typeof current_board == 'undefined')
+		current_board = false;
+
 	// Default values in case oOptions isn't defined.
 	var dOptions = {
 		url: smf_prepareScriptUrl(smf_scripturl) + 'action=uploadAttach;sa=add;' + smf_session_var + '=' + smf_session_id + (current_board ? ';board=' + current_board : ''),


jdarwood007:
Why is current_board coming back undefined?  typeof should only be checking that current_board has a defined type which undefined is how all variables which haven't been set yet return in javascript.

This was done because current_board is used in a few places in the uploader script, but if a 3rd party customization calls it, its possible it may not be on a "board".

Maybe a better fix is to change the url in the oOptions to be 3 urls.  urlUpload, urlDelete, urlThumbnail.  Then we can make those generic in the uploader script but allow 3rd party customizations to overload those urls to better handle the board definition?  There may be some other generalization we should do, because it seems right now that the file uploader is very specific to being a topic and doesn't allow much customization, despite the facte we define a oOptions object to allow overloading.

jdarwood007:
FYI, I broke this with PR #6301

jdarwood007:
To add, I backed out of the work I was doing to implant the file uploader into SimpleDesk here and thus I won't use this anymore, so it if reverts it won't change why I fixed it.  The filer uploader while nice and beautiful, sucks to work with as a customization.  There is very limited support to do any integrations or overloading/hooks in the file uploader.  It is so bad I just gave up after reading the code for hours and seeing how I could port it into SimpleDesk with minimal duplication of code.  SimpleDesk actually is easily able to make use of the rest of the attachment handling system (management and download handles) in SMF because of the support for hooks/customization in it.  But the filer uploader does not work and too specific into handling attachments in boards/topics.

I feel others will find the same story when wanting to bring the file uploader into their customizations.  This means they will either:
1.  Ripping and copying the entire process (php & js) just to customize the few pieces needed to make it work.
2.  They will do like I did and give up, retain the original uploading process as while it requires a good chunk of php code, its still simpler than getting the php + js system working together.

sbulen:
There remain two things that bug me here:
- I don't understand why that code breaks it... But it does.  It looks safe, but it's not.
- I think this large red text box will happen any time a fatal error is triggered.  Not sure the best way to address that...

sbulen:
Took me a while...  This addresses my first concern above, while also leaving @jdarwood007 's safety net in there.

The 'var' made it local.  (The position of a 'var' declaration is irrelevant, via "hoisting", the confusing part...)  So...  Under all circumstances, the local variable current_board was set to false...  While the script completely ignored a valid global variable with the same name...

jdarwood007:
Ahh globals.  I didn't fully check if that variable was used in which scopes.  Javascript is kinda loose when it does variables in scopes.

sbulen:
...and it's the kind of thing you can stare right at & not see it...

PR #6337, opened 2020-11-09
Stop counting spiders twice

BrickOzp:
If spiders are visible they are counted twice towards the
total users online since they are included both in num_guests
and in num_spiders.

### DIFF
--- a/Sources/BoardIndex.php
+++ b/Sources/BoardIndex.php
@@ -119,7 +119,7 @@ function BoardIndex()
 
 	// Track most online statistics? (Subs-MembersOnline.php)
 	if (!empty($modSettings['trackStats']))
-		trackStatsUsersOnline($context['num_guests'] + $context['num_spiders'] + $context['num_users_online']);
+		trackStatsUsersOnline($context['num_guests'] + $context['num_users_online']);
 
 	// Are we showing all membergroups on the board index?
 	if (!empty($settings['show_group_key']))


live627:
what is your spider logging set at?

BrickOzp:
For this issue to occur you need to set the Search Engine Tracking level to at least Standard. If it is Disabled no spiders will be detected.

The input $context['num_guests'], $context['num_spiders'] and $context['num_users_online'] comes from getMembersOnlineStats() in Subs-MembersOnline.php

I have created a script to test that function. Btw are there any unit tests for SMF?
I have attached it here so you can try if yourself, just place it in a parallel folder to Sources.
(Had to change it to .txt to be able to upload. Just change to .php)

The problem is that for all settings for "Show spiders in the online list" where the spiders are shown they will be counted both for num_guests and num_spiders.

For example with the setting Show spiders in the online list = admin only. With one user and two spiders in the log_online table:

[num_guests] => 2
[num_spiders] => 2
[num_buddies] => 0
[num_users_hidden] => 0
[num_users_online] => 1

and with the same data for non admins:

[num_guests] => 2
[num_spiders] => 0
[num_buddies] => 0
[num_users_hidden] => 0
[num_users_online] => 1

[testMembersOnline.txt](https://github.com/SimpleMachines/SMF2.1/files/5515386/testMembersOnline.txt)

jdarwood007:
https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/Sources/Subs-MembersOnline.php#L192
```
	$membersOnlineStats['num_users_online'] = count($membersOnlineStats['users_online']) + $membersOnlineStats['num_users_hidden'] - (isset($modSettings['show_spider_online']) && $modSettings['show_spider_online'] > 1 ? count($spider_finds) : 0);
```

The bug exists here where we set num_users_online but then count spiders and num_users_online both into stats.

jdarwood007 (approved):

Sesquipedalian:
I see no reason not to include this in Final.

PR #6339, opened 2020-11-12
Respects maxlength for textareas in custom profile field editing form

Sesquipedalian:
There's not much to say about this one. If the admin set a maximum length for a custom profile field with the "large text" type, we should add the maxlength attribute to the textarea on the profile edit page. Otherwise, the user might enter too much text and then be surprised when it gets truncated.


live627 (approved):
Damn, that ternary is insaaane.

PR #6340, opened 2020-11-13
Remove child bbc tags from enable settings

### LINKED ISSUE #6023: BBC nit - list & li
#### Description
The list & li BBC codes are enabled separately, but neither work properly unless both are enabled.  Behavior is inconsistent.  

### Steps to reproduce
**_SCENARIO 1:_** Both enabled; both buttons visible in editor; both work:
image
image

**_SCENARIO 2:_** li enabled, list disabled; neither button visible in editor; all lists (numbered or not) appear as bullets:
image

**_SCENARIO 3:_** list enabled, li disabled; both buttons visible in editor; neither work:
image

- Version/Git revision: Now
- Database Type: MySQL
- Database Version: 5.7
- PHP Version: 7.3

Issue exists for both 2.1 & 2.0.  Initially reported here:
https://www.simplemachines.org/community/index.php?topic=547426.0

BrickOzp:
It makes no sense of handling bbc codes that requires other tags
to work. For example li that is a sub tag of list.
Remove all bbc child codes from the enable/disable settings page
and let the parent settings control also the children.

Fixes #6023

### DIFF
--- a/Sources/ManageSettings.php
+++ b/Sources/ManageSettings.php
@@ -355,8 +355,29 @@ function ModifyBBCSettings($return_config = false)
 
 		// Clean up the tags.
 		$bbcTags = array();
+		$bbcTagsChildren = array();
 		foreach (parse_bbc(false) as $tag)
 			$bbcTags[] = $tag['tag'];
+			if (isset($tag['require_children']))
+				$bbcTagsChildren[$tag['tag']] = !isset($bbcTagsChildren[$tag['tag']]) ? $tag['require_children'] : array_unique(array_merge($bbcTagsChildren[$tag['tag']], $tag['require_children']));
+
+		// Clean up tags with children
+		foreach($bbcTagsChildren as $parent_tag => $children)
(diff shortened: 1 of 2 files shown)


Sources/ManageSettings.php
```diff
@@ -355,8 +355,29 @@ function ModifyBBCSettings($return_config = false)
 
 		// Clean up the tags.
 		$bbcTags = array();
+		$bbcTagsChildren = array();
 		foreach (parse_bbc(false) as $tag)
 			$bbcTags[] = $tag['tag'];
+			if (isset($tag['require_children']))
+				$bbcTagsChildren[$tag['tag']] = !isset($bbcTagsChildren[$tag['tag']]) ? $tag['require_children'] : array_unique(array_merge($bbcTagsChildren[$tag['tag']], $tag['require_children']));
```
live627: needs brackets

BrickOzp: Your are right.

MissAllSunday (approved):

PR #6341, opened 2020-11-13
Fix an issue when regular members (not admins) do not see attachments…

dragomano:
… on attaching

#### Description
2020-11-13 20_27_29-Window
2020-11-13 20_27_43-Window

### Steps to reproduce
0. be sure regular members have permissions to view and upload attachments
1. log in as a simple member (not admin)
2. try to upload any image

### DIFF
--- a/Sources/ShowAttachments.php
+++ b/Sources/ShowAttachments.php
@@ -69,7 +69,7 @@ function showAttachment()
 	$attachTopic = isset($_REQUEST['topic']) ? (int) $_REQUEST['topic'] : 0;
 
 	// No access in strict maintenance mode or you don't have permission to see attachments.
-	if ((!empty($maintenance) && $maintenance == 2) || !allowedTo('view_attachments'))
+	if ((!empty($maintenance) && $maintenance == 2))
 	{
 		send_http_status(404, 'File Not Found');
 		die('404 File Not Found');
--- a/Sources/ShowAttachments.php
+++ b/Sources/ShowAttachments.php
@@ -69,7 +69,7 @@ function showAttachment()
 	$attachTopic = isset($_REQUEST['topic']) ? (int) $_REQUEST['topic'] : 0;
 
 	// No access in strict maintenance mode or you don't have permission to see attachments.
-	if ((!empty($maintenance) && $maintenance == 2))
+	if (!empty($maintenance) && $maintenance == 2)
 	{
 		send_http_status(404, 'File Not Found');
 		die('404 File Not Found');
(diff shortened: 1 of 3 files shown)


sbulen:
This issue is specific to the preview only, correct?  

I tried to reproduce the problem & the only function impacted is the preview.  But...

I am still having this issue even after applying this PR:
image

dragomano:
It is one more problem, when attachments, loaded by members, do not have a board id.

Sources/ShowAttachments.php
```diff
@@ -69,7 +69,7 @@ function showAttachment()
 	$attachTopic = isset($_REQUEST['topic']) ? (int) $_REQUEST['topic'] : 0;
 
 	// No access in strict maintenance mode or you don't have permission to see attachments.
-	if ((!empty($maintenance) && $maintenance == 2) || !allowedTo('view_attachments'))
+	if (!empty($maintenance) && $maintenance == 2)
```
live627, line 72: Does this show attachments in threads even without permission?

sbulen:
> It is one more problem, when attachments, loaded by members, do not have a board id.

There is no board id on the attachment record.  It has the message, and that is sufficient.  

Or were you referring to something else?

Sources/ShowAttachments.php
dragomano: No, this does not work here. Permissions in topics work well.

dragomano:
Look at parseAttachBBC function in Subs-Attachments.php. There is no check for admin, but there are for members — each attachment has to have a board id to properly showing even if it is just uploaded

sbulen:
Yes, I see what you mean...  But the attachment has not been attached yet, it's still in preview mode.  In fact, there isn't even a message ID.  

There are multiple routines used here that will get the board_id when needed, but it doesn't exist yet...  In fact, the message id is still blank, too.  

It actually checks the user's board permissions 3x, in 3 different ways.  Against the current board, the attachment's board from getAttachMsgInfo and the attachment's board from loadAttachmentContext...   Those latter two checks are both failing because the board is blank because it hasn't been posted to a board yet...

I hope you don't mind @dragomano , but I am going to propose an alternative, more drastic solution...

Sesquipedalian:
Removing the permission check from ShowAttachments.php is not a good solution. That check needs to exist in that function in order to prevent guests from downloading attachments that they are not supposed to be able to access.

Sesquipedalian:
Since the problem is restricted specifically to previews, because no board id has been set yet, a better solution would be to do the following:

1. ~Use a `$_SESSION` variable to keep track of the ids of the attachments when the user is on post editor page.~ We already do this in `$_SESSION['already_attached']`.
2. Expand [this test](https://github.com/SimpleMachines/SMF2.1/blob/d32819c0ae0549b09dabbba714d3e6bcff05523f/Sources/ShowAttachments.php#L72) so that it also checks whether the requested attachment id is in the `$_SESSION['already_attached']` variable. If it is, then allow the download to proceed.

This should allow the user to view the attachment while on the post editing page even if there is no board id set yet, but without compromising security.

PR #6343, opened 2020-11-14
Don't check attach perms 3 times; breaks previews due to no board id

sbulen:
Alternate approach to #6341 ...  Previews are not working for current attachments.

Board attachment permissions are checked 3x in a row here, which feels like overkill.  I see no value added by these two particular checks.  Removing them fixes the problem.  

This one, I believe, is sufficient, just a couple lines before:
https://github.com/SimpleMachines/SMF2.1/blob/8ebc081cc3c8de6519a699d82e0a090575679a7e/Sources/Subs-Attachments.php#L962

If these checks need to be kept (maybe I'm missing something...) then they must be made smarter to allow for the possible lack of a board_id.

### DIFF
--- a/Sources/Subs-Attachments.php
+++ b/Sources/Subs-Attachments.php
@@ -969,10 +969,6 @@ function parseAttachBBC($attachID = 0)
 	if (empty($attachInfo) || empty($attachInfo['msg']) && empty($context['preview_message']))
 		return 'attachments_no_msg_associated';
 
-	// Hold it! got the info now check if you can see this attachment.
-	if ($view_attachment_boards !== array(0) && !in_array($attachInfo['board'], $view_attachment_boards))
-		return 'attachments_not_allowed_to_see';
-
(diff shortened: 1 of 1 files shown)


sbulen:
@live627 - Are these two edits being removed catching anything that the edit on line 962 might miss?

live627:
The more I think this over, the more messy it gets.

sbulen:
I don’t think these particular edits are adding anything...   

If they are needed, I think it is too difficult to add a check for a new post (msg/board id blank) as is done elsewhere here.   Problem is, I just can't see what it would be protecting.  If we can't articulate the test case, I think it should go....

Sesquipedalian:
The checks that this PR removes are just there for efficiency. There's no point in going further and performing useless SQL queries if we already know that the attachment cannot be viewed. But parts of them could be consolidated to avoid redundancy.

PR #6344, opened 2020-11-15
Revises smf_list_timezones & related langfile; adds Subs-Timezones.php

### LINKED ISSUE #6342: Incompatibility between Timezone.{language}.php and CrowdIn
As reported by m4z in [this topic](https://www.simplemachines.org/community/index.php?topic=573012.0), our Timezones language file is causing difficulties with CrowdIn. 

The trouble boils down to the fact that this special language file isn't just a normal translation file. It also functions like a configuration file that fine-tunes SMF's handling of time zones in order to meet the requirements for any given locale. It does this by allowing the localizer to decide which string variables will be defined in their version of the file. Giving a language file any influence on the program logic is, admittedly, unorthodox, but it actually works quite well to create a system that is at once user-friendly, accurate, and adaptable to varying locale requirements. (FYI, achieving all three of those at once is no simple feat when it comes to time zones.)

But however smoothly this all works within SMF itself, it doesn't play nicely with CrowdIn. This is because CrowdIn (quite reasonably) expects all the same string variables to be defined in each translation of the file. So for full compatibility with CrowdIn, we need to define all the possible strings in the file. Doing that, however, will leave the file unable to serve its secondary role as an ersatz config file. That in turn means that we'll need to define that configurtion information elsewhere.

The cleanest way to do this is to add a Subs-Timezones.php file whose only job is to determine that configuration Information. We will also need to add a whole lot more strings to the existing language file so that all possibilities are covered.

The good news is that I've almost finished coding this, so I'll submit a PR soon.

### LINKED ISSUE #6366: Forum time is wrong unless default time zone is set to UTC
#### Description
If the admin sets the forum's default time zone to anything other than UTC (or to somewhere that uses GMT, such as Africa/Abidjan), then the timestamps calculated by `forum_time()` will be incorrect for logged in users. Specifically, the calculated time will be off by an amount equal to UTC offset of that default time zone.

Guests do not appear to be affected by this bug.

### Steps to reproduce
1. In Administration Center ►  Features and Options ►  General, set the forum's default timezone to, e.g., "America/Denver".
2. In Profile of <user> ►  Look and Layout, set your own time zone to the same value (e.g. "Mountain Time," which corresponds to "America/Denver")
3. Look at the time shown in the forum header and notice how wrong it is. For example, "Nov 21, 2020, 6:36 AM" when it should say "Nov 21, 2020, 1:36 PM"

- Version/Git revision: 5d327fb
- Database Type: N/A
- Database Version: N/A
- PHP Version: Any

This bug appears to have been introduced in #5646.

Sesquipedalian:
- Fixes #6342, although it possibly broke my sanity in the process.
- Also simplifies the event editor template, which was needlessly complicated.
- Fixes #6366


sbulen:
I believe the latest changes will re-introduce the initial issues from #5274  fixed in #5646.  All previously existing logic converted from forum time to user time; but when _a user enters a date_ it need to go the other way around...  

I'd rather leave the bulk of #5646 intact.  The current bug (#6366 ) appears to be have been introduced due to some other conflict in approach between #5646 and other time issues over the last 18 months...

sbulen:
At first glance the root conflict (prior to this PR) causing the issue #6366  is between #5611 and #5646.  I think things look OK until it hits the locale logic in timeformat...   It's almost like the locale logic is adding the offset a 2nd time...

sbulen:
This is what I think is happening...  Both #5611 and #5646 modified how timezones & offsets are calculated.  #5646 kept all the calculations within forum_time, applying any user & any system offset required to any timestamp passed.  #5611 kinda split it, as portions of the date math are still in forum_time and portions are now in timeformat with the addition of the locale logic and strftime.  (I think these would double up if users had offsets migrated from 2.0.)

I'd prefer to keep the offset calcs in one place - I don't like having two techniques in two different places.  That's a big part of the problem addressed in #5646 - the offset calcs were done in multiple places & inconsistently...  If you want to keep the offset calcs in timeformat, then maybe we should get rid of forum_time altogether, or, rewrite to just return the system time & do not do date math at all.  

My recommendation would be to have forum_time do all the offset math, system & user.  One place.  And keep timeformat's role to purely _formatting_ the time, not doing any offset calcs.  I think that would fix #6366 while leaving all the fixes in place from #5646 .

sbulen:
Of course, one could argue that it is all just in the display...  

Let me retest the functions fixed in #5646 and we can work from there.  

Truth be told, a non-zero server time_offset should be *extremely rare* and only used for servers that do not store time relative to gmt...  I haven't seen that since the '90s...  And the _user_ time_offset field should be retired (it's 2.0-think...), as the user's time zone should render that obsolete...  The problem of course with a user time_offset is that it is fixed, when it shouldn't be, due to dst...

sbulen:
And...  After all that....   It seems to work.

sbulen (approved):

Sesquipedalian:
In 3.0 we should remove `forum_time()` and all manual time offset settings, and instead rely on PHP's built-in robust date and time support. Manually trying to manipulate those values is a hangover from ancient, dark days. We should probably also remove all references to knapping stone axes, astrolabes, and treadmill cranes while we're at it. 😉

PR #6347, opened 2020-11-17
Correct SQL for gender count

albertlast:
issue: https://www.simplemachines.org/community/index.php?topic=575645.0


BrickOzp (commented):
LEFT JOIN will catch all members without a value for cust_gender and display an empty label like
Male: 2
: 2

BrickOzp:
That could be fixed by importing the default value for those who hasn't set any value.
That seems correct since the default value is shown in the profile.

SELECT COUNT(id_member) as total_members, gender FROM (
    SELECT mem.id_member, COALESCE(t.value, (
        SELECT default_value FROM {db_prefix}custom_fields WHERE col_name={string:gender_var})) AS gender
    FROM {db_prefix}members AS mem
    LEFT JOIN {db_prefix}themes AS t ON (
    mem.id_member = t.id_member AND t.variable = {string:gender_var} AND
    t.id_theme = {int:default_theme}) 
   WHERE is_activated = {int:is_activated}) as a
GROUP BY gender

albertlast:
Well since the query is very heavy,
I would fav to make a pre query for default and
add a coalesce,
Von: BrickOzp <notifications@github.com>
Gesendet: Donnerstag, 19. November 2020 14:09
An: SimpleMachines/SMF2.1
Cc: albertlast; Author
Betreff: Re: [SimpleMachines/SMF2.1] Correct SQL for gender count (#6347)


That could be fixed by importing the default value for those who hasn't set any value.
That seems correct since the default value is shown in the profile.

SELECT COUNT(id_member) as total_members, gender FROM (
SELECT mem.id_member, COALESCE(t.value, (
SELECT default_value FROM {db_prefix}custom_fields WHERE col_name={string:gender_var})) AS gender
FROM {db_prefix}members AS mem
LEFT JOIN {db_prefix}themes AS t ON (
mem.id_member = t.id_member AND t.variable = {string:gender_var} AND
t.id_theme = {int:default_theme})
WHERE is_activated = {int:is_activated}) as a
GROUP BY gender

BrickOzp:
I think you need that sub query as well. Other wise you could end up with two default value entries, one for those who has selected the default value and one for those who hasn't selected anything.

albertlast:
well in the execution side it make a bit different if the sub query is in the from or select field.
but the behavior in general i testet atm if this on every db so.

okay tested in sqlite, mysql and pg, all behave the same: https://www.db-fiddle.com/f/93Xhi11jceMQy5XdM69RY6/0

albertlast:
Just for my fun,
i tested my and @BrickOzp  version
my https://explain.depesz.com/s/DVqZ
his https://explain.depesz.com/s/IlEW

noteable both run similiar,
because postgres use in his version a initplan (a static sub query).

but i would be not 100% sure if our mysql version 5.0 also got this kind of tricks.

smfbigguy:
Works great. Thank you

BrickOzp (approved):
Tested and it works

MissAllSunday (approved):

MissAllSunday:
Hi @albertlast  can you fix the conflicts so this can be merged, please.

albertlast:
done

MissAllSunday (approved):

PR #6364, opened 2020-11-21
Fix category order

live627:
I have ten categories created via Populate.php, and they are reverse order in the board index verses the admin panel.

This commit fixes that. Sequential scans on both the boards and categories tables are already very fast due to the datatypes of said tables.

### DIFF
--- a/Sources/Subs-Boards.php
+++ b/Sources/Subs-Boards.php
@@ -1219,9 +1219,9 @@ function getTreeOrder()
 
 	$request = $smcFunc['db_query']('', '
 		SELECT b.id_board, b.id_cat
-		FROM {db_prefix}boards AS b
-		ORDER BY b.board_order',
-		array()
+		FROM {db_prefix}categories AS c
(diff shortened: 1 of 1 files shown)


albertlast:
Well i don't understand how the used datatyle related to you sequence scans mentions,
but the category table hav very low amount of columns,
so the read time to this add not no mutch to the overall.
in my 100 boards with 10 category,
the query time changed from 0,080 ms to 0,110 ms.
What i guess get improved the speed of this code,
is the drop of cache logic code,
since the check take more time as the addtional runtime which is needed.

albertlast (approved):
sql code looks fine (didn't check the complete code)

jdarwood007:
Is there a reason we dropped the caching?  If Populate.php is messing this up, it should drop the cache once its done adding the boards.

albertlast:
How mutch time takes a common cache to check and response the data?

live627:
> Is there a reason we dropped the caching?

No. Is an error from pushing too  much code from another branch.

PR #6365, opened 2020-11-21
Remove "size" attribute for input "number"

dragomano:
Input "number" type hasn't "size" attribute
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number


jdarwood007:
It doesn't have a size, but it has some min/max.  Maybe ensure that we can set a min/max and this could make some assumptions like that the ones with a size of 4 didn't intend to have a number more than 9,999.  If we have this in settings, we should ensure we can set min/max and step for a number.

PR #6370, opened 2020-11-23
Set first day of week to date pricker

BrickOzp:
The firstDay property was not set on the date pricker,
resulting in that sunday was always the first day of week
regardless of the first day of week setting.


Sesquipedalian (approved):
Nice.

PR #6371, opened 2020-11-24
Avoid errors for unknown time zones

BrickOzp:
If a time zone is not defined in the tz database installed on
the server, several errors was written to the error log.
Check that tz can be created before using it.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -5987,7 +5987,10 @@ function smf_list_timezones($when = 'now')
 		if ($tzid == 'UTC')
 			continue;
 
-		$tz = timezone_open($tzid);
+		$tz = @timezone_open($tzid);
+
+		if ($tz == null)
+			continue;
 
 		// First, get the set of transition rules for this tzid
 		$tzinfo = timezone_transitions_get($tz, $when, $later);


Sources/Subs.php
```diff
@@ -5987,7 +5987,10 @@ function smf_list_timezones($when = 'now')
 		if ($tzid == 'UTC')
 			continue;
 
-		$tz = timezone_open($tzid);
+		$tz = @timezone_open($tzid);
+
+		if ($tz == NULL)
```
live627: notice that null is lowercase everywhere; for consistency, can you set this lowercase?

Sesquipedalian:
Which time zone identifiers caused the errors in your tests, @BrickOzp? Your change here is a good safety measure, but I would also like to replace the problematic keys defined in `$tzid_metazones` so that we can avoid such errors whenever possible.

BrickOzp:
It was America/Punta_Arenas

Sesquipedalian:
Thanks. I'll implement a fallback for that and for a few others I've now found that might cause similar issues.

Sesquipedalian (approved):

PR #6382, opened 2020-11-30
Do not check post_attachment permission at upload

### LINKED ISSUE #6381: Not possible to upload attachments if needs moderation
#### Description
If user needs requires approval for attachments it is not possible to upload attachments

### Steps to reproduce
1. Enable post moderation
2. Set requires approval for attachments for the user group
3. Try to upload an attachment with the user from the user group.
4. Obscure error message is displayed.
upload error

- Version/Git revision: 40a81b12464fe9
- Database Type: mysql
- Database Version: 
- PHP Version:5.6

BrickOzp:
If the user requires approval for attachments, the
post_attachments permission is not active. So this
permission check will fail. The correct permission
checks are made later so this check is redundant.
Also the response is expected to be json, and this
check will result in a html error page.

Fixes #6381

### DIFF
--- a/Sources/Attachments.php
+++ b/Sources/Attachments.php
@@ -128,8 +128,6 @@ class Attachments
 	{
 		global $smcFunc, $sourcedir;
 
-		isAllowedTo('post_attachment');
-
 		require_once($sourcedir . '/Subs-Attachments.php');
 
 		// Need this. For reasons...


BrickOzp:
Problem was introduced by https://github.com/SimpleMachines/SMF2.1/commit/1ba62ea583e369531aeddb4935717b6423f1aefb

MissAllSunday (approved):

PR #6385, opened 2020-11-30
Fix php8 errors

dragomano:


MissAllSunday (approved):

Sources/Subs-Db-mysql.php
```diff
@@ -728,7 +728,7 @@ function smf_db_error($db_string, $connection = null)
  * @param object $connection The connection to use (if null, $db_connection is used)
  * @return mixed value of the first key, behavior based on returnmode. null if no data.
  */
-function smf_db_insert($method = 'replace', $table, $columns, $data, $keys, $returnmode = 0, $connection = null)
+function smf_db_insert($method = 'replace', $table = '', $columns = array(), $data = array(), $keys = array(), $returnmode = 0, $connection = null)
```
live627: All these parameters are required.

```suggestion
function smf_db_insert($method, $table, $columns, $data, $keys, $returnmode = 0, $connection = null)
```

If $method is empty or unrecognized, it then does a regular INSERT.

https://github.com/SimpleMachines/SMF2.1/blob/8593cf460a0a81dc50df8b4cfdcef03847ad073e/Sources/Subs-Db-mysql.php#L780

live627 (changes requested):

Sources/Subs-Db-mysql.php
dragomano: Done.

live627 (approved):

PR #6389, opened 2020-12-02
Only remove not uploaded attachments on cancel all

### LINKED ISSUE #6388: Cancel all button removes uploaded attachments
#### Description
Using the cancel all button when an attachment already is uploaded will remove also that attachment from the list.
But when posting the message, that attachment is attached.
By using this way you can upload more attachments than the maximum allowed.

The upload all and cancel all buttons are not hidden when not needed.

### Steps to reproduce
1. Attach a file and upload it.
2. Attach another file without uploading it.
3. Press cancel all.
4. Both attachments are removed.
5. Post the message.
6. The first attachment is attached.

- Version/Git revision: d43b0f
- Database Type: mysql
- Database Version:
- PHP Version: 5.6

BrickOzp:
The cancel all button should only cancel not already
uploaded attachments.
Also hide the cancel all and upload all buttons when
there are no files to cancel or upload.
Hide the progress bar when there are no attachments
at all.

Fixes #6388

### DIFF
--- a/Themes/default/scripts/smf_fileUpload.js
+++ b/Themes/default/scripts/smf_fileUpload.js
@@ -106,6 +106,16 @@ function smf_fileUpload(oOptions) {
 				done();
 			}
 		},
+		hideFileProgressAndAllButtonsIfNeeded: function () {
+			// Hide the cancel and upload all buttons if there is nothing to cancel/upload anymore.
+			if (myDropzone.getFilesWithStatus(Dropzone.ADDED).length == 0) {
+				$('div#attachment_upload').find('#attach_cancel_all, #attach_upload_all').hide();
+			}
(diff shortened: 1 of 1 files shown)


Themes/default/scripts/smf_fileUpload.js
```diff
@@ -515,4 +516,15 @@ function smf_fileUpload(oOptions) {
 			myDropzone.emit("complete", mock);
 		});
 	}
+
+	function hide_file_progress_and_all_buttons_if_needed() {
```
MissAllSunday: nitpick: can you change the function name to camelCase please?

Maybe even change it to a closure and use it as a property of the parent function

BrickOzp: Done

MissAllSunday (approved):

PR #6395, opened 2020-12-07
Fixes bug where non-admins could not preview attachments

Sesquipedalian:
Checks ~`$_SESSION['already_attached']`~ `$_SESSION['attachments_can_preview']` to see if the user is currently editing the post that holds this attachment. If so, then we allow the user to preview the attachment. This approach allows us to bypass the problem that no board id has been set yet when previewing messages that have not yet been posted.

Closes #6341
Closes #6343

### DIFF
--- a/Sources/Attachments.php
+++ b/Sources/Attachments.php
@@ -461,6 +461,12 @@ class Attachments
 			$this->_attachSuccess = $_SESSION['already_attached'];
 
 		unset($_SESSION['temp_attachments']);
+
+		// Allow user to see previews for all of this post's attachments, even if the post hasn't been submitted yet.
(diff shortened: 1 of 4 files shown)


dragomano:
And now try to edit a post that already exists. You will see only those attachments that loaded in the current session.
2020-12-08 13_15_53-Редактировать сообщение - Opera

Sesquipedalian:
Good catch. I mistakenly thought that `$_SESSION['already_attached']` would list all the current attachments as well as any newly uploaded attachments. I've fixed this in 2db8458 by introducing `$_SESSION['attachments_can_preview']` to keep track of all the attachments in the post that is being written/edited.

PR #6404, opened 2020-12-11
Mysql mb4

sbulen:


live627:
i   s e e   y o u

sm.org is still on mysql 5.1 according to the mediawiki installation...

Issue #6405, opened 2020-12-13
Search issue - multi-byte words truncated


sbulen:
#### Description
If there are multi-byte terms in a post, they are translated to html entities, which can take up to 8 or 9 bytes per character.  The problem is that search words are truncated at 20 characters.  Thus, any string of multibyte characters can result in truncation - mid html-entity, e.g.:
![image](https://user-images.githubusercontent.com/23568484/102021905-226d8d80-3d38-11eb-9e7c-ecf55a7b3706.png)

Note the truncated html entities, e.g., `&#` and `&#66`.  This is in the log_search_subjects table.

This causes further issues down the road, e.g., executing an html entity to utf8 conversion, you can get:
![image](https://user-images.githubusercontent.com/23568484/102021933-606ab180-3d38-11eb-9e63-236df527c95c.png)
...as that word really isn't unique once it has been truncated.

This issue exists in 2.0 as well.  

In 2.1, this issue is restricted to 4-byte character usage, as anything <4-bytes is no longer converted to html entities - though they may be brought forward during an upgrade.  

### Steps to reproduce
1. Post this: 𐌼𐌰𐌲 𐌲𐌻𐌴𐍃 𐌹̈𐍄𐌰𐌽, 𐌽𐌹 𐌼𐌹𐍃 𐍅𐌿 𐌽𐌳𐌰𐌽 𐌱𐍂𐌹𐌲𐌲𐌹𐌸.
2. Look at log_search_subjects for that post

### Environment (complete as necessary)
- Version/Git revision: current
- Database Type: mysql
- Database Version: 5.7
- PHP Version: 7.4

### Additional information/references
4-byte characters are not common outside the use of emojis, certain symbols, and ancient texts....  
But the SMF crowd is **_exactly_** the kinda crowd to use emojis, certain symbols, and ancient texts...

albertlast:
Based on the description,
The issue exists only in MySQL.

In MySQL could be a workaround convert the database to mb4 and change the parameter in sub mysql file.

sbulen:
Correct, mysql only.  MB4 will require some db changes & increasing minimum version of mysql.  

But mb4 is definitely the right direction.

albertlast:
What i mean,
smf is already prepared for this,
since i added some time ago this parameter, which can be changed by hand:
[...]

PR #6406, opened 2020-12-14
Allow mod added tables to be mb4 when smf is running in mb4

albertlast:
Based on the discussion #6405
i checked the existing mb4 logic of smf and
found a spot which got not covered atm.

In general when someone plan to convert his smf 2.1 to mb4,
the programm logic expect that not only the existing column get changed,
also the default setting of the tables needs to be changed to mb4.

### DIFF
--- a/Sources/DbPackages-mysql.php
+++ b/Sources/DbPackages-mysql.php
@@ -197,8 +197,11 @@ function smf_db_create_table($table_name, $columns, $indexes = array(), $paramet
 	}
 
 	$table_query .= ') ENGINE=' . $parameters['engine'];
-	if (!empty($db_character_set) && $db_character_set == 'utf8')
+	if (!empty($db_character_set) && $db_character_set == 'utf8' && !$smcFunc['db_mb4'])
 		$table_query .= ' DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci';
+
+	if (!empty($db_character_set) && $db_character_set == 'utf8' && $smcFunc['db_mb4'])
+		$table_query .= ' DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci';
 
 	// Create the table!
 	$smcFunc['db_query']('', $table_query,


Sources/DbPackages-mysql.php
```diff
@@ -197,9 +197,12 @@ function smf_db_create_table($table_name, $columns, $indexes = array(), $paramet
 	}
 
 	$table_query .= ') ENGINE=' . $parameters['engine'];
-	if (!empty($db_character_set) && $db_character_set == 'utf8')
+	if (!empty($db_character_set) && $db_character_set == 'utf8' && !$smcFunc['db_mb4'])
```
MissAllSunday, line 200: Optional, might be good move the db_mb4 check into a ternary inside the if to avoid having to double check for $db_character_set

```suggestion
	if (!empty($db_character_set) && $db_character_set == 'utf8')
	{
	    $table_query .=  ' DEFAULT CHARSET=utf8 . (!$smcFunc['db_mb4'] ? '  COLLATE=utf8_general_ci' : 'mb4 COLLATE=utf8mb4_general_ci');
	}
```

Or make it cleaner by defining the charset as a variable

albertlast: both in my eyes make the code harder to read, reason why i choose this way.

MissAllSunday: Same ternary logic was applied on another PR: https://github.com/SimpleMachines/SMF2.1/pull/6273/files

albertlast: Your suggest code changes got more line of changes and
is harder to read as the existing one.

So thanks but no.

MissAllSunday: I think having repetitive code is actually harder to read and can be confusing since both ifs are the same except for the last part but as you wish

Bmgxyzp (approved):
albertlast:aBitMoreMB4

Sesquipedalian:
~Note: This PR targets release-2.1, but it has been assigned to the milestones for SMF 3.0. The PR will need to be updated and resubmitted to target release-3.0. I am leaving it open for now so that it doesn't get forgotten.~

Actually, this PR will become moot once we convert SMF 3.0 to use utf8mb4 by default on MySQL. But this code will likely be for the latter, so I am leaving it open for now to make it easier to find for reference.

PR #6408, opened 2020-12-14
Fix issue when a time bbc tag is not parsed

dragomano:
If we use any language (except English) we have not worked parsing for time tag:

How to see this problem:

1. Select any language, except English.
2. Post a message with a time BBC tag.
3. Edit this message and view it again.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -2078,7 +2078,7 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 					if (is_numeric($data))
 						$data = timeformat($data);
 
-					$tag['content'] = '<strong>$1</strong>';
+					$tag['content'] = '<span class="bbc_time">$1</span>';
 				},
 			),
 			array(
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -2077,8 +2077,8 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 				{
 					if (is_numeric($data))
 						$data = timeformat($data);
-					else
-						$tag['content'] = '[time]$1[/time]';
+
+					$tag['content'] = '<strong>$1</strong>';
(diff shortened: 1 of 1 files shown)


BrickOzp (approved):
This problem exists even if english is selected as language.
I have verified that this fix solves the issue.

However this also adds < strong > tags to time. Is that what we want?

dragomano:
We can add any other tag (`span`, etc), it is just for attraction. I thought that if someone uses a timestamp in a text, he would still want to highlight it.

BrickOzp:
After investigated this further it is related to the selected time format, and not the language.
If using a known time format when writing the tag e.g. [time]1973-11-27, 08:43:00[/time], [time]12234345[/time] or [time]yesterday[/time] it works, but not when using [time]14 December 2020 kl. 00:00[/time]

If the date can be parsed it is stored in the db as unix timestamp [time]12234345[/time].
When viewing and modifying it is displayed in the date time format selected e.g. [time]1973-11-27, 08:43:00[/time] or [time]14 December 2020 kl. 00:00[/time].

One solution could be to use use a different date format when modifying so it could be parsed again.

BrickOzp:
The parsing is done in Subs-Post.php and uses strtotime.
Manual says "The function expects to be given a string containing an English date format and will try to parse that format into a Unix timestamp"
I guess that's why it doesn't work with other languages or time formats.

BrickOzp:
But that is another fix and your change is needed anyway.

dragomano:
Should I change `strong` to `span` or `span class="time"`?

MissAllSunday:
The time class adds a float: right instruction that can cause issues with the rest of the posting template, might be good to change it to something new like bbc_time so theme authors can use it if they want to.

As for strtotime for parsing time tags it is indeed another issue

BrickOzp (approved):

MissAllSunday (approved):

PR #6427, opened 2020-12-27
Introduce Github Actions and remove Travis CI

jdarwood007:
Not much to explain here.  Travis-CI has introduced changes which we believe may not be ideal for the future of open source projects.

It should be noted that Testing PHP 5.4, 5.5 and 5.6 are slow.  It adds a very noticeable time delay to testing.  This is due to the low testing these versions receive.

This PR may initially fail because we are pulling travis CI support.  Will be correcting that.


albertlast:
The lastest ubuntu version shouldn't support sutch old php not any more,
did you double check this if this working?

jdarwood007:
It seems to work fine.  They are compiled, not prebuilt packages.  So they should continue to work on 20.04

https://github.com/actions/virtual-environments/issues/1816

This doesn't even show support for anything prior to 7.1 for 18.04, but it works.
I am using this for php: https://github.com/nanasess/setup-php

Which claims 5.4 support and ubuntu 18.04 and 20.04 support.

I'm using that because it has matrix support and is able to do concurrent runs.  This entire setup for GitHub actions is more efficient than our travis setup.  We only test languages,sign off, etc once.  But we check php syntax for all php versions.  Most of the checks complete in a few minutes.  But 5.4/5.5 did require building so they took the longest at over 10 minutes.

albertlast:
i was already asking myself why this repo is used, but it make now sense.

I wold expect that the output would be already visable in this pr,
is ther any special place to see the github action result?

jdarwood007:
It shows up under my fork: https://github.com/jdarwood007/SMF2.1/actions
Not sure why it doesn't show up here.  SimpleDesk did when I converted it.

jdarwood007:
As a FYI,

If migrating to 20.04 does break it in some unseen way.  Then we can always split the syntax checker script:
```
name: PHP Syntax Check

on: [push, pull_request]

jobs:
  syntax-checker:
    runs-on: ${{ matrix.operating-system }}
    strategy:
      matrix:
        operating-system: [ ubuntu-latest ]
        php: [ '7.1', '7.2', '7.3', '7.4', '8.0' ]
    name: PHP ${{ matrix.php }} Syntax Check
    steps:
      - uses: actions/checkout@master
        with:
          submodules: true
      - name: Setup PHP
        id: SetupPHP
        uses: nanasess/setup-php@master
        with:
          php-version: ${{ matrix.php }}
      - run: php ./other/buildTools/check-php-syntax.php ./
  syntax-checker-legacy:
    runs-on: ${{ matrix.operating-system }}
    strategy:
      matrix:
        operating-system: [ ubuntu-18.04 ]
        php: [ '5.4', '5.5', '5.6' ]
    name: PHP ${{ matrix.php }} Syntax Check
    steps:
      - uses: actions/checkout@master
        with:
          submodules: true
      - name: Setup PHP
        id: SetupPHP
        uses: nanasess/setup-php@master
        with:
          php-version: ${{ matrix.php }}
      - run: php ./other/buildTools/check-php-syntax.php ./
```

albertlast:
because of you example i notice that you miss php 7.0?

.github/workflows/Syntax-Check.yml
```diff
@@ -0,0 +1,22 @@
...
+
+on: [push, pull_request]
+
+jobs:
+  syntax-checker:
+    runs-on: ${{ matrix.operating-system }}
+    strategy:
+      matrix:
+        operating-system: [ ubuntu-latest ]
+        php: [ '5.4', '5.5', '5.6', '7.0', '7.1', '7.2', '7.3', '7.4', '8.0' ]
```
live627, line 11: Is this large matrix really necessary? Sooo many wasted resources on redundant testing.

albertlast: well this the reality of smf,
as dev you should had all version local to test your stuff.

So from my pov, is this the least point where the stuff is important to change.

jdarwood007: It just does a syntax check for all supported php versions.  This should help prevent us from making a mistake of using a invalid syntax in a PHP version we don't have on our dev environments.  We could still introduce those, but less likely to be fatal hopefully.

live627:
@jdarwood007 curious as to the reason you decided not to use the code I sent in https://github.com/SimpleMachines/BuildTools/pull/12

jdarwood007:
I missed them.  We can still merge up your PR there and make it happen here if your PR is still compatible with Travis since this hasn't merged yet.

live627:
Compatible with Travis? I thought we were trying to ditch them.

But yes, what I have works on Travis. I even include [instructions for it in the readme](https://github.com/live627/BuildTools/tree/develop#ci).

PR #6430, opened 2020-12-27
Fix autoload for cache classes

### LINKED ISSUE #6429: The caching does not work
#### Description
The caching does not work — cache dir is always empty (there are .htaccess, index.php, and db_last_error.php only).
The reason - autoloading for Cache APIs classes does not work.

### Steps to reproduce
1. Enable any level of caching, save settings.
2.  Take a tour around your forum, then open cache dir.

- Version/Git revision: 4ba5dd2
- PHP Version: 7.2

dragomano:
Fixes #6429

### DIFF
--- a/SSI.php
+++ b/SSI.php
@@ -87,33 +87,6 @@ $smcFunc = array();
 // Initiate the database connection and define some database functions to use.
 loadDatabase();
 
-// Load installed 'Mods' settings.
-reloadSettings();
-// Clean the request variables.
-cleanRequest();
-
-// Seed the random generator?
-if (empty($modSettings['rand_seed']) || mt_rand(1, 250) == 69)
-	smf_seed_generator();
-
-// Check on any hacking attempts.
-if (isset($_REQUEST['GLOBALS']) || isset($_COOKIE['GLOBALS']))
-	die('No direct access...');
-elseif (isset($_REQUEST['ssi_theme']) && (int) $_REQUEST['ssi_theme'] == (int) $ssi_theme)
-	die('No direct access...');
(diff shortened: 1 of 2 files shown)


MissAllSunday:
I'm somehow worried about moving the autlolad call so early, specially for mods that require being autoloaded early but assume most of SMF's logic is already available at that point. A portal for example.

Might be good to have another "post logic" autoload for those cases.

dragomano:
I tested this. If you leave _spl_autoload_register_ in the same place, the fix will not work. There is no sense to use two autoload functions in different places.

MissAllSunday:
I agree it needs to be changed.

My concern is, since we ultimately just do a require_once for the registered classes, mods that assume SMF's logic is already available at the time they are loaded will have the same issue the cache class currently has.

jdarwood007:
The problem here is that the call to the hook will fail here since we have loaded the database, but have not loaded up any data yet regarding the hooks.  But however, that will be loaded after reloadSettings(); has completed.  So it would only mean that anything wanting to happen between loadDatabase() and the end of reloadSettings(); would fail.  The only integrations we offer there are SMF_INTEGRATION_SETTINGS, which is a simple array key to add additional hooks at runtime, integrate_pre_include which allows adding files and integrate_pre_load which allows calling additional work prior to completing reloadSettings.

I would take it as this is very unlikely to cause any issues that a mod author would have to deal with.  I can't see one.

However this fix also needs to occur in SSI.php as well since it will suffer the same problem.

Also, any reason we can't add the spl_autoload_register into a setup function in Load.php that then simplifies the logic between index.php and SSI.php?

live627:
@dragomano can you also apply this fix to SSI.php?

live627 (approved):

PR #6304, opened 2020-10-17
chore: Remove spell check feature

MissAllSunday:
Fixes #5847

PR #6436, opened 2021-01-02
Remove the tails left after removing the spellcheck feature

dragomano:
Continues #6304


MissAllSunday:
The  `if ($context['show_spellchecking'])` checks

Where intentionally left there to offer support for legacy code and/or allowing a mod to bring back the functionality.

PR #6444, opened 2021-01-08
Add some fixes for MemcacheImplementation and MemcachedImplementation

dragomano:
The first commit fixes invisible settings for Memcached API:
now:
2021-01-09 00_01_54-Server Settings — Firefox Developer Edition
with fix:
2021-01-09 00_00_57-Server Settings — Firefox Developer Edition

The second commit fixes issue when caching (via Memcached API) did not work due to the wrong processing of the server list.

### DIFF
--- a/Sources/Cache/APIs/MemcacheImplementation.php
+++ b/Sources/Cache/APIs/MemcacheImplementation.php
@@ -27,7 +27,7 @@ if (!defined('SMF'))
  */
 class MemcacheImplementation extends CacheApi implements CacheApiInterface
 {
-	const CLASS_KEY = 'cache_memcached';
+	const CLASS_KEY = 'cache_memcache';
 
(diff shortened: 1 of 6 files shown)


live627:
@MissAllSunday the third commit here modifies the settings file; do you think it is ok?

If so, there is more to do here

https://github.com/SimpleMachines/SMF2.1/blob/7ea689343cc5fc667908d3caac6387e9994ae27e/Sources/Cache/APIs/MemcacheImplementation.php#L61

dragomano:
Sure, it is possible to use a common setting "Memcache/memcached servers" (as in SMF 2.0) for both settings, Memcache, and Memcached. In this case, I am ready to change my PR.
2021-01-16 18_15_24-Server Settings — Firefox Developer Edition

Issue #6450, opened 2021-01-11
Users with Update Boards & Categories permissions can't see boards & categories


sbulen:
#### Description
Reported on the forum & confirmed.  

Very odd, I do not see this on my wamp, but I do see it on my hosted linux environment.  

On both wamp & linux, if I view the permissions of the user in question, the restricted boards are listed as not accessible to this user.

### Steps to reproduce
1. On a linux environment....
2. Give 'Global Moderators' the ability to update boards & categories under Admin | Forum | Manage Boards & Categories | Settings
3. Create a category only visible to Global Moderators
4. Create a board under that category only visible to Global Moderators
5. Create a user & assign to Global Moderators as his primary group
6. Logon as that user & the boards are missing...

### Environment (complete as necessary)
- Version/Git revision:  RC3
- Database Type: Mysql
- Database Version: 5.6
- PHP Version: Both 5.4 & 7.3

### Additional information/references
Reported on the support forum here:
https://www.simplemachines.org/community/index.php?topic=576328.0

jdarwood007:
Steps 3-5 mention Global Admin.  Is it supposed to be Global Mod?

I would assume that those with the ability to update boards/cats could only update those they can see.

sbulen:
Yes, I'll correct the above, thanks.  

These boards are configured to be visible to Global Moderators, but aren't.  Unix only, which is odd...

sbulen:
Screenshots:

Global Moderators have permission to Manage Boards & Categories:
![image](https://user-images.githubusercontent.com/23568484/104269645-a60bbe80-544b-11eb-9653-75e37666eca1.png)

Global Moderators have access to the board:
![image](https://user-images.githubusercontent.com/23568484/104269819-0569ce80-544c-11eb-89ba-0bb8770dc722.png)

User is a Global Moderator:
![image](https://user-images.githubusercontent.com/23568484/104269867-1fa3ac80-544c-11eb-96fd-8ccb56cfb412.png)

But cannot see the board at all (unix only, fine on windows):
![image](https://user-images.githubusercontent.com/23568484/104269936-4235c580-544c-11eb-9697-0d212891a23e.png)
![image](https://user-images.githubusercontent.com/23568484/104270029-67c2cf00-544c-11eb-9406-7304b23d0682.png)

Both unix & windows have this message on the user profile permissions tab:
![image](https://user-images.githubusercontent.com/23568484/104270202-cab46600-544c-11eb-832e-397318709b7a.png)

[...]

PR #6452, opened 2021-01-12
Check board_manager_groups when displaying permissions

sbulen:
Now that board managers can see all boards, this needs to be reflected on the 'Show Permissions' view.  

Otherwise, folks with 'Manage Boards and Categories' end up with boards on their restricted boards list, which doesn't make sense.

**_Before change:_**
board_perms_conflict

**_After change:_**
board_perms_conflict-after

Saw this when researching #6450 .

### DIFF
--- a/Sources/Profile-View.php
+++ b/Sources/Profile-View.php
@@ -3109,7 +3109,8 @@ function showPermissions($memID)
 	$context['no_access_boards'] = array();
 	while ($row = $smcFunc['db_fetch_assoc']($request))
 	{
-		if ((count(array_intersect($curGroups, explode(',', $row['member_groups']))) === 0) && !$row['is_mod'] && (!empty($modSettings['board_manager_groups']) && count(array_intersect($curGroups, explode(',', $modSettings['board_manager_groups']))) === 0))
+		if ((count(array_intersect($curGroups, explode(',', $row['member_groups']))) === 0) && !$row['is_mod']
+		&& (!empty($modSettings['board_manager_groups']) && count(array_intersect($curGroups, explode(',', $modSettings['board_manager_groups']))) === 0))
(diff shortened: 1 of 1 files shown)


Sources/Profile-View.php
```diff
@@ -3109,7 +3109,7 @@ function showPermissions($memID)
 	$context['no_access_boards'] = array();
 	while ($row = $smcFunc['db_fetch_assoc']($request))
 	{
-		if (count(array_intersect($curGroups, explode(',', $row['member_groups']))) === 0 && !$row['is_mod'])
+		if ((count(array_intersect($curGroups, explode(',', $row['member_groups']))) === 0) && !$row['is_mod'] && (!empty($modSettings['board_manager_groups']) && count(array_intersect($curGroups, explode(',', $modSettings['board_manager_groups']))) === 0))
```
albertlast: to many logic in one line for my pov.

Sesquipedalian:
```suggestion
		if (count(array_intersect($curGroups, array_merge(explode(',', $row['member_groups']), explode(',' $modSettings['board_manager_groups'])))) === 0 && !$row['is_mod'])
```

Sesquipedalian (changes requested):
Logic is good, but see suggestion for a more parsimonious way to accomplish it.

sbulen:
Recent updates broke this PR...
image

I will simplify the language, retest & resubmit...

sbulen:
I split the logic to make it more readable.
Retested, & tests OK.
I did leave the existing logic, to be consistent with the checks in Load:
https://github.com/SimpleMachines/SMF2.1/blob/6a158f4d4c6d3fb00d165d0537837ef698303b3a/Sources/Load.php#L832

live627 (approved):

PR #6453, opened 2021-01-13
Convert text into string in upgrade.php

### LINKED ISSUE #6422: Language string upgrade.php
SMF RC3 Current build

Missing language string https://github.com/SimpleMachines/SMF2.1/blob/8bb2097366a12c13f0298f4845616cb9f9308107/Themes/default/languages/Install.english.php#L356

in upgrade.php https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/other/upgrade.php#L3995
`<a href="', $_SERVER['PHP_SELF'], '">Click here to try again.</a>`

change to
`<a href="', $_SERVER['PHP_SELF'], '">', $txt['upgrade_respondtime_clickhere'], '</a>`

BrickOzp:
Some strings where hard coded. Convert to strings.

Fixes #6422

### DIFF
--- a/Themes/default/languages/Install.english.php
+++ b/Themes/default/languages/Install.english.php
@@ -302,6 +302,7 @@ $txt['upgrade_json_completed'] = 'Convert to JSON Complete! Click Continue to Pr
 $txt['upgrade_executing'] = 'Executing:';
 $txt['upgrade_of'] = 'of';
 $txt['upgrade_admin_login'] = 'Admin Login:';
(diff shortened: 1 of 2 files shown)


albertlast:
Since you move here in the error handling part,
could it not be that the error txt is not loaded,
so you would create an error in the error handling part?

BrickOzp:
Maybe, but to me it looks like the language file is always loaded when the upgrade.php is run

live627:
Yes, it is.

PR #6471, opened 2021-01-20
Add right padding to language selector for mobile

BrickOzp:
If viewing the forum in a mobile or other small screen
device, the language selector form is right at the edge
of the screen. Add a padding to match the other boxes.


BrickOzp:
Before fix:
Screenshot_20210120-141825

After fix:
Screenshot_20210120-141842

PR #6476, opened 2021-01-23
Add spoiler as a feature

dragomano:
Adds a spoiler tag via html5 details tag.

2021-01-27 21_16_55-Welcome to SMF! — Firefox Developer Edition


BrickOzp:
Nice. I have wanted this function.
Any reason you went for a config of the colors in the admin panel and not in the css? 

I found one bug; If you press cancel on the title insert popup, you get [spoilernull][/spoiler]

dragomano:
> Any reason you went for a config of the colors in the admin panel and not in the css?

1.  For the convenience of users - not everyone likes to edit CSS files directly :)
2.  It is a demo of the capabilities of the color type in the theme settings.

However, if this setting is unnecessary, I can remove it :)

BrickOzp:
Found another thing.
If you modify a post with a spoiler with a title, the title is not shown in the spoiler tag.

dragomano:
I cannot fix this bug yet.

BrickOzp:
It seems that the sceditor does not find the < details > tag.

Edit: It seems to be related to the nested tags

dragomano:
Apparently, we will have to use a hint in the jquery.sceditor.smf.js file  :)
>This fixes GH Bug #2845
			 * As SMF allows "[quote=text]message[/quote]" it is lost during sceditor when it converts bbc to html and then html back to bbc code.  The simplest method is to tell sceditor that this is a "author", which is how the bbc parser treats it in SMF.  This will cause all bbc to be updated to "[quote author=text]message[/quote]".

BrickOzp:
It's somewhat better. If I modify a spoiler with a title it shows as
[spoiler=title]titleContent[/spoiler]

and without title:
[spoiler]SpoilerContent[/spoiler]

It seems that the content param contains both the title and content.

I managed to solve it by changing to this code in jquery.sceditor.smf.js

let titleLength;
if (element.attr('data-title')) {
    title = '=' + element.attr('data-title').php_unhtmlspecialchars();
    titleLength = title.length - 1;
} else
    titleLength = $(element).children('summary:first').text().length;

return '[spoiler' + title + ']' + content.substring(titleLength) + '[/spoiler]';

dragomano:
Did you update all files? I cannot repeat the bug you found.

BrickOzp:
Yes, you are right. I reset my branch and reapplied your commit, and now it works.
Great job.

BrickOzp (approved):

Sesquipedalian:
Sorry, but this will need be a mod, not a feature.

Fortunately, new BBCodes can be added using nothing but integration hooks in 2.1, so it should be simple to convert your code to do that.

PR #6480, opened 2021-01-25
Fix error when using slash in searches

### LINKED ISSUE #6413: From forum: preg_filter(): Unknown modifier '\'
#### Description
"Another new error appeared!  60+ of these all at once:"
> Type of error: General
>  2: preg_filter(): Unknown modifier '\'
>  https://nfggames.com/forum2/index.php?amp;action=search2
>  /var/www/web/forum2/Sources/Search.php (Line 2319)

Reported by two different users on the forum:
https://www.simplemachines.org/community/index.php?topic=575999.msg4075611#msg4075611

BrickOzp:
If using a slash (/) when searching an error was
reported in the error log.
Changed delimiters to fix it.

Fixes #6413

### DIFF
--- a/Sources/Search.php
+++ b/Sources/Search.php
@@ -2315,7 +2315,7 @@ function searchSort($a, $b)
 function highlight($text, array $words)
 {
 	$words = implode('|', array_map('preg_quote', $words));
-	$highlighted = preg_filter('/' . $words . '/i', '<span class="highlight">$0</span>', $text);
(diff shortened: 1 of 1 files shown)


m4z:
Interesting, I didn't know mismatched [delimiters](https://www.php.net/manual/en/regexp.reference.delimiters.php) were possible.

BrickOzp:
Yes, the problem is preg_quote does not handle slash, so this could either be solved by quoting slashes our self or to change the delimiter to something handled by preg_quote

PR #6486, opened 2021-01-29
Set user id after registration

### LINKED ISSUE #6237: Profile Edits - Fields added during registration listed as performed by "deleted member"
#### Description
Somethng's not being initialized correctly in log_actions at registration time.

image

### Steps to reproduce
1. Register a new user
2. Look at the Profile Edit history 

- Database Type:  MySQL
- Database Version:  5.7
- PHP Version:  7.3

BrickOzp:
During the registration of a new user, custom profile fields
could be configured to be set during the registration.
When changed these custom profile field edits are added to the
profile edits log. But since the user id is not set at this time,
the changed by user ended up as 0.
Solve this by setting the user id after it has been created.

Fixes #6237

### DIFF
--- a/Sources/Register.php
+++ b/Sources/Register.php
@@ -251,7 +251,7 @@ function Register($reg_errors = array())
 function Register2()
 {
 	global $txt, $modSettings, $context, $sourcedir;
-	global $smcFunc, $maintenance;
+	global $smcFunc, $maintenance, $user_info;
 
 	checkSession();
 	validateToken('register');
@@ -560,6 +560,8 @@ function Register2()
 	{
 		require_once($sourcedir . '/Profile.php');
 		require_once($sourcedir . '/Profile-Modify.php');
+		// Set user_info id, it is used by log actions
+		$user_info['id'] = $memberID;
 		makeCustomFieldChanges($memberID, 'register');
 	}
 


frandominguezl (approved):

jdarwood007:
Setting the user_info['id'] can have unexpected consequences if done outside of the login/session process.  Its best to avoid changing it the way you did.

frandominguezl:
What are those consequences if I may ask?

jdarwood007:
Well in general, SMF assumes $user_info['is_guest'] = $user_info['id'] == 0;  But some places still use the later logic.  So if a ID exists, there are places in the code that may assume this is a logged in user.  $user_info should always only be filled by the proper process.

We had the alerts system doing weird things because it was manipulating the $user_info of a logged in user.  Which can be dangerous because it is essentially impersonating another user at that point under the correct set of conditions, possibly could lead to an account hijack.  Not that I know of any code that would, but it left the door open for it if somebody found a chain-able set of events for that to happen.

For this I would add the logic to the makeCustomField to when the second param is register, to allow the logged ID to be the new members ID via the $memberID.  That should be as simple as:
```
					'applicator' => $user_info['id'],
```

to
```
					'applicator' => empty($user_info['id']) && $area === 'register' ? $memID : $user_info['id'],

PR #6494, opened 2021-02-03
fix: Limit index size

### LINKED ISSUE #6163: quanda table lngfile key too long

MissAllSunday:
Fixes #6163

### DIFF
--- a/other/install_2-1_mysql.sql
+++ b/other/install_2-1_mysql.sql
@@ -986,7 +986,7 @@ CREATE TABLE {$db_prefix}qanda (
 	question VARCHAR(255) NOT NULL DEFAULT '',
 	answers TEXT NOT NULL,
 	PRIMARY KEY (id_question),
-	INDEX idx_lngfile (lngfile)
+	INDEX idx_lngfile (lngfile(249))
 ) ENGINE={$engine};
 
 #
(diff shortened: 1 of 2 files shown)


sbulen:
There are actually 3 indexes that are too small, this only addresses one of them.

And the DB doesn't actually make use of mb4 until we look into this PR: #6409 

Until then, it's mb3 data stuffed into mb4-capable tables...

sbulen:
And 249 is too big, it needs to be 191, for reasons explained over on #6409 also.

sbulen:
I clearly worry too much...  I am also concerned that the string 'utf8' != the string 'utf8mb4'...

There are several places throughout our code - still old 2.0 code that supports ISO-8859-1 - where the logic treats the contents as _EITHER_ utf8 _OR_ ISO-8859-1...  I.e., a utf8mb4 DB may actually get treated as a latin1 DB, because it's not 'utf8'.  Hence the importance of ripping out all the old "otherwise its ISO-8859-1" logic.

sbulen:
One last change needed - When we change the DB we also add a corrective step to the upgrader.  In this instance, dropping the old index & adding the new one.

An example:
https://github.com/SimpleMachines/SMF2.1/blob/a8f4d6af6465d5990d4df4bd0cb5aed242025a7c/other/upgrade_2-1_mysql.sql#L567

This allows the upgrader to be rerun to fix issues, bring the DB into conformity.

PR #6495, opened 2021-02-04
fix: Escape user input in the editor

live627:
The following commands are affected:

- insert image
- insert link
- insert email

These three commands have a dropdown and collect input. However, said inputs were unfiltered.

Also removes some redundant input prompts to make the UX a bit more uniform.


live627:
Any comments or concerns before I merge this one?

MissAllSunday (approved):

PR #6498, opened 2021-02-05
Translatable fields

### LINKED ISSUE #6435: Custom Profile Fields can not be translated
I had already reported this once https://github.com/SimpleMachines/SMF2.1/issues/4381.

I also remember that it was discussed. In a multilingual forum it should be possible that the input of user-defined fields can also be translated.

There are already language strings, but they do not work
https://github.com/SimpleMachines/SMF2.1/blob/35a6bd91253ca655f2c3bad62033a9998329f2db/Themes/default/languages/index.english.php#L222-L225
https://github.com/SimpleMachines/SMF2.1/blob/35a6bd91253ca655f2c3bad62033a9998329f2db/Themes/default/languages/index.english.php#L231-L232

custom_profile_fields

User profile settings
image

Statistics
custom_profile_fields_3

There should be a possibility to have a separate field for each installed language pack.

custom_profile_fields_2

MissAllSunday:
Starting with 2.1 we offer 4 default custom fields (location, gender, skype and ICQ) currently we hardcoded both names and descriptions for all of them.

We should provide those default cusotm fields to be translated into whatever language the admin has installed. This can be achieved by replacing those hardcoded values with txt tokens. This txt tokens will be dynamically replaced with their corresponding $txt values whenever a cusotm field title or description is shown to the user.

Incidentally, admins can use the same txt token feature to provide multi-language support for their custom fields.

Fixes #6435


Themes/default/languages/Help.english.php
```diff
@@ -625,6 +625,10 @@
 
 $helptxt['field_reg_require'] = 'If this field is required during registration, it will also be required on profile changes.';
 
+$helptxt['translatable_fields'] = 'If you wish to provide multi-language support for your custom profile fields you can use translatable tokens on both the name and description fields.<br>
+Anything enclosed with braces {} will be considered as a token and will be matched against a $txt string, for example, the token: {gender} will be replaced by the value of $txt[\'gender\']<br>
+It is <strong>very important</strong> to note you need to add the matched $txt string to <strong>ALL</strong> currently installed languages, it is recommended to put them on index.language.php for the translation to be available across your forum.';
```
Sesquipedalian, line 630: This is a perfect opportunity to direct the admin to the built-in language editor, so that admins who are reluctant to edit files directly can do the task using a nice, friendly UI instead.

```suggestion
$helptxt['translatable_fields'] = 'If you wish to provide multi-language support for your custom profile fields you can use translatable tokens on both the name and description fields.<br><br>
Anything enclosed with braces <span class="monospace">{}</span> will be considered as a token and will be matched against a $txt string. For example, the token <span class="monospace">{gender}</span> will be replaced by the value of $txt[\'gender\']<br><br>
It is <strong>very important</strong> to add the matched $txt string to <strong>ALL</strong> currently installed languages. To do this, go to the <a href="%1$s?action=admin;area=languages;sa=edit;%3$s=%2$s" class="bbc_link" target="_blank" rel="noopener">Edit Languages</a> page and complete the following steps for each installed language:
<ol class="bbc_list" style="list-style-type: decimal">
	<li>Click the "Edit" button next to the language name.</li>
	<li>In the "Select entries to edit" menu, choose "Modifications".</li>
	<li>Scroll to bottom and click the "Add Another Item" button.</li>
	<li>Follow the prompts to add your new string.</li>
</ol>';
```

Sesquipedalian (changes requested):
Works very nicely! I just have one suggestion regarding the $helptxt string.

Sesquipedalian:
I think it would be a really good idea to include this before the final release of 2.1. It works well, it will make a big difference for our International language support, and the changes involved are quite focused and specific.

Sesquipedalian:
I added my suggested change to `$helptxt['translatable_fields']` and merged this.

PR #6500, opened 2021-02-06
ci: add github workflow/action and remove travis ci

live627:
Depends on https://github.com/SimpleMachines/BuildTools/pull/12

JavaScript linting can be left for the future since our codebase isn't up to snuff yet, with different styles.

This will serve as the first phase of CI.

The version checking script will become problematic as we go to GA/patch only, because then not all files get version updates. See2.0.18.


m4z:
Since #6501 is just ditching PHP 5.4, what's the plan re `legacy-php.yml`? Remove the test once it begins to fail, or still fix 5.4 errors despite not officially supporting it? Or can the check be made "optional"?

live627:
I'll bump it  to 5.6.

live627:
Actually, I got rid of it because of php 7 code. What happened is that overtrue/phplint uses Symfony components which use php 7 code. I'm not ggoing to put in the effort to add something else. Someone else can if they wish. I'm not going to bother with php 5 for cd any longer.

Issue #5594, opened 2019-04-11
Unresponsive tables


SychO9:
a reminder to look at a few more unresponsive tables in the admin section
* hooks
* langs
https://www.simplemachines.org/community/index.php?topic=567096.msg4015563;topicseen#new

frandominguezl:
@sbulen Please, "Has Fix" label on this one O:)

sbulen:
Is this still outstanding?  "Has fix" labels came & went...  Subsequent PRs affecting this have come & gone as well...

DiegoAndresCortes:
I could only find the hooks table as a legitimate issue and since you don't want to play around too much with that info on a phone I decided to hide some elements.
But other than that I couldn't find any other essential sections that are not responsive.
E.g. The languages page is fine for me.

BrickOzp:
This could be closed now

PR #6502, opened 2021-02-06
Some small theme fixes

### LINKED ISSUE #5818: Child boards with new posts: presentation and PHP
Ok, two separate things here. First one is the current presentation for the child board new post indicators. They're a bit washed out and bleh, IMO. Would be hard to read for some people too, if they were new to SMF. The current ones are shown in the top row of the first attached image.

My suggestion is shown in the second row of the first image. I think this is far more legible, and a much better aesthetic match for the board icons at the left side, and better for positioning. It can also be done with a bit less code.

```
/* The "new" button */
.new_posts {
	display: inline-block;
	position: relative;
	top: -2px;
	padding: 0 4px;
	background: linear-gradient(#f97b00,#884d00);
	color: #fff;
	font: 9px/15px verdana, sans-serif;
	font-weight: 700;
	border-radius: 2px;
	opacity: 0.8;
}
.new_posts:hover, .new_posts:focus {
	text-decoration: none;
	opacity: 1;
}
``` 
Note that I removed the transition, as I think in this instance it is not worth the bytes. I also added a minor tweak to BoardIndex.template.php and MessageIndex.template.php. Since you are always going to want a slight gap between the "New" thingy and the board name, I think this is a situation where it makes sense to add a non-breaking space to the HTML. Keeps the CSS simpler, works for any language, and even the most persnickety themer is unlikely to mind.

```
			if (!$child['is_redirect'])
				$child['link'] = '' . ($child['new'] ? '<a href="' . $scripturl . '?action=unread;board=' . $child['id'] . '" title="' . $txt['new_posts'] . ' (' . $txt['board_topics'] . ': ' . comma_format($child['topics']) . ', ' . $txt['posts'] . ': ' . comma_format($child['posts']) . ')" class="new_posts">' . $txt['new'] . '</a>&nbsp;' : '') . '<a href="' . $child['href'] . '" ' . ($child['new'] ? 'class="board_new_posts" ' : '') . 'title="' . ($child['new'] ? $txt['new_posts'] : $txt['old_posts']) . ' (' . $txt['board_topics'] . ': ' . comma_format($child['topics']) . ', ' . $txt['posts'] . ': ' . comma_format($child['posts']) . ')">' . $child['name'] . '</a>';

```
2_1_child_boards_new

-----------------------------------------------------------------------
So that's one thing. The other thing is the content of the title attributes.

When the child board has new posts the "New" thingy links to ?action=unread;board=** while the board name obviously just links to ?board=** yet they both have the same title content, which implies they both do the same thing. I think this is a GUI fail, and it would be better if the child board link always had the total post and topics, and left the "New" count to the "New" thingy title. That way they're not claiming to do the same thing and trying to trick newbies. It'd also save you a ternary in the template, which is not such a bad thing either,
  
2_1_child_boards_PHP

### LINKED ISSUE #6252: Non functioning and probably superfluous code to set height of code blocks
I just came across this in **theme.js**:
```javascript
if (is_webkit && $(tag).height() < 20)
			$(tag).css({height: ($(tag).height + 20) + 'px'});
```
This doesn't do anything, because `$(tag).height` is a function in jQuery.  So `$(tag).height + 20) + 'px'}` never evaluates to anything useful. I could fix it:  `$(tag).height() + 20) + 'px'}` 
And you can review the effect of this fix here: https://jsfiddle.net/KolyaKorruptis/40otp8jr/6/

But in reality Safari of course doesn't need help dealing with _overflow: auto blocks_ these days.  (I actually tested this.) 
Now if browser-support is determined by jQuery [as MissAllSunday said](https://github.com/SimpleMachines/SMF2.1/issues/6245#issuecomment-687156251), then [we're talking about the current Safari version - 1](https://jquery.com/upgrade-guide/3.0/#browser-support). Consequently this code could be removed. But I'd like to hear your opinions.

As a sidenote, I noticed that Chrome gets caught in "is_webkit".

### LINKED ISSUE #6263: responsive.css: .lpr_border class
This should be killed before Final, IMHO. It's a stupid little presentational class (the worst sort :P ) that is only called in two places in the markup, and even then only on screens of 720px or narrower.

Instead of this:
```
	.lpr_border {
		border-top: 1px solid #ddd;
	}
```

...this would be perfectly fine:
```
	.up_contain .lastpost {
		border-top: 1px solid #ddd;
	}
```

Then the templates can just be:
```
			// Show the last post if there is one.
			if(!empty($board['last_post']['id']))
				echo'
					<div class="lastpost">
```

DiegoAndresCortes:
🌵 
Fixes #6263
Fixes #6252 but requires discussion about removing legacy code so for now we can simply fix it
Fixes #5818 but doesn't address the second "issue" which is more like a discussion

### DIFF
--- a/Themes/default/BoardIndex.template.php
+++ b/Themes/default/BoardIndex.template.php
@@ -254,7 +254,7 @@ function template_bi_board_children($board)
 		foreach ($board['children'] as $child)
 		{
 			if (!$child['is_redirect'])
-				$child['link'] = '' . ($child['new'] ? '<a href="' . $scripturl . '?action=unread;board=' . $child['id'] . '" title="' . $txt['new_posts'] . ' (' . $txt['board_topics'] . ': ' . comma_format($child['topics']) . ', ' . $txt['posts'] . ': ' . comma_format($child['posts']) . ')" class="new_posts">' . $txt['new'] . '</a>' : '') . '<a href="' . $child['href'] . '" ' . ($child['new'] ? 'class="board_new_posts" ' : '') . 'title="' . ($child['new'] ? $txt['new_posts'] : $txt['old_posts']) . ' (' . $txt['board_topics'] . ': ' . comma_format($child['topics']) . ', ' . $txt['posts'] . ': ' . comma_format($child['posts']) . ')">' . $child['name'] . '</a>';
+				$child['link'] = '' . ($child['new'] ? '<a href="' . $scripturl . '?action=unread;board=' . $child['id'] . '" title="' . $txt['new_posts'] . ' (' . $txt['board_topics'] . ': ' . comma_format($child['topics']) . ', ' . $txt['posts'] . ': ' . comma_format($child['posts']) . ')" class="new_posts">' . $txt['new'] . '</a> ' : '') . '<a href="' . $child['href'] . '" ' . ($child['new'] ? 'class="board_new_posts" ' : '') . 'title="' . ($child['new'] ? $txt['new_posts'] : $txt['old_posts']) . ' (' . $txt['board_topics'] . ': ' . comma_format($child['topics']) . ', ' . $txt['posts'] . ': ' . comma_format($child['posts']) . ')">' . $child['name'] . '</a>';
 			else
 				$child['link'] = '<a href="' . $child['href'] . '" title="' . comma_format($child['posts']) . ' ' . $txt['redirects'] . ' - ' . $child['short_description'] . '">' . $child['name'] . '</a>';
 
--- a/Themes/default/BoardIndex.template.php
+++ b/Themes/default/BoardIndex.template.php
@@ -108,7 +108,7 @@ function template_main()
 
 			// Show the last post if there is one.
 			echo'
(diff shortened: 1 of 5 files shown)


DiegoAndresCortes:
I also have no clue on what else is broken regarding #5594
I removed some elements from the hooks table as realistically is unlikely that an admin would play around with disabling or removing hooks from the phone, but at least have it looking alright in the phone.
The langs section looks fine to me on a smaller screen, and didn't find any additional tables that look bad.

Also this pr is just straight fixes from the issues, doesn't need much testing at all.

MissAllSunday:
There should be a plugin avilable for .editorConfig support for VSCode: https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig

live627:
@DiegoAndresCortes you planning on leaving those discussions you alluded to for a future version?

live627 (approved):

DiegoAndresCortes:
Just the one about responsive tables which was not mentioned, thanks!

PR #6506, opened 2021-02-07
Drop Subs-Password

albertlast:
Password function was important for php 5.4 env,
since we raised the support to 5.6 we cran dop this logic.

https://www.php.net/manual/en/function.password-hash.php


frandominguezl (approved):

Sources/Subs-Auth.php
```diff
@@ -872,9 +872,7 @@ function smf_setcookie($name, $value = '', $expire = 0, $path = '', $domain = ''
  */
 function hash_password($username, $password, $cost = null)
 {
-	global $sourcedir, $smcFunc, $modSettings;
-	if (!function_exists('password_hash'))
-		require_once($sourcedir . '/Subs-Password.php');
+	global  $smcFunc, $modSettings;
```
live627:
```suggestion
	global $smcFunc, $modSettings;
```

live627 (approved):

PR #6508, opened 2021-02-08
Fix token problem when adding new admins

### LINKED ISSUE #6491: Add Member to Admin Group
#### Description
If you want to add a member to the admin group, this error message appears after logging in.

Token verification failed. Please go back and try again.

### Steps to reproduce
1.  Go to Admin Center =>    Manage Membergroups => Edit membergroups
 

Version/Git revision: Last Commit RC3
Database Type: MySQL
Database Version: 5.7.33
PHP Version: 7.4.14!

https://github.com/SimpleMachines/SMF2.1/blob/a7c4e1bb7f6cb245b623cc67821e8691eaf873e1/Sources/Groups.php#L285-L293

I have changed the above code to 

```
        // Demand an admin password before adding new admins -- every time, no matter what.
        if ($context['group']['id'] == 1)
            validateSession('admin', true);

        checkSession();
        validateToken('mod-mgm');
```

Then it works. The error message no longer appears and the selected member is added to the admin group.

BrickOzp:
Validate token after admin password has been entered,
when adding new admins. This will fix the problem that
the token gets unvalid.

Fixes #6491

### DIFF
--- a/Sources/Groups.php
+++ b/Sources/Groups.php
@@ -285,12 +285,12 @@ function MembergroupMembers()
 	// Must be adding new members to the group...
 	elseif (isset($_REQUEST['add']) && (!empty($_REQUEST['toAdd']) || !empty($_REQUEST['member_add'])) && $context['group']['assignable'])
 	{
-		checkSession();
-		validateToken('mod-mgm');
-
 		// Demand an admin password before adding new admins -- every time, no matter what.
 		if ($context['group']['id'] == 1)
 			validateSession('admin', true);
+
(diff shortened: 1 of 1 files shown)


BrickOzp:
The problem is that verifyToken unsets the token after it is checked. And since it is not transferred to the password form it gets lost.
verifySession will halt the execution when the password is required and the token will only be verified after the admin pass is correct.

live627 (approved):
Good stuff. Thanks.

PR #6509, opened 2021-02-09
fix: validate the ip before attempting to  convert it

live627:
https://www.simplemachines.org/community/index.php?topic=576679.0

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -5017,13 +5017,13 @@ function template_upgrade_complete()
 /**
  * Convert MySQL (var)char ip col to binary
  *
+ * newCol needs to be a varbinary(16) null able field
+ *
  * @param string $targetTable The table to perform the operation on
  * @param string $oldCol The old column to gather data from
  * @param string $newCol The new column to put data in
  * @param int $limit The amount of entries to handle at once.
  * @param int $setSize The amount of entries after which to update the database.
- *
- * newCol needs to be a varbinary(16) null able field
  * @return bool
  */
 function MySQLConvertOldIp($targetTable, $oldCol, $newCol, $limit = 50000, $setSize = 100)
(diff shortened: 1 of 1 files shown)


albertlast:
When i read the code correct, you keep malformed ip addresses in the db instead of deleting them?

sbulen:
~~I think albertlast is right here.  We need something to catch & deal with the invalid values.  We have seen plenty of them...~~

~~So...~~
~~- I think the invalid values need to be dealt with - null?~~
~~- I think we should accept valid ipv6 as well.  No need to throw them out.  I think lots of folks are running SleePy's ipv6 mod.~~

Works fine.  See note below.

w3bdesign (approved):

other/upgrade.php
```diff
@@ -5098,7 +5098,7 @@ function MySQLConvertOldIp($targetTable, $oldCol, $newCol, $limit = 50000, $setS
 		{
 			$arIp[$i] = trim($arIp[$i]);
 
-			if (empty($arIp[$i]))
+			if (!filter_var($arIp[$i], FILTER_VALIDATE_IP, FILTER_FLAG_IPV4))
```
sbulen: I think we should allow for IPV4 & IPV6 here

sbulen (changes requested):
I think we will have a lot of folks converting from ipv6, so we should allow for ipv4 & ipv6.

Tests fine otherwise - properly filters out any non-IP.

sbulen:
@live627 - your thoughts on adding ipv6 here?

other/upgrade.php
live627:
```suggestion
			if (!filter_var($arIp[$i], FILTER_VALIDATE_IP, FILTER_FLAG_IPV4 | FILTER_FLAG_IPV6))
```

live627: why the newline giithub why u ignore our conffig

sbulen (approved):

w3bdesign (approved):

PR #6514, opened 2021-02-11
Sometimes find 0s in mm or dd

### LINKED ISSUE #6316: 2.1 RC3 Upgrader fails on invalid birthdays
#### Description
My site is originally from Xsorbit (ie SMF 1.something), and has been gradually updated all the way up to 2.0.17
Today, whilst testing upgrade to RC3 I ran in to problems with select members. 

The upgrader fails if there are birthdays ending in -00-00. 
I'm not sure how, or from what time in the past 15 years - but I had a handful of members with birthdays like that. 
Easy to fix manually by going in the DB and just changing it to -01-01, 
but could we do something to fix this in the upgrader in case others run in to this?

### Steps to reproduce
1. Have 2.0 with a member whose birthday is for example 1984-00-00
2. Run 2.1 RC3 upgrade
3. See results

I still have a clone of the original DB, so if further details are wanted I can actually redo the upgrade and gather information.

sbulen:
Fixes #6316 

Need to account for some old/migrated DBs with year-only birthdates.

Tests fine - it will now set mm & dd to 1.  

I could not replicate the erroneous data in pg in 2.0, but replicated the fix here anyway & confirmed upgrader still runs fine.

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -1234,6 +1234,9 @@ function UpgradeOptions()
 	if (empty($_POST['upcont']))
 		return false;
 
+	// We cannot execute this step in strict mode - strict mode data fixes are not applied yet
+	setSqlMode(false);
+
 	// Firstly, if they're enabling SM stat collection just do it.
 	if (!empty($_POST['stats']) && substr($boardurl, 0, 16) != 'http://localhost' && empty($modSettings['allow_sm_stats']) && empty($modSettings['enable_sm_stats']))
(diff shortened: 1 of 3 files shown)


sbulen:
Updating karma settings, if requested, also caused issues.  If historical data had 00 in mm or dd, even the Upgrade Options step failed hard, long before it even got to the database changes...

MissAllSunday (approved):

PR #6519, opened 2021-02-12
Fix edit of secondary gravatar email

### LINKED ISSUE #6182: gravatar not taking email address
#### Description
use my gravatar option does not select gravatar
### Steps to reproduce
enable gravatars for forum users is enabled, even tried to disable allowing storing an extra email address for gravatars
2. go in profile and selected option to use my gravatar, only the default avatar is showing.

- Version/Git revision: HEAD is now at 8104e4328 Merge pull request #6178 from SychO9/fix_disabled_files
- Database Type: mariadb
- Database Version:
- PHP Version: 7.1.33

this is a fresh install from github, cache was not enabled.

also checked "Force Gravatars to be used instead of normal avatars?" and that did not force a gravatar to show.

BrickOzp:
If setting "Force Gravatars to be used instead of normal avatars?"
was enabled, the extra email for gravatar field was not populated
and could not be changed, when modifying the profile.

Fixes #6182

### DIFF
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -3220,7 +3220,7 @@ function profileLoadAvatarData()
 		$context['member']['avatar'] += array(
 			'choice' => 'gravatar',
 			'server_pic' => 'blank.png',
-			'external' => $cur_profile['avatar'] == 'gravatar://' || empty($modSettings['gravatarAllowExtraEmail']) || !empty($modSettings['gravatarOverride']) ? $cur_profile['email_address'] : substr($cur_profile['avatar'], 11)
+			'external' => $cur_profile['avatar'] == 'gravatar://' || empty($modSettings['gravatarAllowExtraEmail']) || (!empty($modSettings['gravatarOverride']) && substr($cur_profile['avatar'], 0, 11) != 'gravatar://') ? $cur_profile['email_address'] : substr($cur_profile['avatar'], 11)
 		);
 		$context['member']['avatar']['href'] = get_gravatar_url($context['member']['avatar']['external']);
 	}


BrickOzp:
This doesn't work if an extra email address is configured. And then disable the setting "Enable Gravatars for forum users?" while "Force Gravatars to be used instead of normal avatars?" is enabled. While modifying the profile the extra email is changed to the gravatar url.

Why is even the gravatars used and the setting available if gravatars are disabled?

BrickOzp:
Seems really broken.
If "Enable Gravatars for forum users?" is disabled but "Force Gravatars to be used instead of normal avatars?" is enabled
it is not possible to change the avatar to other options, even if they are shown in the profile editor.

BrickOzp:
Now it works as expected.

Themes/default/languages/Admin.english.php
```diff
@@ -485,12 +485,12 @@
 $txt['gravatar_maxR'] = 'R rated (Restricted)';
 $txt['gravatar_maxX'] = 'X rated (Explicit)';
 $txt['gravatarDefault'] = 'Default image to show when an email address has no matching Gravatar ';
-$txt['gravatar_mm'] = 'A simple, cartoon-style silhouetted outline of a person';
-$txt['gravatar_identicon'] = 'A geometric pattern based on an email hash';
-$txt['gravatar_monsterid'] = 'A generated \'monster\' with different colors, faces, etc';
+$txt['gravatar_mm'] = 'Static, cartoon-style silhouetted outline of a person';
+$txt['gravatar_identicon'] = 'Generated geometric pattern';
+$txt['gravatar_monsterid'] = 'Generated \'monster\' with different colors, faces, etc';
```
live627: More clearer English?

BrickOzp: The reporter of the issue was confused about which gravatars was generated for every user and which ones only showed a static image, so thought it could be made more clear. But I am not a native english speaker, so feel free to propose changes if anything feels wrong.

BrickOzp: I can remove the change of these strings as well. They are not the important part of this PR.

live627: Yes please. I'm thinking we can just keep the first commit, since I definitely cannot come up with a better way to reword those.

BrickOzp: Ok, done

PR #6523, opened 2021-02-15
Add CrowdIn integration

sbulen:
This PR will setup a GitHub action to support CrowdIn integration.

This is a one-way integration that will push the current english sources (from the languages folder + agreement.txt) over to CrowdIn.  The source files in CrowdIn, under the "SMF_2-1" folder will be overwritten.  CrowdIn is smart enough to keep all translations where common keys & strings exist.

This workflow can be triggered in one of two ways:
 - When a release is published, e.g., RC4
 - On request - & ONLY with agreement from the localization team

What needs to be done after this PR is merged:
 - 2 GitHub secrets must be setup for CROWDIN_PROJECT_ID, and CROWDIN_API_TOKEN.  For the CI token, either an API token or a personal token will work.  The CrowdIn token & project ID should reference the primary Simple Machines project in CrowdIn.  I don't have the authority to add the secrets.
 - Since crowdin.yml is in the root directory, I imagine we will need to exclude it from some build tools, e.g., for building install & upgrade packages.  It is possible to move crowdin.yml elsewhere, and would need to reflect this by adding a config: clause in the workflow to point to the new location.


.github/workflows/crowdin_wf.yml
```diff
@@ -0,0 +1,57 @@
...
+        # The title of the new pull request
+        pull_request_title: 'New Crowdin translations by Github Action'
+        # To add labels for created pull request
+        pull_request_labels: 'Language'
+
+      env:
+        # Use GitHub Secrets here, so sensitive stuff isn't in the publicly visible workflow
+        # Note that the CrowdIn API token will accept either a user's personal CrowdIn token
+        # or an "API" token, generated via the CrowdIn console client using the "crowdin init" command.
+        GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
```
live627: None off this is needed because we are only pushing to Crowdin.

```suggestion
      env:
        # Use GitHub Secrets here, so sensitive stuff isn't in the publicly visible workflow
        # Note that the CrowdIn API token will accept either a user's personal CrowdIn token
        # or an "API" token, generated via the CrowdIn console client using the "crowdin init" command.
```

.github/workflows/crowdin_wf.yml
```diff
@@ -0,0 +1,57 @@
...
+        CROWDIN_PROJECT_ID: ${{ secrets.CI_PROJECT }}
+        CROWDIN_API_TOKEN: ${{ secrets.CI_API_TOKEN }}
```
live627: Your tokens should be more descriptive.
```suggestion
        CROWDIN_PROJECT_ID: ${{ secrets.CROWDIN_PROJECT_ID }}
        CROWDIN_API_TOKEN: ${{ secrets.CROWDIN_API_TOKEN }}
```

crowdin.yml
```diff
@@ -0,0 +1,51 @@
...
+  #
+  "source" : "agreement.txt",
+
+  #
+  # Translated agreements include the language
+  #
+  "translation" : "agreement.%language%.txt"
+ }
+
+]
```
live627: I made this to look more yaml and  less like json
```suggestion
#
# Your Crowdin credentials
#  - The _env versions replace the non-_env versions when using a public repository for security
#  - These environment variables must be defined in the env: section in your GitHub workflow
#
project_id_env: CROWDIN_PROJECT_ID
api_token_env: CROWDIN_API_TOKEN
base_path: "."
base_url: "https://api.crowdin.com"

#
# Whether to use the existing file system hierarchy or to flatten it out
#  - true or false, SMF needs this to be true, since our hierarchy isn't flat (agreement.txt vs others)
#
preserve_hierarchy: true

#
# Files configuration
#
files:
  -
   #
   # Source files filter
   #  - Grab all the english files in the languages folder
   #
   source: /Themes/default/languages/*.english.php

   #
   # Where translations will be placed
   #  - Note this honors CrowdIn's "Language Mapping"
   #  - Need to remove the '.english' from the filename on export otherwise things end up like "Admin.english.dutch.php"
   #
   translation: /Themes/default/languages/%file_name%.%language%.php
   translation_replace:
    .english: ""

  -
   #
   # Source files filter
   #  - The english agreement has no language in the name
   #
   source: agreement.txt

   #
   # Translated agreements include the language
   #
   translation: agreement.%language%.txt
```

.github/workflows/crowdin_wf.yml
sbulen: Good input.

crowdin.yml
sbulen: The JSON-ish structure is required here, because we have two sets of sources & translations.  This is because agreement.txt & the rest behave very differently, as described in the comments above.

sbulen: I started with their templates...  The BIGGEST thing that bugs me is the fact they used spaces, not tabs...

.github/workflows/crowdin_wf.yml
sbulen: As noted in the internal discussion, I think we should seriously consider pulling translations into GH....  I added a lot of comments based on many tests, & where their documentation sucks.  

As a first cut, though, I think these things can go.

I'll see what I can do on spaces vs tabs, too...   Again, I used their templates, & they were weird to work with...

crowdin.yml
sbulen: FYI - this file is the output of their config tool....

live627: in the end it doesn't really matter because both versions are valid and they both output the  exact same results (json)

live627:
@liroyvh you are the only one with the ability to add secrets (admin privilege)

frandominguezl (approved):
LGTM

w3bdesign (approved):

sbulen:
Had to remove the wildcards for getting language files to translate.

Otherwise files were in the format: Admin.english.dutch.php, instead of Admin.dutch.php.

jdarwood007:
@CoreISP

LiroyvH:
Secrets added.

jdarwood007:
One final note, does crowdin.yml have to go at the project root or can we put it up in the .github folder to keep our root clean?

sbulen:
We can move it somewhere else.  The problem with using the .github folder is that enables it as an action, but it's not an action.

jdarwood007:
I thought that only would happen if we put it in the .github/workflows not just the .github?
Can we at least make it a .crowdin.yml then?

sbulen:
I pushed it up to the .github folder & it tested OK.  What I had tested earlier was having it in the workflows folder.

jdarwood007 (approved):

PR #6527, opened 2021-02-19
Fix drafts tables for mobiles

BrickOzp:


BrickOzp:
Before fix
drafts org

After fix
drafts fixed

MissAllSunday (approved):

PR #6528, opened 2021-02-21
Fix behavior of [code=xyz] in the editor

live627:
See https://www.simplemachines.org/community/index.php?topic=576369.0

### DIFF
--- a/Themes/default/css/jquery.sceditor.default.css
+++ b/Themes/default/css/jquery.sceditor.default.css
@@ -49,6 +49,9 @@ code::before {
 	top: -1.35em;
 	left: 0;
 }
+code[data-title]::before {
+	content: 'Code: (' attr(data-title) ')';
+}
 code {
 	margin-top: 1.5em;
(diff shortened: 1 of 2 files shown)


Themes/default/css/jquery.sceditor.default.css
```diff
@@ -49,6 +49,9 @@ code::before {
 	top: -1.35em;
 	left: 0;
 }
+code[data-title]::before {
+	content: 'Code: (' attr(data-title) ')';
```
MissAllSunday, line 53: Might be an issue with localization, perhaps it will be better to move  the "Code:" part to a js var filled with its corresponding translation.

MissAllSunday (changes requested):

Themes/default/css/jquery.sceditor.default.css
live627: I will address that localization  problem in the update to SCE v3

PR #6533, opened 2021-02-23
Make custom fields selectable for buddy list

### LINKED ISSUE #2878: custom profile fields showing on buddy list in profile
IMO the custom profile fields should not show on the buddy list unless they also show on the members list. currently all the fields show even if they are not on the members list. this can cause a lot of info to be loaded on that page that may not be important to see.

BrickOzp:
At the moment, all custom fields is shown on the buddylist.
Make it customizable, like for the members list.

Fixes #2878

### DIFF
--- a/Sources/ManageSettings.php
+++ b/Sources/ManageSettings.php
@@ -1788,7 +1788,7 @@ function EditCustomProfiles()
 		$request = $smcFunc['db_query']('', '
 			SELECT
 				id_field, col_name, field_name, field_desc, field_type, field_order, field_length, field_options,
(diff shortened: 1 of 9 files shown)


albertlast:
would it be worth to add/extend a db index?

jdarwood007:
@Sesquipedalian is this something we want for 3.0?  If so, one of us will need to fix the installer/upgrader logic

Sesquipedalian:
It's a good idea, but not during the current milestone. Please create an issue for it, marked for milestone Alpha 5.

jdarwood007:
This will be handled in #9008 for 3.0

PR #6537, opened 2021-02-25
Months fix needed for CrowdIn

sbulen:
This will address the month index issue in CrowdIn.


sbulen:
This is the only place in the language files that I could find affected by this.

PR #6551, opened 2021-03-02
Avoid undefined index error when issue warning

BrickOzp:
If a warning was issued without a reason,
an undefined index error was added to the error log.
Fix that by checking if the text string is defined
before accessing it.

### DIFF
--- a/Themes/default/Profile.template.php
+++ b/Themes/default/Profile.template.php
@@ -2701,7 +2701,7 @@ function template_error_message()
 		{
 			$text_key_error = $error == 'password_short' ?
 				sprintf($txt['profile_error_' . $error], (empty($modSettings['password_strength']) ? 4 : 8)) :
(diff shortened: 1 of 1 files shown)


live627:
Can you add some instructions on how to reproduce? I'm a bit lost here...

BrickOzp:
1. Activate Warning system
2. View a user profile
3. Actions -> Issue a warning
4. Leave the "Reason for Warning" box empty
5. Press "Issue Warning" button
6. "Undefined index: profile_error_You must enter a reason for altering the warning level of a member" in error log

live627:
oh       I see—I failed step 4—didn't know that it was optional

BrickOzp:
It is not. The error happens when the error message is shown.

live627:
Note to self do not  post right before bed

PR #6558, opened 2021-03-04
Add SMF_2-1_NEXT branch for next release

sbulen:
Allow for periodic updates of GH code over on CrowdIn.

This goes for weekly udpates.  It will update a new, separate branch, SMF_2-1_NEXT.

@jdarwood007 - My only question here is whether you need to specify the branch on your download.  If unspecified, it will probably download both branches, SMF_2-1 & SMF_2-1_NEXT, which might interfere with your process.

So please take a look!   THX....

### DIFF
--- a/.github/workflows/crowdin_wf_next.yml
+++ b/.github/workflows/crowdin_wf_next.yml
@@ -1,9 +1,13 @@
 # CrowdIn workflow for SMF
 
-name: CrowdIn - Update SMF_2-1 branch once per release
+name: CrowdIn NEXT - Update SMF_2-1_NEXT branch periodically
 
 # Controls when the action will run. 
 on:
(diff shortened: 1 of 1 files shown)


jdarwood007:
The generation will generate all of them based on how we export from Crowdin, but the package build process will only build from the "SMF_2-1" for the SMF 2.1 branch.  It also can build from the "SMF_2-0" into the SMF 2.0 branch if we import those.  All other folders would be ignored.

sbulen:
If it helps, this is what builds/downloads would look like.  You only want the SMF_2-1 one.

DownloadBothBranches

sbulen:
Based on the internal discussion, sounds like this is good to go!

live627:
Ok then I'm going to merge this one now.

PR #6562, opened 2021-03-08
Remember location after changing language

### LINKED ISSUE #5625: Changing the language redirects to index.php
#### Description
Changing the language from the top-right dropdown menu redirects to `index.php?language=foo`. Is there a technical reason for this? Could it be changed to redirect to `<current deeplink>?language=foo`?
Currently, I either have to re-click my way back into the maze that is, for example, the admin center, or use my browsers back button and reload to see the language change.

### Steps to reproduce
1. Install another language in your forum
2. While n clicks deep into some random forum function, change the language
3. Find yourself being redirected to the forum front page

- Version/Git revision: HEAD

BrickOzp:
Redirect to old url after language change.

Fixes #5625

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -850,6 +850,8 @@ function loadUserSettings()
 				updateMemberData($user_info['id'], array('lngfile' => $user_info['language']));
 			else
 				$_SESSION['language'] = $user_info['language'];
+			// Reload same url with new language
+			redirectexit($_SESSION['old_url']);
 		}
 		elseif (!empty($_SESSION['language']) && isset($languages[strtr($_SESSION['language'], './\\:', '____')]))
 			$user_info['language'] = strtr($_SESSION['language'], './\\:', '____');


live627:
Could this use the old_url from the session similar to the login redirect? Then no more business logic gets thrown into the template.

[some examples](https://github.com/SimpleMachines/SMF2.1/search?q=old_url)

BrickOzp:
Thanks. Much cleaner this way.

Sources/Load.php
```diff
@@ -850,6 +850,9 @@ function loadUserSettings()
 				updateMemberData($user_info['id'], array('lngfile' => $user_info['language']));
 			else
 				$_SESSION['language'] = $user_info['language'];
+			// Reload same url with new language
+			header("Location: " . $_SESSION['old_url']);
+			exit();
```
live627:
```suggestion
			// Reload same url with new language
			redirectexit($_SESSION['old_url']);
```

live627 (approved):

PR #6569, opened 2021-03-14
Properly call timeformat(); get all birthdays

### LINKED ISSUE #5863: Calendar holidays - show wrong dates in different timezones
#### Description
Holidays can display wrong date if viewed from a different timezone than the forum default (server) timezone.  

### Steps to reproduce
Veterans Day - User using forum default timezone in MST:
image

Veterans Day - note wrong date - Same forum, different user, user-specified timezone is PST:
image

- Version/Git revision: Today
- Database Type: MySQL
- Database Version: 5.7
- PHP Version: 7.1

https://www.simplemachines.org/community/index.php?topic=570135.0

sbulen:
Fixes #5863 

The basic problem was that the call to timeformat() was applying timezone offsets, which wasn't appropriate for things like birthdays & holidays.  

While in there discovered it wasn't getting all birthdays, in particular, birthdays without years specified were dropped.

Feedback welcome,

**_BEFORE:  Note dates in list are incorrect & don't match the grid (3/17 & 3/20), birthday missing (3/18):_**
Cal_fix_before-1
Cal_fix_before-2

**_AFTER:  Dates are in alignment; birthday properly displayed:_**
Cal_fix_after-2
Cal_fix_after-1

### DIFF
--- a/Sources/Subs-Calendar.php
+++ b/Sources/Subs-Calendar.php
@@ -39,8 +39,7 @@ function getBirthdayRange($low_date, $high_date)
 		$result = $smcFunc['db_query']('birthday_array', '
 			SELECT id_member, real_name, YEAR(birthdate) AS birth_year, birthdate
 			FROM {db_prefix}members
-			WHERE YEAR(birthdate) != {string:year_one}
-				AND MONTH(birthdate) != {int:no_month}
+			WHERE MONTH(birthdate) != {int:no_month}
 				AND DAYOFMONTH(birthdate) != {int:no_day}
 				AND YEAR(birthdate) <= {int:max_year}
 				AND (
@@ -52,7 +51,6 @@ function getBirthdayRange($low_date, $high_date)
 				'is_activated' => 1,
 				'no_month' => 0,
 				'no_day' => 0,
-				'year_one' => '1004',
 				'year_low' => $year_low . '-%m-%d',
 				'year_high' => $year_high . '-%m-%d',
 				'low_date' => $low_date,
@@ -699,15 +697,25 @@ function getCalendarList($start_date, $end_date, $calendarOptions)
 	// Give birthdays and holidays a friendly format, without the year
 	$date_format = str_replace(array('%Y', '%y', '%G', '%g', '%C', '%c', '%D'), array('', '', '', '', '', '%b %d', '%m/%d'), get_date_or_time_format('date'));
 
+	// timeformat() uses $user_info to get the format
+	if (isset($user_info['time_format']))
+		$saved_dt_format = $user_info['time_format'];
+	$user_info['time_format'] = $date_format;
 	foreach (array('birthdays', 'holidays') as $type)
 	{
 		foreach ($calendarGrid[$type] as $date => $date_content)
 		{
-			$date_local = preg_replace('~(?<=\s)0+(\d)~', '$1', trim(timeformat(strtotime($date), $date_format), " \t\n\r\0\x0B,./;:<>()[]{}\\|-_=+"));
+			// Make sure to apply no offsets
+			$date_local = preg_replace('~(?<=\s)0+(\d)~', '$1', trim(timeformat(strtotime($date), true, true), " \t\n\r\0\x0B,./;:<>()[]{}\\|-_=+"));
 
 			$calendarGrid[$type][$date]['date_local'] = $date_local;
 		}
 	}
+	// Restore it exactly as you found it
+	if (isset($saved_dt_format))
+		$user_info['time_format'] = $saved_dt_format;
+	else
+		unset($user_info['time_format']);
 
 	loadDatePicker('#calendar_range .date_input');
 	loadDatePair('#calendar_range', 'date_input', '');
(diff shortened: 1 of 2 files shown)


live627:
You override `$user_info`? Why? Does this need to show today/yesterday rather than an actual date?

sbulen:
> 
> 
> You override `$user_info`? Why? Does this need to show today/yesterday rather than an actual date?

From above:
> Note that timeformat() actually expects the format code to be in $user_info, not passed. So... Saved it off before the call & restored it after. Open to other approaches, if you have any...

It's showing a date in a shortened format, e.g., Mar 18, as shown in the screenshots above.  But without offsets or tz translations applied.

Note that timeformat() itself also steps on $user_info...

Alternatives are to add a date format string to the function call to timeformat() or to abandon timeformat() altogether for this.  timeformat() is a bear, but it does factor in all variations of OS, user setting, system setting, language prefs, etc., etc., and is based on using our $txt strings.

live627 (approved):

Sources/Subs-Calendar.php
```diff
@@ -699,15 +697,25 @@ function getCalendarList($start_date, $end_date, $calendarOptions)
 	// Give birthdays and holidays a friendly format, without the year
 	$date_format = str_replace(array('%Y', '%y', '%G', '%g', '%C', '%c', '%D'), array('', '', '', '', '', '%b %d', '%m/%d'), get_date_or_time_format('date'));
 
+	// timeformat() uses $user_info to get the format
+	if (isset($user_info['time_format']))
+		$saved_dt_format = $user_info['time_format'];
+	$user_info['time_format'] = $date_format;
```
Sesquipedalian: This part is unnecessary. Although this behaviour is poorly documented, the `$show_today` parameter of `timeformat()` can either accept a boolean or a string containing a `strftime()` format string. There's no need to juggling around the existing value of `$user_info['time_format']`.

Sources/Subs-Calendar.php
```diff
@@ -699,15 +697,25 @@ function getCalendarList($start_date, $end_date, $calendarOptions)
...
+			$date_local = preg_replace('~(?<=\s)0+(\d)~', '$1', trim(timeformat(strtotime($date), true, true), " \t\n\r\0\x0B,./;:<>()[]{}\\|-_=+"));
 
 			$calendarGrid[$type][$date]['date_local'] = $date_local;
 		}
 	}
+	// Restore it exactly as you found it
+	if (isset($saved_dt_format))
+		$user_info['time_format'] = $saved_dt_format;
+	else
+		unset($user_info['time_format']);
```
Sesquipedalian: This part is unnecessary. See previous comment.

Sources/Subs-Calendar.php
```diff
@@ -699,15 +697,25 @@ function getCalendarList($start_date, $end_date, $calendarOptions)
...
+	if (isset($user_info['time_format']))
+		$saved_dt_format = $user_info['time_format'];
+	$user_info['time_format'] = $date_format;
 	foreach (array('birthdays', 'holidays') as $type)
 	{
 		foreach ($calendarGrid[$type] as $date => $date_content)
 		{
-			$date_local = preg_replace('~(?<=\s)0+(\d)~', '$1', trim(timeformat(strtotime($date), $date_format), " \t\n\r\0\x0B,./;:<>()[]{}\\|-_=+"));
+			// Make sure to apply no offsets
+			$date_local = preg_replace('~(?<=\s)0+(\d)~', '$1', trim(timeformat(strtotime($date), true, true), " \t\n\r\0\x0B,./;:<>()[]{}\\|-_=+"));
```
```suggestion
			// Make sure to apply no offsets
			$date_local = preg_replace('~(?<=\s)0+(\d)~', '$1', trim(timeformat(strtotime($date), $date_format, true), " \t\n\r\0\x0B,./;:<>()[]{}\\|-_=+"));
```
All that is needed to fix this issue is to set the `$offset_type` parameter to a value other than `false` or `'forum'`. 

Passing a string to the `$show_today` parameter is supported by the `timeformat()` function, although the function's documentation currently fails to explain this. It is probably a good idea to update the documentation to reflect this, since this behaviour is used in several places throughout the codebase where output in a particular format is required. (To be clear, I think this is counter-intuitive behaviour, but it's been like this since at least SMF 2.0, so we just have to live with it.)

Sesquipedalian (dismissed):

Sources/Subs-Calendar.php
sbulen: Thanks, good input.  I wasn't aware the string could be specified in that field.  

I made that clear in the comments & in the doc block.

Sources/Subs-Calendar.php
sbulen: Done

Sources/Subs-Calendar.php
sbulen: Done.

sbulen:
Retested after applying Sesq's suggestions & it still tests OK.  

Enhanced the docs a bit for timeformat().

MissAllSunday (approved):

PR #6572, opened 2021-03-15
Improve smiley editor

### LINKED ISSUE #5368: UI improvements needed for smiley editor
Add smiley to a single set, example = Image for the Fugue's Set set:  
upload the smiley and it gives the error:  All smiley sets must have an image

If you check the box for, Same image for all sets: then the smiley gets added without problem.

Also does it really need to say set twice when we know it's a set 
Image for the Alienine's Set set:
Image for the Fugue's Set set: 
Image for the Aaron's Set set:
Image for the Akyhne's Set set: 

https://github.com/SimpleMachines/SMF2.1/compare/a31952a...release-2.1

BrickOzp:
* Default "Same image for all sets" to checked, when adding
new smiley.
* Add help text about "Same image for all sets".
* Make it possible to upload new images when editing existing
smileys.

Fixes #5368


Sources/ManageSmileys.php
```diff
@@ -1011,6 +1013,58 @@ function EditSmileys()
 				if (empty($_POST['smiley_code']))
 					fatal_lang_error('smiley_has_no_code');
 
+				// If upload a new smiley image, check that smiley set folders are writable for the sets with new images.
+				$writeErrors = array();
+				foreach ($_FILES['smiley_upload']['name'] as $set => $name)
+				{
+					if (!empty($name))
+						if (!is_writable($context['smileys_dir'] . '/' . $set))
```
live627: Can these two conditions be combined?

Sources/ManageSmileys.php
```diff
@@ -1011,6 +1013,58 @@ function EditSmileys()
...
+							$writeErrors[] = $set['path'];
+				}
+
+				if (!empty($writeErrors))
+					fatal_lang_error('smileys_upload_error_notwritable', true, array(implode(', ', $writeErrors)));
```
live627: General question to whomever: should these errors be logged? (imo they shouldn't, but there may be a reason here that I missed)

BrickOzp: I did also wonder about that. This seems to be used as error handling in many places in the admin section. Is there another way that will just show the error to the user?
The code is mostly copied from AddSmiley()

Sources/ManageSmileys.php
BrickOzp: Sure can

MissAllSunday (approved):

Issue #6546, opened 2021-02-28
Main menu shines through Admin submenu


wintstar:
#### Description
--Description Here--
Bug in Pull request #6502

https://github.com/SimpleMachines/SMF2.1/pull/6502/files#diff-b1a4b2eef9a8e1022a8abbc9ca8789f48779e4693e82d106146d3d506ff31129R2761

```
#main_menu .popup_container, #genericmenu > .popup_container {
    display: block;
    position: relative;
    background: none;
    margin: 0;
    height: auto;
    z-index: 10;
}
```
With this change  **z-index: 10;**, the main menu can be seen in the submenu of the Admin menu. The bug only came with this pull request.

![Screenshot_2021-02-28 Administration Center](https://user-images.githubusercontent.com/2753863/109407808-4bd7a600-7984-11eb-9ae7-ecc0543e5107.png)

Browser Chrome Version 88.0.4324.182 (Offizieller Build) (64-Bit) and Firefox 86.0 (64-Bit).

### Additional information/references

PR #6574, opened 2021-03-15
Popup_container vs sceditor

### LINKED ISSUE #6573: Menu shines through login box
#### Description
The main menu shines through the dimming and the login box.

### Steps to reproduce
1. Logout
2. Click Login
3. Notice the main menu shines through the dimming, and the login box.
menu shines through

- Version/Git revision: 89c3980d3
- Database Type: MySql
- Database Version:
- PHP Version: 5.6

Related to https://github.com/SimpleMachines/SMF2.1/pull/6502/commits/e81ea07a862ec3e7f445db537e901b630c8f0e1a and 
https://github.com/SimpleMachines/SMF2.1/commit/52334481ed17645f07e72d875e928ab1f3d8d61a

Reverting those fixes the problem.

DiegoAndresCortes:
Should fix this pesky issue this time.
PS: Fixes #6573 

@BrickOzp

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -2771,9 +2771,6 @@ dl.addrules dt.floatleft {
 	top: 0;
 	padding: 0;
 }
-#main_menu .popup_container {
-	z-index: 10;
-}
 .main_icons.hide_popup {
 	float: right;
 }
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -4078,6 +4078,7 @@ tr[id^='list_news_lists_'] textarea {
 .word_break {
 	word-break: break-all;
 }
+
 div.sceditor-container {
 	z-index: 4;
 }
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -4078,3 +4078,6 @@ tr[id^='list_news_lists_'] textarea {
 .word_break {
 	word-break: break-all;
 }
+div.sceditor-container {
+	z-index: 4;
+}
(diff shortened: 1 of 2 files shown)


BrickOzp:
Both admin menu and login box works ok with this fix. I don't know what the original problem was, but I have found no new issues after a quick test.

BrickOzp (approved):

DiegoAndresCortes:
Could we merge this one, should be in before rc4

BrickOzp:
I agree, this bug affects the login box and will be the first thing seen after install for new smf users. This must be merged before rc4. @MissAllSunday?

MissAllSunday:
yep!

MissAllSunday:
Looks good although we will have to be very careful when updating the editor not to remove the change on its css file. Perhaps we can also take a look at some other alternatives like overwriten the css rules

jdarwood007:
If our index.css includes a more specific rule than the sceditor, then it should take presence over the built in sceditor rule.

I wonder if "div. sceditor-container div.sceditor-dropdown div" would work?

BrickOzp:
That fixes the login box shine through and alse https://github.com/SimpleMachines/SMF2.1/issues/6546.
But it is still not clear to me what the original problem was.

DiegoAndresCortes:
Do you want to overwrite this on the index.css instead?? not clear.

Fixes a very hidden issue originally caused by #4885

MissAllSunday:
Yes, this will also mean we will have to make sure index.css is loaded after the editor css file.  We could also see if inline styles can be used too or even !important

jdarwood007:
You don't need to move index.css.  Just make the rule more specific.

```
div.sceditor-container div.sceditor-dropdown div {
	z-index: 4;
}
```
adding ```div.sceditor-container``` makes it more specific than the one in the sceditor.css and now it overrides it.

DiegoAndresCortes:
done

BrickOzp (approved):
Looks good to me

PR #6575, opened 2021-03-15
fix resize toggle

### LINKED ISSUE #6541: Resized images cause distortion
#### Description
.bbc.resized on phone sized devices seem to cause a distortion and look bad

See this sample image
Screen Shot 2021-02-26 at 10 16 44 AM

The fix I have found that works is to instead of setting max height/width.  I set the height/width to auto.  This seems to scale the image properly so it doesn't look distorted.

Any unforeseen issues with this fix?

DiegoAndresCortes:
Fixes #6541

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -434,8 +434,8 @@ blockquote cite::before {
 .bbc_img.original_size {
 	height: auto !important;
 	width: auto !important;
-	max-height: none;
-	max-width: 100%;
+	max-height: none !important;
+	max-width: 100% !important;
 }
 /* No image should have a border when linked. */
 a img {
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -434,8 +434,8 @@ blockquote cite::before {
 .bbc_img.original_size {
 	height: auto !important;
 	width: auto !important;
-	max-height: none !important;
-	max-width: 100% !important;
+	max-height: none;
+	max-width: 100%;
 }
 /* No image should have a border when linked. */
 a img {
--- a/Themes/default/css/responsive.css
+++ b/Themes/default/css/responsive.css
@@ -243,8 +243,8 @@
 		float: none;
 	}
 	.bbc_img.resized {
-		max-width: 250px;
-		max-height: 300px;
+		width: auto;
+		height: auto;
 	}
 	img.icon, #forumposts .catbg img {
 		display: none;
--- a/Themes/default/css/responsive.css
+++ b/Themes/default/css/responsive.css
@@ -243,8 +243,7 @@
 		float: none;
 	}
 	.bbc_img.resized {
-		width: auto;
-		height: auto;
+		max-height: 300px;
 	}
 	img.icon, #forumposts .catbg img {
 		display: none;


jdarwood007:
The image does not fit in the frame with this.
Screen Shot 2021-03-15 at 11 08 48 AM


See this example using autos
Screen Shot 2021-03-15 at 11 10 13 AM

DiegoAndresCortes:
I see it like second picture on my phone using that same post.
Also it'd be a different behavior because your images have links and this fixes the toggle resize which I thought was the issue.
e.g When clicking on the link the image will still resize (naturally)

DiegoAndresCortes:
This is the same post on my local forum
Screenshot_14

Maybe I don't fully understand what exactly is the problem you're having and this is fixing something different.

jdarwood007:
The picture is being stretched downwards making it look weird.  It should be more square of a picture

DiegoAndresCortes:
I see now, so the issue I found is caused by these attributes not using auto like you say.
Your proposed changed should be good enough for this.

~~However it now needs to readjust for the resize, will do that in a bit~~  Nvm it's already fine I didn't refresh my browser.
Will do some more testing

DiegoAndresCortes:
@jdarwood007 I think this is the ideal scenario imo
Otherwise images that are already resized in the img tag to very low values like 100x100 would show up in full width

jdarwood007:
Still looks weird to me.

I checked around and found this: https://stackoverflow.com/questions/12991351/css-force-image-resize-and-keep-aspect-ratio
Seems to maintain what I found which is setting height/width with the value auto, but we can still specify a max on those as well.
I tried the object-fit CSS but it adds unnecessary margins around the image, but does force the image to maintain its aspect ratio.  The autos seems to solve this as well.

I also don't think !important is a good usage.  It seems poor practice to use here.

DiegoAndresCortes:
It's required because the javascript toggles a class that is defined on index.css and not responsive.css.

Also this is what I'm referring to:
Screenshot_15
This image is set to 100x100 on the bbc, so when resizing there's no reason for this image to be upscale.

This what happens on mobile if removing min-height:
Screenshot_16
If it has a small size defined, it shouldn't be increasing in size.

-----

With the current code it'd stay on that size.
Screenshot_17

 ---------

If you want these small pictures to be upscaled then I can use your suggestion, but that doesn't seem to be an expected behavior.

BrickOzp:
@jdarwood007 What SMF version are you on?
Your html code shows width and height are set on the img-tag. This is not the case when using auto for with and height, not on latest at least.

BrickOzp:
What about adding 
    height: auto;
    width: auto;
to @media screen and (max-width: 720px) .bbc_img.resized?

jdarwood007:
2.1 RC3  I can upgrade it to a git version though.

That CSS is what I'm doing to achieve desired results.  See https://www.simpledesk.net/

BrickOzp:
> 2.1 RC3 I can upgrade it to a git version though.

No need. Just needed confirmation, because I could not reproduce the problem with width and height set to auto. But that is because RC3 reads the width and height from the image and writes that in the html. But that has changes since.

DiegoAndresCortes:
I provided my concern regarding this with the example above.
I can push the changes (if any) once there's a decision about this.

PR #6578, opened 2021-03-16
Ensure board managers can access all boards

### LINKED ISSUE #5830: Manage boards conflicts
#### Description
If you add 'manage boards and categories' to a group, BUT, that group did not have access to all boards prior, you create an illogical situation.  You are disallowed from modifying board permissions to correct.   

Workaround is to remove 'manage boards and categories' from that group, alter board permissions, then add it back...  

The problem is that 'manage boards and categories' is intended to be a global setting (as in all boards), but you can assign it to groups that don't have all boards...   

Possible solutions?  Disallow selection of that permission if the group doesn't have access to all boards?  Propagate to all boards upon selection (feels a little dangerous...)?

### Steps to reproduce
1. Remove global moderators access from BOARD A
2. Add 'manage boards and categories' to global moderators
3. Now attempt to modify board permissions for BOARD A; you cannot add global moderators because it is grayed out & protected.

- Database Type: mysql
- Database Version: 5.7
- PHP Version: 7.2

This is from the support forum:
https://www.simplemachines.org/community/index.php?topic=569994.msg4034683#msg4034683

I've confirmed in my local WAMP with the latest from GH - it's not just an RC2 issue.

sbulen:
Fixes #5830 

If a group is given the ability to manage boards & categories, this change will follow thru & add board access for all boards to that group.  This is fixing a situation where they could actually be locked out under some circumstances (see the original issue).

Existing help text already says they will be given access to all boards.  This makes that statement true...

Also fixed a bug in which board_manager_groups was still not getting saved as a setting...  

Tested in both pg & mysql.

Note this is a one-way deal - if you remove access, we do not want board access removed for that group across all boards.  You cannot 'undo' this by just flipping the permission for that group.

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -326,7 +326,7 @@ function reloadSettings()
 		require_once($sourcedir . '/Subs-Members.php');
 		$board_managers = groupsAllowedTo('manage_boards', null);
 		$board_managers = implode(',', $board_managers['allowed']);
-		updateSettings(array('board_manager_groups' => $board_managers), true);
+		updateSettings(array('board_manager_groups' => $board_managers));
 	}
 
 	// Is post moderation alive and well? Everywhere else assumes this has been defined, so let's make sure it is.
--- a/Sources/ManagePermissions.php
+++ b/Sources/ManagePermissions.php
@@ -2423,16 +2423,62 @@ function removeIllegalBBCHtmlPermission($reload = false)
 
 /**
  * Makes sure $modSettings['board_manager_groups'] is up to date.
+ * Also ensures all board managers have access to all boards.
  */
 function updateBoardManagers()
 {
-	global $sourcedir;
+	global $sourcedir, $modSettings, $smcFunc;
 
 	require_once($sourcedir . '/Subs-Members.php');
-	$board_managers = groupsAllowedTo('manage_boards', null);
-	$board_managers = implode(',', $board_managers['allowed']);
+	$board_managers = $board_managers_no_admin = groupsAllowedTo('manage_boards', null)['allowed'];
+	$board_managers = implode(',', $board_managers);
 
+	// If no changes, don't bother with the updates
+	if ($board_managers == $modSettings['board_manager_groups'])
+		return;
+
+	// Update the setting
 	updateSettings(array('board_manager_groups' => $board_managers), true);
+
+	// Strip admin from the set, it is assumed on the boards
+	$key = array_search('1', $board_managers_no_admin);
+	if ($key !== false)
+		unset($board_managers_no_admin[$key]);
+
+	// If the only one was an admin, you're done...
+	if (empty($board_managers_no_admin))
+		return;
+
+	// Make sure the board_managers have access to all boards. First, find boards with gaps...
+	$where = '';
+	foreach($board_managers_no_admin as $group)
+		$where .= (empty($where) ? '' : ' OR') . " (FIND_IN_SET({$group}, member_groups) = 0)";	
+
+	$request = $smcFunc['db_query']('', '
+		SELECT id_board, member_groups
+		FROM {db_prefix}boards
+		WHERE' . $where,
+		array()
+	);
+
+	// ...now fix 'em...
+	while ($row = $smcFunc['db_fetch_assoc']($request))
+	{
+		$mg_array = explode(',', $row['member_groups']);
+		$mg_array = array_unique(array_merge($mg_array, $board_managers_no_admin));
+		sort($mg_array, SORT_NUMERIC);
+		$mg_string = implode(',', $mg_array);
+
+		$upd = $smcFunc['db_query']('', '
+			UPDATE {db_prefix}boards
+				SET member_groups = {string:mg_string}
+				WHERE id_board = {int:id_board}',
+			array(
+				'mg_string' => $mg_string,
+				'id_board' => $row['id_board'],
+			)
+		);
+	}
 }
 
 /**


Sesquipedalian:
I really don't think this is a good idea, but I have an alternative proposal.

First, I'll explain my concerns:

The intended behaviour is that having the "Manage boards and categories" permission will function as an override of the normal board access permissions, in order to allow membergroups with that permission to do their job. This override is intentionally designed *not* to grant the group access to all boards via the normal system. The idea here is to allow the "Manage boards and categories" permission to be granted when needed and then easily revoked when no longer needed, without forcing the admin to go through all boards afterwards to manually revoke access to each one for the membergroup in question. But this PR will instead permanently alter the normal board access permissions for the group in question, defeating the purpose of the override aspect.

Now, my alternative proposal:

If I understand the issue correctly, it is that this bit of the UI currently doesn't allow granting normal board access permissions to a group that has the "Manage boards and categories" permission:
Screen Shot 2021-03-27 at 2 00 45 PM

That's simply a UI problem. In the example above, the Global Moderator membergroup has the "Manage boards and categories" permission, and so the checkbox in the HTML has the `checked` and `disabled` attributes applied to it here: https://github.com/SimpleMachines/SMF2.1/blob/84feb66d9497ad309d25766382993573ded57a3f/Themes/default/ManageBoards.template.php#L383
and here:
https://github.com/SimpleMachines/SMF2.1/blob/84feb66d9497ad309d25766382993573ded57a3f/Themes/default/ManageBoards.template.php#L399

The `disabled` attribute is what prevents the admin from changing the membergroup's normal board access permissions. So if we just change that line, the admin will be able to grant or remove normal board access permissions to the membergroup in question.

Of course, as long as the membergroup continues to have the "Manage boards and categories" permission, that will override anything set in the boards normal access permissions, so if we get rid of the `checked disabled` attributes, we should probably add some other indicator or warning to the admin for this situation.

Sesquipedalian:
Okay, I've looked at this further, and I believe the current code is behaving correctly. The UI should indeed give the checkbox the `checked` and `disabled` attributes, because the admin cannot in fact disallow access to the board for membergroups with the "Manage boards and categories" permission.

What we really need is simply to add a clear explanation of the situation to the UI, so that the admin is not confused. Something like this should do the job nicely:
Screen Shot 2021-03-27 at 7 06 54 PM


With that, the admin knows exactly why the Global Moderator group's access is enabled and can't be disabled, and knows exactly what to do about it if they want to change the situation.

sbulen:
I don't agree with this - it's confusing & unfriendly.  Very difficult to fix.  

The whole crux of the original change was to give the board managers access to all the boards.  We should probably do so, otherwise we will continue finding bugs & edge cases in how this was implemented.  We keep saying we are giving them the access, so maybe we should give them the access...  

We should not allow this limbo scenario, where we say they have all the access but it's not enabled for some boards...

OTOH, there is another option that is somewhat similar, but considerably more friendly IMO...  Do not enable 'manage boards & categories' (in the various places it can be selected) unless that group already has access to all boards.  At least that way, they can't get stuck...  And it's a little safer, since the admin must explicitly approve board access to all the boards before granting 'manage boards & categories'.

Sesquipedalian:
Board managers *do* have access to all boards. The "bug" described in #5830 is simply that the admin can't toggle that checkbox, which is the correct behaviour even if it seems counter-intuitive. A real bug would be if board managers could not in fact access all boards, but that isn't what the issue report describes. It just describes not being able to toggle that checkbox.

Sesquipedalian:
To be clear, this statement by Illori from Oct 20, 2019, is not true of the current behaviour:
> in SMF 2.1, it sounds like they should be able to view all boards but it does not actually work that way. the boards are grayed out and if it was not checked already they cannot view the board.

In fact, following the steps to reproduce #5830 creates a situation where:
(1) Global Moderators are in fact able to access all boards.
(2) The checkbox for Global Moderator in the Allowed Groups settings of the Modify Board page shows up as always enabled and unable to be disabled, accurately reflecting the reality that they have access that cannot be disabled using that setting.

If there is a way to create a situation in which the Global Moderator group has the "Manage boards and categories" permission, yet is unable to access certain boards, it is not by following the steps given in that bug report—at least, not in any of my tests.

sbulen:
> If there is a way to create a situation in which the Global Moderator group has the "Manage boards and categories" permission, yet is unable to access certain boards, it is not by following the steps given in that bug report—at least, not in any of my tests.

Actually that's what I was seeing, and what was reported on the support boards as well (see the original post).  Folks actually got locked out...  The only way to provide access was to _disable_ that permission...

> The idea here is to allow the "Manage boards and categories" permission to be granted when needed and then easily revoked when no longer needed, without forcing the admin to go through all boards afterwards to manually revoke access to each one for the membergroup in question.

Reversibility is not doc'd anywhere that I have seen...  What is clearly doc'd is that selecting the option will provide access to all boards.


And...  Of course, I can no longer reproduce the original issue, I have since re-loaded that test environment to test something else.  You are correct that it does not happen in a simple vanilla environment.

The original support writeup was, IIRC, a migrated forum, with subboards, and had deny permissions enabled.  I am going to try to recreate the conditions where this happened...

Sesquipedalian:
If you are able to find a way to reproduce the issue, that will be good. For my part, I cannot based on the information provided.

sbulen:
I am going to close this, as I cannot reproduce the original error at this time (enabling board manager privileges actually removed access for folks in that membergroup).  This has been driving me NUTS as it was easily reproduced earlier... 

I will submit a one-liner for the setting not getting saved.

PR #6583, opened 2021-03-17
Move max attachments logic to javascript

### LINKED ISSUE #6582: Possible to upload more attachment than max attachments setting
#### Description
If uploading max amount of attachments in a post it is possible to modify the post and add just as many attachments again. This could be done multiple times, resulting in an almost infinite number of attachments.

### Steps to reproduce
1. Set a max attachments limit. Default is 4. (Admin -> Forum -> Attachments and Avatars -> Attachments settings -> Max number of attachments per post)
2. Create a new post and upload max amount of attachments.
3. Edit the post
4. You could now add more attachments, even if the limit is reached.

- Version/Git revision: 634fdfcf3
- Database Type: MySql
- Database Version:
- PHP Version: 5.6

BrickOzp:
The logic of how many attachments remaining that could
be uploaded was handled in PHP. This resulted in that if
the max amount of attachments was attached to a post.
If modifying it, you could not remove an attachment and
add another one, without saving it in between.
Move this logic into javascript, to solve this issue.
Also clamped the remaining attatchment counter to 0.

Fixes #6582

### DIFF
--- a/Sources/Post.php
+++ b/Sources/Post.php
@@ -1218,7 +1218,7 @@ function Post($post_errors = array())
 	if ($context['can_post_attachment'])
 	{
 		// If they've unchecked an attachment, they may still want to attach that many more files, but don't allow more than num_allowed_attachments.
-		$context['num_allowed_attachments'] = min(ini_get('max_file_uploads'), (empty($modSettings['attachmentNumPerPostLimit']) ? 50 : $modSettings['attachmentNumPerPostLimit'] - count($context['current_attachments'])));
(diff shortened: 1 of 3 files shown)


Sesquipedalian:
Including logic for this in JavaScript is good, but I suspect it is not enough. If the user has disabled JavaScript and is using the legacy uploader instead, what prevents them from doing the same trick that way? We should have a way of detecting and preventing this type of abuse on the server side in PHP.

BrickOzp:
A server side control already exists if javascript is disabled.

live627 (approved):

PR #6585, opened 2021-03-17
Escape & when writing tags to xml

### LINKED ISSUE #6584: Profile export to html fails if using gravatar
#### Description
If the user has selected a gravatar as avatar the export to html fails.

### Steps to reproduce
1. Set a gravatar as avatar
2. Download Profile Data as HTML
3. Check error log: "2: DOMDocument::load(): EntityRef: expecting ';' in /storage/content/69/176569/dev.swebrick.se/public_html/exports/1_fc5176209570e2d989a0c4034fa2085360afbf51.html, line: 11"
4. Download the created .html file
5. The content is xml and not html as expected.

- Version/Git revision: 634fdfcf3
- Database Type: MySql
- Database Version:
- PHP Version: 5.6

BrickOzp:
Fixes #6584

### DIFF
--- a/Sources/News.php
+++ b/Sources/News.php
@@ -660,10 +660,10 @@ function dumpTags($data, $i, $xml_format = '', $forceCdataKeys = array(), $nsKey
 			}
 			// A string with returns in it.... show this as a multiline element.
 			elseif (strpos($val, "\n") !== false)
-				$context['feed']['items'] .= "\n" . (!empty($element['cdata']) || $forceCdata ? cdata_parse(fix_possible_url($val), $ns, $forceCdata) : fix_possible_url($val)) . "\n" . str_repeat("\t", $i);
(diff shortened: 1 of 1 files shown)


Sesquipedalian (changes requested):
This approach has a high risk of producing double-encoded entities, i.e. `&amp;amp;` in the output for other fields. Generally speaking, News.php is designed to use CDATA tags to handle entities in user supplied data, such as gravatar URLs. So it would be better to add `'cdata' => true,` to the avatar section of the data array in `getXmlProfile()`.

Sesquipedalian:
As I looked at News.php, I noticed several other fields that should have been passed through `cdata_parse()`, so I went ahead and created a separate PR to implement that. Please test #6616 to make sure it solves the issue for you as well, @BrickOzp.

PR #6586, opened 2021-03-17
Remove unused text strings

BrickOzp:


BrickOzp:
Any point in keeping these?

live627:
I'm guessing that some older themes may possibly still use these—but then again, they need completely redone for 2.1, so that point is moot.

DiegoAndresCortes (approved):
One of those isn't even used in SMF 2.0

MissAllSunday (approved):

PR #6588, opened 2021-03-18
Fix error when editing language file with arrays

BrickOzp:
If using PHP 5.6 and editing a language file containing
arrays, an error was reported to the error log. Fixed that.

### DIFF
--- a/Sources/ManageLanguages.php
+++ b/Sources/ManageLanguages.php
@@ -1148,7 +1148,7 @@ function ModifyLanguage()
 				continue;
 
 			// These are arrays that need breaking out.
-			if (strpos($entryValue['entry'], 'array(') === 0 && strpos($entryValue['entry'], ')', -1) === strlen($entryValue['entry']) - 1)
+			if (strpos($entryValue['entry'], 'array(') === 0 && substr($entryValue['entry'], -1) === ')')
 			{
 				// No, you may not use multidimensional arrays of $txt strings. Madness stalks that path.
 				if (isset($entryValue['subkey']))


BrickOzp:
How to reproduce.
1. Admin -> Configuration -> Language -> Edit languages
2. Edit a language (english is ok)
3. Select a language file with arrays in it. e.g. General Strings
4. Error in error log: /Sources/ManageLanguages.php (Line 1151)  2: strpos(): Offset not contained in string

sbulen (commented):
I think the -1 makes sense here...  It is telling strpos to look at the rightmost character.  It is trying to identify arrays by finding 'array(' at the start and ')' at the end.  (In a very convoluted way...)

sbulen:
This is another one where it would really help to test RC4 code with RC4 language strings.

@jdarwood007 - please see my PM.

sbulen:
You know what? I think this is a php version issue.  The -1 fails in php5.x.  Works in 7.x...  Support for negative offsets was added in php 7.1...

sbulen (changes requested):
This is a php version issue.  The -1 fails in 5.6, which we still support.

I recommend changing this line to use substr instead of strpos.  substr supports the negative offset in 5x.  That will make it easy to look at the last character.

My only problem with moving the offset to 0 is that, in theory, you can have a ')' character somewhere _within_ the array.

BrickOzp:
Makes sense why nobody else have seen this problem.

sbulen (approved):

sbulen:
Since this was a php version issue, not a real language issue with arrays, I was able to test this.

Tested before & after in 5.6.  Confirmed issue before, & confirmed fix works.

PR #6589, opened 2021-03-19
Send 403 for banned login attempt

### LINKED ISSUE #6587: Ajax login will fail for banned users on non-english
#### Description
In index.template.php we have this:
```
				<li>', sprintf($txt[$context['can_register'] ? 'welcome_guest_register' : 'welcome_guest'], $context['forum_name_html_safe'], $scripturl . '?action=login', 'return reqOverlayDiv(this.href, ' . JavaScriptEscape($txt['login']) . ');', $scripturl . '?action=signup'), '</li>
```

The second parameter of reqOverlayDiv is $txt['login']
This is set in the index.language.php
```
$txt['login'] = 'Login';
```

In script.js we define the parameters:
```
// Open a overlay div
function reqOverlayDiv(desktopURL, sHeader, sIcon)
```

And finally when we receive an error 500, we throw an error that they are banned if it's from the login page.
```
		statusCode: {
			500: function() {
				if (sHeader == 'Login')
					oPopup_body.html(banned_text);
				else
					oPopup_body.html('500 Internal Server Error');
			}
		}
```

But we are depending on sHeader which is a text string.  So in non english forums, this will never trigger the banned message, but only the internal error message.

### Steps to reproduce
1. Change $txt['login'] to anything else.
2. Ban the member
3. Try to login

- Version/Git revision: latest at time of report
- Database Type: n/a
- Database Version: n/a
- PHP Version: n/a

Tagging this for final, it doesn't need to resolve for RC4 but could.

BrickOzp:
Use HTTP error code 403 (Forbidden) instead of 500
when a banned user tries to login.
This will also avoid using a language string to
differ between banned users and other errors.

Fixes #6587

### DIFF
--- a/Sources/Security.php
+++ b/Sources/Security.php
@@ -345,7 +345,7 @@ function is_not_banned($forceCheck = false)
 		Logout(true, false);
 
 		// You banned, sucka!
-		fatal_error(sprintf($txt['your_ban'], $old_name) . (empty($_SESSION['ban']['cannot_access']['reason']) ? '' : '<br>' . $_SESSION['ban']['cannot_access']['reason']) . '<br>' . (!empty($_SESSION['ban']['expire_time']) ? sprintf($txt['your_ban_expires'], timeformat($_SESSION['ban']['expire_time'], false)) : $txt['your_ban_expires_never']), false);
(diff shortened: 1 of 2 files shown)


live627:
banned_text is now unused. Can you please delete it?


https://github.com/SimpleMachines/SMF2.1/blob/41268e315691c35614d45e96a24f64e5d8313ca7/Sources/Load.php#L2483

BrickOzp:
It is still used

live627:
yeah sorry just noticed

live627:
yeah sorry just noticed

jdarwood007 (approved):

jdarwood007 (approved):

PR #6591, opened 2021-03-19
Check old_url before using it

BrickOzp:

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -850,8 +850,9 @@ function loadUserSettings()
 				updateMemberData($user_info['id'], array('lngfile' => $user_info['language']));
 			else
 				$_SESSION['language'] = $user_info['language'];
-			// Reload same url with new language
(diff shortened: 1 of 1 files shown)


jdarwood007:
Can we add a return so its not up against the previous statement.

BrickOzp:
The redirect was just added in 41268e31 so there should be no problem in continue executing.

live627:
don't needed it because it always exits as the name suggests

BrickOzp:
I got the error when trying to login as a banned user.
You can force it by going to index.php?language=<lang> in a new browser

jdarwood007:
I can reproduce the error as well.

1. Log in as banned user
2. Navigate too ?language=
3. Navigate too ?langauge=english

Don't know if step 3 was needed, but I got the error.  I suspect the true issue is because the ban process is destroying session data.  It isn't worth trying to fix it up to keep that session data and just check that the session is there to be safe.

PR #6592, opened 2021-03-19
Allow CORS requests in SMF

### LINKED ISSUE #5656: CORS policy blocks login popup on sub domains
#### Description
Access to XMLHttpRequest at 'https://www.site.tld/forums/index.php?action=login' from origin 'https://support.site.tld' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.

I suggested a fix for this in #5654 for Cron.php via just a hard code because of the complexity of this.  However this occurs in the main SMF application.  This will also affect other implantations where CORS takes effect due to sub domains.

My suggestion here is we add a feature (YUP!) to allow CORS for Any _(*),_ Referral sub-domain detection (*.site.tld), Referral Multiple domain handling, and disabled (don't see the header).
When we do sub domain detection, we need to know the main domain and may need that value specified for us, (i.e.: site.tld).  Via the referral header we can parse the domain and if its from the sub domain, allow it via returning the matching sub domain for this header.
When we do multiple domain, we should have them manually specify all domains that are allowed.  If we find a match in the referral, we allow it (via returning the matching referral domain for this header.

I believe this should be a blocker issue for release, as this will impact sites needing to specify a CORS header.

### Steps to reproduce
1. Install SMF on www.site.tld/forums/
2. Use SSI.php with $ssi_theme set and call ObExit(), generating a page within SMF template system.
3. Navigate to the page
4. Attempt to login.

See #5654 for how we fixed Cron.php
https://www.moesif.com/blog/technical/cors/Authoritative-Guide-to-CORS-Cross-Origin-Resource-Sharing-for-REST-APIs/
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin

jdarwood007:
This changes checkSession to ignore the HTTP_HOST verification if we have a valid CORS domain found, as HTTP_ORIGIN is trusted: https://stackoverflow.com/questions/4566378/how-secure-is-http-origin.  As such if we have found a valid CORS, we don't need to check the HTTP_HOST because HTTP_REFERER will never validate otherwise.

This also changes send_http_status to send the frame and cors policy headers.  In events where we used send_http_status, sometimes we issued ob_end_clean and that causes all previous headers to be removed.  This ensures we have the headers still to send.  Additionally the 204 status code is now valid default.

This also modifies the index action to move both the Scheduled Task and check that the request method is not an options call.  This is performed by CORS requests and we should avoid doing any extra work other than indicating we have a valid headers for the CORS to accept requests.

We also added a check to the Scheduled Task to ensure during an options request method, we just bail out.  This is performed in addition to the index.php requests incase an external call from SSI.php has called this.

This should fix #5656 

This was bigger than I expected, but not much can be done.  It shows how much work is needed to modify SMF in order to support CORS just for login/out and the basic Profile popup menu.

I recommend due to the size of this, we merge this for RC4 and ensure we can fix any issues later.  Far as I have tested, I expect no regressions for those with CORS disabled.  CORS enabled may have some functionality I missed.  As the size of the changes suggests, this isn't a good candidate to leave out of native support.   There is not enough hooks in the right areas to support this.


sbulen:
I am having a hard time replicating the original issue...

If I allow for global cookies & specify the correct global domain in the cookie settings, I can logon no problem, even outside of the subdomain.  

I am in fact locked out if I disable global cookies - but I think that is expected.

(I'm just using ssi_examples.php, but that has its own little template...)

jdarwood007:
https://stackoverflow.com/questions/26724137/do-i-need-to-enable-cors-when-my-api-is-on-a-subdomain-of-my-main-website/26724206
CORS is required for any requests other than the same domain, even if its a sub domain.

1. Setup SMF on www.domain.tld.
2. Setup a subdomain such as support.domain.tld
3. Setup a SSI page that uses $ssi_theme and build a template_main.
4. Load page
5. Try to use the User info section to login or access user information.

I can send you a private link where you can see this problem in action.

jdarwood007:
I just saw a bug where when the login succeeds, the content of the site is updated with the board index, but the URL remains the old url.  Need to look more into that as it produces odd results to see the board index load on another page it shouldn't.

Sources/Security.php
```diff
@@ -1342,4 +1344,134 @@ function frameOptionsHeader($override = null)
...
+		$boardurl_base = implode('.', array_slice(explode('.', parse_url($boardurl, PHP_URL_HOST)), 1));
+
+		// If we find www, pop one more off.
+		if ('www' === array_slice(explode('.', parse_url($boardurl, PHP_URL_HOST)), 0, 1))
+			$boardurl_base = implode('.', array_slice(explode('.', parse_url($boardurl, PHP_URL_HOST)), 1));
+
+		$origin_base = implode('.', array_slice(explode('.', parse_url($_SERVER['HTTP_ORIGIN'], PHP_URL_HOST)), 1));
+
+		if ('www' === array_slice(explode('.', parse_url($origin_base, PHP_URL_HOST)), 0, 1))
+			$origin_base = implode('.', array_slice(explode('.', parse_url($origin_base, PHP_URL_HOST)), 1));
```
live627: goodness gracious me this code

Sources/ScheduledTasks.php
```diff
@@ -23,6 +23,16 @@ function AutoTask()
...
 	global $smcFunc;
 
+	// We bail out of index.php too early for these to be called.
+	frameOptionsHeader();
+	corsPolicyHeader();
+
+	if (isset($_SERVER['HTTP_X_SMF_AJAX']) && isset($_SERVER['REQUEST_METHOD']) && $_SERVER['REQUEST_METHOD']	=== 'OPTIONS')
+	{
+		send_http_status(204);
+		die;
```
live627, line 34: obExit() seems lonely

Sources/Security.php
jdarwood007: Any suggestions?  Its ugly, but its popping away the www subdomain.  In CORS, www.domain.tld != domain.tld.

Sources/ScheduledTasks.php
jdarwood007: We don't need to call it in obExit because we do it in smf_main, just as our frame security does.

MissAllSunday:
The check on empty for valid_cors_found keys implies there can be logic for when the value is empty and the code changes suggest theres no case for handling that case so perhaps might be good to default to a sane non-empty value for valid_cors_found?  perhaps "same"

Sources/Security.php
```diff
@@ -1342,4 +1344,134 @@ function frameOptionsHeader($override = null)
...
+	if (!empty($modSettings['cors_domains']) && $modSettings['cors_domains'] === '*')
+		$context['cors_domain'] = '*';
+
+	// If subdomain-independent cookies are on, allow sub domains. This will have issues if the forum is at forum.domain.tld and the origin we are at is another.sub.domain.tld.
+	if (empty($context['cors_domain']) && !empty($modSettings['globalCookies']) && !empty($_SERVER['HTTP_ORIGIN']) && filter_var($_SERVER['HTTP_ORIGIN'], FILTER_VALIDATE_URL) !== false)
+	{
+		$boardurl_base = implode('.', array_slice(explode('.', parse_url($boardurl, PHP_URL_HOST)), 1));
+
+		// If we find www, pop one more off.
+		if ('www' === array_slice(explode('.', parse_url($boardurl, PHP_URL_HOST)), 0, 1))
```
BrickOzp: Is this if case needed?
The same code is executed before the if case and inside it.

jdarwood007:
I believe all points are addressed now.

Sources/Security.php
```diff
@@ -1474,4 +1448,33 @@ function corsPolicyHeader($set_header = true)
...
+	// If we find www, pop it out.
+	if ('www' === array_slice(explode('.', $base_domain), 0, 1))
+		$base_domain = implode('.', array_slice(explode('.', $base_domain, 1)), 1);
+
+	// Are we saying this is a sub domain and to remove another domain level?
+	if ($sub_domain)
+		$base_domain = implode('.', array_slice(explode('.', parse_url($domain, PHP_URL_HOST)), 1));
+
+	/*	Note, we do have a TLD regex for the autolinker, however it can not reliably be used here due to
+		Country Code Second Level Domains (example: co.uk) as they are not TLDs.  As these are also
```
BrickOzp, line 1477: Could the list be used here as well?
https://publicsuffix.org/
https://github.com/SimpleMachines/SMF2.1/pull/6764

jdarwood007: I saw it.  I believe that this could work but the benefit doesn't seem to be wroth all the work to determine the limited number of hosts.  I guess we could upon saving of the cors settings, we could rebuild a hidden list of "true" domains if it seems its better for us to figure out the real root domain.

Sources/LogInOut.php
```diff
@@ -45,7 +45,7 @@ function Login()
 
 	$context['sub_template'] = 'login';
 
-	if (!empty($_SERVER['HTTP_X_REQUESTED_WITH']) && $_SERVER['HTTP_X_REQUESTED_WITH'] == 'XMLHttpRequest')
+	if ((!empty($_SERVER['HTTP_X_REQUESTED_WITH']) && $_SERVER['HTTP_X_REQUESTED_WITH'] == 'XMLHttpRequest') || (!empty($context['valid_cors_found']) && !empty($_SERVER['HTTP_X_SMF_AJAX']) && isset($_REQUEST['ajax'])))
```
frandominguezl: New line to improve code readability?

Sources/LogInOut.php
```diff
@@ -99,7 +99,7 @@ function Login2()
 	// Load cookie authentication stuff.
 	require_once($sourcedir . '/Subs-Auth.php');
 
-	if (!empty($_SERVER['HTTP_X_REQUESTED_WITH']) && $_SERVER['HTTP_X_REQUESTED_WITH'] == 'XMLHttpRequest')
+	if ((!empty($_SERVER['HTTP_X_REQUESTED_WITH']) && $_SERVER['HTTP_X_REQUESTED_WITH'] == 'XMLHttpRequest') || (!empty($context['valid_cors_found']) && !empty($_SERVER['HTTP_X_SMF_AJAX']) && isset($_REQUEST['ajax'])))
```
frandominguezl: Same here and also in other places of all proposed changes, if possible

Sources/ScheduledTasks.php
```diff
@@ -23,6 +23,16 @@ function AutoTask()
 {
 	global $smcFunc;
 
+	// We bail out of index.php too early for these to be called.
+	frameOptionsHeader();
+	corsPolicyHeader();
+
+	if (isset($_SERVER['HTTP_X_SMF_AJAX']) && isset($_SERVER['REQUEST_METHOD']) && $_SERVER['REQUEST_METHOD']	=== 'OPTIONS')
```
frandominguezl: Extra space before `===` operator

Sources/Security.php
```diff
@@ -1342,4 +1344,137 @@ function frameOptionsHeader($override = null)
 	header('x-content-type-options: nosniff');
 }
 
+/**
+ * This sets the Access-Control-Allow-Origin header.
+ *
```
frandominguezl, line 1352: Explain the param in the documentation

Sources/Security.php
```diff
@@ -1342,4 +1344,137 @@ function frameOptionsHeader($override = null)
...
+
+		// Be careful with this, your allowing a external site to allow the browser to send cookies with this.
+		if (!empty($modSettings['allow_cors_credentials']))
+			header('Access-Control-Allow-Credentials: true');
+	}
+}
+
+/**
+ * Helper function to figure out a urls base domain.
+ *
```
frandominguezl, line 1456: Explain the param in the documentation

Themes/default/languages/Help.english.php
```diff
@@ -360,6 +360,7 @@
...
+$helptxt['allow_cors'] = 'Cross Origin Resource Sharing or CORS is a policy for modern browsers to trust resources being requests across different domains.  If enabled, this will attempt to resolve CORS policy for scripts accessing the forum resources from different domains.  If subdomain independent cookies (global cookies) are enabled, attempts are made to determine the proper domain.  Supplying the Additional CORS domains with a comma separated list of domains allows for additional domains.  The format should be {SCHEMA}://{DOMAIN}.  If a wildcard (*) is supplied in the domain, attempts are made to resolve the wildcard.  If a wildcard (*) is supplied as the only additional domain, all domains are allowed and this may present a security risk to your forum and should not be used except in extreme cases.  Additional headers can be supplied in a comma separated list and will be allowed during requests containing these headers and wildcard (*) is allowed.';
```
frandominguezl: Rephrase to `to trust resources being requested`.

index.php
```diff
@@ -141,19 +141,26 @@
...
+require_once($sourcedir . '/Session.php');
+require_once($sourcedir . '/Logging.php');
+require_once($sourcedir . '/Security.php');
+require_once($sourcedir . '/Class-BrowserDetect.php');
+
+// If a Preflight is occurring, lets stop now.
+if (isset($_SERVER['REQUEST_METHOD']) && $_SERVER['REQUEST_METHOD'] === 'OPTIONS')
+{
+	send_http_status(204);
+	die;
```
frandominguezl, line 154: Should it be `die()`?

frandominguezl (changes requested):

index.php
jdarwood007: No, die and exit are both language constructs and do not require parenthesis.

Themes/default/languages/Help.english.php
```diff
@@ -360,7 +360,7 @@
...
+$helptxt['allow_cors'] = 'Cross Origin Resource Sharing or CORS is a policy for modern browsers to to trust resources being requested across different domains.  If enabled, this will attempt to resolve CORS policy for scripts accessing the forum resources from different domains.  If subdomain independent cookies (global cookies) are enabled, attempts are made to determine the proper domain.  Supplying the Additional CORS domains with a comma separated list of domains allows for additional domains.  The format should be {SCHEMA}://{DOMAIN}.  If a wildcard (*) is supplied in the domain, attempts are made to resolve the wildcard.  If a wildcard (*) is supplied as the only additional domain, all domains are allowed and this may present a security risk to your forum and should not be used except in extreme cases.  Additional headers can be supplied in a comma separated list and will be allowed during requests containing these headers and wildcard (*) is allowed.';
```
BrickOzp: to to?

MissAllSunday (approved):

jdarwood007:
Also that CI bug can be fixed with:
```
$base_domain = implode('.', array_slice(explode('.', $base_domain, 1), 1));
```

I had them in the wrong position.

PR #6601, opened 2021-03-22
added samesite Cookie settings

### LINKED ISSUE #6594: Add SameSite support to Set-Cookie
SameSite attribute is widely adopted by browsers, why not introduce it?

More info: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite

albertlast:
allow the admin to define the behavior of samesite setting of cookies,
defaul is lax mode,
since none needs a https connection.

fixes #6594


Themes/default/languages/ManageSettings.english.php
```diff
@@ -56,6 +56,10 @@
 $txt['invalid_cookie_domain'] = 'The domain introduced seems to be invalid, please check it and save again.';
 $txt['secureCookies'] = 'Force cookies to be secure';
 $txt['httponlyCookies'] = 'Force cookies to be made accessible only through the HTTP protocol';
+$txt['samesiteCookies'] = 'Force cookies to be send to only first parties';
```
frandominguezl: Force cookies to be sent only to first parties

Themes/default/languages/Help.english.php
```diff
@@ -252,6 +252,7 @@
 $helptxt['enableCompressedOutput'] = 'This setting will compress output in order to lower bandwidth consumption, but it requires
 		zlib to be installed.';
 $helptxt['httponlyCookies'] = 'Cookies won\'t be accessible by scripting languages, such as JavaScript. This setting can help to reduce identity theft through XSS attacks. This can cause issues with third party scripts but should be on wherever possible.';
+$helptxt['samesiteCookies'] = 'The SameSite attribute of the Set-Cookie HTTP response header allows you to declare if your cookie should be restricted to a first-party or same-site context. Further Information <a href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite">MDN</a>';
```
frandominguezl: For further information, please refer to <a href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite">MDN</a>

frandominguezl (changes requested):

Sources/Subs-Auth.php
```diff
@@ -855,11 +856,23 @@ function smf_setcookie($name, $value = '', $expire = 0, $path = '', $domain = ''
 		$httponly = !empty($modSettings['httponlyCookies']);
 	if ($secure === null)
 		$secure = !empty($modSettings['secureCookies']);
+	if ($samesite === null)
+		$samesite = !empty($modSettings['samesiteCookies']) ? $modSettings['samesiteCookies'] : 'lax';
```
jdarwood007, line 860: The documentation states that SameSite=None must be "Secure".  Can you add a logic check for that?

Sources/ManageServer.php
```diff
@@ -482,6 +482,12 @@ function ModifyCookieSettings($return_config = false)
 		array('globalCookiesDomain', $txt['globalCookiesDomain'], 'db', 'text', false, 'globalCookiesDomain'),
 		array('secureCookies', $txt['secureCookies'], 'db', 'check', false, 'secureCookies', 'disabled' => !httpsOn()),
 		array('httponlyCookies', $txt['httponlyCookies'], 'db', 'check', false, 'httponlyCookies'),
+		array('samesiteCookies', $txt['samesiteCookies'], 'db', 'select', array(
+				'none' 		=> $txt['samesiteNone'],
+				'lax' 		=> $txt['samesiteLax'],
+				'strict' 	=> $txt['samesiteStrict']
+			),
+			'samesiteCookies'),
```
jdarwood007, line 490: The documentation states that SameSite=None must be "Secure".  I would add a save logic here that if secure is not checked and samesite is set to none, to set it back to lax.

Themes/default/languages/ManageSettings.english.php
```diff
@@ -56,6 +56,10 @@
...
+$txt['samesiteNone'] = 'None';
+$txt['samesiteLax'] = 'Lax';
+$txt['samesiteStrict'] = 'Strict';
```
jdarwood007, line 62: The word just Lax,Strict,None are not very descriptive, but I don't think there is a better way to explain them without getting wordy or getting something wrong.  I'm mostly wanting to pick on Lax.  I'm assuming that it means "Relaxed".   Ignore this if nobody can think of some better wording.  The documentation describes it, but it is also confusing to new users.

jdarwood007 (changes requested):
Overall, this is a security feature.  I think it should merge in RC4 to ensure it works.  I doubt due to the adoption of it already on the web that we would see anything on normal usages.

Themes/default/languages/ManageSettings.english.php
albertlast: This are the offical states,
to use other words would lead people more in confusion.

Sources/ManageServer.php
albertlast: im aware of this, but i see many edge cases which get to complex for me to handle.
Feel free to provide implementation or do it later on.

isaak654:
According to the Firefox 96.0 [release notes](https://www.mozilla.org/en-US/firefox/96.0/releasenotes/), the SameSite setting is going to be enforced by default:

> Firefox will now enforce the Cookie Policy: Same-Site=lax by default which provides a solid first line of defense against Cross-Site Request Forgery (CSRF) attacks.

isaak654:
According to the Firefox 96.0 [release notes](https://www.mozilla.org/en-US/firefox/96.0/releasenotes), the following SameSite setting is going to be enforced by default:

> Firefox will now enforce the Cookie Policy: Same-Site=lax by default which provides a solid first line of defense against Cross-Site Request Forgery (CSRF) attacks.

sbulen (changes requested):
@albertlast - I've been testing this, and no matter which option I select (none/lax/strict), the ACP displays "lax" as the setting.

Internally, the setting is being updated properly, but the display is incorrect.  Could you correct?

Sources/Subs-Auth.php
jdarwood007: @albertlast You marked this resolved.  I'm not sure I'm reading how?

jdarwood007: FYI,  The documentation I'm reading from is here:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#samesitenone_requires_secure

Should be simple to implant in the logic to enforce if samesite=none to enforce secure.

sbulen (changes requested):
I agree with @jdarwood007 that we cannot allow the combination of "none" without "secure" selected.

I've been testing, and that combo locks everyone out...  It becomes impossible to login...

isaak654:
So, we get back to the starting point.

isaak654:
So, back to the starting point.

albertlast:
feel free to fork the pr,
but i got no time and topic is too old for me to remeber the stuff around this.

PR #6607, opened 2021-03-25
More CrowdIn-friendly line feeds

### LINKED ISSUE #6597: new lines and variables in text string
https://github.com/SimpleMachines/SMF2.1/blob/f03f6efd1b9d3b0cd1bef571389bfbefbba5af3a/Themes/default/languages/index.english.php#L150

Issue reported here:
https://www.simplemachines.org/community/index.php?topic=572767.msg4084946#msg4084946

We should change this string to be compatible with Crowdin, as this will cause bug when translated through it.

sbulen:
Fixes #6597 

When passing arguments to language strings that also had line feeds in them, we sometimes used double quotes around the strings & \n to represent the line feed.  The problem is that the arguments then needed escaping, e.g., %1$s became %1\\$s, and CrowdIn would not export the string properly.  

The fix is to stop using double-quotes around arguments, so we do not have to escape the $.  Fortunately, there are very few strings that need changing.  

**_I had considered 3 options, and <ins>all worked</ins>, but using real line feeds was the cleanest for the translators. The string remained one string, and appeared as expected in the CrowdIn UI:_**
real-lfs-s
In CrowdIn:
real-lfs-t

**_Segregating the line feeds & using double-quotes solely for them worked, but split the string into multiple strings & noted there was whitespace between them:_**
dq-seg-s
In CrowdIn:
dq-seg-t

**_Using PHP_EOL constant was a little cleaner, but still split the string into multiple strings:_**
const-s
In CrowdIn:
const-t

### DIFF
--- a/Themes/default/languages/Admin.english.php
+++ b/Themes/default/languages/Admin.english.php
@@ -130,8 +130,12 @@ $txt['ip_address'] = 'IP address';
 $txt['member_id'] = 'ID';
 
 $txt['unknown'] = 'unknown';
+
 // argument(s): HTTP_REFERER (if applicable), HTTP_USER_AGENT, ip address
-$txt['security_wrong'] = "Administration login attempt!\nReferrer: %1\$s\nUser agent: %2\$s\nIP: %3\$s";
+$txt['security_wrong'] = 'Administration login attempt!
+Referrer: %1$s
+User agent: %2$s
+IP: %3$s';
 
 $txt['email_as_html'] = 'Send in HTML format. (with this you can put normal HTML in the email.)';
 $txt['email_parsed_html'] = 'Add &lt;br&gt;s and &amp;nbsp;s to this message.';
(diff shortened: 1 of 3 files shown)


m4z:
Options 2 and 3 would need sprintf to work (compare #6120).

BrickOzp (approved):

BrickOzp:
It still shows the source string with %1\$s for me on the NEXT branch, but the \n is changed to new lines.
image

sbulen:
You accidentally pulled up the old one on the SMF_2-1 branch.  The SMF_2-1_NEXT branch shows this:
image

sbulen:
> 
> 
> Options 2 and 3 would need sprintf to work (compare #6120).

And would have likely been back to one string, not multiple.  Thanks!

BrickOzp:
Didn't know about the test project.

sbulen:
If someone could take a quick peek - I'd like to have this merged today before the sync early tmw.

sbulen:
I'm going to go ahead & merge this.

PR #6608, opened 2021-03-25
Use correct rotation for thumbnails

### LINKED ISSUE #5165: Inconsistency with inline image attachments
#### Description
There's some inconsistency with image attachments and how they're displayed inline. I tested this with an image that's 4032x1960 (which I can attach here if needed). When inserting the image I leave the height and width at "auto"

If I check the box to resize images when displayed under posts, it shows it as a thumbnail, which can be clicked to view the fullsize image in a new window.

If I uncheck that box and don't set a max height, the larger image appears in the post and don't check the box, the fullsize image is displayed, but is distorted and upside down. If I click on it, it shows smaller than actual size, still upside down, but not distorted.

If I do set a max height, it displays the image with the appropriate height and width, but there is no way to view the original sized image. Nothing happens when I click on the image.

### Steps to reproduce
1. Upload a large image and insert it inline using the "auto" option for height and width
2. Play with image options in the "Attachments and Avatars" section of the admin center.

- Version/Git revision: Latest as of today (11/22/18)
- Database Type: MariaDB
- Database Version: 10.2.x (latest)
- PHP Version: 7.1.x (latest)

In previous versions of SMF, when a larger image was displayed smaller than actual size, you could click on it to view the full size.

Images are as follows - reference image and showing the distortion/incorrect orientation when no max width or height are set. EDIT: It appears something weird is going on either with Chrome or whatever is being used to resize the image as the image shows upside down here too until you click on it.

screenshot_20181122_234402

BrickOzp:
If an image is encoded with a rotation the thumbnail
will be rotated incorrectly. Read the rotation param
from the exif data in tne image and rotate the thumbnail
accordingly.

Fixes #5165

### DIFF
--- a/Sources/Subs-Graphics.php
+++ b/Sources/Subs-Graphics.php
@@ -365,6 +365,26 @@ function resizeImageFile($source, $destination, $max_width, $max_height, $prefer
 		fclose($fp_destination);
 	}
 
+	$rotate = 0;
+	if (($exif_data = @exif_read_data($destination)) !== false)
+	{
+		if (!empty($exif_data['Orientation']))
+		{
+			switch($exif_data['Orientation'])
+			{
+				case 8:
+					$rotate = 90;
+					break;
+				case 3:
+					$rotate = 180;
+					break;
+				case 6:
+					$rotate = -90;
+					break;
+			}
+		}
+	}
+
 	// We can't get to the file. or a previous getimagesize failed.
 	if (empty($sizes))
 		$sizes = array(-1, -1, -1);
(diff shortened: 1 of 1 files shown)


BrickOzp:
I have not tested the MagickWand case since it was not available in my testenvironment, but Imagick and GD works.

Oldiesmann:
I wouldn't worry too much about MagickWand for PHP at this point. It appears that ImageMagick stopped supporting it in favor of IMagick, so as long as that works we should be good.

MissAllSunday (approved):

PR #6609, opened 2021-03-26
Align usage of period in the other options

BrickOzp:
In the other options when creating a topic or post, some
end with period and some don't.
Remove period from those that uses it to align.


BrickOzp:
image

live627 (approved):

PR #6634, opened 2021-03-30
Uses high resolution icons for high resolution screens in Safari

### LINKED ISSUE #6162: Safari doesn't support @media (min-resolution...
index.css - Lines 1727-1732

```
/* Load better icons for higher resolution screens */
@media screen and (min-resolution: 160dpi) {
	.main_icons::before {
		background-image: url(../images/icons/main_icons_sprite_hd.png);
	}
}
```
Problem: Safari, either on desktop or on iOS, doesn't support min-resolution.
See: [https://developer.mozilla.org/en-US/docs/Web/CSS/@media/resolution](url)

This means no nifty hi-res icons on Safari. Workaround for iThingz could be:

```
@supports (-webkit-overflow-scrolling: touch)
```
That will nail any cases of Safari on mobile, but still wouldn't deal with Safari on hi-res desktop. I assume there will be something else that can be used for hi-res desktop (Apple are always inventing heaps of Apple-specific properties) but I haven't looked into that yet.

ETA: This one looks promising.
[https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-mask-attachment](url)

Only supported by Safari, either on desktop or on iOS. Would result in hi-res icons being served even to Safari on lower res desktop, but that's probably not a significant issue. It'd still look fine, and the extra overhead isn't much in the scheme of things.

Sesquipedalian:
Fixes #6162

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -1686,7 +1686,9 @@ tr.windowbg td, tr.bg td, .table_grid tr td {
 	vertical-align: middle;
 }
 /* Load better icons for higher resolution screens */
-@media screen and (min-resolution: 160dpi) {
+@media
+	screen and (-webkit-min-device-pixel-ratio: 1.66),
+	screen and (min-resolution: 160dpi) {
 	.main_icons::before {
 		background-image: url(../images/icons/main_icons_sprite_hd.png);
 	}


DiegoAndresCortes:
Looks good to me

PR #6635, opened 2021-03-31
Convert dates to english

### LINKED ISSUE #6631: Date picker does not work for non english languages
#### Description
When selecting dates with the date picker, and the forum uses a non english language, the selected date is not used.

### Steps to reproduce
1. Install and use a languages that is not english
2. Use a time format with strings for months. e.g DD Month YYYY, (If using all digits time format there is no issue)
3. Open the calendar and select a date.
4. The date is reset to the current date.
5. Create a new event
6. Invalid date

- Version/Git revision: 11d666fe8
- Database Type: mysql
- Database Version:
- PHP Version: 5.6

BrickOzp:
date_parse can only parse english dates.
If the forum is in another language, convert
month names to english before convertion.
Also format dates in Post Event screen in user
defined format.

Fixes #6631

### DIFF
--- a/Sources/Calendar.php
+++ b/Sources/Calendar.php
@@ -113,7 +113,7 @@ function CalendarMain()
 	// Need a start date for all views
 	if (!empty($_REQUEST['start_date']))
 	{
-		$start_parsed = date_parse($_REQUEST['start_date']);
+		$start_parsed = date_parse(convertDateToEnglish($_REQUEST['start_date']));
 		if (empty($start_parsed['error_count']) && empty($start_parsed['warning_count']))
 		{
 			$_REQUEST['year'] = $start_parsed['year'];
@@ -130,7 +130,7 @@ function CalendarMain()
 	// Need an end date for the list view
 	if (!empty($_REQUEST['end_date']))
 	{
-		$end_parsed = date_parse($_REQUEST['end_date']);
+		$end_parsed = date_parse(convertDateToEnglish($_REQUEST['end_date']));
 		if (empty($end_parsed['error_count']) && empty($end_parsed['warning_count']))
 		{
 			$_REQUEST['end_year'] = $end_parsed['year'];
--- a/Sources/Subs-Calendar.php
+++ b/Sources/Subs-Calendar.php
@@ -1060,7 +1060,7 @@ function validateEventPost()
 		// The 2.1 way
 		if (isset($_POST['start_date']))
 		{
-			$d = date_parse($_POST['start_date']);
+			$d = date_parse(convertDateToEnglish($_POST['start_date']));
 			if (!empty($d['error_count']) || !empty($d['warning_count']))
 				fatal_lang_error('invalid_date', false);
 			if (empty($d['year']))
@@ -1070,7 +1070,7 @@ function validateEventPost()
 		}
 		elseif (isset($_POST['start_datetime']))
 		{
-			$d = date_parse($_POST['start_datetime']);
+			$d = date_parse(convertDateToEnglish($_POST['start_datetime']));
 			if (!empty($d['error_count']) || !empty($d['warning_count']))
 				fatal_lang_error('invalid_date', false);
 			if (empty($d['year']))
@@ -1579,7 +1579,7 @@ function setEventStartEnd($eventOptions = array())
 	// If some form of string input was given, override individually defined options with it
 	if (isset($start_string))
 	{
-		$start_string_parsed = date_parse($start_string);
+		$start_string_parsed = date_parse(convertDateToEnglish($start_string));
 		if (empty($start_string_parsed['error_count']) && empty($start_string_parsed['warning_count']))
 		{
 			if ($start_string_parsed['year'] != false)
@@ -1598,7 +1598,7 @@ function setEventStartEnd($eventOptions = array())
 	}
 	if (isset($end_string))
 	{
-		$end_string_parsed = date_parse($end_string);
+		$end_string_parsed = date_parse(convertDateToEnglish($end_string));
 		if (empty($end_string_parsed['error_count']) && empty($end_string_parsed['warning_count']))
 		{
 			if ($end_string_parsed['year'] != false)
@@ -1861,6 +1861,51 @@ function removeHolidays($holiday_ids)
 	updateSettings(array(
 		'calendar_updated' => time(),
 	));
+}
+
+/**
+ * Helper function to convert date string to english
+ * so that date_parse can parse the date
+ *
+ * @param string $date A localized date string
+ * @return string English date string
+ */
+function convertDateToEnglish($date)
+{
+	global $txt, $context;
+
+	if ($context['user']['language'] == 'english')
+		return $date;
+
+	$replacements = array_combine(array_map('strtolower', $txt['months_titles']), array(
+		'January', 'February', 'March', 'April', 'May', 'June',
+		'July', 'August', 'September', 'October', 'November', 'December'
+	));
+	$replacements += array_combine(array_map('strtolower', $txt['months_short']), array(
+		'Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun',
+		'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'
+	));
+	$replacements += array_combine(array_map('strtolower', $txt['days']), array(
+		'Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'
+	));
+	$replacements += array_combine(array_map('strtolower', $txt['days_short']), array(
+		'Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'
+	));
+	// Find all possible variants of AM and PM for this language.
+	$replacements[strtolower($txt['time_am'])] = 'AM';
+	$replacements[strtolower($txt['time_pm'])] = 'PM';
+	if (($am = strftime('%p', strtotime('01:00:00'))) !== 'p' && $am !== false)
+	{
+		$replacements[strtolower($am)] = 'AM';
+		$replacements[strtolower(strftime('%p', strtotime('23:00:00')))] = 'PM';
+	}
+	if (($am = strftime('%P', strtotime('01:00:00'))) !== 'P' && $am !== false)
+	{
+		$replacements[strtolower($am)] = 'AM';
+		$replacements[strtolower(strftime('%P', strtotime('23:00:00')))] = 'PM';
+	}
+
+	return strtr(strtolower($date), $replacements);
 }
 
 ?>
--- a/Themes/default/Calendar.template.php
+++ b/Themes/default/Calendar.template.php
@@ -729,12 +729,12 @@ function template_calendar_top($calendar_data)
 
 	echo '
 			<form action="', $scripturl, '?action=calendar;', $context['calendar_view'], '" id="', !empty($calendar_data['end_date']) ? 'calendar_range' : 'calendar_navigation', '" method="post" accept-charset="', $context['character_set'], '">
-				<input type="text" name="start_date" id="start_date" maxlength="10" value="', $calendar_data['start_date'], '" tabindex="', $context['tabindex']++, '" class="date_input start" data-type="date">';
+				<input type="text" name="start_date" id="start_date" maxlength="10" value="', trim($calendar_data['start_date']), '" tabindex="', $context['tabindex']++, '" class="date_input start" data-type="date">';
 
 	if (!empty($calendar_data['end_date']))
 		echo '
 				<span>', strtolower($txt['to']), '</span>
-				<input type="text" name="end_date" id="end_date" maxlength="10" value="', $calendar_data['end_date'], '" tabindex="', $context['tabindex']++, '" class="date_input end" data-type="date">';
+				<input type="text" name="end_date" id="end_date" maxlength="10" value="', trim($calendar_data['end_date']), '" tabindex="', $context['tabindex']++, '" class="date_input end" data-type="date">';
 
 	echo '
 				<input type="submit" class="button" style="float:none" id="view_button" value="', $txt['view'], '">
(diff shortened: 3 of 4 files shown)


jdarwood007:
Haven't tested but seeing an array merge with the same keys this won't work right.
https://stackoverflow.com/questions/5881443/merging-arrays-with-the-same-keys

Sources/Calendar.php
```diff
@@ -464,6 +464,10 @@ function CalendarPost()
 		}
 	}
 
+	// Convert start and end date to user defined time format
+	$context['event']['start_date'] = timeformat((new DateTime($context['event']['start_date']))->getTimestamp(), get_date_or_time_format('date'));
+	$context['event']['end_date'] = timeformat((new DateTime($context['event']['end_date']))->getTimestamp(), get_date_or_time_format('date'));
+
```
Sesquipedalian: We can't do this, because changing the values of `$context['event']['start_date']` and `$context['event']['end_date']` can break some other calculations and comparisons that happen further down the line. 

Fortunately, we don't need to make this change anyway, because the localized date and time strings are already available in `$context['event']['start_date_local']`, `$context['event']['start_time_local']`, `$context['event']['end_date_local']`, and `$context['event']['end_time_local']`. So, in order to make properly localized date and time strings appear in the UI form, we just need to make the following replacements on the relevant lines in Post.template.php and Calendar.template.php:

`$context['event']['start_date']` -> `$context['event']['start_date_local']`
`$context['event']['start_time']` -> `$context['event']['start_time_local']`
`$context['event']['end_date']` -> `$context['event']['end_date_local']`
`$context['event']['end_time']` -> `$context['event']['end_time_local']`

Sources/Post.php
```diff
@@ -372,6 +372,11 @@ function Post($post_errors = array())
 			}
 		}
 
+		// Convert start and end date to user defined time format
+		$context['event']['start_date'] = timeformat((new DateTime($context['event']['start_date']))->getTimestamp(), get_date_or_time_format('date'));
+		$context['event']['end_date'] = timeformat((new DateTime($context['event']['end_date']))->getTimestamp(), get_date_or_time_format('date'));
+
+
```
Sesquipedalian: See previous comment.

Sources/Subs-Calendar.php
```diff
@@ -1851,4 +1851,21 @@ function removeHolidays($holiday_ids)
...
+ * @return string English date string
+ */
+function convertDateToEnglish($date)
+{
+	global $txt, $context;
+
+	if ($context['user']['language'] == 'english')
+		return $date;
+
+	return str_replace(array_merge($txt['months_titles'], $txt['months_short']), array_merge(array(1 => 'January', 2 => 'February', 3 => 'March', 4 => 'April', 5 => 'May', 6 => 'June', 7 => 'July', 8 => 'August', 9 => 'September', 10 => 'October', 11 => 'November', 12 => 'December'), array(1 => 'Jan', 2 => 'Feb', 3 => 'Mar', 4 => 'Apr', 5 => 'May', 6 => 'Jun', 7 => 'Jul', 8 => 'Aug', 9 => 'Sep', 10 => 'Oct', 11 => 'Nov', 12 => 'Dec')), $date);
```
```suggestion
	return strtr($date, array_combine(array_merge($txt['months_titles'], $txt['months_short']), array_merge(array('January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October', 'November', 'December'), array('Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'))));
```
Using `strtr()` instead of `str_replace()` will ensure that double replacements never occur.

Sesquipedalian (changes requested):

Sources/Subs-Calendar.php
BrickOzp: Thanks for the review, updated

Themes/default/Calendar.template.php
```diff
@@ -824,12 +824,12 @@ function template_event_post()
 						<div class="event_options_left" id="event_time_input">
 							<div>
 								<span class="label">', $txt['start'], '</span>
-								<input type="text" name="start_date" id="start_date" maxlength="10" value="', $context['event']['start_date'], '" tabindex="', $context['tabindex']++, '" class="date_input start" data-type="date">
+								<input type="text" name="start_date" id="start_date" maxlength="10" value="', $context['event']['start_date_local'], '" tabindex="', $context['tabindex']++, '" class="date_input start" data-type="date">
 								<input type="text" name="start_time" id="start_time" maxlength="11" value="', $context['event']['start_time_local'], '" tabindex="', $context['tabindex']++, '" class="time_input start" data-type="time"', !empty($context['event']['allday']) ? ' disabled' : '', '>
```
Sesquipedalian: I know I was the one who suggested `$context['event']['start_date_local']` and `$context['event']['end_date_local']` here and elsewhere, but it now occurs to me that `$context['event']['start_date_orig']` and `$context['event']['end_date_orig']` are actually the better values to use. The `_local` and `_orig` versions are identical in every way, except that the `_orig` version shows the date in the *event's* time zone, whereas the `_local` version shows the date in the *user's* time zone. Since we want to display the date and time of the event in its own time zone in this part of the UI, the `_orig` version is what we want.

For the same reason, we should also replace `start_time_local` and `end_time_local` with `start_time_orig` and `end_time_orig`.

Sorry to give you extra work. 🙂

Sources/Subs-Calendar.php
```diff
@@ -1851,4 +1851,21 @@ function removeHolidays($holiday_ids)
...
+ * @return string English date string
+ */
+function convertDateToEnglish($date)
+{
+	global $txt, $context;
+
+	if ($context['user']['language'] == 'english')
+		return $date;
+
+	return strtr($date, array_combine(array_merge($txt['months_titles'], $txt['months_short']), array_merge(array('January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October', 'November', 'December'), array('Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'))));
```
Sesquipedalian: It now occurs to me that it would be a good idea to perform substitutions for `$txt['days']` and `$txt['days_short']` here as well, just in case.

Sesquipedalian (changes requested):

Themes/default/Calendar.template.php
BrickOzp: Done

Sources/Subs-Calendar.php
BrickOzp: Done

Sesquipedalian (approved):

Sources/Subs-Calendar.php
```diff
@@ -1851,4 +1851,21 @@ function removeHolidays($holiday_ids)
...
+ * @return string English date string
+ */
+function convertDateToEnglish($date)
+{
+	global $txt, $context;
+
+	if ($context['user']['language'] == 'english')
+		return $date;
+
+	return strtr($date, array_combine(array_merge($txt['months_titles'], $txt['months_short'], $txt['days'], $txt['days_short']), array_merge(array('January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October', 'November', 'December'), array('Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'), array('Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'), array('Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'))));
```
frandominguezl: Super long line here. Can you split it to improve code readability?

sbulen:
I have been attempting to test this, with limited success.

I can reproduce the 'Invalid Date' error, in fact, on all languages, even English.  The problem is that I still see that error even with this PR applied.

Part of the problem, I think, is that we do not yet have the ability to test current RC4 code with current RC4 translations.  And there have been PRs merged to address CrowdIn issues in exactly these areas - arrays commonly used in the date processing.

We really need the ability to test this RC4 change with RC4 language translations.  

@jdarwood007 - please see my PM on this.  I think we can make some simple changes & gain the ability to download RC4 translations.

BrickOzp:
Could this also be a php 5.6 problem?

sbulen:
No, I saw invalid date issues on php 7.4.

My concern here is that we need to test the current code with the current (smf_2-1_next) language files.  Trying to test rc4 code with rc3 language files is problematic due to changes in this area.

sbulen:
I dummied up some "RC4" files to test with, and this is still returning an "Ungültiges Datum".

What I am seeing is that it is not converting the am/pm to English.  E.g., this is returned from convertDateToEnglish():
image


I'm using the default time format string of:

> %b %d, %Y, %I:%M %p

BrickOzp:
Tried a new approach instead https://github.com/SimpleMachines/SMF2.1/pull/6803

Sources/Subs-Calendar.php
```diff
@@ -1851,4 +1851,25 @@ function removeHolidays($holiday_ids)
...
+	global $txt, $context;
+
+	if ($context['user']['language'] == 'english')
+		return $date;
+
+	return strtr($date, array_combine(array_merge($txt['months_titles'], $txt['months_short'], $txt['days'], $txt['days_short']),
+		array_merge(array('January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October',	'November', 'December'),
+			array('Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'),
+			array('Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'),
+			array('Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'))));
```
```suggestion
	$replacements = array_combine($txt['months_titles'], array(
		'January', 'February', 'March', 'April', 'May', 'June',
		'July', 'August', 'September', 'October', 'November', 'December'
	));
	$replacements += array_combine($txt['months_short'], array(
		'Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun',
		'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'
	));
	$replacements += array_combine($txt['days'], array(
		'Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'
	));
	$replacements += array_combine($txt['days_short'], array(
		'Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'
	));

	$replacements[$txt['time_am']] = 'AM';
	$replacements[$txt['time_pm']] = 'PM';
	if (strftime('%p', strtotime('01:00:00')) !== 'p')
	{
		$replacements[strftime('%p', strtotime('01:00:00'))] = 'AM';
		$replacements[strftime('%p', strtotime('23:00:00'))] = 'PM';
	}
	if (strftime('%P', strtotime('01:00:00')) !== 'P')
	{
		$replacements[strftime('%P', strtotime('01:00:00'))] = 'AM';
		$replacements[strftime('%P', strtotime('23:00:00'))] = 'PM';
	}

	return strtr($date, $replacements);
```

Sesquipedalian (dismissed):
This should catch all variants of AM/PM that we expect for a given language.

Sources/Subs-Calendar.php
BrickOzp: Why do you need this part strftime('%p', strtotime('01:00:00')?
To get the system am/pm string? Is that string every used?

!== 'p' Will it return 'p' if it is not supported? or '%p'?

Sesquipedalian: > Why do you need this part strftime('%p', strtotime('01:00:00')?
> To get the system am/pm string? Is that string every used?

Yes, that is why. If you look at `timeformat()` in Subs.php, you will see that we only use `$txt['time_am']` and `$txt['time_pm']` as fallbacks in the event that `%p` is not supported in the current locale. If `%p` is supported, then we use the output of `strftime('%p')`.

> !== 'p' Will it return 'p' if it is not supported? or '%p'?

That is correct. It will return `p` if `%p` is not supported.

... Except on Windows, where `strftime()` will return false if given an unsupported format specifier. I forgot about that in my suggested code above. I'll add a new suggestion to deal with that in a review of your latest commit.

Sources/Subs-Calendar.php
```diff
@@ -1863,4 +1863,40 @@ function removeHolidays($holiday_ids)
...
+	));
+	$replacements += array_combine(array_map('strtolower', $txt['days_short']), array(
+		'Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'
+	));
+	$replacements[strtolower($txt['time_am'])] = 'AM';
+	$replacements[strtolower($txt['time_pm'])] = 'PM';
+
+	$replacements = array_map('strtolower', $replacements);
+
+	return strtr(strtolower($date), $replacements);
```
Sesquipedalian: 1. Why are you using `array_map` to apply `strtolower` to the replacement values? Applying `strtolower` to `$date` and to the replacement keys make sense, but why the values?
2. In order to catch all possible variants of "AM" and "PM" for the current language, I suggest the following. It might seem redundant, but it is necessary in order to catch all possible outcomes of the `timeformat()` function.
```suggestion
	// Find all possible variants of AM and PM for this language.
	$replacements[strtolower($txt['time_am'])] = 'AM';
	$replacements[strtolower($txt['time_pm'])] = 'PM';
	if (($am = strftime('%p', strtotime('01:00:00'))) !== 'p' && $am !== false)
	{
		$replacements[strtolower($am)] = 'AM';
		$replacements[strtolower(strftime('%p', strtotime('23:00:00')))] = 'PM';
	}
	if (($am = strftime('%P', strtotime('01:00:00'))) !== 'P' && $am !== false)
	{
		$replacements[strtolower($am)] = 'AM';
		$replacements[strtolower(strftime('%P', strtotime('23:00:00')))] = 'PM';
	}

	return strtr(strtolower($date), $replacements);
```

Sesquipedalian (changes requested):
Updated suggestion that deals with possibly unsupported AM/PM formats on Windows.
For explanation, see my reply to your questions in the comments on my previous suggestion, @BrickOzp.

BrickOzp:
I tested with German. In the time string their am/pm was all caps but the strings was not. So needed to convert to lower case for it to work.

sbulen:
I think we lost the colon fix in the latest logic:
image

BrickOzp:
> I think we lost the colon fix in the latest logic:

https://github.com/SimpleMachines/SMF2.1/pull/6637 isn't merged yet

Sources/Subs-Calendar.php
BrickOzp: 1. Yes, you are right. No point in converting the values.
2. Since we convert it lower case. Do we need to check both %P and %p? Aren't those just different cases?

Sesquipedalian (approved):

PR #6637, opened 2021-03-31
Remove trailing colon from event time

### LINKED ISSUE #6791: default_time_format not honoring seconds
#### Description
I see "Invalid Date" errors entering events into the calendar on my migrated test 2.1 forum.  

Then I noticed the times looked funky:
invalid-date-migrated-forum

It appears the :%S to include seconds in the time is not working.  This is the default time format:
image

Note the DB allows seconds, so I think the :%S should be valid:
image

### Steps to reproduce
1. Add :%S to the default time format
2. Add an event in the calendar

OR....

1. Add an event in the calendar on a migrated forum that had :%S specified (the old default)

- Database Type: mysql
- Database Version: 5.7
- PHP Version: 7.4

This is the **_2.0 default time format_** upon install - unless overridden, it specifies seconds:
image

This is the **_2.1 default time format_** upon install:
image

Not a big priority - workaround is easy, to remove the :%S.

BrickOzp:
When creating a new event, the time box had a trailing colon, like
9:10: or 9:10: am. Removed that.

### DIFF
--- a/Sources/Calendar.php
+++ b/Sources/Calendar.php
@@ -293,6 +293,8 @@ function CalendarPost()
 		'%T' => '%l:%M',
 	));
 
+	$time_string = preg_replace('~:(?=\s|$|%[pPzZ])~', '', $time_string);
+
 	// Submitting?
 	if (isset($_POST[$context['session_var']], $_REQUEST['eventid']))
 	{
--- a/Sources/Post.php
+++ b/Sources/Post.php
@@ -296,6 +296,8 @@ function Post($post_errors = array())
 			'%T' => '%l:%M',
 		));
 
+		$time_string = preg_replace('~:(?=\s|$|%[pPzZ])~', '', $time_string);
+
 		// Editing an event?  (but NOT previewing!?)
 		if (empty($context['event']['new']) && !isset($_REQUEST['subject']))
 		{
--- a/Sources/Subs-Calendar.php
+++ b/Sources/Subs-Calendar.php
@@ -1737,6 +1737,8 @@ function buildEventDatetimes($row)
 			'%T' => '%l:%M',
 		));
 
+	$time_format = preg_replace('~:(?=\s|$|%[pPzZ])~', '', $time_format);
+
 	// Should this be an all day event?
 	$allday = (empty($row['start_time']) || empty($row['end_time']) || empty($row['timezone']) || !in_array($row['timezone'], timezone_identifiers_list(DateTimeZone::ALL_WITH_BC))) ? true : false;
 


Sesquipedalian (approved):

sbulen:
What is very interesting is the row that maps %S to a space...   Do we get seconds if that is removed?

If the string specifies seconds, I think we should honor it instead of stripping it.

BrickOzp:
Made a quick test. Even if I show the seconds they will not get saved. Have not investigated why

BrickOzp:
Turns out seconds are stored in the db, but must get trimmed away on display.
Think we must decide what to do here. Always show seconds (if part of time format) or never show seconds (like now)

sbulen:
Testing this PR and I am still seeing the : when am/pm is specified:
image

This is what we're getting internally, so the trim doesn't work as-is - not on its own anyway:
image

sbulen:
My suggestion is to put back your original mapping of ':%S' to '', plus also keep the trim (a good safety measure since there is a lot of mapping items to '').

Sources/Calendar.php
```diff
@@ -293,6 +293,8 @@ function CalendarPost()
 		'%T' => '%l:%M',
 	));
 
+	$time_string = trim($time_string, ':');
```
Sesquipedalian:
```suggestion
	$time_string = preg_replace('~:(?=\D|$)~', '', $time_string);
```

This will remove any colon that is followed by a non-digit or the end of the string.

| Before | After |
|--------|-------|
| '9:45:00 am' | '9:45:00 am' |
| '9:45: am' | '9:45 am' |
| '9:45:am' | '9:45am' |
| '9:: am' | '9 am' |
| '9::am' | '9am' |
| '9:45:' | '9:45' |

Sources/Post.php
```diff
@@ -296,6 +296,8 @@ function Post($post_errors = array())
 			'%T' => '%l:%M',
 		));
 
+		$time_string = trim($time_string, ':');
```
Sesquipedalian:
```suggestion
		$time_string = preg_replace('~:(?=\D|$)~', '', $time_string);
```

Sources/Subs-Calendar.php
```diff
@@ -1737,6 +1737,8 @@ function buildEventDatetimes($row)
 			'%T' => '%l:%M',
 		));
 
+	$time_format = trim($time_format, ':');
```
```suggestion
	$time_format = preg_replace('~:(?=\D|$)~', '', $time_format);
```

Sesquipedalian (changes requested):
This `preg_replace()` should remove stray colons even in the middle of strings.

sbulen:
That'll work, too!

Sources/Calendar.php
BrickOzp: The time_string contains the time format, so no digits. But I think I managed to modify it to work.

Sources/Calendar.php
```diff
@@ -293,6 +293,8 @@ function CalendarPost()
 		'%T' => '%l:%M',
 	));
 
+	$time_string = preg_replace('~:(?=\s|$)~', '', $time_string);
```
Sesquipedalian: Only checking for a space or end of string could still produce `9:45:am`, `9:45:AM`, `9:45:EST`  or `9:45:-5000` in the output if the original time format contained something like `:%S%p`, `:%S%P`, `:%S%z`, or `:%S%Z` (no space after `%S`)

The following should fix that.
```suggestion
	$time_string = preg_replace('~:(?=\s|$|%[pPzZ])~', '', $time_string);
```

Sources/Post.php
```diff
@@ -296,6 +296,8 @@ function Post($post_errors = array())
 			'%T' => '%l:%M',
 		));
 
+		$time_string = preg_replace('~:(?=\s|$)~', '', $time_string);
```
Sesquipedalian:
```suggestion
		$time_string = preg_replace('~:(?=\s|$|%[pPzZ])~', '', $time_string);
```

Sources/Subs-Calendar.php
```diff
@@ -1737,6 +1737,8 @@ function buildEventDatetimes($row)
 			'%T' => '%l:%M',
 		));
 
+	$time_format = preg_replace('~:(?=\s|$)~', '', $time_format);
```
```suggestion
	$time_format = preg_replace('~:(?=\s|$|%[pPzZ])~', '', $time_format);
```

Sesquipedalian (changes requested):

Sources/Calendar.php
Sesquipedalian: Derp. You are right. See comments below.

Sources/Calendar.php
BrickOzp: Thanks

live627 (approved):

PR #6638, opened 2021-03-31
Correct previous and next month in calendar

BrickOzp:
Sometimes if selecting a date in the beginning or end of
the month, the previous and next month was incorrectly
calculated.
This was because we added or subtracted "1 month" to the
current date. PHP treats 1 month as 31 days. So if the current
month has less than 31 days the calculation is wrong.
Insted use "first day of next month" and
"first day of previous month".


BrickOzp:
More about this behavior in PHP https://stackoverflow.com/questions/3602405/php-datetimemodify-adding-and-subtracting-months

Sesquipedalian (approved):

Issue #6654, opened 2021-04-03
bbcode param mismatch—specific order required


live627:
```
1) PHPTDD\BBCTest::testBBC with data set #22 ('img', '[img alt=some text width=150 ...[/img]', '<img src="http://adomain.tld/...lazy">')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'<img src="http://adomain.tld/an_image.png" alt="some text" title="" width="150" height="100" class="bbc_img" loading="lazy">'
+'<img src="http://adomain.tld/an_image.png" alt="some text width=150 height=100" title="" class="bbc_img" loading="lazy">'
```
This is the bbcodfe in question
https://github.com/SimpleMachines/SMF2.1/blob/8eae9deee9fb0c69571d6654dfac2ecba5c65c2d/Sources/Subs.php#L1746-L1766

And my est inputs are

			array(
				'img',
				'[img alt=some text width=150 height=100]http://adomain.tld/an_image.png[/img]',
				'<img src="http://adomain.tld/an_image.png" alt="some text" title="" width="150" height="100" class="bbc_img resized" loading="lazy">',
			),

This is phpunit. The test method is

	/**
	 * @dataProvider bbcProvider
	 *
	 * @return void
	 */
	public function testBBC($tag, string $test, string $expected): void
	{
		$result = parse_bbc($test);
		$this->assertEquals($expected, $result);

		$this->assertContains($tag, array_column(parse_bbc(false), 'tag'));
	}

https://github.com/live627/SMF2.1/blob/1fdf397dbabc1088b95f03a309a2e4a2939d0465/tests/BBCTest.php#L475-L486

A small debugging session revealed that it matched to the wrong code  it seems like
https://github.com/SimpleMachines/SMF2.1/blob/8eae9deee9fb0c69571d6654dfac2ecba5c65c2d/Sources/Subs.php#L1727-L1745

				if (isset($possible['parameters']['alt']))
				var_dump($matches, $possible['parameters']); 

https://github.com/live627/SMF2.1/blob/7c9a29eef594fb2118d2cf1cfbb76cd4fc57447b/Sources/Subs.php#L2782-L2783

The relevant dump is
```
/home/travis/build/live627/SMF2.1/Sources/Subs.php:2783:
array(3) {
  [0] =>
  string(35) " alt=some text width=150 height=100"
  [1] =>
  string(35) " alt=some text width=150 height=100"
  [2] =>
[...]

PR #6675, opened 2021-04-08
Support negated regex tests for bbcodes

live627:
This is the polar opposite of`test` where it will skip a bbcode if there is a match. 'negative_test` searches the same string as `test`, but doesn't add an anchor.

This is my fix for #6654—a negative lookahead may be possible here but I cannot get it to work. I doubt that regular expressions were made to do that.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -1727,7 +1727,11 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 			array(
 				'tag' => 'img',
 				'type' => 'unparsed_content',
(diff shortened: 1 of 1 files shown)


live627:
This callback method seems like a better idea (also more flexible to boot!)

Sesquipedalian:
I don't think this is necessary. The current `test` regex checks already support negative lookaheads. For example, if `test` is set to `(?![^\]]*?\b(?:width|height)\b)`, then the test will fail if the BBC contains `width` or `height` in its parameter string.

live627:
Closed in favor of #6717

PR #6689, opened 2021-04-12
Backup karma data

BrickOzp:
If selected to delete karma data during upgrade, the karma
data is removed before the database backup is made.
Move deletion of karma data and error log until after backup
is made.

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -1299,55 +1299,10 @@ function UpgradeOptions()
 		);
 
 	// Deleting old karma stuff?
-	if (!empty($_POST['delete_karma']))
-	{
-		// Delete old settings vars.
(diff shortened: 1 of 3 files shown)


MissAllSunday (approved):

jdarwood007:
The truncate on error logs should be left.  So any errors that occurred during the upgrade can be captured incase they contain relevant info.

MissAllSunday:
If I understood correctly the truncate on the log error table wasn't removed but relocated, is that what you were refering to @jdarwood007 ?

BrickOzp:
Yes, like the karma the error log deletion is moved from the UpgradeOptions step to the beginning of DatabaseChanges. The purpose is to not do and changes to the db before it has been backed up.

PR #6691, opened 2021-04-14
Fix autolink when using parentheses

### LINKED ISSUE #6685: the message body was left empty
#### Description
created topic with subject test and body https://en.wikipedia.org/wiki/Kotlin_(programming_language)
clicked on post and was greeted with the message "The message body was left empty."

### Steps to reproduce
1. create new topic, subject test, body https://en.wikipedia.org/wiki/Kotlin_(programming_language)
2. click post see message "The message body was left empty."

- Version/Git revision: as of 12am EDT last night
- Database Type: mariadb
- Database Version: 10.4.10
- PHP Version: 7.3.12

BrickOzp:
When using parentheses in an url, sometimes the
autolinking did not work.

Fixes #6685

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -2447,15 +2447,6 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 							'(?:' .
 								// One or more of:
 								'(?:' .
-									// a run of non-space, non-()<>
-									'[^\s()<>]+' .
-									// or
-									'|' .
-									// balanced parentheses, up to 2 levels
-									'\(([^\s()<>]+|(\([^\s()<>]+\)))*\)' .
-								')+' .
-								// Ending with:
-								'(?:' .
 									// balanced parentheses, up to 2 levels
 									'\(([^\s()<>]+|(\([^\s()<>]+\)))*\)' .
 									// or
@@ -2466,7 +2457,14 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 									'|' .
 									// a trailing slash (but not two in a row)
 									'(?<!/)/' .
-								')' .
+								')+' .
+								// Query, and fragment components
+								'(?:' .
+									// query must start with ?, fragment with #
+									'[?#]' .
+									// Followed by any non space characters
+									'[^\s]+' .
+								')?' .
 							')?' .
 						')?';
 


BrickOzp:
Does anyone have any unit test for the auto linker? To see if this breaks anything.

BrickOzp:
This doesn't catch querys

BrickOzp:
Now querys and fragments should work as well.
example.com?test=test and example.com#test did not work before this fix, but do now.

Why are we so picky about matching parantheses?
https://en.wikipedia.org/wiki/Kotlin_(programming_language does not work
but https://en.wikipedia.org/wiki/Kotlin_programming_language) do

Sesquipedalian:
> Now querys and fragments should work as well.
> example.com?test=test and example.com#test did not work before this fix, but do now.

I would argue that those should not work in the autolinker. Yes, technically `http://example.com?test=test` is a valid URL according to RFC 3986, but the goal of an autolinker is not to detect all valid URLs. The goal of an autolinker is to guess correctly whether the human user _intended_ a given string to be understood as a URL.

Heuristically, when we find a domain name immediately followed by a string that matches the syntax of a query or fragment in some user input, it is more likely that this is the result of user error than of user intent. There are two reasons for this:
1. It is easy for a human to accidentally produce a string that matches this format. For example, it could happen if the user simply forgets to press space at the end of a sentence, like this: `Have you been to example.com?It is exemplary!`
2. It is unusual for a URL to be formatted like that intentionally by a human. As a rule, browsers and such autocorrect `http://example.com?test=test` to the explicitly unambiguous `http://example.com/?test=test`, consistently reinforcing the impression that the latter form is the "right" way.

Between these two factors, it is better to assume that a string like `example.com?It` in user input is not intended to be a URL with a query. This is why GitHub's autolinker did not parse those strings as URLs in your comment above, and SMF shouldn't either.

> Why are we so picky about matching parantheses?
> https://en.wikipedia.org/wiki/Kotlin_(programming_language does not work
> but https://en.wikipedia.org/wiki/Kotlin_programming_language) do

The idea there is to avoid including the closing parenthesis in a case such as `There are examples available for that (see http://example.com/foo) if you are interested.` Again, this is about the heuristics of guessing user intent, rather than about identifying all valid URLs. It would be weird and unusual for a URL to contain unbalanced parentheses, but it is quite common for a URL to be immediately followed by one in user input.

BrickOzp:
Thanks for the explanation @Sesquipedalian.
This regex is not easy to change without knowing if you break something else.

jdarwood007:
This regex is a good one to put lots of testing examples in.  Oneday when we implant proper unit testing, we can convert those over.

Sesquipedalian:
FYI, when I started running some tests and tweaking things, I ended up with some code that I will submit in a PR soon. So, don't waste your time working on this any further, @BrickOzp. :)

BrickOzp:
Nice!

PR #6692, opened 2021-04-14
Add workaround if auto linker returns nothing

BrickOzp:
If a message contains an url that the auto linker can't parse
sometimes the whole message becomes empty. Add a workaround
to restore the message if that happens.

Workaround for #6685

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -2470,7 +2470,9 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 							')?' .
 						')?';
 
-						$data = preg_replace_callback('~' . $url_regex . '~i' . ($context['utf8'] ? 'u' : ''), function($matches)
+						$tmp_data = $data;
+
+						$tmp_data = preg_replace_callback('~' . $url_regex . '~i' . ($context['utf8'] ? 'u' : ''), function($matches)
 						{
 							$url = array_shift($matches);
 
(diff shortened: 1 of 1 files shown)


jdarwood007:
I would rather see the proper fix than a workaround.

BrickOzp:
Generally I agree with you. I have tried to fix this particular url in https://github.com/SimpleMachines/SMF2.1/pull/6691. But if there are one that will break the regex, there can be more urls that breaks it that we do not know about. And I think it is better to display the post without auto linking than totally empty in those cases. That's why I also submitted this PR.

Sources/Subs.php
```diff
@@ -2466,6 +2466,8 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 							')?' .
 						')?';
 
+						$tmp_data = $data;
+
 						$data = preg_replace_callback('~' . $url_regex . '~i' . ($context['utf8'] ? 'u' : ''), function($matches)
```
Sesquipedalian:
```suggestion
						$tmp_data = preg_replace_callback('~' . $url_regex . '~i' . ($context['utf8'] ? 'u' : ''), function($matches)
```

Sources/Subs.php
```diff
@@ -2499,6 +2501,9 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 						}, $data);
 					}
 
+					if (empty($data))
+						$data = $tmp_data;
```
Sesquipedalian:
```suggestion
					if (!is_null($tmp_data))
						$data = $tmp_data;
```

Sesquipedalian (changes requested):
Using `is_null()` will be more precise than `empty()` in this case. 
Should also use the same logic in the email autolinker a few lines below, just in case that regex ever fails with an error.

Sources/Subs.php
```diff
@@ -2470,7 +2470,9 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 							')?' .
 						')?';
 
-						$data = preg_replace_callback('~' . $url_regex . '~i' . ($context['utf8'] ? 'u' : ''), function($matches)
+						$tmp_data = $data;
```
Sesquipedalian, line 2473: I don't think this line serves any purpose now.

Sesquipedalian (changes requested):

PR #6703, opened 2021-04-19
Update txt string icons_filename_all_png

### LINKED ISSUE #6657: Phrasing of $txt['icons_filename_all_{png,gif}']
https://github.com/SimpleMachines/SMF2.1/blob/40e0e3910096b4d72fa7923a0187e5855f6d6a05/Themes/default/languages/ManageSmileys.english.php#L96-L97

I'd suggest either:
```
-All files
-All icons
```

And/Or:
```
-&quot;[...]&quot; files';
+of type &quot;[...]&quot;'; 
```

BrickOzp:
icons_filename_all_gif is not used anymore so it is removed.
Update icons_filename_all_png to use sprintf for future proofing.

Fixes #6657

### DIFF
--- a/Themes/default/languages/ManageSmileys.english.php
+++ b/Themes/default/languages/ManageSmileys.english.php
@@ -93,8 +93,8 @@ $txt['icons_edit_icon'] = 'Edit message icon';
 $txt['icons_new_icon'] = 'New message icon';
 $txt['icons_location_first_icon'] = 'As first icon';
 $txt['icons_location_after'] = 'After';
-$txt['icons_filename_all_gif'] = 'All files must be &quot;gif&quot; files';
-$txt['icons_filename_all_png'] = 'All files must be &quot;png&quot; files';
+// argument(s): the supported file extension
+$txt['icons_extension_must_be'] = 'File extension must be "%1$s"';
 $txt['icons_no_entries'] = 'There are currently no message icons configured.';
 
 ?>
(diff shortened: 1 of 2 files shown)


m4z:
Have you checked that the type is actually verified? Some related strings made me think the check might actually only look at the file extension.

BrickOzp:
Yes I did. And only the file extension is checked. I did change a gif file to .png and it was displayed correctly. But this is a limitation of the database, maybe some legacy stuff. The file extension is not stored in the db and is added by the code. That's why the file needs to be .png

Themes/default/languages/ManageSmileys.english.php
```diff
@@ -93,8 +93,7 @@
 $txt['icons_new_icon'] = 'New message icon';
 $txt['icons_location_first_icon'] = 'As first icon';
 $txt['icons_location_after'] = 'After';
-$txt['icons_filename_all_gif'] = 'All files must be &quot;gif&quot; files';
-$txt['icons_filename_all_png'] = 'All files must be &quot;png&quot; files';
+$txt['icons_filename_all_png'] = 'Files must be of type "png"';
```
m4z: Then this phrasing is misleading, no? The *extension* must be PNG, not the "type" (which could refer to either the image type or the extension).

BrickOzp: Yes, you are right. I will change

jdarwood007:
Just a thought.  Why not pass this through sprintf and ensure its future proofed?

```
$txt['extension_must_be'] = 'File extension must be "%1$s"';
sprintf($txt['extension_must_be'], '.png');
```

MissAllSunday (approved):

PR #6711, opened 2021-04-23
Correctly handle spider actions

BrickOzp:
The function determineActions can return an array.
That was not handled in the spider log list, resulting
in Array to string conversion errors.
Handle returned arrays as well as strings.


BrickOzp:
Error reported here https://www.simplemachines.org/community/index.php?topic=577205.msg4087134#msg4087134

MissAllSunday (approved):

PR #6721, opened 2021-04-30
Remove the option to set height

### LINKED ISSUE #6541: Resized images cause distortion

BrickOzp:
Remove the option to set height for images and embedded
attachments. Having a set height when the width is
responsive, causes aspect ratio to be wrong in some cituations.

Fixes #6541

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -1438,38 +1438,35 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 						// Image.
 						if (!empty($currentAttachment['is_image']))
 						{
-							if (empty($params['{width}']) && empty($params['{height}']))
+							if (empty($params['{width}']))
 								$returnContext .= '<img src="' . $currentAttachment['href'] . '"' . $alt . $title . ' class="bbc_img"></a>';
 							else
 							{
 								$width = !empty($params['{width}']) ? ' width="' . $params['{width}'] . '"': '';
-								$height = !empty($params['{height}']) ? 'height="' . $params['{height}'] . '"' : '';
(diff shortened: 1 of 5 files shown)


BrickOzp:
This is a radical idea.
See investigation in https://github.com/SimpleMachines/SMF2.1/issues/6541

Sources/Subs.php
```diff
@@ -1750,9 +1747,9 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 					'alt' => array('optional' => true),
 					'title' => array('optional' => true),
 					'width' => array('optional' => true, 'value' => ' width="$1"', 'match' => '(\d+)'),
-					'height' => array('optional' => true, 'value' => ' height="$1"', 'match' => '(\d+)'),
+					'height' => array('optional' => true, 'value' => '', 'match' => '(\d+)'),
 				),
-				'content' => '<img src="$1" alt="{alt}" title="{title}"{width}{height} class="bbc_img resized" loading="lazy">',
+				'content' => '<img src="$1" alt="{alt}" title="{title}"{width} class="bbc_img resized" loading="lazy">',
```
albertlast: google and user didn't like lazy load without height.

BrickOzp: Fixed

DiegoAndresCortes:
It's alright imo.
I provided my thoughts on the other pr, so whatever is decided it's fine by me, we just need to be aware that it needs feedback from the average users to understand what behavior they expect.

PR #6722, opened 2021-04-30
Remove size restriction of resized images on mobile

BrickOzp:
Resized images are only allowed to be 250 by 300 pixels
on mobile phones. There is no reason to do so, since not
resized images can be bigger and it can mess up the aspect
ratio of the image.

### DIFF
--- a/Themes/default/css/responsive.css
+++ b/Themes/default/css/responsive.css
@@ -242,10 +242,6 @@
 	.keyinfo h5 .modified {
 		float: none;
 	}
-	.bbc_img.resized {
-		max-width: 250px;
-		max-height: 300px;
-	}
 	img.icon, #forumposts .catbg img {
 		display: none;
 	}


DiegoAndresCortes:
I believe I mentioned that there's a situation where an image/attachment was resized and it should stay like that in phones imo, as the author of the post maybe intended it to be resized.

jdarwood007:
If the bbc tag is setting them, then wouldn't that mean they are defined in the img tag themselves and not via css, thus they would override the call?

Images are supposed to have a defined size to prevent the page from "rubber banding" as it loads content.

DiegoAndresCortes:
I don't remember how it worked, maybe it's the javascript that caused it to overwrite when the image would get expanded/resized.

BrickOzp:
When posting an image och attachment and adding width and/or height the resized class is added to the image.
When clicking on the image it will toggle to original size.

On small screens however the resized images are restricted to max-width: 250px and max-height: 300px
But not resized images are not.

I don't see the point in restricting only resized images when they will resize automatically up to the defined width/height or the width of the available space, which is less.

jdarwood007:
FYI, This does not fix #6541 
This is related to something else?

BrickOzp:
It is somewhat related.
If you attach a portrait image and set width to 500px. When a width or height is set it will be considered resized.
View the post in a mobile, or just resize the browser window.
The image will have incorrect aspect ratio.

Now attach the image again. without setting width or height,
View it in a mobile or resize the browser.
The image is responsive and aspect ratio is correct.

Sesquipedalian (approved):

PR #6723, opened 2021-05-03
Fix text selection in package manager

BrickOzp:
When viewing the search and replace popup the text
selection link does not work. This is because the
js files are not loaded properly.


BrickOzp:
Problem reported here https://www.simplemachines.org/community/index.php?topic=576866.0

live627 (approved):

PR #6724, opened 2021-05-03
Break long attachment dir

BrickOzp:
If the attachment directory is to long in the Attachment Settings,
a scrollbar is shown for whole block of settings.
Add a zero width space after all directory separators to allow
breaking.

### DIFF
--- a/Sources/ManageAttachments.php
+++ b/Sources/ManageAttachments.php
@@ -2381,6 +2381,7 @@ function ManageAttachmentPaths()
 					'data' => array(
 						'db' => 'path',
 						'style' => 'width: 45%;',
+						'class' => 'word_break',
 					),
 				),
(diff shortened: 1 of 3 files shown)


Sesquipedalian:
I'd prefer to see a CSS solution, rather than altering the text content in the paths.

The reason is that if the admin copies and pastes the text of the path, the zero width spaces will be included in the copied text and the path won't work when pasted elsewhere.

I think it would work to add `class="word_break"` to the affected `<dd>` elements, or maybe a custom-made class if you prefer.

BrickOzp:
I did try to fix it in css but word_break did not break at / so didn't look quite as nice. But maybe it is better anyway. I will change.

live627 (approved):

PR #6725, opened 2021-05-03
Add specific id to responsive css selector

BrickOzp:
A specific id was missing from one of the responsive
css selectors. This is probably not on purpose.
Without this id many settings in the admin menu only
uses half the width when viewed on screens with widths
below 480 px.


BrickOzp:
Before
before

After
after

MissAllSunday (approved):

PR #6728, opened 2021-05-05
Fix attachment images when printing topics

BrickOzp:
* When printing without images, attachments where shown
as images. Disabled attachments when printing without
images.
* Hide attachments below posts if they are included
in the post, if the setting "dont_show_attach_under_post"
is enabled.


BrickOzp:
Reported here
https://www.simplemachines.org/community/index.php?topic=577205.msg4087833#msg4087833

MissAllSunday (approved):

PR #6732, opened 2021-05-06
Show original attachment during preview

BrickOzp:
When previewing a post with an inline attachment,
only the thumbnail of the attachment was shown.
Show the original attachment, like how it will be
shown in the posted message.


Sources/ShowAttachments.php
```diff
@@ -291,7 +291,6 @@ function showAttachment()
 	// Does this have a mime type?
 	elseif (!empty($file['mime_type']) && (isset($_REQUEST['image']) || !in_array($file['fileext'], array('jpg', 'gif', 'jpeg', 'x-ms-bmp', 'png', 'psd', 'tiff', 'iff'))))
 		header('content-type: ' . strtr($file['mime_type'], array('image/bmp' => 'image/x-ms-bmp')));
-
```
MissAllSunday: For better readability please leve the empty line separating the conditionals.

BrickOzp: Done

PR #6744, opened 2021-05-14
Fixes to several mentions issues

### LINKED ISSUE #5082: Possible exploits in mentions?
One problem in the 2.0 mentions mod is that you are able to mention an unlimited number of people in the same post. It could be used on purpose to harm the servers (besides the obvious social problem of people mentioning everyone), since you can also edit the post and remove the mentions, and do it many times over, and the mentions notifications still go through upon saving. Without an edit history you also have effectively no way of knowing this is being done without hunting for the data in user's profiles or similar.

Do we have anything in place in 2.1 to prevent this becoming too much of a problem? I'm thinking even a warning in the logs that a post contains a certain amount of mentions could be enough, without making it too complicated with limiters and so on.

**PS: This is based on how the mod for 2.0 works, which was made very similarly to 2.1 mentions. But it should be looked onto whether the same issue exists or not in 2.1 as well. I don't have enough content in my test forum to test it.**

### LINKED ISSUE #6648: Mentions never logged when creating a new post
#### Description
When a new post is created that mentions a member of the forum, that mention is never logged in the mentions table and no notifications or alerts are generated.

### Steps to reproduce
1. Create a new post that mentions a member of the forum using the `@username` syntax.
2. Check your database to see that no record of the mention is created and no notifications are sent.

- Version/Git revision: 505450e
- Database Type: N/A
- Database Version: N/A
- PHP Version: N/A

The problem is here:
https://github.com/SimpleMachines/SMF2.1/blob/505450e24f590b3ba1ef1bfbbd8b0bd27178228c/Sources/Subs-Post.php#L1810
The arguments for `strpos()` are in the wrong order, and the '@' character not be in the test string.

I will submit a fix soon.

### LINKED ISSUE #6649: Behaviour of mention notifications for modified posts wrongly depends on search index settings
The `modifyPost()` function of Subs-Post.php includes this chunk of logic that populates the `$oldmentions` array:
https://github.com/SimpleMachines/SMF2.1/blob/505450e24f590b3ba1ef1bfbbd8b0bd27178228c/Sources/Subs-Post.php#L2192-L2204
The `$oldmentions` array is then used to determine which mentions (if any) are new in the modified version of the post, so that we can avoid sending redundant notifications to members who were already notified about mentions in the old version of the post.

The problem is that `$msgOptions['old_body']` will only be defined if the code inside this preceding if block has been executed:
https://github.com/SimpleMachines/SMF2.1/blob/505450e24f590b3ba1ef1bfbbd8b0bd27178228c/Sources/Subs-Post.php#L2151-L2164
This makes the behaviour of mention notifications in modified posts dependent on the value of `$modSettings['search_custom_index_config']`, which means that these notifications will not behave correctly unless the admin has configured the search settings to use a custom index.

I will submit a fix for this soon.

### LINKED ISSUE #6650: Mentions don't support non-ASCII characters well
Some functions and regular expressions used in `Mentions::getPossibleMentions()` to find mentioned usernames don't play well with non-ASCII characters, potentially creating problems detecting usernames in languages that don't use the Latin alphabet. In particular, these lines are problematic:
https://github.com/SimpleMachines/SMF2.1/blob/505450e24f590b3ba1ef1bfbbd8b0bd27178228c/Sources/Mentions.php#L195
https://github.com/SimpleMachines/SMF2.1/blob/505450e24f590b3ba1ef1bfbbd8b0bd27178228c/Sources/Mentions.php#L226

I will submit a fix soon.

### LINKED ISSUE #6651: Nested quote BBC not filtered correctly in Mentions::getPossibleMentions()
In Mentions::getPossibleMentions(), there is an attempt made to filter out `[quote]` BBCodes so that mentions inside of quotations do not prompt double notifications:
https://github.com/SimpleMachines/SMF2.1/blob/505450e24f590b3ba1ef1bfbbd8b0bd27178228c/Sources/Mentions.php#L190-L192

This regular expression fails when there are nested quotations. For example, if `$body` contains the following string:
```
This is text.
[quote author=foo]Quoted level 1.
[quote]Quote level 2.[/quote]
A @mention in a quote.
[/quote]
End of post.
```
... then the code in question will result in this:
```
This is text.

A @mention in a quote.
[/quote]
End of post.
```
... which obviously isn't the intended result.

I will submit a fix for this soon.

Sesquipedalian:
- Fixes #6648: Aligns test with expected output of Mentions::getBody()
- Fixes #6650: Improves Unicode support in Mentions::getPossibleMentions()
- Fixes #6651: Correctly filters out nested quotes in Mentions::getPossibleMentions()
- Fixes #6649: Simplifies and improves handling of mention notifications for modified posts; Handles quote notifications more like mentions.
- Fixes #5082: Prevents "mention spam."

----------

The last one is the major change. It includes the following:
- Refactors CreatePost-Notity.php to be more OOP-y
- Imposes a delay before sending email notifications about mentions and quotes, and verifies that the quotes and mentions still exist in the post at the end of the delay before sending the email notifications.

The summarized logic flow for the last part of that (imposes a delay and verifies before sending) is as follows.

### When it is time to send out alerts and notifications about a **new** post:
1. Get all the members who are mentioned or quoted in this post.
2. Verify that each member is in fact still mentioned or quoted in the post, and filter out any who are not.
3. Get all the members who are watching for new posts in this topic (or new topics in this board, if the post starts a new topic).
4. Send the appropriate alerts to all the mentioned, quoted, and watching members.
5. Send email notifications to watching members.
    - If a watching member is also a mentioned or quoted member, we send the mention or quote email notification instead of the basic watch notification.
6. If there were any mentioned or quoted members that were not also watching for this post, respawn this task with the following changes:
    - `$msgOptions['mentioned_members']` and `$msgOptions['quoted_members']` only include members that want email notifications but have not yet been sent any about this post.
    - The `claimed_time` for the respawned task is set so that the task will not run until 5 minutes into the future.
#### Then when the second iteration of the task runs:
7. Repeat step 1.
8. Repeat step 2.
9. Filter out any members that have already received email notifications.
10. Send the email notifications to the mentioned and quoted members.

### When it is time to send out alerts and notifications about an **edited** post:
1. Get the mentioned or quoted members of any **new** mentions or quotes in this post.
2. Verify that each member is in fact still mentioned or quoted in the post, and filter out any who are not.
3. Delete mention/quote alerts for any members that are **no longer** mentioned/quoted in this post.
4. Send the appropriate alerts for the **new** mentions and quotes.
5. Respawn this task with the following changes:
    - The `claimed_time` for the respawned task is set so that the task will not run until 5 minutes into the future.
#### Then when the second iteration of the task runs:
6. Repeat step 1.
7. Repeat step 2.
8. Send the email notifications to the mentioned and quoted members.

----------

A secondary benefit of using this method where the task respawns with a delay is that the CreatePost_Notify_Background class is now more resilient against temporary errors while trying to send emails. Previously, if anything went wrong trying to send an email notification, we just gave up immediately on both sending the email *and* on sending an alert:
https://github.com/SimpleMachines/SMF2.1/blob/7a7e13d97372542ce5ea8b7665bae0f10cefb7a6/Sources/tasks/CreatePost-Notify.php#L297-L299
Now we do have "a way to attempt to resend just the email." If anything goes wrong, the email address of the failed send attempt is retained and passed on to the respawned task so that we can try again. (There is a maximum of 10 respawns allowed, so this won't produce an infinite loop if the send attempt failed because of a permanent error.)


Sources/Mentions.php
```diff
@@ -234,6 +301,137 @@ protected static function getPossibleMentions($body)
...
+					$message .= '[/quote]';
+				if ($quote_level >= 1)
+				{
+					$quote_level--;
+					$message .= "\n";
+				}
+			}
+			elseif ($quote_level <= 1)
+				$message .= $block;
+		}
```
live627, line 399: What is the purpose of this code ? I can't wrap my head around it.

Sesquipedalian: This is just copied straight over from its previous location in CreatePost-Notify.php. It appears to be walking through all matches from the regex above in order to make sure nested quotes do not mess it up. The whole thing could probably be replaced with a properly recursive regex like added elsewhere in this PR, but I didn't want to mess with it when it currently isn't broken.

live627 (approved):

Sources/tasks/CreatePost-Notify.php
```diff
@@ -152,71 +248,263 @@ public function execute()
...
+								content_action = {literal:mention}
+								AND id_member IN ({array_int:members_not_mentioned})
+							)
+						)',
+					array(
+						'msg_id' => $msg_id,
+						'members_not_quoted' => $old_alerts['quote'],
+						'members_not_mentioned' => $old_alerts['mention'],
+					)
+				);
```
live627: Looks like this will flail if either of these arrays is empty.

Sesquipedalian: Good catch.

Sesquipedalian:
Look good now, @live627?

live627 (approved):
Yes. Also, in some future version, moving notification triggers to a new function called from `Post2()` (the controller) may be a good idea (separation of concerns and all that jazz).

PR #6746, opened 2021-05-19
Remove hover action for disabled continue button in upgrader

BrickOzp:
Hovering the continue button in the upgrader when the
button was disabled, turned the text orange. Use this style
only when the button is enabled, to not confuse the user.


sbulen:
This seems to work for Chrome, but not for Firefox.  Are you aware of any tweak that will work for both?

BrickOzp:
It works for me in firefox. I have simplified it a bit. Could you test it again?

sbulen (approved):
Looks good in FF, Chrome, Pale Moon!

PR #6759, opened 2021-05-28
Consolidates sm_temp_dir_option() into sm_temp_dir()

Sesquipedalian:
This is just a bit of minor tidying to get rid of an unnecessary extra function. There are no changes to behaviour.

### DIFF
--- a/Sources/Subs-Admin.php
+++ b/Sources/Subs-Admin.php
@@ -2224,6 +2224,8 @@ function emailAdmins($template, $replacements = array(), $additional_recipients 
 */
 function sm_temp_dir()
 {
+	global $cachedir;
+
 	static $temp_dir = null;
 
 	// Already did this.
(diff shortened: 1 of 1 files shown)


Sources/Subs-Admin.php
```diff
@@ -2250,7 +2248,23 @@ function sm_temp_dir()
 	// Search for a working temp directory.
 	foreach ($temp_dir_options as $id_temp => $temp_option)
 	{
-		$possible_temp = sm_temp_dir_option($temp_option);
+		switch ($temp_option) {
+			case 'cachedir':
+				$possible_temp = rtrim($cachedir, '/');
```
BrickOzp, line 2255: Missing global $cachedir?

Sesquipedalian: Good catch.

PR #6761, opened 2021-05-28
Avoids mangling the RSS feed's filename

### LINKED ISSUE #6760: Invalid encoding of RSS feed name for non-latin utf-8 characters
#### Description
Example of invalid feed name:
```
Ð_Ð¾Ñ_Ñ_Ð¼_-recent-rss.xml
```

### Steps to reproduce
1. make forum title in non-latin utf-8 characters, for example: ФОРУМ (FORUM in russian)
2. click RSS link: /index.php?action=.xml;type=rss
3. check xml filename, it contains invalid utf-8 characters: Ð_Ð¾Ñ_Ñ_Ð¼_-recent-rss.xml

- Version/Git revision: RC3

Sesquipedalian:
Fixes #6760

### DIFF
--- a/Sources/News.php
+++ b/Sources/News.php
@@ -263,7 +263,7 @@ function ShowXmlFeed()
 	elseif (!empty($board))
 		$filename[] = 'board=' . $board;
 	$filename[] = $xml_format;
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Hm. Yes, you are right. We have a few lines of code in ShowAttachments.php that deal with this. I could copy that over, I suppose, or maybe move it into a little helper function so we don't have multiple copies scattered around the codebase.

PR #6764, opened 2021-06-02
Improve detection of global cookie settings

### LINKED ISSUE #6762: Installer - cookie setting defaults confused
#### Description
The default cookie setting logic attempts to set the local & global cookie settings and the global cookie domain based on the board url.

It is confusing multi-level tlds with subdomains, e.g., when it sees xxxxx.com.au, it will set global cookies to true and the domain to com.au.

This has come up a few times recently in the support forums:
https://www.simplemachines.org/community/index.php?topic=577727.msg4088673#msg4088673

This needs to be looked at:
https://github.com/SimpleMachines/SMF2.1/blob/ab409dffe23b960cfec2744f09c98734af6284fb/other/install.php#L1294

BrickOzp:
Use the Mozilla Foundation Public Suffix list
to improve detections of domain names using
country code second level domains
(like www.domain.com.au).

Fixes #6762

### DIFF
--- a/other/install.php
+++ b/other/install.php
@@ -1303,15 +1303,46 @@ function DatabasePopulation()
 		{
 			// www isn't really a subdomain in this sense, so strip it out
 			$url_parts['host'] = str_ireplace('www.', '', $url_parts['host']);
-			$pos1 = strrpos($url_parts['host'], '.');
-			if ($pos1 !== false)
+
+			// If host only contains one period there can't be any subdomain
+			if (substr_count($url_parts['host'], '.') > 1)
 			{
-				// 2nd period from the right indicates you have a subdomain
-				$pos2 = strrpos(substr($url_parts['host'], 0, $pos1 - 1), '.');
-				if ($pos2 !== false)
+				$allowed_tld = download_ccSLD_list();
+				if (!empty($allowed_tld))
 				{
-					$globalCookies = '1';
-					$globalCookiesDomain = substr($url_parts['host'], $pos2 + 1);
+					foreach($allowed_tld as $value)
+					{
+						$pattern = '/\.' . strtr($value, array('.' => '\.', '*' => '[^.]*?')) . '$/';
+						if (preg_match($pattern, $url_parts['host'], $matches, PREG_OFFSET_CAPTURE) === 1)
+						{
+							$stripped_domain = substr($url_parts['host'], 0, $matches[0][1]);
+							$tld = $matches[0][0];
+							break;
+						}
+					}
+
+					$pos1 = strrpos($stripped_domain, '.');
+					// If we find a period this indicates you have a subdomain
+					if ($pos1 !== false)
+					{
+						$globalCookies = '1';
+						$globalCookiesDomain = substr($stripped_domain, $pos1 + 1) . $tld;
+					}
+				}
+				else
+				{
+					// Fallback if ccSLD list can't be retrieved
+					$pos1 = strrpos($url_parts['host'], '.');
+					if ($pos1 !== false)
+					{
+						// 2nd period from the right indicates you have a subdomain
+						$pos2 = strrpos(substr($url_parts['host'], 0, $pos1 - 1), '.');
+						if ($pos2 !== false)
+						{
+							$globalCookies = '1';
+							$globalCookiesDomain = substr($url_parts['host'], $pos2 + 1);
+						}
+					}
 				}
 			}
 		}
@@ -1505,6 +1536,18 @@ function DatabasePopulation()
 	}
 
 	return false;
+}
+
+function download_ccSLD_list()
+{
+	// Download list of all known ccSLDs
+	$list = explode("\n", file_get_contents('https://publicsuffix.org/list/public_suffix_list.dat'));
+	$allowed_tld = array();
+	foreach ($list as $value)
+		// Remove all empty lines and comments
+		if (!empty($value) && substr($value, 0, 2) != '//')
+			$allowed_tld[] = $value;
+	return array_reverse($allowed_tld);
 }
 
 // Ask for the administrator login information.


sbulen:
Note that we have an existing function to get valid tlds from iana:
https://github.com/SimpleMachines/SMF2.1/blob/ab409dffe23b960cfec2744f09c98734af6284fb/Sources/Subs.php#L6652

It is stored in regex form in the settings table under tld_regex & used by SMF to validate tlds.  There is even some fallback logic just in case internet is not available.  

The challenge here is that this logic in ForumSettings is called before the DB is even populated...  I'm not sure our existing code can even be called here...  I haven't tried it yet...

sbulen:
**_After some thought, my preference is now to remove the existing default cookie setting logic entirely._**  Make the user set the cookie settings; don't second guess them. 

Another option might be to just set _local_ cookie settings if a valid subfolder is found, and leave any 'global' settings to the user.

I had started working on this & was going to use the following as my test script; you might find it helpful if we keep & enhance setting the defaults.  A list of things we need to consider...


// TESTS.  Basics & subdomains.  NA means doesn't exist.
https://funstuff.com
- localCookies:  NA
- globalCookies:  NA
- globalCookiesDomain:  NA

https://funstuff.com.au
- 	localCookies:  NA
- 	globalCookies:  NA
- 	globalCookiesDomain:  NA

https://www.funstuff.com
- 	localCookies:  NA
- 	globalCookies:  NA
- 	globalCookiesDomain:  NA

https://www.funstuff.com.au
- 	localCookies:  NA
- 	globalCookies:  NA
- 	globalCookiesDomain:  NA

https://support.funstuff.com
- 	localCookies:  NA
- 	globalCookies:  1
- 	globalCookiesDomain:  funstuff.com

https://support.funstuff.com.au
- 	localCookies:  NA
- 	globalCookies:  1
- 	globalCookiesDomain:  funstuff.com.au

// Unfortunately, tlds can actually be domain names.   'com.com' is a real site...
https://com.com
- 	localCookies:  NA
- 	globalCookies:  NA
- 	globalCookiesDomain:  NA

https://com.com.au
- 	localCookies:  NA
- 	globalCookies:  NA
- 	globalCookiesDomain:  NA

https://www.com.com
- 	localCookies:  NA
- 	globalCookies:  NA
- 	globalCookiesDomain:  NA

https://www.com.com.au
- 	localCookies:  NA
- 	globalCookies:  NA
- 	globalCookiesDomain:  NA

https://support.com.com
- 	localCookies:  NA
- 	globalCookies:  1
- 	globalCookiesDomain:  com.com

https://support.com.com.au
- 	localCookies:  NA
- 	globalCookies:  1
- 	globalCookiesDomain:  com.com.au

// local cookies
https://funstuff.com/forum
- 	localCookies:  1
- 	globalCookies:  NA
- 	globalCookiesDomain:  NA

https://funstuff.com.au/forum
- 	localCookies:  1
- 	globalCookies:  NA
- 	globalCookiesDomain:  NA

https://www.funstuff.com/forum
- 	localCookies:  1
- 	globalCookies:  NA
- 	globalCookiesDomain:  NA

https://www.funstuff.com.au/forum
- 	localCookies:  1
- 	globalCookies:  NA
- 	globalCookiesDomain:  NA

SMF normally treats local & global as mutually exclusive, not sure what it should do here... Assuming 'local':
https://support.funstuff.com/forum
- 	localCookies:  1
- 	globalCookies:  NA
- 	globalCookiesDomain:  NA

https://support.funstuff.com.au/forum
- 	localCookies:  1
- 	globalCookies:  NA
- 	globalCookiesDomain:  NA

other/install.php
```diff
@@ -1303,15 +1303,46 @@ function DatabasePopulation()
...
+					foreach($allowed_tld as $value)
+					{
+						$pattern = '/\.' . strtr($value, array('.' => '\.', '*' => '[^.]*?')) . '$/';
+						if (preg_match($pattern, $url_parts['host'], $matches, PREG_OFFSET_CAPTURE) === 1)
+						{
+							$stripped_domain = substr($url_parts['host'], 0, $matches[0][1]);
+							$tld = $matches[0][0];
+							break;
+						}
+					}
```
Sesquipedalian, line 1322: Right now, this requires running possibly thousands of `preg_match()` calls in order to test each ccSLD individually. If instead you use `build_regex()` from Subs.php to create a single regex that will match all possible ccSLD values, you will only need to call `preg_match()` once.

Sesquipedalian (changes requested):

Sesquipedalian:
> Note that we have an existing function to get valid tlds from iana:
> 
> https://github.com/SimpleMachines/SMF2.1/blob/ab409dffe23b960cfec2744f09c98734af6284fb/Sources/Subs.php#L6652
> 
> It is stored in regex form in the settings table under tld_regex & used by SMF to validate tlds. There is even some fallback logic just in case internet is not available.
> 
> The challenge here is that this logic in ForumSettings is called before the DB is even populated... I'm not sure our existing code can even be called here... I haven't tried it yet...

Unfortunately, that regex isn't sufficient for what @BrickOzp is trying to accomplish here.

other/install.php
Sesquipedalian: This should do the job.
```suggestion
if (!empty($allowed_tld))
{
	foreach (build_regex($allowed_tld, '/', true) as $ccSLD_regex)
	{
		preg_match('/\.' . strtr($ccSLD_regex, array('\*' => '[^.]*?')) . '$/u', 'vineyard.ca', $matches, PREG_OFFSET_CAPTURE);

		if (!empty($matches))
			break;
	}
}
```
The ccSLD list is so large that I had to set the `$returnArray` parameter of `build_regex()` to true and the loop through the four regular expressions it returned.

P.S.: Make sure to test this against 5bf8672 or later.

Sesquipedalian: Something like this should do the job.
```suggestion
					require_once(dirname(__FILE__) . '/Sources/Subs.php');
					foreach (build_regex($allowed_tld, '/', true) as $ccSLD_regex)
					{
						preg_match('/\.' . strtr($ccSLD_regex, array('\*' => '[^.]*?')) . '$/u', $url_parts['host'], $matches, PREG_OFFSET_CAPTURE);

						if (!empty($matches))
						{
							$stripped_domain = substr($url_parts['host'], 0, $matches[0][1]);
							$tld = $matches[0][0];							
							break;
						}
					}
```
The ccSLD list is so large that I had to set the `$returnArray` parameter of `build_regex()` to true and then loop through the four regular expressions it returned.

P.S.: Make sure to test this against 5bf8672 or later.

BrickOzp:
> **_After some thought, my preference is now to remove the existing default cookie setting logic entirely._** Make the user set the cookie settings; don't second guess them.

You mean like this?
https://github.com/SimpleMachines/SMF2.1/pull/6768

sbulen:
Yes.  In retrospect, it's a bit of a stretch to assume a local cookie is desired just because a path exists, or that global cookies are desired, just because a subdomain exists.

sbulen:
> Unfortunately, that regex isn't sufficient for what @BrickOzp is trying to accomplish here.

If we were to proceed, you can use the existing regex twice, narrowly focused on the end of the string.  The country codes are all in there.  I confirmed this works, before having doubts that we should even do this at all.

PR #6768, opened 2021-06-03
Remove cookie settings from installer

### LINKED ISSUE #6762: Installer - cookie setting defaults confused

BrickOzp:
Do not try to guess cookie settings during the install.

Fixes #6762
Closes #6764

### DIFF
--- a/other/install.php
+++ b/other/install.php
@@ -1291,44 +1291,6 @@ function DatabasePopulation()
 	if ((!empty($databases[$db_type]['utf8_support']) && !empty($databases[$db_type]['utf8_required'])) || (empty($databases[$db_type]['utf8_required']) && !empty($databases[$db_type]['utf8_support']) && isset($_POST['utf8'])))
 		$newSettings[] = array('global_character_set', 'UTF-8');
 
-	// Auto-detect local & global cookie settings
-	$url_parts = parse_url($boardurl);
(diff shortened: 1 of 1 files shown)


sbulen:
This is my preferred solution.  In retrospect, it's a bit of a stretch to assume a local cookie is desired just because a path exists, or that global cookies are desired, just because a subdomain exists.

sbulen:
Note that this basically brings 2.1 to the 2.0 functionality.  The 2.0 logic doesn't work, due to an error in the custom regex used.  2.0 never set the values for the cookie settings.

sbulen (approved):
I believe this is the proper way to go.  This is a change, so I would like to have some validation from another dev.

sbulen (approved):
LGTM.  Tests fine.

PR #6774, opened 2021-06-08
Make images responsive if max width is set

### LINKED ISSUE #6772: $modSettings['max_image_width'] can cause images to overflow post area
#### Description
If an image's natural width is greater than its container div's current width and the "Max display width of posted or attached images" setting is also set to a value greater than the container div's current width, the image will overflow the container div.

### Steps to reproduce
1. In  Administration Center ►  Attachments and Avatars ►  Attachment Settings, set the "Max display width of posted or attached images" setting to 1500px
2. Create a post with the following content: `[img]https://upload.wikimedia.org/wikipedia/commons/thumb/f/fb/Hotdog_-_Evan_Swigart.jpg/2000px-Hotdog_-_Evan_Swigart.jpg[/img]`
3. View the post.
4. Be sad that part of your hot dog is missing.

- Version/Git revision: 5bf8672
- Database Type: N/A
- Database Version: N/A
- PHP Version: N/A

The problem is here:
https://github.com/SimpleMachines/SMF2.1/blob/5bf867235d86015bd06bdd7bf52cccbd7c0acdfe/Sources/Subs.php#L3731-L3738
These lines insert CSS that overrides the `max-width: 100%` that would otherwise prevent the image from overflowing its container.

What we really want is to limit the width to the value of the "Max display width of posted or attached images" setting or to 100%, whichever is less. But the only way to do that is to programatically set the width parameter to our desired value and to leave the `max-width: 100%` CSS alone. 

Possible solutions are:
1. Revert the relevant changes initially introduced in 36aba93 and go back to programmatically setting the width attribute on the image on the server. Unfortunately, this will mean going back to running `url_image_size()` for each `[img]` BBC, which is less than ideal.
2. Do something clever with JavaScript to limit the image dimensions in the browser.
3. Simply get rid of the "Max display width of posted or attached images" setting and forget about it.

BrickOzp:
If the attachments max width is set the max-width
was set to a px value. This would override the responsive
default value of 100% and would make the images non responsive.
Instead calculate a minimum value of 100% and the selected
max width. This will make the image responsive up to the
selected max value.

Fixes #6772

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -3731,7 +3731,7 @@ function setupThemeContext($forceload = false)
 	// Add max image limits
 	if (!empty($modSettings['max_image_width']))
 		addInlineCss('
-	.postarea .bbc_img, .list_posts .bbc_img, .post .inner .bbc_img, form#reported_posts .bbc_img, #preview_body .bbc_img { max-width: ' . $modSettings['max_image_width'] . 'px; }');
+	.postarea .bbc_img, .list_posts .bbc_img, .post .inner .bbc_img, form#reported_posts .bbc_img, #preview_body .bbc_img { max-width: min(100%,' . $modSettings['max_image_width'] . 'px); }');
 
 	if (!empty($modSettings['max_image_height']))
 		addInlineCss('


BrickOzp:
Somehow this disappeared in a rebase/merge

Sesquipedalian:
This would be the best solution, except that `min()` only exists in the CSS4 draft and to my knowledge we've been limiting ourselves to CSS3. According to caniuse.com, all recently released browsers support `min()`, but we claim to support a number of older browsers, too.

Sesquipedalian:
At the very least, we should probably include some `isBrowser()` tests in the conditions when deciding whether to call `addInlineCSS()`

Sesquipedalian:
... On second thought, I suppose if the browser doesn't understand `max-width: min(100%, 500px);` it will just ignore the line as invalid, and therefore stick with the `max-width: 100%;` defined in index.css. So I guess this is fine after all. 🙂

live627 (approved):

PR #6777, opened 2021-06-08
Respects width and height params of [attach] BBC for non-image attachments

### LINKED ISSUE #6776: [attach] BBC's width and height params ignored for non-images
#### Description
When the [attach] BBCode is used to embed an attachment that is not an image, the width and height parameters are ignored.

### Steps to reproduce
1. Attach a video file to a post.
2. Embed the video in the post using something like this: `[attach id=1 width=560]my_video.mp4[/attach]`
3. Note how that the BBCode's width parameter is ignored when HTML is generated.

- Version/Git revision: 5bf8672
- Database Type:
- Database Version:
- PHP Version:

The problem is that the code inside the [attach] BBC's validation function tests for the values of `$width` and `$height` rather than `$params['{width}']` and `$params['{height}']`.

I will submit a PR shortly.

Sesquipedalian:
Fixes #6776

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -1451,24 +1451,24 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 						// Video.
 						elseif (strpos($currentAttachment['mime_type'], 'video/') === 0)
 						{
-							$width = !empty($width) ? ' width="' . $width . '"' : '';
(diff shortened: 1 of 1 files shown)


BrickOzp (approved):

sbulen:
@BrickOzp , @Sesquipedalian - if testing is completed, this can be merged.

PR #6779, opened 2021-06-09
Correctly store width and height for rotated images

### LINKED ISSUE #6769: Thumbnails in Topic > Print > Text with Images
Two issues:
Images displayed below the message body (ie, thumbnails) are displayed with the same width/height as images in the message body instead of with thumbnail width/height; and
Images that were rotated +/- 90 degrees are distorted when correctly below the message body.

Reported here with suggested fix: https://www.simplemachines.org/community/index.php?topic=577205.msg4089588#msg4089588

Print_Page-With_Images-1

### LINKED ISSUE #6778: Rotated images without thumbnails need width/height swapped
When the _"Administration Center > Attachments and Avatars > Attachment Settings > Thumbnail Settings > Resize images when showing under posts"_ option is disabled the width and height values need to be swapped for images that are rotated +/- 90 degrees ( ie, EXIF values 5, 6, 7 or 8 ).

The first four attachments shown below were saved with the current code.

The second four attachments shown below were saved with the suggested fix in **./Sources/Subs-Attachments.php**.
_Find:_
`		// Get the hash if no hash has been given yet.`

_Add Before:_
```
	// The width and height values need to be swapped for images that are rotated +/- 90 degrees so that they display correctly when the following Attachment Settings are used:
	// Display image attachments as pictures under post - enabled; and
	// Resize images when showing under posts - disabled.
	$orientation = 0;
	if (($exif_data = @exif_read_data($attachmentOptions['tmp_name'])) !== false && !empty($exif_data['Orientation']))
		$orientation = $exif_data['Orientation'];

	if (in_array($orientation, [5, 6, 7, 8]))
	{
		$new_width = $attachmentOptions['height'];
		$new_height = $attachmentOptions['width'];
		$attachmentOptions['width'] = $new_width;
		$attachmentOptions['height'] = $new_height;
	}
```

RotatedImages

BrickOzp:
If attached images have an orientation different than the
viewing orientation the width and height are mixed up,
resulting in wrong aspect ratio when viewed as non thumbnails
under a post or in the print topic view.
Switch width and height params in that case.

Fixes #6778
Fixes #6769

### DIFF
--- a/Sources/Subs-Attachments.php
+++ b/Sources/Subs-Attachments.php
@@ -687,6 +687,15 @@ function createAttachment(&$attachmentOptions)
 	$size = @getimagesize($attachmentOptions['tmp_name']);
 	list ($attachmentOptions['width'], $attachmentOptions['height']) = $size;
 
+	if (function_exists('exif_read_data') && ($exif_data = @exif_read_data($attachmentOptions['tmp_name'])) !== false && !empty($exif_data['Orientation']))
+		if (in_array($exif_data['Orientation'], [5, 6, 7, 8]))
+		{
+			$new_width = $attachmentOptions['height'];
+			$new_height = $attachmentOptions['width'];
+			$attachmentOptions['width'] = $new_width;
+			$attachmentOptions['height'] = $new_height;
+		}
+
 	// If it's an image get the mime type right.
 	if (empty($attachmentOptions['mime_type']) && $attachmentOptions['width'])
 	{


DiegoAndresCortes:
Question @BrickOzp 
I'm uploading pictures with a custom mod and then simply using the **createThumbnail** function, but I get this error after saving

_Uncaught Error: Call to undefined function exif_read_data()_

I noticed that you are adding this function here too, not sure why it's breaking on my end, or did I miss a related PR?

DiegoAndresCortes:
I did not have the extension enabled, now both this and custom code is working.
Not sure if that can be checked as enabled (probably not)

Anyways it's working now

Sources/Subs-Attachments.php
```diff
@@ -661,6 +661,15 @@ function createAttachment(&$attachmentOptions)
 	$size = @getimagesize($attachmentOptions['tmp_name']);
 	list ($attachmentOptions['width'], $attachmentOptions['height']) = $size;
 
+	if (($exif_data = @exif_read_data($attachmentOptions['tmp_name'])) !== false && !empty($exif_data['Orientation']))
```
Sesquipedalian: In case the EXIF extension isn't loaded:
```suggestion
	if (function_exists('exif_read_data') && ($exif_data = @exif_read_data($attachmentOptions['tmp_name'])) !== false && !empty($exif_data['Orientation']))
```

Sesquipedalian (changes requested):

Sources/Subs-Attachments.php
DiegoAndresCortes: The one in Subs-Graphics.php would require the same change then

BrickOzp: Done

Sesquipedalian:
@BrickOzp, it appears that you've added the `function_exists('exif_read_data')` check in Subs-Graphics.php but not in Subs-Attachments.php. Is there a reason for that?

BrickOzp:
Corrected now

BrickOzp:
Corrected now

Sesquipedalian (approved):

PR #6780, opened 2021-06-11
Limit the value of number of posts to match database

### LINKED ISSUE #6771: Setting user's post count to a negative value causes an error when saving
When modifying a user's post count in their profile and setting it to a negative value, an error will be given when attempting to save the profile. SMF will attempt to insert the negative value into the database despite the column being unsigned.

image

To reproduce, go to user's profile and modify "Account Settings" and set the post count to a negative value and then attempt to save.

BrickOzp:
The database column for a users post count is
mediumint(8) unsigned.
However the form did not limit the value resulting in
database errors if a value outside of the exepted range
was entered. Add input_validate and also set min max
value on the input field.

Fixes #6771

### DIFF
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -385,11 +385,15 @@ function loadProfileFields($force_reload = false)
 			'label' => $txt['profile_posts'],
 			'log_change' => true,
 			'size' => 7,
+			'min' => 0,
+			'max' => 2 ** 24 - 1,
 			'permission' => 'moderate_forum',
 			'input_validate' => function(&$value)
 			{
 				if (!is_numeric($value))
 					return 'digits_only';
+				elseif ($value < 0 || $value > 2 ** 24 - 1)
+					return 'posts_out_of_range';
 				else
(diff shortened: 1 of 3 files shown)


albertlast:
The limit depends on the db.
https://github.com/SimpleMachines/SMF2.1/blob/5a5f48db038660666493379e59ccec0b5993da93/other/install_2-1_postgresql.sql#L1064

max on mysql setup: 16777215
max on pg setup: 2147483647

so we could stick by mysql limits.

jdarwood007:
Then the MySQL version should move to int as well to keep them aligned.

albertlast:
well this topic we got many times,
issue is that on mysql side unsinged int kind are used,
which doesn't exists in postgres.

so on pg side it's always need to go one step bigger to match at least mysql max range.

So i would keep it, how it is for 2,1 release.

jdarwood007:
Fair enough.   I think 2^24 is plenty of room though still and don't need to code around for post postgres and mysql.  Any forum with that needed can make the changes to handle this instead.

Sesquipedalian (approved):

PR #6781, opened 2021-06-11
Prioritize avatar max width and height if set

BrickOzp:
If the setting to resize external avatars in css is set,
the avatars could be shown with incorrect aspect ratio,
Prioritize the max-width and max-height setting in code
over other css rules.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -3726,7 +3726,7 @@ function setupThemeContext($forceload = false)
 	// Now add the capping code for avatars.
 	if (!empty($modSettings['avatar_max_width_external']) && !empty($modSettings['avatar_max_height_external']) && !empty($modSettings['avatar_action_too_large']) && $modSettings['avatar_action_too_large'] == 'option_css_resize')
 		addInlineCss('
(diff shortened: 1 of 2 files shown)


Sesquipedalian:
Are we 100% sure that a CSS line to control this will *always* be inserted into the HTML header? If so, great. If not, then we should keep this default value in index.css.

Sesquipedalian:
Another option (and a better one, in my opinion) would be to get rid of that admin setting entirely and just define avatar dimensions in the CSS files. It is a pretty useless setting to begin with, the default value is good, and I've never understood why this value should somehow warrant a customizable admin setting when so many other CSS values don't.

BrickOzp:
You are right. The css is only inserted if avatar_action_too_large is set to option_css_resize
With other values there are no size restriction other than these css rules.

The $modSettings['avatar_max_width_external'] setting is also used
* to specify size when downloading gravatars.
* when setting external avatar by url.
* To resize images in js when previewing avatars. (though it does not work)

Sesquipedalian (approved):

PR #6789, opened 2021-06-17
Makes fixTag() more strict

Sesquipedalian:
Improves efficiency when filtering out URIs with invalid schemes, and explicitly changes invalid schemes to "invalid" just for good measure.


sbulen:
If this is used exclusively for BBC/autolinking, I wonder if we could be even more strict.  Instead of disallowing some, maybe we should only _allow_ some, e.g., http/https/ftp.  Specific to sharing links to sites/files.  

There's a lot of funkiness in this list...
https://en.wikipedia.org/wiki/List_of_URI_schemes

MissAllSunday (approved):

PR #6803, opened 2021-06-21
Send dates in iso format

### LINKED ISSUE #6631: Date picker does not work for non english languages

BrickOzp:
The date picker returns dates in localized format
but only english dates can be parsed.
Add a hidden field that will send the dates in
iso format.
Also format dates in Post Event screen in user
defined format.

Fixes #6631
Closes #6635

### DIFF
--- a/Sources/Calendar.php
+++ b/Sources/Calendar.php
@@ -113,6 +113,8 @@ function CalendarMain()
 	// Need a start date for all views
 	if (!empty($_REQUEST['start_date']))
 	{
+		if (!empty($_REQUEST['iso_start_date']))
+			$_REQUEST['start_date'] = $_REQUEST['iso_start_date'];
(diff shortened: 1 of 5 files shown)


Sesquipedalian:
This approach won't work if the user has disabled JavaScript.

Sesquipedalian:
I've added a suggestion to #6635 that should solve the remaining issue there. I recommend staying with that approach rather than this one.

PR #6804, opened 2021-06-23
Set default board profile when profile set to inherit

### LINKED ISSUE #6798: Out of range value for column 'id_profile' at row 1
#### Description
SQL database error at new board creation with non-standard access settings (not sure which one):
```sql
UPDATE smf_boards
			SET
				id_cat = 13,
				id_parent = 0,
				child_level = 0,
				board_order = 59,
				count_posts = 1,
				id_theme = 0,
				override_theme = 0,
				member_groups = '2,9',
				deny_member_groups = '',
				name = 'zzzzzzzzzzzzzzz',
				description = 'zzzzzzzzzzzzzzzzzzzzzzz',
				id_profile = -1,
				redirect = ''
			WHERE id_board = 129
```

### Steps to reproduce
Choose profile = -1 (first option) on board settings.

- Version/Git revision: RC3
- Database Type: Mysql

BrickOzp:
If creating a new board with the profile set as
"Inherit from parent board" an error will occur.
The board will be created, but without most of the
set options. Fixed this by setting the default profile
if inherit is selected. After the board is created
a check is done to find the parent board profile
setting and copy it.

Fixes #6798

### DIFF
--- a/Sources/ManageBoards.php
+++ b/Sources/ManageBoards.php
@@ -697,7 +697,7 @@ function EditBoard2()
 		$boardOptions['redirect'] = !empty($_POST['redirect_enable']) && isset($_POST['redirect_address']) && trim($_POST['redirect_address']) != '' ? trim($_POST['redirect_address']) : '';
 
 		// Profiles...
-		$boardOptions['profile'] = $_POST['profile'];
+		$boardOptions['profile'] = $_POST['profile'] == -1 ? 1 : $_POST['profile'];
 		$boardOptions['inherit_permissions'] = $_POST['profile'] == -1;
 
 		// We need to know what used to be case in terms of redirection.


Sources/ManageBoards.php
```diff
@@ -697,7 +697,7 @@ function EditBoard2()
 		$boardOptions['redirect'] = !empty($_POST['redirect_enable']) && isset($_POST['redirect_address']) && trim($_POST['redirect_address']) != '' ? trim($_POST['redirect_address']) : '';
 
 		// Profiles...
-		$boardOptions['profile'] = $_POST['profile'];
+		$boardOptions['profile'] = $_POST['profile'] == -1 ? 1 : $_POST['profile'];
```
frandominguezl, line 700: I have a question. Is there a reason why you chose `==` instead of `===`?

BrickOzp: It was like that on the row below

frandominguezl: Enough for me, thanks!

Sesquipedalian (approved):

Oldiesmann:
Does this account for the profile of the parent board changing at some point?

BrickOzp:
This PR does not change the behavior of this. SMF will copy the parent profile id at creation time, it will not be changed later if the parent changes. Subs-Boards.php line 944

PR #6808, opened 2021-06-28
Always load Security.php

BrickOzp:
If the forum was in maintenance mode 2 or
if the database could not be connected,
an undefined function error was displayed,
since Security.php was not loaded. Solve it
by loading it.

### DIFF
--- a/index.php
+++ b/index.php
@@ -75,10 +75,13 @@ require_once($sourcedir . '/Subs.php');
 require_once($sourcedir . '/Subs-Auth.php');
 require_once($sourcedir . '/Errors.php');
 require_once($sourcedir . '/Load.php');
+require_once($sourcedir . '/Security.php');
 
 // If $maintenance is set specifically to 2, then we're upgrading or something.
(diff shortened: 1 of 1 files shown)


BrickOzp:
Reported here https://www.simplemachines.org/community/index.php?topic=578042.msg0

Sesquipedalian (dismissed):

sbulen:
Note this comment in the forum suggests a tweak may be needed:
https://www.simplemachines.org/community/index.php?topic=578068.0

index.php
```diff
@@ -75,10 +75,13 @@
 require_once($sourcedir . '/Subs-Auth.php');
 require_once($sourcedir . '/Errors.php');
 require_once($sourcedir . '/Load.php');
+require_once($sourcedir . '/Security.php');
```
Sesquipedalian, line 78: Now that this is here, please remove the duplicate line of code that exists further down in the file.

Sesquipedalian (changes requested):

Sesquipedalian (approved):

PR #6811, opened 2021-07-02
Fix legacy attachments upgrade

### LINKED ISSUE #6181: Upgrade Bug - setting attachment mime types
#### Description
Not all mime types get calculated due to a bug in the upgrade script.

This doesn't really matter much but should be fixed.

### Steps to reproduce
1. Upgrade from 1.1.21 to 2.1 latest

- Version/Git revision:
- Database Type:mysql
- Database Version:mysql
- PHP Version:7.2

This is caused by your update script skipping rows because you are setting mime types on attachments and then searching for a limit query on where mimetypes != '' on each step into the upgrade for mime types.  It's essentially the same bug I found earlier regarding attachments not being renamed and moved properly.

### LINKED ISSUE #6810: Tiny annoying problem with attachment after upgrade :)
#### Description
After upgrade of a large forum from SMF 2.0 to SMF 2.1 everything works smoothly, except that I see insignificant but annoying  errors inside error log:
`
md5_file(/var/www/html/attachments/23742_12127006f5580f2a4ba08404a1470d8a393f96dd.dat): failed to open stream: No such file or directory
`
This attachment is an image and it is successfully shown inside the topic.
There is file without ".dat" extension present: /var/www/html/attachments/23742_12127006f5580f2a4ba08404a1470d8a393f96dd

How this happened ? Why this file wasn't renamed to .dat extension by upgrade script ?

PS: most of the other attachments are correct and with .dat extension, only some (very small amount) of attachments annoy me with this error in error log

BrickOzp:
During the legacy attachments upgrade, batches of 100
rows that do not have attachment_type of 1, are queried.
But in the prosessing of these rows, attachment_type is
sometimes change to 1. When this happens, rows will be
skipped and those attachments not processed. This can
lead to file not found errors. Fixed this problem by
decreasing the index when changing attachment_type.

Fixes #6181
Fixes #6810
Closes #6177

### DIFF
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -447,6 +447,7 @@ while (!$is_done)
 		SELECT id_attach, id_member, id_folder, filename, file_hash, mime_type
 		FROM {$db_prefix}attachments
 		WHERE attachment_type != 1
+		ORDER BY id_attach
 		LIMIT $_GET[a], 100");
 
 	// Finished?
@@ -513,10 +514,13 @@ while (!$is_done)
 		if ($row['id_member'] != 0)
 		{
 			if (rename($oldFile, $custom_av_dir . '/' . $row['filename']))
+			{
 				upgrade_query("
 					UPDATE {$db_prefix}attachments
 					SET file_hash = '', attachment_type = 1
 					WHERE id_attach = $row[id_attach]");
+				$_GET['a'] -= 1;
+			}
 		}
 		// Just a regular attachment.
 		else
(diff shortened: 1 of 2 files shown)


Sesquipedalian:
I haven't had a chance to test yet, but just looking at the code raises a concern about efficiency for me. This will loop through every single attachment, including

Sesquipedalian:
I haven't had a chance to test yet, but just looking at this code raises a concern about efficiency for me. This will loop through every single row in the table, many of them unnecessarily.

It might be better to keep the WHERE clause as it was and use `$smcFunc['db_fetch_all']` to populate an array with all the relevant rows at once. Then you could free the results of the initial query right away, and just loop through the array to make the updates. That should avoid creating any inconsistencies while still skipping unnecessary rows.

BrickOzp:
> It might be better to keep the WHERE clause as it was and use `$smcFunc['db_fetch_all']` to populate an array with all the relevant rows at once. Then you could free the results of the initial query right away, and just loop through the array to make the updates. That should avoid creating any inconsistencies while still skipping unnecessary rows.

The inconsistency will not occur inside the set, but between batches, since we only process 100 rows at a time.
But I fixed it in another way that will preserve the where clause.

MissAllSunday (approved):

PR #6814, opened 2021-07-04
Improves strip_php_comments()

Sesquipedalian:
In `strip_php_comments()`:
- Uses `token_get_all()` for faster performance.
- Removes unnecessary `$line_ending` parameter.
- No longer rewrites heredoc strings as quoted strings.

Also improves heredoc detection in `updateSettingsFile()`.


albertlast:
How mutch faster is the code?

Sesquipedalian:
About 10% faster.

PR #6824, opened 2021-07-11
Marks alerts as read when clicked

### LINKED ISSUE #6136: Alert no longer marked read after clicking/viewing it
#### Description
Alerts used to get marked read when viewing them, the logic used to be that they take you to `?action=profile;area=showalerts;alert={id}` where it gets marked read and then it redirects to its targeted link, and if no link redirects to the alerts list.

Either way that ensured that the alert got marked read when clicked instead of having to manually do that every time. 

That doesn't work anymore, not sure what changed but they now no longer take to the alert route to mark it read, alert links take directly to the targeted link.

(The previous way of redirecting could have been made better, I should have just included `alert={id}`to the url and then intercepted that and marked the alert read on forum load instead of redirecting but I didn't think of it at the time)

At the moment only some message related alerts get marked read when viewing them.

### Steps to reproduce
1. Produce a non topic/message related alert.
2. View alert.
3. You have to manually mark it read.

Sesquipedalian:
Fixes #6136

### DIFF
--- a/Sources/Profile-View.php
+++ b/Sources/Profile-View.php
@@ -739,7 +739,7 @@ function showAlerts($memID)
 			),
 			'view' => array(
 				'label' => $txt['view'],
-				'href' => $alert['target_href'],
+				'href' => $scripturl . '?action=profile;area=showalerts;alert=' . $id . ';',
 				'icon' => 'move',
 			),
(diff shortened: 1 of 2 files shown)


BrickOzp (approved):
Works

PR #6827, opened 2021-07-12
Support MS-LCID

live627:
This will enable reading the locale on Windows.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -849,9 +849,13 @@ function timeformat($log_time, $show_today = true, $offset_type = false, $proces
 
 	$timeformat = $finalizedFormats[$str][$format_type];
 
+	// Windows requires a slightly different language code identifier (LCID).
+	// https://msdn.microsoft.com/en-us/library/cc233982.aspx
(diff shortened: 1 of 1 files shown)


Sources/Subs.php
```diff
@@ -849,6 +849,11 @@ function timeformat($log_time, $show_today = true, $offset_type = false, $proces
 
 	$timeformat = $finalizedFormats[$str][$format_type];
 
+	// Windows requires a slightly different language code identifier (LCID).
+	// https://msdn.microsoft.com/en-us/library/cc233982.aspx
+	if ($context['server']['is_windows'])
+		$txt['lang_locale'] = strtr($txt['lang_locale'], '_', '-');
```
Sesquipedalian: `$txt['lang_locale']` is used in a lot of other places for various purposes, so we don't want to change it on the fly. Instead, I think it would be better to use something like...
```suggestion
	$lang_locale = $context['server']['is_windows'] ? strtr($txt['lang_locale'], '_', '-') : $txt['lang_locale'];
```
... and then replace `$txt['lang_locale']` with `$lang_locale` on subsequent lines inside this function.

Sesquipedalian (changes requested):

Sesquipedalian (approved):

PR #6834, opened 2021-07-13
Allow sorting the member list by custom fields after search

BrickOzp:
If searching the memberlist by a non custom profile field,
and then sorting the list by a custom profile field result
in an sql error. This is beacuse the custom field is
not joined. Fixed this by adding the custom field to the
query (if not already part since included by the search
fields), just as is done in the display member list case.


BrickOzp:
Reported here https://www.simplemachines.org/community/index.php?topic=577205.msg4091807#msg4091807

live627 (approved):

PR #6837, opened 2021-07-13
sceditor css only in default theme

DiegoAndresCortes:
For instance, the force_current is false by default so it wasn't needed in first place (unless I'm wrong of course)
And just load it from the default theme so themers don't have to update/maintain this file if it's ever updated.

For reference: https://www.simplemachines.org/community/index.php?topic=577856.0

In the topic I mentioned that this file wasn't copied when creating a new theme anyways, I didn't test if it's an expected behavior.

### DIFF
--- a/Sources/Subs-Editor.php
+++ b/Sources/Subs-Editor.php
@@ -1516,7 +1516,7 @@ function create_control_richedit($editorOptions)
 			$context['drafts_autosave_frequency'] = empty($modSettings['drafts_autosave_frequency']) ? 60000 : $modSettings['drafts_autosave_frequency'] * 1000;
 
 		// This really has some WYSIWYG stuff.
-		loadCSSFile('jquery.sceditor.css', array('force_current' => true, 'validate' => true), 'smf_jquery_sceditor');
+		loadCSSFile('jquery.sceditor.css', array('default_theme' => true, 'validate' => true), 'smf_jquery_sceditor');
 		loadTemplate('GenericControls');
 
 		// JS makes the editor go round
--- a/Sources/Subs-Editor.php
+++ b/Sources/Subs-Editor.php
@@ -1516,7 +1516,7 @@ function create_control_richedit($editorOptions)
 			$context['drafts_autosave_frequency'] = empty($modSettings['drafts_autosave_frequency']) ? 60000 : $modSettings['drafts_autosave_frequency'] * 1000;
 
 		// This really has some WYSIWYG stuff.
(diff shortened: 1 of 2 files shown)


Sources/Subs-Editor.php
```diff
@@ -1516,7 +1516,7 @@ function create_control_richedit($editorOptions)
 			$context['drafts_autosave_frequency'] = empty($modSettings['drafts_autosave_frequency']) ? 60000 : $modSettings['drafts_autosave_frequency'] * 1000;
 
 		// This really has some WYSIWYG stuff.
-		loadCSSFile('jquery.sceditor.css', array('force_current' => false, 'validate' => true), 'smf_jquery_sceditor');
+		loadCSSFile('jquery.sceditor.css', array('default_theme' => true, 'validate' => true), 'smf_jquery_sceditor');
```
Sesquipedalian, line 1519: Removing `'force_current' => false` is a good idea, but adding `'default_theme' => true` is not. The jquery.sceditor.css file is used to style the editor toolbar, among other things, and I can easily imagine themes wanting the ability to style that. A dark theme, for example, might want to change the toolbar background to a darker colour.

Sesquipedalian (changes requested):

Sources/Subs-Editor.php
DiegoAndresCortes: I thought we would just encourage to overwrite these styles using a custom file e.g. app.css or just in the index.css
This because the assumption of the sceditor requiring a css update in the future?

Sesquipedalian: That's a fair point. But if we are going to go that route, we should make it as easy and as obvious as possible to theme authors that they are supposed to do that. Something like this should do the job:
```suggestion
		loadCSSFile('jquery.sceditor.css', array('default_theme' => true, 'validate' => true), 'smf_jquery_sceditor');
		
		// THEME AUTHORS: If you want to change the CSS for the editor,
		// include a 'jquery.sceditor.theme.css' file in your theme.
		loadCSSFile('jquery.sceditor.theme.css', array('force_current' => true, 'validate' => true), 'smf_jquery_sceditor_theme');

```

DiegoAndresCortes: Sounds good.
We already have a beefy description in the index.template.php about loading files, should we just slap this note in there instead?

DiegoAndresCortes: Also this is of course just for good measure, I would like for us to add this in the wiki in the future.

Sesquipedalian: > should we just slap this note in there instead?

Why not both? 🙂

Sesquipedalian (approved):

PR #6839, opened 2021-07-14
Show custom fields default value in memberlist

BrickOzp:
In the memberlist, if a custom field is not set for a user
and there exists a default value, show the default value.
This is like the profile summary page.


BrickOzp:
Reported here https://www.simplemachines.org/community/index.php?topic=577205.msg4091876#msg4091876

live627 (approved):

PR #6842, opened 2021-07-14
Add outline to jump target and some scroll padding top

albertlast:
Highlight topic when you jump by using the post id:
old:
grafik

new:
grafik
some padding between the topic and the div,
and a outline around the post

would also like to do this für "new" but i see no solution to provide for both id and new jump (i had to decide)

### DIFF
--- a/Themes/default/Display.template.php
+++ b/Themes/default/Display.template.php
@@ -63,7 +63,7 @@ function template_main()
 	// Show the anchor for the top and for the first message. If the first message is new, say so.
 	echo '
 		</div><!-- #display_head -->
-		<a id="msg', $context['first_message'], '"></a>', $context['first_new_message'] ? '<a id="new"></a>' : '';
+		', $context['first_new_message'] ? '<a id="new"></a>' : '';
 
 	// Is this topic also a poll?
 	if ($context['is_poll'])
--- a/Themes/default/Display.template.php
+++ b/Themes/default/Display.template.php
@@ -474,9 +474,9 @@ function template_single_post($message)
 
 	// Show the message anchor and a "new" anchor if this message is new.
 	echo '
-				<div class="', $message['css_class'], '">
+				<div class="', $message['css_class'], ' targetAble" id="msg' . $message['id'] . '">
 					', $message['id'] != $context['first_message'] ? '
-					<a id="msg' . $message['id'] . '"></a>' . ($message['first_new'] ? '<a id="new"></a>' : '') : '', '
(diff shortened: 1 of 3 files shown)


DiegoAndresCortes:
Maybe use the same theme orange color for the border?

Sesquipedalian:
In theory, I like the idea of setting the id on the `div` element instead of an `a` element. But for consistency, you'd need to go through the rest of the templates and make similar changes in search results, personal messages, lists of recent posts, unread topics, etc., etc.

albertlast:
I had to test this, but my guess would be that for search already works.

Sesquipedalian:
Once you have changed the other templates to match, I'll be glad to see this go in.

Purely in term of aesthetics, I'd prefer a tinted background on the targeted post, rather than a special outline. Something like `background: #ffffdd` would be nice, I think.

albertlast:
I tested the outline function with search and it works perfect,
found one bug about start topic, got fixed.

About the css stuff which color which background etc. please do it by your own,
i don't see this part as scope of this pr to define which css defintion is the right one.

Sesquipedalian:
> I tested the outline function with search and it works perfect,
> found one bug about start topic, got fixed.

Great. 

It would be good to change this part in `template_single_pm()` in PersonalMessage.template.php in the same way that you have changed `template_single_post()` in Display.template.php:
https://github.com/SimpleMachines/SMF2.1/blob/71d52ccd0a8242d2f2be94037bcec7a5551d96b5/Themes/default/PersonalMessage.template.php#L300-L302

> About the css stuff which color which background etc. please do it by your own,
> i don't see this part as scope of this pr to define which css defintion is the right one.

Fair enough.

albertlast:
well i changed the places,
but the right and left side behave strange:
grafik

albertlast:
The source of problem with the missing right and left outline is the flow_hidden setted in the form:
https://github.com/SimpleMachines/SMF2.1/blob/0461d6f646842eeb9b1641eedf72aac8d75a9ff3/Themes/default/PersonalMessage.template.php#L186
since i don't know the reason why is there, i'm not willing to change it.

Sesquipedalian:
FYI, I made some tweaks to this PR before merging. You can see them 8fb5bca. Specifically, I changed the CSS to use `background: #ffffe0;` instead of `outline:3px red dotted;`, and I got rid of the `targetAble` class because `.windowbg:target` is enough to do the job.

albertlast:
to use background on a element which is always switchting between two color i don't find smart,
but this your decision.
.windowbg is used on to many places for my pov -> dangerous to use

PR #6845, opened 2021-07-14
Validates ISO country codes used for time zone lookups

### LINKED ISSUE #6844: Need to validate ISO country codes for time zones
#### Description
If the admin enters an invalid ISO country code in the "Show time zones from these countries first" setting, Subs-Timezones.php will generate errors after `timezone_identifiers_list(DateTimeZone::PER_COUNTRY, $country_code)` fails to find any time zones for the invalid country code:
https://github.com/SimpleMachines/SMF2.1/blob/ad09e71b65e8bc67799b2c7958c33cbf15001577/Sources/Subs-Timezones.php#L1557

Reported here: https://www.simplemachines.org/community/index.php?topic=578191.0

### Steps to reproduce
1. Enter an invalid ISO country code, such as "USA", in the "Show time zones from these countries first" setting.
2. Do something that calls `smf_list_timezones()`, such as reloading the same ACP page where you entered that setting.

- Version/Git revision: 6761679
- Database Type: N/A
- Database Version: N/A
- PHP Version: Any

Apparently we aren't checking the country codes before trying to use them. I will add some validation checks both in the ACP and in the affected functions.

Sesquipedalian:
Fixes #6844

### DIFF
--- a/Sources/ManageSettings.php
+++ b/Sources/ManageSettings.php
@@ -151,7 +151,7 @@ function ModifyModSettings()
  */
 function ModifyBasicSettings($return_config = false)
 {
-	global $txt, $scripturl, $context, $modSettings;
+	global $txt, $scripturl, $context, $modSettings, $sourcedir;
 
 	// We need to know if personal text is enabled, and if it's in the registration fields option.
 	// If admins have set it up as an on-registration thing, they can't set a default value (because it'll never be used)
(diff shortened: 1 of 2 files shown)


Sources/ManageSettings.php
```diff
@@ -253,6 +253,27 @@ function ModifyBasicSettings($return_config = false)
...
+				else
+					$country_codes[$key] = $country_code;
+			}
+
+			if (empty($country_codes))
+				unset($_POST['timezone_priority_countries']);
+			else
+				$_POST['timezone_priority_countries'] = implode(',', $country_codes);
+		}
+
```
live627: Can this code be moved to `Subs-Timezones.php`under a new function name such as `validateTimezones()`?

Sesquipedalian: Good idea.

smfbigguy: Disregard this. Your fix worked.

This is a quote from a member on my site. He says this:


Thanks, I've had a look at his fix. It will cause bugs you can't  take the first 2 letters of a 3 letter code. Some countries 2 and 3 letter codes change.
E.G Bangladesh's 2 letter is BD its 3 letter is BGD if you substitute BGD to BG that is Bulgaria.

They might want to look at the 1st and 3rd letter they might match better?

smfbigguy: Sorry the fix did not work...see text above.

Sesquipedalian: The member on your site misunderstands the purpose of this part of the code. It is not meant to correct three-character codes to two-character codes. It is just removing white space and any extra characters that might have snuck into the input for whatever reason.

In your original "USA" example, truncating to two characters coincidentally happened to produce the correct country code, but that was just a happy accident. Generally speaking, if the truncation happens to produce a valid but unexpected result, it will be the responsibility of the admin to find and enter the correct two-character code for the intended country.

Sesquipedalian: ...Actually, now that I think about it, it's probably cleaner to just fail immediately on overlong strings, rather than truncating them. If the admin accidentally enters an extra character, we'll just reject the whole string.

smfbigguy:
Thanks for taking the time to check things out. :)

Sources/ManageSettings.php
live627: I like that approach to play hard ball. Anything to reduce unexpected behavior is a good thing imo.

live627 (approved):

PR #6848, opened 2021-07-16
Move user actions button to same row as pagelinks for mobiles

BrickOzp:
On mobile screens the pagelinks and the user actions button
was on different lines. Move the user actions button to the
same line as pagelinks if they fit, to avoid some unused space.


BrickOzp:
Before
before

After
after

DiegoAndresCortes (approved):
I noticed it and forgot about it.
It's looking nice.

PR #6852, opened 2021-07-17
Restores the 'no_new_reply_warning' theme option

### LINKED ISSUE #6851: New replies warning can no longer be disabled
#### Description
Reported in the forum:
https://www.simplemachines.org/community/index.php?topic=578187.0
https://www.simplemachines.org/community/index.php?topic=577845.0
https://www.simplemachines.org/community/index.php?topic=537677.0

Yes, in high volume forums this would be extremely annoying.

There was a setting in 2.0 that was removed in 2.1 for some reason.  (Possibly in error? The setting is still referenced in the code...)

Should be logged.  I would not flag it for Final.

Sesquipedalian:
Fixes #6851

### DIFF
--- a/Sources/Post.php
+++ b/Sources/Post.php
@@ -393,7 +393,7 @@ function Post($post_errors = array())
 	// only at preview
 	if (empty($_REQUEST['msg']) && !empty($topic))
 	{
-		if (isset($_REQUEST['last_msg']) && $context['topic_last_message'] > $_REQUEST['last_msg'])
+		if (empty($options['no_new_reply_warning']) && isset($_REQUEST['last_msg']) && $context['topic_last_message'] > $_REQUEST['last_msg'])
 		{
 			$request = $smcFunc['db_query']('', '
 				SELECT COUNT(*)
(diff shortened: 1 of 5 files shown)


Sesquipedalian:
Updated to include changes to the upgrade files, too.

sbulen (approved):
I've confirmed the settings are properly migrated from 2.0, and that they work properly.

I've also confirmed that bulk changes made under Admin | Configuration | Themes and Layout | Member Options propagate properly & work as expected.

A bit of a nit, but the error message is a bit draconian (see below).  
image

Sesquipedalian:
> I've confirmed the settings are properly migrated from 2.0, and that they work properly.
> 
> 
> 
> I've also confirmed that bulk changes made under Admin | Configuration | Themes and Layout | Member Options propagate properly & work as expected.

Excellent.


> A bit of a nit, but the error message is a bit draconian (see below).  
> 
> image

I agree. Classifying and presenting this as a big, scary "post error" is overkill, since there's nothing actually wrong with the post. 

Unfortunately, I think we're stuck with the situation unless we make more substantial changes. Reporting a "post error" is currently the only method we have to deliver this notice, so if we wanted to avoid that we'd need to build some other means of delivery. But it is *way* too late now to put something like that into 2.1.

PR #6854, opened 2021-07-19
Ensures "New" link always uses white text

### LINKED ISSUE #6853: "New" link can sometimes use incorrect text colour
Under certain (rare?) circumstances, the "New" link next to a topic title could end up using the incorrect text colour. Specifically, if the URL in the href attribute of the link is one that has been visited before in the browser, the CSS for `a:visited` has a higher precedence than the CSS for `.new_posts`, and so the text colour for visited links, `#346`, overrides the `#fff` defined for the "New" link.

Screen Shot 2021-07-18 at 7 15 14 PM

Sesquipedalian:
Fixes #6853

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -242,6 +242,9 @@ body#help_popup {
 	border-radius: 2px;
 	opacity: 0.8;
 }
+a.new_posts:visited {
+	color: #fff;
+}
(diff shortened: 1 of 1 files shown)


DiegoAndresCortes (approved):
Nice catch

live627 (approved):

PR #6860, opened 2021-07-20
Improves validation and escaping for JavaScript variables in template_javascript() and addJavaScriptVar()

### LINKED ISSUE #6859: $context['javascript_vars'] and the value of 0
I was working on a personal project and used an integration hook to insert a custom value to $context['javascript_vars']:

```
$context['javascript_vars']['custom_count'] = 0;
```

That will be printed inside script tag as this:

```
var custom_count;
```

So it gets set as null instead of 0.

This is how it works:
for boolean values, we can set strings:

```
$context['javascript_vars']['custom_count'] = 'true';
```

For strings we'll use JavaScriptEscape():
```
$context['javascript_vars']['custom_count'] = JavaScriptEscape('Custom value');
```

I don't think there's a way for the actual zero value.

Subs.php, line 4311:
```
		foreach ($context['javascript_vars'] as $key => $value)
		{
			if (empty($value))
			{
				echo '
		var ', $key, ';';
			}
			else
			{
				echo '
		var ', $key, ' = ', $value, ';';
			}
		}
```

Following the boolean style and setting '0' also won't work because '0' is also empty()

The solution could be:
```
if (empty($value) && !is_numeric($value))
```

Sesquipedalian:
Fixes #6859 and generally improves our support for exporting different variable types from PHP to JavaScript.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -4308,16 +4308,13 @@ function template_javascript($do_deferred = false)
 
 		foreach ($context['javascript_vars'] as $key => $value)
 		{
-			if (empty($value) && !is_numeric($value))
-			{
-				echo '
-		var ', $key, ';';
-			}
-			else
-			{
-				echo '
-		var ', $key, ' = ', $value, ';';
-			}
+			if (!is_string($key) || is_numeric($key))
+				continue;
+
+			if (!is_string($value) && !is_numeric($value))
+				$value = null;
+
+			echo "\n\t\t", 'var ', $key, isset($value) ? ' = ' . $value : '', ';';
 		}
 
 		echo '
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -4308,7 +4308,7 @@ function template_javascript($do_deferred = false)
 
 		foreach ($context['javascript_vars'] as $key => $value)
 		{
-			if (empty($value))
+			if (empty($value) && !is_numeric($value))
 			{
 				echo '
 		var ', $key, ';';
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -2894,8 +2894,49 @@ function addJavaScriptVar($key, $value, $escape = false)
 {
 	global $context;
 
-	if (!empty($key) && (!empty($value) || $value === '0'))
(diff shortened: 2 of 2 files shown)


Sesquipedalian:
Does this look good to you, @asmith20002?

Sources/Subs.php
```diff
@@ -4308,7 +4308,7 @@ function template_javascript($do_deferred = false)
 
 		foreach ($context['javascript_vars'] as $key => $value)
 		{
-			if (empty($value))
+			if (empty($value) && !is_numeric($value))
```
live627: What if the var should be initialized as an empty string, array, object? Would `null` suffice here?
```suggestion
			if ($value === null)
```

Sesquipedalian: I don't think we can handle arrays and objects here in any case, but I suppose initiating as an empty string makes sense. For that matter, initiating as false also makes sense.

Sesquipedalian: As a side issue, it looks like we don't currently support adding empty values except '0' in `addJavaScriptVar()`. I don't see any reason for that limitation, though. A mod might very well want to initiate a var with an empty string or an explicit `false` or even as an empty array. Perhaps we should expand support for those types of empty variables in `addJavaScriptVar()`.

Either way, though, it still makes sense to improve our handling of various empty values here. After all, it is entirely possible for a mod to bypass `addJavaScriptVar()` and its limitations by appending directly to `$context['javascript_vars']`.

live627:
Any reason or advantage to not use the built in json encoding?

Sesquipedalian:
Er, because we aren't writing JSON. This is for dynamically writing literal JavaScript vars. Unless I'm missing something...?

asmith20002:
Thanks for the commit. I wouldn't mind more support for array and simple objects. At the moment I've modified my own project to something like this:

```
is_array($val) ? '[' . implode(', ', array_map('JavaScriptEscape', $val)) . ']' : JavaScriptEscape($val)
```

But that's just supporting simple arrays. Encoding array/objects to  JSON string for Javascript sounds good?

live627 (approved):

PR #6862, opened 2021-07-21
Use data attributes for custom blockquote data

live627:
Firefox and possibly otter browsers now don't shoo non-standard attributes in the HTML DOM. So let's write data attributes instead.

This can be reproduced by clicking to quote a post and switching to WYSIWYG mode.


Sesquipedalian (approved):
Works like a charm.

PR #6867, opened 2021-07-22
Dynamically loads CSS and JavaScript in ssi_examples.php

### LINKED ISSUE #6863: SSI:  Uncaught ReferenceError: $ is not defined
#### Description
Error is present simply loading ssi_examples.php.

Nothing urgent, but should be cleaned up:
image

### Steps to reproduce
1. Load ssi_examples.php with your console browser open

- PHP Version: 7.4
- Browser: Firefox

Sesquipedalian:
Fixes #6863

Rather than hard-coding the links and JavaScript variables, it makes better sense to call `template_css()` and `template_javascript()` and let them do their jobs.

### DIFF
--- a/ssi_examples.php
+++ b/ssi_examples.php
@@ -418,9 +418,12 @@ function template_ssi_above()
 	echo '<!DOCTYPE html>
 <html>
 	<head>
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Darn. This has unexpected side effects. Another solution will be needed.

PR #6878, opened 2021-07-23
update jquery

### LINKED ISSUE #6873: Update SMF 2.1 to use jQuery 3.6.0
jQuery 3.6.0 has been out for several months now. We should probably update SMF (final) to use it instead of 3.5.1. I've been using 3.6.0 since it came out in my test SMF2.1 without issue.

DiegoAndresCortes:
Fixes #6873 

This is all that is needed, right?


Sesquipedalian (approved):

Sesquipedalian:
> This is all that is needed, right?

Yup.

PR #6880, opened 2021-07-24
Improves loading of external CSS and JavaScript files

### LINKED ISSUE #6874: Conflict with minimise files and external theme files
#### Description
When a theme url is set to an external server or subdomain for serving theme scripts and css as static content, if minimise CSS and JavaScript files is enabled links for the mimimised files will be generated for non existent files, as they can not be created on an external server. No theme scripts and css will be loaded 

### Steps to reproduce
1. Set the theme url to an external server or subdomain that has the scripts and css for the theme being used.
2. Enable minimising scripts and css.

Given that it is possible to have theme scripts and css served externally and we don't want to take that away as it's a good (and requested) feature we should have a check that disables the minimise files setting when the theme url is external.

I'm working on a solution, but may need some guidance.

- Version/Git revision: Github latest

### LINKED ISSUE #6876: Should externally loaded scripts and css be receiving an SMF generated query string (seed)
#### Description
Should externally loaded scripts and css be receiving an SMF generated query string (seed)?

I believe the purpose of the query string is for caching purposes (cache busting) but in my testing so far it's not made any positive difference to that, as the string doesn't change after editing any given file that receives a query string, even after forcing a reload it doesn't change (this applied as well to non externally loaded theme scripts and css). 

What it does negatively do though is prevent some content from being used from the browser cache when the user already has the file in the cache. For example, giving the jQuery library a query string as we do now defeats the purpose of caching it as jquery-3.5.1.min.js?smf21rc4_xxxxxx is different to just jquery-3.5.1.min.js that a user may already have in their cache which could have been used.

I propose we should definitely remove the query string from the jQuery library load, and at least offer a setting to turn off the query string for external files (or even all files) unless explicitly set in the load function.

Sesquipedalian:
Fixes #6876 by never appending a seed string to jQuery file links or any external CSS and JavaScript file links.
Fixes #6874 by marking all theme files as external if the theme URL is external.

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -2702,7 +2702,7 @@ function loadSubTemplate($sub_template_name, $fatal = false)
  */
 function loadCSSFile($fileName, $params = array(), $id = '')
 {
-	global $settings, $context, $modSettings;
+	global $settings, $context, $modSettings, $boardurl;
 
 	if (empty($context['css_files_order']))
 		$context['css_files_order'] = array();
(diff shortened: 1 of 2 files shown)


Sources/Load.php
```diff
@@ -2717,9 +2717,15 @@ function loadCSSFile($fileName, $params = array(), $id = '')
 	$params['validate'] = isset($params['validate']) ? $params['validate'] : true;
 	$params['order_pos'] = isset($params['order_pos']) ? (int) $params['order_pos'] : 3000;
 
-	// If this is an external file, automatically set this to false.
-	if (!empty($params['external']))
-		$params['minimize'] = false;
+	// Is this theme always external?
+	if ($themeRef == 'theme' && strpos($settings['theme_url'], $boardurl) !== 0)
```
Sesquipedalian, line 2721: Hm. It just occurred to me that this check (and the identical one below) will cause problems if the admin decides to move the Themes directory outside of the main forum directory. So another solution is needed...

PR #6882, opened 2021-07-25
Doesn't try to guess whether CSS & JS files should always be external

Sesquipedalian:


jdarwood007:
This works, but I was going to suggest maybe we remove the external setup earlier:
```
	$params['external'] = isset($params['external']) ? $params['external'] : false;
```

Then do something like this:
```
	if ((!isset($params['external']) || ($params['external'] === true || $params['external'] === false) || ($themeRef == 'theme' && strpos($settings['theme_url'], $boardurl) !== 0))
```

Not tested, but basically if the external hasn't been specified already or if its not a valid bool, then we try to determine if we need to do some work to make it external.

Sesquipedalian:
Well, the problem is that we can't assume the file is external (meaning, not managed by SMF) merely because the theme URL doesn't begin with the value of $boardurl.

For example, supposed the admin stores the forum's index.php at `public_html/forum/index.php` with the corresponding URL `http://example.com/forum/index.php`, but changes the Themes dir to `public_html/resources/` with the corresponding URL `http://example.com/resources/`. In such a case, the CSS and JavaScript files stored in that custom theme directory are not external in the relevant sense—SMF can still handle the theme files the same way as if the theme's directory was in the default location. But the statement `strpos($settings['theme_url'], $boardurl) !== 0` will evaluate to true, which would cause all the theme files to be marked as external in this case even though they are not.

Unfortunately, I can't think of any good way of automatically distinguishing between the situation that @WhistlingDonkey wants to set up and the situation I described above. URLs with entirely different domains might actually reside on the same server, and URLs that appear to be pointing to the same server could actually use rewrites and redirection to pull files from other servers. We could try to issue a `gethostbyname()` call and compare it with our own IP address, but `gethostbyname()` is ridiculously slow and expensive, and even then such a test would not be foolproof.

So unless someone has an idea that hasn't occurred to me, there is no safe way to automatically make decisions like this. We have no choice but to rely on the admin not to enter bad configuration info (or in the case of a mod that changes SMF's behaviour in this regard, we have to rely on the mod to do its job correctly).

WhistlingDonkey:
It seems I've failed to communicate effectively here. I'm not actually trying to "set anything up" with this, it was just an issue I found that users could find themselves in because there was nothing in SMF to prevent it from happening. I was just trying to prevent others from falling into the trap.

So after a day of testing, I too don't think there is a way to automatically determine if a server is actually external, effectively external, or not. The original commit of checking the theme url against $boardurl would also fail and mark a theme external if both $boardurl and theme_url didn't have a directory as part of the url but were both subdomains (http://www.example.com, http://themes.example.com). If the theme url was checked against the just the root domain (example.com) it would work correctly for both subdomains and urls with directories but testing shows that it would still fail the original issue where the theme scripts and css are separated from the rest of the themes files with the theme_url server having a different different directory to theme_dir, unless the theme url didn't contain the root url.

I understand that it's normal for a user to move the themes to an entirely different directory, and this could even be a different server on the same machine and it will be fine as long as theme_dir points to the right place and both servers are running as the same user:group.

It's also entirely normal to separate scripts, css, images, etc away from a server hosting the actual site and it's dynamic content engine.

So as it seems an impossibility to automatically determine if the theme_url server is physically or user:group external or not, or it's just the separation of the scripts/css from the rest of the themes files I think all we can do is offer a flag in the theme settings for users to mark the theme scripts/css as external to the rest of the themes files, which the original commit could check instead (which I've tested), or as Sesquipedalian writes we'll just have to trust users not to put their themes in a situation where it's not writable and then try to minimise the scripts and css.

Sesquipedalian:
Assigning the CSS and JavaScript directories their own theme variables and adding a flag like you suggest are both good ideas for the future, but it is much too late in SMF 2.1's development process for such substantial changes to the theme system. 

A future version of SMF probably ought to implement these ideas in the core, but if anyone wants to implement them in 2.1, it will have to be via a mod.

PR #6885, opened 2021-07-26
Base pagination on all results

### LINKED ISSUE #6884: Search never returns more than one page of results
#### Description
Search never returns more than one page of results.

The problem is that the total search results used for the paging logic comes from this query:
https://github.com/SimpleMachines/SMF2.1/blob/1d163c4b492f22bc3d92e46c06d5f778a89cd728/Sources/Search.php#L1922
...which I am pretty sure can never exceed max-results-per-page.  So, the paging logic determines you only have one page & will not offer navigation links to other pages.

Looking into alternatives...  

### Steps to reproduce
1. Perform a search with more than results than your max results per page setting under Search | Settings...

- Database Type: na
- Database Version: na
- PHP Version: na

sbulen:
Fixes #6884 

Falls back to the existing query number if the search cache isn't populated.

**_BEFORE:_**
Test search from my guitar fx forum.  Note we have LOTS of references to Fender...
image

**_AFTER:_**
That's more like it...
image

### DIFF
--- a/Sources/Search.php
+++ b/Sources/Search.php
@@ -1919,7 +1919,10 @@ function PlushSearch2()
 		);
 
 		// How many results will the user be able to see?
-		$num_results = $smcFunc['db_num_rows']($messages_request);
+		if (!empty($_SESSION['search_cache']['num_results']))
+			$num_results = $_SESSION['search_cache']['num_results'];
+		else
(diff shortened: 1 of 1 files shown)


Sesquipedalian (approved):
LGTM. An alternative approach would have been to revert 9e74628 and dcb954f, but your approach here produces nicer code than those reversions would leave us with.

PR #6891, opened 2021-07-28
fix cors

live627:
[The params of both `implode()` and  `array_slice()` were mixed up in `FindCorsBaseUrl()`](https://scrutinizer-ci.com/g/SimpleMachines/SMF2.1/code-structure/release-2.1/operation/FindCorsBaseUrl). This introduces a helper function called `get_domain()` which is a little more robust than the current code... errors notwithstanding. 

Some example conversions include:

- domain.com -> domain.com
- sub.domain.com -> domain.com
- www.domain.com -> domain.com
- www.sub.sub.domain.com -> domain.com
- domain.co.uk -> domain.co.uk
- sub.domain.co.uk -> domain.co.uk
- www.domain.co.uk -> domain.co.uk
 - www.sub.sub.domain.co.uk -> domain.co.uk
 
While at it, I took the liberty to split part of `corsPolicyHeader()` into `get_allowed_http_origin()` to help facilitate my testing and to rewrite some off the main code to simplify and make tit easier too understand.

**Potential problems:**

- `$modSettings['cors_domains']` is a CSV of URLs not FQDNs. Feels like a bug to me, or at least unexpected behavior.
- IP addresses are not supported.
- Additionally, special domains such as `localhost` do not get parsed correctly.

**Previous pull requests of interest:**
- https://github.com/SimpleMachines/SMF2.1/pull/6592

### DIFF
--- a/Sources/Security.php
+++ b/Sources/Security.php
@@ -1320,6 +1320,7 @@ else
  * This sets the X-Frame-Options header.
  *
  * @param string $override An option to override (either 'SAMEORIGIN' or 'DENY')
+ *
  * @since 2.1
  */
 function frameOptionsHeader($override = null)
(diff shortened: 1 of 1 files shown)


jdarwood007:
I haven't done any testing yet on this actually in SMF code.  the cors_domains mod setting should support both urls and domains.  parse_url in php should handle that.

Your going to hate this, but SLDs are not uniform..  Which is why I avoided trying to be smart about figuring out what the SLD is.  This was brought up in the original PR.  This site is a good resource: https://publicsuffix.org/list/

I rand the get_domains through some testing to break it to show you.

```
<?php
$domains = array(
	'domain.com',
	'sub.domain.com',
	'www.domain.com',
	'www.sub.sub.domain.com',
	'domain.co.uk',
	'sub.domain.co.uk',
	'www.domain.co.uk',
	'www.sub.sub.domain.co.uk',
	'domain.priv.at',
	'sub.domain.priv.at',
	'www.domain.priv.at',
	'www.sub.sub.domain.priv.at',
	'domain.com.ua',
	'sub.domain.com.ua',
	'www.domain.com.ua',
	'www.sub.sub.domain.com.ua',
	'domain.agents.aero',
	'sub.domain.agents.aero',
	'www.domain.agents.aero',
	'www.sub.sub.domain.agents.aero',
	'domain.mino.gifu.jp',
	'sub.domain.mino.gifu.jp',
	'www.domain.mino.gifu.jp',
	'www.sub.sub.domain.mino.gifu.jp',
	'domain.academy.museum',
	'sub.domain.academy.museum',
	'www.domain.academy.museum',
	'www.sub.sub.domain.academy.museum',
);

foreach ($domains as $domain)
	echo $domain . ' : ' . get_domain($domain) . '<br/>' . "\n";

/**
 * Remove subdomains from hostnames
 *
 * IP addresses are not supported.
 *
 * Some example conversions include:
 *
 * - domain.com -> domain.com
 * - sub.domain.com -> domain.com
 * - www.domain.com -> domain.com
 * - www.sub.sub.domain.com -> domain.com
 * - domain.co.uk -> domain.co.uk
 * - sub.domain.co.uk -> domain.co.uk
 * - www.domain.co.uk -> domain.co.uk
 * - www.sub.sub.domain.co.uk -> domain.co.uk
 *
 * @param string $hostname
 *
 * @return string
 */
function get_domain($hostname)
{
	$hostname = strtolower(trim($hostname));
	$count = substr_count($hostname, '.');
	if($count === 2){
		if(strlen(explode('.', $hostname)[1]) > 3)
			$hostname = explode('.', $hostname, 2)[1];
	}
	else if($count > 2)
		$hostname = get_domain(explode('.', $hostname, 2)[1]);

	return $hostname;
}
```

Which on my testing produces
```
domain.com : domain.com
sub.domain.com : domain.com
www.domain.com : domain.com
www.sub.sub.domain.com : domain.com
domain.co.uk : domain.co.uk
sub.domain.co.uk : domain.co.uk
www.domain.co.uk : domain.co.uk
www.sub.sub.domain.co.uk : domain.co.uk
domain.priv.at : priv.at
sub.domain.priv.at : priv.at
www.domain.priv.at : priv.at
www.sub.sub.domain.priv.at : priv.at
domain.com.ua : domain.com.ua
sub.domain.com.ua : domain.com.ua
www.domain.com.ua : domain.com.ua
www.sub.sub.domain.com.ua : domain.com.ua
domain.agents.aero : agents.aero
sub.domain.agents.aero : agents.aero
www.domain.agents.aero : agents.aero
www.sub.sub.domain.agents.aero : agents.aero
domain.mino.gifu.jp : gifu.jp
sub.domain.mino.gifu.jp : gifu.jp
www.domain.mino.gifu.jp : gifu.jp
www.sub.sub.domain.mino.gifu.jp : gifu.jp
domain.academy.museum : academy.museum
sub.domain.academy.museum : academy.museum
www.domain.academy.museum : academy.museum
www.sub.sub.domain.academy.museum : academy.museum
```

Sesquipedalian:
If we wanted to programmatically fetch and check for ccSLDs in these strings, we could do so using code similar to what I suggested in [this comment](https://github.com/SimpleMachines/SMF2.1/pull/6764#discussion_r644415122). But it is worth noting that the list of ccSLDs at https://publicsuffix.org/list/public_suffix_list.dat is so large that (1) `build_regex()` has to compile them into *four* regular expressions in order to avoid breaking PCRE's maximum length limit for a regex, and therefore (2) that the compiled array of regexes is too big to be stored in the `value` column of the `settings` table (at least on MySQL) unless we change that column from a TEXT to a MEDIUMTEXT.

This amount of effort is simply not worthwhile. We'll do much better to provide solid help text and documentation to guide the admin in entering the correct values for their specific situation, rather than trying to guess this stuff.

jdarwood007:
I believe that to be an inefficient use here.  CORS has no use for 95% of SMF sites.  Only sites using a cross domain request with xhr to their main forum would need this.  Which is where I think it is better we keep the implantation simple and let the admin configure the setup, acknowledging the risks with an improper configuration.

I can help provide some of the basics for a wiki article if we need it.  CORS is an advanced subject to fully understand.  But doing a cross site xhr request isn't something that basic users are doing, but I would see intermediate users attempting.  We can try to bridge that gap with helping admins understand how to configure the domain list the best way.

jdarwood007:
I would see if a mod wants to implant working with that list.  It would work ok over javascript, where we can put the load on the client side to parse any typed urls and determine their SLD with javascript.

Also I only pointed out the initial function in this request to point out it can't solve everything because of the issue with SLDs.  I don't mind improving the code.  @live627 changes do seem to be better.  I haven't tested it though.  If we make a determination if we are ok with these changes, then I can do a test on the simplemachines.org test site

live627:
>  the cors_domains mod setting should support both urls and domains. parse_url in php should handle that.

[Let's test that.](https://3v4l.org/XadOC)

jdarwood007:
Looks like your right.  it think its the path.  If I do a // at the front, it gets it.

live627:
so currently `FindCorsBaseUrl()` returns `null` if the input string is not a URL. Should the data be normalized somehow or should be hope that the input is good if not a URL?

jdarwood007:
We could pass it through a cleanup on save no problem.  It looks like it just needs to be //domain.tld to pass parse_url.

What would be the overall status of what we are doing with this PR then?

Sesquipedalian:
While working on #7084, I noticed that the CORS related code also doesn't deal well with IRIs (URLs with international characters). Some of the changes in my own PR for dealing with IRIs will be helpful not only for dealing with the problems specific to IRIs, but also for dealing with some of the remaining issues discussed in the comments above. Therefore, I am going merge this PR into the one that I am working on, and then make some further changes to address the remaining issues.

jdarwood007:
I don't necessarily agree with the PR, but I can test it and your other PR against the test site to ensure CORS won't break.

Sesquipedalian:
@jdarwood007, a question for you: 

In the current code, these lines in `corsPolicyHeader()` set `$context['cors_domain']` but do not set `$context['valid_cors_found']`:
https://github.com/SimpleMachines/SMF2.1/blob/7143a85777f9926cf2008018d9ce5f596e9e5ef1/Sources/Security.php#L1377-L1378

But then this block of code a little further down will overwrite `$context['cors_domain']` because `$context['valid_cors_found']` is empty:
https://github.com/SimpleMachines/SMF2.1/blob/7143a85777f9926cf2008018d9ce5f596e9e5ef1/Sources/Security.php#L1428-L1441

This doesn't make a lot of sense as it stands. We need either (a) to change `if (empty($context['cors_domain']) || empty($context['valid_cors_found']))` to just `if (empty($context['cors_domain']))`, or else (b) to set `$context['valid_cors_found']` to some non-empty value when `$modSettings['cors_domains'] === '*'`.

The question is, what value, if any, should `$context['valid_cors_found']` have when `$modSettings['cors_domains'] === '*'`? Other parts of the codebase check the value of `$context['valid_cors_found']`, including some places that only check whether it is empty or not, so I need to know what the expected value is in such cases.

Sesquipedalian:
> I don't necessarily agree with the PR, but I can test it and your other PR against the test site to ensure CORS won't break.

Oh, that wasn't quite clear to me from the earlier comments. What exactly do you object to?

jdarwood007:
The first set of code you mentioned should be fixed then.  I didn't test that too heavily and ended up refactoring some stuff, so I must have forgot to test it again.  Note that its very weak security to allow that since you are allowing every website to make xhr requests.

I don't think this fix will do what is expected for all domains.  As noted, the (cc)SLDs are not uniform.  So assuming blindly that if we have a 2 character SLD, that is a cc SLD, won't work.  So while it does attempt to do better auto detection for those using the 2 character SLD (like domain.co.uk), there is others that will not be detected.  Its why I avoided detection and left this up to the admin.  Since CORS is an advanced subject for complex websites, it is not something we need to worry about being "smart".

SM website is a prime example of why we need it.  We have multiple sub domains (downloads, custom, support, and dev) and those need to make xhr requests to our main domain (with www).  The browser security considers these separate "domains" and would prohibit xhr requests upon making a preflight request and not getting a valid OPTIONS header back indicating the request is allowed.  So only sites using SMF that want to make xhr requests to SMF on another subdomain need this.

As a FYI, for SM Website, we just needed to enable the setting and because our cookie settings, CORS works perfectly without any of the other settings.  But on my test site, because its a 3rd level tld (sub2.sub.domain.tld), I have to make it wildcard CORS properly for the domain (in my case *.sub.domain.tld).

jdarwood007:
valid_cors_found can have anything btw.  Its value isn't currently checked, but I would suggest we call it "wildcard".  Its supposed to be there to future proof if we need to add intelligence later based on the matching.

Sesquipedalian:
Okay, thanks.

On further investigation and testing in the meantime, I've decided to take a different approach than in this PR after all. Specifically, I'm going to kill `FindCorsBaseUrl()` entirely, and instead rely on better input sanitization in ManageServer.php to determine the acceptable values for `$modSettings['globalCookiesDomain']` and `$modSettings['cors_domains']`. Once we know that those are sane values, we don't need to do anything clever at all to try to figure out which domains should be allowed.

jdarwood007:
Sounds good.  But remember that CORS check will receive the $_SERVER['HTTP_ORIGIN'] which would be the domain that the request was sent from.  So if I'm on custom.simplemachines.org and send a xhr request to www.simplemachines.org, the $_SERVER['HTTP_ORGIN'] will be custom.simplemachines.org while $_SERVER['HTTP_HOST'] will be www.simplemachines.org

So you will still need to do some "logic" to make the domain.

PR #6893, opened 2021-07-30
Recount large forums

sbulen:
A standalone index on a boolean is generally not used, especially on large tables.  Move it over to the id_board index, where it will be helpful.

The recount maintenance function had a query that can run long on very large forums (1M+ rows).   The index changes address that.

*** Testing completed...


Sesquipedalian:
It would be best to have this in 2.1.0.

other/install_2-1_mysql.sql
```diff
@@ -807,9 +807,8 @@ CREATE TABLE {$db_prefix}messages (
 	approved TINYINT NOT NULL DEFAULT '1',
 	likes SMALLINT UNSIGNED NOT NULL DEFAULT '0',
 	PRIMARY KEY (id_msg),
-	UNIQUE idx_id_board (id_board, id_msg),
+	UNIQUE idx_id_board (id_board, id_msg, approved),
 	UNIQUE idx_id_member (id_member, id_msg),
-	INDEX idx_approved (approved),
```
albertlast, line 812: Why removing this Index?

Sources/ManageMaintenance.php
```diff
@@ -1248,9 +1248,12 @@ function AdminBoardRecount()
 
 	// Update the latest message of each board.
 	$request = $smcFunc['db_query']('', '
-		SELECT m.id_board, MAX(m.id_msg) AS local_last_msg
+		SELECT m.id_board,
+			(SELECT MAX(id_msg)
```
albertlast: Why use of subquery which can be not rewrittrm by old mysql Version?

other/install_2-1_postgresql.sql
```diff
@@ -1218,9 +1218,8 @@ CREATE TABLE {$db_prefix}messages (
 # Indexes for table `messages`
 #
 
-CREATE UNIQUE INDEX {$db_prefix}messages_id_board ON {$db_prefix}messages (id_board, id_msg);
+CREATE UNIQUE INDEX {$db_prefix}messages_id_board ON {$db_prefix}messages (id_board, id_msg, approved);
```
albertlast, line 1221: This is a unique Index, by adding this col, you change his behavior.

albertlast (changes requested):
I don't understand the database changes.

albertlast:
Btw the standalone index, can be combined by good database at runtime: https://www.postgresql.org/docs/current/indexes-bitmap-scans.html

Sources/ManageMaintenance.php
```diff
@@ -1248,9 +1248,12 @@ function AdminBoardRecount()
...
+			FROM {db_prefix}messages
+			WHERE approved = {int:is_approved}
+			AND id_board = m.id_board) AS local_last_msg
```
albertlast: I guess you would like todo a literal join,
```suggestion
		SELECT m.id_board,
			(SELECT MAX(id_msg)
			FROM {db_prefix}messages
			WHERE approved = {int:is_approved}
			AND id_board = m.id_board) AS local_last_msg
		FROM {db_prefix}boards AS m
```

got me from a  ~200ms to a 0,5 ms execute time on a ~600k message table WITHOUT changing index.
My suggestion only works when line 
```
FROM {db_prefix}messages AS m
GROUP BY m.id_board
```
get deleted (1256 - 1257).

Sesquipedalian:
> I don't understand the database changes.

@albertlast, ~please see [here](https://www.simplemachines.org/community/index.php?topic=578294.0) for an explanation.~ Never mind. That topic is in a board that you can't access.

The basic problem is that this query can become quite slow when the messages table is very large. The changes @sbulen is proposing here are ones that seemed to improve execution time quite a bit in his tests.

sbulen:
Sesq - that's an internal discussion...

albertlast - Definitely a valid alternative, I'll test it.  Avoiding db changes would of course be preferred.  

The symptoms we see are pretty bad with large dbs - i.e., millions of rows.  Exec time is not going up in a linear fashion - it's much worse.  I'll share data later when I have time.

other/install_2-1_mysql.sql
sbulen: It is unused.  In general, it is a bad idea to put a standalone index on a boolean.  Low cardinality.  

However, moving it over to the id_board index helps *A LOT*.

Sources/ManageMaintenance.php
sbulen: Subqueries are supported by all of SMF's supported versions of mysql.

other/install_2-1_mysql.sql
albertlast: the problem is here, that you change the defintion of the unqiue index, so a post/message can exsits now more times in the same board.

Sources/ManageMaintenance.php
sbulen: Using your query results in this:
image

Two issues....  The first is the 40 second response.  The second (very minor) issue is too many results, though this is easily corrected.

other/install_2-1_postgresql.sql
sbulen: This index is unique already, as id_msg is unique on the message table.

You are correct in that this index doesn't really need to be UNIQUE...  E.g., it is not helping us enforce a true constraint.  It is primarily helping us get to messages on a board.

jdarwood007:
@albertlast latest query is still slow.  1 min 38.378 sec
Returns nulls as well, which can be handled by code.

Sources/ManageMaintenance.php
albertlast: without explain you benchmark make no sense,
i guess the index:
```sql
CREATE UNIQUE INDEX smf_messages_id_board ON smf_messages (id_board, id_msg);
```
is not used or not exists.

sbulen: Yes, id_board is heavily used, actually.  And is used in both explain plans.  

As noted earlier, the problem is only with larger boards.  We have seen that behavior changes dramatically ~1M rows.  In mysql only.  

(I have dummied up a large pg db and it appears to be OK...  More testing needed though.)

One comparison explain plan is here, before & after this PR:
image

albertlast: my guess is, that your mysql (5.7?) version is to old for the amount of data,
in 8.0 you could use directly lateral join: https://dev.mysql.com/doc/refman/8.0/en/lateral-derived-tables.html
the mention sql from me, is undepend of the amount of data, reason why it's below 0.0005 seconds at a 600k message table.

sbulen: Yes, 5.7.  And yes, I agree mysql8 and pg both handle this MUCH better.

sbulen:
> @albertlast latest query is still slow. 1 min 38.378 sec
> Returns nulls as well, which can be handled by code.

albertlast's other note, buried in a requested change comment above, applies...  

Specifically, the issue appears to be limited to mysql 5.x.  That's why I don't think he sees those response times.

other/install_2-1_mysql.sql
sbulen: (I think you cross posted into the wrong thread, but I'll respond to this concern here, because it is a good one.)

Both the board_id and the approved flag are on the message, ie., dependent on id_message.  So that cannot happen.

sbulen:
After further testing, the index changes were sufficient.

albertlast:
in my test the existing version is very slow:
```sql
SELECT m.id_board, MAX(m.id_msg) AS local_last_msg
		FROM smf_messages AS m
		WHERE m.approved = 1
		GROUP BY m.id_board
```
```
"Finalize GroupAggregate  (cost=35920.70..35946.28 rows=101 width=6) (actual time=129.633..129.711 rows=103 loops=1)"
"  Group Key: id_board"
"  ->  Gather Merge  (cost=35920.70..35944.26 rows=202 width=6) (actual time=129.628..132.688 rows=303 loops=1)"
"        Workers Planned: 2"
"        Workers Launched: 2"
"        ->  Sort  (cost=34920.67..34920.92 rows=101 width=6) (actual time=115.571..115.574 rows=101 loops=3)"
"              Sort Key: id_board"
"              Sort Method: quicksort  Memory: 29kB"
"              Worker 0:  Sort Method: quicksort  Memory: 29kB"
"              Worker 1:  Sort Method: quicksort  Memory: 29kB"
"              ->  Partial HashAggregate  (cost=34916.30..34917.31 rows=101 width=6) (actual time=115.533..115.542 rows=101 loops=3)"
"                    Group Key: id_board"
"                    ->  Parallel Seq Scan on smf_messages m  (cost=0.00..33608.50 rows=261560 width=6) (actual time=0.180..88.915 rows=209248 loops=3)"
"                          Filter: (approved = 1)"
"Planning Time: 0.156 ms"
"Execution Time: 132.767 ms"
```
https://explain.depesz.com/s/2tib

lateral join version:
```sql
SELECT m.id_board,
            (SELECT MAX(id_msg)
            FROM smf_messages
            WHERE approved = 1
            AND id_board = m.id_board) AS local_last_msg
        FROM smf_boards AS m
```
```
"Seq Scan on smf_boards m  (cost=0.00..437.31 rows=105 width=6) (actual time=0.023..0.554 rows=105 loops=1)"
"  SubPlan 2"
"    ->  Result  (cost=4.09..4.10 rows=1 width=4) (actual time=0.005..0.005 rows=1 loops=105)"
"          InitPlan 1 (returns $1)"
"            ->  Limit  (cost=0.42..4.09 rows=1 width=4) (actual time=0.005..0.005 rows=1 loops=105)"
"                  ->  Index Scan Backward using smf_messages_id_board on smf_messages  (cost=0.42..22764.24 rows=6215 width=4) (actual time=0.004..0.004 rows=1 loops=105)"
"                        Index Cond: ((id_board = m.id_board) AND (id_msg IS NOT NULL))"
"                        Filter: (approved = 1)"
"Planning Time: 0.205 ms"
"Execution Time: 0.575 ms"
```
https://explain.depesz.com/s/XVND

sbulen:
Did you test the PR including the index changes?

albertlast:
with the index the query runtime is halfed from the existing version:
```sql
SELECT m.id_board, MAX(m.id_msg) AS local_last_msg
		FROM smf_messages AS m
		WHERE m.approved = 1
		GROUP BY m.id_board
```
```
"Finalize GroupAggregate  (cost=1000.45..19324.79 rows=100 width=6) (actual time=21.381..62.111 rows=103 loops=1)"
"  Group Key: id_board"
"  ->  Gather Merge  (cost=1000.45..19322.79 rows=200 width=6) (actual time=21.355..65.794 rows=301 loops=1)"
"        Workers Planned: 2"
"        Workers Launched: 2"
"        ->  Partial GroupAggregate  (cost=0.42..18299.68 rows=100 width=6) (actual time=0.739..31.820 rows=100 loops=3)"
"              Group Key: id_board"
"              ->  Parallel Index Only Scan using smf_messages_id_board_test on smf_messages m  (cost=0.42..16991.07 rows=261523 width=6) (actual time=0.143..19.220 rows=209248 loops=3)"
"                    Index Cond: (approved = 1)"
"                    Heap Fetches: 0"
"Planning Time: 0.152 ms"
"Execution Time: 65.949 ms"
```
https://explain.depesz.com/s/7hmk

with this pr you only optimize the index, without solving the issue -> more data more runtime,
with literal join the runtime dependy only with amount of bord_id without scaling with the amount of messages/posts.

sbulen:
Some test results:

**_On the 4M DB in question, we now get 1.1 sec._**


**_On my 1.1m row DB:_**
image


**_On my 200k row DB:_**
image


**_Also did a board index comparison, to ensure no negative impact:_**
image

albertlast:
is your 4m db a mysql 8.0 or maria db 10.1+?

sbulen:
> is your 4m db a mysql 8.0 or maria db 10.1+?

Mysql 5.7.  I don't have direct access to that DB.

sbulen:
We renamed indexes, had to account for that.  Discovered while testing variants of upgrades.

sbulen:
Testing completed.  Performance comparisons above.  

Tested various installs & upgrades, pg & mysql, cli & browser.  Confirmed index changes properly reflected.

other/upgrade_2-1_mysql.sql
```diff
@@ -2865,6 +2865,26 @@ ALTER TABLE {$db_prefix}messages
...
+---#
+
+---# Updating messages drop id_board ix alt name
+ALTER TABLE {$db_prefix}messages
+DROP INDEX idx_id_board;
+---#
+
+---# Updating messages add new id_board ix
+ALTER TABLE {$db_prefix}messages
+ADD INDEX idx_id_board (id_board, id_msg, approved);
```
albertlast: unique is missing here,
personaly i would change the index from unique to normal,
since we learned the id_msg is already pk.,

sbulen: Thank you.  Nice catch.

albertlast:
personaly i would like to see them not be unique since it would be better for the performance...
but do what you like.

other/install_2-1_mysql.sql
```diff
@@ -807,9 +807,8 @@ CREATE TABLE {$db_prefix}messages (
 	approved TINYINT NOT NULL DEFAULT '1',
 	likes SMALLINT UNSIGNED NOT NULL DEFAULT '0',
 	PRIMARY KEY (id_msg),
-	UNIQUE idx_id_board (id_board, id_msg),
+	UNIQUE idx_id_board (id_board, id_msg, approved),
```
live627, line 810: Must reorder the columns low cardinality first so that it  gets this instead of `idx_approved`
```suggestion
	UNIQUE idx_id_board (approved, id_board, id_msg),
```

live627 (dismissed):
This new index brought the query down from 2s to 0.04s on localhost

- Total Posts: 216,048
- Total Topics: 45,375

other/install_2-1_mysql.sql
sbulen: I suspect your 2s measurement was an outlier; I would repeat that a few times & take an average.  On my *3M* row db on localhost, that query averages 0.06592s.  On my 200k row db on localhost, that query averages .01784s.  

Changing the order as you recommended above increased the average on the 3M row db to .07108, a decrease in performance.  

A good rule of thumb is to choose the columns as their most restrictive first, which in this case, is id_board.  That helps narrow the selection more quickly.  A boolean value will not help you do that so much:
https://www.percona.com/blog/2009/06/05/a-rule-of-thumb-for-choosing-column-order-in-indexes/

sbulen: I suspect your 2s measurement was an outlier; I would repeat that a few times & take an average.  On my *3M* row db on localhost, that query averages 0.06592s.  On my 200k row db on localhost, that query averages .01784s.  

Changing the order as you recommended above increased the average on the 3M row db to .07108, a decrease in performance.  

A good rule of thumb is to choose the columns as their most restrictive first, which in this case, is id_board.  That helps narrow the selection more quickly.  A boolean value will not help you do that so much:
https://www.percona.com/blog/2009/06/05/a-rule-of-thumb-for-choosing-column-order-in-indexes/

sbulen:
> personaly i would like to see them not be unique since it would be better for the performance...
> but do what you like.

Yes, UNIQUE adds a minor cost to writes, of course.  But also consider that it can offer a minor benefit on reads, as the query knows it can stop after reading the unique row (records are read a lot more than they are written...).  I read an article recently that said if you _can_ make an index unique, you should.  The thinking is that most queries can stop on a hit.  

I believe this was used as a design principle in SMF: Most (maybe all) indexes throughout SMF that include the pk are unique.  Look at attachments, members, categories, subscriptions, etc.  

So I think we should leave it UNIQUE, to be consistent with the approach used throughout SMF.  Also to minimize the change here, reducing risk.

albertlast:
> Yes, UNIQUE adds a minor cost to writes, of course. But also consider that it can offer a minor benefit on reads, as the query knows it can stop after reading the unique row (records are read a lot more than they are written...). I read an article recently that said if you _can_ make an index unique, you should. The thinking is that most queries can stop on a hit.

can you link me the manuel entry where this bahvior is described for mysql or pg?

sbulen:
> > can you link me the manuel entry where this bahvior is described for mysql or pg?

Most manuals don't go there, of course.  Lots of individual articles out there on the benefits of unique indexes.

The article I read recently read was on the Percona site.  I just looked, can't find it atm.

albertlast:
i asked in pg community channel and ther was the awnser that this kind of combination doesn't improved the performance.
So maybe for mysql side?

live627:
Since you got rid of my review:

If I restore the order, the database then uses`idx_approved`.

live627:
I dropped that index and now see a vast improvement  in the optiimizer:
```sql
ANALYZE format=json SELECT m.id_board, MAX(m.id_msg) AS local_last_msg
FROM smf_messages AS m
WHERE m.approved = 1
GROUP BY m.id_board
```

```json
{
  "query_block": {
    "select_id": 1,
    "r_loops": 1,
    "r_total_time_ms": 166.74,
    "table": {
      "table_name": "m",
      "access_type": "index",
      "key": "idx_id_board",
      "key_length": "7",
      "used_key_parts": ["id_board", "id_msg", "approved"],
      "r_loops": 1,
      "rows": 645486,
      "r_rows": 658956,
      "r_total_time_ms": 115.26,
      "filtered": 100,
      "r_filtered": 99.999,
      "attached_condition": "m.approved = 1",
      "using_index": true
    }
  }
}	
```

sbulen:
The pr gets rid of idx_approved on messages.

Please test the pr.  

Share specific times & explain plans if you still have a concern.

sbulen:
Oh ffs, another renamed ix.  Missed it.  

Would only affect rc => rc upgrades, but still.

Commit forthcoming.

sbulen:
@live627 please try again.  Had to account for renamed index...  You will find idx_approved gone this time.

live627 (approved):
Is good now.

live627:
@sbulen we now have a [bot to cleanup after PR Triage](https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/.github/workflows/label-cleanup.yml)—no more [manual intervention required](https://github.com/SimpleMachines/SMF2.1/pull/6905#event-5133240535).

PR #6905, opened 2021-08-03
Modifies the loading of the jQuery library by:

### LINKED ISSUE #6913: Provide more user selectable options for the jQuery library
Forum owners may have their personal preference to not use Google services as much as they can. They may also not be aware other CDN's exists to serve the jQuery library - which we should assume is the case -  which they could use in the custom field.

By providing extra selectable options for major CDN's (jQuerys own CDN and Microsofts CDN) we can help with that personal preference which can only provide goodwill to Simple Machines. It certainly isn't going to provide any harm.

The "auto" option is also removed as it does nothing but use Google anyway.

The linked PR provides more user selectable content delivery networks for the jQuery library and removes potential ambiguity or confusion of the current "auto" setting by removing it.

WhistlingDonkey:
- Providing further major jQuery CDN's as a choice (there is more than Google)
without having to use the custom field.

- Changes the fallback to the local jQuery source instead of Google.

- Removes potential confusion around the use of 'jquery_auto' by removing it, as
it did nothing but use Google anyway.

** Note, this change replaces the modSetting 'cdn' with 'google_cdn',
'jquery_cdn' and 'microsoft_cdn' so if users have set the jQuery source as
'cdn' (Google) they will fallback to the local source until they resave their
settings. ***

Signed by WhistlingDonkey, github@pyro.life


Sources/Load.php
```diff
@@ -2469,18 +2469,20 @@ function loadTheme($id_theme = 0, $initialize = true)
 	);
 
 	// Add the JQuery library to the list of files to load.
-	if (isset($modSettings['jquery_source']) && $modSettings['jquery_source'] == 'cdn')
-		loadJavaScriptFile('https://ajax.googleapis.com/ajax/libs/jquery/' . JQUERY_VERSION . '/jquery.min.js', array('external' => true, 'seed' => false), 'smf_jquery');
+	$jQueryUrls = array ('google_cdn' => 'https://ajax.googleapis.com/ajax/libs/jquery/'. JQUERY_VERSION . '/jquery.min.js', 'jquery_cdn' => 'https://code.jquery.com/jquery-'. JQUERY_VERSION . '.min.js', 'microsoft_cdn' => 'https://ajax.aspnetcdn.com/ajax/jQuery/jquery-'. JQUERY_VERSION . '.min.js');
```
Sesquipedalian: > ... so if users have set the jQuery source as 'cdn' (Google) they will fallback to the local source until they resave their settings.

To avoid this, please change 'google_cdn' back to just 'cdn'.

If we were starting from scratch, I could see an argument for using jQuery's own CDN as the default rather than Google's, but at this point we should avoid changing the current behaviour unless necessary.

Sources/Load.php
```diff
@@ -2469,18 +2469,20 @@ function loadTheme($id_theme = 0, $initialize = true)
...
+	
+	if (array_key_exists($modSettings['jquery_source'], $jQueryUrls))
+		loadJavaScriptFile($jQueryUrls[$modSettings['jquery_source']], array('external' => true, 'seed' => false), 'smf_jquery');
 
 	elseif (isset($modSettings['jquery_source']) && $modSettings['jquery_source'] == 'local')
 		loadJavaScriptFile('jquery-' . JQUERY_VERSION . '.min.js', array('seed' => false), 'smf_jquery');
```
Sesquipedalian, line 2478: Since `local` will be the default that we fall back to a few lines below here, this `elseif` block can be removed.

Sources/ManageSettings.php
```diff
@@ -179,9 +179,10 @@ function ModifyBasicSettings($return_config = false)
 			'select',
 			'jquery_source',
 			array(
-				'auto' => $txt['jquery_auto'],
+				'google_cdn' => $txt['jquery_google_cdn'],
```
Sesquipedalian: Please change 'google_cdn' back to just 'cdn'.

Sources/Profile-Export.php
```diff
@@ -1865,7 +1865,20 @@ function ($a, $b)
...
+
+	elseif (isset($modSettings['jquery_source']) && $modSettings['jquery_source'] == 'local')
+		loadJavaScriptFile('jquery-' . JQUERY_VERSION . '.min.js', array('seed' => false), 'smf_jquery');
+
+	elseif (isset($modSettings['jquery_source'], $modSettings['jquery_custom']) && $modSettings['jquery_source'] == 'custom')
+		loadJavaScriptFile($modSettings['jquery_custom'], array('external' => true, 'seed' => false), 'smf_jquery');
+
+	// Fall back to the local source
+	else
+		loadJavaScriptFile('jquery-' . JQUERY_VERSION . '.min.js', array('seed' => false), 'smf_jquery');
```
Sesquipedalian: Please revert these changes in Profile-Export.php.

In this case, the jQuery library *always* needs to be loaded from a reliable CDN. This is because the export files are meant to function as permanent, standalone files that will continue to work as expected even if the forum they were exported from disappears one day.

Sesquipedalian (changes requested):

Sesquipedalian:
Also, please open an issue explaining why these changes are required (or at least recommended).

WhistlingDonkey:
#6913

Sources/Profile-Export.php
WhistlingDonkey: I can't agree that Google is in anyway more reliable than the other CDN's, or even the local copy since the page is served from the same place, Google are just as fallable as the others and a user would likely have the chosen CDN's copy in cache by the time they land on this page as well but if it needs to be changed back then it shall be done, there is probably more to this one than I know.

Sources/Load.php
WhistlingDonkey: The same will happen if they have the jquery source set to 'auto' or haven't even set the source by saving the page, they will now fallback to the local source instead of Google which was the previous fallback and I guess it would be advisable not to break that too?

Would adding the line `if (isset($modSettings['jquery_source']) && ($modSettings['jquery_source'] == 'cdn' || $modSettings['jquery_source'] == 'auto')) $modSettings['jquery_source'] = 'google_cdn';`or something along those lines, or simply expanding the array to include cdn and auto be a better idea to capture the old settings?

Sources/ManageSettings.php
WhistlingDonkey: See comment for Load.php

Sources/Profile-Export.php
Sesquipedalian: > ... or even the local copy since the page is served from the same place...

This seems to be the source of the misunderstanding. ProfileExport.php doesn't generate dynamic pages to be served to the browser. It generates static HTML files that the user can download to their own hard drive for archival purposes. Therefore, _everything_ in these files must be either (a) served from localhost, (b) embedded directly in the HTML file, or (c) served from a very reliable external source, where "reliable" means "highly unlikely to succumb to link rot any time in the foreseeable future"—in other words, from a major CDN. So yes, this really does need to be changed back.

Sources/Load.php
Sesquipedalian: In this case, the value `auto` means the admin said "I dunno, just use whatever the SMF devs have decided would be a good default", so that's not an issue. But the value `cdn` means that the admin specifically selected "Google CDN" as their jQuery source, so we should maintain that if we can.

... Also, the more I think about it, the more convinced I become that a CDN (I don't really care which one) should be the default, rather than the local file. Using a CDN for this is better both for the server and for the end user, so it should be the default unless the admin has a specific reason not to do so.

Sources/Load.php
Sesquipedalian: On second thought, please keep this block and instead change the default back to a CDN.

Sources/Load.php
```diff
@@ -2469,18 +2469,20 @@ function loadTheme($id_theme = 0, $initialize = true)
...
 
 	elseif (isset($modSettings['jquery_source'], $modSettings['jquery_custom']) && $modSettings['jquery_source'] == 'custom')
 		loadJavaScriptFile($modSettings['jquery_custom'], array('external' => true, 'seed' => false), 'smf_jquery');
 
-	// Auto loading? template_javascript() will take care of the local half of this.
+	// Fall back to the local source
 	else
-		loadJavaScriptFile('https://ajax.googleapis.com/ajax/libs/jquery/' . JQUERY_VERSION . '/jquery.min.js', array('external' => true, 'seed' => false), 'smf_jquery');
+		loadJavaScriptFile('jquery-' . JQUERY_VERSION . '.min.js', array('seed' => false), 'smf_jquery');
```
Sesquipedalian: Looking at this again, I think it is best to keep the default as it was. In the vast, vast majority of cases, using a CDN is the right choice here.

Sesquipedalian (changes requested):

Sources/ManageSettings.php
```diff
@@ -179,9 +179,11 @@ function ModifyBasicSettings($return_config = false)
 			'select',
 			'jquery_source',
 			array(
-				'auto' => $txt['jquery_auto'],
+				'auto' => $txt['jquery_forum_default'],
```
Sesquipedalian:
```suggestion
```

Themes/default/languages/Help.english.php
```diff
@@ -265,7 +265,7 @@
...
 $helptxt['enable_ajax_alerts'] = 'This option allows your members to receive AJAX notifications. This means that members don\'t need to refresh the page to get new notifications.<br><strong>DO NOTE:</strong> This option might cause a severe load at your server with many users online.';
 $helptxt['alerts_auto_purge'] = 'Once an alert has been read, it is rarely needed again. For performance reasons, it is a good idea to automatically delete them after a while.';
-$helptxt['jquery_source'] = 'This will determine the source used to load the jQuery Library. <em>Auto</em> will use the CDN first and if not available fall back to the local source. <em>Local</em> will only use the local source. <em>CDN</em> will only load it from Google CDN network';
+$helptxt['jquery_source'] = 'This will determine the source used to load the jQuery Library. <em>Forum default</em> uses the forum default source, <em>Google CDN, jQuery CDN</em> and <em>Microsoft CDN</em> will load the jQuery library from those respective CDN networks. <em>Local</em> will only use the local source. <em>Custom</em> allows you to specify a custom URL for the library.';
```
Sesquipedalian:
```suggestion
$helptxt['jquery_source'] = 'This will determine the source used to load the jQuery Library. <em>Google CDN</em>, <em>jQuery CDN</em>, and <em>Microsoft CDN</em> will load the jQuery library from those respective CDN networks. <em>Local</em> will only use the local source. <em>Custom</em> allows you to specify a custom URL for the library.';
```

Themes/default/languages/ManageSettings.english.php
```diff
@@ -72,8 +72,10 @@
 $txt['jquery_custom_label'] = 'Custom';
 $txt['jquery_custom'] = 'Custom URL to the jQuery Library';
 $txt['jquery_local'] = 'Local';
-$txt['jquery_cdn'] = 'Google CDN';
-$txt['jquery_auto'] = 'Auto';
+$txt['jquery_google_cdn'] = 'Google CDN';
+$txt['jquery_forum_default'] = 'Forum default';
```
Sesquipedalian:
```suggestion
```

Sesquipedalian (changes requested):
Looking pretty good.

At this point, we can remove the "auto" option entirely. Forums that currently have "auto" selected will end up using the Google CDN, just as they do now, and "Google CDN" will appear as the default option in the select menu for them when they load the admin page. In light of that, we can just ditch "auto" and move on without it.

WhistlingDonkey:
I had reintroduced auto as 'Forum default' as I figured it'd serve as a reminder that the admin hadn't explicitly set a source but with Google CDN as the first and default choice it is not needed. Requested changes applied.

Sesquipedalian (approved):

Sesquipedalian:
Thanks for contributing!

PR #6909, opened 2021-08-03
Changed newicon to use :target css

albertlast:
With this pr a user see now where new is starting,
by using the new :target css

grafik
the newicon use now a static url topic.123456.nm instead of topic.123456.msg1234#new get than redirected to the newest msg when existing.
grafik

in essential the id= #new is not used any more from this jump

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -1031,15 +1031,16 @@ function loadBoard()
 			SELECT
 				m.id_msg
 			FROM
-				smf_messages AS m
+			{db_prefix}messages AS m
 				JOIN (
 					SELECT
 						COALESCE(lt.id_msg, lmr.id_msg, - 1) AS new_from
-					FROM smf_log_topics AS lt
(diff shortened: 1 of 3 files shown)


Sesquipedalian:
Running another query to accomplish this seems like overkill to me. You could just reuse the same logic that inserts the `<a id="new"></a>` element in Display.template.php to add an extra ` first_new` class to the post's container div, and then change `.windowbg:target` to `.windowbg:target, .windowbg.first_new` in index.css.

albertlast:
i see the ui stuff more harmfull than the query approache,
since it effect themes.
specialy in my pov is new/first_new not realy usefull in general,
so i don't want to support this existing.

Sesquipedalian:
Well, we aren't going to change the behaviour of the "new" links at this stage of the development cycle, so the approach proposed in this PR isn't going to be merged. Thanks anyway, though.

PR #6910, opened 2021-08-03
Adds integration hooks to some permission checking functions

### LINKED ISSUE #6902: function removeMessage - add permission override check
function removeMessage - add permission override check

Was working on a mod and today I realized there is no way to override the permission check. I wanted to use this function to delete auto generated messages inside a topic in a board that the user does not have access to delete messages.
I will need to currently write my own function to do so.

Sesquipedalian:
Fixes #6902

### DIFF
--- a/Sources/RemoveTopic.php
+++ b/Sources/RemoveTopic.php
@@ -614,6 +614,9 @@ function removeMessage($message, $decreasePostCount = true)
 	$row = $smcFunc['db_fetch_assoc']($request);
 	$smcFunc['db_free_result']($request);
 
+	// Give mods a heads-up before we do anything.
+	call_integration_hook('integrate_pre_remove_message', array($message, $decreasePostCount, $row));
+
 	if (empty($board) || $row['id_board'] != $board)
 	{
 		$delete_any = boardsAllowedTo('delete_any');
--- a/Sources/Security.php
+++ b/Sources/Security.php
@@ -913,6 +913,9 @@ function allowedTo($permission, $boards = null, $any = false)
 	// This should be a boolean.
 	$any = (bool) $any;
 
+	// Maybe a mod needs to tweak $user_info['permissions'] on the fly?
+	call_integration_hook('integrate_allowed_to', array($permission, $boards, $any));
+
 	// Are we checking the _current_ board, or some other boards?
 	if ($boards === null)
 	{
@@ -1145,6 +1148,9 @@ function boardsAllowedTo($permissions, $check_access = true, $simple = true)
 			}
 		}
 	}
+
+	// Maybe a mod needs to tweak the list of allowed boards on the fly?
+	call_integration_hook('integrate_boards_allowed_to', array(&$boards, $deny_boards, $permissions, $check_access, $simple));
 
 	return $boards;
 }
--- a/Sources/Security.php
+++ b/Sources/Security.php
@@ -913,13 +913,15 @@ function allowedTo($permission, $boards = null, $any = false)
 	// This should be a boolean.
 	$any = (bool) $any;
 
-	// Maybe a mod needs to tweak $user_info['permissions'] on the fly?
-	call_integration_hook('integrate_allowed_to', array($permission, $boards, $any));
-
 	// Are we checking the _current_ board, or some other boards?
 	if ($boards === null)
 	{
-		if (count(array_intersect($permission, $user_info['permissions'])) != 0)
+		$user_permissions = $user_info['permissions'];
+
+		// Allow temporary overrides for general permissions?
+		call_integration_hook('integrate_allowed_to_general', array(&$user_permissions, $permission));
+
+		if (count(array_intersect($permission, $user_permissions)) != 0)
 			return true;
 		// You aren't allowed, by default.
 		else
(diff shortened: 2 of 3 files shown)


live627:
Do integration hooks really belong in core permission checking routinizes? I still hold to https://github.com/SimpleMachines/SMF2.1/issues/6902#issuecomment-891781387. Besides, I have this ugly feeling that a security hole may easily be inadvertently introduced by a mod.

Sesquipedalian:
> Do integration hooks really belong in core permission checking routinizes? I still hold to https://github.com/SimpleMachines/SMF2.1/issues/6902#issuecomment-891781387. Besides, I have this ugly feeling that a security hole may easily be inadvertently introduced by a mod.

Actually, I'm pretty sure that these hooks will make inadvertent security holes _less_ likely, not more. 

Mods can already change the content of `$user_info['permissions']` using _any_ integration hook, because `$user_info` is a global variable. But using other hooks that are not designed for that task is inherently ham-fisted, because the mod has to make the changes in advance, just in case the changes _might_ be needed later. 

In contrast, these hooks allow the mod to make changes with surgical precision. A mod using these hooks will have the maximum possible amount of information about the situation available to it, allowing it to be very precise about whether the changes are appropriate under the current conditions. Moreover, the changes don't need to be made until the moment they are actually called for, narrowing the window of opportunity for unexpected side-effects to occur.

Sesquipedalian:
I've split the hook I added to `allowedTo()` into two distinct hooks. With this change, mods never need to manipulate `$user_info['permissions']` at all when they want to grant a temporary override of the usual permissions. This means that overriding the permissions during one call to `allowedTo()` will never have any effect on subsequent calls to the function or on any other code that relies on the value of `$user_info['permissions']`.

live627:
My hardline stance against this continues. I'm of the opinion that  nothing should modify permissions routines, least of all integration hooks that will probably invite such behavior at some point. And while `$user_info['permissions']` can be modified, it's not exactly documented so is somewhat a black box. Ditto for injection through the package manager. But unscrupulous programmers who want to leak stuff and do some damage will find a way.

My point is that I have a bad feeling that this will create a slippery slope.

Also I should point out that almost all of the mods I've seen in the wild set their entry points only once—during install—and save calls in the database. Even if a hook is added at runtime, the remove call can easily be forgotten. (I sometimes forget to import global variables).

Sesquipedalian:
> My hardline stance against this continues. I'm of the opinion that nothing should modify permissions routines, least of all integration hooks that will probably invite such behavior at some point. And while `$user_info['permissions']` can be modified, it's not exactly documented so is somewhat a black box. Ditto for injection through the package manager. But unscrupulous programmers who want to leak stuff and do some damage will find a way.
> 
> My point is that I have a bad feeling that this will create a slippery slope.

An unscrupulous modification author can already do anything they want, if they can convince the admin to install their mod. That's why we have the customization team review mods before they are listed in our modifications catalogue. These hooks don't change anything about that. All they do is make it easier for well-intentioned mod authors to be as minimally invasive as possible.

> Also I should point out that almost all of the mods I've seen in the wild set their entry points only once—during install—and save calls in the database. Even if a hook is added at runtime, the remove call can easily be forgotten. (I sometimes forget to import global variables).

Er, I'm not sure why this is relevant. Are you referring back to the discussion in #6902? If so, I never proposed using on-the-fly calls to `add_integration_hook()` in @VBGAMER45's mod. I mused about using something like this PR's proposed `integrate_pre_remove_message` hook to, say, set a $context variable or something, and then checking that variable in another hook such as this PR's proposed `integrate_boards_allowed_to` hook. I had envisioned that happening using typical hook behaviour. But none of that is relevant to these proposed hooks in and of themselves, so...? 😕 

-----

At any rate, it sounds like we are at an impasse here, so I suppose we need to get a third dev's opinion as a tie-breaker. @sbulen, @MissAllSunday, @BrickOzp, could one of you please take a look at this PR and give us your thoughts? The matter in contention is whether the proposed `integrate_allowed_to_general`, `integrate_allowed_to_board`, `integrate_boards_allowed_to`, and `integrate_groups_allowed_to` hooks are a good idea or not.

MissAllSunday:
My stand on issues like this one has always been the same, SMF is a tool and we only provide said tool, [we don't have any governance about how such tool would/should be used](https://github.com/SimpleMachines/SMF2.1/blob/release-2.1/LICENSE#L23).

Same idelogy would naturally have to be inherited by the code and subsequentially to hooks.

sbulen:
I agree with @live627 that we are throwing too much stuff into 2.1.  We should be a lot more selective about what gets merged at this point.  I have made this observation numerous times in various discussions.  

We need a stable codebase at some point...  If not now, when?

I also looked at permissions-related mods on the customization site (just the descriptions, not code).  Most seem to add new permissions & utilize the existing permissions logic.  Some 1.1 mods appear to change the nature of the logic (e.g., member-specificity).  But I don't see mods that appear to be fundamentally changing the permissions logic.  I.e., I don't see a lot of pent up need.

So I do not support the change.  And going forward I would not merge much of anything else other than simple bug fixes or hard requirements for 2.1.0.

jdarwood007:
While I agree, hooks generally should be something we can easily add at any point to the code base with no repercussions (at least I hope).  I wouldn't be shocked to see hooks get added in patches.

Sesquipedalian:
Looks like 3-2 in favour.

PR #6916, opened 2021-08-05
Remove unused text string

### LINKED ISSUE #6618: $txt['disable_recent_posts']: Quote reference to $txt['recent_posts']?
https://github.com/SimpleMachines/SMF2.1/blob/ad0ce0244f61a2f803756d503116252f2f594ce9/Themes/default/languages/Themes.english.php#L84

"recent posts" should be quoted, since it refers to the feature of the same name, no?

DiegoAndresCortes:
Fixes #6618 

I actually couldn't find this text string, so seems like it's not even being used anywhere?


live627 (approved):
I can't even find it in 2.0!

PR #6919, opened 2021-08-07
Lists smileys even when smiley_files table is empty

### LINKED ISSUE #6918: It needs to be easier to fix missing smf_smiley_files data
#### Description
From the forum:
https://www.simplemachines.org/community/index.php?topic=578392.0

Here's the background as I understand it...  The smiley url/directory settings were incorrect before running the upgrader.  So, the upgrader could not find any smileys to add, leaving smf_smiley_files empty after the upgrader ran.  

Here's the problem:  There is now no filename information anywhere at all.  _And as near as I can tell, no way to correct this in the UI._  The UI assumes you have smf_smiley_files populated for at least one set.

The initial smileys are there in smf_smileys.  Thus, you cannot manually add them, they get reported as duplicates...  Even though smf_smiley_files is empty.

I could not find any way in the UI to populate smf_smiley_files.  

Maybe I'm missing something???

The only way I could find to fix this was to import smf_smiley_files data from a working forum.  Once smf_smiley_files was properly populated for one set, adding any other existing smiley sets worked fine.  But that first set requires a DB import.

### Steps to reproduce
1. Put invalid values in for the smiley url and directory in a 2.0 DB.
2. Run the upgrader.
3. smf_smiley_files is now empty.
4. Attempt to correct...

- Database Type: na
- Database Version: na
- PHP Version: na

Sesquipedalian:
Fixes #6918

With this, smileys will still be listed in the various "Smileys and Message Icons" lists even if the smiley_files table is empty.

Note: this PR does not try to correct bad smiley settings during upgrade, but it does allow the admin to fix missing smiley_file data in the UI afterward.

### DIFF
--- a/Sources/ManageSmileys.php
+++ b/Sources/ManageSmileys.php
@@ -1392,7 +1392,7 @@ function EditSmileys()
 		$request = $smcFunc['db_query']('', '
 			SELECT s.id_smiley AS id, s.code, f.filename, f.smiley_set, s.description, s.hidden AS location
 			FROM {db_prefix}smileys AS s
-				JOIN {db_prefix}smiley_files AS f ON (s.id_smiley = f.id_smiley)
+				LEFT JOIN {db_prefix}smiley_files AS f ON (s.id_smiley = f.id_smiley)
 			WHERE s.id_smiley = {int:current_smiley}',
 			array(
 				'current_smiley' => (int) $_REQUEST['smiley'],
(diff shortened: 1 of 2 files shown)


sbulen (changes requested):
See note regarding dropdown

Sesquipedalian:
> See note regarding dropdown

Er, what note? Where?

sbulen:
If you see this note on some random unrelated issue, please ignore it over there....   🙄

The browse button doesn't seem to work???
image

Sesquipedalian:
Give it a try now, @sbulen. :)

sbulen:
Works now, thanks.

sbulen (approved):

Sesquipedalian:
Excellent

PR #6926, opened 2021-08-10
Requires appropriate permissions to view warning logs

### LINKED ISSUE #473: Moderation center menu permissions
in the moderation center there are almost no permissions as to who can see what in the menu and the referred to pages, beyond if they are a board moderator.

it could be useful to be able to restrict what moderators can see depending on their permissions.

if a moderator can not issue warnings why should they be able to view the warnings that others have issued?

i think the main point is to limit the warnings currently it only allows board moderators... what if a user is allowed to warn users but is not a board moderator? then they are not even allowed to view the warnings in the moderation center.

### LINKED ISSUE #6925: All menu subsections within an area will be hidden if first subsection is disabled
#### Description
For menus created using `createMenu()`, where the menu contains a section that contains an area that contains a subsection, if the first subsection within the area is disabled, then all of the subsections within that area will be hidden.

### Steps to reproduce
If I looked around long enough, I could probably find a way to reproduce this bug naturally, but I stumbled across it while working on something else, so here is an easy way to trigger it:
1. In Admin.php, change this line:
https://github.com/SimpleMachines/SMF2.1/blob/c30c8ee5acee8e4320fdb32af5fa9780dc4b528e/Sources/Admin.php#L67
... to:
 `'editnews' => array($txt['admin_edit_news'], 'edit_news', 'enabled' => false),`
2. Navigate to the ACP in your browser and open the "Main" menu dropdown of the admin menu.
3. Notice how all the subsections of the "News and Newsletters" area are missing, not just the first item: Screen Shot 2021-08-10 at 4 12 56 PM

- Version/Git revision: c30c8ee
- Database Type: N/A
- Database Version: N/A
- PHP Version: N/A

The problem seems to be a bit of broken logic here:
https://github.com/SimpleMachines/SMF2.1/blob/c30c8ee5acee8e4320fdb32af5fa9780dc4b528e/Sources/Subs-Menu.php#L270-L288

### LINKED ISSUE #6974: Upgrader does not migrate 'warning_show' setting to 'view_warning_*' permissions
after smf site was updated from 2.0.18 to 2.1 rc4
members who have been warned, the warning no longer shows up under their usernames in posts
(muted/watched/etc)

Sesquipedalian:
Fixes #473
Fixes #6974
Also fixes #6925 while we're at it.

### DIFF
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -1158,7 +1158,6 @@ function Display()
 		'can_remove_poll' => 'poll_remove',
 		'can_reply' => 'post_reply',
 		'can_reply_unapproved' => 'post_unapproved_replies',
-		'can_view_warning' => 'profile_warning',
 	);
 	foreach ($anyown_permissions as $contextual => $perm)
 		$context[$contextual] = allowedTo($perm . '_any') || ($context['user']['started'] && allowedTo($perm . '_own'));
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -1446,7 +1446,7 @@ function prepareDisplayContext($reset = false)
 	else
 	{
 		// Define this here to make things a bit more readable
-		$can_view_warning = allowedTo('moderate_forum') || allowedTo('view_warning_any') || ($message['id_member'] == $user_info['id'] && allowedTo('view_warning_own'));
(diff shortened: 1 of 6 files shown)


Sesquipedalian:
Oops. Merged before anyone else reviewed. 😞

PR #6928, opened 2021-08-11
[3.0] Simplify the code a little by removing `hide_subsections`

live627:
Also remove the `disabled` flag as it really wasn't needed

- Alternative fix for #6925
- ref #6926


jdarwood007:
@live627,

This is tagged against 3.0 Alpha 5.  Can you refactor it?

Unless this is still causing an issue in 2.1

PR #6939, opened 2021-08-14
Confirm jquery_source is set

### LINKED ISSUE #6938: 8: Undefined index: jquery_source - Load.php line 2476
#### Description
From the forum, & confirmed:
https://www.simplemachines.org/community/index.php?topic=578482.0
image

Two other notes:
1) I suggest the default be 'Local'.  SMF should be self-contained as much as possible.  Plus, we need to account for the fact that some folks install on internal networks, closed off from internet access. 
2) If you go pull up the settings, the form is a little wonky.  Normally the 'custom URL' field is only enabled when 'custom' is selected, but with no values initialized, it shows 'Google CDN' and the field is enabled.
image

### Steps to reproduce
1. Install current GH

- Database Type: na
- Database Version: na
- PHP Version: na

sbulen:
Fixes #6938


Sesquipedalian (changes requested):
It is better to use a CDN as the default for a library like jQuery. All we need to fix this bug is to add `isset($modSettings['jquery_source']) && ` immediately before `array_key_exists` on line 2476.

sbulen:
I disagree for the reasons written up in the original issue:  SMF should be self-contained. It is used occasionally in offline networks.

PR #6959, opened 2021-08-18
Fix another z-index

DiegoAndresCortes:
I didn't investigate this one that much, it's affecting the topbar in SM site.
Not even sure why it was that high in value.

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -2730,7 +2730,7 @@ dl.addrules dt.floatleft {
 .popup_window {
 	position: relative;
 	width: 480px;
-	z-index: 99;
+	z-index: 5;
 	padding: 0 6px 6px 6px;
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Is the issue uniquely particular to the SM site's custom menu at the top, or have you found places where it occurs in a vanilla SMF install?

If it is unique to the SM site's custom menu, then it'd be better to adjust the site's custom CSS, rather than changing SMF's standard CSS.

DiegoAndresCortes:
I think it could be an issue given a specific case like this one. Changing the custom css would be overwriting this exact same class/attribute.

Thoughts @jdarwood007 ?
Screenshot_26

jdarwood007:
That would be a site specific issue.  We need to apply a different z-index to our site menu.

PR #6996, opened 2021-08-21
rename package files by appending a single number to the filename

### LINKED ISSUE #6995: Mod files get renamed when uploaded
#### Description
Mod files get renamed when uploaded. This becomes annoying when searching for it in FTP manually.

### Steps to reproduce
1. upload a mod with the provided form
2. observe the renamed  file

### Previous topics of interest
- https://github.com/SimpleMachines/SMF2.1/pull/3509
- team topic 577161

live627:
- fix https://github.com/SimpleMachines/SMF2.1/issues/6995

### DIFF
--- a/Sources/PackageGet.php
+++ b/Sources/PackageGet.php
@@ -672,10 +672,10 @@ function PackageUpload()
 		fatal_lang_error('package_upload_error_supports', false, array('zip, tgz, tar.gz'));
 
 	// We only need the filename...
-	$packageName = basename($_FILES['package']['name']);
+	$packageName = substr($_FILES['package']['name'], 0, -strlen($match[0]));
 
 	// Setup the destination and throw an error if the file is already there!
-	$destination = $packagesdir . '/' . package_unique_filename($packagesdir, $packageName, $match[1]);
+	$destination = $packagesdir . '/' . package_unique_filename($packagesdir, $packageName, $match[1]) . $match[0];
 	if (file_exists($destination))
 		fatal_lang_error('package_upload_error_exists');
 
--- a/Sources/PackageGet.php
+++ b/Sources/PackageGet.php
@@ -655,7 +655,7 @@ function PackageUpload()
 
 	// Setup the correct template, even though I'll admit we ain't downloading ;)
 	$context['sub_template'] = 'downloaded';
-	$allowext = array('.zip', '.tgz', '.gz');
+
 	// @todo Use FTP if the Packages directory is not writable.
 
 	// Check the file was even sent!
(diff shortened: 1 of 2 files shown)


Sources/PackageGet.php
```diff
@@ -666,19 +666,16 @@ function PackageUpload()
...
+
+	$found_ext = preg_match('/\.(zip|tgz|tar\.gz)$/i', $_FILES['package']['name'], $match);
+	if ($found_ext === 0)
 		fatal_lang_error('package_upload_error_supports', false, array('zip, tgz, tar.gz'));
-	}
 
 	// We only need the filename...
-	$extension = ($extension == '.gz') ? '.tar.gz' : $extension;
-	$packageName = time() . $extension;
+	$packageName = basename($_FILES['package']['name']);
```
Sesquipedalian:
```suggestion
	$packageName = substr($_FILES['package']['name'], 0, -strlen($match[0]));
```

Sesquipedalian (changes requested):

Sources/Subs-Package.php
```diff
@@ -3007,6 +3007,26 @@ function package_crypt($pass)
...
+/**
+ * @param string $dir
+ * @param string $filename The filename without an extension
+ * @param string $ext
+ * @return string The filename with a number appended but no extension
+ * @since 2.1
+ */
+function package_unique_filename($dir, $filename, $ext)
+{
+	if (file_exists($dir . '/' . $filename . $ext))
```
Sesquipedalian:
```suggestion
	if (file_exists($dir . '/' . $filename . '.' . $ext))
```

Sources/Subs-Package.php
```diff
@@ -3007,6 +3007,26 @@ function package_crypt($pass)
...
+	{
+		$i = 1;
+		while (file_exists($dir . '/' . $filename . '_' . $i . $ext))
```
Sesquipedalian:
```suggestion
		while (file_exists($dir . '/' . $filename . '_' . $i . '.' . $ext))
```

Sources/Subs-Package.php
```diff
@@ -3007,6 +3007,26 @@ function package_crypt($pass)
...
+			$i++;
+		$filename .= '_' . $i;
```
Sesquipedalian, line 3024:
```suggestion
		$filename .= '_' . $i . '.' . $ext;
```

Sesquipedalian (changes requested):

Sesquipedalian:
I clicked submit too soon on the first review. The changes from both reviews are needed. 🙂

Sources/PackageGet.php
live627: Don't trust `basename()`?

Sesquipedalian: I trust it to give us, e.g., `derp.tar.gz`. But in `package_unique_filename()`, the `$filename` variable was being handled as if it contained only `derp`.

Sesquipedalian: To be clear, the version of `package_unique_filename()` in c531cfd was first checking if `packages/derp.tar.gztar.gz` existed, and then if `packages/derp.tar.gz_1tar.gz` existed, before attempting to rename to `derp.tar.gz_2tar.gz`

Sesquipedalian: Perhaps you were thinking of `pathinfo($_FILES['package']['name'], PATHINFO_FILENAME)`? That would in theory have provided what `package_unique_filename()` seems to expect, except that it would only remove `.gz` instead of `.tar.gz` from the end of the name.

Sources/Subs-Package.php
Sesquipedalian: 5af7569 missed this one.

Sesquipedalian:
Missed the change on line 3024. 🙂

Sources/PackageGet.php
live627: I didn't use `pathinfo()` for that very reason. Hence the regex.

Sources/PackageGet.php
```diff
@@ -666,19 +666,16 @@ function PackageUpload()
...
 
 	// Setup the destination and throw an error if the file is already there!
-	$destination = $packagesdir . '/' . $packageName;
-	// @todo Maybe just roll it like we do for downloads?
+	$destination = $packagesdir . '/' . package_unique_filename($packagesdir, $packageName, $match[1]);
```
live627:
```suggestion
	$destination = $packagesdir . '/' . package_unique_filename($packagesdir, $packageName, $match[1]) . $match[0];
``` 
let's see if this works

live627:
I plan to eventually integrate the package backup with this in a future version. The backup routine uses both `.tar` and `.tar.gz` which is why this new function omits returning the file extension.

https://github.com/SimpleMachines/SMF2.1/blob/427f32007c4316cfef413d9bcf5d88c406758319/Sources/Subs-Package.php#L3075-L3087
https://github.com/SimpleMachines/SMF2.1/blob/427f32007c4316cfef413d9bcf5d88c406758319/Sources/Subs-Package.php#L3107-L3117

Sesquipedalian:
> I plan to eventually integrate the package backup with this in a future version. The backup routine uses both `.tar` and `.tar.gz` which is why this new function omits returning the file extension.

Makes sense.

Sesquipedalian (approved):

PR #7009, opened 2021-08-23
Avoid multiple quotes for every selection

### LINKED ISSUE #6960: Quote selected text inserts a second quote with gibberish
I tried out this on SM Site and when clicking on the _quote selected text_ it inserts the correct quote plus an extra quote with some random text.

Sometimes it inserts the rubbish quote **before** the good one, and sometimes **after**.

e.g.
`[quote author=Gary link=msg=4094098 date=1629206116]announced[/quote]
[quote author=Gary link=msg=4094098 date=1629206116]Well, you say that but way back in the day SM.org ran 2.0 when it was still in alpha and before 2.0 itself was even announced, mindsets have changed over the years. [/quote]
`

BrickOzp:
If you do multiple text selections in the same message,
and then click the "Quote selected text" button, you
would get multiple quotes with every text you selected.
Remove the old click listener before adding a new on.
This will only quote the latest selected text.

Fixes #6960

### DIFF
--- a/Themes/default/scripts/quotedText.js
+++ b/Themes/default/scripts/quotedText.js
@@ -129,6 +129,9 @@ $(function() {
 		// Show the "quote this" button.
 		$('#quoteSelected_' + oSelected.msgID).show();
 
+		// Remove any previous selection
+		$(document).off('click', '#quoteSelected_' + oSelected.msgID + ' a');
+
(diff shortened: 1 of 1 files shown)


Sesquipedalian (approved):
> This will only quote the latest selected text.

In my tests, it quoted whichever chunk of text was closest to the beginning of the post, rather than anything to do with the chronological order in which the chunks of text were selected.

Be that as it may, this does seem to solve the issue quite nicely.

PR #7010, opened 2021-08-23
Do not show deleted pms in the pm popup

### LINKED ISSUE #6994: Deleted PMs show in popup
#### Description
Screen Shot 2021-08-20 at 4 03 27 PM

### Steps to reproduce
1. User 1: Send a PM to 2 or more members
2. User 2: Prune Personal Messages and delete all messages
3. User 2: The PM popup still shows these even though they are not visible anywhere in personal messages

- Version/Git revision: Latest git
- Database Type: mysql
- Database Version: 5.7
- PHP Version: 8.0

I noticed this on SM web, but it keeps showing me unread PMs there.  But I believe we just need to run a find and repair errors to correct that.

BrickOzp:
The pm popup listed all unread pms, even if they where
deleted. Changed so that deleted pms don't show up.

Fixes #6994

### DIFF
--- a/Sources/PersonalMessage.php
+++ b/Sources/PersonalMessage.php
@@ -397,10 +397,12 @@ function MessagePopup()
 		FROM {db_prefix}pm_recipients AS pmr
 		WHERE pmr.id_member = {int:current_member}
 			AND is_read = {int:not_read}
+			AND deleted = {int:not_deleted}
 		ORDER BY id_pm',
 		array(
 			'current_member' => $context['user']['id'],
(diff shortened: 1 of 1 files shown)


jdarwood007:
What about the fact that they still show up?  Shouldn't that be corrected as well?

BrickOzp:
> What about the fact that they still show up? Shouldn't that be corrected as well?

New solution should fix that

jdarwood007 (approved):

live627 (approved):

PR #7011, opened 2021-08-23
Remove unnecessary br

BrickOzp:
The br will cause a bottom margin for the legend on the
Message Index on small screens. I have not found any need
for it on bigger screens either.


BrickOzp:
Before
before

After
after

Sesquipedalian (approved):

PR #7013, opened 2021-08-23
Don't break the menu items in the user menu

BrickOzp:
Some languages have problems that some menu items
are too long and gets broken into two lines.
Avoid this.

Reported in
https://www.simplemachines.org/community/index.php?topic=578572.0
https://www.simplemachines.org/community/index.php?topic=578575.0

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -805,6 +805,7 @@ img.sort, .sort {
 	text-indent: -24px;
 	-webkit-hyphens: auto;
 	hyphens: auto;
+	white-space: nowrap;
 }
 /* Fixes bug with border-box on scrollable js */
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
This is not a good solution. The profile menu links need to be able to wrap if they are too long. That's why they have `hyphens: auto` in their CSS.

Sesquipedalian:
There's actually a lot of weirdness and inconsistency with those top menus. The language strings, the user's nickname, the alert and PM counters can all affect the width of the dropdown menus. Moreover, at some screen widths the menus can overflow the viewport.

I've got an idea about how to resolve all these problems and make the layout of these menus consistent in all situations. I'll put together a PR soon.

PR #7019, opened 2021-08-25
Add loop protection to PM Labels upgrade

### LINKED ISSUE #6944: PM Labels upgrade loop protection
#### Description
```
---# Moving label info to new tables and updating rules (May be slow!!!)
```

Upon reviewing this.  There is no loop.  This should be fixed.  If you have a high number of members with labels, this will timeout and break the upgrade.  As well the members table is modified by code later on.  Which with a large forum, this will timeout for sure.

BrickOzp:
To avoid timeouts during upgrades for forums with many users.

Fixes #6944

### DIFF
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -2045,157 +2045,219 @@ ADD COLUMN in_inbox TINYINT(3) NOT NULL DEFAULT '1';
 	$results = $smcFunc['db_list_columns']('{db_prefix}members');
 	if (in_array('message_labels', $results))
 	{
-		// They've still got it, so pull the label info
-		$get_labels = $smcFunc['db_query']('', '
-			SELECT id_member, message_labels
+		$_GET['a'] = isset($_GET['a']) ? (int) $_GET['a'] : 0;
+		$step_progress['name'] = 'Moving pm labels';
+		$step_progress['current'] = $_GET['a'];
+
+		$request = $smcFunc['db_query']('', '
+			SELECT COUNT(*)
 			FROM {db_prefix}members
 			WHERE message_labels != {string:blank}',
 			array(
 				'blank' => '',
 			)
 		);
+		list($maxMembers) = $smcFunc['db_fetch_row']($request);
+		$smcFunc['db_free_result']($request);
 
-		$inserts = array();
-		$label_info = array();
-		while ($row = $smcFunc['db_fetch_assoc']($get_labels))
+		$limit = 5000;
+		$is_done = false;
+
+		while (!$is_done)
 		{
-			// Stick this in an array
-			$labels = explode(',', $row['message_labels']);
-
-			// Build some inserts
-			foreach ($labels AS $index => $label)
-			{
-				// Keep track of the index of this label - we'll need that in a bit...
-				$label_info[$row['id_member']][$label] = $index;
-			}
-		}
-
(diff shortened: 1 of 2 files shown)


other/upgrade_2-1_mysql.sql
```diff
@@ -2045,157 +2045,208 @@ ADD COLUMN in_inbox TINYINT(3) NOT NULL DEFAULT '1';
...
+			$step_progress['current'] = $_GET['a'];
 
-		$smcFunc['db_free_result']($get_pm_rules);
+			if ($step_progress['current'] >= $maxMembers)
+				$is_done = true;
+		}
 
 		// Lastly, we drop the old columns
 		$smcFunc['db_remove_column']('{db_prefix}members', 'message_labels');
 		$smcFunc['db_remove_column']('{db_prefix}pm_recipients', 'labels');
```
jdarwood007, line 2258: Could we move these to separated sub steps?  members table can be fairly large, so it may take time to modify it on larger forums.

BrickOzp: I don't think it is possible to divide it any further since they all depend on the first part where the old index is converted to the new. You have do it all for the same member at the same time.

Sesquipedalian: > Could we move these to separated sub steps? members table can be fairly large, so it may take time to modify it on larger forums.

Between the prudently low `$limit = 5000;` and the call to `nextSubstep()` that @BrickOzp has added in this PR, I think this will be well protected from time outs on even the largest forums.

other/upgrade_2-1_mysql.sql
```diff
@@ -2045,157 +2045,208 @@ ADD COLUMN in_inbox TINYINT(3) NOT NULL DEFAULT '1';
...
+							'actions' => $actions,
+							'id_rule' => $row['id_rule'],
+						)
+					);
 				}
 			}
 
-			// Put this back into a string
-			$actions = serialize($actions);
+			$smcFunc['db_free_result']($get_pm_rules);
```
Sesquipedalian, line 2248: I think it might be good to add the following somewhere around here:
```
			$smcFunc['db_query']('', '
 				UPDATE {db_prefix}members
 				SET message_labels = {string:blank}
 				WHERE id_member IN ({array_int:member_list})',
 				array(
 					'blank' => '',
 					'member_list' => $member_list,
 				)
 			);
```

I know we are just going to drop the column when this step is done, but if the process is interrupted for any reason, this will ensure that members we have already processed will not be retrieved again in the `$get_labels` query when the process resumes.

Sesquipedalian (changes requested):

other/upgrade_2-1_mysql.sql
jdarwood007: But it shouldn't mater because it won't do that substep until it completes the previous substep?

other/upgrade_2-1_mysql.sql
jdarwood007: Good idea.  Only reason sm website made it through is because we used the CLI to upgrade, avoiding any issues with timeouts.  Although we had some crashes in the upgrader that have now been fixed.

BrickOzp: Done

Sesquipedalian (approved):

PR #7023, opened 2021-08-26
Calendar date processing - ensure ints

### LINKED ISSUE #7020: Calendar needs to handle distant future dates more gracefully...
#### Description
If you enter a future date, e.g., in 2099, the Calendar reports an error.

**_PHP 7.4 issues 6 errors, 3 pairs of this:_**
image

**_PHP 8.0 gives a white screen with this error:_**
image

**_Note a different but similar error was reported in the forum:_**
```
ManageCalendar.php (Line 300)
mktime(): Epoch doesn't fit in a PHP integer 

```

### Steps to reproduce
1. Enter a date a century out from today

- Database Type: mysql
- Database Version: 5.7
- PHP Version: 7.4 & 8.0

From the forum:
https://www.simplemachines.org/community/index.php?topic=578657.0

sbulen:
Fixes #7020 

The problem is that if the # is too large for an int on that platform (32-bit platforms) PHP converts it to a float.  Keeping it as an int prevents lots of errors in the log, while making behavior consistent with what is seen on 64-bit platforms.  Values beyond 1/19/2038 are not accepted on either platform; now we get a consistent error message on both.

**_BEFORE, 32-bit, entering large year values, PHP < 8:_**

**_BEFORE, 32-bit, entering large year values, PHP = 8:_**
image

**_After, 32-bit (which is now in alignment with 64-bit behavior):_**

### DIFF
--- a/Sources/Subs-Calendar.php
+++ b/Sources/Subs-Calendar.php
@@ -1769,12 +1769,12 @@ function buildEventDatetimes($row)
 	$end['datetime'] = date_format($start_object, 'Y-m-d H:i:s');
 
 	// ISO formatted datetime string, relative to UTC (e.g. '2016-12-29T05:45:30+00:00')
-	$start['iso_gmdate'] = gmdate('c', $start['timestamp']);
-	$end['iso_gmdate'] = gmdate('c', $end['timestamp']);
+	$start['iso_gmdate'] = gmdate('c', (int) $start['timestamp']);
+	$end['iso_gmdate'] = gmdate('c', (int) $end['timestamp']);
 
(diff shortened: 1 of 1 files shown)


live627:
Do you think that the typecasts could be moved to where the timestamps are parsed a few  lines above?
https://github.com/SimpleMachines/SMF2.1/blob/5bb3d79a49c33c7ec6df2d45c7b724e8392aaa27/Sources/Subs-Calendar.php#L1764-L1765

sbulen:
Yep.  Tests OK.  I was considering doing that.

live627 (approved):

PR #7024, opened 2021-08-27
Fix upload of new packages

### LINKED ISSUE #7022: Cannot upload mods in Package Manager (403 Error)
#### Description
After installing latest Github build 697d1f3 on my test forum, no one mod package cannot be uploaded in Package Manager.
I get 'Error 403' on the white screen. Reinstalling RC4 resolve this problem.

### Steps to reproduce
1. Install latest build 697d1f3
2. Open Package Manager and try to upload some mod package

- Version/Git revision: 697d1f3
- Database Type: MariaDB
- Database Version: 10.3.31
- PHP Version: 8.0.9

BrickOzp:
Also removed logging of errors.

Fixes #7022

### DIFF
--- a/Sources/PackageGet.php
+++ b/Sources/PackageGet.php
@@ -660,9 +660,9 @@ function PackageUpload()
 
 	// Check the file was even sent!
 	if (!isset($_FILES['package']['name']) || $_FILES['package']['name'] == '')
-		fatal_lang_error('package_upload_error_nofile');
+		fatal_lang_error('package_upload_error_nofile', false);
 	elseif (!is_uploaded_file($_FILES['package']['tmp_name']) || (ini_get('open_basedir') == '' && !file_exists($_FILES['package']['tmp_name'])))
-		fatal_lang_error('package_upload_error_failed');
+		fatal_lang_error('package_upload_error_failed', false);
 
 	// Make sure it has a sane filename.
 	$_FILES['package']['name'] = preg_replace(array('/\s/', '/\.[\.]+/', '/[^\w_\.\-]/'), array('_', '.', ''), $_FILES['package']['name']);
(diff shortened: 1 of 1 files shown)


Sources/Subs-Package.php
```diff
@@ -3016,13 +3016,10 @@ function package_crypt($pass)
  */
 function package_unique_filename($dir, $filename, $ext)
 {
-	if (file_exists($dir . '/' . $filename . '.' . $ext))
```
BrickOzp: At the time this function is called, the file was not moved to the destination yet. So it does not exists and the input filename will always be used.

Sources/PackageGet.php
```diff
@@ -673,18 +673,19 @@ function PackageUpload()
...
-		fatal_lang_error('package_upload_error_exists');
+		fatal_lang_error('package_upload_error_exists', false);
 
 	// Now move the file.
 	move_uploaded_file($_FILES['package']['tmp_name'], $destination);
 	smf_chmod($destination, 0777);
 
 	// If we got this far that should mean it's available.
-	$context['package'] = getPackageInfo($packageName);
+	$context['package'] = getPackageInfo($packageFileName);
```
BrickOzp, line 688: Need to pass the filename with extension here

Sources/PackageGet.php
```diff
@@ -700,7 +701,7 @@ function PackageUpload()
 	{
 		while ($package = readdir($dir))
 		{
-			if ($package == '.' || $package == '..' || $package == 'temp' || $package == $packageName || (!(is_dir($packagesdir . '/' . $package) && file_exists($packagesdir . '/' . $package . '/package-info.xml')) && substr(strtolower($package), -7) != '.tar.gz' && substr(strtolower($package), -4) != '.tgz' && substr(strtolower($package), -4) != '.zip'))
+			if ($package == '.' || $package == '..' || $package == 'temp' || $package == $packageFileName || (!(is_dir($packagesdir . '/' . $package) && file_exists($packagesdir . '/' . $package . '/package-info.xml')) && substr(strtolower($package), -7) != '.tar.gz' && substr(strtolower($package), -4) != '.tgz' && substr(strtolower($package), -4) != '.zip'))
```
BrickOzp, line 704: Need to check with extension here as well

BrickOzp:
@live627 can you check this, since it modifies your commit

Sources/Subs-Package.php
live627: The extra number should only be appended if the input file does not exist, similar  to how the package backup does it.
https://github.com/SimpleMachines/SMF2.1/blob/4bbda49cf841a5008539a72b75015dce610e31d4/Sources/Subs-Package.php#L3099-L3105

BrickOzp: Makes sense. I have put it back

live627 (approved):

PR #7029, opened 2021-08-31
"Check All" Should be for all

DiegoAndresCortes:
This guy is showing on each category, should be for checking/selecting all boards.


Sesquipedalian:
Maybe I'm going blind, but for the life of me I can't find anywhere in SMF 2.1 that we actually use this bit of the template. Do you have somewhere that I can look in order to see this change in action?

DiegoAndresCortes:
I couldn't find it either on default SMF sections, came across testing a mod, it's a type of setting for _show_settings_
Should probably mention that this is related to the type **boards** which allows to select/check boards for settings.
The issue is that it's showing under each individual category (of boards)

Sesquipedalian:
Fair enough. The code looks good to me, at any rate.

PR #6754, opened 2021-05-22
Revamps autolinker to fix bugs and improve efficiency

Sesquipedalian:
Fixes #6685
Closes #6691
Closes #6692
Fixes #6497

This uses some more advanced PCRE techniques that I have learned since the last time I revamped the autolinker. Not only does this solve bugs (particularly the aforementioned showstopper bug involving parentheses in an URI), but it also improves efficiency by making more extensive use of PCRE subroutines. Additionally, this takes advantage of PCRE conditional logic to test for specific conditions based on the URI scheme, allowing us to do better at avoiding false positives. Along the way, I also found and fixed a couple of edge case bugs in my `validate_iri()`, `sanitize_iri()`, `iri_to_url()`, and `url_to_iri()` functions.

PR #7032, opened 2021-09-01
Uses idn_to_* functions to ensure domain names are normalized correctly

### LINKED ISSUE #7025: Autolinker (and other places) might incorrectly encode non-ASCII domain names
#### Description
There are several places where we use the Punycode class to convert domain names with non-ASCII characters to/from Punycode encoding. The class itself works fine, but it assumes that the UTF-8 strings that are passed to it have already been normalized using the [NFKC](https://www.unicode.org/reports/tr15/#Norm_Forms) Unicode normalization form—basically meaning that sequences of letters plus combining marks are converted to precomposed characters and/or rearranged to a canonical order. This means that Class-Punycode.php is not a complete polyfill for the intl extension's `idn_to_ascii()` function, which takes care of both the normalization and the actual encoding. But because we have been using it as though it were a complete polyfill, we are sometimes producing incorrectly encoded URLs when we try to convert IRIs into pure ASCII URLs.

For a concise demonstration of the problem, see the results produced [here](https://3v4l.org/pKPAk).

### Steps to reproduce
1. Make sure the auto linker is enabled.
2. Type "täst.de" (using the decomposed `a` and `̈` characters) into the post editor and click the Preview button.
3. Inspect the URL of the link in the preview and notice that it is encoded as `http://xn--tast-wwc.de` rather than `http://xn--tst-qla.de` like is should be.

- Version/Git revision: 2f6e030
- Database Type:
- Database Version:
- PHP Version:

1. We either need to replace Class-Punycode.php with some other polyfill for the `idn_to_ascii()` and `idn_to_utf8()` functions, or else supplement it with a polyfill for the Normalizer class.

2. The cleanest approach to fixing this would be to use Subs-Compat.php to add compatibility functions for `idn_to_ascii()` and `idn_to_utf8()`, which could in turn call whatever polyfill library or libraries we need. Then in the places where we currently use the Punycode class directly, we could just replace those calls with calls to `idn_to_ascii()` or `idn_to_utf8()` as needed.

### LINKED ISSUE #7046: Output of utf8_strtolower() does not match output of mb_strtolower()
#### Description
The utf8_strtolower() and utf8_strtoupper() are supposed to produce the same output as the mb_strtolower() and mb_strtoupper() functions, but they don't.

### Steps to reproduce
The easiest way to reproduce this for testing purposes is to use the following bit of code to compare the output of the mb_strto* functions to the output of the utf8_strto* functions.
```
<?php

define('SMF', 1);
include "/path/to/SMF2.1/Sources/Subs-Charset.php";

for ($i = 0; $i <= 0x10FFFF; $i++) {
	$char = mb_chr($i, 'UTF-8');

	if ($char === false)
		continue;

	$mb_upper = mb_strtoupper($char, 'UTF-8');
	$utf8_upper = utf8_strtoupper($char);

	$mb_lower = mb_strtolower($char, 'UTF-8');
	$utf8_lower = utf8_strtolower($char);

	if ($mb_upper !== $utf8_upper || $mb_lower !== $utf8_lower)
	{
		echo $char, ":\n";

		if ($mb_upper !== $utf8_upper)
		{
			echo "mb_upper: ", var_export(unpack('C*', $mb_upper), true), "\n";
			echo "utf8_upper: ", var_export(unpack('C*', $utf8_upper), true), "\n";
		}

		if ($mb_lower !== $utf8_lower)
		{
			echo "mb_lower: ", var_export(unpack('C*', $mb_lower), true), "\n";
			echo "utf8_lower: ", var_export(unpack('C*', $utf8_lower), true), "\n";
		}

		echo "\n";
	}
}
```

- Version/Git revision: 03bc9f6
- Database Type:
- Database Version:
- PHP Version:

Sesquipedalian:
Fixes #7025

 The main thing to fix the bug is just to [use `idn_to_ascii()` and `idn_to_utf8()`](https://github.com/SimpleMachines/SMF2.1/commit/a605ee951b2e0ed03933df70f766d8829c655656) to ensure that domain names containing non-ASCII characters are normalized to their canonical forms (e.g. `t` `a` `¨` `s` `t` `.` `d` `e`  is normalized to  `t` `ä` `s` `t` `.` `d` `e`).

Why do we need to do this? The IANA imposes such normalization of non-ASCII characters for all registered domain names in order to reduce the possibilities for ambiguous domain names, which means that `t` `a` `¨` `s` `t` `.` `d` `e` can never be registered as a domain name—only  `t` `ä` `s` `t` `.` `d` `e` can. This means that if one of our users types `h` `t` `t` `p` `:` `/` `/` `t` `a` `¨` `s` `t` `.` `d` `e` into a post, we need to perform the same normalization on the domain name in our generated links, or else we will end up pointing users to URLs that don't exist.

... But I ran into a further problem: `idn_to_ascii()` and `idn_to_utf8()` are part of the [intl](https://www.php.net/manual/en/book.intl.php) extension, and many PHP installs do not have that extension enabled. So I created [polyfills for them](https://github.com/SimpleMachines/SMF2.1/commit/3883d0264aabafa2a23df3a3f809a4903ee812ab) in Subs-Compat.php. I didn't bother making these do anything more than what we actually need, so they're pretty short.

... But I ran into a further problem: the native `idn_to_ascii()` function internally calls the Unicode Normalization algorithms that are otherwise exposed via the `normalizer_normalize()` function of the intl extension's [Normalizer](https://www.php.net/manual/en/class.normalizer.php) class. But of course, if the intl extension is unavailable then `normalizer_normalize()` is also unavailable, so I had to create a [polyfill for that](https://github.com/SimpleMachines/SMF2.1/commit/eeea8e0e1752178d8264ef79c0f808a0d28dacf8), too. This one needed to be a complete polyfill, so it is. In this case, it seemed best to create functions in Subs-Charset.php that would serve as wrappers for `normalizer_normalize()` if it is available, or that would implement the Unicode Normalization algorithms themselves if `normalizer_normalize()` is not available.

... But I ran into one more problem: my `normalizer_normalize()` polyfills use `mb_ord()` and `mb_chr()` at a couple of key points, but `mb_ord()` and `mb_chr()` were only added in PHP 7.2.0, so I needed to create [polyfills for them](https://github.com/SimpleMachines/SMF2.1/commit/0b7763a1fc8923f0c28b136b208d00ac47203f57) in Subs-Compat.php, too. These are also complete polyfills.

Now, before anyone gets worried about large PRs, let me point out two things:
1. Almost all of this PR is just polyfills, not changes to SMF's behaviour. **Only a605ee9 actually changes any behaviour.** If you have a PHP system with the intl extension enabled, you can test that lone commit without involving any of the other commits at all.
2. I have delicious, nutritious **unit tests** for the polyfills right here: 
    - [Unit testing for SMF's mb_ord() and mb_char() polyfills](https://gist.github.com/Sesquipedalian/0f8ab97b2b73d68d16eaa8fae56f8ed6)
    - [Unit testing for SMF's normalizer_normalize() polyfill](https://gist.github.com/Sesquipedalian/b0870a303967002adc9db93fc48cd919)
    - For the idn_to_* polyfills, just put the list of URLs [here](https://github.com/SimpleMachines/SMF2.1/pull/6754#issuecomment-846471296) into a post directly in SMF.

-----

Also fixes #7046


live627:
Thanks for all the tests. I just started testing, and in doing so, I noticed that the  third reproduction step did not get transformed at all.

Sesquipedalian:
> Thanks for all the tests. I just started testing, and in doing so, I noticed that the third reproduction step did not get transformed at all.

To clarify, when you say "the third reproduction step" do you mean this part?

>    * For the idn_to_* polyfills, just put the list of URLs [here](https://github.com/SimpleMachines/SMF2.1/pull/6754#issuecomment-846471296) into a post directly in SMF.

If so, I'm not sure what you mean. Here's the result in my tests:
Screen Shot 2021-09-03 at 12 07 41 PM
You need to inspect the source to see the Punycode host name in the href attribute for the relevant links.

Sources/Subs-Charset.php
```diff
@@ -550,4 +550,7794 @@ function utf8_strtoupper($string)
...
+		// Singleton replacements.
+		if (isset($substitutions[$string[$c]]))
+			$string[$c] = $substitutions[$string[$c]];
+
+		// Hangul characters.
+		// See "Hangul Syllable Composition" in the Unicode standard, ch. 3.12.
+		if ($string[$c] >= "\xE1\x84\x80" && $string[$c] <= "\xE1\x84\x92" && $string[$c + 1] >= "\xE1\x85\xA1" && $string[$c + 1] <= "\xE1\x85\xB5")
+		{
+			if (!function_exists('mb_ord'))
+				require_once($sourcedir . '/Subs-Compat.php');
```
live627, line 3050: Undefined global

Sources/Subs-Charset.php
```diff
@@ -550,4 +550,7794 @@ function utf8_strtoupper($string)
...
+	$string = utf8_normalize_kd($string);
+
+	return utf8_compose($string);
+}
+
+/**
+ * Helper function for utf8_normalize_c and utf8_normalize_kc
+ */
+function utf8_compose($string)
+{
```
live627, line 3031:
```suggestion
function utf8_compose($string)
{
	global $sourcedir;
```

live627 (changes requested):

Sources/Subs-Charset.php
Sesquipedalian: Good catch

PR #7037, opened 2021-09-02
fix(notifications): Send user's real name in emails

### LINKED ISSUE #7033: Usernames in email notifications
#### Description
After enabling email notifications, I receive usernames instead of display names on the notifications.
I'm using the "Receive message body in e-mails" option.

I've only tested/used it on SM site.

live627:
We cannot rely on `$posterOptions['name']` because it is populated with `member_name`.

Fix #7033

### DIFF
--- a/Sources/tasks/CreatePost-Notify.php
+++ b/Sources/tasks/CreatePost-Notify.php
@@ -597,7 +597,7 @@ class CreatePost_Notify_Background extends SMF_BackgroundTask
 
 				$replacements = array(
 					'TOPICSUBJECT' => $parsed_message[$localization]['subject'],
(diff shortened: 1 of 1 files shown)


live627:
@Arantor is `member_name` in the alerts table supposed to be populated by `member_name` (matches `poster_name` in the messages table) or `real_name`?

Arantor:
It should be whatever would normally be displayed when the alert is made and what is posted in topic view (should match real_name = display name) so that if the account is deleted, the alert isn't mangled and is still meaningful.

At least that was the design intent, but I don't know if that changed back and forth over the years.

Sesquipedalian (approved):

PR #7040, opened 2021-09-03
Improves performance processing serialized data during upgrade

### LINKED ISSUE #6947: Serialize to json slow
#### Description
Serialize to json is very slow.  On average is 50 every second it converts.  On large forums.  This is horribly slow.

This may not be a bug, but something we want to analyze for performance

```
								$temp = @safe_unserialize($row[$col]);

								// Maybe we can fix the data?
								if ($temp === false)
									$temp = @safe_unserialize(fix_serialized_data($row[$col]));

								// Maybe the data is already JSON?
								if ($temp === false)
									$temp = smf_json_decode($row[$col], true, false);
```

We should run this code block through 1,000 iterations, divide by time and get the average time taken.  Then see if we can optimize it.

Could we combine fix_serialized_data and safe_unserialize together to reduce the times we unserialize data?

Do we want to skip smf_json_decode and use native json_decode?  We use smf_json_decode to future proof ourselves to handling any data issues.  We could forgo that for the upgrader and deal with it differently if needed.

Sesquipedalian:
Fixes #6947

Also closes a potential hole in _safe_serialize().

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -6573,6 +6573,19 @@ function _safe_serialize($value)
 
 	if (is_array($value))
 	{
+		// Check for nested objects or resources.
+		$contains_invalid = false;
+		array_walk_recursive(
+			$value,
+			function($v) use (&$contains_invalid)
+			{
+				if (is_object($v) || is_resource($v))
+					$contains_invalid = true;
+			}
+		);
+		if ($contains_invalid)
+			return false;
+
 		$out = '';
 		foreach ($value as $k => $v)
 			$out .= _safe_serialize($k) . _safe_serialize($v);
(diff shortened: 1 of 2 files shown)


jdarwood007:
I like it much better.

https://gist.github.com/jdarwood007/ecc189e761d629697395af2f764918b7

```
Testing for valid string - Time taken:0.43846988677979
Testing for invalid string - Time taken:1.1400458812714
Testing for JSON string - Time taken:0.064997911453247
Testing for valid string with NEW - Time taken:0.43021416664124
Testing for invalid string with NEW - Time taken:0.42733097076416
Testing for JSON string  with NEW - Time taken:0.05634617805481
Testing for valid string with PR 7040 - Time taken:0.43351912498474
Testing for invalid string with PR 7040 - Time taken:0.43142008781433
Testing for JSON string  with PR 7040 - Time taken:0.024644136428833
```

jdarwood007 (approved):

PR #7042, opened 2021-09-04
Stops us from trying to hash validate a directory.

jdarwood007:
I don't see it as worth it to try to generate a hash for a directory to validate, just check the id

### DIFF
--- a/Sources/Subs-Package.php
+++ b/Sources/Subs-Package.php
@@ -3180,7 +3180,9 @@ function package_validate_installtest($package)
 {
 	global $context;
 
-	$context['package_sha256_hash'] = hash_file('sha256', $package['file_name']);
+	// Don't validate directories.
+	$context['package_sha256_hash'] = is_dir($package['file_name']) ? null : hash_file('sha256', $package['file_name']);
+
 	$sendData = array(array(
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Will the package validation service accept a request without a hash?

jdarwood007:
It will be ignored and continue to check the others like package id.

We could try to hash a folder, but something as simple as a added file or file ending difference could cause a different hash.

Sesquipedalian:
> It will be ignored and continue to check the others like package id.

Fair enough.

> We could try to hash a folder, but something as simple as a added file or file ending difference could cause a different hash.

Looks like just the opposite, actually. In my tests, I've found that *every* directory produces exactly the same hash, regardless of its name or content: `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`. This is the same hash that is generated by `hash('sha256', '')`, which makes sense, since a directory doesn't have any file content of its own.

jdarwood007:
There are methods to hash a directory: https://stackoverflow.com/questions/545387/linux-compute-a-single-hash-for-a-given-folder-contents/545413
But in all its similar, you hash all contents one way or another.

PR #7053, opened 2021-09-08
Fix bbc tags in search results

### LINKED ISSUE #7044: BBC parsed results show in search
#### Description
I think this is related to the fix from #6887

Because we use results after we parse bbc, we sometimes get bbc parsed text strings in search results.
Screen Shot 2021-09-06 at 8 15 06 AM
Screen Shot 2021-09-06 at 8 15 15 AM

I don't know if we can fix this or have to accept this as a side effect from the fix.

### Steps to reproduce
1. Make a post with this content
```
words and words and words and more words... Then Code..

[code]Use my code
[/code]
```
2. Search for "code"

- Version/Git revision: Latest Git
- Database Type:
- Database Version:
- PHP Version:

@sbulen @live627 @Sesquipedalian 
Since you worked with the original PR.  Mentioning you guys on this.

BrickOzp:
Remove code and quote bbc tags from search results that should be truncated. This is beacuse these tags
add some html and that should not be shown in the search results.
Fixes #7044

Avoid the highlighter from matching bbc tags or html attributes.
This will fix search results when searching in a topic.

### DIFF
--- a/Sources/Search.php
+++ b/Sources/Search.php
@@ -2094,11 +2094,11 @@ function prepareSearchContext($reset = false)
 	}
 	else
 	{
-		// Run BBC interpreter on the message.
-		$message['body'] = parse_bbc($message['body'], $message['smileys_enabled'], $message['id_msg']);
-
 		$message['subject_highlighted'] = highlight($message['subject'], $context['key_words']);
 		$message['body_highlighted'] = highlight($message['body'], $context['key_words']);
+
+		// Run BBC interpreter on the highlighted message.
+		$message['body_highlighted'] = parse_bbc($message['body_highlighted'], $message['smileys_enabled'], $message['id_msg']);
 	}
 
 	// Make sure we don't end up with a practically empty message body.
@@ -2319,8 +2319,16 @@ function searchSort($a, $b)
  */
 function highlight($text, array $words)
 {
-	$words = implode('|', array_map('preg_quote', $words));
-	$highlighted = preg_filter('<' . $words . '>iu', '<span class="highlight">$0</span>', $text);
+	$words = implode('[^\]]|[^\[]', array_map('preg_quote', $words));
+
+	$highlighted = preg_replace_callback(
+		'<[^\[]' . $words . '[^\]]>iu',
+		function($matches)
+		{
+			return substr($matches[0], 0, 1) . '<span class="highlight">' . substr($matches[0], 1, -1) . '</span>' . substr($matches[0], -1, 1);
+		},
+		$text
+	);
 
 	if (!empty($highlighted))
 		$text = $highlighted;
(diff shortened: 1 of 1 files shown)


jdarwood007:
Just to be clear, this leaves the code in while stripping just the tags?  Because the code could be a part of the search results.

Sources/Search.php
```diff
@@ -2317,13 +2319,48 @@ function searchSort($a, $b)
  */
 function highlight($text, array $words)
 {
-	$words = implode('|', array_map('preg_quote', $words));
-	$highlighted = preg_filter('<' . $words . '>iu', '<span class="highlight">$0</span>', $text);
+	$words = implode('[^\]]|[^\[]', array_map('preg_quote', $words));
+
+	$highlighted = preg_replace_callback(
+		'<[^\[]' . $words . '[^\]]>iu',
```
Sesquipedalian, line 2325: These regular expressions will not match at the start or end of the string. So if the post begins "Foo bar baz..." and the user searched for "Foo", the initial "Foo" at the start of the post will not be highlighted.

Sesquipedalian (commented):
I'm not a fan of this approach, to be honest. 

First, there's the general problem that running `parse_bbc()` before `highlight()` in one case, but after it in another case, creates inconsistent input and output. The `highlight()` function should either always receive HTML, or always receive BBC, not both.

Second, skipping some matches in the result can cause further problems. For example, if the post only includes the word "Quote" because of the [quote] BBC, and not because of any other occurrences in the text, with the result that nothing is highlighted, this will prevent the truncation to 50 characters on either side of the match from happening.

Finally, there's a user expectations problem. For better or for worse, our search results do return all posts that use the [quote] BBC when the user searches for "Quote". But if the post only includes the word "Quote" because of the [quote] BBC, then the user has no indication *why* the post matched. It looks like a wrong result. (Granted, one could argue that it *is* a wrong result and that the search index should ignore text generated by BBC, but that's a separate matter.)

jdarwood007:
As I mentioned in #7044, this isn't a issue that will be resolved easily.  We have results we need parsed by bbc to make sense and for the highligher to work.  Then we have results we don't want full parsing on and to make the highligther work.  We have gone back and forth a few times on search highlighting issues.  So I think this may be a question for a future version of SMF where we can addresses all the points

1. Search should highlight properly in-between html as expected to produce valid html
2. Search should highlight bbc content correctly but ignoring some html generated input (such as the code/select/expand).  Most users would not expect/want these to be matched/highlighted.  Its not content you are looking for in most cases.
3. Search's may match bbc and these should be clear to the user why.  Such as the word  "quote" could be somebody who said "quote" or the bbc "[quote]"
4. Searches that may match bbc could be excluded, need to be excluded from the search results for a clear reason.  Such as "[b]" shouldn't be searched.

There may be more.

PR #7054, opened 2021-09-08
Remove moved topics from search results

### LINKED ISSUE #7048: Moved topics show up in search results
#### Description
Moved topics show up in search results

### Steps to reproduce
1. Create a topic
2. Move the topic to another board and chose to create a redirect topic
3. Search the entire forum for a word in the topic title
4. Both the topic and the redirect topic show up in the search results

- Version/Git revision: d2f1441c7
- Database Type: MySql
- Database Version:
- PHP Version: 5.6

BrickOzp:
Avoid showing moved topics in search results if possible.
If a topic has id_redirect_topic or redirect_expires set,
they are not shown as search results.

Fixes #7048

### DIFF
--- a/Sources/Search.php
+++ b/Sources/Search.php
@@ -1097,12 +1097,18 @@ function PlushSearch2()
 
 				foreach ($searchWords as $orIndex => $words)
 				{
-					$subject_query_params = array();
+					$subject_query_params = array(
+						'not_redirected' => 0,
+						'never_expires' => 0,
+					);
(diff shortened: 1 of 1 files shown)


Sesquipedalian (approved):

jdarwood007:
Just an added note.  You don't send this logic through any of the APIs.  So the APIs will do their searches as normal and returning the results, which should get filtered here.  But depending on the API, not excluding these could result in longer search times.

PR #7064, opened 2021-09-09
Remove the use of INTL_IDNA_VARIANT_UTS46 constant

### LINKED ISSUE #7063: Autolinking does not work
#### Description
Urls in message does not work.
Possible php 5.6 issue

### Steps to reproduce
1. Have autolinking enabled
2.  Add an url to a message e.g. www.test.com
3. See that the url is not shown in the message 

- Version/Git revision: 77a15caa5ad603c5ecd4a49466719107c2051167
- Database Type: MySql
- Database Version:
- PHP Version: 5.6

INTL_IDNA_VARIANT_UTS46 is undefined, but even if #7062 is applied it does not work.
Reverting a605ee951b2e0ed03933df70f766d8829c655656 solves the issue.

BrickOzp:
This constant is causing problem for php 5.6
Since it is the default parameter there is no need
to send it.

Fixes #7063

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -2785,13 +2785,13 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 
 								if (!empty($host))
 								{
-									$ascii_host = idn_to_ascii($host, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46);
+									$ascii_host = idn_to_ascii($host, IDNA_DEFAULT);
 
 									if ($ascii_host !== $host)
 									{
 										$fullUrl = substr($fullUrl, 0, strpos($fullUrl, $host)) . $ascii_host . substr($fullUrl, strpos($fullUrl, $host) + strlen($host));
 
-										$utf8_host = idn_to_utf8($ascii_host, IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46);
+										$utf8_host = idn_to_utf8($ascii_host, IDNA_DEFAULT);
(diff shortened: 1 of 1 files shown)


sbulen:
Min version for 2.1 is php 7.

No more worrying about php 5.x.

See internal discussion sticky on dev board.

Sesquipedalian:
> Min version for 2.1 is php 7.
> 
> No more worrying about php 5.x.
> 
> See internal discussion sticky on dev board.

Apparently no one ever implemented that decision. See #7065.

Sesquipedalian (changes requested):
> Since it is the default parameter 

It is the default only since PHP 7.4, so it still needs to be set explicitly unless we raise our minimum PHP version to at least 7.4.

Instead of dropping the parameter, it would be better to check whether the constant is defined or not when deciding whether to require Subs-Compat.php. That file defines all the necessary constants if they are not already defined, including INTL_IDNA_VARIANT_UTS46.

Sesquipedalian:
Oh, wait. I see from your comments in #7063 that you already tried my suggestion in #7062 and it didn't work.

Hm...

Sesquipedalian (approved):
On further consideration, this does seem like the best approach. My partial polyfill will only accept INTL_IDNA_VARIANT_UTS46, but the issue only arises when using the native ids_to_* functions, so just rolling with the implicit default should be fine. This does mean that forums running on PHP versions prior to 7.4 with the native idn_to_* functions available will get get different results than everyone else for some strings, but there's not much we can do about that.

PR #7065, opened 2021-09-09
Raises version requirements to PHP 7.0, MySQL 5.6

Sesquipedalian:
As agreed by the team in internal discussion topic 576829.0.


albertlast:
I don't see the pg changes in the commit.

Sesquipedalian:
> I don't see the pg changes in the commit.

The PostgreSQL requirement already was 9.6, so I didn't actually have to change that.

PR #7068, opened 2021-09-10
Turn off autolinking for signatures if tags are disabled

BrickOzp:
If email och url tag are disabled for signatures, but the tags and
autolinking are enabled globally, email and url tags will show up
in the signature. Disable autolinking for signatures if tags are
disabled.


Sesquipedalian:
The iurl BBC should be disabled, too.

Sesquipedalian (approved):

Issue #7060, opened 2021-09-09
Database version not detected correctly in the upgrade.php script (when run through CLI)


websmurf:
#### Description
I'm in the process of testing an upgrade for my SMF 2.0.18 site and ran in to this error when running the upgrade.php script through cli (`php upgrade.php`)

```
Fatal error: Uncaught ArgumentCountError: Too few arguments to function smf_db_query(), 1 passed in /var/www/forum/upgrade.php on line 2594 and at least 2 expected in /var/www/forum/Sources/Subs-Db-mysql.php:361
```

### Steps to reproduce
1. Load in a 2.0.18 database
2. Run the cli upgrade script: `php upgrade.php`

### Environment (complete as necessary)
- Version/Git revision: 2.1 RC4
- Database Type: Mysql 
- Database Version: 8.0.23
- PHP Version: 8.0.5

### Additional information/references

It appears the `$database_version` detection is missing an `eval` call.
At this point the version_check code is being injected into version_compare as is:

In upgrade.php
```php
        if (empty($database_version))
	{
		$database_version = $databases[$db_type]['version_check'];
		$where_field_support = $db_type == 'mysql' && version_compare('5.0', $database_version, '<=');
	}
```

If I change this to:
```php
        if (empty($database_version))
	{
		$database_version = eval($databases[$db_type]['version_check']);
		$where_field_support = $db_type == 'mysql' && version_compare('5.0', $database_version, '<=');
	}
```

The upgrade continues as expected.

websmurf:
There are a few additional errors in the upgrade process, will have a look into that as well:

```
[...]

PR #7072, opened 2021-09-11
Detect database version in `checkChange()`

live627:
Not sure how this was missed before

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -2589,7 +2589,7 @@ function checkChange(&$change)
 			return;
 
 		// Oh where, oh where has my little field gone. Oh where can it be...
-		while ($row = $smcFunc['db_query']($request))
+		while ($row = $smcFunc['db_fetch_assoc']($request))
 			if ($row['Field'] == $temp[1] || $row['Field'] == $temp[2])
 			{
 				$current_type = $row['Type'];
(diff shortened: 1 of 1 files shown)


albertlast:
Instead of eval maybe a function call () would be better?
Also the mysql check the client version not the server version.

sbulen:
Yikes...  Yeah, how has this worked at all...  

Can you reproduce the original issue reported in #7060 ?  If so, how?

sbulen:
I can reproduce #7060 using php8 and current github code.  

I tested with this PR, and I am seeing this error - though it appears to run to completion:
 *  +++  Renaming table columns.Error: Undefined array key 1 File: yada\yada\upgrade.php Line: 2572Error: Undefined array key 1 File: yada\yada\upgrade.php Line: 2572 done.

sbulen:
I think that fixing this will actually break something...

Note that this logic _clearly_ has never worked properly...  For 10 yrs+...  Its purpose is to NOT change the data type of certain fields (passwordSalt, memberGroups) when renaming them (an early 2.0 upgrade step).  Currently, it returns early under all circumstances, meaning, today, it always ensures the column matches the expected data type.  

So...  If we fix it, it may actually leave those fields with an unusual/unexpected datatype going forward.  

At this point, I believe it is safer to remove this function altogether, rather than fix it.

sbulen:
& BTW, the same is true over in SMF 2.0...

Sesquipedalian (dismissed):

Sesquipedalian:
@live627, how much of this still relevant since #7077 was merged? If any is, could you please resolve the conflicts?

PR #7083, opened 2021-09-15
theme cleanup

DiegoAndresCortes:
Fixes:
- https://www.simplemachines.org/community/index.php?topic=578706.0
- https://www.simplemachines.org/community/index.php?topic=578638.0

Some others in that board I flagged as not issues, these ones aren't really either but can be addressed right now.

### DIFF
--- a/Themes/default/BoardIndex.template.php
+++ b/Themes/default/BoardIndex.template.php
@@ -108,7 +108,7 @@ function template_main()
 
 			// Show the last post if there is one.
 			echo'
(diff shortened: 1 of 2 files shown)


StealthWombat:
None of the things I've been reporting are really "issues", in the sense that they don't break default layout or functionality. However, they are going to provide better UX and/or be useful for future custom theming. So, if they can be included before Final, without breaking anything, I think including them would be a good idea.

Sesquipedalian (approved):

PR #7092, opened 2021-09-17
Fixes adding of invisible membergroups (invisible level 2) to a members primary group

### LINKED ISSUE #7089: Invisible Joinable group becomes primary
#### Description
If you have a group that is request-able to join and the group is set to invisible.  Upon approving the join, the group becomes primary.
 
### Steps to reproduce
1.  Make a test group, setting the group type to Requestable and Visibility to Invisible
2. Make a test user
3. Request under test user to join the test group
4. As admin approve this via the moderation center groups
5. User will have the test group become their primary group.

- Version/Git revision: latest git
- Database Type:
- Database Version:
- PHP Version:

Tested on 2.0 and it works as expected with the same steps.

WhistlingDonkey:
Fixes #7089

Fixes adding of invisible membergroups (invisible level 2) to a members primary group from the background task (moderation center). While this fixes the issue at hand there isn't a proper check in the addMembersToGroups() function that could act as a safeguard to check whether a group is only_additional or not.

### DIFF
--- a/Sources/tasks/GroupAct-Notify.php
+++ b/Sources/tasks/GroupAct-Notify.php
@@ -31,7 +31,7 @@ class GroupAct_Notify_Background extends SMF_BackgroundTask
 
 		// Get the details of all the members concerned...
 		$request = $smcFunc['db_query']('', '
-			SELECT lgr.id_request, lgr.id_member, lgr.id_group, mem.email_address,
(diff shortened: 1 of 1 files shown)


WhistlingDonkey:
Something went wrong, it worked and now that I've tidied up my tests now it doesn't. I'll have to try again.

PR #7093, opened 2021-09-17
Fixes adding invisible level 2 membergroups to a members primary group

### LINKED ISSUE #7089: Invisible Joinable group becomes primary

WhistlingDonkey:
Fixes #7089

This is a quick and dirty fix to prevent adding an invisible level 2 group as a members primary group. The culprit of #7089 is the call to addMembersToGroup() in GroupAct-Notify.php where the group type is set to 'auto' without checking if the group type should be set to anything else. I wasn't able to fix the lack of checking in GroupAct-Notify.php effectively so hence this quick and dirty fix.

If someone else can do something more elegant, do it!

### DIFF
--- a/Sources/Subs-Membergroups.php
+++ b/Sources/Subs-Membergroups.php
@@ -544,27 +544,6 @@ function addMembersToGroup($members, $group, $type = 'auto', $permissionCheckDon
 		if ($is_protected == 1)
 			return false;
 	}
-
-	// Safety check for invisible level 2 membergroups to prevent adding it as the members primary group
-	if ($type != 'only_additional')
-	{
-	$request = $smcFunc['db_query']('', '
-			SELECT hidden
-			FROM {db_prefix}membergroups
-			WHERE id_group = {int:current_group}
-			LIMIT {int:limit}',
-			array(
-				'current_group' => $group,
-				'limit' => 1,
-			)
-		);
-		list ($is_hidden) = $smcFunc['db_fetch_row']($request);
-		$smcFunc['db_free_result']($request);
-
-		// Is it hidden?
-		if ($is_hidden == 2)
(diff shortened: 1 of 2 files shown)


jdarwood007:
Don't do a dirty fix, do a proper fix.  In 2.0 this was done all in the initial query.  Can easily be fixed for 2.1

WhistlingDonkey:
I'll rephrase my statement (not having a dig at your comment @jdarwood007, I actually appreciate it). With my limited knowledge of PHP and SQL compared to the developers this is the solution that I ultimately got to work. I did say that if anyone else could do this better to please do so.

Backing up to what I tried, The important parts of `addMembersToGroup()`, ie the queries that act upon `only_additional`, `only_primary` / `force_primary` and `auto` is no different between 2.0 and 2.1 so without adding the safety check in `addMembersToGroup()` like I did then the call in GroupAct-Notify.php has to be corrected. This is where I fell down. 

From what I tried and understand it is not possible to get the hidden value of the group into `lgr` for it to be used in GroupAct-Notify.php without adding a column to the `log_group_requests` table, which I didn't think would be a good idea at this late stage however I think I see where I went wrong with my original fix and will try again.

Sesquipedalian (approved):

PR #7099, opened 2021-09-27
Discourage robots from indexing the credits page

AaronVanGeffen:
Hello, long time no see!

When Googling my name for unrelated reasons, I noticed Google is indexing a lot of SMF credits pages. I think we should discourage this, hence this quick patch. Let me know what you think.

It's been a while since I made a patch to SMF, but it's nice to see the code base still retaining its character. I like that Compuart's weeping angel joke is still in there. :-)


jdarwood007:
Stop by the friends chat sometime!  Nice to see you around.

PR #6786, opened 2021-06-16
Consolidates and simplifies entity handling

Sesquipedalian:
Inspired by discussion in SimpleMachines/SMF2.0#212, this PR consolidates and simplifies a bunch of redundant code that was used for handling HTML entities.

The changes can be summarized as follow:

1. Gets rid of `fixchar__callback()`, `fixchardb__callback()`, `replaceEntities__callback()`, and `entity_fix__callback()`, in favour of the simpler, faster, and more reliable `smf_entity_decode()`, `sanitize_entities()`, and `$smcFunc['entity_fix']()`.
	- Also makes `fix_utf8mb4()` a named function in order to faciliate the above changes.

2. Adds `sanitize_chars()` and `normalize_spaces()` to deal with invalid characters and weird or undesired whitespace.

3. Makes the `$ent_list` regular expression (which is used in many `$smcFunc` functions) much more robust.

As a bonus, this also improves `$smcFunc['html_trim']()` so that it correctly trims `&#160;` and `&#xA0;` as well as `&nbsp;`.

---------

This PR seems a lot bigger than it really is, because there are a lot of files that have a line here or there that can benefit from these improvements. In reality, it is actually not very hard to test all possible routes that are affected by these changes. The following tests will cover literally everything:

- Log in and log out. This tests the changes in `Login2()`.

- Run a search for a string containing entities and characters that should be entities, e.g. `long &amp; short & wide`. This tests the changes in `prepareSearchContext()`.

- Create a poll with an ampersand in the question. This tests the changes in `Post2()`.

- Change the name of the forum to something with a `<`, `>`, or `&` character in it, and then do something that will trigger an email notification to be created. This tests the changes in `mimespecialchars()`.

- Create a new topic, attach an image with an ampersand in its file name to the post, embed that attachment into the post via the `[attach]` BBCode, then view the topic. This tests the changes in `preparsecode()`, `showAttachment()`, and `loadAttachmentContext()`.

- In Administration Center ► Attachments and Avatars ► Browse Files, check that the file name of the attachment from the previous test is displayed correctly. This tests the changes in `BrowseFiles()`.

- In your browser, go to `index.php?action=.xml;type=smf;sa=recent` and `index.php?action=.xml;type=smf;sa=news` to get XML files that contain references to the attachment from the previous test. This tests the changes in `getXmlRecent()` and `getXmlNews()`.

[...]

PR #7102, opened 2021-09-28
Normalizes and sanitizes UTF-8 input

### LINKED ISSUE #7085: Non graphic characters in member names can allow impersonation
#### Description
Couldn't think of a better title...

As discussed at https://www.simplemachines.org/community/index.php?topic=578926 , it is possible for a member to insert non graphic characters into their real name and display name which allows them to have what appears to others the same display name as another member .  This can allow, however futile the attempt, impersonation of another member. At the least it can just cause confusion.

### Steps to reproduce
1. Create two members.
2. Using a non graphic character (eg, 0x0F / u+000F) change the display name of one of the members to match the other member.

eg,
Member 1 display name: TestMember
Member 2 display name: Test`[non graphic character]`Member

Member 2's display name will show the same as member 1's display name.

### Version information
Version: Github latest

It appears that the symbol for a non graphic character is shown on the Account Settings page where you enter the display name but nowhere else.

### LINKED ISSUE #7100: Unicode should be normalized for storage efficiency, better searching, etc.
Right now, we don't normalize Unicode text in posts, user profile fields, etc., etc. Un-normalized Unicode can cause problems anywhere that we perform string comparison, including (but not limited to) searching, word censoring, and input validation. 

For example, suppose one user's browser submitted a post containing a normalized version of the string "Mäppchen", while another user's browser submitted a post containing an un-normalized version of the same string. The two different versions of the string will be saved as-is in their respective posts. Search results for "Mäppchen" will include one post or the other, but not both, depending on whether the submitted search term was normalized or not.

Besides solving possible problems regarding string comparison, performing consistent normalization on Unicode strings will also improve storage efficiency (assuming we normalize to Normalization Form C, which is the recommend form for general purpose use).

### LINKED ISSUE #7101: Better input sanitization needed in posts and profile fields, etc.
This issue is closely related to #7085, but extends it more generally.

SMF currently does not filter out invisible formatting characters, private use characters, or non-character byte sequences in user input. We do filter out ASCII control characters in various places, but that's not sufficient in a Unicode environment.

Filtering out ASCII control characters, private use characters, and non-characters is easy, but invisible formatting characters are a much trickier problem. On the one hand, if we don't filter out these invisible characters, it becomes trivial to insert them into strings in order to circumvent input filters. But on the other hand, if we simply ban all formatting characters in the name of security, we'll break support for emoji and for many non-European languages. Dealing with this problem requires a more nuanced approach than a simple ban, because formatting characters are legitimately necessary in some contexts, but have no business appearing in other contexts.

Sesquipedalian:
Fixes #7085 
Fixes #7100 
Fixes #7101 

This took a long time to finish, not because the code was particularly difficult, but because I had to spend a lot more time in the depths of the Unicode Standard's documentation, annexes, and technical reports than I ever expected. But enough of my whining and moaning....

The first thing this PR does is apply Unicode normalization to strings supplied via user input. A big chunk of that work is done by adding calls to `$smcFunc['normalize]()` in `preparsecode()`, `$smcFunc['htmlspecialchars']()`, `$smcFunc['strtoupper']()`, and `$smcFunc['strtolower']()`. Applying normalization to user inputs for the profile fields makes up the bulk of the remaining changes in this regard, as well as in a few odds and ends like `censorText()` and some admin input fields.

Testing the normalization aspect of this PR is pretty simple. Put some non-ASCII characters into a post, into a profile field (e.g. your displayed name or your signature), and into the description text of a membergroup or a calendar holiday title or whatever. If all goes well, you should see no visible difference whatsoever.

The second thing this PR does is improve input sanitization, particularly in user names and posts. To test this aspect, first try inserting a soft hyphen (U+00A0) or some other invisible formatting character into your displayed name in your user profile, or into the text of a post. If all goes well, the soft hyphen character will be replaced with a `�` (U+FFFD). Next, try inserting this string of text: `نامه‌ای`. If all goes well, the sanitization will leave the string unchanged. If it comes out looking like `⁧نامهای⁩` or `نامه�ای`, that's bad. Third, try inserting this string: `one ‮two‬ three`. In a post, it should appear unchanged, but in the displayed name, it should be changed to `one �two� three`.

### DIFF
--- a/Sources/ManageRegistration.php
+++ b/Sources/ManageRegistration.php
@@ -272,7 +272,7 @@ function EditAgreement()
  */
 function SetReserved()
 {
-	global $txt, $context, $modSettings;
+	global $txt, $context, $modSettings, $smcFunc;
 
 	// Submitting new reserved words.
 	if (!empty($_POST['save_reserved_names']))
@@ -319,7 +319,7 @@ function SetReserved()
  */
 function ModifyRegistrationSettings($return_config = false)
 {
-	global $txt, $context, $scripturl, $modSettings, $sourcedir;
+	global $txt, $context, $scripturl, $modSettings, $sourcedir, $smcFunc;
 	global $language, $boarddir;
 
 	// This is really quite wanting.
--- a/Sources/ManageRegistration.php
+++ b/Sources/ManageRegistration.php
@@ -105,7 +105,7 @@ function AdminRegister()
 
 		foreach ($_POST as $key => $value)
 			if (!is_array($_POST[$key]))
-				$_POST[$key] = htmltrim__recursive(str_replace(array("\n", "\r"), '', $_POST[$key]));
+				$_POST[$key] = htmltrim__recursive(str_replace(array("\n", "\r"), '', $smcFunc['normalize']($_POST[$key])));
 
 		$regOptions = array(
 			'interface' => 'admin',
@@ -224,6 +224,8 @@ function EditAgreement()
 		checkSession();
 		validateToken('admin-rega');
 
+		$_POST['agreement'] = $smcFunc['normalize']($_POST['agreement']);
+
 		// Off it goes to the agreement file.
 		$to_write = str_replace("\r", '', $_POST['agreement']);
 		$bytes = file_put_contents($boarddir . '/agreement' . $context['current_agreement'] . '.txt', $to_write, LOCK_EX);
@@ -277,6 +279,8 @@ function SetReserved()
 	{
 		checkSession();
 		validateToken('admin-regr');
+
+		$_POST['reserved'] = $smcFunc['normalize']($_POST['reserved']);
 
 		// Set all the options....
 		updateSettings(array(
@@ -359,7 +363,7 @@ function ModifyRegistrationSettings($return_config = false)
 			fatal_lang_error('admin_setting_coppa_require_contact');
 
 		// Post needs to take into account line breaks.
-		$_POST['coppaPost'] = str_replace("\n", '<br>', empty($_POST['coppaPost']) ? '' : $_POST['coppaPost']);
+		$_POST['coppaPost'] = str_replace("\n", '<br>', empty($_POST['coppaPost']) ? '' : $smcFunc['normalize']($_POST['coppaPost']));
 
 		call_integration_hook('integrate_save_registration_settings');
 
--- a/Sources/ManageServer.php
+++ b/Sources/ManageServer.php
@@ -198,6 +198,12 @@ function ModifyGeneralSettings($return_config = false)
 	{
 		call_integration_hook('integrate_save_general_settings');
 
+		foreach ($config_vars as $config_var)
+		{
+			if ($config_var[3] == 'text' && !empty($_POST[$config_var[0]]))
+				$_POST[$config_var[0]] = $smcFunc['normalize']($_POST[$config_var[0]]);
+		}
+
 		// Are we saving the stat collection?
 		if (!empty($_POST['enable_sm_stats']) && empty($modSettings['sm_stats_key']))
 		{
@@ -531,6 +537,8 @@ function ModifyCookieSettings($return_config = false)
 	if (isset($_REQUEST['save']))
 	{
 		call_integration_hook('integrate_save_cookie_settings');
+
+		$_POST['cookiename'] = $smcFunc['normalize']($_POST['cookiename']);
 
 		// Local and global do not play nicely together.
 		if (!empty($_POST['localCookies']) && empty($_POST['globalCookies']))
--- a/Sources/ManageServer.php
+++ b/Sources/ManageServer.php
@@ -150,7 +150,7 @@ function ModifySettings()
  */
 function ModifyGeneralSettings($return_config = false)
 {
-	global $scripturl, $context, $txt, $modSettings, $boardurl, $sourcedir;
+	global $scripturl, $context, $txt, $modSettings, $boardurl, $sourcedir, $smcFunc;
 
 	// If no cert, force_ssl must remain 0
 	require_once($sourcedir . '/Subs.php');
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -161,8 +161,10 @@ function reloadSettings()
 			$num = $string[0] === 'x' ? hexdec(substr($string, 1)) : (int) $string;
 			return $num < 0x20 || $num > 0x10FFFF || ($num >= 0xD800 && $num <= 0xDFFF) || $num === 0x202E || $num === 0x202D ? '' : '&#' . $num . ';';
 		},
-		'htmlspecialchars' => function($string, $quote_style = ENT_COMPAT, $charset = 'ISO-8859-1') use ($ent_check, $utf8, $fix_utf8mb4)
+		'htmlspecialchars' => function($string, $quote_style = ENT_COMPAT, $charset = 'ISO-8859-1') use ($ent_check, $utf8, $fix_utf8mb4, &$smcFunc)
 		{
+			$string = $smcFunc['normalize']($string);
+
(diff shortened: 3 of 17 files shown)


Sesquipedalian:
For those wondering, I have a separate PR forthcoming to deal with normalizing IRIs.

jdarwood007:
While I agree with the change, I am worried about it in relations to 2.1 final.  What is the risk impact?

sbulen:
This is high risk, of course.  And shouldn't be thrown in this close to 2.1 final.  

I feel the same way about this as #6786.

I'd even go so far as to say that any procedural => procedural rewrite in the 2.x branches is a waste of time.  

As an example of risks introduced, compare to #5815 that left 2.1 unusable for months. 

Not this close to final.  Write it clean in 3.0.

sbulen:
A smarter approach would be to eliminate the need for entity encoding altogether.  I.e., properly implement mb4.

Sesquipedalian:
You are mistaken, @sbulen. Unicode normalization is just about the safest operation one can perform on a string. The algorithms are idempotent and the only effect they have is to make sure all the characters in a string are consistently normalized to the same form. This is why [both the Unicode Consortium and the W3C recommend](https://unicode.org/reports/tr15/#Norm_Forms) that all strings in web content be normalized to Normalization Form C:

> The W3C Character Model for the World Wide Web 1.0: Normalization [CharNorm] and other W3C Specifications (such as XML 1.0 5th Edition) recommend using Normalization Form C for all content, because this form avoids potential interoperability problems arising from the use of canonically equivalent, yet different, character sequences in document formats on the Web. See the W3C Character Model for the Word Wide Web: String Matching and Searching [CharMatch] for more background.

Also, I am not sure why you mention entity handling. This PR has nothing to do with entities, just raw characters.

@jdarwood007, this will not have any significant effect on our progress toward 2.1.0. All this PR does is make sure that string input submitted during normal forum operation is normalized and sanitized better it was before. Existing content is unaffected.

jdarwood007:
I'm afraid I don't have enough expertise in the charsets/utf8 to give any more than my input then.  In my professional life, we just deal with Spanish, so its utf8 encoding on the pages and nvarchar in the database, the rest is handled by C# natively.  I don't need to do anything special to handle things with it.  I never dug into it deeply with SMF/PHP to get a high knowledge of it.

Hopefully @MissAllSunday, @BrickOzp and @live627 can chime in as well.

Sesquipedalian:
The basic idea of Unicode normalization is actually quite simple, @jdarwood007.

> Essentially, the Unicode Normalization Algorithm puts all combining marks in a specified order, and uses rules for decomposition and composition to transform each string into one of the Unicode Normalization Forms.

Composition and decomposition just mean turning, e.g., `a` + `¨` into `ä` or vice versa. It's really quite straightforward. 🙂

sbulen:
The concept is of course a good idea.  The problems are risk and timing.  

Testing all those string variants in different languages is a big deal.  There's a lot of code in there.

A code-your-own solution for such a complex topic is the wrong solution.  Especially this close to 2.1.

If we must implement unicode normalization at this time, we should use the php functions.
https://www.php.net/manual/en/class.normalizer.php

Sources/ManageRegistration.php
```diff
@@ -359,7 +363,7 @@ function ModifyRegistrationSettings($return_config = false)
 			fatal_lang_error('admin_setting_coppa_require_contact');
 
 		// Post needs to take into account line breaks.
-		$_POST['coppaPost'] = str_replace("\n", '<br>', empty($_POST['coppaPost']) ? '' : $_POST['coppaPost']);
+		$_POST['coppaPost'] = str_replace("\n", '<br>', empty($_POST['coppaPost']) ? '' : $smcFunc['normalize']($_POST['coppaPost']));
```
live627, line 366: Must declare the global `$smcFunc`.

live627, line 366: Must declare the global `$smcFunc`.

Sources/ManageRegistration.php
```diff
@@ -278,6 +280,8 @@ function SetReserved()
 		checkSession();
 		validateToken('admin-regr');
 
+		$_POST['reserved'] = $smcFunc['normalize']($_POST['reserved']);
```
live627, line 283: Must declare the global `$smcFunc`.

Sources/ManageServer.php
```diff
@@ -198,6 +198,12 @@ function ModifyGeneralSettings($return_config = false)
 	{
 		call_integration_hook('integrate_save_general_settings');
 
+		foreach ($config_vars as $config_var)
+		{
+			if ($config_var[3] == 'text' && !empty($_POST[$config_var[0]]))
+				$_POST[$config_var[0]] = $smcFunc['normalize']($_POST[$config_var[0]]);
```
live627, line 204: Must declare the global `$smcFunc`.

live627 (commented):
Courtesy of Scrutiniser.

live627:
> If we must implement unicode normalization at this time, we should use the php functions.

Did you miss where the polyfills were added
- https://github.com/SimpleMachines/SMF2.1/pull/7032

sbulen:
First quick test - it seems to be mangling some languages.

`Sanskrit: ﻿काचं शक्नोम्यत्तुम् । नोपहिनस्ति माम् ॥ `

Appears like this with this PR:
`Sanskrit: �काचं शक्नोम्यत्तुम् । नोपहिनस्ति माम् ॥`

And:
`Nepali: ﻿म काँच खान सक्छू र मलाई केहि नी हुन्‍न् ।`

Appears like this with this PR:
`Nepali: �म काँच खान सक्छू र मलाई केहि नी हुन्‍न् ।`

sbulen:
> 
> Did you miss where the polyfills were added

Yes.

Sesquipedalian:
> First quick test - it seems to be mangling some languages.
> 
> `Sanskrit: ﻿काचं शक्नोम्यत्तुम् । नोपहिनस्ति माम् ॥ `
> 
> Appears like this with this PR: `Sanskrit: �काचं शक्नोम्यत्तुम् । नोपहिनस्ति माम् ॥`
> 
> And: `Nepali: ﻿म काँच खान सक्छू र मलाई केहि नी हुन्‍न् ।`
> 
> Appears like this with this PR: `Nepali: �म काँच खान सक्छू र मलाई केहि नी हुन्‍न् ।`

That is behaving as expected. In both cases your strings contained a Byte Order Mark, a.k.a. Zero Width No-Break Space (U+FEFF), which is both useless and disallowed. Remove that character from you string and you will find that it is still laid out exactly as intended.

Sources/ManageRegistration.php
Sesquipedalian: Thanks.

sbulen:
> That is behaving as expected. In both cases your strings contained a Byte Order Mark, a.k.a. Zero Width No-Break Space (U+FEFF), which is both useless and disallowed. Remove that character from you string and you will find that it is still laid out exactly as intended.

You are correct - I missed the zero-width-no-break on both of those.

This PR really should use the php-supplied functions instead of building our own.  This is still too high-risk a change.

Sesquipedalian:
If you look at the utf8_normalize_*() functions in Subs-Charset.php, you'll see that they do use normalizer_normalize() if it is available. But since many PHP installs do not have that function available, we also have a complete polyfill for it. Moreover, if you look at #7032, you will see that I provided a unit test with it that demonstrates that this polyfill successfully runs all 18000+ normalization tests provided by the Unicode Consortium.

Sources/ManageServer.php
jdarwood007: @Sesquipedalian See #7127

Sesquipedalian: Thanks, @jdarwood007. Will submit fix shortly.

PR #7105, opened 2021-10-04
Fix "copy" theme missing language file

DiegoAndresCortes:
- Create languages dir with the Settings.english.php file
- Remove unnecessary files that might confuse a new theme author:
  - _calendar.css_
  - _calendar.rtl.css_
  - _slider.min.css_
These files are not usually modified and can just override their elements from index.css or similar. Can also manually include them in their themes and directly modify it if they require it.

### DIFF
--- a/Sources/Themes.php
+++ b/Sources/Themes.php
@@ -1343,8 +1343,19 @@ function InstallCopy()
 	mkdir($context['to_install']['theme_dir'] . '/css', 0777);
 	mkdir($context['to_install']['theme_dir'] . '/scripts', 0777);
 
+	// Create subdirectory for language files
+	mkdir($context['to_install']['theme_dir'] . '/languages', 0777);
+
 	// Copy over the default non-theme files.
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Leaving out slider.min.css makes good sense to me, but I am not convinced about calendar.css and calendar.rtl.css. Those files, regardless of whether they are being loaded from the current theme or the default theme, are always loaded after index.css, and therefore always with higher precedence. This means it would be easy to have a situation where the default theme's calendar.css would override the custom theme's index.css. If the theme author is not intimately familiar with SMF's theme system and fallback behaviours, this could be very confusing. For better discovery, therefore, I think we should still duplicate calendar.css and calendar.rtl.css.

Otherwise, this looks good.

Sesquipedalian (changes requested):
Please add calendar.css and calendar.rtl.css back to the list of files to copy. See previous comment for explanation.

Sesquipedalian (approved):

PR #7106, opened 2021-10-05
Improve the hooks list

### LINKED ISSUE #7097: Very long page loading-time (Integration hooks)
#### Description
`Fatal error: Maximum execution time of 60 seconds exceeded in ***\Sources\ManageMaintenance.php on line 2084`

### Steps to reproduce
1. Install several mods that use many hooks (~200+)
2. Go to Maintenance => Forum Maintenance => Integration hooks page in Admin

- Version/Git revision: 2.1 RC4
- Database Type: MariaDB
- Database Version:  10.5.10
- PHP Version: 7.4.23

### Additional information
Of course, this can be fixed, e. g. through `ini_set('max_execution_time', 120)` in the `get_integration_hooks_data` function, but maybe it is better to optimize the function that retrieves information about all the hooks?

live627:
- Greatly reduces the number of calls to `file_exists()` because it's kinda slow on Windows
- Fix sorting
- Functions with the same name but have different namespaces, classes, are now resolved correctly

The previous name resolution algorithm saws tightly wound in several loops (`num_files * nun_hooks *, num_functions`) and therefore could note scale at all. This is now changed so that all filesystem operations are outside of the main loops.

Filesystem operations on Windows can be a bit slow,    (NTFS) even with a SSD. So build a list of files from `$sourcedir` and use that to check for file existence before resorting to reading the disk.

### DIFF
--- a/Sources/ManageMaintenance.php
+++ b/Sources/ManageMaintenance.php
@@ -1852,16 +1852,37 @@ function MaintainRecountPosts()
  */
 function list_integration_hooks()
 {
-	global $sourcedir, $scripturl, $context, $txt;
+	global $boarddir, $sourcedir, $scripturl, $context, $txt;
 
-	$context['filter_url'] = '';
-	$context['current_filter'] = '';
-	$currentHooks = get_integration_hooks();
-	if (isset($_GET['filter']) && in_array($_GET['filter'], array_keys($currentHooks)))
+	$filter_url = '';
+	$current_filter = '';
+	$hooks = get_integration_hooks();
+	$hooks_filters = array();
+
+	if (isset($_GET['filter'], $hooks[$_GET['filter']]))
 	{
-		$context['filter_url'] = ';filter=' . $_GET['filter'];
-		$context['current_filter'] = $_GET['filter'];
+		$filter_url = ';filter=' . $_GET['filter'];
+		$current_filter = $_GET['filter'];
 	}
+	$filtered_hooks = array_filter(
+		$hooks,
+		function($hook) use ($current_filter)
+		{
+			return $current_filter == '' || $current_filter == $hook;
+		},
+		ARRAY_FILTER_USE_KEY
+	);
+	ksort($hooks);
+
+	foreach ($hooks as $hook => $functions)
+		$hooks_filters[] = '<option' . ($current_filter == $hook ? ' selected ' : '') . ' value="' . $hook . '">' . $hook . '</option>';
+
+	if (!empty($hooks_filters))
+		$context['insert_after_template'] .= '
+		<script>
+			var hook_name_header = document.getElementById(\'header_list_integration_hooks_hook_name\');
+			hook_name_header.innerHTML += ' . JavaScriptEscape('<select style="margin-left:15px;" onchange="window.location=(\'' . $scripturl . '?action=admin;area=maintain;sa=hooks\' + (this.value ? \';filter=\' + this.value : \'\'));"><option value="">' . $txt['hooks_reset_filter'] . '</option>' . implode('', $hooks_filters) . '</select>') . ';
+		</script>';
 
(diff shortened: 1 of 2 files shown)


frandominguezl:
Thanks for the explanation! Quite interesting.

Sesquipedalian:
When no hooks have been added to the system, viewing the integration hooks admin page generates errors at the top of the page:
```
Notice
: Undefined index: function in
/Users/jon/Sites/dev/Sources/Subs-List.php
on line
90


Warning
: call_user_func_array() expects parameter 1 to be a valid callback, no array or string given in
/Users/jon/Sites/dev/Sources/Subs-List.php
on line
91
```

Sources/ManageMaintenance.php
```diff
@@ -2015,29 +2051,25 @@ function list_integration_hooks()
...
+	return iterator_to_array(
+		new RecursiveIteratorIterator(
+			new RecursiveCallbackFilterIterator(
+				new RecursiveDirectoryIterator($dirname, FilesystemIterator::UNIX_PATHS),
+				function ($fileInfo, $currentFile, $iterator)
+				{
+					// Allow recursion
+					if ($iterator->hasChildren())
+						return true;
+					return $fileInfo->getExtension() == 'php';
```
Sesquipedalian, line 2068: The existing version of this function returns all files, not just PHP files. I do like the idea of filtering the list on the fly like this, but it would be better not to use a hard-coded extension. I suggest adding a second parameter to the function that takes an array of file extensions to filter by.

Sources/ManageMaintenance.php
```diff
@@ -2291,14 +2212,38 @@ function get_hook_info_from_raw($rawData)
...
+		$hookData['call'] = $hookData['pureFunc'] = $modFunc;
 
 	return $hookData;
 }
 
+function get_defined_functions_in_file(string $file): array
+{
+	$source = file_get_contents($file);
+	// token_get_all() is too slow so use a nice little regex instead.
+	preg_match_all('/^namespace\s([a-z\\\]++)\s*;|class\s+(\w++)[\w\s]*{|function\s+(\w++)\s*\(.*\)[:\|\w\s]*{/im', $source, $matches, PREG_SET_ORDER);
```
Sesquipedalian:
```suggestion
	preg_match_all('/\bnamespace\s++((?P>label)(?:\\\(?P>label))*+)\s*+;|\bclass\s++((?P>label))[\w\s]*+{|\bfunction\s++((?P>label))\s*+\(.*\)[:\|\w\s]*+{(?(DEFINE)(?<label>[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*+))/i', $source, $matches, PREG_SET_ORDER);
```
1. Uses a `(?P>label)` subroutine, which is defined as `(?<label>[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*+)` inside the the `(?(DEFINE)...)` block, to more precisely match valid identifier labels for functions/classes/namespaces. The `[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*` definition comes from the PHP documentation [here](https://www.php.net/manual/en/language.variables.basics.php).
2. Uses `\b` before `namespace`, `class`, and `function` to better protect against, e.g. `not_a_function`. This also allows removing the `^` and makes multiline mode unnecessary.

Sesquipedalian (changes requested):
This is not yet a complete review, but here are a couple of things I noticed.

Sources/ManageMaintenance.php
live627: And this function was consumed by `get_integration_hooks_data()` which checked for that file  extension anyway.

Sesquipedalian (approved):

Sources/ManageMaintenance.php
Sesquipedalian: Fair enough.

PR #7124, opened 2021-10-13
Implements better IRI handling

### LINKED ISSUE #7084: Better handling of IRIs is needed in various places
Various parts of SMF's code need to do a better job handling IRIs.
- Although the autolinker now correctly converts IRIs to pure ASCII URLs in `href` attributes, that doesn't currently happen in BBCodes.
- IRIs entered in the user profile field `website_url` need to be normalized, in order to avoid problems further down the line.
- IRIs entered as the targets of redirection boards need to be normalized for similar reasons.
- If `$boardurl` is an IRI, it needs to be normalized, too. Otherwise, it is possible for the admin to enter an IRI that is perfectly correct to the human eye but could nevertheless immediately render the forum unreachable.
- IRIs in `globalCookiesDomain` and `cors_domains` need to be normalized. **Failing to do so has potential security implications.**
- Likewise, incoming checks for cookies and cross-origin requests need to be make IRI-aware. **Failing to do so has potential security implications.**

Sesquipedalian:
- Updates our polyfills with data from the recently released Unicode 14.0, moves that data from Subs-Charset.php into separate files, and adds an internal development file to make it trivial to update that data again whenever necessary in the future.

- Adds a normalize_iri() function. 
    - When an IRI is fed to this function, the returned value is still an IRI, but sanitized and with all Unicode characters normalized according to the requirements of RFC 3491 for the domain name and according to RFC 3987 for the rest.
    - normalize_iri() internally calls sanitize_iri(), which means that in most cases normalize_iri() can supersede sanitize_iri() (see below). However, there are situations where one would want to call sanitize_iri() directly, so it has been retained as a separate function.

- Adds a parse_iri() wrapper for parse_url(), since the latter doesn't support Unicode very well and will return malformed results for IRIs such as `http://кто.рф/images/search.gif`.

- Fixes #7084
    - Improves handling of IRIs in parse_bbc()
        - In the url, iurl, img, ftp, and flash BBCodes, the value of the generated href or src attribute is always a pure ASCII URL, not an IRI.
        - The autolinker now calls normalize_iri() and iri_to_url() instead of running the bespoke code added in a605ee9

    - Normalizes IRIs in the website_url user profile field.
        - This simply involved calling normalize_iri() in place of sanitize_iri() in Register.php and Profile-Modify.php.

    - Normalizes IRIs when creating redirection boards.
        - This simply involved adding a call to normalize_iri() in ManageBoards.php.

    - Normalizes $boardurl when set or updated in the installer or admin control panel.
        - This simply involved adding calls to normalize_iri() in ManageServer.php and install.php.

    - Normalizes & better validates $modSettings['globalCookiesDomain'] & $modSettings['cors_domains'].
        - For $modSettings['globalCookiesDomain'], this not only makes sure the domain name is correctly normalized, but also improves our overall validation and sanitization of the value. Previously it was quite easy for the admin to set the global cookie domain to `forum.example` if $boardurl was `https://forum.example.com`.
        - For $modSettings['cors_domains'], this applies normalization and sanitization, and also ensures that any wildcards are only present at the start of the domain name. For example, `*.example.com` and `https://*.example.com` will be accepted, but `foo.*.com` and `https://foo.*.com` will not.

    - When checking incoming cookies and cross-origin requests, the checks are now IRI-aware.
        - This closes #6891 (see https://github.com/SimpleMachines/SMF2.1/pull/6891#issuecomment-920489007).

- Fixes an unreported bug in url_to_iri(), wherein urlencoded syntax characters would be incorrectly decoded to the raw characters.

- Fixes an unreported bug wherein the idn_to_ascii() polyfill could produce incorrect results for international domain names using certain non-European scripts.
    - Previously, we just applied mb_strtolower() to IDNs before feeding them to the Punycode algorithm. However, RFC 3491 says that characters should be case folded, not lowercased. Case folding is the process of eliminating case differences by converting cased characters to a default case. In the case of European scripts (and indeed most scripts), case folding does in fact produce lowercase characters, but there are some non-European scripts in which case folding produces uppercase characters instead. IDNs written in such scripts are relatively uncommon today, but that will change in the future, so it is important to implement proper case folding in our idn_to_ascii() polyfill.


Sesquipedalian:
@jdarwood007, if you could check the code related to CORS and global cookies, that'd be great.

live627:
can you check the issues found by Scrutiniser? Some of them look like they may need fixing

https://scrutinizer-ci.com/g/SimpleMachines/SMF2.1/inspections/d8f056d1-375b-4174-83de-fb76c062082e/issues/

Sesquipedalian:
> can you check the issues found by Scrutiniser? Some of them look like they may need fixing
> 
> https://scrutinizer-ci.com/g/SimpleMachines/SMF2.1/inspections/d8f056d1-375b-4174-83de-fb76c062082e/issues/

Thanks. I've now fixed the ones that are legitimate.

Sesquipedalian:
Pinging @jdarwood007 again. I really just need you to verify that the CORS related code works as expected. (Note: In my tests it works as *I* expect, but it is possible that *you* expected and/or intended something different, and you are the authority on CORS here.)

jdarwood007:
I forgot about this when I lost the unread notification in my inbox.

This PR works fine on my test site.

jdarwood007 (approved):

PR #7129, opened 2021-10-20
Fix error with allow_xhjr_credentials not being defined in 3rd party scripting.

jdarwood007:
Ensure if allow_xhjr_credentials is not defined we default too false, which should prevent scripting errors on 3rd party APIs that are using SMF components.

We don't need to merge this until after 2.1.0, but it would be ideal.


Sesquipedalian (approved):

Sesquipedalian:
Seems straightforward enough. We may as well add it now.

PR #7134, opened 2021-10-30
Automatically builds character classes for utf8_sanitize_invisibles()

### LINKED ISSUE #7131: Emoji support in post
#### Description
If post text contains emoji with Variation Selector-16 (for example ❗️ U+2757U+FE0F) it has error ''The message body was left empty.'' on posting or preview
In Admin-Error Log error
`2: mb_ereg_replace_callback(): mbregex compile err: invalid character property name {Regional_Indicator}`
and error stack
`#1: mb_ereg_replace_callback()
Call from /..../Sources/Subs-Charset.php in line 3297`
### Steps to reproduce
1. Copy emoji ❗️
2. Paste in post body
3. Click "Preview"

- Version/Git revision: 2.1RC4 Git https://github.com/SimpleMachines/SMF2.1/commit/aa18d874d3778566c47db02ea5f2ae3f05b10aa1

Sesquipedalian:
- Fixes #7131 by compiling the necessary regular expression classes directly from the data in the Unicode Character Database (rather than relying on whatever regex library version happens to be installed on a server).
- Removes dependency on mb_ereg* functions, allowing us to use just the preg* functions.
- Fixes a typo that caused an unreported bug when sanitizing the Zero Width Non-Joiner in cursive scripts like Arabic.


other/update_unicode_data.php
```diff
@@ -211,37 +477,336 @@
...
+			$funcs['utf8_regex_indic']['data'][$char_script][$insc] = array();
+
+		$funcs['utf8_regex_indic']['data'][$char_script][$insc][] = '\\x{' . str_replace('..', '}-\\x{', $fields[0]) . '}';
+	}
+}
+// Again, sort commonly used scripts before rare scripts.
+uasort($funcs['utf8_regex_indic']['data'], function($a, $b) {
+	if ($a['stats']['age'] == $b['stats']['age']) {
+		return $b['stats']['count'] - $a['stats']['count'];
+	} else {
```
live627: ahem...er... this isn't Allman

Sesquipedalian: I don't understand this comment, @live627.

Sesquipedalian: Oh, you just mean the indentation style. It is true, I just used my text editor's automatic PSR-2 styling when I first made this file for private use, and then I never bothered to change it. I suppose I might as well...

Sesquipedalian: There, done.

PR #7137, opened 2021-11-01
who_admin update

Dzonny:
Fixing the who_admin string.


Sesquipedalian (changes requested):
1. We might as well correct the next line, too.
2. Please remove the trailing newline.

PR #7143, opened 2021-11-07
Improves input sanitization in utf8_* functions

Sesquipedalian:


jdarwood007:
Thanks, I haven't had time to check, but this looks like it should do it based off the logic alone.

PR #7144, opened 2021-11-08
Fix handling of external avatars with parse_iri (fixes #7139)

### LINKED ISSUE #7139: Issue with avatars from integrated [Oxygen] gallery
#### Description
Hello,
Just upgraded two SMF 2.1 forums with latest build ed0f841
Users with avatars selected from integrated [Oxygen] gallery are displayed with default avatar only.

### Steps to reproduce
1.  Upgrade forum with latest build ed0f841
2.  Open some user profile and select an avatar from integrated [Oxygen] gallery
3.  Open user's post and topics and see avatar displayed

- Version/Git revision: ed0f841
- Database Type: MariaDB
- Database Version: 10.3.31
- PHP Version: 8.0.11

Oldiesmann:
The code expected a return of "null" from parse_iri (which the old parse_url call would do if there was no scheme in the given string), but parse_iri will return a string if there's no scheme rather than null Since we know it will be an array when there's a scheme, we can just check whether the function returned an array. Fixes #7139 

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -3883,7 +3883,10 @@ function set_avatar_data($data = array())
 
 			// External url.
 			else
-				$image = parse_iri($data['avatar'], PHP_URL_SCHEME) !== null ? get_proxied_url($data['avatar']) : $modSettings['avatar_url'] . '/' . $data['avatar'];
+			{
+				$parsed_image = parse_iri($data['avatar']. PHP_URL_SCHEME);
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
I'd rather fix `parse_iri()` to make sure its return values exactly correspond to the return values from `parse_url()`. If there's a situation where `parse_url()` returns null but `parse_iri()` doesn't, that's a bug in my code and I need to fix it. 🙂

PR #7151, opened 2021-11-10
Fix comment in emails about reported messages

### LINKED ISSUE #7149: (SMorg) report mail: reporter comment shows message content instead
#### Description
When receiving "Reported post: ${title} by ${author}" mails from SMorg, the text after "The reporter has made the following comment:" actually contains the original text by the author, not the comment by the reporter.
This started around 2021-08-26; on 2021-08-12 it still worked (I did not receive reports between those dates, IIRC notifications were broken/reset for some time and I had to manually set them again).
Probably a 2.1 upgrade issue, and might very well be that I was the one who screwed up the `sprintf` string.

Anyone else seeing this or am I going crazy? :wink: 

### Steps to reproduce
1. Be a moderator
2. Receive a "Reported post:" mail

(Whatever SMorg is running)

BrickOzp:
The comment in the email was not the comment left
by the reporter, but the actual message.

Fixes #7149

### DIFF
--- a/Sources/tasks/MsgReport-Notify.php
+++ b/Sources/tasks/MsgReport-Notify.php
@@ -155,11 +155,14 @@ class MsgReport_Notify_Background extends SMF_BackgroundTask
 			// Second, get some details that might be nice for the report email.
 			// We don't bother cluttering up the tasks data for this, when it's really no bother to fetch it.
 			$request = $smcFunc['db_query']('', '
-				SELECT lr.subject, lr.membername, lr.body
+				SELECT lr.subject, lr.membername, lrc.comment
 				FROM {db_prefix}log_reported AS lr
-				WHERE id_report = {int:report}',
+					INNER JOIN {db_prefix}log_reported_comments AS lrc ON (lr.id_report = lrc.id_report)
+				WHERE lr.id_report = {int:report}
+					AND lrc.id_comment = {int:comment}',
 				array(
 					'report' => $this->_details['report_id'],
+					'comment' => $this->_details['comment_id'],
 				)
 			);
 			list ($subject, $poster_name, $comment) = $smcFunc['db_fetch_row']($request);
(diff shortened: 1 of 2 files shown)


BrickOzp:
The report comments are stored in the log_reported_comments table.
I solved this by including the comment in the background task data.
If the data field is to small for this we could also solve this by including the comment id in the data
and query the log_reported_comments table in the task.

Sesquipedalian:
> If the data field is to small for this

Not a problem, I think. The `task_data` field is a MEDIUMTEXT, so there is plenty of room.

Sources/tasks/MsgReport-Notify.php
```diff
@@ -174,7 +174,7 @@ public function execute()
 					'REPORTERNAME' => $this->_details['sender_name'],
 					'TOPICLINK' => $scripturl . '?topic=' . $this->_details['topic_id'] . '.msg' . $this->_details['msg_id'] . '#msg' . $this->_details['msg_id'],
 					'REPORTLINK' => $scripturl . '?action=moderate;area=reportedposts;sa=details;rid=' . $this->_details['report_id'],
-					'COMMENT' => $comment,
```
Sesquipedalian: Since this is the only place that `$comment` is used, we may as well remove `lr.body` and `$comment` from the query above.

Sesquipedalian (changes requested):

Sesquipedalian:
~Actually, on further consideration, I think it might be better to change the preceding query to something like this:~
```
	SELECT lr.subject, lr.membername, lrc.comment
	FROM {db_prefix}log_reported AS lr
		INNER JOIN {db_prefix}log_reported_comments AS lrc ON (lr.id_report = lrc.id_report)
	WHERE lr.id_report = {int:report}',
	array(
		'report' => $this->_details['report_id'],
	)
);
list ($subject, $poster_name, $comment) = $smcFunc['db_fetch_row']($request);
```

EDIT: Ignore this suggestion. There's a one to many relationship between reports and comments, so this suggestion wouldn't work. Please just implement my originally requested changes in my initial review above.

BrickOzp:
If we pass the id_comment in the task we could do that query

Sesquipedalian:
That sounds like the perfect plan.

Sesquipedalian (approved):

PR #7152, opened 2021-11-12
Cors was non functional

jdarwood007:
Fix an issue where cors wasn't working because valid_cors_found was not set to a valid value.

@Sesquipedalian Not sure what happened in my testing of your PR.  It worked initially.  But this squares away issues we are having with SM Website.  I also forced it to always set a value (using true)

### DIFF
--- a/Sources/Security.php
+++ b/Sources/Security.php
@@ -1475,7 +1475,7 @@ function corsPolicyHeader($set_header = true)
 			if (preg_match('~' . $host_regex . '~u', $origin['host']))
 			{
 				$context['cors_domain'] = trim($_SERVER['HTTP_ORIGIN']);
-				$context['valid_cors_found'] = $origin['type'];
+				$context['valid_cors_found'] = !empty($allowed_origin['type']) ? $allowed_origin['type'] : true;
 				break;
 			}
 		}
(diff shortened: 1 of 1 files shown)


Sources/Security.php
```diff
@@ -1475,7 +1475,7 @@ function corsPolicyHeader($set_header = true)
 			if (preg_match('~' . $host_regex . '~u', $origin['host']))
 			{
 				$context['cors_domain'] = trim($_SERVER['HTTP_ORIGIN']);
-				$context['valid_cors_found'] = $origin['type'];
+				$context['valid_cors_found'] = !empty($allowed_origin['type']) ? $allowed_origin['type'] : true;
```
Sesquipedalian: Yes, this was supposed to be `$allowed_origin['type']` rather than `$origin['type']`.

However, I don't think the ternary is necessary. There's no way that `$allowed_origin['type']` can ever be empty here.

Sesquipedalian (changes requested):

Sources/Security.php
jdarwood007: Well was just making sure it would always validate.  Because valid_cors_found was null, the headers where not getting set.

Sesquipedalian (approved):

PR #7164, opened 2021-11-24
Handle hooks pointing outside the sources directory

live627:
The integration management area should recognise the existence of hooks that specify a file that is outside of the sources directory.

### DIFF
--- a/Sources/ManageMaintenance.php
+++ b/Sources/ManageMaintenance.php
@@ -2104,7 +2104,12 @@ function get_integration_hooks_data($start, $per_page, $sort, $filtered_hooks, $
 	foreach ($filtered_hooks as $hook => $functions)
 		foreach ($functions as $rawFunc)
 		{
-			$hookParsedData = parse_integration_hook($hook, $rawFunc, $files);
+			$hookParsedData = parse_integration_hook($hook, $rawFunc);
+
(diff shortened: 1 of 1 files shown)


Sources/ManageMaintenance.php
```diff
@@ -2104,7 +2104,12 @@ function get_integration_hooks_data($start, $per_page, $sort, $filtered_hooks, $
 	foreach ($filtered_hooks as $hook => $functions)
 		foreach ($functions as $rawFunc)
 		{
-			$hookParsedData = parse_integration_hook($hook, $rawFunc, $files);
+			$hookParsedData = parse_integration_hook($hook, $rawFunc);
+
+			// Handle hooks pointing outside the sources directory.
+			if ($hookParsedData['absPath'] != '' && (!isset($files[$hookParsedData['absPath']]) || file_exists($hookParsedData['absPath'])))
```
Sesquipedalian:
```suggestion
			if ($hookParsedData['absPath'] != '' && !isset($files[$hookParsedData['absPath']]) && file_exists($hookParsedData['absPath']))
```
 This should avoid running files through `get_defined_functions_in_file()` again if we already did them above.

Sesquipedalian (changes requested):

Sesquipedalian (approved):

PR #7165, opened 2021-11-25
Rewrite board permission reports to use less memory

### LINKED ISSUE #7142: Memory exhaustion running board permissions with lots of groups/permissions
#### Description
[Fri Nov 05 20:17:00.814668 2021] [php7:error] [pid 30976] [client xxxx:0] PHP Fatal error:  Allowed memory size of 268435456 bytes exhausted (tried to allocate 20480 bytes) in /Sources/Reports.php on line 976, referer: xxxx/index.php?action=admin;area=reports

### Steps to reproduce
In general
- About 150 boards
- About 150 groups
- About 50 General Permissions (Some groups are using inheritance) 
- 30 Permission Profiles
- Using Post moderation

1. Go to Admin center > Reports
2. Choose Board Permissions
3. Run

- Version/Git revision: b11f632dd
- Database Type: Mysql
- Database Version:
- PHP Version:

live627:
- A side effect is more speed which is expected because memory allocations are slow.
- This retains the same behavior which duplicates permission profiles for the sake of showing one table per board. Should these be grouped?
- The template with the file pointer uses 20 fewer MB than the one with the echo statements—74 MB vs 92 MB—in my test (previous code swallowed over 500 MB)
- Fixes #7142

### DIFF
--- a/Sources/Reports.php
+++ b/Sources/Reports.php
@@ -288,6 +288,8 @@ function BoardReport()
  */
 function BoardPermissionsReport()
 {
+	global $context, $txt;
+
 	global $txt, $modSettings, $smcFunc;
 
 	// Get as much memory as possible as this can be big.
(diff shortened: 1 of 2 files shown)


Sources/Reports.php
```diff
@@ -288,6 +288,8 @@ function BoardReport()
  */
 function BoardPermissionsReport()
 {
+	global $context, $txt;
```
Oldiesmann, line 291: There's already a global statement in this function, which includes $txt. No need to add another one just for one variable.

live627: I'll add a third one just to make you happy.

Sources/Reports.php
```diff
@@ -319,38 +321,24 @@ function BoardPermissionsReport()
 
 	// Fetch all the board names.
 	$request = $smcFunc['db_query']('', '
-		SELECT id_board, name, id_profile
-		FROM {db_prefix}boards
+		SELECT id_board, b.name, id_profile
+		FROM {db_prefix}categories AS c
+			JOIN {db_prefix}boards AS b USING (id_cat)
```
albertlast, line 326: wouldn't recommand to use this (USING) short hand, just keep by smf default ON statment logic.

Sources/Reports.php
```diff
@@ -319,38 +321,24 @@ function BoardPermissionsReport()
 
 	// Fetch all the board names.
 	$request = $smcFunc['db_query']('', '
-		SELECT id_board, name, id_profile
-		FROM {db_prefix}boards
+		SELECT id_board, b.name, id_profile
```
albertlast, line 324: since using join, would be good to use on every column the table alias,
otherwise it's not clear from where the column is coming

Sources/Reports.php
```diff
@@ -319,38 +321,24 @@ function BoardPermissionsReport()
...
+		FROM {db_prefix}categories AS c
+			JOIN {db_prefix}boards AS b USING (id_cat)
 		WHERE ' . $board_clause . '
-		ORDER BY id_board',
+		ORDER BY cat_order, child_level, board_order',
```
albertlast, line 328: same here would add alias,
for better understand from where the column is coming

albertlast (commented):
did only check the sql stuff

jdarwood007:
Updated code: 425k, 2.5 s to get that and 170,000 ms to load the page.
fwrites code: 394kb, 9s to get all that. 230,000 ms to load the page.

The updated code I think is the best, the fwrites had my browser running some cpu for a bit.  not sure why.

Sesquipedalian:
Any updates on this, @live627? I would like to get this fixed as soon as we can.

PR #7166, opened 2021-11-27
Has not worked; safer to remove than to fix

sbulen:
We should leave the DB definition as intended.

Safer to remove this than to fix the syntax errors in it...


live627:
Reviewing this code again, I think it figured out what it is supposed to do: it finds the column type and updates the query if a different type was detected (different width, etc.).

sbulen:
Yep.  But only for 2 columns, salt & membergroups.  

I know there were issues with salt in the past, where it often wasn't big enough.  

But it's plenty big now.  We should run with the current definition.

Sesquipedalian (approved):

PR #7177, opened 2021-12-07
Fix calls to db_affected_rows

### LINKED ISSUE #7176: mysqli_affected_rows() expects parameter 1 to be mysqli, boolean given
#### Description
Error found in error log

mysqli_affected_rows() expects parameter 1 to be mysqli, boolean given

#0: smf_error_handler()
#1: mysqli_affected_rows()
Called from /Sources/Subs-Db-mysql.php on line 513
#2: smf_db_affected_rows()
Called from /Sources/Session.php on line 242
#3: sessionGC()
#4: session_start()
Called from /Sources/Session.php on line 81
#5: loadSession()
Called from /index.php on line 184

### Steps to reproduce
1. Don't know
2. 

- Version/Git revision: 12bf435a8
- Database Type: MySql
- Database Version:
- PHP Version: 7.1

BrickOzp:
db_affected_rows expects a connection object.
Remove parameters of type query result.

Fixes #7176

### DIFF
--- a/Sources/Session.php
+++ b/Sources/Session.php
@@ -187,7 +187,7 @@ function sessionWrite($session_id, $data)
 		array('session_id')
 	);
 
-	return ($smcFunc['db_affected_rows']($session_update) == 0 ? false : true);
+	return ($smcFunc['db_affected_rows']() == 0 ? false : true);
 }
 
 /**
@@ -239,7 +239,7 @@ function sessionGC($max_lifetime)
 		)
 	);
 
-	return ($smcFunc['db_affected_rows']($session_update) == 0 ? false : true);
+	return ($smcFunc['db_affected_rows']() == 0 ? false : true);
 }
 
 ?>


Sources/Session.php
```diff
@@ -239,7 +239,7 @@ function sessionGC($max_lifetime)
 		)
 	);
 
-	return ($smcFunc['db_affected_rows']($session_update) == 0 ? false : true);
+	return ($smcFunc['db_affected_rows']() == 0 ? false : true);
```
albertlast, line 242: what is wrong here?

BrickOzp: db_affected_rows/mysqli_affected_rows expects a connection object, but we pass the response of db_query

albertlast: the deeper issue look like is, that db_affected_rows works in pg and mysql differently, since pg needs the result resource

sbulen: It looks like php 8.1 changed the pg interface for several calls...

@albertlast is correct for php 8.1.  Prior to php 8.1, pg wanted the connect.

But they also changed the corresponding queries to return what was expected....  They tried to make the change 'opaque'...  (Other than your variable names being all wrong...)

More here:
https://php.watch/versions/8.1/PgSQL-resource

So...  

I believe this proposed fix works thru php 8.0 for both pg and mysqli.

It *might* even work in php 8.1, if the 8.1 changes are as opaque as they claim.

Bottom line is testing is needed...

tinoest: I don’t think this is the case, the old PG used a result resource now it uses a result object. 

So the two (MySQL and PgSQL) have always been different.

jdarwood007: @BrickOzp , My PR recently changed this.  I was getting other session issues and I thought this fixed it.  My thought was to ensure we worked with the proper database that had the session data.

I say for now, the revert is ok.  In the future (beyond 2.1), we should move the session handler to a object and store the $db_connection (from when we setup or if we have to reconnect) in a private variable and thus can maintain it safely in envs where multiple sessions or database connections may be present.

sbulen: Yes, this PR is OK, and should probably go in sooner rather than later...  

I've done some testing with pg using php 8.1, and there are some changes that need to be made.  I'd say *MOST* of the 8.1 changes are transparent as ID'd in the article above.  But not all.  See #7190 

This is because many of the affected 2.1 calls to pg are are mapped directly to pg functions in $smcFunc.  So they are working as advertised.  

SMF 2.0 is a completely different monster...   Many of the pg calls were overloaded, & there is additional processing in there that works with the old resource ID.  Since there is no longer a resource ID, 2.0 fails when running pg under php8.1...

sbulen (approved):

PR #7179, opened 2021-12-08
Implements smf_strftime

### LINKED ISSUE #7167: PHP 8.1: strftime() is deprecated
> The strftime() and gmstrftime() functions exhibit similar issues as strptime(), in that the formats they support, as well as their behavior, is platform-dependent. Unlike strptime(), these functions are available on Windows, though with a different feature set than on Linux. Musl-based distributions like Alpine do not support timezone-related format specifiers correctly. These functions are also locale-based, and as such may exhibit thread-safety issues.
> Once again date() or DateTime::format() provide portable alternatives, and IntlDateFormatter::format() provides a more sophisticated, localization-aware alternative.

Ref: https://wiki.php.net/rfc/deprecations_php_8_1#strftime_and_gmstrftime

> The strftime() and gmstrftime() functions have been deprecated in favor of
> date()/DateTime::format() (for locale-independent formatting) or
> IntlDateFormatter::format() (for locale-dependent formatting).

Ref: https://github.com/php/php-src/blob/1cf4fb739f7a4fa8404a4c0958f13d04eae519d4/UPGRADING#L379-L381

Other references:

 * https://www.php.net/manual/en/function.strftime.php (for the old format string characters)
 * https://www.php.net/manual/en/datetime.format.php (for the new format string characters)
 * https://www.php.net/manual/en/datetime.construct.php (see Example 2 for a Unix timestamp code sample)

Verification of the principle: https://3v4l.org/MCURg

---

Instances of strftime() in the code: https://github.com/SimpleMachines/SMF2.1/search?q=strftime

Sesquipedalian:
Fixes #7167

This avoids calling strftime() and uses PHP's built-in DateTime functions instead.

Obviously, this means smf_strftime() does not use the system's strftime library or locale setting, so results may vary in a few cases from the results of strftime():

 - %a, %A, %b, %B, %p, %P: Output will use SMF's language strings to localize these values. If SMF's language strings have not been loaded, PHP's default English strings will be used.

 - %c, %x, %X: Output will always use ISO format.

### DIFF
--- a/SSI.php
+++ b/SSI.php
@@ -2293,7 +2293,7 @@ function ssi_recentEvents($max_events = 7, $output_method = 'echo')
 		ORDER BY cal.start_date DESC
 		LIMIT ' . $max_events,
 		array(
-			'current_date' => strftime('%Y-%m-%d', forum_time(false)),
+			'current_date' => smf_strftime('%Y-%m-%d', forum_time(false)),
 			'no_board' => 0,
 		)
 	);
@@ -2308,8 +2308,8 @@ function ssi_recentEvents($max_events = 7, $output_method = 'echo')
 		// Censor the title.
 		censorText($row['title']);
 
-		if ($row['start_date'] < strftime('%Y-%m-%d', forum_time(false)))
-			$date = strftime('%Y-%m-%d', forum_time(false));
+		if ($row['start_date'] < smf_strftime('%Y-%m-%d', forum_time(false)))
+			$date = smf_strftime('%Y-%m-%d', forum_time(false));
 		else
 			$date = $row['start_date'];
 
--- a/Sources/Calendar.php
+++ b/Sources/Calendar.php
@@ -394,7 +394,7 @@ function CalendarPost()
 		$eventDatetimes = getNewEventDatetimes();
 		$context['event'] = array_merge($context['event'], $eventDatetimes);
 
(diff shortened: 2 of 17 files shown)


albertlast:
since the time stuff take a huge amount of the page creation time,
did you check the performance of the implementation?

Sesquipedalian:
I didn't check that, no. It will almost certainly be slower than the original, but we don't have much choice here. The PHP developers have made this decision for us.

albertlast:
also the other solution was not fast on the start, we had to build optmiztation around this, in the moment we know the issues by this,
same is here we need to know if a issue exists, when yes we could look deeper into this.

Sesquipedalian:
I am surprised and pleased to report that with the efficiency improvements I added in 77d0991, page load times are actually _better_ with this PR than before, at least in my tests. Here are page creation times for the current release-2.1 code and this PR. In both cases, caching was disabled and the values were created by clicking my browser's refresh button while viewing the board index.

|release-2.1|77d0991|
|-----------|---------|
|0.227|0.148|
|0.183|0.047|
|0.161|0.049|
|0.162|0.05|
|0.161|0.065|
|0.164|0.053|
|0.16|0.048|
|0.162|0.051|
|0.156|0.048|
|0.172|0.067|
|0.161|0.046|
|0.16|0.047|
|0.162|0.048|
|0.186|0.05|
|0.166|0.046|
|0.175|0.047|
|0.159|0.046|
|0.16|0.046|
|0.161|0.052|
|0.16|0.046|

albertlast:
i did some fast check on my side too,
and couldn't detect any runtime difference(php 7.4.1) between smf2.1 and your pr -> which is good.

PR #7180, opened 2021-12-10
Manually parse ZIP files

### LINKED ISSUE #7178: internal corruption of phar
### Description
Spanned ZIP files do not get unpacked, but throw warnings instead.

### Steps to reproduce
1. https://custom.simplemachines.org/index.php?mod=4120
2. download it to package dir
3. go to package manager
4. check error log
5. internal corruption of phar "C:\wamp64\www\SMF2.1\Packages\AdSenseModV1.zip" (__HALT_COMPILER(); not found)

### Synopsis
[PHP's `Phar`  library only supports ZIP files with `0x04034b50` as the magic number](https://github.com/php/php-src/blob/90b7bde61507cee1c6b37f153909d72f5b203b8c/ext/phar/phar.c#L1626). This is fine, as normally ZIP files have that magic number. However, empty ZIP files use `0x06054b50` and ones that span multiple disks use `0x07064b50`.

### Additional information
I already have some code prepared that manually goes through the input ZIP file and unpacks it, similar to how 2.0 does it.

### References
1. https://en.wikipedia.org/wiki/ZIP_(file_format)
2. https://www.pkware.com/documents/APPNOTE/APPNOTE-2.0.txt
3. https://github.com/php/php-src/blob/90b7bde61507cee1c6b37f153909d72f5b203b8c/ext/phar/phar.c#L1760
4. https://users.cs.jmu.edu/buchhofp/forensics/formats/pkzip.html

live627:
- Fixes #7178

PHP's Phar library only supports ZIP files with 0x04034b50 as the magic number. This is fine, as normally ZIP files have that magic number. However, empty ZIP files use 0x06054b50 and ones that span multiple disks use 0x07064b50. So manually parse ZIP files (which is what we already did in the early days).

### DIFF
--- a/Sources/Subs-Package.php
+++ b/Sources/Subs-Package.php
@@ -20,8 +20,7 @@ if (!defined('SMF'))
 	die('No direct access...');
 
 /**
- * Reads a .tar.gz file, filename, in and extracts file(s) from it.
- * essentially just a shortcut for read_tgz_data().
+ * Reads an archive from either a remote location or from the local filesystem.
  *
  * @param string $gzfilename The path to the tar.gz file
  * @param string $destination The path to the desitnation directory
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Are there any particular ZIP files I should use to test this, @live627?

live627:
Yes https://github.com/live627/SMF2.1/tree/unittest/tests/packages

However, those are not for the package manager; rather, some custom code must be crafted to test this. Here are my tests (ignore the tarballs) https://github.com/live627/SMF2.1/blob/unittest/tests/PackagesTest.php

Sesquipedalian (approved):
Seems to work nicely.

PR #7185, opened 2021-12-14
Doesn't assume $user_info is populated in getUserTimezone()

### LINKED ISSUE #7184: Undefined index errors in getUserTimezone() when spiders crawl forum
#### Description
When a search engine crawls the forum, `getUserTimezone()` is called before `$user_info['is_guest']` and `$context['user']` have been set, which causes errors because `getUserTimezone()` assumes that they already have been set.

### Steps to reproduce
1. Set the "Search Engine Tracking level" setting to anything other than "Disabled"
2. Set your browser's User Agent string to "Googlebot/2.1"
3. Browse the forum.

- Version/Git revision: e52c55f
- Database Type: N/A
- Database Version: N/A
- PHP Version: N/A

Backtrace:
```
#0: smf_error_handler()
Called from /Users/jon/Sites/dev/Sources/Subs.php on line 6842
#1: getUserTimezone()
Called from /Users/jon/Sites/dev/Sources/Subs.php on line 926
#2: smf_strftime()
Called from /Users/jon/Sites/dev/Sources/ManageSearchEngines.php on line 561
#3: logSpider()
Called from /Users/jon/Sites/dev/Sources/ManageSearchEngines.php on line 541
#4: SpiderCheck()
Called from /Users/jon/Sites/dev/Sources/Load.php on line 803
#5: loadUserSettings()
Called from /Users/jon/Sites/dev/index.php on line 217
#6: smf_main()
Called from /Users/jon/Sites/dev/index.php on line 187
```
It looks like I introduced this in #7179. I will submit a fix shortly.

Sesquipedalian:
Fixes #7184

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -6835,17 +6835,17 @@ function smf_list_timezones($when = 'now')
  */
 function getUserTimezone($id_member = null)
 {
-	global $smcFunc, $context, $user_info, $modSettings, $user_settings;
+	global $smcFunc, $user_info, $modSettings, $user_settings;
 	static $member_cache = array();
 
-	if (is_null($id_member) && $user_info['is_guest'] == false)
-		$id_member = $context['user']['id'];
+	if (is_null($id_member))
+		$id_member = empty($user_info['id']) ? 0 : (int) $user_info['id'];
+	else
+		$id_member = (int) $id_member;
 
 	// Did we already look this up?
-	if (isset($id_member) && isset($member_cache[$id_member]))
-	{
+	if (isset($member_cache[$id_member]))
 		return $member_cache[$id_member];
-	}
 
 	// Check if we already have this in $user_settings.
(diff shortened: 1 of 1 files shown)


Sources/Subs.php
```diff
@@ -6855,19 +6855,16 @@ function getUserTimezone($id_member = null)
...
+	$request = $smcFunc['db_query']('', '
+		SELECT timezone
+		FROM {db_prefix}members
+		WHERE id_member = {int:id_member}',
+		array(
+			'id_member' => $id_member,
+		)
+	);
+	list($timezone) = $smcFunc['db_fetch_row']($request);
+	$smcFunc['db_free_result']($request);
```
jdarwood007, line 6867: You should put back the if logic but use !empty.  We can't query the database for id_member = 0.
Unless I missed something.

Sesquipedalian: You are correct. Running the query doesn't break anything, but it is pointless. I have restored it in #7189.

PR #7193, opened 2021-12-20
Fix array to string conversion notice in PG search

albertlast:
grafik


sbulen (approved):

sbulen:
Tested this, looks good.

PR #7197, opened 2021-12-21
Force to be lowercase as only check lower case later in function scope

tinoest:


albertlast:
should you not fix the code, instead of repairing it in this functions?

tinoest:
The fix would be to force it with a enum. But that’s not an option. Other option is to reject any invalid value. 

This is just sensible to make the input match the check.

albertlast:
i mean the calling code to fix instead here.

tinoest:
> i mean the calling code to fix instead here.

Yes, I understand and my original response answers that question. Best practice is an enum or to reject an invalid value.

This is imo the next best alternative, which doesn’t break existing behaviour.

Sesquipedalian (approved):

PR #7200, opened 2021-12-28
prev/next arrows in page index were visible when all pages was selected

live627:


Sesquipedalian (changes requested):
I'd like to merge this for 2.1.0, actually, but the addition of type hinting to the parameters means we'd have to test this PR everywhere that a page index is ever created before we could do so. If that type hinting is removed, this PR will be able to be merged now.

Sesquipedalian (approved):

Issue #7203, opened 2021-12-28
PHP 8, undefined function shell_exec & hostname lookups


sbulen:
#### Description
It has been reported in the forum lately that folks trying php 8 are receiving "Call to undefined function shell_exec()".

When doing hostname lookups, SMF first attempts shell_exec, and if that fails, it goes on to try other things.  It ignores the errors with an "@shell_exec()".  

This appears to be another facet of the new php 8.x behavior for the error suppressor, "@".  

We need to either rethink that logic, or, treat the shell_exec fail differently in the error routine.  

This logic is the same in 2.1 & 2.0.

### Additional information/references
https://www.simplemachines.org/community/index.php?topic=580071.0

sbulen:
The workaround is to disable hostname lookups, but you really shouldn't have to do so...

jdarwood007:
We can wrap it in a function_exists and a try/catch.  Should let it gracefully handle it.
PHP has moved how the @ operator works.  I think you can get around it by doing error_reporting(0), but I can't recall.

https://www.php.net/releases/8.0/en.php
>The @ operator no longer silences fatal errors.

sbulen:
Yeah, we had to make some other changes a while back: https://github.com/SimpleMachines/SMF2.1/pull/7161/files

But that was for the upgrader only.

It may be sufficient to the main error handler as well:
https://github.com/SimpleMachines/SMF2.1/blob/4502804a886115d25e393b64d05f7b2e3f90c3dd/Sources/Errors.php#L251

tinoest:
error_reporting set to 0 just silences the error, the script still halts there. 

If you want to continue on then it need's to be in a try/catch, other option is function_exists check first. 

e.g.
`try {
echo @shell_exec('uname -a');
} catch (Error $e) {
null;
}`

sbulen:
Actually the script continues - prior to php8.  And if you have an error handler, you can control your response.

A lot of that changes with PHP8.  Prior to PHP8, the error handler was passed an error_reporting level of 0 for suppressed errors.  That is no longer the case for php8+.

I think we need to duplicate the fixes we made to the upgrader error handler to the main smf error handler (links above).

More here:
https://php.watch/versions/8.0/fatal-error-suppression

tinoest:
The link you provided states **All of these errors, if raised, halts the rest of the application from being run. The difference in PHP 8.0 is that the error message is not silenced, which would have otherwise resulted in a silent error.**

[...]

PR #7161, opened 2021-11-22
PHP 8 error suppression change

sbulen:
Fixes #7081 

This fix (or similar) may need to go back to 2.0 as well.

Issue #7081, opened 2021-09-15
Upgrader: php 8 issues upgrading old 1.0 forums


sbulen:
#### Description
The upgrader runs fine in php7.4, but reports lots of errors in php8.  A sampling:
```
	Line 80:  +++  Recounting personal message totals...Error: Undefined array key "m" File: yada\yada\upgrade.php(2155) : eval()'d code Line: 10
	Line 80:  +++  Recounting personal message totals...Error: Undefined array key "m" File: yada\yada\upgrade.php(2155) : eval()'d code Line: 10
	Line 86:  +++  Converting topic notifications...Error: Undefined array key "t" File: yada\yada\upgrade.php(2155) : eval()'d code Line: 9 done.
```

Many undefined index errors converting old themes as well.  

Note 1.1 runs fine.  This is specific to 1.0.  

The upgrade runs to completion.  I'd consider this a pretty low priority.  Would likely impact 2.0 as well - they use almost identical scripts.

### Steps to reproduce
1. Upgrade a smf 1.0 forum to 2.1 using php8

### Environment (complete as necessary)
- Version/Git revision: Current
- Database Type: mysql
- Database Version: 5.7
- PHP Version: php 8

albertlast:
Is this a supported upgrade path?
The 1.0 forum doesn't run with php 8.0 i guess.

sbulen:
Yes.

albertlast:
1.0 is not supported any more,
so a simple yes sounds wrong.

Sesquipedalian:
*Running* SMF 1.0 is unsupported. *Upgrading from* SMF 1.0 is supported. After all, if someone asks for support for SMF 1.0, how can we legitimately tell them that they should upgrade to SMF 2.1 if it is impossible to do so?

Oldiesmann:
Theoretically you should even be able to upgrade all the way from YaBB SE at this point, though you'd likely run into a lot of other problems along the way given how much PHP and MySQL have changed in 17 years...

sbulen:
I test YaBB SE along with SMF 1.0, 1.,1 and 2.0 upgrades when I run a round of upgrade tests.  That's how I found this issue.  This was the first time I did a set using php8; previous tests were 7.4.  Depending where I run the tests, they're using mysql 5.6 or 5.7.  

When issues are found, we've been fixing the 2.1 & the 2.0 upgrade scripts, so they are ~in sync.   -ish.

[...]

PR #7210, opened 2021-12-29
Handle the new error suppression code in PHP 8

live627:
- #7203
- #7204
- #7161
- #7081


sbulen:
This is definitely needed.  Thx.

PR #7211, opened 2021-12-30
Specify memory for upgrader

### LINKED ISSUE #6941: Memory limit breaks upgrader
#### Description
When the upgrade reaches this step: Upgrading post notification settings
It can fail and stop the upgrade.

The forum has a large number of members and the individual php process limit is set to 128M of memory.

The upgrade was ran via command line: php upgrade.php

### Steps to reproduce
1. Have a high number of members (500k+)
2. Have PHP set to 128M
3. Run upgrade.php via command line: php upgrade.php
4. Upgrade bails out at step: Upgrading post notification settings

- Database Type: mysql
- Database Version: 5.7
- PHP Version: 7.4

Bumping the memory limit to 512 fixes this.

sbulen:
Fixes #6941 

Sets aside some memory for the upgrader.

While at it, got rid of the safe_mode check, since safe_mode was removed in php 5.4...

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -113,12 +113,13 @@ $upcontext['steps'] = array(
 );
 // Just to remember which one has files in it.
 $upcontext['database_step'] = 3;
+
+// Secure some resources
 @set_time_limit(600);
-if (!ini_get('safe_mode'))
-{
-	ini_set('mysql.connect_timeout', -1);
-	ini_set('default_socket_timeout', 900);
-}
+@ini_set('mysql.connect_timeout', -1);
+@ini_set('default_socket_timeout', 900);
+@ini_set('memory_limit', '512M');
+
 // Clean the upgrade path if this is from the client.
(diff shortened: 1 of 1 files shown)


m4z:
Once this is merged, we should update the [Server Recommendations](https://wiki.simplemachines.org/smf/SMF2.1:Requirements_and_recommendations), which currently state "`memory_limit` set to at least 32M."

jdarwood007:
I agree.  It should be at least 128MB for most forums.  Only large forums is 512mb recommended.  I would say that.  If your a big board (500k+ posts/200k+ members), 512 otherwise 128mb.  This is just because the amount of data your big board will need to work through at times, not a issues per say with the performance of the code.

Sesquipedalian (approved):

m4z:
I have modified the docs; is [this change](https://wiki.simplemachines.org/smf/index.php?title=SMF2.1%3ARequirements_and_recommendations&type=revision&diff=31439&oldid=31436) correct?

PR #7214, opened 2021-12-30
Improve handling of funky IP data

### LINKED ISSUE #6945: MySQLConvertOldIp fails with invalid IPs
#### Description
If a IP contains any of these.  The conversion process gets stuck in a loop

172.16.60.132,
 10.0.32.24
192.168.0.33, 1

The second one has a space at the start

sbulen:
Fixes #6945

Everything gets updated to either a valid IPv4, a valid IPv6, or NULL.  No more funky strings allowed - it's fixed or NULL.

I simplified the loop logic - it seemed to hang or skip data when different values were set for LIMIT and setSize.  Simplification helped.

Note that new columns to be processed by MySQLConvertOldIPs() must have ~~a default~~ _an initial_ value of ''.  That's how we know there is work remaining.  Can't use NULL, because that's a valid outcome; we need another state.

Note I think our inet conversion logic returns some very odd values sometimes, e.g., "EVe2" and "F>I:".  Always 4 bytes.  Will research more & log a bug.

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -4963,11 +4963,6 @@ function MySQLConvertOldIp($targetTable, $oldCol, $newCol, $limit = 50000, $setS
 
 	$current_substep = !isset($_GET['substep']) ? 0 : (int) $_GET['substep'];
 
-	if (empty($_GET['a']))
-		$_GET['a'] = 0;
-	$step_progress['name'] = 'Converting ips';
-	$step_progress['current'] = $_GET['a'];
-
 	// Skip this if we don't have the column
 	$request = $smcFunc['db_query']('', '
 		SHOW FIELDS
@@ -4985,6 +4980,24 @@ function MySQLConvertOldIp($targetTable, $oldCol, $newCol, $limit = 50000, $setS
 	}
 	$smcFunc['db_free_result']($request);
 
+	// Setup progress bar
+	$request = $smcFunc['db_query']('', '
+		SELECT COUNT(DISTINCT {raw:old_col})
+		FROM {db_prefix}{raw:table_name}',
+		array(
+			'old_col' => $oldCol,
+			'table_name' => $targetTable,
+		)
+	);
+	list ($step_progress['total']) = $smcFunc['db_fetch_row']($request);
+	$smcFunc['db_free_result']($request);
+
+	if (empty($_GET['a']))
+		$_GET['a'] = 0;
+	$step_progress['name'] = 'Converting ips';
+	$step_progress['current'] = $_GET['a'];
+
+	// Main process loop
 	$is_done = false;
 	while (!$is_done)
 	{
@@ -4997,9 +5010,7 @@ function MySQLConvertOldIp($targetTable, $oldCol, $newCol, $limit = 50000, $setS
 		$request = $smcFunc['db_query']('', '
 			SELECT DISTINCT {raw:old_col}
 			FROM {db_prefix}{raw:table_name}
-			WHERE {raw:new_col} IS NULL AND
-				{raw:old_col} != {string:unknown} AND
-				{raw:old_col} != {string:empty}
+			WHERE {raw:new_col} = {string:empty}
 			LIMIT {int:limit}',
 			array(
 				'old_col' => $oldCol,
@@ -5007,7 +5018,6 @@ function MySQLConvertOldIp($targetTable, $oldCol, $newCol, $limit = 50000, $setS
 				'table_name' => $targetTable,
 				'empty' => '',
 				'limit' => $limit,
-				'unknown' => 'unknown',
 			)
 		);
 		while ($row = $smcFunc['db_fetch_assoc']($request))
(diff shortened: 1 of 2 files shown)


sbulen:
Yes, I need to add a commit to fix the default when I'm done...

other/upgrade_2-1_mysql.sql
```diff
@@ -2632,8 +2632,8 @@ if ($doChange)
 
 ---# Add the new ip columns to members
 ALTER TABLE {$db_prefix}members
-ADD COLUMN member_ip VARBINARY(16),
-ADD COLUMN member_ip2 VARBINARY(16);
+ADD COLUMN member_ip VARBINARY(16) DEFAULT '',
```
albertlast: the default '' can't be used here,
since it's not a valid ip adress so we got issues in the long run (can't safe this kind of value in pg for exp.)
same for the other places

other/upgrade.php
```diff
@@ -5049,53 +5056,20 @@ function MySQLConvertOldIp($targetTable, $oldCol, $newCol, $limit = 50000, $setS
 						ELSE NULL
 					END
 					WHERE ' . $oldCol . ' IN ({array_string:whereSet})',
-					$updates
+					array_merge($updates, $new_ips)
```
albertlast, line 5059: to mutch stuff happen here in one line my pov

other/upgrade_2-1_mysql.sql
albertlast: maybe instead write here some ip adress into which is known to be place holder like: 127.255.255.255

jdarwood007 (approved):
Looks good.  Do we need to do anything for postgresql on the defaults?

albertlast:
like i said the default value is wrong since it's not a valid ip adress,
so another solution like mention '127.255.255.255' is needed.

So instead of changing db defintion which i tought is forbidden since long time,
you should fill the column with a place holder without relay on default stuff.

jdarwood007:
Is '::' valid?

I'm more and more about using NULLs if we need defaults and letting our code do is_null checks.

albertlast:
'::' look like a valid one: https://en.wikipedia.org/wiki/IPv6_address#Special_addresses
tested with pg and php 7

jdarwood007:
Lets use that then and if needed update any code handling it to treat :: as empty/invalid ip?  In SMF's context then, :: will not occur

sbulen:
As I stated above, I was going to drop the '' default, as it was only needed for the ConvertIP process.  

But setting it to '' before using it works, too, it's just a bit slower.  Latest commit does that.  No '' default, even temporarily.  A bit safer.

I prefer the current solution to using a dummy IP, because that way we are not displaying a different funky IP forever.  

Note that when ConvertIP is done, there are no more ''s.  It loops until they're gone.  Everything is converted.  Non-translatable IPs are NULL.

sbulen:
BTW, the inet 'bug' I saw earlier was not a bug.  In phpMyAdmin, if it _can_ display a VARBINARY as text, it does...  Some of the IPs actually end up as 4 bytes of ASCII characters...  

EVe2 = 0x45566532 = 69.86.101.50...

other/upgrade.php
sbulen: Works fine.  Nothing wrong with using an array_merge when necessary.

other/upgrade_2-1_mysql.sql
sbulen: Got rid of the default.  Set it with an UPDATE statement instead.  Slower, but a bit safer.

other/upgrade.php
```diff
@@ -4997,49 +5010,43 @@ function MySQLConvertOldIp($targetTable, $oldCol, $newCol, $limit = 50000, $setS
 		$request = $smcFunc['db_query']('', '
 			SELECT DISTINCT {raw:old_col}
 			FROM {db_prefix}{raw:table_name}
-			WHERE {raw:new_col} IS NULL AND
-				{raw:old_col} != {string:unknown} AND
-				{raw:old_col} != {string:empty}
+			WHERE {raw:new_col} = {string:empty}
```
live627, line 5013: Shouldn't this check for `NULL` instead of empty string so that it can skip those on subsequent future runs?

sbulen: No, it should check for empty string.  That is how the field is initialized.

Note that when you pass anything other than an IP to $smcFunc for an inet datatype, is stores NULL.  So we need to differentiate between 'not yet done' and 'done but funky'.  empty = not yet done.  NULL = funky.

albertlast:
was the code already ready for pr?
by looking at the code, i see some issue when runned multiple times... special the mysql upgrade sql part.

sbulen:
Yes, it was ready.  And yes, I tested multiple runs many times.

If you test it and run into an issue, please log a new issue.

sbulen:
@albertlast - I see it now.  Followup PR coming...

PR #7215, opened 2022-01-01
Chunk up log_action member fix

### LINKED ISSUE #6946: Fix missing values in log_actions timeout protection
#### Description
```
Fix missing values in log_actions
```

This does not have any timeout protection.  Current system has over 100k log_actions.  This step attempts to query through all 100k at once and upgrade them.  From testing it can perform about 500 every 10 seconds.  Most of this delay is the process of unserialize and update commands.

Should add some timeout protection to prevent this from being an issue.

sbulen:
Fixes #6946

### DIFF
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -3687,27 +3687,79 @@ foreach($files AS $filename)
 
 ---# Fix missing values in log_actions
 ---{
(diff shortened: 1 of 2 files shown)


jdarwood007 (approved):

jdarwood007:
I see that the software year has hit us.  I will send up PRs shortly.

PR #7218, opened 2022-01-02
Upgrader attachment & avatar fixes

### LINKED ISSUE #6942: Attachments failed eval
#### Description
During a CLI upgrade.  The upgrader crashed.  Restarting the upgrade resulted in this:

```
 *  +++  Adding more space to the mime_type column. done.
 +++  Converting legacy attachments.Error: Trying to access array offset on value of type bool File: /upgrade.php(2148) : eval()'d code Line: 65Error: Trying to access array offset on value of type bool File: /upgrade.php(2148) : eval()'d code Line: 65
```

As a temp patch, the following code was used in upgrade.php
```
				try
				{
					if(@eval('global $db_prefix, $modSettings, $smcFunc, $txt, $upcontext, $db_name; ' . $current_data))
						throw new ErrorException('eval error');
				}
				catch (Throwable $e) {
					$upcontext['error_message'] = 'Error in upgrade script ' . basename($filename) . ' on line ' . $line_number . '!' . $endl;
					if ($command_line)
						echo $upcontext['error_message'];
				}
				catch (ErrorException $e) {
					$upcontext['error_message'] = 'Error in upgrade script ' . basename($filename) . ' on line ' . $line_number . '!' . $endl;
					if ($command_line)
						echo $upcontext['error_message'];
				}
```

This let it continue.  But the original issue still exists.  There is nothing on that step that would eval on line 65 that results in a bool.

More research and testing will be needed.

This bug report essentially is two issues.

First is, we should have that try/catch, but disable it by default.  Its way late in 2.1's process to introduce that with little testing.  But by having it in there but disabled, we can have support enable it for upgrade issues if needed.

The second issue is why the attachments are failing or erroring out.

- Version/Git revision: latet git
- Database Type: mysql
- Database Version: 5.7
- PHP Version: 7.4

### LINKED ISSUE #7082: Upgrader: Tries to delete custom_avatar directory
#### Description
I've never seen this one before...
`+++  Converting legacy attachments.Error: rmdir(yada/yada/custom_avatar): Directory not empty File: yada/yada\upgrade.php(2155) : eval()'d code Line: 31. done.
`

This particular run was via CLI, so I could see the error.  

### Steps to reproduce
1. Upgrade a 2.0 forum that already has a folder for custom_avatar

- Database Type:  mysql
- Database Version:  5.7
- PHP Version:  8.0

This is being run in a test wamp environment.  Initial suspicion is directory separators are confused, so it thinks that the folder names are different when they're not...

sbulen:
Fixes #7082 
Fixes #6942 

The attachment issues were only seen on restarts or reruns for forums with multiple attachment directories.  Just needed to keep better track of when it really completes successfully.  (json_done != attachments_done...)

We used to see 'failed to unserialize' for the attachment dir setting via CLI before, this PR cleans that up as well.  We only need to serialize it once...

The custom av issue only appeared if you had directory separators misaligned (e.g., \custom_avatar vs /custom_avatar...) - pretty rare, but I did in fact have it delete my custom_avatar directory multiple times during test...

### DIFF
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -390,7 +390,7 @@ if (!is_writable($custom_av_dir))
 }
 
 // If we already are using a custom dir, delete the predefined one.
-if ($custom_av_dir != $GLOBALS['boarddir'] .'/custom_avatar')
+if (realpath($custom_av_dir) != realpath($GLOBALS['boarddir'] .'/custom_avatar'))
 {
 	// Borrow custom_avatars index.php file.
 	if (!file_exists($custom_av_dir . '/index.php'))
--- a/other/upgrade_2-1_mysql.sql
+++ b/other/upgrade_2-1_mysql.sql
@@ -585,40 +585,42 @@ if (!empty($attachs))
 
 ---# Fixing attachment directory setting...
 ---{
-if (!is_array($modSettings['attachmentUploadDir']) && is_dir($modSettings['attachmentUploadDir']))
+// If its a directory or an array, ensure it is stored as a serialized string (prep for later serial_to_json conversion)
+// Also ensure currentAttachmentUploadDir is set even for single directories
+// Make sure to do it in memory and in db...
+if (empty($modSettings['json_done']))
 {
-	$smcFunc['db_query']('', '
-		UPDATE {db_prefix}settings
-		SET value = {string:attach_dir}
-		WHERE variable = {string:uploadDir}',
-		array(
-			'attach_dir' => json_encode(array(1 => $modSettings['attachmentUploadDir'])),
-			'uploadDir' => 'attachmentUploadDir'
-		)
-	);
-	$smcFunc['db_insert']('replace',
-		'{db_prefix}settings',
-		array('variable' => 'string', 'value' => 'string'),
-		array('currentAttachmentUploadDir', '1'),
-		array('variable')
-	);
-}
-elseif (empty($modSettings['json_done']))
-{
-	// Serialized maybe?
-	$array = is_array($modSettings['attachmentUploadDir']) ? $modSettings['attachmentUploadDir'] : @unserialize($modSettings['attachmentUploadDir']);
-	if ($array !== false)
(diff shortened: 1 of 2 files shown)


sbulen:
I still need to xlate this over to pg.  commit forthcoming.

jdarwood007 (approved):
The logic looks sound.  I haven't tested as I don't have the test database to verify it though.

sbulen:
@jdarwood007 - any further input?  I'd like to get this merged.

other/upgrade_2-1_mysql.sql
```diff
@@ -585,40 +589,42 @@ if (!empty($attachs))
...
+// Also ensure currentAttachmentUploadDir is set even for single directories
+// Make sure to do it in memory and in db...
+if (empty($modSettings['json_done']))
 {
-	// Serialized maybe?
-	$array = is_array($modSettings['attachmentUploadDir']) ? $modSettings['attachmentUploadDir'] : @unserialize($modSettings['attachmentUploadDir']);
-	if ($array !== false)
+	if (!is_array($modSettings['attachmentUploadDir']) && is_dir($modSettings['attachmentUploadDir']))
 	{
+		$modSettings['attachmentUploadDir'] = serialize(array(1 => $modSettings['attachmentUploadDir']));
```
jdarwood007, line 599: If we do this, do you need to set the current directory as well?

jdarwood007:
The database I have is upgraded at this point.  But if this is clearing that error, it should be fine.

other/upgrade_2-1_mysql.sql
sbulen: We're good here.  This is exactly what the existing logic does (the only change is serialize vs json, which was premature).

sbulen: Sorry if this is a dupe, GH seems to lose my comments....

I think we are good here.  Note this matches previous logic, the only difference is this is serial, not json, which was premature.

sbulen: Also note there is a subsequent step specifically to clean that up - "Fixing attachment directory setting..."

PR #7219, opened 2022-01-03
Set alerts to have a page limit

### LINKED ISSUE #7169: User alerts could cause performance issues
#### Description
```
/**
 * Fetch the alerts a member currently has.
 *
 * @param int $memID The ID of the member.
 * @param mixed $to_fetch Alerts to fetch: true/false for all/unread, or a list of one or more IDs.
 * @param array $limit Maximum number of alerts to fetch (0 for no limit).
 * @param array $offset Number of alerts to skip for pagination. Ignored if $to_fetch is a list of IDs.
 * @param bool $with_avatar Whether to load the avatar of the alert sender.
 * @param bool $show_links Whether to show links in the constituent parts of the alert meessage.
 * @return array An array of information about the fetched alerts.
 */
function fetch_alerts($memID, $to_fetch = false, $limit = 0, $offset = 0, $with_avatar = false, $show_links = false)
```

The $limit allows 0 for all.  The code goes on to do do just that.  If you do 0, you get all of them.  If you somehow have a ton of alerts (such as in a very active forum and alerts tagged on many boards/topics, you could possibly get a bunch of alerts tagged.  

Now combine this with the alerts_popup area/code:
```
	$counter = isset($_REQUEST['counter']) ? max(0, (int) $_REQUEST['counter']) : 0;

	$context['unread_alerts'] = array();
	if ($counter < $cur_profile['alerts'])
	{
		// Now fetch me my unread alerts, pronto!
		require_once($sourcedir . '/Profile-View.php');
		$context['unread_alerts'] = fetch_alerts($memID, false, !empty($counter) ? $cur_profile['alerts'] - $counter : 0, 0, !isset($_REQUEST['counter']));
```

Essentially it is told to load every alert the member has.  Even if that member has lets say 100,000+ alerts.

### Steps to reproduce
1. Generate 100,000+ alerts for a  single member.  You can enable admins to get notifications for new members and then have Populate.php generate a ton of new members.
2.  Hit the url a many times: action=profile;area=alerts_popup;ajax
3. Server will have performance issues.

- Database Type: mysql
- Database Version: mariadb 10.0
- PHP Version: 8.0

This should be fixed before final.

I would like to propose the following changes
In fetch_alerts:
```
	$limit = min((int) $limit > 0 ? (int) $limit : 1000, 1000);
```
This will ensure we can't use more than 1000 ever in any call to fetching alerts.  Pagination will occur if you go to the full alerts page.

In alerts_popup
```
		$limit = min(!empty($counter) ? $cur_profile['alerts'] - $counter : 25, 25);
		$context['unread_alerts'] = fetch_alerts($memID, false, $limit, 0, !isset($_REQUEST['counter']));
```
This ensures the popup is limited to 25 items.  25 is more than enough for a alerts popup that doesn't have pagination.

Any feedback on this?

jdarwood007:
Default to 25 per page, but allow anything from 1 to 1000.  Anything outside of this defaults again to 25
Fixes #7169

Attaching to 2.1.0 since it could be used to perform a DOS.  I crashed my test forum a few times while testing this alone.

### DIFF
--- a/Sources/Profile-View.php
+++ b/Sources/Profile-View.php
@@ -663,7 +663,7 @@ function fetch_alerts($memID, $to_fetch = false, $limit = 0, $offset = 0, $with_
  */
 function showAlerts($memID)
 {
-	global $context, $smcFunc, $txt, $sourcedir, $scripturl, $options;
+	global $context, $smcFunc, $txt, $sourcedir, $scripturl, $options, $modSettings;
 
 	require_once($sourcedir . '/Profile-Modify.php');
 
--- a/Sources/Profile-View.php
+++ b/Sources/Profile-View.php
@@ -236,7 +236,10 @@ function fetch_alerts($memID, $to_fetch = false, $limit = 0, $offset = 0, $with_
 	// Basic sanitation.
 	$memID = (int) $memID;
 	$unread = $to_fetch === false;
-	$limit = max(0, (int) $limit);
+	
+	if (empty($limit) || $limit > 1000)
+		$limit = min(!empty($modSettings['alerts_per_page']) && (int) $modSettings['alerts_per_page'] < 1000 ? (int) $modSettings['alerts_per_page'] : 1000, 1000);
+
 	$offset = !empty($alertIDs) ? 0 : max(0, (int) $offset);
 	$with_avatar = !empty($with_avatar);
 	$show_links = !empty($show_links);
@@ -690,7 +693,7 @@ function showAlerts($memID)
 	}
 
 	// Prepare the pagination vars.
-	$maxIndex = 10;
+	$maxIndex = !empty($modSettings['alerts_per_page']) && (int) $modSettings['alerts_per_page'] < 1000 ? min((int) $modSettings['alerts_per_page'], 1000) : 25;
 	$context['start'] = (int) isset($_REQUEST['start']) ? $_REQUEST['start'] : 0;
 	$count = alert_count($memID);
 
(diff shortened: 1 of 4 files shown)


Sources/Profile.php
```diff
@@ -922,13 +922,14 @@ function alerts_popup($memID)
 
 	// No funny business allowed
 	$counter = isset($_REQUEST['counter']) ? max(0, (int) $_REQUEST['counter']) : 0;
+	$limit = !empty($modSettings['alerts_per_page']) && (int) $modSettings['alerts_per_page'] < 1000 ? min((int) $modSettings['alerts_per_page'], 1000) : 25;
```
Sesquipedalian, line 925: $modSettings needs to be added to the global variables declaration in this function.

Sources/Profile-View.php
```diff
@@ -690,7 +693,7 @@ function showAlerts($memID)
 	}
 
 	// Prepare the pagination vars.
-	$maxIndex = 10;
+	$maxIndex = !empty($modSettings['alerts_per_page']) && (int) $modSettings['alerts_per_page'] < 1000 ? min((int) $modSettings['alerts_per_page'], 1000) : 25;
```
Sesquipedalian, line 696: $modSettings needs to be added to the global variables declaration in this function.

Sesquipedalian (changes requested):

Sources/Profile.php
jdarwood007: My bad. I  tested the logic was working with a db_quote, but I didn't test it in all places.

Sesquipedalian (approved):

PR #7224, opened 2022-01-05
Fix log_action loop encountered when data can't be corrected...

### LINKED ISSUE #7223: Fix missing values in log_actions
Running the upgrade script downloaded from GH today, about an hour ago. Doing this on an already 2.1 forum. The upgrade gets stuck on Fix missing values in log_actions. Been sitting here about an hour. I have restarted the script and to no avail. It stays on this step and does not move.

https://www.simplemachines.org/community/index.php?topic=580182.0

### Steps to reproduce
1. Download GH .zip
2.  Update an existing 2.1 forum

- Version/Git revision:
- Database Type:
- Database Version:
- PHP Version:

sbulen:
Fixes #7223 
Which was fallout from #7215 ...

Two issues:
(1) What if the bad data is in JSON, not in a serialized string?  This caught me by surprise, I thought the bug that left id_member 0 after policy acceptance was only in 2.0, but apparently, it happened _(or happens?)_ in 2.1 also.  This was a problem, because it could never get fixed, and yet it would loop until it was fixed...  So it needs to look at json, too.  
(2) A broader case of (1) above, what if the data is so bad it can't get fixed at all - whether json or serialized?  Now it keeps track of last ID processed, so it can continue from there.  

The problem with any "loop until it's all fixed" loop is that sometimes things aren't fixable...  🙄

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -4986,21 +4986,26 @@ function MySQLConvertOldIp($targetTable, $oldCol, $newCol, $limit = 50000, $setS
 	$smcFunc['db_free_result']($request);
 
 	// Setup progress bar
-	$request = $smcFunc['db_query']('', '
-		SELECT COUNT(DISTINCT {raw:old_col})
-		FROM {db_prefix}{raw:table_name}',
-		array(
-			'old_col' => $oldCol,
-			'table_name' => $targetTable,
-		)
-	);
-	list ($step_progress['total']) = $smcFunc['db_fetch_row']($request);
-	$smcFunc['db_free_result']($request);
+	if (!isset($_GET['total_fixes']) || !isset($_GET['a']))
+	{
+		$request = $smcFunc['db_query']('', '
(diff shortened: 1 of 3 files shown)


jdarwood007:
Might be better to use upgrade_unserialize, it has support for detecting json and attempting to repair some corrupted serialize strings.

sbulen:
I'll give that a shot...

sbulen:
Latest commits:
 - Use upgrade_unserialize()
 - Better account for nextSubstep() falling all the way though during very long running executions.  Turns out you lose $step_progress...  Kept total & current stats so progress is not lost.  This makes the progress bar more accurate, and also under certain circumstances, fixes an issue where the process could have been halted before it was really completed due to the loss of the progress info.

sbulen:
@jdarwood007 - any further input?  I'd like to get it merged.

jdarwood007 (approved):

PR #7226, opened 2022-01-09
Allow us to hide the unread links

jdarwood007:
This may be used in cases where we are using the forum theme from SSI and using our own link tree.

Just a minor thing that would be nice.

### DIFF
--- a/Themes/default/index.template.php
+++ b/Themes/default/index.template.php
@@ -447,7 +447,7 @@ function theme_linktree($force_show = false)
 				<div class="navigate_section">
 					<ul>';
 
-	if ($context['user']['is_logged'])
+	if ($context['user']['is_logged'] && empty($context['hide_unread_links']))
 		echo '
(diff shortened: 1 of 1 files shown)


live627:
This feels like it may belong in `$settings`; what do you think?

jdarwood007:
I thought $settings was more for the theme settings that can be changed.  If there is no way to toggle this on the settings page, does it still make sense?

DiegoAndresCortes:
I'm not a fan of this idea with a themer perspective.
I personally would move the unread links somewhere else as soon as I start working on a theme, so really I wouldn't even bring `$context['hide_unread_links']` with them

Sesquipedalian:
I think #7260 has made this unnecessary. Would you agree, @jdarwood007?

jdarwood007:
technically no, but i've got this customization in our site theme so it works still.

PR #7228, opened 2022-01-09
Exclude non-files

sbulen:
Fixes #7227

PR #7229, opened 2022-01-09
[3.0] Improve tgz file handling

### LINKED ISSUE #7227: Package manager won't parse some packages generated on OSX
#### Description
When I use a package that was generated on OS X, I get these errors during upload:
```
( ! ) Fatal error: Uncaught TypeError: vsprintf(): Argument #2 ($values) must be of type array, string given in /Sources/Errors.php on line 231
( ! ) TypeError: vsprintf(): Argument #2 ($values) must be of type array, string given in /Sources/Errors.php on line 231
```
This doesn't seem to be the root of the error, but what I'm getting back

If I manually upload the file, nothing shows up.

The file was generated using OSX's built in tar uitlity
```
➜  StopForumSpam git:(master) tar -zcf ~/Downloads/StopForumSpam-test.tgz *
```

### Steps to reproduce
1. Attempt to upload the attached file to package manger.  Errors
2. 

- Database Type: N/A
- Database Version: N/A
- PHP Version: 8.0

[StopForumSpam-test.tar.gz](https://github.com/SimpleMachines/SMF2.1/files/7835728/StopForumSpam-test.tar.gz)

OS X tar utility is adding some additional metadata (which is not a file) to the compressed file, which SMF Package Manager is trying to read out.

You should see data like this in the archive.
```
119 LIBARCHIVE.xattr.com.apple.quarantine=MDA4MTs2MWRhMDUxNTtDaHJvbWU7MEY5NERCOTgtQ0M0Ni00ODIwLTk5MEEtM0M0QTRBODY4RkVC 95 SCHILY.xattr.com.apple.quarantine=0081;61da0515;Chrome;0F94DB98-CC46-4820-990A-3C4A4A868FEB
```

This only happens I think if a file exists in the directory that is quarantined by OS X.  https://apple.stackexchange.com/questions/104712/what-causes-os-x-to-mark-a-folder-as-quarantined

live627:
- Fixes #7227

### DIFF
--- a/Sources/Packages.php
+++ b/Sources/Packages.php
@@ -1577,9 +1577,7 @@ function list_getPackages($start, $items_per_page, $sort, $params)
 
 	// Here we have a little code to help those who class themselves as something of gods, version emulation ;)
 	if (isset($_GET['version_emulate']) && strtr($_GET['version_emulate'], array('SMF ' => '')) == $the_version)
-	{
 		unset($_SESSION['version_emulate']);
-	}
 	elseif (isset($_GET['version_emulate']))
 	{
 		if (($_GET['version_emulate'] === 0 || $_GET['version_emulate'] === SMF_FULL_VERSION) && isset($_SESSION['version_emulate']))
@@ -1625,7 +1623,9 @@ function list_getPackages($start, $items_per_page, $sort, $params)
 
 		while ($package = readdir($dir))
 		{
-			if ($package == '.' || $package == '..' || $package == 'temp' || (!(is_dir($packagesdir . '/' . $package) && file_exists($packagesdir . '/' . $package . '/package-info.xml')) && substr(strtolower($package), -7) != '.tar.gz' && substr(strtolower($package), -4) != '.tgz' && substr(strtolower($package), -4) != '.zip'))
+			$basename = strtok($package, '.');
+			$ext = strtolower(strtok('.'));
+			if ($package == '.' || $package == '..' || $package == 'temp' || (!(is_dir($packagesdir . '/' . $package) && file_exists($packagesdir . '/' . $package . '/package-info.xml')) && $ext != 'tar.gz' && $ext != 'tgz' && $ext != 'zip'))
 				continue;
 
 			// Skip directories or files that are named the same.
@@ -1635,17 +1635,11 @@ function list_getPackages($start, $items_per_page, $sort, $params)
 					continue;
 				$dirs[] = $package;
 			}
-			elseif (substr(strtolower($package), -7) == '.tar.gz')
+			elseif ($ext == 'tar.gz' || $ext == 'zip' || $ext == 'tgz')
 			{
-				if (in_array(substr($package, 0, -7), $dirs))
+				if (in_array($basename, $dirs))
 					continue;
-				$dirs[] = substr($package, 0, -7);
-			}
-			elseif (substr(strtolower($package), -4) == '.zip' || substr(strtolower($package), -4) == '.tgz')
-			{
-				if (in_array(substr($package, 0, -4), $dirs))
-					continue;
-				$dirs[] = substr($package, 0, -4);
+				$dirs[] = $basename;
 			}
 
 			$packageInfo = getPackageInfo($package);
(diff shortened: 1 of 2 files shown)


jdarwood007:
Is this compatible with #7228 or an alternative fix?

live627:
This is alternative fix

sbulen:
@live627 - two things:
 - It doesn't fix the initial problem, I am still seeing the same/similar issues when the OSX files are present, e.g., WSOD in the Package Manager:
image

I am also seeing issues trying to analyze a package for the cust site - slightly different, though, it cannot find package-info.xml at all.

 - I am getting thousands of errors in the error log:
      2: Undefined array key "filename"   in Subs-Package, various lines, including 210, 206, 186, 178

I used the same test file we were using on the cust site with the OSX files.  (Since I don't have a mac I need to use @jdarwood007 's example...)

sbulen:
I saw you pushed some commits while I was writing up my earlier test results.  

_I just retested, and all the issues reported above appear to have been addressed._

I must admit, though, that rewrites of stable code always make me nervous...  One of my most repeated complaints with PRs here...

sbulen:
(When the destination is null, that is a request for a list of files in the archive.)

jdarwood007:
If both fix the same thing, then I guess its a mater of choice.

This one feels like it may be more complete in trying to handle the tar file according to the RFC.

However we are late in 2.1 stages.  I see it also added hints.  While I like type hints, we have to be very careful of their usage in 2.1, because of so many places we could possibly break it.  This PR also may need additional testing for open_basedir restriction testing, which the `$destination !== null` checks fixed by preventing us from doing file checks on directories that would start from a path most likely outside of the allowed dirs.

In all honesty, I think #7228 should be merged for 2.1.  This one should go into the future because it is more complete into the RFC spec (as I see it handling some additional types other than files).  @Sesquipedalian may want to make that decision though.

live627:
Testing this one is pretty much the same as https://github.com/SimpleMachines/SMF2.1/pull/7180#issuecomment-994229770

sbulen:
@live627 - I'm not sure this is a hard requirement for 2.1.0?

live627:
Not important enough to hold up the release. The excising code works fine with the ubiquitous ustar format, just not the "newer" pax format.

I'll  let @Sesquipedalian assign a new milestone as he sees fit.

Sesquipedalian:
I'd rather merge the full fix. However, there's really no hurry for dealing with this issue. I've been making SMF compatible tar.gz files on my Mac for years; it's just a matter of setting the right command line switches. Anyway, we can take the time to test this PR as thoroughly as we like after 2.1.0 is out.

jdarwood007:
FYI, this didn't get merged into 2.1.0.  Can we do the #7228 for 2.1.x?

live627:
The speedup mentioned in the latest commit is 0.2 seconds with a package directory filled with 89 objects; package directories, php files, package files.

Sources/Packages.php
```diff
@@ -1587,25 +1587,37 @@ function getPackages()
 
 	$installed_mods = loadInstalledPackages();
 
+	$packagesdir='C:\Users\John\nf\df';
```
jdarwood007, line 1590: Derp

Sources/Subs-Package.php
```diff
@@ -146,56 +146,60 @@ function read_tgz_data(string $data, ?string $destination, bool $single_file = f
...
-				);
-
-				$checksum = 256;
-				for ($i = 0; $i < 512; $i++)
-					if (($i < 148 || $i > 155) && $data[($block << 9) + $i] != "\0")
-						$checksum += ord($data[($block << 9) + $i]);
+			if ($file_info['filename'][-1] !== '/')
+				$file_info['filename'] .= '/';
+		}
+		elseif ($file_info['type'] === 'L')
```
jdarwood007, line 154: Since your updating these.  Can you document the types heres.  Still confusing.

jdarwood007:
@live627 ,  Can you fix the conflict?

Sesquipedalian:
Closed because the original issue has been resolved in 3.0 and I don't see any need to worry about it for 2.1.

Issue #7209, opened 2021-12-29
PHP 8.1 Compatibility 


jdarwood007:
Tracking this as a general task list.  Please comment if we need to add/remove issues.  Marking this as needs to meet 2.1.0.

Pull Requests:
- [x] #7213
- [x] #7192
- [x] #7196
- [x] #7210 
- [x] #7204  
- [x] #7221
- [x] ~#7230~ #7329
- [x] #7231
- [x] #7232

Issues:
- [x] #7212
- [x] #7207
- [x] #7206
- [x] #7195 
- [x] #7190 
- [x] #7187
- [x] #7167 
- [x] #7107 
- [x] #7203 

Anyone with a PHP 8.1 test env, please test as many areas as you can and report issues.

sbulen:
I do not think 8.1 support is a hard requirement for 2.1.0 go-live...  That was never discussed.  

We cannot keep slipping scope.

In all seriousness, 8.2 is slated for a few months away.  Will we add 8.2 to scope of 2.1.0???

jdarwood007:
That is true.  @Sesquipedalian, do you feel we need PHP 8.1 support for 2.1.0 or push it off for 2.1.1?

PHP 8.2 support will also have to come.  I'm not aware of what it deprecated or broke.

live627:
> PHP 8.2 support will also have to come. I'm not aware of what it deprecated or broke.

They just released 8.1 a month ago! 8.2 won't be for another year yet.

sbulen:
> They just released 8.1 a month ago! 8.2 won't be for another year yet.

If we do not change our behavior, we will be having this discussion in 12 months for 2.1.0.

live627:
+5 magic

albertlast:
8.1 look like big hole of work,
so when you plane somekind of release i would say without 8.1 support.

so when a admin want to run 8.1 he had to disable notice about deprecation.

jdarwood007:
I agree.  Lets push this into 2.1.1.  @Sesquipedalian can move it back if he feels it needs to ship into 2.1.0.

We have added a few more PRs/issues already today for 8.1.  So this is a useful thread as is to show how much work we need to do in order to get 8.1 compatible.

Sesquipedalian:
I agree entirely.

SugarD-x:
[...]

PR #7230, opened 2022-01-10
coerce $sprintf to array

live627:
Doing it this way is much easier than fixing all the calls.

- #7209
- #7227


live627:
https://github.com/SimpleMachines/SMF2.1/pull/7329

PR #7231, opened 2022-01-10
pass the correct type to html_entity_decode()

### LINKED ISSUE #7206: php 8.1 notice : html_entity_decode(): Passing null to parameter #2 ($flags) of type int is deprecated
#### Description
while testing i get this error message:
```
Notice
: html_entity_decode(): Passing null to parameter #2 ($flags) of type int is deprecated in
F:\github\SMF2.1\Sources\Subs.php
on line
2864


Notice
: html_entity_decode(): Passing null to parameter #2 ($flags) of type int is deprecated in
F:\github\SMF2.1\Sources\Subs.php
on line
2865


Notice
: html_entity_decode(): Passing null to parameter #2 ($flags) of type int is deprecated in
F:\github\SMF2.1\Sources\Subs.php
on line
2865


Notice
: html_entity_decode(): Passing null to parameter #2 ($flags) of type int is deprecated in
F:\github\SMF2.1\Sources\Subs.php
on line
2866


Notice
: html_entity_decode(): Passing null to parameter #2 ($flags) of type int is deprecated in
F:\github\SMF2.1\Sources\Subs.php
on line
2866

...
```

### Steps to reproduce
1.  index page

- Database Type: pg
- Database Version:
- PHP Version: 8.1

https://www.php.net/manual/de/function.html-entity-decode.php
https://github.com/SimpleMachines/SMF2.1/blob/4502804a886115d25e393b64d05f7b2e3f90c3dd/Sources/Subs.php#L2864-L2873

lines:
2864
2865
2866
2867
2870
2871
2872
2873

live627:
Found while searching for a PM

- #7209

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -2754,8 +2754,8 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 					// An &nbsp; right after a URL can break the autolinker
 					if (strpos($data, '&nbsp;') !== false)
 					{
-						$placeholders[html_entity_decode('&nbsp;', null, $context['character_set'])] = '&nbsp;';
-						$data = strtr($data, array('&nbsp;' => html_entity_decode('&nbsp;', null, $context['character_set'])));
+						$placeholders[html_entity_decode('&nbsp;', 0, $context['character_set'])] = '&nbsp;';
+						$data = strtr($data, array('&nbsp;' => html_entity_decode('&nbsp;', 0, $context['character_set'])));
 					}
 
 					// Some reusable character classes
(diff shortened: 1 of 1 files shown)


Sesquipedalian (changes requested):
Actually, for all these cases the `null` was intended to tell PHP to use the default value. So instead of `0` these should all be changed to that default, i.e. `ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401`.

Oldiesmann:
Note that the default flags changed in 8.1 according to the PHP manual. Prior to 8.1 it was ENT_COMPAT. I'm not sure how this affects us but we need to make sure we don't screw anything up with whichever flag(s) we choose

Sesquipedalian:
> Note that the default flags changed in 8.1 according to the PHP manual. Prior to 8.1 it was ENT_COMPAT. I'm not sure how this affects us but we need to make sure we don't screw anything up with whichever flag(s) we choose

True. Good catch.

... And actually, as I look at this PR again, I now realize that it doesn't make any difference which flags we use in these particular cases. None of the decoded entities here will be affected by any of those flags, so `0` is perfectly fine.

Sesquipedalian (approved):

PR #7233, opened 2022-01-10
Avoid errors if the currentAttachmentUploadDir is invalid

BrickOzp:
If the currentAttachmentUploadDir points to an invalid
attachment index, an error would be added to the error log.
Avoid this by checking if the index exists before using it.

### DIFF
--- a/Sources/Attachments.php
+++ b/Sources/Attachments.php
@@ -116,7 +116,7 @@ class Attachments
 
 		$this->_attachmentUploadDir = $modSettings['attachmentUploadDir'];
 
-		$this->_attchDir = $context['attach_dir'] = $this->_attachmentUploadDir[$modSettings['currentAttachmentUploadDir']];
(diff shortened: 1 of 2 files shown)


sbulen:
Wouldn't we rather know of the error?  I'm not sure I understand why we would want to hide this - it seems important to fix....

BrickOzp:
I don't think "8: Undefined offset: 0" is a good error to indicate something is wrong with the attachment dirs.

It thought of adding a log_error with a nicer message, but it will create an error on every page load.
Then I found out that if a user tries to upload an attachment an error would occur and a readable message will be added to the error log and I think that would be enough.
Do you agree?

sbulen:
No , because that way, the only way the admin knows he somehow broke the attachment settings is after users get burned.  And maybe not even then - you can have a misconfigured folder and still upload attachments ok.  

If the folders are wrong, the admin should know asap

BrickOzp:
I see your point. But we do not detect any other errors with the attachment dir until the user tries to upload anything. e.g. if the path is invalid. Should we add checks for that?

sbulen:
The ACP already confirms paths are valid upon edit.  

If a path became invalid somehow (e.g., improper fix after a move) I believe errors are logged upon retrieval/viewing.  So the admin is aware.

PR #7235, opened 2022-01-11
Improves & simplifies today/yesterday logic in timeformat() 

### LINKED ISSUE #7234: "Today/Yesterday" can be wrong for users not in the same time zone as the forum's default
#### Description
If the user's time zone is not the same as the forum's default time zone, then the `timeformat()` function might end up using the incorrect "Today at" or "Yesterday at" strings for recent dates.

### Steps to reproduce
1. Set your user's time zone to something other than the forum's default time zone.
2. Find (or create) two posts that were created just before and after midnight according to your user's time zone.
3. Notice how the older post (created just before midnight) is correctly time-stamped with "Yesterday at 11:59 pm", but the newer one (created just after midnight) is incorrectly time-stamped as "Yesterday at 12:01 am" instead of "Today at 12:01 am".

- Version/Git revision: 5585668
- Database Type: N/A
- Database Version: N/A
- PHP Version: N/A

This happens because `getdate()` in `timeformat()` always uses the default time zone, but everywhere else in the process we work with the user's time zone.

I will submit a PR shortly.

Sesquipedalian:
Fixes #7234

As a nice side benefit, this also reduced the average page creation time on the board index of my test forum from 0.017 seconds to 0.012 seconds.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -912,7 +912,7 @@ function get_date_or_time_format($type = '', $format = '')
  * @param int|null $timestamp A Unix timestamp.
  *     If null, defaults to the current time.
  * @param string|null $tzid Time zone identifier.
- *     If null, defaults to the user's current time zone.
+ *     If null, uses default time zone.
  * @return string The formatted datetime string.
  */
 function smf_strftime(string $format, int $timestamp = null, string $tzid = null)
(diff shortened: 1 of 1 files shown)


live627 (commented):
Testing completed.

PR #7242, opened 2022-01-19
Handle bogus notification requests

### LINKED ISSUE #7238: database value you're trying to insert does not exist: board
#### Description
/cron.php
/Sources/tasks/CreatePost-Notify.php (Line 126) 

The database value you're trying to insert does not exist: board
Function: execute

- Version/Git revision: 4502804a8
- Database Type: mysql
- Database Version: 5.7
- PHP Version: 7.4

Also:
8: Undefined index: board
 /Sources/tasks/CreatePost-Notify.php (Line 126)

sbulen:
Fixes #7238 

Now, if the board_id isn't specified, a single error is logged & the task returns true so the background task gets deleted.  It no longer causes thousands of errors to be logged due to the background task with the broken request laying around...

While testing, noticed that notifications for quotes & mentions did not go out unless the quoted/mentioned person was watching the topic.  This is incorrect, so it was fixed.

While testing _that_, was occasionally getting undefined errors on poster_time.  Some of the logic handled it not being present, some of the logic did not.  Ensured poster_time is set.  

@Sesquipedalian - I'd appreciate it if you could take a peek since you're more familiar with that notification logic.

### DIFF
--- a/Sources/tasks/CreatePost-Notify.php
+++ b/Sources/tasks/CreatePost-Notify.php
@@ -182,10 +182,6 @@ class CreatePost_Notify_Background extends SMF_BackgroundTask
 		// Modified post, or dealing with delayed mention and quote notifications.
 		if ($type == 'edit' || !empty($this->_details['respawns']))
 		{
-			// Filter out members who have already been notified about this post's topic
-			$this->members['quoted'] = array_intersect_key($this->members['quoted'], $unnotified);
-			$this->members['mentioned'] = array_intersect_key($this->members['mentioned'], $unnotified);
-
 			// Notifications about modified posts only go to members who were mentioned or quoted
 			$this->members['watching'] = $type == 'edit' ? array(): $unnotified;
 
--- a/Sources/tasks/CreatePost-Notify.php
+++ b/Sources/tasks/CreatePost-Notify.php
@@ -84,7 +84,7 @@ class CreatePost_Notify_Background extends SMF_BackgroundTask
 	 */
 	public function execute()
 	{
-		global $smcFunc, $sourcedir, $scripturl, $language, $modSettings, $user_info;
+		global $smcFunc, $sourcedir, $scripturl, $language, $modSettings, $user_info, $txt;
 
 		require_once($sourcedir . '/Subs-Post.php');
 		require_once($sourcedir . '/Mentions.php');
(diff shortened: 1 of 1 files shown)


jdarwood007 (approved):

Sources/tasks/CreatePost-Notify.php
```diff
@@ -182,10 +195,6 @@ public function execute()
 		// Modified post, or dealing with delayed mention and quote notifications.
 		if ($type == 'edit' || !empty($this->_details['respawns']))
 		{
-			// Filter out members who have already been notified about this post's topic
-			$this->members['quoted'] = array_intersect_key($this->members['quoted'], $unnotified);
-			$this->members['mentioned'] = array_intersect_key($this->members['mentioned'], $unnotified);
-
```
Sesquipedalian, line 188: This logic is broken, yes. However, just deleting it will cause two notifications to be sent members who get mentioned in a topic they are watching, so we need to fix the logic here rather than remove it.

I suggest the following:

```
// Filter out members who have already been notified about this post's topic
$already_notified = array_diff_key($this->members['watching'], $unnotified);
$this->members['quoted'] = array_diff_key($this->members['quoted'], $already_notified);
$this->members['mentioned'] = array_diff_key($this->members['mentioned'], $already_notified);
```

Sesquipedalian (changes requested):

Sources/tasks/CreatePost-Notify.php
sbulen: Thanks, I'll take another look at this once we're past all the 2.1.0 stuff!

sbulen: Actually it already keeps track of who has already received a notification in $this->members['done'], and will not send another notice to folks in the 'done' list.  I have not been able to get it to send dupes in test.

I think we're good!

https://github.com/SimpleMachines/SMF2.1/blob/54b854bba9abddf1f351ab48eed0edc347027366/Sources/tasks/CreatePost-Notify.php#L464
https://github.com/SimpleMachines/SMF2.1/blob/54b854bba9abddf1f351ab48eed0edc347027366/Sources/tasks/CreatePost-Notify.php#L632
https://github.com/SimpleMachines/SMF2.1/blob/54b854bba9abddf1f351ab48eed0edc347027366/Sources/tasks/CreatePost-Notify.php#L706

Sesquipedalian: Fair enough then. 🙂

Sesquipedalian (approved):

PR #7247, opened 2022-01-19
Fix edit ban shows wring status of manually disabled

BrickOzp:
If a ban is set to Expired/disabled the status will be shown
as Never in the edit ban view.
Never is represented as NULL in the database and Disabled is
represented by 0. This requires a === null to distinguish.


sbulen (approved):
Issue & fix confirmed.  Looks good.

PR #7248, opened 2022-01-19
Remove all connected ban triggers when deleting a ban

BrickOzp:
Triggers connected to a ban was not removed when the ban was
removed. Corrected that.


sbulen (approved):
Issue & fix confirmed.  Looks good.

PR #7249, opened 2022-01-20
Fix PM rule apply label

BrickOzp:
If a PM rule existed with a label action, all the users
label would be applied regardless of what labels where
selected in the rule. The inbox label would not be removed
even if the user has selected that option.
Fixed this.


BrickOzp:
Reported here: https://www.simplemachines.org/community/index.php?topic=578503.0

DiegoAndresCortes:
Nice, I was curious about this one while cleaning the bugs board yesterday, and was about to test it.
Your fix is working fine for me.

Sesquipedalian (approved):

PR #7252, opened 2022-01-20
Samesite cookies

### LINKED ISSUE #6594: Add SameSite support to Set-Cookie

sbulen:
Fixes #6594 

Builds up from #6601 , I started with those commits, then...
 - Added the edit to make sure that secure cookies was on if samesite cookies were none.  During testing, this combo actually locked everyone out pretty hard...  Important to fix.
 - Fixed the issue with the UI where the selected setting wasn't displayed.
 - Added the new setting to the upgrader & installer.

(While at it, fixed a minor gap in a prior PR - no need to ever repeat attachment migration logic on a 2.1 install...)

Note that albertlast's code works fine for php versions < php7.3, following the notes under option 2.4 in this article:
https://stackoverflow.com/questions/39750906/php-setcookie-samesite-strict


Sources/ManageServer.php
```diff
@@ -555,6 +562,10 @@ function hideGlobalCookies()
 				fatal_lang_error('invalid_cookie_domain', false);
 		}
 
+		// Per spec, if samesite setting is 'none', cookies MUST be secure. Thems the rules. Else you lock everyone out...
+		if (!empty($_POST['samesiteCookies']) && ($_POST['samesiteCookies'] === 'none') && empty($_POST['secureCookies']))
+			fatal_lang_error('samesiteSecureRequired', false);
+
```
jdarwood007, line 568: I would honestly preferred if we "fixed" it instead of errored out.  But I'm fine with this for 2.1.

I think in future versions, we could attempt to "fix" the secure, which if we forced secure we would have to force https.  Maybe in this future version we could do an ajax request that would set the test cookie using the new settings and then issue another ajax request to see if the cookie is working.  Just a thought  here.

Themes/default/languages/Help.english.php
```diff
@@ -252,6 +252,7 @@
 $helptxt['enableCompressedOutput'] = 'This setting will compress output in order to lower bandwidth consumption, but it requires
 		zlib to be installed.';
 $helptxt['httponlyCookies'] = 'Cookies won\'t be accessible by scripting languages, such as JavaScript. This setting can help to reduce identity theft through XSS attacks. This can cause issues with third party scripts but should be on wherever possible.';
+$helptxt['samesiteCookies'] = 'The SameSite attribute of the Set-Cookie HTTP response header allows you to declare if your cookie should be restricted to a first-party or same-site context. For further information, please refer to <a href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite">MDN</a>';
```
jdarwood007: Just a note, and I may be guilty of this as well, the "en-US/" could be omitted from the URL, it looks like Mozilla redirects to the correct website via locale checks they do.  Although our translations can also link to a specific documentation as well here, so it isn't a big worry.  But since english is the fallback, it may produce an undesired link.  For this specific link, they only have a few translations.

other/install_2-1_mysql.sql
```diff
@@ -2145,13 +2145,15 @@ VALUES ('smfVersion', '{$smf_version}'),
...
 	('httponlyCookies', '1'),
+	('samesiteCookies', 'lax'),
 	('tfa_mode', '1'),
 	('export_dir', '{$boarddir}/exports'),
 	('export_expiry', '7'),
 	('export_min_diskspace_pct', '5'),
 	('export_rate', '250'),
 	('allow_expire_redirect', '1'),
 	('json_done', '1'),
+	('attachments_21_done', '1'),
```
jdarwood007, line 2156: Not apart of this PR?

other/install_2-1_postgresql.sql
```diff
@@ -2701,13 +2701,15 @@ VALUES ('smfVersion', '{$smf_version}'),
...
 	('httponlyCookies', '1'),
+	('samesiteCookies', 'lax'),
 	('tfa_mode', '1'),
 	('export_dir', '{$boarddir}/exports'),
 	('export_expiry', '7'),
 	('export_min_diskspace_pct', '5'),
 	('export_rate', '250'),
 	('allow_expire_redirect', '1'),
 	('json_done', '1'),
+	('attachments_21_done', '1'),
```
jdarwood007, line 2712: Not apart of this PR?

jdarwood007 (commented):
Other than my comments, this looks ok to merge.  I haven't actually tested it though.

other/install_2-1_mysql.sql
sbulen: Yes, as noted in the initial post, I caught this minor omission when looking at the install scripts.

Themes/default/languages/Help.english.php
sbulen: Should I change this?  I'm fine either way....

Sources/ManageServer.php
sbulen: I really don't like silently changing settings...   Let 'em know when they've done something wrong so they can research it better and fix it themselves.

Themes/default/languages/Help.english.php
jdarwood007: I'm not sure.  Maybe one of our translators can weigh in on this.  I see it kinda like php.net links.  I try not to link to the "manual/en" in the urls, as they will detect a locale and direct to a translated page automatically.  I kinda actually blame them for not having a "link to this page" type url which provides the most optimized url of that those linking to page can use (getting rid of any junk in the url).

Sources/ManageServer.php
jdarwood007: Yea, which we could address in a future version maybe with a hint that this setting is incompatible with another setting.  But I think again, we can address that in a future version where maybe we have a proper js/php function that can support this all natively out of the box.  The test would be an idea to ensure on something like changing the cookie settings that could break the software, that we believe (via some simple testing), that the cookie change should not break the system.  But again, future version for that since the work it requires is too much to include now.  Just my thoughts as I see this and think we could do better, if we where in the alpha/beta stages..

Themes/default/languages/Help.english.php
sbulen: It really does make sense to get rid of the hardcoded language pref.  Done.

jdarwood007 (approved):

Themes/default/languages/ManageSettings.english.php
```diff
@@ -56,6 +56,11 @@
 $txt['invalid_cookie_domain'] = 'The domain introduced seems to be invalid, please check it and save again.';
 $txt['secureCookies'] = 'Force cookies to be secure';
 $txt['httponlyCookies'] = 'Force cookies to be made accessible only through the HTTP protocol';
+$txt['samesiteCookies'] = 'Force cookies to be sent only to first parties';
+$txt['samesiteNone'] = 'None';
+$txt['samesiteLax'] = 'Lax';
+$txt['samesiteStrict'] = 'Strict';
+$txt['samesiteSecureRequired'] = 'If Same Site cookies is set to none, cookies must be secure.';
```
isaak654: Replace "is set" with "are set".

isaak654 (changes requested):

Themes/default/languages/ManageSettings.english.php
sbulen: It's a single setting, though, not multiple.  Reworded & capitalized to make it clear we are talking about a specific setting & option.

PR #7255, opened 2022-01-22
Fix wrong setting names

dragomano:


live627:
I have rebased this branch against upstream to get rid of that stray merge commit.

Sesquipedalian (approved):

Issue #7259, opened 2022-01-23
Curve2 theme issues


Sesquipedalian:
There are a number of issues with Curve2 that should be addressed. Some only require changes to CSS and could therefore theoretically be put off until 2.1.1. However, many others require HTML or JavaScript changes and therefore need to be addressed before 2.1.0. This is because we don't want theme authors using Curve2 as a base to be thrown for a loop by major changes to the default templates in 2.1.1 or beyond.

The following is not an exhaustive list, and it will probably be updated as more issues are identified.

- [x] Cramped layout around topic info on message index on small screens. (Fixed in #7260)
- [x] Cramped layout around post body on message index on small screens. (Fixed in #7260)
- [x] Post header shows too much data on small screens. (Fixed in #7260)
- [x] Feedback to date suggests people prefer the member’s title and group above the avatar rather than below. (Fixed in #7260)
- [x] Layout around the Like button is too cramped, especially on small screens but even on larger ones sometimes. (Fixed in #7260)
- [x] Posting buttons don’t float to the right on small screens. (Fixed in #7260)
- [x] Quick buttons still appear in the post header area (like in 2.0.x) in the topic summary, search results, the moderation centre, etc. These should be moved to be consistent with the layout used in the topic display.
- [x] The "Unread Posts" and "Updated Topics" links clutter up the linktree on small screens, and even on larger screens when the linktree is long. (Fixed in #7260)
- [x] Empty `lastpost` divs on the message index should be collapsed on small screens. (Fixed in #7260)
- [x] The width of "jump to" select menus changes unpredictably during user interaction. (Fixed in #7260)

- [x] Although the innermost part of the page index template is consistent, the wrapper divs placed around it are quite inconsistent across different templates, potentially causing theme authors all sorts of headaches. (Fixed in #7261)
- [x] The CSS used for the member’s name and position in the profile summary is unlike anything else in the theme. It looks nice in and of itself, sure, but it is inconsistent. (Fixed in #7261)
- [x] The `cat_bar` style for headings is designed to appear immediately above a table or visually similar structure, but in fact it is often appears on its own, which looks weird. (Fixed in #7261)
    - In some cases, rounded bottom corners should be applied to the `cat_bar` div.
    - In other cases, the following div should have the `noup` class applied to it in order to visually connect it with the preceding cat_bar
    - In yet other cases, the `cat_bar` class should be changed to some other class.
    - In a few cases, the entire `cat_bar` div is useless or redundant and can probably be removed.
[...]

PR #7260, opened 2022-01-23
Responsive fixes

Sesquipedalian:
Fixes some of the issues in #7259:

- [x] Cramped layout around topic info on message index on small screens.
- [x] Cramped layout around post body on message index on small screens.
- [x] Post header shows too much data on small screens.
- [x] Feedback to date suggests people prefer the member’s title and group above the avatar rather than below.
- [x] Layout around the Like button is too cramped, especially on small screens but even on larger ones sometimes.
- [x] Posting buttons don’t float to the right on small screens.
- [x] The "Unread Posts" and "Updated Topics" links clutter up the linktree on small screens, and even on larger screens when the linktree is long.
- [x] Empty `lastpost` divs on the message index should be collapsed on small screens.
- [x] The width of "jump to" select menus changes unpredictably during user interaction.


Themes/default/css/index.css
```diff
@@ -3304,33 +3337,38 @@ img.avatar {
...
 	padding: 2px 4px 0 4px;
 }
 .mvisible::before {
-	content: "\00ab ";
+	/* content: "\00ab "; */
 }
 .mvisible::after {
-	content: " \00bb";
+	/* content: " \00bb"; */
 }
```
DiegoAndresCortes: Should be removed instead, but if you are changing opacity in mobile, then we should probably keep them to give notice of a modified post.

Themes/default/css/index.css
```diff
@@ -1409,10 +1430,22 @@ ul li.greeting {
...
+	padding: 0;
+	float: right;
+}
+
+/* The category separator string is 29 dashes */
+#display_jump_to_select,
+#message_index_jump_to_select,
+#quick_mod_jump_to_select {
+	width: 29ch;
+}
```
DiegoAndresCortes: not sure how I feel about this.
I discussed about this months ago on discord, but I'm not happy with either behavior, and with your changes we could end up with this:
Screenshot_31

Not sure I like it, I'm not opposed to it, but I think it'd bring more questions than the previous behavior.

Themes/default/css/index.css
Sesquipedalian: You are correct. I simply forgot to go back and delete them.

Themes/default/css/index.css
Sesquipedalian: Interesting. What browser is that screenshot from? I don't see that in anything I try it in.

Themes/default/css/responsive.css
```diff
@@ -232,25 +245,58 @@
...
+		opacity: 0.6;
+	}
+	.keyinfo .postinfo a.smalltext:hover {
+		opacity: 1;
 	}
 	img.icon, #forumposts .catbg img {
 		display: none;
 	}
-	.user_info li {
+	.poster h4 {
```
DiegoAndresCortes, line 278: I like this idea.
Do you think we should remove custom fields on mobile for this area?
This is a silly example:
Screenshot_32

Themes/default/index.template.php
```diff
@@ -212,22 +212,30 @@ function template_body_above()
...
+					</a>
 					<div id="pm_menu" class="top_menu scrollable"></div>
 				</li>';
 
 		// Thirdly, alerts
 		echo '
 				<li>
-					<a href="', $scripturl, '?action=profile;area=showalerts;u=', $context['user']['id'], '"', !empty($context['self_alerts']) ? ' class="active"' : '', ' id="alerts_menu_top">', $txt['alerts'], !empty($context['user']['alerts']) ? ' <span class="amt">' . $context['user']['alerts'] . '</span>' : '', '</a>
+					<a href="', $scripturl, '?action=profile;area=showalerts;u=', $context['user']['id'], '"', !empty($context['self_alerts']) ? ' class="active"' : '', ' id="alerts_menu_top">
+						<span class="main_icons alerts"></span>
```
DiegoAndresCortes, line 235: This icon is too ugly, I prefer _watched_ or _notify_button_
We could add it as _.main_icons.forum_alerts_ so it remains with some semantic context

Themes/default/css/index.css
DiegoAndresCortes: Firefox using the responsive view

Sesquipedalian: I guess that's just Firefox's algorithm for deciding where to place the expanded select menu when one of the menu items is wider than the select menu itself. It's not overflowing the window even with that long item, though, so it seems more like an aesthetic issue that a functional one. If we have to choose between that or the menu suddenly change size when the user mouses over it, the latter is more important to solve.

Themes/default/css/responsive.css
Sesquipedalian: Perhaps, or maybe make them inline-block as well? I'm open to suggestions either way.

Themes/default/index.template.php
Sesquipedalian: I'm fine with using a different icon, sure. But for the sake of maintaining consistency we should probably just change the icon for `.main_icons.alerts` in index.css, so that the same icon is used for alerts everywhere.

Themes/default/css/responsive.css
DiegoAndresCortes: It's not visible below avatar, with icons, and neither below the poster info. Perhaps we should hide it here too so it's consistent.

Themes/default/index.template.php
DiegoAndresCortes: Ehh I'll leave that up to you, doubt this icon is used in that many places, but it's certainly used for warnings too, which is odd

Themes/default/css/responsive.css
Sesquipedalian: Agreed

Themes/default/index.template.php
Sesquipedalian: It is odd, yes...

Themes/default/css/responsive.css
Sesquipedalian: Done in 8ecac6f

Themes/default/index.template.php
Sesquipedalian: What would really be good is a nice little bell icon. That's more or less standard on other platforms these days.

DiegoAndresCortes: We don’t have, I wouldn’t mind if we add the bell icon or sprite separately, the set has a bunch of bell icons (with the actions too)

Sesquipedalian: Done.

Themes/default/css/index.css
DiegoAndresCortes: I checked this again and you forgot to add `#search_jump_to_select`

DiegoAndresCortes: @Sesquipedalian

Themes/default/css/index.css
```diff
@@ -1174,17 +1185,31 @@ img#smflogo {
 	clear: both;
 }
 /* The upper_section, float the two each way */
+#inner_wrap {
+	display: flex;
+	justify-content: space-between;
+	border-bottom: 1px solid #bbb;
```
BrickOzp, line 1191: This (line created by bottom border) does not look good for guests on mobile (<480px)

BrickOzp (commented):
If there are forum news, the unread links are moved to the left, just after the date. Is this intensional?

If there are no forum news, the unread links are aligned to the right. But the bullet after the date seems wrong IMHO

Themes/default/css/index.css
```diff
@@ -1726,13 +1726,20 @@ tr.windowbg td, tr.bg td, .table_grid tr td {
 	background: url(../images/icons/main_icons_sprite.png) no-repeat -5px -5px / 260px auto;
 	vertical-align: middle;
 }
+.main_icons.alerts::before {
+	background: url(../images/icons/bell.png);
```
BrickOzp, line 1731: Should we use this bell icon for the Notification settings in the user drop down menu?

Themes/default/index.template.php
```diff
@@ -212,22 +212,30 @@ function template_body_above()
...
 					<div id="profile_menu" class="top_menu"></div>
 				</li>';
 
 		// Secondly, PMs if we're doing them
 		if ($context['allow_pm'])
 			echo '
 				<li>
-					<a href="', $scripturl, '?action=pm"', !empty($context['self_pm']) ? ' class="active"' : '', ' id="pm_menu_top">', $txt['pm_short'], !empty($context['user']['unread_messages']) ? ' <span class="amt">' . $context['user']['unread_messages'] . '</span>' : '', '</a>
+					<a href="', $scripturl, '?action=pm"', !empty($context['self_pm']) ? ' class="active"' : '', ' id="pm_menu_top">
+						<span class="main_icons inbox"></span>
```
BrickOzp, line 224: use mail icon instead?

Themes/default/css/index.css
Sesquipedalian: Ah yes. Thanks!

Sesquipedalian:
> If there are forum news, the unread links are moved to the left, just after the date. Is this intensional?

Yes.

> If there are no forum news, the unread links are aligned to the right. But the bullet after the date seems wrong IMHO

Thanks! I'll fix that.

Themes/default/css/index.css
Sesquipedalian: I'd be open to that change. What do you think, @DiegoAndresCortes?

Themes/default/index.template.php
Sesquipedalian: I think using the inbox icon here is more consistent with the rest of the relevant UI. We use the inbox icon for the full PM inbox, and this menu popup simply provides a view of a subset of items in that inbox.

Themes/default/css/index.css
DiegoAndresCortes: Yes, makes sense and alerts are usually first choice

Themes/default/css/index.css
Sesquipedalian: Fixed in d6bac3a

Themes/default/css/index.css
Sesquipedalian: Fixed in acfa658

Themes/default/css/index.css
Sesquipedalian: Done.

DiegoAndresCortes (approved):
Looks good to me

PR #7261, opened 2022-01-23
Improves UI consistency

Sesquipedalian:
Fixes some of the issues in #7259:

- [x] Quick buttons still appear in the post header area (like in 2.0.x) in the topic summary, search results, the moderation centre, etc. These should be moved to be consistent with the layout used in the topic display.
- [x] Although the innermost part of the page index template is consistent, the wrapper divs placed around it are quite inconsistent across different templates, potentially causing theme authors all sorts of headaches.
- [x] The CSS used for the member’s name and position in the profile summary is unlike anything else in the theme. It looks nice in and of itself, sure, but it is inconsistent.
- [x] The `cat_bar` style for headings is designed to appear immediately above a table or visually similar structure, but in fact it is often appears on its own, which looks weird.
    - In some cases, rounded bottom corners should be applied to the `cat_bar` div.
    - In other cases, the following div should have the `noup` class applied to it in order to visually connect it with the preceding cat_bar
    - In yet other cases, the `cat_bar` class should be changed to some other class.
    - In a few cases, the entire `cat_bar` div is useless or redundant and can probably be removed.
- [x] When a category is collapsed on the board index and the user clicks on a category link in the linktree, the browser window may end up autoscrolling to the wrong position on the board index. This happens because the board table below a collapsed `cat_bar` is initially visible when the page loads before being immediately hidden by JavaScript, with the result that the browser initially autoscrolls the window to the original position of the target category. An easy fix is to make the collapsed content already hidden when the page first loads.
- [x] When showing a board, a topic, or even search results or recent posts, very similar information is presented to the user about what is being shown (e.g. board name and description, topic subject and author info, search query string, etc.), but we use at least three entirely different UI conventions to present this info. These should be made consistent. Since the UI used for topic display is the best of the options, the others should probably be changed to resemble that.
- [x] For no apparent reason, topic moderation UI is provided on search results when the “Show results as messages” option is not selected. This makes no sense and should be removed.
- [x] There is significant inconsistency between the presentation of posts in the topic display from the presentation of posts in search results, member profiles, the moderation centre, etc. The presentation should be made as consistent as possible.
- [x] The error log templates are (and need to be) more like post display templates that like the other log templates, but the overall UI for the error log can still be made more consistent with our UI conventions elsewhere than it is now.

### DIFF
--- a/Themes/default/Errors.template.php
+++ b/Themes/default/Errors.template.php
@@ -70,7 +70,7 @@ function template_error_log()
 				</h3>
 			</div>
 			<div class="pagesection">
-				<div class="floatleft">
+				<div class="pagelinks">
 					', $context['page_index'], '
 				</div>';
 
@@ -179,7 +179,7 @@ function template_error_log()
 
 	echo '
 			<div class="pagesection">
-				<div class="floatleft">
+				<div class="pagelinks">
 					', $context['page_index'], '
 				</div>';
 
--- a/Themes/default/Errors.template.php
+++ b/Themes/default/Errors.template.php
@@ -69,31 +69,43 @@ function template_error_log()
 					<a href="', $scripturl, '?action=helpadmin;help=error_log" onclick="return reqOverlayDiv(this.href);" class="help"><span class="main_icons help" title="', $txt['help'], '"></span></a> ', $txt['errorlog'], '
 				</h3>
 			</div>
-			<div class="pagesection">
-				<div class="pagelinks">
-					', $context['page_index'], '
-				</div>';
+			<div class="information flow_hidden">
+				<div class="additional_row">';
 
-	if (!empty($context['errors']))
+	// No errors, so just show a message and be done with it.
+	if (empty($context['errors']))
+	{
 		echo '
+					', $txt['errorlog_no_entries'], '
+				</div>
+			</div>
+		</form>';
+		return;
+	}
+
+	echo '
 				<div class="floatright">
 					<input type="submit" name="removeSelection" value="', $txt['remove_selection'], '" data-confirm="', $txt['remove_selection_confirm'], '" class="button you_sure">
 					<input type="submit" name="delall" value="', ($context['has_filter'] ? $txt['remove_filtered_results'] : $txt['remove_all']), '" data-confirm="', ($context['has_filter'] ? $txt['remove_filtered_results_confirm'] : $txt['sure_about_errorlog_remove']), '" class="button you_sure">
-				</div>';
-
-	echo '
-			</div>
-			<div class="title_bar">
-				<div style="padding: 6px 12px 5px 12px">
-					', $txt['apply_filter_of_type'], ':';
+				</div>
+				', $txt['apply_filter_of_type'], ':';
 
 	$error_types = array();
 
 	foreach ($context['error_types'] as $type => $details)
-		$error_types[] = ($details['is_selected'] ? '<span class="main_icons right_arrow"></span> ' : '') . '<a href="' . $details['url'] . '" ' . ($details['is_selected'] ? 'style="font-weight: bold;"' : '') . ' title="' . $details['description'] . '">' . ($details['error_type'] === 'critical' ? '<span class="error">' . $details['label'] . '</span>' : $details['label']) . '</a>';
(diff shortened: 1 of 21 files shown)


DiegoAndresCortes:
So far so good, you adding anything else to this one?

Themes/default/MessageIndex.template.php
```diff
@@ -69,6 +93,20 @@ function template_main()
...
+	// Let them know why their message became unapproved.
+	if ($context['becomesUnapproved'])
+		echo '
+	<div class="noticebox">
+		', $txt['post_becomes_unapproved'], '
+	</div>';
+
+		// If this person can approve items and we have some awaiting approval tell them.
+		if (!empty($context['unapproved_posts_message']))
+			echo '
```
jdarwood007: Indenting.

Themes/default/ReportedContent.template.php
```diff
@@ -30,6 +30,19 @@ function template_reported_posts()
...
+		</div>
+		<div class="pagesection">';
+
+	if (!empty($context['reports']) && !$context['view_closed'] && !empty($options['display_quick_mod']) && $options['display_quick_mod'] == 1)
+		echo '
+				<ul class="buttonlist floatright">
+					<li class="inline_mod_check">
+						<input type="checkbox" onclick="invertAll(this, this.form, \'close[]\');">
+					</li>
+				</ul>';
```
jdarwood007: Indenting.

Themes/default/ReportedContent.template.php
```diff
@@ -384,12 +397,22 @@ function template_reported_members()
...
-		<div class="pagesection">
+		<div class="pagesection">';
+
+	if (!empty($context['reports']) && !$context['view_closed'] && !empty($options['display_quick_mod']) && $options['display_quick_mod'] == 1)
+		echo '
+				<ul class="buttonlist floatright">
+					<li class="inline_mod_check">
+						<input type="checkbox" onclick="invertAll(this, this.form, \'close[]\');">
+					</li>
+				</ul>';
```
jdarwood007: Indenting.

Themes/default/Search.template.php
```diff
@@ -467,37 +414,19 @@ function template_results()
...
-				bNoRedirect: true,
-				bDisabled: true,
-				sCustomName: "move_to"
-			});';
+		<div class="pagelinks">', $context['page_index'], '</div>';
 
+	// Show a jump to box for easy navigation.
 	echo '
+		<div class="smalltext pagelinks floatright" id="search_jump_to"></div>
+		<script>
```
jdarwood007, line 429: Indenting.

Sesquipedalian:
> So far so good, you adding anything else to this one?

Probably not. 

I have been trying to think of a nice, unobtrusive way to add a "Select all" checkbox on the topic display page, but I haven't come up with any good ideas yet. If I do think of something eventually, I can always put that into another PR.

Themes/default/Search.template.php
Sesquipedalian: Er, sorry, but what's off with this one?

jdarwood007: I will retract that, i missed the closing.

Sesquipedalian: No worries.

Themes/default/Errors.template.php
```diff
@@ -69,31 +69,43 @@ function template_error_log()
...
 					<input type="submit" name="delall" value="', ($context['has_filter'] ? $txt['remove_filtered_results'] : $txt['remove_all']), '" data-confirm="', ($context['has_filter'] ? $txt['remove_filtered_results_confirm'] : $txt['sure_about_errorlog_remove']), '" class="button you_sure">
-				</div>';
-
-	echo '
-			</div>
-			<div class="title_bar">
-				<div style="padding: 6px 12px 5px 12px">
-					', $txt['apply_filter_of_type'], ':';
+				</div>
+				', $txt['apply_filter_of_type'], ':';
```
DiegoAndresCortes, line 91: There's something strange in this area, the indentation doesn't look right?

DiegoAndresCortes (commented):
Everything else seems fine

Themes/default/Errors.template.php
Sesquipedalian: It just looks weird because lines 75-84 above quickly close the wrapper divs and return when there are no errors.

DiegoAndresCortes (approved):

PR #7266, opened 2022-01-24
Avoid "install language table" to overflow the layout

BrickOzp:
In mobiles or other low resolution devices the install
language table overflows outside the screen. Avoid
this by breaking words.


DiegoAndresCortes (approved):
Well I certainly hope no one ever uses a phone to install a language, this is definitely an improvement and very usable in a tablet

PR #7267, opened 2022-01-24
Revamps sign in & sign up UI

Sesquipedalian:
1. Allows showing "Sign In" and "Sign up" buttons either in the user menu area or the main forum menu.
    - For SMF's default Curve2 theme, the buttons are shown in the user menu area.
    - A hidden setting, `$settings['login_main_menu']`, has been added to `template_init()` in index.template.php, in order to make it easy for other themes and/or mods to show the "Sign In" and "Sign Up" buttons in the main menu instead.
    - Adds support for `onclick` attributes on main menu buttons, so that the login overlay will show even if the login button is shown in the main menu.
2. A link to the registration page has been added to the login template, and the margin space has been increased between the submit button and the "Forgot your password?" link on the login form.
3. More consistently uses "sign up", "sign in", and "sign out" in $txt strings.

This PR is in response to feedback from team members on simplemachines.org.


DiegoAndresCortes:
Only noticed the logout missing conditional which you just fixed, works fine for me

Sources/Subs.php
```diff
@@ -5622,8 +5643,8 @@ function setupMenuContext()
 		$current_action = 'search';
 	elseif ($context['current_action'] == 'theme')
 		$current_action = isset($_REQUEST['sa']) && $_REQUEST['sa'] == 'pick' ? 'profile' : 'admin';
-	elseif ($context['current_action'] == 'register2')
-		$current_action = 'register';
+	elseif ($context['current_action'] == 'signup2')
```
albertlast, line 5646: looks for me an unneassary change.

Themes/default/languages/index.english.php
```diff
@@ -76,7 +76,7 @@
 $txt['remove'] = 'Remove';
 $txt['start_new_topic'] = 'Start new topic';
 
-$txt['login'] = 'Login';
+$txt['login'] = 'Sign in';
```
albertlast: i personal run always in the issue that i can't distinguish "sign in" and "sign up" dunno if this a personal issue.

albertlast (commented):
Use less clear words and got more mixed up,
since register is used on some places.

Themes/default/languages/index.english.php
m4z: Do you mean they're too similar visually, or that you mix up their meaning, or something else?

albertlast: "Sign up" and "Sign in" are both looks very similiar and for me is sign up and sign in same meaning, 
only when they stand by side i got a good chance to remeber which is the right one.

Sources/Subs.php
DiegoAndresCortes: But it's signup2 the actual action, only the functions and files are register2. This change makes sense.

Themes/default/languages/index.english.php
m4z: A bit of both, then. For me, it always helps that "sign in" / "log in" is similar (and maybe also "check in", like in a hotel where you already "registered" beforehand). And "sign (me) up" ("I wasn't on board before, but I want to be—where do I put my signature?") is kind of a level-up from being an unregistered guest. :wink:

jdarwood007: Poking around to some websites, it seems "Sign in" is becoming more used.  Even Youtube and GitHub are using "Sign In", but if you look at the url, they usually mention "login".  So it seems to me that they are interchangeable.  I would say the preference based on the url and text comparison to see if they use signin/login in either.  Some are using the same, some are using them one or the other.  Although I didn't find one that used both signup in the url and text.

Sites I looked around at: Reddit, Yahoo, Github, Outlook.com, Youtube, Google, Bitbucket, Netflix, Amazon, Linode, Digitalocean, Apple (ID), Pandora, Spotify and Discord.

Sesquipedalian: I don't honestly care whether we use "sign up", "sign in", and "sign out", or "register", "login", and "logout", but we should consistently use one set of terms or the other.

Sesquipedalian: I've just spent a few minutes looking at that list of sites @jdarwood007 mentioned above, plus Twitter and Facebook, and overall it does seem like a pretty mixed bag.

However, I did notice that most of those sites avoid placing "Sign In" and "Sign Up" buttons right next to each other: 
- The general trend seems to be that if "Sign Up" is used, then "Log In" or some other phrase is used, or conversely that if "Sign In" is used, then "Create Account" or some alternative is used.
-  Google, YouTube, and Yahoo take a different approach by presenting a single "Sign In" button and then relying on a "Create Account" button on their login form to catch and redirect people who don't have an account yet.
- Overall, it seems like the "Sign Up" + "Log In" combination tends to be preferred by sites that appeal to more tech savvy demographics, whereas the "Sign In" + "Create Account" combination tends to be preferred by sites that aim for broader appeal.

Based on these findings, I now suspect that @albertlast's concerns are well founded. In theory, "sign in" and "sign up" are both more user friendly than the old fashioned "login" and "register" terminology. Yet it seems that these companies, who spend ridiculous amounts of money researching such things, have almost all come to the conclusion that using both "sign in" and "sign up" at the same time is a bad idea. It seems quite plausible to me that user confusion, like @albertlast describes, is the reason.

Therefore, I now agree with going back to "Log In" and "Log Out", while keeping "Sign Up". If all other factors were equal, I would probably have preferred "Sign In", "Sign Out", and "Create Account", but our existing URL query parameters already match the first option. I will submit the changes shortly.

Sesquipedalian: I did, however, change "Login" and "Logout" to "Log in" and "Log out". The words without spaces are nouns, naming processes or functions, whereas the words with spaces are verbs, and the latter is what we use those $txt strings for.

albertlast: Like i said in the start, when i'm the only person, than it's fine,
but i got the issue on some pages...

Sesquipedalian: Don't worry about it, @albertlast. I'm already convinced that you were right, and I am glad you mentioned it. 🙂

Sesquipedalian:
So, any objections, or shall I merge this?

Themes/default/scripts/script.js
```diff
@@ -338,11 +338,16 @@ function reqOverlayDiv(desktopURL, sHeader, sIcon)
 {
 	// Set up our div details
 	var sAjax_indicator = '<div class="centertext"><img src="' + smf_images_url + '/loading_sm.gif"></div>';
-	var sIcon = smf_images_url + '/' + (typeof(sIcon) == 'string' ? sIcon : 'helptopics.png');
 	var sHeader = typeof(sHeader) == 'string' ? sHeader : help_popup_heading_text;
 
+	var containerOptions;
+	if (typeof(sIcon) == 'string' && sIcon.match(/\.(gif|png|jpe?g|svg|bmp|tiff)$/) != null)
```
DiegoAndresCortes, line 344: Where does it look for a hard coded icon with this change?
e.g. Before I could do something like this:
`reqOverlayDiv(this.href, \'%2$s\', \'/icons/modify_inline.png\')`
And use the modify_inline.png icon (for example), with the images_url gone, where is it looking for this icon if I were to do the same?

Sesquipedalian: Oops. That was a silly mistake. I will fix that soon.

Sesquipedalian: Fixed in b311437

DiegoAndresCortes: It's working now, thanks.

PR #7268, opened 2022-01-24
Respects $modSettings['securityDisable_moderate'] when issuing warning

Sesquipedalian:
Previous attempt to fix this in #7125 had an incorrect `!` in the logic, which made it do the opposite of what was intended.


jdarwood007 (approved):
Confirmed this fixed it

PR #7269, opened 2022-01-25
fix popwindow and move some styles to main file

DiegoAndresCortes:
Fixes this: https://www.simplemachines.org/community/index.php?topic=578710.0
And this: https://www.simplemachines.org/community/index.php?topic=580392.0

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -2787,7 +2787,7 @@ dl.addrules dt.floatleft {
 #main_menu .popup_window,
 #genericmenu .popup_window {
 	position: relative;
-	width: 480px;
+	width: auto;
 	z-index: 99;
 	box-shadow: 0 3px 6px rgba(0, 0, 0, 0.5);
 	border: 1px solid #777;
@@ -2797,6 +2797,7 @@ dl.addrules dt.floatleft {
 }
 .popup_window {
 	top: 15%;
+	width: 480px;
 }
 .popup_heading {
 	padding: 10px 8px;
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -2787,9 +2787,8 @@ dl.addrules dt.floatleft {
 #main_menu .popup_window,
 #genericmenu .popup_window {
 	position: relative;
-	width: 80vw;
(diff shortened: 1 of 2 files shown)


DiegoAndresCortes:
Should be ready for review in case I missed anything

Sesquipedalian:
This PR removes the margins around the inner div of the popups, like so:
Screen Shot 2022-01-25 at 1 15 08 PM
Is that intentional? I'm not a fan myself.

DiegoAndresCortes:
It's because originally it has 480px width, but once we are below that resolution, it's **95wv**.
Should I reduce the main width a bit more or add  a margin? Although the margin is not great if we are already in the 480px-500px range

Sesquipedalian:
That screenshot was from a desktop view.

I suggest keeping some sort of margin, possibly making the margin thinner on narrow screens but still not zero.

DiegoAndresCortes:
I know what you say, but it's because of the screen size:

_Above 500px or so_
Screen Shot 2022-01-25 at 14 23 24

_Closer to 480px (the base width of the container_
Screen Shot 2022-01-25 at 14 23 32

_Below 481px_
Screen Shot 2022-01-25 at 14 23 37


This you mention is because of the viewport vs the width, not because of the margin.
The current code behaves the same way I didn't change that, hence why I suggested using an even lower width.

Themes/default/css/index.css
```diff
@@ -2783,18 +2783,21 @@ dl.addrules dt.floatleft {
...
 }
-.popup_window {
+.popup_window,
+#main_menu .popup_window,
+#genericmenu .popup_window {
 	position: relative;
-	width: 480px;
+	width: auto;
 	z-index: 99;
-	padding: 0 6px 6px 6px;
```
Sesquipedalian, line 2790: This padding is the bit that I would like to see restored. I previously said margin, but I actually meant this padding.
Screen Shot 2022-01-25 at 4 45 08 PM

DiegoAndresCortes: ohh that makes sense

Sesquipedalian (approved):

PR #7271, opened 2022-01-27
PMs template markup

DiegoAndresCortes:
It adds the recent minimal changes to display template markup, minus the spacer
I also moved the group icons below and removed the conditional on both templates, because it's the same conditional...

### DIFF
--- a/Themes/default/Display.template.php
+++ b/Themes/default/Display.template.php
@@ -687,7 +687,6 @@ function template_single_post($message)
 								</div>';
 
 	echo '
-								
 								', !empty($message['counter']) ? '<span class="page_number floatright">#' . $message['counter'] . '</span>' : '', '
 								<div class="postinfo">
 									<span class="messageicon" ', ($message['icon_url'] === $settings['images_url'] . '/post/xx.png' && !$message['can_modify']) ? ' style="position: absolute; z-index: -1;"' : '', '>
--- a/Themes/default/Display.template.php
+++ b/Themes/default/Display.template.php
@@ -555,14 +555,14 @@ function template_single_post($message)
 			echo '
 								<li class="custom ', $custom['col_name'], '">', $custom['value'], '</li>';
 
-	// Show the post group icons, but not for guests.
-	if (!$message['member']['is_guest'])
-		echo '
-								<li class="icons">', $message['member']['group_icons'], '</li>';
-
 	// Don't show these things for guests.
 	if (!$message['member']['is_guest'])
 	{
+		// Show the post group icons
+		if (!$message['member']['is_guest'])
+			echo '
+								<li class="icons">', $message['member']['group_icons'], '</li>';
+
 		// Show the post group if and only if they have no other group or the option is on, and they are in a post group.
 		if ((empty($modSettings['hide_post_group']) || empty($message['member']['group'])) && !empty($message['member']['post_group']))
 			echo '
--- a/Themes/default/Display.template.php
+++ b/Themes/default/Display.template.php
@@ -559,8 +559,7 @@ function template_single_post($message)
 	if (!$message['member']['is_guest'])
 	{
 		// Show the post group icons
-		if (!$message['member']['is_guest'])
-			echo '
+		echo '
 								<li class="icons">', $message['member']['group_icons'], '</li>';
 
 		// Show the post group if and only if they have no other group or the option is on, and they are in a post group.
--- a/Themes/default/Display.template.php
+++ b/Themes/default/Display.template.php
@@ -687,9 +687,8 @@ function template_single_post($message)
 								</div>';
 
 	echo '
-								<span class="page_number floatright">
-									', !empty($message['counter']) ? ' #' . $message['counter'] : '', ' ', '
-								</span>
+								
+								', !empty($message['counter']) ? '<span class="page_number floatright">#' . $message['counter'] . '</span>' : '', '
 								<div class="postinfo">
 									<span class="messageicon" ', ($message['icon_url'] === $settings['images_url'] . '/post/xx.png' && !$message['can_modify']) ? ' style="position: absolute; z-index: -1;"' : '', '>
 										<img src="', $message['icon_url'] . '" alt=""', $message['can_modify'] ? ' id="msg_icon_' . $message['id'] . '"' : '', '>
--- a/Themes/default/PersonalMessage.template.php
+++ b/Themes/default/PersonalMessage.template.php
@@ -327,15 +327,10 @@ function template_single_pm($message)
 	echo '
 				<ul class="user_info">';
 
-	// Are there any custom fields below the avatar?
-	if (!empty($message['custom_fields']['below_avatar']))
-		foreach ($message['custom_fields']['below_avatar'] as $custom)
-			echo '
-					<li class="custom ', $custom['col_name'], '">', $custom['value'], '</li>';
-
-	if (!$message['member']['is_guest'])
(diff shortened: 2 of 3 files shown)


Themes/default/PersonalMessage.template.php
```diff
@@ -463,11 +464,11 @@ function template_single_pm($message)
 				</ul>
 			</div><!-- .poster -->
 			<div class="postarea">
-				<div class="flow_hidden">
-					<div class="keyinfo">
-						<h5 id="subject_', $message['id'], '">
+				<div class="keyinfo">
+					<div class="postinfo">
```
Sesquipedalian: I think in this case we don't want to use `postinfo`, because we really do want the lines inside this div to remain on separate lines.

Themes/default/PersonalMessage.template.php
```diff
@@ -495,8 +496,8 @@ function template_single_pm($message)
 						<span class="smalltext">&#171; ', $context['folder'] == 'sent' ? $txt['pm_sent_is_replied_to'] : $txt['pm_is_replied_to'], ' &#187;</span>';
```
Sesquipedalian: Please also remove all the useless `&#171;` and `&#187;` entities in this file. These ones escaped my notice when I was hunting down and killing the rest of them in #7261. 🔪💣

Themes/default/PersonalMessage.template.php
```diff
@@ -463,11 +464,11 @@ function template_single_pm($message)
 				</ul>
 			</div><!-- .poster -->
 			<div class="postarea">
-				<div class="flow_hidden">
-					<div class="keyinfo">
-						<h5 id="subject_', $message['id'], '">
```
albertlast, line 468: why the convert of h5 to div?

Sesquipedalian (changes requested):

Themes/default/PersonalMessage.template.php
DiegoAndresCortes: This helps using a single css selector since it’d be the same as the Display template

DiegoAndresCortes: Does it look in two lines for you or just one?

Themes/default/PersonalMessage.template.php
DiegoAndresCortes: Display template got converted to a div, we want it to be consistent, specially for css selectors

Themes/default/PersonalMessage.template.php
DiegoAndresCortes: Ok I finally got some sleep, and so should be fixed in 31a7415
The subject was supposed to go in before the postinfo

Themes/default/PersonalMessage.template.php
DiegoAndresCortes: Yessirrr

Themes/default/PersonalMessage.template.php
```diff
@@ -463,15 +464,15 @@ function template_single_pm($message)
...
-							', $message['subject'], '
-						</h5>';
+				<div class="keyinfo">
+					<div id="subject_', $message['id'], '" class="subject_title">
+						', $message['subject'], '
+					</div>
+					<div class="postinfo">';
```
Sesquipedalian, line 471: Unfortunately, the lines inside this div still aren't being laid out as desired. 
Screen Shot 2022-01-27 at 9 57 29 AM

I suggest adding a rule like this to index.css:
```
#personal_messages .keyinfo .postinfo span {
	width: 100%;
}
```

Themes/default/PersonalMessage.template.php
```diff
@@ -482,21 +483,21 @@ function template_single_pm($message)
 		echo '(', $txt['pm_undisclosed_recipients'], ')';
 
 	echo '
-							<strong> ', $txt['on'], ':</strong> ', $message['time'], ' &#187;
+							<strong> ', $txt['on'], ':</strong> ', $message['time'], ' 
 						</span>';
 
 	// If we're in the sent items, show who it was sent to besides the "To:" people.
 	if (!empty($message['recipients']['bcc']))
 		echo '<br>
```
Sesquipedalian: We can delete this useless `<br>`.

Themes/default/PersonalMessage.template.php
```diff
@@ -482,21 +483,21 @@ function template_single_pm($message)
...
-						<span class="smalltext">&#171; <strong> ', $txt['pm_bcc'], ':</strong> ', implode(', ', $message['recipients']['bcc']), ' &#187;</span>';
+						<span class="smalltext"><strong> ', $txt['pm_bcc'], ':</strong> ', implode(', ', $message['recipients']['bcc']), ' & </span>';
 
 	if (!empty($message['is_replied_to']))
 		echo '<br>
```
Sesquipedalian: We can delete this useless `<br>`, too.

Themes/default/PersonalMessage.template.php
```diff
@@ -482,21 +483,21 @@ function template_single_pm($message)
...
 	echo '
-							<strong> ', $txt['on'], ':</strong> ', $message['time'], ' &#187;
+							<strong> ', $txt['on'], ':</strong> ', $message['time'], ' 
 						</span>';
 
 	// If we're in the sent items, show who it was sent to besides the "To:" people.
 	if (!empty($message['recipients']['bcc']))
 		echo '<br>
-						<span class="smalltext">&#171; <strong> ', $txt['pm_bcc'], ':</strong> ', implode(', ', $message['recipients']['bcc']), ' &#187;</span>';
+						<span class="smalltext"><strong> ', $txt['pm_bcc'], ':</strong> ', implode(', ', $message['recipients']['bcc']), ' & </span>';
```
Sesquipedalian: Missed an `&` at the end of this span.

Sesquipedalian (changes requested):

Themes/default/PersonalMessage.template.php
Sesquipedalian: See new review.

Themes/default/PersonalMessage.template.php
DiegoAndresCortes: I changed the direction instead, should do the same 6738486

Themes/default/PersonalMessage.template.php
DiegoAndresCortes: Should I add the h5 back inside so titles are bold? @Sesquipedalian

DiegoAndresCortes: I changed this back in 7f32adc and removed the bold for the postinfo in pms

Themes/default/PersonalMessage.template.php
Sesquipedalian: Even better. 🙂

Themes/default/Display.template.php
```diff
@@ -688,9 +687,8 @@ function template_single_post($message)
 								</div>';
 
 	echo '
-								<span class="page_number floatright">
-									', !empty($message['counter']) ? ' #' . $message['counter'] : '', ' ', '
-								</span>
+
```
Sesquipedalian: Minor issue: Please remove the empty line here.

Sesquipedalian (changes requested):
One last little nit to pick, and then this should be good to go.

Themes/default/Display.template.php
DiegoAndresCortes: Oops

Sesquipedalian (approved):

PR #7273, opened 2022-01-27
Simplifies and tidies up the markup & CSS for admin search

Sesquipedalian:
Before:
Screen Shot 2022-01-26 at 9 28 57 PM

After:
Screen Shot 2022-01-26 at 9 30 02 PM

### DIFF
--- a/Themes/default/Admin.template.php
+++ b/Themes/default/Admin.template.php
@@ -1327,7 +1327,7 @@ function template_edit_profile_field()
  */
 function template_admin_search_results()
 {
-	global $context, $txt, $scripturl;
+	global $context, $txt;
 
 	echo '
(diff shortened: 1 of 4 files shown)


Themes/default/Admin.template.php
```diff
@@ -1331,19 +1331,12 @@ function template_admin_search_results()
```
DiegoAndresCortes, line 1331: $scripturl not used anymore

DiegoAndresCortes (changes requested):

Themes/default/Admin.template.php
Sesquipedalian: Removed in 2925e9d

DiegoAndresCortes (approved):

PR #7277, opened 2022-01-29
Uses columns in boards list on search & ignore boards pages 

### LINKED ISSUE #7186: Search boards listing
#### Description
This is more of a suggested change.  When you go to ?action=search and have a lot of boards/categories, they all display down the page in a single very long list.  This is poor use of space on wide browsers.

Could we have it properly split things along categories (to not abuse code to do HTML work) and have it float these properly so it looks good?  I'm not the html/css expert here, so just pondering thoughts.

I noticed this while searching on sm website, there is a lot of boards and the list is very long.

Sesquipedalian:
Fixes #7186


DiegoAndresCortes:
I like this, much better

PR #7284, opened 2022-02-01
Save settings fix

### LINKED ISSUE #7282: saveSettings() won't allow custom settings
#### Description
`saveSettings()` uses hard-coded lists of variable names in order to preprocess input before passing it on to `updateSettingsFile()`. This means that any custom settings, such as settings created by a mod, will be rejected. 

These list should not be hard-coded. Instead, `saveSettings()` should use the `$settings_defs` array that is currently defined inside `updateSettingsFile()` to build these lists.

### Steps to reproduce
1. Create and install a mod that tries to add a new setting that should be stored in Settings.php, and that wants to present the UI for that new setting in an existing admin panel that calls `saveSettings()` to save stuff. (See below for example.)
2. 

- Version/Git revision: 39d0943
- Database Type: N/A
- Database Version: N/A
- PHP Version: N/A

Discovered while trying to help someone create a Redis caching mod. Even after the mod author correctly configured their mod to add an input field to Administration Center ►  Server Settings ►  Caching, trying to save produced only [an "Unknown config_var 'cache_redis'" error](https://www.simplemachines.org/community/index.php?msg=4109166).

Sesquipedalian:
Fixes #7282

The main change here is that `saveSettings()` now uses the `$settings_defs` array to determine the expected types for different variables. To accommodate that, I had to move `$settings_defs` into its own helper function.

### DIFF
--- a/Sources/ManageServer.php
+++ b/Sources/ManageServer.php
@@ -1406,39 +1406,94 @@ function saveSettings(&$config_vars)
 		$_POST['boardurl'] = normalize_iri($_POST['boardurl']);
 	}
 
-	// Any passwords?
-	$config_passwords = array(
-		'db_passwd',
-		'ssi_db_passwd',
-	);
+	require_once($sourcedir . '/Subs-Admin.php');
 
-	// All the strings to write.
-	$config_strs = array(
-		'mtitle', 'mmessage',
-		'language', 'mbname', 'boardurl',
-		'cookiename',
-		'webmaster_email',
-		'db_name', 'db_user', 'db_server', 'db_prefix', 'ssi_db_user',
-		'boarddir', 'sourcedir',
-		'cachedir', 'cachedir_sqlite', 'cache_accelerator', 'cache_memcached',
-		'image_proxy_secret',
-	);
+	// Any passwords?
+	$config_passwords = array();
 
 	// All the numeric variables.
-	$config_ints = array(
-		'db_port',
-		'cache_enable',
-		'image_proxy_maxsize',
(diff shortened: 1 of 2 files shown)


Sources/ManageServer.php
```diff
@@ -1467,20 +1518,35 @@ function saveSettings(&$config_vars)
...
+				if ($temp == $_POST[$config_var])
+				{
+					$new_settings[$config_var] = $temp;
+					$is_acceptable_type = true;
+					break;
+				}
+			}
+
+			if (!$is_acceptable_type)
+				fatal_error('Invalid config_var \'' . $config_var . '\'');
```
live627, line 1544: is error logging here intended?

Sesquipedalian: Honestly, I was just following the lead of the previous version of this code. Now that you mention it, yeah, we could simply log the error and drop the invalid value rather than failing so dramatically.

Sesquipedalian: On further consideration, we probably should leave this as a `fatal_error()`. If we just use `log_error()`, we will end up showing a "The settings were successfully saved" message to the admin even though not all the settings were saved. Unfortunately, the current architecture of these functions doesn't allow us to trigger the "Your changes were not saved because: ..." message from within `saveSettings()`, so a fatal error is the only way to give the admin truthful feedback at this time.

It should be possible to change things so that `saveSettings()` becomes able to cause the "Your changes were not saved because: ..." message to appear, but that would need to happen in a future PR.

PR #7292, opened 2022-02-04
fix typo

DiegoAndresCortes:


live627:
[Pane is actually a real word.](https://www.merriam-webster.com/dictionary/pane)

DiegoAndresCortes:
Wow
It was an assumption based on it's single appearance in the files

Sesquipedalian (dismissed):

PR #7299, opened 2022-02-07
Changed a few equal-checks into identity-checks.

AtlasHackert:
Changed a few equal-checks into identity-checks, because identity-checks are (slightly) faster, and are generally better because there's no chance of type confusion. These should be quite uncontroversial, as I only changed instances where either the types were clear from the code, or there were already similar identity-checks used on the same variable elsewhere.

### DIFF
--- a/Sources/Subs-Admin.php
+++ b/Sources/Subs-Admin.php
@@ -228,7 +228,7 @@ function getFileVersions(&$versionOptions)
 		$this_dir = dir($dirname);
 		while ($entry = $this_dir->read())
 		{
-			if (substr($entry, -12) == 'template.php' && !is_dir($dirname . '/' . $entry))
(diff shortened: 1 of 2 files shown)


Sesquipedalian:
Thank you for submitting your pull request! 

While you are right that these changes are unlikely to cause issues, thorough testing would still be required, and they would complicate the process of creating the XML patch files that we issue to allow administrators to update their forums in place. Given that the marginal performance improvements would be imperceptible in real world situations, I'm afraid that the cost-benefit ratio for implementing them is not worthwhile at this stage. We would gladly have accepted them during the pre-release stage, but I'm afraid it is too late now.

Nevertheless, thank you, and if you would like to contribute another PR in the future to address one of the open issues, we will be glad to receive and review it. 🙂

PR #7305, opened 2022-02-09
Fix minor error with package uploading of invalid packages

jdarwood007:


live627:
-  Related to #7230

jdarwood007:
Doing that will cause fatal php errors everywhere its wrong (ie not an array) because of the type hint.  We need to leave these out of 2.1 and focus them into future versions where we have time to write them properly.

Sesquipedalian (approved):

PR #7314, opened 2022-02-12
Lint against all supported PHP versions

### LINKED ISSUE #7311: Should lint against all supported PHP versions
Right now our workflow for checking incoming changes includes a call to phplint, but it only checks against whatever default version it is running. We should check syntax for all supported versions so that we can avoid issues like #7308 in the future.

live627:


.github/workflows/php.yml
```diff
@@ -2,9 +2,9 @@ name: PHP Check
 
 on:
   push:
-   branches:
-      - '*'
-  pull_request:
+    branches:
+      - release-2.1
```
jdarwood007, line 6: Do you know if this means if we branch, will we have to setup a workflow on that branch, or does this one apply to all branches?  Just making sure we understand when we start the next version, how we can modify this workflow to work where its version support will be.

Sesquipedalian: I believe we can just add new branches to this list when necessary. I'll let @live627 answer that definitively, though.

live627: I can confirm.

The PHP workflow on the new feature branch will be very different: unit tests, static analysiss, etc.

jdarwood007 (approved):

PR #7326, opened 2022-02-15
Use SMF_FULL_VERSION as default version for auto suggest

### LINKED ISSUE #7325: Should be 2.1
#### Description
https://github.com/SimpleMachines/SMF2.1/blob/02fe31c8f0458014eb84caf800183eb262f3b7f7/Sources/Subs-Editor.php#L2467

- Version/Git revision: 2.1.1
- Database Type:
- Database Version:
- PHP Version:

BrickOzp:
Fixes #7325

### DIFF
--- a/Sources/Subs-Editor.php
+++ b/Sources/Subs-Editor.php
@@ -2464,7 +2464,7 @@ function AutoSuggest_Search_SMFVersions()
 
 	// Just in case we don't have ANYthing.
 	if (empty($versions))
-		$versions = array('SMF 2.0');
+		$versions = array(SMF_FULL_VERSION);
 
 	foreach ($versions as $id => $version)
 		if (strpos($version, strtoupper($_REQUEST['search'])) !== false)


Sources/Subs-Editor.php
```diff
@@ -2464,7 +2464,7 @@ function AutoSuggest_Search_SMFVersions()
 
 	// Just in case we don't have ANYthing.
 	if (empty($versions))
-		$versions = array('SMF 2.0');
+		$versions = array('SMF 2.1');
```
Sesquipedalian: It makes more sense to use SMF_FULL_VERSION here. If something has gone wrong with retrieving latest-versions.txt, the one version string we do know for sure is SMF_FULL_VERSION. In contrast, `SMF 2.1` is not a version string that has ever or will ever appear in latest-versions.txt.

Sesquipedalian (changes requested):

Sesquipedalian (approved):

PR #7329, opened 2022-02-15
Fixes vsprintf errors

### LINKED ISSUE #7320: vsprintf() passed wrong param type when an error occurs uploading attachments
#### Description
Under certain circumstances, if something goes wrong while uploading an attachment, the code that is supposed to report the attachment error itself causes a more severe error by attempting to pass a non-array to the second parameter of the `vsprintf()` function.

Initially reported on the forum at https://www.simplemachines.org/community/index.php?topic=580699.0

### Steps to reproduce
1. According to the original report, attempting to upload a certain PNG file triggered this.
2. 

- Version/Git revision: 2.1.1
- Database Type:
- Database Version:
- PHP Version:

It looks to me like we may have a number of places in the relevant code where `vsprintf()` is passed scalar values where arrays are expected. The fix will probably be to change `vsprintf()` to `sprintf()` in all these cases.

Sesquipedalian:
Fixes #7320

### DIFF
--- a/Sources/Attachments.php
+++ b/Sources/Attachments.php
@@ -433,7 +433,7 @@ class Attachments
 
 				foreach ($attachment['errors'] as $error)
 				{
-					$attachmentOptions['errors'][] = vsprintf($txt['attach_warning'], $attachment['name']);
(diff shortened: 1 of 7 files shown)


Sesquipedalian:
External review confirms that this works.

PR #7335, opened 2022-02-17
Always unset $_SESSION['temp_attachments'] after post is made.

Sesquipedalian:
Should fix some weird glitches with attachments ending up attached to the wrong post.


DiegoAndresCortes:
Well that was a head scratcher for me and probably @sbulen too since we couldn't reproduce it consistently

PR #7336, opened 2022-02-17
Time zone and calendar fixes

### LINKED ISSUE #7324: Enhancement: Better handling of user default tz on upgrade
#### Description
Just placing this out there...  We've done so much tweaking & fixing with timezones, we may have missed something obvious.

Timezones are "new" in 2.1.  And everything defaults to UTC, _even at the individual user smf_member record level_.  

Let's say you upgrade a forum and want to change the forum default, for this new feature.  Fine, you changed it.  But it doesn't take effect - all users are still on UTC...  Why?

Whatever you do at the forum level almost doesn't matter, because, in the user profile, the tz value is hard-coded _in the DB_ to 'UTC'.  And there is no easy way to update all of your users.

I was looking at this on the support forum & there really must be a better way:
https://www.simplemachines.org/community/index.php?msg=4111262

It's set as the default at the DB level:
https://github.com/SimpleMachines/SMF2.1/blob/02fe31c8f0458014eb84caf800183eb262f3b7f7/other/upgrade_2-1_mysql.sql#L2386
https://github.com/SimpleMachines/SMF2.1/blob/02fe31c8f0458014eb84caf800183eb262f3b7f7/other/install_2-1_mysql.sql#L738

Thinking aloud here, but things to consider:
 - Setting default timezone in upgrader
 - Setting default timezone in installer
 - Defaulting users to some form of 'use forum default'
 - Providing a mechanism for an admin to set all users to 'use forum default'
 - We may also consider removing/zeroing out the old user offset (at least ensuring it's not used anywhere)

### Steps to reproduce
1. Upgrade your existing forum...
2. Everybody is on UTC

- Version/Git revision: 2.1.1

### LINKED ISSUE #7359: Calender events with comma are sometimes truncated and sometimes can't be created
#### Description
Sorry if this is a duplicate, but I couldn't find an existing issue when searching for some/all of "event date comma". I myself can't replicate that part of the issue—event dates with comma can be created fine-ish (read on) on my forum; I don't know what causes the error.

Event creation error reported here:
* https://www.simplemachines.org/community/index.php?topic=578938.0
* https://www.simplemachines.org/community/index.php?topic=579402.0
* https://www.simplemachines.org/community/index.php?topic=580955.0

However, while testing this, I noticed the date strings are sometimes butchered, see below. (Maybe these are two separate issues and we should split them, but for now I suspect they are related.)

In the second, slightly older, forum with the german `time_format`, the dates are butchered and displayed as "`%a.%d.%m.%Y, %H:%M`" ("So.27.02.2022, 11:31 […] CET") instead of "`%a., %d.%m.%Y, %H:%M`" ("So., 27.02.2022, 11:31 […] CET"). For some reason, this doesn't happen in the nightly version forum with the default `time_format` ("Feb 27, 2022, 10:50 VORMITTAG […] CET").

Same thing with whole-day events:
* alpha: "So.27.02.2022" (butchered)
* nightly: "Feb 27, 2022" (fine, AFAICT)

In general, I get the feeling that `time_format` is used for both displaying dates in the forum (this is what the docs say), and programmatically working with the date, at least in the calendar context. I think this is a bad idea, for example I can imagine people even including static strings in the format string, like "year of the lord"-type stuff for religious forums, or "stardate" for other religions.

### Steps to reproduce
1. Create an event
2. Notice the begin and/or end dates contain a comma (depending on `time_format`)
3. (Also notice the date strings are sometimes butchered (some commas and some spaces—apparently only in the date portion, not the time portion—have been removed))
4. Try to post the event
5. Some (users in some) forums will see an error, others won't.

- Version/Git revision:
  - forum 1 ("nightly"): recent nightly tarball (2a2cac7(?)) with `time_format` "`%b %d, %Y, %I:%M %p`" (which might be the default),
  - forum 2 ("alpha", on the same server): 686a12c3644e653afccd2963bf319ecfcd183ab7, with non-default `time_format` "`%a., %d.%m.%Y, %H:%M`"
- Database Type: postgres
- Database Version: 9.6.something / any
- PHP Version: 7.0

I don't think my locale plays a role in this:
```
$ locale
LANG=
LANGUAGE=
LC_CTYPE="POSIX"
LC_NUMERIC="POSIX"
LC_TIME="POSIX"
LC_COLLATE="POSIX"
LC_MONETARY="POSIX"
LC_MESSAGES="POSIX"
LC_PAPER="POSIX"
LC_NAME="POSIX"
LC_ADDRESS="POSIX"
LC_TELEPHONE="POSIX"
LC_MEASUREMENT="POSIX"
LC_IDENTIFICATION="POSIX"
LC_ALL=
```

Sesquipedalian:
Fixes #7324
Fixes #7359 
Also fixes some other glitches I found with time zones and the calendar.

### DIFF
--- a/Sources/Calendar.php
+++ b/Sources/Calendar.php
@@ -125,7 +125,7 @@ function CalendarMain()
 	$month = !empty($_REQUEST['month']) ? (int) $_REQUEST['month'] : $today['month'];
 	$day = !empty($_REQUEST['day']) ? (int) $_REQUEST['day'] : (!empty($_REQUEST['month']) ? 1 : $today['day']);
 
-	$start_object = checkdate($month, $day, $year) === true ? date_create(implode('-', array($year, $month, $day))) : date_create(implode('-', array($today['year'], $today['month'], $today['day'])));
+	$start_object = checkdate($month, $day, $year) === true ? date_create(implode('-', array($year, $month, $day)) . ' ' . getUserTimezone()) : date_create(implode('-', array($today['year'], $today['month'], $today['day'])) . ' ' . getUserTimezone());
 
 	// Need an end date for the list view
 	if (!empty($_REQUEST['end_date']))
@@ -146,14 +146,14 @@ function CalendarMain()
 
 	if (isset($end_month, $end_day, $end_year) && checkdate($end_month, $end_day, $end_year))
 	{
-		$end_object = date_create(implode('-', array($end_year, $end_month, $end_day)));
+		$end_object = date_create(implode('-', array($end_year, $end_month, $end_day)) . ' ' . getUserTimezone());
 	}
 
 	if (empty($end_object) || $start_object >= $end_object)
 	{
 		$num_days_shown = empty($modSettings['cal_days_for_index']) || $modSettings['cal_days_for_index'] < 1 ? 1 : $modSettings['cal_days_for_index'];
 
-		$end_object = date_create(date_format($start_object, 'Y-m-d'));
+		$end_object = date_create(date_format($start_object, 'Y-m-d') . ' ' . getUserTimezone());
 
 		date_add($end_object, date_interval_create_from_date_string($num_days_shown . ' days'));
 	}
--- a/Sources/Calendar.php
+++ b/Sources/Calendar.php
@@ -113,7 +113,7 @@ function CalendarMain()
 	// Need a start date for all views
 	if (!empty($_REQUEST['start_date']))
 	{
-		$start_parsed = date_parse(convertDateToEnglish($_REQUEST['start_date']));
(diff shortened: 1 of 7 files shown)


sbulen:
I will be able to test over the weekend.

sbulen:
Hmm...  I cannot test due to #7366

sbulen:
Let me work on a fix for #7366 and test these together.  I might need a couple days to test properly.

sbulen (approved):
LGTM.

I looked at pg & mysql, installs & upgrades.  In each, I ensured that the forum timezone worked properly, and that an overridden user timezone worked also.

Note this was tested alongside #7376 .  It wouldn't work without it...

PR #7337, opened 2022-02-17
Ensures server settings UI is disabled when Settings.php is not writable

### LINKED ISSUE #7297: Readonly Settings still show image proxy settings changeable
#### Description
Screen Shot 2022-02-06 at 12 35 06 PM

Also, some settings on this page are database ones and could be changeable.  Seems like we could allow those.

This seems related to the force_ssl being enabled which triggers the javascript to toggle those settings disabled status.

### Steps to reproduce
1. See image.  Image proxy settings are not disabled when the Settings.php is readonly
2. 

- Database Type: N/A
- Database Version: N/A
- PHP Version: 8.0

Sesquipedalian:
Fixes #7297

### DIFF
--- a/Sources/Cache/APIs/FileBased.php
+++ b/Sources/Cache/APIs/FileBased.php
@@ -224,7 +224,8 @@ class FileBased extends CacheApi implements CacheApiInterface
 		if (!isset($context['settings_post_javascript']))
 			$context['settings_post_javascript'] = '';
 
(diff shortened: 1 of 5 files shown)


live627:
Do the other cache files need to be updated too?

Sesquipedalian:
Derp. Yes, the Memcache ones do. Good catch.

Sesquipedalian:
Done.

PR #7346, opened 2022-02-22
Explicitly define a default value for credits column in log_packages

### LINKED ISSUE #7345: Upgrade from pre 2.1 versions fails on PG
#### Description
There's a column added to a possibly prefilled table, with NOT NULL constraints - This if I understand everything correctly, should crap itself on MySQL just as well if there are entries already there. 
But it doesn't, as MySQL works around this on it's own.
```
---# Adding support for <credits> tag in package manager
ALTER TABLE {$db_prefix}log_packages
ADD COLUMN credits TEXT NOT NULL;
---#
```

Reported on the forum:
https://www.simplemachines.org/community/index.php?topic=580863

albertlast:
Mysql got an implicit default value on credits col,
to make this point clear for both databases i convert this implicit definition to an explicit defintion.

Test case:
https://dbfiddle.uk/?rdbms=postgres_14&fiddle=664b05272751062e3981ddb055ea7744

fixes: #7345

### DIFF
--- a/other/install_2-1_mysql.sql
+++ b/other/install_2-1_mysql.sql
@@ -463,7 +463,7 @@ CREATE TABLE {$db_prefix}log_packages (
 	failed_steps TEXT NOT NULL,
 	themes_installed VARCHAR(255) NOT NULL DEFAULT '',
 	db_changes TEXT NOT NULL,
-	credits TEXT NOT NULL,
+	credits TEXT NOT NULL DEFAULT '',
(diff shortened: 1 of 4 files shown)


jdarwood007:
Before this merges, we have to ensure we provide this in the upgrade logic for the patch from 2.1.1 to 2.1.2.

Sesquipedalian (approved):

PR #7367, opened 2022-03-01
Force return of config on parse_bbc(false)

MarcoCLA:
If a integration hook (or other functions, looking at `sanitizeMSCutPaste` for example) decides to alter the $message content, a call to parse_bbc(false) can have weird behavior. It can return a string when an array should be returned.

This is guarded by a small hotfix in this pull request.
In my opinion this function should be split in two functions, one for the parsing and one to return the config.

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -1515,9 +1515,10 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 	global $smcFunc, $txt, $scripturl, $context, $modSettings, $user_info, $sourcedir, $cache_enable;
 	static $bbc_lang_locales = array(), $itemcodes = array(), $no_autolink_tags = array();
 	static $disabled, $alltags_regex = '', $param_regexes = array(), $url_regex = '';
+	$returncodes = ($message === false ? true : false);
 
 	// Don't waste cycles
-	if ($message === '')
+	if ($message === '' && !$returncodes)
 		return '';
 
 	// Just in case it wasn't determined yet whether UTF-8 is enabled.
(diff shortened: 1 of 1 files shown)


jdarwood007:
Got a simplified example of what is causing this?

Sesquipedalian:
@MarcoCLA, please [open an issue](https://github.com/SimpleMachines/SMF2.1/issues/new?assignees=&labels=&template=standard_bug.md) to describe in more detail the problem that this PR is meant to address.

MarcoCLA:
### Steps to reproduce
1. 

Add this to `$sourcedir/WeirdBBC.php`
```
<?php
	function WeirdBBC_message(&$message, &$smileys, &$cache_id, &$parse_tags)
	{
		$message = "cheese";
	}
?>
```
2. 
Add hook to database:
`INSERT INTO 'smf_settings' ('variable', 'value') VALUES ('integrate_pre_parsebbc',	'$sourcedir/WeirdBBC.php|WeirdBBC_message');`
Visit page which triggers parse_bbc(false); (/index.php?action=admin;area=featuresettings;sa=bbc)
**Result in Error log:**
```
Invalid argument supplied for foreach()
/var/www/Sources/ManageServer.php (Line 1328)
```

https://github.com/SimpleMachines/SMF2.1/blob/b815ddbc80d1a5419f792299f863dde145d40743/Sources/ManageServer.php#L1326-L1328
Expects an array, not a string ("cheese" in this case).

- Version/Git revision: 2feb78058
- Database Type: unrelated
- Database Version: unrelated
- PHP Version: 8

MarcoCLA:
ping

Sesquipedalian:
`sanitizeMSCutPaste()` won't change `$message` if its value is `false`, but you are right that hooked functions could easily break things. However, I think there is a much simpler way to prevent that: #7393

PR #7368, opened 2022-03-01
Use reverse proxy IP for users

MarcoCLA:
The real client IP is not used in $user_info. This causes incorrect behavior when used behind a reverse proxy.
For example only one guest will be visible (as only the IP of the reverse proxy is used) and all logging is linked to that IP.

Again a small hotfix. A better aproach would be to use a proper getClientIP() function which processes the headers and checks for the reverse proxy IP.

Example: [Symphony getClientIP()](https://github.com/symfony/http-foundation/blob/5ef86ac7927d2de08dc1e26eb91325f9ccbe6309/Request.php#L832)

### DIFF
--- a/Sources/QueryString.php
+++ b/Sources/QueryString.php
@@ -327,12 +327,16 @@ function cleanRequest()
 
 				// Otherwise, we've got an IP!
 				$_SERVER['BAN_CHECK_IP'] = trim($ip);
+				$_SERVER['REMOTE_ADDR'] = trim($ip);
 				break;
 			}
 		}
 		// Otherwise just use the only one.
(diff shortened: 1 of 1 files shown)


jdarwood007:
This is the reason  we have $user_info['ip'] and $user_info['ip2'].  If incorrect stats/guests online are being detected because of a single ip, then we should be fixing that.  In reality, guests online shouldn't be determined by their ip, but by their session.  Some countries alone have a limited number of IPv4 and use what is essentially a CGNAT.

Sesquipedalian:
@MarcoCLA, please [open an issue](https://github.com/SimpleMachines/SMF2.1/issues/new?assignees=&labels=&template=standard_bug.md) to describe in more detail the problem that this PR is meant to address.

MarcoCLA:
What is the use-case of `$user_info['ip']` in the current implementation (`$_SERVER['REMOTE_ADDR']`)? 

If a forwarded ip is verified, I see no reason to use the REMOTE_ADDR instead of the validated forwarded ip.

Sesquipedalian:
Added to "The future" for further consideration. It will help a great deal if we are given a more fulsome description of the problem in a proper issue report.

jdarwood007:
I'm closing in favor of #8697 superseeding this.  It resolves the same issue but also cleans things up. #8696 solved this for 2.1

PR #7376, opened 2022-03-06
Drop defaults for TEXT NOT NULL columns

### LINKED ISSUE #7345: Upgrade from pre 2.1 versions fails on PG

### LINKED ISSUE #7366: BLOB, TEXT, GEOMETRY or JSON column 'credits' can't have a default value
#### Description
--Description Here--

### Steps to reproduce
1. attempted to upgrade a 2.1 RC4 installation to the github version...
2. error in step 4, database changes

- Version/Git revision: 2.1.2 github download from 1-MAR
- Database Type: mySQL
- Database Version:
- PHP Version: 7.4

Database Changes
Executing database changes
Please be patient - this may take some time on large forums. The time elapsed increments from the server to show progress is being made.
Executing: "Package Manager New Features" (7 of 71)
Error!
BLOB, TEXT, GEOMETRY or JSON column 'credits' can't have a default value

I suspect this may be related to a recent commit
https://github.com/SimpleMachines/SMF2.1/pull/7346

sbulen:
Fixes #7366 
Fixes #7345 

Alternate solution to #7346 

Most versions of MySQL will not allow you to put a default on a TEXT column.  I've read that there are some differences between versions < 8.0.13 and otherwise, strict mode or not, etc., etc...  But it's clear that the default has got to go.  MySQL appears to use an implicit default of '' when adding new NOT NULL columns to existing tables with rows and no default is specified.  

Then again, pg will squawk if a table has existing rows, and you attempt to add a NOT NULL column without a default.  Because you haven't specified a default for the existing rows, and they cannot be null.  🙄

So...  Consistent with the rest of SMF, this PR will leave the text fields with a NOT NULL and no default.  The recommended solution for pg is to add the column with a default (so existing rows get set) and then immediately drop the default.  So that's what is done here.

This will provide us with the most consistent DB definition across MySQL and pg, and the most consistent app behavior across them as well.

Note that act_reason appeared to be the only other column with the same situation.  Corrected that as well.  I left the backtrace column alone, it's inconsistent, but has been working fine.

Feedback welcome.  

_Testing completed, checked mysql & pg, installs & upgrades..._

### DIFF
--- a/other/install_2-1_mysql.sql
+++ b/other/install_2-1_mysql.sql
@@ -463,7 +463,7 @@ CREATE TABLE {$db_prefix}log_packages (
 	failed_steps TEXT NOT NULL,
 	themes_installed VARCHAR(255) NOT NULL DEFAULT '',
 	db_changes TEXT NOT NULL,
-	credits TEXT NOT NULL DEFAULT '',
+	credits TEXT NOT NULL,
 	sha256_hash TEXT,
 	PRIMARY KEY (id_install),
 	INDEX idx_filename (filename(15))
--- a/other/install_2-1_postgresql.sql
+++ b/other/install_2-1_postgresql.sql
@@ -733,7 +733,7 @@ CREATE TABLE {$db_prefix}log_packages (
 	failed_steps text NOT NULL,
 	themes_installed varchar(255) NOT NULL DEFAULT '',
 	db_changes text NOT NULL,
-	credits text NOT NULL DEFAULT '',
+	credits text NOT NULL,
 	sha256_hash TEXT,
 	PRIMARY KEY (id_install)
 );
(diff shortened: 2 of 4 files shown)


albertlast:
Looks fine to me

PR #7384, opened 2022-03-12
Fix ability to set null on a column

jdarwood007:
Something minor I noticed.  I wasn't able to use $smcFunc['db_change_column'] to set the default to null for a nullable column.  This is because we force it to a string.  So we check if the default is set to 'NULL' and the column allows nulls, then we set it to the literal NULL.

I thought about the default being set to the literal NULL as well.  But wasn't sure if this would cause confusion or other issues with coding to call this.

@albertlast I didn't have PG setup to test, but I followed the same syntax and it seems it should be fine.  Can you confirm?

### DIFF
--- a/Sources/DbPackages-mysql.php
+++ b/Sources/DbPackages-mysql.php
@@ -420,13 +420,20 @@ function smf_db_change_column($table_name, $old_column, $column_info)
 	// Allow for unsigned integers (mysql only)
 	$unsigned = in_array($type, array('int', 'tinyint', 'smallint', 'mediumint', 'bigint')) && !empty($column_info['unsigned']) ? 'unsigned ' : '';
 
+	// Fix the default.
+	$default = '';
+	if (isset($column_info['default']) && $column_info['default'] == 'NULL' && empty($column_info['not_null']))
+		$default = 'NULL';
+	else
+		$default = '\'' . $smcFunc['db_escape_string']($column_info['default']) . '\'';
+
 	if ($size !== null)
 		$type = $type . '(' . $size . ')';
 
 	$smcFunc['db_query']('', '
 		ALTER TABLE ' . $table_name . '
 		CHANGE COLUMN `' . $old_column . '` `' . $column_info['name'] . '` ' . $type . ' ' . (!empty($unsigned) ? $unsigned : '') . (!empty($column_info['not_null']) ? 'NOT NULL' : '') . ' ' .
-			(!isset($column_info['default']) ? '' : 'default \'' . $smcFunc['db_escape_string']($column_info['default']) . '\'') . ' ' .
+				(empty($default) ? '' : 'default ' . $default) . ' ' .
 			(empty($column_info['auto']) ? '' : 'auto_increment') . ' ',
 		array(
 			'security_override' => true,
--- a/Sources/DbPackages-mysql.php
+++ b/Sources/DbPackages-mysql.php
@@ -422,7 +422,7 @@ function smf_db_change_column($table_name, $old_column, $column_info)
 
 	// Fix the default.
 	$default = '';
-	if (isset($column_info['default']) && $column_info['default'] == 'NULL' && empty($column_info['not_null']))
+	if (isset($column_info['default']) && is_null($column_info['default']) && empty($column_info['not_null']))
 		$default = 'NULL';
 	else
 		$default = '\'' . $smcFunc['db_escape_string']($column_info['default']) . '\'';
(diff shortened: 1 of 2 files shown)


albertlast:
default null make no sense to me, since a column how is null able is null by default when no default is define,
so i don't understand where you try to go.

albertlast:
please had also in mind that db_change_column had changed between 2.0 and 2.1 with this pr #5722 about null able columns.

Oldiesmann:
If the column is nullable, then we should ignore $column['default'] entirely. There's no point in specifying that the column can be null if you always want a value in the column

albertlast:
from a db point of view, is this possible to had a nullable field and a default which is not null,
so from db side it make sense.

jdarwood007:
When a column is nullable, the default should be able to be set as null.  Right now, you can't say that your nullable field has a default value of NULL, but you can say it can have any other value appropriate to its data type. I do this at work.  Typically with modified tracking fields, as when a record is created, there is nothing "modified".  I then coalesce the modified and created fields to make it simple to get the values I care about, especially when working with triggers and logging tables.  Other times its for how data is filled into the system, where the initial creation of the record has fields that are required, but other fields are filled out later on in the record as it progresses through the application.  They will be null until they reach that point in the application process and then contain a value.

I'm not making the default NULL, i'm allowing us to specify that the default can be null.  If you don't specify a default (empty string), its still won't set it, just like before.  You have to specify 'NULL'. 

At least for SQL server (where I work primarily in for my job), if you do all caps NULL in a nullable field, it converts it to the literal NULL.  I figured implanting a similar structure here made sense on how to activate it.

live627 (approved):
I ran into this while updating my Custom Form Mod.

jdarwood007:
@live627 I've changed it to use the literal NULL rather than a string.  I also found another place in the mysql files that also set this.  We should merge the logic but it seems to do something with the auto increment for some reason.

Sources/DbPackages-postgresql.php
```diff
@@ -473,7 +473,14 @@ function smf_db_change_column($table_name, $old_column, $column_info)
...
 	{
-		$action = $column_info['default'] !== null ? 'SET DEFAULT \'' . $smcFunc['db_escape_string']($column_info['default']) . '\'' : 'DROP DEFAULT';
+		// Fix the default.
+		$default = '';
+		if (isset($column_info['default']) && is_null($column_info['default']) && empty($column_info['not_null']))
+			$default = 'NULL';
+		else
+			$default = '\'' . $smcFunc['db_escape_string']($column_info['default']) . '\'';
+
+		$action = $column_info['default'] !== null ? 'SET DEFAULT ' . $default : 'DROP DEFAULT';
```
albertlast: Maybe i miss something, but when in line 478 is it essential that $column_info['default'] is null,
to fill $default with null
and here you check if $column_info['default'] !== null  -> when null than "drop default" so it never read the $default.

jdarwood007: Your right.  I compared it as a string instead.

Sources/DbPackages-mysql.php
```diff
@@ -433,7 +433,7 @@ function smf_db_change_column($table_name, $old_column, $column_info)
 	$smcFunc['db_query']('', '
 		ALTER TABLE ' . $table_name . '
 		CHANGE COLUMN `' . $old_column . '` `' . $column_info['name'] . '` ' . $type . ' ' . (!empty($unsigned) ? $unsigned : '') . (!empty($column_info['not_null']) ? 'NOT NULL' : '') . ' ' .
-				(empty($default) ? '' : 'default ' . $default) . ' ' .
+				($default === '' ? '' : 'default ' . $default) . ' ' .
```
albertlast, line 436: this goes now in direction of nit-picking so feel free to ignore,
but i would like to see that this ``$default === ''`` operator is the like in pg,
so choose what you want === or !== but use this one in both files.

Sources/DbPackages-postgresql.php
```diff
@@ -480,7 +480,7 @@ function smf_db_change_column($table_name, $old_column, $column_info)
 		else
 			$default = '\'' . $smcFunc['db_escape_string']($column_info['default']) . '\'';
 
-		$action = $column_info['default'] !== null ? 'SET DEFAULT ' . $default : 'DROP DEFAULT';
+		$action = $default !== '' ? 'SET DEFAULT ' . $default : 'DROP DEFAULT';
```
albertlast: this goes now in direction of nit-picking so feel free to ignore,
but i would like to see that this ``$default === ''`` operator is the like in pg,
so choose what you want === or !== but use this one in both files.

jdarwood007:
@Sesquipedalian I think this is the PR causing issues in 2.1.2 testing.

PR #7391, opened 2022-03-21
Improve attachment caching

jdarwood007:
We do a bunch of work with the file system that we could cache instead.  This now lets us cache:
- file existence
- modified time
- file size
- etag

This also fixes thumb nails to have the same extension fix that attachments have.

### DIFF
--- a/Sources/ShowAttachments.php
+++ b/Sources/ShowAttachments.php
@@ -120,8 +120,11 @@ function showAttachment()
 		// ensure variant attachment compatibility
 		$filePath = pathinfo($file['filePath']);
 
-		$file['filePath'] = !file_exists($file['filePath']) && isset($filePath['extension']) ? substr($file['filePath'], 0, -(strlen($filePath['extension']) + 1)) : $file['filePath'];
-		$file['etag'] = '"' . md5_file($file['filePath']) . '"';
+		$file['exists'] = file_exists($file['filePath']);
+		$file['filePath'] = !$file['exists'] && isset($filePath['extension']) ? substr($file['filePath'], 0, -(strlen($filePath['extension']) + 1)) : $file['filePath'];
+		$file['mtime'] = $file['exists'] ? filemtime($file['filePath']) : 0;
+		$file['size'] = $file['exists'] ? filesize($file['filePath']) : 0;
+		$file['etag'] = '"' . sha1(md5_file($file['filePath']) . '|' . $file['mtime']) . '"';
 
 		// now get the thumbfile!
 		$thumbFile = array();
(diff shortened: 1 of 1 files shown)


Sources/ShowAttachments.php
```diff
@@ -120,8 +120,11 @@ function showAttachment()
 		// ensure variant attachment compatibility
 		$filePath = pathinfo($file['filePath']);
 
-		$file['filePath'] = !file_exists($file['filePath']) && isset($filePath['extension']) ? substr($file['filePath'], 0, -(strlen($filePath['extension']) + 1)) : $file['filePath'];
-		$file['etag'] = '"' . md5_file($file['filePath']) . '"';
+		$file['exists'] = file_exists($file['filePath']);
+		$file['filePath'] = !$file['exists'] && isset($filePath['extension']) ? substr($file['filePath'], 0, -(strlen($filePath['extension']) + 1)) : $file['filePath'];
+		$file['mtime'] = $file['exists'] ? filemtime($file['filePath']) : 0;
+		$file['size'] = $file['exists'] ? filesize($file['filePath']) : 0;
+		$file['etag'] = '"' . sha1(md5_file($file['filePath']) . '|' . $file['mtime']) . '"';
```
Sesquipedalian: There's no need to include the modification time here if we are already using `md5_file()` to get a checksum of the file contents. In fact, it is probably better not to include it, because the eTag should only change if the actual content of the file has changed. For example, if the file has merely been `touch`-ed by the file system, the eTag should still match.

There's also no point using `sha1()` to re-hash the value. If you would prefer to use `sha1_file()` to get the file checksum, rather than `md5_file()`, that'd be fine, but I don't think it makes much difference for our purposes here.

Sources/ShowAttachments.php
```diff
@@ -143,9 +146,15 @@ function showAttachment()
...
 				$thumbFile['filePath'] = getAttachmentFilename($thumbFile['filename'], $thumbFile['id_attach'], $thumbFile['id_folder'], false, $thumbFile['file_hash']);
-				$thumbFile['etag'] = '"' . md5_file($thumbFile['filePath']) . '"';
+				$thumbPath = pathinfo($thumbFile['filePath']);
+
+				// set filePath and ETag time
+				$thumbFile['exists'] = file_exists($thumbFile['filePath']);
+				$file['filePath'] = !$thumbFile['exists'] && isset($thumbPath['extension']) ? substr($thumbFile['filePath'], 0, -(strlen($thumbPath['extension']) + 1)) : $thumbFile['filePath'];
+				$thumbFile['mtime'] = $thumbFile['exists'] ? filemtime($thumbFile['filePath']) : 0;
+				$thumbFile['size'] = $thumbFile['exists'] ? filesize($thumbFile['filePath']) : 0;
+				$thumbFile['etag'] = '"' . sha1(md5_file($thumbFile['filePath']) . '|' . $thumbFile['mtime']) . '"';
```
Sesquipedalian: See previous comment about `md5_file()`

Sesquipedalian (changes requested):

Sources/ShowAttachments.php
jdarwood007: Done.  I rather us use sha1_file as its 40 characters versus 32 of md5_file.  Anything higher than that isn't worth the cpu cycles for a etag that simply determines modified information on a file.

Sesquipedalian (approved):

PR #7400, opened 2022-03-29
Workaround for Content-Encoding header bug in PHP 8.0.17 & 8.1.4

### LINKED ISSUE #7395: Compressed output in PHP 8.0.17 or 8.1.4 breaks attachments
#### Description
Not sure if there's anything we can do about this, but tracking this here so it doesn't get lost. PHP 8.0.17 and 8.1.4 include a fix for a zlib bug, but the fix breaks attachments when compressed output is enabled. See https://github.com/php/php-src/issues/8218 for more and https://github.com/php/php-src/issues/7953 for the original issue that was patched.

### Steps to reproduce
1. Install SMF on a server with PHP 8.0.17 or 8.1.4 and enable compressed output.
2. Attach an image to a post.

- PHP Version: 8.0.17 or 8.1.4

Sesquipedalian:
Fixes #7395 

Make sure to test on PHP 7.4.x, 8.0.x, and 8.1.x, with the "Enable compressed output" setting enabled and again with it disabled.

### DIFF
--- a/Sources/Profile-Export.php
+++ b/Sources/Profile-Export.php
@@ -493,6 +493,7 @@ function download_export_file($uid)
 		if (strtotime($modified_since) >= $mtime)
 		{
 			ob_end_clean();
+			header_remove('content-encoding');
 
 			// Answer the question - no, it hasn't been modified ;).
 			send_http_status(304);
@@ -505,6 +506,7 @@ function download_export_file($uid)
 	if (!empty($_SERVER['HTTP_IF_NONE_MATCH']) && strpos($_SERVER['HTTP_IF_NONE_MATCH'], $eTag) !== false)
 	{
 		ob_end_clean();
+		header_remove('content-encoding');
 
 		send_http_status(304);
 		exit;
@@ -571,6 +573,8 @@ function download_export_file($uid)
 		while (@ob_get_level() > 0)
 			@ob_end_clean();
 
+		header_remove('content-encoding');
+
 		// 40 kilobytes is a good-ish amount
 		$chunksize = 40 * 1024;
 		$bytes_sent = 0;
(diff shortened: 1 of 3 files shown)


Sources/Profile-Export.php
```diff
@@ -569,7 +571,10 @@ function ($datatype) use ($txt)
 	if (isset($_SERVER['HTTP_RANGE']))
 	{
 		while (@ob_get_level() > 0)
+		{
 			@ob_end_clean();
+			header_remove('content-encoding');
```
jdarwood007: You shouldn't need to do this inside the loop.

jdarwood007:
@Sesquipedalian Did you test uploading files?  I was able to get my dev env updated to 8.0.17 and yea it broke.  I disabled some custom code and got it working again.  But even after testing this PR.  I can't get attachments to upload.

jdarwood007:
Got it working.
There needs to be another call to this on line 40 in ShowAttachments.php as well
```
	// This is done to clear any output that was made before now.
	ob_end_clean();
```

albertlast:
would it be not smarter to give the admin a hint that his php version is broken instead of working around?

Oldiesmann:
> would it be not smarter to give the admin a hint that his php version is broken instead of working around?

Won't help if the admin doesn't have the ability to update PHP. Also, 8.0.17 and 8.1.4 are the most recent versions of PHP so there's not much someone with server admin access can do right now anyway except revert to an earlier version of PHP.

Sources/Profile-Export.php
Sesquipedalian: Fixed

jdarwood007 (approved):

PR #7411, opened 2022-04-05
Prevents moving topics to redirect boards via quick moderation

### LINKED ISSUE #7403: Quick moderation allows moving topics to redirection boards
#### Description
Arantor reported it on the Bug Reports board:
https://www.simplemachines.org/community/index.php?topic=581608.0

### Steps to reproduce
1. Make a normal board.
2. Make a redirection board.
3. Make sure quick moderation (as checkboxes) is enabled in your profile.
4. Make a topic in the normal board.
5. While in the normal board, use checkbox quick moderation to select the topic, select move to, select the redirection board.
6. Press go.

Result:
move redirection

Affects 2.1.1, probably affects 2.0 as well but not tested. 

> I haven't checked but I suspect there's also a two-fer here - in the ACP there is the option to disallow people moving topics to boards where they aren't otherwise able to make new topics. I don't *know* (because haven't tried) whether the restriction is enforced on quick moderation.
> 
> But it's possible it's not.

Sesquipedalian:
Fixes #7403.

Also fixes an unreported bug where the jump_to select menus could be populated multiple times if more than one jump_to appeared on the same page.

### DIFF
--- a/Sources/MessageIndex.php
+++ b/Sources/MessageIndex.php
@@ -947,6 +947,23 @@ function QuickModeration()
 			if (empty($moveCache[1][$topic]))
 				continue;
 
+			// Never move topics to redirect boards
+			$redirect_boards = array();
+			$request = $smcFunc['db_query']('', '
+				SELECT id_board
+				FROM {db_prefix}boards
+				WHERE redirect != {string:blank_redirect}',
+				array(
+					'blank_redirect' => '',
+				)
+			);
+			while ($row = $smcFunc['db_fetch_row']($request))
+				$redirect_boards[] = $row[0];
+			$smcFunc['db_free_result']($request);
+
+			if (in_array($moveCache[1][$topic], $redirect_boards))
+				continue;
+
 			$moveCache[0][] = $topic;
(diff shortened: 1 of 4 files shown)


Oldiesmann:
One minor issue. In the XML, you have a parameter called "is_redirect", but in the JavaScript you're looking for "isRedirect" instead.

Sesquipedalian:
> One minor issue. In the XML, you have a parameter called "is_redirect", but in the JavaScript you're looking for "isRedirect" instead.

I don't believe this is actually an issue. Perhaps you overlooked [this line](https://github.com/SimpleMachines/SMF2.1/pull/7411/files#diff-45b3e1c6309c9b55fad6bbdb0fbfa63c40f21b737f7c34b6100f9ac58f3b1086R1112)?

sbulen:
I will get to these this weekend.  Sorry for the delay!

Sesquipedalian:
Great! Thanks, @sbulen.

jdarwood007 (approved):
Tested and does what it says it does.

Issue #7208, opened 2021-12-29
Slow query on memberlist


jdarwood007:
#### Description

```
SELECT mem.id_member
FROM smf_members AS mem
   
WHERE mem.is_activated = 1
ORDER BY mem.real_name ASC
LIMIT 0, 30
```

id | select_type | table | partitions | type | possible_keys | key | key_len | ref | rows | filtered | Extra
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
1 | SIMPLE | mem |   | ref | is_activated | is_activated | 1 | const | 287230 | 100.00 | Using index condition; Using filesort

### Steps to reproduce
1. 
2. 

### Environment (complete as necessary)
- Version/Git revision: N/A
- Database Type: Mysql
- Database Version: 5.7
- PHP Version: 7.4

### Additional information/references
https://www.simplemachines.org/community/index.php?topic=580082

Attaching this to 2.1.1, as I'm not sure if we can fix it for 2.1.0.  But we may want to fix for 2.1.0, as I suspect it may be a new index issue.

@albertlast See what you can do?

albertlast:
I did look in my test env (pg),
look like that i create a test index very long time ago,
where the colum (is_activated, real_name) as own idx exists for the table smf_members.

So you could try to add this and see what mysql is doing with this.

jdarwood007:
That seems to fix it.

The other sorts such as position, website also do this.

Sorting by membergroup
```
      SELECT mem.id_member
      FROM smf_members AS mem
         LEFT JOIN smf_membergroups AS mg ON (mg.id_group = CASE WHEN mem.id_group = 0 THEN mem.id_post_group ELSE mem.id_group END)
         
      WHERE mem.is_activated = 1
      ORDER BY mg.group_name is not null, mg.group_name ASC
      LIMIT 0, 30
   in /Sources/Memberlist.php line 353, which took 13.64280391 seconds at 0.25429106 into request.
```
id | select_type | table | partitions | type | possible_keys | key | key_len | ref | rows | filtered | Extra
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
1 | SIMPLE | mem |   | ref | is_activated,custom_memberlist_sort | is_activated | 1 | const | 287243 | 100.00 | Using temporary; Using filesort
1 | SIMPLE | mg |   | eq_ref | PRIMARY | PRIMARY | 2 | func | 1 | 100.00 | Using where


Sorting by website
[...]

PR #7415, opened 2022-04-05
Disable memberlist url sort for guests

albertlast:
disable url sort for guest mention in [#7208](https://github.com/SimpleMachines/SMF2.1/issues/7208#issuecomment-1008207268)

### DIFF
--- a/Sources/Memberlist.php
+++ b/Sources/Memberlist.php
@@ -79,7 +79,7 @@ function Memberlist()
 			'label' => $txt['website'],
 			'link_with' => 'website',
 			'sort' => array(
-				'down' =>  $context['user']['is_guest'] ? '1=1' : 'mem.website_url = \'\', mem.website_url is null, mem.website_url DESC',
+				'down' => $context['user']['is_guest'] ? '1=1' : 'mem.website_url = \'\', mem.website_url is null, mem.website_url DESC',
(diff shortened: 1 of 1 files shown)


Sources/Memberlist.php
```diff
@@ -79,8 +79,8 @@ function Memberlist()
 			'label' => $txt['website'],
 			'link_with' => 'website',
 			'sort' => array(
-				'down' => 'mem.website_url = \'\', mem.website_url is null, mem.website_url DESC',
-				'up' => 'mem.website_url != \'\', mem.website_url is not null, mem.website_url ASC'
+				'down' =>  $context['user']['is_guest'] ? '1=1' : 'mem.website_url = \'\', mem.website_url is null, mem.website_url DESC',
```
Sesquipedalian: Nit: Unnecessary extra space between `=>` and `$context`

Sesquipedalian (changes requested):
This works well, but please also add the following code to the `MLAll()` function, immediately after the `// Sort out the column information.` loop:
```
	// Don't offer website sort to guests
	if ($context['user']['is_guest'])
	{
		$context['columns']['website_url']['href'] = '';
		$context['columns']['website_url']['link'] = $context['columns']['website_url']['label'];
	}
```

This addition will prevent the "Website" column from appearing as a link to guests, which will help avoid confusion.

Sesquipedalian (approved):

Issue #7421, opened 2022-04-07
pacman, pg & mysql:  Adding a NOT NULL column to an existing table...


sbulen:
#### Description
Problem 1:  Pg & mysql behave differently with NOT NULL and defaults on text fields.  Mysql will NOT allow you to specify a default on a TEXT column.  pg will allow it.  So...  Right there it is possible to create a mod that will install on one but not the other.  

Problem 2:  If you try to create a column with NOT NULL and NO default, _on a table that has existing records,_ pg will complain, because it feels it must give a value to the column for the existing rows, and you haven't specified one.  

So, even if you tried to be consistent, and specified NOT NULL with no default, it wouldn't work.  

Note we have had a similar issue with the installer/upgrader earlier.  Pacman faces the same issue:
https://github.com/SimpleMachines/SMF2.1/pull/7376

Possible solutions? Thinking aloud here...
 - Give an error when you see a default on a text column, forcing pg & mysql to have the same definition
 - If rows exist, in pg, the update must be done in 2 steps, as in the PR above

### Environment (complete as necessary)
- Version/Git revision:  Current
- Database Type: pg & mysql

### Additional information/references
**_NOTE that if submitting a PR, we will want to see proof of testing on pg & mysql. Before & after screenshots._**

jdarwood007:
I'm not sure of a solution here.  The first problem is a issue with mysql.  However this stackoverflow says 8.0.13 allows it in a weird way: https://stackoverflow.com/questions/3466872/why-cant-a-text-column-have-a-default-value-in-mysql

I would go on saying that we detect if we are trying to add a default to a text column and not allow it for mysql.  But that may not work since it may expect data if its a not null column.  There is no win here.

The second problem, I am guessing for new tables, it just wants that column to have a value then on insert?  Again, this has a no win scenario for us.  We would need a default value to give it a valid setup.  Maybe we just force it for both mysql and postgresql that you can not have a default of none and a not null when you do an add column?  I think Postgresql code does a slow standup if you look at its add_column function and this would have to be stopped early.

sbulen:
In order to have the table structures consistent across pg & mysql, I think we have to do 2 things:
 - Support the lowest common denominator...  Do not allow defaults on text columns.  OR, maybe only allow '', since that is mysql's implicit default...   & then ignore it for mysql.
[...]

PR #7424, opened 2022-04-09
Fixes various issues with DbPackages functions

### LINKED ISSUE #7417: pacman & pg: indexes not removed when mods uninstalled
#### Description
If you install a mod on a pg forum that creates indexes on existing tables, all is OK.  But when you uninstall that mod, the indexes remain.  

Columns and tables are removed properly, but indexes are not.  

### Steps to reproduce
1. Create a pg forum
2. Install a mod that creates indexes on existing tables
3. Uninstall the mod; the indexes remain

- Database Type: pg
- Database Version: 14
- PHP Version: 8.0.13

### LINKED ISSUE #7418: smf_db_list_columns defaults and nulls
I've been working more with SMF's smf_db_list_columns and smf_db_change_column functions and have come to the conclusion we don't have it working right yet for defaults.  I'm starting this to allow a discussion and present the bug.

To start, PHP has a horrible handling for NULLs.  take for example this:
```
<?php
$a = null;

var_dump(isset($a));
var_dump(is_null($a));
var_dump(isset($b));
var_dump(is_null($b));
```

PHP will report $a is not set but is null.  However for $b, PHP will report $b is not set, but throw a variable undefined error on is_null, then return true.  This is documented in PHP's manual,  but I find it a very horrible syntax that is_null will throw a variable undefined error.   If the variable your checking is in an array, you can use array_key_exists as this works as expected.

Now onto MySQL at least.  The default column has some oddities that require checking the null column.
1. If default is "NULL" and the nullable column is set to "NOT NULL", then the field is required to have a value in it and can not contain null.  In PMA this shows up as "NONE"
2. If the default is "NULL" and the nullable column is "NULL", then the column can contain nulls.
3. The value can either be a int or a string as well.  Thanks to PHP's loose type handling, this is only a problem if you want to do a strict type check.  But it makes it more difficult for us to do checks against because if we do strict type checks, it can fail between the '0' === 0.

This creates issues when you are checking and comparing columns.  I don't think SMF's code is up to snuff on getting this all right.  I've wrote some code, but I dislike the changes.  This is the sample of the changes for list_columns
```
				$default = null;
				// Column is nullalbe and the default is NULL.
				if (is_null($row['Default']) && $row['Null'] == 'YES')
					$default = null;
				// Column is not nullable and the default is set to null, meaning no default.
				elseif (is_null($row['Default']) && $row['Null'] == 'NO')
					$default = false;
				elseif (isset($row['Default']) && $row['Default'] == '\'\'')
					$default = '';
				elseif (isset($row['Default']))
					$default = stripslashes($row['Default']);
```

However the code changes for change_columns is:
```
		if (!array_key_exists('default', $column_info))
			$default = $old_info['default'];
		elseif ($old_info['default'] === false)
			$column_info['default'] = null;

[...]

		// Fix the default.
		$default = '';
		if (is_null($column_info['default']) && empty($column_info['not_null']))
			$default = 'NULL';
		elseif (isset($column_info['default']))
			$default = '\'' . $smcFunc['db_escape_string']($column_info['default'], self::$db_connection) . '\'';
```

It is fairly ugly, but it does the work.  Anyone else have a better opinion on trying to get this right?

### LINKED ISSUE #7419: pacman, pg & mysql: TEXT + indexes
#### Description

If a mod needs a TEXT field with an index...  Mysql forces you to specify an index length to create a partial index.  But pg chokes when an index length is specified in pacman.  

I may be missing something...  But, it appears to me that, ATM, it's impossible for a MOD to create a TEXT field with an index that will install on both pg and mysql.  

Possible fix:  Note that where we have partial indexes on text fields in SMF today, the installer & upgrader treat them differently in pg vs mysql.  PG never specifies the length.  It may be sufficient for pacman to drop the length in pg, just like it does in the installer/upgrader.

- Database Type: pg & mysql

Note that VARCHAR(1024), + a full index works for mysql 8 & pg.  
But...  
Mysql 5.6 wouldn't let you do that due to (the index's) index length.  (This restriction was addressed in mysql 5.7.)  In fact, mysql 5.6 doesn't let you do VARCHAR(512) either.

VARCHAR(255) + a full index is the only approach that appears to work across mysql 5.6, mysql 8, & pg where you need a text column with an index today.

**_NOTE that if submitting a PR, we will want to see proof of testing on pg & mysql.  Before & after screenshots._**

### LINKED ISSUE #7420: pacman: Legacy NULL setting doesn't work
#### Description
When the NOT_NULL setting was added to pacman, the prior NULL setting was kept around for legacy support.

It doesn't appear to be working.  We should either make it work or remove it.

- Database Type: mysql (didn't check pg)

**_NOTE that if submitting a PR, we will want to see proof of testing on pg & mysql. Before & after screenshots._**

jdarwood007:
1. We sometimes will pass in `databasename`.`table_name` and have not uses a select database.  So we try to parse our the database info when running select commands that need this such as list tables, list keys, list columns.
2. When working with default, we need to handle nulls properly as a default using the literal NULL.  We should also try to handle numericals better
3. MySQL doesn't allow you to do an index on a text column without a length.  Force one if found.
4. Postgresql doesn't support a length on an index, so strip it out if specified.
5. Postgresql didn't handle indexes well in all scenarios because they are prefixed with the table name and we may not pass the index name prefixed with the table name.
6. Postgresql didn't do NULLs properly because we didn't check the right value returned from the list of columns.
7. Avoid overwriting $table_name, as this can cause unexpected results when the parsed name is sent to other functions.
8. MySQL with text index needs a limit, so we try to add one.

Attached are simple scripts I used to test these.  Please modify the scripts and send them back if you are finding cases where we are still not doing something right.

[test-7417.php.txt](https://github.com/SimpleMachines/SMF2.1/files/8456163/test-7417.php.txt)
[test-7420.php.txt](https://github.com/SimpleMachines/SMF2.1/files/8456164/test-7420.php.txt)
[test-7419.php.txt](https://github.com/SimpleMachines/SMF2.1/files/8456165/test-7419.php.txt)
[test-7418.php.txt](https://github.com/SimpleMachines/SMF2.1/files/8456166/test-7418.php.txt)

Fixes #7417 
Fixes #7418 
Fixes #7419 
Fixes #7420 

I'm doing this as a WIP right now and asking @albertlast and @sbulen to provide feedback.  But once we are happy there, this will want more dev eyes on it.  Its a lot of changes to make.

### DIFF
--- a/Sources/DbPackages-mysql.php
+++ b/Sources/DbPackages-mysql.php
@@ -325,6 +325,7 @@ function smf_db_add_column($table_name, $column_info, $parameters = array(), $if
 	global $smcFunc, $db_package_log, $db_prefix;
 
 	$short_table_name = str_replace('{db_prefix}', $db_prefix, $table_name);
+	$column_info = array_change_key_case($column_info);
 
 	// Log that we will want to uninstall this!
 	$db_package_log[] = array('remove_column', $short_table_name, $column_info['name']);
@@ -407,6 +408,7 @@ function smf_db_change_column($table_name, $old_column, $column_info)
 	global $smcFunc, $db_prefix;
 
 	$short_table_name = str_replace('{db_prefix}', $db_prefix, $table_name);
+	$column_info = array_change_key_case($column_info);
 
 	// Check it does exist!
 	$columns = $smcFunc['db_list_columns']($table_name, true);
@@ -850,6 +852,8 @@ function smf_db_list_indexes($table_name, $detail = false, $parameters = array()
 function smf_db_create_query_column($column)
 {
 	global $smcFunc;
+
+	$column = array_change_key_case($column);
 
 	// Auto increment is easy here!
 	$default = '';
--- a/Sources/DbPackages-mysql.php
+++ b/Sources/DbPackages-mysql.php
@@ -113,20 +113,23 @@ function smf_db_create_table($table_name, $columns, $indexes = array(), $paramet
 
 	// Strip out the table name, we might not need it in some cases
 	$real_prefix = preg_match('~^(`?)(.+?)\\1\\.(.*?)$~', $db_prefix, $match) === 1 ? $match[3] : $db_prefix;
+	$database = !empty($match[2]) ? $match[2] : $db_name;
 
 	// With or without the database name, the fullname looks like this.
 	$full_table_name = str_replace('{db_prefix}', $real_prefix, $table_name);
-	$table_name = str_replace('{db_prefix}', $db_prefix, $table_name);
+	// Do not overwrite $table_name, this causes issues if we pass it onto a helper function.
+	$short_table_name = str_replace('{db_prefix}', $db_prefix, $table_name);
 
 	// First - no way do we touch SMF tables.
-	if (in_array(strtolower($table_name), $reservedTables))
+	if (in_array(strtolower($short_table_name), $reservedTables))
 		return false;
 
 	// Log that we'll want to remove this on uninstall.
-	$db_package_log[] = array('remove_table', $table_name);
+	$db_package_log[] = array('remove_table', $short_table_name);
 
 	// Slightly easier on MySQL than the others...
-	$tables = $smcFunc['db_list_tables']();
+	$tables = $smcFunc['db_list_tables']($database);
+
 	if (in_array($full_table_name, $tables))
 	{
 		// This is a sad day... drop the table? If not, return false (error) by default.
@@ -136,9 +139,9 @@ function smf_db_create_table($table_name, $columns, $indexes = array(), $paramet
 		{
 			$smcFunc['db_transaction']('begin');
 			$db_trans = true;
-			$smcFunc['db_drop_table']($table_name . '_old');
+			$smcFunc['db_drop_table']($short_table_name . '_old');
 			$smcFunc['db_query']('', '
-				RENAME TABLE ' . $table_name . ' TO ' . $table_name . '_old',
+				RENAME TABLE ' . $short_table_name . ' TO ' . $short_table_name . '_old',
 				array(
 					'security_override' => true,
 				)
@@ -150,7 +153,7 @@ function smf_db_create_table($table_name, $columns, $indexes = array(), $paramet
 	}
 
 	// Righty - let's do the damn thing!
-	$table_query = 'CREATE TABLE ' . $table_name . "\n" . '(';
+	$table_query = 'CREATE TABLE ' . $short_table_name . "\n" . '(';
 	foreach ($columns as $column)
 		$table_query .= "\n\t" . smf_db_create_query_column($column) . ',';
 
@@ -219,8 +222,8 @@ function smf_db_create_table($table_name, $columns, $indexes = array(), $paramet
 			GROUP BY column_name
 			HAVING count(*) > 1',
 			array(
-				'table1' => $table_name,
-				'table2' => $table_name . '_old',
+				'table1' => $short_table_name,
+				'table2' => $short_table_name . '_old',
 				'schema' => $db_name,
 			)
 		);
@@ -231,15 +234,15 @@ function smf_db_create_table($table_name, $columns, $indexes = array(), $paramet
 		}
 
 		$smcFunc['db_query']('', '
-			INSERT INTO ' . $table_name . '('
+			INSERT INTO ' . $short_table_name . '('
 			. implode(',', $same_col) .
 			')
 			SELECT ' . implode(',', $same_col) . '
-			FROM ' . $table_name . '_old',
+			FROM ' . $short_table_name . '_old',
 			array()
 		);
 
-		$smcFunc['db_drop_table']($table_name . '_old');
+		$smcFunc['db_drop_table']($short_table_name . '_old');
 	}
 
 	return true;
@@ -255,23 +258,26 @@ function smf_db_create_table($table_name, $columns, $indexes = array(), $paramet
  */
 function smf_db_drop_table($table_name, $parameters = array(), $error = 'fatal')
 {
-	global $reservedTables, $smcFunc, $db_prefix;
+	global $reservedTables, $smcFunc, $db_prefix, $db_name;
 
 	// After stripping away the database name, this is what's left.
 	$real_prefix = preg_match('~^(`?)(.+?)\\1\\.(.*?)$~', $db_prefix, $match) === 1 ? $match[3] : $db_prefix;
+	$database = !empty($match[2]) ? $match[2] : $db_name;
 
 	// Get some aliases.
 	$full_table_name = str_replace('{db_prefix}', $real_prefix, $table_name);
-	$table_name = str_replace('{db_prefix}', $db_prefix, $table_name);
+	// Do not overwrite $table_name, this causes issues if we pass it onto a helper function.
+	$short_table_name = str_replace('{db_prefix}', $db_prefix, $table_name);
 
 	// God no - dropping one of these = bad.
-	if (in_array(strtolower($table_name), $reservedTables))
+	if (in_array(strtolower($short_table_name), $reservedTables))
 		return false;
 
 	// Does it exist?
-	if (in_array($full_table_name, $smcFunc['db_list_tables']()))
+	$tables = $smcFunc['db_list_tables']($database);
+	if (in_array($full_table_name, $tables))
 	{
-		$query = 'DROP TABLE ' . $table_name;
+		$query = 'DROP TABLE ' . $short_table_name;
 		$smcFunc['db_query']('',
 			$query,
 			array(
@@ -300,10 +306,10 @@ function smf_db_add_column($table_name, $column_info, $parameters = array(), $if
 {
 	global $smcFunc, $db_package_log, $db_prefix;
 
-	$table_name = str_replace('{db_prefix}', $db_prefix, $table_name);
+	$short_table_name = str_replace('{db_prefix}', $db_prefix, $table_name);
 
 	// Log that we will want to uninstall this!
-	$db_package_log[] = array('remove_column', $table_name, $column_info['name']);
+	$db_package_log[] = array('remove_column', $short_table_name, $column_info['name']);
 
 	// Does it exist - if so don't add it again!
 	$columns = $smcFunc['db_list_columns']($table_name, false);
@@ -322,7 +328,7 @@ function smf_db_add_column($table_name, $column_info, $parameters = array(), $if
 
 	// Now add the thing!
 	$query = '
-		ALTER TABLE ' . $table_name . '
+		ALTER TABLE ' . $short_table_name . '
 		ADD ' . smf_db_create_query_column($column_info) . (empty($column_info['auto']) ? '' : ' primary key'
 	);
 	$smcFunc['db_query']('', $query,
@@ -347,15 +353,16 @@ function smf_db_remove_column($table_name, $column_name, $parameters = array(), 
 {
 	global $smcFunc, $db_prefix;
 
-	$table_name = str_replace('{db_prefix}', $db_prefix, $table_name);
+	$short_table_name = str_replace('{db_prefix}', $db_prefix, $table_name);
 
 	// Does it exist?
 	$columns = $smcFunc['db_list_columns']($table_name, true);
+
 	foreach ($columns as $column)
 		if ($column['name'] == $column_name)
 		{
 			$smcFunc['db_query']('', '
-				ALTER TABLE ' . $table_name . '
+				ALTER TABLE ' . $short_table_name . '
 				DROP COLUMN ' . $column_name,
 				array(
 					'security_override' => true,
@@ -381,7 +388,7 @@ function smf_db_change_column($table_name, $old_column, $column_info)
 {
 	global $smcFunc, $db_prefix;
 
-	$table_name = str_replace('{db_prefix}', $db_prefix, $table_name);
+	$short_table_name = str_replace('{db_prefix}', $db_prefix, $table_name);
 
 	// Check it does exist!
 	$columns = $smcFunc['db_list_columns']($table_name, true);
@@ -399,6 +406,7 @@ function smf_db_change_column($table_name, $old_column, $column_info)
 		$column_info['not_null'] = !$column_info['null'];
 	if (isset($old_info['null']))
 		$old_info['not_null'] = !$old_info['null'];
+
 	// Get the right bits.
 	if (!isset($column_info['name']))
 		$column_info['name'] = $old_column;
@@ -422,8 +430,10 @@ function smf_db_change_column($table_name, $old_column, $column_info)
 
 	// Fix the default.
 	$default = '';
-	if (isset($column_info['default']) && is_null($column_info['default']) && empty($column_info['not_null']))
+	if (array_key_exists('default', $column_info) && is_null($column_info['default']))
 		$default = 'NULL';
+	elseif (isset($column_info['default']) && is_numeric($column_info['default']))
+		$default = 'DEFAULT ' . (strpos($column_info['default'], '.') ? floatval($column_info['default']) : intval($column_info['default']));
 	else
 		$default = '\'' . $smcFunc['db_escape_string']($column_info['default']) . '\'';
 
@@ -431,7 +441,7 @@ function smf_db_change_column($table_name, $old_column, $column_info)
 		$type = $type . '(' . $size . ')';
 
 	$smcFunc['db_query']('', '
-		ALTER TABLE ' . $table_name . '
+		ALTER TABLE ' . $short_table_name . '
 		CHANGE COLUMN `' . $old_column . '` `' . $column_info['name'] . '` ' . $type . ' ' . (!empty($unsigned) ? $unsigned : '') . (!empty($column_info['not_null']) ? 'NOT NULL' : '') . ' ' .
 				($default === '' ? '' : 'default ' . $default) . ' ' .
 			(empty($column_info['auto']) ? '' : 'auto_increment') . ' ',
@@ -455,11 +465,30 @@ function smf_db_add_index($table_name, $index_info, $parameters = array(), $if_e
 {
 	global $smcFunc, $db_package_log, $db_prefix;
 
-	$table_name = str_replace('{db_prefix}', $db_prefix, $table_name);
+	$short_table_name = str_replace('{db_prefix}', $db_prefix, $table_name);
 
 	// No columns = no index.
 	if (empty($index_info['columns']))
 		return false;
+		
+	// MySQL If its a text column, we need to add a size.
+	$cols = $smcFunc['db_list_columns']($table_name, true);
+	$version = $smcFunc['db_get_version']();
+	foreach ($index_info['columns'] as &$c)
+	{
+		$c = trim($c);
+
+		// If a size was already specified, we won't be able to match it anyways.
+		if (!isset($cols[$c]) || !in_array($cols[$c]['type'], array('text', 'mediumntext', 'largetext')))
+			continue;
+
+		// This is a column we need a size on and we are below 5.7 we have to stick to a smaller size.
+		if (version_compare($version, '5.7', '<'))
+			$c .= '(64)';
+		else
+			$c .= '(255)';
+	}
+
 	$columns = implode(',', $index_info['columns']);
 
 	// No name - make it up!
@@ -469,11 +498,11 @@ function smf_db_add_index($table_name, $index_info, $parameters = array(), $if_e
 		if (isset($index_info['type']) && $index_info['type'] == 'primary')
 			$index_info['name'] = '';
 		else
-			$index_info['name'] = implode('_', $index_info['columns']);
+			$index_info['name'] = trim(implode('_', preg_replace('~(\(\d+\))~', '', $index_info['columns'])));
 	}
 
 	// Log that we are going to want to remove this!
-	$db_package_log[] = array('remove_index', $table_name, $index_info['name']);
+	$db_package_log[] = array('remove_index', $short_table_name, $index_info['name']);
 
 	// Let's get all our indexes.
 	$indexes = $smcFunc['db_list_indexes']($table_name, true);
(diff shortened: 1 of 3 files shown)


jdarwood007:
FYI,
MySQL 8.0 on Ubuntu
PostgreSQL: PostgreSQL 12.9 (Ubuntu 12.9-0ubuntu0.20.04.1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0, 64-bit

Sources/DbPackages-postgresql.php
```diff
@@ -150,28 +151,28 @@ function smf_db_create_table($table_name, $columns, $indexes = array(), $paramet
...
+					DROP SEQUENCE IF EXISTS ' . $short_table_name . '_seq',
 					array(
 						'security_override' => true,
 					)
 				);
 
 			if (!$old_table_exists)
 				$smcFunc['db_query']('', '
-					CREATE SEQUENCE ' . $table_name . '_seq',
+					CREATE SEQUENCE ' . $short_table_name . '_seq',
```
albertlast, line 172: i still would like to see that sequence with owned get created here,
like done in this pr: https://github.com/SimpleMachines/SMF2.1/pull/6323

albertlast:
It's always confusing that the word database is used everywhere,
but not meant to be.
by simple reason mysql didn't support database and you can't "change" database in the way like here done in the code use in pg.

Sources/DbPackages-postgresql.php
```diff
@@ -104,27 +104,28 @@ function db_packages_init()
...
 	$old_table_exists = false;
 
 	// Strip out the table name, we might not need it in some cases
 	$real_prefix = preg_match('~^("?)(.+?)\\1\\.(.*?)$~', $db_prefix, $match) === 1 ? $match[3] : $db_prefix;
+	$database = !empty($match[2]) ? $match[2] : $db_name;
 
 	// With or without the database name, the fullname looks like this.
 	$full_table_name = str_replace('{db_prefix}', $real_prefix, $table_name);
-	$table_name = str_replace('{db_prefix}', $db_prefix, $table_name);
+	$short_table_name = str_replace('{db_prefix}', $db_prefix, $table_name);
```
albertlast, line 118: why is ths not real_table_name like in mysql?

Sources/DbPackages-postgresql.php
```diff
@@ -581,21 +588,21 @@ function smf_db_change_column($table_name, $old_column, $column_info)
...
-				DROP SEQUENCE IF EXISTS ' . $table_name . '_seq',
+				DROP SEQUENCE IF EXISTS ' . $short_table_name . '_seq',
 				array(
 					'security_override' => true,
 				)
 			);
 
 			$smcFunc['db_query']('', '
-				CREATE SEQUENCE ' . $table_name . '_seq',
+				CREATE SEQUENCE ' . $short_table_name . '_seq',
```
albertlast, line 607: like to owned like mention before

albertlast:
run the test script, checked the the log file of pg in data directory and found no error messages,
so test run clean throw pg.

Sources/DbPackages-postgresql.php
jdarwood007: Since that PR wasn't merged.  I don't think this PR should address it.  Even if it is right.

jdarwood007:
> It's always confusing that the word database is used everywhere, but not meant to be. by simple reason mysql didn't support database and you can't "change" database in the way like here done in the code use in pg.

Any suggestions?

jdarwood007:
For your other question on real_table_name.  The first fixes I worked on changed everything to use short_table_name because modifying $table_name was sometimes causing issues when it got passed into other functions.  This was typically with mysql, but I pushed it into the postgresql to be consistent.  The second issue then I found was that in MySQL, when running in SSI, we do not issue a select database command, but run with `database`.`table_name` on all commands.  This caused issue where we expected things like getting a list of tables in a database, but we never selected a database, so the code would use the database configured in SMF.  Which worked until you issued a command in another database.  So for some functions, I have to parse out any possible database match and try to use that, so we issue the commands to the correct database.  So since not all functions needed this fix, I didn't have to do it to all.  But some things like mysql still needed more of a helping hand to get the job done right.

We ignore the schema in postgresql, so at least we don't have to deal with that.

Sources/DbPackages-postgresql.php
sbulen: I agree with @albertlast that an OWNED BY would be appropriate here.

In pg, sequences are standalone things, and must be associated with a column explicitly.  Not like in mysql with AUTO INCREMENT.  In mysql, if you delete the table, the auto increment goes away.  This doesn't happen in pg in SMF today, because we created all the sequences without specifying owners.

So what's the problem???   Assume you installed a pg-based SMF forum, & you need to start over.  So, you delete all the tables and attempt to reinstall.  That produces many errors today, because - among other things - you get duplicates on all the sequences.  Making the columns the owners of the sequences ensures they get deleted when the tables are deleted.  

**_Admittedly - not a major or high priority issue..._**  But the only error-free way to reinstall a pg smf forum today is to drop & recreate the pg db.  Deleting the tables isn't enough.

(Tangent alert:  And once you get past the sequence errors, you may encounter duplicate FUNCTION errors...)

The problem with that other PR is that the approach is highly unusual - it alters the pg database directly.  The equivalent would be to add a table to mysql, but not with a CREATE TABLE command, but by adding rows in the TABLE and COLUMNS tables in mysql's information_schema...  What that PR should really do is simply add the OWNED BYs to the install/upgrade scripts.

albertlast:
> > It's always confusing that the word database is used everywhere, but not meant to be. by simple reason mysql didn't support database and you can't "change" database in the way like here done in the code use in pg.
> 
> Any suggestions?

called schema, since it is that: https://dev.mysql.com/doc/refman/8.0/en/glossary.html#glos_schema

Sources/DbPackages-postgresql.php
jdarwood007: Is it acceptable to say then we should make that PR a prereq for adding it into this PR?  We don't support it today and that PR adds it.  But we don't want to half ask it in this PR with that.

jdarwood007:
In pg we use the public schema.  So wouldn't this get confusing?  Its MySQL in reality that is wrong here.  They should be using a schema on the database.  Which then would align it with most others.

albertlast:
in pg we didn't support schema, for reason i doesn't understand,
so smf get installed/working in the default schema,
which deppends on the database user how is used(common is public).

But the point is still,
that the <schema>.<table> and to call the first part database is wrong.

sbulen:
I have confirmed this fixes #7417 via testing on pg & mysql.

sbulen:
I was testing #7419 but it fails spectacularly on pg...

The package-info.xml of the failed pg test looked like:
```
<?php

global $smcFunc;

if (!isset($smcFunc['db_create_table']))
	db_extend('packages');

$create_tables = array(
	'a_test_record' => array(
		'columns' => array(
		    array(
			   'name' => 'id_atest',
				'type' => 'INT',
				'not_null' => true,
			),
		    array(
			   'name' => 'fails_in_pg',
				'type' => 'TEXT',
				'not_null' => true,
			),
			array(
				'name' => 'fails_in_mysql_56',
				'type' => 'varchar',
				'size' => 512,
				'default' => '',
				'not_null' => true,
			),
		),
		'indexes' => array(
			array(
				'type' => 'primary',
				'columns' => array('id_atest'),
			),
			array(
				'type' => 'index',
				'name' => 'ix_test_text',
				'columns' => array('fails_in_pg (64)'),
			),
			array(
				'type' => 'index',
				'name' => 'ix_test_ix_length',
				'columns' => array('fails_in_mysql_56'),
			),
		),
	),
);

foreach ($create_tables AS $table_name => $data)
	$smcFunc['db_create_table']('{db_prefix}' . $table_name, $data['columns'], $data['indexes']);

```

Mysql works OK.  The attempt to install the same mod in pg looks like:
image

sbulen:
I was testing #7420 , but it appears to fail on both mysql & pg.

To test, I added two columns, one with NULL = true, and one with NULL = false.

The results were such that both were nullable, i.e., the setting didn't have any effect.  This was true in both pg and mysql.

If this particular NULL setting is a legacy setting, i.e., a 2.0 setting, maybe we should just **_remove it entirely._**  That would be a satisfactory result, IMO.  The new setting, the NOT_NULL setting, seems to work fine.  Given that 2.0 mods don't work in 2.1, I'm not sure that we need to migrate support for the older NULL version???

The package-info.xml file looked like:
```
<?php

global $smcFunc;

if (!isset($smcFunc['db_create_table']))
	db_extend('packages');

$create_tables = array(
	'another_test_record' => array(
		'columns' => array(
		    array(
			   'name' => 'id_atest2',
				'type' => 'INT',
				'not_null' => true,
			),
		    array(
				'name' => 'null_false',
				'type' => 'varchar',
				'size' => 50,
				'default' => '',
				'NULL' => false,
			),
			array(
				'name' => 'null_true',
				'type' => 'varchar',
				'size' => 50,
				'default' => '',
				'NULL' => true,
			),
		),
		'indexes' => array(
			array(
				'type' => 'primary',
				'columns' => array('id_atest2'),
			),
		),
	),
);

foreach ($create_tables AS $table_name => $data)
	$smcFunc['db_create_table']('{db_prefix}' . $table_name, $data['columns'], $data['indexes']);
```

The results in mysql indicated both were nullable, i.e., the NULL setting had no effect:
image

The results in pg indicated both were nullable, i.e., the NULL setting had no effect:
image

sbulen (changes requested):
The fixes for 7419 & 7420 both appeared to have failed.  See comments.

jdarwood007:
I will get some more testing in.
I think we will need to accept
```
'default' => null,
'default' => '',
'default' => 0,
'default' => false
```

With false being the default..

sbulen:
> With false being the default..

Since mysql uses an implicit default of '' for text/character columns, I think we need to use that for consistency across pg & mysql...

I think the pg command, when no default is specified in the config to pacman, needs to look something like:
`ALTER TABLE xxxxx NOT NULL default ''; DROP DEFAULT;`

This will mimic mysql's implicit default behavior for text/char, and will work for rows with existing records (#7421 ).  

I have not experimented with ints, etc....

jdarwood007:
if we use false for the default, it shouldn't send any default.  The field would require something then, Unless the nullable is set which would allow it to be null?

albertlast:
> I was testing #7419 but it fails spectacularly on pg...
> 
> The package-info.xml of the failed pg test looked like:
> 
> ```
> <?php
> 
> global $smcFunc;
> 
> if (!isset($smcFunc['db_create_table']))
> 	db_extend('packages');
> 
> $create_tables = array(
> 	'a_test_record' => array(
> 		'columns' => array(
> 		    array(
> 			   'name' => 'id_atest',
> 				'type' => 'INT',
> 				'not_null' => true,
> 			),
> 		    array(
> 			   'name' => 'fails_in_pg',
> 				'type' => 'TEXT',
> 				'not_null' => true,
> 			),
> 			array(
> 				'name' => 'fails_in_mysql_56',
> 				'type' => 'varchar',
> 				'size' => 512,
> 				'default' => '',
> 				'not_null' => true,
> 			),
> 		),
> 		'indexes' => array(
> 			array(
> 				'type' => 'primary',
> 				'columns' => array('id_atest'),
> 			),
> 			array(
> 				'type' => 'index',
> 				'name' => 'ix_test_text',
> 				'columns' => array('fails_in_pg (64)'),
> 			),
> 			array(
> 				'type' => 'index',
> 				'name' => 'ix_test_ix_length',
> 				'columns' => array('fails_in_mysql_56'),
> 			),
> 		),
> 	),
> );
> 
> foreach ($create_tables AS $table_name => $data)
> 	$smcFunc['db_create_table']('{db_prefix}' . $table_name, $data['columns'], $data['indexes']);
> ```
> 
> Mysql works OK. The attempt to install the same mod in pg looks like: image

Maybe the paste of pg log from data directory would be also helpfull.

jdarwood007:
[test-7419b.php.txt](https://github.com/SimpleMachines/SMF2.1/files/8500740/test-7419b.php.txt)

Here is the 7419b to test against @sbulen script.  Looks like we duplicate work in the create table function.  In a future version of SMF, we should move this work to a sub function that then the create tabla and create index functions can both use to generate the correct logic.

I fixed the issue with the transaction.  We have to abort the transaction to log the error.  As it is a fatal error, it should not continue and abort execution anyways.  The error was occurring because it was trying to insert into the error log and couldn't because we are in the transaction that failed.

jdarwood007:
[test-7420b.php.txt](https://github.com/SimpleMachines/SMF2.1/files/8500754/test-7420b.php.txt)

@sbulen Looks like your 7420 script was failing because the key was upper case and the code was looking for lower case.  I just ran it against array_change_key_case which lowercases the key, so we can find it.  Once that was done, it works.

albertlast (commented):
I'm not sure,
if mysql allow the same logic.

Possible that every ddl is an own transaction.

jdarwood007:
We are doing a transaction in pg because we have to issue a create table statement and then the create indexes statement.  If any of those fail, the entire create table should fail.  In MySQL we can do it all in one statement.

jdarwood007:
When we have a auto column (/sequence for pg) we should
1. Enforce that it is not null
2. Ensure it does not have a default (as the auto/sequence will populate it).

I think this is fine restriction to enforce here and ensures all the database systems are happy.

sbulen (approved):
I retested, and #7417 #7419 and #7420 all work now.   

Good stuff...

#7419 still fails on MySQL 5.6, but not sure what we can do about that.

jdarwood007:
Whats your feeling regarding the auto column issue mentioned above?  I think that should be resolved here to round up all the issues we have.

sbulen:
I don't think I've seen a problem with sequences or auto-increments...  

I'm not sure what problem we're trying to solve...

jdarwood007:
I didn't make a test script, but I tested creating a table with a id column that was auto, default of 0 and null.  It failed.  Because well you can't have that.  If you need a test script, I can create one.

sbulen:
I just did a before & after, both mysql & pg.  The column was auto-incremented, default 0, & NULL was true, e.g., it was nullable.

I had no issues with or without this PR.  It always installed clean with no errors, and the column was auto incremented (w/sequence in pg).

It's true that the default & nullable were ignored.  But I saw the same behavior before & after.

I wonder if reverting the prior null-behavior commit changed the behavior.

jdarwood007:
I can't reproduce it anymore.  So at this time, this is ready for approvals.

sbulen:
I just tested the "191" logic, by attempting to create the large index in mysql 5.6, and got the following.  
image

jdarwood007:
Did I do something wrong then?  The 191 is enforced, unless you specify a index length

```
	// MySQL If its a text column, we need to add a size.
	$cols = $smcFunc['db_list_columns']($table_name, true);
	foreach ($index_info['columns'] as &$c)
	{
		$c = trim($c);

		// If a size was already specified, we won't be able to match it anyways.
		if (!isset($cols[$c]) || !in_array($cols[$c]['type'], array('text', 'mediumntext', 'largetext', 'varchar', 'char')))
			continue;

		$c .= '(191)';
	}
```

sbulen:
I suspect the issue was the $default issue?

It is also saying there are conflicts now.  

I'll retest after the conflicts are resolved.

jdarwood007:
I'm not seeing conflicts

sbulen:
Maybe it only applies to rebases?

I'll retest to see if the default fix changed anything.

Screenshot_20220427-180726_Firefox

jdarwood007:
Screen Shot 2022-04-27 at 6 21 29 PM

I did merge release-2.1 to sync the revert of the patch.

jdarwood007:
Your right.. it won't rebase and merge..  but can do a merge request or squash.

sbulen:
That's a merge, not a rebase.  I think the way you merged some commits mid-stream prevents a rebase going forward.  That's all that message means, I think.  You will see it, I think, if you change from merge to rebase.

jdarwood007:
Well I can try to fix it or we just let it do a merge/squash.  I've always used the merge.

sbulen:
OK I retested the 191 fix, and I am still having index length issues in mysql 5.6.   

The $default errors are gone, so they were not related.

It looks like it is not actually using the 191, AFAICT.  (bug_test_2.zip works on mysql 8, but not on mysql 5.6.)

bug_test_2:
```
<?php

global $smcFunc;

if (!isset($smcFunc['db_create_table']))
	db_extend('packages');

$create_tables = array(
	'a_test_record' => array(
		'columns' => array(
		    array(
			   'name' => 'id_atest',
				'type' => 'INT',
				'not_null' => true,
			),
		    array(
			   'name' => 'fails_in_pg',
				'type' => 'TEXT',
				'not_null' => true,
			),
			array(
				'name' => 'fails_in_mysql_56',
				'type' => 'varchar',
				'size' => 512,
				'default' => '',
				'not_null' => true,
			),
		),
		'indexes' => array(
			array(
				'type' => 'primary',
				'columns' => array('id_atest'),
			),
			array(
				'type' => 'index',
				'name' => 'ix_test_text',
				'columns' => array('fails_in_pg (64)'),
			),
			array(
				'type' => 'index',
				'name' => 'ix_test_ix_length',
				'columns' => array('fails_in_mysql_56'),
			),
		),
	),
);

foreach ($create_tables AS $table_name => $data)
	$smcFunc['db_create_table']('{db_prefix}' . $table_name, $data['columns'], $data['indexes']);

```

sbulen:
Backtrace:
image

sbulen:
Note:  To emulate a prefix length of 767 in current mysql, you use innodb tables & change innodb_default_row_format to 'compact', which was the default prior to mysql 5.7.

sbulen:
> Well I can try to fix it or we just let it do a merge/squash. I've always used the merge.

Nope, leave it be.  Sorry for the tangent.

jdarwood007:
Sent up a change that should fix it.  But if not, I will need to get a 5.6 test env setup.

sbulen:
Woohoo!  That worked.  This is from my 5.6 environment:
image

What I think I'll do some time over the next couple of days is repeat all the tests again, just to be safe.

sbulen:
Putting this out here:  With these fixes we are going to support pg far, far better than we have in the past.  Many more mods are going to be able to install & work...

Sources/Subs-Db-postgresql.php
```diff
@@ -913,6 +922,10 @@ function smf_db_error_insert($error_array)
 	if (filter_var($error_array[2], FILTER_VALIDATE_IP) === false)
 		$error_array[2] = null;
 
+	// If we are in a transaction, abort.
+	if ($inTransaction)
+		$smcFunc['db_transaction']('rollback');
```
albertlast: we are here in an error case,
don't expect anything and test everything before use.

Sources/DbPackages-postgresql.php
albertlast: you don't need to add the pr,
you should only add the logic to the code here.

Sources/Subs-Db-postgresql.php
jdarwood007: I'm not sure what your saying here.  Can you clarify that?

I assume your saying something about being in an error state and we are not testing better?  If you have an idea for a better error test let me know.  I simply global the $inTransaction, and set it to true when we start the transaction and false after we commit.  So if a error occurs while inside the transaction, we issue the rollback.

Sources/DbPackages-postgresql.php
jdarwood007: My logic here is your PR adds the owned by for the sequences.  My PR does not touch on this at all as it addresses issues with logic inconsistencies between MySQL and PostgreSQL.    If your PR merges, then I agree with your statement and my PR should be required to have the logic for the sequence owner.  But since it isn't, I don't feel my PR should address this.

If you don't agree with this, you can ping Sesq regarding this as an individual who isn't involved in this PR yet and as the Lead Dev would be the final decision maker regarding this.

Sources/Subs-Db-postgresql.php
albertlast: the reason is clear and fine,
but you use the variables without testing.
smf use empty() for the most cases,
so i would expect to do this in this way here too.

jdarwood007: That makes sense.  Updated

albertlast: dunno if it make sense, but could we also test $smcFunc['db_transaction']('rollback'); if it exists?

jdarwood007: Why would it not exists?  Its defined in the Subs file.  If it doesn't exist, nothing is working.

albertlast: since we are here in the error part so we can't assume that anything is right from here,
reason why we also use prepared statement since it only safe way to hanlde most cases.

Sesquipedalian:
Is this ready to merge?

Sources/Subs-Db-postgresql.php
jdarwood007: @Sesquipedalian Do you want me to add a existence check here as albertlast has suggested

Sesquipedalian: It's hard to imagine a scenario where this code could be executed before `$smcFunc['db_transaction']()` became defined. 

Even so, I'd suggest that you simply call `smf_db_transaction()` directly rather than via `$smcFunc['db_transaction']()`.

Sources/DbPackages-mysql.php
```diff
@@ -420,13 +444,22 @@ function smf_db_change_column($table_name, $old_column, $column_info)
...
+
 	if ($size !== null)
 		$type = $type . '(' . $size . ')';
 
 	$smcFunc['db_query']('', '
-		ALTER TABLE ' . $table_name . '
+		ALTER TABLE ' . $short_table_name . '
 		CHANGE COLUMN `' . $old_column . '` `' . $column_info['name'] . '` ' . $type . ' ' . (!empty($unsigned) ? $unsigned : '') . (!empty($column_info['not_null']) ? 'NOT NULL' : '') . ' ' .
-			(!isset($column_info['default']) ? '' : 'default \'' . $smcFunc['db_escape_string']($column_info['default']) . '\'') . ' ' .
+				($default === '' ? '' : 'DEFAULT ' . $default) . ' ' .
```
Sesquipedalian: Nit: unnecessary extra leading white space

PR #7431, opened 2022-04-16
Better text for error loop

albertlast:
Because of 
https://www.simplemachines.org/community/index.php?topic=581859.0
and
https://www.simplemachines.org/community/index.php?topic=581859.0
i notice that the text of error loop got not enough information for support.
So i extend the error message, to provide admin/support more information about the issue.

Possible we could do more here,
but i guess this is good way to help.

### DIFF
--- a/Sources/Errors.php
+++ b/Sources/Errors.php
@@ -48,7 +48,7 @@ function log_error($error_message, $error_type = 'general', $file = null, $line 
 	if ($error_call > 2)
 	{
 		var_dump($backtrace);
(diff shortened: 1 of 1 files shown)


albertlast:
@sbulen also this could be part of the next release

jdarwood007:
Closing this in favor of #9006
Simply because this targets 2.1, and I don't see this as a bug fix that is causing any problems in 2.1.  We are just making the problem clearer.

PR #7435, opened 2022-04-22
Improves performance when looking up attachments for message previews

### LINKED ISSUE #7434: Looking up new attachments for message previews can be slow
#### Description
Slow queries have been reported when looking up attachments to show in message previews for messages that have not been submitted yet. This is because we currently have to check all attachments where `id_msg` is `0` in order to pick up attachments for unsubmitted messages.

### Steps to reproduce
1. Have lots of avatars in your system.
2. Start writing a new post.
3. Add an attachment to the post, and insert it into the post body using the [attach] BBCode.
4. Request a preview of the post.
5. Notice the long time that it takes to show the preview of the attachment.

- Version/Git revision: 7aa094d
- Database Type: N/A
- Database Version: N/A
- PHP Version: N/A

We can probably fix this by using the $_SESSION['attachments_can_preview'] array to find the attachments we want to preview, rather than looking up every attachment that doesn't have an non-zero `id_msg`.

Sesquipedalian:
Fixes #7434

### DIFF
--- a/Sources/Subs-Attachments.php
+++ b/Sources/Subs-Attachments.php
@@ -1370,9 +1370,6 @@ function prepareAttachsByMsg($msgIDs)
 	else
 		$msgIDs = array_diff($msgIDs, array_keys($context['loaded_attachments']), array(0));
 
-	if (!empty($context['preview_message']))
-		$msgIDs[] = 0;
-
 	if (!empty($msgIDs))
 	{
 		$request = $smcFunc['db_query']('', '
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Pinging @jdarwood007

Sources/Subs-Attachments.php
```diff
@@ -1370,9 +1370,6 @@ function prepareAttachsByMsg($msgIDs)
 	else
 		$msgIDs = array_diff($msgIDs, array_keys($context['loaded_attachments']), array(0));
```
jdarwood007, line 1371: Can we ensure on this line that we don't take in array(0) at all?  Otherwise $msgIDs could still hold id_msg of 0
```
		$msgIDs = array_diff($msgIDs, array_keys($context['loaded_attachments']));
	foreach ($msgIDs as $mID => $mVal)
		if (empty($mVal))
			unset($msgIDs[$mID]);
```

Sesquipedalian: This line already does ensure that `$msgIDs` won't contain a zero. Try this to see what I mean:

```
<?php

$msgIDs = array(123, 456, 789, 0);

$context['loaded_attachments'] = array(
	123 => array('the content of this array is irrelevant for this test'),
	789 => array('the content of this array is irrelevant for this test'),
);

$msgIDs = array_diff($msgIDs, array_keys($context['loaded_attachments']), array(0));

var_export($msgIDs);
```

Output will be:
```
array (
  1 => 456,
)
```

jdarwood007 (approved):

PR #7440, opened 2022-04-27
Adds integrate_word_censor hook

Sesquipedalian:


jdarwood007:
I would assume we want to do that hook before we replace the text?  Otherwise the censored words doesn't provide much use since it was already done.

Sesquipedalian:
🤔 You are right, yes.

PR #7448, opened 2022-05-06
Clamps alert_value within allowed range

### LINKED ISSUE #7430: alert_timeout out of range
#### Description
The id_member was censored, but not the values.  Looks like alert_timeout caused it.
```
/Sources/Subs-Notify.php (Line 90)

Type of error: Database
Error messageSelect Expand
Database Error: Out of range value for column 'alert_value' at row 21


REPLACE INTO smf_user_alerts_prefs(`id_member`, `alert_pref`, `alert_value`)
VALUES
(123456789, 'msg_auto_notify', 1),
(123456789, 'msg_receive_body', 1),
(123456789, 'msg_notify_pref', 1),
(123456789, 'msg_notify_type', 1),
(123456789, 'pm_notify', 1),
(123456789, 'topic_notify', 3),
(123456789, 'board_notify', 3),
(123456789, 'msg_mention', 3),
(123456789, 'msg_quote', 3),
(123456789, 'msg_like', 1),
(123456789, 'unapproved_reply', 3),
(123456789, 'pm_new', 2),
(123456789, 'pm_reply', 2),
(123456789, 'groupr_approved', 3),
(123456789, 'groupr_rejected', 3),
(123456789, 'request_group', 3),
(123456789, 'buddy_request', 1),
(123456789, 'birthday', 3),
(123456789, 'event_new', 3),
(123456789, 'paidsubs_expiring', 3),
(123456789, 'alert_timeout', 391),
(123456789, 'announcements', 1)
````

### Steps to reproduce
1. 
2. 

- Version/Git revision: 2.1.1
- Database Type:
- Database Version:
- PHP Version: 7.4

Sesquipedalian:
Fixes #7430

### DIFF
--- a/Sources/Subs-Notify.php
+++ b/Sources/Subs-Notify.php
@@ -81,7 +81,7 @@ function setNotifyPrefs($memID, $prefs = array())
 
 	$update_rows = array();
 	foreach ($prefs as $k => $v)
-		$update_rows[] = array($memID, $k, $v);
+		$update_rows[] = array($memID, $k, min(max((int) $v, -128), 127));
 
 	$smcFunc['db_insert']('replace',
 		'{db_prefix}user_alerts_prefs',
--- a/Themes/default/Profile.template.php
+++ b/Themes/default/Profile.template.php
@@ -1887,7 +1887,7 @@ function template_alert_configuration()
 						<label for="notify_send_body">', $txt['notify_alert_timeout'], '</label>
 					</dt>
 					<dd>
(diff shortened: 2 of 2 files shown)


jdarwood007:
The logic looks ok.  Although I would say the min can be 0, since we don't need negative alerts.

Sesquipedalian:
The alert_timeout setting should never be negative, sure, but other rows in the table could conceivably need to store a negative value. Basically, this is just making sure we stay within the bounds of a signed TINYINT column in MySQL, rather than trying to impose any particular assumptions about what values might make sense.

jdarwood007:
fair enough.

jdarwood007 (approved):

PR #7459, opened 2022-05-10
Cleans up layout in likes popup

### LINKED ISSUE #7458: Layout is messy in the likes popup list
This has been annoying me for a while:
Screen Shot 2022-05-10 at 5 40 25 PM

Messy, messy, messy.

Sesquipedalian:
Fixes #7458
Screen Shot 2022-05-10 at 5 45 16 PM

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -215,18 +215,15 @@ body#help_popup {
 	align-items: center;
 }
 #likes .avatar {
-	height: 50px;
-	width: 50px;
+	height: 5em;
+	width: 5em;
 	margin: 0 10px 0 0;
 }
-#likes .avatar, #likes li .like_profile {
-	vertical-align: middle;
-	display: inline-block;
+#likes li .like_profile {
+	flex: 1 0 auto;
 }
 #likes li .like_time {
 	padding: 4px 0;
-	flex: 1 0 auto;
-	text-align: right;
 }
 #likes .description {
 	font-style: italic;
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -211,10 +211,12 @@ body#help_popup {
 #likes li {
 	clear: both;
 	padding: 1px 0;
+	display: flex;
+	align-items: center;
 }
 #likes .avatar {
-	max-height: 45px;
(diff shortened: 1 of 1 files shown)


Themes/default/css/index.css
```diff
@@ -211,10 +211,12 @@ body#help_popup {
 #likes li {
 	clear: both;
 	padding: 1px 0;
+	display: flex;
```
albertlast, line 214: looks to me more like a grid layout than a flex layout

Themes/default/css/index.css
```diff
@@ -211,10 +211,12 @@ body#help_popup {
...
+	align-items: center;
 }
 #likes .avatar {
-	max-height: 45px;
-	max-width: 45px;
+	height: 50px;
+	width: 50px;
```
albertlast: i don't understand where this magic numbers come from.
i would expect to had a min height and possible fit content in width in combination of a grid layout.

DiegoAndresCortes: why grid when it's just a list? you are trying to over complicate 3 columns when the only issue was inconsistent size of avatars.

Sesquipedalian: They aren't magic numbers. I just liked 50px better than 45px, and 50px was enough to fit two lines of text comfortably well. But I suppose I could change this to use em units rather than px units.

As for why I used flex rather than grid, that was because flex did the job I needed it to do: aligning everything to the centre on the vertical axis, while also making it easy to justify the date string to the right on the horizontal axis.

Arantor: This may be overwritten by an inline <head> declaration that sets width and height with !important...

Themes/default/css/index.css
```diff
@@ -223,6 +225,8 @@ body#help_popup {
 }
 #likes li .like_time {
 	padding: 4px 0;
+	flex: 1 0 auto;
+	text-align: right;
```
Arantor: This seems like the float_right on the like_time should also be removed if this is a thing.

Sesquipedalian: We could do that, but leaving it in is harmless and requires one less change in the next patch file.

PR #7470, opened 2022-05-19
Avoids errors generating etag when attachment file does not exist

### LINKED ISSUE #7468: sha1_file(/blah/blah/attachments/23993_heapbigblahblah): failed to open stream: No such file or directory
#### Description
Reported in the  forum here:
https://www.simplemachines.org/community/index.php?topic=582243.0
https://www.simplemachines.org/community/index.php?topic=582241.0

I have seen one of these in my forum after our 2.1 upgrade.  In my instance, this file referenced an avatar that had been recently deleted.  I don't think the check to see if the file exists is working somehow.

I get this consistently attempting to download non-existent attachments in 2.1.2.

IP addresses appear to refer to bots; this appears to be crawl activity where reported.  

image

- Version/Git revision: 2.1.2
- Database Type: mysql
- Database Version: 5.6
- PHP Version: 7.4

Sesquipedalian:
Fixes #7468


sbulen (approved):
Confirmed the error message is gone.

PR #7474, opened 2022-05-19
Calls integration hooks for $subActions before setting default subaction

### LINKED ISSUE #7380: Some hooks in the admin area do not work properly
#### Description
Almost all hooks like `call_integration_hook('integrate_<hook_name>', array(&$subActions));` have such a problem, if they have a string like `$_REQUEST['sa'] = isset($_REQUEST['sa']) && isset($subActions[$_REQUEST['sa']]) ? $_REQUEST['sa'] : '<main_action>';` before calling.

### Steps to reproduce
Try to add a new section in the Posts and Topics area with `integrate_manage_posts` hook.

- Version: Current

### Possible solution
1. Remove strings like `$_REQUEST['sa'] = isset($_REQUEST['sa']) && isset($subActions[$_REQUEST['sa']]) ? $_REQUEST['sa'] : '<main_action>';`

2. Move `loadGeneralSettingParameters` function from ManageSettings.php to Admin.php and use where is possible:

ManagePosts.php:
```php
call_integration_hook('integrate_manage_posts', array(&$subActions));

loadGeneralSettingParameters($subActions, 'posts');
```

ManageBans.php
```php
call_integration_hook('integrate_manage_bans', array(&$subActions));

loadGeneralSettingParameters($subActions, 'list');
```

ManageCalendar.php
```php
call_integration_hook('integrate_manage_calendar', array(&$subActions));

loadGeneralSettingParameters($subActions, $default);
```

... etc.

### LINKED ISSUE #7461: integrate_manage_subscriptions runs too late to be useful
integrate_manage_subscriptions exists notionally to allow adding subactions to the manage-paid-subscriptions area in ManagePaid.php. However, it runs too late to actually do this.

The hook allows adding to the sub actions array in the root of ManagePaidSubscriptions but by the time the hook runs, the subaction will have already been validated against the known list and unknown subactions will have been removed.

The call needs to be moved before $_REQUEST['sa'] is sanitised.

Sesquipedalian:
Fixes #7380
Fixes #7461

### DIFF
--- a/Sources/Groups.php
+++ b/Sources/Groups.php
@@ -32,6 +32,8 @@ function Groups()
 		'requests' => array('GroupRequests', 'group_requests'),
 	);
 
+	call_integration_hook('integrate_manage_groups', array(&$subActions));
+
 	// Default to sub action 'index'.
 	$_REQUEST['sa'] = isset($_REQUEST['sa']) && isset($subActions[$_REQUEST['sa']]) ? $_REQUEST['sa'] : 'index';
 
@@ -57,9 +59,6 @@ function Groups()
 			'name' => $txt['groups'],
 		);
 	}
-
-	// CRUD $subActions as needed.
-	call_integration_hook('integrate_manage_groups', array(&$subActions));
 
 	// Call the actual function.
 	call_helper($subActions[$_REQUEST['sa']][0]);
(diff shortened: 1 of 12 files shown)


DiegoAndresCortes:
Bugo also had _integrate_manage_mail_, is that an oversight or was it not affected?

Sesquipedalian:
It wasn't necessary, no. That was one of the cases where Bugo's changes went beyond what was necessary. (For what it's worth, his more extensive changes were all good ideas, but they would need to go into a future point release, not a patch release.)

live627 (approved):

PR #7490, opened 2022-05-25
Fix Tracking IPv6s

jdarwood007:
If you enter a IPv6 in the format:   `2001:abc:123::*`  SMF will fail to convert it to valid IPv6 range. The result is `2001:abc:123::0000` to `2001:abc:123::ffff`
However if you do a it in the format: `2001:abc:123:*` SMF will generate a valid IPv6 range of `2001:abc:123:0000:0000:0000:0000` to `2001:abc:123:ffff:ffff:ffff:ffff`

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -5143,6 +5143,13 @@ function ip2range($fullip)
 	if ($fullip == 'unknown')
 		$fullip = '255.255.255.255';
 
+	// Cleanup the IP as needed.
+	$fullip = strtr($fullip,
+		[
+			// IPv6 ending in :: is valid for a network, but SMF doesn't handle itwell.
+			'::*' => ':*',
+		]);
+
 	$ip_parts = explode('-', $fullip);
 	$ip_array = array();
 


albertlast:
when i see this correctly ``2001:abc:123::ffff`` and ``2001:abc:123::0000`` are valid, but i guess the intention of the user is not this?

jdarwood007:
Those are valid, but remember its only ::0000 to ::ffff or just the last bit is wildcard, when your most likely intending to wildcard the entire network (/64) when you do a ::*

albertlast:
From a techincal pov is ``2001:abc:123::2002``  -> ``2001:abc:123:0000:0000:0000:0000:2002``
based on this the existing logic work
``2001:abc:123::*``  -> ``2001:abc:123:0000:0000:0000:0000:0000`` a. ``2001:abc:123:0000:0000:0000:0000:ffff``
``2001:abc:123::*:*``  -> ``2001:abc:123:0000:0000:0000:0000:0000`` a. ``2001:abc:123:0000:0000:0000:ffff:ffff``
``2001:abc:123:*``  -> ``2001:abc:123:0000:0000:0000:0000:0000`` a. ``2001:abc:123:ffff:ffff:ffff:ffff:ffff``

jdarwood007:
So you believe its fine?  I would more expect ::* to expand the entire range not jus last bits.

albertlast:
Maybe a help text with further information would be helpfull?

jdarwood007:
If thats the case then we could.

live627:
AFAICT, a double-colon expands to zero, so shouldn't `2001:abc:123::*` expand from `2001:abc:123:0000:0000:0000:0000` to `2001:abc:123:0000:ffff:ffff:ffff`? 

According to https://tyler.anairo.com/ipv6-addresses, a double-colon could expand to multiple octet pairs.  Great, now we have ambiguity.

> A fully qualified IPv6 address takes the format of `XXXX:XXXX:XXXX:XXXX:XXXX:XXXX:XXXX:XXXX`. An example fully-qualified address is `FE80:0000:0000:0000:903A:1C1A:E802:11E4`.
Do you see the long string of zeros? With IPv6 addresses, you can compress zeros to shorten the IP address.
You can compress zeros like this: `FE80:0:0:0:903A:1C1A:E802:11E4`, or like this: `FE80::903A:1C1A:E802:11E4`. Take note that you can only compress one set of zeros to `::`.
Like in this IP: `FE80:0000:0000:0000:903A:0000:0000:11E4`, there are 2 sets of zeros. You could write it like this: `FE80::903A:0:0:11E4`, or like this: `FE80:0:0:0:903A::11E4`, but not like this: `FE80::903A::11E4`.

albertlast:
> AFAICT, a double-colon expands to zero, so shouldn't `2001:abc:123::*` expand from `2001:abc:123:0000:0000:0000:0000` to `2001:abc:123:0000:ffff:ffff:ffff`?

Since ``2001:abc:123::1234`` expand to ``2001:abc:123:0000:0000:0000:1234``,
i don't think so, that the wild card operate differently.
Which is free of ambiguity.

jdarwood007:
Looking at my pfsense, it shows my IPv6 like this:

Subnet:   2001:332:cd2h:325::
Subnet Mask:   64 bits
Available Range:   2001:332:cd2h:325:: to 2001:332:cd2h:325:ffff:ffff:ffff:ffff
Range:    2001:332:cd2h:325::1000     -   2001:332:cd2h:325::2000

DNS Servers
2606:4700:4700::1111
2606:4700:4700::1001

So I think if we treat it like a "subnet" the double colon would be correct.
If the code changes are not too difficult, we could leave it as is and implant cidr notation expansion.  Which would clear up the issue right away as the cidr mask can define the range fairly quickly.

albertlast:
still i don't believe that the common admin is known of subnet,
so your pov is minority and most see the input mask als ip address where you mention behavior make no sense.

PR #7495, opened 2022-06-01
Only force a index size if the column is larger than 191 and not specified

### LINKED ISSUE #7492: Database error installing mod
#### Description

```
Incorrect prefix key; the used key part isn't a string, the used length is longer than the key part, or the storage engine doesn't support unique prefix keys
File: /Packages/temp/install-sd.php
Line: 47

CREATE TABLE smf_helpdesk_role_permissions
(
`id_role` smallint(5) unsigned NOT NULL DEFAULT 0,
`permission` varchar(40) NOT NULL ,
`add_type` tinyint(3) unsigned NOT NULL DEFAULT 0,
PRIMARY KEY (id_role,permission(191))) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci


CREATE TABLE smf_helpdesk_preferences
(
`id_member` mediumint(8) unsigned NOT NULL DEFAULT 0,
`variable` varchar(30) NOT NULL ,
`value` text NOT NULL ,
PRIMARY KEY (id_member,variable(191))) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci
```

### Steps to reproduce
1.  Install SimpleDesk from attached file
2. Fails

- Version/Git revision: [5c1766b](https://github.com/SimpleMachines/SMF/commit/5c1766bd571bbddb12b9ae749e5400bf8382cc9c)
- Database Type: Mysql
- Database Version: 8.0
- PHP Version: 8.0

Related to #7424

@sbulen Looks like a BC issue?
There is no collation specified, so its possible i'm defaulting to mb4.  PMA doesn't make it clear which collation is being used.

I was able to work around it by specifying a (30) in the indexes.
[SimpleDesk_2.1.0-broken.zip](https://github.com/SimpleMachines/SMF/files/8790252/SimpleDesk_2.1.0-broken.zip)

jdarwood007:
Fixes #7492

### DIFF
--- a/Sources/DbPackages-mysql.php
+++ b/Sources/DbPackages-mysql.php
@@ -167,7 +167,17 @@ function smf_db_create_table($table_name, $columns, $indexes = array(), $paramet
 
 			// If a size was already specified, we won't be able to match it anyways.
 			$key = array_search($c, array_column($columns, 'name'));
-			if ($key === false || !isset($columns[$key]) || !in_array($columns[$key]['type'], array('text', 'mediumntext', 'largetext', 'varchar', 'char')))
+			$columns[$key]['size'] = isset($columns[$key]['size']) && is_numeric($columns[$key]['size']) ? $columns[$key]['size'] : null;
+			list ($type, $size) = $smcFunc['db_calculate_type']($columns[$key]['type'], $columns[$key]['size']);
+			if (
+				$key === false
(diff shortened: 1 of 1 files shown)


sbulen (approved):
LGTM

PR #7498, opened 2022-06-05
Remove PHP.ini limits from attachments per post in UI context

### LINKED ISSUE #7486: Attachments count inconsistency in the Restrictions line
#### Description
This was reported on the Bug Reports board:
https://www.simplemachines.org/community/index.php?topic=582435.0

jdarwood007:
Fixes #7486

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -415,7 +415,7 @@ function reloadSettings()
 	$file_max_kb = floor(memoryReturnBytes(ini_get('upload_max_filesize')) / 1024);
 	$modSettings['attachmentPostLimit'] = empty($modSettings['attachmentPostLimit']) ? $post_max_kb : min($modSettings['attachmentPostLimit'], $post_max_kb);
 	$modSettings['attachmentSizeLimit'] = empty($modSettings['attachmentSizeLimit']) ? $file_max_kb : min($modSettings['attachmentSizeLimit'], $file_max_kb);
-	$modSettings['attachmentNumPerPostLimit'] = min(ini_get('max_file_uploads'), (empty($modSettings['attachmentNumPerPostLimit']) ? 50 : $modSettings['attachmentNumPerPostLimit']));
+	$modSettings['attachmentNumPerPostLimit'] = !isset($modSettings['attachmentNumPerPostLimit']) ? 4 : $modSettings['attachmentNumPerPostLimit'];
 
 	// Integration is cool.
 	if (defined('SMF_INTEGRATION_SETTINGS'))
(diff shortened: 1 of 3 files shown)


sbulen:
I've done some testing and dz handles it all just fine...  You can GREATLY exceed the php.ini limits using dz.  I've done 100+ on my forum recently, when uploading softball pix...  

I think the fix needs to be simply along the lines of:
`		$context['num_allowed_attachments'] = (empty($modSettings['attachmentNumPerPostLimit']) ? 4 : $modSettings['attachmentNumPerPostLimit']);
`

Because 4 is our default.

Seriously, I've been testing this change & I can't find a way to make it fail at all, even with 'UPLOAD ALL' and a massive over-upload.  dz takes it all in stride, & provides the proper errors where appropriate.

sbulen (changes requested):
After some testing, I think the ONLY change needed is the following:
`$context['num_allowed_attachments'] = (empty($modSettings['attachmentNumPerPostLimit']) ? 4 : $modSettings['attachmentNumPerPostLimit']); `

jdarwood007:
You are right on that.  Spuds did the change on that line last from this commit https://github.com/SimpleMachines/SMF/commit/a8f52e3a12dc77e69c891cac95f5fe8e3fc6fd0f

But if we do like you suggest, I think it is better to handle it in Load.php, so it is handled early on and consistently across the code base we don't have to find all instances of its usage.

sbulen:
I would agree that we should never overwrite `$modSettings['attachmentNumPerPostLimit']` with `ini_get('max_file_uploads')`.

If there are other places that occurs, we should change it.  But only then....

We have a very narrow problem here - and should only make the changes necessary.

sbulen:
This is the only place I see max_file_uploads being used.  So I think we're good with the one-line change.

jdarwood007:
How about these changes.  We did use attachmentNumPerPostLimit in a few places in Post.php.  So if we are setting it to something (even for the dropzone), it needs to be logically set somewhere.  Used a !isset.

sbulen:
Oh, I think I see what you're saying....  Where it's empty, what do you use?  There are a handful of places that will set it to 50 if it's empty...  🙄

sbulen (approved):
LGTM

jdarwood007:
You are right, we are doing some hardcoded checks to force the limit to 50.  Do we just set the max to 50 then?  Thats a lot of attachments per post.

Subs-Attachment.php
```
	// Set a max limit if none exists
	if (empty($modSettings['attachmentNumPerPostLimit']) && $context['attachments']['quantity'] >= 50)
		$modSettings['attachmentNumPerPostLimit'] = 50;
```

sbulen:
It'd be nice to be consistent.  Either 4 or 50 works for me.  

I doubt it will ever be found empty, TBH.  But consistency is good.  

I like 4 because it is consistent with the install & upgrade scripts.

jdarwood007:
Well in the manage, I can set the max to 50.  We can then set the check to do !isset and then ensure its 50 or below.

I'm not sure why dz has a 50 limit.

sbulen:
I don't think dz has a 50 limit...  As noted earlier, I've uploaded more than that before.

Where I see 50 is by doing a grep on attachmentNumPerPostLimit.  There are a couple spots with 50.

jdarwood007:
The above code is the only place where I find 50.  How about I just rip it out?

sbulen:
I'd rather not rip stuff out...  That's a little too FAFO for me.

I think there is one other default to 50 over in Subs-Attachments.

sbulen:
We should probably change the title so it's not confusing later.

sbulen (approved):

PR #7499, opened 2022-06-08
fix: coerce user permissions to array before checking it

live627:
this is a longstanding issue that sometimes rears its ugly head. I tried to parch it once ago (I think it was 2.0.14) which was adequate at the time when PHP was less strict about type mismatches.

**Forum topics of interest**

- https://www.simplemachines.org/community/index.php?topic=580558.0
- https://www.simplemachines.org/community/index.php?topic=582622.0


sbulen (approved):
LGTM.

Sorry I missed this earlier.

PR #7501, opened 2022-06-09
Make sure the function we're trying to call is actually defined first...

Oldiesmann:
Attempting to call utf8_convert_case in Load.php fails because it's defined in Subs-Charset.php, which isn't included prior to calling this function.

**Relevant topic from the forum:**

- https://www.simplemachines.org/community/index.php?topic=582640.0

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -234,7 +234,7 @@ function reloadSettings()
 		{
 			return $smcFunc['convert_case']($string, 'ucwords');
 		},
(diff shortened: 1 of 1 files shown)


Sources/Load.php
```diff
@@ -289,7 +289,10 @@ function reloadSettings()
 				}
 				// Everything else.
 				else
+				{
+					include_once($sourcedir . '/Subs-Charset.php');
```
live627:
```suggestion
					require_once($sourcedir . '/Subs-Charset.php');
```

live627:
```suggestion
					require_once($sourcedir . '/Subs-Charset.php');
```

live627 (approved):

PR #7509, opened 2022-06-25
Double-check alert counter

sbulen:
Try fixing the alert counter with a sledgehammer...

Works, might be overkill.


sbulen:
Not sure the sledgehammer is needed...

PR #7510, opened 2022-06-25
Fix alert counts

sbulen:
Three changes:
 - Addressed an issue where alert counter was not getting updated for all subscribers on replies.  This is a bug that definitely impacts the alert counter and popup.
 - Used the standard calculation in one instance where the logic looked a little sketchy.
 - Removed an erroneous calc, where it was updating the alert counter with _new_ unread, instead of _all_ unread.  Not sure it was necessary there anyway.

### DIFF
--- a/Sources/Display.php
+++ b/Sources/Display.php
@@ -1039,8 +1039,13 @@ function Display()
 					'now' => time(),
 				)
 			);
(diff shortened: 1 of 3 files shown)


sbulen:
I've been running this on my site the last couple of days.  Alerts are delivered much more consistently now.

Also, I periodically audit the alert counter, and it is now perfectly aligned with the alerts, and does not fall out of sync anymore.

Sesquipedalian (approved):

Issue #7423, opened 2022-04-07
Unread alerts should be deleted when relevant content no longer exists.


Sesquipedalian:
#### Description
Suppose that User A writes a post and has like notifications turned on. Then suppose that User B clicks the like button for that post twice (i.e. liking it and then unliking it). An alert is generated for the like event, but the alert is not removed after the unlike event. So when User A checks their alerts, they see one about receiving a like even though that like no longer exists.

Similar situations happen when:
- an alert is issued about a new reply to a topic, but then that reply is deleted
- an alert is issued about a new moderation report, but then the report is closed
- etc.

### Steps to reproduce
1. Do one of the things I described above
2. 

### Environment (complete as necessary)
- Version/Git revision: ae0ab6f
- Database Type: N/A
- Database Version: N/A
- PHP Version: N/A

### Additional information/references
See https://www.simplemachines.org/community/index.php?topic=581756.0

PR #7511, opened 2022-06-26
Delete quote & mention alerts if removed during edit

sbulen:
Partial for #7423 

If a post is edited, & mentions or quotes are removed, delete any alerts for them.  The logic to do this cleanup has been around, but it wasn't being invoked correctly.  It works fine once it's invoked.

It needed to know approval status, because you don't want to generate alerts for unapproved posts.  But for some reason, the approval status wasn't passed.  

Also, it needed to explicitly check for removals.  Previously it was just checking for the presence of quotes or mentions in the edited post...  So...  If they were deleted...  It didn't know it had cleanup to do...  🙄

### DIFF
--- a/Sources/Post.php
+++ b/Sources/Post.php
@@ -2340,10 +2340,6 @@ function Post2()
 			$msgOptions['poster_time'] = $row['poster_time'];
 		}
 
-		// This will save some time...
-		if (empty($approve_has_changed))
-			unset($msgOptions['approved']);
-
 		modifyPost($msgOptions, $topicOptions, $posterOptions);
 	}
 	// This is a new topic or an already existing one. Save it.
@@ -3105,6 +3101,7 @@ function JavaScriptModify()
 			'body' => isset($_POST['message']) ? $_POST['message'] : null,
 			'icon' => isset($_REQUEST['icon']) ? preg_replace('~[\./\\\\*\':"<>]~', '', $_REQUEST['icon']) : null,
 			'modify_reason' => (isset($_POST['modify_reason']) ? $_POST['modify_reason'] : ''),
+			'approved' => (isset($row['approved']) ? $row['approved'] : null),
 		);
 		$topicOptions = array(
 			'id' => $topic,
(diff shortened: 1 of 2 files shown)


Sources/Post.php
```diff
@@ -2340,10 +2340,6 @@ function Post2()
 			$msgOptions['poster_time'] = $row['poster_time'];
 		}
 
-		// This will save some time...
-		if (empty($approve_has_changed))
-			unset($msgOptions['approved']);
-
```
live627, line 2346: I think this was here to stop unwanted approval state change.

https://github.com/SimpleMachines/SMF/blob/acee5b182b8f692c99fb4c6af08eb1d0258a4cfe/Sources/Subs-Post.php#L2392-L2394

sbulen: Hmmm...  I don't think I see how that can happen.  If the approval state didn't change, and it had been approved, it sure looks OK to invoke the approval logic without harm.  Tests OK, too.

Have you found a specific scenario that fails?  How can 'approved' get incorrectly set?

OTOH, if the approval state did not change, and the post had been previously approved, there is a definite issue here...  It needs to get to approvePosts().

sbulen: & BTW, thanks for taking a look...

Sesquipedalian:
We already have code that does this job:
https://github.com/SimpleMachines/SMF/blob/06cdc6cef34672e65f9627a1acc29bf066b046c1/Sources/tasks/CreatePost-Notify.php#L352-L418
That's why I didn't mention this particular situation in #7423.

sbulen:
It's not working.  That's why this change is needed.  From the original note above:

> The logic to do this cleanup has been around, but it wasn't being invoked correctly. It works fine once it's invoked.
> 
> It needed to know approval status, because you don't want to generate alerts for unapproved posts. But for some reason, the approval status wasn't passed.
> 
> Also, it needed to explicitly check for removals. Previously it was just checking for the presence of quotes or mentions in the edited post... So... If they were deleted... It didn't know it had cleanup to do..

Sesquipedalian:
Fair enough then. 🙂

PR #7514, opened 2022-06-28
Fix HTML in theme settings

DiegoAndresCortes:
- Number type doesn't use size anymore
- Regular inputs had the wrong **id**


sbulen (approved):
LGTM

PR #7515, opened 2022-06-29
Mark orphaned alerts as read

sbulen:
Partial for #7423

Unread alerts are not purged.  Alerts that point to messages that have been deleted or moved to boards that the user does not have visibility to, are not displayed.  

So...  These unread alerts will stay around forever as invisible orphans...   (Although a nit, this really bugs me...)

This can be easily detected during the count function, which is frequently invoked, so this PR cleans them up by marking them as read.

_An alternate approach would be to change the purge function to purge alerts whether or not they have been read..._

### DIFF
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -2398,7 +2398,7 @@ function alert_count($memID, $unread = false)
 
 	// We have to do this the slow way as to iterate over all possible boards the user can see.
 	$request = $smcFunc['db_query']('', '
-		SELECT id_alert, content_id, content_type, content_action
+		SELECT id_alert, content_id, content_type, content_action, is_read
 		FROM {db_prefix}user_alerts
 		WHERE id_member = {int:id_member}
 			' . ($unread ? '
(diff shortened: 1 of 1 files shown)


sbulen:
OK, after running this on my site for a while, I think there are two issues with it:
1. It should delete, not mark unread.  The alerts themselves point to invalid messages for that user, so they'll get ugly messages if they click on them.
2. It doesn't update the member count to reflect the change in status, so it can knock the count out of sync (e.g., say there's 1 unread but there are none).  Although this seems like an easy fix via a call to updateMemberData(), the problem is that updateMemberData() calls this function, causing a loop...

So...  

Either I need to update the member record directly via sql, OR, replicate a big chunk of alert_count() elsewhere...  I'm leaning towards the former...  Input welcome.

live627:
> . I'm leaning towards the former

That seems to be quick and easy, the better solution here. And would probably result in fewer operations for the patch, a win in my book.

sbulen:
Change made.

sbulen:
Found another tweak needed while testing...  $user_info isn't populated when this is invoked via cron, so don't use it...

Sources/Profile-Modify.php
```diff
@@ -2495,10 +2495,52 @@ function alert_count($memID, $unread = false)
...
+		$smcFunc['db_query']('', '
+			UPDATE {db_prefix}members
+			SET alerts =
+				CASE
+					WHEN alerts < {int:unread_deletes} THEN 0
+					ELSE alerts - {int:unread_deletes}
+				END
+			WHERE id_member = {int:member}',
+			array(
+				'unread_deletes' => $num_unread_deletes,
```
Sesquipedalian, line 2540: `$num_unread_deletes` is never incremented above, so it will always be 0. I think you can simply get rid of this variable and instead use `count($deletes)` on this line.

Sesquipedalian (changes requested):

Sources/Profile-Modify.php
sbulen: It's incremented on L2505 above.

And it tests fine - this fix has been on my production forum for quite some time.

Note I have a count validation script that runs occasionally to ensure the counts don't fall out of sync again - they aren't.

sbulen: One other note: unread deletes <> deletes, so the count won't work.

Sesquipedalian: Derp. I missed that somehow.

Sesquipedalian (approved):

Sesquipedalian:
A followup question for @sbulen:

Since this PR added code to update the value of the alerts field in the member table every time the `alert_count()` function is called, is there any reason not to make the following change to the `updateMemberData()` function?

Current:
https://github.com/SimpleMachines/SMF/blob/aee494fe17322c24c2076cc084b3f4559e61db5b/Sources/Subs.php#L375-L390

New:
```
		if ($var == 'alerts' && ($val === '+' || $val === '-'))
		{
			include_once($sourcedir . '/Profile-Modify.php');

			foreach ((array) $members as $k => $v)
				alert_count($v, true);

			continue;
		}
```

It seems like setting the new value again further down in `updateMemberData()` is redundant, since we just did it in `alert_count()`. Is there something I am missing here, or would this be a good change for a further PR?

sbulen:
> It seems like setting the new value again further down in updateMemberData() is redundant, since we just did it in alert_count(). Is there something I am missing here, or would this be a good change for a further PR?

Sorry, missed this question earlier.

This is noted above in my 2nd comment above.  Since updateMemberData() in turn calls this function, it causes a tight loop.  

Trust me...  It's not a good thing...

> 2. It doesn't update the member count to reflect the change in status, so it can knock the count out of sync (e.g., say there's 1 unread but there are none).  Although this seems like an easy fix via a call to updateMemberData(), the problem is that updateMemberData() calls this function, causing a loop...

PR #7516, opened 2022-06-30
Delete unread alerts upon an unlike

sbulen:
Partial for #7423 

Just as it says, if there is an unread alert for a like, and the user changes their mind and unlikes it, get rid of the alert & adjust the recipient's alert counter.

### DIFF
--- a/Sources/Likes.php
+++ b/Sources/Likes.php
@@ -286,6 +286,39 @@ class Likes
 		// Are we calling this directly? if so, set a proper data for the response. Do note that __METHOD__ returns both the class name and the function name.
 		if ($this->_sa == __FUNCTION__)
 			$this->_data = __FUNCTION__;
+
+		// Check to see if there is an unread alert to delete as well...
+		$result = $smcFunc['db_query']('', '
+			SELECT id_alert, id_member FROM {db_prefix}user_alerts
+			WHERE content_id = {int:like_content}
+				AND content_type = {string:like_type}
+				AND id_member_started = {int:id_member_started}
+				AND content_action = {string:content_action}
+				AND is_read = {int:unread}',
+			array(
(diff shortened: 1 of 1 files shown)


live627:
Should anything be done with read alerts?

frandominguezl:
Well, you have two choices here and I've seen both of them. For me I actually think it's better to leave the alert when I read it even though the like is gone, I would realize when I visit the post again, but I don't know if this is the same for the general user.

sbulen:
I agree with d3vcho here.

Sesquipedalian:
The problem I see here (and this would apply to any function that automatically deletes alerts) is that the alerts are created by a background task, whereas this code deletes the alert immediately. This creates a race condition where the attempt to delete the alert could easily occur _before_ the alert has been created. In particular, if the user rapidly clicks "Like" and then "Unlike", the attempt to delete the alert might happen before the background task to create it runs. 

To avoid this problem, the delete also needs to happen via a background task.

sbulen:
It cannot do so before the alert is created - because it queries the alerts above, and only deletes it if it exists already...

Sesquipedalian:
Sure, but then when the background task runs, the alert will be created after it was supposed to be deleted.

Put simply, since alerts are created via the background tasks queue, they also need to be deleted via the background tasks queue. That will ensure that the operations happen in the correct chronological order.

Sesquipedalian:
If you have already started to work on code to do the deletes via background tasks, then I will leave you to it, @sbulen. But if you'd prefer, I can take a shot at it instead. I've got an initial version for deleting obsolete unread alerts about likes that seems to work pretty well, and I should be able to implement similar functionality for other alert types fairly quickly.

sbulen:

sbulen:

PR #7517, opened 2022-07-03
added mem.real_name to group by

### LINKED ISSUE #7508: Can't sort Sent PM's by User
#### Description
https://www.simplemachines.org/community/index.php?topic=582802.0

### Steps to reproduce
1.  Go to the pms > Sent items
2. Attempt to sort by user

albertlast:
extraced the group by logic in a own var,
with the hope that it better is to read,
instead of compress more logic into this inline.

fixes: #7508 

tested in pg

### DIFF
--- a/Sources/PersonalMessage.php
+++ b/Sources/PersonalMessage.php
@@ -915,14 +915,19 @@ function MessageFolder()
 			);
 		}
 
+		$group_by = $context['folder'] == 'sent' ? '
+			GROUP BY pm.id_pm, pm.subject, pm.id_member_from, pm.body, pm.msgtime, pm.from_name' . 
+				($context['sort_by'] == 'name' ? ', mem.real_name' : '')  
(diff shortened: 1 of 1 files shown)


live627:
What are your thoughts on changing the sort to `pm.from_name` and leaving the `GROUP BY` alone? This way I think it can sort guests, too.

albertlast:
dunno from ux side,
on programm side we could drop the join in this case, i guess.
so would be faster.

live627:
sounds  good

sbulen:
I have tested this & it works.  

I actually don't think the idea of using pm.from_name will work, because depending on whether you're looking at the inbox or the sent folder, the join to mem is different...  It's not the from_name when looking at the sent folder, which is what is failing here.

Make sense?

sbulen (approved):
I think this is good to go.

PR #7518, opened 2022-07-07
Redis Cache

albertlast:
i notice a thread about redis cache,
so i implement this version,
need to do some kind of test (i'm not able to test all the things like socket or auth)

in my test env is redis slower than pg cache.

I guess this pr need/get some love in the next days.


jdarwood007:
@albertlast , This PR targets 2.1.  I don't see us adding this in 2.1.  Can you retarget this for 3.0?  It should mostly be the same code.

PR #7519, opened 2022-07-07
Don't use the cache if it fails

albertlast:
When the cache fail,
smf is still using the cache,
lead into a lock out from smf.

Change to:
Cache connection failed, cache is not used.


sbulen (approved):
LGTM.  

Thanks, albertlast!

PR #7520, opened 2022-07-09
Restore showing profile areas added with `integrate_profile_areas`

live627:
`integrate_profile_areas ` should  work (retain functionality from  2.0). The problem is that the changed array is lost when `createMenu()` exits, so that the calling code in `Profile.php` thinks that nothing happened.

#### What this PR does
- Move the hook point from `createMenu()` to `Profile()` (like it was in 2.0 which I think is proper).
- Deprecate `integrate_pre_profile_areas` because it is unused now.
- Add a new menu option `disable_hook_call` to make this work right.

#### Alternative fix
Pass `$menuData` as a reference

#### Related issues
- https://github.com/SimpleMachines/SMF/issues/3118
- https://github.com/SimpleMachines/SMF/issues/3145
- https://github.com/SimpleMachines/SMF/issues/3945

### DIFF
--- a/Sources/Profile.php
+++ b/Sources/Profile.php
@@ -491,6 +491,10 @@ function ModifyProfile($post_errors = array())
 	);
 
 	// Let them modify profile areas easily.
+	call_integration_hook('integrate_profile_areas', array(&$profile_areas));
+
+	// Deprecated since 2.1.4 and will be removed in 3.0.0. Kept for compatibility with early versions of 2.1.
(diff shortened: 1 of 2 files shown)


live627:
> Move the hook point from `createMenu()` to `Profile()` (like it was in 2.0 which I think is proper).

The hook call in `createMenu()` is still there, intact,  which means that `integrate_admin_areas` and  any other menu hooks  are unaffected.

PR #7523, opened 2022-07-12
The parameter to `smf_mysql_insert_id()` should be optional

live627:
https://www.simplemachines.org/community/index.php?topic=582937.0


jdarwood007:
Shouldn't we be passing $rs to the mysql_insert_id then?

sbulen:
It's a little different than most other mysqli calls.  It wants the connection, not the result.  So it uses the global for the connection, & doesn't need or use the parameter.

PR #7524, opened 2022-07-12
Fix delete of fulltext index for pg

albertlast:
PG got different syntax for deleting index.

Atm we can't delete the ftx.

### DIFF
--- a/Sources/ManageSearch.php
+++ b/Sources/ManageSearch.php
@@ -267,14 +267,22 @@ function EditSearchMethod()
 		checkSession('get');
 		validateToken('admin-msm', 'get');
 
-		$smcFunc['db_query']('', '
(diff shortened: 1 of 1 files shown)


jdarwood007:
Its hard coded?

albertlast:
It's same the code from above, where the index get deleted.

frandominguezl:
I don't remember the coding standards, but I believe you should actually use curly braces if the if/else have more than one line

albertlast:
@sbulen should this not be a part of the next release?

sbulen (approved):
LGTM.  Tested adds & deletes in pg (14) & mysql (8.0)

sbulen:
& thanks for the contributions, @albertlast !

Issue #7389, opened 2022-03-20
Nit: Altering start value in url can result in fractional page numbers


sbulen:
#### Description
This will happen if you enter a value beyond the range of the list you are viewing as the "start" value in the URL...

This can happen anywhere we use constructPageIndex() - browsing topics, memberlists, etc., whenever the start value requested in the URL > the max value for that list.

![image](https://user-images.githubusercontent.com/23568484/159147834-647922cc-c1fd-406b-8667-91522c2744c7.png)

![image](https://user-images.githubusercontent.com/23568484/159147816-76fb3840-020e-4081-baa0-a117434052d4.png)

### Steps to reproduce
1. Browse the memberlist.  
2. In your browser URL, add ";start=3000001" to the end of the URL.  (Any value beyond the actual range of the list in question.)

(Don't make it TOOOOOO big or the fraction disappears due to rounding... 🙄)

### Additional information/references
Lowest of low priority...  You can ONLY get the odd behavior if you manually edit the URL.  
But... Sometimes folks manually edit the URL...

live627:
> (Don't make it TOOOOOO big or the fraction disappears due to rounding...
🙄)

32-bit floating point precision sure is fun to fiddle with, especially in
video games.

sbulen:
Dupe of #7522 - closing

PR #7526, opened 2022-07-17
fix bounds check for page index

### LINKED ISSUE #7522: Pagination error
while cleaning up some spam, I noticed that some page numbers seemed wrong

see where it says `1.95` instead of  `1`

https://www.simplemachines.org/community/index.php?topic=582201.40

Seems to me that this only happens when manually fiddling with the numbers.

Marking as `Serious bug` with **high priority** because it should load the last page of posts, not be empty.

live627:
- fix #7522

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -566,8 +566,8 @@ function constructPageIndex($base_url, &$start, $max_value, $num_per_page, $flex
 	$start_invalid = $start < 0;
 
 	// $start must be within bounds and be a multiple of $num_per_page.
-	$start = max(0, $start);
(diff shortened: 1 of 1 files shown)


jdarwood007:
It might also make sense to force cast this to an int.

live627:
cast `$start`?

jdarwood007:
Correct.  Just to ensure it should always be a int and not a decimal.

sbulen (approved):
LGTM.

Note recasting to int isn't necessary, because php mod math casts to int.

I tested & confirmed it fixes my page decimal issue as well (#7389 ).

PR #7527, opened 2022-07-17
Cleanup unread alerts after moderation activity

sbulen:
Partial for #7423 

If there are any unread alerts out there related to moderation activity that has been closed out, flag them as read.


Sesquipedalian:
My comment [here](https://github.com/SimpleMachines/SMF/pull/7516#issuecomment-1217173908) also applies to this PR, I'm afraid. 🙁

sbulen:
It cannot attempt the deletion before the alert exists, because this logic actually queries the alert...

Sesquipedalian:
See https://github.com/SimpleMachines/SMF/pull/7516#issuecomment-1218200073 🙂

sbulen:

Sesquipedalian (changes requested):
There are a few other functions in the codebase where reports get closed without calling `updateReport()`, so it would be good to add calls to `clearReportAlerts()` in those other functions. Here's a list of links showing where those other locations are:
- https://github.com/SimpleMachines/SMF/blob/aee494fe17322c24c2076cc084b3f4559e61db5b/Sources/ModerationCenter.php#L708
- https://github.com/SimpleMachines/SMF/blob/aee494fe17322c24c2076cc084b3f4559e61db5b/Sources/ModerationCenter.php#L787
- https://github.com/SimpleMachines/SMF/blob/aee494fe17322c24c2076cc084b3f4559e61db5b/Sources/RemoveTopic.php#L336

sbulen:
I thought I had tested everything; just retested again... 

The first two items in @Sesquipedalian 's list are more difficult to pin down...  I actually don't think that logic is called from anywhere...  I suspect when the 'modblock' changes were implemented, the older variants were cloned and modified, but never removed.  

_**I may be missing something???  Anybody know how to invoke the logic on line 708/787???**_

Still testing the topic deletion...  (I had thought #7515 would deal with deleted topics, and it will, for everyone _except_ admins & global moderators...)

sbulen:
OK, I addressed the first two items by removing the unused logic.

I confirmed that the 3rd item is already addressed by #7515 .  The benefit in that approach is it will not only address deleted/recycled topics, but also deleted/recycled messages, and board permissions changes as well.  All variants of "now you see it, now you don't"...

What confused me earlier is that there was a delay in displaying the updated counter.  It will only fix the counter when a recount occurs.  Most alert-related activities will trigger a recount, e.g., marking alerts read, or viewing alerts.  (Note also that the alert counter gets cached with other member info, and is impacted by the cache timeout as well...)

(At one point I had a "sledgehammer" approach PR, which would always reset the counter in Load.php.  I opted against that as the counter does correct itself...)

sbulen:

PR #7529, opened 2022-07-22
Cleanup approval alerts after approvals

### LINKED ISSUE #7423: Unread alerts should be deleted when relevant content no longer exists.

sbulen:
Fixes #7423 (last one)

Cleanup alerts that posts need approval after posts have been approved.

The wrinkle here was that the post alerts were not identified by msg ids, but by topic ids, which made it impossible to identify the specific alerts to delete for that message.  This required changing the content of the alert a bit, which in turn, required changing the key of the language string, as it is dynamically built.


Sources/tasks/ApprovePost-Notify.php
```diff
@@ -91,8 +91,8 @@ public function execute()
 					'id_member' => $member,
 					'id_member_started' => $posterOptions['id'],
 					'member_name' => $posterOptions['name'],
-					'content_type' => 'topic',
-					'content_id' => $topicOptions['id'],
+					'content_type' => $type == 'topic' ? 'topic' : 'msg',
```
frandominguezl, line 94: What are the possible values of `$type` here? Seems to me like this ternary is not needed.

sbulen: Here, it might be 'topic' for new topics, or 'reply' for replies to topics.  For new topics, you need to be able to look up the topic alerts later.  For replies, you will need to look up message-specific alerts later.  The content type & id should identify the record the alert references.

Sesquipedalian:
My comment [here](https://github.com/SimpleMachines/SMF/pull/7516#issuecomment-1217173908) also applies to this PR, I'm afraid. 🙁

sbulen:
As in the others - this cannot attempt a delete before the alert exists - because it's driven by a query of existing alerts.

Sesquipedalian:
See https://github.com/SimpleMachines/SMF/pull/7516#issuecomment-1218200073 🙂

sbulen:

Sesquipedalian:
Why the change to `content_type` for unapproved replies? I agree that it would make better sense for it to be `msg` rather than `topic` in such cases, but changing this part of the current behaviour doesn't seem strictly necessary to make this code work. Am I missing something?

Sesquipedalian:
I ask because I'd rather not force the translators to redo the $txt string if we can avoid it.

sbulen:
It's necessary to properly identify the alert.  A "topic reply" that only ids the topic isn't specific enough to uniquely the message that was approved.   

Suzie might have 2 unapproved posts in topic 555.  George may have approved post 1 and Randy approved post 2.  Both posts just put alerts out there for topic 555 - nothing message specific.

It's conceivable to find the alert by doing a preg match against the url stuffed in there...  But that's a hackish workaround.   

In all other alerts, the content ids uniquely id the record.  Here, he messages are flagged as topic replies.

Note I only fixed the ones I had to here.  There are other alerts that use the wrong content identifiers.  All post & post approval related.  Sometimes the content points to topics vs messages; sometimes it labeled as a board when the id is a topic id, etc.  But I only cleaned up what I absolutely had to for this task.

So...  Existing alerts at the time this goes in may not be cleaned properly, since they are not uniquely identified.

The "Mark read" link fixes everything, and resets the counters.  It's magic....

Sesquipedalian:
Makes sense. Thanks.

sbulen:

PR #7530, opened 2022-07-31
Improve some lighthouse issues

jdarwood007:
Its against best practices for us to have anchors without herfs.  So try our best to add # links.  Also indicate to robots to not follow these and in some cases, remove them from being tab selected.
Some anchor links to not have text.  This is against best practices for screen readers.  So use some place holders or aria-labels

I didn't hit all pages.  Just a few as I found them.
If this isn't a PR we want to merge, I will just bundle them up into a package to maintain later.

### DIFF
--- a/Sources/Subs-BoardIndex.php
+++ b/Sources/Subs-BoardIndex.php
@@ -194,9 +194,10 @@ function getBoardIndex($board_index_options)
 					'css_class' => ''
 				);
 
-				$categories[$row_board['id_cat']]['link'] = '<a id="c' . $row_board['id_cat'] . '"></a>' . (!$context['user']['is_guest'] ?
-						'<a href="' . $scripturl . '?action=unread;c=' . $row_board['id_cat'] . '" title="' . sprintf($txt['new_posts_in_category'], $row_board['cat_name']) . '">' . $row_board['cat_name'] . '</a>' :
-						$row_board['cat_name']);
+				if (!$context['user']['is_guest'])
+					$categories[$row_board['id_cat']]['link'] = '<a id="c' . $row_board['id_cat'] . '" href="' . $scripturl . '?action=unread;c=' . $row_board['id_cat'] . '" title="' . sprintf($txt['new_posts_in_category'], $row_board['cat_name']) . '">' . $row_board['cat_name'] . '</a>';
+				else
+					$categories[$row_board['id_cat']]['link'] = '<span id="c' . $row_board['id_cat'] . '">' . $row_board['cat_name'] . '</span>';
 
 			}
 
--- a/Themes/default/GenericMenu.template.php
+++ b/Themes/default/GenericMenu.template.php
@@ -25,7 +25,7 @@ function template_generic_menu_dropdown_above()
 	// Load the menu
 	// Add mobile menu as well
 	echo '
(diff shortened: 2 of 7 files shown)


Sources/Subs-BoardIndex.php
```diff
@@ -194,9 +194,10 @@ function getBoardIndex($board_index_options)
 					'css_class' => ''
 				);
 
-				$categories[$row_board['id_cat']]['link'] = '<a id="c' . $row_board['id_cat'] . '"></a>' . (!$context['user']['is_guest'] ?
-						'<a href="' . $scripturl . '?action=unread;c=' . $row_board['id_cat'] . '" title="' . sprintf($txt['new_posts_in_category'], $row_board['cat_name']) . '">' . $row_board['cat_name'] . '</a>' :
-						$row_board['cat_name']);
+				if (!$context['user']['is_guest'])
+					$categories[$row_board['id_cat']]['link'] = '<a id="c' . $row_board['id_cat'] . '" href="' . $scripturl . '?action=unread;c=' . $row_board['id_cat'] . '" title="' . sprintf($txt['new_posts_in_category'], $row_board['cat_name']) . '">' . $row_board['cat_name'] . '</a>';
+				else
+					$categories[$row_board['id_cat']]['link'] = '<span id="c' . $row_board['id_cat'] . '">' . $row_board['cat_name'] . '</span>';
```
DiegoAndresCortes, line 200: Why not just replace it, instead of using if/else?
```
if ($context['user']['is_guest'])
       $categories[$row_board['id_cat']]['link'] = '<span id="c' . $row_board['id_cat'] . '">' . $row_board['cat_name'] . '</span>';
```

Themes/default/GenericMenu.template.php
```diff
@@ -34,7 +34,7 @@ function template_generic_menu_dropdown_above()
 			<div class="popup_window description">
 				<div class="popup_heading">
 					', sprintf($txt['mobile_generic_menu'], $menu_label), '
-					<a href="javascript:void(0);" class="main_icons hide_popup"></a>
+					<a href="#" class="main_icons hide_popup" rel="nofollow" href="#" tabindex="-1"></a>
```
DiegoAndresCortes: Double href

DiegoAndresCortes (changes requested):
This could be included in a future patch but some changes are not essential.
Next version should definitely have more accessibility features

Sources/Subs-BoardIndex.php
jdarwood007: I'm not sure why it would mater.  The unread action doesn't work for guests, so this just prevents linking to a action that the guests can't see.

DiegoAndresCortes (approved):

jdarwood007:
@DiegoAndresCortes I think this is better handled with SMF 3.0 than trying to patch 2.1.  You want this open as a reference or close it and handle it separately?

DiegoAndresCortes:
Yes indeed, I had forgotten we had this open.

I've linked the PR in my notes so can be closed.
We should still open a forum discussion in upcoming months about accessibility enhancements.

jdarwood007:
Thanks, will close this for 2.1 on the grounds we will be making said changes for 3.0.

PR #7533, opened 2022-08-09
Check on UI side website_title length

### LINKED ISSUE #7532: Database Error: Data too long for column 'website_title' at row 1
#### Description
When a member edits their profile to enter their website title, the input allows more characters than the database column will allow.

### Steps to reproduce
1. Edit a user profile on SMF 2.1.2
2. Enter a very long entry for the website title input 

- 2.1.2 stable
- mysqlnd
- 8.1.3
- PHP 7.4 & PHP 8.0

Can be patched in Subs.php

albertlast:
fixes: #7532

### DIFF
--- a/Sources/Profile-Modify.php
+++ b/Sources/Profile-Modify.php
@@ -644,7 +644,7 @@ function loadProfileFields($force_reload = false)
 			'link_with' => 'website',
 			'input_validate' => function(&$value) use ($smcFunc)
 			{
-				if ($smcFunc['strlen']($value) > 250)
+				if (mb_strlen($value) > 250)
 					return 'website_title_too_long';
 
 				return true;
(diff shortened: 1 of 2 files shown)


Sources/Profile-Modify.php
```diff
@@ -642,6 +642,13 @@ function $preload:		A function that is used to load data required for this eleme
 			'size' => 50,
 			'permission' => 'profile_website',
 			'link_with' => 'website',
+			'input_validate' => function(&$value) use ($smcFunc)
+			{
+				if ($smcFunc['strlen']($value) > 250)
```
Sesquipedalian: This should use `mb_strlen($value)` rather than `$smcFunc['strlen']($value)`. This is because `$smcFunc['strlen']()` counts HTML entities like `&amp;` as single characters, but MySQL will count `&amp;` as five characters.

Sesquipedalian (changes requested):

Sesquipedalian (approved):

PR #7538, opened 2022-08-15
Needs a string literal, not a boolean, to work properly in CrowdIn.

### LINKED ISSUE #7537: RTL Crowdin issue
Crowdin does not support boolean in the language strings.  So this code here has issues:
```
// Character set and right to left?
$txt['lang_rtl'] = false;
```

Crowdin wants to make it a string of 'false'.

We can use '0' and '1' as PHP loosely  will handle making these false/true.  But we need to test that using it like this doesn't affect any other code where the conversion happens and if needed, add fixes.

We can have the build script for the download language packs move this back to false/true as well.  But having it able to be supported here works better for translators needing to test language files directly from Crowdin.

sbulen:
Fixes #7537

Note that before the change, the string is not even available for translation at all.  Thus, all forums are stuck in "ltr".

**_BEFORE_** screenshots - the field doesn't appear for translation at all:
image

**_AFTER_** screenshots.  The field is now available for translation (it's the "0").  Conducting an end-to-end test, exporting the language from CI & importing it into a test forum, you can see the theme properly displaying as rtl for Arabic:
image

This shows the setting translated in 2 languages in the test CrowdIn environment:
rtl-crowdin

And this is upon loading the language into SMF:
rtl-testsite

### DIFF
--- a/Sources/ManageLanguages.php
+++ b/Sources/ManageLanguages.php
@@ -991,7 +991,7 @@ function ModifyLanguage()
 	}
 
 	// Saving primary settings?
-	$primary_settings = array('native_name' => 'string', 'lang_character_set' => 'string', 'lang_locale' => 'string', 'lang_rtl' => 'bool', 'lang_dictionary' => 'string', 'lang_recaptcha' => 'string');
+	$primary_settings = array('native_name' => 'string', 'lang_character_set' => 'string', 'lang_locale' => 'string', 'lang_rtl' => 'string', 'lang_dictionary' => 'string', 'lang_recaptcha' => 'string');
 	$madeSave = false;
 	if (!empty($_POST['save_main']) && !$current_file)
 	{
--- a/Themes/default/languages/Help.english.php
+++ b/Themes/default/languages/Help.english.php
@@ -651,7 +651,7 @@ $helptxt['languages_txt_for_email_templates'] = 'These language entries are used
 $helptxt['languages_native_name'] = 'The language\'s name for itself, represented in its own script.';
 $helptxt['languages_character_set'] = 'The character encoding used for this language. As of SMF 2.1, this should always be "UTF-8".';
 $helptxt['languages_locale'] = 'The locale code is used to determine various formatting conventions, etc.<br><br>The locale code typically takes the form of a two character language code, optionally followed by an underscore and a two character country code. For example, <code>en</code> would identify generic English, while <code>en_AU</code> would identify Australian English in particular.';
-$helptxt['languages_rtl'] = 'Check this box if this language reads from right to left (e.g. Hebrew, Arabic).';
+$helptxt['languages_rtl'] = 'Enter a "1" here if this language reads from right to left (e.g. Hebrew, Arabic).  Leave it "0" for left to right languages.';
 $helptxt['languages_dictionary'] = 'This defines the main language dictionary that will be used by the <a href="https://php.net/function.pspell-new" target="_blank" rel="noopener">pSpell</a> spellchecker (if installed).';
 $helptxt['languages_recaptcha'] = 'The language code to use for the reCAPTCHA verification widget.<br><br>See Google\'s reCAPTCHA documentation for its <a href="https://developers.google.com/recaptcha/docs/language" target="_blank" rel="noopener" class="bbc_link">list of supported languages</a>.';
 // The GDPR page of the EU exists in several languages; change the language code at the end of the url
--- a/Themes/default/languages/Install.english.php
+++ b/Themes/default/languages/Install.english.php
@@ -3,7 +3,7 @@
 
 // These should be the same as those in index.language.php.
 $txt['lang_character_set'] = 'UTF-8';
-$txt['lang_rtl'] = false;
+$txt['lang_rtl'] = '0';
 
 $txt['install_step_welcome'] = 'Welcome';
 $txt['install_step_writable'] = 'Writable check';
--- a/Themes/default/languages/index.english.php
+++ b/Themes/default/languages/index.english.php
@@ -15,8 +15,8 @@ $txt['lang_recaptcha'] = 'en';
 
 // Ensure you remember to use uppercase for character set strings.
 $txt['lang_character_set'] = 'UTF-8';
-// Character set and right to left?
-$txt['lang_rtl'] = false;
+// Character set right to left?  0 = ltr; 1 = rtl
+$txt['lang_rtl'] = '0';
 // Number format.
 $txt['number_format'] = '1,234.00';
 
--- a/other/install.php
+++ b/other/install.php
@@ -1942,14 +1942,14 @@ function template_install_above()
 	global $incontext, $txt, $installurl;
 
 	echo '<!DOCTYPE html>
-<html', $txt['lang_rtl'] == true ? ' dir="rtl"' : '', '>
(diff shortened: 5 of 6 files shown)


jdarwood007:
```
./other/install.php:1945:<html', $txt['lang_rtl'] == true ? ' dir="rtl"' : '', '>
./other/install.php:1952:	', $txt['lang_rtl'] == true ? '<link rel="stylesheet" href="Themes/default/css/rtl.css">' : '', '
./other/upgrade.php:3702:					content.write(\'<html', $txt['lang_rtl'] == true ? ' dir="rtl"' : '', '>\n\t<head>\n\t\t<meta name="robots" content="noindex">\n\t\t\');
./other/upgrade.php:3783:<html', $txt['lang_rtl'] == true ? ' dir="rtl"' : '', '>
./other/upgrade.php:3790:	', $txt['lang_rtl'] == true ? '<link rel="stylesheet" href="' . $settings['default_theme_url'] . '/css/rtl.css">' : '', '
./Sources/Load.php:2506:	$context['right_to_left'] = !empty($txt['lang_rtl']);
./Sources/ScheduledTasks.php:1047:	$context['right_to_left'] = !empty($txt['lang_rtl']);
./Sources/ManageLanguages.php:994:	$primary_settings = array('native_name' => 'string', 'lang_character_set' => 'string', 'lang_locale' => 'string', 'lang_rtl' => 'bool', 'lang_dictionary' => 'string', 'lang_recaptcha' => 'string');
./Themes/default/languages/index.english.php:19:$txt['lang_rtl'] = false;
./Themes/default/languages/Install.english.php:6:$txt['lang_rtl'] = false;
```

Did we make sure the checks work to do the dir and the css file as expected?
We could cleanup the checks and update all the checks to use $context['right_to_left'] instead.

The ManageLanguages.php change needs fixed as we have a built in language editor.

live627:
PHP does type juggling for  loose comparisons, which those are.

albertlast:
did you check if still this is working?
https://github.com/SimpleMachines/SMF/blob/6d8394d172041ca1bec3ba249f922462ace8bc78/Sources/ManageLanguages.php#L994

albertlast:
in load.php the var $context['right_to_left'] gets defined with the value,
so also on php the use of this var had to be checked.

jdarwood007:
I think it would be beneficial to change all uses of `$txt['lang_rtl'] == true` to `!empty($context['right_to_left'])`.  Then that way the code in Load.php can handle the comparison safely.

The code in Load.php can then be made to do this:
```
$context['right_to_left'] = !empty($txt['lang_rtl']) && $txt['lang_rtl'] !== 'false';
```

That way it should handle making it RTL for 1, '1', true and 'true'.  All others should make it false.

sbulen:
I don't think I see anything unsafe about the comparison...  Plus, it tests perfectly.  Not sure any further changes are necessary.

Do you have a test case that is failing?

jdarwood007:
I do not.  I would rather we didn't use loose type hinting if possible, but I am not going to push it for the 2.1 release to fix.  For a future SMF release we should correct this.

sbulen:
Loose types are used throughout SMF.  Anywhere we use == instead of ===.   Also in many places we use empty().  (Though I'm aware some academic type folks want to do away with empty() for that very reason...)   It's a core feature of PHP.

Sesquipedalian:
> It's a core feature of PHP.

Indeed.

Sesquipedalian:
If I understand the issue correctly, the problem is that Crowdin is changing `$txt['lang_rtl'] = true;` to `$txt['lang_rtl'] = 'true';` when we export the language packs.

If so, I suggest that we do something like @jdarwood007's idea, but even simpler. Early in `loadLanguage()` do this:

```
$txt['lang_rtl'] = $txt['lang_rtl'] == 'false' ? false : !empty($txt['lang_rtl']);
```

I believe this is the cleanest and least invasive solution possible. On the one hand, we won't need to make any other changes to SMF's codebase. On the other hand, we also won't need to change the existing values in the language files, nor will we need to futz around with Crowdin's behaviour.

sbulen:
The problem is that CrowdIn won't let them translate the boolean at all.  So they cannot specify a language as rtl in Crowdin, not even as 'false' or 'true'.  The field in CI cannot be updated.

**_ALL languages are stuck in ltr..._**

The vendor suggested using '0' and '1', which works.  More discussion & info is on the internal thread (topic 583277).

sbulen:
I added screenshots of before/after behavior, both in CrowdIn and in SMF, above.  Hope this helps.

Back to @jdarwood007 's initial note above, I did not test installs & upgrades in an rtl language, however.

Sesquipedalian:
Thanks for that clarification, @sbulen. Carry on! 🙂

Sesquipedalian:
Based on @jdarwood007's list above, I think the following additional changes are necessary to cover everything we need:

1. Change `false` to `"0"` here as well: https://github.com/SimpleMachines/SMF/blob/aee494fe17322c24c2076cc084b3f4559e61db5b/Themes/default/languages/Install.english.php#L6
2. Change `'true'` and `'false'` to `"1"` and `"0"` on this line: https://github.com/SimpleMachines/SMF/blob/aee494fe17322c24c2076cc084b3f4559e61db5b/Sources/ManageLanguages.php#L1007
3. (Optional) In Load.php's `loadLanguage()` and in install.php's and upgrade.php's `load_lang_file()`, explicitly cast `$txt['lang_rtl']` to boolean. We don't need to do this for our own code, but just in case some mod uses strict comparison on it somewhere, it wouldn't hurt.

jdarwood007:
FYI, while PHP is loosely typed system.  They have been lately tightening that up especially with internal functions.  While I don't see them breaking this, we should work towards being more strict about our type handling and consistent.  PHP 8.0 min requirements in the future will greatly help with this.

sbulen:
OK, requested changes made.  Installer, upgrader & app tested in arabic (after manually switching it to rtl, since the changes aren't in CrowdIn yet...).  Tested under php 8.1.2, mysql 8.0.20.

Note that for the change in ManageLanguages.php, I changed the datatype of the rtl string to string, rather than redefine what 'bool' meant...  I felt that was more logical.

**_Install:_**
rtl-arabic-install

**_Upgrade:_**
rtl-arabic-upgrade
rtl-arabic-upgrade-2

sbulen:

Sesquipedalian:
I don't think that approach to ManageLanguages.php will work. The UI presents a checkbox, so treating the input as a string won't work. We need to write either '0' or '1' depending on whether the checkbox was ticked or not.

sbulen:
Yep...

Why on earth would someone edit the strings there...  Never saw those forms before.  I'd almost prefer to kill those forms, tbh.

jdarwood007:
Well for our released product, we have it.  Maybe for the future, we drop it in favor of supporting Crowdins in place editing UI?
This of course doesn't work well for customizations.  I wonder if we could somehow allow them to be supported.

sbulen:
It turns out that the form adapts to the field type.  Since it was changed to string in the code, it prompts for a string now, not a checkbox.

I do, however, need to update the help text, which uses the term 'checkbox'...

image

sbulen:
(These forms have a couple big bugs...  E.g., some of the fields shown above, when edited, exclusively update the english language files, not the language being displayed...  I will write that up separately...)

sbulen:
I believe this one is ready now.  Again! 😆

PR #7544, opened 2022-09-11
Use template_css in popups/new windows

DiegoAndresCortes:
As discussed previously, this will prevent unexpected situations where a custom theme will load these subtemplates with empty styles.


sbulen (approved):
LGTM

PR #7545, opened 2022-09-14
fix: cors_headers implode

MissAllSunday:
If we enable "Allow CORS (Cross Origin Resource Sharing" setting and setup a custom header on "Additional CORS headers" SMF is appending directly the hardcoded header "X_SMF_AJAX" to the first custom header, resulting on a wrong Access-Control-Allow-Headers value being sent and thus the CORS check is never passed.

Screen Shot 2022-09-13 at 19 47 17

### DIFF
--- a/Sources/Security.php
+++ b/Sources/Security.php
@@ -1499,7 +1499,7 @@ function corsPolicyHeader($set_header = true)
 		foreach ($cors_headers as &$ch)
 			$ch = str_replace(' ', '-', trim($ch));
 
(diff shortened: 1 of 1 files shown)


jdarwood007:
Sounds like something I did..

live627:
❤️ conventional commits!

sbulen (approved):
LGTM...

sbulen:
Thanks @MissAllSunday !

PR #6908, opened 2021-08-03
fix: Hide unwatched topics from the recent posts pages

live627:
Move the unwatched predicates from `LEFT JOIN`ed tables to the `WHERE` clause to ensure that the value is picked up as intended. Use `COALESCE` for when the row/value does not exist.

Fix #6907

Signed-off-by: John Rayes <live627@gmail.com>

PR #7548, opened 2022-09-25
Unwatched fix

### LINKED ISSUE #7402: Not Watching vs Unread topics
#### Description
Arantor reported it on the Bug Reports board:
https://www.simplemachines.org/community/index.php?topic=581574.0

> There are topics on this site I don't care to read - some mod specific topics for example. I've marked them 'not watching' and they don't show up in the regular unread topics. And they don't show up in updated topics, which is ideal.
> 
> But they show up in all unread because they're unread.
> 
> What I'm getting at is that all unread should respect unwatched the way regular unread does.

sbulen:
This extends pr #6908 logic to the temp table (log_topics_unread) used during "ALL UNREAD TOPICS" processing as well.  #6908 fixed "Recent Unread Topics", but not "All Unread Topics".  

This addresses the issue from the forum:  https://www.simplemachines.org/community/index.php?topic=582135.0

The bottom line is that when determining if something is unwatched, you have to have the unwatched field accessible to the query...  It's a little counterintuitive here, though, since the table is normally telling you what you are watching.  BUT...  That same table also tells you when you are explicitly not watching something also...

Note that the temp table log_topics_unread is only used when doing an "ALL UNREAD TOPICS", not when doing an "RECENT UNREAD TOPICS"...  For the issue to be visible, you need to click on "Unread Posts" then subsequently click on the "ALL UNREAD TOPICS" button.  

Initial testing looks good.

@live627 , I'd appreciate your input.

Also...  Fixes #7402

### DIFF
--- a/Sources/Recent.php
+++ b/Sources/Recent.php
@@ -824,13 +824,13 @@ function UnreadTopics()
 			CREATE TEMPORARY TABLE {db_prefix}log_topics_unread (
 				PRIMARY KEY (id_topic)
 			)
-			SELECT lt.id_topic, lt.id_msg
+			SELECT lt.id_topic, lt.id_msg, lt.unwatched
 			FROM {db_prefix}topics AS t
 				INNER JOIN {db_prefix}log_topics AS lt ON (lt.id_topic = t.id_topic)
 			WHERE lt.id_member = {int:current_member}
 				AND t.' . $query_this_board . (empty($earliest_msg) ? '' : '
 				AND t.id_last_msg > {int:earliest_msg}') . ($modSettings['postmod_active'] ? '
-				AND t.approved = {int:is_approved}' : '') . ' AND lt.unwatched != 1',
+				AND t.approved = {int:is_approved}' : ''),
 			array_merge($query_parameters, array(
 				'current_member' => $user_info['id'],
 				'earliest_msg' => !empty($earliest_msg) ? $earliest_msg : 0,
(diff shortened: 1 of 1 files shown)


Bopske:
Tested the fix on my forum and this solves the problem I have been having.

Sources/Recent.php
```diff
@@ -911,6 +912,7 @@ function UnreadTopics()
 				AND t.id_last_msg >= {int:min_message}
 				AND COALESCE(lt.id_msg, lmr.id_msg, 0) < t.id_last_msg' . ($modSettings['postmod_active'] ? '
 				AND ms.approved = {int:is_approved}' : '') . '
+				AND COALESCE(lt.unwatched, 0) != 1
```
albertlast, line 915: possible performance degradation,
someone with a big dataset/setup should test this coalesce approache.

live627: Why? Concerned that it won't use the index?

albertlast: this concerns didn't exists,
since the awnser is known -> it can't use a index.
the question is,
do it matter that no index can be used.

live627: There was never an index on this column to begin with.

live627 (approved):
Tested and fixes the issue.

Sources/Recent.php
albertlast: yea i know, but still someone had to test if the new filter had somekind of negativ impact,
in this cases where filter didn't happen before.

sbulen:
Thanks @live627 !

This fix was tested on two pretty large forums - one with ~245K messages and one with ~300K messages.  No performance issues.  I think we're good.

I do think it's possible to get a performance hit, but only under unusual circumstances, e.g., a user with 100K+ unread messages and thousands of watched/unwatched topics.  But that would be a real edge condition, and even then I suspect the hit would be minor.

albertlast:
instead of using:
```sql
AND COALESCE(lt.unwatched, 0) != 1
```
to use:
```sql
AND (lt.unwatched != 1 or lt.unwatched is null)
```
would allow the database to had a better understand what you request and could better optimize the query when needed.

PR #7550, opened 2022-09-28
Updates time zone data to match current TZDB

Sesquipedalian:
Release 2022b of the TZDB includes a rename of `Europe/Kiev` to `Europe/Kyiv`. This PR updates SMF to match.

See https://data.iana.org/time-zones/tzdb/NEWS

Also expands our coverage for time zone fallbacks. This helps to improve time zone support on forums running on older versions of PHP and therefore of the time zone database.


sbulen:
@Sesquipedalian - is this one ready for review?  (I see you added some changes.)

Sesquipedalian:
Actually, I think a few fixes are still needed. My previous testing missed some things. I'll @ you when it is ready. 🙂

Sesquipedalian:
@sbulen, this is ready for review now. I suggest testing on PHP 7.0.0 (or the earliest version after 7.0.0 that you have available to you), since the relevant code will only kick in when using an outdated version of the TZDB.

sbulen (approved):
LGTM.

Tested on php 7.1/mysql 5.7.   Kyiv would come up as EEST/EET.

(For some reason couldn't get my php7.0.33 env up & running...🙄)

PR #7554, opened 2022-10-01
MySQL remove NO_AUTO_CREATE_USER

### LINKED ISSUE #7552: MySQL 8+ - SQL_MODE not getting properly set
#### Description
Noticed that in my MySQL 8+ test environments, SQL_MODE wasn't getting set...

At issue is this line:
https://github.com/SimpleMachines/SMF/blob/7401bafa89dcea72cdcd3da60372f58d0e2031b2/Sources/Subs-Db-mysql.php#L108

If run manually, I get the error: "1231 - Variable 'sql_mode' can't be set to the value of 'NO_AUTO_CREATE_USER'"

This appears to be a new edit in 8.0; this works fine in my 5.7 environment.  Since the response isn't getting checked, it is failing silently.  The SQL_MODE isn't getting set at all; it will remain set to whatever the environment default is.  Thus, users may see different symptoms/behavior (as in #7508 - some folks saw ONLY_FULL_GROUP_BY behavior & some didn't...).

- Database Type: mysql
- Database Version: 8.0+
- PHP Version: any

Found this testing #7517 & #7508

albertlast:
Since the setting create issues #7552 and
only define the bahavior of replication: https://stackoverflow.com/questions/71110454/no-auto-create-user-in-mysql-8-0
this should be dropped.

fixes: #7552


sbulen (approved):
Thanks!

PR #7557, opened 2022-10-04
PHP 8.1 doesn't like implicit float to int conversions

### LINKED ISSUE #7556: 8192: Implicit conversion from float ###.#### to int loses precision
#### Description
I saw 16 of these errors after an upgrade:
8192: Implicit conversion from float to int loses precision
yada/yada/Subs-Charset.php (Line 454)

This appears to be a php 8.1 issue.

### Steps to reproduce
1. Saw this after a test upgrade using current code
2. It appears to have been reloading the tld regex

Note that I can also reproduce running the Weekly Maintenance scheduled task, which results in the same 16 errors.

- Database Type: mysql
- Database Version: 8.0
- PHP Version: 8.1

image

image

sbulen:
Fixes #7556 

Explicit int recast gets rid of the errors.  Screenshots confirm the correct letters are extracted.

**_Dump of the internal data - BEFORE:_**
image

**_Dump of the internal data - AFTER:_**
image

### DIFF
--- a/Sources/Subs-Charset.php
+++ b/Sources/Subs-Charset.php
@@ -440,6 +440,7 @@ function utf8_decompose($chars, $compatibility = false)
 	for ($i=0; $i < count($chars); $i++)
 	{
 		// Hangul characters.
+		// http://unicode.org/L2/L2006/06310-hangul-decompose9.pdf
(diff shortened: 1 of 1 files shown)


Sesquipedalian (approved):

sbulen:
Thanks, @Sesquipedalian !

PR #7560, opened 2022-10-12
PHP 8 email line break change

sbulen:
Emails were not getting sent properly due to malformed headers under the following circumstances:
 - PHP 8+
 - Unix only (Windows worked OK)
 - Mail type PHP only (SMTP worked OK)

Due to a change in PHP 8, the PHP mail() function now expects line endings to be "\r\n", even on linux systems.  The old SMF logic that changed the line endings per the OS was actually breaking things when using the mail type of PHP on linux.

This fix has been tested on all variants of PHP7.4 vs PHP8.1, windows vs linux, PHP vs SMTP.  

See forum discussion:
https://www.simplemachines.org/community/index.php?topic=583842.0

See PHP 8 bug report & discussion.  Note the bug report was closed w/o action, saying the new behavior is in line with the RFC:
https://github.com/php/php-src/issues/8086

### DIFF
--- a/Sources/Subs-Post.php
+++ b/Sources/Subs-Post.php
@@ -572,8 +572,8 @@ function sendmail($to, $subject, $message, $from = null, $message_id = null, $se
 	// Use sendmail if it's set or if no SMTP server is set.
 	$use_sendmail = empty($modSettings['mail_type']) || $modSettings['smtp_host'] == '';
 
-	// Line breaks need to be \r\n only in windows or for SMTP.
-	$line_break = $context['server']['is_windows'] || !$use_sendmail ? "\r\n" : "\n";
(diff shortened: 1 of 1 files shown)


sbulen:
Opted for a slightly safer solution that only tweaked the behavior for 8x.  

7x is left as-is.  

Although my tests in 7x were OK, it's possible that some other environments really won't like the crlf going to mail().  E.g., in the php bug report above, some users complained that crlf was translated to crcrlf in their systems, since their envs were performing a lf => crlf conversion, resulting in broken headers...  Best to play it safe...

BrickOzp (approved):

sbulen:
Thanks, @BrickOzp !

PR #7564, opened 2022-10-17
Make the hidden members visible in the online list

### LINKED ISSUE #7543: Hidden members hide names for admins
https://github.com/SimpleMachines/SMF/blob/aee494fe17322c24c2076cc084b3f4559e61db5b/Themes/default/Who.template.php#L68

When members are hidden and a admin browses action=who, their username does not show at all, just their IP.

This is because of a .hidden generic css we have: https://github.com/SimpleMachines/SMF/blob/aee494fe17322c24c2076cc084b3f4559e61db5b/Themes/default/css/index.css#L4030-L4032

There is a more specific CSS already: https://github.com/SimpleMachines/SMF/blob/aee494fe17322c24c2076cc084b3f4559e61db5b/Themes/default/css/index.css#L3922-L3924   which was added in #3949

DiegoAndresCortes:
Fixes #7543

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -3920,6 +3920,7 @@ p.information img {
 	margin-top: -1px !important;
 }
 span.member.hidden {
+	display: inline-block;
 	font-style: italic;
 }
 


jdarwood007:
I look at my own suggestion now and think its a poor one.  Maybe it would be better if the template did: `$member['is_hidden'] && !$user_info['is_admin']`?  Implying it would only apply the hidden class to those who are hidden and the user viewing this is not an admin.

DiegoAndresCortes:
Not sure I get that, it already skips the hidden users if you can't see them.

I think it's this condition?
```
if (!allowedTo('moderate_forum'))
		$conditions[] = '(COALESCE(mem.show_online, 1) = 1)';
```

jdarwood007:
I was thinking from the template:
https://github.com/SimpleMachines/SMF/blob/aee494fe17322c24c2076cc084b3f4559e61db5b/Themes/default/Who.template.php#L68

So then it wouldn't apply the hidden css class.

DiegoAndresCortes:
So we would be removing the visual hint that they are hidden?
Also it'd be `!allowedTo('moderate_forum')` because that's what the query uses, so not only admins can see hidden users, apparently.

jdarwood007:
Oh, I guess the hint would be appropriate.  I thought about the fact that we are using a .hidden and it is using a css display property to make it visible.  Ignore and continue.

DiegoAndresCortes:
Alright so I suppose the single commit is the easiest approach to fix this

sbulen (approved):
LGTM

PR #7565, opened 2022-10-17
Only disable bbc when its not a message

### LINKED ISSUE #7404: Setting for signatures affects the `url` BBC in entire posts
#### Description
This was reported on the Bug Reports board (relevant details are on page 4):
https://www.simplemachines.org/community/index.php?topic=580848.60

If you disable `url` BBC, all links in topics stop working and not just in the signature:
Capturar

So signatures should be separate from posts when it comes to the `url` BBC.

DiegoAndresCortes:
Fixes #7404

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -1592,7 +1592,7 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 				);
 		}
 
-		if (!empty($parse_tags))
+		if (!empty($parse_tags) && $message === false)
 		{
 			if (!in_array('email', $parse_tags))
 				$disabled['email'] = true;


jdarwood007 (approved):
Logic looks good, but didn't test.  $message === false, is when we are testing what BBC is available.

DiegoAndresCortes:
These were the results, I posted them somewhere else by mistake.

**Disabled:**
Screenshot_12
Screenshot_11

**Enabled**
Screenshot_14
Screenshot_13

sbulen:
LGTM.  Thanks for the test pix.  Really helps.

Ready to merge?

BrickOzp:
Can you reproduce the original issue if removing my fix?

Otherwise I think your fix looks good.

DiegoAndresCortes:
I didn’t test for it, I assume your fix is still needed because nothing changed prior as far as I could tell.

PR #7567, opened 2022-10-20
Updates Unicode data files to Unicode 15

Sesquipedalian:
Unicode 15 was released on Sep 15, 2022. Files were updated by running `php other/update_unicode_data.php`.

Also breaks up long Unicode regex strings into their constituent parts. Aside from improving code legibility, this should make it easier to construct patch releases in the future, since the patches will only need to modify the bits that changed rather than the entire regex string.


sbulen:
@Sesquipedalian - is this ready?  Has testing been completed?

Sesquipedalian:
~Yes~
Crud. I take that back. Looks like something is up with a few of the normalization tests. Not sure why yet.

sbulen:
Thanks, let me know when it's ready.  

FYI, I sent a PM over on SMF with a question, when you have a moment.

jdarwood007:
Just a note here.  With the amount of updates to this file, we should just have Package Manager replace the file.  It does make it so it can't be undone during uninstall (unless we package it), but it really doesn't make a difference there.

sbulen:
I don't know if Sesq is plugged into that!  Individual diffs in 3 files:  CaseFold.php, CaseLower.php & CaseUpper.php, are so big that pacman chokes on 'em.  The patch installs OK, but on de-install, regex fails hard & reports the pattern exceeds 64k (a php compile-time setting).  I did find a way to chunk them up differently that is working now.  

This particular PR doesn't touch those 3 files.

And yeah, given the nature of these files, I agree, it would be OK for the patch to just copy them over.

Sesquipedalian:
Never mind. There's nothing wrong with this PR. It was just a derp on my part while testing. 

Specifically, I was testing on a version of PHP that has the intl extension enabled and is using an older version of Unicode. Since the utf8_normalize_* functions in Subs-Charset.php prefer to use the intl extension's normalizer_normalize() function when available, rather than doing the work ourselves, the fact that my PHP's intl extension was outdated caused my tests to fail for the new changes Unicode 15. But when I edited Subs-Charset.php to force it to use our own normalization code (so that I was testing our code instead of the intl extension's code), all tests passed with flying colours.

Sesquipedalian:
Regarding patches: Yes, I agree that the data files in ./Sources/Unicode should all just be copied into the directory rather than modified when the patch is installed, and then just left alone if the patch is uninstalled. 

Even if there were no issues with pacman, these files should never be reverted anyway. This is because reverting from Unicode 15 to Unicode 14 in our data files could cause a string that was previously processed and accepted using Unicode 15 to subsequently be rejected after reverting back to Unicode 14. That could cause garbled output and/or other unexpected issues. Generally speaking, Unicode versions should never be rolled back in any context.

sbulen (approved):
LGTM.   Tests look good.

PR #7568, opened 2022-10-21
Check attachment and avatar folders before upgrade

### LINKED ISSUE #7555: Confirm attachment dirs (and avatar dir, & custom avatar dir) exist before upgrade
#### Description
There have been a number of instances where folks lose their avatars & attachments during a 2.1 upgrade.

The only known way to reproduce this at this time is to have incorrectly specified attachment folders.  E.g., this can happen after a forum move, restore, or folder restructure or rename, etc.

We should consider confirming the existence of the folders at the start of the upgrade.  (Attachments, avatars, and, if specified, custom avatars.)  

Issue a warning if they do not.

Another possible consideration is to allow for attachment processing on rerun.  Currently, a flag is set and it will not attempt the processing on subsequent executions of the upgrader (as attachment processing is often the lengthiest part of the upgrade...).  But if folks are flailing away & rerunning the upgrade multiple times, etc., we do not want a situation where attachments & avatars don't get processed the first time thru, and the processing is bypassed the second time thru, making it impossible for the user to properly upgrade attachments after a botched start.  (This might be tricky as the attachment folders are converted from serialized to JSON during the upgrade...)

### Steps to reproduce
1. Start with a 2.0 forum with attachments and _uploaded_ custom avatars.  However, do NOT specify a custom avatar folder (in 2.0, if no custom avatar folder is specified, custom avatars are stored as attachments).  
2. Rename the attachment folder.
3. Run the 2.1 upgrade.  
4. All attachments & custom avatars will be "lost".  The only way to recover is to manually update the files or run a script.

From the forum:
https://www.simplemachines.org/community/index.php?topic=582805.0

sbulen:
Fixes #7555 

If folders are not found, warnings are produced:
image

Also, if we are rerunning the upgrader, a new option has been added to allow for rerunning the attachment conversion:
att_dir_option

### DIFF
--- a/Themes/default/languages/Install.english.php
+++ b/Themes/default/languages/Install.english.php
@@ -271,6 +271,7 @@ $txt['upgrade_customize'] = 'Customize';
 $txt['upgrade_debug_info'] = 'Output extra debugging information.';
 $txt['upgrade_empty_errorlog'] = 'Empty error log before upgrading.';
 $txt['upgrade_delete_karma'] = 'Delete all karma settings and info from the DB';
+$txt['upgrade_reprocess_attachments'] = 'Rerun attachment conversion';
 $txt['upgrade_stats_collection'] = 'Allow Simple Machines to collect basic stats monthly.';
 $txt['upgrade_stats_info'] = 'If enabled, this will allow Simple Machines to visit your site once a month to collect basic statistics. This will help us make decisions as to which configurations to optimise the software for. For more information please visit our <a href="%1$s" target="_blank" rel="noopener">info page</a>.';
 $txt['upgrade_migrate_settings_file'] = 'Migrate to a new Settings file.';
--- a/Themes/default/languages/Install.english.php
+++ b/Themes/default/languages/Install.english.php
@@ -372,9 +372,9 @@ $txt['warning_lang_old'] = 'The language files for your selected language, %1$s,
 $txt['warning_lang_missing'] = 'The upgrader could not find the &quot;Install&quot; language file for your selected language, %1$s. Upgrade will continue with the forum default, %2$s.';
 
 // Attachment & Avatar folder checks
-$txt['warning_av_missing'] = 'Warning!  Avatar directory not found.  Continuing may be unsafe.  Please confirm folder settings before proceeding.';
-$txt['warning_custom_av_missing'] = 'Warning!  Custom avatar directory not found.  Continuing may be unsafe.  Please confirm folder settings before proceeding.';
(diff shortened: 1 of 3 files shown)


sbulen:
Testing looks good.  I'd appreciate a peek before I merge!

other/upgrade.php
```diff
@@ -1061,6 +1064,111 @@ function WelcomeLogin()
...
+				if (!is_dir($dir))
+					$attdr_problem_found = true;
+			}	
+		}
+		else
+		{
+			if (!is_dir($json_test))
+				$attdr_problem_found = true;
+		}
+	}
```
DiegoAndresCortes, line 1134: They all three seem to not depend on each other? It works though

Themes/default/languages/Install.english.php
```diff
@@ -371,6 +371,11 @@
 $txt['warning_lang_old'] = 'The language files for your selected language, %1$s, have not been updated to the latest version. Upgrade will continue with the forum default, %2$s.';
 $txt['warning_lang_missing'] = 'The upgrader could not find the &quot;Install&quot; language file for your selected language, %1$s. Upgrade will continue with the forum default, %2$s.';
 
+// Attachment & Avatar folder checks
+$txt['warning_av_missing'] = 'Warning!  Avatar directory not found.  Continuing may be unsafe.  Please confirm folder settings before proceeding.';
+$txt['warning_custom_av_missing'] = 'Warning!  Custom avatar directory not found.  Continuing may be unsafe.  Please confirm folder settings before proceeding.';
+$txt['warning_att_dir_missing'] = 'Warning!  One or more attachment directories not found.  Continuing may be unsafe.  Please confirm folder settings before proceeding.';
```
DiegoAndresCortes: Are the double spaces intended?

DiegoAndresCortes (commented):
The original issue didn't happen, but only tested one upgrade. I'll try to check again later

other/upgrade.php
sbulen: Not sure the question here.  The string is either plain text, serialized, or json.  Can't be more than one.

The only way I can reproduce the original issue (attachments/avatars not converted) is by starting with an incorrect attachment folder.

Themes/default/languages/Install.english.php
sbulen: YES THEY ARE NECESSARY!!!!

OK, I removed 'em...   You young'uns...

sbulen:
So, to test, you first gotta break the upgrader...  I usually just rename the attachment folder.  This will show you the new warnings.

To test the restart option, you can subsequently fix the attachment folder name then rerun the upgrader.  You should get the new option to rerun the attachment conversion.

sbulen:
Lots of testing completed, I am going to go ahead & merge this.
image

PR #7571, opened 2022-10-22
A few more Unicode-related changes

### LINKED ISSUE #7569: Unicode-related functions should always use the same version of Unicode data
[Recent experience](https://github.com/SimpleMachines/SMF/pull/7567#issuecomment-1286466524) made me realize that it is currently possible for some of our Unicode processing functions to use different versions of Unicode data than other functions. (Specifically, on servers with PHP's intl extension enabled, our normalization functions will use whichever Unicode version that extension was built with, whereas our case conversion and sanitization functions always use the latest Unicode version.) This creates the possibility for inconsistent results. 

To fix this, we need to make sure that we always use a consistent Unicode version for all functions.

One simple way to do that would be to never use the intl extension's Normalizer functions at all. But using the Normalizer functions gives better performance, so it would be better to use a version check rather than getting rid of the option entirely.

### LINKED ISSUE #7570: Unicode normalization performance could be improved
The performance of our Unicode normalization functions could be improved by implementing the Unicode ["Quick Check" algorithm](https://www.unicode.org/reports/tr15/#Detecting_Normalization_Forms).

Sesquipedalian:
Fixes #7569 
Fixes #7570


Sesquipedalian:
Oh, and in case anyone wants to test for themselves: https://gist.github.com/Sesquipedalian/b0870a303967002adc9db93fc48cd919

Sesquipedalian:
FYI, with the latest performance improvements, calls to our normalization functions now return about 20 to 50 times faster than they used to when passed strings that were already normalized. Since in the great majority of cases, the strings will indeed already have been normalized, this is a significant improvement.

jdarwood007:
I think we should tag these for 2.1.4.  2.1.3 is already large enough as it is and I'm not versed enough in the charset stuff to feel confident of any BC issues.  @Sesquipedalian Is that an ok assessment here?

sbulen:
I think this might be related to some of this work?  https://github.com/SimpleMachines/SMF/issues/7572

jdarwood007:
ok, to fix that error, I guess.

Sesquipedalian:
I'm not sure what is going on in #7572, but it looks like the problem in that issue is that files are somehow ending up in an inconsistent state. This PR is not intended to deal with whatever is going on with that.

Leaving that aside, I still would not agree with @jdarwood007's suggestion about this PR.

At the very least, the changes in [74afc97](https://github.com/SimpleMachines/SMF/pull/7571/commits/74afc97ed0703723532bcfcaea944ad569fc697f) should be included as soon as possible, since they address a bug (even if it is a bug that we may not have received complaints about yet, since English speaking people are most likely to encounter it only when using recently released emojis and therefore are most likely to disregard it as just one of those things that happens when using cutting edge tech).

Regarding the `utf8_is_normalized()` function and the supporting regex data that has been added to ./Sources/Unicode/RegularExpressions.php, well, I suppose it *could* be left out, but I see absolutely no reason why we would do that. If I understand correctly, it has already been decided that the ./Sources/Unicode/*.php files will simply be overwritten with new copies by the 2.1.3 patch, rather than modified in place. With that being the case, it makes no difference whether we have this additional supporting data in the file or not.

Of course, it is up to you fine folks to decide about this. But that's my input on the matter. 🙂

Sesquipedalian:
Another thought: 

If people really are concerned about including the Unicode data files in the patch packages, I suppose that we could try writing up a background task to upgrade them instead. This would essentially mean wrapping up the code in ./other/update_unicode_data.php into a background task class and running it as part of weekly maintenance. Even if we did that, though, we would still want to have the latest data in our install and upgrade packages, so merging this PR would still be the right course of action.

But honestly, I don't think this alternative would be worthwhile. The Unicode database is updated annually (usually in September). I don't think the effort involved in creating and testing a new background task is worthwhile in order to avoid including 168 kilobytes of compressed data in one patch file per year.

sbulen:
I was ultimately able to build packages that worked.

I think we can and should continue to use pacman for all the code changes.  Any change may introduce bugs or cause some form of env conflict for some users.  Folks should be able to revert should issues arise.

Building the package was problematic here due to the size of some of the changes in the unicode PRs.   PHP itself has some compiled-in limits that were exceeded by the search/replaces.  Chunking up the updates to work around those limits was tricky, due in part to some pacman quirks.  But we are past that issue.

Sesquipedalian:
Good to hear.

Still, I do strongly encourage that the changes in [74afc97](https://github.com/SimpleMachines/SMF/pull/7571/commits/74afc97ed0703723532bcfcaea944ad569fc697f) be included in 2.1.3, even if you decide not to include other changes in this PR until later. If necessary, I can split this PR into two.

Sesquipedalian:
You know what? Since this PR is addressing two different issues, I really ought to have broken it up into two different PRs anyway. I will do that now.

Sesquipedalian:
See #7573 for the version consistency part that I strongly recommend for inclusion in 2.1.3.
See #7574 for the performance improvement part, which I also recommend for 2.1.3, but could be optional if people really object.

PR #7573, opened 2022-10-24
Ensures normalization always uses consistent Unicode version

### LINKED ISSUE #7569: Unicode-related functions should always use the same version of Unicode data

Sesquipedalian:
Fixes #7569

Unit test: https://gist.github.com/Sesquipedalian/b0870a303967002adc9db93fc48cd919

### DIFF
--- a/Sources/Subs-Charset.php
+++ b/Sources/Subs-Charset.php
@@ -14,6 +14,8 @@
 if (!defined('SMF'))
 	die('No direct access...');
 
+require_once($sourcedir . '/Unicode/Metadata.php');
+
 /**
  * Converts the given UTF-8 string into lowercase.
  * Equivalent to mb_strtolower($string, 'UTF-8'), except that we can keep the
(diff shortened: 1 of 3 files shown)


sbulen (approved):
LGTM.

PR #7574, opened 2022-10-24
Improves performance in normalization polyfills

### LINKED ISSUE #7570: Unicode normalization performance could be improved

Sesquipedalian:
Fixes #7570

Unit test: https://gist.github.com/Sesquipedalian/b0870a303967002adc9db93fc48cd919


Sesquipedalian:
Any chance that this could be merged soon? It's not a big deal, but it would help me keep my local dev environment tidy.

jdarwood007:
@Sesquipedalian This is a fairly big looking commit, although not much changed in existing files.  What is the risk if this is included in a patch releases?

Sesquipedalian:
Negligible risk. These changes simply perform some tests before executing the full Unicode normalization process on a string, and skips the full process if the string is already normalized. 

You can use the unit test that I provided to confirm this. It runs the complete suite of normalization tests provided by the Unicode Consortium itself.

Sesquipedalian:
If you're interested in more information on how these tests work, see https://unicode.org/reports/tr15/#Detecting_Normalization_Forms.

jdarwood007:
I think thats good enough to say it should be safe for release.  Along with the unit test, it proves far as we know it shouldn't break anything.

jdarwood007 (approved):

PR #7578, opened 2022-10-26
Avoids partial domain name matches in autolinker

### LINKED ISSUE #7576: URL parser doesn't recognize some .example URLs
#### Description
If I [post](https://www.simplemachines.org/community/index.php?topic=583784.msg4137766#msg4137766) something like "`https://forum.my.example`", the auto-linker writes this code:
```
<a href="https://forum.my" class="bbc_link" target="_blank" rel="noopener">https://forum.my</a>.example
```

I tried to reproduce this with other obscure/long TLDs (like `.blackfriday`), but only `.example` seems to be affected.

These work (the final slash, added by my Chromium browser during c&p, indicates where the autolinker ends the link):
* https://foo.example/
* https://www.example/

These don't work (the final slash, added by my Chromium browser during c&p, indicates where the autolinker ends the link):
* https://foo.bar/.example
* https://www.foo/.example
* https://www.foo.bar/.example

### Steps to reproduce
1. See above

- Version/Git revision: Whatever currently runs on SMorg
- Database Type: Whatever currently runs on SMorg
- Database Version: Whatever currently runs on SMorg
- PHP Version: Whatever currently runs on SMorg

Sesquipedalian:
Fixes #7576

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -2887,7 +2887,7 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 
 							$pcre_subroutines['bracket_quote'] = '[' . $bracket_quote_chars . ']|&' . build_regex($bracket_quote_entities, '~');
 							$pcre_subroutines['allowed_entities'] = '&(?!' . build_regex(array_merge($bracket_quote_entities, array('lt;', 'gt;')), '~') . ')';
-							$pcre_subroutines['excluded_lookahead'] = '(?![' . $excluded_trailing_chars . ']*(?>[\h\v]|<br>|$))';
+							$pcre_subroutines['excluded_lookahead'] = '(?![' . $excluded_trailing_chars . ']*(?' . '>[\h\v]|<br>|$))';
 
 							foreach (array('path', 'query', 'fragment') as $part)
 							{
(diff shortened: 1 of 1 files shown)


BrickOzp (approved):

Sesquipedalian:
You might want to review again, @BrickOzp. Upon further consideration, I realized that the correct approach was (1) to recognize both `example` and `invalid` as known TLDs, but (2) to *not* autolink URLs with those TLDs, because they are intentionally forbidden from resolving as live URLs.

BrickOzp (approved):

Sesquipedalian:
Hopefully, the third time's the charm. 

I wasn't happy with my previous approach of explicitly detecting and then rejecting `example` and `invalid`, because we would still end up with partial domain name matches in cases like `http://www.com.urboiuwr/derp.txt`. This would get transformed into `<a href="http://www.com">http://www.com</a>.urboiuwr/derp.txt`, effectively cutting the URL in two and linking only the first part. This is the same undesirable behaviour as described in the original bug report, so merely making special cases for `example` and `invalid` clearly isn't enough to solve the issue in the more general case.

With my latest push, I've reverted my previous attempts and instead fixed two subtle mistakes in the autolinker's regular expression. With this change, we no longer get partial domain name matches. Now `http://www.com.urboiuwr/derp.txt` in its entirety will fail to match, even though the substring `http://www.com` would be a valid URL.

Importantly, I also tested this against my full suite of other URL tests, and everything else continues to work as expected. So I am finally certain that this PR is done.

PR #7580, opened 2022-10-27
Automate updating of version numbers & years in .js, .css, and LICENCE files, too.

Sesquipedalian:
Previously, update_version_numbers.php only updated PHP files.

This PR also makes the script more friendly to Windows.

----------

For those who aren't familiar with it, update_version_numbers.php is a file I added when SMF 2.1.2 was released. Running it from the command line will:
1. Update the version number and copyright year in the licence block of index.php, cron.php, proxy.php, SSI.php, install.php, upgrade.php, upgrade-helper.php, Settings.php, Settings_bak.php, and every other file that has changed since the last tagged release.
2. Update the version number in every language file that has changed since the last tagged release.
3. Update the SMF_VERSION and SMF_SOFTWARE_YEAR constants in index.php, cron.php, proxy.php, SSI.php, install.php, and upgrade.php.
4. Update SMF_LANG_VERSION in upgrade.php to make sure it matches the version number in index.english.php.
5. Update the copyright year in the LICENCE file.

### DIFF
--- a/other/update_version_numbers.php
+++ b/other/update_version_numbers.php
@@ -37,18 +37,25 @@ $release_branch = 'release-2.1';
 // The root directory of the working tree for this Git repository
 $basedir = dirname(__DIR__);
 
+// Make sure we are working in the right directory
+chdir($basedir);
+
 // Do nothing if on wrong branch or working tree is dirty
-if (trim(shell_exec('cd "' . $basedir . '"; if [[ "$(git rev-parse --abbrev-ref HEAD)" != "' . $release_branch . '" || -n "$(git status --porcelain)" ]]; then echo "abort"; else echo; fi')) !== '')
-{
-	die('Could not continue. Wrong branch or dirty working tree.');
-}
+if (trim((string) shell_exec('git status --porcelain')) !== '')
+	die('Could not continue. Dirty working tree.');
+
+if (trim(shell_exec('git rev-parse --abbrev-ref HEAD')) !== $release_branch)
+	shell_exec('git checkout "' . $release_branch . '"');
+
+if (trim(shell_exec('git rev-parse --abbrev-ref HEAD')) !== $release_branch)
+	die('Could not continue. Wrong branch is checked out.');
 
 // Matches all standard SMF version strings
 $version_regex = '\d+\.\d+[. ]?(?:(?:(?<= )(?>RC|Beta |Alpha ))?\d+)?';
 
 // Get previous version based on the most recent Git tag
 // This assumes we are using proper sematic versioning in our tags
-$prev_version = ltrim(trim(shell_exec('cd "' . $basedir . '"; git checkout "' . $release_branch . '" &>/dev/null; git describe --tags --abbrev=0')), 'v');
+$prev_version = ltrim(trim(shell_exec('git describe --tags --abbrev=0')), 'v');
 
 // Was the new version passed as a command line argument?
 if (!empty($argv[1]))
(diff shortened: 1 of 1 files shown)


sbulen:
Yeah, it didn't work at all for me!  I had to hack it up a bit...  😂

Thanks,

Sesquipedalian:
Yeah, the previous version of this file assumed a Unix or Linux shell was available for the `shell_exec()` commands. I think this version will work on Windows without issue. I got rid of all shell commands except simple `git` commands, and I believe `git` takes all the same options and arguments on Windows as it does on *nix. But if you still encounter any problems on Windows, just let me know. I'll be glad to fix them.

Sesquipedalian:
By the way, when I run update_version_numbers.php against [cb154e6](https://github.com/SimpleMachines/SMF/commit/cb154e6d4c2b37176f0ef7dfa3478a95cdea1efa), it finds ten more files that still need to have their version numbers updated to 2.1.3.

sbulen:
Language files?

I was deliberately leaving those out, so folks can test - even upgrades - using their languages.  Was going to do a 2nd pass for languages after tests.

Sesquipedalian:
> Language files?
> 
> I was deliberately leaving those out, so folks can test - even upgrades - using their languages. Was going to do a 2nd pass for languages after tests.

Fair enough. 🙂

Sesquipedalian:
Any chance that this could be merged soon? It's not a big deal, but it would help me keep my local dev environment tidy.

sbulen:
@Sesquipedalian - this works great, though I have one issue with the approach.

This script seems to want to update the default branch.  I'd strongly prefer it to allow me to work in a new feature branch, so the changes can be submitted just like any PR.  I can work around it, but it seems like a non-standard workflow.

Sesquipedalian:
Well, you can change the value of `$release_branch` to whatever you like locally, @sbulen. So you could create yourself a local branch and use that if you prefer the more roundabout workflow. But I honestly don't see the point in this case. You'll just be merging those changes into the release branch immediately afterwards, so why bother with the extra steps?

jdarwood007:
I think it is fine for now.  There is other logic in here that will break in the future.  Such as when we start the new release branch and maintain 2 releases off this repo, this code pulls the latest "tag" to figure out things.  Which will fail with 2 active release branches.

Sesquipedalian:
It shouldn't be too hard to add some code to filter out tags that are not ancestors of the tip of the release branch.

EDIT: Actually, some minor tweaking of the existing `git describe` command may be all that is necessary.

jdarwood007:
I'm sure we can.  I just wanted to point out that while this may be flawed in attaching to a specific release branch, that we have other flaws in the code as well.

Sesquipedalian:
Well, attaching it to a specific release branch is not a flaw in the logic; it is intentional. The idea was that when we make a `release-2.2` branch (or whatever name it is given), we just change the value of `$release_branch` to that branch's name and commit that change to that branch. That way, when release-2.1 is checked out, this script will have `$release_branch = 'release-2.1';`, and when release-2.2 is checked out, it will have `$release_branch = 'release-2.2';`, etc.

sbulen:
> You'll just be merging those changes into the release branch immediately afterwards, so why bother with the extra steps?

It's the "immediately" that might be the issue.  Folks might point out other updates needed for one reason or another.  It may take time.  I like keeping a pristine copy of upstream as my default branch, in case another feature branch is needed.  

I am a firm believer in following the normal process!  It should be like any other PR.

That said, this is fine.  I'll continue to tweak it as needed.  It's helpful as-is.

sbulen (approved):
LGTM.  Good for current usage.

Sesquipedalian:
> Folks might point out other updates needed for one reason or another. It may take time.

When that happens, this script can be run again. It can be run and re-run as often as you like between releases. The key thing is just to make sure that you run it one last time immediately before you tag a release.

Sesquipedalian:
> I think it is fine for now. There is other logic in here that will break in the future. Such as when we start the new release branch and maintain 2 releases off this repo, this code pulls the latest "tag" to figure out things. Which will fail with 2 active release branches.

> It shouldn't be too hard to add some code to filter out tags that are not ancestors of the tip of the release branch.
> 
> EDIT: Actually, some minor tweaking of the existing `git describe` command may be all that is necessary.

I had a moment to test this, and in fact no changes to the current code are necessary to handle different tags for different branches. The command `git describe --tags --abbrev=0` always returns the most recent tag applied to the currently checked out branch. If different branches have different tags (e.g. `v2.1.3` in the `release-2.1` branch vs. `v2.2-alpha.1` in the `release-2.2` branch), the command will return the correct tag for whichever branch is checked out. So, no, nothing is broken in this logic.

PR #7598, opened 2022-11-04
Fix various issue with the upgrader running PHP 8.1

jdarwood007:
While testing for 2.1.3, I ran into the following issues on my git env.  This fixes them up. I did borrow code from a PHP RFC, but due to the simplicity of the code, I don't know if it can really be 'copyrighted'.  Included a url at least to the RFC for now.  This should go away in the future once we have moved to a PHP version that can handle unserialize properly.

( ! ) Fatal error: Uncaught TypeError: unserialize(): Argument #1 ($data) must be of type string, array given in /smf21git/upgrade.php on line 1091 ( ! ) TypeError: unserialize(): Argument #1 ($data) must be of type string, array given in /smf21git/upgrade.php on line 1091

( ! ) Fatal error: Uncaught TypeError: json_decode(): Argument #1 ($json) must be of type string, array given in /smf21git/upgrade.php on line 1104 ( ! ) TypeError: json_decode(): Argument #1 ($json) must be of type string, array given in /smf21git/upgrade.php on line 1104

( ! ) Deprecated: Implicit conversion from float 0.03333333333333333 to int loses precision in /smf21git/upgrade.php on line 5080

FYI, I am on PHP 8.1

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -1118,6 +1118,13 @@ function checkFolders()
 	{
 		// OK...
 	}
+	// An array already?
+	elseif (is_array($modSettings['attachmentUploadDir']))
+	{
+		foreach($modSettings['attachmentUploadDir'] AS $dir)
+			if (!empty($dir) && !is_dir($dir))
+				$attdr_problem_found = true;
+	}
 	// Serialized?
 	elseif ($ser_test !== false)
 	{
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -1088,8 +1088,28 @@ function checkFolders()
 
 	// Finally, attachment folders.
 	// A bit more complex, since it may be json or serialized, and it may be an array or just a string...
-	$ser_test = @unserialize($modSettings['attachmentUploadDir']);
-	$json_test = @json_decode($modSettings['attachmentUploadDir'], true);
+
+	// PHP currently has a terrible handling with unserialize in which errors are fatal and not catchable.  Lets borrow some code from the RFC that intends to fix this
+	// https://wiki.php.net/rfc/improve_unserialize_error_handling
+	try {
+    	set_error_handler(static function ($severity, $message, $file, $line) {
+			throw new \ErrorException($message, 0, $severity, $file, $line);
+		});
+		$ser_test = @unserialize($modSettings['attachmentUploadDir']);
+	} catch (\Throwable $e) {
+		$ser_test = false;
+	}
+	finally {
+	 	restore_error_handler();
+	}
+
+	// Json is simple, it can be caught.
+	try {
+		$json_test = @json_decode($modSettings['attachmentUploadDir'], true);
+	} catch (\Throwable $e) {
+		$json_test = null;
+	}
+
 	$string_test = !empty($modSettings['attachmentUploadDir']) && is_string($modSettings['attachmentUploadDir']) && is_dir($modSettings['attachmentUploadDir']);
 
 	// String?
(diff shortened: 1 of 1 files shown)


sbulen:
Hmmm...  I did all my testing in php 8.1 & never saw these errors.  I'll try to reproduce.

sbulen:
This all looks safe.  Let me run a bunch of tests....

jdarwood007:
I hadn't updated my git checkout in a while.  I was testing against git directly.

These should be safe as I simply wrap try/catches around these, as @ does not work for fatal errors since 8.0.  Those errors are fatal.  We could also I guess we do a string test before attempting any decoding.  The error is because $modSettings['attachmentUploadDir'] is already an array by the time the testing is done and thus fails.  But that would require some additional heavy testing as to not break the logic already there.

The later check is a !empty check on a is_dir.  If for some reason we get a empty string var into into the is_dir(), it is invalid.

sbulen:
Were you running CLI?  On a forum with lots of attachments?

I still don't see where those errors came from.   attachmentUploadDir is not converted to arrays before the attachment step anymore...  It used to do that, but that caused failures when restarting large upgrades (one of the changes made for large upgrades with failures & restarts after we converted smorg).  

Still trying to understand how you're getting there...

jdarwood007:
Negative on both.
I do get the `Warning! One or more attachment directories not found. Continuing may be unsafe. Please confirm folder settings before proceeding.`

I do have one or more invalid directories.
```
{"1":"\/smf21git\/attachments","2":"\/smf21git\/attachments\/2\/","3":"\/smf21git\/attachments\/2022\/02","4":"\/smf21git\/attachments\/2022\/03","5":"\/smf21git\/attachments\/2022\/06"}
```

Exists
- smf21git/attachments
- attachments/2022/02
- attachments/2022/03
- attachments/2022/06

So I am missing attachments/2

sbulen:
It bugs me that I cannot reproduce this.  And it bugs me that I do not know how you are seeing an array in attachmentsUploadDir at this point in the upgrade.  

But if we are truly seeing an array here, we need to check it.  We need another block of checks, I'd place it underneath the string test & above the serialized test.  Something like:
```
	// An array already?
	elseif (is_array($modSettings['attachmentUploadDir']))
	{
		foreach($modSettings['attachmentUploadDir'] AS $dir)
			if (!empty($dir) && !is_dir($dir))
				$attdr_problem_found = true;
	}

```

sbulen:
E.g., if it's a straight-up array, how does the attachment conversion work?
https://github.com/SimpleMachines/SMF/blob/aa2f168c69e5286222bae88a11c36edd3ce9368f/other/upgrade_2-1_mysql.sql#L436

jdarwood007:
Well thats skipping the decoding.

Looking at upgrade.php, reloadSettings() is called when we have $_GET['ssi'].  Maybe that is it? I don't recall setting that.

sbulen:
When you saw those errors - did they happen at the very beginning, before the database changes?   

Or did they come later?

jdarwood007:
It was doing it right at the start.  But now that I have tried to run the upgrader again, it doesn't do it.

sbulen:
OK, I've done a few tests & it seems safe.

BUT...  If we really think it may be in array format here, we need to check the contents of the array, per my note above.

Make that change & I'll do another round of tests & merge.

I'm still totally baffled how you got an array there...

sbulen (approved):
Looks safe.  Have run several tests & it works well.

I wish I knew how to reproduce the issue though.  Very odd.

jdarwood007:
Same here.  I didn't think to save what was in my settings table, which may have indicated a corrupted attachment settings.

PR #7600, opened 2022-11-04
Pacman tries to read a directory as a compressed file

jdarwood007:
Fixes error caused by package manager trying to read a directory as a compressed file

```
Notice
: file_get_contents(): Read of 8192 bytes failed with errno=21 Is a directory in
/Sources/Subs-Package.php
on line
36
```


live627:
How to reproduce this?

jdarwood007:
I would have to poke around to be sure, but I caught this in one of my error logs.
The fix is fairly straightforward and has our extraction function just bail out if it detects its a directory.

Issue #7601, opened 2022-11-05
Invalid argument supplied for foreach()


vsc55:
#### Description
2: Invalid argument supplied for foreach()
Type: general
Level: 2
File (/var/www/html/Sources/Subs.php) in line 4545
URL:  https://foro/index.php?action=dlattach;attach=2433

sbulen:
Steps to reproduce?

I don't think we've seen this.

Environment information - verison of SMF, PHP, etc.?

vsc55:
### Steps to reproduce
I don't know exactly how it happened, I saw it in the error logs.
It could be caused by creating a new subforum but I can't confirm it.

### Environment (complete as necessary)
- Version/Git revision: SMF 2.1.2
- Database Type: MariaDB
- Database Version: 10.3.32
- PHP Version: 7.4.1

PR #7602, opened 2022-11-05
#7601 Check is exist template_layers.

vsc55:


sbulen:
See comments in issue writeup.

We need to see steps to reproduce, so we can properly test before/after fix.

jdarwood007:
For a fix I would suggest:
```foreach ($context['template_layers'] ?? [] as $layer)```

sbulen:
The problem here is we don't know what the actual issue is.  Both the issue & PR writeup are blank.

I believe we were seeing this error early in the attachment changes, i.e., I believe this was a 2.1.1 or early 2.1.2 issue that has since been resolved.  

To confirm this, we need more info, e.g., steps to reproduce.  And to confirm it still exists with current code.

vsc55:
Hi @sbulen,
The version of smf is 2.1.2.
I don't know exactly how it happened, I saw it in the error logs.
It could be caused by creating a new subforum but I can't confirm it.

PR #7603, opened 2022-11-07
Protect from overlong USER_AGENTs

### LINKED ISSUE #7596: Database Error: Data too long for column 'url' at row 1
#### Description
Guest user.  Logging.php line 128

Not frequent - I've seen it in my production log only once, ever...

We should probably substring to make the insert safe - not sure we need THAT much URL...  

- Version/Git revision:  Current GH
- Database Type: MySQL
- Database Version: 5.6
- PHP Version: 7.4

Backtrace:
image

sbulen:
Fixes #7596 

Avoid storing/truncating strings that are too long in the url column in log_online.

Note that the truncate_array() function used already attempts to make sure the $_GET part is < 1900,  So in theory we have 2048 - 1900 = 148 bytes to work with.  But that's all before encoding, escaping, control characters, etc.

### DIFF
--- a/Sources/Logging.php
+++ b/Sources/Logging.php
@@ -52,7 +52,7 @@ function writeLog($force = false)
 
 	if (!empty($modSettings['who_enabled']))
 	{
-		$encoded_get = truncate_array($_GET) + array('USER_AGENT' => $_SERVER['HTTP_USER_AGENT']);
+		$encoded_get = truncate_array($_GET) + array('USER_AGENT' => mb_substr($_SERVER['HTTP_USER_AGENT'], 0, 128));
 
 		// In the case of a dlattach action, session_var may not be set.
 		if (!isset($context['session_var']))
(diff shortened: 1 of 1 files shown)


sbulen:
Before change, easy to break by altering USER_AGENT:
image

It's easy to putz with your USER_AGENT, most browsers let you just key in a new one:
image

Afterwards, with the same USER_AGENT, there are no errors...

Sources/Logging.php
```diff
@@ -52,14 +52,18 @@ function writeLog($force = false)
...
 		// In the case of a dlattach action, session_var may not be set.
 		if (!isset($context['session_var']))
 			$context['session_var'] = $_SESSION['session_var'];
 
 		unset($encoded_get['sesc'], $encoded_get[$context['session_var']]);
 		$encoded_get = $smcFunc['json_encode']($encoded_get);
+
+		// Sometimes folks mess with USER_AGENT & $_GET data, so one last check to avoid 'data too long' errors
+		if (mb_strlen($encoded_get) > 2048)
+			$encoded_get = '';
```
albertlast, line 66: instead of empty the encodie, you could write somekind of info,
"to long enconding" or something else,
so that this line get distinguishable to line 69.

sbulen: As near as I can tell this is not displayed or shown anywhere.  It only exists in the db for admin/debugging purposes.

Not worth inserting another string to translate.

PR #7604, opened 2022-11-07
Update col default & null handling

### LINKED ISSUE #7594: Defaults are still not what the condition expects?
From #7424

In this case, I was testing with db_change_column function and MySQL.
If you send `default => ''`

This line
https://github.com/SimpleMachines/SMF/blob/aa2f168c69e5286222bae88a11c36edd3ce9368f/Sources/DbPackages-mysql.php#L464

Will return the following: `"''",` which is not the same that is expected below:
https://github.com/SimpleMachines/SMF/blob/aa2f168c69e5286222bae88a11c36edd3ce9368f/Sources/DbPackages-mysql.php#L473

This causes the query to fail when changing a column with a primary key. If the column is **auto**, it doesn't have an empty value, which brings up the same issue as above because it will get that assigned and attempt to set a `DEFAULT ''`, which fails.

sbulen:
Fixes #7594 

Testing for both mysql & pg are complete.

Changes needed were:
 - Properly distinguish between NULL and not-specified defaults in db_list_columns
 - Fix lots of issues with nullable settings not being honored...
 - Honor the 'carry forward what's not overridden' rule consistently, for both pg & mysql
 - Add explicit support for dropping defaults (a new 'drop_default' option)

### DIFF
--- a/Sources/DbPackages-mysql.php
+++ b/Sources/DbPackages-mysql.php
@@ -436,7 +436,7 @@ function smf_db_change_column($table_name, $old_column, $column_info)
 	// Get the right bits.
 	if (!isset($column_info['name']))
 		$column_info['name'] = $old_column;
-	if (!isset($column_info['default']))
+	if (!array_key_exists('default', $column_info) && isset($old_info['default']) && empty($column_info['drop_default']))
 		$column_info['default'] = $old_info['default'];
 	if (!isset($column_info['not_null']))
 		$column_info['not_null'] = $old_info['not_null'];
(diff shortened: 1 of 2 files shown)


DiegoAndresCortes:
No changes for PG then?
I'll test it at some point this week

sbulen:
Not yet - pg changes coming.  That's why it's still in draft form.

sbulen:
The tests I have performed are as follows:
pg_mysql_chg_col_tests

sbulen:
OK, I am done with the pg side as well.  

In db_change_column, both pg & mysql had issues with changes with default values & changes in whether or not the column was nullable.  Also, mysql & pg behaved very differently with regards to what needed to be specified in db_change_column: pg wanted everything re-specified, where mysql only wanted the changes.  These issues have all been addressed & the behavior of the two have been brought in sync.

**_When given the same request, they must make the same updates, otherwise, mods don't have a chance of behaving properly._**

live627:
The commit I just pushed handles `null` defaults properly in `db_list_columns` on pgsql whereas before it was a string.

sbulen:
Hmmm...  I was properly getting null returned - & confirmed, otherwise, none of the other logic would have worked.  

But I like your logic.  It reads more clean.  And my tests still all work properly.  

And most important, under the covers, the old logic would pass null values to preg_match, and generate an internal error that was suppressed.  Your logic doesn't do that.  Cleaner flow.

live627:
I was using an adaptation of `test-7420.php` by @jdarwood007 

tbl

jdarwood007:
That refers to issue #7420 I named all the test files after the issue they where related to fixing.

sbulen:
Those tests appear to be nullable flag tests - and are checking whether the nullable flag is being set properly on a dummy varchar(1) column.  And on the nullable fields, it looks like the default was in fact NULL.  

We're looking at the _defaults_, not the _types_.  You cannot have a data type of null.  

And the good news is that in my tests, the _defaults_ were null as expected, where expected, both before & after your change.  

We're good!  😀

jdarwood007:
So the test file is flawed?  Got a revision we should put on it then?

sbulen:
I think the above is probably OK...  Depends on if the test was trying to make 3 nullable & set the default to null, and make the other 2 not nullable.  That's what we're seeing. 

Couldn't find the test source.

jdarwood007:
https://github.com/SimpleMachines/SMF/issues/7594#issuecomment-1303541915

I sent up a zip here with all the test files I had.

live627:
> So the test file is flawed?

No,  it's fine. The problem here was that this  PR  initially failed your test.

PR #7613, opened 2022-11-14
Update dependencies

live627:


live627:
- previously failing test now succeeds
- linting checks now run on pull requests
- no more warnings from running old versions of actions

This is a go

PR #7616, opened 2022-11-14
Show original topic title in ssi_recentTopics like in SMF 2.0

### LINKED ISSUE #7612: ssi_recentTopics returns post subject instead of topic title
#### Description
It seems that the output of `ssi_recentTopics` changed slightly in SMF 2.1: In SMF 2.0, `$post['subject']` contained the topic title (e.g. "Hello world!"), while in SMF 2.1, it contains the title of the last post in that topic (e.g. "Re: Hello world!").  Arguably, this behaviour makes more sense for `ssi_recentPosts` because it's about seeing what the latest posts are, while `ssi_recentTopics` is about seeing what the latest topics are. For my integration into another website, the "Re: " just adds unnecessary clutter, and I really want people to see the topic titles, not the reply subjects. Just removing the  "Re: " part might be a good temporary workaround, but if possible, it would be great if this function could return topic titles again.

### Steps to reproduce
1.  Call `$posts = ssi_recentTopics(10, null, null, 'not_echo');` from some SSI code
2.  See observation above

- Version/Git revision: 2.1.2
- Database Type: MySQL
- Database Version: 5.8
- PHP Version: 7.3

sagamusix:
Fixes #7612. I think for some of the selected columns it might not be 100% clear whether `mf` or `ml` should be used (e.g. `ml.icon`, I'm really not sure if it makes more sense semantically to show the icon of the first or last post and I'm not using that part of the feature), so that's up for debate.

### DIFF
--- a/SSI.php
+++ b/SSI.php
@@ -666,14 +666,14 @@ function ssi_recentTopics($num_recent = 8, $exclude_boards = null, $include_boar
 	// Find all the posts in distinct topics.  Newer ones will have higher IDs.
 	$request = $smcFunc['db_query']('substring', '
 		SELECT
-			mf.poster_time, mf.subject, ml.id_topic, mf.id_member, ml.id_msg, t.num_replies, t.num_views, mg.online_color, t.id_last_msg,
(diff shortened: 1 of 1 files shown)


sbulen (approved):
LGTM

sbulen:
Thanks for the submission!

PR #7638, opened 2022-12-02
Add permission for using Gravatars

### LINKED ISSUE #7635: Gravatars are enabled even when avatars are disabled
#### Description
Gravatars should be governed by avatar permissions.  There are no separate gravatar permissions.  On a forum with gravitars enabled, but a usergroup has no avatar permissions, that group can still use gravitars.  

image

image

### Steps to reproduce
1. Disable avatars for one membergroup
2. Enable gravitars globally
3. Logon as a member of that membergroup - gravitars are allowed

- Version/Git revision: 2.1.3
- Database Type: MySQL
- Database Version: 8.0
- PHP Version: 8.1

From the forum:
https://www.simplemachines.org/community/index.php?topic=584338.0

BrickOzp:
Add a separate group permission for using Gravatars.

Fixes #7635

### DIFF
--- a/Sources/ManagePermissions.php
+++ b/Sources/ManagePermissions.php
@@ -1157,6 +1157,7 @@ function setPermissionLevel($level, $group, $profile = 'null')
 		'profile_displayed_name',
 		'profile_upload_avatar',
 		'profile_remote_avatar',
+		'profile_gravatar',
 		'profile_remove_own',
 		'report_user',
 	));
@@ -1506,6 +1507,7 @@ function loadAllPermissions()
 			'profile_server_avatar' => array(false, 'profile'),
 			'profile_upload_avatar' => array(false, 'profile'),
 			'profile_remote_avatar' => array(false, 'profile'),
+			'profile_gravatar' => array(false, 'profile'),
 			'report_user' => array(false, 'profile'),
 			'profile_identity' => array(true, 'profile_account'),
 			'profile_displayed_name' => array(true, 'profile_account'),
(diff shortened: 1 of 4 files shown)


BrickOzp:
A drawback with this solution is that all forums that uses Gravatar need to set this permission for all groups that should be able to use Gravatars after the upgrade.

Oldiesmann:
Since the current default behavior in SMF is to allow Gravatars even when other avatar options aren't allowed, the solution to that could just be to set that permission by default (or at least try to - may not always be possible depending on how complicated the setup is), and tell people they can remove the permission if they don't want it.

jdarwood007:
I wouldn't approve of this for release until we handle the permission issue.  As we are adding a new permission mid release, I feel we need to address what the expected "default" would be.  I would go off of the permission to choose a remotely stored avatar and clone that.  But only if we have detected no existing permissions exists for gravatars.

@BrickOzp Do you think you could make something like that happen or should we delay this for a future release?
If you can make it for this release, send up the PR and move the milestone.

BrickOzp:
Is the upgrader run for minor releases?

live627:
2.1.x -> 2.1.x + 1? No; patches are run from the package manager.

jdarwood007:
Well we can execute a script during install.

The issue will be functionality will appear to break or not work due to this upgrade that previously worked.
A quick scetch of the code that I think would work
```
If (count of select * from permissions where permission = 'gravatar') == 0)
   insert into permissions select permission = 'gravatar, * from permissions where permission = 'remotely stored avatar'
```
Shouldn't hurt to run this as its permissions on a general level and not board level.

Sesquipedalian (approved):

Sesquipedalian:
I will add code to the patch to clone the existing profile_remote_avatar permission to profile_gravatar.

PR #7659, opened 2022-12-23
Fix no such file errors

### LINKED ISSUE #7657: fopen(/xxxx/cache/data_blahblahblah-SMF-theme_settings-1.cache): Failed to open stream: No such file or directory
#### Description
Since 2.1.3, I periodically get this failed to open stream.  It happens a few times a week.

Problem is there is in fact time between the 'file_exists' check and the 'fopen'.:
https://github.com/SimpleMachines/SMF/blob/d3a8569377639fd559068477a05c5cf2bc0579fb/Sources/Cache/APIs/FileBased.php#L62

In a high-volume retrieval environment, that cache may have a new key within that timeframe.  

I have a solution.  Will submit a PR.

### Steps to reproduce
1. Pretty sure you need high volume retrieval, i.e., lots of activity.  & SMF file caching on.

- Version/Git revision: 2.1.3
- Database Type: MySQL
- Database Version: 8.0
- PHP Version: 8.1

Screenshot-2
Screenshot-1

sbulen:
Fixes #7657 

Testing that change in a petri dish...  Before:
old_code

After:
new_code

### DIFF
--- a/Sources/Cache/APIs/FileBased.php
+++ b/Sources/Cache/APIs/FileBased.php
@@ -59,7 +59,7 @@ class FileBased extends CacheApi implements CacheApiInterface
 
 	private function readFile($file)
 	{
-		if (file_exists($file) && ($fp = fopen($file, 'rb')) !== false)
+		if (($fp = @fopen($file, 'rb')) !== false)
 		{
 			if (!flock($fp, LOCK_SH))
 			{


Sesquipedalian (approved):

jdarwood007:
Looks like I authored this change that caused this issue.  Wonder if it would make sense to wrap these in a try/catch/finally and have it more gracefully just bail out of errors.

sbulen:
> Looks like I authored this change that caused this issue. Wonder if it would make sense to wrap these in a try/catch/finally and have it more gracefully just bail out of errors.

What this change does is basically that - if it cannot find the cache, it ends up creating a new one.  I think we're good here.

jdarwood007:
Yea, I just don't trust @ operator to do its job anymore since they changed it in PHP 8.

PR #7666, opened 2023-01-06
Properly handle disabled hooks & includes

### LINKED ISSUE #7634: Hooks - squirrely behavior upon disable/uninstall/reinstall
#### Description
While looking at this forum topic:
https://www.simplemachines.org/community/index.php?topic=584301.0

I realized there are certain circumstances that the hook maintenance under Admin | Maintenance | Forum Maintenance | Integration Hooks doesn't play nice when installing/uninstalling mods.

Strange things can happen, were you end up with duplicate hook entries, some enabled & some disabled:
image

image

There is logic in the hooks processing to avoid creating dupes, however, I don't think it factors in the enabled/disabled status.  It appears to only work 100% of the time on enabled hooks.

**_A reinstall should re-enable any pre-existing, disabled, hooks, rather than adding a new entry._**

### Steps to reproduce
1. Disable all the hooks (either by repair_settings.php or in the admin console)
2. Uninstall the package
3. Reinstall the package

- Version/Git revision: 2.1.3
- Database Type: MySQL
- Database Version: 8.0
- PHP Version: 8.1

The workaround that works all the time is to:
1. deinstall
2. manually delete all the hooks in the admin console
3. reinstall

sbulen:
Fixes #7634

This addresses a number of problems where adding/removing mods did not properly deal with disabled hooks.  It also addresses numerous issues with the Integration Hooks admin function when dealing with disabled hooks & also when dealing with includes. 

Changes:
 - When adding a hook, if a disabled entry previously exists, it should get rid of the old disabled entry, & avoid creating a dupe.
 - When removing a hook, it should look for disabled entries as well, otherwise they will be left in place, e.g., even after a mod de-install.
 - Some file operations (file exists, etc.) were including the disabled indicator (!) in the file/function name, which would make it look like the file/function did not exist when it did.  This prevented some operations in the Integration Hooks maintenance function from working.
 - Use the path when it's available.  Some add/remove hook operations did not check for the path.  In particular, enable/disable operations use a remove/add approach, so removing an entry resulted in no change (due to no hit, since the path wasn't specified), but the add resulted in adding a partial entry (no path).  The result was an erroneous partial entry that could not be fixed.
 - Eat dupes, to help when cleaning things up.  There should never be duplicates.
 - Allow for enabling/disabling of includes.  Needed for when they are disabled elsewhere, e.g., in repair_settings.php.

At this point, adding mods no longer creates dupes, and results in one clean set of integration hooks, even when cleaning up after a broken install/uninstall.  Removing mods removes all the hooks, no matter what the status.  

Also, as intended, you can now use the Integration Hooks function to do most hook cleanup activity.  Even if you were to accidentally disable all hooks in repair_settings.php.

### DIFF
--- a/Sources/ManageMaintenance.php
+++ b/Sources/ManageMaintenance.php
@@ -1884,8 +1884,9 @@ function list_integration_hooks()
 
 		else
 		{
-			$function_remove = urldecode($_REQUEST['function']) . (($_REQUEST['do'] == 'disable') ? '' : '!');
-			$function_add = urldecode($_REQUEST['function']) . (($_REQUEST['do'] == 'disable') ? '!' : '');
+			// Disable/enable logic; always remove exactly what was passed
+			$function_remove = urldecode($_REQUEST['function']);
+			$function_add = urldecode(rtrim($_REQUEST['function'], '!')) . (($_REQUEST['do'] == 'disable') ? '!' : '');
 
 			remove_integration_function($_REQUEST['hook'], $function_remove);
 			add_integration_function($_REQUEST['hook'], $function_add);
--- a/Sources/ManageMaintenance.php
+++ b/Sources/ManageMaintenance.php
@@ -1979,11 +1979,8 @@ function list_integration_hooks()
 					{
 						$change_status = array('before' => '', 'after' => '');
 
-						if ($data['can_disable'])
-						{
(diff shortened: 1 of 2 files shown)


sbulen:
## Some before/after tests...

### Test #1 - disable one hook:
**_Before - mutant partial hook entry (no path specified):_**
before_disable_one

**_After - hook properly disabled:_**
after_disable_one

### Test #2 - uninstall with a disabled hook:
**_Before - hook fragment left behind:_**
before_uninstall_w_disabled

**_After - all hooks removed:_**
after_uninstall_w_disabled

### Test #3 - disable all via repair_settings.php:
**_Before - note it thinks the include is missing:_**
before_disable_all

**_After - it no longer thinks the include is missing, so it can be re-enabled:_**
after_disable_all

sbulen:
### Some more "before" tests to demonstrate the problems:
**_Before - deinstall when disabled hooks exist leave them in place (this PR properly deletes them):_**
before_disabled_not_uninstalled

**_Before - install when disabled hooks exist result in dupes (this PR removes the dupes):_**
before_disabled_dupes on install

jdarwood007:
I tested this with a few of my mods, including dev tools.  Nothing seemed to break and the hooks did seem to remain clean.

PR #7667, opened 2023-01-06
Updates SMF_SOFTWARE_YEAR to 2023

Sesquipedalian:


sbulen:
In all prior years, we did all 200+ files.

I am actually not sure of the legal requirement.  Do we have to do all, or only the DEFINEs?

Sesquipedalian:
> In all prior years, we did all 200+ files.
> 
> 
> 
> I am actually not sure of the legal requirement.  Do we have to do all, or only the DEFINEs?

Well, the file headers and the constants are two separate matters.

During the pre-release stage of development we updated the copyright year in all file headers every year. But post-release, the longstanding practice with SMF has been that file headers are only updated when the file content changes. (FYI, I believe this practice is driven merely by the desire to keep patches as slim as possible, rather than any legal requirement one way or the other. We *could* update the copyright year in all file headers if we wanted; we just don't.)

The SMF_SOFTWARE_YEAR constant, on the other hand, always needs to be up to date (which also means updating the file headers in the files where the constant is defined). We also make an exception to the practice described in the previous paragraph for ./other/Settings.php and ./other/Settings_bak.php, in that we always keep the copyright year in the file headers of those two files up to date with the current year. Finally, because of the way check-smf-license.php was written, we also need to update the copyright year in all ./other/*.php files merely in order to satisfy the check. 

As an aside, it would probably make sense to tweak the code in check-smf-license.php so that the last sentence of the previous paragraph was no longer true. But that would require a PR to the SMF/BuildTools repository, and is therefore obviously out of scope for this PR.

jdarwood007:
The check license script should most likely be updated to accept a range of dates, starting from the first date of 2.1 to the current year and accept all years in that range.  We could update it to check the current tag of the previous release and then compare all modified files and ensure they have the proper year.  Its supposed to help remind us for the little things we don't remember.

Sesquipedalian:
I don't think we need to specify a date range. I think we just need to replace this line with two lines that explicitly name Settings.php and Settings_bak.php:

https://github.com/SimpleMachines/BuildTools/blob/f5de83f89ec17f55927b31b910c87603e4aed6f1/check-smf-license.php#L70

live627 (approved):

PR #7673, opened 2023-01-13
Remove erroneous verification check

### LINKED ISSUE #7597: Verification questions not configured properly. Please report this error to an administrator.
#### Description
I am getting these errors in my log occasionally.  These errors are usually, but not always, for guests.  Sometimes it looks like it's during a search, sometimes during a preview...

But the verification questions are in fact configured properly...

There was a recent PR that uniquely ID'd the error recently.  I don't think this PR caused the error, but I do think it made it easier to identify exactly which edit is failing:  #7536 

Prior to that PR, the error was the "you don't have access" one.  However, it was not logged.  Since #7536, it's getting logged.

It looks like when you just change the language in the middle of a captcha test, it's missing the $_REQUEST part of this test:
https://github.com/SimpleMachines/SMF/blob/aa2f168c69e5286222bae88a11c36edd3ce9368f/Sources/Subs-Editor.php#L2115

### Steps to reproduce
1. Have verification questions setup for guest search.
2. Do a guest search, but answer the verification questions incorrectly, so you are provided error messages.
3. Change languages in the dropdown in the top menu bar.

- Version/Git revision:  Current GH
- Database Type: mysql
- Database Version: 5.6
- PHP Version: 7.4

Backtrace of all variants of this error include the Post.php line 1651 & the Subs-Editor.php line 2116:
image

sbulen:
Fixes #7597 

This error appeared whenever a user changed languages while a incorrect answer to a verification question error was displayed.  

Note that this error said the verification questions were incorrectly configured, however, the verification questions were in fact configured properly.  So, completely erroneous errors were logged and presented to the user, and the user was left in a weird state having lost their input.

This could happen anywhere there were verification questions: e.g., guest search, registration, new user first posts.  

The old edit was looking at $_REQUEST, which might get clobbered upon a submit - and changing languages issues another form submit on the same page.

With this PR, the erroneous error is no longer logged or presented.  The users may still lose their input (due to changing the language in the middle of a form submit), but at least they now get a meaningful error message.

### DIFF
--- a/Sources/Subs-Editor.php
+++ b/Sources/Subs-Editor.php
@@ -2111,9 +2111,6 @@ function create_control_verification(&$verificationOptions, $do_test = false)
 		// This cannot happen!
 		if (!isset($_SESSION[$verificationOptions['id'] . '_vv']['count']))
 			fatal_lang_error('no_access', false);
-		// Verification question does not exist for this language.
-		if ($thisVerification['number_questions'] && (!isset($_SESSION[$verificationOptions['id'] . '_vv']['q']) || !isset($_REQUEST[$verificationOptions['id'] . '_vv']['q'])))
-			fatal_lang_error('registration_no_verification_questions');
 		// Hmm, it's requested but not actually declared. This shouldn't happen.
 		if ($thisVerification['empty_field'] && empty($_SESSION[$verificationOptions['id'] . '_vv']['empty_field']))
 			fatal_lang_error('no_access', false);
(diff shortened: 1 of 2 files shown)


BrickOzp:
What will happen if the questions aren't configured for the language?
What is presented to the user?

sbulen:
Try it!

sbulen:
The worst symptom I saw was questions in English.

BrickOzp:
Finally had time to test this.
If verification questions are not set up for the language the registration will just go through without showing any questions.
Don't know if that is a good solution or not.

This was after upgrade from 2.0 to 2.1. The verification questions in 2.0 was not language specific, but in 2.1 they are. The upgrade copied them to english. Meaning that swedish did not have any questions set up.

But we might need to change this error message?
"You need to add a question and answer for your forum's default language (Svenska) otherwise users will not be able to fill in a CAPTCHA, meaning no registration."

jdarwood007:
Confirmed fixed the issue.

PR #7688, opened 2023-02-24
Uses built-in 'matcher' callback of At.js

### LINKED ISSUE #7687: Inserting mentions in WYSIWYG mode can mess with cursor position
#### Description
Inserting mentions in WYSIWYG mode messes with the cursor position when trying to edit the same paragraph that the mention appears in. Specifically, under certain conditions (see below) the cursor will keep jumping to the end of the paragraph when trying to edit a paragraph that contains a mention.

### Steps to reproduce
1. Put the following text into the quick reply (replace "@test1" with a username on your test forum):
```
Lorem ipsum dolor sit amet, consectetuer adipiscing elit. @test1  Aenean tempus justo sed arcu. Sed ut neque. Nam quis odio in nulla dapibus laoreet. Suspendisse interdum volutpat ligula. 

Nullam sagittis nunc quis massa. @test1 Curabitur a pede vel risus mollis dignissim. Aenean nec massa. Vivamus at tellus. Sed ac nunc. Pellentesque feugiat ultricies quam. Ut porta turpis commodo pede dignissim laoreet.
```
2. Click the preview button to go to the full post authoring form.
3. Click somewhere after the mention in the second paragraph.
4. Observe how the cursor jumps to the end of the paragraph.

- Version/Git revision: 5bc137d
- Database Type: N/A
- Database Version: N/A
- PHP Version: N/A

Original report: https://www.simplemachines.org/community/index.php?topic=585125.0

Sesquipedalian:
Fixes #7687

For those who are curious, the issue was that the custom 'matcher' callback we were using would capture everything from the `@` to the end of the line, rather than just the actual matching string. Reverting to the default behaviour fixes this, and I can find no negative side effects in my tests.

### DIFF
--- a/Themes/default/scripts/mentions.js
+++ b/Themes/default/scripts/mentions.js
@@ -4,6 +4,7 @@ var atwhoConfig = {
 	at: '@',
 	data: [],
 	show_the_at: true,
+	startWithSpace: true,
 	limit: 10,
 	callbacks: {
 		remoteFilter: function (query, callback) {
--- a/Themes/default/scripts/mentions.js
+++ b/Themes/default/scripts/mentions.js
@@ -6,25 +6,6 @@ var atwhoConfig = {
 	show_the_at: true,
 	limit: 10,
 	callbacks: {
-		matcher: function(flag, subtext, should_start_with_space) {
-			var match = '', started = false;
-			var string = subtext.split('');
-			for (var i = 0; i < string.length; i++)
-			{
-				if (string[i] == flag && (!should_start_with_space || i == 0 || /[\s\n]/gi.test(string[i - 1])))
-				{
-					started = true;
-					match = '';
-				}
-				else if (started)
-					match = match + string[i];
-			}
-
-			if (match.length > 0)
-				return match;
-
-			return null;
-		},
 		remoteFilter: function (query, callback) {
 			if (typeof query == 'undefined' || query.length < 2 || query.length > 60)
 				return;


jdarwood007:
#1794 was the PR that added that code.  Any issues you foresee from what that PR was trying to fix?

Sesquipedalian:
No. There was no description in that PR and I could not find a related issue, so I had to try to figure out what the purpose of the code changes was myself. As far as I can tell, the other changes in that PR did improve some stuff, whereas the custom 'matches' callback doesn't appear to have accomplished anything except introducing this bug. Perhaps the custom code originally provided a workaround for a bug in At.js that was later solved upstream, but either way it doesn't do anything useful now. 🤷🏻‍♂️

For what its worth, I initially tried seeing if I could fix the custom code, and was making progress until I realized that my fixes were essentially just reimplementing what was in the default code anyway.

Themes/default/scripts/mentions.js
```diff
@@ -6,25 +6,6 @@ var atwhoConfig = {
 	show_the_at: true,
```
live627, line 6:
```suggestion
	show_the_at: true,
	startWithSpace: true,
```

live627 (changes requested):
I have found that the custom matcher required whitespace before the keyword to mat ch. This restores that behavior.

Sesquipedalian:
Nice. Suggestion committed.

live627 (approved):

jdarwood007:
Thanks for the answer, I wasn't sure what the original code was intending to do either, so I was seeing if somebody else understood.

live627:
I only figured it out because of the change in behavior after I applied the patch.

PR #7691, opened 2023-02-28
set sent based on the notification type

### LINKED ISSUE #7680: Not receiving some notifications
#### Description
One of my users was complaining he's not getting notifications for one of the subscribed threads - it was verified, that forum does not send any email (ie. not going to spam). I was getting the notifications for the very same thread (so notifications work). After long testing it _seems_ that if user has both BOARD and THREAD notifications on, no notification gets through.

### Steps to reproduce
1. Set board notifications to emails & notifications (this board contains thread from point 2)
2. Set thread notifications to emails & notifications
3. Wait for any message to the thread
4. No notification gets through. 
5. If board notifications are turned off and new message is posted, the notification appears.

- Version/Git revision: 2.1.3

I wasn't able to find any other explanation for the behavior described and I also wasn't able to find any other settings which explain it.

live627:
fix #7680

### DIFF
--- a/Sources/tasks/CreatePost-Notify.php
+++ b/Sources/tasks/CreatePost-Notify.php
@@ -280,7 +280,7 @@ class CreatePost_Notify_Background extends SMF_BackgroundTask
 			$smcFunc['db_query']('', '
 				UPDATE {db_prefix}log_notify
 				SET sent = {int:is_sent}
-				WHERE (id_topic = {int:topic} OR id_board = {int:board})
+				WHERE ' . ($type == 'topic' ? 'id_board = {int:board}' : 'id_topic = {int:topic}') . '
 					AND id_member IN ({array_int:members})',
 				array(
 					'topic' => $topicOptions['id'],


Sources/tasks/CreatePost-Notify.php
```diff
@@ -280,7 +280,7 @@ public function execute()
 			$smcFunc['db_query']('', '
 				UPDATE {db_prefix}log_notify
 				SET sent = {int:is_sent}
-				WHERE (id_topic = {int:topic} OR id_board = {int:board})
+				WHERE ' . ($type == 'topic' ? 'id_board = {int:board}' : 'id_topic = {int:topic}') . '
```
Sesquipedalian, line 283: Just checking, but did you actually mean this instead?
```suggestion
				WHERE ' . ($type !== 'topic' ? 'id_board = {int:board}' : 'id_topic = {int:topic}') . '
```
It seems odd to mark the board notification as sent when the notification type was topic.

live627: Odd perhaps, but is a new topic notif; therefore `ln.id_topic` won't be set.

https://github.com/SimpleMachines/SMF/blob/6060f363db575314b4c59a9a20485a1081439fbb/Sources/Subs-Post.php#L2170

sbulen: @Sesquipedalian - this is correct.  The internal naming of the types can be a bit confusing...  But a board watch triggers a 'topic' notification type (i.e., upon a new topic) and a topic watch triggers a 'reply' notification type.  So when you see a 'topic' notification, you gotta look up the board to set the sent flag.

This very confusion was at the heart of some of the 2.1.3 notification PRs...

sbulen: Cross post...

sbulen:
@live627 going forward please check if there was someone assigned to the issue...  (I've been testing a similar PR the last 2 days!)

live627:
oh whoops, sorry about that.

Sources/tasks/CreatePost-Notify.php
Sesquipedalian: Fair enough. Thanks for the explanation. 🙂

sbulen:
No problem.  

What I've been testing is whether quotes/mentions/edit notifications still work, and it appears they do.   Lots of types to test...

sbulen (approved):
Tests out OK...

live627:
@sbulen we have a bot running that automatically removes these labels.

PR #7692, opened 2023-02-28
fix adding to board_permissions_view

### LINKED ISSUE #7689: Permissions issue creating admin-only boards
#### Description
Removing all permissions groups should create an admin-only board.  But the board remains visible on the board index.  Note you cannot access the board, but it is visible.

Note also that if you assign a group permission to that board, and subsequently remove it, it behaves properly & is no longer visible on the board index...  

### Steps to reproduce
1. Remove all permissions for a board, including guests.
2. Logout & access the board index as a guest.  

- Version/Git revision: Current GH
- Database Type: MySQL
- Database Version: 8.0
- PHP Version: 8.1

From the forum:
https://www.simplemachines.org/community/index.php?topic=585181.0

live627:
fix #7689

### DIFF
--- a/Sources/Subs-Boards.php
+++ b/Sources/Subs-Boards.php
@@ -680,7 +680,7 @@ function modifyBoard($board_id, &$boardOptions)
 		)
 	);
 
-	if ($board_permissions_inserts != [])
+	if ($board_permissions_inserts != array())
 		$smcFunc['db_insert']('insert',
 			'{db_prefix}board_permissions_view',
 			array('id_group' => 'int', 'id_board' => 'int', 'deny' => 'int'),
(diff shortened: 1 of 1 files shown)


Sources/Subs-Boards.php
```diff
@@ -664,45 +671,22 @@ function modifyBoard($board_id, &$boardOptions)
...
+	// Before we add new access_groups or deny_groups, remove all of the old entries
+	$smcFunc['db_query']('', '
+		DELETE FROM {db_prefix}board_permissions_view
+		WHERE id_board = {int:selected_board}',
+		array(
+			'selected_board' => $board_id,
+		)
+	);
 
+	if ($board_permissions_inserts != [])
```
Sesquipedalian: Nit: SMF coding style prefers the `array()` syntax.
```suggestion
	if ($board_permissions_inserts != array())
```

jdarwood007:
FYI to test this you simply need to setup a board with guest only group.  Modify and remove guest group.  Then view the board index as a guest.  After this fix, it works as expected and does not show.

PR #7698, opened 2023-03-06
Runs scheduled tasks via cron.php

### LINKED ISSUE #7696: Cron check doesn't work properly
#### Description
There is a check, based on `cron_is_real_cron` to verify that cron.php is being triggered by some external trigger, and this check uses the scheduled tasks table to verify if there are any late running tasks (more than 1 day old).

But cron.php doesn't trigger any of the scheduled tasks handling, nor clearing the mail queue that I can see.

- Version/Git revision: 2.1.3
- Database Type: n/a
- Database Version: n/a
- PHP Version: n/a

cron.php probably should explicitly trigger the scheduled tasks routine and the mail queue routine if there is nothing else for it to do in its alloted time period. It's currently only checking the background tasks queue, which is as I intended it to do originally.

Sesquipedalian:
Fixes #7696

This is an alternative to #7697. The approach in this PR is superior, but it is a more substantial change than in the other PR. Therefore I am offering both for consideration.

### DIFF
--- a/Sources/Load.php
+++ b/Sources/Load.php
@@ -2614,34 +2614,6 @@ function loadTheme($id_theme = 0, $initialize = true)
 	loadJavaScriptFile('script.js', array('defer' => false, 'minimize' => true), 'smf_script');
 	loadJavaScriptFile('theme.js', array('minimize' => true), 'smf_theme');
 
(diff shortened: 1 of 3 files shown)


sbulen:
I definitely prefer this one.

jdarwood007:
Works as expected.

PR #7704, opened 2023-03-15
Automatically checks out new branch when updating version numbers

Sesquipedalian:
As requested by @sbulen in https://github.com/SimpleMachines/SMF/pull/7580#issuecomment-1465302306 🙂

As before, it is still up to you to commit and merge the changes (whether directly or via a PR). But this should be more convenient for your preferred workflow.


sbulen:
Perfect!

The prior version violated the "Don't poop where you eat!" rule!

sbulen (approved):

PR #7717, opened 2023-03-24
Improve temporary hook handling

sbulen:
Followup on & enhance #7666

The UI was still confusing, in particular when dealing with temporary hooks.  Tightened up on allowed actions based on status.  E.g., it _looked_ like it was updating temporary hooks, but the updates were ignored.

Specific changes:
 - Added ability to discern which hooks are temporary.
 - Added two new statuses to the legend, for temporary hooks & for temporary hooks that are missing.
 - No actions are allowed on temporary hooks.  The only way to remove or disable a temporary hook is to deinstall the mod or theme.
 - The only appropriate action for permanent hooks that are not found is removal.

**_With Changes - all is good, one temp hook:_**
Hooks-1

**_With Changes - some hooks disabled:_**
Hooks-2

**_With Changes - source code is missing for some hooks:_**
Hooks-3

### DIFF
--- a/Sources/ManageMaintenance.php
+++ b/Sources/ManageMaintenance.php
@@ -1977,12 +1977,24 @@ function list_integration_hooks()
 				'data' => array(
 					'function' => function($data) use ($txt, $scripturl, $context, $filter_url)
 					{
-						$change_status = array('before' => '', 'after' => '');
+						// Cannot update temp hooks in any way, really.  Just show the appropriate icon.
+						if ($data['status'] == 'temp')
+						{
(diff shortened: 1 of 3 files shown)


dragomano (approved):

jdarwood007:
The code logic itself looks good.  I don't have any mods around to test the temp hook issue on.

I have a rewriting logic I was going to submit for the future code that handles the entire hook system way better, but its a major feature overhaul, including a new database table, not suitable for a patch release.

sbulen:
Thanks for looking at this @dragomano & @jdarwood007 !

PR #7718, opened 2023-03-24
Fixes Issue #7655: Correct notification mail template for "Immediate"

### LINKED ISSUE #7655: Notification mail template is inconsistent with "Immediately" notification setting
#### Description

Continued from https://www.simplemachines.org/community/index.php?topic=584475.0

Even if you set the notification settings to "Immediately" rather than "Immediately - First time only", the mail template says the following:

> More replies may be posted, but you won't receive any more notifications for this topic until you visit it.

I certainly got a bit confused because I got more than one mail, but I also didn't get a mail for every reply. This sentence should probably be omitted from the mail if notifications are set to "Immediate". Maybe it should also mention that not every reply will trigger an immediate mail notification, as SMF tries to avoid being too spammy.

nc3man:
The value of $frequency must be checked in order to select the correct message based on user notification frequency setting

### DIFF
--- a/Sources/tasks/CreatePost-Notify.php
+++ b/Sources/tasks/CreatePost-Notify.php
@@ -529,7 +529,7 @@ class CreatePost_Notify_Background extends SMF_BackgroundTask
 
 				$content_type = 'board';
 
-				$message_type = !empty($frequency) ? 'notify_boards_once' : 'notify_boards';
+				$message_type = (!empty($frequency) && $frequency==2) ? 'notify_boards_once' : 'notify_boards';
 
 				if (empty($modSettings['disallow_sendBody']) && !empty($this->prefs[$member_id]['msg_receive_body']))
 					$message_type .= '_body';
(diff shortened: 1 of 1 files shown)


nc3man:
Sorry for the first bolluxed PR attempt. Thank you for closing that, although I see I could have done it as well. I was able to learn a little more and was able to rebase my repo by removing all 3 previous commits.
```
git reset --hard 7c7593518
git push --force origin HEAD:release-2.1
```
where 7c7593518 was the SHA in reflog prior to the first commit. Learned something :). I then committed the simple one line change. 
Thanks for your patience.

Sources/tasks/CreatePost-Notify.php
```diff
@@ -529,7 +529,7 @@ protected function handleWatchedNotifications()
 
 				$content_type = 'board';
 
-				$message_type = !empty($frequency) ? 'notify_boards_once' : 'notify_boards';
+				$message_type = (!empty($frequency) && $frequency==2) ? 'notify_boards_once' : 'notify_boards';
```
live627:
```suggestion
				$message_type = !empty($frequency) && $frequency == 2 ? 'notify_boards_once' : 'notify_boards';
```

live627 (changes requested):
just some formatting

PR #7721, opened 2023-03-25
Improve handling of SVG image attachments and WYSIWYG attachment previews

### LINKED ISSUE #7719: SVG attachments handled inconsistently
#### Description
SVG attachments are not treated like images in topic display, but they are treated like images by the editor. As a result, if an SVG is uploaded and then inserted into the post using the [attach] BBCode, it will appear as an image in the editor preview, but will then appear as a link when the post is displayed after submission.

### Steps to reproduce
1. Make sure you have allowed SVG file attachments in the attachment settings.
2. On the post authoring page, upload an SVG file as attachment. Notice that the SVG image is displayed as an image in the attachments area.
3. Click the Post button to submit the post.
4. View the post. Notice that the attachment is displayed below the post like a non-image file (i.e. no preview image).
5. Click the post's More... > Modify button to go back to the post authoring page.
6. Click the insert button to insert the SVG into the post. Notice the SVG appears as an image in the editor.
7. Submit and view the post. Notice that the SVG appears as a link, not an image.

- Version/Git revision: 2.1.4
- Database Type: N/A
- Database Version: N/A
- PHP Version: N/A

Back during the beta development phase, I meant to add full support for SVG images throughout the codebase. Apparently I missed this part. 😖

Anyway, I have a fix forthcoming.

### LINKED ISSUE #7720: Inconsistent presentation of embedded image attachments in editor vs. topic display
#### Description
When the [attach] BBCode to embed an image attachment into the text of a post, the image is displayed at full size in the topic display (which is good). However, when the editor is WYSIWYG mode, the thumbnail image is used. The thumbnail usually has much smaller dimensions than the full image, which can confuse the end user who (1) rightly expects the full size image to be shown in the post and (2) not unreasonably expects that when using a WYSIWYG editor, what they see in the editor will in fact be what they get in the final output.

### Steps to reproduce
1. Embed an image attachment into a post.
2. Notice how it uses the image's thumbnail in the WYSIWYG editor, but the full image in the topic display.

- Version/Git revision: 2.1.4
- Database Type: N/A
- Database Version: N/A
- PHP Version: N/A

There's no benefit to using the thumbnail in the editor rather than the original image. The purpose of thumbnails is to save on bandwidth, but we'll be sending out the original image anyway, so why bother?

One could argue that the users should know that they need to click the Preview button to get an accurate preview, but I know from experience as an admin that users get confused and annoyed by the inconsistency between the WYSIWYG presentation and the final presentation.

Sesquipedalian:
Fixes #7719 by showing SVG attachments as images.

The fiddliest part was handling thumbnails. SVG images don't need thumbnails for obvious reasons, but SMF assumes that images do need them. So I had to add some special handling for SVGs to tell SMF that the SVG image is its own thumbnail, and then slightly tweak the JavaScript used to expand thumbnails when the user clicks on them.

---

Fixes #7720 by showing original images for embedded attachments in WYSIWYG editor. 

The basic task of switching from thumbnails to original images merely required changing `thumb` to `image` on the relevant line in jquery.sceditor.smf.js. However, to prevent the download counter from incrementing unexpectedly, I also had to resurrect the `preview` URL parameter in ShowAttachment.php as a way to tell SMF that this shouldn't count as a download.

### DIFF
--- a/Sources/Subs-Attachments.php
+++ b/Sources/Subs-Attachments.php
@@ -687,6 +687,12 @@ function createAttachment(&$attachmentOptions)
 	$size = @getimagesize($attachmentOptions['tmp_name']);
 	list ($attachmentOptions['width'], $attachmentOptions['height']) = $size;
 
+	if (!empty($attachmentOptions['mime_type']) && $attachmentOptions['mime_type'] === 'image/svg+xml')
+	{
+		foreach (getSvgSize($attachmentOptions['tmp_name']) as $key => $value)
+			$attachmentOptions[$key] = $value === INF ? 0 : $value;
+	}
+
 	if (function_exists('exif_read_data') && ($exif_data = @exif_read_data($attachmentOptions['tmp_name'])) !== false && !empty($exif_data['Orientation']))
 		if (in_array($exif_data['Orientation'], [5, 6, 7, 8]))
 		{
@@ -1305,7 +1311,7 @@ function loadAttachmentContext($id_msg, $attachments)
 			if (!empty($attachment['id_thumb']))
 				$attachmentData[$i]['thumbnail'] = array(
 					'id' => $attachment['id_thumb'],
-					'href' => $scripturl . '?action=dlattach;attach=' . $attachment['id_thumb'] . ';image',
+					'href' => $scripturl . '?action=dlattach;attach=' . $attachment['id_thumb'] . ';image;thumb',
 				);
 			$attachmentData[$i]['thumbnail']['has_thumb'] = !empty($attachment['id_thumb']);
 
@@ -1334,6 +1340,17 @@ function loadAttachmentContext($id_msg, $attachments)
 
 			if (!$attachmentData[$i]['thumbnail']['has_thumb'])
 				$attachmentData[$i]['downloads']++;
+
+			// Describe undefined dimensions as "unknown".
+			// This can happen if an uploaded SVG is missing some key data.
+			foreach (array('real_width', 'real_height') as $key)
+			{
+				if (!isset($attachmentData[$i][$key]) || $attachmentData[$i][$key] === INF)
+				{
+					loadLanguage('Admin');
+					$attachmentData[$i][$key] = ' (' . $txt['unknown'] . ') ';
+				}
+			}
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


Sources/Subs-Attachments.php
```diff
@@ -1420,4 +1457,125 @@ function prepareAttachsByMsg($msgIDs)
...
+			// Percentage.
+			// SVG spec says to use viewBox dimensions in this case.
+			default:
+				unset($$dimension);
+				break;
+		}
+	}
+
+	// Width and/or height is missing or a percentage, so try the viewBox attribute.
+	if ((!isset($width) || !isset($height)) && preg_match('/\bviewBox\s*=\s*(["\'])\s*[\d.]+[,\s]+[\d.]+[,\s]+([\d.]+)[,\s]+([\d.]+)\s*\\1/', $svg, $matches, PREG_UNMATCHED_AS_NULL))
```
live627:
```suggestion
	if ((!isset($width) || !isset($height)) && preg_match('/\bviewBox\s*=\s*(["\'])\s*[\d.]+[,\s]+[\d.]+[,\s]+([\d.]+)[,\s]+([\d.]+)\s*\\1/', $svg, $matches))
```
This is, I believe, PHP 7.2+

Sources/Subs-Attachments.php
```diff
@@ -1420,4 +1457,125 @@ function prepareAttachsByMsg($msgIDs)
...
 
+/**
+ * Gets the dimensions of an SVG image (specifically, of its viewport).
+ *
+ * See https://www.w3.org/TR/SVG11/coords.html#IntrinsicSizing
+ *
+ * @param string $filepath The path to the SVG file.
+ * @return array The width and height of the SVG image in pixels.
+ */
+function getSvgSize($filepath)
```
live627: Should this function not be moved to Subs-Graphics?

Sesquipedalian: Good idea.

Sources/Subs-Attachments.php
Sesquipedalian: Derp. Yes, it is.

Sesquipedalian: ...and I don't actually need it anyway.

Sesquipedalian:
Ready for another review, @live627.

PR #7722, opened 2023-03-26
Shows "Last Edit" info (if applicable) when editing a post

Sesquipedalian:
The logic for setting `$context['last_modified_text']` was broken. This fixes it, and also simplifies how we handle and display that information in the posting fields on the editor page.

### DIFF
--- a/Sources/Post.php
+++ b/Sources/Post.php
@@ -748,6 +748,7 @@ function Post($post_errors = array())
 		// When was it last modified?
 		if (!empty($row['modified_time']))
 		{
+			$modified_reason = $row['modified_reason'];
 			$context['last_modified'] = timeformat($row['modified_time']);
 			$context['last_modified_reason'] = censorText($row['modified_reason']);
 			$context['last_modified_name'] = $row['modified_name'];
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
@jdarwood007: I'm not 100% sure, but I get the impression that @sbulen actually plans to put this into 2.1.4 rather than 2.1.5, since it is connected with #7676. I could be entirely wrong, of course.

jdarwood007:
Sounded like it was an existing bug, not one introduced by that PR?

Sesquipedalian:
The bug itself already existed, yes. But the changes needed in order to fix it also affect some of the new code introduced in #7676. Because of that, it wouldn't make a whole lot of sense to leave this out, at least to my mind. But it's up to you folks to decide, of course.

jdarwood007:
Ok.  I will leave it up for @sbulen then to decide.  I interrupted it as a existing bug so we can leave it out while we work to finalize 2.1.4

DiegoAndresCortes:
Why is it using  `$context['last_modified_reason']` in the value, wouldn't it be censored?

Sesquipedalian:
> Why is it using `$context['last_modified_reason']` in the value, wouldn't it be censored?

Excellent catch! It would indeed. I have fixed that now.

DiegoAndresCortes (approved):

Issue #6920, opened 2021-08-09
Slow query in the action to find forum errors


live627:
https://github.com/SimpleMachines/SMF2.1/blob/0f375f1b9b6682e9cf8415624296281b1559441e/Sources/RepairBoards.php#L603-L618

```sql
ANALYZE format=json SELECT
	t.id_topic, t.id_first_msg, t.id_last_msg,
	CASE WHEN MIN(ma.id_msg) > 0 THEN
		CASE WHEN MIN(mu.id_msg) > 0 THEN
			CASE WHEN MIN(mu.id_msg) < MIN(ma.id_msg) THEN MIN(mu.id_msg) ELSE MIN(ma.id_msg) END ELSE
		MIN(ma.id_msg) END ELSE
	MIN(mu.id_msg) END AS myid_first_msg,
	CASE WHEN MAX(ma.id_msg) > 0 THEN MAX(ma.id_msg) ELSE MIN(mu.id_msg) END AS myid_last_msg,
	t.approved, mf.approved, mf.approved AS firstmsg_approved
FROM smf_topics AS t
	LEFT JOIN smf_messages AS ma ON (ma.id_topic = t.id_topic AND ma.approved = 1)
	LEFT JOIN smf_messages AS mu ON (mu.id_topic = t.id_topic AND mu.approved = 0)
	LEFT JOIN smf_messages AS mf ON (mf.id_msg = t.id_first_msg)
WHERE t.id_topic BETWEEN 2 AND 3
GROUP BY t.id_topic, t.id_first_msg, t.id_last_msg, t.approved, mf.approved
ORDER BY t.id_topic
```
I had to tightly limit the range because otherwise it would timeout.
```json
{
  "query_block": {
    "select_id": 1,
    "r_loops": 1,
    "r_total_time_ms": 8154.2,
    "filesort": {
      "sort_key": "t.id_topic, t.id_first_msg, t.id_last_msg, t.approved, mf.approved",
      "r_loops": 1,
      "r_total_time_ms": 0.0109,
      "r_used_priority_queue": false,
      "r_output_rows": 2,
      "r_buffer_size": "360",
      "temporary_table": {
        "table": {
          "table_name": "t",
          "access_type": "range",
          "possible_keys": ["PRIMARY"],
          "key": "PRIMARY",
          "key_length": "3",
          "used_key_parts": ["id_topic"],
          "r_loops": 1,
          "rows": 2,
          "r_rows": 2,
[...]

PR #7731, opened 2023-04-20
Update RepairBoards.php

atagmester:
Improve performance

### DIFF
--- a/Sources/RepairBoards.php
+++ b/Sources/RepairBoards.php
@@ -629,8 +629,8 @@ function loadForumTests()
 					CASE WHEN MAX(ma.id_msg) > 0 THEN MAX(ma.id_msg) ELSE MIN(mu.id_msg) END AS myid_last_msg,
 					t.approved, mf.approved, mf.approved AS firstmsg_approved
 				FROM {db_prefix}topics AS t
-					LEFT JOIN {db_prefix}messages AS ma ON (ma.id_topic = t.id_topic AND ma.approved = 1)
-					LEFT JOIN {db_prefix}messages AS mu ON (mu.id_topic = t.id_topic AND mu.approved = 0)
+					LEFT JOIN {db_prefix}messages AS ma ON (ma.id_topic = t.id_topic AND ma.approved = 0)
+					LEFT JOIN {db_prefix}messages AS mu ON (mu.id_topic = t.id_topic AND mu.approved = 1)
 					LEFT JOIN {db_prefix}messages AS mf ON (mf.id_msg = t.id_first_msg)
 				WHERE t.id_topic BETWEEN {STEP_LOW} AND {STEP_HIGH}
 				GROUP BY t.id_topic, t.id_first_msg, t.id_last_msg, t.approved, mf.approved


Sources/RepairBoards.php
```diff
@@ -1854,4 +1854,4 @@ function createSalvageArea()
 	loadLanguage('Admin');
 }
 
-?>
```
BrickOzp, line 1857: Remove this change

sbulen:
I am closing this.  It doesn't make sense to rename unapproved messages to approved & vice-versa.  No actual impact on performance.

Best guess is this was a test of some sort.

atagmester:
It makes sense, the query takes 5 minutes as is, and 2 seconds with the modification. Both produce the same output…

Sesquipedalian:
It seems like what you are trying to do is, in effect, to switch the order in which the CASE conditions on lines 624 and 625 are evaluated.

Changing that order quite possibly would improve performance in most cases, since unapproved messages are less common than approved ones.

However, the approach that you took in this pull request is problematic, because it also flips the meaning of other parts of the conditional logic in the query. It will almost certainly do the wrong thing. 

If your changes happened to produce the same results in your situation, you either got very lucky, or you else you first ran the existing code to repair your tables and then ran your modified code against the same, already repaired data.

live627:
I identified this problem two years ago and documented in https://github.com/SimpleMachines/SMF/issues/6920

Rereading it, I see that I never shared my rewritten query. I'll try to find it in my rats nest of local branches.

jdarwood007:
Looks like topics is not hitting a index and using a temp table/file sort.  A large forum could suffer here.  I tried to play with adding a index, but even if I get it to use a index, it still does a temp table/file sort.

live627:
A filesort is never the problem on slow queries that I analyze. Temporary tables, on the other hand...

live627:
The filesort in that thread took 0.01ms.

jdarwood007:
Screen Shot 2023-04-20 at 9 35 25 PM

I tried a index on topics for id_topic, id_first_msg, id_last_msg and approved.  It got the query to find a index.  But still doing a temp table.

live627:
GitKraken erroneously states that my repo is private and is holding me hostage until I cough up the dough.  Anyway, here is my code:

```diff

diff --git a/Sources/RepairBoards.php b/Sources/RepairBoards.php
index f9198acd08..c75e256132 100644
--- a/Sources/RepairBoards.php
+++ b/Sources/RepairBoards.php
@@ -601,27 +601,30 @@ function loadForumTests()
 			),
 			'check_query' => '
 				SELECT
-					t.id_topic, t.id_first_msg, t.id_last_msg,
-					CASE WHEN MIN(ma.id_msg) > 0 THEN
-						CASE WHEN MIN(mu.id_msg) > 0 THEN
-							CASE WHEN MIN(mu.id_msg) < MIN(ma.id_msg) THEN MIN(mu.id_msg) ELSE MIN(ma.id_msg) END ELSE
-						MIN(ma.id_msg) END ELSE
-					MIN(mu.id_msg) END AS myid_first_msg,
-					CASE WHEN MAX(ma.id_msg) > 0 THEN MAX(ma.id_msg) ELSE MIN(mu.id_msg) END AS myid_last_msg,
-					t.approved, mf.approved, mf.approved AS firstmsg_approved
-				FROM {db_prefix}topics AS t
-					LEFT JOIN {db_prefix}messages AS ma ON (ma.id_topic = t.id_topic AND ma.approved = 1)
-					LEFT JOIN {db_prefix}messages AS mu ON (mu.id_topic = t.id_topic AND mu.approved = 0)
-					LEFT JOIN {db_prefix}messages AS mf ON (mf.id_msg = t.id_first_msg)
+					t.id_topic, t.id_first_msg, t.id_last_msg, t.approved,
+					myid_first_msg, myid_last_msg, firstmsg_approved
+				FROM smf_topics AS t
+					JOIN
+					(
+						SELECT
+							id_topic, MIN(id_msg) AS myid_first_msg, approved AS firstmsg_approved
+						FROM smf_messages
+						WHERE id_topic BETWEEN {STEP_LOW} AND {STEP_HIGH}
+						GROUP BY id_topic, approved
+					) AS mf USING (id_topic)
+					JOIN
+					(
+						SELECT
+							id_topic, MAX(id_msg) AS myid_last_msg
+						FROM smf_messages
+						WHERE id_topic BETWEEN {STEP_LOW} AND {STEP_HIGH}
+						AND approved = 1
+						GROUP BY id_topic
+					) AS ml USING (id_topic)
 				WHERE t.id_topic BETWEEN {STEP_LOW} AND {STEP_HIGH}
-				GROUP BY t.id_topic, t.id_first_msg, t.id_last_msg, t.approved, mf.approved
 				ORDER BY t.id_topic',
 			'fix_processing' => function($row) use ($smcFunc)
 			{
-				$row['firstmsg_approved'] = (int) $row['firstmsg_approved'];
-				$row['myid_first_msg'] = (int) $row['myid_first_msg'];
-				$row['myid_last_msg'] = (int) $row['myid_last_msg'];
-
 				// Not really a problem?
 				if ($row['id_first_msg'] == $row['myid_first_msg'] && $row['id_last_msg'] == $row['myid_last_msg'] && $row['approved'] == $row['firstmsg_approved'])
 					return false;
```

sbulen:
It makes more sense to continue this discussion over on #6920 .

And answer the questions raised there.  

This PR is closed.

PR #7733, opened 2023-04-22
Remove menu item if all submenu items are disabled

### LINKED ISSUE #7559: 8: Undefined index: label, xxx/Themes/default/GenericMenu.template.php (Line 121)
#### Description
On occasion, I get an "Undefined index: label" error in the log.  In this instance, the user is a global moderator, but doesn't have many of the moderation permissions.  

- Version/Git revision: Current GH
- Database Type: MySQL
- Database Version: 8.0
- PHP Version: 8.1

At issue is this logic:
https://github.com/SimpleMachines/SMF/blob/e8c7800eadb6a41e5b87ec2d11a1e34f959bc392/Sources/Subs-Menu.php#L233

Under certain (rare) circumstances, it can get to that logic without $first_sa and $last_sa being set.  E.g., if the user doesn't have permission to do any of the activities in the submenu.  First & last end up getting set on an undefined/null new entry.  The template can't find the label, because it is null:
image

sbulen:
Fixes #7559 

This occurs if all submenu items are disabled.  You get a weird submenu list, and an error in the error log.

_**BEFORE...**  You can see that all submenu items were disabled via permissions._
label_issue_1
label_issue_3

**_AFTER..._**
label_issue_4

### DIFF
--- a/Sources/Subs-Menu.php
+++ b/Sources/Subs-Menu.php
@@ -229,6 +229,13 @@ function createMenu($menuData, $menuOptions = array())
 									$menu_context['sections'][$section_id]['areas'][$area_id]['subsections'][$sa]['disabled'] = true;
 							}
 
+							// If permissions removed/disabled for all submenu items, remove the menu item
+							if (empty($first_sa) && empty($last_sa))
+							{
+								unset($menu_context['sections'][$section_id]['areas'][$area_id]);
+								continue;
+							}
+
 							// Set which one is first, last and selected in the group.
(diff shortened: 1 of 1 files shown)


Sesquipedalian (approved):

live627 (approved):

live627:
which app you use to view the array like that?

sbulen:
That's a screenshot from the Eclipse php step debugger.  Using xdebug.

PR #7735, opened 2023-04-25
Attachments fixes

DiegoAndresCortes:
- The insert popup has a higher height than what a small image could have:
These are examples uploading images with low height:
Screenshot_88
Screenshot_86

- Fixes a typo in the file.

### DIFF
--- a/Themes/default/css/attachments.css
+++ b/Themes/default/css/attachments.css
@@ -129,7 +129,7 @@ div#post_attachments_area.roundframe {
 	display: flex;
 	justify-content: center;
 	margin: 0.5em 0;
-	min-height: 175px;
+	min-height: 75px;
 	max-height: 200px;
 	width: auto;
 }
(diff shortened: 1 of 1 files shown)


DiegoAndresCortes:
@Sesquipedalian

Sesquipedalian:
~I can probably submit a fix in the next day or two.~

Sorry. I misread this as a issue that I was being asked to solve, rather than a PR that I was being asked to review.

sbulen:
As-is, this causes the attachment area to get quite large:
image

DiegoAndresCortes:
Oops as per the discussion topic, it was 75px, this has a typo

Sesquipedalian:
What would happen if you tried raising the z-index of the menu instead of changing the min-height of the image?

DiegoAndresCortes:
Probably would need to remove/play with the overflows since it’s hidden in one or two containers. I don't want to mess around with it too much as of course it may have a purpose for the current behavior (if it ain’t broke…).
You're welcome to tweak it or provide the best solution if it’s safe to do so, no problem with that.

Sesquipedalian:
#7737 gives an alternative approach.

PR #7736, opened 2023-04-26
Send real_name in quote notification emails

### LINKED ISSUE #7734: Notification emails should use display name in all circumstances
#### Description
Reported in the forum here:
https://www.simplemachines.org/community/index.php?topic=585650.0

Confirmed with quotes - display name was not used as expected

- Version/Git revision:  2.1.3
- Database Type: mysql
- Database Version: 8.0
- PHP Version: 8.2

BrickOzp:
The username was used in the quote notification emails instead of real_name.

Fixes #7734

### DIFF
--- a/Sources/tasks/CreatePost-Notify.php
+++ b/Sources/tasks/CreatePost-Notify.php
@@ -628,6 +628,8 @@ class CreatePost_Notify_Background extends SMF_BackgroundTask
 		$msgOptions = &$this->_details['msgOptions'];
 		$posterOptions = &$this->_details['posterOptions'];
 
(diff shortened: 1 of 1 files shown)


live627:
looks like the relevant info is in MEMBERNAME and that we may need to fix the relevant language string instead.

BrickOzp:
No MEMBERNAME and QUOTENAME are different users
From the email template

Hello {MEMBERNAME},

You have been quoted in the post titled "{CONTENTSUBJECT}" by {QUOTENAME}, you can see the post here

live627:
you right, sorry

PR #7737, opened 2023-04-26
Tweaks CSS for insert menu for small images

Sesquipedalian:
Closes #7735

Rather than changing the min-height of the preview wrapper, this adjusts the z-index of the menu div and the overflow properties of the relevant container divs. It also adjusts the border-radius on the corners of the file size progress bar, because I was formerly relying on the overflow properties to take care of hiding those corners where they would otherwise spill outside the rounded corners of the container.

### DIFF
--- a/Themes/default/css/attachments.css
+++ b/Themes/default/css/attachments.css
@@ -36,7 +36,7 @@
 
 /* Stuff unique to Post.template.php */
 div#post_attachments_area.roundframe {
-	overflow: auto;
+	overflow: visible;
 	border-top: none;
 	padding: 0;
 }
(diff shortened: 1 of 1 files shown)


sbulen:
@Sesquipedalian - I'm testing & comparing this to the other solution.

I don't think this one fixes the typo on line ~33 that is causing console warnings (and invalidates the css).

Sesquipedalian:
I'll add a fix for that later today.

Sesquipedalian:
@sbulen: Done.

sbulen:
After comparing the two, I prefer the alternate approach.  Looks less cluttered.

Sesquipedalian:
As you wish.

DiegoAndresCortes (approved):

jdarwood007:
Is this still making 2.1.5?  I forgot if another decision was made.

sbulen:
Yes.  We've seen other examples of overflow reported on the forum.  This acts as a safety net...

Sesquipedalian:
> Yes.  We've seen other examples of overflow reported on the forum.  This acts as a safety net...

In light of @DiegoAndresCortes approval and this statement, I will go ahead and merge now.

PR #7772, opened 2023-06-15
Banlist fixes

### LINKED ISSUE #7278: Formatting on ban list needs some tlc
#### Description
image

### Steps to reproduce
1. Enter very long strings with no breaks in ban reason & notes

Not a massive issue, but should be addressed at some point...

DiegoAndresCortes:
Fixes #7278 

- Word break on ban list.
- Long reasons would throw db error.

### DIFF
--- a/Sources/ManageBans.php
+++ b/Sources/ManageBans.php
@@ -158,7 +158,7 @@ function BanList()
 				),
 				'data' => array(
 					'db' => 'notes',
-					'class' => 'smalltext',
+					'class' => 'smalltext word_break',
 				),
 				'sort' => array(
 					'default' => 'LENGTH(bg.notes) > 0 DESC, bg.notes',
@@ -171,7 +171,7 @@ function BanList()
 				),
 				'data' => array(
 					'db' => 'reason',
-					'class' => 'smalltext',
+					'class' => 'smalltext word_break',
 				),
 				'sort' => array(
 					'default' => 'LENGTH(bg.reason) > 0 DESC, bg.reason',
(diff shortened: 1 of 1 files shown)


Sources/ManageBans.php
```diff
@@ -869,6 +869,9 @@ function banEdit2()
 
 		call_integration_hook('integrate_edit_bans', array(&$ban_info, empty($_REQUEST['bg'])));
 
+		// Limit 'reason' characters
+		$ban_info['reason'] = substr($ban_info['reason'], 0, 255);
```
Sesquipedalian:
```suggestion
		$ban_info['reason'] = $smcFunc['truncate']($ban_info['reason'], 255);
```

The string may have entities in it. A simple `substr()` could easily end up cutting an entity in half, resulting in an invalid string. Calling `$smcFunc['truncate']()` will handle this correctly.

Sesquipedalian (changes requested):

Sources/ManageBans.php
DiegoAndresCortes: I did forget to use our functions.
What's the difference between truncate and substr in $smcFunc, the entities?

Sesquipedalian: Both functions handle entities as single characters when counting. The difference is in how they interpret their $length arguments 

$smcFunc['substr'] produces a string that visually appears to be the specified length to a human reader. The real string length will be greater if the string contains any entities.

In contrast, $smcFunc['truncate'] produces a string whose real length will be equal to or less than the specified value. If the string contains any entities, it will visually appear to a human reader to be shorter than the specified length.

Since you are trying to ensure that the string fits into a database field, the real string length is the one that matters. Therefore, $smcFunc['truncate'] is the function you want.

DiegoAndresCortes: Alright thanks

Sesquipedalian (approved):

PR #7774, opened 2023-06-15
[2.1] Set quick edit to have same behavior as modify

### LINKED ISSUE #7642: Cannot change "reason for editing" text in quick edit mode
#### Description
Apparently the small explaining text that can be added when the grace period has expired cannot be changed in quick edit mode (i.e. modifications don't get saved)

### Steps to reproduce
1. wait for the grace period to expire
2. try to add or change the explanation text in quick edit mode

- Version/Git revision: SMF 2.1.3
- Database Type: MySQL
- Database Version: 5.7.35-38-log
- PHP Version: 7.1.11

Haven't checked as regular user yet, only admin. To actually save changes to the database I have to switch to standard edit mode. Happens on both my posts and other users post.

DiegoAndresCortes:
Fixes #7642 

It now matches full modify:
- No reason set when still in grace period
- Set a modified time if the reason changed

### DIFF
--- a/Sources/Post.php
+++ b/Sources/Post.php
@@ -3208,7 +3208,12 @@ function JavaScriptModify()
 		}
 		// If nothing was changed there's no need to add an entry to the moderation log.
 		else
+		{
 			$moderationAction = false;
+			
+			// Clear the reason
+			$msgOptions['modify_reason'] = '';
+		}
 
 		modifyPost($msgOptions, $topicOptions, $posterOptions);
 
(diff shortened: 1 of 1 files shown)


DiegoAndresCortes:
I will change the behavior of this.
Apparently the full modify allows you to edit a message by just changing the modify reason, doesn't make sense in my opinion but I will edit it to match that behavior.

Sesquipedalian:
It makes sense if you made a typo in the edit reason and want to fix that without making any further changes in the message itself.

DiegoAndresCortes:
Had to do more changes to match the behavior of modify.

Sesquipedalian (approved):

Sesquipedalian:
@DiegoAndresCortes, could you please also submit a PR to make the same changes in 3.0?

PR #7775, opened 2023-06-15
[2.1] Fix issue with wrong date on event creating

### LINKED ISSUE #7630: Invalid date when posting calendar event with default french format
#### Description
We enabled the calendar feature on our forum. It worked fine, until an "Invalid date" error appeared when posting an event past the 12th of the month… which is almost certainly a parsing error of french date format (dd/mm/YYYY instead of american mm/dd/YYYY).

Changing the forum default date format to "%d %B %Y, %H:%M:%S" fixed the problem, but now the calendar date picker displays dates as "24 Novembre 20…" (long french format that does not fit in the box).
Also manually typing the date into the picker using french short format dd/mm/YYYY causes the "Invalid date" error.
Users of our forum are all french and I expect them to input "invalid dates" frequently.

Expected behaviour :
- picking the date using the date picker component should allow to send the form with a date format that can be parsed by the server
- manually entering the date should be disallowed or mark the field as invalid if the typed value can’t be parsed
- french users (and locales with similar date formats) should be able to input dates in a short, familiar format

### Steps to reproduce
1. Enable calendar on an smf forum with french translation
2. Post an event with a date past the 12th of the month (25th of december 2022) OR type directly the date in usual french format (25/12/2022) => the date picker is not marked as invalid, the preview works, but submitting the event leads to an "Invalid date" error page

- Version/Git revision: 2.1.2
- Database Type: mysql
- Database Version:
- PHP Version: 7.3

Don’t every brother support input type="date" nowadays?

dragomano:
This issue is actual only for some languages (French, Russian, etc)

Fixes #7630 
Closes #7630

### DIFF
--- a/Themes/default/Post.template.php
+++ b/Themes/default/Post.template.php
@@ -155,12 +155,12 @@ function template_main()
 								<div class="event_options_left" id="event_time_input">
 									<div>
 										<span class="label">', $txt['start'], '</span>
-										<input type="text" name="start_date" id="start_date" value="', trim($context['event']['start_date_orig']), '" tabindex="', $context['tabindex']++, '" class="date_input start" data-type="date">
+										<input type="text" name="start_date" id="start_date" value="', trim($context['event']['start_date']), '" tabindex="', $context['tabindex']++, '" class="date_input start" data-type="date">
 										<input type="text" name="start_time" id="start_time" maxlength="11" value="', $context['event']['start_time_orig'], '" tabindex="', $context['tabindex']++, '" class="time_input start" data-type="time"', !empty($context['event']['allday']) ? ' disabled' : '', '>
 									</div>
 									<div>
 										<span class="label">', $txt['end'], '</span>
-										<input type="text" name="end_date" id="end_date" value="', trim($context['event']['end_date_orig']), '" tabindex="', $context['tabindex']++, '" class="date_input end" data-type="date"', $modSettings['cal_maxspan'] == 1 ? ' disabled' : '', '>
+										<input type="text" name="end_date" id="end_date" value="', trim($context['event']['end_date']), '" tabindex="', $context['tabindex']++, '" class="date_input end" data-type="date"', $modSettings['cal_maxspan'] == 1 ? ' disabled' : '', '>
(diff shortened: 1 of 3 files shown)


Themes/default/Post.template.php
```diff
@@ -362,7 +362,7 @@ function addPollOption()
 							<div class="files" id="attachment_previews">
 								<div id="au-template">
 									<div class="attachment_preview_wrapper">
-										<div class="attach-ui">
+										<div class="attach-ui roundframe">
```
albertlast: why is an ui change needed for this issue?

dragomano: It's not necessary. Maybe it's because I haven't updated the fork in a while.

Sesquipedalian: I suspect that part of the recent attachment UI changes were accidentally merged into this PR.

isaak654:
```suggestion
										<div class="attach-ui">
```

isaak654 (changes requested):

Sesquipedalian:
This won't work correctly. `$context['event']['start_date_orig']` gives the start date according to the event's own time zone, whereas `$context['event']['start_date']` gives the start date according to the viewing user's time zone. So if the event's internal data says that it starts at `2024-05-21 23:00:00 Europe/Dublin` and user's time zone is `Europe/Moscow`, then `$context['event']['start_date_orig']` will say `2024-05-21`, but `$context['event']['start_date']` will say `2024-05-22`, because `2024-05-21 23:00:00 Europe/Dublin` is `2024-05-22 01:00:00 Europe/Moscow`.

dragomano:
The problem is that `start_date_org` contains the date not in the format that you mention in the example. In the case of English, there will be no problems, but with Cyrillic dates, an error `Invalid date` will occur when creating an event. If you are not sure about `start_date`, I can use `date('Y-m-d', $context['event']['start_timestamp'])` in a hidden field,. Likewise for `end_date_orig` and `end_timestamp`.

1) Open "Post event" page:
sshot-28
2) Select any date via date widget - as you see, the format changes
sshot-29

Results for English
sshot-27
Results for Russian
sshot-26

PR #7778, opened 2023-06-20
Prevent null value from $user_info in logActions

### LINKED ISSUE #7546: logActions can get invalid id_member when called from Scheduled tasks
When triggering _removeMembersFromGroups_ using a scheduled task, it calls logAction > logActions to log the changes.
When using a scheduled task, id_member might be 0, causing the following:

```
The database value you're trying to insert does not exist: id_member
Function: logActions
```

Culprit: 
https://github.com/SimpleMachines/SMF/blob/release-2.1/Sources/Logging.php#L539-L542

DiegoAndresCortes:
Fixes #7546

### DIFF
--- a/Sources/Logging.php
+++ b/Sources/Logging.php
@@ -543,7 +543,7 @@ function logActions(array $logs)
 		if (isset($log['extra']['member_affected']))
 			$memID = $log['extra']['member_affected'];
 		else
-			$memID = $user_info['id'];
+			$memID = (isset($user_info['id']) && !empty($user_info['id']) ? $user_info['id'] : (!empty($log['extra']['member']) ? $log['extra']['member'] : 0));
 
 		if (isset($user_info['ip']))
 			$memIP = $user_info['ip'];
--- a/Sources/Logging.php
+++ b/Sources/Logging.php
@@ -543,7 +543,7 @@ function logActions(array $logs)
 		if (isset($log['extra']['member_affected']))
 			$memID = $log['extra']['member_affected'];
 		else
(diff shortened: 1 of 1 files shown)


Sources/Logging.php
```diff
@@ -543,7 +543,7 @@ function logActions(array $logs)
 		if (isset($log['extra']['member_affected']))
 			$memID = $log['extra']['member_affected'];
 		else
-			$memID = $user_info['id'];
+			$memID = (isset($user_info['id']) && !empty($user_info['id']) ? $user_info['id'] : (!empty($log['extra']['member']) ? $log['extra']['member'] : 0));
```
jdarwood007: !empty implies isset, you don't need to specify both.  As another note.  If the value being checked is set to null, it is considered empty and not set, you have to use is_null checks in that case, except this doesn't work for array keys, you have to use array_key_exists.  We technically could use null coalescing here as well.  That is a PHP 7.0 feature.

DiegoAndresCortes: I think is because I initially had the condition reversed.
$log['extra']['member'] might not need to be checked since it's already checked.

`if (isset($log['extra']['member']) && !is_numeric($log['extra']['member']))`

Perhaps like this?
`$memID = $user_info['id'] ?? $log['extra']['member']`

jdarwood007: 2.1 does say 7.0 or higher, so I would say yes, just like that for null coalescing. Could even go further:

```$memID = $user_info['id'] ?? $log['extra']['member'] ?? 0;```

DiegoAndresCortes: I updated it.

Sesquipedalian (approved):

PR #7788, opened 2023-07-08
SMTP Mail Queueing improvements

### LINKED ISSUE #7787: SMTP Email - live server edits vs relay
#### Description
We have seen a couple of instances where SMTP servers are performing edits when SMF connects to queue up an email for sending.  I.e., it is validating the _**recipient**_ in real time, while adding the entry to the mail queue on the SMTP server.

Some SMTP servers apparently allow this behavior, and do not act as a relay.  

The far more common behavior is to queue it up, and later, upon sending, if it fails, it fails.  

The problem with this behavior, though, is that the request fails, the entry remains in the mail queue.  So, it gets processed again, the next time the SMF mail queue gets processed.  

So...  If the recipient does not exist, the email can stay in the SMF queue forever...

From the forum: https://www.simplemachines.org/community/index.php?topic=586346.0

jdarwood007:
Fixes #7787

Trying to keep inside the database schema, i'm abusing the priority level to slowly raise it to the max (127 due to signed tinyint).  It does this over a period of 3 days (which can be changed) by calculating the difference between the current time and the time the email was sent.

### DIFF
--- a/Sources/ScheduledTasks.php
+++ b/Sources/ScheduledTasks.php
@@ -779,7 +779,7 @@ function ReduceMailQueue($number = false, $override_limit = false, $force_send =
 	foreach ($emails as $email)
 	{
 		// This seems odd, but check the priority if we should try again so soon. Do this so we don't DOS some poor mail server.
-		if ($email['priority'] > $priority_offset && (time() - $email['time_sent']) % $priority_offset != 0)
+		if ($email['priority'] > $priority_offset && (time() - $email['time_sent']) % $priority_offset != rand(0, $priority_offset))
 		{
 			$failed_emails[] = array($email['to'], $email['body'], $email['subject'], $email['headers'], $email['send_html'], $email['time_sent'], $email['private'], $email['priority']);
 			continue;
@@ -805,8 +805,12 @@ function ReduceMailQueue($number = false, $override_limit = false, $force_send =
 		else
 			$result = smtp_mail(array($email['to']), $email['subject'], $email['body'], $email['headers']);
 
+		// Old emails should expire
+		if (!$result && $email['priority'] >= $max_priority)
+			$result = true;
+
 		// Hopefully it sent?
-		if (!$result && $email['priority'] < $max_priority)
+		if (!$result)
 		{
 			// Determine the "priority" as a way to keep track of SMTP failures.
 			$email['priority'] = max($priority_offset, $email['priority'], min(ceil((time() - $email['time_sent']) / $smtp_expire * ($max_priority - $priority_offset)) + $priority_offset, $max_priority));
--- a/Sources/ScheduledTasks.php
+++ b/Sources/ScheduledTasks.php
@@ -775,8 +775,16 @@ function ReduceMailQueue($number = false, $override_limit = false, $force_send =
 	$failed_emails = array();
 	$max_priority = 127;
 	$smtp_expire = ($modSettings['smtp_day_expiry'] ?? 3) * 60 * 60 * 24;
+	$priority_offset = 4;
 	foreach ($emails as $email)
 	{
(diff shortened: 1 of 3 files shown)


jdarwood007:
I'm still working on this.  I had a issue where the mail was deleted and it shouldn't have been deleted yet.  Still looking into it.

Sources/ScheduledTasks.php
```diff
@@ -772,8 +773,18 @@ function ReduceMailQueue($number = false, $override_limit = false, $force_send =
 
 	// Send each email, yea!
 	$failed_emails = array();
+	$max_priority = 127;
+	$smtp_expire = ($modSettings['smtp_day_expiry'] ?? 3) * 60 * 60 * 24;
```
Sesquipedalian: If the value of the setting changes while there are unsent emails in the queue, it will mess up the calculations. I suggest ditching the setting and just hard coding the value. It's not the sort of thing that a forum admin would normally ever need to adjust anyway.

jdarwood007: I'm fine with that.

jdarwood007:
I believe this is ready.  I'm doing a final round of verification.
I also fixed a issue where the mail queue viewer will show UTF-8 encoded subjects.

Sources/Subs-Post.php
```diff
@@ -1592,9 +1592,17 @@ function server_parse($message, $socket, $code, &$response = null)
 	if ($code === null)
 		return substr($server_response, 0, 3);
 
-	if (substr($server_response, 0, 3) != $code)
+	$resonse_code = (int) substr($server_response, 0, 3);
```
Sesquipedalian:
```suggestion
	$response_code = (int) substr($server_response, 0, 3);
```

Sources/Subs-Post.php
```diff
@@ -1592,9 +1592,17 @@ function server_parse($message, $socket, $code, &$response = null)
...
+	if ($resonse_code != $code)
```
Sesquipedalian:
```suggestion
	if ($response_code != $code)
```

Sources/Subs-Post.php
```diff
@@ -1592,9 +1592,17 @@ function server_parse($message, $socket, $code, &$response = null)
...
 	{
-		log_error($txt['smtp_error'] . $server_response);
+		// Ignoreable errors that we can't fix should not be logged.
+		/*
+		 * 550 - cPanel rejected sending due to DNS issues
+		 * 450 - DNS Routing issues
+		 */
+		if ($resonse_code < 500 && !in_array($resonse_code, [450]))
```
Sesquipedalian:
```suggestion
		if ($response_code < 500 && $response_code != 450))
```

Sesquipedalian (changes requested):

Sesquipedalian:
Made a few typo fixes

jdarwood007:
Looks like you broke the syntax.

Sources/Subs-Post.php
```diff
@@ -1592,9 +1592,17 @@ function server_parse($message, $socket, $code, &$response = null)
...
+	$response_code = (int) substr($server_response, 0, 3);
+	if ($response_code != $code)
 	{
-		log_error($txt['smtp_error'] . $server_response);
+		// Ignoreable errors that we can't fix should not be logged.
+		/*
+		 * 550 - cPanel rejected sending due to DNS issues
+		 * 450 - DNS Routing issues
+		 */
+		if ($response_code < 500 && $response_code != 450))
```
Sesquipedalian:
```suggestion
		if ($response_code < 500 && $response_code != 450)
```

Sesquipedalian:
> Looks like you broke the syntax.

I suppose that's what I get for using GitHub's web-based editor. Fixed now.

Sesquipedalian (approved):

PR #7789, opened 2023-07-08
Improve error logging while in SSI

jdarwood007:
When a error occurs on a remote script, the full url is not logged. For example.  Given a forum at:  https://example.com/forum/ and site content at https://example.com/about/

Some examples here:
A error at /about/ will log just the /about/ if no query string is provided. A error at /about/?foo=bar will log just ?foo=bar
A error at https://sub.example.com/ will log no url

SMF will then when displaying the error log, give false urls due to its handling of urls in the viewer.

This will for SSI, attempt to use the correct URLs, falling back to various methods and default responses.

This is draft for now, I am performing testing with this.

### DIFF
--- a/Sources/Errors.php
+++ b/Sources/Errors.php
@@ -79,14 +79,17 @@ function log_error($error_message, $error_type = 'general', $file = null, $line 
 		$user_info['ip'] = '';
 
 	// Find the best query string we can...
-	$query_string = empty($_SERVER['QUERY_STRING']) ? (empty($_SERVER['REQUEST_URL']) ? '' : str_replace($scripturl, '', $_SERVER['REQUEST_URL'])) : $_SERVER['QUERY_STRING'];
+	if (SMF == 'SSI')
+		$query_string = $_SERVER['REQUEST_URL'] ?? (($_SERVER['REQUEST_SCHEME'] ?? 'http') . '://' . ($_SERVER['SERVER_NAME'] ?? 'unknown') . '/' . ($_SERVER['REQUEST_URI'] ?? (($_SERVER['DOCUMENT_URI'] ?? $_SERVER['SCRIPT_NAME']	?? 'unknown.php') . !empty($_SERVER['QUERY_STRING']) ? '?' . $_SERVER['QUERY_STRING'] : '')));
+	else
+		$query_string = empty($_SERVER['QUERY_STRING']) ? (empty($_SERVER['REQUEST_URL']) ? '' : str_replace($scripturl, '', $_SERVER['REQUEST_URL'])) : $_SERVER['QUERY_STRING'];
 
 	// Don't log the session hash in the url twice, it's a waste.
 	if (!empty($smcFunc['htmlspecialchars']))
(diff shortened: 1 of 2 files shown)


Sources/ManageErrors.php
```diff
@@ -180,7 +180,7 @@ function ViewErrorLog()
 			'time' => timeformat($row['log_time']),
 			'timestamp' => $row['log_time'],
 			'url' => array(
-				'html' => $smcFunc['htmlspecialchars'](strpos($row['url'], 'cron.php') === false ? (substr($row['url'], 0, 1) == '?' ? $scripturl : '') . $row['url'] : $row['url']),
+				'html' => $smcFunc['htmlspecialchars'](strpos($row['url'], 'cron.php') === false && !strpos($row['url'], 'https://') && !strpos($row['url'], 'http://')  ? (substr($row['url'], 0, 1) == '?' ? $scripturl : '') . $row['url'] : $row['url']),
```
live627:
```suggestion
				'html' => $smcFunc['htmlspecialchars'](strpos($row['url'], 'cron.php') === false && !strpos($row['url'], 'https://') && !strpos($row['url'], 'http://')  ? ($row['url'][0] === '?' ? $scripturl : '') . $row['url'] : $row['url']),
```
perf: Check string inline; do not check copy

live627: How's that for premature optimization, lol

Sources/Errors.php
```diff
@@ -79,14 +79,17 @@ function log_error($error_message, $error_type = 'general', $file = null, $line
 		$user_info['ip'] = '';
 
 	// Find the best query string we can...
-	$query_string = empty($_SERVER['QUERY_STRING']) ? (empty($_SERVER['REQUEST_URL']) ? '' : str_replace($scripturl, '', $_SERVER['REQUEST_URL'])) : $_SERVER['QUERY_STRING'];
+	if (SMF == 'SSI')
```
live627:
```suggestion
	if (SMF === 'SSI')
```
perf:  Identity check is slightly faster

Sources/ManageErrors.php
jdarwood007: Looks good.  I didn't even think about it using more memory that way as its copying it.  Really wish PHP had kept {} as the way to navigate strings by character offset.  So confusing think its an array.  Yes I know technically in C/C++ a string is an array of chars, but still...

jdarwood007: I sent up a change for the strpos stuff I did.  http:// could exist somewhere in the string and break this.  So I check it starts with 0 and then flip the check. 

```
<?php
$url = 'https://google.com';

var_dump(strpos($url, 'cron.php') === false);
var_dump(!strpos($url, 'https://'));
var_dump(!strpos($url, 'http://'));

$url = 'bad-actor/https://google.com';

var_dump(strpos($url, 'cron.php') === false);
var_dump(!(strpos($url, 'https://') === 0));
var_dump(!(strpos($url, 'http://') === 0));
```

Issue #7173, opened 2021-12-01
MySQL 8.0 deprecation warnings


sbulen:
#### Description
Receiving the following warnings when restoring an SMF DB using MySQL 8.0.20:
```
WARNING: mysqli_warning  'utf8' is currently an alias for the character set UTF8MB3, but will be an alias for UTF8MB4 in a future release. Please consider using UTF8MB4 in order to be unambiguous.
WARNING: mysqli_warning  Integer display width is deprecated and will be removed in a future release.
WARNING: mysqli_warning  'NO_ZERO_DATE', 'NO_ZERO_IN_DATE' and 'ERROR_FOR_DIVISION_BY_ZERO' sql modes should be used with strict mode. They will be merged with strict mode in a future release. 
```

All simple fixes, just tedious.  Changes needed throughout installer & upgrader, & also in settings management for $db_character_set.  

### Steps to reproduce
1. Restore an SMF DB from a mysqldump export into a MySQL 8.0.20 DB

### Environment (complete as necessary)
- Version/Git revision: current
- Database Type: MySQL
- Database Version: 8.0.20
- PHP Version: 8.0.13

sbulen:
I really don't think this is a 2.1.0 requirement.  Could we remove or change the milestone?

sbulen:
I see this as being in 3 parts:
 - INT display width cleanup.  Soon is good, just not a 2.1.0 dependency.
 - STRICT mode cleanup.  Soon is good, just not a 2.1.0 dependency.
 - UTF8/UTF8MB3/UTF8MB4 clarifications.  Should be incorporated into #6409 .

Sesquipedalian:
I'm fine with changing the milestone, so long as no actual database changes will be required to fix this.

sbulen:
Some more reading regarding these changes:

Today, "utf8" is an alias for "utf8mb3", but that is going to change to "utf8mb4".  In fact, utf8mb3 itself - the SMF current default - is being deprecated:
https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-utf8.html

Benefits, including utf8mb4 is faster:
https://dev.mysql.com/blog-archive/mysql-8-0-when-to-use-utf8mb3-over-utf8mb4/

Integer display width is deprecated:
https://dev.mysql.com/worklog/task/?id=13127
https://dev.mysql.com/doc/refman/8.0/en/numeric-type-attributes.html
[...]

PR #7791, opened 2023-07-09
Installer & Upgrader int cleanup for mysql 8+

sbulen:
Partial for #7173

Integer display width is deprecated:
https://dev.mysql.com/worklog/task/?id=13127
https://dev.mysql.com/doc/refman/8.0/en/numeric-type-attributes.html
https://stackoverflow.com/questions/58938358/mysql-warning-1681-integer-display-width-is-deprecated

### DIFF
--- a/other/install_2-1_mysql.sql
+++ b/other/install_2-1_mysql.sql
@@ -1,5 +1,4 @@
 #### ATTENTION: You do not need to run or use this file!  The install.php script does everything for you!
-#### Install script for MySQL 4.0.18+
 
 #
 # Table structure for table `admin_info_files`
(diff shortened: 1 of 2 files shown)


live627:
This seems to me 2.2 territory. We don't introduce database changes in  patches, unless some policy change happened...?

sbulen:
We're good.  It's not a structural change.  No code impact.  These values aren't used at all.

If it is technically in violation of a policy somewhere, we can rewrite that policy.

jdarwood007:
I believe if you upgrade to mysql 8 with these, they are just ignored.  So I agree its just aligning with MySQL 8 (and technically other RDBMS) to ignore the size on the int based fields, since they already explicitly have a size.  MySQL did some funky logic with how to display it with the size specified and zerofill specified.  See for more history https://dev.mysql.com/doc/refman/8.0/en/numeric-type-attributes.html

jdarwood007 (approved):

PR #7794, opened 2023-07-11
Implant PHP SessionHandlerInterface

### LINKED ISSUE #7793: PHP 8.3  session_set_save_handler 6 arguments deprecated
#### Description
The PHP team has deprecated calling session_set_save_handler with 6 or more arguments in 8.3
https://wiki.php.net/rfc/deprecate_functions_with_overloaded_signatures#session_set_save_handler

We call this in 
Session.php
```
			session_set_save_handler('sessionOpen', 'sessionClose', 'sessionRead', 'sessionWrite', 'sessionDestroy', 'sessionGC');
```

The suggestion is to use a object or implant SessionHandlerInterface (PHP 5.4+)

This does affect 2.0 as well

jdarwood007:
Fixes #7793

### DIFF
--- a/Sources/Session.php
+++ b/Sources/Session.php
@@ -25,7 +25,7 @@ if (!defined('SMF'))
  */
 function loadSession()
 {
-	global $modSettings, $boardurl, $sc, $smcFunc, $cache_enable;
+	global $context, $modSettings, $boardurl, $sc, $smcFunc, $cache_enable;
 
 	// Attempt to change a few PHP settings.
 	@ini_set('session.use_cookies', true);
@@ -68,7 +68,10 @@ function loadSession()
 			@ini_set('session.serialize_handler', 'php_serialize');
 			if (ini_get('session.serialize_handler') != 'php_serialize')
 				@ini_set('session.serialize_handler', 'php');
-			session_set_save_handler('sessionOpen', 'sessionClose', 'sessionRead', 'sessionWrite', 'sessionDestroy', 'sessionGC');
+			
+			$context['session_handler'] = new SmfSessionHandler();
+			session_set_save_handler($context['session_handler'], true);
+
 			@ini_set('session.gc_probability', '1');
 		}
 		elseif (ini_get('session.gc_maxlifetime') <= 1440 && !empty($modSettings['databaseSession_lifetime']))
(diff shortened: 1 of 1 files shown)


Sources/Session.php
```diff
@@ -95,151 +98,157 @@ function loadSession()
...
-		if (!empty($db_port))
-			$db_options['port'] = $db_port;
+	/**
+	 * Implementation of SessionHandler::write() replacing the standard write handler.
+	 *
+	 * @param string $id The session ID
+	 * @param string $data The data to write to the session
+	 * @return boolean Whether the info was successfully written
+	 */
+	#[\ReturnTypeWillChange]
```
live627, line 170: Weren't attributes only introduced in PHP 8.2? Maybe do a version check to keep using the old code in older php versions, similar to the db layer in 2.0.14.

jdarwood007: Correct, but # is also supported for comments.  So for anything below 8.2, its ignored as a comment.  For 8.2, this is telling PHP to ignore the fact the signature doesn't match. 8.1 and 8.0 don't seem to care the signature doesn't match the implantation interfaces.  Thus left 7.x, which complained about the signatures not matching since it was type hinted and the interface in 7.x was without type hints.

Sources/Session.php
```diff
@@ -95,151 +98,157 @@ function loadSession()
...
-	);
-
-	return ($smcFunc['db_affected_rows']() == 0 ? false : true);
+	/**
+	 * Validates a given string conforms to our testing for a valid session id.
+	 *
+	 * @param string $id The session ID
+	 * @return boolean Whether the string is valid format or not
+	 */
+	private function IsValidSessionID(string $id): bool
```
dragomano: Why not `isValidSessionID`? It's just a method, not a class.

Sources/Session.php
```diff
@@ -95,151 +98,157 @@ function loadSession()
...
+		// Clean up after yerself ;).
+		$session_update = $smcFunc['db_query']('', '
+			DELETE FROM {db_prefix}sessions
+			WHERE last_update < {int:last_update}',
+			array(
+				'last_update' => time() - $max_lifetime,
+			)
+		);
+
+		return ($smcFunc['db_affected_rows']() == 0 ? false : true);
```
dragomano: `return $smcFunc['db_affected_rows']() !== 0;` or `return $smcFunc['db_affected_rows']() > 0;` is shorter

Sources/Session.php
jdarwood007: Fixed

Sources/Session.php
jdarwood007: I most likely copied from a previous function and didn't check the return type properly.  Fixed

dragomano (approved):

Sesquipedalian (approved):

PR #7795, opened 2023-07-11
[2.1] Add a requirements file

jdarwood007:
Doing this because I have to keep looking it up every time I want to know.  It really should be in our repo in addition to our wiki/site.  Also gives a place for our wiki/site to make references/get updated from when we update here.

### DIFF
--- a/other/requirements.md
+++ b/other/requirements.md
@@ -1,0 +1,57 @@
+# SMF Miniumn Requirements
+
+## PHP
+### PHP Version Support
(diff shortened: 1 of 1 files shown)


sbulen:
I always use the online manual!

frandominguezl:
Step forward into modernizing the repo and making it more contributor friendly 👍🏼

live627:
Next step: security policy

Issue #7539, opened 2022-08-17
Board disappears if moved


BrickOzp:
#### Description
When a board is moved with the Move Board functions it is not shown to regular users.

### Steps to reproduce
1. Admin -> Forum -> Board and Categories -> Modify Boards
2. Select any existing board and press "MOVE"
3. Move the board to a different position
4. View board index as regular user

Expected
5. The board is visible

Actual
5. The board is not visible

### Environment (complete as necessary)
- Version/Git revision: 2.1.2
- Database Type: Mysql
- Database Version:
- PHP Version: 7.4

### Additional information/references

Reported here https://www.simplemachines.org/community/index.php?topic=583120

BrickOzp:
The problem seems to be that all entries for the board is removed from board_permissions_view when the board is moved.
But it is only populated again if $boardOptions['deny_groups'] or $boardOptions['access_groups'] is passed to modifyBoard().

PR #7802, opened 2023-07-20
[2.1] Only delete board_permissions_view if we have modified the access

jdarwood007:
This bug keeps reappearing in various ways.

To reproduce this one, simply move a board.  When you do, only the new position for the board is passed.  Resulting in modifyBoard deleting the board_permissions_view data, but has no data to rebuild with.  Simply modifying the board and saving again rebuilds the access.

@sbulen, @live627, @BrickOzp, @albertlast can you all test this

Related #7689
Related #7692
Related #7539
Related #7540
Related #5514
Related #5515
Related #5265
Related #5152
Related #5153
Related #4727

### DIFF
--- a/Sources/Subs-Boards.php
+++ b/Sources/Subs-Boards.php
@@ -672,13 +672,14 @@ function modifyBoard($board_id, &$boardOptions)
 		);
 
 	// Before we add new access_groups or deny_groups, remove all of the old entries
-	$smcFunc['db_query']('', '
-		DELETE FROM {db_prefix}board_permissions_view
-		WHERE id_board = {int:selected_board}',
-		array(
-			'selected_board' => $board_id,
-		)
-	);
+	if (isset($boardOptions['access_groups']) || isset($boardOptions['deny_groups']))
+		$smcFunc['db_query']('', '
+			DELETE FROM {db_prefix}board_permissions_view
+			WHERE id_board = {int:selected_board}',
+			array(
+				'selected_board' => $board_id,
+			)
+		);
 
 	if ($board_permissions_inserts != array())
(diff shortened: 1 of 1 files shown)


Flavio93Zena:
Fix confirmed on vanilla test forum, boards moving correctly 5/5 tested - Ninja ZX-10RR on sm.

live627:
I won't be available for a week. Going for a road trip soon to visit my
brother

jdarwood007:
np,  You had previously touched the code here in another PR.  So wanted to ensure you didn't see anything wrong or you tested something I may have not tested.

albertlast:
Well your changes look similiar to the old pr: https://github.com/SimpleMachines/SMF/commit/8c03499a40771d767f2e534f189b4fa519df2dbc

so instead being "smart" and try to guess which part needs to be deleted we could truncate the table and rebuild complete,
like we do in the upgrade process:
https://github.com/SimpleMachines/SMF/blob/fd30b3714c17b30bb0821d79268caa8b9462a9c9/other/upgrade_2-1_mysql.sql#L3418-L3452

jdarwood007:
They are similar, but instead of checking if they contain anything, I just check to see if its set, which means it was passed into.  It also means that we did update those.

Technically this could possibly have a error if you only passed access or deny groups (and not both).

We could do the insert select, as long as postgresql follows the same syntax.

albertlast:
On the first glance it look similiar:
https://github.com/SimpleMachines/SMF/blob/fd30b3714c17b30bb0821d79268caa8b9462a9c9/other/upgrade_2-1_postgresql.sql#L3554-L3588

live627 (approved):

Issue #7805, opened 2023-07-23
Inconsistent/wrong uses of "forums"


m4z:
I don't currently have Git set up on this machine, so here's a dump for myself or anybody else to pick up:

* [Inconsistent use of "forums", should be "forum".](https://github.com/SimpleMachines/SMF/blob/fd30b3714c17b30bb0821d79268caa8b9462a9c9/Themes/default/languages/Help.english.php#L463)
* [Indicates possession, should be "forum's"](https://github.com/SimpleMachines/SMF/blob/fd30b3714c17b30bb0821d79268caa8b9462a9c9/Themes/default/languages/ManageSettings.english.php#L256) (though I'm never sure with these)
* [Indicates possession, should be "forum's"](https://github.com/SimpleMachines/SMF/blob/fd30b3714c17b30bb0821d79268caa8b9462a9c9/Themes/default/languages/ManageMaintenance.english.php#L163)

Sesquipedalian:
You are correct in all three cases. Please submit a PR for them.

jdarwood007:
#7807 fixed these

PR #7807, opened 2023-07-26
Fix possessive form

m4z:
Adresses issue 2/3 and 3/3 described in #7805

### DIFF
--- a/Themes/default/languages/ManageMaintenance.english.php
+++ b/Themes/default/languages/ManageMaintenance.english.php
@@ -160,7 +160,7 @@ $txt['maintain_run_now'] = 'Run task now';
 $txt['maintain_return'] = 'Back to Forum Maintenance';
 
 $txt['maintain_backup'] = 'Backup Database';
-$txt['maintain_backup_info'] = 'Download a backup copy of your forums database in case of emergency.';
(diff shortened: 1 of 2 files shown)


DiegoAndresCortes (approved):

jdarwood007:
Crowdin won't blow up and get this wrong because of the escaped character right?

DiegoAndresCortes:
We have others like
https://github.com/SimpleMachines/SMF/blob/fd30b3714c17b30bb0821d79268caa8b9462a9c9/Themes/default/languages/ManageBoards.english.php#L92

So probably safe.

sbulen:
The gotcha with escaping & control characters in language strings is I recall .js vs .php needing things a bit different.  And email templates might be different as well...

We have had "LeT's mAKe tHinGs ConSIsTEnt!!!" PRs break things in the past.

Testing is critical.

I only believe it's working when I see it...

Screenshots, please.

m4z:
* `$txt['maintain_backup_info']`:
Apparently, the backup functionality was completely removed in 63f90bfa813bddf7f885f953d64387d4755b0e91. Should I change this PR to remove all `maintain_backup*` strings or do a separate PR? Should I undo touching the string in this PR?


* `$txt['custom_profile_desc']`:
Screenshot_20230727_075426

Sesquipedalian (approved):

PR #7814, opened 2023-07-30
Remove inconsistent double spaces in some strings.

m4z:
I assume they aren't visible in HTML anyway, but I haven't verified this.


jdarwood007:
https://www.grammarly.com/blog/spaces-after-period/#:~:text=How%20many%20spaces%20come%20after,space%20is%20correct%2C%20not%20two.

I am one who does this, I got in a habit of it. Hard to break now. Good reminder though, I will try not to do this.

m4z:
Even if it was the right thing to do, 99% of strings do not use it.

Sesquipedalian:
Double spaces between sentences are indeed a relic of a bygone era and should be avoided. However, I don't think it is worthwhile to change them in existing strings. As noted above, HTML collapses them to a single space anyway, so it doesn't make a material difference to output. But moreover, if we change the strings in the English files, that will trigger Crowdin to mark them as needing re-translation for all the other languages. Even if the translators won't have to do anything more than click some buttons to deal with that, why waste their time for something that makes no difference?

live627:
I believe this convention is only for fixed width typefaces.

m4z:
> But moreover, if we change the strings in the English files, that will trigger Crowdin to mark them as needing re-translation for all the other languages. Even if the translators won't have to do anything more than click some buttons to deal with that, why waste their time for something that makes no difference?

Currently, I believe Crowdin forces translators to include the double space in their translation. I'm not 100% sure, but I believe only managers can override the warning about the missing double space in the translation. I have no way to test that, though, but I'm relatively sure I have experienced this in other Crowdin projects, where I'm not a manager.

Sesquipedalian:
Again, why waste their time for something that makes no difference?

m4z:
My goal is consistency. We're talking about wasted seconds for every language, and we currently have 16 languages with >90% translated.

DiegoAndresCortes:
Correcting the strings on Crowdin will take a couple of minutes tops, I'm not opposed to this change.

jdarwood007:
@m4z Can you retarget this for 3.0?  I am in the boat that we leave it for 2.1, but I see no issue targeting it with 3.0 and the other language improvements we have done.

PR #7820, opened 2023-08-04
Shows multiple events on the same day correctly

### LINKED ISSUE #7803: Calendar - events dropped when there are multiple overlapping events
#### Description
Odd one from the forum...  But it can be verified.  If you have multiple overlapping events, then the various date range views seem to inexplicably drop some of the events.  

Any event on its own is displayed perfectly, no matter what the view.  But as soon as there are multiple, some may get dropped.

### Steps to reproduce
0. Forum & user timezone to EDT
1. Add an all-day event, for 3 days, from Aug 3 - Aug 5
2. Add an all-day event, on Aug 5.
3. Choose list view, Aug 3 - Aug 10, both events show properly for Aug 5
4. Choose list view, Aug 5 - Aug 12, only the 3 day event is shown.  The Aug 5 event is dropped from the list...

There are similar variations in month view, week view, etc.  

- Version/Git revision: 2.1.4
- Database Type: mysql
- Database Version: 8.0
- PHP Version: 8.2

From the forum:  https://www.simplemachines.org/community/index.php?topic=586490.0

Sesquipedalian:
Fixes #7803

### DIFF
--- a/Sources/Subs-Calendar.php
+++ b/Sources/Subs-Calendar.php
@@ -175,7 +175,7 @@ function getEventRange($low_date, $high_date, $use_permissions = true)
 		$start_date_string = date_format($start_object, 'Y-m-d');
 		$end_date_string = date_format($end_object, 'Y-m-d');
 
-		$cal_date = ($start_object >= $low_object) ? $start_object : $low_object;
+		$cal_date = ($start_object >= $low_object) ? (clone $start_object) : (clone $low_object);
 		while ($cal_date <= $end_object && $cal_date <= $high_object)
 		{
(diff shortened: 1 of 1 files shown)


jdarwood007:
I didn't even realize DateTime in PHP could be compared, given the lack of documentation showing it has comparison operators.  That is another one for PHP hidden logic.

Sesquipedalian:
It's not entirely hidden: https://www.php.net/manual/en/datetime.diff.php#example-2094
It could be more obvious, though.

live627:
they should expose that like in c++

indyrcpilot:
It looks like it fixed the problem with events not showing up!
Thanks!

PR #7822, opened 2023-08-09
[2.1] Hide PHP errors from failing to remove files/directories while removing a theme

jdarwood007:
Also.  If you choose to install a theme by a path on the server and it fails, the logic tries to remove the directory.  I'm not certain of the proper fix.  The logic is done in the wrong place.  The get_theme_info does the deleting, instead of the higher up functions.

### DIFF
--- a/Sources/Subs-Themes.php
+++ b/Sources/Subs-Themes.php
@@ -487,12 +487,12 @@ function remove_dir($path)
 					remove_dir($path . '/' . $object);
 
 				else
-					unlink($path . '/' . $object);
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
@jdarwood007, this PR targets release-2.1, but you assigned it to the milestones for SMF 3.0. Does this need to be resubmitted to target release-3.0, does it need to be reassigned to a milestone for SMF 2.1, or both?

Sesquipedalian (approved):

PR #7825, opened 2023-08-11
Do not attempt to convert backup tables to UTF-8

### LINKED ISSUE #7823: Upgrade converting UTF-8 may sometimes convert backup tables
#### Description
I noticed this while doing a test upgrade on a forum
Screen Shot 2023-08-09 at 12 41 03 PM

The issue is because the forum has no prefix.

### Steps to reproduce
1. Use no prefix on `$db_prefix = '';`
2.  Perform a upgrade, instructing to create backup tables
3. During the UTF-8 conversion, backup tables are converted.

- Version/Git revision: 2.1.4
- Database Type: MySQL
- Database Version: 8.0
- PHP Version: 8.1

I don't know if we can come up with a fix at the moment.  This is a really unique edge case.  I'm documenting it at least so we know this occurs.

jdarwood007:
Fixes #7823

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -3252,6 +3252,10 @@ function ConvertUtf8()
 		db_extend();
 		$queryTables = $smcFunc['db_list_tables'](false, $db_prefix . '%');
 
+		$queryTables = array_values(array_filter($queryTables, function($v){
+			return !(stripos($v, 'backup_') === 0);
+		}));
+
 		$upcontext['table_count'] = count($queryTables);
 
 		// What ones have we already done?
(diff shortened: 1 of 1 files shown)


other/upgrade.php
```diff
@@ -3252,6 +3252,10 @@ function ConvertUtf8()
 		db_extend();
 		$queryTables = $smcFunc['db_list_tables'](false, $db_prefix . '%');
 
+		$queryTables = array_values(array_filter($queryTables, function($v){
+			return !(stripos($v, 'backup_') === 0);
```
live627:
```suggestion
			return stripos($v, 'backup_') !== 0;
```

just a silly change

jdarwood007: Well that could mean it would match `my_backup_table`.  Instead of just matching `backup_table`.  That was my intention at least.

Sesquipedalian: No, they are equivalent statements.

jdarwood007: oh yea, its still strict type checking.

live627 (approved):

PR #7827, opened 2023-08-15
SMTP sometimes may send 451 errors.

jdarwood007:
I'm seeing this now.  Wasn't seeing it earlier.  Not sure what changed.  But the SMTP sending server seems to be generating this, not the receiving server.


Sesquipedalian (changes requested):
SMF's coding style uses the full `array()` syntax rather than the `[]` short syntax. Otherwise, looks good.

Sesquipedalian (approved):

PR #7839, opened 2023-09-22
Allow ignoring file errors if requested

jdarwood007:
This is simple.
The following will have SMF say an error will exist during package install

```
		<remove-file name="$sourcedir/Test.php" error="ignore" />
```

With this change, SMF ignores this as we requested in the xml.


live627 (approved):

Sesquipedalian:
Oops. This targeted release-2.1, but was on the milestone for 3.0 Alpha 4. @jdarwood007, can you please submit a version of this PR targeting release-3.0?

PR #7840, opened 2023-09-25
code tag changes

live627:
- wrap code in preformatted html tags
- do the same in the sceditor
- bonus: fix php tag processing in the sceditor
- php 8.3 changes highlight_string(), so update our integrations
- do not add php start and end php tags to the final output of the php bbcode
- add buttons for selecting and expanding code blocks with js; don't output them from php
- remove smfSelectText( ) as it is unused now

(cherry picked from commit f9d90a8f7f299f10be8160ca0b6a3a062390c3e6)

### DIFF
--- a/Sources/Subs.php
+++ b/Sources/Subs.php
@@ -1860,75 +1860,55 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 			),
 			array(
 				'tag' => 'code',
-				'type' => 'unparsed_content',
-				'content' => '<div class="codeheader"><span class="code floatleft">' . $txt['code'] . '</span> <a class="codeoperation smf_select_text">' . $txt['code_select'] . '</a> <a class="codeoperation smf_expand_code hidden" data-shrink-txt="' . $txt['code_shrink'] . '" data-expand-txt="' . $txt['code_expand'] . '">' . $txt['code_expand'] . '</a></div><code class="bbc_code">$1</code>',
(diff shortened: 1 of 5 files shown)


Sources/Subs.php
```diff
@@ -1860,75 +1860,55 @@ function parse_bbc($message, $smileys = true, $cache_id = '', $parse_tags = arra
 			),
 			array(
 				'tag' => 'code',
-				'type' => 'unparsed_content',
-				'content' => '<div class="codeheader"><span class="code floatleft">' . $txt['code'] . '</span> <a class="codeoperation smf_select_text">' . $txt['code_select'] . '</a> <a class="codeoperation smf_expand_code hidden" data-shrink-txt="' . $txt['code_shrink'] . '" data-expand-txt="' . $txt['code_expand'] . '">' . $txt['code_expand'] . '</a></div><code class="bbc_code">$1</code>',
-				// @todo Maybe this can be simplified?
-				'validate' => isset($disabled['code']) ? null : function(&$tag, &$data, $disabled) use ($context)
+				'type' => 'unparsed_content',,
```
jdarwood007: Interesting that older PHP doesn't like the double comma.  I'm sure thats a typoe.

jdarwood007:
Is this targeted at 2.1?

Sources/Subs.php
live627:
```suggestion
				'type' => 'unparsed_content',
```

live627:
No

PR #7846, opened 2023-11-04
Objectification

Sesquipedalian:
Refactors SMF to use the object oriented code paradigm.


Sesquipedalian:
Bleh. I chose the wrong target branch.

PR #7849, opened 2023-11-05
Fix the installer/upgrader

jdarwood007:
```
( ! ) Fatal error: Uncaught Error: Typed static property SMF\Db\DatabaseApi::$db must not be accessed before initialization in /Sources/Config.php on line 1326
( ! ) Error: Typed static property SMF\Db\DatabaseApi::$db must not be accessed before initialization in /Sources/Config.php on line 1326
Call Stack
#	Time	Memory	Function	Location
1	0.0055	1343864	{main}( )	.../upgrade.php:0
2	0.0213	1703288	SMF\Config::generateSeed( )	.../upgrade.php:277
3	0.0213	1703664	SMF\Config::updateModSettings( $change_array = ['rand_seed' => 1108340315], $update = ??? )	
```

This is occuring because `$modSettings['rand_seed']` is not set.  All of `$modSettings` is not set, besides the default theme that gets set later in upgrade.php

This logic is called by this in ugprade.php
```
// Old DBs won't have this
if (!isset(Config::$modSettings['rand_seed']))
	Config::generateSeed();
```

We only reload the data when ssi is in the url, so this will always reseed the settings every time.
```
// Are we going to be mimic'ing SSI at this point?
if (isset($_GET['ssi']))
{
	User::load();
	User::$me->loadPermissions();
	Config::reloadModSettings();
}
```

Note also prior to this we call loadEssentialData, which also attempts to make use of $modSettings.

Well digging back further I locate this call:
```
		return throw_error(sprintf(Lang::$txt['error_sourcefile_missing'], 'Db/APIs/' . Config::$db_type . '.php'));
```
Change that over to a die and..

```
Unable to find the Sources/Db/APIs/mysql.php file. Please make sure it was uploaded properly, and then try again.
```

It can't find it because in my Settings.php, `$db_type = 'mysql';`

To work around this, added a helper method in DatabaseApi to determine the type based on the passed type.  Choose to pass a type so we can always get back a result even for other database types.

This also does fix the seed issue since the database is now able to connect.


jdarwood007:
I'm on a linux os and its case sensitive:
```
Linux, the UNIX system, and Windows® observe different conventions for case-sensitivity in file name lookup: Linux and the UNIX system are case sensitive; Windows is not case sensitive.
```

As well, all I did so far is copy my 2.1 forum into a new database and directory, then uploaded the 3.0 files.

Sesquipedalian:
I haven't tested yet, but it looks good to the eye. 

One request: please change the name of the method to something more descriptive, such as `getClass`. Generally speaking, property names should be nouns while method names should use verbs. There's no programming reason for this; it just makes it easier for humans reading the code later to understand. It's the reason why, for example, PHP offers `\DateTime::getTimestamp()` rather than `\DateTime::timestamp()`.

jdarwood007:
Done.  I've also included some other fixes that I needed to get the ugprade.php to run.

Sesquipedalian:
Nice. I will look forward to testing as soon as I can. It might take me a little to get to it, though; this week is going to be a _very_ busy one for me. Of course, someone else can always test and approve instead.

jdarwood007:
Added fixes for #7852 
The autoloader needs to ignore trying to load integrations during install.

Sesquipedalian:
Conflicts need to be fixed.

Sources/Autoloader.php
```diff
@@ -40,7 +40,7 @@
 		$sourcedir = __DIR__;
 
 	// Do any third-party scripts want in on the fun?
-	if (class_exists('SMF\\Config', false) && $hook_value !== (Config::$modSettings['integrate_autoload'] ?? ''))
+	if (defined('SMF') && !defined('SMF_INSTALLING')  && class_exists('SMF\\Config', false) && $hook_value !== (Config::$modSettings['integrate_autoload'] ?? ''))
```
Sesquipedalian: I don't think any scenario exists in which the `SMF` constant wouldn't be defined by the time we got here. Am I missing something?

other/install.php
```diff
@@ -868,9 +868,9 @@ function DatabaseSettings()
 		Config::load();
 
 		// Better find the database file!
-		if (!file_exists(Config::$sourcedir . '/Db/APIs/' . Config::$db_type . '.php'))
+		if (!file_exists(Config::$sourcedir . '/Db/APIs/' . Db::getClase(Config::$db_type) . '.php'))
```
Sesquipedalian: Typo

other/install.php
```diff
@@ -868,9 +868,9 @@ function DatabaseSettings()
...
 		{
-			$incontext['error'] = sprintf(Lang::$txt['error_db_file'], 'Db/APIs/' . Config::$db_type . '.php');
+			$incontext['error'] = sprintf(Lang::$txt['error_db_file'], 'Db/APIs/' . Db::getClase(Config::$db_type) . '.php');
```
Sesquipedalian: Typo

Sesquipedalian (changes requested):
Ran into problems due to typos when testing the installer.

After correcting the typos in my local test copy, I then received this error during step 2 of the install:
```
Fatal error: Uncaught TypeError: mysqli_query(): Argument #1 ($mysql) must be of type mysqli, null given in /Users/jon/Sites/temp/Sources/Db/APIs/MySQL.php:237 Stack trace: #0 /Users/jon/Sites/temp/Sources/Db/APIs/MySQL.php(237): mysqli_query(NULL, '\n\t\t\tSELECT VERS...', 0) #1 /Users/jon/Sites/temp/Sources/Db/APIs/MySQL.php(1066): SMF\Db\APIs\MySQL->query('', '\n\t\t\tSELECT VERS...', Array) #2 /Users/jon/Sites/temp/Sources/Db/APIs/MySQL.php(1933): SMF\Db\APIs\MySQL->get_version() #3 /Users/jon/Sites/temp/Sources/Db/DatabaseApi.php(334): SMF\Db\APIs\MySQL->__construct(Array) #4 /Users/jon/Sites/temp/install.php(882): SMF\Db\DatabaseApi::load(Array) #5 /Users/jon/Sites/temp/install.php(178): DatabaseSettings() #6 {main} thrown in /Users/jon/Sites/temp/Sources/Db/APIs/MySQL.php on line 237
```
I have not tried to investigate that error further.

jdarwood007:
I can't get your error to occur.  But that sounds like a mysql error occurred, most likely in the database connection since mysqli param should be the connection resource.

I'm even more certain seeing the stacktrace shows myqi_query first param is null.  The connection failed but it continued. Need to investigate to ensure that the installer does not allow/try to continue if the database connection fails.

tyrsson:
Just as a FYI, I ran the install late last night. 

Windows 10, php 8.2, 10.10.2-MariaDB

I did not tie it to an existing DB. Install ran without issue.

jdarwood007:
See my earlier note about case sensitivity.  Windows won't see this issue.  Only linux/unix based oses.

tyrsson:
I saw it, I was just confirming it runs in windows and providing what its running on ;)

Sources/Db/APIs/MySQL.php
```diff
@@ -2020,23 +2020,26 @@ protected function __construct(array $options = [])
 			$this->prefixReservedTables();
 		}
 
-		$this->get_version();
-		$this->supports_pcre = version_compare($this->version, strpos($this->version, 'MariaDB') !== false ? '10.0.5' : '8.0.4', '>=');
+		if (empty($this->connection) && !$non_fatal) {
+			$this->get_version();
+			$this->supports_pcre = version_compare($this->version, strpos($this->version, 'MariaDB') !== false ? '10.0.5' : '8.0.4', '>=');
+		}
```
Sesquipedalian: Why have these lines been wrapped in this `if` block? The test here, `(empty($this->connection) && !$non_fatal)` will never be true, because if it were we would already have bailed out higher up. As a result, the lines of code in this block will never be executed.

```suggestion
		$this->get_version();
		$this->supports_pcre = version_compare($this->version, strpos($this->version, 'MariaDB') !== false ? '10.0.5' : '8.0.4', '>=');
```

Sources/Db/APIs/MySQL.php
```diff
@@ -2020,23 +2020,26 @@ protected function __construct(array $options = [])
...
-		);
+		if (empty($this->connection) && !$non_fatal)
+			$this->query(
+				'',
+				'
+				SET NAMES {string:db_character_set}',
+				[
+					'db_character_set' => $this->character_set,
+				],
+			);
```
Sesquipedalian: As above, why have these lines been wrapped in this `if` block? The test here, `(empty($this->connection) && !$non_fatal)` will never be true, because if it were we would already have bailed out higher up. As a result, the lines of code in this block will never be executed.

```suggestion
		$this->query(
			'',
			'SET NAMES {string:db_character_set}',
			[
				'db_character_set' => $this->character_set,
			],
		);
```

Sesquipedalian (changes requested):

Sources/Db/APIs/MySQL.php
jdarwood007: Oh, I guess it should be !empty.  If there isn't a database connection, we can't exactly get a database version/pcre info.

Sources/Db/APIs/MySQL.php
jdarwood007: As mentioned, if the connection fails it would still try to execute this.  I may have the if logic wrong if it doesn't make sense.

Sesquipedalian: `if (!empty($this->connection)) { ... }` would make a lot more sense, yes. 

In fact, I suggest putting everything from `$this->get_version();` through to the end of the `SET NAMES` query inside an `if` block with that test, like so:

```
		if (!empty($this->connection)) {
			$this->get_version();
			$this->supports_pcre = version_compare($this->version, strpos($this->version, 'MariaDB') !== false ? '10.0.5' : '8.0.4', '>=');

			// For backward compatibility.
			if (!is_object(self::$db_connection)) {
				self::$db_connection = $this->connection;
			}

			// Ensure database has UTF-8 as its default input charset.
			$this->query(
				'',
				'SET NAMES {string:db_character_set}',
				[
					'db_character_set' => $this->character_set,
				],
			);
		}
```

Sesquipedalian: Should do the same in `SMF\Db\APIs\PostgreSQL::__construct()`, too.

jdarwood007: Would it be better to just do something like:

```
if (empty($this->connection) {
    return;
}
```

No need to nest the rest, just bail at this point, no need to execute anything.  I believe my code was attempting to mimic the earlier code, but I didn't make the right changes in regards to handling the `$non_fatal`.  I think its best to ignore it and yes if we don't have a connection, just stop.

Maybe even this:

```
		if (empty($this->connection) && !$non_fatal) {
			ErrorHandler::displayDbError();
		}
/* ADDED THIS */
		elseif (empty($this->connection)) {
			return;
		}
```

Essentially make it so we do the following checks
1. Check if we are missing a connection and have a fatal connection, if both are true; display error
2. Check if we have a missing connection, if true; return, no further logic can be done.

Only other logic being skipped is the SSI logic, but this won't mater really since there is no database connection.

Sources/Db/APIs/MySQL.php
```diff
@@ -1984,18 +1984,24 @@ protected function __construct(array $options = [])
...
 		// For backward compatibility.
 		if (!is_object(self::$db_connection)) {
 			self::$db_connection = $this->connection;
 		}
 
+		// At this point, if we don't have a connection, nothing else can be done.
+		if (empty($this->connection)) {
+			return;
+		}
+
```
```suggestion
		// At this point, if we don't have a connection, nothing else can be done.
		if (empty($this->connection)) {
			return;
		}

		// For backward compatibility.
		if (!is_object(self::$db_connection)) {
			self::$db_connection = $this->connection;
		}

```

There's no point trying to set `self::$db_connection` if `$this->connection` is empty.

Sources/Db/APIs/MySQL.php
```diff
@@ -1984,18 +1984,24 @@ protected function __construct(array $options = [])
...
+		$this->get_version();
+		$this->supports_pcre = version_compare($this->version, strpos($this->version, 'MariaDB') !== false ? '10.0.5' : '8.0.4', '>=');
+
 		// Ensure database has UTF-8 as its default input charset.
 		$this->query(
 			'',
-			'SET NAMES {string:db_character_set}',
+			'
+			SET NAMES {string:db_character_set}',
```
Sesquipedalian: Please revert this change.

Sources/Db/APIs/PostgreSQL.php
```diff
@@ -2117,6 +2117,11 @@ protected function __construct(array $options = [])
 			self::$db_connection = $this->connection;
 		}
 
+		// At this point, if we don't have a connection, nothing else can be done.
+		if (empty($this->connection)) {
+			return;
+		}
+
```
Sesquipedalian, line 2119: As with MySQL.php, this should go before the attempt to set `self::$db_connection`.

Sources/Db/APIs/MySQL.php
Sesquipedalian: Makes sense. See my new review.

Sesquipedalian (changes requested):

Sources/Db/APIs/MySQL.php
jdarwood007: My bad, bad rebase.

Sesquipedalian (approved):
Works nicely for me. :)

PR #7851, opened 2023-11-05
Calls integrate_autoload hook only once

### LINKED ISSUE #7850: 3.0 autoloader calls integrate_autoload hook needlessly
SMF 3.0's autoloader calls the integrate_autoload hook every time. It should only need to do that once.

This behaviour could cause a performance hit.

Sesquipedalian:
Fixes #7850

### DIFF
--- a/Sources/Autoloader.php
+++ b/Sources/Autoloader.php
@@ -20,9 +20,9 @@ namespace SMF;
  */
 spl_autoload_register(function ($class)
 {
-	static $level = 0;
+	static $hook_value = '';
 
-	$class_map = array(
+	static $class_map = array(
 		// Some special cases.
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
We should probably merge this regardless, but I'd be curious to know what difference it makes to your performance benchmarks, @albertlast.

jdarwood007 (approved):
Looks good.  Only comment is if a integration hook is added temporarily live, it won't be captured here.

Sesquipedalian:
> Looks good. Only comment is if a integration hook is added temporarily live, it won't be captured here.

Good catch. I've just pushed a change that should help with that.

jdarwood007 (approved):

PR #7853, opened 2023-11-05
[3.0] Languages Overhaul

jdarwood007:
Move the languages directory form `/Themes/default/languages` to `/Languages`
Additionally, this also moves agreement.txt from `/` to `/Languages`

- We introduce a new config variable for Settings.php, `$languagesdir` which defaults to `__DIR__ . '/Languages'`
- We now use the locale instead of a custom name
- Languages are now in a sub-directory. I.e. `/Languages/en_US/`
- We no longer use a suffix on filenames (i.e. we no longer have index.english.php)
- Rename index to General
- This does not change that we can still have languages in `/Themes/*/languages`, we do deprecate the usage of this for override.
- Upgrader will attempt to clean up the old languages directory, it's a best effort.  I coded this in a way for us to add additional cleanup steps in the future.
- We will rename `Settings.{language}` to `{locale}/ThemeStrings` to encourage themes to use this as an alternative to modifying base languages

@DiegoAndresCortes Can you review the theme's logic is sound.

Todo:

- [x] Upgrade's convert UTF-8 logic needs to handle the new locales
- [x] Test installing a mod that is using `$themesdir/default/languages`
- [x] TSources/PackageManager/SubsPackage.php still calls `$languagedir` (and `$languages_dir`) with the old paths.  The issue is how we handle the logic for migration to the 'new' paths.
- [x] Test installing a mod that uses `$languagedir`


Sesquipedalian:
I've moved this to the 3.0 Alpha 2 milestone. For Alpha 1, the only goals are:

- Merge OOP code
- Find and fix bugs introduced via OOP refactoring. (Goal is to have all functionality just like in SMF 2.1.)
- Basic installer and upgrader updates. (Create upgrade_3.0_*.sql files, etc.)

Sesquipedalian:
The syntax check failed, but that should be an easy fix. Looks like you're just missing some trailing commas.

In order to pass the build tools check, you'll need to update the BuildTools repository so that the scripts look in the correct place for the language files depending on the version, and then update composer.lock so that it uses the correct version of BuildTools for the checks.

Sources/Theme.php
DiegoAndresCortes, line 1: You didn't deal with this:

https://github.com/SimpleMachines/SMF/blob/323322e3f56df3223a75298a9ef5716968069671/Sources/Theme.php#L2601-L2602

jdarwood007:
[Antechinus](https://www.simplemachines.org/community/index.php?action=profile;u=129726) suggested we do a per directory language.  So we would do /Languages/english/index.english.php.  We could drop the language out of the filename as well.

This may also be a good time to discuss.  Should we also go locale for our naming instead of our custom method.  So index.english.php becomes index.en-us.php. @Dzonny any opinion on that?

DiegoAndresCortes:
I like the idea and support it.
Could we still support subdirectories? E.g languages/english/LargeMod/

jdarwood007:
You have an idea of what you think it would be the best source of truth regarding how we associate a locale to a language?
Use the unicode data? https://www.unicode.org/cldr/charts/42/delta/en.html
Use Crowdin data? https://developer.crowdin.com/language-codes/


You can already do sub folders with languages.  Just call it with the folder name.  SimpleDesk already do this:
```
load_language('sd_language/SimpleDeskAdmin');
```

DiegoAndresCortes:
> You can already do sub folders with languages. Just call it with the folder name. SimpleDesk already do this:
> 
> ```
> load_language('sd_language/SimpleDeskAdmin');
Yes I'm aware of that I do the same and Bugo does it too. I was referring to the folder per language format. For example, a MOD would now have this structure:
- /languages/
  - /english/SimpleDesk/
  - /spanish/SimpleDesk/
Is this correct?

> You have an idea of what you think it would be the best source of truth regarding how we associate a locale to a language?

I know some of our language translations are mixed in this regard, so I'm not sure either which one would be preferred.
The spanish_latin translation is using Crowdin locale for some reason, es-419. I personally like unicode more.

Sesquipedalian:
The CLDR is standard. Using it to designate our locales and languages will greatly simplify the interactions between our language files and other PHP functions.

jdarwood007:
@DiegoAndresCortes Correct, they would be that way.  Although with the locale discussion it would be /en-us/Simpledesk and /es-mx/SimpleDesk.

@Sesquipedalian I agree, which is why I suggested it.  We would also be able to build in support to read the users locale in their request, check it against our known language list of locales, and then change to that language.  No funky workarounds needed.  I don't think I need to target that with this PR, but moving us over to en-us and the folder structure seems to be a win all around that we can make for this major release.
I think I see you pointed out that we should follow the CLDR standards.  I didn't know if Crowdin's differed enough for us to want to use theirs, or if we are happy with what Crowdin has and we can just go forward with using what they have.

Sesquipedalian:
Yeah, I would want us to follow the CLDR rather than any vendor-specific variant.

Reorganizing all the language files into subdirectories named after CLDR locale strings is an _excellent_ plan. 

I would prefer to see the same structure enforced for theme-specific language files as well. Stricter consistency is actually easier for everyone to understand and work with in the long run.

jdarwood007:
Moving this to draft.  I still need to finish the upgrader work.  The upgrader will have to handle updating the members table with a old => new.
I also need to test editing and copying functions that are handled under the Themes admin action.

This got things working for me.

Sesquipedalian:
One thing: It would be best to use the canonical forms of locale labels in all cases. The canonical forms use underscores rather than hyphens between label parts, and the country code part is always uppercase. So, `en_US` rather than `en-us`.

An easy way to get the canonical form of any locale label is by using [Locale::Canonicalize](https://www.php.net/manual/en/locale.canonicalize).

(I also posted this comment on the related BuildTools pull request.)

jdarwood007:
I think we can make that work.  I wanted it to be able to handle the browser information as sent by the browser.  Appears to be able to work:

```
<?php
var_dump($_SERVER['HTTP_ACCEPT_LANGUAGE']);

$langs = array_map(function($l) {
	[$v,] = explode(';', $l);
	return $v;
}, explode(',', $_SERVER['HTTP_ACCEPT_LANGUAGE']));

var_dump($langs);

foreach ($langs as $locale)
	var_dump(Locale::canonicalize($locale));
```
Passing in `en-US,en;q=0.9` returns `en_US` on the first result.

jdarwood007:
I'm going to need help from @Dzonny and the rest of our translators.  To handle the upgrade logic, we need to convert our custom naming format to the proper locales.
What I have so far, which is most likely wrong:
```
	'english' => 'en_US',
	'afrikaans' => 'af_ZA',
	'albanian' => 'sq_AL',
	'arabic' => 'ar_SA',
	'bulgarian' => 'bg_BG',
	'catalan' => 'ca_ES',
	'chinese_simplified' => 'zh_CN',
	'chinese_traditional' => 'zh_TW',
	'croatian' => 'hr_HR',
	'czech_informal' => 'cs',
	'czech' => 'cs_CZ',
	'danish' => 'da_DK',
	'dutch' => 'nl_NL',
	'english_british' => 'en_GB',
	'esperanto' => 'eo_UY',
	'estonian' => 'et_EE',
	'finnish' => 'fi_FI',
	'french' => 'fr_FR',
	'galician' => 'gl_ES',
	'german_informal' => 'de',
	'german' => 'de_DE',
	'greek' => 'el_GR',
	'hebrew' => 'he_IL',
	'hungarian' => 'hu_HU',
	'indonesian' => 'id_ID',
	'italian' => 'it_IT',
	'japanese' => 'ja_JP',
	'kurdish_kurmanji' => 'kmr_TR',
	'lithuanian' => 'lt_LT',
	'macedonian' => 'mk_MK',
	'malay' => 'ms_MY',
	'norwegian' => 'no_NO',
	'persian' => 'fa_IR',
	'polish' => 'pl_PL',
	'portuguese_brazilian' => 'pt_BR',
	'portuguese_pt' => 'pt_PT',
	'romanian' => 'ro_RO',
	'russian' => 'ru_RU',
	'serbian_cyrillic' => 'sr_SP',
	'serbian_latin' => 'sr_CS',
	'slovak' => 'sk_SK',
	'slovenian' => 'sl_SI',
	'spanish_es' => 'es_ES',
	'spanish_latin' => 'es_MX',
	'swedish' => 'sv_SE',
	'thai' => 'th_TH',
	'turkish' => 'tr_TR',
	'ukrainian' => 'uk_UA',
	'urdu' => 'ur_PK',
	'vietnamese' => 'vi_VN',
	'welsh' => 'cy_GB',
	'armenian_east' => 'hy',
	'armenian_west' => 'hy',
	'basque' => 'eu_ES',
	'cambodian' => 'km_KH',
	'bangla' => 'bn',
	'bosnian' => 'bs_BA',
	'chinese-simplified' => 'zh_Hans',
	'chinese-traditional' => 'zh_Hant',
	'filipino_tagalog' => 'fil',
	'filipino_visayan' => 'fil',
	'georgian' => 'ka_GE',
	'hindi' => 'hi',
	'icelandic' => 'is',
	'korean' => 'ko_KR',
	'kurdish_sorani' => 'ckb_IR',
	'lao' => 'lo',
	'latvian' => 'lv_LV',
	'malayalam' => 'ml_IN',
	'mongolian' => 'mn',
	'sakha' => 'sah',
	'sinhala' => 'si',
	'spanish' => 'es',
	'telugu' => 'te',
	'turkmen' => 'tk',
	'uzbek_latin' => 'uz_UZ',
	'serbian' => 'sr',
	'filipino_taglog' => 'fil',
	'filipino' => 'fil',
	'lituanian' => 'lt',
	'catala' => 'ca',
	'norsk' => 'nn',
	'brazilian' => 'es_BR',
	'portuguese' => 'pt',
	'azerbaijani' => 'az_AZ',
	'belarusian' => 'be_BY',
	'english_pirate' => 'en_PT',
	'uyghur' => 'ug',
```

Going to start with the localization team and go from there.

jdarwood007:
I realize that list will have to be in the source code. To maintain compatibility with older mods, we will have to handle the old languages.  They would also install into the root language folders, instead of in the sub directories.  But we only really need to do this for non system files or mods that are not fully compatible.

@Sesquipedalian I think it would be best if I triggered something that only happened if we call from a combat situation.  Seems like we could modify the loadLanguage legacy call to pass an additional param, which would then allow the fallback.

DiegoAndresCortes (changes requested):
No testing yet from me atm, but this also needs to be removed:
https://github.com/SimpleMachines/SMF/blob/b730ca91d91ba34f7c6d2d88f9e67937c9aad5f1/Themes/default/index.template.php#L56-L57

.gitignore
```diff
@@ -1,6 +1,6 @@
 # SMF Generated Files #
 #######################
-agreement.txt
+/Languages/en-us/agreement.txt
```
Sesquipedalian:
```suggestion
/Languages/en_US/agreement.txt
```

Sources/Actions/Admin/Languages.php
```diff
@@ -872,13 +884,14 @@ function ($val1, $val2) {
 		];
 
 		$madeSave = false;
+		$index_filename = Config::$languagesdir . '/' . $lang_id . '/' . 'index.' . $lang_id . '.php';
```
Sesquipedalian:
```suggestion
		$general_filename = Config::$languagesdir . '/' . $lang_id . '/' . 'General.' . $lang_id . '.php';
```

Sources/Actions/Admin/Languages.php
```diff
@@ -872,13 +884,14 @@ function ($val1, $val2) {
...
 
 		if (!empty($_POST['save_main']) && !$current_file) {
 			User::$me->checkSession();
 			SecurityToken::validate('admin-mlang');
 
 			// Read in the current file.
-			$current_data = implode('', file(Theme::$current->settings['default_theme_dir'] . '/languages/index.' . $lang_id . '.php'));
+			$current_data = implode('', file($index_filename));
```
Sesquipedalian:
```suggestion
			$current_data = implode('', file($general_filename));
```

Sources/Actions/Admin/Languages.php
```diff
@@ -889,7 +902,7 @@ function ($val1, $val2) {
 
 			$current_data = preg_replace(array_keys($replace_array), array_values($replace_array), $current_data);
 
-			$fp = fopen(Theme::$current->settings['default_theme_dir'] . '/languages/index.' . $lang_id . '.php', 'w+');
+			$fp = fopen($index_filename, 'w+');
```
Sesquipedalian:
```suggestion
			$fp = fopen($general_filename, 'w+');
```

Sources/Actions/Admin/Languages.php
```diff
@@ -899,9 +912,9 @@ function ($val1, $val2) {
 		// Quickly load index language entries.
 		$old_txt = Lang::$txt;
 
-		require Theme::$current->settings['default_theme_dir'] . '/languages/index.' . $lang_id . '.php';
+		require $index_filename;
```
Sesquipedalian:
```suggestion
		require $general_filename;
```

Sources/Actions/Admin/Languages.php
```diff
@@ -899,9 +912,9 @@ function ($val1, $val2) {
...
 
-		Utils::$context['lang_file_not_writable_message'] = is_writable(Theme::$current->settings['default_theme_dir'] . '/languages/index.' . $lang_id . '.php') ? '' : sprintf(Lang::$txt['lang_file_not_writable'], Theme::$current->settings['default_theme_dir'] . '/languages/index.' . $lang_id . '.php');
+		Utils::$context['lang_file_not_writable_message'] = is_writable($index_filename) ? '' : sprintf(Lang::$txt['lang_file_not_writable'], $index_filename);
```
Sesquipedalian:
```suggestion
		Utils::$context['lang_file_not_writable_message'] = is_writable($general_filename) ? '' : sprintf(Lang::$txt['lang_file_not_writable'], $general_filename);
```

Sources/Actions/Admin/Registration.php
```diff
@@ -224,7 +224,7 @@ public function agreement(): void
 
 		// Try to figure out if we have more agreements.
 		foreach (Utils::$context['languages'] as $lang) {
```
Sesquipedalian:
```suggestion
		foreach (Utils::$context['languages'] as $lang_id => $lang) {
```

Sources/Actions/Admin/Registration.php
```diff
@@ -224,7 +224,7 @@ public function agreement(): void
...
-			if (file_exists(Config::$boarddir . '/agreement.' . $lang['filename'] . '.txt')) {
+			if (file_exists(Config::$langaugesdir . '/agreement.' . $lang['filename'] . '.txt')) {
```
Sesquipedalian:
```suggestion
			if (file_exists(Config::$langaugesdir . '/' . $lang_id . '/agreement.txt')) {
```

Sesquipedalian: In fact, the contents of `Utils::$context['editable_agreements']` and `Utils::$context['current_agreement']` on the lines below need to change to reflect the new directory structure. That in turn requires changes to more lines of code in the rest of this method—and possibly other places where those context variables are used.

Sources/Actions/Register.php
```diff
@@ -190,10 +190,10 @@ public function show(): void
...
 		if (!empty(Config::$modSettings['requireAgreement'])) {
 			// Have we got a localized one?
-			if (file_exists(Config::$boarddir . '/agreement.' . User::$me->language . '.txt')) {
-				Utils::$context['agreement'] = BBCodeParser::load()->parse(file_get_contents(Config::$boarddir . '/agreement.' . User::$me->language . '.txt'), true, 'agreement_' . User::$me->language);
-			} elseif (file_exists(Config::$boarddir . '/agreement.txt')) {
-				Utils::$context['agreement'] = BBCodeParser::load()->parse(file_get_contents(Config::$boarddir . '/agreement.txt'), true, 'agreement');
+			if (file_exists(Config::$languagesdir . '/agreement.' . User::$me->language . '.txt')) {
+				Utils::$context['agreement'] = BBCodeParser::load()->parse(file_get_contents(Config::$languagesdir . '/agreement.' . User::$me->language . '.txt'), true, 'agreement_' . User::$me->language);
+			} elseif (file_exists(Config::$languagesdir . '/agreement.txt')) {
+				Utils::$context['agreement'] = BBCodeParser::load()->parse(file_get_contents(Config::$languagesdir . '/agreement.txt'), true, 'agreement');
```
Sesquipedalian:
```suggestion
			if (file_exists(Config::$languagesdir . '/' . User::$me->language . '/agreement.txt')) {
				Utils::$context['agreement'] = BBCodeParser::load()->parse(file_get_contents(Config::$languagesdir .'/' . User::$me->language . '/agreement.txt'), true, 'agreement_' . User::$me->language);
			} elseif (file_exists(Config::$languagesdir . '/agreement.txt')) {
				Utils::$context['agreement'] = BBCodeParser::load()->parse(file_get_contents(Config::$languagesdir . '/agreement.txt'), true, 'agreement');
```

Sources/Lang.php
```diff
@@ -255,6 +261,11 @@ public static function load(string $template_name, string $lang = '', bool $fata
 				break;
 			}
 
+			// Keep track of what we're up to, soldier.
+			if (!empty(Config::$db_show_debug)) {
+				Utils::$context['debug']['language_files'][] = $file[2] . '.' . $file[1] . ' (' . (Config::$languagesdir == $file[0] ? 'Base' : basename(Theme::$current->settings['theme_url'] ?? 'unknown')) . ')';
```
Sesquipedalian:
```suggestion
				Utils::$context['debug']['language_files'][] = $file[2] . '/' . $file[1] . ' (' . (Config::$languagesdir == $file[0] ? 'Base' : basename(Theme::$current->settings['theme_url'] ?? 'unknown')) . ')';
```

I think that's what you meant here.

Sources/Lang.php
```diff
@@ -404,7 +407,7 @@ public static function get(bool $use_cache = true): array
 							// Set the language's name.
 							if (!empty($matchNative) && !empty($matchNative[1])) {
 								// Don't mislabel the language if the translator missed this one.
-								if ($langName !== 'English' && $matchNative[1] === 'English') {
+								if ($entry !== 'en_US' && $matchNative[1] === 'English') {
```
Sesquipedalian:
```suggestion
								if ($entry !== 'en_US' && $matchNative[1] === 'English (US)') {
```

Sources/User.php
```diff
@@ -5271,6 +5271,21 @@ protected function setLanguage(): void
 		elseif (!empty($_SESSION['language']) && isset($languages[strtr($_SESSION['language'], './\\:', '____')])) {
 			$this->language = strtr($_SESSION['language'], './\\:', '____');
 		}
+		// Can we locate it in the accept language?
+		elseif (!empty($_SERVER['HTTP_ACCEPT_LANGUAGE'])) {
```
```suggestion
		elseif (!empty($_SERVER['HTTP_ACCEPT_LANGUAGE']) && class_exists('\Locale')) {
```

The `\Locale` class is provided by the intl extension, but many hosts out there don't have the intl extension installed. Since trying to guess the right language based on `$_SERVER['HTTP_ACCEPT_LANGUAGE']` is a pleasant addition rather than a mission critical feature, I say we just skip the attempt if the class is unavailable.

other/install.php
```diff
@@ -389,21 +419,21 @@ function load_lang_file()
...
 	// Make sure it exists, if it doesn't reset it.
-	if (!isset($_SESSION['installer_temp_lang']) || preg_match('~[^\\w_\\-.]~', $_SESSION['installer_temp_lang']) === 1 || !file_exists(Config::$boarddir . '/Themes/default/languages/' . $_SESSION['installer_temp_lang'])) {
+	if (!isset($_SESSION['installer_temp_lang']) || preg_match('~[^\\w_\\-.]~', $_SESSION['installer_temp_lang']) === 1 || !file_exists(Config::$languagesdir . '/' . $_SESSION['installer_temp_lang'])) {
 		// Use the first one...
 		list($_SESSION['installer_temp_lang']) = array_keys($incontext['detected_languages']);
 
 		// If we have english and some other language, use the other language.  We Americans hate english :P.
-		if ($_SESSION['installer_temp_lang'] == 'Install.english.php' && count($incontext['detected_languages']) > 1) {
-			list(, $_SESSION['installer_temp_lang']) = array_keys($incontext['detected_languages']);
+		if ($_SESSION['installer_temp_lang'] == 'Install.en_US.php' && count($incontext['detected_languages']) > 1) {
```
Sesquipedalian: This certainly appears to be wrong, but I'm not sure whether `en_US/Install.php` will work here either. We might need to do more work in this function.

other/install.php
```diff
@@ -564,14 +594,14 @@ function CheckFilesWritable()
 		'Packages',
 		'Smileys',
 		'Themes',
-		'agreement.txt',
+		'Languages/agreement.txt',
```
Sesquipedalian:
```suggestion
		'Languages/en_US/agreement.txt',
```

other/update_version_numbers.php
```diff
@@ -147,7 +147,7 @@ function ($filename) {
 }
 
 // Update SMF_LANG_VERSION
-preg_match($lang_pattern, file_get_contents("{$basedir}/Themes/default/languages/index.english.php"), $matches);
+preg_match($lang_pattern, file_get_contents("{$basedir}/Languages/index.english.php"), $matches);
```
Sesquipedalian:
```suggestion
preg_match($lang_pattern, file_get_contents("{$basedir}/Languages/en_US/General.php"), $matches);
```

other/upgrade.php
```diff
@@ -536,21 +537,17 @@ function load_lang_file()
...
-	} elseif (isset(Config::$language)) {
-		$_SESSION['upgrader_langfile'] = 'Install.' . Config::$language . '.php';
+	if (isset($upcontext['language']) && file_exists($lang_dir . '/' . $upcontext['language'] . '/Install.' . $upcontext['language'] . '.php')) {
+		$_SESSION['upgrader_langfile'] = $upcontext['language'] . '/Install.' . $upcontext['language'] . '.php';
+	} elseif (isset($upcontext['lang']) && file_exists($lang_dir . '/' . $upcontext['lang'] . '/Install.' . $upcontext['lang'] . '.php')) {
+		$_SESSION['upgrader_langfile'] = $upcontext['lang'] . '/Install.' . $upcontext['lang'] . '.php';
+	} elseif (isset(Config::$language) && file_exists($lang_dir . '/' . Config::$language . '/Install.' . Config::$language . '.php')) {
+		$_SESSION['upgrader_langfile'] = Config::$language . '/Install.' . Config::$language . '.php';
+	} else {
+		$_SESSION['upgrader_langfile'] = Config::$language . '/Install.en_US.php';
```
Sesquipedalian:
```suggestion
	if (isset($upcontext['language']) && file_exists($lang_dir . '/' . $upcontext['language'] . '/Install.php')) {
		$_SESSION['upgrader_langfile'] = $upcontext['language'] . '/Install.php';
	} elseif (isset($upcontext['lang']) && file_exists($lang_dir . '/' . $upcontext['lang'] . '/Install.php')) {
		$_SESSION['upgrader_langfile'] = $upcontext['lang'] . '/Install.php';
	} elseif (isset(Config::$language) && file_exists($lang_dir . '/' . Config::$language . '/Install.php')) {
		$_SESSION['upgrader_langfile'] = Config::$language . '/Install.php';
	} else {
		$_SESSION['upgrader_langfile'] = 'en_US/Install.php';
```

other/upgrade.php
```diff
@@ -566,19 +563,43 @@ function load_lang_file()
...
-				// Skip any old '-utf8' language files that might be lying around
-				if (strpos($entry, '-utf8') !== false) {
+				if (!is_dir(Config::$languagesdir . '/' . $entry) || !file_exists(Config::$languagesdir . '/' . $entry . '/' . 'Install.php') || !file_exists(Config::$languagesdir . '/' . $entry . '/' . 'General.php')) {
 					continue;
 				}
 
-				if (substr($entry, 0, 8) == 'Install.' && substr($entry, -4) == '.php') {
-					$detected_languages[$entry] = ucfirst(substr($entry, 8, strlen($entry) - 12));
+				// Get the line we need.
+				$fp = @fopen($language_dir . '/' . $entry . '/' . 'index.php', 'r');
```
Sesquipedalian:
```suggestion
				$fp = @fopen($language_dir . '/' . $entry . '/' . 'General.php', 'r');
```

other/upgrade.php
```diff
@@ -566,19 +563,43 @@ function load_lang_file()
...
+						if (strpos($line, '$txt[\'native_name\']') === false) {
+							continue;
+						}
+
+						preg_match('~\$txt\[\'native_name\'\]\s*=\s*\'([^\']+)\';~', $line, $matchNative);
+
+						// Set the language's name.
+						if (!empty($matchNative) && !empty($matchNative[1])) {
+							// Don't mislabel the language if the translator missed this one.
+							if ($entry !== 'en_US' && $matchNative[1] === 'English') {
```
Sesquipedalian:
```suggestion
							if ($entry !== 'en_US' && $matchNative[1] === 'English (US)') {
```

other/upgrade.php
```diff
@@ -566,19 +563,43 @@ function load_lang_file()
...
 				}
+
+				$detected_languages[$entry] = $langName ?? $entry;
 			}
 			$dir->close();
 		}
-		// Our guess was wrong, but that's fine. We'll try again after Config::$modSettings['theme_dir'] is defined.
-		elseif (!isset(Config::$modSettings['theme_dir'])) {
+		// Our guess was wrong, but that's fine. We'll try again after Config::$Languagesdir is defined.
+		elseif (!isset(Config::$Languagesdir)) {
```
Sesquipedalian:
```suggestion
		// Our guess was wrong, but that's fine. We'll try again after Config::$languagesdir is defined.
		elseif (!isset(Config::$languagesdir)) {
```

other/upgrade.php
```diff
@@ -665,16 +686,16 @@ function load_lang_file()
 		list($_SESSION['upgrader_langfile']) = array_keys($detected_languages);
 
 		// If we have English and some other language, use the other language.
-		if ($_SESSION['upgrader_langfile'] == 'Install.english.php' && count($detected_languages) > 1) {
-			list(, $_SESSION['upgrader_langfile']) = array_keys($detected_languages);
+		if ($_SESSION['upgrader_langfile'] == 'Install.en_US.php' && count($detected_languages) > 1) {
```
Sesquipedalian:
```suggestion
		if ($_SESSION['upgrader_langfile'] == 'en_US/Install.php' && count($detected_languages) > 1) {
```

Again, I'm not sure if this suggestion will work.

other/upgrade.php
```diff
@@ -1008,12 +1029,11 @@ function WelcomeLogin()
 
 	quickFileWritable($cachedir_temp . '/db_last_error.php');
 
-	if (!file_exists(Config::$modSettings['theme_dir'] . '/languages/index.' . $upcontext['language'] . '.php')) {
+	if (!file_exists(Config::$languagesdir . '/' . $upcontext['language'] . '/index.' . $upcontext['language'] . '.php')) {
```
Sesquipedalian:
```suggestion
	if (!file_exists(Config::$languagesdir . '/' . $upcontext['language'] . '/General.php')) {
```

other/upgrade.php
```diff
@@ -1008,12 +1029,11 @@ function WelcomeLogin()
...
 		return throw_error(sprintf(Lang::$txt['error_lang_index_missing'], $upcontext['language'], $upgradeurl));
-	}
+	} elseif (!isset($_GET['skiplang'])) {
+		$temp = substr(@implode('', @file(Config::$languagesdir . '/' . $upcontext['language'] . '/index.' . $upcontext['language'] . '.php')), 0, 4096);
```
Sesquipedalian:
```suggestion
		$temp = substr(@implode('', @file(Config::$languagesdir . '/' . $upcontext['language'] . '/General.php')), 0, 4096);
```

other/upgrade.php
```diff
@@ -1026,7 +1046,7 @@ function WelcomeLogin()
 	}
 
 	// Check agreement.txt. (it may not exist, in which case $boarddir must be writable.)
-	if (isset(Config::$modSettings['agreement']) && (!is_writable(Config::$boarddir) || file_exists(Config::$boarddir . '/agreement.txt')) && !is_writable(Config::$boarddir . '/agreement.txt')) {
+	if (isset(Config::$modSettings['agreement']) && (!is_writable(Config::$languagesdir) || file_exists(Config::$languagesdir . '/agreement.txt')) && !is_writable(Config::$languagesdir . '/agreement.txt')) {
```
Sesquipedalian:
```suggestion
	if (isset(Config::$modSettings['agreement']) && (!is_writable(Config::$languagesdir . '/en_US') || file_exists(Config::$languagesdir . 'en_US/agreement.txt')) && !is_writable(Config::$languagesdir . 'en_US/agreement.txt')) {
```

other/upgrade.php
```diff
@@ -1347,14 +1367,14 @@ function checkLogin()
 			$upcontext['upgrade_status']['pass'] = $upcontext['user']['pass'];
 
 			// Set the language to that of the user?
-			if (isset($user_language) && $user_language != $upcontext['language'] && file_exists(Config::$modSettings['theme_dir'] . '/languages/index.' . basename($user_language, '.lng') . '.php')) {
+			if (isset($user_language) && $user_language != $upcontext['language'] && file_exists(Config::$languagesdir . '/index.' . basename($user_language, '.lng') . '.php')) {
```
Sesquipedalian:
```suggestion
			if (isset($user_language) && $user_language != $upcontext['language'] && file_exists(Config::$languagesdir . '/' . basename($user_language, '.lng') . '/General.php')) {
```

other/upgrade.php
```diff
@@ -1347,14 +1367,14 @@ function checkLogin()
...
 				$user_language = basename($user_language, '.lng');
-				$temp = substr(@implode('', @file(Config::$modSettings['theme_dir'] . '/languages/index.' . $user_language . '.php')), 0, 4096);
+				$temp = substr(@implode('', @file(Config::$languagesdir . '/index.' . $user_language . '.php')), 0, 4096);
```
Sesquipedalian:
```suggestion
				$temp = substr(@implode('', @file(Config::$languagesdir . '/' . $user_language . '/General.php')), 0, 4096);
```

other/upgrade.php
```diff
@@ -1347,14 +1367,14 @@ function checkLogin()
...
 				preg_match('~(?://|/\*)\s*Version:\s+(.+?);\s*index(?:[\s]{2}|\*/)~i', $temp, $match);
 
 				if (empty($match[1]) || $match[1] != SMF_LANG_VERSION) {
 					$upcontext['upgrade_options_warning'] = sprintf(Lang::$txt['warning_lang_old'], $user_language, $upcontext['language']);
-				} elseif (!file_exists(Config::$modSettings['theme_dir'] . '/languages/Install.' . $user_language . '.php')) {
+				} elseif (!file_exists(Config::$languagesdir . '/Install.' . $user_language . '.php')) {
```
Sesquipedalian:
```suggestion
				} elseif (!file_exists(Config::$languagesdir . '/' . $user_language . '/Install.php')) {
```

other/upgrade.php
```diff
@@ -1506,7 +1526,7 @@ function UpgradeOptions()
 	}
 
 	// If we're overriding the language follow it through.
-	if (isset($upcontext['lang']) && file_exists(Config::$modSettings['theme_dir'] . '/languages/index.' . $upcontext['lang'] . '.php')) {
+	if (isset($upcontext['lang']) && file_exists(Config::$languagesdir . '/index.' . $upcontext['lang'] . '.php')) {
```
Sesquipedalian:
```suggestion
	if (isset($upcontext['lang']) && file_exists(Config::$languagesdir . '/' . $upcontext['lang'] . '/index.php')) {
```

other/upgrade.php
```diff
@@ -1578,6 +1598,10 @@ function UpgradeOptions()
 		$changes['packagesdir'] = fixRelativePath(Config::$boarddir) . '/Packages';
 	}
 
+	// Languages have moved!
+	if (empty(Config::$languagesdir))
+		$changes['languagesdir'] = fixRelativePath(Config::$boarddir) . '/Languages';
+
```
Sesquipedalian: PHP-CS-Fixer needed

other/upgrade.php
```diff
@@ -2957,7 +2981,7 @@ function cmdStep0()
 		print_error('Error: Unable to obtain write access to "' . basename(SMF_SETTINGS_BACKUP_FILE) . '".');
 	}
 
-	if (isset(Config::$modSettings['agreement']) && (!is_writable(Config::$boarddir) || file_exists(Config::$boarddir . '/agreement.txt')) && !is_writable(Config::$boarddir . '/agreement.txt')) {
+	if (isset(Config::$modSettings['agreement']) && (!is_writable(Config::$languagesdir) || file_exists(Config::$languagesdir . '/agreement.txt')) && !is_writable(Config::$languagesdir . '/agreement.txt')) {
```
Sesquipedalian:
```suggestion
	if (isset(Config::$modSettings['agreement']) && (!is_writable(Config::$languagesdir . '/en_US') || file_exists(Config::$languagesdir . '/en_US/agreement.txt')) && !is_writable(Config::$languagesdir . '/en_US/agreement.txt')) {
```

other/upgrade.php
```diff
@@ -2993,22 +3017,22 @@ function cmdStep0()
 		print_error('Error: Unable to obtain write access to "db_last_error.php".');
 	}
 
-	if (!file_exists(Config::$modSettings['theme_dir'] . '/languages/index.' . $upcontext['language'] . '.php')) {
+	if (!file_exists(Config::$languagesdir . '/index.' . $upcontext['language'] . '.php')) {
```
Sesquipedalian:
```suggestion
	if (!file_exists(Config::$languagesdir . '/' . $upcontext['language'] . '/General.php')) {
```

other/upgrade.php
```diff
@@ -2993,22 +3017,22 @@ function cmdStep0()
...
 		print_error('Error: Unable to find language files!', true);
 	} else {
-		$temp = substr(@implode('', @file(Config::$modSettings['theme_dir'] . '/languages/index.' . $upcontext['language'] . '.php')), 0, 4096);
+		$temp = substr(@implode('', @file(Config::$languagesdir . '/index.' . $upcontext['language'] . '.php')), 0, 4096);
```
Sesquipedalian:
```suggestion
		$temp = substr(@implode('', @file(Config::$languagesdir . '/' . $upcontext['language'] . '/General.php')), 0, 4096);
```

other/upgrade.php
```diff
@@ -2993,22 +3017,22 @@ function cmdStep0()
...
 		preg_match('~(?://|/\*)\s*Version:\s+(.+?);\s*index(?:[\s]{2}|\*/)~i', $temp, $match);
 
 		if (empty($match[1]) || $match[1] != SMF_LANG_VERSION) {
 			print_error('Error: Language files out of date.', true);
 		}
 
-		if (!file_exists(Config::$modSettings['theme_dir'] . '/languages/Install.' . $upcontext['language'] . '.php')) {
+		if (!file_exists(Config::$languagesdir . '/Install.' . $upcontext['language'] . '.php')) {
```
Sesquipedalian:
```suggestion
		if (!file_exists(Config::$languagesdir . '/' . $upcontext['language'] . '/Install.php')) {
```

other/upgrade.php
```diff
@@ -2993,22 +3017,22 @@ function cmdStep0()
...
 			print_error('Error: Install language is missing for selected language.', true);
 		}
 
 		// Otherwise include it!
-		require_once Config::$modSettings['theme_dir'] . '/languages/Install.' . $upcontext['language'] . '.php';
+		require_once Config::$languagesdir . '/Install.' . $upcontext['language'] . '.php';
```
Sesquipedalian:
```suggestion
		require_once Config::$languagesdir . '/' . $upcontext['language'] . '/Install.php';
```

Sesquipedalian (dismissed):

Sources/User.php
jdarwood007: Makes sense.  It was a nice addition since we can now properly sniff the browser data to change the language.

other/upgrade.php
jdarwood007: I got an idea that should work for both install/upgrade. It is, we only need the language.  No reason to specify install.php anywhere.  I need to make all the changes and test this.

jdarwood007: @Sesquipedalian I sent up some changes that should still work.  Note that the upgrade.php logic isn't complete.  I have to change over the UTF-8 stuff somehow.  I say somehow because there isn't a one-to-one translation on those.

jdarwood007:
So just to recap four pending issues need to be finished here.

1. Upgrade's convert UTF-8 logic needs to handle the new locales
2. I have not tested installing a mod that is using `$themesdir/default/languages`
3. Sources/PackageManager/SubsPackage.php still calls `$languagedir` (and `$languages_dir`) with the old paths.  Issue is how we handle the logic for migration to the 'new' paths.
4. I have not tested installing a mod that uses `$languagedir`

jdarwood007:
@Sesquipedalian I feel this is ready now.

Sesquipedalian:
Cool. I'll review it in the next day or two.

Sesquipedalian:
I found a few issues. It was easier to fix them than it was to document them all in review comments. How do my changes look to you, @jdarwood007?

jdarwood007:
Looks good.  I thought I had tested the language editor, but I must have broken it since I last tested it.  Do we need another set of eyes on this since you made changes?

jdarwood007:
Also note, that I haven't reviewed the changes we need to make to Crowdin.  We haven't even discussed that yet on how we handle this.  I will get a discussion going on the team boards as we also need to get the branch going in Crowdin.

Sesquipedalian:
> Looks good. I thought I had tested the language editor, but I must have broken it since I last tested it. Do we need another set of eyes on this since you made changes?

If you can just take a moment to make sure that file select box populates correctly on your machine, that should be enough.

> Also note, that I haven't reviewed the changes we need to make to Crowdin. We haven't even discussed that yet on how we handle this. I will get a discussion going on the team boards as we also need to get the branch going in Crowdin.

A discussion will be good. I had no involvement with setting up the CrowdIn integration, so I have no more idea than the next guy about what needs to be done there. As I recall, @sbulen was the primary developer involved with that at the time, so maybe start the discussion in the Development Contributors board so that he can participate in the discussion.

jdarwood007:
Sent up a small fix.  The charset and locale didn't show correctly in the list

Screen Shot 2024-01-29 at 4 22 42 PM

Because `$txt` is local, not from `Lang::$txt`

Sesquipedalian:
Nice catch. I think we can go ahead a merge this now.

PR #7863, opened 2023-11-10
Call functions instead of evaluating strings

live627:
- do not evaluate strings to execute code
- ends usage of sSelf
- autosuggest callbacks are slightly reworked
- removing some old compatibility code

### DIFF
--- a/Themes/default/scripts/suggest.js
+++ b/Themes/default/scripts/suggest.js
@@ -14,35 +14,45 @@ function smc_AutoSuggest(oOptions)
 	this.aCache = [];
 	this.aDisplayData = [];
 
-	this.sRetrieveURL = 'sRetrieveURL' in this.opt ? this.opt.sRetrieveURL : '%scripturl%action=suggest;suggest_type=%suggest_type%;search=%search%;%sessionVar%=%sessionID%;xml;time=%time%';
+	this.sRetrieveURL = this.opt.sRetrieveURL || '%scripturl%action=suggest;suggest_type=%suggest_type%;search=%search%;%sessionVar%=%sessionID%;xml;time=%time%';
+	this.oRetrieveTokens = {
+		scripturl: smf_prepareScriptUrl(smf_scripturl),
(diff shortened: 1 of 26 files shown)


live627:
The autosuggest plugin seems to support comma-separated values,  where it can send a search to the server based on the last value.

Screenshot 2023-11-09 at 12-58-32 Modify Boards

Not sure if this hidden feature should be removed.

jdarwood007:
This may be outside the scope of what you intend to to here, but how much along does this help us to being able to use defer on our javascript files?  It couldn't happen for 2.1, but I would love it if we could add defer, aligning with modern suggestions for javascript to be either async or defer.  We would async the main script, defer the execution until the body is essentially loaded and then wait for dom to report its ready.

Themes/default/scripts/suggest.js
```diff
@@ -338,24 +282,40 @@ smc_AutoSuggest.prototype.removeLastSearchString = function ()
...
 	// If there's a callback then call it.
-	if ('oCallback' in this && 'onBeforeAddItem' in this.oCallback && typeof(this.oCallback.onBeforeAddItem) == 'string')
+	if (typeof this.oCallback.onBeforeAddItem == 'function')
 	{
 		// If it returns false the item must not be added.
-		if (!eval(this.oCallback.onBeforeAddItem + '(' + this.opt.sSelf + ', \'' + sItemId + '\');'))
+		if (!this.oCallback.onBeforeAddItem.call(this, sItemId))
+			return;
+	}
+	// Backward compatibility; to be removed in the future
```
live627, line 294: This line is too vague. It needs to specify a milestone.

```suggestion
	// Backward compatibility; to be removed in v4.0.0.
```

Opinions? I hereby suggest the next major version.

live627:
Beyond the scope of this branch. I am initially sending in my backlog of JavaScript changes. Script deferral https://github.com/SimpleMachines/SMF/issues/6539 comes in the future because the current setup isn't ready.

PR #7867, opened 2023-11-11
Remove MagickWand support

Oldiesmann:
As far as I can tell, the MagickWand extension for PHP is no longer developed. There is no reference to it on the ImageMagick website (except on their now-archived forum) and the only source code I can find for it (on GitHub) hasn't been updated since it was uploaded over a decade ago. Given that, there's no point in continuing to support it.

Note that this also makes a few language file changes, so I'm not sure if that needs to be changed on CrowdIn or not.


Sesquipedalian (approved):

Sesquipedalian:
I've changed the milestone for this to 3.0 Alpha 1. I remember now that I already removed MagicWand support in other places where it was previously used, which means that having any remnants of it around is actually a possible source of bugs.

@Oldiesmann, please resolve the conflicts so that this can be merged.

Oldiesmann:
Should be good to go now

PR #7869, opened 2023-11-11
[3.0] Update our password conversion

### LINKED ISSUE #7860: mhash deprecated in php 8.1
#### Description

We only keep it around for legacy logon checks...  But even that may need to go.

https://www.simplemachines.org/community/index.php?topic=587369.0

jdarwood007:
Fixes #7860

I reorganized the passwords to only check if the strlen matches first using a switch statement.  No sense in trying to check passwords that won't produce a hash of the correct length.
I moved SMF's passwords from 1.1 to 2.0 outside of the checks better, since they are still supported as part of a native upgrade without conversion.  SMF 1.0/YabbSE requires password conversions.

I thought about looking around at what password mechanisms are being used by forums nowadays, but I hope they have all switched over to the native password_* functions.

### DIFF
--- a/Sources/Actions/Login2.php
+++ b/Sources/Actions/Login2.php
@@ -522,8 +522,7 @@ class Login2 implements ActionInterface
 		$other_passwords = [];
 
 		// SMF 1.1 and 2.0 password styles.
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Delightfully cleaner and clearer!

PR #7910, opened 2023-11-26
Fixes problem getting the number of search results

### LINKED ISSUE #7877: Searching causes errors
#### Description
I created a large custom search index on my test board (which currently has all of 3 topics and posts), then tried searching, which in turn led to this error:
image

I also got a completely different error while searching for something else:
image

### Steps to reproduce
1. Create a custom search index
2. Try searching for something

- Version/Git revision: 3.1 Alpha (current as of sometime yesterday afternoon)
- Database Type: MariaDB
- Database Version: 11.1.2
- PHP Version: 8.2.11

I have a topic with the subject of "Poll" (created while debugging the poll issue yesterday), but no posts have that text
Welcome should return the "Welcome to SMF" post. This also generates an error about an undefined array index.

Sesquipedalian:
Fixes #7877

Since a generator doesn't start executing its code until the first time it is asked to do so, we had to call it at least once in order to cause SearchResult::$messages_request to be set. The easiest way to do that was just to call `self::$getter->current();` in order to kick off the query.

### DIFF
--- a/Sources/Actions/Search2.php
+++ b/Sources/Actions/Search2.php
@@ -251,9 +251,7 @@ class Search2 implements ActionInterface
 		// Number of pages hard maximum - normally not set at all.
 		Config::$modSettings['search_max_results'] = empty(Config::$modSettings['search_max_results']) ? 200 * Config::$modSettings['search_results_per_page'] : (int) Config::$modSettings['search_max_results'];
 
-		if (isset($_REQUEST['start'])) {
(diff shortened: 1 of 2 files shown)


jdarwood007 (approved):
That is what I was missing.  Calling the current method to initiate the generator.

Works for me with no issue now.  @Oldiesmann Can you also confirm?

Oldiesmann:
No errors on my end either and it returned the expected result.

PR #7911, opened 2023-11-26
Fixes issue where SMF\Pm tried to use $this->current_label_redirect

### LINKED ISSUE #7907: Error when sending PMs
#### Description
When sending a PM to a user, I got a blank screen. The domain's error log shows ""PHP Fatal error: Uncaught Error: Using $this when not in object context" related to this line:
https://github.com/SimpleMachines/SMF/blob/88ef4a09f9edad42dd13ec22653a07705df12769/Sources/PersonalMessage/PM.php#L1119

There is no "current_label_redirect" property for the PM class. There is one defined in the PersonalMessage class, but PM does not extend PersonalMessage.

- Version/Git revision: Latest as of 2:48pm on November 25th
- Database Type: MariaDB
- Database Version: 11
- PHP Version: 8.2.13

$this->current_label_redirect is also used a few lines after this.

Sesquipedalian:
Fixes #7907

### DIFF
--- a/Sources/Actions/PersonalMessage.php
+++ b/Sources/Actions/PersonalMessage.php
@@ -381,7 +381,10 @@ class PersonalMessage implements ActionInterface
 	 */
 	public function send2(): void
 	{
-		PM::compose2();
+		// Message sent successfully?
+		if (PM::compose2()) {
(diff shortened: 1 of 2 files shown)


Sesquipedalian:
Please test, @Oldiesmann

PR #7922, opened 2023-11-28
add script shortcut to `php-cs-fixer`

live627:
the new command is `composer lint`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,11 @@
     "minimum-stability": "dev",
     "prefer-stable": true,
     "require-dev": {
-        "simplemachines/build-tools": "dev-release-3.0"
+        "simplemachines/build-tools": "dev-release-3.0",
+        "friendsofphp/php-cs-fixer": "^3.40"
+    },
+    "scripts": {
+        "lint": "php-cs-fixer fix -v"
     },
     "config": {
         "platform": {
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         "friendsofphp/php-cs-fixer": "^3.40"
     },
     "scripts": {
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -9,11 +9,15 @@
     "minimum-stability": "dev",
     "prefer-stable": true,
     "require-dev": {
-        "simplemachines/build-tools": "dev-release-3.0"
+        "simplemachines/build-tools": "dev-release-3.0",
+        "friendsofphp/php-cs-fixer": "^3.40"
+    },
+    "scripts": {
+        "lint": "php-cs-fixer fix -v"
```
Sesquipedalian:
```suggestion
        "lint": "php-cs-fixer check -v"
```

I don't think these workflows are the place to make changes to files. Unless I misunderstand something fundamental about how these workflows function, all we can do here is check the files.

Sesquipedalian (changes requested):

jdarwood007:
https://github.com/OskarStark/php-cs-fixer-ga

Add this while we are at it.  Let Actions handle it.  Set it up to only do it on changed files.

composer.json
Sesquipedalian: That is, unless these workflows can automatically add an extra commit with the changes. If that's the case, that would be awesome.

jdarwood007: I think this is just so you can run 'composer lint' in your command line and it will run the php-cs-fixer.  But I would suggest it run it only against your changed files.  Existing files not compliant should be ignored.

live627:
all requested changes have been applied

Sesquipedalian (approved):

PR #7923, opened 2023-11-28
Defined types and strict typing

### LINKED ISSUE #7916: Type hinting needed in DatabaseApi classes
Neither SMF\Db\DatabaseApiInterface, nor any of the classes that implement it, use typed parameters and return values. They really should. Implementing them will probably expose a number of obscured bugs, but that's a good thing.

jdarwood007:
Start adding types and enforcing strict typing
A few bugs I found while doing this

@Sesquipedalian How is the progress on this look.  I think this is what we want to do.  I picked caching and Db to start with as the heaviest ones we need.  But more progress needs made on others.  I was going to work with lower level stuff before working into actions.
Interestingly enough, for the most part this hasn't shown any issues.  I'm testing VScode with PHP intelisense and it has found more issues than this.

@albertlast Can you give me some more pg testing?  I did basic testing and pg seems to function normally.  I haven't jumped into operations with things like package manager, but I feel more confident about those working right than the basics.

Fixes #7916


Sesquipedalian:
Adding type hints is great. _Almost_ everywhere that doesn't currently have type hinting is due to human forgetfulness on my part. But just FYI, there are a (very) few places throughout the codebase where type hints were intentionally left out. Unfortunately, I don't remember now exactly which places those were. 🙁 But basically that just means that as you work on this, don't just do batch changes assuming it's fine; check each case to make sure it makes sense.

`declare(strict_types = 1);` is another matter. In theory, it's a great idea, and strongly recommended for new code. But SMF has a *lot* of code that relies on type juggling for parameters. I'm not categorically opposed to using strict typing.  Not at all. But because strict typing can so easily break existing code (sometimes in unexpected ways), I will insist that every single change to use strict typing must be accompanied by thorough testing data to demonstrate that the change doesn't break anything.

jdarwood007:
I think we can still get through strict typing if we do it well.  I picked some of the most core aspects of our code to do it on to start with.  Database stuff is handled everywhere by everything.  So it will get a lot of testing.  In theory if we get our core stuff, as we bubble up to the upper level stuff, they should be easier to test and ensure we are using the proper typing.

I was next going to go into PackageManager, Personal Message, Search, Tasks, TOTP and WebFetch.  I don't have a good way to test Subscriptions right now, but it is perhaps the harder of all of them to test.

If we get this in and merged, it helps testing this while we continue to dev 3.0.

jdarwood007:
FYI any place type hinting is left out, we should at least be able to use the type 'mixed', I already used it in a few places.  There is only one case where you can't use mixed and that is if you don't specify any return.  That is considered for void, which can't be mixed with any other return type.  PHP accepts that loosely as a return type of null, so the code can be updated to return null in that case.

jdarwood007:
@Sesquipedalian Going to stop here with this.  More needs converted, but I think this is a good start for us to test and fix anything that has a bad type juggling.  I fixed a few things found during my testing already.

tyrsson:
There is a lot of places, especially in actions where we end up either returning an array or the method terminates without returning anything, effectively making it a void return type. We can not union void and array so we need to make a decision on these. Technically speaking a method that returns void, still returns a null value.

jdarwood007:
Correct.  So the method I went was to make it null.  Only a little code would be affected by this sort of change anyways.  If it must return a non null, we should return the default values: '', 0, [], etc.

Speaking from a architecture standpoint, Actions themselves that return to the UI should always return void.  If they should return a value, a working/helper functions should be handling that work.  There is no issue with the working/helper functions returning a type.

I also used the mixed on types that does just to much to get a specific type.  Someday we will want to move beyond that to ensure we return a more defined type.

tyrsson:
I would really like to see this merged before I move on since I am about to start Sources/Cache in the backcompat work. It would simplify my life :)

I would also say that as long as the typing is wide enough to allow the required types that SMF needs to juggle that is where we would want to go. All other types should not be allowed. 

I would rather not throw mixed at a lot of things just to have a type. I would rather take the time to verify the required types and type it correctly to those types and no others. It's the whole point of typing the code.

Also, I would say we are at greater risk implementing typing without strict types than we are by using strict types as at least we will know exactly where and why the code breaks and then it can be fixed. Using typing without strict types usually leads to more bugs than strict types causes in problems. Not to mention the issues caused by strict typing are usually easily fixed.

jdarwood007:
A few places that use mixed are because the usage is to wide.  One such place is caching.  Typically we do expect an array return, but because its possible to cache anything, who know what it is.

I think at least starting with what we have in this and merging, means we can test.  If any issues arrise we have two situations:
1. Either actions are not handling the logic properly and need fixed.  Logic is correct (sometimes a simple cast is all).
2. The core code is not accepting a type it should be, phpdoc and its type updated to resolve this.

By doing type casting, we will cause fatal php errors when they occur, but this is alpha 1 and we are not at a release point yet.  I'm confident we can find this by the time we get near releasing.

I also don't know what @Tyrsson has done for the stuff in /Sources/* directly, as those need typing as well.  I hit mostly the sub stuff here, fixing anything else outside of it I saw needed resolution because of my changes.  I was going to work on those in another PR to give this one time to settle as I worked in it.  I'm fairly confident 99% of the use cases are safe though, we will only find very odd things I suspect with this PR.  Things that static analysis can't see.

tyrsson:
I changed very little (if any) in the actual Sources/* files I think. Any change I made was according to the phpdoc typing for the method, or if there was a missing void return. Possibly a few instances changed to ?array per our earlier discussion above. I typed the back compat functions to match the signatures they are proxying too.

I was actually looking at Caching while you were typing your last reply. In instances like that mixed is the only thing you can return... I mean it could literally be nearly anything.

jdarwood007:
Yea, only way would be to add more methods that would do a getAsInt('my_cache_key') which would just wrap a `return (int) self::get($key);`  One thing I do miss from C# stuff is that I can define a type for my return in my call.

Sources/PackageManager/SubsPackage.php
```diff
@@ -342,7 +342,7 @@ public static function read_tgz_data($data, $destination, $single_file = false,
 	 * @param array $files_to_extract
 	 * @return mixed If destination is null, return a short array of a few file details optionally delimited by $files_to_extract. If $single_file is true, return contents of a file as a string; false otherwise
 	 */
-	public static function read_zip_data($data, $destination, $single_file = false, $overwrite = false, $files_to_extract = null)
+	public static function read_zip_data(string $data, string $destination, bool $single_file = false, bool $overwrite = false, array$files_to_extract = null): mixed
```
tyrsson: There is a typo in this signature. Need a space between the array type and $files_to_extract

tyrsson (approved):
I support the type changes. The rest of the changes  I would have to deep dive into the code so my approval is in respect to the typing. We have approved mods with better typing than we are providing. It will only help us in the long run.

PR #7927, opened 2023-12-02
Various Fixes

jdarwood007:
Just some fixes found by intelisense.  I worked through some stuff.  Didn't get into actions.

As well, I saw we have magic quotes stuff.  It looks like get_magic_quotes_gpc and various other stuff was removed in 8.0.  We can safely remove this now.

### DIFF
--- a/Sources/Attachment.php
+++ b/Sources/Attachment.php
@@ -1469,7 +1469,7 @@ class Attachment implements \ArrayAccess
 			);
 		}
 
-		if (empty(Config::$modSettings['attachmentThumbnails']) || (empty($attachmentOptions['width']) && empty($attachmentOptions['height']))) {
+		if (empty($image) || empty(Config::$modSettings['attachmentThumbnails']) || (empty($attachmentOptions['width']) && empty($attachmentOptions['height']))) {
 			return true;
 		}
 
--- a/Sources/Attachment.php
+++ b/Sources/Attachment.php
@@ -717,7 +717,7 @@ class Attachment implements \ArrayAccess
 			}
 		}
 
-		if (!$doit) {
+		if (empty($doit)) {
 			return;
 		}
 
@@ -939,7 +939,7 @@ class Attachment implements \ArrayAccess
 	{
 		// Make sure we're uploading to the right place.
 		if (!empty(Config::$modSettings['automanage_attachments'])) {
-			automanage_attachments_check_directory();
+			self::automanageCheckDirectory();
 		}
 
 		if (!is_array(Config::$modSettings['attachmentUploadDir'])) {
@@ -1469,7 +1469,7 @@ class Attachment implements \ArrayAccess
 			);
 		}
 
-		if (!$this->is_image || empty(Config::$modSettings['attachmentThumbnails']) || (empty($attachmentOptions['width']) && empty($attachmentOptions['height']))) {
+		if (empty(Config::$modSettings['attachmentThumbnails']) || (empty($attachmentOptions['width']) && empty($attachmentOptions['height']))) {
 			return true;
 		}
 
@@ -2228,8 +2228,8 @@ class Attachment implements \ArrayAccess
 				// This can happen if an uploaded SVG is missing some key data.
 				foreach (['real_width', 'real_height'] as $key) {
 					if (!isset($attachmentData[$i][$key]) || $attachmentData[$i][$key] === INF) {
-						loadLanguage('Admin');
-						$attachmentData[$i][$key] = ' (' . $txt['unknown'] . ') ';
+						Lang::load('Admin');
+						$attachmentData[$i][$key] = ' (' . Lang::$txt['unknown'] . ') ';
 					}
 				}
 			}
--- a/Sources/Attachment.php
+++ b/Sources/Attachment.php
@@ -1316,6 +1316,7 @@ class Attachment implements \ArrayAccess
 	{
 		// If this is an image we need to set a few additional parameters.
 		$image = new Image($attachmentOptions['tmp_name']);
+		$is_image = !empty($image);
 
 		// Source is empty for non-images.
 		if (!empty($image->source)) {
@@ -1469,7 +1470,7 @@ class Attachment implements \ArrayAccess
 			);
 		}
 
-		if (empty($image) || empty(Config::$modSettings['attachmentThumbnails']) || (empty($attachmentOptions['width']) && empty($attachmentOptions['height']))) {
+		if (empty($is_image) || empty(Config::$modSettings['attachmentThumbnails']) || (empty($attachmentOptions['width']) && empty($attachmentOptions['height']))) {
 			return true;
 		}
 
--- a/Sources/BBCodeParser.php
+++ b/Sources/BBCodeParser.php
@@ -1143,7 +1143,7 @@ class BBCodeParser
 	 *
 	 * @param string|bool $message The string to parse.
 	 * @param bool $smileys Whether to parse smileys. Default: true.
-	 * @param string $cache_id The cache ID.
+	 * @param string|int $cache_id The cache ID.
 	 *    If $cache_id is left empty, an ID will be generated automatically.
 	 *    Manually specifiying a ID is helpful in cases when an integration hook
 	 *    wants to identify particular strings to act upon, but is otherwise
@@ -1151,7 +1151,7 @@ class BBCodeParser
 	 * @param array $parse_tags If set, only parses these tags rather than all of them.
 	 * @return string The parsed string.
 	 */
-	public function parse(string $message, bool $smileys = true, string $cache_id = '', array $parse_tags = []): string
+	public function parse(string $message, bool $smileys = true, string|int $cache_id = '', array $parse_tags = []): string
 	{
 		// Don't waste cycles
 		if (strval($message) === '') {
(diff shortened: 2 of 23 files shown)


Sources/Msg.php
```diff
@@ -1231,8 +1231,7 @@ public static function spellCheck(): void
 
 		// Free resources for enchant...
 		if (isset(Utils::$context['enchant_broker'])) {
-			enchant_broker_free_dict($dict);
-			enchant_broker_free(Utils::$context['enchant_broker']);
+			unset($dict, Utils::$context['enchant_broker']);
```
jdarwood007, line 1234: Note here, the enchant_*free functions are deprecated, they recommend just destroying the object.

Sources/Attachment.php
```diff
@@ -1469,7 +1469,7 @@ public static function create(&$attachmentOptions)
 			);
 		}
 
-		if (!$this->is_image || empty(Config::$modSettings['attachmentThumbnails']) || (empty($attachmentOptions['width']) && empty($attachmentOptions['height']))) {
+		if (empty(Config::$modSettings['attachmentThumbnails']) || (empty($attachmentOptions['width']) && empty($attachmentOptions['height']))) {
```
Sesquipedalian: Why was this change made? If the attached file isn't an image, a thumbnail should never be made.

Sources/BBCodeParser.php
```diff
@@ -2940,7 +2940,7 @@ function ($matches) {
 							return $url;
 						}
 
-						return '[url=&quot;' . str_replace(['[', ']'], ['&#91;', '&#93;'], $full_url->toAscii()) . '&quot;]' . $url . '[/url]';
+						return '[url=&quot;' . str_replace(['[', ']'], ['&#91;', '&#93;'], strval($full_url->toAscii())) . '&quot;]' . $url . '[/url]';
```
Sesquipedalian: This isn't necessary.

Sources/Cookie.php
```diff
@@ -393,8 +393,8 @@ public static function setLoginCookie($cookie_length, $id, $password = '')
 				}
 
 				$alias_cookie = clone $cookie;
-				$alias_cookie->data[3] = $alias_cookie->domain = $domain;
-				$alias_cookie->data[4] = $alias_cookie->path = $path;
+				$alias_cookie->domain = $domain;
+				$alias_cookie->path = $path;
```
Sesquipedalian:
```suggestion
				$alias_cookie->custom_data[3] = $alias_cookie->domain = $domain;
				$alias_cookie->custom_data[4] = $alias_cookie->path = $path;
```

Sources/Profile.php
```diff
@@ -2170,7 +2170,7 @@ public static function backcompat_makeCustomFieldChanges($id, $area, $sanitize =
 		self::$member->save();
 
 		if (!empty($return_errors)) {
-			return $this->cf_save_errors;
+			return self::$cf_save_errors;
```
Sesquipedalian:
```suggestion
			return self::$member->cf_save_errors;
```

Sources/Url.php
```diff
@@ -443,6 +443,8 @@ public function proxied(): object
 			return $proxied;
 		}
 
+		$parsed = $this->parse();
+
```
Sesquipedalian: There is a mistake in this method, but this isn't the right solution. Instead, all occurrences of `$parsed` in the rest of the method should be replaced with `$proxied`.

Sources/Url.php
```diff
@@ -779,7 +781,7 @@ public static function urlToIri(string $url): object
 	 * @param string $url The original URL of the requested resource.
 	 * @return object A new instance of this class for the proxied URL.
 	 */
-	public static function getProxiedUrl(string $url): string
+	public static function getProxiedUrl(string $url): object
```
Sesquipedalian:
```suggestion
	public static function getProxiedUrl(string $url): Url
```

Sources/User.php
```diff
@@ -3174,17 +3176,17 @@ public static function delete(int|array $users, bool $check_not_admin = false):
 			list($user) = $users;
 
 			if ($user == User::$me->id) {
-				isAllowedTo('profile_remove_own');
+				self::isAllowedTo('profile_remove_own');
```
```suggestion
				self::$me->isAllowedTo('profile_remove_own');
```

Sources/User.php
```diff
@@ -3174,17 +3176,17 @@ public static function delete(int|array $users, bool $check_not_admin = false):
...
 			} else {
-				isAllowedTo('profile_remove_any');
+				self::isAllowedTo('profile_remove_any');
```
```suggestion
				self::$me->isAllowedTo('profile_remove_any');
```

Sources/User.php
```diff
@@ -3174,17 +3176,17 @@ public static function delete(int|array $users, bool $check_not_admin = false):
...
 			}
 		} else {
 			foreach ($users as $k => $v) {
 				$users[$k] = (int) $v;
 			}
 
 			// Deleting more than one?  You can't have more than one account...
-			isAllowedTo('profile_remove_any');
+			self::isAllowedTo('profile_remove_any');
```
```suggestion
			self::$me->isAllowedTo('profile_remove_any');
```

Sources/User.php
```diff
@@ -3880,7 +3882,7 @@ public static function isBannedEmail(string $email, string $restriction, string
 
 		// You're in biiig trouble.  Banned for the rest of this session!
 		if (isset($_SESSION['ban']['cannot_access'])) {
-			$this->logBan($_SESSION['ban']['cannot_access']['ids']);
+			User::$me->logBan($_SESSION['ban']['cannot_access']['ids']);
```
```suggestion
			self::$me->logBan($_SESSION['ban']['cannot_access']['ids']);
```

Sources/User.php
```diff
@@ -3889,7 +3891,7 @@ public static function isBannedEmail(string $email, string $restriction, string
 
 		if (!empty($ban_ids)) {
 			// Log this ban for future reference.
-			$this->logBan($ban_ids, $email);
+			User::$me->logBan($ban_ids, $email);
```
```suggestion
			self::$me->logBan($ban_ids, $email);
```

Sesquipedalian (changes requested):
Just a few things.

Sources/Attachment.php
jdarwood007: `$this->image` doesn't exist.  I couldn't figure out why.  2.1 doesn't have this check.  Also we are in a static function, so `$this` can't exist.  We do have `$image` which is a object, but there is no `is_image` property.

Sources/BBCodeParser.php
jdarwood007: This was mostly a complaint by static analysis.  Because toAscii() returns a object: `	public function toAscii(): object`.  So static analysis says you can't use a object in str_replace for the subject.  `strval` can take a object in and is going to return a string.  We could expose a the __toString as a non magic method.  But as well, toAscii seems like it should return a string, but I couldn't be sure why we are still returning the object.

Sources/Cookie.php
jdarwood007: Ahh, I missed that.  It was data and I couldn't find data.  I totally missed custom_data

Sources/Url.php
jdarwood007: Ok, I was thinking that but wasn't sure.  I checked 2.1 and it was using `$parsed` as well, so I assumed that is what we meant.

Sources/User.php
jdarwood007: There is lots of instances of `User::$me` being used here, Should I just fix them all?

Sources/BBCodeParser.php
Sesquipedalian: The static analysis is mistaken in this case. An instance of Stringable will be cast to string automatically when used as a string by str_replace().

The toAscii() method changes the Url object itself. It returns $this for the sake of method chaining.

Sources/User.php
Sesquipedalian: Yeah.

Sources/Attachment.php
jdarwood007: Changed it to check $image.  $image is set by the Image class which returns null if its not a image.  That should do what you are asking?

Sources/BBCodeParser.php
jdarwood007: Well I don't think relying on the __ToString magic method to stringify the object is the best practice to do for production code.  But I reverted the change.

Sources/Attachment.php
```diff
@@ -1469,7 +1469,7 @@ public static function create(&$attachmentOptions)
 			);
 		}
 
-		if (!$this->is_image || empty(Config::$modSettings['attachmentThumbnails']) || (empty($attachmentOptions['width']) && empty($attachmentOptions['height']))) {
+		if (empty($image) || empty(Config::$modSettings['attachmentThumbnails']) || (empty($attachmentOptions['width']) && empty($attachmentOptions['height']))) {
```
Sesquipedalian: This won't quite work, because $image is always unset on line 1346 above.

But you could just add `$is_image = true;` inside the if block that starts on line 1321, and then change this line here to use `empty($is_image)`.

Sources/Cookie.php
Sesquipedalian: Not your fault. I'm the one who made the mistake to begin with.

Sources/Attachment.php
Sesquipedalian: I've commented on the new version below.

Sesquipedalian (approved):

PR #7930, opened 2023-12-03
Support HTTP 1.1 Chunked transfer

### LINKED ISSUE #7929: latest-news.js is butchered in SMF 3
#### Description
Viewing the admin center in SMF 3, you just see the standard "You are unable to connect to simplemachines.org's latest news file." placeholder instead of the latest news.

### Steps to reproduce
1. Go to the admin center
2. Look at where the latest news should be

- Database Type: MariaDB (also occurs with PostgreSQL)
- Database Version:
- PHP Version: 8.3.0

A comparison of the latest-news file in 2.1 and 3 shows that in SMF 3, there's a "2002" and blank line before the window.smfAnnouncements definition and a blank line and a "0" at the end of the code after the window.setInnerHTML call.

Here's what I see in SMF 3:
```
2002

window.smfAnnouncements = [
	{
		subject: 'SMF 2.1.4 has been released',
		href: 'https://www.simplemachines.org/community/index.php?topic=586097.0',
		time: 'Jun 10, 2023, 05:21 PM',
		author: 'shawnb61',
		message: 'SMF 2.1.4 includes security updates and numerous bug fixes. We recommend updating as soon as possible.'
	},
...
}

if (document.getElementById('credits'))
	setInnerHTML(document.getElementById('credits'), getInnerHTML(document.getElementById('credits')).replace(/anyone we may have missed/, '<span title="And you thought you had escaped the credits, hadn\'t you, Zef Hemel?">anyone we may have missed</span>'));

0
```

jdarwood007:
Fixes #7929

@Oldiesmann Test and confirm

### DIFF
--- a/Sources/WebFetch/APIs/SocketFetcher.php
+++ b/Sources/WebFetch/APIs/SocketFetcher.php
@@ -94,6 +94,27 @@ class SocketFetcher extends WebFetchApi
 	private bool $keep_alive;
 
 	/**
+	 * @var bool
+	 *
+	 * Whether the response is transfered as chunked data.
+	 */
+	private bool $is_chunked = false;
+
+	/**
+	 * @var int
+	 *
+	 * Whether to keep the socket connection open after the initial request.
+	 */
+	private int $buffer_size = 4096;
+
+	/**
+	 * @var string
+	 *
+	 * The line break
+	 */
+	private string $line_break = "\r\n";
+
+	/**
 	 * @var resource
 	 *
 	 * File pointer for the open socket connection.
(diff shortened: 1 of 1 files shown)


Oldiesmann:
Something else seems to be broken with scheduled tasks now unfortunately. If I choose to run the task now, it says it was successful but the "next due" never changes and nothing is updated in the task log. I also have several tasks with "next due" dates in the past, and nothing has been updated in the task log in a while.

jdarwood007:
I suspect there is another issue then.  The data is updating correctly at least though?  That is the focus on this PR, which is to correct the data not being handled properly due to chunking in HTTP 1.1.


I do get a error when I run it myself still, but its due to a missing Use statement.  I fixed that in: https://github.com/SimpleMachines/SMF/pull/7927/files#diff-0110925af7e28153225ee796276bc6329a889f3631c500c05527ad06a756b223

Oldiesmann:
Tried it locally on my test board on Windows and it worked fine (your fix fixed the issue). Not sure why scheduled tasks aren't working on my server.

Oldiesmann:
Already mentioned this in Discord but will post here as well for anyone who didn't see that. The issue with tasks not getting run on my board was because the Tasks directory was never renamed. I deleted the "tasks" directory, uploaded the one from my local copy ("Tasks" with a capital T) and updated the path in Settings.php and everything is working now. The change in this PR also fixes the issue with the latest-news.js file.

PR #7931, opened 2023-12-03
BackwardCompatibility Refactor

tyrsson:
Removes BackwardCompatibility Trait function exporting
Provides function mapping via Subs-Compat.php
Improves typing for methods that explicitly have phpdoc comments stating what typing should be.


jdarwood007:
Looking good!

As mentioned to you in chat.  Since you fixed some of the type handling here.  Take note I have done work in both #7923 and #7927 (only fixes phpdoc) which are not merged yet which also fix types.  I haven't touched anything inside actions yet as the core code needs fixed before we handle it into actions.
For those wanting to know more why, when the core is fixed, actions are either using the core code correctly, or we need to fix the core to handle the types and update the phpdoc.

tyrsson:
Well, I was thinking on this. If you work those changes and I work the actions then we shouldn't have conflicts and we get better coverage. That however is up to @Sesquipedalian and whether we decide to go this route. Its still kinda up in the air at the moment. I just opened this Draft so he could take a look around.

Sesquipedalian:
I like where this is going. 

My first suggestion is that the right place for all these wrapper functions is the Subs-Compat.php file. Things like this are the reason for that's file's existence.

My second suggestion is that it is premature to kill of the BackwardCompatibilty trait just yet. We need an alternative way to deal with exporting static properties to global variables before we can do that.

So for now, I suggest putting the trait back the way it was and moving all the wrapper functions to Subs-Compat.php. All the places where you have stripped out references to the trait so far can stay that way, and as you work through more of them you can continue in the same vein.

You'll need to leave the trait in place in any classes that have a `prop_names` subarray in their `$backcompat` property until we come up with a good alternative for dealing with the properties.

Also, don't forget to remove all the calls to the `exportStatic()` at the end of each class's file. It looks like some have been missed in this draft.

tyrsson:
The export Static calls were just missed in the process of just getting the forum to load so I could continue :) I'm aware that they need cleaned up :). Will pull off the move to Subs-Compat here shortly and cleanup those files as well as reinstate the Trait.

Sources/Actions/BoardIndex.php
```diff
@@ -38,8 +38,6 @@
  */
 class BoardIndex implements ActionInterface
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian, line 42: Forgot to remove the $backcompat property and the call to exportToStatic() at the end of the file.

Sources/Actions/Moderation/ReportedContent.php
```diff
@@ -38,8 +38,6 @@
  */
 class ReportedContent implements ActionInterface
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian: Forgot to remove the $backcompat property and the call to exportToStatic() at the end of the file.

Sources/Actions/Notify.php
```diff
@@ -32,8 +32,6 @@
  */
 abstract class Notify
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian, line 36: Forgot to remove the $backcompat property and the call to exportToStatic() at the end of the file.

Sources/Alert.php
```diff
@@ -20,7 +20,6 @@
  */
 class Alert implements \ArrayAccess
 {
-	use BackwardCompatibility;
```
Sesquipedalian, line 25: Forgot to remove the $backcompat property and the call to exportToStatic() at the end of the file.

Sources/Attachment.php
```diff
@@ -25,7 +25,6 @@
  */
 class Attachment implements \ArrayAccess
 {
-	use BackwardCompatibility;
```
Sesquipedalian, line 30: Forgot to remove the $backcompat property and the call to exportToStatic() at the end of the file.

Sources/BBCodeParser.php
```diff
@@ -51,8 +51,6 @@
  */
 class BBCodeParser
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian, line 57: Forgot to remove the $backcompat property and the call to exportToStatic() at the end of the file.

Sources/BrowserDetector.php
```diff
@@ -28,8 +28,6 @@
  */
 class BrowserDetector
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian, line 34: Forgot to remove the $backcompat property and the call to exportToStatic() at the end of the file.

Sources/Cookie.php
```diff
@@ -18,8 +18,6 @@
  */
 class Cookie
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian, line 24: Forgot to remove the $backcompat property and the call to exportToStatic() at the end of the file.

Sources/Db/DatabaseApi.php
```diff
@@ -24,7 +24,6 @@
 abstract class DatabaseApi
 {
 	use BackwardCompatibility;
-
```
Sesquipedalian, line 29: Please restore the missing blank line.

Can also remove the 'func_names' subarray from the $backcompat property below.

Sources/ErrorHandler.php
```diff
@@ -28,8 +28,6 @@
  */
 class ErrorHandler
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian, line 34: Forgot to remove the $backcompat property.

Sources/Forum.php
```diff
@@ -193,6 +193,7 @@ class Forum
 	 */
 	public function __construct()
 	{
+		$test = sanitize_chars('testing');
```
Sesquipedalian: Probably not meant to be here.

Sources/Graphics/Image.php
```diff
@@ -32,8 +31,6 @@
  */
 class Image
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian, line 38: Forgot to remove the $backcompat property.

Sources/IP.php
```diff
@@ -20,7 +20,6 @@
  */
 class IP implements \Stringable
 {
-	use BackwardCompatibility;
```
Sesquipedalian, line 25: Forgot to remove the $backcompat property.

Sources/IntegrationHook.php
```diff
@@ -20,8 +20,6 @@
  */
 class IntegrationHook
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian, line 26: Forgot to remove the $backcompat property.

Sources/Lang.php
```diff
@@ -21,7 +21,6 @@
 class Lang
 {
 	use BackwardCompatibility;
-
```
Sesquipedalian: Please restore the missing blank line.

Can also remove the 'func_names' subarray from the $backcompat property below.

Sources/Logging.php
```diff
@@ -25,8 +25,6 @@
  */
 class Logging
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian, line 31: Forgot to remove the $backcompat property.

Sources/Mail.php
```diff
@@ -20,8 +20,6 @@
  */
 class Mail
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian, line 27: Forgot to remove the $backcompat property.

Sources/Menu.php
```diff
@@ -96,7 +96,6 @@
  */
 class Menu implements \ArrayAccess
 {
-	use BackwardCompatibility;
```
Sesquipedalian, line 101: Forgot to remove the $backcompat property.

Sources/Msg.php
```diff
@@ -27,7 +27,6 @@
  */
 class Msg implements \ArrayAccess
 {
-	use BackwardCompatibility;
```
Sesquipedalian, line 32: Forgot to remove the $backcompat property.

Sources/PageIndex.php
```diff
@@ -20,8 +20,6 @@
  */
 class PageIndex implements \Stringable
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian, line 26: Forgot to remove the $backcompat property.

Sources/QueryString.php
```diff
@@ -20,8 +20,6 @@
  */
 class QueryString
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian, line 26: Forgot to remove the $backcompat property.

Sources/Security.php
```diff
@@ -20,8 +20,6 @@
  */
 class Security
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian, line 26: Forgot to remove the $backcompat property.

Sources/Security.php
```diff
@@ -552,13 +550,4 @@ public static function kickGuest(): void
 	}
 }
 
-// Export public static functions and properties to global namespace for backward compatibility.
-if (is_callable(__NAMESPACE__ . '\\Security::exportStatic')) {
-	Security::exportStatic();
-}
-
-// Some functions have moved.
-class_exists('SMF\\SecurityToken');
```
Sesquipedalian, line 561: If this line is removed here, it needs to be added in the appropriate section of Subs-Compat.php.

Sources/SecurityToken.php
```diff
@@ -24,8 +24,6 @@
  */
 class SecurityToken extends \ArrayObject
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian, line 30: Forgot to remove the $backcompat property.

Sources/Session.php
```diff
@@ -26,8 +26,6 @@
  */
 class Session implements \SessionHandlerInterface
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian, line 32: Forgot to remove the $backcompat property.

Sources/Subs-Compat.php
```diff
@@ -13,7 +13,7 @@
  *
  * @version 3.0 Alpha 1
  */
-
+namespace SMF;
```
Sesquipedalian: This puts all the functions below into the SMF namespace. But we need them to be in the global namespace, lest we defeat the whole point of this exercise.

Sources/TaskRunner.php
```diff
@@ -38,8 +38,6 @@
  */
 class TaskRunner
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian, line 44: Forgot to remove the $backcompat property.

Sources/Theme.php
```diff
@@ -34,8 +34,6 @@
  */
 class Theme
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian, line 40: Forgot to remove the $backcompat property.

Sources/Time.php
```diff
@@ -18,7 +18,6 @@
  */
 class Time extends \DateTime implements \ArrayAccess
 {
-	use BackwardCompatibility;
```
Sesquipedalian, line 23: Forgot to remove the $backcompat property.

Sources/TimeZone.php
```diff
@@ -18,8 +18,6 @@
  */
 class TimeZone extends \DateTimeZone
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian, line 24: Forgot to remove the $backcompat property.

Sources/Unicode/Utf8String.php
```diff
@@ -28,8 +26,6 @@
  */
 class Utf8String implements \Stringable
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian: Forgot to remove the $backcompat property.

Sources/Url.php
```diff
@@ -26,8 +26,6 @@
  */
 class Url implements \Stringable
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian: Forgot to remove the $backcompat property.

Sources/Utils.php
```diff
@@ -19,15 +19,14 @@
 class Utils
 {
 	use BackwardCompatibility;
-
```
Sesquipedalian: Please restore the missing blank line.

Can also remove the 'func_names' subarray from the $backcompat property below.

Sources/WebFetch/WebFetchApi.php
```diff
@@ -22,8 +21,6 @@
  */
 abstract class WebFetchApi implements WebFetchApiInterface
 {
-	use BackwardCompatibility;
-
```
Sesquipedalian, line 28: Forgot to remove the $backcompat property.

Sesquipedalian (dismissed):
Most comments are just reminders to remove the $backcompat property, but there are a few more substantial questions and comments in there.

Sources/Subs-Compat.php
tyrsson: This is simply to make life simpler until I alias the namespaces with use statements ;)

Sources/Forum.php
tyrsson: It was only for testing, has been removed since then. I forgot to remove it before my initial commit.

Sources/TimeZone.php
tyrsson: Im not even working on those yet, where they have been left is a marker for project wide search ;)

Sources/Security.php
tyrsson: Ok, but I must admit, I am having a hard time grasping the purpose of this... Maybe I just can't see the forest for the trees here.

Sources/Actions/Notify.php
tyrsson: Look at the Sources/Actions/Admin/* classes. I have not worked on hardly any of these top level classes other than to get the forum to load. As I commit I note which the commit is finishing in the commit message. Those are the ones ready for review. Will save us both a lot of time :)

Sources/Utils.php
tyrsson: I only targeted one function/method pair from this file for testing very early on. The single function has been defined as first in the list will move when I get to that class.

Sources/Subs-Compat.php
tyrsson: i.e. 
use SMF\Actions; (covers all the classes that have been refactored so far ;) )

Sources/Security.php
Sesquipedalian: As with most things that might make one say, "That's odd. What's that for?", this is for backward compatibility support. This particular line ensures that when SMF\Security is loaded, SMF\SecurityToken is loaded at the same time. This is done because some of the backward compatibility functions supplied by SecurityToken used to reside in Security. 

This line itself was added before I created the section of Subs-Compat where various classes are loaded when Confg::$backward_compatibility is enabled (i.e. the same one where you are now adding the backward compatibility function wrappers). Now that we are centralizing all of that compatibility loading into one place, this line is an excellent candidate to be moved there—or better yet, to be replaced with wrapper functions to call SecurityToken's various static methods.

Sources/Subs-Compat.php
Sesquipedalian: Okay. But when I'm asked to give a review, I can only review the code that is actually in front of me. So that's what you're getting comments on. 🙂

Sources/Security.php
tyrsson: My intention is to move all required backcompat function to that file. The only responsibility for the trait after that is global variables since I have not thought of a better way to accomplish that. The variable usage from what I have seen so far is nowhere near as prolific as the function usage though so the cleanup after when the time comes to remove it will be far less pervasive.

This should also allow us to remove all of the class_exist calls from Subs-Compat as well since they will be imported with use statements. Or am I missing something?

Sesquipedalian: No, not missing anything. Once all the backcompat wrapper functions are added to Subs-Compat.php, the class_exists() calls be removed from it.

tyrsson:
@Sesquipedalian @live627 @jdarwood007 
Note: I have marked all of the various methods that this deprecates with a deprecation notice since 3.0 so you can search for that project wide and any that intersect with a "Backward compatibility wrapper" comment this covers. Searching it in my branch reports 210 methods as deprecated.
I've done some initial testing, but it needs tested in-depth and I would appreciate all the help I can get on it. Any suggestions on mods that use the largest number of require compat would also be helpful :)

tyrsson:
@Sesquipedalian how well was the back compat tested before I started work on the refactor?

tyrsson:
@Sesquipedalian 

I can confirm that this mod:
https://custom.simplemachines.org/index.php?mod=1836

Is working with this PR.  It does not however work with release-3.0.

Also, @DiegoAndresCortes ST Shop mod does not, nor does LightPortal. Still looking for the root cause of that.

tyrsson:
I can also confirm that this mod:

Works as it should (after fixing the fact that it uses a deprecated hook). 

https://custom.simplemachines.org/index.php?mod=4314

tinoest:
Can I check how you access the old globals that used to exist? user_info for example, or is the idea those are dropped?

There is an error when calling loadTemplate, you need to return something in loadTemplate in Theme.php if loaded is 1.

tinoest:
Error message

2: Undefined variable $filename

File /var/www/html/Sources/Subs-Compat.php Line 4409

It's fileName in the function call.

tinoest:
Another one.

`function loadMemberData($users = [], int $type = self::LOAD_BY_ID, ?string $dataset = null): array`

You can't use self outside of the object status, so this is what's needed. 

`function loadMemberData($users = [], int $type = 0, ?string $dataset = null): array`

tyrsson:
> Can I check how you access the old globals that used to exist? user_info for example, or is the idea those are dropped?
> 
> There is an error when calling loadTemplate, you need to return something in loadTemplate in Theme.php if loaded is 1.

user_info is dropped in favor of User::$me iirc.

If there is a error is will be in the loadTemplate method. As you see the function loadTemplate is simply a proxy to that class method. I will look into it though. What is the exact error you are getting, it would help a lot.


> Error message
> 
> 2: Undefined variable $filename
> 
> File /var/www/html/Sources/Subs-Compat.php Line 4409
> 
> It's fileName in the function call.
Yes, there was that one and one for $key where it was $Key.  Typo's




> Another one.
> 
> `function loadMemberData($users = [], int $type = self::LOAD_BY_ID, ?string $dataset = null): array`
> 
> You can't use self outside of the object status, so this is what's needed.
> 
> `function loadMemberData($users = [], int $type = 0, ?string $dataset = null): array`

Actually what is needed there is SMF\User::LOAD_BY_ID since it is supposed to use the class const as the default value. I fixed two instances of where it should have been using constants from SMF\Group as well.

tinoest:
The theme one was bool is meant to be returned null provided or something to that effect. I just added return true to the bottom of the function to clear it.

Was aware it wasn’t your code breaking it, just logged it as an error. 

I was checking that backwards compatibility was being maintained with user_info global, worked out the load order has changed, but I can amend the mod to suit.

tyrsson:
If there is any chance that the function may exit void (which would have provided the null return) then I need to modify the typing.

tyrsson:
@tinoest sorry I didn't reply to the backcompat question.

If you check the BackwardCompatibility Trait that is where the class property to global var exporting is handled. You can run a project wide search for $backcompat and any class property that has a prop_names key is exporting those to global scope.

Sesquipedalian:
> @Sesquipedalian how well was the back compat tested before I started work on the refactor?

It is possible that I missed some things here and there, since we don't have unit testing yet.



> Can I check how you access the old globals that used to exist? user_info for example, or is the idea those are dropped?

> user_info is dropped in favor of User::$me iirc.

The old $user_info global has been deprecated but it still exists. It is just a reference to User::$me and is not used by SMF 3.0's own code.

Sesquipedalian:
I see that you've asked for a new review, @Tyrsson. Is that (a) because you believe this PR is ready to merge and you want me to review everything, or (b) because you have specific things you want me to look at? If the latter, what do you want me to look at in particular?

tyrsson:
It's about as ready for review as it can be. Its coverage is at minimum as good as the original back compat layer. I covered the few cases that I saw where things had not been included but should. But, @Sesquipedalian you know the code A LOT better than I do so if someone is going to catch it, it will probably be you.

Bare in mind that none of the deprecated methods have yet to be removed. I wanted a couple rounds of testing before performing that last step since it effectively leaves the prior functionality as it was in place until this is reviewed and approved. Its a simple task then to remove the deprecated methods.

Please look over the Backcompat Trait very closely. Any thoughts on improvement are more than welcome :)

Sesquipedalian:
Cool. Thanks for the clarification. I'll look through it all soon. Might take a bit, due to the time of year and then scope of there changes.

dragomano:
Some checks with the SMF Tracy Debugger mod
```php
public function modifyModifications(array &$subActions)
{
    // It causes "argument #2 ($method) must be of type string, array given"
    $subActions['tracy_debugger'] = [$this, 'settings'];

    // It works as expected
    // $subActions['tracy_debugger'] = __CLASS__ . '::settings#';
}
```

```php
public function settings(bool $return_config = false)
{
    global $context, $txt;

    // It causes "Indirect modification of overloaded element of SMF\Menu has no effect"
    $context[$context['admin_menu_name']]['tab_data']['description'] = $txt['tracy_description'];

    // It works as expected
    // \SMF\Menu::$loaded['admin']->tab_data['description'] = $txt['tracy_description'];
}
```

tinoest:
Other than the main loading of SMF being slightly different (not caused by this PR)

TinyPortal works ok with a few minor changes to it's logic.

Sesquipedalian (changes requested):
1. I don't think it makes sense to have those factory and provider methods in the BackwardCompatibilty trait. Putting them into that trait makes them available for all classes that use the trait, but they are each only applicable to a subset of those classes. 

	For the cases where these factory methods use direct maps from passed function names to static method names, I think it makes more sense to simply call the relevant static methods directly from the Subs-Compat functions. 

	For example:

	```
	function utf8_strtolower(string $string): string
	{
		return (string) Utf8String::utf8StringFactory(__FUNCTION__, $string);
	}
	```

	should simply be:

	```
	function utf8_strtolower(string $string): string
	{
		return (string) Utf8String::create($string)->convertCase('lower');
	}
	```

	The situation is a bit different for the subActionProvider and profileSubActionProvider methods. They make sense to me, since they condense a lot of stuff nicely. But they still belong somewhere else since they aren't applicable to every class that would use the BackwardCompatibility trait.

	Perhaps subActionProvider should go into a new SMF\Actions\BackwardCompatibility trait and profileSubActionProvider into a new SMF\Actions\Profile\BackwardCompatibility trait, with each of those traits in turn using the base SMF\BackwardCompatibility trait. Or perhaps this would be a good case for using abstract classes. At any rate, moving these methods out of the base SMF\BackwardCompatibility trait into something more specific would be good.

2. There are a number of cases where classes have static methods that exist solely for the purpose of facilitating backward compatibility. For most of these, it would make sense to move the entire contents of the method into the relevant Subs-Compat function.

	For example, if this function in Subs-Compat.php:

	```
	function call_helper(mixed $input, bool $return = false): mixed
	{
		return SMF\Utils::call_helper($input, $return);
	}
	```

	were changed to this:

	```
	function call_helper(mixed $input, bool $return = false): mixed
	{
		$callable = Utils::getCallable($input);

		// Just return the callable if that's all we were asked to do.
		if ($return) {
			return $callable;
		}

		call_user_func($callable);
	}
	```

	... then the Utils::call_helper method could be deleted entirely.

tyrsson:
@Sesquipedalian I considered the suggestions on the composed trait approach, but thought better of it since this is basically just a temp shim. I will implement those sometime over the weekend.

Also, agree on the other suggestions.

Hope to have it completed by maybe Tuesday of this coming week.

Sources/Subs-Compat.php
```diff
@@ -24,28 +33,4918 @@
...
-	class_exists('SMF\\User');
-	class_exists('SMF\\Graphics\\Image');
-	class_exists('SMF\\WebFetch\\WebFetchApi');
+	 * functions to be available.
+	 */
+
+	function Activate()
+	{
+		return Actions\Activate::call();
+	}
```
dragomano, line 56: Here and below, there is a lack of indentation between functions

Sources/Subs-Compat.php
```diff
@@ -24,28 +33,4918 @@
...
+	{
+		return SMF\Theme::load($id, $initialize);
+	}
+	function loadEssentialThemeData(): void
+	{
+		SMF\Theme::loadEssential();
+	}
+	function loadTemplate(string $template_name, string|array $style_sheets = [], bool $fatal = true): bool
+	{
+		return SMF\Theme::loadTemplate($template_name, $style_sheets, $fatal);
```
dragomano, line 5108: Try to go to Smileys and Message Icons, and you will get "Return value must be of type bool, null returned"

Sources/Subs-Compat.php
tyrsson: Fixed

Sources/Subs-Compat.php
tyrsson: Fixed with return type ?bool

I tested it. Please let me know if this does not fix it on your end.

tyrsson:
@Sesquipedalian I think I covered  your requested changes.

I maintained all of the factory/provider methods although I moved them to their respective classes so they are not available via composition to classes that can not and should not have access to them. In that context I think they make sense since it allows us to remove many no longer needed methods and provides a consistent pattern to back compat usage throughout the application.

I introduced the two new traits and implemented those. One usage of particular note is that Actions\TrackIP uses the Profile\BackwardCompatibility Trait instead of the Actions\ trait. Reason for this is that the Profile traits subActionProvider signature already had support for the $memID argument and I did not want to introduce a signature change for a single method and argument.

I moved the discussed methods from Utils to Subs-Compat and removed them from Utils. Once this is approved I will then remove the related deprecated methods that the traits replace.

dragomano:
To make backward compatibility work in the case of permissions, we need to modify the code of the integrateLoadPermissions method in the Sources/Actions/Admin/Permissions.php:

find
```php
// Provide a practical way to modify permissions.
IntegrationHook::call('integrate_load_permissions', [&self::$permission_groups, &$permissions_by_scope, &self::$left_permission_groups, &$hidden_permissions, &$relabel_permissions]);
```

replace with
```php
// Provide a practical way to modify permissions.
IntegrationHook::call('integrate_load_permissions', [&self::$permission_groups, &$permissions_by_scope, &self::$left_permission_groups, &$hidden_permissions, &$relabel_permissions]);

if (isset(self::$permission_groups['membergroup'])) {
	self::$permission_groups['global'] = array_merge(self::$permission_groups['global'], self::$permission_groups['membergroup']);
	unset(self::$permission_groups['membergroup']);
}

if (isset($permissions_by_scope['membergroup'])) {
	$permissions_by_scope['global'] += $permissions_by_scope['membergroup'];
	unset($permissions_by_scope['membergroup']);
}
```

tyrsson:
> To make backward compatibility work in the case of permissions, we need to modify the code of the integrateLoadPermissions method in the Sources/Actions/Admin/Permissions.php:
> 
> find
> 
> ```
> // Provide a practical way to modify permissions.
> IntegrationHook::call('integrate_load_permissions', [&self::$permission_groups, &$permissions_by_scope, &self::$left_permission_groups, &$hidden_permissions, &$relabel_permissions]);
> ```
> 
> replace with
> 
> ```
> // Provide a practical way to modify permissions.
> IntegrationHook::call('integrate_load_permissions', [&self::$permission_groups, &$permissions_by_scope, &self::$left_permission_groups, &$hidden_permissions, &$relabel_permissions]);
> 
> if (isset(self::$permission_groups['membergroup'])) {
> 	self::$permission_groups['global'] = array_merge(self::$permission_groups['global'], self::$permission_groups['membergroup']);
> 	unset(self::$permission_groups['membergroup']);
> }
> 
> if (isset($permissions_by_scope['membergroup'])) {
> 	$permissions_by_scope['global'] += $permissions_by_scope['membergroup'];
> 	unset($permissions_by_scope['membergroup']);
> }
> ```

@Sesquipedalian, @jdarwood007  what's your thoughts on this?

Sesquipedalian:
> @Sesquipedalian, @jdarwood007 what's your thoughts on this?

On the one hand, @Bugo usually knows what he is talking about. On the other hand, we don't have an explanation of why these suggested changes are necessary. 

So, @Bugo, please open an issue to explain what is wrong with this hook and why we need to change it.

Regarding this PR, it doesn't change anything about that integration hook, so I don't think we need to worry about it here.

Sources/Actions/BackwardCompatibility.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+declare(strict_types=1);
```
Sesquipedalian: Moving this below the license block will keep some of our other continuous integration scripts happier.

Sources/Actions/Profile/BackwardCompatibility.php
```diff
@@ -0,0 +1,73 @@
+<?php
+
+declare(strict_types=1);
```
Sesquipedalian: Moving this below the license block will keep some of our other continuous integration scripts happier.

jdarwood007:
@Tyrsson You have a conflict

Otherwise, I think we should merge and work out the issue with the permissions when we have more details.  I'm not certain what the problem is at the moment.

Sources/Actions/BackwardCompatibility.php
```diff
@@ -0,0 +1,55 @@
...
+		self::load();
+
+		if (is_string($sa)) {
+			// make sure its lowecase
+			if (array_key_exists($sa, self::$subactions)) {
+				self::$obj->subaction = $sa;
+			}
+		}
+
+		if(is_string($activity)) {
```
Sesquipedalian:
```suggestion
		if (is_string($activity)) {
```

Sources/Actions/BackwardCompatibility.php
```diff
@@ -0,0 +1,55 @@
...
+	{
+		if ($return_config) {
+			return self::getConfigVars();
+		}
+
+		self::load();
+
+		if (is_string($sa)) {
+			// make sure its lowecase
+			if (array_key_exists($sa, self::$subactions)) {
```
Sesquipedalian: Based on the comment, I assume you meant to do this:

```suggestion
			// Make sure it's lowercase.
			$sa = strtolower($sa);
			
			if (array_key_exists($sa, self::$subactions)) {
```

Sources/Subs-Compat.php
```diff
@@ -24,28 +33,5724 @@
...
+	 *
+	 * @param int $min Minumum value. Default: 0.
+	 * @param int $max Maximum value. Default: PHP_INT_MAX.
+	 * @return int A random integer.
+	 */
+	function randomInt(int $min = 0, int $max = PHP_INT_MAX): int
+	{
+		return random_int($min, $max);
+	}
+
```
Sesquipedalian: These two need be moved back to SMF\Utils. 

The comments in their documentation about only existing for backward compatibility purposes refers to the fact we can now use `random_int()` and `random_bytes()` directly in our code. In SMF 2.1 we couldn't be sure that those functions would be available, and so we used `$smcFunc['random_int']()` and `$smcFunc['random_bytes']()` as wrapper functions that would call a userland library for us if the real `random_int()` and `random_bytes()` weren't available.

Because these two methods are used to backstop the current iteration of `$smcFunc`, they are structurally different than other backward compatibility functions. One could say that they exist for backward compatibility purposes, but not for BackwardCompatibility purposes. 😉

Sesquipedalian (changes requested):
1. In my previous review, I made this comment:

	> There are a number of cases where classes have static methods that exist solely for the purpose of facilitating backward compatibility. For most of these, it would make sense to move the entire contents of the method into the relevant Subs-Compat function.

	This was meant to include to the various `backCompatProvider` methods that are called from some of the functions in Subs-Compat.php.

	For example:
	```
	function AutoSuggestHandler(?string $suggest_type = null): ?bool
	{
		return Actions\AutoSuggest::backCompatProvider(suggest_type: $suggest_type, callHandler: true);
	}
	```
	should simply be:
	```
	function AutoSuggestHandler(?string $suggest_type = null): ?bool
	{
		if (isset($suggest_type)) {
			return Actions\AutoSuggest::checkRegistered($suggest_type);
		}

		Actions\AutoSuggest::call();
	}
	```

2. You forgot to delete the existing backward compatibility static methods from many of the classes that no longer need them.

3. Don't forget to run PHP-CS-Fixer. It flagged a bunch of little things when I checked out this PR most recently. 🙂

Sources/Actions/BackwardCompatibility.php
tyrsson: Actually that should have been removed because we can't normalize it since there is like 3-4 subactions that are not all lowercase. Seems like every time I try to normalize based of something always being done a certain way... wrong answer cause there is always just a handful of cases that violates it.

tyrsson:
I'll work on this first thing in the morning.

dragomano:
> On the one hand, @Bugo usually knows what he is talking about. On the other hand, we don't have an explanation of why these suggested changes are necessary.
> 
> So, @Bugo, please open an issue to explain what is wrong with this hook and why we need to change it.
> 
> Regarding this PR, it doesn't change anything about that integration hook, so I don't think we need to worry about it here.

Currently, SMF 2.1 has no problem with this hook. So, I can't create an issue now. But when this PR will be merged, the problem will appear. In particular, I've been tested adding permissions with Light Portal.

Before PR changes:

sshot-14

After PR changes:

First of all, we see a missing title on the permissions block:

sshot-15

And we have some new errors in the Error Log:

sshot-17
sshot-16

In SMF 3.0, for some reason, the self::$permission_groups parameter in this hook stops having a `membergroup` key and uses the `global` key. My code just adds a little tweak to get around this :)

Sources/Actions/BackwardCompatibility.php
tyrsson: updated the comment to reflect actual usage which is to verify its a supported subaction.

tyrsson:
@Sesquipedalian @jdarwood007 @dragomano 

I *think* it covered everything. Please let me know if I missed anything.

dragomano:
I think, randomBytes and randomInt should be static.

Sesquipedalian:
> I think, randomBytes and randomInt should be static.

Yes, they should.

Sesquipedalian:
There were a few changes to make, but it was simpler to just make them than to document them.

Thanks for this, @Tyrsson. Excellent work!

PR #7939, opened 2023-12-04
Renames tasks directory to Tasks during upgrade

### LINKED ISSUE #7932: Upgrader does not fix $tasksdir
#### Description
Just a reference from #7930.  When tasks was renamed to 'Tasks' to fix other issues and align to more standards, the underlaying setting is not updated in Settings.php.  Upgrader needs to resolve this.
Setting this to Alpha 2 as we don't have the new upgrader logic, but it needs tracked so we fix it.

Sesquipedalian:
Fixes #7932

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -1839,6 +1839,18 @@ function DeleteUpgrade()
 		'upgradeData' => null,
 	);
 
+	// Fix case of Tasks directory.
+	if (
+		is_dir(Config::$tasksdir)
+		&& realpath(Config::$tasksdir) !== realpath(dirname(Config::$tasksdir) . '/Tasks')
+		&& is_writable(Config::$tasksdir)
+		&& is_writable(dirname(Config::$tasksdir))
+	) {
+		rename(Config::$tasksdir, dirname(Config::$tasksdir) . DIRECTORY_SEPARATOR . 'Tasks_temp');
+		rename(dirname(Config::$tasksdir) . DIRECTORY_SEPARATOR . 'Tasks_temp', dirname(Config::$tasksdir) . DIRECTORY_SEPARATOR . 'Tasks');
+		$changes['tasksdir'] = dirname(Config::$tasksdir) . '/Tasks';
+	}
+
 	// Are we in maintenance mode?
 	if (isset($upcontext['user']['main']))
 	{
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -1842,12 +1842,30 @@ function DeleteUpgrade()
 	// Fix case of Tasks directory.
 	if (
 		is_dir(Config::$tasksdir)
-		&& realpath(Config::$tasksdir) !== realpath(dirname(Config::$tasksdir) . '/Tasks')
+		&& basename(Config::$tasksdir) !== 'Tasks'
 		&& is_writable(Config::$tasksdir)
 		&& is_writable(dirname(Config::$tasksdir))
 	) {
+		// Do 'tasks' and 'Tasks' both exist?
+		if (
+			!empty(fileinode(realpath(dirname(Config::$tasksdir) . '/tasks')))
+			&& !empty(fileinode(realpath(dirname(Config::$tasksdir) . '/Tasks')))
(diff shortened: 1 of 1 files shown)


other/upgrade.php
```diff
@@ -1839,6 +1839,13 @@ function DeleteUpgrade()
 		'upgradeData' => null,
 	);
 
+	// Fix case of Tasks directory.
+	if (is_dir(Config::$tasksdir) && is_writable(Config::$tasksdir) && is_writable(dirname(Config::$tasksdir))) {
+		rename(Config::$tasksdir, dirname(Config::$tasksdir) . DIRECTORY_SEPARATOR . 'Tasks_temp');
+		rename(dirname(Config::$tasksdir) . DIRECTORY_SEPARATOR . 'Tasks_temp', dirname(Config::$tasksdir) . DIRECTORY_SEPARATOR . 'Tasks');
+		$changes['tasksdir'] = dirname(Config::$tasksdir) . '/Tasks';
```
jdarwood007, line 1869: I assume this should be $sourcedir?

Sesquipedalian: In 99.999% of cases, `dirname(Config::$tasksdir)` will be the same as `Config::$sourcedir`, yes. But in theory someone might have moved it, so it's best to use `dirname(Config::$tasksdir)` to be sure.

jdarwood007: I missed the dirname() call.

Sesquipedalian: Ah. Got it. Easy enough to overlook. 🙂

jdarwood007: Only thing I would suggest is maybe we check if we need to rename it or not.  I wonder if realpath check would suffice.  `realpath($taskdir) != realpath(dirname($taskdir) . '/Tasks')`?

jdarwood007: Also, is_dir('tasks') will not be true on case sensitive systems.  Is this intended to handle that case?

Sesquipedalian: Good idea.

Sesquipedalian: > Also, is_dir('tasks') will not be true on case sensitive systems. Is this intended to handle that case?

If `./Sources/tasks` has already been renamed to `./Sources/Tasks`, then getting false from that check is exactly what we want.

jdarwood007: Well not if `$tasksdir` is set to `$sorucedir . '/tasks' and we have now uploaded it as 'Tasks'.  Resulting in on case sensitive OS having both a 'tasks' and 'Tasks' directory.  Happened on one of my test boards when I manually updated it.

Sesquipedalian: Ah. Yes, now I see what you mean. Hm... that does pose a bit of a problem, doesn't it? Got any ideas?

jdarwood007: Not sure yet.  Windows doesn't care, it sees it as the same.  Linux will see this as if they where named 2 different things.  PHP we can use strcasecmp to check them.  Maybe `strcasecmp($taskdir) != strcasecmp(dirname($taskdir) . '/Tasks');`?

Sesquipedalian: I think I figured out a way to consistently handle all cases.

Sesquipedalian: Oh, wait, nope. I thought realpath() would report the true case of the dir's name on case insensitive file systems. Apparently that was wrong, which means I need to think about this further. Ugh.

Sesquipedalian: Got it now. This is exactly what inodes are for.

jdarwood007:
Honestly, just thinking on this.  Do we really need to specify `$tasksdir`?  Can we just eliminate this as a option for 3.0?  Tasks will always be `$sourcedir . '/Tasks'`.  The entire issue goes away completely.  To add to this, why don't we make Actions directory a config option, Unicode? Cache? Db? Graphics?  Where does it end.

Sesquipedalian:
That's not a bad argument. But in that case, we need to add logic to the upgrader to move a custom tasks dir back to the standard location.

Sesquipedalian:
At any rate, such a change would be Alpha 2, not Alpha 1.

Sesquipedalian:
Ugh, I see this got closed. Apparently that happens automatically if you rename the branch that you want to merge. I'll open a new PR.

PR #7941, opened 2023-12-05
Renames tasks directory to Tasks during upgrade

### LINKED ISSUE #7932: Upgrader does not fix $tasksdir

Sesquipedalian:
Same as #7939 
Fixes #7932

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -1839,6 +1839,18 @@ function DeleteUpgrade()
 		'upgradeData' => null,
 	);
 
+	// Fix case of Tasks directory.
+	if (
+		is_dir(Config::$tasksdir)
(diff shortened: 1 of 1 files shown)


jdarwood007:
Logic looks sound, I will test it tomorrow.

Sesquipedalian:
Thanks.

jdarwood007 (approved):

jdarwood007:
The logic you added works, but I can't get the upgrader to run right now due to other errors.  I think your other PR for upgrade changes for 3.0 files would fix it.  But I can pull the changes out and test them in a separate script and it does the changes.

PR #7943, opened 2023-12-05
Removes random_compat library

Sesquipedalian:
Removes unnecessary random_compat library and improves various attempts to get random data.

Also fixes a bug where we weren't using bin2hex() on the random bytes inserted into the verification code URL in SMF\Actions\VerificationCode::execute().

Labelled as "PHP version support" because this was a polyfill library for PHP versions that we no longer need to support in 3.0.


jdarwood007 (approved):
Looks good.

PR #7945, opened 2023-12-06
Fix Detailed Versions

jdarwood007:
This fixes issues needed to get detailed versions to work correctly.

I've done a change to include a 'root' directory that will hold the cron, subscriptions, ssi, proxy, and any future files at the root of the forum.

To support sources, I changed over to a iterator for sources.  Made things a bit easier actually.  I can do this for the rest if we want.

There is a related PR on the build tools: https://github.com/SimpleMachines/BuildTools/pull/28

SimpleMachines website is updated with that content at the moment.

When the build tool PR merges, this pr can be updated to reference the updated BuildTools repo and then merged.  Or that can be updated any time after as long as its updated before the alpha release.

### DIFF
--- a/Sources/Actions/Admin/ACP.php
+++ b/Sources/Actions/Admin/ACP.php
@@ -1578,8 +1578,9 @@ class ACP implements ActionInterface
 		];
 
 		foreach ($sources_dir as $filename => $file) {
-			if (!$file->isFile() || $file->getFilename() === 'index.php' || $file->getExtension() !== 'php')
+			if (!$file->isFile() || $file->getFilename() === 'index.php' || $file->getExtension() !== 'php') {
 				continue;
+			}
 			foreach ($ignore_sources as $if)
 				if (preg_match('~' . $if . '~i', $filename)){
 					continue 2;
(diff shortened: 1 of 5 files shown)


Sources/Actions/Admin/ACP.php
```diff
@@ -1526,66 +1526,82 @@ public static function getFileVersions(&$versionOptions)
...
-		while ($entry = $sources_dir->read()) {
-			if (substr($entry, -4) === '.php' && !is_dir(Config::$sourcedir . '/' . $entry) && $entry !== 'index.php') {
-				// Read the first 4k from the file.... enough for the header.
-				$fp = fopen(Config::$sourcedir . '/' . $entry, 'rb');
+				$fp = fopen(Config::$boarddir . '/' . $file, 'rb');
 				$header = fread($fp, 4096);
 				fclose($fp);
-
-				// Look for the version comment in the file header.
+
```
live627:
```suggestion

```

jdarwood007: My bad.  New editor, still getting used to it for things like copy and paste.

Sesquipedalian (approved):

PR #7947, opened 2023-12-09
Fix some PM bugs

Oldiesmann:
This fixes four different PM-related bugs:
* Error while trying to view a label for which no messages existed (including inbox)
* Error while viewing search results if $_GET['start'] wasn't set
* Error while searching PMs if no messages matched the given parameters
* The PM popup wouldn't load because of a bad parameter being passed to the function to load messages

### DIFF
--- a/Sources/PersonalMessage/Received.php
+++ b/Sources/PersonalMessage/Received.php
@@ -525,6 +525,7 @@ class Received implements \ArrayAccess
 		}
 
 		$joins = [];
+		self::$recent[$paramskey] = [];
 
 		$where = [
 			'pmr.id_member = {int:me}',
--- a/Sources/PersonalMessage/Received.php
+++ b/Sources/PersonalMessage/Received.php
@@ -524,8 +524,9 @@ class Received implements \ArrayAccess
 			return self::$recent[$paramskey];
 		}
 
-		$joins = [];
 		self::$recent[$paramskey] = [];
+
+		$joins = [];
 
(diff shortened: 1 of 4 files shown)


Sources/PersonalMessage/Search.php
```diff
@@ -321,7 +321,7 @@ public function performSearch(): void
 		// Sort out the page index.
 		Utils::$context['page_index'] = new PageIndex(
 			Config::$scripturl . '?action=pm;sa=search2;params=' . $this->compressed_params,
-			$_GET['start'],
+			((int) $_GET['start'] ?? 0),
```
Sesquipedalian:
```suggestion
			(int) ($_GET['start'] ?? 0),
```

Sources/PersonalMessage/Received.php
```diff
@@ -525,6 +525,7 @@ public static function getRecent(int $label = -1, string $sort = 'pmr.id_pm', bo
 		}
 
 		$joins = [];
+		self::$recent[$paramskey] = [];
```
Sesquipedalian:
```suggestion
		}

		self::$recent[$paramskey] = [];

		$joins = [];
```

Sesquipedalian (changes requested):

Sesquipedalian (approved):

Sesquipedalian:
I like this new feature where multiple suggestions can be committed in a single batch.

PR #7948, opened 2023-12-10
The "(no primary membergroup)" option was missing when editing profiles

Oldiesmann:
The "(no primary membergroup)" option was missing from the profile, preventing you from being able to remove a member's primary group. I discovered this while trying to demote a test account from administrator to a regular user.

### DIFF
--- a/Sources/Profile.php
+++ b/Sources/Profile.php
@@ -1218,7 +1218,6 @@ class Profile extends User implements \ArrayAccess
 			&& (
 				stristr($this->avatar['url'], 'gravatar://')
 				|| !empty(Config::$modSettings['gravatarOverride'])
-				// Gravatar?
 			)
 		) {
 			$this->formatted['avatar'] += [
--- a/Sources/Profile.php
+++ b/Sources/Profile.php
@@ -1212,6 +1212,7 @@ class Profile extends User implements \ArrayAccess
 			'allow_gravatar' => !empty(Config::$modSettings['gravatarEnabled']),
 		];
 
+		// Gravatar?
 		if (
 			$this->formatted['avatar']['allow_gravatar']
 			&& (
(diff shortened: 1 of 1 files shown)


Sources/Profile.php
```diff
@@ -1212,12 +1212,12 @@ public function loadAvatarData(): bool
 			'allow_gravatar' => !empty(Config::$modSettings['gravatarEnabled']),
 		];
 
-		// Gravatar?
 		if (
 			$this->formatted['avatar']['allow_gravatar']
 			&& (
 				stristr($this->avatar['url'], 'gravatar://')
 				|| !empty(Config::$modSettings['gravatarOverride'])
+				// Gravatar?
```
Sesquipedalian, line 1220: This looks like a mistake.

Sources/Profile.php
```diff
@@ -1212,12 +1212,12 @@ public function loadAvatarData(): bool
 			'allow_gravatar' => !empty(Config::$modSettings['gravatarEnabled']),
 		];
 
-		// Gravatar?
 		if (
```
Sesquipedalian, line 1215:
```suggestion
		// Gravatar?
		if (
```

Sources/Profile.php
Sesquipedalian, line 1220:
```suggestion
```

Sesquipedalian (changes requested):
Moved comment seems to be a mistake. Otherwise, looks good.

Sources/Profile.php
Sesquipedalian, line 1220:
```suggestion
		// Gravatar?
		if (
			$this->formatted['avatar']['allow_gravatar']
			&& (
				stristr($this->avatar['url'], 'gravatar://')
				|| !empty(Config::$modSettings['gravatarOverride'])
```

Oldiesmann:
Not sure how that happened. I'll fix it when I get home tonight 🙂

jdarwood007:
PHP check is failing because of lint failure.  Nothing introduced by this PR far as I can tell.

PR #7960, opened 2023-12-22
Fix an issue with the cache quickGet function

Oldiesmann:
quickGet was typed to only expect a string for the function, causing an error when trying to load latest posts from cache on the boardIndex


jdarwood007:
The fail can be ignored here.  It's the lint that is failing.  A separate PR should be ran to fix the lint issues.

tyrsson (approved):

PR #7966, opened 2023-12-23
Implements SMF\Uuid

Sesquipedalian:
Based on https://github.com/Sesquipedalian/Uuid

This will be useful for several planned changes to SMF 3.0

### DIFF
--- a/Sources/Uuid.php
+++ b/Sources/Uuid.php
@@ -1,0 +1,979 @@
+<?php
+
+/**
+ * Simple Machines Forum (SMF)
+ *
+ * @package SMF
+ * @author Simple Machines https://www.simplemachines.org
+ * @copyright 2024 Simple Machines and individual contributors
+ * @license https://www.simplemachines.org/about/smf/license.php BSD
+ *
+ * @version 3.0 Alpha 1
+ */
+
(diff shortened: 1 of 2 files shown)


live627:
```
> php-cs-fixer check --diff
Loaded config default from "/home/runner/work/SMF/SMF/.php-cs-fixer.dist.php".
   1) Sources/Uuid.php
      ---------- begin diff ----------
--- /home/runner/work/SMF/SMF/Sources/Uuid.php
+++ /home/runner/work/SMF/SMF/Sources/Uuid.php
@@ -561,6 +561,7 @@
 
 			if ((string) $forum_uuid === Config::$modSettings['forum_uuid']) {
 				self::$namespace = $forum_uuid->getBinary();
+
 				return;
 			}
 		}

      ----------- end diff -----------
````

Sesquipedalian:
Ugh. I didn't see PHP-CS-Fixer complain about this issue because it was too busy complaining about an issue in some other file that had nothing to do with this PR.

We **_need_** to fix the PHP-CS-Fixer job so that (1) it reports _all_ the problems, not just the first it finds, and (2) it only checks the files that were changed in the PR, not all the files in the repository.

PR #7972, opened 2023-12-24
Makes the [center] BBC work when wrapped around tables

Sesquipedalian:
Without this CSS tweak, `[center][table]...[/table][/center]` doesn't work as expected.

See https://www.simplemachines.org/community/index.php?msg=4166332


DiegoAndresCortes:
It was referenced in the bugs report board as well.
https://www.simplemachines.org/community/index.php?topic=586445.0

DiegoAndresCortes (approved):

PR #7973, opened 2023-12-25
remove error handling in query function

### LINKED ISSUE #7969: [3.0]: Query Datatyp doesn't match error case
### Basic Information

# Expected behavior:
-------------------
Both db code do on an error case (when return value is false),
collect the error message and return this:
https://github.com/SimpleMachines/SMF/blob/e31624b3a96f62b2b52eeb6bc7631eda3449c0f9/Sources/Db/APIs/MySQL.php#L244-L246
error return a string

query function is defined that it return bool or object,
which result in an db error case also in a datatyp error.

I'm not sure in which direction the problem should be solved,
add datatype? or write the error message somewhere else and return false?
throw an exception?
# Actual behavior:
-------------------
TypeError: SMF\Db\APIs\PostgreSQL::query(): Return value must be of type object|bool, string returned in F:\github\SMF2.1\Sources\Db\APIs\PostgreSQL.php on line 320

### Steps to reproduce

1. create any smf query error (like wrong table name)

grafik

### Version/Git revision

3.0 Alpha 1

### Database Engine

All

### Database Version

_No response_

### PHP Version

_No response_

### Logs

_No response_

### Additional Information

_No response_

albertlast:
closed: #7969

### DIFF
--- a/Sources/Db/APIs/MySQL.php
+++ b/Sources/Db/APIs/MySQL.php
@@ -242,10 +242,6 @@ class MySQL extends DatabaseApi implements DatabaseApiInterface
 
 		$ret = @mysqli_query($connection, $db_string, self::$unbuffered ? MYSQLI_USE_RESULT : MYSQLI_STORE_RESULT);
 
-		if ($ret === false && empty($db_values['db_error_skip'])) {
-			$ret = $this->error($connection);
-		}
-
 		// Debugging.
 		if ($this->show_debug) {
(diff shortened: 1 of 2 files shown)


jdarwood007:
The build check fails because PSR alignment from a previous PR that was merged, not this one.

Sesquipedalian:
> The build check fails because PSR alignment from a previous PR that was merged, not this one.

We **really** need to fix that ASAP.

Sesquipedalian (approved):

PR #7976, opened 2023-12-27
Raises minimum database versions

Sesquipedalian:
Does anyone see a reason not to do this?

FYI:
- MySQL 8.0.x is still supported, and 8.0.35 is the latest release in that branch.
- PostgreSQL 12.x is still supported, and 12.17 is the latest release in that branch.

### DIFF
--- a/other/install.php
+++ b/other/install.php
@@ -67,7 +67,7 @@ require_once Config::$sourcedir . '/Subs-Compat.php';
 $databases = [
 	'mysql' => [
 		'name' => 'MySQL',
-		'version' => '5.6.0',
+		'version' => '8.0.35',
 		'version_check' => function () {
 			if (!function_exists('mysqli_fetch_row')) {
 				return false;
@@ -96,7 +96,7 @@ $databases = [
 	],
 	'postgresql' => [
 		'name' => 'PostgreSQL',
-		'version' => '9.6',
+		'version' => '12.17',
 		'version_check' => function () {
 			$request = pg_query(Db::$db->connection, 'SELECT version()');
 			list($version) = pg_fetch_row($request);
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -53,7 +53,7 @@ $GLOBALS['required_php_version'] = '8.0.0';
 $databases = [
 	'mysql' => [
 		'name' => 'MySQL',
-		'version' => '5.6.0',
+		'version' => '8.0.35',
 		'version_check' => function () {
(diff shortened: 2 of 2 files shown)


albertlast:
My rule of the thumb for pg is, 
newst eol is min pg version on >release< of smf.
would be smf release today would be pg 11.

Feel free to change to rule.

albertlast (approved):
i'm fine with pg 12

jdarwood007:
Do we have a technical reason to change?

MySQL 5.7 has more usage than 8.0.  At least according to some stats here: https://wordpress.org/about/stats/

tyrsson (approved):
I can't think of any off the top of my head. Authentication is a little trickier when using 8.0+ via docker but it has been so long I can not recall the details iirc it was due to a native password module being deprecated or something. I'll test it since I have a couple of docker files I can throw at it in the next few days. I do not see a reason why that should prevent this however.

Sesquipedalian:
> Do we have a technical reason to change?

I would like SMF to be able to take advantage of real JSON data types. I know MySQL 8.0 has support for that. But if some version of MySQL 5.x has JSON support (I didn't check), then I suppose we could do that instead.

jdarwood007:
https://dev.mysql.com/doc/refman/5.7/en/json.html
Seems to indicate it does, unless there is a feature you notice we are missing.  I would want to suggest 5.7 to be on the safe side.

albertlast:
Reason for mysql 8.0: https://dev.mysql.com/doc/refman/8.0/en/mysql-nutshell.html
Many Json improvments (look like 5.7 json support is not good)
- json_table
- json operator
- json merge
- last duplicate key win
- etc.

Common Table Expression
Window Function
Datatype "Text" and "Json" can now hav default values (always a struggle in smf since this issue is very hidden)
mb4 is default utf
descending index
index of json by using function index

--> for me is mysql 8.0 the recommanded min,
because of json issue befor, added cte and window function

Sesquipedalian:
Those are _all_ going to be helpful.

Oldiesmann:
MySQL 5.7 just hit "Oracle Sustaining Support" a couple months ago (which, as far as I can tell, means they're not going to support it anymore unless you pay for one of their support plans, and that seems to be more for enterprise environments), so by the time 3.0 final is released hopefully more people will be on MySQL 8 or whatever the MariaDB equivalent is.
We could also use SMF's internal stats collection feature for this, assuming it still works (I believe there was a way for developers/team members to see general details about collected stats but I don't remember)

PR #7978, opened 2023-12-28
Fix caching error on Profile loading

jdarwood007:
To reproduce this
1. Set Caching to level 3
2. Load the board index

Warning: Trying to access array offset on value of type bool in /Sources/User.php on line 5372
Warning: Undefined array key "" in /Sources/User.php on line 5372

This is because $data is not set, which also causes a null compare.

### DIFF
--- a/Sources/User.php
+++ b/Sources/User.php
@@ -5307,6 +5307,10 @@ class User implements \ArrayAccess
 	 */
 	protected static function loadUserData(array $users, int $type = self::LOAD_BY_ID, string $dataset = 'normal'): array
 	{
+		if (!isset(self::$dataset_levels[$dataset])) {
+			$dataset = 'normal';
+		}
+
 		// Keep track of which IDs we load during this run.
 		$loaded_ids = [];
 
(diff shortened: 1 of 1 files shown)


Sources/User.php
```diff
@@ -5370,7 +5370,7 @@ protected static function loadUserData(array $users, int $type = self::LOAD_BY_I
 				}
 
 				// Does the cached data have everything we need?
-				if (self::$dataset_levels[$data['dataset']] >= self::$dataset_levels[$dataset]) {
+				if (is_array($data) && !empty(self::$dataset_levels[$data['dataset']]) && !empty(self::$dataset_levels[$dataset]) && self::$dataset_levels[$data['dataset']] >= self::$dataset_levels[$dataset]) {
```
Sesquipedalian: The value of `self::$dataset_levels['minimal']` is zero, so this won't work correctly. 

Instead, I suggest a two-part solution.

First, on this line:

```suggestion
				if (is_array($data) && self::$dataset_levels[$data['dataset'] ?? 'minimal'] >= self::$dataset_levels[$dataset]) {
```

Then, at the top of this method, add the following in order to ensure that $dataset is valid:

```
		if (!isset(self::$dataset_levels[$dataset])) {
			$dataset = 'normal';
		}
```

Sesquipedalian (changes requested):

Sources/User.php
jdarwood007: Yep, and I even noted it was an int and didn't think of that. I have applied your suggestion.

Sesquipedalian (approved):

PR #7982, opened 2023-12-29
Ensure the member is always an int when setting TFA cookies

Oldiesmann:
#### Description
The member ID was being passed as a string in some cases when calling setTFACookie, causing an error.

### DIFF
--- a/Sources/Cookie.php
+++ b/Sources/Cookie.php
@@ -179,7 +179,7 @@ class Cookie
 
 		// Special case for the login and TFA cookies.
 		if (in_array($this->name, [Config::$cookiename, Config::$cookiename . '_tfa'])) {
-			$this->member = (int) $this->custom_data[0] ?? (int) User::$me->id;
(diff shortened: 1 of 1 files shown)


Sources/Cookie.php
```diff
@@ -179,7 +179,7 @@ public function __construct(
 
 		// Special case for the login and TFA cookies.
 		if (in_array($this->name, [Config::$cookiename, Config::$cookiename . '_tfa'])) {
-			$this->member = $this->custom_data[0] ?? User::$me->id;
+			$this->member = (int) $this->custom_data[0] ?? (int) User::$me->id;
```
Sesquipedalian:
```suggestion
			$this->member = (int) ($this->custom_data[0] ?? User::$me->id);
```

Sesquipedalian (changes requested):

Sesquipedalian (approved):

Issue #7985, opened 2023-12-29
[3.0]: Time format needs to accept a format.


jdarwood007:
### Basic Information

# Expected behavior:
-------------------
No parsing error:

# Actual behavior:
-------------------
```
				return '[time]' . Time::create('@' . $matches[1])->setTimezone(new \DateTimeZone(User::getTimezone()))->format(null, false) . '[/time]';
```
This results in the following errors:
```
Argument '1' passed to format() is expected to be of type string, null given
Too many arguments to function format(). 2 provided, but 1 accepted.
```

This is occurring because setTimezone returns DateTimeZone, which the format() method expects a $format

### Version/Git revision

0fd130ded

### Database Engine

All

Sesquipedalian:
1. What file is this in?
2. Can we have a better title?

Oldiesmann:
> 1. What file is this in?

Sources/Msg.php line 949.

jdarwood007:
Added a fix in PR #7988.  I think it will work

The issue is that setTimezone returns a datetime object, not a SMF\Time object, so the format method needs the format specified as the first argument.  I did a fix by calling the Time::getDateFormat() as the argument.  Another option to fix would be for SMF\Time to overload the DateTime::setTimezone just to return a SMF\Time, which then would allow the SMF\Time::format() to work.

PR #7988, opened 2023-12-30
[3.0] Strict types

jdarwood007:
This is another PR for more strict typing.  This completes (hopefully) most of the sources except Actions.

This PR has a good chance to break things because it touches many of Actions' used code.

A few things I noticed.

1. $row returns from the database always return strings, which gets annoying where ints are expected.  In the future, we will want to look into a PHP ORM.

2. Dynamic Properties are a bit annoying.  Because we let PHP do some force types, it worked, but with strict typing, it does not like it.  I did what I could to work around it.  But for now, the dynamic properties file does not have strict typing turned on.

3. Our usage of __ToString()
https://www.php.net/manual/en/language.oop5.magic.php#object.tostring

> A Stringable object will not be accepted by a string type declaration if strict typing is enabled. If such behaviour is wanted the type declaration must accept Stringable and string via a union type. 

This means that even with __ToString, unless the function we pass to accepts the Stringable class, it won't work.  So (string) casting needs to be done.

4. There are places where the IDE can't determine the variable.  Its unsure.  Things like psalm and other analyzers as we get to adopting them will also fail.  We will need to add `/** @var type $var */` such comments before their usage to help out the analyzer.


Sources/Actions/PersonalMessage.php
```diff
@@ -595,7 +595,7 @@ public function report(): void
 			User::$me->checkSession();
 
 			// Remove the line breaks...
-			$body = preg_replace('~<br ?/?' . '>~i', "\n", $this->body);
+			$body = preg_replace('~<br ?/?' . '>~i', "\n", $pm->body);
```
jdarwood007, line 587: All the changes of $this to $pm in this file fix reporting PM isssue.

Sources/Actions/Groups.php
```diff
@@ -501,7 +501,7 @@ public function requests(): void
 
 				$request = Db::$db->query(
 					'',
-					'SELECT lgr.id_request
+					'SELECT lgr.id_request, lgr.id_group, lgr.id_member
```
jdarwood007, line 504: Fixes a bug regarding approving group membership

Sources/Actions/Admin/SearchEngines.php
```diff
@@ -439,7 +439,8 @@ public function logs(): void
 
 					Utils::$context['spider_logs']['rows'][$k]['data']['viewing']['class'] = $new_url['class'];
 				} else {
-					Utils::$context['spider_logs']['rows'][$k]['data']['viewing']['value'] = $new_url;
+					// @TODO: Indirect modification of overloaded element of SMF\ItemList has no effect in
+					@Utils::$context['spider_logs']['rows'][$k]['data']['viewing']['value'] = $new_url;
```
jdarwood007, line 443: I didn't want to touch this right now, but I got that error.  We will need to rework the logic.

Sources/Actions/Post.php
```diff
@@ -745,7 +745,7 @@ protected function initiateEvent(): void
 			list(Utils::$context['event']) = Event::load((int) $_REQUEST['eventid']);
 		}
 
-		if (!(Utils::$context['event'] instanceof Event)) {
+		if (!isset(Utils::$context['event']) || !(Utils::$context['event'] instanceof Event)) {
```
jdarwood007, line 748: Got this error when I went to a post and tried to make an event.

Sources/Actions/Profile/ShowPermissions.php
```diff
@@ -78,8 +78,8 @@ public function execute(): void
 		// Load all the permission profiles.
 		Permissions::loadPermissionProfiles();
 
-		Board::$info->id = empty(Board::$info->id) ? 0 : (int) Board::$info->id;
-		Utils::$context['board'] = Board::$info->id;
+		$board = empty(Board::$info->id) ? 0 : (int) Board::$info->id;
```
jdarwood007, line 81: This error I got while trying to see a user's permissions.  We can reassign it back like that, so I localized it.  2.1 code reassigned it back into $board,

Sources/Actions/Profile/Tracking.php
```diff
@@ -337,9 +337,12 @@ public function activity(): void
 		Db::$db->free_result($request);
 
 		// Find other users that might use the same IP.
-		$ips = array_unique($ips);
+		$ips = array_filter(array_unique($ips), function($ip) {
+			return !empty($ip);
```
jdarwood007, line 341: This was annoying, but the IP can be a empty string, which we then try to use later and things where not happy.

tyrsson:
In reference to __toString usage. I would actually prefer to type it Stringable|string.

In reference to the Dynamic properties. I would like to remove it.  Its bad practice in general.

In reference to ORM usage. I'm not so sure that would be the way to go. They are usually extremely heavy. A good DBAL, for sure. an ORM.... I'm not convinced.

In reference to:
/** @var type $var */
Psalm will pretty much replace our IDE's SA usage. It's awesome. I look to have us a working PR ready for merge before Tuesday so just hang in there and this will get a lot easier in the next few days. I'm trying to get the back compat merged in so that its included with our baseline.

Sesquipedalian:
> In reference to the Dynamic properties. I would like to remove it.

That's not going to happen in 3.0. Dynamic properties, and the $prop_aliases feature of SMF\DynamicPropertyHelper in particular, are required to enable great swathes of the backward compatibility functionality.

Sources/Msg.php
```diff
@@ -946,7 +950,7 @@ function ($matches) {
 		$message = preg_replace_callback(
 			'~\[time\](\d{0,10})\[/time\]~i',
 			function ($matches) {
-				return '[time]' . Time::create('@' . $matches[1])->setTimezone(new \DateTimeZone(User::getTimezone()))->format(null, false) . '[/time]';
+				return '[time]' . Time::create('@' . $matches[1])->setTimezone(new \DateTimeZone(User::getTimezone()))->format(Time::getDateFormat()) . '[/time]';
```
jdarwood007, line 953: #7985

Sources/Event.php
```diff
@@ -1115,7 +1115,7 @@ protected function getNumDays(): int
 			return 0;
 		}
 
-		return date_interval_format(date_diff($this->start, $this->end), '%a') + ($this->end->format('H') < $this->start->format('H') ? 2 : 1);
+		return ((int) $this->start->diff($this->end)->format('%a')) + ($this->end->format('H') < $this->start->format('H') ? 2 : 1);
```
jdarwood007, line 1118: I refactored this when I came across it because my IDE didn't like the date_diff returns a DateInterval and passed that into date_interval_format, which expects param 1 to b ea string.  A bit of reworking and its doing what it needs to do, type-safe.

Sources/Utils.php
```diff
@@ -1727,11 +1727,12 @@ public static function checkMimeType(string $data, string $type_pattern, bool $i
...
 	 *
 	 * Wrapper function for PHP's chmod().
-	 *
+	 * 
 	 * @param string $path The full path of the file or directory.
+	 * @param array|string|null $chmod Specify a chmod, if left blank, SMF will auto pick the writable values.
 	 * @return bool Whether the file/dir exists and is now writable.
 	 */
-	public static function makeWritable(string $path): bool
+	public static function makeWritable(string $path, array|string|null $chmod = null): bool
```
jdarwood007, line 1735: We have many places that call makeWritable and passing a second argument.  But it doesn't exist.  The code that called this was in Subs-Package.php.  Because it needs to be able to specify the chmod, I opted to fix it here by allowing us to pass it.  Both a string or array of chmod can be passed, then it will work.  However because the function returns true when writable, if your sending a chmod that should remove writable, make sure to expect that you will want a bool of false returned.

jdarwood007:
> In reference to __toString usage. I would actually prefer to type it Stringable|string.

According to PHP docs, the [\Stringable](https://www.php.net/stringable) interface __ToString() must return a type of string.

Psalm should work with that comment format correct?  That is the [official PHP Doc @var format](https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/var.html).  Makes Visual Studio Code happy and it understands some of the more complex variables of what type they are, so it understands the usage later on.

tyrsson:
> > In reference to __toString usage. I would actually prefer to type it Stringable|string.
> 
> According to PHP docs, the [\Stringable](https://www.php.net/stringable) interface __ToString() must return a type of string.
> 
> Psalm should work with that comment format correct? That is the [official PHP Doc @var format](https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/var.html). Makes Visual Studio Code happy and it understands some of the more complex variables of what type they are, so it understands the usage later on.

Yes, psalm should work with that iirc.

tyrsson:
@Sesquipedalian is there any particular reason why we are dragging our feet on merging this?

Sesquipedalian:
Because I have children who have needed my time and attention lately.

Sesquipedalian:
Other members of the dev team are empowered to review and merge PRs. So if you have had a chance to do a thorough review and you approve the PR, submit your review and then merge it. 🙂

tyrsson:
Oh, please don't take my response as criticism my friend. I am aware but I am the "new guy"  :) I'm blessed that my kids are grown (maybe, maybe not). I'm just trying hard not to merge things that I support overall until those things are approved by general consensus.

Sources/PersonalMessage/Received.php
```diff
@@ -296,13 +298,13 @@ public function save(): void
...
-	public function __set(string $prop, $value): void
+	public function __set(string $prop, mixed $value): void
 	{
 		// This is a bitmap where the lowest bit is the read status and the
 		// second bit is the replied status.
 		if ($prop == 'is_read') {
-			$this->unread = !($value & 0b01);
-			$this->replied = $value & 0b10;
+			$this->unread = !((int) $value & 0b01);
+			$this->replied = (boolean) ((int) $value & 0b10);
```
tyrsson: Typo this should be (bool)

Sources/PersonalMessage/Received.php
```diff
@@ -640,9 +642,9 @@ protected function getLabels(): void
 	 * @param int|string $limit Maximum number of results to retrieve.
 	 *    If this is left empty, all results will be retrieved.
 	 *
-	 * @return Generator<array> Iterating over the result gives database rows.
+	 * @return \Generator<array> Iterating over the result gives database rows.
```
tyrsson, line 645: I think we could be more specific here with \Generator array<int|string, Received>

Sources/Tasks/UpdateUnicode.php
```diff
@@ -618,7 +621,7 @@ private function make_temp_dir()
 	 *
 	 * @return string Path to locally saved copy of the file.
```
tyrsson: need @return string|bool

Sources/Group.php
```diff
@@ -2471,32 +2477,32 @@ protected static function canDelete(object $group): bool
...
 
 	/**
 	 * Returns whether people can search for members of this group.
 	 *
-	 * @param object $group An instance of this class.
+	 * @param self $group An instance of this class.
 	 * @return bool Whether the group can be deleted.
 	 */
-	protected static function canSeach(object $group): bool
+	protected static function canSeach(self $group): bool
```
tyrsson: Possible type in the method name? I would expect canSearch but can recall what it used to be.

Sources/PersonalMessage/Received.php
jdarwood007: Static analysis does not like that because it changes the type then detected elsewhere.  Its correct to say its an array because elsewhere in that file we use it, we pass the $row into a `new self($row)`, which the constructor takes as the properties to load up the object.

Sources/Group.php
jdarwood007: Its compat code:
```
		'can_search' => __CLASS__ . '::canSearch',
```
Just a typo.  I corrected it

Sources/ServerSideIncludes.php
```diff
@@ -1039,7 +1066,7 @@ public static function topTopics($type = 'replies', $num_topics = 10, $output_me
 	 * @param string $output_method The output method. If 'echo', displays a list of topics, otherwise returns an array of info about them
 	 * @return void|array Either displays a list of top topics or returns an array of info about them, depending on output_method.
 	 */
-	public static function topTopicsReplies($num_topics = 10, $output_method = 'echo')
+	public static function topTopicsReplies(int $num_topics = 10, string $output_method = 'echo'): array
```
tyrsson: This must be returned typed for ?array since it returns self::topTopics, which could return null

Sources/ServerSideIncludes.php
```diff
@@ -1057,7 +1084,7 @@ public static function topTopicsReplies($num_topics = 10, $output_method = 'echo
 	 * @param string $output_method The output method. If 'echo', displays a list of topics, otherwise returns an array of info about them
 	 * @return void|array Either displays a list of top topics or returns an array of info about them, depending on output_method.
 	 */
-	public static function topTopicsViews($num_topics = 10, $output_method = 'echo')
+	public static function topTopicsViews(int $num_topics = 10, string $output_method = 'echo'): array
```
tyrsson: Also must be ?array

Sources/ServerSideIncludes.php
```diff
@@ -1097,7 +1125,7 @@ public static function latestMember($output_method = 'echo')
 	 * @param string $output_method The output method. If 'echo', displays a link to the member's profile, otherwise returns an array of info about them.
 	 * @return void|array Displays a link to a random member's profile or returns an array of info about them depending on output_method.
 	 */
-	public static function randomMember($random_type = '', $output_method = 'echo')
+	public static function randomMember(string $random_type = '', string $output_method = 'echo'): array
```
tyrsson: must return ?array since queryMembers can possibly return null

Sources/ServerSideIncludes.php
```diff
@@ -1514,7 +1546,7 @@ public static function login($redirect_to = '', $output_method = 'echo')
 	 * @param string $output_method The output method. If 'echo', displays the poll, otherwise returns an array of info about it
 	 * @return void|array Either shows the top poll or returns an array of info about it, depending on output_method.
 	 */
-	public static function topPoll($output_method = 'echo')
+	public static function topPoll(string $output_method = 'echo'): array
```
tyrsson: return type should align with self::recentPoll

Sources/Url.php
```diff
@@ -370,9 +374,9 @@ public function validate(int $flags = 0): object
 	 * characters (a.k.a. IRIs)
 	 *
 	 * @param int $component Optional flag for parse_url's second parameter.
-	 * @return mixed Same as parse_url(), but with unmangled Unicode.
+	 * @return string|int|array Same as parse_url(), but with unmangled Unicode.
 	 */
-	public function parse(int $component = -1): mixed
+	public function parse(int $component = -1): string|int|array|null|false
```
tyrsson: This should end with |bool

Sources/Url.php
```diff
@@ -697,7 +701,7 @@ function ($line) {
 	 * @param int $component Optional flag for parse_url's second parameter.
 	 * @return mixed Same as parse_url(), but with unmangled Unicode.
 	 */
-	public static function parseIri(string $iri, int $component = -1): mixed
+	public static function parseIri(string $iri, int $component = -1): string|int|array|null|false
```
tyrsson: should end with |bool

Sources/Url.php
```diff
@@ -709,10 +713,10 @@ public static function parseIri(string $iri, int $component = -1): mixed
 	 *
 	 * @param string $iri The IRI to parse.
 	 * @param int $flags Optional flags for filter_var's third parameter.
-	 * @return object|false A reference to an object for the IRI if it is valid,
+	 * @return self|false A reference to an object for the IRI if it is valid,
 	 *    or false if the IRI is invalid.
 	 */
-	public static function validateIri(string $iri, int $flags = 0): object|false
+	public static function validateIri(string $iri, int $flags = 0): self|false
```
tyrsson: should end with |bool

Sources/Utils.php
```diff
@@ -1760,7 +1763,7 @@ public static function makeWritable(string $path): bool
 	/**
 	 * Emits a file for download. Mostly used for attachments.
 	 *
-	 * @param array|object $file Information about the file. Must be either an
+	 * @param array|\ArrayAccess $file Information about the file. Must be either an
```
tyrsson, line 1766: If we used traversable instead we could just type hint it to Interable.

tyrsson (approved):
Approved

PR #7989, opened 2023-12-30
Fix @see per the phpDocs docs

Oldiesmann:
#### Description
phpDoc complained that "\SMF\IntegrationHook::add" is not a valid FQSEN, so fix that.

### DIFF
--- a/Sources/IntegrationHook.php
+++ b/Sources/IntegrationHook.php
@@ -271,7 +271,7 @@ class IntegrationHook
 	 * Does nothing if the function is not available.
 	 * Cleans up enabled/disabled variants before taking requested action.
 	 *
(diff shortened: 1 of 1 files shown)


Sources/IntegrationHook.php
```diff
@@ -271,7 +271,7 @@ public static function add(string $name, string $function, bool $permanent = tru
 	 * Does nothing if the function is not available.
 	 * Cleans up enabled/disabled variants before taking requested action.
 	 *
-	 * @see IntegrationHook:add
+	 * @see \SMF\IntegrationHook\IntegrationHook:add
```
Sesquipedalian:
```suggestion
	 * @see IntegrationHook::add
```

Sesquipedalian (changes requested):

PR #7996, opened 2024-01-02
fix typos

RainRat:


jdarwood007:
As these are just minor typos in comments and the PR is against 2.1, I don't think we should merge these.  If this was made against 3.0, which is in alpha development, I wouldn't see an issue with it

live627:
Sorry @RainRat, but we are only accepting bug fixes to SMF 2.1. Typo fixes do not count as they do not affect functionality.

RainRat:
I will make a new change against 3.0

PR #8003, opened 2024-01-06
[3.0] Fix caching issue with objects

### LINKED ISSUE #7991: [3.0]: Caching fails on board index
### Basic Information

Turning on caching causes the board index to die with a fatal error (see logs section below).

### Steps to reproduce

1. Turn on caching
2. Try to view the board index

### Version/Git revision

3.0 Alpha 1

### Database Engine

All

### Database Version

_No response_

### PHP Version

_No response_

### Logs

```shell
Fatal error: Uncaught TypeError: strlen(): Argument #1 ($string) must be of type string, false given in /Users/jon/Sites/dev/Sources/Cache/CacheApi.php:558 Stack trace: #0 /Users/jon/Sites/dev/Sources/Cache/CacheApi.php(558): strlen(false) #1 /Users/jon/Sites/dev/Sources/Cache/CacheApi.php(512): SMF\Cache\CacheApi::put('boardindex-late...', Array, 60) #2 /Users/jon/Sites/dev/Sources/Actions/BoardIndex.php(92): SMF\Cache\CacheApi::quickGet('boardindex-late...', '', Array, Array) #3 /Users/jon/Sites/dev/Sources/Actions/BoardIndex.php(275): SMF\Actions\BoardIndex->execute() #4 [internal function]: SMF\Actions\BoardIndex::call() #5 /Users/jon/Sites/dev/Sources/Forum.php(260): call_user_func('SMF\\Actions\\Boa...') #6 /Users/jon/Sites/dev/index.php(144): SMF\Forum->execute() #7 {main} thrown in /Users/jon/Sites/dev/Sources/Cache/CacheApi.php on line 558
```

### Additional Information

This is the due to the caching system really not being designed for strict types.

jdarwood007:
Fixes #7991

4 things are being done to correct this:
1. use a native object in the FileBased setup (We could skip this, but why not)
2. The CacheApi::Get() now handles returning the $value if its a non string. 
3. The hits tracker needs to convert the $value to a string always.
4. Utils::jsonEncode and Utils::jsonDecode now align to PHP's native json_encode and json_decode.

### DIFF
--- a/Sources/Utils.php
+++ b/Sources/Utils.php
@@ -139,7 +139,7 @@ class Utils
 		'normalize' => __CLASS__ . '::normalize',
 		'truncate' => __CLASS__ . '::truncate',
 		'json_encode' => __CLASS__ . '::jsonEncode',
-		'json_decode' => __CLASS__ . '::jsonDecode',
+		'json_decode' => __CLASS__ . '::backcompat_smf_json_decode',
 		'random_int' => __CLASS__ . '::randomInt',
 		'random_bytes' => __CLASS__ . '::randomBytes',
 	];
@@ -1228,6 +1228,10 @@ class Utils
 	 * @param ?bool $associative Whether to force JSON objects to be returned as
 	 *    associative arrays. SMF nearly always wants this to be true, but for
 	 *    the sake of consistency with json_decode(), the default is null.
+	 * @param int $depth Maximum nesting depth of the structure being decoded.
+	 *    The value must be greater than 0, and less than or equal to 2147483647.
+	 * @param int $flags Bitmask of JSON_BIGINT_AS_STRING, JSON_INVALID_UTF8_IGNORE,
+	 *     JSON_INVALID_UTF8_SUBSTITUTE, JSON_OBJECT_AS_ARRAY, JSON_THROW_ON_ERROR
 	 * @param bool $should_log Whether to log errors. Default: true.
 	 * @return mixed The decoded data.
 	 */
@@ -2271,6 +2275,27 @@ class Utils
 		}
 
 		return $callable;
+	}
+
+	/**
+	 * Backward compatibilty wrapper for the smf_json_decode() method.
+	 *
+	 * @deprecated 3.0 - Only exists for $smcFunc backwards wrapper.
+	 * @param mixed $json The string to decode.
+	 * @param bool $associative Whether to force JSON objects to be returned as
+	 *    associative arrays. SMF nearly always wants this to be true, but for
+	 *    the sake of consistency with json_decode(), the default is false.
+	 * @param bool $should_log Whether to log errors. Default: true.
+	 * @return mixed The decoded data.
+	 */
+	public static function backcompat_smf_json_decode(mixed $json, bool $associative = false, bool $should_log = true): mixed
+	{
+		// In older versions, we accepted a mixed $json and would return if it was not a string.
+		if (empty($json) || !is_string($json)) {
+			return $json;
+		}
+
+		return Utils::jsonDecode($json, $associative, 512, 0, $should_log);
 	}
 
 	/*************************
--- a/Sources/Utils.php
+++ b/Sources/Utils.php
@@ -139,7 +139,7 @@ class Utils
 		'normalize' => __CLASS__ . '::normalize',
 		'truncate' => __CLASS__ . '::truncate',
 		'json_encode' => __CLASS__ . '::jsonEncode',
-		'json_decode' => __CLASS__ . '::backcompat_smf_json_decode',
+		'json_decode' => 'smf_json_decode',
 		'random_int' => __CLASS__ . '::randomInt',
 		'random_bytes' => __CLASS__ . '::randomBytes',
 	];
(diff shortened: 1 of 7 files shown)


Sources/Utils.php
```diff
@@ -1255,14 +1255,14 @@ function ($value, $key) use (&$length) {
...
-	 * @param string $json The string to decode.
+	 * @param mixed $json The string to decode.
 	 * @param bool $associative Whether to force JSON objects to be returned as
 	 *    associative arrays. SMF nearly always wants this to be true, but for
 	 *    the sake of consistency with json_decode(), the default is false.
 	 * @param bool $should_log Whether to log errors. Default: true.
 	 * @return mixed The decoded data.
 	 */
-	public static function jsonDecode(string $json, bool $associative = false, bool $should_log = true)
+	public static function jsonDecode(mixed $json, bool $associative = false, bool $should_log = true)
```
Sesquipedalian: When would this ever be something besides a string, and why would we want to accept a non-string value here?

jdarwood007: It was happening on the board index. a int/bool was getting in there.  Best to let it just realize it doesn't need to decode those and return.

Sesquipedalian: It would be better to cast the value to string before passing it in. Utils::jsonDecode() is supposed to accept the same input (for the first two parameters) as json_decode().

jdarwood007: I agree. I didn't add the is_string checks, it was already in there. So we have been already accepting it like this. We would have to adapt the logic further and not send it to this function if it's not a string. If it's an int/bool/float, et don't need to do anything.

jdarwood007: Also to add:
>  Returns the value encoded in json as an appropriate PHP type. Unquoted values true, false and null are returned as true, false and null respectively. null is returned if the json cannot be decoded or if the encoded data is deeper than the nesting limit. 

We should make jsonDecode and then return null instead of [] on error if we are wanting to match specs here.
I can align that.  Maybe best to do that in a separate PR because its possible it will break things.

What do you think/want?
I have no problem with moving things back into the get to check for non strings and skipping json decode in that case.

Sesquipedalian: > We should make jsonDecode and then return null instead of [] on error if we are wanting to match specs here. I can align that. Maybe best to do that in a separate PR because its possible it will break things.
> 
> What do you think/want? I have no problem with moving things back into the get to check for non strings and skipping json decode in that case.

That sounds like an excellent plan. Please do.

jdarwood007: I aligned it.  I also aligned jsonEncode (the best I could since the return false type isn't supported in 8.0).

To align it, I had to update calls to jsonDecode so I could add the parameters required.  I added a wrapper for the legacy $smcFunc calls.

Sources/Utils.php
```diff
@@ -2264,6 +2277,27 @@ public static function getCallable(mixed $input, ?bool $ignore_errors = null): m
...
+	public static function backcompat_smf_json_decode(mixed $json, bool $associative = false, bool $should_log = true): mixed
+	{
+		// In older versions, we accepted a mixed $json and would return if it was not a string.
+		if (empty($json) || !is_string($json)) {
+			return $json;
+		}
+
+		return Utils::jsonDecode($json, $associative, 512, 0, $should_log);
+	}
+
```
Sesquipedalian: It would be better to paste the contents of this method into the `smf_json_decode()` function in Subs-Compat.php, and then have `$smcFunc['json_decode']` point to that.

Sesquipedalian (changes requested):

Sources/Utils.php
jdarwood007: I think It is correct now on how we would expect it to work.

Interesting that `$smcFunc['json_decode']` was mapping to `Utils::jsonDecode` instead of the `smf_json_decode`.

jdarwood007:
Unrelated to the PR (kinda).  But if we don't like the new additional parameters for jsonDecode.  PHP 8.0 added support for [named parameters](https://php.watch/versions/8.0/named-parameters).  So you could call it like:
```
Utils::jsonDecode(
    json: '{"hello":"world"}',
    should_log: true
);
```

jdarwood007:
Also, while the depth is 512.  A lot of them don't need that high.  The cache one returns a single level (cached value is still encoded).  The cookie one as well.  So if we want to optimize/maybe call it security, we could lower those to a depth of 2.  I don't even think our cookies have any nested arrays.

Sources/Utils.php
Sesquipedalian: > Interesting that `$smcFunc['json_decode']` was mapping to `Utils::jsonDecode` instead of the `smf_json_decode`.

That's due to the history and evolution of the code. In my original OOP refactoring, `$smcFunc['json_decode']` and `smf_json_decode` were both just aliases of `Utils::jsonDecode` that were `eval`'d into existence at runtime. But in @Tyrsson's refactoring of BackwardCompatibilty, `smf_json_decode` became a real function again. As long as the signatures of all three were the same, it made sense for `$smcFunc['json_decode']` to continue to be a direct alias of `Utils::jsonDecode` rather than detouring through `smf_json_decode`. But now that the signature of `Utils::jsonDecode` has changed, making `$smcFunc['json_decode']` an alias of `smf_json_decode` makes the most sense.

Sesquipedalian (approved):

Sesquipedalian:
> Also, while the depth is 512.  A lot of them don't need that high.

It's fine.

PR #7933, opened 2023-12-04
[3.0] Theme

live627:
This is the new theme started by @twitchismental

I still need to convert several form grids to use CSS grids instead of floats. Also the grids abuse definition list markup,  something tat has always annoyed me from the very beginning.

Submitting as a draft now to get early feedback.

PR #8004, opened 2024-01-06
[3.0] Add an option for cdnjs to jQuery CDNs

Oldiesmann:
This PR adds built-in support for loading jQuery from Cloudflare's CDN (aka "cdnjs"). While it doesn't add any additional benefits, it gives us another solid option in addition to the ones provided.


DiegoAndresCortes:
We implemented cloudflare cdn in #7933 when adding fontawesome, we can add it to jquery there

jdarwood007:
@DiegoAndresCortes, would you prefer yours or Oldies then?  If you want yours, closeout and merge the theme one later.

At another layer.  I think we need to abstract the CDN logic.  So it's easy to add more providers in the future or remove them.  As well as identify which files they can serve, then allow passing or grabbing which 'version' they should be using as defined by our software.  A "local" provider is always the SMF install.

tyrsson:
Personally, I think it would be best to push this to alpha 4 when the other template changes are targeted.

Oldiesmann:
The only theme-related files this touches are a couple of language files. It doesn't modify the templates at all.

DiegoAndresCortes:
Both PR's need to wait for #7853 anyways so will just have to wait and decide later.

jdarwood007:
@Oldiesmann This has a conflict that needs resolved.

Oldiesmann:
> @Oldiesmann This has a conflict that needs resolved.

Should be fixed now.

PR #8005, opened 2024-01-06
[3.0] Fix Group Membership Requests

### LINKED ISSUE #7949: [3.0] Cannot join group from Group Membership section of profile
#### Description
There are two different but related bugs here. If you create a group that someone can request to join (or one that they can join/leave at will), you can't join it from the Group Membership section of the profile, even if you're an admin.

There are two bugs here:
* This code isn't checking whether the person can join/request to join the group before hitting the permission check:
https://github.com/SimpleMachines/SMF/blob/de43b89cc5ac90c7b000e6ae1fc4f29bcb403b42/Sources/Actions/Profile/GroupMembership.php#L175C1-L178C4
```php
        // Let's be extra cautious...
        if (!User::$me->is_owner || empty(Config::$modSettings['show_group_membership'])) {
            User::$me->isAllowedTo('manage_membergroups');
        }
```

* The "is_admin" check in allowedTo() is failing for admins when the above permission check is run (as I said, if I go to my profile as an admin, then go to the Group Membership section and try to join a group, I'm still told I can't modify or manage membergroups.

### Steps to reproduce
1. Create a new group and set its type to either "Requestable" or "Free"
2. Go to the group membership page in your profile (click username -> My Groups) and try to join the newly created group, either as a regular user or as an admin.

- Database Type: MariaDB
- Database Version: 11
- PHP Version: 8.3

### LINKED ISSUE #7950: [3.0] Can't add/remove group members from Mod Center
#### Description
Discovered this one while trying to "demote" a second test account on my test board in order to test the issue reported in #7949 (couldn't do it from the profile due to yet another bug; I submitted #7948 to fix that particular issue). When you choose which member(s) to remove from a group, you are prompted for your password, but the token from the groups form isn't carried over through the login process (or a new token is generated in the process), so you get a "token verification failed" error instead

### Steps to reproduce
1. Go to Moderate -> Groups -> View Membergroups
2. Either check the box next to one or more members and click the "remove from group" button or try to add a new member to the group
3. Enter your password when prompted to do so

- Database Type: MariaDB
- Database Version: 11
- PHP Version: 8.3

This happens regardless of whether moderation security is enabled.

jdarwood007:
Fixes #7949
Fixes #7950

@Oldiesmann Please test this
@Sesquipedalian I don't like how the fix turned out, but I didn't see a better fix.

The major issue stems from a concept in Sources\Group.php, the `load()`.
In various places in our group logic, we assumed `load()` would return
- An array of IDs for the group
- The keys would be the group ID
- That when calling load with only one ID, we could do a list to pop off the first group.

Neither case was true.  We returned an array of objects of SMF\Group.  The key was not the group ID, but the next ID, which makes list() work, but breaks other logic.
Some of the logic in `load()` assumed that `$loaded[$id_group]`, yet then assigned it to the array with `[]`.  Only lines apart. 
I aligned it with how SMF\User does it just for uniformity.  But because of this change, it could have far-reaching effects.

The second major issue was that when you do array_merge, you lose key associations.  So I had to use `+=` logic. This means that if any of the same keys exist on both sides, the right side of the logic is lost.

A third issue was that the `addMembers` logic was silently failing and SMF just happily assumed it worked.  I added a check but didn't add a better error message right now.  If we need one, we can.

A fourth issue is that we assumed `Db::$db->fetch_all()` would return an array of the ids in such fashion `[1,2,3,5]`, when in fact it returned them as `[0 => ['id_group' = 1], 1 => ['id_group' = 2']....]`.  A simple array map fixes this.

In many places, because we assumed that `load()` is returning IDs, not objects, we incorrectly assumed we could just use it like that.  However, that is causing the issues.  I added hints that should help most IDEs recognize that we are working with an object, not an int/array.

### DIFF
--- a/Sources/Actions/Profile/GroupMembership.php
+++ b/Sources/Actions/Profile/GroupMembership.php
@@ -106,8 +106,8 @@ class GroupMembership implements ActionInterface
 		$open_requests = Db::$db->fetch_all($request);
 		Db::$db->free_result($request);
 		
-		$open_requests = array_map(function($v) {
-			return (int) $v['id_group'];
+		$open_requests = array_map(function($request) {
+			return (int) $request['id_group'];
 		}, $open_requests);
 
 		// Show the assignable groups in the templates.
--- a/Sources/Actions/Profile/GroupMembership.php
+++ b/Sources/Actions/Profile/GroupMembership.php
@@ -105,6 +105,10 @@ class GroupMembership implements ActionInterface
 		);
 		$open_requests = Db::$db->fetch_all($request);
 		Db::$db->free_result($request);
+		
+		$open_requests = array_map(function($v) {
+			return (int) $v['id_group'];
+		}, $open_requests);
 
 		// Show the assignable groups in the templates.
 		foreach (Profile::$member->current_and_assignable_groups as $id => $group) {
@@ -114,17 +118,17 @@ class GroupMembership implements ActionInterface
 			}
 
 			// Are they in this group?
-			$member_or_available = in_array($id, Profile::$member->groups) ? 'member' : 'available';
+			$member_or_available = in_array($group->id, Profile::$member->groups) ? 'member' : 'available';
 
 			// Can't join private or protected groups.
 			if ($group->type < Group::TYPE_REQUESTABLE && $member_or_available == 'available') {
 				continue;
 			}
 
-			Utils::$context['groups'][$member_or_available][$id] = $group;
+			Utils::$context['groups'][$member_or_available][$group->id] = $group;
 
 			// Do they have a pending request to join this group?
-			Utils::$context['groups'][$member_or_available][$id]->pending = in_array($id, $open_requests);
+			Utils::$context['groups'][$member_or_available][$group->id]->pending = in_array($group->id, $open_requests);
 		}
 
 		// If needed, add "Regular Members" on the end.
--- a/Sources/Group.php
+++ b/Sources/Group.php
@@ -1017,20 +1017,24 @@ class Group implements \ArrayAccess
 		// new additional group.
 		$set_primary = [];
 		$set_additional = [];
+		$member_ids = [];
 
+		/** @var \SMF\User $member */
 		foreach ($members as $key => $member) {
+			$member_ids = $member->id;
+
 			// Forcing primary.
 			if ($type === 'force_primary') {
-				if (User::$loaded[$member]->group_id !== $this->id) {
+				if (User::$loaded[$member->id]->group_id !== $this->id) {
 					$set_primary[] = $member;
 				}
 			}
 			// They're already in this group.
-			elseif (in_array($this->id, User::$loaded[$member]->groups)) {
+			elseif (in_array($this->id, User::$loaded[$member->id]->groups)) {
 				continue;
 			}
 			// They have a different primary group.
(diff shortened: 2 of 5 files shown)


Oldiesmann:
Works well for the most part. I was able to join/leave a free group (both as an admin and as a regular user) and add someone to the group via the mod center (without being prompted for my password since mod security is off).

However, I did find a couple more bugs. I'm not sure if either of these is related to your changes but I'll mention them here anyway.

Trying to remove someone from a group via the groups interface (Admin -> Members -> Membergroups, click group name, check box next to a member and click the button to remove them):
` Uncaught TypeError: Cannot access offset of type SMF\\User on array in /.../Sources/Group.php:1226`

When you go to edit a group (Admin -> Members -> Membergroups, click "Modify" next to a group), you now get an error saying "The membergroup does not exist or is invalid".

I was unable to test requestable groups (eg a group where someone has to approve your request to join) since I didn't have any and am unable to edit any new groups I create.

Oldiesmann:
Everything seems to be working now. The second bug I mentioned above seems to be due to me not copying the changes to Sources/Actions/Admin/Membergroup.php. I've been able to join/leave both a free and a requestable group (and approve a request to join the requestable group), as well as add and remove group members from the admin center.

jdarwood007:
Can another dev merge this? I can't test the functionality of group requests with strict typing because of this.

Sources/Group.php
```diff
@@ -1077,15 +1081,16 @@ public function addMembers(int|array $members, string $type = 'auto', bool $perm
...
-				$new_additional_groups = array_diff(User::$loaded[$member]->groups, [$this->id, User::$loaded[$member]->post_group_id]);
+				$new_additional_groups = array_diff(User::$loaded[$member->id]->groups, [$this->id, User::$loaded[$member->id]->post_group_id]);
 				sort($new_additional_groups);
 
 				$to_set[implode(',', $new_additional_groups)][] = $member->id;
 			}
 
-			foreach ($to_set as $new_additional_groups => $member_ids) {
-				User::updateMemberData($member_ids, [
+			foreach ($to_set as $new_additional_groups => $mems) {
```
tyrsson: Not to be picky here but can we get a more descriptive $var here, such as $mem_ids and $mem_id. or $ids and $id. Anything to communicate what its actually using.

Sources/Actions/Profile/GroupMembership.php
```diff
@@ -120,6 +120,10 @@ public function show(): void
 		);
 		$open_requests = Db::$db->fetch_all($request);
 		Db::$db->free_result($request);
+		
+		$open_requests = array_map(function($v) {
```
tyrsson: I could be blind and if so please forgive me. It appears $v is undefined here?

jdarwood007: $v will be the value of the array.  Its a callback.

Sources/Group.php
jdarwood007: I will put it back to `$member_ids`.  It is an array of ints containing the `id_member` to be processed.

Sources/Actions/Profile/GroupMembership.php
jdarwood007: Changed it to $request so its more clear?

tyrsson (approved):

PR #8012, opened 2024-01-12
Unicode background task error

jdarwood007:
Fixes a error with the unicode background task failing because $func_name is sometimes an int.


tyrsson (approved):
Approved. Have run the related code through xdebug prior to approval and merging.

PR #8020, opened 2024-01-14
[2.1] Fix an issue where ban name length is not checked

### LINKED ISSUE #8019: [2.1.4]: Renaming a ban fails with Database Error if name is too long
### Basic Information

Trying to rename a ban will fail with an opaque Database error if the name is too long, even though the input field itself has no length limit. Reading the admin error logs reveals the problem: `Database Error: Data too long for column 'name' at row 1` (source location: ManageBans.php, line 1564)

### Steps to reproduce

1. Edit an existing ban
2. Rename the ban name to something moderately lengthy
3. Observe Database Error.

### Expected result

The input field should be limited to maximum permitted length.

### Actual result

_No response_

### Version/Git revision

2.1.4

### Database Engine

MySQL

### Database Version

_No response_

### PHP Version

_No response_

### Logs

_No response_

### Additional Information

_No response_

dragomano:
Fixes #8019 
Closes #8019


Oldiesmann:
FYI, you don't need both "fixes" and "closes" as they both do the same thing

dragomano:
Just in case :)

Sesquipedalian (approved):

Themes/default/ManageBans.template.php
```diff
@@ -184,7 +184,7 @@ function template_ban_edit()
 			</div><!-- .windowbg -->
 		</form>';
 
-	if (!$context['ban']['is_new'] && empty($context['ban_suggestions']))
+	if (!$context['ban']['is_new'] && empty($context['ban_suggestions']) && !empty($context['ban_list']))
```
jdarwood007, line 187: This change is causing a problem seen in #8746

LexArma: Reverthing the change in this one line fixed #8746 for me, and I feel the last !empty is a mistake, and should not be there at all - The list is already equipped to handle a situation where no triggers exist. I tested, a ban with no triggers simply states "No ban triggers specified", and allows for adding new triggers.

PR #8021, opened 2024-01-15
[3.0] Strict Type for actions

### LINKED ISSUE #8026: [3.0]: htmlspecialcharsRecursive receives integers
### Basic Information

In certain places, htmlspecialcharsRecursive and htmlspecialchars receive integers.
For example, in the registration process, the following data is sent as a unix timestamp:

https://github.com/SimpleMachines/SMF/blob/fc6b11d9591122d5a2d0d02b2c2cf96668106c96/Sources/Actions/Register2.php#L301-L308

There might be more places.
It seems it always has received integers too, either intentionally or by accident.
Should it skip integers, convert them, or keep accepting them?

### Steps to reproduce

1. Attempt to register a new account
2. `Uncaught TypeError: SMF\Utils::htmlspecialcharsRecursive(): Argument #1 ($var) must be of type array|string, int given`

### Expected result

_No response_

### Actual result

_No response_

### Version/Git revision

3.0 Alpha 1

### Database Engine

MySQL

### Database Version

_No response_

### PHP Version

8.1

### Logs

_No response_

### Additional Information

_No response_

### LINKED ISSUE #8027: [3.0]: Alert::fetch() $to_fetch is receiving integers
### Basic Information

When fetching a specific alert:
https://github.com/SimpleMachines/SMF/blob/fc6b11d9591122d5a2d0d02b2c2cf96668106c96/Sources/Actions/Profile/ShowAlerts.php#L57

It sends the alert id. But fetch is expecting a boolean or an array in $to_fetch and receives neither.
This will cause for Alert:loadForMember to fail in this check:
https://github.com/SimpleMachines/SMF/blob/fc6b11d9591122d5a2d0d02b2c2cf96668106c96/Sources/Alert.php#L846

Thus returning the entire list of alerts and always grabbing the oldest one when you go to ?action=profile;area=showalerts;alert=2;

### Steps to reproduce

1. Receive some alerts
2. Click on any alert, all of them go to the oldest one's _target_href_

### Expected result

It should redirect to the correct alert.

### Actual result

It redirects to the wrong alert.

### Version/Git revision

3.0 Alpha 1

### Database Engine

MySQL

### Database Version

_No response_

### PHP Version

8.1

### Logs

_No response_

### Additional Information

Two possible solutions (for me anyways).

- Cast the single alert where this happens, in an array:
`$alerts = Alert::fetch(User::$me->id, [$alert_id]);`

- Add integer to the types of $to_fetch for both Alert::fetch and Alert:loadForMember

### LINKED ISSUE #8028: [3.0] Backards compat strict type error
@Tyrsson 

Sources/Board.php
```
public static $info;
```

Set the type to self.
Refresh the forum

```
( ! ) Fatal error: Uncaught Error: Cannot access uninitialized non-nullable property SMF\Board::$info by reference in /Sources/BackwardCompatibility.php on line 67
```

We may have this issue with other areas.  I suspect we will have to make these variables nullable.

jdarwood007:
Work In Progress

Bug Fixes:
- Various calls to `ErrorHandler::fatalLang()` 3rd param where strings, not array as expected.
- Actions\Admin\ErrorLog: `self::LOAD_BY_ID` instead of `User::LOAD_BY_ID`.
- Actions\Admin\Reports: `addSeparator()` called `$table` instead of `$$custom_table`.
- Actions\Admin\Server: `cookie()` incorrectly called Url logic.
- Actions\Admin\Server: `export()` `$diskspace_disabled` was undefined.
- Actions\Admin\Subscriptions: `add()` made an incorrect call to `getSubs()`.
- Actions\Admin\Warnings: `$currently_enabled` was incorrectly set.
- Actions\Admin\Registration: Fix the error if you have never set a privacy policy.
- Actions\Moderation\Warnings: `templates()` made an incorrect call to `ModifyWarningTemplate`
- Actions\ Moderation\Watched Users: `execute()` made a incorrect call to `list_getWatchedUserPostsCount()`
- Actions\Profile\Export: `execute()` called `$current_datatype` which is never defiend.
- Actions\Profile\GroupMembership `groupMembership2()` didn't return anything.
- Actions\Profile\Profile Fix some logic with avatars.
- Actions\Profile\TrackIP Used wrong calls for the tracking list
- Actions\Activate Incorrect call to `ErrorHandler::fatalLang` last param.
- Actions\Calendar `export` tried to do a strlen on an array.
- Actions\CoppaForm Used `count` instead of `strlen`
- Actions\Feed `setMember()` method had unreacable code.
- Actions\Help Used `$subactions` instead of `self::$subactions`
- Actions\Login2 Incorrectly called `Cookie::setTFACookie` 2nd param with the user's salt rather than id.
- Actions\Login2 `phpBB3_password_check()` needed parameters rather than `$_POST`
- Actions\Notify `setAlertPref()` used incorrect logic for `self::MODE_NO_EMAIL` case
- Actions\Post Made an incorrect call to `ErrorHandler::fatalLang` status param.
- Actions\Post2 `submit()` made incorrect call to `Utils::htmlTrim`
- Actions\Search2 Incorrect call for `SearchRequest::$message_request`
- Actions\Search2 Removed spell check calls.
- Actions\Profile `Main()` didn't check if a profile existed before continuing after `load`
- /subscriptions.php made call to `Subscriptions::getAll()` that doesn't exist, use `Subscriptions::getSubs()`

Resolved bugs:
Fixes #8026
Fixes #8027
Fixes  #8028


Sources/Actions/Profile/ShowAlerts.php
```diff
@@ -70,7 +72,7 @@ public function execute(): void
 			}
 
 			// Mark the alert as read while we're at it.
-			Alert::mark(User::$me->id, $alert_id, 1);
+			Alert::mark(User::$me->id, $alert_id, true);
```
jdarwood007, line 75: See #8027

Sources/Utils.php
```diff
@@ -440,18 +441,18 @@ public static function htmlspecialchars(string $string, int $flags = ENT_COMPAT,
...
 	 *
-	 * @param array|string $var The string or array of strings to add entities to
+	 * @param mixed $var The string or array of strings to add entities to
 	 * @param int $flags Bitmask of flags to pass to standard htmlspecialchars().
 	 *    Default is ENT_COMPAT.
 	 * @param string $encoding Character encoding. Default is UTF-8.
 	 * @return array|string The string or array of strings with entities added
 	 */
-	public static function htmlspecialcharsRecursive(array|string $var, int $flags = ENT_COMPAT, string $encoding = 'UTF-8'): array|string
+	public static function htmlspecialcharsRecursive(mixed $var, int $flags = ENT_COMPAT, string $encoding = 'UTF-8'): array|string
```
jdarwood007, line 450: See #8026

tyrsson: This should not be expanded to accept mixed. The calling code should pass either an array or string. The underlying function is typed for string. If we are sending anything other than
string|array<array-key, string> 
then the calling code needs refactored.

jdarwood007: It's receiving ints/float/other basic types as well.  Seems easier to let it just return those (we don't need to handle it).

tyrsson: In this case I can see that. Lets just not use it as a crutch to not clean up the related calling code as needed when we can.

jdarwood007: Affirmative.  It was easier to use mixed than define a long list of types.  I believe only strings need cleaning through `htmlspecialchars`, but since we accept an array, we may have other types being processed in it due to recursion.  If anyone knows of other types that also need cleansing, let me know so we can do those.

Sources/Board.php
```diff
@@ -326,14 +326,14 @@ class Board implements \ArrayAccess
...
+	public static ?int $board_id;
 
 	/**
-	 * @var object
+	 * @var self
 	 *
 	 * Instance of this class for board we are currently in.
 	 */
-	public static $info;
+	public static ?self $info;
```
jdarwood007, line 336: #8028

tyrsson (approved):

jdarwood007:
This should be ready.  I tried to do every action I could think of, as guest, user, moderator, and admin.  But I'm certain I missed stuff because not all the right conditions were set.  When reviewing, do your best to test things.  But we may have to accept it will be merged and more issues found.

The good part is with this, we should be correctly always sending the right types to functions, so PHP is not juggling types.  Helps us know our expected input and outputs.  It also helps static analysis tools know more about what is going on. Another phase of this would be to use objects for returns, so we can get full-type hinting, instead of arrays.

Issue #8023, opened 2024-01-16
[3.0]: Avatar changing


jdarwood007:
### Basic Information

When trying to use a SVG for a avatar, a fatal PHP error occurs and the avatar is not saved

### Steps to reproduce

1. Go to Forum Profile
2. Choose an external image
3. Enter: https://www.php.net/images/logos/php-logo-white.svg

### Expected result

Forum Profile is saved and avatar changed

### Actual result

Fatal PHP error

### Version/Git revision

3054fc2

### Database Engine

All

### Database Version

N/A

### PHP Version

8.1.24

### Additional Information

@Sesquipedalian I'm seeing multiple avatar-related issues.  Not just this one, but its the one I'm not certain on how we intend to fix, since we are passing things through Image handler and then trying to use $image->type, which is not defined.

jdarwood007:
Changing the title here.  I found a few more problems

- Can not use a webp image.
- When you do set an image, the preview seems to be 'sticky' for that page load and does not refresh until the next page load
- The avatar does not detect which type was set correctly.

frandominguezl:
I couldn't reproduce with the latest build

SVG test:
![imagen](https://github.com/SimpleMachines/SMF/assets/10365222/9fb452b2-cbb3-4f98-8814-7abe4dd57cfb)

Webp test:
![imagen](https://github.com/SimpleMachines/SMF/assets/10365222/19ea0279-d646-48a9-8c32-258e0aaa9435)
_Test image: https://tmpr.s3.eu-central-1.amazonaws.com/assets/imgs/webp-test.webp_

DiegoAndresCortes:
Maybe #8226 fixed this too?

Sesquipedalian:
> Maybe #8226 fixed this too?

Maybe. At any rate, it seems that we can close this.

PR #8033, opened 2024-01-21
Fix type mismatch when checking for gravatar

Oldiesmann:
#### Description
The various parts of code checking for gravatar expect `$data['avatar']` to be a string, but it might be a Url object instead. This casts it to a string before we start working with it. The code that expects it to be a Url object will convert it (back) to one if needed.

### DIFF
--- a/Sources/User.php
+++ b/Sources/User.php
@@ -2816,6 +2816,9 @@ class User implements \ArrayAccess
 		// Set a nice default var.
 		$image = '';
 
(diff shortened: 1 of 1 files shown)


tyrsson (approved):

jdarwood007:
Avatars have a lot of problems, but I haven't dug into it since I had strict typing PR going on.  See #8023 
Will merge this for now, we can continue and get it fixed later.  This may even fix things since lots of the problems revolve around the differences in URL object and strings.

PR #8052, opened 2024-01-28
[3.0] Remove tasksdir configuration option

jdarwood007:
We no longer need this nor should we have it.  All other sub directories in Sources do not have their own option to move the Tasks directory, so it doesn't make sene to allow this

### DIFF
--- a/Sources/Config.php
+++ b/Sources/Config.php
@@ -792,7 +792,7 @@ class Config
 				if (!is_dir(realpath($languagesdir)) && is_dir($boarddir . '/Languages'))
 					$languagesdir = $boarddir . '/Languages';
 				END,
-			'search_pattern' => '~\n?(#[^\n]+)?(?:\n\h*if\s*\((?:\!file_exists\(\$(?' . '>boarddir|sourcedir|packagesdir|cachedir|languagesdir)\)|\!is_dir\(realpath\(\$(?' . '>boarddir|sourcedir|packagesdir|cachedir|languagesdir)\)\))[^;]+\n\h*\$(?' . '>boarddir|sourcedir|packagesdir|cachedir|languagesdir)[^\n]+;)+~sm',
+			'search_pattern' => '~\n?(#[^\n]+)?(?:\n\h*if\s*\((?:\!file_exists\(\$(?' . '>boarddir|sourcedir|tasksdir|packagesdir|cachedir|languagesdir)\)|\!is_dir\(realpath\(\$(?' . '>boarddir|sourcedir|tasksdir|packagesdir|cachedir|languagesdir)\)\))[^;]+\n\h*\$(?' . '>boarddir|sourcedir|tasksdir|packagesdir|cachedir|languagesdir)[^\n]+;)+~sm',
 		],
 		'db_character_set' => [
 			'text' => <<<'END'
--- a/Sources/Config.php
+++ b/Sources/Config.php
@@ -777,18 +777,6 @@ class Config
 			'raw_default' => true,
 			'type' => 'string',
 		],
-		'tasksdir' => [
-			'text' => <<<'END'
-				/**
-				 * @var string
-				 *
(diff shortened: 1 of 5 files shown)


other/upgrade.php
```diff
@@ -1918,36 +1913,6 @@ function DeleteUpgrade()
...
-		}
-
-		// Rename 'tasks' to 'Tasks'.
-		// Do this in two steps to make sure it works on case insensitive file systems.
-		rename(Config::$tasksdir, dirname(Config::$tasksdir) . DIRECTORY_SEPARATOR . 'Tasks_temp');
-		rename(dirname(Config::$tasksdir) . DIRECTORY_SEPARATOR . 'Tasks_temp', dirname(Config::$tasksdir) . DIRECTORY_SEPARATOR . 'Tasks');
-
-		$changes['tasksdir'] = dirname(Config::$tasksdir) . '/Tasks';
-	}
-
```
Sesquipedalian: This all needs to stay except for the final `$changes['tasksdir'] = dirname(Config::$tasksdir) . '/Tasks';` line. It's here because we renamed `Sources/tasks` to `Sources/Tasks`.

Sources/Config.php
```diff
@@ -779,14 +767,12 @@ class Config
...
 					$sourcedir = $boarddir . '/Sources';
-				if (!is_dir(realpath($tasksdir)) && is_dir($sourcedir . '/Tasks'))
-					$tasksdir = $sourcedir . '/Tasks';
 				if (!is_dir(realpath($packagesdir)) && is_dir($boarddir . '/Packages'))
 					$packagesdir = $boarddir . '/Packages';
 				if (!is_dir(realpath($cachedir)) && is_dir($boarddir . '/cache'))
 					$cachedir = $boarddir . '/cache';
 				END,
-			'search_pattern' => '~\n?(#[^\n]+)?(?:\n\h*if\s*\((?:\!file_exists\(\$(?' . '>boarddir|sourcedir|tasksdir|packagesdir|cachedir)\)|\!is_dir\(realpath\(\$(?' . '>boarddir|sourcedir|tasksdir|packagesdir|cachedir)\)\))[^;]+\n\h*\$(?' . '>boarddir|sourcedir|tasksdir|packagesdir|cachedir)[^\n]+;)+~sm',
+			'search_pattern' => '~\n?(#[^\n]+)?(?:\n\h*if\s*\((?:\!file_exists\(\$(?' . '>boarddir|sourcedir|packagesdir|cachedir)\)|\!is_dir\(realpath\(\$(?' . '>boarddir|sourcedir|packagesdir|cachedir)\)\))[^;]+\n\h*\$(?' . '>boarddir|sourcedir|packagesdir|cachedir)[^\n]+;)+~sm',
```
Sesquipedalian: `tasksdir` needs to be remain part of this regex, even though the relevant lines of code have been removed from the code block's definition. Otherwise, Config::updateSettingsFile() will break when trying to update an existing Settings.php file that contains those lines.

Sesquipedalian (commented):
In addition to the issues mentioned in my review comments, this PR also needs some code to take care of removing the $tasksdir setting from existing Settings.php files. However, I have written some code that I can add to this PR to take care of that part. So once conflicts have been resolved and the issues I commented on have been fixed, I will add that code to this PR and then merge.

other/upgrade.php
jdarwood007: Could you help me understand why it is still needed? I removed the entire option.  It's now hard coded (for backward compatibility reasons) to `$sourcedir . '/Tasks'`

Sesquipedalian: Because we need to rename the directory itself on disk.

Sources/Config.php
jdarwood007: Your right.  Reversed.  Conflicts were also removed.

jdarwood007:
Looks good.  Don't see any issues with those changes.

PR #8053, opened 2024-01-28
[3.0] Introduce Sapi Class

jdarwood007:
Move some logic from both Config and Utils that fit better here. This should be considered low-level code and should handle operating with the code in a very error-safe way.

It additionally, fixed some warnings from static analysis.

### DIFF
--- a/Sources/Actions/Admin/ACP.php
+++ b/Sources/Actions/Admin/ACP.php
@@ -1615,7 +1615,6 @@ class ACP implements ActionInterface
 			$directories += ['template_versions' => Theme::$current->settings['theme_dir']];
 		}
 
-		/** @var string $dirname */
 		foreach ($directories as $type => $dirname) {
 			$this_dir = dir($dirname);
 
--- a/Sources/Actions/Admin/ACP.php
+++ b/Sources/Actions/Admin/ACP.php
@@ -1615,6 +1615,7 @@ class ACP implements ActionInterface
 			$directories += ['template_versions' => Theme::$current->settings['theme_dir']];
 		}
 
+		/** @var string $dirname */
 		foreach ($directories as $type => $dirname) {
 			$this_dir = dir($dirname);
 
--- a/Sources/Actions/Admin/Find.php
+++ b/Sources/Actions/Admin/Find.php
@@ -25,6 +25,7 @@ use SMF\PackageManager\XmlArray;
 use SMF\User;
 use SMF\Utils;
 use SMF\WebFetch\WebFetchApi;
+use SMF\Sapi;
 
 /**
  * Provides the search functionality inside the admin control panel.
@@ -194,7 +195,7 @@ class Find implements ActionInterface
 	public function internal(): void
 	{
 		// Try to get some more memory.
-		Config::setMemoryLimit('128M');
+		sapi::setMemoryLimit('128M');
 
 		IntegrationHook::call('integrate_admin_search', [&$this->language_files, &$this->include_files, &$this->settings_search]);
 
--- a/Sources/Actions/Admin/Find.php
+++ b/Sources/Actions/Admin/Find.php
@@ -195,7 +195,7 @@ class Find implements ActionInterface
 	public function internal(): void
 	{
 		// Try to get some more memory.
-		sapi::setMemoryLimit('128M');
+		Sapi::setMemoryLimit('128M');
 
 		IntegrationHook::call('integrate_admin_search', [&$this->language_files, &$this->include_files, &$this->settings_search]);
 
--- a/Sources/Actions/Admin/Permissions.php
+++ b/Sources/Actions/Admin/Permissions.php
@@ -3258,6 +3258,7 @@ class Permissions implements ActionInterface
 
 		foreach (self::getPermissions() as $permission => $perm_info) {
 			// If this permission shouldn't be given to certain groups (e.g. guests), don't.
+			/** @var string $perm */
 			foreach ([$permission, $perm_info['generic_name']] as $perm) {
 				if (isset(Utils::$context['group']['id']) && in_array(Utils::$context['group']['id'], self::$excluded[$perm] ?? [])) {
 					continue 2;
--- a/Sources/Actions/Admin/Permissions.php
+++ b/Sources/Actions/Admin/Permissions.php
@@ -3258,7 +3258,6 @@ class Permissions implements ActionInterface
 
 		foreach (self::getPermissions() as $permission => $perm_info) {
 			// If this permission shouldn't be given to certain groups (e.g. guests), don't.
-			/** @var string $perm */
 			foreach ([$permission, $perm_info['generic_name']] as $perm) {
 				if (isset(Utils::$context['group']['id']) && in_array(Utils::$context['group']['id'], self::$excluded[$perm] ?? [])) {
 					continue 2;
--- a/Sources/Actions/Feed.php
+++ b/Sources/Actions/Feed.php
@@ -26,6 +26,7 @@ use SMF\ErrorHandler;
 use SMF\IntegrationHook;
 use SMF\IP;
 use SMF\Lang;
+use SMF\Sapi;
 use SMF\Theme;
 use SMF\Time;
 use SMF\Url;
@@ -774,7 +775,7 @@ class Feed implements ActionInterface
 
 			// Limit the length of the message, if the option is set.
 			if (!empty(Config::$modSettings['xmlnews_maxlen']) && Utils::entityStrlen(str_replace('<br>', "\n", $row['body'])) > Config::$modSettings['xmlnews_maxlen']) {
-				$row['body'] = strtr(Utils::entitySubstr(str_replace('<br>', "\n", $row['body']), 0, Config::$modSettings['xmlnews_maxlen'] - 3), ["\n" => '<br>']) . '...';
+				$row['body'] = strtr(Utils::entitySubstr(str_replace('<br>', "\n", $row['body']), 0, (int) (Config::$modSettings['xmlnews_maxlen'] - 3)), ["\n" => '<br>']) . '...';
 			}
 
 			$row['body'] = BBCodeParser::load()->parse($row['body'], $row['smileys_enabled'], $row['id_msg']);
@@ -1207,7 +1208,7 @@ class Feed implements ActionInterface
 
 			// Limit the length of the message, if the option is set.
 			if (!empty(Config::$modSettings['xmlnews_maxlen']) && Utils::entityStrlen(str_replace('<br>', "\n", $row['body'])) > Config::$modSettings['xmlnews_maxlen']) {
-				$row['body'] = strtr(Utils::entitySubstr(str_replace('<br>', "\n", $row['body']), 0, Config::$modSettings['xmlnews_maxlen'] - 3), ["\n" => '<br>']) . '...';
+				$row['body'] = strtr(Utils::entitySubstr(str_replace('<br>', "\n", $row['body']), 0, (int) (Config::$modSettings['xmlnews_maxlen'] - 3)), ["\n" => '<br>']) . '...';
 			}
 
 			$row['body'] = BBCodeParser::load()->parse($row['body'], (bool) $row['smileys_enabled'], (int) $row['id_msg']);
@@ -2915,7 +2916,7 @@ class Feed implements ActionInterface
 				$pos = empty($positions) ? $n : min($positions);
 
 				if ($pos - $old > 0) {
-					$cdata .= substr($data, $old, $pos - $old);
+					$cdata .= substr($data, $old, (int) ($pos - $old));
 				}
 
 				if ($pos >= $n) {
@@ -2930,9 +2931,9 @@ class Feed implements ActionInterface
 					}
 
 					if (substr($data, $pos + 1, 1) == '/') {
-						$cdata .= ']]></' . $ns . ':' . substr($data, $pos + 2, $pos2 - $pos - 1) . '<![CDATA[';
+						$cdata .= ']]></' . $ns . ':' . substr($data, $pos + 2, (int) ($pos2 - $pos - 1)) . '<![CDATA[';
 					} else {
-						$cdata .= ']]><' . $ns . ':' . substr($data, $pos + 1, $pos2 - $pos) . '<![CDATA[';
+						$cdata .= ']]><' . $ns . ':' . substr($data, $pos + 1, (int) ($pos2 - $pos)) . '<![CDATA[';
 					}
 
 					$pos = $pos2 + 1;
@@ -3118,13 +3119,13 @@ class Feed implements ActionInterface
 		if (
 			empty(Config::$modSettings['queryless_urls'])
 			|| (
-				Utils::$context['server']['is_cgi']
+				Sapi::isCGI()
 				&& ini_get('cgi.fix_pathinfo') == 0
 				&& @get_cfg_var('cgi.fix_pathinfo') == 0
 			)
 			|| (
-				!Utils::$context['server']['is_apache']
-				&& !Utils::$context['server']['is_lighttpd']
+				!Sapi::isSoftware(Sapi::SERVER_APACHE)
+				&& !Sapi::isSoftware(Sapi::SERVER_LIGHTTPD)
 			)
 		) {
 			return $val;
--- a/Sources/Actions/Feed.php
+++ b/Sources/Actions/Feed.php
@@ -775,7 +775,7 @@ class Feed implements ActionInterface
 
 			// Limit the length of the message, if the option is set.
 			if (!empty(Config::$modSettings['xmlnews_maxlen']) && Utils::entityStrlen(str_replace('<br>', "\n", $row['body'])) > Config::$modSettings['xmlnews_maxlen']) {
-				$row['body'] = strtr(Utils::entitySubstr(str_replace('<br>', "\n", $row['body']), 0, (int) (Config::$modSettings['xmlnews_maxlen'] - 3)), ["\n" => '<br>']) . '...';
+				$row['body'] = strtr(Utils::entitySubstr(str_replace('<br>', "\n", $row['body']), 0, Config::$modSettings['xmlnews_maxlen'] - 3), ["\n" => '<br>']) . '...';
 			}
 
 			$row['body'] = BBCodeParser::load()->parse($row['body'], $row['smileys_enabled'], $row['id_msg']);
@@ -1208,7 +1208,7 @@ class Feed implements ActionInterface
 
 			// Limit the length of the message, if the option is set.
 			if (!empty(Config::$modSettings['xmlnews_maxlen']) && Utils::entityStrlen(str_replace('<br>', "\n", $row['body'])) > Config::$modSettings['xmlnews_maxlen']) {
-				$row['body'] = strtr(Utils::entitySubstr(str_replace('<br>', "\n", $row['body']), 0, (int) (Config::$modSettings['xmlnews_maxlen'] - 3)), ["\n" => '<br>']) . '...';
+				$row['body'] = strtr(Utils::entitySubstr(str_replace('<br>', "\n", $row['body']), 0, Config::$modSettings['xmlnews_maxlen'] - 3), ["\n" => '<br>']) . '...';
 			}
 
 			$row['body'] = BBCodeParser::load()->parse($row['body'], (bool) $row['smileys_enabled'], (int) $row['id_msg']);
@@ -2916,7 +2916,7 @@ class Feed implements ActionInterface
 				$pos = empty($positions) ? $n : min($positions);
 
 				if ($pos - $old > 0) {
-					$cdata .= substr($data, $old, (int) ($pos - $old));
+					$cdata .= substr($data, $old, $pos - $old);
 				}
 
 				if ($pos >= $n) {
@@ -2931,9 +2931,9 @@ class Feed implements ActionInterface
 					}
 
 					if (substr($data, $pos + 1, 1) == '/') {
-						$cdata .= ']]></' . $ns . ':' . substr($data, $pos + 2, (int) ($pos2 - $pos - 1)) . '<![CDATA[';
+						$cdata .= ']]></' . $ns . ':' . substr($data, $pos + 2, $pos2 - $pos - 1) . '<![CDATA[';
 					} else {
-						$cdata .= ']]><' . $ns . ':' . substr($data, $pos + 1, (int) ($pos2 - $pos)) . '<![CDATA[';
+						$cdata .= ']]><' . $ns . ':' . substr($data, $pos + 1, $pos2 - $pos) . '<![CDATA[';
 					}
 
 					$pos = $pos2 + 1;
--- a/Sources/Actions/Groups.php
+++ b/Sources/Actions/Groups.php
@@ -273,7 +273,7 @@ class Groups implements ActionInterface
 		Utils::$context['can_send_email'] = User::$me->allowedTo('moderate_forum');
 
 		// Load all the group moderators, for fun.
-		$group->loadModerators();
+		User::load($group->loadModerators(), User::LOAD_BY_ID, 'minimal');
 
 		foreach ($group->moderator_ids as $mod_id) {
 			$group->moderators[] = [
--- a/Sources/Actions/Groups.php
+++ b/Sources/Actions/Groups.php
@@ -273,7 +273,7 @@ class Groups implements ActionInterface
 		Utils::$context['can_send_email'] = User::$me->allowedTo('moderate_forum');
 
 		// Load all the group moderators, for fun.
-		User::load($group->loadModerators(), User::LOAD_BY_ID, 'minimal');
+		$group->loadModerators();
 
 		foreach ($group->moderator_ids as $mod_id) {
 			$group->moderators[] = [
--- a/Sources/Actions/Moderation/EndSession.php
+++ b/Sources/Actions/Moderation/EndSession.php
@@ -65,9 +65,9 @@ class EndSession implements ActionInterface
 	/**
 	 * Static wrapper for constructor.
 	 *
-	 * @return object An instance of this class.
+	 * @return self An instance of this class.
 	 */
-	public static function load(): object
+	public static function load(): self
 	{
 		if (!isset(self::$obj)) {
 			self::$obj = new self();
--- a/Sources/Actions/Moderation/EndSession.php
+++ b/Sources/Actions/Moderation/EndSession.php
@@ -65,9 +65,9 @@ class EndSession implements ActionInterface
 	/**
 	 * Static wrapper for constructor.
 	 *
-	 * @return self An instance of this class.
+	 * @return object An instance of this class.
 	 */
-	public static function load(): self
+	public static function load(): object
 	{
 		if (!isset(self::$obj)) {
 			self::$obj = new self();
--- a/Sources/BBCodeParser.php
+++ b/Sources/BBCodeParser.php
@@ -62,63 +62,63 @@ class BBCodeParser
 	 *
 	 * If not empty, only these BBCode tags will be parsed.
 	 */
-	public $parse_tags = [];
+	public array $parse_tags = [];
 
 	/**
 	 * @var bool
 	 *
 	 * Whether BBCode should be parsed.
 	 */
-	public $enable_bbc;
+	public bool $enable_bbc;
 
 	/**
 	 * @var bool
 	 *
 	 * Whether to allow certain basic HTML tags in the input.
 	 */
-	public $enable_post_html;
+	public bool $enable_post_html;
 
 	/**
 	 * @var array
 	 *
 	 * List of disabled BBCode tags.
 	 */
-	public $disabled = [];
+	public array $disabled = [];
 
 	/**
 	 * @var bool
 	 *
 	 * Whether to autolink plain text URLs.
 	 */
-	public $autolink_enabled;
+	public bool $autolink_enabled;
 
 	/**
 	 * @var bool
 	 *
 	 * Whether smileys should be parsed.
 	 */
-	public $smileys = true;
+	public bool $smileys = true;
 
 	/**
 	 * @var string
 	 *
 	 * The smiley set to use when parsing smileys.
 	 */
-	public $smiley_set;
+	public string $smiley_set;
 
 	/**
 	 * @var bool
 	 *
 	 * Whether custom smileys are enabled.
 	 */
-	public $custom_smileys_enabled;
+	public bool $custom_smileys_enabled;
 
 	/**
 	 * @var string
 	 *
 	 * URL of the base smileys directory.
 	 */
-	public $smileys_url;
+	public string $smileys_url;
 
 	/**
 	 * @var string
@@ -132,35 +132,35 @@ class BBCodeParser
 	 *
 	 * Shorthand check for whether character encoding is UTF-8.
 	 */
-	public $utf8 = true;
+	public bool $utf8 = true;
 
 	/**
 	 * @var string
 	 *
 	 * Language locale to use.
 	 */
-	public $locale = 'en_US';
+	public string $locale = 'en_US';
 
 	/**
 	 * @var int
 	 *
 	 * User's time offset from UTC.
 	 */
-	public $time_offset;
+	public int $time_offset;
 
 	/**
 	 * @var string
 	 *
 	 * User's strftime format.
 	 */
-	public $time_format;
+	public string $time_format;
 
 	/**
 	 * @var bool
 	 *
 	 * Enables special handling if output is meant for paper printing.
 	 */
-	public $for_print = false;
+	public bool $for_print = false;
 
 	/**************************
 	 * Public static properties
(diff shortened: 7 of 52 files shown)


Sources/Actions/Admin/ACP.php
```diff
@@ -1615,6 +1615,7 @@ public static function getFileVersions(array &$versionOptions): array
 			$directories += ['template_versions' => Theme::$current->settings['theme_dir']];
 		}
 
+		/** @var string $dirname */
```
Sesquipedalian: This does not seem necessary to me.

Sources/Sapi.php
```diff
@@ -0,0 +1,318 @@
...
+	 * - `sys_get_temp_dir()`
+	 * - `upload_tmp_dir`
+	 * - `session.save_path`
+	 * - `cachedir`
+	 *
+	 * @return string Path to a temporary directory.
+	 */
+	public static function getTempDir(): string
+	{
+		return Config::getTempDir();
```
Sesquipedalian, line 208: ~Is there any reason why you didn't just move the code inside Config::getTempDir() directly into this method?~

Never mind. I see why now. Config needs to be able to get the correct temp dir without any dependencies on other classes.

Sources/Utils.php
```diff
@@ -2018,14 +2004,12 @@ public static function redirectexit(string $setLocation = '', bool $refresh = fa
...
-				empty(Utils::$context['server']['is_cgi'])
+				!Sapi::isCGI()
 				|| ini_get('cgi.fix_pathinfo') == 1
 				|| @get_cfg_var('cgi.fix_pathinfo') == 1
 			)
 			&& (
-				!empty(Utils::$context['server']['is_apache'])
-				|| !empty(Utils::$context['server']['is_lighttpd'])
-				|| !empty(Utils::$context['server']['is_litespeed'])
+				Sapi::isOS([Sapi::SERVER_APACHE, Sapi::SERVER_LIGHTTPD, Sapi::SERVER_LITESPEED])
```
Sesquipedalian:
```suggestion
				Sapi::isSoftware([Sapi::SERVER_APACHE, Sapi::SERVER_LIGHTTPD, Sapi::SERVER_LITESPEED])
```

Sources/Sapi.php
```diff
@@ -0,0 +1,318 @@
...
+
+	public const SERVER_IIS = 'iis';
+	public const SERVER_APACHE = 'apache';
+	public const SERVER_LITESPEED = 'litespeed';
+	public const SERVER_LIGHTTPD = 'lighttpd';
+	public const SERVER_NGINX = 'nginx';
+
+	public const OS_WINDOWS = 'Windows';
+	public const OS_MAC = 'Darwin';
+	public const OS_LINUX = 'linux';
```
Sesquipedalian:
```suggestion
	public const OS_LINUX = 'Linux';
```

Sources/Actions/Admin/Find.php
```diff
@@ -194,7 +195,7 @@ public function execute(): void
 	public function internal(): void
 	{
 		// Try to get some more memory.
-		Config::setMemoryLimit('128M');
+		sapi::setMemoryLimit('128M');
```
Sesquipedalian:
```suggestion
		Sapi::setMemoryLimit('128M');
```

Sources/Actions/Admin/Permissions.php
```diff
@@ -3258,6 +3258,7 @@ protected static function loadAllPermissions(): void
 
 		foreach (self::getPermissions() as $permission => $perm_info) {
 			// If this permission shouldn't be given to certain groups (e.g. guests), don't.
+			/** @var string $perm */
```
Sesquipedalian: This does not seem necessary to me.

Sources/Actions/Feed.php
```diff
@@ -774,7 +775,7 @@ public function getXmlNews(): array
 
 			// Limit the length of the message, if the option is set.
 			if (!empty(Config::$modSettings['xmlnews_maxlen']) && Utils::entityStrlen(str_replace('<br>', "\n", $row['body'])) > Config::$modSettings['xmlnews_maxlen']) {
-				$row['body'] = strtr(Utils::entitySubstr(str_replace('<br>', "\n", $row['body']), 0, Config::$modSettings['xmlnews_maxlen'] - 3), ["\n" => '<br>']) . '...';
+				$row['body'] = strtr(Utils::entitySubstr(str_replace('<br>', "\n", $row['body']), 0, (int) (Config::$modSettings['xmlnews_maxlen'] - 3)), ["\n" => '<br>']) . '...';
```
Sesquipedalian: Casting the result of the expression to int is unnecessary. The result of the arithmetic operation will already be an integer.

Sources/Actions/Feed.php
```diff
@@ -1207,7 +1208,7 @@ public function getXmlRecent(): array
...
```
Sesquipedalian: Casting the result of the expression to int is unnecessary. The result of the arithmetic operation will already be an integer.

Sources/Actions/Feed.php
```diff
@@ -2915,7 +2916,7 @@ public static function cdataParse(string $data, string $ns = '', bool $force = f
 				$pos = empty($positions) ? $n : min($positions);
 
 				if ($pos - $old > 0) {
-					$cdata .= substr($data, $old, $pos - $old);
+					$cdata .= substr($data, $old, (int) ($pos - $old));
```
Sesquipedalian: Casting the result of the expression to int is unnecessary. The result of the arithmetic operation will already be an integer.

Sources/Actions/Feed.php
```diff
@@ -2930,9 +2931,9 @@ public static function cdataParse(string $data, string $ns = '', bool $force = f
 					}
 
 					if (substr($data, $pos + 1, 1) == '/') {
-						$cdata .= ']]></' . $ns . ':' . substr($data, $pos + 2, $pos2 - $pos - 1) . '<![CDATA[';
+						$cdata .= ']]></' . $ns . ':' . substr($data, $pos + 2, (int) ($pos2 - $pos - 1)) . '<![CDATA[';
```
Sesquipedalian: Casting the result of the expression to int is unnecessary. The result of the arithmetic operation will already be an integer.

Sources/Actions/Feed.php
```diff
@@ -2930,9 +2931,9 @@ public static function cdataParse(string $data, string $ns = '', bool $force = f
...
 					} else {
-						$cdata .= ']]><' . $ns . ':' . substr($data, $pos + 1, $pos2 - $pos) . '<![CDATA[';
+						$cdata .= ']]><' . $ns . ':' . substr($data, $pos + 1, (int) ($pos2 - $pos)) . '<![CDATA[';
```
Sesquipedalian: Casting the result of the expression to int is unnecessary. The result of the arithmetic operation will already be an integer.

Sources/Actions/Groups.php
```diff
@@ -273,7 +273,7 @@ public function members(): void
 		Utils::$context['can_send_email'] = User::$me->allowedTo('moderate_forum');
 
 		// Load all the group moderators, for fun.
-		User::load($group->loadModerators(), User::LOAD_BY_ID, 'minimal');
+		$group->loadModerators();
```
Sesquipedalian: This is a good change to make, but it probably belongs in a separate PR.

Sources/Actions/Moderation/EndSession.php
```diff
@@ -65,9 +65,9 @@ public function execute(): void
 	/**
 	 * Static wrapper for constructor.
 	 *
-	 * @return object An instance of this class.
+	 * @return self An instance of this class.
 	 */
-	public static function load(): object
+	public static function load(): self
```
Sesquipedalian: This is a good change to make, but it probably belongs in a separate PR.

Sources/BBCodeParser.php
```diff
@@ -62,63 +62,63 @@ class BBCodeParser
 	 *
 	 * If not empty, only these BBCode tags will be parsed.
 	 */
-	public $parse_tags = [];
+	public array $parse_tags = [];
```
Sesquipedalian: All of the type hinting changes in the BBCodeParser class are good changes to make, but they probably belong in a separate PR.

Sources/BBCodeParser.php
```diff
@@ -2730,7 +2730,7 @@ protected function parseMessage(): void
 				$this->last_pos = max($this->last_pos, 0);
 
 				// Pick a block of data to do some raw fixing on.
-				$data = substr($this->message, $this->last_pos, $this->pos - $this->last_pos);
+				$data = substr($this->message, $this->last_pos, (int) ($this->pos - $this->last_pos));
```
Sesquipedalian: Casting the result of the expression to int is unnecessary. The result of the arithmetic operation will already be an integer.

Sources/BBCodeParser.php
```diff
@@ -2742,7 +2742,7 @@ protected function parseMessage(): void
 				$data = strtr($data, ["\t" => '&nbsp;&nbsp;&nbsp;']);
 
 				// If it wasn't changed, no copying or other boring stuff has to happen!
-				if ($data != substr($this->message, $this->last_pos, $this->pos - $this->last_pos)) {
+				if ($data != substr($this->message, $this->last_pos, (int) ($this->pos - $this->last_pos))) {
```
Sesquipedalian: This is unnecessary. The arithmetic operation will automatically cast the resulting value to int anyway.

Sources/BBCodeParser.php
```diff
@@ -3619,7 +3619,10 @@ protected function closeTags(): void
 			return;
 		}
 
-		$look_for = strtolower(substr($this->message, $this->pos + 2, $pos2 - $this->pos - 2));
+		// @todo: $pos2 could be false?
+		$pos2 = (int) $pos2;
+
+		$look_for = strtolower(substr($this->message, $this->pos + 2, (int) ($pos2 - $this->pos - 2)));
```
Sesquipedalian:
```suggestion
		$look_for = strtolower(substr($this->message, $this->pos + 2, (int) $pos2 - $this->pos - 2));
```

But moreover, this probably belongs in a separate PR.

Sources/BBCodeParser.php
```diff
@@ -3691,9 +3694,9 @@ protected function closeTags(): void
 		}
 
 		foreach ($to_close as $tag) {
-			$this->message = substr($this->message, 0, $this->pos) . "\n" . $tag['after'] . "\n" . substr($this->message, $pos2 + 1);
+			$this->message = substr($this->message, 0, $this->pos) . "\n" . $tag['after'] . "\n" . substr($this->message, (int) ($pos2 + 1));
```
Sesquipedalian: Casting the result of the expression to int is unnecessary. The result of the arithmetic operation will already be an integer.

Sources/BBCodeParser.php
```diff
@@ -3691,9 +3694,9 @@ protected function closeTags(): void
...
 			$this->pos += strlen($tag['after']) + 2;
-			$pos2 = $this->pos - 1;
+			$pos2 = (int) ($this->pos - 1);
```
Sesquipedalian: Casting the result of the expression to int is unnecessary. The result of the arithmetic operation will already be an integer.

Sources/BBCodeParser.php
```diff
@@ -3745,7 +3748,7 @@ protected function detectTag(string $tag_character): array
 			}
 
 			// A test validation?
-			if (isset($possible['test']) && preg_match('~^' . $possible['test'] . '~', substr($this->message, $this->pos + 1 + $pt_strlen + 1)) === 0) {
+			if (isset($possible['test']) && preg_match('~^' . $possible['test'] . '~', substr($this->message, (int) ($this->pos + 1 + $pt_strlen + 1))) === 0) {
```
Sesquipedalian: Casting the result of the expression to int is unnecessary. The result of the arithmetic operation will already be an integer.

Sources/BBCodeParser.php
```diff
@@ -3773,7 +3776,7 @@ protected function detectTag(string $tag_character): array
 				}
 
 				// Maybe we just want a /...
-				if ($possible['type'] == 'closed' && $next_c != ']' && substr($this->message, $this->pos + 1 + $pt_strlen, 2) != '/]' && substr($this->message, $this->pos + 1 + $pt_strlen, 3) != ' /]') {
+				if ($possible['type'] == 'closed' && $next_c != ']' && substr($this->message, (int) ($this->pos + 1 + $pt_strlen), 2) != '/]' && substr($this->message, (int) ($this->pos + 1 + $pt_strlen), 3) != ' /]') {
```
Sesquipedalian: Casting the result of the expression to int is unnecessary. The result of the arithmetic operation will already be an integer.

Sources/BBCodeParser.php
```diff
@@ -4099,7 +4102,7 @@ protected function transformToHtml(array $tag, array $params): void
 				return;
 			}
 
-			$data = substr($this->message, $this->pos1, $pos2 - $this->pos1);
+			$data = substr($this->message, $this->pos1, (int) ($pos2 - $this->pos1));
```
Sesquipedalian: Casting the result of the expression to int is unnecessary. The result of the arithmetic operation will already be an integer.

Sources/BBCodeParser.php
```diff
@@ -4147,7 +4151,7 @@ protected function transformToHtml(array $tag, array $params): void
 
 			$data = [
 				substr($this->message, $pos2 + ($quoted == false ? 1 : 7), $pos3 - ($pos2 + ($quoted == false ? 1 : 7))),
-				substr($this->message, $this->pos1, $pos2 - $this->pos1),
+				substr($this->message, $this->pos1, (int) ($pos2 - $this->pos1)),
```
Sesquipedalian: Casting the result of the expression to int is unnecessary. The result of the arithmetic operation will already be an integer.

Sources/BBCodeParser.php
```diff
@@ -4195,7 +4199,7 @@ protected function transformToHtml(array $tag, array $params): void
 			}
 
 			// We want $1 to be the content, and the rest to be csv.
-			$data = explode(',', ',' . substr($this->message, $this->pos1, $pos2 - $this->pos1));
+			$data = explode(',', ',' . substr($this->message, $this->pos1, (int) ($pos2 - $this->pos1)));
```
Sesquipedalian: Casting the result of the expression to int is unnecessary. The result of the arithmetic operation will already be an integer.

Sources/BBCodeParser.php
```diff
@@ -4220,7 +4224,7 @@ protected function transformToHtml(array $tag, array $params): void
 				return;
 			}
 
-			$data = explode(',', substr($this->message, $this->pos1, $pos2 - $this->pos1));
+			$data = explode(',', substr($this->message, $this->pos1, (int) ($pos2 - $this->pos1)));
```
Sesquipedalian: Casting the result of the expression to int is unnecessary. The result of the arithmetic operation will already be an integer.

Sources/BBCodeParser.php
```diff
@@ -4282,7 +4286,7 @@ protected function transformToHtml(array $tag, array $params): void
 				return;
 			}
 
-			$data = substr($this->message, $this->pos1, $pos2 - $this->pos1);
+			$data = substr($this->message, $this->pos1, (int) ($pos2 - $this->pos1));
```
Sesquipedalian: Casting the result of the expression to int is unnecessary. The result of the arithmetic operation will already be an integer.

Sources/BBCodeParser.php
```diff
@@ -4321,7 +4325,7 @@ protected function transformToHtml(array $tag, array $params): void
 
 		// Are we trimming outside this tag?
 		if (!empty($tag['trim']) && $tag['trim'] != 'outside' && preg_match('~(<br>|&nbsp;|\s)*~', substr($this->message, $this->pos + 1), $matches) != 0) {
-			$this->message = substr($this->message, 0, $this->pos + 1) . substr($this->message, $this->pos + 1 + strlen($matches[0]));
+			$this->message = substr($this->message, 0, $this->pos + 1) . substr($this->message, (int) ($this->pos + 1 + strlen($matches[0])));
```
Sesquipedalian: Casting the result of the expression to int is unnecessary. The result of the arithmetic operation will already be an integer.

Sources/Graphics/Gif/Image.php
```diff
@@ -179,9 +179,9 @@ public function deInterlace(): void
 				$this->m_data = substr($this->m_data, $this->m_gih->m_nWidth);
 
 				$data =
-					substr($data, 0, $y * $this->m_gih->m_nWidth) .
+					substr($data, 0, (int) ($y * $this->m_gih->m_nWidth)) .
```
Sesquipedalian: Casting the result of the expression to int is unnecessary. The result of the arithmetic operation will already be an integer.

Sources/Graphics/Gif/Image.php
```diff
@@ -179,9 +179,9 @@ public function deInterlace(): void
...
 					$lne .
-					substr($data, ($y + 1) * $this->m_gih->m_nWidth);
+					substr($data, (int) (($y + 1) * $this->m_gih->m_nWidth));
```
Sesquipedalian: Casting the result of the expression to int is unnecessary. The result of the arithmetic operation will already be an integer.

Sources/PackageManager/FtpConnection.php
```diff
@@ -293,7 +293,7 @@ public function passive(): bool
 		}
 
 		// This is pretty simple - store it for later use ;).
-		$this->pasv = ['ip' => $match[1] . '.' . $match[2] . '.' . $match[3] . '.' . $match[4], 'port' => $match[5] * 256 + $match[6]];
+		$this->pasv = ['ip' => $match[1] . '.' . $match[2] . '.' . $match[3] . '.' . $match[4], 'port' => (int) ($match[5] * 256 + $match[6])];
```
Sesquipedalian: Casting the result of the expression to int is unnecessary. The result of the arithmetic operation will already be an integer.

Sources/Sapi.php
```diff
@@ -0,0 +1,318 @@
...
+		$oses = (array) $os;
+
+		/** 
+		 * Technically we could simplify this down using PHP_OS_FAMILY,
+		 * but to ensure backwards compatibility, we won't yet.
+		 */
+		foreach ($oses as $os) {
+			switch ($os) {
+				case self::OS_WINDOWS:
+					$is_os = PHP_OS_FAMILY === 'Windows' || DIRECTORY_SEPARATOR === '\\';
```
Sesquipedalian:
```suggestion
					$is_os = PHP_OS_FAMILY === self::OS_WINDOWS || DIRECTORY_SEPARATOR === '\\';
```

Sources/Sapi.php
```diff
@@ -0,0 +1,318 @@
...
+					break;
+				case self::OS_MAC:
+					$is_os = PHP_OS_FAMILY === 'Darwin';
```
Sesquipedalian:
```suggestion
					$is_os = PHP_OS_FAMILY === self::OS_MAC;
```

Sources/Sapi.php
```diff
@@ -0,0 +1,318 @@
...
+					break;
+				// This may result in false positives because 'linux' is very broad.
+				case self::OS_LINUX:
+					$is_os = PHP_OS_FAMILY === 'Linux';
```
Sesquipedalian:
```suggestion
					$is_os = PHP_OS_FAMILY === self::OS_LINUX;
```

Sources/Sapi.php
```diff
@@ -0,0 +1,318 @@
...
+		return empty($_SERVER['REQUEST_METHOD']);
+	}
+
+	/**
+	 * Checks if the server is able to support case folding.
+	 * 
+	 * @see https://www.w3.org/TR/charmod-norm/#definitionCaseFolding
+	 * @return bool True if it does, false otherwise
+	 */
+	public static function isCaseFolding(): bool
```
Sesquipedalian:
```suggestion
	public static function supportsIsoCaseFolding(): bool
```

This name would be clearer, in my opinion,

Sources/Sapi.php
```diff
@@ -0,0 +1,318 @@
...
+		Utils::$context['server'] = [
+			'is_iis' => self::isSoftware(self::SERVER_IIS),
+			'is_apache' => self::isSoftware(self::SERVER_APACHE),
+			'is_litespeed' => self::isSoftware(self::SERVER_LITESPEED),
+			'is_lighttpd' => self::isSoftware(self::SERVER_LIGHTTPD),
+			'is_nginx' => self::isSoftware(self::SERVER_NGINX),
+			'is_cgi' => self::isCGI(),
+			'is_windows' => self::isOS(self::OS_WINDOWS),
+			'is_mac' => self::isOS(self::OS_MAC),
+			'iso_case_folding' => self::isCaseFolding(),
```
Sesquipedalian:
```suggestion
			'iso_case_folding' => self::supportsIsoCaseFolding(),
```

Sources/Unicode/Utf8String.php
```diff
@@ -199,7 +199,7 @@ public function convertCase(string $case, bool $simple = false): object
 
 			switch ($case) {
 				case 'title':
-					$this->convertCase($this->string, 'lower', $simple);
+					$this->convertCase('lower', $simple);
```
Sesquipedalian: This is a good change to make, but it probably belongs in a separate PR.

Sources/Unicode/Utf8String.php
```diff
@@ -1047,7 +1047,7 @@ protected function normalizeKC(): object
 	 */
 	protected function normalizeKCFold(): object
 	{
-		if ($this->isNormalized($this->string, 'kc_casefold')) {
+		if ($this->isNormalized('kc_casefold')) {
```
Sesquipedalian: This is a good change to make, but it probably belongs in a separate PR.

ssi_examples.php
```diff
@@ -475,7 +475,7 @@ function template_ssi_above()
 		</style>
 		<script>
 			var smf_scripturl = "', SMF\Config::$scripturl, '";
-			var smf_iso_case_folding = ', SMF\Utils::$context['server']['iso_case_folding'] ? 'true' : 'false', ';
+			var smf_iso_case_folding = ', SMF\Sapi::isCaseFolding() ? 'true' : 'false', ';
```
Sesquipedalian:
```suggestion
			var smf_iso_case_folding = ', SMF\Sapi::supportsIsoCaseFolding() ? 'true' : 'false', ';
```

Sesquipedalian (changes requested):

Sources/Actions/Admin/ACP.php
jdarwood007: The IDE is thinking that `$dirname` is an array in its static analysis.  This gives it a hint that it's a string.  Note that this will help psalm as well when we implant it to understand the variable.  We can also do `/** @var string[] $directories */` earlier in the code.  The IDE (and static analysis) will also understand then its an array of strings.

Sources/Sapi.php
jdarwood007: I tried.  But we call the getTempDir very early and it breaks Config from loading (autoloader isn't running yet).  I wanted to take another stab at it someday, but the most logical place is for it to be in the Sapi class since it is talking to the server about where to locate a temporary working directory.

If you want it in the Config class only we can. I had wrote all the changes when I came across this, so I simply just pulled it back and updated the one reference in Config class that was broken.

Sources/Actions/Admin/Permissions.php
jdarwood007: Static Analysis hint.

Sources/Actions/Feed.php
jdarwood007: Static analysis says it is a float.  I'm not certain where it gets a float from, I tried to work backwards to figure it out and gave up.

Sources/Sapi.php
Sesquipedalian: No, it's fine this way. What you did makes sense. It's not quite ideal in terms of the purpose of the Sapi class, but it is the best approach in this case.

Sources/Actions/Groups.php
jdarwood007: Well it's just a bug, I came across.  `$group->loadModerators();` returns null.  So calling `User::load()` is not necessary.  From what I can tell the original code thought that `loadModerators` returned an array of user IDs.  However `loadModerators` already calls `User::load()`

Sources/Actions/Feed.php
Sesquipedalian: Well, the static analysis is mistaken. I mean, the result theoretically _could_ be a float if Config::$modSettings['xmlnews_maxlen'] were a float. But it isn't.

Sources/Actions/Groups.php
Sesquipedalian: Oh, I agree that it looks like a good and reasonable change. It's just off-topic for this PR. 🙂

Sources/Actions/Admin/Permissions.php
Sesquipedalian: 😛 Useless clutter.

jdarwood007: I'm removing it, but you have to consider this will be a major problem for static analysis when we introduce psalm for checking our code.  We know/understand things, but if the static analysis is confused, it is going to have the wrong type of information and warn us about incorrect usages in functions.  The hints help it understand what we know.  This is typically true of arrays where the members of the array are unknown to it unless it can see the array and its members defined earlier.  @Tyrsson can provide a peak, but he said there was quite a bit that psalm will warn about if we introduced it today.

Sources/Actions/Feed.php
jdarwood007: It doesn't know that though.  It knows `Config::$modSettings` is an array with unknown members.  It doesn't know the member `xmlnews_maxlen` is an int.  Technically it isn't an int either, because we didn't cast it (I assume), so it's a string.  PHP is still loosely inferring it is an int by coercion.

Sources/Actions/Admin/Permissions.php
Sesquipedalian: Bleh. Fine.

Sources/Actions/Feed.php
Sesquipedalian: Then this would be a better approach:

```
$row['body'] = strtr(Utils::entitySubstr(str_replace('<br>', "\n", $row['body']), 0, (int) Config::$modSettings['xmlnews_maxlen'] - 3), ["\n" => '<br>']) . '...';
```

Cast the original value to int rather than the result of the expression. After all, if the concern is that there could be unpredictable output, then casting the result after the arithmetic operation is done is going to be more unpredictable than casting the original value to the expected type.

jdarwood007: I pulled the changes out for now, so we can focus on the Sapi class.  I will send it up in another PR later.  Seems we need to have discussions on handling the strict typing for things like $modSettings which is causing an unknown int.  I wonder if I could just do it in the PHPDoc with `array|mixed[]|null` and appease the type hinting.

Sesquipedalian (approved):

Sources/Sapi.php
```diff
@@ -0,0 +1,318 @@
...
+	 */
+	public static function isOS(string|array $os): bool
+	{
+		$oses = (array) $os;
+
+		/** 
+		 * Technically we could simplify this down using PHP_OS_FAMILY,
+		 * but to ensure backwards compatibility, we won't yet.
+		 */
+		foreach ($oses as $os) {
```
live627, line 103: multiline comments start with `/*`; phpdoc starts with `/**`

Sesquipedalian: There were a bunch of things in this PR that made PHP-CS-Fixer sad. Apparently the GitHub action didn't catch them for some reason. I push a commit to fix them all. However, we clearly need to fix something in the GitHub action so that it flags these problems like it is supposed to.

jdarwood007: And I have a habit of not checking php-cs-fixer before sending.  But I don't see any logic in our github actions that runs cs fixer.  Just a linter

Sesquipedalian: The problem is that the linter should have flagged the formatting issues and therefore failed the check. It didn't do that.

jdarwood007:
I don't think most things are being checked properly by it
```
2024-02-01T00:43:49.6317556Z ##[group]Run composer lint
2024-02-01T00:43:49.6318036Z [36;1mcomposer lint[0m
2024-02-01T00:43:49.6344510Z shell: /usr/bin/bash -e {0}
2024-02-01T00:43:49.6344910Z ##[endgroup]
2024-02-01T00:43:49.8334027Z > php-cs-fixer check --diff --config .php-cs-fixer.dist.php --path-mode=intersection $(git diff --name-only "*.php")
2024-02-01T00:43:50.0463406Z PHP CS Fixer 3.48.0 Small Changes by Fabien Potencier, Dariusz Ruminski and contributors.
2024-02-01T00:43:50.0464942Z PHP runtime: 8.1.2-1ubuntu2.14
2024-02-01T00:43:50.1026517Z Loaded config default from ".php-cs-fixer.dist.php".
2024-02-01T00:43:50.1027692Z     0 [░░░░░░░░░░░░░░░░░░░░░░░░░░░░]
2024-02-01T00:43:50.1065546Z ##[group]Run php ./vendor/simplemachines/build-tools/check-signed-off.php
2024-02-01T00:43:50.1066513Z [36;1mphp ./vendor/simplemachines/build-tools/check-signed-off.php[0m
2024-02-01T00:43:50.1096272Z shell: /usr/bin/bash -e {0}
2024-02-01T00:43:50.1096675Z ##[endgroup]
2024-02-01T00:43:50.1752223Z fatal: bad object 5f3a3e2f0d754d0ea544a24464f8bee9b5b0aafa
2024-02-01T00:43:50.1773935Z fatal: bad object 5f3a3e2f0d754d0ea544a24464f8bee9b5b0aafa
2024-02-01T00:43:50.1796667Z fatal: bad object 5f3a3e2f0d754d0ea544a24464f8bee9b5b0aafa
2024-02-01T00:43:50.1820262Z fatal: bad object 5f3a3e2f0d754d0ea544a24464f8bee9b5b0aafa
2024-02-01T00:43:50.1841886Z fatal: bad object 5f3a3e2f0d754d0ea544a24464f8bee9b5b0aafa
2024-02-01T00:43:50.1863929Z fatal: bad object 5f3a3e2f0d754d0ea544a24464f8bee9b5b0aafa
2024-02-01T00:43:50.1885141Z fatal: bad object 5f3a3e2f0d754d0ea544a24464f8bee9b5b0aafa
2024-02-01T00:43:50.1906571Z fatal: bad object 5f3a3e2f0d754d0ea544a24464f8bee9b5b0aafa
```

It went through with no errors..  But the checked sign off generated errors and passed through.

Sesquipedalian:
Hm. Those 'bad object' errors look like _Git_ errors. If so, that's not good at all. But I don't know where they would even be coming from; it's not from our own repository, or else we'd be seeing them ourselves constantly.

jdarwood007:
I've dug into it.  I can fix the errors, but it doesn't fix the issue well.  The commit errors may be fixable.  But the cs-fixer isn't.  It needs to produce no output when things are ok and produce output when errors exist.  I don't see a way to do that.  If we can do that, I think we can fix the action.

I'm looking at https://github.com/squizlabs/PHP_CodeSniffer  as a replacement.  A Github Action exists that does line comments.  So that would be awesome.  I've been playing around with creating some custom rules to offset what rules we don't have.  It may take me a while to get through it and configure all the rules.  I have to create custom rules for some cases where we do things differently from the standards.

Sesquipedalian:
Before you do too much with PHP-CodeSniffer, I believe PHP-CS-Fixer has a --quiet option. I'm on my phone right now, so I can't check at the moment, but I suggest looking at that before doing anything else.

jdarwood007:
I think we can fix it then.  I was still testing but I got something to work for the work flow which was essentially
```
    - name: Run Script
      run: |
        php ./check-signed-off.php > check-signed-off.txt 2>&1
        echo 'var<<EOF' >> $GITHUB_OUTPUT
        echo "$(cat check-signed-off.txt)\n" >> $GITHUB_OUTPUT
        echo 'EOF' >> $GITHUB_OUTPUT
        cat check-signed-off.txt

    - name: Fail If we had a message
      if: env.GITHUB_OUTPUT != ''
      run: |
        echo "$env.GITHUB_OUTPUT"
        exit 1
```

That was for sign-off, but we can repeat for linter.  I'm not an expert in GitHub Actions @live627 seems to know more than me and may have a better idea to correct it.

jdarwood007:
Also looks like we may be able to make it add to the comment:
```
    - name: Post result to PR
      uses: mshick/add-pr-comment@v1
      if: env.GITHUB_OUTPUT != ''
      with:
        message-path: ${{ env.GITHUB_OUTPUT }}
        repo-token: ${{ secrets.GITHUB_TOKEN }}
        repo-token-user-login: 'github-actions[bot]' # The user.login for temporary GitHub tokens
        allow-repeats: true # This is the default
```

Needs testing

tyrsson:
Actually https://github.com/squizlabs/PHP_CodeSniffer has been forked into a new project.

Issue detailing it can be found here:
https://github.com/squizlabs/PHP_CodeSniffer/issues/3932

I have been needing to update some of my projects as well just have not had the time lately.

jdarwood007:
Yep, but the composer install is still the "old name".  To be fair, working with it.  There is stuff we cna't do I have found so far, but I agree with what its fixing isn't the best standards anyways.

Sesquipedalian:
I think #8072 will at least solve the problem where we don't want any output from PHP-CS-Fixer unless something is wrong.

Sesquipedalian:
I also suggest that we move further discussion of the GitHub actions and continuous integration to that PR, rather than here in the comments of a closed PR.

PR #8064, opened 2024-01-31
[3.0] Uses ICU MessageFormat for translatable strings

### LINKED ISSUE #2020: grammar issue... 1 days
when you have the max number of days to keep a draft set at 1 days and manually save a draft you get a nice grammar issue.

The contents have been saved as a draft, and will be accessible from the Show Drafts area of your profile. Please note that all drafts will be automatically deleted after 1 days.

also in the admin panel it says days :P

Sesquipedalian:
As discussed internally in topic 587985, SMF 3.0 is changing the format of our translatable strings to use ICU MessageFormat.

This PR implements the necessary code to support ICU MessageFormat, including fallback code to support servers that don't have the intl extension's MessageFormatter class available. It also rewrites all relevant strings in our language files to use MessageFormat instead of sprintf() format.

~This should not be merged until our CrowdIn integration has been correctly configured for it.~ Never mind. Our CrowdIn integration will ignore everything on the release-3.0 branch until we proactively set it up to start paying attention to it. So this PR can be merged whenever ready.

Fixes #2020


Sources/Lang.php
```diff
@@ -505,6 +499,60 @@ public static function get(bool $use_cache = true): array
...
+			if (isset($target[$key])) {
+				$target = &$target[$key];
+			} else {
+				// Can't do anything with a bad key.
+				return '';
+			}
+		}
+
+		if (empty($args)) {
+			return $target;
```
jdarwood007, line 559: Because `$target` is possibly an array and the return type is a string, you should force it at this point to be a string return.  Just in case.  Or this function should allow returning array

Sources/Lang.php
```diff
@@ -505,6 +499,60 @@ public static function get(bool $use_cache = true): array
...
+		}
+
+		return self::formatText($target, $args);
```
jdarwood007, line 562: The same possible issue could occur here.  Take for example if we called `Lang::getTxt(['days'], []);`  We would have `$target` essentially pointing to `$txt['days']`.  Logic would not be happy here.

Sources/Lang.php
```diff
@@ -662,29 +701,51 @@ public static function sentenceList(array $list): string
 	public static function numberFormat(int|float|string $number, ?int $decimals = null): string
 	{
 		if (is_string($number)) {
-			$number = intval($number);
+			$number = $number + 0;
```
jdarwood007, line 704: Any reason for that? I know it's the same result, but it's just loose type converting.  I'm used to (from my job working in other languages) that you can't do that to force them to convert over like that without a casting.  Don't mean to get up on it, just that I'm used to casting things and its not allowed to mix incompatible types.

Sesquipedalian: Because $number might be something like '1.25'.

Sources/Lang.php
Sesquipedalian: Good catch. I hadn't thought of that scenario. If $target is an array at this point, we should throw a ValueError.

Sources/Lang.php
jdarwood007: Ahh, floatval might be the best option then.

Sesquipedalian: Not if $number is something like '1'. 

For the purposes of MessageFormat strings, 1 and 1.0 are distinct values that need different handling in many languages. Therefore, neither intval() nor floatval() are suitable. Adding zero is the simplest, cleanest way to get the necessary behaviour in PHP.

tyrsson: Might be a good place for is_numeric instead of just a is_string check.

Sesquipedalian: Good idea.

dragomano:
We need to use plural for these lines, too:
```
$txt['guest'] = 'Guest';
$txt['guests'] = 'Guests';
$txt['user'] = 'User';
$txt['users'] = 'Users';
$txt['hidden'] = 'Hidden';

// Plural form of hidden for languages other than English
$txt['hidden_s'] = 'Hidden';
$txt['buddy'] = 'Buddy';
$txt['buddies'] = 'Buddies';
```
```
$txt['spider'] = 'Spider';
$txt['spiders'] = 'Spiders';
```

Sesquipedalian:
> We need to use plural for these lines, too:
> 
> ```
> $txt['guest'] = 'Guest';
> $txt['guests'] = 'Guests';
> $txt['user'] = 'User';
> $txt['users'] = 'Users';
> $txt['hidden'] = 'Hidden';
> 
> // Plural form of hidden for languages other than English
> $txt['hidden_s'] = 'Hidden';
> $txt['buddy'] = 'Buddy';
> $txt['buddies'] = 'Buddies';
> ```
> 
>  ```
> $txt['spider'] = 'Spider';
> $txt['spiders'] = 'Spiders';
> ```

Yes, we do.

Themes/default/MessageIndex.template.php
```diff
@@ -34,18 +34,35 @@ function template_main()
...
+		if (!empty(Utils::$context['view_num_guests'])) {
+			$list_of_viewers[] = Lang::getTxt('guest_plural', [Utils::$context['view_num_guests']]);
+		}
+
+		echo '
+			<p>
+				', Lang::getTxt(
+					'who_viewing_board',
+					[
+						'list_of_viewers' => Lang::sentenceList($list_of_viewers),
```
dragomano: Result: `{0} is viewing this board. `
I think, sentenceList should get an array with values only:
`'list_of_viewers' => Lang::sentenceList(array_values($list_of_viewers)),`

The same in Display.template.php

dragomano (changes requested):

Themes/default/MessageIndex.template.php
Sesquipedalian: Ah yes, you are right.

Sources/Actions/Post.php
```diff
@@ -1494,13 +1490,13 @@ protected function showAttachments(): void
 
 			foreach ($attachmentRestrictionTypes as $type) {
 				if (!empty(Config::$modSettings[$type])) {
```
dragomano, line 1492: Add here `Config::$modSettings[$type] = (int) Config::$modSettings[$type];` to prevent possible errors next

Sesquipedalian:
The "Replaces many concatenated strings with MessageFormat strings" commit took a ridiculous amount of time and mind-numbing effort, but I have rewritten nearly all the cases of concatenated strings with full MessageFormat strings. I probably missed a few, but the great majority have been done.

Sesquipedalian:
I'm going to go ahead and merge this monstrous PR now. If any errors crop up, we can fix them as we find them.

jdarwood007:
Perfect.  It's best to do it now while in alpha, we should catch most of them before beta as we progress through the alpha stages.

PR #8069, opened 2024-02-01
[3.0] Removed calls to Settings language

### LINKED ISSUE #8068: [3.0]: Unable to load the 'Settings.' language file.
### Basic Information

This is generated in the error log.
From previous discussion, I assume it should be loading ThemeStrings instead.

However, in the areas where it's needed it appears to be already loaded so the calls to load Settings language file could be dropped instead of replaced.

### Steps to reproduce

_No response_

### Expected result

_No response_

### Actual result

_No response_

### Version/Git revision

3.0 Alpha 1

### Database Engine

MySQL

### Database Version

_No response_

### PHP Version

8.1

### Logs

_No response_

### Additional Information

_No response_

jdarwood007:
Fixes #8068

@Oldiesmann and @DiegoAndresCortes ,  This should resolve it.

### DIFF
--- a/Sources/Actions/Admin/Themes.php
+++ b/Sources/Actions/Admin/Themes.php
@@ -1377,7 +1377,7 @@ class Themes implements ActionInterface
 		// Load the important language files...
 		Lang::load('Admin');
 		Lang::load('Themes');
-		Lang::load('Settings');
+		Lang::load('ThemeStrings');
 		Lang::load('Drafts');
 
 		// Default the page title to Theme Administration by default.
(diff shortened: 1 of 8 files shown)


DiegoAndresCortes:
Given that ThemeStrings might be the main local language file for all themes, should it be loaded globally instead?

jdarwood007:
I have no complaint here.  @Sesquipedalian, are you ok if we set up ThemeSttrings as a global call similar to General+Modifications?

Sesquipedalian:
Sure

Sesquipedalian (approved):

PR #8072, opened 2024-02-02
Fixes to GitHub actions and continuous integration scripts.

Sesquipedalian:
- [x] Only outputs anything from PHP-CS-Fixer if issues were detected
- [ ] Fix 'bad object' messages during sign-off check

@jdarwood007, please feel free to push your own commits to this branch as well.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         "friendsofphp/php-cs-fixer": "^3.40"
     },
     "scripts": {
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Are there other things that need to be fixed, @jdarwood007?

jdarwood007:
I think we can start here.  I was trying to message with actions to get it to post the results of the data to a message on the PR, which would be way more helpful than what we have now.  But I need more time to figure it out.

PR #8077, opened 2024-02-03
Setup action for CS Fixer

jdarwood007:
This will now run PHP-CS-Fixer as a dedicated GitHub Action.  When it finds errors with the changed files, it will fail the check.

This is almost a copy and paste from: https://github.com/OskarStark/php-cs-fixer-ga

### DIFF
--- a/.github/phpcs/SMFClosingTag.php
+++ b/.github/phpcs/SMFClosingTag.php
@@ -1,16 +1,17 @@
 <?php
-
-declare(strict_types=1);
-
-/*
+/**
  * This file is modified from original CS fixer source code.
  *
- * (c) Fabien Potencier <fabien@symfony.com>
- *	 Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ * Simple Machines Forum (SMF)
  *
- * This source file is subject to the MIT license that is bundled
- * with this source code in the file LICENSE.
+ * @package SMF
+ * @author Simple Machines https://www.simplemachines.org
+ * @copyright 2024 Simple Machines and individual contributors
+ * @license https://www.simplemachines.org/about/smf/license.php BSD
+ *
+ * @version 3.0
  */
+declare(strict_types=1);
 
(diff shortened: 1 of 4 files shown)


jdarwood007:
FYI, it ran on this PR as well if you want to see the check details.

jdarwood007:
Additionally,  Just added a custom cs fixer that does the fixing of the line endings, replacing the need for the EOL script.  As a bonus, it means it can also run locally easily for devs.

.github/phpcs/SMFClosingTag.php
```diff
@@ -0,0 +1,114 @@
...
+		// We have closing bracket then closing barcket, single white space and then closing tag. Add one more return.
+		if ($count > 0 && $tokens[$count - 1]->isGivenKind(T_CLOSE_TAG) && $tokens[$count - 2]->isGivenKind(T_WHITESPACE) && $tokens[$count - 3]->getContent() !== ';') {
+			$tokens[$count - 2] = new Token([
+				T_WHITESPACE,
+				$this->whitespacesConfig->getLineEnding() . $this->whitespacesConfig->getLineEnding()
+			]);
+		}
+
+		// We have a closing bracket, and then closing tag, no white space, add returns.
+		// There is no ID/Name for closing curely bracket or semi-colon.
```
tyrsson, line 105: Not sure if this is what you're looking for but check out T_CLOSE_CURLY_BRACKET

jdarwood007: I tried, it doesn't exist.  I checked getId and getName for the char and both came back with nothing.

tyrsson: Yea, appears to be some custom defined const. I just remember seeing it, I think maybe php codesniffer. I guess you could always define it if need be.

Sesquipedalian (approved):

Sesquipedalian:
Thanks for doing this, @jdarwood007

PR #8082, opened 2024-02-05
[3.0] Move external libraries to a separate directory (Libs)

### LINKED ISSUE #7977: ReCaptcha Lib needs updated for 3.0
Just opening this as a reminder for us.

It appears our version is at 1.2.4 (according to the last commit message regarding it) and the libs current version is 1.3 which adds php 8 support (I only had a minute or two to look over it).

dragomano:
This PR moves third-party libraries (Minify and ReCaptcha) to a separate subdirectory within Sources — `Libs`.

It also adds a command (see composer.json in the root) to update these libraries (experts can to configure this command to run before building the package so that the Libs folder is included in the SMF distribution package).

Also implemented are 2 of SMF's own classes, JS and CSS, which extend the external classes MatthiasMullie\Minify\JS and MatthiasMullie\Minify\CSS respectively. This is done to add an additional comment at the top of each minimized file (previously this required changes to the updated Minify.php each time).

This way, external libraries can be easily updated with each release.

Fixes #7977


live627:
> Also implemented are 2 of SMF's own classes, JS and CSS, which extend the external classes MatthiasMullie\Minify\JS and MatthiasMullie\Minify\CSS respectively. This is done to add an additional comment at the top of each minimized file (previously this required changes to the updated Minify.php each time).

- https://github.com/SimpleMachines/SMF/pull/8240 does not add this.

PR #8090, opened 2024-02-07
[3.0] Fix an issue with wrong date on event creating

dragomano:
The same as #7775, but for 3.0

### DIFF
--- a/Sources/Actions/Calendar.php
+++ b/Sources/Actions/Calendar.php
@@ -1804,7 +1804,7 @@ class Calendar implements ActionInterface
 		]);
 
 		// Fix for languages those do not use AM/PM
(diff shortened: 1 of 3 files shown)


Sources/Actions/Calendar.php
```diff
@@ -1802,6 +1802,12 @@ public static function convertDateToEnglish(string $date): string
 		$replacements += array_combine(array_map('strtolower', Lang::$txt['days_short']), [
 			'Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat',
 		]);
+
+		// Fix for languages those do not use AM/PM
+		if (Lang::$txt['time_am'] == ' ' && Lang::$txt['time_pm'] == ' ') {
```
jdarwood007: Use `trim()` here and `=== ''`

PR #8093, opened 2024-02-08
Installer and upgrader rewrite

### LINKED ISSUE #8261: [3.0]: Check for "calendar_holidays" table during upgrade could return false positives
### Basic Information

The check to see if a "calendar_holidays" table exists prior to attempting to convert holidays to the new events format could return a false positive if there are old backup tables, another (older) SMF installation on the same database or something else using the same database that happens to have a table with a name ending in "calendar_holidays". This will lead to an error on the next step since SMF will be trying to pull from {db_prefix}calendar_holidays when the table doesn't exist

### Steps to reproduce

1. Run the upgrader when the database that SMF is using doesn't have the "{prefix}calendar_holidays" table but has another table with a name ending in "calendar_holidays" (such as "backup_calendar_holidays")

### Expected result

Code should determine that the specific table it's looking for isn't there

### Actual result

Code checks for ''any'' table ending in "calendar_holidays", causing an error if there are old backup tables or another table matching "%calendar_holidays".

### Version/Git revision

3.0 Alpha 1

### Database Engine

All

### Database Version

10.11.8-MariaDB-ubu2204

### PHP Version

8.3.8

### Logs

_No response_

### Additional Information

Filing this instead of submitting a PR because I can't tell if the db_list_tables method supports the standard {db_prefix} replacement syntax or not.

Sesquipedalian:
@jdarwood007, let's use this to work collaboratively on the installer/upgrader rewrite.

- [x] Add Schema for 3.0
- [x] Add Initial framework for Maintenance logic
- [x] Build Installer logic 
- [x] Setup Migration logic for upgrades from SMF 2.0 to 2.1
- [x] Setup Migration logic for upgrades from SMF 2.1 to 3.0
- [ ] Test Installer logic
- [ ] Build Upgrader logic
- [ ] Test Upgrader logic
- [ ] Build Converter Logic
- [ ] Setup Converter logic for YabbSE
- [ ] Setup Converter logic for SMF 1.0
- [ ] Setup Converter logic for SMF 1.1
- [ ] Build Repair Settings tool
- [ ] Build logic for self-contained tools to utilize the maintenance logic.
- [ ] Add logic back for installer to work with compression (?obgz)
- [ ] Allow upgrader to support login or entering in the database password.

Fixes #8261


Sources/Db/Schema/Column.php
```diff
@@ -0,0 +1,202 @@
...
+	 * @param ?string $charset The character set for string data.
+	 *    Only applicable to string types. If null, will be set automatically.
+	 */
+	public function __construct(
+		string $name,
+		string $type,
+		?int $size = null,
+		?bool $unsigned = null,
+		?bool $not_null = null,
+		mixed $default = null,
```
live627: Could this type be more specific? Example: `mixed` is shorthand for `object|resource|array|string|float|int|bool|null`

```suggestion
		string|float|int|bool|null $default = null,
```

Sources/Db/Schema/Column.php
```diff
@@ -0,0 +1,202 @@
...
+	 * Whether the column disallows null values.
+	 */
+	public ?bool $not_null;
+
+	/**
+	 * @var mixed
+	 *
+	 * Default value of the column.
+	 */
+	public mixed $default;
```
live627:
```suggestion
	/** @var string|float|int|bool|null Default value of the column. */
	public string|float|int|bool|null $default;
```

jdarwood007:
A lot is incoming, it's my initial work I started to write the installer logic.  You should be able to get through an install.  But the base logic should get it going for an upgrader and converter logic.

jdarwood007:
I've also added some task notes based on what I was thinking, please amend or adjust.

Sources/Db/APIs/MySQL.php
```diff
@@ -1362,19 +1360,11 @@ public function calculate_type(string $type_name, ?int $type_size = null, bool $
...
 			} elseif ($type_name == 'varbinary' && $reverse && $type_size == 16) {
 				$type_name = 'inet';
 				$type_size = null;
-			} elseif ($type_name == 'binary' && $reverse && $type_size == 16) {
-				$type_name = 'uuid';
-				$type_size = null;
 			} elseif ($type_name == 'varbinary') {
 				$type_name = 'varbinary';
-			} elseif ($type_name == 'binary') {
-				$type_name = 'binary';
```
Oldiesmann: Did you mean to remove all of this?

Sources/Db/Schema/Index.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+// Try to handle it with the upper level index.php. (it should know what to do.)
+if (file_exists(dirname(__DIR__) . '/index.php'))
+	include (dirname(__DIR__) . '/index.php');
+else
+	exit;
+
+?>
```
dragomano: Don't forget to rename this file to `index.php`.

Sesquipedalian:
Please remember not to force-push to a shared branch, @jdarwood007. I was able to reconcile the differences in my local repository fairly easily this time, but another time I might not be so lucky.

jdarwood007:
I will try to remember that.  Would you prefer a merge then?  I was bringing it up to the current 3.0 branch.

Sesquipedalian:
> I will try to remember that. Would you prefer a merge then? I was bringing it up to the current 3.0 branch.

For that situation, yeah, a merge would be the better option.

In general, I'm just asking that you don't amend commits after they've been pushed to this branch. Feel free to amend commits and rebase stuff as much as you like in your local working copy before you push them. That's cool. But once a commit has been pushed to this shared branch, it shouldn't be changed.

jdarwood007:
Will do merge then.  Rebase wants me to do a force push to get it into sync.

jdarwood007:
@Sesquipedalian Minor detail, the check script check-smf-license.php fails, because its looking for SMF_VERSION in upgrade.php.  I'm wondering why this file was checked instead of index.php.  Any issue if we change that?

Sesquipedalian:
None at all.

Sources/Maintenance/Tools/Install.php
```diff
@@ -0,0 +1,1753 @@
...
+			}
+		}
+		Config::$modSettings['disableQueryCheck'] = true;
+
+		// Windows likes to leave the trailing slash, which yields to C:\path\to\SMF\/attachments...
+		if (Sapi::isOS(Sapi::OS_WINDOWS)) {
+			$attachdir = $path . 'attachments';
+		} else {
+			$attachdir = $path . '/attachments';
+		}
```
Sesquipedalian: This seems wrong. Line 857 above has this:

```
$path = rtrim(str_replace(DIRECTORY_SEPARATOR, '/', Maintenance::getBaseDir()), '/');
```

I don't see how there could ever be a trailing slash in the $path variable on any operating system.

Sesquipedalian:
In my tests, the installer now works nicely.

However, I have not tested on a Windows server (and cannot do so). @jdarwood007, if you can confirm that the installer works on Windows, we can mark off that task on the checklist.

Sources/Maintenance/Tools/Install.php
```diff
@@ -1503,6 +1525,30 @@ private function createImageProxySecret(): string
...
+	 *
+	 * @return bool True if we could update our settings file, false otherwise.
+	 */
+	private function saveUpgradeData(): bool
+	{
+		return Config::updateSettingsFile(['upgradeData' => json_encode([
+			'started' => $this->time_started,
+		])]);
+	}
+
```
jdarwood007: Technically should be called save/get InstallData, but doesn't mater.  Didn't even think of yet that the installer now has a timer since it is getting it from the upgrader.

jdarwood007:
> In my tests, the installer now works nicely.
> 
> However, I have not tested on a Windows server (and cannot do so). @jdarwood007, if you can confirm that the installer works on Windows, we can mark off that task on the checklist.

I will have to setup an instance on my windows box but can.  I still need to get through upgrade process and confirm it works.  Will do it after that since we should just have conversion logic to build for YabbSE/SMF 1.x.  After that its cleanup and getting CLI working (I'm adding code but not testing it yet).

Sources/Maintenance/Tools/Install.php
jdarwood007: I think this is a poor conversion of the code.  https://github.com/SimpleMachines/SMF/blob/d5943104856f57851415c56c1b392dabb72a7d18/other/install.php#L1194

Sources/Maintenance/Tools/Install.php
Sesquipedalian: Yeah, I decided to just stick with "upgradeData" in order to avoid having to add something else to check for in SMF\Config.

Sources/Maintenance/Tools/Install.php
Sesquipedalian: > I think this is a poor conversion of the code.  https://github.com/SimpleMachines/SMF/blob/d5943104856f57851415c56c1b392dabb72a7d18/other/install.php#L1194

Yeah, in light of that, I think we can get rid of the special case for Windows.

Sesquipedalian:
FYI, I removed the $initial_columns property from SMF\Db\Schema\Table and all the table classes that extended it. Between the data in the $initial_data property and the column definitions in the Column objects, we already have all we need.

jdarwood007:
@Sesquipedalian Might be all the time I have for today, but I added some initial logic to show how the upgrade processes.

Already ran into an instance where a query is not both MySQL and PostgreSQL safe.  But others I was able to make it safe.  I haven't tried to execute the code yet, just working on converting the steps from .sql files into a .php file.

Sources/Maintenance.php
```diff
@@ -0,0 +1,836 @@
...
+
+	/**
+	 * See if we think they have already installed it?
+	 *
+	 * @return bool Whether we believe SMF has been installed.
+	 */
+	public static function isInstalled(): bool
+	{
+		foreach (['image_proxy_secret', 'db_passwd', 'boardurl'] as $var) {
+			if (Config::${$var} === Config::$settings_defs[$var]['default']) {
```
jdarwood007: @Sesquipedalian $settings_defs is protected and is not accessible.

Sesquipedalian: Oh, right.

Sources/Maintenance/Migration/v2_1/Migration1023.php
```diff
@@ -0,0 +1,110 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace SMF\Maintenance\Migration\v2_1;
+
+use SMF\Db\DatabaseApi as Db;
+use SMF\Maintenance\Migration;
+
+class Migration1022 extends Migration
```
Oldiesmann: This should be Migration1023

Sesquipedalian:
I made some changes, @jdarwood007. 

First, I moved some stuff around and adjusted some names to make naming and organization more consistent. 

Second, instead of using numbered migration sub-steps, I gave them all names and created an array constant (SMF\Maintenance\Tools\Upgrade::MIGRATIONS) that lists them all in the intended order. I did this because (1) it is more reliable to use an order that is explicitly defined in code rather than an implicit order that depends on number file names, and (2) this will make it easier to adjust the order whenever necessary. The latter consideration will become important once we start introducing more database changes as 3.0 develops.

jdarwood007:
That works as well.  My intention for the numbering was to just glob the directory, which then with them named in an order, would return a list for us to use to iterate through with an iterator seek (https://www.php.net/manual/en/arrayiterator.seek.php) when we restart the script.

I have had no time for coding lately.  It spring time, so house work to be done and I'm releasing a new app for work.  Trying to take time off, but its going to be for more house work.

jdarwood007:
I lost the last half hour of work I did to custom fields.  Will try to work on this more as I get time.  I am currently in the 'Messenger fields' section in the SQL files.  You have to compare both MySQL and Postgresql because we appear to sometimes do different things or skip steps in some.  The SQL files are still about 2500 lines of code to review.  We still need to build converter logic as well to handle YabbSE, SMF 1.x, and SMF 2.0 (if we don't want to rewrite its upgrade).  As well we will have to completely revalidate the entire 2.0 to 2.1 upgrade logic here.

Oldiesmann:
Found an install-related bug that should be fixed in this PR assuming it still applies (I haven't looked at the code yet). The MessageFormatter stuff relies on $txt['lang_locale'], which is defined in the index language file. However, this file isn't loaded in install.php, which will cause errors.
Normally I'd just submit a PR for the fix but since you're in the process of rewriting the installer anyway I didn't think that was necessary.

tyrsson (commented):
In reference to the add method in the Column class.

Imho this should not be allowed from a column class. Table classes should consume column classes. 

Providing means for a column to add itself to a table is kinda like the tail wagging the dog.

Sources/Db/Schema/Column.php
```diff
@@ -0,0 +1,203 @@
...
+	 * Adds this column to the specified table.
+	 *
+	 * @see SMF\Db\DatabaseApi::add_column
+	 *
+	 * @param string $table_name The name of the table to add the column to.
+	 * @param string $if_exists What to do if the column exists.
+	 *    If 'update', column is updated.
+	 * @return bool Whether or not the operation was successful.
+	 */
+	public function add(string $table_name, string $if_exists = 'update'): bool
```
tyrsson: This method is the one I reference in my other comment. 

Why should columns be concerned with tables? Tables should be concerned about columns. Not vice-versa.

The Table class should expose an addColumn method. The Table class already knows which db table its tied to by its $name property. Well, I say it should expose an addColumn. Honestly there should be a group of DDL objects that expose this, and Column should be the parent.

Question:
How, in the future, do you propose to test that the column this instance represents is representative of a Binary, DateTime or TimeStamp for example? We cant test based on type as that is apparent because we do not have objects that represent those "types". So the only choice is to test the instances state since this Column instance can possibly represent all "types". It handles all responsibilities. It should handle only the common responsibilities shared for all column types. Int and text are very different creatures. SRP in OO is a thing for a reason.

The behavioral contract here is that tables contain columns. You do not write queries of the nature (sudo code):
COULMN ALTER TABLE 
its ALTER TABLE ADD, DROP, MODIFY
You can not get there from a column. Any code / API modeling it should not be able to get there either. It's a breach of contract in respect to the objects responsibility. What it represents in the context of the system should not be able to perform the action this method provides.

jdarwood007:
I know ICQ was removed, but the logic was for 2.0 to 2.1 upgrade.  Shouldn't it be preserved?  Let the 3.0 upgrader remove it later?

Sesquipedalian:
You are correct. I will revert that change.

Sesquipedalian:
"Fix 2.1 schema namespace"

Wow. I can't believe I missed that. Good catch!

jdarwood007:
I didn't either until I remembered I was writing the upgrader for 2.0 to 2.1 and I was referencing 3.0 for the schema.  Tried to use 2.1 namespace and my editor complained.  Tracked it down to that.

jdarwood007:
Finally got through the 2.1 upgrader sql logic.  What a mess that was.

I started on the migration logic itself.

I also moved the Maintenance.php from Sources to the Maintenance folder.  Had to fix some references for that.  If we don't want it there, we can move it back.

I also deleted the upgrade .SQL files for both 2.1 and 3.0.  It will make a merge conflict if we make changes, but forces us to find and fix anything.  I put the install.php in place.  You can swap out install for the upgrade call to see it.  I didn't want to overwrite upgrade.php until I have finished using it for referencing while building the migrations,cleanup and done steps to the upgrader.

But I think having the install.php will let us fully test the installer to make sure we are happy with it and can cross that off the list.

There is a note where I left off on the migration logic I am working on for the upgrade itself.  I need to get back but wanted to get loaded what I have so far.

jdarwood007:
@Sesquipedalian 
I'm running into an error: 
`( ! ) Fatal error: Uncaught Error: Typed static property SMF\User::$me must not be accessed before initialization in /Sources/BrowserDetector.php on line 131`

Its being called in the Migration/v2_1/BoardDescriptions
```
					'name' => Utils::htmlspecialchars(strip_tags(BBCodeParser::load()->unparse($row['name']))),
```

What is the optimal call here to put the best effort that SMF can do to safely startup the User::$me object?  Keeping in mind we may be coming from a env in which it may not be able to startup things.

jdarwood007:
```
		// Load up the current user safely.
		$p = User::load(0);
		User::$me = $p[0];
```

This seems to call ok.

Sesquipedalian:
> ```
> 		// Load up the current user safely.
> 		$p = User::load(0);
> 		User::$me = $p[0];
> ```
> 
> This seems to call ok.

This would be better:

```
		// Load up the current user safely.
		User::setMe(0);
```

Also, is it indeed the intention here to load a guest user? Because that's what loading user 0 does.

Sesquipedalian:
Moved to Alpha 4 as per the internal roadmap.

Sesquipedalian:
Due to of a lot of ugliness and misery involved in trying to resolve conflicts, and due to @jdarwood007 informing me that he has limited time to work on it for the next while, I've decided to close this PR and reimplement this code in a new one. The fundamental logic is all there now thanks to @jdarwood007's work, so I will just need to make a few adjustments and finish off the remaining migration step classes, etc., that have not been written yet.

PR #8102, opened 2024-02-13
Minor fixes for Lang.php

jdarwood007:
Declare ValueError usage
formatText does not accept 3 parameters.

### DIFF
--- a/Sources/Lang.php
+++ b/Sources/Lang.php
@@ -17,6 +17,7 @@ namespace SMF;
 
 use SMF\Cache\CacheApi;
 use SMF\Localization\MessageFormatter;
+use ValueError;
 
 /**
  * Handles the localizable strings shown in SMF's user interface.
@@ -326,7 +327,7 @@ class Lang
 
 			// That couldn't be found!  Log the error, but *try* to continue normally.
 			if (!$found && $fatal) {
-				ErrorHandler::log(self::formatText(self::$txt['theme_language_error'] ?? 'Unable to load the {filename} language file.', ['filename' => $lang . '/' . $template_name], 'template'));
+				ErrorHandler::log(self::formatText(self::$txt['theme_language_error'] ?? 'Unable to load the {filename} language file.', ['filename' => $lang . '/' . $template_name]));
 				break;
 			}
 
--- a/Sources/Lang.php
+++ b/Sources/Lang.php
@@ -327,7 +327,7 @@ class Lang
 
 			// That couldn't be found!  Log the error, but *try* to continue normally.
 			if (!$found && $fatal) {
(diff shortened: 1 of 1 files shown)


Sources/Lang.php
```diff
@@ -326,7 +327,7 @@ public static function load(string $template_name, string $lang = '', bool $fata
 
 			// That couldn't be found!  Log the error, but *try* to continue normally.
 			if (!$found && $fatal) {
-				ErrorHandler::log(self::formatText(self::$txt['theme_language_error'] ?? 'Unable to load the {filename} language file.', ['filename' => $lang . '/' . $template_name], 'template'));
+				ErrorHandler::log(self::formatText(self::$txt['theme_language_error'] ?? 'Unable to load the {filename} language file.', ['filename' => $lang . '/' . $template_name]));
```
Sesquipedalian:
```suggestion
				ErrorHandler::log(self::formatText(self::$txt['theme_language_error'] ?? 'Unable to load the {filename} language file.', ['filename' => $lang . '/' . $template_name]), 'template');
```

Sesquipedalian (changes requested):

Sources/Lang.php
```diff
@@ -17,6 +17,7 @@
 
 use SMF\Cache\CacheApi;
 use SMF\Localization\MessageFormatter;
+use ValueError;
```
Sesquipedalian: If we are adding this, then we might as well replace all instances of `\ValueError` with just `ValueError` in the rest of the file.

Sesquipedalian (changes requested):

Sources/Lang.php
jdarwood007: I could go either way.  Just an IDE warning while I was looking at it. 

What is the preferred method here for PHP's built-ins?  Use \ or define the use?

Sesquipedalian: I generally prefer using fully qualified names (e.g. `\ValueError`) for built-in classes, for two reasons. 

1. There is less to maintain. Creating aliases of classes in the global namespace in the local namespace via `use` statements accomplishes very little except requiring us to spend more time keeping track of `use` statements.
2. It makes the code clearer. When I see a FQN in the code, I can immediately know that I am dealing with a built-in class, without having to waste any time scrolling to the top of the file to see if there are any unexpected class aliases in play.

All of which is to say that I'd actually prefer to see this `use` statement removed.

jdarwood007: Thats fine with me.  Pick a standard.  So for built-ins,  we use \, for all others we define the use.

Sesquipedalian (approved):

PR #8111, opened 2024-02-20
[2.1] Move hooks filter above table

dragomano:
Sometimes the hooks table layout is broken due to too long hook names. Moving the filter from the table header upwards solves this problem.

Before this change:
sshot-11

After:
sshot-12


DiegoAndresCortes:
Would #8059 work on this scenario too? (as a similar solution)

dragomano:
Yeah, if we update this rule:
```css
#display_jump_to_select,
#message_index_jump_to_select,
#search_jump_to_select,
#quick_mod_jump_to_select,
#list_integration_hooks select {
	width: 29ch;
	overflow: hidden;
}
```

Can I use selector `#list_integration_hooks select` or should I add any ID for filter select field?

sshot-4

DiegoAndresCortes:
That selector should be good

Sesquipedalian (approved):

PR #8117, opened 2024-02-21
Fixes PHP-CS runner

jdarwood007:
This should resolve an issue with the php-cs runner where it is failing because of deleted and non-PHP files.


Sesquipedalian:
A nice thing about this PR is that it proves itself to be good.

jdarwood007:
I hope it does. I downloaded and ran 'act' to run the github actions locally until I could get it to work without errors.

PR #8127, opened 2024-03-04
[3.0] Fix an issue with mediumtext type

### LINKED ISSUE #8126: [3.0]: Adding columns with type = 'mediumtext' does not work
### Basic Information

This script fine works in 2.1, but doesn't work in 3.0.

It must add two tables - lp_page_tags, and lp_pages, but adds only lp_page_tags, for unknown reasons.

### Steps to reproduce

1. Try to run this database.php:
```php
<?php

if (file_exists(__DIR__ . '/SSI.php') && ! defined('SMF')) {
	require_once __DIR__ . '/SSI.php';
} elseif (! defined('SMF')) {
	die('<b>Error:</b> Cannot install - please verify that you put this file in the same place as SMF\'s index.php and SSI.php files.');
}

if ((SMF === 'SSI') && ! $user_info['is_admin']) {
	die('Admin privileges required.');
}

$tables = [];

$tables[] = [
	'name'    => 'lp_page_tags',
	'columns' => [
		[
			'name'     => 'page_id',
			'type'     => 'int',
			'size'     => 10,
			'unsigned' => true
		],
		[
			'name'     => 'tag_id',
			'type'     => 'int',
			'size'     => 10,
			'unsigned' => true
		]
	],
	'indexes' => [
		[
			'type'    => 'primary',
			'columns' => ['page_id', 'tag_id']
		]
	]
];

$tables[] = [
	'name' => 'lp_pages',
	'columns' => [
		[
			'name'     => 'page_id',
			'type'     => 'smallint',
			'size'     => 5,
			'unsigned' => true,
			'auto'     => true
		],
		[
			'name' => 'description',
			'type' => 'varchar',
			'size' => 255,
			'null' => true
		],
		[
			'name' => 'content',
			'type' => 'mediumtext',
			'null' => false
		],
		[
			'name'    => 'type',
			'type'    => 'varchar',
			'size'    => 10,
			'default' => 'bbc',
			'null'    => false
		],	
	],
	'indexes' => [
		[
			'type'    => 'primary',
			'columns' => ['page_id']
		],
		[
			'type'    => 'unique',
			'columns' => ['alias']
		]
	],
];

db_extend('packages');

foreach ($tables as $table) {
	$smcFunc['db_create_table']('{db_prefix}' . $table['name'], $table['columns'], $table['indexes']);
}

if (SMF === 'SSI') {
	echo 'Database changes are complete! Please wait...';
}

```

### Expected result

Two tables in the database: `lp_page_tags`, `lp_pages`

### Actual result

One table in the database: `lp_page_tags`

### Version/Git revision

d594310

### Database Engine

MySQL

### Database Version

10.11.6-MariaDB

### PHP Version

8.2.16

### Logs

_No response_

### Additional Information

_No response_

dragomano:
Fixes #8126

### DIFF
--- a/Sources/Db/APIs/MySQL.php
+++ b/Sources/Db/APIs/MySQL.php
@@ -1268,7 +1268,7 @@ class MySQL extends DatabaseApi implements DatabaseApiInterface
 			// If a size was already specified, we won't be able to match it anyways.
 			if (
 				!isset($cols[$c])
-				|| !in_array($cols[$c]['type'], ['text', 'mediumntext', 'largetext', 'varchar', 'char'])
+				|| !in_array($cols[$c]['type'], ['text', 'mediumtext', 'largetext', 'varchar', 'char'])
 				|| (
 					isset($size)
 					&& $size <= 191
@@ -1575,7 +1575,7 @@ class MySQL extends DatabaseApi implements DatabaseApiInterface
 				if (
 					$key === false
 					|| !isset($columns[$key])
-					|| !in_array($columns[$key]['type'], ['text', 'mediumntext', 'largetext', 'varchar', 'char'])
+					|| !in_array($columns[$key]['type'], ['text', 'mediumtext', 'largetext', 'varchar', 'char'])
 					|| (
 						isset($size)
 						&& $size <= 191
--- a/Sources/Db/APIs/MySQL.php
+++ b/Sources/Db/APIs/MySQL.php
@@ -1268,7 +1268,7 @@ class MySQL extends DatabaseApi implements DatabaseApiInterface
 			// If a size was already specified, we won't be able to match it anyways.
 			if (
 				!isset($cols[$c])
-				|| !in_array($cols[$c]['type'], ['text', 'mediumtext', 'largetext', 'varchar', 'char'])
+				|| !in_array($cols[$c]['type'], ['text', 'mediumtext', 'longtext', 'varchar', 'char'])
 				|| (
 					isset($size)
 					&& $size <= 191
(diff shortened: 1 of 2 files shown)


Sources/Db/APIs/MySQL.php
```diff
@@ -1349,6 +1349,8 @@ public function calculate_type(string $type_name, ?int $type_size = null, bool $
 			$types = [
 				'inet' => 'varbinary',
 				'uuid' => 'binary',
+				'mediumtext' => 'text',
+				'largetext' => 'text',
```
Sesquipedalian: This doesn't seem correct to me. Why would we want to change MEDIUMTEXT and LARGETEXT data types to regular TEXT?

Sesquipedalian (changes requested):

Sources/Db/APIs/MySQL.php
dragomano: You're right. This means something changed somewhere, because in Alpha 1 `database.php` worked as expected.

dragomano: Now a query like this is generated:
```sql
CREATE TABLE `smf_alpha`.smf_lp_pages ( `page_id` smallint(5) unsigned auto_increment, `alias` varchar(255) NOT NULL , `description` varchar(255) , `content` mediumtext(0) NOT NULL , `type` varchar(10) NOT NULL DEFAULT 'bbc', PRIMARY KEY (page_id), UNIQUE alias (alias(191))) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci
```
And it needs to be like this:
```sql
CREATE TABLE `smf_alpha`.smf_lp_pages ( `page_id` smallint(5) unsigned auto_increment, `alias` varchar(255) NOT NULL , `description` varchar(255) , `content` mediumtext NOT NULL , `type` varchar(10) NOT NULL DEFAULT 'bbc', PRIMARY KEY (page_id), UNIQUE alias (alias(191))) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci
```

jdarwood007: type hinting issue is defaulting the size to 0 and it needs to be nullable so when null we don't pass a size.

Sesquipedalian (approved):

Issue #8128, opened 2024-03-04
[2.1.4]: GMail semicolon encoding


itspoonmkohv:
### Basic Information

GMail encodes activation links with uppercase urlencode. SMF checks for lowercase urlencoded semicolons. Thus activation links doesn't work for GMail.

### Steps to reproduce

1. Create new account with GMail
2. Click on activation link

### Expected result

Account get's verified

### Actual result

An error has occurred. Sorry, this section isn't available at this time.

### Version/Git revision

2.1.4

### Database Engine

MySQL

### PHP Version

10.3.28-MariaDB

DiegoAndresCortes:
FYI: This was reported for 2.1.4 but the fix merged was for 3.0

Sesquipedalian:
You're right. Thanks.

PR #8129, opened 2024-03-04
GMail semicolon encoding issue #8128 bugfix

itspoonmkohv:
Fixes issue, where GMail encodes semicolons with %3B but SMF only detects %3b. 
For example /index.php?action=activate%3Bu=1358%3Bcode=dxxxxxxxxx9


live627:
can you amend your commit to include your `signed-off-by`, please?

PR #8132, opened 2024-03-10
[3.0] Replace AT.js (@mentions) with Tribute.js

dragomano:
The script used to mention users was updated 9 years ago, it would be nice to replace it.

The AT.js developer himself suggests using Tribute.js as a replacement, so I chose [one of the latest forks](https://github.com/SrPatinhas/tribute).

Features
* single script (`tribute.js`) instead of 2 (`jquery.atwho.min.js`, `jquery.caret.min.js`)
* no dependencies on jQuery
* there is no need to change the PHP backend code (we only change `mentions.js`)
* there is a possibility of fine tuning (see [options](https://github.com/SrPatinhas/tribute))

sshot-2


live627:
Thanks! (just the emoji isn't enough)

Sesquipedalian (changes requested):
I definitely like the idea of updating to a more recent and maintained script. 

However, more work needs to be done on this PR in order to make it fully compatible with SCEditor. In particular, the script only works when the editor is in source mode. It doesn't interact at all with the editor when it is in WYSIWYG mode.

Could you get this working with WYSIWYG mode, @dragomano? Once that is done, I will be glad to merge this PR.

dragomano:
If SCEditor is planned to be kept, it would be better to write a separate mentions plugin specifically for it. As it turns out, Tribute.js has issues with [support for iframe and WYSIWYG editors](https://github.com/zurb/tribute/issues?q=is%3Aissue+is%3Aopen+iframe).

Sesquipedalian:
> If SCEditor is planned to be kept

It is. If you want to find or create a solution that works with the WYSIWYG editor, that'd be lovely.

dragomano:
The modern solutions have the tendency to abandon BBCode. For example, CKEditor 4 has BBCode addon, and perfect [Mentions/ Tags and Emoji](https://ckeditor.com/docs/ckeditor4/latest/examples/mentions.html) addon (3 in 1), but CKEditor 4 reached its End of Life (EOL) in June 2023. And CKEditor 5 doesn't have BBCode addon.

Sesquipedalian:
Sorry if I wasn't clear. We have no plans to get rid of SCEditor. I meant that if you can find a solution to the problem that tribute.js doesn't work with SCEditor's WYSIWYG mode, that would be great. If that isn't possible, then we'll need to stick with at.js.

jdarwood007:
@dragomano,  If you still plan this, a conflict exists.  But ensure that the issue you raised in #8820 doesn't return here.

live627:
- #7933  completely rewrites at.js to be  native  ECMAScript.

PR #8141, opened 2024-03-17
Only delete board_permissions_view if we have modified the access

live627:
- see #7802


jdarwood007:
There was a reason you couldn't use the empty array check and I believe it was for when you delete all access (admin only board).  I resorted to the isset check, knowing we are attempting to modify access, but have no access to set.

live627:
oh right

PR #8146, opened 2024-03-25
Show query errors

live627:
I got tired of seeing bad queries catastophically fail with a frustratingly obscure `TypeError`

Code is copied from 2.x ~because I lazy~ because it works.

### DIFF
--- a/Sources/Db/APIs/MySQL.php
+++ b/Sources/Db/APIs/MySQL.php
@@ -242,6 +242,30 @@ class MySQL extends DatabaseApi implements DatabaseApiInterface
 
 		$ret = @mysqli_query($connection, $db_string, self::$unbuffered ? MYSQLI_USE_RESULT : MYSQLI_STORE_RESULT);
 
+		if ($ret === false && empty($db_values['db_error_skip'])) {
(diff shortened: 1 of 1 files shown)


Sources/Db/APIs/MySQL.php
```diff
@@ -242,6 +242,30 @@ public function query(string $identifier, string $db_string, array $db_values =
 
 		$ret = @mysqli_query($connection, $db_string, self::$unbuffered ? MYSQLI_USE_RESULT : MYSQLI_STORE_RESULT);
 
+		if ($ret === false && empty($db_values['db_error_skip'])) {
+			list($file, $line) = $this->error_backtrace('', '', 'return', __FILE__, __LINE__);
+			$query_error = $this->error();
+
+			// Nothing's defined yet... just die with it.
+			if (empty(Utils::$context) || empty(Lang::$txt))
+				die($query_error);
```
```suggestion
			if (empty(Utils::$context) || empty(Lang::$txt)) {
				die($query_error);
			}
```

Keep PHP-CS-Fixer happy

Sesquipedalian (changes requested):

Sesquipedalian (approved):

PR #8149, opened 2024-03-27
Do not round query times

### LINKED ISSUE #8148: query times lose precision; verry fast one reports `0` seconds
1. enable the intl extension
2.. observe the query log with very fast queries
3. problematic query times in big circle
4. smaller circle shows slower queries still reports its time ok

Screenshot 2024-03-26 at 05-52-33 Board Number 3

live627:


Languages/en_US/General.php
```diff
@@ -1136,12 +1136,12 @@
 }.';
 $txt['debug_query_in_line'] = 'in {file} on line {line}, ';
 $txt['debug_query_which_took'] = '{0, plural,
-	one {which took # second}
-	other {which took # seconds}
+	one {which took {0, number, ::precision-unlimited} second}
+	other {which took {0, number, ::precision-unlimited} seconds}
```
Sesquipedalian, line 1140:
```suggestion
	one {which took {0, number, :: .000 second}
	other {which took {0, number, :: .000} seconds}
```

This will give exactly three decimal places in the output every time.

Languages/en_US/General.php
```diff
@@ -1136,12 +1136,12 @@
...
 }.';
 $txt['debug_query_which_took_at'] = '{0, plural,
-	one {which took # second}
-	other {which took # seconds}
+	one {which took {0, number, ::precision-unlimited} second}
+	other {which took {0, number, ::precision-unlimited} seconds}
```
Sesquipedalian, line 1144:
```suggestion
	one {which took {0, number, :: .000} second}
	other {which took {0, number, :: .000} seconds}
```

This will give exactly three decimal places in the output every time.

Sesquipedalian (changes requested):

Languages/en_US/General.php
live627: but 2.1 showed us all eight

Screenshot 2024-03-27 at 18-37-26 Screenshot

Sesquipedalian: Then never mind. 🙂

Sesquipedalian (approved):

PR #8150, opened 2024-03-27
Updates membersAllowedTo() to fetch members within moderator groups

live627:


Sesquipedalian:
Could you give a little more explanation of what this is meant to solve? The PR title makes sense in and of itself, but without explanatory context it is difficult to understand why the change is being made.

live627:
`User::membersAllowedTo()` has a glaring defect in that it wouldn't fetch moderator groups. Also unravels the sql query to be simpler and easier to understand with a few unions to retain intended behavior.

One slight change to the public facing call; passing `0` as the board id will fetch moderators from all boards.

Sesquipedalian (approved):

Sesquipedalian:
Thanks for the explanation. 🙂

PR #8156, opened 2024-03-29
Do not make up decimal places

### LINKED ISSUE #8155: `$significant_decimals` generates noise
1. disable `intl`
2. view the query log
3. chaos ensures now that it uses `unlimied-precision`

Screenshot 2024-03-28 at 16-42-06 My Community

5. the second image shows the result when `intl` is enabled again

Screenshot 2024-03-28 at 16-40-03 My Community

live627:
Lots of noise is generated when using the stem `unlimited-precision`

### DIFF
--- a/Sources/Localization/MessageFormatter.php
+++ b/Sources/Localization/MessageFormatter.php
@@ -626,7 +626,7 @@ class MessageFormatter
 			// Float precision format.
 			if (str_starts_with($stem, '.')) {
 				$significant_integers = strlen(strval(intval($number + 0)));
-				$significant_decimals = is_int($number + 0) ? 0 : max(0, strlen(rtrim(sprintf('%.53F', abs($number + 0)), '0')) - $significant_integers - 1);
+				$significant_decimals = (int) strpos(strrev(strval($number)), '.');
 
 				preg_match('/\.(0*)(#*)(\*?)/', $stem, $matches);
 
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Might as well change the code on this line, too.
https://github.com/SimpleMachines/SMF/blob/a182d918c1db9c6c87f84adfeaafdefe14453f40/Sources/Localization/MessageFormatter.php#L868

live627:
added, unsure if it's right or how to test if there is no off by one error.

Sources/Localization/MessageFormatter.php
```diff
@@ -865,7 +865,7 @@ protected static function applyNumberSkeleton(int|float|string $number, string $
 
 		// Ensure $number is a string.
 		if (is_float($number)) {
-			$precision = max(0, strlen(rtrim(sprintf('%.53F', abs($number + 0)), '0')) - strlen(strval(intval($number + 0))) - 1);
+			$precision = (int) strpos(strrev(strval($number)), '.') - strlen(strval(intval($number + 0))) - 1;
```
Sesquipedalian:
```suggestion
			$precision = (int) strpos(strrev(strval($number)), '.');
```

^ This is what you replaced `max(0, strlen(rtrim(sprintf('%.53F', abs($number + 0)), '0')) - strlen(strval(intval($number + 0))) - 1);` with in the other two cases, so I would think it is what we need in this case as well. Or am I missing something here?

Sesquipedalian (changes requested):

Sources/Localization/MessageFormatter.php
live627: Probably. I was unsure on this, so wanted your input.

Sesquipedalian (approved):

PR #8158, opened 2024-03-30
Implements SMF\Tasks\GenericTask

Sesquipedalian:
A class for running background tasks with custom callable functions. Similar to SMF\Tasks\GenericScheduledTask, except that this is for tasks that are not scheduled.

### DIFF
--- a/Sources/Tasks/GenericTask.php
+++ b/Sources/Tasks/GenericTask.php
@@ -1,0 +1,44 @@
+<?php
+
+/**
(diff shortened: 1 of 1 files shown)


live627:
Might want to pass `$details` to the `callable`

Sesquipedalian:
Done in c49445be066db5ed3879a950733b6217000e963e and a182d918c1db9c6c87f84adfeaafdefe14453f40

PR #8160, opened 2024-03-30
Shows search results and improves code separation between search APIs and search admin action

### LINKED ISSUE #8159: [3.0]: Search results are not showing
### Basic Information

Run any search, and the results are not displayed.

### Steps to reproduce

1. Run any search.

### Expected result

Search results are displayed.

### Actual result

Search results are not displayed.

### Version/Git revision

3.0 Alpha 1

### Database Engine

All

### Database Version

_No response_

### PHP Version

_No response_

### Logs

_No response_

### Additional Information

_No response_

Sesquipedalian:
Fixes #8159

Also disentangles methods and settings for specific search APIs from the code in SMF\Actions\Admin\Search.

### DIFF
--- a/Sources/Search/APIs/Custom.php
+++ b/Sources/Search/APIs/Custom.php
@@ -17,8 +17,13 @@ namespace SMF\Search\APIs;
 
 use SMF\Config;
 use SMF\Db\DatabaseApi as Db;
+use SMF\Lang;
+use SMF\Menu;
+use SMF\Sapi;
 use SMF\Search\SearchApi;
 use SMF\Search\SearchApiInterface;
+use SMF\SecurityToken;
+use SMF\User;
 use SMF\Utils;
 
 /**
@@ -27,6 +32,24 @@ use SMF\Utils;
  */
 class Custom extends SearchApi implements SearchApiInterface
 {
+	/**************************
+	 * Public static properties
+	 **************************/
+
+	/**
+	 * @var array
+	 *
+	 * Sub-actions to add for SMF\Actions\Admin\Search::$subactions.
+	 */
+	public static array $admin_subactions = [
+		'createmsgindex' => __CLASS__ . '::build',
+		'removecustom' => __CLASS__ . '::remove',
+	];
+
+	/*********************
+	 * Internal properties
+	 *********************/
+
 	/**
 	 * @var array Index settings
 	 */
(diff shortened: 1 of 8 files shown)


Sources/Search/APIs/Custom.php
```diff
@@ -337,6 +364,377 @@ public function postModified(array &$msgOptions, array &$topicOptions, array &$p
 			}
 		}
 	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public function postRemoved(int $id_msg): void
```
live627, line 373: this should be added to `supportsMethod()`

Sesquipedalian: Good catch

live627: Looking further into this confusing mess, tit appears as if it returns  false, it bubbles up to the parent

https://github.com/SimpleMachines/SMF/blob/b11d8c4178b3d8e18a5243c33174524d216b8e4e/Sources/Search/APIs/Custom.php#L91-L99

This interface violates the Interface Segregation Principle. Perhaps we can break out the post and the topic methods into their own interfaces.

Sources/Search/APIs/Custom.php
```diff
@@ -27,6 +32,24 @@
...
+
+	/**
+	 * @var array
+	 *
+	 * Sub-actions to add for SMF\Actions\Admin\Search::$subactions.
+	 */
+	public static array $admin_subactions = [
+		'createmsgindex' => __CLASS__ . '::build',
+		'removecustom' => __CLASS__ . '::remove',
+	];
```
live627, line 47: Can this be a method?

I have these three methods in my  custom action interface. Perhaps they could be of use.

	public function addSubAction(string $sa, callable $cb): void;
	public function hasSubAction(string $sa): bool;
	public function callSubAction(string $sa): void;

Sources/Search/APIs/Custom.php
Sesquipedalian: > confusing mess

Indeed. The search APIs date way back to the early days of SMF 2.1, and they were less than rigorously structured, to say the least. This PR is working toward cleaning up the mess, but there's still going to be more to do after this.

Sources/Search/APIs/Custom.php
Sesquipedalian: 🤔 Maybe. Those methods do look interesting. 

Since you already have code and an idea for this, how about I just leave it to you to make a PR implementing that idea at your leisure after this is merged?

PR #8165, opened 2024-04-01
Fixes return types for various Utils methods

Sesquipedalian:
In particular, methods that internally use preg functions could return false.


jdarwood007:
FYI, the Return type of false is PHP 8.2+.  Which is why it wasn't used.
https://www.php.net/manual/en/language.types.declarations.php

Sesquipedalian:
Actually, false and null can be used as return types in 8.0 if they are part of a union type. 8.2 added the ability for them to be standalone return types.

jdarwood007:
Ahh, good to know.

PR #8167, opened 2024-04-03
More improvements to search API

Sesquipedalian:
1. Adds SearchApiInterface::getSize() method and moves code for getting the size of each API's index into its class. This helps to further disentangle the APIs from the administrative code in SMF\Actions\Admin\Search.
2. Adds SearchApiInterface::getStatus() method, which simplifies the process of keep track of whether an API is ready to be used, etc.
3. In SearchApi::detect(), includes the loaded instance of each API in the returned array of info.
4. Labels and descriptions for each API are now accessible via SearchApiInterface::getLabel() and SearchApiInterface::getDescription().
5. As requested by @live627, replaces $admin_subactions property used by some APIs with a new SearchApiInterface::getAdminSubactions() method.
6. Template improvements. We no longer need special templates for the fulltext and custom indexes.


Sources/Search/APIs/Custom.php
```diff
@@ -70,6 +71,13 @@ class Custom extends SearchApi implements SearchApiInterface
 	 */
 	protected $supported_databases = ['mysql', 'postgresql'];
 
+	/**
+	 * @var in
```
live627:
```suggestion
	 * @var int
```

Sources/Search/APIs/Fulltext.php
```diff
@@ -62,6 +63,13 @@ class Fulltext extends SearchApi implements SearchApiInterface
 	 */
 	protected $supported_databases = ['mysql', 'postgresql'];
 
+	/**
+	 * @var in
```
live627:
```suggestion
	 * @var int
```

live627:
ran a search in the forum for `lorem ipsum` and got crash.

```
<b>Fatal error</b>:  Uncaught Error: Class &quot;SMF\Search\BBCodeParser&quot; not found in C:\wamp64\www\smf-dev\Sources\Search\SearchApi.php:1004
Stack trace:
#0 C:\wamp64\www\smf-dev\Sources\Search\SearchApi.php(565): SMF\Search\SearchApi-&gt;setBlacklistedWords()
#1 C:\wamp64\www\smf-dev\Sources\Actions\Search2.php(94): SMF\Search\SearchApi-&gt;initializeSearch()
#2 C:\wamp64\www\smf-dev\Sources\Actions\Search2.php(223): SMF\Actions\Search2-&gt;execute()
#3 [internal function]: SMF\Actions\Search2::call()
#4 C:\wamp64\www\smf-dev\Sources\Forum.php(263): call_user_func(Array)
#5 C:\wamp64\www\smf-dev\index.php(148): SMF\Forum-&gt;execute()
#6 {main}
  thrown in <b>C:\wamp64\www\smf-dev\Sources\Search\SearchApi.php</b> on line <b>1004</b><br />

**Comment by Sesquipedalian** [comment]
Derp. Will push a fix shortly

**Comment by live627** [comment]
I attempted to create a custom  index, and after a long time,  it ended in error


```
<b>Fatal error</b>:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 12584025 bytes) in <b>C:\wamp64\www\smf-dev\Sources\Db\APIs\MySQL.php</b> on line <b>710</b><br />
<br />
<b>Fatal error</b>:  Uncaught TypeError: substr(): Argument #1 ($string) must be of type string, bool given in C:\wamp64\www\smf-dev\Sources\ErrorHandler.php:88
Stack trace:
#0 C:\wamp64\www\smf-dev\Sources\ErrorHandler.php(88): substr(false, -2)
#1 C:\wamp64\www\smf-dev\Sources\ErrorHandler.php(141): SMF\ErrorHandler-&gt;__construct(2, 'Unknown: Cannot...', 'Unknown', 0)
#2 [internal function]: SMF\ErrorHandler::call(2, 'Unknown: Cannot...', 'Unknown', 0)
#3 {main}
  thrown in <b>C:\wamp64\www\smf-dev\Sources\ErrorHandler.php</b> on line <b>88</b><br />

Sesquipedalian:
That's odd. I don't believe that I made any significant changes to the code that builds the custom index. I just moved it to a different file. Does that happen with the version currently in release-3.0?

live627:
I can retest later today.

Sesquipedalian:
Please do. :)

Sesquipedalian:
Actually, this PR doesn't even touch the code to build the custom index. That was moved from SMF\Actions\Admin\Search to SMF\Search\APIs\Custom in #8160, so you will need to test with f1d7263d77d955f61908def2a36bb52ea3c6b958.

live627:
> 4. Labels and descriptions for each API are now accessible via SearchApiInterface::getLabel() and SearchApiInterface::getDescription().

IMO, only  controllers (and templates) should know about language files. Will also make for easier time in the future whenever we decide to use dependency injection, the D in SOLID where we get PSR-11 compliance.

Sesquipedalian:
> > 4. Labels and descriptions for each API are now accessible via SearchApiInterface::getLabel() and SearchApiInterface::getDescription().
> 
> IMO, only controllers (and templates) should know about language files. Will also make for easier time in the future whenever we decide to use dependency injection, the D in SOLID where we get PSR-11 compliance.

Fair point. I will change these methods so that they return an expected $txt key rather than the string itself. That way, the methods will give the object a way to say "Here's where I think my label/description should be found when someone else looks it up," rather than "Here's what my label/description is."

live627:
awesome, thanks

PR #8174, opened 2024-04-17
Improve query log display

### LINKED ISSUE #8170: First line of query gets cut off in query log
Queries in the view query action have their first few characters cut off. The ones at the bottom of every page are unaffected .
Screenshot 2024-04-05 at 05-54-38 My Community

I see two ways to fix this

1. fix the trimming algorithm
2. prepend newlines to each query in the code like in 2.x

live627:
- Add syntax highlighting in the dedicated area
- Show query analysis in JSON below the EXPLAIN table
- Fix algorithm to trim indent off queries
- Also trim queries in main pages when  enabled


Sesquipedalian (approved):
Very nice!

PR #8177, opened 2024-04-23
Implements recurring calendar events

### LINKED ISSUE #7935: Support recurring calendar events
Add support for recurring calendar events.

Sesquipedalian:
Closes #7935

This is a major upgrade to SMF's calendar system:

- Implements support for recurring events in accordance with [RFC 5545](https://datatracker.ietf.org/doc/html/rfc5545) (the iCalendar specification).
	- This is accomplished via changes to SMF\Actions\Calendar, SMF\Calendar\Event, and four new classes: SMF\Calendar\RRule, SMF\Calendar\RecurrenceIterator, SMF\Calendar\EventOccurrence, and SMF\Calendar\EventAdjustment.
	- Updates the UI to allow users to specify recurrence rules for events. The UI supports all the recurrence rule options that one might find in full-fledged iCalendar applications such as Apple's Calendar app or Google Calendar.
	- When a recurring event is linked to a topic, all individual occurrences of the event link to the same topic, and the topic always shows the next upcoming occurrence of the event (or the last occurrence if there are no more upcoming ones).

- Events, holidays, and birthdays are now all classes, with SMF\Calendar\Holiday and SMF\Calendar\Birthday being extensions of SMF\Calendar\Event.
	- Events and holidays are now stored in the same table and handled in almost all the same ways. The only difference between them from a database perspective is the value of the "type" column. This reflects the RFC 5545 approach, wherein holidays are just events.

- Adds support for importing iCalendar data from exteral sources. 
	- For example, admins can now import holiday information for their countries' holidays from iCalendar URLs with a couple of clicks.
	- Admins can similarly import events from an iCalendar URL whenever they like. 
	- SMF will even check for duplicates in the imported data, so reimporting a calendar multiple times is safe.
	- Additionally, it is possible to subscribe to an external calendar in order to update the imported data automatically via a scheduled task.

- Improves support for exporting iCalendar data. 
	- Fully complies with the iCalendar standard (we weren't before).
	- Adds support for exporting recurring events.
	- Adds support for exporting all currently visible events.
		- For example, in month view it is is now possible to download all events in the month being viewed.
	- Adds support for subscribing to the forum's calendar in a third party calendar app.

- Includes support for "special recurrence rules" that call custom PHP functions to calculate recurrence rules that are too complicated even for RFC 5545 to model. 
	- This is used in vanilla SMF to calculate the date of Easter (whether Western or Eastern). 
	- It could also be used by mods to add support for holidays that are calculated according to non-Gregorian calendars—although it is probably easier to simply import such data from an external holiday calendar.

Other changes include:

- Makes Time::isStrftimeFormat() public.
- Implements SMF\TimeInterval, which extends \DateInterval with some extra features for SMF.
- Kills off the deprecated 'span' event property.
- Moves SMF\Event to SMF\Calendar\Event.
- Supports attaching multiple events to one topic. (Nearly all of the underlying code for this was already there; I just put some finishing touches on it and exposed it in the UI.)
- Uses HTML5 date and time inputs throughout the calendar UI.
- Improves the code to export events in iCal format.
- Moves calendar strings to their own file.
- Adds new code to the installer and upgrader files in order to implement the new database structures required to support recurring events.

### UI for creating recurring events
Screenshot 2024-04-29 at 6 24 51 PM

Screenshot 2024-04-29 at 6 22 05 PM

Screenshot 2024-04-29 at 6 23 39 PM

Screenshot 2024-04-29 at 6 23 55 PM

### Recurrence info is included when displaying events in a topic or in list view
Screenshot 2024-04-29 at 6 28 29 PM

### Improved export options
Screenshot 2024-04-29 at 6 30 30 PM

### Admin panel for importing (and optionally subscribing to) external iCalendar data
Screenshot 2024-04-29 at 6 35 55 PM


Sesquipedalian:
To heck with it. I've tested this PR a ton, and I am getting tired of resolving conflicts...

PR #8182, opened 2024-04-25
Performs autolinking in the editor

### LINKED ISSUE #7937: Do autolinking during post composition, not during post display
The continuing proliferation of TLDs is causing a steady increase in complaints about unexpected links in text. For example, writing [files.zip](https://files.zip/) in a post now creates a link, because zip is now a TLD.

We could monkey around with the autolinker's regex in order to try to avoid such situations, but that's a perpetually losing game of catch-up. Moreover, even if we tried that, we could never get the right results for every situation because we simply can't read the user's mind. We need a different approach.

The [Universal Acceptance Steering Group](https://www.icann.org/en/blogs/details/universal-acceptance-steering-group-a-year-of-progress-26-10-2016-en), "a community-based team of industry leaders supported by ICANN," published some [recommendations regarding autolinking](https://uasg.tech/download/uasg-010-quick-guide-to-linkification-en/) a few years ago. The most relevant part for our purposes is their recommendation that autolinking should happen during the input stage, not the output stage.

Currently, SMF does its autolinking during the output stage (specifically, while parsing BBC). This creates three closely related issues:

1. Users will sometimes see unexpected links in their posts after submitting them, without any warning that this would happen during the authoring phase.
2. Users have no easy means of removing unexpected links from their posts.
3. New links can appear in old posts when new TLDs are added by ICANN.

All of these problems would be solved if SMF did its autolinking during the input stage. This would mean creating our own custom plugin for SCEditor that would detect plain text URLs and automatically wrap them in url BBC tags. The user would therefore see the link being created on the fly during the authoring phase, which would solve issue 1. If the user didn't want a particular string to be treated as a link, they would then be able to select that string and remove the link using the editor's toolbar button, which would solve issue 2. Moreover, because submitted URLs would now always be explicitly tagged using the url BBC, we could safely assume that untagged URLs in posts were not supposed to be linked at all, which would solve issue 3.

The one wrinkle in this plan would be the question of how to deal with plain text URLs in old posts that were written with the expectation of SMF's current autolinking behaviour. The possible solutions I see are:

1. We record the date when the forum was upgraded to SMF 3.0, and then only perform autolinking during the output phase if the post was last modified before that date.
2. We do in fact monkey around with the output-phase autolinker code to make it much more restrictive (e.g. only recognizing basic TLDs), so that our fallback when dealing with plain text URLs is very conservative in what it links.
3. We do both (1) and (2).

Sesquipedalian:
Fixes #7937 

With these changes, autolinking of URLs and email addresses now happens in the editor while the user is composing a post or PM, rather than during the output phase when the post is being prepared for display.

- The autolinking code has been moved out of SMF\BBCodeParser into its own class, SMF\Autolinker.
- A new SCEditor plugin has been added that takes care of the JavaScript for autolinking in the editor. It works in both WYSIWYG and source mode.
- A new `[nolink]` BBCode has been added in order to allow users to actively prevent certain strings from becoming links. Its only function is to mark a string of text as an area where no autolinking should ever happen. Users won't normally need to type this manually; it is inserted when the user clicks the "unlink" toolbar button (see below).
- An "unlink" toolbar button has been added. SCEditor actually includes a version of this button by default, but we weren't showing it for some reason. At any rate, our new version not only unlinks an existing link, but moreover inserts a `[nolink]` BBCode around the string that was previously linked so that the autolinker knows that it shouldn't try to link it again.
- Where possible, the editor will prefer using the simple forms of the `[url]`, `[iurl]`, and `[email]` BBCodes.
- SMF\Msg::preparsecode() now validates `[url]`, `[iurl]`, and `[email]` BBCodes to make sure they are syntactically valid. If they aren't, it attempts to fix them (or removes them on failure). Additionally, if this happens when the user is trying to submit a post or PM, SMF will go back to the posting page and ask the user to confirm that the changes are correct.
- SMF\Msg::preparsecode() has been given a new `$autolink` parameter that tells the method to perform autolinking of plain-text URLs. This is used for cases where input is provided in plain text only (e.g. when entering news items, warning messages, topic moved reasons, etc.), because without a full-fledged editor there is no opportunity to do autolinking while the user is writing.
- Version info has been added for posts and PMs (as suggested by @jdarwood007 in internal discussion). This includes both the addition of a `version` column to the relevant database tables, and the addition of a `$this->version` property to the relevant classes. 
	- This info allows us to decide whether to perform autolinking during the output phase for posts and PMs that are being displayed. Specifically, we only do so for posts and PMs that were written for earlier versions of SMF. 
		- We do this in order to maintain the behaviour that was expected when that old content was written.
		- During output for this old content we only autolink URLs that use "basic" TLDs (meaning, TLDs from the 2012 list, before ICANN started adding all the new gTLDs). This allows us to futureproof the old content against unexpected new gTLDs, thereby avoiding the [files.zip](https://files.zip/) problem mentioned in #7937.
		- If old content is edited using 3.0, the version will updated when the content is saved. This makes sense because editing the content in 3.0 means it has now had the opportunity to be autolinked in the editor.
	- @jdarwood007 has further ideas about ways that this version info could be used, so it will be helpful for more than just this specific application.
- Also fixes a minor bug in the editor where the toolbar buttons wouldn't initially be active.


Sources/Actions/Profile/ShowPosts.php
```diff
@@ -859,6 +860,12 @@ protected function loadPosts(bool $is_topics = false): void
 			Lang::censorText($row['body']);
 			Lang::censorText($row['subject']);
 
+			// Old SMF versions autolinked during output rather than input,
+			// so maintain expected behaviour for those old messages.
+			if (version_compare($this->version, '3.0', '<')) {
```
live627: I don't see this property defined at a very cursory glance. And it  looks like this isn't the only one that looks out of place.

Sesquipedalian: Oof. That should be `$row['version']`.

Sesquipedalian: Fixed.

live627:
some possible errors 

https://scrutinizer-ci.com/g/live627/SMF2.1/inspections/ab50bbfc-9da9-47d9-a5dc-f43e4f342b54/issues/files/Sources/Autolinker.php?status=new&orderField=path&order=asc&honorSelectedPaths=0

live627:
URIs posted into messages via the source editor do not turn  into links.

Sources/Autolinker.php
```diff
@@ -0,0 +1,1251 @@
...
+	{
+		// An &nbsp; right after a URL can break the autolinker
+		if (strpos($string, '&nbsp;') !== false) {
+			$string = strtr($string, ['&nbsp;' => str_repeat(html_entity_decode('&nbsp;', 0, $this->encoding), 3)]);
+		}
+
+		$this->setUrlRegex();
+
+		if ($plaintext_only) {
+			$no_autolink_regex = Utils::buildRegex(self::$no_autolink_tags);
```
live627: should this be a static variable for reuse when parsing multiple messages?

Sesquipedalian: Good idea.

Sesquipedalian:
> some possible errors
> 
> https://scrutinizer-ci.com/g/live627/SMF2.1/inspections/ab50bbfc-9da9-47d9-a5dc-f43e4f342b54/issues/files/Sources/Autolinker.php?status=new&orderField=path&order=asc&honorSelectedPaths=0

Some of those are legitimate, some aren't. 

In particular, "The variable $bbc does not seem to be defined for all execution paths leading up to this point" is incorrect. Scrutinizer just doesn't quite understand what is happening in that loop.

Sesquipedalian:
> URIs posted into messages via the source editor do not turn into links.

They do once you press space or return after them.

live627 (approved):

PR #8184, opened 2024-04-26
Improves on custom search index

### LINKED ISSUE #8175: [3.0]: Links not rendered in search results
### Basic Information

The post link and member link aren't rendered in the search results.

### Steps to reproduce

1. Search for something
2. Look at the results

### Expected result

Expect to see correct post_link and member_link

### Actual result

You see "{post_link} by {member_link}" instead
image

### Version/Git revision

a540d9e

### Database Engine

MySQL

### Database Version

10.11.7-MariaDB-1:10.11.7+maria~ubu2204

### PHP Version

8.3.6

### Logs

_No response_

### Additional Information

_No response_

### LINKED ISSUE #8176: [3.0]: Custom search index weirdness
### Basic Information

While playing around with the search the other day trying to reproduce #7870 I discovered some weird issues with the custom search index - sometimes it doesn't return any results when it should and other times it doesn't return the expected results.

I have the following on my board:
A post with the word "testing"
A post with the phrase "trying to reproduce the error"
A post that quotes the "error" post (same topic)

Search tests:
Searching for "testing" - no results
Searching for "trying" - no results
Searching for "reproduce" or "error" - single result in the quote (but not the post that is being quoted)

### Steps to reproduce

1. Create some topics/posts
2. Create a custom search index
3. Attempt to search for various words within the posts/topics created

### Expected result

Expect to find all posts/topics containing the word searched for

### Actual result

Inconsistent/unexpected results (see above)

### Version/Git revision

a540d9e

### Database Engine

MySQL

### Database Version

10.11.7-MariaDB-1:10.11.7+maria~ubu2204

### PHP Version

8.3.6

### Logs

_No response_

### Additional Information

The MariaDB version doesn't seem to have any play in this. I initially ran into these issues while on MariaDB 10.6 but upgraded to MariaDB 10.11 last night (not because of this problem but because Plesk now has a feature to upgrade to the latest LTS version).
The size of the custom search index doesn't seem to make a difference here either, though this forum only has 6 posts in 3 topics so I doubt there's much if any difference between the actual index size regardless of which option you choose when creating the index.
This occurs both with the basic search and the advanced search.

Sesquipedalian:
- Fixes #8175
- Fixes #8176
- Fixes an unreported bug where log_search_subject could end up with invalid HTML entities in it.
- Replaces Utils::text2words() with simpler, discrete methods.
	- SMF\Utils::extractWords() handles the use case where the caller wants to get the individual words from a string.
	- SMF\Search\APIs\Custom::getWordNumbers() handles the use case, which is specific to the custom search API, where the caller wants to get an array of integer values for the unique words in a string.
- Tweaks SMF\Unicode\Utf8String::sanitizeInvisibles() in order to ignore Emoji_Modifier characters in search strings, the word censor, etc. This means, for example, that searching for `👩` will match any of `👩`, `👩🏻`, `👩🏼`, `👩🏽`, `👩🏾`, or `👩🏿`.
- Improves SMF\Search\SearchResult::highlight() so that it can highlight matched phrases even when the whitespace, punctuation marks, etc., between the words of the phrase are a bit different than what the user originally entered.
	- For example, if the user searched for `eats shoots and leaves`, we now successfully highlight the string `eats, shoots, and leaves` in matching results.
- Moves some logic out of the templates into Sources, which allows search APIs to have some control over what is shown in the search UI forms.
	- For example, if a particular search API wants to add some addtional options to the search form, it can. (See below.)
- ### Deprecates SMF\Search\APIs\Custom and implements a new, better API: SMF\Search\APIs\Parsed.
	- This new API is more efficient than the old custom index.
		- Whereas the old custom index only served to narrow down the list of possibly matching messages and still had to check the body text of each possibly matching message in order to find the actual matches, the parsed index is fully authoritative in itself and can get the exact set of matches without needing to check the messages table.
	- This new API is more precise than both the fulltext index and the old custom index.
		- Both the fulltext index and the old custom index are built using the raw BBCode text of the message body. In contrast, the parsed index uses the parsed BBCode (thus the name). This means that the search results fully align with what the user actually sees when reading the matching posts, and that there are no weird side effects from indexing raw BBCode tags and the like.
	- This new API supports accent-sensitive *and* accent-insensitive search, at the user's discretion.
		- For example, a user searching for the Spanish word `peña` would not want matches for `pena` to show up in the results, but a user searching for `cliché` probably would want matches for `cliche` as well. The "Ignore accents, etc." search option allows users to choose the behaviour they want on a case-by-case basis.
		- The default value for the "Ignore accents, etc." search option is determined on a per-language basis, so that languages that don't normally use accents (e.g. Englsh) can have that option turned on by default, while languages that normally do use accents (e.g. French, Spanish) can have it turned off by default.
	- This new API supports glob style wildcards (i.e. `*` and `?`), allowing the user more flexibility than the admin-level "Match whole words only" setting could otherwise permit.
		- The structure of the parsed index means that this support comes at virtually no additional cost to performance.
		- Wildcards are only supported when the admin-level "Match whole words only" setting is turned on, because turning that option off is tantamount to wrapping every word in the search query with leading and trailing `*` wildcards anyway.
	- The old custom index is still supported, but the UI now shows a message advising the admin to upgrade to a parsed index. Moreover, the old custom index will only be shown in the list of available options either if it is currently in use, or else if Config::$db_show_debug is true.


Sesquipedalian:
I've added another commit that takes advantage of the [newly introduced Autolinker class](https://github.com/SimpleMachines/SMF/pull/8182) in order to improve the handling of URLs in the parsed search index. URL schemes (e.g. `http`, `https`, etc.), `www.`, and basic gTLDs (e.g. `.com`, `.org`, etc.) are all removed from the URL before indexing and/or searching, because those parts are generally just noise for search purposes. New gTLDs  (e.g. `.community`, `.ninja`, etc.) and ccTLDs (e.g. `.ca`, `.nz`, etc.) are not removed, however, because those are uncommon enough to still be useful in search results.

Sesquipedalian:
@Oldiesmann, can you confirm that this PR fixes the problems you were having while using the custom index? Even though the new parsed index deprecates the old custom index, we still need to fix the latter.

PR #8186, opened 2024-04-26
Report

live627:
- The Report controller uses the  new objects instead of raw SQL queries. Insert reference to Pinocchio saying "I'm a Real Boy now!"
- Add a new report type for board access. Well  actually it was moved from the board report in an attempt to be more accessible.
- Use JavaScript to split long tables, increasing readability and eliminating the need to scroll vertically . Who wants to do that?

I also have some code that adds checkboxes to filter the results by category mainly useful for board permissions. Shall I add a commit for that?


Sources/Actions/Admin/Reports.php
```diff
@@ -335,11 +300,11 @@ public function boards(): void
 			$boardData['groups'] = implode(', ', $allowedGroups);
 
 			if (!empty(Config::$modSettings['deny_boards_access'])) {
-				$disallowedGroups = explode(',', $row['deny_member_groups']);
+				$disallowedGroups = explode(',', $board->deny_member_groups);
```
Sesquipedalian: This line causes PHP to die with the following fatal error:
```
Fatal error: Uncaught TypeError: explode(): Argument #2 ($string) must be of type string, array given in /path/to/forum/Sources/Actions/Admin/Reports.php:303 Stack trace: #0 /path/to/forum/Sources/Actions/Admin/Reports.php(303): explode(',', Array) #1 [internal function]: SMF\Actions\Admin\Reports->boards() #2 /path/to/forum/Sources/Actions/Admin/Reports.php(194): call_user_func(Array) #3 /path/to/forum/Sources/Actions/Admin/Reports.php(642): SMF\Actions\Admin\Reports->execute() #4 [internal function]: SMF\Actions\Admin\Reports::call() #5 /path/to/forum/Sources/Actions/Admin/ACP.php(786): call_user_func(Array) #6 /path/to/forum/Sources/Actions/Admin/ACP.php(813): SMF\Actions\Admin\ACP->execute() #7 [internal function]: SMF\Actions\Admin\ACP::call() #8 /path/to/forum/Sources/Forum.php(263): call_user_func(Array) #9 /path/to/forum/index.php(146): SMF\Forum->execute() #10 {main} thrown in /path/to/forum/Sources/Actions/Admin/Reports.php on line 303
```

Themes/default/scripts/reports.js
```diff
@@ -0,0 +1,53 @@
...
+	for (let i = 1; i < els.length; i++) {
+	console.log(els[i].getBoundingClientRect().right > curPageWidth,els[i].getBoundingClientRect().right,curPageWidth) 
+		if (els[i].getBoundingClientRect().right > curPageWidth) {
+			positions.push(i - 1);
+			curPageWidth += pageWidth - pos;
+		}
+	}
+
+	//~ positions.push(positions[positions.length - 1] + positions[positions.length - 2]);
+	//~ return;
```
Sesquipedalian: Delete?

Sesquipedalian (dismissed):

Sesquipedalian:
> I also have some code that adds checkboxes to filter the results by category mainly useful for board permissions. Shall I add a commit for that?

Sounds nice. Yes, please!

Sources/Actions/Admin/Reports.php
live627: fixed

live627:
added those extra checkboxes

Themes/default/Reports.template.php
```diff
@@ -66,11 +66,39 @@ function template_main()
 	echo '
 		</div>';
 
-	// Go through each table!
-	foreach (Utils::$context['tables'] as $table)
+	if (count(Utils::$context['tables']) > 1)
 	{
 		echo '
-		<table class="table_grid report_results">';
+		<style>';
```
Sesquipedalian, line 72:
```suggestion
		<script>';
```

Themes/default/Reports.template.php
```diff
@@ -66,11 +66,39 @@ function template_main()
...
+
+		foreach (Utils::$context['tables'] as $i => $table)
+			echo '
+			body:has(#report_check_', $i + 1, ':not(:checked)) #report_', $i + 1, ' {
+				display: none;
+			}';
+
+	echo '
+		</style>
```
Sesquipedalian, line 81:
```suggestion
		</script>
```

Sesquipedalian (changes requested):
1. The Board Access report uses the wrong title in the cat_bars.
3. There is a useless duplicate table for each board in the Boards report. 
Screenshot 2024-04-30 at 2 49 17 PM Perhaps some more work needs to go into the logic that breaks up wide tables?
4. The checkboxes don't do anything. I suggested changing `<style>` to `<script>` below, which is necessary. However, even after I did that in my own tests, the JavaScript ran into syntax errors that will need to be fixed.

Themes/default/Reports.template.php
live627: thesee are actually css rules inspired by https://codepen.io/jlengstorf/pen/YzMwNrp

Sesquipedalian (approved):

PR #8187, opened 2024-04-27
[3.0] Change likes to reactions...

Oldiesmann:
This PR changes SMF's likes feature into a full-fledged reactions system.

Features:
Admins can add/remove reactions as desired (only limit is defined by the limits of the SMALLINT data type, which should be plenty)
For simplicity, it will use images in a reactions folder in the theme's images folder and will look for a $txt['react_{name}'] string for i18n (defaults to 'name' defined in the database if not found)

Work done so far:
- Database changes for install and upgrade
- Modified language strings

Still to do:
- Code modifications/additions to handle things
- Theme stuff
- Testing (this is alpha so not too worried about that yet)

Note that the upgrade uses a combination of db API functions and upgrade_query. In some situations (such as creating a table), upgrade_query is far simpler than the associated db API call.


Oldiesmann:
Also cc @Sesquipedalian since this will require changes to the DB schema stuff in #8093

asmith20002:
Good job. IMO this also should be done for 2.1

live627:
2.1 is feature frozen and receives bug fixes only.

Sources/Actions/Admin/Reactions.php
```diff
@@ -285,6 +287,39 @@ public function editreactions(): void
...
+				'position' => 'bottom_of_list',
+				'data' => [
+					'value' => '<input type="text" name="reacts_add[]">'
+				]
+			],
+			[
+				// Clicking this magic link adds a new row...
+				'position' => 'bottom_of_list',
+				'data' => [
+					'value' => '<a href="javascript:void(0);" onclick="addrow()">' . Lang::$txt['reacts_add'] . '</a>'
```
DiegoAndresCortes: Can you make this a <button type="button" instead so we don't have to come back to it when fixing for wcag sutff

Oldiesmann: Sure. I'm just writing up basic code at this point and haven't even tested any of this yet (I plan to start that soon so I don't end up with a ton of stuff to fix later)

other/install_3-0_MySQL.sql
```diff
@@ -991,6 +991,17 @@ CREATE TABLE {$db_prefix}qanda (
...
 ) ENGINE={$engine};
 
+#
+# Table structure for table `reactions`
+#
+
+CREATE TABLE {$db_prefix}reactions (
+	id_reaction SMALLINT UNSIGNED AUTO_INCREMENT,
+	name VARCHAR(255) NOT NULL DEFAULT '',
+	order SMALLINT UNSIGNED NOT NULL DEFAULT '0',
```
DiegoAndresCortes:
```suggestion
	`order` SMALLINT UNSIGNED NOT NULL DEFAULT '0',
```

DiegoAndresCortes (changes requested):

other/install_3-0_MySQL.sql
Oldiesmann: I dropped that column entirely for the time being. I was considering adding functionality to choose the order of the reactions but forgot about that when I wrote all the admin code and didn't want to go to the trouble of adding that in as well. I'll do that if I add it back at some point though.

Sesquipedalian:
This is still marked as a draft, @Oldiesmann. Is it ready for review?

Oldiesmann:
Not yet. The back end stuff is done but I haven't done anything on the user side yet because theme-related stuff isn't my strong point. It's at a point now where the admin can add, remove and modify reactions as well as choose who can react to things, but I don't have any of the UI stuff done - the part that actually allows someone to react to something.

jdarwood007:
I would suggest putting something in that looks decent and letting the theme PR take care of improving it then.

Oldiesmann:
> I would suggest putting something in that looks decent and letting the theme PR take care of improving it then.

@tyrsson is going to help me out with the UI part when he gets a chance.

tyrsson:
> > I would suggest putting something in that looks decent and letting the theme PR take care of improving it then.
> 
> @tyrsson is going to help me out with the UI part when he gets a chance.

I hope to get started on it in the next couple weeks. It's on the schedule. The markup really is not the part that will take some doing it will be the JS required to make it really usable that needs sorting out.

Oldiesmann:
@DiegoAndresCortes did you mean to request changes? If so, what do you suggest? GitHub only says that you "requested changes with read-only permissions"

DiegoAndresCortes:
No recently, only the two older ones where you commented

Oldiesmann:
Going to go ahead and mark this as ready for review. At the very least we can merge the initial backend functionality in, which will allow others to contribute without having to fork my branch.

This currently has all the functionality to allow admins to manage reactions - enable/disable them, choose who can use them and add/edit/delete them. The user-facing stuff doesn't work yet as I'm not good with UI stuff at all. I have designed it with i18n in mind as well. The plan is that, in the UI area, we can use `$txt['react_{name}']` for the reaction name, or just fall back to the name from the DB if no such text string exists.

tyrsson:
Sorry @Oldiesmann I have just not been able to get to this :( As you know from our Discord convo's I've got a lot of irons in the fire at the moment.

Sources/Maintenance/Migration/v3_0/HolidaysToEvents.php
```diff
@@ -679,4 +679,4 @@ public function execute(): bool
 
 		return true;
 	}
-}
+}
```
jdarwood007: Revert this change

Sources/Maintenance/Migration/v3_0/UserReactions.php
```diff
@@ -0,0 +1,161 @@
...
+
+	/**
+	 *
+	 */
+	public string $name = 'Converting likes to reactions';
+
+	/****************
+	 * Public methods
+	 ****************/
+
```
jdarwood007, line 36:
```suggestion
	public function isCandidate(): bool
	{
		// Make sure we haven't already done this...
		$cols = Db::$db->list_columns('messages');

		// If the reactions column exists in the messages table, there's nothing to do
		return !in_array('reactions', $cols))
    }

```

If you add this check, you can remove it from the execute() function.  The upgrader will indicate "skip" when it comes across it and isCandidate() returns false.

Sources/Maintenance/Migration/v3_0/UserReactions.php
```diff
@@ -0,0 +1,161 @@
...
+		
+		// If the reactions column exists in the messages table, there's nothing to do
+		if(!in_array('reactions', $cols))
+		{
+			// Does the user_likes table exist?
+			$table_exists = Db::$db->list_tables(false, '%user_likes');
+			if (!empty($table_exists))
+			{
+				// Rename the table
+				Db::$db->query('', 'ALTER TABLE {db_prefix}user_likes RENAME TO {db_prefix}user_reacts');
```
jdarwood007: @Sesquipedalian 
This isn't cross DB safe, but we don't have a method to rename a table.  It seems like we should have such logic.

Sources/Maintenance/Migration/v3_0/UserReactions.php
```diff
@@ -0,0 +1,161 @@
...
+				
+				// Add the new column
+				Db::$db->add_column('{db_prefix}user_reacts', ['name' => 'id_reaction', 'type' => 'smallint', 'not_null' => true, 'default' => '0']);
```
jdarwood007: Ensure that you have a use statement to `UserReacts` and you could do:
```
$tbl = new UserReacts();

Db::$db->add_column('{db_prefix}user_reacts', $tbl->columns['id_reaction']);
```

Thus ensuring that you are asking it to do the changes as needed to match the schema.

Sources/Maintenance/Migration/v3_0/UserReactions.php
```diff
@@ -0,0 +1,161 @@
...
+
+				// Default reaction is like for now
+				Db::$db->query('', 'UPDATE {db_prefix}user_reacts SET id_react={int:one}', ['one' => 1]);
+
+				// Rename the like_time column
+				Db::$db->change_column('{db_prefix}user_reacts', 'like_time', ['name' => 'react_time']);
+
+				// Rename the index
+				Db::$db->query('', 'ALTER TABLE {db_prefix}user_reacts RENAME idx_liker TO idx_reactor');
```
jdarwood007: Use `Db::$db->rename_index`

Sources/Maintenance/Migration/v3_0/UserReactions.php
```diff
@@ -0,0 +1,161 @@
...
+
+		$reactions_idxs = [
+			'id_reaction' => [
+				'name' => 'id_reaction',
+				'columns' => ['id_reaction'],
+				'type' => 'primary'
+			]
+		];
+
+		Db::$db->create_table('{db_prefix}reactions', $reactions_cols, $reactions_idxs);
```
jdarwood007:
```suggestion
			$table = new Schema\v3_0\UserReacts();
			$table->create();
```

The upgrader will handle all the logic to setting up the table as long as the schema has all the correct defs.

Also you could change your entire check logic at the start to do if statement this way:
```
			$existing_tables = Db::$db->list_tables();

			if (!in_array(Config::$db_prefix . $table->name, $existing_tables)) {
				$table->create();
			} else {
			[....]
```

jdarwood007 (commented):
Just some upgrader comments

Sources/Maintenance/Migration/v3_0/UserReactions.php
Sesquipedalian: Yes, we should.

Sesquipedalian: Added in #8828

Sources/Maintenance/Migration/v3_0/HolidaysToEvents.php
Oldiesmann: I'm not sure what specifically I need to revert here. It looks like I just removed a } and added one back?

This file got messed up when I pulled in the changes from the install/upgrade stuff, Git was marking the file as "changed in both" even though it was new to my branch and I didn't touch it. I ended up copying the raw file from GitHub and replacing my version with that one because I couldn't figure out how to resolve the conflict otherwise.

jdarwood007: I think its the line ending that changed.

Oldiesmann: I'm not sure what I can do to fix that but I'll see.

Oldiesmann: Replaced it with the raw file from GitHub. Looks like it was an empty line at the end of the file that was missing, but this should be fixed now

Oldiesmann:
Fixed several bugs:
- Fixed admin center - fixed admin search referencing non-existent function; removed likes; added reactions menu item
- Fixed queries that still used the old format which had an optional identifier as the first parameter
- Fixed query in Stats that referenced a non-existent "reacts" column

Still need to update the Msg.php code to populate the variable which tracks how many of each reaction was chosen for a particular post

Languages/en_US/Profile.php
```diff
@@ -136,7 +136,7 @@
 $txt['alert_board_notify'] = 'When a new topic is created in a board I follow, I normally want to know via...';
 $txt['alert_msg_mention'] = 'When my @name is mentioned in a post';
 $txt['alert_msg_quote'] = 'When one of my posts is quoted';
-$txt['alert_msg_like'] = 'When one of my posts is liked';
+$txt['alert_msg_react'] = 'When someoe reacts to one of my posts';
```
tyrsson: Got a typo in the string. "someoe"

Oldiesmann:
@jdarwood007 @Sesquipedalian Can either of you explain what's causing the PHP-CS-Fixer check to fail here? It's not showing me an error and I can't get it to run locally (it just outputs whatever HTML is generated by SMF)

Sesquipedalian:
I see you are working to fix code style issue, @Oldiesmann. I recommend that you simply [install PHP-CS-Fixer](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer?tab=readme-ov-file#installation) and then use it to fix your files, like so:

```
php-cs-fixer fix --config=".php-cs-fixer.dist.php" --allow-risky="yes"
```

The `--allow-risky="yes"` bit is needed because we do use one rule that PHP-CS-Fixer classifies as "risky."

Oldiesmann:
> I see you are working to fix code style issue, @Oldiesmann. I recommend that you simply [install PHP-CS-Fixer](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer?tab=readme-ov-file#installation) and then use it to fix your files, like so:
> 
> ```
> php-cs-fixer fix --config=".php-cs-fixer.dist.php" --allow-risky="yes"
> ```

It doesn't work if I run it locally because of all the index.php files SMF sticks within vendor and its various subdirectories - it just ends up pulling in SMF's own index.php file and generates whatever output is generated from that.

Sesquipedalian:
Have you tried it recently? The .php-cs-fixer.dist.php file is configured to ignore anything inside the vendor dir.

Oldiesmann:
If I run it from the directory with the now-fixed files, it complains that Settings.php isn't readable, and if I copy Settings.php to the directory, it'll just spit out the HTML for the error message saying SMF can't connect to the database. If I run it from a live local test board, it spits out HTML.

image

image

Sesquipedalian:
Hm. You are correct. I will fix this soon.

Sesquipedalian:
Please see #9114 for the fix to your problem, @Oldiesmann.

Sesquipedalian:
@Oldiesmann, you need to make some changes to comply with the code standard.

Oldiesmann:
> @Oldiesmann, you need to make some changes to comply with the code standard.

I'll fix it when I get home.

Oldiesmann:
@Sesquipedalian All fixed now :smiley:

jdarwood007:
Is any of the TODO stuff still relevant?

Oldiesmann:
> Is any of the TODO stuff still relevant?
> 

Most of it is. I've added the functionality on the admin side (add/edit/remove/reorder reactions, enable/disable the feature and choose who can react) but none of the user side functionality is there there yet. @tyrsson is/was going to handle most of that at some point.

PR #8194, opened 2024-05-02
Fixes some upgrader issues

### LINKED ISSUE #8191: [3.0]: Upgrade doesn't work
### Basic Information

After fixing the upgrader to call Sapi::httpsOn() instead of Config::httpsOn() (or commenting out those three lines of code), the upgrader now just shows a blank line with a "Click here to try again" link

### Steps to reproduce

1. Apply changes from #8189 or comment out the three relevant lines of code
2. Attempt to upgrade your forum

### Expected result

Upgrade should run as normal

### Actual result

Upgrade shows but instead of being able to begin the process you get a blank line followed by "Click here to try again".
image

### Version/Git revision

492d519

### Database Engine

All

### Database Version

10.11.7-MariaDB-1:10.11.7+maria~ubu2204

### PHP Version

8.3.6

### Logs

_No response_

### Additional Information

Something is causing template_error_msg() to be loaded, but $upcontext['error_msg'] isn't set when the function is called. I haven't looked into it further than that.

Sesquipedalian:
Fixes #8191

@Oldiesmann, please test this to confirm.


Oldiesmann:
No change for me. Unfortunately, MariaDB doesn't support the "IF NOT EXISTS ()" syntax. It __does__ support "IF NOT EXISTS" as part of ADD COLUMN, but MySQL doesn't. Either we have to check for MySQL vs MariaDB or we have to use the database API instead (either check whether it exists with list_columns or just use add_column and tell it to ignore it if it already exists)

Sesquipedalian:
How about now, @Oldiesmann?

Oldiesmann:
Looks like we've been chasing the wrong bug all along. Still not working but running it via command line gives me a useful error. Why I don't get any errors when running it in the browser is beyond me, but at least this is getting us somewhere...
```
PHP Fatal error:  Uncaught TypeError: strpos(): Argument #1 ($haystack) must be of type string, false given in /.../Sources/QueryString.php:70
Stack trace:
#0 /.../Sources/QueryString.php(70): strpos()
#1 /.../upgrade.php(840): SMF\QueryString::cleanRequest()
#2 /.../upgrade.php(242): loadEssentialData()
#3 {main}
  thrown in /.../Sources/QueryString.php on line 70
```

Oldiesmann:
Unfortunately it's still not working. I do get an error when running it  via the shell, but that appears to be due to $_SERVER['QUERY_STRING'] not existing (since there is no query string when running it directly). Not getting any errors at all when running it via the browser, which is very annoying. It's not even showing me the login form, so I have no idea what the issue is.

Oldiesmann:
Finally got somewhere with this. After fixing the error with $_SERVER['QUERY_STRING'], I ran it on the shell again and got a message saying "Cannot find language files!".

I then went through upgrade.php and threw in a die() statement before each call to `throw_error()` and finally tracked it down to this code:
```php
	if (!file_exists(Config::$languagesdir . '/' . $upcontext['language'] . '/General.php')) {
		return throw_error(Lang::getTxt('error_lang_general_missing', ['lang' => $upcontext['language'], 'url' => $upgradeurl]));
	}
```

As you can see from this screenshot, the file in question most definitely exists.
image

Oldiesmann:
Further analysis... The problem is $upcontext['language'] is still being set to "english" and not "en_US", so it's never finding it because Languages/english/General.php doesn't exist.

Sesquipedalian:
That is very helpful sleuthing, @Oldiesmann. Thanks!

Sesquipedalian:
How about now, @Oldiesmann?

Oldiesmann:
It works fine now, though the cleanup step at the end fails because there's still a call to Config::httpsOn() in there somewhere that I missed when I fixed the other one.

Sesquipedalian:
> It works fine now, though the cleanup step at the end fails because there's still a call to Config::httpsOn() in there somewhere that I missed when I fixed the other one.

Great. I'v fixed those, and I will merge now.

PR #8196, opened 2024-05-05
Change the return type of ActionInterface to match that of ActionTrait

Oldiesmann:
phpStorm was complaining that the implementation of ActionInterface::load() wasn't compatible with that of ActionTrait::load(), so this fixes that. @Tyrsson can explain more.


tyrsson:
I'm pretty sure that this issue is due to the underlying changes in how late static binding happens within expressed traits when the concrete class is composed.

live627 (approved):

PR #8200, opened 2024-05-06
Replace hardcoded table prefix with {db_prefix}.

Oldiesmann:


Sesquipedalian (approved):
Oops. Thanks for catching that.

PR #8201, opened 2024-05-06
Fixes three things in SMF\Search\APIs\Parsed

Sesquipedalian:
1. Uses the database API to create tables for the parsed search index. This provides better support of PostgreSQL.
2. Fixes some backwards logic in SMF\Search\APIs\Parsed::accentSensitive().
3. Simplifies handling of blacklisted words in SMF\Search\APIs\Parsed. Specifically, instead of doing some of the work in the constructor, we now do it all in SMF\Search\APIs\Parsed::setBlacklistedWords().

### DIFF
--- a/Sources/Search/APIs/Parsed.php
+++ b/Sources/Search/APIs/Parsed.php
@@ -113,23 +113,6 @@ class Parsed extends SearchApi implements SearchApiInterface
 	 ****************/
 
 	/**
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
@Oldiesmann, since you were recently testing the search code, you probably want to pull in these latest changes to continue your tests.

PR #8205, opened 2024-05-09
Fix problems determining size of parsed index on Postgres

Oldiesmann:
Fixes two minor issues while trying to calculate the size of a parsed search index in PostgreSQL:
1. PostgreSQL doesn't like trying to calculate the size of non-existent things. Check whether the tables actually exist first both because it's easy and it allows us to properly handle actual database errors (at least in the future - as it is the code will still croak if $request is false)
2. The column in information_schema.TABLES containing the table name is "table_catalog". The schema is entirely different (see the Postgres manual on schemas for more).


Oldiesmann:
Accidentally branched this off the wrong branch. Will resubmit without the other changes in a bit.

PR #8207, opened 2024-05-10
Use PHP 8 string functions

Oldiesmann:
This changes the code to use `str_starts_with`, `str_contains` and `str_ends_with` instead of `strpos` and `substr` for finding strings within other strings. It looks nicer and is less confusing.


live627:
Riding the (PHP) storm?

Oldiesmann:
> Riding the (PHP) storm?

How did you guess? :stuck_out_tongue:

Sesquipedalian:
If you like, sure. I personally don't mind the old fashioned way for these, but I have no objection.

jdarwood007:
I think its a micro improvement:
```
<?php

$iterations = 100000;

$start = microtime(true);
for($i = 0; $i < $iterations; $i++) {
	$v = substr('Hello World', 0, 5) === 'Hello';
}
$end = microtime(true);

echo 'First:' . $end - $start . "\n";


$start = microtime(true);
for($i = 0; $i < $iterations; $i++) {
	$v = str_starts_with('Hello World', 'Hello');
}
$end = microtime(true);

echo 'Second:' . $end - $start . "\n";
```

> First:0.0059850215911865
> Second:0.0034339427947998

Somebody can check the work, but it shows that it does perform slightly better.

live627:
Makes sense, because the new functions don't copy strings.



> > Riding the (PHP) storm?
> 
> How did you guess? 😛

PHPStorm automates this. If it were me, I wouldn't do this otherwise because of the tedium, which I do not like.

live627:
Just discovered this goodie https://cs.symfony.com/doc/rules/alias/modernize_strpos.html

jdarwood007:
Neat.  Can be added, but requires us to enable risky, but the risk is not a concern:

```
diff --git a/.php-cs-fixer.dist.php b/.php-cs-fixer.dist.php
index 6fd874e2b..a3b3944a4 100644
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -34,6 +34,7 @@ return (new PhpCsFixer\Config())
     ->registerCustomFixers([
         new \SMF\Fixer\Whitespace\closing_tag_fixer(),
     ])
+    ->setRiskyAllowed(true)
        ->setRules([
                '@PSR12' => true,

@@ -156,6 +157,7 @@ return (new PhpCsFixer\Config())
                'explicit_string_variable' => true,
                'simple_to_complex_string_variable' => true,
                'single_quote' => true,
+               'modernize_strpos' => true,

                // Whitespace.
                'array_indentation' => true,
```

Sesquipedalian:
Let's go ahead and enable that option in PHP-CS-Fixer. Then you can just run PHP-CS-Fixer once to get all the remaining instances in one go, @Oldiesmann.

Oldiesmann:
I've already finished everything at this point

Sesquipedalian:
Ah. Very good, then!

other/upgrade.php
```diff
@@ -5757,4 +5757,4 @@ function upgradeGetColumnInfo($targetTable, $column): array
 	return [];
 }
 
-?>
\ No newline at end of file
+https://youtu.be/u9Dg-g7t2l4?>
```
Sesquipedalian: # 🤨

Sesquipedalian (changes requested):

other/upgrade.php
Oldiesmann: That's what happens when I accidentally middle click on my laptop instead of left click. Not sure how I didn't notice but I'll get rid of it

Sesquipedalian (approved):

PR #8218, opened 2024-05-11
Apply native_function_invocation in PHPCS

live627:
Some built-in functions can be turned into specific opcodes at compile time when the compiler is sure that they will be used (i.e. no namespace fallback is needed). This can have a major performance impact when they appear in hot paths.


tyrsson:
use function is_array;

tyrsson:
I setup and ran a benchmark on this. Now granted its not using SMF code because I set it up in a repo I had at hand and really that's irrelevant. The sample size is small but even with the low number of revs and its a benefit can be seen.

Results can be found here:
https://gist.github.com/Tyrsson/138566046b96576707afd719cd09ed81

live627:
- existing calls to built in classes are prefixed with the backslah, so this commit follows that pattern
- I didn't find an option to change this fixer's behavior, and there is no way that I'll do any of this by hand

jdarwood007:
I want to implement this.  @Sesquipedalian do you agree with adding this?

It's a compile-time optimization in PHP that means when you call a function inside a namespace, it first searches that namespace for that function and when that fails, it falls back to the global namespace.  So a bit of time is wasted searching the namespace.  The compiler can optimize the lookups this way and speed up execution.  This is a micro improvement and won't improve any page load times noticeably on the human level.

Another workaround people do is to add "use function name" to import functions from the global namespace into the local namespace.

tyrsson:
Use function is the standard. Not a work around.

jdarwood007:
@tyrsson So its more common to do `use function array_exists` rather than `\array_exists`?  I would think the latter.

tyrsson:
Its specified in PER coding standard. PER replaced the PSR. (although it was part of the PSR standard as well)

Just like there should be 

use const (statements)

jdarwood007:
Ok, well then, looks like we use [GlobalNamespaceImportFixer](https://cs.symfony.com/doc/rules/import/global_namespace_import.html)

live627:
This commit follows the example set by @Sesquipedalian where internal namespaced calls are not imported, but are instead prefixed with backslashes.

Sesquipedalian:
> Its specified in PER coding standard. PER replaced the PSR. (although it was part of the PSR standard as well)
> 
> 
> 
> Just like there should be 
> 
> 
> 
> use const (statements)

Maybe I'm missing something, but I see no such specification in https://www.php-fig.org/per/coding-style/

That said, we don't aim for complete PER compliance. For example, they can pry my tab indentation from my cold, dead hands.

Sesquipedalian:
Regarding this PR, I like the idea. Unfortunately, I entirely overlooked it before and now it has a gazillion conflicts that would need to be resolved. Rather than doing all that, I suggest redoing it based on the current HEAD of release-3.0.

tyrsson:
Ah, I suppose you are correct. I suppose I "read it in" since the vast majority of the code I deal with on a regular basis uses that usage. I find it better to be explicit. I just find it odd that folks see a difference between use Some\Class; and use some_func;. I prefer the code to be consistent throughout.

> For example, they can pry my tab indentation from my cold, dead hands.

SMF is the only project I contribute to that uses tabs :(

Sesquipedalian:
Cold. Dead. Hands.

composer.json
```diff
@@ -13,8 +13,8 @@
         "friendsofphp/php-cs-fixer": "^3.40"
     },
     "scripts": {
-        "lint": "php-cs-fixer --quiet check --config .php-cs-fixer.dist.php --path-mode=intersection $(git diff --name-only \"*.php\") || php-cs-fixer check --diff --config .php-cs-fixer.dist.php --path-mode=intersection $(git diff --name-only \"*.php\")",
-        "lint-fix": "php-cs-fixer fix -v --config .php-cs-fixer.dist.php --path-mode=intersection $(git diff --name-only \"*.php\")"
```
Sesquipedalian, line 17: Why was `--path-mode=intersection $(git diff --name-only \"*.php\")` removed from the `lint-fix` command?

live627: To get it to work for me, but then I forgot to put it back.

Sesquipedalian: Can you please add it back?

PR #8229, opened 2024-05-18
spoofdetector_name added to wrong table

Oldiesmann:
The upgrade script for MySQL adds spoofdetector_name and related indexes to the messages table instead of the members table


Sesquipedalian (approved):

Sesquipedalian:
Derp

PR #8235, opened 2024-05-22
[3.0] Fix exdates not having a default value

frandominguezl:
I was getting an error upon installing that this field didn't have a default value.


DiegoAndresCortes:
Spotted this with @Sesquipedalian 
But ` BLOB and TEXT columns cannot have DEFAULT values. ` is still in the documentation.

Something must've changed in 8.0 because SMF 2.1 was already using STRICT_TRANS_TABLES

frandominguezl:
Indeed, the MySQL 8.0.13 [release notes](https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-13.html#:~:text=.%20(WL%20%2311652)-,Data%20Type%20Notes,-MySQL%20now%20supports) include this feature. I added the default value to Postgres as well, but I don't know if that's needed.

Sesquipedalian:
The real question is, *why* is the installer complaining about this for the `exdates` column but not for the `rdates` column that is identical in every way except for the name itself? Even if recent versions of MySQL have added support for default values for text columns, default values are not normally required for them. So we must be doing something to trigger this error message. We need to figure out what are we doing and how to fix it properly.

frandominguezl:
Understood, I will try to dig into this this weekend

Sesquipedalian:
Cool. Thanks!

Oldiesmann:
Sounds like we might be trying to insert data into that particular table without setting a value for that column. That's usually what triggers the "column doesn't have a default value" error.

frandominguezl:
Oldiesmann is actually right:
https://github.com/SimpleMachines/SMF/blob/release-3.0/other/install_3-0_MySQL.sql#L1576

My preference for this kind of things is to actually set the default value upon table creation, but as I understand it, it won't be compatible for MySQL versions prior to 8.0.13 so, might just tweak that query to include the default value for `exdates`.

Am I right? :)

Oldiesmann:
Either option will work actually. The current minimum MySQL version for 3.0 is 8.0.35, and it would appear that Postgres also supports default values for text columns. Since it's just one query though it might be easier to just fix that query.

sbulen:
The real problem here is we are now inconsistent.

IMO, we should be adding defaults on *ALL* the text values.  This requires updates throughout the installer & upgrader.

albertlast:
We didn't done this,
because of the mysql version we supported in 2.1.
But yea 3.0 got new rules.

Oldiesmann:
To get SMF 3 to install for now, I modified the INSERT queries in the install file to insert '' (empty string) into that field, whih fixed the issue of course. That might just be the simplest solution.

Oldiesmann:
I don't agree with the solution of giving exdates a default value unless we do that with all the text fields. It doesn't make sense to do that for a single field - consistency is key.

sbulen:
> I don't agree with the solution of giving exdates a default value unless we do that with all the text fields. It doesn't make sense to do that for a single field - consistency is key.

Agreed.

PR #8240, opened 2024-05-31
[3.0] Manage third party libs via composer

MissAllSunday:
This is a simple POC for allowing SMF to handle their internal third party libraries  via composer.

Some key features  to take into consideration for this POC

- Composer is solely used by SMF, that is, no access to mod authors. Mod authors can still ship their own vendor folder if [they provide their respective configuration for it](https://github.com/MissAllSunday/Breeze/blob/30f1d88cb86149d45570c8e50de8de5b8f2afc1c/composer.json#L17)
- Modifications to Simplemachines/build-tools will have to be made to further remove hardcoded instances of third party libs, happy to do those changes myself if required.
- SMF release process will need to change to accommodate composer commands.


.scrutinizer.yml
```diff
@@ -1,13 +1,8 @@
 filter:
     excluded_paths:
         - '*.min.js'
-        - 'Sources/minify/'
-        - 'Sources/random_compat/'
-        - 'Sources/ReCaptcha/'
+        - 'vendor/' # Just to be sure
     dependency_paths:
```
live627: should we remove this entry if empty?

```suggestion
```

Sources/Config.php
```diff
@@ -235,6 +235,8 @@ class Config
 	 * Path to the Sources directory.
 	 */
 	public static $sourcedir;
+
```
live627, line 247:
```suggestion

	/**
	 * Path to where our dependencies are located.
	 */
```

Sources/Actions/Admin/ACP.php
```diff
@@ -1523,8 +1523,7 @@ public static function getFileVersions(array &$versionOptions): array
 		);
 
 		$ignore_sources = [
-			Config::$sourcedir . '/minify/*',
-			Config::$sourcedir . '/ReCaptcha/*',
+			Config::$vendordir . '/*',
```
live627: If I'm reading the code right, this line can go because ``$vendordir` is outside of `$sourcedir`, no?

Sources/Tasks/ExportProfileData.php
```diff
@@ -1678,21 +1678,21 @@ protected function loadCssJs(): void
...
+				include_once implode(DIRECTORY_SEPARATOR, [Config::$vendordir, 'minify', 'src', 'Minify.php']);
 
-				include_once implode(DIRECTORY_SEPARATOR, [Config::$sourcedir, 'minify', 'path-converter', 'src', 'Converter.php']);
+				include_once implode(DIRECTORY_SEPARATOR, [Config::$vendordir, 'minify', 'path-converter', 'src', 'Converter.php']);
 
-				include_once implode(DIRECTORY_SEPARATOR, [Config::$sourcedir, 'minify', 'src', 'CSS.php']);
+				include_once implode(DIRECTORY_SEPARATOR, [Config::$vendordir, 'minify', 'src', 'CSS.php']);
 
-				include_once implode(DIRECTORY_SEPARATOR, [Config::$sourcedir, 'minify', 'src', 'JS.php']);
+				include_once implode(DIRECTORY_SEPARATOR, [Config::$vendordir, 'minify', 'src', 'JS.php']);
```
live627: would it not be  better to just include the Composer autoload file here?

other/update_unicode_data.php
```diff
@@ -54,7 +54,7 @@
 }
 
 // 4. Get some more stuff we need.
-require_once $sourcedir . '/Autoloader.php';
+require_once '/vendor/autoload.php';
```
live627: why the leading slash? Wouldn't that try to look in the top directory?

index.php
```diff
@@ -131,8 +134,32 @@
...
+	}
+
+	if (class_exists(IntegrationHook::class, false)) {
+		IntegrationHook::call('integrate_autoload', [&$third_party_mappers]);
+	}
+}
+
+foreach ($third_party_mappers as $prefix => $dirname) {
+	$loader->addPsr4($prefix, $dirname);
+}
```
Sesquipedalian: I'd prefer to see this code moved back into an autoloader ~class~ <ins>file</ins> rather than having it here in index.php. 

Part of the reason for moving the existing autoloading code into a separate class was to avoid polluting the global namespace with variables in index.php. (Note: when backward compatibility support is enabled we end up putting references to class variables from SMF's namespace into the global namespace anyway, but the idea is to eliminate that in a future version.)

Sesquipedalian (changes requested):

Sources/Actions/Admin/ACP.php
MissAllSunday: Yeah, you're right, lines above specify the target which is Sources

live627 (approved):
Tested working. Normal forum operations still work as expected.

live627:
@MissAllSunday this is still marked as draft. You plan on more changes here?

MissAllSunday:
> @MissAllSunday this is still marked as draft. You plan on more changes here?

Nope, will mark it as ready and resolve the conflicts

live627:
Upon further testing, `integrate_autoload` doesn't work because it is called before settings are loaded. My fix https://github.com/live627/SMF2.1/commit/cfdfe79546def2a2de6e92b8d6b2d9924493c9a5 should  I submit this as another PR?

MissAllSunday:
Not really sure, it has to do with this PR so it makes sense to include it here but it could also be a separate PR specially if this one doesn't get merged, up to you. Best of options would be include it here and also send it as a separate PR I suppose.

MissAllSunday:
I did a rebase from release-3.0 will take a look at this again

Oldiesmann:
> I've been waiting for this PR for 5 years. Please merge it soon.

3.0 is still in alpha so it will still be a while before it's ready for use on a production forum

MissAllSunday:
> > I've been waiting for this PR for 5 years. Please merge it soon.
> 
> 3.0 is still in alpha so it will still be a while before it's ready for use on a production forum

Probably just a spammer account, I wouldn't pay too much attention to it

22522404:
> > > I've been waiting for this PR for 5 years. Please merge it soon.
> > 
> > 
> > 3.0 is still in alpha so it will still be a while before it's ready for use on a production forum
> 
> Probably just a spammer account, I wouldn't pay too much attention to it

22522404:
> > > I've been waiting for this PR for 5 years. Please merge it soon.
> > 
> > 
> > 3.0 is still in alpha so it will still be a while before it's ready for use on a production forum
> 
> Probably just a spammer account, I wouldn't pay too much attention to it

You are a fighter among garbage, you bastard are so slow, you have done such a simple thing and you haven't finished it yet, you can go to hell. I decided to develop it myself, you garbage bastard. You have no work ethic. You are a rude garbage low-level developer. You don't deserve to serve on this project.

sbulen:
Abuse reported.

Wildly inappropriate.  First time I've seen anything like that on GH.

22522404:
You spent more than a year doing a job that can be done in 7 hours. You are garbage. You are the garbage among the garbage. If you can't do this, shut it down and let someone who is capable do it. You are garbage.

Sesquipedalian (changes requested):
The changes to ./Sources/Autoloader.php will pollute the global namespace with several variables. Could you please wrap all of that code in a closure in order to keep it contained? The code in ./index.php for loading ./Settings.php would be a good model for this.

Moreover, with these changes, how does autoloading of our own classes work? Perhaps I am missing something, but it looks to me that by removing the call to spl_autoload_register() from ./Sources/Autoloader.php, this PR removes our ability to autoload our own classes.

jdarwood007:
@MissAllSunday, can you also ensure that the readme is updated with relevant information for those looking to start using the development version?  Nothing fancy, I would add a bullet to the how to contribute section.

MissAllSunday:
@Sesquipedalian @jdarwood007 

How about this approach? essentially removing the Autoloader file and the need to call Integration classes and instead, add the autoloader logic inside the already established closure which already checks and loads all what we need for using composer's autoloader.  Since autoloading happens quite early and its a special case, I don't think we need to fully load Integration class to handle it, I used the same approach we already used when loading integrate_pre_include hooks on Config.

Regarding autoloading own classes, the SMF namespace was moved to composer's configuration, meaning all classes that has that namespace will still be autoloaded just like it was before this change.  If there is a class that doesn't use the SMF namespace or lives outside its scope then it will have to be changed or its unique namespace to be added to composer's autoload setting

MissAllSunday:
rebased against latest release-3.0

jdarwood007:
#8801 adds another library.  So, figuring this out is becoming more important.  Hopefully, we can get it tagged to a milestone.

live627:
Figuring this out will be needed if we want psr interoperability.

MissAllSunday:
Figuring this out as in there is still something to be resolved on this PR?

live627:
As in, getting this assigned to a milestone; sorry for not being clear there.

MissAllSunday:
This branch and release-3.0 has diverged quite significantly so a rebase was causing a lot of problems, better start fresh so a new PR was submitted at: https://github.com/SimpleMachines/SMF/pull/9015

PR #8242, opened 2024-06-07
Add missing documentation

Oldiesmann:
This PR adds missing descriptions/missing docblocks.

### DIFF
--- a/Sources/Actions/Admin/Smileys.php
+++ b/Sources/Actions/Admin/Smileys.php
@@ -2561,7 +2561,8 @@ class Smileys implements ActionInterface
 	}
 
 	/**
-	 *
+	 * @param string $dir The directory to create
+	 * @param string $name The name of the set
 	 */
 	protected function createDir(string $dir, string $name): void
 	{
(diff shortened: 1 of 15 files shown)


Sources/Actions/Profile/GroupMembership.php
```diff
@@ -26,7 +26,7 @@
 use SMF\User;
 use SMF\Utils;
 
-/**
+/**buildquic
```
Sesquipedalian: ???

Sources/Actions/Admin/Smileys.php
```diff
@@ -2723,7 +2729,9 @@ protected static function saveSets(): void
 	}
 
 	/**
-	 *
+	 * Trims unnecessary whitespace from the given string
```
Sesquipedalian:
```suggestion
	 * Sanitizes the string and trims unnecessary whitespace.
```

Sesquipedalian (changes requested):

Sources/Actions/Profile/GroupMembership.php
Oldiesmann: Not sure how that got there

Sesquipedalian (approved):

PR #8258, opened 2024-06-25
Improves detection & removal of mods' lang files during uninstall

### LINKED ISSUE #8094: [3.0]: Language files are not removed on mods deleting
### Basic Information

sshot-7

### Steps to reproduce

1. Install any mod with hooks and language files for SMF 2.1
2. Try to uninstall it

### Expected result

`<remove-dir name="$languagedir/modname" />`

I think that we need to check both paths on deleting: `$boarddir/Languages`, and `$themedir/languages` (if compatibility mode is enabled)

### Actual result

_No response_

### Version/Git revision

3.0 Alpha 2

### Database Engine

MySQL

### Database Version

10.11.6-MariaDB

### PHP Version

8.0.30

### Logs

_No response_

### Additional Information

_No response_

Sesquipedalian:
Fixes #8094

@dragomano, please test to confirm.

### DIFF
--- a/Sources/PackageManager/SubsPackage.php
+++ b/Sources/PackageManager/SubsPackage.php
@@ -1393,6 +1393,8 @@ class SubsPackage
 							'filename' => $this_action['filename'],
 						];
 					}
+
+					self::fixLangFilePathForRemoval($return, $this_action);
 				} elseif ($actionType == 'remove-file') {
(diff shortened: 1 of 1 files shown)


dragomano:
It works, but this repeated code should probably be moved in a separate function.

Sesquipedalian:
True enough.

PR #8259, opened 2024-06-26
Upgr fixes

### LINKED ISSUE #7799: Upgrader: Old protected_alter logic is producing deprecation warnings
#### Description

The upgrader is producing deprecation warnings when upgrading from SMF 2.0.x or prior.  This logic is coming from the protected_alter function in upgrade.php:
https://github.com/SimpleMachines/SMF/blob/4d7fbd840b244b1940b41e1023e366d2b7ad4dbc/other/upgrade.php#L2597

The errors look like:
> *  +++  Renaming table columns.Error: strpos(): Passing null to parameter # 1 ($haystack) of type string is deprecated File: D:\wamp64\www\yabbse\upgrade.php Line: 2597.Error: strpos(): Passing null to parameter # 1 ($haystack) of type string is deprecated File: D:\wamp64\www\yabbse\upgrade.php Line: 2597

I haven't looked too closely yet...  It appears to think $row['Info'] is null, which doesn't make sense to me...

But the bigger question is:  **_Do we even need the protected alter function???_**

I am not quite sure what that old function is there to do...

### Steps to reproduce
1.  Perform an upgrade using PHP 8.1 or PHP 8.2
2. Run the upgrade from the command line - so errors are preserved/displayed
3. Run an upgrade from 2.0 or prior

- Database Type: MySQL
- Database Version: 8.0
- PHP Version: 8.2 (confirmed in 8.1 also)
- WAMP - haven't tried linux yet...

sbulen:
Fixes #7799 

Also adds $smcFunc['truncate'], used when building log_search_subjects via text2words() in old 1.0.x migrations.  This was causing fatal errors when running the upgrader against 1.0.x & yabbse1.5.5.

Needed to get this line:
https://github.com/SimpleMachines/SMF/blob/454e67f2dc49f05bfa1e978d6d5079ba889116c1/Sources/Subs.php#L5332

To behave like the prior version for the upgrade:
https://github.com/SimpleMachines/SMF/blob/19a9de63d8b895c6f931137bba09c3d5a79caf52/Sources/Subs.php#L5325

Note that $smcFunc['truncate'] is actually pretty involved & factors in entity encoding...  
https://github.com/SimpleMachines/SMF/blob/454e67f2dc49f05bfa1e978d6d5079ba889116c1/Sources/Load.php#L222

But to port over all the portions of Load.php into upgrade.php to make that work felt like overkill.  Opted to just match the prior logic, which has been used for years.

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -2594,7 +2594,7 @@ function protected_alter($change, $substep, $is_test = false)
 			SHOW FULL PROCESSLIST');
 		while ($row = $smcFunc['db_fetch_assoc']($request))
 		{
-			if (strpos($row['Info'], 'ALTER TABLE ' . $db_prefix . $change['table']) !== false && strpos($row['Info'], $change['text']) !== false)
+			if ($row['Info'] !== null && (strpos($row['Info'], 'ALTER TABLE ' . $db_prefix . $change['table']) !== false && strpos($row['Info'], $change['text']) !== false))
 				$found = true;
 		}
 
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -715,6 +715,12 @@ function loadEssentialData()
 		return random_bytes($bytes);
 	};
 
+	// This is used in text2words() for old 1.0.x conversions; restoring old logic
+	$smcFunc['truncate'] = function($word, $max_chars)
+	{
+		return substr($word, 0, $max_chars);
+	};
+
 	// We need this for authentication and some upgrade code
 	require_once($sourcedir . '/Subs-Auth.php');
 	require_once($sourcedir . '/Class-Package.php');
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -718,7 +718,17 @@ function loadEssentialData()
 	// This is used in text2words() for old 1.0.x conversions; restoring old logic
 	$smcFunc['truncate'] = function($word, $max_chars)
 	{
-		return substr($word, 0, $max_chars);
+		$new_string = '';
+
+		foreach (preg_split('/((?>&.*?;|\X))/u', $string, -1, PREG_SPLIT_DELIM_CAPTURE | PREG_SPLIT_NO_EMPTY) as $char)
+		{
+			if (strlen($new_string . $char) > $max_chars)
+				break;
+
+			$new_string .= $char;
+		}
+
+		return $new_string;
 	};
 
 	// We need this for authentication and some upgrade code


other/upgrade.php
```diff
@@ -715,6 +715,12 @@ function loadEssentialData()
 		return random_bytes($bytes);
 	};
 
+	// This is used in text2words() for old 1.0.x conversions; restoring old logic
+	$smcFunc['truncate'] = function($word, $max_chars)
+	{
+		return substr($word, 0, $max_chars);
+	};
```
Sesquipedalian:
```suggestion
	$smcFunc['truncate'] = function($word, $max_chars)
	{
		$new_string = '';

		foreach (preg_split('/((?>&.*?;|\X))/u', $string, -1, PREG_SPLIT_DELIM_CAPTURE | PREG_SPLIT_NO_EMPTY) as $char)
		{
			if (strlen($new_string . $char) > $max_chars)
				break;

			$new_string .= $char;
		}

		return $new_string;
	};
```


This will avoid chopping in the middle of entities or multi-byte characters. It's still not quite as cushy as the one in Load.php, but it will at least prevent malformed output.

Sesquipedalian (changes requested):

Sesquipedalian (approved):

sbulen:
@Sesquipedalian -

The latest change introduced an error:
> +++  Indexing topic subjects...Error: Undefined variable $string File: \yadayada\van1023\upgrade.php Line: 723Error: preg_split(): Passing null to parameter # 2 ($subject) of type string is deprecated File: \yadayada\van1023\upgrade.php Line: 723

Wrong var name used:
image

Sesquipedalian:
D'oh!

sbulen:
Please note also that I'd be hesitant to assume the "u" modifier is OK - a lot of these old DBs are probably not utf8.  

Many aren't even ISO-8859-1...  Some are ISO-8859-2+, for other language support back in the day...

I helped one VERY LARGE forum (lcdtvthailand) migrate forward from TIS-620...

I'm not sure I know what the "u" modifier will do to that.

Part of this is why I left the truncate function alone, as noted in my comment.

OTOH...  I do think that it is quite possible that the EXACT logic used in Load.php might work, because all the vars are defined at that point in the upgrader (maybe moved a few lines lower after $utf8 is declared)...
https://github.com/SimpleMachines/SMF/blob/6996e045d959f4af2d0b185c44344b5be724ee4c/Sources/Load.php#L222

sbulen:
I know I am often too conservative on these things, but...  I'd rather run with what's been working for the last umpteen years...

Sesquipedalian:
It will be fine. All those old 8-bit character sets will be divided up by this regex correctly. If we were trying to *analyze* the characters, we'd get into trouble, because the bytes beyond 7F would mean different things in different character sets. But we're just *splitting* them, so as long as each character is just one byte (which they are), it doesn't matter what they map to in this or that character set. Indeed, the only reason I used the `u` flag was to get access to the `\X` shorthand, which is like `.` except that it will match newlines, too. Beyond that, the `u` flag will make no discernible difference in this context.

> I know I am often too conservative on these things, but... I'd rather run with what's been working for the last umpteen years...

That's the thing. It hasn't been. It cuts in the middle of entities. That's how the problems with broken entities in the log_search_subjects table happened.

Sesquipedalian:
... Actually, just in case, I will go ahead and change it so that it doesn't use the `u` flag after all.

PR #8262, opened 2024-06-28
Remaps \SMF namespace to forum's root directory

Sesquipedalian:
Any objections?

### DIFF
--- a/Languages/en_US/Errors.php
+++ b/Languages/en_US/Errors.php
@@ -518,7 +518,7 @@ $txt['undefined_xml_attribute'] = 'Undefined XML attribute: {0}';
 $txt['undefined_xml_element'] = 'Undefined XML element: {0}';
 
 // User::load() error
-$txt['invalid_member_data_set'] = 'SMF\User::load(): Invalid member set: "{0}"';
+$txt['invalid_member_data_set'] = 'SMF\Sources\User::load(): Invalid member set: "{0}"';
 
 // logActions() errors
 $txt['logActions_not_array'] = 'logActions(): data is not an array with action "{0}"';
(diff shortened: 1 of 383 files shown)


jdarwood007:
I'm not certain why. It doesn't need to match our folder structure.  @Tyrsson any thoughts?

Sesquipedalian:
Part of the reason is explained in https://www.simplemachines.org/community/index.php?msg=4176355, but more generally, the thinking was that this would allow us to bring ./index.php, ./cron.php, ./Themes, etc., into the SMF namespace.

tyrsson:
> Part of the reason is explained in https://www.simplemachines.org/community/index.php?msg=4176355, but more generally, the thinking was that this would allow us to bring ./index.php, ./cron.php, ./Themes, etc., into the SMF namespace.

Why would one want to namespace files that do not define a class or function?

I'm not sure what this is meant to accomplish or what issue its intended to mitigate..

jdarwood007:
Ahh, I hadn't stepped onto the site, been busy.

I can see why the idea comes up.  We could still move the Maintenance folder outside of the sources.  We would just have a call for our autoloader to start somewhere different when the SMF\Maintenance namespace is reached.  The root files should if they have to have a namespace, be just SMF.  They shouldn't since its the entry point into the program.

Themes make sense in high lever terms, but I think implementation gets a bit difficult.  Currently, SMF just checks for the files existent when loading another theme and if it doesn't, resorts to the default theme file.  For a class and/or namespace system, we would have to track that to make the correct calls.  I think procedural themes are easier for the theme system than having it class and/or namespaced.

Sesquipedalian:
Hm. Yeah, fair point.

PR #8278, opened 2024-07-01
3.0/caching issues

Sesquipedalian:
Fixes some issues with caching.

First, there were some problems with the Sqlite cache API that have been fixed in 17f9186.

Second, and more importantly, 17fdb6e changes how we convert cached data to strings. Instead of using json_encode() and json_decode(), we are now using serialize() and unserialize(). This is necessary because in SMF 3.0 we need to cache objects. For example, when caching recent events, we pass a collection of SMF\Calendar\EventOccurrence objects to be cached. Encoding that to JSON loses data and the objects cannot be reconstructed when retrieved, resulting in nasty errors.

### DIFF
--- a/Sources/Cache/APIs/Sqlite.php
+++ b/Sources/Cache/APIs/Sqlite.php
@@ -109,7 +109,7 @@ class Sqlite extends CacheApi implements CacheApiInterface
 		if ($value === null) {
 			$query = 'DELETE FROM cache WHERE key = \'' . $this->cacheDB->escapeString($key) . '\';';
 		} else {
-			$query = 'REPLACE INTO cache VALUES (\'' . $this->cacheDB->escapeString($key) . '\', \'' . $this->cacheDB->escapeString($value) . '\', ' . $ttl . ');';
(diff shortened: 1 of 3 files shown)


tyrsson:
> Fixes some issues with caching.
> 
> First, there were some problems with the Sqlite cache API that have been fixed in [17f9186](https://github.com/SimpleMachines/SMF/commit/17f918693d5aa11493afb320b13478af3d743295).
> 
> Second, and more importantly, [17fdb6e](https://github.com/SimpleMachines/SMF/commit/17fdb6e07a6595d2ed494f47d19c6ac65404ba62) changes how we convert cached data to strings. Instead of using json_encode() and json_decode(), we are now using serialize() and unserialize(). This is necessary because in SMF 3.0 we need to cache objects. For example, when caching recent events, we pass a collection of SMF\Calendar\EventOccurrence objects to be cached. Encoding that to JSON loses data and the objects cannot be reconstructed when retrieved, resulting in nasty errors.

Sounds familiar.... Sorry, I have not been able to finish the psr compliance work :)

I would follow the usage here:
**https://wiki.php.net/rfc/custom_object_serialization#:~:text=The%20__serialize()%20and,if%20it%20represented%20object%20properties.**

Sesquipedalian:
> Sounds familiar....

Yup

PR #8287, opened 2024-07-06
Drop single dimensional array support for data array in insert function

albertlast:
Since it bugged since 2.1 -> no one mention the issue, we could drop the feature and make the interface more clear.

### DIFF
--- a/Sources/Db/APIs/MySQL.php
+++ b/Sources/Db/APIs/MySQL.php
@@ -361,11 +361,6 @@ class MySQL extends DatabaseApi implements DatabaseApiInterface
 			}
 		}
 
-		// Inserting data as a single row can be done as a single array.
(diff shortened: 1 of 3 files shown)


live627:
nope,  not ready for thsi major breaking    change because this ends up  breaking half of the entire application, let alone third party mods.

albertlast:
maybe you didn't read the mayor point, but it doesn't work atm.
Or miss i something?

live627:
How is it bugged? Can I have an example?

Sesquipedalian:
Checking only one random element in $data makes no sense. I think @albertlast is correct that the current code could cause bugs. 

However, the solution is not to just remove this code. Instead, we should check all the elements in $data to make sure that all the elements in $data are as expected.

PR #8303, opened 2024-08-11
Initial requirements for 3.0

jdarwood007:
@Sesquipedalian I just pulled initial data, we will amend this as 3.0 progresses.  This will be used to assist our doc team and our site information for information about minimum requirements for SMF.

### DIFF
--- a/other/requirements.md
+++ b/other/requirements.md
@@ -1,0 +1,57 @@
+# SMF Miniumn Requirements
+
+## PHP
+### PHP Version Support
+| MIN SMF VERSION | MAX SMF VERSION | MIN PHP VERSION | MAX PHP VERSION |
+| ------ | ------ | ------ | ------ |
+| 3.0    | LATEST | 8.0.0  | LATEST  |
+
+### Undocumented Version Support
+- Versions below miniumn listed above are not supported
+- Versions above the maxiumn listed above:
+	- Will be supported in a future date unless this version has reached End of Life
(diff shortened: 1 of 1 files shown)


other/requirements.md
```diff
@@ -0,0 +1,57 @@
+# SMF Miniumn Requirements
+
+## PHP
+### PHP Version Support
+| MIN SMF VERSION | MAX SMF VERSION | MIN PHP VERSION | MAX PHP VERSION |
+| ------ | ------ | ------ | ------ |
+| 3.0    | LATEST | 8.0.0  | LATEST  |
```
live627, line 7: Should we also include 2.1 here? Thoughts?
```suggestion
| 2.1    | LATEST | 7.0.0  | LATEST  |
| 3.0    | LATEST | 8.0.0  | LATEST  |
```

jdarwood007: We can, but the 2.1 branch can also hold it.

DiegoAndresCortes: Are the branches mentioned somewhere? This may cause confusion since different projects tend to do one repo per version thus making it non-trivial unless you previously contributed to 2.1 or you check the branches.

Sesquipedalian: That's not really correct. Some projects might do that, but projects that correctly understand how Git works don't typically do that.

Sesquipedalian (approved):

Issue #8293, opened 2024-07-17
[3.0]: Config quirks


sbulen:
### Basic Information

1.  In Windows, Settings.php is being reloaded constantly, mainly due to it's not being found in a get_included_files() check.  This is due to the different DIRECTORY_SEPARATORs between Windows & linux.
2. eval returns null, which can result in null being assigned to a setting under some circumstances.  E.g., attempting an upgrade, received the following error:

![image](https://github.com/user-attachments/assets/9c8d0f08-e5cb-49be-a94f-29e6171e3b9c)

### Steps to reproduce

1.  Install 2.1.4
3.  Upgrade to 3.0

### Expected result

A 3.0 forum

### Actual result

WSOD error

### Version/Git revision

4.0 Alpha 2 - current GH

### Database Engine

All

### Database Version

8.4.0

### PHP Version

8.3.8

### Logs

```shell
Fatal error: Uncaught TypeError: Cannot assign null to property SMF\Config::$languagesdir of type string in D:\wamp64\www\84van30\Sources\Config.php:915 Stack trace: #0 D:\wamp64\www\84van30\Sources\Config.php(886): SMF\Config::set(Array) #1 D:\wamp64\www\84van30\upgrade.php(188): SMF\Config::load() #2 {main} thrown in D:\wamp64\www\84van30\Sources\Config.php on line 915
```

sbulen:
I'm using my WAMP...

I think at least part of the problem is directory separators...  It keeps thinking things aren't loaded...
![image](https://github.com/user-attachments/assets/00b2b91d-d10d-4ce0-bb9e-dba9f75bcb16)

sbulen:
I believe the rest of the issue is that eval() is used in Config.php against the definition default value strings.
But...  **_eval() RETURNS NULL_** unless there is an explicit return in the eval'd code...  
https://github.com/SimpleMachines/SMF/blob/7788cb5c26ab7bc7ea77d3fac6ec2b29bef3f854/Sources/Config.php#L915

Using double quotes, ", will likely work better.

sbulen:
Lots of options on the DIRECTORY_SEPARATOR question...

Option 1: Bite the bullet & use DIRECTORY SEPARATOR consistently throughout; never hardcode '/' again...  Ensure folder settings are OS-appropriate.  

Proper solution.  Least fun.  Like having a paladin on the team.  

[...]

PR #8316, opened 2024-10-06
[3.0] Fix eval return; set dir consistently throughnout

sbulen:
Partial for #8293 - this addresses the null assignment.

This also addresses some inconsistencies, sometimes __DIR__ was used and sometimes dirname(__FILE__) was used, when they were both doing the same thing.

There was some logic in there that oddly reversed them...(???)  Tore that out.

### DIFF
--- a/Sources/Config.php
+++ b/Sources/Config.php
@@ -1697,6 +1697,14 @@ class Config
 
 					if (!empty($setting_def['raw_default']) && $setting_def['default'] !== '') {
 						$var_pattern[] = preg_replace('/\s+/', '\s+', preg_quote($setting_def['default'], '~'));
+
+						if (str_contains($setting_def['default'], 'dirname(__FILE__)')) {
+							$var_pattern[] = preg_replace('/\s+/', '\s+', preg_quote(str_replace('dirname(__FILE__)', '__DIR__', $setting_def['default']), '~'));
+						}
+
+						if (str_contains($setting_def['default'], '__DIR__')) {
+							$var_pattern[] = preg_replace('/\s+/', '\s+', preg_quote(str_replace('__DIR__', 'dirname(__FILE__)', $setting_def['default']), '~'));
+						}
 					}
 
 					$var_pattern = array_unique($var_pattern);
--- a/Sources/Config.php
+++ b/Sources/Config.php
@@ -682,7 +682,7 @@ class Config
 				 * Path to the cache directory for the file-based cache system.
 				 */
 				END,
-			'default' => 'dirname(__FILE__) . \'/cache\'',
+			'default' => '__DIR__ . \'/cache\'',
 			'raw_default' => true,
 			'type' => 'string',
 		],
@@ -745,7 +745,7 @@ class Config
 				 * The absolute path to the forum's folder. (not just '.'!)
 				 */
 				END,
-			'default' => 'dirname(__FILE__)',
+			'default' => '__DIR__',
 			'raw_default' => true,
 			'type' => 'string',
 		],
@@ -757,7 +757,7 @@ class Config
 				 * Path to the Sources directory.
 				 */
 				END,
-			'default' => 'dirname(__FILE__) . \'/Sources\'',
+			'default' => '__DIR__ . \'/Sources\'',
 			'raw_default' => true,
 			'type' => 'string',
 		],
@@ -769,7 +769,7 @@ class Config
 				 * Path to the Packages directory.
 				 */
 				END,
-			'default' => 'dirname(__FILE__) . \'/Packages\'',
+			'default' => '__DIR__ . \'/Packages\'',
 			'raw_default' => true,
 			'type' => 'string',
 		],
@@ -920,7 +920,7 @@ class Config
 						'__DIR__' => var_export(dirname(SMF_SETTINGS_FILE), true),
 					]);
 
-					self::${$var} = eval($default . ';');
+					self::${$var} = eval('return ' . $default . ';');
 				} else {
 					self::${$var} = $def['default'];
 				}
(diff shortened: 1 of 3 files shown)


Sources/Config.php
```diff
@@ -920,7 +920,7 @@ public static function set(array $settings): void
 						'__DIR__' => var_export(dirname(SMF_SETTINGS_FILE), true),
 					]);
 
-					self::${$var} = eval($default . ';');
+					self::${$var} = eval('return ' . $default . ';');
```
Sesquipedalian, line 923: This is good.

Sources/Config.php
```diff
@@ -682,7 +682,7 @@ class Config
 				 * Path to the cache directory for the file-based cache system.
 				 */
 				END,
-			'default' => 'dirname(__FILE__) . \'/cache\'',
+			'default' => '__DIR__ . \'/cache\'',
```
Sesquipedalian, line 685: This is acceptable, but only if the same change is made in ./other/Settings.php.

Sources/Config.php
```diff
@@ -745,7 +745,7 @@ class Config
 				 * The absolute path to the forum's folder. (not just '.'!)
 				 */
 				END,
-			'default' => 'dirname(__FILE__)',
+			'default' => '__DIR__',
```
Sesquipedalian, line 748: This is acceptable, but only if the same change is made in ./other/Settings.php.

Sources/Config.php
```diff
@@ -757,7 +757,7 @@ class Config
 				 * Path to the Sources directory.
 				 */
 				END,
-			'default' => 'dirname(__FILE__) . \'/Sources\'',
+			'default' => '__DIR__ . \'/Sources\'',
```
Sesquipedalian, line 760: This is acceptable, but only if the same change is made in ./other/Settings.php.

Sources/Config.php
```diff
@@ -769,7 +769,7 @@ class Config
 				 * Path to the Packages directory.
 				 */
 				END,
-			'default' => 'dirname(__FILE__) . \'/Packages\'',
+			'default' => '__DIR__ . \'/Packages\'',
```
Sesquipedalian, line 772: This is acceptable, but only if the same change is made in ./other/Settings.php.

Sources/Config.php
```diff
@@ -1697,14 +1697,6 @@ public static function updateSettingsFile(array $config_vars, ?bool $keep_quotes
 
 					if (!empty($setting_def['raw_default']) && $setting_def['default'] !== '') {
 						$var_pattern[] = preg_replace('/\s+/', '\s+', preg_quote($setting_def['default'], '~'));
-
```
Sesquipedalian: This is bad. These lines do in fact perform an important function, and must be restored.

Sesquipedalian (changes requested):

Sources/Config.php
sbulen: 👍

Sources/Config.php
sbulen: All references to dirname(__FILE__) changed to __DIR__ in Settings.php & Settings_bak.php.

Sources/Config.php
sbulen: All references to dirname(__FILE__) changed to __DIR__ in Settings.php & Settings_bak.php.

Sources/Config.php
sbulen: All references to dirname(__FILE__) changed to __DIR__ in Settings.php & Settings_bak.php.

Sources/Config.php
sbulen: All references to dirname(__FILE__) changed to __DIR__ in Settings.php & Settings_bak.php.

Sources/Config.php
sbulen: Change made.  I think I'm missing something there, because it just adds replacements to change dirname(__FILE__) to __DIR__ & vice-versa...  Which won't do anything meaningful AFAICT?

If you could briefly explain what I'm missing, I'd appreciate it!

Sesquipedalian: The part you are missing is that these statements are not changing the content of anything. Rather, they are building a series of regex patterns that will all be tested later.

For example, let's take the definition for the `$cachedir` variable:
```php
'cachedir' => [
	'text' => <<<'END'
		/**
		 * @var string
		 *
		 * Path to the cache directory for the file-based cache system.
		 */
		END,
	'default' => '__DIR__ . \'/cache\'',
	'raw_default' => true,
	'type' => 'string',
],
```

When we get to the following block of code, the outer `if` condition is satisfied, so the contents are executed in order to create some regex patterns and add them to the `$var_pattern` array:
```php
if (!empty($setting_def['raw_default']) && $setting_def['default'] !== '') {
	$var_pattern[] = preg_replace('/\s+/', '\s+', preg_quote($setting_def['default'], '~'));

	if (str_contains($setting_def['default'], 'dirname(__FILE__)')) {
		$var_pattern[] = preg_replace('/\s+/', '\s+', preg_quote(str_replace('dirname(__FILE__)', '__DIR__', $setting_def['default']), '~'));
	}

	if (str_contains($setting_def['default'], '__DIR__')) {
		$var_pattern[] = preg_replace('/\s+/', '\s+', preg_quote(str_replace('__DIR__', 'dirname(__FILE__)', $setting_def['default']), '~'));
	}
}
```

The first statement, `$var_pattern[] = preg_replace('/\s+/', '\s+', preg_quote($setting_def['default'], '~'));`, causes the following regex to be added to the array: `__DIR__\s+\.\s+'/cache'`.

The next statement is an `if` test that fails, since the definition for `$cachedir`'s default value does not contain `dirname(__FILE__)`, so its content is skipped.

The third statement is an `if` that passes, since the definition for `$cachedir`'s default value does contain `__DIR__`, so its content is executed. 

That content consists of the statement `$var_pattern[] = preg_replace('/\s+/', '\s+', preg_quote(str_replace('__DIR__', 'dirname(__FILE__)', $setting_def['default']), '~'));`, which causes the following regex to be added to the array: `dirname(__FILE__)\s+\.\s+'/cache'`.

The upshot is that `$var_pattern` now contains regular expressions that will correctly detect both the `dirname(__FILE__) . '/cache'` and the `__DIR__ . '/cache'` versions of the default value.

Sesquipedalian (approved):

PR #8318, opened 2024-10-07
[3.0]: add $possible_topic_columns to integrate_modify_post

live627:
### Motivation

I have a modification (Topic Descriptions) that adds a column to `smf_topics`. Currently, a separate query is necessary to update the topics table. This commit allows for adding the new column to the existing query.

### Description
1. **Introduction of `$possible_topic_columns`**:
   - A new associative array, `$possible_topic_columns`, is added to map topic option keys (e.g., `sticky_mode`, `lock_mode`, `poll`) to their corresponding database column names and data types.
   - This replaces hardcoded column and parameter assignments for topic modifications.

2. **Enhancements to the `integrate_modify_post` Hook**:
   - The `integrate_modify_post` hook now includes the new `$possible_topic_columns` array, enabling mods to interact with these columns more effectively.

The `$possible_topic_columns` array introduced in the patch is an associative array. Its structure is as follows:

```php
$possible_topic_columns = [
    'key_in_topicOptions' => ['type' => 'db_column_name'],
];
```

### Example from the Patch

```php
$possible_topic_columns = [
    'sticky_mode' => ['int' => 'is_sticky'], // Maps 'sticky_mode' from $topicOptions to the 'is_sticky' database column with type 'int'.
    'lock_mode'   => ['int' => 'locked'],   // Maps 'lock_mode' from $topicOptions to the 'locked' database column with type 'int'.
    'poll'        => ['int' => 'id_poll'],  // Maps 'poll' from $topicOptions to the 'id_poll' database column with type 'int'.
];
```

### Details:
1. **Keys (`'sticky_mode'`, `'lock_mode'`, `'poll'`)**:
   - These are the keys expected in the `$topicOptions` array, passed as input to the function.

2. **Values**:
   - The value is an array where:
     - The first element is the **data type** (`'int'`, `'string'`, etc.) for parameter binding in the query.
     - The second element is the **database column name** (`'is_sticky'`, `'locked'`, `'id_poll'`).

### DIFF
--- a/Sources/Msg.php
+++ b/Sources/Msg.php
@@ -1641,10 +1641,6 @@ class Msg implements \ArrayAccess
 	{
 		$searchAPI = SearchApi::load();
 
-		$topicOptions['poll'] = isset($topicOptions['poll']) ? (int) $topicOptions['poll'] : null;
-		$topicOptions['lock_mode'] = $topicOptions['lock_mode'] ?? null;
-		$topicOptions['sticky_mode'] = $topicOptions['sticky_mode'] ?? null;
-
 		// This is longer than it has to be, but makes it so we only set/change what we have to.
 		$messages_columns = [];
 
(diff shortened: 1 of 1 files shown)


live627:
request for comment @jdarwood007 @Sesquipedalian or should I merge this as-is

Sesquipedalian:
Sorry, I'm still on hiatus so I haven't had a chance to look at this. Can you add a description for the PR so that we know what it intends to accomplish?

live627:
summady added courtesy of  chat gpt

Sesquipedalian:
Thanks, but I wasn't really asking for a description of _what_ this PR changes. I was asking _why_ you want to make these changes. What's the goal? What do these changes make possible?

live627:
I have a modification (Topic Descriptions) that adds a column to `smf_topics`. Currently, a separate query is necessary to update the topics table. This commit allows for adding the new column to the existing query.

Sesquipedalian:
Ah. That makes sense. Thanks. I will review this as soon as I can.

Sesquipedalian (approved):

Issue #8307, opened 2024-08-18
[3.0] & [2.1]: php 8.3 - Blank screen viewing source in error log code popup window


sbulen:
### Basic Information

The popup window to view source in the error log doesn't work in PHP 8.3.

This is true for 2.1 & 3.0.  PHP 8.2 works fine.

No issue logged, in smf, php or apache error logs.  Nothing in the console either.

From the forum:
https://www.simplemachines.org/community/index.php?topic=589529.0

### Steps to reproduce

1. Set php version to 8.3.
2. Open the error log, click on the source/line link to view source.

### Expected result

This is what it looks like in php 8.2:
![image](https://github.com/user-attachments/assets/0c1c1e07-cf76-4340-95aa-50b6553db0ee)

### Actual result

This:
![image](https://github.com/user-attachments/assets/aa739d84-d49e-40c7-bbcf-7dc7494ff8c3)

### Version/Git revision

3.0 Alpha 2 & 2.1.4

### Database Engine

All

### Database Version

8.4

### PHP Version

8.3.8

### Logs

```shell
Nothing found in any log...
```

sbulen:
The root of the issue here is that in PHP 8.3, significant changes were made to php's to highlight_string()...
https://php.watch/versions/8.3/highlight_file-highlight_string-html-changes

highlight_string() is used in our highlightPhpCode() function.

In particular, highlight_string() no longer returns line feeds as `<br />`.

jdarwood007:
Looks like we need to version sniff to handle the output properly then.

sbulen:
There is logic there already for 8.3, but I think it needs a tweak or two.

I'm working on a PR.

PR #8319, opened 2024-10-08
[3.0] Fix blank screens in error log

sbulen:
Partial for #8307 

Prior to php 8.3, highlight_string() added some wrapper code before & after the requested code - code tags, spans, etc.  Including line feeds...  Since eols were returned as br tags _within_ the requested code, you could easily strip the wrapper line feeds by mapping \n to ''.  

With 8.3, everything comes back using newlines, both the requested code & the wrapper code, so stripping newlines broke stuff by turning the whole program into one large line...

Two tweaks needed:
 - Translate all newlines into br tags, like in php < 8.3, and as expected by downstream processing
 - Strip br tags inserted before & after the requested code, so code line numbers are still accurate when displayed
 
Tested in php 8.l, 8.2, and 8.3.

### DIFF
--- a/Sources/BBCodeParser.php
+++ b/Sources/BBCodeParser.php
@@ -2077,7 +2077,7 @@ class BBCodeParser
 
 		$oldlevel = error_reporting(0);
 
-		$buffer = str_replace(["\n", "\r"], '', @highlight_string($code, true));
+		$buffer = str_replace(["\n", "\r"], ['<br />', ''], @highlight_string($code, true));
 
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
I think this has already been taken care of in bbde2cd. If not, feel free to submit a new, non-conflicting PR.

Sesquipedalian:
Oh, actually no, the issue wasn't fixed in bbde2cd. But I've implemented a fix for it in #8438

PR #8321, opened 2024-10-09
[3.0] Cast objs to string when used as keys

### LINKED ISSUE #8306: [3.0]: New user signup WSOD - Uncaught TypeError: Cannot access offset of type SMF\IP on array
### Basic Information

At a couple points in a new user signup and approval process, the admin may get a WSOD.

### Steps to reproduce

1. Config so new users require admin approval & email activation.
2. Test user signs up.  Alert is sent to admin.
3. Admin approves.  *** May WSOD upon clicking the alert here ***
4. Admin gets to the approval screen (can sometimes get that far), and clicks on the "CHECK FOR DUPLICATES" button, *** WSOD here ***

### Expected result

Admin should be able to get to the approval screen & approve.

### Actual result

Sometimes, WSODs just trying to get to the screen.  Sometimes, WSODs trying to take actions there, e.g., "CHECK FOR DUPLICATES".

### Version/Git revision

3.0 Alpha 2

### Database Engine

All

### Database Version

8.4

### PHP Version

8.3.8

### Logs

```shell
Fatal error: Uncaught TypeError: Cannot access offset of type SMF\IP on array in D:\wamp64\www\84van30\Sources\Actions\Admin\Members.php:1527 
Stack trace: 
#0 D:\wamp64\www\84van30\Sources\Actions\Admin\Members.php(1300): SMF\Actions\Admin\Members::populateDuplicateMembers(Array) 
#1 [internal function]: SMF\Actions\Admin\Members::list_getMembers(0, 30, 'date_registered...', 'is_activated = ...', Array, true) 
#2 D:\wamp64\www\84van30\Sources\ItemList.php(496): call_user_func_array(Array, Array) 
#3 D:\wamp64\www\84van30\Sources\ItemList.php(208): SMF\ItemList->getItems() 
#4 D:\wamp64\www\84van30\Sources\Actions\Admin\Members.php(1014): SMF\ItemList->__construct(NULL) 
#5 [internal function]: SMF\Actions\Admin\Members->browse() 
#6 D:\wamp64\www\84van30\Sources\Actions\Admin\Members.php(144): call_user_func(Array) 
#7 D:\wamp64\www\84van30\Sources\ActionTrait.php(55): SMF\Actions\Admin\Members->execute() 
#8 [internal function]: SMF\Actions\Admin\Members::call() 
#9 D:\wamp64\www\84van30\Sources\Actions\Admin\ACP.php(781): call_user_func(Array) 
#10 D:\wamp64\www\84van30\Sources\ActionTrait.php(55): SMF\Actions\Admin\ACP->execute() 
#11 [internal function]: SMF\Actions\Admin\ACP::call() 
#12 D:\wamp64\www\84van30\Sources\Forum.php(264): call_user_func(Array) 
#13 D:\wamp64\www\84van30\index.php(149): SMF\Forum->execute() 
#14 {main} thrown in D:\wamp64\www\84van30\Sources\Actions\Admin\Members.php on line 1527
```

### Additional Information

_No response_

sbulen:
Fixes #8306 

Cannot use the object itself as an array key.  Using the __toString() method instead wherever used as a key.

Tests OK.

### DIFF
--- a/Sources/Actions/Admin/Members.php
+++ b/Sources/Actions/Admin/Members.php
@@ -1524,11 +1524,11 @@ class Members implements ActionInterface
 			];
 
 			if (in_array($row['member_ip'], $ips)) {
-				$duplicate_members[$row['member_ip']][] = $member_context;
+				$duplicate_members[$row['member_ip']->__toString()][] = $member_context;
 			}
 
 			if ($row['member_ip'] != $row['member_ip2'] && in_array($row['member_ip2'], $ips)) {
-				$duplicate_members[$row['member_ip2']][] = $member_context;
+			    $duplicate_members[$row['member_ip2']->__toString()][] = $member_context;
 			}
 		}
 		Db::$db->free_result($request);
@@ -1561,7 +1561,7 @@ class Members implements ActionInterface
 
 			$had_ips[$row['poster_ip']][] = $row['id_member'];
 
-			$duplicate_members[$row['poster_ip']][] = [
+			$duplicate_members[$row['poster_ip']->__toString()][] = [
 				'id' => $row['id_member'],
 				'name' => $row['member_name'],
 				'email' => $row['email_address'],
@@ -1575,12 +1575,12 @@ class Members implements ActionInterface
 		// Now we have all the duplicate members, stick them with their respective member in the list.
 		if (!empty($duplicate_members)) {
 			foreach ($members as $key => $member) {
-				if (isset($duplicate_members[$member['member_ip']])) {
-					$members[$key]['duplicate_members'] = $duplicate_members[$member['member_ip']];
+				if (isset($duplicate_members[$member['member_ip']->__toString()])) {
+					$members[$key]['duplicate_members'] = $duplicate_members[$member['member_ip']->__toString()];
 				}
 
-				if ($member['member_ip'] != $member['member_ip2'] && isset($duplicate_members[$member['member_ip2']])) {
-					$members[$key]['duplicate_members'] = array_merge($member['duplicate_members'], $duplicate_members[$member['member_ip2']]);
+				if ($member['member_ip'] != $member['member_ip2'] && isset($duplicate_members[$member['member_ip2']->__toString()])) {
+					$members[$key]['duplicate_members'] = array_merge($member['duplicate_members'], $duplicate_members[$member['member_ip2']->__toString()]);
 				}
 
 				// Check we don't have lots of the same member.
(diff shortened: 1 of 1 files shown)


live627:
The explicit calls to that magic method makes the code look funny. Since that routine appears to work only with strings,  could you instead change `new Ip(/* ... */)` to either `new Ip(/* ... */)->ip` or `(string) new Ip(/* ... */)`?

sbulen:
Found a few spots I missed.

> The explicit calls to that magic method makes the code look funny. Since that routine appears to work only with strings, could you instead change `new Ip(/* ... */)` to either `new Ip(/* ... */)->ip` or `(string) new Ip(/* ... */)`?

I looked into this, & considered using the existing ->expand() method, but then some of the comparisons/dupe checks would fail.

I also considered adding an ->ip() method, but that would be inconsistent with the other classes we have, which explicitly define __toString() for string conversion such as this.

Sesquipedalian:
Just cast to string. Not only is that the simplest and most concise approach, but it also communicates the most information to developers reading the code later. When we see an object being cast to a string, we know we are getting a string and that the object must implement the Stringable interface, and we know it without needing to look up property types or method return types.

sbulen:
Done & retested.

It's a nit, but, something bugs me about requiring a type cast, it seems inconsistent, e.g., IP.php has toHex() and toBinary(), but instead of providing a similar toString(), we require a cast instead.  Requiring a cast on our own object feels like we missed something to me.

Then again, there aren't magic methods for __hex() & __binary()...  

And as long as we're consistent going forward, we're good.

Sesquipedalian:
Well, it's not that we _require_
a cast. Calling `__toString()` works. It's just a style preference that the majority of us share.

Personally, I wish PHP would add some more magic methods for casting to scalar values (e.g. `__toInt()`, `__toFloat()`, etc.) so that developers could use this approach more broadly for the reasons stated above.

Sesquipedalian (approved):

jdarwood007:
To be fair, in my opinion, our force casting should be fixed in the future.  For 3.0, using force casting is acceptable as rewriting lots to get the app into a app supported by singletons, so we can move to a more modern approach later that uses factories, compartmentalizes logic into smaller replaceable objects or other methods will be easier.

I would see the IP class as having a `ToString()`, `ToInet`, `ToBinary`, etc method to support use cases in our code.

live627:
I prefer explicit casting because it communicates intent in a familiar way
common to C-style languages such  as PHP,  C, C++, and C#..

PR #8325, opened 2024-10-15
Update min PHP version to 8.3

MissAllSunday:
Purposefully left out composer.json

`    "config": {
        "platform": {
            "php": "8.0.0"
        }
    }
`
    
    
Since it seems its used for internals


live627:
Can you clarify, please? What part of the code requires this?

MissAllSunday:
None, I opened a topic about it on Development Contributors board

live627:
The decision to have the min  at 8.0 was based on the max officially supported by SMF 2.0. Thus, 2.0 to 3.0 upgrades could be done on the same runtime.

Issue #7870, opened 2023-11-12
Search this topic does not deliver any results


7wells:
#### Description
Search `This topic` does not deliver any results.

### Steps to reproduce
1. Visit a topic.
2. Select `This topic` from drop-down menu.
3. Enter a search term in the search field.
4. Click `SEARCH` button.

![smf](https://github.com/SimpleMachines/SMF/assets/65889763/4148341b-bd70-429b-88ea-bd8bd39394b7)

### Environment (complete as necessary)
- Version/Git revision: 2.1.3
- Database Type: ??
- Database Version: ??
- PHP Version: ??

As I am not the board (https://forum.fhem.de) owner, I cannot provide details about database and PHP, sorry.

### Additional information/references
None

Oldiesmann:
Works fine for me on 2.1.4 at seniorsandfriends.org

Sesquipedalian:
Since no one else can reproduce the issue and @7wells gave a thumbs up to @Oldiesmann's reply, I am interpreting that to mean that the issue can be closed. If that is incorrect and the issue is still occurring for you, @7wells, please feel free to reply to this with more information about how we can reproduce the problem.

isaak654:
Issue confirmed by @sbulen in https://www.simplemachines.org/community/index.php?msg=4172803

> Yep, confirmed.  I think this is a bug.
> 
> Guests get the not allowed message.
> 
> Regular users get a no results message.
> 
> Admins get proper results.

Oldiesmann:
This affects SMF 3.0 as well.

Oldiesmann:
Seems to be fixed for 3.0 now with the latest search improvements

isaak654:
Please consider removing the “Confirmation needed” label and assigning a milestone for 2.1.x.

isaak654:
@Sesquipedalian Please consider removing the “Confirmation needed” label and assigning a milestone for 2.1.x.

sbulen:
From: https://www.simplemachines.org/community/index.php?msg=4181419
> Pretty sure it's not tied to the php or mysql version.  My test copy (php 8.3, mysql 8.4; I also tried php 7.4) shows the same behavior as production (php 8.2, mysql 8.0 percona).  My forum uses a custom index. 
> 
> In my forum, it doesn't always fail...  But there are patterns...
> 
> - It fails for certain terms in certain topics/boards, for non-admins & guests, in a repeatable fashion.
[...]

PR #8332, opened 2024-10-25
[2.1] Use real topic id

sbulen:
Partial for #7870 

This addresses 2.1, but not 3.0.


sbulen:
Did some additional testing to ensure unapproved messages were not displayed to users - & that works.  All good.

sbulen:
I'm kinda shocked how big the hole is here...

I'd recommend this for 2.1.6, if not 2.1.5.

sbulen:
I think this was superseded by #8458 & needs to be closed...

PR #8335, opened 2024-11-01
Fetch message for quotefast action via Msg object

live627:
### Motivation

I have a modification (Topic Descriptions) that needs access to this, so we'll use `Msg::load()` which eventually calls `integrate_query_message`

### Idea

`$query_customizations` could become its own object.

### DIFF
--- a/Sources/Actions/QuoteFast.php
+++ b/Sources/Actions/QuoteFast.php
@@ -80,7 +80,7 @@ class QuoteFast implements ActionInterface
 		$bq = User::$me->mod_cache['bq'];
 
 		if (isset($_REQUEST['modify']) || $bq != '1=1') {
-			$query_customizations['where'][] = 't.locked = {int:not_locked}' . ($bq == '0=1' ? '' : 'm.' . $bq);
+			$query_customizations['where'][] = 't.locked = {int:not_locked}' . ($bq == '0=1' ? '' : ' OR m.' . $bq);
 		}
 
 		$row = current(Msg::load((int) $_REQUEST['quote'], $query_customizations));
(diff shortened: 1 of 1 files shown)


jdarwood007:
Haven't run the code but looks good.

I'm in support of more objects for everything.  It makes things much easier to develop.  But we have to contend with how it would support customizations.  Magic methods maybe can help there.

DiegoAndresCortes:
I like it, same thing we got for messageindex, the more places the better imo

tyrsson:
From a customization perspective. If you model the instances that need to support customization as Observable then mods can be notified via their own registered listeners.

Sources/Actions/QuoteFast.php
```diff
@@ -47,52 +48,68 @@ class QuoteFast implements ActionInterface
...
-		Db::$db->free_result($request);
+		];
 
-		Utils::$context['sub_template'] = 'quotefast';
+		$bq = User::$me->mod_cache['bq'];
 
-		if (!empty($row)) {
-			$can_view_post = $row['approved'] || ($row['id_member'] != 0 && $row['id_member'] == User::$me->id) || User::$me->allowedTo('approve_posts', $row['id_board']);
+		if (isset($_REQUEST['modify']) || $bq != '1=1') {
+			$query_customizations['where'][] = 't.locked = {int:not_locked}' . ($bq == '0=1' ? '' : 'm.' . $bq);
```
Sesquipedalian:
```suggestion
			$query_customizations['where'][] = 't.locked = {int:not_locked}' . ($bq == '0=1' ? '' : ' OR m.' . $bq);
```

Is this what you meant to write?

Sesquipedalian (changes requested):

Sesquipedalian (approved):

PR #8336, opened 2024-11-03
Another strict types fix for SVGs

Sesquipedalian:


dragomano:
Just a note...

Using a [PHP-Typed package](https://github.com/WPLake/PHP-Typed), you can simplify such constructs - 
```php
!empty(Config::$modSettings['attachmentThumbHeight']) ? (int) Config::$modSettings['attachmentThumbHeight'] : 1000
```
to the following form:
```php
Typed::int(Config::$modSettings, 'attachmentThumbHeight', 1000)
```

PR #8346, opened 2024-11-26
Lets our entity handling methods work with all named HTML entities

Sesquipedalian:


jdarwood007:
Is there a source this comes from to know when it needs updates?  I don't think most humans could keep up with that.

Sesquipedalian:
It won't need updates. The values came from `get_html_translation_table(HTML_ENTITIES, ENT_QUOTES | ENT_HTML5)`, and the list of named entities for HTML5 is a fixed value that [won't be updated or expanded in the future](https://html.spec.whatwg.org/multipage/named-characters.html) (see note at bottom of page). If it weren't for that guarantee, I would have built the regex at runtime. But since the list is immutable, we can just hard-code the regex.

PR #8348, opened 2024-11-28
Improvements to iCalendar support for SMF's calendar

Sesquipedalian:
1. Fixes a bug where we were incorrectly handling negative BYYEARDAY values in iCalendar recurrence rules.
2. Includes additional relevant properties when exporting holidays as iCalendar files.
3. Improves conformance with RFC 5545 by includes VTIMEZONE data in exported iCalendar files.

Item 3 was a pain in the butt, but necessary to conform to the spec. (Personally, I think it is stupid that the spec requires including information about all the timing and rules that govern when a given time zone switches between standard time and daylight saving time, rather than just giving the time zone identifier and letting the calendar software handle it directly, but that's just the way it is.)


jdarwood007:
How annoying will it be to update those files?  Especially for patch releases?

sbulen:
There's a script to update them,  just like the unicode tables.

Auto correct really really wanted to make this post about unicorns.

Sesquipedalian:
> There's a script to update them, just like the unicode tables.

Correct. I have now added documentation notes to remind future developers about this.

PR #8370, opened 2024-12-23
Updates PHP-CS-Fixer rules from PSR-12 to PER, and applies them

Sesquipedalian:
Like it says on the tin...

https://www.php-fig.org/per/coding-style/


Sesquipedalian:
I'll take those thumbs up reactions as approval of the PR.

PR #8372, opened 2024-12-24
[2.1] Hooks list add support for leading backslash namespaces

MissAllSunday:
By default, hook list  in maintenance relies on the file property to determinate if a hook is  enable/active via setting its abs path.

For larger mods or mods that relies on autoloaders the file property might not be defined for all hooks, making them appear as "the hook has not been found" on the hook list.

This PR adds a check to see if a hook has its correspondence with the list of already loaded functions, if found it sets its absPath which in return the hook list will correctly list them as active.

### DIFF
--- a/Sources/ManageMaintenance.php
+++ b/Sources/ManageMaintenance.php
@@ -2109,10 +2109,11 @@ function get_integration_hooks_data($start, $per_page, $sort, $filtered_hooks, $
 	foreach ($filtered_hooks as $hook => $functions)
 		foreach ($functions as $rawFunc)
 		{
-			$hookParsedData = parse_integration_hook($hook, $rawFunc);
+			$hookParsedData = parse_integration_hook($hook, $rawFunc, $function_list);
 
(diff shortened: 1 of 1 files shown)


live627:
Which mod must I install to find the error? It seems to work with my Custom Form mod. What am I missing here?

MissAllSunday:
Ahhh I see it now, upon further inspection the only difference between your mod and mine (where the issue was reported) is a leading backslash in the namespace:

Yours:  'CustomForm\Integration::helpadmin'
Mine: '\PostLimit\PostLimitAlerts::handle'

Trimming it resolves the issue altogether so no need to pass the previous function_list to match.

I prefer usage of leading backslash to make it clear its an absolute path instead of a relative one but I left it to the team if this case should be taken into consideration or not.

live627:
well, sine php accepts that syntax perhaps we should too.

Sesquipedalian (approved):

PR #8375, opened 2024-12-25
[3.0] rewrite message links to use long form

### LINKED ISSUE #8367: [2.1], [3.0]: Google/Crawler Inefficiencies
### Basic Information

I'm sharing some observations here & a suggestion.  

A couple notes to start...
 - Google does not honor crawl-delay, they will hit your forum as often as they want.  Sometimes quite a lot...
 - Google does not honor nofollow.  So when crawling a page, it will follow all the links on the page.
 - Google only **indexes** pages with urls that follow the canonical form - for all SMF _topics & messages,_ that is: `https://forum.com/index.php?topic=######.##`.
 - Google search console documents the excluded non-canonical urls as duplicates, with a message:  "Alternate page with proper canonical tag - These pages aren't indexed or served on Google".

I believe most other crawlers exhibit similar behavior.  Some honor crawl-delay.  I don't see anybody honoring nofollow.  And from what I can see (e.g., Bing), they also only index canonicals.

Some notes on SMF...
 - For each message displayed on a page, SMF provides a link in the format `https://forum.com/index.php?msg=######`, with a nofollow.
 - When provided a "msg=" format url, SMF will issue a 302 redirect to a url with the format `https://forum.com/index.php?topic=######.msg######msg#####`.  
 - The redirect is OK, and that request returns a status code of 200.
 - When the page is built, the canonical format provided by SMF in the head is in the format `https://forum.com/index.php?topic=######.##`

So...  Putting this all together...  For a forum that is configured to show 25 posts on a page, the sequence of events looks like this:
 - Let's assume Google starts a crawl with `https://forum.com/index.php?topic=######.##`.  This is canonical form, so it is indexed.
 - This page has 25 messages on it, including one url per message, each of the form: `https://forum.com/index.php?msg=######`.
 - Google will then, 25x:
    - Request all message urls provided `https://forum.com/index.php?msg=######`
    - SMF says hold on, 302, you should really be using `https://forum.com/index.php?topic=######.msg######msg#####`
    - Google then requests the page `https://forum.com/index.php?topic=######.msg######msg#####`, but that page has a canonical of `https://forum.com/index.php?topic=######.##` in the head, so Google discards it.  Note this canonical URL will match what was on the very first line in this sequence.

So...  There was one successful page load & index.  However, there were 25 subsequent requests to SMF that resulted in redirects, and 25 more requests to SMF that were discarded...

**_Literally 50x the number of requests to the site for the actual content indexed._**  

**_Note that all the content was already properly indexed - all 25 messages were on the original `https://forum.com/index.php?topic=######.##` request.  It's just requesting that same content an additional 50x._**

Suggestions:
 - When building pages, don't post the links in `https://forum.com/index.php?msg=###### format`, use the `https://forum.com/index.php?topic=######.msg######msg#####` format, as that will avoid the 302s.  
 - An alternate approach would be to not issue a 302, just build the page.

I believe either of the above suggestions will reduce the # of crawler requests to the site by half.  

I don't know how to get Google to honor nofollow...  But they do honor robots.txt.  Maybe site admins can disallow msg= via robots.txt.  I don't believe there is a msg= valid canonical anywhere that would be indexed.

### Steps to reproduce

You can see all of the above in:
 - Your web access logs
 - Google Search Console
 - Google search results

### Expected result

Closer to a 1:1 relationship between requests & indexed URLs

### Actual result

Crawlers hit the site 50x times more than is necessary.

### Version/Git revision

3.0 alpha 2 & 2.1.4

### Database Engine

All

### Database Version

8.4

### PHP Version

8.3.8

### Logs

_No response_

### Additional Information

_No response_

live627:


Sesquipedalian:
Please see comments in the original issue.

Sesquipedalian:
Preferred solution: https://github.com/SimpleMachines/SMF/issues/8367#issuecomment-2562004948

Sesquipedalian:
This wouldn't solve the issue. https://github.com/SimpleMachines/SMF/issues/8367#issuecomment-2562002855

PR #8376, opened 2024-12-25
[3.0] Replaces uses of deprecated constant SID with calls to session_id()

### LINKED ISSUE #8365: [2.1, 3.0]: Constant SID is deprecated
### Basic Information

This error occurs periodically and is related to the code where the constant is used.

Example:

```php
if (defined('SID') && SID != '')
```

### Steps to reproduce

_No response_

### Expected result

_No response_

### Actual result

_No response_

### Version/Git revision

3.0 Alpha 2

### Database Engine

All

### Database Version

11.2.2-MariaDB

### PHP Version

8.4.1

### Logs

_No response_

### Additional Information

_No response_

Sesquipedalian:
Fixes #8365

### DIFF
--- a/Sources/QueryString.php
+++ b/Sources/QueryString.php
@@ -414,14 +414,14 @@ class QueryString
 	 */
 	public static function ob_sessrewrite(string $buffer): string
 	{
-		// If Config::$scripturl is set to nothing, or the SID is not defined (SSI?) just quit.
-		if (Config::$scripturl == '' || !defined('SID')) {
+		// If Config::$scripturl is set to nothing, or the session ID is not defined (SSI?) just quit.
+		if (Config::$scripturl == '' || session_id() === false) {
 			return $buffer;
 		}
 
 		// Do nothing if the session is cookied, or they are a crawler - guests are caught by redirectexit().
-		if (empty($_COOKIE) && SID != '' && !BrowserDetector::isBrowser('possibly_robot')) {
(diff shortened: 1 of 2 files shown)


sbulen:
I think this is an issue...  I believe the constant SID includes more than just session_id...   

From the PHP manual:
[SID](https://www.php.net/manual/en/session.constants.php#constant.sid) ([string](https://www.php.net/manual/en/language.types.string.php))
    Constant containing either the session name and session ID in the form of "name=ID" or empty string if session ID was set in an appropriate session cookie. This is the same id as the one returned by [session_id()](https://www.php.net/manual/en/function.session-id.php).

In other words, I think the replacement for the constant SID must look something like:  "PHPSESSID=yadayadayada".  session_id is only the yadayadayada part....

Note also that the "PHPSESSID" part can also be renamed in session.name as well, e.g., by the site admin.

Sesquipedalian:
Derp. You're right. The issue is more complicated than I realized.

jdarwood007:
session_name() gets the other part.

jdarwood007:
The bigger issue is that session_id() returns to possible invalid values.  false and an empty string.  Empty string if the session never started and false if an error occurred.

Maybe this is a good candidate to define in Session.php as a static function that handles things cleanly.

Sesquipedalian:
For now, I have reverted these changes and added a stop-gap solution for SMF 3.0 in #8387

PR #8392, opened 2025-01-02
exdates doesn't have a default value...

### LINKED ISSUE #8246: [3.0]: Installer - Line #1604: Field 'exdates' doesn't have a default value
### Basic Information

Get an error upon install.

### Steps to reproduce

1. Install current GitHub

### Expected result

No errors upon install, & calendar entries for holidays should be created.

### Actual result

Get this error:
Line # 1604: Field 'exdates' doesn't have a default value

And calendar entries for holidays are not created.

### Version/Git revision

3.0 Alpha 1 - current GH

### Database Engine

MySQL

### Database Version

8.4.0

### PHP Version

8.3.8

### Logs

_No response_

### Additional Information

image

### LINKED ISSUE #8380: [3.0]: "Field 'exdates' doesn't have a default value" when installing
### Basic Information

When installing SMF 3.0 with MySQL 9.1, you are given an error message stating that some queries didn't execute properly, with a message of "Line #1606: Field 'exdates' doesn't have a default value". This is because the "exdates" field is defined as NOT NULL, but the INSERT query to add initial data to the calendar table doesn't attempt to insert anything in that field

### Steps to reproduce

1. Try to install SMF 3.0 with MySQL 9.1 (default setup with WAMPServer in this case)
2.

### Expected result

Expect forum to install properly

### Actual result

The following error/warning is given instead:
image

### Version/Git revision

2eeafd1

### Database Engine

MySQL

### Database Version

9.1.0

### PHP Version

8.4.2

### Logs

_No response_

### Additional Information

We need to either explicitly insert an empty string into that column, specify a default value or allow it to be NULL, though I don't know if SMF will handle NULL values.

Oldiesmann:
Fixes #8235. Closes #8380 and closes #8246


Sesquipedalian (approved):

Sesquipedalian:
Nicely done, @Oldiesmann. I figured there had to be something we were doing wrong regarding the exdates field in particular, and you found it.

PR #8395, opened 2025-01-04
Extending actions

live627:
### Motivation

The goal is to reduce reliance on static properties for actions requiring special behavior.

---

### Interface Definition

Below is the interface for defining action-specific behavior (`ActionInterface`). Method names probably could be improved:

```php
interface ActionInterface
{
	/**
	 * Checks if this action allows guest access even when guest access is disabled.
	 *
	 * @return bool True if access is allowed, false otherwise.
	 */
	public function isRestrictedGuestAccessAllowed(): bool;

	/**
	 * Determines if this action can be logged in the online activity log.
	 *
	 * @return bool True if the action can be logged, false otherwise.
	 */
	public function canBeLogged(): bool;

	/**
	 * Identifies whether this action qualifies as a "simple action."
	 *
	 * @return bool True if the action is simple, false otherwise.
	 */
	public function isSimpleAction(): bool;

	/**
	 * Checks whether this action requires XML mode to be enabled.
	 *
	 * @return bool True if XML mode is required, false otherwise.
	 */
	public function isXmlAction(): bool;

	// Additional methods can be added here as needed.
}
```

### Deprecations

1. **`integrate_pre_log_stats` (modifying `Forum::$unlogged_actions`)**  
   - **Deprecated**: Modifying global `$unlogged_actions` via hooks.  
   - **Replacement**: Implement `ActionInterface::canBeLogged()` to manage logging at the action level.

2. **`integrate_guest_actions` (modifying `Forum::$guest_access_actions`)**  
   - **Deprecated**: Adding guest-access actions via hooks.  
   - **Replacement**: Use `ActionInterface::isRestrictedGuestAccessAllowed()` for guest access control.

3. **`integrate_simple_actions` (modifying `$simpleActions`, `$simpleAreas`, `$simpleSubActions`)**  
   - **Deprecated**: Managing "simple actions" dynamically through hooks and global arrays.  
   - **Replacement**: Implement `ActionInterface::isSimpleAction()` for explicit control.

4. **`integrate_simple_actions` (modifying `$extraParams`, `$xmlActions`)**  
   - **Deprecated**: Using hooks to flag XML actions or manage extra parameters.  
   - **Replacement**: Use `ActionInterface::isXmlAction()` for defining XML requirements.

### Updates to `integrate_actions`

- **Current Behavior**: String-based method references in the `$actions` array will continue to work as they currently do.  
- **New Feature**: Actions can now specify class references, such as `Actions\SomeClass::class`, if the class implements the `ActionInterface`.

This addition provides a smooth transition to the new approach while maintaining compatibility with existing logic.

### Backward Compatibility Impact

No backward compatibility issues are expected here. The old logic based on hooks and global properties will remain functional. Developers can incrementally adopt the new interface without requiring immediate changes to legacy code.

### Request for Comments

Please share your thoughts or additional suggestions.

### Question

Should we add a method for default action selection in the admin panel? This method would allow an action to specify itself as the default option for related settings in the admin panel.


jdarwood007:
I would change `isXmlAction` to something like `returnType` and define a few constants
```
const RETURN_TYPE_HTTP = 0;
const RETURN_TYPE_XML = 1;
const RETURN_TYPE_JSON = 2;
```

This would allow us to define that we have multiple types that can be easily modified.

Sesquipedalian:
I like what I'm seeing here!

Sources/Forum.php
```diff
@@ -143,26 +160,12 @@ class Forum
...
+	public static array $unlogged_actions = [
 		'about:unknown' => true,
+		'calendar' => ['sa' => ['clock']],
 		'clock' => true,
-		'dlattach' => true,
-		'findmember' => true,
-		'helpadmin' => true,
 		'jsoption' => true,
-		'likes' => true,
 		'modifycat' => true,
```
live627: These actions cannot use the new approach because they are simple callbacks that piggyback off other actions. any reason that they do not use their own objects? The only reason I can think of is that they are easter eggs and therefore require hunting around.

live627:
> I would change `isXmlAction` to something like `returnType` and define a few constants
> 
> ```
> const RETURN_TYPE_HTTP = 0;
> const RETURN_TYPE_XML = 1;
> const RETURN_TYPE_JSON = 2;
> ```
>
> This would allow us to define that we have multiple types that can be easily modified.

This is a good start, but is a limiting design. There are more MIME types to use, so something like the Strategy pattern might be a better fit

jdarwood007:
I'm open to whatever you think works best.  I am mostly thinking that limiting it to just Xml acction prevents us from migrating to JSON (and still supporting XML for BC), indicating other output formats like js/css/wav/etc.  We don't need to implement the entire mime-type spectrum though.

live627:
Right. We just need three classes: HTML, XML, JSON. A mod will  be able  to  add more, like JSON+LD or whatever.

jdarwood007:
Well we don't even need json right now since we don't have it in our code.  It will just set up the extendability of it for when we do.

live627:
the likes interface uses json.  I only discovered that after digging around after your suggestion of return types.

jdarwood007:
So we do have some.  Well that changes my plan to introduce json later.  We can then work on replacing our XML-based responses with json objects, which should simplify/improve some js logic.

The simple actions you mentioned you are unsure what to do, are easter eggs as you suspected.

live627:
This one's ready for review

Sources/Utils.php
```diff
@@ -2400,17 +2400,24 @@ public static function obExit(?bool $header = null, ?bool $do_footer = null, boo
...
-				// (since this is just debugging... it's okay that it's after </html>.)
-				if (!isset($_REQUEST['xml'])) {
-					Logging::displayDebug();
+				if (Forum::getCurrentAction()?->isSimpleAction() === false
+					|| (
+						Forum::getCurrentAction() === null
+						&& !isset($_REQUEST['xml'])
+					)
+				) {
+					DebugUtils::displayDebug();
```
dragomano: There is no `displayDebug` method in DebugUtils

Sources/ActionInterface.php
```diff
@@ -28,6 +28,58 @@ interface ActionInterface
...
+	 * @return bool
+	 */
+	public function isSimpleAction(): bool;
+
+	/**
+	 * Gets the output type for this action.
+	 *
+	 * @return OutputTypeInterface
+	 */
+	public function getOutputType(): OutputTypeInterface;
```
dragomano, line 66: Where is this interface?

Sources/Forum.php
```diff
@@ -353,79 +417,51 @@ protected function main(): array|string
...
-					// Sorry, only one default action is needed.
-					$defaultAction = $defaultAction[0];
-
-					$call = Utils::getCallable($defaultAction);
+					$default_action = explode(',', Config::$modSettings['integrate_default_action']);
 
-					if (!empty($call)) {
-						return $call;
-					}
+					return Utils::getCallable($default_action[0]);
```
dragomano: Considering that `Utils::getCallable` returns `callable|false`, shouldn't this method also return `callable|false`?

dragomano (commented):
It looks beautiful. I hope backward compatibility with mods for SMF 2.1 won't be affected.

tyrsson:
I think canBeLogged and isRestrictedGuestAccessAllowed should be split into their own interfaces. Well, the guest access check.... I'm not 100% certain on that one, but the canBeLogged should 100% be its own interface. Psr-3 already exposes an interface for this. LoggerAwareInterface, and a Trait for its implementation LoggerAwareTrait.

The interface:
```php
<?php

namespace Psr\Log;

/**
 * Describes a logger-aware instance.
 */
interface LoggerAwareInterface
{
    /**
     * Sets a logger instance on the object.
     */
    public function setLogger(LoggerInterface $logger): void;
}
```
The Trait
```php
<?php

namespace Psr\Log;

/**
 * Basic Implementation of LoggerAwareInterface.
 */
trait LoggerAwareTrait
{
    /**
     * The logger instance.
     */
    protected ?LoggerInterface $logger = null;

    /**
     * Sets a logger.
     */
    public function setLogger(LoggerInterface $logger): void
    {
        $this->logger = $logger;
    }
}
```
Please, take a moment and seriously consider the implications of this. They are a lot farther reaching than they first appear.

live627:
How would that  be applicable to writing the current action to the log of online users? Wouldn't  a PSR-3 logger be better suited for the error log? Or the upgrader?

Sources/Forum.php
live627: Good point. Should probably be `string|callable|false` to accommodate class names .

Sesquipedalian:
I've added some more changes to clean up and further standardize actions. The main thing is that all cases of `?action=...` URL parameters are now mapped to proper SMF\ActionInterface classes.

Sesquipedalian:
Assuming you approve of my changes, @live627, I think we can merge this.

Sources/Forum.php
```diff
@@ -353,78 +637,61 @@ protected function main(): array|string
...
 			elseif (empty(Topic::$topic_id)) {
-				return 'SMF\\Actions\\MessageIndex::call';
+				$_REQUEST['action'] = 'messageindex';
 			}
-
-			// Board is not empty... topic is not empty... action is empty.. Display!
+			// Board is not empty... topic is not empty... action is empty... Display!
 			else {
-				return 'SMF\\Actions\\Display::call';
+				$_REQUEST['action'] = 'display';
```
live627: Any reason to fix the request here? I'm going to change this to a parameter variable.

Sources/Forum.php
```diff
@@ -336,14 +611,23 @@ protected function main(): array|string
...
+	 *
+	 * This method handles the main forum logic, such as checking permissions,
+	 * logging user activity, and tracking forum statistics.
+	 */
+	protected function main(): void
+	{
 		// If we are in a topic and don't have permission to approve it then duck out now.
 		if (!empty(Topic::$topic_id) && empty(Board::$info->cur_topic_approved) && !User::$me->allowedTo('approve_posts') && (User::$me->id != Board::$info->cur_topic_starter || User::$me->is_guest)) {
 			ErrorHandler::fatalLang('not_a_topic', false);
 		}
```
live627, line 626: Would this be able to be decoupled from here now?

Sources/Forum.php
Sesquipedalian: Generally speaking, I was going for greater consistency. Given that BoardIndex, MessageIndex, and Display behave like actions in every way except that they usually aren't accessed via `?action=...`, it's odd to have them behave differently in this regard.

More specifically, I have been working for the last little while on implementing a robust routing system (a.k.a. pretty URLs), and as part of that it is helpful to have fully defined and consistent behaviour for mapping between URL query parameters and the called ActionInterface instances. Even if `$_REQUEST['action']` isn't set from the real query string, it is a lot easier to build routes for BoardIndex, MessageIndex, and Display if it is.

Sesquipedalian: That said, if you are planning to ditch direct use of `$_REQUEST['action']` in favour of something better, by all means do so. As long as BoardIndex, MessageIndex, and Display work like every other action in your new system, that will be fine.

For example, if your plan is to set a property on SMF\Forum (or SMF\QueryString or something) and then replace all references to `$_REQUEST['action']` in the rest of the codebase with references to that property, then all I would like to see is for that property to be set to the appropriate key in SMF\Forum::$actions for  BoardIndex, MessageIndex, or Display as necessary.

Sources/Forum.php
Sesquipedalian: Perhaps, yeah. Where would you rather do this?

live627: Probably the equivalent in `Topic` of `Board::load()`

Sources/Forum.php
live627: I just want to  avoid writing to superglobals if at all possible.

Sources/Forum.php
Sesquipedalian: Sounds good

live627:
There's also code here that skips loading the theme for `dlattach`... I want to make a new method in the interface.  Elkarte calls it `needsTheme` but I should probably name it `canLoadTheme`

Sources/Forum.php
Sesquipedalian: That makes sense. The best way to do that would be to replace all references to `$_REQUEST['action']` with something like `QueryString::get('action')`... which is a change that would definitely go outside of the scope of this PR. 

So, sure, revert those writes to `$_REQUEST['action']` for now, and I will come up with something to implement a getter (and maybe a setter) for QueryString in a different PR later.  But please leave the entries for `boardindex`, `messageindex`, and `display` in Forum::actions.

live627: I already pushed a  commit that shunted it to a variable.

live627:
> There's also code here that skips loading the theme for `dlattach`... I want to make a new method in the interface. Elkarte calls it `needsTheme` but I should probably name it `canLoadTheme`

Oh, problem. Several actions do things with the theme among other non-constructory things in their constructors. This addition will have to wait for another pull request where I can find all of the occurrences and send them off to be `execute()`d.



> Assuming you approve of my changes, @live627, I think we can merge this.

Aye, aye, cap'n!

tyrsson:
> How would that be applicable to writing the current action to the log of online users? Wouldn't a PSR-3 logger be better suited for the error log? Or the upgrader?

SMF already uses a Logging class does it not?

The corresponding LogLevel would be 'info'.

As an example if the logger leverages "channels" such as Monolog does, then the channel would be "stats" and the factory would push a StatsHandler. This handler would be responsible for determining which table to write too. 

Interestingly I see that SMF is already using an "extra" array.

tyrsson:
> > I would change `isXmlAction` to something like `returnType` and define a few constants
> > ```
> > const RETURN_TYPE_HTTP = 0;
> > const RETURN_TYPE_XML = 1;
> > const RETURN_TYPE_JSON = 2;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This would allow us to define that we have multiple types that can be easily modified.
> 
> This is a good start, but is a limiting design. There are more MIME types to use, so something like the Strategy pattern might be a better fit

Note: This only addresses the overall response types, not all mime types for such things as images etc.

I would say this should exist as Response type classes. Ie as extenders of a Response class which implements the Psr\Http\Message\ResponseInterface.

Two implementers that come to mind (just the two I use most often):
https://github.com/laminas/laminas-diactoros
https://github.com/guzzle/guzzle 

They are all http response content types. Even in this usage context I would say that they would be more useful if they atleast contained the actual content type header field value that ultimately they will or should be used for ie text/html, application/json etc.

Short of actually implementing any of that. I would suggest something like the following.

Just a quick 5 minute skeleton (its all in a single file just as an example)

```php
<?php

declare(strict_types=1);

interface ContentTypeInterface
{
    public final const HEADER_FIELD = 'Content-Type';
    public final const TYPE_HTML    = 'text/html';
    public final const TYPE_JSON    = 'application/json';
    public final const TYPE_TEXT    = 'text/plain';
    public final const TYPE_XML     = 'application/xml';

    public final const SUPPORTED_TYPES = [
        self::TYPE_HTML,
        self::TYPE_JSON,
        self::TYPE_TEXT,
        self::TYPE_XML,
    ];

    public const RESPONSE_TYPE = self::TYPE_HTML;

    public function getContentType(): string;
}

abstract class ContentType implements ContentTypeInterface
{
    public function __construct()
    {
        \header(self::HEADER_FIELD . ': ' . $this->getContentType());
    }

    final public function getContentType(): string
    {
        if (!\in_array(static::RESPONSE_TYPE, self::SUPPORTED_TYPES)) {
            throw new \RuntimeException('Unsupported content type');
        }
        return static::RESPONSE_TYPE;
    }
}

class Html extends ContentType
{
    public const RESPONSE_TYPE = self::TYPE_HTML;
    // do other stuffz
}

echo (new Html())->getContentType();
```

The typed response class SHOULD enforce the content type on the response. Hence the header. I know SMF can make this extremely difficult at times. That, itself, is something that needs to be addressed when the template rewrite begins. You will find, and I hate to sound like a broken record, but SMF often does the wrong things, at the wrong times, I suppose I should say in the wrong layer of the application. Which is why there is markup in source files among other issues. If I remember correctly and there is a good chance I may not. Seems like the json response handling was one of the issues I faced when running SMF as a Not Found Handler for a psr7/15 application to allow refactoring to that paradigm ahead of SMF proper.... Its been awhile so forgive me if recall wrong there. Basically the issue was "when" in processing the response was started iirc.

Or maybe I am completely misunderstanding the problem being solved and if so, please ignore me.

live627:
> I would say this should exist as Response type classes.

Very similar to what I ended up designing https://github.com/SimpleMachines/SMF/pull/8395/files#diff-050252f54994f6af000b9a45fef8c33777e03e33d7ab04ff931a3c1ea065139c seperate objects for each of the three output types:  `OutputTypes/Html`, `OutputTypes/Xml`, and `OutputTypes/Json`.

> hate to sound like a broken record, but SMF often does the wrong things, at the wrong times, I suppose I should say in the wrong layer of the application.

Implementing the various interfaces proposed by PHP-FIG is a noble goal but should probably wait until application layers can be detangled. Careful planning must be done to fix some of this  tight coupling. I'm not so great at quick grandiose changes, but rather slow methodical changes,like an evolution. Jumping the gun and implementing some PSRs  immediately might lead to some bad implementations.

b23fx1wtmzac1

tyrsson:
I never said that you should actually have a dependency on the PSR. What I was saying is that if you can, at this moment, align the code that is being developed today with the PSR they will be easier to adopt in the future. I suppose I should have spelled that out. I didn't think that would be necessary. Lesson learned.

I had a very long reply typed for this and then decided to approach it from a different direction.

Please explain to me the significance of this:
```php
     final public function getContentType(): string
    {
        if (!\in_array(static::RESPONSE_TYPE, self::SUPPORTED_TYPES)) {
            throw new \RuntimeException('Unsupported content type');
        }
        return static::RESPONSE_TYPE;
    }
```
What does it accomplish and why?

What are the implications for your mod authors?

What are the implications for consumers of your interface vs the consumers of my implementation?

Answer those questions fully and you will understand the difference in the code I suggested and the code that was actually committed. They are really only similar in respect to class names. 

P.S.
I find your choice of meme usage very ironic.

tyrsson:
> Implementing the various interfaces proposed by PHP-FIG is a noble goal but should probably wait until application layers can be detangled

One of the reasons those interfaces exist is for the purpose of detangling application layers.

tyrsson:
>  Careful planning must be done to fix some of this tight coupling.

Do you really think I do not understand this? I probably understand it at least as well as anyone involved and probably better than most.

See the below as a natural progression of my suggestions above where $html would be the return of the SMF template engine rendering.
```php
<?php

declare(strict_types=1);

interface ContentTypeInterface
{
    public final const HEADER_FIELD = 'Content-Type';
    public final const TYPE_HTML    = 'text/html';
    public final const TYPE_JSON    = 'application/json';
    public final const TYPE_TEXT    = 'text/plain';
    public final const TYPE_XML     = 'application/xml';

    public final const SUPPORTED_TYPES = [
        self::TYPE_HTML,
        self::TYPE_JSON,
        self::TYPE_TEXT,
        self::TYPE_XML,
    ];

    public const RESPONSE_TYPE = self::TYPE_HTML;

    public function getContentType(): string;

    public function getBody(): string;
}

abstract class ContentType implements ContentTypeInterface
{
    protected string $body;

    public function __construct($body)
    {
        \header(self::HEADER_FIELD . ': ' . $this->getContentType());
        $this->body = $body;
    }

    final public function getContentType(): string
    {
        if (!\in_array(static::RESPONSE_TYPE, self::SUPPORTED_TYPES)) {
            throw new \RuntimeException('Unsupported content type');
        }
        return static::RESPONSE_TYPE;
    }

    public function getBody(): string
    {
        return $this->body;
    }
}

class Html extends ContentType
{
    public const RESPONSE_TYPE = self::TYPE_HTML;
    // do other stuffz

}

class Text extends ContentType
{
    public const RESPONSE_TYPE = self::TYPE_TEXT;
    // do other stuffz

}

class Json extends ContentType
{
    public const RESPONSE_TYPE = self::TYPE_JSON;
    // do other stuffz

    public function getBody(): string
    {
        return \json_encode($this->body);
    }
}

$html = '<!DOCTYPE html><html><head><title>Hello, World!</title></head><body><h1>Hello, World!</h1></body></html>';
$text = 'Hello, World!';
$json = ['foo' => 'bar'];

echo (new Html($html))->getBody();
```

Sesquipedalian:
If I may, @tyrsson, I think what is going on here is simply this: 

You are presenting useful and valid concepts as ideals and goals to aim for, in the hope and conviction that they will be helpful. And as ideal concepts and structures to aim for, they are helpful.

The disconnect arises from the fact that the work the rest of us are doing is trying to evolve and modernize a codebase that was written under a very different paradigm, _without breaking it in the process_. 

That last constraint makes all the difference here. The simple truth is that, within the confines of that constraint, theoretical discussions and proofs of concept about how this or that job could be done if we were starting from scratch have a limited degree of utility.

More valuable to us than a thousand theoretical discussions would be a pull request from you in which you took some element or structure of SMF's current code and refactored it in order to implement these concepts in fully working code. If you were to do that, I entirely expect that the resulting PR would be awesome and very warmly received.

PR #8397, opened 2025-01-06
Update docs for `Utils::getCallable()`

live627:


Sources/Utils.php
```diff
@@ -2428,94 +2428,81 @@ public static function obExit(?bool $header = null, ?bool $do_footer = null, boo
...
 		}
 
-		// No full objects, sorry! pass a method or a property instead!
-		if (is_object($input)) {
-			return false;
-		}
-
-		// Stay vitaminized my friends...
+		// Sanitize and trim the input.
 		$input = Utils::htmlspecialchars(Utils::htmlTrim($input));
```
live627, line 2458: Should this  be  here?

Sources/Utils.php
```diff
@@ -2428,94 +2428,81 @@ public static function obExit(?bool $header = null, ?bool $do_footer = null, boo
...
 	 *
-	 * @param mixed $input Input to parse to find a callable.
-	 * @return mixed Either a callable, or false on failure.
+	 * @param string|array $input Input to parse as a callable.
+	 * @param bool|null $ignore_errors Optional. Whether to suppress errors if the callable is invalid. Defaults to the value of `Utils::$context['ignore_hook_errors']`.
+	 *
+	 * @return callable|false Returns the callable if valid, or false on failure.
 	 */
-	public static function getCallable(mixed $input, ?bool $ignore_errors = null): mixed
+	public static function getCallable(string|array $input, ?bool $ignore_errors = null): callable|false
```
Sesquipedalian, line 2449: Is callable a supported return type in PHP 8.0? (I can't check at the moment.)

live627: Yes; IIRC it was introduced in 7.0 along with the basic scalar types.

jdarwood007: I believe so and all checks pass.  Perhaps it is the first class callable syntax that you are thinking of which was 8.1: https://www.php.net/manual/en/functions.first_class_callable_syntax.php

Sesquipedalian: Yes, thanks.

Sesquipedalian (approved):

PR #8399, opened 2025-01-07
[3.0] Remove PHPSESSID

### LINKED ISSUE #8383: [2.1] & [3.0]: Consider disabling URL-based sessions, getting rid of PHPSESSID from URLs

sbulen:
Partial for https://github.com/SimpleMachines/SMF/issues/8383 (addresses 3.0 only)

This PR removes PHPSESSID URL handling altogether from SMF. It also addresses the 8.4 SID & session parameter deprecation issues, that are related. The point of the 8.4 changes was to eventually remove PHP's PHPSESSID URL processing.

It goes a bit further and will not write sessions where cookies are not present/allowed. I believe this last part can significantly decrease MySQL CPU workload for websites that are getting crawled heavily (like mine).

Testing thus far is fine. Things work OK with & without cookies. No impact to logon/logoff. No impact to queryless URLs. And there are DEFINITELY far fewer session records, that I believe were all unused anyway.

**_Loss of functionality:_**
I believe the only functions that are no longer supported are:
- Guest posts with cookies disallowed
- You cannot view who likes posts with cookies disallowed
- Search verification with cookies disallowed (search works OK without verification)
- New user registrations with cookies disallowed (not really a loss, because they can't login with cookies disallowed)


sbulen:
See #8394  for 2.1 version.

Sources/Session.php
```diff
@@ -96,6 +96,11 @@ public function read(string $session_id): string
 	 */
 	public function write(string $session_id, string $data): bool
 	{
+
+		// Don't both writing the session if cookies are diabled
```
```suggestion
		// Don't bother writing the session if cookies are disabled
```

sbulen: Done!

Sources/Session.php
```diff
@@ -96,6 +96,11 @@ public function read(string $session_id): string
 	 */
 	public function write(string $session_id, string $data): bool
 	{
+
+		// Don't bother writing the session if cookies are diabled
```
isaak654:
```suggestion
		// Don't bother writing the session if cookies are disabled
```

sbulen: Done, thanks

sbulen:
Resolved conflicts.

Sources/Session.php
```diff
@@ -96,6 +96,11 @@ public function read(string $session_id): string
 	 */
 	public function write(string $session_id, string $data): bool
 	{
+
+		// Don't bother writing the session if cookies are disabled
+		if (empty($_COOKIE))
+			return true;
```
Sesquipedalian: This makes PHP-CS-Fixer sad. Please change to:

```suggestion
		if (empty($_COOKIE)) {
			return true;
		}
```

Sesquipedalian (changes requested):
> If we want something cleaner, another option would be to not display the search & post (& login) buttons when cookies are not found. Or an error popup, saying cookies are required. Open to feedback.

An error popup is a good idea. This should be implemented in this PR before we merge it.

Sources/Session.php
sbulen: I'll take your word for it!  I figured I passed php-cs-fixer given the green checkmark on the PR.

I was never able to get php-cs-fixer to work acceptably locally.  I have it installed, but it seems to want to do very strange stuff.  E.g., in this example, it changed 90% of tabs to spaces.  (Not all of them...)  It did add the braces, though.

No plugins for the tools I use either (Notepad++, Eclipse).  I can integrate by hand, but I get the same odd results, e.g., lots of tabs become spaces.   So I don't use it.

jdarwood007: Shouldn't have to install it if composer has ran
```
composer lint
composer lint-fix
```

I also added it to my git as an alias
```
        phpcs = "!f() { ./vendor/bin/php-cs-fixer --config=.php-cs-fixer.dist.php list-files | xargs -n 50 -P 8 ./vendor/bin/php-cs-fixer fix --config=.php-cs-fixer.dist.php --path-mode intersection -v; }; f"```
```
Note that the 8 is concurrent threads.

I started to test adding it to .git/hooks/pre-commit
```
#!/bin/sh
git phpcs
```
I hope to get it to run before I issue a commit, but I haven't gotten around to testing it more. I need to work on adding it when I stage the commit.

sbulen: Our instructions said to install it, so I installed it.  I had to install composer, also...  I'm not familiar with either tool.

But it did install, & it does make the changes.  But it also screws up the whitespace.  

Does it change most of your tabs to spaces also?  Or is that just a me thing?  😆

sbulen: Hmmm....  i see with your example above, you tell it to use our rules with a config override (--config=.php-cs-fixer.dist.php)....

That wasn't mentioned anywhere...  I'll experiment with that as well.

jdarwood007: Correct, it needs our config or it applies the defaults.  Why people like spaces over tabs is beyond me as tabs can be adjusted in the editor to favor whatever spacing you prefer.

sbulen: And yes, I just confirmed that it works.  THANX!!!

The notes over on the Development Contributors board probably should reflect this.

I wrote a little .bat file that is invoked by Notepad++ and it works just fine now.

Sources/Session.php
```diff
@@ -96,6 +96,12 @@ public function read(string $session_id): string
 	 */
 	public function write(string $session_id, string $data): bool
 	{
+
```
Sesquipedalian:
```suggestion
```

Sesquipedalian (dismissed):
We might as well rename `ob_sessrewrite` to something like `obDebug` while we're at it.

Also, it would be nice to add that error popup you suggested earlier, if you are willing to write the code for that.

sbulen:
I'm looking into the popups.  While testing this code with various options, I discovered two new wrinkles:
 - The captcha for new user registrations doesn't work either
 - The login error message for no cookies changed...  It used to give you a nice 'please check your cookies' error, now it gives a session error.

The latter issue is a bit of a headscratcher, as the session issue is encountered far earlier, before the cookie check or the logon action is invoked.  I.e., can't just fix it with a simple edit in the logon action.  And 3.0 behaves differently than 2.1 here - the session error is buried in the browser console & not actually displayed.

...So, logically, I spent the last 3 days trying to get xdebug to work without cookies to debug properly.  Most of the info out there is out of date & predates xdebug3, and actually broke my IDE pretty bad, leading me to upgrade my out-of-date IDE...  🙄

(I put the water on to boil & started peeling the potatoes, while peeling the potatoes, the phone rang, while on the phone, the doorbell rang, opening the door, the dog escaped...)

Sesquipedalian:
Oof. I didn't think it would be quite so miserable to do that. Thanks for being willing to slog through it, @sbulen!

sbulen:
Question...  in Subs-Compat.php, ob_sessrewrite points to ob_sessrewrite.  But ob_sessrewrite only does part of what it used to do - it no longer does the queryless URLs function.

In Subs-Compat.php, should ob_sessrewrite call both functions?

Sesquipedalian:
Good catch. Yes, it probably should.

sbulen:
Notes on the recent commits (same as in 2.1):

With cookies disabled, & PHPSESSID removed from URLs:
 - Captchas did not work - the image was missing on the screens. This affected guest posts, guest searches, & new user registrations.
 - Even when captchas were not required, guest posts did not work & registrations did not work - they errored out on session errors.
 - Guest searches work fine if captcha is not required.

Note that if you went to the action under these circumstances, the screen was incomplete - it was missing the capcha image. Since it can look quite odd/confusing, partially painting the screen of a function they can't use, it is a little tidier to just not let them go there.

Note also that the login logic needed a little tweaking - checking the session before loading the theme made it difficult to provide a meaningful 'check your cookies' error upon login. A workaround is proposed here.

_**The posting function cannot be tested at the moment in 3.0, I believe there are permissions issues for guests...**_  See #8615

Themes/default/Display.template.php
```diff
@@ -813,9 +813,19 @@ function template_single_post($message)
...
+
+			// Remove link if no cookies; session reference won't work
+			if (empty($_COOKIE)) {
+				$like_text = Lang::getTxt($base, ['url' => '', 'num' => $count], file: 'General');
+				$like_text = strtr($like_text, ['<a href="">' => '', '</a>' => '']);
+			}
+			else {
+				$like_text = Lang::getTxt($base, ['url' => Config::$scripturl . '?action=likes;sa=view;ltype=msg;like=' . $message['id'] . ';' . Utils::$context['session_var'] . '=' . Utils::$context['session_id'], 'num' => $count], file: 'General');
+			}
+
```
Sesquipedalian:
```suggestion
			$like_text = Lang::getTxt($base, ['url' => Config::$scripturl . '?action=likes;sa=view;ltype=msg;like=' . $message['id'] . ';' . Utils::$context['session_var'] . '=' . Utils::$context['session_id'], 'num' => $count], file: 'General');

			// Remove link if no cookies; session reference won't work
			if (empty($_COOKIE)) {
				$like_text = preg_replace('~</?a\b[^>]*>~', '', $like_text);
			}

```

This will ensure that the anchor tags are removed correctly even if the strings are changed in the future.

Sesquipedalian (changes requested):

Themes/default/Display.template.php
sbulen: Done

PR #8401, opened 2025-01-07
[3.0] GitHub Action to Update Year in Files

live627:
Copy of https://github.com/SimpleMachines/SMF/pull/8400

Merge this before that.

### DIFF
--- a/.github/workflows/update-year.yml
+++ b/.github/workflows/update-year.yml
@@ -1,0 +1,28 @@
+name: Update Year
+
+on:
+  workflow_dispatch:
+
+jobs:
+  update-year:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Set current year environment variable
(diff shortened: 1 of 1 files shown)


Sesquipedalian (changes requested):
Please also update SMF_SOFTWARE_YEAR in `other/install.php` and `other/upgrade.php`, and also update the year in the license blocks in all PHP files everywhere.

Once 3.0 is released and enters the maintenance phase, we will change this action so that it only updates the license blocks for the files listed [here](https://github.com/SimpleMachines/BuildTools/blob/75964af7cd4a7ed44130ba6daeeaf9780f7b9aa3/check-smf-license.php#L62-L68). But during the pre-release phase, we want to update license blocks of all files.

live627:
I see your specification and I raise you a smarter bot that will transition from updating all files to only the few based on `SMF_VERSION` in `./index.php`. We might forget to update it when the time comes.

Sesquipedalian:
> I see your specification and I raise you a smarter bot that will transition from updating all files to only the few based on `SMF_VERSION` in `./index.php`. We might forget to update it when the time comes.

Very nice!

Sesquipedalian (approved):

PR #8403, opened 2025-01-13
Moderators should see when users were last online

MegaBrutal:
Moderators should see when users were last online, even if they are hidden. In the off-chance you don't agree, I suggest it to be a configurable option or a permission.


Oldiesmann:
Interesting. The help text for the moderate_forum permission indicates that this should already be the case, but obviously it isn't. Given everything else that permission gives you access to though, this makes sense.
image

jdarwood007:
It was changed here: https://github.com/SimpleMachines/SMF/commit/19b875368b5e2b41caaaa7717fa5c24a9665b494

Sesquipedalian:
> It was changed here: https://github.com/SimpleMachines/SMF/commit/19b875368b5e2b41caaaa7717fa5c24a9665b494

The discussion in the issue that was fixed by that commit shows debate over whether this should require admin_forum or moderate_forum permission. The commit used admin_forum, but there doesn't seem to be any indication of why that one was chosen—or at least, none that I noticed.

To me, moderate_forum makes more sense. Does anyone object? In particular, @live627, since you authored 19b8753, do you have any thoughts on this?

Oldiesmann (commented):
You don't need to change the copyright year. The developers will fix it themselves in the future

Sesquipedalian:
@MegaBrutal, to resolve the issue with the CI check regarding the copyright year, please merge the latest changes from [SimpleMachines:release-2.1](https://github.com/SimpleMachines/SMF/tree/release-2.1) into your pull request.

live627 (approved):

MegaBrutal:
I'm getting conflicting messages about whether I should fix the copyright year or not. Or is it optional? I could do it, but I already have 2 reviews, I'm afraid they would be dropped which would restart the process.

live627:
No,  don't update the year. Your branch should just be focused on the
isssue initially set out to solve.

Sesquipedalian:
1. In your local repository, drop commit 39625eef724011c3dba2c88c5e693b7b9065868a.
2. In your local repository, merge the upstream changes from the release-2.1 branch.
3. Force push your local repository to your GitHub fork, which will update this PR.

Note: generally speaking, force pushing should be avoided. In this particular case, however, it is what we want. The alternative would be to close this PR and submit another with only the changes you intended to make, but as you rightly noted, that would require a new set of reviews, etc.

Sesquipedalian:
Great! Thanks for your contribution, @MegaBrutal!

PR #8419, opened 2025-01-23
Improves display of popups with large amounts of content

Sesquipedalian:
- Better accommodation for large amounts of text.
- Fixes issue where excess content would be cut off when displayed on small device screens.


DiegoAndresCortes:
In #7933 it's already using max-height

PR #8421, opened 2025-01-24
[3.0] Implements URL routing (i.e. better queryless/pretty URLs) 

Sesquipedalian:
The main thing this does is to enable SMF native support for URLs to be rewritten as routes (i.e. virtual paths that are interpreted as well-structured queries). In other words, it implements a robust form of pretty URLs.

Examples:

| # | Standard URL | Routed URL |
| - | ------------ | ---------- |
| 1 | `https://example.com/index.php?board=1.0` | `https://example.com/boards/general-discussion-1` |
| 2 | `https://example.com/index.php?board=1.20` | `https://example.com/boards/general-discussion-1/20` |
| 3 | `https://example.com/index.php?topic=123.15` | `https://example.com/topics/my-great-topic-123/15` |
| 4 | `https://example.com/index.php?action=moderate;area=modlog` | `https://example.com/moderate/modlog` |
| 5 | `https://example.com/index.php?action=calendar;year=2025;month=1;day=23` | `https://example.com/calendar/2025/01/23/` |
| 6 | `https://example.com/index.php?action=post;board=1.0` | `https://example.com/boards/general-discussion-1/post` |
| 7 | `https://example.com/index.php?action=markasread;sa=board;board=1.0` | `https://example.com/boards/general-discussion-1/markasread/` |
| 8 | `https://example.com/index.php?action=profile;area=account;u=1` | `https://example.com/members/sesquipedalian-1/account` |

-----

There are four types of routes:
1. Routes to boards and topics, which use the form `/boards/<slug>-<id>` or `/topics/<slug>-<id>`, with an optional `/<start>` value appended for pagination purposes. See examples 1 through 3 above. There are also routes to individual posts using `/msgs/<id>`, but these redirect to the canonical URL just like `?msg=<id>` does.
2. Routes to typical actions, which use the form `/<action>/<area>/<sa>` (with the area and sub-action being optional). Example 4 above shows this. Some actions (see example 5) provide additional routing elements for commonly used query parameters that are specific to those particular actions.
3. Routes to actions that apply to a specific board or topic, which use the form `/boards/<slug>-<id>/<action>`. See examples 6 and 7 above.
4. Routes to members, which use the form `/members/<slug>-<id>/<area>/<sa>` (with the area and sub-action being optional). See example 8 above.

-----

Internally, routes are translated into query strings very early in the process (specifically, during `QueryString::cleanRequest()`), so that all other code works with the query string just like it always has. Similarly, forum URLs are rewritten as routes only near the very end of the process, during `Utils::obExit()`, `Utils::redirectexit()`, or `Mail::send()`.

-----

There are two settings that control this behaviour:
1. The existing `queryless_urls` setting. This is the primary setting. As has always been the case, it is only supported on the Apache, LiteSpeed, and lighttpd web servers.
2. A new `hide_index_php` setting. When this is enabled, `/index.php`will be removed from URLs pointing to pages within the forum. This setting is supported on any web server.

These two settings can operate independently. If queryless URLs are enabled but the option to hide index.php is disabled, then `https://example.com/index.php?board=1.0` will become `https://example.com/index.php/boards/general-discussion-1`. If the option to hide index.php is enabled but queryless URLs are disabled, then`https://example.com/index.php?board=1.0` will become `https://example.com/?board=1.0`.

If the admin enables both of these settings together, SMF writes a small mod_rewrite rule to the .htaccess file in order to make everything work correctly. This write operation performs safety checks; if the write operation fails, SMF will show an error message and will refuse to enable the settings.

-----

Backward compatibility for the old-school queryless URLs has been maintained. Those forms of queryless URLs will no longer be generated, but they will still be recognized and parsed. This ensures that existing links from external sites, browser bookmarks, etc., will continue to work.

-----

Slugs are generated on the fly for boards, topics, and members. Strictly speaking, these slugs are just fluff and don't matter for the functioning of the system. (Indeed, the slug can be left out or changed to any random string; all we need is the ID value that appears at the end). However, including a memorable slug in the URL allows the user to start typing the part of the URL that they remember into their browser's address bar and then have autocomplete suggest the remainder of the URL (including the harder-to-remember ID value) for them.

EDTIT: Based on @live627's feedback, we now redirect to the correct URL if an incorrect slug is given.

-----

@VBGAMER45 will likely be particularly interested in the new `integrate_rewrite_as_queryless` and `integrate_parse_route` hooks. With these hooks, you should be able to create a hooks-only version of your Pretty URLs mod in order to support people who have used your mod with SMF 2.1 and below and/or who just prefer the format of the URLs that your mod produces.


live627:
>  Backward compatibility for the old-school queryless URLs has been maintained. Those forms of queryless URLs will no longer be generated, but they will still be recognized and parsed. This ensures that existing links from external sites, browser bookmarks, etc., will continue to work.

Are these permanent redirects? I read somewhere that some browsers can silently update the bookmark if it targets a permanent redirect.

> Strictly speaking, these slugs are just fluff and don't matter for the functioning of the system. (Indeed, the slug can be left out or changed to any random string

Slugs cannot be arbitrary because users with a vendetta against the site will seed bad words into the address bar. I remember over a decade ago, a story broke where users would do this to news site and it was a big story because of f-bombs in the address bar that led to the article.

Don't let `example.com/member/nitwitt-1` point to `Admin`.

Sesquipedalian:
> Are these permanent redirects? I read somewhere that some browsers can silently update the bookmark if it targets a permanent redirect.

They're not redirects at all. We just parse them and display the content, just like we do now.

> > Strictly speaking, these slugs are just fluff and don't matter for the functioning of the system. (Indeed, the slug can be left out or changed to any random string
> 
> 
> 
> Slugs cannot be arbitrary because users with a vendetta against the site will seed bad words into the address bar. I remember over a decade ago, a story broke where users would do this to news site and it was a big story because of f-bombs in the address bar that led to the article.
> 
> 
> 
> Don't let `example.com/member/nitwitt-1` point to `Admin`.

Hm. I hadn't thought of that. I suppose I could add a check to compare the provided slug to the canonical slug, and redirect if they don't match.

dragomano:
Why are files like `.htaccess_20250124T031621Z` being created (the same thing happens when trying to add SMF rules to `robots.txt`) — shouldn’t they be deleted automatically?

Why not use a `textarea` for the option `Meta keywords associated with forum`? I’m sure users will find a text field with `size=50` pretty limiting.

In the help section, it would be great to show a hint with the code that needs to be manually inserted into `.htaccess` if it can’t be done automatically — at least on Windows, I started having issues with that. For example, here’s how it looks in SimpleSEF:

sshot-5

What about the `space replacement` option?

sshot-6

Missing character transliteration: `/boards/новыи-раздел-3` => `/boards/noviy-razdel-3`. You could use ready-made libraries like behat/transliterator or symfony/string:

```php
$slug = \Behat\Transliterator\Transliterator::transliterate($string, $this->spaceChar);

// or

$slug = (string) (new \Symfony\Component\String\Slugger\AsciiSlugger())->slug($string, $this->spaceChar)->lower();
```

Sesquipedalian:
> Why are files like `.htaccess_20250124T031621Z` being created (the same thing happens when trying to add SMF rules to `robots.txt`) — shouldn’t they be deleted automatically?

Backups. That's intentional behaviour.

> Why not use a `textarea` for the option `Meta keywords associated with forum`? I’m sure users will find a text field with `size=50` pretty limiting.

Good idea. The current input field is the same as it has been since at least SMF 2.1, but for those who do use meta keywords (basically just Russian users, since Yandex is the only search engine that still pays any attention to them) a textarea would certainly be nicer to work with.

> In the help section, it would be great to show a hint with the code that needs to be manually inserted into `.htaccess` if it can’t be done automatically — at least on Windows, I started having issues with that. For example, here’s how it looks in SimpleSEF:
> 
> sshot-5

Another good idea.

> What about the `space replacement` option?
> 
> sshot-6

That seems like an option for the sake of an option to me. I could be persuaded otherwise, though.

> Missing character transliteration: `/boards/новыи-раздел-3` => `/boards/noviy-razdel-3`. You could use ready-made libraries like behat/transliterator or symfony/string:
> 
> ```
> $slug = \Behat\Transliterator\Transliterator::transliterate($string, $this->spaceChar);
> 
> // or
> 
> $slug = (string) (new \Symfony\Component\String\Slugger\AsciiSlugger())->slug($string, $this->spaceChar)->lower();
> ```

The current behaviour is intentional. There's no reason not to support non-ASCII characters in URLs in 2025. In fact, I was considering the idea of localizing the boilerplate parts of the route, too. For example, on a forum that uses Russian as its default language the route would be something like `/разделы/новыи-раздел-3`. (I don't speak or read Russian, so that might have been the wrong word or grammatically incorrect, but I'm sure you get the idea.) That would require adding more language strings, of course.

Sources/Utils.php
```diff
@@ -998,6 +1004,78 @@ public static function ucwords(string $string): string
...
+		);
+
+		// Convert back to composed form.
+		$string = self::normalize($string, 'c');
+
+		// Truncate to fit within the allowed length.
+		while (mb_strlen($string) > $max_length) {
+			$string = mb_substr($string, 0, str_contains($string, '-') ? mb_strrpos($string, '-') : $max_length);
+		}
+
```
dragomano: What if we add a special hook of this kind here?

```php
// Give mods a chance to change slug string.
IntegrationHook::call('integrate_make_slug', [&$string]);
```

Sesquipedalian: Good idea.

Sesquipedalian:
@live627, I believe the changes in 7c6fe9f will take care of your concerns about arbitrary slugs. ~If no slug at all was provided, we just let it ride, but~ if a URL with an incorrect slug was provided, we redirect to the canonical URL. Unfortunately, this intervention cannot happen at early stages in the process, because we can't validate the incoming slug until we have loaded the necessary data about the requested board, topic, member, or membergroup.

@dragomano, I have implemented a bunch of your excellent suggestions. Thanks.

(Edited to strike out information that is no longer correct after subsequent commits.)

Sesquipedalian:
> > Missing character transliteration: `/boards/новыи-раздел-3` => `/boards/noviy-razdel-3`. You could use ready-made libraries like behat/transliterator or symfony/string:
> > ```
> > $slug = \Behat\Transliterator\Transliterator::transliterate($string, $this->spaceChar);
> > 
> > // or
> > 
> > $slug = (string) (new \Symfony\Component\String\Slugger\AsciiSlugger())->slug($string, $this->spaceChar)->lower();
> > ```
> 
> The current behaviour is intentional. There's no reason not to support non-ASCII characters in URLs in 2025. In fact, I was considering the idea of localizing the boilerplate parts of the route, too. For example, on a forum that uses Russian as its default language the route would be something like `/разделы/новыи-раздел-3`. (I don't speak or read Russian, so that might have been the wrong word or grammatically incorrect, but I'm sure you get the idea.) That would require adding more language strings, of course.

To be clear, I would very much like feedback from @dragomano, @live627, @jdarwood007, @sbulen, @BrickOzp, @LexArma, @Kindred-999, @MissAllSunday, @Oldiesmann, @MissAllSunday, @Arantor, or anyone else who wants to comment on this matter. I'm open to being persuaded to transliterate all slugs to ASCII, if given good arguments for it. I'm also open to considering other ideas about this, if someone wants to suggest something else.

Sources/ActionRouter.php
```diff
@@ -0,0 +1,161 @@
...
+	 *
+	 * @param array &$params URL query parameters.
+	 * @return array The route path elements.
+	 */
+	protected static function buildActionRoute(array &$params): array
+	{
+		if (str_starts_with(self::class, 'SMF\\Actions')) {
+			if (!isset($params['action'])) {
+				foreach (Forum::$actions as $action => $info) {
+					if (str_starts_with($info[1], self::class)) {
```
live627: These are callables and can be arrays.

```suggestion
					if (is_string($info[1]) && str_starts_with($info[1], self::class)) {
```

Oldiesmann:
The only issue I can see is that browsers will paste the URLs with HTML encoding even if displayed normally. See for instance the main page of the Russian version of Wikipedia, where "Заглавная_страница"
becomes this when pasted:
```
%D0%97%D0%B0%D0%B3%D0%BB%D0%B0%D0%B2%D0%BD%D0%B0%D1%8F_%D1%81%D1%82%D1%80%D0%B0%D0%BD%D0%B8%D1%86%D0%B0
```
I'm not sure how that will matter for international users though.

Sources/ActionRouter.php
Sesquipedalian: Good catch!

dragomano:
Users prefer to transliterate URLs rather than use national characters in URLs for several reasons:

- Compatibility: Not all browsers and systems handle URLs with national characters correctly, which can lead to access issues.
- Convenience: Transliteration helps avoid problems with keyboard layout, especially if the user doesn't know how to switch to the desired language.
- Memorability: Addresses written in Latin characters can be easier to remember and type, especially for users who do not speak the language with national characters.
- International Access: Transliterated addresses may be more understandable for an international audience that may not know how to read or input characters from another alphabet.
- Historical Habits: Many users are accustomed to using Latin characters on the internet, and this has become a standard for many web resources.

Yes, there are [“.рф” domains](https://en.wikipedia.org/wiki/.%D1%80%D1%84), those addresses are completely in Cyrillic. But many still hate such URLs and prefer to use solutions like the Pretty URLs mod.

Here’s an example with Japanese characters. Do you want such URLs on forums?

`https://www.simplemachines.org/あ-う-ん-な-い-そ-ん-に-ち-は-ひ-が-て-ら-し-い`

Sesquipedalian:
> The only issue I can see is that browsers will paste the URLs with HTML encoding even if displayed normally. See for instance the main page of the Russian version of Wikipedia, where "Заглавная_страница" becomes this when pasted:
> 
> ```
> %D0%97%D0%B0%D0%B3%D0%BB%D0%B0%D0%B2%D0%BD%D0%B0%D1%8F_%D1%81%D1%82%D1%80%D0%B0%D0%BD%D0%B8%D1%86%D0%B0
> ```
> 
> I'm not sure how that will matter for international users though.

That seems to depends on the browser. In Safari, for example, copying the URL directly from address bar copies the Unicode characters, whereas in Firefox the percent-encoded values are copied.

Based on Wikipedia's example, though, I'm now wondering whether I should actually be less aggressive in stripping out diacritical marks (accents and such). They leave the all the diacritics in their URLs.

LexArma:
I do like the idea of optional built in pretty-urls functionality. Not sure if localization is worth the effort though, but I'll keep an eye on this discussion. I'm not for or against.

Sources/Utils.php
```diff
@@ -998,6 +1004,81 @@ public static function ucwords(string $string): string
...
+				// Replace all characters that aren't letters or numbers with hyphens.
+				// Note: ideographic characters count as letters.
+				'/[^\p{L}\p{N}]/u' => fn($matches) => '-',
+
+				// Collapse runs of hyphens.
+				'/-+/u' => fn($matches) => '-',
+
+				// Trim leading and trailing hyphens.
+				'/^-|-$/u' => fn($matches) => '',
+			],
```
live627: This can be `preg_replace;() ` as it accepts arrays. The Unicode flag is not required for the last two patterns.

Sesquipedalian: Both true

dragomano:
When entering the topic, an error occurs:

sshot-8

live627:
I use `cocur/slugify` to generate slugs from titles.

dragomano:
> I use `cocur/slugify` to generate slugs from titles.

Yes, it's a good library too.

jdarwood007:
How is the url consumed into the app? Looks like we just dump it into the query string and then parse it back out later?

Just checking as while we can write .htaccess, I want to understand how we would support other web servers like nginx.  If we just dump the request URI into the query string, that can be done easily for other web servers.  Just not a oob since those typically require editing a server config, not a local .htaccess

Sesquipedalian:
> How is the url consumed into the app? Looks like we just dump it into the query string and then parse it back out later?

Yes, that's exactly what it does. So extending support to other web servers is certainly possible, as you say.

live627:
I click a link to view a topic and I get error

```
<br />
<b>Fatal error</b>:  Uncaught Error: Typed static property SMF\User::$me must not be accessed before initialization in C:\wamp64\www\smf-dev\Sources\Topic.php:1632
Stack trace:
#0 C:\wamp64\www\smf-dev\Sources\Topic.php(602): SMF\Topic-&gt;loadTopicInfo()
#1 C:\wamp64\www\smf-dev\Sources\Topic.php(1492): SMF\Topic::load(606)
#2 [internal function]: SMF\Topic::buildRoute(Array)
#3 C:\wamp64\www\smf-dev\Sources\QueryString.php(609): call_user_func('SMF\\Topic::buil...', Array)
#4 C:\wamp64\www\smf-dev\Sources\QueryString.php(536): SMF\QueryString::buildRoute(Array)
#5 [internal function]: SMF\QueryString::SMF\{closure}(Array)
#6 C:\wamp64\www\smf-dev\Sources\QueryString.php(519): preg_replace_callback('~(?:(?:(?&gt;\\b(?:...', Object(Closure), 'http://localhos...')
#7 C:\wamp64\www\smf-dev\Sources\Utils.php(2369): SMF\QueryString::rewriteAsQueryless('http://localhos...')
#8 C:\wamp64\www\smf-dev\Sources\QueryString.php(798): SMF\Utils::redirectexit('http://localhos...')
#9 C:\wamp64\www\smf-dev\Sources\QueryString.php(223): SMF\QueryString::redirectFromMsg()
#10 C:\wamp64\www\smf-dev\Sources\Forum.php(381): SMF\QueryString::cleanRequest()
#11 C:\wamp64\www\smf-dev\index.php(153): SMF\Forum-&gt;__construct()
#12 {main}
  thrown in <b>C:\wamp64\www\smf-dev\Sources\Topic.php</b> on line <b>1632</b><br />
```

EDIT: This is redirected from the old format.

Sesquipedalian:
Yeah, I believe that both @dragomano's error and @live627's error were introduced by the redirection code I added for dealing with non-canonical slugs. I'll try to fix it in the next couple of days.

Sesquipedalian:
I believe I have fixed those two bugs now.

I am still looking into options for ASCII slugs. I am marking this as a draft until that part has been dealt with.

dragomano:
The fact that users can link to topics using any phrases results in the canonical address matching those phrases, leading to different links to the same topic appearing in search results:

```php
/topics/some-title-28/
/topics/other-title-28/
/topics/title-28/
/topics/28/
```

Am I right?

Sesquipedalian:
No. If the slug is incorrect, the redirection logic will always force a redirect to the correct canonical URL.

tyrsson:
I'm just going to point out the obvious. Do with it what you will.

Actions should not be aware of or take part in their own routing. I can not imagine a way to more tightly couple it than that.

Sesquipedalian:
Having them tightly coupled was an intentional choice. My first draft had a separate route map, but I quickly realized that it would be extremely easy for that to get out of sync with the expected parameters of the actual actions. Instead, my thinking was that each action knows best what its own expected parameters are, and therefore what its route should look like. That's why QueryString::buildRoute() and QueryString::parseRoute() ask the appropriate action how its route should be built/parsed.

But if you have a different way of thinking about it, please do explain. I'm entirely open to revamping this system.

tyrsson:
Will post back my thoughts on it. I am currently working on something related, for SMF.

tyrsson:
Yes, I am aware there is a lot this does not address. I'm only addressing what is below to see how it differed from your initial draft. If its very similar then I'm not sure how much good any of my suggestions will be.

> My first draft had a separate route map, but I quickly realized that it would be extremely easy for that to get out of sync with the expected parameters of the actual actions.

How so? The path or route would define the parameters. Changes to the params would require an inherent change to the consumer would it not? I'm curious if your route map looked anything like what's below?

What the route definitions themselves could look like. Its pretty common among nearly every router I have ever used.

```php
return [
	[
		'path'       => '/',
		'action'     => Actions\BoardIndex::class,
		'name'       => 'board.index',
		'allowed_methods' => ['GET'],
	],
	[
		'path'       => '/board/{slug:[a-zA-Z0-9_-]+}[/{start:[0-9]+}]',
		'action'     => Actions\MessageIndex::class,
		'name'       => 'board',
		'allowed_methods' => ['GET'],
	],
	[
		'path'       => '/topic/{slug:[a-zA-Z0-9_-]+}[/{msg:[a-zA-Z0-9_-]+}]',
		'action'     => Actions\Display::class,
		'name'       => 'display',
		'allowed_methods' => ['GET'],
	],
];
```

For /board slug is not optional and will accept any combination of letters and numbers an underscore or a dash. Start is optional and accepts only digits 0-9 as I am sure you know.

If slug is not present in the uri then a match should not occur and a 404 would be generated.

For all of the example route definitions above, they would only match for GET request.

The action key is what ties the route to whatever is supposed to run when the route is matched.

Sesquipedalian:
Yes, that's similar to what I started with. If routes were only used for boards, topics, members, and membergroups, it would have been fine. 

The problem came with actions. Different actions can have so many possible variations of parameters and thus of the corresponding routes. Because of that, it quickly became clear that hardcoding all the possible routes into an enormous array would have been very brittle. Next I fiddled with making a multidimensional array to serve as a tree that the system could walk down in order to find a matching path. But that still required hardcoding, so it didn't help much. I needed something that could build and parse the routes programmatically by asking the relevant action what its route should look like.

I next tried building small hardcoded route maps in each action's properties, with the idea that a Router class would consume those. But that was really just repeating the hardcoded approach from before, except that now I was scattering the branches of the tree across many files. It didn't address the fundamental problem.

Thus I came to the conclusion that I needed methods to programmatically and dynamically determine the pattern(s) of the possible routes for each action. At that point, the most logical place to put those methods was in the action classes themselves. Now the top level code that handles routing (i.e. `QueryString::buildRoute()` and `QueryString::parseRoute()`) consults the relevant action rather than trying to keep track of everything itself.

One of the nice advantages of this dynamic approach is that if, for example, an action gains or loses a sub-action, whether from a mod or from future changes to SMF itself, the routing will automatically know that.

Still, perhaps there is a better way to accomplish these goals. If you have ideas for a superior way to accomplish the fundamental task of building and parsing routes dynamically rather than by hardcoding them, I'd be glad to hear them. 🙂

live627:
Mods that add pages might want to use the slug generator. How will they be able to specify their slug type? Also, should slugs be persisted somewhere as they will rarely change?

tyrsson:
I didnt have time to read it all because I am running late for an appt this morning.

A few quick thoughts.

Every current action is known. They should all have their own route.

Mods, can add an custom action by adding a route.

But, with the work that has already been put in and considering its working I think I may another solution. That will.

* Not introduce changes to the actions themselves beyond a const or two.
* Will not add methods to every action
* Will centralize the usage in SMF and keep it decoupled.

Give me a few hours and I will post an example.

Sesquipedalian:
> Mods that add pages might want to use the slug generator. How will they be able to specify their slug type? 

That's a good question. I will adjust SMF\Slug to support arbitrary types.

> Also, should slugs be persisted somewhere as they will rarely change?

I don't think so. They can be built programmatically using the existing data, so why bother storing a bunch of extra strings?

We could cache them, though, I suppose.

Sesquipedalian:
> Give me a few hours and I will post an example.

I look forward to it.

Sesquipedalian:
Oops. Pushed too soon.

Sesquipedalian:
FYI, I'm still looking into options for ASCII slugs. 

- The cocur/slugify library is nice, but it doesn't cover all the languages we need. 

- The Behat\Transliterator library would be excellent except that it is licensed under the LGPL, which is not two-way compatible with the BSD-3 license that SMF uses. (LGPL projects can use BSD-3 packages, but not vice-versa.)

- Symphony's AsciiSlugger depends on the intl extension's \Transliterator class, but we don't include intl as part of our system requirements (and for good reason). Symphony might provide a fallback library for intl (I haven't checked), but if so it will almost certainly be *enormous*.

jdarwood007:
If it is complicated or slower logic, then I would say yes, we could cache it. We could even cache it as a generated file. I had some rough ideas about letting SMF have a separate folder in our sources that holds generated PHP files (so we don't need to know or have all files writable).  A generated file we include would be faster than a cache include and parse.  This file wouldn't change often and we could let it generate if its missing or when a package manager action takes place.

Sesquipedalian:
> This file wouldn't change often

It would change every time a topic is created, actually.

jdarwood007:
Well if you're keeping the entire parsed output.  But not if we are saving the regex patterns and which action they route to.

Sesquipedalian:
> Well if you're keeping the entire parsed output. But not if we are saving the regex patterns and which action they route to.

Wait, are you talking about caching routes or caching slugs? I assumed you were talking about caching slugs, since that is what we were talking about caching until now. But if you want to cache regular expression patterns, that would make more sense if you meant caching routes.

jdarwood007:
Yes caching routes then.  Caching the slugs doesn't make sense, since once we identify what the route is, we can let the action handle determining how to handle things

tyrsson:
> > Give me a few hours and I will post an example.
> 
> I look forward to it.

I've approached it from several different directions and with the current implementation to work with I can't see a better path forward in respect to the traits etc. I would make the suggestion that ActionInterface can extend Routable and ActionTrait can use ActionRouter. Which would leave only adding traits to those that need the suffix router trait and provide the narrowest scope of changes.

My initial thought was to simply decorate actions at the call line. Where the decorator would take an ActionInterface instance. In the context of SMF this does not work very well. I thought I could get around it, maybe, but testing did not prove it out and I have no more time to spend on it.

dragomano:
What if users want to override some rules? cocur/slugify allows [extending existing rules](https://github.com/dragomano/SimpleSEF/blob/master/src/Sources/SimpleSEF-Db/CustomRuleProvider.php) by adding your own. In the case of your solution, will it require creating additional hooks for that?

Sesquipedalian:
@jdarwood007 

> Yes caching routes then. Caching the slugs doesn't make sense, since once we identify what the route is, we can let the action handle determining how to handle things

There's really nothing to cache in that regard. We don't use regular expressions to parse the route. We just explode it and then walk through the components of the path in order to translate them directly into URL query parameters.

Sesquipedalian:
@tyrsson 

> I would make the suggestion that ActionInterface can extend Routable and ActionTrait can use ActionRouter.

That's what I thought at first, too, but not everything that implements ActionInterface can or should implement Routable.

Still, thanks for taking a stab at it.

Sesquipedalian:
@dragomano 

> What if users want to override some rules? cocur/slugify allows [extending existing rules](https://github.com/dragomano/SimpleSEF/blob/master/src/Sources/SimpleSEF-Db/CustomRuleProvider.php) by adding your own. In the case of your solution, will it require creating additional hooks for that?

I've now updated AsciiTransliterator with a couple of hooks to allow that. 

However, two very different approaches are needed depending on whether the \Transliterator class is being used or not:

1. When \Transliterator is not being used, mods can directly apply their own rules on a character-by-character basis using the `integrate_ascii_transliterator_chars` hook.

2. When \Transliterator is being used, mods can use the `integrate_ascii_transliterator_id` hook to affect the results by changing the transliterator identifiers that are used to construct the Transliterator instance. 

    - It is not possible to directly modify rules on a character-by-character basis when using \Transliterator, because  \Transliterator does not operate on a simple character-by-character basis. It uses much more intelligent logic and thus produces much better output.

    - I expect that there will rarely be a need to adjust anything when using \Transliterator anyway, since the rules for transliterating the forum's default language are always applied before the generic transliteration rules. For example, if Turkish is the forum's default language, the specific rules for transliterating Turkish will be applied first, and then the generic rules for transliterating other languages will be applied only after that. This should result in optimal transliteration for the forum's primary language without any extra work.

Sesquipedalian:
In case anyone cares, the data for AsciiTransliterator comes from the same source that Behat\Transliterator got its data from. The results produced by AsciiTransliterator are thus identical to the results produced by Behat\Transliterator.

Sources/Msg.php
```diff
@@ -1719,6 +1714,14 @@ public static function modify(array &$msgOptions, array &$topicOptions, array &$
 			$searchAPI->postRemoved((int) $msgOptions['id']);
 		}
 
+		// If this is the first post of a topic, remove any cached slug string for the topic.
+		if (
+			!empty($topicOptions['id'])
+			&& ($msgOptions['id'] ?? NAN) === ($topicOptions['first_msg'] ?? NAN)
+		) {
+			CacheApi::put('slug_type-topic_id-' . $topicOptions['id'], null, 0);
```
dragomano, line 1723: You forgot to add `use SMF\Cache\CacheApi;`

Sesquipedalian: Oops! Fixed now.

Sesquipedalian:
I've tidied up the commit history for this PR.

dragomano:
I still haven't figured out how to add a custom action with my parameters to this routing system. I need a practical example with new hooks. I tried using urls replacement in the XML sitemap with `SMF\QueryString::rewriteAsQueryless`, but in the result, the sitemap stopped being generated altogether.

It would be interesting to try implementing route creation using class attributes:

```php
#[Route('/topics/{topic}', 'topic', ['topic' => '\d+', 'start' => '\d+'])]
class Topic
{
  //
}

#[Attribute(Attribute::TARGET_CLASS)]
class Route
{
  public function __construct(string $path, string $name, array $params = [])
  {
    //
  }

  public function build(array $params): array
	{
    //
  }

  public function parse(array $route, array $params = []): array
  {
    //
  }
}
```

Somewhere in code:

```php
$reflectionClass = new ReflectionClass(Topic::class);
$attributes = $reflectionClass->getAttributes(Route::class);

foreach ($attributes as $attribute) {
    $instance = $attribute->newInstance();
    $instance->someMethod(); // build, parse, etc
}
```

dragomano:
And it seems there is a bug if you try to click on the link to a specific post in the topic.

Sesquipedalian:
> I still haven't figured out how to add a custom action with my parameters to this routing system. I need a practical example with new hooks.

1. Your custom action must implement the SMF\Routable interface. 
    - If you just want basic routing in the form of `/action`, `/action/sa`, or `/action/area/sa`, then you just need to use the SMF\ActionRouter trait.
   - If you want to customize the route for your action, then:
       - Add a `buildRoute()` method to your action. Inside that method you can do whatever is appropriate to translate $_GET parameters into a route path. 
       - Add a `parseRoute()` method to your action. Inside that method you can do whatever is appropriate to translate the route path back into $_GET parameters.
 
2. Use the integrate_parse_route hook to add an entry to SMF\QueryString::$route_parsers, like this:
    ```
    QueryString::$route_parsers['custom_action'] = Actions\CustomAction::class;
    ```
    The key should be the string used as the value of `$_GET['action']`, and the value should be the fully qualified class name of your custom action's class (the example above uses the `::class` keyword to get the FQCN).

Sesquipedalian:
> And it seems there is a bug if you try to click on the link to a specific post in the topic.

Could you please elaborate? All the links to individual messages that I have tried work just fine.

dragomano:
> Could you please elaborate? All the links to individual messages that I have tried work just fine.

I get an error when I try to access a URL like /msgs/id.

sshot-12

If there is a poll in the topic, the error looks different

sshot-13

Sesquipedalian:
Thanks!

live627:
I did some digging by isolating the queries needed to build routes and have discovered that 71 queries were executed to open the message index https://dev.live627.com/index.php?board=6.0 of a populated board. The dump is attached [queries.txt](https://github.com/user-attachments/files/18653739/queries.txt) or at the bottom of the page and he code is at https://github.com/live627/SMF2.1/commit/8369dd6642d3b63ac912a8c2b6f12c2d89267976. Is there a way to combine similar queries?

Side note: Updating my custom form mod to use routes led to a blank page due to an error in my code. Turns out that errors thrown while running a buffer will cause this. (I run into this issue quite a bit when developing mods that operate  in the buffer)

live627:
If you want to move on and work on other code, we can merge this now and I can try to combine these calls in a future PR.

Sesquipedalian:
Looks like it's happening because it is loading all those members. I have an idea about how to fix that.

Sesquipedalian:
I believe the latest commit will fix that, @live627.

PR #8427, opened 2025-01-31
Moves code for handling redirects and unexpected request URLs out of Theme and into more appropriate locations

### LINKED ISSUE #8426: [3.0]: HTTP redirection loop occurs when user needs to agree to updated terms of service
### Basic Information

When the registration agreement or privacy policy have been updated, the code that redirects a member to ?action=agreement ends up going into an infinite redirection loop.

### Steps to reproduce

1. Update the registration agreement or privacy policy
2. Attempt to log in as an existing member who is not an admin.

### Expected result

The form that asks the member to accept the updated registration agreement or privacy policy will be shown.

### Actual result

Infinite redirection loop occurs.

### Version/Git revision

3.0 Alpha 2

### Database Engine

All

### Database Version

_No response_

### PHP Version

_No response_

### Logs

```shell

```

### Additional Information

Bug was introduced in #8395. 

It happens because Theme::load() is called in Forum::init(), before Forum::$current_action has been set by Forum::findAction(). This causes problems is because Theme::load() calls Theme::initialize(), which in turns calls Theme::requireAgreement(), which tries to check Forum::$current_action to decide whether to redirect or not.

The solution is to move the requireAgreement() out of Theme entirely. In fact, there are several methods called during Theme::initialize() that have absolutely no business being in Theme to begin with.

### LINKED ISSUE #8452: [3.0]: XML Parsing Error when trying to preview posts
### Basic Information

Clicking the Preview button either from the main post screen or the Quick Reply box results in an error. If you do it from the main post screen, it just sits there saying "Fetching preview...". If you do it from the Quick Reply box you get an XML Parse Error message in the browser (at least in Firefox)

### Steps to reproduce

1. Start a new post or a reply to an existing one
2. Click the "Preview" button

### Expected result

A preview of your post is displayed

### Actual result

XML Parsing Error if using Quick Reply (at least in Firefox) or "_Fetching preview..._" instead of the preview, with errors in console

### Version/Git revision

4e54200

### Database Engine

All

### Database Version

_No response_

### PHP Version

8.3.4

### Logs

```shell

```

### Additional Information

Firefox error: 
Image

Chrome error:
Image

Tested both locally and on a couple boards on my server. Same result.

Sesquipedalian:
Fixes #8426
Fixes #8452

[9662156](https://github.com/SimpleMachines/SMF/pull/8427/commits/9662156a8381f50f5beaeabb2a5af72834a6021e) will probably also help avoid issues like [this](https://www.simplemachines.org/community/index.php?topic=590802.0) perennial problem when admins inconsistently use 'www.' in some of their settings but not all.


live627:
The URL is no longer fixed when I request the page via `127.0.0.1` instead of `localhost`. Maybe we should ask @jdarwood007 to test aliases, too.

Sesquipedalian:
Derp. Fixed now.

Sesquipedalian:
Look good to you now, @live627?

live627:
How do I add a forum alias? A search for `alias` in the ACP yields nothing. Do I need to ad one manually in the database then?

Sesquipedalian:
> How do I add a forum alias? A search for `alias` in the ACP yields nothing. Do I need to ad one manually in the database then?

Looks like it, yeah. It appears to be a hidden setting.

Sesquipedalian:
I have added more commits to fix #8452, because the fixes for that issue need to build on the other changes in this PR.

Sesquipedalian:
@Oldiesmann, can you confirm that this fixes #8452?

Oldiesmann:
> @Oldiesmann, can you confirm that this fixes #8452?

PM previews are still broken but everything else works now. (Error from Firefox and Chrome)
image
image

Oldiesmann:
Previewing newsletters is also broken with the same error as above (just with a different URL)

live627:
I set up a new virtual host (wamp server makes this very easy) pointing to SMF but none of the links change. I add a new setting `forum_alias_urls` with the value `dev.local`. Did I do it right? Am I missing something?

Sesquipedalian:
@Oldiesmann, those should both be fixed now.

Sesquipedalian:
> I set up a new virtual host (wamp server makes this very easy) pointing to SMF but none of the links change. I add a new setting `forum_alias_urls` with the value `dev.local`. Did I do it right? Am I missing something?

Try now. I tweaked that code to be a bit smarter when I moved it, but in the process I forgot to consider the possibility that the values in that array might be just naked domain names or schemeless URLs.

Oldiesmann:
> @Oldiesmann, those should both be fixed now.

- I'll test when I get home tonight and let you know

live627:
Will test soon. This is the first time I set an alias.

live627 (approved):
it works now

PR #8430, opened 2025-01-31
Faster package backup creation

live627:
Now backups can be created in half a second whereas the old code took nearly two seconds in my tests. YMMV. Busy servers probably won't see the same processing times.

This feature must specifically be enabled in the package manager settings.

### DIFF
--- a/Sources/PackageManager/PackageUtils.php
+++ b/Sources/PackageManager/PackageUtils.php
@@ -3414,22 +3414,30 @@ class PackageUtils
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Assigned to Alpha 5 as per the internal roadmap.

Sesquipedalian:
Reassigned to Alpha 4 in accordance with changes to the internal roadmap.

jdarwood007:
@live627 Can you fix the merge conflict?

live627:
done

PR #8431, opened 2025-02-02
[3.0] Adds ability to anonymize accounts upon deletion

### LINKED ISSUE #7934: Anonymize posts upon account deletion
When a user account is deleted, add the ability to anonymize the posts made by that user.

Sesquipedalian:
This adds the ability to anonymize personally identifying information (PII) when an account gets deleted.

- Members can choose an option to have their PII anonymized when deleting their accounts. If account deletion requires admin approval, the anonymization will not occur until the deletion is approved.
- Admins and moderators with the ability to delete accounts can always choose to anonymize the PII of a member when deleting an account, even if the member did not request that.
- A new admin-level setting has been added that will always force PII to be anonymized when deleting an account. When this setting is enabled, no checkbox to choose anonymization is shown. Instead, we just show a message stating that anonymization will occur.

During anonymization, the following changes are made:

- An $anonymous_name value is generated. It will look similar to `user_4nx8fxcx`.
- An $anonymous_email value is generated. It will look similar to `4nx8fxcxvdam599vq5x5apnhv1@email.invalid`.
- The `poster_name` and `poster_email` fields for all of the member's posts are replaced with $anonymous_name and $anonymous_email.
- The member's name is replaced with $anonymous_name in the `modified_name` field of any posts that the member edited.
- Any mentions of the member within the body of forum posts is replaced with $anonymous_name.
- Any occurrences of the member's name in the following logs is replaced with $anonymous_name.
    - log_group_requests
    - log_packages
    - log_reported
    - log_reported_comments

Fixes #7934


jdarwood007 (commented):
I've reviewed the code and it looks fine.  Have not tested it yet.

PR #8432, opened 2025-02-02
Removes Config::$tasksdir setting entirely

### LINKED ISSUE #7942: Remove $tasksdir
#### Description
From discussion here #7939

We have had a few issues related to background tasks due to the naming conventions.  We renamed 'tasks' to 'Tasks' to fix these issues.  But more issues existed due to case sensitivity issues on typically Linux distros.

The proposal is to:
1. Move the Tasks directory back into Sources (if not already).
2. remove `$tasksdir` from Settings.php and Config.php
3. Update any reference to `$tasksdir` to use `$sourecedir . '/Tasks'`

Sesquipedalian:
Fixes #7942

Also moves some backward compatibility functions from Config to Subs-Compat, because there's no reason not to.

Also makes our handling of `use` statements in Subs-Compat consistent. Specifically, this gets rid of all `use` statements for classes in that file and therefore always gives the fully qualified class name in all references to classes. Previously we were rather haphazardly `use`-ing some classes but not others.


Sesquipedalian:
Since @jdarwood007 appears to be busy and this is a very simple change, I'm just going to go ahead and merge it now.

PR #8433, opened 2025-02-02
A couple of database fixes

Sesquipedalian:
Closes #8287


live627:
It's dead, Jim!!

```

Error: Invalid data structure sent to the database.(User.php-1426) in C:\wamp64\www\smf-dev\Sources\Db\APIs\MySQL.php on line 2406

Error: Invalid data structure sent to the database.(Session.php-121) in C:\wamp64\www\smf-dev\Sources\Db\APIs\MySQL.php on line 2406
```

Sesquipedalian:
Good. We can find those and change them so that they pass correctly structured data.

PR #8436, opened 2025-02-04
Always pass correctly structured data to Db::insert()

Sesquipedalian:
A followup for #8433


albertlast:
Is this done by hand or is that somekind of prettier/formatter?

Sesquipedalian:
By hand.

live627 (approved):

albertlast:
Would be Something like that: https://github.com/prettier/plugin-php not better?

live627:
we use php-cs-fixer

albertlast:
php-cs-fixer doesn't look like to do formatting?

Sesquipedalian:
It does. We would just need to write a custom rule if we wanted to do this, and I don't think it's worth the effort in this case.

PR #8443, opened 2025-02-05
Displays friendly URLs correctly in error log

### LINKED ISSUE #8440: [3.0]: Friendly URLs not displayed/logged properly in Error Log
### Basic Information

When the new "Use Friendly URLs" feature is enabled, they're not logged properly in the error log - instead it shows it as the query string (<url>/index.php?<friendly_url>)

### Steps to reproduce

1. Enable "Use friendly URLs" 
2. Do something that causes an error, then view the error log

### Expected result

Friendly URL displayed correctly

### Actual result

Invalid URL depending on settings:
Friendly URLs without hiding index.php: `<forum_url>/?/welcome-smf-1`
Friendly URLs and index.php hidden: friendly version of the URL treated as a query string (`<forum_url>/index.php?<forum_url>/welcome-smf-1`)

If you're just hiding index.php without friendly URLs, it shows the normal URL with index.php. That's not really an issue since that URL will still function regardless. The other two won't.

### Version/Git revision

7b217e4

### Database Engine

All

### Database Version

10.11.10-MariaDB-ubu2204

### PHP Version

8.4.3

### Logs

```shell

```

### Additional Information

_No response_

Sesquipedalian:
Fixes #8440

Previously we had attempted to be clever about figuring out what parts of the URL to record in the log and then about how to reconstruct the complete URL when displaying the log.

Now we keep it simple: 

- When recording, if the URL starts with Config::$boardurl, we chop off Config::$boardurl and save the remainder. If the URL does not start with Config::$boardurl for whatever reason, we record the whole URL. 

- When displaying, if the recorded value is a complete URL, we display it as-is. Otherwise, we prepend Config::$boardurl to it.

This does mean that people who have not enabled the option to hide index.php in their URLs will have `/index.php` at the start of all the recorded values in the `url` field of the error log table. But that's really not a big deal.

### DIFF
--- a/Sources/Actions/Admin/ErrorLog.php
+++ b/Sources/Actions/Admin/ErrorLog.php
@@ -27,6 +27,7 @@ use SMF\Parser;
 use SMF\SecurityToken;
 use SMF\Theme;
 use SMF\Time;
+use SMF\Url;
 use SMF\User;
 use SMF\Utils;
 
@@ -232,7 +233,7 @@ class ErrorLog implements ActionInterface
 				'time' => Time::create('@' . $row['log_time'])->format(),
 				'timestamp' => $row['log_time'],
 				'url' => [
-					'html' => Utils::htmlspecialchars(!str_contains($row['url'], 'cron.php') ? (str_starts_with($row['url'], '?') ? Config::$scripturl : '') . $row['url'] : $row['url']),
+					'html' => Utils::htmlspecialchars((Url::create($row['url'])->isValid() ? '' : Config::$boardurl) . $row['url']),
 					'href' => base64_encode(Db::$db->escape_wildcard_string($row['url'])),
 				],
 				'message' => [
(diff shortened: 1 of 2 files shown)


Sources/Actions/Admin/ErrorLog.php
```diff
@@ -232,7 +233,7 @@ public function view(): void
 				'time' => Time::create('@' . $row['log_time'])->format(),
 				'timestamp' => $row['log_time'],
 				'url' => [
-					'html' => Utils::htmlspecialchars(!str_contains($row['url'], 'cron.php') ? (str_starts_with($row['url'], '?') ? Config::$scripturl : '') . $row['url'] : $row['url']),
+					'html' => Utils::htmlspecialchars((Url::create($row['url'])->isValid() ? '' : Config::$boardurl) . $row['url']),
```
jdarwood007, line 236: So if the url is valid, we show nothing, otherwise we assume its a partial url and add in the boardurl?  Why not show it if its a valid url?

Sesquipedalian: No. If the URL is valid then we show it as-is. If it isn't then we prepend Config::$boardurl.

PR #8444, opened 2025-02-07
Fixes error with data structure in Display

Oldiesmann:
Fixes "Invalid data structure sent to the database" error when trying to view a post


Oldiesmann:
Apparently this was already fixed and my local copy wasn't up to date.

PR #8458, opened 2025-02-11
[2.1] Returns correct search results when post moderation is enabled

### LINKED ISSUE #7870: Search this topic does not deliver any results

Sesquipedalian:
Fixes #7870 for SMF 2.1.

Closes #8332

### DIFF
--- a/Sources/Search.php
+++ b/Sources/Search.php
@@ -1282,6 +1282,7 @@ function PlushSearch2()
 					'select' => array(
 						'id_search' => $_SESSION['search_cache']['id_search'],
 						'relevance' => '0',
+						'id_topic' => 't.id_topic',
 					),
 					'weights' => array(),
 					'from' => '{db_prefix}topics AS t',
@@ -1306,7 +1307,6 @@ function PlushSearch2()
 
 				if (empty($search_params['topic']) && empty($search_params['show_complete']))
 				{
-					$main_query['select']['id_topic'] = 't.id_topic';
 					$main_query['select']['id_msg'] = 'MAX(m.id_msg) AS id_msg';
 					$main_query['select']['num_matches'] = 'COUNT(*) AS num_matches';
 
@@ -1316,8 +1316,6 @@ function PlushSearch2()
 				}
 				else
 				{
-					// This is outrageous!
-					$main_query['select']['id_topic'] = 'm.id_msg AS id_topic';
 					$main_query['select']['id_msg'] = 'm.id_msg';
 					$main_query['select']['num_matches'] = '1 AS num_matches';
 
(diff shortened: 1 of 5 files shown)


Sesquipedalian:
@sbulen

sbulen:
Assuming everyone's on board with the db change!

The behavior matches current behavior.  Whether that's good or bad is up to you!

When searching within a topic, or when _showing results as messages_, you will see all messages that have the term.  However, under all other circumstances, you will only see one message per topic that includes the term.

Some folks like the current behavior - though I'm not sure how much of that is "that's how it's always been".  

We do get occasional complaints on the support board, though, that search results are incomplete.  So not everybody likes the current behavior.  

I had thought that changing this index would change this behavior, but it didn't...   This PR matches current behavior; the one-result-per-topic is enforced elsewhere.

To demonstrate...  In this test env, there are 3 topics, across two boards:
 - Topic: "Searched Topic" is within the board "Searched Board"
 - Topic: "Not Searched Topic" is within the board "Searched Board"
 - Topic: "Not Searched Board" is within the board "Not Searched Board"

Each topic has 3 posts, "Lions", "Lions and Tigers", and "Lions and Tigers and Bears".  

**_So there are 3 x 3 = 9 total posts include the term "Lions" across the 3 topics:_**
These are only visible if you check the box "Show results as messages" on the search form.  

across-topics-as-msgs

**_Searching across topics only gives you 3 results, one per topic (to the consternation of some):_**
If you do not check the box "Show results as messages" on the search form.  

across-topics

**_Searching within a topic gives you all 3 within that topic:_**

within-topics

sbulen:
Part of the challenge with the above is that "Show results as messages" really means something else...  It means "show all messages, not just one per topic, and show the entire message, don't truncate it"...

I wonder if "Show complete results" addresses the confusion (in English anyway)...

Sesquipedalian:
> Assuming everyone's on board with the db change!

It's just dropping and re-adding a primary key index. Nothing substantial.

> The behavior matches current behavior. Whether that's good or bad is up to you!

There's no plan to change the intended behaviour, at least not in 2.1. The goal is just to fix the bug.

In 3.0 we could talk about changing how search results are filtered. I agree that it would be better to always show all matching messages and relabel the "Show results as messages" search option as "Show full message text" or something like that. Perhaps we could even pretty it up a bit by grouping messages from the same topic together.

PR #8459, opened 2025-02-11
[3.0] Returns correct search results when post moderation is enabled

### LINKED ISSUE #7870: Search this topic does not deliver any results

Sesquipedalian:
Fixes #7870 for SMF 3.0.

Closes #8333.

### DIFF
--- a/Sources/Search/SearchApi.php
+++ b/Sources/Search/SearchApi.php
@@ -640,7 +640,7 @@ abstract class SearchApi implements SearchApiInterface
 		$request = Db::$db->search_query(
 			'',
 			'
-			SELECT ' . (empty($this->params['topic']) ? 'lsr.id_topic' : $this->params['topic'] . ' AS id_topic') . ', lsr.id_msg, lsr.relevance, lsr.num_matches
+			SELECT lsr.id_topic, lsr.id_msg, lsr.relevance, lsr.num_matches
 			FROM {db_prefix}log_search_results AS lsr' . ($this->params['sort'] == 'num_replies' || !empty($approve_query) ? '
 				INNER JOIN {db_prefix}topics AS t ON (t.id_topic = lsr.id_topic)' : '') . '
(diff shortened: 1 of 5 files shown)


Sesquipedalian:
@sbulen

sbulen:
Note there is a deprecation warning:
Deprecated
: str_ireplace(): Passing null to parameter 3 ($subject) of type array|string is deprecated in
D:\wamp64\www\84search30\upgrade.php
on line
959

Same behavior under PHP 8.3 & 8.4.

I don't know if that is the cause, but I cannot run a 3.0 => 3.0 upgrade to test this with my existing 3.0 test env/data.  I get stuck on this screen:
(And yes, I did try clicking on the "try again" link!)
image

sbulen:
Note that if I manually update the index, I get the same results as in #8458 , which match the current results.

**That may be good enough for now,** as we have at least two major upgrader.php PRs coming up, the mb4 one and the rewrite.  Not sure we want to invest a lot of time in this incarnation of the upgrader.  The logic appears to work.

But prior to those two PRs, looks like we can't do a 3.0 => 3.0 upgrade at the moment.  Not a major loss, but, it's a test result...

Sesquipedalian:
> Note that if I manually update the index, I get the same results as in #8458 , which match the current results.
> 
> **That may be good enough for now,** as we have at least two major upgrader.php PRs coming up, the mb4 one and the rewrite. Not sure we want to invest a lot of time in this incarnation of the upgrader. The logic appears to work.

Excellent.

We'll deal with upgrader stuff elsewhere.

PR #8472, opened 2025-02-13
[3.0] Fixes pagination issue

### LINKED ISSUE #8471: [3.0]: Paginator displays an extra page
### Basic Information

Image

### Steps to reproduce

1. Create 2-3 topics with 2-3 posts
2. Set "Number of topics per page in the message index": 1
3. Set "Number of posts per page in a topic page": 1

### Expected result

_No response_

### Actual result

_No response_

### Version/Git revision

dda4503

### Database Engine

All

### Database Version

11.2.2-MariaDB

### PHP Version

8.2.27

### Logs

```shell

```

### Additional Information

_No response_

Sesquipedalian:
Fixes #8471

Technically, only the first commit is necessary to fix the issue, but the second commit is also a good idea.

### DIFF
--- a/Sources/Actions/Admin/ErrorLog.php
+++ b/Sources/Actions/Admin/ErrorLog.php
@@ -188,6 +188,11 @@ class ErrorLog implements ActionInterface
 		$start = (int) $_GET['start'];
 		Utils::$context['page_index'] = new PageIndex(Config::$scripturl . '?action=admin;area=logs;sa=errorlog' . (Utils::$context['sort_direction'] == 'down' ? ';desc' : '') . (isset($this->filter) ? $this->filter['href'] : ''), $start, (int) $num_errors, (int) Config::$modSettings['defaultMaxListItems']);
 
+		// If the supplied start value was invalid, redirect to the correct one.
+		if ($_GET['start'] != $start) {
+			Utils::redirectexit(Utils::$context['page_index']->base_url . ';start=' . $start);
+		}
+
 		Utils::$context['start'] = $_GET['start'];
 
 		// Update the error count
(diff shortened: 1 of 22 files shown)


Sesquipedalian:
@dragomano, please confirm that this fixes the issue for you.

dragomano:
Yes, this fixes pagination in the list of topics within a board, but the bug occurs everywhere, not just there.

dragomano:
Although it looks like there is an error somewhere in the calculations of `$this->last_page_value` or `$this->last_page_num`.

This change helped:
```php
$this->last_page_value = (int) (($this->max_value - 1) / $this->num_per_page) * $this->num_per_page;
```
PageIndex.php, line 271

Sesquipedalian:
> Yes, this fixes pagination in the list of topics within a board, but the bug occurs everywhere, not just there.

Hm. Yes, you are correct.



> Although it looks like there is an error somewhere in the calculations of `$this->last_page_value` or `$this->last_page_num`.
> 
> This change helped:
> 
> ```
> $this->last_page_value = (int) (($this->max_value - 1) / $this->num_per_page) * $this->num_per_page;
> ```
> 
> PageIndex.php, line 271

I think `$this->last_page_value` is a red herring. The fundamental problem is actually that `$start` gets a bad value. I will investigate further.

Sesquipedalian:
> Yes, this fixes pagination in the list of topics within a board, but the bug occurs everywhere, not just there.

Hm. Yes, you are correct.



> Although it looks like there is an error somewhere in the calculations of `$this->last_page_value` or `$this->last_page_num`.
> 
> This change helped:
> 
> ```
> $this->last_page_value = (int) (($this->max_value - 1) / $this->num_per_page) * $this->num_per_page;
> ```
> 
> PageIndex.php, line 271

I think we will do better to simply adjust `$this->max_value`. The logic in the PageIndex class assumes that `$this->max_value` is a zero-based value, but everything that gets passed to it is always a count of items. So we need to subtract one from it to get the right value in our calculations. This affects not only `$this->last_page_value`, but also `$start` in PageIndex::fixStart()

Sesquipedalian:
The issue should now be well and truly fixed.

PR #8474, opened 2025-02-15
[3.0]: Add `integrate_build_route` hook

dragomano:
While testing the ability to add new routes, I discovered that there is no way to add routes for URLs like `/index.php?page=somepage` and similar ones. This hook provides that capability.

### DIFF
--- a/Sources/ActionTrait.php
+++ b/Sources/ActionTrait.php
@@ -27,7 +27,7 @@ trait ActionTrait
 	 * An instance of this class.
 	 * This is used by the load() method to prevent multiple instantiations.
 	 */
-	protected static self $obj;
+	protected static $obj;
 
 	/****************
 	 * Public methods
--- a/Sources/ActionRouter.php
+++ b/Sources/ActionRouter.php
@@ -59,13 +59,11 @@ trait ActionRouter
 	 */
 	public static function parseRoute(array $route, array $params = []): array
 	{
-		$params = array_merge($params, self::parseActionRoute($route));
-
-		return $params;
+		return array_merge($params, self::parseActionRoute($route));
 	}
 
 	/************************
-	 * Interal static methods
+	 * Internal static methods
 	 ************************/
 
 	/**
@@ -84,6 +82,8 @@ trait ActionRouter
 	 */
 	protected static function buildActionRoute(array &$params): array
 	{
+		$route = [];
+
 		if (str_starts_with(self::class, 'SMF\\Actions')) {
 			if (!isset($params['action'])) {
 				foreach (Forum::$actions as $action => $info) {
@@ -131,7 +131,7 @@ trait ActionRouter
 	 * For all other actions, the second element of the route path is mapped to
 	 * the 'sa' parameter.
 	 *
-	 * The route is passed by reference and route path elements are removed from
+	 * The route passed by reference and route path elements are removed from
 	 * the route when recognized. This lets wrapper methods continue parsing the
 	 * remainder of the route without worrying about duplicate elements.
 	 *
--- a/Sources/QueryString.php
+++ b/Sources/QueryString.php
@@ -617,6 +617,8 @@ class QueryString
 			$route_base = 'msgs';
 		}
 
+		IntegrationHook::call('integrate_build_route', [&$route_base, $params]);
+
 		if (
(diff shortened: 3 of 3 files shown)


Sesquipedalian:
Yes there is. When integrate_parse_route is called, which is very early in the process, you add your routable class to QueeyString::$route_parsers. Then when the time comes to build the route, which is very late in the process, your routable class is already in the list and will be called upon to build the route.

Sesquipedalian:
Perhaps I should change the name of the hook to make it clearer that it is for adding route parsers in general, not just for parsing routes specifically.

Sesquipedalian:
FYI, @dragomano, I reverted 20f64f3 and then added 65fef3f instead. You can simply use `integrate_route_parsers` to add a class that implements SMF\Routable to QueryString::$route_parsers. That will ensure that your class that implements SMF\Routable is available both for building routes and parsing routes when needed.

dragomano:
Something isn't working. In my version, I could use the hooks like this:

```php
	public function buildRoute(&$route_base, $params): void
	{
		if (isset($params['page'])) {
			$route_base = 'pages';
		}
	}

	public function parseRoute(): void
	{
		QueryString::$route_parsers['pages'] = Page::class;
	}
```

after which I would get the desired transformation: `/index.php?page=smithchester => /pages/smithchester`  

```php
class Page implements Routable
{
	use ResponseTrait;

	public static function buildRoute(array $params): array
	{
		$route = [];

		if (isset($params['page'])) {
			$route[] = 'pages';
			$route[] = $params['page'];

			unset($params['page']);
		}

		return ['route' => $route, 'params' => $params];
	}

	public static function parseRoute(array $route, array $params = []): array
	{
		array_shift($route);

		// We need to redirect from `/pages` to `/`
		if (empty($route)) {
			(new Response())->redirect();
		}

		$params['page'] = array_shift($route);

		return $params;
	}
}
```

Sources/ActionTrait.php
```diff
@@ -27,7 +27,7 @@ trait ActionTrait
 	 * An instance of this class.
 	 * This is used by the load() method to prevent multiple instantiations.
 	 */
-	protected static self $obj;
+	protected static $obj;
```
jdarwood007, line 30: Why do we need to do this?

Sesquipedalian:
@dragomano, what are you doing in the function that hooks into `integrate_route_parsers`?

It should look similar to this:

```
function add_route_parser(): void
{
	SMF\QueryString::$route_parsers['page'] = SMF\Page::class;
}
```

Sources/ActionTrait.php
dragomano: PHPStorm says `Property type does not match` because `static self` and `@var static` are not the same.

dragomano:
> @dragomano, what are you doing in the function that hooks into `integrate_route_parsers`?
> 
> It should look similar to this:
> 
> ```
> function add_route_parser(): void
> {
> 	SMF\QueryString::$route_parsers['page'] = SMF\Page::class;
> }
> ```

I need to use `QueryString::$route_parsers['pages'] = Page::class;`, but id doesn't help. And `QueryString::$route_parsers['page'] = Page::class;` doesn't work too. I'm interested in the following replacement:  `/index.php?page=slug` => `/pages/slug`

dragomano:
Once again, the following code is currently used in `QueryString.php`:  

```php
		if (isset($params['action'])) {
			$route_base = $params['action'];
		} elseif (isset($params['board'])) {
			$route_base = 'boards';
		} elseif (isset($params['topic'])) {
			$route_base = 'topics';
		} elseif (isset($params['msg'])) {
			$route_base = 'msgs';
		}
```
It works for actions, and boards, and topics, and msgs, but I need to add this case:

```php
		} elseif (isset($params['page'])) {
			$route_base = 'pages';
		}
```

Sources/ActionTrait.php
jdarwood007: Interesting, I would have said correct the @var as the documentation above says it should be an instance of the class.

Sesquipedalian:
I understand now. Yes, that needs to be addressed.

Sesquipedalian:
#8477 has reapplied commit 20f64f3.

PR #8475, opened 2025-02-15
[3.0] integrate_parse_route → integrate_route_parsers

Sesquipedalian:
Closes #8474

@dragomano, simply use this hook to add a class that implements SMF\Routable to QueryString::$route_parsers.


Sesquipedalian:
Actually, I will update #8474 with this change, rather than submitting an alternate PR, because I want the other commit you included in it.

PR #8484, opened 2025-02-21
Add formatText to ItemList

live627:
####     Introducing `formatText` to format values

Is somewhat similar to the `sprintf` formatter in that it has `format` and `params`, but there is no option for HTML encoding values and arguements must be provided as a hashmap/dictionary. It includes the rows from the `get_items` function.

### DIFF
--- a/Sources/ItemList.php
+++ b/Sources/ItemList.php
@@ -308,7 +308,7 @@ class ItemList implements \ArrayAccess
 		} else {
 			$this->sort = [
 				'id' => $this->options['default_sort_col'],
-				'desc' => (!empty($this->options['default_sort_dir']) && $this->options['default_sort_dir'] == 'desc') || (!empty($this->options['columns'][$this->options['default_sort_col']]['sort']['default']) && str_ends_with($this->options['columns'][$this->options['default_sort_col']]['sort']['default'], 'desc')) ? true : false,
+				'desc' => (!empty($this->options['default_sort_dir']) && $this->options['default_sort_dir'] == 'desc') || (!empty($this->options['columns'][$this->options['default_sort_col']]['sort']['default']) && str_ends_with($this->options['columns'][$this->options['default_sort_col']]['sort']['default'], 'desc')),
 			];
 		}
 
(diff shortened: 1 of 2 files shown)


Sesquipedalian:
Since Lang::formatText() internally handles both MessageFormat strings and sprintf strings, I suggest that you process both the sprintf type and the formatText type with the same code, rather than having two different blocks of code.

The following is untested, but it illustrates the idea:
```
				// Using formatText is probably the most readable way of injecting data.
				elseif (isset($column['data']['formatText']) || isset($column['data']['sprintf'])) {
					$params = [];

					if (isset($column['data']['sprintf'])) {
						$column['data']['formatText'] = $column['data']['sprintf'];
					}

					foreach ($column['data']['formatText']['params'] as $param => $htmlsafe) {
						$params[] = $htmlsafe ? Utils::htmlspecialchars((string) $list_item[$param]) : $list_item[$sprintf_param];
					}

					$cur_data['value'] = Lang::formatText($column['data']['formatText']['format'], $params);
				}
```

This would replace the existing code for handling sprintf format parameters. Then along with this change we could (1) rename `$column['data']['sprintf']` to `$column['data']['formatText']` in all the various `$listOptions` arrays scattered around the codebase and (2) declare `$column['data']['sprintf']` to be deprecated in favour of `$column['data']['formatText']`.

Sesquipedalian (changes requested):
Since Lang::formatText() internally handles both MessageFormat strings and sprintf strings, I suggest that you process both the sprintf type and the formatText type with the same code, rather than having two different blocks of code.

The following is untested, but it illustrates the idea:
```
// Using formatText is probably the most readable way of injecting data.
elseif (isset($column['data']['formatText']) || isset($column['data']['sprintf'])) {
	$params = [];

	if (isset($column['data']['sprintf'])) {
		$column['data']['formatText'] = $column['data']['sprintf'];
	}

	foreach ($column['data']['formatText']['params'] as $param => $htmlsafe) {
		$params[] = $htmlsafe ? Utils::htmlspecialchars((string) $list_item[$param]) : $list_item[$param];
	}

	$cur_data['value'] = Lang::formatText($column['data']['formatText']['format'], $params);
}
```

This would replace the existing code for handling sprintf format parameters. Then along with this change we could (1) rename `$column['data']['sprintf']` to `$column['data']['formatText']` in all the various `$listOptions` arrays scattered around the codebase and (2) declare `$column['data']['sprintf']` to be deprecated in favour of `$column['data']['formatText']`.

A further benefit is that there is no need to introduce another format with slightly different behaviour than the existing type. Regardless whether the format string uses MessageFormat or sprintf, the expected input structure and the resultant behaviour will be the same. For example, it will still be possible to choose for each individual parameter whether to use htmlspecialchars() on the value.

live627:
Took a bit to figure out a good way to combine processing those data types . Might be less  readable now; probbably doesn't matter. 

**Testing**

Verify  `sprintf` still works by visting the scheduled tasks page . Verify `getTxt` by visiting task log.

**Question**

Shall it      instead use underscores for consistency? `format_text`

Sesquipedalian:
> Shall it instead use underscores for consistency? `format_text`

Sure. Might as well change `getTxt` to `get_txt` while we're at it.

Sources/ItemList.php
```diff
@@ -402,82 +402,72 @@ protected function buildRows(): void
...
+					$call = 'vsprintf';
+					$format = 'sprintf';
+
+					if (isset($data['getTxt'])) {
+						$call = 'SMF\Lang::getTxt';
+						$format = 'getTxt';
+					} elseif (isset($data['formatText'])) {
+						$call = 'SMF\Lang::formatText';
+						$format = 'formatText';
 					}
```
Sesquipedalian:
```suggestion
					if (isset($data['getTxt'])) {
						$call = 'SMF\Lang::getTxt';
						$format = 'getTxt';
					} else {
						$call = 'SMF\Lang::formatText';
						$format = isset($data['formatText']) ? 'formatText' : 'sprintf';
					}
```

SMF\Lang::formatText() handles both MessageFormat and sprintf format, so it can be called for both format types.

Sesquipedalian (changes requested):

Sesquipedalian (approved):

PR #8492, opened 2025-03-01
[2.1] Fix mail queue issues 

### LINKED ISSUE #8314: [2.1]: Mail queue not emptying
### Basic Information

The mail queue was updated in 2.1.5 to address emails that can't be sent by adding a priority. I missed the code to have it delete emails after they reached the timeout (127).  This should be a release blocker and a simple fix.

### Steps to reproduce

1.
2.

### Expected result

_No response_

### Actual result

_No response_

### Version/Git revision

2.1.5 GIT

### Database Engine

All

### Database Version

_No response_

### PHP Version

_No response_

### Logs

_No response_

### Additional Information

_No response_

jdarwood007:
By randomizing mail out of the queue for better processing.

@ Sesquipedalian, this is what I have working, and it keeps the queue fairly low and maintainable.

The random pull from the queue does it. So, on a chance, we will pull older items from the queue and try them, then push their priority up using the logic that assumes we want to hold them for about three days before giving up. The next send time helps prevent us from dosing a mail server if the queue is low enough to try to send the same email every time this function is called, which could be every page load.

3.0 needs 2 column changes, next_send and tries.  Both are fairly descriptive of what I intend for them to do, which is the same thing here, but without working with the priority column to avoid database changes.

Fixes #8314

### DIFF
--- a/Sources/ScheduledTasks.php
+++ b/Sources/ScheduledTasks.php
@@ -717,11 +717,40 @@ function ReduceMailQueue($number = false, $override_limit = false, $force_send =
 		ORDER BY priority ASC, id_mail ASC
 		LIMIT {int:limit}',
 		array(
-			'limit' => $number,
+			'limit' => ceil($number * 0.7),
 		)
 	);
 	$ids = array();
 	$emails = array();
+	while ($row = $smcFunc['db_fetch_assoc']($request))
+	{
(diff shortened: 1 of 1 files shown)


Sesquipedalian (approved):

isaak654:
> Fixes #8314

For some reason, the issue has not yet been closed.

PR #8509, opened 2025-03-15
[3.0] Lock all actions to a specific version

jdarwood007:
In light of recent compromise of a github action, I am proposing we change all actions to reference the SHA-1 immutable id rather than a tag which can be overridden.


Sesquipedalian:
Makes sense to me

PR #8511, opened 2025-03-16
Revert "Lock all actions to a specific version"

Sesquipedalian:
This reverts commit 642c591c8921030561580b8f6c86eeaf8852c5f9.

The changes in that commit were causing the action to fail. A different approach will be needed. 
https://github.com/SimpleMachines/SMF/actions/workflows/php.yml

@jdarwood007


Sesquipedalian:
I'm going to go ahead and merge this reversion now, @jdarwood007. If and when you are able to take a another crack at it, please do so.

PR #8512, opened 2025-03-16
Revert "[2.1] Lock all actions to a specific version"

Sesquipedalian:
This reverts commit 8aa0c6f71e177df9f32f23942fcb3a08ff413d1d.

The changes in that commit were causing the action to fail. A different approach will be needed. 
https://github.com/SimpleMachines/SMF/actions/workflows/php.yml

@jdarwood007


Sesquipedalian:
I'm going to go ahead and merge this reversion now, @jdarwood007. If and when you are able to take a another crack at it, please do so.

PR #8519, opened 2025-03-19
[3.0]: Add support of placeholder attribute

dragomano:
A small addition for the settings.

Does it make sense to create a similar PR for the 2.1 branch?

### DIFF
--- a/Themes/default/Admin.template.php
+++ b/Themes/default/Admin.template.php
@@ -899,6 +899,7 @@ function template_show_settings()
 				$javascript = $config_var['javascript'];
 				$disabled = !empty($config_var['disabled']) ? ' disabled' : '';
 				$subtext = !empty($config_var['subtext']) ? '<br><span class="smalltext"> ' . $config_var['subtext'] . '</span>' : '';
+				$placeholder = !empty($config_var['placeholder']) ? ' placeholder="' . $config_var['placeholder'] . '"' : '';
 
 				// Various HTML5 input types that are basically enhanced textboxes
 				$text_types = array('color', 'date', 'datetime', 'datetime-local', 'email', 'month', 'time');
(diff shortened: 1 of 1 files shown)


Oldiesmann:
I don't think it makes sense to add this to 2.1 really. At this point most mods compatible with 2.1 final or newer will work fine with any version of 2.1. If we added this to 2.1, mod authors wouldn't be able to take advantage of it as easily since they'd either have to release a new version that's only "compatible" with versions that have this feature or they'd have to do version checking somewhere in the code to only output a placeholder if that version of SMF supports it. Many probably wouldn't see it as being worth the trouble.

It's fine in 3.0 since it's early in the development stages and mod authors aren't concerned with 3.0 at this point.

dragomano:
As far as I'm concerned, if a feature isn't supported in Admin.template.php, it's just ignored, so it's not a big deal. Right now, when I need to add a placeholder for a field, I have to use some hacky code like `'size' => '100" placeholder="Some text'`, even though the size isn't always necessary.

DiegoAndresCortes (approved):

DiegoAndresCortes:
> If we added this to 2.1, mod authors wouldn't be able to take advantage of it as easily since they'd either have to release a new version that's only "compatible" with versions that have this feature or they'd have to do version checking somewhere in the code to only output a placeholder if that version of SMF supports it. Many probably wouldn't see it as being worth the trouble.

I'm under the impression that it's just an array key so it would be ignored if it doesn't exist?
Anyways, I don't think it's necessary for 2.1, we can probably leave on 3.0 only.

Sesquipedalian:
Added to the roadmap for Alpha 4, which is the one for theme changes.

Sesquipedalian (approved):

PR #8526, opened 2025-03-24
[3.0] Improvements to language string handling

### LINKED ISSUE #8531: [3.0]: Scheduled Tasks Enabled flags not displaying correctly
### Basic Information

If you visit the 3.0 Scheduled Tasks page, it appears no tasks are enabled.

If you attempt to enable one of the tasks, it remains unselected.

I think this is a display issue, not a setting issue, as it appears some of them are running automatically.

### Steps to reproduce

1. Install 3.0
2. Enable some Scheduled Tasks

### Expected result

Enabled flag should reflect setting

### Actual result

Enabled flag is never displayed

### Version/Git revision

3.0 Alpha 3 - Current GitHub

### Database Engine

All

### Database Version

8.4.0

### PHP Version

8.4.2

### Logs

```shell

```

### Additional Information

_No response_

Sesquipedalian:
SMF\Lang::getTxt() was introduced in #8064 in order to give us an abstracted method to retrieve and format the localizable strings that we present in the UI. However, until now we have continued to access and use the raw variables in many, many places rather than working through the methods provided by SMF\Lang. This PR fixes that.

In addition to using SMF\Lang::getTxt() in pretty much all the places where we were previously using the variables directly, this PR also expands the functionality of SMF\Lang in the following ways:

1. Allows SMF\Lang::getTxt() to return an array when asked for one. Previously, it would throw a ValueError if asked to do that. For example, `SMF\Lang::getTxt('days')` will now return the array of strings stored in `SMF\Lang::$txt['days']`. Previously, it was only possible to retrieve individual elements of the array, like so: `SMF\Lang::getTxt(['days', 2])`.

2. Adds optional $file and $lang parameters to SMF\Lang::getTxt(). These parameters allow use to indicate exactly which file and language we want to retrieve the value from. Loading the correct file is handled automatically. There is no longer any uncertainly about which value will be retrieved, and there is no longer any need to worry about juggling back and forth between language files when doing complicated things like sending out notifications in multiple languages.

3. Implements SMF\Lang::txtExists(). As one might guess from the method name, this checks whether the specified string exists in our language files. When given a `$file` (and optional `$lang`) argument, it will load the specified file before checking whether the string exists.

4. Implements SMF\Lang::setTxt(). This allows us to dynamically add or overwrite any arbitrary language string. Using this method is generally safer and better than manipulating the variables directly. It is primarily used for working with dynamically generated strings.

5. Because of the changes mentioned above, it is now almost never necessary to call SMF\Lang::load() directly. Indeed, I strongly recommend that henceforth we all avoid calling SMF\Lang::load() unless absolutely necessary. We'll get better, more efficient, and more reliable results by sticking to using the file parameter of SMF\Lang::getTxt().

The PR also includes various fixes for some minor bugs that I discovered along the way.

Also fixes #8531.


jdarwood007:
What about overloading in `Modifications.{lang}.php`?

Sesquipedalian:
> What about overloading in `Modifications.{lang}.php`?

There's code to handle that in Lang::getTxt():

```php
		// Check whether we need to load the specified file.
		if (
			is_string($file)
			&& (
				// If we haven't loaded the file yet, do so now.
				!isset(self::$loaded_keys[$txt_key[0]])

				// If we loaded it for a different language, reload for the right language.
				|| self::$loaded_keys[$txt_key[0]]['lang'] !== $lang

				// In the event of key conflicts between different files, give
				// them the string from the requested file. HOWEVER, if the key
				// was overwritten in the Modifications or ThemeStrings language
				// files, then keep that version instead.
				|| (
					self::$loaded_keys[$txt_key[0]]['file'] !== $file
					&& self::$loaded_keys[$txt_key[0]]['file'] !== 'ThemeStrings'
					&& self::$loaded_keys[$txt_key[0]]['file'] !== 'Modifications'
				)
			)
		) {
			self::load($file . (!str_contains($file, 'ThemeStrings') ? '+ThemeStrings' : '') . (!str_contains($file, 'Modifications') ? '+Modifications' : ''), $lang, force_reload: true);
		}

```

ThemeStrings will always take precedence over a typical language file, and Modifications will take precedence over ThemeStrings.

Sesquipedalian:
Does that seem reasonable to you, @jdarwood007?

Sesquipedalian:
I've added a fix to this PR for #8531. I've included that fix here merely in order to avoid creating merge conflicts later.

PR #8529, opened 2025-03-25
[3.0] Informs reader when a post has bumped an old topic

Sesquipedalian:
This one comes from a list of good ideas shared with me by @Arantor a while ago. Notice the nice little "_6 months later..._" message:

Screenshot 2025-03-24 at 8 02 23 PM

This message will be shown between posts only when the time elapsed between their post dates is greater than the "Time before topic is warned as old on reply" setting. 

The message doesn't state the exact amount of time that has passed. Instead, it rounds the value down to the nearest reasonable unit of time: 
- If more than a year has passed, the message says "X years later..."
- If more than a month has passed, the message says "X months later..."
- If more than 14 days have passed, the message says "X weeks later..."
- Otherwise, the message says "X days later..."

If the user has enabled the option to show the most recent posts first (i.e. reverse chronological order), then the message changes from "X years later..." to "X years earlier..."

### DIFF
--- a/Languages/en_US/General.php
+++ b/Languages/en_US/General.php
@@ -984,6 +984,45 @@ $txt['error_old_topic'] = 'Warning: {0, plural,
 	other {this topic has not been posted in for at least # days.}
 }<br>Unless you are sure you want to reply, please consider starting a new topic.';
 
+$txt['bump_notice'] = '{invert, select,
+	true {{unit, select,
+		year {{num, plural,
+			one {# year earlier...}
+			other {# years earlier...}
+		}}
+		month {{num, plural,
+			one {# month earlier...}
+			other {# months earlier...}
+		}}
+		week {{num, plural,
+			one {# week earlier...}
+			other {# weeks earlier...}
+		}}
+		other {{num, plural,
+			one {# day earlier...}
+			other {# days earlier...}
+		}}
+	}}
+	other {{unit, select,
+		year {{num, plural,
+			one {# year later...}
+			other {# years later...}
+		}}
+		month {{num, plural,
+			one {# month later...}
+			other {# months later...}
+		}}
+		week {{num, plural,
+			one {# week later...}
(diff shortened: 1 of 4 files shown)


Sesquipedalian:
Any objections?

22522404:
Shouldn't it be: ago?
like: X months ago

Sesquipedalian:
No. If the posts are being displayed in the normal chronological order, then as one reads down the page one is going forward in time. Thus, "later" is the appropriate word.

However, if the posts are being displayed in reverse chronological order, then "earlier" is the appropriate word.

The word "ago" could only ever be used in the second case. However, it would not be as appropriate as "earlier" because of the subtle difference in meaning between those two words. "Ago" generally means a time in the past relative to the present moment, whereas "earlier" just means a time in the past relative to some other arbitrary time. Since these messages are indicating that there was a long period of time between two different posts, both of which were in the past, "earlier" is the appropriate word.

jdarwood007:
Why query this at all?  In the loop, hold the previous timestamp and when the message your processing has a timestamp greater than x time, then you show the bump.

Sesquipedalian:
> hold the previous timestamp

Um, that already is what this code does in nearly all cases. I think you overlooked the `if (!isset($prev_timestamp)) {` line.

The extra query is only executed for the first post on the second and following pages of a multipage topic. We have to do it in that one specific situation because the timestamp of the last post on the previous page of the topic is not known.

jdarwood007:
Ahh, I didn't see that the previous timestamp was an already an existing variable, thought it was newly created and didn't make sense.  Pagination makes sense for why.

Sesquipedalian:
So, any objection to me merging this, @jdarwood007?

Oldiesmann:
IP Community (aka IPB) does it this way. An example from a forum I'm a member of. The previous post in this topic was from November of 2021.
Screenshot_20250328_115139

PR #8530, opened 2025-03-25
[3.0] Removes letter links from member list

Sesquipedalian:
SMF is an international software package, but this feature on the member list assumed the English alphabet and doesn't play nicely with non-ASCII characters at all. Plus, this feature honestly isn't very useful anyway.

### DIFF
--- a/Sources/Actions/Memberlist.php
+++ b/Sources/Actions/Memberlist.php
@@ -294,30 +294,7 @@ class Memberlist implements ActionInterface, Routable
 			$_REQUEST['sort'] = 'real_name';
 		}
 
-		if (!is_numeric($_REQUEST['start'])) {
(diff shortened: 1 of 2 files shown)


Sesquipedalian:
Any objections?

Sources/Actions/Memberlist.php
```diff
@@ -294,30 +294,7 @@ public function all(): void
...
-			);
-			list($start) = Db::$db->fetch_row($request);
-			$start = (int) $start;
-			Db::$db->free_result($request);
-		} else {
-			$start = (int) $_REQUEST['start'];
-		}
+		$start = (int) $_REQUEST['start'];
 
 		Utils::$context['letter_links'] = '';
```
live627, line 299: gonna get rid of this, too?

https://github.com/SimpleMachines/SMF/blob/6a949d59fc356e456845e440af952c1840815d27/Sources/Actions/Memberlist.php#L325

live627:
No objections from me.

PR #8532, opened 2025-03-27
spriintf params in ItemList should start empty

live627:
Multiple areas in the admin panel and elsewhere use  `sprintf` to format items in `ItemList`. This restores expected behavior. Testing in the admin memberlist.

- Introduced in #8484
- Bug report: #8531
- Reverts last minute fix in  #8526

### DIFF
--- a/Sources/Actions/Admin/Tasks.php
+++ b/Sources/Actions/Admin/Tasks.php
@@ -202,9 +202,9 @@ class Tasks implements ActionInterface
 						'style' => 'width: 40%;',
 					],
 					'data' => [
-						'format_text' => [
+						'sprintf' => [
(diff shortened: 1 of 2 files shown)


live627:
I have noticed  that trailing `0` is dropped. Something in `Lang::formatText` is doing it.

live627:
nvm just me confused over numbers.

Sesquipedalian:
[Revert "Fixes display of scheduled tasks"](https://github.com/SimpleMachines/SMF/pull/8532/commits/06f0f13eeca240b6a272342286c0cc0f290c9aac)

Why? The `format_text` approach is technically superior to the `sprintf` approach and slightly faster.

PR #8536, opened 2025-03-29
Used `format_text` instead of `sprintf` for ItemList formatting

Sesquipedalian:


jdarwood007:
data => format_text => format.
Maybe it would be better to do:
data => format => [text, params]

PR #8556, opened 2025-04-02
Add methods for granular manipulation of debugging info

live627:
- Add custom info to show with all the other debug info
- `SMF\Debug\DebugUtils` (renamed from SMF\DebugUtils`)
- Each entry to show is a `DebugContextEntry ` object
- Use the HTML `<details>` tag instead of the old fashioned show/hide links
- Adds a new debug entry to show a few properties of the current `Action`

### Why

The portal mod I'm developing adds debug items for each loaded module. I use a buffer hook right now. Be great if there was a dedicated way that's not a hack.


Sources/Subs-Compat.php
```diff
@@ -8097,7 +8097,7 @@ function getMembersOnlineStats(array $membersOnlineOptions): array
 	 */
 	function displayDebug(): void
 	{
-		SMF\Logging::displayDebug();
+		SMF\Debug\DebugUtils::displayDebug();
```
Sesquipedalian, line 8100: I believe a similar change will be needed in Utils::obExit() as well.

Sesquipedalian (changes requested):
I like it. I think there is just one small change needed.

Sources/Subs-Compat.php
live627: commit amended. I seem to have missed copying that file over from the document root to the working tree.

Sesquipedalian (approved):

PR #8563, opened 2025-04-05
[3.0] Address install loop

sbulen:
Fixes #8559 

I think taking the substring here is a byproduct of ye olde bygone days...  

What it's doing today is taking substr('en_US', 8, -4), which is an empty string.  

Confirmed by testing that this gets rid of my loop issues.

PR #8564, opened 2025-04-05
Automatically fall back to en_US when Config::$language is invalid

Sesquipedalian:
This is a follow-up to #8563. It should catch any situation where something tries to set Config::$language to an invalid value.


Sesquipedalian:
@sbulen

sbulen (approved):
LGTM!

PR #8575, opened 2025-04-12
[2.1] Accepts wildcards in email-based ban triggers

### LINKED ISSUE #8571: [2.1|3.0] Unable to ban domains in emails
https://github.com/SimpleMachines/SMF/blob/69a8ac4373764a66facc27073ea27c401402072d/Themes/default/ManageBans.template.php#L138

Due to `type="email"`, you can not ban for instance `*@*.evil`.

This affects 2.1 and 3.0

Sesquipedalian:
Fixes #8571 for 2.1

### DIFF
--- a/Themes/default/ManageBans.template.php
+++ b/Themes/default/ManageBans.template.php
@@ -133,7 +133,7 @@ function template_ban_edit()
 							<label for="email_check">', $txt['ban_on_email'], '</label>
 						</dt>
 						<dd>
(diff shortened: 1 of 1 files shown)


jdarwood007 (approved):
I expected this was going to be the fix.  I suspect we did type of email to try to validate, but it doesn't work well for bans since we need to introduce wildcards.

For 3.0, I was thinking we could try to use the pattern syntax, but hadn't tested it: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/pattern

PR #8576, opened 2025-04-12
[3.0] Accepts wildcards in email-based ban triggers

### LINKED ISSUE #8571: [2.1|3.0] Unable to ban domains in emails

Sesquipedalian:
Fixes #8571 for 3.0


jdarwood007:
This fixes it and as I mentioned on the 2.1 fix, I wanted to try for 3.0 if we could handle validation with the https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/pattern syntax.

PR #8578, opened 2025-04-13
[3.0] Adds userland fallbacks for array functions added in PHP 8.4

Sesquipedalian:


jdarwood007:
Should we add documentation stating what version they polyfill up to, so when our min exceeds that, its easy to remove these?
```
/*
 * @Polyfill: PHP 8.3 and below
*/
```

Sesquipedalian:
Could, I suppose. Feel free to send up a PR for that.

PR #8580, opened 2025-04-16
[3.0] Refactors BBCode definitions into classes plus other improvements

Sesquipedalian:
The main change in this PR is [Refactors BBCode definitions into classes](https://github.com/SimpleMachines/SMF/pull/8580/commits/3a3e7c93ee9e22891af906df3839805eb756bff0)

Also fixes an unreported bug:
1. [Fully respects changes made by a parsed content BBC's validate method](https://github.com/SimpleMachines/SMF/pull/8580/commits/7f6863e0691213de86844c7f59b1b0ebd0484c26)

Additional feature improvements include:
1. [Adds support for 'quoted' => 'optional' for BBCode parameters](https://github.com/SimpleMachines/SMF/pull/8580/commits/91b061702f3c4f5d7b1c06230301d337e3f5c549)
2. [Makes BBCodeParser::insertTxt() static](https://github.com/SimpleMachines/SMF/pull/8580/commits/10f693c8327e539a36dfa6da6a51f339c779e629) and [Supports both txt and editortxt in BBCodeParser::insertTxt()](https://github.com/SimpleMachines/SMF/pull/8580/commits/25be611d9f6be48f256251f457360736f0f4b32a)
3. [Gives BBCode validate() methods read access to BBCodeParser properties](https://github.com/SimpleMachines/SMF/pull/8580/commits/86e56837b69327144238a2cdcb198adde3e6cbaa)
4. [Rearranges a couple of editor buttons](https://github.com/SimpleMachines/SMF/pull/8580/commits/05664be3f022c879ac5bfb62c5ba5acc4df15013)


Sources/Parsers/BBCodeParser.php
```diff
@@ -222,565 +226,95 @@ class BBCodeParser extends Parser
...
+		['class' => 'SMF\BBCode\Td'],
+		['class' => 'SMF\BBCode\Time1'],
+		['class' => 'SMF\BBCode\Time2'],
+		['class' => 'SMF\BBCode\Tr'],
+		['class' => 'SMF\BBCode\Tt'],
+		['class' => 'SMF\BBCode\U'],
+		['class' => 'SMF\BBCode\Url1'],
+		['class' => 'SMF\BBCode\Url2'],
+		['class' => 'SMF\BBCode\White'],
+		['class' => 'SMF\BBCode\YouTube'],
```
jdarwood007: Shouldn't this be calls like:
```
['class' => SMF\BBCode\YouTube::class],
```

Strictly typed and allows for CI/CD pipelines to identify that it exists.

Sources/BBCode/BBCode.php
```diff
@@ -0,0 +1,300 @@
...
+	public static array $types = [
+		null,
+		'unparsed_equals',
+		'parsed_equals',
+		'unparsed_content',
+		'closed',
+		'unparsed_commas',
+		'unparsed_commas_content',
+		'unparsed_equals_content',
+	];
```
jdarwood007: Just asking why we don't define constants here.

Sources/Parsers/BBCodeParser.php
Sesquipedalian: Derp. Yes, it should.

Sources/BBCode/BBCode.php
Sesquipedalian: It was easier to loop through them as an array in a few places.

jdarwood007: Would self:: not work? I'm not saying anything is wrong; I'm just ensuring we are trying to fit into the uniform code design we have been using so far in 3.0. I would prefer enums, but we have to wait until our minimum is 8.1 or higher.

```
	public const null TYPE_UNDEFINED = null;
	public const string TYPE_CLOSED = 'closed';
	public const string TYPE_PARSED_EQUALS = 'parsed_equals';
	public const string TYPE_UNPARSED_CONTENT = 'unparsed_content';
	public const string TYPE_UNPARSED_COMMAS = 'unparsed_commas';
	public const string TYPE_UNPARSED_COMMAS_CONTENT = 'unparsed_commas_content';
	public const string TYPE_UNPARSED_EQUALS = 'unparsed_equals';
	public const string TYPE_UNPARSED_EQUALS_CONTENT = 'unparsed_equals_content';


	public static array $types = [
		self::TYPE_UNDEFINED,
		self::TYPE_UNPARSED_EQUALS,
		self::TYPE_PARSED_EQUALS,
		self::TYPE_UNPARSED_CONTENT,
		self::TYPE_CLOSED,
		self::TYPE_UNPARSED_COMMAS,
		self::TYPE_UNPARSED_COMMAS_CONTENT,
		self::TYPE_UNPARSED_EQUALS_CONTENT,
	];
```

Sesquipedalian: I suppose that would work, yeah. 🙂

Sesquipedalian:
Look good, @jdarwood007?

jdarwood007:
Yes.  I haven't tested any code yet, though.  I was going to review how mods still work, since BBC mods are fairly common.
Looking forward to seeing how we support both using the class and the pre-3.0 style.

Sesquipedalian:
Oh, yeah, that's simple enough to explain.

1. In `BBCodeParser::integrateBBC()`, after the hook itself has been called, we immediately call `BBCodeParser::loadBBCodeClasses()`.
2. `BBCodeParser::loadBBCodeClasses()` loops through all items in `BBCodeParser::$codes`. When it finds an array using the old-style format, it feeds that array into a new instance of `GenericBBCode` and then replaces the entry in `BBCodeParser::$codes` with the resulting object.
3. `GenericBBCode::__construct()` examines and validates all the information in the array and sets the object's properties to the appropriate values.

The one bit that is slightly more complicated is the handling of the `validate` key in old-style BBCode definitions. For those, the object's `validationCallback` property is set to whatever the value of the `validate` key in the array is. Then the object's real `validate()` method simply calls that callback.

PR #8586, opened 2025-04-17
Logs AJAX errors to the JavaScript console

Sesquipedalian:
This will help with debugging and support


jdarwood007:
I would have suggested using `console.error`

Sesquipedalian:
That would make more sense, yes. Not sure why I didn't do that.

PR #8593, opened 2025-04-20
[3.0] Adds details and spoiler BBCodes

Sesquipedalian:
This adds two new BBCodes.

# Details

The format for this BBCode is `[details summary="summary text"]parsed content[/details]`. When transformed into HTML, it is rendered as a `<details>` and `<summary>` structure. It is suitable for block-level spoilers or any other content that could be considered as optional details, such as a changelog or some off-topic discussion.

Editor button and dropdown:
Screenshot 2025-04-20 at 6 26 05 PM

Using `details` in the WYSIWYG editor:
Screenshot 2025-04-20 at 2 11 05 PM

Displaying a spoiler:
Screenshot 2025-04-20 at 2 12 19 PM
Screenshot 2025-04-20 at 2 12 28 PM

Another example of suitable content:
Screenshot 2025-04-20 at 2 23 13 PM
Screenshot 2025-04-20 at 2 23 23 PM

# Spoiler

This BBCode has a more complicated implementation. It supports three forms for the sake of compatibility with spoiler mods for previous versions of SMF:
```
[spoiler]parsed content[/spoiler]
```
```
[spoiler text="summary text"]parsed content[/spoiler]
```
```
[spoiler="summary text"]parsed content[/spoiler]
```

The second and third forms are always rendered like a `details` BBCode.

The first form is more complex:
- If the first form contains multiple lines of text it will be rendered as if it were a `details` BBCode using the default summary string "Spoiler". 
- Otherwise, it will be rendered as an inline spoiler. 
     - These inline spoilers appear like a span of redacted text (see images below) until the user clicks on them in order to reveal the content.
    - The "redacted" effect for inline spoilers always uses the same colour as the surrounding text's colour. So for a dark-mode theme, where light text is used on a dark background, the "redacted" effect will appear as a span of text that has been "whited-out" rather than "blacked-out."
    - In order to support assistive technologies (e.g. screen readers) and people using keyboard navigation, an invisible button is embedded within each inline spoiler span. This button can be targeted and pressed in order to show or hide the spoiler content. The button has the title "Show or hide spoiler content" in order to indicate to screen readers what the button does.
    - The cursor changes to a pointer icon (just like for links, etc.) when hovering over the spoiler in order to indicate to the user that the content is clickable. (Unfortunately, the cursor is not visible in the screenshots below, but that's what it does.)

Editor button and display in WYSIWYG mode:
Screenshot 2025-04-20 at 2 58 07 PM
(Note how the spoiler text is visible inside a grey dashed outline. This is so that the author can see the text.)

Display in the hidden state:
Screenshot 2025-04-20 at 3 00 53 PM

Display in the hidden state while using a screen reader to target the invisible button:
Screenshot 2025-04-20 at 3 04 55 PM

Display after clicking on the spoiler to reveal the content:
Screenshot 2025-04-20 at 3 01 55 PM

I put a good deal of effort into making sure that this SMF-native implementation of a spoiler BBCode will support spoiler tags created by mods in previous versions. As far as I can tell, this supports the tags created by all known spoiler mods, so forums that used those mods will see their existing spoiler tags work as expected in SMF 3.0.


Themes/default/scripts/script.js
```diff
@@ -34,6 +34,19 @@ if (!('getElementsByClassName' in document))
...
+	for (const elem of document.querySelectorAll('.bbc_inline_spoiler')) {
+		elem.addEventListener('click', toggleSpoiler);
+	}
+});
+
+// Shows or hides the content of a spoiler tag.
+function toggleSpoiler() {
+	const button = this.firstElementChild;
+	button.nextElementSibling.style.visibility = (button.nextElementSibling.style.visibility === 'visible' ? 'hidden' : 'visible');
+	this.style.backgroundColor = (this.style.backgroundColor === 'transparent' ? 'currentcolor' : 'transparent');
```
ThistleSifter: Can both of these colours be moved somewhere where they can be customised (eg a custom property)? `currentcolor` won't necessarily be desirable as the spoiler colour, especially in a dark theme. As for using `transparent` as the 'shown' background, I'd suggest using a background that is slightly visible to indicate that it came from a spoiler, especially since you can toggle back and forth.

ThistleSifter (dismissed):

Themes/default/scripts/script.js
Sesquipedalian: I think you misunderstand what `currentcolor` will do here. It forces the background to match the text colour. So if the text is a light colour on a dark background, the background colour applied to the hidden spoiler will also be the same light colour. I chose `currentcolor` precisely because it is the best way to support both light and dark themes with the same CSS.

It would be possible to apply some non-transparent background to a revealed spoiler, yes, but I don't see the benefit to that. Using `transparent` both guarantees compatibility with all themes and keeps the JavaScript used to toggle between states simple. It would be possible on both counts to change that, but it would be more complex and thus easier for a custom theme to mess up.

Sesquipedalian: Still, I suppose I could change this to toggle a class instead.

ThistleSifter: What about something like this?
```js
function toggleSpoiler() {
    const button = this.firstElementChild;
    const showing = button.nextElementSibling.style.visibility === 'visible';
    button.nextElementSibling.style.visibility = (showing ? 'hidden' : 'visible');
    this.style.backgroundColor = (showing ? 'var(--bbc-spoilerinline-show-color, currentcolor)' : 'var(--bbc-spoilerinline-hide-color, transparent)');
}
```

Sesquipedalian: That could have worked, but [f82d5e7](https://github.com/SimpleMachines/SMF/pull/8593/commits/f82d5e76c97c3d5eaa1b529312a942f7671d3b86) is simpler. 🙂

Sources/BBCode/Spoiler2.php
```diff
@@ -0,0 +1,92 @@
...
+
+	/**
+	 *
+	 */
+	public string $tag = 'spoiler';
+
+	/**
+	 *
+	 */
+	public ?string $type = 'unparsed_equals';
```
ThistleSifter: Should this be `BBCode::TYPE_UNPARSED_EQUALS` instead?

Themes/default/css/index.css
```diff
@@ -481,6 +481,59 @@ blockquote cite::before {
...
+	cursor: pointer;
+	border-radius: 3px;
+	transition: background 0.5s;
+	background-color: currentcolor;
+	position: relative;
+}
+.bbc_inline_spoiler.revealed {
+	background-color: transparent;
+}
+.bbc_inline_spoiler button {
```
ThistleSifter: To be on the safe side, since this BBCode can include other tags, should this be `.bbc_inline_spoiler > button` (and similar in line 522, 525 & 534)?

Themes/default/css/index.css
```diff
@@ -481,6 +481,59 @@ blockquote cite::before {
...
+	outline: 2px solid #7fb0d8;
+}
+.bbc_inline_spoiler span {
+	/*
+		The visibility value is vital to making spoiler tags work correctly with
+		assistive technologies (e.g. screen readers). If you are customizing the
+		CSS for your theme, you should not change this part.
+	*/
+	visibility: hidden;
+	transition: visibility 0.5s;
```
ThistleSifter: You could use the following so that there is only one duration value to override for customising:
```css
transition-property: visibility;
transition-duration: inherit;
```

Themes/default/css/noscript.css
```diff
@@ -0,0 +1,9 @@
+/* This file holds CSS that is only used when JavaScript is disabled */
+
+/* Fallback method to show spoiler content */
+.bbc_inline_spoiler:hover {
+	background-color: transparent;
+}
+.bbc_inline_spoiler:hover span {
```
ThistleSifter: `.bbc_inline_spoiler:hover > span` ?

ThistleSifter (changes requested):
Looks good with the class. Spotted a couple of little things.

Sources/BBCode/Spoiler2.php
Sesquipedalian: Yes it should. Good catch!

Themes/default/css/index.css
Sesquipedalian: Yes, that is a good idea.

Themes/default/css/index.css
Sesquipedalian: Nice idea

ThistleSifter (commented):
I can't test it currently but that looks good to me.

PR #8606, opened 2025-04-25
[3.0] No longer uses closing PHP tags in SMF files

### LINKED ISSUE #7804: "No newline at end of file" policy
While I'm not strictly a developer, I'd like to discuss this policy.

As of now, our [README](https://github.com/SimpleMachines/SMF/blob/fd30b3714c17b30bb0821d79268caa8b9462a9c9/README.md?plain=1#L22C1-L22C1) doesn't even mention it, which raises the barrier to contribute, as it has caused [rework](https://github.com/search?q=repo%3ASimpleMachines%2FSMF+newline&type=pullrequests) for new and old contributors—myself included—in the past, and (AFAIK) still makes it impossible to use the GitHub web editor (because that adds a newline, a behavior which can't be changed). If the policy is documented elsewhere, I didn't find it.

I'm not a PHP expert, but [this SO discussion](https://stackoverflow.com/questions/9921927/php-is-there-a-reason-to-put-a-newline-at-the-end-of-file) makes me think we use the policy based on a misinterpretation of the unclearly phrased and outdated [PSR-2](https://www.php-fig.org/psr/psr-2/#22-files), which has since been superseded by [PSR-12](https://www.php-fig.org/psr/psr-12/#22-files) and received better phrasing. (See also [this very helpful comment elsewhere](https://github.com/squizlabs/PHP_CodeSniffer/issues/2645), which links to a [discussion on that specific—phrasing, not policy—change](https://github.com/php-fig/fig-standards/pull/825).)

So I'd like us to either:
* document the current no-newline policy for new contributors
* drop the policy (my preference)

Sesquipedalian:
Fixes #7804 

Aside from removing the tags themselves, there have also been changes to the SMF\Config, SMF\ErrorHandler, SMF\PackageManager\PackageManager, SMF\Tasks\UpdateUnicode, and SMF\Security classes, as well as changes to ./other/update_timezones.php, .editor_config, and .php-cs-fixer.dist.php in order to support this.


jdarwood007 (approved):
Given our move to supporting more hooks, this is becoming less of an issue all the time.  We still need to do good testing with package manager related actions, as I worry we bust something that makes upgrading and downgrading impossible.  But with all the changes we have prepared, hopefully none arise.

PR #8627, opened 2025-05-01
Fix a few "undefined" errors

Oldiesmann:
#### Description
This fixes a few "undefined ..." errors:
1. "Undefined variable: $row" in Sources/Actions/BoardIndex.php
2. "Undefined variable: $groups_data" in Sources/Actions/Admin/Reports.phhp
3. Missing "use SMF\Mail;" statement in Sources/Actions/Profile/Activate.php, resulting in undefined SMF\Actions\Profile\Mail error
4. Call to undefined function sanitizeInputs in Sources\Actions\PollEdit2.php - Poll::sanitizeInputs was a protected method but PollEdit2 doesn't extend the Poll class

### DIFF
--- a/Sources/Poll.php
+++ b/Sources/Poll.php
@@ -1362,14 +1362,10 @@ class Poll implements \ArrayAccess
 		return (int) $most_active;
 	}
 
-	/*************************
-	 * Internal static methods
-	 *************************/
-
 	/**
 	 * Validates and sanitizes $_POST input for creating or editing a poll.
 	 */
-	protected static function sanitizeInput(array &$errors): void
+	public static function sanitizeInput(array &$errors): void
 	{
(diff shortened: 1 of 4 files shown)


Sources/Poll.php
```diff
@@ -1362,14 +1362,10 @@ protected function getMostActive(int &$options): int
...
 
-	/*************************
-	 * Internal static methods
-	 *************************/
-
 	/**
 	 * Validates and sanitizes $_POST input for creating or editing a poll.
 	 */
-	protected static function sanitizeInput(array &$errors): void
+	public static function sanitizeInput(array &$errors): void
```
Sesquipedalian: This is a good change, but please also move this method into the public static methods section higher up in the file.

Sesquipedalian (changes requested):

Sesquipedalian (approved):

PR #8629, opened 2025-05-03
Fixes a few more undefined variables

Oldiesmann:

#### Description
Fixes a couple more undefined variable issues. Also fixes missing parameters in in_array.

Note that due to how getAllWithPermissions works we may need to tweak the integration hook call since board_or_profile could be the ID of a permissions profile or a board, but this at least passes useful info to the hook rather than an undefined variable.

### DIFF
--- a/Sources/Group.php
+++ b/Sources/Group.php
@@ -1563,7 +1563,7 @@ class Group implements \ArrayAccess
 		// Reset all cached permissions.
 		Config::updateModSettings(['settings_updated' => time()]);
 
-		$set = current(GroupPermissionSet::load($profile, (int) $group));
+		$set = current(GroupPermissionSet::load($profile, (int) $this->id));
 
 		foreach ($set->permissions as $permission_name => $value) {
 			$permission = Permission::get($permission_name);
@@ -2240,10 +2240,10 @@ class Group implements \ArrayAccess
 		}
 
 		// Maybe a mod needs to tweak the list of allowed groups on the fly?
-		IntegrationHook::call('integrate_groups_with_permissions', [&$groups, $permissions, $board]);
+		IntegrationHook::call('integrate_groups_with_permissions', [&$groups, $permissions, $board_or_profile]);
 
 		// Call the deprecated integrate_groups_allowed_to hook.
-		self::integrateGroupsAllowedTo($groups, $permissions, $board);
+		self::integrateGroupsAllowedTo($groups, $permissions, $board_or_profile);
 
 		return $groups;
 	}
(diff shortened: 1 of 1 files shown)


Sources/Group.php
```diff
@@ -2240,10 +2240,10 @@ public static function getAllWithPermissions(string|array $permissions, ?int $bo
 		}
 
 		// Maybe a mod needs to tweak the list of allowed groups on the fly?
-		IntegrationHook::call('integrate_groups_with_permissions', [&$groups, $permissions, $board]);
+		IntegrationHook::call('integrate_groups_with_permissions', [&$groups, $permissions, $board_or_profile]);
 
 		// Call the deprecated integrate_groups_allowed_to hook.
-		self::integrateGroupsAllowedTo($groups, $permissions, $board);
+		self::integrateGroupsAllowedTo($groups, $permissions, $board_or_profile);
```
Sesquipedalian, line 2246: Hm. Yes, there is a problem here, but this solution won't suffice because `$board_or_profile` might be a profile. We need to make some more substantial changes to Group::getAllWithPermissions() to fix it. I will add some changes to this PR in order to do that.

Sources/Group.php
```diff
@@ -2786,9 +2786,9 @@ protected static function integrateGroupsAllowedTo(array &$groups, array $permis
 			IntegrationHook::call('integrate_groups_allowed_to', [&$allowed_denied, $permission, $board]);
 
 			foreach ($groups as $group => $group_permissions) {
-				if (in_array($allowed_denied['allowed'])) {
+				if (in_array($group, $allowed_denied['allowed'])) {
```
Sesquipedalian, line 2789:
```suggestion
				if (in_array($group_permissions, $allowed_denied['allowed'])) {
```

Sources/Group.php
```diff
@@ -2786,9 +2786,9 @@ protected static function integrateGroupsAllowedTo(array &$groups, array $permis
...
 					$groups[$group][$permission] = 1;
-				} elseif (in_array($allowed_denied['denied'])) {
+				} elseif (in_array($group, $allowed_denied['denied'])) {
```
Sesquipedalian, line 2791:
```suggestion
				} elseif (in_array($group_permissions, $allowed_denied['denied'])) {
```

Sesquipedalian (changes requested):

Sources/Group.php
Sesquipedalian: Wait, no. Ignore this.

Sources/Group.php
Sesquipedalian: Ignore this too. Derp.

PR #8637, opened 2025-05-08
More backcompat fixes

Oldiesmann:
#### Description
Fixes files that have the SMF\Actions namespace with a "use BackwardCompatibility" statement but no "use SMF\BackwardCompatibility".

### DIFF
--- a/Sources/Actions/Announce.php
+++ b/Sources/Actions/Announce.php
@@ -18,6 +18,7 @@ namespace SMF\Actions;
 use SMF\ActionInterface;
 use SMF\ActionRouter;
 use SMF\ActionTrait;
+use SMF\BackwardCompatibility;
 use SMF\Board;
 use SMF\BrowserDetector;
 use SMF\Config;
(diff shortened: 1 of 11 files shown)


Sesquipedalian:
Actually, in all of these cases the `use BackwardCompatibility;` statement should be removed, because none of these classes use the SMF\BackwardCompatibility trait's code at all.

While we're at it, we should also delete the obsolete SMF\Actions\Profile\BackwardCompatibility and remove any use statements for that, too.

Sesquipedalian (changes requested):
Actually, in all of these cases the `use BackwardCompatibility;` statement should be removed, because none of these classes do anything with the SMF\BackwardCompatibility trait's code.

While we're at it, we should also delete the obsolete SMF\Actions\Profile\BackwardCompatibility and remove any use statements referring to it in the SMF\Actions\Profile\\* classes, too.

Sesquipedalian:
@Oldiesmann, I checked out a copy of this branch, made those changes, and then merged.

Oldiesmann:
> @Oldiesmann, I checked out a copy of this branch, made those changes, and then merged.

That works. I'm at work right now so I didn't have a chance to get to it before you did.

Sesquipedalian:
No worries. Thanks for your continuing help!

Issue #8624, opened 2025-05-01
[3.0]: Guests cannot see boards when cookies are disabled


sbulen:
### Basic Information

Guests cannot see boards when cookies are disabled.  

Note the board has the proper permissions, -1,0,2.

This is followup to #8615 .  It's interfering with testing of #8399 .  Works fine in 2.1.

![Image](https://github.com/user-attachments/assets/3ad25f5e-ac5e-4940-9893-9afcc26c8f2d)

### Steps to reproduce

1.  Install from current GitHub.
2.  Logoff.
3.  Disable cookies on your browser.
4.  Navigate to newly installed forum.

### Expected result

As in prior SMF versions, normally the guests can see boards & navigate around.

### Actual result

No boards are visible.  See screenshot above.

### Version/Git revision

3.0 Alpha 3

### Database Engine

All

### Database Version

8.4.0

### PHP Version

8.4.5

### Logs

```shell

```

Sesquipedalian:
@sbulen, is this still happening? I cannot reproduce it.

sbulen:
Hmmm...  I think the behavior has changed a little...  This is what I am seeing today:

If I just visit the forum with cookies disabled, e.g., `https://myforum.org/index.php`, I see the above, i.e., no boards or categories. 

If I hit pretty much any form button, PHPSESSID is added to the url, and I can see all the boards/categories.

In 2.1, no such button press (to add PHPSESSID) is necessary.  Upon first visit, boards are visible.

So...  It seems the difference between 2.1 & 3.0 is that a session is now REQUIRED to see public boards & categories, and that wasn't the case before.

I could attempt to fix as part of #8399.  But something changed upstream, e.g., there is no similar issue testing #8394 .  

(It might take a while...  looks like I spawned another round of merge conflicts with #8399...)

sbulen:
I've been testing this & have discovered that navigating 100% by url, e.g., index.php?board=1 or index.php?topic=1 or index.php?msg=1 works fine.

So...  It's probably not a permissions issue.  Or even a session issue...  Possibly just a simple board index display issue...

Or possibly a category issue.  The only URLs that don't work are index.php & index.php#c1

sbulen:
I've narrowed it down to this line of code here:
https://github.com/SimpleMachines/SMF/blob/1b525178316533f807ca40dba1e28e6234dbacd3/Sources/Actions/BoardIndex.php#L464

[...]

Issue #8603, opened 2025-04-24
[3.0]: Permissions errors on new install


sbulen:
### Basic Information

Running into various permissions errors on a new install:

In SMF error log:  Permission "send_email_to_members" does not exist 

In PHP error log:  Undefined array key "postmod_active" in D:\wamp64\www\84van30\Sources\Permissions\Permission.php on line 1134

### Steps to reproduce

1. Install
2.

### Actual result

Installation completes, but forum is not usable.

### Version/Git revision

3.0 alpha 3

### Database Engine

All

### Database Version

8.4.4

### PHP Version

8.4.5

### Logs

```shell

```

Oldiesmann:
Similar things happen on an existing install from the latest code. The permission error prevents you from even being able to use the forum. I can't get to the error log to see if there's an error about "postmod_active", but I did run into an issue related to that while trying to run upgrade.php to see if that would fix the permission error.

sbulen:
I was able to get past the permission error by deleting all of the "send_email_to_members" rows from the smf_permissions table...

Got me working!

Oldiesmann:
That fixed it for me as well.

Oldiesmann:
Did some more digging. It looks like the "send_email_to_members" permission was renamed to "send_mail". There are still two references to the old permission in the language files but none in the code itself.

On a related note, an unknown permission should not break the forum. Log an error or just ignore it entirely.

Sesquipedalian:
> an error about "postmod_active"

> deleting all of the "send_email_to_members" rows

> an unknown permission should not break the forum. Log an error or just ignore it entirely.

Thanks, gentlemen. I will submit a PR shortly to address all of those.

Issue #8241, opened 2024-06-04
[3.0]:  Installer does not provide minimum database version upon failure due to unsupported version.


tyrsson:
### Basic Information

On running install.php against mysql 8.0.31 The installer does not provide any feedback as to what the minimum version should be.

### Steps to reproduce

1. Start a new install with any MySQL version < 8.0.35

### Expected result

That the UI would provide the minimum required version.

### Actual result

Just provides a notice to upgrade.

### Version/Git revision

51a1e38

### Database Engine

MySQL

### Database Version

8.0.31

DiegoAndresCortes:
Odd, I did get the error message.

https://github.com/SimpleMachines/SMF/blob/51a1e385257b484c11271df33a72d672b687423a/other/install.php#L933-L939

tyrsson:
I got the error. It just does not tell you what the minimum required version is. It should.

PR #8652, opened 2025-05-19
[3.0] Rewrites installer and upgrader

### LINKED ISSUE #7157: Upgrader forgets language override
#### Description
I recently attempted to do an upgrade of my "nightly" test forum, which I last did around May (:roll_eyes:). The upgrader detected outdated language files and gave me the option to use a `?lang=english` (or similar) override, but a couple of "Continue"s later, it threw that URL parameter away and I was back at an inconsistent mess of english and translated strings.

### Steps to reproduce
1. Have outdated language files
2. Click the upgrader link to "use/try english instead"
3. Continue through the upgrade
4. At around step 4, substep 1, the parameter is gone from the URL and there's non-english strings.

- Version/Git revision: HEAD
- Database Type: postgres
- Database Version: 9.4 (yes, the upgrade later failed for that reason; only 9.6+ is supported)
- PHP Version: 5.6.something

### LINKED ISSUE #7954: [3.0] Upgrader error reporting
#### Description
If there is *ONE THING* we need to address with the upgrader, it's the error reporting.

The only way in 2.0 to know what an error is is to run it in CLI, so errors are on the command line & are visible in the output.  When running via browser, as most folks do, the errors are hidden from the user since the code is all run via eval().  

Fine, if there are no issues...   But a big problem when there *are* issues.  You're flying blind.

### LINKED ISSUE #8261: [3.0]: Check for "calendar_holidays" table during upgrade could return false positives

### LINKED ISSUE #8570: [3.0]: 2.1 => 3.0 Upgrade Issues
### Basic Information

I realize there is a lot of open work on the upgrader now, just logging this so folks are aware.

If you run the upgrader from 2.1 to 3.0 today, it will properly do the mb4 udpates.

It will *NOT* properly do the other DB updates.  It's a mixed bag:
 - spoofdetector_name ends up on messages, not on members???  I have no idea how this happens...
 - calendar updates don't happen
 - various version columns aren't added

The bottom line is that the DB from an upgrade doesn't match the DB from an install.

### Steps to reproduce

1.
2.

### Expected result

_No response_

### Actual result

_No response_

### Version/Git revision

3.0 Alpha 3

### Database Engine

All

### Database Version

8.4.4

### PHP Version

8.4.5

### Logs

```shell

```

### Additional Information

_No response_

### LINKED ISSUE #8599: [3.0]: Upgrade fails with generic MySQL error
### Basic Information

Trying to run upgrade.php results in a white screen with the error ```Access denied; you need (at least one of) the SUPER privilege(s) for this operation```

### Steps to reproduce

1. Move/copy upgrade files from other to the forum root
2. Attempt to run upgrade.php

### Expected result

Upgrade.php completes successfully

### Actual result

As soon as it gets to the first step (or past the backup step if you chose to backup you tables), you get a screen saying "Access denied; you need (at least one of) the SUPER privilege(s) for this operation". Doesn't matter if you have debug info enabled or not.

### Version/Git revision

b6174e9

### Database Engine

MySQL

### Database Version

10.11.11

### PHP Version

8.4.6

### Logs

```shell

```

### Additional Information

Unfortunately I'm not able to debug this any further. It happens immediately after logging in/after completing the backups, so the step number varies depending on the option you've chosen.

### LINKED ISSUE #8616: [3.0]: Install error: Unable to load the en_US/ThemeStrings language file.
### Basic Information

I know we talked about this before, but now it's showing up in the error log!  😆

Error text:  Unable to load the en_US/ThemeStrings language file.

### Steps to reproduce

1.  Create an empty MySQL DB.
2.  Create an empty folder for your new forum.
3.  Sync GitHub so release-3.0 is current; checkout release-3.0.
4.  Copy all files & folders from your GitHub working folder EXCEPT: system files (no files starting with dots), composer files, DCO.txt.  The goal here is to mimic the contents of an install package.
5. Copy the 3 install*.* files down from /other to the forum root
6. Copy the 2 Settings*.* files down from /other to the forum root 
7. Execute the installer, pointing to your new DB, accepting all defaults (except stat collection)
8. When done, you have an error logged in the SMF error log

### Expected result

No errors

### Actual result

Error logged:  Unable to load the en_US/ThemeStrings language file.

### Version/Git revision

3.0 Alpha 3

### Database Engine

All

### Database Version

8.4

### PHP Version

8.4.5

### Logs

```shell
Error

Type of error
Template

Error message
Unable to load the en_US/ThemeStrings language file.

URL of page causing the error
http://localhost/84van30an/index.php

Backtrace information

    #0: SMF\Lang::load()
    Called from D:\wamp64\www\84van30an\Sources\Lang.php on line 600
    #1: SMF\Lang::txtExists()
    Called from D:\wamp64\www\84van30an\Sources\Permissions\Permission.php on line 931
    #2: SMF\Permissions\Permission->__construct()
    Called from D:\wamp64\www\84van30an\Sources\Permissions\Permission.php on line 1,215
    #3: SMF\Permissions\Permission::getAll()
    Called from D:\wamp64\www\84van30an\Sources\Permissions\GroupPermissionSet.php on line 328
    #4: SMF\Permissions\GroupPermissionSet::loadGlobalPermissionData()
    Called from D:\wamp64\www\84van30an\Sources\Permissions\GroupPermissionSet.php on line 300
    #5: SMF\Permissions\GroupPermissionSet::load()
    Called from D:\wamp64\www\84van30an\Sources\Group.php on line 2,272
    #6: SMF\Group::getAllowedTo()
    Called from D:\wamp64\www\84van30an\Sources\Config.php on line 1,124
    #7: SMF\Config::reloadModSettings()
    Called from D:\wamp64\www\84van30an\install.php on line 1,560
    #8: AdminAccount()
    Called from D:\wamp64\www\84van30an\install.php on line 172
```

### Additional Information

_No response_

### LINKED ISSUE #8648: Upgrade error due to missing global permissions
https://github.com/SimpleMachines/SMF/blob/1b525178316533f807ca40dba1e28e6234dbacd3/other/upgrade.php#L1995

```
Access denied; you need (at least one of) the SUPER or SYSTEM_VARIABLES_ADMIN privilege(s) for this operation
```

@Sesquipedalian

### LINKED ISSUE #8683: [2.1 & 3.0]: Consider disabling cron during patches, mod installs, etc, when source code is in flux
### Basic Information

It may make sense to disable cron during patching.  During patch testing, I occasionally see cron errors.  If you do enough patches, it'll happen...

2 examples during a recent test are on the forum test board.

I just got another one now.  When these ones come, they come in bursts of ~11 (which I think are the unicode updates):
Image

### Steps to reproduce

1. Intermittent issue, when cron kicks off while doing patch tests
2.

### Expected result

No errors

### Actual result

Cron errors as listed above

### Version/Git revision

2.1 (but can happen in any)

### Database Engine

All

### Database Version

8.4.4

### PHP Version

8.4.5

### Logs

```shell

```

### Additional Information

_No response_

Sesquipedalian:
This is the new installer and upgrader code. It is based on the work in #8093. 

It works in my tests, but it probably has bugs. Let's find 'em and fix 'em.

Fixes #7954
Fixes #8261
Fixes #8570
Fixes #8599
Fixes #8616
Fixes #8648
Fixes #8683


Sources/Db/Schema/Table.php
```diff
@@ -0,0 +1,278 @@
...
+	 *
+	 * @param DbIndex $index The index to update.
+	 * @return bool Whether or not the operation was successful.
+	 */
+	public function alterIndex(DbIndex $index): bool
+	{
+		// This method is really just a convenient way to replace an existing index.
+		$this->dropIndex($index);
+
+		return $this->addIndex($index);
```
jdarwood007: I should learn how to suggest change.  This should be.
```
		return $this->dropIndex($index) && $this->addIndex($index);
```

As we don't have a successful alter if it doesn't drop and add.

Sources/Db/Schema/Table.php
```diff
@@ -0,0 +1,278 @@
...
+
+/**
+ * Simple Machines Forum (SMF)
+ *
+ * @package SMF
+ * @author Simple Machines https://www.simplemachines.org
+ * @copyright 2023 Simple Machines and individual contributors
+ * @license https://www.simplemachines.org/about/smf/license.php BSD
+ *
+ * @version 3.0 Alpha 2
```
jdarwood007: Not going to comment them all.  Says Alpha 2.

Sources/Tasks/Utf8EntityDecode.php
```diff
@@ -0,0 +1,398 @@
...
+	/**
+	 * This executes the task.
+	 *
+	 * @return bool Always returns true.
+	 * @todo PHP 8.2: This can be changed to return type: true.
+	 */
+	public function execute(): bool
+	{
+		// Avoid leaving data in an inconsistent state.
+		ignore_user_abort(true);
```
jdarwood007, line 46: Minor detail, another PR maybe.  We should have a handler that goes into Sapi do this since we are instructing the Server API to ignore if the request aborts.

Sources/Tasks/Utf8EntityDecode.php
```diff
@@ -0,0 +1,398 @@
...
+
+		$time_limit = (int) (((ini_get('safe_mode') === false && Sapi::setTimeLimit(Taskrunner::MAX_CLAIM_THRESHOLD) !== false) ? Taskrunner::MAX_CLAIM_THRESHOLD : (int) ini_get('max_execution_time')) / 2);
```
jdarwood007: `safe_mode` was deprecated in 5.3.  We are long past needing to worry about it.

jdarwood007 (commented):
Is 2.0 and before stuff going up separately?

Sesquipedalian:
> Is 2.0 and before stuff going up separately?

This PR already includes the necessary migration steps to upgrade from 2.0. We had previously discussed the plan to make upgrading from 1.x a separate converter.

Sources/Tasks/Utf8EntityDecode.php
Sesquipedalian: That makes sense to me, sure.

Sesquipedalian: On further consideration, I'm not sure that would be necessary after all. It's such a simple function. Why would we need to wrap it?

jdarwood007: We give instructions to the Sapi more or less consistently and ensure we perform them in class. But it's a minor thing; we could ignore it for now and handle it later.

jdarwood007:
> > Is 2.0 and before stuff going up separately?
> 
> This PR already includes the necessary migration steps to upgrade from 2.0. We had previously discussed the plan to make upgrading from 1.x a separate converter.

I had planned to bake the converter logic into the installer/upgrader to make it still easy to convert Yabb, SMF 1.x and 2.0.  The reason is that we are still supporting them, but only through a conversion process.  Yabb and SMF 1.x cost a lot of technical debt to maintain the upgrader logic for.  So by moving into converters, we reduce that to maintaining how to convert data, which becomes easier with a select this and insert that process.

Sesquipedalian:
Sure. But that would all just be a matter of writing Migration and Cleanup classes, plus a Tools class to call them, right?

sbulen:
I just ran a 2.1 =>3.0 upgrade & it went very well.

_**EDIT:  I had to rerun for... reasons... Different issues reported below...**_

On completion, I got this error in the Cleanup phase:
Critical Error!
fileinode(): Argument # 1 ($filename) must be of type string, false given
D:\wamp64\www\van2130\Sources\Maintenance\Cleanup\v3_0\TasksDirCase.php:62

sbulen:
Just tried a plain 3.0 install & it failed on an error loop:
```
array(5) { 
[0]=> array(5) { ["file"]=> string(38) "D:\wamp64\www\84van30\Sources\Lang.php" ["line"]=> int(368) ["function"]=> string(3) "log" ["class"]=> string(16) "SMF\ErrorHandler" ["type"]=> string(2) "::" } 
[1]=> array(5) { ["file"]=> string(38) "D:\wamp64\www\84van30\Sources\Lang.php" ["line"]=> int(756) ["function"]=> string(4) "load" ["class"]=> string(8) "SMF\Lang" ["type"]=> string(2) "::" } 
[2]=> array(5) { ["file"]=> string(59) "D:\wamp64\www\84van30\Sources\Maintenance\Tools\Install.php" ["line"]=> int(182) ["function"]=> string(6) "getTxt" ["class"]=> string(8) "SMF\Lang" ["type"]=> string(2) "::" } 
[3]=> array(5) { ["file"]=> string(57) "D:\wamp64\www\84van30\Sources\Maintenance\Maintenance.php" ["line"]=> int(268) ["function"]=> string(8) "getSteps" ["class"]=> string(29) "SMF\Maintenance\Tools\Install" ["type"]=> string(2) "->" } 
[4]=> array(5) { ["file"]=> string(33) "D:\wamp64\www\84van30\install.php" ["line"]=> int(31) ["function"]=> string(7) "execute" ["class"]=> string(27) "SMF\Maintenance\Maintenance" ["type"]=> string(2) "->" } 
} Error loop.
```

sbulen:
_**EDIT:  I had to rerun the 2.1 => 3.0 upgrade for... reasons... Slightly different issues reported below...**_

Also just noticed on my 2.1 => 3.0 test that the upgrade files were left in the forum root directory.  They weren't deleted, likely because the cleanup phase didn't end cleanly.

Related, got this issue, the version wasn't updated:
_SMF file version (3.0 Alpha 3) does not match SMF database version (2.1.4).
Run the SMF upgrader to fix this._

Also noticed on the 2.1 => 3.0 test that one DB change was not applied:
 - smf_version not added to log_packages

sbulen:
I was really hoping the rewrite would include better error reporting.  Ideally a persistent log to help diagnose upgrader issues in the support forum.  I can't find any record of what actions were done or not done.

Per request here: https://github.com/SimpleMachines/SMF/issues/7954

sbulen:
FYI, you may find these helpful...  My tools that I use during testing & support activities were all udpated for SMF3.0.  

Including the database structure check (matches current release-3.0 install, & I'll try to keep it current going forward):
https://raw.githubusercontent.com/sbulen/sjrbTools/refs/heads/master/smf_db_compare.php

The settings dump (just displays them):
https://raw.githubusercontent.com/sbulen/sjrbTools/refs/heads/master/smf_settings_diag.php

The utf8(mb4) verification (ensure all columns, tables, & settings are correct):  
https://raw.githubusercontent.com/sbulen/sjrbTools/refs/heads/master/smf_utf8_diag.php

Just right click & download & run from forum root.

sbulen:
I am not seeing the ThemeStrings.php language file error.  That's been fixed.

jdarwood007:
Maybe the upgrader could drop and create a `{db_prefix}log_upgrade` that would store any errors that occurred during that run?  Suggesting it like this so we always ensure we don't try ever to upgrade that table, rebuild it.

sbulen:
I'd want positive or negative...  A log of what was done & what wasn't. 

In my mind, the text output from CLI is perfect.

I have more faith in a text log...  So many issues are db-related, and you stand a better chance of logging that.

sbulen:
Regarding the installer, it's still trying to load $txt from Install.php, but it appears that was renamed to Maintenance.php, & doesn't exist anymore.  Maintenance/Tools/Install.php still references "file: 'Install'" throughout.

sbulen:
I had to rerun my 2.1 => 3.0 upgrader test due to an issue on my side.  I updated the comments above accordingly.

The main problem is the cleanup hangs for some reason, & there's fallout from that.

Question:
 - On the 2.1 => 3.0 upgrade, it didn't rerun the 2.1 upgrade steps.  Are we changing this going forward?   I suspect they did this because there were DB changes introduced in point releases during 1.x and had to clean things up.  I don't think they could assume a given DB state.  Not sure we need to do this anymore.

sbulen:
In CLI, the logs report other errors during the UTF8 step.  

Minor nit, but possibly related... The counter on the UTF8 step when run from the browser always displays 0, e.g., it always says "Working on file 0 of 72" and never increments.  (I think it's overrunning the array, & the browser is hiding these errors...)

```
 +++ Converting table smf_user_alerts_prefs to utf8mb4... done.
 +++ Converting table smf_user_drafts to utf8mb4... done.
 +++ Converting table smf_user_likes to utf8mb4... done.
Warning: Undefined array key 72 in D:\wamp64\www\van2130\Sources\Maintenance\Utf8ConverterStep.php on line 651

Call Stack:
    0.0005     394264   1. {main}() D:\wamp64\www\van2130\upgrade.php:0
    0.0207    1413888   2. SMF\Maintenance\Maintenance->execute($type = 2) D:\wamp64\www\van2130\upgrade.php:24
   13.9752    8924544   3. call_user_func:{D:\wamp64\www\van2130\Sources\Maintenance\Maintenance.php:287}($callback = [0 => class SMF\Maintenance\Utf8ConverterStep { private int ${SMF\Maintenance\Step}id = 5; private string ${SMF\Maintenance\Step}name = 'Convert to UTF-8'; private ?string ${SMF\Maintenance\Step}title = NULL; private array|string ${SMF\Maintenance\Step}function = ...; private string ${SMF\Maintenance\Step}template = 'convertUtf8'; private int ${SMF\Maintenance\Step}progress = 30; public array $supported_charsets = [...]; public string $lang_charset = 'UTF-8'; public array $charset_maps = [...] }, 1 => 'convertDatabase']) D:\wamp64\www\van2130\Sources\Maintenance\Maintenance.php:287
   13.9752    8924728   4. SMF\Maintenance\Utf8ConverterStep->convertDatabase() D:\wamp64\www\van2130\Sources\Maintenance\Maintenance.php:287


Warning: Attempt to read property "name" on null in D:\wamp64\www\van2130\Sources\Maintenance\Utf8ConverterStep.php on line 654

Call Stack:
    0.0005     394264   1. {main}() D:\wamp64\www\van2130\upgrade.php:0
    0.0207    1413888   2. SMF\Maintenance\Maintenance->execute($type = 2) D:\wamp64\www\van2130\upgrade.php:24
   13.9752    8924544   3. call_user_func:{D:\wamp64\www\van2130\Sources\Maintenance\Maintenance.php:287}($callback = [0 => class SMF\Maintenance\Utf8ConverterStep { private int ${SMF\Maintenance\Step}id = 5; private string ${SMF\Maintenance\Step}name = 'Convert to UTF-8'; private ?string ${SMF\Maintenance\Step}title = NULL; private array|string ${SMF\Maintenance\Step}function = ...; private string ${SMF\Maintenance\Step}template = 'convertUtf8'; private int ${SMF\Maintenance\Step}progress = 30; public array $supported_charsets = [...]; public string $lang_charset = 'UTF-8'; public array $charset_maps = [...] }, 1 => 'convertDatabase']) D:\wamp64\www\van2130\Sources\Maintenance\Maintenance.php:287
   13.9752    8924728   4. SMF\Maintenance\Utf8ConverterStep->convertDatabase() D:\wamp64\www\van2130\Sources\Maintenance\Maintenance.php:287


 +++ ... 
Warning: Attempt to read property "test_args" on null in D:\wamp64\www\van2130\Sources\Maintenance\Utf8ConverterStep.php on line 658

Call Stack:
    0.0005     394264   1. {main}() D:\wamp64\www\van2130\upgrade.php:0
    0.0207    1413888   2. SMF\Maintenance\Maintenance->execute($type = 2) D:\wamp64\www\van2130\upgrade.php:24
   13.9752    8924544   3. call_user_func:{D:\wamp64\www\van2130\Sources\Maintenance\Maintenance.php:287}($callback = [0 => class SMF\Maintenance\Utf8ConverterStep { private int ${SMF\Maintenance\Step}id = 5; private string ${SMF\Maintenance\Step}name = 'Convert to UTF-8'; private ?string ${SMF\Maintenance\Step}title = NULL; private array|string ${SMF\Maintenance\Step}function = ...; private string ${SMF\Maintenance\Step}template = 'convertUtf8'; private int ${SMF\Maintenance\Step}progress = 30; public array $supported_charsets = [...]; public string $lang_charset = 'UTF-8'; public array $charset_maps = [...] }, 1 => 'convertDatabase']) D:\wamp64\www\van2130\Sources\Maintenance\Maintenance.php:287
   13.9752    8924728   4. SMF\Maintenance\Utf8ConverterStep->convertDatabase() D:\wamp64\www\van2130\Sources\Maintenance\Maintenance.php:287


Warning: Trying to access array offset on null in D:\wamp64\www\van2130\Sources\Maintenance\Utf8ConverterStep.php on line 658

Call Stack:
    0.0005     394264   1. {main}() D:\wamp64\www\van2130\upgrade.php:0
    0.0207    1413888   2. SMF\Maintenance\Maintenance->execute($type = 2) D:\wamp64\www\van2130\upgrade.php:24
   13.9752    8924544   3. call_user_func:{D:\wamp64\www\van2130\Sources\Maintenance\Maintenance.php:287}($callback = [0 => class SMF\Maintenance\Utf8ConverterStep { private int ${SMF\Maintenance\Step}id = 5; private string ${SMF\Maintenance\Step}name = 'Convert to UTF-8'; private ?string ${SMF\Maintenance\Step}title = NULL; private array|string ${SMF\Maintenance\Step}function = ...; private string ${SMF\Maintenance\Step}template = 'convertUtf8'; private int ${SMF\Maintenance\Step}progress = 30; public array $supported_charsets = [...]; public string $lang_charset = 'UTF-8'; public array $charset_maps = [...] }, 1 => 'convertDatabase']) D:\wamp64\www\van2130\Sources\Maintenance\Maintenance.php:287
   13.9752    8924728   4. SMF\Maintenance\Utf8ConverterStep->convertDatabase() D:\wamp64\www\van2130\Sources\Maintenance\Maintenance.php:287


Fatal error: Uncaught Error: Call to a member function isCandidate() on null in D:\wamp64\www\van2130\Sources\Maintenance\Utf8ConverterStep.php on line 660

Error: Call to a member function isCandidate() on null in D:\wamp64\www\van2130\Sources\Maintenance\Utf8ConverterStep.php on line 660

Call Stack:
    0.0005     394264   1. {main}() D:\wamp64\www\van2130\upgrade.php:0
    0.0207    1413888   2. SMF\Maintenance\Maintenance->execute($type = 2) D:\wamp64\www\van2130\upgrade.php:24
   13.9752    8924544   3. call_user_func:{D:\wamp64\www\van2130\Sources\Maintenance\Maintenance.php:287}($callback = [0 => class SMF\Maintenance\Utf8ConverterStep { private int ${SMF\Maintenance\Step}id = 5; private string ${SMF\Maintenance\Step}name = 'Convert to UTF-8'; private ?string ${SMF\Maintenance\Step}title = NULL; private array|string ${SMF\Maintenance\Step}function = ...; private string ${SMF\Maintenance\Step}template = 'convertUtf8'; private int ${SMF\Maintenance\Step}progress = 30; public array $supported_charsets = [...]; public string $lang_charset = 'UTF-8'; public array $charset_maps = [...] }, 1 => 'convertDatabase']) D:\wamp64\www\van2130\Sources\Maintenance\Maintenance.php:287
   13.9752    8924728   4. SMF\Maintenance\Utf8ConverterStep->convertDatabase() D:\wamp64\www\van2130\Sources\Maintenance\Maintenance.php:287


```

Sesquipedalian:
Thanks for the reports, @sbulen! I think they have all been fixed with the latest commits.

Sesquipedalian:
> I was really hoping the rewrite would include better error reporting. Ideally a persistent log to help diagnose upgrader issues in the support forum. I can't find any record of what actions were done or not done.
> 
> Per request here: #7954

Ask and ye shall receive. When the debug option is enabled in the upgrader, the final step will show a complete log of all actions. This log contains the exact same output as the CLI logging. Also, the contents of the log have been further improved.

Edit: However, the log file is not persistent. It is automatically deleted once the installation or upgrade process has completed successfully. This is because the log file might contain sensitive information and therefore should not be left lying around for some random attacker to find through a simple HTTP request to `http://example.com/forum/upgrade.log`.

sbulen:
Not sure if I should record this here or elsewhere, but a slight variation of #8603 has popped up after my first test.

My error log has hundreds of these - about 80 per page load:
> http s://localhost/van2130/index.php?action=admin;area=logs;sa=errorlog;desc=
> [D:/wamp64/www/van2130/Sources/Permissions/Permission.php](http s://localhost/van2130/index.php?action=admin;area=logs;sa=errorlog;file=RDovd2FtcDY0L3d3dy92YW4yMTMwL1NvdXJjZXMvUGVybWlzc2lvbnMvUGVybWlzc2lvbi5waHA=;line=1180) (Line 1,180)
> [Backtrace information](http s://localhost/van2130/index.php?action=admin;area=logs;sa=errorlog;backtrace=178)
> 
> Type of error: General
> Error message Select
> Tried to check an unknown permission: send_email_to_members.

Sesquipedalian:
> send_email_to_members

Ugh. This thing again. 

Unfortunately, the SMF 2.1 installer and upgrader are inconsistent regarding this permission. SMF 2.1 itself doesn't use it for anything, and the 2.1 upgrader removes it, but the 2.1 installer still creates it.

We can add an upgrade step to remove it if found.

sbulen:
I want to come back to this...

>  On the 2.1 => 3.0 upgrade, it didn't rerun the 2.1 upgrade steps.  Are we changing this going forward?   
> I suspect they did this because there were DB changes introduced in point releases during 1.x and had to clean 
> things up.  I don't think they could assume a given DB state. 

A couple cases in point:  
 - #8212 
 - #8458 

If I upgrade from 2.1.4 (or 2.1.1 - 2.1.3...) to 3.0, neither of those DB changes will be applied.  

This is the reason that in the past, all steps from the current to the target were applied - they couldn't assume you had all of the current (2.1.x in this example) DB updates applied before moving to the 3.0 step.  

So...  In my mind, either (A) we rerun the 2.1 step before the 3.0 step - OR - (B) we make ABSOLUTELY NO DB changes, not even indexes, in point releases (and revert the two changes above prior to 2.1.5) - OR - (C) we add a whole new mechanism to the upgrader to handle point release DB changes.

I'd prefer A or B.  I think C would be a waste of time - that's what option A does.  We just need to stick to the rule that the upgrader must be re-runnable.

sbulen:
Note a corrollary to the above:  The upgrade_x-y_MySQL.sql files must be kept in perfect sync across all releases.  I.e., any change in the 3.0 branches to the .sql files must be made in the 2.1 branch files & vice-versa.

Ideally, they're separate & we just maintain one set.   2.0 and 2.1 had drifted significantly over time - I spent a lot of time bringing the 1.x & 2.x .sql files back in sync across the 2.0 & 2.1 repos...  

It was painful...  Different upgrade paths produced different DB structures...

Sesquipedalian:
> I want to come back to this...
> 
> > On the 2.1 => 3.0 upgrade, it didn't rerun the 2.1 upgrade steps.  Are we changing this going forward?
> > I suspect they did this because there were DB changes introduced in point releases during 1.x and had to clean
> > things up.  I don't think they could assume a given DB state.
> 
> A couple cases in point:
> 
> * [Adds index on id_board to log_notify table #8212](https://github.com/SimpleMachines/SMF/pull/8212)
> * [[2.1] Returns correct search results when post moderation is enabled #8458](https://github.com/SimpleMachines/SMF/pull/8458)
> 
> If I upgrade from 2.1.4 (or 2.1.1 - 2.1.3...) to 3.0, neither of those DB changes will be applied.
> 
> This is the reason that in the past, all steps from the current to the target were applied - they couldn't assume you had all of the current (2.1.x in this example) DB updates applied before moving to the 3.0 step.
> 
> So... In my mind, either (A) we rerun the 2.1 step before the 3.0 step - OR - (B) we make ABSOLUTELY NO DB changes, not even indexes, in point releases (and revert the two changes above prior to 2.1.5) - OR - (C) we add a whole new mechanism to the upgrader to handle point release DB changes.
> 
> I'd prefer A or B. I think C would be a waste of time - that's what option A does. We just need to stick to the rule that the upgrader must be re-runnable.

Actually, this is trivial to take care of with the new installer and upgrader system. 4305a83 has fixed the issue for the two PRs you mentioned.

sbulen:
> Actually, this is trivial to take care of with the new installer and upgrader system. [4305a83](https://github.com/SimpleMachines/SMF/commit/4305a835d29121156402d9a2132da7281094cc10) has fixed the issue.

Hmmm...  So, any updates made in point releases going forward, must be replicated in SMF\Migration\v2_1\ and also SMF\Migration\v3_0\?  

What about non-index updates, e.g., field lengths or new columns, etc.?

Sesquipedalian:
> > Actually, this is trivial to take care of with the new installer and upgrader system. [4305a83](https://github.com/SimpleMachines/SMF/commit/4305a835d29121156402d9a2132da7281094cc10) has fixed the issue.
> 
> Hmmm... So, any updates made in point releases going forward, must be replicated in SMF\Migration\v2_1\ and also SMF\Migration\v3_0?

Not necessarily. If a particular migration step (ie. a database change) should be applied in more than one version of SMF, it can be listed in both appropriate sections of SMF\Maintenance\Tools\Upgrade::MIGRATIONS. In practice, though, it won't normally be necessary to list any given migration step more than once.

> What about non-index updates, e.g., field lengths or new columns, etc.?

Any changes related to table structure can be applied by using code similar to these examples:

```
		$table = new Schema\v3_0\LogSearchResults();
		$table->alterIndex($table->indexes['primary']);
```
```
		$table = new Schema\v3_0\Calendar();
		$table->alterColumn($table->columns['did']);
```

The first example will ensure that the primary key index of the log_search_results table conforms to the definition for it in the SMF\Db\Schema\v3_0\LogSearchResults class. The second example will ensure that the uid column of the calendar table conforms to the definition for it in the SMF\Db\Schema\v3_0\Calendar class.

In each case, the complete definition of the relevant column or index is given by an SMF\Db\Schema\Column or SMF\Db\Schema\DbIndex object. Calling the alterColumn or alterIndex method changes the existing column or index in the database so that all of its attributes match the definition given by the defining object.

This sort of thing is the whole point of the SMF\Db\Schema\* classes. 🙂

sbulen:
> Not necessarily. If a particular migration step (ie. a database change) should be applied in more than one version of SMF, it can be listed in both appropriate sections of SMF\Maintenance\Tools\Upgrade::MIGRATIONS. In practice, though, it won't normally be necessary to list any given migration step more than once.

I don't think I follow.  Why was it needed in two places in only this instance?  *Any* tweaks in 2.1 will cause a similar tweak in 3.0, correct?

Most db changes apply to a specific release & all subsequent releases when you make a change...

If the approach is "take me to THIS structure", as it sounds, then shouldn't it check all tables, columns & indexes?  Especially for major releases?  

That would make more sense to me than having to pick & choose what might need to be checked in subsequent releases.

The old upgrader, clunky & problematic as it was, would deal with any tweaks, at any time, and allow you to migrate forward from any point release.  All you had to do was make one update to one .sql file.  I'm not yet comfortable we've matched that functionality...

Sesquipedalian:
> I don't think I follow. Why was it needed in two places in only this instance?

It wasn't. SMF\Migration\v3_0\IndexUpdates is only applied in the 2.1 → 3.0 phase.

> If the approach is "take me to THIS structure", as it sounds, then shouldn't it check all tables, columns & indexes? Especially for major releases?

Well, we could add a final migration step for each point release that simply walks through all tables in that version's schema and applies alterColumn or alterIndex for each column and index in each table. That would ensure that we were always dealing with a known table structure. However, it would also increase the likelihood of blocking errors during upgrade—specifically, in situations where a column has gotten out of whack and now has data in it that can't be coerced into the expected form.

sbulen:
> > I don't think I follow. Why was it needed in two places in only this instance?
> 
> It wasn't. SMF\Migration\v3_0\IndexUpdates is only applied in the 2.1 → 3.0 phase.

But it is needed in 2.1 also - its a 2.1 change.  

I'm thinking about 3.0 & 3.1 as well, when both are being worked on.  The change should really only live in one release.

sbulen:
> Well, we could add a final migration step for each point release that simply walks through all tables in that version's schema and applies alterColumn or alterIndex for each column and index in each table. That would ensure that we were always dealing with a known table structure. However, it would also increase the likelihood of blocking errors during upgrade—specifically, in situations where a column has gotten out of whack and now has data in it that can't be coerced into the expected form.

TBH, I like this idea.  Though it suggests a need for 'report only' changes vs 'required structural' changes.

And it would also need to factor in the DB changes SMF makes due to user setting changes, e.g., text to mediumtext, custom search index, reports, etc.  

Anyway, food for thought.  

I'll run a few more tests tonight.  Then I'm out of town for 8 days, coming back next weekend.

sbulen:
No luck with the installer...  

I copied down 3 files from /other: install.php, Settings.php & Settings_bak.php.  If I should be doing something different than that, let me know.

PHP 8.4.5, MySQL 8.4.4.

Also tried with PHP 8.4.5 & Postgresql 14, & got the same result.

image

sbulen:
Similar trying a 2.1 => 3.0 upgrade.

WSOD with:
File not readable: Settings.php

sbulen:
Just tried a 2.1 => 3.0 upgrade via CLI.  It completed OK, but several issues:

 - After upgrade, it took me a few attempts to login.  At times I saw the same symptoms as the 'no session' error (#8624): no categories or boards.  But I had no restrictions on cookies.
 - I'm missing the id_board index on smf_log_notify
 - I'm missing smf_version on smf_log_packages
 - I'm getting cron errors on entity decode, see pic below

image

sbulen:
Just ran the 2.1 => 3.0 upgrade via browser...  *All the same comments as above*, but a minor nit about the graphics - the weird two dark grey lines below the 'please be patient' line.  Note I did not ask to have details displayed.  I think this is the box for the details:
image

sbulen:
Just ran a 3.0 => 3.0 upgrade via CLI and this is the bottom part of the log:

> ...
> ...
>  +++ Converting "smf_topics" to utf8mb4... skipped.
>  +++ Converting "smf_user_alerts" to utf8mb4... skipped.
>  +++ Converting "smf_user_alerts_prefs" to utf8mb4... skipped.
>  +++ Converting "smf_user_drafts" to utf8mb4... skipped.
>  +++ Converting "smf_user_likes" to utf8mb4... skipped.
> 
> Warning: Undefined array key 72 in D:\wamp64\www\van2130\Sources\Maintenance\Utf8ConverterStep.php on line 667
> 
> Call Stack:
>     0.0005     394264   1. {main}() D:\wamp64\www\van2130\upgrade.php:0
>     0.0219    1417704   2. SMF\Maintenance\Maintenance->execute($type = 2) D:\wamp64\www\van2130\upgrade.php:24
>     1.3207    6775704   3. call_user_func:\{D:\wamp64\www\van2130\Sources\Maintenance\Maintenance.php:294($callback = [0 => class SMF\Maintenance\Utf8ConverterStep  private int $SMF\Maintenance\Step\}id = 5; private string $SMF\Maintenance\Step\}name = 'Convert to UTF-8'; private string $SMF\Maintenance\Step}title = 'Convert to UTF-8'; private array|string $SMF\Maintenance\Step function = ...; private string ${SMF\Maintenance\Steptemplate = 'convertUtf8'; private int $SMF\Maintenance\Step\}progress = 30; public array $supported_charsets = [...]; public string $lang_charset = 'UTF-8'; public array $charset_maps = [...] , 1 => 'convertDatabase']) D:\wamp64\www\van2130\Sources\Maintenance\Maintenance.php:294
>     1.3207    6775888   4. SMF\Maintenance\Utf8ConverterStep->convertDatabase() D:\wamp64\www\van2130\Sources\Maintenance\Maintenance.php:294
> 
> 
> Warning: Attempt to read property "name" on null in D:\wamp64\www\van2130\Sources\Maintenance\Utf8ConverterStep.php on line 667
> 
> Call Stack:
>     0.0005     394264   1. {main}() D:\wamp64\www\van2130\upgrade.php:0
>     0.0219    1417704   2. SMF\Maintenance\Maintenance->execute($type = 2) D:\wamp64\www\van2130\upgrade.php:24
>     1.3207    6775704   3. call_user_func:D:\wamp64\www\van2130\Sources\Maintenance\Maintenance.php:294($callback = [0 => class SMF\Maintenance\Utf8ConverterStep private int ${SMF\Maintenance\Stepid = 5; private string $SMF\Maintenance\Step}name = 'Convert to UTF-8'; private string $SMF\Maintenance\Steptitle = 'Convert to UTF-8'; private array|string $SMF\Maintenance\Step}function = ...; private string $SMF\Maintenance\Steptemplate = 'convertUtf8'; private int $SMF\Maintenance\Stepprogress = 30; public array $supported_charsets = [...]; public string $lang_charset = 'UTF-8'; public array $charset_maps = [...] }, 1 => 'convertDatabase']) D:\wamp64\www\van2130\Sources\Maintenance\Maintenance.php:294
>     1.3207    6775888   4. SMF\Maintenance\Utf8ConverterStep->convertDatabase() D:\wamp64\www\van2130\Sources\Maintenance\Maintenance.php:294
> 
> Step 7: Finalize Upgrade
> Upgrade complete!
>

sbulen:
That's enough for tonight...  See you in a week.

Your meditation exercise for the week: contemplate a 2.1.2 => 3.1.3 upgrade.  Will the id_board index be there?

And when working on 3.1, do the 3.0 & 3.1 migration release branches for 3.0 need to be kept in sync?  How?

sbulen:
One last question: Where's the log?

Ah...  Saw your note above, it's only around when it fails...  

Hmmm...  I'm thinking support board questions...  Not sure I'd want it deleted.  Might want to consult @Kindred-999 ???

sbulen:
A thought...

If you had migration folders for the point releases, you're probably golden...   Just step through each.

In 2.1, you started saving the version in the settings table that way - e.g., "2.1.4".  So you know the precise starting point.

We'd need a 2.1.5 one for the two changes in question.

And that way, the exact same process could be used in point releases as in the upgrader.   All locked in step.  The only difference is the starting point.

Sesquipedalian:
> But it is needed in 2.1 also - its a 2.1 change.
> 
> I'm thinking about 3.0 & 3.1 as well, when both are being worked on. The change should really only live in one release.

> Your meditation exercise for the week: contemplate a 2.1.2 => 3.1.3 upgrade. Will the id_board index be there?
> 
> And when working on 3.1, do the 3.0 & 3.1 migration release branches for 3.0 need to be kept in sync? How?

I have thought about these things. I respectfully suggest that you are thinking about them wrongly, my friend, and in two key ways.

First, you seem to be thinking as though the upgrader needs to replay the evolution of our database structure with a patch release level of granularity, as though we need to model 2.1.0 → 2.1.1 → 2.1.2 → 2.1.3 → 2.1.4 → 2.1.5 → ... → 3.0.0 in the upgrader. Because of this way of thinking about it, you think that 2.1.5's fix to the log_notify table's primary index needs to be handled in an upgrade step that happens *before* the 2.1 → 3.0 upgrade.

However, the upgrader actually replays the evolution of the database structure with a point release level of granularity, e.g. 2.0.x → 2.1.x, or 2.1.x → 3.0.x. That means a database change introduced in a patch release needs to be reflected in two places: in a 2.1.x → 2.1.x upgrade (i.e. rerunning the upgrader), and in a 2.1.x → 3.0.x upgrade.  

That brings us to the second point. You seem to be thinking of the upgrader as though it were an un-versioned, universal thing that applies to all SMF versions equally—we just tack on new sets of changes with each new version of SMF. 

But that's not the case. Each version of SMF has a corresponding version of the upgrader, and these versions are indeed distinct. Specifically, only the 2.1 version of the upgrader will ever perform a 2.1.x → 2.1.x upgrade, while only the 3.0 version will ever perform a 3.0.x → 3.0.x upgrade, etc. This difference is why the change to the database in 2.1.5 needs to be reflected in two different places. In the 2.1 upgrader, it needs to be reflected in the 2.1.x → 2.1.x steps. But in the 3.0 upgrader (and beyond), it needs to be reflected in the 2.1.x → 3.0.x steps, because no 2.1.x → 2.1.x steps are or could ever be performed by the 3.0 upgrader.

Sesquipedalian:
> And when working on 3.1, do the 3.0 & 3.1 migration release branches for 3.0 need to be kept in sync? How?

This, like everything else, is managed by Git. As I said in my previous reply, the upgrader is, in fact, versioned along with SMF as a whole and as part of SMF as a whole. When 3.1 is forked from 3.0, it will inherit the migration steps from the 3.0 branch. As new database changes are introduced in 3.1, migration steps will be added for them in the 3.1 branch. If we ever find ourselves needing to introduce a database change in a patch to 3.0, a migration step for that will end up being added in both the 3.0 branch as part of the 3.0x → 3.0.x steps and, separately, in the 3.1 branch as part of the 3.0.x → 3.1.x steps. That's all there is to it.

Sesquipedalian:
> * After upgrade, it took me a few attempts to login.

That's actually not the upgrader's fault. It's due to the changes introduced in #8464 (specifically, in fed16cc). At the time when I was working on that PR, I was flipping back and forth so much between the old and new password hash values that I didn't notice that it would take two tries to log in successfully once the change was made. We need to make some more changes— probably in SMF\Actions\Login2—in order to handle the change seamlessly.

Sesquipedalian:
> * I'm missing the id_board index on smf_log_notify
> * I'm missing smf_version on smf_log_packages

Both should be fixed with the latest commits.

Sesquipedalian:
> One last question: Where's the log?
> 
> Ah... Saw your note above, it's only around when it fails...
> 
> Hmmm... I'm thinking support board questions... Not sure I'd want it deleted. Might want to consult @Kindred-999 ???

Log files are now saved in a secured ./logs directory. If that directory can't be created or can't be secured, then the log file is deleted after the tool completes its process.

Sesquipedalian:
> * I'm getting cron errors on entity decode, see pic below

[b2a365f](https://github.com/SimpleMachines/SMF/pull/8652/commits/b2a365f05e51009c344d9f01f3f5191e5409998c) should fix that.

> Just ran a 3.0 => 3.0 upgrade via CLI and this is the bottom part of the log:
> 
> > Warning: Undefined array key 72 in D:\wamp64\www\van2130\Sources\Maintenance\Utf8ConverterStep.php on line 667

[54c268a](https://github.com/SimpleMachines/SMF/pull/8652/commits/54c268a20a1bbac1ff8754785980da5cde9d337a) should fix that.

Sesquipedalian:
> No luck with the installer...
> 
> I copied down 3 files from /other: install.php, Settings.php & Settings_bak.php. If I should be doing something different than that, let me know.
> 
> PHP 8.4.5, MySQL 8.4.4.
> 
> Also tried with PHP 8.4.5 & Postgresql 14, & got the same result.
> 
> image

Hm. Weird. I'll need to experiment to see if I can reproduce that and figure out a cause.

Sesquipedalian:
Latest commits should fix all reported issues with the installer (and several unreported ones, too).

Sesquipedalian:
> > Well, we could add a final migration step for each point release that simply walks through all tables in that version's schema and applies alterColumn or alterIndex for each column and index in each table. That would ensure that we were always dealing with a known table structure. However, it would also increase the likelihood of blocking errors during upgrade—specifically, in situations where a column has gotten out of whack and now has data in it that can't be coerced into the expected form.
> 
> TBH, I like this idea.

I've now implemented it. Doing so was quite salutary. I found and fixed a lot of small issues that would have likely continued to go unnoticed otherwise.

jdarwood007:
For the fix on #8648 is it only the first part or the second part of the error?
The second part was the complex fix.

Sesquipedalian:
I changed things so that the part where we try to set the default row type to dynamic is now optional and will be silently skipped if it cannot be performed. This is because setting that default is nice, but not absolutely necessary.

sbulen:
> I changed things so that the part where we try to set the default row type to dynamic is now optional and will be silently skipped if it cannot be performed. This is because setting that default is nice, but not absolutely necessary.

Why is it no longer necessary?  Did we change the structure so there are no longer index length issues?  I.e., SMF early InnoDB adopters, with older row formats, are still going to fail on 1071 "Specified key is too long..." errors on the mb4 conversion.

Sesquipedalian:
Using dynamic rows is mandatory, but seeing that default is optional. We explicitly set the rows to dynamic when creating or updating the table. Setting that default is a good idea if we can, just in case someone creates a table without using our API methods, but we can live without it.

jdarwood007:
What about the second part with the function and super privs?

I am honestly not sure how much we would see that in a shared env to know how common it will be.  But I know having super is a deal breaker for most if not all, shared envs.

Sesquipedalian:
> What about the second part with the function and super privs?
> 
> I am honestly not sure how much we would see that in a shared env to know how common it will be. But I know having super is a deal breaker for most if not all, shared envs.

That's what I addressed above. Setting that default is what requires those high level privileges. We try to set the default if we can, but if that fails we just silently ignore it and move on.

jdarwood007:
Screen Shot 2025-06-05 at 7 05 23 PM

Odd that this happens.  I was on another 3.0 Alpha.  But it should restart from 3.0 alpha upgrades at least.
Also note the nans that occurred

> PHP: 8.3.21 | Real Time: 0.166 | Timer: DS: 0 | Engine: mysql | Git: [Branch:"pr/8652" | Revision:"c46cf5b" - Modified Files]

sbulen:
Is this OK to test again?

Sesquipedalian:
Yes, although I haven't yet addressed the bug in @jdarwood007's last comment. ~(It's a simple division by zero problem that just needs some `if` checks to test for when there are no database changes to make, so it won't be hard to fix. But I haven't done that yet, so you can expect to run into it.)~

EDIT: I was wrong. It was a  bit more involved than that.

Sesquipedalian:
Latest commits do two things:

1. Merge it the recent changes from the main branch, including the change to the parameters for Db::$db->query().
2. Fix the issue @jdarwood007 reported [above](https://github.com/SimpleMachines/SMF/pull/8652#issuecomment-2947809535) where re-running the upgrader would not perform the expected steps for a 3.0 → 3.0 upgrade and would instead show a broken UI with NANs and such.

jdarwood007:
Sent a patch for a identifier up for you.

Currently getting this
```
Ensuring "smf_moderators" is structured correctly... done.

Critical Error!
Invalid use of NULL value 
```

Which, points out a good case where we need the UI to update that properly.
The log shows it  like this:
```
 +++ Ensuring "smf_package_servers" is structured correctly... 
```

Here's a `show create table`
```
CREATE TABLE `smf_package_servers` (
  `id_server` smallint unsigned NOT NULL AUTO_INCREMENT,
  `name` varchar(255) NOT NULL DEFAULT '',
  `url` varchar(255) NOT NULL DEFAULT '',
  `validation_url` varchar(255) NOT NULL DEFAULT '',
  `extra` mediumtext,
  PRIMARY KEY (`id_server`)
) ENGINE=InnoDB AUTO_INCREMENT=3 DEFAULT CHARSET=utf8mb3 ROW_FORMAT=DYNAMIC
```

sbulen:
Not getting very far on an install or upgrade with the current code (MySQL 8.4.4, PHP 8.4.5).  

On install, tables get created, but they are empty, also admin account is not created & app won't load.  The upgrader has similar errors.

Getting a bunch of these:
```
Warning
: Undefined variable $table in
D:\wamp64\www\84van30\Sources\Db\Schema\Table.php
on line
440

Warning
: Attempt to read property "columns" on null in
D:\wamp64\www\84van30\Sources\Db\Schema\Table.php
on line
440

Warning
: foreach() argument must be of type array|object, null given in
D:\wamp64\www\84van30\Sources\Db\Schema\Table.php
on line
440

Warning
: Undefined variable $table in
D:\wamp64\www\84van30\Sources\Db\Schema\Table.php
on line
440
```

Underneath the above errors, the installer displays this:
```
Your settings have now been saved and the database has been populated with all the data required to get your forum up and running. Summary of population:

    Created 72 tables.

Some of the queries were not executed properly. This could be caused by an unsupported (development or old) version of your database software.

Technical information about the queries:

    admin_info_files:Class "SMF\Db\Schema\Lang" not found
    board_permissions:Class "SMF\Db\Schema\Lang" not found
    board_permissions_view:Class "SMF\Db\Schema\Lang" not found
    boards:Class "SMF\Db\Schema\Lang" not found
    calendar:Class "SMF\Db\Schema\Lang" not found
    categories:Class "SMF\Db\Schema\Lang" not found
    custom_fields:Class "SMF\Db\Schema\Lang" not found
    membergroups:Class "SMF\Db\Schema\Lang" not found
    message_icons:Class "SMF\Db\Schema\Lang" not found
    messages:Class "SMF\Db\Schema\Lang" not found
    package_servers:Class "SMF\Db\Schema\Lang" not found
    permission_profiles:Class "SMF\Db\Schema\Lang" not found
    permissions:Class "SMF\Db\Schema\Lang" not found
    scheduled_tasks:Class "SMF\Db\Schema\Lang" not found
    settings:Class "SMF\Db\Schema\Lang" not found
    smiley_files:Class "SMF\Db\Schema\Lang" not found
    smileys:Class "SMF\Db\Schema\Lang" not found
    spiders:Class "SMF\Db\Schema\Lang" not found
    themes:Class "SMF\Db\Schema\Lang" not found
    topics:Class "SMF\Db\Schema\Lang" not found
    user_alerts_prefs:Class "SMF\Db\Schema\Lang" not found

Click "Continue" to progress to the admin account creation page.
```
And if I press Continue, or otherwise try to load the new forum, I get this:
```
Warning: Undefined array key "attachmentUploadDir" in D:\wamp64\www\84van30\Sources\Config.php on line 1054

Warning: Undefined array key "attachmentUploadDir" in D:\wamp64\www\84van30\Sources\Config.php on line 1055

Fatal error: Uncaught TypeError: SMF\Utils::jsonDecode(): Argument #1 ($json) must be of type string, null given, called in D:\wamp64\www\84van30\Sources\Config.php on line 1055 and defined in D:\wamp64\www\84van30\Sources\Utils.php:1474 Stack trace: #0 D:\wamp64\www\84van30\Sources\Config.php(1055): SMF\Utils::jsonDecode(NULL, true, 512, 0, false) #1 D:\wamp64\www\84van30\Sources\Maintenance\Tools\Install.php(867): SMF\Config::reloadModSettings() #2 [internal function]: SMF\Maintenance\Tools\Install->adminAccount() #3 D:\wamp64\www\84van30\Sources\Maintenance\Maintenance.php(294): call_user_func(Array) #4 D:\wamp64\www\84van30\install.php(31): SMF\Maintenance\Maintenance->execute(1) #5 {main} thrown in D:\wamp64\www\84van30\Sources\Utils.php on line 1474
```

Sesquipedalian:
> Warning
> : Undefined variable $table in
> D:\wamp64\www\84van30\Sources\Db\Schema\Table.php
> on line
> 440

Derp. Fixed now.

Oldiesmann:
The "SMF\Db\Schema\Lang not found" errors indicate we're trying to call methods in the SMF\Lang class while in the SMF\Db\Schema namespace. Should be an easy fix though.

Sesquipedalian:
> The "SMF\Db\Schema\Lang not found" errors indicate we're trying to call methods in the SMF\Lang class while in the SMF\Db\Schema namespace. Should be an easy fix though.

Fixed now.

jdarwood007:
Why are we deleting files?  The purpose was to leave them around.

Sesquipedalian:
We are only deleting the ones that will never be needed again. For example, once the upgrade to 3.0 has been completed, we will never need the migration steps for 2.0 → 2.1 again, nor the database schema files for 2.1. Conversely, we do keep the schema files and migration steps for 3.0, because we might very well use them again.

jdarwood007:
This PR should address #8683 as well

Sesquipedalian:
Good idea

jdarwood007:
I found one more, #8241 makes sense to handle here.  Not for 2.1, but something we can do here by updating the error message to include it.  I am not seeing any other issue for installer that isn't linked to this PR.

jdarwood007:
I also see #7615 
But I think that is resolved with the rewrite in the first place.

I also see #7551 
I need to check on that one. Maybe related to #8241

Sesquipedalian:
I'm going to go ahead and merge this now. If any new issues with it arise afterward, we can fix them then.

PR #8664, opened 2025-06-09
Optional identifier

jdarwood007:


jdarwood007:
Thanks for the catch.  It looks like we have some slight variance in how we called `Db::$db->query`.

PR #8675, opened 2025-06-16
PHP 8.4 compatibility

### LINKED ISSUE #8676: [2.1]: PHP 8.4 compatibility
PHP 8.4 deprecates implicit nullable types, meaning that a function signature like this will thow deprecation notices:

```
function my_func(int $foo = null)
{
    ...
}
```

Instead, PHP 8.4 requires the `int` type declaration to be replaced with `?int` or `int|null`. However, neither  `?int` nor `int|null` are supported by PHP 7.0, which means that the only way to support implicitly nullable types in a way that is compatible with both PHP 7.0 and PHP 8.4 is to not use type declarations at all.

This breaking change affects the ReCaptcha library we use, and Google has not yet updated the library to fix this. 

Moreover, unless Google decides to update the library in a way that stops using implicitly nullable types entirely, it will be impossible for them (or anyone) to add support for PHP 8.4 without dropping support for PHP 7.0.

In light of this, I see three possible responses:

1. We tweak our copy of the included library to make it compatible with PHP 8.4 until such time as Google updates the official repository with a fix. This will require dropping support for PHP 7.0.
2. We replace the included library with some other one. This would also require other changes elsewhere in our code, and would require extensive testing. This might or might not require dropping support for PHP 7.0.
3. We decide not to support PHP 8.4.

I don't think it makes any sense to keep support for PHP 7.0 at the cost of not supporting 8.4, so option three is off the table. 

Option 2 might be worth looking into, but it would take time to test and I want to resolve this in 2.1.5. 

Option 1, then, seems like the best route forward to me. I don't like the idea of having to make changes to a third party library, but that seems like the lesser evil.

Sesquipedalian:
Updates the PHP versions that we lint against.

Fixes #8676. In the process, this also raises our minimum supported version of PHP from 7.0 to PHP 7.1

If and when Google updates the official ReCaptcha library to make it compatible with PHP 8.4, we can drop it in.

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -41,7 +41,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php: [ 7.0, 7.1, 7.2, 7.3, 7.4, 8.0, 8.1, 8.2, 8.3, 8.4 ]
+        php: [ 7.1, 7.2, 7.3, 7.4, 8.0, 8.1, 8.2, 8.3, 8.4 ]
 
     name: PHP ${{ matrix.php }} Syntax Check
     steps:
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -41,7 +41,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php: [ 7.0, 7.1, 7.2, 7.3, 7.4, 8.0, 8.1 ]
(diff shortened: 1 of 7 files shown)


jdarwood007 (approved):
Option 1 is our best option and I agree its a really old PHP version so we should not worry.

jdarwood007:
@illori Be aware of these changes for when 2.1.5 is released that it updates minimum requirements and the docs will need to be updated to reflect that.  I updated other/requirements.md

PR #8697, opened 2025-06-26
[3.0] Overhaul User IP handling

### LINKED ISSUE #7684: Proxy headers don't work properly
#### Description
Proxy detection doesn't actually work properly

### Steps to reproduce
1. Have a proxy
2. Observe the IP address isn't respected properly

- Version/Git revision: 2.1.3
- Database Type: n/a
- Database Version: n/a
- PHP Version: n/a

Bug report: https://www.simplemachines.org/community/index.php?topic=585020.0
Related PR: https://github.com/SimpleMachines/SMF/pull/7368

The issue is that $_SERVER['REMOTE_ADDR'] isn't actually updated correctly to be used (something the PR sort of handles, but incorrectly)

The intent should be that the real IP as per the proxy headers gets dumped into $_SERVER['REMOTE_ADDR'] (and thus into $user_info['ip']), with the next IP up the chain (if present, that *isn't* the proxy server) into $_SERVER['BAN_CHECK_IP'] (and thus into $user_info['ip2']) as the fallback ban check so that proxies that aren't expected can be banned if necessary.

jdarwood007:
Fixes #7684

This PR addresses the above issue but also isolates the handling of the IP determination into a dedicated function as well as removing all places where we are changing the $_SERVER['REMOTE_ADDR'] or setting $_SERVER['BAN_CHECK_IP'], except if we are in compatibility mode.  The later purpose is to have SMF stop modifying variables on the server, leaving them as was and cleaning them to its standards and using that when needed.

This fixes the PR by flipping how we have done the checks.  In the past we have determined the user being passed through the reverse proxy and setting that users real IP into BAN_CHECK_IP, leaving REMOTE_ADDR as the proxy.  We then proceeded to check and use most of the time REMOTE_ADDR.  By flipping this, we get the users real IP and use that and use the alternative IP as the original requesting IP (reverse proxy).


Sesquipedalian (changes requested):
Two things:

1. Needs to be updated to work with the new installer and upgrader.
2. I think it would be more appropriate to put these new static methods and properties in SMF\IP rather than SMF\QueryString.

jdarwood007:
In reality, the user IP has little to do with IP, but we don't have a logical place yet for all input ($_GET/$_POST/$FILES/etc).

Sources/Actions/Login2.php
```diff
@@ -793,7 +794,7 @@ protected function DoLogin(): void
 		}
 
 		// You've logged in, haven't you?
-		$update = ['member_ip' => User::$me->ip, 'member_ip2' => $_SERVER['BAN_CHECK_IP']];
+		$update = ['member_ip' => User::$me->ip, 'member_ip2' => QueryString::getUserIPAlternative()];
```
Sesquipedalian:
```suggestion
		$update = ['member_ip' => User::$me->ip, 'member_ip2' => IP::getUserIPAlternative()];
```

Sources/Actions/Login2.php
```diff
@@ -24,6 +24,7 @@
 use SMF\ErrorHandler;
 use SMF\IntegrationHook;
 use SMF\Lang;
+use SMF\QueryString;
```
Sesquipedalian:
```suggestion
use SMF\IP;
```

This line will also need to be moved up in order to keep PHP-CS-Fixer happy.

Sources/Actions/Register2.php
```diff
@@ -607,7 +607,7 @@ public static function registerMember(array &$reg_options, bool $return_errors =
 			'posts' => 0,
 			'date_registered' => time(),
 			'member_ip' => $reg_options['interface'] == 'admin' ? '127.0.0.1' : User::$me->ip,
-			'member_ip2' => $reg_options['interface'] == 'admin' ? '127.0.0.1' : $_SERVER['BAN_CHECK_IP'],
+			'member_ip2' => $reg_options['interface'] == 'admin' ? '127.0.0.1' : IP::getUserIPAlternative(),
```
Sesquipedalian, line 611: I believe this requires a `use SMF\IP;` statement to be added to the top of the file.

Sources/ErrorHandler.php
```diff
@@ -275,7 +275,7 @@ public static function log(string $error_message, string|bool $error_type = 'gen
 		$error_info = [
 			User::$me->id ?? User::$my_id ?? 0,
 			time(),
-			User::$me->ip ?? $_SERVER['REMOTE_ADDR'] ?? '',
+			User::$me->ip ?? QueryString::getUserIP(),
```
Sesquipedalian:
```suggestion
			User::$me->ip ?? IP::getUserIP(),
```

Sesquipedalian (changes requested):
I've noted a few spots that still need changes. 

Beyond that, I'm wondering whether any of the changes to the installer and upgrader that were part of the initial commit in this PR will need to be reflected in the new installer and upgrader. Perhaps they don't, but I figured it was worthwhile to double check.

Sesquipedalian:
> In reality, the user IP has little to do with IP, but we don't have a logical place yet for all input ($_GET/$_POST/$FILES/etc).

You're not wrong. Perhaps we should have a general SMF\InputProcessor class to handle all input. Then we could move not only these methods and properties to it, but also a bunch of other stuff in SMF\QueryString that really has little to do with query strings. But such a change would certainly need to be a different PR, so nothing to worry about now.

Sources/Actions/Register2.php
jdarwood007: Added

jdarwood007:
> > In reality, the user IP has little to do with IP, but we don't have a logical place yet for all input ($_GET/$_POST/$FILES/etc).
> 
> You're not wrong. Perhaps we should have a general SMF\InputProcessor class to handle all input. Then we could move not only these methods and properties to it, but also a bunch of other stuff in SMF\QueryString that really has little to do with query strings. But such a change would certainly need to be a different PR, so nothing to worry about now.

If we do that, we are going to talk about a major change.  Since we should use that InputProcessor to handle all input and no longer touch anything in any superglobals.  The goal is to get there in the future.  The best part about doing that is we will only sanitize data when requested, so instead of the entire superglobals being cleansed, we would do it when requested the input.  The problem is, so many places in our code touch and update the superglobals that it requires refactoring the logic to use local variables and then for other places that assumed the updated superglobal was already done need to be fixed.  Then, doing something for backwards compatibility means we need to read from PHP://stdin to ensure we don't read from the modified superglobal when we pollute them.

Sesquipedalian:
Very good points. Perhaps a task for 3.1, then.

Sesquipedalian (changes requested):
Needs to be updated for compatibility with the new installer and upgrader

jdarwood007:
I can't figure out where the changes to the installer/upgrader come from.  None of my commits did that...  Its clear as day if I diff between the two branches, but when I just look at the history above, none of them shows the installer/ugprader commit that reinstates the entire thing.

PR #8704, opened 2025-06-26
[3.0] Issue 8134

arlomedia:
Simplify the session lifetime calculations and always use the setting from the user interface.

From https://github.com/SimpleMachines/SMF/issues/8134

### DIFF
--- a/Sources/Session.php
+++ b/Sources/Session.php
@@ -227,7 +227,7 @@ class Session implements \SessionHandlerInterface
 
 				@ini_set('session.gc_probability', '1');
 			}
-			@ini_set('session.gc_maxlifetime', max($modSettings['databaseSession_lifetime'], 60));
+			@ini_set('session.gc_maxlifetime', max(Config::$modSettings['databaseSession_lifetime'], 60));
 
 			// Use cache setting sessions?
 			if (empty(Config::$modSettings['databaseSession_enable']) && !empty(CacheApi::$enable) && php_sapi_name() != 'cli') {
--- a/Sources/Session.php
+++ b/Sources/Session.php
@@ -158,11 +158,6 @@ class Session implements \SessionHandlerInterface
 	 */
 	public function gc(int $max_lifetime): int|false
 	{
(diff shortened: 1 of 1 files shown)


Sources/Session.php
```diff
@@ -231,9 +226,8 @@ public static function load(): void
 				session_set_save_handler(new self(), true);
 
 				@ini_set('session.gc_probability', '1');
-			} elseif (ini_get('session.gc_maxlifetime') <= 1440 && !empty(Config::$modSettings['databaseSession_lifetime'])) {
-				@ini_set('session.gc_maxlifetime', max(Config::$modSettings['databaseSession_lifetime'], 60));
 			}
+			@ini_set('session.gc_maxlifetime', max($modSettings['databaseSession_lifetime'], 60));
```
```suggestion
			@ini_set('session.gc_maxlifetime', max(Config::$modSettings['databaseSession_lifetime'], 60));
```

jdarwood007:
I would prefer to see it like this, anyone concur?
```
@ini_set('session.gc_maxlifetime', max(min(Config::$modSettings['databaseSession_lifetime'], Config::$modSettings['databaseSession_lifetime']), 60));
```

Get the lowest of the server or SMF settings and ensure they have a sane value of at least 60.

arlomedia:
This seems like a mistake: `min(Config::$modSettings['databaseSession_lifetime'], Config::$modSettings['databaseSession_lifetime'])`. It sounds like you meant to write `min(ini_get('session.gc_maxlifetime'), Config::$modSettings['databaseSession_lifetime'])`.

How about just `max(Config::$modSettings['databaseSession_lifetime'], 60)`? Then users could set it higher or lower than the server default setting.

jdarwood007:
Ooops, I did typo that, but yes:
```
@ini_set('session.gc_maxlifetime', min(ini_get('session.gc_maxlifetime'), Config::$modSettings['databaseSession_lifetime']), 60));
```

isaak654:
Aside from editing the title, if this pull request cannot be repurposed for branch 2.1, it may be preferable to submit a new one.

PR #8709, opened 2025-06-27
[3.0] Use a anonymous function for db callbacks

jdarwood007:
Just an idea I had while reading the code for another bug I was looking at. Doesn't really change much programmatically, but aesthetically looks better and easier to follow a debugger through.

If we don't want it, close the PR.


Sources/Db/APIs/MySQL.php
```diff
@@ -282,14 +266,10 @@ public function quote(string $db_string, array $db_values, ?object $connection =
...
-			$this->temp_values = $db_values;
-			$this->temp_connection = $connection ?? $this->connection;
-
 			// Do the quoting and escaping
-			$db_string = preg_replace_callback('~{([a-z_]+)(?::([a-zA-Z0-9_-]+))?}~', [$this, 'replacement__callback'], $db_string);
-
-			unset($this->temp_values, $this->temp_connection);
+			$db_string = preg_replace_callback('~{([a-z_]+)(?::([a-zA-Z0-9_-]+))?}~', function ($matches) use ($db_values, $connection) {
+				return $this->replacement__callback($matches, $db_values, $connection);
+			}, $db_string);
```
```suggestion
			$db_string = preg_replace_callback(
				'~{([a-z_]+)(?::([a-zA-Z0-9_-]+))?}~', 
				fn ($matches) => $this->replacement__callback($matches, $db_values, $connection ?? $this->connection), 
				$db_string,
			);
```

Sources/Db/APIs/MySQL.php
```diff
@@ -2261,11 +2241,15 @@ protected function initiate(string $user, string $passwd, array $options = []):
...
 	 */
-	protected function replacement__callback(array $matches): string
+	protected function replacement__callback(array $matches, array $db_values, ?object $connection = null): string
 	{
-		if (!is_object($this->temp_connection)) {
+		$connection ??= $this->connection;
+
+		if (!is_object($connection)) {
 			ErrorHandler::displayDbError();
 		}
```
```suggestion
	 * @param array $db_values The values to be inserted into the string.
	 * @param object $connection The connection to use.
	 * @return string The appropriate string depending on $matches[1].
	 */
	protected function replacement__callback(array $matches, array $db_values, object $connection): string
{
```

Since this method is only called internally, we know that we will always pass a non-null $connection value.

Sources/Db/APIs/PostgreSQL.php
```diff
@@ -317,14 +317,10 @@ public function quote(string $db_string, array $db_values, ?object $connection =
...
-			$this->temp_values = $db_values;
-			$this->temp_connection = $connection ?? $this->connection;
-
 			// Do the quoting and escaping
-			$db_string = preg_replace_callback('~{([a-z_]+)(?::([a-zA-Z0-9_-]+))?}~', [$this, 'replacement__callback'], $db_string);
-
-			unset($this->temp_values, $this->temp_connection);
+			$db_string = preg_replace_callback('~{([a-z_]+)(?::([a-zA-Z0-9_-]+))?}~', function ($matches) use ($db_values, $connection) {
+				return $this->replacement__callback($matches, $db_values, $connection);
+			}, $db_string);
```
Sesquipedalian:
```suggestion
			$db_string = preg_replace_callback(
				'~{([a-z_]+)(?::([a-zA-Z0-9_-]+))?}~', 
				fn ($matches) => $this->replacement__callback($matches, $db_values, $connection ?? $this->connection), 
				$db_string,
			);
```

Sources/Db/APIs/PostgreSQL.php
```diff
@@ -2258,11 +2254,15 @@ function ($errno, $errstr) {
...
+	 * @param null|object $connection = null The connection to use (null to use $db_connection)
 	 * @return string The appropriate string depending on $matches[1]
 	 */
-	protected function replacement__callback(array $matches): string
+	protected function replacement__callback(array $matches, array $db_values, ?object $connection = null): string
 	{
-		if (!is_object($this->temp_connection)) {
+		$connection ??= $this->connection;
+
+		if (!is_object($connection)) {
```
Sesquipedalian:
```suggestion
	 * @param array $db_values The values to be inserted into the string.
	 * @param object $connection The connection to use.
	 * @return string The appropriate string depending on $matches[1].
	 */
	protected function replacement__callback(array $matches, array $db_values, object $connection): string
{
```

Since this method is only called internally, we know that we will always pass a non-null $connection value.

Sesquipedalian (changes requested):
In addition to the suggested changes, conflicts need to be resolved.

Sources/Db/APIs/PostgreSQL.php
Sesquipedalian: Oops. Don't just merge this suggestion. I meant to include the following three lines in the selection for the suggestion. They should be removed along with the lines that are removed in this suggestion.

jdarwood007:
I totally forgot that `fn()` exists in the scope of the function that called it and thus doesn't need use.  Much better.

Sesquipedalian (approved):

PR #8710, opened 2025-06-28
Exit out of thumbnail creation if it fails

### LINKED ISSUE #8649: [3.0]: .tiff issue - cannot assign false to property
### Basic Information

If you attempt to attach a .tiff in 3.0, you get a "cannot assign false to property" error.

Image

### Steps to reproduce

New install, all default settings, however sizes increased to allow for larger attachments.

1.  Attach a .tiff

### Expected result

Attached .tiff

### Actual result

If you ignore the error, you can actually create the post and the attachment record is created.  

However...  No thumbnail, no attachment record created for the thumbnail.

Similar issues with thumbnails on .tiffs exist in 2.1: #8650 

### Version/Git revision

3.0 alpha 3

### Database Engine

All

### Database Version

8.4.0

### PHP Version

8.4.5

### Logs

```shell

```

### Additional Information

_No response_

jdarwood007:
Fixes #8649

The issue found in #8649 is realpath() is returning false and attempting to assign that to $this->source, which is not allowed.  Both GD and ImageMagick failed to create a thumbnail.

### DIFF
--- a/Sources/Attachment.php
+++ b/Sources/Attachment.php
@@ -2073,7 +2073,7 @@ class Attachment implements \ArrayAccess
 
 						$image = new Image($filename, true);
 
-						if (!empty($image->source) && ($thumb = $image->createThumbnail(Config::$modSettings['attachmentThumbWidth'], Config::$modSettings['attachmentThumbHeight'])) !== false) {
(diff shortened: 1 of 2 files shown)


Oldiesmann:
Might want to check if the format is even supported before we try. I had no issue with a test .tiff image using IMagick on my server though.

Sesquipedalian (approved):

PR #8716, opened 2025-07-01
[2.1] Fix mail queue

### LINKED ISSUE #8712: [2.1]: Mail getting stuck in queue after upgrading
### Basic Information

After upgrading from SMF 2.1.4 to 2.1.5 and then to 2.1.6 mail is getting stuck in the queue (not sending) , and when selecting "Send Mail Queue Now" nothing at all seems to happen. There are no errors in the error log. The method I am using is SMTP - STARTTLS . Before upgrading everything worked as it expected.
P.S. Sending a test mail works fine.
P.P.S. I don't know if this started to happen with 2.1.5 or 2.1.6 as I wasn't using 2.1.5 long enough to notice.

### Steps to reproduce

1. E.G. register an account to receive an activation email, request a password reset link, send registration reminder mail, etc.

### Expected result

Mail should be sent from the queue

### Actual result

Mail gets sent to the queue and remains stuck there (unless removed)

### Version/Git revision

2.1.6

### Database Engine

All

### Database Version

_No response_

### PHP Version

8.1

### Logs

```shell

```

### Additional Information

_No response_

jdarwood007:
Fixes #8712
Needs some more testing, but this should be the resolution we are looking for.

### DIFF
--- a/Sources/ScheduledTasks.php
+++ b/Sources/ScheduledTasks.php
@@ -811,8 +811,10 @@ function ReduceMailQueue($number = false, $override_limit = false, $force_send =
 	{
 		// First, figure out when the next send attempt should happen based on the current priority.
 		$next_send_time = $email['time_sent'];
-		for ($i = 0; $i < $email['priority']; $i++) {
-			$next_send_time += 20 * max(0, $email['priority'] - $priority_offset);
+		if ($email['priority'] >= $priority_offset) {
+			for ($i = 0; $i < $email['priority']; $i++) {
+				$next_send_time += 20 * max(0, $email['priority'] - $priority_offset);
+			}
 		}
 
 		// If the email is too old, discard it.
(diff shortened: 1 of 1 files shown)


jdarwood007:
@sbulen , @Oldiesmann , @beckzyb,

Confirm that even with this, we are looking good. I am monitoring a mail queue now, and it seems to be running ok with these changes for the past few hours.

Oldiesmann:
That fixed it. Applied patch, then went in and sent the queue and that cleared it out. I also got the email that was waiting for me.

beckzyb:
This seems to be working great. As a test I've just sent reminder emails to all forum members who have been awaiting activation for more than 14 days, and the mail is leaving the queue at 10 per minute (which is what I have it set to) . All is good now!

sbulen:
I created one of my little modlets & have it installed on my forum.  I'll monitor.

qcybb:
This fixes the issue. The backlog in mail queue was sent out, and all new emails are being delivered within a minute.

PR #8740, opened 2025-07-02
[3.0] Ensure Atom title is marked as html

### LINKED ISSUE #8338: [2.1]: Atom feed `entry` `title` contains HTML but `type` is `text`
### Basic Information

The `type` of the `title` element in Atom defaults to `text`, so it should not contain HTML.  Either change the `type` to `html` (as is done with `summary`) or do not HTML-escape it.

### Steps to reproduce

1. Post topic with subject containing a quote (`"`)
2. [Validate](https://validator.w3.org/feed/) the Atom feed for the board

### Expected result

No validation issues.

### Actual result

"title should not contain HTML unless declared in the type attribute: &amp;quot;"

### Version/Git revision

2.1.4

### Database Engine

All

### Database Version

_No response_

### PHP Version

_No response_

### Logs

_No response_

### Additional Information

* [The "atom:title" Element](https://www.rfc-editor.org/rfc/rfc4287.html#section-4.2.14) is a text construct
* [Text Constructs](https://www.rfc-editor.org/rfc/rfc4287.html#section-3.1) have a `type`
* [The "type" Attribute](https://www.rfc-editor.org/rfc/rfc4287.html#section-3.1.1) defaults to `text`

jdarwood007:
Fixes #8338

Note that the Atom RFC (4287) also says that the ID should be a URI, so we can't use a UUID.

### DIFF
--- a/Sources/Actions/Feed.php
+++ b/Sources/Actions/Feed.php
@@ -923,6 +923,7 @@ class Feed implements ActionInterface, Routable
 						[
 							'tag' => 'title',
 							'content' => $row['subject'],
+							'attributes' => ['type' => 'html'],
 							'cdata' => true,
 						],
 						[
@@ -973,7 +974,7 @@ class Feed implements ActionInterface, Routable
 						],
 						[
 							'tag' => 'id',
-							'content' => $uuid,
+							'content' => Config::$scripturl . '?msg=' . $row['id_msg'],
 						],
 						[
 							'tag' => 'link',
@@ -1363,6 +1364,7 @@ class Feed implements ActionInterface, Routable
 						[
 							'tag' => 'title',
 							'content' => $row['subject'],
+							'attributes' => ['type' => 'html'],
 							'cdata' => true,
 						],
 						[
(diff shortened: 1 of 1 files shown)


Sources/Actions/Feed.php
```diff
@@ -973,7 +974,7 @@ function ($a, $b) {
 						],
 						[
 							'tag' => 'id',
-							'content' => $uuid,
+							'content' => Config::$scripturl . '?msg=' . $row['id_msg'],
```
Sesquipedalian:
```suggestion
							'content' => 'urn:uuid:' . $uuid,
```

Sources/Actions/Feed.php
```diff
@@ -1413,7 +1415,7 @@ function ($a, $b) {
...
```
Sesquipedalian:
```suggestion
							'content' => 'urn:uuid:' . $uuid,
```

Sesquipedalian (dismissed):
UUIDs can be turned into valid URIs by prepending `urn:uuid:` to them.

jdarwood007:
That works for me.  I saw it was failing spec on those and figured we would return to the URL.

PR #8749, opened 2025-07-03
[2.1] Prevents multiple task instances from clobbering Unicode files

### LINKED ISSUE #8713: [3.0]: Scheduled task issue
### Basic Information

If you request Weekly Maintenance twice, you can end up with a WSOD with this text:
> Fatal error: Uncaught Error: Undefined constant "SMF_UNICODE_VERSION" in D:\wamp64\www\84van21\Sources\Subs-Charset.php:353 Stack trace: # 0 D:\wamp64\www\84van21\Sources\Load.php(342): utf8_normalize_c('admin') # 1 D:\wamp64\www\84van21\Sources\Load.php(169): {closure:reloadSettings():326}('admin') # 2 D:\wamp64\www\84van21\Sources\QueryString.php(519): {closure:reloadSettings():167}('admin', 3) # 3 D:\wamp64\www\84van21\Sources\QueryString.php(523): htmlspecialchars__recursive('admin', 1) # 4 D:\wamp64\www\84van21\Sources\QueryString.php(152): htmlspecialchars__recursive(Array) # 5 D:\wamp64\www\84van21\index.php(145): cleanRequest() # 6 {main} thrown in D:\wamp64\www\84van21\Sources\Subs-Charset.php on line 353

Other errors include:
Image

Basically the process steps on itself...  

The bigger issue here is that the source code is fractured, one execution interferes with the next & the .php files are non-functional.

The only way to get your forum running is to actually restore the /Unicode files either from GitHub or a 2.1.6 upgrade package.

Not sure what the fix is here - force execution serialization somehow (I thought they already were...)?  Use hashed temp filenames, unique to that execution (just as effective & probably easier)?

### Steps to reproduce

1. Request Weekly Maintenance "Run Now" twice in succession.
2.

### Expected result

_No response_

### Actual result

_No response_

### Version/Git revision

2.1.6

### Database Engine

All

### Database Version

8.4.4

### PHP Version

8.4.5

### Logs

```shell

```

### Additional Information

_No response_

Sesquipedalian:
Fixes the real problem in #8683
Fixes #8713

### DIFF
--- a/Sources/Subs-Admin.php
+++ b/Sources/Subs-Admin.php
@@ -1930,10 +1930,13 @@ function safe_file_write($file, $data, $backup_file = null, $mtime = null, $appe
 		}
 	}
 
-	// We're done with these.
-	@unlink($temp_sfile);
+	clearstatcache();
 
-	if (!empty($temp_bfile))
+	// We're done with these.
+	if (is_file($temp_sfile))
+		@unlink($temp_sfile);
+
+	if (!empty($temp_bfile) && is_file($temp_bfile))
 		@unlink($temp_bfile);
 
 	if ($failed)
--- a/Sources/tasks/UpdateUnicode.php
+++ b/Sources/tasks/UpdateUnicode.php
@@ -558,7 +558,7 @@ class Update_Unicode extends SMF_BackgroundTask
 				$file_paths['temp'] = $this->temp_dir . DIRECTORY_SEPARATOR . $func_info['file'];
 
 				// If the temp file went missing, bail out immediately.
-				if (!file_exists($file_paths['temp']))
+				if (!is_readable($file_paths['temp']) || !is_writable($file_paths['temp']))
 					return true;
 
 				// Add closing PHP tag to the temp file.
--- a/Sources/tasks/UpdateUnicode.php
+++ b/Sources/tasks/UpdateUnicode.php
@@ -420,6 +420,18 @@ class Update_Unicode extends SMF_BackgroundTask
 		if (empty($this->temp_dir))
 			return true;
 
+		// Prevent race conditions.
+		if (is_file($this->temp_dir . DIRECTORY_SEPARATOR . 'lock'))
+			return true;
+
+		if (!@touch($this->temp_dir . DIRECTORY_SEPARATOR . 'lock'))
+			return true;
+
+		register_shutdown_function(function () {
+			if (file_exists($this->temp_dir . DIRECTORY_SEPARATOR . 'lock'))
+				unlink($this->temp_dir . DIRECTORY_SEPARATOR . 'lock');
+		});
+
 		// Do we even need to update?
 		if (!$this->should_update())
 		{
--- a/Sources/tasks/UpdateUnicode.php
+++ b/Sources/tasks/UpdateUnicode.php
@@ -553,20 +553,23 @@ class Update_Unicode extends SMF_BackgroundTask
 		{
 			require_once($sourcedir . '/Subs-Admin.php');
 
-			$done_files = array();
+			foreach ($this->funcs as $func_name => $func_info)
+			{
+				$file_paths['temp'] = $this->temp_dir . DIRECTORY_SEPARATOR . $func_info['file'];
+
+				// If the temp file went missing, bail out immediately.
+				if (!file_exists($file_paths['temp']))
+					return true;
+
+				// Add closing PHP tag to the temp file.
+				if (!preg_match('/[?]>$/', file_get_contents($file_paths['temp'])))
+					file_put_contents($file_paths['temp'], '?' . '>', FILE_APPEND);
+			}
 
 			foreach ($this->funcs as $func_name => $func_info)
 			{
 				$file_paths['temp'] = $this->temp_dir . DIRECTORY_SEPARATOR . $func_info['file'];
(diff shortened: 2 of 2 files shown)


Sesquipedalian:
@sbulen, please test this by running the Unicode update again. In addition to checking that the error messages no longer occur, I am also hoping for confirmation that the Unicode files are in fact updated successfully.

sbulen:
OK, so just repeated all tests...

The good news is that the 11 errors from the Weekly Maint have been addressed.   (And yes, the real reason the error occurs is that the file was renamed a few lines earlier...)

Further, the unicode files are being correctly updated.  (I restored /Unicode from v.2.1.4 before each test...) 

The ONLY difference I'm seeing is version #s.  Current Git = 2.1.5, rebuilt = 2.1.6.  Contents are identical.

I guess I was off on a red herring thinking it was associated with updates, because I tended to apply updates to recently unused, or recently rebuilt, or recently restored test forums.  I.e., navigating around the test forums triggered weekly maint...  🙄  I suspect folks reporting the 11 errors on the support boards were similar in one way or another.  

The back-to-back Weekly Maint executions are still problematic though:
image

Sesquipedalian:
Just to verify, was that test conducted using [256d3f6](https://github.com/SimpleMachines/SMF/pull/8749/commits/256d3f69cf4f7ccc9b6a03cfeb9969d14fc2de72)?

sbulen:
Yep.
image

Sesquipedalian:
Latest commit should solve that.

Sesquipedalian:
I realize that I wasn't clear in my last comment. Can you confirm that no more errors are logged with the latest commit, @sbulen?

jdarwood007:
Shouldn't we also use file locking (flock) so that the task handling it has exclusive access to the files, and another task shouldn't be able to interfere with it?

sbulen:
I'm still seeing these when I resubmit back-to-back:
image

sbulen:
And this particular file is corrupted as described earlier.   Doc block is missing, including the opening php tag.

This is comparing my current Idna.php (left), after the failure above, with what's in GitHub (right):
image

All other files look good.  The only problem I see right now is with Idna.php.

sbulen:
(I got a screenshot this time...  SeE?  i'M nOt cRaZy!  🤪)

sbulen:
Also got this one time:
image

Note I see different results under different circumstances.  Test variants include:
 - Restoring old unicode before running maint (to confirm the files get updated)
 - Running Weekly Maint 2x in quick succession, vs 3x, vs 4x, etc.

sbulen:
If I do a bunch in a row, I might see many errors, like this.  The line in question is an unlink, with no @:
image

sbulen:
> Shouldn't we also use file locking (flock) so that the task handling it has exclusive access to the files, and another task shouldn't be able to interfere with it?

@jdarwood007 - the safe file write function does use flock.  

I think there are layers of challenges here - first, the updates are in a set of several files that come from a remote source.   So...  There is a somewhat lengthy time duration, from beginning to end, of connecting & downloading multiple files.   

The process may even resubmit itself to resume work later if it thinks it runs too long.  
https://github.com/SimpleMachines/SMF/blob/5da8ca889a801e7a5245d609a68697fe4153cb84/Sources/Tasks/UpdateUnicode.php#L641

Also, it's not one file, it's 12.  

So I'm not sure we can actually have a 'stop it here' within the code itself.  I'm wondering if we need to stop duplication within cron itself.  Define a class of tasks (anything that updates source code, e.g., this one, maybe the tld regex one?) that may go long, and if they're already in the queue, have cron.php delete the subsequent request.  

Just thinking aloud.  I suspect we need to stop it at the traffic cop level, not all the individual scooters.

jdarwood007:
In task runner we could add a flock doing something like `flock($cachedir . '/'. md5($task_name) . '.lock'` and do the checks.  So if somehow another task runner came in and found it was running this, it would exit (or wait its turn?)  What your describing is a race condition and we have to prevent multiple instances from trying to do the same work.  Since the unicode updater can restart itself that is a problem.

We could abuse the task runner a bit.  It has a protection against grabbing the same instance with a `claimed_time` that is checked against `MAX_CLAIM_THRESHOLD`, which is 300 seconds (5 minutes).  When the unicode updater goes to do its insert for restarting, we could insert it with a claim_time of `time()-60*4`, giving the system an entire minute to finish up if it's slow.

Sesquipedalian:
> And this particular file is corrupted as described earlier. Doc block is missing, including the opening php tag.
> 
> This is comparing my current Idna.php (left), after the failure above, with what's in GitHub (right): image
> 
> All other files look good. The only problem I see right now is with Idna.php.

What the...???

Sesquipedalian:
> In task runner we could add a flock doing something like `flock($cachedir . '/'. md5($task_name) . '.lock'` and do the checks. So if somehow another task runner came in and found it was running this, it would exit (or wait its turn?) What your describing is a race condition and we have to prevent multiple instances from trying to do the same work. Since the unicode updater can restart itself that is a problem.

This was a great suggestion. Thanks, @jdarwood007!

@sbulen, I think the latest commit will fully solve the problem. Please confirm.

sbulen:
That fixed it!  I can run 10 in a row now...  I've confirmed updates are properly made.  

@jdarwood007 with the orthogonal thinking!

PR #8756, opened 2025-07-05
[3.0] Allow tasks to be defined as single running instances

jdarwood007:
@Sesquipedalian I had started to write this before the other PRs went up, but figured I should send this to see if there is interest.
This was my solution to the race condition by allowing tasks to define if they should only run once and if so, issue a lock file and hold onto it while they run.
We can improve on it and define other tasks that should also be single instance runners.

### DIFF
--- a/Sources/TaskRunner.php
+++ b/Sources/TaskRunner.php
@@ -616,7 +616,7 @@ class TaskRunner
 	 * Checks if a task is locked into a single instance.
 	 * While checking, we obtain a exclusive lock to run this task.
 	 *
-	 * @param \SMF_BackgroundTask $task
+	 * @param SMF\Tasks\BackgroundTask $task
 	 * @return bool True if we obtained a lock, false otherwse.
 	 */
 	protected function tryTaskLock(SMF_BackgroundTask $task): bool
--- a/Sources/TaskRunner.php
+++ b/Sources/TaskRunner.php
@@ -647,7 +647,7 @@ class TaskRunner
 	 *
 	 * @param \SMF_BackgroundTask $task
 	 */
-	protected function unlockTask(SMF_BackgroundTask $task): void
+	protected function unlockTask(BackgroundTask $task): void
 	{
 		$token_file = Config::$cachedir . DIRECTORY_SEPARATOR . md5($task::class) . '.lock';
 
--- a/Sources/TaskRunner.php
+++ b/Sources/TaskRunner.php
@@ -17,7 +17,7 @@ namespace SMF;
 
 use SMF\Db\DatabaseApi as Db;
 use SMF\Tasks\ScheduledTask;
-use SMF_BackgroundTask;
+use SMF\Tasks\BackgroundTask;
 
 /**
  * Runs background tasks (a.k.a. cron jobs), including scheduled tasks.
--- a/Sources/TaskRunner.php
+++ b/Sources/TaskRunner.php
@@ -573,7 +573,7 @@ class TaskRunner
 		}
 
 		// Normally, the class should be specified using its fully qualified name.
-		if (class_exists($task_details['task_class']) && is_subclass_of($task_details['task_class'], SMF_BackgroundTask::class)) {
+		if (class_exists($task_details['task_class']) && is_subclass_of($task_details['task_class'], BackgroundTask::class)) {
 			$details = empty($task_details['task_data']) ? [] : Utils::jsonDecode($task_details['task_data'], true);
 
 			$bgtask = new $task_details['task_class']($details);
--- a/Sources/TaskRunner.php
+++ b/Sources/TaskRunner.php
@@ -16,6 +16,8 @@ declare(strict_types=1);
 namespace SMF;
 
 use SMF\Db\DatabaseApi as Db;
+use SMF\Tasks\ScheduledTask;
+use SMF_BackgroundTask;
 
 /**
  * Runs background tasks (a.k.a. cron jobs), including scheduled tasks.
@@ -108,6 +110,12 @@ class TaskRunner
 			'class' => 'SMF\\Tasks\\PruneLogTopics',
 		],
 	];
+
+	/*********************
+	 * Internal properties
+	 *********************/
+
+	protected $fp = null;
 
 	/****************
 	 * Public methods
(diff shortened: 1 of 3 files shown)


sbulen:
We could either treat these special, or, ensure _everything_ single threads.  

The vast majority of tasks are fine because they're one-and-done.

I suspect the issue is only with tasks that run long, and discontinue to resume work mid-task later.

jdarwood007:
Well, not all tasks would be single-threaded.  Consider the notification for a new post (CreatePost_Notify).  It is fine if we are processing this multiple times because the data set passed in would be to process (in theory) different data sets.  The Unicode updater on the other hand, has one job and does not accept different data sets to change what it would do.

I am sure a few other tasks should be defined as a single instance if we want to proceed with this idea.

We could flip the logic, make it required to mark a runner as multi-instance capable, otherwise tasks would default to single instance only.

sbulen:
I was thinking it more along the lines of KISS...  If everything is single threaded, nothing can ever step on itself.  

I like simple.

OTOH, I doubt we have a lot of tasks that literally split work up in chunks that must be processed serially like the unicode updater.  It's possible the unicode task is the only one that will ever face this issue.

Sources/TaskRunner.php
```diff
@@ -16,6 +16,8 @@
 namespace SMF;
 
 use SMF\Db\DatabaseApi as Db;
+use SMF\Tasks\ScheduledTask;
+use SMF_BackgroundTask;
```
Sesquipedalian:
```suggestion
use SMF\Tasks\BackgroundTask;
```

Sources/TaskRunner.php
```diff
@@ -565,22 +573,18 @@ protected function performTask(array $task_details): bool
 		}
 
 		// Normally, the class should be specified using its fully qualified name.
-		if (class_exists($task_details['task_class']) && is_subclass_of($task_details['task_class'], 'SMF\\Tasks\\BackgroundTask')) {
+		if (class_exists($task_details['task_class']) && is_subclass_of($task_details['task_class'], SMF_BackgroundTask::class)) {
```
Sesquipedalian:
```suggestion
		if (class_exists($task_details['task_class']) && is_subclass_of($task_details['task_class'], BackgroundTask::class)) {
```

Sources/TaskRunner.php
```diff
@@ -565,22 +573,18 @@ protected function performTask(array $task_details): bool
...
 			$details = empty($task_details['task_data']) ? [] : Utils::jsonDecode($task_details['task_data'], true);
 
 			$bgtask = new $task_details['task_class']($details);
-
-			$success = $bgtask->execute();
 		}
 		// Just in case a mod or something specified a task without giving the namespace.
-		elseif (class_exists('SMF\\Tasks\\' . $task_details['task_class']) && is_subclass_of('SMF\\Tasks\\' . $task_details['task_class'], 'SMF\\Tasks\\BackgroundTask')) {
+		elseif (class_exists('SMF\\Tasks\\' . $task_details['task_class']) && is_subclass_of('SMF\\Tasks\\' . $task_details['task_class'], SMF_BackgroundTask::class)) {
```
Sesquipedalian:
```suggestion
		elseif (class_exists('SMF\\Tasks\\' . $task_details['task_class']) && is_subclass_of('SMF\\Tasks\\' . $task_details['task_class'], BackgroundTask::class)) {
```

Sources/TaskRunner.php
```diff
@@ -590,15 +594,73 @@ protected function performTask(array $task_details): bool
...
+
+		$success = $bgtask->execute();
+
+		$this->unlockTask($bgtask);
+
 		// For scheduled tasks, log it and update our next scheduled task time.
-		if (is_subclass_of($bgtask, 'SMF\\Tasks\\ScheduledTask')) {
+		if (is_subclass_of($bgtask, ScheduledTask::class)) {
 			$bgtask->log();
 			Tasks\ScheduledTask::updateNextTaskTime();
```
Sesquipedalian:
```suggestion
			ScheduledTask::updateNextTaskTime();
```

Sources/TaskRunner.php
```diff
@@ -590,15 +594,73 @@ protected function performTask(array $task_details): bool
...
 		}
 
 		return $success;
 	}
 
+	/**
+	 * Checks if a task is locked into a single instance.
+	 * While checking, we obtain a exclusive lock to run this task.
+	 *
+	 * @param \SMF_BackgroundTask $task
```
Sesquipedalian:
```suggestion
	 * @param SMF\Tasks\BackgroundTask $task
```

Sources/TaskRunner.php
```diff
@@ -590,15 +594,73 @@ protected function performTask(array $task_details): bool
...
+	 * @return bool True if we obtained a lock, false otherwse.
+	 */
+	protected function tryTaskLock(SMF_BackgroundTask $task): bool
```
Sesquipedalian:
```suggestion
	protected function tryTaskLock(BackgroundTask $task): bool
```

Sources/TaskRunner.php
```diff
@@ -590,15 +594,73 @@ protected function performTask(array $task_details): bool
...
+			&& ($this->fp = fopen($token_file, 'c'))
+			&& flock($this->fp, LOCK_EX)
+		);
+	}
+
+	/**
+	 * Releases a lock on a task.
+	 * Does not check if the task allows mulitple instances, checks if we have any resource or file in use.
+	 *
+	 * @param \SMF_BackgroundTask $task
```
Sesquipedalian:
```suggestion
	 * @param SMF\Tasks\BackgroundTask $task
```

Sources/TaskRunner.php
```diff
@@ -590,15 +594,73 @@ protected function performTask(array $task_details): bool
...
+	 */
+	protected function unlockTask(SMF_BackgroundTask $task): void
```
Sesquipedalian:
```suggestion
	protected function unlockTask(BackgroundTask $task): void
```

Sesquipedalian (dismissed):
In 3.0, `SMF_BackgroundTask` has been renamed to `SMF\Tasks\BackgroundTask`

Sesquipedalian:
I'm not sure about this idea yet. I do like it, but I'll need to ponder it a bit more.

> OTOH, I doubt we have a lot of tasks that literally split work up in chunks that must be processed serially like the unicode updater. It's possible the unicode task is the only one that will ever face this issue.

The ExportProfileData task is another example.

jdarwood007:
Oh my bad.  VScode suggested `SMF_Background` task for me, and I took its word for it.

You're welcome to dismiss the PR or write up something else.  I think the idea is sound that we want to ensure our task runners are able to lock out a duplicate instance when needed.  How we go about that can be done in a few ways.

Export profile is a good example as it can be executed by multiple members, so it needs to allow multiple instances, but we don't want it to accidentally run the same export.  Perhaps we can define in the data set a unique key for exporting (generate a uuid), then have a way to define the instance id (a function) which would return that uuid, and we check that instead of just a class name.

Sesquipedalian:
The idea is sound, yes. 

However, I think all of the relevant logic needs to be implemented in BackgroundTask and its descendants, not in TaskRunner. That way, tasks can customize what sort of locking and unlocking they do (if any) as required for their specific needs.

Sesquipedalian:
I'm working on an alternative version of this idea.

jdarwood007:
Ok. I had an idea I was thinking of.  It was defining in the interface
```
public function instanceName: ?string {return null;}
```

Then, each task can dynamically generate a name.

Such as some of them may just do themselves:
```
public function instanceName: ?string {return static::class;}
```

Or more complex like:
```
public function instanceName: ?string {return static::class . $this->opts['id_member'];}
```

We would use that to generate the lock file.  If it's null, ignore the lock; otherwise, we compare to determine if we have run into the same task running.

PR #8760, opened 2025-07-05
[3.0] Don't save a comment on non existent mod report

### LINKED ISSUE #7429: Moderator reported posts comments
#### Description
/index.php?action=moderate;area=reportedposts;sa=handlecomment;rid=xxxx
/Sources/Subs-ReportedContent.php (Line 609)

Type of error: General
Error messageSelect
8: Trying to access array offset on value of type bool

### Steps to reproduce
1. Unknown, in error log
2. 

- Version/Git revision: 2.1.1
- Database Type:
- Database Version:
- PHP Version: 7.4

```
			'msg_id' => $report['id_msg'],
```

jdarwood007:
Fixes #7429

### DIFF
--- a/Sources/Actions/Moderation/ReportedContent.php
+++ b/Sources/Actions/Moderation/ReportedContent.php
@@ -1215,6 +1215,10 @@ class ReportedContent implements ActionInterface
 
 		$report = $this->getReportDetails($report_id);
 
+		if (empty($report)) {
+			return;
+		}
+
 		if ($this->type == 'members') {
 			$prefix = 'Member';
 			$data = [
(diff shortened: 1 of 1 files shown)


Sources/Actions/Moderation/ReportedContent.php
```diff
@@ -1215,6 +1215,10 @@ protected function saveModComment(int $report_id, array $data): void
 
 		$report = $this->getReportDetails($report_id);
 
+		if (empty($report)) {
+			return;
+		}
+
```
Sesquipedalian: While we're at it, we should probably move all of this chunk to the top of the function, before we even try to insert the comment. After all, if the report doesn't exist, what are we commenting on?

Sesquipedalian (changes requested):

Sesquipedalian (approved):

PR #8762, opened 2025-07-05
[3.0] Allow nulls on db_insert

### LINKED ISSUE #7504: Can not insert null values.
#### Description
Lets assume you have a table that allows nulls
```
CREATE TABLE `test` ( `id` INT NOT NULL AUTO_INCREMENT , `mycol` INT NULL , PRIMARY KEY (`id`)) ENGINE = InnoDB;
```

Try to use a $smcFunc['insert'] on this.

```
$smcFunc['db_insert']('insert',
	'test',
	[
		'id' => 'int',
		'mycol' => int'
	],
	[
		[
			'id' => 1,
			'mycol' => null
		]
	]
);
```

This will fail with a error
The database value you're trying to insert does not exist: mycol

This is because of this line of code
```
	if (!isset($values[$matches[2]]))
		smf_db_error_backtrace('The database value you\'re trying to insert does not exist: ' . (isset($smcFunc['htmlspecialchars']) ? $smcFunc['htmlspecialchars']($matches[2]) : htmlspecialchars($matches[2])), '', E_USER_ERROR, __FILE__, __LINE__);
```

PHP returns false for isset on nulls.  A way around this is to use in_array($matches[2], $values).  This will return true.

jdarwood007:
Fixes #7504

Additional testing requested to ensure nothing is broken by this


Sesquipedalian:
Beyond this, we would also need to handle null values correctly in the following switch block. I'm currently on my phone and unable to test, but just looking at the code I see no way that a null value would pass through the switch block without triggering an error.

jdarwood007:
You are correct.  I could do before the start of the switch do this.

```
if ($matches[2] === null && ! str_starts_with($matches[1], 'array') {
	return null;
}
```

Which would allow the possibility of an unknown type, but a minor issue in my mind.
Then add logic to the array_* matches to handle the nulls in their foreach loop of something like:
```
if ($value === null) {
	$replacement[$key] = null;
}
```

Sesquipedalian:
Returning a null value here would result in a malformed SQL statement in the end. We would want to return the string `NULL`.

As for the array_* types, those should absolutely refuse to accept null values, just like they do now. Scalars can be nullable in SQL; arrays can only be empty.

live627:
I use the `raw` datatype whenever I must set ~fire~`null` to the database.  Works great even in 2.0.

jdarwood007:
Yup, the raw workaround works.  This should eliminate the need for it.  If we accept nulls on the database handling, we should accept it for inserts and other places where we handle database stuff.

The select queries can use `col = NULL` or `col = {string:null}` and then pass `'NULL'` as the string.  Databases I worked with accept either literal `NULL` or the string of `'NULL'` to mean the same thing.  The only issue that gets people is remembering that `NULL` does not equal an empty string (as it does in PHP), a typical fix is to use something like `COALESCE(col, {empty}) <> = {empty}` to ensure you're comparing a non-null value against your other value.

Sources/Db/APIs/MySQL.php
```diff
@@ -2584,11 +2584,17 @@ protected function replacement__callback(array $matches): string
...
 		}
 
-		$replacement = $this->temp_values[$matches[2]];
+		$replacement = $this->temp_values[$matches[2]] ?? null;
+
+		if ($replacement === null && str_starts_with($matches[1], 'array_')) {
+			$this->error_backtrace('The database value you\'re trying to insert does not exist: ' . Utils::htmlspecialchars($matches[2]), '', E_USER_ERROR, __FILE__, __LINE__);
+		} elseif ($replacement === null) {
+			return 'NULL';
+		}
```
```suggestion
		$replacement = $this->temp_values[$matches[2]];

		if ($replacement === null) {
			if (str_starts_with($matches[1], 'array_')) {
				$this->error_backtrace('The database value you\'re trying to insert does not exist: ' . Utils::htmlspecialchars($matches[2]), '', E_USER_ERROR, __FILE__, __LINE__);
			}
			
			return 'NULL';
		}
```

Don't need the null coalescing operator since `array_key_exists()` was used above.

Sources/Db/APIs/PostgreSQL.php
```diff
@@ -2516,11 +2516,17 @@ protected function replacement__callback(array $matches): string
...
 		}
 
-		$replacement = $this->temp_values[$matches[2]];
+		$replacement = $this->temp_values[$matches[2]] ?? null;
+
+		if ($replacement === null && str_starts_with($matches[1], 'array_')) {
+			$this->error_backtrace('The database value you\'re trying to insert does not exist: ' . Utils::htmlspecialchars($matches[2]), '', E_USER_ERROR, __FILE__, __LINE__);
+		} elseif ($replacement === null) {
+			return 'NULL';
+		}
```
Sesquipedalian:
```suggestion
		$replacement = $this->temp_values[$matches[2]];

		if ($replacement === null) {
			if (str_starts_with($matches[1], 'array_')) {
				$this->error_backtrace('The database value you\'re trying to insert does not exist: ' . Utils::htmlspecialchars($matches[2]), '', E_USER_ERROR, __FILE__, __LINE__);
			}
			
			return 'NULL';
		}
```

Sesquipedalian (changes requested):

Sesquipedalian (approved):

Issue #8707, opened 2025-06-27
[3.0] Better error handling needed for failed image re-encode operations


mwegithub:
### Basic Information

I'm using a plain vanilla SMF 2.15 without any plug-ins or custom themes. It runs on Linux with PHP 8.4 and  FastCGI enabled.

I've upgraded from the 2.14 to the 2.15 release yesterday. Immediately, the error log started to fill up.

I could fix the issue causing hundreds of log entries about _$id being undefined_ in QueryString.php myself (there is already a pull request from another user for this). Also fixed it in Subs.php.

Now I have several hundred other log file entries (in a low-volume community) like this:

```Bogus marker length `/www/htdocs/w0181784/example.com/community/attachments/13876_85d5997b92bca48be9188ac930fa3f11b9051feb.dat_thumb.tmp' @ error/jpeg.c/JPEGErrorHandler/339```

coming from

`/www/htdocs/w0181784/example.com/community/Sources/Subs-Graphics.php, line 503.`

It seems every time ImageMagick is initialized with a temporary JPG SMF creates (?) it logs this entry. Is this a known issue? Can this be caused by the ImageMagick version my hoster has installed? I'm not sure where to start looking.

Another repeating error message I get with 2.15 but had never before is:

`Unable to verify referring URL. Please go back and try again.`

The backtrace shows:

```https://www.example.com/community/index.php?action=likes;sa=view;ltype=msg;like=96800;a38606a896d=a8c260d8069ebb30c4224b32ff64b5f8```

````    #0: fatal_lang_error()
    Called from /www/htdocs/w0181784/example.com/community/Sources/Security.php on line 710
    #1: checkSession()
    Called from /www/htdocs/w0181784/example.com/community/Sources/Likes.php on line 152
    #2: call()
    Called from /www/htdocs/w0181784/example.com/community/index.php on line 185
````
I recall some PHP from older projects, but I don't know enough about SMF to figuring out why these new errors show up so often with the 2.15 release.

### Steps to reproduce

1. Just let the community run for a while and observe the log file.

### Expected result

None of these error log entries should show. They did not with the 2.14 release.

### Version/Git revision

2.15 from June 26.

### Database Engine

All

### Logs

```shell

```

jdarwood007:
Can you go to Admin > Support and Credits and provide us with all the version information listed?

mwegithub:
Of course, happy to help. Here it is:

[...]

PR #8763, opened 2025-07-05
[3.0] Use attachment type constants

jdarwood007:
Update instances of numerical ids with constants for the Attachment type

While looking into #8707, I see that we had a constant for attachment type, but the code references the numerical values for all its checks.  I updated what I could find.

### DIFF
--- a/Sources/Config.php
+++ b/Sources/Config.php
@@ -2772,7 +2772,7 @@ class Config
 		}
 
 		if (!isset($db_last_error)) {
-			self::updateDbLastError(0, true);
+			self::updateDbLastError(0);
 		} else {
 			self::$db_last_error = (int) $db_last_error;
 		}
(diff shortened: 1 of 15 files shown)


Sources/Config.php
```diff
@@ -2772,7 +2772,7 @@ public static function getDbLastError(): int
 		}
 
 		if (!isset($db_last_error)) {
-			self::updateDbLastError(0, true);
+			self::updateDbLastError(0);
```
Sesquipedalian: This seems like it belongs in a separate PR.

Sesquipedalian (changes requested):

Sources/Config.php
jdarwood007: My bad, the editor noticed that while I had the file open, I meant to send that up separately.

Sesquipedalian (approved):

PR #8767, opened 2025-07-09
[3.0] Implements concurrency protection for tasks

Sesquipedalian:
1. Generalizes the fix from #8570, using ideas from @jdarwood007's #8756, so that tasks that should not run concurrent instances can configure themselves to prevent that.
2. Implements SMF\Tasks\BackgroundTask::respawn() so that any task that needs that ability can access it simply and consistently. (Previously, several tasks implemented their own, slightly differing versions of this.)

### DIFF
--- a/Sources/TaskRunner.php
+++ b/Sources/TaskRunner.php
@@ -138,12 +138,6 @@ class TaskRunner
 				ErrorHandler::displayMaintenanceMessage();
 			}
 
-			// Have we already turned this off? If so, exist gracefully.
-			// @todo Remove this? It's a bad idea to ever disable background tasks.
-			if (file_exists(Config::$cachedir . '/cron.lock')) {
-				$this->obExit();
-			}
-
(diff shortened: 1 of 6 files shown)


Sources/TaskRunner.php
```diff
@@ -565,21 +561,35 @@ protected function performTask(array $task_details): bool
...
 
 			$task_class = 'SMF\\Tasks\\' . $task_details['task_class'];
 
 			$bgtask = new $task_class($details);
 
+			if (!$bgtask->canExecute()) {
+				return true;
+			}
+
 			$success = $bgtask->execute();
```
jdarwood007: This and the previous section are repeated.  Can we move this logic outside the if/elseif statements to reduce code?

jdarwood007 (approved):
This is where I was planning on heading.  Left a comment about moving a block of code but that isn't required.

PR #8768, opened 2025-07-12
[3.0] Improve load average detection

### LINKED ISSUE #7397: Package manager, load_average undefined errors
Installing packages on SMF 2.1.1 causes undefined index errors for me. 
I got these two on the same second while installing a package, the installation is succesful though.

?action=admin;area=packages;sa=install2;package=1613849140.zip;pid=0
/httpdocs/Sources/Load.php (Line 341)  
"8: Undefined index: load_average"

?action=admin;area=packages;sa=install2;package=1613849140.zip;pid=0
/httpdocs/Sources/Load.php (Line 337)
"8: Undefined index: load_average"

I would suspect the cause and solution to be the same or similar as here
https://github.com/SimpleMachines/SMF2.1/pull/4804

jdarwood007:
This improves the load average detection by moving the logic into Sapi. A new function is added to simplify calls to checking if the server is overloaded Additionally it attempts to make use of the built in sys_getloadavg(), with checks incase a server admin has disabled it.

This will Fix #7397, although that is marked as a 2.1 bug.

A note about linux was added and how load is calculated.  We could run `nproc` and divide that by the load.  But Linux admins are used to the logic and it would just confuse more server admins than help.

A note about Windows support was included.  I attempted to check the speed of commands, but they where slow.  On my i9-10850k running at idle, I ran those commands through Measure-Command powershell function and they would take 100+ milliseconds to complete.  Not ideal to add 0.1+ load time just to calculate the server load.  Wmic is known to be slow to process commands.

### DIFF
--- a/Sources/Actions/Admin/Server.php
+++ b/Sources/Actions/Admin/Server.php
@@ -603,7 +603,7 @@ class Server implements ActionInterface
 				$_SESSION['adm-save'] = Utils::$context['settings_message']['label'];
 			}
 		} elseif (!self::$loadAverageDisabled) {
-			Utils::$context['settings_message']['label'] = Lang::getTxt('loadavg_warning', [Config::$modSettings['load_average']], file: 'ManageSettings');
+			Utils::$context['settings_message']['label'] = Lang::getTxt('loadavg_warning', [Sapi::getLoadAverage()], file: 'ManageSettings');
 		}
 
 		$config_vars = self::loadBalancingConfigVars();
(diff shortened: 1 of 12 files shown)


Sesquipedalian:
I like it, but I wonder if we should also do something like [this](https://www.php.net/manual/en/function.sys-getloadavg.php#129847) to get the number of CPU cores and divide the load average by that number in order to get a more meaningful value on multicore machines.

We wouldn't need to check the number of CPU cores often. Indeed, in this case I wouldn't even use the cache; I would save it in the settings table and update it once per week during weekly maintenance.

jdarwood007:
I covered that in the fact that Linux administrators know and understand it.  So the question is, do we break that to make it more understandable to the average user?  Another note being that 25% load with 4 cores could mean 1 CPU at 100% and 3 at 0 or 4 cpus at 6.25%.  One is an overloaded core that will cause problems because a threaded SAPI would have children on the saturated core that wouldn't respond.  The other is a fairly fine system load.

Sesquipedalian:
> I covered that in the fact that Linux administrators know and understand it.

Except when they don't. 

Moreover, many forum admins have no significant Linux admin knowledge at all. Their forum might be hosted on a Linux server, but they just installed SMF following instructions and have no idea how to admin the server, let alone what load average means.

> So the question is, do we break that to make it more understandable to the average user? Another note being that 25% load with 4 cores could mean 1 CPU at 100% and 3 at 0 or 4 cpus at 6.25%. One is an overloaded core that will cause problems because a threaded SAPI would have children on the saturated core that wouldn't respond. The other is a fairly fine system load.

I don't think this is really about what is or isn't understandable to the forum admin. It's about how we calculate when to limit certain forum features for the sake of load balancing. What matters for that purpose is knowing the overall system load as a percentage of the system's total load capacity. 

If the forum admin is indeed knowledgable about load averages, knows the number of CPUs available on their server, and has good historical knowledge of load and usage patterns for their website to draw on, then yes, they can set good values for SMF's various load management settings. But if the admin does not have all the necessary knowledge required to customize those settings to fit the reality of their particular server, they will almost certainly leave the default values as they are. 

The problem is, those [default values](https://github.com/SimpleMachines/SMF/blob/e03a1abaa37646f6258d50bd276418b98a944c3e/Sources/Actions/Admin/Server.php#L97-L106) may be wildly inappropriate for any particular server. For example, the default setting will disable search if the load average is 2.5. If the server has 2 CPU cores, then we're disabling search when the system load is at 125% capacity, which is a bit conservative but not unreasonable. But if the server has 8 CPU cores, then we're disabling search when the system load is at only 31.25% capacity, which is ridiculously low.

So rather than raw load average values, we should be making these decisions based on the current load average as a percentage of the system's total load capacity.

Sesquipedalian:
Still, I suppose all of those concerns go beyond the scope of this PR. This PR is only concerned with improving how we get the current load average, not with how we use it.

Sesquipedalian (approved):

jdarwood007:
I see how the defaults can be wrong as they are making an assumption (in my eyes) on the number of cores.  I am willing to work on adding the logic to figure out the CPUs and then divide that by the load to get the true load in a factor out of 100%.

Sesquipedalian:
That would be awesome! Thanks!

PR #8770, opened 2025-07-12
[3.0] Various fixes

jdarwood007:
Odds and ends I found while reading some of the code.  If I need to revert any changes or do separate PRs, I can.  Just a lot of little stuff and didn't want to create a dozen PRs

- .gitignore
    - 	Add Autolinker.js to ignore files
- Action\Unread
    - 	$start was undefined
- Calendar\Birthday
    - 	PHP Doc update to use \Generator to prevent IDE confusion
- Diff\Diff
    - 	A few instances where the type casting may lead to a non-integer.
- Diff\FullDiff
    - 	Time was not declared in our Use statement
    - 	Can not use a positional argument after a named argument
    - 	Changed $path1 and $path2 as they where undefined
    - 	Added $section to fixFinalLineEndings as it was missing
- Localization\MessageFormater
    - 	Ensure that country_currencies exists before using it, or fall back for safety.
- Markdown Parser
    - 	Change the usage of $this->message to $string as it does not exist
    - 	load() Attempted to use a 2nd parameter for the construct, which does not exist
    - 	Improve type hinting and PHP Doc on getMethod so our IDE knows what is going on
- Autolinker
    - 	Ensure $pos is an int
- User
    - 	Always ensure that display_fields is an array if null.

### DIFF
--- a/Sources/Diff/FullDiff.php
+++ b/Sources/Diff/FullDiff.php
@@ -1192,7 +1192,7 @@ class FullDiff extends Diff
 		$diff->changes = self::redistributeContextLines($diff->changes);
 
 		// Deal with '\ No newline at end of file'
-		$diff->changes = self::fixFinalLineEndings($no_newline_at_end, $diff->changes, $section);
+		$diff->changes = self::fixFinalLineEndings($no_newline_at_end, $diff->changes);
 
 		// Do all the cleanup the constructor normally would.
 		$diff->changes = $diff->consolidate($diff->changes);
@@ -1339,7 +1339,7 @@ class FullDiff extends Diff
 		$diff->changes = self::redistributeContextLines($diff->changes);
 
 		// Deal with '\ No newline at end of file'
-		$diff->changes = self::fixFinalLineEndings($no_newline_at_end, $diff->changes, $section);
+		$diff->changes = self::fixFinalLineEndings($no_newline_at_end, $diff->changes);
 
 		// Do all the cleanup the constructor normally would.
 		$diff->changes = $diff->consolidate($diff->changes);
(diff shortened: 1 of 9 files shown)


Sources/Diff/FullDiff.php
```diff
@@ -1518,9 +1518,10 @@ protected static function redistributeContextLines(array $changes): array
 	 *
 	 * @param array $where Info about which change parts have no final newline.
 	 * @param array $changes The diff's complete set of changes.
+	 * @param string $section Which section
 	 * @return array Updated $changes.
 	 */
-	protected static function fixFinalLineEndings(array $where, array $changes): array
+	protected static function fixFinalLineEndings(array $where, array $changes, string $section): array
```
Sesquipedalian: This is not the right way to fix this. The problem is simply that all occurrences of `$section` inside this function should actually be `$part`.

Sources/Parsers/MarkdownParser.php
```diff
@@ -893,10 +893,8 @@ public function parse(string $string, bool $from_bbcode_parser = false, array $o
...
 	public static function load(int $output_type = self::OUTPUT_HTML, ?int $hard_breaks = null): self
 	{
-		$hard_breaks = $hard_breaks ?? (int) (Config::$modSettings['markdown_brs'] ?? 0);
-
 		if (!isset(self::$parsers[$output_type][$hard_breaks])) {
-			self::$parsers[$output_type][$hard_breaks] = new self($output_type, $hard_breaks);
+			self::$parsers[$output_type][$hard_breaks] = new self($output_type);
 		}
 
 		return self::$parsers[$output_type][$hard_breaks];
```
Sesquipedalian: Please remove all references to `$hard_breaks` entirely.
1. Please remove the `$hard_breaks` input parameter and the documentation for it.
2. Change `self::$parsers[$output_type][$hard_breaks]` to just `self::$parsers[$output_type]` in all cases.

Sources/User.php
```diff
@@ -1104,7 +1104,7 @@ public function format(bool $display_custom_fields = false): array
 			$this->formatted['custom_fields'] = [];
 
 			if (!isset(Utils::$context['display_fields'])) {
-				Utils::$context['display_fields'] = Utils::jsonDecode(Config::$modSettings['displayFields'], true);
+				Utils::$context['display_fields'] = Utils::jsonDecode(Config::$modSettings['displayFields'], true) ?? [];
```
```suggestion
				Utils::$context['display_fields'] = Utils::jsonDecode(Config::$modSettings['displayFields'] ?? '[]', true);
```

Sesquipedalian (changes requested):

Sources/User.php
jdarwood007: I will double-check this one, but it's fine.  I encountered a case where the mod setting was not provided, leading to an error during JSON decoding.  This issue stems from custom profile fields, which I suspect are corrupted on my installation.

Sources/Parsers/MarkdownParser.php
```diff
@@ -317,15 +317,6 @@ class MarkdownParser extends Parser
...
 
-	/**
-	 * @var int
-	 *
-	 * How to render line breaks.
-	 *
-	 * Value should be a bitmask of this class's BR_* constants.
-	 */
-	public int $hard_breaks = 0;
-
```
Sesquipedalian: This still needs to exist. When I said in my last comment to remove all references to `$hard_breaks`, I only meant within `MarkdownParser::load()`. This property is set in `MarkdownParser::parse()` and is used in many places.

Sesquipedalian (changes requested):

Sesquipedalian (approved):

PR #4411, opened 2017-12-17
Fix Hacking check by temp converting Fixes #3766

albertlast:
pg escape('') to mysql escape (\\')
Signed-off-by: albertlast albertlast@hotmail.de

Fixes #3766

PR #8783, opened 2025-07-18
[2.1] Fixes "No direct access..." error when slash at end of message

### LINKED ISSUE #8309: [2.1]: Smileys at end of PMs cause 'No Direct Access' Error
### Basic Information

In 2.1, a PM with a smiley at the end causes a 'No Direct Access' error.

Not an issue in 3.0.

### Steps to reproduce

1. Send a PM with a smiley at the end

### Expected result

Sent PM

### Actual result

On screen:
image

In the error log:
image

### Version/Git revision

2.1.4

### Database Engine

All

### Database Version

8.4

### PHP Version

8.3.8

### Logs

_No response_

### Additional Information

From the forum: 
https://www.simplemachines.org/community/index.php?topic=589648.0

Sesquipedalian:
Fixes #8309

### DIFF
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -368,12 +368,14 @@ function smf_db_query($identifier, $db_string, $db_values = array(), $connection
 
 	// Comments that are allowed in a query are preg_removed.
 	static $allowed_comments_from = array(
+		'~(?<![\'\\\\])\'\X*?(?<![\'\\\\])\'~',
 		'~\s+~s',
 		'~/\*!40001 SQL_NO_CACHE \*/~',
 		'~/\*!40000 USE INDEX \([A-Za-z\_]+?\) \*/~',
 		'~/\*!40100 ON DUPLICATE KEY UPDATE id_msg = \d+ \*/~',
 	);
 	static $allowed_comments_to = array(
+		' %s ',
 		' ',
 		'',
 		'',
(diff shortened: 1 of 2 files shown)


jdarwood007:
I can test later next week.

@albertlast are you able to confirm we didn't undo the previous fix?

albertlast:
Im stuck in the upgrade progress so got a hardtime to check, could maybe install a new env...
At least for mysql i didn't understand the change, since it just roll back the change from #4411 without adding new logic.
For postgresql i didn't understrand what the change does.

In my eyes the pr #5180 would be fixed but i see no willing to want this.

jdarwood007 (approved):

jdarwood007:
@albertlast Any hints on reproducing the original report?  The PR was attempting to fix #3766 but seems to cause some side effects.

albertlast:
Look like there was a time or is still, where smf.org send js code to the smf setup,
the pure return look like this: https://github.com/SimpleMachines/SMF/issues/3766#issuecomment-332950388

jdarwood007:
I still don't know how that output would happen.  That looks like json based response.

Sesquipedalian:
I've replaced the while loop with a simple regex that finds single quoted strings and replaces them with the ` %s ` placeholder. Using this regex is much easier to follow than the loop, and it will be much more resilient against any future code changes.

The regex can handle both MySQL escaping syntax (i.e. `\'`) and PostgreSQL escaping syntax (i.e. `''`) to detect escaped single quote characters inside the strings.

Sesquipedalian:
> I still don't know how that output would happen. That looks like json based response.

As @albertlast described in the initial issue description in #3766, the content returned from the URL https://www.simplemachines.org/smf/latest-news.js is full JavaScript, not just a JSON array.

Sesquipedalian:
FYI, the fundamental problem in #3766 was that the while loop method of finding and replacing strings is simply not robust enough to catch all the strings it should when those strings contain escaped single quotes. #4411 attempted to fix that, and it did work for the particular strings that @albertlast was testing against, but it introduced problems for other strings.

By getting rid of the while loop entirely and instead using a regular expression to find the strings, this PR solves the problem permanently.

PR #8790, opened 2025-07-20
[3.0] Change load average to percent based results

jdarwood007:
Still returns a float, however, it's now based on CPU load/processors to give you the percent of the CPU it's based on. I removed the mac os restrictions, as I don't see why we would otherwise be preventing it from working there.

I adjusted the defaults, but we may want to discuss them. I didn't add an upgrader step for this.  I don't feel we should change it.  If we do anything, I would suggest just turning it off and letting the admin reset things.

### DIFF
--- a/Sources/Actions/Admin/Server.php
+++ b/Sources/Actions/Admin/Server.php
@@ -95,14 +95,14 @@ class Server implements ActionInterface
 	 * Default values for load balancing options.
 	 */
 	public const LOADAVG_DEFAULT_VALUES = [
-		'loadavg_auto_opt' => 20.0,
-		'loadavg_search' => 30.0,
-		'loadavg_allunread' => 40.0,
-		'loadavg_unreadreplies' => 40.0,
-		'loadavg_show_posts' => 40.0,
-		'loadavg_userstats' => 50.0,
-		'loadavg_bbc' => 60.0,
-		'loadavg_forum' => 85.0,
+		'loadavg_auto_opt' => 50.0,
+		'loadavg_search' => 300.0,
+		'loadavg_allunread' => 200.0,
+		'loadavg_unreadreplies' => 200.0,
+		'loadavg_show_posts' => 200.0,
+		'loadavg_userstats' => 150.0,
+		'loadavg_bbc' => 700.0,
+		'loadavg_forum' => 1000.0,
 	];
 
 	/*******************
--- a/Sources/Actions/Admin/Server.php
+++ b/Sources/Actions/Admin/Server.php
@@ -95,14 +95,14 @@ class Server implements ActionInterface
 	 * Default values for load balancing options.
 	 */
 	public const LOADAVG_DEFAULT_VALUES = [
-		'loadavg_auto_opt' => 1.0,
-		'loadavg_search' => 2.5,
-		'loadavg_allunread' => 2.0,
-		'loadavg_unreadreplies' => 3.5,
-		'loadavg_show_posts' => 2.0,
-		'loadavg_userstats' => 10.0,
-		'loadavg_bbc' => 30.0,
-		'loadavg_forum' => 40.0,
+		'loadavg_auto_opt' => 20.0,
+		'loadavg_search' => 30.0,
+		'loadavg_allunread' => 40.0,
+		'loadavg_unreadreplies' => 40.0,
+		'loadavg_show_posts' => 40.0,
+		'loadavg_userstats' => 50.0,
+		'loadavg_bbc' => 60.0,
+		'loadavg_forum' => 85.0,
 	];
 
 	/*******************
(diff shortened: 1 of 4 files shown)


Sources/Actions/Admin/Server.php
```diff
@@ -95,14 +95,14 @@ class Server implements ActionInterface
...
-		'loadavg_bbc' => 30.0,
-		'loadavg_forum' => 40.0,
+		'loadavg_auto_opt' => 20.0,
+		'loadavg_search' => 30.0,
+		'loadavg_allunread' => 40.0,
+		'loadavg_unreadreplies' => 40.0,
+		'loadavg_show_posts' => 40.0,
+		'loadavg_userstats' => 50.0,
+		'loadavg_bbc' => 60.0,
+		'loadavg_forum' => 85.0,
```
jdarwood007: @LiroyvH,

Do you have any feedback regarding my suggestions to change the load average detection defaults in SMF?
This is now based on "100%" meaning completely using the entire system across all cores.

Sesquipedalian: I suppose the real question is, how many CPUs did the old values assume? 

From what I can tell, the default values for these loadavg_* settings were introduced sometime during the pre-release development of SMF 2.0, so sometime before 2012. My memory is a little fuzzy about common web server CPUs back in those days, but I suspect that those numbers would have been chosen on the assumption that a decent-to-good server web server would have 2 CPU cores.

If that's a fair guess, then the old value for 'loadavg_auto_opt', would mean about 50% load, whereas the value for 'loadavg_forum' (i.e. the point at which the forum just gives up and refuses to run) would mean about 2000% load. Obviously, 2000% load is insanely overloaded, but since we're talking about the level at which SMF should assume that the server will not be able to run its thread at all, it actually makes sense.

On some modestly provisioned servers I have administered, it wasn't uncommon to see load averages hitting 200-300% of capacity. That was fine, really. It just meant that responses were a little slower because threads had to queue up to execute.

In light of all that, here's my tentative suggestions for the default values. I've assumed the old values were for 2 CPU cores. I've raised the relative values for some of the thresholds (`loadavg_search`, `loadavg_allunread`, `loadavg_unreadreplies`, and `loadavg_show_posts`) and lowered the relative values for the really high ones (`loadavg_userstats`, `loadavg_bbc`, and `loadavg_forum`). I have drastically lowered the threshold for `loadavg_userstats` in particular. That feature is really not worth the cost when the server is overloaded. Meanwhile, I've raised the threshold for search a fair bit; it's an important feature, even if its expensive.

```suggestion
		'loadavg_auto_opt' => 50.0,
		'loadavg_search' => 300.0,
		'loadavg_allunread' => 200.0,
		'loadavg_unreadreplies' => 200.0,
		'loadavg_show_posts' => 200.0,
		'loadavg_userstats' => 150.0,
		'loadavg_bbc' => 700.0,
		'loadavg_forum' => 1000.0,
```

I would also be very interested to get @LiroyvH's thoughts on this, as well as yours @jdarwood007.

jdarwood007: Those should be based on 100% being the max since we take the average and divide by CPU cores.  A 4-core system on Linux would have the ability to go up to 400% (100% * 4 cpus).  I based it on that.  It should be impossible to hit 1000% percent or even 150% with this new logic.

Sesquipedalian: No, the load average can, often does, exceed 100% capacity. When that happens, tasks are put in a queue to await processing. Rather, issues arise when that queue gets too large—meaning that the load is growing faster than the CPUs can work through it—because the waiting tasks will take so long to be processed that the system might time out. (That's fundamentally what happens in a DDOS, for example.) 

The sweet spot for efficiency is to be operating at just below 100% capacity, because then you are making good use of your resources. It's fine to spike above 100%—meaning that tasks get added to the queue to await execution—just as long as you don't _stay_ above 100% for a long time. 

It's sort of like having enough cashiers at the grocery store. Sometimes there's a rush of customers at once, and queues form at each till because there are more customers than there are cashiers to process them. That's okay as long as those queues don't grow too large and can be processed in a timely manner. It only becomes a problem if the queues keep growing.

Sesquipedalian (changes requested):
This all looks great, except that it would be good to cache the value of `Sapi::$cpu_count` rather than running the commands inside `Sapi::getCpuCount()` every time SMF executes. The value will very rarely change, after all.

I think it would be best to cache this value in a similar way to how we cache the regex that we use to match TLDs in URLs. Specifically, we save that value in the settings table so that we can refer to it as `Config::$modSettings['tld_regex']`, and then we refresh it periodically. 

In a similar fashion, we could record `Config::$modSettings['cpu_count']` in the settings table and then refresh it as part of daily maintenance.

If this approach is taken, I suggest adding a `bool $update = false` parameter to `Sapi::getCpuCount()` and then using logic similar to the logic in `Url::setTldRegex()`, such that the commands to find the number of CPUs are only executed if either (a) `Config::$modSettings['cpu_count']` is empty or (b) the `$update` parameter is true. Then the daily maintenance task would just need to make a one line call to `Sapi::getCpuCount(true);` in order to force an update of the value in the settings table.

Sesquipedalian (approved):

PR #8796, opened 2025-07-23
[2.1] Uses HTML5 date and time inputs for calendar events

### LINKED ISSUE #7630: Invalid date when posting calendar event with default french format

Sesquipedalian:
Drops the JavaScript date and time pickers we were using before and implements standard HTML5 date and time input fields instead. We can do this now that all the browsers we support have full support for `<input type="date">` and `<input type="time">`.

Fixes #7630


jdarwood007:
Should we remove this from 2.1 or leave it?  It does simplify things, but just asking if we remove this from the stable version.

Sesquipedalian:
Yeah, it's fine. It fixes the bug in #7630, and removing this stuff is harmless now that we can use `<input type="date">` and `<input type="time">`. The JavaScript date picker and time picker were only there because the support for those input types was not universal back in 2016. Removing them actually makes things more stable.

jdarwood007 (approved):
I approve of the code.  Any samples here should have been sending dates in using the standard format expected to ensure no issues with data processing.

Sesquipedalian:
Agreed. The JavaScript date and time pickers were not normalizing the input adequately, which was the cause of #7630.

PR #8801, opened 2025-07-24
[3.0] Implement File System handler

### LINKED ISSUE #5316: FTPS/SFTP support
SMF currently does not appear to support FTP over SSL (FTPS) nor SFTP ("FTP over SSH").
More and more providers are rejecting plain-text authentication. They may even demand the data channel to be encrypted as well and thus reject any non-encrypted connection.
And some others, though this is more rare, don't support FTP nor FTPS but only SFTP.

With SMF only supporting plain-text FTP, tools such as the Package Manager and installer may not be able to perform certain tasks such as changing permissions (chmod). 
It would be nice if we can at least get FTPS support. Having SFTP support as well would be nice.
Hence this feature request. :)

jdarwood007:
This introduces a file system handler for package manager, bringing in FTPS and SFTP support to our package manager operations.

FTPS is handled with PHP ftp library and SFTP is handled with Phpseclib (MIT License). This simplifies and centralizes the logic for determining how to perform file system operations by using the root as a substitution for the boarddir.

The test page has been changed from xml supported to json and most of the logic moved into admin.js.

The settings page now operates as config vars.  Not sure why it wasn't ever this way.

The legacy FtpConnection was moved to basically be a wrapper for the new file system caller.  I didn't want to remove it as it needs to remain backwards compatible.

The code for Maintenance was updated, but since none of our tools yet use it, the logic is only in theory based on matching changes.

Additionally, I had no luck getting FTP stuff to work before hand.  I made changes until things started to operate.  I also tried to support using ssl:// like we did before, but this never worked for me, so I forced the use of the dedicated FTPS logic.  If anyone is able to get it to work, we can better identify the file system handlings.

Finally, I know there is leagues's file system (MIT License) handler that does FTP(s), SFTP, Webdav and a few others.  I thought about using it, but the only additional library I see useful for SMF to support would be webdav.  If we want to redirect to using it, then we can.

Fixes #5316


Sources/Maintenance/Migration/v3_0/FileSystemHandler.php
```diff
@@ -0,0 +1,84 @@
...
+				continue;
+			}
+
+			$newSettings[$newKey] = Config::$modSettings[$oldKey];
+			$newSettings[$oldKey] = null;
+		}
+
+		// Hold on, do we have a 'ssl://' or 'ftps://' in the server name?
+		if (str_starts_with(Config::$modSettings['package_server'], 'ssl://') || str_starts_with(Config::$modSettings['package_server'], 'ftps://')) {
+			$server_addr = preg_replace('~^((ft|htt)ps?|ssl)?://~i', '', Config::$modSettings['package_server']);
```
live627, line 73: This regex also matches `://`. Is this intentional?

Sources/PackageManager/FileSystem/APIs/FtpSSL.php
```diff
@@ -0,0 +1,363 @@
...
+	}
+
+	/**
+	 * Changes a files attributes (chmod)
+	 *
+	 * @param int|string $chmod The value for the CHMOD operation
+	 * @param string $ftp_file The file to CHMOD
+	 * @return bool Whether or not the operation was successful
+	 */
+	public function changePermissions(string $filename, string $chmod): bool
```
live627, line 164: docblock mismatch

Sources/Maintenance/Migration/v3_0/FileSystemHandler.php
jdarwood007: It wasn't intended, but I am also now wondering, would it be harmful?  Ideally, we strip out any schema from the provided server address.

Themes/default/scripts/admin.js
```diff
@@ -629,4 +606,82 @@ function toggleBaseDir ()
...
+		this.testResults({
+			success: 0,
+			message: error.message
+		});
+	}
+}
+
+smf_fsTest.prototype.testResults = function (data)
+{
+	console.log(data);
```
MissAllSunday: Debug code

PR #8802, opened 2025-07-24
[3.0] Utf8EntityDecode task did not follow implementation of Background task

jdarwood007:
I noticed this error while using xdebug and step debugging.

### DIFF
--- a/Sources/Tasks/Utf8EntityDecode.php
+++ b/Sources/Tasks/Utf8EntityDecode.php
@@ -194,6 +194,30 @@ class Utf8EntityDecode extends BackgroundTask
 	}
 
 	/**
+	 * Adds a new instance of this task to the task list.
+	 */
+	protected function respawn(array $details, int $time = 0): void
+	{
+		Db::$db->insert(
+			method: 'insert',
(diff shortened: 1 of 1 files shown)


Sources/Tasks/Utf8EntityDecode.php
```diff
@@ -193,6 +193,30 @@ protected function decode(string $string): string
...
+				[
+					get_class($this),
+					json_encode($this->_details),
+					0,
+				],
+			],
+			keys: [],
+		);
+	}
+
```
Sesquipedalian: This doesn't need to be added here. We can just use the inherited method from the base class. Instead, the only change we need to make is to change `$this->respawn();` to `$this->respawn($this->_details);` on line 166 above.

Sesquipedalian (changes requested):

Sources/Tasks/Utf8EntityDecode.php
jdarwood007: That's what I get for not checking the reason and just changing the visibility to resolve the error.

Sesquipedalian (approved):

PR #8804, opened 2025-07-25
[3.0] Add additional fields to support better mail queueing

jdarwood007:
This resolves the mail queue needing to handle failed emails better by implementing 3 additional columns for tracking failures.

- This code does not yet make use of the `extra` column.  This is a placeholder should the need arise to add additional changes to the queue after a stable release.
- `tries` column will hold the number of times this was put back into the queue, if the max is reached, the email is removed from the queue.
- `next_try` column holds the time we will next try.  The select statement is updated to only grab stuff at the current time for before.
- `calculateNextTry()` sets up the next time we send by a exponential increase based on the number of tries we have done ensuring we increase our delay ever greater.


Sesquipedalian (changes requested):
There's no need for a migration step. The upgrader will automatically take care of normalizing the table (meaning, in this case, adding the new columns):

https://github.com/SimpleMachines/SMF/blob/fcd3cc4b599952ae605d264475dac936f58bfba5/Sources/Maintenance/Tools/Upgrade.php#L1096-L1103

jdarwood007:
Should we rely on that or ensure that the upgraded schema is in place and let the catchall be the final cleanup?  I had assumed it was the latter and would be our safety catch.

Sesquipedalian:
If the new table needs to exist in order to perform additional migration steps, then creating it is fine. But if that isn't the case, then leaving it until that final step makes good sense. Why have redundant code?

PR #8807, opened 2025-07-25
[3.0] Add constants for insert return mode

jdarwood007:
Just a little quality of life on the code here.  I couldn't remember what the 1 meant when looking at a query.

I didn't conduct an extensive review of all cases involving a return mode, so others may exist.

### DIFF
--- a/Sources/Db/APIs/MySQL.php
+++ b/Sources/Db/APIs/MySQL.php
@@ -339,7 +339,7 @@ class MySQL extends DatabaseApi implements DatabaseApiInterface
 	/**
 	 *
 	 */
-	public function insert(string $method, string $table, array $columns, array $data, array $keys, int $returnmode = 0, ?object $connection = null): int|array|null
+	public function insert(string $method, string $table, array $columns, array $data, array $keys, int $returnmode = DatabaseApi::INSERT_RETURN_MODE_OFF, ?object $connection = null): int|array|null
 	{
 		$connection = $connection ?? $this->connection;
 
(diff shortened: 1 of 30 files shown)


Sources/Db/DatabaseApi.php
```diff
@@ -29,6 +29,21 @@ abstract class DatabaseApi
...
+	 * Class constants
+	 *****************/
+
+	/**
+	 * Insert return modes.
+	 * 	- Off returns null.
+	 * 	- Single returns the last id.
+	 * 	- Multi returns all ids.
+	 * @var int
+	 */
```
Sesquipedalian:
```suggestion
	/**
	 * @var int
	 *
	 * Insert return modes.
	 * 	- Off returns null.
	 * 	- Single returns the last id.
	 * 	- Multi returns all ids.
	 */
```

Sources/Db/APIs/MySQL.php
```diff
@@ -339,7 +339,7 @@ public function free_result(object $result): bool
 	/**
 	 *
 	 */
-	public function insert(string $method, string $table, array $columns, array $data, array $keys, int $returnmode = 0, ?object $connection = null): int|array|null
+	public function insert(string $method, string $table, array $columns, array $data, array $keys, int $returnmode = DatabaseApi::INSERT_RETURN_MODE_OFF, ?object $connection = null): int|array|null
```
Sesquipedalian, line 342: Should probably make the same change in the signature for this method in SMF\Db\DatabaseApiInterface.

Sesquipedalian (changes requested):

Sesquipedalian (approved):

PR #8808, opened 2025-07-25
[3.0] Move to dedicated Hooks table

jdarwood007:
Sets up a dedicated hooks table for our integrations.  Hooks have become very important to our code base to ensure our mod authors are able to tap into the code where needed without needing to alter the source code.  As such we should build a more efficient setup for storing and maintaining those hooks.

- Deprecates using modSettings to store integrations.
- Deprecates `integrate_` prefix to hook names
- Adds compatibility with SMF 2.1 logic.
- Sets up some special handling logic for handling the pre_include (and possibly others) logic.
- Adds logic to the package manager to handle tagging all hooks registered by a mod to it to ensure removal on uninstall.

I have toyed with the idea of fully removing all `integrate_` from all files.  Not sure if it's relevant to this PR.  But since we are now storing hooks in their table, we don't need to have the prefix.  Looking for any feedback on that, I am thinking it should be done and let the IntegrationHooks logic handle restoring it for compatibility with SMF 2.1 where needed.

I need to do more testing here as this changes a lot and I have only done some basic mod tests.


Sources/IntegrationHook.php
```diff
@@ -166,100 +186,339 @@ public function execute(array $parameters = []): array
...
+		if (empty($id_hook)) {
+			$key = array_find_key(self::$hooks[$name], function ($val) use ($function, $file, $class, $is_object) {
+				return
+					$val['function'] === $function
+					&& $val['class'] === $class
+					&& $val['file'] === $file
+					&& $val['is_object'] === $is_object;
+			});
+		} else {
+			$key = array_find_key(self::$hooks[$name], fn($val) => $val['id_hook'] === $id_hook);
```
live627, line 447: Did we raise our requirements?

Sources/Maintenance/Migration/v3_0/MailQueueExentsion.php
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace SMF\Maintenance\Migration\v3_0;
+
+use SMF\Db\Schema;
+use SMF\Maintenance\Migration\MigrationBase;
+
+class MailQueueExentsion extends MigrationBase
```
live627: Extra file committed as a mistake?

Sources/PackageManager/PackageManager.php
```diff
@@ -1147,15 +1147,22 @@ public function install(): void
...
 					if (!isset(Utils::$context['ignore_hook_errors'])) {
 						Utils::$context['ignore_hook_errors'] = true;
 					}
 
 					if ($action['reverse']) {
-						IntegrationHook::remove($action['hook'], $action['function'], true, $action['include_file'], $action['object']);
+						IntegrationHook::unregister($action['hook'], 0, $action['function'], true, $action['include_file'], $action['object']);
 					} else {
-						IntegrationHook::add($action['hook'], $action['function'], true, $action['include_file'], $action['object']);
+						IntegrationHook::register($action['hook'], $action['function'], $action['include_file'], $action['class'], !empty($action['obect']), !empty($action['is_enabled']), $packageInfo['id']);
```
live627: Actually, this is a bool so no need for the empty check

https://github.com/SimpleMachines/SMF/blob/6b0f9915e128b5740c79136c13cb4b0ba7a69a3e/Sources/PackageManager/PackageUtils.php#L1229

```suggestion
						IntegrationHook::register($action['hook'], $action['function'], $action['include_file'], $action['class'], $action['object'], $action['is_enabled'], $packageInfo['id']);
```

Sources/Config.php
```diff
@@ -1178,22 +1181,24 @@ public static function reloadModSettings(): void
...
-
-			foreach ($pre_includes as $include) {
-				$include = strtr(trim($include), ['$boarddir' => self::$boarddir, '$sourcedir' => self::$sourcedir]);
-
-				if (file_exists($include)) {
-					require_once $include;
-				}
-			}
-		}
+		IntegrationHook::call('integrate_pre_include');
```
live627, line 1202: This logic must be preserved.

live627: For an example,, see

https://github.com/live627/smf-custom-forms/blob/bf6527ba66bd53a7a1c60e6734deeab20c603c2b/src/package-info.xml#L50

Sources/IntegrationHook.php
jdarwood007: No, I just didn't check that it's 8.4 only.  VScode doesn't warn me.  I should see if it can.  I will come up with an alternative approach.  There is a polyfill on the array_find docs.

Sources/Config.php
jdarwood007: I forgot, I never use pre_include.

What if we fix it in the package manager so that when we encounter "_include" ending when performing a hook install, it automatically moves from `function=""` to `file=""`? Would also have to add this to the `add()` (or just in subs-combat).  Then deprecate using function="" for 3.0?  It supports 2.1, but transitions it into a standard setup.

Sources/IntegrationHook.php
Sesquipedalian: This is fine. We already have a fallback function for it in Subs-Compat.php:

https://github.com/SimpleMachines/SMF/blob/6b0f9915e128b5740c79136c13cb4b0ba7a69a3e/Sources/Subs-Compat.php#L12084-L12095

Sources/PackageManager/PackageManager.php
jdarwood007: `$action` comes from the parsed XML data. To be safe, it should be empty to ensure that the data sends a boolean value.

Sources/Config.php
dragomano: In SMF 2.1, there are some hooks without prefix: `mention_insert_`, `pre_cache_quick_get`, `cache_put_data`, etc

jdarwood007: > In SMF 2.1, there are some hooks without prefix: `mention_insert_`, `pre_cache_quick_get`, `cache_put_data`, etc

You're right, I've added a workaround to ensure we update modSettings properly.  Everywhere else should be calling through the hook calls so that it will handle it blindly.

jdarwood007: > This logic must be preserved.

@live627 , I believe the workaround I just sent up should resolve this.  I detect when we are calling an include hook and that `file` is empty, but `function` is not, then push `function` into `file`, so the logic is handled the same.
Should fix it to work the same and consistently.  Future calls to our hooks registration for `*_include` hooks should be with `function` empty and `file` specified.

Sources/Maintenance/Migration/v3_0/Hooks.php
```diff
@@ -0,0 +1,58 @@
...
+declare(strict_types=1);
+
+namespace SMF\Maintenance\Migration\v3_0;
+
+use SMF\Config;
+use SMF\Db\DatabaseApi as Db;
+use SMF\Db\Schema;
+use SMF\Maintenance\Migration\MigrationBase;
+
+class Hooks extends MigrationBase
```
Sesquipedalian: This migration step is unnecessary. The upgrader already contains functionality to create any new tables:

https://github.com/SimpleMachines/SMF/blob/fcd3cc4b599952ae605d264475dac936f58bfba5/Sources/Maintenance/Tools/Upgrade.php#L1096-L1103

Table::normalize() will create the table if it doesn't exist:

https://github.com/SimpleMachines/SMF/blob/fcd3cc4b599952ae605d264475dac936f58bfba5/Sources/Db/Schema/Table.php#L109-L111

As a general rule, it is only necessary to explicitly create a new table during a migration step if you need the table to exist in order to perform other migration logic. In this case there is no further logic that uses the new table.

Sesquipedalian (changes requested):

Sources/IntegrationHook.php
```diff
@@ -662,13 +662,15 @@ final protected static function loadFile(string $file, bool $silent = false): bo
...
+		$path = Config::canonicalPath($path);
+
 		// Load the file if it can be loaded.
 		if (is_file($path)) {
 			require_once $path;
 		}
 		// No? Try a fallback to Config::$sourcedir.
 		else {
-			$path = Config::$sourcedir . '/' . $file;
+			$path = Config::canonicalPath(Config::$sourcedir . '/' . $file;
```
Oldiesmann: You're missing a `)` here

PR #8815, opened 2025-07-29
[3.0] MySQL doesn't support format=json on analyze until 8.3

jdarwood007:
Prior to 8.3 you get: ERROR 1235 (42000): This version of MySQL doesn't yet support 'FORMAT=JSON with EXPLAIN ANALYZE'

Support for this was added in 8.3: https://dev.mysql.com/blog-archive/new-json-format-for-explain/

Minor indentation cleanup included.

### DIFF
--- a/Sources/Actions/ViewQuery.php
+++ b/Sources/Actions/ViewQuery.php
@@ -131,19 +131,19 @@ class ViewQuery implements ActionInterface, Routable
 			}
 
 			echo '
-		<div id="qq', $q, '" style="margin-bottom: 2ex;">';
+			<div id="qq', $q, '" style="margin-bottom: 2ex;">';
 
 			if ($is_select_query) {
 				echo '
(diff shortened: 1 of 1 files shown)


Sources/Actions/ViewQuery.php
```diff
@@ -157,58 +157,70 @@ public function execute(): void
...
-					'EXPLAIN ' . ($vendor == 'PostgreSQL' ? '(ANALYZE, FORMAT JSON) ' : 'ANALYZE FORMAT=JSON ') . $select,
-				);
-			}
+				if ($vendor === 'MariaDB') {
+					$result = Db::$db->query('ANALYZE FORMAT=JSON ' . $select);
+				}
 
-			echo '
-		<pre>' . DebugUtils::highlightJson(Db::$db->fetch_row($result)[0]) . '</pre>';
+				if ($vendor == 'PostgreSQL') {
```
Sesquipedalian: Shouldn't this be an `elseif`?

Sesquipedalian (changes requested):

Sesquipedalian (approved):

PR #8830, opened 2025-08-08
[3.0] Regular users could not edit own profile

jdarwood007:
Permissions is an array, and we want to allow them if they have any of the permissions, not all.

To reproduce this, simply try to edit your "own profile" as a standard user with permissions granted to do so.

### DIFF
--- a/Sources/Actions/Profile/Main.php
+++ b/Sources/Actions/Profile/Main.php
@@ -1033,7 +1033,7 @@ class Main implements ActionInterface, Routable
 		}
 
 		if (isset($security_checks['permission'])) {
-			User::$me->isAllowedTo($security_checks['permission']);
+			User::$me->isAllowedTo($security_checks['permission'], any: true);
 		}
 
 		// Create a token if needed.
--- a/Sources/Menu.php
+++ b/Sources/Menu.php
@@ -451,7 +451,7 @@ class Menu implements \ArrayAccess
 			return false;
 		}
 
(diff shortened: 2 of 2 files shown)


Sources/Actions/Profile/Main.php
```diff
@@ -1033,7 +1033,7 @@ protected function securityChecks(): void
 		}
 
 		if (isset($security_checks['permission'])) {
-			User::$me->isAllowedTo($security_checks['permission']);
+			User::$me->isAllowedTo($security_checks['permission'], any: true);
```
Oldiesmann, line 1036: Should be `User::$me->isAllowedTo($security_checks['permission'], null, true);`

Sources/Menu.php
```diff
@@ -451,7 +451,7 @@ protected function enabledAndAllowed(array $menu_item): bool
 			return false;
 		}
 
-		return !(isset($menu_item['permission']) && !User::$me->allowedTo($menu_item['permission']));
+		return !(isset($menu_item['permission']) && !User::$me->allowedTo($menu_item['permission'], any: true));
```
Oldiesmann, line 454: Same thing here.

Sources/Actions/Profile/Main.php
jdarwood007: Its the same thing. https://www.php.net/manual/en/functions.arguments.php#functions.named-arguments

Oldiesmann: I wasn't aware that was a thing. Interesting.

Sesquipedalian (approved):

PR #8831, opened 2025-08-08
[3.0] Users couldn't toggle word censor if the admin allowed it

### LINKED ISSUE #8810: [2.1 & 3.0]: Users cannot turn off word censoring when admin allows them to
### Basic Information

When the admin enables the option to allow users to turn off word censoring, that option doesn't show up in the user's profile, whether looking at the options as an admin or as the user themselves.

### Steps to reproduce

1. Go to Admin -> Forum -> Posts and Topics -> Censored Words
2. Check the box next to the "Allow users to disable word censoring" option and save the settings.
3. Go to the look and layout section of any user's profile

### Expected result

The option to turn off word censoring should be there

### Actual result

No such option appears. 

Image

### Version/Git revision

2.1.6

### Database Engine

All

### Database Version

_No response_

### PHP Version

_No response_

### Logs

```shell

```

### Additional Information

In 2.1, the code to display the option is still there. I haven't checked 3.0 yet but I would assume similar code exists in that version as well.

https://github.com/SimpleMachines/SMF/blob/a0bb20705606c46804da848629e247b16077b883/Sources/Profile-Modify.php#L556-L571

jdarwood007:
Fixes #8810

### DIFF
--- a/Sources/Actions/Profile/ThemeOptions.php
+++ b/Sources/Actions/Profile/ThemeOptions.php
@@ -52,14 +52,6 @@ class ThemeOptions implements ActionInterface
 			Profile::$member->loadCustomFields('theme');
 		}
 
-		if (empty(Utils::$modSettings['allow_no_censored'])) {
-			$k = array_find_key(Utils::$context['theme_options'], fn($v) => is_array($v) && $v['id'] === 'show_no_censored');
-
-			if ($k !== null) {
-				unset(Utils::$context['theme_options'][$k]);
-			}
-		}
-
 		Utils::$context['page_desc'] = Lang::getTxt('theme_info', file: 'Profile');
 
 		Profile::$member->setupContext(
--- a/Sources/Actions/Profile/ThemeOptions.php
+++ b/Sources/Actions/Profile/ThemeOptions.php
@@ -52,6 +52,14 @@ class ThemeOptions implements ActionInterface
 			Profile::$member->loadCustomFields('theme');
 		}
 
+		if (empty(Utils::$modSettings['allow_no_censored'])) {
+			$k = array_find_key(Utils::$context['theme_options'], fn($v) => is_array($v) && $v['id'] === 'show_no_censored');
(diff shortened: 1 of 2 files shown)


Themes/default/Settings.template.php
```diff
@@ -67,6 +67,11 @@ function template_options()
 			'label' => Lang::getTxt('show_no_signatures', file: 'Profile'),
 			'default' => true,
 		),
+		array(
+			'id' => 'show_no_censored',
+			'label' => Lang::getTxt('show_no_censored'),
+			'default' => false,
+		),
```
Sesquipedalian:
```suggestion
		array(
			'id' => 'show_no_censored',
			'label' => Lang::getTxt('show_no_censored'),
			'default' => false,
			'enabled' => !empty(Config::$modSettings['allow_no_censored']),
		),
```

Sources/Actions/Profile/ThemeOptions.php
```diff
@@ -52,6 +52,14 @@ public function execute(): void
...
 		}
 
+		if (empty(Utils::$modSettings['allow_no_censored'])) {
+			$k = array_find_key(Utils::$context['theme_options'], fn($v) => is_array($v) && $v['id'] === 'show_no_censored');
+
+			if ($k !== null) {
+				unset(Utils::$context['theme_options'][$k]);
+			}
+		}
+
```
Sesquipedalian:
```suggestion
```

As said in #8833,  I don't think we should be unsetting the theme option. It is already the case that If the admin turns off the ability to disable the censor, then the censor will be applied and the theme option will be ignored no matter what its value is. But if the admin turns back on the ability to disable the censor, we should go back to whatever the user had chosen for themselves previously. We can't do that if we forcibly discard the user's chosen preference.

Sesquipedalian (changes requested):

Sesquipedalian (approved):

PR #8837, opened 2025-08-08
[3.0] Fix quotes in names for mention handling

### LINKED ISSUE #8820: [2.1 & 3.0]: Mentions don't work for names with apostrophes
### Basic Information

Image

### Steps to reproduce

1. Edit any name - add apostrophe
2. Try to mention that member

### Expected result

_No response_

### Actual result

_No response_

### Version/Git revision

70c0b0a80224d2bec667e8a7278bc8233372d4d8

### Database Engine

All

### Database Version

11.7.2-MariaDB

### PHP Version

8.3.22

### Logs

```shell

```

### Additional Information

I noticed that when a user edits their display name in the profile and adds an apostrophe, the value is saved differently:
- 2.1: `Test&#39;Name`
- 3.0: `Test'Name`

jdarwood007:
Fixes #8820

Big note, 2.1 handled names differently, so a bit of compatibility had to be added here.

If #8132 is accepted, then it needs to ensure it doesn't reintroduce this issue.

@Sesquipedalian I am not happy with how the fix had to come out for handling the entities issue.  If you have a better idea, please chime in or introduce an alternative PR.
I tested by changing a member's name to `Apasta'fee`.  In SMF 2.1, the database value is `Apasta&#39;fee` and in 3.0 it is `Apasta'fee`.  2.1 and 3.0 both handle a double quote the same by encoding it (i.e., `Double&quot;Quotes`)

### DIFF
--- a/Sources/Mentions.php
+++ b/Sources/Mentions.php
@@ -246,6 +246,7 @@ class Mentions
 		while ($row = Db::$db->fetch_assoc($request)) {
 			if (!isset($existing_mentions[$row['id_member']]) && stripos($body, static::$char . $row['real_name']) === false) {
 				$break = true;
+
 				// SMF 2.1 would store names containing single quotes escaped, 3.0 does not. Compatbility.
 				if (Config::$backward_compatibility && !isset($existing_mentions[$row['id_member']]) && stripos($body, Utils::htmlspecialchars(static::$char . $row['real_name'], ENT_QUOTES)) !== false) {
 					$row['real_name'] = Utils::htmlspecialchars($row['real_name'], ENT_QUOTES);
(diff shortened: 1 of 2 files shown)


Sesquipedalian:
> in 3.0 it is `Apasta'fee`. 

Hm. That wasn't an intentional change. I think it indicates a deeper problem that must be fixed first. I will look into it. In the meantime, I will leave this PR here.

jdarwood007:
I guessed that we had stopped mucking up entities somewhat and that we would see less of this in the future.  Seems now that during the transition of code, some `ENT_QUOTES` got dropped.

Sesquipedalian:
Yeah. I have a PR nearly ready.

PR #8842, opened 2025-08-08
[3.0] Improve admin search time

### LINKED ISSUE #8812: [3.0]: Admin search incredibly slow
### Basic Information

The admin search works in PostgreSQL, but is incredibly slow. I'm seeing page load times of between 15 and 20 seconds.

### Steps to reproduce

1. Install SMF on PostgreSQL
2. Search for a task/setting in the admin center

### Expected result

Page loads in a reasonable amount of time

### Actual result

Page takes 15+ seconds to load

### Version/Git revision

6b0f991

### Database Engine

PostgreSQL

### Database Version

_No response_

### PHP Version

_No response_

### Logs

```shell

```

### Additional Information

It appears it's bogging down after pulling text search config info. None of the queries are taking long, but the page load time always jumps dramatically following that query.

Image

jdarwood007:
Fixes #8812

The issue exists for a few reasons.
`lang:txtExists` checks for already loaded differently than `getTxt`.  `txtExists` only checks the `$filename` and doesn't take into account that when we called the `load` we set the checks on existing files per file, not per batch/group when using the `+`.  `getTxt` checks by seeing if the actual key exists, well at least the first one in the array.

The main solution is the first as @live627 pointed out and to call the `load` outside the loop.  Then remove the references to the `$file` parameter, which prevents the logic from including and loading the same language files a few thousand times.  This brought us from about 10 seconds to half a second.

A secondary fix that does offer a improvement (from about 10 seconds to 3.5 seconds without the previous fix) is when loading a batch/group of files, store that result as well.

### DIFF
--- a/Sources/Actions/Admin/Find.php
+++ b/Sources/Actions/Admin/Find.php
@@ -240,6 +240,10 @@ class Find implements ActionInterface
 		Utils::$context['page_title'] = Lang::getTxt('admin_search_results', file: 'Admin');
 		Utils::$context['search_results'] = [];
 
+		// Load all the languages first, not doing so slow.
+		// Do not use the $file param in any Lang calls.
+		Lang::load(implode('+', $this->language_files));
+
 		$search_term = strtolower(Utils::htmlspecialcharsDecode(Utils::$context['search_term']));
 
 		// Go through all the search data trying to find this text!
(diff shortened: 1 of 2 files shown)


Sesquipedalian:
The fact that this is happening at all means that there is a deeper issue with Lang::getTxt() itself that needs to be fixed. I am currently working on that problem.

PR #8845, opened 2025-08-10
[3.0] Various fixes

jdarwood007:
Working on another PR, but these issues popped up in static analysis.  Lots of issues in Subs-Compat due to a bad copy and paste when setting it up.

- Actions\BoardIndex
	- 'ref' was defined as 0 but used as an array elsewhere.
- Actions\Calendar
	- Attempted to call getCalendarGrid with 4 parameters, only 3 accepted
	- Ensure we have a string for strlen()
- Autoloader
	- Only variables can be passed by reference and array_pop requires a reference.
- Board
	- Do not need to break as we throw a fatal error;
- Subs-Compat
	- alert_mark tried to use $memID which doesn't exist.
	- url_to_iri tried to use $iri which doesn't exist.
	- Various functions called for wrong config vars.
	- Utils,User,Image was not "used"
	- getCalendarGrid tried to use the picker which no longer exists.
	- removeHolidays didn't fully qualify Calendar\Holiday
	- FindMember was removed as it was deprecated in 2.0.
	- recursiveBoards only variables can be passed by reference.
	- ShowDrafts and showProfileDrafts had wrong call for draft pms.
	- get_proxied_url is intended to return string
	- boardsAllowedTo should always return an array
	- call_helper does not always return
- Cache\APIs\Postgres
	- Improve docblock hint as it has been \Pgsql\Connection for some time
- Db\APIs\MySQL
	- $this->user and passwd didn't exist, but we expected it in some conditions.
- Db\APIs\Postgresql
	- Tried to use Db::$db instead of $this
- Db\DatbaseApi
	- Improve hints that we mixin the interface
- Permissions\UserPermissionSet
	- post_ban_permissions didn't exist under self or always.
- Theme
	- Ensure string passed to strlen
- Uuid
	- getHexV2 had unreachable code

### DIFF
--- a/Sources/Db/APIs/MySQL.php
+++ b/Sources/Db/APIs/MySQL.php
@@ -103,6 +103,20 @@ class MySQL extends DatabaseApi implements DatabaseApiInterface
 	 */
 	protected $supports_cte;
 
+	/**
+	 * @var string
+	 *
+	 * MySQL username.
+	 */
+	private $user;
+
+	/**
+	 * @var string
+	 *
+	 * MySQL password.
+	 */
+	private $passwd;
+
 	/****************
 	 * Public methods
 	 ****************/
@@ -2561,6 +2575,10 @@ class MySQL extends DatabaseApi implements DatabaseApiInterface
 			$this->connection,
 			'SET SESSION sql_mode = \'' . implode(',', $sql_mode) . '\'',
 		);
+
+		// We will need this for a autoreconnect later.
+		$this->user = $user;
+		$this->passwd = $passwd;
 	}
 
 	/**
@@ -2766,6 +2784,9 @@ class MySQL extends DatabaseApi implements DatabaseApiInterface
 				$this->error_backtrace('Undefined type used in the database query. (' . $matches[1] . ':' . $matches[2] . ')', '', false, __FILE__, __LINE__);
 				break;
 		}
+
+		// We reached a impossible location, but static anlaysis doesnt know that.
+		throw new \Exception();
 	}
 
 	/**
--- a/Sources/Subs-Compat.php
+++ b/Sources/Subs-Compat.php
@@ -174,7 +174,7 @@ if (!empty(SMF\Config::$backward_compatibility)) {
 	function ModifyAntispamSettings(bool $return_config = false): ?array
 	{
 		if (!empty($return_config)) {
-			return SMF\Actions\Admin\Attachments::getConfigVars();
+			return SMF\Actions\Admin\AntiSpam::getConfigVars();
 		}
 
 		return SMF\Actions\Admin\AntiSpam::call();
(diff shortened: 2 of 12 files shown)


Sources/Uuid.php
```diff
@@ -721,9 +721,6 @@ protected function getHexV2(array $input): string
 						// On non-POSIX systems, fall back to user ID because
 						// getmygid() returns nothing useful on non-POSIX systems.
```
Sesquipedalian:
```suggestion
```

Might as well get rid of these comment lines, too.

Sources/Subs-Compat.php
```diff
@@ -8002,7 +7996,7 @@ function ShowDrafts(int $member_id, int $reply_to = 0, int $draft_type = 0): boo
 	function showProfileDrafts(int $memID, int $draft_type = 0): void
 	{
 		if ($draft_type === 1) {
-			SMF\DraftPM::showInProfile($memID);
+			\SMF\PersonalMessage\DraftPM::showInProfile($memID);
```
Sesquipedalian:
```suggestion
			SMF\PersonalMessage\DraftPM::showInProfile($memID);
```

Sources/Subs-Compat.php
```diff
@@ -7985,7 +7979,7 @@ function DeleteDraft(int $id_draft, bool $check = true): bool
 	function ShowDrafts(int $member_id, int $reply_to = 0, int $draft_type = 0): bool
 	{
 		if ($draft_type === 1) {
-			return SMF\DraftPM::showInEditor($member_id, $reply_to);
+			return \SMF\PersonalMessage\DraftPM::showInEditor($member_id, $reply_to);
```
Sesquipedalian:
```suggestion
			return SMF\PersonalMessage\DraftPM::showInEditor($member_id, $reply_to);
```

Sources/Db/APIs/MySQL.php
```diff
@@ -2766,6 +2784,9 @@ protected function replacement__callback(array $matches, array $db_values, objec
 				$this->error_backtrace('Undefined type used in the database query. (' . $matches[1] . ':' . $matches[2] . ')', '', false, __FILE__, __LINE__);
 				break;
 		}
+
+		// We reached a impossible location, but static anlaysis doesnt know that.
+		throw new \Exception();
```
Sesquipedalian, line 2789:
```suggestion
		return '';
```

This will match SMF\Db\APIs\PostgreSQL::replacement__callback()

Sesquipedalian (changes requested):
Just a few adjustments needed.

Sources/Db/APIs/MySQL.php
jdarwood007: We don't need that.  Static analysis thinks that we don't return in all instances, because it doesn't understand that the switch is the last code in the function and will ensure we do something.
The throw helps it understand we will not be executing beyond that point, and would stop executing code if we reach it as a fatal error.

Sesquipedalian: Then please also change the PostgreSQL file to match this. I don't really care which one we use, just as long as we do it consistently.

jdarwood007: My bad for missing that, done.

I did look at using the match statement, but I don't see it improving anything.  It would involve a try/catch block and a significant return on the match (`return match[....]`) and anonymous functions to handle more complex checks we do.  Didn't seem like a win just to help static analysis know.

Sesquipedalian (approved):

PR #8847, opened 2025-08-14
[3.0] Implement AntiSpam System

jdarwood007:
This overhauls the entire AntiSpam system in SMF to provide a modular-based system, which should allow adding and removing methods for handling AntiSpam easily.  The design was set up so we could implement solutions like hCaptcha or Cloudflare Turnstile by simply dropping in the needed integration file and adding any needed vendor files.

A first note.  The AntiSpam system is currently broken in 3.0.  Due to it never verifying the verification questions, it only generates them.  I could send a separate PR just to fix that, but I fixed it in this to get my initial testing done.

The old "Verifier" was moved into the AntiSpam namespace and called "Verification" now.  It differs from the AntiSpam class itself in that it's a wrapper to provide the GUI to the standardized GenericControls, while the AntiSpam logic itself should be callable by any code that wants to implement its templating handling.

The BlankField Verification method may be better renamed to HiddenField, I haven't decided really.  I also thought about making this a required field for verification, because it's a very simple bot check, if they don't check that the CSS is hiding the field.

There is no restriction on having competing agents even if they offer the same deal (i.e., multiple types of CAPTCHA).

I want to work on getting reCAPTCHA V3 to work; we currently only support V2.  I don't want to do its enterprise captcha, but we may need to add support for that.

The upgrader does not yet handle converting from the old data to the new.

Looking for feedback to ensure I am going in the right direction here.  Things are named properly, and it looks like we can achieve the goal of extending the system.

Current TODO on this PR is

- [ ] Add upgrader steps
- [ ] Add support for reCAPTCHA v3
- [ ] Cleanup extra logic.
- [ ] Improve docs


live627:
> The BlankField Verification method may be better renamed to HiddenField, I haven't decided really. I also thought about making this a required field for verification, because it's a very simple bot check, if they don't check that the CSS is hiding the field.

Isn't it supposecd to be required to be blank anyway?

jdarwood007:
Yes it should always be empty.  A non-empty value is a failure.

PR #8838, opened 2025-08-08
[2.1] Fix quotes in names for mention handling

jdarwood007:
Fixes #8820

PR #8860, opened 2025-08-19
[3.0] Correctly handles quotes in mentions, plus other entity-related fixes

### LINKED ISSUE #8820: [2.1 & 3.0]: Mentions don't work for names with apostrophes

Sesquipedalian:
Fixes #8820 
Closes #8837

- The underlying issue that made this more difficult to fix in 3.0 was that member names were being saved with unescaped apostrophe characters rather than `&#39;` entities, but the mention code assumed the entities would be used. This issue affected not only member names, but various other things that are processed via `Utils::htmlspecialcharsRecursive()`. In previous versions of SMF, that function always forced the `ENT_QUOTES` flag to be used, but in 3.0 it was changed to allow the caller to set the flags and to use `ENT_COMPAT` by default, just like `Utils::htmlspecialchars()` does. This issue has been [fixed](9b3a25120e5ef0140f7c3c13ff7d36d66056604d) by specifying `ENT_QUOTES` in all calls to `Utils::htmlspecialcharsRecursive()` where it was expected.
    - **NOTE TO DEVS, TESTERS, AND EARLY ADOPTERS: This PR does not magically replace any unencoded apostrophe characters in your database with `&#39;` entities. You will need to fix them yourself.**

- Makes several other improvements to our handling of HTML entities, including:
    - [Adds $double_encode param to Utils::htmlspecialchars() and friends](https://github.com/SimpleMachines/SMF/commit/f37b4392ca0fa83e09c79101b58dafab4937dfb0)
    - [Adds SMF\Utils::htmlspecialcharsDecodeRecursive()](https://github.com/SimpleMachines/SMF/commit/14df9ad90c8bda6cff8fee7250847aad51f81d40)
    - [Simplifies entity processing methods](https://github.com/SimpleMachines/SMF/commit/78c3cc09df9d42ec28b4adb147801a5b8aeb8ccc). This one adds some polyfills for the various `grapheme_*` functions from the `intl` extension, and then uses them to simplify code in the `SMF\Utils::entity*` methods.

- This PR also [updates jquery.atwho.min.js to version 1.5.4](https://github.com/SimpleMachines/SMF/commit/f264aa3ec9c0729955a7f5ff3c4df4578c83c374). Doing this made it unnecessary to add the custom override that was added to mentions.js in #8838. We may want to update jquery.atwho.min.js in SMF 2.1 as well so that we can get rid of that customization there, too.

- Finally, based on all the changes above, [Correctly handles quotes in mentions](https://github.com/SimpleMachines/SMF/commit/352f3bd1527b8f7c1f9d58e38cfc5a04c9d0c451)


jdarwood007:
It's fine to update at.js for now.  @live627 commented in #8132 that the theme PR #7933 will replace this with something native.

PR #8865, opened 2025-08-22
[3.0] Remove deprecated mysqli_ping

### LINKED ISSUE #8659: mysql_ping deprecated
https://github.com/SimpleMachines/SMF/blob/c1038f6be04d29b87fc11f6d7be117cda40fc015/Sources/Db/APIs/MySQL.php#L773-L776

> The [mysqli_ping()](https://www.php.net/manual/en/mysqli.ping.php) function and [mysqli::ping()](https://www.php.net/manual/en/mysqli.ping.php) method are now deprecated as the reconnect feature was removed in PHP 8.2.0.

jdarwood007:
In PHP 8.4 this was deprecated
With the mysqlnd driver, reconnect is ignored
A SO post indicates that as of PHP 8.2, it throws an error when the connection is down

Another solution can be just to remove the ping entirely.

Fixes #8659

### DIFF
--- a/Sources/Db/APIs/MySQL.php
+++ b/Sources/Db/APIs/MySQL.php
@@ -731,7 +731,19 @@ class MySQL extends DatabaseApi implements DatabaseApiInterface
 	 */
 	public function ping(?object $connection = null): bool
 	{
-		return mysqli_ping($connection ?? $this->connection);
+		try {
+			mysqli_query($connection ?? $this->connection, 'DO 1');
+		} catch (\Exception) {
+			// If we specified a connection, don't connect.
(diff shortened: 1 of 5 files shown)


Sesquipedalian:
> Another solution can be just to remove the ping entirely.

Honestly, that's probably the best solution. `Db::$db->ping()` is called in only one location across SMF's entire codebase, and even there it's unnecessary.

jdarwood007:
Done.

The only reason for it would be, for example, a long-running upgrade, and the database decided to disconnect from us.  The ping would attempt to reconnect to continue

Sesquipedalian (approved):

PR #8869, opened 2025-08-23
[3.0] Allows members to see own login history and profile edit history

### LINKED ISSUE #7790: Members should be able to see login history and profile edits
Profile > Track User > Login

This page is only accessible if you have moderate_forum permissions.
This should operate where members can see this data.  It is useful information for members to know when a login occurred under a IP address they do not recognize.
This of course should not link them to track user/ip (unless they have moderate permissions.

Additionally,
This profile edit history should be visible to members.  With exception for additional groups which may contain member groups that are hidden (or hide them).  This is also useful information for a member to see.

Sesquipedalian:
Fixes #7790 

With this change, members can now see the history of logins and the history of profile edits for their own accounts. The permissions that allow a member to edit their login info are the same ones that allow the member to view their login history. Likewise, the permission that allows a member to edit their forum profile is the same one that allows the member to view the history of edits to their profile.

This PR also fixes a number of smaller issues I found in related code.


jdarwood007:
```
SMF\Lang::tokenTxtReplace(): Argument #1 ($string) must be of type string, int given, called in /Sources/Actions/Profile/Tracking.php on line 775
```

Not really related to the PR.  I would fix it by allowing `tokenTxtReplace` to accept an int and then cast it to string.

```
: Trying to access array offset on null in
/Sources/Actions/Profile/Tracking.php
on line
723
```
Fixed it by null coalescing that and the next line.  Interesting data, not edited:
Screen Shot 2025-08-24 at 8 44 03 AM

Additionally, if an admin changes their additional member groups and views the profile edits, they don't see this.  Admins should still see hidden/invisible group changes.

Sesquipedalian:
> ```
> SMF\Lang::tokenTxtReplace(): Argument #1 ($string) must be of type string, int given, called in /Sources/Actions/Profile/Tracking.php on line 775
> ```
> 
> Not really related to the PR. I would fix it by allowing `tokenTxtReplace` to accept an int and then cast it to string.

Fixed in 9c5e8d5


> ```
> : Trying to access array offset on null in
> /Sources/Actions/Profile/Tracking.php
> on line
> 723
> ```
> 
> Fixed it by null coalescing that and the next line. 

Fixed in 43f82cd.

> Interesting data, not edited: Screen Shot 2025-08-24 at 8 44 03 AM

Weird, but out of scope for this PR.

> Additionally, if an admin changes their additional member groups and views the profile edits, they don't see this. Admins should still see hidden/invisible group changes.

Fixed in 43f82cd.

jdarwood007 (approved):

PR #8884, opened 2025-08-26
Introduce smc_Request

live627:
* **AJAX refactor:** Replaced all jQuery `$.ajax` calls with the Fetch API and Promises.

  * Introduced a new `smc_Request` class with static `fetch` and `fetchXML` methods, supporting timeouts and credential options.
  * `getServerResponse`, `getXMLDocument`, and `sendXMLDocument` now use Fetch and return Promises, calling callbacks if provided.
* **Improved error handling:** Uses `Promise.reject` and catches network errors; callbacks receive `false` on failure.
* **Popup and menu handling:**

  * `reqOverlayDiv` and `smc_PopupMenu` now use `fetch` and native DOM methods instead of jQuery for content loading and event handling.


jdarwood007:
Can you also add the logic for handling JSON?
I had envisioned that we would replace our XML responses with JSON responses to take advantage of the more native handling of data.  One step at a time, but if this is in place here, those steps become easier to do.

sbulen:
Almost like we need a response type, to make it easier to avoid rendering http themes.

live627:
```js
		smc_Request.fetch(url)
			.then(res => res.json())
			.then(function(jsonArray) { /* ... */ });

**Comment by jdarwood007** [comment]
> ```js
> 		smc_Request.fetch(url)
> 			.then(res => res.json())
> 			.then(function(jsonArray) { /* ... */ });
> ```

Fair enough.  Mentioned because of the static function just for handling xml.

**Comment by live627** [comment]
right, that's because there is no quick built in function call

**Review by jdarwood007 — COMMENTED** [review]
The code looks good.

However, I can't complete a quick edit.  All I did was click quick edit and then save.  Never completes the save.

**Comment by live627** [comment]
I just got the  upgrader to work for me. I'll test this now.

**Comment by live627** [comment]
Oh I see what's going on. It sends an object, which this new userland code treats as JSON. We must send a string so it sends the proper header.

**Comment by jdarwood007** [comment]
That works but the Jump to never loads

**Comment by jdarwood007** [comment]
With your jumpto changes, can that JavaScript now be called with a defer?

Some JS goals I had were to move everything into defer and use the calls for inline JavaScript to send them to the bottom of the page.  Then look at doing noonce on all scripts and CSS.  One step at a time would be if JS is getting updated, can it be deferred, which makes the later steps easier to accomplish later.

File: Themes/default/Display.template.php [hunk 1]
```diff
@@ -288,63 +288,45 @@ function template_main()
...
-				sRestoreButtonLabel: \'', Lang::getTxt('quick_mod_restore', file: 'General'), '\',
-				sRestoreButtonImage: \'restore_selected.png\',
-				sRestoreButtonConfirm: \'', Lang::getTxt('quickmod_confirm', file: 'General'), '\',
-				bCanSplit: ', Utils::$context['can_split'] ? 'true' : 'false', ',
-				sSplitButtonLabel: \'', Lang::getTxt('quickmod_split_selected', file: 'General'), '\',
-				sSplitButtonImage: \'split_selected.png\',
-				sSplitButtonConfirm: \'', Lang::getTxt('quickmod_confirm', file: 'General'), '\',
-				sFormId: \'quickModForm\'
-			});';
+				var strips = [
```
jdarwood007: Should use `const` instead of `var`

live627:
I was planning to defer scripts in a future PR, but I decided to defer topic.js here 9ece101d190410f3ed8864ac807a0ca497ac6ae5 so you would be able to see that each individual call in the templates would need to be deferred as well.

jdarwood007:
That's fine.  As long as we are making the right progress towards where we can make that step happen more easily without rewriting them again.  The ultimate goal for our JS in my mind is that our libraries can be loaded in the head with async, and the calls that set up the logic in the app are deferred until the end.

Themes/default/Display.template.php
live627:
```suggestion
				const strips = [
```

Themes/default/Display.template.php
```diff
@@ -288,63 +288,45 @@ function template_main()
...
+					{ id: "moderationbuttons", display: "moderationbuttons_strip", varName: "oInTopicModeration" },
+					{ id: "moderationbuttons_mobile", display: "moderationbuttons_strip_mobile", varName: "oInTopicModerationMobile" }
+				];
+
+				for (var i = 0; i < strips.length; i++) {
+					var strip = strips[i];
```
live627:
```suggestion
				for (let i = 0; i < strips.length; i++) {
					const strip = strips[i];
```

jdarwood007:
Safe to review again or do you have more incoming?

live627:
go ahead

jdarwood007:
Two issue.

1. Jump to a board and it takes you to the wrong url
2. When you try to quick edit a post: Uncaught ReferenceError: oQuickModify is not defined

live627:
- I'll fix the quick modify error in #8886 if you don't mind.

jdarwood007 (approved):

PR #8886, opened 2025-08-26
[3.0] Update quick modify to support Ctrl+Enter

live627:


jdarwood007:
Just for clarification, why are we removing tabindex?

Sesquipedalian:
As a rule, we should keep the tabindex attribute itself, but we should set its value to 0 everywhere:

> Warning: You are recommended to only use 0 and -1 as tabindex values. Avoid using tabindex values greater than 0 and CSS properties that can change the order of focusable HTML elements ([Ordering flex items](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flexible_box_layout/Ordering_flex_items)). Doing so makes it difficult for people who rely on using keyboard for navigation or assistive technology to navigate and operate page content. Instead, write the document with the elements in a logical sequence.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/tabindex

frandominguezl:
> As a rule, we should keep the tabindex attribute itself, but we should set its value to 0 everywhere:
> 
> > Warning: You are recommended to only use 0 and -1 as tabindex values. Avoid using tabindex values greater than 0 and CSS properties that can change the order of focusable HTML elements ([Ordering flex items](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flexible_box_layout/Ordering_flex_items)). Doing so makes it difficult for people who rely on using keyboard for navigation or assistive technology to navigate and operate page content. Instead, write the document with the elements in a logical sequence.
> 
> https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/tabindex

According to the tabindex spec, there is no need to specify it to 0, since some elements have 0 as a default:

https://html.spec.whatwg.org/multipage/interaction.html#dom-tabindex

jdarwood007:
Screen Shot 2025-09-14 at 3 42 45 PM

I get an error 500 when I try to save the quick edit.
Also, get a js error for length.  The js error is because of the error 500 and it isn't handling the error properly.

The error 500 is related to a diff error.  This is synced up to the current revision, correct?

jdarwood007:
@live627,
You still working on this?

PR #8888, opened 2025-08-26
[3.0] Upgrader fixes

jdarwood007:
- Add a way for devs to catch which file is failing the file checker
- Table checking for existing indexes would fail if fixIndexName could not resolve it.
- Use isCandidate() on some upgrade steps to ensure we don't attempt to run it again.
- Ensure message id order when running EditHistory
- Search Results primary key was not cross database safe.

### DIFF
--- a/Sources/Maintenance/Migration/v3_0/SearchResultsPrimaryKey.php
+++ b/Sources/Maintenance/Migration/v3_0/SearchResultsPrimaryKey.php
@@ -16,10 +16,12 @@ declare(strict_types=1);
 namespace SMF\Maintenance\Migration\v3_0;
 
 use SMF\Db\DatabaseApi as Db;
+use SMF\Db\Schema;
 use SMF\Maintenance\Migration\MigrationBase;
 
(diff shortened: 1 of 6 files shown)


Sources/Maintenance/Migration/v3_0/SearchResultsPrimaryKey.php
```diff
@@ -16,10 +16,12 @@
 namespace SMF\Maintenance\Migration\v3_0;
 
 use SMF\Db\DatabaseApi as Db;
+use SMF\Db\Schema;
 use SMF\Maintenance\Migration\MigrationBase;
 
 class SearchResultsPrimaryKey extends MigrationBase
 {
+	private static array $columns = ['id_search', 'id_topic', 'id_msg'];
```
Sesquipedalian: Please run PHP-CS-Fixer.

Sesquipedalian (changes requested):

Sesquipedalian (approved):

PR #8891, opened 2025-08-27
[3.0] code tag changes

live627:
- wrap code in preformatted html tags
- do the same in the sceditor
- bonus: fix php tag processing in the sceditor
- add buttons for selecting and expanding code blocks with js; don't output them from php
- remove smfSelectText( ) as it is unused now

### DIFF
--- a/Sources/BBCode/Code1.php
+++ b/Sources/BBCode/Code1.php
@@ -39,7 +39,7 @@ class Code1 extends BBCode
 	/**
 	 *
 	 */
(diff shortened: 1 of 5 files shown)


ThistleSifter:
For semantics and accessibility, what about replacing the div header with figure/figcaption?

```html
<figure>
    <figcaption>PHP code: <button>Select</button> <button>Expand></button></figcaption>
    <pre><code>$foo</code></pre>
</figure>
```

jdarwood007:
@live627 This is in draft.  Is that intended?

live627:
Superseded by https://github.com/SimpleMachines/SMF/pull/9332

PR #8913, opened 2025-09-04
[2.1] Fixes bug showing guest avatar on board index

### LINKED ISSUE #8572: [2.1]:  guest avatar on the home page
### Basic Information

Hello, If guest can send messages, the avatar on the home page is taken from the images in the topic (random) 
https://www.simplemachines.org/community/index.php?msg=4188316

### Steps to reproduce

1.
2.

### Expected result

_No response_

### Actual result

_No response_

### Version/Git revision

 2.1.4

### Database Engine

All

### Database Version

_No response_

### PHP Version

8.1

### Logs

```shell

```

### Additional Information

_No response_

Sesquipedalian:
Fixes #8572


jdarwood007:
This shouldn't do anything.  It uses the member table instead of messages.  But we bring in the members table by using the reference from the messages table id_member.

Sesquipedalian:
One would think that it doesn't matter, but it does. When we joined on `am.id_member = m.id_member`, we ended up joining some arbitrary row from the attachments table where `am.id_member` was 0. When we join on `am.id_member = mem.id_member`, no attachment row ends up being joined at all. This is because `mem.id_member` ends up being null for guests when we try to join the row from the members table on `mem.id_member = m.id_member`. That in turn means that `mem.id_member` doesn't match the `am.id_member` value of any row in the attachments table, and so no attachment row is joined.

PR #8917, opened 2025-09-04
[3.0] Always use canonical path when including files

### LINKED ISSUE #8293: [3.0]: Config quirks

Sesquipedalian:
On Windows servers, the paths that were used to include files often contained a mix of backslashes and forward slashes, because our code typically uses forward slashes as per Unix/Linux conventions, but saved paths like `Config::$sourcedir` typically used backslashes as per Windows conventions.

This has been known to cause issues in situations where the path of an included file has to be checked for whatever reason. For example, #8293 documents a case where Settings.php is included over and over due to this issue. 

This PR fixes the problem by always using the correct directory separator character in our paths when including files:
- In cases where the path to be included was a hard-coded string, this could be achieved most efficiently by simply replacing literal `/` characters in paths with the `DIRECTORY_SEPARATOR` constant. 7b9a16d implements this.
- In cases where the path to be included was determined programmatically, it was easier to implement a function to canonicalize the path just before including the file. b3be59d implements this.

The new `Config::canonicalpath()` method added in b3be59d is a wrapper for `realpath()` that is also able to handle paths that don't exist.

Fixes #8293.

### DIFF
--- a/Sources/Actions/Admin/Logs.php
+++ b/Sources/Actions/Admin/Logs.php
@@ -156,7 +156,7 @@ class Logs implements ActionInterface
 		];
 
 		if (!empty(self::$subactions[$this->subaction][0])) {
-			require_once Config::$sourcedir . '/' . self::$subactions[$this->subaction][0];
+			require_once Config::canonicalPath(Config::$sourcedir . '/' . self::$subactions[$this->subaction][0]);
 		}
 
 		$call = \is_string(self::$subactions[$this->subaction][1]) && method_exists($this, self::$subactions[$this->subaction][1]) ? [$this, self::$subactions[$this->subaction][1]] : Utils::getCallable(self::$subactions[$this->subaction][1]);
--- a/Sources/Config.php
+++ b/Sources/Config.php
@@ -1190,7 +1190,7 @@ class Config
 				$include = strtr(trim($include), ['$boarddir' => self::$boarddir, '$sourcedir' => self::$sourcedir]);
 
 				if (file_exists($include)) {
-					require_once $include;
+					require_once self::canonicalPath($include);
 				}
 			}
 		}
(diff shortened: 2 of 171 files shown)


Sesquipedalian:
@sbulen, if you can test this, that would be appreciated. In particular, I am looking for confirmation that `Config::canonicalpath()` works as expected on Windows, since I don't have a Windows machine available to test on.

Sources/Config.php
```diff
@@ -2845,6 +2850,129 @@ public static function checkCron(): void
...
+	 *    - If false, relative paths are processed as given.
+	 *    Default: false.
+	 * @param bool $real Whether to get the real path for existing files. This
+	 *    can be set to false if the caller wants to canonicalize a hypothetical
+	 *    path without any possibility of the real file structure interfering
+	 *    with the result.
+	 *    Default: true.
+	 * @return string The canonical file path.
+	 */
+	public static function canonicalpath(string $path, string|bool $base_dir = false, bool $real = true): string
```
jdarwood007: Following naming standards, shouldn't this be `canonicalPath`?

Sources/Actions/Admin/Logs.php
```diff
@@ -156,7 +156,7 @@ public function execute(): void
 		];
 
 		if (!empty(self::$subactions[$this->subaction][0])) {
-			require_once Config::$sourcedir . '/' . self::$subactions[$this->subaction][0];
+			require_once Config::canonicalpath(Config::$sourcedir . '/' . self::$subactions[$this->subaction][0]);
```
jdarwood007: Just nitpicking a random example.  I'm debating whether it would look better like that or as
```
require_once Config::$sourcedir . DIRECTORY_SEPARATOR . Config::canonicalpath(self::$subactions[$this->subaction][0]);
```
Since we should only need to sanitize the input received from the sub-action itself, knowing that we have already cleaned `$sourcedir`.

jdarwood007: Another quick idea is we could do
```
function buildPath(...$args)
{
	return Config::canonicalpath(implode(DIRECTORY_SEPARATOR, $args));
}
```
Then a simple call to it:
```
Config::buildPath([Config::$sourcedir, self::$subactions[$this->subaction][0]));
```

I've seen it in other languages I've worked with where you provide the directory names and it handles the rest.

Sesquipedalian: > Just nitpicking a random example. I'm debating whether it would look better like that or as
> 
> ```
> require_once Config::$sourcedir . DIRECTORY_SEPARATOR . Config::canonicalpath(self::$subactions[$this->subaction][0]);
> ```
> 
> Since we should only need to sanitize the input received from the sub-action itself, knowing that we have already cleaned `$sourcedir`.

Not if the relative path in `self::$subactions[$this->subaction][0]` contains directory separators itself.

Sesquipedalian: > Another quick idea is we could do
> 
> ```
> function buildPath(...$args)
> {
> 	return Config::canonicalpath(implode(DIRECTORY_SEPARATOR, $args));
> }
> ```
> 
> Then a simple call to it:
> 
> ```
> Config::buildPath([Config::$sourcedir, self::$subactions[$this->subaction][0]));
> ```
> 
> I've seen it in other languages I've worked with where you provide the directory names and it handles the rest.

It's a good idea, sure.

Sources/Config.php
Sesquipedalian: Meh. We could, I suppose. I modelled that name directly on `realpath()`. If it feels more consistent to you to capitalize the `P`, sure, we can do that.

Sources/Actions/Admin/Logs.php
jdarwood007: > > Since we should only need to sanitize the input received from the sub-action itself, knowing that we have already cleaned `$sourcedir`.
> 
> Not if the relative path in `self::$subactions[$this->subaction][0]` contains directory separators itself.

It's why I had it wrapped to use the `canonicalpath`, but left the rest.
Not really an issue, just me nitpicking at styling, and wanted to throw it out there if it was liked better.

Sources/Config.php
jdarwood007: Won't force it, just a comment that we seemed to be using camelCasing for our naming conventions (or at least trying to go that way) and wanted to ensure if it was an oversight, we address it.

jdarwood007 (approved):
Looks good.  Made some comments about coding styles, which don't impact anything if we decide to not do them.

Sources/Actions/Admin/Logs.php
Sesquipedalian: Oh, I see. Well, since the same `str_replace()` calls will be made either way, it won't make any performance difference, so it's just a matter of aesthetics. Personally, I find the `/` character is easier to read, so I only manually replaced in with `DIRECTORY_SEPARATOR` where necessary.

PR #8924, opened 2025-09-05
[3.0] Gets the correct number of CPUs in SMF\Sapi::getCpuCount()

Sesquipedalian:
Before this, it always returned 1.


jdarwood007:
Well, that was stupid of me...

Sesquipedalian:
I've done it before too.

PR #8925, opened 2025-09-06
[2.1] Add option to open board redirect in a new tab

MisterVector:
This PR adds an option to open board redirects in a new tab.


jdarwood007:
2.1 is feature locked and we are not accepting features for it.
It could be considered for 3.0 but would need to be approved.  However, if this were added for 3.0, instead of a new specific column, I would add an `options` column instead and store this as a JSON array in it, allowing extensibility of the boards without needing to muck up too many queries.

PR #8929, opened 2025-09-07
[3.0] Fix Fatal error: Cannot redeclare grapheme_strstr() while installing SMF

MisterVector:
I was going to install SMF 3.0 and ran into a function redeclaration. This PR fixes the following error:

Fix Fatal error: Cannot redeclare grapheme_strstr()


jdarwood007:
Can you sign off on the commit?

MisterVector:
Sorry about that! It's signed now.

PR #8937, opened 2025-09-09
feat(utils): support character classes in regex builder

live627:
- Remove phantom leaf nodes from trie construction
- Collapse single-character alternations into `[...]` character classes
- Avoid redundant `(?> … )` wrapping for character classes

### Before
```
(?>HAVING|GROUP|JOIN|KEY|PR(?>OCEDURE|IMARY)|SE(?>LECT|T)|A(?>DD|L(?>TER|L)|N(?>D|Y)|S(?>C|))|B(?>ETWEEN|ACKUP|Y)|C(?>URRENT_(?>DATE|TIME(?>STAMP|))|HECK|ASE|O(?>NSTRAINT|LUMN)|R(?>EATE|OSS))|D(?>ATABASE|ISTINCT|ROP|E(?>FAULT|LETE|SC))|E(?>LSE|ND|X(?>CEPT|ISTS))|F(?>OREIGN|ROM|ULL)|I(?>F|N(?>SERT|DEX|NER|T(?>ERSECT|O)|)|S)|L(?>EFT|I(?>MIT|KE))|N(?>ULL|OT)|O(?>UTER|N|R(?>DER|))|R(?>EFERENCES|IGHT|O(?>LLBACK|WNUM))|T(?>RUNCATE|ABLE|OP)|U(?>PDATE|NI(?>QUE|ON))|V(?>ALUES|IEW)|W(?>HERE|ITH))
```
###  After

```
(?>HAVING|GROUP|JOIN|KEY|PR(?>OCEDURE|IMARY)|SE(?>LECT)|ADD|SC|L(?>TER)|N[DY]|B(?>ETWEEN|ACKUP)|C(?>URRENT_(?>TIMESTAMP|DATE)|HECK|ASE|O(?>NSTRAINT|LUMN)|R(?>EATE|OSS))|D(?>ATABASE|ISTINCT|ROP|E(?>FAULT|LETE|SC))|E(?>LSE|ND|X(?>CEPT|ISTS))|F(?>OREIGN|ROM|ULL)|I(?>N(?>SERT|DEX|NER|T(?>ERSECT)))|L(?>EFT|I(?>MIT|KE))|N(?>ULL|OT)|O(?>RDER|UTER)|R(?>EFERENCES|IGHT|O(?>LLBACK|WNUM))|T(?>RUNCATE|ABLE|OP)|U(?>PDATE|NI(?>QUE|ON))|V(?>ALUES|IEW)|W(?>HERE|ITH))
```

Observe the character class `/*...*/ ADD|SC|L(?>TER)|N[DY]|B(?>ETWEEN  /*...*/`

### DIFF
--- a/Sources/Utils.php
+++ b/Sources/Utils.php
@@ -1083,8 +1083,6 @@ class Utils
 
 				$node = &$node[$char];
 			}
-
-			$node[''] = '';
 		}
 
 		// This recursive closure turns the trie into a regular expression.
@@ -1097,6 +1095,7 @@ class Utils
 
 			$regex = [];
 			$length = 0;
+			$single_chars = [];
 
 			foreach ($trie as $key => $value) {
 				$key_regex = preg_quote((string) $key, $delim);
@@ -1107,24 +1106,34 @@ class Utils
 				} else {
 					$sub_regex = $trie_to_regex($value, $delim);
 
-					if (\count(array_keys($value)) == 1) {
-						$new_key_array = explode('(?' . '>', $sub_regex);
+					if (\count($value) == 1) {
+						$new_key_array = explode('(?>', $sub_regex);
 						$new_key .= $new_key_array[0];
-					} else {
-						$sub_regex = '(?' . '>' . $sub_regex . ')';
+					} elseif (!str_starts_with($sub_regex, '[') && !str_ends_with($sub_regex, ']')) {
+						$sub_regex = '(?>' . $sub_regex . ')';
 					}
 				}
+				$str = $key_regex . $sub_regex;
 
 				if ($depth > 1) {
-					$regex[$new_key] = $key_regex . $sub_regex;
+					// Collect single-char leaves for character class
+					if (empty($value) && mb_strlen($str, $encoding) === 1 && \count($trie) !== 1) {
+						$single_chars[] = $str;
+					} else {
+						$regex[$new_key] = $str;
+					}
 				} else {
-					if (($length += \strlen($key_regex . $sub_regex) + 1) < $max_length || empty($regex)) {
-						$regex[$new_key] = $key_regex . $sub_regex;
+					if (($length += \strlen($str) + 1) < $max_length || empty($regex)) {
+						$regex[$new_key] = $str;
 						unset($trie[$key]);
 					} else {
 						break;
 					}
 				}
+			}
+
+			if ($single_chars !== [] && $regex === []) {
+				return '[' . implode('',  $single_chars) . ']';
 			}
 
 			// Sort by key length and then alphabetically


Sesquipedalian:
Unfortunately, this doesn't work.

1. In the new regex, `SE(?>LECT|T)` no longer appears, but rather only `SE(?>LECT)`. Thus, the regex no longer matches `SET`, but only `SELECT`.

2. The empty leaf nodes are there for a reason. When they are inserted into an atomic group, it is because they need to be there in order to match a whole word.

	Consider this subsection of the regex as compiled by the existing code: `I(?>F|N(?>SERT|DEX|NER|T(?>ERSECT|O)|)|S)`. For ease of reading, let's break it up across multiple lines like so:
    ```php
	$regex = 'I' .
	'(?>' .
		'F' .
		'|' .
		'N' .
		'(?>' .
			'SERT' .
			'|' .
			'DEX' .
			'|' .
			'NER' .
			'|' .
			'T' .
			'(?>' .
				'ERSECT' .
				'|' .
				'O' .
			')' .
			'|' .
			''. // <-- empty leaf node
		')' .
		'|' .
		'S' .
	')';
	```

	Then let's test that regex against a string like so:

	```php
	var_export((bool) preg_match('/' . $regex . '/', 'IN A BIT'));
	```

	This will output `true` because it matched the word `IN`. If we step through the regex's logic, it starts by matching the initial `I` in the string. Then it enters the first level atomic group and tries to match `F`. That fails, so next it tries `N`. That matches, so it enters the second level atomic group and tries to match `SERT`. That fails, so it tries `DEX`. That fails, so it tries `NER`. That fails, so it tries `T`. That fails, so it skips the third level atomic group and moves on to the last option of the second level atomic group, which is just a zero-length string. That matches after `IN`, and so an overall match for `IN` has been found.

	Now let's try a version without that empty leaf node:
   ```php
	$regex = 'I' .
	'(?>' .
		'F' .
		'|' .
		'N' .
		'(?>' .
			'SERT' .
			'|' .
			'DEX' .
			'|' .
			'NER' .
			'|' .
			'T' .
			'(?>' .
				'ERSECT' .
				'|' .
				'O' .
			')' .
			// '|' .
			// '' .
		')' .
		'|' .
		'S' .
	')';

	var_export((bool) preg_match('/' . $regex . '/', 'IN A BIT'));
	```

	This will output `false` because it fails to match the word `IN`. The logic of this regex is the same as before, except there is no option to match a zero-length string after `IN`. Lacking that option, the regex concludes that it can't match any of the options that start with `IN`, and so it moves on to trying `IS`, which obviously fails.

Sesquipedalian:
It's also worth noting that adding character classes in place of atomic groups won't really accomplish much. A character class _is_ an atomic group. The notation is different, that's all.

PR #8941, opened 2025-09-11
[3.0] Fix a fatal error due to numerical keys

jdarwood007:
I was receiving the following error:
```
( ! ) Fatal error: Uncaught TypeError: mb_str_split(): Argument #1 ($string) must be of type string, int given in /Sources/Utils.php on line 1076
```

This occurred while bringing in the `$itemcodes` from the BBC parser.  After some debugging, I realized that `$normalized_strings` had a numerical index of 0 instead of a string of 0.

Some more debugging later, I found out how to reproduce it in a very simple statement: `var_dump(["0" => 1]);`

I finally discovered the reason after locating this note in the PHP documentation:
> [String](https://www.php.net/manual/en/language.types.string.php)s containing valid decimal [int](https://www.php.net/manual/en/language.types.integer.php)s, unless the number is preceded by a + sign, will be cast to the [int](https://www.php.net/manual/en/language.types.integer.php) type. E.g. the key "8" will actually be stored under 8. On the other hand "08" will not be cast, as it isn't a valid decimal integer.

https://www.php.net/manual/en/language.types.array.php#language.types.array.syntax

This is unrelated to #8937, but it may conflict with it.

### DIFF
--- a/Sources/Utils.php
+++ b/Sources/Utils.php
@@ -1049,7 +1049,7 @@ class Utils
 
 		// Optimizing is faster when we sort by length.
 		asort($normalized_strings);
-		$strings = array_keys($normalized_strings);
+		$strings = array_map('\strval', array_keys($normalized_strings));
 
 		// Can we trim common characters from the end?
 		$trailing = '';
--- a/Sources/Utils.php
+++ b/Sources/Utils.php
@@ -1049,7 +1049,7 @@ class Utils
 
 		// Optimizing is faster when we sort by length.
 		asort($normalized_strings);
-		$strings = array_map('\strval', array_keys($normalized_strings));
+		$strings = array_map('strval', array_keys($normalized_strings));
 
 		// Can we trim common characters from the end?
 		$trailing = '';


live627:
We should avoid unnecessary allocations in potentially hot code. Would adding the`strval` suffice on 


```php
				$s = (string) $str;
```

jdarwood007:
It would not.  PHP appears to do a form of try cast on the key, and if it can become an int, it does.

Sources/Utils.php
```diff
@@ -1049,7 +1049,7 @@ public static function buildRegex(array $strings, ?string $delim = null, bool $r
 
 		// Optimizing is faster when we sort by length.
 		asort($normalized_strings);
-		$strings = array_keys($normalized_strings);
+		$strings = array_map('\strval', array_keys($normalized_strings));
```
live627:
```suggestion
		$strings = array_map('strval', array_keys($normalized_strings));
```

live627 (commented):
I needed to enable strict types in my test file to get that error.

Sources/Utils.php
jdarwood007: Funny thing is, I don't know if it matters or not.  I don't know if the zend optimizations apply on array_map.

live627: I don't know either , but I do know that this backslash  is unnecessary because string callbacks are always in the global namespace.

live627 (approved):

PR #8949, opened 2025-09-19
Rewrite our editor integration

### LINKED ISSUE #7636: Selection lost in editor after bbc button clicked
#### Description
Selection lost in editor after bbc button clicked.    

Workaround is to re-select the text after each click.  Annoying, but it works.

### Steps to reproduce
1. Select text.
2. Press the bold bbc button
3. Press the italics bbc button; there is no text within the italics bbc tags

- Version/Git revision: 2.1.3
- Database Type: MySQL
- Database Version: 8.0
- PHP Version: 8.1

Reported in forum:
https://www.simplemachines.org/community/index.php?topic=584335.0

### LINKED ISSUE #7681: Can't load more styles to sceditor
#### Description
When toggling the source view in the sceditor, it is possible to load a provided file:
https://www.sceditor.com/documentation/options/#style
https://github.com/SimpleMachines/SMF/blob/5bc137dabe88aada50f2b450957f169e058f224a/Sources/Subs-Editor.php#L1912

We are loading _jquery.sceditor.default.css_ and it's fine. However, this is a limitation for custom themes using multiple files (css variables, variants, dark mode, etc), it's currently not possible to load multiple files or pass data attributes to the iframe (so it can be targeted inside a single file).

live627:
Completely rewrite  our integration in the style of SCEditor plugin.

Custom SCEditor options:
   - `commandsWithDropdown`: Identifies buttons that use dropdown menus.
   - `textOnlyCommands`: Configures buttons to display text without icons.
   - `commandsWithText`: Configures buttons to show text alongside icons.

New Hooks:
- Hook: `integrate_sceditor_locale`
- Parameters:
  - `array &$translation_map`: Reference to the array of localization strings.


jdarwood007:
Does this rewrite break compatibility with 2.1 usage of SC Editor if they are using the standard methods to create a rich text editor?

live627:
No; any existing parameter expected by `Editor::create()` remain unchanged and still work as expected.

jdarwood007:
@live627 All functions are working correctly, but the UI looks a bit off.

1. Empty `<div>` causes a return, breaking the BBC buttons into two rows
2. There isn't a rounding box around the BBC buttons/smileys, which indicates the UI area.
3. When in the editor mode, the background color is the color of the parent div, which makes it less noticeable that it's an editor box
4. Smiley images do not change the mouse pointer to indicate a clickable.

Screen Shot 2025-09-28 at 6 08 50 PM

live627:
> 1. Empty `<div>` causes a return, breaking the BBC buttons into two rows

That's how it was in very early 2.1 alpha circa 2012 when @emanuele45 (or @Spuds not too sure on who) was in charge of the editor. Kinda annoyed me that SCEditor had no way of doing so and couldn't figure out a way to patch that in in time for the big 2.1 release date.

Sources/Editor.php
```diff
@@ -373,6 +498,180 @@ public static function getMessageIcons(int $board_id): array
...
+					'description' => Lang::getTxt('font_color', var: 'editortxt'),
+				],
+				[],
+				[
+					'code' => 'removeformat',
+					'description' => Lang::getTxt('remove_formatting', var: 'editortxt'),
+				],
+			],
+			[
+				[
```
jdarwood007, line 597:
```suggestion
				],
				[
```

Removes the extra group, letting the flex box handle it.

Sources/Editor.php
```diff
@@ -490,188 +791,24 @@ protected function buildButtons(): void
...
-			],
-		];
-
-		if (empty(Config::$modSettings['disable_wysiwyg'])) {
-			self::$bbc_tags[\count(self::$bbc_tags) - 1][] = [
-				'code' => 'removeformat',
-				'description' => Lang::getTxt('remove_formatting', var: 'editortxt'),
-			];
-		}
-
```
jdarwood007, line 594: Is there any intention to add back the remove formatting option which appears to be used when the editor is disabled?

jdarwood007:
Otherwise if you have no commits left, it looks good.  The two rows thing just looks off in my mind and should be fixed to use the flexbox.

Sources/Editor.php
live627: The logic to show or hide it was backwards; should be fixed n w in 5e643a94ccaff81b8c8d0b8a9c2b4898b21b7c6e

Sources/Editor.php
live627: This actually IS what triggers the breaking divider . Also, it'll probably break mods that expect the array structure to be like this. Although I've never seen any that specifically look for this index; instead, they rely on `count()`.

live627:
I might just undo the double row thing since you probably won't be the only one to complain about it.

jdarwood007:
Could ajdust the css to do it from a ".parent > div" scenario instead.

PR #8954, opened 2025-09-20
[3.0] Admin info files data should be medium text

### LINKED ISSUE #8952: [3.0]: "Data too long for column data ..." when installing
### Basic Information

When installing SMF, after getting past the admin account creation, you get a blank screen with an error saying "Data too long for column data at row 1".

### Steps to reproduce

1. Attempt to install SMF 3
2.

### Expected result

Installation succeeds

### Actual result

Blank screen with error message on step 5

### Version/Git revision

3c73aee

### Database Engine

MySQL

### Database Version

9.3.0

### PHP Version

8.4.12

### Logs

```shell

```

### Additional Information

After further investigation,  I determined that this is due to the admin version check JS - the large number of files in SMF 3 makes that JS exceed the maximum length of 65535 characters for the "data" column in the admin_info_files table. Changing the data column to "mediumtext" in the schema file allows the installation to succeed, though I don't know if there's' a better solution.

jdarwood007:
Fixes #8952

@Oldiesmann and @MisterVector ,
Can you test this?  You can test either a fresh install or, after pulling in this change, run upgrade.php.

### DIFF
--- a/Sources/Db/Schema/v3_0/AdminInfoFiles.php
+++ b/Sources/Db/Schema/v3_0/AdminInfoFiles.php
@@ -110,7 +110,7 @@ class AdminInfoFiles extends Table
 			),
 			'data' => new Column(
 				name: 'data',
-				type: 'text',
+				type: 'mediumtext',
 				not_null: true,
 			),
 			'filetype' => new Column(


Oldiesmann:
That's what fixed it on my test install

PR #8958, opened 2025-09-21
[3.0] Installer not replacing values

### LINKED ISSUE #8951: [3.0]: Incorrect Theme/avatar paths and URLs on install
### Basic Information

When installing SMF3, the theme and avatar directories are not set properly because {$boardurl} and {$boarddir} aren't being parsed somewhere.

### Steps to reproduce

1. Install SMF3
2. Notice that the theme URL is wrong

### Expected result

Theme and avatar directory paths and URLs should be set correctly

### Actual result

"{$boardurl}" and "{$boarddir}" are literally part of the URL/path

### Version/Git revision

3c73aee

### Database Engine

All

### Database Version

9.3.0

### PHP Version

8.4.12

### Logs

```shell

```

### Additional Information

This has not been tested with PostgreSQL or MariaDB but it's assumed that the same issue occurs with those as well

jdarwood007:
Fixes #8951

We can't use just a simple array_map because we have more complexities, such as
```
			'value' => '{$boardurl}/Themes/default/images',
```

We can't just call a straight strtr, because sometimes the data is an int:
```
			'id_theme' => 1,
```

So we have to be a bit more verbose.

Additionally found some old code in the installer related to this.

### DIFF
--- a/Sources/Db/Schema/Table.php
+++ b/Sources/Db/Schema/Table.php
@@ -513,6 +513,11 @@ class Table
 			'{$registration_method}' => \defined('SMF_INSTALLING') ? ((int) !empty($_POST['reg_mode'])) : (Config::$modSettings['registration_method'] ?? 0),
 		];
 
+		// Sometimes its a string, sometimes its an array, sometimes its json.
+		if (\is_array($replacements['{$attachdir}'])) {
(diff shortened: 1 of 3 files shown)


MisterVector:
Installs fine for me and I don't notice any CSS issues.

PR #8964, opened 2025-09-29
[3.0] Category order

jdarwood007:
Fixes a bug where the drop down for where to place a new category does not show the category name.


Sesquipedalian:
The problem with the relevant language strings is actually more extensive that just this instance. I recommend the attached patch instead:

[mboards_order.patch](https://github.com/user-attachments/files/22602752/mboards_order.patch)

Sesquipedalian (changes requested):
See patch file.

jdarwood007:
Applied the suggestion.

I initially didn't go that way since we used the string in board management as well, but the proposed patch works.

Sesquipedalian (approved):

PR #8969, opened 2025-09-30
[3.0]: Implements Board::unload() & Category::unload()

### LINKED ISSUE #8943: [3.0]: Some things hidden when no session exists
### Basic Information

This has been an issue for a while, and I'm not sure if it's been reported or not... When you first visit a forum when no previous session exists (logged in or otherwise), some things are hidden that should otherwise be shown.

Guest: No boards shown, even though recent posts might be visible
User: Blank area to the right of boards where the most recent post would be shown

### Steps to reproduce

1. Visit a forum that you haven't visited in a while, whether you're logged in there or not - just so that no active session exists
2.

### Expected result

Everything should load like usual

### Actual result

Guest: No boards shown
User: No "most recent post" info shown for boards

Image

Image

### Version/Git revision

11827fe

### Database Engine

All

### Database Version

_No response_

### PHP Version

8.4.12

### Logs

```shell

```

### Additional Information

_No response_

Sesquipedalian:
Based on a patch by @jdarwood007.

These methods allow us to do a better job of ensuring that we clean up unnecessary categories when we unset items in Board::$loaded.

Also makes some changes to ensure that we always load all essential properties for a board (such as $last_msg) in Board::load(). This is a quick, short-term fix for the problem described in https://github.com/SimpleMachines/SMF/issues/8943#issuecomment-3344620641, although I will still want to do a broader refactoring of the Board and Category classes soon.

### DIFF
--- a/Sources/Board.php
+++ b/Sources/Board.php
@@ -926,7 +926,7 @@ class Board implements \ArrayAccess, Routable
 		}
 
 		// Can we unload the category too?
-		if (self::$loaded[$id]->cat?->id > 0 && array_filter(self::$loaded, fn($b) => $b->id !== $id && $b->cat?->id === self::$loaded[$id]->cat->id) === []) {
+		if ($id > 0 && self::$loaded[$id]->cat?->id > 0 && array_filter(self::$loaded, fn($b) => $b->id !== $id && $b->cat?->id === self::$loaded[$id]->cat->id) === []) {
 			Category::unload(self::$loaded[$id]->cat->id);
 		}
 
--- a/Sources/Board.php
+++ b/Sources/Board.php
@@ -914,6 +914,26 @@ class Board implements \ArrayAccess, Routable
 	}
 
 	/**
+	 * Removes an instance of this class, handling any necessary cleanups and
+	 * memory optimizations.
+	 *
+	 * @param int $id The ID number of the board to unload.
+	 */
+	public static function unload(int $id): void
+	{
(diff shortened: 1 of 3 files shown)


Sources/Board.php
```diff
@@ -913,6 +913,26 @@ public static function init(?int $id = null, array $props = []): ?self
...
+	 * @param int $id The ID number of the board to unload.
+	 */
+	public static function unload(int $id): void
+	{
+		if (!isset(self::$loaded[$id])) {
+			return;
+		}
+
+		// Can we unload the category too?
+		if (self::$loaded[$id]->cat?->id > 0 && array_filter(self::$loaded, fn($b) => $b->id !== $id && $b->cat?->id === self::$loaded[$id]->cat->id) === []) {
```
jdarwood007:
```suggestion
		if ($id > 0 && self::$loaded[$id]->cat?->id > 0 && array_filter(self::$loaded, fn($b) => $b->id !== $id && $b->cat?->id === self::$loaded[$id]->cat->id) === []) {
```

Sesquipedalian:
Latest commit makes some changes to ensure that we always load all essential properties for a board (such as $last_msg) in Board::load(). This is a quick, short-term fix for the problem described in https://github.com/SimpleMachines/SMF/issues/8943#issuecomment-3344620641, although I will still want to do a broader refactoring of the Board and Category classes soon.

jdarwood007 (approved):

PR #8983, opened 2025-10-16
Lint PHP files faster

live627:
- Only setup a new PHP environment if versions differ 
- Run the newest (much faster) version of phplint on supported environments

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -47,8 +47,16 @@ jobs:
     steps:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 #4.2.2
 
+    - name: Get PHP version
+      id: php-version
+      run: |
+        php -v
+        PHP_VERSION=$(php -r "echo PHP_MAJOR_VERSION . '.' . PHP_MINOR_VERSION;")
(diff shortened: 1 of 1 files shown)


.github/workflows/php.yml
```diff
@@ -66,4 +74,11 @@ jobs:
       run: composer install --prefer-dist --no-progress --ansi
 
     - name: Lint PHP files
+      if: ${{ matrix.php == 8 }}
       run: vendor/bin/phplint -w --exclude .git --exclude vendor --ansi .
+
+    - name: Check PHP syntax errors
+      if: ${{ matrix.php != 8 }}
+      uses: overtrue/phplint@9.6.2
```
jdarwood007:
```suggestion
      uses: overtrue/phplint@e2ebdbac93fa6ffe45d724e350f77df34c1f8fb4 #9.6.2
```

You should avoid tags and use commit ids.  Tags can be recreated with malicious code injected
The hashtag is just a comment, so we know what version we intended it to be agaisnt

live627:
Very pleased with the results.  Faster builds always make me happy. I strongly dislike wasted resources.

Screenshot 2025-10-15 at 20-28-59 Lint PHP files faster · SimpleMachines_SMF@21b8a39

Issue #8986, opened 2025-10-18
[2.1 & 3.0]: `<move-file>` in `package-info.xml` incorrectly reports "File not found"


Daaaav:
### Basic Information

The `<move-file>` tag in a mod package (not `<require-file>`), as far as I can tell, is meant to move and/or rename a file _from_ somewhere inside the SMF folder _to_ somewhere inside the SMF folder. For example creating a backup file of a standard file you will overwrite, so mod uninstallation can restore the original file again. In SMF 2.1 and 3.0-alpha, this seems to always report that an error in mod installation will happen (File not found). After bypassing the warning, the actual move does work though.

### Steps to reproduce

1. Have a mod package with a `package-info.xml` containing something like:
   ```xml
	<install for="2.0 - 2.1.99">
		<move-file from="$boarddir/ssi_examples.php" name="ssi_examples2.php" destination="$boarddir" />
	</install>
   ```
2. Install the package

### Expected result

This should rename `ssi_examples.php` to `ssi_examples2.php`, after confirming the mod install will be successful. As SMF 2.0 indicates:

<img width="887" height="173" alt="Image" src="https://github.com/user-attachments/assets/32e53e90-2739-412a-b7ce-6f7d78c5677c" />

### Actual result

SMF 2.1 and 3.0-alpha indicate an error is about to happen, even though the file is there and the rename will work:

<img width="901" height="500" alt="Image" src="https://github.com/user-attachments/assets/27589930-a745-4420-aebf-ec4de72adf2e" />

### Version/Git revision

2.1.6 & 3.0 Alpha 4 (08de9e7)

### Database Engine

MySQL

### Database Version

MySQL 8.0.43 & MariaDB 10.11.13

### PHP Version

7.3.33 & 8.1.33

### Logs

```shell

```

### Additional Information

The syntax of `<move-file>` is [documented](https://wiki.simplemachines.org/smf/Category:Package_SDK) a bit ambiguously (which arguments are full paths, parent folder paths or filenames, and what is old and new) but as confirmed by [a topic from 2005](https://www.simplemachines.org/community/index.php?topic=41752.msg305010#msg305010), it is:
- `from`: old full path (e.g. `$boarddir/ssi_examples.php`)
- `name`: new filename (e.g. `ssi_examples2.php`)
- `destination`: new parent folder (e.g. `$boarddir`)

Mod file to test with:
[...]

PR #8987, opened 2025-10-21
[2.1] Fix `<create-file>`, `<move-file>` and `<move-dir>` in `package-info.xml`

### LINKED ISSUE #8970: [2.1 & 3.0]: `<create-file>` in `package-info.xml` is broken
### Basic Information

The `<create-file>` tag in a mod package is meant to create a blank file, but this doesn't work properly in both 2.0 and 2.1 (haven't tested 3.0).

### Steps to reproduce

1. Have a mod package with a `package-info.xml` containing something like:
   ```xml
	<install for="2.0 - 2.1.99">
		<create-file name="new_file.txt" destination="$boarddir" />
	</install>
   ```
2. Install the package

### Expected result

The blank file is created without errors.

### Actual result

Testing on an old PHP 7.3.33, the mod seems to install correctly and the blank file is created, but two errors appear in the logs:

[Subs-Package.php:1395](https://github.com/SimpleMachines/SMF/blob/release-2.1/Sources/Subs-Package.php#L1395): `8: Undefined index: source`
[Subs-Package.php:2613](https://github.com/SimpleMachines/SMF/blob/release-2.1/Sources/Subs-Package.php#L2613): `2: file_get_contents(): Filename cannot be empty`

On PHP 8.1.33, the mod install crashes halfway:

Path cannot be empty

Similar errors appear in the log (`2: Undefined array key "source"` and `Path cannot be empty`) pointing to the same lines.

### Version/Git revision

2.1.6

### Database Engine

All

### Database Version

_No response_

### PHP Version

8.1.33

### Logs

```shell

```

### Additional Information

Both the [Package SDK page](https://wiki.simplemachines.org/smf/Category:Package_SDK) and the [DTD](http://www.simplemachines.org/xml/package-info) say that `name` and `destination` are the correct arguments.

Seems to me the following would fix it, but I haven't tested that:
```php
			// Create an empty file.
			package_put_contents($action['destination'], package_get_contents($action['source']), $testing_only);
```
Change to:
```php
			// Create an empty file.
			package_put_contents($action['destination'], '', $testing_only);
```

Mod file to test with:
[test_create_file_v1.0.zip](https://github.com/user-attachments/files/22625287/test_create_file_v1.0.zip)

Daaaav:
The fixes are pretty simple:

- `<create-file>` was trying to create a file with the contents of the "source" file, instead of an empty file as intended.
- For `<move-file>` and `<move-dir>`, the test was trying to confirm that the destination file existed, when it's actually supposed to not exist since we're moving something to that name.

  Turns out a similar fix was done earlier to stop the warning from showing up for `<create-dir>` and `<create-file>` (#3009), so I just extended that fix.

Fixes #8970 and #8986 for 2.1, lemme know if you'd like me to make a 3.0 PR too :)

### DIFF
--- a/Sources/Packages.php
+++ b/Sources/Packages.php
@@ -663,7 +663,7 @@ function PackageInstallTest()
 		}
 
 		// Don't fail if a file/directory we're trying to create doesn't exist...
-		if (isset($action['filename']) && !file_exists($file) && !in_array($action['type'], array('create-dir', 'create-file')) && $action['error'] != 'ignore')
+		if (isset($action['filename']) && !file_exists($file) && !in_array($action['type'], array('create-dir', 'create-file', 'move-dir', 'move-file')) && $action['error'] != 'ignore')
 		{
 			$context['has_failure'] = true;
 
--- a/Sources/Subs-Package.php
+++ b/Sources/Subs-Package.php
@@ -1392,7 +1392,7 @@ function parsePackageInfo(&$packageXML, $testing_only = true, $method = 'install
 				$failure |= !mktree(dirname($action['destination']), 0777);
 
 			// Create an empty file.
-			package_put_contents($action['destination'], package_get_contents($action['source']), $testing_only);
+			package_put_contents($action['destination'], '', $testing_only);
 
 			if (!file_exists($action['destination']))
 				$failure = true;


Sesquipedalian (approved):

Sesquipedalian:
> lemme know if you'd like me to make a 3.0 PR too :)

Please do, yes.

PR #8991, opened 2025-10-23
[2.1] Fix string literal detection

sbulen:
Fixes #8980

pg & mysql escape literals a bit differently.  The rules that work for one won't work for the other.  pg will escape single quotes by doubling them, `''`, mysql will escape them with a backslash, `\'`.  Also, mysql escapes backslashes, `\\`, where pg doesn't.

So you cannot identify & clean the literals with the same logic.

The 'query check' logic makes a fully fleshed out copy of the current query, inserting values & smf substitutions, in order to do some syntax checks for behavior that SMF forbids.  It wants to focus on the sql syntax itself, not anything that might only look like syntax, so it strips all string literals.  Then it can easily check for forbidden syntax - internal comments, stacked sql commands, sleep() & benchmark() functions.  The cleaned copy is discarded after this check.

Here are two properly escaped strings in pg:
`'Hello world \'`
`'subject: ''SMF 2.0.17 has been released'','`

And here are the same strings in mysql:
`'Hello world \\'`
`'subject: \'SMF 2.0.17 has been released\','`

So...  The solution here was to remove the respective escapes first, making removing the string literals more straightforward.  For mysql, this was a two step process - had to get rid of escaped backslashes first, in order to make the distinction between an escaped backslash with a valid end-of-string single quote (`\\'`) and an escaped mid-string single quote (`\'`).

An aside: during A/B testing, I ran across one of those elusive 'mysql result expected false given' errors.  This may have been a (the?) source of them.

Another benefit of these fixes: Performance.  On very large inserts, with lots of literals, the strpos() calls in the old method were insanely inefficient.  (I have a mod that loads large files, and disabling the query check improves load time ~10x...  Viewing the execution profile in WinCacheGrind, virtually all the time was spent executing strpos()...  MUCH faster now.)

PR #8996, opened 2025-10-25
[2.1] Improve cleaning of string literals

### LINKED ISSUE #8993: [2.1]: "No direct access" bug still there, in both pg & in mysql
### Basic Information

Issue #8309 is still open.

I suggest that we either reconsider #8991, or at least, incorporate the info painstakingly explained in it into the latest approach.

We still get 'No direct access' on both mysql and on pg using the simple PM test laid out in #8309.  

Further, The related "hidden" issues are still present as well, as in this example from parsesmileys (Subs.php line 3902):

```
				SELECT s.code, f.filename, s.description
				FROM smf_smileys AS s
					JOIN smf_smiley_files AS f ON (s.id_smiley = f.id_smiley)
				WHERE f.smiley_set = 'fugue'
					AND s.code IN ('>:D', ':D', '::)', '>:(', ':))', ':)', ';)', ';D', ':(', ':o', '8)', ':P', '???', ':-[', ':-X', ':-*', ':\'(', ':-\\', '^-^', 'O0', 'C:-)', 'O:-)')
```

The clean version does not have all literals properly identified, the substitution spans literals, and in fact, eats part of the query:
`select s.code, f.filename, s.description from smf_smileys as s join smf_smiley_files as f on (s.id_smiley = f.id_smiley) where f.smiley_set = %s and s.code in ( %s , %s , %s , %s , %s , %s , %s , %s , %s , %s , %s , %s , %s , %s , %s , %s , %s , %s ^-^ %s o0 %s c:-) %s o:-) %s
`

The above fails in both pg and in mysql as well.

When testing, I suggest using AT LEAST these 3 tests, in both mysql and in postgresql:
 - Sleepy's test spelled out in #8309
 - Fetch Simple Machines Files
 - The query in parsesmileys, Subs.php line 3902

#8991 passes all these tests.  It's helpful to place a breakpoint after the cleaning routines in Subs-Db-xxxxx.php and observe dozens of queries, to make sure everything is caught & properly cleaned.

That smiley query is a very good test...

### Steps to reproduce

1. Send a PM with `Hello World \` as the subject and `I'll` as the body
2.

### Expected result

Sent PM

### Actual result

"No direct access" error.

### Version/Git revision

2.1 - current GitHub

### Database Engine

All

### Database Version

mysql 8.4; pg 14

### PHP Version

8.4.5

### Logs

```shell

```

### Additional Information

_No response_

sbulen:
Fixes #8993 

Just placing this out there again for consideration.  It's an update of #8991

### DIFF
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -368,12 +368,14 @@ function smf_db_query($identifier, $db_string, $db_values = array(), $connection
 
 	// Comments that are allowed in a query are preg_removed.
 	static $allowed_comments_from = array(
+		'~\'\X*?\'~s',
 		'~\s+~s',
 		'~/\*!40001 SQL_NO_CACHE \*/~',
 		'~/\*!40000 USE INDEX \([A-Za-z\_]+?\) \*/~',
 		'~/\*!40100 ON DUPLICATE KEY UPDATE id_msg = \d+ \*/~',
 	);
 	static $allowed_comments_to = array(
+		' %s ',
 		' ',
 		'',
 		'',
@@ -415,19 +417,9 @@ function smf_db_query($identifier, $db_string, $db_values = array(), $connection
 	// First, we clean strings out of the query, reduce whitespace, lowercase, and trim - so we can check it over.
 	if (empty($modSettings['disableQueryCheck']))
 	{
-		$clean = preg_split('/(?<![\'\\\\])\'(?![\'])/', $db_string);
-
-		for ($i = 0; $i < count($clean); $i++)
-		{
-			if ($i % 2 === 1)
-				$clean[$i] = ' %s ';
-		}
-
-		$clean = trim(strtolower(preg_replace(
-			$allowed_comments_from,
-			$allowed_comments_to,
-			implode('', $clean)
-		)));
+		// Clear out escaped backslashes & single quotes first, to make it simpler to ID & remove string literals
+		$clean = str_replace(array('\\\\', '\\\''), array('', ''), $db_string);
+		$clean = trim(strtolower(preg_replace($allowed_comments_from, $allowed_comments_to, $clean)));
 
 		// Comments?  We don't use comments in our queries, we leave 'em outside!
 		if (strpos($clean, '/*') > 2 || strpos($clean, '--') !== false || strpos($clean, ';') !== false)
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -368,14 +368,12 @@ function smf_db_query($identifier, $db_string, $db_values = array(), $connection
 
 	// Comments that are allowed in a query are preg_removed.
 	static $allowed_comments_from = array(
-		'~\'\X*?\'~s',
 		'~\s+~s',
 		'~/\*!40001 SQL_NO_CACHE \*/~',
 		'~/\*!40000 USE INDEX \([A-Za-z\_]+?\) \*/~',
 		'~/\*!40100 ON DUPLICATE KEY UPDATE id_msg = \d+ \*/~',
 	);
 	static $allowed_comments_to = array(
-		' %s ',
 		' ',
 		'',
 		'',
@@ -417,9 +415,19 @@ function smf_db_query($identifier, $db_string, $db_values = array(), $connection
 	// First, we clean strings out of the query, reduce whitespace, lowercase, and trim - so we can check it over.
 	if (empty($modSettings['disableQueryCheck']))
 	{
-		// Clear out escaped backslashes & single quotes first, to make it simpler to ID & remove string literals
-		$clean = str_replace(array('\\\\', '\\\''), array('', ''), $db_string);
-		$clean = trim(strtolower(preg_replace($allowed_comments_from, $allowed_comments_to, $clean)));
+		$clean = preg_split('/(?:\\\\{2})*\K(?<![\'\\\\])\'(?![\'])/', $db_string);
+
+		for ($i = 0; $i < count($clean); $i++)
+		{
+			if ($i % 2 === 1)
+				$clean[$i] = ' %s ';
+		}
+
+		$clean = trim(strtolower(preg_replace(
+			$allowed_comments_from,
+			$allowed_comments_to,
+			implode('', $clean)
+		)));
 
 		// Comments?  We don't use comments in our queries, we leave 'em outside!
 		if (strpos($clean, '/*') > 2 || strpos($clean, '--') !== false || strpos($clean, ';') !== false)
--- a/Sources/Subs-Db-postgresql.php
+++ b/Sources/Subs-Db-postgresql.php
@@ -379,12 +379,14 @@ function smf_db_query($identifier, $db_string, $db_values = array(), $connection
 
 	// Comments that are allowed in a query are preg_removed.
 	static $allowed_comments_from = array(
+		'~\'\X*?\'~s',
 		'~\s+~s',
 		'~/\*!40001 SQL_NO_CACHE \*/~',
 		'~/\*!40000 USE INDEX \([A-Za-z\_]+?\) \*/~',
 		'~/\*!40100 ON DUPLICATE KEY UPDATE id_msg = \d+ \*/~',
 	);
 	static $allowed_comments_to = array(
+		' %s ',
 		' ',
 		'',
 		'',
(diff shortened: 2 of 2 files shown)


Sesquipedalian:
I want to review this carefully. I will make time to do so in the next few days.

sbulen:
Another note on these escapes...  I initially thought - if php regex allowed it (uh... nope...) - of putting a 2nd lookbehind for mysql, so `\\'` would be understood as a single quote whereas `\'` would be understood as an escaped single quote...

The problem is that it can keep going...  e.g., this is a single quote that terminates a string:
`\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\'`
...but this is an escaped single quote that does not terminate a string:
`\\\\\\\\\\\\\\\\\\\\\\\\\\\\\'`

So it'd need to check for an even vs odd # of chars in a lookbehind in regex...  Which is why I ultimately gave up on using a regex for that.  A similar problem exists in pg - the number of single quotes in a row determines whether it's escaped or not...

Sources/Subs-Db-mysql.php
```diff
@@ -368,12 +368,14 @@ function smf_db_query($identifier, $db_string, $db_values = array(), $connection
 
 	// Comments that are allowed in a query are preg_removed.
 	static $allowed_comments_from = array(
+		'~\'\X*?\'~s',
```
Sesquipedalian:
```suggestion
```

Sources/Subs-Db-mysql.php
```diff
@@ -368,12 +368,14 @@ function smf_db_query($identifier, $db_string, $db_values = array(), $connection
...
 		'~\s+~s',
 		'~/\*!40001 SQL_NO_CACHE \*/~',
 		'~/\*!40000 USE INDEX \([A-Za-z\_]+?\) \*/~',
 		'~/\*!40100 ON DUPLICATE KEY UPDATE id_msg = \d+ \*/~',
 	);
 	static $allowed_comments_to = array(
+		' %s ',
```
Sesquipedalian:
```suggestion
```

Sources/Subs-Db-mysql.php
```diff
@@ -415,19 +417,9 @@ function smf_db_query($identifier, $db_string, $db_values = array(), $connection
...
-		}
-
-		$clean = trim(strtolower(preg_replace(
-			$allowed_comments_from,
-			$allowed_comments_to,
-			implode('', $clean)
-		)));
+		// Clear out escaped backslashes & single quotes first, to make it simpler to ID & remove string literals
+		$clean = str_replace(array('\\\\', '\\\''), array('', ''), $db_string);
+		$clean = trim(strtolower(preg_replace($allowed_comments_from, $allowed_comments_to, $clean)));
```
Sesquipedalian:
```suggestion
		$clean = preg_split('/(?:\\\\{2})*\K(?<![\'\\\\])\'(?![\'])/', $db_string);

		for ($i = 0; $i < count($clean); $i++)
		{
			if ($i % 2 === 1)
				$clean[$i] = ' %s ';
		}

		$clean = trim(strtolower(preg_replace(
			$allowed_comments_from,
			$allowed_comments_to,
			implode('', $clean)
		)));
```

Sesquipedalian (changes requested):
The proposed changes for PostgreSQL work well, but the proposed changes for MySQL ran into issues with some complex strings. This is because the SQL standard `''` syntax for escaping apostrophe characters is unambiguous and can therefore be handled accurately with a simple `str_replace()` call, whereas the C-style `\'` syntax that MySQL supports can create more complexity that requires taking the surrounding context into account when doing this job.

That said, I did find a case while testing for MySQL where the existing regex used in the `preg_split()` call wasn't quite good enough and needs a tweak. 

Thus, I have made code suggestions in the comments for this review that revert the proposed changes made to the MySQL code but also adjust the regex used in the `preg_split()` call.

Sources/Subs-Db-postgresql.php
```diff
@@ -413,19 +415,9 @@ function smf_db_query($identifier, $db_string, $db_values = array(), $connection
...
-				$clean[$i] = ' %s ';
-		}
-
-		$clean = trim(strtolower(preg_replace(
-			$allowed_comments_from,
-			$allowed_comments_to,
-			implode('', $clean)
-		)));
+		// Clear out escaped single quotes first, to make it simpler to ID & remove string literals
+		$clean = str_replace('\'\'', '', $db_string);
```
Sesquipedalian:
```suggestion
		$clean = str_replace('\'\'', ' ', $db_string);
```

Here is one more important recommendation. By replacing the escaped quotation mark with a space rather than nothing, we can avoid accidentally creating strings that contain unexpected syntax structures that were not present in the original.

Sesquipedalian (changes requested):
One more important change to make.

sbulen:
> whereas the C-style `\'` syntax that MySQL supports can create more complexity that requires taking the surrounding context into account when doing this job.

Can you provide an example of this?  Where the PR would fail?

Sesquipedalian:
> > whereas the C-style `\'` syntax that MySQL supports can create more complexity that requires taking the surrounding context into account when doing this job.
> 
> Can you provide an example of this? Where the PR would fail?

Start with this as the incoming query string passed to the $db_string parameter of smf_db_query():
```
SELECT 'hel''lo hel\\''lo /*'
```
... which uses a mix of `''` syntax for escaping apostrophes and `\\` syntax for escaping backslashes in order to escape the raw string `hel'lo hel\'lo /*`. Using a mix of syntaxes like that would be a weird thing to do, but it is perfectly legal in MySQL.

When the MySQL code proposed in this PR tries to make a cleaned version of this query string, it will produce this:
```
select %s %s lo /*'
```
... which is a problem.

Now, _in theory_ such a situation isn't supposed to arise because people are supposed to be pass the raw string (`hel'lo hel\'lo /*` in this case) as a value in the $db_values parameter of smf_db_query() rather than writing it directly into $db_string. But someone _could_ do that if they chose to, and our tests would be wrong to reject it since it is in fact a perfectly valid query string.

EDIT: Appended ` /*` to the string literal in the test query in order to make the problem completely obvious.

sbulen:
The proposed change to the PR fails simple tests - e.g., this PM:
Subject: `Hello World I\'m home`
Body: `I'll`

Not unusual; folks share code.

If the only way to get the current PR to fail is by manually hacking the code &, further, not calling the appropriate $smcFunc...  I don't think that's a problem.  Don't hack the code, use SMF standards, you'll be fine.

Sesquipedalian:
"Don't do that" isn't an adequate approach here. 

I will test your example soon. It is entirely possible that the preg_split() approach remains insufficient. If so, we will need to find something else.

Also, calling $smcFunc['db_query'] vs. smf_db_query makes no difference regarding my comments above. I referred to the latter for clarity, that's all.

sbulen:
If they're not calling smcFunc, they're not here at all, though...   If they are calling smcFunc, it'll get escaped accordingly...

I've tested that, too.

I think we're covered with this PR.

Sesquipedalian:
```php
$request = $smcFunc['db_query']('', 'SELECT \'hel\'\'lo hel\\\\\'\'lo /*\'', array());
```

Merely using `$smcFunc['db_query']()` does not not guarantee that string literals in the query will be escaped consistently. If the string were passed in the $db_values parameter, then it would be escaped consistently, sure. But putting a manually escaped string literal in the $db_string parameter is entirely possible with `$smcFunc['db_query']()`.

Sesquipedalian:
> The proposed change to the PR fails simple tests - e.g., this PM: Subject: `Hello World I\'m home` Body: `I'll`

Yup, I was able to reproduce that. Somewhat surprisingly, though, the PM still sent and the error appeared when trying to write the session data to the database after the PM had been sent. At any rate, that error message did allow me to find the issue with the regular expression I was using in the pre_split() call.

Instead of this:
```php
$clean = preg_split('/(?<![\'\\\\])\'(?![\'])|(?<=[\\\\]{2})\'/', $db_string);
```
... we need this:
```php
$clean = preg_split('/(?:\\\\{2})*\K(?<![\'\\\\])\'(?![\'])/', $db_string);
```

The `(?:\\\\{2})*` part matches zero or more escaped backslashes (i.e. zero or more sets of two backslashes). Then the `\K` drops those escaped backslashes from the match and proceeds to look for an unescaped apostrophe. If it finds an unescaped apostrophe, it splits the string around it.

This approach (using `(?:\\\\{2})*\K`) is the way to accomplish this:
> So it'd need to check for an even vs odd # of chars in a lookbehind in regex...

I have updated my suggestion in my review comments to use this new regex.

If you can find any other cases where my latest suggestion fails, please let me know, @sbulen.

Sesquipedalian (approved):

PR #9002, opened 2025-10-26
[3.0] Popup setBody function

### LINKED ISSUE #8989: [3.0]: Popup overlays no longer work
### Basic Information

Clicking on a link that is supposed to create a popup overlay, such as a help icon, no longer opens the content in the expected overlay. Instead, the entire web page goes to the target URL, just like a normal link would.

### Steps to reproduce

1. Navigate to any admin page and click on a help icon.
2.

### Expected result

The help text overlay appears and shows the help text, like this:

Image

### Actual result

Browser goes to a separate page containing the help text, like this:

Image

### Version/Git revision

3.0 Alpha 4

### Database Engine

All

### Database Version

N/A

### PHP Version

N/A

### Logs

```shell
JavaScript console reports the following error:

[Error] ReferenceError: Can't find variable: $
	Global Code (script.js:1377)
```

### Additional Information

This bug was introduced in #8884

jdarwood007:
The problem with the existing code is that the cover object doesn't exist. On a more logical sense, we shouldn't be manipulating the pop-up like this without going through its object.

Introduces a setBody, which allows us to populate the contents of a pop-up.

Fixes #8989

### DIFF
--- a/Themes/default/scripts/script.js
+++ b/Themes/default/scripts/script.js
@@ -275,8 +275,8 @@ function reqOverlayDiv(desktopURL, sHeader, sIcon)
 	})
 		.then((res, rej) => res.ok ? res.text() : rej(res))
 		.then(data => {
-			const parser = new DOMParser()
-			const doc = parser.parseFromString(data, "text/html")
+			const parser = new DOMParser();
+			const doc = parser.parseFromString(data, "text/html");
 			const body = doc?.querySelector('body div')?.innerHTML ?? data;
 			oContainer.setBody(body);
 		})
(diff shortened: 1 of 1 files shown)


Themes/default/scripts/script.js
```diff
@@ -276,11 +275,14 @@ function reqOverlayDiv(desktopURL, sHeader, sIcon)
 	})
 		.then((res, rej) => res.ok ? res.text() : rej(res))
 		.then(data => {
-			oPopup_body.innerHTML = data;
+			const parser = new DOMParser()
+			const doc = parser.parseFromString(data, "text/html")
```
live627:
```suggestion
			const parser = new DOMParser();
			const doc = parser.parseFromString(data, "text/html");
```

live627 (approved):

PR #9003, opened 2025-10-26
[3.0] Fix PM logic for groups

### LINKED ISSUE #8994: [3.0]: Cannot send PM - array_intersect(): Argument #1 ($array) must be of type array, null given
### Basic Information

Cannot send PM - get an error:  array_intersect(): Argument # 1 ($array) must be of type array, null given 

### Steps to reproduce

1. Install 3.0
2. Register a new user
3. Try to send a PM to that user

### Expected result

Sent PM

### Actual result

There are actually two errors in the logs:
Image

### Version/Git revision

3.0 Alpha 4 - current GH

### Database Engine

MySQL

### Database Version

8.4.4

### PHP Version

8.4.5

### Logs

```shell

```

### Additional Information

_No response_

jdarwood007:
Fixes #8994

@Sesquipedalian
This was introduced in #8574 

The fix I did is maybe more than a patch.  But it seems that you deprecated the 'allowed'/'denied' return of the getAllowedTo and added instructions for another function, which doesn't return data the same way.
Updating the PM logic to fully use that logic would have it do much more work.

### DIFF
--- a/Sources/PersonalMessage/PM.php
+++ b/Sources/PersonalMessage/PM.php
@@ -1295,7 +1295,15 @@ class PM implements \ArrayAccess
 		}
 
 		// Load the groups that are allowed to read PMs.
-		$pmReadGroups = Group::getAllowedTo('pm_read');
+		$allGroups = Group::getAllWithPermissions('pm_read', null);
+
+		foreach ($allGroups as $k => $g) {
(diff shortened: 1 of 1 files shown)


sbulen:
Tested this, & it does address the reported error.  Thanks,

PR #9006, opened 2025-10-26
[3.0] More detailed loop message

jdarwood007:
Replaces #7431

### DIFF
--- a/Sources/ErrorHandler.php
+++ b/Sources/ErrorHandler.php
@@ -219,7 +219,7 @@ class ErrorHandler
 		if ($error_call > 2) {
 			var_dump($backtrace);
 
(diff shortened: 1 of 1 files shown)


Sources/ErrorHandler.php
```diff
@@ -219,7 +219,7 @@ public static function log(string $error_message, string|bool $error_type = 'gen
 		if ($error_call > 2) {
 			var_dump($backtrace);
 
-			die('Error loop.');
+            die('Error, loop detected. The database may have failed or crashed.');
```
live627:
```suggestion
			die('Error: loop detected. The database may have failed or crashed.');
```

PR #9014, opened 2025-10-27
[3.0] Fix key identification to address data population

### LINKED ISSUE #9012: [3.0]: pg install: database not properly populated
### Basic Information

After a 3.0 install for pg, the database is not fully populated.

You'll see lots of errors along the lines of:
```
Warning: Undefined array key "attachmentUploadDir" in D:\wamp64\www\pgvan30\Sources\Config.php on line 1060

Warning: Undefined array key "attachmentUploadDir" in D:\wamp64\www\pgvan30\Sources\Config.php on line 1061

Fatal error: Uncaught TypeError: SMF\Utils::jsonDecode(): Argument #1 ($json) must be of type string, null given, called in D:\wamp64\www\pgvan30\Sources\Config.php on line 1061 and defined in D:\wamp64\www\pgvan30\Sources\Utils.php:1454 Stack trace: #0 D:\wamp64\www\pgvan30\Sources\Config.php(1061): SMF\Utils::jsonDecode(NULL, true, 512, 0, false) #1 D:\wamp64\www\pgvan30\Sources\Maintenance\Tools\Install.php(850): SMF\Config::reloadModSettings() #2 D:\wamp64\www\pgvan30\Sources\Maintenance\Maintenance.php(294): SMF\Maintenance\Tools\Install->adminAccount() #3 D:\wamp64\www\pgvan30\install.php(31): SMF\Maintenance\Maintenance->execute(1) #4 {main} thrown in D:\wamp64\www\pgvan30\Sources\Utils.php on line 1454
```

I've traced the error to this line:
https://github.com/SimpleMachines/SMF/blob/6e810cb97c8db7e7be2d42c25b55f57056994735/Sources/Db/Schema/Table.php#L550

Which will only pass a key to a database for INSERTs if the key is an auto incremented key.  But many of our tables do not have auto incremented keys.  All inserts for those tables fail during database population.

E.g., the settings table.  

Under the covers, the errors look like:
```
INSERT INTO ato_board_permissions_view("id_group", "id_board", "deny")
				VALUES
					(-1, 1, 0),
					(0, 1, 0),
					(2, 1, 0) ON CONFLICT () DO NOTHING

error:ERROR:  syntax error at or near ")"
LINE 5:      (2, 1, 0) ON CONFLICT () DO NOTHING
                                    ^

```
Because nothing was passed as the key, because the key wasn't an auto increment.

### Steps to reproduce

1.Install 3.0 under pg
2.

### Expected result

Fully populated database.

### Actual result

Many INSERTS fail during the install.

### Version/Git revision

3.0 Alpha 4 - current GitHub

### Database Engine

PostgreSQL

### Database Version

14

### PHP Version

8.4.5

### Logs

```shell

```

### Additional Information

_No response_

sbulen:
Fixes #9012 

Note that it was important to pass the whole key, not just an auto inc column, to address the query issues.  Not all keys are just auto incs.

Further, the old logic wouldn't add a key unless it was a passed column in the query, but pg didn't like that because it wasn't always the complete unique/primary key (e.g., auto incs usually excluded).  So had to ensure it passed all pk elements, all the time.

Another possible solution would have been to remove the conflict target (key list in parens) entirely.  This solution felt a little more holistic, however, as it addressed the broader issue with Table.php as well.

Tests great; no more hidden errors.  Freshly installed pg 3.0 database is fully populated.

### DIFF
--- a/Sources/Db/APIs/PostgreSQL.php
+++ b/Sources/Db/APIs/PostgreSQL.php
@@ -413,17 +413,11 @@ class PostgreSQL extends DatabaseApi implements DatabaseApiInterface
 		if ($method == 'replace' || $method == 'ignore') {
 			$key_str = implode(',', $keys);
 			$col_str = '';
-
 			$count = 0;
-			$count_pk = 0;
 
+			// Make a list of the non-pk fields.
 			foreach ($columns as $columnName => $type) {
-				// Check pk field.
-				if (\in_array($columnName, $keys)) {
-					$count_pk++;
-				}
-				// Normal field.
-				elseif ($method == 'replace') {
+				if (!\in_array($columnName, $keys) && ($method == 'replace')) {
 					$col_str .= ($count > 0 ? ',' : '');
 					$col_str .= $columnName . ' = EXCLUDED.' . $columnName;
 					$count++;
--- a/Sources/Db/APIs/PostgreSQL.php
+++ b/Sources/Db/APIs/PostgreSQL.php
@@ -411,7 +411,7 @@ class PostgreSQL extends DatabaseApi implements DatabaseApiInterface
 
 		// PostgreSQL doesn't support replace: we implement a MySQL-compatible behavior instead
 		if ($method == 'replace' || $method == 'ignore') {
-			$key_str = '';
+			$key_str = implode(',', $keys);
 			$col_str = '';
 
 			$count = 0;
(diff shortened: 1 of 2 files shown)


Sources/Db/APIs/PostgreSQL.php
```diff
@@ -420,8 +420,6 @@ public function insert(string $method, string $table, array $columns, array $dat
 			foreach ($columns as $columnName => $type) {
 				// Check pk field.
 				if (\in_array($columnName, $keys)) {
-					$key_str .= ($count_pk > 0 ? ',' : '');
-					$key_str .= $columnName;
 					$count_pk++;
```
jdarwood007: If your doing that, remove the $count_pk logic.  It doesn't seem to be used anywhere else.

sbulen: Done

PR #9024, opened 2025-11-01
[3.0] Remove stray variable causing undefined variable warning which slipped in by a merge

MisterVector:
This PR removes a variable that slipped in by the following merge:

image

The variable is highlighted in red. The commit highlighted in green originally removed the variable.


jdarwood007:
Ooops, I fixed it with #9025

PR #9025, opened 2025-11-01
[3.0] Fix a undefined query_string

jdarwood007:
We can and will blame Vector


Sesquipedalian:
Way to go ~@jdarwood007~ @MisterVector 😉😆

PR #9028, opened 2025-11-02
[3.0] Extend template system with callable support

dragomano:
Adds support for callable strings (e.g., `Class::method#`) in the `loadSubTemplate` method, allowing both static methods and instance method calls with object caching. Adds the same support for the callback option in `Admin.template.php`

**Example with callback functions:**

```php
<?php

namespace SMF;

class SomeClass
{
	public static function getConfigVars(): array
	{
		return [
			['callback', 'legacy_callback_name'], // looking for `template_callback_legacy_callback_name` function
			['callback', __CLASS__ . '::test#'], // looking for `test` method in the current class
			['callback', TestClass::class . '::test#'], // looking for `test` method in other class
			['callback', TestClass::class . '::test2'], // looking for `test2` static method in other class
		];
	}

	public function test(): void
	{
		echo __METHOD__;
	}
}

class TestClass
{
	public function test(): void
	{
		echo __METHOD__;
	}

	public static function test2(): void
	{
		echo __METHOD__;
	}
}
```

**Example with subaction:**

```php
Utils::$context['sub_template'] = 'show_settings'; // current (legacy)
Utils::$context['sub_template'] = TestClass::class . '::test#'; // new
```

**Example with layers:**

```php
Utils::$context['template_layers'][] = TestClass::class . '::test#';
```

```php
<?php

namespace SMF;

class TestClass
{
	public function testAbove(): void
	{
		echo 'above';
	}

	public function testBelow(): void
	{
		echo 'below';
	}
}
```

### DIFF
--- a/Sources/Theme.php
+++ b/Sources/Theme.php
@@ -536,7 +536,7 @@ class Theme
 			}
 		}
 
-		if (! $template_loaded) {
+		if (!$template_loaded) {
 			$theme_function = 'template_' . $template_name;
 
 			$callable = Utils::getCallable($theme_function);
@@ -548,7 +548,7 @@ class Theme
 			}
 		}
 
-		if (! $template_loaded) {
+		if (!$template_loaded) {
 			// Handle errors based on the $fatal parameter.
 			if ($fatal === false) {
 				ErrorHandler::fatalLang(
--- a/Sources/Theme.php
+++ b/Sources/Theme.php
@@ -517,35 +517,21 @@ class Theme
 		$template_loaded = false;
 
 		if (\is_string($template_name) && \str_contains($template_name, '::')) {
-			$template_name = preg_replace_callback(
+			$template_name = \preg_replace_callback(
 				'/(#?)_(above|below)$/i',
-				fn($m) => ucfirst($m[2]) . $m[1],
+				fn($m) => \ucfirst($m[2]) . $m[1],
 				$template_name
 			);
 
-			$callable = Utils::getCallable($template_name);
-
-			if ($callable !== false) {
-				if (empty($function_params)) {
-					\call_user_func($callable);
-				} else {
-					\call_user_func_array($callable, $function_params);
-				}
-
+			if ($callable = Utils::getCallable($template_name, true)) {
+				\call_user_func_array($callable, $function_params);
 				$template_loaded = true;
 			}
 		}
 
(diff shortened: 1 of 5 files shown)


Sources/Theme.php
```diff
@@ -507,25 +507,48 @@ public static function loadSubTemplates(): void
...
+
+				$template_loaded = true;
+			}
 		}
 
-		// Attempt to call the sub-template function.
-		if (\is_callable($theme_function)) {
-			\call_user_func_array($theme_function, $function_params);
-		} else {
+		if (! $template_loaded) {
```
live627:
```suggestion
		if (!$template_loaded) {
```

Sources/Theme.php
```diff
@@ -507,25 +507,48 @@ public static function loadSubTemplates(): void
...
+			$callable = Utils::getCallable($theme_function);
+
+			if ($callable !== false) {
+				\call_user_func_array($theme_function, $function_params);
+
+				$template_loaded = true;
+			}
+		}
+
+		if (! $template_loaded) {
```
live627:
```suggestion
		if (!$template_loaded) {
```

Themes/default/Admin.template.php
```diff
@@ -872,8 +872,13 @@ function ($v)
...
 		if (is_array($config_var) && $config_var['type'] == 'callback')
 		{
-			if (function_exists('template_callback_' . $config_var['name']))
-				call_user_func('template_callback_' . $config_var['name']);
+			$callbackName = $config_var['name'];
+			if (function_exists('template_callback_' . $callbackName)) {
+				$callbackName = 'template_callback_' . $callbackName;
+			}
+
+			$callable = Utils::getCallable($callbackName);
```
live627:
```suggestion
			$callback_name = $config_var['name'];
			if (function_exists('template_callback_' . $callback_name)) {
				$callback_name = 'template_callback_' . $callback_name;
			}

			$callable = Utils::getCallable($callback_name);
```

live627 (changes requested):
Should similar code in the profile area also be updated?

Sesquipedalian:
This is a cool idea, but we might end up making _profound_ changes to the theme system in 3.0. For that reason, I recommend that you don't invest much effort into theme-related code until that has been settled.

PR #9036, opened 2025-11-08
Improve space efficiency for regex building by using a radix tree

live627:
My [test of over 1000 cities](https://gist.github.com/live627/e01affb5ca52fa154119d6bb875e8c474) shows a sizeable memory reduction. Sidebar: This gist was accidentally left private in the last update, yet no one mentioned it.

### DIFF
--- a/Sources/Utils.php
+++ b/Sources/Utils.php
@@ -1053,10 +1053,11 @@ class Utils
 
 		// Can we trim common characters from the end?
 		$trailing = '';
+		$i = -1;
 		unset($normalized_strings);
 
-		while (mb_strlen($strings[0], $encoding) > 1) {
-			$last_char = mb_substr($strings[0], -1, 1, $encoding);
+		while (\strlen($strings[0]) > 1) {
+			$last_char = mb_substr($strings[0], $i, null, $encoding);
 
 			foreach ($strings as $string) {
 				if (!str_ends_with($string, $last_char)) {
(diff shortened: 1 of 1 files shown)


Sources/Utils.php
```diff
@@ -1053,48 +1053,96 @@ public static function buildRegex(array $strings, ?string $delim = null, bool $r
...
+						$len++;
+					}
+
+					if ($len === 0) {
+						continue; // no prefix match, try next branch
+					}
+
+					$prefix = mb_substr($key, 0, $len, $encoding);
+					$key_remainder = mb_substr($key, $len, null, $encoding);
+					$remaining_remainder = mb_substr($remaining, $len, null, $encoding);
```
live627: I have some code which cuts these string based on byte boundaries,   but that will make this function UTF-8 only. Would that be acceptable?

Sources/Utils.php
```diff
@@ -1031,7 +1031,7 @@ public static function buildRegex(array $strings, ?string $delim = null, bool $r
 		$normalized_strings = [];
 
 		foreach ($strings as $str) {
-			if (\is_scalar($str)) {
+			if (is_scalar($str)) {
```
jdarwood007: why did this change?

Sesquipedalian (dismissed):
~Splitting into bytes rather than characters is indeed more space efficient, but sadly it's not an option for us. That's because splitting into bytes will introduce byte sequences that are not valid UTF-8 into the overall regex. Although the preg_* functions themselves can handle that, having invalid byte sequences will cause problems when trying to store the compiled regular expressions.~

~For example, we store the TLD regex in the database. If the regex contains invalid byte sequences, the database will reject it.~

Never mind. I misread the proposed code.

Sesquipedalian:
> My [test of over 1000 cities](https://gist.github.com/live627/e01affb5ca52fa154119d6bb875e8c474) shows a sizeable memory reduction. Sidebar: This gist was accidentally left private in the last update, yet no one mentioned it.

I'm pretty sure I did see the gist before, but now I get a 404 at that link. Do you have the privacy setting backwards?

Sesquipedalian:
In my testing, these proposed changes are slower than the existing code. However, there are some good ideas in here for improving performance.

PR #9043, opened 2025-11-11
[2.1] Fix string literal cleanup - take 3

### LINKED ISSUE #9031: [2.1 & 3.0]: No direct access error - for PMs and messages - still there
### Basic Information

Sorry I was unavailable for a while (& I'm about to head out of town again...).  I just tried this again, and it's still happening, even with #8996 merged.

See the log below.

This affects messages as well as PMs.

For the record, my PR as originally submitted deals with this (#8991 and #8996 as originally submitted).

See also #8993 (which I thought was closed...??)

### Steps to reproduce

1. Send a PM (or start a new topic) with the subject `Hello World \`
2. Add the word `I'll` as the body
3. Press SEND

### Expected result

PM sent

### Actual result

Image

### Version/Git revision

2.1 - current GitHub as of today

### Database Engine

All

### Database Version

8.4.4

### PHP Version

8.4.5

### Logs

```shell
No direct access...

			INSERT INTO smf_personal_messages(`id_pm_head`, `id_member_from`, `deleted_by_sender`, `from_name`, `msgtime`, `subject`, `body`)
			VALUES
				(0, 1, 0, SUBSTRING('admin_shawnb', 1, 255), 1762455918, SUBSTRING('Hello world \\', 1, 255), SUBSTRING('I'll', 1, 65534))
Function: sendpm
```

### Additional Information

_No response_

sbulen:
Fixes #9031 

Third attempt!  I believe this one addresses concerns raised by @Sesquipedalian in the discussion on #8996 .

The only quirk in this version is that it will strip empty strings from the query prior to the search for forbidden syntax, which isn't a problem given what it's looking for.

Spent time testing with:
 - The primary test case in the bug reports (subject `Hello World \`, body `I'm`), as a PM and as a new topic
 - The smileys test
 - Fetching SMF files (the original, original bug...)
 - Several attempts to break it with content with repeating quotes & slashes & apostrophes & entities of various lengths, etc.
 - Observing a few dozen cleaned misc queries just browsing around the forum
 - An install; running all scheduled tasks; several admin functions

### DIFF
--- a/Sources/Subs-Db-mysql.php
+++ b/Sources/Subs-Db-mysql.php
@@ -368,12 +368,14 @@ function smf_db_query($identifier, $db_string, $db_values = array(), $connection
 
 	// Comments that are allowed in a query are preg_removed.
 	static $allowed_comments_from = array(
+		'~\'\X*?\'~s',
 		'~\s+~s',
 		'~/\*!40001 SQL_NO_CACHE \*/~',
 		'~/\*!40000 USE INDEX \([A-Za-z\_]+?\) \*/~',
 		'~/\*!40100 ON DUPLICATE KEY UPDATE id_msg = \d+ \*/~',
 	);
 	static $allowed_comments_to = array(
+		' %s ',
 		' ',
 		'',
 		'',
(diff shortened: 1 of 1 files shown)


jdarwood007:
I requested @Sesquipedalian but haven't heard from him.  Beyond my testing, which covers the same tests as those done in this PR, I don't have enough regex skills to fully understand it, see any side effects, or identify edge cases not covered.

if nobody gives me a reason not to merge this tomorrow, then it will be done, and I will build and start testing the 2.1 patch file again.

Sesquipedalian:
I don't have capacity to test this at the moment. It looks like this version should address my concerns with @sbulen's previous versions, though. So I think we can merge it and then see if the beta testers find any problems with it.

jdarwood007:
@sbulen can you make the 3.0 version as well?
want to ensure we have both since the issue affects both and the issue opened for it needs to reference both before it closes.

PR #9047, opened 2025-11-14
[2.1] mail pref error

### LINKED ISSUE #9046: [2.1 & 3.0]: Undefined variable $pref, CreatePost-Notify.php - Line 612 & 590
### Basic Information

I am getting this using current GH, but not in 2.1.6.  Pretty sure it needs to be fixed prior to 2.1.7.

If you look at the if here:
https://github.com/SimpleMachines/SMF/blame/f820d70b1f5886f6e5fd814de8145f772e4a10db/Sources/tasks/CreatePost-Notify.php#L524

...there is a nested if/elseif, without an else anywhere...  

I.e., $pref might not get set at all.

_I think the error happens any time a reply is made to any unwatched topic within a watched board.  There should be no notifications when this happens._  

### Steps to reproduce

1. Install a new forum
2. Add a new user
3. Logon as that new user, & ask for Alerts and Email Notifications for "General Discussion"
4. Log out & log back in as admin
5. Add a new topic to "General Discussion"  (new user will properly receive an alert)
6. Now, post a reply to that new topic - error will occur

### Expected result

No alert.

### Actual result

Image

### Version/Git revision

2.1 - current GitHub

### Database Engine

MySQL

### Database Version

8.4.4

### PHP Version

8.4.5

### Logs

```shell

```

### Additional Information

_No response_

sbulen:
Fixes #9046

### DIFF
--- a/Sources/tasks/CreatePost-Notify.php
+++ b/Sources/tasks/CreatePost-Notify.php
@@ -554,6 +554,9 @@ class CreatePost_Notify_Background extends SMF_BackgroundTask
 				if (empty($modSettings['disallow_sendBody']) && !empty($this->prefs[$member_id]['msg_receive_body']))
 					$message_type .= '_body';
 			}
+			// No notification request for this user for this board/topic.
+			else
+				continue;
 
 			// We need to fake some of $user_info to make BBC parsing work correctly.
 			if (isset($user_info))


sbulen:
I confirmed this addresses the errors, and also the spurious emails.

Sesquipedalian:
~Do we need a version of this for 3.0? If so, can you please submit one, @sbulen?~

Never mind. I see it has already been done.

PR #9064, opened 2025-11-22
[3.0] Sqlite caching didn't always create a table

jdarwood007:
Due to the journal mode changes, it seems that the table doesn't always get created because the file size can be larger than 0.  But it still seems to be exactly 4096. Changed the check and added an exists to trying to create the table/index.

### DIFF
--- a/Sources/Cache/APIs/Sqlite.php
+++ b/Sources/Cache/APIs/Sqlite.php
@@ -98,9 +98,9 @@ class Sqlite extends CacheApi implements CacheApiInterface
 				$this->cacheDB->exec('PRAGMA journal_mode = wal;');
 			}
 
-			if (filesize($database) == 0) {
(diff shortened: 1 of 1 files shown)


sbulen:
Looks like mine is 16K...
image

jdarwood007:
Is that the initial empty file with no tables?

I would have the file created, but no tables, and the file size was reported at 4096.  That was the minimum file size, empty with no tables.  Its possible this is tied into the block size on the disk, but my linux system says that 512 kb.  The sqlite documentation says 512 as well.  I'm not sure why I hit this.

You can test by deleting the cache file and see what it recreates to, then check its contents to see if it contains the table and index.

jdarwood007:
Going to abandon this.  The reason its 4096 and 16 kb on your side is the block size on the disk.  The reason the 512kb is the block size.  There isn't a method that seems worth determining the block size.  This will just be one of those known issues with caching that results in the needed to rebuild the cache by deleting the cache database.

PR #9075, opened 2026-01-17
[3.0] Fix invalid timezone

### LINKED ISSUE #9037: [3.0]: Unknown or bad timezone
### Basic Information

`DateTimeZone::__construct(): Unknown or bad timezone ()`

### Steps to reproduce

1. Go to the member list (`?action=mlist`)

### Expected result

_No response_

### Actual result

_No response_

### Version/Git revision

2ed6754

### Database Engine

MySQL

### Database Version

11.8.3-MariaDB

### PHP Version

8.2.29

### Logs

```shell

```

### Additional Information

Possible fix:

User.php, replace
```php
'local_time' => Time::create('now', $this->timezone)->format(null, false),
```
with
```php
'local_time' => Time::create()->format(),
```

jdarwood007:
The main issue is the user has an invalid timezone in their profile.  Doing a check for an empty profile is simple, but a garbage entry is another issue.

https://www.php.net/manual/en/datetimezone.construct.php

According to the documentation an invalid timezone throws an error. Our code is assuming in many places that a valid time zone will be returned.  The exception is not being handled

Create a wrapper and attempt to fallback the forum default time zone if we fail to get a valid user time zone Anywhere we use UTC, I ignored using the wrapper.

Fixes #9037

### DIFF
--- a/Sources/Actions/Calendar.php
+++ b/Sources/Actions/Calendar.php
@@ -860,7 +860,7 @@ class Calendar implements ActionInterface, Routable
 		$occurrences = [];
 
 		$one_day = new \DateInterval('P1D');
-		$tz = new \DateTimeZone(User::getTimezone());
+		$tz = TimeZone::create(User::getTimezone());
 		$high_date = (new \DateTimeImmutable($high_date . ' +1 day'))->format('Y-m-d');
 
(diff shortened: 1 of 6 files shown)


Sesquipedalian:
SMF\TimeZone::create() should return an instance of SMF\TimeZone, not \DateTimeZone.

PR #9077, opened 2026-01-17
[3.0] Database Initializer support

### LINKED ISSUE #9011: [3.0]: pg install: members table not created due to "ERROR:  function indexable_month_day(date) does not exist"
### Basic Information

Installing pg, there a syntax error on the members table, and the members table is not created.

Two related issues here:
1. The function indexable_month_day() is present in 2.1 & upon upgrades to 3.0, but it's not present on installs.
2. This error is not displayed (I can only see it via debugging utilities).  

Error text is as follows:

> ERROR:  function indexable_month_day(date) does not exist
> LINE 1: ... INDEX ato_members_idx_birthdate2 ON ato_members (indexable_...
> HINT:  No function matches the given name and argument types. You might need to add explicit type casts.

**_Note:  None of the pg functions are created in 3.0 upon install..._**

### Steps to reproduce

1.Install pg, the members table won't exist

2.NOTE:  DROP and re-CREATE the database before starting the test if you are reusing the db.  Otherwise the functions may still be present from a prior install/upgrade.  To test the install, you need a fully empty database, including no functions, sequences, etc.  

Much of this stuff is still laying around, even if you drop all the TABLEs in a DB. 

### Expected result

All tables created & populated; SMF pg functions created

### Actual result

The members table does not exist.  

You will see this on the Database Population screen during installation:

```
Some of the queries were not executed properly. This could be caused by an unsupported (development or old) version of your database software.

Technical information about the queries:

    ERROR: relation "ato_members" does not exist

```

**_Note:  None of the pg functions are created in 3.0 upon install..._**

### Version/Git revision

3.0 Alpha 4 - current GitHub

### Database Engine

PostgreSQL

### Database Version

14

### PHP Version

8.4.5

### Logs

```shell
ERROR:  function indexable_month_day(date) does not exist
LINE 1: ... INDEX ato_members_idx_birthdate2 ON ato_members (indexable_...
                                                             ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

### Additional Information

_No response_

jdarwood007:
PostgreSQL needs to initialize some functions and operators before it works in SMF

Added support for this and ensured that on upgrade, we run this.

Fixes #9011

### DIFF
--- a/Sources/Db/Schema/Table.php
+++ b/Sources/Db/Schema/Table.php
@@ -605,7 +605,7 @@ class Table
 			$fully_qualified_class_name = __NAMESPACE__ . '\\' . $schema_version . '\\Initialize\\' . $title;
 
 			if (!class_exists($fully_qualified_class_name)) {
-				return ['class'];
+				return [];
 			}
 
 			/**
--- a/Sources/Db/Schema/Table.php
+++ b/Sources/Db/Schema/Table.php
@@ -592,4 +592,30 @@ class Table
 
 		return $tables;
 	}
+
+	/**
+	 * Gets all known table schemas.
+	 *
+	 * @return array All known table schemas.
+	 */
+	final public static function getInitializers(string $schema_version, string $title): array
+	{
+		if (file_exists(__DIR__ . '/' . $schema_version . '/Initialize/' . $title . '.php')) {
+
+			$fully_qualified_class_name = __NAMESPACE__ . '\\' . $schema_version . '\\Initialize\\' . $title;
+
+			if (!class_exists($fully_qualified_class_name)) {
+				return ['class'];
+			}
+
+			/**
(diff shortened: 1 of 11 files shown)


Sources/Db/Schema/Table.php
```diff
@@ -592,4 +592,30 @@ final public static function getAll(string $schema_version): array
...
+	 * @return array All known table schemas.
+	 */
+	final public static function getInitializers(string $schema_version, string $title): array
+	{
+		if (file_exists(__DIR__ . '/' . $schema_version . '/Initialize/' . $title . '.php')) {
+
+			$fully_qualified_class_name = __NAMESPACE__ . '\\' . $schema_version . '\\Initialize\\' . $title;
+
+			if (!class_exists($fully_qualified_class_name)) {
+				return ['class'];
```
jdarwood007:
```suggestion
				return [];
```

PR #9082, opened 2026-01-17
[2.1] Fix IPv6 reverse hostname display on member profiles

MegaBrutal:
Hostnames for IPv6 addresses didn't appear on member profiles because the regex to filter invalid IP values only matched for IPv4 addresses. I changed the test to the PHP [filter_var()](https://www.php.net/manual/en/function.filter-var.php) function which should be more robust.

### DIFF
--- a/Sources/Profile-View.php
+++ b/Sources/Profile-View.php
@@ -86,7 +86,7 @@ function summary($memID)
 	if (allowedTo('moderate_forum'))
 	{
 		// Make sure it's a valid ip address; otherwise, don't bother...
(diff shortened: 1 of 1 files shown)


jdarwood007:
@MegaBrutal 
Thank you for your contribution.

Do you have the related PR for SMF 3.0?

We can look at including this in a future 2.1 update, but we also need to patch it in 3.0.

MegaBrutal:
@jdarwood007 SMF 3.0 already uses ``filter_var``, no fix is needed there.
https://github.com/SimpleMachines/SMF/blob/4dd6381088e0c6f7a1a1e5bfcd10b67bd24b7130/Sources/Actions/Profile/Summary.php#L107

PR #9090, opened 2026-01-24
[3.0] Add doAutoSubmit

jdarwood007:
Pulled from #7933

### DIFF
--- a/Themes/default/UpgradeTemplate.php
+++ b/Themes/default/UpgradeTemplate.php
@@ -93,22 +93,7 @@ class UpgradeTemplate extends MaintenanceTemplate
 
 		echo '
 			<script>
-				let countdown = ' . (Maintenance::$tool->isDebug() ? '-1' : '3') . ';
-				let dontSubmit = false;
-
-				function doAutoSubmit()
-				{
-					if (countdown == 0 && !dontSubmit) {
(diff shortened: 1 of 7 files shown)


Themes/default/UpgradeTemplate.php
```diff
@@ -93,22 +93,7 @@ public static function lower(): void
...
-					} else if (countdown == -1) {
-						return;
-					}
-
-					document.getElementById("contbutt").value = "', Lang::getTxt('action_continue', file: 'Maintenance'), ' (" + countdown + ")";
-					countdown--;
-
-					setTimeout("doAutoSubmit();", 1000);
-				}
+				setTimeout(\'doAutoSubmit(' . (Maintenance::$tool->isDebug() ? '-1' : '3') . ', "', Lang::getTxt('action_continue', file: 'Maintenance'), '", "upform");\', 1000);
```
live627: this should be moved and should specify name of b utton

https://github.com/SimpleMachines/SMF/blob/92e5b4bc172642531c579a5e76a1f090e4646394/Themes/default/UpgradeTemplate.php#L114-L120

jdarwood007: Will fix the form, but I can't honestly remember why I have that defer there.  I added a todo and someyear we will look at it.

PR #9108, opened 2026-02-10
[3.0] Dependency Injection migration: Service Provider and first migration

MissAllSunday:
This PR adds a way to register services as dependencies and migrates ErrorHandler to be served via the Container

Heres a Graph TB for reference on how ErrorHandler is been used and how backward compatibility is achieved by leaving the Original ErrorHandler as a Facade and mark it as deprecated in favor of the new Service. I'm aware original ErrorHandler has some duplicated code for calling the container, however, this is done to maximize backward compatibility.

```  mermaid

graph TB
    subgraph "Legacy Code"
        A[Legacy Code Calls]
        A1["ErrorHandler::log()"]
        A2["ErrorHandler::fatal()"]
        A3["ErrorHandler::catch()"]
        A4["new ErrorHandler()"]
    end
    
    subgraph "Facade Layer"
        B[ErrorHandler Class]
        B1[getService Method]
        B2[Static Methods]
    end
    
    subgraph "DI Container"
        C[Container::getInstance]
        D[ServiceProvider]
        E[ServicesList.php]
    end
    
    subgraph "Service Layer"
        F[ErrorHandlerService]
        F1[handleError]
        F2[log]
        F3[fatal]
        F4[catch]
    end
    
    A --> B
    A1 --> B2
    A2 --> B2
    A3 --> B2
    A4 --> B
    
    B --> B1
    B2 --> B1
    
    B1 -->|Try Container First| C
    C --> D
    D --> E
    E -->|Returns Shared Instance| F
    
    B1 -->|Fallback if Container Unavailable| F
    
    F --> F1
    F --> F2
    F --> F3
    F --> F4
    
    style B fill:#e1f5ff
    style F fill:#fff4e1
    style C fill:#e8f5e9
    style A fill:#fce4ec
``` 

Of course arch and folders/naming convention should be subject to modifications to better suit current patterns if needed.

### DIFF
--- a/index.php
+++ b/index.php
@@ -133,7 +133,7 @@ call_user_func(function () {
 	SMF\Config::$loader->setPsr4('SMF\\', $sourcedir);
 
 	// Initialize the container.
-	\SMF\Infrastructure\Container::init();
+	SMF\Infrastructure\Container::init();
 
 	// Ensure $db_last_error is set, too.
 	SMF\Config::getDbLastError();
--- a/index.php
+++ b/index.php
@@ -133,7 +133,7 @@ call_user_func(function () {
 	SMF\Config::$loader->setPsr4('SMF\\', $sourcedir);
 
 	// Initialize the container.
-	SMF\Container::init();
+	\SMF\Infrastructure\Container::init();
 
 	// Ensure $db_last_error is set, too.
 	SMF\Config::getDbLastError();
--- a/Sources/ErrorHandler.php
+++ b/Sources/ErrorHandler.php
@@ -19,201 +19,222 @@ use SMF\Services\ErrorHandlerService;
  */
 class ErrorHandler
 {
-    /**
-     * @var array
-     *
-     * What types of categories do we have for logging errors?
-     *
-     * @deprecated Use ErrorHandlerService::$known_error_types instead.
-     *             Kept for backward compatibility.
-     */
-    public static array $known_error_types = [
-        'general',
-        'critical',
-        'database',
-        'undefined_vars',
-        'user',
-        'ban',
-        'template',
-        'debug',
-        'cron',
-        'paidsubs',
-        'backup',
-        'login',
-    ];
+	/**************************
+	 * Public static properties
+	 **************************/
 
-    /**
-     * @var ErrorHandlerService|null
-     *
-     * Cached service instance.
-     */
-    protected static ?ErrorHandlerService $service = null;
+	/**
+	 * @var array
+	 *
+	 * What types of categories do we have for logging errors?
+	 *
+	 * @deprecated Use ErrorHandlerService::$known_error_types instead.
+	 *             Kept for backward compatibility.
+	 */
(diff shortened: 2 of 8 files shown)


dragomano:
Examples of two different container implementations to demonstrate just how powerful this thing is. Both versions are fully working:

- [League Container](https://gist.github.com/dragomano/d4cb242d374e0a83566805afc631ab72)
- [Laminas Service Manager](https://gist.github.com/dragomano/87319a52eecf738ec35092ed6ced7329)

index.php
```diff
@@ -133,7 +133,7 @@
 	SMF\Config::$loader->setPsr4('SMF\\', $sourcedir);
 
 	// Initialize the container.
-	SMF\Container::init();
+	\SMF\Infrastructure\Container::init();
```
Sesquipedalian:
```suggestion
	SMF\Infrastructure\Container::init();
```

nit

Sesquipedalian (dismissed):
In addition to fixing the noted minor nit, please run PHP-CS-Fixer on all modified and added files in order to make them compliant with the SMF code standard.

(It appears that our CI script that is supposed to run PHP-CS-Fixer as a linter is not working right now. If it were, this issue would have been flagged automatically.)

index.php
MissAllSunday: I applied the change.  As for the lint, unfortunately, it seems composer lint has been setup to work with a git diff command to get the staged changes and since I already committed them the linter no longer sees them.

Running php-cs-fixer without the git diff constrain yields more than 1k files needed to be corrected so perhaps it would be a good idea to create a new branch just for linting and start fresh.

That or change the instruction to git diff release-3.0 to make sure the linter works on all changes against the latest release-3.0 commit.

MissAllSunday: I tried to manually execute php-cs-fixer on the modified files of this PR but it seems there are some discrepancies between SMF needed PHP version (8.0)  and what the linter needs (8.1)

When I tried to force run the linter on php 8 via docker:

```
docker run --rm -it -v "$(pwd)":/app -w /app php:8.0-cli sh -c "
    apt-get update -qq && apt-get install -y git unzip -qq > /dev/null && \
    curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer -q && \
    git config --global --add safe.directory /app && \
    composer install --no-interaction --quiet && \
    ./vendor/bin/php-cs-fixer fix --config=.php-cs-fixer.dist.php --allow-risky=yes --show-progress=dots -vvv \
    Sources/Infrastructure/Container.php
```

 I get the following error:

```
[Error]                                                  
        Undefined constant "SMF\Fixer\ClassNotation\T_READONLY" 
```
        
Which indicates some of the rules for php-cs-fixer are too new for the PHP version SMF needs

Sesquipedalian: Hm. I'll look into that. Thanks, @MissAllSunday.

Sesquipedalian (approved):

PR #9110, opened 2026-02-13
[3.0] Fixes PHP-CS-Fixer workflow 

Sesquipedalian:


jdarwood007:
@Sesquipedalian 
I had migrated away from tj-actions because of the security issue that occurred: https://github.com/advisories/ghsa-mrrh-fwg8-r2c3

In reality, we have ways to do the same thing without adding a dependency. 

Also, this is why we don't use tag ids when referencing dependencies in our CI, but rather the commit hash.  Because as the security issue above, the tag was overwritten, allowing the vulnerability.

Sesquipedalian:
I'll try reverting that part of the change and test whether it still works.

Sesquipedalian:
It did not work. That shell code did not do the job it needed to.

PR #9123, opened 2026-02-24
[3.0] Demotes smf_crc32() to a backward compatibility function

Sesquipedalian:
1. [Removes unnecessary smf_crc32() calls in SMF\Graphics\Gif\File](https://github.com/SimpleMachines/SMF/pull/9123/commits/3661a5ca45dd0ceb4a4e54b90920022509910f35)
    - Although the integer value returned by crc32() can be different on 32-bit vs. 64-bit systems, `pack('N', crc32($tmp)` will always result in the same value on both 32-bit and 64-bit systems. This is because pack() only cares about the raw binary value, not the integer interpretation. Thus, there is no need to call smf_crc32() here; we can just use normal crc32() on both 32-bit and 64-bit systems.
    - Since SMF\Graphics\Gif\File is only retained for backward compatibility and is itself never used by SMF 3.0, there is no easy way to test this change within SMF itself. However, it is possible to verify that `pack('N', smf_crc32($tmp))` and `pack('N', crc32($tmp))` always produce the same results on both 32-bit and 64-bit platforms by going to [https://3v4l.org/#live](https://3v4l.org/#live) and running the following script:
	```php
	<?php
	
	echo PHP_INT_SIZE === 4 ? '32-bit' : '64-bit', PHP_EOL;
	
	$str = 'derp';
	
	$p1 = pack('N', smf_crc32($str));
	$p2 = pack('N', crc32($str));
	
	var_export($p1 === $p2);
	
	function smf_crc32($number): int
	{
		$crc = crc32($number);
	
		// On a 32-bit system, PHP_INT_SIZE === 4.
		// On a 64-bit system, PHP_INT_SIZE === 8.
		if (PHP_INT_SIZE === 8 && $crc & 0x80000000) {
			$crc ^= 0xffffffff;
			$crc += 1;
			$crc = -$crc;
		}
	
		return $crc;
	}
	```
	At least as of Feb 24, 2026, when you first go to https://3v4l.org/#live and paste in that script, the live preview will be initially executed using a 32-bit build of PHP. If you then manually click the `eval();` button, it will execute again using a 64-bit build of PHP.
2. [Removes unnecessary smf_crc32() call in PackageUtils::readTgzData()](https://github.com/SimpleMachines/SMF/pull/9123/commits/785eca45823dfbbc62ecf77126586da1d6bf5eb4)
    - To test this, simply upload any .tar.gz package via SMF's package manager and then go to the package manager's Browse Packages page. If the .tar.gz package appears in the list, then you have confirmed that this commit works.
3. [Deprecates smf_crc32()](https://github.com/SimpleMachines/SMF/pull/9123/commits/7a382b1cf12c54b573519ce509e06a4f1e25f679)
    - We no longer use it for anything in SMF itself.
    - This commit also improves the documentation for the smf_crc32() function, explaining what the function does and why it has been deprecated.

### DIFF
--- a/Sources/Graphics/Gif/File.php
+++ b/Sources/Graphics/Gif/File.php
@@ -23,9 +23,6 @@ declare(strict_types=1);
 
 namespace SMF\Graphics\Gif;
 
-use SMF\Config;
-use SMF\Sapi;
-
 class File
 {
 	/*******************
@@ -149,13 +146,13 @@ class File
 		// Now, we want the header...
 		$out .= "\x00\x00\x00\x0D";
 		$tmp = 'IHDR' . pack('N', (int) $this->header->m_nWidth) . pack('N', (int) $this->header->m_nHeight) . "\x08\x03\x00\x00\x00";
-		$out .= $tmp . pack('N', smf_crc32($tmp));
+		$out .= $tmp . pack('N', crc32($tmp));
 
 		// The palette, assuming we have one to speak of...
 		if ($colors > 0) {
 			$out .= pack('N', (int) $colors * 3);
 			$tmp = 'PLTE' . $pal;
-			$out .= $tmp . pack('N', smf_crc32($tmp));
+			$out .= $tmp . pack('N', crc32($tmp));
 		}
 
 		// Do we have any transparency we want to make available?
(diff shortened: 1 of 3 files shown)


Sources/Graphics/Gif/File.php
```diff
@@ -149,13 +146,13 @@ public function get_png_data(string $background_color): string|bool
 		// Now, we want the header...
 		$out .= "\x00\x00\x00\x0D";
 		$tmp = 'IHDR' . pack('N', (int) $this->header->m_nWidth) . pack('N', (int) $this->header->m_nHeight) . "\x08\x03\x00\x00\x00";
-		$out .= $tmp . pack('N', smf_crc32($tmp));
+		$out .= $tmp . pack('N', crc32($tmp));
```
jdarwood007, line 149: Are there any worries about switching to just crc32 for our image handling?

Sesquipedalian: Nope. As shown in the testing instructions I gave above, `pack('N', crc32($tmp))` and `pack('N', smf_crc32($tmp))` produce identical results. 

This is because the 'N' format treats the value as a 32-bit unsigned long, which means that it truncates all the extra bits that `smf_crc32()` prepends on 64-bit systems, leaving only the original 32-bit checksum returned by plain `crc32()`. This in turn means that calling `smf_crc32()` here is and has always been pointless.

PR #9127, opened 2026-02-25
[2.1] Bots - view likes attacks

### LINKED ISSUE #9112: [2.1]: Feature request - ability to disable public access to likes list
### Basic Information

Hello.

I would like to request a new option in the SMF 2.1 likes system:

ability to hide the list of users who liked a post from guests (non-logged users).

Discussion on this topic:

https://www.simplemachines.org/community/index.php?topic=593808.0

Bots are abusing action=likes;sa=view to generate massive session creation and error logs. SMF 2.1 needs an option to prevent guests from accessing the likes list without breaking functionality for logged users.

The only current mitigation options are:

blocking via .htaccess:
https://www.simplemachines.org/community/index.php?msg=4192411

removing links from templates
manually modifying Likes.php
https://www.simplemachines.org/community/index.php?topic=593808.msg4198896#msg4198896

Behaviour

When disabled (recommended default):

guests cannot access action=likes;sa=view

the link to the likes list is not shown to guests

optionally show a message:

You must be logged in to see who liked this post

This would eliminate the botnet attack vector while preserving flexibility for administrators who want public likes visibility.

SMF 2.1 should include this protection because bots already know and target the likes endpoint.

### Steps to reproduce

1.
2.

### Expected result

_No response_

### Actual result

_No response_

### Version/Git revision

2.1.6

### Database Engine

MySQL

### Database Version

_No response_

### PHP Version

_No response_

### Logs

```shell

```

### Additional Information

_No response_

sbulen:
This PR removes the link allowing users to drill down to see who liked a post.  Only guests & bots are affected.

This link is a bot magnet during botnet attacks.  During some attacks, these are the ONLY requests made - in the tens of thousands.  Note that the request includes the session var & value in the URL.  (I believe this is the only guest link that does so...???)  When bots pass this bogus session info, the existing session is destroyed and a new one is created, ultimately causing TWO session writes for each bot request.  I.e., double the impact.  

I believe this should be removed from guests/bots altogether.  If that user wants to drill down, they can register.

Fixes #9112

For more discussion see:
https://www.simplemachines.org/community/index.php?topic=590069.0

Feedback welcome.

### DIFF
--- a/Themes/default/Display.template.php
+++ b/Themes/default/Display.template.php
@@ -462,7 +462,7 @@ function template_main()
  */
 function template_single_post($message)
 {
-	global $context, $settings, $options, $txt, $scripturl, $modSettings;
+	global $context, $settings, $options, $txt, $scripturl, $modSettings, $user_info;
 
 	$ignoring = false;
 
@@ -847,7 +847,8 @@ function template_single_post($message)
 			$base .= (isset($txt[$base . $count])) ? $count : 'n';
 
 			// Remove link if no cookies; session reference won't work
-			if (empty($_COOKIE))
+			// Also, don't show link to guests
+			if (empty($_COOKIE) || !empty($user_info['is_guest']))
 				$txt[$base] = preg_replace('~</?a\b[^>]*>~', '', $txt[$base]);
 
 			echo '


jsqx:
Will this fix be included in SMF 2.1.7?

sbulen:
> Will this fix be included in SMF 2.1.7?

Probably not.  First, I imagine there needs to be some internal discussion among the team.  There was a lot of debate in the past about what guests & likes & views.  IMO, the fact that it has become a bot magnet changes the discussion though, hence this PR.

Also, the next release has been in process for a while.  Scope has crept & a good case could be made to stop scope creep.

Also, it's possible the team comes back with "OK, but make it an option", due to concern about removing a feature.  

It's not even approved yet.

jdarwood007:

sbulen:

I think this is very low risk & very easy.  My only thought is some folks on the team might want this one to be an option.

I don't think it should be, but some might.  We'll see what the feedback is.

jdarwood007:
I've tagged this on the 3.0 project with the 2.1.8 milestone.  It makes sense to help with bots and performance.  Will test this later.

Sesquipedalian (approved):

PR #9128, opened 2026-02-25
[2.1] Bots - reduce topic view updates

sbulen:
This PR aims to minimize I/O during a bot attack by eliminating the tracking of guest & bot views in the smf_topics record.  Note this counter is incremented under all circumstances, and causes yet another I/O during bot attacks.

An option is added, allowing the user to turn this behavior on or off.  Users may wish to disable this only during high-CPU bot attacks.  

For more discussion see:

Feedback welcome.


jdarwood007:

sbulen:

This one has clear tradeoffs, but given there's a switch, folks can turn it on only when needed.

jdarwood007:
This is another one: do we need to implement the installer/upgrader logic if the feature is disabled by default?

I've tagged this on the 3.0 project with the 2.1.8 milestone.

Sesquipedalian (changes requested):
Please remove unnecessary installer and upgrader changes.

sbulen:
> Please remove unnecessary installer and upgrader changes.

> This is another one: do we need to implement the installer/upgrader logic if the feature is disabled by default?

I had deliberately set it up so behavior is unchanged if no var is found.  Felt safest.  Also, easiest to patch, no update needed.

But I did add to the installer & upgrader, mainly because I like seeing all the 'real' SMF vars doc'd in that fashion.  To my knowledge, we've always added them.

I'll go either way, just letting you know my thoughts.

Sesquipedalian:
There are already a number of other settings that aren't included in the install/upgrade scripts because they are disabled by default.

sbulen:
OK, I'll remove.

sbulen:
Done.

jdarwood007 (approved):
Approved, will wait for a 3.0 version before I merge.

PR #9137, opened 2026-03-05
[3.0] Moves input sanitization of BBCode (a.k.a. pre-parsing) from SMF\Msg to SMF\Parser

Sesquipedalian:
1. [Moves Msg::un_preparsecode() to Parser::getEditableString()](https://github.com/SimpleMachines/SMF/commit/dd6a6bef88cc560de9c9c7eb877fced9862bf39f)
    - Renamed the method because `getEditableString` is more descriptive of what the method actually does.
2. [Moves Msg::preparsecode() to Parser::sanitize()](https://github.com/SimpleMachines/SMF/commit/0f8752fbb4c4af122eca49735b051a62ab7e5e7f)
    - Again, renamed to give a better indication of what the method really does.

**METHOD SIGNATURE CHANGED:** Whereas `Msg::preparsecode()` received the `$message` as a reference and returned `void`, Parser::sanitize() receives the `$message` as a normal argument and returns `string`.


Sesquipedalian:
Hm. The CI workflows are not running for some reason.

PR #9144, opened 2026-03-11
[2.1] Dont let bots flood logs

### LINKED ISSUE #9142: [2.1 & 3.0] "Unable to verify referring URL" errors can flood the SMF error log
### Basic Information

When bots attempt to spoof the referrer field, this can trigger "Unable to verify referring URL" errors.  

Normally, SMF smartly avoids logging all of these, otherwise, you'd be at the mercy of the bots.  And they're merciless.

It looks like SMF attempts to only log these errors upon receipt of valid requests where invalid referrers are detected, and it doesn't log when there are invalid requests.  But...  The problem is that bots can still flood the logs with valid looking requests.

Related: https://github.com/SimpleMachines/SMF/issues/9112

Note that #9112 requests we should disable the view likes link for guests, which is good.  This particular issue is subtly different - maybe we shouldn't allow the log to be flooded with errors - even when bots attempt valid links.  Two aspects of the same issue - get rid of the tempting link, & get rid of these errors in the log.

Which begs the question - do we need "Unable to verify referring URL" errors logged at all - ever???  Why???  My current thinking is to stop logging them altogether.

If you give a bot the ability to generate errors at will, they will do so, and that will eat up your resources.

I have learned a few lessons testing #9126 that I will log separately.  Pretty sure these affect both 2.1 & 3.0.

### Steps to reproduce

1.
2.

### Expected result

_No response_

### Actual result

_No response_

### Version/Git revision

2.1 & 3.0 Alpha 4

### Database Engine

All

### Database Version

_No response_

### PHP Version

8.4

### Logs

```shell

```

### Additional Information

_No response_

sbulen:
Partial for #9142 

2.1 portion only.  Just don't log "Unable to verify referring URL" errors.  This gives a handle for bots to flood your logs.  Plus, I'm not sure the info is usable or actionable.  Note also that the other similar session errors throughout this logic are not logged.  

I kept the one a few lines below this as logged (line 691) - that condition appears to be more of a trap for an internal issue.  So it's kept, for now...

If approved, I'll submit the 3.0 equivalent.

### DIFF
--- a/Sources/Security.php
+++ b/Sources/Security.php
@@ -680,7 +680,6 @@ function checkSession($type = 'post', $from_action = '', $is_fatal = true)
 		if (isset($parsed_url['host']) && strtolower($referrer['host']) != strtolower($parsed_url['host']) && strtolower($referrer['host']) != strtolower($real_host))
 		{
 			$error = 'verify_url_fail';
-			$log_error = true;
 		}
 	}
 


jdarwood007:
Makes sense.  I was going to suggest we do `!$context['user']['is_guest']`, but then I thought, why, its garbage either way.

sbulen:
Yep.

Sesquipedalian (approved):

jdarwood007:
@sbulen Can you submit a PR for 3.0 on this change as well?

sbulen:
> @sbulen Can you submit a PR for 3.0 on this change as well?

Will do.

Issue #9146, opened 2026-03-12
[3.0]: Guests cannot access forum


sbulen:
### Basic Information

New install, fresh from GH as of today.

On a new install, guests cannot access the forum:
"The topic or board you are looking for appears to be either missing or off limits to you."

Kinda holding up testing of some other items.

### Steps to reproduce

1. Install 3.0
2. Allow guest access (default)
3. Browse as guest

### Version/Git revision

3.0 Alpha 4 (current)

### Database Engine

All

### Database Version

8.4.5

### PHP Version

8.5

### Logs

```shell

```

Sesquipedalian:
To clarify, are you seeing this message along with a login form or not?

sbulen:
Guest access is allowed - no login required.

All default permissions.  No ACP changes.

Just install & go there as a guest.

Sesquipedalian:
Yes, but when the error message is displayed, is it shown along with a login form or without one? 

Basically, I'm trying to figure out whether you are passing through User::$me->kickIfGuest() or just being sent through ErrorHandler::fatalLang().

sbulen:
Just did a brand new install, current GH as of this moment.  Install went clean.  

BTW:  Install log looks excellent...

**Logged off as admin; now guest.  Guest can see board index just fine.**  

<img width="819" height="642" alt="Image" src="https://github.com/user-attachments/assets/74f8b712-dd01-4dc0-8145-bf4928e1057c" />


**But once I click on "General Discussion", I get the following.**

<img width="1037" height="673" alt="Image" src="https://github.com/user-attachments/assets/af296f3d-c26a-46a0-a0fc-4cf94006203c" />

Sesquipedalian:
Okay, that's helpful. I now know the path that you are passing through. Thanks!

sbulen:
Found this in the PHP error log:

```
[19-Mar-2026 23:16:24 UTC] PHP Fatal error:  Uncaught Error: Typed static property SMF\Db\DatabaseApi::$db must not be accessed before initialization in D:\wamp64\www\84van30\Sources\Config.php:1078
Stack trace:
# 0 D:\wamp64\www\84van30\Sources\Maintenance\Tools\ToolsBase.php(644): SMF\Config::reloadModSettings()
[...]

PR #9124, opened 2026-02-25
[2.1] Bots - reduce permissions queries

sbulen:
This PR aims to minimize unnecessary I/O by not doing moderator permissions lookups for guests and bots, who we already know will not have any moderator or admin privileges...  This can add a lot of unnecessary permissions queries during normal operation, but especially during bot attacks.

I've been running this code on my prod forum with no issues.

If this is approved, I can submit a 3.0 version.  

For more discussion see:
https://www.simplemachines.org/community/index.php?topic=592442.0

Feedback welcome.

PR #9149, opened 2026-03-13
[3.0] Bots - reduce permissions queries

sbulen:
3.0 version of #9124 - More discussion may be found there.

This PR aims to minimize unnecessary I/O by not doing moderator permissions lookups for guests and bots, who we already know will not have any moderator or admin privileges... This can add a lot of unnecessary permissions queries during normal operation, but especially during bot attacks.

For more discussion see:

NOTE I cannot test this due to #9146 , but it looks simple enough...

NOTE2 I cannot run php fixer due to errors, e.g., 

> 	The rules contain risky fixers ("native_function_invocation"), but they are not allowed to run. Perhaps you forget to use --allow-risky=yes option?

I hope I did the new syntax correctly.  

Feedback welcome.

### DIFF
--- a/Sources/User.php
+++ b/Sources/User.php
@@ -1411,7 +1411,11 @@ class User implements \ArrayAccess
 		}
 
 		// What groups can they moderate?
-		$group_query = $this->allowedTo('manage_membergroups') ? '1=1' : '0=1';
+		if ($this->is_guest) {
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
@sbulen, the logic is inverted in this PR. Both instances of `if ($this->is_guest)` should have been `if (!$this->is_guest)`. Can you please submit a followup PR to fix that?

sbulen:
Oof...

Issue #9147, opened 2026-03-12
[3.0]: Users cannot search forum - DB error


sbulen:
### Basic Information

Using current GH as of today, users cannot search:
"SMF\Db\APIs\MySQL::query(): Argument # 2 ($db_values) must be of type array, string given, called in D:\wamp64\www\84van30\Sources\Board.php on line 1123"

### Steps to reproduce

1. Install
2. Sign in as admin
3. Attempt to search

### Version/Git revision

3.0 Alpha 4 - current as of today

### Database Engine

All

### Database Version

8.4.5

### PHP Version

8.5

### Logs

```shell

```

PR #9150, opened 2026-03-13
[3.0] Bots - reduce topic view updates

sbulen:
3.0 version of #9128 - More discussion may be found there.

This PR aims to minimize I/O during a bot attack by eliminating the tracking of guest & bot views in the smf_topics record. Note this counter is incremented under all circumstances, and causes yet another I/O during bot attacks.

An option is added, allowing the user to turn this behavior on or off. Users may wish to disable this only during high-CPU bot attacks.

For more discussion see:


sbulen:
Note I'm having a hard time testing in 3.0 due to #9146 and #9147.  

I'm tied up right now...  Will hopefully be able to direct my attention there next week.  If anybody else has a theory, please dive in!

PR #9159, opened 2026-03-18
[3.0] Requires 64-bit PHP for installation

Sesquipedalian:
See [#9049](https://github.com/SimpleMachines/SMF/pull/9049#issuecomment-4085912000)


jdarwood007:
Adding some supporting documentation

https://wiki.php.net/rfc/drop_32bit_support

Most distros do not support 32-bit, and PHP itself is looking to discontinue it.  So the likelihood this affects many people is very low at this point.

PR #9172, opened 2026-03-27
`composer install -dev` → `COMPOSER_NO_DEV=0 composer install`

Geremia:


Geremia:
Actually, this is valid for Composer 2.9.5, but in ≥3, it's deprecated:
```
You are using the deprecated option "--dev". It has no effect and will break in Composer 3.
```

live627:
from https://stackoverflow.com/a/73351476

> You can set the environment variable `COMPOSER_NO_DEV` to `0` or `1` to change the default behaviour of `composer install` and `composer update`. 
> 
> see: documentation - [COMPOSER_NO_DEV](https://getcomposer.org/doc/03-cli.md#composer-no-dev)
> 
>  * If you want `composer install` to **not** install the dev dependencies by default
>    * `export COMPOSER_NO_DEV=1`
>  * If you want `composer install` to install dev dependencies (the default)
>    * `export COMPOSER_NO_DEV=0`
>    * or `unset COMPOSER_NO_DEV`
> 
> Depending on how you develop (i.e. in a container) there are various options to set a default value for the environment variable.
> 
> On the other hand you can instruct your engineer colleagues that do not require any dev dependencies to run the command:
> 
> ```bash
> composer install --no-dev
> 
> # .. or ..
> COMPOSER_NO_DEV=1 composer install
> ```
> ... instead of ...
> 
> ```bash
> composer install
> 
> # .. or ..
> COMPOSER_NO_DEV=0 composer install
> ```

Geremia:
@live627 Thanks. I've changed to `COMPOSER_NO_DEV=0` to instead.

jdarwood007:
So the default composer run is to include developer dependencies now?
Seems a bit backwards.

live627:
it's primarily a dev tool

Sesquipedalian (approved):

Sesquipedalian:
@jdarwood007, we should probably check our package build scripts to make sure any calls to composer are explicitly using `--no-dev`.

PR #9175, opened 2026-03-28
PHPMailer for SMTP

Geremia:
Run `composer require phpmailer/phpmailer` in root directory (not Sources) to setup.


Sesquipedalian:
Hi @Geremia! Thanks for this pull request. We're always glad to receive new contributions from the community!

Regarding this PR in particular, we won't be able to accept it because SMF 2.1 is no longer accepting new features such as this.

Instead, I invite you to submit a new pull request implementing this new feature in SMF **3.0**. You will most likely want to do so by either (a) making changes to SMF\MailAgent\APIs\SMTP so that it uses PHPMailer, or (b) adding a new MailAgent API class to use PHPMailer. Either way, you will also need to make some user interface changes so that forum admins can add the necessary OAuth credentials.

PR #9176, opened 2026-04-02
[2.1] Call autoload integration hook only once

live627:
Before this change, the autoloading function would take on average 5ms PER CLASS.  This quickly adds up when running moods that use many different classes to build pages.

### DIFF
--- a/index.php
+++ b/index.php
@@ -98,19 +98,21 @@ loadDatabase();
  *
  * @param string $class The fully-qualified class name.
  */
-spl_autoload_register(function ($class) use ($sourcedir)
-{
-	$classMap = array(
-		'ReCaptcha\\' => 'ReCaptcha/',
-		'MatthiasMullie\\Minify\\' => 'minify/src/',
-		'MatthiasMullie\\PathConverter\\' => 'minify/path-converter/src/',
-		'SMF\\Cache\\' => 'Cache/',
-	);
+$class_map = null;
+spl_autoload_register(function($class) use ($sourcedir, &$class_map) {
+	if ($class_map === null) {
+		$class_map = [
+			'ReCaptcha\\' => 'ReCaptcha/',
+			'MatthiasMullie\\Minify\\' => 'minify/src/',
+			'MatthiasMullie\\PathConverter\\' => 'minify/path-converter/src/',
+			'SMF\\Cache\\' => 'Cache/',
+		];
 
-	// Do any third-party scripts want in on the fun?
-	call_integration_hook('integrate_autoload', array(&$classMap));
+		// Call hook only ONCE, not on every autoload
+		call_integration_hook('integrate_autoload', array(&$class_map));
+	}
 
-	foreach ($classMap as $prefix => $dirName)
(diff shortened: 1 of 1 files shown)


index.php
```diff
@@ -98,19 +98,21 @@
...
+		$class_map = [
+			'ReCaptcha\\' => 'ReCaptcha/',
+			'MatthiasMullie\\Minify\\' => 'minify/src/',
+			'MatthiasMullie\\PathConverter\\' => 'minify/path-converter/src/',
+			'SMF\\Cache\\' => 'Cache/',
+		];
+
+		// Call hook only ONCE, not on every autoload
+		call_integration_hook('integrate_autoload', array(&$class_map));
+	}
```
Sesquipedalian:
```suggestion
spl_autoload_register(function($class) use ($sourcedir) {
	global $modSettings;

	static $hook_value = '';
	static $class_map = array(
		'ReCaptcha\\' => 'ReCaptcha/',
		'MatthiasMullie\\Minify\\' => 'minify/src/',
		'MatthiasMullie\\PathConverter\\' => 'minify/path-converter/src/',
		'SMF\\Cache\\' => 'Cache/',
	);

	// Do any third-party scripts want in on the fun?
	if ($hook_value !== (isset($modSettings['integrate_autoload']) ? $modSettings['integrate_autoload'] : ''))
	{
		call_integration_hook('integrate_autoload', array(&$class_map));
		$hook_value = $modSettings['integrate_autoload'];
	}
```

This will keep the logic entirely contained within the autoload function, and if `$modSettings['integrate_autoload']` is empty, it will reduce the number of calls to the hook to zero.

index.php
```diff
@@ -141,6 +143,10 @@
 $context = array();
 reloadSettings();
 
+// Force the autoload to refresh its class map now that enough data has
+// been loaded for the integration hook to work as expected.
+$class_map = null;
+
```
Sesquipedalian:
```suggestion
```

When the suggestion above is applied, these lines are no longer necessary.

Sesquipedalian (changes requested):

live627:
Much cleaner; thanks. And my integrations still work as expected.

Sesquipedalian:
Excellent!

index.php
```diff
@@ -120,15 +126,15 @@
...
 		// separators with directory separators in the relative class name, append
 		// with .php
-		$fileName = $dirName . strtr($relativeClass, '\\', '/') . '.php';
+		$filename = $dirname . strtr($relative_class, '\\', '/') . '.php';
 
 		// if the file exists, require it
-		if (file_exists($fileName = $sourcedir . '/' . $fileName))
+		if (file_exists($filename = $sourcedir . '/' . $filename))
 		{
 			require_once $fileName;
```
Sesquipedalian: Was about to approve when I noticed the mismatched case here. Technically it will still work because PHP variable names are not case sensitive, but it would be better to be consistent.

Sesquipedalian (changes requested):

Sesquipedalian (approved):

PR #9183, opened 2026-04-11
[3.0] Add logException to ErrorHandler

jdarwood007:
We can now call errorHandler::LogException(\Exception) and have SMF handle the work of putting the pieces together to log the data.

Additionally a new known_exception_types map is added to allow us to natively map some known or excepted types to the known_error_types.  This wasn't passed into our hooks as we want to carefully add to this list since a generalization of a exception class to a error type may lead to false categorization.

### DIFF
--- a/Sources/Services/ErrorHandlerService.php
+++ b/Sources/Services/ErrorHandlerService.php
@@ -264,7 +264,6 @@ class ErrorHandlerService
 		// Exceptions may get mapped back into our common error types.
 		if (isset($this->known_exception_types[$error_type])) {
 			$error_type = $this->known_exception_types[$error_type];
-
 		}
 
 		if (empty($tried_hook)) {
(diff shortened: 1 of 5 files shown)


Sources/Services/ErrorHandlerService.php
```diff
@@ -251,6 +261,12 @@ public function log(string $error_message, string|bool $error_type = 'general',
 		// This prevents us from infinite looping if the hook or call produces an error.
 		$other_error_types = [];
 
+		// Exceptions may get mapped back into our common error types.
+		if (isset($this->known_exception_types[$error_type])) {
+			$error_type = $this->known_exception_types[$error_type];
+
```
live627:
```suggestion
```

live627 (approved):

PR #9187, opened 2026-04-16
[3.0] Improve performance of XmlArray

live627:
One of the main things slowing down the package lists is our XML parser. I have identified several repeated operations and have fixed them to not repeat as much.

The final commit needs a little more testing because I couldn’t test installing a mod due to my built cURL extension not working right.

### DIFF
--- a/Sources/PackageManager/XmlArray.php
+++ b/Sources/PackageManager/XmlArray.php
@@ -157,13 +157,15 @@ final class XmlArray
 
 		// For each element in the path.
 		foreach ($path as $el) {
+			$pos = strpos($el, '[');
+
 			// Deal with sets....
-			if (str_contains($el, '[')) {
-				$lvl = (int) substr($el, strpos($el, '[') + 1);
-				$el = substr($el, 0, strpos($el, '['));
+			if ($pos !== false) {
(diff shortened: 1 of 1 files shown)


Sources/PackageManager/XmlArray.php
```diff
@@ -569,8 +576,12 @@ protected function _parse(string $data): array
...
 	protected function _xml(array $array, ?int $indent): string
 	{
-		$indentation = $indent !== null ? '
-' . str_repeat('	', $indent) : '';
+		static $indentation_cache;
+
+		$indentation_cache ??= $indent !== null
+			? "\n" . str_repeat("\t", $indent)
+			: '';
+		$indentation = $indentation_cache;
```
Sesquipedalian: I don't think this works. $indentation is supposed to be set to a number of tabs equal to the value of $indent, but this change will set $indentation only on the first call to the method and then ignore $indent in any subsequent calls.

Sesquipedalian (changes requested):
Partial review

Sources/PackageManager/XmlArray.php
live627: oh you''re right. I'll revert this one; static cache here probably not worth the effort.

PR #9194, opened 2026-04-29
Fix number formatting issues in message formatter

live627:
The board index stats would truncate after the first thousands separator.


live627:
I also noticed that the PHP build with the intl extension enabled added the thousands separator, while the custom implementation did not.

Sesquipedalian (approved):

Issue #7107, opened 2021-10-05
PHP 8.1 MySQLi error mode


VBGAMER45:
I noticed after looking at the new release for PHPBB they added mysqli_report(MYSQLI_REPORT_OFF);  to handle mysqli errors being outputted to the screen which is the default for PHP 8.1 now...  https://wiki.php.net/rfc/mysqli_default_errmode.

Looks like you can do set before the connection is set.

Sesquipedalian:
We should address this in SMF 2.0 as well.

PR #9201, opened 2026-05-03
don't let the mysql driver throw exceptions when debugging is enabled

live627:


jdarwood007:
Any reason we undid #8816 in this?

live627:
Now my database errors are actually reported as database errors.

Sources/Db/APIs/MySQL.php
```diff
@@ -2513,7 +2513,7 @@ protected function initialize(array $options = []): void
 
 		// Either we aren't in SSI mode, or it failed.
 		if (empty($this->connection)) {
-			if (empty($options)) {
+			if (empty($options)) {
```
live627:
```suggestion
			if (empty($options)) {
```

live627:
- Original issue #7107
- Original fix #7126
- From #8816

> We have since improved our database standards to fix any issues that strict mode would generate.

> This change sets it so when using db_show_debug, we enable the errors/strict mode. This should ensure that during development processes, we do not introduce any new issues that are being silently ignored.

The MySQL strict errors are forwarded to PHP either way.

jdarwood007:
As long as we ensure that mysql errors are being captured.  The focus of my PR was to ensure that we capture errors while in debug mode to ensure we handle them properly before they cause a issue in production.

sbulen:
FYI, debug mode has other problems as well.  My main peeve is it causes lots of token errors, since asking for details eats your token.  

IMO, it's a targeted get in there, look at what you wanna look at, & get immediately out.  Too problematic to just leave on.

jdarwood007:
True for normal operations.  For us developers, our installs typically have debug on all the time.  If it's causing a token issue, we need to look at that.  I haven't experienced a token issue myself.  When I developed tokens for 2.1, I was always in debug mode.  If we have broken tokens since, it has to be how we are reading them out for the debug section below.

sbulen:
In my experience, debug mode has always caused token problems.  For years, since I first used SMF.  I suspect it's one of the reasons we get so many token errors reported on the forum; I think folks turn it on & forget about it.  I've brought this up before on the forum.

I'll log it separately.  

I'm not sure there's anything we can do, actually.  Certain actions cause a rewrite, eating the token.

live627:
I’ve seen the token issue before, but only on repair boards. That makes me think that the query log might have gotten in the way and overflowed the DB column. I have not tested my hypothesis.

PR #9204, opened 2026-05-04
fix variable mismatch in groupsAllowedTo()

live627:


Sources/Subs-Compat.php
```diff
@@ -10513,19 +10513,15 @@ function groupsAllowedTo(string $permission, ?int $board_id = null): array
...
-						$result[$perm]['allowed'][] = $group;
-						break;
-
-					case 0:
-						$result[$perm]['denied'][] = $group;
-						break;
+				if ($value === 1) {
+					$result[$perm]['allowed'][] = $group;
+				} elseif ($value === 0) {
+					$result[$perm]['denied'][] = $group;
```
Sesquipedalian: Why? Sure, `switch` is more verbose, but it only evaluates `$value` once.

Sesquipedalian (changes requested):

Sources/Subs-Compat.php
live627: I'll drop this commit then.

Sesquipedalian (approved):

PR #9206, opened 2026-05-04
build first class callables for hooks

live627:
This is the first of a series of pull requests that will convert string callables to first-class callables. Each branch will focus on a single related unit. For example, this one addresses hooks, while another one focuses on menus, and so on.

Fully backward compatible. I have compiled a script testing all the various callable types that I can think of:

[https://github.com/live627/benchmark-tests/blob/main/callables.php](https://github.com/live627/benchmark-tests/blob/main/callables.php)

It’s quite interesting how the static `class::method` syntax lost the horse race. Any string method will be slower because it is internally converted to a callable, and the first-class syntax can be used directly:

[https://github.com/live627/benchmark-tests/blob/main/callables.txt](https://github.com/live627/benchmark-tests/blob/main/callables.txt)

### DIFF
--- a/Sources/IntegrationHook.php
+++ b/Sources/IntegrationHook.php
@@ -118,7 +118,7 @@ class IntegrationHook
 		foreach ($this->callables as $func_string => $callable) {
 			// Is it valid?
 			if (\is_callable($callable)) {
-				$this->results[$func_string] = \call_user_func_array($callable, $parameters);
+				$this->results[$func_string] = $callable(...$parameters);
 			}
 			// This failed, but we want to do so silently.
 			elseif ($this->ignore_errors) {
--- a/Sources/IntegrationHook.php
+++ b/Sources/IntegrationHook.php
@@ -116,39 +116,7 @@ class IntegrationHook
 
 		// Loop through each callable.
 		foreach ($this->callables as $func_string => $callable) {
-			// Is it valid?
-			if (\is_callable($callable)) {
-				$this->results[$func_string] = $callable(...$parameters);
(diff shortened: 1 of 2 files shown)


Sources/IntegrationHook.php
```diff
@@ -116,39 +116,7 @@ public function execute(array $parameters = []): array
...
-					}
-
-					ErrorHandler::log(Lang::getTxt('hook_fail_call_to', [$func, $path], file: 'Errors'), 'general');
-				}
-				// Assume the file resides on Config::$boarddir somewhere...
-				else {
-					ErrorHandler::log(Lang::getTxt('hook_fail_call_to', [$func_string, Config::$boarddir], file: 'Errors'), 'general');
-				}
-			}
+			$this->results[$func_string] = $callable(...1$parameters);
```
live627:
```suggestion
			$this->results[$func_string] = $callable(...$parameters);
```

tyrsson:
@live627 might wanna try adding static to the arrow function and closure and see if that improves the benchmark. There has been some optimization around that in the more recent PHP versions (although I can't recall the details off the top of my head). Not sure if it will make a difference at all.


```php
$fn = static fn($x) => $x;
$fn = static function ($x) {
    return $x;
};
```

live627:
I think that'll make no difference to the benchmark because none of the closures are created in an object context and therefore cannot bind `$this`.

Relevant article: https://f2r.github.io/en/static-closures

tyrsson:
Will the hooks be binding $this?

PR #9207, opened 2026-05-04
use first class syntax for menu callbacks

live627:
regex pattern

```
__NAMESPACE__\s*\.\s*'\\\\([A-Za-z_][A-Za-z0-9_]*)::([A-Za-z_][A-Za-z0-9_]*)'
```

### DIFF
--- a/Sources/Actions/Admin/ACP.php
+++ b/Sources/Actions/Admin/ACP.php
@@ -93,17 +93,17 @@ class ACP implements ActionInterface, Routable
 			'areas' => [
 				'index' => [
 					'label' => 'admin_center',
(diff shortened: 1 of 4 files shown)


live627:
I have no idea why the linting CI check is complaining on lines that I didn’t change. My ACP still loads as expected.

Sesquipedalian:
> I have no idea why the linting CI check is complaining on lines that I didn’t change. My ACP still loads as expected.
> 
> 

It's pointing to the opening lines of the arrays you changed. The linter doesn't like first class callables in properties. I'm guessing that is because they look to the linter as though they break the rule that properties cannot contain dynamic values.

Issue #9136, opened 2026-03-03
[2.1 & 3.0]: Denying guest access is still costly


sbulen:
### Basic Information

The check whether guest access is disabled is, unfortunately, late in the overall process - AFTER the session has been loaded & AFTER the guest's/user's online status has been logged.

I.e., it doesn't help.  Although it effectively blocks the guest from reading anything, the problematic I/O and high CPU caused by bots/guests remains.

So...  Folks on the forum attempt to handle bot issues by blocking guest access, but that won't work.

What's worse, is that log_online shows info about the request, not the end result.  I.e., in the log it appears they're still browsing when they're not.

We need to find a way to make blocking guests more effective, & to stop all of the I/O associated with guests/bots, BEFORE the session & logging kick in.

https://www.simplemachines.org/community/index.php?topic=593888.0

### Steps to reproduce

1.
2.

### Version/Git revision

2.1.6 & 3.0 Alpha 4

### Database Engine

All

### Logs

```shell

```

jdarwood007:
Going to tag for a future 2.1.8, but I'm not certain if this would make it.  I'm not entirely certain what we can optimize here.

We could, when we determine that they don't have a cookie, not try to load the session.

sbulen:
My suggestion here is that we move the KickGuest check earlier, before the logging, but after the session check.  At that time we'll know if they're a guest or not.  We'll still get the session writes, however, we won't get the log_online writes.  Many folks find these confusing, as they have kicked guests, but the Who's Online still shows them all.  

Not perfect, but it would reduce IO, and make Who's Online make sense...

sbulen:
I think this is related to #7453

PR #9215, opened 2026-05-06
[2.1] Don't log guests if disallowed

### LINKED ISSUE #7453: Who's Online doesn't always check whether bots actually have permission to log in
Source: https://www.simplemachines.org/community/index.php?topic=582176.0

Image 5-6-22 at 9 22 PM

As reported by Arantor:
> As per the screenshot, Google did get an error trying to view registration (because you can see the little red (!) icon there), but I don't see it getting a little (!) icon for the others...
I just tried it myself and correctly got stopped so maybe the logging code doesn't work so well there? 

Maybe it's minor, but it's still an inconsistency.

sbulen:
Partial for #9136 
Fixes #7453

Do any guest logging AFTER checking if guests should be kicked...

This will relieve I/O during bot attacks - no need to log folks who aren't doing anything.

Also, the logging was confusing - it made it look like they were doing something in Who's Online.

### DIFF
--- a/index.php
+++ b/index.php
@@ -240,6 +240,30 @@ function smf_main()
 	if (!empty($topic) && empty($board_info['cur_topic_approved']) && !allowedTo('approve_posts') && ($user_info['id'] != $board_info['cur_topic_starter'] || $user_info['is_guest']))
 		fatal_lang_error('not_a_topic', false);
 
+	// Do some logging, unless this is an attachment, avatar, toggle of editor buttons, theme option, XML feed, popup, etc.
+	$no_stat_actions = array(
+		'about:unknown' => true,
+		'clock' => true,
+		'dlattach' => true,
+		'findmember' => true,
+		'helpadmin' => true,
+		'jsoption' => true,
+		'likes' => true,
+		'modifycat' => true,
+		'pm' => array('sa' => array('popup')),
+		'profile' => array('area' => array('popup', 'alerts_popup', 'download', 'dlattach')),
+		'requestmembers' => true,
+		'smstats' => true,
+		'suggest' => true,
+		'uploadAttach' => true,
+		'verificationcode' => true,
+		'viewquery' => true,
+		'viewsmfile' => true,
+		'xmlhttp' => true,
+		'.xml' => true,
+	);
+	call_integration_hook('integrate_pre_log_stats', array(&$no_stat_actions));
+
 	// Make sure that our scheduled tasks have been running as intended
 	check_cron();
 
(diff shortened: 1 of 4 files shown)


sbulen:
Note that the 'User has been taken to the login page' functionality didn't work in 2.1...  Worked OK in 3.0.

It may have been less confusing to users if it were displayed properly in 2.1...  Maybe not...  In 3.0 it said, "User is viewing board 1.0" (or whatever action) with an icon, which, when the icon was clicked, provided a popup that said "User has been taken to the login page".  Contradictory info, and the real state was dependent on folks thinking of clicking on that tiny icon.

index.php
```diff
@@ -240,42 +240,6 @@ function smf_main()
...
-		'smstats' => true,
-		'suggest' => true,
-		'uploadAttach' => true,
-		'verificationcode' => true,
-		'viewquery' => true,
-		'viewsmfile' => true,
-		'xmlhttp' => true,
-		'.xml' => true,
-	);
-	call_integration_hook('integrate_pre_log_stats', array(&$no_stat_actions));
```
live627, line 265: We should probably leave this array here just in case some site admin decided to modify it, unlikely as it is.

live627:
This looks like it might stop logging for valid topic and board requests.

sbulen:
> This looks like it might stop logging for valid topic and board requests.

All this did was move the logging to AFTER the kickGuest test.   What it does today is log invalid requests.  This is very confusing, comes up in the support boards regularly.  Folks deny guest access, and during a bot attack, Who's Online shows hundreds of guests/bots browsing the forum & doing various actions.  When they aren't...

So it's the other way around - today it logs invalid requests, and this change stops that.  The log will make more sense.

index.php
sbulen: Extremely unlikely that's been changed, IMO.

And to me it makes sense to keep the logging logic together, it's more readable.

In fact, it could (should?) be a single function call.  An argument could be made to put that logic in Logging.php...

live627:
That function has a very weird design where it returns a function name. The problem is that the complex branching logic includes topics.

3.0 thankfully unravels it somewhat...

index.php
jdarwood007: Any harm in leaving the array definition here and just moving the logic?

This is for 2.1 and we have to consider that it will add to the patch size to move the array.

jdarwood007:
I think this looks good after we resolve the conversation about moving that array definition.

index.php
sbulen: Mainly readability.  I like having the related logic together, so you know what it's doing without having to scroll all over the place. 

Edits like that are like running your code thru a paper shredder, and taping it back together out of sequence but somehow still functional.  Maybe it still works fine, but it makes the code incomprehensible.

IMO, readability is important.

But that's two team members now who want to split it out...  I'll make the change & retest...

jdarwood007: I understand the logical order.  I was mostly being considerate to what our patching operations will look like for those upgrading from 2.1.7 to 2.1.8.  A big block of code will have to be removed and re-added, essentially.

If this were for 3.0 that is not been released yet, no issue with moving blocks around.

sbulen:
OK, array & hook both restored to where they were.

I think this one is ready now.

PR #9216, opened 2026-05-06
[3.0] Don't log guests if disallowed

sbulen:
Partial for #9136 

Do any guest logging AFTER checking if guests should be kicked...

This will relieve I/O during bot attacks - no need to log folks who aren't doing anything.

Also, the logging was confusing - it made it look like they were doing something in Who's Online.


sbulen:
I think this one is ready as well.

Due to the way it's structured, the array wasn't moved.

Issue #9179, opened 2026-04-05
[2.1]: Full PHP 8.5 compatibility audit and fixes


gitgost-anonymous:
### Basic Information

The `release-2.1` branch has multiple PHP 8.5 incompatibilities that need to be addressed. PHP 8.5 promotes several deprecations from PHP 8.1–8.4 to errors/removals, and the current codebase — originally targeting PHP 7.0 — is affected in numerous places. The CI workflow (`.github/workflows/php.yml`) only lints syntax up to PHP 8.4 and does not run any runtime tests, so these issues are not currently caught.

This issue consolidates **all** known and discoverable PHP 8.5 incompatibilities into a single tracking issue with exact code references.

---

### 1. `trigger_error()` with `E_USER_ERROR` (deprecated PHP 8.4, error in 8.5)

PHP 8.4 deprecated passing `E_USER_ERROR` to `trigger_error()`. In PHP 8.5, this will throw a `ValueError`. SMF's database error backtrace function passes `E_USER_ERROR` through from callers and also directly invokes `trigger_error()` with it.

**Related issue:** #9141 (open, in 2.1.8 milestone — but only covers the DB layer)

#### 1a. `Sources/Subs-Db-mysql.php` — `smf_db_error_backtrace()`

https://github.com/SimpleMachines/SMF/blob/7438c2b04807bca3b3160801a3d40fd2f0f0ca7b/Sources/Subs-Db-mysql.php#L876-L916

The function accepts `$error_type` and passes it directly to `trigger_error()` on lines 913–915. Every caller that passes `E_USER_ERROR` triggers the deprecation:

- **Line 195** — `smf_db_error_backtrace('Invalid value inserted...', '', E_USER_ERROR, ...)`
- **Line 209** — Integer type check failure
- **Line 222** — Empty array_int
- **Line 227** — Non-integer in array_int
- **Line 235** — Non-array for array_int
- **Line 243** — Empty array_string
- **Line 251** — Non-array for array_string
- **Line 258** — Invalid date
- **Line 265** — Invalid time
- **Line 274** ��� Invalid datetime
- **Line 279** — Invalid float
- **Line 296** — Invalid IPv4/IPv6
- **Line 304** — Empty array_inet
- **Line 311** — Invalid inet value
- **Line 318** — Non-array for array_inet

[...]

PR #9219, opened 2026-05-07
[2.1] uninitialized permissions should not give error spam

live627:
- Install any mod that uses inline permissions
- I grabbed Message Bookmarks
- Run PHP 8.5
```
8192: Using null as an array offset is deprecated, use an empty string instead
```


live627:
- #9179

PR #9244, opened 2026-05-29
[2.1] Detects multiline func signatures in get_defined_functions_in_file()

### LINKED ISSUE #9243: [2.1]: get_defined_functions_in_file() fails to detect multiline method signatures, marking valid integration hooks as missing
### Basic Information

Some IDEs (e.g., PHPStorm) automatically suggest using a multiline format for long parameters. Therefore, in code you can often find entries like this:

```php
public function modifyBoard(
	int $id,
	array $boardOptions,
	array &$boardUpdates,
	array &$boardUpdateParameters
): void
{
	// Code
}
```

However, currently SMF only supports the single-line format for such parameters:

```php
public function modifyBoard(int $id, array $boardOptions, array &$boardUpdates, array &$boardUpdateParameters): void
{
	// Code
}
```

As a result, some hooks are marked as missing, even though they actually exist.

### Steps to reproduce

1. Install Optimus 3.0
2. Check the Integration Hooks table in the forum's settings

### Expected result

_No response_

### Actual result

_No response_

### Version/Git revision

2.1.7

### Database Engine

All

### Database Version

_No response_

### PHP Version

_No response_

### Logs

```shell

```

### Additional Information

Possible fix: https://www.simplemachines.org/community/index.php?msg=4202555

Sesquipedalian:
Fixes #9243

### DIFF
--- a/Sources/ManageMaintenance.php
+++ b/Sources/ManageMaintenance.php
@@ -2239,7 +2239,7 @@ function get_defined_functions_in_file(string $file): array
 {
 	$source = file_get_contents($file);
 	// token_get_all() is too slow so use a nice little regex instead.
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
Please confirm that this works in your tests, @dragomano

dragomano:
Yes, everything works.

PR #9249, opened 2026-05-31
Further improves getDefinedFunctionsInFile()

live627:
Support functions with union types and classes that   extend fully qualified types

Can now parse nearly the entire SMF codebase as seen in my very long result set https://github.com/live627/benchmark-tests/blob/main/smf/fns.txt

My comparison script https://github.com/live627/benchmark-tests/blob/main/smf/fns.php

### DIFF
--- a/Sources/Actions/Admin/Maintenance.php
+++ b/Sources/Actions/Admin/Maintenance.php
@@ -2300,12 +2300,20 @@ class Maintenance implements ActionInterface
 	protected static function getDefinedFunctionsInFile(string $file): array
 	{
 		$source = file_get_contents($file);
-		// token_get_all() is too slow so use a nice little regex instead.
-		preg_match_all('/\bnamespace\s++((?P>label)(?:\\\(?P>label))*+)\s*+;|\bclass\s++((?P>label))[\w\s]*+{|\bfunction\s++((?P>label))\s*+\(.*?\)[:\|\w\s]*+{(?(DEFINE)(?<label>[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*+))/is', $source, $matches, PREG_SET_ORDER);
 
+		if (!str_contains($source, 'function')) {
+			return [];
+		}
+
+		// Remove multiline comments so regex does not
+		// match fake functions/classes inside them.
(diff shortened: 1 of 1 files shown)


Sources/Actions/Admin/Maintenance.php
```diff
@@ -2300,13 +2300,21 @@ protected static function parseIntegrationHook(string $hook, string $rawData): a
...
-		// token_get_all() is too slow so use a nice little regex instead.
-		preg_match_all('/\bnamespace\s++((?P>label)(?:\\\(?P>label))*+)\s*+;|\bclass\s++((?P>label))[\w\s]*+{|\bfunction\s++((?P>label))\s*+\(.*?\)[:\|\w\s]*+{(?(DEFINE)(?<label>[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*+))/is', $source, $matches, PREG_SET_ORDER);
 
+		if (!str_contains($source, 'function')) {
+			return [];
+		}
+
+		// Remove multiline comments so regex does not
+		// match fake functions/classes inside them.
+		$source = preg_replace('~//[^\h]+|/\*.*?\*/~s', '', $source);
```
Sesquipedalian, line 2310:
```suggestion
		// Remove multiline comments so regex does not
		// match fake functions/classes inside them.
		$source = Config::stripPhpComments($source);
```

This would produce more thorough and reliable results.

Can I get you to submit a follow-up PR making this change, @live627?

live627: I hadn't noticed that function before now. I'm testing it now. It's quite a bit slower: the one with the token takes 0.43s, while the custom algorithm takes 0.38s. The one in this branch takes 0.08s.

That custom implementation also missed a function.

(new regex = mine, old regex = your suggested change without the tokenizer)
```
Files:        1599
Iterations:   1

New regex total: 79.910 ms
Old regex total: 378.003 ms

New regex avg:   79.910 ms
Old regex avg:   378.003 ms

Speedup:         4.73x (new regex faster)

New regex count: 4420
Old regex count: 4419

Differences detected:

../../smf2.1/sources\Subs-Compat.php
  New regex only:
    + package_crypt
```

PR #9258, opened 2026-06-06
Use Tokens::isAnyTokenKindsFound() in SectionComments::isCandidate()

live627:
This change replaces manual token iteration in `SectionComments::isCandidate()` with PHP-CS-Fixer's built-in `Tokens::isAnyTokenKindsFound()` method.

The previous implementation scanned every token until a class-like structure was found. The new implementation delegates this check to the tokenizer's optimized lookup mechanism, which is just a simple hash lookup. `isCandidate( )` is called `34694` times.

### DIFF
--- a/.github/phpcs/SectionComments.php
+++ b/.github/phpcs/SectionComments.php
@@ -232,14 +232,7 @@ final class SectionComments extends AbstractFixer
 		$tokens->clearEmptyTokens();
 
 		// Does this file contain an enumeration?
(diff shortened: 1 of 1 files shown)


live627:
also adding a commit to move regex build to the constructor, which is called `282` times, whereas `applyFix()` is called `21860` times. Should reduce repetitive work.

Sesquipedalian:
Merged via the command line in 4ace4cd

PR #9265, opened 2026-06-11
Further improves SectionComments

live627:
* Use a more forgiving regex to detect slightly malformed section comments. Idea for further improvement: Use `Utils::buildRegex()` if available, because each comment is stripped to the bare words before building the final regex.
* Avoids removing and reinserting section comments that are already correct and in the proper location. This is explicitly checked for in the vendor-provided test suite. 
* Defers removal of obsolete section comments until after analysis, reducing token stream mutations during processing. This improves fixer idempotency by avoiding unnecessary token modifications when no changes are required.
* Correctly preserves valid section comments while removing duplicates and misplaced comments.

My tests https://github.com/live627/php-cs-fixer-custom-fixers/blob/main/tests/SectionCommentsFixerTest.php

### DIFF
--- a/.github/phpcs/SectionComments.php
+++ b/.github/phpcs/SectionComments.php
@@ -99,10 +99,10 @@ final class SectionComments extends AbstractFixer
 		];
 
 		foreach ($this->comments as $type => $string) {
-			$regexes[$type] = preg_replace('/\s+/', '\s+', preg_quote($string, '/'));
+			$words[] = trim($string, "/* \n\t");
 		}
 
-		$this->comment_regex = implode('|', $regexes);
+		$this->comment_regex = '/^\/[*\s]+(?:' . implode('|', $words) . ')[*\s]+\/$/';
 	}
 
 	public function getName(): string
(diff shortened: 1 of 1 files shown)


live627:
```
Case                                       Tokens
-------------------------------------------------
all sections with phpdocs and enum            401
all sections and enum                         341
section comments missing a space              361
misplaced and duplicate section comments       93
completely wrong section comments              34
```

.github/phpcs/SectionComments.php
```diff
@@ -254,57 +239,58 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens): void
...
-					'T_PUBLIC',
-					'T_PROTECTED',
-					'T_PRIVATE',
-					$is_enum && !$exists['case'] ? 'T_CASE' : NAN,
+				$id,
+				$in === [] ? [
+					T_PUBLIC,
+					T_PROTECTED,
+					T_PRIVATE,
+					$is_enum && !$exists['case'] ? T_CASE : -1,
```
Sesquipedalian:
```suggestion
					$is_enum && !$exists['case'] ? T_CASE : NAN,
```

NAN was an intentional choice here, because it is never equal to anything (not even itself).

.github/phpcs/SectionComments.php
```diff
@@ -99,10 +100,10 @@ public function __construct()
 		];
 
 		foreach ($this->comments as $type => $string) {
-			$regexes[$type] = preg_replace('/\s+/', '\s+', preg_quote($string, '/'));
+			$words[] = trim($string, "/* \n\t");
```
Sesquipedalian:
```suggestion
			$words[] = preg_replace('/\s+/', '\s+', preg_quote(trim($string, "/* \n\t"), '/'));
```

This needs to keep the part where we replace literal whitespace with `\s+`. Otherwise this supposedly more forgiving regex will fail if, for example, the comment's indentation changed.

Sesquipedalian (changes requested):
Partial review

.github/phpcs/SectionComments.php
```diff
@@ -207,156 +208,194 @@ public function isCandidate(Tokens $tokens): bool
...
-							$tokens[$insert_at - 1]->isGivenKind([T_FINAL, T_READONLY, T_ABSTRACT])
-							|| $tokens[$insert_at - 1]->isWhitespace()
-							|| $tokens[$insert_at - 1]->isComment()
-						)
-					) {
-						$insert_at--;
-					}
+				$elements[] = [
+					'type' => $insert_type,
+					'start' => array_first($in),
```
Sesquipedalian:
```suggestion
					'start' => $in[array_key_first($in)];,
```

Sesquipedalian (changes requested):

Sesquipedalian (approved):

.github/phpcs/SectionComments.php
```diff
@@ -207,156 +208,194 @@ public function isCandidate(Tokens $tokens): bool
...
-							$tokens[$insert_at - 1]->isGivenKind([T_FINAL, T_READONLY, T_ABSTRACT])
-							|| $tokens[$insert_at - 1]->isWhitespace()
-							|| $tokens[$insert_at - 1]->isComment()
-						)
-					) {
-						$insert_at--;
-					}
+				$elements[] = [
+					'type' => $insert_type,
+					'start' => $in[array_key_first($in)];,
```
Sesquipedalian:
```suggestion
					'start' => $in[array_key_first($in)],
```

Sesquipedalian (changes requested):

Sesquipedalian (approved):

Issue #9259, opened 2026-06-07
[2.1 & 3.0]: Upgrader serialize-to-json logic issues


sbulen:
### Basic Information

I've been helping a forum member with a very large & problematic upgrade, 3.3M+ posts.  This is one of the issues faced.  This is a 2.0 => 2.1 upgrade.  Forum is MariaDB, MyISAM, latin1.  

The upgrade crashes repeatedly at the json conversion step, before the utf8 conversion.  The fatal error is caused by this query:
<img width="1441" height="68" alt="Image" src="https://github.com/user-attachments/assets/c3dc6081-6016-42d9-9fc8-d7910bf8403f" />

Note that it is trying to decrement the value of the 'memberlist_cache' setting via updateSettings.  The DB query is failing because it's a serialized string, not a numeric value...  Why did it do that?  The json_encode of the unserialized string comes back false.  The call to updateSettings with a new value of false tells upgradeSettings to set the string to 'value - 1', which fails.

**_EDIT:  After further tests & diagnosing the issue, I don't think this has anything to do with MariaDB vs MySQL, nor CLI vs browser.  It has nothing to do with how things are escaped._**

The root, I believe, is more basic than that...  json_encode ONLY works on utf8...  And we're not on utf8 yet, at this stage of the upgrade...  When you try to json_encode a string with a singe-byte, latin1 umlaut in it, it fails & returns false.

### Steps to reproduce

1.
2.

### Version/Git revision

2.1.7 & 3.0 Alpha 4

### Database Engine

All

### Database Version

MySQL 8.4.4 & MariaDB 11.5.2

### PHP Version

8.4.5

### Logs

```shell

```

### Additional Information

This is followup to the forum threads regarding failed upgrades & restores:
https://www.simplemachines.org/community/index.php?topic=594485.0
https://www.simplemachines.org/community/index.php?topic=594487.0

sbulen:
Peeling the onion...

At start, this upgrade took 18 hours & would not complete.  It timed out repeatedly, & crashed during the serialize_to_json step.  The serialize_to_json step was partially completed, and the parts that were completed were mostly incorrect.

Steps taken:
 - Convert to InnoDB before upgrading.  This reduced the time of execution from 18 hours to < 5 hours.  The reason is that MyISAM tables must be rebuilt during most ALTER TABLE operations.  And the upgrade has A LOT of ALTER TABLE operations on the smf_messages table.  Don't do that.
[...]

PR #9271, opened 2026-06-17
[2.1] Upgrader - Fix issues with json conversion

sbulen:
Partial for #9259

The json conversion step had some issues.  
 - Due to a typo, safe_unserialize() string fixes were never applied.
 - json_encode can ONLY work on UTF8.  Nothing else.  Anytime a non-utf8 character was passed, the json_encode would fail & return false.
 - The result was not checked, & "false" was passed to updates, which, for example, might tell updateSettings() to decrement the value of the setting.  MariaDB would crash rather than comply; MySQL would change affected settings to -1, e.g., memberlist_cache.  
 - The json conversion occurs before the UTF8 conversion...  So this would happen even in latin1 databases (umlauts & accents, etc.).

The net result is that many strings were either not converted or blanked out.  Settings might be affected, e.g,. memberlist_cache.  Most of the issues were with various logs, e.g., log_actions.  qanda.  Places where user-entered text ended up in a serialized string.  

I use a simple "cheezy 2.1 encoding" technique to preserve the non-utf8 strings across json_encode.  (Trademark pending.)  Got a better idea, let me know...  But it works.  The UTF8 conversion, which comes later, works as desired & the content is properly migrated to UTF8.

This one is ready to go.  If approved, I'll work on the 3.0 version.

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -3629,7 +3629,7 @@ function serialize_to_json()
 							echo "\n - Failed to unserialize the '" . $var . "' setting. Skipping.";
 						elseif ($temp !== false)
 						{
-							$new_settings[$var] = json_encode($temp);
+							$new_settings[$var] = json_encode($temp, JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE);
 							if ($new_settings[$var] === false)
 								$new_settings[$var] = '';
 							else
@@ -3681,7 +3681,7 @@ function serialize_to_json()
 
 						if ($temp !== false)
 						{
-							$row['value'] = json_encode($temp);
+							$row['value'] = json_encode($temp, JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE);
 							if ($row['value'] === false)
 								$row['value'] = '';
 							else
@@ -3771,7 +3771,7 @@ function serialize_to_json()
 										echo "\nFailed to unserialize " . $row[$col] . ". Setting to empty value.\n";
 								}
 
-								$row[$col] = json_encode($temp);
+								$row[$col] = json_encode($temp, JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE);
 								if ($row[$col] === false)
 									$row[$col] = '';
 								else
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -3510,22 +3510,27 @@ function upgrade_unserialize($string)
 		$data = @safe_unserialize($string);
 
 		// The serialized data is broken.
-		if ($data === false)
+		// OR... Has strings that are not utf8.
+		if (($data === false) || (mb_check_encoding($string, 'UTF-8') === false))
 		{
 			// This bit fixes incorrect string lengths, which can happen if the character encoding was changed (e.g. conversion to UTF-8)
 			$new_string = preg_replace_callback(
 				'~\bs:(\d+):"(.*?)";(?=$|[bidsaO]:|[{}}]|N;)~s',
 				function ($matches)
 				{
+					// If not utf8, use cheezy-21-encoding, because json_encode ONLY works on utf8
+					// Will decode this after json_encode; utf8 conversion can then proceed properly on the non-utf8 data
+					if (mb_check_encoding($matches[2], 'UTF-8') === false)
+					{
+						$matches[2] = 'czy21enc:' . bin2hex($matches[2]);
+					}
 					return 's:' . strlen($matches[2]) . ':"' . $matches[2] . '";';
 				},
 				$string
 			);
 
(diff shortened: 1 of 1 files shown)


sbulen:
Examples of the types of issues you might see...

### _**In smf_settings:**_
nax_mysql_settings2

### _**In smf_log_actions:**_
nax_mysql_log_actions

sbulen:
### _**If json_encode *DID* find proper unicode, its default was to escape it, which was bad for SMF:**_
json_unicode_goof

sbulen:
Found an issue, need to do some retests...

sbulen:
The issue I have found is specific to older 2.0 DBs that were not converted to utf8.

Any forum that has $db_character_set properly specified is fine.  

Note that 2.0 didn't set $db_character_set upon install; 2.0 would only set it upon utf8 conversion.  2.0 supported other charsets, e.g., latin5 - if a forum did so, my understanding is that it was up to the admin to set $db_character_set properly and set the db defaults properly before install.  _(I can't find this doc'd anywhere, but it seems to work...)_  So I am pretty sure most non-latin1 forums are OK, because they'd have $db_character_set set (or they wouldn't work...).  So I _believe_ the issue is effectively confined to (1) latin1, (2) no $db_character_set, (3) forum DB upgraded & somewhat current.  

Without $db_character_set, SMF never executes a SET NAMES.  Without SET NAMES, character_set_results, character_set_connection, character_set_client all default to character_set_server, which is typically utf8mb4 these days.  So when the upgrader is communicating with the DB, there is unwanted translation occurring.

Seems like it'd be an obvious fix, just set $db_character_set internally under the proper conditions.  But this isn't working for some reason, I haven't figured out why yet...  Also note that MariaDB seems to be ignoring SMF's specified collation.  Things aren't ending up utf8mb3_general_ci at the end of the upgrade (it's all utf8mb3_uca1400_ai_ci)...  So there are other things going on here.  

Note things are far better with this PR - no more crashes, serialized string fix logic works, fewer blank/garbled strings.  Just not 100%.  

Still testing.

sbulen:
Hmmm...  Remaining issues are specifically with log_actions, which gets multiple rounds of data fixes during the database changes steps, prior to json conversion...

sbulen:
Got it...

If you check the box 'Migrate to a new Settings file', with no value for $db_character_set, then in step 2, Upgrade Options, a new file is written with $db_character_set = 'utf8'.  

All subsequent steps load the new Settings.php and issue a SET NAMES to utf8.  But the DB is not utf8 yet, so this causes issues.   There are a few updates to log_actions during the DB updates.  This is where the issue lies - the strings don't even make it to the json conversion.  

To confirm, I just ran a conversion _without_ the 'Migrate to a new Settings file' box checked & it ran just fine, the data is all preserved.  

Should have a fix tomorrow.

jdarwood007:
There has to be a way to verify the data is in the proper charset we expect prior to trying to deserialize it.

sbulen:
Yep.  Testing a solution now.  So far, so good.

The problem isn't due to the serialize/deserialize part.  It's the actual DB query & subsequent update...  With SET NAMES to utf8, it is having issues with single-byte (non-utf8) latin1 extended chars, eg umlauts & these guys: 'àáäå'.  The problematic DB I've gotten my hands on has many such strings.  (MySQL is happy mixing & matching... MariaDB not so much...)

Since it's not related to the json conversion, I'll open a separate PR when it's ready, tonight or tomorrow am.  The restores & upgrades for each test take a while...

jdarwood007:
I was hoping we could walk the array to check for sane characters.

sbulen:
In these examples, they're all good characters.  There's nothing wrong with the json strings, or even the string lengths.  There was never a utf8 conversion; the string lengths are correct.

The problem is that we issued SET NAMES to utf8, but the data is actually encoded latin1 and the DB is latin1.  We gotta set the $db_character_set and do the correct SET NAMES.

jdarwood007:
Gotcha.  Hopefully your solution doesn't rely on the set names then.  We are bound to break this again in the future as we expect utf8.  In fact the serialization to json conversion should occur as a migration step in the future rather than a separate process to ensure its handled properly and we don't attempt it again after 3.0.

sbulen:
#9278 was submitted for the $db_character_set/SET NAMES issue.   One correction to my notes above: I redid some tests, and after ensuring $db_character_set was NOT set in Settings.php, found that $MySQL was affected as well.  The issue was not restricted to just MariaDB.  

In all my testing, the json conversion looks good now, so I'm flagging this as ready.

Sesquipedalian (approved):

Sesquipedalian:
Looks great, @sbulen! I look forward to the 3.0 version as well.

sbulen:
Let me know if you want a 3.0 version of this.

Sesquipedalian:
Yes

PR #9272, opened 2026-06-17
[2.1] Upgrader - Performance

sbulen:
See #9259 for other issues with large DB upgrades

The upgrader has several extremely expensive database operations, all to the messages table:
 - Adding likes column
 - Adding modified_reason column
 - Renaming poster_ip to poster_ip_old
 - Adding column poster_ip
 - Dropping column poster_ip_old

For ***each*** of the above operations, under the covers, ALTER TABLE makes a copy of the table then rebuilds it from scratch.  
 - Everything gets rebuilt, even fulltext indexes when present
 - MySQL implicit table rebuilds can be VERY costly; I've seen them take as long as 10 hours ***each*** on a table with 3.3M records

It doesn't makes sense for it to take many hours to do a simple operation, even on 3.3M messages...  ALTER TABLE clearly has issues with these implicit table rebuilds; it is much faster to rebuild the table explicitly yourself.

MariaDB handles this much better than MySQL, I believe due to a better implementation of ALGORITHM=INPLACE & ALGORITHM=INSTANT.  MySQL has more constraints on these timesavers.  HOWEVER...  If you restructure the activity a bit, to account for such constraints (e.g., only dropping columns at the end of the column list) MySQL is almost as quick as MariaDB.

This PR:
 - Reduces message table rebuilds by making an empty copy of messages, then doing most DB changes to the empty copy.
 - The table is then repopulated explicitly, which is far faster than ALTER TABLE's implicit copy.
 - Removes the fulltext index prior to any DB activity.  This had been part of the utf8 conversion; this PR moves it prior to the DB changes.  This cleans up many warnings & errors in the MySQL/MariaDB logs when FTS is present.  _More importantly, the upgrader isn't rebuilding the fulltext index over & over & over..._  
 - Where necessary (because the operation requires populated data), moved column to be dropped to the end of the column list.

The IPv6 & UTF8 conversions remain time-consuming.  No way around that.

Still, the time savings are significant.  My "before" executions on MySQL were in the ~60 hour range, and varied on a number of factors, e.g., InnoDB vs MyISAM & environment/debug settings.  MariaDB "before" executions varied from 5-18 hours.  Tests were 2.0 => 2.1 upgrades, DB has 3.3M rows, no attachments.  

With this change, MySQL now takes ~1 hour, & MariaDB takes ~45 mins.

If necessary, we could do the same to members.  I've never seen members updates go near as long though.

I'm still doing tests on this.

If approved, I'll work on the 3.0 version.

### DIFF
--- a/other/upgrade.php
+++ b/other/upgrade.php
@@ -3052,51 +3052,6 @@ function ConvertUtf8()
 		list($upcontext['database_charset']) = explode('_', $column_info['Collation']);
 		$upcontext['database_charset'] = in_array($upcontext['database_charset'], $charsets) ? array_search($upcontext['database_charset'], $charsets) : $upcontext['database_charset'];
 
-		// Detect whether a fulltext index is set.
-		$request = $smcFunc['db_query']('', '
-			SHOW INDEX
-			FROM {db_prefix}messages',
-			array(
-			)
-		);
-
-		$upcontext['dropping_index'] = false;
-
-		// If there's a fulltext index, we need to drop it first...
-		if ($request !== false || $smcFunc['db_num_rows']($request) != 0)
-		{
-			while ($row = $smcFunc['db_fetch_assoc']($request))
-				if ($row['Column_name'] == 'body' && (isset($row['Index_type']) && $row['Index_type'] == 'FULLTEXT' || isset($row['Comment']) && $row['Comment'] == 'FULLTEXT'))
-					$upcontext['fulltext_index'][] = $row['Key_name'];
-			$smcFunc['db_free_result']($request);
-
-			if (isset($upcontext['fulltext_index']))
-				$upcontext['fulltext_index'] = array_unique($upcontext['fulltext_index']);
-		}
-
-		// Drop it and make a note...
-		if (!empty($upcontext['fulltext_index']))
-		{
-			$upcontext['dropping_index'] = true;
-
-			$smcFunc['db_query']('', '
-				ALTER TABLE {db_prefix}messages
-				DROP INDEX ' . implode(',
-				DROP INDEX ', $upcontext['fulltext_index']),
-				array(
-					'db_error_skip' => true,
-				)
-			);
-
-			// Update the settings table
-			$smcFunc['db_insert']('replace',
-				'{db_prefix}settings',
-				array('variable' => 'string', 'value' => 'string'),
-				array('db_search_index', ''),
-				array('variable')
-			);
-		}
-
 		// Figure out what charset we should be converting from...
 		$lang_charsets = array(
 			'arabic' => 'windows-1256',
@@ -3470,7 +3425,7 @@ function ConvertUtf8()
 		}
 		$smcFunc['db_free_result']($request);
 
-		if ($upcontext['dropping_index'] && $command_line)
+		if (!empty($_SESSION['dropping_index']) && $command_line)
 		{
 			echo "\n" . '', $txt['upgrade_fulltext_error'], '';
 			flush();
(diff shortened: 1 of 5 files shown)


sbulen:
Without this PR...

Here is an example of an ALTER TABLE, adding the "likes" column.  The PROCESSLIST shows that the current state is it is making a tmp copy of the table.  

MySQL 8.4.4, an InnoDB table, 3.3M messages, this single ALTER TABLE command is still ongoing at ~7 hours.  

This is still early in the upgrade process - there are at least another 4 similar operations to go.  I.e., this particular upgrade will take AT LEAST another 28 hours, simply on things like adding & removing columns.  Plus it still has the IP conversion & the UTF8 conversion to go. 

I believe part of the problem here is the default value; when completed, the default value needs to be on each record, so it cannot take any shortcuts, i.e., there is no opportunity for an ALGORITHM=INSTANT to help out (for MySQL anyway).  It MUST update every row in the table...

nax_restructure_likes_tmp_copy

sbulen:
OTOH, with this PR...

This upgrade is only 16 minutes in and the DB updates are almost complete.  In fact, it's already 40% thru the explicit rebuild of messages.  It still has the IPv6, JSON & UTF8 conversion to go, though.  Still, it will be done in ~1hr total.

Same setup:  MySQL 8.4.4, InnoDB, 3.3M messages.

nax_restructure_2

sbulen:
Some before & after comparisons...  I've always understood that MyISAM is faster than InnoDB.  However, in my initial upgrader tests, InnoDB was faster, due to the table copy efficiencies.  So it depends on the activity.

MariaDB 11.5.2.  MySQL 8.4.4.
Test DB: 3.3M rows, fulltext index on message body.  No attachments.
The bottleneck is disk I/O, which is pegged to 100% during upgrades.  Old school HDD.

**_Before_** - 2.0.19 => 2.1.7, with #9271 only:
 - MyISAM - MySQL: 60 hours (estimated, I killed it...)
 - MyISAM - MariaDB: 18 hours
 - InnoDB - MySQL: 60 hours
 - InnoDB - MariaDB: 5 hours

Note that without the json fix #9271, Maria DB cannot complete with this dataset.

**_After_** - 2.0.19 => 2.1.7, with #9270, #9271, #9272:
 - MyISAM - MySQL: 20 mins
 - MyISAM - MariaDB: 15 mins
 - InnoDB - MySQL: 65 mins
 - InnoDB - MariaDB: 45 mins

It's pretty shocking how quick MyISAM can be if you sidestep the table restructures w/fulltext.

jdarwood007:
I like the changes, but I am worried about these in the stable release.  @Sesquipedalian, what are your thoughts on this?

live627:
This doesn't affect the patch though.

jdarwood007:
i'm worried we are changing the upgrader in a way that we can't be certain we didn't introduce a breaking upgrade bug that can be caught in the limited testing that occurs during a patch.  Not the actual patch file itself.

sbulen:
I've made a few changes, mostly to improve look & feel of the upgrade, to make it smoother.  Especially with large forums, it was too choppy, pausing & restarting.  

Changes:
 - Moved the fulltext index drop to the database changes.  The problem with it being in the upgrade options is that it took a few minutes to execute, and that left the upgrade options screen hanging.  Felt like the upgrade froze.
 - Smaller bites for a couple operations, including ipv6 conversion & message table rebuild.  That way nextSubstep updates are reflected more often, & it's more visible that it's moving along.  
 - Found some bugs in the .sql layout, that affected things like the progress count & status display.  The count being wrong sometimes affected the CONTINUE button being enabled/disabled properly.  

Minor tweaks only.  Most of this is unnoticeable on smaller upgrades, but the start/stop choppiness was pronounced on larger upgrades.

Still testing.  Nothing found so far other than the above.

sbulen:
I wrote a little QnD utility to populate a 2.0 DB with a few million rows populated with heaps of latin1 (one-byte, non-utf8) data:
https://github.com/sbulen/sjrbTools/blob/master/smf_populate_db_20.php

Ugly, but effective...

This will help folks trying to reproduce issues such as those discussed in #9259, #9270, #9271, #9272 & #9278 including:
 - various issues caused by sheer volume & execution time
 - charset issues
 - json encoding issues

If you would like to experience horror & woe, run that tool against a 2.0 latin1 DB, then run the 2.1.7 upgrader from the SMF site against it...

other/upgrade.php
```diff
@@ -3476,7 +3431,7 @@ function ConvertUtf8()
 		require_once($sourcedir . '/Subs-Admin.php');
 		updateSettingsFile(array('db_character_set' => 'utf8'));
 
-		if ($upcontext['dropping_index'] && $command_line)
+		if (!empty($_SESSION['dropping_index']) && $command_line)
```
jdarwood007, line 3434: Don't use session data.  Incase the upgrader gets paused long enough a session is lost.

PR #9283, opened 2026-07-08
Replace deprecated backtick operator with shell_exec() in load averag…

NukeWorker:
…e check

The backtick operator is deprecated as of PHP 8.5 (E_DEPRECATED). Sources/Load.php uses @`uptime` to read the system load average, which emits a deprecation notice on 8.5+. Swapped to the equivalent @shell_exec('uptime') — same return value, no behavior change.


live627:
Thanks for this! Can you modify the commit message to add a `signed-off-by`?

NukeWorker:
I tried, but I screwed it up. I don’t use git much.

 

Sorry.  Can you add the PR properly for me.

Sesquipedalian:
> Can you add the PR properly for me.

Done. 🙂

NukeWorker:
Thank you very much!

PR #9286, opened 2026-07-10
[2.1] Replaces backticks with shell_exec()

Sesquipedalian:
The backtick operator as an alias for shell_exec() has been deprecated in PHP 8.5.


Sesquipedalian:
Reimplements #9283

jdarwood007:
Just to comment, I checked for @` in the 3.0 source code, only comments found.

Sesquipedalian:
> Just to comment, I checked for @` in the 3.0 source code, only comments found.

I similarly did a regex search for all backtick quoted string and found only comments. With two slightly different searches both finding the same result, I think we can be confident in our conclusion.

PR #9305, opened 2026-07-25
Update `friendsofphp/php-cs-fixer` to `v3.95.17`

live627:
Includes my performance fixes (yes, I've decided to work with upstream instead of jumping ship)

- ref https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/releases/tag/v3.95.17


live627:
The CI check took 10 fewer seconds to run.

Sesquipedalian:
Nice!

PR #9306, opened 2026-07-25
40 files per process

live627:
Testing to see whether the overhead of creating processes in `php-cs-fixer` is greater than the time taken to run the fixers.


live627:
That cut the time nearly in half.

PR #9323, opened 2026-07-29
[3.0] Add AGENTS.md instructions for AI coding agents

albertlast:
> [!NOTE]
> **This change was produced by an LLM.** The file, the commit message and this
>
> Unlike the bug fixes in #9319–#9322, there is no runtime behaviour to verify here.
> What was checked is that every claim in the file was read out of this repository's
> own configuration rather than written from habit, and that the two new files do not
> disturb the existing checks. Please review it as untrusted work, and in particular
> as a proposal rather than a fix: whether SMF wants agent instructions in-tree at all
> is a project decision, not a defect.

#### Description

`AGENTS.md` is the cross-tool convention for giving AI coding agents project context.
GitHub Copilot reads it natively, as do Codex, Cursor, Gemini CLI, Aider, Windsurf and
Zed. Claude Code reads only `CLAUDE.md`, so that file is a two-line import:

```markdown
@AGENTS.md

## Claude Code
...
```

A symlink is the other documented option, but it needs Developer Mode or Administrator
on Windows, and `.gitattributes` already goes out of its way to keep Windows checkouts
working, so the import is the portable choice. The result is one file to maintain and
no duplicated instructions.

The contents are derived from this repository rather than from assumption:

| Source | What it contributes |
| --- | --- |
| `.editorconfig`, `.gitattributes` | tabs at width 4, LF, final newline; 2 spaces for YAML |
| `.php-cs-fixer.dist.php` | `@PER-CS2x0`, the `ordered_class_elements` order, `composer lint` / `lint-fix` |
| `.github/phpcs/SectionComments.php` | that the section banners are generated, and their names |
| `.github/workflows/*.yml` | what CI does and does not check |
| `composer.json` | PHP 8.4 platform, PSR-4 mapping, lint scripts |
| `DCO.txt`, `check-signed-off.php` | sign-off requirements |
| `.github/PULL_REQUEST_TEMPLATE/standard_pr.md` | PR body shape |
| `compose.yaml` | how to run the forum to check a change |
| `Sources/Db/Schema/v3_0/` | where to confirm a column still exists |

Two things are stated explicitly because a tool cannot infer them, and both were learned
the hard way while working on #9319–#9322:

- **There is no test suite.** CI proves the code parses and is formatted; it never
  executes SMF. An agent that treats four green checks as "this works" will be wrong,
  and all four of those bugs were runtime failures in code that parsed perfectly.
- **The sign-off check can pass on a pull request even when the commits are unsigned.**
  `check-signed-off.php` looks for `Signed-off-by:` on the last line, and when it does
  not find one it walks up to the parent commits — where `release-3.0` is signed off. Run
  against a feature branch directly it exits 1; run against `release-3.0` it exits 0. The
  file tells agents to verify locally rather than trust the badge.

The remaining sections cover the codebase traps that produced the recent run of bugs,
generalised rather than listed: typed properties that throw when read before their load
step, static properties shared with descendant classes, `empty()` hiding a null object,
and columns that migrations removed while a query still names them.

#### What was checked

- All four integrity scripts the `build` job runs pass with the files added:
  `check-smf-license`, `check-smf-index`, `check-smf-languages`, `check-version`.
- PHP-CS-Fixer over the whole tree: `Found 0 of 1725 files that can be fixed`. Markdown
  is outside the finder, so the new files are not linted, but nothing else shifted.
- Both files are LF with a final newline, matching `.editorconfig`, and together come to
  155 lines, within the 200-line guidance for instruction files.
- The only `@` in `AGENTS.md` is `` `@PER-CS2x0` `` inside a code span, so it is not
  mistaken for an import when Claude Code expands the file.

**Depends on #9317.** The "Verifying a change" section documents `docker compose up`,
the service ports and the two `docker exec` helpers, all of which come from the Docker
environment added there. If #9317 is not merged, that section describes something that
does not exist and should be dropped or rewritten before this lands. Merging #9317 first
is the natural order.

Related to #9319, #9320, #9321 and #9322 by origin rather than by content: the "Things
that bite in this codebase" section is a generalisation of those four root causes, and
the "no test suite" note is the reason each of them had to be verified by hand against a
running forum. No file overlap with any of them.

This PR touches only new files, so it conflicts with nothing.

1. Depends on: #9317
2. Related: #9319, #9320, #9321, #9322

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -83,22 +83,33 @@ param, throws, return.
 CI only proves that the code parses (`phplint` on 8.4 and 8.5) and is formatted
 correctly. It never executes SMF. Do not assume green checks mean a change works.
 
-So verify by running the forum. The repository ships a Docker environment:
+So verify by running the forum. The repository ships a Docker environment, documented
+in full in `.docker/README.md`:
 
 ```bash
 docker compose up -d --build
 # http://localhost:8080          the forum (install.php on first run)
 # http://localhost:8081          Adminer
 # http://localhost:8025          Mailpit, catches all outgoing mail
+# localhost:3307                 MySQL 8.4
 # localhost:5433                 PostgreSQL 17
 ```
+
+Both database services always start. `SMF_DB_TYPE` decides which one the generated
+`Settings.php` points at, and it defaults to `mysql`. Because raw SQL has to work on
+both engines, verify anything touching SQL on the other one as well: delete
+`Settings.php` and `Settings_bak.php`, restart `web` with `SMF_DB_TYPE` set the other
+way, then reinstall.
 
 The checkout is bind-mounted into the web container, so edits are live with no rebuild.
 Useful while working:
 
 ```bash
-docker exec smf-dev-web-1 php -l /var/www/html/Sources/Whatever.php
-docker exec smf-dev-db-1 psql -U smf -d smf -c 'SELECT * FROM smf_log_errors ORDER BY id_error DESC LIMIT 5;'
+docker compose exec web php -l /var/www/html/Sources/Whatever.php
+
+# Whichever engine the forum is installed on:
+docker compose exec mysql mysql -usmf -psmf smf -e 'SELECT * FROM smf_log_errors ORDER BY id_error DESC LIMIT 5;'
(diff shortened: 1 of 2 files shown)


sbulen:
This is actually a helpful read for human coders, too...

PR #9326, opened 2026-07-29
[3.0][Testing] Add a PHPUnit suite for the parts that need no database

albertlast:
> [!NOTE]
>
> The suite was run rather than only reasoned about, on both PHP versions CI targets.
> Please review it as untrusted work, and as a proposal rather than a fix: whether SMF
> wants a test suite, and whether this is the right shape for one, is a project
> decision.

#### Description

SMF has no automated tests. CI proves that the code parses and that it is formatted;
it never executes anything. Every bug in the recent run — #9319, #9320, #9321, #9322,
#9324, #9325 — parsed cleanly and passed every check.

More of 3.0 is testable than the absence of a suite suggests. The 3.0 rewrite into
namespaced classes did most of the work already. `tests/bootstrap.php` defines the
constants `index.php` defines, points the autoloader at `Sources/`, and sets the paths
and default language that the Unicode and entity helpers use to find their data files.
That is the whole bootstrap: no `Settings.php`, no database, no request.

**108 tests, 157 assertions**, in well under a second.

| Test class | Covers |
| --- | --- |
| `UtilsTest` | `buildRegex`, entity-aware `strlen`/`substr`/`strpos`/`split`, `htmlTrim`, `truncate`, `shorten`, `normalize`, `convertCase`, `sanitizeChars`, `sanitizeEntities`, JSON |
| `IPTest` | v4 and v6 normalisation, family flags, binary and hex round trips, validity |
| `UrlTest` | component parsing, `toAscii`/`toUtf8`, IDN, scheme handling |
| `UuidTest` | v4 and v7, variant, nil, `compress`/`expand`, short form, strict parsing |
| `SecurityTest` | bcrypt round trip, salting, cost factor, generators |
| `SapiTest` | `canonicalPath`, memory size parsing |
| `TimeIntervalTest` | ISO 8601 round trip, calendar-aware `toSeconds` |
| `ActionTraitTest` | `load()` returning the right class, in both orders, two hierarchies |
| `CreatePostNotifyTest` | `getTimeOffset()`, including half and quarter hour zones |
| `PunycodeTest` | IDN round trip |

What is **not** covered is anything reaching `Config::$modSettings`, `User::$me` or
`Db::$db`, which is most of the forum. That boundary is deliberate and is stated in
`tests/bootstrap.php`: work needing those belongs in an integration suite running
against a real install, which the Docker environment in #9317 would be the natural
place to build. Production code should not be contorted to make it reachable from here.

`AGENTS.md`, added in #9323, said there was no test suite. It now describes the suite as
something to write for rather than something that merely exists: the expectation that a
change to reachable code carries a test in the same commit, the cases that are reachable
— pure helpers, value objects, class-level behaviour, protected helpers through
reflection, and `modSettings` keys a test sets and restores itself — each pointed at a
worked example in `tests/Unit/`, and the cases that are not. It also names the traps that
only show up once you write a test here: `beStrictAboutOutputDuringTests` turns a stray
`echo` into a failure, `failOnRisky` fails a test that asserts nothing, and the style
fixer sorts data providers below the public test methods rather than above them.

#### These are regression tests, not tests written to fit

Each set was checked against the unfixed code:

- Reverting only `Sources/ActionTrait.php` and `Sources/Tasks/CreatePost_Notify.php`:
  `Tests: 16, Assertions: 12, Errors: 11`. The four `ActionTraitTest` cases and all
  seven `CreatePostNotifyTest` cases fail; the `Utils` cases pass, because #9318 is
  already merged.
- Reverting only `Sources/Sapi.php` and `Sources/Url.php`: exactly six failures, three
  in `SapiTest` and three in `UrlTest`, and nothing else.

Two behaviours were found while writing this and are now fixed in #9324 and #9325.
Until they were fixed the suite described them in comments rather than asserting them,
since pinning a defect down in a test preserves it.

One behaviour is pinned down deliberately, in `testAnyFourByteStringIsReadAsAPackedAddress`:
`IP`'s constructor accepts the packed binary form and cannot tell it apart from any
other four or sixteen byte string, so `new IP('nope')` is `110.111.112.101` and passes
validation. That is genuine ambiguity in the input rather than a defect, and it is the
sort of thing worth having written down. It caught the test out first: the initial
invalid-input case was `'not an ip at all'`, which is exactly sixteen bytes and
therefore a valid IPv6 address.

#### How this was verified

- `vendor/bin/phpunit`: **OK (108 tests, 157 assertions)** on PHP 8.4.23 and on 8.5,
  the two versions the existing syntax check targets.
- All four integrity scripts in the `build` job still pass: `check-smf-license`,
  `check-smf-index`, `check-smf-languages`, `check-version`.
- PHP-CS-Fixer over the whole tree: `Found 0 of 1738 files that can be fixed`. The
  `tests/` directory is inside the fixer's finder, so the tests follow the same rules
  as the rest of the codebase.
- The forum still serves normally with the branch checked out.

Those runs cover the suite as it stands. The last commit on the branch touches `AGENTS.md`
only and changes no test, so the figures above still describe it, but it has not itself
been through a run.

Two practical notes for whoever reviews the plumbing:

- `tests/` is already in the license-header ignore list in BuildTools, so the test
  files need no header. `check-smf-index` does still walk the directory, so `tests/`
  and `tests/Unit/` carry the standard `index.php`.
- PHPUnit's cache is set to `.phpunit.cache` rather than the default under `cache/`,
  because `check-smf-index` reports a missing index file for it the moment anyone runs
  the tests locally. CI would not have caught that, since the jobs do not share a
  workspace.

Separately, the custom `SMF/section_comments` fixer is not attribute-aware: when the
first member of a group carries a PHP attribute, the banner is inserted between the
attribute and the declaration. `tests/Unit/CreatePostNotifyTest.php` orders a plain
method first to avoid it, with a comment saying why. Not addressed here, but worth
knowing as attributes become more common.

#### Coverage

Measured with pcov on PHP 8.4.23, against the whole of `Sources/`, which is what
`<source>` in `phpunit.xml.dist` declares:

```
Classes:  0.00% (0/597)
Methods:  1.11% (40/3596)
Lines:    0.31% (625/200176)
```

That headline number is deliberately not flattering. The denominator is the entire
codebase, and this suite only reaches the stateless part of it, so 0.31% is a fair
description of how much of SMF is under automated test today. It is reported here
rather than tuned, because a scope narrowed to "the files that happen to have tests"
would produce a much nicer figure that tells nobody anything.

Per class actually exercised:

| Class | Lines | Methods |
| --- | --- | --- |
| `Punycode` | 80.30% (159/198) | 21.43% (3/14) |
| `TimeInterval` | 64.14% (93/145) | 33.33% (2/6) |
| `Url` | 34.82% (78/224) | 16.67% (3/18) |
| `Uuid` | 34.15% (84/246) | 25.00% (6/24) |
| `ActionTrait` | 33.33% (4/12) | 20.00% (2/10) |
| `Sapi` | 20.00% (36/180) | 12.50% (2/16) |
| `Utils` | 19.18% (145/756) | 23.21% (13/56) |
| `IP` | 4.91% (11/224) | 22.22% (4/18) |
| `CreatePost_Notify` | 2.04% (8/393) | 16.67% (1/6) |
| `Security` | 1.64% (7/428) | 23.53% (4/17) |

Where the line percentage is low but the method percentage is not, the class is mostly
database-bound: `CreatePost_Notify` at 2% means `getTimeOffset()` is covered and the
other 385 lines need a running forum.

The number worth watching is not the total. It is whether a class that could be tested
has been.

Three practical notes:

- The workflow runs with `coverage: none`, so CI does not collect it. The suite is
  meant to stay fast, and there is no threshold to enforce against yet. Collecting it
  is a local, on-demand thing.
- Coverage needs a driver that is not installed by default. `pecl install pcov` in a
  throwaway `php:8.4-cli` container is enough.
- It also needs `memory_limit` well above the 128M default. Building the coverage
  target map over 200k lines exhausts that; 2G is comfortable.

Running it also turned up a defect in this branch, now fixed. `#[CoversClass]` on
`ActionTrait` is not a valid target, because it is a trait, and PHPUnit only says so
when coverage is actually collected. The suite passed on its own and failed all five
`ActionTraitTest` cases the moment anyone added `--coverage`:

```
Class SMF\ActionTrait is not a valid target for code coverage
```

It is `#[CoversTrait]` now, which is also why `ActionTrait` appears in the table above.

This branch is stacked. It contains, and therefore depends on:

- **#9323**, the `AGENTS.md` it amends, which in turn depends on **#9317** for the
  Docker environment it documents.
- **#9321** and **#9322**, whose fixes the `ActionTrait` and `CreatePost_Notify` tests
  require. Without them this branch is red by construction.
- **#9324** and **#9325**, likewise for the `Sapi` and `Url` tests.

The four fix PRs are all independent of this one and can be merged on their own. This
one should go last: until they land, the diff shown here is theirs as well as its own,
and rebasing on `release-3.0` afterwards leaves only `tests/`.

1. Depends on: #9317, #9321, #9322, #9323, #9324, #9325


live627:
> Separately, the custom `SMF/section_comments` fixer is not attribute-aware: when the first member of a group carries a PHP attribute, the banner is inserted between the attribute and the declaration

- https://github.com/SimpleMachines/SMF/pull/9265

albertlast:
When this and pr https://github.com/SimpleMachines/SMF/pull/9317 got merged, i would look into creating tests framework with database stuff

PR #9330, opened 2026-07-30
[2.1][Testing] Add a Docker development environment and a baseline forum for upgrade testing

albertlast:
> [!NOTE]
>
> Everything stated below was verified by running it rather than only reasoning about
> it — full installs from an empty database on both engines, and the committed dumps
> restored back over the forums that produced them. Even so, please review it as
> untrusted work, and as a proposal: whether SMF wants generated fixtures committed to
> a release branch is a project decision, not a technical one.

#### Description

**This is a proposal rather than a fix, and it is tooling only — no forum code
changes.** Please close it if a bundled dev environment and committed fixtures are not
wanted on the 2.1 branch.

Testing the 2.1 → 3.0 upgrade needs something nobody currently has: a **real 2.1 forum,
with content in it, that everyone can start from and that is identical every time**.
Building one by hand takes an afternoon of clicking through the installer and produces a
different forum each time, so upgrade work either happens against an empty schema —
which exercises almost none of the migrations — or against whatever private database the
person testing happens to own.

That starting point can only be built here. 3.0 cannot create a 2.1 forum; its installer
writes 3.0 schema. So this branch adds the environment and the builder, and commits the
result:

```sh
.docker/baseline/restore.sh --engine mysql --profile small
```

Roughly ten seconds later there is an installed SMF 2.1.7 forum with 400 members and
6,000 posts, on either engine, identical to the one anyone else restores. Point a 3.0
checkout at that database, run `upgrade.php`, and the migration has something to
migrate.

The environment half is the same one as #9317, ported to 2.1. Ports are offset
(8180/8181/8125/3407/5533) so both stacks can run at once, which upgrade testing needs.

#### What makes the scenario a complex one

A stock install exercises almost nothing: SMF 3.0 runs 87 `v2_1` migrations and 17
`v3_0` ones, and a fresh 2.1 database leaves most of the tables they read completely
empty. Content generation alone does not help much either — the upstream `Populate.php`
tool fills six tables, and hardly any migration touches them.

So the builder seeds the rest deliberately, and each script names the migrations it
exists to feed:

| Seeded | Why it matters to the upgrade |
|---|---|
| IPv4, IPv6 **and absent** addresses on messages, members, logins and every log table | one migration per IP-bearing column. On MySQL these are `VARBINARY(16)` and the conversion is real work; on PostgreSQL 2.1 already used `inet` and they should skip. Both engines are built, so both paths are covered |
| Custom profile fields **and their values** in 2.1's layout (`themes.variable = 'cust_<name>'`) | that indirection is the whole of `CustomFieldsPart1`–`Part3` |
| Calendar events — all-day, timed, multi-day, topic-linked — and holidays on the **1004 sentinel** | `HolidaysToEvents`, `RecurringEvents`, `EventUids`. A holiday dated any other year is a one-off, and the recurrence conversion never fires |
| Polls (one open, one expired) with votes, PMs with labels and a filing rule, drafts of both types | `PersonalMessageLabels`, `UserDrafts` and friends |
| Real attachments on disk **and** in the table: a normal one, a thumbnail linked to its parent, an avatar | `LegacyAttachments`, `AttachmentSizes`, `AttachmentDirectory` |
| Bans: IPv4 range, **IPv6 range**, email mask, member ban, plus logged hits | `Ipv6BanItem`, `Ipv6LogBanned` |
| Error, action, online, reported, flood, spider and search logs — some rows with backtrace and session, some deliberately without | the remaining IPv6 log migrations, `ErrorLogSession`, `SearchResultsPrimaryKey` |
| An installed package row, legacy settings (`karmaMode`, `mail_type`, `cookieTime`, `enable_mod_prefs`, `time_offset`), TFA secrets, non-zero time offsets in both signs | `PackageVersion`, `MailType`, `RemoveCookieTime`, `DropModPrefs`, `DropTimeOffset` are all no-ops unless the old values are present |
| Two **MyISAM** tables on MySQL | a 2.1 install on any modern server creates everything as InnoDB, so `ConvertToInnoDb` would otherwise find nothing to convert |
| A **utf8mb3** database, matching 2.1's own DDL | this is what makes it a genuine pre-utf8mb4 forum, and converting it is what 3.0's UTF-8 step does. Changing it to utf8mb4 would quietly delete the thing under test |
| Names differing only by case or homoglyph | `SpoofDetector` |

Also seeded, because the alternative is a forum nobody can browse: board access. 2.1's
`createBoard()` leaves `member_groups` empty, so every board the generator makes is
admin-only until something opens it up.

#### Sizes

| Profile | Members | Topics | Messages | Dump per engine | Committed |
|---|---:|---:|---:|---:|:-:|
| `tiny` | 50 | ~150 | 600 | ~0.9 MB | yes |
| `small` | 400 | ~1,200 | 6,000 | ~8 MB | yes |
| `medium` | 2,000 | ~8,000 | 40,000 | ~70 MB | no |
| `large` | 20,000 | ~40,000 | 250,000 | ~450 MB | no |

`tiny` is for a quick upgrade run, `small` for a realistic one. `medium` and `large`
exist as profiles but are gitignored by directory name, so producing one cannot
accidentally add half a gigabyte to the repository.

**The committed artifacts total about 17 MB**, which is the main thing to weigh in
reviewing this. If that is too much for a release branch, the builder works without them
— `make-baseline.sh` regenerates everything in one command — and the artifacts can
simply be dropped from the branch.

#### Installing without a browser

2.1's installer turns out to be drivable with `curl`: it reads its step from `?step=`,
runs steps in order within one request, and stops at the first that needs input. There
is no CSRF token and no server-side progress. Three details are not obvious, and are
documented in `install-forum.sh`:

- **`DatabasePopulation()` returns `false` even when it succeeds** — it pauses to show
  its "N duplicate tables ignored" report, and that form posts `pop_done` to skip past
  it. That is why the sequence is five requests, not four.
- **`dbsession` and `reg_mode` must travel with `boardurl`**, because forum settings
  returning true falls straight through into database population in the same request.
- **`DeleteInstall()` deletes nothing.** The congratulations page has a checkbox that
  points the browser at `?delete`; that request is what removes `install.php` and the
  two schema files.

The driver never sends the `stats` field, which would register the site with
simplemachines.org.

#### Populate.php is fetched, not vendored

Content comes from `Populate.php` in
[SimpleMachines/tools](https://github.com/SimpleMachines/tools). It is downloaded at a
pinned commit and checksummed rather than committed here: it is third-party licensed
(MPL 1.1, with a BSD lorem ipsum generator) and does not belong in an SMF release
branch. Four patches adapt it, each asserting it matched exactly once, so an upstream
change breaks the build loudly instead of quietly producing a different baseline.

Two of those are worth flagging upstream: `makeBoards()` and `makeMessages()` both reach
`mt_rand(1, 0)` on their first item, which PHP 8 rejects outright where PHP 7 silently
swapped the arguments. **`Populate.php` cannot run at all on PHP 8 unpatched.**

#### It does not disturb the existing CI checks

Everything lives under `.docker/`, which the integrity scripts skip. Every PHP file added
is **7.1-syntax-clean**, since the lint job runs `phplint` over the whole checkout from
7.1 through 8.4 and it is not worth betting on whether it descends into dot directories —
verified by running `php -l` over all of them under both `php:7.1-cli` and `php:8.4-cli`.
`shellcheck -S warning` is clean on every shell script. No file under `.github/` is
touched.

Outside `.docker/` and `compose.yaml` the only tracked changes are `.gitignore` and
`.gitattributes`: the former because 2.1 ignores `*.sql`, `*.gz` and `attachments/` and
would otherwise swallow the entire deliverable, the latter to mark the dumps `-text
-diff` so line-ending rewriting cannot corrupt their hex-encoded IP columns.

#### Verification

Both engines, both profiles, built from an empty database and then restored from the
committed files:

- MySQL 8.4 and PostgreSQL 17, PHP 8.2, full headless install each time.
- Every dump restored back over the database that produced it, with checksums matching
  the manifest and all 41 recorded table counts matching.
- `check-coverage.php` — **78 assertions that the columns the migrations read are
  actually filled**, not merely that rows exist — passes on both engines.
- The restored forum browses: index, boards, topics and calendar all return 200, with
  zero deprecation notices in the response body.

That coverage check exists because it was needed. The custom field *definitions* were
silently absent on PostgreSQL for several builds: `db_insert()` builds its `ON CONFLICT`
clause only from key columns that also appear in the column list, and `id_field` is a
sequence the caller does not supply, so PostgreSQL received `ON CONFLICT () DO NOTHING`,
rejected it as a syntax error, and SMF logged that rather than raising it. Row counts
looked healthy — a stock install ships four custom fields of its own — and only reading
the data showed it. `make-baseline.sh` now runs the check before dumping.

#### Notes for review

- Credentials are `admin` / `baseline` and `smf` / `smf` / `smf`, deliberately trivial
  and published; the whole stack is throwaway and binds only to localhost. Generated
  members cannot log in at all, and every address is `@example.com`, which RFC 2606
  reserves.
- `$auth_secret` and `$image_proxy_secret` are **pinned** by the builder. They live in
  `Settings.php` and not in the database, so a restored dump paired with fresh random
  secrets would invalidate every session and two-factor backup code inside it.
- MySQL is pinned to 8.4 and must not float to 9: MySQL 9 removes the `utf8`/`utf8mb3`
  aliases 2.1's DDL is written against.
- Regenerating writes a **new** artifact directory and bumps `CURRENT`; it never
  overwrites, so a test can pin a revision.
- `.docker/baseline/README.md` documents the cross-branch recipe for consuming these from
  a 3.0 branch — `git archive` from a pinned commit, so the artifacts have exactly one
  home and no copy can drift.

1. Related: #9317 (the 3.0 Docker environment this is the 2.1 counterpart of)


albertlast:
something i ever wanted,
a source for migration test,
so maybe a little bit confusing, why this pr target branche 2.1,
but in his core of motivation it's a feature for 3.0.
since the biggest userbase wouldn't be install 3.0 fresh, they would like to migrate from 2.1,
this pr build the ground work for migration test from 2.1 to 3.0.

albertlast:
note for my futue me,
i don't like that the admin account is factor enabled, it make acessing the admin account harder, also since it not mention,
on the other hand i like that admin account got als two factor enabled for testing the upgrade process since it provide more complexity

PR #9333, opened 2026-08-01
[3.0] Theme split (wave 1, part 2) — List the maintenance tasks as one pick-one form

albertlast:
Wave 1, part 2 of breaking up #7933. Touches one theme template, and it is the one the change is about.

## What changes

`template_maintain_routine()` and `template_maintain_database()` each hand-wrote a heading, a form and a submit button for every task. Adding a maintenance task meant editing the template. Both are replaced by one `template_maintain_options()`:

```php
foreach (Utils::$context['options'] as $activity => $option) {
	// radio + maintain_<activity> + maintain_<activity>_info
}
```

`Utils::$context['options']` is built from the sub-action's own `activities` list, so the form and the dispatcher cannot drift apart. A task whose strings are not named `maintain_<activity>` supplies its own `title` and `info`, and `after` appends markup below the description. The shared wrapper and the "task finished" notice move into a `maintain` template layer.

## Fixes the reported template error

@dragomano reported this on #7933:

> Unable to load the '' template at Forum Maintenance - Routine

`routine` and `database` are the two sub-actions with no `'template'` key, and the dispatcher did

```php
Utils::$context['sub_template'] = !empty(self::$subactions[...]['template']) ? ... : '';
```

so they asked for a sub-template named `''`. They now default to `maintain_options`.

## Language strings

- `maintain_errors` → `maintain_repair`, so it matches its activity name. Its only consumer was the template being replaced.
- `maintain_cache` keeps its name — `Admin.template.php` and `cleanCache()` also use it — so `cleancache` supplies its title and info explicitly. Same for `convertentities`, since `template_convert_entities()` still uses the `entity_convert_*` strings.
- Adds `text_title`. Converting the messages body column is offered in both directions now, and the TEXT direction had no heading.
- Adds `convert_to_suggest_text`, shown when the maximum message length is at or above 65535, since shrinking the column would truncate posts members can currently make.

## Note on scope

#7933 also converts the board picker on the Topics tab to `<details>` plus a shared `template_choose_boards()` helper. I left that out, for two reasons: the helper lives in `GenericControls.template.php`, which belongs to the editor part of the split, and it emits `name="brd[]"` while `TopicRemove::old()` reads `$_POST['boards']` as an associative array — so on that branch, pruning by selected boards would not do what the form says. That conversion should come with the part that owns the helper, with the field name sorted out.

## Testing

- Routine tab: all six tasks render with real strings, no unresolved language keys.
- Database tab on PostgreSQL: `optimize` and `convertentities` only, `convertmsgbody` correctly absent.
- Ran "Recount all forum totals and statistics" through the new form — task executes, the confirmation notice renders, and `#manage_maintenance` appears exactly once.
- Members, Topics and Integration Hooks tabs all still render; they do not use the new layer.
- 108/108 unit tests pass.

### DIFF
--- a/Languages/en_US/ManageMaintenance.php
+++ b/Languages/en_US/ManageMaintenance.php
@@ -162,8 +162,8 @@ $txt['errortype_login_desc'] = 'Errors caused by failed login attempts or brute 
 
 $txt['maintain_recount'] = 'Recount all forum totals and statistics';
 $txt['maintain_recount_info'] = 'Should the total replies of a topic or the number of PMs in your inbox be incorrect: this function will recount all saved counts and statistics for you.';
-$txt['maintain_errors'] = 'Find and repair any errors';
-$txt['maintain_errors_info'] = 'If, for example, posts or topics are missing after a server crash, this function may help in finding them again.';
+$txt['maintain_repair'] = 'Find and repair any errors';
+$txt['maintain_repair_info'] = 'If, for example, posts or topics are missing after a server crash, this function may help in finding them again.';
 $txt['maintain_logs'] = 'Empty out unimportant logs';
 $txt['maintain_logs_info'] = 'This function will empty out all unimportant logs. This should be avoided unless something is wrong, but it does not hurt anything.';
 $txt['maintain_cache'] = 'Empty SMF’s cache';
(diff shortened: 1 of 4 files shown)


live627:
- Some code that was removed in  https://github.com/SimpleMachines/SMF/pull/8787 has resurfaced.

albertlast:
should be better now

live627 (approved):

PR #9334, opened 2026-08-01
[3.0] Theme split (wave 1, part 3) — Drop the unused jQuery Custom Scrollbar

albertlast:
Wave 1, part 3 of breaking up #7933. No theme templates touched.

`Theme::loadJavaScript()` loads `jquery.custom-scrollbar.js` and its stylesheet on every page for every logged-in member:

```php
if (!User::$me->is_guest) {
	self::loadJavaScriptFile('jquery.custom-scrollbar.js', ['minimize' => true], 'smf_jquery_scrollbar');
	self::loadCSSFile('jquery.custom-scrollbar.css', ['force_current' => false, 'validate' => true], 'smf_scrollbar');
}
```

Nothing calls `customScrollbar()`. Grepping `Sources/` and `Themes/default/` for `customScrollbar`, `custom-scrollbar`, `smf_jquery_scrollbar` and `smf_scrollbar` finds only the plugin itself and these two lines. So it is 942 lines of JS and CSS downloaded and parsed for nothing.

Removes both files, both loader calls, and the credit entry, since the library is no longer bundled. #7933 leaves the credit behind; that looks like an oversight rather than a decision.

Not included: #7933 also adds `defer` to `smf_jquery_plugins.js` and `theme.js` in the same block. That is a separate change with its own risks, so it is left for the part that deals with script loading.

## Testing

- Board index, a topic, profile, personal messages, credits and the admin centre all load with no scrollbar references in the markup, and no console errors.
- `jQuery.fn.customScrollbar` is `undefined` — confirming nothing was relying on it.


live627:
this related CSS rule can go

https://github.com/SimpleMachines/SMF/blob/fcbc3cba323f6e8dbbe1302fb329cf41b156d14e/Themes/default/css/index.css#L926-L928

live627 (approved):

PR #9346, opened 2026-08-02
[3.0][Testing] Make the code style check runnable locally

albertlast:
#### Description

Two related things, both about being able to reproduce a CI result before
pushing rather than after.

##### `composer lint` cannot run on the PHP this project supports

`.github/phpcs/SectionComments.php` called `array_first()`, which arrived in
**PHP 8.5**. `composer.json` pins the platform to **8.4.1**, and AGENTS.md tells
contributors to run `composer lint` before every commit.

So on the lowest PHP the project claims to support, the documented pre-commit
check could not run at all — it died with `Call to undefined function
array_first()` on every file, changed or not, so it was not obvious that the
problem was the tool rather than your branch. CI never noticed, because
`php-cs-fixer.yml` runs in a container that happens to ship 8.5.

`array_key_first()` has been available since 7.3 and says the same thing here:
what the call wants is the first *value* of an array keyed by token type.

Verified the whole tree is clean on both — 8.4 in the dev container (which could
not get past the first file before), and 8.5 in the same image CI uses.

##### `.docker/ci.sh`

Reproducing a CI failure otherwise means reading the workflow YAML and
reconstructing the invocation by hand. This runs the lot:

```sh
.docker/ci.sh              # everything CI checks
.docker/ci.sh --full       # style check over the whole tree, not just changes
.docker/ci.sh --fix        # apply the style fixes rather than reporting them
```

Mirrors `php.yml` (sign-off, the four integrity checks, phplint) and
`php-cs-fixer.yml`, plus the test suite when the branch has one. Every check runs
even after one fails and the failures are listed at the end, because finding out
about the second problem on the next push is the thing this exists to stop.

`--full` is the interesting flag. The style workflow only looks at the files a PR
changed — **except** when `composer.lock` or the fixer config is in the diff, in
which case it checks all seventeen hundred. So a branch that touches a dependency
silently inherits anything already non-compliant on `release-3.0`. Running
`--full` on this branch reports exactly one file, `Sources/PersonalMessage/Search.php`,
which nothing here touches and which #9345 fixes.

It deliberately does not pretend to cover the whole matrix: CI lints and tests on
8.4 **and** 8.5, and the container is whichever `PHP_VERSION` built it. The script
header and the README both say so, and give the one-line rebuild for the other.

##### Verified

- `.docker/ci.sh` exits 0 on a clean tree and 1 with a violation present —
  checked by planting one, including an **untracked** file, which an earlier
  draft of the changed-file detection missed.
- Full tree clean under php-cs-fixer on PHP 8.4 and on PHP 8.5.
- `shellcheck` clean on every script in `.docker/`.

#### Merge order

**Merge #9317 and #9344 before this one.** Both are contained in this branch, so the
diff shown here is theirs as well as its own; once they land and this is rebased on
`release-3.0`, what is left is `.docker/ci.sh` and the `array_first()` fix.

1. Depends on #9317 — `ci.sh` lives in `.docker/`.
2. Depends on #9344 — `ci.sh` sources `.docker/lib.sh`.
3. Related to #9345, which fixes the one pre-existing violation `--full` reports.


jdarwood007:
Not going to even review this until #9317 is merged.  I have no idea what this adds over the main docker support.

PR #9350, opened 2026-08-03
[3.0] Theme split (wave 2, part 1) — Add variables.css and move the base values into it

albertlast:
#### Description

Part of the #7933 split. Wave 1 was the six PRs that touched no theme templates. Wave 2 is the CSS foundation, and this is the first part of it.

`Theme::loadCss()` has been asking for `variables.css` since 3.0 Alpha:

```php
// Variables
self::loadCSSFile('variables.css', ['minimize' => true, 'order_pos' => -2], 'smf_variables');
```

The default theme has never shipped the file. A missing stylesheet is dropped without complaint, so nothing breaks and nothing happens — the call has been a no-op for as long as it has been there. This adds the file it was waiting for. No change to `Theme.php` is needed; `order_pos -2` already puts it ahead of `index.css` at `1`.

(`icons.css` is loaded the same way two lines above and is also missing. That one belongs to a later part of this wave.)

#### What is in it

The tokens `index.css` needs before it reaches any particular part of the forum:

| Group | Covers |
| --- | --- |
| `--html-*`, `--body-*` | page and body background, body colour, font family and size |
| `--selection-*` | `::selection` |
| `--input-*` | form controls in their normal, hover, focus and disabled states |
| `--fieldset-*` | fieldsets and their legends |
| `--strong-*`, `--heading-*` | `strong` / `.strong`, `h1`–`h6` |
| `--horizontalrule-*` | `hr` |

Names are the ones the new theme uses, so the parts after this one convert their own section of `index.css` without having to invent a vocabulary, and #7933's own `variables.css` drops onto this without renaming anything.

#### Every value is the one that rule already had

This changes no pixels, deliberately. Deciding what the forum should look like is a separate job from giving its current values names, and doing them together is a large part of why #7933 became hard to review. So the palette does not move here; a later part changes the values, and that diff will be nothing but values.

The substitutions are 1:1 and diffable by eye — 29 declarations:

```diff
 html {
-	background: #3e5a78;
+	background: var(--html-bg);
 	scroll-padding-top: 3rem;
 }
```

#### Testing

Fresh install on MySQL, Docker environment.

Computed styles captured before and after across eight pages — board index, a board, a topic, the posting form, search, a profile, an admin settings page and help — for the eighteen selectors these rules reach (`html`, `body`, `input`, `select`, `textarea`, `button`, `fieldset`, `fieldset legend`, `strong`, `h1`–`h3`, `hr`, `a`, `.button`, `input[disabled]`, `.windowbg`, `.cat_bar`) and the fourteen properties they set. **1316 comparisons, no differences.**

Also confirmed the tokens survive `minimize`: the generated bundle carries the `:root` block ahead of the rules that use it. Error log clean.

The unit suite does not reach CSS, so there is no test to add here.

Related to #7933

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -1,13 +1,13 @@
 /* You can find detailed information at https://wiki.simplemachines.org/smf/Curve2_CSS
 /* Index */
 html {
-	background: #3e5a78;
+	background: var(--html-bg);
 	scroll-padding-top: 3rem;
 }
 body {
-	background: #e9eef2;
-	font: 83.33%/150% "Segoe UI", "Helvetica Neue", "Nimbus Sans L", Arial, "Liberation Sans", sans-serif;
-	color: #4d4d4d;
+	background: var(--body-bg);
+	font: var(--body-font-size)/150% var(--body-font-family);
+	color: var(--body-color);
 	display: flex;
 	flex-direction: column;
 	min-height: 100vh;
@@ -15,8 +15,8 @@ body {
 }
 ::selection {
 	text-shadow: none;
-	background: #99d4ff;
-	color: rgba(0, 0, 0, 0.6);
+	background: var(--selection-bg);
+	color: var(--selection-color);
 }
 /* General reset */
 * {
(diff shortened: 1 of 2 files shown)


albertlast:

Sesquipedalian (approved):

Issue #346, opened 2013-04-30
Linktree - No URL Param


matthew-kerle:
Linktree items without a url param mess up the style due to the span that wraps it...should be a simple tweak to either the CSS or the markup that comprises it.

matthew-kerle:
When the Pull Request is merged, we should be all set to close this issue. I can't find any outstanding RTL issues in Curve II with this PR applied.

PR #222, opened 2012-10-23
! backport some SCeditor 1.4 fixes to 1.3.7

Spuds:
! Fix bug when pasting from Word
! Fixed bug causing scrollbar to change when inserting BBCode in source mode
! Fixed bug causing cursor position to be lost in FF when getting the editors value

Signed-off-by: Spuds spuds@simplemachines.org

PR #9351, opened 2026-08-03
[3.0] Theme split (wave 2, part 2) — Move the button values into variables.css

albertlast:
#### Description

Part of the #7933 split, wave 2 part 2. **Stacked on #9350** — that one adds `variables.css`; this one adds the next group of tokens to it. Until #9350 lands, this PR's diff shows both commits. Nothing else depends on the order.

The `.button` family. `index.css` styles buttons, the quickbuttons and the inline moderation checkbox through the same set of rules, so they move together:

| Rule | Tokens |
| --- | --- |
| `.button, .quickbuttons > li > a, .inline_mod_check` | `--button-color`, `--button-font-size`, `--button-text-transform`, `--button-cursor`, `--button-border-*`, `--button-box-shadow` |
| `.button:hover, .button:focus, .quickbuttons …` | `--button-border-color_hover`, `--button-box-shadow_hover` |
| `.button:hover, .button:focus` | `--button-color_hover` |
| `.button.active` | `--button-bg_active`, `--button-color_active`, `--button-font-weight_active`, `--button-border-color_active`, `--button-text-shadow_active` |
| `.button.active:hover, .button.active:focus` | `--button-color_active_hover`, `--button-box-shadow_active` |

Names again match the ones #7933 uses, and every value is the one the rule already had. Nothing changes on screen.

#### Two values left hard-coded on purpose

- `.pagesection .button { color: #346 }` is the body link colour, not a button colour. It happens to equal `--body-link-color`, and pointing it there would be a guess about intent rather than a rename.
- `color: #222` in the shared hover rule only ever reaches the quickbuttons: the very next rule overrides it for `.button`. There is no honest name for it in this vocabulary yet.

Both are better decided by whoever changes the palette than by a mechanical substitution.

#### Testing

Fresh install on MySQL, Docker environment. Two passes:

1. **Each token resolves to the literal it replaced.** Read all nineteen back from `:root` in the browser and compared against the strings they were lifted from — no mismatches. Since the substitutions are textually 1:1, that alone settles the normal case and the states that cannot be simulated.
2. **Computed styles before and after** for eleven button variants — plain, active, quickbutton, inline mod check, first and last in a `.buttonrow`, inside a `.cat_bar`, `.smalltext`, and the three of those that can hold focus — across twenty properties including all four border colours, the shadow and the text shadow. **220 comparisons, no differences.**

Related to #7933, #9350

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -1114,16 +1114,16 @@ a[class^="mobile_generic_menu_"] {
 .button, .quickbuttons > li > a, .inline_mod_check {
 	display: inline-block;
 	padding: 0 8px;
-	color: #444;
-	font-size: 0.7rem;
+	color: var(--button-color);
+	font-size: var(--button-font-size);
 	line-height: 2em;
-	text-transform: uppercase;
-	cursor: pointer;
+	text-transform: var(--button-text-transform);
+	cursor: var(--button-cursor);
 	min-height: calc(2em + 2em * (0.9 - 0.85)); /* "input" font size minus ".button" font size */
-	border: 1px solid;
-	border-color: #ddd #bbb #aaa #ddd;
-	border-radius: 3px;
-	box-shadow: 1px 1px 1px rgba(221, 221, 221, 0.57);
+	border: var(--button-border-width) var(--button-border-style);
+	border-color: var(--button-border-color);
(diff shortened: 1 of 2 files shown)


albertlast:

Sesquipedalian (approved):

PR #9352, opened 2026-08-03
[3.0] Theme split (wave 2, part 3) — Move the bar and frame values into variables.css

albertlast:
#### Description

Part of the #7933 split, wave 2 part 3. **Stacked on #9350 and #9351** — all three add to `variables.css`, so each branches from the one before and the diff shows the earlier commits until they land. Merge order is 9350, 9351, then this.

The containers every page is built out of:

| Rule | Tokens |
| --- | --- |
| `div.cat_bar`, `h3.catbg`, `.cat_bar .desc` | `--catbar-bg`, `--catbar-border-*`, `--catbar-box-shadow`, `--catbar-color`, `--catbar-font-family`, `--catbar-font-size`, `--catbar-text-shadow`, `--catbar-text-shadow_hover` |
| `.title_bar` | `--titlebar-bg`, `--titlebar-border-*`, `--titlebar-color` |
| `.sub_bar` | `--subbar-bg`, `--subbar-border-*`, `--subbar-box-shadow`, `--subbar-text-shadow` |
| `.roundframe` | `--roundframe-*` |
| `.information` | `--information-*` |
| `.windowbg` and friends | `--window-bg`, `--window-bg_hover`, `--window-border-*`, `--window-box-shadow`, `--window-odd-bg`, `--window-target-bg`, `--window-locked-bg`, `--window-sticky-bg`, `--window-stickylocked-bg`, `--window-approvetopic-bg`, `--window-approvepost-bg` |

#### One thing worth knowing for the parts that follow

Three of these rules set a border on a single side — `div.cat_bar` and `.sub_bar` have a `border-bottom`, `.information` has all four minus the top. Rewriting those as the `border-color` / `border-style` / `border-width` triple *looks* equivalent and is not:

```css
/* Not the same thing */
border-bottom: 1px solid #777;
border-color: #777; border-style: solid; border-width: 0 0 1px;
```

The second leaves `style` and `color` set on the sides whose width is zero, so a later rule that gives one of those sides a width paints a border that could not have appeared before. The single-side shorthand is kept.

This is not theoretical: the first attempt at these rules produced exactly that, and the check below caught 14 differences from it.

#### Two values left hard-coded

- The shared rule for `h3.titlebg, h4.titlebg, .titlebg, h3.subbg, h4.subbg, .subbg` sets one colour and one font family for **both** bars. Tokenising it means choosing which of the two names it belongs to, which is a palette decision rather than a rename.
- What #7933 calls the *even* row is this theme's *odd* one — today `even` shares the base background and `odd` is `#fdfdfd`. Rather than borrow `--window-even-bg` and have it mean the opposite, that value is `--window-odd-bg` here. Worth reconciling when the palette changes.

#### Testing

Fresh install on MySQL, Docker environment. Same two passes as the earlier parts:

1. **All 48 tokens resolve to the literal they replaced**, read back from `:root`. No mismatches.
2. **Computed styles before and after** for 22 container variants — cat bar plain and collapsed, its heading, its link, its description, title bar, titlebg, sub bar, subbg, roundframe, information, windowbg plain / locked / sticky / sticky+locked / approvetopic / approvepost, generic list wrapper, approvebg, `.bg.odd`, `.bg.even` — across 23 properties, with all four border widths and colours and each corner radius read separately. **506 comparisons, no differences.**

Related to #7933, #9350, #9351

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -3029,30 +3029,30 @@ dl.addrules dt.floatleft {
 }
 
 .generic_list_wrapper, .windowbg, .approvebg, .approvebg2 {
-	background: #f0f4f7;
+	background: var(--window-bg);
 	margin: 12px 0 0 0;
 	padding: 12px 16px;
-	border: 1px solid #ddd;
-	border-radius: 6px;
-	box-shadow: 0 -2px 2px rgba(0, 0, 0, 0.1);
+	border: var(--window-border-width) var(--window-border-style) var(--window-border-color);
+	border-radius: var(--window-border-radius);
+	box-shadow: var(--window-box-shadow);
 	overflow: auto;
 }
 /* Here comes the glory... */
 .windowbg:nth-of-type(even), .bg.even {
-	background: #f0f4f7;
+	background: var(--window-bg);
 }
 .windowbg:nth-of-type(odd), .bg.odd {
-	background: #fdfdfd;
+	background: var(--window-odd-bg);
 }
 
 /* Highlight the target item */
 .windowbg:target {
-	background: #ffffe0;
+	background: var(--window-target-bg);
 }
 
 /* Add some hover on table rows */
 tr.windowbg:hover {
-	background: #e2eef8;
+	background: var(--window-bg_hover);
 }
 
 /* Special treatment for #forumposts area */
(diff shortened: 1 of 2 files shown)


albertlast:

Sesquipedalian (approved):

PR #9353, opened 2026-08-03
[3.0] Theme split (wave 2, part 4) — Move the message box and bar values into variables.css

albertlast:
#### Description

Part of the #7933 split, wave 2 part 4. **Stacked on #9350, #9351 and #9352**, in that merge order — they all add to `variables.css`, so each branches from the one before and shows the earlier commits until they land.

The boxes SMF talks to people through, and the two bars:

| Rule | Tokens |
| --- | --- |
| `.errorbox` | `--errorbox-bg`, `--errorbox-border-*` |
| `.noticebox` | `--noticebox-bg`, `--noticebox-border-*`, `--noticebox-color` |
| `.infobox` | `--infobox-bg`, `--infobox-border-*`, `--infobox-color` |
| `.descbox` | `--descbox-border-*` |
| `.generic_bar`, its span, its inner `.bar`, the vertical variant | `--genericbar-*`, `--genericbar-inner-*` |
| `.progress_bar` and the yellow, green, red and blue variants | `--progressbar-*`, `--progressbar-inner-bg*` |

#### Notes

- The error, notice and info boxes each carry a border on the **top and bottom only**, so they keep their two single-side shorthands, for the reason #9352 ran into: writing them as the `border-color` / `border-style` / `border-width` triple leaves a style and a colour on the left and right sides, where today there is none, and a later rule giving those sides a width would then paint a border that could not have appeared before.
- `.errorbox` sets no colour of its own — it inherits. So there is no `--errorbox-color` here. Better to leave that visible than to invent a value for it.

#### Testing

Fresh install on MySQL, Docker environment. The same two passes as the earlier parts:

1. **All 40 tokens resolve to the literal they replaced**, read back from `:root`. No mismatches.
2. **Computed styles before and after** for 23 elements — each box, the error box's heading, its alert paragraph and its span, both bars, their inner `.bar`, the vertical variant and the four progress colours — across 27 properties, with every border side read separately. **621 comparisons, no differences.**

Related to #7933, #9350, #9351, #9352

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -3029,30 +3029,30 @@ dl.addrules dt.floatleft {
 }
 
 .generic_list_wrapper, .windowbg, .approvebg, .approvebg2 {
-	background: #f0f4f7;
+	background: var(--window-bg);
 	margin: 12px 0 0 0;
 	padding: 12px 16px;
-	border: 1px solid #ddd;
-	border-radius: 6px;
-	box-shadow: 0 -2px 2px rgba(0, 0, 0, 0.1);
+	border: var(--window-border-width) var(--window-border-style) var(--window-border-color);
+	border-radius: var(--window-border-radius);
+	box-shadow: var(--window-box-shadow);
 	overflow: auto;
 }
 /* Here comes the glory... */
 .windowbg:nth-of-type(even), .bg.even {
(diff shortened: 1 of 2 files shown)


albertlast:

Sesquipedalian (approved):

PR #9354, opened 2026-08-03
[3.0] Theme split (wave 2, part 5) — Move the quote and code block values into variables.css

albertlast:
#### Description

Part of the #7933 split, wave 2 part 5. **Stacked on #9350, #9351, #9352 and #9353**, in that merge order.

What BBC produces inside a post:

| Rule | Tokens |
| --- | --- |
| `blockquote` | `--quote-border-*`, `--quote-font-size` |
| `blockquote cite` | `--quote-cite-border-*`, `--quote-cite-font-size` |
| `.bbc_standard_quote`, `.bbc_alternate_quote` | `--quote-bg`, `--quote-alternate-bg` |
| `.bbc_code` | `--code-bg`, `--code-border-*`, `--code-font-size`, `--code-height` |
| `.bbc_link`, `.bbc_link:hover` | `--bbc-link-border-*` |

#### On the borders again

Unlike the rules in #9352 and #9353, `blockquote` and `.bbc_code` do carry a border on **every** side — they just use different widths and colours per side:

```css
border: 1px solid #d6dfe2; border-left: 2px solid #aaa; border-right: 2px solid #aaa;
/* is exactly */
border-color: #d6dfe2 #aaa; border-style: solid; border-width: 1px 2px;
```

So here the triple is the right shape and reproduces the shorthands exactly. `.bbc_code` likewise becomes `2px 1px 3px` in `#bbb #dfdfdf #aaa` — which happens to be the same shape #7933's `--code-border-width` uses. The genuinely single-sided borders, on `blockquote cite` and `.bbc_link`, keep their shorthand.

#### One value left hard-coded

`.codeheader` and `.quoteheader` share one rule and one set of values. Tokenising it means deciding whether it belongs to the code vocabulary or the quote one — #7933 has both, with identical values. Same situation as the title bar and sub bar headings in #9352, and left the same way, for whoever changes the palette.

#### Testing

Fresh install on MySQL, Docker environment. The same two passes:

1. **All 21 tokens resolve to the literal they replaced**, read back from `:root`. No mismatches.
2. **Computed styles before and after** for 11 elements — standard quote, alternate quote, a plain blockquote, the cite, a code block, an expanded code block, both headers, a BBC link plain and focused, and inline `.bbc_tt` — across 23 properties, every border side read separately. **253 comparisons, no differences.**

Related to #7933, #9350, #9351, #9352, #9353

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -3029,30 +3029,30 @@ dl.addrules dt.floatleft {
 }
 
 .generic_list_wrapper, .windowbg, .approvebg, .approvebg2 {
-	background: #f0f4f7;
+	background: var(--window-bg);
 	margin: 12px 0 0 0;
 	padding: 12px 16px;
-	border: 1px solid #ddd;
-	border-radius: 6px;
-	box-shadow: 0 -2px 2px rgba(0, 0, 0, 0.1);
+	border: var(--window-border-width) var(--window-border-style) var(--window-border-color);
+	border-radius: var(--window-border-radius);
+	box-shadow: var(--window-box-shadow);
 	overflow: auto;
 }
 /* Here comes the glory... */
 .windowbg:nth-of-type(even), .bg.even {
-	background: #f0f4f7;
+	background: var(--window-bg);
 }
 .windowbg:nth-of-type(odd), .bg.odd {
-	background: #fdfdfd;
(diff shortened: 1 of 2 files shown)


albertlast:

Sesquipedalian (approved):

PR #9355, opened 2026-08-03
[3.0] Theme split (wave 2, part 6) — Move the overlay and callout values into variables.css

albertlast:
#### Description

Part of the #7933 split, wave 2 part 6. **Stacked on #9350, #9351, #9352, #9353 and #9354**, in that merge order.

The things that sit on top of a page rather than inside it:

| Rule | Tokens |
| --- | --- |
| `#ajax_in_progress` and its link | `--ajaxinprogress-*`, `--ajaxinprogress-link-*` |
| `.popup_container`, `.popup_window`, `.popup_heading`, `.popup_content` | `--popup-bg`, `--popup-window-*`, `--popup-heading-color`, `--popup-content-color`, `--popup-content-line-height` |
| `.highlight` | `--searchhighlight-color`, `--searchhighlight-font-size`, `--searchhighlight-font-weight` |
| `a.moderation_link` | `--moderationlink-bg`, `--moderationlink-font-weight` |

#### One value that cannot be a token

`.highlight` sets `background-color: inherit`, and a custom property cannot hold that. `inherit` is a CSS-wide keyword, so

```css
:root { --searchhighlight-bg: inherit; }   /* asks :root to inherit from its parent */
```

has nothing to inherit from, resolves to the guaranteed-invalid value, and every `var()` reading it falls back to `transparent`. That is close enough to invisible to pass a screenshot and wrong wherever the highlight sits inside something coloured. It stays a literal here, with a comment saying why.

Worth flagging because #7933's own `variables.css` has `--searchhighlight-bg: inherit`, so the same declaration is waiting there.

#### Testing

Fresh install on MySQL, Docker environment.

1. **All 24 tokens resolve to the literal they replaced**, read back from `:root`. No mismatches. This is the check that caught the `inherit` problem — it came back as an empty string rather than the value it was given.
2. **Computed styles before and after** for nine elements — the AJAX bar and its link, popup container, window, large window, heading and content, the highlight and the moderation link — across 26 properties, every border side separately. **234 comparisons, no differences.**
3. For the `inherit` case specifically, a `.highlight` inside a parent painted `rgb(1, 2, 3)` still computes to `rgb(1, 2, 3)` and not `transparent`. The probe in point 2 could not have caught this on its own, since both answers look the same against a transparent parent.

Related to #7933, #9350, #9351, #9352, #9353, #9354

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -3029,30 +3029,30 @@ dl.addrules dt.floatleft {
 }
 
 .generic_list_wrapper, .windowbg, .approvebg, .approvebg2 {
-	background: #f0f4f7;
+	background: var(--window-bg);
 	margin: 12px 0 0 0;
 	padding: 12px 16px;
-	border: 1px solid #ddd;
-	border-radius: 6px;
-	box-shadow: 0 -2px 2px rgba(0, 0, 0, 0.1);
+	border: var(--window-border-width) var(--window-border-style) var(--window-border-color);
+	border-radius: var(--window-border-radius);
+	box-shadow: var(--window-box-shadow);
 	overflow: auto;
 }
 /* Here comes the glory... */
 .windowbg:nth-of-type(even), .bg.even {
-	background: #f0f4f7;
+	background: var(--window-bg);
 }
 .windowbg:nth-of-type(odd), .bg.odd {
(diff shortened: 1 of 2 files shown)


albertlast:

Sesquipedalian (approved):

PR #9357, opened 2026-08-03
[3.0] Theme split (wave 2, part 8) — Use logical properties for the settings lists

albertlast:
#### Description

Part of the #7933 split, wave 2 part 8 — the second RTL part. **Stacked on #9350 through #9356**, in that merge order.

`dl.settings` is the two-column label-and-value layout most of the admin and profile screens are built from. Its `clear`, the floats on `dt`, `dt.windowbg` and `dd`, and the gap after an image in a `dt` are all direction-dependent, and every one of them had an override in `rtl.css`:

```css
/* index.css */                  /* rtl.css */
dl.settings   { clear: right }   { clear: left }
dl.settings dt         { float: left }    { float: right }
dl.settings dt.windowbg{ float: left }    { float: right }
dl.settings dd         { float: right }   { float: left }
dl.settings img        { margin: 0 10px 0 0 } { margin: 0 0 0 10px }
```

Saying `inline-start` and `inline-end` removes the need for all five. `rtl.css` is down to **646** lines, from the 686 it started this wave with.

The image margin becomes `margin-block: 0; margin-inline: 0 10px`, which is the same declaration in both directions rather than two that have to be kept in step.

#### Testing

Fresh install on MySQL, Docker environment, with `lang_rtl` toggled on `en_US`. Geometry rather than computed styles, for the reason given in #9356: `getComputedStyle` returns logical keywords as authored and cannot tell you where anything actually landed.

Seven probes — the list itself, a `dt`, a `dt.windowbg`, both `dd`s, the image inside a `dt`, and a list clearing a float sitting beside it — each recording distance from both edges of a fixed 400px parent, vertical offset, width and height.

| | probes | differences |
| --- | --- | --- |
| LTR | 7 | 0 |
| RTL | 7 | 0 |

Nothing moves in either direction. Unlike #9356, this part has no behaviour change to report — the overrides it removes were complete, so the logical version reproduces them exactly.

Related to #7933, #9350, #9351, #9352, #9353, #9354, #9355, #9356

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -720,31 +720,32 @@ a.moderation_link, a.moderation_link:visited {
 /* Lists with settings use these a lot.
 ------------------------------------------------------- */
 dl.settings {
-	clear: right;
+	clear: inline-end;
 	overflow: auto;
 	margin: 0 0 10px 0;
 	padding: 5px;
 }
 dl.settings dt {
 	width: 56%;
-	float: left;
+	float: inline-start;
 	margin: 0 0 10px 0;
 	clear: both;
 }
 dl.settings dt.windowbg {
 	width: 98%;
-	float: left;
+	float: inline-start;
 	margin: 0 0 3px 0;
 	padding: 0 0 5px 0;
 	clear: both;
 }
 dl.settings dd {
 	width: 42%;
-	float: right;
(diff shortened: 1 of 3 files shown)


albertlast:

Sesquipedalian (approved):

PR #9359, opened 2026-08-03
[2.1] Respects literal `left`/`right` float direction for float BBC in RTL

Sesquipedalian:
The `[float]` BBCode takes a parameter to indicate whether it should float to the left or the right (i.e. `[float=left]` or `[float=right]`). End users writing their posts generally expect that the `left`/`right` parameter value will function as described regardless of the writing direction of the text in which the BBCode is inserted. This change ensures that the float still moves to the expected side even for RTL languages.

### DIFF
--- a/Themes/default/css/rtl.css
+++ b/Themes/default/css/rtl.css
@@ -4,8 +4,14 @@
 .floatright {
 	float: left;
 }
+.bbc_float.floatright {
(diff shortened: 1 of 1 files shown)


Sesquipedalian:
@jdarwood007 Tagging you in this so that you are aware of it.

PR #9365, opened 2026-08-04
[3.0] Improves SMF\TimeInterval thanks to property hooks

Sesquipedalian:
Previously, it was all but impossible to fully and reliably extend \DateInterval due to some strange quirks that it has. Most notably, it was extremely difficult to work with the days property in an extending class, because some sort of underlying magic would clobber anything the extending class tried to do with that property. However, with the advent of property hooks in PHP 8.4, it is now possible to create reliable workarounds for \DateInterval's quirky behaviour, so this PR does that.

### DIFF
--- a/Sources/Time.php
+++ b/Sources/Time.php
@@ -739,6 +739,19 @@ class Time extends \DateTime implements \ArrayAccess
 		return $this;
 	}
 
+	/**
+	 * Like \DateTimeInterface::diff(), except it returns SMF\TimeInterval
+	 * instances instead of \DateInterval instances.
+	 *
+	 * @param DateTimeInterface $target The date to compare to.
+	 * @param bool $absolute Whether the interval be forced to be positive.
+	 * @return TimeInterval Representation of the difference between the dates.
+	 */
+	public function diff(DateTimeInterface $target, bool $absolute = false): TimeInterval
+	{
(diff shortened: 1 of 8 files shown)


Sources/Time.php
```diff
@@ -739,6 +739,19 @@ public function setTimezone(\DateTimeZone|string $timezone): static
 		return $this;
 	}
 
+	/**
+	 * Like \DateTimeInterface::diff(), except it returns SMF\TimeInterval
+	 * instances instead of \DateInterval instances.
+	 *
+	 * @param DateTimeInterface $target The date to compare to.
```
jdarwood007, line 746:
```suggestion
	 * @param \DateTimeInterface $target The date to compare to.
```

Sources/Time.php
```diff
@@ -739,6 +739,19 @@ public function setTimezone(\DateTimeZone|string $timezone): static
...
+	 * @param bool $absolute Whether the interval be forced to be positive.
+	 * @return TimeInterval Representation of the difference between the dates.
+	 */
+	public function diff(DateTimeInterface $target, bool $absolute = false): TimeInterval
```
jdarwood007, line 750:
```suggestion
	public function diff(\DateTimeInterface $target, bool $absolute = false): TimeInterval
```

jdarwood007 (commented):
The suggested changes should resolve #9367

PR #9372, opened 2026-08-05
[3.0] Theme split (wave 3, part 4) — Give the width-limited wrappers a shared class

albertlast:
#### Description

Part 4 of wave 3 of the #7933 split. Stacked on #9371, #9370 and #9369 — the diff of this one is its last commit.

The rule that keeps the page centred named four wrappers in a single selector:

```css
#top_section .inner_wrap, #wrapper, #header, #footer .inner_wrap {
	max-width: 1200px;
	margin: 0 auto;
	width: 90%;
}
```

Anything else wanting the same treatment had to be added to that list, and the same list is repeated in `responsive.css` for the narrow override. It becomes a `.content-wrapper` class instead, and the width becomes `--wrapper-width`. This is what the rest of the header work hangs off: once the wrappers are collapsed, the elements that need the width limit are not the ones named above.

Still no visual change. This is the second of the two structural steps, and the reparenting itself comes after it.

#### Three templates, not one

`Themes/default/MaintenanceTemplate.php` and the XSLT in `Sources/Tasks/ExportProfileData.php` each build their own markup but load this same `index.css`, so their `#header`, `#wrapper` and `.inner_wrap` elements were being centred by that selector too. They get the class as well; without it the upgrade and maintenance pages and every exported profile would lose the width limit and run the full width of the window.

The maintenance page's own `#footer` has no `.inner_wrap` and so was never centred by this rule. That is left as it is.

#### Verification

Board index captured on the parent branch and again here, at two viewport widths so that both sides of the 720px breakpoint are exercised — the wide one is where `max-width: 1200px` bites and the narrow one is where `responsive.css` overrides it to `width: 100%`. The whole DOM is walked, recording tag, `getBoundingClientRect` and 25 computed properties per element, with `max-width` among them since that is the property that would break if the class failed to apply:

```
index  (innerWidth 1401): .content-wrapper 0 -> 4    116 elements, 0 style/geometry differences
narrow (innerWidth  701): .content-wrapper 0 -> 4    116 elements, 0 style/geometry differences

compared 6960 values
style/geometry differences: 0
```

`composer lint` cannot run against the two files with classes in this environment — the custom `SMF/section_comments` fixer calls `array_first()`, which needs PHP 8.5, and the container is on 8.4. `php -l` passes on all three files and CI covers the rest.

Related to #7933.

### DIFF
--- a/Sources/Tasks/ExportProfileData.php
+++ b/Sources/Tasks/ExportProfileData.php
@@ -135,7 +135,7 @@ class ExportProfileData extends BackgroundTask
 						</head>
 						<body>
 							<div id="footerfix">
-								<div id="header">
+								<div id="header" class="content-wrapper">
 									<h1 class="forumtitle">
 										<a id="top">
 											<xsl:attribute name="href">
@@ -145,7 +145,7 @@ class ExportProfileData extends BackgroundTask
 										</a>
 									</h1>
 								</div>
-								<div id="wrapper">
+								<div id="wrapper" class="content-wrapper">
 									<div id="upper_section">
 										<div id="inner_section">
 											<div id="inner_wrap">
@@ -167,7 +167,7 @@ class ExportProfileData extends BackgroundTask
 								</div>
 							</div>
 							<div id="footer">
-								<div class="inner_wrap">
+								<div class="inner_wrap content-wrapper">
 									<ul>
 										<li class="floatright">
 											<a>
--- a/Sources/Tasks/ExportProfileData.php
+++ b/Sources/Tasks/ExportProfileData.php
@@ -135,7 +135,7 @@ class ExportProfileData extends BackgroundTask
 						</head>
 						<body>
 							<div id="footerfix">
-								<div id="header" class="content-wrapper">
+								<div id="header" class="content_wrapper">
 									<h1 class="forumtitle">
 										<a id="top">
 											<xsl:attribute name="href">
@@ -145,7 +145,7 @@ class ExportProfileData extends BackgroundTask
 										</a>
 									</h1>
 								</div>
-								<div id="wrapper" class="content-wrapper">
+								<div id="wrapper" class="content_wrapper">
 									<div id="upper_section">
 										<div id="inner_section">
 											<div id="inner_wrap">
@@ -167,7 +167,7 @@ class ExportProfileData extends BackgroundTask
 								</div>
 							</div>
 							<div id="footer">
-								<div class="inner_wrap content-wrapper">
+								<div class="inner_wrap content_wrapper">
 									<ul>
 										<li class="floatright">
 											<a>
--- a/Themes/default/MaintenanceTemplate.php
+++ b/Themes/default/MaintenanceTemplate.php
@@ -52,11 +52,11 @@ abstract class MaintenanceTemplate
 	</head>
 	<body>
 		<div id="footerfix">
-		<div id="header">
+		<div id="header" class="content-wrapper">
 			<h1 class="forumtitle">', Maintenance::$tool->getScriptName(), '</h1>
 			<img id="smflogo" src="', Maintenance::$theme_url, '/images/smflogo.svg" alt="Simple Machines Forum" title="Simple Machines Forum">
 		</div>
-		<div id="wrapper">';
+		<div id="wrapper" class="content-wrapper">';
 
 		// Have we got a language drop down - if so do it on the first step only.
 		if (!empty(Maintenance::$languages) && \count(Maintenance::$languages) > 1 && Maintenance::getCurrentStep() == 0) {
--- a/Themes/default/MaintenanceTemplate.php
+++ b/Themes/default/MaintenanceTemplate.php
@@ -52,11 +52,11 @@ abstract class MaintenanceTemplate
 	</head>
 	<body>
 		<div id="footerfix">
-		<div id="header" class="content-wrapper">
+		<div id="header" class="content_wrapper">
 			<h1 class="forumtitle">', Maintenance::$tool->getScriptName(), '</h1>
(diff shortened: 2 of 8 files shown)


Sources/Tasks/ExportProfileData.php
```diff
@@ -135,7 +135,7 @@ class ExportProfileData extends BackgroundTask
 						</head>
 						<body>
 							<div id="footerfix">
-								<div id="header">
+								<div id="header" class="content-wrapper">
```
live627:
```suggestion
								<div id="header" class="content_wrapper">
```

Sources/Tasks/ExportProfileData.php
```diff
@@ -167,7 +167,7 @@ class ExportProfileData extends BackgroundTask
 								</div>
 							</div>
 							<div id="footer">
-								<div class="inner_wrap">
+								<div class="inner_wrap content-wrapper">
```
live627:
```suggestion
								<div class="inner_wrap content_wrapper">
```

Sources/Tasks/ExportProfileData.php
```diff
@@ -145,7 +145,7 @@ class ExportProfileData extends BackgroundTask
 										</a>
 									</h1>
 								</div>
-								<div id="wrapper">
+								<div id="wrapper" class="content-wrapper">
```
live627:
```suggestion
								<div id="wrapper" class="content_wrapper">
```

Themes/default/css/index.css
```diff
@@ -1207,9 +1207,11 @@ html[lang="el-GR"] .inline_mod_check {
 	box-shadow: 0 2px 3px rgba(0, 0, 0, 0.14);
 }
 
-/* Set maximum width limit for content */
-#top_section .inner_wrap, #wrapper, #header, #footer .inner_wrap {
-	max-width: 1200px;
+/* Set maximum width limit for content.
+/* Anything that should line up with the rest of the page gets this class,
+/* rather than each wrapper being named individually in the selector. */
+.content-wrapper {
```
live627:
```suggestion
.content_wrapper {
```

Themes/default/css/responsive.css
```diff
@@ -156,7 +156,7 @@
 	#footer {
 		padding: 10px;
 	}
-	#top_section .inner_wrap, #wrapper, #header, #footer .inner_wrap {
+	.content-wrapper {
```
live627:
```suggestion
	.content_wrapper {
```

Themes/default/index.template.php
```diff
@@ -231,7 +231,7 @@ function template_body_above()
 	// Wrapper div now echoes permanently for better layout options. h1 a is now target for "Go up" links.
 	echo '
 	<div id="top_section">
-		<div class="inner_wrap">';
+		<div class="inner_wrap content-wrapper">';
```
live627:
```suggestion
		<div class="inner_wrap content_wrapper">';
```

Themes/default/index.template.php
```diff
@@ -427,8 +427,8 @@ function template_body_above()
 		', empty(Theme::$current->settings['site_slogan']) ? '<img id="smflogo" src="' . Theme::$current->settings['images_url'] . '/smflogo.svg" alt="Simple Machines Forum" title="Simple Machines Forum">' : '<div id="siteslogan">' . Theme::$current->settings['site_slogan'] . '</div>', '';
 
 	echo '
-	</div>
-	<div id="wrapper">
+	</header>
+	<div id="wrapper" class="content-wrapper">
```
live627:
```suggestion
	<div id="wrapper" class="content_wrapper">
```

Themes/default/index.template.php
```diff
@@ -418,7 +418,7 @@ function template_body_above()
 	</div><!-- #top_section -->';
 
 	echo '
-	<div id="header">
+	<header id="header" class="content-wrapper">
```
live627:
```suggestion
	<header id="header" class="content_wrapper">
```

Themes/default/index.template.php
```diff
@@ -498,15 +498,15 @@ function template_body_above()
...
 		</div><!-- #content_section -->
 	</div><!-- #wrapper -->
 </div><!-- #footerfix -->';
 
 	// Show the footer with copyright, terms and help links.
 	echo '
-	<div id="footer">
-		<div class="inner_wrap">';
+	<footer id="footer">
+		<div class="inner_wrap content-wrapper">';
```
live627:
```suggestion
		<div class="inner_wrap content_wrapper">';
```

Themes/default/MaintenanceTemplate.php
```diff
@@ -52,11 +52,11 @@ public static function header(): void
 	</head>
 	<body>
 		<div id="footerfix">
-		<div id="header">
+		<div id="header" class="content-wrapper">
 			<h1 class="forumtitle">', Maintenance::$tool->getScriptName(), '</h1>
 			<img id="smflogo" src="', Maintenance::$theme_url, '/images/smflogo.svg" alt="Simple Machines Forum" title="Simple Machines Forum">
 		</div>
-		<div id="wrapper">';
+		<div id="wrapper" class="content-wrapper">';
```
live627:
```suggestion
		<div id="wrapper" class="content_wrapper">';
```

Themes/default/MaintenanceTemplate.php
```diff
@@ -52,11 +52,11 @@ public static function header(): void
 	</head>
 	<body>
 		<div id="footerfix">
-		<div id="header">
+		<div id="header" class="content-wrapper">
```
live627:
```suggestion
		<div id="header" class="content_wrapper">
```

live627 (changes requested):
use snake_case  for CSS class names instead of kebab-case

live627 (approved):

PR #9405, opened 2026-08-08
[3.0] Let SMF\Time do arithmetic with an SMF\TimeInterval

### LINKED ISSUE #9384: [3.0] Calendar 500s on any forum with an event: SMF\TimeInterval is not a usable \DateInterval
### Summary

`SMF\TimeInterval` can no longer be passed to `\DateTime::add()` or `sub()`, so
**any forum with a calendar event returns a 500 on `?action=calendar`**, and the
event editor cannot be opened at all.

```php
(new DateTimeImmutable('2026-01-01'))->add(new SMF\TimeInterval('P1Y'));
// Error: Object of type SMF\TimeInterval (inheriting DateInterval) has not been
//        correctly initialized by calling parent::__construct() in its constructor
```

`SMF\Calendar\Event` and `EventOccurrence` do exactly this in eight places:

| file | line |
|---|---|
| `Sources/Calendar/Event.php` | 493, 883, 1125, 1311 |
| `Sources/Calendar/EventOccurrence.php` | 250, 373, 630 |
| `Sources/Actions/Calendar.php` | 681 |

### Reproducing it

On a clean install of `release-3.0` (`a7ac468b1`), MySQL, PHP 8.4.24:

```sql
INSERT INTO smf_calendar
  (id_board, id_topic, title, id_member, start_date, end_date, start_time,
   end_time, timezone, location, duration, rrule, rdates, exdates, sequence,
   uid, type, enabled)
VALUES
  (0, 0, 'Test event', 1, '2026-08-12', '2026-08-12', '14:00:00', '15:00:00',
   'UTC', 'Somewhere', 'PT1H', 'FREQ=YEARLY;COUNT=1', '', '', 0, 'test-1', 0, 1);
```

`?action=calendar;viewmonth;year=2026;month=8` → **HTTP 500**, and
`smf_log_errors` has the message above pointing at `Event.php:1125`.

With `cal_allow_unlinked` on, `?action=calendar;sa=post` fails the same way from
`Event.php:493`, so a new event cannot be created through the UI either. With it
off the calendar is fine only because it has no events to load.

### Cause

`7b31585f8` ("Improves SMF\TimeInterval thanks to property hooks", 2026-08-03)
moved the class to holding a private `\DateInterval $base` and exposing its
values through property hooks. Before that commit the constructor called
`parent::__construct()`; now it does not, so the object *is* an uninitialised
`\DateInterval` even though it *reports* correct values. PHP's own date
arithmetic reads the real object, not the hooks, and rejects it outright.

### Why I have not just sent a patch

Adding `parent::__construct()` back re-breaks the thing that commit was written
to fix. I tried it: initialising the parent makes PHP's internal read handler win
over the `days` hook, so `days` goes back to `false` for exactly the intervals the
commit taught it to report:

| duration | `$days` with hooks alone | `$days` after `parent::__construct()` |
|---|---|---|
| `P1D` | `1` | `false` |
| `PT1H30M` | `0` | `false` |

So the two requirements are in direct conflict, and picking between them is a
call about this class's design rather than a bug fix. The options I can see:

1. Expose the wrapped `\DateInterval` and pass that at the eight call sites.
   Keeps the hooks; but it is an API change, and any mod doing date arithmetic
   with a `TimeInterval` still breaks with no hint as to why.
2. Call `parent::__construct()` and find another way to keep `days` working.
3. Something that has not occurred to me — @Sesquipedalian, this is your design
   and you will see it faster than I will.

Happy to implement whichever you prefer.

A separate, unrelated bug in the same class — every fractional duration throwing —
is fixed in #9383.

Found while testing the calendar area for the #7933 split.

albertlast:
#### Description

**The calendar is a fatal error on a stock install.** Not on a forum that has had an event added to it — on every forum, because the installer seeds 28 holidays, and drawing any of them goes through `Calendar\Event::__set()`:

```php
$end = (clone $this->start)->add($this->duration);   // Event.php:1125
```

`$this->duration` is an `SMF\TimeInterval`, and PHP refuses it:

```
Object of type SMF\TimeInterval (inheriting DateInterval) has not been correctly
initialized by calling parent::__construct() in its constructor
```

This is the bug reported in #9384. It has been sitting there since, so here is a fix that does not require settling the design question first.

##### Why the constructor is not simply restored

PHP's date arithmetic reads a `\DateInterval`'s internal state directly, not the property hooks. `TimeInterval` keeps its values in a private `\DateInterval` and serves them through hooks precisely so that `$days` can work — as its own docblock says, that is the only reliable way to get a usable `$days` in a subclass.

Calling `parent::__construct()` fixes the arithmetic and breaks that, because PHP then serves **every** hooked property from the internal state and ignores the hooks. Confirmed on the image's PHP 8.4.24:

```php
class TI extends DateInterval {
	public mixed $days { get => $this->base->days; }   // base->days === 45
	private DateInterval $base;
	public function __construct(string $d) { …; parent::__construct($d); }
}

$ti = new TI('P45D');
// y=0 m=0 d=45 days=false      ← the hook is declared, and never consulted
// $dt->add($ti) → works
```

Assigning `$this->days` afterwards does not help either — it creates a deprecated dynamic property that shadows nothing, and reads still return `false`. `Calendar\Holiday.php:214` reads `->days` off a `Time::diff()` result, so losing it is not free.

##### What this does instead

`TimeInterval::toDateInterval()` hands out the instance it is wrapping, and `SMF\Time::add()` / `::sub()` unwrap before calling PHP:

```php
public function add(\DateInterval $interval): static
{
	parent::add($interval instanceof TimeInterval ? $interval->toDateInterval() : $interval);

	return $this;
}
```

Every caller that passes a `TimeInterval` — `Event.php` ×2, `EventOccurrence.php` ×3, `Actions/Calendar.php` — has an `SMF\Time` on the left-hand side, so they are all covered. A plain `\DateInterval` takes exactly the path it took before. `TimeInterval`'s public API is unchanged, and `$days` still works.

It is a fix at the boundary rather than in the type, and I would rather you had the choice: the alternative is to make `TimeInterval` a genuine `\DateInterval` and give up the hooked `$days`, which changes a public property's behaviour and is your call, not mine. This unblocks the calendar either way.

##### Checked

On the running forum, before: `action=calendar` returns the fatal error page, `Event.php:1125` in `smf_log_errors`. After: renders, and December 2026 marks the 21st and the 25th as `class="days windowbg holidays"`, November the 11th and the 26th. Nothing new in the error log.

Rebased on the merged #9383 — the fractional duration fix — which this needs to construct the seeded holidays in the first place.

Fixes #9384
Related to #7933

### DIFF
--- a/Sources/Calendar/RecurrenceIterator.php
+++ b/Sources/Calendar/RecurrenceIterator.php
@@ -339,7 +339,12 @@ class RecurrenceIterator implements \Iterator
 
 		// We were given a view duration.
 		if (isset($view)) {
-			$this->view_end = (clone $this->view_start)->add($view);
+			// $view_start is a plain \DateTime or \DateTimeImmutable here, so
+			// unlike everywhere else in the calendar it cannot unwrap an
+			// SMF\TimeInterval for itself.
+			$this->view_end = (clone $this->view_start)->add(
+				$view instanceof TimeInterval ? $view->toDateInterval() : $view,
+			);
 		}
 		// Rule contained an UNTIL value, so use that as our view end.
 		elseif (isset($this->rrule->until)) {
--- a/Sources/Time.php
+++ b/Sources/Time.php
@@ -740,6 +740,37 @@ class Time extends \DateTime implements \ArrayAccess
 	}
 
 	/**
+	 * Like \DateTime::add(), except that it also accepts the SMF\TimeInterval
+	 * instances that SMF\Time::diff() and the calendar hand around.
+	 *
+	 * Those are not usable by PHP's date arithmetic on their own, so they are
+	 * unwrapped first. See SMF\TimeInterval::toDateInterval().
+	 *
+	 * @param \DateInterval $interval The interval to add.
+	 * @return static reference to this object.
+	 */
+	public function add(\DateInterval $interval): static
+	{
+		parent::add($interval instanceof TimeInterval ? $interval->toDateInterval() : $interval);
+
+		return $this;
+	}
+
+	/**
+	 * Like \DateTime::sub(), except that it also accepts SMF\TimeInterval
+	 * instances. See self::add().
+	 *
+	 * @param \DateInterval $interval The interval to subtract.
+	 * @return static reference to this object.
+	 */
+	public function sub(\DateInterval $interval): static
+	{
+		parent::sub($interval instanceof TimeInterval ? $interval->toDateInterval() : $interval);
+
+		return $this;
+	}
+
+	/**
 	 * Like \DateTimeInterface::diff(), except it returns SMF\TimeInterval
 	 * instances instead of \DateInterval instances.
 	 *
--- a/Sources/TimeInterval.php
+++ b/Sources/TimeInterval.php
@@ -291,6 +291,23 @@ class TimeInterval extends \DateInterval implements \Stringable
 	}
 
 	/**
+	 * Returns the underlying \DateInterval instance.
+	 *
+	 * PHP's date arithmetic reads a \DateInterval's internal state directly
+	 * rather than going through the property hooks above, and this class never
+	 * calls parent::__construct(), so it cannot be handed to
+	 * \DateTimeInterface::add() or ::sub() itself. Anything doing date
+	 * arithmetic wants this instead. SMF\Time::add() and ::sub() call it for
+	 * you.
+	 *
+	 * @return \DateInterval The underlying \DateInterval.
+	 */
+	public function toDateInterval(): \DateInterval
+	{
+		return clone $this->base;
+	}
+
+	/**
 	 * Formats the object as a string so it can be reconstructed later.
 	 *
 	 * @return string A ISO 8601 duration string suitable for reconstructing


albertlast:
Pushed a second commit: the first one was incomplete.

`Calendar\RecurrenceIterator` holds its view start as a plain `\DateTime` or `\DateTimeImmutable`, depending on the recurrence type, rather than an `SMF\Time` — so it does not get the unwrapping that `SMF\Time::add()` does, and it threw the same error at `RecurrenceIterator.php:342`. **Posting a calendar event** (`?action=calendar;sa=post`) was still a fatal after the first commit. It now unwraps the view duration itself, with a comment saying why that one is different.

Found by re-sweeping the calendar's own pages after the first fix rather than stopping at `?action=calendar`. All of these now render with nothing in the error log:

```
?action=calendar
?action=calendar;viewlist
?action=calendar;viewweek
?action=calendar;year=2026;month=12
?action=calendar;sa=post
```

This is the leak in fixing it at the boundary rather than in the type, and it is worth being plain about: every `\DateTime`/`\DateTimeImmutable` that meets a `TimeInterval` needs the same unwrapping. I checked the rest — `Event.php` ×2, `EventOccurrence.php` ×3, `Actions/Calendar.php` all have an `SMF\Time` on the left, and the remaining `->add()`/`->sub()` calls in `RecurrenceIterator` pass `$this->frequency_interval`, which is a plain `\DateInterval`. So the set is closed today, but nothing stops it reopening.

Separately, `?action=admin;area=managecalendar` still fails after this — but on something unrelated: `Holiday.php:278`, *"Failed to parse time string (FREQ=YEARLY)"*. That is its own bug and I will raise it separately rather than widen this PR.

Sesquipedalian:
We're not going to use this solution. I will submit a PR soon that restores full compatibility of SMF\TimeInterval with \DateInterval.

Issue #9423, opened 2026-08-08
[3.0]: Theme header messed up


Oldiesmann:
### Basic Information

On the latest version of SMF 3, the theme header is messed up. The forum name and logo are shoved into the top bar where the username/"my messages"/"alerts" and search bar are.

### Steps to reproduce

1. Install latest GitHub version of 3.0
2.

### Expected result

Forum name and logo in light blue area between menu bars (screenshot from an older revision of 3.0):

<img width="1196" height="93" alt="Image" src="https://github.com/user-attachments/assets/ed661bbe-02e7-43ab-a8b4-dee232e39301" />

### Actual result

Forum name and logo shoved into top menu bar area (screenshot from latest revision):

<img width="1220" height="118" alt="Image" src="https://github.com/user-attachments/assets/3be923b1-6fc9-4431-bf83-4f80e5a8516f" />

### Version/Git revision

d41ccea

### Database Engine

All

### PHP Version

8.5.9

### Logs

```shell

```

jdarwood007:
Most likely related to the ongoing theme changes @albertlast has sent up.
I don't know if the current wave of changes will address or not yet.

live627:
C'est la vie.

albertlast:
Most of this is intended, but you did catch one real defect in it. Splitting the two apart:

**The merged header is the intended layout.** #9373 ("Moves the top bar into the header as a user panel") removed the separate `#top_section` band that used to sit above the header. Before it, that band held the user links, the language picker and the search box, which left `#header` holding nothing but the forum title and the logo. They are now one row: title at the start, user panel at the end. #9374 then made the main menu a full-width band under it, which is why the light-blue gap in your "expected" screenshot is gone.

That is the end state of #7933, not something introduced along the way. The theme branch has the same structure:

```php
// live627/SMF2.1:theme — index.template.php
<header id="header">
    <div class="content-wrapper">
        <h1 class="forumtitle">…</h1>
        <div class="user_panel">
```
```css
#header .content-wrapper { align-items: center; display: flex; justify-content: space-between; }
```

Both PRs are part of the #7933 split (wave 3, parts 5 and 6).

[...]

PR #9425, opened 2026-08-09
[3.0] Centres the forum title and logo in the header

albertlast:
#### Description

Two rules in `index.css` are leftovers from the header as it was before #9373, when it was a band of its own that bottom-aligned its contents:

```css
h1.forumtitle { padding: 22px 12px 6px 10px; }
img#smflogo   { margin: 16px 0 0 0; }
```

The header lines its contents up in the middle now (`#header .content_wrapper { align-items: center }`), so both of those push the title and the logo below the user panel sitting beside them. Measured centres on the board index at 1280px, before and after:

| | wrapper | title | logo | user panel |
|---|---|---|---|---|
| before | 47 | 47 | **55** | 47 |
| after | 47 | 47 | 47 | 47 |

The title keeps the horizontal padding that lines it up with the page content below it — as `padding-inline`, so the RTL sheet does not need its own rule. Its start edge stays at x=76 against the breadcrumb's 75. The logo needs nothing; RTL's own `img#smflogo { margin-right: 1em }` still applies.

Header height is unchanged at 93px either way — `--header-height: 80px` sets the floor.

This is the visible half of what #9423 reports. The rest of that report is the intended layout from #9373 and #9374, which merged the old `#top_section` into the header; that matches the end state of #7933 and is not changed here.

Related #9423
Related #7933

### DIFF
--- a/Themes/default/css/index.css
+++ b/Themes/default/css/index.css
@@ -1288,11 +1288,13 @@ html[lang="el-GR"] .inline_mod_check {
 	gap: .5em;
 	justify-content: flex-end;
 }
-/* The main title. */
+/* The main title. The header lines its contents up in the middle now, so the
+/* title only keeps the padding that lines it up with the page content below
+/* it. The old top-heavy padding pushed it below everything beside it. */
 h1.forumtitle {
 	font-size: 1.8em;
 	font-family: "Tahoma", sans-serif;
-	padding: 22px 12px 6px 10px;
+	padding-inline: 10px 12px;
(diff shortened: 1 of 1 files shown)


albertlast:
Superseded by #9427. That one puts the header back to a band of its own between the top bar and the menu, per the decision on #9423, so the row this centred things in no longer exists — the top-heavy padding on `h1.forumtitle` and the top margin on `img#smflogo` are correct again against `align-items: flex-end`. Nothing here to carry over.

PR #9448, opened 2026-08-09
[3.0] Theme split (wave 4, part 22) — point admin.css at the design tokens

albertlast:
#### Description

Where the stylesheets stand today:

| file | hard-coded colours | `var(--…)` uses |
|---|---|---|
| `index.css` | 181 | 227 |
| **`admin.css`** | **33** | **0** |
| `calendar.css` | 33 | 0 (#9443 converts these) |

`index.css` has had tokens applied to it for a while. `admin.css` was never touched, so restyling the admin centre means editing a stylesheet rather than a variable, and `dark.css` and the colour variants have nothing to reach for. This is the same treatment #9443 gives `calendar.css`, and the two are best judged together.

Two colours already had a token that means the right thing, so they point at it rather than gaining a duplicate:

- `.search_results li a strong` → `--body-link-color` (it is a link)
- `#manage_boards li.windowbg` borders → `--window-border-color` (the items carry `.windowbg`)

The other 28 get named tokens of their own under a new `/** Admin **/` block, each holding the value it has now.

Where one value appears twice under two different meanings it gets two tokens: `#d0e7f8` is both the hover on a board row and the `.highlight2` background, and `#cacdd3` borders both the package list and the permission board list. A restyle would very plausibly move one and not the other.

##### Verification

The acceptance test is that nothing changes on screen.

Every painting rule in the loaded stylesheet was applied to a probe element and its computed values read, before and after — **511 selectors, 9234 declarations**.

27 differed, all inside three rules that set a `border` shorthand and then override one side of it:

```css
.action_admin .table_grid td {
	border: 1px solid var(--admin-table-grid-border-color);
	border-top: 0;
}
```

That is a limitation of the measurement, not a regression: once the shorthand contains a `var()`, `CSSRule.style.cssText` cannot serialise it alongside the longhand override, so the probe received nothing to apply. Those three rules were measured instead on real elements — including the `:first-child` and `:last-child` variants that carry their own border rules — giving **6 elements × 12 border properties, 0 differences**.

`admin.css` also gains the trailing newline `.editorconfig` asks for, which is why the diff shows a closing brace as changed.

Related: #7933

### DIFF
--- a/Themes/default/css/admin.css
+++ b/Themes/default/css/admin.css
@@ -10,7 +10,7 @@
 	float: right;
 }
 .action_admin .table_grid td {
-	border: 1px solid #ddd;
+	border: 1px solid var(--admin-table-grid-border-color);
 	border-top: 0;
 }
 .action_admin .generic_list .flow_auto {
@@ -28,7 +28,7 @@ a.help span {
 	margin-right: 2px;
 }
 .table_caption, tr.table_caption td {
-	color: #000;
+	color: var(--admin-table-caption-color);
 	font-size: 10px;
 	font-weight: bold;
 }
@@ -43,7 +43,7 @@ fieldset dl {
 }
 legend {
 	font-weight: bold;
-	color: #000;
+	color: var(--admin-legend-color);
 }
 
 /* Styles for the admin home screen.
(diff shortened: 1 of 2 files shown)


albertlast:
Note for whoever merges these: this and #9443 both add a new section to the end of `variables.css` — `/** Admin **/` here, `/** Calendar **/` there — so they conflict with each other even though both merge cleanly against `release-3.0` on their own. The conflict is purely the two blocks landing at the same anchor; there is no overlap in what they define.

Whichever goes in first, ping me and I will rebase the other. They are independent otherwise and can be reviewed in either order.

